Issue #3379, opened 2016-07-03
BEP: Consider adding a field in Hello to indicate whether the device intends to keep the connection open


canton7:
Currently, there's no distinction between "Hello, I need to send this message but I'm intending to close the connection straight afterwards because I don't know you, or because I have you paused" and "Hello, all looks good from my side, let's sync".

This would then also be exposed in the Events API's Connected event.

It would be useful to know the difference: wrappers which generate a "Connected to Device" user notification would know not to show one for devices which are going to disconnect straight away, and the GUI could show a message other than "Disconnected".

calmh:
Sounds reasonable to me. I would propose something like

```
--- a/lib/protocol/bep.proto
+++ b/lib/protocol/bep.proto
@@ -18,6 +18,13 @@ message Hello {
     string           device_name    = 1;
     string           client_name    = 2;
     string           client_version = 3;
+    ConnectionStatus status         = 4;
+}
+
+enum ConnectionStatus {
+    ACCEPTED      = 0;
+    DENIED_AUTH   = 1;
+    DENIED_POLICY = 2;
 }
 ```

where `ACCEPTED` is what happens today (of course the connection can be closed at any point anyway), `DENIED_AUTH` specifically means "we don't like your device ID" and `DENIED_POLICY` means "we know who you are and would accept your connection, but for the moment we are not" perhaps due to pause status, resource shortage, you name it.

AudriusButkevicius:
So all sounds good, apart from this POLICY thing, we have enough enums to indicate every possible reason I think, or we could just send back a string.

calmh:
It needs an enum at least for machine decisions. Certainly we can add more enums at any point, but it's the usual "protocol vs one implementation" kind of thing. *We* don't have many different reasons, but the next implementation might have "denied due to time of day", "denied due to being on 3g and not 4g", etc

AudriusButkevicius:
Ok, but then let's add a string reason so that it's atleast explained.

lkwg82:
Is this still relevant?

AudriusButkevicius:
Yes

PR #3390, opened 2016-07-05
lib/protocol: Allow unknown message types

calmh:
### Purpose

This lets us add message types in the future, for authentication or
other purposes, without completely breaking old clients. I see this as
similar behavior to adding fields to messages - newer clients must
simple be aware that older ones may ignore the message and act
accordingly.

Related to, for example, #3379.

### Testing

None whatsoever. But it looks safe.

### DIFF
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -106,6 +106,7 @@ type rawConnection struct {
 	once        sync.Once
 	pool        sync.Pool
 	compression Compression
+	warnOnce    sync.Once
 }
 
 type asyncResult struct {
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Not sure if it should warn. To be honest, the more advanced peer should be aware which messages other side support from versiom defined in hello?

calmh:
It should, yes. Silent skip?

AudriusButkevicius:
Magic cookie would be a field in hello message most likely.

AudriusButkevicius:
I am not sure we should warn about this as this is something that is not necesserily actionable by the user. What do you expect the user to do when he sees this message?

calmh:
Upgrade, probably, but I'm absolutely fine with not even warning. But I do think we'll save ourselves a headache down the road by allowing unknown messages. Other mechanisms I've considered are ranges on the message type (types <256 are mandatory, understand them or close the connection; >=256 are optional and can be skipped if you don't understand) or an explicit flag in Header (type=47, mandatory=true => skipping this message is not allowed).

AudriusButkevicius:
So I am fine ignoring, just not sure about warning.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as e440d3002877a5fb7ae0ada5b5843b011b00c9bd. Thanks, @calmh!

PR #3411, opened 2016-07-18
lib/model: Sort outgoing index updates by LocalVersion

calmh:
### Purpose

Preparation for delta indexes. To review this part individually and not mix it up unnecessarily with the other changes. Adds some overhead to the initial index exchange, especially if the index is large, but not enough to affect the ManyFiles benchmark negatively in my testing.

### Testing

The index sorter is fully covered by unit testing apart from the `panic()`ing lines. I've run the benchmarks and those work, so indexes are being sent.

### Things to Review

I'm panicing in quite a few places in the sorter as I don't want to add error returns that would anyway be annoying to handle. I think the cases where there are panics are those where we would panic anyway - out of disk space, disk I/O failure, unable to marshal a struct, and so on. But I would appreciate an extra eye on it.

### DIFF
--- a/lib/model/sorter.go
+++ b/lib/model/sorter.go
@@ -1,0 +1,197 @@
+// Copyright (C) 2016 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package model
+
+import (
+	"encoding/binary"
+	"io/ioutil"
+	"os"
+	"sort"
+
+	"github.com/syncthing/syncthing/lib/protocol"
+	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/opt"
+)
+
+const (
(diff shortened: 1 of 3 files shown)


lib/model/sorter.go
```diff
@@ -0,0 +1,197 @@
...
+		if !fn(f) {
+			break
+		}
+	}
+}
+
+func (s *onDiskIndexSorter) Close() {
+	l.Debugf("onDiskIndexSorter %p closes", s)
+	s.db.Close()
+	os.RemoveAll(s.dir)
```
AudriusButkevicius: osutil?

calmh: :+1: fixd

AudriusButkevicius:
The build failed?

AudriusButkevicius:
Build machine oom?

calmh:

calmh:
I think the builder ran out of memory

calmh:
@AudriusButkevicius please take another look on this one :)

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 8ab6b60778c59b700c5e622b007cdd0669a9371b. Thanks, @calmh!

PR #3427, opened 2016-07-22
cmd/syncthing, lib/db, lib/model, lib/protocol: Implement delta indexes (fixes #438)

### LINKED ISSUE #438: Implement delta-only index exchange at connect
As stated in BEPv1, every connection has to start with an index exchange.
Every index exchange contains a BlockInfo structure for (almost) every 128KiB in the repository. A BlockInfo structure has a size of (currently) 40Bytes. This results in an initial transfer of data according to ~1/3276 of the repository size. That doesn't sound like a lot, but repositories can and will be quite large.

Isn't this a huge problem for the scalability of BEP? A repository with 4GiB of data in it will result in the transfer of roughly *a Megabyte* of information at the beginning of every connection. In *both directions*. Depending on the connection speed, this may take quite a bit of time (e.g., 5 seconds on ADSL2 at full speed).

One of the bigger problems of this is, that people don't expect to have a scalability issue with the size of their content. "It's only a few files ... why does it struggle with that?"

Of course, this isn't easily changeable in v1. But, perhaps for BEPv2, it would be necessary to look into tree hashes (e.g., Tiger Tree Hash as used in Bittorrent). That would reduce the overhead to a constant hash per file and not per block. Telling people that it takes a while because they have millions of files is easier than telling them, that it takes so long because they have one large file in their repository.

calmh:
### Purpose

This is the delta index implementation. There are a few main parts to it:

- The "IndexID" stuff which is to ensure that we can detect when a given LocalVersion counter has reset or doesn't mean what we think it means. It also tells us whether the other side actually implements this part of the protocol or not (IndexID==0 meaning no support).

- A tweak around how we handle incrementing LocalVersion in the database. I think this is simpler, and it's now just a regular incrementing counter per folder.

- How we send indexes is changed, and it's now triggered from inside the code that takes care of receiving the ClusterConfig. There are a number of checks for the announced IndexID and MaxLocalVersion and if it checks out we start sending index updates only from that MaxLocalVersion and onwards.

### Testing

I've ran the integration tests that are suitable, including the various "restart during transfer" tests and it checks out. I'm running it myself on my device since a week or so, and it works as far as I can tell.

### Docs

Spec update: https://github.com/syncthing/docs/pull/208. It's a backwards compatible change.


lib/db/set.go
```diff
@@ -95,42 +98,53 @@ func (s *sizeTracker) Size() (files, deleted int, bytes int64) {
...
 	s.db.checkGlobals([]byte(folder), &s.globalSize)
 
 	var deviceID protocol.DeviceID
 	s.db.withAllFolderTruncated([]byte(folder), func(device []byte, f FileInfoTruncated) bool {
 		copy(deviceID[:], device)
-		if f.LocalVersion > s.localVersion[deviceID] {
-			s.localVersion[deviceID] = f.LocalVersion
-		}
 		if deviceID == protocol.LocalDeviceID {
+			if f.LocalVersion > s.localVersion {
```
Zillode: `s.localVersion` is never set?

calmh: Hm? I set it on the line below your comment. :)

calmh: Ah you mean initialized perhaps? It's initialized to zero as we allocate the "s" struct above.

AudriusButkevicius:
@st-review merge

AudriusButkevicius:
Nothing to pick on, clear and understandable.

st-review:
:ok_hand: Merged as 47fa4b0a2cc2f9bb25965110649062e3e8fe1c9e. Thanks, @calmh!

calmh:
Lets hope it works

rumpelsepp:
Wow! That's really exciting! Updating now...

PR #3431, opened 2016-07-23
lib/protocol, lib/model: Implement high precision time stamps (fixes #3305)

### LINKED ISSUE #3305: Increase time stamp granularity
We currently only track, store and compare file time stamps with second granularity. But ntfs, ext4, apfs and so on store much more fine grained time stamps. I suggest that we add a nanoseconds field to the time stamp, and set time stamps to the highest precision possible. We should still have a (configurable?) comparison window, to allower comparisons to succeed on lower precision systems. Or this could be handled by the mtimefs.

calmh:
### Purpose

This renames the existing FileInfo.modified field to modified_s while retaining the same meaning (UNIX epoch seconds), and adds a new field int32 modified_ns that holds the nanoseconds within the second.

When scanning, modified_ns is populated from Stat() data when it's provided by the OS/filesystem. When syncing, we set the timestamp with the full precision.

This also adds a new advanced config option mtimeWindowS which is the time window within which we consider to timestamps equal. It defaults to one second which is the same behavior as before - a file with timestamp 1234567890.000000000 is considered the same as the same file with a timestamp 123456789.123456789.

The window is special as the purpose is to handle filesystems truncating timestamps - 1234567890.000000000 is not considered equal to 1234567889.999999999 with the default settings even though the time difference is only a nanosecond, as they become different when both
are truncated to the 1s window.

This config can be set to 2s to account for FAT32 shenanigans.

At least on Solaris+ZFS there's an interesting issue where we can read and understand nanosecond precision timestamps but when set by Chtimes the precision gets truncated to microseconds. Not sure if this is a Go bug or a Solaris bug or what, but the integration test accounts for it.

### Testing

The integration tests now expect syncing to fully copy the time stamps up to microsecond precision, instead of just the second value. This only means something on systems that keep that precision, but I tested it on Solaris and there it passes and manual inspection shows that the timestamps are copied.

### Documentation

- [ ] Must update the spec


AudriusButkevicius:
I think this is somewhat pointless.
The right fix to the problem is to defer scanning by one second after reading the timestamp, no?

calmh:
This is not to fix whatever issue that would fix, just to retain more metadata. It's kinda arbitrary to truncate all timestamps at the one second level when many or most systems can do better.

AudriusButkevicius:
Do we really need nanosecond predicion?

I was just thinking that we could fit the whole thing into one uint64 instead?

calmh:
We could just redefine the existing time stamp to be epoch nanoseconds and cover the relevant time period, but that would break compatibility. Hence the new subsecond field, and I don't really think it hurts to separate the two.

I'm not actually arguing we need nanosecond precision for anything (I certainly don't), mostly just that if the metadata is there in the filesystem there's no reason for us to intentionally truncate it while copying.

As a side effect, if `mtimeWindowS` is set to zero we can detect multiple modifications within the same second but I suspect there's other stuff that's racy enough that that doesn't matter...

Apparently `os.Chtimes` always truncates to microseconds at the moment, regardless if the underlying OS/FS supports better, but seems like something that might change in Go 1.8

AudriusButkevicius:
Is there any good reason why this shouldn't default to 0?

calmh:
I was thinking to retain the current behavior and avoid unnecessary "sync, discover changed mtime, rescan, sync changed mtime back to source" cycles when two devices have differing time stamp precision. 

But what we really should do is take stock of who supports what. Ntfs is high precision, and I think so is ext4, and those are probably the most common... Potentially we could also probe it and auto detect - set time stamp of .stfolder to something.99999999 and read it back to check.

calmh:
Hfs is unfortunately crap precision as far as I know, so Mac installations should somehow default to 1s. 

Actually I think probing when mtimeWindowS=0 might be the cleanest.

AudriusButkevicius:
Or we could double check when we adjust the time to verify where we are at, and use the virtual mtime thing.

calmh:
Yes, I think that's better. I tried some of the probing and it's nontrivial as `2016-07-25 10:00:11.111111111 +0200 CEST` gets rounded to `2016-07-25 10:00:11.111112 +0200 CEST` for example for some reason I don't understand.

calmh:
Ok this is now quite simplified from before. I add the new `modified_ns` field to hold the nanoseconds part, create `ModTime()` method on `protocol.FileInfo` like exists on an `os.FileInfo` and then generally just juggle `time.Time`s everywhere instead of the old epoch seconds fields.

We gain being able to detect changes at whatever resolution the filesystem supports, and sync mtimes with up to microsecond precision (not nanosecond, as the standard library truncates in chtimes...).

lib/db/structs.proto
```diff
@@ -26,10 +26,11 @@ message FileInfoTruncated {
...
     int64                 size           = 3;
     uint32                permissions    = 4;
-    int64                 modified       = 5;
+    int64                 modified_s     = 5;
     bool                  deleted        = 6;
     bool                  invalid        = 7;
     bool                  no_permissions = 8;
     protocol.Vector       version        = 9 [(gogoproto.nullable) = false];
     int64                 sequence       = 10;
+    int32                 modified_ns    = 11;
```
AudriusButkevicius: Put this next to modified just with a higher id maybe

lib/protocol/bep.proto
```diff
@@ -97,12 +97,13 @@ message FileInfo {
...
     int64        size           = 3;
     uint32       permissions    = 4;
-    int64        modified       = 5;
+    int64        modified_s     = 5;
     bool         deleted        = 6;
     bool         invalid        = 7;
     bool         no_permissions = 8;
     Vector       version        = 9 [(gogoproto.nullable) = false];
     int64        sequence       = 10;
+    int32        modified_ns    = 11;
```
AudriusButkevicius: same comment as above

lib/scanner/walk.go
```diff
@@ -323,7 +323,8 @@ func (w *walker) walkRegular(relPath string, info os.FileInfo, fchan chan protoc
 		Version:       cf.Version.Update(w.ShortID),
 		Permissions:   curMode & uint32(maskModePerm),
 		NoPermissions: w.IgnorePerms,
-		Modified:      info.ModTime().Unix(),
+		ModifiedS:     info.ModTime().Unix(),
+		ModifiedNs:    int32(info.ModTime().UnixNano() % 1e9),
```
AudriusButkevicius: what is this magic constant

lib/db/structs.proto
calmh: Sure, will fix.

lib/scanner/walk.go
calmh: Stupidity. Will correct.

lib/scanner/walk.go
```diff
@@ -357,7 +358,8 @@ func (w *walker) walkDir(relPath string, info os.FileInfo, dchan chan protocol.F
 		Version:       cf.Version.Update(w.ShortID),
 		Permissions:   uint32(info.Mode() & maskModePerm),
 		NoPermissions: w.IgnorePerms,
-		Modified:      info.ModTime().Unix(),
+		ModifiedS:     info.ModTime().Unix(),
+		ModifiedNs:    int32(info.ModTime().UnixNano() % 1e9),
```
AudriusButkevicius: And what about this one?

AudriusButkevicius:
@st-review 

merge

st-review:
I'm sorry, @AudriusButkevicius. I'm afraid I don't know what you mean. I know how to `don't`, `lgtm`, `merge`, `prevent`, `squash`, `stop`.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as ea87bcefd6b83bcf343c785391bc709f867d4b3c. Thanks, @calmh!

PR #3437, opened 2016-07-25
lib/model: Stricter temporary file permissions

calmh:
### Purpose

We could have a file to sync with permissions `rw-------` but we'd create the temp file with `rw-rw-rw-` minus umask, usually `rw-r--r--`. This potentially exposes private data while the file is being synced.

Similarly, when ignorePerms was set and we were reusing a temp files we would set the permissions to `rw-r--r--` explicitly, potentially overriding a strict umask that would otherwise have had the file be `rw-------`.

### Testing

I ran some integration tests, but please stare intently at this code to make sure I haven't made a thinko.

### DIFF
--- a/lib/model/sharedpullerstate.go
+++ b/lib/model/sharedpullerstate.go
@@ -121,24 +121,40 @@ func (s *sharedPullerState) tempFile() (io.WriterAt, error) {
 		}
 	}
 
+	// The permissions to use for the temporary file should be those of the
+	// final file, except we need user read & write at minimum. The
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
This might cause issues on these retard NASes

calmh:
I'm not sure how, and also not sure if that matters. :)

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 6ed22d08851c62ae5c1b9ec98c81dada6686341d. Thanks, @calmh!

PR #3438, opened 2016-07-25
cmd/syncthing: Generate FolderCompletion events for folders shared with a connecting device (fixes #3436)

### LINKED ISSUE #3436: FolderCompletion is not sent when connecting with delta index
When nothing changed (not tested with changes, I assume that works) no FolderCompletion event is sent after connecting.

### Version Information

syncthing v0.14.0+3-g694da60 "Dysprosium Dragonfly" (go1.6.3 linux-amd64) unknown-user@lx-buildslave.syncthing.net 2016-07-23 18:32:15 UTC

calmh:
### Purpose

This used to happen by itself as the connecting device always sent an Index message and we triggered on that. Nowadays there's no guarantee for that, but we anyway need to send out one event to let listeners know the state of folders shared with the device.

### Testing

I ran it with `-verbose` and observed the events being generated when a device connected, a device that had no new index info to share.

### DIFF
--- a/cmd/syncthing/summaryservice.go
+++ b/cmd/syncthing/summaryservice.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/syncthing/syncthing/lib/events"
+	"github.com/syncthing/syncthing/lib/protocol"
 	"github.com/syncthing/syncthing/lib/sync"
 	"github.com/thejerf/suture"
 )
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 0f28626bb496dcf069db829a25f902f279d05d43. Thanks, @calmh!

PR #3443, opened 2016-07-25
lib/scanner: Recheck file size and modification time after hashing

calmh:
To catch the case where the file changed. Also make sure we never let a
size-vs-blocklist mismatch slip through.

https://forum.syncthing.net/t/since-0-14-lots-of-sync-conflict-files-appearing/7853/21

### DIFF
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -941,7 +941,7 @@ func (f *rwFolder) handleFile(file protocol.FileInfo, copyChan chan<- copyBlocks
 
 	// Check for an old temporary file which might have some blocks we could
 	// reuse.
-	tempBlocks, err := scanner.HashFile(tempName, protocol.BlockSize, 0, nil)
+	tempBlocks, err := scanner.HashFile(tempName, protocol.BlockSize, nil)
 	if err == nil {
 		// Check for any reusable blocks in the temp file
 		tempCopyBlocks, _ := scanner.BlockDiff(tempBlocks, file.Blocks)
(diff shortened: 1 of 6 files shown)


AudriusButkevicius:
feels a bit of hack, but I guess it's cheap

plouj:
Would this not keep hashing until the file stops changing faster than the
hashing can finish?

AudriusButkevicius:
It will just skip the file all together.

plouj:
Ah, ok. I guess outbox must be a list of files to hash.

FWIF I think the size != f.Size assertion should be done by unit tests for
HashFile.

AudriusButkevicius:
Outbox is what is already hashed.

plouj:
Of course when I say "outbox" I mean "inbox" :)

calmh:
I moved the check to where we open the file so we can check the fd, and then actually recalculate the size based on the block list after the hashing. Because we've stat:ed at some point long ago before putting the file in the inbox, and that might be out of date. The state just before hashing gets the latest data, and the state of the same fd afterwards ensures that the file didn't change while we were reading it.

I removed the sizehint to HashFile because it doesn't need it any more, and changed the size hint logic in the actual hasher to allow zero as a valid size hint, as that is a valid file size. Otherwise, a file that was zero bytes at the start but grew during hashing would cause an inconsistent result.

Phew... Lets see how this fares when hammered on.

AudriusButkevicius:
Good to merge, or are we waiting for more tests by someone?

calmh:
No, lets merge. I've ran the test script for a few hours and it looks good. I'll think on how we can add more regular tests for this. I'll release a 0.14.1 with this merged.

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 7aaa1dd8a3a2b8e6b9c69b5e147858cefbb4e708. Thanks, @calmh!

PR #3453, opened 2016-07-27
lib/model: Trigger a puller iteration on connection (fixes #3451)

### LINKED ISSUE #3451: Receiving device gets stuck in "out of sync" when sending device disconnects/reconnects
The files are available and could be pulled, but they are not:

screen shot 2016-07-27 at 21 55 34

The reason is that the puller loop runs when the device is disconnected, doesn't queue any files because they are all unavailable so it concludes there is nothing to do, and goes into holding pattern waiting for the index information to change (which would indicate new files available). But that never happens, because delta indexes.

calmh:
### Purpose

Fix it...

### Testing

The `TestRestartSenderDuringTransfer` integration test hangs without this fix, not with it.


calmh:

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 7c7e8648ffcd49222f5c3c358a820506b0c7a786. Thanks, @calmh!

PR #3454, opened 2016-07-27
lib/db, lib/model: Create temp sorting database in config dir (fixes #3449)

### LINKED ISSUE #3449: Index sorting should spill in home dir.
Why not?

If we are afarid we'll leave trash, we can clean up on start up.

calmh:
### Purpose

Create temp databases beside main database.

A design issue here is that the index sorter is created at a point quite far in the call stack from the closest place or even package that knows anything about disk paths. My solution takes a quite circuitous path via letting the main database know where it lives and have the model check that when it kicks of the index sending routines... I'm all open for suggestions on better ways to do this.

### Testing

`STTRACE=model`, the log "opened sorter" line shows the expected path.

### DIFF
--- a/lib/db/leveldb_dbinstance.go
+++ b/lib/db/leveldb_dbinstance.go
@@ -29,6 +29,7 @@ type deletionHandler func(t readWriteTransaction, folder, device, name []byte, d
 type Instance struct {
 	committed int64 // this must be the first attribute in the struct to ensure 64 bit alignment on 32 bit plaforms
 	*leveldb.DB
+	location  string
 	folderIdx *smallIndex
 	deviceIdx *smallIndex
 }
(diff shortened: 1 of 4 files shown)


plouj:
I might be confunsing golang with python, but Is it possible to store the db location in a package-global variable within the model package rather than needlessly passing it around the sendIndexes and sendIndexTo functions?

AudriusButkevicius:
Global state is evils

calmh:
That could be one option. Another would be to move the "locations" stuff that now lives in package `main` to package `config`, possibly.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 72026db5999f2276f4386c4972c3efa0ba93744e. Thanks, @calmh!

PR #3462, opened 2016-07-29
all: Rename LocalVersion to Sequence (fixes #3461)

### LINKED ISSUE #3461: Rename "local version" to "sequence number"
The name "local version" stems from when there was a single integer version per file that was cluster-global, and then this "local" version number beside it. Nowadays the versioning is much more advanced than just an integer, and the local version isn't really a "version" in any real sense of the word.

It is however a sequence number that is incremented on every change, and lets devices know where in the sequence to resume index sending and so on. So renaming it to that would make more sense to me.

```diff
--- a/lib/protocol/bep.proto
+++ b/lib/protocol/bep.proto
@@ -68,7 +68,7 @@ message Device {
     repeated string addresses    = 3;
     Compression     compression  = 4;
     string          cert_name    = 5;
-    int64           max_local_version = 6;
+    int64           max_sequence = 6;
     bool            introducer   = 7;
     uint64          index_id     = 8 [(gogoproto.customname) = "IndexID", (gogoproto.customtype) = "IndexID", (gogoproto.nullable) = false];
 }
@@ -102,7 +102,7 @@ message FileInfo {
     bool         invalid        = 7;
     bool         no_permissions = 8;
     Vector       version        = 9 [(gogoproto.nullable) = false];
-    int64        local_version  = 10;
+    int64        sequence       = 10;
 
     repeated BlockInfo Blocks = 16 [(gogoproto.nullable) = false];
 }
```

calmh:
### Purpose

Clarity.

Mostly machine replaced with some touchups.

### Testing

It still builds, I think.

### Documentation

Some things to follow:

- [ ] Specs
- [ ] Manpages etc

This is a change in the REST API format for `/rest/db/get` and `/rest/db/need` but I doubt anyone cares. @canton7 @Zillode @Nutomic ? For the folder summary event someone might conceivably care so I've retained the old field `version` in addition to the new `sequence`.


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 297240facfaccdf575a9edb1895c0660628743af. Thanks, @calmh!

PR #3463, opened 2016-07-30
minor accessibility fixes #3297

### LINKED ISSUE #3297: Add minor accessibility fixes to the web GUI
I'm adding some minor accessibility fixes (mainly cosmetic) to your web gui (I'm a blind computer user and active member of the [NVDA](http://www.nvaccess.org) project) and use syncthing for maintaining many files on a subdomain of http://derekriemer.com . First off, great job on your User interface. It's really easy to understand and structured well, and because of syncthing, I have a much easier way of adding files to my website than having to jump through my many layers of authentication to connect via sftp. The minor fixes I can include are:

- There should be a heading level one somewhere on the page to comply with HTML best practices.
- Either an alt tag should be added to each of the syncthing logos that are created on line 35 of index.htm, or the span should be marked with aria-hidden="true" to convey to assistive technology to ignore these two logos.
- the language menu incorrectly has aria-expanded="true" rather than aria-expanded="false" when it starts up. Therefore, Syncthing reports "English expanded" rather than "english collapsed" as it should when the drop down is not shown.
- The actions menu does not have aria-expanded="false" on it, which causes it to report as a link, and not a link that can be expanded. This doesn't cause the link you click to activate to have any accessibility issues (The angular dropdown I think is fixing this after an expand/collapse cycle as well). Once the user activates the link, the dropdown is accessible, this is simply a cosmetic fix to alert users of Assistive Technologies that a dropdown menu exists.
- Optionally, we could add roles of menu and menuitem to each of the dropdowns code (on the ul and li tags) to turn the dropdowns into menus that appear when they are activated. If we do this, we should hide the a tags inside the list items from assistive technology.
- role="main" should be added around the area that contains the folders, and devices, to convey to assistive technology that this is the main contents.
- The folders are buttons (as of the last release), but do not have a collapsed state. I could add this by adding aria-expanded="true". I've tried this with the source, and it works great.
- We should add aria-expanded="true" to the button that displays the device name under the heading level 3. This way, assistive technology users know you can collapse this.
- Add collapsed states to each device present in the HTML code under remote devices.
- Some of the tabs on the advanced settings page need cosmetic fixes to the collapsed and expanded states. For example, currently they by default all say expanded, even if they should be collapsed. Note that after you expand and then collapse them, they take on the correct state.
- There are some modal dialogs like the simlinks are disabled that appear and don't ever get focus to set to them. Currently, a user must find them manually.

Thanks.

derekriemer:
### Purpose

This fixes a few accessibility issues, mainly the state of certain controls.

### Testing

Testing in mozilla firefox, Google Chrome, and Internet explorer with the NVDA screen reader have been done. I'm not sure if this project has integration tests.

### Screenshots

This is a GUI change, but the GUI should not visually have changed.


derekriemer:
Hi, Do you guys prefer rebasing or merging onto master?
I can merge master into this to fix the issues, but would a rebase be prefered?

calmh:
Merging master into the branch as necessary is preferred. It'll get squashed at merge time anyhow.

calmh:
@st-jenkins add to whitelist

gui/default/index.html
```diff
@@ -32,11 +32,11 @@
 
   <nav class="navbar navbar-top navbar-default" role="navigation">
     <div class="container">
-      <span class="navbar-brand">
+      <span class="navbar-brand" aria-hidden="true">
         <img class="logo hidden-xs" src="assets/img/logo-horizontal.svg" height="32" width="117"/>
         <img class="logo hidden visible-xs" src="assets/img/favicon-default.png" height="32"/>
       </span>
-      <p class="navbar-text hidden-xs" ng-class="{'hidden-sm':upgradeInfo && upgradeInfo.newer}">{{thisDeviceName()}}</p>
+      <h1 class="navbar-text hidden-xs" ng-class="{'hidden-sm':upgradeInfo && upgradeInfo.newer}">{{thisDeviceName()}}</h1>
```
calmh: This change doesn't seem good though, it increases the size of the device name in the header drastically. Perhaps there is an aria tag that could be used instead to indicate it's headerness, or the styles need to be tweaked to not do this for h1-in-navbar.

screen shot 2016-07-30 at 10 08 02

AudriusButkevicius:
Also, the html gains double spaces in some places, which I am not fully happy with.

gui/default/index.html
derekriemer: It's not a huge deal if that isn't a heading, so I'll remove it. I meant to remove it, and forgot before pushing.

gui/default/index.html
```diff
@@ -514,9 +514,9 @@ <h4 class="panel-title">
 
         <!-- Remote devices -->
         <h3 translate>Remote Devices</h3>
-        <div class="panel-group" id="devices">
+        <div class="panel-group" id="devices"></div>
```
derekriemer: I will remove this mistaken div closing tag I introduced.

gui/default/index.html
```diff
@@ -437,10 +437,10 @@ <h4 class="panel-title">
 
       <!-- This device -->
 
-      <div class="col-md-6">
+      <div class="col-md-6" aria-label="{{'Devices' | translate}}" role="region">
         <h3 translate>This Device</h3>
-        <div class="panel panel-default" ng-repeat="deviceCfg in [thisDevice()]">
-          <button class="btn panel-heading" data-toggle="collapse" data-target="#device-this">
+        <div class="panel panel-default"  ng-repeat="deviceCfg in [thisDevice()]">
```
AudriusButkevicius: Double space here

gui/default/syncthing/settings/advancedSettingsModalView.html
```diff
@@ -27,9 +27,9 @@ <h4 class="panel-title" translate>GUI</h4>
       </div>
 
       <div class="panel panel-default">
-        <div class="panel-heading" role="tab" id="optionsHeading" data-toggle="collapse" data-parent="#advancedAccordion" href="#optionsConfig" aria-expanded="true" aria-controls="optionsConfig" style="cursor: pointer">
-          <h4 class="panel-title" translate>Options</h4>
-        </div>
+        <div class="panel-heading" role="tab" id="optionsHeading" data-toggle="collapse" data-parent="#advancedAccordion" href="#optionsConfig" aria-expanded="false" aria-controls="optionsConfig" style="cursor: pointer;">
+            <h4  class="panel-title" tabindex="0" translate>Options</h4>
+        </div>
```
AudriusButkevicius: Trailing spaces, double spaces, and double indentation

AudriusButkevicius:
I'm fine to merge once last comments are addressed.

AudriusButkevicius:
@st-review merge

gui: Improve accessibility (fixes #3297)

skip-check: authors

st-review:
:ok_hand: Merged as a8cd9d0154d39e589ca876c662a50c4ef36560ca. Thanks, @derekriemer!

Issue #3458, opened 2016-07-28
Odd behaviour when syncing a folder ending with a space on Windows


canton7:
Interesting issues with syncthing folders which containing a trailing space between Linux and Windows.

Setup is one Linux machine running Syncthing, and one Windows machine, both running 0.14. Assume a rescan after each command.

#### Linux
Create a folder called "test " (with the trailing space), with a file (called "file") inside

#### Windows
```
[UAL5T] 20:27:59 WARNING: File name "test " (folder "Docs") contains invalid characters; marked as invalid.
[UAL5T] 20:27:59 INFO: Resurrecting directory \\?\C:\<Path>\Docs\test 
```
... but the folder is created with a trailing space anyway.

Rename the folder to "test" (with no trailing space) - I had to use bash to do this, but there are probably neater ways.

#### Linux

Folders "test" and "test " both exist now.

Remove "test "

#### Windows

Syncthing tries to delete "file", but that fails with the error `GetFileAttributesEx \\?\C:\<Path>\Docs\test : The system cannot fine the file specified.`

There is a way to get Syncthing to delete the contents of that folder, but I haven't managed to reproduce it. It probably has something to do with two Windows computers and a Linux computer, all being on at different points.

I've also had an issue where Syncthing kept resurrecting the folder with a space in on the Windows side, but again, not (yet) reproducible.

Thanks!

AudriusButkevicius:
There was a patch marking files with trailing spaces as invalid. Looking at the log it seems to be the case, not sure where the resurrecting thing comes from.

AudriusButkevicius:
Ah ok, I see how this happens.

calmh:
Right. If the folder is invalid, so should the files inside be...

Issue #3457, opened 2016-07-28
Delta indexes doesn't play well with changing ignores


calmh:
We process incoming index updates according to the ignore patterns, to filter out unwanted entries. When ignore patterns changed to allow files that were previously filtered out things would be OK on the next index exchange / reconnect. Currently, there is no new index exchange so the new ignore patterns don't get matched against the remote index data.

We need to either save all the data and process it differently, or as a workaround drop the delta index data when ignore patterns change.

plouj:
Is it potentially a lot of data to keep that may never be used? Do people
change ignores very often?

canton7:
Would it make sense to always accept (and never filter) incoming index entries, but to make ones we don't care about as "ignored" in the database (and not try and pull them)?

I have a suspicion that would also help solve the "my folder's out of sync because I have ignore patterns" issue...

calmh:
Yes, I think something like that.

AudriusButkevicius:
We could also scan (as in walk, and mark as invalid with no blocks) files that are ignored locally be of any benefit? And potentially advertising them?

canton7:
Does it make subsequent scans cheaper, as you're not having to match ignored files against the ignore rules each time?

scienmind:
Is there any way to work around it manually for now? 
(i.e. force an index exchange for a certain repo)

I had to exclude few subfolders from syncing, now all related repos are stuck on "Out of sync" due to excluded folders. And it seems that this issue is the cause.
(the exclusion done via `#include...`+`.stglobalignore`)

AudriusButkevicius:
Remove the folder ans readd it, which will force a rescan.

Issue #3012, opened 2016-04-26
Files with invalid file names should fail louder


BethAr:
*Edit by calmh because tl;dr:* Files with invalid characters in the file name are dropped from incoming updates. With a warning, but that warning goes away and then the device looks happy in the GUI. Instead, we should accept the invalid files into the index, display a non-100% sync status, and warn repeatedly (every pull iteration, unless we do something else?).

----

Hello,

I have a Linux (Mint 17.3) machine synching a backup folder to a Windows 7 computer. I thought the backup was running correctly for a couple of months, but when I had to format I found out many files never arrived at the Windows machine.

Obviously the problem was Windows stupid filepath limitation, but I was a bit scared that Syncthing didn't warn me in any way that some files were never synched. The fact that it told me all that time that the folder was _up to date_ is very unnerving.

I would like to request some sort of warning when this happens. The folder on the Linux machine was set as a master folder, and no file was ever added to the Windows machine's folder manually, so it should tell me that there were files missing on the other end, not that the folder was up to date.

I can't specify the version because I went through several Syncthing updates in that time.

Thank you.

calmh:
That sounds very odd indeed. It will fail to sync filenames that can't be represented on windows, but it does so quite loudly and and gets visibly stuck in "out of sync". Not showing an error and looking "up to date" is surprising - can you reproduce this?

AudriusButkevicius:
So I think we should close this, atleast until steps of reproduction have been provided, because this is based on some old memories of some old version, which might not even be relevant anymore.

calmh:
I don't think we've *ever* behaved as the OP describes, or I hope so, so I would like at least some comments on this before closing. Perhaps it's reproducible.

BethAr:
Ok,

The last version that was installed on the machine was v0.12.21, and I noticed the problem just last week, so it isn't old. I don't know which version I started with, but I know the dates, so maybe you can tell.

Here is my setup in more detail. It began like this:

**Source machine**
HP Pavillion g4 notebook running Windows 7 Professional x64.
Installed Syncthing on 2016-01-20 (I know because I mailed me the ID)
3-4gb backup folder set as Master: C:\Beta\Backup
320gb Samsung HDD, single NTFS partition.
Running Avast antivirus.

**Target machine**
Dell desktop running WIndows 7 Professional x64.
Installed Syncthing on 2016-01-20 (same day)
Folder synched to: F:\Apoio\Beta\Backup
[...]

PR #3464, opened 2016-07-30
Correct handling of ignores with delta indexes

calmh:
This replaces the filtering done in index updates with filtering when traversing the need lists.

It leaves the `invalid` bit completely unused. A future PR can use it to indicate files that are known but ignored and let other devices update their in sync status based on it.


lib/model/model.go
```diff
@@ -2235,27 +2253,6 @@ func mapDeviceCfgs(devices []config.DeviceConfiguration) map[protocol.DeviceID]s
...
-			fs[i] = fs[len(fs)-1]
-			fs = fs[:len(fs)-1]
-		} else {
-			i++
-		}
-	}
-	return fs
-}
-
 func symlinkInvalid(folder string, fi db.FileIntf) bool {
```
AudriusButkevicius: You could still use a simillar function to perform filtering in both places. Also symlinks invalid in this case might not be needed anymore? (Unless we want the warning to stay).

AudriusButkevicius:
This does however change how the completion looks like. So we'll look up to date locally and out of sync remotely. Can we add a new explicit ignored bit, send that across and update the bits as we scan/update ignores if they change?

Then we can have two completion numbers

calmh:
I think this would be a way to return the `Invalid` bit to it's original meaning - "do not attempt to download this file from me". For ignored files we would process the file as usual, but instead of doing anything with it (like pulling or deleting) we simply set the invalid bit and do the database update. For other devices it looks like we processed it but are simply unwilling to provide the file to anyone else. And we never set the invalid bit under other circumstances, so it could be renamed "ignored" instead...

lib/model/rwfolder.go
```diff
@@ -424,20 +424,11 @@ func (f *rwFolder) pullerIteration(ignores *ignore.Matcher) int {
...
-		}
-		if !symlinks.Supported && file.IsSymlink() {
-			return true
-		}
-		if ignores.Match(file.Name).IsIgnored() {
+		if shouldIgnore(intf, ignores, f.ignoreDelete) {
 			return true
 		}
 
-		l.Debugln(f, "handling", file.Name)
```
AudriusButkevicius: I find thiz message very useful when debugging.

AudriusButkevicius:
We do set invalid for files we can't handle. Does that case work fine or should we make a simillar change?

calmh:
I was thinking of removing that filtering as well and doing it in the same place as here. Mostly to clear up the handling of the invalid bit.

AudriusButkevicius:
Makes sense.

lib/model/model.go
calmh: (done)

calmh:
I think this is ready for review now. It's not super easy to test this stuff automatically unfortunately. I've done manual testing to make sure ignores behave as I expect (which is essentially like they did pre-deltas), but not yet verified the invalid name handling on Windows.

Give me until tomorrow or so to do that as well.

lib/model/model.go
```diff
@@ -1278,11 +1285,6 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
 			maxSequence = f.Sequence
 		}
 
-		if ignores.Match(f.Name).IsIgnored() || symlinkInvalid(folder, f) {
-			l.Debugln("not sending update for ignored/unsupported symlink", f)
-			return true
```
calmh: I'm not super sure why we had this filtering here, but I'm fairly confident we shouldn't do it. If it does any difference at all it would screw up delta indexes as things with low sequence numbers would be mysteriously missing and never get sent.

However I don't see what we would ever have ignored files in our "have" list (unless they were not ignored previously, and then they should been sent then so we are fine) nor invalid symlinks (they are invalid so we can't have them locally).

calmh:
Appears to work for the Windows stuff as well; the lines are logged once at pull time.

screen shot 2016-08-03 at 10 10 08

Is the INFO level too subtle here? I made it info as it could potentially be a lot of lines (like every file in an invalid folder) but it might slip by undetected.

The Windows device will look "syncing" from the other side as it doesn't process these files, obviously.

AudriusButkevicius:
We should spit a warning atleast one, of a generic nature.

AudriusButkevicius:
Or we could fail these at pull time, and they would just appear in failed items.

calmh:
That would probably be better from a user point of view. The downside is that it would leave the puller continuously retrying them (causing a database scan every minute if nothing else).

calmh:
OK, here's a second try. This moves the invalid checks to the puller, resulting in the errors showing up as normal sync errors in the GUI:

screen shot 2016-08-05 at 08 38 51

screen shot 2016-08-05 at 08 38 57

As a bonus this now also fixes #3458

calmh:
Rebased to update commit subject

calmh:
Also fixes #3012 so the subject should be

`cmd/syncthing, lib/...: Correctly handle ignores & invalid file names (fixes #3012, fixes #3457, fixes #3458)`

when merged

AudriusButkevicius:
@st-review merge

cmd/syncthing, lib/...: Correctly handle ignores & invalid file names (fixes #3012, fixes #3457, fixes #3458)

st-review:
:ok_hand: Merged as 1eb6db6ca86f5c885777bf1809d790c5ede39794. Thanks, @calmh!

PR #3467, opened 2016-08-02
cmd/syncthing, lib/config: Enable HTTP CPU/heap profile collection for users

calmh:
### Purpose

This adds a config to enable debug functions on the API server, which is by default disabled. When enabled, the /rest/debug things become available and become available without requiring a CSRF token (although authentication is required if configured).

We also add a new endpoint /rest/debug/cpuprof?duration=15s (with the duration being configurable, defaulting to 30s). This runs a CPU profile for the duration and returns it as a file. It sets headers so that a browser will save the file with an informative name.

The same is done for heap profiles, /rest/debug/heapprof, which does not take any parameters. The heap profile won't contain historical allocations, but a snapshot of live data at the time it's taken.

The purpose of this is that any user can enable debugging under advanced, then point their browser to the endpoint above and get a file that contains a CPU/heap profile we can use, with the filename telling us what version and architecture the profile is from.

On the command line, this becomes

    curl -O -J http://localhost:8082/rest/debug/cpuprof?duration=5s
    curl: Saved to filename
    'syncthing-cpu-darwin-amd64-v0.14.3+4-g935bcc0-110307.pprof'

### Testing

Works for me

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -17,6 +17,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"runtime"
+	"runtime/pprof"
 	"sort"
 	"strconv"
 	"strings"
@@ -267,9 +268,15 @@ func (s *apiService) Serve() {
 	postRestMux.HandleFunc("/rest/system/resume", s.postSystemResume)          // device
 	postRestMux.HandleFunc("/rest/system/debug", s.postSystemDebug)            // [enable] [disable]
 
-	// Debug endpoints, not for general use
-	getRestMux.HandleFunc("/rest/debug/peerCompletion", s.getPeerCompletion)
-	getRestMux.HandleFunc("/rest/debug/httpmetrics", s.getSystemHTTPMetrics)
+	guiCfg := s.cfg.GUI()
+
+	if guiCfg.EnableDebugging {
+		// Debug endpoints, not for general use
+		getRestMux.HandleFunc("/rest/debug/peerCompletion", s.getPeerCompletion)
+		getRestMux.HandleFunc("/rest/debug/httpmetrics", s.getSystemHTTPMetrics)
+		getRestMux.HandleFunc("/rest/debug/cpuprof", s.getCPUProf) // duration
+		getRestMux.HandleFunc("/rest/debug/heapprof", s.getHeapProf)
+	}
 
 	// A handler that splits requests between the two above and disables
 	// caching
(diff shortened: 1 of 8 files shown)


calmh:
@st-jenkins retest this now

calmh:

lib/config/guiconfiguration.go
```diff
@@ -21,6 +21,7 @@ type GUIConfiguration struct {
 	APIKey              string `xml:"apikey,omitempty" json:"apiKey"`
 	InsecureAdminAccess bool   `xml:"insecureAdminAccess,omitempty" json:"insecureAdminAccess"`
 	Theme               string `xml:"theme" json:"theme" default:"default"`
+	EnableDebugging     bool   `xml:"debugging,attr" json:"debugging"`
```
AudriusButkevicius: Inconsistent naming. Debugging vs enabledebugging

cmd/syncthing/gui.go
```diff
@@ -267,9 +268,15 @@ func (s *apiService) Serve() {
...
-	getRestMux.HandleFunc("/rest/debug/httpmetrics", s.getSystemHTTPMetrics)
+	guiCfg := s.cfg.GUI()
+
+	if guiCfg.EnableDebugging {
+		// Debug endpoints, not for general use
+		getRestMux.HandleFunc("/rest/debug/peerCompletion", s.getPeerCompletion)
+		getRestMux.HandleFunc("/rest/debug/httpmetrics", s.getSystemHTTPMetrics)
+		getRestMux.HandleFunc("/rest/debug/cpuprof", s.getCPUProf) // duration
+		getRestMux.HandleFunc("/rest/debug/heapprof", s.getHeapProf)
+	}
```
AudriusButkevicius: Should we always add these, and then just return 500s if the var is not true? Prevents need to restart.

calmh: Maybe. Let me see how that would look.

calmh:
Yeah that looks OK to me as well.

AudriusButkevicius:
Do we need to copy the value of the bool somewhere in config saving to prevent restart messages?

calmh:
No, the restart triggering thing in `model` doesn't look at the GUI config at all.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as ffe7a2fcd70cb731712a662daf2b11d3e3aae30a. Thanks, @calmh!

PR #3471, opened 2016-08-03
lib/config, lib/connections: Refactor handling of ignored devices (fixes #3470)

### LINKED ISSUE #3470: Incorrectly logs "connection from ignored device" for non-ignored devices
```
[I6KAH] 10:06:00 INFO: Connection from [fe80::5d2c:53ce:2c:3e97%en0]:50891 (TCP (Server)) with ignored device ID JKYS3V4-U3F2KJ7-6C76GJQ-LMKI7SL-H75YSLB-X4JVUCG-CPIBMFU-VUMJJA2
```

That's just a connection from a regular, new, device. The popup in the UI works fine. But the log line is confusing.

0.14.3

calmh:
### Purpose

Fix the logging around ignored devices. But I also noticed that I think we do too much processing for ignored devices and redid that to drop them earlier. That removed the only thing that `model.OnHello` did, so I removed that method. I'm not at all opposed to having it, but since it doesn't do anything after this change I don't think it makes sense to keep it around - we can reinstate it when necessary?

### Testing

Added a test to make sure we handle the case of manually adding a device to override an older ignore. Manually verified to have a device connect, ignore it, then add it manually.

### DIFF
--- a/lib/config/config.go
+++ b/lib/config/config.go
@@ -273,6 +273,16 @@ func (cfg *Configuration) prepare(myID protocol.DeviceID) error {
 		cfg.GUI.APIKey = rand.String(32)
 	}
 
+	// The list of ignored devices should not contain any devices that have
+	// been manually added to the config.
+	newIgnoredDevices := []protocol.DeviceID{}
+	for _, dev := range cfg.IgnoredDevices {
+		if !existingDevices[dev] {
+			newIgnoredDevices = append(newIgnoredDevices, dev)
+		}
+	}
+	cfg.IgnoredDevices = newIgnoredDevices
+
 	return nil
 }
 
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -697,3 +697,20 @@ func TestV14ListenAddressesMigration(t *testing.T) {
 		}
 	}
 }
+
+func TestIgnoredDevices(t *testing.T) {
+	// Verify that ignored devices that are also present in the
+	// configuration are not in fact ignored.
+
+	wrapper, err := Load("testdata/ignoreddevices.xml", device1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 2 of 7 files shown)


lib/config/config.go
```diff
@@ -270,6 +270,16 @@ func (cfg *Configuration) prepare(myID protocol.DeviceID) error {
 		cfg.GUI.APIKey = rand.String(32)
 	}
 
+	// The list of ignored devices should not contain any devices that have
+	// been manually added to the config.
+	newIgnoredDevices := []protocol.DeviceID{}
```
calmh: We want the items here to not be nil for nicer JSON serialization, which is why this is not just a `var newIgnoredDevices []protocol.DeviceID`.

AudriusButkevicius:
Prepare() gets called not that often, so I suspect this won't work on a 'live' system, just because the ignoredness is checked first. I'd rather you just made OnHello return an error, and remove the ! from the if condition.

calmh:
Prepare should be called on any config load or save though, so that should be fine. You mean returning a specific "this devices is ignored" type error from OnHello, or moving more of the connection service handling in to the OnHello method?

AudriusButkevicius:
Well either a specific error or generic error. Doesn't matter, an error is an error, so drop the connection.

calmh:
OK, a second take on this. I moved more checks into OnHello instead, and I think it became a bit cleaner. I also made a convenience method for finding the device config as we had a couple of loops for that - that means the connection service stuff got outdented so use [?w=1](https://github.com/syncthing/syncthing/pull/3471/files?w=1) on this one...

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as f368d2278f843b6f8aae072fc489adab37590630. Thanks, @calmh!

PR #3479, opened 2016-08-05
lib/db, lib/fs, lib/model: Introduce fs.MtimeFS, remove VirtualMtimeRepo

calmh:
### Purpose

This changes the mtime handling to be a layer that intercepts Lstat and Chtimes to always present a consistent view to the application. It's a bit more ambitious in getting it right than the previous implementation as it will always check what the resulting mtime was after Chtimes, regardless of if an error was returned or not, and record any difference.

This correctly handles filesystems that truncate times to 2s and stuff like that as well.

It's a small step towards a full `fs` implementation in the future.

It's also a building block for the high precision timestamps as the new mtimefs always presents (virtual) nanosecond precision, with best effort precision on the actual filesystem.

The database format is unchanged from before so the old values are still good.

### Testing

There are unit tests on the MtimeFS. I ran some integration tests to make sure things weren't unexpectedly broken.

I'm running this plus the high precision timestamps stuff (not yet pushed) on my main sync boxes and it seems to work as expected.


lib/fs/mtimefs.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (t *dbMtime) Unmarshal(bs []byte) error {
+	if err := t.real.UnmarshalBinary(bs[:len(bs)/2]); err != nil {
+		return err
+	}
+	if err := t.virtual.UnmarshalBinary(bs[len(bs)/2:]); err != nil {
+		return err
+	}
+	return nil
+}
```
AudriusButkevicius: I know this is not your code, but this relies on the binary representation to return a fixed number of bytes, which is suboptimal.

calmh: Yeah I know, but looking at the actual marshalbinary method it is sane and stable and doesn't look likely to change anytime soon. If you look at the first commit I actually changed to use a protobuf representation of two nanosecond fields, but in the end I thought it made more sense to keep the existing format and avoid issues on upgrade on Android etc

AudriusButkevicius:
So actually, this whole commit does literally nothing, functionality wise, which makes me question if we need it at all?

This fs concept is half backed as it's just a partial abstraction

calmh:
No, it's a necessary step for the high precision time stamps, it solves the issue with 2s granularity on FAT, and it cleans up the virtual mtime code and starts down the path of a file system abstraction.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 0655991a19e23dd585ee0c697a48931362cdae21. Thanks, @calmh!

PR #3480, opened 2016-08-05
lib/connections: Fix connection switching

AudriusButkevicius:
### Purpose

It seems that it would be impossible to drop down to relay after establishing a direct connection
Also, we should not drop the existing connection until after we've passed the validation steps,
and it seems it's being dropped in two places unnecesserily at the moment.

### Testing

None what so ever.

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -193,15 +193,17 @@ next:
 
 		// If we have a relay connection, and the new incoming connection is
 		// not a relay connection, we should drop that, and prefer the this one.
+		connected := s.model.ConnectedTo(remoteID)
 		s.curConMut.Lock()
 		ct, ok := s.currentConnection[remoteID]
 		s.curConMut.Unlock()
+		priorityKnown := ok && connected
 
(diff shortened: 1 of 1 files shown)


calmh:
Lgtm but let me do some actual testing since I anyway intended to do that

calmh:
I have a vague suspicion that something like the following can happen, but it's not confirmed at all:

- relay connection exists
- direct connection comes in
- we drop relay connection
- we add direct connection
- other side drops relay connection, we get some error, call close
- close is based on device id, we remove the direct connection for that device is that we added milliseconds ago
- index sending starts for the direct connection that we just added and closed

But needs investigation...

AudriusButkevicius:
Totally possible

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as a4f052ad31fa160ad376598a920a5938699f58b9. Thanks, @AudriusButkevicius!

calmh:
No testing yet but it looks safe. We should look again into the possible race condition described above or something similar.

PR #3483, opened 2016-08-08
build, lib: Correct total test coverage calculation

calmh:
### Purpose

High numbers are nice, but truthful ones are more useful.

### Testing

Lets see if this builds.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -104,13 +104,18 @@ case "${1:-default}" in
 		# For every package in the repo
 		for dir in $(go list ./lib/... ./cmd/...) ; do
 			# run the tests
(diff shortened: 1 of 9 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as a4ed50ca85e670a5f0ff064e220915f310ddb553. Thanks, @calmh!

calmh:
This will, if I can hack it, result in st-review saying things like

---

This PR increases test coverage by 0.001 percentage points! Good work @AudriusButkevicius!

---

when it does change. With some suitable polite wording when the coverage goes down...

calmh:
Hah. Made complicated by the fact that the test coverage varies randomly from run to run...

PR #3489, opened 2016-08-10
lib/connections: Add KCP support (fixes #804)

### LINKED ISSUE #804: Consider using reliable UDP for block transfers
This is a long-shot, over-engineering and is not required, but a interesting topic nonetheless which could be a unique selling point to Syncthing.

Some interesting ideas can be found [here](http://stackoverflow.com/questions/107668/what-do-you-use-when-you-need-reliable-udp)

I seems that the easiest one to adopt would be [UDT](http://udt.sourceforge.net/) with bindings already available for most languages.

AudriusButkevicius:
This is mostly for benchmarks and testing.

Seems to connect, and not crash immediately, which is good news.

Known issues:
1. No timeouts as part of the protocol, so we usually end up with a dead connection for 5-7.5 minutes until it times out. This could be addressed by always using the new connection that we get, as long as the priority is right
2. Has some weirdness around closing connections. See https://github.com/xtaci/kcp-go/issues/18

The diff is massive due to deps, so I suggest you pull the PR locally and diff inside the lib and cmd/syncthing dirs.


calmh:

I diddled the API keys which some random post on the internet said had helped once to get commit status reporting working again...

calmh:
> No timeouts as part of the protocol, so we usually end up with a dead connection for 5-7.5 minutes until it times out.

This is fine by me as this is how we treat TCP as well.

AudriusButkevicius:
Well there is no FIN or RST so Close pretty much does nothing, which means we have to wait for 7 minutes for the connection to timeout at the other end and stop getting Already connected to there. This will be very painful with adding folders etc etc. Apparently I was suggested to use yamux on top of kcp to bring all of these control messages back in. Will try that.

AudriusButkevicius:
Also, luckily my new router uses restricted nat rather than full cone nat, so I jist realized that I the way I implemented it, it won't punch through anything else than full cone.

calmh:
I'm not 100% up to speed on the various cones, but I suspect punching anything other than full cone is quite difficult.

I read but didn't fully understand hos comments about lack of tcp-like handshaking, as that doesn't make sense to me in the context of having the session manager with Listen, Accept, Close and so on on top of the kcp layer, which he does. You'll probably make more sense of that than me :)

AudriusButkevicius:
Kcp is mostly about framing and re-delivery. It's nothing about being actually connected (checking that we are connected) or actually disconnecting as the protocol only defines a way to move data, but no way to indicate that we want to disconnect for example.

With the new code that replaces connections it might be fine though.

AudriusButkevicius:
And with the change I am planning, the only unpunchable nat should be symetric nat

calmh:
Cool

AudriusButkevicius:
@st-jenkins stfu

calmh:
Hah. I can see that getting annoying, yes.

calmh:
Disabled that for now, will rethink it.

calmh:
Something odd with the listener?

```
jb@syno:~/s/g/s/s/test $ grep listen h1/config.xml
        <listenAddress>dynamic+https://relays.syncthing.net/endpoint</listenAddress>
        <listenAddress>kcp://127.0.0.1:22001</listenAddress>
jb@syno:~/s/g/s/s/test $ syncthing -home h1
[monitor] 08:24:25 INFO: Starting syncthing
[I6KAH] 08:24:25 INFO: syncthing v0.14.4+4-g6850dc5-pr-3489 "Dysprosium Dragonfly" (go1.6.3 darwin-amd64) jb@syno.nym.se 2016-08-10 19:55:19 UTC
[I6KAH] 08:24:25 INFO: My ID: I6KAH76-66SLLLB-5PFXSOA-UFJCDZC-YAOMLEK-CP2GB32-BV5RQST-3PSROAU
[I6KAH] 08:24:25 INFO: Single thread hash performance is ~204 MB/s
[I6KAH] 08:24:25 INFO: Starting deadlock detector with 20m0s timeout
[I6KAH] 08:24:25 INFO: Ready to synchronize default (readwrite)
[I6KAH] 08:24:25 INFO: Ready to synchronize ¯\_(ツ)_/¯ Räksmörgås 动作 Адрес (readwrite)
[I6KAH] 08:24:25 INFO: listen (BEP/kcp): listen udp 127.0.0.1:22001: bind: address already in use
2016/08/12 08:24:25 Supervisor connections.Service: connections.Service: Failed service 'kcp://127.0.0.1:22001' (1.000000 failures of 5.000000), restarting: true, error: "kcp://127.0.0.1:22001 returned unexpectedly", stacktrace: [unknown stack trace]
[I6KAH] 08:24:25 INFO: listen (BEP/kcp): listen udp 127.0.0.1:22001: bind: address already in use
2016/08/12 08:24:25 Supervisor connections.Service: connections.Service: Failed service 'kcp://127.0.0.1:22001' (1.999984 failures of 5.000000), restarting: true, error: "kcp://127.0.0.1:22001 returned unexpectedly", stacktrace: [unknown stack trace]
2016/08/12 08:24:25 Supervisor connections.Service: connections.Service: Failed service 'kcp://127.0.0.1:22001' (2.999980 failures of 5.000000), restarting: true, error: "kcp://127.0.0.1:22001 returned unexpectedly", stacktrace: [unknown stack trace]
[I6KAH] 08:24:25 INFO: listen (BEP/kcp): listen udp 127.0.0.1:22001: bind: address already in use
[I6KAH] 08:24:25 INFO: listen (BEP/kcp): listen udp 127.0.0.1:22001: bind: address already in use
2016/08/12 08:24:25 Supervisor connections.Service: connections.Service: Failed service 'kcp://127.0.0.1:22001' (3.999976 failures of 5.000000), restarting: true, error: "kcp://127.0.0.1:22001 returned unexpectedly", stacktrace: [unknown stack trace]
[I6KAH] 08:24:25 INFO: listen (BEP/kcp): listen udp 127.0.0.1:22001: bind: address already in use
2016/08/12 08:24:25 Supervisor connections.Service: connections.Service: Failed service 'kcp://127.0.0.1:22001' (4.999971 failures of 5.000000), restarting: true, error: "kcp://127.0.0.1:22001 returned unexpectedly", stacktrace: [unknown stack trace]
[I6KAH] 08:24:25 INFO: listen (BEP/kcp): listen udp 127.0.0.1:22001: bind: address already in use
2016/08/12 08:24:25 Supervisor connections.Service: Entering the backoff state.
2016/08/12 08:24:25 Supervisor connections.Service: connections.Service: Failed service 'kcp://127.0.0.1:22001' (5.999964 failures of 5.000000), restarting: false, error: "kcp://127.0.0.1:22001 returned unexpectedly", stacktrace: [unknown stack trace]
[I6KAH] 08:24:25 INFO: Completed initial scan (rw) of folder ¯\_(ツ)_/¯ Räksmörgås 动作 Адрес
[I6KAH] 08:24:25 INFO: Device EJHMPAQ-OGCVORE-ISB4IS3-SYYVJXF-TKJGLTU-66DIQPF-GJ5D2GX-GQ3OWQK is "s4" at [kcp://127.0.0.1:22004]
[I6KAH] 08:24:25 INFO: Device I6KAH76-66SLLLB-5PFXSOA-UFJCDZC-YAOMLEK-CP2GB32-BV5RQST-3PSROAU is "s1" at [kcp://127.0.0.1:22001]
[I6KAH] 08:24:25 INFO: Device MRIW7OK-NETT3M4-N6SBWME-N25O76W-YJKVXPH-FUMQJ3S-P57B74J-GBITBAC is "s2" at [kcp://127.0.0.1:22002]
[I6KAH] 08:24:25 INFO: Device 373HSRP-QLPNLIE-JYKZVQF-P4PKZ63-R2ZE6K3-YD442U2-JHBGBQG-WWXAHAU is "s3" at [kcp://127.0.0.1:22003]
[I6KAH] 08:24:25 INFO: Device 7PBCTLL-JJRYBSA-MOWZRKL-MSDMN4N-4US4OMX-SYEXUS4-HSBGNRY-CZXRXAT is "s4" at [kcp://127.0.0.1:22004]
[I6KAH] 08:24:25 INFO: No automatic upgrades; STNOUPGRADE environment variable defined.
[I6KAH] 08:24:25 INFO: Completed initial scan (rw) of folder default
[I6KAH] 08:24:25 INFO: GUI and API listening on 127.0.0.1:8081
[I6KAH] 08:24:25 INFO: Access the GUI via the following URL: http://127.0.0.1:8081/
^C[I6KAH] 08:24:27 INFO: Exiting
[monitor] 08:24:27 INFO: Signal 2 received; exiting
jb@syno:~/s/g/s/s/test $ 
```

AudriusButkevicius:
Strange, the only issue I observed is not accepting new connections. Might be an OS X thing.

AudriusButkevicius:
Or OS X firewall thing. Did you try signing the binary

calmh:
No, but we don't have any issues listening on UDP for discovery. I'll look into it later!

AudriusButkevicius:
So fixed the issue with listener blocking forever after the first connection.

Seems to work every time, 50% of the time.

Probably leaks go routines, but there is so many layers of the onion, making it very painful to debug.
kcp seems to be fairly persistent in keeps flooding udp packets like there is no tomorrow, perhaps related to the leak...

```
net.PacketConn -> pfilter.FilteredConnection splits into:
    -> Stun connection
    -> KCP Listener -> Yamux -> TLS -> BEP
    -> KCP Dialer 0 -> Yamux -> TLS -> BEP
    ...
    -> KCP Dialer N -> Yamux -> TLS -> BEP
```

pfilter allows multiplexing packets into multiple virtual connections based on the packet signature
kcp only provides reliable delivery, no control flow
yamux provides control flow, close, keepalive etc.

Anyways, it would be nice to review/test/benchmark before I waste more effort on this just to be told that it won't work for us.

AudriusButkevicius:
I think I've addressed the routine leak

calmh:
Works and connects locally for me at least, although not benchmarked as there is a cpu utilization issue. The STUN thing seems to spin and take up a full core, when just starting up with the h1/config.xml config: https://nym.se/r/lbq1.svg

AudriusButkevicius:
Will fix

calmh:
```
jb@syno:~ $ syncthing -home /Volumes/Untitled
[monitor] 09:20:39 INFO: Starting syncthing
[start] 09:20:39 INFO: Generating ECDSA key and certificate for syncthing...
[XE7ET] 09:20:39 INFO: syncthing v0.14.4+18-g7f5bb04-dirty-pr-3489 "Dysprosium Dragonfly" (go1.7rc6 darwin-amd64) jb@syno.nym.se 2016-08-18 18:53:55 UTC
[XE7ET] 09:20:39 INFO: My ID: XE7ET75-XUA3TYO-VQGD5V2-FE2KBS3-2MWLBUP-NLVHHYM-I5HCCWV-EM3A6QW
[XE7ET] 09:20:40 INFO: Single thread hash performance is ~374 MB/s
[XE7ET] 09:20:40 INFO: Default folder created and/or linked to new config
[XE7ET] 09:20:40 FATAL: get free port (BEP/kcp): listen udp 0.0.0.0:0: unexpected address type 0.0.0.0:0
[monitor] 09:20:40 INFO: Syncthing exited: exit status 1
```

AudriusButkevicius:
Another thing to point out, if we use the listeners connection for dialing, the dialed connections will break if someone removes the listener (which will close it's socket). Do we care about this?

calmh:
No. I'd consider that expected, to the point that I was for a moment slightly surprised to see the TCP connections stick around when I disabled the TCP listener.

AudriusButkevicius:
Well, I am talking about dialed connections, not accepted connections

calmh:
Yes, either way - I don't mind them being closed (or not closed for that matter).

AudriusButkevicius:
This is brewing on getting my changes merged upstream.

AudriusButkevicius:
So all changes upstream have been accepted.
I've switched the flow control library to one that the kcp guy implemented, which supposedly doesn't have a 27Mbps performance cap.

Benching on windows is a pain as the processes run away, and on a linux VM I get the same 6MiBs on tcp and kcp, so I'd be nice if you could bench.

If it performs much worse, I might want to try a few builds, one stripping out the packet filter stuff, one switching back to yamux to see which one offers the best speed and where the bottleneck is.

calmh:
Exciting, I shall take another look!

calmh:
@st-jenkins test this please if you don't mind?

calmh:
I... guess not.

calmh:
I get 15 MB/s locally on my laptop, which is a tiny fraction of the TCP performance but also completely acceptable given the constrains, in my opinion. I see it's nicely rebased on top of latest master, so let me deploy this to a couple of machines and test it over the actual internet for a couple of days

AudriusButkevicius:
I assume this is the 2G benchmark?

So do you mind if I make a few more PRs with different tweaks to the same setup for you to run benchmarks as?

I'd do it myself, but on Windows the benchmarks are very inconsistent, and the only feasible Linux env I have reports 6Mbps for TCP....

Unless I could do benchmarks on the DO build machine or something.

calmh:
On Laptop->Wifi->Switch->Server at home it sticks at around 1.8-2.0 MB/s which is less, perhaps not completely unacceptable (I get around 50 MB/s with TCP here).

There is some weirdness though - from my home server it connected out to another server in Germany, then timed out (I think, as it's exactly one minute after the connect) and hasn't reconnected since. I'll add some tracing to the mix to see what happens.

```
Nov 25 19:51:03 pseudo syncthing: [ITZRN] INFO: kcp://:34567 detected NAT type: Symetric NAT
Nov 25 19:51:03 pseudo syncthing: [ITZRN] INFO: kcp://:34567 resolved external address kcp://94.255.169.149:17142
Nov 25 19:51:07 pseudo syncthing: [ITZRN] INFO: Established secure connection to SYNO4VL-TKVFF25-QKC2T3V-TTK4BGU-D2ZA5OT-PDKSI3R-3P4JQVS-BL4VXAC at [::]:34567-[2001:7a0:8003:32:196f:6406:8172:d9c1]:34567 (KCP (Client))
Nov 25 19:51:07 pseudo syncthing: [ITZRN] INFO: Device SYNO4VL-TKVFF25-QKC2T3V-TTK4BGU-D2ZA5OT-PDKSI3R-3P4JQVS-BL4VXAC client is "syncthing v0.14.12+32-g3742439-pr-3489" named "syno.nym.se"
Nov 25 19:51:07 pseudo syncthing: [ITZRN] INFO: Puller (folder "documents", dir "Web Receipts"): delete: remove /data/documents/Web Receipts: directory not empty
Nov 25 19:51:10 pseudo syncthing: [ITZRN] INFO: Established secure connection to GB33MBV-EKIA4IA-RNNU2YP-MHYRNL7-AWPB5JR-CDJNPRN-2NJXZUA-X4Z7FAL at [::]:34567-176.9.46.83:34502 (KCP (Client))
Nov 25 19:51:10 pseudo syncthing: [ITZRN] INFO: Device GB33MBV-EKIA4IA-RNNU2YP-MHYRNL7-AWPB5JR-CDJNPRN-2NJXZUA-X4Z7FAL client is "syncthing v0.14.12+32-g3742439" named "flk1.kastelo.net"
Nov 25 19:51:10 pseudo syncthing: [ITZRN] INFO: Puller (folder "documents", dir "Web Receipts"): delete: remove /data/documents/Web Receipts: directory not empty
Nov 25 19:51:15 pseudo syncthing: [ITZRN] INFO: Completed initial scan (ro) of folder foto
Nov 25 19:52:10 pseudo syncthing: [ITZRN] INFO: Connection to GB33MBV-EKIA4IA-RNNU2YP-MHYRNL7-AWPB5JR-CDJNPRN-2NJXZUA-X4Z7FAL closed: reading length: broken pipe
(silence since then, it's now 20:03)
```

Yes, the benchmark above was the 2G one, and I'm doing roughly the same (just adding a 2GB file of random data to a folder) in my real setup above. Do add whatever and I'll test.

AudriusButkevicius:
Can you run the same 2G benchmark on each of the two commits I've added?

AudriusButkevicius:
@st-jenkins test this please

calmh:
Looks like a factor two improvement at least:

```plain
jb@unu:~/s/g/s/s/test $ go test -v -tags integration\ benchmark -run LargeFile2G
=== RUN   TestBenchmarkTransferLargeFile2G
2016/12/14 10:54:14 Cleaning...
2016/12/14 10:54:14 Generating files...
2016/12/14 10:54:21 Total 2048.0 MiB in 1 files
2016/12/14 10:54:21 Starting instance 1...
2016/12/14 10:54:29 Starting instance 2...
2016/12/14 10:54:30 {3 2016-12-14 10:54:30.935495003 +0100 CET StateChanged map[folder:default from:idle to:scanning]}
2016/12/14 10:54:30 {8 2016-12-14 10:54:30.936461025 +0100 CET StateChanged map[from:scanning to:idle duration:0.000966473 folder:default]}
2016/12/14 10:54:31 {229 2016-12-14 10:54:31.062619676 +0100 CET StateChanged map[folder:default from:idle to:syncing duration:0.12615848200000002]}
2016/12/14 10:55:37 {456 2016-12-14 10:55:37.303130878 +0100 CET StateChanged map[duration:66.240510769 folder:default from:syncing to:idle]}
2016/12/14 10:55:37 Verifying...
2016/12/14 10:55:42 Result: Wall time: 1m6.240511202s
2016/12/14 10:55:42 Result: 30.9 MiB/s synced
2016/12/14 10:55:42 Receiver: Utime: 44.038398s
2016/12/14 10:55:42 Receiver: Stime: 28.091541s
2016/12/14 10:55:42 Receiver: MaxRSS: 63644 KiB
2016/12/14 10:55:42 Sender: Utime: 26.421233s
2016/12/14 10:55:42 Sender: Stime: 23.716282s
2016/12/14 10:55:42 Sender: MaxRSS: 43740 KiB
--- PASS: TestBenchmarkTransferLargeFile2G (87.94s)
PASS
ok  	github.com/syncthing/syncthing/test	87.955s
```

AudriusButkevicius:
Thats on the last commit, can you run the benchmark on previous to last (where we use smux but not my packet filter magic), as I suspect thats the bottleneck.

calmh:
jb@unu:~/s/g/s/s/test $ go test -v -tags integration\ benchmark -run LargeFile2G
=== RUN   TestBenchmarkTransferLargeFile2G
2016/12/15 17:08:17 Cleaning...
2016/12/15 17:08:17 Generating files...
2016/12/15 17:08:25 Total 2048.0 MiB in 1 files
2016/12/15 17:08:25 Starting instance 1...
2016/12/15 17:08:33 Starting instance 2...
2016/12/15 17:08:34 {4 2016-12-15 17:08:34.662534496 +0100 CET StateChanged map[to:scanning folder:default from:idle]}
2016/12/15 17:08:34 {7 2016-12-15 17:08:34.663243389 +0100 CET StateChanged map[duration:0.0007090460000000001 folder:default from:scanning to:idle]}
2016/12/15 17:08:34 {226 2016-12-15 17:08:34.85003731 +0100 CET StateChanged map[duration:0.18679330700000002 folder:default from:idle to:syncing]}
2016/12/15 17:10:48 {301 2016-12-15 17:10:48.324956455 +0100 CET StateChanged map[to:idle duration:133.474918888 folder:default from:syncing]}
2016/12/15 17:10:48 Verifying...
2016/12/15 17:10:53 Result: Wall time: 2m13.474919145s
2016/12/15 17:10:53 Result: 15.3 MiB/s synced
2016/12/15 17:10:53 Receiver: Utime: 49.413962s
2016/12/15 17:10:53 Receiver: Stime: 33.458984s
2016/12/15 17:10:53 Receiver: MaxRSS: 56288 KiB
2016/12/15 17:10:53 Sender: Utime: 30.835115s
2016/12/15 17:10:53 Sender: Stime: 27.874s
2016/12/15 17:10:53 Sender: MaxRSS: 45228 KiB
--- PASS: TestBenchmarkTransferLargeFile2G (155.79s)
PASS
ok  	github.com/syncthing/syncthing/test	155.806s
jb@unu:~/s/g/s/s/test $ git describe
v0.14.12-33-g832290a

calmh:
```
jb@unu:~/s/g/s/s/test $ go test -v -tags integration\ benchmark -run LargeFile2G
=== RUN   TestBenchmarkTransferLargeFile2G
2016/12/15 17:11:42 Cleaning...
2016/12/15 17:11:42 Generating files...
2016/12/15 17:11:49 Total 2048.0 MiB in 1 files
2016/12/15 17:11:49 Starting instance 1...
2016/12/15 17:11:58 Starting instance 2...
2016/12/15 17:11:59 {2 2016-12-15 17:11:59.189166384 +0100 CET StateChanged map[folder:default from:idle to:scanning]}
2016/12/15 17:11:59 {6 2016-12-15 17:11:59.191219989 +0100 CET StateChanged map[duration:0.0020534710000000003 folder:default from:scanning to:idle]}
2016/12/15 17:11:59 {225 2016-12-15 17:11:59.352776507 +0100 CET StateChanged map[folder:default from:idle to:syncing duration:0.16155608400000002]}
2016/12/15 17:14:14 {302 2016-12-15 17:14:13.901569244 +0100 CET StateChanged map[from:syncing to:idle duration:134.548792129 folder:default]}
2016/12/15 17:14:14 Verifying...
2016/12/15 17:14:18 Result: Wall time: 2m14.548792737s
2016/12/15 17:14:18 Result: 15.2 MiB/s synced
2016/12/15 17:14:18 Receiver: Utime: 46.847315s
2016/12/15 17:14:18 Receiver: Stime: 32.208597s
2016/12/15 17:14:18 Receiver: MaxRSS: 56724 KiB
2016/12/15 17:14:18 Sender: Utime: 29.523931s
2016/12/15 17:14:18 Sender: Stime: 26.390305s
2016/12/15 17:14:18 Sender: MaxRSS: 43764 KiB
--- PASS: TestBenchmarkTransferLargeFile2G (156.37s)
PASS
ok  	github.com/syncthing/syncthing/test	156.392s
jb@unu:~/s/g/s/s/test $ git describe
v0.14.12-33-g832290a
```

calmh:
So looks a bit variable, results are a bit different now than before and no difference between the last two commits that I can measure.

calmh:
```
jb@unu:~/s/g/s/s/test $ go test -v -tags integration\ benchmark -run LargeFile2G
=== RUN   TestBenchmarkTransferLargeFile2G
2016/12/15 17:16:28 Cleaning...
2016/12/15 17:16:28 Generating files...
2016/12/15 17:16:35 Total 2048.0 MiB in 1 files
2016/12/15 17:16:35 Starting instance 1...
2016/12/15 17:16:44 Starting instance 2...
2016/12/15 17:16:45 {5 2016-12-15 17:16:45.0674912 +0100 CET StateChanged map[to:scanning folder:default from:idle]}
2016/12/15 17:16:45 {7 2016-12-15 17:16:45.068028802 +0100 CET StateChanged map[to:idle duration:0.000537854 folder:default from:scanning]}
2016/12/15 17:16:45 {226 2016-12-15 17:16:45.241558734 +0100 CET StateChanged map[duration:0.173529341 folder:default from:idle to:syncing]}
2016/12/15 17:18:07 {276 2016-12-15 17:18:07.577178516 +0100 CET StateChanged map[duration:82.33561961 folder:default from:syncing to:idle]}
2016/12/15 17:18:07 Verifying...
2016/12/15 17:18:12 Result: Wall time: 1m22.335619782s
2016/12/15 17:18:12 Result: 24.9 MiB/s synced
2016/12/15 17:18:12 Receiver: Utime: 40.336608s
2016/12/15 17:18:12 Receiver: Stime: 25.907658s
2016/12/15 17:18:12 Receiver: MaxRSS: 58096 KiB
2016/12/15 17:18:12 Sender: Utime: 25.280612s
2016/12/15 17:18:12 Sender: Stime: 22.356797s
2016/12/15 17:18:12 Sender: MaxRSS: 44236 KiB
--- PASS: TestBenchmarkTransferLargeFile2G (104.19s)
PASS
ok  	github.com/syncthing/syncthing/test	104.205s
jb@unu:~/s/g/s/s/test $ git describe
v0.14.12-34-g1b35aa1
```

AudriusButkevicius:
Thats strange, I would have expected a different result, so I guess yamux is better, yet smux guy wrote it to work around issues of yamux.

 I'll make another version which essentially removes the  8322 commit to see how much my packet filterig stuff contributes to slowing the 30MBps down.

AudriusButkevicius:
Can you bench again please. Thanks.

calmh:
```
jb@unu:~/s/g/s/s/test $ go test -v -tags integration\ benchmark -run LargeFile2G
=== RUN   TestBenchmarkTransferLargeFile2G
2016/12/19 11:00:31 Cleaning...
2016/12/19 11:00:31 Generating files...
2016/12/19 11:00:38 Total 2048.0 MiB in 1 files
2016/12/19 11:00:38 Starting instance 1...
2016/12/19 11:00:46 Starting instance 2...
2016/12/19 11:00:47 {3 2016-12-19 11:00:47.565910348 +0100 CET StateChanged map[folder:default from:idle to:scanning]}
2016/12/19 11:00:47 {7 2016-12-19 11:00:47.566549817 +0100 CET StateChanged map[duration:0.00064124 folder:default from:scanning to:idle]}
2016/12/19 11:00:47 {15 2016-12-19 11:00:47.646642443 +0100 CET StateChanged map[duration:0.080092586 folder:default from:idle to:syncing]}
2016/12/19 11:02:05 {59 2016-12-19 11:02:05.19472695 +0100 CET StateChanged map[duration:77.548084092 folder:default from:syncing to:idle]}
2016/12/19 11:02:05 Verifying...
2016/12/19 11:02:10 Result: Wall time: 1m17.548084507s
2016/12/19 11:02:10 Result: 26.4 MiB/s synced
2016/12/19 11:02:10 Receiver: Utime: 44.213309s
2016/12/19 11:02:10 Receiver: Stime: 28.213175s
2016/12/19 11:02:10 Receiver: MaxRSS: 57076 KiB
2016/12/19 11:02:10 Sender: Utime: 26.368207s
2016/12/19 11:02:10 Sender: Stime: 24.299111s
2016/12/19 11:02:10 Sender: MaxRSS: 41876 KiB
--- PASS: TestBenchmarkTransferLargeFile2G (99.13s)
PASS
ok  	github.com/syncthing/syncthing/test	99.140s
jb@unu:~/s/g/s/s/test $ git describe
v0.14.12-35-ged537b5
```

Honestly I expect that the framing has fairly little impact given the massive per packet syscall overhead or so.

AudriusButkevicius:
Anyways, it seems that the performance combo we've got now makes most sense. 
So the last remaining complain is that it does not reconnect?

calmh:
I think it did not connect at all in at least one scenario for me. Do you have this working somewhere?

AudriusButkevicius:
```
[K732B] 2016/12/21 17:43:56.619803 service.go:301: INFO: Established secure connection to F72F5AO-37T2W7K-BC43BAN-ZDMWZF6-KR7SSHD-O6YQ7GM-CR2XWJG-UOTCTA5 at [::]:22040-xxx:22040 (kcp-client) (TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384)
```

connect fine for me

AudriusButkevicius:
can you repro?

AudriusButkevicius:
So what's the plan with this one?

calmh:
I want it to be actually running for a while in a setup to iron out issues. Last time I tried in mine it didn't work. Do you have one where you can do so? Otherwise I'll happily retry, but if it doesn't connect I need to revert (after suitable amounts of troubleshooting).

AudriusButkevicius:
Well for me it seems to work. Can you set me up in a place where it doesn't work?

AudriusButkevicius:
Take it for another spin, with STTRACE=connections and let me know if anything obvious crops up.

calmh:
Deploying now

calmh:
This is the setup:

```
ITZRN -- [evil firewall] -- [the internet] -- GB33M
```

ITZRN is at my home, behind NAT and firewall. It's expected that no inbound connections to it can succeed. GB33M is on the internet, public IP, no filtering at all. Both are configured with `kcp://34567` or similar as the listen address, and have global discovery running.

Things did actually connect just fine and is shuffling data now over kcp over ipv6.

screen shot 2017-01-26 at 13 11 42

Supposedly that side is "kcp-client" which isn't technically possibly but does show that they managed to rendezvous with each other using the same ports and get through the firewall. I'll do some more experiments over the coming days.

calmh:
Connection is still established a day later, so it's not entirely broken.

lib/connections/kcp_listen.go
```diff
@@ -0,0 +1,263 @@
...
+		for _, addr := range t.cfg.StunServers() {
+			client.SetServerAddr(addr)
+
+			natType, extAddr, err = client.Discover()
+			if err != nil || extAddr == nil {
+				l.Debugf("%s stun discovery on %s: %s (%v)", t.uri, addr, err, extAddr)
+				continue
+			}
+
+			t.mut.Lock()
```
calmh: What is the locking for?

lib/connections/kcp_misc.go
```diff
@@ -0,0 +1,172 @@
...
+	"bytes"
+	"encoding/binary"
+	"net"
+	"sort"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/hashicorp/yamux"
+
```
calmh: blank line

lib/connections/kcp_misc.go
```diff
@@ -0,0 +1,172 @@
...
+	if host, _, err := net.SplitHostPort(f[j].LocalAddr().String()); err == nil {
+		jIsUnspecified = net.ParseIP(host).IsUnspecified()
+	}
+	return (iIsUnspecified && !jIsUnspecified) || (iIsUnspecified && jIsUnspecified)
+}
+
+// As we open listen connections, we register them here, so that Dial calls
+// could reuse them. This way we will hopefully work around restricted NATs
+// by dialing via the same connection we are listening on, creating a mapping
+// on our NAT to that IP, and hoping that the other end will try to dialour
```
calmh: "dialour"

lib/connections/kcp_misc.go
```diff
@@ -0,0 +1,172 @@
...
+
+func deregisterFilter(filter *pfilter.PacketFilter) {
+	mut.Lock()
+	defer mut.Unlock()
+
+	for i, f := range filters {
+		if f == filter {
+			copy(filters[i:], filters[i+1:])
+			filters[len(filters)-1] = nil
+			filters = filters[:len(filters)-1]
```
calmh: break here?

lib/connections/kcp_misc.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (f *kcpConversationFilter) Outgoing(out []byte, addr net.Addr) {
+	if !f.isKCPConv(out) {
+		panic("not a kcp conversation")
+	}
+	atomic.StoreUint32(&f.convID, binary.LittleEndian.Uint32(out[:4]))
+}
+
+func (kcpConversationFilter) isKCPConv(data []byte) bool {
+	// First 4 bytes convID
```
calmh: `if len(data) < 5 { return false}`

lib/connections/kcp_misc.go
```diff
@@ -0,0 +1,172 @@
...
+	// 5th byte is cmd
+	// IKCP_CMD_PUSH    = 81 // cmd: push data
+	// IKCP_CMD_ACK     = 82 // cmd: ack
+	// IKCP_CMD_WASK    = 83 // cmd: window probe (ask)
+	// IKCP_CMD_WINS    = 84 // cmd: window size (tell)
+	return 80 < data[4] && data[4] < 85
+}
+
+func (f *kcpConversationFilter) ClaimIncoming(in []byte, addr net.Addr) bool {
+	if f.isKCPConv(in) {
```
calmh: check len here as well

lib/connections/kcp_misc.go
```diff
@@ -0,0 +1,172 @@
...
+		_, ok := f.ids[id]
+		f.reap()
+		f.mut.Unlock()
+		return ok
+	}
+	return false
+}
+
+func (f *stunFilter) isStunPayload(data []byte) bool {
+	// First two bits always unset, and should always send magic cookie.
```
calmh: check len first, and maybe also that the length is sufficient for the things that will try to use the stun package if we return true

lib/connections/tcp_dial.go
```diff
@@ -31,7 +31,7 @@ type tcpDialer struct {
 }
 
 func (d *tcpDialer) Dial(id protocol.DeviceID, uri *url.URL) (internalConn, error) {
-	uri = fixupPort(uri)
+	uri = fixupPort(uri, 22000)
```
calmh: constant

lib/connections/tcp_listen.go
```diff
@@ -181,7 +180,7 @@ type tcpListenerFactory struct{}
 
 func (f *tcpListenerFactory) New(uri *url.URL, cfg *config.Wrapper, tlsCfg *tls.Config, conns chan internalConn, natService *nat.Service) genericListener {
 	return &tcpListener{
-		uri:        fixupPort(uri),
+		uri:        fixupPort(uri, 22000),
```
calmh: constant

lib/connections/util.go
```diff
@@ -0,0 +1,29 @@
...
+package connections
+
+import (
+	"net"
+	"net/url"
+	"strconv"
+	"strings"
+)
+
+func fixupPort(uri *url.URL, replacementPort int) *url.URL {
```
calmh: `defaultPort` I think, we are not replacing anything.

vendor/github.com/AudriusButkevicius/pfilter/conn.go
```diff
@@ -0,0 +1,93 @@
...
+	select {
+	case <-r.closed:
+		return 0, nil, errClosed
+	default:
+	}
+
+	var timeout <-chan time.Time
+
+	if deadline, ok := r.deadline.Load().(time.Time); ok && !deadline.IsZero() {
+		timeout = time.After(deadline.Sub(time.Now()))
```
calmh: Will most of our calls have a read deadline or no? If no, this if fine. If yes, we should probably not spawn a goroutine for every read.

vendor/github.com/AudriusButkevicius/pfilter/misc.go
```diff
@@ -0,0 +1,36 @@
+package pfilter
+
+import (
+	"fmt"
+	"net"
+	"sync"
+)
+
+var (
+	maxPacketSize = 2048
```
calmh: How is this decided?

lib/config/config.go
```diff
@@ -37,6 +37,7 @@ var (
 	// config.
 	DefaultListenAddresses = []string{
 		"tcp://0.0.0.0:22000",
+		"kcp://0.0.0.0:22020",
```
AudriusButkevicius: Not sure, but for example "port 22000 in use" can mean either, hence less confusion?

lib/config/config.go
```diff
@@ -56,7 +57,12 @@ var (
 	// DefaultDiscoveryServers should be substituted when the configuration
 	// contains <globalAnnounceServer>default</globalAnnounceServer>.
 	DefaultDiscoveryServers = append(DefaultDiscoveryServersV4, DefaultDiscoveryServersV6...)
-
+	// DefaultStunServers should be substituted when the configuration
+	// contains <stunServer>default</stunServer>.
+	DefaultStunServers = []string{
+		"stun.ekiga.net:3478",
+		"stun1.voiceeclipse.net:3478",
+	}
```
AudriusButkevicius: We probably need to do something about these, as I am sure we'll  just get banned somehow by this.

lib/connections/kcp_dial.go
```diff
@@ -0,0 +1,109 @@
...
+func (d *kcpDialer) Dial(id protocol.DeviceID, uri *url.URL) (internalConn, error) {
+	uri = fixupPort(uri, 22020)
+
+	// Try to dial via an existing listening connection
+	// giving better changes punching through NAT.
+	f := getDialingFilter()
+	var conn *kcp.UDPSession
+	var err error
+	if f != nil {
+		conn, err = kcp.NewConn(uri.Host, nil, 0, 0, f.NewConn(20, &kcpConversationFilter{}))
```
AudriusButkevicius: this is this filter priority thingy, but I guess all of these should just be constants.

lib/connections/kcp_dial.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (d *kcpDialer) Dial(id protocol.DeviceID, uri *url.URL) (internalConn, error) {
+	uri = fixupPort(uri, 22020)
+
+	// Try to dial via an existing listening connection
+	// giving better changes punching through NAT.
+	f := getDialingFilter()
+	var conn *kcp.UDPSession
+	var err error
+	if f != nil {
```
AudriusButkevicius: Not sure I follow, given getDialingFilter returns the value which we decide upon if we enter the if?

lib/connections/kcp_dial.go
```diff
@@ -0,0 +1,109 @@
...
+		conn, err = kcp.NewConn(uri.Host, nil, 0, 0, f.NewConn(20, &kcpConversationFilter{}))
+		// We are piggy backing on a listener connection, no need for keepalives.
+		// Futhermore, keepalives just send garbage, which will flip our filter.
```
AudriusButkevicius: I'll change this to be more understandable, but what I meant was that it will cause us to receive packets that we cannot attribute to any filter we have set up which will cause them be logged/dropped.

lib/connections/kcp_dial.go
```diff
@@ -0,0 +1,109 @@
...
+	} else {
+		conn, err = kcp.DialWithOptions(uri.Host, nil, 0, 0)
+	}
+	if err != nil {
+		l.Debugln(err)
+		conn.Close()
+		return internalConn{}, err
+	}
+
+	conn.SetWindowSize(128, 128)
```
AudriusButkevicius: it's send receive windows. I ripped those out of the kcptun as defaults, but they changed. I'll move them to constants and copy a few other things he added since.

lib/connections/kcp_listen.go
```diff
@@ -0,0 +1,263 @@
...
+		l.Infoln("listen (BEP/kcp):", err)
+		return
+	}
+	filterConn := pfilter.NewPacketFilter(packetConn)
+	kcpConn := filterConn.NewConn(100, nil)
+	stunConn := filterConn.NewConn(10, &stunFilter{
+		ids: make(map[string]time.Time),
+	})
+
+	filterConn.Start()
```
AudriusButkevicius: It exits on first ReadFrom(), which should happen after a .Close():

https://github.com/AudriusButkevicius/pfilter/blob/master/filter.go#L118

vendor/github.com/AudriusButkevicius/pfilter/misc.go
AudriusButkevicius: Random number larger than most MTUs? 
Have a better idea? I guess I could explicitly set MTU to 1500 in the KCP library.

vendor/github.com/AudriusButkevicius/pfilter/conn.go
AudriusButkevicius: Probably.

lib/connections/kcp_misc.go
AudriusButkevicius: Checking inside anyway.

lib/connections/kcp_misc.go
AudriusButkevicius: added it, so it's definately there, but github failing to update display?

lib/config/config.go
calmh: Maybe add a few more from the lists of public ones, and randomize the list before using? I think we are small enough to not have that much of an impact, and these things seem public enough.

lib/config/optionsconfiguration.go
```diff
@@ -131,6 +131,8 @@ type OptionsConfiguration struct {
 	UnackedNotificationIDs  []string                `xml:"unackedNotificationID" json:"unackedNotificationIDs"`
 	TrafficClass            int                     `xml:"trafficClass" json:"trafficClass"`
 	WeakHashSelectionMethod WeakHashSelectionMethod `xml:"weakHashSelectionMethod" json:"weakHashSelectionMethod"`
+	StunServers             []string                `xml:"stunServer" json:"stunServers" default:"default"`
+	StunKeepaliveS          int                     `xml:"stunKeepaliveSeconds" json:"stunKeepaliveSeconds" default:"24"`
```
calmh: What's the procedure for disabling STUN?

lib/connections/kcp_dial.go
calmh: I was thinking `if f := getDialingFilter(); f != nil { ...` because there was some distance between the assignment and the `if` and it doesn't look used afterwards

vendor/github.com/AudriusButkevicius/pfilter/misc.go
calmh: There should be some correlation at least. A random UDP datagram can be up to 65507 bytes in size, which this does not handle. Maybe KCP never creates datagrams that big, but then we should know and have a not about what that maximum size is.

lib/connections/constants.go
```diff
@@ -0,0 +1,27 @@
...
+	kcpNoFilterPriority           = 100
+	kcpConversationFilterPriority = 20
+	kcpStunFilterPriority         = 10
+
+	kcpNoDelay                 = 1
+	kcpInterval                = 10
+	kcpResend                  = 2
+	kcpNoCongestion            = 1
+	kcpKeepAlive               = 10
+	kcpClientSendWindowSize    = 128
```
calmh: Why have the client and server values different?

AudriusButkevicius: Because that is what kcptun is using.

calmh: But aren't the two sides equivalent and it's random who gets to be server and client? That might not be the case in their gaming use case.

AudriusButkevicius: Makes sense.

vendor/github.com/AudriusButkevicius/pfilter/misc.go
AudriusButkevicius: Are you ok to set this to 65507?

lib/config/optionsconfiguration.go
AudriusButkevicius: Should we have a separate boolean?

AudriusButkevicius:
So things oustanding are:
1. Stun disablement, how we want to handle it.
2. pfilter MTU set to 1500 given [https://github.com/xtaci/kcp-go/blob/master/kcp.go#L903](this) and given we do not specify a custom MTU.

Can you give it a second proper shake and see what's outstanding?

AudriusButkevicius:
Ghrr the shuffle thing requires Go 1.8, which I guess we'll have available everywhere by the time this gets merged.

calmh:
Even with 1.8 planned for release mid feb, I would really like to remain buildable with 1.7 for a while. Is this something essential? I haven't reviewed the new changes yet.

AudriusButkevicius:
I'll try to copy the implementation of swapper from 1.8 into 1.7. i'll look into logging.
What about disabling stun?

calmh:
Yeah given that it talks all the time to foreign servers I think we need to be able to do that. Can KCP run without STUN? If not I guess they need to be disabled together. I don't particularly care if it's a new boolean or happens when the interval is set to zero or something like that.

AudriusButkevicius:
Well if the other side has stun enabled it might still succeed.

calmh:
Yes, or we might even be behind not-NAT (although no good reason to run KCP then). So yeah, no need to tie the two together. Then I think setting the stun interval to zero to disable would at least be consistent with how we do other things. In the GUI, it could still be represented as a checkbox if we wanted to represent it.

AudriusButkevicius:
Fine I'll do it tonight. Give the PR another grilling so I could hopefully wrap it up.

calmh:
One thing I want tested before merging, that I can't currently easily test, is what happens when there is limited bandwidth. Like, on the old 10/0.5 ADSL I had, when uploading it saturated the 0.5 Mbps uplink to the point where internet was for all intents and purposes "down" for everyone else. That's not a good thing if it still happens.

AudriusButkevicius:
tc to the rescue?

calmh:
I could set a rate limit on the firewall and see what happens, yeah. Syncthing rate limits are nice and all, but sort of beside the point. I'm not saying it does, but killing someones internet connection really isn't acceptable default behavior. :)

AudriusButkevicius:
Well what exactly are the options in that case? This requires congestion control implementations which I am sure KCP does not support.

calmh:
It must have some sort of rate back off in reaction to packet loss or delay, otherwise it wouldn't work at all basically?

AudriusButkevicius:
Well I suspect SetNoDelay controls that, but it's not automatic.

calmh:
I'll see if i can figure out how it behaves.

calmh:
Good new, and bad news. I set the rate on my router to 500Kbps, and it behaved just fine - hovering around 250-350 Kbps of actual transfer and not fucking up my regular surfing. If anything it may be *too* sensitive as it could have used more bandwidth, but I don't really care.

However, there are bugs. Changing the shaping rate on my router resets "something" (security flows, NAT mappings, I don't know) so that the connection is dropped. After that, it doesn't re-establish. Both sides are trying to dial, but hit a timeout and/or that yamux error:

```
[GB33M] 2017/02/11 11:09:15.784683 service.go:314: DEBUG: Reconnect loop
[GB33M] 2017/02/11 11:09:15.784800 service.go:341: DEBUG: Reconnect loop for ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM
[GB33M] 2017/02/11 11:09:19.468081 service.go:390: DEBUG: dial ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM kcp://94.255.169.149:24615
[GB33M] 2017/02/11 11:09:19.468143 kcp_dial.go:46: DEBUG: dial kcp://94.255.169.149:24615 using existing conn on [::]:34502
2017/02/11 11:09:34 [ERR] yamux: Failed to read header: broken pipe
[GB33M] 2017/02/11 11:09:34.469324 service.go:395: DEBUG: dial failed ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM kcp://94.255.169.149:24615 i/o deadline reached
[GB33M] 2017/02/11 11:09:34.469399 service.go:390: DEBUG: dial ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM kcp://94.255.169.149:32497
[GB33M] 2017/02/11 11:09:34.469450 kcp_dial.go:46: DEBUG: dial kcp://94.255.169.149:32497 using existing conn on [::]:34502
2017/02/11 11:09:49 [ERR] yamux: Failed to read header: broken pipe
[GB33M] 2017/02/11 11:09:49.470558 service.go:395: DEBUG: dial failed ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM kcp://94.255.169.149:32497 i/o deadline reached
[GB33M] 2017/02/11 11:09:49.470882 service.go:390: DEBUG: dial ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM kcp://94.255.169.149:34567
[GB33M] 2017/02/11 11:09:49.470944 kcp_dial.go:46: DEBUG: dial kcp://94.255.169.149:34567 using existing conn on [::]:34502
2017/02/11 11:10:04 [ERR] yamux: Failed to read header: broken pipe
[GB33M] 2017/02/11 11:10:04.472238 service.go:395: DEBUG: dial failed ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM kcp://94.255.169.149:34567 i/o deadline reached
[GB33M] 2017/02/11 11:10:04.472326 service.go:390: DEBUG: dial ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM kcp://[2001:7a0:8003:32::11]:34567
[GB33M] 2017/02/11 11:10:04.472384 kcp_dial.go:46: DEBUG: dial kcp://[2001:7a0:8003:32::11]:34567 using existing conn on [::]:34502
2017/02/11 11:10:19 [ERR] yamux: Failed to read header: broken pipe
[GB33M] 2017/02/11 11:10:19.473442 service.go:395: DEBUG: dial failed ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM kcp://[2001:7a0:8003:32::11]:34567 i/o deadline reached
[GB33M] 2017/02/11 11:10:19.473466 service.go:411: DEBUG: sleep until next dial 24s
```

Also, while in this state it's burning a full core of CPU power (on both sides):

screen shot 2017-02-11 at 11 06 45

Here's a `web` of the profile: https://kastelo.net/t/pprof001.svg

It's not super easy to interpret, other than that it looks like it's KCP related and probably caused by creating millions of timers that then need handling. Or so it looks to me at least. Here is a file with two CPU profiles and the binary that created them, if you want to look closer into it: https://kastelo.net/t/profiles.tar.gz

Restarting the side behind NAT made them connect again. When connected, CPU usage goes down to ~0 as it should be, on both sides - even the one that was not restarted. So I'm guessing there's a tight timer-creating loop somewhere while it is not connected to anything, or retrying, or something like that... Curiously, just shutting one device down when they were connected and in sync didn't trigger this again. So maybe there's something more to it than just the reconnect loop.

AudriusButkevicius:
How long did you keep it in that broken state?

I

calmh:
I noticed it maybe five minutes after doing the shaping change, when it didn't connect again. Then it persisted for maybe fifteen more minutes while I futzed around looking at logs and stuff. The two profiles I took have timestamps roughly ten minutes apart, so at least that long.

AudriusButkevicius:
So I've killed the yamux logs, made it build with 1.7, and added a super super ugly way of disabling stun (happy to take advice how to make it less ugly).

calmh:
FWIW it happened again (disconnected; spinning CPU) and this time I did QUIT it. Log here: https://kastelo.net/t/messages.log
As you can see from the startup version it's gc9f807a (for interpreting line numbers) plus "dirty" which was my ignore-need-files patch at that time, I don't think it screws with the relevant line numbers.

AudriusButkevicius:
The only runnable things look like:

```
Feb 11 12:49:06 pseudo syncthing: goroutine 5654 [runnable]:
Feb 11 12:49:06 pseudo syncthing: 	/usr/local/go/src/time/sleep.go:88 +0x52
Feb 11 12:49:06 pseudo syncthing: github.com/syncthing/syncthing/vendor/github.com/xtaci/kcp-go.(*UDPSession).updateTask(0xc4213ae270)
```

AudriusButkevicius:
I think I worked out what's causing it.

The fault is essentially choosing the "fast3" NoDelay options from kcptun, which essentially fires off a timer every 10ms.
I essentially set everything back to defaults, which means the timer fires every 100ms now.

Check out config.go, I know what NoDelay, FastResend and CongestionControl is, yet I am not sure if we should enable or disable them.

I guess you can tweak them around and run speed benchmarks to see what's the end speed you end up with.

AudriusButkevicius:
Can you please take it for another spin?

calmh:
Will do. As for STUN, how about making it an actual service that starts and stops with the KCP listener?

calmh:
What sort of maniac has a function signature that takes a bunch of ints and where 1 means disable something...

AudriusButkevicius:
Apparently @xtaci :D

calmh:
First impression is that it still works just fine, but the transfer rate (in my specific network setup) is reduced from ~3.5 MiB/s before to ~1.5 MiB/s now.

AudriusButkevicius:
Try tweaking the config to your likings.

xtaci:
Impressive work! Any help I can do?

calmh:
@xtaci Oh hi! I think we're mostly good, thanks for putting together this package! One thing you might be able to assist with is give us your opinion on recommended parameters for our use case. We're well aware that our use case probably differs a fair bit from the original KCP use case and we may be doing some shoehorning here... But what we're doing is using it mostly for the purpose of a reliable NAT-busting connection. We run yamux and TLS some other stuff on top of it. I'd like the KCP layer to mostly

 - Be fairly resource constrained when there is not much to do. That is, infrequent keepalives and slower polling/loops, if that applies. We do higher level keepalives on top of KCP every 60 seconds or so anyway.

 - Behave considerately to other traffic and people's internet connections. Since we are not a real time or streaming implementation, we may have a huge amount of data that we're pushing essentially as fast as possible on to KCP. It would be good if this did not result in drowning out short lived TCP flows or fill peoples router buffers. This seems to work perfectly fine as it is now, with congestion control enabled, but if there's something else we should consider please let us know!

 - Otherwise, be as fast as possible :) We seem to get a little different values depending on parameters. Is the transfer speed closely related to some of the timer values? Is there a clear trade off here between performance and CPU usage? (We care more about low CPU usage when idle; when actually shuffling data it doesn't matter as much.)

AudriusButkevicius:
Also from our benchmarks it seems yamux outperformes smux whichwas a bit surprising.

xtaci:
`interval` controls the periodically wake-up of the `updateTask` goroutine for each connection(`UDPSession`). it's about the scheduling in golang, according to my tests, 33ms interval for 3000 connections costs 20% of CPU just for **goroutine scheduling**.  By setting to 40ms interval, costs will be much lower.

`interval` is the `Clock Granularity` in TCP; when estimating round-trip time. The `jitters` of RTT will at last be set to `interval`(the wakeup period). For example, if when measured some RTTs, as 50ms, 49ms, 51ms, and interval is set to 40ms, the RTO(retransmission timeout) will be 90ms; this indirectly affects the **Responsiveness** of the connection，but it's mostly fine for non-interactive applications.

The hard problem for congestion control is the jitters caused by network devices(deep queueing, bursts of sending, propagation latency), which leads to `false positive` of RTO and unexpected window shrinking, and finally stalled connections, that happens everyday on long fat pipe, the longer the road, the more RTT jitters, this leads back to letting the network go idle(under utilization).

In my opinion, it may be necessary to allow users to slightly adjust `nc` and `sndwnd/rcvwnd` options.


As for smux, smux is designed for memory-constrained devices, especially low-end routers. Memory is strictly limited to buffer size in smux, that's why yamux may outperforms smux, the stream flow-control in yamux is quit loose.

calmh:
OK! Here's what I think we should do.

- Merge this.
- Add some advanced configs for these settings so they can be tweaked without rebuild
- Add a feature flag advanced config "enableDefaultKCP" which defaults to false. We only add the kcp listeneder to the "default" listen address set when enabled. Effectively this means KCP will not be used, as it wont be listened on or announced. (But if the other side announces it, we'll connect to it).

We then release an RC with this in, an announce to interested parties that they can turn it on for testing. We catch any disastrous mishaps without killing peoples connections.

Then in the next month or so we set enableDefaultKCP=true for RC builds, false in release builds.

Then we just remove the flag and enable it everywhere.

calmh:

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 0da0774ce4eb5ac9e338dcb21f568e6f07e2101a. Thanks, @AudriusButkevicius!

PR #3515, opened 2016-08-16
Added Core Infrastructure Initiative Badge

dwvisser:
### Purpose

Show off that Syncthing has a 100% passing CII badge.

### Testing

N/A

### Screenshots

N/A

### Documentation

N/A

### Authorship

Dale Visser, a happy user of Syncthing. :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 [![Latest Build (Official)](https://img.shields.io/jenkins/s/http/build.syncthing.net/syncthing.svg?style=flat-square&label=unix%20build)](http://build.syncthing.net/job/syncthing/lastBuild/)
 [![API Documentation](https://img.shields.io/badge/api-Godoc-blue.svg?style=flat-square)](http://godoc.org/github.com/syncthing/syncthing)
 [![MPLv2 License](https://img.shields.io/badge/license-MPLv2-blue.svg?style=flat-square)](https://www.mozilla.org/MPL/2.0/)
+[![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/88/badge)](https://bestpractices.coreinfrastructure.org/projects/88)
 
 This is the Syncthing project which pursues the following goals:
 


calmh:
While I did fill out the form in question, I'm not 100% sure of the value added at this point.

AudriusButkevicius:
I'd rather not have this. Soon our readme will lool like my parents IE with 12 toolbars.

AudriusButkevicius:
Also, I don't really see any benefit to the user, as it's some arbitarty website which I hear about for the first time.

dwvisser:
I would respectfully disagree with the Core Infrastructure Initiative's Best Practices Badge project being "some arbitrary website". You can read all about the purpose of the project at length at https://github.com/nodejs/node/blob/master/README.md, so I won't repeat anything about its virtues here. I will point out, though, that a couple of well-known OSS projects with GitHub presence have already placed the badge on their project pages:

* [Node](https://github.com/nodejs/node/blob/master/README.md)
* [Curl](https://github.com/curl/curl/blob/master/README.md)

AudriusButkevicius:
Hmm, ok, curl and node are pretty big, so perhaps it's a thing nowadays.

calmh:
Yeah it was from Curl (Daniel) that I learned about it, and I think the questions are good starting points for things to consider. I'd need to take another look to see if the answers were entirely truthful before putting up a badge for it though.

david-a-wheeler:
I think the badge *should* be added to the website.  The Linux Foundation isn't just a random organization.

@calmh: A look-through to ensure that the answers are correct makes sense.  I encourage you to fill in the "justifications" for each question to justify why they're correct - then it's much easier to verify later.

(Full disclosure: I'm technical lead for the badging project.  But hopefully you'll agree with me anyway :-) ).

calmh:
@st-review merge it

readme: Add Core Infrastructure Initiative Badge

Skip-check: authors, pr-build

st-review:
:ok_hand: Merged as 3a3c8ec6b85369e8955f102e7eedf00dc4744547. Thanks, @dwvisser!

PR #3519, opened 2016-08-16
lib/model: Correct virtual mtime handling (fixes #3516)

### LINKED ISSUE #3516: Master folder on Linux shared to Macos spurious "out of sync"
This one is easy for me to reproduce... 
(Both machines have ST0.14.4 running.)
* Create empty folder master on Linux (I tried ext4 and vintage ext3 filesystems, no difference in result.)
* Share that master to a MACOS box (with filevault enabled, didn't test on non-filevault fs).
* Create a file on the master "echo >bazbaz"

"bazbaz" is copied to the mac fine, but once the mac does its periodic scan, the
linux box will show "out of sync" with "override changes" button lit up. Once this happens
the folder is essentially "stuck" until you hit the override button. No changes occurred on the
Mac. Since there were no changes on the mac, I didn't expect the "out of sync" to pop up.
Once I hit the override changes, the out-of-sync goes away and doesn't appear to come back
for that file (though subsequent new files will cause it to pop up again.)
Maybe something to do with the new higher-res file times?

This problem didn't exist in 0.13.10.

### Version Information

Syncthing Version: 0.14.4
OS Version: ubuntu 14.04/MacOSX (whatever the latest is)

calmh:
### Purpose

We previously set the mtime on the temp file, and then renamed it to the real path. Unfortunately that means we'd save the real timestamp under the under the temp name ".syncthing.foo.tmp" when the actual file that we will look up on the next scan is "foo". This moves the Chtimes later, ensuring that it gets recorded correctly under the right name.

### Testing

Verified in my FreeBSD + Mac setup. Unfortunately we don't have a good way to test the puller, especially not cross platform... :/

### DIFF
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -1245,9 +1245,6 @@ func (f *rwFolder) performFinish(state *sharedPullerState) error {
 		}
 	}
 
-	// Set the correct timestamp on the new file
-	f.mtimeFS.Chtimes(state.tempName, state.file.ModTime(), state.file.ModTime()) // never fails
-
 	if stat, err := f.mtimeFS.Lstat(state.realName); err == nil {
 		// There is an old file or directory already in place. We need to
 		// handle that.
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
This introduces a race.

T1: Rename to realName
T2: Scan realName detect mtime change
T1: Adjust mtime

calmh:
We don't scan in parallell with pulling, or do you mean that someone not-syncthing might make that change in the meantime?

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 0b1475169f933e6f15f55bc50a663b6aa6339ca9. Thanks, @calmh!

PR #3520, opened 2016-08-16
all: Add folder pause, make pauses optionally permanent (fixes #3407, fixes #215, fixes #3001)

### LINKED ISSUE #215: Per folder "pause sync" toggle
When sync is paused, no data should be transmitted.

There should also be a Rest API for this (pause and resume).
#192

Edit: I think there should also a `-pause` command line flag to launch in paused mode (and/or remember the state on shutdown).

### LINKED ISSUE #3001: Paused/resume device API calls can hang for long periods
I've had an issue reported (https://github.com/canton7/SyncTrayzor/issues/252) where calls to `/rest/system/resume` and `/rest/system/pause` can sometimes hang (for more than 90 seconds, at which point I time out). It appears to happen unpredictably.

It appears that the device in question is never actually paused/resumed, although I'm not 100% sure of that.

I notice that `PauseDevice` and `ResumeDevice` both [acquire a lock](https://github.com/syncthing/syncthing/blob/ea54525a33e4d59be1025d740da6c23c069913ea/lib/model/model.go#L1069) - maybe something's holding this for an extended period?

Syncthing version v0.12.22, Windows.

### LINKED ISSUE #3407: Pause should be permanent
When I pause a device, and restart synching, the pause is removed and the devices are used to share files agin.

I would expect it to be permanent until I press the continue button.

AudriusButkevicius:
### Purpose

Addresses most wanted issue numer 3.

CC: @kozec, @canton7, @Nutomic, @capi
### Testing

Not much, pushed some buttons around in the UI, the UI looked the way it was supposed to, saw log messages that I expected to see.
### Documentation
- [ ] Removal of DevicePaused/Unpaused events
- [ ] Removal of /pause, /resume REST endpoints
- [ ] Removal of 'paused' in connection info
- [ ] Addition of unpaused command line option
- [ ] Addition of UnpauseOnStart option
- [ ] Addition of Paused attribute on Folder/Device segments


AudriusButkevicius:
+119 −116  - 2 features in 3 lines.

canton7:
This will screw up many, many things on my side. 

In particular, I use device pausing and unpausing to disable devices which are connected to over a metered connection. I use the endpoints and events which you've just removed... 

Please don't merge until 0.15!

AudriusButkevicius:
If anything, I think life is easier now, because you just listen to ConfigSaved, and set/read Paused on each device/folder.

Agree, it's not a zero effort change, but I think it's for the better.

canton7:
Not quite. In order to set the config, I need to model it exactly using my data structures, so that I can parse it in its entirety, then set it back with no loss of information. Since the config format changes often, and I need to be backwards compatible, this would mean a lot of different models, and a lot of very similar-looking code around each, to support. 

I've avoided setting the config so far for this reason. 

Before you release this, I need the chance add support for setting config, in parallel with the current approach, and add a switch which selects between the two approaches based on the Syncthing version (which means I need to know ahead of time which version will contain these changes). 

If I don't do this, 5000 Windows users will insta-crash, which won't be pretty.

AudriusButkevicius:
We could maintain backwards compatibility, and deprecate the stuff I removed later to make life easier, but not sure how much of a benefit that is, as it just defers work which will have to be done anyway.

calmh:
Could we not do this and retain the REST methods, having them internally just doing the xo dig updates? The difference would be that the actions are persisted, but otherwise the behavior would be the same

canton7:
Maintaining backwards compatibility, at least for a while, would be a huge help for me. Yes it's just deferring work which needs to be done anyway, but that gives me time to get a fix in place, while allowing you to get your new feature out earlier. 

How hard would it be to keep the events for a while? I won't crash if they're removed, bur I'll do strange things if the user manually pauses and unpauses devices.

Nutomic:
I'm not familiar with the Syncthing source code, so I'm just going by the referenced issues, and it sounds really good! Some questions:
- To pause Syncthing globally, I will have to iterate over all folders in the config and pause them, right?
- Can I add a folder in paused state? So that people can add folders while Syncthing is paused, without transferring data.

AudriusButkevicius:
Answer to both questions is yes. You can also pause syncthing by pausing all devices, not only folders. Or start witg -paused command line.

Folder being paused will still make syncthing connect and exchange pings etc etc, but it will be minimal cost.

Nutomic:
Sounds perfect!

AudriusButkevicius:
Added backwards compatibility. I guess it will be easy just to revert that once we decide to drop 0.15.

cmd/syncthing/main.go
```diff
@@ -46,6 +46,8 @@ import (
 	"github.com/syncthing/syncthing/lib/upgrade"
 
 	"github.com/thejerf/suture"
+
+	_ "net/http/pprof"
```
calmh: ?

calmh:
I only glanced at the diff briefly now on my phone, so take this with a grain of salt. My previous attempts at folder pausing were shot down for the following reasons:

- Pausing a folder required a reconnect, which caused an index resend and consumed data. Not an issue any more. 

- A paused folder should not receive index updates, so it doesn't consume data for that reason. We can handle this easily by setting a paused flag in the cluster config and the other side won't start an index sender. Not sure if the diff does this?

- A paused folder should not receive requests. We can handle this in the same way, by remembering that flag and filtering out the device in the Available() check. Not sure if implemented?

AudriusButkevicius:
Paused folders simply stop existing. So nothing is received, nothing is sent (absent from CC), and nothing is scanned.

calmh:
That doesn't stop another device from requesting files from me for that folder though, if they see that I'm connected and they have an index for it since before?

AudriusButkevicius:
You can't request blocks from a paused folder, because it's not running, hence folderSharedWith returns false.

Regarding the introducer issue, good point. Let's decide which one are we merging first and I'll patch up the other

calmh:
I don't understand what you mean about folderSharedWith. That check is just based on the config. If the other device pauses the folder, I'm still sharing it with them so I have reason to request blocks from them if they are connected and the index says they have the blocks I need?

AudriusButkevicius:
folderSharedWith checks m.deviceFolders, which gets populated as folders are started. Paused folders are not started.

calmh:
But I'm not talking about the paused side, I'm talking about the other side that might want to request blocks from the paused side.

AudriusButkevicius:
Yes, requests will go out, but will not be answered,  i guess that should be addressed.

cmd/syncthing/main.go
AudriusButkevicius: the STPROFILER=":9090" doesn't work if this is not imported.

AudriusButkevicius:
Yet adding a paused boolean to protocol.Folder breaks backwards compatibility to some extent, as old clients will ignore the fact that it's paused and keep sending requests. Is that fine?

AudriusButkevicius:
perhaps we should add a separate ClusterConfig.PausedFolders and use that in the de-introducer branch to work around the folders being missing in the CM?

calmh:
> old clients will ignore the fact that it's paused and keep sending requests. Is that fine?

I think that's fine

AudriusButkevicius:
Actually it's fairly trivial, so I've "just added another map"™ to the soup.
Question, if I have A and B, and B connects paused, yet B has previously advertised to A that it has some files that A doesn't? Will A go into berserk mode trying to download from nowhere?

calmh:
As long as the Available() call doesn't return the device I think it should be fine, it's similar to B not being connected at all

cmd/syncthing/main.go
calmh: Right, I had mentally deprecated that, woops

lib/config/optionsconfiguration.go
```diff
@@ -41,6 +41,7 @@ type OptionsConfiguration struct {
 	OverwriteRemoteDevNames bool     `xml:"overwriteRemoteDeviceNamesOnConnect" json:"overwriteRemoteDeviceNamesOnConnect" default:"false"`
 	TempIndexMinBlocks      int      `xml:"tempIndexMinBlocks" json:"tempIndexMinBlocks" default:"10"`
 	UnackedNotificationIDs  []string `xml:"unackedNotificationID" json:"unackedNotificationIDs"`
+	UnpauseOnStart          bool     `xml:"unpauseOnStart" json:"unpauseOnStart"`
```
calmh: What's the intention for this config? It's essentially a bool you can set to have other bools changed in the config on startup?

calmh:
Apart from having confusingly many ways to alter the pause state this looks good to me, I'd like to merge it before 0.14.5 on tuesday, will try to do some testing on my own before then. :)

lib/config/optionsconfiguration.go
AudriusButkevicius: If this is set and you don't start with -(un)paused, it will unpause everything, just as it does in the current release. This just gives feature parity

calmh:
I get the point about feature parity, but I think the unpauseonrestart option is confusion and unnecessary and vote we don't include it. 

@canton7 is it an issue on your side that Syncthing can start up with things disabled?

canton7:
I'd have to check

AudriusButkevicius:
Well I feel that different people might want different things. What's the big deal of keeping the option?

calmh:
It just feels weird to me to have a value in the config that makes sure other values in the config are changed to something else on startup. It's too many ways to control the same thing, making it complicated when it should be simple and introducing paradoxes like `UnpauseOnStartup=true`, `syncthing -paused` -- what happens? I would much rather we just decided on a behavior and stuck to it as I don't think anyone has asked for this yet.

AudriusButkevicius:
Well we had people asking for the other behaviour (one of the tickets this is closing). I bet my socks we'll have people from the other camp who are currently satisfied moaning very soon.

calmh:
Probably. Nevertheless this seems like much to simple a feature to warrant the complexity of two command line flags, one odd global config setting that is different from any other setting we have, plus the per item configs.

I suggest keeping `-paused` (sets everything to paused at startup), adding `-unpaused` (unpauses everything at start, retains the old behavior if required) and not having the global config.

AudriusButkevicius:
Removed the option

calmh:

or just, like, test it to start with like it's your actual job to do

calmh:
okay... I'm getting a bit annoyed with Jenkins now 😠

calmh:
Something is bad with unpausing folders. Using h1 and h2 configs,

- Start both
- Add some files to s1/, they sync to s2/
- Pause default on instance one (i.e. s1)
- Add some files to s1
- Unpause default on instance one

New files are scanned and visible on instance 1, but not sent to instance 2.

screen shot 2016-09-15 at 12 53 14

screen shot 2016-09-15 at 12 53 17

AudriusButkevicius:
Addressed the issue.

cmd/syncthing/gui.go
```diff
@@ -270,10 +269,12 @@ func (s *apiService) Serve() {
 	postRestMux.HandleFunc("/rest/system/restart", s.postSystemRestart)        // -
 	postRestMux.HandleFunc("/rest/system/shutdown", s.postSystemShutdown)      // -
 	postRestMux.HandleFunc("/rest/system/upgrade", s.postSystemUpgrade)        // -
-	postRestMux.HandleFunc("/rest/system/pause", s.postSystemPause)            // device
-	postRestMux.HandleFunc("/rest/system/resume", s.postSystemResume)          // device
 	postRestMux.HandleFunc("/rest/system/debug", s.postSystemDebug)            // [enable] [disable]
 
+	// Deprecated
+	postRestMux.HandleFunc("/rest/system/pause", s.deprecatedMakePauseHandler(true))   // device
```
calmh: I don't think deprecating these things is moving in the right direction, to be honest. The initial GUI/API thing was to get, edit and post a complete config on every change. That's easy and it works. However it's also quite low level. From an API perspective it would make more sense to me to move in a direction where there are actions like these pause and unpause calls, similar to PUT/POST/DELETE devices and folders and toggling config options etc. So instead of removing (or deprecating) these, why not accept their existence and slowly move towards a more fine grained API? That they make persistent config changes internally is perfectly fine of course.

lib/events/events.go
```diff
@@ -45,6 +43,11 @@ const (
 	ListenAddressesChanged
 	LoginAttempt
 
+	// Deprecated events
+
+	DeprecatedDevicePaused
```
calmh: Likewise I see no reason in deprecating these events. Having more fine grained events is more useful to the consumer than just "the config changed. figure it out for yourself! ppppfft!" :)

cmd/syncthing/gui.go
AudriusButkevicius: Well because they are values that live in the config (oppose to being some runtime values), to me what we have now makes perfect sense.
Also, I am perfectly fine to have one and one way only to set these, oppose to two, one via config (as that's where they need to live sadly, in order to persist), and one via pause unpause calls.

lib/events/events.go
AudriusButkevicius: See above comment.

cmd/syncthing/gui.go
AudriusButkevicius: Sure, but our API for most of the things is posting a config as a whole, hence putting a sugar coating on a pretty shit api still makes it a shit api.

I guess we can add these, hide the option from the config, but then we'd have to query each folders pause state upon refresh, and for a guy with 209 folders that will probably be a bit painful.

Having a monolithic one endpoint API makes things really simple however, 0 extra calls as the config is already there.

AudriusButkevicius: So just to be clear, I am fine if we want to add new calls and new events, I am just worried about the 209 rest calls to get current folder state for some people.

scienmind:
A small request for the UI design while it's still WIP:
implement "Pause All Folders" and "Resume All Folders" buttons

Background:
On some devices (Laptops) I have a lot of stuff held in sync (>1 TB) in few dozens of repositories (folders). 
Often I'd like to save up on battery while still syncing one or two crucial folders. What i'm doing so far to achieve that is have 2 ST instances, one with few crucial folders, and another one with the rest (that I'd shut down if needed). Being able to just pause "less critical" folders will eliminate the need to mess  around with two instances. 
BUT i'd need for example to click "pause" 30 times to leave 2 folders active, and 30 clicks more to reactivate everything. Simply "Pause all" and manually enable a couple, and the "Unpause all" will make it much easier for daily use at the cost of a small button add...

AudriusButkevicius:
What's the score here? Want to drive this home.

AudriusButkevicius:
@st-jenkins test this please

cmd/syncthing/gui.go
calmh: Given that we generate events when things change, there is no need to refresh all folders.

AudriusButkevicius: What about when you refresh the ui?

calmh:
I don't see a need to hide anything from the config. I think we should do the following:

- Keep stuff in the config, as you do.
- Generate "config changed" event when these things change, as you do.
- Update the existing pause/resume methods to do the needful.
- Update the code for the paused/resumed events to still generate those events like they are today.
- Don't mark anything deprecated.

In the long run, having people depend on `POST /rest/system/pause/whatever` instead of our exact config layout should be a win for both parties. As such, I'd suggest adding endpoints for folder pausing similar to the existing ones for device pausing, and doing similar things when moving forward, in an attempt to evolve the API away from always posting the config.

When we do a .15 or a .16 we might consider whether we want to make a bigger change to the API or not.

AudriusButkevicius:
So what you seem to be proposing is to add apis and then not use it ourselves, which I find bizzare. 
We don't do 200 gets to get pause state, we peak into the config.
How do we do 'pause all'? Set it in the config too? Or do 200 posts?

canton7:
> So what you seem to be proposing is to add apis and then not use it ourselves, which I find bizzare.

Being nice to other consumers of your API is much appreciated! :smile:

AudriusButkevicius:
I want to be nice, as I agree "one call to rule them all" is bad, yet I am struggling to imagine an alternative way which doesn't result in 200 round trips in one case or another.

canton7:
Bearing in mind that the 95th percentile of devices in the cluster is 10, I think 200 is probably an edge case...

Let's not forget that there are other endpoints which are per-folder as well, and the 95th and 100th percentiles for number of folders and number of devices are very similar. If you think that the 200-device case needs special support to avoid multiple requests, we'll also need to add special support for the 200-folder case. In that case, let's add something across *all* affected endpoints which e.g. allows someone to specify a list of devices / folders.

AudriusButkevicius:
Fine.

calmh:
> So what you seem to be proposing is to add apis and then not use it ourselves

No, I think we should use it ourselves as well. We already do, for the existing pause/resume buttons.

AudriusButkevicius:
Yeah but we don't use for getting the paused'ness

calmh:
No, we get that from /rest/system/connections, which does include all the device connection status in one call. But for folders, for example, we already call /rest/db/status for every folder on startup, including pausedness there for each folder seems reasonable maybe, or in another place. We poll /rest/stats/folder for every folder as well. And /stats/device and so on. So I'm not sure I'm following what the concern is here.

AudriusButkevicius:
The concern here is that techically we should do GET id/pause to get the pausedness and make it "restful" and put or post to the same url to adjust the pausedness. Oppose to adjust it via X, lookup state via Y.

calmh:
Right, OK, that makes sense. I'm not certain that we need to completely fanatical in splitting up every little thing into it's own endpoint. I guess one middle ground could be a GET/POST on a `/rest/system/folder` endpoint, with a JSON payload that is same or similar in both directions.

I.e, GET /rest/system/folder/abcd-1234 returns:

```
{
    "id": "abcd-1234",
    "label": "My folder",
    ...
    "paused": false
}
```

and you can either POST a complete new folder config (replaces the existing one, similar to what we do today with the full config), DELETE it to remove the folder or PATCH with a document containing only attributes to change. The latter may be annoying for us to handle on the server side, but would mean the pause operation on the client side becomes

```
PATCH /rest/system/folder/abcd-1234
{
    "paused": true
}
```

This is possibly a bit out of scope for this specific issue though, and I'm getting tired of discussing so you'll soon have won by wearing me down to a point where I don't really care what API looks like any more 😑

canton7:
>  so you'll soon have won by wearing me down to a point where I don't really care what API looks like any more

No! Think of the other API consumers!

AudriusButkevicius:
Rebased on master, added events for folder pauses, yet no HTTP API for pausing unpausing, so you need to throw the whole config around for that. Reason for that is that the mux we use now does not support extracting arguments from the path which would be the right way to do it. We'll leave that for the future.

lib/config/deviceconfiguration.go
```diff
@@ -17,6 +17,7 @@ type DeviceConfiguration struct {
 	Introducer               bool                 `xml:"introducer,attr" json:"introducer"`
 	SkipIntroductionRemovals bool                 `xml:"skipIntroductionRemovals,attr" json:"skipIntroductionRemovals"`
 	IntroducedBy             protocol.DeviceID    `xml:"introducedBy,attr" json:"introducedBy"`
+	Paused                   bool                 `xml:"paused" json:"pause"`
```
calmh: paused/pause?

calmh:
Otherwise LGTM, let me take it for a spin as well.

lib/model/model.go
```diff
@@ -365,13 +372,16 @@ func (m *Model) RestartFolder(cfg config.FolderConfiguration) {
 	m.pmut.Lock()
 
 	m.tearDownFolderLocked(cfg.ID)
-	m.addFolderLocked(cfg)
-	folderType := m.startFolderLocked(cfg.ID)
+	if !cfg.Paused {
+		m.addFolderLocked(cfg)
+		folderType := m.startFolderLocked(cfg.ID)
+		l.Infoln("Restarted folder", cfg.ID, fmt.Sprintf("(%s)", folderType))
```
calmh: These should use the `cfg.Description()`

lib/model/model.go
```diff
@@ -2384,6 +2393,24 @@ func (m *Model) CommitConfiguration(from, to config.Configuration) bool {
...
+	fromDevices := mapDeviceConfigs(from.Devices)
+	toDevices := mapDeviceConfigs(to.Devices)
+	for deviceID, toCfg := range toDevices {
+		fromCfg, ok := fromDevices[deviceID]
+		if !ok || fromCfg.Paused == toCfg.Paused {
+			continue
+		}
+
+		if toCfg.Paused {
+			l.Infoln("Pausing", deviceID)
```
calmh: Note for the future, we should have something like a `deviceCfg.Description()` as well with the device ID (perhaps in short form) plus device name. Not in this PR though.

calmh:
Appears to work as designed while clicking around 👍

AudriusButkevicius:
Addressed, merge if happy.

calmh:
@st-review merge the shit out of this

all: Add folder pause, make pauses permanent (fixes #3407, fixes #215, fixes #3001)

st-review:
:ok_hand: Merged as bab7c8ebbff24ab50326ead5693cfcabcc87ca0a. Thanks, @AudriusButkevicius!

calmh:
Good stuff

tobiastom:
Thank you everyone for your work on this!

PR #3523, opened 2016-08-18
cmd/syncthing, gui, lib/config: Added 'Parent Folder' option (fixes #…

### LINKED ISSUE #2157: Default path for new folders
It would be great to choose a Default folder, where syncthing can create folders for new shares. 

e.g under Windows "%USERPROFILE%\Syncthing" or Linux "$HOME/Syncthing"

If a Device share now a new Folder with me, i can add the Folder without add a path and syncthing create the folder in this default folder. 

Maybe with a trusted (e.g. for devices under my control) flag for devices, that new shares automatically added.

I have some devices now and add a new folder to all devices manual is annoying. 

This feature exist in Bittorrent sync and is very handy.

mdevey:
### Purpose

Set the common parent path of most shares to reduce typing when adding a new share.  'Folder Path' is prefilled when '+ Add Folder' is pressed.

I personally want STPARENTFOLDER for a docker build.   I have no use for -parent-folder but was attempting to mirror existing code in main.go for completeness.

The name Parent Folder was chosen after I saw possible confusion with the Default Folder.
### Testing

Not alot, Mac only.  The config/gui changes are trivial.  Not so confident of any accidental side effects in main.go.
### Screenshots

image
### Documentation

If you accept the need for STPARENTFOLDER and '-parent-folder' the commandline operations page would need to duplicate the descriptions from main.go. 
### Newbie

Happy to rework, let me have it!

cmd/syncthing, gui, lib/config: Added 'Parent Folder' option (fixes #2157)

Ease '+ Add Folder' with a common parent folder.
The 'Folder Path' input is prefilled with this option.
Useful for users with most shares in one folder. eg /mnt/data/

Added command line options also:
  Environment variable STPARENTFOLDER
  Runtime switch -parent-folder


gui/default/syncthing/core/syncthingController.js
```diff
@@ -1296,6 +1296,7 @@ angular.module('syncthing.core')
 
         $scope.addFolder = function () {
             $scope.currentFolder = {
+                path: $scope.config.options.parentFolder,
```
calmh: Just at first glance I think this might be a recipe for disaster as it will pass validation if the user just accepts the default. Do that more than once and you have multiple folders all syncing directly into to the `parentFolder`. If we're to suggest anything here, perhaps `$parentFolder/$folderID` or something like that.

calmh:
On a nomenclature note I think we want to try to keep "folder" referring to our own folder concept (the one that has an ID and name and so on) and use "path" or "directory" to refer to the thing on disk. Otherwise it's confusing for a folder to have a parent folder etc.

calmh:
@st-jenkins add to whitelist

gui/default/syncthing/core/syncthingController.js
mdevey: Saw you post about nomenclature folder vs directory, thank you.
s/parentFolder/parentDirectory/

I tend to use $parentDirectory/$folderLabel for incoming shares.  With $folderID I would have no idea when using a file manager.  But for new shares... I guess a small bit of js to duplicate the 'Folder Label' into 'Folder Path' as it is being typed for the first time.  A bit more involved but a much better user experience.

I'll get back to you :)

AudriusButkevicius:
I wouldn't call it parent, as it's not clear what you mean, I'd just call it defaultPath, ir defaultLocation.

AudriusButkevicius:
Also the "incoming" folders have a label which I think should be prepopulated, so the code should be the same.

mdevey:
Ok defaultPath.  

Should I close this now, or leave it around for a day?

AudriusButkevicius:
Well either do a forced push to update the current pull request, or alternatively you can open a new one.

AudriusButkevicius:
Or make additional commits on top addressing the comments.

calmh:
@mdevey ping? :)

mdevey:
Sorry, will update this weekend or close it. Going on holidays next week.

mdevey:
Lets see if I'm heading the right direction?

Incoming folders) 
path autofilled from either:
defaultPath + folderLabel or 
defaultPath + folderID

New Folders)
So ... I stopped the auto Folder ID generation.  (it's a button instead)
This 1) Stops people hitting ok immediately and getting multiple folders with same path 2) allows us to pull the basename of the folderPath to use as the folderID.

screen shot 2016-09-05 at 12 19 34 am

Simple use is what I'm aiming for: Select an existing directory and you're done.  There is no need when adding a new (existing on filesystem) folder to do any typing in the Label or ID fields.

calmh:
> So ... I stopped the auto Folder ID generation. (it's a button instead)
This 1) Stops people hitting ok immediately and getting multiple folders with same path 2) allows us to pull the basename of the folderPath to use as the folderID.

Couldn't this equally be solved by auto populating the folder ID, and updating the folder path to be `defaultPath + folderLabel` whenever the folder label changes? It then gets a default that the user can just accept, or they can change it. There is an angular attribute `$dirty` or something per field that indicates whether the user has touched it, which could be used to update the path from the label as above, unless the user has changed the path in which case we don't touch it any more.

wweich:
I also think that the ID should stay auto generated. It was introduced to prevent duplicate folder IDs (like pictures) which prevents those folders to be shared with someone else having the same folder ID for something/someone else,
Only the label should be taken from the path, not the ID.

mdevey:
Fair enough, I can adjust when I return from holidays.

I think the messages/logging need improving to include Labels.  When trouble shooting random id's led me to cycle through folders looking for an id.  I've recently found my life is simplified when I don't use a random id.

Might it be possible to hide the existence of id's from the user until there is a conflict?

calmh:
Certainly error messages and so on should show labels in addition to ids at the very least.

AudriusButkevicius:
I assume this is dead, should we close it?

mdevey:
I'll open another if I progress

Issue #1284, opened 2015-01-22
smarter block splitting like rsync


rwcarlsen:
A rolling checksum algorithm could be used to compute split points instead of
using a fixed block size for files.  The blocks would still be hashed and
identified with sha256 - the rolling checksum would be used only for computing
split points.  The gain from this change would be smarter block tranmission
between nodes.  When a file is modified in a way that changes its length, this
algorithm results in nearly all blocks before and after the changed area being
identical - no need to re-transmit.  This is possible because the rolling
checksum operates over a relatively small byte window (~10's of bytes).  Split
points are identified as locations where the rolling sum has so many
consecutive trailing zero bits. All regions in a file where the window doesn't
include modifications, the rolling checksum values are the same - resulting
in split points being identical. This will result in variable sized blocks.

* Add rolling checksum splitting to `internal/scanner` - probably something
  like what bup does (http://github.com/bup/bup).  It might also be a good
  idea to enforce hard minimum and maximum block size limits (e.g. 0.1x
  blocksize and 3x blocksize).

* reduce (avg) protocol.BlockSize to ~65k.  This works better with
  cheaper rolling checksum algorithms.

* Remove fixed-length block code assumptions.  This primarily involves (starts
  with) code in `internal/model/puller.go`'s `copierRoutine` method and it
  spiders out a bit from there.  The most elegant way I have come up with so
  far involves changing the value stored for block keys in leveldb to be the
  byte offset of the block instead of an index.  This is necessary since the
  index can no longer be used to compute the offset for verifying blocks -
  i.e.  `fd.ReadAt(buf, protocol.BlockSize*int64(index)` wouldn't work
  anymore.

With a naive implementation, a version of syncthing doing this would work rather
inefficiently with older versions of syncthing.  Consider a node A running with
this new block splitting communicating with an older node B:

1. A sees a new local file X created and sends its block list to B.

2. B then pull over missing blocks - just like we want.

3. B sees a local modification to file X. It recomputes the block list and
  sends it to A.

4. All the received blocks look new to A because of different split points -
  even though the file might have just had a small change (even an append).  It pulls over all the blocks.

5. A sees a local modification to file X.  It recomputes the block list and sends it to B.

6. All the received blocks look new to B because of different split points -
  even though the file might have just had a small change (even an append).  It pulls over all the blocks.

7. Repeat from 3 ad infinitum.

So new versions are still compatible, but their communication is not very
efficient. One possible solution would be to have nodes change their splitting
[...]

Issue #107, opened 2014-03-31
rsync/delta transfer


pgega:
Documentation (feature list) does not make it clear if syncthing supports diff/delta (rsync like) file transfer.

Is diff transfer supported?

MarkusTeufelberger:
Could this issue be reopened or clarified? As Far as I understand the current model, it works efficiently only in some specific cases:

* Data is appended at the end of a file
* An exact multiple of 128 kiB is added anywhere in the file (start/middle)
* Data is being replaced by different data with exactly the same size anywhere in the file
* An exact multiple of 128 kiB is deleted from anywhere in the file

In all these cases, this would cause only very few blocks to be transmitted.

Where this method fails, is in any other case.

A potential worst case would be e.g. a video file format that stores metadata right at the beginning of the file. If your movie from "last christmas" should now be called "Christmas video 2014" or you fix a typo in the subtitles, Syncthing will have to transmit the whole file, even if only a few bytes were added. Rsync prevents this by using the file content itself to decide where to place block boundaries, not only size.

calmh:
That's true, but the creators of said video format should be tried for crimes against humanity. Not only would it be an inefficient transfer, it would be an inefficient file write and and an inefficient backup. My point being, I don't think it's a common case - compare for example (my closest personal experience) the Adobe DNG format for raw photos, which includes metadata. Said metadata is stored in XML at the start of the file. But it's padded with a lot of nuls, just for this purpose - normally there's a kilobyte or two of them after the end of the XML. Writing a larger or smaller XML into this area is just a small rewrite and does not touch the rest of the file.

There are other reasons not to do the rsync thing as well. Primarily, it requires syncing a given file between exactly two parties who both read and hash the file pretty much in lockstep. There's no simple clustering here because how we chop up the file depends on it's contents, and if I have a set of block hashes for different block offsets than you, there's no reconciling them.

MarkusTeufelberger:
This was just the first example that came to my mind, I'm sure there are better ones out there. My point is that the current solution does not handle the general case and I'm not 100% sure if the current cases are really the most common ones.

Any change can max. impact 2 existing block hashes (+ the amount of blocks within that change). Rsync splits files based on a fixed size sliding window, so introducing a new split somewhere in the beginning does not have an impact on later splits. See http://manpages.ubuntu.com/manpages/maverick/man1/bup-split.1.html for a similar implementation.

csindle:
My 2¢ scenario where this also hurts us:
[...]

PR #3527, opened 2016-08-20
lib/model, lib/scanner: Efficient inserts/deletes in the middle of the file

AudriusButkevicius:
### Purpose

Review with &w=1
Uses a weak rolling hash to identify shifted blocks.
Potentially a DoS surface, by forcing us to hash the file `size_of_file` times with SHA256 since the weak hash can collide fairly easily?

Yet there are other ways to do a similar thing I guess, such as providing blocks with the incorrect hash to start with, but that is somewhat limited by the network speed.

Also probably needs some tests, as I haven't checked if it actually works.


AudriusButkevicius:
LUL:

```
run compiler with -v for register allocation sites
lib/weakhash/weakhash.go:77: internal compiler error: out of fixed registers
```

calmh:
So it looks sort of reasonable on the face of it. Why, though? I mean what's the compelling use case here? And what are the performance tradeoffs, i guess it results in up to blocksize number of rolling hash rounds and then one or more extra sha256 rounds?

Apart from that, what is the `weakhash` algo? There could be a comment about what it is and where it comes from.

calmh:
I've filed an extra bug on the compiler issue, in the meantime this compiles:

```
func (d *digest) Write(data []byte) (int, error) {
	for _, c := range data {
		d.a -= uint16(d.buf[d.j])
		d.a += uint16(c)
		d.b -= uint16(d.size) * uint16(d.buf[d.j])
		d.b += d.a
		d.buf[d.j] = c
		d.j = (d.j + 1) % d.size
	}
	return len(data), nil
}
```

AudriusButkevicius:
The use case is somewhat obvious, better performance for files that get content appended in the middle.
It results in `length_of_file` rounds of rolling checksum (yet the round itself is 4-5 operations per byte as it's a sliding window hash), and yes, potentially a few more SHA256 rounds.

So the algorithm is http://tutorials.jenkov.com/rsync/checksums.html (or simillar to this http://blog.liw.fi/posts/rsync-in-python/#comment-fee8d5e07794fdba3fe2d76aa2706a13)
which I believe is what rsync used before they switched to a bastardized version of adler32.
We could use adler32, yet we'd have to make a copy of the implementation in the stdlib as it does not support a rolling window, so a lot of meaningless recomputation is involved.

AudriusButkevicius:
If you are fine with the general idea/approach, I'll add tests add comments, and potentially add a flag to disable this.

calmh:
> The use case is somewhat obvious, better performance for files that get content appended in the middle.

Sure. But when is that required? I mean under what real world situations is this a win. I've heard of Photoshop files, which may be large-ish (I guess on the scale of ~100 MB?) and are binary and rewritten on each save so can benefit from this. But are there any other real life cases?

> It results in length_of_file rounds of rolling checksum (yet the round itself is 4-5 operations per byte as it's a sliding window hash), and yes, potentially a few more SHA256 rounds.

I don't see why we'd need to do more than `blockSize` rounds as doing one with offset `blockSize+1` is equivalent to doing one with offset `1`, just the resulting list of block hashes being off by one?

But my real concern is how much does this cost in the general case where it is not needed, that is how much slower will it be to sync a change of some piece of data in the middle of some file.

In the end I'm trying to judge if this makes sense based on the advantages for some use case I'm not sure of outweighing increasing the cost of every sync operation.

AudriusButkevicius:
Yes we do blockSize rounds per block, but in total we do length_of_file rounds, as we produce a hash for the first block, then add the next byte (which removes the last byte as the window slides) produce a hash again, etc, etc until we reach the end of the file, so in total it will be length_of_file rounds.

Sadly, I don't have a usecase. I mean rsync does it, and that's like a defacto best tool around, hence why I am trying to make syncthing just as good . I know large word files, large PDFs, large SVGs would benefit from this too.

In terms of cost, I feel that computing the rolling hash as we scan will have pretty much no noticeable impact, compared to SHA256.

When downloading, we'll have to compute the rollmap for the file, which again, I suspect will have no noticeable impact when compared having to access the database, ask for stuff over the network and verify it with SHA256. Yes, there might be a few extra rounds of SHA256 involved in case the weak hash collides, but I suspect it's improbable for the hash to collide within the same file. Otherwise, if we don't collide, we are not going over the network and already benefiting from savings.


Btw, if we get variable block sizes, and micro delta indexes syncthing will be better than rsync in terms of features and network bandwidth required by the protocol.

AudriusButkevicius:
I mean, perhaps we should ask the community if they would find this useful

canton7:
Given that you've done the work, I feel it would be a shame to lose it, if the cost is negligible and it's nice and encapsulated...

canton7:
... Maybe only enable on files over a certain size if the cost does become an issue?

calmh:
Right, I misunderstood how it worked and interpreted one "round" as one pass over the file, when in fact it's a total of one pass over the file and that's it?

Given that, I think this is viable. 

However I think we should only trigger this based on some half way intelligent heuristic, for example by detecting that allnor most blocks have changed after a given point in a file. If we're just going append data, replace the last block, or replace a block in the middle there is no reason to compute a roll hash. 

And it would be good to keep a few metrics on this, some sort of hit ratio, that we could add to the report and make future decisions based on.

AudriusButkevicius:
> Right, I misunderstood how it worked and interpreted one "round" as one pass over the file, when in fact it's a total of one pass over the file and that's it?

Yes, we compute `number of bytes in file` hashes, each hash covering protocol.BlockSize, and each hash represents byte index in the file.

I just realized that it won't work with big files, as it's 4 bytes per hash, and if we have a 8GB file, that's 32GB worth of hashes.

I'll change the API so that it takes a list of hashes we are interested in, and returns offsets for each hash if it finds it during the roll.

Yes, perhaps enabling this for anything that will potentially have to find xMB from somewhere else would make sense.
That somewhere else could be the blockmap though, but perhaps checking if we can get anything by rolling is better than accessing the block map for every hash?

AudriusButkevicius:
Anyways, lets settle if this is something we want to go for, and if it is, I'll drive it home.

calmh:
I don't *mind* this, but I would like to know about at least *one* real world case where it will make a positive difference. As in an actual user saying "this will help a lot in my situation because *x*" where *x* is something more concrete than "I think the algorithm would be cool" :)

After that it's "just" technical details and I'm sure we can solve them.

AudriusButkevicius:
Well I can't give you one sadly. Perhaps ask on the forum, yet I found #1284 #107 that talk about this and "better delta sync", "noob question - incremental sync" which talk about this and the interest of having this .

calmh:
The gzipped-sql-dumps one sounds vaguely plausible, if that does result in the sort of data shifts we are talking about here. Could be interesting to check.

uok:
Do PST-files (MS Outlook) fall into this category?

Ferroin:
PST files are not likely in this category, they work like a database file, expanding to the desired size by allocating space at the end, with rewrites happening in the middle without shifting allocations in the middle of the file.  In essence, a PST file (or a SQLite database) is like a small filesystem which grows in size to fit internal allocations, without moving anything already allocated unless it's told to do so.

The biggest likely benefactors are the few odd applications that use fallocate(2) with FALLOC_FL_INSERT_RANGE or FALLOC_FL_COLLAPSE_RANGE (I know such applications exist, but I don't know of any specific examples, and they are essentially non-existent on Windows because it doesn't have a syscall that works that way).

In theory though, anything that shifts data around inside a file could benefit.  The most obvious case would be plain text files with extra data inserted in the middle, but it should also benefit some compressed formats such as PSD, XCF, DOCX, ODT, and similar, as well as some archive formats.

calmh:
Indeed. But all of those file formats are usually small enough that this doesn't matter. If the compressed gigabytes-sized SQL dump *does* benefit from this, that would be fair.

Ferroin:
The thing is, once you start to see files bigger than a few megabytes, this type of thing does improve the network bandwidth utilization measurably.  Big files are of course going to benefit the most, but even smaller ones will benefit.  Using rsync as an example (because it has the ability to behave like this), take a 256MB file, copy it, and then add one byte at the beginning of the file.  Re-syncing this file will require copying the whole 256MB using a traditional copy, which will mean close to 270MB of data over the network using scp, while using a method like this proposal with rsync will require sending maybe a few KB of data over the network o achieve the same thing.

AudriusButkevicius:
The point is that we are unaware of anything that prepends or inserts to the file while also being of reasonable size.

mdevey:
Use case: pack files.  Media heavy games, preprocess 1000s of assets into the memory format used in ram.  So instead of 1000 individual file loads and processing, one file is dumped into ram.

Use case: mng (sequence of pngs).  Artists typically add a new action (fade in, new loop) to the same file and update a text file to reference the new frames with a new action verb.  Alot easier for programmers to deal with than a whole new asset.

Rsync is the king for updating iterated game assets.  I can't stress that observation enough. Aspera etc are woeful.

Ferroin:
There are all kinds of other things too though.  Zip archives for example will benefit from this if new files are added or old ones are removed, or even if the archive is rewritten.  ISO images would fall into the same category.

You can't think of this like a filesystem developer would.  A file that gets replaced by a new version with the same name is seen as the same file with different data by Syncthing, not a new separate file (and that is exactly how it should be in userspace).  This means that things that rewrite files using replace by rename method still get a benefit form this because Syncthing still sees it as the same file.  There are very few things that prepend to or insert data into the middle of a file in-place, but there are quite a few which rewrite the whole file but still cause existing data to shift around inside it.

Taking the example of a zip archive:  Create a zip archive with a bunch of files in it.  Now create a second with the same files, except remove one or add one more to it.  Now do a hex-dump of the files and do a side-by-side comparison (for example with vimdiff or winmerge).  Most of the data will be the same between the files, just with parts of it shifted.  The same applies to ISO 9660 filesystem images, and, with slightly less frequency and in more specific conditions, to most other archive formats.

fti7:
What about VM Disk Images (e.g. VMware .vmdk) or TrueCrypt Containers? ;-)

AudriusButkevicius:
These I suspect change in place, or are append only, so are already handled. We are only interested in prepend or insert cases.

AudriusButkevicius:
So sadly the gzipped SQL dumps argument is invalid:

```
bash@smash:~$ diff -y <(cat x.sql | gzip -c | xxd) <(cat y.sql | gzip -c | xxd) |
...
0000410: b4f2 8ff7 d12d 23db f4d4 3585 3ee5 af6b  .....-#...5   0000410: b4f2 8ff7 d12d 23db f4d4 3585 3ee5 af6b  .....-#...5
0000420: 9d9f f4da b5de 2bbe 5b2d 1e21 f1eb a509  ......+.[-.   0000420: 9d9f f4da b5de 2bbe 5b2d 1e21 f1eb a509  ......+.[-.
0000430: 97e3 9e92 9e4c 01f3 925c f3e8 b858 b08b  .....L...\.   0000430: 97e3 9e92 9e4c 01f3 925c f3e8 b858 b08b  .....L...\.
0000440: 56e3 892e e5a2 557f b4cb a19b f88a 8efc  V.....U.... | 0000440: 56e3 892e e5a2 55a7 2e5c ba92 6ffe ec1d  V.....U..\.
0000450: 14d9 5974 fb73 b8f0 9cf2 cced bb7d f9e5  ..Yt.s..... | 0000450: 77e8 cebe a277 3fc5 8016 ddfe 1c7e 3de7  w....w?....
0000460: 2e1c 23e9 7f7a 824d 3f9e 8fe6 593a 8e12  ..#..z.M?.. | 0000460: 4173 a36f 5f7e b95f c791 3c60 824d e79e  As.o_~._..<
0000470: 1d8c c947 1723 76f4 ea09 444e ff40 62b4  ...G.#v...D | 0000470: 8fe6 593a 8e12 1da1 c971 1723 76f4 ea09  ..Y:.....q.
0000480: 3134 4bfc 03c9 d7c6 c0f6 9bcf 7c20 71a8  14K........ | 0000480: 8453 ff40 b6b4 3134 4bfc 0319 d9c6 c0f6  .S.@..14K..
0000490: 2f63 52ab 33f4 4641 fafb 6619 4461 8abf  /cR.3.FA..f | 0000490: 9bcf 7c20 11ab 2fa3 57ab 33f4 4641 fafb  ..| ../.W.3
...
```

Same goes for xz

Yet it works for .zip files:


```
bash@smash:~$ diff -y <(cat x.sql | zip -q | xxd) <(cat y.sql | zip -q | xxd) 
...
0000440: dbf4 d435 853e e5af 6b9d 9ff4 dab5 de2b  ...5.>..k..   0000440: dbf4 d435 853e e5af 6b9d 9ff4 dab5 de2b  ...5.>..k..
0000450: be5b 2d1e 21f1 eba5 0997 e39e 929e 4c01  .[-.!......   0000450: be5b 2d1e 21f1 eba5 0997 e39e 929e 4c01  .[-.!......
0000460: f392 5cf3 e8b8 58b0 8b56 e389 2ee5 a255  ..\...X..V.   0000460: f392 5cf3 e8b8 58b0 8b56 e389 2ee5 a255  ..\...X..V.
0000470: 7fb4 cba1 9bf8 8a8e fc14 d959 74fb 73b8  ........... | 0000470: a72e 5cba 926f feec 1d77 e8ce bea2 773f  ..\..o...w.
0000480: f09c f2cc edbb 7df9 e52e 1c23 e97f 7a82  ......}.... | 0000480: c580 16dd fe1c 7e3d e741 73a3 6f5f 7eb9  ......~=.As
0000490: 4d3f 9e8f e659 3a8e 121d 8cc9 4717 2376  M?...Y:.... | 0000490: 5fc7 913c 6082 4de7 9e8f e659 3a8e 121d  _..<`.M....
00004a0: f4ea 0944 4eff 4062 b431 344b fc03 c9d7  ...DN.@b.14 | 00004a0: a1c9 7117 2376 f4ea 0984 53ff 40b6 b431  ..q.#v....S
00004b0: c6c0 f69b cf7c 2071 a82f 6352 ab33 f446  .....| q./c | 00004b0: 344b fc03 19d9 c6c0 f69b cf7c 2011 ab2f  4K.........
...
```

AudriusButkevicius:
Perhaps I should invent a better way to measure this to be honest.

I'll come back soon.

AudriusButkevicius:
Ok, so my second test consists of computing a rolling sum for each of the compressed files, sorting the sums numerically, and then diffing the resulting file. Sadly none of the compression algorithms would seem to benefit from rolling sums, as compressed content after the change is fairly different, yet for the raw files it's a different story.

calmh:
Yeah that makes sense and matches my expectations. Perhaps the use case mentioned in the ticket was raw sql with the row values themselves compressed somehow (maybe just base64 encoded or something). That's back to being way out there in niche land though, as far as I'm concerned. Remember, rsync is still out there and will remain the superior solution for any number of use case, not just due to this difference. 

> pack files

For game developers syncing their source?

> mng 

Unsure, sounds like append?

> disk images

In place updates, no benefit. 

> zip archive updates

Append only unless you mean rewritten completely, in which case Audrius test above. For the case with a new zip file created with one file less than another I guess that can result in shifted blocks, but again, when does this happen out there with any real frequency?

> ISO images

Maybe. What's the use case for frequently changing ISO images with slightly differing contents?

I'm not trying to be an asshole here or saying that this can never be useful, and the code is sound, but if it's not clearly useful for at least a measurable percentage of use cases out there...

I don't the mind adding one off things for odd use cases when the "damage" is isolated. This however complicates and potentially slows down our absolutely most central operation so I'm applying a higher standard for whether it should go in.

AudriusButkevicius:
We could make it yet another option though

Ferroin:
Having the option would be nice.  This is something what will benefit some specific use cases, but will provide no net benefit to others.

As far as compressed files, they theoretically could benefit, but for most compression formats you need very specific situations for this to happen (the insertion in the uncompressed file must be a multiple of the compression block size, which can be pretty big depending on your parameters).  Archive files though should still benefit, ZIP and many other archive formats which natively support compression compress each file individually, so adding a file usually means that the rest of the content shifts (depending on where the file is in the archive).  Even tar archives work like this (albeit at a block level, not file level), except they need to be uncompressed for this to be seen (or meet the same requirements I listed above for plain compression formats).

calmh:

AudriusButkevicius:
So it's broken unfinished and potentially unmergable. I was throwing it out there for discussion. Whats the conclusion, do we want to have this with an option, or should we have it on by default given speed of the hash is 20gbps?

calmh:
OK so lets say we do this. I might want it optional, and I'll be very happy if it records some sort of hit ratio. I'd say the main blocker at this point is the memory requirement? That rollmap gets big quickly. Do it in segments perhaps?

AudriusButkevicius:
No we just look for a predefined set of sums as we roll, so not much of a memory requirement.

sjau:
I'd love that as well.

I have a 100GB veracrypt container with stuff in it. I think it does write in the middle when it's mounted and stuff is changed.

AudriusButkevicius:
It does write in the middle, but not an append in the middle, so current mechanism should be good enough.

nwf:
Just pitching in my $0.02.

I have seen this kind of weak rolling checksum (under the guise of "Rabin splitting", which cuts blocks when the rolling sum is N mod M, for fixed N and M, or some-such) be very useful when storing uncompressed tarballs and SQL and AFS-volume dumps in bup.  These are files of variable-length records with arbitrary traversal order (in particular, creates and appends in the filesystem do not land at the end of the dumps; they appear in the middle of the dump file) and the rolling checksum does a pretty reasonable job at recovering structure that a fixed-length block system misses.

Note that the lack of compression on the input is key; a compressor does a great job of also detecting structure (that is, after all, its job) and producing a stream that differs essentially everywhere after the first difference.  That is, unless the compressor is also doing Rabin splitting or something similar, it will not re-synchronize with its previous runs.  For this reason, bup splits and then compresses each chunk from the splitter in isolation.

That said, I am not sure that this is a compelling use case for syncthing.  At least, we generate our tarballs and AFS dumps and immediately shove them into our archival store (bup) and then replicate *that*, rather than synchronizing the raw dumps to a remote machine.  Anyone synchronizing a directory full of uncompressed tarballs?

Ferroin:
@nwf ZIP files also benefit, they work in a manner similar to tar, but they compress prior to archiving, not after (which leads the the rather comical situation that you can gzip a ZIP archive and often make it a bit smaller).  I'm not sure about any other archive formats, but most of them likely do the same thing, as most are designed for random access, unlike TAR which is designed for purely sequential access (which is why it is possible for tar to process compressed archives without needing a temporary file or insane amounts of memory).  I know some games use archives like that (and I think Steam does for distribution and it's local backup format), and there are other odd cases where it may benefit (Minecraft's Anvil format region files appear to update like this sometimes, but not always for example).  If anything, this is a feature that will likely help a small number of people, and should probably be optional and default to being off.

Tstassin:
I think this PR is relevant for media files (music, video, images), see  [here](https://forum.syncthing.net/t/no-delta-sync-while-updating-metadata-of-flac-files/8244).

In my case, activating this option, even if only available through advanced options, could mean 99% less bandwidth overhead

AudriusButkevicius:
So this should now be less heavy, and keeps stats (yet doesn't use them).
Are you fine with what we have and I should start writing tests, or do you want some exposure of the stats somewhere or some extra bits done?

calmh:
Probably fine is as. I do want the stats stored and / or exposed somewhere, but that probably needs to be a part of the usage reporting revamp. Some more tests and/or benchmarks would be awesome of course.

calmh:
Speed stripes:

```diff
commit 27f9c2b275afc99e4a0102a4eb79ffe660a6434c
Author: Jakob Borg <jakob@kastelo.net>
Date:   Sat Nov 5 20:12:10 2016 +0100

    wip: Make it 93% faster

diff --git a/lib/weakhash/benchmark_test.go b/lib/weakhash/benchmark_test.go
new file mode 100644
index 0000000..d9de813
--- /dev/null
+++ b/lib/weakhash/benchmark_test.go
@@ -0,0 +1,22 @@
+package weakhash
+
+import (
+	"os"
+	"testing"
+)
+
+const testFile = "../model/testdata/~syncthing~file.tmp"
+
+func BenchmarkFind1MFile(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		fd, err := os.Open(testFile)
+		if err != nil {
+			b.Fatal(err)
+		}
+		_, err = Find(fd, []uint32{0, 1, 2}, 128<<10)
+		if err != nil {
+			b.Fatal(err)
+		}
+		fd.Close()
+	}
+}
diff --git a/lib/weakhash/weakhash.go b/lib/weakhash/weakhash.go
index eeaa7f1..39b9560 100644
--- a/lib/weakhash/weakhash.go
+++ b/lib/weakhash/weakhash.go
@@ -7,6 +7,7 @@
 package weakhash
 
 import (
+	"bufio"
 	"hash"
 	"io"
 	"os"
@@ -30,6 +31,15 @@ func Find(r io.Reader, hashesToFind []uint32, size int) (map[uint32][]int64, err
 	if r == nil {
 		return nil, nil
 	}
+
+	// Check if the reader implements ReadByte() (i.e., is probably a
+	// buffered reader), otherwise wrap it in a bufio.Reader.
+	if _, ok := r.(interface {
+		ReadByte() (byte, error)
+	}); !ok {
+		r = bufio.NewReader(r)
+	}
+
 	hf := NewHash(size)
 
 	n, err := io.CopyN(hf, r, int64(size))
```

lib/model/rwfolder.go
```diff
@@ -1109,41 +1128,61 @@ func (f *rwFolder) copierRoutine(in <-chan copyBlocksState, pullChan chan<- pull
...
 				_, err = dstFd.WriteAt(buf, block.Offset)
 				if err != nil {
 					state.fail("dst write", err)
+
 				}
-				if file == state.file.Name {
-					state.copiedFromOrigin()
-				}
-				return true
+				state.copiedFromOriginShifted()
```
calmh: If the found offset is equal to the old block offset this was just a regular copy-from-origin and not a shifted one, so the weak hash doesn't get to claim a point. (This matters not at all except when we some day come to the point of looking at the stats to evaluate the performance of the algorithm.)

AudriusButkevicius:
Added some tests addressed comments and squashed a few bugs along the way.

calmh:
OK LGTM except I very much dislike adding another megabyte of binary crap to the repo just for testing. It should be enough with like a kilobyte of data?

AudriusButkevicius:
Sure.

AudriusButkevicius:
Done.

calmh:
This gets stuck on the simple syncing integration test, where `master` doesn't. Not sure what's wrong, could be something about the reporting or events that the test doesn't see it as having become in sync?

```
jb@unu:~/s/g/s/s/test $ go test -v -tags integration -run SyncCluster
=== RUN   TestSyncClusterWithoutVersioning
2016/11/08 14:10:48 Testing with numFiles=100, fileSizeExp=20, timeLimit=5m0s
2016/11/08 14:10:48 Cleaning...
2016/11/08 14:10:48 Generating files...
2016/11/08 14:10:48 Starting Syncthing...
2016/11/08 14:10:48 Starting instance 1...
2016/11/08 14:10:49 Starting instance 2...
2016/11/08 14:10:51 Starting instance 3...
2016/11/08 14:10:52 Iteration 1
2016/11/08 14:10:52 Forcing rescan...
2016/11/08 14:10:52 Syncing...
(stuck here)
```

AudriusButkevicius:
So even if I run on master, I get barely 10-20KB/s with these tests.
strace shows most of the time spinning around futexes.

AudriusButkevicius:
Can you double check that the integration tests pass fine for you on master?

calmh:
I seem to have accidentally commited some crap in the test configs form earlier testing. I fixed that on master now.

AudriusButkevicius:
Rebased on master, and passes fine.

calmh:
This still does not pass for me, does it for you?

```
jb@unu:~/s/g/s/s/test $ go test -v -tags integration -run SyncCluster
=== RUN   TestSyncClusterWithoutVersioning
2016/11/12 09:25:23 Testing with numFiles=100, fileSizeExp=20, timeLimit=5m0s
2016/11/12 09:25:23 Cleaning...
2016/11/12 09:25:25 Generating files...
2016/11/12 09:25:25 Starting Syncthing...
2016/11/12 09:25:25 Starting instance 1...
2016/11/12 09:25:27 Starting instance 2...
2016/11/12 09:25:28 Starting instance 3...
2016/11/12 09:25:29 Iteration 1
2016/11/12 09:25:29 Forcing rescan...
2016/11/12 09:25:29 Syncing...
/// eternal wait here
```

AudriusButkevicius:
```
Audrius@Audrius syncthing $ git show -1 -q
commit 30a7b8bfd76521b825fb8c882a6a2c13ed43b923
Author: Audrius Butkevicius <audrius.butkevicius@gmail.com>
Date:   Thu Nov 10 21:40:10 2016 +0000

    lint
Audrius@Audrius syncthing $ ./build.sh
github.com/syncthing/syncthing/lib/protocol
github.com/syncthing/syncthing/lib/weakhash
github.com/syncthing/syncthing/lib/discover
github.com/syncthing/syncthing/cmd/stdiscosrv
github.com/syncthing/syncthing/lib/scanner
github.com/syncthing/syncthing/lib/config
github.com/syncthing/syncthing/cmd/stfileinfo
github.com/syncthing/syncthing/lib/db
github.com/syncthing/syncthing/cmd/stdisco
github.com/syncthing/syncthing/cmd/stfinddevice
github.com/syncthing/syncthing/cmd/stindex
github.com/syncthing/syncthing/lib/relay/protocol
github.com/syncthing/syncthing/lib/nat
github.com/syncthing/syncthing/lib/relay/client
github.com/syncthing/syncthing/lib/pmp
github.com/syncthing/syncthing/lib/upnp
github.com/syncthing/syncthing/cmd/stvanity
github.com/syncthing/syncthing/cmd/strelaypoolsrv
github.com/syncthing/syncthing/cmd/strelaysrv/testutil
github.com/syncthing/syncthing/cmd/strelaysrv
github.com/syncthing/syncthing/lib/connections
github.com/syncthing/syncthing/lib/stats
github.com/syncthing/syncthing/lib/model
github.com/syncthing/syncthing/cmd/syncthing
Audrius@Audrius syncthing $ cd test
Audrius@Audrius test $ go test -v -tags integration -run TestSyncClusterWithoutVersioning
=== RUN   TestSyncClusterWithoutVersioning
2016/11/12 08:32:35 Testing with numFiles=100, fileSizeExp=20, timeLimit=5m0s
2016/11/12 08:32:35 Cleaning...
2016/11/12 08:32:37 Generating files...
2016/11/12 08:32:41 Starting Syncthing...
2016/11/12 08:32:41 Starting instance 1...
2016/11/12 08:32:42 Starting instance 2...
2016/11/12 08:32:44 Starting instance 3...
2016/11/12 08:32:45 Iteration 1
2016/11/12 08:32:45 Forcing rescan...
2016/11/12 08:32:47 Syncing...
2016/11/12 08:32:47 Checking...
2016/11/12 08:32:48 Altering...
2016/11/12 08:32:49 Iteration 2
2016/11/12 08:32:49 Forcing rescan...
```

calmh:
Hmm OK I'll investigate

AudriusButkevicius:
Bump. Not to be forgotten.

timofonic:
So is this going to be merged? :D

calmh:
This fails the basic sync test on my computer, which passes on master. Given that it apparently passes for Audrius I need to investigate. I haven't had time to do so.

AudriusButkevicius:
So I'd ask if you had time for this but I guess it's:

AudriusButkevicius:
@st-jenkins test this please

calmh:
Yeah so now that the other urgent things are maybe sorted for the moment, let me take another look.

calmh:
@st-jenkins test this please, i thought audrius fixed you

calmh:
@AudriusButkevicius you have a couple of conflicts to resolve at your leisure, but i'll try to test the branch at the state it was in the meantime

calmh:
OK I've narrowed it down to us losing events somewhere, which is probably not something you've caused here but is triggered by timing or whatever, looking into that.

calmh:
Now it works for me. Lets get master merged into this so it's clean and make some progress.

AudriusButkevicius:
@st-jenkins test this please

AudriusButkevicius:
@st-jenkins test this please

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 05828368200d519e3ddc18f22ebf05c4a6c5e836. Thanks, @AudriusButkevicius!

AudriusButkevicius:
not sure if you were fine to merge this, but given this is "approved" I assume you were.

calmh:
I was going to review it another time, but yeah this is fine. However, I still want stats on this. I don't think anyone will mind adding a usage reporting variable on the hit rate for this block finder, at the same time the feature is added. So some time during the next two weeks, before release, I would very much like something that reports the blocks stats from this thing. I can implement it myself as well if you don't have the bandwidth.

PR #3532, opened 2016-08-22
cmd/syncthing: Remove old temp index dbs on startup (fixes #3529)

### LINKED ISSUE #3529: lib/model: Temp index sorting databases remain on disk
I just noticed my backup program (attic/borg backup) backing up a number of syncthing folders, which are called tmp-index-sorter.##### and live in the config directory. I understand these are db files enabling the delta-indexing algorithm. In the past I would ignore folders by placing a "cachedir.tag" file into folders I don't want backed up.
In this case I was puzzled by their naming scheme though- they are called "tmp*" -something , but don't seem to be more temporal than the actual database-folder. My initial thought was thus not to place the cachedir.tag -files there, since "tmp"-folders tend to get cleaned up at times (?!)
Also, there are quite a number of "tmp-index-sorter"-folders, and it looks untidy. O:-) At least much less tidy than I am used to from syncthing.
Wouldn't it make more sense to at least put all the index-related db-folders into an extra subfolder say $st_home/index/ ?

So I guess I am proposing both a naming-scheme change and a location change...
EDIT: Just saw this related https://github.com/syncthing/syncthing/issues/3449 closed issue, so I guess you guys are somewhat aware of the issue

calmh:
### Purpose

Remove temp index databases left over after a crash or similar incident.

### Testing

I ran it, and it removed an old index db I had lying around.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -1124,15 +1124,16 @@ func autoUpgrade(cfg *config.Wrapper) {
 // suitable time after they have gone out of fashion.
 func cleanConfigDirectory() {
 	patterns := map[string]time.Duration{
-		"panic-*.log":      7 * 24 * time.Hour,  // keep panic logs for a week
-		"audit-*.log":      7 * 24 * time.Hour,  // keep audit logs for a week
-		"index":            14 * 24 * time.Hour, // keep old index format for two weeks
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 00a654845f8ea338dc87c1f3ca600e091ced2f6d. Thanks, @calmh!

PR #3554, opened 2016-08-30
Update syncthing.plist

matlegit:
### Purpose

I was looking into how to make Syncthing run on login and came across this plist from Syncthing's documentation. After reading Apple's own documentation on plists and trying it out, I believe that adding the SuccessfulExit subkey would be a great addition. 

If the user exits Syncthing, launchd won't forcibly try to restart it which can be useful if the user doesn't want Syncthing to be constantly running. This saves the user from having to manually unload Syncthing. Upon next login though, Syncthing will start up on login as usual. Let me know what you think, thanks!

For reference:
http://launchd.info/
https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man5/launchd.plist.5.html#//apple_ref/doc/man/5/launchd.plist
### Testing

All manually by loading and unloading via the launchctl command. Also logged in and out to test.
### Screenshots

None
### Documentation

### Authorship

### DIFF
--- a/etc/macosx-launchd/syncthing.plist
+++ b/etc/macosx-launchd/syncthing.plist
@@ -25,7 +25,10 @@
 		</dict>
 
 		<key>KeepAlive</key>
-		<true/>
+		<dict>
+			<key>SuccessfulExit</key>
+			<true/>
+		</dict>
 
 		<key>LowPriorityIO</key>
 		<true/>
--- a/etc/macosx-launchd/syncthing.plist
+++ b/etc/macosx-launchd/syncthing.plist
@@ -23,7 +23,10 @@
 			<key>STNORESTART</key>
 			<string>1</string>
 		</dict>
-
+		
+		<key>RunAtLoad</key>
+		<true/>
+		
 		<key>KeepAlive</key>
 		<dict>
 			<key>SuccessfulExit</key>


calmh:
This sounds reasonable to me. Don't mind all the failing build checks at the moment, those aren't your fault.

calmh:

calmh:
With this change Syncthing no longer restarts when selecting "Restart" in the menu. Which is odd as that is exit code 3?

matlegit:
Thanks for getting back so quickly, huge fan of your work.

As to the odd behavior, I believe that i forgot to add "RunAtLoad" since the documentation states "This key implies that "RunAtLoad" is set to true, since the job needs to run at least once before we can get an exit status."

calmh:
This still does not restart for me. Does it do that for you?

matlegit:
Unfortunately not, sorry about submitting a faulty PR. It's strange because i'm using a modified plist for my own project and Syncthing does restart with my project. It's running Syncthing 14.4 though. 

When I try with Syncthing 14.5 it doesn't restart like you said. I'll get down to the bottom of this, sorry again. Attached is my project plist

[my_plist.txt](https://github.com/syncthing/syncthing/files/444429/my_plist.txt)

calmh:
You're running without STNORESTART, meaning you have the monitor process inbetween launchd and Syncthing. That one only exists on actual requested shutdown from within the UI, which explains why it works. But it *should* work with launchd managing restarts and not having the monitor process in between I think...

PR #3556, opened 2016-09-02
lib/model, gui: Correct completion percentages when there are lots of deletes (fixes #3496)

### LINKED ISSUE #3496: Percentage shown parenthetically after the word 'syncing' can be a very large negative number
Please see screenshot.
untitled

I apologise if this problem - which will be, in part, a symptom of a more basic problem - has been reported before.

Windows 8.1 x64
Syncthing v0.14.4, Windows (64 bit)

EDIT: perhaps the problem owes to something shown in the screenshot, namely, that two different versions of Syncthing are in play (one on one device, the other on another).

calmh:
### Purpose

We used to consider deleted files & directories 128 bytes large. After the delta indexes change a bug slipped in where deleted files would be weighted according to their old non-deleted size. Both ways are incorrect (but the latest change made it worse), as if there are more files deleted than remaining data in the repo the needSize can be greater than the globalSize, resulting in a negative completion percentage.

This change makes it so that deleted items are zero bytes large, which makes more sense. Instead we expose the number of files that we need to delete as a separate field in the Completion() result, and hack the percentage down to 95% complete if it was 100% complete but we need to delete files. This latter part is sort of ugly, but necessary to give the user some sort of feedback.

Also, we must make sure to use the `(*FileInfo).FileSize` method to get the size of something when it can be a directory or a deleted item, as the `Size` field may be a lie in those cases.

### Testing

I added a unit test for the calculation, which failed before the fix. I haven't tested the corresponding fixup in the GUI because it's hard...

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -588,6 +588,7 @@ func (s *apiService) getDBCompletion(w http.ResponseWriter, r *http.Request) {
 		"completion":  comp.CompletionPct,
 		"needBytes":   comp.NeedBytes,
 		"globalBytes": comp.GlobalBytes,
+		"needDeletes": comp.NeedDeletes,
 	})
 }
 
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -439,18 +439,26 @@ angular.module('syncthing.core')
         }
 
         function recalcCompletion(device) {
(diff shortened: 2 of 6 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 5b37d0356cb3ff03edb45a79899dd7096403208a. Thanks, @calmh!

PR #3557, opened 2016-09-02
lib/model: Handle deleted-then-ignored files (fixes #3502)

### LINKED ISSUE #3502: Older files lost during update from v0.14.3 to 0.14.4
### Setup

Here's what I have:

 *  a Fedora 23 where syncthing and syncthing-inotify are started as user service.  Update is enabled so syncthing updates from web.  this has two folders: one called 'st' and other 'file-vornet.cz-rpm'
 *  a Debian Jessie VPS, syncing 'st' fully (also syncing it to two other machines that are offline now and also holding other folders for other machines)
 *  a Debian Jessie VM, hosted on the Fedora machine: syncing 'st' partially (just few megs)

All folders are "peers" (no "master"), and all 

### Incident

 1. opened QTSyncthingTray and noticed ST wants to update
 2. clicked the red update button (ignoring common advice about red buttons)
 3. ssh'd to the other two machines and started updating there via APT

Now at some point soon, I noticed in QTSyncthingTray that the 'st' folder is too small (~7M instead of ~1.7G) "Out Of Sync", "Failed items" loading forever, and items that are "Out of sync" are few seemingly random folders (some of them not accessed for long time). Also, Devices showed "syncing" status with negative percentage. Also "Last file received" shows "Deleted <some file>" but that's not the case.

See screenshot: http://file.vornet.cz/tmp/st-badstate.png

I jumped to the VPS, ran `du -s` (size) repeatedly and I could observe how the folder is shrinking.  I would not be able to see anything on the local VM as that one uses only tiny slice of the 'st' folder.

Luckily, I had relatively fresh backup (actually this machine is on dual boot with debian and there's separate instance of this 'st' folder so I used that.

### aftermath

Now what I have found about the issue:

 1. No folders seem to have gone: structure is preserved, but files are gone.

 2. Files/folders touched recently were OK.

 3. There's one folder which holds my payrolls (each in own sub-folder) and a script for decrypting them (I get encrypted zip via email).  Yesterday I added a payroll here: the payroll was present, old payrolls gone, and the script gone.  This shows that there's age-based system in which files got lost.

 4. There's another interesting folder "st/_sin" ("syncthing in", not "sin" :)): I recently moved into it another folder "ean-pdf" (from elsewhere in 'st' folder).  The "ean-pdf" is there with all its files, but other files in the "_sin" folder are gone. This seems strange: the "_sin" was techincally "changed recently", but got damaged; but the ean-pdf -- a "new arrival" or "subject of the change" did not.

 5. In the Fedora's log, soon after start of new syncthing, there's lot of lines like:

        Aug 14 03:35:53 fullmoon.brq.redhat.com syncthing[22127]: [ZINNF] INFO: Puller (folder "xvndu-bvmzg", dir "archive/personal"): delete: remove /home/amahdal/st/archive/personal: directory not empty

    where the dir is one that contains empty subdirectories.

 6.  'the file-vornet.cz-rpm' dir does not seem affected, maybe because it's recently changed and just holds couple of RPMS and a script...

### Version Information

Syncthing Version: v0.14.4
OS Version: Fedora 23 (other devices: Debian Jessie)

I'm assuming the culprit happened on Fedora: from the "Deleted" info it appears as if the VPS deleted it, but the local VM never had it and as I said I could observe the file loss on the VPS; it looks as if there was already some misunderstanding going.

calmh:
### Purpose

When files that were previously marked as deleted became ignored, we used to do nothing at all. This changes that behavior to set the Invalid bit (that we should rename to Ignored). This then becomes an update to other devices that they should not trust our knowledge about the file in question.

Read this diff without whitespace...

### Testing

Tested by
- creating a bunch of files on s1
- letting them sync to s2
- shutting down s2
- deleting the files on s1 and rescanning
- adding the files to .stignore on s1 and rescanning
- starting up s2 and letting it sync
- observing the files are not deleted on s2, and it considers itself up
  to date.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1735,41 +1735,46 @@ func (m *Model) internalScanFolderSubdirs(folder string, subDirs []string) error
 		subDirs = []string{""}
 	}
 
-	// Do a scan of the database for each prefix, to check for deleted files.
+	// Do a scan of the database for each prefix, to check for deleted and
+	// ignored files.
 	batch = batch[:0]
 	for _, sub := range subDirs {
 		var iterError error
 
 		fs.WithPrefixedHaveTruncated(protocol.LocalDeviceID, sub, func(fi db.FileIntf) bool {
 			f := fi.(db.FileInfoTruncated)
-			if !f.IsDeleted() {
-				if len(batch) == batchSizeFiles {
-					if err := m.CheckFolderHealth(folder); err != nil {
-						iterError = err
-						return false
-					}
-					m.updateLocalsFromScanning(folder, batch)
-					batch = batch[:0]
+			if len(batch) == batchSizeFiles {
+				if err := m.CheckFolderHealth(folder); err != nil {
+					iterError = err
+					return false
 				}
+				m.updateLocalsFromScanning(folder, batch)
+				batch = batch[:0]
+			}
 
-				if !f.IsInvalid() && (ignores.Match(f.Name).IsIgnored() || symlinkInvalid(folder, f)) {
-					// File has been ignored or an unsupported symlink. Set invalid bit.
-					l.Debugln("setting invalid bit on ignored", f)
-					nf := protocol.FileInfo{
-						Name:          f.Name,
-						Type:          f.Type,
-						Size:          f.Size,
-						ModifiedS:     f.ModifiedS,
-						ModifiedNs:    f.ModifiedNs,
-						Permissions:   f.Permissions,
-						NoPermissions: f.NoPermissions,
-						Invalid:       true,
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Did you repro the actual removal tho?

calmh:
Actually no, I just think that I understand it, so let me do that as well.

calmh:
Yes. The exact same procedure described above results in the files gone from s2, without this patch.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 46a143e80e13e2fface4ac22d6b0beaf21c2e2ed. Thanks, @calmh!

PR #3558, opened 2016-09-02
lib/config, cmd/syncthing: Enforce localhost only connections

calmh:
### Purpose

When the GUI/API is bound to localhost, we enforce that the Host header looks like localhost. This can be disabled by setting insecureSkipHostCheck in the GUI config.

### Testing

Verified by a new unit test!

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -313,6 +313,11 @@ func (s *apiService) Serve() {
 	// Add the CORS handling
 	handler = corsMiddleware(handler)
 
+	if addressIsLocalhost(guiCfg.Address()) && !guiCfg.InsecureSkipHostCheck {
+		// Verify source host
+		handler = localhostMiddleware(handler)
+	}
+
 	handler = debugMiddleware(handler)
 
 	srv := http.Server{
@@ -495,6 +500,17 @@ func withDetailsMiddleware(id protocol.DeviceID, h http.Handler) http.Handler {
 	})
 }
 
+func localhostMiddleware(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if addressIsLocalhost(r.Host) {
+			h.ServeHTTP(w, r)
+			return
+		}
+
+		http.Error(w, "Host check error", http.StatusForbidden)
+	})
+}
+
 func (s *apiService) whenDebugging(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if s.cfg.GUI().Debugging {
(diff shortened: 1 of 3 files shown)


cmd/syncthing/gui.go
```diff
@@ -1292,3 +1307,17 @@ func dirNames(dir string) []string {
 	sort.Strings(dirs)
 	return dirs
 }
+
+func addressIsLocalhost(addr string) bool {
+	host, _, err := net.SplitHostPort(addr)
```
AudriusButkevicius: What about ":8080"

AudriusButkevicius: What about ":8080"

calmh: Let me write a test for this function

calmh: Covered!

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 49910a1d853a8c392e5faa00c3c2aee0fc060d78. Thanks, @calmh!

sbraz:
@calmh Correct me if I'm wrong but I think you forgot to update the doc in case syncthing is behind a reverse proxy: https://docs.syncthing.net/users/reverseproxy.html#nginx does not work any more, I had to remove the `#proxy_set_header        Host $host;` line.

AudriusButkevicius:
Read the FAQ.

D4rk4:
@AudriusButkevicius where?
https://docs.syncthing.net/users/faq.html#why-do-i-get-host-check-error-in-the-gui-api

Plz, write docs, for new features... Also this feature break Syncthing-GTK....

AudriusButkevicius:
If you open the config with a text editor it will be set to false in the options section.
Or do what the faq explains and disabled it via advanced options in the GUI

D4rk4:
@AudriusButkevicius realy? It's not new installation. And no options like this in config:
grep -ic insecureSkipHostcheck config.xml                                                            
0
Me try add "<insecureSkipHostcheck>false</insecureSkipHostcheck>" in "options" section. But it's not working.

AudriusButkevicius:
Use the UI.

D4rk4:
@AudriusButkevicius UI? It's server. UI say "Host check error". LOL you broke Syncthing on many servers. Bravo!

canton7:
The change was made to fix the possibility of a DNS rebinding attack. Laugh all you like, the change was made to keep users safe.

D4rk4:
@canton7 okay. How i can disable this feature in config.xml? UI not accessible now...

canton7:
> @canton7 okay. How i can disable this feature in config.xml? UI not accessible now...

Please actually read my previous message, where I said:

> You need to add <insecureSkipHostcheck>true</insecureSkipHostcheck> to the gui section.

D4rk4:
@canton7 thx. It's working!

PR #3560, opened 2016-09-03
cmd/syncthing: Copy config on upgrade, instead of renaming (fixes #3525)

### LINKED ISSUE #3525: Syncthing upgrade can end up with no config.xml when there is not enough space on device
Hello there,

I somehow managed to completely fill up my /home drive on one server. I noticed it too late and as result, syncthing failed to write to `config.xml` somewhere in the process and truncated it to 0b.
Config was then replaced with default one upon next start.

Luckily, following is not my case, but this could be rather unpleasant for someone without backups.

Usual way how to _not_ end up with invalid config is to write `config.xml.tml` and mv it only after successful write.

### Version Information

Syncthing Version: syncthing v0.14.4 "Dysprosium Dragonfly" (go1.6.3 linux-amd64) builduser@svetlemodry 2016-08-10 19:32:43 UTC
OS Version: Manjaro Linux, fully updated as on Fri Aug 19 07:38:46 CEST 2016
(It's headless machine, so I'm not using any browser nor Syncthing-GTK there)

calmh:
### Purpose

Make sure that if we're out of space when it comes to upgrade, we don't end up not having a config.xml at all.

### Testing

```
[BN3ZH] 15:43:11 INFO: Archiving a copy of old config file format at: /Volumes/Untitled//config.xml.v14
[BN3ZH] 15:43:11 FATAL: Config archive: open /Volumes/Untitled//config.xml.v14: no space left on device
[monitor] 15:43:11 INFO: Syncthing exited: exit status 1
jb@retro:~ $ ls -l /Volumes/Untitled/config.xml 
-rw-------  1 jb  staff  3041 Sep  3 15:42 /Volumes/Untitled/config.xml
```

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -889,17 +889,30 @@ func loadOrCreateConfig() *config.Wrapper {
 }
 
 func archiveAndSaveConfig(cfg *config.Wrapper) error {
-	// To prevent previous config from being cleaned up, quickly touch it too
-	now := time.Now()
-	_ = os.Chtimes(cfg.ConfigPath(), now, now) // May return error on Android etc; no worries
-
+	// Copy the existing config to an archive copy
 	archivePath := cfg.ConfigPath() + fmt.Sprintf(".v%d", cfg.Raw().OriginalVersion)
 	l.Infoln("Archiving a copy of old config file format at:", archivePath)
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 7990ffcc6099509e1fc34c53c343d7a075af4e71. Thanks, @calmh!

PR #3606, opened 2016-09-17
Fix a bug on windows when using external versioner

nrm21:
### Purpose

Fixes a bug on windows where, when using an external versioner (such as a batch file), filenames that contain spaces are simply ignored because the cmd.run() command errors out silently and won't call the batch script.

### Testing

To test simply setup a folder on windows with external versioning and call a batch file.  Place two files (one containing space and the other without space in the filename) into the dir and sync to another.  Then delete the files one by one (on another machine) and sync between each.  You should observe the batch script runs for the one file and not the other.  The batch file doesn't have to delete, it can do anything (like echo hello world to a file on the desktop just as an example).  You will see the batch file fire for one and not the other.

### Documentation

This MIGHT be just a Windows 10 issue since I didn't test with Windows 7 but I imagine it is for both.  Also I didn't test with Linux (since I don't use ST on that OS) and if the issue exists for Linux this might also fix for that as well by removing the if runtime.GOOS == "windows" check.

If the pull req is accepted I can provide the batch script I use as an example of how to delete to recycle bin that I use by modifying this document: https://docs.syncthing.net/users/versioning.html

### Authorship

Nate Morrison

### DIFF
--- a/lib/versioner/external.go
+++ b/lib/versioner/external.go
@@ -65,7 +65,7 @@ func (v External) Archive(filePath string) error {
 
 	// Windows Go has a bug where it won't properly run the external versioner
 	// if there is a space in the filename.  This code fixes that bug.
-    if runtime.GOOS == "windows" {
+	if runtime.GOOS == "windows" {
 		filenameparts := strings.Split(inFolderPath, " ")
 		// Remove the last element
 		cmd.Args = cmd.Args[:len(cmd.Args)-1]
--- a/lib/versioner/external.go
+++ b/lib/versioner/external.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/syncthing/syncthing/lib/osutil"
@@ -61,6 +62,19 @@ func (v External) Archive(filePath string) error {
 	}
 
 	cmd := exec.Command(v.command, v.folderPath, inFolderPath)
+
+	// Windows Go has a bug where it won't properly run the external versioner
+	// if there is a space in the filename.  This code fixes that bug.
+    if runtime.GOOS == "windows" {
+		filenameparts := strings.Split(inFolderPath, " ")
+		// Remove the last element
+		cmd.Args = cmd.Args[:len(cmd.Args)-1]
+		// And reappend all filename parts as seperate args
+		for _, v := range filenameparts {
+			cmd.Args = append(cmd.Args, v)
+		}
+	}
+
 	env := os.Environ()
 	// filter STGUIAUTH and STGUIAPIKEY from environment variables
 	filteredEnv := []string{}


nrm21:
It's possible it's in the interaction between go and the script, or even in the exe I'm using to send file to recycle bin (for which I don't have access to the code).  But I can tell you when I echo out exactly what is sent to the script from ST version handler and the run that manually... it works exactly as expected... just not when ST send's it by itself.  So I propose we change the behavior of the handler slightly and just document it (mind you there is no change in behavior for files with no spaces).

Currently the handler runs:
<scriptpath> <folderpath> <filepiece1 filepiece2 filepiece3.txt>
and I'm proposing:
<scriptpath> <folderpath> <filepiece1> <filepiece2> <filepiece3.txt> ...

How do you suggest I

calmh:
What happens when you do the echo test with the current setup then, and what does your script look like when you do so?

> Currently the handler runs:
> `<scriptpath> <folderpath> <filepiece1 filepiece2 filepiece3.txt>`
> and I'm proposing:
> `<scriptpath> <folderpath> <filepiece1> <filepiece2> <filepiece3.txt>`

On the face of it this looks like the difference between `command folderpath "filepiece1 filepiece2 filepiece3.txt"` and `command folderpath filepiece1 filepiece2 filepiece3.txt` where the first behaviour *should* be the correct one from my point of view. Is there something specific about Windows or Go that makes this not the case?

calmh:
This test passes, which seems to indicate to me that the bat file in that test can access and delete the long file named file using the current code: https://github.com/syncthing/syncthing/pull/3609

calmh:

nrm21:
Ok so I retested and yes I can get the bat to run either way if just echoing the command it will run to a file.  however when I try to run the command from ST with a file with spaces, it does not work.  It does however work fine on that same script with a file that contains no spaces.

And I agree with you that the first behavior SHOULD be the correct one.  I'm baffled why it's not working for me.

nrm21:
For your benefit here is the batch script I am running:

`:: Change to script dir
cd /d %~dp0

:: Set variables and get all remaining variables starting from
:: %2 into the filepath variable (in case of spaces in filename)
for /f "tokens=1*" %%a in ("%*") do (
  set folderpath=%%a
  set filepath=%%b
)
:: Remove any quotes from filename
set filepath=%filepath:"=%
:: Remove strange \\?\ from beginning of windows path
set folderpath=%folderpath:~4%

"%cd%\Recycle.exe" -f "%folderpath%\%filepath%"`

calmh:
Check out the bat file in the test I just added, as it seems to do the right thing. I copied it from the example on our docs site as I don't really speak `.bat` fluently.

nrm21:
Ok so I tried that bat file.  Same effect it runs fine for a file with no spaces... doesn't run for a file with.  I'm kinda surprised no one else has mentioned or run into this bug.  Then again I don't know how many people on Windows might be using an external verisoner.

canton7:
Can you put a pause in the batch file, get syncthing to call it, and use task manager to see what command line parameters it was called with? (you'll need to enable an extra column I'm the process tab).

nrm21:
I already tried that, pause doesn't seem to respond even when the batch file runs every other line properly (like echo to file or delete file).

canton7:
Probably because there's no interactive console attached. A long sleep will do the trick too.

nrm21:
Ok after a lot of testing it seems I have fixed my issue.  The ultimate problem somehow related to the recycle script and exe being located in the "C:\Program Files (x86)\" dir.  Having those parenthesis in the exe path was causing windows to freak out (it wasn't even the spaces).

GRRRR sometimes I just hate windows and their stupid attitude of (add every possible special char into our friggin SYSTEM directory paths)!

As soon as I moved my script into a directory with no special chars in it everything seemed to work correctly.  You can close this pull request.  Sorry for wasting your time.

ProactiveServices:
This should be re-opened as it's not fixed. Batch files that use paths with parenthesis need a bit of special treatment. I can also double-check if/where quotes are being passed down the chain using Sysinternals Process Monitor (it'll take a record of all the commands and their parameters) but don't have the time just yet.

calmh:
There appears to be an issue executing the external versioner script, on Windows, when it's in a path with spaces or parenthesis in it. An issue should be opened for that problem, though this is not that so this should remain closed.

There is no problem giving paths with spaces or parenthesis to the bat file.

nrm21:
Just tested with "Program Files" same symptoms as "Program Files (x86)".

Interestingly, I think ST is still trying to delete file in the background on rescan, because after failure to delete I can move the directory that recycle script and exe is in out of either of those dirs to the main root C:\ and as soon as I change the path fo the version script in the folder edit settings.... within seconds it then deletes the file to recycle bin.

calmh:
I think this is a bug in Go's `exec.Command`, actually looking into it just now for fun but it looks hairy and way too Windowsy for my taste so I'll probably leave it at just filing an issue.

nrm21:
Yeah don't worry about it on my behalf. You have bigger issues to deal with I'm sure.  I was sort of scratching my own itch here and since I found another way to accomplish same, I'm happy.  :)

PR #3618, opened 2016-09-24
lib/upnp: Correct the result deduplication mechanism (fixes #3578)

### LINKED ISSUE #3578: UPnP not doing port mapping on all discovered routers
I've two routers in the network to access internet with UPnP (NAT) enabled and Syncthing, and most of the times it adds the port mapping to the incorrect router.

The mapping will be added to the default gateway used by the interface.

### Version Information

Syncthing Version: v0.14.6
OS Version: Windows 10 (64 bit)

calmh:
### Purpose

The deduplication was broken as we kept a pointer to the loop variable, but the loop variable is overwritten on each iteration.

### Testing

None, unfortunately.

### DIFF
--- a/lib/upnp/upnp.go
+++ b/lib/upnp/upnp.go
@@ -108,19 +108,21 @@ func Discover(renewal, timeout time.Duration) []nat.Device {
 		close(resultChan)
 	}()
 
+	seenResults := make(map[string]bool)
 nextResult:
 	for result := range resultChan {
-		for _, existingResult := range results {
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
@st-review merge

st-review:

st-review:
:ok_hand: Merged as 3cabecda04d43352200a41e54d08fa1d54a7abc9. Thanks, @calmh!

Issue #3586, opened 2016-09-10
Feature request: Show countdown until next folder scan


eliempje:
Currently for a sync folder one can only see the time at witch interval the folder is scanned. E.g. 600 seconds.
Idea is to add a countdown timer.

scienmind:
As in "live display" for the seconds left until the next scan?

While I can see that is can be "kinda nice" to see the internal clocks of it and see the time left, I can't see any practical purpose.
It'll just add extra frequent-changing element to the GUI, taxing the CPU.

If you need to scan now, and wonder if wait until the next scan, just hit "rescan".
Besides, the best practice is to use inotify anyways so that the scanner will be auto-invoked upon file modifications. In that case, I also can't see why would it be useful.

canton7:
JS is pretty good at not doing anything if you're not looking at the tab, and we don't need to hit syncthing, so I can't see the issue.

eliempje:
Ok I get the point

calmh:
Syncthing could just send an "next rescan at: $timestamp" attribute in the state change event, and the gui could do the rest. So this should be fairly simple and clean to do.

uok:
I think a simple (not ticking) info "next scan: (timestamp = last scan time + scan interval)" is helpful

calmh:
We randomly fuzz the scan interval a bit (about +/- 30% iirc), so it needs to be announced after each scan to be reliable. (This is just to avoid folders staying in lockstep with each other)

uok:
correct, I just remembered this too (see https://github.com/syncthing/syncthing/commit/54c3caad530fb16a3764ee70955280f042f2860f)

canton7:
There's the fun around syncthing-inotify randomly delaying scans as well?

calmh:
Yes, but Syncthing knows that and could in fact reflect it in this way. That would make the ui less of a lie on this point than it is now.

canton7:
Ah, no, I guess the GUI gets events when the next scan is delayed, because a scan just happened... duh. Carry on.

Ferroin:
I like the idea of having a time-stamp showing when the next scan will be as opposed to a counter showing remaining time.  Doing the time-stamp would allow a mostly static element instead of one that needs to update every second, and makes more sense when you start talking about scan delays longer than a few minutes.  Having the ability to choose between the two, and possibly also just disable it would be even better (especially considering that people are not likely to agree on which is better).

canton7:
[...]

PR #3624, opened 2016-09-25
Add "Next Scan" timestamp to folder view #3586

hfichtenberger:
### Purpose

Adds an entry _Next Scan_ to each folder in the GUI that shows when the next scan will happen (similar to the _Last Scan_ entry) as requested by #3586.
### Testing

I didn't find any tests I could hook into. Currently, I don't think I can write a test in the way I should to test this feature from scratch (i.e., set up the environment, create a dummy share and check whether the time shows right) because I'm just starting to read the syncthing code. I checked that the next scan is correct on my installation manually. I also checked that if the timestamp is not greater than the last scan's timestamp the entry is hidden.

However, it seems to me that `func (f *folderScanner) Reschedule() time.Time` is called just after the StateChange event in `func (m *Model) internalScanFolderSubdirs(folder string, subDirs []string) error` is fired. I think that, in theory, the GUI could query the new folder state just before the next scan timestamp is updated. I would suggest to fire another StateChanged event right after updating the next scan timestamp in the folderState object. It seems to me that there is no intended way to do this without actually changing the state to another value (see `func (s *stateTracker) setState(newState folderState)`). What would be the right way to do this?
### Screenshots

firefox_2016-09-27_22-29-32
### Documentation

A new screenshot of the GUI may be added to docs/intro/gui.rst (the current one seems to be outdated anyway).

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -390,13 +390,12 @@
                       <th><span class="fa fa-fw fa-clock-o"></span>&nbsp;<span translate>Last Scan</span></th>
                       <td translate ng-if="folderStats[folder.id].lastScanDays >= 365" class="text-right">Never</td>
                       <td ng-if="folderStats[folder.id].lastScanDays < 365" class="text-right">
-                        <span>{{folderStats[folder.id].lastScan | date:'yyyy-MM-dd HH:mm:ss'}}</span>
-                      </td>
-                    </tr>
-                    <tr ng-if="folderStats[folder.id].nextScanDelta > 0">
-                      <th><span class="fa fa-fw fa-clock-o"></span>&nbsp;<span translate>Next Scan</span></th>
-                      <td class="text-right">
-                        <span>{{folderStats[folder.id].nextScan | date:'yyyy-MM-dd HH:mm:ss'}}</span>
+                        <span tooltip ng-if="folderStats[folder.id].nextScanDelta > 0" data-original-title="<strong><span class='fa fa-fw fa-angle-double-right'></span><span translate>Next scan</span></strong><br> {{folderStats[folder.id].nextScan | date:'yyyy-MM-dd HH:mm:ss'}}">
+                          <span>{{folderStats[folder.id].lastScan | date:'yyyy-MM-dd HH:mm:ss'}}</span>
+                        </span>
+                        <span ng-if="folderStats[folder.id].nextScanDelta <= 0">
+                          <span>{{folderStats[folder.id].lastScan | date:'yyyy-MM-dd HH:mm:ss'}}</span>
+                        </span>
                       </td>
                     </tr>
                     <tr ng-if="folder.type != 'readonly' && folderStats[folder.id].lastFile && folderStats[folder.id].lastFile.filename">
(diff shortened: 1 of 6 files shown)


lib/stats/folder.go
```diff
@@ -15,6 +15,7 @@ import (
 type FolderStatistics struct {
 	LastFile LastFile  `json:"lastFile"`
 	LastScan time.Time `json:"lastScan"`
+	NextScan time.Time `json:"nextScan"`
```
AudriusButkevicius: This number will be a lie after restarting most likely.

gui/default/index.html
```diff
@@ -393,6 +393,12 @@ <h4 class="panel-title">
                         <span>{{folderStats[folder.id].lastScan | date:'yyyy-MM-dd HH:mm:ss'}}</span>
                       </td>
                     </tr>
+                    <tr ng-if="folderStats[folder.id].nextScanDelta > 0">
+                      <th><span class="fa fa-fw fa-clock-o"></span>&nbsp;<span translate>Next Scan</span></th>
+                      <td class="text-right">
+                        <span>{{folderStats[folder.id].nextScan | date:'yyyy-MM-dd HH:mm:ss'}}</span>
+                      </td>
+                    </tr>
```
AudriusButkevicius: Can we just have next time as a tooltip for the last scan? UI space is precious.

calmh:
Haven't looked at the code yet, but to answer the question in the description... we should send the next scan time in the state change to "idle". That may mean we need to calculate it earlier, I don't know. We should also consider that an ongoing pull is going to potentially delay the next scan, just so that this doesn't get confusing in the GUI.

uok:
@hfichtenberger thanks for working on that improvement! Suggestion for icon: fa-angle-double-right or fa-forward

hfichtenberger:
@calmh As I said I'm not too familiar with the high-level philosophy of the code but I guess the next scan time shouldn't be calculated and set in `func (m *Model) internalScanFolderSubdirs(folder string, subDirs []string)` because this is not the semantic scope of this method. Actually, it's also called for on-demand scans so that rescheduling might not be intended in this case. I can imagine too possible solutions: One would be to pull calls to `func (s *stateTracker) setState(newState folderState)` into methods on a higher level, probably into `func (f *rwFolder) Serve()` and `func (f *roFolder) Serve()` (which means that the actual scanning would be detached from the management of the folder state, so one would have to check the call graph very carefully for all calls to the scan methods). The other solution would be to introduced a method like `func (s *stateTracker) updateState()`, which could fire a StateChanged "from x to x" event that indicates a change of the _state metadata_. This one seems less invasive to me with the downside of firing to events.

lib/stats/folder.go
hfichtenberger: I see that this is saved into some LevelDB file. As the next scan time is calculated in any case after restarting, I guess that this shouldn't be saved in the first place. Do you know by chance where this serialization happens?

AudriusButkevicius: So I suggest you wedge it in here:

https://sourcegraph.com/github.com/hfichtenberger/syncthing@faf3a96db78f3a17449d3abbc0ec2c3dd8dab6a8/-/blob/cmd/syncthing/gui.go#L612

So just add a new method to the model m.NextScan("folder") which is potentially stored on the runner, and then returned in the summary.

calmh:
I hear you. Propose a solution. :)

calmh:
Timeout, please feel free to reopen if this gets closer to completion.

PR #3626, opened 2016-09-26
Diskchange api

nrm21:
### Purpose

Adds a URL handler for the 'LocalDiskChange' events, so it has it's own buffered subscription.

### Testing

curl -H "X-API-Key: Z0mgS3cr3tAPIkey" http://localhost:8384/rest/events/disk | yourfavjsonprettyprinter

### Documentation

This should keep the main events sub from getting polluted with local disk change events and vice versa.  Also it gives people a programmatic way to view what files were modified on what server by polling each nodes API service (Of course they would need to set nodes to listen on 0.0.0.0 instead of 127.0.0.1 first and prob would then want to set SSL as well).  So if someone were to create a program that gathered all that data to create a global log, they could.

### Authorship

Nate Morrison (nrm21)

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -231,8 +231,8 @@ func (s *apiService) Serve() {
 	getRestMux.HandleFunc("/rest/db/need", s.getDBNeed)                          // folder [perpage] [page]
 	getRestMux.HandleFunc("/rest/db/status", s.getDBStatus)                      // folder
 	getRestMux.HandleFunc("/rest/db/browse", s.getDBBrowse)                      // folder [prefix] [dirsonly] [levels]
-	getRestMux.HandleFunc("/rest/events", s.getEvents)                           // since [limit]
-	getRestMux.HandleFunc("/rest/events/disk", s.getEvents)                      // since [limit]
+	getRestMux.HandleFunc("/rest/events", s.getIndexEvents)                      // since [limit]
+	getRestMux.HandleFunc("/rest/events/disk", s.getDiskEvents)                  // since [limit]
 	getRestMux.HandleFunc("/rest/stats/device", s.getDeviceStats)                // -
 	getRestMux.HandleFunc("/rest/stats/folder", s.getFolderStats)                // -
 	getRestMux.HandleFunc("/rest/svc/deviceid", s.getDeviceID)                   // id
@@ -996,23 +996,21 @@ func (s *apiService) postDBIgnores(w http.ResponseWriter, r *http.Request) {
 	s.getDBIgnores(w, r)
 }
 
-func (s *apiService) getEvents(w http.ResponseWriter, r *http.Request) {
-	// If this is a query to '/rest/events/disk' then we set the isDiskUpdate flag here
-	isDiskUpdate := false
-	if strings.Contains(r.URL.String(), "/disk") {
-		isDiskUpdate = true
-	}
+func (s *apiService) getIndexEvents(w http.ResponseWriter, r *http.Request) {
+	s.fss.gotEventRequest()
+	s.getEvents(w, r, s.eventSub)
+}
 
+func (s *apiService) getDiskEvents(w http.ResponseWriter, r *http.Request) {
+	s.getEvents(w, r, s.diskEventSub)
+}
+
+func (s *apiService) getEvents(w http.ResponseWriter, r *http.Request, eventSub events.BufferedSubscription) {
 	qs := r.URL.Query()
 	sinceStr := qs.Get("since")
 	limitStr := qs.Get("limit")
 	since, _ := strconv.Atoi(sinceStr)
 	limit, _ := strconv.Atoi(limitStr)
-
-	// Only needed if this is a regular events update
-	if !isDiskUpdate {
-		s.fss.gotEventRequest()
-	}
 
 	// Flush before blocking, to indicate that we've received the request and
 	// that it should not be retried. Must set Content-Type header before
@@ -1021,13 +1019,7 @@ func (s *apiService) getEvents(w http.ResponseWriter, r *http.Request) {
 	f := w.(http.Flusher)
 	f.Flush()
 
-	// Call whichever sub according to the query used
-	var evs []events.Event
-	if isDiskUpdate {
-		evs = s.diskEventSub.Since(since, nil)
-	} else {
-		evs = s.eventSub.Since(since, nil)
-	}
+	evs := eventSub.Since(since, nil)
 	if 0 < limit && limit < len(evs) {
 		evs = evs[len(evs)-limit:]
 	}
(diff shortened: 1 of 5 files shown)


nrm21:
Ok well st-review is still not happy.  Says there's no test but I added one in http-test.go

nrm21:
In addition to the API commit, I saw that my original code only included folder ID.  I suspect back then we weren't using ID's and labels so I added that as well for clarity.  Most people will want to see the folder label for clarity not the machine readable ID.

AudriusButkevicius:
It could do with an explanation why this is useful

nrm21:
For those who want a global log (of what node changed what file) they can now programmatically query an API on each node for that info.

The API structure was there already and the local machine was already logging this info to the audit log, this just formalizes a way someone can remotely fetch it.

nrm21:
Ok so I'm ready to pull this unless someone has objection.

lib/model/model.go
```diff
@@ -1467,10 +1467,7 @@ func (m *Model) updateLocalsFromScanning(folder string, fs []protocol.FileInfo)
 
 	// Fire the LocalChangeDetected event to notify listeners about local
 	// updates.
-	m.fmut.RLock()
```
calmh: Locking still required here.

cmd/syncthing/gui.go
```diff
@@ -1017,6 +1020,30 @@ func (s *apiService) getEvents(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func (s *apiService) getDiskEvents(w http.ResponseWriter, r *http.Request) {
+	qs := r.URL.Query()
+	sinceStr := qs.Get("since")
+	limitStr := qs.Get("limit")
+	since, _ := strconv.Atoi(sinceStr)
+	limit, _ := strconv.Atoi(limitStr)
+
+	s.fss.gotEventRequest()
```
calmh: Not needed here, as listening to these events doesn't imply an interest in the folder summary events.

cmd/syncthing/gui.go
```diff
@@ -1017,6 +1020,30 @@ func (s *apiService) getEvents(w http.ResponseWriter, r *http.Request) {
 	sendJSON(w, evs)
 }
 
+func (s *apiService) getDiskEvents(w http.ResponseWriter, r *http.Request) {
```
calmh: Given that this is identical to the other events handler, apart from the "gotEventRequest" comment below, I think it should be refactored into a common function/method that just takes the relevant event subscription as a parameter.

nrm21: Heh, I was wondering if you'd want that or not.  I gambled that you wouldn't, so what just an extra bool parameter then in the input to differentiate?

calmh:
Also, please write up the relevant thing on the docs repo for this while it's still fresh in everyone's memory.

cmd/syncthing/gui.go
```diff
@@ -997,37 +997,22 @@ func (s *apiService) postDBIgnores(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *apiService) getEvents(w http.ResponseWriter, r *http.Request) {
```
calmh: Ah, no, this is not what I meant, sorry. :) It would be cleaner to have the two handlers as you did, but they both call out to a third method that does most of the work, with the specific event subscription to return as a parameter.

lib/model/model.go
```diff
@@ -1499,7 +1499,9 @@ func (m *Model) updateLocals(folder string, fs []protocol.FileInfo) {
 
 func (m *Model) localChangeDetected(folder config.FolderConfiguration, files []protocol.FileInfo) {
 	// For windows paths, strip unwanted chars from the front
+	m.fmut.RLock()
```
calmh: The locking is needed where you access m.folderCfgs, as that is a non-concurrent-safe map that may be changed by other routines in parallell.

nrm21:
How about now?  I put the mutex lock inside the deepest function.  I saw before it was only around the .Path() call anyway so I figured to keep it locked as short a time as necessary.

Also getEvents() merged.  I used a flag based on if "/disk" is in the URL query

lib/model/model.go
nrm21: Ahh I see... ok I'll fix that.

nrm21:
Ok try now.

nrm21:
Also I added the appropriate doc files in syncthing/docs

nrm21:
So are we good for a merge?

lib/model/model.go
```diff
@@ -1465,12 +1465,10 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
 	m.updateLocals(folder, fs)
 
-	// Fire the LocalChangeDetected event to notify listeners about local
-	// updates.
+	// Fire the LocalChangeDetected event to notify listeners about local updates.
 	m.fmut.RLock()
-	path := m.folderCfgs[folder].Path()
+	m.localChangeDetected(m.folderCfgs[folder], fs)
 	m.fmut.RUnlock()
-	m.localChangeDetected(folder, path, fs)
```
AudriusButkevicius: It looks like this was called intentionally outside of the lock before. Why is this ok now?

nrm21: Before I was calling the lock on the 'folderConfiguration.Path()' method and immediately releasing.  Now I am passing in the whole folder object 'm.folderCfgs[folder]' so I can get a few more things from it.  I was originally going to run the lock again only on that part but then Jakob wrote 

> The locking is needed where you access m.folderCfgs, as that is a non-concurrent-safe map that may be changed by other routines in parallell.

So I translated that to mean... lock everything that touches this until it stops touching it.

AudriusButkevicius: I'd rather it got what it needed under the lock, and passed into the function outside of the lock. If there are a lot of files, the lock could be pointlessly held for long periods of time.

nrm21: Ok done.

lib/model/model.go
```diff
@@ -1526,14 +1523,16 @@ func (m *Model) localChangeDetected(folder, path string, files []protocol.FileIn
 			action = "deleted"
 		}
 
-		// The full file path, adjusted to the local path separator character.
-		path := filepath.Join(path, filepath.FromSlash(file.Name))
+		// The full file path, adjusted to the local path separator character.  Also
+		// for windows paths, strip unwanted chars from the front.
+		path := filepath.Join(strings.Replace(folderCfg.Path(), `\\?\`, "", 1), filepath.FromSlash(file.Name))
```
calmh: Last comment, I think we should keep the `strings.Replace(folderCfg.Path(), `\\?\`, "", 1)` outside of the loop as it does not need to be done for each iteration and this stuff gets called a lot after all.

calmh:
@st-review lgtm after that last comment is fixed

st-review:

lib/model/model.go
nrm21: fixed

AudriusButkevicius:
@st-review lgtm

st-review:
:ok_hand: Merged as d322ebd0b9d5a6c1025c29d4519c4cde21def78d. Thanks, @nrm21!

PR #3627, opened 2016-09-26
there is no network.target for user instances

peterhoeg:
### Purpose

network.target is only available to system services and not user services so there is no point specifying a dependency.

Further to: https://github.com/syncthing/syncthing-inotify/pull/130

### Testing

The previous depedency never worked - it was silently dropped. It runs fine here with and without.

### Screenshots

n/a

### Documentation

n/a

### DIFF
--- a/etc/linux-systemd/user/syncthing.service
+++ b/etc/linux-systemd/user/syncthing.service
@@ -1,7 +1,6 @@
 [Unit]
 Description=Syncthing - Open Source Continuous File Synchronization
 Documentation=man:syncthing(1)
-After=network.target
 Wants=syncthing-inotify.service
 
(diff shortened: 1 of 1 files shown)


canton7:
Ping @rumpelsepp

calmh:
@st-review merge it

etc/linux-systemd: Remove bogus dependency on networking for user unit

st-review:
:ok_hand: Merged as 891ff383ec25a33bd3c04f4382d8754629e0c6e9. Thanks, @peterhoeg!

rumpelsepp:
Lgtm. Thx for merging, I was in the US for 1 month and not plugged into an "internet cable".

PR #3629, opened 2016-09-27
Add support to TLS renegotiation

GodUseVPN:
I have the discovery server running in http mode behind SSL-secured apache.
Configuration as:
```
        <Location /v2>
                SSLVerifyClient optional_no_ca
        </Location>
        RequestHeader set X-SSL-Cert "%{SSL_CLIENT_CERT}s"
        ProxyPass /ping http://127.0.0.1:8484/ping
        ProxyPass /v2 http://127.0.0.1:8484/v2
```
This case cause tls renegotiation, and announceClient will failed with "tls: no renegotiation"

The change require Go 1.7.1 or higher to compile. Refer to [crypto/tls: does not support renegotiation #5742](https://github.com/golang/go/issues/5742)

### DIFF
--- a/lib/discover/global.go
+++ b/lib/discover/global.go
@@ -89,6 +89,7 @@ func NewGlobal(server string, cert tls.Certificate, addrList AddressLister) (Fin
 			Proxy: http.ProxyFromEnvironment,
 			TLSClientConfig: &tls.Config{
 				InsecureSkipVerify: opts.insecure,
+				Renegotiation: tls.RenegotiateFreelyAsClient,
 				Certificates:       []tls.Certificate{cert},
 			},
 		},


AudriusButkevicius:
Is it not easier just to use a set of secure ciphers?

GodUseVPN:
Secure ciphers settings can't avoid renegotiation.

AudriusButkevicius:
So I don't understand why this would cause a problem in the first place.
I suspect the reverse proxy connection from apache sends the renegotiation extension for no good reason?

calmh:
No. TLS renegotiation is hairy business and we have no good reason to support it. Even if we wanted to, we are maintaining compatibility with Go 1.5 which also disqualifies this. Sorry.

GodUseVPN:
Ok, I think Server-Initiated Renegotiation is useful.
For someone who has the same problem, `Renegotiation: tls.RenegotiateFreelyAsClient` should add to both announceClient and queryClient, I have missed queryClient in the commit.
In my case, discovsrv is deployed in a vm, behind transparent proxy.
haproxy(443) -->openvpn(1194)/apache(8443) --> virtualbox(8484) --> discovsrv(8443)

calmh:
Note that we run the global discovery servers in production between a nginx proxy and this works fine without renegotiation. Which boosts my opinion that this is unnecessary.

PR #2939, opened 2016-04-04
lib/ignore: Remove pattern for foo/** which is already covered by foo/

calmh:
### Purpose

Actual speed difference according to benchmarks is hidden in the noise

Also make the "pattern" field for each entry match what is actually
evaluated, or it looks like we have duplicates in the `expanded` JSON field.

### Testing

Those unit tests, they rock.

PR #3640, opened 2016-09-30
Match directory contents for patterns ending in /

### LINKED ISSUE #3155: Expected behavior of the ignore pattern "dir/"?
Syncthing Version: v0.13.2
OS Version: Arch Linux

I am not sure whether this is a bug or an intentional change, but I am reporting it anyways. In Syncthing 0.12 I discovered I could use
```
!/dir/file
/dir/
!/dir
```
to sync only ``/dir/file``. Now in 0.13 the files under ``/dir/`` do not seem to match the ``/dir/`` rule, and so I have to use the following instead:
```
!/dir/file
/dir/*
!/dir
```
The documentation is currently unclear about what would happen when a rule ends with ``/``.

### LINKED ISSUE #3639: Patterns with terminal slash (e.g., "dir/") do not ignore directory contents
If a pattern with a trailing slash is included in `.stignore`, it is expected to match the directory contents, but not the directory itself. Instead, the directory and contents are synced.

This seems to be due to the removal of these three lines: https://github.com/syncthing/syncthing/pull/2939/files#diff-a0ccca3b95e35f5785db8ae0886a570dL303
That commit says it removes pattern for `foo/**`, which is covered by `foo/`.
But it should be the other way around. `foo/` can be removed because it doesn't match anything (not the directory, which is matched by `foo`, and not the contents, which are matched by `foo/**`).

Related: #3155

### Version Information
Syncthing Version: v0.14.7
Ubuntu 16.04

timhowes:
Appends "**" to patterns with a terminal slash, so that directory
contents are ignored, but not the directory itself.

Addresses issue related to https://github.com/syncthing/syncthing/pull/2939
Closes #3639, resolves #3155

### DIFF
--- a/lib/ignore/ignore.go
+++ b/lib/ignore/ignore.go
@@ -345,7 +345,7 @@ func parseIgnoreFile(fd io.Reader, currentFile string, seen map[string]bool) ([]
 		case strings.HasSuffix(line, "/**"):
 			err = addPattern(line)
 		case strings.HasSuffix(line, "/"):
-			err = addPattern(line)
+			err = addPattern(line + "**")
 		default:
 			err = addPattern(line)
 			if err == nil {
--- a/test/ignore_test.go
+++ b/test/ignore_test.go
@@ -130,7 +130,37 @@ func TestIgnores(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	expected = len(all)*7/8 + 1 // seven out of eight items of each type should remain, plus the foo.TXT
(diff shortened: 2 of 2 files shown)


calmh:
Could you add a test to verify this, so we don't regress on this point again in the future?

calmh:
@st-jenkins add to white list

timhowes:
OK, I added an additional section to `TestIgnores` that checks that ignoring `d1/` ignores the file `d1/f1.TXT` but not `d1` itself.

calmh:
I meant a unit test. :) Merged, with that tweaked. Thanks!

PR #3650, opened 2016-10-04
Update to accept relay clients listening on all interfaces

damajor:
Update to accept relay clients listening on all interfaces  …

Listening on all interfaces is the default behavior of running the relay server without any parameters.
This commit will allow those clients to connect instead of rejecting them with a HTTP-401.

### DIFF
--- a/cmd/strelaypoolsrv/main.go
+++ b/cmd/strelaypoolsrv/main.go
@@ -326,7 +326,7 @@ func handlePostRequest(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	ip = net.ParseIP(host)
+	ip := net.ParseIP(host)
 	// The client did not provide an IP address, use the IP address of the client.
 	if ip == nil || ip.IsUnspecified() {
 		uri.Host = net.JoinHostPort(rhost, port)
--- a/cmd/strelaypoolsrv/main.go
+++ b/cmd/strelaypoolsrv/main.go
@@ -287,6 +287,7 @@ func handleGetRequest(w http.ResponseWriter, r *http.Request) {
 
 func handlePostRequest(w http.ResponseWriter, r *http.Request) {
 	var newRelay relay
+	var ip net.IP
 	err := json.NewDecoder(r.Body).Decode(&newRelay)
 	r.Body.Close()
 
@@ -326,8 +327,9 @@ func handlePostRequest(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	ip = net.ParseIP(host)
 	// The client did not provide an IP address, use the IP address of the client.
-	if host == "" || host == "0.0.0.0" {
+	if host == "" || ip.IsUnspecified() {
 		uri.Host = net.JoinHostPort(rhost, port)
 		newRelay.URL = uri.String()
 	} else if host != rhost {
--- a/cmd/strelaypoolsrv/main.go
+++ b/cmd/strelaypoolsrv/main.go
@@ -327,7 +327,7 @@ func handlePostRequest(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// The client did not provide an IP address, use the IP address of the client.
-	if host == "" {
+	if host == "" && host == "0.0.0.0" {
 		uri.Host = net.JoinHostPort(rhost, port)
 		newRelay.URL = uri.String()
 	} else if host != rhost {
--- a/cmd/strelaypoolsrv/main.go
+++ b/cmd/strelaypoolsrv/main.go
@@ -327,7 +327,7 @@ func handlePostRequest(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// The client did not provide an IP address, use the IP address of the client.
-	if host == "" && host == "0.0.0.0" {
+	if host == "" || host == "0.0.0.0" {
 		uri.Host = net.JoinHostPort(rhost, port)
 		newRelay.URL = uri.String()
 	} else if host != rhost {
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
No, this actually makes it worse and prevents all clients from joining the pool.

damajor:
I just have tested again and you are right. I just write it to quickly. Now it should work as intended.

AudriusButkevicius:
That still doesn't answer my second question.

damajor:
I explained why it is a problem on the discussion link. It is because when
you do not provide explicit IP then the relay send 0.0.0.0, not an empty
string.

AudriusButkevicius:
If it doesn't then it's a problem, as it should send '' or ':1234' indicating just a port.

damajor:
If you inspect the relay server source code the address sent seems good to me (0.0.0.0 is good when listening on every interfaces), but it has nothing to do with an empty string.

AudriusButkevicius:
https://github.com/syncthing/syncthing/blob/master/cmd/strelaysrv/main.go#L91

The default on the relaysrv is empty string, so I still don't underatand why this is an issue in the default case.

damajor:
If you look closely to the code that send the connection at
https://github.com/syncthing/syncthing/blob/master/cmd/strelaysrv/pool.go#L27
You can see that what is transmitted is a full URI like
"relay://X.X.X.X:PORT/?id=INWONPC-E66F7IQ-3VJ4BJ2-EHKRL3G-ZBQ7TVO-L2333X2-MIOCCLK-BWFQ6QR&pingInterval=1m0s&networkTimeout=2m0s&sessionLimitBps=0&globalLimitBps=0&statusAddr=&providedBy="

This string is coming from
https://github.com/syncthing/syncthing/blob/master/cmd/strelaysrv/main.go#L213

Just follow how the URI is build and from which variables it is build and
you'll get it.

damajor:
I am not able to test it right now, but I am pretty sure that this call
return 0.0.0.0
https://github.com/syncthing/syncthing/blob/master/cmd/strelaysrv/main.go#L118
This good to me and behave correctly (IP cannot be blank in a URI afaik).

As stated by "Jakob Borg"  in
https://forum.syncthing.net/t/how-global-announce-works/6401 unspecified IP
address can be 0.0.0.0 and this is handled at server side.

canton7:
> Just follow how the URI is build and from which variables it is build and you'll get it.

Top tip: if you want your changes to be merged, don't come here with an attitude. Assume the maintainers have zero free time, and spell everything out, step-by-step, so there can be no doubt about what the problem is and where it comes from. Provide a line-by-line analysis, provide the logs you captured which indicate where the issue is, spell *everything* out, and everyone's lives will be smoother.

damajor:
Hi Anthony,

As you can see I do my best to provide as many information as possible.
Just flaming about attitude is a bit out of scope.
Do you have something to say about the issue ?

AudriusButkevicius:
Ok, so the listen address gets mangled by the UPnP addition in relaysrv

AudriusButkevicius:
Can you add yourself to authors and nicks

damajor:
I'll do that as soon as I can this evening. Thanks.

AudriusButkevicius:
This could be made more robust tho. We could wrap host in net.ParseIP(host) and then check if ip == nil or ip.IsUnspecified()

damajor:
Good idea, I will look at that later today.

calmh:
+1 on the nil/unspecified stuff, there is no particular reason to nail ourselves down to IPv4 here.

-1 on the adding to authors, I'll do that before mergin (or Audrius could; the correct procedure is to add to AUTHORS in the correct format and run scripts/authors.go)

damajor:
Ok I will do it with the abstract method.
Regarding the authors, honestly it is not important to me as it is a little
contrib, maybe later if I do more contribs.

damajor:
I have updated the pull request. But I have not removed the test for an empty string as I dont know what is the point there.

cmd/strelaypoolsrv/main.go
```diff
@@ -287,6 +287,7 @@ func handleGetRequest(w http.ResponseWriter, r *http.Request) {
 
 func handlePostRequest(w http.ResponseWriter, r *http.Request) {
 	var newRelay relay
+	var ip net.IP
```
AudriusButkevicius: No need for this, just 

`ip := net.ParseIP(...)`

cmd/strelaypoolsrv/main.go
```diff
@@ -326,8 +327,9 @@ func handlePostRequest(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	ip = net.ParseIP(host)
 	// The client did not provide an IP address, use the IP address of the client.
-	if host == "" {
+	if host == "" || ip.IsUnspecified() {
```
AudriusButkevicius: Just `if ip == nil || ip.IsUnspecified()`

damajor:
Ok thanks for the comments, I have cleaned the code as mentioned.

AudriusButkevicius:
You need a colon with the equals to declare a new var on the stack

damajor:
Fixed again :)
Always learning new things is nice thanks for the explanation.

AudriusButkevicius:
@st-review merge

skip-check: authors

st-review:
:ok_hand: Merged as 96e8f94833fc3a023b998685b106fcb241f65101. Thanks, @damajor!

AudriusButkevicius:
my bad, should have amended subject, but life is hard and then we all die.

damajor:
No problem :)

AudriusButkevicius:
Do you have a name for the authors list?

damajor:
I do have a name :) Xavier O.

Issue #3651, opened 2016-10-05
Documentation in .deb package has HTML content in PDF files


fheinle:
Turns out, ``/usr/share/doc/syncthing/FAQ.pdf`` and ``/usr/share/doc/syncthing/Getting-Started.pdf`` only contain HTML content:
    
    $ cat FAQ.pdf 
    <html>
    <head><title>301 Moved Permanently</title></head>
    <body bgcolor="white">
    <center><h1>301 Moved Permanently</h1></center>
    <hr><center>nginx/1.10.0 (Ubuntu)</center>
    </body>
    </html>

    $ cat Getting-Started.pdf 
    <html>
    <head><title>301 Moved Permanently</title></head>
    <body bgcolor="white">
    <center><h1>301 Moved Permanently</h1></center>
    <hr><center>nginx/1.10.0 (Ubuntu)</center>
    </body>
    </html>

I expected to, well, find PDF files. 

### Version Information

```
LANG=C apt-cache policy syncthing
syncthing:
  Installed: 0.14.8
  Candidate: 0.14.8
  Version table:
 *** 0.14.8 500
        500 http://apt.syncthing.net syncthing/release amd64 Packages
        100 /var/lib/dpkg/status
     0.14.7 500
        500 http://apt.syncthing.net syncthing/release amd64 Packages
     0.14.6 500
        500 http://apt.syncthing.net syncthing/release amd64 Packages

```
Syncthing Version: v0.14.8
OS Version: ElementaryOS 0.4 Loki (Ubuntu 16.04) with syncthing repo added

MikeLund:
Thanks for noticing this! I assume the syncthing server config was recently changed to require https, so the old http urls would give that response. https://github.com/syncthing/syncthing/pull/3652 should be a quick fix for it.

PR #3652, opened 2016-10-05
Update docs to use https

MikeLund:
Simple change to resolve https://github.com/syncthing/syncthing/issues/3651

### DIFF
--- a/jenkins/common.bash
+++ b/jenkins/common.bash
@@ -38,8 +38,8 @@ function clean {
 function fetchExtra {
     echo Fetching extra resources
     mkdir extra
-    curl -s -o extra/Getting-Started.pdf http://docs.syncthing.net/pdf/Getting-Started.pdf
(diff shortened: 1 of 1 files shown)


wweich:
@st-review merge it

jenkins: use https when downloading docs (fixes #3651)

st-review:
:ok_hand: Merged as 1e54a3e801b2856e5a17e9e29570a190e3db45a1. Thanks, @MikeLund!

fheinle:
It works. As a side note: passing `-L` to curl will follow HTTP redirects to their new destination, allowing for more flexibility on your webserver's side.

wweich:
As we are all very security conscious, the servers will keep https (as long as there are free certs from Let's Encrypt or an alternative) and using https directly is imho the way to go for now.

canton7:
Sure, but I think that passing `-L` can only do good.

PR #3674, opened 2016-10-17
lib/ignore: Add directory separator to glob.Compile call

imsodin:
The current implementation of [gobwas/glob](https://github.com/gobwas/glob) in lib/ignore does not account for directory separators ('/'). Thus the distinction between '\*' and '\*\*' is meaningless. Currently both the patterns 'a\*b' and 'a\*\*b' match the path 'a/b'.

I confirmed the current behavior looking at the last file received entry in the GUI. But as I am only familiar with syncthing-inotify and it uses lib/ignore for its filtering as well, I tested this PR with debug information from syncthing-inotify. The test setup is trivial, just put something like the strings mentioned into .stignore.

### DIFF
--- a/lib/ignore/ignore.go
+++ b/lib/ignore/ignore.go
@@ -279,14 +279,14 @@ func parseIgnoreFile(fd io.Reader, currentFile string, seen map[string]bool) ([]
 		var err error
 		if strings.HasPrefix(line, "/") {
 			// Pattern is rooted in the current dir only
-			pattern.match, err = glob.Compile(line[1:])
+			pattern.match, err = glob.Compile(line[1:], '/')
 			if err != nil {
 				return fmt.Errorf("invalid pattern %q in ignore file (%v)", line, err)
 			}
 			patterns = append(patterns, pattern)
 		} else if strings.HasPrefix(line, "**/") {
 			// Add the pattern as is, and without **/ so it matches in current dir
-			pattern.match, err = glob.Compile(line)
+			pattern.match, err = glob.Compile(line, '/')
 			if err != nil {
 				return fmt.Errorf("invalid pattern %q in ignore file (%v)", line, err)
 			}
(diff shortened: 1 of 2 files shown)


calmh:
@st-jenkins add to white list

calmh:
LGTM but could we get a unit test for this as well to catch regressions in the future? Somewhere in lib/ignore/ignore_test.go should be appropriate

imsodin:
Like this?

AudriusButkevicius:
I wonder if the pattern separators are normalised somewhere along the way, so that this gets fixed on windows.

imsodin:
@AudriusButkevicius By normalised do you mean converted to slashs? If yes, this happens [here](https://github.com/syncthing/syncthing/blob/master/lib/ignore/ignore.go#L341).

AudriusButkevicius:
@st-review lgtm

st-review:

calmh:
How about this:

```go
func TestIssue3674(t *testing.T) {
	stignore := `
	a*b
	a**c
	`

	testcases := []struct {
		file    string
		matches bool
	}{
		{"ab", true},
		{"asdfb", true},
		{"ac", true},
		{"asdfc", true},
		{"as/db", false},
		{"as/dc", true},
	}

	pats := New(true)
	err := pats.Parse(bytes.NewBufferString(stignore), ".stignore")
	if err != nil {
		t.Fatal(err)
	}

	for _, tc := range testcases {
		res := pats.Match(tc.file).IsIgnored()
		if res != tc.matches {
			t.Errorf("Matches(%q) == %v, expected %v", tc.file, res, tc.matches)
		}
	}
}
```

Which fails now, and should hopefully pass with your patch:

```bash
jb@unu:~/s/g/s/s/l/ignore $ go test -v -run TestIssue3674
=== RUN   TestIssue3674
--- FAIL: TestIssue3674 (0.00s)
	ignore_test.go:768: Matches("as/db") == true, expected false
FAIL
FAIL	github.com/syncthing/syncthing/lib/ignore	0.006s
```

imsodin:
Changed it to calmh's test function and yes, it does pass.

calmh:
👍

calmh:

calmh:
@st-review lgtm

st-review:

calmh:
wat

calmh:
@st-review merge it then

st-review:

st-review:
:ok_hand: Merged as 7c37301c910dd3cf54a44a7ac677aef52f402e6d. Thanks, @imsodin!

PR #3681, opened 2016-10-21
gui: update translation

JayceChant:
### Purpose

With reference to the context and translation in other languages,
"Override changes"  means the folder changed on other device(s),
user choose to accept the changes,
and override to the local master folder.
"撤销改变" means "Revert changes",
is not a proper translation.
So I modify the translation,
"接受其他设备的修改" means
"accept changes from other devices",
which I think most close to the original meaning.

### Testing

gui : translation change only.


AudriusButkevicius:
Read the readme file in that directory.

PR #3687, opened 2016-10-26
Remove focus outline on active panel header (fixes #3684)

### LINKED ISSUE #3684: suggestion: remove standard webkit focus
As you can see in the picture attached at the bottom, Chrom(e|ium) has a standard outline for focused elements. I personally hate this outline and I think it would be good to simly set `outline: none` to buttons, especially because there is a custom box-shadow active effect and it's obvious that the section in the screenshot is currently active. 

EDIT: This only occurs in the light theme, I don't see any blue outline in the dark theme.

selection_999 557

Syncthing Version: v0.14.9
OS Version: Linux Mint (64 bit) (Ubuntu 16.04)
Browser Version: Chromium v52

uok:
Remove focus outline on active panel header (webkit browsers, e.g. Chrome)

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -50,6 +50,10 @@ ul+h5 {
     word-wrap:break-word;
 }
 
+.btn:focus {
+    outline: 0;
+}
+
 .text-monospace {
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
See original ticket.

Issue #1683, opened 2015-04-22
More detailed sync logs in the web gui


lovetox:
Hi there,

Something i wished for since im using syncthing, are data like:

which files did a node already sync and which not

right now i only know the last time the node was online and that it finished (?) at least one file.
thats not very much. 
last synced file is cute, but i dont see why it has a place in the gui when other more meaningful data doesnt.

i guess some of that information can be obtained through log files, but i hope its planned to move some  of the data like this into the gui

i tried to search for a similar thread but didnt find anything, sorry if that topic was already requested.

keep going with the good work.

AudriusButkevicius:
What exactly would you expect to see?

calmh:
I think this may be #1550...

calmh:
Which in turn was a duplicate of #1437...

lovetox:
@calmh the other tickets are mostly about errors, definitely also an issue, though i didnt see many errors until now :D

for example:

if you klick on some history button for a folder

you can see in rows, all the files that were synced with the date, time, size and finish.

so someone could see for example: the node was online synced these files but the most important one is still left etc.

or for example, that i could see, the node finished all the files, and i can delete the node from my syncthing cause that was all was needed.

AudriusButkevicius:
Well if the node ia offline, we don't even know what it has as it's offline hence we cannot make any assumptions on what he has and what portion of his files we have. Furthermore the files might have changed while he was offlinr, so the fact that we think we have everything might be a lie. We can only deduce this once connected, and at that point you have the out of sync list/status bar which gives you what you want.

lovetox:
im not talking about "status" in percent or something like this, i know this was discussed in other topics and it looks to be a problem.

im not a programmer, but i cant believe you need the other node online to determine what your own computer transfered. this information surley has to be already in some logs? 

date time, file x started transfering
date time, node x signaled file x is finished - move on to the next

this information is stored, and can be viewed, if the node is later online or not does not matter

i hope i expressed myself better.

calmh:
Currently, what the other devices do is not logged. Actually, not even what we do ourselves, though that could at least be done. For other devices though, we don't know what they did. Maybe the synced file A from us and file B from someone else. We could show file A as the latest file synced by them, but it could be a lie.

lovetox:
hm ok.

yeah i only want to see what my machine did.
if a node gets a file from another node than me, that of course should not show up in my log / or at least thats not what i need.
[...]

Issue #2924, opened 2016-04-01
Feature request: Show me the most recently synced files, both globally, and per share/device.


seansimonanimation:
Syncthing Version: v0.12.21
OS Version: Windows 7/WIndows 10
Browser Version: SyncTrayzor browser

I love the flexibility that Syncthing provides me. It allows me to manage files between my team members, and the freedom that nested sharing allows lets me set up my organization the way I want.

There is one bit of functionality that Dropbox provides that I would like to see: When you click on the Dropbox tray icon, you get a list of the 30 or so most recently updated files. I haven't seen anything like this in Syncthing, and I haven't seen it as a feature request here on the Github, but would it be possible to add in a row to remote devices and shared folders that allow me to see the most recently updated files, and when they were updated?

I feel like this is something that is already most of the way there, it's just a matter of logging it, and showing the log.

Adding it into the GUI would make it an even better feature.
Thanks :)

AudriusButkevicius:
It already shows the last file. 
Anyways, I think there already exists a ticket like this, but I am not in a position to find it over the phone.

exokkk:
I would support to see more than only the last file.

AudriusButkevicius:
If you want to support something, make a pull request.

rmarskell:
I noticed that when I single-click on the icon in my taskbar, it shows a recently updated file list. It would be great to have a full log somewhere, but this is a great start.

AudriusButkevicius:
Syncthing does not provide task bar icons, so its probably some wrapper you are using.

rmarskell:
Ah, touché. I found this on a Google search for SyncTrayzor and assumed it was the same repo. Anyone looking for this feature should check it out: https://github.com/canton7/SyncTrayzor

AudriusButkevicius:
Closed by #3694

PR #3694, opened 2016-10-29
GUI display of global changes

nrm21:
### Purpose

This PR adds a button to the device section of the GUI that the user can click on to see all filesystem changes that originated/occurred on that machine.  That is to say, not changed copied from the network.  This should help people discover more easily what computer made a certain change that was propagated in larger P2P swarms.
### Screenshots

[image](http://i.imgur.com/tGNhoCr.png)


gui/default/index.html
```diff
@@ -501,6 +501,15 @@ <h4 class="panel-title">
...
                   </tr>
+                  <!--<tr>
+                    <th><span class="fa fa-fw fa-exchange"></span>&nbsp;<span translate>Last Local Changes</span></th>
+                    <td class="text-right">
+                      <span tooltip data-original-title="{{localChangeLog}}">
+                      {{localChangeEvents[localChangeEvents.length-1].data.path}}
+                      </span>
+                      <a ng-href="localChangeLog()">{{localChangeEvents[localChangeEvents.length-1].data.path}}</a>
+                    </td>
+                  </tr>-->
```
AudriusButkevicius: Remove the dead code

gui/default/syncthing/core/syncthingController.js
```diff
@@ -629,6 +631,32 @@ angular.module('syncthing.core')
...
+                    }
+
+                    $scope.localChangeEventLog += "<tr>";
+                    $scope.localChangeEventLog += "<td>" + $scope.localChangeEvents[i].data.type + "</td>";
+                    $scope.localChangeEventLog += "<td>" + $scope.localChangeEvents[i].data.action + "</td>";
+                    $scope.localChangeEventLog += "<td>" + $scope.localChangeEvents[i].data.path + "</td>";
+                    $scope.localChangeEventLog += "<td>" + $scope.localChangeEvents[i].time.toLocaleString() + "</td>";
+                    $scope.localChangeEventLog += "</tr>";
+                }
+                $scope.localChangeEventLog += "</table>";
```
AudriusButkevicius: Please no HTML building in the code, that's why we have templates to start with

uok:
@nrm21 thanks for working on this - some ideas

**button ideas**
- button below panel like "add remote device" below remote devices
- instead of button an entry in top menu
- button with just icon or icon+text floated right in header of own device (e.g. Win10-VM)

**icon ideas**
- in my opinion icon `fa-history` is a better fit

**label ideas**
- "last changes" or "log"

calmh:
@AudriusButkevicius floated the idea in another ticket of tagging a file info with the device id of the one who last changed it. This seems like a simple and concise addition, we essentially just set that field in the scanner and it'll propagate everywhere. The LocalIndexUpdated event could then report which device originated the file that was synced. Would that fulfill this, or change this, or be relevant to this work somehow?

nrm21:
Yeah I was thinking about a change sorta like that myself as well a while back.  This change basically will just read from the [/rest/event/disk](https://github.com/syncthing/docs/blob/master/rest/events-disk-get.rst) API and display it to the user in the GUI.

So what you're mentioning would likely enhance and probably solve the global log problem permanently.  We could just add some code to append events to that event buffer as well but with type being RemoteChangeDetected instead or something like that.

gui/default/index.html
nrm21: Done.

gui/default/syncthing/core/syncthingController.js
nrm21: Done.

nrm21:
@uok I moved the button to the bottom and gave it the icon you like.  :)

nrm21:
Also guys I updated the image link in the OP to show the way it looks now.

uok:
Thanks! Please do not use "change log" as it is usually related to software releases (and also already exists in the bottom menu for that reason). I think "last changes" is just fine,

How many more "steps" than 10 back can changes be tracked?

nrm21:
I assume the answer is... as many back as that nodes syncthing has been online for.  I actually changed it to show the last 20 changes (not 10 anymore).  But that's just so it will fit on screen and not spend enormous amounts of CPU showing many many changes.

I just ran the API on my main client at home (where my ST has been online for weeks) and although my screen buffer only has a few thousand lines my event "id" is currently at approx 31k so I'm hoping that's enough events for your liking.  :)

gui/default/syncthing/device/changeLogModalView.html
```diff
@@ -0,0 +1,24 @@
...
+        <table>
+          <tr>
+            <th translate>Type</th>
+            <th translate>Action</th>
+            <th translate>Path</th>
+            <th translate>Time</th>
+          </tr>
+          <tr ng-repeat="ChangeLog in localChangeEvents">
+            <td translate>&nbsp;{{ChangeLog.data.type}}&nbsp;</td>
+            <td translate>&nbsp;{{ChangeLog.data.action}}&nbsp;</td>
```
AudriusButkevicius: This can't be translated as it's a variable.

AudriusButkevicius: Use CSS rather than whitespaces from both ends.

gui/default/syncthing/device/changeLogModalView.html
```diff
@@ -0,0 +1,24 @@
...
+<modal id="localDiskChanges" status="default" icon="{{'history'}}" heading="{{'Last Changes' | translate}}" large="yes" closeable="yes">
+    <div class="modal-body">
+        <table>
+          <tr>
+            <th translate>Type</th>
+            <th translate>Action</th>
+            <th translate>Path</th>
+            <th translate>Time</th>
+          </tr>
+          <tr ng-repeat="ChangeLog in localChangeEvents">
```
AudriusButkevicius: Incorrect casing + variable name

nrm21:
So I'm wondering if I should just also go ahead and add that field to the fileinfo and tack that feature onto this before pulling.  What do you think?  Older versions of the software will just ignore the field anyway right?

nrm21:
Ok I did it anyway... still not working yet but almost there.  Basically a local node shows all changes local and remote but doesn't see the Devicename of the remote (only his own) and vice versa on the other node.

Sooo... close !!   :)

lib/model/model.go
```diff
@@ -1467,18 +1467,32 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
 }
 
 func (m *Model) updateLocalsFromScanning(folder string, fs []protocol.FileInfo) {
+	// If we're running this code it's because our node last add/moved/changed
+	// this file.  So we set LastModBy to our devices friendly name
+	for i := range fs {
+		fs[i].LastModBy = m.deviceName
+	}
```
AudriusButkevicius: This needs to be done at the place where stuff comes back from the scanner.

lib/model/model.go
```diff
@@ -1467,18 +1467,32 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
 	m.updateLocals(folder, fs)
 
 	m.fmut.RLock()
 	folderCfg := m.folderCfgs[folder]
 	m.fmut.RUnlock()
 
 	// Fire the LocalChangeDetected event to notify listeners about local updates.
-	m.localChangeDetected(folderCfg, fs)
+	localChange := true
+	m.diskChangeDetected(folderCfg, fs, localChange)
```
AudriusButkevicius: No point allocating a variable if its only used once.

canton7: I think the intention is to document to the reader that the final parameter is 'localChange'? If you read `m.diskChangeDetected(folderCfg, fs, true)`, it's not clear what the `true` refers to.

lib/model/model.go
calmh: In walk.go, where the updated FileInfos are created, would be my preference.

lib/protocol/bep.pb.go
```diff
@@ -297,6 +297,7 @@ type FileInfo struct {
 	Permissions   uint32       `protobuf:"varint,4,opt,name=permissions,proto3" json:"permissions,omitempty"`
 	ModifiedS     int64        `protobuf:"varint,5,opt,name=modified_s,json=modifiedS,proto3" json:"modified_s,omitempty"`
 	ModifiedNs    int32        `protobuf:"varint,11,opt,name=modified_ns,json=modifiedNs,proto3" json:"modified_ns,omitempty"`
+	LastModBy     string       `protobuf:"bytes,12,opt,name=lastmodby,proto3" json:"lastmodby,omitempty"`
```
calmh: I don't see where the change is to bep.proto, but this should almost certainly be a `protocol.DeviceID` field, which is now possible natively since my last change. Don't forget the same change in lib/db/structs.proto

lib/model/model.go
nrm21: I actually only put it cause I thought you guys would want it for clarity.  I prefer brevity myself as long as it's documented somewhere, and I assume most programmers use an IDE to read code so they can just jump to the function declaration from that line to see what the variable is used for.

nrm21:
Not all changes we made yet... hold up for a day or so while I collapse from exhaustion for 7 hours then go to work then maybe come back and do more.  :P

lib/model/model.go
```diff
@@ -1549,21 +1546,21 @@ func (m *Model) diskChangeDetected(folderCfg config.FolderConfiguration, files [
...
 			})
 		} else {
 			events.Default.Log(events.RemoteChangeDetected, map[string]string{
-				"lastModBy": file.LastModBy,
-				"folderID":  folderCfg.ID,
-				"label":     folderCfg.Label,
-				"action":    action,
-				"type":      objType,
-				"path":      path,
+				"lastModifiedBy": file.LastModifiedBy.String(),
```
AudriusButkevicius: Put it at the bottom

lib/model/model.go
```diff
@@ -1546,21 +1546,21 @@ func (m *Model) diskChangeDetected(folderCfg config.FolderConfiguration, files [
 
 		if localChange {
 			events.Default.Log(events.LocalChangeDetected, map[string]string{
-				"lastModifiedBy": file.LastModifiedBy.String(),
 				"folderID":       folderCfg.ID,
 				"label":          folderCfg.Label,
 				"action":         action,
 				"type":           objType,
 				"path":           path,
+				"lastModifiedBy": file.LastModifiedBy.String(),
```
AudriusButkevicius: Also, I don't think you need to call string, it will get converted automatically.

lib/db/structs.proto
```diff
@@ -28,6 +28,7 @@ message FileInfoTruncated {
     uint32                permissions    = 4;
     int64                 modified_s     = 5;
     int32                 modified_ns    = 11;
+    protocol.DeviceID     lastmodifiedby = 12;
```
AudriusButkevicius: this should be last_modified_by probably

AudriusButkevicius: Or maybe we could cut it down to modified_by given we have modified_s and modified_ns

lib/model/model.go
nrm21: nope, compiler complains

nrm21:
Ok I made a few more changes including putting the code to tag modified_by in walk.go.  But it still is not showing on the other side (other side sees the shortID as AAAAAAA).  I may need some help finding exactly what needs to be done to copy that bit of info over the network.

AudriusButkevicius:
Did you recompile the protobuf file?

nrm21:
Is this a separate recompile than the normal one?   Right now I'm feeling safe going with "no I didn't" LOL!

Pardon my ignorance this is my first protobuf related code.  Any doc links on specifically what I should be doing there?  I'm just running "go run build.go" as per usual each time I recompile.  If there's more I need to be doing than that.  I may need some quick guidance.

AudriusButkevicius:
Can't recall from the top of my head/phone, but check the build file or run go generate in lib/protocol

nrm21:
Well it looks as if the correct command is "go run build.go proto" but when I run it (even after fresh reboot) I get:

2016/11/03 10:48:06 rename structs.proto.tmp structs.proto: The process cannot access the file because it is being used by another process.

AudriusButkevicius:
Check the script that runs when you run that command, I bet it's not closing the file it just read, it should be an easy fix somewhere in the scripts directory. And you are probably right, @calmh wrote it on most likely osx/unix so it probably hasn't been exercised on windows,

nrm21:
Well I'm nearly at my wits end with protobuf.  After getting around 5 or 6 errors I'm about ready to give up.  But this is what I finally get where I can't get past.

`2016/11/04 19:45:14 protoc-gen-go: error:no files to generate
lib\db\structs.go:8: running "protoc": exit status 1`
`2016/11/04 19:45:15 protoc-gen-go: error:no files to generate
lib\discover\local.go:8: running "protoc": exit status 1`
`2016/11/04 19:45:15 protoc-gen-go: error:no files to generate
lib\protocol\bep_extensions.go:4: running "protoc": exit status 1`


Any ideas?  How hard would it be to just compile, on Linux, the windows version of whatever files PB wants to compile and that I could then copy to my windows box to finish the compile?

How do you guys make the windows version when you compile?  On your unix/linux box?

AudriusButkevicius:
You need to install protobuf 3 compiler (protoc) from https://github.com/google/protobuf/releases/tag/v3.0.0 and have it somehwere along your PATH env var.

nrm21:
Well I did "go get -u github.com/golang
/protobuf/protoc-gen-go" instead but, it did install it... However I never got a protoc.exe file just a protoc-gen-go.exe file.  I put it in my path but now it's telling me there are no files to configure when I run.

AudriusButkevicius:
Can you paste the output here?

AudriusButkevicius:
I fixed a bug with the way the files were defined breaking on Windows, so you might need to rebase on master/merge master into your branch.

AudriusButkevicius:
https://github.com/syncthing/syncthing/commit/14937e7dd250c5bed1eb0274b7965e58faf97c65

nrm21:
Ok I rebased on yours.  Interestingly it seemed to fix the first error I would get where the files were "already in use" when renaming.  But I had previous fixed that where I jut ran it once... commented out all the code that opened/read the file and only left the "rename code" uncommented and reran the script.

But now I at least don't have to do that.  :)


However I still get this:

`2016/11/04 21:48:00 protoc-gen-go: error:no files to generate
lib\db\structs.go:8: running "protoc": exit status 1`
`2016/11/04 21:48:03 protoc-gen-go: error:no files to generate
lib\discover\local.go:8: running "protoc": exit status 1`
`2016/11/04 21:48:06 protoc-gen-go: error:no files to generate
lib\protocol\bep_extensions.go:4: running "protoc": exit status 1`
`exit status 1`
`exit status 1`

AudriusButkevicius:
In lib/protocol, run `protoc -I ../../../../../ -I ../../../../gogo/protobuf/protobuf -I . --gogofast_out=. bep.proto` and provide the output.

nrm21:
It's not returning from the command.

AudriusButkevicius:
well it might take some time, but it should return. If it's not returning there is not much I can suggest apart from googling around.

nrm21:
I will keep trying with different paths... anyway thanks for the help in trying.

I may even consider seeing if I can get the protobuf recompile on Linux and copy the needed files back over to my windows box.  That might work right?  Or is it somehow tying that file into the exe as it compiles it?

AudriusButkevicius:
The generated file should be the same across all platforms.

AudriusButkevicius:
I'd say modified by could be a full ID

nrm21:
Ok so... at the cost of much of my sanity, I resetup the whooooole environment in Linux and reinstalled go, git, protobuf, everything.  Also I had to fix a bajillion little errors caused by this thing not being in the right dir and that thing not being at ../../../../somewhere/gogogogoohgodwhyaretheresomany'go's/protobuf/protoc-gogowaitwasit2gosor3gos-gogofast

LOL, anyway...  I can now finally compile the protobuf files and ST.  But once I copy over these files:
lib/db/structs.pb.go
lib/protocol/bep.pb.go

Then I get this compile error:
lib\db\structs.pb.go:24:8: cannot find package "protocol" in any of ...<my GOPATH locations>

Ok so... I guess that's because I put "protocol.ShortID" in the protobuf file.  But shouldn't it be able to find that as a custom type in part of the lib?

AudriusButkevicius:
No, I don't think we have that. Check the vector definition, as that uses shortid. Also, not sure, maybe we should use full id. @calmh ?

nrm21:
Any reason why we need full?  It's MUCH shorter to display in GUI only 7 chars and it should still be a unique enough string.  I personally prefer the friendly name of the client myself since it is more human readable (for GUI purposes).  But I'll do it how ever you guys decide you want it.

AudriusButkevicius:
We can always opt-in to show the short one, and yeah short will work for most of the cases, and I guess there is no need to have the full one if it's modified by a device you are not aware off for example.

nrm21:
Ok now the errors I get are these when I copy in the recompiled .bp.go and .proto files:

C:\Users\Nate\Documents\src\github.com\syncthing\syncthing>go run build.go
github.com/syncthing/syncthing/lib/protocol
# github.com/syncthing/syncthing/lib/protocol
lib\protocol\bep.pb.go:49: undefined: proto.GoGoProtoPackageIsVersion2
lib\protocol\bep.pb.go:865: m.ModifiedBy.ProtoSize undefined (type *ShortID has no field or method ProtoSize)
lib\protocol\bep.pb.go:866: m.ModifiedBy.MarshalTo undefined (type *ShortID has no field or method MarshalTo)
lib\protocol\bep.pb.go:939: invalid operation: m.ModifiedBy != 0 (mismatched types *ShortID and int)
lib\protocol\bep.pb.go:942: cannot convert m.ModifiedBy (type *ShortID) to type uint64
lib\protocol\bep.pb.go:993: invalid operation: m.ID != 0 (mismatched types ShortID and int)
lib\protocol\bep.pb.go:996: cannot convert m.ID (type ShortID) to type uint64
lib\protocol\bep.pb.go:1438: m.ModifiedBy.ProtoSize undefined (type *ShortID has no field or method ProtoSize)
lib\protocol\bep.pb.go:1470: invalid operation: m.ModifiedBy != 0 (mismatched types *ShortID and int)
lib\protocol\deviceid.go:22: ShortID redeclared in this block
        previous declaration at lib\protocol\bep.pb.go:326
lib\protocol\bep.pb.go:1470: too many errors

AudriusButkevicius:
As I explained, shortid is not a custom type., Check the vector class, I think we use a int64 for storage of short id.

nrm21:
Ok I'll try again later when I get some time.  I thought it was custom because I saw somewhere in the same file another use of that field and it was coded as custom type.

nrm21:
@AudriusButkevicius This are the errors I have left after a clean redo of those files (took an older version of the proto files so I wasnt letting anything unecessary in them).  I have uploaded my code

lib\scanner\walk.go:328: cannot use w.Config.ShortID (type protocol.ShortID) as type uint64 in field value
lib\scanner\walk.go:364: cannot use w.Config.ShortID (type protocol.ShortID) as type uint64 in field value

how then should I deal with these shortID fields...try to typecast them to uint64 somehow?

nrm21:
@calmh @AudriusButkevicius  Ok so... I almost have this thing but I may need a bit of help.  I have it where it is correctly reporting who made a change to what file/folder for adds and modifies.  But for file/folder deletes it reports shortid 'AAAAAAA' (which I suspect is the default when the field is NULL).  I think this is because walk.go is not running through my code on discovery that a file is not there, only one when a new file IS that we didn't have before.

My code is within walkRegular() and walkDir(), which is within walkAndHashFiles() which is within walk() in the walk.go file.  Where would you suggest I add code for the identification of who deleted files that could be displayed in GUI?  My thoguhts are within the updateGlobal() function in the leveldb_transactions.go file since that also seems to run for both adds and deletes.  But I'm open to better suggestions.

AudriusButkevicius:
Deletions are handled outside of the scanner/walker, in the model. Try looking where it's handling the changes from the scanner.

nrm21:
When you say where handling changes from tehe scanner do you mean in scanner.go?

My methodology FYI:
I'm basically doing STTRACE=all... run program... note current time (to the second)... delete file/rescan all... kill program... look through trace log for everywhere it runs.

AudriusButkevicius:
These two places:
https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L1952
https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L1793

nrm21:
Thank you, I believe that might have been the piece I was missing.  I will do some more testing and let you know later today.

AudriusButkevicius:
Actually, the override case probably needs setting the last modified by in both cases of the if.

nrm21:
Not sure I understand why that function even needs my code at all.  The only place my code is needed is when we have detected our own copy on disk of a unique file.  Then we add our ShortID to the tag in prep that protobufs will send it off to others.

So if this override function is for files we need, wouldn't we would want to leave the tag alone here?  Trusting that some other client would make the necessary tag change since they were the one to last create/modify/delete the file?

AudriusButkevicius:
Override puts the files back to some old state, which counts as a modification

nrm21:
@calmh @AudriusButkevicius  Ok this is finally ready for PR.

Final image:  http://i.imgur.com/wt7dwF7.png

canton7:
Small request: can the gui display device names if available, rather than short IDs?

AudriusButkevicius:
Yes, good idea

nrm21:
I can try.

AudriusButkevicius:
No, but each ID has a ShortID() function I think.

canton7:
[`$scope.findDevice()`](https://github.com/syncthing/syncthing/blob/master/gui/default/syncthing/core/syncthingController.js#L915) or `$scope.deviceName()` (just below) look promising?

I wouldn't bother putting the device name in the stuff reported by the API - just convert it in the GUI. I think that's how the rest of the GUI does it...

AudriusButkevicius:
Yeah, in JS land it's easier just to do prefix matching.

nrm21:
FindDevice() requires ID, I'm just passing shortID, I could pass that instead of short.  Might even be better if that's not what we're going to be displaying anyway.

nrm21:
Ok so since I'm not that strong with Angular my question:

do I change this in html:
<td>{{localChangeEvent.data.modifiedBy}}</td>

To something like:
<td>{{ $scope.findDevice(localChangeEvent.data.modifiedBy) }}</td>

AudriusButkevicius:
Just `findName(xyz)` and then define $scope.findName somewhere in the JS land.

nrm21:
Ok done.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1263,6 +1275,10 @@ angular.module('syncthing.core')
             }
         }
 
+        $scope.localDiskChanges = function () {
+            $('#localDiskChanges').modal();
+        }
```
AudriusButkevicius: the local in the names is now misleading

nrm21: your right... I'll fix it.

lib/protocol/bep.proto
```diff
@@ -100,6 +100,7 @@ message FileInfo {
     uint32       permissions    = 4;
     int64        modified_s     = 5;
     int32        modified_ns    = 11;
+    string       modified_by    = 13;
```
nrm21: Well I don't know if I can do that.  I had short ID before but you guys wanted friendly name and that means I need the full ID.   I tried using []byte with protobuf but it was giving me wierd errors so I had to compromise with it and just go with string (of full ID).

AudriusButkevicius: We want short id here for the sake of saving database space.
Short ID is essentially first 5 characters of long ID, so you can look up device name by matching short ID as prefix to each devices lond ID.

nrm21:
Where is @calmh anyway?  I see him committing recently but haven't heard from him in a while on this thread.

AudriusButkevicius:
Yeah he seems to have become selective on answering ;)

gui/default/syncthing/device/changeLogModalView.html
```diff
@@ -0,0 +1,26 @@
...
+        <table>
+          <tr>
+            <th translate>Device</th>
+            <th translate>Action</th>
+            <th translate>Type</th>
+            <th translate>Path</th>
+            <th translate>Time</th>
+          </tr>
+          <tr ng-repeat="changeEvent in globalChangeEvents">
+            <td>{{deviceName(findDevice(changeEvent.data.modifiedBy))}}</td>
```
nrm21: Ok so I just changed this back to short ID and it broke predictably.  This line here needs the full device ID and not a short ID.

I want this to work in a way that satisfies you so I can get this PR'ed.  Any chance you can help me find the function I'm looking for in javascript to search for the full ID using the short?  I'm looking here but coming up blank.

AudriusButkevicius: There is nome, you have to write one that does prefix matching.

canton7: (Loop through all known devices. For each one, look at the device ID. If the device ID starts with your short ID, return its name. If none are found, return the short ID. For bonus points, see if more than one device ID has the same prefix, and do something sensible in that case).

lib/protocol/bep.proto
calmh: Yeah I was thinking about this too, and was first thinking full ID, but short ID is enough and has advantages.

(We try to guarantee that short IDs are unique in a cluster, but there are some corner cases where this may not be the case. In those cases our versioning is also broken however, so we should not worry about that for this specific case...)

calmh:
Working hard and traveling lots in weird time zones, and optimizing my engagement levels or something... :)

gui/default/syncthing/device/changeLogModalView.html
calmh: The config wrapper should be able to provide a `DeviceFromShort(id ShortID) DeviceConfiguration` or similar based on an internal lookup table. The short ID to long ID mapping is guaranteed to be unique per config.

nrm21:
Ok should be done now, give it a look and let me know.

nrm21:
For the check not sucessful...
I'm not exactly sure what that failure in model_test.go is.  I never even touched that file.

calmh:
That failure isn't your fault, and fixed in master. Don't worry about it.

calmh:
Please don't squash commits, it makes it much harder to review as we now need to go through the whole thing again rather than just look at the latest changes. No worries for this time though. :)

lib/db/structs.proto
```diff
@@ -28,6 +28,7 @@ message FileInfoTruncated {
     uint32                permissions    = 4;
     int64                 modified_s     = 5;
     int32                 modified_ns    = 11;
+    uint64                modified_by    = 12;
```
calmh: This, and in bep.proto, should be

``` 
uint64       modified_by    = 12 [(gogoproto.customtype) = "ShortID", (gogoproto.nullable) = false];
```

to get the correct type on the Go side of things.

lib/model/model.go
```diff
@@ -1559,18 +1559,38 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
 	return maxSequence, err
 }
 
+func (m *Model) makeUsLastModified(fs []protocol.FileInfo) []protocol.FileInfo {
+	fsMod := make([]protocol.FileInfo, 0, len(fs))
+	for _, file := range fs {
+		file.ModifiedBy = uint64(m.id.Short())
```
calmh: I think it would be cleaner to set this where we create the new FileInfo in walk.go instead?

lib/db/structs.proto
nrm21: Are we sure we want this nullable = false?  Don't we want to allow the field to be blank for reverse compatibility or something?

lib/model/model.go
nrm21: I tried that at first.  Unfortunately it only worked for object adds and modifies.

When I put the code in the location you want it at, then when I delete an object it would report the name of the last modified client as the one who deleted instead of the actual deleter.  I suppose the deletion code doesn't go through walk.go so I tried finding somewhere else I could add this hook also for deletes but never found a clean spot.

Perhaps you know of somewhere else outside walk.go?

nrm21:
Sorry about the squashing, I'll keep that in mind next time.

nrm21:
So you don't want me to squash but is it ok the rebase on master every so often if my branch gets too stale?

lib/db/structs.proto
calmh: Nullable false is the default anyway for uint64 and is compatible, it's just that witht he custom type the Go protobuf compiler would want to make it a *ShortID otherwise which we don't want. So yeah, nullable=false.

calmh:
Ideally, merge master into the pull request branch. That'll save you work over rebasing it, as well. But if it's just a couple of commits, rebasing is fine too.

lib/model/model.go
calmh: The corresponding place for deletes is here:

https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L1873

We should probably have methods on FileInfo for doing this instead, i.e.,

```
func (f FileInfo) Deleted(by ShortID) FileInfo
func (f FileInfo) Updated(by, newBlocks, newSize, newModtime etc)
```

Future improvement!

nrm21:
Ok check it out now.  Sorry for the week delay (if you were waiting) it's the holidays here in America.

nrm21:
@calmh @AudriusButkevicius  Any word on this?  Ready for pull?  Other probs?

gui/default/syncthing/core/syncthingController.js
```diff
@@ -51,6 +51,9 @@ angular.module('syncthing.core')
         $scope.failedPageSize = 10;
         $scope.scanProgress = {};
         $scope.themes = [];
+        $scope.globalChangeEvents = {};
+        $scope.findDevice = {};
+        $scope.deviceName = {};
```
AudriusButkevicius: Don't see the two vars used.

build.go
```diff
@@ -541,10 +541,20 @@ func buildSnap(target target) {
...
+	if matched, _ := regexp.MatchString(`^\d+\.\d+\.\d+$`, snapver); matched {
+		snapgrade = "stable"
+	}
 	err = tmpl.Execute(f, map[string]string{
-		"Version": version,
-		"Architecture": snaparch})
+		"Version":      snapver,
+		"Architecture": snaparch,
+		"Grade":        snapgrade,
+	})
```
AudriusButkevicius: Not sure why this is here

AudriusButkevicius:
Also rebase or bmerge master into this.

gui/default/syncthing/core/syncthingController.js
nrm21: I'll have to look into this.  I believe I just need to delete those as I decided not to use them anyway.

build.go
AudriusButkevicius: I think something went wrong when you merged some remote branch in, as that was proposed in some PR but then got removed I think, so you should remove it.

The proto compiler might be going nuts as you're using the wrong version of it?

nrm21:
Mistake

nrm21:
@calmh @AudriusButkevicius  Not to be a continual pest but it seems if I don't "bump the thread" then my PR gets forgotten.  :)

Are we good for merge yet?

gui/default/syncthing/core/syncthingController.js
```diff
@@ -912,6 +922,16 @@ angular.module('syncthing.core')
             return '';
         };
 
+        $scope.findDevicefromShort = function (shortID) {
+            var matches = $scope.devices.filter(function (n) {
```
AudriusButkevicius: From

AudriusButkevicius:
So apart from the capitalization, lgtm

lib/model/model.go
```diff
@@ -1602,13 +1608,25 @@ func (m *Model) localChangeDetected(folderCfg config.FolderConfiguration, files
...
 
-		events.Default.Log(events.LocalChangeDetected, map[string]string{
-			"folderID": folderCfg.ID,
-			"label":    folderCfg.Label,
-			"action":   action,
-			"type":     objType,
-			"path":     path,
-		})
+		if localChange {
+			events.Default.Log(events.LocalChangeDetected, map[string]string{
```
calmh: These appear identical, perhaps just make the event type a variable instead?

calmh:
@st-jenkins test this please what are you waiting for come on

nrm21:
There should be a screenshot at the very top of the thread.  I've updated it once or twice since I first dev'ed the feature.  Is there something specific you would like seen screenshotted?

AudriusButkevicius:
The screenshot is old, when there did not exist who made the change etc.

Question/Discussion point:
We have a buffered subscription of 1000, if we are only showing last 20.
I think wither we should show last all available using pagination, or perhaps reduce the buffering?

nrm21:
Ahh yes your right... and in fact that's the oldest screenshot, i have a newer one halfway down in the thread but that's also old already.  I'll upload another and fix these issues when I get a chance.


Buffered Sub 1000:
I can lower the buffered sub to 20 but my thoughts when making it were that the buffer of 1000 is for the API functionality.  That way anyone who wants to later, can make their own application (Nutomic's Android ST or something other perhaps?) that consumes the API and can use as many as they want of those events for their purpose.  I just happen, right now, to be the only consumer of that API func and I choose to use/show only the last 20.  So even though it's the same PR, I sort of see the API functionality and the GUI functionality as separate things.  Thats what an API is for anyway right, for others to share in the functionality of your code?  :)

AudriusButkevicius:
Actually, can you address the two nitpicks, get a screenshot and lets get this merged. We can iterate on other things in a separate PR if you ever feel bored.

nrm21:
[Screenshot](http://i.imgur.com/gs1QS97.png)

Read for merge.

nrm21:
Ok sorry I had to clobber my entire branch go back rebase and resolve some conflicts.

Apparently someone added something to protobufs recently and it wasn't cleanly merging with my code.  Sorry about that...

Ok NOW it should be ready for merge.

AudriusButkevicius:
Sorry but I'll make a few last nitpicks, as I've spotted them just now.

gui/default/index.html
```diff
@@ -613,9 +613,14 @@ <h4 class="panel-title">
           </div>
         </div>
         <div class="form-group">
-          <button type="button" class="btn btn-sm btn-default pull-right" ng-click="addDevice()">
-            <span class="fa fa-plus"></span>&nbsp;<span translate>Add Remote Device</span>
-          </button>
+          <span class="pull-right">
+            <button type="button" class="btn btn-sm btn-default" ng-click="globalDiskChanges()">
```
AudriusButkevicius: We could drop the word `Disk` as I don't think it brings any value.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -186,6 +187,7 @@ angular.module('syncthing.core')
 
         $scope.$on(Events.LOCAL_INDEX_UPDATED, function (event, arg) {
             refreshFolderStats();
+            refreshLastLocalChanges();
```
AudriusButkevicius: Not sure why this is local given the whole thing is called global.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -629,6 +631,14 @@ angular.module('syncthing.core')
             }).error($scope.emitHTTPError);
         }, 2500);
 
+        var refreshLastLocalChanges = debounce(function () {
+            $http.get(urlbase + "/events/disk?limit=20").success(function (data) {
+                $scope.globalChangeEvents = data;
```
AudriusButkevicius: You might want to store them (or display) in reverse order, as then newest change will be on top.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -629,6 +631,14 @@ angular.module('syncthing.core')
...
+
+                console.log("refreshLastLocalChanges", data);
```
AudriusButkevicius: Again Local yet it's global.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1270,6 +1290,10 @@ angular.module('syncthing.core')
             }
         }
 
+        $scope.globalDiskChanges = function () {
+            $('#globalDiskChanges').modal();
+        }
```
AudriusButkevicius: I know the above line is already inconsistent (looking at the above function), but can we have a `;` at the end of the function declaration.

gui/default/syncthing/device/changeLogModalView.html
```diff
@@ -0,0 +1,26 @@
...
+    <div class="modal-body">
+        <table>
+          <tr>
+            <th translate>Device</th>
+            <th translate>Action</th>
+            <th translate>Type</th>
+            <th translate>Path</th>
+            <th translate>Time</th>
+          </tr>
+          <tr ng-repeat="changeEvent in globalChangeEvents">
```
AudriusButkevicius: As I said above, perhaps displaying these in reverse order would feel more natural.

lib/model/model.go
```diff
@@ -1602,12 +1608,18 @@ func (m *Model) localChangeDetected(folderCfg config.FolderConfiguration, files
...
-		events.Default.Log(events.LocalChangeDetected, map[string]string{
-			"folderID": folderCfg.ID,
-			"label":    folderCfg.Label,
-			"action":   action,
-			"type":     objType,
-			"path":     path,
+		typeOfEvent := events.RemoteChangeDetected
+		if localChange {
+			typeOfEvent = events.LocalChangeDetected
+		}
```
AudriusButkevicius: Given this boolean is used only to decide the event type, I feel that `diskChangeDetected` should just accept a event type instead of the boolean. 
This way you don't need to specify what the boolean means as it's obvious from the event value being passed.

gui/default/index.html
```diff
@@ -647,6 +652,7 @@ <h4 class="panel-title">
   <ng-include src="'syncthing/core/shutdownDialogView.html'"></ng-include>
   <ng-include src="'syncthing/device/idqrModalView.html'"></ng-include>
   <ng-include src="'syncthing/device/editDeviceModalView.html'"></ng-include>
+  <ng-include src="'syncthing/device/changeLogModalView.html'"></ng-include>
```
AudriusButkevicius: This is a bit inconsistent as well, globalDiskChanges vs LastLocalChagne vs changeLog

gui/default/index.html
AudriusButkevicius: Or call it changeLog like you call it elsewhere.

AudriusButkevicius: Yet changelog might be used for displaying syncthing version changes in the future, so not sure if using it would be clever. Activity Log?

gui/default/index.html
nrm21: What would you like it called?

AudriusButkevicius: I don't know, but something consistent. Maybe Activity Log as I suggested above?

gui/default/syncthing/core/syncthingController.js
nrm21: I think this is a personal preference, I like it the way it is.  Half will want it one way half the other.

gui/default/syncthing/core/syncthingController.js
nrm21: I'll just take the opportunity to fix them both.

gui/default/syncthing/device/changeLogModalView.html
nrm21: So I'm going to leave this as is for personal pref if you don't mind and maybe in future I can add a sort.

AudriusButkevicius:
Seems the device name display is broken:

nrm21:
Wierd... it seems working for me.  I did:

go run build.go assets
go run build.go

Are you using a custom GUI folder?

AudriusButkevicius:
No, but on the other end I am using a version that does not set modifiedBy, which results in 'AAAAA', which results in an empty display value, so I think the case of device not being there is not handled well.

nrm21:
Ahh perhaps then it's blank because findDeviceFromShort() cant find from AAAAA.

```
        $scope.findDeviceFromShort = function (shortID) {
            var matches = $scope.devices.filter(function (n) {
                return n.deviceID.substr(0, 7) === shortID;
            });
            if (matches.length !== 1) {
                return undefined;
            }
            return matches[0];
        };

...

        $scope.deviceName = function (deviceCfg) {
            if (typeof deviceCfg === 'undefined' || typeof deviceCfg.deviceID === 'undefined') {
                return "";
            }
            if (deviceCfg.name) {
                return deviceCfg.name;
            }
            return deviceCfg.deviceID.substr(0, 6);
        };
```

nrm21:
The real question is why wasn't it finding a device.  I'm thinking maybe your protobufs is missing the field perhaps?

AudriusButkevicius:
I think your findDeviceFromShort should return the name or the short id if it can't find anything by itself without tapping into the other function. And for AAAAA it should return the special "Unknown" value.

AudriusButkevicius:
Amd yes, the other side does not set modifiedBy as its an older version, but its a normal usecase that we should support.

AudriusButkevicius:
Actually a better case would be for ShortID not to return AAAAA (return empty string) when the field is unset, and then we just do `|| 'Unknown'` in the UI code. We already do that for normal long device ID, check there.

aviau:
Nice feature :+1:

nrm21:
@AudriusButkevicius I'm not exactly sure what you mean by "then we just do || 'Unknown' in the UI code. We already do that for normal long device ID, check there."  Can you explain where exactly?

deviceName() would have to be changed for this I think.

AudriusButkevicius:
Look for EmptyDeviceID which alters how empty device IDs are serialized. We should do the same for shirt device ids. Once we do that, and unset ModifiedBy field returns an empty string oppose to AAAAA (which is what you get it you do ShortDeviceID(0).String()) we can modify the template for UI code to fallback to showing 'unknown' oppose to an empty string.

nrm21:
Well if the whole point is to show 'Unknown' to user when its AAAA, why not just do if (n.deviceID == AAAAA) return "unknown".  That's MUCH simpler and less code change (thus less likely to breakage).

nrm21:
Re-re-re-re-re-re-rebase (what is this the 6th time now? I'm losing count)

nrm21:
Ok see if that works for you.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -912,6 +923,17 @@ angular.module('syncthing.core')
...
+            var matches = $scope.devices.filter(function (n) {
+                return n.deviceID.substr(0, 7) === shortID;
+            });
+            if (matches.length !== 1) {
+                n.deviceID.name = "Unknown";
+                return n;
+            }
+            return matches[0];
+        };
+
```
AudriusButkevicius: This is a bad idea, you are modifying attributes on a random object that is persisted, so this will affect other things.

I'd replace:
`{{deviceName(findDeviceFromShort(changeEvent.data.modifiedBy))}}`

With:
```
{{deviceNameFromShort(changeEvent.data.modifiedBy))}}
```

and then just do:

```
for (i= 0; i< a.length; i++) {
    if ($scope.devices[i].deviceID.substr(0, 7) == shortID) {
        return $scope.devices[i].name;
    }
}
return 'Unknown';
```

canton7: Won't that literal 'Unknown' make localization impossible?

AudriusButkevicius: Actually it should not be 'Unknown' it should be shortID, so we fall back to displaying the ID.

nrm21: Ok so you're saying "If we cant display friendly name, only then we display shortID" right?

But isn't that was it was doing before for you?  Only it didn't know the ID so it used 'AAAAAA' right?

AudriusButkevicius: No, because A <-> B <-> C, A will never know C's name. The way we solve the AAAAA case is we makeit return an empty string instead of AAAAA, in which case we display unknown.

Sure people will upgrade, but not everyone atthe same pace.

calmh: I think this is what we're talking about, right, combined with just handling the empty string in the GUI:

```diff
diff --git a/lib/protocol/deviceid.go b/lib/protocol/deviceid.go
index cab70e3..e2b5d15 100644
--- a/lib/protocol/deviceid.go
+++ b/lib/protocol/deviceid.go
@@ -88,6 +88,9 @@ func (n *DeviceID) MarshalText() ([]byte, error) {
 }
 
 func (s ShortID) String() string {
+       if s == 0 {
+               return ""
+       }
        var bs [8]byte
        binary.BigEndian.PutUint64(bs[:], uint64(s))
        return base32.StdEncoding.EncodeToString(bs[:])[:7]
```

gui/default/syncthing/core/syncthingController.js
```diff
@@ -923,14 +923,14 @@ angular.module('syncthing.core')
...
             var matches = $scope.devices.filter(function (n) {
                 return n.deviceID.substr(0, 7) === shortID;
             });
             if (matches.length !== 1) {
-                return undefined;
+                return "undefined";
             }
-            return matches[0];
+            return matches[0].name;
         };
```
AudriusButkevicius: No, modifiedBy in Go code is an integer equal to 0, which when serialized to JSON becomes AAAAA (and the snippet @calmh posted changes that to be an empty string instead, which should be part of this PR).

We said that we want to:

1. Display device name if possible
2. Display short ID (if showing device name is not possible as we are not directly connected to the device)
3. Show 'Unknown' if it's a modification by a legacy client which hasn't set the field (resulting in AAAAA now and empty string once you apply calmh's snippet)

AudriusButkevicius:
The unknown case should probably be handled like:

```
<td ng-if="changeEvent.data.modifiedBy">{{friendlyNameFromShort(changeEvent.data.modifiedBy)}}</td>
<td ng-if="!changeEvent.data.modifiedBy"><span translate>Unknown</span></td>
```

So that the Unknown string could be translated.

AudriusButkevicius:
Let me know if you need any more help with this. Given this is very close to being good to go, I can finish it if you don't have time.

nrm21:
I can finish most likely...⁣ and not ignoring, just been busy with life lately (job).  I might have time tonight after work to work on it.

Thx I will let you know if I need help.



On Dec 20, 2016, at 1:48 PM, Audrius Butkevicius <notifications@github.com> wrote:

Let me know if you need any more help with this. Given this is very close to being good to go, I can finish it if you don't have time.

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub, or mute the thread.

AudriusButkevicius:
Great, thanks.

nrm21:
Another rebase

gui/default/syncthing/core/syncthingController.js
```diff
@@ -912,6 +923,16 @@ angular.module('syncthing.core')
             return '';
         };
 
+        $scope.friendlyNameFromShort = function (shortID) {
+            var matches = $scope.devices.filter(function (n) {
+                return n.deviceID.substr(0, 7) === shortID;
+            });
+            if (matches.length !== 1) {
+                return undefined;
```
AudriusButkevicius: `return shortID` here.

So we return name if we have the device, short id if we don't, and unknown inf modified by is unset.

AudriusButkevicius:
@st-review merge

all: Add a global change list

st-review:
:ok_hand: Merged as 0725e3af382a186495a856fe111346019c71b1e0. Thanks, @nrm21!

AudriusButkevicius:
Do you know if this closes any issue?

nrm21:
I think it does but I will have to look it up.

AudriusButkevicius:
Thanks for your patience and a valuable feature.

nrm21:
No prob, sorry it took so long at the end.  My life gets busy in waves.  Sometimes I have lots of time to program and sometimes not.  Back around September when I started I had time... these days not so much.

Actually one thing I want to do in future (but no time right now) is to show per folder sync log.  But I wanna do it at the API level so that showing it in GUI is a cinch (by passing in a URL GET parameter for folder ID).

IIRC from looking at it for a few mins it was not that easy because the JSON "data" field is a blank interface.  So I cant just look inside it for a FolderID = whatever, since Go doesn't know until runtime what fields would be in that interface (could be anything I guess).  Maybe there's another easy way.  Gather all the events in buffer and then pluck out the ones we want maybe?  Probably that's the way.

Or maybe someone else could do that if they have time.

nrm21:
Partial fulfillment of #2924
Closes #1683 also maybe

nrm21:
Heh, beat me too it.   ;)

lib/protocol/bep.proto
```diff
@@ -100,6 +100,7 @@ message FileInfo {
     uint32       permissions    = 4;
     int64        modified_s     = 5;
     int32        modified_ns    = 11;
+    uint64       modified_by    = 12 [(gogoproto.customtype) = "ShortID", (gogoproto.nullable) = false];
```
calmh: This never made it into the documentation/spec. Someone mind taking care of that?

nrm21: Done.

calmh: Eh... Yeah. But it's not live on the web. Which is not your fault. I'll go fix it, sorry for the noise.

PR #3695, opened 2016-10-29
lib/protocol: Use DeviceID in protocol messages, with custom marshalling

calmh:
### Purpose

This makes the device ID a real type that can be used in the protobuf schema. That avoids the juggling back and forth from []byte in a bunch of places and simplifies the code.

### Testing

There is a new unit test with protobuf schema that verifies that the old serialization is identical to the new. I've also manually verified that the discovery messages look the same before and after.


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 0296c2368578234649626caef929194a415ddf17. Thanks, @calmh!

PR #3709, opened 2016-11-05
Keep GUI wording consistent

uok:
Keep GUI wording consistent


calmh:
Negatory on this I think as our vocabulary is "folder" for the thing that is shared and "directory" or "path" for the thing that organizes files on disk...

calmh:
But the whole line of text there is too long and we should have the UI present that info differently somehow... Maybe just in a tooltip or whatever (but I dislike tooltips for discoverability...)

uok:
ok, I hope that this is clear for all the translators. I suggested this change as I saw this word pop up on transifex as untranslated and I like to keep things simple, e.g. user shares a folder in syncthing that has x subfolders.

PR #3715, opened 2016-11-08
Sync files to disk when STFSYNC set (fixes #3711)

### LINKED ISSUE #3711: fsync files
1. New versions of files get synced to a computer by syncthing.
2. The computer crashes, before the changes were written to disk.
3. At the next start the filesystem shows the old versions of the files.
4. Syncthing detects that the files have changed.
5. The old versions of the files are synced to all other computers by syncthing. The new versions are lost on all computers.

Android seems to be very lazy at writing dirty buffer cache pages to disk. It only happens when asked explicitly by fsync/sync or free memory gets low.

Unrud:
Sync files to disk before commiting changes to the database, when the environment variable STFSYNC is set. It solves the problems described in #3711.
To improve performance the syncs are performed when doing database batch commits.
The database itself is not synced to disk.


calmh:
Envvars are typically for stuff that is either only for developers or need to take effect before reading ocnfig. Should this not be a global or possibly per-folder config instead?

calmh:
BTW this had no measurable performance impact on `STFSYNC=1 go test -v -tags "integration benchmark" -run ManyFiles` on my Mac w/ SSD. It may be worse on rotating disks, but we should check. If not, and if it's clean enough, I'm leaning towards defaulting to it on despite prior reservations against it.

calmh:
@st-jenkins add to white list

lib/config/folderconfiguration.go
```diff
@@ -84,6 +84,11 @@ func (f *FolderConfiguration) CreateMarker() error {
 			return err
 		}
 		fd.Close()
+		if os.Getenv("STFSYNC") != "" {
+			if syncErr := osutil.SyncDir(filepath.Dir(marker)); syncErr != nil {
```
calmh: The `syncErr`s should typically all be called `err`, otherwise I'm all confused wondering why it's different and if there's an `err` declared somewhere we must not shadow etc.

lib/model/rwfolder.go
```diff
@@ -1372,12 +1372,33 @@ func (f *rwFolder) dbUpdaterRoutine() {
...
+
 	handleBatch := func() {
 		found := false
 		var lastFile protocol.FileInfo
 
 		for _, job := range batch {
 			files = append(files, job.file)
+			if fsync {
+				// collect changed files
+				if job.jobType == dbUpdateHandleFile || job.jobType == dbUpdateShortcutFile {
```
calmh: These would probably be better as a switch?

lib/config/folderconfiguration.go
```diff
@@ -84,6 +84,11 @@ func (f *FolderConfiguration) CreateMarker() error {
...
+				l.Infoln("Syncing directory failed:", syncErr)
```
AudriusButkevicius: Call it fsyncing not to confuse people

cmd/syncthing/main.go
```diff
@@ -172,6 +172,8 @@ are mostly useful for developers. Use with care.
                    "minio" for the github.com/minio/sha256-simd implementation,
                    and blank (the default) for auto detection.
 
+ STFSYNC           Sync files to disk.
+
```
AudriusButkevicius: I'd rather this was a config option or always on.

lib/model/rwfolder.go
```diff
@@ -1372,12 +1372,33 @@ func (f *rwFolder) dbUpdaterRoutine() {
...
+					changedFiles = append(changedFiles, filepath.Join(f.dir, job.file.Name))
+				}
+				// collect changed dirs
+				if job.jobType == dbUpdateHandleDir {
+					changedDirs = append(changedDirs, filepath.Join(f.dir, job.file.Name))
+				} else if job.jobType != dbUpdateShortcutFile {
+					changedDirs = append(changedDirs,
+						filepath.Dir(filepath.Join(f.dir, job.file.Name)))
```
AudriusButkevicius: This can have thousands of dupes.

lib/model/rwfolder.go
```diff
@@ -1372,12 +1372,33 @@ func (f *rwFolder) dbUpdaterRoutine() {
...
+				}
```
AudriusButkevicius: The logic is convoluted, use a switch.

lib/model/rwfolder.go
```diff
@@ -1390,6 +1411,33 @@ func (f *rwFolder) dbUpdaterRoutine() {
...
 
+		if fsync {
+			// sync files to disk
+			sort.Strings(changedFiles)
+			var lastChangedFile string
+			for _, changedFile := range changedFiles {
+				if lastChangedFile != changedFile {
+					lastChangedFile = changedFile
+					if syncErr := osutil.SyncFile(changedFile); syncErr != nil {
+						l.Infoln("Syncing file failed:", syncErr)
```
AudriusButkevicius: Again fsyncig

Unrud:
> Should this not be a global or possibly per-folder config instead?

I made it a global option. Setting it for every folder separately seems tedious.

> BTW this had no measurable performance impact

MacOS doesn't flush the on-disk-cache when fsync is called. That might be the reason.

Why are you calling Sync() [here](https://github.com/syncthing/syncthing/blob/master/lib/osutil/osutil.go#L178)?

lib/model/rwfolder.go
Unrud: It can't be more than ``maxBatchSize``. I thought that it's faster to just add all and sort them once at the end.

Ferroin:
@calmh I just finished some testing on both a 7200RPM SATA3 HDD and a slow replicated storage cluster (about 20MB/s bulk write speed), both on 64-bit Linux, and it both cases the performance difference is negligible (it's single digit percentages of the time it takes to write the data in the first place).  i only got a handful of data points for each, so it's by no means an exhaustive test, but it's still better than nothing.

As far as defaulting to having it on, I think that's a good idea, and I kind of feel that having the ability to control it per folder would be nice (there are some folders I have that I really don't care about this, while there are others where I do).

lib/config/optionsconfiguration.go
```diff
@@ -41,6 +41,7 @@ type OptionsConfiguration struct {
 	OverwriteRemoteDevNames bool     `xml:"overwriteRemoteDeviceNamesOnConnect" json:"overwriteRemoteDeviceNamesOnConnect" default:"false"`
 	TempIndexMinBlocks      int      `xml:"tempIndexMinBlocks" json:"tempIndexMinBlocks" default:"10"`
 	UnackedNotificationIDs  []string `xml:"unackedNotificationID" json:"unackedNotificationIDs"`
+	Fsync                   bool     `xml:"fsync" json:"fsync" default:"false"`
```
AudriusButkevicius: I'd set this to true, unless @calmh disagrees?
Also, I am not sure if we even should do this, perhaps we should enable this by default, given it's nice a little price to pay in order to guarantee we don't corrupt someones files.

lib/osutil/sync.go
```diff
@@ -0,0 +1,15 @@
+package osutil
```
AudriusButkevicius: Needs proper headers.

lib/osutil/sync_unix.go
```diff
@@ -0,0 +1,11 @@
+// +build !windows
```
AudriusButkevicius: Headers.

lib/osutil/sync_windows.go
```diff
@@ -0,0 +1,14 @@
+// +build windows
```
AudriusButkevicius: Headers.

Yet perhaps it's easier just to share the implementation across all platforms, and do `runtime.GOOS == "windows"` in the sync dir case?

lib/model/rwfolder.go
```diff
@@ -1372,12 +1374,49 @@ func (f *rwFolder) dbUpdaterRoutine() {
...
+				case dbUpdateShortcutFile:
+					changedFiles = append(changedFiles, filepath.Join(f.dir, job.file.Name))
+				case dbUpdateHandleFile:
+					changedFiles = append(changedFiles, filepath.Join(f.dir, job.file.Name))
+					fallthrough
+				case dbUpdateDeleteDir:
+					fallthrough
+				case dbUpdateDeleteFile:
+					changedDirs = append(changedDirs,
+						filepath.Dir(filepath.Join(f.dir, job.file.Name)))
```
AudriusButkevicius: The line break is a bit ugly here.
I'd store the value in a variable or just leave one longer line.

Personally, if I do have to break brackets across multiple lines, I usually follow the same principle as with curly braces such as:

```
foo = append(
  bar_long_baz_nuff
)
```

argument being that nobody does:

```
if (x) {
    do_y() }
```

Yet that's just me.

lib/model/rwfolder.go
```diff
@@ -1372,12 +1374,49 @@ func (f *rwFolder) dbUpdaterRoutine() {
...
+	}
+
+	syncFilesOnce := func(files []string, syncFn func(string) error) {
+		sort.Strings(files)
+		var lastFile string
+		for _, file := range files {
+			if lastFile != file {
+				lastFile = file
+				if err := syncFn(file); err != nil {
+					l.Infoln("fsync failed:", err)
```
AudriusButkevicius: Also, printing the path might be nice.

lib/osutil/atomic.go
```diff
@@ -97,6 +102,8 @@ func (w *AtomicWriter) Close() error {
 		return err
 	}
 
+	SyncDir(filepath.Dir(w.next.Name()))
```
AudriusButkevicius: Could check the error I guess.

lib/config/folderconfiguration.go
```diff
@@ -84,6 +84,9 @@ func (f *FolderConfiguration) CreateMarker() error {
 			return err
 		}
 		fd.Close()
+		if syncErr := osutil.SyncDir(filepath.Dir(marker)); syncErr != nil {
+			l.Infoln("fsync failed:", syncErr)
```
AudriusButkevicius: Nice to have a path.

lib/model/rwfolder.go
Unrud: **gofmt** doesn't approve.

calmh:
Why not make this a folder level option instead? We can do it unconditionally for global things like saving the config and folder marker anyway.

Unrud:
> Why not make this a folder level option instead?

I thought that it's tedious to set it for every folder separately. I made it a folder level option now.

calmh:
The requirements might differ though, and I agree with Audrius above that this should default to "true" instead. However that's somewhat annoying to do for folders, because we don't apply defaults to those in the same way as in the options case (for reasons, but still). You'll need to:

- Update `$scope.addFolder` and `$scope.addFolderAndShare` in `syncthingController.js` to add the new field.
- Update the config version number in `config.go` and add a migration there from version 16 to version 17 that sets it to true for existing folders at upgrade time.

Sorry :)

Apart from that it looks mostly good now, I think my `s/syncErr/err/` recommendation in most places still applies though.

AudriusButkevicius:
I suggest we invert the option, disableFsync with default beinf false, and it works fine.

lib/model/rwfolder.go
AudriusButkevicius: This should still be addressed.

lib/config/folderconfiguration.go
```diff
@@ -84,6 +85,9 @@ func (f *FolderConfiguration) CreateMarker() error {
 			return err
 		}
 		fd.Close()
+		if err := osutil.SyncDir(filepath.Dir(marker)); err != nil {
+			l.Infoln("fsync failed:", err)
```
AudriusButkevicius: Potentially print the path too.

calmh:
I think this will be better. I have working, normal, defaults for folder and device configurations in a branch so this will make more sense down the road.

lib/model/rwfolder.go
```diff
@@ -1372,12 +1374,49 @@ func (f *rwFolder) dbUpdaterRoutine() {
...
+	if f.fsync {
+		changedFiles = make([]string, 0, maxBatchSize)
+		changedDirs = make([]string, 0, maxBatchSize)
+	}
+
+	syncFilesOnce := func(files []string, syncFn func(string) error) {
+		sort.Strings(files)
+		var lastFile string
+		for _, file := range files {
+			if lastFile != file {
```
calmh: Minor style nit, could reduce indentation here by

```go
if file == lastFile {
    continue
}
```

lib/model/rwfolder.go
```diff
@@ -1372,12 +1374,49 @@ func (f *rwFolder) dbUpdaterRoutine() {
...
 	handleBatch := func() {
 		found := false
 		var lastFile protocol.FileInfo
 
 		for _, job := range batch {
 			files = append(files, job.file)
+			if f.fsync {
+				// collect changed files and dirs
+				switch job.jobType {
+				case dbUpdateShortcutFile:
```
calmh: "shortcut file" means we just changed the metadata, i.e., timestamp or permissions, so there is no need to fsync the file. We might want to fsync the containing directory though if we're being pedanting about it, which I honestly don't think we need to be.

lib/model/rwfolder.go
```diff
@@ -1372,12 +1374,49 @@ func (f *rwFolder) dbUpdaterRoutine() {
...
+					changedFiles = append(changedFiles, filepath.Join(f.dir, job.file.Name))
+				case dbUpdateHandleFile:
+					changedFiles = append(changedFiles, filepath.Join(f.dir, job.file.Name))
+					fallthrough
+				case dbUpdateDeleteDir:
+					fallthrough
+				case dbUpdateDeleteFile:
```
calmh: This could be merged to a single

`case dbUpdateDeleteDir, dbUpdateDeleteFile:`

instead of the fallthrough.

lib/model/rwfolder.go
Unrud: Metadata is **not** stored in the directory entry. fsync of the file also syncs metadata.

The directory entry only contains the filename and a pointer to the file structure. Syncing the directory is only necessary after creating, renaming or removing a file.

Unrud: Symlinks are not explicitly handled. Only MacOS supports syncing the content (the path it points to). I left it out for simplicity but can add it if you want.

lib/model/rwfolder.go
```diff
@@ -1403,19 +1404,19 @@ func (f *rwFolder) dbUpdaterRoutine() {
...
-					changedFiles = append(changedFiles, filepath.Join(f.dir, job.file.Name))
-					fallthrough
-				case dbUpdateDeleteDir:
-					fallthrough
-				case dbUpdateDeleteFile:
-					changedDirs = append(changedDirs,
-						filepath.Dir(filepath.Join(f.dir, job.file.Name)))
+				case dbUpdateHandleFile, dbUpdateShortcutFile:
+					// syncing symlinks is only supported by MacOS
+					if !job.file.IsSymlink() {
```
AudriusButkevicius: This confused me a bit. You mean fsyncing symlinks, right?

Unrud: Yes

AudriusButkevicius:
@st-review lgtm

st-review:

calmh:
Unrud, do you have a name that you would like to add to the AUTHORS file?

Unrud:
> Unrud, do you have a name that you would like to add to the AUTHORS file?

No, I don't care.

calmh:
I do though, so it needs to be something. Is "Unrud" how you want to be named, or is there a last name or first name I can attach to that? I don't particularly care what on your drivers license or birth certificate, just that we have something with a name and an email address to attach the copyright to. :)

Unrud:
> Is "Unrud" how you want to be named

Yes.

> and an email address

unrud@openaliasbox.org

calmh:

calmh:
Or don't. Whatever. :(

calmh:
@st-review lgtm

lib/model: Add fsync of files and directories, option to disable (fixes #3711)

Skip-check: authors

calmh:
@st-review merge it!

lib/model: Add fsync of files and directories, option to disable (fixes #3711)

Skip-check: authors

calmh:
@st-review merge it!

lib/model: Add fsync of files and directories, option to disable (fixes #3711)

Skip-check: authors

calmh:
Looks like github webhooks are taking ages to fire at the moment. Fixing this up manually.

st-review:

st-review:
@calmh: There's already a merge pending for this PR.

st-review:
@calmh: There's already a merge pending for this PR.

st-review:
@calmh: There's already a merge pending for this PR.

st-review:

calmh:
Merged in 1574b7d

calmh:
Thanks Unrud!

st-review:
@calmh: Merge failed:

```
$ git fetch -f origin refs/pull/3715/head:pr-3715
   1392589..1574b7d  master     -> orig/master
   0d55d8c..1574b7d  master     -> origin/master
$ git reset --hard
HEAD is now at 0d55d8c gui: Convert URLs in warning messages to HTML links (fixes #3241)
$ git checkout master
Already on 'master'
Your branch is behind 'origin/master' by 2 commits, and can be fast-forwarded.
HEAD is now at 1574b7d lib/model: Add fsync of files and directories, option to disable (fixes #3711)
$ git clean -fxd
$ git merge --squash --no-commit pr-3715
$ git commit -F -
On branch master

nothing to commit, working directory clean

```

AudriusButkevicius:
This is very useful. Thank you for spending time on this.

PR #3717, opened 2016-11-08
Build the snaps for armhf and arm64

elopio:
### Purpose

Support building snaps for armhf and arm64.

### Testing

I built the snaps in a xenial lxc with:
go run build.go -goarch armhf snap
go run build.go -goarch arm64 snap
 
Then tested the armhf in an raspberrypi2 and arm64 in a dragonboard 410c, both with ubuntu core snappy images.

### Documentation

There are no docs for building debs, so I haven't written anything yet about building snaps. However, I think a new section should be added for this. I can do it in a follow up branch.

### DIFF
--- a/build.go
+++ b/build.go
@@ -539,8 +539,8 @@ func buildSnap(target target) {
 	}
 
 	snaparch := goarch
-	if goarch == "arm" {
-		snaparch = "armhf"
+	if snaparch == "armhf" {
+		goarch = "arm"
 	}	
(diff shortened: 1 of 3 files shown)


calmh:
@st-review merge it

build: Build snaps for ARM

st-review:
:ok_hand: Merged as afd15392b135c95644b713127ef0b63946a522bf. Thanks, @elopio!

elopio:
Thanks!

PR #3722, opened 2016-11-10
lib/config: Subscribers get a copy of the config

AudriusButkevicius:
As notifications are not under a lock, and the config might be modified as subscribers access it.
Plus, from the API perspective people might modify it locally for some reason, affecting the value others get.

I believe this solves the race in de-introduction stuff, so we can revert the revert.


calmh:
@st-review merge merge merge

st-review:
:ok_hand: Merged as 56e2ba29d0f973b285868131e8226885f72d79f5. Thanks, @AudriusButkevicius!

PR #3727, opened 2016-11-11
Fixes to relay stuff

AudriusButkevicius:
Probably best merged unsquashed.

### DIFF
--- a/cmd/strelaysrv/listener.go
+++ b/cmd/strelaysrv/listener.go
@@ -172,7 +172,7 @@ func protocolConnectionHandler(tcpConn net.Conn, config *tls.Config) {
 					if debug {
 						log.Println("Sent invitation from", id, "to", requestedPeer)
 					}
-				default:
+				case <-time.After(time.Second):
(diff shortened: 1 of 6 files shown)


cmd/strelaysrv/listener.go
```diff
@@ -259,8 +258,8 @@ func protocolConnectionHandler(tcpConn net.Conn, config *tls.Config) {
 			}
 			conn.Close()
 
-		case msg := <-outbox:
-			if msg == nil {
+		case msg, ok := <-outbox:
+			if !ok {
```
calmh: This will never happen if the outbox isn't closed?

calmh: Is the idea that the return always happens via the `<-errors` branch?

AudriusButkevicius: Yes

AudriusButkevicius:
Addressed comment.

calmh:
Merged

PR #3730, opened 2016-11-12
Set the sync home to SNAP_USER_COMMON

elopio:
### Purpose

Set the home to a path that's not versioned with snap updates.

### Testing

Installed in a kvm, and confirm that the sync path is now in $HOME/snap/syncthing/common

### Documentation

When we document the snap, we need to mention the sync path. I prefer to give a little testing on edge before documenting, because things are likely to change.

### DIFF
--- a/snapcraft.yaml.template
+++ b/snapcraft.yaml.template
@@ -13,7 +13,7 @@ confinement: strict
 
 apps:
   syncthing:
-    command: syncthing
+    command: syncthing -home $SNAP_USER_COMMON
     plugs: [home, network, network-bind]
 
 parts:
--- a/snapcraft.yaml.template
+++ b/snapcraft.yaml.template
@@ -13,7 +13,7 @@ confinement: strict
 
 apps:
   syncthing:
-    command: syncthing -home $SNAP_USER_COMMON
+    command: env HOME=$SNAP_USER_COMMON $SNAP/syncthing
     plugs: [home, network, network-bind]
 
 parts:


calmh:
Something similar seems needed, but leaves the user without the ability to control it?

elopio:
@calmh what about this?

The thing that I can't seem to get working is syncthing -paths.
If you start it like this:
$ syncthing -home $HOME/test
it will work. But then you do:
$ syncthing -paths, and it will not report that the paths are in the test directory, it will report $SNAP_USER_COMMON.

Maybe I'm just misunderstanding how -paths should work.

calmh:
That's expected, `-home` just overrides the one invocation, it's not persisted.

I would expect the right thing here to be setting `XDG_CONFIG_DIF=$SNAP_USER_COMMON`, as that's where we store the config and database.

Why is not `$HOME` set to the actual `$HOME` given that it has the `home` plug?

calmh:
With

```diff
--- a/snapcraft.yaml.template
+++ b/snapcraft.yaml.template
@@ -13,7 +13,7 @@ confinement: strict
 
 apps:
   syncthing:
-    command: syncthing
+    command: env XDG_CONFIG_HOME="$SNAP_USER_COMMON" "$SNAP/syncthing"
     plugs: [home, network, network-bind]
 
 parts:
```

we get:

```
jb@ubuntu:~/s/g/s/syncthing $ syncthing -paths
Configuration file:
	/home/jb/snap/syncthing/common/syncthing/config.xml

Database directory:
	/home/jb/snap/syncthing/common/syncthing/index-v0.14.0.db

Device private key & certificate files:
	/home/jb/snap/syncthing/common/syncthing/key.pem
	/home/jb/snap/syncthing/common/syncthing/cert.pem

HTTPS private key & certificate files:
	/home/jb/snap/syncthing/common/syncthing/https-key.pem
	/home/jb/snap/syncthing/common/syncthing/https-cert.pem

Log file:
	/home/jb/snap/syncthing/common/syncthing/syncthing.log

GUI override directory:
	/home/jb/snap/syncthing/common/syncthing/gui

Default sync folder directory:
	/home/jb/snap/syncthing/x4/Sync

jb@ubuntu:~/s/g/s/syncthing $ 
```

Which is fine, except the "default sync folder directory" is broken as we have no working `$HOME` apparently.

calmh:
Another question - how do we integrate with systemd or whatever properly using the snap?

calmh:
It looks like the snap format has a provision for `daemon`s, but syncthing would need to be either that or a user command, and `$HOME` gets set differently depending on how it's run. I'm starting to feel like a square peg in the round snap hole here...

elopio:
Hey @calmh.

Here's a nice trick:
$ snap run --shell syncthing  # This will open a shell in the same context where the snaps are executed.
$ env | grep HOME

The idea is to make every snap as isolated as possible, so they get their own home and tmp, that are not accessible by other snaps. Of course in a project like syncthing we can't follow the strictest rules, because it should be possible to sync your music directory, for example, which shouldn't be owned by any particular snap. So we allow snaps to get out of their confinement with interfaces. But then the home interface doesn't change the context where the snap runs, it just adds the additional permissions to read and write the user's home. And also note that the $HOME in the context of the snap doesn't alter the $HOME of the user, which is why `syncthing -home $HOME/test` leaves the files in /home/user/test, not in /home/user/snap/syncthing/[...].

Our approach to get started with interfaces is to make first a broad one, and then start limiting the permissions until you have only what's necessary. In the example from above, it doesn't make sense to let syncthing write to all your home if you only want to sync your Music directory. But we'll get there.

Anyway, back to -home, I didn't understand what you found wrong with
`env HOME=$SNAP_USER_COMMON $SNAP/syncthing`
Actually, after looking at your comment I thought this could be a good idea:
`env  XDG_CONFIG_HOME=$SNAP_USER_DATA HOME=$SNAP_USER_COMMON $SNAP/syncthing`
That way the sync directory is shared between revisions, but the database and config files are revision dependent which gives you the opportunity to modify the schemas on updates. By default, they will just be copied.

One more thing that we haven't used here, but might be useful is the snap config, which would work something like this: `snap set syncthing default_sync_dir=/home/elopio/sync`.

About the daemon, I thought that it would be easier to get the snap working first in user space before enabling it system wide. That will let you gather some stats to see if your users like it and if it's worth the effort. I'm not sure if you will agree here, we can take a different approach for sure. 

So all of this makes me thing that perhaps I'm following the wrong approach, specially because of your comment of the square peg :) Please note that we are reaching out first to upstreams which we consider important for the platform. We want you to help us define the right set of interfaces, and polish the current features, so if you feel that something doesn't fit your needs, this is the perfect time to get the right combination of interfaces for your project. Maybe it will be a better approach for you to tell us how you would like to see syncthing working on strict isolation, secure from the changes of other snaps, but making it easy for your users to choose whatever directory they want to put under your control.

This is our main mailing list: https://lists.ubuntu.com/mailman/listinfo/snapcraft
You'll see there that we are constantly explaining the parts that are confusing, making notes to better document them, and adding user requests to our backlog. I invite you to join, your feedback has been enormously useful already.

pura vida.

calmh:
The thing I don't like about the `$HOME` approach is essentially that `$HOME` and `~` suddenly means something entirely different that users are used to. E.g., check out this add folder dialog:

screen shot 2016-11-14 at 15 46 24

That's not my `~`, but Syncthing can't know that. In this case it makes less sense than usual as it points to the versioned directory, and we could fix that by pointing it to the common directory, but the user still can't share `~/Documents` by pointing Syncthing to `~/Documents` and that's going to be confusing and a support issue...

I honestly don't know how Syncthing would be usable in strict isolation, unless it's used as a write-only backup or something where the user doesn't need to interact with the files managed by Syncthing.

elopio:
I understand your problem. So you basically want to keep ~ to expand to the real user $HOME. I think that's a reasonable request, I'm going to raise this to the mailing list.

A quick hack that will get us there now would be:
command: env XDG_CONFIG_HOME=$SNAP_USER_COMMON HOME=/home/$USER $SNAP/syncthing.

Of course, that breaks if the user's home is in a different directory, like in /mnt or something, so it's not yet good material to release to stable. But IMO, it could be good enough to get further testing of the snap while the snap devs provide us with a better solution. What do you think?

About the vim example, I think it is a different case. I wouldn't like my vim config to be accessible by every other app. Let's say that for some reason you have your email address in there, or the password for your smtp server. In the current world, when you install a closed app like skype, or an open app with a security vulnerability, they can just steal all your information. One of the ideas with snaps is to limit the surface of attack. That's why I find it nice to isolate the syncthing config in /home/elopio/snap/syncthing.

Now the editor case is really interesting, because you should be able to edit every file in your system, even the ones that will break it or expose it, like bootloader conf. We are working to be able to do that in a nice and reasonably secure way. The user should understand the risks, and that they are trusting their machine to those developers. The synchronization services are similar, because I'm sure many people will want to synchronize their dot files, or their full home. My main use case for syncthing is to sync my keepass database (I trust you <3 :)

So, thanks for your time, again. I'm going to raise your points to make sure that they are in our roadmap.

pura vida.

elopio:
https://lists.ubuntu.com/archives/snapcraft/2016-November/001652.html

calmh:
Thanks, that'll have to do for now. Running as a user-owned service is also a thing that needs solving, moving forward, as syncthing is typically not launched interactively yet should run as a normal user.

PR #3739, opened 2016-11-17
lib/model: Minor cleanup to not fondle cfg.Raw things in handleDeintroductions

calmh:
### Purpose

This is a bit cleaner in my opinion as it avoid pulling in a RawCopy and uses the existing methods to iterate over devices. Adds a new `RemoveDevice` on the config wrapper, similar to `SetDevice`, which abstracts away the copying and replacing stuff.

### Testing

Same as always...

### DIFF
--- a/lib/config/wrapper.go
+++ b/lib/config/wrapper.go
@@ -209,6 +209,27 @@ func (w *Wrapper) SetDevice(dev DeviceConfiguration) error {
 	return w.replaceLocked(newCfg)
 }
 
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
@st-review merge

st-review:

st-review:
:ok_hand: Merged as a8a0bc356aff9871839a97198ca28dbab19d2f4b. Thanks, @calmh!

PR #3741, opened 2016-11-17
WIP: Log folder's ID and Label (fixes #3724)

### LINKED ISSUE #3724: Also use folder label when emitting log entries
Ref. [forum discussion](https://forum.syncthing.net/t/adding-folder-name-alongside-folder-id-for-some-log-entries/8570)

Log entries (particularly those used to emit problems) currently only mention the folder ID. This can make diagnosis difficult when one uses folder labels and has many shared folders. If a folder has a label it would be helpful to include this when emitting logs.

zaynetro:
### Purpose

This pull requests fixes issue #3724. Log entries should print folder's ID and Label.

### Testing

Testing wasn't necessary for this change.


calmh:
So there are two formatting types in use here, inconsistently, since before.

- `%s` which just formats a string normally, and can be surrounded by quotes to quote the string: `Printf("\"%s\"", "foo")` becomes `"foo"`.
- `%q` which prints an escaped string, in quotes. `Printf("%q", "foo")` becomes `"foo"`, same as the example above. But also, `Printf("%q", "fo\"o")` becomes `"fo\"o"` while with the `%s` formatting above it would print as `"fo"o"`.

So anyway, long story short, I think we should probably use `"... folder %q (%s)", label, id"` to generate something like `... folder "My folder Label" (abcde-12345)`.

calmh:
@st-jenkins add to white list

AudriusButkevicius:
To be honest, I think we should just print "%s" % cfg and then add a .String() on folder config returning whatever, in case we change something yet again.

calmh:
Actually that's a good idea. I was thinking that there were lots of places where the folder label and ID didn't come via a a FolderConfiguration object but that doesn't really seem to be the case, from the diff. So yeah.

Although I wouldn't use String() for it, but something else like Description() instead, or printing the actual object for testing and debugging gets annoying.

zaynetro:
OK, now I am using "%q (%s)" format in all log entries. Folder configuration struct has now a method `Description()` which prints folder's label and ID.

calmh:
Neat! Looks good to me, thanks! There is a merge conflict on the authors change (one of the reasons I prefer doing it), so I'll take care of merging this manually.

calmh:
Merged in d3a251e, thanks for the cleanup!

zaynetro:
Thanks for the quick replies!

lib/model/model.go
```diff
@@ -806,19 +806,19 @@ func (m *Model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
...
 						// happened. We send a full index to reset the
 						// situation.
-						l.Infof("Device %v folder %q is delta index compatible, but seems out of sync with reality", deviceID, folder.ID)
+						l.Infof("Device %v folder %q (%s) is delta index compatible, but seems out of sync with reality", deviceID, folder.Label, folder.ID)
 						startSequence = 0
 						continue
 					}
 
-					l.Debugf("Device %v folder %q is delta index compatible (mlv=%d)", deviceID, folder.ID, dev.MaxSequence)
+					l.Debugf("Device %v folder %q (%s) is delta index compatible (mlv=%d)", deviceID, folder.Label, folder.ID, dev.MaxSequence)
```
AudriusButkevicius: I was saying (over the phone, not sure if it got delivered), that in protocol/bep_extensions.go, we could add:

```
func (f Folder) Description() string {
    return fmt.Sprintf("%q (%s)", f.Label, f.ID)
}
```

and use folder.Description() in the remaining cases.

PR #3745, opened 2016-11-19
lib/connections: Slow down failing listeners

calmh:
### Purpose

This adds a supervisor with a less aggressive restart policy for listeners. Listeners would fail for essentially a few reasons I can think of

- the port not being free to bind to

- a statically configured relay is down

- a dynamic relay endpoint is misconfigured or down

None of them strikes me as something that needs to be retried in bursts every few seconds, and in fact that behavior is harmful to the network in the last two cases.

### Testing

Tested with a misconfigured dynamic relay config, which gives two retries to load and parse every ten minutes.

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -59,9 +59,10 @@ type Service struct {
 	natService           *nat.Service
 	natServiceToken      *suture.ServiceToken
 
-	listenersMut   sync.RWMutex
-	listeners      map[string]genericListener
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 548a324256bb412f7635a1975b3af5ea75db7bb0. Thanks, @calmh!

PR #3748, opened 2016-11-21
lib/fs: The interface and basicfs

calmh:
### Purpose

I'm making a new run for the filesystem and new puller code. I'm working on integrating the `Filesystem` Everywhere(tm), but I think it's easier to review and push it in parts. These are the main interfaces and basic simple os-wrapping implementation. I already have code in lots of places that uses them so I'm fairly confident they don't need to change too much.

One thing that will change at a later date is how we handle paths. All the `os` etc functions take absolute paths today, while the desired end state is that the `Filesystem` has a root (the folder root) and all other operations are on folder-relative paths. That's probably the last thing I'll change.

The next steps after this are something along the lines of:

1. The `ExtendedFilesystem` that grabs the relevant parts from osutil (this is essentially done)
2. A modified `MtimeFilesystem` that is a layer on top of the ExtendedFilesystem (this is done)
3. Integrating these new filesystems everywhere we use raw `os`, `osutil` and `symlinks` packages today. (work in progress)
4. Removing the now obsolete `symlinks` package and parts of `osutil`
5. Injecting the new puller, which is mostly plug and play once all this is done
6. Refactoring our path handling to always be folder-relative instead of absolute, with the folder root living in the `Filesystem`.

I'm hoping to get this done piece by piece over the next couple of weeks or months.


AudriusButkevicius:
Uuuuuuuuuiu i'm super excited. Can't wait to get home to review this.

calmh:
This part is not exciting but I'm glad you're looking forward to it :)

lib/fs/basicfs_walk.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"errors"
+	"path/filepath"
+)
+
+// SkipDir is used as a return value from WalkFuncs to indicate that
+// the directory named in the call is to be skipped. It is not returned
+// as an error by any function.
+var errSkipDir = errors.New("skip this directory")
+var SkipDir = errSkipDir // silences the lint warning...
```
AudriusButkevicius: Shouldn't this live in filesystem.go?
I suspect all filesystems should return the same error?

lib/fs/basicfs_symlink_windows.go
```diff
@@ -0,0 +1,197 @@
...
+
+	r0, _, err := syscall.Syscall(procCreateSymbolicLink.Addr(), 3, uintptr(unsafe.Pointer(srcp)), uintptr(unsafe.Pointer(trgp)), uintptr(mode))
+	if r0 == 1 {
+		return nil
+	}
+	return err
+}
+
+func (fs BasicFilesystem) ChangeSymlinkType(path string, tt LinkTargetType) error {
+	LinkTarget, exTt, err := fs.ReadSymlink(path)
```
AudriusButkevicius: the uppercase here is weird, also this exTt is weird.

target, existingTargetType, err ?

lib/fs/filesystem.go
```diff
@@ -0,0 +1,68 @@
...
+const (
+	LinkTargetFile LinkTargetType = iota
+	LinkTargetDirectory
+	LinkTargetUnknown
+)
+
+// The Filesystem interface abstracts access to the file system.
+type Filesystem interface {
+	ChangeSymlinkType(name string, tt LinkTargetType) error
+	Chmod(name string, mode os.FileMode) error
```
AudriusButkevicius: I say os.FileMode should be a redefined locally, copying stuff from stdlib or potentially even using an interface.
I know there is no need now, but we might add more bits to it?

lib/fs/basicfs_walk.go
```diff
@@ -0,0 +1,90 @@
...
+
+// WalkFunc is the type of the function called for each file or directory
+// visited by Walk. The path argument contains the argument to Walk as a
+// prefix; that is, if Walk is called with "dir", which is a directory
+// containing the file "a", the walk function will be called with argument
+// "dir/a". The info argument is the os.FileInfo for the named path.
```
AudriusButkevicius: no `os.`

lib/fs/basicfs_walk.go
```diff
@@ -0,0 +1,90 @@
...
+// incoming error will describe the problem and the function can decide how
+// to handle that error (and Walk will not descend into that directory). If
+// an error is returned, processing stops. The sole exception is when the function
+// returns the special value SkipDir. If the function returns SkipDir when invoked
+// on a directory, Walk skips the directory's contents entirely.
+// If the function returns SkipDir when invoked on a non-directory file,
+// Walk skips the remaining files in the containing directory.
+type WalkFunc func(path string, info FileInfo, err error) error
+
+// walk recursively descends path, calling w.
```
AudriusButkevicius: w -> walkFn

AudriusButkevicius:
So I wanted to complain about this not being rooted to some specific directory, but I see you are addressing it in point 6.

calmh-sandbox:
Test, please ignore.

calmh-sandbox:
Test again sorry

calmh:
Addressed review comments. I changed the `os.FileMode` to be a raw `uint32` instead. I figure we use the modes for two things: to figure out things about the file we're looking at, which we can/should use the `FileInfo.IsSymlink()` and similar methods for, and to get/set the bits verbatim while syncing which we might as well use an `uint32` for.

AudriusButkevicius:
Can we just have a typedef for the uint32 and use that so that there is less refactoring when the type becomes a bit more special?

calmh:
Certainly ^^

AudriusButkevicius:
@st-review merge

st-review:

st-review:
:ok_hand: Merged as fc1430aa92c364cbad169d89abd2a17f80821afd. Thanks, @calmh!

PR #3751, opened 2016-11-22
cmd/syncthing: Make the default folder default again

calmh:
The current way is quite confusing for new users - we create a default folder, but it's not usable with the default folder created somewhere else. Instead, when setting up for the first time with two devices, the default folder must be removed and recreated on one of them. This comes up on IRC and the forum now and then.

I think this matches expectactions better.

Another alternative would be to remove it entirely (not create a default folder), but then we should also add some guidance in the UI on how to proceed.

Relates to, for example, https://github.com/syncthing/docs/pull/250

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -955,9 +955,8 @@ func defaultConfig(myName string) config.Configuration {
 
 	if !noDefaultFolder {
 		l.Infoln("Default folder created and/or linked to new config")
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 462fde5e7d48128a92524a1fd63e890bcb6a9a80. Thanks, @calmh!

PR #3754, opened 2016-11-22
lib/ignore: Fast reload of unchanged ignores (fixes #3394)

### LINKED ISSUE #3394: Reloading ignores is hugely inefficient
We reload the ignores before each folder scan. This involves parsing and compiling the ignore patterns. If there are a lot of patterns that can be expensive. If running with syncthing-inotify we might get a lot of scan requests.

We should just check the .stignore file timestamp and not do anything unless it's changed.

It's a mystery to me why we don't already do this. I must have been drunk when I wrote this (or reviewed it, if someone else is responsible. I forget.).

calmh:
### Purpose

This changes the "seen" map that we're anyway keeping around to track the modtimes of loaded files instead. When doing a Load() we check that

1) the file we are loading is in the modtime set, and
2) that none of the files in the modtime set have changed modtimes.

If that's the case we do a quick return without parsing anything or clearing the cache.

This required adding ~~two~~three one seconds sleeps in the tests to make sure the modtimes were updated when we expect cache reloads, because I'm on a crappy filesystem with one second timestamp granularity. That also proves it works...

### DIFF
--- a/lib/ignore/ignore.go
+++ b/lib/ignore/ignore.go
@@ -69,6 +69,7 @@ type Matcher struct {
 	matches   *cache
 	curHash   string
 	stop      chan struct{}
+	modtimes  map[string]time.Time
 	mut       sync.Mutex
 }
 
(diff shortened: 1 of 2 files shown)


st-review:
:ok_hand: Merged as a2b8485a898adeb1d7ad16a739d56cd7a8b13168. Thanks, @calmh!

AudriusButkevicius:
However I'd use Chtimes or reach into the x.modtimes to reduce the test time.

calmh:
Good point, I'll fix that

PR #3775, opened 2016-11-30
lib/protocol: Apply input filtering on file names

calmh:
### Purpose

Stricter filtering of incoming filenames. Essentially, names on the wire should be in canonical shortest possible form -- no internal `.` or `..` components, no double slashes, not be one of `.` or `..`, not be absolute (start with `/`), not end with `/`, not start with `../`, and not be empty.

(.stignore and similar special files should not be served, but it is not a *protocol* error to talk about them, it's a Syncthing specific limitation that we will not serve them. Coming in another PR.)

On Windows, files in index messages containing backslashes are filtered out with a warning. Requests for such files get an error return and a warning logged.

### Testing

The filter methods have some unit tests on them.

### Documentation

Requires additions to the spec to explain the restrictions we apply here.

@Unrud for review.

### DIFF
--- a/lib/protocol/nativemodel_windows.go
+++ b/lib/protocol/nativemodel_windows.go
@@ -6,29 +6,64 @@ package protocol
 
 // Windows uses backslashes as file separator
 
-import "path/filepath"
+import (
+	"path/filepath"
+	"strings"
+)
 
 type nativeModel struct {
 	Model
 }
 
 func (m nativeModel) Index(deviceID DeviceID, folder string, files []FileInfo) {
-	fixupFiles(folder, files)
+	files = fixupFiles(files)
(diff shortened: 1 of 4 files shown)


calmh:
@st-review don't merge this until we've considered it carefully

st-review:

calmh:
Metalint is `lib/protocol/protocol.go:55:1:warning: errInternalFilename is unused (deadcode)`, I'll fix that.

AudriusButkevicius:
Lgtm

Unrud:
I'm fine with this.

calmh:
@st-review merge

st-review:
:ok_hand: Merged as 3266aae1c3ffd298ea2aada170b14ede05260296. Thanks, @calmh!

PR #3776, opened 2016-11-30
lib/model: Double check results in filepath.Join where needed

calmh:
### Purpose

Wherever we have untrusted relative paths, make sure they are not escaping their folder root.

### Testing

Unit tests on the new functions.

@Unrud to review.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -119,6 +119,7 @@ var (
 	errDeviceUnknown       = errors.New("unknown device")
 	errDevicePaused        = errors.New("device is paused")
 	errDeviceIgnored       = errors.New("device is ignored")
+	errNotRelative         = errors.New("not a relative path")
 )
 
 // NewModel creates and starts a new model. The model starts in read-only mode,
@@ -1105,12 +1106,15 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset 
 	//     4. Eliminate .. elements that begin a rooted path:
 	//        that is, replace "/.." by "/" at the beginning of a path,
 	//        assuming Separator is '/'.
-	fn := filepath.Join(folderPath, name)
-
-	if !strings.HasPrefix(fn, folderPath) {
+	fn, err := rootedJoinedPath(folderPath, name)
+	if err != nil {
 		// Request tries to escape!
 		l.Debugf("%v Invalid REQ(in) tries to escape: %s: %q / %q o=%d s=%d", m, deviceID, folder, name, offset, len(buf))
 		return protocol.ErrInvalid
+	}
+
+	if isInternal(name) {
+		return protocol.ErrNoSuchFile
 	}
 
 	if folderIgnores != nil {
@@ -1152,7 +1156,7 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset 
 		// file has finished downloading.
 	}
 
-	err := readOffsetIntoBuf(fn, offset, buf)
+	err = readOffsetIntoBuf(fn, offset, buf)
 	if os.IsNotExist(err) {
 		return protocol.ErrNoSuchFile
 	} else if err != nil {
@@ -1700,7 +1704,9 @@ func (m *Model) ScanFolderSubdirs(folder string, subs []string) error {
 func (m *Model) internalScanFolderSubdirs(folder string, subDirs []string) error {
 	for i, sub := range subDirs {
 		sub = osutil.NativeFilename(sub)
-		if p := filepath.Clean(filepath.Join(folder, sub)); !strings.HasPrefix(p, folder) {
+		// We test each path by joining with "root". What we join with is
+		// not relevant, we just want the dotdot escape detection here.
+		if _, err := rootedJoinedPath("root", sub); err != nil {
 			return errors.New("invalid subpath")
 		}
 		subDirs[i] = sub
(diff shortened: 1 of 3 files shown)


calmh:
@st-review don't merge this until after due consideration

st-review:

AudriusButkevicius:
Lgtm

lib/model/model.go
```diff
@@ -2550,24 +2556,39 @@ func makeForgetUpdate(files []protocol.FileInfo) []protocol.FileDownloadProgress
...
+		// deeper place in the hierarchy (i.e. someOtherFolder/.stignore is
+		// just a file as far as we're concerned.)
+		if name == internal {
+			return true
+		}
+		if strings.HasPrefix(name, internal+string(os.PathSeparator)) {
+			return true
+		}
+	}
+	return false
```
Unrud: Shouldn't this also check for temporary files?
This function can also be used for checking paths in [scanner/walk.go](https://github.com/syncthing/syncthing/blob/master/lib/scanner/walk.go#L260). Maybe it should be public.

lib/model/model.go
```diff
@@ -2550,24 +2556,39 @@ func makeForgetUpdate(files []protocol.FileInfo) []protocol.FileDownloadProgress
...
 	case matcher.Match(file.FileName()).IsIgnored():
-		// ignore patterns second because ignoring them is a valid way to
-		// silence warnings about them being invalid and so on.
 		return true
 	}
 
 	return false
 }
 
+func isInternal(name string) bool {
```
Unrud: The path in **name** must be clean for this function to work, I would add that as documentation.

calmh: Yeah. I'll look at the parent trimming thing. That's a somewhat different thing though, as it gets fed via a path that cleans names and should only be looking at things that are already below the specified root.

lib/model/model.go
```diff
@@ -2606,3 +2627,37 @@ func (s folderDeviceSet) sortedDevices(folder string) []protocol.DeviceID {
...
+
+	// The relative path should be clean from internal dotdots and similar
+	// funkyness.
+	rel = filepath.FromSlash(rel)
+	if filepath.Clean(rel) != rel {
+		return "", errInvalidFilename
+	}
+
+	// It is not acceptable to attempt to traverse upwards.
+	if strings.HasPrefix(rel, ".."+string(os.PathSeparator)) {
```
calmh: It's tested for though, gets caught by the prefix check below. This check is needed anyway to catch `join("foo/bar", "../bar/baz")` which becomes "foo/bar/baz" (but should still not be allowed).

lib/model/model.go
calmh: Temp files I'm not convinced about belongs in the same place, but yes there should be a central place where the knowledge about what files are internal lives. This is a good as any. But then it needs to live somewhere else - `config` perhaps, or `scanner`. Hmm.

lib/model/model.go
Unrud: This depends on the value of **root**. ```Join("/", "..")``` becomes "/".

lib/model/model.go
```diff
@@ -2606,3 +2627,37 @@ func (s folderDeviceSet) sortedDevices(folder string) []protocol.DeviceID {
...
 }
+
+// rootedJoinedPath takes a root and a supposedly relative path inside that
+// root and returns the joined path. An error is returned if the joined path
+// is not in fact inside the root.
+func rootedJoinedPath(root, rel string) (string, error) {
+	// The expected prefix for the resulting path is the root, with a path
+	// separator at the end.
+	expectedPrefix := filepath.FromSlash(root)
+	if !strings.HasSuffix(expectedPrefix, string(os.PathSeparator)) {
```
Unrud: If **root** is "" the value of **expectedPrefix** must also be "" not os.PathSeparator. Otherwise the check at the end will always fail.

lib/model/model.go
```diff
@@ -2606,3 +2627,37 @@ func (s folderDeviceSet) sortedDevices(folder string) []protocol.DeviceID {
...
+		expectedPrefix += string(os.PathSeparator)
+	}
+
+	// The relative path should be clean from internal dotdots and similar
+	// funkyness.
+	rel = filepath.FromSlash(rel)
+	if filepath.Clean(rel) != rel {
+		return "", errInvalidFilename
+	}
+
```
Unrud: Should also check that ```filepath.VolumeName(rel) == ""``` to prevent stuff like "\\\\bla" on Windows.

calmh:
Adressed. Obviously, with root="/" you can access arbitrary paths but the same syntax rules for the relative part should apply. I've removed the isInternal check, keeping this focused on the join-path-with-root sanitation -- will move those checks into a common place in the ignore handling instead.

Unrud:
I'm fine with this.

lib/model/model.go
```diff
@@ -2606,3 +2590,57 @@ func (s folderDeviceSet) sortedDevices(folder string) []protocol.DeviceID {
...
+	// It is not acceptable to attempt to traverse upwards or refer to the
+	// root itself.
+	switch rel {
+	case ".", "..", pathSep:
```
Unrud: Does checking for ``"."`` break **internalScanFolderSubdirs**?

calmh:
@st-review merge

st-review:
:ok_hand: Merged as 63194a37f668630f7040bc15df39f9d1c45965fa. Thanks, @calmh!

PR #3778, opened 2016-12-01
Prevent collisions in the progressemitter registry

Unrud:
### Purpose

Prevent collisions in the progressemitter registry.
Using ``filepath.Join`` can cause collisions. The folder ID could be something like ``".."`` or ``"../.."``.

### DIFF
--- a/lib/model/progressemitter.go
+++ b/lib/model/progressemitter.go
@@ -8,7 +8,6 @@ package model
 
 import (
 	"fmt"
-	"path/filepath"
 	"time"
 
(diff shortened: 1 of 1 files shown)


calmh:
@st-review merge it

lib/model: Prevent collisions in the progressemitter registry

Using filepath.Join can cause collisions. The folder ID could be something like ".." or "../..".

st-review:
:ok_hand: Merged as cabe94552af5810e435b22d1ed9582ef2e833d3e. Thanks, @Unrud!

PR #3780, opened 2016-12-01
Add write-only folder type

Smiley73:
### Purpose

This change adds a receive-only folder type feature to Syncthing, as discussed under [8852](https://forum.syncthing.net/t/pull-only-folder-type/8592).
The receiveonlyfolder type has an advanced boolean flag (DeleteLocalChanges), which enables the automatic deletion of locally detected changes. This needs to specifically be enabled. Versioning and sync conflicts work.

### Testing
Automated tests have been added to `receiveonlyfolder_test.go`.

_**Manual Testing**_

**With DeleteLocalChanges** 
2 devices, 1 folder. Folder as sendreceivefolder on one side and receiveonlyfolder on the other side.
Enable DeleteLocalChanges on receiveonlyfolder.
File changed on receiveonlyfolder . Hit sync. Message in log. File gets reverted to original version
Created file on receiveonlyfolder . Hit sync. Message in log. File gets removed.
Created directory on receiveonlyfolder . Hit sync. Message in log. Directory gets removed.

**Without DeleteLocalChanges** 
2 devices, 1 folder. Folder as rwfolder on one side and wofolder on the other side.
Disable DeleteLocalChanges on wofolder on
File changed on receiveonlyfolder . Hit sync. Message in log. File gets reverted to original version
Created file on receiveonlyfolder . Hit sync. Message in log. No local changes.
Created directory on receiveonlyfolder . Hit sync. Message in log. No local changes.

**Repeat all tests with versioning on & conflicts off.
Repeat all tests with versioning off & conflicts on.**

### Screenshots

image

### Documentation

[Pull Request](https://github.com/syncthing/docs/pull/273)

### Authorship
Heiko Zuerker

# Remaining work to be completed

- [ ] Reporting - Folders shouldn't claim to be in-sync if one side is ignoring local changes. (see calmh's response from 2017-03-11)


calmh:
@st-jenkins add to whitelist

lib/model/model.go
```diff
@@ -1539,13 +1540,121 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
+			if file.IsDirectory() {
+				objType = "dir"
+			}
+
+			if file.IsDeleted() {
+				action = "deleted"
+			}
 
+			if len(file.Version.Counters) == 1 && file.Version.Counters[0].Value == 1 {
+				// A file, directory or symlink was added, which we'll have to remove again
```
calmh: It can have any number of counters and stuff the first time it's added. Your check covers the first time a file is discovered anywhere, but it might already be well known and modified many times when *we* hear about it for the first time.

lib/model/model.go
```diff
@@ -1539,13 +1540,121 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
+					} else {
+						fileDeletions = append(fileDeletions, file)
+					}
+				} else {
+					correctiveaction = "none"
+				}
+			} else {
+				file.Deleted = false
+				file.Invalid = true
+				file.Version = file.Version.SetZero(m.shortID)
```
calmh: If you're trying to ensure that this index entry is older than anything else, I don't think this will do. Setting the version to the empty vector should do it though.

lib/model/model.go
```diff
@@ -1539,13 +1540,121 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
+	m.fmut.RLock()
+	folderCfg := m.folderCfgs[folder]
+	m.fmut.RUnlock()
+
+	folderrunner := m.folderRunners[folder]
+	var newfs []protocol.FileInfo
+
+	// if this folder is a "write/pull only" folder, then we'll have to undo any local changes
+	if folderCfg.Type == config.FolderTypeWriteOnly {
+		// reject and undo all local changes
```
calmh: This whole thing should be broken out into something other than updateLocals

lib/model/util.go
```diff
@@ -58,3 +67,95 @@ func (d *deadlockDetector) Watch(name string, mut sync.Locker) {
 		}
 	}()
 }
+
+// moveforconflict renames a file to deal with sync conflicts
+func moveforconflict(name string, maxConflicts int) error {
```
calmh: Generally speaking, camel case your names.

lib/model/model.go
```diff
@@ -1539,13 +1540,121 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
+				action = "added"
+				if folderCfg.DeleteLocalChanges {
+					correctiveaction = "deleted"
+					if file.IsDirectory() {
+						dirDeletions = append(dirDeletions, file)
+					} else {
+						fileDeletions = append(fileDeletions, file)
+					}
+				} else {
+					correctiveaction = "none"
```
calmh: Not doing anything with the file, and not accepting it into the database, means it will be rehashed on every scan. That's probably not great.

calmh:
Sorry, trying to avoid nitpicking until we get to that stage, the naming and method extraction things are borderline. I think the more important parts are making sure things work in terms of getting the right actions and avoiding eternal rescans etc at this point.

Smiley73:
No worries. I'll start hammering away at the changes as time permits. Stay tuned for updates (might be a week or more).

wweich:
From the previous discussion, I understood that the "deleteLocalChanges" setting would be a "real" advanced setting (Actions -> Advanced) and not right beside the folder type selection. That way, it would be for the real advanced users.

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -99,6 +101,14 @@
               </div>
               <p translate class="help-block">File permission bits are ignored when looking for changes. Use on FAT file systems.</p>
             </div>
+            <div ng-if="currentFolder.type == 'writeonly'" class="form-group">
+              <div class="checkbox">
+                <label>
+                  <input type="checkbox" ng-model="currentFolder.deleteLocalChanges"> <span translate>Delete local changes</span>
+                </label>
+              </div>
+              <p translate class="help-block">Files or folders that have been added locally, will automatically be removed on the next scan. Enable versioning or sync-conflict handling to not lose your data! </p>
```
AudriusButkevicius: Extra space before the closing element

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -99,6 +101,14 @@
               </div>
               <p translate class="help-block">File permission bits are ignored when looking for changes. Use on FAT file systems.</p>
             </div>
+            <div ng-if="currentFolder.type == 'writeonly'" class="form-group">
+              <div class="checkbox">
+                <label>
+                  <input type="checkbox" ng-model="currentFolder.deleteLocalChanges"> <span translate>Delete local changes</span>
```
AudriusButkevicius: Extra space before the span.

lib/model/model.go
```diff
@@ -1539,13 +1540,121 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
 }
 
 func (m *Model) updateLocalsFromScanning(folder string, fs []protocol.FileInfo) {
-	m.updateLocals(folder, fs)
+	m.fmut.RLock()
+	folderCfg := m.folderCfgs[folder]
+	m.fmut.RUnlock()
+
+	folderrunner := m.folderRunners[folder]
+	var newfs []protocol.FileInfo
```
AudriusButkevicius: newFs

lib/model/model.go
```diff
@@ -1539,13 +1540,121 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
 }
 
 func (m *Model) updateLocalsFromScanning(folder string, fs []protocol.FileInfo) {
-	m.updateLocals(folder, fs)
+	m.fmut.RLock()
+	folderCfg := m.folderCfgs[folder]
+	m.fmut.RUnlock()
+
+	folderrunner := m.folderRunners[folder]
```
AudriusButkevicius: folderRunner

lib/model/model.go
```diff
@@ -1539,13 +1540,121 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
+		// reject and undo all local changes
+		// GAP: need to handle file renaming also (see rwfolder line 511+ comparing blocks)
+
+		fileDeletions := []protocol.FileInfo{}
+		dirDeletions := []protocol.FileInfo{}
+
+		for _, file := range fs {
+			if strings.Contains(file.Name, ".sync-conflict-") {
+				// this is a conflict copy, let's move on to the next file
+				continue
```
AudriusButkevicius: These shouldn't happen, it's a one way street isn't it?

lib/model/model.go
```diff
@@ -1539,13 +1540,121 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
+			}
+
+			objType := "file"
+			action := "modified"
+			correctiveaction := "resync"
```
AudriusButkevicius: correctiveAction

lib/model/model.go
```diff
@@ -1539,13 +1540,121 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
+						fileDeletions = append(fileDeletions, file)
+					}
+				} else {
+					correctiveaction = "none"
+				}
+			} else {
+				file.Deleted = false
+				file.Invalid = true
+				file.Version = file.Version.SetZero(m.shortID)
+				//file.Sequence = 0
```
AudriusButkevicius: Not a meaningful comment

lib/model/model.go
```diff
@@ -1539,13 +1540,121 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
+				newfs = append(newfs, file)
+			}
+
+			// we better tell the user on the UI and in the log that we had to take corrective actions
+			l.Warnln("Rejecting local change on folder", folderCfg.Description(), objType, file.Name, "was", action, "corrective action:", correctiveaction)
```
AudriusButkevicius: I accidently dropped my pictures folder in, now my UI is flooded.

lib/model/model.go
```diff
@@ -1539,13 +1540,121 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
+		m.updateLocals(folder, fs)
+
+		// Fire the LocalChangeDetected event to notify listeners about local updates.
+		m.localChangeDetected(folderCfg, fs)
+	}
+
+}
+
+// deleteRejectedDir attempts to delete the given directory
+func (m *Model) deleteRejectedDir(folder string, file protocol.FileInfo, ver versioner.Versioner) {
```
AudriusButkevicius: ver never used.

lib/model/rwfolder.go
```diff
@@ -690,20 +692,9 @@ func (f *rwFolder) deleteDir(file protocol.FileInfo, matcher *ignore.Matcher) {
...
-			fullDirFile := filepath.Join(file.Name, dirFile)
-			if defTempNamer.IsTemporary(dirFile) || (matcher != nil && matcher.Match(fullDirFile).IsDeletable()) {
-				os.RemoveAll(filepath.Join(f.dir, fullDirFile))
-			}
-		}
-		dir.Close()
-	}
 
-	err = osutil.InWritableDir(os.Remove, realName)
+	err = deletedir(f.dir, file, matcher)
```
AudriusButkevicius: The fact that this logic has moved from rwfolder to some function that is not part of the struct, that is now used by the model smells really bad.
As I said, I think you should make a:

```
type rofolder struct {
  rwfolder
}
```

and override the few places that are relevant to you.

We print folder type somewhere in model.StartFolder() which will definately misreport in your case.

lib/model/model.go
Smiley73: Not handling this case caused issues, since it detects the sync conflict file as a new local file and once again wants to reject the local change, creating a new sync conflict file, ...

calmh:
Let's stick to the structural and big things first, and we can pick apart style things once it's generally sound. Lest we overwhelm @smiley73 at the starting line. :)

Smiley73:
Yeah please go easy on that poor guy! :-)

calmh:

calmh:

lib/model/model.go
AudriusButkevicius: Em, why don't we just delete the file?

lib/model/model.go
Smiley73: Done

lib/model/model.go
Smiley73: Agreed, that could cause quite a bit of unnecessary load on the system.
I'm currently trying to understand how this all ties together. Do you have any suggestions on what the best place to catch these is?

chylli:
`Write-Only` this name is a little odd

wweich:
Thats the internal name in contrast to read-only (for master). For that reason, he started a [discussion](https://forum.syncthing.net/t/renaming-of-folder-types/8728) for renaming the folder types.

calmh:
Adding a note here as well as it relates to #3793. *This* change may well require changing the config format and mean a 0.15 release. That's fine.

Smiley73:
We should be able to get away without any new config format for this. We'd just have to add the "writeonly" type and rename it in 0.15 to "receiveonly". Your call.
I'll get the other PR taken care of first, then we can tackle this one.

lib/model/rwfolder.go
Smiley73: The folder type is correctly displayed by model.StartFolder():
`[YKC4E] 14:38:32 INFO: Restarted folder wvd4j-ksyym (receiveonly)`

I've tried making the change you suggested, but I can't get it to work. The embedding of the other type only partially works. Might simply be my limited understanding of golang...

AudriusButkevicius:
You should drop a message as you update stuff, as I don't think github notifies about new commits.

lib/model/model.go
```diff
@@ -1558,13 +1559,129 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
 
 func (m *Model) updateLocalsFromScanning(folder string, fs []protocol.FileInfo) {
-	m.updateLocals(folder, fs)
+	m.fmut.RLock()
+	folderCfg := m.folderCfgs[folder]
+	m.fmut.RUnlock()
+
+	// if this folder is a "write/pull only" folder, then we'll have to undo any local changes
+	if folderCfg.Type == config.FolderTypeReceiveOnly {
+		m.rejectLocalChanges(folder, fs)
```
AudriusButkevicius: I still strongly suggest you create a:
```
type receiveOnlyFolder type {
  sendOnlyFolder
}
```

and override the bit that handles the scanning to do reverts instead, then this whole "detach deleteDir from the struct" business goes away.

AudriusButkevicius:
A bit of a broken record, but here it goes.

So I guess the code is sane, and probably works, yet I am still of a strong opinion that the functionality is being wedged in in the wrong place.

I guess I am fine to duplicate code/functionality/perform some nasty type embedding to get the job done in the right place, as I fully understand that the current code is not easy to mend to fit the purpose, and I don't expect you be the superhero that fixes that, yet I am against wedging in the functionality where it doesn't belong.

I wouldn't mind if you did a literal `cp rwfolder.go rofolder.go`, and then ripped out what you don't need, added what you do need, even if it resulted in tons of duplicate code. This way decomposing things into separate reusable components would be a much easier task when someone decides to undertake it in the future, as it's obvious which parts of code overlap.

I don't know how @calmh wants to spin this.

Smiley73:
I wanted to stay away from duplicating code, since it will make the maintenance going forward harder. You will always have to remember doing the same code change in two places. There must be a better way. 
To my previous message, I couldn't figure out how to "inherit" or "enhance" the sendOnlyFolder. Everything I tried didn't work and all the examples I found didn't seem to work. I'm pretty sure it's my understanding of golang. I have a couple off days coming up, hopefully I can give it another try. If you know the solution, then I wouldn't mind a pointer.

AudriusButkevicius:
```
type receiveOnlyFolder type {
  sendOnlyFolder
}
```
This is how you inherit.

And this is how you override/call parent:

```
func (f *receiveOnlyFolder) foo() {
   Do stuff in child...
   f.sendOnlyFolder.foo() // call parent
}
```

calmh:
Let me be the bad guy and throw out another roadblock here. Regardless of the organizational points above, this will not be merged without some sort of reasonable test suite, given that it has the potential to be disruptive. I don't particularly care if there are unit tests, integration tests (see the `test` directory) that remote controls syncthing instances, a shell script to do the same, or what - and I can help suggest how to do things, but we there needs to be tests.

Smiley73:
@calmh & @AudriusButkevicius 

I've been trying to figure out how I can prevent the rehashing of the locally rejected files on every scan. The walkRegular function in walk.go rehashes a file if it has the invalid flag set. I believe this is the right place to prevent the hashing for rejected files.
Obviously the easiest would be if I could just change it to not hash when the invalid bit is set, but I'm pretty sure it's not the behavior you guys want.
Any suggestions on how to handle this, without having to introduce a new FileInfo flag?

calmh:
> without having to introduce a new FileInfo flag?

I'd consider introducing a new FileInfo flag. Or possibly redefining what the invalid flag means. I'm on mobile right now and not going to grep through the code but I think it has very few uses today. A flag that meant something like "do not send this index entry" and that wasn't actually part of the defined on-the-wire protocol maybe a solution here. Syncthing could then hash and index the files and set that bit, so that it could track the state locally without the file entry being propagated to anyone else. If that helps. I'm not sure. Experiment, and if you find a better solution that's great but in the grand scheme of things I think a new flag is probably not the biggest hurdle here.

AudriusButkevicius:
I wonder if there is anything we could get from the index.
I suspect you could check for an existing "current" entry in the db, yet having an empty version vector, implying that i was rejected previously.

Smiley73:
I was able to fix the rehashing issue with just using the invalid flag.
Please take a closer look at lib/db/leveldb_dbinstance.go/updateFiles . I had to remove an if statement there and want to make sure it doesn't cause problems. This function should never be called unless there was a change to the file (confirmed by my testing), so it should be hopefully good.

calmh:
This looks like a valid use of the invalid flag to me.

Smiley73:
@calmh test cases have been added. Thanks for the pointer to requests_test.go, that was very helpful.

I did not add any test cases with versioning enabled, because I'm not doing anything special and just rely on the existing functionality.

Smiley73:
@calmh & @AudriusButkevicius 
How are we looking on this pull request? Anything else you want me to change?

lib/db/leveldb_dbinstance.go
```diff
@@ -217,19 +219,19 @@ func (db *Instance) updateFiles(folder, device []byte, fs []protocol.FileInfo, l
...
-				localSize.addFile(f)
-			}
 
-			t.insertFile(folder, device, f)
-			if f.IsInvalid() {
-				t.removeFromGlobal(folder, device, name, globalSize)
-			} else {
-				t.updateGlobal(folder, device, f, globalSize)
-			}
+		// update the DB entry if the file has changed
```
AudriusButkevicius: Can you explain why we don't need the if anymore?

AudriusButkevicius:
The functionality is still in the model which is wrong, so I am a bit puzzled why you have the new type at all, as ot serves not much purpose. I though the idea was to move the functionality into the folder

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -86,10 +86,13 @@
               <label translate>Folder Type</label>
               &nbsp;<a href="https://docs.syncthing.net/users/foldertypes.html" target="_blank"><span class="fa fa-book"></span>&nbsp;<span translate>Help</span></a>
               <select class="form-control" ng-model="currentFolder.type">
+
```
AudriusButkevicius: Blank line?

lib/db/leveldb_dbinstance.go
Smiley73: I had to remove an if statement here, because otherwise it would never update the record for the file and constantly rehash. This was due to the fact that only the checksum was changed. I could have added another check to the if statement, but it didn't seem to make sense. I did some testing and this function never gets called unless there was an actual change to the file.

Smiley73:
@AudriusButkevicius yeah I dropped the ball on that one. I moved everything over to the new type. The change is a lot cleaner now.

lib/model/util.go
```diff
@@ -59,3 +68,95 @@ func (d *deadlockDetector) Watch(name string, mut sync.Locker) {
 		}
 	}()
 }
+
+// moveforconflict renames a file to deal with sync conflicts
+func moveForConflict(name string, maxConflicts int) error {
```
AudriusButkevicius: These all should now live with the folder as we no longer call them from the model

lib/model/receiveonlyfolder.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return f
+}
+
+func (f *receiveOnlyFolder) String() string {
+	return fmt.Sprintf("receiveOnlyFolder/%s@%p", f.folderID, f)
+}
+
+// rejectLocalChanges reverts all local changes
+func (f *receiveOnlyFolder) rejectLocalChanges(fs []protocol.FileInfo) []protocol.FileInfo {
```
AudriusButkevicius: I still think this should live in a modified version of Scan() so there would be not be `if type == x do y` type of branching in the model.

lib/model/util.go
Smiley73: This function is used in rwfolder also, that's why I moved it to util.go. Are you suggesting to move it to folder.go instead?

AudriusButkevicius: Yeah, it's not used outside of the folder

lib/model/receiveonlyfolder.go
Smiley73: Trying to do this via Scan would be much more complicated I believe. I'm also trying very hard to avoid duplication of code.
updateLocalsFromScanning seems to be the most efficient place to catch the local changes.

AudriusButkevicius: As I said, if I have to choose duplicate code versus bleeding functionality, I'll choose duplicate code any day.

lib/model/util.go
Smiley73: Okay I'll move the functions from util.go to folder.go as requested. Should hopefully have some time this weekend.

lib/model/receiveonlyfolder.go
AudriusButkevicius: @calmh agree?

Smiley73: Yeah sorry, but this just doesn't sound like the right approach to me. I got a big issue with code duplication. This not only doubles the maintenance going forward, but also adds the risk of forgetting to do things in two places and thus causing even more problems.

Is the if statement what bothers you? I can change it up so we have a function in folders that always gets called. Something like "validateLocalChanges". We then would simply overwrite it for receiveonly to do some real action. Or we could even simply call rejectLocalChanges, which won't do anything on the other folder types.

AudriusButkevicius: I am fine to add a function which is noop for all existing folder types, apart from yours, then the functionality is contained within the folder implementation, and there is no branching in the model based on the type of folder it is.

That's essentially what bothers me.

AudriusButkevicius: As in, "doMyRevertsOrWhatever" which is noop for existing folders.

calmh: > I am fine to add a function which is noop for all existing folder types, apart from yours, then the functionality is contained within the folder implementation, and there is no branching in the model based on the type of folder it is.

👍

lib/model/rwfolder.go
```diff
@@ -789,20 +789,8 @@ func (f *sendReceiveFolder) deleteDir(file protocol.FileInfo, matcher *ignore.Ma
...
-		files, _ := dir.Readdirnames(-1)
-		for _, dirFile := range files {
-			fullDirFile := filepath.Join(file.Name, dirFile)
-			if defTempNamer.IsTemporary(dirFile) || (matcher != nil && matcher.Match(fullDirFile).IsDeletable()) {
-				os.RemoveAll(filepath.Join(f.dir, fullDirFile))
-			}
-		}
-		dir.Close()
-	}
+	err = deleteDir(f.dir, file, matcher)
```
AudriusButkevicius: These should move back to the folder definition.

lib/model/rwfolder.go
```diff
@@ -847,18 +835,7 @@ func (f *sendReceiveFolder) deleteFile(file protocol.FileInfo) {
...
-		// of deleting. Also merge with the version vector we had, to indicate
-		// we have resolved the conflict.
-		file.Version = file.Version.Merge(cur.Version)
-		err = osutil.InWritableDir(f.moveForConflict, realName)
-	} else if f.versioner != nil {
-		err = osutil.InWritableDir(f.versioner.Archive, realName)
-	} else {
-		err = osutil.InWritableDir(os.Remove, realName)
-	}
+	err = deleteFile(f.dir, file, f.versioner, f.MaxConflicts)
```
AudriusButkevicius: And this.

lib/model/rwfolder.go
```diff
@@ -1652,47 +1629,7 @@ func removeAvailability(availabilities []Availability, availability Availability
...
-				if gerr != nil {
-					l.Debugln(f, "removing extra conflict", gerr)
-				}
-			}
-		} else if gerr != nil {
-			l.Debugln(f, "globbing for conflicts", gerr)
-		}
-	}
-	return err
+	return moveForConflict(name, f.MaxConflicts)
```
AudriusButkevicius: and this.

lib/model/rwfolder.go
Smiley73: Done

lib/model/rwfolder.go
Smiley73: Done

lib/model/rwfolder.go
Smiley73: Done

lib/model/receiveonlyfolder.go
Smiley73: Done. I have to add an if statement for the folder runner being nil, due to some test cases not actually starting the folder.

Smiley73:
@calmh & @AudriusButkevicius 
Anything else you guys want me to change? I got some time today and would like to get close to wrapping this up.

lib/model/receiveonlyfolder.go
```diff
@@ -0,0 +1,158 @@
...
+func (f *receiveOnlyFolder) deleteRejectedDir(file protocol.FileInfo) {
+	err := f.folder.deleteDir(f.Path(), file, nil)
+	if err != nil && !os.IsNotExist(err) {
+		l.Infof("deleteRejectedDir (folder %q, file %q): delete: %v", f, file.Name, err)
+	}
+}
+
+// deleteRejectedFile attempts to delete the given file
+func (f *receiveOnlyFolder) deleteRejectedFile(file protocol.FileInfo, ver versioner.Versioner) {
+	err := f.folder.deleteFile(f.Path(), file, ver, f.MaxConflicts)
```
AudriusButkevicius: Not sure if it's possible to change sendReceiveFolder to embed folder fully, then these could just be f.deleteFile, if not, then it's fine as it is, though yeah, it's not for this pull request either.

lib/model/model.go
```diff
@@ -1569,9 +1569,13 @@ func (m *Model) updateLocalsFromScanning(folder string, fs []protocol.FileInfo)
...
-	fs = runner.validateLocalChanges(fs)
-
-	m.updateLocals(folder, fs)
+	if runner == nil {
+		l.Infoln("bug: folderrunner is nil (should only happen in tests)")
+		m.updateLocals(folder, fs)
+	} else {
+		// validate all local changes and update the database
+		fs = runner.validateAndUpdateLocalChanges(fs)
+	}
```
Smiley73: Unfortunately we do need this. Test cases in model_test.go fail, since they never start the folder and thus runner will be nil.

lib/model/folder.go
```diff
@@ -51,3 +62,105 @@ func (f *folder) scanSubdirsIfHealthy(subDirs []string) error {
...
+
+// a function to provide the ability to validate and modify local changes
+// before they are committed to the database
+// Default behavior is to apply the changes as-is, overwrite function as needed
+func (f *folder) validateAndUpdateLocalChanges(fs []protocol.FileInfo) []protocol.FileInfo {
+	// update the database
+	f.model.updateLocals(f.folderID, fs)
+
+	return fs
+}
```
Smiley73: Phew... You started to scare me for a second while I was reading this.

lib/model/model.go
AudriusButkevicius: We can just wedge in a folder in the test in that case.

Smiley73: Done

lib/model/model.go
```diff
@@ -1575,8 +1575,7 @@ func (m *Model) updateLocalsFromScanning(folder string, fs []protocol.FileInfo)
 	m.fmut.RUnlock()
 
 	if runner == nil {
-		l.Infoln("bug: folderrunner is nil (should only happen in tests)")
-		m.updateLocals(folder, fs)
+		panic("updateLocalsFromScanning: can not find folder " + folder)
```
AudriusButkevicius: I'd remove the if statement, and let the code blow up trying to access a method on nil.

lib/model/rwfolder_test.go
```diff
@@ -64,6 +64,7 @@ func setUpModel(file protocol.FileInfo) *Model {
 	db := db.OpenMemory()
 	model := NewModel(defaultConfig, protocol.LocalDeviceID, "device", "syncthing", "dev", db, nil)
 	model.AddFolder(defaultFolderConfig)
+	model.StartFolder("default")
```
AudriusButkevicius: probably defer model.tearDownFolder or whatever it is.

AudriusButkevicius: nvm...

lib/model/model.go
Smiley73: You wish is my command. ;-)

AudriusButkevicius:
So this looks good to me. 
Some comments start with a lowercase letter which is unusual.
I am not sure I understand the changes in lib/db, as that's usually not my territory, but more or less looks good to me.

Time for @calmh

Smiley73:
I went through an (hopefully) fixed all the comments that started with lower case

Smiley73:
@calmh 
Gentle reminder to chime in on this PR.

calmh:
Okay! I'll try to take a look tomorrow, then I'm heading out to shanghai and back on tuesday, so if you don't hear anything during that dime don't despair.

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -99,6 +101,14 @@
               </div>
               <p translate class="help-block">File permission bits are ignored when looking for changes. Use on FAT file systems.</p>
             </div>
+            <div ng-if="currentFolder.type == 'receiveonly'" class="form-group">
+              <div class="checkbox">
+                <label>
+                  <input type="checkbox" ng-model="currentFolder.deleteLocalChanges"><span translate>Delete local changes</span>
```
calmh: I first wrote this comment:

> I don't think "delete" is the right word here, in the description and config name - isn't "revert" more what we're doing? (i.e., we can't delete a change that deletes a file, but we can revert it)

then later realized that this is because it always reverts something that is not an addition, hence this only controls deletes of additions. That is not in the spirit of how I want things to work in order to minimize damage. I would expect this to be *be* a true toggle of `revertLocalChanges` so that when it's false we do not stomp on the user's changes - we just don't send them onward towards the cluster.

lib/db/leveldb_dbinstance.go
calmh: Yeah I think the "if" is vestigial so this should be fine. The comment above is incorrect though, there is no "if the file has changed" condition here, other than maybe the fact that we're here at all. Remove, I think,

lib/db/set_test.go
```diff
@@ -659,13 +659,13 @@ func TestCommitted(t *testing.T) {
...
-	// Updating with something identical should not do anything
+	// Updating with something identical should also increase the counter
 
 	s.Update(protocol.LocalDeviceID, local)
 
 	c2 := ldb.Committed()
-	if c2 > c1 {
-		t.Errorf("replace with same contents should do nothing but %d > %d", c2, c1)
+	if c2 <= c1 {
+		t.Errorf("replace with same contents should do increase the counter but %d != %d", c2, c1)
```
calmh: Broken sentence structure and this test is now mostly useless; I'd delete everything after line 661. This used to test the if statement you removed, because historically we sent unnecessary updates for ignored files or something and these were filtered out by that one.

lib/model/folder.go
```diff
@@ -51,3 +62,105 @@ func (f *folder) scanSubdirsIfHealthy(subDirs []string) error {
...
+			}
+		} else if gerr != nil {
+			l.Debugln("globbing for conflicts", gerr)
+		}
+	}
+	return err
+}
+
+// deletedir attempts to delete the given directory
+func (f *folder) deleteDir(path string, file protocol.FileInfo, matcher *ignore.Matcher) error {
```
calmh: "path" here should be called "folderPath" or "root" I think to avoid confusion. From the outside, `deleteDir(path, someStruct)` looks to me like it would delete `path`.

Doc comment must use same case as the method name, `deletedir` -> `deleteDir`.

lib/model/folder.go
```diff
@@ -51,3 +62,105 @@ func (f *folder) scanSubdirsIfHealthy(subDirs []string) error {
...
+		}
+		dir.Close()
+	}
+
+	return osutil.InWritableDir(os.Remove, realName)
+}
+
+// deletefile attempts to delete the given file
+// Takes sync conflict and versioning settings into account
+func (f *folder) deleteFile(path string, file protocol.FileInfo, ver versioner.Versioner, maxConflicts int) error {
```
calmh: Same comment

lib/model/folder.go
```diff
@@ -51,3 +62,105 @@ func (f *folder) scanSubdirsIfHealthy(subDirs []string) error {
...
+	var err error
+
+	realName, err := rootedJoinedPath(path, file.Name)
+	if err != nil {
+		return err
+	}
+
+	if maxConflicts > 0 {
```
calmh: Uh, this is incorrect. The original code checks whether there is a conflict between the update and the index version.

lib/model/folder.go
```diff
@@ -51,3 +62,105 @@ func (f *folder) scanSubdirsIfHealthy(subDirs []string) error {
 	}
 	return nil
 }
+
+// A function to provide the ability to validate and modify local changes,
+// before they are committed to the database
+// Default behavior is to apply the changes as-is, overwrite function as needed
+func (f *folder) validateAndUpdateLocalChanges(fs []protocol.FileInfo) []protocol.FileInfo {
```
calmh: I'm not super keen on this method. It gets called from `model` (where it used to call `updateLocals`) then does something (potentially) and calls back into the model while returning the  same thing it was passed, maybe. Would this not be cleaner as just a `filterLocalChanges()` with the same signature, that just modified the list of changes as appropriate and then let the model do `updateLocals` as usual? Easier to test too.

lib/model/model.go
```diff
@@ -1567,12 +1568,15 @@ func sendIndexTo(minSequence int64, conn protocol.Connection, folder string, fs
...
 func (m *Model) updateLocalsFromScanning(folder string, fs []protocol.FileInfo) {
-	m.updateLocals(folder, fs)
-
 	m.fmut.RLock()
 	folderCfg := m.folderCfgs[folder]
+	runner, _ := m.folderRunners[folder]
 	m.fmut.RUnlock()
 
+	// Validate all local changes and update the database
+	fs = runner.validateAndUpdateLocalChanges(fs)
```
calmh: As commented above, I think this should just be a filter method called before updateLocal gets called from here.

lib/model/receiveonlyfolder.go
```diff
@@ -0,0 +1,158 @@
+// Copyright (C) 2014 The Syncthing Authors.
```
calmh: 2017

lib/model/receiveonlyfolder.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/syncthing/syncthing/lib/config"
+	"github.com/syncthing/syncthing/lib/events"
+	"github.com/syncthing/syncthing/lib/fs"
+	"github.com/syncthing/syncthing/lib/protocol"
+	"github.com/syncthing/syncthing/lib/sync"
+	"github.com/syncthing/syncthing/lib/versioner"
+)
+
+type receiveOnlyFolder struct {
+	// WO folders are really just RW folders where we reject local changes...
```
calmh: Terminology - WO/RW/send-only/receive-only

lib/model/receiveonlyfolder.go
```diff
@@ -0,0 +1,158 @@
...
+					dirDeletions = append(dirDeletions, file)
+				} else {
+					fileDeletions = append(fileDeletions, file)
+				}
+			} else {
+				correctiveAction = "none"
+			}
+		}
+		// Let's update the record to reflec that this is invalid and should be pulled again if possible
+		fs[i].Deleted = false
```
calmh: As discussed in the first comment, this is too destructive for me. I'm fine with the `resync` corrective action, but not as a default that cannot be disabled.

lib/model/rwfolder.go
```diff
@@ -847,18 +835,7 @@ func (f *sendReceiveFolder) deleteFile(file protocol.FileInfo) {
 		return
 	}
 
-	cur, ok := f.model.CurrentFolderFile(f.folderID, file.Name)
-	if ok && f.inConflict(cur.Version, file.Version) {
```
calmh: As noted in `deleteFile`, conflict check needs to be moved there or happen here and the result passed on.

lib/model/receiveonlyfolder.go
```diff
@@ -0,0 +1,158 @@
...
+				if file.IsDirectory() {
+					dirDeletions = append(dirDeletions, file)
+				} else {
+					fileDeletions = append(fileDeletions, file)
+				}
+			} else {
+				correctiveAction = "none"
+			}
+		}
+		// Let's update the record to reflec that this is invalid and should be pulled again if possible
```
calmh: "reflect"

lib/model/receiveonlyfolder.go
AudriusButkevicius: I am sure this is a feature people will want, so I suspect we need another flag.

calmh: I'm not convinced of the need for an "override all my changes, except for new files" mode, but I don't care strongly about that. By all means introduce whatever flags necessary to create such a mode if it's desired. But I do care very much about the default mode not undoing my changes, regardless if they are to existing files or new ones.

AudriusButkevicius: Well I suspect the functionality people want is revert everything and delete everything extra.

calmh: Yeah that's what I would expect as well, I just want that to be turned off by default and enabled by something like `revertLocalChanges`.

AudriusButkevicius: I think that was the intent of the flag, but perhaps the implementation is not fully ironed out.

lib/model/rwfolder.go
Smiley73: Good catch. I must have screwed that up when I was copy'n'pasting things around. I added this to the deleteFile.

Smiley73:
I've been fighting with getting the requested `revertLocalChanges ` to work. Unfortunately setting a file to invalid, makes it show up during the next `withNeed` run. It doesn't seem to find the file entry in the db and thus sets `have = false`,  causing it to get pulled again.

lib/model/model.go
Smiley73: I did this, because I need to a `IndexUpdated()` to force the modified files being downloaded again after the database gets updated. I wanted to avoid to executions of `updateLocals`.

lib/model/folder.go
```diff
@@ -141,19 +141,21 @@ func (f *folder) deleteDir(path string, file protocol.FileInfo, matcher *ignore.
...
 	if err != nil {
 		return err
 	}
 
-	if maxConflicts > 0 {
+	cur, ok := f.model.CurrentFolderFile(f.folderID, file.Name)
+	if ok && f.inConflict(cur.Version, file.Version) && maxConflicts > 0 {
 		// There is a conflict here. Move the file to a conflict copy instead
 		// of deleting.
+		file.Version = file.Version.Merge(cur.Version)
```
calmh: This unfortunately doesn't accomplish what it needs to, given that this `file` doesn't get passed into the database in the end. That'll leave other devices without the conflict resolved, causing another conflict when they get our change.

lib/model/folder.go
```diff
@@ -51,3 +62,123 @@ func (f *folder) scanSubdirsIfHealthy(subDirs []string) error {
...
+	realName, err := rootedJoinedPath(folderPath, file.Name)
+	if err != nil {
+		return err
+	}
+
+	cur, ok := f.model.CurrentFolderFile(f.folderID, file.Name)
+	if ok && f.inConflict(cur.Version, file.Version) && maxConflicts > 0 {
+		// There is a conflict here. Move the file to a conflict copy instead
+		// of deleting.
+		file.Version = file.Version.Merge(cur.Version)
```
calmh: This unfortunately doesn't accomplish what it needs to, given that this file doesn't get passed into the database in the end. That'll leave other devices without the conflict resolved, causing another conflict when they get our change.

calmh:
Maybe if you don't mess with the version vector, we could see that we have a file that is newer than the global version (so no need to pull anything) but it's invalid (so no-one else should adopt it).

Smiley73:
>Maybe if you don't mess with the version vector, we could see that we have a file that is newer than the global version (so no need to pull anything) but it's invalid (so no-one else should adopt it).
I tried leaving the vector alone, setting the previous one, and emptying it. It's always the same end-result. After the invalid flag is set, it always shows up under `withneed` with `have = false`.

lib/model/folder.go
Smiley73: We never send changes to other devices, so this won't be an issue.
It's a receive-only folder after all. ;-)

Smiley73:
@calmh 
The culprit is the `updateFiles` function.
It has 2 sections, where it removes the invalid file from global. This causes it to be download again since it can't find it and sets`have=false`.

Any concerns with messing with this portion of `updateFiles`? Other devices should not download the file as long as the invalid flag is set.
`if f.IsInvalid() {`
` t.removeFromGlobal(folder, device, name, globalSize)`
`} else {`
` t.updateGlobal(folder, device, f, globalSize)`
`}`

As soon as I comment the above out and only keep ` t.updateGlobal(folder, device, f, globalSize)`, everything works as expected.

calmh:
However this means that a file announced by someone else with the invalid flag set is eligible for promotion to the "global" version, and everyone will be out of sync because they want a file they can't have. I don't know without looking if we have any paths today that results in files with the invalid bit being announced. I would expect recieve-only devices to announce their locally updated files with the invalid bit set, or it won't be visible from the outside that they are out of sync?

Smiley73:
I believe I got it now working the way you described. The test cases `TestInvalidAvailability` and `TestGlobalNeedWithInvalid` in lib/db are now failing, since invalid files get promoted to global. I didn't want to change or remove them, until we're in agreement on how it should behave.

> I don't know without looking if we have any paths today that results in files with the invalid bit being announced.
You do not. The two test cases mentioned above make sure of that. 

The latest changes currently live in another [branch](https://github.com/Smiley73/syncthing/compare/pullonly...Smiley73:disablerevert?expand=1). I can merge them and update the PR if you think the behavior is what we want.

calmh:
I guess

> I would expect recieve-only devices to announce their locally updated files with the invalid bit set, or it won't be visible from the outside that they are out of sync?

remains a question. That is, when running in non-undo-immediately mode, are local changes exposed to the outside? Or does it still look 100% in sync even when it's not? I think the latter is not desirable?

Smiley73:
Changes are exposed and it showed 99% during my testing. I can merge the changes if you want to try it out. Might have to wait until the weekend. Not sure when I get time to fix the db test cases.

calmh:
Cool, please do that. Worry less about fixing up old test cases - those can be hairy, and I can help out if needed.

Smiley73:
The changes have been merged to this PR.

Smiley73:
I got some time this Saturday to tinker a bit more. Any chance to get you to take another peak at this?

I was also wondering if I should add a feature, so that it rechecks all invalid files on the folder startup. This way we'd handle changed settings better.

Smiley73:
I was bored... It now forces a rescan of each invalid file on folder startup. This way, changed settings will actually take effect on existing files.
Before this change, it would never have deleted a locally added file after the 'delete local changes' setting was enabled.

Smiley73:
@calmh just getting this back up on your radar.

lib/model/folder.go
calmh: This code runs for every folder type though, not just for receive only folders?

calmh:

builds have expired

lib/model/folder.go
Smiley73: Yes. This is the code I moved from (f *sendReceiveFolder) deleteFile to (f *folder) deleteFile. It should be identical to what was in the original function.

calmh: Yes. But the problem is it modifies a copy of the `file`, which has no effect.

Smiley73:
The issue with the db updates in deleteFile is resolved now.
I took this opportunity and did some more cleanup. I moved more into the folder struct itself (like the FolderConfiguration and a couple other things).

Smiley73:
I removed the TestInvalidAvailability test case, since it's now obsolete based on the new behavior.
TestGlobalNeedWithInvalid I need help with. I don't completely understand how this works.

Smiley73:
@calmh something isn't right with the Jenkins slaves.
The Linux build fails with a race condition. It might be because of go 1.8.0-rc2. It seems to work fine on the windows build (1.7.4) and solaris (1.8.0-rc3) slaves. I tested at home on both OSX and Windows and can't reproduce the issue (don't have rc2, just rc3).
The Windows and Solaris slaves also don't seem to be using the latest version of the pull request. Those builds still fail on a test case I actually removed.

calmh:
The data race warning is correct, you're reading the `done` variable from a background goroutine at line 101 (rather, the fakeConnection does when called by the model) and writing to it concurrently at line 133.

lib/db/set_test.go
```diff
@@ -658,15 +620,6 @@ func TestCommitted(t *testing.T) {
...
 	}
-
-	// Updating with something identical should not do anything
-
-	s.Update(protocol.LocalDeviceID, local)
-
-	c2 := ldb.Committed()
-	if c2 > c1 {
-		t.Errorf("replace with same contents should do nothing but %d > %d", c2, c1)
-	}
```
AudriusButkevicius: Why is this going away?

lib/model/rwfolder.go
```diff
@@ -1653,47 +1600,7 @@ func removeAvailability(availabilities []Availability, availability Availability
...
-					l.Debugln(f, "removing extra conflict", gerr)
-				}
-			}
-		} else if gerr != nil {
-			l.Debugln(f, "globbing for conflicts", gerr)
-		}
-	}
-	return err
+	return f.folder.moveForConflict(name, f.MaxConflicts)
 }
```
AudriusButkevicius: You don't need this, f.moveForConflict = f.folder.moveForConflict as folder is an unnamed embedded.

lib/model/rwfolder.go
```diff
@@ -789,21 +785,8 @@ func (f *sendReceiveFolder) deleteDir(file protocol.FileInfo, matcher *ignore.Ma
...
-		for _, dirFile := range files {
-			fullDirFile := filepath.Join(file.Name, dirFile)
-			if ignore.IsTemporary(dirFile) || (matcher != nil &&
-				matcher.Match(fullDirFile).IsDeletable()) {
-				os.RemoveAll(filepath.Join(f.dir, fullDirFile))
-			}
-		}
-		dir.Close()
-	}
+	err = f.folder.deleteDir(f.dir, file, matcher)
```
AudriusButkevicius: why not just f.deleteDir? 
I think it's fine to emit delete events for the receiveonly folder?

lib/db/set_test.go
calmh: It tests something we don't do any more and I think there was some reasonable behavior change around this in the database now.

lib/model/rwfolder.go
Smiley73: I completely removed the moveForConflict in rwfolder.go. With the other changes I introduced to folder.go I don't need this clutch anymore to pass MaxConflicts.

lib/model/rwfolder.go
Smiley73: I didn't want to emit an event that matches the "normal" folder sync operations.
Within validateAndUpdateLocalChanges it already emits an event "LocalChangeRejected".

Smiley73:
The race condition is fixed and I incorporated the other feedback.
TestGlobalNeedWithInvalid  is the only test failing now. I'll need help with that one.

AudriusButkevicius:
Does it otherwise function as expected?

Smiley73:
I believe so. I added a whole bunch of new test cases to validate all kinds of scenarios, in addition to all the manual testing I did previously.
I also added all the additional features that were requested. I'll be updating the docs this morning.

AudriusButkevicius:
So this is the part of code where parts of it are mysterious to me, so sadly @calmh will have to help you with the test case.

calmh:
Yep, acknowledged. I'll take a look at this and the other things tomorrow, barring regular work day disasters preventing it.

calmh:
I did some testing and noticed a few things I think are confusing. Some may be by design, others I think not.

Assume two devices sharing a folder, "s1" is in normal mode and "s2" is receive only, and the "revert local changes" mode is off.

- Creating a conflict on s2, by creating a file on s2 and then creating the same file on s1, makes the conflict file sync back to s1. This may be intended but makes it seem not "receive only".

- Likewise deleting the conflict file on s2 now deletes it from s1. This is consistent with the previous point though, conflict files being exempt from receive-only status.

- Deleting files on s2 is weird. The delete is propagated to s1 but not acted upon, yet no-one considers the other out of sync. For example, if there were 200 files before and everyone was in sync (local=200, global=200 on both sides), and I delete a file on s2, s2 has global=199, local=199 (OK), s1 gets global=199, local=200, still "up to date" (odd) and no out of sync indication for s1 (also odd - either of the two should be out of sync now). I guess this might come from sending delete entries with the invalid bit set, which is then the newest version when seen by s1 but invalid so it can't be installed as as the "global". I think this would mean that another normal device s3 (for example) would not sync these files from s1, even if it already didn't have them.

- Modifying files on s2 has about the same effect - they effectively disappear from the global index on s1.

Now I'm turning on "revert local changes" (but not "delete local changes").

- Previously deleted files are not restored. I guess this is expected as the handling of them differs at *scan* time? But it's still a bit odd from a user point of view, because now that folder is supposedly "in sync" (according to the GUI) but I have local changes that are not undone, which the config indicates they should have been.

- Creating a file on s2 and renaming some on s1 to the same name does not create a conflict on s2, it just overwrites the file just created. I'm not 100% sure we do this correctly in the normal case either though - it's more difficult to trigger there.

Turning the folder on s2 back to "send & receive".

- Previously ignored changes remain ignored - that is, s1 and s2 show different counts and different contents, yet both consider the other "up to date".

Smiley73:
> Turning the folder on s2 back to "send & receive".
Previously ignored changes remain ignored - that is, s1 and s2 show different counts and different contents, yet both consider the other "up to date".

I added f.resetInvalidFiles() to the other folder types. It'll get executed on folder initialization. Of course this means the local changes will immediately get push to the other devices.

The rest will take me a bit longer....

Smiley73:
> Creating a conflict on s2, by creating a file on s2 and then creating the same file on s1, makes the conflict file sync back to s1. This may be intended but makes it seem not "receive only".
Likewise deleting the conflict file on s2 now deletes it from s1. This is consistent with the previous point though, conflict files being exempt from receive-only status.

Both of these are fixed now.

Smiley73:
@calmh 
I think I got all the problems tackled now. Deleted files are also properly restored if you turn on "revert local changes" at a later point.
Deleted files now also get restored if you switch the folder type to "send & receive". Deletion will not get propagated. 
Switching to "send only" will obviously not restore the files, but also not propagate the change. I could force that the deletions get sent out to other devices, but wanted to check first if that's what we really want.

Smiley73:
@calmh getting this back up on your radar.

calmh:
Taking this for another spin. I still think the in sync reporting needs some work. For example, I spin up two devices with a folder, one normal ("s1") and one receive-only ("s2"). There are 20 files, in sync, and I change two of them on the receive-only side. It correctly logs "ignoring local change" and does in fact to do so, so the operation works. But, "normal" side:

screen shot 2017-03-11 at 08 27 00

Receive-only side:

screen shot 2017-03-11 at 08 27 55

At a glance, everything is perfectly in sync. However, the normal side has subtracted two files from the "global" state which looks odd. It also thinks s2 is in sync. S2 thinks it's in sync as well. Somewhere here there should be an indication that s2 is not in phase with the desired reality?

The 18/20 on the s1 sides seems mostly cosmetic; bringing up a third device correctly syncs all the 20 files, although it also displays 18/20.

Perhaps the receive only side should announce the changed file with an all zeroes version vector, in addition to the invalid bit? This would say "not only is this file invalid, it's also older than anything else in the universe".

calmh:
Feel free to fix or just delete the NeedWithInvalid test as appropriate, so we get a greener panel and builds that can be downloaded.

Smiley73:
The empty version vector triggers the file to be downloaded again from the peer. I'm using that to force the re-sync. 
I'll have to do some more digging into this. Not sure when I'll have time. This might take a bit...

Noki:
Any updates here? I could really use this feature. ;-)

lkwg82:
@Noki Are you offering help?

Smiley73:
Yeah sorry guys, but I really don't have time right now to continue this work. 
This change is "almost there", but I really could use some help to complete it. If you're waiting for me to do it, then it'll be a while...

lkwg82:
Is there somebody, who is interested and able to takeover?

lkwg82:
@Smiley73

maybe you can list the task needed to be done?
This will lower the hurdles.

Smiley73:
I updated the description and added the only thing I found still outstanding, which is:
Reporting - Folders shouldn't claim to be in-sync if one side is ignoring local changes. (see calmh's response from 2017-03-11).
Hopefully somebody else can help get this completed.

lkwg82:
I failed merging it in the web ui :(

lkwg82:
I struggle how to take over. What is the proposed procedure?

calmh:
I guess resolve the merge conflicts and move on from that point with the outstanding issue(s)?

calmh:
Closed for lack of movement, please feel free to continue and we'll reopen whenever.

ellnic:
:( Shame

PR #3783, opened 2016-12-02
lib/model: Prevent symlink traversals

calmh:
### Purpose

This adds two checks. When pulling files, we only process items that have a parent in the local index or the current pulling batch, and where that parent is a directory. When serving requests, we require that the item we are serving the request for itself exists in the local index.

In the second commit I add an LRU cache for the database check in Request(), as this otherwise kills performance for large transfers as we would do a database looking for every block. The cache holds the last few files that we've successfully verified and is purged when we update the local index. It's in a separate commit just for review purposes.

I switched out the groupcache/lru package to another LRU package that is apparently based on that one from the beginning, because it had a `Purge()` method that the first lacked. We don't need two different LRU packages.

### Testing

I've added unit tests that attempt to do the operations that this blocks. Those tests fail (i.e., the operations pass) before implementing this.

Don't mind if @Unrud takes a peek here as well...


calmh:
Ah, this has one more race condition I need to take care of as well...

lib/model/requests_test.go
```diff
@@ -62,6 +63,11 @@ func TestRequestSimple(t *testing.T) {
 func TestSymlinkTraversalRead(t *testing.T) {
 	// Verify that a symlink can not be traversed for reading.
 
+	if runtime.GOOS == "windows" {
```
AudriusButkevicius: could check symlinks.Supported.

lib/model/requests_test.go
```diff
@@ -57,6 +60,192 @@ func TestRequestSimple(t *testing.T) {
 	}
 }
 
+func TestSymlinkTraversalRead(t *testing.T) {
+	// Verify that a symlink can not be traversed for reading.
+
+	if runtime.GOOS == "windows" {
```
AudriusButkevicius: should check symlinks.Supported

lib/model/requests_test.go
```diff
@@ -57,6 +60,192 @@ func TestRequestSimple(t *testing.T) {
...
+	fc.addFile("symlink", 0644, protocol.FileInfoTypeSymlinkDirectory, contents)
+	fc.sendIndexUpdate()
+	<-done
+
+	// Request a file by traversing the symlink
+	buf := make([]byte, 10)
+	err := m.Request(device1, "default", "symlink/requests_test.go", 0, nil, false, buf)
+	if err == nil || !bytes.Equal(buf, make([]byte, 10)) {
+		t.Error("Managed to traverse symlink")
+	}
```
AudriusButkevicius: I'd add more cases apart from ../, some of the foo/../../foo stuff etc.

lib/model/rwfolder.go
```diff
@@ -458,6 +494,11 @@ func (f *rwFolder) pullerIteration(ignores *ignore.Matcher) int {
 			}
 		}
 
+		if file.IsDirectory() {
+			// Mark the directory we just processed as cleared for having
+			// files inside it.
+		}
```
AudriusButkevicius: if true then nothing?

calmh: Woops, fixed

lib/model/requests_test.go
calmh: No point I think, that's just the contents of the symlink. It can point to `.` or to the moon, we should never traverse it regardless.

cmd/stdiscosrv/querysrv.go
```diff
@@ -18,7 +18,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/golang/groupcache/lru"
+	lru "github.com/hashicorp/golang-lru"
```
calmh: This is however how `goimports` inserts it when it adds the import, as the package name is different from the import path... I have no strong opinion either way, really.

lib/model/requests_test.go
AudriusButkevicius: Agree.

Unrud:
Disclaimer: I didn't test any of this, just the first ideas that came to my mind.

  * Temporary file names are not filtered from the Index. Consequently it would be possible to create a Symlink like ``.syncthing.tmp.foo -> /home/user/bashrc``. If in the next step an IndexUpdate with the file "foo" gets received, the fake temporary file will be opened and written to. (It would require a bit more effort than this, because temporary files are deleted if no blocks match.)

  * Send an Index with the directories "foo" and "foo/bin". Send an IndexUpdate deleting "foo". Send an IndexUpdate with the Symlink ``foo -> /home/user``. Send an IndexUpdate with the file "foo/bin/echo".
    Hopefully the entry "foo/bin" is still in the database and the checks pass.
    **Edit:** I realize that this doesn't work because only empty (besides temporary files) folders are deleted. I guess on Windows it would be possible to abuse some name collisions, e.g. deleting foo/Bin. Another possibility would be tricking the user into deleting it manually or even just using a temporary name for the inner folder.

  * This is not protected against local changes. On a multi-user system someone might has write access to a folder that's synced by Syncthing (running under a different user account). It's easy to create a folder structure and wait until it's scanned. In the next step replace the folders by a Symlink and send an IndexUpdate, before the local change is scanned.

  * scanner/walk.go contains race conditions. Stat is called on a file and it's type is checked, further processing follows. There is no guarantee that the file type doesn't change between the call to Stat and the next steps. It's possible that Symlinks are followed and the files get inserted into the database.
    I guess it's not possible to exploit remotely, but local changes can cause this even accidentally.

A solution would be to re-implement all the file system functions (Open, Delete, Rename, Stat...) in a "symlink race"-safe way. In Unix this could be done by opening the folder directory, using openat with the O_NOFOLLOW flag to reach the parent directory (one step at a time) and use openat/unlinkat/renameat/fstatat/... to perform the operation.
I think that it's impossible to handle Symlinks securely on Windows? (They can only created by the administrators anyway.)

calmh:
I'll consider some of the above points a bit more, but in general I think we are not trying to protect against someone fooling Syncthing by racing changes on the local filesystem. If the attacker has write access to the local folder but different access permissions than the Syncthing user there will always be funny business they can do - for example hard link to a file that Syncthing can read but they cannot, and have it sync that file to somewhere else.

Syncthing should be safe against whatever someone else speaking to it over the network comes up with, though.

Unrud:
> If the attacker has write access to the local folder but different access permissions than the Syncthing user there will always be funny business they can do

Not if Syncthing behaves correctly. Linux/MacOS provide the means to avoid races on the filesystem.

It seems like Symlinks can be used by ordinary users on Windows soon, that will be interesting:
https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/

> for example hard link to a file that Syncthing can read but they cannot, and have it sync that file to somewhere else.

You need write access to a file if you want to create a hard link to it.

The current solution relies on the contents of the database always being consistent and in sync with the file system. Additionally bad things can happen if the file system changes during a scan. There is just a lot that can potentially go wrong and it's difficult to verify.

calmh:
> Not if Syncthing behaves correctly. Linux/MacOS provide the means to avoid races on the filesystem.

I wonder if this is supported well enough on the other platforms we support, as well. (By Go and so on, if nothing else.) If so, we should do that and save us some pain.

> You need write access to a file if you want to create a hard link to it.

This is not the case on my Mac, at least.

calmh:
We can get the O_NOFOLLOW flag where supported and pass it to os.OpenFile (I think), but that requires C and kills cross compilation...

```go
package main

// #include <fcntl.h>
// int nofollow() {
//    return O_NOFOLLOW;
// }
import "C"
import "fmt"

func main() {
	fmt.Printf("%x\n", C.nofollow())
}
```

Unrud:
> I wonder if this is supported well enough on the other platforms we support, as well

Obviously it's not supported on Windows. A platform-independent solution can protect against remote attacks and make local attacks very difficult (because the time window is small). Special implementations for Linux/MacOS can be added later.

``fs.BasicFilesystem`` seems like a good starting point. Maybe pass the folder path to the constructor and have one Filesystem per folder. You have to check that the sub-paths don't contain Symlins with ``Lstat`` before every operation.
E.g. for ``Open("foo/bar")`` the paths ``/home/user/Sync/foo`` and ``/home/user/Sync/foo/bar`` must be checked for Symlinks by ``Lstat`` before ``os.Open`` is called.
E.g. for ``Remove("foo/bar")`` the path ``/home/user/Sync/foo`` must be checked for Symlink before ``os.Remove`` is called.
Additionally a RW-Lock is required to make sure that dangerous operations are not executed at the same time. It should be enough if ``Rename`` and ``CreateSymlink`` are exclusive, all other operations can be executed safely at the same time.

This should not hurt performance. The slow part is loading blocks from the disk to RAM. It doesn't really matter if that cached data is only used for e.g. ``Open`` or additionally for some ``Lstat`` calls.

> Edit: ``syscall.Openat`` only exists for Linux, in Go. :(

It's also missing ``fstatat`` and some flags.

> This is not the case on my Mac, at least.

Sorry, I only checked Linux. On BSD it's the same as MacOS. You can still prevent access to your files by not granting the execute right on directories.

PR #3792, opened 2016-12-09
lib/model, lib/protocol, lib/scanner: Include symlink target in index, pull symlinks synchronously

calmh:
### Purpose

Handle symlinks like we handle directories and deletions - synchronously before files.

Adds a `symlinkTarget` field to the `FileInfo`, with ID 17 after `blocks` as it will very seldom be set. This is used when creating symlinks. Symlink handling itself is somewhat simplified, removing the `ChangeType` path by always simply recreating it instead.

### Testing

The scanner unit test verifies that the destination gets set.

This passes the symlink-related integration tests.

I've tested manually against an older version of Syncthing, where symlinks from the newer side gets synced to the older (using the old method) and symlinks from the older to the newer error out. The error message says to rescan with a newer Syncthing on the source, but this should be fairly rare as it will only show up on installations where there is a version mismatch or symlinks where out of sync when the upgrade happened.

### Documentation

I'll add this to the spec when merged.


lib/db/structs.proto
```diff
@@ -33,4 +33,5 @@ message FileInfoTruncated {
     bool                  no_permissions = 8;
     protocol.Vector       version        = 9 [(gogoproto.nullable) = false];
     int64                 sequence       = 10;
+    string                symlinkTarget  = 17;
```
AudriusButkevicius: Proto uses snake case.

lib/protocol/bep.proto
```diff
@@ -106,7 +106,8 @@ message FileInfo {
     Vector       version        = 9 [(gogoproto.nullable) = false];
     int64        sequence       = 10;
 
-    repeated BlockInfo Blocks = 16 [(gogoproto.nullable) = false];
+    repeated BlockInfo Blocks        = 16 [(gogoproto.nullable) = false];
+    string             symlinkTarget = 17;
```
AudriusButkevicius: Snake case

lib/protocol/bep_extensions.go
```diff
@@ -30,8 +30,8 @@ func (m Hello) Magic() uint32 {
 }
 
 func (f FileInfo) String() string {
-	return fmt.Sprintf("File{Name:%q, Type:%v, Sequence:%d, Permissions:0%o, ModTime:%v, Version:%v, Length:%d, Deleted:%v, Invalid:%v, NoPermissions:%v, Blocks:%v}",
-		f.Name, f.Type, f.Sequence, f.Permissions, f.ModTime(), f.Version, f.Size, f.Deleted, f.Invalid, f.NoPermissions, f.Blocks)
+	return fmt.Sprintf("File{Name:%q, Type:%v, Sequence:%d, Permissions:0%o, ModTime:%v, Version:%v, Length:%d, Deleted:%v, Invalid:%v, NoPermissions:%v, Blocks:%v, SymlinkTarget:%q}",
+		f.Name, f.Type, f.Sequence, f.Permissions, f.ModTime(), f.Version, f.Size, f.Deleted, f.Invalid, f.NoPermissions, f.Blocks, f.SymlinkTarget)
 }
```
AudriusButkevicius: Lets make the return value type conditional.

AudriusButkevicius:
Maybe we should just kill the blocks bit on the new version, and just say that its broken.

calmh:
Yes. It is not used by any of the new symlink code, so no need to maintain it.

lib/protocol/bep_extensions.go
```diff
@@ -30,8 +30,19 @@ func (m Hello) Magic() uint32 {
...
 
 func (f FileInfo) String() string {
-	return fmt.Sprintf("File{Name:%q, Type:%v, Sequence:%d, Permissions:0%o, ModTime:%v, Version:%v, Length:%d, Deleted:%v, Invalid:%v, NoPermissions:%v, Blocks:%v, SymlinkTarget:%q}",
-		f.Name, f.Type, f.Sequence, f.Permissions, f.ModTime(), f.Version, f.Size, f.Deleted, f.Invalid, f.NoPermissions, f.Blocks, f.SymlinkTarget)
+	switch f.Type {
+	case FileInfoTypeDirectory:
+		return fmt.Sprintf("Directory{Name:%q, Sequence:%d, Permissions:0%o, ModTime:%v, Version:%v, Deleted:%v, Invalid:%v, NoPermissions:%v}",
+			f.Name, f.Sequence, f.Permissions, f.ModTime(), f.Version, f.Deleted, f.Invalid, f.NoPermissions)
```
calmh: ACK. We don't even set those fields. Keeping NoPermissions, which is always true, for clarity.

AudriusButkevicius:
I don't think we use mtime for folders either, do we?

calmh:
We don't set it, but the index entry still has one.

lib/model/rwfolder.go
```diff
@@ -666,6 +674,77 @@ func (f *rwFolder) handleDir(file protocol.FileInfo) {
...
+		events.Default.Log(events.ItemFinished, map[string]interface{}{
+			"folder": f.folderID,
+			"item":   file.Name,
+			"error":  events.Error(err),
+			"type":   "symlink",
+			"action": "update",
+		})
+	}()
+
+	realName, err := rootedJoinedPath(f.dir, file.Name)
```
AudriusButkevicius: Is var err above actually non nil here?

AudriusButkevicius: Asking as I am on the phone and don't recall how shadowing works in this case.

AudriusButkevicius:
Apart from thatquestion, lgrm

lib/model/rwfolder.go
calmh: Oh. Yes, that actually works, although it's quite non-obvious and not really something I'd recommend, even though it's probably me who wrote it... Since err is already declared in the same scope, the `:=` does nothing on it.

calmh: That defer construction is still broken though, because we have `if _, err := ...` below which *does* declare a new err. Hang on and I'll fix it.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 7b07ed6580e9d39ffbb91a7cb8e0946a88ded830. Thanks, @calmh!

AudriusButkevicius:
Btw this still needs an evalsymlinks wedged in somewhere, I assume its coming as a separate pr

Issue #2679, opened 2016-01-09
Clean up "master" terminology


calmh:
I.e., the concept of a folder that does not accept changes from the other devices in the cluster. Started as "read only" but no-one understood what is was read only against. Became "master" as in "master copy" or "golden master", but is still confusing for many. The config option is still called "readOnly". In discussions about folders that act the opposite way, "slave" is invariably proposed.

I suggest that we should clean all this up to a terminology centered around "changes" (or similar) which is what happens to files or directories, and is what we synchronize between devices. We then have folders which "send changes" and "receive changes". The current "Master" checkbox then becomes "Don't receive changes", or possibly "Only send changes" (but the "only" is ambiguous there), with a help text like "Don't receive changes to files or folders sent by other devices. Changes made locally are always sent." or so. The config option would be "sendChanges", true by default.

The "Override" button would become "Overwrite remote changes", "Force send changes" or something similar.

Possible new folder modes would be expressed in the same terms, e.g. "Don't send changes" for something where we expect no local changes to happen.

Feel free to bikeshed.

This changes our external API so would be a major version change.

This ticket obsoletes #1731.

generalmanager:
I think this is the only reasonable way to phrase this. Especially with additional features like "don't accept/receive deletes" the current phrasing would be very ambiguous and lead to even more support requests based on misunderstandings.

CoupeWare:
I've always wondered what happens if you have 2 master folders. I've been meaning to test this. Based on what you have written here I suspect they "Don't receive changes" from each other.

I use master folders when I'm using Syncthing for backup. In most cases the master syncs to 2 or more backup devices located in different places. I think of the backup folders as "Archive Folders" rather than the source folder as a "Master Folder". One thing I would find very useful is an option to have normal folders on more than one source device that sync with each other in read/write mode and one or more devices with the folders set as "Archive Folders" which get changes from both source devices and never send changes. The downside to this is that multiple Archive devices would not be able to share with each other and keep each other up to date when the source devices are offline.

As things are I would like to be able to override changes on the Archive devices. I don't always have access to the source devices but I can always get to the various Archive devices.

BethAr:
Here is a suggestion:

Text on folder --> "Master Out of Sync" / "Slaves Out of Sync"
Yellow Button --> "Override Slaves" (no confirmation)
Red Button --> "Override Master" (open confirmation dialog)

Tralen:
[...]

PR #3793, opened 2016-12-09
Rename folder types

Smiley73:
### Purpose

Rename the folder types to "Send & Receive" and "Send Only".

### Testing

Compiled and executed test suite. 
Executed bin/syncthing with v17 configuration. Verified change of folder type string to new value.

### Screenshots

N/A

### Documentation

Folder type configuration name changed. Documentation update still to be done.

### Authorship

Heiko Zuerker


lib/config/foldertype.go
```diff
@@ -9,15 +9,15 @@ package config
...
+	FolderTypeSendOnly
 )
 
 func (t FolderType) String() string {
 	switch t {
-	case FolderTypeReadWrite:
+	case FolderTypeSendReceive:
 		return "readwrite"
-	case FolderTypeReadOnly:
+	case FolderTypeSendOnly:
```
AudriusButkevicius: I though we willbe migrating the configs?

lib/model/model.go
```diff
@@ -1949,7 +1949,7 @@ func (m *Model) generateClusterConfig(device protocol.DeviceID) protocol.Cluster
 		protocolFolder := protocol.Folder{
 			ID:                 folder,
 			Label:              folderCfg.Label,
-			ReadOnly:           folderCfg.Type == config.FolderTypeReadOnly,
+			ReadOnly:           folderCfg.Type == config.FolderTypeSendOnly,
```
AudriusButkevicius: Tbh this should become a protocol defined thing.

lib/config/foldertype.go
Smiley73: Definitely. I just haven't figured out yet how the API works, so I didn't want to touch it yet.

lib/model/model_test.go
```diff
@@ -1779,7 +1779,7 @@ func TestIssue3164(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	fl := rwFolder{
+	fl := sendreceiveFolder{
```
AudriusButkevicius: sendReceive...

lib/model/rofolder.go
```diff
@@ -15,15 +15,15 @@ import (
 )
 
 func init() {
-	folderFactories[config.FolderTypeReadOnly] = newROFolder
+	folderFactories[config.FolderTypeSendOnly] = newSendOnlyFolder
 }
 
-type roFolder struct {
+type sendonlyFolder struct {
```
AudriusButkevicius: sendOnly

lib/model/rwfolder_test.go
```diff
@@ -66,8 +66,8 @@ func setUpModel(file protocol.FileInfo) *Model {
 	return model
 }
 
-func setUpRwFolder(model *Model) rwFolder {
-	return rwFolder{
+func setUpSendreceiveFolder(model *Model) sendreceiveFolder {
```
AudriusButkevicius: SendReceive

lib/model/model_test.go
Smiley73: Should I rename all sendreceiveFolder to sendReceiveFolder, even in rwfolder.go ?

lib/model/model.go
Smiley73: I'm unclear what you want me to do.

Smiley73:
I've been trying to decipher the API and what I have to change there.
According to the documentation, I believe I only have to worry about these 2:
GET /rest/system/config
POST /rest/system/config

Is that correct?

AudriusButkevicius:
Well thats what we'll break, but we will probably step on our own toes reading old config.xml off disk.

Smiley73:
I could add code to replace sendreceive with readwrite and sendonly with readonly within the API calls. This way we could switch to using the new type name in the config.

Whenever you guys release the next major version, then you could take that clutch out.

lib/model/model.go
calmh: I guess so, yes. However that does not need to be part of *this* PR, as it does not introduce another folder type.

calmh:
@st-jenkins add to white list

cmd/syncthing/usage_report.go
```diff
@@ -134,7 +134,7 @@ func reportData(cfg configIntf, m modelIntf) map[string]interface{} {
 	for _, cfg := range cfg.Folders() {
 		rescanIntvs = append(rescanIntvs, cfg.RescanIntervalS)
 
-		if cfg.Type == config.FolderTypeReadOnly {
+		if cfg.Type == config.FolderTypeSendOnly {
 			folderUses["readonly"]++
```
AudriusButkevicius: We should probably update the usage server too? Or f
Just folderUses[type]++

lib/config/config.go
```diff
@@ -325,6 +328,10 @@ func convertV14V15(cfg *Configuration) {
 	cfg.Version = 15
 }
 
+func convertV17V18(cfg *Configuration) {
+	cfg.Version = 18
+}
```
AudriusButkevicius: This will probably need a magic struct that maps the same xml element into a string and then perform substitution somehow, without breaking the Unmarshal on the custom type.

lib/config/foldertype.go
```diff
@@ -30,12 +30,12 @@ func (t FolderType) MarshalText() ([]byte, error) {
...
 func (t *FolderType) UnmarshalText(bs []byte) error {
 	switch string(bs) {
-	case "readwrite":
-		*t = FolderTypeReadWrite
-	case "readonly":
-		*t = FolderTypeReadOnly
+	case "readwrite", "sendreceive":
+		*t = FolderTypeSendReceive
+	case "readonly", "sendonly":
+		*t = FolderTypeSendOnly
```
AudriusButkevicius: Ah ok, this might work actually, instead of the magic extra struct, the question is however when do we remove these old values, I suspect never :/ or we could remove them if we did the extra struct migration, but then the struct would have to live forever. However then the code to deal with legacy stuff is nicely living in coversion section opose an unmarshal method of a random type. @calmh any strong preferences?

calmh: My strong preference is to not change the on disk or JSON serialized config at this point. If we want to accept the synonyms as above in case someone writes them manually into the config i'm fine with that.

We will need to change the config format at the point another folder type is introduced. That will need to be a 0.15 release and it's fine to remove the old format and migrate it.

AudriusButkevicius: Fine

cmd/syncthing/usage_report.go
Smiley73: I did the change, but this would need some server-side changes also. Where's the code for that?

lib/config/foldertype.go
Smiley73: Yeap, very confusing... I'm trying to help you guys here with way more than I planned to. I really don't want to spend a lot of time and nothing comes out of it.

I thought the purpose of this PR is to do the actual folder type name changes, since they've been confusing a lot of people. I don't think there's a lot of benefit to do this only partially. 
So far I was able to do all changes so that the user doesn't even need to do anything (API still pending). Obviously there will be new names displayed in the GUI, so there is user impact.

Of course I'm not familiar with the guidelines you follow what can and can't go into a minor release.
When are you planning to release 0.15?

lib/config/foldertype.go
Smiley73: To my earlier point: If we don't make these changes, what's the benefit or purpose of this PR then?

AudriusButkevicius: Well the biggest pain point is the confusion in the UI, oppose to confusion in the API, however I do agree that we should do it in once sweep.

Smiley73:
@calmh and @AudriusButkevicius 
So what should I be doing? I have refrained from making any further modifications, until I get some marching orders.
When are you guys planning to release a 0.15?
Should I undo all the visible changes? (Which won't be of any benefit with the user confusion)
Should I add code to ensure the API compatibility with 0.14?

AudriusButkevicius:
I'm a bit lost on this one, I'd say we cut the 0.15 with this, together with any other incompatible protocol changes required to make this right, yet I agree its much harder than we though. 


Perhals we do a display value only change for now and cut the 0.15 branch with the bigger changes.
Yet lets see what @calmh thinks first.

calmh:
I'm not dropping a 0.15 for an internal cleanup change.

The purpose of this change, as I see it, is to clean up the nomenclature visible to the user. That means GUI and docs. Which is what I recommended from the start. :)

We've survived thus far with the config calling it "read only" even when the GUI calls it "master". We can survive with "read only" in the config while the GUI calls it "send only".

Smiley73:
Okay. I'll revert the config change. I'd suggest leaving the parts that will make a migration to the future config names easier. I'll also take a closer look at the docs.
Should have some time this weekend to work on it.

Smiley73:
@calmh and @AudriusButkevicius 
I believe I'm done with the code changes for this. Can you please review?
Next step would be the documentation....

AudriusButkevicius:
man pages are auto-generated, the rest looks good

Smiley73:
Ah crap. Guess I should have read the README..... reverting it right now

AudriusButkevicius:
Something wrong went with the revert.

AudriusButkevicius:
Last thing, can you add yourself to NICKS and AUTHORS in alphabetical order.

AudriusButkevicius:
@st-review test this please

st-review:
I'm sorry, @AudriusButkevicius. I'm afraid I don't know what you mean. I know how to `don't`, `lgtm`, `merge`, `prevent`, `squash`, `stop`.

AudriusButkevicius:
ghrr

AudriusButkevicius:
@st-jenkins test this please

AudriusButkevicius:
@st-review merge

lib/model: Clarify master terminology (fixes #2679)

st-review:

st-review:
:ok_hand: Merged as 398c356f22fb0e19898d498cf9235ac5f51596fa. Thanks, @Smiley73!

AudriusButkevicius:
tyvm for the patience.

PR #3798, opened 2016-12-09
lib/model, lib/osutil: Verify target directory before pulling / requesting

calmh:
### Purpose

Ensure symlinks are not followed when pulling or requesting data.

This simply does an `lstat` walk of the components from the folder dir to the target directory, ensuring that all of them are real directories. This is of course not race safe, so assumes that the user does not locally replace a directory with a symlink at the same time as a pull is in progress against that directory. We should document this as a "don't do that". It is safe against ourselves, as we do not do directory/symlink replacement concurrently with pulling files.

This is cheaper than the stdlib `filepath.EvalSymlinks` which does a lot more work to find the actual absolute path of a file (which we don't care about).

It is also not cached at all. I don't think that matters in the puller, and I'm not currently measuring any difference in the request handling either, although I'm doing a couple of more tests. If it's an issue I think we can re-introduce fd caching in the request handler.

### Testing

A couple of new tests added for the pulling stage. It passes the existing integration tests for symlinks and basic syncing.

### DIFF
--- a/lib/osutil/isdir.go
+++ b/lib/osutil/isdir.go
@@ -1,0 +1,44 @@
+package osutil
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// IsDir returns true if base and every path component of name up to and
+// including filepath.Join(base, name) is a directory (and not a symlink or
+// similar). Base and name must both be clean and name must be relative to
+// base.
+func IsDir(base, name string) bool {
+	defer func() {
+		fmt.Println(base, name)
+	}()
+	path := base
+	info, err := Lstat(path)
+	if err != nil {
+		return false
+	}
+	if !info.IsDir() {
+		return false
+	}
+
+	if name == "." {
+		// The result of calling IsDir("some/where", filepath.Dir("foo"))
+		return true
+	}
+
+	parts := strings.Split(name, string(os.PathSeparator))
+	for _, part := range parts {
+		path = filepath.Join(path, part)
+		info, err := Lstat(path)
+		if err != nil {
+			return false
+		}
+		if !info.IsDir() {
+			return false
+		}
+	}
+	return true
+}
--- a/lib/osutil/isdir.go
+++ b/lib/osutil/isdir.go
@@ -1,3 +1,9 @@
+// Copyright (C) 2016 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at http://mozilla.org/MPL/2.0/.
+
 package osutil
 
 import (
--- a/lib/osutil/isdir.go
+++ b/lib/osutil/isdir.go
@@ -38,6 +38,9 @@ func IsDir(base, name string) bool {
 		if err != nil {
 			return false
 		}
+		if info.Mode()&os.ModeSymlink != 0 {
+			return false
+		}
 		if !info.IsDir() {
 			return false
 		}
--- a/lib/osutil/isdir.go
+++ b/lib/osutil/isdir.go
@@ -1,7 +1,6 @@
 package osutil
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -12,9 +11,6 @@ import (
 // similar). Base and name must both be clean and name must be relative to
 // base.
 func IsDir(base, name string) bool {
-	defer func() {
-		fmt.Println(base, name)
-	}()
 	path := base
 	info, err := Lstat(path)
 	if err != nil {
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -120,6 +120,7 @@ var (
 	errDevicePaused        = errors.New("device is paused")
 	errDeviceIgnored       = errors.New("device is ignored")
 	errNotRelative         = errors.New("not a relative path")
+	errNotDir              = errors.New("parent is not a directory")
 )
 
 // NewModel creates and starts a new model. The model starts in read-only mode,
@@ -1113,20 +1114,15 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset 
 		return protocol.ErrNoSuchFile
 	}
 
-	if info, err := osutil.Lstat(fn); err == nil && info.Mode()&os.ModeSymlink != 0 {
-		target, _, err := symlinks.Read(fn)
-		if err != nil {
-			l.Debugln("symlinks.Read:", err)
-			if os.IsNotExist(err) {
-				return protocol.ErrNoSuchFile
-			}
-			return protocol.ErrGeneric
-		}
-		if _, err := strings.NewReader(target).ReadAt(buf, offset); err != nil {
-			l.Debugln("symlink.Reader.ReadAt", err)
-			return protocol.ErrGeneric
-		}
-		return nil
+	if !osutil.IsDir(folderPath, filepath.Dir(name)) {
+		l.Debugf("%v REQ(in) for file not in dir: %s: %q / %q o=%d s=%d", m, deviceID, folder, name, offset, len(buf))
+		return protocol.ErrNoSuchFile
+	}
+
+	if info, err := osutil.Lstat(fn); err != nil || !info.Mode().IsRegular() {
+		// Reject reads for anything that doesn't exist or is something
+		// other than a regular file.
+		return protocol.ErrNoSuchFile
 	}
 
 	// Only check temp files if the flag is set, and if we are set to advertise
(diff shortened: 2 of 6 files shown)


calmh:
I added a benchmark for just a Request() call for a block in a file a few levels deep (it's in master).

```
master:
BenchmarkRequestInSingleFile-8   	  100000	     20330 ns/op	6447.12 MB/s

this branch:
BenchmarkRequestInSingleFile-8   	   30000	     42265 ns/op	3101.19 MB/s

with my secret fd caching:
BenchmarkRequestInSingleFile-8   	  100000	     10399 ns/op	12603.61 MB/s
```

Clearly, makes fast. But also clearly this is not the bottleneck, even at 42265 ns per call, given that we have a network roundtrip on top of that.

AudriusButkevicius:
Where's the secret fd caching?

lib/osutil/isdir.go
```diff
@@ -0,0 +1,46 @@
...
+	if !info.IsDir() {
+		return false
+	}
+
+	if name == "." {
+		// The result of calling IsDir("some/where", filepath.Dir("foo"))
+		return true
+	}
+
+	parts := strings.Split(name, string(os.PathSeparator))
```
AudriusButkevicius: I think most of the paths passed into this are OS specific, so we should be fine.

lib/osutil/isdir.go
```diff
@@ -0,0 +1,46 @@
...
+	for _, part := range parts {
+		path = filepath.Join(path, part)
+		info, err := Lstat(path)
+		if err != nil {
+			return false
+		}
+		if !info.IsDir() {
```
Unrud: If the type of a Symlink on Windows is directory, ``IsDir()`` returns true. You have to check if the Symlink bit is set.

Unrud:
This does not address the temporary file problem.
An attacker can create a Symlink like ``.syncthing.foo.tmp --> ../bar`` and then send an IndexUpdate for the file **foo** containing a manipulated version of **bar**. Syncthing will write to the file ``../bar`` outside of the synced folder.

AudriusButkevicius:
There are no more temp files for symlinks.

Unrud:
> There are no more temp files for symlinks.

Right, but you can still create symlinks that are named like temp files.

Unrud:
The solution relies on ``folderFiles.WithNeed`` delivering parents before their children. Otherwise an attacker can create the folder ``foo`` and then send an Index with the file ``foo/bar`` and a Symlink ``foo -> ..``. If the FileInfo for ``foo/bar`` is handled first, the folder ``foo`` exists and the checks pass. Next the FileInfo for the Symlink is handled and ``handleSymlink`` deletes the folder ``foo`` and replaces it by the symlink.

I think that you can exploit upper and lower case letters in file names to place children before their parents on Windows.

AudriusButkevicius:
> Right, but you can create symlinks that are named like temp files. 

Not anymore, https://github.com/syncthing/syncthing/commit/7b07ed6580e9d39ffbb91a7cb8e0946a88ded830#diff-feb9fde15f8dd977fe9b279a36a447a2R680

> The solution relies on folderFiles.WithNeed delivering parents before their children.

The rest of the implementation relies on this, and I don't see a way for an attacker to exploit this.
Furthermore, there is a change in the pipeline which turns dependencies into a graph before executing, so the order becomes not important, which also addresses some weird issues where syncthing just flops.

Unrud:
> Not anymore

You send a FileInfo with the name ``.syncthing.foo.tmp`` and type Symlink. Syncthing will then create the symlink ``.syncthing.foo.tmp``. I've tested it and was able to manipulate files outside of the folder.

> and I don't see a way for an attacker to exploit this.

I've described one possibility but I may have missed something.

> Furthermore, there is a change in the pipeline which turns dependencies into a graph before executing

How does it prevent the check ``if !osutil.IsDir(f.dir, filepath.Dir(file.Name)) {`` from being circumvented by ordering the children before the parents in the database on Windows?

```go
func (db *Instance) withNeed(folder, device []byte, truncate bool, fn Iterator) {
    ...
    dbi := t.NewIterator(util.BytesPrefix(db.globalKey(folder, nil)[:keyPrefixLen+keyFolderLen]), nil)
    ...
                if cont := fn(gf); !cont {
```
⬇
```go
func (s *FileSet) WithNeed(device protocol.DeviceID, fn Iterator) {
    ...
    s.db.withNeed([]byte(s.folder), device[:], false, nativeFileIterator(fn))
```
⬇
```go
folderFiles.WithNeed(protocol.LocalDeviceID, func(intf db.FileIntf) bool {
    ...
    if !osutil.IsDir(f.dir, filepath.Dir(file.Name)) {
        ...
        return true
    }
    ...
    if !handleItem(file) {
```
⬇
```go
handleItem := func(fi protocol.FileInfo) bool {
    ...
    case fi.IsSymlink():
        ...
        f.handleSymlink(fi)
```
⬇
```go
func (f *rwFolder) handleSymlink(file protocol.FileInfo) {
    ....
    if _, err = f.mtimeFS.Lstat(realName); err == nil {
        ...
	err = osutil.InWritableDir(os.Remove, realName)
        ...
    }
    ...
    createLink := func(path string) error {
        return symlinks.Create(path, file.SymlinkTarget, tt)
    }

    if err = osutil.InWritableDir(createLink, realName); err == nil {
````

AudriusButkevicius:
> I've tested it and was able to manipulate files outside of the folder.

Was this done on master with yesterday's commit?

> I've described one possibility but I may have missed something.

Yes the reordering would work, yet the iterator from the database always returns entries sorted lexically, so unless I am missing something I don't see how an attacker could force a different order coming from the db.

> How does it prevent the check if !osutil.IsDir(f.dir, filepath.Dir(file.Name)) { from being circumvented by ordering the children before the parents in the database on Windows?

I'd say because the graph performs ordering/sorting of it's own, irrespective of what order the remote device sends updates, yet given you are referring to windows specifically, I think I am misunderstanding something. Are you talking about someone sending a path with `\` in it from non-windows, to windows machine and then abusing that?

Unrud:
> Was this done on master with yesterday's commit I've linked to?

5070d52f2f4763d363508d10c8f5f6eaf6e20646

> yet the iterator from the database always returns entries sorted lexically

Windows doesn't care for upper and lower case letters in paths. That can be used to sort children before parents like ``foo`` and ``FOO/bar`` because ``F < f``. I've not tested that, it's possible that I missed something.

AudriusButkevicius:
Thats a good point.

calmh:
> If the type of a Symlink on Windows is directory, IsDir() returns true. You have to check if the Symlink bit is set.

Fixed.

> Right, but you can create symlinks that are named like temp files.

Fixed. Also, the same check is performed before opening a file in the read-from-temporary path.

> That can be used to sort children before parents like foo and FOO/bar because F < f. 

Not fixed yet.

I am, again, starting to reconsider the value of supporting symlinks at all vs the pain they cause us...

AudriusButkevicius:
That would be sad, given case insensitive indexes were on the list in the future anyway.

calmh:
OK, how about this. While processing the Need list, we divide things into files (those get queued) and "everything else". The everything else pile gets sorted by number of paths components and then handled in that order. Once complete, files are handled.

AudriusButkevicius:
Isn't that what we already do?

AudriusButkevicius:
Ah the order is not by path component...

AudriusButkevicius:
I guess this might be problematic as the need list might be huge

calmh:
Yes, but there's not clear way around it to me. We already queue the files, this just adds the same for non-files.

One other way that doesn't work is to do this already in the database key format, that is instead or storing names as `foo/bar/baz` and sorting based on that we could have used `<2>foo/bar/baz` with an initial int16 or so for the number of path components. But that kills prefix scans and gets generally annoying.

AudriusButkevicius:
Or we could use the spill sorter thing to create an intermediate index sorted by path component, yet I'd argue that it's perhaps better to make indexes non-casesensitive indexes as it squashes a few other bugs along the way.

Unrud:
> I am, again, starting to reconsider the value of supporting symlinks at all vs the pain they cause us...

Just make a file system abstraction that is protected against symlinks (as I suggested previously).
All the security relevant code is one place and later you can even add platform specific code to protect against local symlink races.

> non-casesensitive indexes as it squashes a few other bugs

Make the file system abstraction case-sensitive on Windows.

> That would be sad, given case insensitive indexes were on the list in the future anyway.

No, don't submit to Windows.

calmh:
Never start a sentence with "Just".

calmh:
@st-review merge it for now. we will do more work in this area.

st-review:
:ok_hand: Merged as 3582783972d37543d36b15c400bb89f3fd4f9bbb. Thanks, @calmh!

PR #3814, opened 2016-12-17
Add several new linters

calmh:
And fix their complaints, including one genuine but fairly harmless bug. Best merged as is, not via st-review, but I want the checks run on the build servers.


calmh:
Couple of more things discovered by newer versions of the linters on the builder, that I'll fix.

PR #3818, opened 2016-12-18
lib/connections: Fix port fixup in Go 1.8 (fixes #3817)

### LINKED ISSUE #3817: "missing port in address" error since v0.14.15
### Version Information
Syncthing Version: v0.14.15
OS Version: Ubuntu 16.04
Browser Version: (if applicable, for GUI issues)

Hello,
since update to v0.14.15 I get this error when trying to connect to other device:
2016/12/18 11:15:36.004319 service.go:391: DEBUG: dial failed TF2HZQ3-... tcp4://192.168.71.1 dial tcp4: address 192.168.71.1: missing port in address

The address in the configuration is actually missing the port number but as far as I know it should default to port 22000 in that case and it worked fine until last update.
It is not a big problem for me. I added the port and it works, but I have a lot of devices connecting to my server whose users are not able to edit the configuration. It would mean for to remotely connect to every device and fix the configuration.
I don't know if it is a bug or you changed it intentionally.
Please let me know.
Thanks in advance.

calmh:
The test for the error string is fragile, and the error string changed in Go 1.8 so the relevant part is no longer a prefix. This covers it with a test though, so it should be fine in the future as well.


AudriusButkevicius:
@st-review merge

st-review:

st-review:
:ok_hand: Merged as 9c67bd2550125427444d7320b77991b6e3543378. Thanks, @calmh!

wweich:
Is this part of the current release?

https://forum.syncthing.net/t/upgrade-v0-14-16-breaks-connection-between-linux-installations/8853

calmh:
No, unfortunately

PR #3825, opened 2016-12-19
lib/model: Consistently show folder description in startup messages

calmh:
We missed a couple of places to show the folder description (id + label) instead of just the ID.

Since we anyway need the folderConfig for this I'm skipping the copying of all it's attributes that rwfolder did and just keeping the original around instead.

### DIFF
--- a/lib/model/rofolder.go
+++ b/lib/model/rofolder.go
@@ -20,6 +20,7 @@ func init() {
 
 type sendOnlyFolder struct {
 	folder
+	cfg config.FolderConfiguration
 }
 
 func newSendOnlyFolder(model *Model, cfg config.FolderConfiguration, _ versioner.Versioner, _ *fs.MtimeFS) service {
@@ -30,6 +31,7 @@ func newSendOnlyFolder(model *Model, cfg config.FolderConfiguration, _ versioner
 			stop:         make(chan struct{}),
 			model:        model,
 		},
+		cfg: cfg,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


lib/model/rofolder.go
```diff
@@ -20,6 +20,7 @@ func init() {
 
 type sendOnlyFolder struct {
 	folder
+	cfg config.FolderConfiguration
 }
```
calmh: Yeah, that'd be cleaner. And fix the tests. I'll get to it.

lib/model/rofolder.go
```diff
@@ -20,7 +20,7 @@ func init() {
 
 type sendOnlyFolder struct {
 	folder
-	cfg config.FolderConfiguration
+	config.FolderConfiguration
 }
```
AudriusButkevicius: There was a joke about how Latvian president asked Lithuanian president to lend them tanks when Russian invaded Latvia.

The Lithuanian president asked whether the Latvian president would like one of them, or all of them... while Lithuania only had two...

TL;DR:
Given all of our two folders are embedding `folder`, I suggest we embed config in `folder` instead.

calmh: I have a funny joke about where exactly you can embed your config...

:trollface: anyway, that folder thing is a bit of a weird appending as it has almost no functionality. Should sort that out same day.

AudriusButkevicius:
Fine.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 5c067661f4894fc386ed03f4e05ccd60370cad2a. Thanks, @calmh!

PR #3831, opened 2016-12-20
gui: Re-word introducer text (fixes #1819)

### LINKED ISSUE #1819: UI text explaining "Introducer" setting on devices is misleading
Currently, the text reads:

> Any devices configured on an introducer device will be added to this device as well.

But that's not actually what happens.  If you have:
- Three devices A, B, and C
- A knows about B and C
- B and C only know about A (not each other)
- C has A configured as a introducer

Then C will only become aware of B if all three devices A, B, and C have _the same shared folder in common between them_.

The UI text appears to indicate that C would become aware of B regardless of any folders configured on any of the devices.  The text should be updated to make the actual behavior more clear.

ProactiveServices:
Now explicitly mentions that devices will only be added for mutually-shared folders.

### DIFF
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -40,7 +40,7 @@
           <label>
             <input type="checkbox" ng-model="currentDevice.introducer"> <span translate>Introducer</span>
           </label>
-          <p translate class="help-block">Any devices configured on an introducer device will be added to this device as well.</p>
+          <p translate class="help-block">Add devices from the introducer to our device list, for mutually shared folders.</p>
         </div>
       </div>
       <div class="row">


calmh:
I'm going to bikeshed this as it will cause work for ~34 other people as well (the translators) once merged.

> Add devices from the introducer to our device list, for mutually shared folders.

The term "device list" is unique to this sentence AFAIK. How about

"Previously unknown devices sharing our folders on the introducer will be added here as well."

ProactiveServices:
Sounds great. Squeezing a concept like this into one sentence isn't easy!

calmh:
Yeah it's a sentence i hardly understand myself when reading it so honestly I'm not sure it's an enormous improvement. I'll console myself with the thought that there will be 34 translations of it that will make either more or less sense to the poor readers on the other side...

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 79eac61b09ea360c31634ff26903d854e63417a9. Thanks, @ProactiveServices!

AudriusButkevicius:
I merged based on the fact that this is better than what we have now.

PR #3840, opened 2016-12-24
lib/model: Allow updates to deleted files (fixes #3839)

### LINKED ISSUE #3839: Resync of offline machine stuck in "Out of Sync" state after dirs+files were added and then removed elsewhere
## Current Behavior

## Expected Behavior

### Version Information

Syncthing Version: v0.14.16
OS Version: Arch Linux (x64)
Browser Version: N/A

AudriusButkevicius:
To be honest, I still think this somewhat contains holes.
This whole "resurrecting dir" stuff is now impossible.

### DIFF
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -382,7 +382,7 @@ func (f *sendReceiveFolder) pullerIteration(ignores *ignore.Matcher) int {
 		if err := fileValid(intf); err != nil {
 			// The file isn't valid so we can't process it. Pretend that we
 			// tried and set the error for the file.
-			f.newError(intf.FileName(), err)
+			f.newError(intf.FileName(), "need validation", err)
 			changed++
 			return true
 		}
@@ -431,7 +431,7 @@ func (f *sendReceiveFolder) pullerIteration(ignores *ignore.Matcher) int {
 		// Verify that the thing we are handling lives inside a directory,
 		// and not a symlink or empty space.
 		if !osutil.IsDir(f.dir, filepath.Dir(fi.Name)) {
-			f.newError(fi.Name, errNotDir)
+			f.newError(fi.Name, "preprocess", errNotDir)
 			continue
 		}
 
@@ -520,7 +520,7 @@ nextFile:
 		// Verify that the thing we are handling lives inside a directory,
 		// and not a symlink or empty space.
 		if !osutil.IsDir(f.dir, filepath.Dir(fi.Name)) {
-			f.newError(fi.Name, errNotDir)
+			f.newError(fi.Name, "process", errNotDir)
 			continue
 		}
 
@@ -608,7 +608,7 @@ func (f *sendReceiveFolder) handleDir(file protocol.FileInfo) {
 
 	realName, err := rootedJoinedPath(f.dir, file.Name)
 	if err != nil {
-		f.newError(file.Name, err)
+		f.newError(file.Name, "handleDir rj", err)
 		return
 	}
 	mode := os.FileMode(file.Permissions & 0777)
@@ -629,8 +629,7 @@ func (f *sendReceiveFolder) handleDir(file protocol.FileInfo) {
 	case err == nil && (!info.IsDir() || info.Mode()&os.ModeSymlink != 0):
 		err = osutil.InWritableDir(os.Remove, realName)
 		if err != nil {
-			l.Infof("Puller (folder %q, dir %q): %v", f.folderID, file.Name, err)
-			f.newError(file.Name, err)
+			f.newError(file.Name, "handleDir delete old", err)
 			return
 		}
 		fallthrough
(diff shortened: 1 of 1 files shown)


calmh:
I see this but I have to think more carefully about it than I'm capable of tonight. 

https://s-media-cache-ak0.pinimg.com/564x/cd/d5/e4/cdd5e4858eac95d440512d9ea2f747a2.jpg

lib/model/rwfolder.go
```diff
@@ -429,9 +429,11 @@ func (f *sendReceiveFolder) pullerIteration(ignores *ignore.Matcher) int {
 
 	for _, fi := range processDirectly {
 		// Verify that the thing we are handling lives inside a directory,
-		// and not a symlink or empty space.
-		if !osutil.IsDir(f.dir, filepath.Dir(fi.Name)) {
-			f.newError(fi.Name, errNotDir)
+		// and not a symlink or empty space. If it's already deleted, handle
+		// the check inside deleteFile deleteDir only in the scenario where
+		// the subject we are supposed to delete actually exists.
+		if !fi.IsDeleted() && !osutil.IsDir(f.dir, filepath.Dir(fi.Name)) {
```
Unrud: This seems dangerous, because a path that traverses Symlinks might end up in ``buckets``. I think this allows an attacker to delete known files outside of the folder.

lib/model/rwfolder.go
```diff
@@ -518,9 +520,11 @@ nextFile:
 		}
 
 		// Verify that the thing we are handling lives inside a directory,
-		// and not a symlink or empty space.
-		if !osutil.IsDir(f.dir, filepath.Dir(fi.Name)) {
-			f.newError(fi.Name, errNotDir)
+		// and not a symlink or empty space. If it's already deleted, handle
+		// the check inside deleteFile deleteDir only in the scenario where
+		// the subject we are supposed to delete actually exists.
+		if !fi.IsDeleted() && !osutil.IsDir(f.dir, filepath.Dir(fi.Name)) {
```
Unrud: This change is unnecessary. ([Line 515](https://github.com/AudriusButkevicius/syncthing/blob/f3dedf6a59822dbad2d12a4c85a0a58e32ec8c7f/lib/model/rwfolder.go#L515))

lib/model/rwfolder.go
```diff
@@ -784,10 +781,19 @@ func (f *sendReceiveFolder) deleteDir(file protocol.FileInfo, matcher *ignore.Ma
...
-		f.newError(file.Name, err)
+		f.newError(file.Name, "deleteDir rj", err)
 		return
 	}
 
+	if _, serr := os.Lstat(realName); !os.IsNotExist(serr) {
+		// Verify that the thing we are handling lives inside a directory,
+		// and not a symlink or empty space.
+		if !osutil.IsDir(f.dir, filepath.Dir(file.Name)) {
+			f.newError(file.Name, "deleteDir isDir", errNotDir)
```
Unrud: The delete was successful at this point, the error is pointless.
E.g. ``foo/bar`` was renamed to ``baz/bar`` and a Symlink ``foo -> bar`` was created. The ``Lstat`` check succeeds but the real file ``foo/bar`` doesn't exist anymore.

lib/model/rwfolder.go
```diff
@@ -842,10 +847,19 @@ func (f *sendReceiveFolder) deleteFile(file protocol.FileInfo) {
...
-		f.newError(file.Name, err)
+		f.newError(file.Name, "deleteFile rj", err)
 		return
 	}
 
+	if _, serr := os.Lstat(realName); !os.IsNotExist(serr) {
+		// Verify that the thing we are handling lives inside a directory,
+		// and not a symlink or empty space.
+		if !osutil.IsDir(f.dir, filepath.Dir(file.Name)) {
+			f.newError(file.Name, "deleteFile isDir", errNotDir)
```
Unrud: Same as deleteDir

lib/model/rwfolder.go
```diff
@@ -1532,7 +1540,7 @@ func (f *sendReceiveFolder) dbUpdaterRoutine() {
 			}
 			lastFile = file
 			if err := syncFn(file); err != nil {
```
Unrud: The above changes allow ``file`` to traverse a Symlink. You have to check with ``isDir``, before calling ``syncFn``. It prevents fsyncing files outside of the folder and the fsync errors disappear.

calmh:
Instead of handling the inconsistency on the filesystem, could we fix it in the database directly?

AudriusButkevicius:
Not sure what you mean. Wasn't the point thatthe db is racey compared with the fs?

calmh:
Maybe I'm misunderstanding. But isn't the problem that we have:

- dir
  - dir/file

and we get deletes for them:

- dir (deleted)
  - dir/file (deleted)

so we remove both. We then get an update:

  - dir/file (updated)

but we can't resurrect it because the dir is gone and fails the parent check? Similar thing when dir is ignored and dir/file unignored.

My alternative idea for handling this is in `WithNeed`, where it would be able to track that it's descended into the deleted `dir` and emit a need for that when it emits the need for `dir/file`. Let me try and write that up and see if it fixes the issue.

AudriusButkevicius:
I don't think thats the problem as we'd get the dir entry too.

The issue as far as I understand it is:

A and B has dir/file
A deletes dir and dir/file
B deletes it too
B goes offline
A recreates dir and dir/file
A deletes dir and dir/file (version bump)
B comes online
B can never accept the new index entries for stuff that does not exist.

PR #3856, opened 2017-01-01
lib/protocol: Revert protobuf encoder changes in v0.14.17 (fixes #3855)

### LINKED ISSUE #3855: Connection ... closed: unmarshalling message: proto: wrong wireType = 2 for field BlockIndexes
Not seen myself, but reported:

```
INFO: Connection to ... closed: unmarshalling message: proto: wrong wireType = 2 for field BlockIndexes
```

It looks like the encoding of block indexes was changed in 0725e3af, maybe due to a newer or older version of the protobuf encoder, and this is an incompatibility.

```diff
func (m *BlockInfo) Reset()                    { *m = BlockInfo{} }
@@ -380,7 +383,7 @@ type FileDownloadProgressUpdate struct {
        UpdateType   FileDownloadProgressUpdateType `protobuf:"varint,1,opt,name=update_type,json=updateType,proto3,enum=protocol.FileDownloadProgressUpdateType" json:"update_type,omitempty"`
        Name         string                         `protobuf:"bytes,2,opt,name=name,proto3" json:"name,omitempty"`
        Version      Vector                         `protobuf:"bytes,3,opt,name=version" json:"version"`
-       BlockIndexes []int32                        `protobuf:"varint,4,rep,name=block_indexes,json=blockIndexes" json:"block_indexes,omitempty"`
+       BlockIndexes []int32                        `protobuf:"varint,4,rep,packed,name=block_indexes,json=blockIndexes" json:"block_indexes,omitempty"`
 }
```

(note addition of `packed`, which should anyway be the default in proto3)

```diff
@@ -1140,11 +1163,22 @@ func (m *FileDownloadProgressUpdate) MarshalTo(data []byte) (int, error) {
        }
        i += n3
        if len(m.BlockIndexes) > 0 {
-               for _, num := range m.BlockIndexes {
-                       data[i] = 0x20
-                       i++
-                       i = encodeVarintBep(data, i, uint64(num))
+               data5 := make([]byte, len(m.BlockIndexes)*10)
+               var j4 int
+               for _, num1 := range m.BlockIndexes {
+                       num := uint64(num1)
+                       for num >= 1<<7 {
+                               data5[j4] = uint8(uint64(num)&0x7f | 0x80)
+                               num >>= 7
+                               j4++
+                       }
+                       data5[j4] = uint8(num)
+                       j4++
                }
+               data[i] = 0x22
+               i++
+               i = encodeVarintBep(data, i, uint64(j4))
+               i += copy(data[i:], data5[:j4])
        }
        return i, nil
 }
```

(actual marshalling code changed, field type changing from `0` (varint) to `2` (Length-delimited	string, bytes, embedded messages, packed repeated fields))

calmh:
The protobuf encoder now produces packed arrays for things like []int32, which is actually correct according to the proto3 spec. However Syncthing v0.14.16 and earlier doesn't support this. This reverts the encoding change, but keeps the updated decoder so that we are both more compatible with other proto3 implementations and can move to the updated encoder in the future.

I added a test that it can decode both v0.14.16 and v0.14.17 variants.

Also removes the .pb.go line from .gitattributes as we apparently need to keep an eye on diffs to these files. For a while we need to be careful with regenerating the files as that would change our encoder again.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,4 +6,3 @@ vendor/**   -text=auto
 
 # Diffs on these files are meaningless
 *.svg           -diff
-*.pb.go         -diff
--- a/lib/protocol/bep.pb.go
+++ b/lib/protocol/bep.pb.go
@@ -383,7 +383,7 @@ type FileDownloadProgressUpdate struct {
 	UpdateType   FileDownloadProgressUpdateType `protobuf:"varint,1,opt,name=update_type,json=updateType,proto3,enum=protocol.FileDownloadProgressUpdateType" json:"update_type,omitempty"`
 	Name         string                         `protobuf:"bytes,2,opt,name=name,proto3" json:"name,omitempty"`
 	Version      Vector                         `protobuf:"bytes,3,opt,name=version" json:"version"`
-	BlockIndexes []int32                        `protobuf:"varint,4,rep,packed,name=block_indexes,json=blockIndexes" json:"block_indexes,omitempty"`
+	BlockIndexes []int32                        `protobuf:"varint,4,rep,name=block_indexes,json=blockIndexes" json:"block_indexes,omitempty"`
 }
 
 func (m *FileDownloadProgressUpdate) Reset()                    { *m = FileDownloadProgressUpdate{} }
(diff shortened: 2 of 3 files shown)


calmh:
Going to verify this manually as well with .16, .17 and the new build.

calmh:
Yep, confirmed.

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 6809d38cdeef7c0e8f6eb2c880fbd5cf7f4e8559. Thanks, @calmh!

PR #3862, opened 2017-01-02
lib/connections: Allow on the fly changes to rate limits (fixes #3846)

### LINKED ISSUE #3846: Changing BW limits should not involve a restart
Currently setting or unsetting bandwidth limits is a little cumbersome (Actions > Settings > set incoming/outgoing limits, and you have to remember what they were set to before - plus it involves a restart). It would be nice if bandwidth limits could be toggled on/off (to a preset level) from the Actions menu, and if Syncthing didn't have to be restarted.

(I'd previously asked for this on the [SyncTrayzor repo](https://github.com/canton7/SyncTrayzor/issues/328), but it's complex to implement at that level. I thought it might be easier to add as functionality in core Syncthing.)

calmh:
Also replaces github.com/juju/ratelimit with golang.org/x/time/rate as the latter supports changing the rate on the fly.

The `LimitBandwidthInLAN` option already takes effect on the fly, but only for connections created after that point as we make the decision on whether it's a LAN connection or not at connection time. I think this is fine.

### Testing

Tested manually by changing the rate to and fro.


lib/connections/service.go
```diff
@@ -282,16 +271,11 @@ next:
...
-			wr = NewWriteLimiter(c, s.writeRateLimit)
-		}
-
 		rd := io.Reader(c)
-		if limit && s.readRateLimit != nil {
-			rd = NewReadLimiter(c, s.readRateLimit)
+		if s.shouldLimit(c.RemoteAddr()) {
+			wr = s.limiter.newWriteLimiter(c)
+			rd = s.limiter.newReadLimiter(c)
 		}
```
AudriusButkevicius: I say we should wrap all connections and then limit them based on LimitBwInLan which would allow live toggling of that without reconnects?

calmh: That would require doing the is-in-LAN lookup on every read/write though

AudriusButkevicius: The classification could be done on connect, the application of classification could be done on each read write. As in, we decide if its lan or not at the point we wrap, and limit based on whats set inthe config and if the conn is classified as lan

calmh: Right, yes, can do.

lib/connections/service.go
```diff
@@ -268,15 +267,8 @@ next:
...
-		// limit the connection, then we wrap it in a limiter.
-
-		wr := io.Writer(c)
-		rd := io.Reader(c)
-		if s.shouldLimit(c.RemoteAddr()) {
-			wr = s.limiter.newWriteLimiter(c)
-			rd = s.limiter.newReadLimiter(c)
-		}
+		wr := s.limiter.newWriteLimiter(c, s.isLAN(c.RemoteAddr()))
+		rd := s.limiter.newReadLimiter(c, s.isLAN(c.RemoteAddr()))
```
AudriusButkevicius: I'd store the return value in a bool, just in case that function does something more expensive in thefuture

AudriusButkevicius:
@st-review merge

st-review:

st-review:
:ok_hand: Merged as ec62888539e2e77cf3dc6f64f0c18b45eab40b04. Thanks, @calmh!

PR #3864, opened 2017-01-02
vendor: Update github.com/gogo/protobuf

calmh:
Also tweaks the proto definitions:

 - `[packed=false]` on the block_indexes field to retain compat with v0.14.16 and earlier. This didn't work earlier today but was fixed just now.

 - Uses the vendored protobuf package in include paths.

And, "build.go setup" will install the vendored protoc-gen-gogofast. This should ensure that a proto rebuild isn't so dependent on whatever version of the compiler and package the developer has installed...

The protobuf package needs to be added/updated with `gvt fetch -a` to retain all the non-Go files it contains. Or it needs to be done manually or whatever, but regular gvt fetch will clean out the proto definitions that the proto files include.


calmh:
Oh. They broke the unmarshaller. Huh. Guess we'll be back on this one then.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 987718baf89f78718b44f345429c9412920645d8. Thanks, @calmh!

PR #3870, opened 2017-01-04
Add a systemd script for Syncthing relay

kadogo:
### Purpose

I create a systemd script for syncthing relay for our private relay so maybe it can be usefull to add an example in the repository.

### DIFF
--- a/etc/linux-systemd/system/syncthing-relay.service
+++ b/etc/linux-systemd/system/syncthing-relay.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Syncthing - Relay
+After=network.target
+ 
+[Service]
+User=relaysrv
+ExecStart=/usr/local/bin/relaysrv -keys /usr/local/etc/relaysrv/ -pools ""
+Restart=on-failure
+SuccessExitStatus=3 4
+RestartForceExitStatus=3 4
+ 
+[Install]
+WantedBy=multi-user.target


etc/linux-systemd/system/syncthing-relay.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Syncthing - Relay
+After=network.target
+ 
+[Service]
+User=relaysrv
+ExecStart=/usr/local/bin/relaysrv -keys /usr/local/etc/relaysrv/ -pools ""
+Restart=on-failure
+SuccessExitStatus=3 4
+RestartForceExitStatus=3 4
```
AudriusButkevicius: It does not use exit codes in any meaningful way

etc/linux-systemd/system/syncthing-relay.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Syncthing - Relay
+After=network.target
+ 
+[Service]
+User=relaysrv
+ExecStart=/usr/local/bin/relaysrv -keys /usr/local/etc/relaysrv/ -pools ""
```
AudriusButkevicius: Pools should be left as is for the public consumption I suspect.

etc/linux-systemd/system/syncthing-relay.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Syncthing - Relay
+After=network.target
+ 
+[Service]
+User=relaysrv
```
AudriusButkevicius: What if this user does not exist?

kadogo: It's the user when we follow the documentation (https://docs.syncthing.net/users/strelaysrv.html#setting-up)
(I use /usr/local/etc in place of /etc because it's not install with a package, but we can replace it with /etc/ if you prefer).

etc/linux-systemd/system/syncthing-relay.service
kadogo: So we need to remove the 3 last lines ?
I don't know systemd well I only adapt the script for syncthing.

PR #3872, opened 2017-01-04
lib/model: Change weak hash algo, add heuristic

AudriusButkevicius:


lib/model/rwfolder.go
```diff
@@ -1216,7 +1220,7 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
 		f.model.fmut.RUnlock()
 
 		var weakHashFinder *weakhash.Finder
-		if !f.DisableWeakHash {
+		if (len(state.file.Blocks)-state.have)*100/len(state.file.Blocks) >= f.WeakHashThresholdPct {
```
calmh: this expression should definitely be a variable of some sort with a name

calmh:
```
[RJBTF] 21:27:44 INFO: Single thread hash performance is 345 MB/s using minio/sha256-simd (345 MB/s using crypto/sha256).
...
[RJBTF] 21:27:58 VERBOSE: Scanning folder "default", 9% done (218.0 MiB/s)
```

still quite a lot of hashing performance we are losing here.

lib/model/rwfolder.go
calmh: Also, potential div by zero?

AudriusButkevicius:
I guess we could go for xxhash, which would potentially raise that a bit, but then the rolling would be incredibly expensive

lib/model/rwfolder.go
```diff
@@ -1216,7 +1220,8 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
 		f.model.fmut.RUnlock()
 
 		var weakHashFinder *weakhash.Finder
-		if !f.DisableWeakHash {
+		blocksPercentChanged := (len(state.file.Blocks) - state.have) * 100 / len(state.file.Blocks)
```
calmh: what about `len(state.file.Blockes) == 0`

AudriusButkevicius: fixed

lib/model/rwfolder.go
```diff
@@ -1216,7 +1220,12 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
 		f.model.fmut.RUnlock()
 
 		var weakHashFinder *weakhash.Finder
-		if !f.DisableWeakHash {
+		blocksPercentChanged := 0
+		if len(state.file.Blocks) != 0 {
```
calmh:
```go
if tot := len(state.file.Blocks); tot > 0 {
    blocksPercentChanged = (tot - state.have) * 100 / tot
}
```

🚲 🏡

calmh:
This is vast improvement on both side, it's good enough for now.

AudriusButkevicius:
🚂 🚂

calmh:
@st-review merge it yo

lib/model, lib/weakhash: Hash using adler32, add heuristic in puller

Adler32 is much faster, and the heuristic avoid the obvious cases where it will not help.

st-review:

calmh:
obtypo in the commit message, which will look like you did it 💯 😑

st-review:
:ok_hand: Merged as 29d010ec0e7430089ed417581c31b2b503e57dd9. Thanks, @AudriusButkevicius!

Issue #3857, opened 2017-01-01
lib/symlinks: Problem with symlinks if syncthing for windows is running


n86io:
Here the problem.

![syncthing](https://cloud.githubusercontent.com/assets/8514256/21582523/6a29c276-d05c-11e6-9ae6-b94726153196.png)

In fact, no single change are happened, but the global changes show it and synced every displayed "change".

This happened only, if syncthing for windows is running. If i shutdown the windows-syncthing and use only unix-based (Ubuntu 16.04 and macOS Sierra) syncthing versions, this error not happened.

It only happened with folder-symlinks, not with file-symlinks.

### Version Information

Syncthing Version: v0.14.18 (I have noticed that already in v0.14.17)
OS Version: Windows 10 (Pro and Home)

calmh:
.17 or .18 shouldn't matter here. This looks like a real problem to me.

AudriusButkevicius:
I suspect one of the nodes participating is pre-symlink field, and uses blocks instead, hence permanently bounces as one stores it the field, other stores in the block.

n86io:
The version information is in my opinion clear enough, to know how version i use and it's clear to me, that an upgrade can possible clear some problems, if new updates available.

I had the problem with .17 and have it now also with .18

With .14 i have also delete this symlinks and created new.

AudriusButkevicius:
I am not talking about a single version, I am talking about all devices involved. The fact that changes come from unknown device imply an old version.

n86io:
No, all devices use .18. And if i make a change on a windows device, in global changes it shows, that it come from the windows device, but not for the symlink problem.

calmh:
I'm fairly sure this is a genuine bug so we can stop doubting @n86io :) There is code to prevent sending new format symlink entries to old versions, even.

AudriusButkevicius:
Yet I am still suspicious about the unset ModifiedBy in the screenshots.

n86io:
Sorry @AudriusButkevicius, but i have no reason, to mislead you or other developers and i have no time for such silly things. ;-)

nrm21:
I'm having a hard time duplicating this:

Client 1:   Windows 10  14393.576
Client 2:   Windows 10  14393.576

When I make either dir or regular symlink on either client I get no sync of any file and no mention of it in the log.  ST just ignores them, which I understand to be the correct behavior.

calmh:
[...]

PR #3875, opened 2017-01-05
lib/scanner: Never, ever descend into symlinks (ref #3857)

calmh:
On Windows we would descend into SYMLINKD type links when we scanned them successfully, as we would return nil from the walk function and the filepath.Walk iterator apparently thought it OK to descend into the symlinked directory.

With this change we always return filepath.SkipDir no matter what.

Tested on Windows 10 as admin, does what it should.


st-review:
:ok_hand: Merged as 926c88cfc4c9c34b9e31fe13121f82b3ca58104f. Thanks, @calmh!

AudriusButkevicius:
This got broken n
In the scanner refactor.

PR #3883, opened 2017-01-07
lib/osutil: IsDir returns true if path does not exist (fixes #3839)

### LINKED ISSUE #3839: Resync of offline machine stuck in "Out of Sync" state after dirs+files were added and then removed elsewhere

AudriusButkevicius:
The intent of the function is to verify that we are not traversing symlinks.
If the path along any component does not exist, we are not traversing anything, and it should be fine to let the updates through (as they are most likely only metadata updates)

/cc @Unrud


calmh:
Note that this is a bugfix for the known cases when we are synchronizing a file in a directory that does not exist:

 - because the directory was deleted on one device, but the file updated on another
 - because the directory is ignored but the file inside it unignored

calmh:
Should rename it if we do this check as the name is otherwise misleading.

AudriusButkevicius:
Yeah I had like 3 minutes before I had to leave the house. Probably should be TraversesSymlink

lib/osutil/isdir.go
```diff
@@ -14,13 +14,14 @@ import (
...
-// similar). Base and name must both be clean and name must be relative to
+// similar), but returns true if the path is inaccessible,
+// Base and name must both be clean and name must be relative to
 // base.
 func IsDir(base, name string) bool {
 	path := base
 	info, err := Lstat(path)
 	if err != nil {
-		return false
+		return os.IsNotExist(err)
```
calmh: I don't think we should change this return. If the base directory doesn't exist we're in deep water, probably "folder path missing" etc.

lib/osutil/isdir.go
```diff
@@ -14,13 +14,14 @@ import (
 
 // IsDir returns true if base and every path component of name up to and
 // including filepath.Join(base, name) is a directory (and not a symlink or
-// similar). Base and name must both be clean and name must be relative to
+// similar), but returns true if the path is inaccessible,
```
calmh: This is broken. :) Maybe just `... is a directory or non-existent (and not a symlink`?

calmh:
@st-review lgtm

st-review:

AudriusButkevicius:
Merge this or the other implementation, whatever you prefer.

calmh:
@st-review merge it!

st-review:
:ok_hand: Merged as 1a1e35d99872b86ec29d43aef61f9eac39de6b1a. Thanks, @AudriusButkevicius!

PR #3885, opened 2017-01-07
lib/osutil: Replace IsDir with TraversesSymlink (fixes #3839)

### LINKED ISSUE #3839: Resync of offline machine stuck in "Out of Sync" state after dirs+files were added and then removed elsewhere

AudriusButkevicius:
This is an alternative, which just checks for symlink traversal, which was our main security issue

/cc @Unrud


AudriusButkevicius:
On android this might cause a bit of stupidness, as EvalSymlinks is not using our magical lstat, but it's probably fine.

AudriusButkevicius:
I guess we could fix Lstat with:

https://github.com/bouk/monkey

calmh:
I'm skeptical as EvalSymlinks does a lot of work that we don't need, much more than our simple lstat loop. It's also checking for a negative rather then ensuring a positive (i.e., a black list rather than a white list). I think your other approach to verify that everything along the way is a directory (or potentially nothing at all) is better.

AudriusButkevicius:
But this better represents what exactly we are trying to catch, and I wouldn't say its doing much more work.

calmh:
Fair enough, it's not exactly as bad as I was afraid of;

```
BenchmarkIsDir-8              	  200000	     11022 ns/op	    2096 B/op	      21 allocs/op
BenchmarkTraversesSymlink-8   	  100000	     15993 ns/op	    2784 B/op	      35 allocs/op
```

AudriusButkevicius:
So this renames stuff, yet keeps the functionality as discussed.

AudriusButkevicius:
Anyways, both PRs now fix the issue. Tested manually with the repro provided. Merge whichever one you want.

lib/osutil/traversessymlink_test.go
```diff
@@ -35,40 +35,42 @@ func TestIsDir(t *testing.T) {
...
 		{"a/x", false},
 		{"a/b/x", false},
 		{"a/x/c", false},
 		// Symlink or behind symlink
-		{"a/l", false},
-		{"a/l/c", false},
+		{"a/l", true},
+		{"a/l/c", true},
+		// Non-existing behind a symlink, but does not fail as does not exist.
+		{"a/l/x", false},
```
calmh: But it should fail, or we will attempt `mkdir("a/l/x")` which we should not.

calmh:
Right, you said as much in your comment above, sorry.

Issue #2299, opened 2015-09-21
Auto-Accept shared folders from trusted devices


Getron:
This is a proposal of auto-accepting shared folders from a trusted devices.

For example, if i want to add more and more shares on my servers, i must on every server to accept the shares, every share, on every device.

If the share come from the trusted (added) device, maybe to have an option on device, to check a box, if the device is trusted (or own), and to give an standard folder location on the server or machine, where the auto-accepted shares would go (maybe, if the auto-accepted share is on, then, on this standard folder, the syncthing maybe create an subfolder with the name of shared folder, to have it sorted, if there are many shared folders), and would be accepted automaticaly.

Like this, user must not manualy accept every share, on every sync machines (devices), separatelly. Just add the shares on local machine, or local server, and have it done on all other machines (devices, or servers).

Devs, what you think? Is this an good idea?

marrie:
I agree. I've been having a  similar discussion on the forums but about a queue to except devices if you don't have the web UI up, maybe to auto except devices as well. A lot of people think the idea sucks, but I personally think both your idea and mine would make for a better work flow. Of corse these could be toggles for the user who does not want the option.

AudriusButkevicius:
The question for the original post, would the location where folders get created would be per device or global? Also, what happens if a folder under the same name already exists? Do we share the existing one?

AudriusButkevicius:
As for devices, I don't think it will happen as adding a device does not share any folders hence adds no value. And adding new device and sharing folders automatically is just plain damn dangerous.

Getron:
I think, just auto-add the share from "Trusted" devices.

Maybe, to add an option on the device adding, a check aquare, to mark the device as own, or trusted. Not the devices itself to be added. Just shares from a trusted devices. This is safe.

Example:
I have 2 devices: 1 PC and one Server

I share a folder from my PC with server. On the server side, i should just once add the PC, and on adding (or later), on the device options, i should able to mark the PC device as a trusted one.

All future shared folders from my PC should automaticaly added on my server.

The location where the folder where created (in my example, server), i should maybe choose on the main options of syncthing. Or maybe in the "Sync" folder itself to add a folder "Auto-added"(this is a standard one), but, it should be great, if the user can choose one.

After this, on this main folder, if, for example, my PC syncthing name is "Main-PC", on the server side, the syncthing should create a folder with the name of devices, on the main folder.

It should look like this:

[...]

PR #3892, opened 2017-01-10
Simple solution for issue #2299 (Auto-accepted shared folders from trusted devices)

cherylquirion:
### Purpose
This provides a simple solution to add a detected device to syncthing automatically and enables sharing of existing folders (with correct folder ID). This solution does not create new folders. It will only add devices to folders that already exist given their folder ID.

I have made a flag available in the optionsconfiguration to be able to configure syncthing to accept devices and folders automatically. By default, it is disabled.

### Use case 
We have a large Raspberry Pi (mobile) installation. We use a "sync server" (a Raspberry Pi cluster) in the garages where the vehicles enter for service (mostly daily but some only come in once a week). The 'sync server' needs to be able to add mobile devices on-the-fly when they are detected in the local network. Due to the sheer volume of devices we need to be able to do this without having to know/approve the device ID in advance.

Btsync does this well as it synchronises only based on the FileID. In this case, if the devices have the same FileID (which we control via Puppet) then the files can be safely synchronised. We already have over hundreds of devices in the field and will roll out hundreds more over the coming months. Manually adding device IDs and folder IDs is untenable. We want to be able to install a new device in the vehicle and have it start synching automatically once it discovers the 'sync server' in the network.


calmh:
Hi! This sounds like a really neat use case, happy to hear about it. One thing that strikes me (regardless of the merit of the pull request, which I haven't yet looked at) is that I wonder what the scalability constraints here are - your Syncthing instance will add potentially thousands of devices, all of which it will need to keep track of completion status for and attempt to connect to every second. That is, this moves into the "public sync" area that we've so far avoided, beyond the pure "auto add devices" feature.

It also seems like a potentially dangerous option - a user might think "hey, that sounds like it'll save me a bunch of work", and now any user can connect to them, join their folders, and access their data? Perhaps there is something more to this as the title mentions "trusted devices". I'll glance at the code now! :)

lib/model/model.go
```diff
@@ -685,8 +685,21 @@ func (m *Model) NeedFolderFiles(folder string, page, perpage int) ([]db.FileInfo
 func (m *Model) Index(deviceID protocol.DeviceID, folder string, fs []protocol.FileInfo) {
 	l.Debugf("IDX(in): %s %q: %d files", deviceID, folder, len(fs))
 
-	if !m.folderSharedWith(folder, deviceID) {
-		l.Debugf("Unexpected folder ID %q sent from device %q; ensure that the folder exists and that this device is selected under \"Share With\" in the folder configuration.", folder, deviceID)
+	autoAcceptDevicesFolders := m.cfg.Options().AutoAcceptDevicesFolders
+
+	if autoAcceptDevicesFolders == true {
```
calmh: I'm not really sure why this needs to happen in `Index` and `IndexUpdate` in addition to in `ClusterConfig`? We can be assured that the first message on the wire will always be a ClusterConfig so it should be enough to fix up any folder memberships there I think?

lib/model/model.go
```diff
@@ -770,12 +798,21 @@ func (m *Model) folderSharedWithLocked(folder string, deviceID protocol.DeviceID
 	return false
 }
 
+func (m *Model) ThrowUnexpectedFolderError (folder protocol.Folder, deviceID protocol.DeviceID) {
```
calmh: Nitpick, but we don't really talk about throwing in Go and this could be an unexported method, so `unexpectedFolderError` would be OK. If it needs to exist, given that I think we should only need to do this processing in one place.

AudriusButkevicius:
To be very blunt and honest, I think you are using the wrong tool for the job.
Syncthing is not great for things that are considered to be public, atleast not in the current format it's in.

calmh:
Mmm. This dodges the whole "default path" crap of adding unknown folders, and that's good. But allowing any device at all to connect and join folders makes the folder ID a weak shared secret, and that undermines my intention of how the security should work.

But there may be ways to solve this use case all the same. Assuming that the number of remote devices is too large to manage manually, yet small enough that Syncthing will actually scale to solve this problem.

For example, perhaps the roaming devices could all have device IDs based on certificates that are signed by the central Syncthing instance. It would then allow any device to join, but with the caveat that they need to present a certificate that is signed by itself. That may complicate your provisioning process of those devices, and it's certainly more code, but it would be something that wouldn't be that "off switch" for security.

cherylquirion:
Hi guys, happy to chat about the use case in more detail. I am not a Go programmer so I am not accustomed to the style (sorry!). My strengths are C, Java, Python. However, I tried my hand at Go to nut out the concept quickly.

I agree with the security concerns. The folder ID is a weak secret but I guess something we are willing to wear in the short term for the convenience of not having to add device manually. Our network traffic is restricted via a VPN. We aim to have around 1000 devices operational by the end of this year. So there is no way that manually adding these devices will work. 

I am about to move this into a scaled test with many Raspberry Pi devices next week. From there, we will do a limited production test running it side-by-side with btsync. I will report back the results.

calmh:
Yeah don't worry about the style things - we'll point them all out, but it's just superficial. The main thing is the fundamental changes to how it might get used by people. From what you say there a few things that might change the situation:

- Your vehicles are behind VPN. We could add a restriction that we allow unknown devices from a known IP range only, 172.16.0.0/20 for example, and that would increase security significantly and still handle your use case I think?

- You're using Raspberries. That's not really been a great thing to do with Syncthing in my experience, at least when it comes to scaling up the amount of data. I suspect that scaling up the amount of disconnected devices won't matter nearly as much, but it's definitely a very low powered device and I'd be wary of using it as a central point.

I get having Raspberries in vehicles and that's probably super fine. But something more powerful, even the smallest, cheapest, least power hungry "real" CPU would be a huge boost.

From a configuration standpoint I guess you could set the reconnection interval to 86400 or something on the central server(s) so they almost never spend any time trying to connect to the vehicles and just accept incoming connections.

cherylquirion:
Hi... I have responded to your questions in line. Let me know if you need more info. As I said we plan to test extensively in our staging environment soon (probably early next week). 

> "Your vehicles are behind VPN. We could add a restriction that we allow unknown devices from a known IP range only, 172.16.0.0/20 for example, and that would increase security significantly and still handle your use case I think?" 

- I think that this would be perfect for our use case. We use a VPN for all comms. We would need to make that configurable obviously.

> You're using Raspberries. That's not really been a great thing to do with Syncthing in my experience, at least when it comes to scaling up the amount of data. I suspect that scaling up the amount of disconnected devices won't matter nearly as much, but it's definitely a very low powered device and I'd be wary of using it as a central point......

- We have multiple buses working out of a large garage. We have our "sync-server" (which is a cluster of 10 RPis) to handle synchronisation of the local network when the buses are in for fueling/cleaning. We would never be synching hundreds or thousands of buses at a time to one server. We have a sync-server in 1 garage handling 200 buses. We will deploy another sync-server cluster to another garage that will handle about the same number. We will also install sync-servers to remote garages where buses are serviced around the country. HOWEVER there are only ever ~20-25 buses maximum being serviced at a time because that is all there is physically space for.

Our topology is as follows:

- Organisation A:    buses  <----> sync-server (garage) <------> master sync server (amazon) 

- Organisation B:    buses  <----> sync-server (garage) <------> master sync server (amazon) 

Each organisation will have  one or more sync-server cluster (usually in their major garage) to service their buses. Typical bus fleets are ~200-300 buses. Each sync-server syncs content to amazon (our content master) via VPN. The buses only ever talk to a known sync-server via VPN.

> From a configuration standpoint I guess you could set the reconnection interval to 86400 or something on the central server(s) so they almost never spend any time trying to connect to the vehicles and just accept incoming connections.

- For the sync-server I think that makes sense if I understand your question correctly. However, when the buses enter the garage we need to work fast. There is a short window of opportunity where the vehicles are being fueled/serviced (a couple of hours). So we need a fairly frequent (but not too frequent) polling interval - I would start testing with every 15 minutes. This is basically the only time that the bus may be in internet range. They work out of range most of the time.

lib/model/model.go
cherylquirion: My misunderstanding. I will pull it all back to only ClusterConfig.

lib/model/model.go
cherylquirion: Not nitpicky at all. I am learning Go obviously! :)

calmh:
I think the cleanest way to do this would be to hook into the event listener. On the central device, you'll get notifications of rejected connections, which you can then react to by adding the device to the config. Next time it reconnects (in 60 seconds by default) it'll be accepted. I'm happy to discuss ways to get auto accepting connections to work directly in syncthing, it's something other people have asked for as well, but disabling security completely is probably not the right way.

PR #3901, opened 2017-01-13
lib/ignore: Centralize handling of temporary filenames (fixes #3899)

### LINKED ISSUE #3899: Move TempNamer to ignores too
This is similar to #3779:
While fixing a race in the FS notification PR (fswatcher package), I stumpled over `TempNamer`, which is needed for its `IsTemporary` function. It is currently used in model and and scanner packages. It is passed as part of a struct from the model to the scanner package and for the case it is not passed, an interface and replacement type had to be implemented in the scanner package. The same would have to be done for the fswatcher package. However the whole thing is completely static: It is defined at init and then never changed.
Therefore I propose to move it to the ignores package, such that it is easily accessible without passing it around (ignores is imported in all places already).

What do you think about this - do I get the green light to do this?

imsodin:
For purpose see #3899, short recap:
Creating temporary names and testing whether a name is temporary is a "static" test, meaning it is set at `init()` and never changed again. Passing it around from the models module is overly complicated and pointless. So this PR centralizes these functions in the ignore package for easy access for all libraries.

I also refactored the function `shouldIgnore` in models.go: The part that checks whether the filename is either internal, temporary or ignored is handled by `ShouldIgnore` in the ignores package. This might seem unnecessary, as it is currently only used in models.go, but it contextually belongs there and such a function will be needed in fswatcher as well. Shouldn't hurt, right? 

The model tests (in specific rwfolder_test.go) use static test data with temporary windows prefixes. These were set in `init()` of the test itself, which worked as the prefix is an internal variable of the same package. Now it is an internal variable of the ignore package. I resolved this by adding an exported function to the ignore package, that manually sets the prefix to the windows type. I considered this less bad than exporting the prefix itself.
Is there any better way to both protect this variable in production and allow to set it for tests?

### DIFF
--- a/lib/ignore/ignore.go
+++ b/lib/ignore/ignore.go
@@ -249,6 +249,22 @@ func (m *Matcher) clean(d time.Duration) {
 	}
 }
 
+// ShouldIgnore returns true when a file is temporary, internal or ignored
+func (m *Matcher) ShouldIgnore(filename string) bool {
+	switch {
+	case IsTemporary(filename):
+		return true
+
+	case IsInternal(filename):
+		return true
+
+	case m.Match(filename).IsIgnored():
+		return true
+	}
+
+	return false
+}
+
 func hashPatterns(patterns []Pattern) string {
 	h := md5.New()
 	for _, pat := range patterns {
@@ -416,21 +432,5 @@ func IsInternal(file string) bool {
 			return true
 		}
 	}
-	return false
-}
-
-// ShouldIgnore returns true when a file is temporary, internal or ignored
-func ShouldIgnore(filename string, matcher *Matcher) bool {
-	switch {
-	case IsTemporary(filename):
-		return true
-
-	case IsInternal(filename):
-		return true
-
-	case matcher.Match(filename).IsIgnored():
-		return true
-	}
-
 	return false
 }
--- a/lib/ignore/ignore.go
+++ b/lib/ignore/ignore.go
@@ -418,3 +418,19 @@ func IsInternal(file string) bool {
 	}
 	return false
 }
+
+// ShouldIgnore returns true when a file is temporary, internal or ignored
+func ShouldIgnore(filename string, matcher *Matcher) bool {
+	switch {
+	case IsTemporary(filename):
+		return true
+
+	case IsInternal(filename):
+		return true
+
+	case matcher.Match(filename).IsIgnored():
+		return true
+	}
+
+	return false
+}
--- a/lib/ignore/tempname.go
+++ b/lib/ignore/tempname.go
@@ -4,7 +4,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at http://mozilla.org/MPL/2.0/.
 
-package model
+package ignore
 
 import (
 	"crypto/md5"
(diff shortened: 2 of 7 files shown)


lib/ignore/tempname.go
```diff
@@ -33,33 +28,39 @@ const maxFilenameLength = 160 - len(unixTempPrefix) - len(".tmp")
...
 	}
-	tname := fmt.Sprintf("%s%s.tmp", t.prefix, tbase)
+	tname := fmt.Sprintf("%s%s.tmp", tempPrefix, tbase)
 	return filepath.Join(tdir, tname)
 }
+
+// SetWindowsPrefix is entirely reserved for testing, it should never be called
+// in production code.
+// It is needed to allow static test files independent from underlying os.
+func SetWindowsPrefix() {
```
calmh: Looks good to me except for this method which is ugly and clobbers any other usage of temp files after usage, in tests by other packages for example. Better alternatives imho are exporting the variables themselves, or a more general setprefixes method that can be called before and after tests to restor.

AudriusButkevicius:
God damn github with it's new  "oh I really want to comment" button that I keep forgetting to press.

lib/ignore/ignore.go
```diff
@@ -418,3 +418,19 @@ func IsInternal(file string) bool {
 	}
 	return false
 }
+
+// ShouldIgnore returns true when a file is temporary, internal or ignored
+func ShouldIgnore(filename string, matcher *Matcher) bool {
```
imsodin: done

imsodin:
@AudriusButkevicius I was already wondering about your strangely long silence xD

Regarding the testing uglyness: I can easily go ahead with calmh's proposition, but I agree with AudriusButkevicius: Ideally there should not be any exported function just for testing. Unfortunately I don't know the reflect hack or the monkey patching library, and google wasn't educating. Any hints?

AudriusButkevicius:
See if you can get it to work:

https://github.com/bouk/monkey

AudriusButkevicius:
Or actuall, as @calmh suggests, just expose the variable.

imsodin:
Monkey (in its current form) would be pointless, as it can neither replace internal stuff nor variables, only exposed functions.
So I will expose the variable.

Regarding @calmh's comment
> Looks good to me except for this method which is ugly and clobbers any other usage of temp files after usage, in tests by other packages for example. 
Are imported packages only loaded (i.e. `init()` called) once or whenever they are explicitly imported?
In case of the former, resetting to the original prefix should not be necessary, right? If it is, is there any way to run a function that resets the prefix to the original value on exiting tests (sort of like the opposite of `init()`)?

AudriusButkevicius:
They are called once per binary invocation, which happens once for all tests.

imsodin:
I rubberducked it and found TestMain (and didn't read Audrius' comment early enough), thus did it through that. Is it ok?
I also exported the const vars `WindowsTempPrefix` and `UnixTempPrefix` for ease of use in testing and consistency.

calmh:
@st-review lgtm

st-review:

AudriusButkevicius:
@st-review lgtm

st-review:
:ok_hand: Merged as dbb3a34887b40b730c045700c1705859187f9016. Thanks, @imsodin!

Issue #3902, opened 2017-01-13
pathIsSubFolder


bugith:
When pathIsSubFolder is true, the warning message should also display otherfolderlabel if exists, in addition of currently only folderID (otherfolder). Something like this was already be done recently for when we an receive invitation to a new shared folder, although this was not related to pathIsSubFolder.
Thanks

Sorry, still not a coder ;-)

bugith:
Thanks so much guys, you're beep-beep :-) Trying to do it myself, I'd still be wandering in which file this code is.
Great team

PR #3904, opened 2017-01-14
gui: Warn if a new folder is a parent of an existing one (fixes #3197)

### LINKED ISSUE #3197: gui: Should warn when adding a folder that is a parent of an existing folder
Hi,

I work with a team of animators, and our Syncthing share has a huge, nested share structure that currently has about 80GB in it, and it's growing a little bit every day. 
I don't know how difficult this would be to make happen (I imagine very), but there have been a couple times in working with Syncthing with a team that I have had to deal with infinite directory loops. Oftentimes they are a result from erring on some Syncthing share algebra, but the fact that it's happened twice in the past couple months could be cause for concern. In those instances, a popup warning that lets us know something along the lines of "Continuing with what you are doing could result in an infinite directory loop!" would have been really nice to keep us from making a that terrible mistake.

If you can't pull the directory structure at share creation, then maybe while it's starting its initial sync, run some pattern recognition that kills the sync process for the affected shares? This might help a lot since not all of my people are very technically inclined. Just a thought. Thanks!

Syncthing Version: v0.13.0
OS Version: Windows 7,10

kamadak:
This patch add a check if a new folder is a parent of an existing one.
The might be a bit verbose because the warning is displayed as soon as a user types a "/" to input a full path.


AudriusButkevicius:
So this looks good to me, but there is another ticket for actually showing the other path and I wonder if I could free-load that on you given you are poking in that area? 

If you don't have time that I am fine to merge this.

kamadak:
Yes, I will take care of that while I'm here.

kamadak:
The patch for #3902 introduces two more messages, resulting in four slightly different ones...  I'd like to reduce them for less translation/maintenance work, but have no good idea for now.

AudriusButkevicius:
@st-review merge

gui: Improve warnings when creating folder in a subdirectory (fixes #3197, fixes #3902)

st-review:
:ok_hand: Merged as 929a4d0c0cced14a2770e7b526cdeb34b8522547. Thanks, @kamadak!

PR #3915, opened 2017-01-18
cmd/syncthing: Fail if -logfile and -no-restart passed together (fixes #3912)

### LINKED ISSUE #3912: The -logfile flag doesn't work together with -no-restart (STNORESTART=1)
### Version Information

Syncthing Version: v0.14.19
OS Version: OSX

Running with the following command: `/usr/local/opt/syncthing/bin/syncthing -no-browser -no-restart -logflags=7 -verbose -logfile=/var/log/syncthing.log`

No log file is created, no matter what settings I use for `logfile` or `logflags`. Running the command from a shell shows the expected logging to stdout with the VERBOSE logging (when specified). It just never writes to the logfile. I have run this on multiple machines running El Capitan and Sierra with same results.

canton7:
### Purpose

Fail if -no-restart and -logflags passed together, since the result is unintuitive.

### Testing

Ran it, saw that it failed.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -289,6 +289,12 @@ func main() {
 		os.Setenv("STGUIAPIKEY", options.guiAPIKey)
 	}
 
+	// Check for options which are not compatible with each other
+	// Have to check logfile before it's set to the default below - we only want to complain if they set -logfile explicitly, not if it's set to its default location
+	if options.noRestart && options.logFile != "" {
+		l.Fatalln("-logfile may not be used with -no-restart or STNORESTART")
+	}
+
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
I recall there being some back about file being - for stdout -- for stderr. Maybe thats for the audit files, but should be checked by someone who is not on the phone.

calmh:
Yes, that's for event audit only (and should work regardless of the other options).

calmh:
@st-review merge it!

st-review:
:ok_hand: Merged as 4c8aa14e07f2b18b340a906326500711fdc7f208. Thanks, @canton7!

PR #3918, opened 2017-01-18
cmd/syncthing: Report real hashing performance, including weakhash

calmh:
Instead of

    [I6KAH] 19:05:56 INFO: Single thread hash performance is 359 MB/s using minio/sha256-simd (354 MB/s using crypto/sha256).

it now says

    [I6KAH] 19:06:16 INFO: Single thread SHA256 performance is 359 MB/s using minio/sha256-simd (354 MB/s using crypto/sha256).
    [I6KAH] 19:06:17 INFO: Actual hashing performance is 299.01 MB/s

which is more informative. This is also the number it reports in usage reporting.


scienmind:
What is the fundamental difference between the two kinds of measurements?
(seems like `sha256.SelectAlgo()` and `cpuBench()` called sequentially)

calmh:
One is SHA256 only, the other is the full hash process which now is SHA256 and adler32 in sequence.

AudriusButkevicius:

AudriusButkevicius:
I suspect it's broken due to the newly added boolean to blocks indicating if we want weak hash or not.

calmh:
fixd

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 6d4fa27ea709e4c15b89353bd2fa50483ffac5e2. Thanks, @calmh!

PR #3922, opened 2017-01-23
lib/connections: Fix rate limiting on arm64 (fixes #3921)

### LINKED ISSUE #3921: Rate limiting is broken on ARM64
Apparently the default rate limit of zero (no limit) breaks on arm64. The cause chain seems to be:

- [we create a zero rate limiter](https://github.com/syncthing/syncthing/blob/master/lib/connections/limiter.go#L68) (that `<` should be a `<=`, thanks to @AudriusButkevicius for noticing)
- [the zero rate is documented to drop all events](https://godoc.org/golang.org/x/time/rate#Limit)
- there is a bug (???) in the rate limiter package that makes this not actually happen, instead a zero rate limiter allows all events
- except, apparently on arm64 (???)

Last point to be confirmed, but seems to be the case...

AudriusButkevicius:
Blah

### DIFF
--- a/lib/connections/limiter.go
+++ b/lib/connections/limiter.go
@@ -65,7 +65,7 @@ func (lim *limiter) CommitConfiguration(from, to config.Configuration) bool {
 		lim.read.SetLimit(1024 * rate.Limit(to.Options.MaxRecvKbps))
 	}
 
-	if to.Options.MaxSendKbps < 0 {
+	if to.Options.MaxSendKbps <= 0 {
 		lim.write.SetLimit(rate.Inf)
 	} else {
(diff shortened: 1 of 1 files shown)


calmh:
@st-review merge it

lib/connections: Fix rate limiting on arm64 (fixes #3921)

Skip-check: metalint

st-review:

calmh:
(Updated stuff on the build servers to apparently metalint got smarter or more stupid, will fix...)

st-review:
:ok_hand: Merged as 2c4b92d410d33f848e46d2b7373f7aa1222ea1db. Thanks, @AudriusButkevicius!

PR #3925, opened 2017-01-24
lib/weakhash: Limit number of hits for any given weakhash

calmh:
### Purpose

Don't run out of memory when we find the desired block everywhere, as in a file of all zeroes when looking for the zero block.

### Testing

None, just eyeballing it.

### DIFF
--- a/lib/weakhash/weakhash.go
+++ b/lib/weakhash/weakhash.go
@@ -16,6 +16,9 @@ import (
 
 const (
 	Size = 4
+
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as ee36e2d46d3392a6a40891cce59c53556a6f16b3. Thanks, @calmh!

PR #3927, opened 2017-01-24
cmd/syncthing: Fix rescan delay in /rest/db/scan

imsodin:
### Purpose

Currently scanning is delayed even when the scan fails but never when all folders are scanned (i.e. no folder keyword given). This changes the behaviour such that the next scan is only delayed when the scan is successful, but regardless whether a specific folder or all folders were scanned.

### Testing

Nothing new.

### Documentation

This brings the actual behaviour closer to the docs. Related: syncthing/docs#283

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -1131,14 +1131,8 @@ func (s *apiService) postDBScan(w http.ResponseWriter, r *http.Request) {
 	qs := r.URL.Query()
 	folder := qs.Get("folder")
 	if folder != "" {
-		nextStr := qs.Get("next")
-		next, err := strconv.Atoi(nextStr)
-		if err == nil {
-			s.model.DelayScan(folder, time.Duration(next)*time.Second)
-		}
-
 		subs := qs["sub"]
-		err = s.model.ScanFolderSubdirs(folder, subs)
+		err := s.model.ScanFolderSubdirs(folder, subs)
 		if err != nil {
 			http.Error(w, err.Error(), 500)
 			return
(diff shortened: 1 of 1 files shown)


cmd/syncthing/gui.go
```diff
@@ -1151,6 +1145,11 @@ func (s *apiService) postDBScan(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
+	nextStr := qs.Get("next")
+	next, err := strconv.Atoi(nextStr)
+	if err == nil {
+		s.model.DelayScan(folder, time.Duration(next)*time.Second)
```
calmh: I don't think this part makes sense when `folder == ""`?

imsodin: Right. Would it be better to add a function DelayScanAll or handle an empty folder string in the existing DelayScan function?

calmh: I don't think it makes sense to delay scan of all folders, really.

imsodin: Ok. And only delaying if the scan succeeded is a good idea or not either?

calmh: That part does make sense, I think. Probably.

AudriusButkevicius:
@st-review lgtm

st-review:

calmh:
One question though, what does this fix? Is there a situation where a delay is asked for and the scan fails and the delay happening causes a problem?

imsodin:
@calmh I should have mentioned the reason for my PR: I was just looking at the code for the docs update and it struck me as odd that syncing is delayed even though the scan failed (and the API returns an error).

calmh:

calmh:
@st-review merge it please

cmd/syncthing: Only delay next scan (via REST) if the scan succeeds

st-review:

st-review:
:ok_hand: Merged as 237893ead3f0ac923145e6a55b6f41b828808db1. Thanks, @imsodin!

Issue #899, opened 2014-10-23
ETA and bytes left for synchronization


uok:
Since current speed and bytes left are already here it would be great to use them to calculate an ETA for download/upload of each folder and also show the time left for the sync status of each device. After successful transfers this line is hidden.

Another helpful information would be to see the total remaining size (sum of all remaining transfers) of all connected folders for each device.

label: [enhancement]

matsp:
From time to time I miss this feature, too. I want to add that it would be a nice feature to see the actual file upload to each device that is connected to my own client.

marco-m:
I would like to mention that there is a concept called "Exponentially Weighted Moving Average", that looks like exactly fit to the use case that at the beginning you don't really know "how much you need to fetch until later". 

There is this Go implementation https://github.com/VividCortex/ewma that is used by, for example, by https://github.com/vbauerster/mpb and it works well.

ebanDev:
Any news about this issue ?

AudriusButkevicius:
No

KraXen72:
what is preventing this to be merged?

tomasz1986:
> what is preventing this to be merged?

This is an issue. There is no pull request that would fix it yet.

KraXen72:
i see, and https://github.com/syncthing/syncthing/pull/5361 was frozen due to age.

PR #3928, opened 2017-01-24
gui: Remaining sync bytes in folder header (fixes #3908)

### LINKED ISSUE #3908: "Syncing (X MB to go)" instead of "Syncing (99%)"
I have a number of folders with lots of data (pictures, music, documents), that I make frequent small changes to. The status line in the UI "Syncing (99%)" is then almost useless, since the 1% that is unsynced can represent anything up to 700MB. It would be much more useful if this could be replaced with something like "Syncing (50 MiB to go)".

benshep:
The progress indicator in the folder header in the GUI now shows the
remaining bytes to sync as well as the percentage.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -285,7 +285,7 @@
                   <span ng-switch-when="idle"><span class="hidden-xs" translate>Up to Date</span><span class="visible-xs">&#9724;</span></span>
                   <span ng-switch-when="syncing">
                     <span class="hidden-xs" translate>Syncing</span>
-                    ({{syncPercentage(folder.id)}}%)
+                    ({{syncPercentage(folder.id)}}%, {{syncRemaining(folder.id)}})
                   </span>
                   <span ng-switch-when="outofsync"><span class="hidden-xs" translate>Out of Sync</span><span class="visible-xs">&#9724;</span></span>
                 </div>
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -285,7 +285,7 @@
                   <span ng-switch-when="idle"><span class="hidden-xs" translate>Up to Date</span><span class="visible-xs">&#9724;</span></span>
                   <span ng-switch-when="syncing">
                     <span class="hidden-xs" translate>Syncing</span>
-                    ({{syncPercentage(folder.id)}}%, {{syncRemaining(folder.id) | binary}}B)
+                    ({{syncPercentage(folder.id)}}%, &#9662;{{syncRemaining(folder.id) | binary}}B)
                   </span>
                   <span ng-switch-when="outofsync"><span class="hidden-xs" translate>Out of Sync</span><span class="visible-xs">&#9724;</span></span>
                 </div>
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -285,7 +285,7 @@
                   <span ng-switch-when="idle"><span class="hidden-xs" translate>Up to Date</span><span class="visible-xs">&#9724;</span></span>
                   <span ng-switch-when="syncing">
                     <span class="hidden-xs" translate>Syncing</span>
-                    ({{syncPercentage(folder.id)}}%, {{syncRemaining(folder.id) | binary}}B)
+                    <span ng-show="syncRemaining(folder.id)">({{syncPercentage(folder.id)}}%, {{syncRemaining(folder.id) | binary}}B)</span>
                   </span>
                   <span ng-switch-when="outofsync"><span class="hidden-xs" translate>Out of Sync</span><span class="visible-xs">&#9724;</span></span>
                 </div>
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -285,7 +285,7 @@
                   <span ng-switch-when="idle"><span class="hidden-xs" translate>Up to Date</span><span class="visible-xs">&#9724;</span></span>
                   <span ng-switch-when="syncing">
                     <span class="hidden-xs" translate>Syncing</span>
-                    ({{syncPercentage(folder.id)}}%, &#9662;{{syncRemaining(folder.id) | binary}}B)
+                    ({{syncPercentage(folder.id)}}%, {{syncRemaining(folder.id) | binary}}B)
                   </span>
                   <span ng-switch-when="outofsync"><span class="hidden-xs" translate>Out of Sync</span><span class="visible-xs">&#9724;</span></span>
                 </div>
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -285,7 +285,7 @@
                   <span ng-switch-when="idle"><span class="hidden-xs" translate>Up to Date</span><span class="visible-xs">&#9724;</span></span>
                   <span ng-switch-when="syncing">
                     <span class="hidden-xs" translate>Syncing</span>
-                    ({{syncPercentage(folder.id)}}%, {{syncRemaining(folder.id)}})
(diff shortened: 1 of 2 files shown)


gui/default/syncthing/core/syncthingController.js
```diff
@@ -722,6 +722,26 @@ angular.module('syncthing.core')
...
+			// KiB, MiB, etc.
+            if (typeof $scope.model[folder] === 'undefined') {
+                return 0;
+            }
+            if ($scope.model[folder].globalBytes === 0) {
+                return 0;
+            }
+
+            var bytes = $scope.model[folder].globalBytes - $scope.model[folder].inSyncBytes;
+			if (isNaN(bytes)) {
```
canton7: Here too.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -722,6 +722,26 @@ angular.module('syncthing.core')
             return Math.floor(pct);
         };
 
+        $scope.syncRemaining = function (folder) {
+			// Formats the remaining sync bytes into a string with a suffix
```
canton7: Got a bit of a tab/space mixup here, by the looks of things.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -722,6 +722,26 @@ angular.module('syncthing.core')
...
+            }
+            if ($scope.model[folder].globalBytes === 0) {
+                return 0;
+            }
+
+            var bytes = $scope.model[folder].globalBytes - $scope.model[folder].inSyncBytes;
+			if (isNaN(bytes)) {
+				return 0;
+			}
+			var k = 1024;
```
calmh: We have a directive for this, `{{ ... | binary}}` I think.

AudriusButkevicius:
A screenshot how this looks would be nice.

gui/default/index.html
```diff
@@ -285,7 +285,7 @@ <h4 class="panel-title">
                   <span ng-switch-when="idle"><span class="hidden-xs" translate>Up to Date</span><span class="visible-xs">&#9724;</span></span>
                   <span ng-switch-when="syncing">
                     <span class="hidden-xs" translate>Syncing</span>
-                    ({{syncPercentage(folder.id)}}%)
+                    ({{syncPercentage(folder.id)}}%, {{syncRemaining(folder.id)}})
```
calmh: It seems that it might be unclear what the amount means. Size? Transferred? Remaining?

benshep:
Here's a screenshot.

image

I fixed the tab thing, and also put in an extra check for zero.

gui/default/index.html
benshep: I was trying to make it concise! Could add "to go" I suppose, but I think it should be pretty clear when you watch it for a few seconds and it goes down.

calmh:
I think my comment about the directive got lost in the indentation change. There is no need to do formatting, we have a `binary` directive. Check the other places that display byte counts.

AudriusButkevicius:
I am still of an opinion that we're duplicating stuff, and I agree that it's not obvious what that number actually means, but I am not against it given you've produced a PR.

uok:
Looks like a good improvement, but I'm not sure about screen space on smaller screens. Also note that some translations are a lot longer than "syncing". I was thinking about two lines, also adding ETA (#899) to the mix:
e.g. first line "Syncing (55%)", second line "40 MB, 1h 3min"

calmh:
There is literally no way to guess ETA from the outside. We have no idea at what rate their sync progresses. Also, they may have 250 GiB out of sync, and we'll show "250 GiB" with this change, but that might be a change to the permissions bits of a 250 GiB file which takes literally zero time and nothing to transfer or copy. This is one of the things I mentioned in the original issue - this number doesn't actually mean what most people would think it means.

imsodin:
The same argument applies to the sync percentage, making it useless in such a scenario, right?
The reason why I like displaying the absolute number of bytes more than percentage is mentioned in the original issue: No matter how big the folder, you will get some progress report. With the relative percentage it just duplicates the "syncing" (=99%) - "idle" (=100%) indication.

fti7:
+1 i think it could be an improvement in terms of showing progress to the User. You might want to use an icon to indicate what the Number does.
Of course it should be self-explanatory if the refresh interval is high enough/the decrease is fluid/snappy

ETA / Transferate would also be nice... but of course we need to gather this Data somehow first.
Also multiline might be an good Option. 

singleline

Or Multiline:
multiline

multiline2

benshep:
@uok The longest I can find is Turkish! But the GUI seems to resize very gracefully on narrower displays. The box first expands vertically to accommodate the text:

image

Then the "Syncing" text is hidden altogether, so it's less of an issue:

image

The only GUI issue I can see is the "Syncing" text overlaps with the computer name in the right-hand box:

image

But that's a totally separate issue.

benshep:
@calmh Yes, I get that showing ETA is hard. That should be a separate issue if people want it. I see your point that 250GiB might not mean 250GiB to transfer. I have some multi-GiB virtual machine files, and they usually sync very fast due to the hashing thing (thanks for that, BTW!). But in my "normal" use-case (which I admit might not be everyone's), 250GiB usually does mean 250GiB to transfer, and so it's usually meaningful. In any case, it's a conservative estimate, right?

benshep:
@fti7 Nice mockups, but that text is a bit small. The idea of this is really an at-a-glance indication of how much is left to transfer.

benshep:
@fti7 Here's one with a little icon.

image

uok:
Nice mockups, I really like the multi line with ETA ver much! 👍 
For me "ETA = (remaining size per folder / current global speed)" would be enough although I know that Syncthing re-uses blocks, etc. it would just a rought estimate. But if the user is on a mobile connection it would give you something to work with.

If we just keep % and MB at the moment, I think the little icon is not needed, maybe use `|` or `-` as a separator.

calmh:
> current global speed

There is no such thing that we're currently aware of.

Ferroin:
I'd argue that the icon shown could be misunderstood to indicate that the given number is the download rate, not the amount remaining.

I also kind of agree with @uo, regarding the ETA, but some slightly more complicated math would give a better estimate,  Something along the lines of:
(total remaining data for this folder / (current download speed / % of total state across all folders that need to be synced that this folder requires))
should give a slightly more accurate representation.  FWIW that's essentially the math used by most web browsers and download managers, they just don't need the division to calculate the stream bandwidth for each file because each file is it's own stream.

benshep:
OK, removed the icon.

calmh:
Ok, lets shelf the discussions on a theoretical ETA and transfer rate that is not currently in evidence and focus on the PR at hand...

@benshep I'm mostly fine with this. I'd like to preemptively suggest only showing it when the number is >0 though. There are known cases where it will turn negative, which already affects the percentage weirdly. We don't need to add insult to injury by also claiming that there are -542 KiB left to sync.

benshep:
Slightly out of scope of this issue, but now the percentage and bytes are hidden when there's nothing useful to show (100%, 0 B).

image

calmh:
All good

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as e03be9158b03b3ad3c880ace569d64b633d687d9. Thanks, @benshep!

calmh:
(For the record, in much of the discussion above I was, erroneously, thinking about showing this for remote devices, not local folders. My misunderstanding. We can of course calculate a sync rate and an ETA locally. Carry on.)

PR #3929, opened 2017-01-24
cmd/syncthing: Send Ping events to the disk events API

canton7:
The Ping event is important, as it means that requests complete within
a sensible time. The disk events API didn't have the Ping event, so
if there were no disk events, the request would keep taking forever.
Unless, of course, there's a reverse proxy which times the request out
after a suitably large interval (or something else aborts it), in which
case Syncthing isn't very happy.

### Testing

Ran Syncthing, saw that the request to the disk events API completed with a ping event after a while.


calmh:
@st-review merge it

st-review:
:ok_hand: Merged as ddca8d91fa137ea39e40274167398a68eb4c3bc4. Thanks, @canton7!

PR #3939, opened 2017-01-26
cmd/syncthing, gui, lib/config, lib/upgrade: Add option to upgrade to pre-releases

calmh:
### Purpose

This replaces the old behavior where "beta versions" (those with a dash somewhere in the version number) would upgrade to pre-releases and no-one else would. Instead there is an explicit checkbox for it, defaulting to off. No importance is given to what kind of release is running *now* - if you're running a pre-release and the checkbox is off, you'll get the next real *release* and then stay on that track.

screen shot 2017-01-26 at 11 38 10

### Testing

Unit testing covers the actual upgrading behavior. I manually tested the checkbox works as intended.

### Further work

A few more things need to happen, in separate PRs:

- When upgrading to this version, we should display a notice saying "hey, there are now pre-releases!" with a short explanation of pros and cons. Otherwise no-one will ever end up running pre-releases.

- When installing from scratch from a pre-release, it should default to upgrading to other pre-releases.

- When installing from scratch from a pre-release, it should turn on usage reporting by default and also notify the user that this happened and why it's awesome.


AudriusButkevicius:
@st-review merge

st-review:

calmh:
Updated UI. I combined the choices to a select, with the proposed effect on the usage reporting. The help link will go to the docs page describing the different releases and their tradeoffs. That page should have a short note about auto upgrades as well given the help link placement.

No upgrades:

screen shot 2017-01-27 at 12 05 08

Stable releases, can enable usage reporting:

screen shot 2017-01-27 at 12 04 48

Candidate releases, always with usage reporting:

screen shot 2017-01-27 at 12 04 59

Of course the user may be on a stable release and enabling candidate release upgrades then enables usage reporting as well. I think it's worth it to show that up front and not only when the upgrade happens.

AudriusButkevicius:
@st-review merge

st-review:

imsodin:
Maybe still include the preview link when usage reporting is forced on for candidate releases?

calmh:
@imsodin Yes.

screen shot 2017-01-27 at 12 21 19

@AudriusButkevicius One does not simply merge this pull request... Apparently. Not sure what the builders are doing....

st-review:
@AudriusButkevicius: Patiently waited 30m0s for the build status to turn green, but enough is enough.

calmh:
@st-review IT IS GREEN

st-review:
I'm sorry, @calmh. I'm afraid I don't know what you mean. I know how to `don't`, `lgtm`, `merge`, `prevent`, `squash`, `stop`.

calmh:
@calmh slaps st-review around a bit with a large trout.

AudriusButkevicius:
@st-review merge it

st-review:
:ok_hand: Merged as 35e87e23fda4b696f6f5eed82483e0e0ea4bebfe. Thanks, @calmh!

uok:
Good improvement, hope it helps to expand users for testing.
Sorry for the late reply, I was thinking there is no real translation for "candidate releases" - my suggestion for the wording would be 
- "Disabled"
- "Stable releases only" 
- "All releases" or "stable and test releases"

imsodin:
Sorry for my ignorance (which might be due to too limited/wrong language knowledge), but what is the problem with "candidate releases"? For the languages I know the translations are a bit weird as I am used to just using the technical terms in English even when not talking English, but they are possible: "Vorläufige /Kandidats-Version" and "version/sortie candidate" (both are likely not ideal, more a prove of concept).

uok:
My idea was in the direction of an average user who is not a programmer but might be interested to help test new versions. I just wanted to prevent workarounds like "Release-Kandidat-Version" and "Golden-Master-Kandidat-Version" 😄

calmh:
@uok I sort of agree, first wrote it as "Stable and candidate releases" which sort is equivalent to "All releases". But the other effects make less sense that way - "Usage reporting is always enabled for candidate releases" when I selected "All releases"....

PR #3941, opened 2017-01-27
lib/events: Add timeout to events API, remove Ping (fixes #3933)

### LINKED ISSUE #3933: Empty entries in global log
There are [empty entries in the global log](http://i.imgur.com/eNcWlB7.png) appearing every minute, so I suspect #3929 is to blame. The ping events have to be filtered out somehow.

canton7:
Previously, the Ping event was used to ensure that something appered on the events API at least every minute, so that requests to the events API would take at most a minute. However this is less than ideal in some cases. If there are very few events, then ping events can dominate, meaning it's hard to find other events. Filtering out ping events from the disk events API also makes consumers more complex.

Instead, allow a timeout to be specified when requesting events. If no events are found within this timeout, then an empty array is returned. The timeout defaults to 60 seconds (for backwards compatibility), and any invalid value (<0, non-numeric) is also treated as 60 seconds. A timeout of zero is allowed, in which case the events API because request-response instead of long-polling, returning immediately if no events are available.

In order to facilitate this, a new structure in lib/sync, TimeoutCond, was created. This behaves similarly to Cond, except that it allows the user to specify a timeout. One timeout will span multiple calls to .Wait().

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -233,8 +233,8 @@ func (s *apiService) Serve() {
 	getRestMux.HandleFunc("/rest/db/need", s.getDBNeed)                          // folder [perpage] [page]
 	getRestMux.HandleFunc("/rest/db/status", s.getDBStatus)                      // folder
 	getRestMux.HandleFunc("/rest/db/browse", s.getDBBrowse)                      // folder [prefix] [dirsonly] [levels]
-	getRestMux.HandleFunc("/rest/events", s.getIndexEvents)                      // since [limit]
-	getRestMux.HandleFunc("/rest/events/disk", s.getDiskEvents)                  // since [limit]
+	getRestMux.HandleFunc("/rest/events", s.getIndexEvents)                      // since [limit] [timeout]
+	getRestMux.HandleFunc("/rest/events/disk", s.getDiskEvents)                  // since [limit] [timeout]
 	getRestMux.HandleFunc("/rest/stats/device", s.getDeviceStats)                // -
 	getRestMux.HandleFunc("/rest/stats/folder", s.getFolderStats)                // -
 	getRestMux.HandleFunc("/rest/svc/deviceid", s.getDeviceID)                   // id
@@ -1019,8 +1019,14 @@ func (s *apiService) getEvents(w http.ResponseWriter, r *http.Request, eventSub 
 	qs := r.URL.Query()
 	sinceStr := qs.Get("since")
 	limitStr := qs.Get("limit")
+	timeoutStr := qs.Get("timeout")
 	since, _ := strconv.Atoi(sinceStr)
 	limit, _ := strconv.Atoi(limitStr)
+
+	timeout := defaultEventTimeout
+	if timeoutSec, timeoutErr := strconv.Atoi(timeoutStr); timeoutErr == nil && timeoutSec >= 0 { // 0 is a valid timeout
+		timeout = time.Duration(timeoutSec) * time.Second
+	}
 
 	// Flush before blocking, to indicate that we've received the request and
 	// that it should not be retried. Must set Content-Type header before
@@ -1029,7 +1035,8 @@ func (s *apiService) getEvents(w http.ResponseWriter, r *http.Request, eventSub 
 	f := w.(http.Flusher)
 	f.Flush()
 
-	evs := eventSub.Since(since, nil)
+	// If there are no events available return an empty slice, as this gets serialized as `[]`
+	evs := eventSub.Since(since, []events.Event{}, timeout)
 	if 0 < limit && limit < len(evs) {
 		evs = evs[len(evs)-limit:]
 	}
--- a/cmd/syncthing/summaryservice.go
+++ b/cmd/syncthing/summaryservice.go
@@ -169,7 +169,7 @@ func (c *folderSummaryService) foldersToHandle() []string {
 	c.lastEventReqMut.Lock()
 	last := c.lastEventReq
 	c.lastEventReqMut.Unlock()
-	if time.Since(last) > pingEventInterval {
+	if time.Since(last) > defaultEventTimeout {
 		return nil
 	}
 
--- a/cmd/syncthing/verboseservice.go
+++ b/cmd/syncthing/verboseservice.go
@@ -66,7 +66,7 @@ func (s *verboseService) WaitForStart() {
 
 func (s *verboseService) formatEvent(ev events.Event) string {
 	switch ev.Type {
-	case events.Ping, events.DownloadProgress, events.LocalIndexUpdated:
+	case events.DownloadProgress, events.LocalIndexUpdated:
 		// Skip
 		return ""
 
--- a/lib/events/events.go
+++ b/lib/events/events.go
@@ -275,7 +275,7 @@ type bufferedSubscription struct {
 	next int
 	cur  int // Current SubscriptionID
 	mut  sync.Mutex
-	cond sync.TimeoutCond
+	cond *sync.TimeoutCond
 }
 
 type BufferedSubscription interface {
@@ -284,11 +284,11 @@ type BufferedSubscription interface {
 
 func NewBufferedSubscription(s *Subscription, size int) BufferedSubscription {
 	bs := &bufferedSubscription{
-		sub:  s,
-		buf:  make([]Event, size),
-		mut:  sync.NewMutex(),
-		cond: sync.NewTimeoutCond(),
+		sub: s,
+		buf: make([]Event, size),
+		mut: sync.NewMutex(),
 	}
+	bs.cond = sync.NewTimeoutCond(bs.mut)
 	go bs.pollingLoop()
 	return bs
 }
@@ -316,35 +316,21 @@ func (s *bufferedSubscription) pollingLoop() {
 }
 
 func (s *bufferedSubscription) Since(id int, into []Event, timeout time.Duration) []Event {
-
-	// BE VERY CAREFUL modifying this section - ensure that the locking still ties up.
-	// The outermost 'if' statement MUST exit with s.mut LOCKED
 	s.mut.Lock()
-	// Check once before creating the TimeoutCond (and its associated cost)
+	defer s.mut.Unlock()
+
+	// Check once first before generating the TimeoutCondWaiter
 	if id >= s.cur {
-		s.mut.Unlock()
+		waiter := s.cond.SetupWait(timeout)
+		defer waiter.Stop()
 
-		condWaiter := s.cond.SetupWait(timeout)
-		defer condWaiter.Stop()
-
-		for {
-			if eventAvailable := condWaiter.Wait(); eventAvailable {
-				s.mut.Lock()
-
-				if id < s.cur {
-					// The event we want is available. The lock IS held at this point
-					break
-				}
-
-				s.mut.Unlock()
-			} else {
-				// Timed out. The lock is NOT held at this point
+		for id >= s.cur {
+			if eventsAvailable := waiter.Wait(); !eventsAvailable {
+				// Timed out
 				return into
 			}
 		}
 	}
-
-	defer s.mut.Unlock()
 
 	for i := s.next; i < len(s.buf); i++ {
 		if s.buf[i].SubscriptionID > id {
--- a/lib/events/events.go
+++ b/lib/events/events.go
@@ -10,7 +10,6 @@ package events
 import (
 	"errors"
 	"runtime"
-	stdsync "sync"
 	"time"
 
 	"github.com/syncthing/syncthing/lib/sync"
@@ -19,8 +18,7 @@ import (
 type EventType int
 
 const (
-	Ping EventType = 1 << iota
-	Starting
+	Starting EventType = 1 << iota
 	StartupComplete
 	DeviceDiscovered
 	DeviceConnected
@@ -55,8 +53,6 @@ var runningTests = false
 
 func (t EventType) String() string {
 	switch t {
-	case Ping:
-		return "Ping"
 	case Starting:
 		return "Starting"
 	case StartupComplete:
@@ -274,16 +270,16 @@ func (s *Subscription) C() <-chan Event {
 }
 
 type bufferedSubscription struct {
-	sub  *Subscription
-	buf  []Event
-	next int
-	cur  int // Current SubscriptionID
-	mut  sync.Mutex
-	cond *stdsync.Cond
+	sub                  *Subscription
+	buf                  []Event
+	next                 int
+	cur                  int // Current SubscriptionID
+	mut                  sync.Mutex
+	newEventNotification chan struct{} // If non-nil, this is closed when a new event becomes available. It is lazily created when needed.
 }
 
 type BufferedSubscription interface {
-	Since(id int, into []Event) []Event
+	Since(id int, into []Event, timeout time.Duration) []Event
 }
 
 func NewBufferedSubscription(s *Subscription, size int) BufferedSubscription {
@@ -292,7 +288,6 @@ func NewBufferedSubscription(s *Subscription, size int) BufferedSubscription {
 		buf: make([]Event, size),
 		mut: sync.NewMutex(),
 	}
-	bs.cond = stdsync.NewCond(bs.mut)
 	go bs.pollingLoop()
 	return bs
 }
@@ -314,18 +309,21 @@ func (s *bufferedSubscription) pollingLoop() {
 		s.buf[s.next] = ev
 		s.next = (s.next + 1) % len(s.buf)
 		s.cur = ev.SubscriptionID
-		s.cond.Broadcast()
+		if s.newEventNotification != nil {
+			close(s.newEventNotification)
+			s.newEventNotification = nil
+		}
 		s.mut.Unlock()
 	}
 }
 
-func (s *bufferedSubscription) Since(id int, into []Event) []Event {
+func (s *bufferedSubscription) Since(id int, into []Event, timeout time.Duration) []Event {
+	if eventAvailable := s.waitForNewEventWithTimeout(id, timeout); !eventAvailable {
+		return into
+	}
+
 	s.mut.Lock()
 	defer s.mut.Unlock()
-
-	for id >= s.cur {
-		s.cond.Wait()
-	}
 
 	for i := s.next; i < len(s.buf); i++ {
 		if s.buf[i].SubscriptionID > id {
@@ -350,4 +348,57 @@ func Error(err error) *string {
 	}
 	str := err.Error()
 	return &str
+}
+
+// Wait for the given event to become available, or for a timeout, whichever is sooner.
+// Returns true if the event became available, or false if we timed out.
+func (s *bufferedSubscription) waitForNewEventWithTimeout(id int, timeout time.Duration) bool {
+	// Checks whether 'id' is available. If it is, returns nil. If it isn't, returns newEventNotification (constructing it if necessary)
+	checkIDAndConstructNotificationIfNecessary := func() <-chan struct{} {
+		s.mut.Lock()
+		defer s.mut.Unlock()
+
+		if id < s.cur {
+			return nil
+		}
+		if s.newEventNotification == nil {
+			s.newEventNotification = make(chan struct{})
+		}
+		return s.newEventNotification
+	}
+
+	// Early exit, in the case where events are already available - saves creating the timer
+	newEventNotification := checkIDAndConstructNotificationIfNecessary()
+	if newEventNotification == nil {
+		return true
+	}
+	if timeout <= 0 {
+		return false
+	}
+
+	// From the point on, we know we're going to have to wait.
+
+	// Don't use time.After, as that (temporarily) leaks the timer if we don't timeout, and that might
+	// make a difference when there's a large volume of events.
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+
+	for {
+		// If the timer fires, then we've definitely timed out.
+		// If the newEventNotification channel is closed, then the event we want may or may not be present.
+		// It's likely that it will be, because people don't tend to ask for events in the future,
+		// but check anyway. If it isn't, we'll need to make sure that a new newEventNotification has been
+		// created.
+
+		select {
+		case <-timer.C:
+			return false
+		case <-newEventNotification:
+		}
+
+		newEventNotification = checkIDAndConstructNotificationIfNecessary()
+		if newEventNotification == nil {
+			return true
+		}
+	}
 }
--- a/lib/events/events.go
+++ b/lib/events/events.go
@@ -270,12 +270,12 @@ func (s *Subscription) C() <-chan Event {
 }
 
 type bufferedSubscription struct {
-	sub                  *Subscription
-	buf                  []Event
-	next                 int
-	cur                  int // Current SubscriptionID
-	mut                  sync.Mutex
-	newEventNotification chan struct{} // If non-nil, this is closed when a new event becomes available. It is lazily created when needed.
+	sub  *Subscription
+	buf  []Event
+	next int
+	cur  int // Current SubscriptionID
+	mut  sync.Mutex
+	cond sync.TimeoutCond
 }
 
 type BufferedSubscription interface {
@@ -284,9 +284,10 @@ type BufferedSubscription interface {
 
 func NewBufferedSubscription(s *Subscription, size int) BufferedSubscription {
 	bs := &bufferedSubscription{
-		sub: s,
-		buf: make([]Event, size),
-		mut: sync.NewMutex(),
+		sub:  s,
+		buf:  make([]Event, size),
+		mut:  sync.NewMutex(),
+		cond: sync.NewTimeoutCond(),
 	}
 	go bs.pollingLoop()
 	return bs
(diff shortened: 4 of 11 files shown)


cmd/syncthing/summaryservice.go
```diff
@@ -169,7 +169,8 @@ func (c *folderSummaryService) foldersToHandle() []string {
 	c.lastEventReqMut.Lock()
 	last := c.lastEventReq
 	c.lastEventReqMut.Unlock()
-	if time.Since(last) > pingEventInterval {
+	// TODO: Hard-coded value of 60, used to be pingEventInterval
```
canton7: Not sure what best to do here.

canton7:
While I'm here, is it worth reducing the size of the disk events buffered subscription to 25, as we only ever fetch 25 events? Or not worth the effort?

cmd/syncthing/gui.go
```diff
@@ -1019,8 +1019,15 @@ func (s *apiService) getEvents(w http.ResponseWriter, r *http.Request, eventSub
...
 	limitStr := qs.Get("limit")
+	timeoutStr := qs.Get("timeout")
 	since, _ := strconv.Atoi(sinceStr)
 	limit, _ := strconv.Atoi(limitStr)
+	var timeoutSec int
+	if timeoutStr == "" {
+		timeoutSec = 60 // Default to 60 for backwards compatibility
+	} else {
+		timeoutSec, _ = strconv.Atoi(timeoutStr)
+	}
```
AudriusButkevicius: It might be better to always try and parse the time (even if its an empty string) and then always default to 60s in case of error.

cmd/syncthing/verboseservice.go
```diff
@@ -66,10 +66,6 @@ func (s *verboseService) WaitForStart() {
 
 func (s *verboseService) formatEvent(ev events.Event) string {
 	switch ev.Type {
-	case events.Ping, events.DownloadProgress, events.LocalIndexUpdated:
-		// Skip
-		return ""
-
 	case events.Starting:
```
AudriusButkevicius: This removes skips for a bunch of other events

cmd/syncthing/summaryservice.go
AudriusButkevicius: Maybe we should always generate the summaries?

lib/events/events.go
```diff
@@ -314,19 +309,26 @@ func (s *bufferedSubscription) pollingLoop() {
...
-func (s *bufferedSubscription) Since(id int, into []Event) []Event {
+func (s *bufferedSubscription) Since(id int, into []Event, timeout time.Duration) []Event {
+	eventAvailable := s.waitForNewEventWithTimeout(id, timeout)
+	if !eventAvailable {
+		if into == nil {
+			return make([]Event, 0)
+		}
+		return into
+	}
+
```
AudriusButkevicius: Just return into, semantics of nil are the same as of an empty slice.

lib/events/events.go
```diff
@@ -314,19 +309,26 @@ func (s *bufferedSubscription) pollingLoop() {
...
+			s.newEventNotification = nil
+		}
 		s.mut.Unlock()
 	}
 }
 
-func (s *bufferedSubscription) Since(id int, into []Event) []Event {
+func (s *bufferedSubscription) Since(id int, into []Event, timeout time.Duration) []Event {
+	eventAvailable := s.waitForNewEventWithTimeout(id, timeout)
+	if !eventAvailable {
```
AudriusButkevicius: The call could be evaluated as part of the if

lib/events/events.go
```diff
@@ -351,3 +353,47 @@ func Error(err error) *string {
...
+	cur := s.cur
+	s.mut.Unlock()
+	if id < cur {
+		return true
+	}
+
+	// From the point on, we know we're going to have to wait
+
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
```
AudriusButkevicius: Just use time.After

lib/events/events.go
```diff
@@ -351,3 +353,47 @@ func Error(err error) *string {
...
+
+	for {
+		s.mut.Lock()
+		if s.newEventNotification == nil {
+			s.newEventNotification = make(chan struct{})
+		}
+		newEventNotification := s.newEventNotification
+		s.mut.Unlock()
```
AudriusButkevicius: I think there is a race here that will break it 

Cur=5
Id=5

T1: goes past the early exit
T2: emits new event, cur=6
T1: sets up notification channel at which point its too late
T1: times out without notifying about cur=6

cmd/syncthing/gui.go
```diff
@@ -233,8 +233,8 @@ func (s *apiService) Serve() {
 	getRestMux.HandleFunc("/rest/db/need", s.getDBNeed)                          // folder [perpage] [page]
 	getRestMux.HandleFunc("/rest/db/status", s.getDBStatus)                      // folder
 	getRestMux.HandleFunc("/rest/db/browse", s.getDBBrowse)                      // folder [prefix] [dirsonly] [levels]
-	getRestMux.HandleFunc("/rest/events", s.getIndexEvents)                      // since [limit]
-	getRestMux.HandleFunc("/rest/events/disk", s.getDiskEvents)                  // since [limit]
+	getRestMux.HandleFunc("/rest/events", s.getIndexEvents)                      // since [limit] [timeout]
+	getRestMux.HandleFunc("/rest/events/disk", s.getDiskEvents)                  // since [limit] [timeout]
```
canton7: My interpretation was that the comment listed the required and [optional]  query parameters for that endpoint. Since we've added an optional timeout parameter to these endpoints, I added it to the comment. Did I misunderstand what's going on here?

cmd/syncthing/gui.go
canton7: Yeah, I can buy that

cmd/syncthing/verboseservice.go
canton7: Damnit

lib/events/events.go
canton7: It could, but since it returns a "magic" bool with no obvious meaning I wanted to give it a name, for readability. Happy to remove it if you care strongly.

lib/events/events.go
canton7: The JSON serializer doesn't think so - it serializes nil as emptystring, but an empty slice as `[]`. I wanted to return an empty JSON array in the case of no events, to be as friendly to consumers as possible. I'm happy to move this logic out into gui.go if you'd prefer.

lib/events/events.go
canton7: That means that we keep the timer running even if we discovery a new event. It's a small, temporary resource leak. Happy to change if you're OK with that.

lib/events/events.go
canton7: Yep

cmd/syncthing/gui.go
AudriusButkevicius: The problem is that my phone screen is too narrow and I didn't even see there is a comment there, so just looked like spaces to tabs change or something like that.

lib/events/events.go
AudriusButkevicius: Personally I am fine to leak resources temporarily if we can make the code prettier, but I don't have a strong preference either.

lib/events/events.go
AudriusButkevicius: I am not sure its worth it. I'd rather get this fixed on the caller side.

cmd/syncthing/summaryservice.go
calmh: Add a constant `defaultEventTimeout` which is what we use when there is none given by the client, and also what we *expect* the GUI to use in order to keep triggering summary events?

lib/events/events.go
canton7: Righto, I'll change it on the caller.

lib/events/events.go
canton7: Fine :smile:

lib/events/events.go
calmh: I think Audrius meant

```go
if eventAvailable := s.waitForNewEventWithTimeout(id, timeout); !eventAvailable {
```

which both documents the bool and limits its scope.

lib/events/events.go
calmh: No. This is correct. With lots of events happening this thing may be called hundreds of times per second, every time creating a new 60 second timer.

cmd/syncthing/gui.go
```diff
@@ -1019,8 +1019,13 @@ func (s *apiService) getEvents(w http.ResponseWriter, r *http.Request, eventSub
 	qs := r.URL.Query()
 	sinceStr := qs.Get("since")
 	limitStr := qs.Get("limit")
+	timeoutStr := qs.Get("timeout")
 	since, _ := strconv.Atoi(sinceStr)
 	limit, _ := strconv.Atoi(limitStr)
+	timeoutSec, timeoutErr := strconv.Atoi(timeoutStr)
+	if timeoutErr != nil || timeoutSec < 0 { // A timeout of 0 is valid
```
calmh: I'd simplify to just

```go
timeoutSec, _ := strconv.Atoi(timeoutStr)
if timeoutSec <= 0 {
    timeoutSec = DefaultEventTimeoutS
}
```

but I don't care strongly about it (apart from having a constant which is already mentioned).

lib/events/events.go
calmh: Yeah. The `evs := eventSub.Since(since, nil)` in gui.go can become `evs := eventSub.Since(since, []Event{})` and the intention is explicit there.

lib/events/events.go
```diff
@@ -351,3 +353,54 @@ func Error(err error) *string {
 	str := err.Error()
 	return &str
 }
+
+// Wait for the given event to become available, or for a timeout, whichever is sooner.
+// Returns true if the event became available, or false if we timed out.
+func (s *bufferedSubscription) waitForNewEventWithTimeout(id int, timeout time.Duration) bool {
+	// Checks whether 'id' is available. If it is, returns nil. If it isn't, returns newEventNotification (constructing it if necessary)
+	checkIDAndConstructNotificationIfNecessary := func() <-chan struct{} {
```
calmh: I'm not super fond of this construction but I can't exactly pinpoint the problem.

Really I think I'd want this whole thing moved out into `sync.TimeoutCond` that is like a `sync.Cond` with the behavior we need...

lib/events/events.go
canton7: Ah, gotcha

lib/events/events.go
```diff
@@ -351,3 +353,54 @@ func Error(err error) *string {
...
+
+	// From the point on, we know we're going to have to wait
+
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+
+	for {
+		// If the timer fires, then we've definitely timed out.
+		// If the newEventNotification channel is closed, then the event we want may or may not be present.
+		// It's likely that it will be, because people don't tend to ask for events in the future,
```
calmh: I wouldn't worry about it, as we already return the empty array if that happens.

lib/events/events.go
canton7: If hundreds of events are being created a second, then most of the polls should result in events already being available, in which case we don't create the timer. I'll leave as-is (and add a comment) though

cmd/syncthing/gui.go
canton7: OK, sure. I thought it might be nice to support a timeout of 0 (since I can imagine a caller just wanting to know whether any events are available - the Global Changes stuff in the GUI might want to do this, for instance).

calmh: Ah, right, in that case keep it as is!

lib/events/events.go
canton7: OK, I'll see what I can cook up.

lib/events/events.go
canton7: Worry about what, exactly? The current behaviour will not return an empty array if the Cond is signalled before the event we're after is available - it'll loop around and wait on it again.

lib/events/events.go
calmh: I could, for example imagine a fairly clean TimeoutCond that does something like (pseudo code)

```
wait(timeout):
  - create a channel
  - with locking, add the channel to a list of waiting channels
  - select on the channel and a timeout

broadcast:
  - iterate over the list of waiting channels
    - close each
```

lib/events/events.go
calmh: Erm, yes. You're right, I'm wrong. Carry on. :)

lib/events/events.go
canton7: Why do you need a list of waiting channels? Can't you share one channel between all waiters, as I've done?

calmh: Possibly yes.

lib/events/events.go
canton7: Right, added a commit which attempts to do this. It gets messy as the timer needs to live through different calls to Wait(), but it's not too bad

canton7:
I'm failing a deadlock test, so I've screwed something up. I've run out of time for now, but I'll come back to this.

lib/sync/sync.go
```diff
@@ -227,3 +227,67 @@ func goid() int {
...
+	c.mut.Lock()
+	defer c.mut.Unlock()
+
+	if c.ch != nil {
+		close(c.ch)
+		c.ch = nil
+	}
+}
+
+func (c *TimeoutCond) SetupWait(timeout time.Duration) TimeoutCondWaiter {
```
canton7: Taking ownership of another Mutex, and unlocking it at the start / locking it again at the end (similar to what Cond does) would make the use of this a lot cleaner, but I'm worried about the effects of too much magic.

lib/events/events.go
```diff
@@ -319,12 +315,19 @@ func (s *bufferedSubscription) pollingLoop() {
...
+func (s *bufferedSubscription) Since(id int, into []Event, timeout time.Duration) []Event {
 	s.mut.Lock()
 	defer s.mut.Unlock()
 
-	for id >= s.cur {
-		s.cond.Wait()
+	// Check once first before generating the TimeoutCondWaiter
+	if id >= s.cur {
+		waiter := s.cond.SetupWait(timeout)
+		for id >= s.cur {
```
imsodin: What is the reason to create `TimeoutCondWaiter` instead of setting up the timer in one call to (an adapted) `s.cond.Wait(timeout)`?

canton7: From the commit message:

> This gets a bit messy, as the timeout's Timer needs to span calls to Wait()

If we're waiting for an event in the future, we need to keep waiting until we see the event we're after. There's one timeout which spans all of these waits.

canton7:
OK, that's looking a lot smarter now, and I'm significantly more happy with it.

lib/sync/sync.go
```diff
@@ -227,3 +227,70 @@ func goid() int {
...
+// TimeoutCondWaiter is a type allowing a consumer to wait on a TimeoutCond with a timeout. Wait() may be called multiple times,
+// and will return true every time that the TimeoutCond is broadcast to. Once the configured timeout
+// expires, Wait() will return false.
+// Call Stop() to release resources once this TimeoutCondWaiter is no longer needed.
+type TimeoutCondWaiter struct {
+	c     *TimeoutCond
+	timer *time.Timer
+}
+
+func NewTimeoutCond(l sync.Locker) TimeoutCond {
```
calmh: Given that the methods are all pointer methods, I think this should return a `*TimeoutCond` as well, even though the difference in practice is negligible. It also *looks* safer, as it doesn't look like L is copied.

lib/sync/sync.go
```diff
@@ -227,3 +227,70 @@ func goid() int {
...
+func (c *TimeoutCond) Broadcast() {
+	// ch.L must be locked when calling this function
+
+	if c.ch != nil {
+		close(c.ch)
+		c.ch = nil
+	}
+}
+
+func (c *TimeoutCond) SetupWait(timeout time.Duration) TimeoutCondWaiter {
```
calmh: Same here, return a `*TimeoutCondWaiter` I think.

lib/events/events.go
canton7: Righty, done this in a follow-up commit

lib/sync/sync.go
canton7: Righto, I wasn't sure about this. My logic was that this saved a heap allocation, but I'm not aware of the other considerations here.

canton7:
Ah, damnit, we lost special support for a timeout of 0 along the way. Incoming...

lib/sync/sync.go
calmh: No, it'll end up on the heap either way in this case. The only difference is if it gets copied a couple of times on the way or not. The copies are essentially free so this doesn't matter, but it looks less surprising to have the pointer return type (and just `return &TimeoutCond{ ... ` etc).

calmh:
Not in the sense that the extra timer is much to worry about, but by setting up a timer with a zero interval (that'll fire immediately) it might become a race condition agains the channel read?

canton7:
Hmm. There's only a race there if someone requests events with a zero timeout at the same instant that an event is created. In that case there's a race regardless of what we do, so it doesn't matter..?

cmd/syncthing/gui.go
nrm21: > While I'm here, is it worth reducing the size of the disk events buffered subscription to 25, as we only ever fetch 25 events? Or not worth the effort?

When I wrote that code I made the API buffer 1000 events and just use the first 25.  My thoughts were that maybe someone else could use more info if they wanted to include a feature that used more.  Or something else I just hadn't thought of.  I was future proofing, but I have no problem with you limiting it to just 25 if you feel that's better.

canton7:
Anything else you want me to do to this?

lib/events/events.go
```diff
@@ -319,12 +315,21 @@ func (s *bufferedSubscription) pollingLoop() {
...
+		waiter := s.cond.SetupWait(timeout)
+		defer waiter.Stop()
+
+		for id >= s.cur {
+			if eventsAvailable := waiter.Wait(); !eventsAvailable {
+				// Timed out
+				return into
+			}
+		}
 	}
```
AudriusButkevicius: I'd say that the optimization of not creating a waiter makes mental stack management too hard, and I'd sa that we should just create the waiter in all cases to make the code easier to read.

AudriusButkevicius:
@st-review lgtm

st-review:

lib/events/events.go
canton7: I'll leave that one up to @calmh - when the events are coming thick and fast Syncthing will hit the events endpoint a *lot* in quick succession.

calmh:
I started writing some tests to make sure I understand the timoutcond. Should this pass?

https://gist.github.com/calmh/c712bf46d26acb56d3adfe87101eb771

calmh:
The timeoutcond seems to work at least; here's a test case if you want to add it. If you can think of better ways to test it, please add to it.

https://gist.github.com/14f6ff12ffbedccaf5e531ba045cd6b7

canton7:
Cheers - I'll add it. 

@st-review don't merge this just yet, please

canton7:
Added, thanks!

canton7:
Any idea what could be going on with that test failure? I can't reproduce it locally...

canton7:
Hmm, increasing the sleep in TestBufferedSub seems to have fixed it. Build server more loaded than normal, perhaps?

calmh:
It dropped 46 events... I can't immediately reproduce it either but I would guess it's caused by the timing that you're already on to. Maybe the build server was extra loaded there for a while.

calmh:
New failure is probably Windows timer or clock granularity not being great...

AudriusButkevicius:
```
--- FAIL: TestBufferedSub (0.43s)
	events_test.go:225: Incorrect ID; 602 != 556
FAIL
FAIL
```

canton7:
Grr, hate timing-based tests...

I think I'm going to revert that modification I made to TestBufferedSub - since it "fixed" it, I don't think that that failure is anything to do with me.

calmh:
This one is interesting though:

sync_test.go:298: Wait failed early, 33.378��s < 4ms

That's mangled unicode for microseconds. Wonder why it immediately returned false instead of blocking for at least 4 ms?

canton7:
That could happen if the thread's stalled between the `c.SetupWait(d)` and the `t0 := time.Now()`, since the timer starts at the point that `SetupWait` is called.

calmh:
Lets change the order of those

canton7:
I was going to, but I'm not sure that it helps. If the thread is stalled between the `time.Now()` and the `SetupWait()`, then you'll get the opposite issue, no?

canton7:
I will however add a call to `w.Stop()` - we might get a bazillion timers all finishing at roughly the same time, which might be messing up the scheduling.

calmh:
Yeah. Can be worked around I guess by increasing the values so that any stalls are in the noise. For now, maybe comment out the section that does the test failures with a comment that it's not reliable, and leaving the rest in. Even without those checks the final table can be eye-balled for plausibleness and it gets exercise under the race detector.

But looking at this it seems we might as well call `c.SetupWait(timeout).Wait()` since the waiter isn't usefully recycled, which brings me to the question of whether we need to expose the setup step at all?

canton7:
The setup step is needed in practice, because the buffered subscription needs to call `.Wait()` multiple times for the same timeout (and the same timer). It needs to do this because `.Wait()` returning doesn't necessarily mean that an event with a high enough ID is available yet, so we might need to loop around and try again.

calmh:
Okay, carry on.

canton7:
Reducing the number of iterations seems to help, but that's probably because we vastly lower the changes of hitting this.

I'll do what you suggested.

calmh:
@st-review lgtm

st-review:

st-review:
@calmh: Merge failed:

```
$ git fetch -f origin refs/pull/3941/head:pr-3941
   35e87e2..20f8b4f  master     -> orig/master
   1c9361a..20f8b4f  master     -> origin/master
$ git reset --hard
HEAD is now at 1c9361a cmd/syncthing: Implement "release candidate" logic
$ git checkout master
Already on 'master'
Your branch is behind 'origin/master' by 1 commit, and can be fast-forwarded.
HEAD is now at 20f8b4f gui: Wording of auto upgrade selections
$ git clean -fxd
$ git merge --squash --no-commit pr-3941
Auto-merging cmd/syncthing/main.go
Auto-merging cmd/syncthing/gui.go
$ git commit -F -
Commit message subject:

    WIP towards timeout on events subscriptions

match this pattern:

    ^[\w/,\. ]+: \w

```

canton7:
@st-review merge lib/events: Add timeout to events API, remove Ping (fixes #3933)

Previously, the Ping event was used to ensure that something appered on the events API at least every minute, so that requests to the events API would take at most a minute. However this is less than ideal in some cases. If there are very few events, then ping events can dominate, meaning it's hard to find other events. Filtering out ping events from the disk events API also makes consumers more complex.

Instead, allow a timeout to be specified when requesting events. If no events are found within this timeout, then an empty array is returned. The timeout defaults to 60 seconds (for backwards compatibility), and any invalid value (<0, non-numeric) is also treated as 60 seconds. A timeout of zero is allowed, in which case the events API because request-response instead of long-polling, returning immediately if no events are available.

In order to facilitate this, a new structure in lib/sync, TimeoutCond, was created. This behaves similarly to Cond, except that it allows the user to specify a timeout. One timeout will span multiple calls to .Wait().

st-review:
@canton7: Merge failed:

```
$ git fetch -f origin refs/pull/3941/head:pr-3941
$ git reset --hard
HEAD is now at 20f8b4f gui: Wording of auto upgrade selections
$ git checkout master
Already on 'master'
HEAD is now at 20f8b4f gui: Wording of auto upgrade selections
$ git clean -fxd
$ git merge --squash --no-commit pr-3941
Auto-merging cmd/syncthing/main.go
Auto-merging cmd/syncthing/gui.go
$ git commit -F -
Commit message subject:

    Previously, the Ping event was used to ensure that something appered on the events API at least every minute, so that requests to the events API would take at most a minute. However this is less than ideal in some cases. If there are very few events, then ping events can dominate, meaning it's hard to find other events. Filtering out ping events from the disk events API also makes consumers more complex.

match this pattern:

    ^[\w/,\. ]+: \w

```

st-review:
@canton7: Merge failed:

```
$ git fetch -f origin refs/pull/3941/head:pr-3941
$ git reset --hard
HEAD is now at 20f8b4f gui: Wording of auto upgrade selections
$ git checkout master
Already on 'master'
HEAD is now at 20f8b4f gui: Wording of auto upgrade selections
$ git clean -fxd
$ git merge --squash --no-commit pr-3941
Auto-merging cmd/syncthing/main.go
Auto-merging cmd/syncthing/gui.go
$ git commit -F -
Commit message subject:

    Previously, the Ping event was used to ensure that something appered on the events API at least every minute, so that requests to the events API would take at most a minute. However this is less than ideal in some cases. If there are very few events, then ping events can dominate, meaning it's hard to find other events. Filtering out ping events from the disk events API also makes consumers more complex.

match this pattern:

    ^[\w/,\. ]+: \w

```

calmh:
@st-review merge it like this then

cmd/syncthing, lib/events, lib/sync: Add timeout to REST event API, remove Ping (fixes #3933)

st-review:
:ok_hand: Merged as ac510b26e259271d2d0b530a7d3e46a3169a8cf1. Thanks, @canton7!

canton7:
How come it works when you do it, but not when I do it? :disappointed:

calmh:
Commit subject needs to be on a line by itself after the command. I didn't see your post in time or I've have injected the longer message, sorry.

canton7:
Hmm, I edited it to put the subject on its own line, and the bot responded to it, but I think it took the old version of my comment again...

calmh:
Possibly

canton7:
Anyway, thanks for your help getting this in!

@nrm21 Now that we've got `?timeout=0`, do you think it makes sense to fetch from the disk events API with a zero timeout whenever the "Global Changes" dialog is opened, rather than doing a long poll every time LOCAL_INDEX_UPDATED is received?

calmh:
I wouldn't worry too much about the timeout, it'll only fire when there are zero events in the buffer (i.e. at startup), and then it'll anyway return an empty list in the end.

canton7:
True, fair enough.

PR #3942, opened 2017-01-27
Average bps

benshep:
Calculate average bps using a FIFO buffer instead of just basing it on the last two byte counts. Use this to show an esimate of the ETA for syncing folders.


benshep:
Oops, sorry - that's more commits than I was expecting to see.

calmh:
Better, I think, to do the calculation on the "server" side using the `EWMA` we have in use in other places and exposing it in the connection stats struct.

benshep:
Hmm, OK. So you would replace AddInt64 in the countingReader and countingWriter objects with a byteCounter object? Seems sensible.

calmh:
That sounds reasonable. Might need some scaffolding or moving a type or two or something, not sure, but it's an approach I would 👍 at this point

PR #3945, opened 2017-01-30
gui, lib/config: Add notification about new release channels

calmh:
### Purpose

As above

### Testing

Clicketyclick

### Screenshots

screen shot 2017-01-30 at 23 12 09

The link goes to https://build.syncthing.net/docs/223/users/releases.html


lib/config/config_test.go
```diff
@@ -200,7 +200,7 @@ func TestOverriddenValues(t *testing.T) {
 		AlwaysLocalNets:         []string{},
 		OverwriteRemoteDevNames: true,
 		TempIndexMinBlocks:      100,
-		UnackedNotificationIDs:  []string{},
+		UnackedNotificationIDs:  []string{"channelNotification"}, // added in 17->18 migration
```
AudriusButkevicius: Suggest you break this over multiple lines so that we end up with:

```
[]string{
  "foo", //bar
  "baz". //nuf
}
```

AudriusButkevicius:
@st-review lgtm

st-review:

lib/config/testdata/v18.xml
```diff
@@ -0,0 +1,15 @@
+<configuration version="17">
```
imsodin: v18?

AudriusButkevicius: Its 17 to 18 migration

imsodin: Ok. I was simply comparing to v17.xml where the numbers matched, no "deeper insights" :)

calmh: No you're right it should be 18. Fixed. Also another thing, the override test was flaky.

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 3eb7a9373ace20017b2bb4f781e94431f24089d9. Thanks, @calmh!

PR #3948, opened 2017-01-31
lib/config: Disable weak hashing by default (fixes #3938)

### LINKED ISSUE #3938: Weak hashing is extremely slow on ARMv8 (Android)
```
[ZYPSC] 03:41:31 INFO: Single thread SHA256 performance is 559 MB/s using minio/sha256-simd (10 MB/s using crypto/sha256).
[ZYPSC] 03:41:33 INFO: Actual hashing performance is 84.14 MB/s
```

Weak hashing slows the actual hashing down almost an order of magnitude. It seems there is no hardware acceleration for it. I'm not sure what the requirements for a weak hash function are, but it could be much faster if it's hardware accelerated.

According to  `/proc/cpuinfo`, these features are supported:
```
Features        : fp asimd evtstrm aes pmull sha1 sha2 crc32
```

Maybe we can use one of those for the weak hash?

### Version Information

    Syncthing Version: v0.14.21
    Android Version: Android 7.0
    Device: LG G4 (Snapdragon 808)

calmh:
### Purpose

More fast

### Testing

Not really...

### DIFF
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -108,7 +108,7 @@ func TestDeviceConfig(t *testing.T) {
 				Versioning: VersioningConfiguration{
 					Params: map[string]string{},
 				},
-				WeakHashThresholdPct: 25,
+				WeakHashThresholdPct: 101,
 			},
 		}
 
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
If you plan a release now sure, otherwise I'd like to do the benchmark based version instead?

calmh:
Feel free to do that, there's no immediate rush. But I think that needs some intelligence on the receiving side as well. There's no point to be reading and weakhashing a file to look for matches if all of the weakhashes we have from the other side are zero.

AudriusButkevicius:
Sounds good.

PR #3949, opened 2017-01-31
lib/weakhash, lib/model, cmd/syncthing: Decide if to use weakhash on startup (fixes #3938)

### LINKED ISSUE #3938: Weak hashing is extremely slow on ARMv8 (Android)

AudriusButkevicius:
...


lib/weakhash/weakhash.go
```diff
@@ -45,7 +49,7 @@ func Find(ir io.Reader, hashesToFind []uint32, size int) (map[uint32][]int64, er
 
 	offsets := make(map[uint32][]int64)
 	for _, hashToFind := range hashesToFind {
-		offsets[hashToFind] = nil
+		offsets[hashToFind] = make([]int64, 0, maxWeakhashFinderHits)
```
AudriusButkevicius: This is to reduce allocations:

```
	var foo []int64

	for cap(foo) < 10 {
		foo = append(foo, 1)
		fmt.Println(len(foo), cap(foo))
	}
```

```
1 1
2 2
3 4
4 4
5 8
6 8
7 8
8 8
9 16
```

cmd/syncthing/usage_report.go
```diff
@@ -113,7 +113,8 @@ func reportData(cfg configIntf, m modelIntf) map[string]interface{} {
 	var mem runtime.MemStats
 	runtime.ReadMemStats(&mem)
 	res["memoryUsageMiB"] = (mem.Sys - mem.HeapReleased) / 1024 / 1024
-	res["sha256Perf"] = cpuBench(5, 125*time.Millisecond)
+	res["sha256Perf"] = cpuBench(5, 125*time.Millisecond, false)
+	res["hashPerf"] = cpuBench(5, 125*time.Millisecond, true)
```
AudriusButkevicius: Not sure if we should reset UR because of this, as technically the value of sha256Perf has changed when we started using scanner.Blocks oppose to sha256 package.

lib/weakhash/weakhash.go
calmh: Mmm. Back of envelope calculation says this allocates ~640 KiB per GiB of input file (1024*1024/128*80), which should be fine.

cmd/syncthing/usage_report.go
calmh: No

lib/config/optionsconfiguration.go
```diff
@@ -43,6 +43,7 @@ type OptionsConfiguration struct {
 	TempIndexMinBlocks      int      `xml:"tempIndexMinBlocks" json:"tempIndexMinBlocks" default:"10"`
 	UnackedNotificationIDs  []string `xml:"unackedNotificationID" json:"unackedNotificationIDs"`
 	TrafficClass            int      `xml:"trafficClass" json:"trafficClass"`
+	WeakHashSelectionMethod int      `xml:"weakHashSelectionMethod" json:"weakHashSelectionMethod"`
```
calmh: Magic numbers are magic. This should be an enum type with constants that serialize to a string in the config. And probably something like `WeakHashEnabled` with possible values `auto`, `yes`, `no` (yes and no being called something else if you prefer)? Yes there's some overlap with the weak hash threshold, but it would then be back to defaulting to 25 and it was never very intuitive to have to set a percentage to >100 to disable something.

AudriusButkevicius: Yeah, sadly util.SetDefaults won't be able to handle that, unless some really dark witchcraft.

AudriusButkevicius: Nvm, worked it out.

lib/config/optionsconfiguration.go
```diff
@@ -6,43 +6,132 @@
...
+import (
+	"encoding/json"
+	"encoding/xml"
+	"fmt"
+)
+
+type WeakHashSelectionMethod int
+
+const (
+	WeakHashAuto WeakHashSelectionMethod = 1 << iota
```
calmh: No need for bitmasking here I think; just `= iota` and the default value becomes zero, no SetDefaults shenanigans required.

AudriusButkevicius:
I've left the ParseDefault as it might turn out to be useful in the future.

calmh:
This is in principle good apart from not actually compiling and stuff

AudriusButkevicius:
@st-review merge if it passes

st-review:

st-review:
:ok_hand: Merged as 67acef1794c88879452e3752ab16f313dc15609e. Thanks, @AudriusButkevicius!

PR #3960, opened 2017-02-05
lib/events: Be more resilient against dropping events (fixes #3952)

### LINKED ISSUE #3952: lib/events: TestBufferedSub fails spuriously
Since the new buffered sub, but not super clear why. Is it a little bit slower?

```
Test with coverage
ok  	github.com/syncthing/syncthing/lib/auto	0.019s	coverage: 100.0% of statements
ok  	github.com/syncthing/syncthing/lib/beacon	0.004s	coverage: 4.9% of statements
ok  	github.com/syncthing/syncthing/lib/config	0.151s	coverage: 70.3% of statements
ok  	github.com/syncthing/syncthing/lib/connections	0.009s	coverage: 3.4% of statements
ok  	github.com/syncthing/syncthing/lib/db	0.486s	coverage: 57.2% of statements
ok  	github.com/syncthing/syncthing/lib/dialer	0.004s	coverage: 10.1% of statements [no tests to run]
ok  	github.com/syncthing/syncthing/lib/discover	5.290s	coverage: 46.1% of statements
--- FAIL: TestBufferedSub (0.08s)
	events_test.go:225: Incorrect ID; 92 != 71
FAIL
coverage: 66.7% of statements
```

calmh:
Instead of just immediately dropping the event if the subscription isn't
ready to receive it, give it 15 ms to catch up. The value 15 ms is
grabbed out of thin air - it just seems reasonable to me.

The timer juggling makes the event send pretty much exactly twice as
slow as it was before, but we're still under a microsecond. I think it's
negligible compared to whatever event that just happened that we're
interested in logging (usually a file operation of some kind).

	benchmark                  old ns/op     new ns/op     delta
	BenchmarkBufferedSub-8     475           950           +100.00%

	benchmark                  old allocs     new allocs     delta
	BenchmarkBufferedSub-8     4              4              +0.00%

	benchmark                  old bytes     new bytes     delta
	BenchmarkBufferedSub-8     104           117           +12.50%

### DIFF
--- a/lib/events/events.go
+++ b/lib/events/events.go
@@ -51,6 +51,8 @@ const (
 
 var runningTests = false
 
+const eventLogTimeout = 15 * time.Millisecond
+
 func (t EventType) String() string {
 	switch t {
 	case Starting:
@@ -122,6 +124,7 @@ type Logger struct {
 	subs                []*Subscription
 	nextSubscriptionIDs []int
 	nextGlobalID        int
+	timeout             *time.Timer
 	mutex               sync.Mutex
 }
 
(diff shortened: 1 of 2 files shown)


calmh:
@st-review merge it, i believe in myself

PR #3972, opened 2017-02-08
lib/model: Increase in-flight buffer from 2mb to 8mb, download 2 files at a time

AudriusButkevicius:
Because why not.

### DIFF
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -67,8 +67,8 @@ const (
 )
 
 const (
-	defaultCopiers     = 1
-	defaultPullers     = 16
+	defaultCopiers     = 2
+	defaultPullers     = 64
 	defaultPullerSleep = 10 * time.Second
 	defaultPullerPause = 60 * time.Second
 )


calmh:
I agree increasing pullers is probably at worst harmless. Increasing copiers can help but also significantly hurt - copying two large files in parallel on a spinning disk usually sucks. I'd like some sort of reasoning behind why it's a win.

AudriusButkevicius:
It's not really a win, it was more to address the one fast one slow peer problem.
We are already fetching blocks in random order which is suboptimal, so I don't think it makes it much worse. Once we solve the random download order issue, we can address this too, but I somehow think we rarely hit disk throughput, and readahead should be able to cope with reading a few mb from separate files.

calmh:
Let's try it and see who screams

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 72de47df006d2fa1a4f4a6797aa94a1e73f80d5c. Thanks, @AudriusButkevicius!

calmh:
Hah. I can merge via email. \o/

Issue #3964, opened 2017-02-06
GUI: Add "Pause All Folders" and "Resume All Folders" buttons


scienmind:
Implement "Pause All Folders" and "Resume All Folders" buttons

Use cases:
https://github.com/syncthing/syncthing/pull/3520#issuecomment-263859409
https://forum.syncthing.net/t/pause-all-folders/9170

imsodin:
Fixed in #3973 with https://github.com/syncthing/syncthing/commit/0c68e1e510e2e92beed870b3ea30fb047d6dabcf. Also relevant: #3982

PR #3973, opened 2017-02-08
GUI: Added "Pause All Folders" and "Resume All Folders buttons

Niller303:
As per scienmind's post (#3964) i though it was a good idea to add such buttons and did so.


gui/default/index.html
```diff
@@ -437,6 +437,12 @@ <h4 class="panel-title">
           </div>
         </div>
         <span class="pull-right">
+          <button type="button" class="btn btn-sm btn-default" ng-click="pauseAllFolders()">
+            <span class="fa fa-refresh"></span>&nbsp;<span translate>Pause All Folders</span>
+          </button>
+          <button type="button" class="btn btn-sm btn-default" ng-click="resumeAllFolders()">
+            <span class="fa fa-refresh"></span>&nbsp;<span translate>Resume All Folders</span>
+          </button>
```
AudriusButkevicius: I wonder if we should hide the buttons if all folders are resumed or paused etc?

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1661,6 +1661,26 @@ angular.module('syncthing.core')
             $http.post(urlbase + "/db/scan?folder=" + encodeURIComponent(folder));
         };
 
+        $scope.pauseAllFolders = function () {
+            // loop through all folders and pause them
+            var folderListCache = $scope.folderList();
+            for (var i = 0; i < folderListCache.length; i++) {
+                if (!folderListCache[i].paused) {
+                    $scope.setFolderPause(folderListCache[i].id, true);
```
AudriusButkevicius: If you have 100s of folders, this will be very inefficient.
If you have a 100 folders, it will post the config, emit events, reload half of the services for each folder.
If you have 100s of folders, this will be very inefficient.
If you have a 100 folders, it will post the config, emit events, reload half of the services for each folder.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1664,21 +1664,31 @@ angular.module('syncthing.core')
...
 
         $scope.resumeAllFolders = function() {
             // loop through all folders and start them
             var folderListCache = $scope.folderList();
+
             for (var i = 0; i < folderListCache.length; i++) {
-                if (folderListCache[i].paused) {
-                    $scope.setFolderPause(folderListCache[i].id, false);
+                var folder = folderListCache[i];
+                if (folder && !folder.paused) {
```
imsodin: The `!` shouldn't be there.

AudriusButkevicius:
I think it's better to have a single `setAllFolderPause(bool)`, also no need to check for `folder && folder.paused`. Folder has to be there as you are indexing, and you are overwriting the paused state so it doesn't matter what it's set to.

gui/default/index.html
Niller303: I just don't know how this behaviour would function, would it just default to pause/resume all?
Or would it switch attribute when you pressed it?

Niller303: Maybe a new dropdown menu is also an option for managing all items

gui/default/index.html
```diff
@@ -437,6 +437,12 @@ <h4 class="panel-title">
           </div>
         </div>
         <span class="pull-right">
+          <button type="button" class="btn btn-sm btn-default" ng-click="pauseAllFolders()">
+            <span class="fa fa-refresh"></span>&nbsp;<span translate>Pause All Folders</span>
+          </button>
+          <button type="button" class="btn btn-sm btn-default" ng-click="resumeAllFolders()">
```
AudriusButkevicius: Just setAllFolderPause(false) and setAllFolderPause(true) in the other one, and remove the two functions.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1661,6 +1661,20 @@ angular.module('syncthing.core')
             $http.post(urlbase + "/db/scan?folder=" + encodeURIComponent(folder));
         };
 
+        $scope.setAllFolderPause = function(pause) {
+            var folderListCache = $scope.folderList();
+
+            for (var i = 0; i < folderListCache.length; i++) {
+                var folder = folderListCache[i];
+                if (folder) {
```
calmh: There should be no circumstances where an entry in the folderlist is null/undefined?

calmh:
@Niller303 Do you have a name for the AUTHORS file?

gui/default/syncthing/core/syncthingController.js
Niller303: I don't know, in $scope.setFolderPause it also checks if the folder is valid.
I also don't have my name in the authors file.

uok:
Helpful improvement! 
For good user experience it would be great if the "resume all" button is only displayed if one or more folders are paused

scienmind:
> For good user experience it would be great if the "resume all" button is only displayed if one or more folders are paused

Welcomed tweak :)
Same goes for "pause all" : available if one or more is unpaused (otherwise, dimmed/unavailable).

Niller303:
To my experience i find buttons disappearing and appearing annoying.
I would much rather have the buttons have a "light down" effect when its unavailable.

For the AUTHORS file i would have:: 
Niels Peter Roest (niller303) <nielsproest@hotmail.com>

calmh:
That's valid reasoning, but not how the rest of the GUI already works.

(Added to authors.)

canton7:
> That's valid reasoning, but not how the rest of the GUI already works.

It isn't? Don't the Remove / Ignore Patterns / Save disable themselves (rather than disappearing) if they can't be clicked?

calmh:
Yes, form save buttons get disabled by the form being in an invalid state. But if you expand a folder you'll see a pause button only if it's not paused, a resume button when it's paused, a scan button when it's idle, etc.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1670,7 +1670,19 @@ angular.module('syncthing.core')
...
             $scope.saveConfig();
-        }
+        };
+
+        $scope.getAllFolderPauseState = function(pause) {
+            var folderListCache = $scope.folderList();
+
+            for (var i = 0; i < folderListCache.length; i++) {
+                if (folderListCache[i].paused == pause) {
+                    return true;
```
AudriusButkevicius: Should this be not equals and return false?

AudriusButkevicius: Also from the naming convention maybe the method should start with is/are such as areAllFolderPausedStateSetTo(), but don't care much about this.

Niller303: The name is now unreasonably long as to explain what it does

AudriusButkevicius:
@st-review lgtm

st-review:

gui/default/syncthing/core/syncthingController.js
calmh: Dat name though :D

calmh:
@st-review lgtm

st-review:
:ok_hand: Merged as 0c68e1e510e2e92beed870b3ea30fb047d6dabcf. Thanks, @Niller303!

Niller303:
Perhaps the text should be changed to just "Pause All" and "Resume All" without the "Folder part". So it dosen't use as much space.

uok:
Good point, this would fit with existing "Rescan All"

AudriusButkevicius:
Feel free to make a pr?

PR #3981, opened 2017-02-10
lib/model: Use ignore patters in remote completion calculation

calmh:
Basically, if we don't care about the sync status of the file we should
not tag someone else out of sync because they don't have the latest
version. This solves *my* "Syncing - 100%" scenario at least.

The reason this happens seems to be like this, in my situation. I have
three devices, connected in a "line": A-B-C. A is a Mac and litters
.DS_Store files everywhere. I've ignored these, but some escaped into
the folders before I did so. I've also ignored them on B and C but at
different stages. B was flagging C as out of sync, because at the point
the ignores were introduced C had a lower version of .DS_Store than A.
Now none of them are sending updates about it any more since it's
ignored...

### Testing

Manually verified... This shit is cumbersome to test mechanically, we should fix that...

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -486,6 +486,7 @@ type FolderCompletion struct {
 func (m *Model) Completion(device protocol.DeviceID, folder string) FolderCompletion {
 	m.fmut.RLock()
 	rf, ok := m.folderFiles[folder]
+	ignores := m.folderIgnores[folder]
 	m.fmut.RUnlock()
 	if !ok {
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Why not

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 89f8be40c6513292c5a53f5807591c40be75684d. Thanks, @calmh!

PR #3995, opened 2017-02-21
cmd/syncthing, lib/model: Handle rel/abs paths for config/protected paths (Fixes #3183)

### LINKED ISSUE #3183: lib/model: "overwriting protected files" warning doesn't properly take ignores into account
### What happened

When I'm syncing my home directory I need to exclude the syncthing configuration files. So I add the following to my ignore list:
`/.config/syncthing`

However, this results in the following warning on start:

```
Mai 25 12:53:08 book syncthing[28921]: [6EVKI] WARNING: Some protected files may be overwritten and cause issues. See http://docs.syncthing.net/users/config.html#syncing-configuration-files f
or more information. The at risk files are: /home/hashworks/.config/syncthing/index-v0.13.0.db, /home/hashworks/.config/syncthing/config.xml, /home/hashworks/.config/syncthing/cert.pem, /home
/hashworks/.config/syncthing/key.pem
```

When I change the ignore list entry to the following the warning won't occur after a restart.
`.config/syncthing`
### What I expect

The warning shouldn't occur when I'm syncing `~` and I ignored `/.config/syncthing`, just as with `.config/syncthing`.
### Steps to reproduce
1. Sync your home directory
2. Watch the configuration sync warning appear
3. Stop syncthing
4. Add `.config/syncthing` to your ignore list
5. Start syncthing
6. Make sure the configuration sync warning doesn't appear
7. Stop syncthing
8. Append `.config/syncthing` with `/`
9. Start syncthing
10. Watch the configuration sync warning appear
### Additonal

I haven't testet if the configuration files actually get synced with `/.config/syncthing`, they won't with `.config/syncthing`. Could be that only the configuration sync check is missing a rule.
Edit: I've testet the `/.config/syncthing` entry, as I expected the rule works as it should, the folder isn't getting synced. It's just the warning message that is missing a rule for that case.
### Version Information

Syncthing Version: v0.13.2
OS Version: Archlinux

imsodin:
This ensures that the protected paths are relative to the folder root when checking if they are ignored.

While testing I did not get the "overwriting protected files" warning as I set the config directory via `-home=./config`. This path is then kept in relative form and therefore isn't considered as a subfolder ([model.go:283](https://github.com/syncthing/syncthing/compare/master...imsodin:protectedIgnore?expand=1#diff-1df9805b2381b6288b220226f82d5dc9R283)). This is fixed by calling `filepath.Abs` on the path given on the command line.


calmh:
The commentary and stuff in main seems off. In general I don't think we can require an absolute config path. For example it would break my workflow:

$ cd test
$ syncthing -home h1

(Unless I misunderstand what this does)

imsodin:
Your use case still works, this does not require an absolute path, it converts the relative one to an absolute path, by using `filepath.Abs(relpath)`, which essentially prepends the current working dir. Consistent absolute paths seem to be expected in parts of the code, as the `string.HasPrefix` checks don't work with a mixture of relative and absolute paths.

That comment entered by copy&paste from folderconfig.go, can remove that of course.

calmh:
Yeah, please do. Now that I read it again it makes sense, but it discusses a performance aspect that is  irrelevant when called once in main. It seems to work fine though so my other concern is a non-issue 👍

calmh:
@st-review merge it

st-review:

st-review:
:ok_hand: Merged as c20d612736aee5e4472deb20008bf9b7a18ad5f2. Thanks, @imsodin!

Issue #2491, opened 2015-11-19
Next Gen Ignores: Requirements


calmh:
This issue replaces the ones that relate to fundamental issues with how our ignores work. The following is the up to date list of requirements:

Requirements:

- Should be understandable by someone who understands the interface in use by DropBox, OneDrive, etc.
- Should be a simple tree view with checkboxes
  - Root of tree (=folder root) checked by default
  - Can uncheck root and then check just the branches that we want synced
- Should be able to white or black list extensions
  - Globally, or per directory?
  - "Never sync `*.tmp`"
  - "Only sync `*.doc`"?
- Should live in config (not .stignore files)
- Should sync between devices
  - Maybe?
  - But must also be able to keep it separate

Comments and discussion below. A project maintainer will keep the list above in sync as the requirements change or are clarified.

crccheck:
Why not also keep `.stignore`, but make it sync between devices as `.gitignore` and `.hgignore` do? That way, there's a _developer friendly_ way to keep files out. As someone who had to delete 100k `node_modules` files yesterday (not to mention `.DS_Store`, `packages`, `build`, `.tox`, etc), that would have save me 3 gigs of bandwidth and time.

There's two distinct user stories:
1. As a user to Syncthing who also knows Dropbox, I want a familiar interface for selectively syncing files so I only sync what I care about
2. As a developer using Syncthing, I want a familiar interface for selectively syncing files globally so I only sync what I care about without additional set up every new install.

calmh:
A `.fooignore` file might come naturally to developers, but that's really the only demographic that expects a hidden file with a bizarre extension (think Windows). I think it's a developer accident that it exists at all in it's current form today - it's really a piece of folder configuration as anything else. Whether it syncs or not is really somewhat beside the point of the existence of the file, but I do think there should be at least an option to sync the patterns, yes.

(You can do that today with a little #include hackery as well.)

cdhowie:
I don't care where the data is stored, but patterns allow me to express ignores in a concise manner and with more flexibility than a tree option.  Please at least retain this functionality, if not in a file called `.stignore` then somewhere else.  None of the options provided in this ticket summary will meet my needs.

If the current pattern-based ignore option is removed I'm afraid I may have to stop upgrading Syncthing.

calmh:
@cdhowie What are the needs, that are not covered by the tree options and the extension white/black lists?

cdhowie:
[...]

PR #3996, opened 2017-02-21
lib/model, gui: Allow creating and editing ignores of paused folders

imsodin:
### Purpose

This makes it possible to add ignores in the webGUI before the first scan happens to prevent unwanted files to enter the index. The problem with editing ignores before actually creating the folder is, that the folder path is not yet definitive while it is necessary to get the contents of a possibly existing `.stignore` file there. Therefore I went along with @bugith's suggestion to add an option to save a folder in paused state.
When introducing folder pause the folder was only added to the model if it wasn't paused. That prevents calls to the rest API to get/set ignores. I didn't see any adverse affects of adding paused folders to the model and tests passed. Is there a particular reason for leaving paused folders out (@AudriusButkevicius)?

### Testing

Manual poking around in GUI.

### Screenshots

http://i.imgur.com/Vjz9Zqm.png


AudriusButkevicius:
I am pretty sure there is a reason, yet I can't check them on the phone. One that comes to mind is remote peers being able to ask data from a folder that is paused. Others might include sending the folder as being "available" in clusterconfig.

I think the better option here would be to fix ignores endpoints, as they don't need anything more than just the path to be able to work.

canton7:
I think the model of "Save as paused, then edit ignores, then unpause" is unnecessarily convoluted. The presence of the button, and the tooltip, raise more questions than they answer, I think.

imsodin:
Regarding Audrius' remark: I will undo that change and change the ignore endpoints as suggested. My plan is to move reading/writing ".stignore" entirely to the ignores package. Currently it is theoretically possible to write to the same ".stignore" at the same time - there is no locking. 

Regarding the GUI: I am aware it is not very straight forward and I am open to suggestions how to solve this more elegantly.
The problem I can't figure out in order to directly edit ignores: Unless an actual folder created (i.e. save pressed) loading existing ignores is not possible via the API. Maybe it would be possible to give a path instead of a folder id to the getIgnores endpoint to populate the ignores modal at folder creation time, but that's ugly as well (what happens when the user subsequently changes the folder path, does everything vanish?).

AudriusButkevicius:
It's probably easiest to somehow relax the ignores endpoint so that we use the same button, but fire off at the ignore endpoint before at the config endpoint, or add a new /folder/create endpoint that takes folder config plus ignores and handles that internally.

imsodin:
@AudriusButkevicius That still does not solve the problem of a possible preexisting ."stignore" file. When there is an option to edit ignores during folder creation, there must be a way to read the ".stignore" file at the currently given path. And if that was made possible, how would you handle a scenario like edit ignores -> change folder path to a directory where there already is a ".stignore" file -> edit ignores again. Just display the contents of ".stignore" and drop what the user entered earlier?

imsodin:
I adjusted the endpoints such that in case there is no folder config found in the model, they directly read/write to the ".stignore". I did not adjust the tests, I first wanted to get feedback on whether this approach is viable.

To that end I introduced an additional member `list` to the `ignores.Matcher` struct which contains the lines of ".stignore". Previously `getIgnores` in model.go would return patterns from the ignores struct and lines from the ".stignore" file, which is inconsistent. Now both can be obtained from `Matcher`. To avoid duplication, I created a function to write to ".stignore" in the ignores package. Is this fine or is there a more appropriate place?

bugith:
Would the following path (or something similar) on folder creation be a workaround ? : create all folders with a simple "*" ignore pattern list, then show in GUI a dedicated message in this very folder section stating the folder was "Just created or is restricted to everything - Remove this restriction now/LATER ?"

AudriusButkevicius:
No that makes adding a folder in the generic case much worse.

bugith:
I felt this too. OTOH I thought, well let's see that as the ultimate warning : "This is you last chance to not share everything already in this folder" ;-)

imsodin:
I went with an additional button as it does not disrupt the existing workflow and it allows to create ignores before first scan. It isn't as nice as directly being able to edit ignores when creating a folder, but for already explained reasons, I didn't see how to do that and have not yet gotten any new ideas.

That said the GUI change could be factored out into a separate PR if this is entirely not suitable. The underlying change to allow getting/setting ignores of paused folders and making returned patterns and line contents consistent (both from either last loaded ignores or directly from .stignore, not a mixture of both) is IMO still valid. Any comments on that part?

uok:
Is it correct that the usage of ignores is not tracked/reported on [https://data.syncthing.net/](https://data.syncthing.net)?
IMHO adding a second button for a small percentage of users is strange as the rest now suddenly has a choice that needs to be documented. I also recommend not to use the same primary color as with "save"

imsodin:
Actually there is a simple solution: Add an option "Save & Set Ignores Patterns" instead of "Save Paused" that will bring up the ignores modal right after (which is now not a problem, as the folder both exists and its path is fixed).

It currently looks like this, but the name is lengthy and maybe there should be a visual distinction from normal save as well - I hope for suggestions.

AudriusButkevicius:
Why not just have exactly the same UI as we have now and  based on if ignores are set or not, do the logistics?

imsodin:
There is two problems:  
Firstly, as there is no folder yet, there is no way to get a possibly preexisting ".stignore" file at the path given. That could be circumvented by allowing the ignores endpoint to take a path instead of a folder id, but that seems out of place (but certainly easily doable).
Secondly, as there may or may not be an existing ".stignore" file at the path given for the folder. This is not a problem for an existing folder, where the path can no longer change, but it is for a new folder, where the folder path is still subject to change. This can lead to a scenario like user not setting the folder path but editing ignores, the setting the folder path to a location, where there is already a ".stignore" file. Now there is a collision where we need to either automatically merge the two (user input and existing file), e.g. by appending, or require another user interaction.

Both can easily be avoided by requiring the user to finish the folder setup before he can create the ignores, which is what the additional button does. Obviously instead of an additional button, there could be a checkbox option with a text like "Create ignore patterns after saving" or something similar. Would that be more suitable?

AudriusButkevicius:
The logistics of how the ignores end up there doesn't have to be related to the UX. You could hold the ignores in memory as user saves then, create paused folder, set ignores from memory, resume the folder.

imsodin:
Sure, one can do that. How do you propose to handle the case where there is already a .stignore file, i.e. there is one version entered by the user on creation in memory and another on disk in ".stignore"? Automatically overwrite or append (with possible notification)?  
Alternatively an interactive warning would be good, but requires some kind of diff/merge tool and I certainly don't have the time/UI skills to do that (especially when there is the easy option of just creating ignores after setting up the rest).

AudriusButkevicius:
I'd just overwrite it I guess.

imsodin:
Now it uses the usual interface for editing ignores and just overwrites a preexisting .stignore file (which is mentioned in the ignores modal). I needed to change more javascript code than I originally intended (I don't know it). It does work for me, but please look at it carefully.

scienmind:
Why not the following logic:

1. Enable a button to edit ignores (as present in "edit folder" after creation) in the creation dialog
When user confirms the creation by clicking "save":
2. Look for .stignore file in the target directory, if present - **append** the "fresh" user-entered ignores to it, if not present, create a new ,stignore.
3. Load the updated .stignore, and start scanning accounting for it.

It keeps the old flow.
Allows to create a folder with new ignores in relatively simple way.
Allows for users to "seed" `.stignore`s (useful for OS-junk templates), and even add to them during creation.

* one caveat: if .stignore is present with data, and user enters ingores as well during creation (that will be appended), the present .stignore data won't show up in the UI during the creation. Showing it introduces mentioned issues with possible directory path change. 
But I don't think it's a bit deal. Since if someone manually put .stignore file there, and manually entering more ignores in the creating dialog, it's probably intentional to keep doth ignores data.

calmh:
I tried to say this a while back but it got lost in cyberspace... reposting although it's now late in the game:

Would it make things easier or more complicated to store the ignores in the config instead?

(Some complications from the top of my head: handling includes; the transition (maybe read both, do the transition on first save, if there are ignores in the config those have precedence))

imsodin:
@scienmind That's what it does except that it overwrites instead of appends. I could add an "append" option to the ignores REST endpoint, but I first want to be sure this is what "the majority" wants.
http://i.imgur.com/8rRV6JW.png
http://i.imgur.com/7O4QnMv.png

@calmh This would remove the problem of preexisting ".stignore" file at the location, but I think it would be an annoyance for some users who primarily edit ignores via the ".stignore" file not the webGUI (for example me :) ). So I wouldn't go down that path (at least not until something like  #2491 gets tackled).

Looking at the second screenshot there is an issue with path separators. I tried to fix it with https://github.com/syncthing/syncthing/pull/3996/files#diff-d181f8a81bd64b810deead2ee11355ffR18 but for some reason that doesn't seem to work.

scienmind:
> Would it make things easier or more complicated to store the ignores in the config instead?

For:
1 - Sounds cleaner in general; centralized configuration store
2 - Allows to solve issues like this one with ease

Against:
1 - The current way of batch deployment of "OS-junk defaults" becomes more cumbersome (i.e. can't just paste files into a bunch of directories)
BUT - it depends on what becomes of "includes", if they'll still be supported, the above point is a non-issue.
2 - Potential downsides of a very big config (can't think of any :thinking: )

Overall sounds like a good idea to me :+1: 
Although more in the territory of this never-ending monster: [Next Gen Ignores: Requirements #2491](https://github.com/syncthing/syncthing/issues/2491)

scienmind:
@imsodin timing, ha :wink: 

> I could add an "append" option to the ignores REST endpoint, but I first want to be sure this is what "the majority" wants.

My thought is that appending data to an existing file won't surprise people who pre-seed ignores in folders today, and won't wipe off their .stignores. If somehow .stignore is already present in the target directory, my guess would be that that it was done on purpose.

I may be wrong though, can't speak for "the majority"...

Also there is a chance of double entry with appending (same stuff in .stignore + manually entered). I don't know how ST handles it today, but if it does it well, this is non-issue then.

AudriusButkevicius:
endsWitch is a typo, hence why it does not work.

imsodin:
@scienmind That "majority" thing sounds pretty awful. What I wanted to say was that I want to wait if there was any opposition ( @AudriusButkevicius ?) to that.

gui/default/syncthing/folder/editIgnoresModalView.html
```diff
@@ -14,12 +14,13 @@
     </dl>
   </div>
   <div class="modal-footer">
-    <div class="pull-left"><span translate>Editing</span> <code>{{currentFolder.path}}{{system.pathSeparator}}.stignore</code></div>
+    <div class="pull-left" ng-show="!editingExisting"><span translate>Creating ignore patterns at</span> <code>{{currentFolder.path}}{{!currentFolder.path.endsWith(system.pathSeparator) ? system.pathSeparator : "" }}.stignore</code> <span translate>(overwrites existing file).</span></div>
```
uok: translation strings without punctuation marks please

imsodin: Thanks. This might change if it is decided to append instead of ignore, I will fix it then.

imsodin:
Now I added the append option to the ignores rest endpoint and made appending the default at folder creation, while overwriting can be activated. For existing folders nothing changed.
http://i.imgur.com/WPUn3nQ.png

scienmind:
I like the optional "overwrite" there :+1: 

Maybe add the filename of the file in question to the strings (to be more clear about it):
   Instead of : "By default this is appended to an existing file"
   Should it be : "By default this is appended to an existing `.stignore` file" ?

AudriusButkevicius:
I think you are wasting effort on the append stuff and should just back it out, and instead make sure we delete .stfolder, .stignores and even .stversions upon removing the folder (which has a separate ticket I think).

scienmind:
> make sure we delete .stfolder, .stignores and **even .stversions** upon removing the folder

deleting .stversions sounds quite dangerous, as it will wipe out stuf for people who use it for backup of any kind, and sometimes folder is being remove just to be later added again. But that's way off the topic for this PR...

imsodin:
@AudriusButkevicius Well the effort is already spent/wasted. As scienmind noted, removing internal files/dirs on folder removal is a separate issue (admittedly, adding an option to append to ignores wasn't originally intended either).

AudriusButkevicius:
I summon @calmh to decide, but I say we back out the append commit and merge this, if it works.

calmh:
*calmh appears, as summoned*

Aight let me take a look now that I'm back in the connected universe

cmd/syncthing/gui.go
```diff
@@ -968,10 +969,17 @@ func (s *apiService) getRandomString(w http.ResponseWriter, r *http.Request) {
...
 	qs := r.URL.Query()
 
-	ignores, patterns, err := s.model.GetIgnores(qs.Get("folder"))
+	folder := qs.Get("folder")
+	ignores, patterns, err := s.model.GetIgnores(folder)
 	if err != nil {
-		http.Error(w, err.Error(), 500)
-		return
+		matcher := ignore.New(false)
+		path := filepath.Join(s.cfg.Folders()[folder].Path(), ".stignore")
```
calmh: This looks dangerous. If the folder doesn't exist, `s.cfg.Folders()[folder].Path()` is the empty string and `.stignore` ends up in the current directory, whatever that happens to be for the running syncthing instance... It's probably a "file not found" here but the `POST` case below will create the file if it's missing, as far as I can see.

lib/model/model.go
```diff
@@ -1241,65 +1240,40 @@ func (m *Model) ConnectedTo(deviceID protocol.DeviceID) bool {
...
+	lines := m.folderIgnores[folder].Lines()
 	patterns := m.folderIgnores[folder].Patterns()
 	m.fmut.RUnlock()
 
 	return lines, patterns, nil
 }
 
-func (m *Model) SetIgnores(folder string, content []string) error {
+func (m *Model) SetIgnores(folder string, content []string, append bool) error {
+	m.fmut.RLock()
```
calmh: Good catch

lib/model/model_test.go
```diff
@@ -952,6 +952,9 @@ func TestIgnores(t *testing.T) {
 	m.ServeBackground()
 	defer m.Stop()
 
+	// Let initial setup operations (scan, ...) happen
+	time.Sleep(time.Duration(10) * time.Millisecond)
```
calmh: Very suspicious here. What are we waiting for exactly? If it's a scan, we should just do a scan instead as the 10 ms will probably not be enough on some random computer out there.

lib/osutil/atomic.go
```diff
@@ -50,6 +51,24 @@ func CreateAtomic(path string) (*AtomicWriter, error) {
...
+	w, err := CreateAtomic(path)
+	if err != nil {
+		return nil, err
+	}
+
+	orig, err := os.Open(path)
+	if err == nil {
+		io.Copy(w, orig)
+	}
+	orig.Close()
```
calmh: This is probably a nil pointer crash waiting to happen when the open fails; the close should be inside the if.

imsodin:
Why IMO loading ignores while creating a folder doesn't work:
While creating the folder the ignores cannot be obtained by the GET /rest/db/ignores endpoint, as there is not folder yet. That might be mitigated by adding a possibility to this endpoint to read ignores from an arbitrary path, but then the following (potential) problem arises due to the folder path not being fix yet:  
What happens when the user sets the folder path (already containing a .stignore) and then edits the ignores, after which he edits the folder path again (new preexisting .stignore)?

I will wait on the append comments until this is decided.

calmh:
So frankly I would just not worry about this. If the user edits ignores when creating the folder we overwrite whatever might already be present. If they don't, we don't. Someone having the foresight to create a .stignore file with important contents before adding a folder, and also editing the ignores to something else when adding the folder seems like a rare occurrence and they can suffer the consequences of their actions on their own. :)

lib/model/model_test.go
imsodin: Totally agree and I must admit, I don't know what it's waiting for exactly. Originally the error occurred at the second and third arrEqual tests. I suspected a race between setting and getting ignores first (as now `GetIgnores` picks the ignores up from memory not disk), but there should be no problem, as setIgnores should block until the new ignores are picked up in the scan it prompts in the end. The delay was introduces by trial and error. Now I removed it again for testing and it works locally, lets see what jenkins means.

imsodin:
So I removed the sleep and locally tests passed, but apparently not for jenkins

```
--- FAIL: TestIgnores (0.12s)
	model_test.go:982: Incorrect ignores: [.* quux] == [.* quux]
	model_test.go:986: Incorrect ignores: [.* quux] != [.* quux pox]
```

I am at a loss as to what might cause this. The obvious change is that now not only patterns but the file contents are kept in the `Matcher` struct, while [contents were loaded from file before](https://github.com/syncthing/syncthing/pull/3996/files#diff-1df9805b2381b6288b220226f82d5dc9R1250). However that should not cause this error, as `SetIgnores` should block until the scan finishes and thus the ignores file is reloaded.

imsodin:
Is there still something left to modify on this or is it good to go?

calmh:
Was the thing fixed were an .stignore file is created in some random place if I post a set of ignores for a folder that doesn't exist? (On mobile, suboptimal code reading env)

imsodin:
@calmh Yes it was: https://github.com/syncthing/syncthing/pull/3996/commits/528457946eb66fdb682688fae7b13a2f81e40d3b

calmh:
Then lgtm, I'll let Audrius decide and merge as he's the first reviewer here

calmh:
That commit is just in the "get" path though, I was thinking of "set" which is not super obvious how it behaves from the code I see

calmh:
(A test for that would be awesome and clear it up, could be among the other gui/rest tests to make sure that an expected error happens on unknown folder :)

AudriusButkevicius:
There is too much logistics in the JS code which could easily be moved to the template and simplify things, such as having separate buttons for different saves depending on the mode.
I am not sure but perhaps even if the dialog is not visible, I think the dom element is still accessible after the modal is gone?

imsodin:
@AudriusButkevicius I am sorry, but I need more explanation - I don't understand what you are talking about. My JS knowledge is limited to what I need to write this.

@calmh Fixed and added tests. Also the weird inconsistently appearing error appeared again - as said earlier I have no clue what this is about as it is not locally reproducible and randomly appearing or not on jenkins.

cmd/syncthing/gui.go
```diff
@@ -969,12 +970,27 @@ func (s *apiService) getRandomString(w http.ResponseWriter, r *http.Request) {
 func (s *apiService) getDBIgnores(w http.ResponseWriter, r *http.Request) {
 	qs := r.URL.Query()
 
-	ignores, patterns, err := s.model.GetIgnores(qs.Get("folder"))
-	if err != nil {
-		http.Error(w, err.Error(), 500)
+	folderName := qs.Get("folder")
+
+	l.Infoln("DBIgnores: folders:", s.cfg.Folders())
```
AudriusButkevicius: This needs to go

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1402,6 +1402,11 @@ angular.module('syncthing.core')
                 autoNormalize: true
             };
             $scope.editingExisting = false;
+            $scope.newIgnores = {
+                edited: false,
+                ignores: "",
+                path: $scope.currentFolder.path
```
AudriusButkevicius: You have this value later, as you create the folder.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1402,6 +1402,11 @@ angular.module('syncthing.core')
                 autoNormalize: true
             };
             $scope.editingExisting = false;
+            $scope.newIgnores = {
+                edited: false,
```
AudriusButkevicius: No need to track this. If content is not '', you can always set the patterns regardless if it was edited or not, don't see any harm. I guess you can skip if it's empty string.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1586,11 +1607,17 @@ angular.module('syncthing.core')
 
         $scope.saveIgnores = function () {
             if (!$scope.editingExisting) {
+                $scope.newIgnores.edited = true;
+                $scope.newIgnores.ignores = $('#editIgnores textarea').val().split('\n')
                 return;
```
AudriusButkevicius: I'd just have a different button with a different callback if not editing existing

gui/default/syncthing/folder/editIgnoresModalView.html
```diff
@@ -14,12 +14,14 @@
     </dl>
   </div>
   <div class="modal-footer">
-    <div class="pull-left"><span translate>Editing</span> <code>{{currentFolder.path}}{{system.pathSeparator}}.stignore</code></div>
```
AudriusButkevicius: Why doesn't this just work? Why do we need new strings?

gui/default/syncthing/core/syncthingController.js
imsodin: I will get rid of this entirely, it is redundant to currentFolder.path.

gui/default/syncthing/core/syncthingController.js
imsodin: The use case I had in mind that required this was to explicitly overwrite a preexisting .stignore by saving an empty ignores modal (versus not using it at all -> keep existing file). However I am probably over-thinking this again - on the other hand does it hurt to track it?

gui/default/syncthing/core/syncthingController.js
imsodin: Just for my education: What is the advantage of adding an addtional function and button for save?
My reasoning was that `saveIgnores` already handles editing an existing function. I simply extended it to not just stop, but actually process the input in case of editing an existing folder.

gui/default/syncthing/folder/editIgnoresModalView.html
imsodin: We are not actually editing the file at path but creating or overwriting it. I consider this important enough to warn the user about it.

The second new string is because when adding a new folder the path is not guaranteed to be set. This results in a message saying you are editing `/.stignore`. This could be gotten rid of by disabling ignores until the configuration is valid (like the "Save" button).

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: As I said elsewhere. We should probably clean up .stignore and .stfolder as we remove folders, in which circumstances this would never exist. Also, wouldn't the getIgnores backpopulate whats on disk if there is anything?

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: I'd say it's separation of concerns. It's one function that does two things based on some magic flag. Also, saveIgnores that actually doesn't save anything in this case (just stores them somewhere) sounds fishy.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1402,6 +1402,11 @@ angular.module('syncthing.core')
                 autoNormalize: true
             };
             $scope.editingExisting = false;
+            $scope.newIgnores = {
+                edited: false,
+                ignores: "",
```
AudriusButkevicius: Can we not access the text area while it's not visible, making this storage redundant?

gui/default/syncthing/folder/editIgnoresModalView.html
AudriusButkevicius: Yeah I think we should disable ignore button until the form is valid. Which reduces the spectrum of messages

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: It wouldn't back populate on adding an new folder as it does not exist.

cmd/syncthing/gui.go
```diff
@@ -969,12 +970,27 @@ func (s *apiService) getRandomString(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		matcher := ignore.New(false)
+		path := filepath.Join(folder.Path(), ".stignore")
+		if err := matcher.Load(path); err != nil {
+			http.Error(w, err.Error(), 500)
+			return
+		}
+		ignores = matcher.Lines()
+		patterns = matcher.Patterns()
+	}
```
AudriusButkevicius: Maybe we should move this to the model (or make this the default implementation in the model), so that there are is a single way to get ignores?

gui/default/syncthing/core/syncthingController.js
imsodin: It can exist even if we cleaned up: I exclusively set ignores by editing the .stignore file directly and I always create it before adding the folder to prevent unwanted files making it into the index.
And setIgnores is called at the end on saving the new folder, then it overwrites the file on disk and only then is getignores called to reply to the REST call with the new config. The difficulty in this PR was always that you can't populate what is on disk as the folder does not yet exist and the path is not fixed yet. Otherwise we wouldn't even be talking about overwriting.

AudriusButkevicius: Well at the point you press save things are set it stone. You could easily add folder paused, set ignores and unpause the folder as part of the addition process

AudriusButkevicius: Also, I feel the end state would either be load stuff for unadded folders and allow modifying (but that's a security hole), or overwriting

imsodin: > Well at the point you press save things are set it stone. You could easily add folder paused, set ignores and unpause the folder as part of the addition process

That is exactly what I do in this PR. It means that at the moment when the user is editing ignores, the existing ignores cannot be loaded. At your request when saving the ignores a preexisting ".stignore" is overwritten, not appended to.

> Also, I feel the end state would either be load stuff for unadded folders and allow modifying (but that's a security hole), or overwriting

I half-heartedly proposed such an extension to the ignores API point but that was not met by acclaim. As you and calmh pointed out and I completely agree, if someone puts a ".stignore" in place he wont just overwrite by accident, so I think just overwriting an existing file as it is currently done is simplest and safest.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1511,6 +1513,11 @@ angular.module('syncthing.core')
             $scope.config.folders = folderList($scope.folders);
 
             $scope.saveConfig();
+
+            if (!$scope.editingExisting && $scope.editedIgnores) {
+                $scope.editingExisting = true;
```
AudriusButkevicius: I'd remove the check condition in the save ignores, and just saveIgnores()

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1584,6 +1591,21 @@ angular.module('syncthing.core')
...
+                return;
+            }
+
+            if ($scope.currentFolder.path.endsWith($scope.system.pathSeparator)) {
+                $scope.currentFolder.path = $scope.currentFolder.path.slice(0, -1);
+            };
+            $('#editIgnores').modal().one('shown.bs.modal', function () {
+                textArea.focus();
+            });
+            $scope.editedIgnores = true;
```
AudriusButkevicius: I'd do it based if textArea content is not empty or something like that.

lib/model/model.go
```diff
@@ -1242,65 +1241,55 @@ func (m *Model) ConnectedTo(deviceID protocol.DeviceID) bool {
...
-		fmt.Fprintln(fd, line)
-	}
-
-	if err := fd.Close(); err != nil {
-		l.Warnln("Saving .stignore:", err)
-		return err
+	if paused {
+		return nil
 	}
-	osutil.HideFile(path)
```
AudriusButkevicius: This can stay.

lib/model/model.go
```diff
@@ -1242,65 +1241,55 @@ func (m *Model) ConnectedTo(deviceID protocol.DeviceID) bool {
...
 }
 
 func (m *Model) SetIgnores(folder string, content []string) error {
+	m.fmut.RLock()
 	cfg, ok := m.folderCfgs[folder]
+	m.fmut.RUnlock()
+	var paused bool
 	if !ok {
-		return fmt.Errorf("Folder %s does not exist", folder)
+		cfg, paused = m.cfg.Folders()[folder]
```
AudriusButkevicius: I guess this is still `ok` or `exists` not actually `paused`. The access can also be folded into the if as it's not used anywhere as in ` if x, y := foo; y == bar { }`

lib/model/model.go
```diff
@@ -1242,65 +1241,55 @@ func (m *Model) ConnectedTo(deviceID protocol.DeviceID) bool {
...
+		if !paused {
+			return fmt.Errorf("Folder %s does not exist", folder)
+		}
 	}
 
-	path := filepath.Join(cfg.Path(), ".stignore")
-
-	fd, err := osutil.CreateAtomic(path)
+	err := ignore.WriteIgnores(filepath.Join(cfg.Path(), ".stignore"), content)
```
AudriusButkevicius: Can fold into the if too.

lib/model/model.go
```diff
@@ -1242,65 +1241,55 @@ func (m *Model) ConnectedTo(deviceID protocol.DeviceID) bool {
...
+	lines := m.folderIgnores[folder].Lines()
 	patterns := m.folderIgnores[folder].Patterns()
 	m.fmut.RUnlock()
 
 	return lines, patterns, nil
 }
 
 func (m *Model) SetIgnores(folder string, content []string) error {
+	m.fmut.RLock()
 	cfg, ok := m.folderCfgs[folder]
```
AudriusButkevicius: Maybe we don't even need this, just check the config?

lib/model/model_test.go
```diff
@@ -947,11 +947,13 @@ func TestIgnores(t *testing.T) {
...
 	db := db.OpenMemory()
 	m := NewModel(defaultConfig, protocol.LocalDeviceID, "device", "syncthing", "dev", db, nil)
-	m.AddFolder(defaultFolderConfig)
-	m.StartFolder("default")
 	m.ServeBackground()
+	m.cfg.SetFolder(defaultFolderConfig)
 	defer m.Stop()
 
+	// wait for goroutine wrapper.notifyListener to finish
+	time.Sleep(time.Duration(100) * time.Millisecond)
```
AudriusButkevicius: 5ms should be enough?

lib/model/model.go
```diff
@@ -1242,65 +1241,55 @@ func (m *Model) ConnectedTo(deviceID protocol.DeviceID) bool {
...
-	if !cfg.HasMarker() {
-		return lines, nil, fmt.Errorf("Folder %s stopped", folder)
-	}
-
-	fd, err := os.Open(filepath.Join(cfg.Path(), ".stignore"))
-	if err != nil {
-		if os.IsNotExist(err) {
-			return lines, nil, nil
+		cfg, ok = m.cfg.Folders()[folder]
+		if ok {
```
AudriusButkevicius: I guess we can just check if it's in the config and that's it?

imsodin: There is a difference between the folder config being in the model or the config: In case it is in the model the ignores can be directly loaded from memory. In case iti is only in the config (i.e. paused), it has to be read from ".stignore".

lib/model/model.go
imsodin: In this case it is definitely necessary, as in the paused case no rescan can happen in the end.

lib/model/model.go
imsodin: It is used at the end of the function and it means paused, i.e. folder exists but is not currently in the model. In case a folder is unpaused, the new ignores are written to ".stignore" and then a rescan is scheduled to pick up the new ignores and the resulting changes to the index. This cannot happen when a folder is paused.

lib/model/model.go
imsodin: Why? All the writing is handled in `ignore.WriteIgnores`.

lib/model/model_test.go
imsodin: Maybe, locally it works. I have just grown vary of timing issues and therefore like to be especially prudent - lets give it a spin on jenkins.

lib/model/model.go
```diff
@@ -1242,65 +1241,54 @@ func (m *Model) ConnectedTo(deviceID protocol.DeviceID) bool {
...
-	scanner := bufio.NewScanner(fd)
-	for scanner.Scan() {
-		lines = append(lines, strings.TrimSpace(scanner.Text()))
+	if !cfg.HasMarker() {
+		return nil, nil, fmt.Errorf("Folder %s stopped", folder)
 	}
 
 	m.fmut.RLock()
+	lines := m.folderIgnores[folder].Lines()
 	patterns := m.folderIgnores[folder].Patterns()
```
AudriusButkevicius: You should store the return value of the map access, and then call the two functions.

lib/model/model.go
AudriusButkevicius: I'd try to access the runner instead, and send the can to the runner, so that we don't get the config from two different places (which is very weird).

Also if a folder overall does not exist, you end up checking two places, the map, and then the config.
Instead I'd  use the runner and move checking if the runner exists to the very end of the function, and make that deciding factor if we scan or not, otherwise I'd just trust the config to do the right thing.

```
cfg, ok := Folders()[folder]
if !ok {
   blah
}

do stuff

if runner, ok := m.runners[folder]; ok {
   return runner.Scan()
}
return nil
```

lib/model/model.go
AudriusButkevicius: I wonder if that is still relevant.

We used to have to compile regular expressions, and comparing them would take a lot of CPU time. Now we use a generic globbing package which is much faster, so I wonder if caching yields much benefits.

But fine, let's leave it like that.

AudriusButkevicius: I guess I'd invert the if condition doing the following:

```
runner, ok  = foo
if ok {
   return runner.Lines()
}

cfg, ok := getfolder
if ok {
  return readFile()
}
return nil
```

This way you don't have to follow to levels of ok's and terminates early which makes it more readable.

lib/model/model.go
AudriusButkevicius: Ah, ok.

lib/ignore/ignore.go
```diff
@@ -360,11 +371,12 @@ func parseIgnoreFile(fd io.Reader, currentFile string, modtimes map[string]time.
 		} else if strings.HasPrefix(line, "#include ") {
 			includeRel := line[len("#include "):]
 			includeFile := filepath.Join(filepath.Dir(currentFile), includeRel)
-			includes, err := loadIgnoreFile(includeFile, modtimes)
+			inclLines, inclPatterns, err := loadIgnoreFile(includeFile, modtimes)
```
AudriusButkevicius: These are short enough that we can call them includeLines, includePatterns

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: This is still relevant.

lib/model/model.go
imsodin: Did it, definitely much cleaner - thanks a lot for your feedback!

imsodin:
@AudriusButkevicius 

I think I discovered an existing problem with paused folders:
A new folder is added and started regardless of whether it is paused: https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L2381
An existing folder is restarted, where it is neither added nor started when paused: https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L379
There is code in model.go that expects paused folders to be present in the model: https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L1772

I think we had this discussion already but I can't find it any more: I still think that it would be clearest to keep paused folders in the model. In that case it would never be necessary to check in two places for the config. And e.g. `m.deviceFolders` would always contain all folders, also the paused onces. Also folder resume events are not emitted when it happens via a configuration change (i.e. via GUI), as `CommitConfiguration` then just sees a new, not a changed folder.
In any case either `CommitConfiguration` or `StartFolder` needs to check if the folder is paused and not add/start it.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1509,13 +1495,18 @@ angular.module('syncthing.core')
                 delete folderCfg.versioning;
             }
 
+            if (!$scope.editingExisting && $('#editIgnores textarea').val()) {
```
AudriusButkevicius: You should store the return value in a variable, as the $() lookup is quite expensive.

AudriusButkevicius: Also, potentially trim the value.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1413,26 +1414,11 @@ angular.module('syncthing.core')
...
-                simpleKeep: 5,
-                staggeredMaxAge: 365,
-                staggeredCleanInterval: 3600,
-                staggeredVersionsPath: "",
-                externalCommand: "",
-                autoNormalize: true,
-                viewFlags: {
-                    importFromOtherDevice: true
-                }
+            $scope.currentFolder = $scope.folderDefaults;
```
AudriusButkevicius: I think this gets a reference, not a copy, so you'd be altering the defaults too.

AudriusButkevicius: There is angular.copy used somewhere else in the code I think.

lib/model/model.go
```diff
@@ -1244,52 +1244,48 @@ func (m *Model) GetIgnores(folder string) ([]string, []string, error) {
...
 
 	if err := ignore.WriteIgnores(filepath.Join(cfg.Path(), ".stignore"), content); err != nil {
 		l.Warnln("Saving .stignore:", err)
 		return err
 	}
 
-	if paused {
-		return nil
+	m.fmut.RLock()
+	_, ok = m.folderCfgs[folder]
```
AudriusButkevicius: You can get the runner, which would have a scan method on it, instead of calling model.Scan() (which does that anyway).

lib/model/model.go
```diff
@@ -1244,52 +1244,48 @@ func (m *Model) GetIgnores(folder string) ([]string, []string, error) {
...
-	patterns := m.folderIgnores[folder].Patterns()
-	m.fmut.RUnlock()
+	if cfg, ok := m.cfg.Folders()[folder]; ok {
+		matcher := ignore.New(false)
+		path := filepath.Join(cfg.Path(), ".stignore")
+		if err := matcher.Load(path); err != nil {
+			return nil, nil, err
+		}
+		return matcher.Lines(), matcher.Patterns(), nil
+	}
```
AudriusButkevicius: I'd also add a comment in both places explaining why this is checked twice.

AudriusButkevicius:
The first one is a good catch, and needs fixing.

The second one, I am not sure I understand the issue?

The third one is a non-issue as you don't scan paused folders anyway, just unpaused folders when setting ignores.

imsodin:
You are right, only the first is an actual issue. I wanted to point out that the current code is not consistent in whether it expects paused folders to be in the model or not and add that to my case for having all folders in the model whether paused or not. What is the benefit of not including paused folders?

AudriusButkevicius:
So I think paused folders are not expected to be in the model, and yeah, it's possible that there are cases that are inconsistent.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1507,10 +1495,20 @@ angular.module('syncthing.core')
                 delete folderCfg.versioning;
             }
 
+            var ignores = $('#editIgnores textarea').val().trim()
```
AudriusButkevicius: missing semicolon

AudriusButkevicius:
@st-review lgtm

st-review:

AudriusButkevicius:
@calmh have a last look

AudriusButkevicius:
Btw, does it work correctly now? If you add a single file to the folder ahead of time, add the folder, does it end up being with 0 items?

imsodin:
> Btw, does it work correctly now? If you add a single file to the folder ahead of time, add the folder, does it end up being with 0 items?

Yes, anything that is ignored at folder creation does not turn up in the Local/Global State.

gui/default/syncthing/folder/editIgnoresModalView.html
```diff
@@ -14,12 +14,13 @@
...
-    <button type="button" class="btn btn-primary btn-sm" ng-click="saveIgnores()" data-dismiss="modal">
+    <div class="pull-left" ng-show="editingExisting"><span translate>Editing</span> <code>{{currentFolder.path}}{{system.pathSeparator}}.stignore</code></div>
+    <div class="pull-left" ng-show="!editingExisting"><span translate>Creating ignore patterns, overwriting an existing file at</span> <code>{{currentFolder.path}}{{system.pathSeparator}}.stignore</code>.</div>
```
calmh: I know this is now a sad never ending story so it hurts a bit to bring this up now, but this string should really not assume that the path is the last word of the sentence and has a period after it. Some languages may require a different sentence structure so the path should be a placeholder.

(The "Editing <foo>" thing is a historical accident.)

lib/ignore/ignore.go
```diff
@@ -274,27 +284,28 @@ func hashPatterns(patterns []Pattern) string {
 	return fmt.Sprintf("%x", h.Sum(nil))
 }
 
-func loadIgnoreFile(file string, modtimes map[string]time.Time) ([]Pattern, error) {
+func loadIgnoreFile(file string, modtimes map[string]time.Time) ([]string, []Pattern, error) {
 	if _, ok := modtimes[file]; ok {
-		return nil, fmt.Errorf("Multiple include of ignore file %q", file)
+		return nil, nil, fmt.Errorf("Multiple include of ignore file %q", file)
```
calmh: Nitpick, errors should not begin with a capital letter (because they'll get logged in the end like `l.Warnln("Couldn't load ignores: %v", err)` and we don't want a spurious mid-sentence capital).

lib/model/model_test.go
```diff
@@ -947,11 +947,13 @@ func TestIgnores(t *testing.T) {
...
 	db := db.OpenMemory()
 	m := NewModel(defaultConfig, protocol.LocalDeviceID, "device", "syncthing", "dev", db, nil)
-	m.AddFolder(defaultFolderConfig)
-	m.StartFolder("default")
 	m.ServeBackground()
+	m.cfg.SetFolder(defaultFolderConfig)
 	defer m.Stop()
 
+	// wait for goroutine wrapper.notifyListener to finish
+	time.Sleep(time.Duration(5) * time.Millisecond)
```
calmh: OK! Now that we know what we're waiting for:

```
type configWaiter struct {
    done chan struct{}
}

func (configWaiter) VerifyConfiguration(from, to Configuration) error { return nil }
func (w *configWaiter) CommitConfiguration(from, to Configuration) (handled bool) {
    close(w.done)
    return true
}
func (configWaiter) String() string { return "configWaiter" }
```

and then instead of the sleep

```go
...
defer m.Stop()

configDone := make(chan struct{})
m.cfg.Subscribe(&configWater{configDone})
m.cfg.SetFolder(defaultFolderConfig)
<-configDone
```

(type directly into browser, probably buggy, but principle should be sound)

calmh: Alternatively, subscribe to and then wait for the `ConfigChanged` event, I guess.

gui/default/syncthing/folder/editIgnoresModalView.html
imsodin: Never mind, rather do it right than do it twice. Two things:

Currently the path is formatted as `<code>` which is not possible within a translated string due to https://github.com/syncthing/syncthing/blob/master/gui/default/syncthing/app.js#L27. I guess this can't/shouldn't be changed for security reasons. So just add the path unformatted?

And lets get rid of this historical accident then. As this already introduces a new string, a second one (well an additional changed one) shouldn't hurt much, right?

lib/model/model_test.go
imsodin: There is no `ConfigChanged` and `ConfigSaved` is not emitted as the config isn't saved to disk. So I adopted your solution - thanks!

gui/default/syncthing/folder/editIgnoresModalView.html
imsodin: I went ahead with it, just complain if it isn't wanted.

lib/model/model_test.go
imsodin: @calmh
Well that actually isn't reliable as the `configWaiter` is triggered shortly after `model.CommitConfiguration` starts, not when it finished -> racy again. So I tried something else: I added a `ConfigChanged` event emitted at the end of `model.CommitConfiguration` and wait for that event. Well it is not enough, the good old weirdo error turns up again https://build.syncthing.net/job/syncthing-pr/3544/console So back to square one: No clue why it happens. Now we are even sure that the folder was added and started. Is there any way to find out what the program does in the 5ms sleep that are apparently enough to fix it?

calmh:
Eh, is odd. I can't look at it right now but I will. Another possibility could be to add an even `IgnoredPatternsChanged` and react to that for example, but it feels a bit so-so to add the event just for test purposes.

calmh:
Is it just starting up and scanning the folder? In which case doing a manual scan from the test is a way to wait for that to happen

imsodin:
@calmh Is https://github.com/syncthing/syncthing/pull/3996/commits/29c7ed836967a7a5812f5e0a2fb50e62f9682f84 what you meant? It doesn't solve the issue either...
Is there any way to control the log level on jenkins for this PR? I have little hope that there is anything helpful in there, but maybe we get lucky.

calmh:
Hmmm yeah. There is now, I gave you some credentials and info on IRC.

imsodin:
After way too many attempts this test is now under control. There was the one problem that was always suspected (correctly) to be the cause (initial scan) and there was a race that was to blame for making it harder than it should be to wait for the initial scan to finish.

The race is between a manually scheduled scan and the initial scan. The initial scan happens after one millisecond. While extremely unlikely, it was apparently possible to squeeze a manual scan into this little interval and this was damaging enough to wreak havoc on these tests. I prevented this by waiting for the initial scan to finish when scheduling a scan, which needed a little refactor on the folders (move `initialScanCompleted` to folder.go).

calmh:
@st-review lgtm @AudriusButkevicius hade some comments and has looked more at it

st-review:

AudriusButkevicius:
@st-review lgtm

st-review:
:ok_hand: Merged as 25b314f5f1448355dc9779ad260dcaa843dcd40c. Thanks, @imsodin!

calmh:
I hesitate to even mention this, but https://build.syncthing.net/job/syncthing-mac/977/console :/

imsodin:
Nice... Apparently mac got the powers to trigger yet another weird problem when initializing a folder.
Honestly I am not really motivated to again search for something that is unlikely to be ever relevant in reality. This is (probably?) "fixable" by the mystery 10ms sleep.

calmh:
No worries, I can squash this one. The problem is that we use the file modtime to determine whether to reload the ignores or not, and the Mac only has one second resolution on the modtime. The test changes ignores multiple times per second.

PR #4018, opened 2017-03-04
Encrypt local discovery

calmh:
### Purpose

This is a quick hack I just thought off that may or may not be a good idea - it's more a "because we can" than anything else. It prevents the information leak that is the current local discovery, where each device broadcasts its own device ID, and in the future possibly more information (i.e., the "discovery gossiping" we've considered).

The scheme is that discovery packets are now tailored for each known peer and encrypted with their device ID. Only configured peers can thus understand our discovery announcements. On incoming announcements we first try to unlock them with our device ID, otherwise try to parse the message as a common unencrypted one.

This does mean that two devices can't "see" each other locally until at least one has added the other. This could be a gotcha, I haven't thought it through fully... Please consider this before deciding if this is good, bad, or pointless.

### Testing

Tested locally, things discover each other.


AudriusButkevicius:
We could do the same with normal discovery too I guess, posting a device specific record. I don't see much downside to this, nor upsides. In the public disco case you'd still have to verify client certs to prevent DoS, beating the purpose.

calmh:
Maybe later

Issue #3679, opened 2016-10-21
"Failed Items" dialogue doesn't allow you to select/copy the path of the failed item


sexybern:
When presented with "Failed Items", I want to be able to copy the path of the failed item to the clipboard so that I can paste it into a shell to investigate.

Currently, the "Failed Items" dialogue only presents the path to the user as a hovering tooltip, which doesn't allow you to select or copy the path:

![image](https://cloud.githubusercontent.com/assets/3932675/19589543/934c44c0-9764-11e6-8d9b-06f0b5133051.png)

Could you please add a feature that allows path copying, or a more persistent display of the path so that it can be copied manually?
### Version Information

Syncthing Version: Currently on v0.14.9, Linux (64 bit), but it's always behaved this way.

nov1n:
Hellow, I'm new around here.

How about clicking to copy to clipboard using either
- [clippy](https://github.com/mojombo/clippy)
- [document.execCommand('copy')](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand)

I can work on this.

AudriusButkevicius:
Sounds good, but also need some indication that it gets copied.

nov1n:
Proof of concept implementation in #4021 . Let me know what you think.

PR #4021, opened 2017-03-05
Copy failed file path

nov1n:
### Purpose

Addresses #3679.

### Testing

Tested on Chrome and Firefox on Windows and Ubuntu.

### Implementation

Files that failed to sync may be clicked to copy the full path to clipboard.
[Clipboard.js](https://clipboardjs.com/) was used.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -689,6 +689,7 @@
   <script src="vendor/angular/angular-translate-loader-static-files.js"></script>
   <script src="vendor/angular/angular-dirPagination.js"></script>
   <script src="vendor/bootstrap/js/bootstrap.js"></script>
+  <script src="vendor/clipboard/clipboard.js"></script>
   <!-- / vendor scripts -->
 
   <!-- gui application code -->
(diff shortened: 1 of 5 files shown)


AudriusButkevicius:
There is no indication that it's copyable nor that it was copied.

nov1n:
There is none indeed, that sounds like a good idea. I'll update it tomorrow.

calmh:
More generally, I'm not convinced this is a good idea. Why special case copy behavior for one place a path is shown, and not also do it in the download progress dialog, with the device ID in the device ID dialog, etc?

nov1n:
Issue #3679 raised the issue that there is no way to copy the full file path in the failed files modal. If this is a problem in other places as well, should we add a way to copy there too?

calmh:
Maybe the lack of ability to copy is limited to this case, but adding click-to-copy-automatically seems a bit arbitrary (and hard to discover). Perhaps the real problem is just that the dialog is badly designed to begin with so that the path needs to be hidden in a tooltip, and it should be displayed easier so people can select and Ctrl-C it like anything else?

nov1n:
Okay, I think it was just for convenience purposes as it is a common usecase to copy this path to do investigation. A right aligned icon would make it more easily discoverable, although I agree that it is a arbitrary, especially because it is the only place in the UI where this is possible.

Shall I modify the PR such that the full path is displayed in the dialog (rather than the basepath) such that users may simply Ctrl-C it?

nov1n:
@calmh , see #4023

PR #4025, opened 2017-03-07
lib/events, cmd/syncthing: Correct GlobalID in debug and make since arameter optional

imsodin:
These are minor fixes as a side product of writing documentation.
The since parameter is not actually required for /rest/events and this is already used in the gui code.
The debug log is simply offset by one compared to /rest/events due to first logging then incrementing GlobalID,


calmh:
@st-review merge it

lib/events, cmd/syncthing: Correct GlobalID in debug and mark "since" parameter optional

Skip-check: metalint

st-review:
:ok_hand: Merged as 1c29a93013b6e3109f76cf7717d5dab0be8d98d9. Thanks, @imsodin!

PR #4028, opened 2017-03-07
lib/scanner: Stopped outputting rescan debug message if file doesn't …

### LINKED ISSUE #1350: Debug output while scanning needs improvement
If we fail to have a file locally we output:

```
DEBUG: rescan: File{Name:"", Flags:00, Modified:0, Version:0, Size:0, Blocks:[]} 1416172353 -rw-rw-rw-
```

Which is useless.

We should check the ok condition:
https://github.com/syncthing/syncthing/blob/master/internal/scanner/walk.go#L268

kwhite17:
…exist locally (fixes #1350)

### Purpose

To stop logging 'rescan' debug messages on files that don't exist locally.

### Testing

I ran the scanner test suite to make sure every still functioned as normal.

### DIFF
--- a/lib/scanner/walk.go
+++ b/lib/scanner/walk.go
@@ -311,7 +311,9 @@ func (w *walker) walkRegular(relPath string, info os.FileInfo, fchan chan protoc
 		return nil
 	}
 
-	l.Debugln("rescan:", cf, info.ModTime().Unix(), info.Mode()&os.ModePerm)
+	if ok {
+		l.Debugln("rescan:", cf, info.ModTime().Unix(), info.Mode()&os.ModePerm)
+	}
 
 	f := protocol.FileInfo{
 		Name:          relPath,


kwhite17:
I took a look at the metalint console output and it all seems to be unrelated to the change I made. Is that okay?

calmh:
Yeah the metalint error is something else that just popped u, ignore it. What's the point of this change though, I don't really get it?

kwhite17:
I'm guessing Audrius thought this would be a good way to reduce some noise in the log. According to #1350, syncthing was producing unhelpful debug output while scanning for changed files.  Syncthing will now only log the rescan debug statements for files that exist locally, eliminating the log statements that contain default or empty file information because the file doesn't exist locally.

AudriusButkevicius:

AudriusButkevicius:
@st-review merge this

st-review:

st-review:
:ok_hand: Merged as 1caa683ec1e4634bad5f8169d3b477c7df513d01. Thanks, @kwhite17!

PR #4029, opened 2017-03-07
lib/osutil: Ugly hack to avoid staticcheck complaints

calmh:
### Purpose

Less red, more green


calmh:

calmh:
I have no idea where that error comes from given this code. I even wiped and restarted the containers to ensure there's nothing cached or whatever...

calmh:
Seems buggy, I disabled the check.

PR #4032, opened 2017-03-08
Use SMUX

AudriusButkevicius:
Bench again vs master and let me know what the results are. Potentially try tweaking the KCP settings and see if it makes any difference.


calmh:
Can you fix the formatting issue?

calmh:
The new code is faster. I did three tests each of transferring a 2G file from scratch between two computers on my LAN, on both current `master` (syncthing-linux-amd64-v0.14.24+22-gf35e1ac) and this PR (syncthing-linux-amd64-v0.14.24+27-g2a97825). I graphed transfer rate and CPU load on the receiving side. It's quite noisy so I added an average as well. There are some pauses in the data visible as sawtooth patterns - I think this is the whole program being I/O blocked or something, although I don't fully understand how that could block the perfstat loop from printing it's stuff.

On master the transfer completed in about 340 seconds or so, with the PR in about 240.

https://kastelo.net/t/master-1.pdf
https://kastelo.net/t/master-2.pdf
https://kastelo.net/t/master-3.pdf

https://kastelo.net/t/pr-1.pdf
https://kastelo.net/t/pr-2.pdf
https://kastelo.net/t/pr-3.pdf

AudriusButkevicius:
Did you try tweaking the kcp settings in some way to see if you can squeeze more out?

AudriusButkevicius:
I'll address formatting in the evening.

calmh:
No tweaks so far, except that I've changed the update interval from 100 to 25 ms as a default. I'll take care of the formatting and merge, I want to release an RC and have this is in it

PR #4034, opened 2017-03-09
script, gui: Silence useless warnings in translation script

imsodin:
When I ran `go run build.go translate` I was irritated at first about all the output and then just ignored it and looked at the actual effect it has.
This adds rules to silence untranslated strings warnings related to strings, that are not to be translated, such that in the future relevant warnings about untranslated strings don't go unnoticed.

I decided to ignore the "errors" and "warn" strings in the yellow top bar in case of javascript errors instead of translate them, as they are part of a kind of warning symbol, not really a textual information.

### DIFF
--- a/gui/default/syncthing/core/aboutModalView.html
+++ b/gui/default/syncthing/core/aboutModalView.html
@@ -18,7 +18,7 @@ Jakob Borg, Audrius Butkevicius, Alexander Graf, Anderson Mesquita, Antony Male,
     <hr/>
 
     <p translate>Syncthing includes the following software or portions thereof:</p>
-    <ul class="list-unstyled two-columns">
+    <ul class="list-unstyled two-columns" id="copyright-notices">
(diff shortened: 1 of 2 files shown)


calmh:
Lgtm. Possibly we could simplify by using an attribute in the HTML, like `data-translatable="no"` on things we know should not be translated?

calmh:
@st-review lgtm

st-review:

imsodin:
This sounds like a much cleaner approach. I quickly googled the data-... parameter option and I found that this is only valid in html5, but invalidates earlier html, but actually works anyhow. Do we have to care about such "formalities"? Otherwise I would probably want to change this PR to that method.

AudriusButkevicius:
@st-review lgtm

st-review:

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 88dfd634e5a24f15bf2622beb797f6aa81675a22. Thanks, @imsodin!

PR #4042, opened 2017-03-15
lib/model, lib/scanner: Properly ignore symlinks on Windows (fixes #4035)

### LINKED ISSUE #4035: Syncing fails with 'A required privilege is not held by the client' on Windows with symlinks
After upgrading to Syncthing v0.14.24 I'm no longer able to finish syncing between two Windows machines (it won't ignore symbolic links anymore which it isn't able to sync.)

Machine A has a 'send only' folder that is synchronized with Machine B. Some files on machine A are symbolic links. Previously I'd turned off symlinks under the advanced configuration options so the symbolic links would be ignored. The option to ignore symbolic links has been removed in v0.14.24 and now it won't ignore them and it can't synchronize them either.

### Version Information

Machine A:
Syncthing Version: v0.14.24
OS Version: Windows 10

Machine B:
Syncthing Version: v0.14.24
OS Version: Windows 10

Probably unrelated but I should note I use tray icon Synctrayzor (at latest, v1.1.14) on both machines which starts Syncthing.

calmh:
Adds a unit test to ensure we don't scan symlinks on Windows. For the
rwfolder, trusts that the logic in the invalid check is correct and that
the check is actually called from the need loop.

(Need to make the build server able to run the symlink test, but it works on my Windows 10 as admin)

### DIFF
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -53,8 +53,9 @@ type copyBlocksState struct {
 const retainBits = os.ModeSetgid | os.ModeSetuid | os.ModeSticky
 
 var (
-	activity    = newDeviceActivity()
-	errNoDevice = errors.New("peers who had this file went away, or the file has changed while syncing. will retry later")
+	activity               = newDeviceActivity()
+	errNoDevice            = errors.New("peers who had this file went away, or the file has changed while syncing. will retry later")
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 5c27796471d2da8e9dd890b476361941c4447e5b. Thanks, @calmh!

PR #4059, opened 2017-03-23
gui, lib/events, lib/model: Introduce device state events based on activity (fixes #623)

### LINKED ISSUE #623: "Syncing" status should be based on activity, not completion percentage
To handle the case where a node ignores a bunch of files, we should not assume a node is "syncing" just because it doesn't have all files we think it should have. It should be based on something else, like whether we've seen a file request from that node in the last x seconds or similar.

calmh:
This introduces a new event DeviceState and some new new GUI mechanics
to make use of it. Essentially we keep track of whether a device is
syncing or idle, based on whether we've seen a Request from them lately.
(They can of course be syncing stuff from someone else without ever
sending us any requests, but it's fairly unlikely if it's on a folder
we're sharing.)

In addition I add the states "preparing index" and "sending index". The
point of this is to show what's going on when we connect to a new device
and have a large index to send. For example, on my photo catalog, when I
add a new device it takes about 5 minutes to prepare and sort the index,
during which time (in the old GUI) exactly nothing appears to happen.
Now it correctly shows "preparing index" during that time, and then
"sending index" while it's being transmitted, before presumably going to
"syncing" when data is flowing.

It changes the problematic "Syncing (100%)" to "Idle (100%)" for the
cases where the other device isn't downloading stuff yet still isn't
really in sync from our point of view. It's still not optimal, but it's
a small improvement in that case as well.


AudriusButkevicius:
This is confusing.

Half of the events are for local state and half are for remote.

Ideally we'd start annoying devices own state over the wire, and remove the guess work, at which point I am not sure if it's relevant to signify that we are doing an index sorting locally.

calmh:
It's relevant because it can take a shitload of time, during which everything just looks idle or stalled otherwise. But we could certainly send activity updates. We could in that case also send our locally perceived completion percentages, which the other side could show, avoiding some issues around that... Lets leave this on ice for a moment and sketch out how that could look, then see.

AudriusButkevicius:
This idea or this pr?

calmh:
Leave this PR on ice, and lets sketch out how activity announcements could look

AudriusButkevicius:
Where do you want to discuss it?
I'd imagine we'd could extend the events mechanism over the protocol connection, annoucing folder state changes and what not. I guess the hard bit would be reconciling stuff, as in X started syncting, then disconnects etc etc.

calmh:
https://forum.syncthing.net/t/activity-update-messages/9567

PR #4070, opened 2017-03-30
lib/scanner: Use fs.Filesystem for all operations

calmh:
### Purpose

One more step on the path of the great refactoring. Touches rwfolder a little bit since it uses the Lstat from fs as well, but mostly this is just on the scanner as rwfolder is scheduled for a later refactor.

There are a couple of usages of fs.DefaultFilesystem that will in the end become a filesystem injected from the top, but that comes later.

### Testing

Much of this is still unit tested, and I ran a couple of manual integration tests on top.


lib/fs/basicfs.go
```diff
@@ -71,11 +71,32 @@ func (f *BasicFilesystem) DirNames(name string) ([]string, error) {
...
+}
+
+// fsFile implements the fs.File interface on top of an os.File
+type fsFile struct {
+	*os.File
+}
+
+func (f fsFile) Stat() (FileInfo, error) {
+	info, err := f.File.Stat()
+	if err != nil {
```
AudriusButkevicius: Doesn't this need the magic stat treatment?

AudriusButkevicius: I guess you could pass a handle to the underlying fs, and it could do f.fs.Stat()

lib/fs/mtimefs_test.go
```diff
@@ -25,7 +25,7 @@ func TestMtimeFS(t *testing.T) {
 	// a random time with nanosecond precision
 	testTime := time.Unix(1234567890, 123456789)
 
-	mtimefs := NewMtimeFS(make(mapStore))
+	mtimefs := NewMtimeFS(make(mapStore), DefaultFilesystem)
```
AudriusButkevicius: Can we swap the args around given mtimefs needs a backing fs.

lib/model/rwfolder.go
```diff
@@ -678,7 +678,7 @@ func (f *sendReceiveFolder) handleDir(file protocol.FileInfo) {
 	// It's OK to change mode bits on stuff within non-writable directories.
 	if f.ignorePermissions(file) {
 		f.dbUpdates <- dbUpdateJob{file, dbUpdateHandleDir}
-	} else if err := os.Chmod(realName, mode|(info.Mode()&retainBits)); err == nil {
+	} else if err := os.Chmod(realName, mode|(os.FileMode(info.Mode())&retainBits)); err == nil {
 		f.dbUpdates <- dbUpdateJob{file, dbUpdateHandleDir}
```
AudriusButkevicius: On the phone so ignore my ignorance. This is just an int, so why do we need to change the type?

lib/fs/basicfs.go
calmh: I don't think so. We haven't treated these specifically before, just done Stat() on the os-returned fd.

That said, I was considering removing the android-specialcasing anyway to see if it's still broken or not... But opted to bring it along in the end.

lib/model/rwfolder.go
calmh: Chmod wants an os.FileMode, and we return an fs.FileMode. The others are just numeric constants, so in the end the expression becomes fs.FileMode unless converted.

lib/scanner/walk.go
```diff
@@ -80,18 +80,14 @@ type CurrentFiler interface {
...
 func Walk(cfg Config) (chan protocol.FileInfo, error) {
 	w := walker{cfg}
 
 	if w.CurrentFiler == nil {
 		w.CurrentFiler = noCurrentFiler{}
 	}
-	if w.Lstater == nil {
-		w.Lstater = defaultLstater{}
+	if w.Filesystem == nil {
+		w.Filesystem = fs.DefaultFilesystem
```
calmh: Not by the tests. It seems sane to have a fallback, but in the end I don't feel strongly about it either way.

lib/fs/basicfs.go
AudriusButkevicius: We should definately keep it, unless stdlib addresses it.

lib/model/rwfolder.go
AudriusButkevicius: I guess this will become a fs.Chmod at which point these will go away. Fine.

lib/scanner/walk.go
AudriusButkevicius: Ok let's keep it.

lib/fs/basicfs.go
calmh: Yep. It is kept. The code above is equal to what we had before.

lib/scanner/blockqueue.go
```diff
@@ -42,7 +42,7 @@ func newParallelHasher(dir string, blockSize, workers int, outbox, inbox chan pr
 
 // HashFile hashes the files and returns a list of blocks representing the file.
 func HashFile(path string, blockSize int, counter Counter, useWeakHashes bool) ([]protocol.BlockInfo, error) {
-	fd, err := os.Open(path)
+	fd, err := fs.DefaultFilesystem.Open(path)
```
AudriusButkevicius: Hmm, this should probably bubble down all the way from the scanner if you want to declare that "scanner conversion is finished".

calmh: Mmmyeah that's a bit lazy, let me refactor that walker thing.

AudriusButkevicius:
@st-review lgtm

st-review:

lib/fs/filesystem.go
```diff
@@ -57,12 +59,20 @@ type FileInfo interface {
 // FileMode is similar to os.FileMode
 type FileMode uint32
 
+// ModePerm is the equivalent of os.ModePerm
+const ModePerm = FileMode(os.ModePerm)
+
 // DefaultFilesystem is the fallback to use when nothing explicitly has
 // been passed.
 var DefaultFilesystem Filesystem = new(BasicFilesystem)
```
imsodin: Not directly related to this PR, but why isn't this ` = NewBasicFilesystem()`?

calmh: Probably I introduced that method later and forgot. Good catch, fixed!

imsodin:
@st-review lgtm

st-review:
:ok_hand: Merged as 4253f22680dfc700759529c553554a106fa16a6a. Thanks, @calmh!

PR #4073, opened 2017-03-30
lib/connections, lib/model, gui: Specify allowed networks per device (fixes #219)

### LINKED ISSUE #219: Allowed subnet list per node
Is it possible to introduce an option to specify which repositories should be synchronized using only LAN connection?

I have some huge data don't need to be synchronized immediately and it could be synchronized within a higher speed connection.

calmh:
### Purpose

This adds a new config AllowedNetworks per device, which when set should contain a list of network prefixes (192.168.0.0/126 etc) that are allowed for the given device. The connection service will not attempt connections to addresses outside of the given networks and incoming connections will be rejected as well.

I've added the config to the normal device editor and shown it (when set) in the device summary on the main screen.

### Testing

There's a unit test for the IsAllowedNetwork method, I've done some manual sanity testing on top of that.

### Screenshots

```
[MRIW7] INFO: Connection from I6KAH76-66SLLLB-5PFXSOA-UFJCDZC-YAOMLEK-CP2GB32-BV5RQST-3PSROAU
 at 127.0.0.1:61159 (tcp-server) rejected: network not allowed
```

screen shot 2017-03-30 at 18 28 40

screen shot 2017-03-30 at 18 28 32


imsodin:
Maybe this option is something for the advanced configuration to keep the already biggish "Edit Device" modal simple(r). Other than that, lgtm.

AudriusButkevicius:
Agree it should be in advanced or we should add an expansion thingy.

calmh:
TBH this is why I fixed slices for advanced config, but then I thought "why be lazy" and added it to the real config settings. But if you guys think this is too niche for that, I'm fine with dropping it and letting it be advanced only for now. The downside to that is that there is no good place to document the format then, other than in the documentation (which I can do...).

lib/connections/service.go
```diff
@@ -641,3 +646,30 @@ func tlsTimedHandshake(tc *tls.Conn) error {
 	defer tc.SetDeadline(time.Time{})
 	return tc.Handshake()
 }
+
+func IsAllowedNetwork(host string, allowed []string) bool {
+	if len(allowed) == 0 {
```
AudriusButkevicius: Perhaps this should be on the deviceCfg struct, as the special case of empty list makes it somewhat a shit generic function.

calmh: Maybe... Or I could move the len()==0 check out to the call site instead and make this a strict match only function?

AudriusButkevicius: Whichever works for me.

calmh:
Related: https://github.com/syncthing/docs/pull/312

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as c5e0c479893c4c988ce83ad3cd07361a227a0dc3. Thanks, @calmh!

PR #4074, opened 2017-03-30
gui: Allow toggleable units for transfer rate (fixes #234)

### LINKED ISSUE #234: Make units for transfer rates configurable
Change the Down-/Upload rate from "bps" to "KB/s" or "MB/s" because it's more common, isn't it?

EDIT: Also "Outgoing Rate Limit (KBps)" and "Max File Change Rate (KBps)"

calmh:
Click the transfer rate to toggle between binary-exponent bytes (KiB/s, MiB/s) and metric based bits (kbps, Mbps).

No, this is not persisted anywhere so someone with a strong preference for the "bits" case will still be sad. But it's literally a ten minute fix of a contentious issue, so let the next PR handle that.

rates

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -221,6 +221,13 @@ identicon {
     height: 1em;
 }
 
+a.toggler {
+    color: inherit;
+}
+a.toggler:hover {
+    border-bottom: 1px dashed;
+    text-decoration: none;
+}
 
 /**
  * Progress bars with centered text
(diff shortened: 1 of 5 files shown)


AudriusButkevicius:
Use local storage for persistence. Most sane browsers support it. I think this is where we save the language selection.

calmh:
Local storage'd!

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as c3820fbbf237091d988827907dd5e74d0f2681fc. Thanks, @calmh!

PR #3986, opened 2017-02-13
Filesystem Notification - Continuation

imsodin:
This is the continuation of plouj's PR #2807 on a branch in the syncthing repo for ease of maintenance. Refer to that PR for general information what this is about and commits previous to https://github.com/syncthing/syncthing/commit/1d424d84609a40d2a11ac8e06cb87aac6b130629.

PR #4083, opened 2017-04-06
lib/model: TestIgnores: Add ms sleep on all platforms (ref #3986 #3996)

imsodin:
Apparently no sleep at all between changes to ".stignore" is potentially too fast even for non-darwin systems.
https://build.syncthing.net/job/syncthing-pr/3605/console
Unfortunately it isn't reproducible as with the mac builds, but hopefully a tiny 1ms sleep will fix it.
This PR exists primarily to get rid of unrelated test failures in #3986.


calmh:
@st-review merge it (voodoo, but doesn't hurt)

st-review:
:ok_hand: Merged as 5521759b23b05ecf5c3cb73ee633370bb8de06a0. Thanks, @imsodin!

PR #4087, opened 2017-04-06
gui, lib/config, lib/model: Allow absolute values for minimum disk free space (fixes #3307)

### LINKED ISSUE #3307: Set "Minimum Free Disk Space" in MB/GB
I set "Minimum Free Disk Space" to `0,001` but still get `disk has insufficient free space`. My HDD has over 5 GB free and the Folder i want to sync has 2,4 MB of files, changes are one line in a text file.

I think the setting would make more sense to set to "1 GB" or "500 MB".

calmh:
### Purpose

This deprecates the current minDiskFreePct setting and introduces minDiskFree. The latter is, in it's serialized form, a string with a unit. We accept percentages ("2.35%") and absolute values ("250 k", "12.5 Gi"). Common suffixes are understood. The config editor lets the user enter the string, and validates it.

We still default to "1 %", but the user can change that to an absolute value at will.

### Testing

Unit tests on the parsing. Manual testing that the checks themselves still work.

### Screenshots

screen shot 2017-04-06 at 14 02 32


calmh:
@Nutomic @Zillode FYI on the config format change.

@canton7 @kozec Don't think you guys parse or post the config?

calmh:
Github, why didn't you link this to #3307 already?

canton7:
> Don't think you guys parse or post the config?

I parse bits of it, but this bit doesn't affect me.

AudriusButkevicius:
I think the size bits should live in a separate package. Also I wonder if there is already something in Go that deals with this (same like quantities package for python, with pretty printing, rescaling, parsing)

calmh:
Not that I'm aware of, but I'm open for suggestions? :)

AudriusButkevicius:
Can't find anything useful.let's just move it to a separate package so we can build upon it.

calmh:
I'd say lets do that when we need it for something. As it is there is only a single user and it fits in like other config-specific things (FolderType and so on). It's also a fairly ugly and special kind if thing that can be either a percentage or a byte count... So yeah. No. :)

Nutomic:
Thanks, will keep this in mind for the next version.

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -73,11 +73,11 @@
...
               </p>
             </div>
-            <div class="form-group" ng-class="{'has-error': folderEditor.minDiskFreePct.$invalid && folderEditor.minDiskFreePct.$dirty}">
-              <label for="minDiskFreePct"><span translate>Minimum Free Disk Space</span> (0.0 - 100.0%)</label>
-              <input name="minDiskFreePct" id="minDiskFreePct" class="form-control" type="number" ng-model="currentFolder.minDiskFreePct" required min="0.0" max="100.0">
+            <div class="form-group" ng-class="{'has-error': folderEditor.minDiskFree.$invalid && folderEditor.minDiskFree.$dirty}">
+              <label for="minDiskFree"><span translate>Minimum Free Disk Space</span></label>
+              <input name="minDiskFree" id="minDiskFree" class="form-control" type="text" ng-model="currentFolder.minDiskFree" required ng-pattern="/^\s*[0-9.]+\s*(k|M|G|T|Ki|Mi|Gi|Ti|%|)\s*$/">
               <p class="help-block">
-                <span translate ng-if="!folderEditor.minDiskFreePct.$valid && folderEditor.minDiskFreePct.$dirty">The minimum free disk space percentage must be a non-negative number between 0 and 100 (inclusive).</span>
```
imsodin: This string could probably be removed from `lang-en.json` (can't comment there).

AudriusButkevicius:
The only weird thing is preventing the units to be suffixed with B, as in KB or KiB. I've never bought 2 K of chicken... Nor my internet speed was ever 20M/s

AudriusButkevicius:
I think we should stick with what people are used to.

calmh:
That's fair, and also I think the settings UI would be better as one number field and a separate unit drop down; less need for explanatory text, less room for user error.

calmh:
Here's an updated variant, with a more user friendly interface. In the config it's separated, json:

```json
"minDiskFree": {"value": 12.34, "unit": "%"}
```

and in xml:

```xml
<minDiskFree unit="%">12.34</minDiskFree>
```

Screenshot:

screen shot 2017-04-07 at 09 06 35

String parsing retained just for the ParseDefault... :/

calmh:
Pinging reviewers

AudriusButkevicius:
I don't understand why units are a space string in some cases?

gui/default/syncthing/core/syncthingController.js
```diff
@@ -62,7 +62,7 @@ angular.module('syncthing.core')
                 selectedDevices: {},
                 type: "readwrite",
                 rescanIntervalS: 60,
-                minDiskFreePct: 1,
+                minDiskFree: "1 %",
```
imsodin: In the folder edit modal this is represented as a struct with value and unit members, might be the culprit for the space units Audrius mentioned.

calmh:
I'm not sure what you mean, sorry? Do you mean in the parsing tests?

AudriusButkevicius:
Either my phone had something cached or failed to draw the % sign.

AudriusButkevicius:
@st-review lgtm

st-review:

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: Yes, this seems wrong.

calmh: Oh. Yeah. That's wrong. Let me fix.

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -65,19 +65,28 @@
...
+              <select class="form-control" ng-model="currentFolder.minDiskFree.unit">
+                <option value="%">%</option>
+                <option value="kB">kB</option>
+                <option value="MB">MB</option>
+                <option value="GB">GB</option>
+                <option value="TB">TB</option>
+              </select>
               <p class="help-block">
-                <span translate ng-if="!folderEditor.minDiskFreePct.$valid && folderEditor.minDiskFreePct.$dirty">The minimum free disk space percentage must be a non-negative number between 0 and 100 (inclusive).</span>
+                <span translate>Enter a non-negative number (e.g., "2.35") and select a unit. Percentages are as part of the total disk size.</span>
```
imsodin: In my opinion this should either also only show when the user input is invalid or the info test on the rescan interval should be always visible as well.

imsodin:
And full nit-pick mode (ignore at will): The input fields are generally as wide as possible, while this new field isn't. Not that I particularly like the wide input fields, just for consistency.

calmh:
Ok, further layout tweaks. Full width, to the extent that I can accomplish with bootstrap. I'd have liked less margin between the controls, but at this point I give up. No help text when valid, to the price of it not being absolutely obvious that fractions are allowed:

screen shot 2017-04-12 at 08 14 11

With error:

screen shot 2017-04-12 at 08 14 18

On tiny screens:

screen shot 2017-04-12 at 08 14 42

calmh:
Oh, and we need a corresponding editor for the home disk threshold visible in normal settings as well, because the advanced config editor doesn't understand the object type that now represents the value-and-unit...

calmh:
In main config editor, bottom left:

screen shot 2017-04-12 at 08 24 27

With error:

screen shot 2017-04-12 at 08 24 41

imsodin:
Looks good!
Just one more question: Are old, no longer used strings intentionally kept?
https://github.com/calmh/syncthing/blob/c1e08b8662ad095050c13df19b3f385b7c68f13f/gui/default/assets/lang/lang-en.json#L233

calmh:
They are intentionally not removed because it's simpler and I don't think it's required in practice.

imsodin:
@st-release lgtm

calmh:
("st-review" :)

calmh:
@imsodin I don't think the bot picks up on edits

imsodin:
@st-review lgtm

st-review:
:ok_hand: Merged as da34f27546f27e5048df1175d4de1fafce6d687c. Thanks, @calmh!

PR #4091, opened 2017-04-11
cmd/syncthing: Handle XML format configs over the API

calmh:
### Purpose

This simplifies backing up and restoring config remotely. By setting "Accept: application/xml" when GETing the config we get the config in the same format as on disk. Conversely, we can clone a device config to a remote device by POSTing an existing config from disk using "Content-Type: application/xml".

When accepting an XML format config we take it as-is: no funny business about the password etc, it should already be in hashed format.

### Testing

Works for me

### Documentation

Needs adding to the API docs

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -7,8 +7,10 @@
 package main
 
 import (
+	"bytes"
 	"crypto/tls"
 	"encoding/json"
+	"encoding/xml"
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
So I'd be against this. Why not accept other calls in xml then?

calmh:
Yes, why not indeed? In practice, because I needed to this to do a thing, and I don't need it for other stuff. But then again the config is the only thing we have in XML so it doesn't seem unreasonable to special case it.

AudriusButkevicius:
However bad this is going to sound, the fact that you needed to do something is not really a compelling enough argument to wedge in some unexplainable unicorn in the API, especially when the semantics between the two calls that look the same are actually different.

I am sure you are clever enough to convert things to json and not require some sort of cancer in side the code that returns content in xml and errors in json? This is like an invitation to make a joke of the API. If we want to add XML support, we should do it throughout the code base, or nowhere at all. I'd even be in favour convert the config to json to keep it consistent.

calmh:
:D Mmm OK let me think on it. I don't particularly want to XMLize the entire API, nor honestly change the config format - especially not to JSON.

PR #4092, opened 2017-04-11
cmd/syncthing: Allow custom event subscriptions (fixes #1879)

### LINKED ISSUE #1879: Feature: Filter events by type
It would be nice to have a feature to filter the events, e.g. by type using the rest api.

`/rest/events?since=<lastSeenID>&type=DeviceConnected`

or 

`/rest/events?since=<lastSeenID>&type=DeviceConnected&type=DeviceDisconnected`

calmh:
### Purpose

This adds a parameter "events" to the /rest/events endpoint. It should
be a comma separated list of the events the consumer is interested in.
When not given it defaults to the current set of events, so it's
backwards compatible.

The API service then manages subscriptions, creating them as required
for each requested event mask. Old subscriptions are not "garbage
collected" - it's assumed that in normal usage the set of event
subscriptions will be small enough. Possibly lower than before, as we
will not set up the disk event subscription unless it's actually used.

### Testing

The event mask calculation has a couple of tests.

### Documentation

Needs to be added to the api docs.


AudriusButkevicius:
This looks cool, yet I think we should pre-create the disk and other even subs not to miss startup and other important events? The disk sub will just end up empty showing that no files changed after startup?

I assume this does not introduce gaps in the events index?

cmd/syncthing/main.go
```diff
@@ -634,12 +634,6 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 	errors := logger.NewRecorder(l, logger.LevelWarn, maxSystemErrors, 0)
 	systemLog := logger.NewRecorder(l, logger.LevelDebug, maxSystemLog, initialSystemLog)
 
-	// Event subscription for the API; must start early to catch the early
-	// events. The LocalChangeDetected event might overwhelm the event
-	// receiver in some situations so we will not subscribe to it here.
-	apiSub := events.NewBufferedSubscription(events.Default.Subscribe(events.AllEvents&^events.LocalChangeDetected&^events.RemoteChangeDetected), 1000)
-	diskSub := events.NewBufferedSubscription(events.Default.Subscribe(events.LocalChangeDetected|events.RemoteChangeDetected), 1000)
```
imsodin: Unless the reasoning in the above comment is no longer valid, wouldn't it make sense to keep these two subscriptions and add them to the new `eventSubs` map as default subscriptions? I suspect most API users will keep using the default call.
Also in general this approach differs significantly from the old behaviour, by not providing past events on the first call with a specific event type. On a second call, it will, which I think is confusing from a user perspective.

calmh:
You're both right and we should get those subscriptions up and running early. I'll fix that. That said, there was never a guarantee, as depending on what happens the early messages might already have "scrolled away" by the time you come to poll the subscription.

There shouldn't be an impact on the index numbers. The "since" and so on things are per-subscription and have been for a while, and the events themselves have two IDs, the per-subscription one to be able to detect dropped events within the subscription and the global ones that are somewhat less interesting. These already didn't match up with each other since we introduced the disk events that don't end up in the standard subscription.

cmd/syncthing/main.go
canton7: :+1: at the very least I think we need to keep the buffering of past events for the default subscription

cmd/syncthing/gui.go
```diff
@@ -1047,6 +1056,31 @@ func (s *apiService) getEvents(w http.ResponseWriter, r *http.Request, eventSub
...
+	return eventMask
+}
+
+func (s *apiService) getEventSub(mask events.EventType) events.BufferedSubscription {
+	s.eventSubsMut.Lock()
+	bufsub, ok := s.eventSubs[mask]
+	if !ok {
+		evsub := events.Default.Subscribe(mask)
+		bufsub = events.NewBufferedSubscription(evsub, eventSubBufferSize)
+		s.eventSubs[mask] = bufsub
```
canton7: I don't see an expiration policy. I'm assuming you've decided that's not an issue?

AudriusButkevicius:
As long as we expose the per sub event numbers, this looks fine.

calmh:
On 11 Apr 2017, at 11:03, Antony Male <notifications@github.com> wrote:
> 
> I don't see an expiration policy. I'm assuming you've decided that's not an issue?

Yeah, as noted in the commit message. It wouldn't be super hard to do, but I don't think it is an issue in practice - and it can be added if it is.

calmh:
Updated to retain the early creation of the default event subscription. I'm not worried about the disk event thing as there was never an assumption that it would grab everything from the start.

AudriusButkevicius:
I say we keep it like it was before, there is no harm in doing so.

calmh:
This is like it was before?

AudriusButkevicius:
I don't see the initialization of the disk sub, which is going to confuse people as changes are not recorded if the UI has not been loaded, which is not something you'd expect.

calmh:
Oh, that! Yeah, sorry, I'd forgotten that this is actually used by the GUI itself. I'll add that back.

calmh:

those ignores again... :/

AudriusButkevicius:
Wut the build failure looks amusing.

imsodin:
I don't consider them so amusing anymore...
As it isn't failing consistently (has it ever?), shall we increase the sleep to 10ms [here](https://github.com/syncthing/syncthing/blob/master/lib/model/model_test.go#L961) just to make it a bit less likely and thus even more frustrating should it ever come up again :P

calmh:
Maybe we should just trust the whole ignore system as an article of faith and prohibit any tests of it

calmh:
So, apart from the luls around the spurious ignore failures, are we good here?

AudriusButkevicius:
Will this work I wonder.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as d48e46a29c386b2e857f73b13d3273b41871d01b. Thanks, @calmh!

calmh:
Looks like no. I was actually thinking about having the bot pick up github's "approval" as an lgtm though.

PR #4095, opened 2017-04-13
lib/model: Create root directory for paused folders (fixes #4094)

### LINKED ISSUE #4094: When adding ignores at folder creation time, .stignore file can't be created
When creating a folder share, specified Ignore Patterns won't save unless the _path/.stignore_ yet exists.
Since #3608 the Ignore Patterns button logic is OK (grayed when no path is given), but if it is not possible to create the full path to .stignore on create, then the button should not even show (or remain grayed) as long as the folder share is not created (then needing to re-open/edit the folder share to set Ignores, allowing to drop synching selected Ignores... and go to the directory tree to do some clean-up)

imsodin:
see #4094

One possible caveat: Paused folders in config.xml are not handled (e.g. https://github.com/syncthing/syncthing/blob/master/cmd/syncthing/main.go#L817), but possibly other places too (?).

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -812,6 +812,16 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 		setPauseState(cfg, true)
 	}
 
+	// Add and start folders
+	for _, folderCfg := range cfg.Folders() {
+		if folderCfg.Paused {
+			folderCfg.CreateRoot()
+			continue
+		}
+		m.AddFolder(folderCfg)
+		m.StartFolder(folderCfg.ID)
+	}
+
 	mainService.Add(m)
 
 	// Start discovery
@@ -856,8 +866,6 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 			cachedDiscovery.Add(mcd, 0, 0, ipv6LocalDiscoveryPriority)
 		}
 	}
-
-	cfg.Replace(cfg.RawCopy())
 
 	// GUI
 
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -812,15 +812,6 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 		setPauseState(cfg, true)
 	}
 
-	// Add and start folders
-	for _, folderCfg := range cfg.Folders() {
-		if folderCfg.Paused {
-			continue
-		}
-		m.AddFolder(folderCfg)
-		m.StartFolder(folderCfg.ID)
-	}
-
 	mainService.Add(m)
 
 	// Start discovery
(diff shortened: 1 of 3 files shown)


lib/model/model.go
```diff
@@ -231,18 +231,9 @@ func (m *Model) startFolderLocked(folder string) config.FolderType {
...
-			// begin with.
-			permBits = 0700
-		}
 
-		if _, err := os.Stat(cfg.Path()); os.IsNotExist(err) {
-			if err := osutil.MkdirAll(cfg.Path(), permBits); err != nil {
-				l.Warnln("Creating folder:", err)
-			}
-		}
+		createRootFolder(cfg)
```
calmh: I've moved off the real computer for the evening so I'm not gong to test tonight, but I can bitch about the naming :D it's a folder root, not a root folder. ;)

AudriusButkevicius:
Lgtm

calmh:
Sorry for letting stew. I want to test this (and the other PR) a bit before merging, and haven't had time to do so yet.

calmh:
This looks sane and probably works, but it doesn't solve the underlying problem. That is, if I 

 - "add folder"
 - fill in a path
 - hit "ignore patterns"
 - fill in some crap
 - "save"
 - "save"

the folder gets added, the directory is created, no error is shown, but no ignore patterns are created in the folder

calmh:
Friendly reminder here just to make sure it's not forgotten or anything :)

imsodin:
Not forgotten, but off skitouring = offline.
I cannot reproduce your problem, the .stignore file is created with the entered content when I follow these steps.

calmh:
Ah, cool, have fun. Let me see if I can narrow it down and fix it.

calmh:
Found the issue, there is a race condition in the javscript which I'll fix in a followup commit.

calmh:
@st-review merge

st-review:
:ok_hand: Merged as 914b09fd1f77ecefdb9b52af607ca12f5d51e756. Thanks, @imsodin!

cmd/syncthing/main.go
```diff
@@ -815,6 +815,7 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 	// Add and start folders
 	for _, folderCfg := range cfg.Folders() {
 		if folderCfg.Paused {
+			folderCfg.CreateRoot()
```
calmh: @imsodin Happened to be looking back at this... This is a problem if we're starting up, a folder is paused, and the folder is unmounted or otherwise missing. We will then create a new, empty, root for it. If the folder is then unpaused, we'll assume it became empty and sync a complete delete of it...

imsodin: @calmh I seem to remember that this came up before but was considered not a problem, as the missing folder marker prevents it?

AudriusButkevicius: CreateRoot does not create the marker though?

imsodin: It doesn't, that's done by `cfg.CreateMarker`.

PR #4102, opened 2017-04-14
Reducing height of textarea in editIgnoresModalView modal

jonathancross:
### Purpose

The height of the 'ignore patterns' modal is excessively large and can cause unnecessary display issues on smaller screens.  Important items like the *Save* button get pushed below the bottom of the screen.  The screen confusion is amplified by the fact that this is a modal inside another modal.

All of this is unnecessary because an HTML `textarea` element will scroll if the user adds more than the 5 lines.  Additionally, in modern browsers like Chrome, user can also grab the bottom-right corner and drag to make the textarea larger in case that is desirable.

### Testing

Tested in all major browsers. A very simple change to standard HTML attribute.
I experimented with several heights, 5 lines was chosen because it clearly indicates multiple lines are possible and will be plenty to get any user started.

### Screenshots

syncthing-modal

### Documentation

Documentation is not necessary.

### Authorship

**I agree to the above.**

### DIFF
--- a/gui/default/syncthing/folder/editIgnoresModalView.html
+++ b/gui/default/syncthing/folder/editIgnoresModalView.html
@@ -1,7 +1,7 @@
 <modal id="editIgnores" status="default" heading="{{'Ignore Patterns' | translate}}" large="yes" closeable="yes">
   <div class="modal-body">
     <p translate>Enter ignore patterns, one per line.</p>
-    <textarea class="form-control" rows="15"></textarea>
+    <textarea class="form-control" rows="5"></textarea>
 
     <hr/>
 


calmh:
Thanks! The change itself looks fine to me. We don't add authorship for these kinds of patches though, so please don't do the changes to AUTHORS, NICKS etc here. (It's always better to let a maintainer handle that, even though you've managed it better than most do on their first attempt :D)

jonathancross:
> please don't do the changes to AUTHORS, NICKS...

Okay, I'll remove then. :-)

jonathancross:
Okay, removed AUTHORS (and related files).

calmh:
@st-review merge it

gui: Reduce height of textarea in ignores modal

Skip-check: authors

st-review:

st-review:
:ok_hand: Merged as f689512a3fb3522c5633682a99a2cb138b204c93. Thanks, @jonathancross!

PR #4104, opened 2017-04-15
Mark initial scan as finished even if failed and refactor (fixes #4103)

### LINKED ISSUE #4103: Folders with missing path are not restarted by "Rescan all"
Dear Team

Usecase:
I share a few larger, slow changing folders from inside a truecrypt container. I use a high rescan interval, because I use to work with the data and I dont want to sync before I am done. (work is ususally below 10 min, like copying pictures inside and renaming them.) I set the other side to "versioning" and I dont want copies of all that renaming on that side.
After mounting, I usually click "rescan all" to get them from stopped to alive.

what happend:
With the change from v0.14.26 to v0.14.27, the functionality of the "Rescan all" Button changed.
Before: 
Stopped Folders would be checked if they are "alive" again
Now:
Stopped Folders are skipped during the "Rescan all" function

Folders can only get "alive" again, if the rescan interval is due.

what you expected to happen instead:
Stopped Folders would be checked if they are "alive" again

any steps to reproduce the problem:
unmount a shared folder 
start Syncthing
rescan the folder > status is "Stopped"
mount folder again
press "Rescan all" or "Rescan" on the individual folder. (Same effect: No effect)

### Version Information

Syncthing Version: v0.14.27 (rc2 and rc3)
OS Version: Windows 7 
Browser Version: Using Syncthing Trazor
grafik

imsodin:
The mechanism to disallow manual scans before the initial scan completed (#3996) , had the side effect, that if the initial scan failed, no further scans are allowed. So this marks the initial scan as finished regardless of whether it succeeded or not.

There was also redundant code in rofolder and a pointless check for folder health in `scanSubsIfHealthy` (happens in `internalScanFolderSubdirs` as well). This also moves logging from folder.go to ro/rw-folder.go to include the information about whether it is send-only or send-receive.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -11,10 +11,10 @@ import "time"
 type folder struct {
 	stateTracker
 
-	scan                 folderScanner
-	model                *Model
-	stop                 chan struct{}
-	initialScanCompleted chan struct{}
+	scan                folderScanner
+	model               *Model
+	stop                chan struct{}
+	initialScanFinished chan struct{}
 }
 
 func (f *folder) IndexUpdated() {
@@ -25,7 +25,7 @@ func (f *folder) DelayScan(next time.Duration) {
 }
 
 func (f *folder) Scan(subdirs []string) error {
-	<-f.initialScanCompleted
+	<-f.initialScanFinished
 	return f.scan.Scan(subdirs)
 }
 func (f *folder) Stop() {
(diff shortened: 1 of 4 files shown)


imsodin:
@st-jenkins test this please

calmh:
lgtm, although the s/Completed/Finished/ seems spurious

calmh:
@st-review merge it

lib/model: Mark initial scan as finished even if failed and refactor (fixes #4103)

The mechanism to disallow manual scans before the initial scan completed (#3996) , had the side effect, that if the initial scan failed, no further scans are allowed. So this marks the initial scan as finished regardless of whether it succeeded or not.

There was also redundant code in rofolder and a pointless check for folder health in scanSubsIfHealthy (happens in internalScanFolderSubdirs as well). This also moves logging from folder.go to ro/rw-folder.go to include the information about whether it is send-only or send-receive

st-review:
:ok_hand: Merged as 6b188ebcf34f7c82e636c7bec772537a839f1a2d. Thanks, @imsodin!

AudriusButkevicius:
Shouldn't we log the errors though?

PR #4114, opened 2017-04-22
lib/db, lib/model: Always use reasonable sized batches (fixes #2250, fixes #4112)

### LINKED ISSUE #2250: Removing large shared folder causes very high RAM usage
I removed a shared folder of 509GB in 350 items. It took 15 minutes for Syncthing to finish "Cleaning data for dropped folder", during which it peaked at 5.2GB private bytes, and spent most of the time above 4GB. This strikes me as being a bit excessive!

### LINKED ISSUE #4112: "Override changes" should use less memory
Hi there,

it might be that what I'm trying to do simply isn't possible. I'm running syncthing on an (old) Atom N230 system with only 2Gb of ram running WindowsServer2008, 32bit, yes I know it does not make sense =) but it only runs Syncthing and CrashPlan and has been working fine for years. Most of the time the memory usage hoovers around 800Mb and although the CPU isn't very powerful when it comes down to hashing etc, it's I/O subsystem is sufficiently adequate to make it work. In total I've been syncing between 1 and 2 TB between that machine and another one across the internet. Like I said, this has worked fine for a very long time. Thank you very much for such a nice piece of software!

But.... recently one of the larger folders got out of sync and I'm not sure why. Although the folder is "Send Only" it shows the button [Override Changes]

299946ffbc7b7bf74d65fd5d791d95f9d3d04c5c_1_690x385

Now, when I press this button memory usage goes through the roof quickly and within a minute or so Syncthing throws out a panic log and restarts.

02ea5745437483db26374c019f688e35160969a1

This happens every time I try to [Override Changes]. Reading [this ](https://forum.syncthing.net/t/out-of-memory-panic-log/9072) post the issue might be related to something with the Weak Hashes. I could try to apply the proposed fix and trick the system into not using Weak Hashes and see where this brings me but I was mostly wondering if it would be useful for you guys to have me try something else first? (I probably could take a backup of %APPDATA%\Syncthing first so we can 'revert back' to the 'bad situation' if needed?)

Panic log in attachment, redacted some parts to avoid script kiddies going wild on my ADSL line =)

Thanks upfront for having a look at this,
Roby

[panic-20170420-132111.txt](https://github.com/syncthing/syncthing/files/946590/panic-20170420-132111.txt)

### Version Information

Syncthing Version: syncthing v0.14.26 "Dysprosium Dragonfly" (go1.8 windows-386) jenkins@build.syncthing.net 2017-03-23 11:04:35 UTC
OS Version: Microsoft Windows [Version 6.0.6002] (Server 2008 SP2, x86)

calmh:
### Purpose

Harmonize how we use batches in the model, using ProtoSize() to judge the actual weight of the entire batch instead of estimating. Use smaller batches in the block map - I think we might have though that batch.Len() in the leveldb was the batch size in bytes, but it's actually number of operations.

### Testing

Tried both the override and remove operations on my collection of sample data. With the old version memory usage ballooned ~80 -> ~390 MB. With this change, it stays at ~100 MB instead.

There could conceivably be a performance impact from the smaller batches. I didn't see any statistically relevant change on the ManyFiles benchmark though, and if it's small it's probably worth it anyway.

### DIFF
--- a/lib/db/blockmap.go
+++ b/lib/db/blockmap.go
@@ -19,7 +19,7 @@ import (
 
 var blockFinder *BlockFinder
 
-const maxBatchSize = 256 << 10
+const maxBatchSize = 1000
 
 type BlockMap struct {
 	db     *Instance
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -40,8 +40,8 @@ import (
 
 // How many files to send in each Index/IndexUpdate message.
 const (
-	indexTargetSize = 250 * 1024 // Aim for making index messages no larger than 250 KiB (uncompressed)
-	indexBatchSize  = 1000       // Either way, don't include more files than this
+	maxBatchSizeBytes = 250 * 1024 // Aim for making index messages no larger than 250 KiB (uncompressed)
+	maxBatchSizeFiles = 1000       // Either way, don't include more files than this
 )
 
 type service interface {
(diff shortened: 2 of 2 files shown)


AudriusButkevicius:
@st-review merge

st-review:

st-review:
:ok_hand: Merged as e9f05d138f802075e58b4e3688d735bfcd087fc0. Thanks, @calmh!

PR #4116, opened 2017-04-23
gui: Restrict GUI listening ports to >1024, add help (fixes #4020)

### LINKED ISSUE #4020: Prevent binding the GUI to a privileged port
It seems a fairly common mistake to set the GUI to port 80, after which it can't start and is inaccessible. The settings dialog should probably not allow setting ports below 1024. To do that anyway, the advanced settings dialog can be used.

calmh:
This adds a pattern validator to the GUI listen port field that checks for port numbers 1024 and above. Also adds a help link pointing to the (new) page talking about GUI listen port numbers. That page has information on how to work around the restriction, in general terms. The related pull request is here: https://github.com/syncthing/docs/pull/315

Also changes the header from "GUI Listen Addresses" to the singular version, because we only support one listen address today.

screen shot 2017-04-23 at 10 40 47

screen shot 2017-04-23 at 10 40 57


AudriusButkevicius:
Seems it accepts 01025, does go accept that?

calmh:
Yup: https://play.golang.org/p/7Ewrf2QaZw

There is however one bug, 127.0.0.1:00080 is accepted by the regexp. Let me fix that.

calmh:
Fixed, now passes the tests I can think of, manually.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 26654df48c11380b70780c30ba77046bbcc596c6. Thanks, @calmh!

PR #4117, opened 2017-04-25
lib/fs, lib/model, lib/scanner: Make scans cancellable (fixes #3965)

### LINKED ISSUE #3965: lib/model: Duplicate scans can happen when folder is restarted
If the folder takes too long to complete the fist scanning, a second process seems to kick in a while after.

calmh:
### Purpose

The folder already knew how to stop properly, but the fs.Walk() didn't
and can potentially take a very long time. This adds context support to
Walk and the underlying scanning stuff, and passes in an appropriate
context from above. The stop channel in model.folder is replaced with a
context for this purpose.

### Testing

To test I added an infiniteFS that represents a large amount of data
(not actually infinite, but close) and verify that walking it is
properly stopped. For that to be implemented smoothly I moved out the
Walk function to it's own type, as typically the implementer of a new
filesystem type might not need or want to reimplement Walk.

It's somewhat tricky to test that this actually works properly on the
actual sendReceiveFolder and so on, as those are started from inside the
model and the filesystem isn't easily pluggable etc. Instead I've tested
that part manually by adding a huge folder and verifying that pause,
resume and reconfig do the right things by looking at debug output.

### DIFF
--- a/cmd/stfileinfo/main.go
+++ b/cmd/stfileinfo/main.go
@@ -7,6 +7,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"log"
 	"os"
@@ -70,7 +71,7 @@ func main() {
 		if *standardBlocks || blockSize < protocol.BlockSize {
 			blockSize = protocol.BlockSize
 		}
-		bs, err := scanner.Blocks(fd, blockSize, fi.Size(), nil, true)
+		bs, err := scanner.Blocks(context.TODO(), fd, blockSize, fi.Size(), nil, true)
 		if err != nil {
 			log.Fatal(err)
 		}
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -17,7 +17,7 @@ type folder struct {
 	scan                folderScanner
 	model               *Model
 	ctx                 context.Context
-	cancel              func()
+	cancel              context.CancelFunc
 	initialScanFinished chan struct{}
 }
 
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -6,14 +6,18 @@
 
 package model
 
-import "time"
+import (
+	"context"
+	"time"
+)
 
 type folder struct {
 	stateTracker
 
 	scan                folderScanner
 	model               *Model
-	stop                chan struct{}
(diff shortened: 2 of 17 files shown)


lib/scanner/walk_test.go
```diff
@@ -458,3 +459,44 @@ func initTestFile() {
...
+	// Cancel the walker.
+	cancel()
+
+	// Empty out any waiting entries and wait for the channel to close.
+	// Count them, they should be zero or very few.
+	extra := 0
+	for range fchan {
+		extra++
+	}
+	if extra > 4 {
```
AudriusButkevicius: The 4 is mysterious. I understand it's number of cores, but still.

cmd/stfileinfo/main.go
```diff
@@ -70,7 +71,7 @@ func main() {
 		if *standardBlocks || blockSize < protocol.BlockSize {
 			blockSize = protocol.BlockSize
 		}
-		bs, err := scanner.Blocks(fd, blockSize, fi.Size(), nil, true)
+		bs, err := scanner.Blocks(context.TODO(), fd, blockSize, fi.Size(), nil, true)
```
AudriusButkevicius: I wonder if we could stuff some of the cruft (sizeHint, Counter, useWeakHash, Counter) into context.WithValue.
For Counter and sizeHint it mostly makes sense as these are optional

lib/model/folder.go
```diff
@@ -6,14 +6,18 @@
...
+)
 
 type folder struct {
 	stateTracker
 
 	scan                folderScanner
 	model               *Model
-	stop                chan struct{}
+	ctx                 context.Context
+	cancel              func()
```
AudriusButkevicius: context.CancelFunc

calmh:
I pushed fixes for the comments. We could maybe move some things into the context, but preferably not as part of this PR.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as d6fbfc35452224c0103d7e2297d0df1988dc2193. Thanks, @calmh!

PR #4119, opened 2017-04-26
Filesystem Notifications - UI

imsodin:
This is about the changes in the web UI due to #3986.

Not that this is based on the branch plouj-fswatcher and not master to visually separate the changes. It is generally not polished yet - thus should be considered WIP.

PR #4120, opened 2017-04-26
script: Ignore units and allow translated strings as translate-values

imsodin:
This extends the regex of ignored strings (i.e. no warning about being untranslated). This regex gets increasingly sillyish, maybe at some point it would be worth to switch to something like the earlier proposed `data-translatable="no"` attribute.

Also I removed a break statement to also check translate-values for translated strings. This allows to separate terms that are "names" and should be consistent in all translations. An example can be seen in #4119 where I added each occurrence of "filesystem notifications" as a translate-value. This could also be used for "Folder ID" and similar terms.
Doing this is a trade-off between simplicity and consistent translations. I think it is worth doing. Opinions?

### DIFF
--- a/script/translate.go
+++ b/script/translate.go
@@ -24,7 +24,7 @@ var attrRe = regexp.MustCompile(`\{\{'([^']+)'\s+\|\s+translate\}\}`)
 
 // exceptions to the untranslated text warning
 var noStringRe = regexp.MustCompile(
-	`^((\W*\{\{.*?\}\} ?.?\/?.?(bps)?\W*)+(\.stignore)?|[^a-zA-Z]+.?[^a-zA-Z]*|(k|M|G|T)B|Twitter|JS\W?|DEV|https?://\S+)$`)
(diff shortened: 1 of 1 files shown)


script/translate.go
```diff
@@ -24,7 +24,7 @@ var attrRe = regexp.MustCompile(`\{\{'([^']+)'\s+\|\s+translate\}\}`)
 
 // exceptions to the untranslated text warning
 var noStringRe = regexp.MustCompile(
-	`^((\W*\{\{.*?\}\} ?.?\W*)+(\.stignore)?|[^a-zA-Z]+.?[^a-zA-Z]*|Twitter|JS\W?|DEV|https?://\S+)$`)
+	`^((\W*\{\{.*?\}\} ?.?\/?.?(bps)?\W*)+(\.stignore)?|[^a-zA-Z]+.?[^a-zA-Z]*|(k|M|G|T)B|Twitter|JS\W?|DEV|https?://\S+)$`)
```
canton7: `[kMGT]B` ?

imsodin: Thanks. Kind of highlights what I meant by it being sillyish :P

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 9de6cdddfdfbc6c88cbbca248c51341ffa5aff8a. Thanks, @imsodin!

PR #4122, opened 2017-05-01
lib/ignore: Don't match root (".")

imsodin:
When the line "/*" or "*" is present in ".stignore", the folder root (i.e. ".") is ignored. This is a problem when files are explicitly included (e.g. "!/a*") and it generally doesn't make sense to ignore the entire folder.

This surfaced in syncthing/syncthing/inotify#161 and wasn't a problem for Syncthing, as the root folder case was already handled in walk.go.


calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 0b854dff9d23159f399da1e77d360b3d9be0c938. Thanks, @imsodin!

PR #3221, opened 2016-05-30
gui: Be more explicit about how assets should be cached (fixes #3182)

canton7:
### Purpose

With the previous setup, browsers were free to use a local cache for any
length of time they pleased: we didn't set an 'Expires' header (or
max-age directive), and Cache-Control just said "you're free to cache
this".

Therefore be more explicit: we don't mind if browsers cache things, but
they MUST revalidate everything on every request.

### Testing

Set `STTRACE=http`, reloaded lots in some browsers, saw that browsers always revalidated.

PR #4129, opened 2017-05-03
gui: Don't reload concurrently with saving config when changing theme (fixes #4127)

### LINKED ISSUE #4127: Theme change results in "Decoding posted config: unexpected EOF"
Running Syncthing v0.14.27, Windows (64 bit) on Windows 7 I tried to change the GUI theme to Black.

Syncthing seems to restrart / the browser windows reloads and then shows the message (Notice)

`2017-05-03 13:07:27: Decoding posted config: unexpected EOF `

I can only click OK.
The theme does not change.

See also [in this forum thread](https://forum.syncthing.net/t/error-when-changing-theme/8091)

### Version Information
Syncthing v0.14.27, Windows (64 bit)
Internet Explorer 11

calmh:
### Purpose

Avoid the error saving config that happens when the POST gets cancelled due to reload.

### Testing

Changed theme a few times...

There is another problem though related to caching that is not fixed by this PR. The reload done by the UI didn't actually *change* the theme for me. I needed a manual, cache-clearing, shift-reload for that to happen...

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1101,14 +1101,14 @@ angular.module('syncthing.core')
                     });
                 });
 
-                $scope.saveConfig();
+                $scope.saveConfig(function () {
+                    if (themeChanged) {
+                        document.location.reload(true);
+                    }
+                });
             }
 
             $('#settings').modal("hide");
-
-            if (themeChanged) {
-              document.location.reload(true);
-            }
         };
 
         $scope.saveAdvanced = function () {


AudriusButkevicius:
Can you check why it doesn't clear the cache?
I can't repro locally and we don't seem to set cache headers for theme files, so it should work.

Alternatively we could wrap it in a middleware that does mtime checks and force mtime changes between themes (irrespective of compilation time)

If you append a noop query string, does that force it to clear?

canton7:
I thought we set a bunch of no-cache headers for all gui assets? On mobile so can't check right now, but I'm pretty sure there was an issue with the gui not being refreshed properly after an upgrade, which I fixed with some headers. So if those headers are no longer being applied, that might be a regression.

AudriusButkevicius:
noCacheMiddleware is only used for rest API calls it seems.

canton7:
Here it is: https://github.com/syncthing/syncthing/pull/3221/files . Looks like it has gone?

calmh:
I accidentally pushed this past the process, sorry... But I think we were talking about something other than this PR in here anyway.

canton7:
Yeah, it's this: 

> There is another problem though related to caching that is not fixed by this PR. The reload done by the UI didn't actually change the theme for me. I needed a manual, cache-clearing, shift-reload for that to happen...

calmh:
Yes, I mentioned that in passing and you've probably located the issue as well.

PR #4140, opened 2017-05-09
(WIP) folderscanner refactoring

calmh:
Seeing as this branch escaped into the public consciousness and you're sneak-reviewing it already, might as well be  a PR so the comments get collected for posterity.

Expect all builds to fail and everything to be crap for the time being.


calmh:
aint' gonna happen, at least not at the moment.

PR #4148, opened 2017-05-14
cmd/syncthing: Warn when running as a super user (fixes #4123)

### LINKED ISSUE #4123: Warn against running as a privileged user
We should generally not run as a privileged user. We can't suddenly make that an error as that would break a lot of installations. But we could add a warning with similar behavior as the already existing one for unprotected GUI access. With a similar advanced config to tone it down. 

Practically, detecting uid 0 is trivial. I don't know how to detect SYSTEM on Windows.

calmh:
UID 0 on Unixes, SYSTEM SID on Windows.

### Testing

I've verified that this works on Unixes. **I'd like help to verify this is the right thing on Windows. Also, are there other SIDs we should be looking for?**

### Screenshot

screen shot 2017-05-14 at 18 12 53

### DIFF
--- a/lib/protocol/bep.proto
+++ b/lib/protocol/bep.proto
@@ -108,6 +108,7 @@ message FileInfo {
 
     repeated BlockInfo Blocks         = 16 [(gogoproto.nullable) = false];
     string             symlink_target = 17;
+    bytes encryption_iv = 18;
 }
 
 enum FileInfoType {
@@ -124,6 +125,7 @@ message BlockInfo {
     int32  size      = 2;
     bytes  hash      = 3;
     uint32 weak_hash = 4;
+    bytes encryption_iv = 5;
 }
 
 message Vector {
--- a/lib/protocol/bep.proto
+++ b/lib/protocol/bep.proto
@@ -108,7 +108,6 @@ message FileInfo {
 
     repeated BlockInfo Blocks         = 16 [(gogoproto.nullable) = false];
     string             symlink_target = 17;
-    bytes encryption_iv = 18;
 }
 
 enum FileInfoType {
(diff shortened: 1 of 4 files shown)


ProactiveServices:
Syncthing v0.14.28-rc.3+13-gd25b1526 Windows 7 x64, running as NT AUTHORITY\SYSTEM SID S-1-5-18 (both the helper syncthing.exe and "actual" syncthing.exe) is not causing the warning. Nothing in the Syncthing logs to suggest any warning either.

calmh:
Ugh. Can you run with the latest update, and `STTRACE=main`?

calmh:
Oh, ah, `gd25b1526` doesn't have this feature. This is the one to test: https://build.syncthing.net/job/syncthing-pr-windows/2398/

lib/protocol/bep.proto
```diff
@@ -108,7 +108,6 @@ message FileInfo {
 
     repeated BlockInfo Blocks         = 16 [(gogoproto.nullable) = false];
     string             symlink_target = 17;
-    bytes encryption_iv = 18;
```
AudriusButkevicius: hmm

AudriusButkevicius: I still think this is best done via a custom filesystem translator ;)

ProactiveServices:
Warning is present with the linked exe, both in GUI and console.

calmh:
\o/

calmh:
@st-review merge it, it's been tested

st-review:
:ok_hand: Merged as 1d15b8be9bfcd1e7e4063b3169e7d8bd727b7450. Thanks, @calmh!

PR #4151, opened 2017-05-16
Adding simple init script for older linux releases like CentOS 6

viable-hartman:
### Purpose

Providing a simple init script for older Linux distros like CentOS 6

### Testing

I put this init script on a CentOS 6 system with syncthing installed and used it to start, stop, restart, and check the status of syncthing.

### DIFF
--- a/etc/linux-init/syncthing
+++ b/etc/linux-init/syncthing
@@ -1,0 +1,94 @@
+#!/bin/sh
+#
+# syncthing          Start/Stop the SyncThing daemon.
+#
+# chkconfig: 2345 90 60
+# description: SyncThing lets you sync
+#              and share an unlimited number of files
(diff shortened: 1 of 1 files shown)


calmh:
Hi! Thanks for the contribution. However, this init script has several problems:

 - It misspells "Syncthing"
 - It runs Syncthing as root, which you should never do
 - It explicitly logs everything to /dev/null which makes things hard to troubleshoot
 - It does a very dubious dance with pid files, given that Syncthing runs with two processes and I think it's random which gets into the pid file

viable-hartman:
Yeah, its certainly not the best init script in the world.  I just needed something that works for older Linux distros and was quick to make, so I hacked this one together.  Figured other people might find it useful.  When I get more time I'll clean it up a bit.

Issue #3524, opened 2016-08-18
Device name in conflicts?


Lillecarl:
When conflicts occur, we get a file & timestamp like this: Testfile.sync-conflict-20160818-134336.txt and the original Testfile.txt

What would be cool would be (at least as an option) to include the device name in the conflict file, so that you can easily determine which file is from "your node" so to speak.

This would be useful since we don't know who "initated" the file changes first (if there's a conflict).

This will also be useful for GUI tools that, so they can tell which is "your" file.

Best regards
Carl Hjerpe

AudriusButkevicius:
I am pretty sure it's a duplicate, yet I can't find the original ticket.

calmh:
There was a pull request. It's not super trivial. We could add the local device name to the conflict file when creating it though, which could be useful when it's synced to other devices.

Lillecarl:
@calmh I haven't read any code yet in Syncthing, I'm thinking about a way for all syncthings to know which file is which (as in renaming both files to wherever they came from). But i don't yet know how things work, so i can't say too much.
But indeed, creating the file with the devicename could be v0.1, there can always be further improvements lateron :)

AudriusButkevicius:
I actually think it''s impossible in some cases, as we don't track where the file originated from, and if you were offline for N hours, come back online, there are 10 other devices which have a conflicting copy which would be considered inferior to the one you have, which device are you conflicting with? All 10 of them?

AudriusButkevicius:
Also we would never use device name, as these things vary between devices and can be user adjusted, so might make no sense on other machines. Lastly when we create a conflict it becomes just "another file" with a different name, so having different names on different sides makes no sense, as the conflict resolution logic is the same on all devices, therefore the outcome is also the same on all sides, hence the name should be the same.

Lillecarl:
@AudriusButkevicius When a device detects a conflict, it creates another file. Which will be synced with everyone, then including some kind of name would help others know who it originated from.

Imaginary example:
User 1 has his computer offline, writes things to document & save, then comes online (without instantly noticing the conflict).
User 2 has his computer online, writes things to document & save.
User 3 has his computer online, sees that there's a conflict, sees that the conflict is with User1, which is a respectable member of this folder family, and therefore he manually merges them, then remove the conflict files.

I think it could be valuable, but might not fit syncthings goals :)

AudriusButkevicius:
It's hard to understand if you are describing of how you think it works, or how youdl'd loke it to work.

[...]

PR #4161, opened 2017-05-20
Nodename in conflictfile

nrm21:
### Purpose

Fulfills issue #3524 request

### Testing

Changed a file on 3 different (ntp synced) computers at a 2 second internal difference.  Client A, B, C in that order.  Client C's file remained the master, the other 2 reacted accordingly.

All 3 had different info thus 2 conflicts were created (in addition to master).
If only 2 computers have conflict then only one conflict will be produced.

### Screenshots

After 2 pull intervals (so all clients could receive all copies):
[http://i.imgur.com/m3TkNog.png](http://i.imgur.com/m3TkNog.png)

### Authorship

Nate Morrison (nrm21)

### DIFF
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -1692,7 +1692,7 @@ func (f *sendReceiveFolder) moveForConflict(name string) error {
 		err = nil
 	}
 	if f.MaxConflicts > -1 {
-		matches, gerr := osutil.Glob(withoutExt + ".sync-conflict-????????-??????" + ext)
+		matches, gerr := osutil.Glob(withoutExt + ".sync-conflict-????????-??????-???????" + ext)
 		if gerr == nil && len(matches) > f.MaxConflicts {
 			sort.Sort(sort.Reverse(sort.StringSlice(matches)))
 			for _, match := range matches[f.MaxConflicts:] {
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -1694,7 +1694,7 @@ func (f *sendReceiveFolder) moveForConflict(name string, lastModBy string) error
 		err = nil
 	}
 	if f.MaxConflicts > -1 {
-		matches, gerr := osutil.Glob(withoutExt + ".sync-conflict-????????-??????.*" + ext)
+		matches, gerr := osutil.Glob(withoutExt + ".sync-conflict-????????-??????*" + ext)
 		if gerr == nil && len(matches) > f.MaxConflicts {
 			sort.Sort(sort.Reverse(sort.StringSlice(matches)))
 			for _, match := range matches[f.MaxConflicts:] {
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -86,6 +86,7 @@ type sendReceiveFolder struct {
 
 	mtimeFS   *fs.MtimeFS
 	dir       string
+	lastModBy string
 	versioner versioner.Versioner
 	sleep     time.Duration
 	pause     time.Duration
@@ -852,6 +853,7 @@ func (f *sendReceiveFolder) deleteFile(file protocol.FileInfo) {
 		// of deleting. Also merge with the version vector we had, to indicate
 		// we have resolved the conflict.
 		file.Version = file.Version.Merge(cur.Version)
+		f.lastModBy = file.ModifiedBy.String()
 		err = osutil.InWritableDir(f.moveForConflict, realName)
 	} else if f.versioner != nil {
 		err = osutil.InWritableDir(f.versioner.Archive, realName)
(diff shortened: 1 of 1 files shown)


nrm21:
Possibly could use client friendly name but... I was having trouble with that so.  Lets just stick with shortID.

lib/model/rwfolder.go
```diff
@@ -1454,6 +1456,7 @@ func (f *sendReceiveFolder) performFinish(state *sharedPullerState) error {
 			// we have resolved the conflict.
 
 			state.file.Version = state.file.Version.Merge(state.version)
+			f.lastModBy = state.file.ModifiedBy.String()
 			if err = osutil.InWritableDir(f.moveForConflict, state.realName); err != nil {
```
AudriusButkevicius: Remove the f.lastModBy (as that is racey as hell, if we handle multiple files at the time for example, the attribute might change to some undefined value which makes no sense)

And use:

```
osutil.InWritableDir(func(name string) {
 f.moveForConflict(name, state.ModifiedBy.String())
}, state.realName);
```
and replace the signature of moveForConflict to accept ModifiedBy.

lib/model/rwfolder.go
```diff
@@ -1689,7 +1692,7 @@ func (f *sendReceiveFolder) moveForConflict(name string) error {
 		err = nil
 	}
 	if f.MaxConflicts > -1 {
-		matches, gerr := osutil.Glob(withoutExt + ".sync-conflict-????????-??????" + ext)
```
canton7: Probably want to not ignore old conflict files here?

AudriusButkevicius:
Also, we need to take into account existing conflict files (for cleanup purposes).

lib/model/rwfolder.go
nrm21: maybe I should change it too ".sync-conflict-????????-??????.*"

lib/model/rwfolder.go
nrm21: It gives me this if I try:

cannot use func literal (type func(string)) as type func(string) error in argument to osutil.InWritableDir

It might be because I changed moveForConflict(string) to moveForConflict(string, string).  But how else will I get the modby in there?

AudriusButkevicius: it should be ```return f.moveForConflict(string, string)```

nrm21: still not working, Now I'm getting:
have too many args to return   have (error)   want ()

			`state.file.Version = state.file.Version.Merge(state.version)
			err = osutil.InWritableDir(func(name string) {
				return f.moveForConflict(name, state.file.ModifiedBy.String())
			}, state.realName)
			if err != nil {
				return err
			}`

AudriusButkevicius:
```
err = osutil.InWritableDir(func(name string) error {
	return f.moveForConflict(name, state.file.ModifiedBy.String())
}, state.realName)
```

nrm21: Thank you that was it.

lib/model/rwfolder.go
```diff
@@ -1689,7 +1694,7 @@ func (f *sendReceiveFolder) moveForConflict(name string) error {
...
+		matches, gerr := osutil.Glob(withoutExt + ".sync-conflict-????????-??????.*" + ext)
```
AudriusButkevicius: I'm not convinced this works. Did you test it?

nrm21: no not yet... busy day today.  But isn't there a unit test for that code?

nrm21: Yeah according to this page it looks like it's not regex that the function accepts (which I just assumed it was) https://golang.org/pkg/path/filepath/#Glob

lib/model/rwfolder.go
canton7: Is there any disadvantage to doing `.sync-conflict-*`?

lib/model/rwfolder.go
canton7: Indeed: `?????` is not a valid regex.

nrm21:
Anyone know what we're waiting on for this?  I'm not totally sure what Glob does but the code suggests it has to do with max conflicts.  So when I set it to 2 to won't make another conflict if there are already 2 conflicts for the file.

Isn't that what it's suppose to do?  Or should it behave another way?

AudriusButkevicius:
@st-review lgtm

st-review:

AudriusButkevicius:
@calmh ping

calmh:
@st-review lgtm

st-review:
@calmh: Merge failed:

```
$ git fetch -f origin refs/pull/4161/head:pr-4161
   91d37f3..dd6bb6d  master     -> orig/master
$ git reset --hard
HEAD is now at dd6bb6d gui: Reset warnings about nested folder roots (fixes #3433)
$ git checkout master
Already on 'master'
HEAD is now at dd6bb6d gui: Reset warnings about nested folder roots (fixes #3433)
$ git clean -fxd
$ git merge --squash --no-commit pr-4161
$ git commit -F -
Commit message subject:

    Add name of latest modifier node to conflict file

match this pattern:

    ^[\w/,\. ]+: \w

```

calmh:
Aww :(

calmh:
@st-review merge it

lib/model: Add name of latest modifying device to conflict file (fixes #3524)

st-review:
:ok_hand: Merged as 1235cead3530985e8e6700db869bffa25755bf61. Thanks, @nrm21!

PR #4168, opened 2017-05-22
lib/model: Use up to date device name, do not provide name to unknown devices (fixes #4164)

### LINKED ISSUE #4164: Changed system name requires restart to take effect in outgoing hello messages
Synchthing ignores user Device Name change and advertises user System Name. A Security Risk

The synchthing user can change their device name from what appears to be the coding pulling that directly from the system platform name. However the system platform name is still shown to all remote peers your ID is shared with. That's a real problem. 

Urgent suggestion, when the synchthing user changes the Device Name in the Webui, that is the Device Name that shows to any and all remote peers you share your ID with downstream.

AudriusButkevicius:
### Purpose

As per ticket.

### Testing

Em, it compiles.

### DIFF
--- a/lib/config/wrapper.go
+++ b/lib/config/wrapper.go
@@ -436,10 +436,5 @@ func (w *Wrapper) MyName() string {
 	w.mut.Lock()
 	myID := w.cfg.MyID
 	w.mut.Unlock()
-	device, ok := w.Device(myID)
-	if !ok {
-		l.Warnln("BUG: No own device present")
-		return ""
-	}
-	return device.Name
+	return w.Device(myID).Name
 }
--- a/lib/config/wrapper.go
+++ b/lib/config/wrapper.go
@@ -431,3 +431,15 @@ func (w *Wrapper) StunServers() []string {
 
 	return addresses
 }
+
+func (w *Wrapper) MyName() string {
+	w.mut.Lock()
+	myID := w.cfg.MyID
+	w.mut.Unlock()
+	device, ok := w.Device(myID)
+	if !ok {
+		l.Warnln("BUG: No own device present")
+		return ""
+	}
+	return device.Name
+}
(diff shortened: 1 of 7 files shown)


lib/model/model.go
```diff
@@ -1320,9 +1318,13 @@ func (m *Model) OnHello(remoteID protocol.DeviceID, addr net.Addr, hello protoco
 }
 
 // GetHello is called when we are about to connect to some remote device.
-func (m *Model) GetHello(protocol.DeviceID) protocol.HelloIntf {
+func (m *Model) GetHello(id protocol.DeviceID) protocol.HelloIntf {
+	name := ""
+	if _, ok := m.cfg.Device(id); ok {
+		name = m.cfg.MyName()
```
calmh: Use the device you got from m.cfg.Device() here and there is no point in having the cfg.MyName method any more at all?

AudriusButkevicius: The ID is the remote ID in this case

calmh: Oh. Right.

lib/config/wrapper.go
```diff
@@ -431,3 +431,15 @@ func (w *Wrapper) StunServers() []string {
 
 	return addresses
 }
+
+func (w *Wrapper) MyName() string {
+	w.mut.Lock()
+	myID := w.cfg.MyID
+	w.mut.Unlock()
+	device, ok := w.Device(myID)
```
calmh: I think just skip the comma0ok and return device.Name regardless. It should /always/ exist, and if it doesn't it's something really weird with the config that we'll fix on next startup so no need to tell the user something they can't do anything about.

calmh:
Otherwise all lgtm good stuff

AudriusButkevicius:
@st-review merge

st-review:

st-review:

calmh:
dde34b13fa7f0de5421cca8888ff29b1_nelson-ha-ha-meme-ha-memes_400-373

AudriusButkevicius:
@st-review merge

st-review:

st-review:
:ok_hand: Merged as 91d37f35bc7dc6bd08999e85c9d91d54c48fab32. Thanks, @AudriusButkevicius!

PR #4173, opened 2017-05-23
gui: Reset warnings about nested folder roots (fixes #3433)

### LINKED ISSUE #3433: gui: Warning "path is a subdirectory of other folder" persists between dialogs
Seen in ubuntu firefox
Starting to create a new share, then giving a path through a yet shared one the warning (string 139) triggers OK. Then cancelling the creation and immediately starting a new creation, the warning still shows even though the path is empty. The warning definitively disappears OK when typing / ~ or double-clic the path field... until we type again a path through an existing share path, which is OK.

Seems to be some variable not reset by Close button. Same issue when validating with Save.

imsodin:
Refer to #3433 for a description of the problem.

To reset the warning on modal open the variables set in the `pathIsSubDir`-directive need to be reset. To make this easier I collected them in one object called `folderPathErrors`.

I also found one small mistake in the help text about tilde expansion (path wasn't included in the ng-if condition). And indentation polish.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -53,27 +53,30 @@ angular.module('syncthing.core')
         $scope.themes = [];
         $scope.globalChangeEvents = {};
         $scope.metricRates = false;
+        $scope.folderPathErrors = {};
 
         try {
             $scope.metricRates = (window.localStorage["metricRates"] == "true");
         } catch (exception) { }
 
         $scope.folderDefaults = {
-                selectedDevices: {},
-                type: "readwrite",
-                rescanIntervalS: 60,
-                minDiskFree: {value: 1, unit: "%"},
-                maxConflicts: 10,
-                fsync: true,
-                order: "random",
-                fileVersioningSelector: "none",
-                trashcanClean: 0,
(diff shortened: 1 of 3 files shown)


gui/default/syncthing/core/syncthingController.js
```diff
@@ -53,27 +53,30 @@ angular.module('syncthing.core')
...
-                simpleKeep: 5,
-                staggeredMaxAge: 365,
-                staggeredCleanInterval: 3600,
-                staggeredVersionsPath: "",
-                externalCommand: "",
-                autoNormalize: true
+            selectedDevices: {},
+            type: "readwrite",
+            rescanIntervalS: 60,
+            fsNotifications: $scope.fsNotificationsAvailable,
```
AudriusButkevicius: Hmm

imsodin: My git workspace seems to be a bit messy - thanks for noticing.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as dd6bb6d5fdd01882078d2ef1e57ca74a91daf217. Thanks, @imsodin!

Issue #3993, opened 2017-02-19
 When a share is removed from one device, other devices should be notified


alessandro-g89:
I noticed something unexpected when I remove a shared folder from a device (let's call it A): soon, all my other devices that share the same folder try to push it again to A, as if A never had that folder in the first place. The only way to stop them sending notifications is to go to each device's UI and un-select A from the list of devices sharing the folder.

I believe this behaviour is troublesome and counter-intuitive: to stop one device from sharing that folder, I also need to tweak all the other devices. A non-expert user could get very confused and not know how to fix the issue at all. What I would expect is for device A to tell the others: "I had that folder, I don't want it anymore, stop sharing it with me".
Technically, I'd say that upon deletion, A should remember the devices that were sharing the folder at that time, and try to notify each of them of the deletion. When a device is actually notified, A can forget about him having the folder. When all the devices are notified, A can finally forget about the folder altogether.

I experienced this with device A being an android smartphone, with syncthing-android v0.9.6 (latest on f-droid) and syncthing v0.14.22-rc.1+12-g5a1ee7c. Other devices are laptops and ARM devices with auto-update enabled (currently using syncthing v0.14.23).

I did some searching before posting, but I didn't find any reference of what should happen in this case, either confirming or disproving my experience.

calmh:
Just thinking out loud, we could have each device also mention folders it knows about but *does not* share, in the initial cluster config message. That way the other side can know which folders are "symmetrically" shared, which are intentionally not shared, and which are pending. It could display that on the device in the gui. It would mean that when deleting / unsharing a folder we would need to save that somewhere instead of just remove it like now.

dcarosone:
No thanks. I don't particularly want to leak the existence of other shares
to peers that shouldn't know about them.

Not every network is single user, single purpose, or fully trusted for all
data sets.

On 20 Feb. 2017 14:58, "Jakob Borg" <notifications@github.com> wrote:

Just thinking out loud, we could have each device also mention folders it
knows about but *does not* share, in the initial cluster config message.
That way the other side can know which folders are "symmetrically" shared,
which are intentionally not shared, and which are pending. It could display
that on the device in the gui. It would mean that when deleting / unsharing
a folder we would need to save that somewhere instead of just remove it
like now.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/syncthing/syncthing/issues/3993#issuecomment-280982700>,
or mute the thread

[...]

PR #4179, opened 2017-05-29
Add ability to ignore folders offered by other nodes

nrm21:
### Purpose

Add ability to ignore folders permanently (fixes request #3993)

### Screenshots

[http://i.imgur.com/LAlN5lY.png](http://i.imgur.com/LAlN5lY.png)

### Authorship

nrm21

### Notes

So here's the thing... The button still doesn't work.  I need a javascript/angular/whatever front end guy to look at it and tell me what I'm doing wrong.   It's suppose to add the folder to the config when you press the button but, it's not adding anything to the config.  When you add to the config manually (via rest API or editing the xml file in program dir) then everything works fine.  So it's just the UI code that's not working.  probably just an easy one line of code somewhere.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -216,6 +216,9 @@
               <button type="button" class="btn btn-sm btn-default" ng-click="dismissFolderRejection(event.data.folder, event.data.device)">
                 <span class="fa fa-clock-o"></span>&nbsp;<span translate>Later</span>
               </button>
+              <button type="button" class="btn btn-sm btn-default" ng-click="ignoreRejectedFolder(event.data.folder, event.data.device)">
+                <span class="fa fa-clock-o"></span>&nbsp;<span translate>Ignore</span>
+              </button>
             </div>
           </div>
         </div>
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -217,7 +217,7 @@
                 <span class="fa fa-clock-o"></span>&nbsp;<span translate>Later</span>
               </button>
               <button type="button" class="btn btn-sm btn-default" ng-click="ignoreRejectedFolder(event.data.folder, event.data.device)">
-                <span class="fa fa-clock-o"></span>&nbsp;<span translate>Ignore</span>
+                <span class="fa fa-times"></span>&nbsp;<span translate>Ignore</span>
               </button>
             </div>
           </div>
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1533,7 +1533,7 @@ angular.module('syncthing.core')
         };
 
         $scope.ignoreRejectedFolder = function (folder, device) {
-            $scope.config.ignoredFolders.push(folder.id);
+            $scope.config.ignoredFolders.push(folder);
             $scope.saveConfig();
             $scope.dismissFolderRejection(folder, device);
         };
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1532,6 +1532,12 @@ angular.module('syncthing.core')
             delete $scope.folderRejections[folder + "-" + device];
         };
 
+        $scope.ignoreRejectedFolder = function (folder, device) {
+            $scope.config.ignoredFolders.push(folder.id);
+            $scope.saveConfig();
+            $scope.dismissFolderRejection(folder, device);
+        };
+
         $scope.sharesFolder = function (folderCfg) {
             var names = [];
             folderCfg.devices.forEach(function (device) {
--- a/lib/config/config.go
+++ b/lib/config/config.go
@@ -155,6 +155,7 @@ type Configuration struct {
 	GUI            GUIConfiguration      `xml:"gui" json:"gui"`
 	Options        OptionsConfiguration  `xml:"options" json:"options"`
 	IgnoredDevices []protocol.DeviceID   `xml:"ignoredDevice" json:"ignoredDevices"`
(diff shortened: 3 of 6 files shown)


AudriusButkevicius:
@calmh so I had this when I first added the feature. You've asked to remove this as the solution to this is to get the other side to unshare it. What has changed that we now think its a good idea?

calmh:
Maybe just that I've become more used to asymmetric configs being a reality and have ceased to fight it. Look far enough back and it was a protocol error for the configs not to be identical.

AudriusButkevicius:
@st-review lgtm

st-review:

calmh:
looks good to me too but apparently doesn't work so lets have someone chime in on that first :D

gui/default/index.html
```diff
@@ -216,6 +216,9 @@ <h3 class="panel-title">
               <button type="button" class="btn btn-sm btn-default" ng-click="dismissFolderRejection(event.data.folder, event.data.device)">
                 <span class="fa fa-clock-o"></span>&nbsp;<span translate>Later</span>
               </button>
+              <button type="button" class="btn btn-sm btn-default" ng-click="ignoreRejectedFolder(event.data.folder, event.data.device)">
+                <span class="fa fa-clock-o"></span>&nbsp;<span translate>Ignore</span>
```
uok: use "fa-times" instead of "fa-clock-o"

AudriusButkevicius:
well looks to me like it should work, what's did you try adding breakpoints to the javascript debugger, and are you sure you compiled the assets in the build and did a ctrl+f5 reload?

lib/model/model.go
```diff
@@ -809,6 +809,11 @@ func (m *Model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
 			continue
 		}
 
+		if m.cfg.IgnoredFolder(folder.ID) {
+			l.Infof("Ignoring folder %s since we are configured to", folder.Label)
```
imsodin: Should probably use folder.Description().

lib/config/config.go
```diff
@@ -155,6 +155,7 @@ type Configuration struct {
 	GUI            GUIConfiguration      `xml:"gui" json:"gui"`
 	Options        OptionsConfiguration  `xml:"options" json:"options"`
 	IgnoredDevices []protocol.DeviceID   `xml:"ignoredDevice" json:"ignoredDevices"`
+	IgnoredFolders []string              `xml:"ignoredFolder" json:"ignoredFolders"`
```
calmh: Yep, it's a matter of the structure of the generated markup.

nrm21: I just copied the IgnoredDevices as much as I could on this one.

gui/default/index.html
nrm21: Done.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1532,6 +1532,12 @@ angular.module('syncthing.core')
             delete $scope.folderRejections[folder + "-" + device];
         };
 
+        $scope.ignoreRejectedFolder = function (folder, device) {
+            $scope.config.ignoredFolders.push(folder.id);
```
nrm21: On why the button isn't working Could it be that I'm using folder.id here?  I do want only a string and not the whole folder struct in the portion of the config I wrote.  But maybe I'm not understanding how this feature works.

AudriusButkevicius: event.data.folder that you are passing into this function I believe is already a string, so "string".id ends up undefined.

nrm21: Ok well either that did it OR... just a rebuild did.  Don't know which but it's working now.  :D

nrm21: Ready for PR.

calmh:
@st-review are you there?

st-review:
I'm sorry, @calmh. I'm afraid I don't know what you mean. I know how to `don't`, `lgtm`, `merge`, `prevent`, `squash`, `stop`.

lib/model/model.go
calmh: Y, and maybe "... from device %s" as well

nrm21: Done.

nrm21:
Only issue with this code is... what if a user wants to remove an ignore.  I mean it's easy enough to remove from the config.xml and restart but... will users like that?

Then again I guess that's how it is for unignore of devices also.

lib/model/model.go
imsodin: s/folder.Label/folder.Description()/ ?

nrm21: fixed.

lib/model/model.go
```diff
@@ -809,6 +809,16 @@ func (m *Model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
 			continue
 		}
 
+		if m.cfg.IgnoredFolder(folder.ID) {
+			for _, dev := range folder.Devices {
+				// Find friendly name instead of dev id and report that to console
```
calmh: Sorry, one last nitpick :/ We always log the device ID to console. (We should, at some point, gain a deviceDescription method like we have for folders, to print a short ID and label, but we don't currently have that.)

calmh:
@st-review lgtm

gui, lib/config, lib/model: Add ability to ignore folders offered by other nodes (fixes #3993)

st-review:
:ok_hand: Merged as b49bbe82ddf3ab0eae2c1482f0974a116ac50108. Thanks, @nrm21!

PR #4185, opened 2017-05-31
cmd/syncthing: Start CPU usage monitoring not from init (fixes #4183)

### LINKED ISSUE #4183: Shows zero CPU usage when started with -no-restart
CPU meter works when running normally, or when started with `STNORESTART=1`. Most likely it's due to the `!innerProcess` check that somehow fails for the command line parameter.

calmh:
Starting stuff from init() is an antipattern, and the innerProcess
variable isn't 100% reliable. We should sort out the other uses of it as
well in due time.

Also removing the hack on innerProcess as I happened to see it and the
affected versions are now <1% users.


AudriusButkevicius:
@st-review merge

st-review:

st-review:
:ok_hand: Merged as 803da92ca93c44439eaae8faf0afbbfb4d60928b. Thanks, @calmh!

PR #4186, opened 2017-05-31
gui: Add dropdown menu to choose discovered devices to device modal (fixes #4157)

### LINKED ISSUE #4157: Show "nearby devices" in add device dialog
Please add an option to easily share publicly across my network subnet and search my network subnet for these Syncthing shares.
This would make rapid deployment much more possible and effective in both a business or homegroup setting.

imsodin:
### Purpose
#4157

### Testing
Manual poking around.

### Screenshots

http://i.imgur.com/8TjO0Q8.png
http://i.imgur.com/iEIPV0w.png

### Implementation

I needed the isEmptyObject function within the modal view. I don't know what the "correct" way of doing this in JS/Angular is, I just assigned it to $scope.isEmptyObject.


gui/default/syncthing/core/syncthingController.js
```diff
@@ -1807,4 +1811,6 @@ angular.module('syncthing.core')
                 window.localStorage["metricRates"] = $scope.metricRates;
             } catch (exception) { }
         }
+
+        $scope.isEmptyObject = isEmptyObject;
```
AudriusButkevicius: You don't need this, empty object is already false'ish.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1226,6 +1227,9 @@ angular.module('syncthing.core')
 
         $scope.saveDevice = function () {
             $('#editDevice').modal('hide');
+            if ($scope.currentDevice.deviceID === '') {
+                $scope.currentDevice.deviceID = $scope.manualDeviceID;
+            }
```
AudriusButkevicius: I wonder if something like this works instead, removing the need for a scope variable

```
         <select ng-if="!editingExisting && discovery" class="form-control" ng-model="currentDevice.deviceID">
            <option ng-repeat="(id, data) in discovery" value="{{id}}">{{id}}</option>
            <option ng-value='currentDevice.deviceID' translate>Enter manually below</option>
          </select>
```

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -3,13 +3,15 @@
...
+          <option ng-repeat="(id, data) in discovery" value="{{id}}">{{id}}</option>
+        </select>
+        <input ng-if="!editingExisting && (!currentDevice.deviceID || isEmptyObject(discovery))" name="deviceID" id="deviceID" class="form-control text-monospace" type="text" ng-model="manualDeviceID" required valid-deviceid />
         <div ng-if="editingExisting" class="well well-sm text-monospace" select-on-click>{{currentDevice.deviceID}}</div>
         <p class="help-block">
-          <span translate ng-if="deviceEditor.deviceID.$valid || deviceEditor.deviceID.$pristine">The device ID to enter here can be found in the "Actions > Show ID" dialog on the other device. Spaces and dashes are optional (ignored).</span>
+          <span translate ng-if="!currentDevice.deviceID && !isEmptyObject(discovery) && deviceEditor.deviceID.$pristine">Choose a discovered device ID or enter one manually.</span>
+          <span translate ng-if="!currentDevice.deviceID && deviceEditor.deviceID.$pristine">The device ID can be found in the "Actions > Show ID" dialog on the other device. Spaces and dashes are optional (ignored).</span>
```
AudriusButkevicius: Why removal of the $valid and the false'ish check?

AudriusButkevicius:
So this is a nice feature an all, but UX/UI wise it's a car crash.

I think it might be better to make some sort of expandable thing "Devices nearby" like advanced settings in folder editor that lists the devices. We should also improve local discovery to put friendly name in the local discovery packet and display that if available.

Another alternative would be CSS HAX such as https://stackoverflow.com/a/31734324
We should check how that looks on the mobile though.

gui/default/syncthing/device/editDeviceModalView.html
imsodin: I don't see the point of telling the user where to find the device ID when a valid one has already been entered or selected -> removal of $valid and checking whether an ID has been selected in from the list.

gui/default/syncthing/core/syncthingController.js
imsodin: What would you base the ng-if of the text input field on then?

imsodin:
> So this is a nice feature an all, but UX/UI wise it's a car crash.

Agreed, could be much nicer, I just wanted a compact and quick solution because I really think that feature helps a lot for usability and I really totally missed it before. I would be happy if someone with actual frontend skills did something nice.

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: I am not sure why you need that to be honest. What's the exact problem?

imsodin: I don't want to display a text input field unless "Enter manually below" is selected. its ugly(er) and then the user could theoretically both choose an ID from the list and enter one below.

AudriusButkevicius:
> if someone with actual frontend skills did something nice.

Don't we all? like to the whole UI perhaps maybe?

calmh:
How about something more HTMLy? Maybe not the most beautiful, but quite clear?

```html
        <p class="help-block" ng-if="!editingExisting && discovery">
          <span translate>You can also select one of these nearby devices:</span>
          <ul>
            <li ng-repeat="(id, data) in discovery"><a ng-click="currentDevice.deviceID = id">{{id}}</a>
            <span class="text-muted">{{data.addresses.join(", ")}}</span></li>
          </ul>
        </p>
```

jun-01-2017 07-46-36

AudriusButkevicius:
I wonder how that looks on the fossdem wifi

calmh:
Similar to the wifi network dropdown in the middle of a busy city, if we are successful. I don't think it matters.

AudriusButkevicius:
Fine. I think think we should ramp it up a notch by adding friendly name to local discovery packets.

uok:
My vote is against friendly names unless it is clear to the user that friendly name is broadcast. "local discovery" means "in the same network", e.g. public wifi, hotel wifi - correct?

AudriusButkevicius:
Usually these sort of networks block multicast. Furthermore, on Windows, as you connect to a network it asks you what type of network it is, if you select public (like you should) local discovery would be blocked.

I also think you can probe the hostname of a machine on the local network without needing syncthing anyway, so omitting the name is more of a nuisance oppose to a privacy measure.

uok:
Ok, thanks for info. I just wanted to make sure that privacy is kept since most users only have 3 devices and adding a new device is a rare action.

calmh:
I suggest doing something with what we have now, and potentially adding info to the discovery packets as a separate stage.

imsodin:
> I wonder how that looks on the fossdem wifi

That's why I chose a dropdown menu. What about combining @calmh's version with @AudriusButkevicius proposal, to make it somewhat "scalable":

> I think it might be better to make some sort of expandable thing "Devices nearby" like advanced settings in folder editor that lists the devices.

AudriusButkevicius:
Another way to do this would be to have a separate item in the settings dropdown menu (with a bootstrap badge listing the count of devices found and also only visible if the count is non zero) which shows a modal dialog with devices nearby. As you click on one of them, it shows you the add device dialog.

Another question, does this hide devices that are already added?

uok:
I suggest to keep it simple. It would be great success for Syncthing to see a long list of Syncthing devices around them but at the moment I expect max. 5 devices to show up in this list. And only if the user has configured network, etc. correctly

calmh:
Any collective decision here or 🖌🚲🏡 ?

AudriusButkevicius:
What you suggested is better than whats in the PR if we don't want to go for something more advanced.

uok:
I think @calmh's idea is simple and fast: click/tap and done.

imsodin:
Done. I am still concerned that users from the top 5% in the usage statistics may have to do quite some scrolling on adding a device, but maybe that's not really an issue.

calmh:
We should not show devices here that are already added, which mitigates that to a certain extent.

calmh:
Also I think my idea to show the addresses needs to be somewhat restrained, at the least to not show relay addresses...

screen shot 2017-06-09 at 11 58 43

AudriusButkevicius:
I don't think we should show addresses, and cap at max 5 devices.

calmh:
Works for me. Extra points for a `... and {{items.length - 5}} other devices` or something under the list if it overflows. But really, we're almost never going to need it.

imsodin:
I forgot it in the commit message: Already connected devices are filtered out as well.

AudriusButkevicius:
so looks good to me, someone who actually runs more than one syncthing should test.

AudriusButkevicius:
@st-review lgtm

st-review:

calmh:
@st-review lgtm

st-review:
:ok_hand: Merged as 98cfc204ca366e455b0996096b9f440b5257c089. Thanks, @imsodin!

dinosore:
For me, the list shows an already connected device.  v0.14.32-rc.1, Linux (64 bit).

PR #4187, opened 2017-06-01
cmd/syncthing, lib/logger: Set debug flags on SetDebug instead of main for tests

imsodin:
### Purpose

If you currently run tests with STTRACE set the flags for filenames/precise timestamps/... are not activated, as these get set in cmd/syncthing. Now those flags get set as well when SetDebug is called. This results in flags being set redundantly, if multiple facilities are activated, but I think this is a non-issue and cleaner than e.g. checking for the length of l.debug.

### Testing

Tests still passing and tests now display the additional info.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -269,7 +269,7 @@ func defaultRuntimeOptions() RuntimeOptions {
 	}
 
 	if os.Getenv("STTRACE") != "" {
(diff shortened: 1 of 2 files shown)


calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 719c313b23656a47545e199073db583771a59278. Thanks, @imsodin!

PR #4208, opened 2017-06-11
lib/ignore, lib/model: Use an interface to detect file changes, improving tests

calmh:
### Purpose

This solves the erratic test failures on model.TestIgnores by ensuring that the ignore patterns are reloaded even in the face of unchanged timestamps.

Also updates the constructor to functional options style, as I realized how annoying it was to have to change the calls everywhere when adding an optional parameter, and how nice it would have been not to have to do that to begin with.

Also use a fake clock for the cache test, as that can otherwise fail under load (as shown in one of the failures).

### Testing

This is adequately unit tested I think.

### DIFF
--- a/lib/ignore/cache.go
+++ b/lib/ignore/cache.go
@@ -8,6 +8,12 @@ package ignore
 
 import "time"
 
+type nower interface {
+	Now() time.Time
+}
+
+var clock = nower(defaultClock{})
+
 type cache struct {
 	patterns []Pattern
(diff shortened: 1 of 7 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 68c1a0b9b4066862b03263ebd6584274ebde17fb. Thanks, @calmh!

Issue #109, opened 2014-04-04
Support for file encryption (e.g. non-trusted servers)


Natanji:
So I have had a look at BitTorrent sync, syncthing and alternatives and what I always wondered about was the possibility to not only sync between resources I own and trust, but also external resources/servers which I do NOT trust with my data, up to a certain extent.

One way to do this is using ecryptfs or encfs, but this has many obvious downsides: it is not an interoperable solution (only works on Linux), the files are actually stored in encrypted form on the disk (even if the resource is trusted and this is not necessary, for instance because of the file system being encrypted already), etc.

What I propose is somehow configuring nodes which are only sent the files in an encrypted format, with all file contents (and potentially file/directory names as well; or even permissions) being encrypted. This way, if I want to store my private files on a fast server in a datacenter to access them from anywhere, I could do this with syncthing without essentially giving up ownership of those files. I could also prevent that particular sync node from being allowed/able to make any changes to the files without me noticing.

I realize that this requires a LOT of additional effort, but it would be a killer feature that seems to not be available in any other "private cloud" solution so far. What are your thoughts on this feature?

EDIT: BitTorrent sync mentions a feature like this in their API docs: "Encryption secret
API users can generate folder secrets with encrypted peer support. Encryption secrets are read-only. They make Sync data encrypted on the receiver’s side. Recipients can sync files, but they can’t see file content, and they can’t modify the files. Encryption secrets come in handy if you need to sync to an untrusted location." (from http://www.bittorrent.com/intl/de/sync/developers/api)

vincenzoml:
Whoa, what a long and interesting thread. I've been thinking about zero-knowledge hosting for a long time when a famous service was shut down in new Zealand 2011 if I recall correctly. I even started some prototypes that eventually went dead. But before I explain my own ideas, may I ask for a summary of the thread? Is there a proposed / accepted project for the implementation of this feature?

neolefty:
A tangential feature would be read-only sharing. Non-trusted, but without encryption.

For example, KA Lite (an offline fork of Khan Academy) offers the option to use btsync to download its videos. A core of editors have read-write access, and a much larger group of consumers have read-only access, which is very conveniently supported by btsync. I think KA Lite would love to use an open-source alternative such as Syncthing, but currently Syncthing has 2 problems:

1. No read-only participation, to prevent accidental deletion or corruption by consumers

2. Too much administrative overhead for mass sharing of files (no auto-peering) -- a separate issue!

[...]

Issue #24, opened 2014-01-08
Add -v --version options


jpjp:
I can't seem to find where the version number of syncthing is stored in the source, so didn't attempt a patch.

Would be useful to have an option to view the version from the binary.

Issue #23, opened 2014-01-07
Max out TLS security


jpjp:
Since the cryptography in syncthing is entirely self-enclosed and you don't need to worry about breaking external stuff, there is a good opportunity to make TLS as secure as possible.

The 30c3 talk about better cryptography might be interesting https://bettercrypto.org/

PR #4210, opened 2017-06-13
At rest file encryption

nrm21:
### Purpose

At rest file encryption.  Shouldn't need explanation why, its the most requested feature we get pretty much. 
 This is just a preliminary alpha to receive comments (and if necessary help).  It's very much not ready for PR yet.

Fulfills most of #109 I think
Also semi-related to this forum thread: https://forum.syncthing.net/t/encrypted-nodes-please-test/5295/42

### Testing

Ok so far it works like this (and I realize this is not in some finished usable state yet, it's mostly PoC for now):
* You setup a directory that you leave unsynced w/ anyone (let's call it 'Sync')
* You put files and folders in it, and wait for a scan
* After scan it will create a Sync.ENC dir in a lateral dir to it in the same root (that is the dir you setup sync pair with your untrusted client)
* It will place a file in that dir for every file you have in your 'Sync' dir (it will traverse sub folders since it's using the standard scanner but ignore creation of encrypted objects for folders, yet it WILL save the full file path inside the encrypted portion of the file)
* All those files will be placed in the Sync.ENC dir at top level, there will be no depth and the filename of each will be some fraction (prob first quarter) of the sha256 hash of file.
* Decryption happens pretty much on the very next scan of that Sync.ENC folder, where it tries to recreate the path back in the non .ENC folder (at least that's the idea, I haven't yet gotten it working for multiple files at a time yet).
* On decryption, if file exists it should ignore and not rewrite back to disk

### Notes
* Filenames can be changed for now and it should have no bearing on the decryption capability since all relevant info needed for it is stored inside the file (possibly subject to change).
* Determination to encrypt/decrypt is made based on location of file (is it in rootdir or rootdir.ENC) AND existence of the file sig (aka magic numbers) we attach to it after encryption.
* Puts file MAC times in the file but currently ignore this on decrypt (possibly we do something in future with this)
* Currently the engine decrypts a file and finds out of it is needed or the file already exists.  We might need to explore ways to decide if we need a file enc/dec operation or not BEFORE we perform it (hash comparisons here might be better)
* Giving strong consideration to JUST doing encryption and putting the decryption code in a separate executable for now as a first round.  The logic here is that it avoids quite a few problems and MOST of the time people probably just want an encrypted backup of their stuff on an untrusted client (that's always updated older file version with newer file versions/modification) and the ONE time they need backup from catastrophe, they can use the decryption executable

Here's a general idea of encrypted file structure (picture since I can't get this post to format as fixed length font):
http://imgur.com/a2V8In3

### Authorship

nrm21

### DIFF
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -26,6 +26,7 @@ type FolderConfiguration struct {
 	RescanIntervalS       int                         `xml:"rescanIntervalS,attr" json:"rescanIntervalS"`
 	IgnorePerms           bool                        `xml:"ignorePerms,attr" json:"ignorePerms"`
 	AutoNormalize         bool                        `xml:"autoNormalize,attr" json:"autoNormalize"`
+	EncryptionKey         string                      `xml:"encryptionKey" json:"encryptionKey"`
 	MinDiskFree           Size                        `xml:"minDiskFree" json:"minDiskFree"`
 	Versioning            VersioningConfiguration     `xml:"versioning" json:"versioning"`
 	Copiers               int                         `xml:"copiers" json:"copiers"` // This defines how many files are handled concurrently.
--- a/lib/config/foldertype.go
+++ b/lib/config/foldertype.go
@@ -11,6 +11,7 @@ type FolderType int
 const (
 	FolderTypeSendReceive FolderType = iota // default is sendreceive
 	FolderTypeSendOnly
+	FolderTypeEncrypted
 )
 
 func (t FolderType) String() string {
@@ -19,6 +20,8 @@ func (t FolderType) String() string {
 		return "readwrite"
 	case FolderTypeSendOnly:
 		return "readonly"
+	case FolderTypeEncrypted:
+		return "encrypted"
 	default:
 		return "unknown"
 	}
@@ -34,6 +37,8 @@ func (t *FolderType) UnmarshalText(bs []byte) error {
 		*t = FolderTypeSendReceive
 	case "readonly", "sendonly":
 		*t = FolderTypeSendOnly
+	case "encrypted":
+		*t = FolderTypeEncrypted
 	default:
 		*t = FolderTypeSendReceive
 	}
--- a/lib/model/fileencryption.go
+++ b/lib/model/fileencryption.go
@@ -1,0 +1,181 @@
+package model
+
+import (
+	"bytes"
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/sha256"
+	"encoding/binary"
+	"encoding/hex"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/syncthing/syncthing/lib/protocol"
+)
+
+// This function is used to encrypt or decrypt the file based on existance of the right file signature
+func EncOrDecFiles(rootDir string, files []protocol.FileInfo) {
+	for _, file := range files {
+		if !file.IsDeleted() && !file.IsDirectory() { // non deleted files only
+			l.Infoln("file:", file.Name, file.IsDeleted())
+
+			if file.Type == 0 || file.Type == 4 { // files and symlinks only (ignore dirs and all the other bs for now)
+				// some key for testing (32 bytes exactly, otherwise we need to truncate/pad to length)
+				key := checkKeyLength([]byte("ThisiswaytoolongakeyforSyncthing"))
+				aesblock, err := aes.NewCipher(key)
+				if err != nil {
+					l.Infoln(err)
+				}
+				pathToFile := rootDir + string(os.PathSeparator) + filepath.FromSlash(file.Name)
+
+				plainOrCiphertext, err := ioutil.ReadFile(pathToFile)
+				if err != nil {
+					l.Infoln(err)
+				}
+
+				// if file sig exists we should decrypt instead
+				if string(plainOrCiphertext[:5]) == "stenc" {
+					if !strings.HasSuffix(rootDir, ".ENC") { // make sure we're not running decryption dir created to hold the enc files
+						decryptFile(rootDir, pathToFile, key, aesblock, plainOrCiphertext)
+					}
+				} else {
+					encryptFile(rootDir, pathToFile, key, aesblock, plainOrCiphertext)
+				}
+			}
+		}
+	}
+}
+
+func decryptFile(rootDir string, pathToFile string, key []byte, aesblock cipher.Block, ciphertext []byte) {
+	ciphertext = ciphertext[5:] // strip file sig
+
+	iv := ciphertext[:16]             // extract IV from next 16 bytes
+	ciphertext = ciphertext[len(iv):] // and strip it
+
+	plaintext := make([]byte, len(ciphertext))
+	stream := cipher.NewCFBDecrypter(aesblock, iv)
+	stream.XORKeyStream(plaintext, ciphertext)
+
+	filepathNull := bytes.IndexByte(plaintext, 0) // find first null char (end of filename)
+	if filepathNull == -1 {
+		l.Infoln("Cannot find null byte, we may have the wrong key or a corrupt file")
+	}
+	newfilepath := string(plaintext[:filepathNull]) // extract filepath
+	newfilepath = filepath.Base(newfilepath)
+	plaintext = plaintext[filepathNull+1:] // remove filepath from file contents
+	plaintext = plaintext[32:]             // strip off sha256 plaintext hash (since we're not using them right now)
+	plaintext = plaintext[(8 * 3):]        // strip off file access times (since we're not using them right now)
+
+	f, err := os.Create(rootDir + string(os.PathSeparator) + newfilepath)
+	if err != nil {
+		l.Infoln(err)
+	}
+	_, err = io.Copy(f, bytes.NewReader(plaintext))
+	if err != nil {
+		l.Infoln(err)
+	}
+}
+
+func encryptFile(rootDir string, pathToFile string, key []byte, aesblock cipher.Block, plaintext []byte) {
+	// Get file MAC times and make byte arrays out of them
+	mtime, atime, ctime, err := fileStatTimes(pathToFile)
+	bmtime, batime, bctime := make([]byte, 8), make([]byte, 8), make([]byte, 8)
+	binary.LittleEndian.PutUint64(bmtime, uint64(mtime))
+	binary.LittleEndian.PutUint64(batime, uint64(atime))
+	binary.LittleEndian.PutUint64(bctime, uint64(ctime))
+	// To revert to int again ==>     i = int64(binary.LittleEndian.Uint64(b))
+
+	// The IV needs to be unique but not secure, so put it at beginning of ciphertext
+	h := sha256.New()
+	h.Write([]byte(pathToFile))
+	var iv []byte
+	iv = append(iv, h.Sum(nil)...)
+
+	if _, err := io.Copy(h, bytes.NewReader(plaintext)); err != nil {
+		log.Fatal(err)
+	}
+	sha256HashBytes := h.Sum(nil) // sha256 of just the filedata by itself
+
+	var fileHeader []byte
+	fileHeader = append([]byte(pathToFile), 0)          // null byte terminated
+	fileHeader = append(fileHeader, sha256HashBytes...) // plaintext hash
+	fileHeader = append(fileHeader, bmtime...)          // mod time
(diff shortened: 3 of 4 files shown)


AudriusButkevicius:
I think this is orders of magnitude worse than just using encfs, what's the point?

calmh:
Well, I wouldn't put it like that. But I don't think it's a desirable property that each device carrying the unencrypted data must also carry the encrypted data. It's implementation convenient and avoids some problems that are otherwise harder to solve, but it's not convenient from a usage point of view.

Ferroin:
Having an encrypted backup on an untrusted node can't be done with encfs (or eCryptFS, or whatever else you decide to use that's a filesystem layer) because:
1. Running encfs on the untrusted node exposes unencrypted data there.
2. For at least encfs (and I think eCryptFS too) you can't sync the back-end storage in real-time, even pulling data off of it, both because of how FUSE works, and because you're highly likely to get partial data.

nrm21:
> I think this is orders of magnitude worse than just using encfs, what's the point?
Tell us how you really feel LOL.

> But I don't think it's a desirable property that each device carrying the unencrypted data must also carry the encrypted data. It's implementation convenient and avoids some problems that are otherwise harder to solve, but it's not convenient from a usage point of view.
Agreed and that is why I did it.  My hope is to change that eventually but looking for a more iterative approach to start.  Something that requires the least amount of change to codebase to start and go from there.

nrm21:
Not to be negative but I, like most, been following the whole debate on this file encryption at rest thing off and on for a few years now and I feel like the demand for complete refactorization and redesign to make it the "right way" (in quotes because that's a highly subjective thing, maybe not from a security PoV from a usability PoV), is a daunting hurdle and powerful disincentive for anyone to start on this.

I favor a more iterative 'agile' approach.  Making the least amount of changes needed to get something working and secure.  Fewer changes = fewer chances for bugs and other issues.

Perhaps I'm wrong, but hopefully we can get the discussion going again.

AudriusButkevicius:
Sorry, my reply is harsh. This is cool and all but its fundamentally wrong design wise.

If a stunt like this is to be done in syncthing, it should done at the filesystem access time, and not semi randomly on scans consuming double the storage and forcing you to rescan to decrypt/encrypt your files...

I am not sure how all the <flavor of the month> fuse encryption thingies work and what they provide, but I'd hope that there exists one that encrypts/decrypts things at access time on the fly filling the gap this PR is trying to fill.

nrm21:
Lol that's ok I'm a big boy, I can take criticism.  ;)
However you are right to a large degree, it is somewhat ugly to circumvent certain hard challenges that would require rewrite/refactor large parts of the codebase.  I put up this PR for a few reasons:

1) I kinda wanted others to look at this and give their thoughts and possibly direction on where to go with this, because I'm not exactly sure.

2) This is just quick and dirty code and I'm not opposed to complete rewire of this, to the best of my ability.  But rather than writing tons of code only to have it rejected... better write as little as possible, have it rejected and maybe a "but if you wrote it like this... maybe we would take it".

3) I definitely don't expect this PR to be accepted (I wrote that at time top) just more of a restarting of the conversation.  I could have just recreated a subject but I feel without some starter code it would be another guy bloviating on "Hey people lets start on file encryption right?" and everyone going "YEAH!" and then everyone promptly doing nothing about it... yet again.  :P

Ferroin:
@AudriusButkevicius Unless you're talking about pulling the implementation out of EncFS or some other FUSE-based encrypting filesystem, it's not an option for one of the primary use cases for this.

That said, EncFS does have the type of on-access {de,en}cryption that you're talking about, as does eCryptFS (to a lesser degree, that's an odd chimera with both a userspace FUSE component and a kernel-side component), but they use _a lot_ of memory for what they do.  If Syncthing takes that approach, it does have some negative aspects:
1. Each time a file changes, the entire file will need to be re-encrypted so that the hashes can be checked/recalculated.  Without this, the untrusted node will need to temporarily decrypt blocks as they come in to verify the hashes, which is not good from a security POV.
2. A decision will need to be made as to whether we value low CPU usage or low memory usage more.  In the CPU case, we then need to cache encrypted blocks in memory, in the memory case we need to encrypt blocks every time they are requested (I would personally advocate having this be a tunable, small networks without many devices would be better served by not caching, while large ones with lots of devices will likely require it for any kind of scalability in terms of performance).  In essence, the current implementation in the PR values both above disk space, as it's essentially caching everything on disk.

nrm21:
More or less.  However the current implementation does require a complete reencryption of the file anytime there is a change, but it's almost unavoidable to do this I think.

Otherwise we could encrypt each block seperately but I'm not sure what this would gain us.  Say you made a change to a file near the beginning.  Wouldn't the resulting bit shifting for all the rest of the file STILL cause a rewrite of all the blocks.  All that extra complication for what?  Good file security does not lend itself well to dedup and other cool tricks I think.

MyPod:
About the complete re-encryption issue, https://github.com/syncthing/syncthing/pull/3527 would lead me to believe it doesn't happen too often, so it shouldn't be a big deal.

Ferroin:
If we're only storing hashes for the encrypted copy of each block, we will always have to re-encrypt completely, because we have no way to know where in the file anything changed.

To avoid the requirement to always re-encrypt, Syncthing needs to store 2 hashes for each block, one for the unencrypted copy, and one for the encrypted copy.  Doing so would reduce the overhead in the case where a block hasn't changed, but increase it in the case where a block has changed, and would also increase the space requirements in the database.

nrm21:
In the interest of simplicity, this implementation is not storing ANY block hashes... only hashes for the entire file.

I chose not to do anything at the block level to avoid having to change much of anything in the underlying scanner code.  This makes compatibility with 'encryption agnostic' clients easier and puts separation between the code layer that encrypts and the one that distributes files to other clients.  That separation makes troubleshooting where a problem could be much easier going into the future.

The most obvious downside being, we need to store everything twice on trusted clients.  A pretty big downside but not the end of the world if you're dealing with small files that are sensitive (doc, jpg, png, xls, gsheet, etc).

calmh:
So what this has going for it is that it's almost trivially simple. A `+244` diff that implements encrypted folders is pretty damn cool. Nonetheless, requiring an encrypted extra copy on disk is impractical and not, in my opinion, the way to go about it.

I think the correct solution is protocol level support for encrypted index entries and encrypted blocks, where a device knows that's it is not trusted with the real data and can act accordingly. I have several half hearted designs in my head, we might get there some day if noone else does it first. Regardless, this is not the way, sorry.

nrm21:
@calmh Ok fair enough.  So translating this into psudocode in my mind it sounds like you are saying (please correct if I'm wrong):

- Just before sending on the wire a requested block... Encrypt each 128k block (treat each file chunk of ~128k independent of any others)... then sent directly to untrusted client and never stored on our (trusted client) disk

- Once received a block from untrusted client... decrypt before writing to disk

It's implied that this system would require a new mechanism like FolderConfig[folder].Encrypted == true to differentiate how clients treat blocks from these from normal folders (which should have no change in behavior from how it is today).

nrm21:
Probably (and maybe I'm overthinking this), we would need to attach IV, unencrypted data hash and encrypted data hash to the whole block which would make the size bigger that 128k... woudln't that break things?  Aren't there things that expect blocks to be that exact size or smaller, or am I wrong?

calmh:
Mmhm. I'm envisioning several new message types in the protocol. Consider that I, in the config, can set a preshared key for a given device and folder. For that folder, we then no longer send `IndexUpdate`s with `FileInfo`s in them. Instead, we send `EncryptedIndexUpdate`s with `EncryptedFileInfo`s:

```protobuf
message EncryptedIndexUpdate {
    string                     folder = 1;
    repeated EncryptedFileInfo files  = 2 [(gogoproto.nullable) = false];
}

message EncryptedFileInfo {
    string identifier         = 1; // "file name" for this encrypted instance of the file, something like the sha256 of the original filename plus the shared key or similar
    bytes  encrypted_fileinfo = 2; // the original protobuf FileInfo, encrypted
    bytes  nonce              = 3; // the nonce or IV that goes with it
    int32  num_blocks         = 4; // how many blocks the file contains, so the other side can know how much data to pull (roughly)
    Vector version            = 5; // original version vector, because they need to do conflict resolution etc
    int64  sequence           = 6; // as above
}
```

In our outgoing index sender we "simply" know that the other side is supposed to receive encrypted data and encrypt the file infos on the fly.

The other side receives these and stores them as is. It knows that it doesn't have the full data, but it has an encrypted FileInfo it can pass on to others, it has an `identifier` it can use to store things on disk, and it has the `version` etc. It can pull all the required blocks of the file by using the new `EncryptedRequest` method to get `EncryptedResponse`s:

```protobuf
message EncryptedRequest {
    int32  id           = 1;
    string folder       = 2;
    string identifier   = 3;
    int32  block_number = 4;
}

message EncryptedResponse {
    int32     id    = 1;
    bytes     data  = 2; // an encrypted 128k block
    bytes     nonce = 3; // the nonce or iv that goes with it
    ErrorCode code  = 4;
}
```

On the sending side it just reads the corresponding cleartext block, encrypts it and sends it. On the receiving side it stores the block and corresponding nonce/iv. The last block is padded to full size.

Devices that have the key use these same methods to request data from encrypted devices, where the other side just reads it verbatim from disk. It can then decrypt the data using the key it has and save it in cleartext. It knows the padding etc for the last block either because it's obvious from the decrypted data or by looking in the decrypted fileinfo.

Simple, no? I'm sure there's a hundred details I haven't though of that will bite you during implementation, of course.

nrm21:
Some thoughts (prob related to your last sentence):
* Why do we need an identifier even (unless I misunderstand the point)?  Can we just use block number as id?  This is block #23, this is block #24, this is....etc.  What does the untrusted node care since he can't read it anyway?

* Receiving just ciphertext and nonce sure does sound simple but is it?  What about mod times, permissions and all that?  Or is that what the identifier is for (maybe that's why I misunderstand it as just a simple "I'm THIS particular block" field).  I assume we're NOT putting data we need to read back on decryption in the identifier field since you mentioned sha256 hash.  Which I translate to... "one-way irreversible loss of data we can never recover" function. :)

* Someone with thousands of small files in a folder would find this grossly inefficient since "The last block is padded to full size" would mean that all those "one block" files would become ~128k on disk.  Prob not a deal breaker and the price to pay for security I guess, but does the padding need to be that much?  It brings up the larger question really.  How much padding of anything is enough.  I feel like even slightly changing the file size is enough to mask most files pretty well.  If you go back and look at the structure of my file in this PR, you will see that near the beginning that filename field was variable (w/ null byte to demarc)  That variability was on purpose to make those without decryption capabilities (but forensically looking at the file) have difficulty knowing where one field ends and another begins (not so hard for them if all the fields are a set length in size).  I understand part of this is cause protobufs wants "descriptive-ness" of the fields.  But isn't there some exception for a variable field in protobufs?  Maybe we could look into incorporating something like that to up security.

nrm21:
One last thing not mentioned here: Say you have 3 or more nodes in an encrypted folder pair (2 are trusted).  Should we send the connection between the two trusted to each other as a normal connection, or just encrypt and send knowing the other guy can decrypt?  Honestly I think that if we did the latter it would be simpler to code since we're not behaving 2 different ways for 2 different client connections.  Thought it might make a connection between those 2 slower than it would have been.  Oh well, people can always choose to not use the feature between trusted hosts if they want.

nrm21:
Also do you mind if I add your structs to:
https://github.com/syncthing/syncthing/wiki/Crypto-Proposal

nrm21:
*After looking at docs again*

Oh I see... the encrypted fileinfo has all that modtime and file perms inside it.  Got it.

calmh:
So yeah the identifier is in place of the file name, which we are not telling the untrusted device. We still need something so that we can refer to the same file.

So yeah the padding might not need to be to a full block. Given that the untrusted device isn't told the actual size and just grabs the number of blocks we tell it to, it's up to the trusted device how much data to return for the last block. It could return a full block, padded to 64 bytes boundary or something random. When a trusted device is downloading from an untrusted device it has the FileInfo and knows the exact length of the block and file.

I think the encryption should be per device and folder. That is, if you have two trusted devices A and B and an untrusted device U:

- A and B add each other normally, like today.
- A and B adds U as untrusted and sets a key for the connection+folder, which must be the same on both A and B

Communication between A and B is normal. Communication between A and U or B and U is with the encrypted data.

An additional untrusted device U2 could be added and joined to U in the normal manner. U wouldn't see any difference between U2 and A or B - the fact that A and B can decrypt the data only matters locally.

Syncthing should be smart enough to just store the encrypted data and index entries when it receives them, and pass them on like any other data. So an untrusted device doesn't need to know that it's untrusted.

calmh:
Another thing to keep in mind here is that any file change (even metadata, unless we do something smart) means a full download for every untrusted device.

PR #4215, opened 2017-06-15
Set the secure flag to true for the session cookie

dojodominator:
If the HTTP proto used is HTTPS, then the secure flag must be set on the session cookie.  Not using secure could allow the cookie to be sent over an unencrypted connection which would allow a malicious actor to steal the session information.

This also removes a trivial memory allocation in the session map.

### Purpose

Fixes a potential session leak vulnerability by enabling the secure flag for session cookies over HTTPS connections or when behind a reverse proxy that is presenting HTTPS connections to the internet.

### Testing

Tested behind an HTTP reverse proxy that sets the "X-FORWARDED-PROTO" header to "https".  To test, run syncthing with HTTPS enabled or behind an HTTPS enabled reverse proxy that uses one of the supported headers.

### Authorship

### DIFF
--- a/cmd/syncthing/gui_auth.go
+++ b/cmd/syncthing/gui_auth.go
@@ -21,7 +21,7 @@ import (
 )
 
 var (
-	sessions    = make(map[string]bool)
+	sessions    = make(map[string]struct{})
 	sessionsMut = sync.NewMutex()
 )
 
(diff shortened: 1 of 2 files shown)


calmh:
Hi! This seems reasonable, but doesn't seem to do what it says when I test it:

```
jb@unu:~ $ curl -ski https://testuser:testpass@localhost:8081/ | head -10
HTTP/1.1 200 OK
Content-Length: 47188
Content-Type: text/html
Set-Cookie: sessionid-I6KAH=ZCeU9s37HdNfLydfFQvp2Qx2SeFxMEhj
Set-Cookie: CSRF-Token-I6KAH=SQwd7exrSwLrMJtb7jkz7yz22jXq6VNh
X-Syncthing-Id: I6KAH76-66SLLLB-5PFXSOA-UFJCDZC-YAOMLEK-CP2GB32-BV5RQST-3PSROAU
X-Syncthing-Version: v0.14.31-rc.1+3-g5bde8283-dirty-pr-4215
Date: Fri, 16 Jun 2017 05:51:54 GMT

<!DOCTYPE html>
```

Perhaps a test could be added to gui_test.go to verify this?

calmh:
That said, I'm unsure how much this helps. We use basic authentication. The cookie is session scoped. If the browser is still in the same session (so the cookie is still there) and gets redirected to HTTP, Syncthing will do a basic auth challenge. The browser will respond to this and send the username and password in clear text instead.

calmh:
Closing for lack of activity, feel free to reactivate whenever

PR #4220, opened 2017-06-17
lib/config: Ignored folders that are added to the config should not remain ignored (fixes #4219)

### LINKED ISSUE #4219: Ignored folder share attempts stick around for too long
If a folder is ignored, it then doesn't work to add and share it manually. You can do all the steps in the GUI and config, but the folder still gets ignored in the clusterconfig handling and we don't start the appropriate routines for it.

calmh:
### Purpose

Manually adding a folder that was ignored should work.

### Testing

Unit test.

### DIFF
--- a/lib/config/config.go
+++ b/lib/config/config.go
@@ -268,6 +268,14 @@ func (cfg *Configuration) clean() error {
 		seenFolders[folder.ID] = struct{}{}
 	}
 
+	// Remove ignored folders that are anyway part of the configuration.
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 8f214fe4a95184ad2f011559f3a1ae808ba91640. Thanks, @calmh!

adamryczkowski:
Thank you very much!

PR #4226, opened 2017-06-25
lib/upgrade: 0.x to 1.0 is a minor upgrade

calmh:
This removes the special handling of minor versions as major when the
actual major is zero, and adds the special case that upgrades from 0.x
to 1.x are considered minor. 0.x to 2.x or 1.x to 2.x etc are still
considered major.


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as db1dc9985aea5407663bff354b3bb83bb4ad26af. Thanks, @calmh!

PR #4228, opened 2017-06-25
Convert folders to use filesystem abstraction

AudriusButkevicius:
It builds and starts (to some extent)

Haven't even started looking at the tests.
Probably makes sense to look through package by package.

A few notable things:

1. BasicFilesystem.rooted is the old rootPathJoined, yet it now permits accessing root of the folder for stuff like mkdir('.')
2. locations in UI could also be converted to use filesystems
3. staggered versioning now forced at .stversions

todo: 

- [x] undo staggered versioning path selection
- [x] convert external versioner to use templating
- [ ] update external versioner docs
- [x] update external versioner UI
- [x] add one off ack warning for staggered versioning changes
- [x] potentially allow people escaping .stversions via symlink in basic filesystem? (already allowed it seems...)
- [X] fix tests on windows
- [X] fix tests on linux/mac


lib/fs/filesystem.go
```diff
@@ -76,3 +107,9 @@ var IsExist = os.IsExist
 
 // IsNotExist is the equivalent of os.IsNotExist
 var IsNotExist = os.IsNotExist
+
+// IsPermission is the equivalent of os.IsPermission
+var IsPermission = os.IsPermission
+
+// IsPathSeparator is the equivalent of os.IsPathSeparator
+var IsPathSeparator = os.IsPathSeparator
```
AudriusButkevicius: IsPathSeparator is a function

lib/fs/filesystem.go
```diff
@@ -22,23 +22,39 @@ type Filesystem interface {
...
 	Rename(oldname, newname string) error
 	Stat(name string) (FileInfo, error)
 	SymlinksSupported() bool
 	Walk(root string, walkFn WalkFunc) error
+	Show(name string) error
+	Hide(name string) error
+	Glob(pattern string) ([]string, error)
+	SyncDir(name string) error
+	Roots() ([]string, error)
+	Usage(name string) (Usage, error)
```
AudriusButkevicius: I guess this is clever enough to check usage in a specific directory, but in practice it's never used.

AudriusButkevicius:
So I am not going to start fixing the tests until we settle what the code looks like.
This already took a week to produce, so I am not going to spend another week fixing tests just to be told it's the wrong approach.

cmd/syncthing/gui.go
```diff
@@ -1259,23 +1259,38 @@ func (s *apiService) getPeerCompletion(w http.ResponseWriter, r *http.Request) {
 func (s *apiService) getSystemBrowse(w http.ResponseWriter, r *http.Request) {
 	qs := r.URL.Query()
 	current := qs.Get("current")
+	fsType := qs.Get("filesystem")
+	if fsType == "" {
+		http.Error(w, "filesystem not specified", 500)
```
calmh: Compatibility: default to `"basic"`

calmh:
General observation after a first glance at the first few things: I think you're going overboard with it. The `Filesystem` is for folders. There is no point, imho, to using it for things like csrf cookies. We can assume that we have a regular, file based, filesystem to store our config in.

lib/config/folderconfiguration.go
```diff
@@ -8,19 +8,17 @@ package config
...
-	"github.com/syncthing/syncthing/lib/osutil"
+	"github.com/syncthing/syncthing/lib/fs"
 	"github.com/syncthing/syncthing/lib/protocol"
 )
 
 type FolderConfiguration struct {
 	ID                    string                      `xml:"id,attr" json:"id"`
 	Label                 string                      `xml:"label,attr" json:"label"`
-	RawPath               string                      `xml:"path,attr" json:"path"`
+	FilesystemType        string                      `xml:"filesystemType,attr" json:"filesystemType"`
```
calmh: Minor style nit, maybe these should not be `attr`s as it's getting a bit long, but no strong opinion really.

lib/config/folderconfiguration.go
```diff
@@ -39,12 +37,12 @@ type FolderConfiguration struct {
...
 	DisableSparseFiles    bool                        `xml:"disableSparseFiles" json:"disableSparseFiles"`
 	DisableTempIndexes    bool                        `xml:"disableTempIndexes" json:"disableTempIndexes"`
-	Fsync                 bool                        `xml:"fsync" json:"fsync"`
 	Paused                bool                        `xml:"paused" json:"paused"`
 	WeakHashThresholdPct  int                         `xml:"weakHashThresholdPct" json:"weakHashThresholdPct"` // Use weak hash if more than X percent of the file has changed. Set to -1 to always use weak hash.
 
-	cachedPath string
+	cachedFilesystem fs.Filesystem
 
+	DeprecatedPath           string  `xml:"path,attr,omitempty" json:"-"`
```
calmh: Compatibility: we talked about this, and I can indeed fix it up later if you prefer, but I would aim for `json:"path,omitempty"` and keep that attribute populated as long as the type is `basic`. On input, if `DeprecatedPath` is set and the `Filesystem*` are empty, set the filesystem stuff appropriately.

lib/weakhash/weakhash.go
```diff
@@ -72,27 +71,21 @@ func Find(ir io.Reader, hashesToFind []uint32, size int) (map[uint32][]int64, er
 	return offsets, nil
 }
 
-func NewFinder(path string, size int, hashesToFind []uint32) (*Finder, error) {
-	file, err := os.Open(path)
+func NewFinder(ir io.ReadSeeker, size int, hashesToFind []uint32) (*Finder, error) {
```
calmh: ❤️

lib/versioner/versioner.go
```diff
@@ -8,11 +8,13 @@
...
 package versioner
 
+import "github.com/syncthing/syncthing/lib/fs"
+
 type Versioner interface {
 	Archive(filePath string) error
 }
 
-var Factories = map[string]func(folderID string, folderDir string, params map[string]string) Versioner{}
+var Factories = map[string]func(folderID string, filesystem fs.Filesystem, params map[string]string) Versioner{}
```
calmh: Does this mean the versioning directory needs to live in the same filesystem as the folder, that is, underneath the folder root? I'm not necessarily opposed to that (it's even hardcoded into everything but the staggered thing) but it is a change

lib/versioner/external.go
```diff
@@ -51,16 +50,11 @@ func (v External) Archive(filePath string) error {
...
-	if err != nil {
-		return err
-	}
-
 	if v.command == "" {
 		return errors.New("Versioner: command is empty, please enter a valid command")
 	}
 
-	cmd := exec.Command(v.command, v.folderPath, inFolderPath)
+	cmd := exec.Command(v.command, v.filesystem.Type(), v.filesystem.URI(), filePath)
```
calmh: Compatibility: if the filesystem type is `basic`, call it with the old parameter format?

AudriusButkevicius: We should move to templated commands and as part of config migration prepend a template that defaults to the current args we add to the command.

AudriusButkevicius:
So I don't think it's that of a bad idea to convert everything to use filesystems, as in load my config and certs from encrypted s3,or put audit logs in s3 etc, it was just hard to implement around the locations concept as half of them are files other half folders.

calmh:
Where do you store the config telling you how to access the s3 buckets? The database?

AudriusButkevicius:
No, you specify it on the command line


syncthing -home-fs s3 -home-path user:looser@mybucket.aws.com/syncthing-config

calmh:
No, I meant do you expect to put the database in s3 as well

AudriusButkevicius:
That probably won't go down well, but I don't see why we shouldn't support everything else apart from the database?

The database could also be a case, but not for s3, for something else potentially.

calmh:
I think it increases complexity and diff size for no good purpose. We don't need it now, we don't know if it will work when we do need it - better to do it then instead.

AudriusButkevicius:
Agree, hence all the todos. Or do you want me to unwind some specifics?

calmh:
I'd suggest unwinding, and not sprinkling with comments, all the file i/o that is not folder related (cookies, config, etc.). Ignores and versioning are clearly folder related and must be under the fs umbrella.

I'm not at all opposed to being able to do what you imagine with the config and stuff in remote storage, by the way.

AudriusButkevicius:
Which bits would you unwind?

calmh:
I can point to specifics later, but don't worry too much about it. I think most were pieces that you had already commented as "convert?" where I think the answer is "no, not right now at least". I'm currently on vacation (-ish...) so sorry fo rnot giving you more review time, I've yet to look at the complex packages but I suspect it's probably mostly fine...

calmh:
Model stuff looks good as well. I think nothing here is really controversial, once we've sorted out what goes into a filesystem and what is just local unabstracted i/o. Scary part is getting the rooted path stuff right and tested so there are no escapes or screwups? That will require some testing I guess, but we also have our beta testers...

AudriusButkevicius:
So I didn't have time to look at this on holiday. I'll try to do it in the evenings.

lib/fs/basicfs_unix.go
```diff
@@ -0,0 +1,91 @@
...
+func (f *BasicFilesystem) MkdirAll(name string, perm FileMode) error {
+	name, err := f.rooted(name)
+	if err != nil {
+		return err
+	}
+	return os.MkdirAll(name, os.FileMode(perm))
+}
+
+func (f *BasicFilesystem) Show(name string) error {
+	_, err := f.rooted(name)
```
AudriusButkevicius: Hiding stuff on linux is noop (apart from trying to escape the root, in which case it errors).
Only way to hide is to prefix it with a dot (which changes the name so we don't want to do that).

AudriusButkevicius:
so made some progress, pls skim through see if you are happy with things.

AudriusButkevicius:

AudriusButkevicius:
So tests pass for me on windows and linux:
```
root@you:~/st# ./build.sh test
ok      github.com/syncthing/syncthing/lib/auto 1.239s
ok      github.com/syncthing/syncthing/lib/beacon       1.008s
ok      github.com/syncthing/syncthing/lib/config       1.179s
ok      github.com/syncthing/syncthing/lib/connections  1.016s
ok      github.com/syncthing/syncthing/lib/db   1.544s
ok      github.com/syncthing/syncthing/lib/dialer       1.031s
ok      github.com/syncthing/syncthing/lib/discover     2.501s
ok      github.com/syncthing/syncthing/lib/events       6.484s
ok      github.com/syncthing/syncthing/lib/fs   1.015s
ok      github.com/syncthing/syncthing/lib/ignore       1.046s
ok      github.com/syncthing/syncthing/lib/logger       1.017s
ok      github.com/syncthing/syncthing/lib/model        22.142s
ok      github.com/syncthing/syncthing/lib/nat  1.033s
ok      github.com/syncthing/syncthing/lib/osutil       1.040s
ok      github.com/syncthing/syncthing/lib/pmp  1.015s
ok      github.com/syncthing/syncthing/lib/protocol     1.462s
ok      github.com/syncthing/syncthing/lib/rand 6.366s
?       github.com/syncthing/syncthing/lib/rc   [no test files]
ok      github.com/syncthing/syncthing/lib/relay/client 1.017s
ok      github.com/syncthing/syncthing/lib/relay/protocol       1.009s
ok      github.com/syncthing/syncthing/lib/scanner      8.015s
?       github.com/syncthing/syncthing/lib/sha256       [no test files]
ok      github.com/syncthing/syncthing/lib/signature    1.468s
ok      github.com/syncthing/syncthing/lib/stats        1.013s
ok      github.com/syncthing/syncthing/lib/sync 2.051s
ok      github.com/syncthing/syncthing/lib/tlsutil      1.019s
ok      github.com/syncthing/syncthing/lib/upgrade      1.014s
ok      github.com/syncthing/syncthing/lib/upnp 1.017s
ok      github.com/syncthing/syncthing/lib/util 1.008s
ok      github.com/syncthing/syncthing/lib/versioner    1.029s
ok      github.com/syncthing/syncthing/lib/weakhash     1.014s
?       github.com/syncthing/syncthing/cmd/stbench      [no test files]
?       github.com/syncthing/syncthing/cmd/stcli        [no test files]
?       github.com/syncthing/syncthing/cmd/stcompdirs   [no test files]
?       github.com/syncthing/syncthing/cmd/stdisco      [no test files]
?       github.com/syncthing/syncthing/cmd/stdiscosrv   [no test files]
?       github.com/syncthing/syncthing/cmd/stevents     [no test files]
?       github.com/syncthing/syncthing/cmd/stfileinfo   [no test files]
?       github.com/syncthing/syncthing/cmd/stfinddevice [no test files]
?       github.com/syncthing/syncthing/cmd/stgenfiles   [no test files]
?       github.com/syncthing/syncthing/cmd/stindex      [no test files]
?       github.com/syncthing/syncthing/cmd/strelaypoolsrv       [no test files]
?       github.com/syncthing/syncthing/cmd/strelaypoolsrv/auto  [no test files]
?       github.com/syncthing/syncthing/cmd/strelaysrv   [no test files]
?       github.com/syncthing/syncthing/cmd/strelaysrv/testutil  [no test files]
?       github.com/syncthing/syncthing/cmd/stsigtool    [no test files]
?       github.com/syncthing/syncthing/cmd/stvanity     [no test files]
?       github.com/syncthing/syncthing/cmd/stwatchfile  [no test files]
ok      github.com/syncthing/syncthing/cmd/syncthing    18.382s
```

Seem to be hitting some flaky test (that I have written)/speed of the machines.

Probably ready for a second round of reviews.

AudriusButkevicius:

lib/fs/basicfs.go
```diff
@@ -105,6 +317,14 @@ func (f fsFile) Stat() (FileInfo, error) {
 	return fsFileInfo{info}, nil
 }
 
+func (f fsFile) Sync() error {
+	err := f.File.Sync()
+	if err != nil && !strings.Contains(err.Error(), "handle is invalid") {
```
calmh: What's the magic handle is invalid thing here?

lib/fs/filesystem.go
```diff
@@ -22,23 +22,38 @@ type Filesystem interface {
...
+	URI() string
 }
 
 // The File interface abstracts access to a regular file, being a somewhat
 // smaller interface than os.File
 type File interface {
 	io.Reader
+	io.Writer
+	io.ReaderAt
+	io.Seeker
```
calmh: ObOCD: maybe sort the interface embeddings into some sort of order

lib/fs/filesystem.go
```diff
@@ -22,23 +22,38 @@ type Filesystem interface {
 	DirNames(name string) ([]string, error)
 	Lstat(name string) (FileInfo, error)
 	Mkdir(name string, perm FileMode) error
+	MkdirAll(name string, perm FileMode) error
```
calmh: Could we reduce the API surface by having just one of mkdir/mkdirall (preferably the latter)?

lib/fs/filesystem.go
```diff
@@ -22,23 +22,38 @@ type Filesystem interface {
...
 	Open(name string) (File, error)
+	OpenFile(name string, flags int, mode FileMode) (File, error)
 	ReadSymlink(name string) (string, error)
 	Remove(name string) error
+	RemoveAll(name string) error
 	Rename(oldname, newname string) error
 	Stat(name string) (FileInfo, error)
 	SymlinksSupported() bool
 	Walk(root string, walkFn WalkFunc) error
+	Show(name string) error
```
calmh: The name `Show()` makes perfect sense in context, but is somewhat unclear out of context - does it pop up a GUI preview of the file maybe? I wonder if `Unhide()` would be better, even if it's a double negative?

lib/fs/filesystem.go
```diff
@@ -59,12 +74,27 @@ type FileInfo interface {
...
+// Usage represents filesystem space usage
+type Usage struct {
+	Free  int64
+	Total int64
+}
 
-// DefaultFilesystem is the fallback to use when nothing explicitly has
-// been passed.
-var DefaultFilesystem Filesystem = NewWalkFilesystem(NewBasicFilesystem())
+// Variable equivalents from os package.
```
calmh: Comment nitpick: none of these things are variable 👨‍🏫

lib/fs/registry.go
```diff
@@ -0,0 +1,41 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package fs
+
+import "errors"
+
+func init() {
+	registry[FilesystemTypeBasic] = func(root string) Filesystem {
```
calmh: I've become somewhat skeptical of the magic registry kind of global state. I wonder if it wouldn't be cleaner to skip the registry thing and just have `NewFilesystem` explicitly know about the filesystem types we have and just `switch{}` on the `FilesystemType`

cmd/syncthing/gui.go
```diff
@@ -1259,23 +1259,35 @@ func (s *apiService) getPeerCompletion(w http.ResponseWriter, r *http.Request) {
 func (s *apiService) getSystemBrowse(w http.ResponseWriter, r *http.Request) {
 	qs := r.URL.Query()
 	current := qs.Get("current")
+	var fsType fs.FilesystemType
+	fsType.UnmarshalText([]byte(qs.Get("filesystem")))
```
calmh: This is fine as we want to fallback to `basic`, which is the default, and we can ignore the error. But I had to go look at the implementation of `FilesystemType` to figure that out so it would be awesome to have a comment here that this is how the defaulting is handled

lib/model/model.go
```diff
@@ -271,7 +269,12 @@ func (m *Model) warnAboutOverwritingProtectedFiles(folder string) {
 		return
 	}
 
-	folderLocation := m.folderCfgs[folder].Path()
+	// This is a bit of a hack.
```
calmh: :D

lib/model/rwfolder.go
```diff
@@ -1290,11 +1239,8 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
 
 			if !found {
 				found = f.model.finder.Iterate(folders, block.Hash, func(folder, file string, index int32) bool {
-					inFile, err := rootedJoinedPath(folderRoots[folder], file)
-					if err != nil {
-						return false
-					}
-					fd, err := os.Open(inFile)
+					fs := folderFilesystems[folder]
+					fd, err := fs.Open(file)
```
calmh: The parameter being named `file` confused the fuck out of me when reading the diff as there is now a `file` declared at the function scope which is of a completely different type. Rename one?

lib/model/rwfolder.go
```diff
@@ -1345,7 +1291,12 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
 				state.copyDone(block)
 			}
 		}
-		weakHashFinder.Close()
```
calmh: Does the `weakHashFinder` no longer require closing? Can we drop in the file close with a defer at the point we do the open instead of in the nil check here?

lib/model/sharedpullerstate.go
```diff
@@ -292,9 +293,12 @@ func (s *sharedPullerState) finalClose() (bool, error) {
 	}
 
 	if s.fd != nil {
+		// This is our errors if we weren't errored before. Otherwise we
```
calmh: Spurious sudden plural `errors` in comment

lib/osutil/atomic.go
```diff
@@ -25,26 +25,36 @@ var (
 // returned on Close, so a lazy user can ignore errors until Close.
 type AtomicWriter struct {
```
calmh: Not necessarily part of this PR but we should probably consider renaming this as there is no guarantee whatsoever that this operation is "atomic" after involving the `Filesystem`. Maybe it should become a WithTempFileWriter or some crap in the future.

lib/versioner/staggered.go
```diff
@@ -48,23 +48,12 @@ func NewStaggered(folderID, folderPath string, params map[string]string) Version
...
-		versionsDir = filepath.Join(folderPath, ".stversions")
-		l.Debugln("using default dir .stversions")
-	} else if filepath.IsAbs(params["versionsPath"]) {
-		l.Debugln("using dir", params["versionsPath"])
-		versionsDir = params["versionsPath"]
-	} else {
-		versionsDir = filepath.Join(folderPath, params["versionsPath"])
-		l.Debugln("using dir", versionsDir)
-	}
+	versionsDir := ".stversions"
```
calmh: Silent deprecation of the custom directory? I wonder what the best way to handle this is...

calmh:
All seems sane to me, only issue apart from the nitpicks is the custom path in the staggered versioner. I'm not sure how to manage that. Doing it properly requires allowing file operations between `Filesystems` which is sort of the opposite of what we want to do. Throw a big fat warning and disable the folder on upgrade, if the versions path is outside the folder? Not very user friendly...

lib/fs/basicfs.go
AudriusButkevicius: On windows fsyncing directories fails with this error. I should probably add an OS check and a comment to make that easier to understand.

lib/fs/filesystem.go
AudriusButkevicius: Same goes for Open I guess?

lib/fs/filesystem.go
AudriusButkevicius: Same goes for Open? Or you are fine with that?

lib/model/rwfolder.go
AudriusButkevicius: Yes, it doesn't need closing anymore as we pass in the io.Reader rather than the name.

lib/osutil/atomic.go
AudriusButkevicius: It's still atomic due to renaming?

lib/versioner/staggered.go
AudriusButkevicius: Yes, best thing I can come up with is permitting a symlink stversions, and symlink that to the old path outside of the folder

AudriusButkevicius:
Logfs and errorfs are both used in code. Logfs wraps any fs for debugging, errorfs returns errors for everything (only used when you pass an invalid type to NewFilesystem call. I wanted to skip the fs, err idiom as I'd have to pepper if branches throughout the code. Alternative would be to panic if its unknown type).

AudriusButkevicius:
Also see comment above on addressing staggered versioning.

lib/fs/filesystem.go
calmh: I'm still thinking about that. The real problem, imho, is `OpenFile` which takes a bunch of flags and stuff and is very *file*-specific. That is, does it make sense to be able to send append-only and flags like that? Maybe we can restrict the possibilities

calmh:
Symlinks could be a solution. Will probably fail spectacularly on Windows though. (Either from us trying to do it in code of failing to instruct the user about a complicated thing they need to do.)

AudriusButkevicius:
Ah yes, ok, then perhaps a big red warning and paused folders.

lib/model/rwfolder.go
AudriusButkevicius: We process this in a loop, so we can't defer file closing, unless we move all of the body of the loop to a function.

lib/fs/filesystem.go
AudriusButkevicius: I'll leave it as there is a comment in rwfolder.go:

```
        // We declare a function that acts on only the path name, so
		// we can pass it to InWritableDir. We use a regular Mkdir and
		// not MkdirAll because the parent should already exist.
```

lib/versioner/staggered.go
```diff
@@ -30,15 +30,15 @@ type Interval struct {
 type Staggered struct {
 	versionsPath  string
 	cleanInterval int64
-	folderPath    string
+	filesystem    fs.Filesystem
```
imsodin: Again: fs?

lib/versioner/simple.go
```diff
@@ -22,18 +22,18 @@ func init() {
 
 type Simple struct {
 	keep       int
-	folderPath string
+	filesystem fs.Filesystem
```
imsodin: Why not just fs as for folders?

lib/scanner/walk.go
```diff
@@ -426,54 +417,53 @@ func (w *walker) walkSymlink(ctx context.Context, absPath, relPath string, dchan
...
 
 		// We will attempt to normalize it.
-		normalizedPath := filepath.Join(w.Dir, normPath)
-		if _, err := w.Filesystem.Lstat(normalizedPath); fs.IsNotExist(err) {
+		if _, err := w.Filesystem.Lstat(normPath); fs.IsNotExist(err) {
 			// Nothing exists with the normalized filename. Good.
-			if err = w.Filesystem.Rename(absPath, normalizedPath); err != nil {
-				l.Infof(`Error normalizing UTF8 encoding of file "%s": %v`, relPath, err)
+			if err = w.Filesystem.Rename(path, normPath); err != nil {
+				l.Infof(`Error normalizing UTF8 encoding of file "%s": %v`, path, err)
```
AudriusButkevicius: This might cause a large number of warnings though, plus this PR is not really changing the error message.

lib/fs/basicfs_unix.go
```diff
@@ -0,0 +1,63 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// +build !windows
+
+package fs
+
+import "os"
+
+var symlinksSupported = true
```
AudriusButkevicius: Removing

lib/scanner/walk.go
```diff
@@ -426,54 +417,53 @@ func (w *walker) walkSymlink(ctx context.Context, absPath, relPath string, dchan
...
 				return "", true
 			}
-			l.Infof(`Normalized UTF8 encoding of file name "%s".`, relPath)
+			l.Infof(`Normalized UTF8 encoding of file name "%s".`, path)
 		} else {
 			// There is something already in the way at the normalized
 			// file name.
-			l.Infof(`File "%s" has UTF8 encoding conflict with another file; ignoring.`, relPath)
+			l.Infof(`File "%s" path has UTF8 encoding conflict with another file; ignoring.`, path)
```
AudriusButkevicius: See above.

lib/scanner/walk.go
```diff
@@ -426,54 +417,53 @@ func (w *walker) walkSymlink(ctx context.Context, absPath, relPath string, dchan
...
 }
 
 func (w *walker) checkDir() error {
-	if info, err := w.Filesystem.Lstat(w.Dir); err != nil {
+	if info, err := w.Filesystem.Lstat("."); err != nil {
 		return err
 	} else if !info.IsDir() {
-		return errors.New(w.Dir + ": not a directory")
+		return errors.New(w.Filesystem.URI() + ": not a directory")
 	} else {
```
AudriusButkevicius: Addressed

lib/versioner/staggered.go
AudriusButkevicius: to be fixed

lib/versioner/simple.go
AudriusButkevicius: to be fixed

lib/model/rwfolder.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2014 The Syncthing Authors.
+2// Copyright (C) 2014 The Syncthing Authors.
```
AudriusButkevicius: to be fixed

AudriusButkevicius:
So I think all the points are addressed apart from:

[ ] Update docs for versioners once this is merged
[ ] At the moment there is 3 ways of using filesystem in names: MtimeFS, logFilesystem and here folderFs. Should probably stick to one, (which needs discussions/deciding)

AudriusButkevicius:
@st-jenkins retest

calmh:
So the one point here that becomes an incompatibility is the removal of the custom versions dir for the staggered versioner. This has a fairly significant user base (11%) although we don't know how many of those have custom versioning dirs. I'm not super happy with just removing that functionality outright and posting a GUI notice "we messed up your versioning config for you, sorry not sorry 🖕".

I know it sucks, but what do you think about adding a hack to the staggered versioner to keep it working with a custom dir (i.e., not filesystem aware), and post a notification saying "you have a deprecated setup; this will be removed soon" and then finally killing it in, say, three months or so?

AudriusButkevicius:
How is deprecating now different than in 3 months? What are the corrective actions (apart from stopping using syncthing) can the users take?

calmh:
Removing is different from deprecating. Having an automatic update drop functionality with no prior warning is evil. Doing it with prior warning is still annoying but a *necessary* evil in this case. The corrective action the user can take is to use the default versioning dir or realize ahead of time that they need to change whatever it is they do that depends on a custom versioning dir.

calmh:
Or, we can "hope" that it doesn't affect too many users and deal with the backlash. But then I think it warrants at least a forum post ahead of time.

AudriusButkevicius:
Ok, let me look into wedging filesystem stuff for this. It will probably be ugly.

calmh:
I'd suggest something along the lines of letting the staggered versioner do it's thing old style, just use the filesystem code to get the absolute path of the file it's about to version. In the config migration, we add an alert if we detect a custom versioner path being set. Down the line, we change it to exactly what you have now.

AudriusButkevicius:
I am not against supporting it though, versioning to s3 sounds cool, we should just be consistent across the board for all versioners.

imsodin:
Naive question: Why not have a separate filesystem for that use case?  
Sure you can't rename then and need to copy and del, but that's anyway the case if the versioning dir is on a separate underlying filesystem (which I assume is the most frequent reason for a custom dir).

AudriusButkevicius:
I think in the current case it just fails, as we try to rename.

imsodin:
Sure, but that could simply be solved in an ugly way by adding a condition and then either renaming or just opening the file on folder filesystem and writing on versioner's filesystem. Or create an interface or osutil function for interactions between filesystems to avoid copy and delete when the two filesystem types allow renaming.

AudriusButkevicius:
Agree.

lib/osutil/osutil.go
```diff
@@ -26,12 +26,54 @@ var renameLock = sync.NewMutex()
...
+		// See if one is a prefix of the other filesystem.
+		if from.Type() == to.Type() {
+			fromUri := from.URI()
+			toUri := to.URI()
+			if strings.HasPrefix(toUri, fromUri) {
+				newToPath := filepath.Join(strings.TrimPrefix(toUri, fromUri), toPath)
+				if err := from.Rename(fromPath, newToPath); err == nil {
+					return err
+				}
+			} else if strings.HasPrefix(fromPath, toPath) {
```
imsodin: s/Path/Uri/ ?

lib/osutil/osutil.go
```diff
@@ -26,12 +26,54 @@ var renameLock = sync.NewMutex()
...
+				newFromPath := filepath.Join(strings.TrimPrefix(fromUri, toUri), fromPath)
+				if err := to.Rename(newFromPath, toPath); err == nil {
+					return err
+				}
+			} else {
```
imsodin: Couldn't you just create a new fs with empty uri and rename by concatenating uri and path for source and target instead of searching a prefix?

AudriusButkevicius: This whole thing in general doesn't give me a lot of confidence. 

Was thinking about cases where it's C drive vs D drive, or /a/b/c renamed to /a/b/../../d. In blank

calmh: I don't think we should spend energy on cross filesystem renames. In general we shouldn't do them. There is a special case where this may be useful in some legacy cases like the staggered versioner when both filesystems are "basic", ad we can have a `RenameBetweenBasicFSs` that does the same as `os.Rename` - fails if the two `Filesystems` aren't in fact backed by the same actual underlying filesystem.

I don't want to think too much about what it would mean to "rename" a file from S3 into Upspin or whatever. There is no relevant sense in which it is still the same object with just a different name. The reasonable operation is copy followed by delete, with separate error handling on each.

AudriusButkevicius: Sure, but that's just renaming the function to Move instead of Rename, or is there a bigger issue?

calmh: The bigger issue is that I don't think we should pretend to provide something rename-like when in the background it's hundreds of network operations and syscalls and whatnot, where a failure leaves you in an unknown state with objects present on both sides, maybe fully copied maybe not, etc. It's not something we can successfully abstract. Instead, we should expose the fact that a cross-filesystem "rename" needs to be an open plus a create plus a copy plus a delete and let the user of the API decide what to do with failures at each of those points.

Except in the special case where an actual rename syscall can succeed.

AudriusButkevicius: And whats wrong with having an abstraction that deals with that on our behalf? (Instead of peppering "rename, if fails - copy" everywhere?)

calmh:
Because the abstraction can't deal with it without knowing the appropriate way to deal with it, which depends on the situation it's used in.

Where are all these cross filesystem renames everywhere you refer to anyway?

lib/osutil/osutil.go
imsodin: On IRC calmh brought up this option:

>[...] i'd encapsulate it as something like fs1.MoveTo(fs2) where it attempts magic and if that fails does the ioutil.Copy

To me this looks like: Do essentially the same, but delegate it to the implementation of the filesystem interface. Like this every filesystem can decide whether it even attempts magic or just does copy&delete, making it more transparent what actually happens.

AudriusButkevicius:
Currently in staggered versioner, but I envision all versioners supporting custom filesystems.
I guess what you are saying is that we only have one case of this now,  so lets handle that as the special case.

calmh:
Yeah

lib/versioner/staggered.go
```diff
@@ -309,3 +302,64 @@ func (v *Staggered) Archive(filePath string) error {
...
+			if err = from.Rename(fromPath, newToPath); err == nil {
+				return err
+			}
+		} else if strings.HasPrefix(fromUri, toUri) {
+			newFromPath := filepath.Join(strings.TrimPrefix(fromUri, toUri), fromPath)
+			if err = to.Rename(newFromPath, toPath); err == nil {
+				return err
+			}
+		} else {
+			// Find a prefix, if we can't find it, we can't copy.
```
imsodin: s/copy/rename/

AudriusButkevicius:
So I think this is "finished" of what I had to do, so please give it another proper shake, and if happy, merge.

I'd prefer if we'd merge the "conceptually ok" code and keep the head potentially buggy, and weed out bugs in separate PRs as this one is becoming unbearable in terms Chrome being able to render it.

cmd/syncthing/gui.go
```diff
@@ -1259,23 +1259,36 @@ func (s *apiService) getPeerCompletion(w http.ResponseWriter, r *http.Request) {
 func (s *apiService) getSystemBrowse(w http.ResponseWriter, r *http.Request) {
 	qs := r.URL.Query()
 	current := qs.Get("current")
+	// Default value or in case of error unmarshalling ends up being basic fs.
+	var fsType fs.FilesystemType
+	fsType.UnmarshalText([]byte(qs.Get("filesystem")))
+
```
imsodin: empty line

cmd/syncthing/gui.go
```diff
@@ -1259,23 +1259,36 @@ func (s *apiService) getPeerCompletion(w http.ResponseWriter, r *http.Request) {
...
 	if current == "" {
-		if roots, err := osutil.GetFilesystemRoots(); err == nil {
+		filesystem := fs.NewFilesystem(fsType, "")
+		if roots, err := filesystem.Roots(); err == nil {
+
```
imsodin: gain

lib/versioner/staggered.go
```diff
@@ -309,3 +302,65 @@ func (v *Staggered) Archive(filePath string) error {
...
 }
+
+func renameOrCopy(from fs.Filesystem, fromPath string, to fs.Filesystem, toPath string) (err error) {
+	defer from.Remove(fromPath)
+
+	if from.Type() == to.Type() {
+		fromUri := from.URI()
+		toUri := to.URI()
+		// See if one is a prefix of the other filesystem.
+		if strings.HasPrefix(toUri, fromUri) {
```
calmh: I suspect this suffers from the usual path prefix vulnerability (`c:\foo` being a prefix of `c:\foobar` but actually unrelated). I also (still) think this whole thing is overthinking it. To retain the previous behavior we can just `if from.Type() == fsTypeBasic && from.Type() == to.Type() { return os.Rename(...) }`. For non-basic filesystems (hypothetical at this point) there is no reason to assume that a shared prefix means a rename is possible. I don't think that's the case between S3 buckets for example, even though they may share a region or whatever as prefix.

So all in all I think we take damage from trying to be clever here. This is my only real complaint and that I would want adressed before merging. Otherwise it looks good to me and I think the proper course of action is to merge and get it out there in a candidate for some real world testing.

AudriusButkevicius:
Addressed.

calmh:
@st-review merge it

all: Convert folders to use filesystem abstraction

st-review:
:ok_hand: Merged as 3d8b4a42b72de26f6e15fdc2b2a7feeb116f9577. Thanks, @AudriusButkevicius!

calmh:
🎉

AudriusButkevicius:
So many lines that literally add no functionality.

calmh:
Indeed

calmh:
Might introduce a whole bunch of bugs though!

PR #4229, opened 2017-06-26
Build: Add maximum compression

rasa:
### Purpose

Add maximum compression to .tar and .zip files.
Removed per @calmh:
~~Remove /etc files from zip file, as they are unneeded on Windows.~~
~~Fix tar builds on Windows, as some Windows users might prefer tars over zips (I do).~~
~~Add 'all' option to tar and zip targets, so all executables can be shipped, if desired.~~

### Testing

Results of `go run build.go zip` before and after (1.1% improvement):
````
-rw-r--r-- 1 ross 197121  5694763 Jun 26 13:37 syncthing-windows-amd64-v0.14.31-rc.1+5-gdb1dc998.zip
-rw-r--r-- 1 ross 197121  5629660 Jun 26 13:38 syncthing-windows-amd64-v0.14.31-rc.1+9-g7a9e7183-build-max-compression.zip
````

### Authorship

Ross Smith II (rasa) <ross@smithii.com>

### DIFF
--- a/build.go
+++ b/build.go
@@ -445,7 +445,7 @@ func build(target target, tags []string) {
 
 	tags = append(target.tags, tags...)
 
-	rmr(target.binaryName)
+	rmr(target.BinaryName())
 	args := []string{"build", "-i", "-v", "-ldflags", ldflags()}
 	if len(tags) > 0 {
 		args = append(args, "-tags", strings.Join(tags, " "))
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by calmh** [comment]
Hi! Thanks. In principle I'm fine with this, apart from excluding etc and extra from Windows. The extra stuff includes documentation. The etc stuff may currently be useless, but it's not guaranteed to be so in the future and it's equally useless for some other platforms we build for.

**Comment by calmh** [comment]
Also clean up the comments on the `all` target to explain what happens with binaries instead of the current "can't build archives" thing.

**Comment by rasa** [comment]
@calmh Thanks for reviewing, and considering this PR. I removed the exclusion of the /etc folder, and fixed the 'all' comment. Let me know if there's anything else needed. Thanks!

File: build.go [hunk 1]
```diff
@@ -63,8 +65,8 @@ type archiveFile struct {
 
 var targets = map[string]target{
 	"all": {
-		// Only valid for the "build" and "install" commands as it lacks all
-		// the archive creation stuff.
+		// Only valid for the "build", "install", "tar" and "zip" commands
```
calmh: Actually, why not add the required stuff to this target, maybe keep the later magic to add all binaries automatically, instead of reusing the syncthing target? The filenames etc will make more sense that way.

rasa: Yeah, that makes more sense. Fixing now...

rasa: Sorry, I started to hack on it, and I'm not sure the best way forward, sorry.

calmh:
Maybe scale it back to just the compression changes and add the other stuff at a later time when appropriate.

rasa:
OK, I backed out the build all changes. Let me know if this fits the bill.

calmh:

calmh:
@st-review merge it

build: Use maximum compression when archiving

calmh:
@rasa Thanks!

st-review:
:ok_hand: Merged as 0ca2ed7ad7aa7cd33d657e37233c3805b401c128. Thanks, @rasa!

PR #4232, opened 2017-06-28
Home page: Домашная страница → Сайт

maximal:
1. `Домашная` is a typo. Correct is `Домашняя`.

2. `Домашняя страница` is a bit verbose and causes the footer to be placed in two text rows instead of one (on maximum page width):

image
↓
image


AudriusButkevicius:
Please read the readme at the root of the directory.
Translations are managed via an external service

maximal:
Okay.
You definitely should change `README-FIRST.txt` to `README.md` for better rules visibility.

maximal:
Seems like Russian translation section got stuck there. Editor doesn’t respond for months.

https://www.transifex.com/syncthing/syncthing/language/ru/

PR #4233, opened 2017-06-28
README.md in addition to README-FIRST.txt

maximal:
For better visibility in GitHub web pages.

### DIFF
--- a/gui/default/assets/lang/README.md
+++ b/gui/default/assets/lang/README.md
@@ -1,0 +1,7 @@
+All files in this directory are auto generated. Do not change any of
+them. To contribute translations, please head over to
+
+https://www.transifex.com/projects/p/syncthing/
+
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Having one file is enough, I'd just scrap readme first.

calmh:
I renamed the file instead. Thanks for suggesting this.

maximal:
I would use Markdown there though.
You can add some formatting in future.

AudriusButkevicius:
Yeah like a "Don't modify files here" in the largest possible font in red with many bang signs.

maximal:
!!

AudriusButkevicius:
An idea for bot, auto close PRs referring to files in there.

calmh:
It's not that big of a problem, we can answer them manually for now I think.

PR #4236, opened 2017-06-29
lib/model, gui/default: Add Receive Only Folder option with which a p…

kumartej:
…eer will not advertise local changes to the rest of the peers

### Purpose

Currently only 2 folder types are supported e.g: Send Only, Send & Receive. But these two options do not support the scenario where only fewer node should introduce new changes to the peers. The other nodes/peers should not publish any local changes to the network. This is a typical use case where repositories/files can be synced to multiple peers but writes to these repositories/files are only accepted in a single or fewer nodes. 

In some use cases, it would be dangerous to have a non-designated peer/node advertise local changes globally. For an example, if we are to sync critical configuration files using Syncthing, if we changed anything in one of these files by mistake/accidentally, that change would propagate to **all** of the other peers in no time. This is unacceptable in certain environments.

With this patch, I am introducing a 3rd folder type: Receive Only. Only few nodes with Send Only share would be responsible to introduce new changes to the network. Rest of the nodes/peers would have the Receive Only folder type set. So any local changes in these non-designated peers would not be advertised across the network. Local changes would rather be deleted/overwritten. Thus protecting critical data from accidental changes.

This new folder type is exposed via config file as well as GUI.

### Testing

**Host A**
Folder Type: Send & Receive

**Host B**
Folder Type: Receive Only

In Host A: 
```
$ touch ~/Sync/new_file
$ cat > ~/Sync/new_file
This is the first line
CTRL+D
# content of this file never changes or write on this file from Host B never gets advertised
```

In Host B:
```
$ ls -l ~/Sync/new_file  (new_file is synced)
-rw-r--r-- 1 xxx xxx 0 Jun 28 21:53 /home/xxx/Sync/new_file
$ cat ~/Sync/new_file
This is the first line
$ cat >>~/Sync/new_file
This change will not propagate & will be over written
CTRL+D
# .. after a few seconds ..
$ cat ~/Sync/new_file
This is the first line
```

### Screenshots

image

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -359,7 +359,8 @@
                       <th><span class="fa fa-fw fa-lock"></span>&nbsp;<span translate>Folder Type</span></th>
                       <td class="text-right">
                         <span ng-if="folder.type == 'readonly'" translate>Send Only</span>
-                        <span ng-if="folder.type != 'readonly'">{{ folder.type.charAt(0).toUpperCase() + folder.type.slice(1) }}</span>
+                        <span ng-if="folder.type == 'writeonly'" translate>Write Only</span>
+                        <span ng-if="folder.type != 'readonly' && folder.type != 'writeonly'">{{ folder.type.charAt(0).toUpperCase() + folder.type.slice(1) }}</span>
                       </td>
                     </tr>
                     <tr ng-if="folder.ignorePerms">
--- a/gui/default/syncthing/folder/editFolderModalView.html
+++ b/gui/default/syncthing/folder/editFolderModalView.html
@@ -99,9 +99,11 @@
               &nbsp;<a href="https://docs.syncthing.net/users/foldertypes.html" target="_blank"><span class="fa fa-book"></span>&nbsp;<span translate>Help</span></a>
               <select class="form-control" ng-model="currentFolder.type">
                 <option value="readwrite" translate>Send &amp; Receive</option>
-                <option value="readonly" translate>Send Only</option>
+		<option value="readonly" translate>Send Only</option>
+		<option value="writeonly" translate>Receive Only</option>
               </select>
               <p ng-if="currentFolder.type == 'readonly'" translate class="help-block">Files are protected from changes made on other devices, but changes made on this device will be sent to the rest of the cluster.</p>
+              <p ng-if="currentFolder.type == 'writeonly'" translate class="help-block">Changes made on this device will not be sent to the rest of the cluster.</p>
             </div>
             <div class="form-group">
               <div class="checkbox">
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -210,6 +210,10 @@ func (f *FolderConfiguration) cleanedPath() string {
 	return cleaned
 }
 
+func (f *FolderConfiguration) IsReceiveOnlyFolder() bool {
+	return f.Type.IsReceiveOnlyFolder()
+}
+
 type FolderDeviceConfigurationList []FolderDeviceConfiguration
 
 func (l FolderDeviceConfigurationList) Less(a, b int) bool {
--- a/lib/config/foldertype.go
+++ b/lib/config/foldertype.go
@@ -11,6 +11,7 @@ type FolderType int
 const (
 	FolderTypeSendReceive FolderType = iota // default is sendreceive
 	FolderTypeSendOnly
+	FolderTypeReceiveOnly
 )
 
 func (t FolderType) String() string {
@@ -19,6 +20,8 @@ func (t FolderType) String() string {
 		return "readwrite"
 	case FolderTypeSendOnly:
 		return "readonly"
+	case FolderTypeReceiveOnly:
+		return "writeonly"
 	default:
 		return "unknown"
 	}
@@ -34,8 +37,19 @@ func (t *FolderType) UnmarshalText(bs []byte) error {
 		*t = FolderTypeSendReceive
 	case "readonly", "sendonly":
 		*t = FolderTypeSendOnly
+	case "writeonly", "receiveonly":
+		*t = FolderTypeReceiveOnly
 	default:
 		*t = FolderTypeSendReceive
 	}
 	return nil
+}
+
+func (t FolderType) IsReceiveOnlyFolder() bool {
+	switch t {
+	case FolderTypeReceiveOnly:
+		return true
+	default:
+		return false
+	}
 }
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -181,9 +181,9 @@ func (s *FileSet) Update(device protocol.DeviceID, fs []protocol.FileInfo) {
 		discards := make([]protocol.FileInfo, 0, len(fs))
 		updates := make([]protocol.FileInfo, 0, len(fs))
 		for i, newFile := range fs {
-			fs[i].Sequence = atomic.AddInt64(&s.sequence, 1)
 			existingFile, ok := s.db.getFile([]byte(s.folder), device[:], []byte(newFile.Name))
 			if !ok || !existingFile.Version.Equal(newFile.Version) {
+				fs[i].Sequence = atomic.AddInt64(&s.sequence, 1)
 				discards = append(discards, existingFile)
 				updates = append(updates, newFile)
 			}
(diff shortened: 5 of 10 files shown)


calmh:
Hi and welcome! Without looking at the code, let me just throw some things out there that need to be checked off as part of doing a "receive only" folder. You may already have solved a few or many of these, but please bear with me.

- [ ] Automatic deletions and overrides need to be optional and very clearly warned against.

It's not OK that the user selects "receive only" thinking it sounds good and we suddenly blow away all their data that is not present somewhere else. Consider that the devices may be out of sync when the change happens.

- [ ] What does it look like from the outside (i.e., what index data do we send) when things are out of sync due to local changes? Either before we have managed to revert it back, or when not configured to automatically revert.

You will need to consider conflicts:

- [ ] what happens when there is a conflict between a remote and a local change,
- [ ] in case a conflict between two remote changes is resolved locally, what happens to the "losing" change (it is normally sent back to the cluster as a conflict copy),

kumartej:
Hi @calmh 

After going through your comment. I guess this comment will answer most of the concerns.
- As of now overwrites and deletions are automatic but can be made as optional, giving an override option in Gui.
- To add a receive only folder, it doesn't delete any files when connecting to a cluster. During the initial scan it send the indexes which the folder has and gets synced with the other peers. After initial scan is done the local changes would be overwritten if any.
- There is no question of out of sync due to local change as the local change is being over written.
- I can think of two cases if we are considering conflicts between remote and local changes considering we have to implement override as an option:
   
1. If the user selects an automatic override option then there is no case of conflict as the local changes will be over written and there will be no conflict.
2. If the user doesn't select automatic override, then we can have a conflict file which indicates the conflict until the user clicks override button in Gui.

- In case of conflict in two remote changes, it could work similar to send & receive folder how it could resolve the conflict.

I am curious to know what is the use case of allowing multiple nodes/peers to write or introduce changes that can propagate to the entire cluster. For an example, I share a document with the network, some rouge peer can then introduce a change in the document resulting loss of my "golden" copy. With current implementation of Send and Send & Receive folder types, there's a high risk of this happening. I don't think that sounds like a good idea. This patch particularly address that problem.

calmh:
> 	• As of now overwrites and deletions are automatic but can be made as optional, giving an override option in Gui.

Cool.

> 	• To add a receive only folder, it doesn't delete any files when connecting to a cluster. During the initial scan it send the indexes which the folder has and gets synced with the other peers. After initial scan is done the local changes would be overwritten if any.

I don't fully understand the behavior here. Assuming a completely blank / default configured instance, I:

 - Add a folder with existing files in it, type="receive only". You say it does an initial scan and indexes are sent and nothing changes.
 - I share this folder with another device, which is type="send and receive" and has no contents. What happens?
 - I share this folder with another device, which is type="receive only" and has different contents. What happens?

> 	• There is no question of out of sync due to local change as the local change is being over written.

Except that it is optional and may not be enabled.

> 	• I can think of two cases if we are considering conflicts between remote and local changes considering we have to implement override as an option:
> 	• If the user selects an automatic override option then there is no case of conflict as the local changes will be over written and there will be no conflict.
> 	• If the user doesn't select automatic override, then we can have a conflict file which indicates the conflict until the user clicks override button in Gui.

Right, so a local conflict is probably not a thing here. I guess the "override button" in this case would be an inverse of the override button we have on send only folders: when pressed it would wipe out any local changes, conflicts or not?

> 	• In case of conflict in two remote changes, it could work similar to send & receive folder how it could resolve the conflict.

In this case it would create a conflict copy and send it to the other devices, but this folder is "receive only" so it can't send that conflict copy. The behavior would be the same as if the user created a new file that is not yet known (the conflict copy).

> I am curious to know what is the use case of allowing multiple nodes/peers to write or introduce changes that can propagate to the entire cluster. For an example, I share a document with the network, some rouge peer can then introduce a change in the document resulting loss of my "golden" copy. With current implementation of Send and Send & Receive folder types, there's a high risk of this happening. I don't think that sounds like a good idea. This patch particularly address that problem.

Yes, the use case for this is somewhat clear. I'm not arguing against the existence of a receive only folder, just ironing out the details of how it's supposed to work.

kumartej:
> I don't fully understand the behavior here. Assuming a completely blank / default configured instance, I:
> - Add a folder with existing files in it, type="receive only". You say it does an initial scan and indexes are sent and nothing changes.
> - I share this folder with another device, which is type="send and receive" and has no contents. What happens?
> - I share this folder with another device, which is type="receive only" and has different contents. What happens?

Index messages and building a global index from the peers will be the same as done for send & receive folder type. Receive only folder is the one which doesn't advertise local changes (Index update messages will not be sent for local changes).

> Right, so a local conflict is probably not a thing here. I guess the "override button" in this case would be an inverse of the override button we have on send only folders: when pressed it would wipe out any local changes, conflicts or not?

yeah you are exactly right.

> In this case it would create a conflict copy and send it to the other devices, but this folder is "receive only" so it can't send that conflict copy. The behavior would be the same as if the user created a new file that is not yet known (the conflict copy).

As it has the same behavior of user creating a new file it would be deleted. But the file tries to achieve the global version.

calmh:
Either we are talking around each other or you are envisioning something different than what I expect. The expected behavior is that if a user creates a file on a receive only folder, it will be deleted? If so, why (and how) does this not apply to files already there? And why does it not apply when two receive only devices talk to each other? If index updates are not sent, can other devices not download anything from receive only devices?

kumartej:
This is what happens now:

The first time the receive only folder is up, then it deletes the files in the folder and it tries to achieve the global version of the folder. If there are any local changes after syncing up it deletes the files the changes and does not send any index update message for these local changes. If there is any remote change it tries to sync up with the global version of the folder.

 If a new peer connects to this receive only folder it shares the current index file which it has. If there are only two receive only folders in the cluster then all the files in both the cluster will be deleted (as both of the index files doesn't contain anything).

Index update messages will be sent for all the remote changes that happens in the receive only folder. These message will not be sent for any local changes that happens in the receive only folder. This way we can protect the advertisement of local changes to other peers but the advertisement of remote changes would still continue. So this way other peers can download the file.

calmh:
Right, so that's what I was expecting (things getting wiped out), and that part is unacceptable and needs to be reconsidered.

imsodin:
For simplicity, couldn't that all fall under the manual override default with option to do it automatically? I.e.  don't delete/overwrite any changes by default, but display an out-of-sync message and an override button (potentially with an additional warning that this will wipe your data). Then in advanced settings have a checkbox allowing for unconditional/automatic overwriting. I think this is much more consistent and easy to maintain for devs/understand for users than introducing special cases to protect the data.

calmh:
Yes, that sounds like a reasonable way to do it.

calmh:
This is also an issue:

> Index update messages [...] will not be sent for any local changes that happens in the receive only folder.

That means that other devices think they can sync the file from us, but we will return incorrect data. It would need to advertise the change regardless, but with the invalid bit set to avoid other devices adopting it.

kumartej:
> That means that other devices think they can sync the file from us, but we will return incorrect data. It would need to advertise the change regardless, but with the invalid bit set to avoid other devices adopting it.

I couldn't understand why it returns incorrect data If we over write the local changes. After over writing the changes it will be in sync with the global version, so this means other devices can sync the file.

imsodin:
Yes, there is no problem when you overwrite local changes automatically, but this cannot be the default behaviour. Not because of a technical limitation, but because of [Syncthing's first goal to be safe from data loss](https://github.com/syncthing/syncthing/blob/master/GOALS.md#1-safe-from-data-loss), i.e. this is a policy decision. So when local changes happen but are not overwritten yet (because the user neither pressed a manual button nor set the non-default setting to overwrite automatically), the peers need to know that you don't have the correct version of a file, otherwise they will try to get it from you while it isn't there.

kumartej:
I want some feedback for my changes in the codebase to proceed further. Could anyone go through my changes in the pr.

AudriusButkevicius:
Before we go to the code the generic behaviour is flawed.

This folder will permanently look out of sync and will never allow other devices to download from this device (as the device never advertises anything).

Furthermore, as it was pointed out above, this can lead to data loss that's not intentional (accidentially set folder type to receive only and all my files are gone).

calmh:
What Audrius says. We need to agree on what the high level behavior is supposed to be. Then we can talk code.

PR #4237, opened 2017-06-29
Made an macOS icns (icon) file

NoLooseEnds:
Made a macOS icns (icon) file to use in Finder for differentiating the different directories (same way like i.e Dropbox and Google Drive).

icon_512x512

Info on how to make and use an icns file here: https://blog.macsales.com/28492-create-your-own-custom-icons-in-10-7-5-or-later


NoLooseEnds:
Forgot to _not_ convert the color profile. Fixing.

calmh:
This is neat looking, but we currently don't set any folder icons on things. We'll probably *sync* the folder icon if someone does, though.

NoLooseEnds:
@calmh So far you have to set it manually in Finder. But that's a one time thing, so it doesn't bother me too much. It was more a thing I made now for my self, but wanted to make available for others.

I'm not sure how to set it in the Finder sidebar (like Dropbox/Google Drive), but I think that involves code and hooking into Finder in some way.

I made a new pull request with the correct color profile (i.e not converting to SRGB) here: https://github.com/syncthing/syncthing/pull/4238

I'm still pretty new to GitHub, so I'm not sure how to update a pull request. I ended up deleting the whole thing and starting over with a new pull request.

PR #4239, opened 2017-06-29
gui: Fix discovered devices list (follow-up to #4186)

imsodin:
Thanks @dinosore for discovering the problem. Filtering devices didn't work (`in` is unfortunately not the same in JS as in python...). Also the list of device ids was shown when editing an existing device - no idea why, but cleaning up the html structure also got rid of that glitch.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1186,13 +1186,17 @@ angular.module('syncthing.core')
             return $http.get(urlbase + '/system/discovery')
                 .success(function (registry) {
                     $scope.discovery = [];
+                    outer:
                     for (var id in registry) {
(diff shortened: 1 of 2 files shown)


imsodin:
@st-jenkins retest

imsodin:
@st-jenkins test this please

calmh:
@st-review merge

st-review:
:ok_hand: Merged as 8d13e01342bf74c1006c7527c16e088f7a66d0c1. Thanks, @imsodin!

calmh:
New RC tomorrow.

dinosore:
Thanks @imsodin for quick fix.
Mildly disappointed that only device ids and not names also shown in list on previous iteration.
[
image
](https://en.wikipedia.org/wiki/Winnie-the-Pooh#Characterl)

PR #4252, opened 2017-07-07
meta: Move metadata checks into meta directory, make them tests

calmh:
This moves a few things from script/ to a new directory meta/, and makes
them real Go tests. These are the authors, copyright, metalint and gofmt
checks. That means that they can now be run by

go test -v ./meta

and optionally filtered by the usual -run thing to go test. Also -short
will cut down on the metalint stuff and exclude the authors check (which
is slow because it runs git lots of times).

Mainly this makes everything easier on things like build servers where
we can now just run tests instead of do a bunch of scripting.

### DIFF
--- a/build.go
+++ b/build.go
@@ -170,41 +170,6 @@ var targets = map[string]target{
 	},
 }
 
-var (
-	// fast linters complete in a fraction of a second and might as well be
-	// run always as part of the build
-	fastLinters = []string{
-		"deadcode",
-		"golint",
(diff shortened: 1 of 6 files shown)


calmh:

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 200a7fc844895ad4f12cce66bf54deff9583460d. Thanks, @calmh!

PR #4253, opened 2017-07-09
build: Support builds outside of GOPATH

calmh:
This adds support for building with the source placed anywhere and no GOPATH set. The build script handles this by creating a temporary GOPATH in the system temp dir (or another specified location) and mirroring the source there before building. The resulting binaries etc still end up in the same place as usual, meaning at least the "build", "install", "tar", "zip", "deb", "snap", "test", "vet", "lint", "metalint" and "clean" commands work without a GOPATH. To this end these commands internally use fully qualified package paths like "github.com/syncthing/syncthing/cmd/..." instead of "./cmd/..." like before.

There is a new command "gopath" that prepares and echoes the directory of the temporary GOPATH. This can be used to run other non-build go commands:

```
export GOPATH=$(go run build.go gopath)  // GOPATH is now set
go test -v -race github.com/syncthing/syncthing/cmd/...
```

There is a new option "-no-build-gopath" that prevents the check-and-copy step, instead assuming the temporary GOPATH is already created and up to date. This is a performance optimization for build servers running multiple builds commands in sequence:

```
go run build.go gopath // creates a temporary GOPATH
go run build.go -no-build-gopath -goos=... tar // reuses GOPATH
go run build.go -no-build-gopath -goos=... tar // reuses GOPATH
```

The temporary GOPATH is placed in the system temporary directory (os.TempDir()) unless overridden by the STTMPDIR variable. It is named after the hash of the current directory where build.go is run. The reason for this is that the name should be unique to a source checkout without risk for conflict, but still persistent between runs of build.go.

There should be no difference for people who have the source in the right place in the GOPATH. The new behavior is triggered by GOPATH being unset or empty, when we would previously try to guess a value of GOPATH.

### DIFF
--- a/build.go
+++ b/build.go
@@ -213,6 +213,7 @@ func main() {
 			}
 		}
 		os.Setenv("GOPATH", gopath)
+		log.Println("GOPATH is", gopath)
 	}
 
 	// Set path to $GOPATH/bin:$PATH so that we can for sure find tools we
--- a/build.go
+++ b/build.go
@@ -42,6 +42,7 @@ var (
 	race          bool
 	debug         = os.Getenv("BUILDDEBUG") != ""
 	noBuildGopath bool
+	extraTags     string
 )
 
 type target struct {
@@ -260,6 +261,7 @@ func runCommand(cmd string, target target) {
 		if noupgrade {
 			tags = []string{"noupgrade"}
 		}
+		tags = append(tags, strings.Fields(extraTags)...)
 		install(target, tags)
 		metalintShort()
 
@@ -268,6 +270,7 @@ func runCommand(cmd string, target target) {
 		if noupgrade {
 			tags = []string{"noupgrade"}
 		}
+		tags = append(tags, strings.Fields(extraTags)...)
 		build(target, tags)
 
 	case "test":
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
@st-review lgtm

st-review:

imsodin:
I actually liked the previous behaviour of just assuming the working directory is within a valid go dir structure and setting GOPATH accordingly, because I usually forget to set GOPATH. Obviously my laziness is no valid argument against this improvement, but please print a warning to the command line that GOPATH isn't set and building happens in a temporary dir.

calmh:
@imsodin I reinstatated the print that was there before when it guessed the GOPATH:

```
$ go run build.go 
GOPATH is /var/folders/9l/prphpj9n431dsb2myvdjpj4w0000gp/T/syncthing-c0ad971a
ok  	_/Users/jb/src/github.com/syncthing/syncthing/meta	3.197s
```

I think that now looks odd enough to serve as a warning to those who expected some other GOPATH. :)

imsodin:
Yes, that is fine. Another question: What exactly is the intention for the -no-build-gopath flag?
Don't get me wrong, I am not trying to get back the previous behaviour, but if this is intended as a "legacy-mode", wouldn't it make sense to set GOPATH to "../../../../" as before.

calmh:
It's not intended as a legacy mode, it's just to avoid extra work when running build.go multiple times in a row. The dotdot stuff fails when you're running in /tmp/buildagent/A378E55C or whatever, which is what I want to work.

imsodin:
As I said, I was just wondering what the purpose was, so completely fine by me (lazy me now uses an alias to set the GOPATH and change the working dir to the syncthing base dir, which I should have done in the first place :) ).

imsodin:
@st-review lgtm

st-review:
:ok_hand: Merged as 7c8418f493398aea5d5c3c4cdbd3b4e46762b56a. Thanks, @calmh!

PR #4265, opened 2017-07-21
lib/model: resolve file conflicts for new file (#3742)

### LINKED ISSUE #3742: lib/model: Newly created conflicting files can be silently overwritten
Syncthing overwrote a conflicting file during a sync without noticing the conflict.

I expected Syncthing to notice the file it was about to overwrite had not yet been seen through a local scan and either rename it, or create the new file with a "conflict" filename.

Syncthing Version: v0.4.11 both sides
OS Version: Windows 10 / FreeBSD 9.3 / Fedora 24

Steps:
1) Add a new file X on machine A (Windows 10).
2) Add a new file X on machine B (FreeBSD 9.3) with different content.
3) Rescan on machine A. Triggers sync to machine B.

Machine B overwrites file X without warning since it hasn't discovered it via a scan yet.
I've tested this between a few different OSes in different directions.
Other observations:
- Modifying existing files on both sides before a single rescan appears to work correctly.
- Deleting the test files then rescanning/recreating on both sides/rescanning appears to generate the expected conflict file.

It is slightly contrived I know, but it is more likely when using longer rescan intervals to conserve battery / IOPS. There is no safety net if multiple machines create a file with the same name within a rescan interval.

liusy182:
### Purpose

Always use `currentFolderFile()`'s version which is more up-to-date than
cached version in `sendReceiveFolder` struct.
Fixes  #3742

### Testing

Tested with steps described in the mentioned issue and it worked properly with `xxx-conflict-xxx` file generated.

### DIFF
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -853,7 +853,8 @@ func (f *sendReceiveFolder) deleteFile(file protocol.FileInfo) {
 		// we have resolved the conflict.
 		file.Version = file.Version.Merge(cur.Version)
 		err = osutil.InWritableDir(func(name string) error {
-			return f.moveForConflict(name, file.ModifiedBy.String())
+			_, err := f.moveForConflict(name, file.ModifiedBy.String())
+			return err
 		}, realName)
 	} else if f.versioner != nil {
 		err = osutil.InWritableDir(f.versioner.Archive, realName)
@@ -1450,8 +1451,26 @@ func (f *sendReceiveFolder) performFinish(state *sharedPullerState) error {
 			if err = osutil.InWritableDir(os.Remove, state.realName); err != nil {
 				return err
 			}
-		case state.info == nil || !state.info.ModTime().Equal(stat.ModTime()) || state.info.Size() != stat.Size() ||
-			f.inConflict(state.version, state.file.Version):
+		case state.info == nil || !state.info.ModTime().Equal(stat.ModTime()) || state.info.Size() != stat.Size():
+			// `info` represents file stat before pulling start. stat represents
+			// file stat now. nil info but valid stat means the file is newly created
+			// during pulling. diff between info and stat means the file exists
+			// previously but modified during pulling.
+			state.file.Version = state.file.Version.Merge(state.version)
+			err = osutil.InWritableDir(func(name string) error {
+				newName, err := f.moveForConflict(name, state.file.ModifiedBy.String())
+				if err != nil {
+					return err
+				}
+				// trigger a scan for the moved file to update its version
+				scanDir := filepath.Join(filepath.Dir(state.file.Name), filepath.Base(newName))
+				go f.Scan([]string{scanDir})
+				return nil
+			}, state.realName)
+			if err != nil {
+				return err
+			}
+		case f.inConflict(state.version, state.file.Version):
 			// The new file has been changed in conflict with the existing one. We
 			// should file it away as a conflict instead of just removing or
 			// archiving. Also merge with the version vector we had, to indicate
@@ -1459,7 +1478,8 @@ func (f *sendReceiveFolder) performFinish(state *sharedPullerState) error {
 
 			state.file.Version = state.file.Version.Merge(state.version)
 			err = osutil.InWritableDir(func(name string) error {
-				return f.moveForConflict(name, state.file.ModifiedBy.String())
+				_, err := f.moveForConflict(name, state.file.ModifiedBy.String())
+				return err
 			}, state.realName)
 			if err != nil {
 				return err
@@ -1668,20 +1688,20 @@ func removeAvailability(availabilities []Availability, availability Availability
 	return availabilities
 }
 
-func (f *sendReceiveFolder) moveForConflict(name string, lastModBy string) error {
+func (f *sendReceiveFolder) moveForConflict(name, lastModBy string) (string, error) {
 	if strings.Contains(filepath.Base(name), ".sync-conflict-") {
 		l.Infoln("Conflict for", name, "which is already a conflict copy; not copying again.")
 		if err := os.Remove(name); err != nil && !os.IsNotExist(err) {
-			return err
+			return "", err
 		}
-		return nil
+		return "", nil
 	}
 
 	if f.MaxConflicts == 0 {
 		if err := os.Remove(name); err != nil && !os.IsNotExist(err) {
-			return err
+			return "", err
 		}
-		return nil
+		return "", nil
 	}
 
 	ext := filepath.Ext(name)
@@ -1709,7 +1729,8 @@ func (f *sendReceiveFolder) moveForConflict(name string, lastModBy string) error
 			l.Debugln(f, "globbing for conflicts", gerr)
 		}
 	}
-	return err
+
+	return newName, err
 }
 
 func (f *sendReceiveFolder) newError(path string, err error) {
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -1151,7 +1151,6 @@ func (f *sendReceiveFolder) handleFile(file protocol.FileInfo, copyChan chan<- c
 		available:        reused,
 		availableUpdated: time.Now(),
 		ignorePerms:      f.ignorePermissions(file),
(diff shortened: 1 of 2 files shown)


calmh:
It's not super clear to me how this helps. In theory the file info might be more up to date, but in practice I don't think it can be, because it is only updated by scanning and we do not scan while we are pulling. The root cause in #3742 is that if a file is created after we've started pulling it is not present in the database at all and we don't know to check for a conflict.

lib/model/rwfolder.go
```diff
@@ -1151,7 +1151,6 @@ func (f *sendReceiveFolder) handleFile(file protocol.FileInfo, copyChan chan<- c
 		available:        reused,
 		availableUpdated: time.Now(),
 		ignorePerms:      f.ignorePermissions(file),
-		version:          curFile.Version,
```
calmh: If we were to do this, you most also remove this field from the struct declaration so there's nothing else mistakenly using the now uninitialized field.

liusy182: Opps I thought I have done that but I did not. Thanks for catching this.

liusy182:
Thanks for explaining that. I think I am still a newbie to the code :)
For some reason the scenario mentioned by #3742 started to work between my Mac and Windows after applying this fix. Maybe the right approach is to utilize `f.mtimeFS.Lstat`? I will take a closer look on that.

liusy182:
@calmh I have updated PR to compare file Stat right before pulling v.s. right before pulling is about to finish. Tested with 2 mac machines syncing across each other and working.

By the way, the previous solution was wrong as you have pointed out. I have tested again with 2-Macs set up and it didn't work :(

calmh:
I'm suspicious. Stat()ing somewhere is probably the correct approach, but here you're saving the result of the initial stat() which may be nil if the file doesn't exist at that point (new file) and then expecting to run ModTime() on it later which should panic. But I haven't tested.

AudriusButkevicius:
You should stat just before the conflict check/rename I think.

liusy182:
Sorry for my negligence on the nil pointer. The recovery mechanism worked pretty well that I did not even notice the panic. I have added the `nil` check

AudriusButkevicius:
I think this os wrong as the lstat is potentially old by the time the conflict check happens

liusy182:
correct me if I am wrong, `stat()` check happens in the [same goroutine](https://github.com/syncthing/syncthing/pull/4265/files#diff-feb9fde15f8dd977fe9b279a36a447a2R1436) not too far away from where it is used. It should be quite safe to use it?

AudriusButkevicius:
So I think this is fundamentally wrong.

You stat the file, then go about modifying it, downloading stuff from the internet, which takes a long time, and then say, does the mtime of the file I modified match it before I modified it (which is always false), if it does not then you move the file for conflict.

The code might be not that far off apart in terms of distance, but it's definately on a different routine and it's miles away in terms of time.

AudriusButkevicius:
It also moves the file for conflict if we didn't hit the branch that populates info, unconditionally.

liusy182:
Sorry I am not able to get the use case you have given clearly. Could you clarify the steps a bit more so that I can try to reproduce the problem?

AudriusButkevicius:
It's not a use case. If you follow the steps I described you will understand why you fix does not work (or rather makes it wrong). If you think it works, you can walk me step by step explaining how it works.

liusy182:
> You stat the file, then we go about modifying it, downloading stuff from the internet, which takes a long time, and then say, does the mtime of the file I modified match it before I modified it (which is always false), if it does not then you move the file for conflict.

To rephrase it: 
Stat the file -> then start downloading from internet, while downloading we modify the local file -> file gets downloaded -> file downloaded has different mtime from local file -> we move file for conflict

Isn't this the desired behavior or am I missing something here?

AudriusButkevicius:
But we modify the file ourselves modify the file (as we download stuff into it), so there is no way mtime will match.

liusy182:
My understanding was that content is downloaded to the temp file, not the actual file. Could you pinpoint me in code where the actual file gets modified during download?

AudriusButkevicius:
Ok, yes, that is true, but you don't hit this branch if the file does not currently exist, or if it's a directory:

https://github.com/syncthing/syncthing/pull/4265/files#diff-feb9fde15f8dd977fe9b279a36a447a2L1056

info is set to nil. In the finisher you hit this branch, as info is nil:

https://github.com/syncthing/syncthing/pull/4265/files#diff-feb9fde15f8dd977fe9b279a36a447a2R1453

And you try to remove/move for conflict a file that does not exist.

I think you should replace storing version + info with just storing curFile, and then in the finisher compare stat.ModTime() with curFile.ModTime() + sizes, if anything does not match, instead of trying to move things for conflict or whatever, just abort, schedule a rescan, and hope that next time around the files does not get modified.

This is because the way do you it now, the versions do not indicate a conflict (as the file that was modified was not rescanned, so the version counter hasn't been incremented), yet you move it away for conflict anyway.

AudriusButkevicius:
Also, we could probably remove the pre download stat check that schedules the scan, and just do it afterwards.

liusy182:
Yes. If the file does not exist before download, info will be nil. However after download,
 - if the file still not exist, we will not enter this [if branch](https://github.com/syncthing/syncthing/pull/4265/files#diff-feb9fde15f8dd977fe9b279a36a447a2R1436). This means we will simply rename tempfile to file.
 - if the file appears now, it means we have modified it while download, hence we move for conflict.

Do you see any flaw here?

> I think you should replace storing version + info with just storing curFile, and then in the finisher compare stat.ModTime() with curFile.ModTime() + sizes, if anything does not match, instead of trying to move things for conflict or whatever, just abort, schedule a rescan, and hope that next time around the files does not get modified.

This is probably a better approach. I am not sure whether aborting the entire pulling process at final stage and reschedule a scan will cause any wastage on network / processing, and whether this can be a concern?

AudriusButkevicius:
Ok, this now makes sense, but we should add comments, as it's not obvious, because a lot of the assumptions are made around if branches which are miles away from the statements.

I might be wrong, and @calmh might have to correct me, but I think we should schedule a rescan, like in the case before the download (as we could do the same, rename and steam ahead), as now the versions don't seem to conflict, and we don't bump the versions (yet we should).

calmh:
(Disclaimer: didn't read code, just thinking out loud based on the comments)

Maybe we can simplify the "trigger rescan" case to just always do a conflict rename instead. Any local change concurrent with us trying to sync a file should be a conflict so maybe we can save some time here.

calmh:
@st-review rebuild this

AudriusButkevicius:
We should rescan ir patch up things manually. As a minimum to get the device that caused the conflict in the version vector right.

liusy182:
Based on discussion, I see there are a few ways to handle file that is newly created / modified during pulling, at the finisher stage:

1. abort pulling of the current file (do not rename, do not move for conflict), trigger a scan for this file immediately. The problem is, if the conflicting file is the only file that gets changed, we will trigger [10 scan goroutine](https://github.com/syncthing/syncthing/pull/4265/files#diff-feb9fde15f8dd977fe9b279a36a447a2R228) before we finally able to do the scan. This is probably bad and I do not see an easy way to fix this.

2. move the fie for conflict. Update the Version of the moved file. I am not too sure whether this is possible since the moved file is new and not yet in DB. Could you provide some input here?

3. move file for conflict, do nothing but let the next scheduled scan to update its version.  This seems to be harmless because the conflicting file for sure will get its version updated first before it is synced with other devices.

liusy182:
@AudriusButkevicius do you have any input on option 2 or you would agree that option 3 is the way to go so far? I will add more comments to document the code if it is agreed.

AudriusButkevicius:
I think #3 is the right option in this case. I don't think we should try and inject files manually, and just let the scan do it. This also applies for the case where we schedule the scan manually, as in we could move the file as a conflict before we trigger a rescan and then make sure the scan happens right after pulling.

liusy182:
Thanks @AudriusButkevicius  I have added comments and now triggers a rescan when file version is outdated based on your comments.

AudriusButkevicius:
Fixed in my pull request.

Issue #2660, opened 2016-01-05
Add top level Hash field to FileInfo


calmh:
We should have a file-total Hash field directly on the FileInfo structures. This has a couple of use cases:
- It enables comparison of two files to determine if they are identical without loading and walking the entire block list (assuming the hash is secure).
- It can be kept also for deleted files, thus indicating what the contents were of the file that was deleted. This could be used as input to smarter conflict detection around deletes.
- It could (at some point, in the future) be used as a one level Merkle tree, in that we we wouldn't need to send the full block list in the index exchange. The device could then instead ask for the block list for files that have unknown top level hashes.

The top level hash should not be a straight hash of the file contents, it should be a hash of the block hashes. This enables recomputing it quickly when only a few blocks have changed.

Assuming we have configurable hash algorithms, should this follow the configured hash algorithm or always be SHA256? I'm leaning towards the latter as it makes things simpler if we can assume the hash is secure, and the cost is negligible anyway.

AudriusButkevicius:
Well a secure hash over insecure hashes doesn't make it secure.

calmh:
That depends... Even with an insecure hash like Murmur3 we assume that it's secure *enough* that for a given block at a given offset in a given file the same hash implies the same data. If we don't feel we can assume even that, we can't use that hash algo.

PR #4269, opened 2017-07-25
lib/protocol: Add SummaryHash field (ref #2660)

calmh:
This field is SHA256(all block hashes). It gives us a top level field that can be quickly compared to check whether two files are identical. This doesn't actually put it to use, but I ran into the issue and figured this is a five minute fix. With this in place we are prepared to use it and the data is in the database, for very little additional cost.


AudriusButkevicius:
So I am super skeptical of this. This just shifts the problem to the other side. I suspect comparison is much cheaper than hashing in this case. Whats the benefit here actually, or the problem we are solving?

calmh:
Fine. Given that this is trivial it can be recreated if required. Then I'm closing the related issue as well.

calmh:
Well, actually, the point where comparison is more expensive is when there are tens of thousands of blocks and we can avoid loading from db to begin with as we can see from the hashhash that we don't need to. But lets do this when we do that, in that case.

PR #4289, opened 2017-08-08
lib/versioner: Clean the versions dir of symlinks, not the full folder

calmh:
Prevents the damage, while not recovering from it.

### DIFF
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -142,6 +142,10 @@ func TestRequest(t *testing.T) {
 }
 
 func TestSymlinkRecovery(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("symlinks not supported on Windows")
+	}
+
 	ldb := db.OpenMemory()
 
 	fs := db.NewFileSet("default", ldb)


AudriusButkevicius:
Why can't we do a one-off cleanup as part of config migrations? If someone sets up a symlink explicitly perhaps we should not nuke it?

AudriusButkevicius:
Regarding recovery, we could check time of

AudriusButkevicius:
Sry, pocket actions.

calmh:
This thing should be removed in the long run anyway. I'm not super keen on trusting this to a migration; there might be folders that are offline or whatnot.

AudriusButkevicius:
Offline?
So when are we removing this?

calmh:
So what we could do for recovery is list our "have" index, then look for any other older entries in the index for deleted symlinks where the delete is recent. I'm not super hopeful that many places will have those entries intact but it's worth a try I guess.

"Offline" as is on a removable disk for example.

I'd remove this when we have a small-digit number of <0.14.34 clients out there.

AudriusButkevicius:
In my phone so can't check, but I don't think we can classify deletion as recent as we don't update the mtime?

AudriusButkevicius:
If we knew they are recent, we could check mtime of syncthing.old binary or time when release hit the shelves as the cutoff period, and time when the release was pulled as the other cutoff

calmh:
We set the mtime to the delete time. I'm just less hopeful that we find an old index entry. But I'm hacking it currently.

calmh:
OK, here's another attempt that's cleaner. It does the cleanup in a migration after all. Later, as part of the same migration, it attempts a recovery sweep.

calmh:
@st-review merge

st-review:
:ok_hand: Merged as fa5c890ff6a7fe9eb8b2a34900058fe6637ba682. Thanks, @calmh!

lib/model/model.go
```diff
@@ -2721,3 +2726,76 @@ func rootedJoinedPath(root, rel string) (string, error) {
...
+		}
+
+		// It's deleted. Check if it was deleted in the bad window.
+		if fi.ModTime().Before(startDate) || !fi.ModTime().Before(endDate) {
+			return true
+		}
+
+		// Try to find an older index entry.
+		for deviceID := range m.cfg.Devices() {
+			olderFI, ok := fs.Get(deviceID, fi.Name)
```
AudriusButkevicius: Do we unset target on delete?

calmh: Yes

calmh:
@AudriusButkevicius I merged it because I need to drop something within the hour, but since you're here I'll give you a few minutes to find serious problems with this as well.

Not minor nits though, at this point.

AudriusButkevicius:
Lgtm

calmh:
Oh. We also don't actually set the modified time like I though we do, so the recovery step is broken.

calmh:
I don't think there's anything reliable we can do do undo that then unfortunately

AudriusButkevicius:
Fine, drop the extra code in that case, and make an enhancement for the future?

calmh:
Yes

PR #4301, opened 2017-08-11
lib/model: Fix race in GetIgnores (fixes #4300)

### LINKED ISSUE #4300: Querying ignores is racy leading to potential nil deref
https://forum.syncthing.net/t/reason-for-panic-maybe-too-little-ram/10346

The lock is not held between getting the value for the queried folder from `folderCfgs` and from `folderIgnores`. Therefore the latter does not necessarily exist and it's existence isn't checked.

PR incoming.

imsodin:
While searching for the cause of #4300  I found two other minor issues:

 - Some exported functions on `*Matcher` do a "nil-check" at the beginning, others don't. I changed it to be consistent in favor of not checking, because it is the responsibility of the caller to not call a function on nil (I didn't find any instance in the code, where it is the other way around). To make tests work and to be true to [this comment](https://github.com/imsodin/syncthing/blob/0f00bec72c63aeec165a84e5da08a71350bcd1bc/lib/scanner/walk.go#L51), the check whether the Matcher is nil is now done in walk.go.

 - `GetIgnores` returned an error, if .stignore does not exist, while this is perfectly valid.

I still have two separates locks, as the `checkFolderPath` must be called in any case, but doesn't require the lock (it doesn't make sense to load .stignore when the folder marker is missing). The original problem is fixed by checking the second map retrieval.

Originally reported in:
https://forum.syncthing.net/t/reason-for-panic-maybe-too-little-ram/10346

Regression from #3996

### DIFF
--- a/lib/ignore/ignore.go
+++ b/lib/ignore/ignore.go
@@ -177,7 +177,7 @@ func (m *Matcher) parseLocked(r io.Reader, file string) error {
 }
 
 func (m *Matcher) Match(file string) (result Result) {
-	if m == nil || file == "." {
+	if file == "." {
 		return resultNotMatched
 	}
 
@@ -232,10 +232,6 @@ func (m *Matcher) Lines() []string {
 
 // Patterns return a list of the loaded patterns, as they've been parsed
 func (m *Matcher) Patterns() []string {
-	if m == nil {
-		return nil
-	}
-
 	m.mut.Lock()
 	defer m.mut.Unlock()
 
--- a/lib/ignore/ignore.go
+++ b/lib/ignore/ignore.go
@@ -177,7 +177,7 @@ func (m *Matcher) parseLocked(r io.Reader, file string) error {
 }
 
 func (m *Matcher) Match(file string) (result Result) {
-	if file == "." {
+	if m == nil || file == "." {
 		return resultNotMatched
 	}
 
@@ -232,6 +232,10 @@ func (m *Matcher) Lines() []string {
 
 // Patterns return a list of the loaded patterns, as they've been parsed
 func (m *Matcher) Patterns() []string {
+	if m == nil {
+		return nil
+	}
+
 	m.mut.Lock()
 	defer m.mut.Unlock()
 
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1245,33 +1245,31 @@ func (m *Model) ConnectedTo(deviceID protocol.DeviceID) bool {
 
 func (m *Model) GetIgnores(folder string) ([]string, []string, error) {
 	m.fmut.RLock()
+	defer m.fmut.RUnlock()
+
 	cfg, ok := m.folderCfgs[folder]
-	m.fmut.RUnlock()
-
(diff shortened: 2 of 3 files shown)


AudriusButkevicius:
So I'd personally do the nil checks inside the member functions, as it makes most of things cleaner for not having to do a nil check.

This is still racey, as you access two variables under two different locks, but I guess it now just returns nothing as second ok will be false.

imsodin:
No longer racy, but arguably uglier - better?

```golang
func (m *Model) GetIgnores(folder string) ([]string, []string, error) {
	m.fmut.RLock()
	cfg, cfgOk := m.folderCfgs[folder]
	ignores, ignoresOk := m.folderIgnores[folder]
	m.fmut.RUnlock()

	if !cfgOk {
		cfg, cfgOk = m.cfg.Folders()[folder]
		if !cfgOk {
			return nil, nil, fmt.Errorf("Folder %s does not exist", folder)
		}
	}

	if err := m.checkFolderPath(cfg); err != nil {
		return nil, nil, err
	}

	if !ignoresOk {
		ignores = ignore.New()
		path := filepath.Join(cfg.Path(), ".stignore")
		if err := ignores.Load(path); err != nil && !os.IsNotExist(err) {
			return nil, nil, err
		}
	}

	return ignores.Lines(), ignores.Patterns(), nil
}
```

The main reason why I didn't want to do the nil check in ignores.go was, that it never happens in "real" code, only in tests, where the matcher was just omitted instead of created without loading anything. This could also be fixed explicitly in the tests or in `Walk` (in walker.go)[https://github.com/syncthing/syncthing/blob/master/lib/scanner/walk.go#L82] by adding another check for the Matcher.
I feel like the last option would be cleanest, as it prevents nil deref in the first place thus preventing nil-checks everywhere.

AudriusButkevicius:
I think a nil matcher should just be a noop like it was before, on all methods.
Your code can be as it was before just change the locking to:

```
m.fmut.RLock()
defer m.fmut.RUnlock()

... do stuff ...

```

Then it's not ugly and the lock is held as long as it needs to be.

imsodin:
There is more about ignores that I would like to change/optimize/clean up, that probably needs more discussion and is not directly related to the race to be fixed. So I will clean this PR up such that it only fixes the race and leaves everything else for another PR.

AudriusButkevicius:
This acquires the rlock twice in a nested fashion.

AudriusButkevicius:
So as a good citizen, can you add if m == nil check in Lines() and Hash() (should return value of hashPatterns(nil))

imsodin:
@AudriusButkevicius Can we make that part of the announced separate PR? Or more to the point: I want to put that up for discussion at a more appropriate place. I still have reservations against this that need to be argued away or overpowered by a second opinion in order to do that myself.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as ab8c2fb5c7189722e4fb2afe53ccdc7b0a3771c3. Thanks, @imsodin!

PR #4310, opened 2017-08-21
lib/ignore: Consistent behaviour for nil *Matcher

imsodin:
Some exported functions on `*Matcher` check whether it is nil, others don't. This PR adds such check consistently.
As the first match determines the result when ignoring, duplicate lines have no effect. Therefore these are now filtered out when parsing the file.


lib/ignore/ignore.go
```diff
@@ -125,40 +127,42 @@ func New(fs fs.Filesystem, opts ...Option) *Matcher {
 }
 
 func (m *Matcher) Load(file string) error {
+	if m == nil {
+		return IsNil
+	}
```
AudriusButkevicius: I'd just return nil instead of IsLink everywhere, a nil matcher is a noop.

calmh:
Honestly, wouldn't it be simpler to just have it never check for nils and make sure that it is not nil when used from the `Model`? I bet that originally it was a neat shortcut to be able to pass a nil matcher when there were no ignores, and have Match() and Patterns() return something sane anyway. But apparently we don't do that any more (I guess?) since Lines() would explode when called on a nil matcher.

lib/ignore/ignore.go
calmh: But Load() on a nil matcher is a clear error and should (imho) panic as it's always a bug.

AudriusButkevicius: well I'd panic anyway if we just removed a check with a nil pointer dereference.

imsodin:
I should have referenced earlier discussions on this:  
#4301  
https://forum.syncthing.net/t/how-to-expose-ignore-package/10356

In a nut-shell: Audrius considers nil-Matcher a valid noop-Matcher, my opinion is/was to consider a nil `Matcher` as not valid everywhere. No other feedback (should have pinged) -> I was going with his option.

imsodin:
The last commit is the other variant where no nil checks are done, making it "illegal" to use `Matcher` nil pointers (panic will happen on locking). The only other adaption needed is to add a check for nil in `scan.Walk` as ignores aren't set in tests.

imsodin:
I split the unrelated part about .stignore line dedupliation into it's own PR #4350, such that this is now exclusively about the nil business.

AudriusButkevicius:
I don't have an opinion on this, so I'll leave to to @calmh

calmh:
@st-review merge it

lib/ignore: Consistent behaviour for nil *Matcher

st-review:
:ok_hand: Merged as 9dbc50999607833b4b938aeb7b588a7d234ff712. Thanks, @imsodin!

PR #4311, opened 2017-08-21
lib/model: GetIgnores: Don't return error for no .stignore file

imsodin:
GetIgnores returned an error if the folder is not currently in model and there is no ".stignore" file, while this is valid.


calmh:
@st-review merge

st-review:
:ok_hand: Merged as 0ca0e3e9bdc40e5f198f0a4c3e98c855eb3ad3b7. Thanks, @imsodin!

PR #4312, opened 2017-08-21
lib/upnp: Disable confusing messages

AudriusButkevicius:
The Write on Windows to some virtual adapters always times out.
And the invalid ID makes people think something is wrong.

### DIFF
--- a/lib/upnp/upnp.go
+++ b/lib/upnp/upnp.go
@@ -168,7 +168,9 @@ USER-AGENT: syncthing/1.0
 
 	_, err = socket.WriteTo(search, ssdp)
 	if err != nil {
-		l.Infoln(err)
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 606fce09ca4dc06fc91302d0e811926e4196acbe. Thanks, @AudriusButkevicius!

AudriusButkevicius:
@calmh this merged a failed build it seems?

calmh:
Uh, yeah, wtf.

calmh:
@AudriusButkevicius There was a race condition of sorts, as statuses are not visible in the github API until they have been reported (even if they have been tagged as "required") and tc doesn't report status until it starts the build. I've taught the mergebot to understand and wait for "required" statuses now so this should work as expected.

PR #4315, opened 2017-08-21
all: Better subscription lifetime handling

AudriusButkevicius:
Not that it changes anything, but conceptually is the right thing to do.

Usage service is unhandled, but it never seems to stop.


calmh:
@st-review rebuild now that github is back up

AudriusButkevicius:
Fuck it.

Issue #4305, opened 2017-08-14
Android: Sync conflict not detected: newest copy is propagated to other clients


tomtom:
Let's say I have three devices (1 linux [A] + 2 android [B, C] if this is relevant). All devices are in sync. I switch off syncthing on all 3 devices.

1. I edit file X on device B, then C, then A. So file X has changed on all 3 devices.
2. I switch on syncthing on B, C, A.
3. I wait
4. After a while, the version of file X edited last on device A is propagated to B and C. There is not copy of the old version of X on B and C indicating a synchronization conflict.

I personally would expect to end up with at least two files indicating a synchronization conflict.

Is the situation describe in step 4 the expected outcome? Are my expectations about how synchting works wrong?

calmh:
It is expected that you would get sync conflicts. I'm 99% sure that 

> 2 android [B, C] if this is relevant

is the problem as Android doesn't keep proper mtime and we have a bunch of hacks around that. I thought those hacks worked, but there may be something else interacting here as well maybe.

calmh:
I tested your scenario manually with three instances, one file changed on all three sides while Syncthing is off, then starting all three and got the expected result:

```
-rw-r--r--   1 jb  staff   36 Aug 17 12:00 test-appendfile
-rw-r--r--   1 jb  staff   36 Aug 17 12:00 test-appendfile.sync-conflict-20170817-120105-373HSRP
-rw-r--r--   1 jb  staff   36 Aug 17 12:00 test-appendfile.sync-conflict-20170817-120105-MRIW7OK
```

So I'm fairly sure this is specific to either Android in general or your test on Android in particular.

tomtom:
Does the DB also contain info on file sizes? If so, a file could be considered unchanged only if the mtime is the same (or unknown on Android?) _and_ if the file size hasn't changed. This wouldn't get mere replacements but that's better than the current situation.

calmh:
The check is done on mtime, size and permissions (already).

AudriusButkevicius:
So I think this is fixed in the RC.

calmh:
I doubt it as the description includes stopping syncthing, doing the modification everywhere, then starting syncthing. The scan should unconditionally happen when it starts.

AudriusButkevicius:
@tomtom can you please test the scenario again once android version is released?

tomtom:
It tried the following (similar) scenario: 3 devices (1 linux [A] + 2 android [B, C]).

[...]

PR #4317, opened 2017-08-21
lib/model: Move stale scan check info finisher (ref #4305, fix #3742)

### LINKED ISSUE #3742: lib/model: Newly created conflicting files can be silently overwritten

AudriusButkevicius:
Defers the "hey stuff just appeared" check to finished, which I think potentially addresses both issues, but not sure.

### DIFF
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -1002,27 +1002,6 @@ func (f *sendReceiveFolder) handleFile(file protocol.FileInfo, copyChan chan<- c
 
 	tempName := ignore.TempName(file.Name)
 
-	if hasCurFile && !curFile.IsDirectory() && !curFile.IsSymlink() {
-		// Check that the file on disk is what we expect it to be according to
-		// the database. If there's a mismatch here, there might be local
-		// changes that we don't know about yet and we should scan before
-		// touching the file. If we can't stat the file we'll just pull it.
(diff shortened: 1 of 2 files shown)


calmh:
@st-review merge

st-review:
:ok_hand: Merged as b8c249cddcaad1e22e19adeabd850ffae7799b75. Thanks, @AudriusButkevicius!

PR #4323, opened 2017-08-23
sync percentage process max is 100%

SCys:
### Purpose

sync percentage process max is 100%.


calmh:
I'd rather we found and fixed the cases where we report that we have more data in sync than actually exists...

AudriusButkevicius:
Agree

SCys:
i will try other way change the ui show progress percentage.
origin code show large than 231%.

calmh:
If you can reproduce this, please file the issue with the steps and we can fix that instead.

PR #4325, opened 2017-08-23
gui: Round down in devices completion (consistent with folders)

imsodin:
The completion percentage is already rounded down to avoid the potentially confusing situation where the status is "Syncing" but the completion 100%. This PR does the same for the devices.


calmh:
@st-review merge

st-review:
:ok_hand: Merged as c7221b035d94d459fe8bbe1e88df67b7d320b216. Thanks, @imsodin!

Issue #4245, opened 2017-07-04
relay server eats ram


khumarahn:
Hi, 

continuing https://forum.syncthing.net/t/relay-server-ram-memory-usage/10086/6 here.

I stopped the relay server with -SIGQUIT and here is the log file:
https://163.172.132.71/strelaysrv.log

AudriusButkevicius:
The relevant leaking bit:

```
goroutine 465377 [IO wait, 309656 minutes]:
net.runtime_pollWait(0xb31ea2c8, 0x72, 0x131af000)
	/usr/lib/go/src/runtime/netpoll.go:160 +0x60
net.(*pollDesc).wait(0x1422f8f8, 0x72, 0x0, 0x0)
	/usr/lib/go/src/net/fd_poll_runtime.go:73 +0x34
net.(*pollDesc).waitRead(0x1422f8f8, 0x0, 0x0)
	/usr/lib/go/src/net/fd_poll_runtime.go:78 +0x30
net.(*netFD).Read(0x1422f8c0, 0x131af000, 0x1000, 0x1000, 0x0, 0x4f2a20, 0x10872004)
	/usr/lib/go/src/net/fd_unix.go:243 +0x1f0
net.(*conn).Read(0x10abfc30, 0x131af000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
	/usr/lib/go/src/net/net.go:173 +0xb8
bufio.(*Reader).fill(0x14d40510)
	/usr/lib/go/src/bufio/bufio.go:97 +0x1c4
bufio.(*Reader).Read(0x14d40510, 0x11b2f400, 0x400, 0x400, 0x0, 0x0, 0x0)
	/usr/lib/go/src/bufio/bufio.go:209 +0x22c
github.com/syncthing/syncthing/lib/tlsutil.(*UnionedConnection).Read(0x1537c4c0, 0x11b2f400, 0x400, 0x400, 0x244, 0x0, 0x0)
	/var/tmp/portage/net-p2p/syncthing-0.14.11/work/syncthing-0.14.11/src/github.com/syncthing/syncthing/lib/tlsutil/tlsutil.go:153 +0x6c
crypto/tls.(*block).readFromUntil(0x1453d260, 0xb5545218, 0x1537c4c0, 0x5, 0x0, 0x0)
	/usr/lib/go/src/crypto/tls/conn.go:476 +0xc8
[...]

PR #4335, opened 2017-08-31
lib/tlsutil: Remove undesired bufio from UnionedConnection (ref #4245)

calmh:
### Purpose

This reduces memory pressure on the relay server by removing an unnecessary bufio.Reader on every connection. We used the buffered reader to peak at the first byte and deduce if it was a TLS connection or not, but then we needed to keep it around forever. Now, instead, read one byte explicitly and return that as the first real read from the client, then just send the reads to the conn directly.

### Testing

I've added a unit test! Also, manually it still works to talk both HTTP and HTTPS to Syncthing.


lib/tlsutil/tlsutil.go
```diff
@@ -130,27 +128,36 @@ func (l *DowngradingListener) AcceptNoWrapTLS() (net.Conn, bool, error) {
 		return nil, false, err
 	}
 
-	br := bufio.NewReader(conn)
+	var first [1]byte
```
calmh: (Making this a `[1]byte` instead of just a `byte` is more convenient as we can `Read()` into it)

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 4b6e7e7867047eb41e63b6798c401c10675d9cec. Thanks, @calmh!

PR #4338, opened 2017-09-01
lib/model: make puller retrying and logging less aggressive

imsodin:
Currently all errors during pulling and the first of these errors again on finishing are logged to info. Besides that the errors logged when finishing are stored in f.errors. This PR moves all logging during pulling to the debug channel (they might still be relevant in some obscure debugging case) and uses the stored errors to log the main error per fail when all pulling iterations are done and failed.

Additional instead of trying 11 times it now only tries 3 times.

This is the first part of what is discussed here:
https://forum.syncthing.net/t/reduce-verboseness-of-puller/10261

### DIFF
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -239,20 +239,23 @@ func (f *sendReceiveFolder) Serve() {
 					break
 				}
 
-				if tries > 10 {
+				if tries > 2 {
 					// We've tried a bunch of times to get in sync, but
 					// we're not making it. Probably there are write
 					// errors preventing us. Flag this with a warning and
 					// wait a bit longer before retrying.
-					l.Infof("Folder %q isn't making progress. Pausing puller for %v.", f.folderID, f.pause)
-					l.Debugln(f, "next pull in", f.pause)
-
 					if folderErrors := f.currentErrors(); len(folderErrors) > 0 {
+						for path, err := range folderErrors {
+							l.Infof("Puller (folder %q, dir %q): %v", f.Description(), path, err)
+						}
 						events.Default.Log(events.FolderErrors, map[string]interface{}{
 							"folder": f.folderID,
 							"errors": folderErrors,
 						})
 					}
+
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
I think we should get rid of the Debugf as we print in the end anyway, and change newError to have a context, so that we have:

```
newError(file.Name, "shortcut", err)
newError(file.Name, "rename", err)
newError(file.Name, "delete", err)
```

imsodin:
Just to be sure about this: We don't log everything in the end: We only log from the last iteration and we only log the first error occurring per iteration (there may be more, but newError drops them). I feel like dropping this information is fine, as this code is pretty "old" as in there will not likely come up weird stuff that requires this level of verboseness. So this is a judgment call about removing some (very little) information, it is not completely redundant.

Adding context sounds useful.

AudriusButkevicius:
So I'd be fine with dropping the Debug statements, I guess I'll leave it for @calmh to decide.

calmh:
I agree about the context to newError to make it maybe have more info.

(I also think we should adopt github.com/pkg/errors with errors.Wrap in general, which would have given us this for free, maybe even without that we can just decorate the errors along the way.)

I think this is fine as is and the rest is room for future improvements.

calmh:
@st-review merge it

lib/model: Make puller retrying and logging less aggressive

Currently all errors during pulling and the first of these errors again on finishing are logged to info. Besides that the errors logged when finishing are stored in f.errors. This PR moves all logging during pulling to the debug channel (they might still be relevant in some obscure debugging case) and uses the stored errors to log the main error per fail when all pulling iterations are done and failed.

Additional instead of trying 11 times it now only tries 3 times.

This is the first part of what is discussed here:
https://forum.syncthing.net/t/reduce-verboseness-of-puller/10261

st-review:
:ok_hand: Merged as 55e0ac3e2435836c43e1c567e992e442391e962c. Thanks, @imsodin!

PR #4339, opened 2017-09-01
lib/connetions: Use our own fork of kcp (fixes #4063)

### LINKED ISSUE #4063: KCP doesn't reconnect after connection is broken
Trivial to reproduce:

Start up s1 and s2 (`cd test; syncthing -home h1` etc) and let them connect. Restart one side. The other will time out the connection after ~ eight minutes - this is expected, as we don't have any `RST`-equivalent in this protocol (although I think the interval is unexpectedly long).

After that they never manage to connect again. To get them to be able to connect, the other side must also be restarted.

This has nothing to do with NAT etc as these are two devices with hardcoded addresses to each other on the same host. Presumably it's about some sort of connection state that is never cleared out.

0.14.26-rc.1

AudriusButkevicius:
This updates kcp and uses our own fork which:

1. Keys sessions not just by remote address, but by remote address + conversation id
2. Allows not to close connections that were passed directly to the library.
3. Resets cache key if the session gets terminated.

Please run this on your setup and see if you still have issues.


calmh:
This updates kcp and uses our own fork which:

  1.  Keys sessions not just by remote address, but by remote address + conversation id
  2.  Allows not to close connections that were passed directly to the library.
  3.  Resets cache key if the session gets terminated.

calmh:
@st-review lgtm apart from gofmt and typo in subject "connetions"

st-review:

calmh:
Hah. I can push to your branches when they are part of a PR on Syncthing. I didn't know that. So the

screen shot 2017-09-02 at 07 54 24

is literally true. Neat.

calmh:
@st-review rebuild

calmh:
@st-review merge it

lib/connections: Use our own fork of kcp (fixes #4063)

This updates kcp and uses our own fork which:

1. Keys sessions not just by remote address, but by remote address + conversation id
2. Allows not to close connections that were passed directly to the library.
3. Resets cache key if the session gets terminated.

st-review:

st-review:
:ok_hand: Merged as cbcc3ea132e1374e43296564900e3b9d4d1c896a. Thanks, @AudriusButkevicius!

calmh:
I've moved my main server pair to kcp only again and will see how it behaves!

PR #4340, opened 2017-09-01
lib/model: Trigger pulls instead of pulling periodically

imsodin:
This changes the current behaviour of pulling periodically and aborting early when the sequence number hasn't changed to only pull when triggered by a call to `SchedulePull()`.


lib/model/model.go
```diff
@@ -1780,6 +1768,10 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 		return err
 	}
 
+	defer func() {
+		runner.SchedulePull()
+	}()
```
AudriusButkevicius: `defer runner.SchedulePull()`

lib/model/model.go
```diff
@@ -1748,18 +1748,6 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
...
-	// request things we might have suddenly become unignored and so on.
-
-	oldHash := ignores.Hash()
-	defer func() {
-		if ignores.Hash() != oldHash {
-			l.Debugln("Folder", folder, "ignore patterns changed; triggering puller")
-			runner.IndexUpdated()
-		}
-	}()
-
```
AudriusButkevicius: How is this compensated?

lib/model/model.go
AudriusButkevicius: Also, what's the point of pulling after a scan?
Just to do the sequence checks?

lib/model/rwfolder.go
```diff
@@ -194,13 +172,11 @@ func (f *sendReceiveFolder) Serve() {
 			curSeq, ok := f.model.RemoteSequence(f.folderID)
 			if !ok || curSeq == prevSec {
 				l.Debugln(f, "skip (curSeq == prevSeq)", prevSec, ok)
-				f.pullTimer.Reset(f.sleep)
 				continue
 			}
 
 			if err := f.model.CheckFolderHealth(f.folderID); err != nil {
 				l.Infoln("Skipping pull of", f.Description(), "due to folder error:", err)
-				f.pullTimer.Reset(f.sleep)
```
AudriusButkevicius: What's going to trigger this now, given the timer is not doing it anymore?

lib/model/rwfolder.go
```diff
@@ -254,7 +230,9 @@ func (f *sendReceiveFolder) Serve() {
 						})
 					}
 
-					f.pullTimer.Reset(f.pause)
+					time.AfterFunc(f.pause, f.SchedulePull)
+					f.increasePause()
```
AudriusButkevicius: Given this function is only called once, I'd put it's code here for clarity and to remove indirection.

lib/model/model.go
imsodin: By https://github.com/syncthing/syncthing/pull/4340/files/d06035c89cab60277cf849c03aff2ddf2395d754#diff-1df9805b2381b6288b220226f82d5dc9R1771 and then https://github.com/syncthing/syncthing/pull/4340/files/d06035c89cab60277cf849c03aff2ddf2395d754#diff-feb9fde15f8dd977fe9b279a36a447a2R163

lib/model/model.go
imsodin: E.g. for the scenario where puller is failing due to a file1 (whatever reason, e.g. permission error). User notices this in logs, resolves it manually and the scan pick the change up. Generally it changes the index so check that there is nothing changed in our needs.

lib/model/rwfolder.go
imsodin: A call to SchedulePull, i.e. an index update, a scan or a change in cluster config. The point of this PR is to not pull every 10s but only if something changed.

AudriusButkevicius: Sure, but let's say we have an outdated folder that was unmounted, with inotify or something like that that has watches setup for an unmounted folder. We mount the folder, the inotify is somewhat useless, never triggers a scan (as fd's point at the wrong directory), how would we know a directory is alive again?

imsodin: If a folder is stopped, it gets started -> scanned. If a folder is in state error, this error can only be removed during a scan.

AudriusButkevicius: Right, ok.

lib/config/folderconfiguration.go
```diff
@@ -44,6 +43,7 @@ type FolderConfiguration struct {
 
 	DeprecatedReadOnly       bool    `xml:"ro,attr,omitempty" json:"-"`
 	DeprecatedMinDiskFreePct float64 `xml:"minDiskFreePct,omitempty" json:"-"`
+	DeprecatedPullerSleepS   int     `xml:"pullerSleepS" json:"pullerSleepS"`
```
AudriusButkevicius: I think remove this, as it's not used anywhere.

AudriusButkevicius:
@calmh I'll leave this to you, as this is complicated.

imsodin:
Also it is quite a bit lower priority than #4426 to me, so if time is scarce, I would prefer if that got handled instead.

calmh:
Let's not do everything at once. We have one intrusive change to the core functionality in already, I'd like to get that out into an rc and have it debugged for a while before adding more oddities into the mix. That goes for 4426 as well.

AudriusButkevicius:
We are screwing it up anyway, so I'd do it all in one go, and cut a 0.15.

imsodin:
Can't we have the best of two worlds and get a new RC out - then master can run ahead if we have capacity beside patching up KCP and filesystem notifications in the RC? Is there still something blocking the release of the RC?

AudriusButkevicius:
Usage report server hasn't been brought up to scratch yet, other than that, perhaps @calmh should be the test rat for a day or two, and we should unleash it.

calmh:
Full negatory on screwing everything up and going for a 0.15. This is a regular 0.14 release once the usage reporting stuff is in. And yes, we can merge more stuff once the RC is out the door.

lib/model/rwfolder.go
```diff
@@ -305,6 +213,99 @@ func (f *sendReceiveFolder) String() string {
...
+	default:
+		// Once the initial scan finished, a pull will be scheduled
+		return prevSec, prevIgnoreHash, true
+	}
+
+	f.model.fmut.RLock()
+	curIgnores := f.model.folderIgnores[f.folderID]
+	f.model.fmut.RUnlock()
+
+	if newHash := curIgnores.Hash(); newHash != prevIgnoreHash {
```
calmh: Can I ask to have the variable naming tweaked a bit? This is not your fault, it's historical, but might as well fix it while we're anyway poking round.

 - The method signature here is complex enough (three return values) that I think it would add clarity to name the return values. I *don't* suggest using bare returns or anything - just giving the return values names so it's clear what they are from looking at the method signature.

- The `prevSec` variable/parameter is misspelled, it should be `prevSeq` as it's about a sequence number and not a time or something.

- The method *takes* `prevSeq` and `prevIgnoreHash` but the values it calculates and compares them to in the method should be `curSeq` and `curIgnoreHash`, and those should be what it returns. It's confusing to return `prevSeq` etc at the end of the method, as if those were the same things we got in.

calmh:
Looks reasonable to me. Is it tested somehow? Have you run it? :)

lib/model/rwfolder.go
imsodin: Done. The `curSeq` feels a bit weird, as it is resulted in `curSeq = prevSeq`, 's/curSeq/remoteSeq/g` and 's/prevSeq/curSeq/g` - but I guess it is clearer.

imsodin:
Nah, this hasn't seen daylight yet - now that it has been called "reasonable" I will take it for a spin.

lib/config/folderconfiguration.go
imsodin: Ok. Just for my information: What is the purpose of keeping any of these deprecated variables?

calmh: Usually we keep them to be able to migrate old configuration. If there is nothing to migrate we don't need to keep it.

lib/model/rwfolder.go
```diff
@@ -308,6 +219,99 @@ func (f *sendReceiveFolder) String() string {
...
+
+	for {
+		tries++
+
+		changed = f.pullerIteration(curIgnores)
+		l.Debugln(f, "changed", changed)
+
+		if changed == 0 {
+			// No files were changed by the puller, so we are in
+			// sync. Remember the local version number and
```
calmh: Comment out of date

lib/model/rwfolder.go
```diff
@@ -308,6 +219,99 @@ func (f *sendReceiveFolder) String() string {
...
+			// schedule a resync a little bit into the future.
+
+			if lv, ok := f.model.RemoteSequence(f.folderID); ok && lv < remoteSeq {
+				// There's a corner case where the device we needed
+				// files from disconnected during the puller
+				// iteration. The files will have been removed from
+				// the index, so we've concluded that we don't need
+				// them, but at the same time we have the local
```
calmh: This one too I think

lib/model/rwfolder.go
```diff
@@ -308,6 +219,99 @@ func (f *sendReceiveFolder) String() string {
...
+				remoteSeq = lv
+			}
+			curSeq = remoteSeq
+
+			f.pause = f.basePause()
+
+			break
+		}
+
+		if tries > 2 {
```
calmh: How about naming the 2 somewhere

lib/model/rwfolder.go
imsodin: I am not entirely sure I understand the corner case describe here correctly. It seems to be about a device disconnecting during the pull, which can still happen with this PR.

imsodin: Never mind, you are probably just referring to the `curVer` usage that I missed.

imsodin:
@calmh Now fully lgtm?

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as f6ea2a7f8e5077b7b279cd680400509bccbf557e. Thanks, @imsodin!

PR #4341, opened 2017-09-02
lib: Folder marker is now a folder

AudriusButkevicius:
So we can store all sorts of stuff there.
Also moves some functions to fs package.

### DIFF
--- a/lib/config/testdata/v23.xml
+++ b/lib/config/testdata/v23.xml
@@ -1,0 +1,16 @@
+<configuration version="22">
+    <folder id="test" path="testdata" type="readonly" ignorePerms="false" rescanIntervalS="600" autoNormalize="true">
+        <filesystemType>basic</filesystemType>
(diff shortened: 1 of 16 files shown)


calmh:
@st-review lgtm from the phone, someone else to agree or I'll merge it from a computer later

st-review:

calmh:
@st-review merge

st-review:
:ok_hand: Merged as ab132ff6fe056324a023d89324350427e14c09e0. Thanks, @AudriusButkevicius!

lib/config/testdata/v23.xml
```diff
@@ -0,0 +1,16 @@
+<configuration version="22">
```
imsodin: Forgot to increment number. Shouldn't a test pick this up?

AudriusButkevicius: It doesn't matter, the test does the migration, we just check that things after the migration are in the right places where we expect them to be.

calmh: and specifically that test is pointless busywork when the config format hasn't changed. it was good the first few iterations and we changed things like how we represent devices etc.

Issue #3628, opened 2016-09-26
Usage reporting update


calmh:
I'd like to make some additions to the anonymous usage data. Since this will require re-asking for permission, and thus likely losing a significant base of our current reporters, we should make it count. I want the following data:
### Ignore patterns
- [ ] Number of patterns (i.e., lines in the ignore file plus all included files).
- [x] Whether `#include` is used or not

We report one data point, the max of any folder, and the "or" of the "uses includes" flags.
### Connection statistics
- [x] Direct vs relay connections. I want to gauge how much devices are connected over the various connection types we have. For this purpose I think the metric we want is something like connection-minutes, not connection events or bytes transferred. 
- [ ] Max number of established connections. I.e., how many devices are we connected to at most, reset per connection report.
### General data
- [ ] Uptime. (Syncthing's, not the system's.)
- [ ] Has used external scanner, i.e. inotify? Just a boolean that gets set when the REST method is called. 
- [ ] CPU utilization, as some sort of max, average or percentage...
### GUI settings
- [x] localhost only listening or wider?
- [x] Password set or not?
- [x] All the various booleans (TLS, skip host check, ...)
- [x] Selected theme
### Puller stats
- Separate ticket on what we should track here because it will require some implementation on its own. But generally speaking data on pulled vs reused blocks and stuff like that. 

#3806

uok:
That's a lot of (additonal) statistics... Do you plan to remove features/functions if they do not show up as used?

uok:
That's a lot of (additonal) statistics... Is there any part of the software left that is not tracked? 
Do you plan to remove features/functions if they do not show up as used?

AudriusButkevicius:
memory usage

Ferroin:
Memory usage is already reported.

AudriusButkevicius:
Yes, I am a muppet.

calmh:
@uok is there something in these metrics that you see as a concern in collecting?

nrm21:
Direct vs relay connections:
Might want to check first if at least one one the clients in a connection has manually disabled relay.  No point in skewing the stats with a lot of obvious "100% direct" entries of those.

Or maybe you do... if those are the kinda stats you're looking for... but I'm guessing not.

Ferroin:
For CPU usage:
[...]

PR #4347, opened 2017-09-03
cmd/syncthing: Add more stats to usage reports (ref #3628)

AudriusButkevicius:
Tested manually.
Not sure if we need to bump the version here.


AudriusButkevicius:
Apart from tests, what else needs to be done here @calmh?

calmh:
I think either implement what we discussed on IRC about retaining the previous report version until the new one is accepted, or a huge push to get all the stuff we really want into the report in in one swoop... Not necessarily in this PR of course, but before we let this into the wild...

AudriusButkevicius:
So added incremental support, but did it without paying too much attention or thinking about it.
Tested here and there, seems to work, but give it a proper review pls.

cmd/syncthing/usage_report.go
```diff
@@ -264,27 +234,44 @@ func (s *usageReportingService) sendUsageReport() error {
...
 	}
 }
 
+func (s *usageReportingService) VerifyConfiguration(from, to config.Configuration) error {
+	return nil
+}
+
+func (s *usageReportingService) CommitConfiguration(from, to config.Configuration) bool {
+	if from.Options.URAccepted != to.Options.URAccepted || from.Options.URUniqueID != to.Options.URUniqueID || from.Options.URURL != to.Options.URURL {
+		s.timer.Reset(time.Duration(to.Options.URInitialDelayS) * time.Second)
```
calmh: I suspect this is a data race on `s.timer`

AudriusButkevicius: Why? s.timer is not touched once it's initialized.

AudriusButkevicius:
@st-review retest

st-review:
I'm sorry, @AudriusButkevicius. I'm afraid I don't know what you mean. I know how to `don't`, `lgtm`, `merge`, `prevent`, `rebuild`, `squash`, `stop`.

AudriusButkevicius:
@st-review rebuild

calmh:
It's used by Serve() and reset by the config commit.

On 10 Sep 2017, at 11:14, Audrius Butkevicius <notifications@github.com<mailto:notifications@github.com>> wrote:


@AudriusButkevicius commented on this pull request.

________________________________

In cmd/syncthing/usage_report.go<https://github.com/syncthing/syncthing/pull/4347#discussion_r137944522>:

>               }
        }
 }

+func (s *usageReportingService) VerifyConfiguration(from, to config.Configuration) error {
+       return nil
+}
+
+func (s *usageReportingService) CommitConfiguration(from, to config.Configuration) bool {
+       if from.Options.URAccepted != to.Options.URAccepted || from.Options.URUniqueID != to.Options.URUniqueID || from.Options.URURL != to.Options.URURL {
+               s.timer.Reset(time.Duration(to.Options.URInitialDelayS) * time.Second)


Why? s.timer is not touched once it's initialized.

AudriusButkevicius:
Sure but that's just reads, which means it should be fine?

calmh:
It's not, but don't take my word for it

```go
package main

import (
	"fmt"
	"testing"
	"time"
)

func TestTimer(t *testing.T) {
	test := &test{
		t: time.NewTimer(0),
	}
	go test.Serve()
	for i := 0; i < 5; i++ {
		test.Commit()
		time.Sleep(time.Second)
	}
}

type test struct {
	t *time.Timer
}

func (t *test) Serve() {
	t.t.Reset(0)
	for {
		<-t.t.C
		fmt.Println("tick")
		t.t.Reset(time.Second)
	}
}

func (t *test) Commit() {
	fmt.Println("resetting timer")
	t.t.Reset(10 * time.Millisecond)
}
```

```text
jb@unu:~/t/timer $ go test -race
resetting timer
tick
tick
==================
WARNING: DATA RACE
Write at 0x00c420088650 by goroutine 6:
  time.(*Timer).Reset()
      /usr/local/go1.9/src/time/sleep.go:127 +0x98
  _/Users/jb/tmp/timer.(*test).Commit()
      /Users/jb/tmp/timer/timer_test.go:35 +0xb1
  _/Users/jb/tmp/timer.TestTimer()
      /Users/jb/tmp/timer/timer_test.go:15 +0xb8
  testing.tRunner()
      /usr/local/go1.9/src/testing/testing.go:746 +0x16c

Previous write at 0x00c420088650 by goroutine 8:
  time.(*Timer).Reset()
      /usr/local/go1.9/src/time/sleep.go:127 +0x98
  _/Users/jb/tmp/timer.(*test).Serve()
      /Users/jb/tmp/timer/timer_test.go:25 +0x59

```

AudriusButkevicius:
Is that a race internally setting something in the timer or accessing the timer in general.
As in, if I wrap things in a lock, do I need to wrap the channel retrieval with a lock too?

calmh:
I think it's just the Reset() that does internal unlocked writes, so protecting that should be enough. There's a whole discussion in the Reset() documentation but I think most of it is not relevant as we don't really care if timer fires at exactly the same time we call Reset()

AudriusButkevicius:
Anyway, this uses a channel now.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -133,7 +135,11 @@ angular.module('syncthing.core')
             }).error($scope.emitHTTPError);
 
             $http.get(urlbase + '/svc/report').success(function (data) {
-                $scope.reportData = data;
+                $scope.reportDataPreview = $scope.reportData = data;
+                if ($scope.system && $scope.config.options.urSeen < $scope.system.urVersionMax) {
+                    // Usage reporting format has changed, prompt the user to re-accept.
+                    $('#ur').modal();
+                }
```
AudriusButkevicius: This is a bit racy, as we rely on $scope.system being populated first.

calmh:
The block stats and connection types objects are shown also in the version two preview:

screen shot 2017-09-20 at 08 28 28

AudriusButkevicius:
Fixed, I hope.

https://stackoverflow.com/questions/31200906/custom-child-directive-accessing-scope-of-parent

gui/default/syncthing/usagereport/usageReportPreviewModalView.html
```diff
@@ -6,7 +6,7 @@
     <p translate>The aggregated statistics are publicly available at the URL below.</p>
     <p><a href="https://data.syncthing.net/" target="_blank">https://data.syncthing.net/</a></p>
     <label translate>Version</label>
-    <select class="form-control" ng-model="reportDataPreviewVersion" ng-change="refreshReportDataPreview()">
+    <select id="urPreviewVersion" class="form-control" ng-model="$parent.$parent.reportDataPreviewVersion" ng-change="refreshReportDataPreview()" >
```
calmh: ... 😷

cmd/syncthing/usage_report.go
```diff
@@ -234,13 +234,14 @@ func (s *usageReportingService) sendUsageReport() error {
 
 func (s *usageReportingService) Serve() {
 	s.stop = make(chan struct{})
-	s.timer.Reset(time.Duration(s.cfg.Options().URInitialDelayS) * time.Second)
-
+	t := time.NewTimer(time.Duration(s.cfg.Options().URInitialDelayS) * time.Second)
```
calmh: defer t.Stop()

gui/default/syncthing/core/syncthingController.js
```diff
@@ -133,7 +135,11 @@ angular.module('syncthing.core')
             }).error($scope.emitHTTPError);
 
             $http.get(urlbase + '/svc/report').success(function (data) {
-                $scope.reportData = data;
+                $scope.reportDataPreview = $scope.reportData = data;
+                if ($scope.system && $scope.config.options.urSeen < $scope.system.urVersionMax) {
```
calmh: This might not work for candidate releases. For those we set `urAccepted = urVersionMax` (or the equivalent what the variables are a called) on startup, but we don't tweak `urSeen` so this dialog would be triggered unnecessarily.

AudriusButkevicius:
Updated and added usage stats for features of the ignore patterns.

lib/model/model.go
```diff
@@ -406,25 +406,102 @@ func (m *Model) RestartFolder(cfg config.FolderConfiguration) {
...
+					} else {
+						seenPrefix[0] = false
+						seenPrefix[1] = false
+						seenPrefix[2] = false
+						break
+					}
+				}
+
+				// Noops, remove
+				if strings.HasPrefix(line, "**") {
```
imsodin: I don't think doube star prefixes are noops, e.g. `**/temp` isn't rooted, is it?

AudriusButkevicius:
I need to remove **/ I guess as that is equivalent behaviour without the prefix.

AudriusButkevicius:
@st-review rebuild

calmh:
This is good to go I think?

AudriusButkevicius:
Yes

calmh:
@st-review merge

st-review:
:ok_hand: Merged as 2760d032ca672435e06a1eeb7832f5fe1af14f92. Thanks, @AudriusButkevicius!

AudriusButkevicius:
I assume this need schema changes in the db?

calmh:
Yes, it needs handling on the receiving side. I think we have a couple of migrations in that code already, basically `ALTER TABLE ... ADD COLUMN` and new fields in the struct it unmarshals to etc

calmh:
I guess if you're looking at a new challenge maybe refactor it to use a json document type in the database so that we'd just need to hack the various analytics parts in the future and not the whole storage part

PR #4349, opened 2017-09-04
cmd/syncthing, lib/sha256: Skip CPU benchmarks when user decided (fixes #4348)

### LINKED ISSUE #4348: Hashing Benchmarks slow down Startup Considerably
I just did some tests to see how we can speedup the Syncthing start on Android. For that, I added log statements around the following code block in main.go:
```
sha256.SelectAlgo()
sha256.Report()
perfWithWeakHash := cpuBench(3, 150*time.Millisecond, true)
l.Infof("Hashing performance with weak hash is %.02f MB/s", perfWithWeakHash)
perfWithoutWeakHash := cpuBench(3, 150*time.Millisecond, false)
l.Infof("Hashing performance without weak hash is %.02f MB/s", perfWithoutWeakHash)
```

According to the logs, this takes more than 3 seconds to run on my Nexus 5. When I commented these lines out and hardcoded `weakhash.Enabled = false`, the app was available almost instantly.

From what I can see, there is no way to disable the benchmarks via parameters or config. So there are a few options:
- add an option, to disable all benchmarks (via environment variable or parameter)
- run the benchmark only on first start, and cache the result as long as the CPU doesn't change (but that will still slow down the initial start)
- optimize the benchmark, so that it runs much faster
- run the benchmark asynchronously, so it doesn't block startup (but that would still be a waste of cpu/battery)

### Version Information

Syncthing Version: v0.14.32
OS Version: Android 6

calmh:
### Purpose

When STHASHING is set, don't benchmark as it's already decided. If weak
hashing isn't set to "auto", don't benchmark that either.

### Testing

I haven't even had coffee yet.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -638,12 +638,10 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 	l.Infoln(LongVersion)
 	l.Infoln("My ID:", myID)
 
+	// Select SHA256 implementation and report. Affected by the
+	// STHASHING environment variable.
 	sha256.SelectAlgo()
 	sha256.Report()
-	perfWithWeakHash := cpuBench(3, 150*time.Millisecond, true)
-	l.Infof("Hashing performance with weak hash is %.02f MB/s", perfWithWeakHash)
-	perfWithoutWeakHash := cpuBench(3, 150*time.Millisecond, false)
-	l.Infof("Hashing performance without weak hash is %.02f MB/s", perfWithoutWeakHash)
 
 	// Emit the Starting event, now that we know who we are.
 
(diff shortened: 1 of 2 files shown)


Nutomic:
Just tested this on Android with `STHASHING=minio` `<weakHashSelectionMethod>never</weakHashSelectionMethod>`, hashing is disabled and Syncthing starts in ~300ms :+1:

calmh:
@syncthing/core I think this is uncontroversial?

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 4d3e0de4ba14d49667fe5442693981fc7272cefe. Thanks, @calmh!

PR #4350, opened 2017-09-04
lib/ignore: discard duplicate lines in .stignore

imsodin:
As the first match determines the result when ignoring, duplicate lines have no effect. Therefore these are now discarded when parsing the file.

### DIFF
--- a/lib/ignore/ignore.go
+++ b/lib/ignore/ignore.go
@@ -132,18 +132,12 @@ func (m *Matcher) Load(file string) error {
 		return nil
 	}
 
-	fd, err := m.fs.Open(file)
+	fd, info, err := loadIgnoreFile(m.fs, file, m.changeDetector)
 	if err != nil {
 		m.parseLocked(&bytes.Buffer{}, file)
 		return err
 	}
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
@st-review lgtm

st-review:

calmh:
This could be a teensy tiny bit surprising if I have some patterns in my ignore file, then add an include of common files to ignore (at the top, lets say), and next time I open my ignore patterns half of them are gone (because they are also present in the included file, which is invisible from the built in editor)?

imsodin:
@calmh This does not remove any lines from the actual file, it just doesn't create the patterns.
https://github.com/syncthing/syncthing/pull/4350/files#diff-a0ccca3b95e35f5785db8ae0886a570dR434

calmh:
@st-review lgtm

lib/ignore: Ignore duplicate lines in .stignore

st-review:
:ok_hand: Merged as c41aaad3bb19d10095700c9c4dd2159da996bb2e. Thanks, @imsodin!

PR #4352, opened 2017-09-06
lib/model: Deduplicate folder loops

imsodin:
Purely (almost) cosmetic to reduce code duplication, remove unnecessary code, handle scanning the same for both folder types and use new terminology (ro/rw -> sendonly/sendreceive) for info logging.
The only difference was that `sendOnlyFolder` called `Reschedule` after the initial scan finished, `sendReceiveFolder` before. Now both do it after.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -71,3 +71,20 @@ func (f *folder) scanSubdirs(subDirs []string) error {
 	}
 	return nil
 }
+
+func (f *folder) scanTimerFired() {
+	err := f.scanSubdirs(nil)
+
+	select {
+	case <-f.initialScanFinished:
+	default:
+		status := "Completed"
+		if err != nil {
+			status = "Failed"
+		}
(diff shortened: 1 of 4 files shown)


AudriusButkevicius:
The special has interval case disappears, I guess this means we force a double rescan in both cases? Why do we need a second scan in the first place?

imsodin:
@AudriusButkevicius The has interval check is simply redundant, as reschedule already does it:
https://github.com/syncthing/syncthing/pull/4352/files#diff-9a2872b93819d219be3f7f23070bff2bR38

AudriusButkevicius:
@st-review lgtm

st-review:

calmh:
@st-review lgtm

st-review:
:ok_hand: Merged as f7d2c58783167d4249e5e547dafa3c8d4030d53a. Thanks, @imsodin!

PR #4361, opened 2017-09-10
cmd/syncthing: Add some common security releated HTTP headers (fixes #4360)

### LINKED ISSUE #4360: XSS-related HTTP headers
Syncthing does not appear to produce XSS-related HTTP headers that are designed to prevent some basic browser-side attacks. ([Clickjacking](https://en.wikipedia.org/wiki/Clickjacking), [XSRF](https://en.wikipedia.org/wiki/Cross-site_request_forgery)) Those should probably be added to prevent RFC1918-style attacks. 

ref: 
* https://www.owasp.org/index.php/Security_Headers
* https://securityheaders.io/

calmh:
These:

1. X-Frame-Options: SAMEORIGIN
2. X-XSS-Protection: 1; mode=block
3. X-Content-Type-Options: nosniff
4. Content-Type: text/html; charset=utf-8

Number 4 we already do of course, but I sprinkled some utf-8 where appropriate.

Numbers 2-3 are uncontroversial and shouldn't do any harm.

Number 1 might be an issue. Potentially for apps that wrap Syncthing (@nutomic, @kozec, @canton7), also I recall from forums that there are people who have their own web pages that load the Syncthing UI in a part of them. I added a new GUI config option `InsecureAllowFrameLoading` to turn off this header.

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -332,7 +332,7 @@ func (s *apiService) Serve() {
 	}
 
 	// Add the CORS handling
-	handler = corsMiddleware(handler)
+	handler = corsMiddleware(handler, guiCfg.InsecureAllowFrameLoading)
 
 	if addressIsLocalhost(guiCfg.Address()) && !guiCfg.InsecureSkipHostCheck {
 		// Verify source host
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 429cc20eb7935b3e3f9ee945a130c3b72c63b50d. Thanks, @calmh!

PR #4390, opened 2017-09-21
Add -device-id command line option (fixes #4387)

### LINKED ISSUE #4387: Feature Request: Add -node-id command line switch
I'm trying to setup syncthing in an automated way and I want to write the configuration myself. This would include all devices and all folders I want to share.

For this I need the node ID of each host. It would be very helpful to have access to the node ID through the command line interface. 

Would it make sense to add

    syncthing -node-id

which will read the configuration it currently has, get the node ID and just print it out.

tobiastom:
This pull request adds a new command line switch: `-device-id`. It will display the device ID of the Syncthing instance.
The new option is documented inside the command line interface.

I did build and test it locally and it displays the device ID, based on the certificates used. If more tests are necessary, please let me know. 

This closes #4387.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -236,6 +236,7 @@ type RuntimeOptions struct {
 	resetDeltaIdxs bool
 	showVersion    bool
 	showPaths      bool
+	showDeviceId   bool
 	doUpgrade      bool
 	doUpgradeCheck bool
 	upgradeTo      string
@@ -301,6 +302,7 @@ func parseCommandLineOptions() RuntimeOptions {
 	flag.BoolVar(&options.doUpgradeCheck, "upgrade-check", false, "Check for available upgrade")
 	flag.BoolVar(&options.showVersion, "version", false, "Show version")
 	flag.BoolVar(&options.showPaths, "paths", false, "Show configuration paths")
+	flag.BoolVar(&options.showDeviceId, "device-id", false, "Show the device ID")
 	flag.StringVar(&options.upgradeTo, "upgrade-to", options.upgradeTo, "Force upgrade directly from specified URL")
 	flag.BoolVar(&options.auditEnabled, "audit", false, "Write events to audit file")
 	flag.BoolVar(&options.verbose, "verbose", false, "Print verbose log output")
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
It should be -node-id or just -id becaise we have -reset-database not -resetdatabase.

This also creates the certificate if it does not exist which I suspect is unintended.

tobiastom:
I'll change the naming. I wasn't sure which one is correct.

Regarding generating a new certificate: for my case this is intended. I could generate and get the ID before I have to start syncthing as a service. 

I can understand that this might not be the desired default behaviour.

AudriusButkevicius:
There is already -generate for that.

tobiastom:
I've updated the pull request to not generate a new certificate. It will report an error now.

tobiastom:
Thanks. I'll update the pull request later today.

cmd/syncthing/main.go
```diff
@@ -390,6 +392,17 @@ func main() {
 		return
 	}
 
+	if options.showDeviceId {
+		cert, err := tls.LoadX509KeyPair(locations[locCertFile], locations[locKeyFile])
+		if err != nil {
+			l.Fatalln("Error reading node ID:", err)
```
imsodin: Also "device ID".

cmd/syncthing/main.go
```diff
@@ -625,7 +638,6 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 	// Ensure that we have a certificate and key.
 	cert, err := tls.LoadX509KeyPair(locations[locCertFile], locations[locKeyFile])
 	if err != nil {
-		l.Infof("Generating ECDSA key and certificate for %s...", tlsDefaultCommonName)
```
imsodin: Why?

cmd/syncthing/main.go
tobiastom: Changed it. Thanks for the reminder.

cmd/syncthing/main.go
tobiastom: That was not intentional. I did add the line again.

calmh:
@st-review rebuild

calmh:
@st-review merge it

Add -device-id command line option (fixes #4387)

st-review:
:ok_hand: Merged as 5aade9a4a5506e5b71b2a9ab863790dca488700e. Thanks, @tobiastom!

tobiastom:
Thank you!

PR #4394, opened 2017-09-23
lib/model: Hide temporary files on Windows while they are in use (fixes #4382)

### LINKED ISSUE #4382: Hide temporary files on Windows
Version 0.14.38 unhid hidden folders on multiple Windows 10 64-bit machines. After the automatic update to 0.14.38 .stfolder, .stignore (when applicable), .stversions (when applicable) were all unhidden in Syncthing folders. Other hidden files in operating system also became unhidden after upgrade. Although unsure if this spread across operating system or just in Syncthing folders because became aware of this when saw lots of ".tmp" files (which were unhidden) in Syncthing folders. Not sure if . (dot) files that found in user profile folder were already unhidden or not.

Not entirely bad thing because learned a new windows command as a result. Running below commands in administrator command prompt solved issue.

ATTRIB +H /s /d C:\.*
ATTRIB +H /s /d C:\*.tmp

A backslash needs to come after C: in above command but is being edited out. Not sure how to force what is written verbatim to be shown.

This occurred on three Windows 10 64-bit machines so expect can be replicated.

### Version Information

Syncthing Version: v0.14.38
OS Version: Windows 10 Professional 64-bit

calmh:
### Purpose

Hide the temp files while we are using them.

### Testing

None whatsoever. I'm just assuming that the Hide() call works and that it's valid to call Hide() on an open file. If someone on Windows could confirm that it would be lovely.

I don't think there are any ways out of a temp file other than via sharedPullerState.finalClose - if there are this would result in files becoming permanently hidden which would be bad.

### DIFF
--- a/lib/model/sharedpullerstate.go
+++ b/lib/model/sharedpullerstate.go
@@ -164,6 +164,9 @@ func (s *sharedPullerState) tempFile() (io.WriterAt, error) {
 		return nil, err
 	}
 
+	// Hide the temporary file
+	s.fs.Hide(s.tempName)
+
 	// Don't truncate symlink files, as that will mean that the path will
 	// contain a bunch of nulls.
 	if s.sparse && !s.file.IsSymlink() {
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
I'd argue to leave the temp files hidden in case of error?

calmh:
Possibly, but then we also need to audit all the paths where we might have a fast path directly to rename and so on, which I don't want to do at the moment. :)

AudriusButkevicius:
@st-review merge

st-review:

st-review:
:ok_hand: Merged as a365ae51c47fa2bad19a0500c277b22f799bf85e. Thanks, @calmh!

Issue #4392, opened 2017-09-22
"Not making progress" error lacks detail


calmh:
We recently cleaned up redundant error messages from the puller. Apparently we cleaned it up so much there are now no errors at all under som circumstances:

```
[2JOAH] 19:02:56 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:07:10 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:12:46 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:15:40 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:17:06 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:18:33 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:21:51 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:26:12 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:30:35 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:33:32 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:35:01 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:36:29 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:39:26 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:42:21 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:45:19 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:46:47 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
[2JOAH] 19:48:14 INFO: Folder "Shared" isn't making progress. Pausing puller for 1m0s.
```

There should be at least one error printed for each time it gives up.

imsodin:
All the logging to info that was changed to debug is accompanied by a call to `f.newError`, so I don't understand why this happens.

[...]

PR #4395, opened 2017-09-23
Small changes

AudriusButkevicius:
Merge unsquashed.

### DIFF
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -1904,48 +1904,69 @@ func TestIssue3164(t *testing.T) {
 
 func TestIssue4357(t *testing.T) {
 	db := db.OpenMemory()
-	m := NewModel(defaultConfig, protocol.LocalDeviceID, "syncthing", "dev", db, nil)
+	cfg := defaultConfig.RawCopy()
+	// Create a separate wrapper not to polute other tests.
+	wrapper := config.Wrap("/tmp/test", config.Configuration{})
+	m := NewModel(wrapper, protocol.LocalDeviceID, "syncthing", "dev", db, nil)
 	m.ServeBackground()
 	defer m.Stop()
-	cfg := m.cfg.RawCopy()
-	m.CommitConfiguration(config.Configuration{}, cfg)
+
+	// Replace kicks off CommitConfiguration in multiple routines which we
+	// have no idea when they get called, so sleep a bit :(
+	// We could make replace blocking for all routines to come back without
+	// holding the lock, but that's a bit much just to make a test.
+	replaceCfg := func(cfg config.Configuration) {
+		t.Helper()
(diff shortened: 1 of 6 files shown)


lib/model/rwfolder.go
```diff
@@ -1645,8 +1633,8 @@ func (f *sendReceiveFolder) newError(path string, err error) {
 	if _, ok := f.errors[path]; ok {
 		return
 	}
-
-	f.errors[path] = err.Error()
+	l.Infof("Puller (folder %q, file %q): %s: %v", f.folderID, path, context, err)
```
imsodin: This should be debug, otherwise it will be logged repeatedly on puller iterations.

imsodin:
While you are at it, you could also change the logging at info level to debug here:
https://github.com/syncthing/syncthing/blob/master/lib/model/sharedpullerstate.go#L222
See this comment for why:
https://github.com/syncthing/syncthing/issues/4392#issuecomment-331467140

lib/model/rwfolder.go
AudriusButkevicius: It shouldn't I thinkn as we check the map before?

imsodin: Ah sorry, sloppy code reading.
Then this loop I recently added isn't need anymore, right?
https://github.com/syncthing/syncthing/blob/master/lib/model/rwfolder.go#L248

AudriusButkevicius: I guess we can keep that and scrap my print.

imsodin: I actually like your print better, as it preserves temporal order ;)

AudriusButkevicius:
@calmh have a look

calmh:
Merged!

lib/model/model_test.go
```diff
@@ -1904,48 +1904,69 @@ func TestIssue3164(t *testing.T) {
...
 	defer m.Stop()
-	cfg := m.cfg.RawCopy()
-	m.CommitConfiguration(config.Configuration{}, cfg)
+
+	// Replace kicks off CommitConfiguration in multiple routines which we
+	// have no idea when they get called, so sleep a bit :(
+	// We could make replace blocking for all routines to come back without
+	// holding the lock, but that's a bit much just to make a test.
+	replaceCfg := func(cfg config.Configuration) {
+		t.Helper()
```
imsodin: This is new in go1.9, so that may create problems for packagers (until this 1.8 worked fine for me). That's not a problem, as we require latest go, but it might be nice to mention it in the release (@calmh).

AudriusButkevicius: Actually this should be removed and the comment should be adjusted.

calmh: Yeah doit. (We build with an old compiler, but don't test with it...)

Issue #3439, opened 2016-07-25
Ship .desktop files


aviau:
Related Debian bug: https://bugs.debian.org/832387

Syncthing should ship ``.desktop`` files.

I'm opening this issue here because we probably want to update the upstream package to also ship ``.desktop`` files.

Suggestion from the Debian bug:
```
One slightly kludgy possibility would be to include two .desktop files:
syncthing.desktop:
    [Desktop Entry]
    Name=Syncthing
    GenericName=Syncthing
    Comment=
    Exec=/usr/bin/syncthing
    Terminal=false
    Type=Application
    X-Gnome-Autostart=true

syncthing-ui.desktop:
    [Desktop Entry]
    Encoding=UTF-8
    Name=Syncthing UI
    Type=Link
    URL=http://localhost:8384
    Icon=www
```

I think I would prefer to merge the two files.

However I would like for that to be compatible with running syncthing with systemd ``syncthing@aviau``, which I think is great.

Any ideas?

calmh:
What is a `.desktop` file and where/why should we ship one?

kozec:
> X-Gnome-Autostart=true

Won't this make Syncthing auto-started by default?

> Terminal=false

... and in background on top of that?

Generally, stuff that doesn't have any sort of GUI doesn't have desktop files. From user perspective, clicking on such icon does nothing. If starting Syncthing daemon from applications menu is desired, at least set Terminal to true, so clicking will open _something_.

aviau:
Documentation/Specification of the .desktop files can be found here: https://specifications.freedesktop.org/desktop-entry-spec/latest/index.html#introduction

They are files that define how applications should appear/behave in menus, etc.

> Generally, stuff that doesn't have any sort of GUI doesn't have desktop files

Exactly, this is why I said that I would prefer merging the two files. The above file was just a suggestion by a user of the package.

I think clicking on the Syncthing should launch the daemon (if its not running) and open the UI in a browser. Do we agree with that?

aviau:
Suggestion:

The desktop entry of Syncthing can use a mechanism to prevent launching multiple daemons for a user. I am sure a quick script can check if there is already a ``syncthing@user`` service running, and then launch the daemon if not. Then we can avoid launching two non-systemd-managed daemons with pid files in ``/var/run/syncthing/syncthing-user.pid``

calmh:
`syncthing -browser-only` starts the GUI.

[...]

Issue #755, opened 2014-09-28
Syncthing Linux: Syslog events


dtouzeau:
Is it possible to send events to the syslog system ?
Thanks.

PR #4398, opened 2017-09-23
chore: gnome-shell autostart

typekpb:
### Purpose

Provides sample for gnome-shell startup integration

### Testing

Has been tested on RHEL 7.4 with Gnome shell.

### DIFF
--- a/etc/linux-gnome-shell/README.md
+++ b/etc/linux-gnome-shell/README.md
@@ -1,0 +1,5 @@
+This directory contains a configuration for running syncthing as user startup application in the gnome-shell Linux environment.
+ 1. Copy the file `synthing.desktop` to `$HOME/.config/autostart/syncthing.desktop`
+
+Log output is sent to syslogd.
+
--- a/etc/linux-gnome-shell/syncthing.desktop
+++ b/etc/linux-gnome-shell/syncthing.desktop
@@ -1,0 +1,9 @@
+[Desktop Entry]
+Version=1.0
+Name=syncthing
+GenericName=syncthing
+Comment=syncthing
+Exec=/usr/bin/syncthing -no-browser -no-restart -logflags=0
+Terminal=false
+Type=Application
+X-Desktop-File-Install-Version=0.22


etc/linux-gnome-shell/syncthing.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Version=1.0
+Name=syncthing
+GenericName=syncthing
+Comment=syncthing
+Exec=/usr/bin/syncthing -no-browser -no-restart -logflags=0
+Terminal=false
+Type=Application
+X-Desktop-File-Install-Version=0.22
```
imsodin: Just out of curiosity and because a google search doesn't bring anything up: What is this option about?

typekpb: @imsodin not sure which option are you referring to, probably `X-Desktop-File-Install-Version`, honestly I don't understand it either, just found a sample at: https://unix.stackexchange.com/questions/209898/running-program-on-startup-fedora-22-2015/228376#228376 and adopted it to syncthing needs

imsodin: As its use it unknown and the file works without it, I would remove it.

imsodin:
Sorry for the huge delay, I thought I already wrote the following:
This is not exclusive to gnome-shell, it works in every desktop environment that is XDG compliant (so virtually every DE). Therefore I would rename the folder to something like "linux-desktop-autostart" and mention gnome-shell as one of the DEs that support this (and maybe mention that it works with any DE compliant with the xdg-spec: https://specifications.freedesktop.org/autostart-spec/latest/ar01s02.html )

etc/linux-gnome-shell/README.md
```diff
@@ -0,0 +1,5 @@
+This directory contains a configuration for running syncthing as user startup application in the gnome-shell Linux environment.
+ 1. Copy the file `synthing.desktop` to `$HOME/.config/autostart/syncthing.desktop`
+
+Log output is sent to syslogd.
```
imsodin: Did you test that? I thought there wasn't any logging by default and that you had to add it to the exec comand if you wanted it.

typekpb: @imsodin I haven't tested that, you're right

So should I end up with (as suggested in #755): `Exec=/usr/bin/syncthing -no-browser -no-restart | logger -t syncthing`?

as logging sounds always like a good idea

etc/linux-gnome-shell/syncthing.desktop
usta: Hello, X-Desktop-File-Install-Version tag comes from "desktop-file-install" application which is under "desktop-file-utils". It is a utility to install, or optionally edit desktop files. That 0.22 is the version number which belongs to in which version of "desktop-file-install" application used. (Currently "desktop-file-utils" package version is 0.23 ) : https://www.freedesktop.org/wiki/Software/desktop-file-utils/

imsodin: Thanks for the input. From that link:
> desktop-file-install: installs a desktop file to the applications directory, optionally munging it a bit in transit.
I assume adding this line is part of the munging, which to me is another reason to not have this line in the distributed desktop file. If the user then installs the desktop file with this utility, it gets added with the correct version, if he just copies it manually, it doesn't.

lkwg82:
What is needed to finish this?

etc/linux-gnome-shell/README.md
imsodin: Sorry, I completely forgot about this.  
I don't think we should do any scripting in distributed files - logger may be available on all xdg compatible linuxes (?) but this will e.g. miss logs going to stderr and 2>&1 only works in a shell, so you had to wrap everything into a `sh -c 'syncthing ...` and I am sure there is more pitfalls.  
Unless there is a clear reference, I would either omit any mentioning of logging or change this to a notice that logs may not be stored at all.

etc/linux-gnome-shell/syncthing.desktop
imsodin: Actually you could use syncthing's own logging at default location for this, i.e. `Exec=/usr/bin/syncthing -no-browser -logfile=""`. In any case I think the `-no-restart` should be dropped as there is no daemon management.

imsodin:
All the above comments still need addressing.  
Also this partly addresses #3439, if you (@typekpb) want to address all of that as well, that would be great, otherwise that will be for a followup-PR.

PR #4402, opened 2017-09-26
add multiple listeners support for the Webui

lstordeur:
### Purpose

As advised in #3983, I have tackled the problem of multiple listeners support in the web interface as a necessary step prior to adding support for UNIX domain sockets.

The configuration format has evolved to a new version `v24` bringing the following enhancements:
- several webui listeners support;
- per-listener properties.

Here is an example of the new configuration format:

Examples
```
    <gui enabled="true" debugging="false">
        <listener tls="true">
          <address>127.0.0.1:8384</address>
	  <insecureAdminAccess/>
	  <insecureSkipHostCheck/>
        </listener>
        <listener>
          <address>127.0.0.1:44444</address>
	  <insecureAllowFrameLoading/>
        </listener>
        <apikey>BLtGw3CM9AnBSBaISeq0FPdHCzaY5zG5</apikey>
        <theme>dark</theme>
    </gui>
```

### Testing

- automatic migration from v23 to v24 has been tested
- basic test with two listeners on the loopback interface listening of a couple of different ports (one being TLS-enabled, the other not).

### Documentation

The documentation will have to be updated to reflect this new feature.


lib/config/guiconfiguration.go
```diff
@@ -13,20 +13,20 @@ import (
...
-	Password                  string `xml:"password,omitempty" json:"password"`
-	RawUseTLS                 bool   `xml:"tls,attr" json:"useTLS"`
-	APIKey                    string `xml:"apikey,omitempty" json:"apiKey"`
-	InsecureAdminAccess       bool   `xml:"insecureAdminAccess,omitempty" json:"insecureAdminAccess"`
-	Theme                     string `xml:"theme" json:"theme" default:"default"`
-	Debugging                 bool   `xml:"debugging,attr" json:"debugging"`
-	InsecureSkipHostCheck     bool   `xml:"insecureSkipHostcheck,omitempty" json:"insecureSkipHostcheck"`
-	InsecureAllowFrameLoading bool   `xml:"insecureAllowFrameLoading,omitempty" json:"insecureAllowFrameLoading"`
+	Enabled                   bool     `xml:"enabled,attr" json:"enabled" default:"true"`
+	RawAddress                []string `xml:"address" json:"address" default:"127.0.0.1:8384"`
```
AudriusButkevicius: Should be plural.

cmd/syncthing/gui.go
```diff
@@ -367,12 +375,20 @@ func (s *apiService) Serve() {
...
+		Handler: handler,
+		// ReadTimeout must be longer than SyncthingController $scope.refresh
+		// interval to avoid HTTP keepalive/GUI refresh race.
+		ReadTimeout: 15 * time.Second,
+	}
+	for _, listener := range listeners {
+		go func(listener net.Listener) {
+			serveError <- srv.Serve(listener)
+		}(listener)
+	}
```
AudriusButkevicius: This will leave zombie routines hanging on channel send for every other listener apart from the first one that fails.

cmd/syncthing/gui.go
```diff
@@ -334,29 +338,33 @@ func (s *apiService) Serve() {
...
+		for _, addr := range guiCfg.Addresses() {
+			if !addressIsLocalhost(addr) {
+				hasRemoteAddr = true
+				break
+			}
+		}
+		// Verify source host if all listening addresses are local.
+		if !hasRemoteAddr {
+			handler = localhostMiddleware(handler)
+		}
```
AudriusButkevicius: This should be a separate handler per listener, as now a single address taints all handlers.

Also, I am thinking that InsecureSkipHostCheck and UseTLS to be per address (which means we'd need Address to be a struct with additional parameters, not just a string)

cmd/syncthing/gui.go
```diff
@@ -233,13 +235,15 @@ func (s *apiService) Serve() {
 		}
 	}
 
-	if listener == nil {
+	if 0 == len(listeners) {
```
AudriusButkevicius: This is not C, no need for yoda notation.

cmd/syncthing/gui.go
```diff
@@ -189,16 +189,18 @@ func (s *apiService) getListener(guiCfg config.GUIConfiguration) (net.Listener,
...
-		TLSConfig: tlsCfg,
+	listeners := []net.Listener{}
+	for _, addr := range guiCfg.Addresses() {
+		rawListener, err := net.Listen("tcp", addr)
+		if err != nil {
+			return nil, err
+		}
+		listeners = append(listeners, &tlsutil.DowngradingListener{
+			Listener:  rawListener,
+			TLSConfig: tlsCfg,
```
AudriusButkevicius: Feature request to specify a separate TLS certificate per listener incoming in 3, 2, 1...

cmd/stcli/cmd_gui.go
```diff
@@ -48,7 +48,7 @@ func guiDump(c *cli.Context) {
 	writer := newTableWriter()
 	fmt.Fprintln(writer, "Enabled:\t", cfg.Enabled, "\t(enabled)")
 	fmt.Fprintln(writer, "Use HTTPS:\t", cfg.UseTLS(), "\t(tls)")
-	fmt.Fprintln(writer, "Listen Addresses:\t", cfg.Address(), "\t(address)")
+	fmt.Fprintln(writer, "Listen Addresses:\t", strings.Join(cfg.Addresses(), ","), "\t(address)")
```
AudriusButkevicius: Should probably call it "addresses" instead of "address" in all key checks.

cmd/syncthing/gui.go
```diff
@@ -398,15 +414,20 @@ func (s *apiService) String() string {
 }
 
 func (s *apiService) VerifyConfiguration(from, to config.Configuration) error {
-	_, err := net.ResolveTCPAddr("tcp", to.GUI.Address())
-	return err
+	for _, addr := range to.GUI.Addresses() {
+		_, err := net.ResolveTCPAddr("tcp", addr)
+		if err != nil {
+			return err
+		}
```
AudriusButkevicius: Can be written as:
```
if _, err := net.ResolveTCPAddr("tcp", addr); err != nil {
  return err
}
```

cmd/syncthing/gui.go
```diff
@@ -398,15 +414,20 @@ func (s *apiService) String() string {
...
 }
 
 func (s *apiService) CommitConfiguration(from, to config.Configuration) bool {
 	// No action required when this changes, so mask the fact that it changed at all.
 	from.GUI.Debugging = to.GUI.Debugging
 
-	if to.GUI == from.GUI {
+	if reflect.DeepEqual(to.GUI, from.GUI) {
 		return true
 	}
```
AudriusButkevicius: Why does this need a reflect?

lib/config/guiconfiguration.go
```diff
@@ -13,20 +13,20 @@ import (
...
+	User                      string   `xml:"user,omitempty" json:"user"`
+	Password                  string   `xml:"password,omitempty" json:"password"`
+	RawUseTLS                 bool     `xml:"tls,attr" json:"useTLS"`
+	APIKey                    string   `xml:"apikey,omitempty" json:"apiKey"`
+	InsecureAdminAccess       bool     `xml:"insecureAdminAccess,omitempty" json:"insecureAdminAccess"`
+	Theme                     string   `xml:"theme" json:"theme" default:"default"`
+	Debugging                 bool     `xml:"debugging,attr" json:"debugging"`
+	InsecureSkipHostCheck     bool     `xml:"insecureSkipHostcheck,omitempty" json:"insecureSkipHostcheck"`
+	InsecureAllowFrameLoading bool     `xml:"insecureAllowFrameLoading,omitempty" json:"insecureAllowFrameLoading"`
```
AudriusButkevicius: InsecureAllowFrameLoading, InsecureSkipHostCheck, InsecureAdminAccess, RawUseTLS should probably be per-address.

lib/config/guiconfiguration.go
```diff
@@ -36,12 +36,12 @@ func (c GUIConfiguration) Address() string {
...
 			if err != nil {
-				return override
+				return []string{override}
 			}
-			return url.Host
+			return []string{url.Host}
 		}
 
-		return override
+		return []string{override}
```
AudriusButkevicius: Should probably add support for comma notation for multiple addresses.

lib/config/guiconfiguration.go
```diff
@@ -54,10 +54,10 @@ func (c GUIConfiguration) UseTLS() bool {
 	return c.RawUseTLS
 }
 
-func (c GUIConfiguration) URL() string {
+func (c GUIConfiguration) UrlFromAddress(address string) string {
```
AudriusButkevicius: URLFromAddress

lib/config/guiconfiguration.go
```diff
@@ -79,6 +79,14 @@ func (c GUIConfiguration) URL() string {
 	return u.String()
 }
 
+func (c GUIConfiguration) URLs() []string {
+	Urls := []string{}
```
AudriusButkevicius: urls

cmd/syncthing/gui.go
```diff
@@ -334,29 +338,33 @@ func (s *apiService) Serve() {
...
-	l.Infoln("GUI and API listening on", listener.Addr())
-	l.Infoln("Access the GUI via the following URL:", guiCfg.URL())
+	for _, listener := range listeners {
+		l.Infoln("GUI and API listening on", listener.Addr())
+		l.Infoln("Access the GUI via the following URL:", guiCfg.UrlFromAddress(listener.Addr().String()))
+	}
 	if s.started != nil {
 		// only set when run by the tests
-		s.started <- listener.Addr().String()
+		s.started <- listeners[0].Addr().String()
```
AudriusButkevicius: Indexing into a potentially empty array feels a bit scary.

cmd/syncthing/gui.go
lstordeur: Because the GUI configuration now has a slice field which is a reference type and does not support native structural equality.

lib/config/guiconfiguration.go
lstordeur: Does it mean that we need to promote the `RawAddresses` field as a structure ?

lib/config/guiconfiguration.go
lstordeur: I am not sure to see your point at the level of this function. Where would you precisely introduce support for comma notation?

lib/config/guiconfiguration.go
lstordeur: Fine

lib/config/guiconfiguration.go
lstordeur: Fine

lib/config/guiconfiguration.go
lstordeur: Fine

cmd/syncthing/gui.go
AudriusButkevicius: It's sad but true :(

cmd/syncthing/gui.go
lstordeur: I see your point. Do you expect this feature as part of this pull request?

cmd/syncthing/gui.go
lstordeur: Fine

lib/config/guiconfiguration.go
AudriusButkevicius: STGUIADDRESS env var override should probably support comma notation, as in `STGUIADDRESS=http://0.0.0.0:80,https://0.0.0.0:443`

cmd/syncthing/gui.go
lstordeur: Maybe I am wrong but the listeners are obtained from the `GUIConfiguration` structure which specifies a default listener value (to `127.0.0.1:8384`).

cmd/syncthing/gui.go
lstordeur: Fine

lib/config/guiconfiguration.go
AudriusButkevicius: That, plus a migration, converting the old string to a new struct.
So `Deprecated_RawAddress`, `Deprecated_UseTLS`, etc etc, gets converted to:
```
GUIAddress{ 
  Addresss string
  UseTLS bool
  ...
}
```

I guess asking you to do that is unfair, as you are only interested in getting support for 2nd'ary addresses, but this is the way I'd imagine it works under ideal circumstances, so I am just throwing it out there.

cmd/syncthing/gui.go
AudriusButkevicius: I guess, but I'd write a test to verify this, as in ensure that loading a config with no address section we still end up with atleast a single address entry.

This will furthermore need UI improvements, as in the UI currently listen address is a single field.

cmd/stcli/cmd_gui.go
lstordeur: Which key checks are you talking about?

cmd/syncthing/gui.go
AudriusButkevicius: Not necessarily, just throwing it out there in case you are willing to add that.

cmd/stcli/cmd_gui.go
AudriusButkevicius: The switch statement, `case "address"` should become `case "addresses"` (as well as address -> addresses in all other places)

lib/config/guiconfiguration.go
lstordeur: Fine.

cmd/syncthing/gui.go
lstordeur: I will investigate that point.

lib/config/guiconfiguration.go
lstordeur: I will investigate that point.

lstordeur: Here is a proposal for a new GUIConfiguration structuration. What's your opinion?

```
type GUIListener struct {
	RawAddress                string `xml:"address" json:"address" default:"127.0.0.1:8384"`
	RawUseTLS                 bool   `xml:"tls,attr" json:"useTLS"`
	InsecureAdminAccess       bool   `xml:"insecureAdminAccess,omitempty" json:"insecureAdminAccess"`
	InsecureSkipHostCheck     bool   `xml:"insecureSkipHostcheck,omitempty" json:"insecureSkipHostcheck"`
	InsecureAllowFrameLoading bool   `xml:"insecureAllowFrameLoading,omitempty" json:"insecureAllowFrameLoading"`
}

type GUIConfiguration struct {
	Listener  GUIListener `xml:"listener" json:"listener"`
	Enabled   bool        `xml:"enabled,attr" json:"enabled" default:"true"`
	User      string      `xml:"user,omitempty" json:"user"`
	Password  string      `xml:"password,omitempty" json:"password"`
	APIKey    string      `xml:"apikey,omitempty" json:"apiKey"`
	Theme     string      `xml:"theme" json:"theme" default:"default"`
	Debugging bool        `xml:"debugging,attr" json:"debugging"`

	// Deprecated. Old listener configuration style.
	RawAddresses              string `xml:"address" json:"address" default:"127.0.0.1:8384"`
	RawUseTLS                 bool   `xml:"tls,attr" json:"useTLS"`
	InsecureAdminAccess       bool   `xml:"insecureAdminAccess,omitempty" json:"insecureAdminAccess"`
	InsecureSkipHostCheck     bool   `xml:"insecureSkipHostcheck,omitempty" json:"insecureSkipHostcheck"`
	InsecureAllowFrameLoading bool   `xml:"insecureAllowFrameLoading,omitempty" json:"insecureAllowFrameLoading"`
}
```

cmd/stcli/cmd_gui.go
lstordeur: Fine

cmd/syncthing/gui.go
lstordeur: Go concurrency paradigm is still an area where I do not yet feel comfortable. ;-)
How would you better handle all these goroutine lifetimes?

lib/config/guiconfiguration.go
AudriusButkevicius: I'd also rename the old deprecated fields as Deprecated_, and add a config migration which converts old values to a single item to Listeners, and nulls them out. Also, you'd want to add omitEmpty to them, to remove them from showing up.

cmd/syncthing/gui.go
AudriusButkevicius: So there are a few approaches, depending on your appetite to learn.

Easiest is to do:
```
err := srv.Serve(listener)
select {
  serverError <- err:
  default:
}
```

So that we discard errors once the channel is full.
As the loop exits, the channel would get GCed.

More idiomatic way would be to have a wait group, add 1 for every listener, create error channel with a buffer of `len(listeners)`. Upon receiving the first error (or the stop), close all listeners, wait for the waitgroup to resolve (all listeners to exit), drain the error channel (discarding errors), close the channel, and return.

lib/config/guiconfiguration.go
lstordeur: Updated according to your observations:

```
type GUIListener struct {
	Address                   string `xml:"address" json:"address" default:"127.0.0.1:8384"`
	UseTLS                    bool   `xml:"tls,attr" json:"useTLS"`
	InsecureAdminAccess       bool   `xml:"insecureAdminAccess,omitempty" json:"insecureAdminAccess"`
	InsecureSkipHostCheck     bool   `xml:"insecureSkipHostcheck,omitempty" json:"insecureSkipHostcheck"`
	InsecureAllowFrameLoading bool   `xml:"insecureAllowFrameLoading,omitempty" json:"insecureAllowFrameLoading"`
}

type GUIConfiguration struct {
	Listeners []GUIListener `xml:"listener" json:"listeners"`
	Enabled   bool          `xml:"enabled,attr" json:"enabled" default:"true"`
	User      string        `xml:"user,omitempty" json:"user"`
	Password  string        `xml:"password,omitempty" json:"password"`
	APIKey    string        `xml:"apikey,omitempty" json:"apiKey"`
	Theme     string        `xml:"theme" json:"theme" default:"default"`
	Debugging bool          `xml:"debugging,attr" json:"debugging"`

	// Deprecated. Old listener configuration style.
	Deprecated_RawAddress                string `xml:"address,omitempty" json:"address"`
	Deprecated_RawUseTLS                 bool   `xml:"tls,attr,omitempty" json:"useTLS"`
	Deprecated_InsecureAdminAccess       bool   `xml:"insecureAdminAccess,omitempty" json:"insecureAdminAccess"`
	Deprecated_InsecureSkipHostCheck     bool   `xml:"insecureSkipHostcheck,omitempty" json:"insecureSkipHostcheck"`
	Deprecated_InsecureAllowFrameLoading bool   `xml:"insecureAllowFrameLoading,omitempty" json:"insecureAllowFrameLoading"`
}
```

lstordeur: Maybe we should rename the `Addresses()` method to `Listeners()` for coherency.

lstordeur: Also, do we consider to rename the environment variable `STGUIADDRESS` to `STGUIADDRESSES`? This will of course introduce a backward incompatibility. Or maybe we could support both forms.

AudriusButkevicius: Sure. Yet you'll have to change the tags on the deprecated fields to prevent json serialization.

AudriusButkevicius: And as I said, `default` value has no meaning in your case.

lstordeur: Ok. I propose to handle both deprecated config migration and default listener injection in the `loadConfig()` function. Fine for you or do you see a better place?

lib/config/guiconfiguration.go
lstordeur: Here is my proposal to handle environment overriding with multiple listeners:
1. First we check for a new environment variable `STGUIADDRESSES` which supports a comma separated list of individual urls;
2. If the first one is not defined, we fallback to the legacy `STGUIADDRESS` environment variable which only support one url;
3. If the second one is not defined, we use the listeners as defined in the configuration.

lib/config/guiconfiguration.go
AudriusButkevicius: Migrations are handled in /lib/config/config.go or wrapper.go, you should just add a migration there.
Injecting a default listener should probably happen somewhere in prepare(), given Enabled is true.

cmd/syncthing/main.go
```diff
@@ -451,7 +451,7 @@ func main() {
 func openGUI() {
 	cfg, _ := loadConfig()
 	if cfg.GUI().Enabled {
-		openURL(cfg.GUI().URL())
+		openURL(cfg.GUI().URLs()[0])
```
AudriusButkevicius: This won't work if the first address will be UNIX socket for example. Not saying you have to fix this now, but keep this in mind.

lib/config/guiconfiguration.go
lstordeur: Fine.
My last blocking point concerns the JSON serialization part related to the deprecated fields. Where concretely should we prevent them to be serialized?

lib/config/guiconfiguration.go
AudriusButkevicius: Sounds good to me. @calmh?

lib/config/guiconfiguration.go
lstordeur: Oh. The new configuration restructuration implies to rework the dump/get/set operations at the `stcli' command level. How should we cope with the new structuration at this level?

lib/config/guiconfiguration.go
calmh: sgtm

cmd/syncthing/gui.go
calmh: One of the things I thought while glancing at the config diff was that the addition of `RawAddresses []string` might instead have been a change to have a slice of GUI configurations instead of just a single one. I'm not sure if this makes it more or less difficult.

lib/config/guiconfiguration.go
AudriusButkevicius: The tool is half broken anyway, so we can many any needed breaking changes there.

lstordeur: Work on the go code looks well. Now I definitely need some pointers for the Javascript part.
From my current comprehension, I have identified the following tasks:
- update the HTML layout for the Settings modal view. Maybe we could dynamically handle one   `<div>` per listener and gather all the listener properties (address, UseTLS, ...) inside each division. But I have no idea on how to handle the dynamic side;
- I suspect that the test related to `$scope.openNoAuth` should be performed on the right listener but I have no idea on how to identify the right one.
- and definitely some other impacts I have not yet identified ;-)

cmd/syncthing/gui.go
lstordeur: Fine

lib/config/guiconfiguration.go
AudriusButkevicius: Yeah, this will be an interesting problem as a lot of these settings will become conditional.
Best idea I have is to have a $scope.listeners [{useTls: true, addess:bar}, ..] then render a small template per listener, and have - button next to each entry to remove and a plus at the bottom to add.

lstordeur:
I have updated both:
- pull request description;
- my development branch

according to our discussion and the resulting proposals.

Note that my development branch only contains work on the Go code part, to make move forward the review process on this side. I have not yet started to work on the Javascript part but will start to investigate very soon.

I have tried to address nearly all of your comments in the new commits.

cmd/syncthing/gui.go
```diff
@@ -314,49 +351,22 @@ func (s *apiService) Serve() {
...
 	s.fss = newFolderSummaryService(s.cfg, s.model)
 	defer s.fss.Stop()
 	s.fss.ServeBackground()
 
-	l.Infoln("GUI and API listening on", listener.Addr())
-	l.Infoln("Access the GUI via the following URL:", guiCfg.URL())
+	for _, guiListener := range guiCfg.GUIListeners() {
+		l.Infoln("GUI and API listening on", guiListener.Address)
+		l.Infoln("Access the GUI via the following URL:", guiCfg.URLFromGUIListener(guiListener))
+	}
```
AudriusButkevicius: So this is ok for this PR I guess, but in general, I'd expect the Access the GUI to be printed only once for something that is not a unix socket. I know non-unix socket is not possible now, but just thinking ahead.

cmd/syncthing/gui.go
AudriusButkevicius: Same comment as above, opening unix://foo/bar is not very useful, so you'd want to make sure you pick something that is http.

lib/config/config.go
```diff
@@ -375,6 +384,41 @@ func (cfg *Configuration) clean() error {
...
+		cfg.GUI.Deprecated_InsecureAllowFrameLoading = false
+	}
+
+	// Migrate if at least one deprecated field had been filled.  If both
+	// deprecated and new configuration style are used, merged them
+	// together.
+	if guiListener != emptyListener {
+		cfg.GUI.Listeners = append(cfg.GUI.Listeners, guiListener)
+	}
+	cfg.Version = 24
```
AudriusButkevicius: So this is too complicated.
I'd just do
```
cfg.GUI.Listeners = append(cfg.GUI.Listeners, GUIListener{
    UseTLS: cfg.GUI.Deprecated_RawUseTLS
    ...
})
cfg.GUI.Deprecated_RawUseTLS = False
...
```

lib/config/guiconfiguration.go
```diff
@@ -12,56 +12,74 @@ import (
...
-	return c.RawAddress
+func GUIListenerFromEnv(envAddr string) (l GUIListener) {
+	if strings.Contains(envAddr, "/") {
+		url, err := url.Parse(envAddr)
+		if err != nil {
+			l.Address = envAddr
+		} else {
+			l.Address = url.Host
+		}
+	}
```
AudriusButkevicius: This will end up with empty address in case of `STGUIADDRESSES=127.0.0.1:1234`

lib/config/guiconfiguration.go
```diff
@@ -12,56 +12,74 @@ import (
...
+	// Deprecated. Old listener configuration style.
+	Deprecated_RawAddress                string `xml:"address,omitempty" json:"-"`
+	Deprecated_RawUseTLS                 bool   `xml:"tls,attr,omitempty" json:"-"`
+	Deprecated_InsecureAdminAccess       bool   `xml:"insecureAdminAccess,omitempty" json:"-"`
+	Deprecated_InsecureSkipHostCheck     bool   `xml:"insecureSkipHostcheck,omitempty" json:"-"`
+	Deprecated_InsecureAllowFrameLoading bool   `xml:"insecureAllowFrameLoading,omitempty" json:"-"`
+}
 
-	return c.RawAddress
+func GUIListenerFromEnv(envAddr string) (l GUIListener) {
```
AudriusButkevicius: I'd call this GUIListenerFromString or something like that.

lib/config/guiconfiguration.go
```diff
@@ -12,56 +12,74 @@ import (
...
 		Scheme: "http",
-		Host:   c.Address(),
+		Host:   guiListener.Address,
 		Path:   "/",
 	}
 
-	if c.UseTLS() {
+	if guiListener.UseTLS {
 		u.Scheme = "https"
 	}
```
AudriusButkevicius: You should create a:

```
func (l GUIListener) URL() string {
```
As now you call into GUIConfiguration that is not used in this function at all.

AudriusButkevicius:
So generally this looks ok, but conceptually, I suspect we'll have requests for having separate endpoints that are authenticated versus non-authenticated, using different certificates, etc, so perhaps @calmh's suggestion for a slice []GUIConfiguration was probably a better idea, yet I suspect we can address that in the future (with yet another migration potentially).

cmd/syncthing/gui.go
lstordeur: So, for this PR, I propose to only add a couple of comments to keep in mind that further work will need to be done here when we add UNIX socket support.

Also, to be sure : if one user configures several TCP listeners, do you definitely expect to print the Access URL only once for the first TCP listener ?

lib/config/config.go
lstordeur: You're so right ! ^^

lib/config/guiconfiguration.go
lstordeur: Right!

cmd/syncthing/gui.go
AudriusButkevicius: Perhaps you are right, it's most likely fine to print them all. We already print a ton of stuff, so this won't make it worse.

lstordeur:
Concerning the idea for a slice of GUIConfiguration:

In the context of this PR, I guess I can easily extend the slice flavor from `GUIListener' to `GUIConfiguration', leaving out of this work any immediate feature that could be added (like multiple certificate support).

One could see that as a preliminary work to the possible further feature requests you mentioned above. I think it makes sense if we can avoid one additional configuration migration (although the configuration migration support is very smooth and painless).

AudriusButkevicius:
Sounds good, I think that makes it more future proof.

lstordeur:
Of course, this is finally not so easy as it means at first sight. Especially, I do not know how to handle one point:

Okay, so I know handle a slice of `GUIConfiguration`. Obviously, I had to update several points where one used to expect only one GUI instance. However, the `Theme` field is a bit harder to update. Indeed, this field is directly used to instantiate the API Service:

```
func newAPIService(id protocol.DeviceID, cfg configIntf, httpsCertFile, httpsKeyFile, assetDir string, m modelIntf, defaultSub, diskSub events.BufferedSubscription, discoverer discover.CachingMux, connectionsService connectionsIntf, errors, systemLog logger.Recorder, cpu rater) *apiService {
	service := &apiService{
		id:            id,
		cfg:           cfg,
		httpsCertFile: httpsCertFile,
		httpsKeyFile:  httpsKeyFile,
		statics:       newStaticsServer(cfg.GUIs().Theme, assetDir),
		model:         m,
		eventSubs: map[events.EventType]events.BufferedSubscription{
			defaultEventMask: defaultSub,
			diskEventMask:    diskSub,
		},
		eventSubsMut:       sync.NewMutex(),
		discoverer:         discoverer,
		connectionsService: connectionsService,
		systemConfigMut:    sync.NewMutex(),
		stop:               make(chan struct{}),
		configChanged:      make(chan struct{}),
		startedOnce:        make(chan struct{}),
		guiErrors:          errors,
		systemLog:          systemLog,
		cpu:                cpu,
	}

	return service
}
```

Does it mean that I now have to instantiate several API Service endpoints, once per `GUIConfiguration`?

AudriusButkevicius:
The static service interface should be changed to accept the theme as an argument to calls rather than the constructor. Alternatively, we might go for static service per listener for example.

lstordeur:
I have takem the path to static service per listener.

However, the same kind of problem raises a bit further with debugging handlers. I do not see how to address this one:

```
func (s *apiService) whenDebugging(h http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if s.cfg.GUI().Debugging {
			h.ServeHTTP(w, r)
			return
		}

		http.Error(w, "Debugging disabled", http.StatusBadRequest)
	})
}
```
The test on the `Debugging` field must be achieved in a dynamic way (so definitely in the closure). Meaning, I do not see how to handle that without instantiating one API Service per GUI instance.

AudriusButkevicius:
We generate separate handlers per gui anyway, so this has to be modified to the gui config as a parameter and check debugging on that.

AudriusButkevicius:
I'd have to check at home (not on my phone), perhaps api service per listener makes sense.

AudriusButkevicius:
So API service per listener is a big rough, as then it would mean even subscriptions per listener, etc, which would be a bit of an overkill. 

We will generate a separate set of handlers per Listener, so the debug stuff can just be decided there, 
Again, whenDebugging only refers to `s` (the service) to access the Listener config, which implies that in multiple listener case, it shouldn't even live on the `s`.

lstordeur:
I have created a new PR #4433 to track developments related to multiple gui configuration support.

lkwg82:
As there is a superseeding PR #4433, I close this.

PR #4417, opened 2017-10-13
Add -pkgDir parameter for build script

Nutomic:
### Purpose

Allow passing the -pkgDir parameter to `go build` via build.go. Required for https://github.com/syncthing/syncthing-android/issues/873#issuecomment-336356056

### Testing

Ran build.go with and without -pkgDir and verified expected behaviour.

### DIFF
--- a/build.go
+++ b/build.go
@@ -45,6 +45,7 @@ var (
 	noBuildGopath bool
 	extraTags     string
 	installSuffix string
+	pkgdir	      string
 )
 
 type target struct {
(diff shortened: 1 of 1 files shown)


build.go
```diff
@@ -427,6 +429,9 @@ func build(target target, tags []string) {
 
 	rmr(target.BinaryName())
 	args := []string{"build", "-i", "-v", "-ldflags", ldflags()}
+	if (pkgDir != "") {
```
AudriusButkevicius: No need for braces around the if, it's not java.

build.go
```diff
@@ -342,6 +343,7 @@ func parseFlags() {
 	flag.BoolVar(&noBuildGopath, "no-build-gopath", noBuildGopath, "Don't build GOPATH, assume it's OK")
 	flag.StringVar(&extraTags, "tags", extraTags, "Extra tags, space separated")
 	flag.StringVar(&installSuffix, "installsuffix", installSuffix, "Install suffix, optional")
+	flag.StringVar(&pkgDir, "pkgDir", "", "Set -pkgdir parameter for `go build`")
```
AudriusButkevicius: pkgdir flag pls, like passed to the compiler. Not fussed on the variable name

Nutomic:
Fixed. Also added the pkgdir parameter to install() for consistency. I don't think there are any other commands where it is needed?

AudriusButkevicius:
@st-review merge

build: Support passing -pkgdir

st-review:
:ok_hand: Merged as cdbb32d0f043c56c6787df3f036feabc8891fd3c. Thanks, @Nutomic!

PR #4422, opened 2017-10-14
cmd/syncthing: Expand usage stats even more (ref #3628)

AudriusButkevicius:
Also add diffing functionality

### DIFF
--- a/cmd/syncthing/usage_report.go
+++ b/cmd/syncthing/usage_report.go
@@ -12,6 +12,7 @@ import (
 	"crypto/rand"
 	"crypto/tls"
 	"encoding/json"
+	"net"
 	"net/http"
 	"runtime"
 	"sort"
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
Don't merge yet

AudriusButkevicius:
Now it's good to merge.

@imsodin wants to add usage stats for fs watcher, so I guess we should get his stuff merged before making an rc or a release.

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 059185b32569aaec3f95d4fe5bab829f8df1104e. Thanks, @AudriusButkevicius!

calmh:
If the fs notifier thing is close to merging I'm willing to break the rc process a little bit - we can release several RCs while iterating. Tweaking the usage reporting between two RCs is fine as it's always on an never asked about (don't remember if we verified that's still the case with the usage reporting upgrades, I remember mentioning it).

AudriusButkevicius:
The only problem with reiterating is the migration on ur reporting server can only be done once.

PR #4425, opened 2017-10-16
build: Warn when being executed outside of gopath

imsodin:
Currently if you have Syncthing checked out in two directories, once at the proper place in the gopath, once somewhere outside. If you then run e.g. `go run build.go build` in the source dir outside of gopath, assets will be generated in the local working dir (i.e. outside of gopath) while building happens within gopath. This can then lead to the build failing, due to missing assets. This is due to commands from the go toolchain operating within gopath and generating assets within the current working directory.

This patch removes this inconsistency by moving the working directory into the gopath. I am not entirely convinced, this is the cleanest way to go, but I couldn't think of anything else that is really consistent. It also means that the `install` operations cannot install to `./bin` anymore, but always to `gopath/bin`.

### DIFF
--- a/build.go
+++ b/build.go
@@ -204,30 +204,31 @@ func main() {
 		}()
 	}
 
-	gopath := gopath()
-	if gopath == "" {
-		var err error
-		gopath, err = temporaryBuildDir()
+	if gopath := gopath(); gopath == "" {
+		gopath, err := temporaryBuildDir()
 		if err != nil {
 			log.Fatal(err)
 		}
 		if !noBuildGopath {
 			lazyRebuildAssets()
-			if err = buildGOPATH(gopath); err != nil {
+			if err := buildGOPATH(gopath); err != nil {
 				log.Fatal(err)
 			}
 		}
 		os.Setenv("GOPATH", gopath)
 		log.Println("GOPATH is", gopath)
+	} else {
+		gopath, _ = filepath.EvalSymlinks(gopath)
+		wd, _ := os.Getwd()
+		wd, _ = filepath.EvalSymlinks(wd)
+		if filepath.Join(gopath, "src/github.com/syncthing/syncthing") != wd {
+			fmt.Println("You are running this outside of GOPATH/src/github.com/syncthing/syncthing, this might cause failure!")
+		}
 	}
 
 	// Set path to $GOPATH/bin:$PATH so that we can for sure find tools we
 	// might have installed during "build.go setup".
-	os.Setenv("PATH", fmt.Sprintf("%s%cbin%c%s", gopath, os.PathSeparator, os.PathListSeparator, os.Getenv("PATH")))
-
-	if wd, _ := os.Getwd(); !strings.HasPrefix(wd, gopath) {
-		os.Chdir(filepath.Join(gopath, "src/github.com/syncthing/syncthing"))
-	}
+	os.Setenv("PATH", fmt.Sprintf("%s%cbin%c%s", os.Getenv("GOPATH"), os.PathSeparator, os.PathListSeparator, os.Getenv("PATH")))
 
 	checkArchitecture()
 
--- a/build.go
+++ b/build.go
@@ -45,7 +45,7 @@ var (
 	noBuildGopath bool
 	extraTags     string
 	installSuffix string
-	pkgdir	      string
+	pkgdir        string
 )
 
 type target struct {
@@ -204,14 +204,16 @@ func main() {
 		}()
 	}
 
-	if gopath() == "" {
-		gopath, err := temporaryBuildDir()
+	gopath := gopath()
+	if gopath == "" {
+		var err error
+		gopath, err = temporaryBuildDir()
 		if err != nil {
 			log.Fatal(err)
 		}
 		if !noBuildGopath {
 			lazyRebuildAssets()
-			if err := buildGOPATH(gopath); err != nil {
+			if err = buildGOPATH(gopath); err != nil {
 				log.Fatal(err)
 			}
 		}
@@ -221,7 +223,11 @@ func main() {
 
 	// Set path to $GOPATH/bin:$PATH so that we can for sure find tools we
 	// might have installed during "build.go setup".
-	os.Setenv("PATH", fmt.Sprintf("%s%cbin%c%s", os.Getenv("GOPATH"), os.PathSeparator, os.PathListSeparator, os.Getenv("PATH")))
+	os.Setenv("PATH", fmt.Sprintf("%s%cbin%c%s", gopath, os.PathSeparator, os.PathListSeparator, os.Getenv("PATH")))
+
+	if wd, _ := os.Getwd(); !strings.HasPrefix(wd, gopath) {
+		os.Chdir(filepath.Join(gopath, "src/github.com/syncthing/syncthing"))
+	}
 
 	checkArchitecture()
 


calmh:
If the issue is just the assets thing, I think we should just make sure to generate assets before jumping into our "shadow-GOPATH"

imsodin:
I am not entirely sure, I didn't check thoroughly, but as far as I can see it is the only thing.

If it really is, the following should solve the problem as well:

```diff
--- a/build.go
+++ b/build.go
@@ -642,8 +637,9 @@ func listFiles(dir string) []string {
 }
 
 func rebuildAssets() {
-	runPipe("lib/auto/gui.files.go", "go", "run", "script/genassets.go", "gui")
-	runPipe("cmd/strelaypoolsrv/auto/gui.go", "go", "run", "script/genassets.go", "cmd/strelaypoolsrv/gui")
+	gopath := os.Getenv("GOPATH")
+	runPipe(filepath.Join(gopath, "lib/auto/gui.files.go"), "go", "run", "script/genassets.go", "gui")
+	runPipe(filepath.Join(gopath, "cmd/strelaypoolsrv/auto/gui.go"), "go", "run", "script/genassets.go", "cmd/strelaypoolsrv/gui")
 }
 
 func lazyRebuildAssets() {
@@ -653,6 +649,8 @@ func lazyRebuildAssets() {
 }
 
 func shouldRebuildAssets(target, srcdir string) bool {
+	target = filepath.Join(os.Getenv("GOPATH"), target)
+	srcdir = filepath.Join(os.Getenv("GOPATH"), srcdir)
 	info, err := os.Stat(target)
 	if err != nil {
 		// If the file doesn't exist, we must rebuild it
```

calmh:
That said, for me this already seems to work out of the box:

```
jb@unu:/t/syncthing $ pwd
/tmp/syncthing
jb@unu:/t/syncthing $ git clean -fxd
Removing bin/
Removing cmd/strelaypoolsrv/auto/gui.go
Removing lib/auto/gui.files.go
jb@unu:/t/syncthing $ ls -l lib/auto/
total 16
-rw-r--r--  1 jb  wheel  719 Oct 16 15:33 auto_test.go
-rw-r--r--  1 jb  wheel  327 Oct 16 15:33 doc.go
jb@unu:/t/syncthing $ export | grep GOPATH
jb@unu:/t/syncthing $ ./build.sh 
GOPATH is /var/folders/9l/prphpj9n431dsb2myvdjpj4w0000gp/T/syncthing-578c0728
github.com/syncthing/syncthing/cmd/stcompdirs
github.com/syncthing/syncthing/cmd/stgenfiles
github.com/syncthing/syncthing/cmd/stevents
github.com/syncthing/syncthing/cmd/stcli
github.com/syncthing/syncthing/cmd/stbench
github.com/syncthing/syncthing/cmd/stdisco
github.com/syncthing/syncthing/cmd/stdiscosrv
github.com/syncthing/syncthing/cmd/stfileinfo
github.com/syncthing/syncthing/cmd/stfinddevice
github.com/syncthing/syncthing/cmd/stindex
github.com/syncthing/syncthing/cmd/strelaypoolsrv
github.com/syncthing/syncthing/cmd/strelaysrv
github.com/syncthing/syncthing/cmd/strelaysrv/testutil
github.com/syncthing/syncthing/cmd/stsigtool
github.com/syncthing/syncthing/cmd/stvanity
github.com/syncthing/syncthing/cmd/stwatchfile
github.com/syncthing/syncthing/cmd/syncthing
ok  	_/tmp/syncthing/meta	3.693s
jb@unu:/t/syncthing $ ls -l bin/
total 196016
-rwxr-xr-x  1 jb  staff   6187932 Oct 16 15:34 stbench*
-rwxr-xr-x  1 jb  staff   7692876 Oct 16 15:34 stcli*
-rwxr-xr-x  1 jb  staff   1594192 Oct 16 15:34 stcompdirs*
-rwxr-xr-x  1 jb  staff   5555196 Oct 16 15:34 stdisco*
-rwxr-xr-x  1 jb  staff   9306684 Oct 16 15:34 stdiscosrv*
-rwxr-xr-x  1 jb  staff   4470692 Oct 16 15:34 stevents*
-rwxr-xr-x  1 jb  staff   6620716 Oct 16 15:34 stfileinfo*
-rwxr-xr-x  1 jb  staff   5751836 Oct 16 15:34 stfinddevice*
-rwxr-xr-x  1 jb  staff   1577824 Oct 16 15:34 stgenfiles*
-rwxr-xr-x  1 jb  staff   6217564 Oct 16 15:34 stindex*
-rwxr-xr-x  1 jb  staff   6442252 Oct 16 15:34 strelaypoolsrv*
-rwxr-xr-x  1 jb  staff   8732396 Oct 16 15:34 strelaysrv*
-rwxr-xr-x  1 jb  staff   4553492 Oct 16 15:34 stsigtool*
-rwxr-xr-x  1 jb  staff   3295936 Oct 16 15:34 stvanity*
-rwxr-xr-x  1 jb  staff   1539024 Oct 16 15:34 stwatchfile*
-rwxr-xr-x  1 jb  staff  15006188 Oct 16 15:34 syncthing*
-rwxr-xr-x  1 jb  staff   5780908 Oct 16 15:34 testutil*
jb@unu:/t/syncthing $ 
```

calmh:
Oh, wait, is this about *having* a GOPATH set, and *having* a Syncthing in the right place, but building *another* Syncthing somewhere else? Yeah, no, that's not supported. We make some attempts to build "anywhere" on a system that does not have a GOPATH, but if there is a GOPATH set it needs to be correct and Syncthing needs to be in the right place within it. This is Go, after all.

imsodin:
> Oh, wait, is this about having a GOPATH set, and having a Syncthing in the right place, but building another Syncthing somewhere else? Yeah, no, that's not supported. We make some attempts to build "anywhere" on a system that does not have a GOPATH, but if there is a GOPATH set it needs to be correct and Syncthing needs to be in the right place within it. This is Go, after all.

Exactly that! It may not be supported, but it appears to work except for throwing unclear errors because the generated assets are missing. If running any command of build.go from outside of `GOPATH/src/github.com/synthing/syncthing` isn't supported, we could check that and abort early, with a helpful error message.

calmh:
It's super unclear in general as you can never be sure what source you're building. Basically you'll be building the source in GOPATH, except for tests which do `./...` and the like. Possibly we could try to catch it and prevent it, but I can't think of a single reasonable situation in which this would actually happen and I definitely don't think we should attempt to "handle" it somehow...

imsodin:
Ok, I get the we shouldn't handle it now, I only tried that in the first place as we already have a little bit of GOPATH mangling. The new commit just prints a warning when the current working directory is not as expected, such that potentially resulting cryptic errors get some context.

The reason I brought this up is, that in syncthing/syncthing-android#968 some time was spent due to this error:
https://travis-ci.org/syncthing/syncthing-android/builds/288434755#L2387
If you don't have Syncthing asset generation in mind and know that it happens locally, this does not really point to a problem with GOPATH. There are other pointers, but usually with a wrong GOPATH building fails immediately. This is definitely a very specific case, but GOPATH in general is such a error-prone system that I think everything that can provide some meaningful feedback is helpful.

Nutomic:
Yep, I ran into this problem and got very confused, as there was no hint at all what was causing the error. Thanks @imsodin for figuring this all out!

AudriusButkevicius:
@st-review merge

build: Warn when being executed outside of gopath

st-review:
:ok_hand: Merged as d84b6bb8226540f53e74b9da2e24d6cddbe5acda. Thanks, @imsodin!

calmh:
This doesn't work when using a split GOPATH, like I do:

```
jb@unu:~/s/g/s/syncthing $ echo $GOPATH
/Users/jb/ext:/Users/jb
jb@unu:~/s/g/s/syncthing $ pwd
/Users/jb/src/github.com/syncthing/syncthing
jb@unu:~/s/g/s/syncthing $ ./build.sh 
You are running this outside of GOPATH/src/github.com/syncthing/syncthing, this might cause failure!
ok  	github.com/syncthing/syncthing/meta	4.167s
```

imsodin:
I should have known better than to pick a fight with GOPATH, nevertheless: #4447

calmh:
To the tune of The Clash's *I fought the law*... 🎶 I fought the $GOPATH, but the $GOPATH won 🎵

PR #4426, opened 2017-10-16
lib/db: filter unchanged files when updating and polish

imsodin:
The behaviour of `db.updateFiles` and `db.WithNeed` is completely unchanged, I just made the code (subjectively) more readable and succinct.

The higher level `FileSet.Update` did not previously check whether the invalid flag has changed to update the blockmap (i.e. did not add/remove blocks when they are un-/ignored). In addition the file is looked up both here and then later in `db.updateFiles`. To prevent the second lookup just to be dropped in case the file hasn't changed, it is filtered out already here.

Testing:  
Extended `TestUpdateToInvalid` to also check the state of the blockmap -> fails without PR, passes now. Existing tests for updating files still work.

### DIFF
--- a/lib/db/leveldb_dbinstance.go
+++ b/lib/db/leveldb_dbinstance.go
@@ -206,33 +206,23 @@ func (db *Instance) updateFiles(folder, device []byte, fs []protocol.FileInfo, l
 			err = ef.Unmarshal(bs)
 		}
 
-		if err != nil {
-			if isLocalDevice {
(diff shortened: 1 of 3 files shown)


calmh:
@st-review merge

lib/db: Filter unchanged files when updating and polish

st-review:
:ok_hand: Merged as 7ba9e7c322cc18cfa82f8236cf770530c4f19e35. Thanks, @imsodin!

PR #4430, opened 2017-10-17
Configurable staggered versioning

trmfreitas:
### Purpose

This change allows configuration of staggered versioning mode. Specifically the intervals.

### Testing

Testing was done to check if input boxes on gui was showing correct information and that interval configuration actually changed something.

### Screenshots

image

### Documentation

I am not sure if the GUI should be changed. The change I did is not very user friendly. For me, and perhaps most advanced uses of versioning feature, a configurable section in the XML is good enough.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,11 +1,17 @@
-Do not report security issues in this bug tracker. Instead, contact
-security@syncthing.net directly - see https://syncthing.net/security.html
-for more information.
+### DO NOT REPORT SECURITY ISSUES IN THIS ISSUE TRACKER
 
-If your issue is a support request ("How do I get my devices to connect?"
-or similar), please use the support forum at https://forum.syncthing.net/
-where a large number of helpful people hang out. This issue tracker is for
-reporting bugs or feature requests directly to the developers.
+Instead, contact security@syncthing.net directly - see
+https://syncthing.net/security.html for more information.
+
+### DO NOT POST SUPPORT REQUESTS OR GENERAL QUESTIONS IN THIS ISSUE TRACKER
+
+Please use the forum at https://forum.syncthing.net/ where a large number of
+helpful people hang out. This issue tracker is for reporting bugs or feature
+requests directly to the developers. Worst case you might get a short
+"that's a bug, please report it on GitHub" response on the forum, in which
+case we thank you for your patience and following our advice. :)
+
+### Please do post actual bug reports and feature requests.
 
 If your issue is a bug report, replace this boilerplate with a description
 of the problem, being sure to include at least:
--- a/build.go
+++ b/build.go
@@ -45,6 +45,7 @@ var (
 	noBuildGopath bool
 	extraTags     string
 	installSuffix string
+	pkgdir	      string
 )
 
 type target struct {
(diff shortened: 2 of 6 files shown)


AudriusButkevicius:
So I suggest you implement this as an external versioner, there is no reason for the average user to have access to this.

trmfreitas:
So, you don't think it is usable to have this config in the xml only?

AudriusButkevicius:
It could be there, but given it's the first time I've heard anyone interested in tweaking these, I'd just suggest to have an external versioner instead.

calmh:
I'm inclined to agree; it's difficult enough to understand already without adding this level of customization.

Ferroin:
I agree this should probably be config only, not in the UI, it's pretty cluttered, and I doubt many people will use it (on the grounds that the default intervals are pretty standard for things like this).  That said, If I had the ability to configure intervals like this, and ideally control the number of intervals and frequency of copies kept as well, I would probably be using the staggered versioning for most things.

trmfreitas:
So, should I invest some time in checking that we can have this in config only and not have problems with GUI?

AudriusButkevicius:
So I am actually against having this in the config too, as I don't think it would get much meddling. The config already has a lot of cruft, but most of the cruft is there to be able to switch off a feature, in this case there is nothing to switch off, and tweaks can be done by just extracting it into an external versioner.

AudriusButkevicius:
Yet that's me, let's see what others have to say

imsodin:
I agree this should not be in the folder edit modal, that would cause more confusion than it would be useful. I am hesitant to shut something down entirely just because there will be few users. Especially when the alternative is scripting, which is a huge barrier for many users. On the other hand I am skeptical that the kind of user that doesn't want to do scripting is interested in tweaking this parameters.

TLDR:  
I don't have any decisive new arguments to bring to the table. I slightly tend towards allowing it in the config on the basis that it won't do much harm.

trmfreitas:
Ok I will experiment having only the config.xml option. 
Thank you.

imsodin:
Not much to experiment there: Remove all the gui stuff and you are set. However at the moment nothing seems yet to be decided whether this will make it in or not (and overall tendency is still not).

trmfreitas:
Yes, thanks.

Em 19/10/2017 7:04 p.m., "Simon Frei" <notifications@github.com> escreveu:

AudriusButkevicius:
@calmh make the call

calmh:
No. The staggered versioner is already too complicated; we should not make it worse.

AudriusButkevicius:
Senior Techical Architect has spoken, amen.


Sorry about that @trmfreitas

lkwg82:
Thx for your PR. Next time I encourage you to scribble and discuss the idea in the forum to avoid declined PR.

I went the same way ;).

trmfreitas:
No problem. I did this cause I needed this kind of versioning to use
syncthing as a backup solution on one of my machines. With this I am able
to schedule backups to be kept for one week, one per day, one month, one
per week, and every month of the year. That is why I did it and then I
thought that it would be great if I didn't have to use my own compiled
version so I thought about checking if this was interesting for someone
else.

Thank you all for your time and a very nice product. 😀

Issue #3136, opened 2016-05-21
Toggling HTTP/HTTPS breaks GUI/API


calmh:
Toggling the setting and saving the config in the GUI results in the log line:

```
[SYNO4] 20:46:01 WARNING: API: accept tcp 127.0.0.1:8385: use of closed network connection
```

And then there is nothing listening on the GUI port any more.

v0.13.1

calmh:
So now of course I can't reproduce this myself any more...

norgeous:
I can get this same error / similar by rapidly setting `Use HTTPS for GUI` on and off or by rapidly switching between themes. It happens if you ignore the 'restart needed' warning message, and just keep saving settings.

norgeous:
just got the following, using the above method. idk if helpful.
```log
[W6SBY] 01:00:36 WARNING: API: accept tcp 127.0.0.1:8384: use of closed network connection
2016/05/29 01:00:45 http: panic serving 127.0.0.1:58129: runtime error: invalid memory address or nil pointer dereference
goroutine 618 [running]:
net/http.(*conn).serve.func1(0xc0839e6000)
	c:/go/src/net/http/server.go:1389 +0xc8
panic(0xb792e0, 0xc082002070)
	c:/go/src/runtime/panic.go:443 +0x4f7
main.(*apiService).CommitConfiguration(0xc082053e00, 0xf, 0xc083bf0400, 0x1, 0x4, 0xc082037d40, 0x2, 0x4, 0x1, 0xc0843b9cb0, ...)
	c:/jenkins/workspace/syncthing-release-windows/src/github.com/syncthing/syncthing/cmd/syncthing/gui.go:402 +0x145
github.com/syncthing/syncthing/lib/config.(*Wrapper).replaceLocked(0xc082001b80, 0xf, 0xc0848bcc00, 0x1, 0x4, 0xc0830501a0, 0x2, 0x4, 0x1, 0xc0826396e0, ...)
	c:/jenkins/workspace/syncthing-release-windows/src/github.com/syncthing/syncthing/lib/config/wrapper.go:153 +0x73b
github.com/syncthing/syncthing/lib/config.(*Wrapper).Replace(0xc082001b80, 0xf, 0xc0848bcc00, 0x1, 0x4, 0xc0830501a0, 0x2, 0x4, 0x1, 0xc0826396e0, ...)
[...]

PR #4433, opened 2017-10-18
Webui: add support for multiple gui configurations

lstordeur:
### Purpose

This new PR supersedes #4402 according to the technical choices which has been discussed.

This PR brings support for multiple GUI configurations (and as a side effect, multiple gui listeners).

Detailed changes:
- in the configuration file, the `gui` tag can now be given several times;
- the `STGUIADDRESS` environment variable will now override a whole GUI configuration object instead of just the listener part;
- a new environment variable `STGUIADDRESSES`, accepting a comma-separated syntax, allows to override the default GUI configuration with multiple GUI configuration objects.

Example:

```
    <gui enabled="true" debugging="false" tls="true">
      <address>127.0.0.1:8384</address>
      <insecureAdminAccess>true</insecureAdminAccess>
      <insecureSkipHostCheck>true</insecureSkipHostCheck>
      <apikey>BLtGw3CM9AnBSBaISeq0FPdHCzaY5zG5</apikey>
      <theme>dark</theme>
    </gui>
    <gui enabled="true" debugging="true">
      <address>127.0.0.1:44444</address>
      <insecureAllowFrameLoading>true</insecureAllowFrameLoading>
      <apikey>BLtGw3CM9AnBSBaISeq0FPdHCzaY5zG5</apikey>
      <theme>dark</theme>
    </gui>
```

```
STGUIADDRESSES="https://localhost:8384,http://localhost:44444" syncthing
```

### Testing

- defining a  couple of gui configurations in the config file with different ports, one config TLS-enabled, the other not;
- overriding the configuration file throufh `STGUIADDRESS`;
- overriding the configuration file through the new variable `STGUIADDRESSES` by giving a couple of URLs listening on different ports, with one TLS-enabled, the other not.

### Documentation

The documentation will have to be updated to reflect this new feature.


lstordeur:
So, in its current state, I believe I have broken GUI theme update support when committing a new configuration. Indeed (and this is still the same issue as discussed before), from `CommitConfiguration()`, I cannot access the static server instance as it no longer lives in the API service structure but is instantiated when building the handlers.

cmd/syncthing/gui.go
```diff
@@ -191,16 +191,18 @@ func (s *apiService) getListener(guiCfg config.GUIConfiguration) (net.Listener,
...
-	listener := &tlsutil.DowngradingListener{
-		Listener:  rawListener,
-		TLSConfig: tlsCfg,
+	listeners := make([]net.Listener, 0, len(guiCfgs))
+	for _, guiCfg := range guiCfgs {
+		rawListener, err := net.Listen("tcp", guiCfg.Address)
+		if err != nil {
+			return nil, err
+		}
+		listeners = append(listeners, &tlsutil.DowngradingListener{
```
AudriusButkevicius: We should unwind and close all listeners in case of an error, or skip listeners that fail and abort if none exist.

cmd/syncthing/gui.go
```diff
@@ -369,12 +383,23 @@ func (s *apiService) Serve() {
 		close(s.startedOnce)
 	}
 
-	// Serve in the background
+	// Serve in the background for each listener
 
 	serveError := make(chan error, 1)
-	go func() {
```
AudriusButkevicius: I don't think the channel needs to be buffered? Can you check thw history why its buffered?

lstordeur: Blamed to the following commit:
```
commit 5aacfd16395efc6a375e21113ded6cb3b0d5783e
Author: Jakob Borg <jakob@nym.se>
Date:   Mon Jun 6 22:12:23 2016 +0000

    cmd/syncthing: Make API serve loop more robust (fixes #3136)
    
    This sacrifices the ability to return an error when creating the service
    for being more persistent in keeping it running.
    
    GitHub-Pull-Request: https://github.com/syncthing/syncthing/pull/3270
    LGTM: AudriusButkevicius, canton7
```
The `serveError` channel has been buffered since its birth.

AudriusButkevicius: @canton7, @calmh remember why?

cmd/syncthing/gui.go
```diff
@@ -306,16 +290,15 @@ func (s *apiService) Serve() {
 	// The main routing handler
 	mux := http.NewServeMux()
 	mux.Handle("/rest/", restMux)
+
 	mux.HandleFunc("/qr/", s.getQR)
 
 	// Serve compiled in assets unless an asset directory was set (for development)
-	mux.Handle("/", s.statics)
+	mux.Handle("/", newStaticsServer(guiCfg.Theme, s.assetDir))
```
AudriusButkevicius: You could have map[config.GUIConfiguration]staticServer (or some other better key for the map) map on the API service which calls setTheme on the right one as it detects the change. Or to be honest it could just call it on every one every config save, as all that changes is just a string.

cmd/syncthing/gui.go
```diff
@@ -789,16 +829,19 @@ func (s *apiService) postSystemConfig(w http.ResponseWriter, r *http.Request) {
...
-	if to.GUI.Password != s.cfg.GUI().Password {
-		if to.GUI.Password != "" {
-			hash, err := bcrypt.GenerateFromPassword([]byte(to.GUI.Password), 0)
-			if err != nil {
-				l.Warnln("bcrypting password:", err)
-				http.Error(w, err.Error(), http.StatusInternalServerError)
-				return
-			}
+	for i, guiCfg := range s.cfg.GUIs() {
+		toGuiCfg := &to.GUIs()[i]
```
AudriusButkevicius: What happens if we add or remove a config... This might end up out of index.
I think you need some reasonable way to compare the two types of structs (or have some reliable map key) to do lookups.

cmd/syncthing/gui.go
```diff
@@ -400,24 +428,34 @@ func (s *apiService) String() string {
...
+	}
+	return nil
 }
 
 func (s *apiService) CommitConfiguration(from, to config.Configuration) bool {
-	// No action required when this changes, so mask the fact that it changed at all.
-	from.GUI.Debugging = to.GUI.Debugging
+	configChanged := false
+	for i, toGuiCfg := range from.GUIs() {
+		fromGuiCfg := &to.GUIs()[i]
```
AudriusButkevicius: Smae out of index argument

cmd/syncthing/main.go
```diff
@@ -559,14 +564,15 @@ func performUpgrade(release upgrade.Release) {
...
+	firstGuiCfg := cfg.GUIs()[0]
+	u, err := url.Parse(firstGuiCfg.URL())
 	if err != nil {
 		return err
 	}
 	u.Path = path.Join(u.Path, "rest/system/upgrade")
 	target := u.String()
 	r, _ := http.NewRequest("POST", target, nil)
-	r.Header.Set("X-API-Key", cfg.GUI().APIKey)
+	r.Header.Set("X-API-Key", firstGuiCfg.APIKey)
```
AudriusButkevicius: I'd do this as a loop until first success.

cmd/syncthing/main.go
```diff
@@ -1083,7 +1095,7 @@ func setupGUI(mainService *suture.Supervisor, cfg *config.Wrapper, m *model.Mode
 		// Can potentially block if the utility we are invoking doesn't
 		// fork, and just execs, hence keep it in its own routine.
 		<-api.startedOnce
-		go openURL(guiCfg.URL())
+		go openURL(cfg.GUIs()[0].URL())
```
AudriusButkevicius: This needs the // WARNING

cmd/syncthing/usage_report.go
```diff
@@ -245,7 +245,6 @@ func reportData(cfg configIntf, m modelIntf, connectionsService connectionsIntf,
 		}
 		res["folderUsesV3"] = folderUsesV3Interface
 
-		guiCfg := cfg.GUI()
 		// Anticipate multiple GUI configs in the future, hence store counts.
```
AudriusButkevicius: You can remove this comment in this commit.

lib/config/config.go
```diff
@@ -225,6 +244,12 @@ found:
 		return err
 	}
 
+	// Add a default GUI configuration if no one has been given.
+	if len(cfg.RawGUIs) == 0 {
+		cfg.RawGUIs = append(cfg.RawGUIs, GUIConfiguration{})
+		util.SetDefaults(&cfg.RawGUIs[0])
+	}
```
AudriusButkevicius: This is a bit wishy washy. Create, add to slice, then index access back from the slice.

Maybe this:

```
         // Add a default GUI configuration if no one has been given.
        if len(cfg.RawGUIs) == 0 {
                var guiCfg GUIConfiguration
                util.SetDefaults(&guiCfg)
		cfg.RawGUIs = []GUIConfiguration{guiCfg}
        }
```

cmd/syncthing/usage_report.go
```diff
@@ -260,40 +259,42 @@ func reportData(cfg configIntf, m modelIntf, connectionsService connectionsIntf,
...
+				guiStats["enabled"]++
+				if guiCfg.UseTLS {
+					guiStats["useTLS"]++
+				}
+				if len(guiCfg.User) > 0 && len(guiCfg.Password) > 0 {
+					guiStats["useAuth"]++
+				}
+				if len(guiCfg.APIKey) > 0 {
+					guiStats["useAPIKey"]++
+				}
```
AudriusButkevicius: Can you also remove this, it seems the API key is always set.

cmd/syncthing/gui.go
canton7: #3136 seems to have a detailed breakdown.

AudriusButkevicius: That explains the issue, yet I am not sure I understand why it needs to be buffered.

cmd/syncthing/gui.go
lstordeur: Do we really wish to support addition/removal of gui configuration in a dynamic way?

AudriusButkevicius: Why not?

AudriusButkevicius: We support changing most of the settings online as it stands now.

lstordeur: Fine. So, this implies to implement some sort of gui config lookup. I propose to use the `Address` field as key for the lookup. Does it sound good for you?

AudriusButkevicius: Makes sense as these have to be unique.

cmd/syncthing/main.go
lstordeur: To be sure sure we're on the same page: should I loop on the whole `upgradeViaRest()` code for each gui (so including request sending and response status code checking) or just the part which forges the request (so up to the `r.Header.Set()` call)?

cmd/syncthing/gui.go
lstordeur: Here again, maybe we can use the `Address` as key for the `staticServer` map.

cmd/syncthing/main.go
AudriusButkevicius: No, try all UI urls, in case some of them are unix sockets for example.

lstordeur: So I suggest to add the `// WARNING` here to address the case when we'll have UNIX socket support.

cmd/syncthing/gui.go
AudriusButkevicius: Sure

cmd/syncthing/main.go
AudriusButkevicius: Fine

cmd/syncthing/gui.go
```diff
@@ -369,12 +390,23 @@ func (s *apiService) Serve() {
...
+		srv := http.Server{
+			Handler: handlers[i],
+			// ReadTimeout must be longer than SyncthingController $scope.refresh
+			// interval to avoid HTTP keepalive/GUI refresh race.
+			ReadTimeout: 15 * time.Second,
+		}
+		go func(listener net.Listener) {
+			select {
+			case serveError <- srv.Serve(listener):
+			default:
```
imsodin: Why the select-default? AFAIK this means that there will never be sent anything to `serveError` as `srv.Serve` does not return immediately.

lib/config/guiconfiguration.go
```diff
@@ -26,42 +28,27 @@ type GUIConfiguration struct {
...
-		// an URL if it contains a slash. If that fails, return it as is and let
-		// some other error handling handle it.
-
-		if strings.Contains(override, "/") {
-			url, err := url.Parse(override)
-			if err != nil {
-				return override
-			}
-			return url.Host
+func GUIConfigFromString(envAddr string) (c GUIConfiguration) {
```
imsodin: I think you need to run `c` through `util.SetDefaults` first, otherwise it will e.g. be disabled due as all members will be initialized at their zero values.

lib/config/config.go
```diff
@@ -162,6 +159,38 @@ type Configuration struct {
...
+		// it.
+		var overrideGUIs []GUIConfiguration
+		for _, overrideEntry := range strings.Split(override, ",") {
+			overrideGUIs = append(overrideGUIs, GUIConfigFromString(overrideEntry))
+		}
+		return overrideGUIs
+	} else if override := os.Getenv("STGUIADDRESS"); override != "" {
+		// Legacy overriding form which only supports one address.
+		return []GUIConfiguration{GUIConfigFromString(override)}
+	}
```
imsodin: As far as I understand it before only the address was set by the env var, all other GUI settings were still taken from the config file. I don't really see how to do that with a variable amount of GUIs in a sensible way.
One option would be to match given addresses with the list of GUI configs and use default values if there are more addresses given than there are GUIs in the config. That seems a bit random though. Another thing would be to just use default values when the env var is used. Honestly I don't really know the use case for this, so I am fishing in the dark.

cmd/syncthing/gui.go
```diff
@@ -400,24 +432,50 @@ func (s *apiService) String() string {
...
 
-	if to.GUI.Theme != from.GUI.Theme {
-		s.statics.setTheme(to.GUI.Theme)
+		configChanged = true
 	}
+	for _, fromGuiCfg := range from.GUIs() {
+		toGuiCfg := to.GUIByAddress(fromGuiCfg.Address)
+		if toGuiCfg == nil {
+			// A GUI configuration has been removed. GUI restart is needed.
+			return false
```
imsodin: Wouldn't `configChanged=true` be enough? I don't think a full restart of Syncthing is needed, rerunning `Serve` is enough, isn't it?

cmd/syncthing/gui.go
AudriusButkevicius: Good catch.
What needs to happen is
```
err := srv.Serve(listener)
select {
case serverError <- err:
default:
}
```

AudriusButkevicius: Actually it might still work, as I suspect srv.Serve needs to be evaluated.
Needs testing.

cmd/syncthing/gui.go
```diff
@@ -329,36 +316,70 @@ func (s *apiService) Serve() {
...
+			// This is during initialization. A failure here should be fatal
+			// as there will be no way for the user to communicate with us
+			// otherwise anyway.
+			l.Fatalln("Starting API/GUI:", err)
+		}
+	}
+
+	if len(listeners) == 0 {
+		// Not much we can do here other than exit quickly. The supervisor
+		// will log an error at some point.
```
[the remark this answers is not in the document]
AudriusButkevicius: To be honest, the UI could be disabled.
We didn't scream loud before, and we probably should, but it needs to happen with sanity checks that we are expected to have a UI running.

cmd/syncthing/gui.go
```diff
@@ -329,36 +316,70 @@ func (s *apiService) Serve() {
...
+
+func (s *apiService) Serve() {
+	guiCfgs := s.cfg.GUIs()
+	listeners, err := s.getListeners(guiCfgs)
+	if err != nil {
+		select {
+		case <-s.startedOnce:
+			// We let this be a loud user-visible warning as it may be the only
+			// indication they get that the GUI won't be available.
+			l.Warnln("Starting API/GUI:", err)
```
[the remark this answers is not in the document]
AudriusButkevicius: I wouldn't define the string outside of the variable, this get's put in the data area anyway, as it's a constant.

cmd/syncthing/gui.go
```diff
@@ -293,11 +276,15 @@ func (s *apiService) Serve() {
 
 	// Debug endpoints, not for general use
 	debugMux := http.NewServeMux()
-	debugMux.HandleFunc("/rest/debug/peerCompletion", s.getPeerCompletion)
```
AudriusButkevicius: debugMux only needs to be declared in the if branch.
In the debugging = False case you can replace Handler with HandlerFunc that just returns the error inline,
instead of having a function that returns a function that does nothing (debugMux in the function is not used)

cmd/syncthing/gui.go
AudriusButkevicius: Yeah, service restart would rewind all UIs

cmd/syncthing/gui.go
```diff
@@ -789,16 +849,19 @@ func (s *apiService) postSystemConfig(w http.ResponseWriter, r *http.Request) {
...
-		if to.GUI.Password != "" {
-			hash, err := bcrypt.GenerateFromPassword([]byte(to.GUI.Password), 0)
-			if err != nil {
-				l.Warnln("bcrypting password:", err)
-				http.Error(w, err.Error(), http.StatusInternalServerError)
-				return
-			}
+	for _, toGuiCfg := range to.GUIs() {
+		fromGuiCfg := s.cfg.GUIByAddress(toGuiCfg.Address)
+		if (fromGuiCfg == nil) || toGuiCfg.Password != fromGuiCfg.Password {
```
AudriusButkevicius: No need for braces

lib/config/config.go
```diff
@@ -152,7 +149,7 @@ type Configuration struct {
 	Version        int                   `xml:"version,attr" json:"version"`
 	Folders        []FolderConfiguration `xml:"folder" json:"folders"`
 	Devices        []DeviceConfiguration `xml:"device" json:"devices"`
-	GUI            GUIConfiguration      `xml:"gui" json:"gui"`
+	RawGUIs        []GUIConfiguration    `xml:"gui" json:"gui"`
```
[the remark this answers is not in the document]
AudriusButkevicius: So old XML I suspect is still fine, as it's one element out of many with the same tags. Json needs renaming though, and notify people about the breaking change.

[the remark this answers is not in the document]
AudriusButkevicius: GUIs() is taken by the function that adds them from env vars.

lib/config/config.go
AudriusButkevicius: This is overridden by wrappers, instead of passing in the command line I think.
Use TLS and other crap had assessors would check the env var, so I suspect this is still ok.

lib/config/guiconfiguration.go
AudriusButkevicius: Good catch, the fact that the author hasn't caught this gets me a bit worried about the level of testing done.

cmd/syncthing/gui.go
lstordeur: I believe your right concerning your last point. Actually, I have wondered whether I really have to evaluate the `Serve()` call before the `select`. So, I had a look on the semantic specification of the `select` statement. Here it is (as picked up from https://golang.org/ref/spec#Select_statements)

 Execution of a "select" statement proceeds in several steps:

1. For all the cases in the statement, the channel operands of receive operations and the channel and right-hand-side expressions of send statements are evaluated exactly once, in source order, upon entering the "select" statement. The result is a set of channels to receive from or send to, and the corresponding values to send. Any side effects in that evaluation will occur irrespective of which (if any) communication operation is selected to proceed. Expressions on the left-hand side of a RecvStmt with a short variable declaration or assignment are not yet evaluated.
2. If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection. Otherwise, if there is a default case, that case is chosen. If there is no default case, the "select" statement blocks until at least one of the communications can proceed.
3. Unless the selected case is the default case, the respective communication operation is executed.
4. If the selected case is a RecvStmt with a short variable declaration or an assignment, the left-hand side expressions are evaluated and the received value (or values) are assigned.
5. The statement list of the selected case is executed.

cmd/syncthing/gui.go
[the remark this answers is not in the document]
imsodin: The main point is about the information in the error message, the rest is just nit-picking on an ultra-nit-picky level.
My "reasoning" (on the nit-pick) is about convenience, nothing technical: String in variable means just having one place to modify the string, compared to altering two identical strings in both `l.Fatalln` calls.

cmd/syncthing/gui.go
imsodin: Point taken, everything is correct as is. And probably according to typical Go-style (as terse as possible) this is the right way to do it. I would prefer Audrius variant as it is more obvious what happens, but then again I should just remember how select works.

cmd/syncthing/gui.go
imsodin: I believe that without the 1-buf it is a race between `srv.Serve` returning/sending to `serveError` and control flow in `apiService.Serve` reaching the final select statement where it can read from `serveError`.

cmd/syncthing/gui.go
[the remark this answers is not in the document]
imsodin: Agreed, not something to be done in this PR. I mostly just jumped at commenting when reading this, because as far as my experience goes the suture stuff is nice in that it keeps running but (in it's current form) does hide serious problems.

In short: Nothing to be solved/discussed in this PR (if at all).

lib/config/config.go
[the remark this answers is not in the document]
imsodin: > So old XML I suspect is still fine, as it's one element out of many with the same tags. Json needs renaming though, and notify people about the breaking change.

I don't understand why there is a breaking change. As far as I see the GUI config stays the same, there is just potentially more of it, so an automatic conversion could just pick up the existing config, write it as a single entry into an array of configs and be done with it.

lib/config/guiconfiguration.go
imsodin: That seems a bit harsh, there is still lots of reviewing going on and tests don't even pass yet. I am positive in that stadium lots of my PRs have errors of that magnitude in them, exactly because there wasn't much (any) real word testing done yet.

lib/config/config.go
imsodin: What I mean is that e.g. `apiService.getHandler` gets the config from a `cfg.Wrapper` (as everything else I assume) and that queries this function, which will return a default gui config except for the address which is set by the env var. Previously it got the "normal" gui config and got the address by calling `GUIConfiguration.Address` which respected the env var. Thus options like user and password were preserved when using the env var, which they aren't anymore.

lib/config/guiconfiguration.go
AudriusButkevicius: Agree, perhaps it's not yet tested as it's deemed early, so I am too quick to jump to conclusions.

lib/config/config.go
AudriusButkevicius: Ok, that's a fair argument, but I can't think of how to handle this properly.
Perhaps we should support `http(s)?://user:pass@host.com`. I think there is no way to set the API key from env var?

AudriusButkevicius: Regardless what we do, this is going to end up a breaking change, so perhaps we just remove STGUIADDRESS, leave STGUIADDRESSESS that supports specifying user:pass.

calmh:
Not following the reviews and stuff really, just jumping in to say that I think it's probably a valid config to not have any gui config / listener at all, once the config is that there are 0..n listeners. A Syncthing instance can run fine from a config file generated  by something else, and with `-verbose` or similar does provide all the relevant info as log output. So I don't think we should spend effort to prevent this case.

cmd/syncthing/gui.go
lstordeur: Fine

lib/config/config.go
lstordeur: Fine for me. I will just wait a definitive confirmation to move forward in this direction. I am just wondering whether we really have to remove `STGUIADDRESS`. We could also extend the supported syntax for this legacy environment variable by adding support for the `user:pass` form.

lib/config/guiconfiguration.go
lstordeur: Indeed. I would have caught this bug if I had tested the `STGUIADDRESS(ES)` environment overriding feature. The point is that as this stadium, my validation procedure is still very light, given the yet high rate of review activity on this PR. I'll try to harden a bit my validation steps.

cmd/syncthing/gui.go
lstordeur: Fine

lib/config/config.go
AudriusButkevicius: @calmh agree?
This changes gui config from object to array in the json world anyway, which is a breaking change, so I think we can axe that old syntax.

lib/config/config.go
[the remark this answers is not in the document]
AudriusButkevicius: Json now is:
```
gui: { 
   User...
}
```
Becomes
```
gui: [
  {
      User: ...
  }
]
```

So synctrayzor et all break, hence this breaks api compatability.

lib/config/config.go
calmh: I think the STGUIADDRESS variable is intended for wrappers and similar things mostly? As such I think it's only used/useful to say "I want to be able to access the GUI/API *here*" and I have a hard time imagining a situation where being able to specify multiple would make sense. I would just keep STGUIADDRESS and have it be in addition to whatever else is configured in the config. 

The API key can currently be set with STGUIAPIKEY. I think it makes sense to be able to specify that from the outside, again for wrappers. I'm not sure of the use case of being able to specify username and password from an environment variable.

lib/config/config.go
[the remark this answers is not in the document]
calmh: Pinging @canton7 @Nutomic @kozec to have an opinion here. We can potentially continue to *accept* a single object in input, being flexible in what we output would probably be more annoying.

lib/config/config.go
AudriusButkevicius: My point is that this is a breaking change anyway, so I don't see a reason why we could not remove the old thing and add a new thing that does a bit more, such as extracts username and password from the URL, and supports multiple. Not saying that it will be used for multiple listeners, but the ability to specify that does not do any damage.

AudriusButkevicius: Yeah, we can keep the old thing for compatability,but to me it seems like a good opportunity to deprecate now, rather than later.

imsodin: Assuming we want to keep compatibility, I still don't see a consistent way to do this yet:
Before: All GUI config parameters are taken from config except url/port and key if given by env.
Now: Url/port and key can still be set by env, but then all other values are default, regardless of what is in the config.
To have the same/similar behaviour we could for example just combine the values given with the parameters in the first GUI config. That is close to what it was before, but seems a bit random in case of multiple GUI configs (what designates the "first" config). Actually that might not even be consistent, as order in json is random, if I remember correctly. I don't know whether this matters at all for those who actually use this env vars, I am just pointing out the change in behaviour which doesn't seem to be on the radar.

st-review:
:robot: *beep boop*

PR #4435, opened 2017-10-19
gui: versioner->versioning and add missing translations

imsodin:
All over the UI and docs we use "versioning", except for this one instance where it is "versioner". Minor issue, but it caused [confusion for translation](https://www.transifex.com/syncthing/syncthing/translate/#de/gui/117529706?q=external). Also strings for the new usage reporting were missing.


gui/default/syncthing/settings/settingsModalView.html
```diff
@@ -142,7 +142,8 @@
             <div ng-if="tmpOptions.upgrades != 'candidate'">
               <label translate for="urVersion">Anonymous Usage Reporting</label> (<a href="" translate data-toggle="modal" data-target="#urPreview">Preview</a>)
               <select class="form-control" id="urVersion" ng-model="tmpOptions._urAcceptedStr">
-                <option ng-repeat="n in urVersions()" value="{{n}}">Version {{n}}</option>
+                  <option ng-repeat="n in urVersions()" value="{{n}}">{{'Version' | translate}} {{n}}</option>
+                  <!-- <option ng-repeat="n in urVersions()" value="{{n}}"><translate>Version</translate> {{n}}</option> -->
                 <!-- 1 does not exist, as we did not support incremental formats back then. -->
```
AudriusButkevicius: Whats up with the comments?

AudriusButkevicius: And indent?

AudriusButkevicius:
This will cause everyone to have to retranslate, is it really worth it?

gui/default/syncthing/settings/settingsModalView.html
imsodin: No idea what it is with the comment, not of my doing, I didn't even notice before. Indenting is my wrongly adjusted editor -> will fix.

imsodin:
I do believe so. The two translations will be side by side, so it is a simple copy and paste and it will be both more consistent and less confusing. Often in translation meaning gets slightly altered even for consistent "initial state", so such little problems can already cause much bigger changes in meaning - that's what happened to the German translation which was simply wrong.

calmh:
@st-review merge

gui: "versioner" -> "versioning" and add missing translations

st-review:
:ok_hand: Merged as c704ba9ef9d18730d3c3a48f6c1ad78d9b16b950. Thanks, @imsodin!

PR #4449, opened 2017-10-21
lib/fs: Sync does not fail (fixes #4432)

### LINKED ISSUE #4432: fsync on directories fails if directory is located in a samba share
Since samba does not support fsync on directories the syncthing daemon gives a lot of "invalid argument" error. It would be great if fsync could be disabled/ignored for directories. 
I think this behaviour is already the default case for windows (5ec1490), but I would like to use it on Linux.

Syncthing Version: v0.14.39
OS Version: Linux / Ubuntu 17.04

AudriusButkevicius:
...

### DIFF
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -96,7 +96,7 @@ func (f *FolderConfiguration) CreateMarker() error {
 		}
 		if dir, err := fs.Open("."); err == nil {
 			if serr := dir.Sync(); err != nil {
-				l.Infof("fsync %q failed: %v", ".", serr)
+				l.Debugln("fsync %q failed: %v", ".", serr)
 			}
 		} else {
 			l.Infof("fsync %q failed: %v", ".", err)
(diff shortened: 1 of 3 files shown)


calmh:
Well this is the lazy one leaving all kinds of `if err != nil { log and shit }` that are now dead code... If we don't care whether this works (which I'm fine with in the regular sync path) the clean solution is to explicitly ignore the error at the call site.

AudriusButkevicius:
Fair point.

AudriusButkevicius:
Updated

lib/config/folderconfiguration.go
```diff
@@ -96,7 +96,7 @@ func (f *FolderConfiguration) CreateMarker() error {
 		}
 		if dir, err := fs.Open("."); err == nil {
 			if serr := dir.Sync(); err != nil {
```
calmh: There is a different bug here where the `if` sets `serr` bug checks `err`

calmh: Eh, I can fix that

calmh:
@st-review merge

st-review:
:ok_hand: Merged as 622b614f315b1ba145f81720e46d31ed7407c569. Thanks, @AudriusButkevicius!

PR #4450, opened 2017-10-21
lib/model: Start/stop fs watch when folder error is cleared/set (fixes #4445)

### LINKED ISSUE #4445: Filesystem watcher doesn't handle restarted folders
I have my photos on an external SSD. It's a folder by itself, and usually not connected. When it is connected it gets picked up by a scan and then things work. FS watching doesn't seem compatible with this.

Expected behavior: watching starts when a scan notices the folder is alive again.

Actual behavior: no watching, ever.

imsodin:
The approach seems a bit hackish, but the only alternative I see is in my opinion worse: Subscribe to folder state events in the folder itself (which are generated by an embedded type).

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -25,7 +25,7 @@ type folder struct {
 	initialScanFinished chan struct{}
 	watchCancel         context.CancelFunc
 	watchChan           chan []string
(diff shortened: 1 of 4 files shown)


imsodin:
Will soon be replaced by a PR also addressing #4451.

Issue #4446, opened 2017-10-21
KCP gets confused when restarting a device


calmh:
I think we're back to (or just still at) some issue with connections not being established after being broken. For example, when I restart Syncthing on my laptop (`SYNO4VL`) connected to my server (`ITZRNXE`):

Laptop logs:

```
[SYNO4] 19:08:27 INFO: Started filesystem watcher for folder "Transfer" (4vcqq-sciqk)
[SYNO4] 19:08:27 INFO: Started filesystem watcher for folder "foto/2017" (2fyjq-jrfpr)
[SYNO4] 19:08:27 INFO: Started filesystem watcher for folder "Lightroom" (lightroom)
[SYNO4] 19:08:27 INFO: Started filesystem watcher for folder documents
[SYNO4] 19:08:27 INFO: Completed initial scan of readwrite folder "Transfer" (4vcqq-sciqk)
[SYNO4] 19:08:27 INFO: Completed initial scan of readonly folder "Lightroom" (lightroom)
[SYNO4] 19:08:27 INFO: Completed initial scan of readonly folder documents
[SYNO4] 19:08:28 INFO: Completed initial scan of readonly folder "foto/2017" (2fyjq-jrfpr)
[SYNO4] 19:08:30 INFO: Device ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM is "pseudo.nym.se" at [kcp://pseudo.int.nym.se:12345 dynamic]
[SYNO4] 19:08:30 INFO: Device SYNO4VL-TKVFF25-QKC2T3V-TTK4BGU-D2ZA5OT-PDKSI3R-3P4JQVS-BL4VXAC is "syno.nym.se" at [dynamic]
[SYNO4] 19:08:30 INFO: No automatic upgrades; STNOUPGRADE environment variable defined.
[SYNO4] 19:08:30 INFO: GUI and API listening on 127.0.0.1:8385
[SYNO4] 19:08:30 INFO: Access the GUI via the following URL: http://127.0.0.1:8385/
[SYNO4] 19:08:30 INFO: Established secure connection to ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM at [::]:22020-172.16.32.11:12345 (kcp-client) (TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305)
[...]

PR #4452, opened 2017-10-22
lib/connections: Use own KCP fork, move listener setup earlier (ref #4446)

AudriusButkevicius:
...


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 0d3016635744d79fc9a94bca8aff25e42a7718d8. Thanks, @AudriusButkevicius!

AudriusButkevicius:
@calmh pls take this for a spin and see if you can still repro the issue.

calmh:
I think it mostly works. It takes a while for the connection to time out on the non-restarted side, and in the meantime it drops new attempts as "already connected" which looks like "EOF" to the other side. After about five minutes it connected.

AudriusButkevicius:
Did you pull the extra commit I pushed on master? If not then you are just failing to repro.

calmh:
The blacklist thing? Nope, pulled just before it. So apparently it's not 100% either way. 🤷‍♂️

AudriusButkevicius:
Its definately not 100%, but tye blacklist prevents the symptoms when I see it.

AudriusButkevicius:
Can you with the blacklist and see if it still happens?

calmh:
Not yet, but I think the NAT thing and this is the cause (or, another cause) for reconnects not happening in a timely fashion:

screen shot 2017-10-22 at 23 55 48

Trying and timing out each of those would take a while probably

AudriusButkevicius:
I think we are lacking parallel dials? 😆

AudriusButkevicius:
Technically, with my change, even symmetric NAT shouldn't provide so many addresses.

calmh:
Yes I think this is fixed

Issue #4451, opened 2017-10-22
Repeated "stopping folder" when the folder is already stopped


calmh:
Not sure if this is new or old.

```
[SYNO4] 00:05:59 INFO: Device ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM client is "syncthing v0.14.39" named "pseudo.nym.se"
[SYNO4] 00:55:26 INFO: Stopping folder "foto/2017" (2fyjq-jrfpr) due to error: folder path missing
[SYNO4] 02:09:03 INFO: Stopping folder "foto/2017" (2fyjq-jrfpr) due to error: folder path missing
[SYNO4] 03:18:50 INFO: Stopping folder "foto/2017" (2fyjq-jrfpr) due to error: folder path missing
[SYNO4] 04:24:07 INFO: Stopping folder "foto/2017" (2fyjq-jrfpr) due to error: folder path missing
[SYNO4] 05:27:31 INFO: Stopping folder "foto/2017" (2fyjq-jrfpr) due to error: folder path missing
[SYNO4] 06:37:45 INFO: Stopping folder "foto/2017" (2fyjq-jrfpr) due to error: folder path missing
[SYNO4] 07:32:30 INFO: Stopping folder "foto/2017" (2fyjq-jrfpr) due to error: folder path missing
[SYNO4] 08:44:55 INFO: Stopping folder "foto/2017" (2fyjq-jrfpr) due to error: folder path missing
[SYNO4] 10:27:38 INFO: Connection to ITZRNXE-YNROGBZ-HXTH5P7-VK5NYE5-QHRQGE2-7JQ6VNJ-KZUEDIU-5PPR5AM closed: read timeout
```

(those are all the log entries for that time span)

No reason to mention stopping the folder when it hasn't been started in between. I have a one hour scan interval, so it looks like it logs on every scan (what with the randomization of the scan times etc).

v0.14.39+25-gb1ade6d0

PR #4455, opened 2017-10-22
lib/model, lib/config: Refactor folder health/error handling (fixes #4445 #4451)

### LINKED ISSUE #4445: Filesystem watcher doesn't handle restarted folders

imsodin:
While fixing the two linked bugs the result grew uglier and uglier, so I decided to refactor. This splits code regarding folder error/health handling from model.go into folder.go and the config package. The only change in behaviour regards logging and start/stopping fs watch on error state change.


lib/config/size.go
```diff
@@ -73,3 +75,24 @@ func (s Size) String() string {
...
+			return fmt.Errorf("insufficient space in %v %v: %f %% < %v", fs.Type(), fs.URI(), freePct, s)
+		}
+	} else {
+		if err == nil && float64(usage.Free) < val {
+			return fmt.Errorf("insufficient space in %v %v: %v < %v", fs.Type(), fs.URI(), usage.Free, s)
+		}
+	}
+
+	return nil
+}
```
AudriusButkevicius: All of this should live in `func (f *FolderConfiguration) CheckFreeSpace() (err error) {`, and keep using a type, as a type.

lib/config/wrapper.go
```diff
@@ -468,3 +470,15 @@ func (w *Wrapper) MyName() string {
 	cfg, _ := w.Device(myID)
 	return cfg.Name
 }
+
+// CheckHomeFreeSpace returns nil if the home disk has the required amount of
+// free space, or if home disk free space checking is disabled.
+func (w *Wrapper) CheckHomeFreeSpace() error {
+	w.mut.Lock()
+	minFree := w.cfg.Options.MinHomeDiskFree
+	w.mut.Unlock()
```
AudriusButkevicius: Could just call `w.Options()`

lib/model/rwfolder.go
```diff
@@ -214,8 +208,8 @@ func (f *sendReceiveFolder) Serve() {
 				continue
 			}
 
-			if err := f.model.CheckFolderHealth(f.folderID); err != nil {
-				l.Infoln("Skipping pull of", f.Description(), "due to folder error:", err)
+			if err := f.CheckHealth(); err != nil {
+				l.Debugln("Skipping pull of", f.Description(), "due to folder error:", err)
```
AudriusButkevicius: We already log this in info or higher, so I'd just remove this

lib/model/model.go
```diff
@@ -2020,13 +2015,13 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
...
-			l.Infof("Stopping folder %s mid-scan due to folder error: %s", folderCfg.Description(), iterError)
+			l.Debugln("Stopping scan of folder %s due to: %s", folderCfg.Description(), err)
 			return iterError
 		}
 	}
 
-	if err := m.CheckFolderHealth(folder); err != nil {
-		l.Infof("Stopping folder %s mid-scan due to folder error: %s", folderCfg.Description(), err)
+	if err := runner.CheckHealth(); err != nil {
+		l.Debugln("Stopping scan of folder %s due to: %s", folderCfg.Description(), err)
```
AudriusButkevicius: We already log this in info or higher, so I'd just remove this

lib/model/model.go
```diff
@@ -2020,13 +2015,13 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 		})
 
 		if iterError != nil {
-			l.Infof("Stopping folder %s mid-scan due to folder error: %s", folderCfg.Description(), iterError)
+			l.Debugln("Stopping scan of folder %s due to: %s", folderCfg.Description(), err)
```
AudriusButkevicius: We already log this in info or higher, so I'd just remove this

lib/model/model.go
```diff
@@ -1914,7 +1909,7 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 		// The error we get here is likely an OS level error, which might not be
 		// as readable as our health check errors. Check if we can get a health
 		// check error first, and use that if it's available.
-		if ferr := m.CheckFolderHealth(folder); ferr != nil {
+		if ferr := runner.CheckHealth(); ferr != nil {
 			err = ferr
 		}
 		runner.setError(err)
```
AudriusButkevicius: Isn't this pointless?

lib/model/model.go
```diff
@@ -1926,8 +1921,8 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 
 	for f := range fchan {
 		if len(batch) == maxBatchSizeFiles || batchSizeBytes > maxBatchSizeBytes {
-			if err := m.CheckFolderHealth(folder); err != nil {
-				l.Infof("Stopping folder %s mid-scan due to folder error: %s", folderCfg.Description(), err)
+			if err := runner.CheckHealth(); err != nil {
+				l.Debugln("Stopping scan of folder %s due to: %s", folderCfg.Description(), err)
```
AudriusButkevicius: We already log this in info or higher, so I'd just remove this

lib/model/model.go
```diff
@@ -1938,8 +1933,8 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 		batchSizeBytes += f.ProtoSize()
 	}
 
-	if err := m.CheckFolderHealth(folder); err != nil {
-		l.Infof("Stopping folder %s mid-scan due to folder error: %s", folderCfg.Description(), err)
+	if err := runner.CheckHealth(); err != nil {
+		l.Debugln("Stopping scan of folder %s due to: %s", folderCfg.Description(), err)
```
AudriusButkevicius: We already log this in info or higher, so I'd just remove this

lib/config/size.go
imsodin: This code is also needed to get free space of home.

lib/model/model.go
imsodin: Not sure, the reason given in the comments seems sound enough to me?

lib/model/model.go
imsodin: Agreed, the only added information is, that it happened during scan, which should also be clear from context.

lib/config/size.go
AudriusButkevicius: Fine, so have a utility function or something, don't just wedge something on a type.

lib/model/model.go
AudriusButkevicius: A right, if it's err not ferr, it wouldn't be set otherwise.

lib/model/model.go
AudriusButkevicius: you can do CheckHealth("during scan") or something to explain where the health check is happening, and then log that.

lib/config/size.go
imsodin: Wedging was more or less a random decision, I just thought this makes sense here as it is used in the config package (before it was wedged on `Model`, but it has no connection to that at all). Would it make more sense as a utility function in the fs package?

AudriusButkevicius: I think so.

imsodin: config fs import cycle...
So either we leave it in config or I move the entire Size thing into the fs package.

AudriusButkevicius: Make a utility in config.

lib/model/model.go
imsodin: Actually, we do not necessarily log anything on info or higher unless the error is new or changed. So in a debug scenario, I think it is valid that we log the error anyway and from the position of origin.

AudriusButkevicius: Fine whatever.

AudriusButkevicius:
@st-review lgtm

st-review:

calmh:
@st-review lgtm

lib/model, lib/config: Refactor folder health/error handling (fixes #4445, fixes #4451)

st-review:
:ok_hand: Merged as dc42db444b17145b7de9a0d1c56a42b9d64f3dea. Thanks, @imsodin!

Issue #4443, opened 2017-10-21
lib/model: Filter out invalid symlinks on windows


Phoenix616:
Currently the folder puller will pause for a minute if it isn't making progress anymore. This happens for example if you have symlinks in your folder which will fail as they are not supported.

This request is for the puller to completely stop if all items that would be left are either not supported or fail so that it does not unnecessary run again and again and cause CPU spikes.

### Version Information

Syncthing Version: v0.14.39
OS Version: Windows 7 (64 bit)
Browser Version: SyncTrayzor 1.1.18

AudriusButkevicius:
And how do we ever recover if we stop forever?

Phoenix616:
Well it kinda seems that it is trying to transfer the same unsupported items every time instead of ignoring them?

AudriusButkevicius:
What if it's not unsupported items but files that are in use, or directories that cannot be deleted because they have files that are ignored, etc etc?

It shouldn't be trying to transfer unsupported files in the first place, what's the actual error you are seeing?

imsodin:
I assume the folder goes to error state and pulling is paused due to the `fileValid` check, which fails for symlinks on windows:
https://github.com/syncthing/syncthing/blob/master/lib/model/rwfolder.go#L380

I think this should not cause an error, as there is nothing the user can do. Removing the symlink may not be desirable in a mixed linux-windows cluster. I think there wouldn't be any harm in just ignoring symlinks on windows or just warn about it. I don't see any reason to put the entire folder in an error state and keep pulling to no apparent benefit.

AudriusButkevicius:
I think we filter out symlinks on the indexes we receive, hence why I want to see the error message, whether it's caused by us, or perhaps the underlying filesystem doesn't support symlinks making it an error of a different flavor.

imsodin:
I can't remember any checks at all when receiving indexes from when I recently fiddled with the ignores/invalid stuff.

Phoenix616:
The main issue is that it is retrying to sync the folder with the failed/unsupported items in it every minute (as the puller is paused for a minute by default) which results in a [CPU spike](https://i.imgur.com/JTh15Xs.png). (Orange is the syncthing.exe process, all other folders are paused) Pausing the syncing of that folder removes the spike. The spike also does not happen with other folders that do not have these unsupported files in them.

imsodin:
Can you please post logs to confirm where exactly the problem is as Audrius suggested - thanks.

[...]

PR #4460, opened 2017-10-24
all: add invalid/ignored files to global list (fixes #623)

### LINKED ISSUE #623: "Syncing" status should be based on activity, not completion percentage

imsodin:
This more or less just does what it says: It adds all already recorded and new invalid files into the global list. Then I adjusted the rest of the code, such that it works the same way as before except for remote completion calculation, which now knows about remote invalid/ignored files and thus does not consider them for completion.

Testing:
Manual short time testing, i.e. I ran it on production system until no activity, then checked that it wasn't totally broken (i.e. syncs happened) and then reverted back to "normal". 
The db transition went actually very fast, but the ensuing index exchange was very heavy in my case (I apparently have had ignores on and off a lot). Heavy as in it took quite some time/cpu. In the end I had only in sync remote devices, which was a nice change :)
I think existing tests cover this area already quite well, but I didn't look into that much yet, I wanted to get general feedback on the approach first.


imsodin:
I reproduced #4443 by accident, meaning I will fix it in here.

imsodin:
I still fixed #4443 in this PR, as my approach for fixing it requires this PR and this PR requires the fix to pass tests -> commit will have to be adjusted.

lib/model/rwfolder.go
```diff
@@ -369,22 +372,47 @@ func (f *sendReceiveFolder) pullerIteration(ignores *ignore.Matcher) int {
...
+			f.newError("need", intf.FileName(), fs.ErrInvalidFilename)
 			changed++
 			return true
 		}
 
 		file := intf.(protocol.FileInfo)
 
 		switch {
+		case ignores.ShouldIgnore(file.Name):
+			invalidFile := protocol.FileInfo{
```
calmh: This desperately needs a helper method on FileInfo or so. To be used also in the other one or two places we duplicate and slightly change a fileinfo. Otherwise we are sure to forget some of these when we add fields.

lib/model/rwfolder.go
```diff
@@ -444,19 +489,22 @@ func (f *sendReceiveFolder) pullerIteration(ignores *ignore.Matcher) int {
 				// number, hence the deletion coming in again as part of
 				// WithNeed, furthermore, the file can simply be of the wrong
 				// type if we haven't yet managed to pull it.
-				if ok && !df.IsDeleted() && !df.IsSymlink() && !df.IsDirectory() {
+				// An invalid file may or may not contain block information,
```
calmh: I think invalid / ignored filed should never have block info?

lib/db/leveldb_dbinstance.go
```diff
@@ -444,11 +435,15 @@ func (db *Instance) withNeed(folder, device []byte, truncate bool, fn Iterator)
...
 		have := false // If we have the file, any version
 		need := false // If we have a lower version of the file
-		var haveVersion protocol.Vector
+		var haveFileVersion FileVersion
 		for _, v := range vl.Versions {
 			if bytes.Equal(v.Device, device) {
 				have = true
-				haveVersion = v.Version
+				haveFileVersion = v
+				if v.Invalid {
```
calmh: I'm suspicious. Why does the need decision depend only on the invalidness and not at all on the version comparison etc that happens later?

lib/model/model.go
```diff
@@ -626,6 +622,11 @@ func (m *Model) Completion(device protocol.DeviceID, folder string) FolderComple
 			return true
 		}
 
+		if file, ok := rf.Get(device, ft.Name); ok && file.Invalid {
```
calmh: This, while being the point of the exercise, is scary. It makes completion calculations at least twice as expensive as they already were. Do we understand the impact here?

imsodin: Argh this entire thing shouldn't be here, this is legacy of my first proof-of-concept. The entire point of adding invalid to the global version list was to get rid of these additional lookups by excluding invalid files in the db iteration if requested. So I will have to test again whether my entire scheme actually works.

imsodin: Hurrey, it is working without it and now it is blazing fast. I was worried why it took so long to get idle before, now I know why - thanks for catching this!

lib/db/leveldb_dbinstance.go
imsodin: The version isn't relevant for invalid files, e.g. when ignores changed we want to get all invalid files (`needinvalid == true`) while on "normal" pulls we don't consider invalid files as needed. The latter is the same behaviour as before this PR where invalid files weren't in the global version list, i.e. `have == false` which is equivalent to `need == true`.

lib/model/rwfolder.go
imsodin: Seems likely indeed. One question: The existing code that "invalidates" a FileInfo creates a new object instead of modifying the existing one:
https://github.com/imsodin/syncthing/blob/8cf6c6e94d273471fb715e25abeff152d3221b06/lib/model/model.go#L1975
The original FileInfo variable isn't used later on. Is there a side effect of changing it in place that I am missing or can such a new helper function modify the FileInfo in place like this:

```go
// Invalidate sets the Invalid property to true, discards block information and
// sets ModifiedBy to the ID of the device that invalidates it.
func (f FileInfo) Invalidate(invalidatedBy ShortID) {
	f.Invalid = true
	f.ModifiedBy = invaliatedBy
	f.Blocks = nil
}
```

lib/db/leveldb_dbinstance.go
calmh: But previously we would just slide over the invalid files, continuing along the list of available versions. With this, I guess we might have a version list of `[{version=42, invalid=true}, {version=42, invalid=false}]` and when running with `needInvalid=false` we would then immediately conclude we don't need the file, while in the old code we would see the second entry and actually evaluate the version to see if we need it or not.

lib/model/rwfolder.go
calmh: If you don't reuse it later you can modify it in place. I think in other places we do need to modify a *copy*, and a copy isn't trivial to get as there are slices and stuff internally. I think the helper function we really need is a copy function for FileInfos.

lib/db/leveldb_dbinstance.go
imsodin: We first check whether the version is from the device in question, so there is only one version to be considered. Also before this we did break out of the loop regardless of versions. Now we do find a version for this device even if it is invalid, so we need to decide whether to include (new behaviour) or ignore (old behaviour) this invalid file.

lib/model/rwfolder.go
imsodin: I didn't find a place where a copy is needed, however it still wasn't that simple as `db.FileInfoTruncated` is involved, so I had to create a separation function for that type. Hardly the most elegant version, but that's what I could come up with.

lib/db/leveldb_dbinstance.go
imsodin: While I still think this would work if invalid was equal to ignored, there is a problem with symlinks also being marked as invalid. I will have to rethink this.

imsodin:
@calmh I adjusted the approach to consider the version of invalid files. If a (locally) invalid file is update remotely, it is considered needed locally and will receive a local version bump if it is still invalid or get pulled. The new function `WithNeedOrInvalid` on 'FileSet` considers all invalid files as needed, regardless of their version. This is necessary when local ignore patterns change, as ignored and invalid files might become unignored, but have an up-to-date version.

calmh:
@imsodin do you mind solving the merge conflicts and i'll take another pass at this?

imsodin:
I will, but except for "high-level" review you rather don't invest time at the moment: https://github.com/syncthing/syncthing/pull/4460/commits/60d2a648ee394c114368152f806e8ef549276726 worked fine when testing it manually, but yesterday I unfortunately found that the current state behaves badly (takes really long, triggers lots of scanning) - I haven't yet have time to investigate why. It might just be a slipup where I messed my db up, I am doing a complete rescan (with master) for know to make sure that isn't the problem. I will let you know once I figured it out.

calmh:
OK!

imsodin:
@calmh ready!
It works: Depending on how many ignored files you have, the index exchange takes some moments but after that it is nice and calm and shows correct remote device completions :)

calmh:
Awesome, let me test and review/merge this evening!

calmh:
all: Add invalid/ignored files to global list, announce to peers (fixes #623)

This lets us determine accurate completion status for remote peers when they have ignored files.

calmh:
@st-review merge it

all: Add invalid/ignored files to global list, announce to peers (fixes #623)

This lets us determine accurate completion status for remote peers when they have ignored files.

st-review:
:ok_hand: Merged as c080f677cb0d536d6bdb1465da89653f990d0449. Thanks, @imsodin!

PR #4482, opened 2017-11-04
all: Configurable folder marker name (fixes #1126)

### LINKED ISSUE #1126: Allow syncing read-only directories as "Master Folders"
The reason for creating this is strict KitKat [permissions](http://www.androidcentral.com/kitkat-sdcard-changes) for an external SD Card. As a result Syncthing cannot sync data on SD Card except one folder /sdcard1/Android/data/com.nutomic.syncthingandroid/files/

My suggestion is not to stop syncing read-only folders if "Folder Master" option is selected.
An obvious example why we may need it is one-way syncing photos _from_ an Android device.

In this case file .stfolder also cannot be written and a "permission denied" error rises. This needs to be resolved somehow.

Of course, this would be useful not only on Android platform. Generally , this is quite logical to allow one-way sync of read-only directories when "Master Folder" option is selected.

calmh:
### Purpose

With this change there is a "marker name" folder config. When set it is used instead of .stfolder, and .stfolder is the default.

Instead of having the static function fs.IsInternal I moved that check into the ignore patterns, and made the ignore matcher handle the internal files directly. It gets the marker name when constructed and adds this to the list of internal names. We keep .stfolder as an internal name no matter what for historical reasons.

Giving this some more attention now as an escape hatch for the ".stfolder is a directory" haters... The given marker name will be created as a directory, but we don't really care what it is as long as a stat returns that it exists.

Handling this in combination with whatever we intended to store in the marker in the future is an exercise for future us...

### Testing

One unit tests and a quick manual check...

### Screenshots

Advanced config only:

screen shot 2017-11-04 at 14 00 20


lib/scanner/walk.go
```diff
@@ -242,12 +244,7 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
 		}
 
-		if fs.IsInternal(path) {
-			l.Debugln("ignored (internal):", path)
-			return skip
-		}
-
-		if w.Matcher.Match(path).IsIgnored() {
+		if w.Matcher.ShouldIgnore(path) {
 			l.Debugln("ignored (patterns):", path)
```
imsodin: " (patterns)" is not meaningful anymore.

lib/ignore/ignore.go
```diff
@@ -281,6 +289,22 @@ func (m *Matcher) ShouldIgnore(filename string) bool {
 	return false
 }
 
+// isInternal returns true if the file, as a path relative to the folder
+// root, represents an internal file that should always be ignored. The file
+// path must be clean (i.e., in canonical shortest form).
+func (m *Matcher) isInternal(file string) bool {
```
imsodin: > This relocation will cause me issues with the encrypted folder stuff.

To solve this, couldn't this code be moved into `Match` and `Result` be extended with a member function `IsInternal()` and accordingly an additional state `resultInternal`? As far as I understand your approach was to "interface" `Match` into the filesystem to first decrypt the filename, which would then still work.

calmh:
Actually this thing is fundamentally wrong. We should not ignore the marker. What we should do:

- Ignore `.stfolder`, perhaps as we already did with the `IsInternal` call
- Have the marker be configurable and default to `.stfolder`
- If it's set to something that is not `.stfolder` we:
  1. Do not attempt to create it, as it's something the user manages themselves
  2. Do not ignore it, as it's probably one of the users actual data items

calmh:
Reopen with new approach

calmh:
OK apparently it can't be reopened once force pushed... New one then.

PR #4483, opened 2017-11-05
lib/config, lib/model: Configurable folder marker name (fixes #1126)

### LINKED ISSUE #1126: Allow syncing read-only directories as "Master Folders"

calmh:
### Purpose

Make the folder marker name configurable. When it is set to something custom we do not attempt to create, hide, delete or ignore it. It can still be ignored in the usual way if the user so wishes.

### Testing

Unit test that the health check still passes.

### Screenshots

Advanced config only:

32405590-980b0058-c168-11e7-960a-a63e4c316ba8

### DIFF
--- a/lib/config/config.go
+++ b/lib/config/config.go
@@ -399,17 +399,21 @@ func convertV22V23(cfg *Configuration) {
 		// begin with.
 		permBits = 0700
 	}
+
+	// Upgrade code remains hardcoded for .stfolder despite configurable
+	// marker name in later versions.
+
 	for i := range cfg.Folders {
 		fs := cfg.Folders[i].Filesystem()
 		// Invalid config posted, or tests.
 		if fs == nil {
 			continue
 		}
-		if stat, err := fs.Stat(".stfolder"); err == nil && !stat.IsDir() {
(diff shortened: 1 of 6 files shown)


calmh:
Also https://github.com/syncthing/docs/pull/353

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 9c855ab22e9f0e895db44fd63e0460c0c9451f6f. Thanks, @calmh!

PR #4486, opened 2017-11-06
lib/connections:  Indicate stack in transport (fixes #4463)

### LINKED ISSUE #4463: Split transport usage reporting per stack
So we'd know our v6 reach, versus v4 reach.

AudriusButkevicius:
As the code explains.

### DIFF
--- a/lib/connections/structs.go
+++ b/lib/connections/structs.go
@@ -93,7 +93,15 @@ func (c internalConn) Type() string {
 }
 
 func (c internalConn) Transport() string {
-	return c.connType.Transport()
+	transport := c.connType.Transport()
+	ip := net.ParseIP(c.LocalAddr().String())
+	if ip == nil {
+		return transport
+	}
+	if ip.To4() != nil {
+		return transport + "4"
+	}
+	return transport + "6"
 }
 
 func (c internalConn) String() string {
--- a/lib/connections/structs.go
+++ b/lib/connections/structs.go
@@ -94,7 +94,11 @@ func (c internalConn) Type() string {
 
 func (c internalConn) Transport() string {
 	transport := c.connType.Transport()
-	ip := net.ParseIP(c.LocalAddr().String())
+	host, _, err := net.SplitHostPort(c.LocalAddr().String())
+	if err != nil {
+		return transport
+	}
+	ip := net.ParseIP(host)
 	if ip == nil {
 		return transport
 	}


lib/connections/structs.go
```diff
@@ -93,7 +93,15 @@ func (c internalConn) Type() string {
 }
 
 func (c internalConn) Transport() string {
-	return c.connType.Transport()
+	transport := c.connType.Transport()
+	ip := net.ParseIP(c.LocalAddr().String())
```
calmh: I don't think this works, as `LocalAddr().String()` should include a port number. (If you think it does, a test to prove it would be sweet.)

AudriusButkevicius: As usual, this is Well Tested(TM)

calmh: Apart from, like, not working, it looks good to me.

AudriusButkevicius:
internalConn needs a *tls.Conn which is annoying to mock, so no tests. Worst case this won't work and have no effect :)

AudriusButkevicius:
@imsodin your test failures.

imsodin:
As it won't be reproducible, there is nothing to investigate here. Should I simply disable this test on mac or can we live with having to rerun this mac test once every 20 builds (teamcity currently reports 5% flakyness)?

calmh:
@st-review merge

st-review:
:ok_hand: Merged as 0ee1146e1c996e9056c9ae0af71a6cec4cf9388b. Thanks, @AudriusButkevicius!

calmh:
@imsodin is there no way to unflake it?

imsodin:
@calmh Probably, but I have no clue how. The test timed out after 10min, while it has an "internal" timeout of 2s - there should be now way at all for it to hang up (and it only does such things on mac, but there is nothing mac specific except for extra delays in these tests).

AudriusButkevicius:
When it times out, it usually dumps the stacks atleast on my box. Perhaps teamcity swallows that (or I have it setup that way specifically)

imsodin:
Right, didn't think about that. I even vaguely remember that I have seen such a stacktrace on teamcity as well, but I am not sure. This time there is nothing however (or I'm blind):
https://build.syncthing.net/viewLog.html?buildId=12326&buildTypeId=Syncthing_BuildMac&tab=buildLog&_focus=976#_state=976

PR #4487, opened 2017-11-06
lib/connections: Fix lan detection (fixes #4421)

### LINKED ISSUE #4421: Local network classification doesn't always work
Firstly, we check these once on startup, which is not ideal.
Secondly, they are only checked if rate limiting is enabled, yet if rate limiting is enabled post startup, lans are never resolved.

AudriusButkevicius:
...

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -725,24 +725,6 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 		weakhash.Enabled = true
 	}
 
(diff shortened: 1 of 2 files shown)


calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 62a4106a794bd88c6f0a86593ee0c4f4736a4d72. Thanks, @AudriusButkevicius!

calmh:
Are you asking me or setting reminders for yourself? :D

screen shot 2017-11-06 at 15 06 54

AudriusButkevicius:
Nah, it was supposed to be a branch reducing strain on discovery servers by not reannouncing many kcp addresses, but then I got bored and did something else.

PR #4489, opened 2017-11-06
lib/connections: Parallel dials in the same priority (fixes #4456)

### LINKED ISSUE #4456: Dial device addresses in parallel
<a href="https://github.com/syncthing/syncthing/pull/4452#issuecomment-338512994">screen shot 2017-10-23 at 08 11 48</a>

If we have multiple tcp/kcp addresses available for a device we could dial them all in parallel, picking the best that succeeded after a second or so.

(This will need better handling of multiple device connections on the receiving side, as we otherwise run the risk of having the best connection being dropped as "already connected to device" from the other side.)

AudriusButkevicius:
Well Tested(TM)

Introduces a potential issue where we always pick some connectable but dodgy connection that breaks
soon after the TLS handshake.


lib/connections/service.go
```diff
@@ -391,23 +395,62 @@ func (s *Service) connect() {
...
+				res := make(chan internalConn, len(funcs))
+				for _, dialFunc := range funcs {
+					go func(df func() (internalConn, error), result chan internalConn) {
+						if c, err := df(); err == nil {
+							result <- c
+						}
+						wg.Done()
+					}(dialFunc, res)
+				}
+				wg.Wait()
```
calmh: It seems wasteful to wait for all connections to fail before proceeding with one that succeeded?

AudriusButkevicius: It would be awesome if we'd have context support in the dialer package, yet we don't :(

calmh:
Yep that's all good then

AudriusButkevicius:
I'm rewriting this to make it more readable, and to address the speed issue you've pointed out.

AudriusButkevicius:
Should be more readable.

lib/connections/service.go
```diff
@@ -693,3 +698,45 @@ func IsAllowedNetwork(host string, allowed []string) bool {
...
+		dialTargetBuckets[tgt.priority] = append(dialTargetBuckets[tgt.priority], tgt)
+	}
+
+	// Get all available priorities
+	priorities := make([]int, 0, len(dialTargetBuckets))
+	for prio := range dialTargetBuckets {
+		priorities = append(priorities, prio)
+	}
+
+	// Sort the priorities so that we dial form lowest (which means highest...)
```
imsodin: Good to know :)
Also s/form/from/ or maybe "dial lowest first" would be clearer anyway.

lib/connections/service.go
```diff
@@ -693,3 +698,45 @@ func IsAllowedNetwork(host string, allowed []string) bool {
...
+		go func(deviceID protocol.DeviceID, prio int) {
+			wg.Wait()
+			close(res)
+			l.Debugln("discarding", len(res), "connections while connecting to", deviceID, prio)
+			for conn := range res {
+				conn.Close()
+			}
+		}(deviceID, prio)
+
+		if conn, ok := <-res; ok {
```
imsodin: Isn't this a race with range above in case there is only once dialer for this prio (or all except the last one fail)? As the channel is buffered, the send in the dialer will return immediately, the wait group condition is fulfilled and the channel closed. While unlikely, this could happen before this if condition receives a valid connection. Or is there something preventing this that I miss?

AudriusButkevicius: Yes this can happen. This should be handled differently.

AudriusButkevicius:
Fixed.

lib/connections/structs.go
```diff
@@ -165,3 +166,19 @@ func (o *onAddressesChangedNotifier) notifyAddressesChanged(l genericListener) {
...
+	uri      *url.URL
+	priority int
+}
+
+func (t dialTarget) Dial(deviceID protocol.DeviceID, wg sync.WaitGroup, res chan<- internalConn) {
+	l.Debugln("dialing", deviceID, t.uri, "prio", t.priority)
+	conn, err := t.dialer.Dial(deviceID, t.uri)
+	l.Debugln("dialing", deviceID, t.uri, "outcome", conn, err)
+	if err == nil {
+		res <- conn
```
calmh: How about, instead of having `res` be `len(targets)`-buffered, it is just 1-buffered and we do

```
select {
  case res <- conn:
    // we won
  default:
    // we lost
    conn.close()
}
```

then I think you don't need the discarder routine above?

lib/connections/structs.go
```diff
@@ -165,3 +166,19 @@ func (o *onAddressesChangedNotifier) notifyAddressesChanged(l genericListener) {
 		callback(l)
 	}
 }
+
+type dialTarget struct {
+	dialer   genericDialer
+	uri      *url.URL
+	priority int
+}
```
calmh: It seems to me that the rationale between what goes into the struct and what goes as parameters is odd - the stuff that is the same for every invocation is in the parameters, and the stuff that varies is in the struct. I realize that this is because we're collecting the stuff that varies into the struct in a loop, and the rest we just "know" and throw in as parameters, but it feels ugly...

It would be nice if the `uri` was the parameter and the rest is the context the dialer works on, making it more liek a normal dialer... I don't have a concrete suggestion on sorting that out though so take it as a suggest/comment and do with it what you like.

I also think it's not worth passing the wait group just for a single done call, it could be done less magically by the caller as `go func() { thedial ; wg.Done() }`

lib/connections/service.go
```diff
@@ -693,3 +698,59 @@ func IsAllowedNetwork(host string, allowed []string) bool {
...
+			wg.Add(1)
+			go tgt.Dial(deviceID, wg, res)
+		}
+
+		// Spawn a routine which will unblock main routine in case we fail
+		// to connect to anyone.
+		go func() {
+			wg.Wait()
+			close(res)
+
```
calmh: Blank line on the wrong side of `}()`

lib/connections/structs.go
AudriusButkevicius: Not really. It will panic on send to closed channel, and second dial can slip through to the channel which still has to be discarded.

lib/connections/structs.go
AudriusButkevicius: How about I wedge everything in the struct, and do the dial in an anonymous function?

AudriusButkevicius:
So I've replied to your comments, but kept the implementation pretty much the same.

calmh:
Fine. Is it tested somehow?

AudriusButkevicius:
It's Well Tested(TM), which means I didn't even try compiling it on my own machine. Sadly I am on holiday with lack of monitors, so launching multiple instances in an single 13" screen is too hard.

calmh:
That's what I suspected. Let me take it for a run some time then.

AudriusButkevicius:
Zing zing.

imsodin:
Meager data point: My laptop with this PR does not blow up and connects fine to other devices on LAN.

lib/connections/service.go
```diff
@@ -710,3 +716,65 @@ func IsAllowedNetwork(host string, allowed []string) bool {
...
+		tgts := dialTargetBuckets[prio]
+		res := make(chan internalConn, len(tgts))
+		wg := sync.NewWaitGroup()
+		for _, tgt := range tgts {
+			wg.Add(1)
+			go func() {
+				conn, err := tgt.Dial()
+				if err == nil {
+					res <- conn
+				}
```
imsodin: Dial failure was previously logged to debug. Is this removed on purpose?

AudriusButkevicius: It's logged in Dial()

calmh:
@st-review merge

st-review:
:ok_hand: Merged as aecd7c64ced9bff41c0def7b669d22385d53d368. Thanks, @AudriusButkevicius!

PR #4491, opened 2017-11-06
cmd/syncthing: No global options (fixes #4474)

### LINKED ISSUE #4474: Spurious "allow anonymous usage reporting?" dialog on RC
It should not ask about the new format on RC:s, just use it.

screen shot 2017-10-31 at 09 55 28

```
$ grep ur .config/syncthing/config.xml
<configuration version="25">
        <urAccepted>2</urAccepted>
        <urSeen>2</urSeen>
        <urUniqueID>kLP6zS6x</urUniqueID>
        <urURL>https://data.syncthing.net/newdata</urURL>
        <urPostInsecurely>false</urPostInsecurely>
        <urInitialDelayS>1800</urInitialDelayS>
</configuration>
```

AudriusButkevicius:
Currently it's a bit wishy washy, some use the options in the local scope, some fetch them on the fly,
this does a bunch of copying every time.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -702,9 +702,7 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 		},
 	}
 
-	opts := cfg.Options()
-
-	if opts.WeakHashSelectionMethod == config.WeakHashAuto {
+	if opts := cfg.Options(); opts.WeakHashSelectionMethod == config.WeakHashAuto {
 		perfWithWeakHash := cpuBench(3, 150*time.Millisecond, true)
 		l.Infof("Hashing performance with weak hash is %.02f MB/s", perfWithWeakHash)
 		perfWithoutWeakHash := cpuBench(3, 150*time.Millisecond, false)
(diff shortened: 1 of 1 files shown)


imsodin:
Why do you get rid of the "global" opts? Wouldn't the two additional lines to set and save options be enough: https://github.com/syncthing/syncthing/pull/4491/files#diff-90be85495f283f39a39308cf8ffb765cR860

AudriusButkevicius:
I feel that rechecking the options every time we need them is probably the more correct thing todo.
What if one of these things that we call into saves options itself, and then we bulldoze whatever changes the subroutine has made?

imsodin:
@st-review lgtm

st-review:

calmh:
I see no global options here, just a function local variable used to reduce some typing? Rechecking is a bit moot as this function is only ever run once and lasts for the length of the program...

calmh:
Oh I see, it fixes the UR thing. Yeah, OK.

calmh:
@st-review lgtm

cmd/syncthing: Don't cache stale options in main (fixes #4474)

st-review:
:ok_hand: Merged as 88180904f24a1274abbd9204082b2742d7f4bcf0. Thanks, @AudriusButkevicius!

Issue #4476, opened 2017-10-31
Dir not empty can return a better error/rescan


AudriusButkevicius:
Given we know there are files that are not deletable and ignored, we could return a more descriptive error.
Also of we discover stuff we don't have in the index, we could force rescan.

generalmanager:
I just created a similar ticket for the case where local files can't be read or are ignored because they are symlinks (https://github.com/syncthing/syncthing/issues/4480). We should probably use the same kind of UI signaling in both cases. Also :+1:

AudriusButkevicius:
This has nothing todo with UI.

generalmanager:
What I meant is that the kind of UI I described in #4480 could also be appropriate for the case where ignored files/subdirectories or ACL prevent the deletion of a directory.
I didn't grasp all the details of #4475 and this tickets connection to it before reading it in it's entirety, so I get if you want to keep the scope of this issue focused on the error message and triggering a rescan. 

I just wanted to leave a note to a possible UI improvement in case #4480 is accepted, so that this and similar cases where local stuff can't be deleted would be treated similarly to local stuff not beeing readable. Apoligies if you felt that I sidetracked from the topic.

AudriusButkevicius:
It seems you are confused. This ticket doesn't need a UI nor does #4475.

generalmanager:
I understand that neither an improved error message nor triggering a rescan or bumping a version on the db needs any kind of UI. I just wanted to leave a note saying that *IF* it should be decided to build some kind of user-facing message in regard to unaccessible files (which the issue I opened is about) then it could also make sense to use the same UI to inform the user about beeing unable to delete local folders (with ignored stuff inside it beeing one possible reason).

I agree that not beeing able to delete something isn't necessarily of utmost importance to the user, this was really only meant as a small annotation, mainly to get the tickets linked so when this is built, one takes a second to think if the UI which might be built for #4480 would also be useful for this, preventing additional work in the future. But now with all the back-and-forth it's really starting to distract from the main issue. So please feel free to delete my comments here.

calmh:
Closed by #4493

calmh:
@imsodin what was the resolution in relation to this issue, I'm not sure I understand the action here

AudriusButkevicius:
It returns more readable errors and schedules scans from within delete if there are unscanned entries.

imsodin:
[...]

PR #4493, opened 2017-11-07
lib/model: Improve scan scheduling and dir del during pull (fixes #4475 #4476)

### LINKED ISSUE #4475: Discovering new files in a deleted directory does not resurrect the directory
So for some unknown reason the puller gets stuck on some directory being unable to delete it. The directory was deleted on the remote device (don't know when exactly, maybe yesterday) but can't be removed on the local one. Here's the log:

```
[xxxx] 2017/10/31 13:10:51.599131 logfs.go:61: DEBUG: mtimefs.go:55 basic /home/dina/Dropbox/ Lstat . {0xc4203e21a0} <nil>
[xxxx] 2017/10/31 13:10:51.599188 logfs.go:61: DEBUG: mtimefs.go:55 basic /home/dina/Dropbox/ Lstat моё {0xc4203e2270} <nil>
[xxxx] 2017/10/31 13:10:51.599230 rwfolder.go:570: DEBUG: Deleting dir моё/МедСправки орто
[xxxx] 2017/10/31 13:10:51.599326 logfs.go:55: DEBUG: <autogenerated>:1 basic /home/dina/Dropbox/ DirNames моё/МедСправки орто [Справка.bmp Выписка1.bmp Thumbs.db Выписка2.bmp Акт.bmp] <nil>
[xxxx] 2017/10/31 13:10:51.599353 logfs.go:115: DEBUG: <autogenerated>:1 basic /home/dina/Dropbox/ Stat моё {0xc4203e2340} <nil>
[xxxx] 2017/10/31 13:10:51.599385 logfs.go:97: DEBUG: <autogenerated>:1 basic /home/dina/Dropbox/ Remove моё/МедСправки орто remove /home/dina/Dropbox/моё/МедСправки орто: directory not empty
[xxxx] 2017/10/31 13:10:51.599399 logfs.go:61: DEBUG: mtimefs.go:55 basic /home/dina/Dropbox/ Lstat моё/МедСправки орто {0xc4203e2410} <nil>
[xxxx] 2017/10/31 13:10:51.599426 rwfolder.go:1675: INFO: Puller (folder "dina-dropbox", file "моё/МедСправки орто"): delete dir: remove /home/dina/Dropbox/моё/МедСправки орто: directory not empty
[xxxx] 2017/10/31 13:10:51.599454 rwfolder.go:227: DEBUG: sendReceiveFolder/dina-dropbox@0xc420087400 changed 1
[xxxx] 2017/10/31 13:10:51.599466 rwfolder.go:325: DEBUG: sendReceiveFolder/dina-dropbox@0xc420087400 c 1 p 16
[xxxx] 2017/10/31 13:10:51.629165 logfs.go:61: DEBUG: mtimefs.go:55 basic /home/dina/Dropbox/ Lstat . {0xc42148a0d0} <nil>
[xxxx] 2017/10/31 13:10:51.629228 logfs.go:61: DEBUG: mtimefs.go:55 basic /home/dina/Dropbox/ Lstat моё {0xc42148a4e0} <nil>
[xxxx] 2017/10/31 13:10:51.629289 rwfolder.go:570: DEBUG: Deleting dir моё/МедСправки орто
[xxxx] 2017/10/31 13:10:51.629371 logfs.go:55: DEBUG: <autogenerated>:1 basic /home/dina/Dropbox/ DirNames моё/МедСправки орто [Справка.bmp Выписка1.bmp Thumbs.db Выписка2.bmp Акт.bmp] <nil>
[xxxx] 2017/10/31 13:10:51.629406 logfs.go:115: DEBUG: <autogenerated>:1 basic /home/dina/Dropbox/ Stat моё {0xc42148a5b0} <nil>
[xxxx] 2017/10/31 13:10:51.629437 logfs.go:97: DEBUG: <autogenerated>:1 basic /home/dina/Dropbox/ Remove моё/МедСправки орто remove /home/dina/Dropbox/моё/МедСправки орто: directory not empty
[xxxx] 2017/10/31 13:10:51.629458 logfs.go:61: DEBUG: mtimefs.go:55 basic /home/dina/Dropbox/ Lstat моё/МедСправки орто {0xc42148a680} <nil>
[xxxx] 2017/10/31 13:10:51.629503 rwfolder.go:227: DEBUG: sendReceiveFolder/dina-dropbox@0xc420087400 changed 1
[xxxx] 2017/10/31 13:10:51.629518 rwfolder.go:325: DEBUG: sendReceiveFolder/dina-dropbox@0xc420087400 c 1 p 16
[xxxx] 2017/10/31 13:10:51.656975 logfs.go:61: DEBUG: mtimefs.go:55 basic /home/dina/Dropbox/ Lstat . {0xc42148a750} <nil>
[xxxx] 2017/10/31 13:10:51.657036 logfs.go:61: DEBUG: mtimefs.go:55 basic /home/dina/Dropbox/ Lstat моё {0xc42148a820} <nil>
[xxxx] 2017/10/31 13:10:51.657110 rwfolder.go:570: DEBUG: Deleting dir моё/МедСправки орто
[xxxx] 2017/10/31 13:10:51.657216 logfs.go:55: DEBUG: <autogenerated>:1 basic /home/dina/Dropbox/ DirNames моё/МедСправки орто [Справка.bmp Выписка1.bmp Thumbs.db Выписка2.bmp Акт.bmp] <nil>
[xxxx] 2017/10/31 13:10:51.657265 logfs.go:115: DEBUG: <autogenerated>:1 basic /home/dina/Dropbox/ Stat моё {0xc420cbe000} <nil>
[xxxx] 2017/10/31 13:10:51.657314 logfs.go:97: DEBUG: <autogenerated>:1 basic /home/dina/Dropbox/ Remove моё/МедСправки орто remove /home/dina/Dropbox/моё/МедСправки орто: directory not empty
[xxxx] 2017/10/31 13:10:51.657330 logfs.go:61: DEBUG: mtimefs.go:55 basic /home/dina/Dropbox/ Lstat моё/МедСправки орто {0xc420cbe0d0} <nil>
[xxxx] 2017/10/31 13:10:51.657376 rwfolder.go:227: DEBUG: sendReceiveFolder/dina-dropbox@0xc420087400 changed 1
[xxxx] 2017/10/31 13:10:51.657399 rwfolder.go:264: INFO: Folder dina-dropbox isn't making progress. Pausing puller for 1m0s.
[xxxx] 2017/10/31 13:10:51.657412 rwfolder.go:265: DEBUG: sendReceiveFolder/dina-dropbox@0xc420087400 next pull in 1m0s
[xxxx] 2017/10/31 13:10:51.687779 progressemitter.go:240: DEBUG: progress emitter: bytes completed for dina-dropbox: 0
```

The sync progress is stuck seemingly forever. Of course I can remove the directory, rescan and it'll work again for some days or weeks only to get stuck on another directory again. This happens regularly but I see no pattern as that happens to another Syncthing user, I only provide the server. The directories and files are renamed and deleted pretty often, not always there's a connection to the server, sometimes Syncthing disconnects and never reconnects again though Internet is working fine (subject to other issue). It's pretty long-standing now, I remember this happening even a year ago.

### Version Information

Syncthing Version: v0.14.40-rc.4 (server), 0.14.39 (client)
OS Version: Debian Jessie (server), Ubuntu 14.04 (client)
Browser Version: Chrome latest

imsodin:
This does not prevent the inconsistent state discussed in #4475, as it would require lots of db lookups for a relatively rare situation. It instead detects this situation when a directory should be deleted due to pulling/remote change, but there exists known valid files inside this directory. In this case the version of the local dir is bumped such that overrides the deletion from the other device.

This change broke one test and I added one more to target this specific scenario. This lead to feature creep in connected code:
 - The broken test `TestIssue3164` does nothing in the model, it operates solely on the sendreceive folder, so I fixed and moved it in rwfolder_test.go.
 - s/f.dbUpdates/dbUpdateChan/ because all other channel involved in pulling are not part of the folder struct and are anyways redefined on every puller iteration.
 - Don't call `ScanFolder` in goroutines on every puller iteration, add paths to `toBeScanned` and "schedule" the scan at the end of pulling (actual scan will anyways only happen after pulling finished). Saves some redundant scans and seems less hackish to me.

Testing: One new test simulating #4475, which in addition also tests the case of an unexpected, unscanned file in a to be deleted folder. No "testing in production" so far, I will do that when I examine the invalid PR situation.


lib/model/model_test.go
```diff
@@ -2502,6 +2469,78 @@ func TestCustomMarkerName(t *testing.T) {
...
+		t.Fatal(err)
+		return
+	}
+	fd.Close()
+
+	dbi := db.OpenMemory()
+	m := NewModel(defaultConfig, protocol.LocalDeviceID, "syncthing", "dev", dbi, nil)
+	m.AddFolder(defaultFolderConfig)
+	m.StartFolder("default")
+	m.ServeBackground()
```
AudriusButkevicius: Doesn't this (or folder start) kick off an initial scan?

lib/model/model_test.go
```diff
@@ -2502,6 +2469,78 @@ func TestCustomMarkerName(t *testing.T) {
...
+	}
+	fd.Close()
+
+	dir, ok := m.CurrentFolderFile("default", "baz")
+	if !ok {
+		t.Fatalf("Can't get dir \"baz\" after initial scan")
+	}
+
+	dir.Deleted = true
+	version := dir.Version.Update(device1.Short()).Update(device1.Short())
```
AudriusButkevicius: Why twice?

lib/model/rwfolder.go
```diff
@@ -1378,31 +1425,14 @@ func (f *sendReceiveFolder) performFinish(state *sharedPullerState) error {
...
-			// non-empty directories).
-
-			// TODO: This is the place where we want to remove temporary files
-			// and future hard ignores before attempting a directory delete.
-			// Should share code with f.deletDir().
-
-			if err = osutil.InWritableDir(f.fs.Remove, f.fs, state.file.Name); err != nil {
-				return err
-			}
+			panic("Symlinks and directories should be handled directly")
```
AudriusButkevicius: This can cause a panic if I replace a file with a directory as it's being downloaded.
I think this should be a conflict, or do what it does now.

lib/model/rwfolder.go
```diff
@@ -760,33 +770,70 @@ func (f *sendReceiveFolder) deleteDir(file protocol.FileInfo, matcher *ignore.Ma
...
 		}
 	}
 
+	if resurrect {
+		l.Debugln(f, "resurrecting dir", file.Name)
+		// A new file has been added before the dir was deleted -> revive the dir in index
+		file.Version = file.Version.Update(f.model.shortID)
+		dbUpdateChan <- dbUpdateJob{file, dbUpdateResurrectDir}
+		return
+	}
```
AudriusButkevicius: Isn't it easier on each dbHandle/ShortcutFile to traverse up each directory up to the root and bump the version?

The rest of checks and errors etc look good.

lib/model/rwfolder.go
imsodin: I misread this, will have a second look at it. the todo comment is confusing

imsodin: I misread this, will have a second look at it. the todo comment is confusing

lib/model/model_test.go
imsodin: yes. the scanfolder call below is used to make sure it has finished before continuing

lib/model/model_test.go
imsodin: because the original version does not have device1 in it at all, so nothing->1->2 higher than 1 on local device

lib/model/rwfolder.go
imsodin: maybe easier but very expensive: on every file change you need as many additional database lookups as there is path components.

AudriusButkevicius: As, isn't it easier to handle this at the scanner.

AudriusButkevicius: Yeah, but technically that's the right thing to do to prevent races. This seems like a hack.
Furthermore, if we go for a graph based puller, doing this makes perfect sense?

imsodin: In my opinion this is a pretty clean hack and preferable to the performance hit of updating all parent folders on every detected change via pull or scan, but that's obviously a judgement call.
As for the graph based puller: I don't know what it is and how it relates to this, but I hear about it every now and then whenever something complex comes up - this sounds like one of these future wonders to be completed yesterday things. More to the point: I don't think it is a good idea to postpone something because of it.

AudriusButkevicius: I'll leave for @calmh to make the judgement call. This just feels like technical debt that potentially will never be repaid.

imsodin: While fixing https://github.com/syncthing/syncthing/pull/4493#discussion_r149401957 I thought of another very unlikely but possible corner case which would make this entire thing even more complex/ugly. I am getting increasingly convinced that this is not the way to go. You might want to hold off reviewing/judgment calling this PR for now, I'll do some more digging/thinking into the issue itself.

imsodin:
I found the source of the problem. This now prevents the problem on the "device of origin" by not only silently creating missing parent dirs, but also scheduling a scan on them. This needed a slight adjustment to `osutil.TraversesSymlinks` and as an additional feature creep I followed the TODO comment in `performFinish`.

lib/osutil/traversessymlink.go
```diff
@@ -32,47 +32,46 @@ func (e NotADirectoryError) Error() string {
...
 }
 
-// TraversesSymlink returns an error if base and any path component of name up to and
-// including filepath.Join(base, name) traverses a symlink.
+// TraversesSymlink returns an error and the part of path causing it, if base
+// and any path component of name up to and including name traverses a symlink,
+// is not a directory or is missing.
 // Base and name must both be clean and name must be relative to base.
-func TraversesSymlink(filesystem fs.Filesystem, name string) error {
+func TraversesSymlink(filesystem fs.Filesystem, name string) (error, string) {
```
calmh: I haven't read the rest yet, just a small cleanliness note here. The `(error, string)` return type and `err, _ := ...` calls look *really* odd to me. I would prefer extending the error with the path information. That is, add a `Path string` to `NotADirectoryError` and `TraversesSymlinkError` that the caller can use after type assertion if interested. Or maybe reuse `os.PathError` which is essentially this...

imsodin:
I now also tested it manually and the scenario of 4475 isn't reproducible anymore, as the deleted dir is now detected in a scan immediately after a file was created in it from pulling.

lib/model/rwfolder.go
```diff
@@ -1723,6 +1727,66 @@ func (f *sendReceiveFolder) IgnoresUpdated() {
...
+		return errDirHasToBeScanned
+	}
+	if hasIgnored {
+		return errDirHasIgnored
+	}
+	if hasKnown {
+		return errDirHasKnown
+	}
+
+	for _, del := range toBeDeleted {
```
calmh: I think we could do this part before returning, to make some progress regardless?

lib/model/rwfolder.go
```diff
@@ -1397,15 +1412,8 @@ func (f *sendReceiveFolder) performFinish(state *sharedPullerState) error {
...
-			l.Debugln("file modified but not rescanned; not finishing:", state.curFile.Name)
-			// Scan() is synchronous (i.e. blocks until the scan is
-			// completed and returns an error), but a scan can't happen
-			// while we're in the puller routine. Request the scan in the
-			// background and it'll be handled when the current pulling
-			// sweep is complete. As we do retries, we'll queue the scan
-			// for this file up to ten times, but the last nine of those
-			// scans will be cheap...
-			go f.Scan([]string{state.curFile.Name})
+			l.Debugln("file unknow or modified but not scanned; not finishing:", state.curFile.Name)
```
calmh: typo in "unknown". It's in a debug message so I would hardly care except someone else is going to find it during some automated scan and file a separate pull request on it which I'll then have to handle

lib/model/rwfolder.go
```diff
@@ -262,11 +264,12 @@ func (f *sendReceiveFolder) pull(prevSeq int64, prevIgnoreHash string) (curSeq i
 	f.clearErrors()
 	var changed int
 	tries := 0
+	toBeScanned := make(map[string]struct{})
```
calmh: This map appears to be passed around everywhere during the pulling. Given we have multiple pullers this should be a data race, probably with a crash when the runtime detects simultaneous modification. I think we just don't tickle this in the tests...

I suggest a similar mechanism to the db updater thing. A channel for items to be scanned (that can be passed around willy nilly), a loop that collects items from it, channel is closed when we're done, the loop exits and does the scan.

lib/model/rwfolder.go
imsodin: Maybe in the case where there is only something ignored in the dir that cannot be deleted, but in case there is something to be scanned or valid in there, I don't think we should delete anything, as we might not delete the dir in the end.

lib/model/rwfolder.go
```diff
@@ -295,14 +298,7 @@ func (f *sendReceiveFolder) pull(prevIgnoreHash string) (curIgnoreHash string, s
...
 
-	if len(toBeScanned) != 0 {
-		scanList := make([]string, len(toBeScanned))
-		for path := range toBeScanned {
-			l.Debugln(f, "scheduling scan after pulling for", path)
-			scanList = append(scanList, path)
-		}
-		go f.Scan(scanList)
-	}
+	close(scanChan)
```
AudriusButkevicius: This probably should be after the FolderIdle as you might bulldoze FolderScan state.

lib/model/rwfolder.go
```diff
@@ -1611,6 +1614,32 @@ loop:
...
+loop:
+	for {
+		select {
+		case path, ok := <-scanChan:
+			if !ok {
+				break loop
+			}
+			toBeScanned[path] = struct{}{}
+		}
+	}
```
AudriusButkevicius: just:

``
for path := <-scanChan {
  toBeScanned[path[ = struct{}{}
}
```

File: lib/model/rwfolder.go (hunk 9 above)
Comment (imsodin) [reply lib/model/rwfolder.go]: It doesn't matter as the actual scan will happen via the main folder for loop, i.e. after this function finished. I'll change it to prevent that confusion regardless.

**Comment by imsodin** [comment]
What about this?

File: lib/model/rwfolder.go [hunk 11]
```diff
@@ -1618,6 +1618,25 @@ loop:
...
+		toBeScanned[path] = struct{}{}
+	}
+
+	if len(toBeScanned) != 0 {
+		scanList := make([]string, len(toBeScanned))
+		for path := range toBeScanned {
+			l.Debugln(f, "scheduling scan after pulling for", path)
+			scanList = append(scanList, path)
+		}
+		f.Scan(scanList)
```
AudriusButkevicius: This will be `toBeScanned` x empty strings + toBeScanned content, as make uses size and not capacity hint.

lib/model/rwfolder.go
```diff
@@ -60,6 +60,9 @@ var (
 	activity               = newDeviceActivity()
 	errNoDevice            = errors.New("peers who had this file went away, or the file has changed while syncing. will retry later")
 	errSymlinksUnsupported = errors.New("symlinks not supported")
+	errDirHasToBeScanned   = errors.New("directory contains unexpected files, scheduling scan")
+	errDirHasIgnored       = errors.New("directory contains ignored files (see ignore documentation for (?d) prefix)")
+	errDirHasKnown         = errors.New("directory contains valid files, if it is not fixed in a subsequent pull check remote")
```
AudriusButkevicius: What is this intending to say?

lib/model/rwfolder.go
```diff
@@ -1731,6 +1750,66 @@ func (f *sendReceiveFolder) IgnoresUpdated() {
 	f.IndexUpdated()
 }
 
+// deleteDir attempts to delete a directory. It checks for files/dirs inside
+// the directory and removes them if possible or returns an error if it fails
+func (f *sendReceiveFolder) deleteDir(dir string, ignores *ignore.Matcher, scanChan chan<- string) error {
+	files, _ := f.fs.DirNames(dir)
+
+	toBeDeleted := make([]string, 0)
```
AudriusButkevicius: You are not going to delete more than there are files, and there ideally should not be any files, so perhaps it's worth using len(files) as a hint?

AudriusButkevicius:
I don't remember what this looked like, but I can't see any hacks here, apart from the comments.

lib/model/rwfolder.go
```diff
@@ -524,9 +533,26 @@ nextFile:
...
+			// fail to create the file and end up looping forever on it. This
+			// breaks that by creating the directory and scheduling a scan,
+			// where it will be found and the delete bit on it removed.  The
+			// user can then clean up as they like...
+			l.Debugln("%v resurrecting parent directory of %v: %v", f, fi.Name, tErr.Path)
+			if err := f.fs.MkdirAll(filepath.Dir(fi.Name), 0755); err != nil {
+				f.newError("resurrecting parent dir", fi.Name, err)
+				continue
+			}
+			scanChan <- tErr.Path
```
AudriusButkevicius: Why this whole lot is under TraversesSymlink if?

lib/model/rwfolder.go
imsodin: This is a failed attempt to compactify the information that there are files inside the folder, which according to the db and ignores should really be there, but their parent dir is still to be deleted and this might be transient (if it isn't, it definitely is a bug). Probably this should just be a generic "directory contains files error".

lib/model/rwfolder.go
imsodin: Because if it returns a nil error there is nothing to do, if it reports a missing parent directory, we recreate and scan it.

lib/model/rwfolder.go
imsodin: This is fixed -thanks.

lib/model/rwfolder.go
AudriusButkevicius: It's a bit strange to rely on traverses symlink implementation to validate that the whole path is there. What if it checks if there is no symlinks up to the point in the path where something actually exists?

imsodin: I see that the naming is a bit confusing, I don't understand the question though: That is exactly what it does: If it encounters a symlink up to the missing directory, it returns an error signifying there is a symlink -> abort. If it does not, it returns an error signifying that a parent is missing (before it would just return nil), and we create the directory.
It would probably better be named something like "PathTraverseCheck" as already before, it also reports an error when the path contains a file instead of a directory. Separating these checks into their own function would mean iterating several times over the path components, which while cheap still seems bad.

AudriusButkevicius: My point is, I could easily see this function unintentionally changing the way it behaves and only checking if no symlinks exist up to the point the final path exists.

If you check for /foo/bar/baz and only /foo exists as a non-link, I wouldn't be surprised if this function didn't fail, as no symlinks are traversed, regardless that the final path does not exist.

Lstating the final directory you land in is not hard, given, you're mkdirilall'ing anyway.
Alternative would be to change name of TraversesSymlinks to explain that it also checks existance of the destination.

imsodin: Ok, I am leaving TraversesSymlink alone.

imsodin:
> I don't remember what this looked like, but I can't see any hacks here, apart from the comments.

Hackyness has been removed due to a comment by calmh (along with previously unkown racyness).

calmh:
This is also pending me and I'll get to it again soonish...

calmh:
LGTM. Is good to go?

AudriusButkevicius:
Lgtm too

imsodin:
@st-review merge

lib/model: Improve scan scheduling and dir del during pull (fixes #4475 #4476)

st-review:
:ok_hand: Merged as cce634f340b472b781f66bd86a1b6954c34ed7fd. Thanks, @imsodin!

calmh:
Github doesn't understand multiple issues like that, you need `fixes #4475, fixes #4476`. I fixed 4476 manually.

imsodin:
Oh, annoying - that makes it even harder to pack anything meaningful into the commit header. I'll have to go with short keywords and a description in the body in the future - thanks for catching and fixing this.

PR #4498, opened 2017-11-08
lib/model: Incremental block stats usage reporting

AudriusButkevicius:
...

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -104,7 +104,7 @@ type modelIntf interface {
 	CurrentSequence(folder string) (int64, bool)
 	RemoteSequence(folder string) (int64, bool)
 	State(folder string) (string, time.Time, error)
-	UsageReportingStats(version int) map[string]interface{}
+	UsageReportingStats(version int, preview bool) map[string]interface{}
 }
 
 type configIntf interface {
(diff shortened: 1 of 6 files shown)


imsodin:
What is the plan to account for the difference if the folder was restarted since the last report or not?
I don't know exactly what you do when analyzing the data, but you could record the folder start time/block stat reset time to get information about the duration in which this data was accumulated.

AudriusButkevicius:
Probably the stats need to live somewhere outside of the folder, and the sharedPullerState should just increment some global.

calmh:
Since we only care about the ratio between the numbers I don't think it's a big deal. This is probably sane anyway, we'll get less but fresher data and if we want more aggregation we can do it over several days server side.

calmh:
@st-review lgtm

st-review:

imsodin:
Sure, for the important metrics this makes sense. I was bringing this up because Audrius already added this nice new "Totals to date" statistics, which is obviously not very important, but flashy, and this particular metric will be hampered by this. So I guess lgtm from my side unless @AudriusButkevicius rather wants to implement his "outside of folder" proposal?

AudriusButkevicius:
Moved the stats to the global variable.

AudriusButkevicius:
@st-review merge

st-review:

st-review:
:ok_hand: Merged as 386cb274bdd3bbf53662420c0eebdd16a5ee9abc. Thanks, @AudriusButkevicius!

PR #4500, opened 2017-11-09
lib/db: Refactor away the large genericReplace thing

calmh:
### Porpoise

This removes a significant, complex chunk of database code. The "replace" operation walked both the old and new in lockstep and made the relevant changes to make the new situation correct. But since delta indexes we pretty much never need this - we just used replace to drop the existing data and start over.

This makes that explicit and removes the complexity.

(This is one of those things that would be annoying to make case insensitive, while the actual "drop and then insert" that we do is easier.)

### Testing

This is fairly well unit tested...

The one change to the tests is to cover the fact that previously replace with something identical didn't bump the sequence number, while obviously removing everything and re-inserting does. This is not behavior we depend on anywhere.


AudriusButkevicius:
Lgtm, but I'll leave to @imsodin to comment, as he was knee deep in the db stuff recently.

lib/db/set.go
```diff
@@ -146,29 +152,39 @@ func NewFileSet(folder string, fs fs.Filesystem, db *Instance) *FileSet {
...
 	normalizeFilenames(fs)
 
 	s.updateMutex.Lock()
 	defer s.updateMutex.Unlock()
 
+	s.dropFilesLocked(device)
+	s.updateLocked(device, fs)
+}
+
+func (s *FileSet) DropFiles(device protocol.DeviceID) {
```
imsodin: We don't use "Flles anywhere else", just `Drop` would be the consistent thing.

lib/db/set.go
```diff
@@ -146,29 +152,39 @@ func NewFileSet(folder string, fs fs.Filesystem, db *Instance) *FileSet {
...
-		}
+		s.sequence = 0
 	} else {
-		s.remoteSequence[device] = maxSequence(fs)
+		s.remoteSequence[device] = 0
 	}
-	s.db.replace([]byte(s.folder), device[:], fs, &s.localSize, &s.globalSize)
+}
+
+func (s *FileSet) dropFilesLocked(device protocol.DeviceID) {
```
imsodin: Same as above with leaving the "Files" away.

lib/db/set.go
```diff
@@ -146,29 +152,39 @@ func NewFileSet(folder string, fs fs.Filesystem, db *Instance) *FileSet {
 
 func (s *FileSet) Replace(device protocol.DeviceID, fs []protocol.FileInfo) {
```
imsodin: Given that it does now something else than before and is only used once, I would personally remove it entirely and call `Drop` and `Update` explicitly the one time it is used in the model.

imsodin:
@st-review lgtm

st-review:

AudriusButkevicius:
@st-review lgtm

st-review:
:ok_hand: Merged as d7d45d8092e973dac2fd3cac58ba0a5a45c41f51. Thanks, @calmh!

PR #4502, opened 2017-11-10
lib/watchaggregator: Don't care about timings during testing on darwin

imsodin:
This now stops checking the timings of notifications on darwin, but still does it on other platforms. I guess that's an ok compromise - internal timings will hardly vary between platforms.
While at it I renamed s/new/newAggregator/, that was a leftover of un-exporting `New`. And I sped up timings a bit (not waiting anymore for timeouts once everything we need was received).

### DIFF
--- a/lib/watchaggregator/aggregator.go
+++ b/lib/watchaggregator/aggregator.go
@@ -110,7 +110,7 @@ type aggregator struct {
 	ctx                   context.Context
 }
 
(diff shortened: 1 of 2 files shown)


calmh:
@st-review merge

st-review:
:ok_hand: Merged as ec4c3bae0d432e352375eb6fd3bf5a6e65326d1c. Thanks, @imsodin!

PR #4508, opened 2017-11-13
lib/model: Properly schedule pull on reconnect (fixes #4504)

### LINKED ISSUE #4504: Pulls not triggered correctly on reconnection
Since f6ea2a7f there is an issue where a disconnect during sync, which causes "peer went away" errors, does not get resolved on reconnect.

Reproducer:

```
jb@unu:~/s/g/s/s/test $ go test -v -tags integration -run Reconnect
=== RUN   TestReconnectReceiverDuringTransfer
2017/11/13 00:35:23 Cleaning...
2017/11/13 00:35:24 Generating files...
2017/11/13 00:35:24 Starting up...
2017/11/13 00:35:24 Starting instance 1...
2017/11/13 00:35:27 Starting instance 2...
2017/11/13 00:35:31 Pausing receiver...
2017/11/13 00:35:31 Resuming receiver...
2017/11/13 00:35:32 Pausing receiver...
2017/11/13 00:35:32 Resuming receiver...
(hangs here, s2 is out of sync, connected, and not pulling)
```

To be resolved before 0.14.41-rc.1

calmh:
### Purpose

We need to reset prevSeq so that we force a full check when someone reconnects - the sequence number may not have changed due to the reconnect. (This is a regression; we did this before f6ea2a7.)

Also add an optimization: we schedule a pull after scanning, but there is no need to do so if no changes were detected. This matters now because the scheduled pull actually traverses the database which is expensive.

This, however, makes the pull not happen on initial scan if there were no changes during the initial scan. Compensate by always scheduling a pull after initial scan in the rwfolder itself.

### Testing

Previously failing integration test now passes

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1877,8 +1877,6 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 		return err
 	}
 
-	defer runner.SchedulePull()
-
 	// Clean the list of subitems to ensure that we start at a known
 	// directory, and don't scan subdirectories of things we've already
 	// scanned.
@@ -1918,6 +1916,15 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 
 	batch := make([]protocol.FileInfo, 0, maxBatchSizeFiles)
 	batchSizeBytes := 0
+	changes := 0
+
+	// Schedule a pull after scanning, but only if we actually detected any
+	// changes.
+	defer func() {
+		if changes > 0 {
+			runner.SchedulePull()
+		}
+	}()
 
 	for f := range fchan {
 		if len(batch) == maxBatchSizeFiles || batchSizeBytes > maxBatchSizeBytes {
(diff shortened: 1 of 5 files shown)


imsodin:
I think this makes `prevSeq` completely moot, as it will always be zero (except for failed pulls, but there it will always be different from the remote sequence, otherwise the pull wouldn't fail in the first place). This is actually just a consequence of https://github.com/syncthing/syncthing/commit/f6ea2a7f8e5077b7b279cd680400509bccbf557e that I missed together with the connection case that you fix here.

imsodin:
Actually, the integration test in the issue works on my machine and I don't understand the issue conceptually: When pulling does not succeed, `prevSeq` is not updated, so the remote sequence should be higher and thus a pull should happen after reconnect.

AudriusButkevicius:
So, wazzup with zis?

imsodin:
As a pull is now only scheduled when really necessary, the `prevSeq` is not needed any more and can be disposed of. It is set to 0 on every pull anyway.

calmh:
I've incorporated that commit

imsodin:
I haven't really looked at the integration test changes - they run on my machine, that's good enough for me.

imsodin:
@st-review lgtm

st-review:

AudriusButkevicius:
@st-review lgtm

st-review:
:ok_hand: Merged as 5f4ed66aa1b0cd22f96503cf91279b64f9f9ffdc. Thanks, @calmh!

PR #4511, opened 2017-11-14
lib/connections: Consistent log levels & polish (fixes #4510)

### LINKED ISSUE #4510: Logging at info level and above should always include context
This is very low importance, it just came up with a recent report of connection issues:

Logging like `l.Infoln(err)` leads to errors like 

```
[NPNSQ] 22:22:01 INFO: set tcp 192.168.178.24:6736->192.168.178.20:22000: protocol not available
```

whose origin and context are not immediately clear. As this is user facing information, it should contain a minimum of information. In the example of the tcp listener this could be in line with existing logging:

```
			l.Infoln("Listen (BEP/tcp): failed to set options:", err)
```

The alternative is to downgrade it to debug logging level, where the context is evident from file names and line numbers. In the case of the tcp listener, [similar failures are on different levels](https://github.com/syncthing/syncthing/blob/master/lib/connections/tcp_listen.go#L105). Is this on purpose or should all of them be either debug or info level?

imsodin:
As Audrius mentioned in the linked issue, these functions would probably benefit from a general overhaul. Until that happens, this makes logging more consistent and less confusion to the user (info). It still logs to info level when the service fails at setup. During operation there was a mix of info and debug level logging, which now all happens on debug.

### DIFF
--- a/lib/connections/relay_listen.go
+++ b/lib/connections/relay_listen.go
@@ -62,6 +62,9 @@ func (t *relayListener) Serve() {
 
 	oldURI := clnt.URI()
 
+	l.Infof("Relay listener (%v) starting", t)
+	defer l.Infof("Relay listener (%v) shutting down", t)
+
 	for {
 		select {
 		case inv, ok := <-invitations:
@@ -71,13 +74,13 @@ func (t *relayListener) Serve() {
 
 			conn, err := client.JoinSession(inv)
 			if err != nil {
-				l.Infoln("Joining relay session (BEP/relay):", err)
+				l.Debugln("Listen (BEP/relay): failed to join session:", err)
 				continue
 			}
 
 			err = dialer.SetTCPOptions(conn)
 			if err != nil {
-				l.Infoln(err)
+				l.Debugln(err)
 			}
 
 			err = dialer.SetTrafficClass(conn, t.cfg.Options().TrafficClass)
@@ -95,7 +98,7 @@ func (t *relayListener) Serve() {
 			err = tlsTimedHandshake(tc)
 			if err != nil {
 				tc.Close()
-				l.Infoln("TLS handshake (BEP/relay):", err)
+				l.Debugln("Listen (BEP/relay): TLS handshake:", err)
 				continue
 			}
 
--- a/lib/connections/relay_listen.go
+++ b/lib/connections/relay_listen.go
@@ -50,7 +50,7 @@ func (t *relayListener) Serve() {
 		t.mut.Lock()
 		t.err = err
 		t.mut.Unlock()
-		l.Warnln("listen (BEP/relay):", err)
+		l.Warnln("Listen (BEP/relay):", err)
 		return
 	}
 
@@ -74,18 +74,18 @@ func (t *relayListener) Serve() {
 
 			conn, err := client.JoinSession(inv)
 			if err != nil {
-				l.Debugln("Listen (BEP/relay): failed to join session:", err)
+				l.Infoln("Listen (BEP/relay): joining session:", err)
 				continue
 			}
 
 			err = dialer.SetTCPOptions(conn)
 			if err != nil {
-				l.Debugln("Listen (BEP/relay): failed to set options:", err)
+				l.Debugln("Listen (BEP/relay): setting tcp options:", err)
 			}
 
 			err = dialer.SetTrafficClass(conn, t.cfg.Options().TrafficClass)
 			if err != nil {
-				l.Debugln("Listen (BEP/relay): failed to set traffic class:", err)
+				l.Debugln("Listen (BEP/relay): setting traffic class:", err)
 			}
 
 			var tc *tls.Conn
@@ -98,7 +98,7 @@ func (t *relayListener) Serve() {
 			err = tlsTimedHandshake(tc)
 			if err != nil {
 				tc.Close()
-				l.Debugln("Listen (BEP/relay): TLS handshake:", err)
+				l.Infoln("Listen (BEP/relay): TLS handshake:", err)
 				continue
 			}
 
--- a/lib/connections/relay_listen.go
+++ b/lib/connections/relay_listen.go
@@ -80,12 +80,12 @@ func (t *relayListener) Serve() {
 
 			err = dialer.SetTCPOptions(conn)
 			if err != nil {
-				l.Debugln(err)
+				l.Debugln("Listen (BEP/relay): failed to set options:", err)
 			}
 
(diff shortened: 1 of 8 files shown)


lib/connections/relay_listen.go
```diff
@@ -71,13 +74,13 @@ func (t *relayListener) Serve() {
...
 			if err != nil {
-				l.Infoln("Joining relay session (BEP/relay):", err)
+				l.Debugln("Listen (BEP/relay): failed to join session:", err)
 				continue
 			}
 
 			err = dialer.SetTCPOptions(conn)
 			if err != nil {
-				l.Infoln(err)
+				l.Debugln(err)
```
AudriusButkevicius: Need context

lib/connections/tcp_listen.go
```diff
@@ -104,7 +104,7 @@ func (t *tcpListener) Serve() {
 
 		err = dialer.SetTCPOptions(conn)
 		if err != nil {
-			l.Infoln(err)
+			l.Debugln(err)
 		}
```
AudriusButkevicius: Context

lib/connections/tcp_dial.go
```diff
@@ -39,7 +39,7 @@ func (d *tcpDialer) Dial(id protocol.DeviceID, uri *url.URL) (internalConn, erro
 
 	err = dialer.SetTCPOptions(conn)
 	if err != nil {
-		l.Infoln(err)
+		l.Debugln(err)
 	}
```
AudriusButkevicius: Context

lib/upnp/upnp.go
```diff
@@ -159,7 +159,7 @@ USER-AGENT: syncthing/1.0
 
 	err = socket.SetDeadline(time.Now().Add(timeout))
 	if err != nil {
-		l.Infoln(err)
+		l.Debugln(err)
```
AudriusButkevicius: Context

lib/upnp/upnp.go
```diff
@@ -168,7 +168,7 @@ USER-AGENT: syncthing/1.0
 	_, err = socket.WriteTo(search, ssdp)
 	if err != nil {
 		if e, ok := err.(net.Error); !ok || !e.Timeout() {
-			l.Infoln(err)
+			l.Debugln(err)
```
AudriusButkevicius: Context

lib/upnp/upnp.go
```diff
@@ -168,7 +168,7 @@ USER-AGENT: syncthing/1.0
...
 		}
```
AudriusButkevicius: You should also tweak SetTCPOptions to not fail on first failure.

lib/connections/relay_listen.go
imsodin: Actually I considered debug logging without context as ok, as the context can be established from file names and line numbers. I will add some (generic) context anyway.

lib/upnp/upnp.go
imsodin: I don't really know what you want here.

lib/connections/relay_listen.go
```diff
@@ -71,18 +74,18 @@ func (t *relayListener) Serve() {
 
 			conn, err := client.JoinSession(inv)
 			if err != nil {
-				l.Infoln("Joining relay session (BEP/relay):", err)
+				l.Debugln("Listen (BEP/relay): failed to join session:", err)
```
AudriusButkevicius: Should we really hide these? It's one of the only ways to debug the issues that are happening now.

imsodin: That was what I was asking originally. I was confused by what looked like randomly chosen info or debug levels with or without context (e.g. SetTCPOptions was info but without context, SetTrafficClass was debug but with context). I can't really judge their relevance, so pointers whether a log statement should be info or debug like this one are appreciated.

AudriusButkevicius: Other errors that are returned by dial functions etc are logged a layer up, in the loop, so it's ok, yet this is the only place that reports relay issues, so it's not cool to hide it.

lib/connections/relay_listen.go
```diff
@@ -95,7 +98,7 @@ func (t *relayListener) Serve() {
 			err = tlsTimedHandshake(tc)
 			if err != nil {
 				tc.Close()
-				l.Infoln("TLS handshake (BEP/relay):", err)
+				l.Debugln("Listen (BEP/relay): TLS handshake:", err)
```
AudriusButkevicius: Same here, only place that reports listen errors, so not cool to hide it.

lib/connections/tcp_listen.go
```diff
@@ -104,18 +104,18 @@ func (t *tcpListener) Serve() {
...
 		if err != nil {
-			l.Debugf("failed to set traffic class: %s", err)
+			l.Debugln("Listen (BEP/tcp): failed to set traffic class:", err)
 		}
 
 		tc := tls.Server(conn, t.tlsCfg)
 		err = tlsTimedHandshake(tc)
 		if err != nil {
-			l.Infoln("TLS handshake (BEP/tcp):", err)
+			l.Debugln("TLS handshake (BEP/tcp):", err)
```
AudriusButkevicius: Not cool to hide.

lib/connections/relay_listen.go
imsodin: I will make everything log on info level that results in an early return/continue/break, does that sound sensible?

AudriusButkevicius: Well I think I pointed out the relevant places in this change, but yeah, anything that doesn't return an error should probably be logged.

lib/connections/relay_listen.go
imsodin: The dial logging was only on debug, I changed that to info and removed some redundant debug log statements in the dial functions themselves.

lib/connections/service.go
```diff
@@ -401,7 +401,7 @@ func (s *Service) connect() {
 
 				conn, err := dialer.Dial(deviceID, uri)
 				if err != nil {
-					l.Debugln("dial failed", deviceCfg.DeviceID, uri, err)
+					l.Infof("%v for %v at %v: %v", dialerFactory, deviceCfg.DeviceID, uri, err)
```
AudriusButkevicius: This might flood, as we might have old addresses, so better this stays debug.

imsodin: ok

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 4efff736b3e0d368afc0cda34c6baeef0c6be425. Thanks, @imsodin!

PR #4514, opened 2017-11-16
lib/versioner: Purge the empty directories in .stversions (fixes #4406)

### LINKED ISSUE #4406: Empty directories in .stversions not purged
When using staggered file versioning folder structure in .stversions is never deleted. Even after all files in a given folder expire in .stversions. Can test behaviour with other file versioning methods if needed but assume the same.

On another note enabled usage reporting on all devices. Not a decision made lightly but trust Syncthing and transparent usage reporting and is (very) small way to give back.

### Version Information

Syncthing Version: v0.14.38

dsaveliev:
This PR is intended to improve cleanup logic in Trash Can and Staggered File Versioning (fixes #4406)

In the current code `(*Staggered).clean()` method removes only the bottom directories in the nesting hierarchy. 
`(*Trashcan) cleanoutArchive()` method has similar behaviour, moreover, it is trying to remove a directory with non-empty subdirectory (fortunately, `Filesystem.Remove` is unable to make recursive deletion and just returns an error).

This PR introduces an auxiliary `filesCounter` struct,  which is able to track files count for all directories of `.stversions`. 
It remembers bypass order of `Filesystem.Walk` method and checks folders in reverse order - from the bottom to the root, so we can find all empty directories without recursive logic and heavy system calls.

Here are the tests for `filesCounter` as well as additional cases for Trash Can tests.

I would appreciate any thoughts and advice.


calmh:
Cool, looks reasonable and good first (here, at least) PR!

I think we can do it a little bit more efficiently though. In principle we could do deletes while walking, as we know on the way "out" from a directory whether it was empty or not, but that gets tricky when we've deleted directories and so on so I think the "one pass to look, one pass to clean" approach is fine.

The files counter doesn't actually need to count files though - it just needs to check if there were any at all. As such, I think your addDir can do roughly what it does, but addFile can just remove the parent dir from the map. The map then just tracks dirs that are, as far as we know, empty. This avoid keeping state for the entire tree, instead just keeping the dirs that will in fact be removed.

In the end, just collecting the keys of that map and doing a reverse sort will get them in the order they can be removed.

What do you think?

AudriusButkevicius:
Can't we just DirNames after a delete to check if there is anything left? This seems like a lot of moving parts.

dsaveliev:
I agree with you. 
Cleanup logic has one subtlety - we need to go from the bottom to the top of nesting hierarchy to remove directories containing only empty subdirectories.
I could keep bypass order in the simple slice and check each directory with `DirNames` at the end of cleanup methods. It should simplify this PR.

Does it make sense?

AudriusButkevicius:
Just do a second walk removing empty dirs, no extra classes/trackers needed.

dsaveliev:
Ok, I will do it.

imsodin:
> Just do a second walk removing empty dirs, no extra classes/trackers needed. Or keep the list as you suggested, and recheck later.

I think that wont work because the paths are walked bottom up, so you can only remove empty leaf directories.

dsaveliev:
> The files counter doesn't actually need to count files though - it just needs to check if there were any at all. As such, I think your addDir can do roughly what it does, but addFile can just remove the parent dir from the map. The map then just tracks dirs that are, as far as we know, empty. This avoid keeping state for the entire tree, instead just keeping the dirs that will in fact be removed.

It sounds reasonable, the only thing that every time we find a file we need to remove all parent dirs in the file path from the map, not only direct parent directory.

I made a simplified version of this logic using `DirNames`, as suggested by @AudriusButkevicius.
I think it could make sense to move this code to `util.go` to remove code duplication.

AudriusButkevicius:
So this looks mostly fine, I guess I wouldn't Warn() in the removal code, as it's not a fatal thing, so I'd just debug log.

Perhaps moving this whole process into a utility function what walks the tree a second time is not such a bad idea, so that the intrusion into the versioners is minimal, and all they have to do is call one function.

imsodin:
I don't think it is a good idea to walk the tree again and use DirNames on it just to remove empty directories. This might be cheap on the current one and only basic filesystem type, but if the fs is e.g. something in the cloud, such calls can be expensive/slow. As we have all the information from the first walk, I think we should use that (as done in the first commit).

AudriusButkevicius:
Fine.

dsaveliev:
Here is the original version with some improvements - implemented logic recommended by @calmh. 
The counters have been deleted and I just keep track of the dirs which should be removed.

AudriusButkevicius:
I still think this is too much for what it's doing. Extra structs, sorting interface (which I am not sure why it's even there as strings get sorted alphabetically and by length anyway), just to check if a directory is empty

imsodin:
@AudriusButkevicius What do you propose instead - writing out the (admittedly small) code snippets of the filesCounter directly in the versioners, using fs operations as proposed earlier or something different?

lib/versioner/files_counter.go
```diff
@@ -0,0 +1,61 @@
...
+func (s byLength) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+func (s byLength) Less(i, j int) bool {
+	return len(s[i]) > len(s[j])
+}
+
+// filesCounter is intended to track empty directories
+type filesCounter struct {
```
imsodin: just `type filesCounter map[string]bool/` would do.

AudriusButkevicius:
Actually, the struct is fine, yet I don't understand why we care about sort order.
Also you should use `map[string]struct{}` which is smaller than bool.

AudriusButkevicius:
Otherwise, this is probably as good as it gets.

lib/versioner/trashcan.go
```diff
@@ -164,10 +157,13 @@ func (t *Trashcan) cleanoutArchive() error {
 		return err
 	}
 
-	// The last directory seen by the walkFn may not have been removed as it
-	// should be.
-	if currentDir != "" && filesInDir == 0 {
-		t.fs.Remove(currentDir)
+	for _, path := range filesCounter.emptyDirs() {
```
imsodin: While your at it: This entire for loop could also be a convenience function like `func (c *filesCounter) deleteEmptyDirs(fs fs.filesystem)`.

lib/versioner/files_counter.go
imsodin: Also on the very nitpicky level: This doesn't count files anymore, maybe emptyDirTracker (or something nicer) would be more appropriate.

AudriusButkevicius: Agree

lib/versioner/trashcan.go
dsaveliev: Done. I decided not to return an error from `deleteEmptyDirs` and instead left a log warning if you don't mind.

dsaveliev:
> Also you should use map[string]struct{} which is smaller than bool.
Yep, I forgot about it. Fixed.

AudriusButkevicius:
@st-review lgtm

st-review:

lib/versioner/files_counter.go
```diff
@@ -0,0 +1,58 @@
...
+	"path/filepath"
+	"sort"
+
+	"github.com/syncthing/syncthing/lib/fs"
+)
+
+// filesCounter is intended to track empty directories
+type filesCounter struct {
+	dirs map[string]struct{}
+}
```
imsodin: The previous comments still apply:
The type can simply be
```
type filesCounter map[string]struct{}
```

And it should probably be renamed, as it is no longer counting files (e.g. "emptyDirTracker").

dsaveliev: Done. Naming is hard so I picked up your variant.

imsodin:
@st-review lgtm

st-review:
@imsodin: Build status is `failure` -- refusing to merge.

AudriusButkevicius:
@dsaveliev can you add yourself to authors and nicks files?

dsaveliev:
> @dsaveliev can you add yourself to authors and nicks files?

Sure, done.

AudriusButkevicius:
@st-review merge

st-review:

st-review:
:ok_hand: Merged as 9471b9f6af2aa84720ae92f4f0b748acb84ff306. Thanks, @dsaveliev!

dsaveliev:
Thanks for your patience!

PR #4517, opened 2017-11-17
Fix discovery in the absence of listen addresses (fixes #4418)

### LINKED ISSUE #4418: No Global Discovery without Sync Protocol Listen Address
When setting "Sync Protocol Listen Address" to a blank value, Syncthing starts without listening ports, as intended. This is a laptop always behind NAT anyway, and it feels more secure this way.

However, all Discovery servers fail with `nothing to announce`, which seems unexpected, as it should still be possible to use Global Discovery to connect to reachable servers or via relay.

### Version Information

Syncthing Version: v0.14.38
OS Version: Mac OS X (64 bit)

calmh:
### Purpose

This makes it OK to not have any listeners working. Specifically,

- We don't complain about an empty listener address
- We don't complain about not having anything to announce to global
  discovery servers
- We don't send local discovery packets when there is nothing to
  announce.

The last point also fixes a thing where the list of addresses for local
discovery was set at startup time and never refreshed.

### Testing

I fiddled with it manually and it seemed to work.

### DIFF
--- a/lib/discover/global.go
+++ b/lib/discover/global.go
@@ -121,6 +121,7 @@ func NewGlobal(server string, cert tls.Certificate, addrList AddressLister) (Fin
 		noAnnounce:     opts.noAnnounce,
 		stop:           make(chan struct{}),
 	}
+	cl.setError(errors.New("not announced"))
 
 	return cl, nil
 }
@@ -210,6 +211,7 @@ func (c *globalClient) sendAnnouncement(timer *time.Timer) {
 		// There are legitimate cases for not having anything to announce,
 		// yet still using global discovery for lookups. Do not error out
 		// here.
+		c.setError(nil)
 		timer.Reset(announceErrorRetryInterval)
 		return
 	}
--- a/lib/discover/global.go
+++ b/lib/discover/global.go
@@ -121,7 +121,6 @@ func NewGlobal(server string, cert tls.Certificate, addrList AddressLister) (Fin
 		noAnnounce:     opts.noAnnounce,
 		stop:           make(chan struct{}),
 	}
-	cl.setError(errors.New("not announced"))
 
 	return cl, nil
 }
@@ -208,8 +207,9 @@ func (c *globalClient) sendAnnouncement(timer *time.Timer) {
 	}
 
 	if len(ann.Addresses) == 0 {
-		c.setError(errors.New("nothing to announce"))
-		l.Debugln("Nothing to announce")
+		// There are legitimate cases for not having anything to announce,
+		// yet still using global discovery for lookups. Do not error out
+		// here.
 		timer.Reset(announceErrorRetryInterval)
 		return
 	}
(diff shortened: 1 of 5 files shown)


lib/discover/local.go
```diff
@@ -112,24 +112,41 @@ func (c *localClient) Error() error {
...
 	msg = append(msg, bs...)
 
+	return msg, true
+}
+
+func (c *localClient) sendLocalAnnouncements() {
+	var msg []byte
 	for {
-		c.beacon.Send(msg)
+		var ok bool
```
AudriusButkevicius: why not move the ok outside of the loop next to msg?

lib/discover/local.go
```diff
@@ -112,24 +112,41 @@ func (c *localClient) Error() error {
...
+	}
 	binary.BigEndian.PutUint32(msg, Magic)
 
-	var pkt = c.announcementPkt()
+	pkt := Announce{
+		ID:         c.myID,
+		Addresses:  addrs,
+		InstanceID: rand.Int63(),
+	}
 	bs, _ := pkt.Marshal()
```
AudriusButkevicius: Perhaps not throw away this error, and return err == nil as the boolean?

calmh: It's impossible for this marshal to fail afaik

lib/discover/local.go
```diff
@@ -132,7 +132,7 @@ func (c *localClient) announcementPkt(msg []byte) ([]byte, bool) {
 	pkt := Announce{
 		ID:         c.myID,
 		Addresses:  addrs,
-		InstanceID: rand.Int63(),
+		InstanceID: instanceID,
 	}
```
AudriusButkevicius: Why this change?

calmh: Previously the packet was allocated outside of the loop, so got the instance ID at startup as was intended. My first change moved the randomness inside the loop, which is wrong. This fixes that.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 7ebf58f1bcc1874bc05cdf4ad0d780a854ca8095. Thanks, @calmh!

lib/discover/global.go
```diff
@@ -208,8 +208,10 @@ func (c *globalClient) sendAnnouncement(timer *time.Timer) {
 	}
 
 	if len(ann.Addresses) == 0 {
-		c.setError(errors.New("nothing to announce"))
-		l.Debugln("Nothing to announce")
+		// There are legitimate cases for not having anything to announce,
+		// yet still using global discovery for lookups. Do not error out
+		// here.
+		c.setError(nil)
 		timer.Reset(announceErrorRetryInterval)
```
imsodin: Shouldn't this be `defaultReannounceInterval`?
I know this PR is already merged (are we very productive today? ;) ), but still seems like the right place to nag about it.

calmh: I think not, because it is sort of an error that we have no addresses to announce, possibly a temporary one. If so we gain by checking again a bit sooner. Clearly this is subjective, but it was at least intentional and not by mistake I retained that interval. :)

PR #4521, opened 2017-11-17
lib/fs: Add case insensitivity to MtimeFS

calmh:
This is step one of a hundred fifty on the path to case insensitivity.
It brings in the basic case folding mechanism and adds it to the
mtimefs, as this is something outside the fileset that touches stuff in
the database based on name. No effort to convert or handle existing
entries when the insensitivity is changed, I don't think we need it...

Useless by itself but includes tests and will reduce the review load
along the way.

### DIFF
--- a/lib/fs/folding.go
+++ b/lib/fs/folding.go
@@ -1,3 +1,9 @@
+// Copyright (C) 2017 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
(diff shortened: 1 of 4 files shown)


AudriusButkevicius:
@st-review merge

st-review:

st-review:
@AudriusButkevicius: Patiently waited 30m0s for the build status to turn green, but enough is enough.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as ee5d0dd43fe13788ef6ac86faf12f42029f455ea. Thanks, @calmh!

PR #4522, opened 2017-11-17
lib/model: Don't update ignore hash when pull fails

imsodin:
While reviewing the PR about pull scheduling on reconnect I realized that the ignores hash is incorrectly updated even if a pull fails. This means on a subsequent pull, ignored files aren't included anymore, which is relevant due to #4460, and should thus also be in v0.14.41.

### DIFF
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -239,17 +239,17 @@ func (f *sendReceiveFolder) pull(prevIgnoreHash string) (curIgnoreHash string, s
 		return prevIgnoreHash, true
 	}
 
+	if err := f.CheckHealth(); err != nil {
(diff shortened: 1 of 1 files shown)


calmh:
Won't we have set the correct bits on all the ignored files as part of the pulling, even if some other files fail to pull for whatever reason?

imsodin:
Yes, that case is fine, the problem is the other way round: If a file got un-ignored and fails to be pulled, it won't get queried again.

calmh:
Ah

calmh:
@st-review merge

st-review:
:ok_hand: Merged as 6cf01c1d30ad4d74ca4ecb3ccf34716cca9b93db. Thanks, @imsodin!

PR #4523, opened 2017-11-17
lib/connections: Only announce punchable nats (fixes #4519)

### LINKED ISSUE #4519: Avoid lots and lots of announced addresses in the presence of symmetric NAT
When behind symmetric NAT we will get a new port mapping for every STUN lookup, which is (worst case) every 24 seconds. Worst case this means we accumulate up to 150 useless KCP addresses (one hour's worth) on the discovery server.

We should probably just not announce any KCP addresses to the global discovery system when we are behind an unfriendly NAT type.

Relates to #4459

AudriusButkevicius:
Well Tested (TM)

### DIFF
--- a/lib/connections/kcp_listen.go
+++ b/lib/connections/kcp_listen.go
@@ -246,6 +246,13 @@ func (t *kcpListener) stunRenewal(listener net.PacketConn) {
 			if oldType != natType {
 				l.Infof("%s detected NAT type: %s", t.uri, natType)
 				t.nat.Store(natType)
+				oldType = natType
+			}
+
+			// We can't punch through this one, so no point doing keepalives
+			// and such, just try again in a minute and hope that the NAT type changes.
+			if !isPunchable(natType) {
+				break
 			}
 
 			for {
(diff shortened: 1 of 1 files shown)


lib/connections/kcp_listen.go
```diff
@@ -285,11 +292,10 @@ func (t *kcpListener) stunRenewal(listener net.PacketConn) {
...
 				}
 			}
-
-			oldType = natType
 		}
 
-		// We failed to contact all provided stun servers, chillout for a while.
+		// We failed to contact all provided stun servers or the nat is not punchable.
+		// Chillout for a while.
 		time.Sleep(time.Minute)
```
calmh: Constantify and raise? a minute is quite a short time isn't it? or maybe that's fine it's just opinion, but constantify at least.

calmh:
@st-review merge

st-review:

st-review:
:ok_hand: Merged as 0518a92cdb025e952a1a405e19306b51e1736595. Thanks, @AudriusButkevicius!

PR #4533, opened 2017-11-20
lib/connections: Trust the model to tell us if we are connected

AudriusButkevicius:
Well Tested(TM)

This should address issue as described in https://forum.syncthing.net/t/stun-nig-party-with-paused-devices/10942/13
Essentially the model and the connection service goes out of sync in terms of thinking if we are connected or not.
Resort to model as being the ultimate source of truth.

I can't immediately pin down how this happens, yet some ideas.

ConfigSaved happens in separate routine, so it's possbile that we have some sort of device removed yet connection comes in parallel kind of thing.
However, in this case the connection exists in the model, and does not exist in the connection service and the only way for the connection to be removed
in the connection service is device removal from the config.

Given the subject, this might also be related to the device being paused.

Also, adds more info to the logs

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -89,9 +89,6 @@ type Service struct {
 	listeners          map[string]genericListener
 	listenerTokens     map[string]suture.ServiceToken
 	listenerSupervisor *suture.Supervisor
-
-	curConMut         sync.Mutex
-	currentConnection map[protocol.DeviceID]completeConn
 }
 
 func NewService(cfg *config.Wrapper, myID protocol.DeviceID, mdl Model, tlsCfg *tls.Config, discoverer discover.Finder,
@@ -129,9 +126,6 @@ func NewService(cfg *config.Wrapper, myID protocol.DeviceID, mdl Model, tlsCfg *
 			FailureThreshold: 2,
 			FailureBackoff:   600 * time.Second,
 		}),
-
-		curConMut:         sync.NewMutex(),
-		currentConnection: make(map[protocol.DeviceID]completeConn),
 	}
 	cfg.Subscribe(service)
 
@@ -228,15 +222,11 @@ next:
 
 		// If we have a relay connection, and the new incoming connection is
 		// not a relay connection, we should drop that, and prefer this one.
-		connected := s.model.ConnectedTo(remoteID)
-		s.curConMut.Lock()
-		ct, ok := s.currentConnection[remoteID]
-		s.curConMut.Unlock()
-		priorityKnown := ok && connected
+		ct, connected := s.model.Connection(remoteID)
 
 		// Lower priority is better, just like nice etc.
-		if priorityKnown && ct.internalConn.priority > c.priority {
-			l.Debugln("Switching connections", remoteID)
+		if connected && ct.Priority() > c.priority {
+			l.Debugf("Switching connections %s (existing: %s new: %s)", remoteID, ct, c)
 		} else if connected {
 			// We should not already be connected to the other party. TODO: This
 			// could use some better handling. If the old connection is dead but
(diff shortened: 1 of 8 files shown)


lib/model/model.go
```diff
@@ -1331,14 +1331,14 @@ func (cf cFiler) CurrentFile(file string) (protocol.FileInfo, bool) {
 }
 
 // ConnectedTo returns true if we are connected to the named device.
```
imsodin: needs updating

imsodin:
Sounds very reasonable to keep this information in a central place instead of relying on the config commits to remove old connections. However I cannot say I am able to connect the logs in the forum and/or recent changes with this.

lib/connections/service.go
```diff
@@ -394,8 +377,8 @@ func (s *Service) connect() {
 
 				prio := dialerFactory.Priority()
 
-				if priorityKnown && prio >= ct.internalConn.priority {
-					l.Debugf("Not dialing using %s as priority is less than current connection (%d >= %d)", dialerFactory, dialerFactory.Priority(), ct.internalConn.priority)
+				if ok && prio >= ct.Priority() {
```
imsodin: The `ok` needs a proper name. This panics because ok was used for something else since its creation.

calmh:
inb4 details this looks like a good idea, I'd sort of forgotten that we even had connections in more than one place. Do we think this is 0.14.41 material?

AudriusButkevicius:
> Do we think this is 0.14.41 material?

As in, do we think it's 0.14.41 release right now worthy?
I don't think so, as this was here for ages, yet I think we should cut an RC with this.

imsodin:
I think the question was more should this still go into 41 or wait for 42. I believe it should be in 41. It has been like this for a long time, but there were only issues since the 41 RCs and maybe since 40 and my only theory is, that the parallel dial speedup made it problematic (maybe Audrius has more concrete/plausible theories). In my opinion the additional delay to 41 final is worth the chance of more stable connections.

AudriusButkevicius:
There are other problems this could cause, for example it might never fallback to relays when disconnected, because it's delusional that it's connected over TCP (even if it's not).

I suspect this is definately 41 material, and we should shove it in the RC to see what happens.

AudriusButkevicius:
Of course, you two should be the test rats before we cut the RC, as I don't run syncthing.

imsodin:
> Of course, you two should be the test rats before we cut the RC, as I don't run syncthing.

Are you cheating on Syncthing? :P 

First lab rat already died and I don't want to loose another one to the panic I commented on before ;)

AudriusButkevicius:
I fixed the panics, hopefully.

imsodin:
@calmh Your servers are overflowing.

calmh:
@st-review rebuild

imsodin:
Now it works.

On a completely unrelated note: I just realized that the connection was established to the global IP, not the local one. I don't know whether routers are smart enough to shortcut that, but wouldn't it make sense to give local IPs a higher prio even if just for asthetics?

AudriusButkevicius:
We do by sorting the order of addresses, yet with parallel dials that goes to shits...

calmh:
@st-review merge

st-review:
:ok_hand: Merged as 44a542391e534d3836bce9c2ea13ff21b11e806e. Thanks, @AudriusButkevicius!

scienmind:
Is `[IDIDID] 12:01:18 INFO: Connected to already connected device...` expected to be fixed in 0.14.42?
Cause I still get those a lot...

AudriusButkevicius:
No, that's misconfiguration on your side.

scienmind:
> that's misconfiguration on your side.

Can you clarify a bit what may be misconfigured? 

Wherever I have 2+ devices on one LAN talking to 2+ devices on some other LAN over the internet, I get a flood of `Connected to already connected device...`. It started only after KCP went live in stable channel. And I haven't altered the configurations there for a long long time.

imsodin:
I don't know what the cause for your messages are, just that the right place to find out is in the forum not in an already merged PR.

PR #4534, opened 2017-11-20
lib/connections: Fix local address priority

AudriusButkevicius:
Well Tested (TM).

Do not merge, because there are a few points to discuss.

There are two caveats to this:
1. We no longer prefer IPv6 over IPv4 locally (not sure why we did in the past).
2. Connections that do connect, still end up at the base priority, so for example if we connect over WAN for some reason, we will not suddenly switch to a LAN connection, yet given the prioritization is moving, we could make this happen now. One problem could be where each side sees the other one as on LAN on different addresses, and we keep switching between connections.

Yet I suspect this is very unlikely.

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -336,6 +336,8 @@ func (s *Service) connect() {
 				}
 			}
 
+			addrs = util.UniqueStrings(addrs)
+
 			l.Debugln("Reconnect loop for", deviceID, addrs)
 
 			seen = append(seen, addrs...)
@@ -375,9 +377,9 @@ func (s *Service) connect() {
 					continue
 				}
 
-				prio := dialerFactory.Priority()
+				priority := dialerFactory.Priority()
 
-				if connected && prio >= ct.Priority() {
+				if connected && priority >= ct.Priority() {
 					l.Debugf("Not dialing using %s as priority is less than current connection (%d >= %d)", dialerFactory, dialerFactory.Priority(), ct.Priority())
 					continue
 				}
@@ -385,9 +387,18 @@ func (s *Service) connect() {
 				dialer := dialerFactory.New(s.cfg, s.tlsCfg)
 				nextDial[addr] = now.Add(dialer.RedialFrequency())
 
+				// For LAN addresses, increase the priority so that we
+				// try these first.
+				switch {
+				case dialerFactory.AlwaysWAN():
+					// Do nothing.
+				case s.isLANHost(uri.Host):
+					priority -= 1
+				}
+
 				dialTargets = append(dialTargets, dialTarget{
 					dialer:   dialer,
-					priority: prio,
+					priority: priority,
 					deviceID: deviceID,
 					uri:      uri,
 				})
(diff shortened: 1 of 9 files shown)


lib/discover/cache.go
```diff
@@ -134,7 +119,6 @@ func (m *cachingMux) Lookup(deviceID protocol.DeviceID) (addresses []string, err
 	}
 	m.mut.RUnlock()
 
-	addresses = uniqueSortedAddrs(paddresses)
```
imsodin: Shouldn't addresses still be made unique?

imsodin:
I don't understand the scenario when making the LAN/WAN prio permanent with seeing each other on different addresses causing switching between LAN and WAN. If LAN is successful, we switch to that and stay there, if not we stay on WAN. What am I missing?

AudriusButkevicius:
So it's an imaginary scenario where you have two routes between nodes, and not A to B considers P1 as local P2 as WAN, and B to A conders P2 as local, P1 as WAN.

You can also imagine this with AlwaysLocalNets misconfiguration

calmh:
This looks reasonable, needs a small rebase now

imsodin:
I believe the "temporary" priorisation in the current state of the PR is the right way to go for 0.14.41. I think permanently prioritizing local over global connections is the right way to go, but only later. We should delay all new(ish) behaviour for 42 and try to stabilize 41.

@AudriusButkevicius I would prefer successive commits to force pushing - makes secondary reviews much faster (I hardly noticed the uniqueness change).

AudriusButkevicius:
Well in this case we'll end up dialing local addresses first, so it's likely this will do what we want anyway, even. If we don't switch from WAN to LAN.

lib/connections/service.go
```diff
@@ -392,19 +394,29 @@ func (s *Service) connect() {
...
 
+				// For LAN addresses, increase the priority so that we
+				// try these first.
+				offset := 0
+				switch {
+				case dialerFactory.AlwaysWAN():
+					// Do nothing.
+				case s.isLANHost(uri.Host):
+					offset += 1
+				}
```
imsodin: What's the motivation to use this offset instead of just modifying the prio?

imsodin:
Right, I guess it's anyway impossible for a device to get available on LAN without dropping the WAN connection. Anyway, this is lgtm from my side except for the conflict (my code comment is just a style question).

AudriusButkevicius:
Updated.

I usually do forced pushes for small commits, so that you see the whole picture (as it's small enough) rather than incremental progress which might be confusing me renaming A to B and then B to C etc.

imsodin:
@st-review lgtm

st-review:

lib/connections/service.go
```diff
@@ -412,6 +423,14 @@ func (s *Service) connect() {
 	}
 }
 
+func (s *Service) isLANHost(host string) bool {
+	addr, err := net.ResolveIPAddr("ip", host)
```
calmh: Host here comes from `uri.Host` which includes a port number which, again, probably does not resolve here so everything will be considered non-LAN...

calmh:
@st-review lgtm now

st-review:
:ok_hand: Merged as 4922b46fbdc920f499e208738a4829c2a08335f4. Thanks, @AudriusButkevicius!

calmh:
So this was still broken. How about some tests in the future, saving me from releasing broken RCs all the time?

AudriusButkevicius:
Just ask for tests in the PR.

PR #4540, opened 2017-11-22
lib/events: Fix unmarshaling of EventType

monstermunchkin:
### Purpose

This allows to properly unmarshal an `Event`. Prior to this change, unmarshaling would have failed with:
```
json: cannot unmarshal string into Go struct field Event.type of type events.EventType
```

### Testing

A Test has been added which tries unmarshaling an `Event`.

### DIFF
--- a/lib/events/events.go
+++ b/lib/events/events.go
@@ -8,6 +8,7 @@
 package events
 
 import (
+	"encoding/json"
 	"errors"
 	"runtime"
 	"time"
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
When is it actually unmarshalled?

monstermunchkin:
Not in syncthing directly, but when you want to unmarshal a JSON object from the REST API into an `events.Event` struct. If this is not handled here, you would have to redefine the `Event` struct in your own code which is more work than actually needed.

AudriusButkevicius:
@st-review merge

lib/events: Fix unmarshaling of EventType

skip-check: authors

st-review:
:ok_hand: Merged as b2af8f135bcf5f3d76edb44a25bf677dfc13a9c3. Thanks, @monstermunchkin!

PR #4545, opened 2017-11-23
lib/config: Support symlinked root (fixes #4542, fixes #4353)

### LINKED ISSUE #4353: Symlinks as folder root broken on Windows
After updating to 0.14.37, all root folders which are referenced by symlink result in "Failed initial scan (rw) of...". Verified by creating two new sync folders, one referencing an actual folder's location, the other a symlink to a folder on another drive.

A current workaround of the problem is to set the path directly to the folder (no symlink).

Syncthing Version: v0.14.37
OS Version: Windows 10

### LINKED ISSUE #4542: Support OneDrive folders on recent Windows 10
Syncthing (v0.14.40) does not work with Onedrive on Windows 10 Fall Creator Update (version 1709) properly in some cases. 

In Win 10 Fall Creators Update, Microsoft changed how Onedrive operates NTFS filesystem. Now OneDrive creates [reparsepoints](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365503(v=vs.85).aspx) (an NTFS feature). Syncthing forum user "geeksquad" explained it [here](https://forum.syncthing.net/t/folder-path-missing-even-the-path-exists-onedrive-win10-fall-creators-update/10913/4). It seems Syncthing is refusing to recognize a OneDrive directory using _reparsepoints_ and reports an error message "folder path missing". 

> [ZEER7] 12:49:24 WARNING: Creating folder marker: folder path missing
> [ZEER7] 12:49:24 INFO: Restarted folder "EXAMPLE" (xxxxx-yyyyy) (readwrite)
> [ZEER7] 12:49:24 INFO: Connection to WJXMAFI closed: reading length: read tcp 192.168.1.2:51413->192.168.1.3:22000: use of closed network connection
> [ZEER7] 12:49:24 WARNING: Error on folder "EXAMPLE" (xxxxx-yyyyy): folder path missing
> [ZEER7] 12:49:24 INFO: Failed initial scan of readwrite folder "EXAMPLE" (xxxxx-yyyyy)

I have 2 OneDrive synced directories on a same computer synced with Syncthing to a Linux server. Based on my experiments the following factors may impact on the success of getting Syncthing to sync a OneDrive folder: 

1. I have OneDrive Business and standard OneDrive folders and only the ordinary OneDrive folder syncs properly. There might or might not be a issue related to Onedrive Business. 
2.  In case of the (standard) OneDrive, the synced folder is not the root OneDrive folder, but a folder under the Onedrive root folder. Syncthing syncs this folder under the OneDrive root folder OK.
3. In the case of the OneDrive Business folder, I am trying to sync the OneDrive Business root folder, but after updating to Win10 Fall Creators Update, Syncthing stopped syncing this folder. 
4. Syncthing does not sync properly any OneDrive or OneDrive Business folder IF _Files On-demand_ feature has been enabled on the OneDrive. Microsoft introduced this feature in Win 10 Fall Creators Update and it is enabled by default. I have disabled it manually. 

### Version Information

Syncthing Version: v0.14.40 Win 64-bit
OS Version: Windows 10 Fall Creator Update (Version >= 1709 )

AudriusButkevicius:
Well Tested (TM)

### DIFF
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -457,35 +457,29 @@ func TestFolderCheckPath(t *testing.T) {
 	}
 	fd.Close()
 
-	eq := func(err error) func(error) bool {
-		return func(ierr error) bool {
-			return err == ierr
-		}
-	}
-
 	testcases := []struct {
-		path    string
-		checker func(error) bool
+		path string
+		err  error
 	}{
 		{
-			path:    "",
-			checker: eq(nil),
+			path: "",
+			err:  errMarkerMissing,
 		},
 		{
-			path:    "file",
-			checker: eq(errPathNotDirectory),
+			path: "file",
+			err:  errPathNotDirectory,
 		},
 		{
-			path:    "does not exist",
-			checker: fs.IsNotExist,
+			path: "does not exist",
+			err:  errPathMissing,
 		},
 		{
-			path:    "dir",
-			checker: eq(nil),
+			path: "dir",
+			err:  nil,
 		},
 		{
-			path:    "link",
-			checker: eq(nil),
+			path: "link",
+			err:  nil,
 		},
 	}
 
@@ -495,8 +489,8 @@ func TestFolderCheckPath(t *testing.T) {
 			MarkerName: DefaultMarkerName,
 		}
 
-		if err := cfg.CheckPath(); !testcase.checker(err) {
-			t.Errorf("unexpected error in case %s: %s", testcase.path, err)
+		if err := cfg.CheckPath(); testcase.err != err {
+			t.Errorf("unexpected error in case %s: %s != %s", testcase.path, err, testcase.err)
 		}
 	}
 }
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -446,7 +446,7 @@ func TestFolderCheckPath(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	err = os.Symlink(filepath.Join(n, "dir"), filepath.Join(n, "link"))
+	err = os.Symlink(filepath.Join(n, "link"), filepath.Join(n, "dir"))
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -496,7 +496,7 @@ func TestFolderCheckPath(t *testing.T) {
 		}
 
 		if err := cfg.CheckPath(); !testcase.checker(err) {
-			t.Error("unexpected error %s", err)
+			t.Errorf("unexpected error in case %s: %s", testcase.path, err)
 		}
 	}
 }
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -451,12 +451,6 @@ func TestFolderCheckPath(t *testing.T) {
 		t.Fatal(err)
 	}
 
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
Also, interestingly:

Is it a regular: No
Is it a directory: No
Is it a symlink: Yes

lib/config/folderconfiguration.go
```diff
@@ -124,12 +124,20 @@ func (f *FolderConfiguration) CreateMarker() error {
...
-	if err != nil || !fi.IsDir() {
+	if err != nil {
+		l.Debugln("folder root:", err) // Log the error for debugging purposes
+		return errPathMissing
+	}
+
+	// Users might have the root directory as a symlink or reparse point.
+	// Furthermore, OneDrive bullcrap uses a magic reparse point to the cloudz...
+	if !fi.IsDir() && !fi.IsSymlink() {
 		return errPathMissing
```
calmh: Given that there is now a separate check I think it would be nice to have an `errPathNotDir` or something that is more descriptive than "missing", as users will not consider it missing.

Maybe even check if the first error `os.IsNotExist` and only then return `errPathMissing`, otherwise the error as is? In the name of getting a possibly more concrete and specific error in front of the user.

Oh, and how about a test for this? Just something that creates a directory, creates a symlink to it, and verifies that checkpath on the symlink succeeds. If goos=="windows" and the symlink create fails, t.Skip

lib/config/config_test.go
```diff
@@ -430,6 +431,76 @@ func TestFolderPath(t *testing.T) {
 	}
 }
 
+func TestFolderCheckPath(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("windows")
```
calmh: The test should run on Windows, this already worked afaik on other platforms. The build server runs in developer mode and we should be able to create symlinks just fine, so this test should pass. Not everywhere is the same of course which is why I suggest a t.Skip on error from os.Symlink

calmh: ... in theory. I tested it now, and while I can mklink just fine from a command prompt without any elevation I still get `symlink foo ~windows-symlink-test-8366981916958631613: A required privilege is not held by the client.` from tests which is annoying

calmh: Ah. Not only must symlinks be enabled for normal users, the *application* must also say "please let me create a symlink even though I'm just a normal process", using the SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE flag in the symlink call. That's so fucking stupid.

https://msdn.microsoft.com/en-us/library/windows/desktop/aa363866(v=vs.85).aspx

calmh: So I guess this is fine, once working, I filed the other one on Go. Maybe you want to contribute the fix there. :)

calmh:
With #4548 merged, this test can run on Windows as well.

imsodin:
@st-review lgtm

st-review:

calmh:
@st-review lgtm

st-review:
:ok_hand: Merged as 95a65bf0d0d7c54be83f2d894cda54f10508a1b2. Thanks, @AudriusButkevicius!

PR #4548, opened 2017-11-25
lib/osutil, lib/scanner: Run symlink test on Windows when possible

calmh:
### Purpose

Adds a shit-ugly function to create symlinks on Windows in tests only. The walk test now runs on Windows instead of being skipped (not visible in the code; but the build server now has one more passed test and one fewer skipped). Other tests could do the same.

### DIFF
--- a/lib/osutil/symlink_windows.go
+++ b/lib/osutil/symlink_windows.go
@@ -1,0 +1,137 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package osutil
+
+import (
+	"os"
+	"path/filepath"
+	"syscall"
+)
+
+// DebugSymlinkForTestsOnly is os.Symlink taken from the 1.9.2 stdlib,
+// hacked with the SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE flag to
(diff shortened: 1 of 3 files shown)


lib/osutil/symlink_windows.go
```diff
@@ -0,0 +1,137 @@
...
+func DebugSymlinkForTestsOnly(oldname, newname string) error {
+	// CreateSymbolicLink is not supported before Windows Vista
+	if syscall.LoadCreateSymbolicLink() != nil {
+		return &os.LinkError{"symlink", oldname, newname, syscall.EWINDOWS}
+	}
+
+	// '/' does not work in link's content
+	oldname = filepath.FromSlash(oldname)
+
+	// need the exact location of the oldname when its relative to determine if its a directory
```
imsodin: If you're really worried about the dictionary attacks: Its->it's

calmh: :D none of this code is written by us.

imsodin: Oh damn, I don't read the copyright stuff. But that's a relieve, because that functions is s**t xD

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 429b3a042938da3494215a509c7ce51a10b6ed4e. Thanks, @calmh!

Issue #4224, opened 2017-06-20
Editable default values for folders, devices


scienmind:
When new folder is created/added various advanced settings have a default values (rescan interval, pull order, versioning, etc.)

Making these defaults editable would help a lot to streamline deployments.

Usually I use a custom values for the above, but they are always consistent per machine:
For example:
On a file server: certain rescan rate, specific versioning scheme.
On a mobile clients: other rescan rate, other versioning scheme.

And for every folder added to any of the devices there are many clicks to set the corresponding parameters.
Allowing to "save current settings as default" (or something like that) will save a lot of time and improve UX.

(also, the back end for this will be useful once "auto-accept folders" is a reality)

AudriusButkevicius:
You should really use the rest API in my opinion. The defaults are there for common use cases. If your usecase is not common, you should use the rest api.

Ferroin:
Is there any practical reason that the defaults have to be hard-coded in Syncthing itself and can't be loaded from the config?  Yes, this isn't something that would be a super common case, but there are all kinds of perfectly legitimate reasons to want some way to specify defaults for a device in otherwise regular use cases.

AudriusButkevicius:
There isn't, but I'd argue that you already have a way to modify the defaults, that is by specifying a non-default value when creating the folder. Defaults currently exist in both Javascript land and Go land, so I question the implementation complexity of achieving this versus just calling the API with the values you actually want. Also, what happens if you created a folder with defaults and change the value of a default, are we going to have users expect the value on the folder to change too? I guess we'll have users in both camps and someone will always be left unhappy.

I vote close as wont-fix, as this makes things ambiguous, and just leads to unhappy people just in different ways.

@calmh ?

calmh:
Defaults need to live somewhere, for example we use them when creating the initial folder before there is a config, and when they are missing from the config itself - e.g., if someone hand added a folder with just the id and path into the config. 

I'm not absolutely opposed to having the GUI get the defaults as an object from the backend, and the backend reading them from the config, with the config having defaults itself if that section is missing.

But I also don't think most users spend to much time adding folders that this is a big deal either way.

zatricky:
I recently started moving over from resilio to syncthing - and find this to be an essential missing feature adding much frustration to the migration process.

Bear in mind I am synchronising 8 Devices with 18 Folders. 3 Devices don't have all the folders (mobile phone and storage-limited laptops for example):
2x servers (sites A+B)
3x desktops (sites A+C)
[...]

PR #4551, opened 2017-11-25
Autoaccept

AudriusButkevicius:
Fuck asynchronous config commits.
Spent 20 mins writing the feature, 4 hours debugging why the tests are flaky.

In the next rewrite, we should make config commits synchronous and just fucking be adults and deal with not holding locks when we mess around with the config. We could even tap into sync package, add `var IsAnyLockHeld bool`, and in replaceConfigLocked do ` if sync.IsAnyLockHeld { panic("reconsider your life") }`.


calmh:
:D 

Although... promises? D: (yeah I'll read the code before complaining too much)

lib/model/model.go
```diff
@@ -1486,7 +1541,7 @@ func (m *Model) AddConnection(conn connections.Connection, hello protocol.HelloR
 	conn.ClusterConfig(cm)
 
 	device, ok := m.cfg.Devices()[deviceID]
-	if ok && (device.Name == "" || m.cfg.Options().OverwriteRemoteDevNames) {
+	if ok && (device.Name == "" || m.cfg.Options().OverwriteRemoteDevNames) && hello.DeviceName != "" {
```
AudriusButkevicius: This is where 4 hours were spent essentially.

We setup the model in newState, AddConnection kicks off m.cfg.SetDevice(), we start poking at the model ourselves, adding folders, configs bla bla, and both newState and our poking kicks off a ton of routines which call CommitConfiguration in random fucking order, which leads to random fucking behaviour.

AudriusButkevicius: We could .Wait() on the SetDevice promise to fix this too, but the amount of places where in practice we should wait for stuff, is through the fucking roof.

calmh:
So maybe we should reconsider the configuration commit stuff even more than you've even done. As in remove all notifications and let each thing that is interested make sure to use live values all the time, being prepared for the fact that they may change from one access to the other. Or the opposite, where a configuration commit is a more invasive thing that stops and restarts all affected services, instead of them subscribing... Or something different.

AudriusButkevicius:
Well, in all logical places we already query live values, but for adding devices or folders it doesn't work, unless you poll, which is also shit.

I've added just enough band aid for me to be able to finish writing the feature/tests, as that was the ultimate goal.

If you are interested in refactoring this, we can discuss, but I am definately not re-engineer the world in this PR.

lib/model/model.go
```diff
@@ -1170,18 +1226,17 @@ func (m *Model) introduceDevice(device protocol.Device, introducerCfg config.Dev
 	m.cfg.SetDevice(newDeviceCfg)
 }
 
-func (m *Model) introduceDeviceToFolder(device protocol.Device, folder protocol.Folder, introducerCfg config.DeviceConfiguration) {
-	l.Infof("Sharing folder %s with %v (vouched for by introducer %v)", folder.Description(), device.ID, introducerCfg.DeviceID)
-
-	m.deviceFolders[device.ID] = append(m.deviceFolders[device.ID], folder.ID)
-	m.folderDevices.set(device.ID, folder.ID)
+func (m *Model) shareFolderWithDeviceLocked(deviceID protocol.DeviceID, folder protocol.Folder, introducer protocol.DeviceID) config.Promise {
```
imsodin: When you're changing the device: `folder` could also be just `folderID`.

lib/model/model.go
```diff
@@ -1142,6 +1152,52 @@ func (m *Model) handleDeintroductions(introducerCfg config.DeviceConfiguration,
 	return changed
 }
 
+// handleAutoAccepts handles adding and sharing folders for devices that have
+// AutoAcceptFolders set to true.
+func (m *Model) handleAutoAccepts(deviceCfg config.DeviceConfiguration, folder protocol.Folder) bool {
+	if _, ok := m.cfg.Folder(folder.ID); !ok {
+		defaultPath := m.cfg.Options().DefaultFolderPath
+		defaultPathFs := fs.NewFilesystem(fs.FilesystemTypeBasic, defaultPath)
+		for _, path := range []string{folder.Label, folder.ID} {
```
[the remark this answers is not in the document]
AudriusButkevicius: Then your preseeding thing goes out the window.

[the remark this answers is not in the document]
imsodin: A bit of self-contradiction doesn't hurt :P I was throwing that in just for discussion. I am not sure myself whether the benefits of allowing existing dirs (facilitating obscure workflows) outweighs the disadvantages of causing messed up folders.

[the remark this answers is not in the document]
AudriusButkevicius: I think let's start with this, if people complain, we can adjust.

AudriusButkevicius:
updated.

imsodin:
I didn't expect my "Waiter" idea to get instantly adopted, it may just be fine, but I should have made the following clear clear: "Waiter! was just a proposition - if "Promise" is a known programming standard, you should stick with that (I simply don't know). I am not saying I don't actually like "Waiter", I just think that we shouldn't create new terminology if there is a matching existing one.

Also repeating myself one last time, then I am going to stop: I would really appreciate successive commits, especially when the diff is lengthy like here (the main part is compact here, but still).

lib/model/model.go
```diff
@@ -1142,6 +1152,52 @@ func (m *Model) handleDeintroductions(introducerCfg config.DeviceConfiguration,
...
+			if _, err := defaultPathFs.Lstat(path); !fs.IsNotExist(err) {
+				continue
+			}
+
+			fcfg := config.NewFolderConfiguration(folder.ID, fs.FilesystemTypeBasic, filepath.Join(defaultPath, path))
+			fcfg.Label = folder.Label
+
+			// Need to wait for the promise, as this calls CommitConfiguration,
```
imsodin: Still "promise" plus all the `p` variables don't really make sense anymore with the waiter name change - see my more general comment about that: https://github.com/syncthing/syncthing/pull/4551#issuecomment-347137797

AudriusButkevicius:
This PR should ideally be merged unsquashed, as the commits represent two separate ideas, hence I keep force pushing rather than committing on top.

AudriusButkevicius:
Alternative would be to have two PR's but we're past that point now.

AudriusButkevicius:
What are we missing here?

lib/model/model.go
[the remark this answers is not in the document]
imsodin: Why would a single naming scheme like label_id make preseeding impossible? I still think a single and always descriptive directory name is better than either label (fine) or just id (not very helpful) - after all this path is pretty much set in stone afterwards.

calmh:
Missing time on my side, sorry. I want to look at this and don't have the time/energy now unfortunately. But, weekend coming up! \o/

AudriusButkevicius:
https://giphy.com/gifs/firstandmonday-bored-dude-waiting-5wWf7H0qoWaNnkZBucU

calmh:
So I've glanced at it and it seems sort of reasonable-ish. However; how come we need the waiter thing for this, and we did not for regular introducer action? Was that broken previously? Because this seems like the same thing just based on a different criterium?

AudriusButkevicius:
I think the comments explain this, essentially we add new folders from cluster config message, and cluster config starts poking at things that it expects to exist for the running folders, yet they are started in a separate configcomitter routine for the model, so we need to make sure folders are started before progressing with ClusterConfig.

AudriusButkevicius:
Adding a device doesn't immediately connect to it, and we didn't add folders before, just shared then with someone.

calmh:
This seems to work, but

- We should now expose the default path setting somewhere among the usual options, and the auto accept folders thing as a checkbox beside "introducer" on the device settings. This is good enough that we should not hide it...

- The GUI gets confused, perhaps as a result of some cached config or whatever. The folders that have been auto added are listed as "unshared", even though they are clearly shared with the other side and have even pulled a file:

screen shot 2017-12-05 at 13 46 30

This does not fix itself by reloading the GUI, although restarting Syncthing does fix it.

AudriusButkevicius:
Isn't this a bid advanced though, and should be controlled via advanced config?

calmh:
I have a hard time imagining something less advanced than a checkbox that then "just works". I see this being used on a server to allow new folders created on a phone or laptop to automatically be synced/backed up, or vice versa that every share added on a server type device automatically gets pushed to my laptop. No mind control required. :)

AudriusButkevicius:
So can't repro the UI bug, can repro a slightly different one, where there is no rescan interval:

image


Also, happened to restructure settings UI:

image
image
image
image

AudriusButkevicius:
Also, your UI bug must be browser cache related I suspect, as it clearly states stuff is shared in the Shared with section. Perhaps dump of the config would be nice?

scienmind:
> I can only see commercial conglomerates that control our minds and free will using this, not mortals.

I actually have previously had inquiries on "can't i just tell it where to dump all the shared folders?" and "can't it just always approve new shares from known devices?" from very much mortal peasants.
More so, I believe many "non-techy" users would appreciate being able to make ST "work with least involvement required".

So I agree with calmh, the "default path" and the "auto accept folders thing" should be in usual options for mortals

calmh:
OK I figured some things out and noticed some others.

- The "unshared" thing is because the folder is created without *our* device ID - a restart fixes that as part of the cleanup stage in the config loader. The "unshared" thing triggers on just one device being added to the folder, assuming it is always ourselves in that case.

- The auto accepted folder is created without a bunch of default values, as those typically gets set by the GUI... Rescan interval as you noted, normalisation, max conflicts, min disk free, etc. Ties into the above.

- Changing default folder path triggers "restart needed" dialog. Need to filter it out in the place in the model.

GUI looks fine to me.

scienmind:
> The auto accepted folder is created without a bunch of default values, as those typically gets set by the GUI... 

It will be very handy to be able to set/edit those (will also fix https://github.com/syncthing/syncthing/issues/4224 as well)

gui/default/assets/lang/lang-en.json
```diff
@@ -156,6 +161,8 @@
    "Override Changes": "Override Changes",
    "Path": "Path",
    "Path to the folder on the local computer. Will be created if it does not exist. The tilde character (~) can be used as a shortcut for": "Path to the folder on the local computer. Will be created if it does not exist. The tilde character (~) can be used as a shortcut for",
+   "Path where new auto accepted folders will be created, as well as the default suggested path when adding new folders via the UI.\n                Tilde character (~) expands to {%tilde%}.": "Path where new auto accepted folders will be created, as well as the default suggested path when adding new folders via the UI.\n                Tilde character (~) expands to {{tilde}}.",
```
imsodin: Doesn't seem to be used anywhere.

lib/connections/service.go
```diff
@@ -758,7 +759,7 @@ func dialParallel(deviceID protocol.DeviceID, dialTargets []dialTarget) (interna
 	for _, prio := range priorities {
 		tgts := dialTargetBuckets[prio]
 		res := make(chan internalConn, len(tgts))
-		wg := sync.NewWaitGroup()
+		wg := stdsync.WaitGroup{}
```
imsodin: What is this about?

AudriusButkevicius:
@scienmind go away

gui/default/assets/lang/lang-en.json
AudriusButkevicius: Sorry, this is prior reformatting.

lib/connections/service.go
AudriusButkevicius: our sync package is used for debugging long lasting locks/wg's. In this case the wg is expected to take time as it waits up to 10s to timeout.

AudriusButkevicius:
This is now neck deep into `Well Tested(TM)` territory.

lib/util/utils.go
```diff
@@ -70,6 +71,31 @@ func SetDefaults(data interface{}) error {
...
+		panic(fmt.Sprintf("non equal types: %s != %s", fromType, toType))
+	}
+
+	for i := 0; i < toStruct.NumField(); i++ {
+		fromField := fromStruct.Field(i)
+		toField := toStruct.Field(i)
+
+		tag := toType.Field(i).Tag
+
+		v := tag.Get("restart")
```
imsodin: Function argument `tag` isn't used, should probably be in place of "restart" (or removed entirely, as "restart" seems to be the only used tag/keyword).

AudriusButkevicius: surprised lint didn't flag that up.

lib/util/utils_test.go
```diff
@@ -169,3 +169,61 @@ func TestAddress(t *testing.T) {
...
+		},
+		CopyD: &Nested{
+			A: 3,
+		},
+		NoCopy: 4,
+	}
+
+	to := Test{
+		CopyA: 11,
+		CopyB: []string{"b*tch", "hoe"},
```
calmh: unsuitable example values

lib/model/model.go
```diff
@@ -2492,17 +2496,15 @@ func (m *Model) CommitConfiguration(from, to config.Configuration) bool {
...
-	from.Options.MaxRecvKbps = to.Options.MaxRecvKbps
-	from.Options.MaxSendKbps = to.Options.MaxSendKbps
-	from.Options.LimitBandwidthInLan = to.Options.LimitBandwidthInLan
-	from.Options.StunKeepaliveS = to.Options.StunKeepaliveS
-	from.Options.StunServers = to.Options.StunServers
+
+	// Copy fields that do not have the field set to true
+	util.CopyMatchingTag(&from.Options, &to.Options, "restart", func(v string) bool {
+		if len(v) > 0 && v != "true" {
+			panic(fmt.Sprintf(`unexpected struct value: %s. expected untagged or "true"`, v))
```
calmh: Why be so strict in these checks? A simple `return v != "true"` (and `return v == "true"` in the folder case above) should accomplish the same and not suddenly panic once someone decides to be extra explicit in the config struct for whatever reason

AudriusButkevicius: To prevent confusion, as in one place you want falses in the other you want trues, due to the imbalanced number of things.

calmh: Hmm OK. Just as long as this is exercised by one of the tests (I didn't check, all good if it already is) so it doesn't blow up later in production if someone does it wrong.

AudriusButkevicius: Well, we do save config while in the model and start and stop folders, so I am pretty sure it's covered.

calmh:
@st-review gui, lib/config, lib/model: Support auto-accepting folders (fixes #2299)

Also introduces a new Waiter interface for config changes and segments the configuration GUI.

st-review:
I'm sorry, @calmh. I'm afraid I don't know what you mean. I know how to `don't`, `lgtm`, `merge`, `prevent`, `rebuild`, `squash`, `stop`.

calmh:
@st-review merge

gui, lib/config, lib/model: Support auto-accepting folders (fixes #2299)

Also introduces a new Waiter interface for config changes and segments the configuration GUI.

st-review:
:ok_hand: Merged as 445c4edeca8f70bec8ae02a17db234ed06e88d62. Thanks, @AudriusButkevicius!

PR #4553, opened 2017-11-26
gui/default: Add sorting Devices by status (fixes #2636)

### LINKED ISSUE #2636: Sort order of devices by their status
It'd be nice to have an option to sort by the status of the device (such as Disconnected, Up to Date), instead of having it alphabetically.

Kropekk:
## Purpose

Add sorting Devices by status (fixes #2636)

### Testing
No tests - but the code is quite easy and follows standards.

### Screenshots

sortdevicesbynames
sortdevicesbystatus

### Documentation

The feature is self-explanatory. Maybe I could just add the order of sorting:

1. insync
2. syncing 
3. paused
4. disconnected
5. unused
6. unknown

### DIFF
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -10,6 +10,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
How does this look on a mobile/small screen?

uok:
My preference would be a simple order by connection status and without extra button: 
1) connected
2) disconnected 

If two devices are e.g. connected, order them alphabetically. Anything else is not important in my opinion. Too many sorting rules could also cause jumping entries during short syncs.

PR #4554, opened 2017-11-26
Add sorting Devices by status, second attempt (fixes #2636)

### LINKED ISSUE #2636: Sort order of devices by their status

Kropekk:
### Purpose
Add sorting Devices by status, second attempt (fixes #2636).
Also, I believe there was a bug in DEVICE_CONNECTED handler, hence line 248 (setting connected to true)
It's a second attempt: My previous approach was to add a button, which can be seen here: https://github.com/syncthing/syncthing/pull/4553

### Testing
No tests - but the code is quite easy and follows standards.

### Documentation
https://docs.syncthing.net/intro/gui.html
The sentence "The local device is always at the top, with remote devices in alphabetical order below." should be corrected. (I can try to do it after successful pull)

### DIFF
--- a/gui/default/syncthing/app.js
+++ b/gui/default/syncthing/app.js
@@ -36,18 +36,6 @@ syncthing.config(function ($httpProvider, $translateProvider, LocaleServiceProvi
 
 // @TODO: extract global level functions into separate service(s)
 
-function deviceCompare(a, b) {
-    if (typeof a.name !== 'undefined' && typeof b.name !== 'undefined') {
-        if (a.name < b.name)
-            return -1;
-        return a.name > b.name;
-    }
-    if (a.deviceID < b.deviceID) {
-        return -1;
-    }
-    return a.deviceID > b.deviceID;
-}
-
 function folderCompare(a, b) {
     var labelA = a.id;
     if (typeof a.label !== 'undefined' && a.label !== null && a.label.length > 0) {
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -222,11 +222,13 @@ angular.module('syncthing.core')
         $scope.$on(Events.LOCAL_INDEX_UPDATED, function (event, arg) {
             refreshFolderStats();
             refreshGlobalChanges();
+
         });
 
         $scope.$on(Events.DEVICE_DISCONNECTED, function (event, arg) {
             $scope.connections[arg.data.id].connected = false;
             refreshDeviceStats();
+            $scope.devices.sort(deviceCompare);
         });
 
         $scope.$on(Events.DEVICE_CONNECTED, function (event, arg) {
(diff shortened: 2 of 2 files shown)


imsodin:
You don't need to create a new pull request when doing a change, you can just push changes as additional commits to the branch where your PR lives on. This keeps the discussion intact, i.e. keeps previous and new comments in one place.

Personally I think we should either sort all just alphabetically or sort by *all* different status and only secondarily alphabetically. Sorting only by some statuses seems confusing and inconsistent to me. In any case I think "syncing" devices should get priority over "insync" devices, as they might need user attention.

uok:
Please keep in mind that most users only have **3 or less** devices ([see stats](https://data.syncthing.net/)) in their cluster!
My preference would be a simple order by connection status and without extra button: 
1) connected
2) disconnected 

If two devices have same status, order them alphabetically. Anything else is not important in my opinion. Too many sorting rules could also cause jumping entries during short syncs or disconnects.

imsodin:
I guess in the end this is a matter of taste and as I was reminded lately "De gustibus non est disputandum". That is kind of an argument for having this configurable with the sort button :P 

> If two devices have same status, order them alphabetically. Anything else is not important in my opinion. 

My problem with this is that disconnected and connected are not user facing device statuses (there is a "Disconnected" status, but "Paused" and "Unused" would also fall into that category too). To my feeling any sorting by "hidden" or non-obvious (alphabetically) factors is cause for confusion. Again, this is just my opinion.

Jumping only is a concern for "Up to Date" vs "Syncing", so could be prevented by just disabling sorting by status for these two.

> Please keep in mind that most users only have 3 or less devices (see stats) in their cluster!

What disadvantages would these users have due to sorting by (more) statuses?

uok:
"De gustibus non est disputandum" - latin for feature creep? 😄 
I see sorting as giving the user a quicker overview. If average is 3 devices, then sorting does not solve a problem and is simply not an usability gain that justifies an additional button I think. For me the existing sorting (alphabetically) and colored status text are enough, even on a machine with 10 devices (which puts it in 95 percentile of the stats)

calmh:
Here's my opinion, which is weakly held and could potentially change: we should not do sorting apart from the alphabetical one. My reasoning behind that is that the device sections are click targets. Moving them around when there are things like pause, edit and resume buttons that users will try to click is counter productive. Likewise that clicking pause on a connected device will re-sort the list and move away the device from where I just clicked on it.

(The last argument is weak because as it is we reload the list and close the device section I just clicked in, which is just as bad. We should not do that either.)

calmh:
I'm going to close this, and the corresponding issue, as I now think we should not do this. Sorry, @Kropekk, nothing wrong with your submission - especially as it resolves an issue that I have already considered valid once.

PR #4559, opened 2017-11-28
gui, lib/model: Display list of files needed by remote (fixes #4369)

### LINKED ISSUE #4369: GUI should indicate which files are out of sync on remote devices
Roughly similar UI to the local out of sync modal, per remote device. Helps solve 99% synced crap.

imsodin:
### Purpose

Issue and screenshots speak for themselves.
While doing this I noticed a little "logging bug" (i.e. no functionality affected): A folder was never correctly reported as paused, but as missing. I anyway put that check in a separate function and added a test.

### Testing

Manual testing -> see Screenshots (luckily an old unconnected phone provides some index garbage that shows up).  
The paused folder error got its own test.

### Screenshots

http://i.imgur.com/H92Oqx4.png
http://i.imgur.com/NgmzllD.png


gui/default/syncthing/core/syncthingController.js
```diff
@@ -1734,6 +1746,23 @@ angular.module('syncthing.core')
...
 
+        $scope.showRemoteNeed = function (device) {
+            $scope.remoteNeed = {};
+            $scope.remoteNeedFolders = [];
+            var folderIDCache = $scope.deviceFolders(device);
+            for (var i = 0; i < folderIDCache.length; i++) {
+                var pos = 0
+                if ($scope.completion[device.deviceID][folderIDCache[i]].needCount > 0) {
+                    getRemoteNeed(pos, device.deviceID, folderIDCache[i]);
+                    pos++;
```
imsodin: If anyone with any javascript skills can tell me how to more nicely keep folder order despite the fact that http.get is asynchronous, please do so.

AudriusButkevicius:
You should show mtime, size of files,last modified by just to make the UI less bland.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -476,29 +478,31 @@ angular.module('syncthing.core')
...
                 $scope.completion[device]._needBytes = needed
+                $scope.completion[device]._needCount = count;
             }
 
             if (needed == 0 && deletes > 0) {
                 // We don't need any data, but we have deletes that we need
                 // to do. Drop down the completion percentage to indicate
                 // that we have stuff to do.
                 $scope.completion[device]._total = 95;
-                $scope.completion[device]._needBytes = 0;
```
AudriusButkevicius: What's the rationale behind this change?

cmd/syncthing/gui.go
```diff
@@ -254,6 +255,7 @@ func (s *apiService) Serve() {
 	getRestMux.HandleFunc("/rest/db/file", s.getDBFile)                          // folder file
 	getRestMux.HandleFunc("/rest/db/ignores", s.getDBIgnores)                    // folder
 	getRestMux.HandleFunc("/rest/db/need", s.getDBNeed)                          // folder [perpage] [page]
+	getRestMux.HandleFunc("/rest/db/remoteneed", s.getDBRemoteNeed)              // device folder
```
AudriusButkevicius: Would it not make sense to paginate?

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius:
```
var promises = [];
for (var i = 0; i < folders.length, i++) {
    promises.push($http.get(url))
}
$q.all(promises).then(function (results) {
}).error(function (results) {});
```
https://docs.angularjs.org/api/ng/service/$q

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1734,6 +1746,23 @@ angular.module('syncthing.core')
             });
         };
 
+        $scope.showRemoteNeed = function (device) {
+            $scope.remoteNeed = {};
+            $scope.remoteNeedFolders = [];
+            var folderIDCache = $scope.deviceFolders(device);
```
AudriusButkevicius: Just `folders`, it's not really a cache as it's discarded.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1734,6 +1746,23 @@ angular.module('syncthing.core')
...
+            for (var i = 0; i < folderIDCache.length; i++) {
```
AudriusButkevicius:
```
angular.each(folders, function(folder) {
  if completion.needCount == 0: return...
})
```

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: To be honest, perhaps I'd contract the folder sections, and fetch as they get expanded.

gui/default/syncthing/transfer/remoteNeededFilesModalView.html
```diff
@@ -0,0 +1,20 @@
...
+      <button class="btn panel-heading" data-toggle="collapse" data-target="#remoteNeed-{{$index}}" aria-expanded="false">
+        <h4 class="panel-title">
+          <span>{{folderLabel(folder)}}</span>
+        </h4>
+      </button>
+      <div id="remoteNeed-{{$index}}" class="panel-collapse collapse">
+        <div class="panel-body">
+          <table class="table table-condensed table-striped">
+            <tr ng-repeat="file in remoteNeed[folder]">
+              <td>{{file.name}}</td>
```
AudriusButkevicius: I'd add pagination as this list can be in the millions. 
Out of sync dialog has most of that implemented.
We should also do it for failed items.

AudriusButkevicius: Also more data would be nice.

lib/model/model.go
```diff
@@ -2457,6 +2475,48 @@ func (m *Model) CommitConfiguration(from, to config.Configuration) bool {
...
+		}
+		return errFolderMissing
+	}
+	return nil
+}
+
+// checkFolderDeviceStatusLocked first checks the folder and then whether the
+// given device is connected and shares this folder.
+// Need to hold (read) lock on both m.fmut and m.pmut when calling this.
+func (m *Model) checkDeviceFolderConnectedLocked(folder string, device protocol.DeviceID) error {
```
AudriusButkevicius: Doesn't look like it's used?

lib/model/model.go
```diff
@@ -182,15 +183,11 @@ func (m *Model) StartFolder(folder string) {
 }
 
 func (m *Model) startFolderLocked(folder string) config.FolderType {
-	cfg, ok := m.folderCfgs[folder]
-	if !ok {
-		panic("cannot start nonexistent folder " + cfg.Description())
+	if err := m.checkFolderRunningLocked(folder); err == nil || err == errFolderMissing {
```
AudriusButkevicius: `nil.Error()` will panic itself with a random thing, so you want something better.

lib/model/model.go
```diff
@@ -1806,22 +1831,30 @@ func (m *Model) ScanFolder(folder string) error {
...
-
-	if !okRunner {
-		if okCfg && cfg.Paused {
-			return errFolderPaused
-		}
-		return errFolderMissing
+	m.fmut.RLock()
+	if err := m.checkFolderRunningLocked(folder); err != nil {
+		return err
+		m.fmut.RUnlock()
```
AudriusButkevicius: deadlock

lib/model/model.go
```diff
@@ -1806,22 +1831,30 @@ func (m *Model) ScanFolder(folder string) error {
...
+	m.fmut.RUnlock()
 
 	return runner.Scan(subs)
 }
 
 func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, subDirs []string) error {
+	m.fmut.RLock()
+	if err := m.checkFolderRunningLocked(folder); err != nil {
+		return err
+		m.fmut.RUnlock()
```
AudriusButkevicius: deadlock

lib/model/model.go
```diff
@@ -784,6 +785,30 @@ func (m *Model) NeedFolderFiles(folder string, page, perpage int) ([]db.FileInfo
...
+// RemoteNeedFolderFiles returns paginated list of currently needed files in
+// progress, queued, and to be queued on next puller iteration, as well as the
+// total number of files currently needed.
+func (m *Model) RemoteNeedFolderFiles(device protocol.DeviceID, folder string) ([]db.FileInfoTruncated, error) {
+	m.pmut.RLock()
+	m.fmut.RLock()
+	if err := m.checkFolderRunningLocked(folder); err != nil {
+		return nil, err
+		m.fmut.RUnlock()
+		m.pmut.RUnlock()
```
AudriusButkevicius: mega deadlock.
I don't think you need pmut at this point either.

AudriusButkevicius: I am surprised lint doesn't flag this up as dead code.

lib/model/model.go
```diff
@@ -784,6 +785,30 @@ func (m *Model) NeedFolderFiles(folder string, page, perpage int) ([]db.FileInfo
...
+	}
+	rf := m.folderFiles[folder]
+	m.fmut.RUnlock()
```
AudriusButkevicius: pmut never unlocked.

gui/default/syncthing/transfer/remoteNeededFilesModalView.html
imsodin: I was considering either grouping and hiding by folders (as done now) or using pagination, but then only sorted by folder. I preferred grouping without pagination, as with pagination it would be difficult to see changes by folder. I'll see whether I find a solution for both. And I will add some more info.

gui/default/syncthing/core/syncthingController.js
imsodin: That it is pointless because the condition and the assignment some lines above already make sure that `_needBytes == needed == 0`.

gui/default/syncthing/transfer/remoteNeededFilesModalView.html
AudriusButkevicius: You can have pagination within the accordion I guess?

imsodin:
Ok, so now there is both an accordion and pagination:
screenshot_2017-11-29_21-42-50
And no, the dates aren't a UI problem, the files are really like this (coming from an Android device that wasn't online for ages - so nothing to care about).

AudriusButkevicius:
We should probably not show size for directories, or atleast somehow indicate that it's a directory.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -56,6 +56,16 @@ angular.module('syncthing.core')
...
+        resetRemoteNeed();
+        // $scope.remoteNeed = [];
+        // $scope.remoteNeedFolders = [];
+        // $scope.remoteNeedCurrentPage = [];
+        // $scope.remoteNeedPageSize = [];
+        // // $scope.remoteNeed = {};
+        // // $scope.remoteNeedFolders = [];
+        // // $scope.remoteNeedCurrentPage = {};
+        // // $scope.remoteNeedPageSize = {};
+        // $scope.remoteNeedDevice = '';
```
AudriusButkevicius: Not needed.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -638,6 +658,18 @@ angular.module('syncthing.core')
...
 
+        function resetRemoteNeed() {
+            $scope.remoteNeed = [];
+            $scope.remoteNeedFolders = [];
+            // $scope.remoteNeedCurrentPage = [];
+            // $scope.remoteNeedPageSize = [];
+            // $scope.remoteNeed = {};
+            // $scope.remoteNeedFolders = [];
+            // $scope.remoteNeedCurrentPage = {};
+            // $scope.remoteNeedPageSize = {};
```
AudriusButkevicius: ... not needed.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1734,6 +1769,33 @@ angular.module('syncthing.core')
             });
         };
 
+        $scope.showRemoteNeed = function (device) {
+            var promises = [];
+            $scope.remoteNeedDevice = device.deviceID;
+            var i = 0;
+            $scope.deviceFolders(device).forEach(function(folder) {
```
AudriusButkevicius: I think forEach has a hack, where by if you do `function(folder, idx)` it should give you the index.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1734,6 +1769,33 @@ angular.module('syncthing.core')
             });
         };
 
+        $scope.showRemoteNeed = function (device) {
+            var promises = [];
+            $scope.remoteNeedDevice = device.deviceID;
+            var i = 0;
```
AudriusButkevicius: Put the vars next to each other.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1734,6 +1769,33 @@ angular.module('syncthing.core')
...
+                }
+                $scope.remoteNeedFolders.push(folder);
+                promises.push(getRemoteNeedPromise(1, 10, i));
+                i++;
+            });
+            $q.all(promises).then(function (results) {
+                results.forEach(function(res) {
+                    $scope.remoteNeed.push(res.data);
+                    // // Using dicts because updates will be asynchronous
+                    // $scope.remoteNeed[$scope.remoteNeedFolders[i]] = res.data;
```
AudriusButkevicius: not needed

gui/default/syncthing/transfer/remoteNeededFilesModalView.html
```diff
@@ -0,0 +1,41 @@
...
+            <thead>
+              <tr>
+                <th translate>Name</th>
+                <th translate>Size</th>
+                <th translate>Mod. Time</th>
+                <th translate>Mod. Device</th>
+              </tr>
+            </thead>
+            <tr dir-paginate="file in remoteNeed[$index].files | itemsPerPage: remoteNeed[$index].perpage" current-page="remoteNeed[$index].page" total-items="remoteNeed[$index].total" pagination-id="'remoteNeed-' + $index">
+                <!-- <tr ng-repeat="file in remoteNeed[$index]"> -->
```
AudriusButkevicius: Don't need this?

gui/default/syncthing/transfer/remoteNeededFilesModalView.html
```diff
@@ -0,0 +1,41 @@
...
+      </button>
+      <div id="remoteNeed-{{$index}}" class="panel-collapse collapse">
+        <div class="panel-body">
+          <table class="table table-striped table-dynamic">
+            <thead>
+              <tr>
+                <th translate>Name</th>
+                <th translate>Size</th>
+                <th translate>Mod. Time</th>
+                <th translate>Mod. Device</th>
```
AudriusButkevicius: Last modification, potentially add tooltips explaining in full what it is?

lib/model/model.go
```diff
@@ -784,6 +787,40 @@ func (m *Model) NeedFolderFiles(folder string, page, perpage int) ([]db.FileInfo
 	return progress, queued, rest, total
 }
 
+// RemoteNeedFolderFiles returns paginated list of currently needed files in
+// progress, queued, and to be queued on next puller iteration, as well as the
+// total number of files currently needed.
+func (m *Model) RemoteNeedFolderFiles(device protocol.DeviceID, folder string, page, perpage int) ([]db.FileInfoTruncated, int, error) {
+	m.pmut.RLock()
+	m.fmut.RLock()
```
AudriusButkevicius: So this might be correct, but you should double check that the order these are acquired is the same as somewhere else.

lib/model/model.go
```diff
@@ -784,6 +787,40 @@ func (m *Model) NeedFolderFiles(folder string, page, perpage int) ([]db.FileInfo
...
+			skip--
+			return true
+		}
+		if get > 0 {
+			files = append(files, f.(db.FileInfoTruncated))
+		}
+		return true
+	})
+
+	return files, total, nil
```
AudriusButkevicius: Don't you have the total in the needCount from completion?
In which case you could decrement get every time you append a file, and `return get > 0` to stop iteration when you've got enough.

I guess if this list of out of sync keeps changing, pagination has the nice affect of readjusting the pages count every time you switch the page, yet I suspect this might be sluggish for large amounts of files.

imsodin:
The tooltip causes this weird column shift, no idea why:
screenshot_2017-11-30_01-21-12

gui/default/syncthing/core/syncthingController.js
imsodin: Yes, but I don't need the index of all that I iterate over, only of those that aren't complete.

uok:
@imsodin, Can you please explain what the "Mod. Device" column displays? I expect it to always show the same device name after clicking on "remote out of sync items"

AudriusButkevicius:
Tooltip did not cause that, you have a mismatching column count due to html changes.

AudriusButkevicius:
Yet I can't immediately see the issue.

imsodin:
Tooltips are at least responsible in the sense that without them the take is just fine, the weirdness only appears when howering the header and the tooltip appears.

@uok the tooltip describes it like this:

"Device which has done the latest modification"

is this clear?

uok:
In my opinion since this "out of sync items" info is only visible during sync, the columns "mod. time" and "mod. device" are not really helping the user (as you cannot really influence them except hit "pause" but then maybe it is already "too late"). For me seeing a) which file b) size of file is out of sync is more than enough info to decide if I can turn off  (e.g. most important files are done) my computer or not. This would also free up screen space to display longer file names and paths.

Other small suggestions:
+) Put the device name into the purple header like "Devicename - out of sync items"
+) drop the "remote" wording, because I think it is clear that if you click on a device you get "out of sync items" for this device.

AudriusButkevicius:
Apart from the broken html, this looks fine to me.

imsodin:
Html is unbroken:
screenshot_2017-12-01_13-13-07
Tooltips shouldn't be on `<td>` but the text inside it.

I added the devicename in the title instead of "remote". I left the additional information, but I don't care much either way: They aren't really needed when it is a "normal" sync, but when something is out of the order, they could be helpful. E.g. in my case, the "problem" (constant syncing) can be solved by removing this obsolete device (or bringing it online, but I can't do that). And the values should wrap, so the full path is visible regardless of size.

Please read the strings, I tend to produce not overly easy understandable explanations in general.

AudriusButkevicius:
@st-review lgtm

st-review:

AudriusButkevicius:
@st-review lgtm

st-review:

lib/model/model.go
```diff
@@ -585,6 +584,7 @@ func (m *Model) FolderStatistics() map[string]stats.FolderStatistics {
 type FolderCompletion struct {
 	CompletionPct float64
 	NeedBytes     int64
+	NeedCount     int64
```
calmh: What is Count counting? Can we be more specific here (and everywhere it flows)

gui/default/assets/lang/lang-en.json
```diff
@@ -257,6 +261,7 @@
    "This is a major version upgrade.": "This is a major version upgrade.",
    "This setting controls the free space required on the home (i.e., index database) disk.": "This setting controls the free space required on the home (i.e., index database) disk.",
    "Time": "Time",
+   "Time when this item has been last modified": "Time when this item has been last modified",
```
calmh: This string, and the one about "device which has done the latest modification" both sound like broken or at least non-idiomatic English. Though I can't off hand pinpoint issue - spider sense tingling. Anyone native speaking here? (sorry if that's you @imsodin and I'm barking up the wrong tree)

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1729,11 +1748,35 @@ angular.module('syncthing.core')
...
+                }
+                $scope.remoteNeedFolders.push(folder);
+                promises.push(getRemoteNeedPromise(1, 10, i));
+                i++;
+            });
+            $q.all(promises).then(function (results) {
+                results.forEach(function(res) {
+                    $scope.remoteNeed.push(res.data);
+                });
+                $('#remoteNeed').modal().on('hidden.bs.modal', function () {
```
calmh: How long does it take to load the remote need thing for large databases or slow devices? Given we don't show the modal until everything has resolved this might leave the user clicking then having absolutely nothing happen for seconds and seconds?

lib/model/model.go
```diff
@@ -715,15 +718,15 @@ func (m *Model) NeedSize(folder string) db.Counts {
 // NeedFolderFiles returns paginated list of currently needed files in
 // progress, queued, and to be queued on next puller iteration, as well as the
 // total number of files currently needed.
-func (m *Model) NeedFolderFiles(folder string, page, perpage int) ([]db.FileInfoTruncated, []db.FileInfoTruncated, []db.FileInfoTruncated, int) {
+func (m *Model) NeedFolderFiles(folder string, page, perpage int) ([]db.FileInfoTruncated, []db.FileInfoTruncated, []db.FileInfoTruncated) {
 	m.fmut.RLock()
 	defer m.fmut.RUnlock()
 
 	total := 0
```
calmh: Remove total here as well?

lib/model/model.go
```diff
@@ -2457,6 +2486,48 @@ func (m *Model) CommitConfiguration(from, to config.Configuration) bool {
 	return true
 }
 
+// checkFolderRunningLocked returns nil if the folder is up and running and a
+// descriptive error if not.
+// Need to hold (read) lock on m.fmut when calling this.
+func (m *Model) checkFolderRunningLocked(folder string) error {
+	_, ok := m.folderRunners[folder]
+	if !ok {
```
calmh: could reverse pretty much all of these conditions and get a flatter indentation with early returns

gui/default/assets/lang/lang-en.json
imsodin: I am Swiss - no trace of native English here :)

lib/model/model.go
imsodin: NeedItemCount? That's not much more specific, but I mean it is anything we track - files, dirs and symlinks.

calmh: `NeedItems` then I think is fine. We have "items" as a word for files-or-folders in a couple of places already, but just a count can be anything.

gui/default/syncthing/core/syncthingController.js
imsodin: Could be a moment indeed - now the modal is shown immediately displaying a simple "loading..." string and data as soon as the rest call for the first folder returns.

gui/default/assets/lang/lang-en.json
imsodin: I tried to come up with alternatives, but they are all worse than this - so unless suddenly someone with an English language intuition turns up, we may have to go with this?

AudriusButkevicius: Doesn't trigger my spider senses.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1908,4 +1939,8 @@ angular.module('syncthing.core')
                 window.localStorage["metricRates"] = $scope.metricRates;
             } catch (exception) { }
         }
+
+        $scope.sizeof = function (dict) {
```
AudriusButkevicius: sizeOf

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1729,11 +1747,24 @@ angular.module('syncthing.core')
...
+            $scope.deviceFolders(device).forEach(function(folder) {
+                if ($scope.completion[device.deviceID][folder].needItems === 0) {
+                    return;
+                }
+                $scope.remoteNeedFolders.push(folder);
+                $scope.refreshRemoteNeed(folder, 1, 10);
+            });
+            $('#remoteNeed').modal().on('hidden.bs.modal', function () {
+                resetRemoteNeed()
+            });
```
AudriusButkevicius: This will screw up if you close the dialog before the request returns. Not sure how to do it right. Move reset early?

imsodin: A solution might be to wrap [saving the return value from the request](https://github.com/syncthing/syncthing/pull/4559/files/131d3daa64ceae10263348eff8eb5a4b26f85c4d#diff-10f45c616bfdd371810aecdd203d0f36R673) inside a `if ($scope.remoteNeedDevice !== '')` thus no new data will be added after the modal is closed. That is probably racy though, so additionally resetting early to catch that is probably still a good idea. I don't like to exclusively reset early to not keep that data pointlessly in memory.

gui/default/assets/lang/lang-en.json
calmh: "Last modification time", "Time the item was last modified" sound better to me.

(We can bike shed the wording a bit because it sort of gets set in stone due to translations)

imsodin: The first one sounds wrongish to me, but the second is nicer (shorter) than mine and google agrees with your tense ("was last modified" is much more common than "has been last modified" - very scientific decision method). So that's settled then :)

gui/default/assets/lang/lang-en.json
```diff
@@ -47,12 +47,14 @@
...
    "Creating ignore patterns, overwriting an existing file at {%path%}.": "Creating ignore patterns, overwriting an existing file at {{path}}.",
    "Danger!": "Danger!",
+   "Data is loading...": "Data is loading...",
    "Deleted": "Deleted",
    "Device": "Device",
    "Device \"{%name%}\" ({%device%} at {%address%}) wants to connect. Add new device?": "Device \"{{name}}\" ({{device}} at {{address}}) wants to connect. Add new device?",
    "Device ID": "Device ID",
    "Device Identification": "Device Identification",
    "Device Name": "Device Name",
+   "Device which has done the latest modification": "Device which has done the latest modification",
```
calmh: "Device that made the last modification"?

gui/default/assets/lang/lang-en.json
```diff
@@ -47,12 +47,14 @@
    "Copyright © 2014-2017 the following Contributors:": "Copyright © 2014-2017 the following Contributors:",
    "Creating ignore patterns, overwriting an existing file at {%path%}.": "Creating ignore patterns, overwriting an existing file at {{path}}.",
    "Danger!": "Danger!",
+   "Data is loading...": "Data is loading...",
```
AudriusButkevicius: Data is loading?

How is the data loading itself?

Loading data.... ?

gui/default/assets/lang/lang-en.json
imsodin: "Device on which the last modification was done"? Or closer to yours "Device where the last modification originated?"  
And "originated"/"was done"/"has happened" could be used interchangeably. Originated is the most elegant, but as you say maybe technical. Has happened sounds so passive (it wasn't happenstance but the user - hopefully). So was done seems most appropriate, but also ugly.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1623,17 +1646,14 @@ angular.module('syncthing.core')
...
+                for (var i = 0; i < folder.devices.length; i++) {
+                    if (folder.devices[i].deviceID === deviceCfg.deviceID) {
+                        folders.push(folder.id);
                         break;
                     }
                 }
-            }
-
-            folders.sort(folderCompare);
+            });
```
AudriusButkevicius: Why sort gone?

imsodin: Because it doesn't work: `folders` at this point is a list of IDs, not folderconfigs -> not sortable by label. The loop above is over `folderList()`, which is already sorted, so the result is sorted too.

AudriusButkevicius: ah ok.

gui/default/syncthing/core/syncthingController.js
imsodin: As this isn't marked as "outdated" I addressed this by only adding data to the object if the modal is still open and resetting information at the beginning as well just for good measure.

gui/default/assets/lang/lang-en.json
```diff
@@ -59,6 +59,7 @@
    "Device ID": "Device ID",
    "Device Identification": "Device Identification",
    "Device Name": "Device Name",
+   "Device where the last modification was done": "Device where the last modification was done",
```
imsodin: @calmh How about this?

uok: I still find the two modification columns confusing and somehow out of place in the out of sync modal.  If you really think they add value for the user I would (not native speaker) suggest "last modified by this device", "last modifying device" or "device that lastly modified this file"

calmh: Agree with Ben, except s/lastly/last/. I think it's the past perfect tense (?) in "was done" instead of something more direct like "did" that sounds more German than English to me (as a swede, mind). :)

imsodin: Well, that's again closer to my first proposal, which @calmh didn't like (and I agree in hindsight), because the device didn't modify anything, the user did...  
Also at least the which device last modified were useful in one scenario: One device that isn't in use anymore got some index data to one device which then show up as out of sync, Without the information about the device, I would have never known why they are even there. I agree this is a very remote scenario, but "debugging" usually involves weird scenarios which you didn't think of before, so I believe it is good to stay on the verbose side of things.

calmh:
OK barring the language stuff is this good to go?

imsodin:
Yes

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1729,11 +1749,25 @@ angular.module('syncthing.core')
             $('#needed').modal().on('hidden.bs.modal', function () {
                 $scope.neededFolder = undefined;
                 $scope.needed = undefined;
-                $scope.neededTotal = 0;
                 $scope.neededCurrentPage = 1;
             });
         };
 
+        $scope.showRemoteNeed = function (device) {
+            resetRemoteNeed()
```
AudriusButkevicius: Semicolon.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -656,6 +665,20 @@ angular.module('syncthing.core')
...
+            var url = urlbase + '/db/remoteneed?device=' + $scope.remoteNeedDevice.deviceID;
+            url += '&folder=' + encodeURIComponent(folder);
+            url += "&page=" + page + "&perpage=" + perpage;
+            $http.get(url).success(function (data) {
+                if ($scope.remoteNeedDevice !== '') {
+                    $scope.remoteNeed[folder] = data;
+                }
+            }).error(function (err) {
+                $scope.remoteNeed[folder] = undefined;
+                $scope.emitHTTPError(err)
```
AudriusButkevicius: Semicolon

cmd/syncthing/gui.go
```diff
@@ -731,20 +731,52 @@ func (s *apiService) getDBNeed(w http.ResponseWriter, r *http.Request) {
...
+	deviceID, err := protocol.DeviceIDFromString(device)
+	if err != nil {
+		http.Error(w, err.Error(), 500)
+		return
+	}
+
+	page, perpage := getPagingParams(qs)
+
+	if files, err := s.model.RemoteNeedFolderFiles(deviceID, folder, page, perpage); err != nil {
+		http.Error(w, err.Error(), http.StatusNotFound)
```
AudriusButkevicius: We don't know what the underlying problem was, so 404 might be confusing.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1729,11 +1749,25 @@ angular.module('syncthing.core')
...
+            $scope.remoteNeedDevice = device;
+            $scope.deviceFolders(device).forEach(function(folder) {
+                if ($scope.completion[device.deviceID][folder].needItems === 0) {
+                    return;
+                }
+                $scope.remoteNeedFolders.push(folder);
+                $scope.refreshRemoteNeed(folder, 1, 10);
+            });
+            $('#remoteNeed').modal().on('hidden.bs.modal', function () {
+                resetRemoteNeed()
```
AudriusButkevicius: Semicolon

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1908,4 +1942,8 @@ angular.module('syncthing.core')
                 window.localStorage["metricRates"] = $scope.metricRates;
             } catch (exception) { }
         }
+
+        $scope.sizeOf = function (dict) {
+            return Object.keys(dict).length;
+        }
```
AudriusButkevicius: I'd say semicolon, but we should ger a js linter.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -656,6 +665,20 @@ angular.module('syncthing.core')
...
+            url += "&page=" + page + "&perpage=" + perpage;
+            $http.get(url).success(function (data) {
+                if ($scope.remoteNeedDevice !== '') {
+                    $scope.remoteNeed[folder] = data;
+                }
+            }).error(function (err) {
+                $scope.remoteNeed[folder] = undefined;
+                $scope.emitHTTPError(err)
+            });
+        }
```
AudriusButkevicius: Semicolon I guess, we have it after most scope functions.

cmd/syncthing/gui.go
imsodin: I don't know about http error code conventions, that was an guess based off wikipedia. We do know the source of the error: It can only be an error regarding the given folder or device (nonexistent, paused, not connected, ...) and that will be explained by `err.Error()`. So the client can do something about it, that's why I

gui/default/assets/lang/lang-en.json
uok: Another suggestion: simply use "device" and "time" as table header - much easier for translators because I think not everybody gets the abbreviation if he does not see the code. Optional tooltips e.g. "device that last modified this file" - optional because I think we should only include them if we get a torrent of support requests regarding these columns.

If we leave both columns in place my only request would be to have "time" column before "device" column, thanks.

calmh:
Feel free to merge whenever you guys have sorted out any remaining style or language issues (column headers maybe as discussed above), I'm cool with it.

AudriusButkevicius:
This already has my lgtm

gui/default/assets/lang/lang-en.json
imsodin: @uok @calmh @AudriusButkevicius 
I am honestly slightly annoyed due to many different proposals but no clear consensus. I really don't mind much either way, I'd just like to get to a decision (which could also be do whatever you want, but that's also a decision). I will use the following unless there is clear opposition (as in I get overruled by authority/majority or my previous arguments are addressed and refuted :P ):

A. Should the "Mod." be dropped?  
I agree having the abbreviation is a bit ugly, "Modification" is just very long, so I considered it the lesser evil. I am against dropping this, as it introduces ambiguity (e.g. time when sync will finish/started, time created, ...). Yes if you actually think about it, such alternatives don't make much sense, but assuming users think upfront is a leap of faith.  
I will keep "Mod." as it is currently.

B. What tooltip to use:
**"Device that last modified this file"** is perfectly fine with me, I assume @calmh is now also ok with the slight discrepancy in the meaning between the device vs the user making the change.

@uok I will swap the time and device columns. I did not bring up dropping the columns alltogether as they were originally proposed by @AudriusButkevicius (which I blatantly interpret as approval) and I am for having them. And I don't understand the idea of dropping the the tooltips until there is confusion instead of preventing that upfront: What is the problems with tooltips (except finding consensus on their wording xD )?

calmh:
I'll add mine so it merges when the column stuff is sorted then :)

gui/default/assets/lang/lang-en.json
uok: I only stumbled upon "mod." as a word when thinking about translations as I'm not sure all translators are programmers. Let's got with what you suggested, thanks.

calmh:
@st-review lgtm

st-review:
:ok_hand: Merged as 8fa2b7765ab59cba24996cb0a084dcf1315c25f8. Thanks, @imsodin!

PR #4563, opened 2017-11-30
gui/default: Add confirmation dialog when removing folder (fixes #4543)

### LINKED ISSUE #4543: Confirmation on Remove Folder Button
Three times already I've deleted a folder by accident trying to click on Advanced settings. It takes hours for the reindex, which really is a pain. So I want to suggest adding a simple "Are you sure? Yes/No" confirmation on the folder remove button, as is pretty much an industry standard.

qepasa:
### Purpose

Hello everyone. Just as stated in the issue, I added dialog window  in GUI when removing folder.
I added some text which is not translated already so that definitively could be a point of concern.

### Testing

Changes are pretty straightforward and to be honest I don't know how I could test a thing like that
besides simple manual tests.

### Screenshots

https://i.imgur.com/6HYusgB.png


uok:
Thanks for the PR! A few comments on GUI (I have not tested functionality):
+) Header "Remove folder" is better than "Removing folder"
+) It would be great to read "folder ABC" (actual folder name) instead of "this folder"
+) Maybe buttons "yes" (with trash icon) and "cancel" (with X icon)
+) Translations are done via Transifex (you don't need to edit translation files)

imsodin:
As @uok said, you don't need to handle translations manually. Just run `go run build.go translate` after you made changes involving strings that need translations.

gui/default/syncthing/core/removeFolderDialogView.html
```diff
@@ -0,0 +1,15 @@
+<modal  id="remove-confirmation" status="danger" icon="exclamation-circle" heading="{{'Remove folder' | translate}}" large="no" closeable="yes">
+    <div class="modal-body">
+        <p ng-model="currentFolder.label" style=" overflow : hidden; text-overflow: ellipsis;">
+            <span translate>Are you sure you want to remove folder</span> {{currentFolder.label}}?
```
AudriusButkevicius: This should probably be done with variable substitution, as not in all languages the folder name can go on the end of the sentence.

gui/default/index.html
```diff
@@ -725,8 +725,9 @@ <h4 class="panel-title">
   <ng-include src="'syncthing/core/majorUpgradeModalView.html'"></ng-include>
   <ng-include src="'syncthing/core/aboutModalView.html'"></ng-include>
   <ng-include src="'syncthing/core/discoveryFailuresModalView.html'"></ng-include>
+  <ng-include src="'syncthing/core/removeFolderDialogView.html'"></ng-include>
 
-  <!-- vendor scripts -->
+<!-- vendor scripts -->
```
imsodin: This indentation is actually intended.

gui/default/syncthing/core/removeFolderDialogView.html
```diff
@@ -0,0 +1,15 @@
+<modal  id="remove-confirmation" status="danger" icon="exclamation-circle" heading="{{'Remove folder' | translate}}" large="no" closeable="yes">
```
imsodin: We usually capitalize titles, so "Remove Folders"

qepasa:
Thanks for your comments. The dialog hasn't changed much after those changes but still, here's an example with long folder label in case there is something that still needs to be changed: https://i.imgur.com/ipIsNMo.png

imsodin:
Thanks for the screenshot, it still made me notice something: The different symbols in the folder edit modal and this modal for the same operation (remove folder) seems wrong to me. In my opinion it should both times be the minus symbol thing or both times the trashcan - I don't really have a preference either way. The difference in color seems fine, as in the edit modal it isn't definitive (less severe, yellow) while in the confirmation it is (red).

And if I were to be very nitpicky (though this is deep into wishland, definitely not required for this PR): The modal could be lower, e.g. with it's lower border right above the control buttons (Save etc) of the folder edit modal or central over the edit folder modal. That would require less mouse travel to confirm. Again this is just a suggestion, please feel free to ignore it if you don't feel like messing with this.

AudriusButkevicius:
Also add yourself to AUTHORS and NICKS files.

uok:
@qepasa 
I agree with @imsodin about the symbol, the minus symbol was better (sorry about that, I forgot that removing the folder does not actually delete files in which case the trash can symbol would be a better fit)

calmh:
(just AUTHORS and run `go run script/authors.go`, or leave it to me at merge time)

calmh:
@st-review lgtm (needs commit subject tweak)

st-review:

AudriusButkevicius:
so looks good to me too.

qepasa:
I just noticed that there is no confirmation dialog when removing a device. Should I add it as well? 
@imsodin To be honest I'm not all that familiar with front end and I don't really know how to do positioning in a proper way (first guess would be something like add top: xx% to this dialog style)

AudriusButkevicius:
Adding the same for devices makes sense

calmh:
Yeah along those lines the remove button is yellow color (warning-ish) not red (danger!) and I think the dialog should follow.

qepasa:
I added dialog when removing device (screenshot 2) and moved html files for those dialog views to corresponding folders as I thought it makes more sense, if not I'll put them back into core/

Moreover when fiddling around and testing those dialogs I noticed that there are some problems with text wrapping in Edit Device view (sreenshot 2 - Share Folders With Device checkbox list) and remote device name (screenshot 3). I don't know how relevant these issues are. I tried to fix the one with checkbox list by adding usual styles (text-overflow and overflow) but it made tooltips disappear for items in the list. 

Screenshots:
https://imgur.com/sWe5wNg (added message when removing folder)
https://imgur.com/yXrXSNQ
https://imgur.com/eQnyj7f

calmh:
@st-review merge it

gui: Add confirmation on removing devices and folders (fixes #4543)

st-review:
:ok_hand: Merged as 205426a9c6bb3b60a06f4aece95fd0562cd9910e. Thanks, @qepasa!

PR #4568, opened 2017-12-04
lib/model: Mark files below newly created symlinks as deleted (fixes #2571)

### LINKED ISSUE #2571: Replacing a directory with a symlink or vice versa is buggy

> ## 1
> 
> A:
> a
> B:
> a b

> ## 2
> 
>   A:
>   a b
>   B:
>   a b

> ## 3
> 
> Two:~$ ls -R 
> A:
> a
> B:
> a b

> ## 4
> 
> Two:~$ rm -r A; ls B
> a b

> ## 5
> 
> One:~$ ls -R
> A:
> b
> B:
> b
> 
> Two:~$ ls -R
> A:
> b
> B:
> b

> ## 2bis
> 
>   A:
>   a b
>   B:
>   a b
> ## 3bis
> 
> Two:~$ ls -R 
> A:
> a b
> B:
> a b

imsodin:
### Purpose

The scenario is well described in #2571, in short: If a dir is replaced by a symlink, the files within the dir aren't marked as deleted in the database if they exist at the location the new symlink points to. The other device will therefore fail to remove the directory because it isn't empty.  
The issue further describes that when you manually delete the directory, contents from the symlink target gets deleted: This isn't the case even previous to this PR, as we check that we don't descend into symlinks.

### Testing

New unit test for this scenario.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1944,8 +1944,6 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 	if err := runner.CheckHealth(); err != nil {
 		l.Debugln("Stopping scan of folder %s due to: %s", folderCfg.Description(), err)
 		return err
-	} else if len(batch) > 0 {
-		m.updateLocalsFromScanning(folder, batch)
 	}
 
 	if len(subDirs) == 0 {
@@ -1956,8 +1954,6 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 
 	// Do a scan of the database for each prefix, to check for deleted and
 	// ignored files.
-	batch = batch[:0]
-	batchSizeBytes = 0
 	for _, sub := range subDirs {
 		var iterError error
 
@@ -1986,13 +1982,17 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 				// The file is valid and not deleted. Lets check if it's
 				// still here.
 
-				if _, err := mtimefs.Lstat(f.Name); err != nil {
+				if _, err := mtimefs.Lstat(f.Name); err != nil || osutil.TraversesSymlink(mtimefs, filepath.Dir(f.Name)) != nil {
 					// We don't specifically verify that the error is
 					// fs.IsNotExist because there is a corner case when a
 					// directory is suddenly transformed into a file. When that
 					// happens, files that were in the directory (that is now a
 					// file) are deleted but will return a confusing error ("not a
 					// directory") when we try to Lstat() them.
+					// If a parent directory was replaced by a symlink, the
+					// Lstat may succeed but anything below that symlink
+					// should still be considered deleted, as we never descend
+					// into symlinks.
 
 					nf := protocol.FileInfo{
 						Name:       f.Name,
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
This looks mostly ok, but I guess it's still racey if we download more than one file at a time?
Also other changes are a bit confusing why they are happening.

imsodin:
The other change is explained in the commit message, I forgot to repeat it in the PR:
If an index exchange is forced after scanning and before looking for deleted files, the other device will have the problem described in the issue being fixed, and only after the second index update comes in, it will be resolved. By not sending the index update in between those two stages, this temporary problem is avoided.

lib/model/model.go
```diff
@@ -1944,8 +1944,6 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 	if err := runner.CheckHealth(); err != nil {
 		l.Debugln("Stopping scan of folder %s due to: %s", folderCfg.Description(), err)
 		return err
-	} else if len(batch) > 0 {
```
calmh: I don't see the point of removing this. There is still the batch sending in the loop above, this is just to handle the last, incomplete batch.

lib/model/model.go
```diff
@@ -1956,8 +1954,6 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 
 	// Do a scan of the database for each prefix, to check for deleted and
 	// ignored files.
-	batch = batch[:0]
```
calmh: These should really be inside the if, above (and not removed, imho).

lib/model/model.go
imsodin: This is for small batches, e.g. the example case of replacing a directory with a symlink/file: Without this the new symlink/file will be transmitted in a single batch, then the database is walked and the deletion below that sent in another batch. In this time, the remote device will have the problem that it should delete a directory, but can't because there is valid content.  
Also I didn't see the point of having this in the first place: Why send twice before reaching the max batch size or end of scanning?

imsodin:
>Another question. This makes us do the symlink traversal check for every file. That check tests each component up to the target. That means we recheck a lot of stuff a lot of times, essentially making the scan O(d*n) where d is the tree depth instead of O(n).

>Is this a problem? Can we do something about that?

We could move this entire cleanup stage to the scanner, where we already know both the new and old file (i.e. replace dir with file/symlink -> delete stuff below), so no need to walk at all. Another approach would be to make the scanner pass a pair of files (replaced and new), such that we can do the cleanup without database traversal and lstating in the model.

lib/model/model.go
calmh: OK, makes sense.

calmh:
The scanner doesn't have database access other than via the CurrentFiler interface, so it doesn't know what files should be marked deleted...

I'm thinking that since this is anyway racy per definition and sort of best effort, we might get away with a traverses symlink cache. That is, the symlink traversal checker has an in memory cache, created at the beginning of the walk, marking entries as checked. Each path can then be checked from the end until we reach a level that has already been decided one way or the other. That makes the leaf checks on average just one stat plus one cache hit for the directory, instead of umpteen stats.

imsodin:
That certainly fixes the lstat performance issue.  
What is your opinion on the second suggested approach to pass both new and old file from scanner to model? This has the additional benefit that child removals can be added to the batch before the parent dir removal, thus even for changesets exceeding the batch size there won't be any intermediate failed state.

imsodin:
@calmh I just came across a related but more serious issue: #4573. This PR will depend on how that issue is resolved, so we should focus on that and put this PR on ice for now.

PR #4576, opened 2017-12-06
lib/fs: Prolong test timeout on darwin, hopefully fixing flakyness

imsodin:
Teamcity reports long durations when the flaky test passes on darwin, so extending the timeout seems to have a good chance of minimizing flakyness - lets see (and hope).

### DIFF
--- a/lib/fs/basicfs_watch_test.go
+++ b/lib/fs/basicfs_watch_test.go
@@ -218,7 +218,11 @@ func testScenario(t *testing.T, name string, testCase func(), expectedEvents []E
 
 	go testWatchOutput(t, name, eventChan, expectedEvents, allowOthers, ctx, cancel)
 
-	timeout := time.NewTimer(2 * time.Second)
+	timeoutDuration := 2 * time.Second
+	if runtime.GOOS == "darwin" {
+		timeoutDuration *= 2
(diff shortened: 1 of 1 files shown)


imsodin:
@st-review merge

st-review:
@imsodin: Build status is `pending`. I'll wait until it goes green and then merge!

AudriusButkevicius:
I wonder if the tests could be refactored not to rely on timeouts?

st-review:
:ok_hand: Merged as c005b8dcb0724a9b7f4cd337037f2deb5d6e297d. Thanks, @imsodin!

calmh:
I suspect the issue is not Darwin as such either, but that the Darwin "build server" is an laptop in my garage that also occasionally does other stuff. Whereas the windows and linux builders are mostly dedicated vm:s.

imsodin:
Mentioning "darwin" is just shorter than "slow darwin build server" - that's all there is to it.

And these test don't "rely" on timeout, they just give up after a certain amount of time, such that the "global" timeout isn't triggered and other tests can run as well. It would have probably been better to extend the timeout time regardless of OS, as it doesn't matter any when the test doesn't fail.

PR #4584, opened 2017-12-10
lib/model, lib/scanner: More precise deletion detection (fixes #2571, fixes #4573)

### LINKED ISSUE #2571: Replacing a directory with a symlink or vice versa is buggy

### LINKED ISSUE #4573: Inaccessible files during scan are marked as deleted
Repro:

1. Create `dir/file` and make sure they are synced between devices A and B.
2. Change ownership/permission for `dir` on device A such that the user syncthing is running as is no longer able to access `dir` (e.g. `sudo chown root:root dir && sudo chmod 700 dir`).
3. Hit the rescan button on device A and wait for "changes" to propagate to B.

Result: `dir/file` gets deleted from device B.

Expected: Some kind of error or at least "nothing" (see #4480), but certainly not deletion of the file on the remote.

Origin of bug:  
Lstat returning any error is considered equal to file having been deleted: https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L1989

Possible solutions:  
Maybe this can be handled with stricter checking of the error returned by Lstat. If we do this, I believe we need to add `IsNotExist(err) bool` to the filesystem interface to make sure all filesystem implementations are compatible with this check.  
The following idea doesn't actually work (we also detect "ordinary" deletes this way).<strike>It might be both more consistent and performant to drop these lstat checks and replace them by comparing the new file info with the one it has replaced. Thus cases like replacing a dir with a file can still be detected and handled, while the case described here wont result in data loss.</strike>  
Another possibility would be to implement #4480 and skip the current lstat tests on paths that failed to be scanned.

imsodin:
### Purpose

Fixes #2571 and #4573 by extending the file info returned from a scan by a "ReplacedDir" parameter, which signifies that the old file info in the db was a directory (and now is a file/symlink). This information is used to detect deleted children, for which "IsNotExist" returns false.  
In addition this makes sure that these deleted children are updated in the db before their parent is replaced, thus preventing intermediate pull (dir not empty) failures when the entire update exceeds the batch size.  
Compared to #4568 (which is obsolete if going forward with the approach of this PR) this uses information from scanning instead of doing more lstat calls.

### Testing

Two unit tests specific to the fixed issues and `TestInternalScan`. The last one is there because all tests passed with an earlier version of this PR, while it was severely broken - this test would have failed.


AudriusButkevicius:
I think ScanResult should be:

```
type ScanResult struct {
  Old fs.FileInfo
  New fs.FileInfo
}
```

And then whatever checks we want to do can be done in the model.

We can then add utility functions

```
func (r *ScanResult) TypeChanged() {
   return r.Old.IsDir() != r.New.IsDir() || r.Old.IsRegular() != r.New....
}
```

AudriusButkevicius:
This Old/New stuff would also help detect renames better, as we could defer submitting stuff that doesn't have Old until we go traversing deletes in the model.

AudriusButkevicius:
Or even better, scanner could perhaps do the WithHave walk itself and return Scan result where either New or Old is a nil pointer.

imsodin:
> Or even better, scanner could perhaps do the WithHave walk itself and return Scan result where either New or Old is a nil pointer.

That sounds like a good idea, gets rid of some duplicate db lookups. I don't understand though why not to return results for replacements, but only for creations or deletions? I will see whether I get that to work this way with reasonable effort.  
To detect renames for a future protocol extension you anyway need to go over all (file) infos again after scanning/walking - you can never know whether something is part of a rename before you know about all creations and deletions (unless I am missing something?).

It should probably have been obvious that chmoding to 0000 isn't possible on Windows:
https://build.syncthing.net/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=13971&_focus=823
https://github.com/syncthing/syncthing/pull/4584/files#diff-9accdb1b2ab06368238d103a1f87b3c6R2742
Is there anything similar to make a file inaccessible for Windows or should I just skip this on Windows?

AudriusButkevicius:
I was working on something similar, and it's impossible to make files inaccessible on windows it seems.
You can set them to write only (not sure if that works), yet most of our functions do some sort of osutil.GODMODE=ON and chmods everything along the way the way it needs.

I am not suggesting not to return results for replacements, I am only saying replacement can be expressed as a pair of *New *Old.

To better detect renames,  we could reuse @calmh's index sorter spiller thingy, and do this:
1. If we walk and find something that was modified but existed before, we can send it to the model straight away, it's unlikely to be a rename.
3. If we find something new that did not exist, we scan it and store that in this temporary disk spilling map, essentially mapping `add + first block hash` to the FileInfo.
3. If we find something that was deleted, we store it in `del + first block hash` in the old file info in the disk spilling map.

After scan is finished, we go through every delete in this temp spill db (delete first as we want those to go out first in the index messages I guess), and try to match a delete with new addition, and which point we return a scan result where Old.Name doesn't match New.Name, which implies rename.
Then we make sure these two are commited to our db one after another (Add then Delete),
and in the index sender we make sure we include all deletes that come after an addition, regardless of the batch size.

We could ramp this up a notch and use this tempdb spiller to map every file (even existing ones being modified) by their hash to catch cases such as A and B files exist, and someone does mv A B.

This removes the nice feature of starting to send stuff for others to download as soon as things are discovered, but I think not redownloading stuff in general is better plus existence of inotify should make this less tragic (yet might need to be changed to accommodate the scanner changes).

calmh:
Sounds reasonable (although possibly quite a lot of work). We might want to make an exception for newly added folders, for some definition of newly added (flag of some kind when we actually finish a scan?). When adding a folder it's a win if we start sending out stuff a bit earlier and avoid the whole scan-to-one-database-then-copy-it-to-the-real-index work.

I think with the scheme above we only need to send add + the directly following delete (ordered by sequence, but we already do that) in the same batch, not all deletes? The latter could theoretically exceed the max message size, although we could check specifically for that as well.

imsodin:
Agree, that sounds nice. As far as I understand I can still go ahead with this PR/fixing these issues without any renaming stuff, this came just up because of `ScanResult` which should return both new and old file infos, correct? Then I guess we should do further discussion of design details about renaming which do not have direct implications on this PR in a dedicated issue to separate concerns and preserve the information.

AudriusButkevicius:
> I think with the scheme above we only need to send add + the directly following delete (ordered by sequence, but we already do that) in the same batch, not all deletes? The latter could theoretically exceed the max message size, although we could check specifically for that as well.

I am suggesting to include the next delete, not all of them.

AudriusButkevicius:
@imsodin so my suggestion was just to make ScanResult have a *Old *New to avoid db lookups, and for potential extension. I am not anticipating to rewrite the scanner in this PR.

imsodin:
@AudriusButkevicius Now walking the filesystem and db happens in parallel and the old file info is returned in `ScanResult`. As a new file info is always required, I embedded the new file info:

```go
type ScanResult struct {
	protocol.FileInfo
	Old *protocol.FileInfo
}
```

This approach relies on both db and fs walking happening ordered alphabetically by path. The fs walk was not ordered before (but the comment said it should). Also the operations done in the walkfs aren't logged when debugging is enabled.

lib/scanner/walk.go
```diff
@@ -299,67 +409,71 @@ func (w *walker) walkRegular(ctx context.Context, relPath string, info fs.FileIn
...
 	//  - was a directory previously (not a file or something else)
 	//  - was not a symlink (since it's a directory now)
 	//  - was not invalid (since it looks valid now)
-	cf, ok := w.CurrentFiler.CurrentFile(relPath)
-	permUnchanged := w.IgnorePerms || !cf.HasPermissionBits() || PermsEqual(cf.Permissions, uint32(info.Mode()))
-	if ok && permUnchanged && !cf.IsDeleted() && cf.IsDirectory() && !cf.IsSymlink() && !cf.IsInvalid() {
+	permUnchanged := cf != nil && (w.IgnorePerms || !cf.HasPermissionBits() || PermsEqual(cf.Permissions, uint32(info.Mode())))
+	if cf != nil && permUnchanged && !cf.IsDeleted() && cf.IsDirectory() && !cf.IsSymlink() && !cf.IsInvalid() {
 		return nil
 	}
```
AudriusButkevicius: All of this cf stuff could be guarded by a single if.

lib/scanner/walk.go
```diff
@@ -248,43 +293,108 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
+			}
+			for haveChanOpen && strings.HasPrefix(currDBFile.Name, path+string(fs.PathSeparator)) {
+				if !currDBFile.IsInvalid() {
+					finishedChan <- ScanResult{
+						FileInfo: currDBFile.InvalidatedCopy(w.ShortID),
+						Old:      currDBFile,
+					}
+				}
+				currDBFile, haveChanOpen = <-w.HaveChan
+			}
```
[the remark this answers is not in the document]
AudriusButkevicius: To be honest, perhaps it's usefull to change to to two channels, each walking it's own thing, which then pulls a value from each, if matches, does stuff, if not, handles addition/removal and  tugs off the other channel. atleast it would be less confusing.

lib/scanner/walk.go
```diff
@@ -71,16 +72,18 @@ type Config struct {
 	UseWeakHashes bool
 }
 
-type CurrentFiler interface {
-	// CurrentFile returns the file as seen at last scan.
-	CurrentFile(name string) (protocol.FileInfo, bool)
+type ScanResult struct {
+	protocol.FileInfo
```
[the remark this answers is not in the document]
imsodin: I added the explicit `New`, but I still do the deletedCopy in scanner. My reasoning is, that in the scanner package it is already known that it is deleted, while in the model I need to add `if r.New == nil` handling. Did you have any specific purpose in mind for setting it to nil?

lib/scanner/walk.go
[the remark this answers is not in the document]
imsodin: I don't immediately see how this would be any different from what is done right now: New path is coming from fs walk, then db file infos are received and handled until they catch up with the fs path, then magic happens including handling child db items if necessary. How would that be different with two channels? (I am not saying the current state isn't confusing)

lib/model/model.go
```diff
@@ -1945,13 +1935,39 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
...
+	subPrefixes := subDirs
+	if len(subPrefixes) == 0 {
+		subPrefixes = []string{""}
+	}
+	haveChan := make(chan *protocol.FileInfo)
+	dbCtx, dbCancel := context.WithCancel(ctx)
+	go func() {
+		for _, sub := range subPrefixes {
+			select {
+			case <-dbCtx.Done():
```
AudriusButkevicius: wonder if evaluating Done() once doesn't make more sense.

lib/fs/filesystem.go
```diff
@@ -170,7 +170,8 @@ func NewFilesystem(fsType FilesystemType, uri string) Filesystem {
 	if l.ShouldDebug("filesystem") {
 		fs = &logFilesystem{fs}
 	}
-	return fs
+
+	return NewWalkFilesystem(fs)
```
AudriusButkevicius: This will have an undesired side effect of logging tons of stuff in logFs, yet not a killer.

lib/model/model.go
```diff
@@ -1945,13 +1935,39 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
...
+	}()
+
+	rchan, err := scanner.Walk(ctx, scanner.Config{
 		Folder:                folderCfg.ID,
 		Subs:                  subDirs,
 		Matcher:               ignores,
 		BlockSize:             protocol.BlockSize,
 		TempLifetime:          time.Duration(m.cfg.Options().KeepTemporariesH) * time.Hour,
-		CurrentFiler:          cFiler{m, folder},
+		HaveChan:              haveChan,
```
AudriusButkevicius: Perhaps it's better to pass the FileSet and just let the scanner deal with getting it whatever it needs?
This way both iteration loops (subdirs here, and subdirs in the scanner) are in the same place.

lib/scanner/walk.go
```diff
@@ -201,35 +207,58 @@ func (w *walker) walk(ctx context.Context) (chan protocol.FileInfo, error) {
...
 	return func(path string, info fs.FileInfo, err error) error {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
 		default:
 		}
 
+		if path == "." {
+			return err
+		}
```
AudriusButkevicius: I think this should be handled by fs.IsInternal()

lib/fs/filesystem.go
imsodin: I agree it is very noisy and I will happily undo that change, I just wanted to bring it up for discussion. It took me a while to figure out that nothing is working because `fs.Walk` isn't ordered, and I first had to figure out that the walkfs doesn't produce any debug logging to get there.

lib/scanner/walk.go
imsodin: I don't think it does, it only checks for files matching `/.st**` as far as I see.

lib/model/model.go
imsodin: I would assume that this is the same after compiling, but yeah, why not.

lib/scanner/walk.go
AudriusButkevicius: Right, I'm thinking about my modified version in encryoted fs.

lib/model/model.go
imsodin: To me It looked like it was deliberate that the scanner package does not use db at all, so I kept it that way. If there is nothing to it, I agree that it would be better to have both iterations in one place.

calmh: As long as it takes a reasonable interface it and isn't a direct dependency on the database that's probably fine either way.

lib/model/model.go
AudriusButkevicius: I don't think it is, as the compiler can't know that done doesn't have side-effects, perhaps it increments some internal counter, in which case it can't evaluate once.

lib/scanner/walk.go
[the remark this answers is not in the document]
AudriusButkevicius: I would have more confidence in this working, debugging and extending this, as the lockstepping is now peppered throughout the file in multiple places.

AudriusButkevicius:
Idk, github sucks, my comments about locktatepping are gone as far as I can see.

imsodin:
@AudriusButkevicius I separated walking the filesystem tree and processing the results from that and the db - I hope this is now more approachable/less scary (?).

lib/model/model.go
```diff
@@ -1984,98 +2003,43 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
...
-		if len(batch) == maxBatchSizeFiles || batchSizeBytes > maxBatchSizeBytes {
-			if err := runner.CheckHealth(); err != nil {
+	delDirStack := make([]protocol.FileInfo, 0)
+	for r := range rchan {
+		if err := checkBatch(); err != nil {
+			l.Debugln("Stopping scan of folder %s due to: %s", folderCfg.Description(), err)
+			return err
+		}
+
+		for len(delDirStack) != 0 && !strings.HasPrefix(r.New.Name, delDirStack[len(delDirStack)-1].Name+string(fs.PathSeparator)) {
```
calmh: Comment on what's being accomplished here would be appreciated. It's not clear to me at first glance. Also the `delDirStack[len(delDirStack)-1]` thing could have a name as it's repeated a lot.

lib/model/model.go
```diff
@@ -1984,98 +2003,43 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 		}
 	}()
 
-	for f := range fchan {
-		if len(batch) == maxBatchSizeFiles || batchSizeBytes > maxBatchSizeBytes {
-			if err := runner.CheckHealth(); err != nil {
+	delDirStack := make([]protocol.FileInfo, 0)
```
calmh: This could probably be `var delDirStack []protocol.FileInfo`

lib/scanner/walk.go
```diff
@@ -98,30 +111,61 @@ type walker struct {
...
-	toHashChan := make(chan protocol.FileInfo)
-	finishedChan := make(chan protocol.FileInfo)
+	fsChan := make(chan fsWalkResult)
+
+	haveChan := make(chan *protocol.FileInfo)
+	haveCtx, haveCancel := context.WithCancel(ctx)
+
+	// A routine which walks the db and returns file infos to be compared
+	// to scan results.
+	go func() {
```
calmh: I think this, and its peer below, would do nicely as actual named methods?

(And we can probably defer the channel close / context cancel in those functions.)

lib/scanner/walk.go
```diff
@@ -98,30 +111,61 @@ type walker struct {
...
+		if len(w.Subs) == 0 {
+			w.Have.Walk("", haveCtx, haveChan)
+		} else {
+			haveCtxChan := haveCtx.Done()
```
calmh: I don't think this loop is tight enough that it's worth temp-variabling the context channel, but it's a matter of taste.

lib/scanner/walk.go
```diff
@@ -218,18 +256,23 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
 		}
 
-		if err != nil {
-			l.Debugln("error:", path, info, err)
-			return skip
-		}
-
-		if path == "." {
-			return nil
+		if err == nil {
```
calmh: This stuff could just be below the `if err != nil` below?

lib/scanner/walk.go
```diff
@@ -247,44 +290,186 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
-			l.Debugln("ignored (patterns):", path)
-			return skip
-		}
-
 		if !utf8.ValidString(path) {
+			select {
+			case fsChan <- fsWalkResult{
+				path: path,
+				info: nil,
+				err:  errors.New("Path isn't a valid utf8 string"),
```
calmh: Don't capitalize errors (here and other places)

lib/scanner/walk.go
```diff
@@ -247,44 +290,186 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
+			case <-ctx.Done():
+				return ctx.Err()
+			}
 			l.Warnf("File name %q is not in UTF8 encoding; skipping.", path)
 			return skip
 		}
 
 		path, shouldSkip := w.normalizePath(path)
 		if shouldSkip {
+			select {
```
calmh: This little select is repeated often enough that it would be a nice `func walkError(ctx context.Context, dst chan WalkResult, path string, err error)` I think

lib/scanner/walk.go
```diff
@@ -247,44 +290,186 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
+		if info.IsSymlink() && info.IsDir() {
+			l.Debugln("ignored (symlinked directory):", path)
+			return skip
+		}
+
+		return err
+	}
+}
+
+func (w *walker) processWalkResults(ctx context.Context, fsChan <-chan fsWalkResult, haveChan <-chan *protocol.FileInfo, toHashChan, finishedChan chan<- ScanResult) {
```
calmh: (I stopped reviewing here, leaving bookmark for myself)

lib/scanner/walk.go
imsodin: Inside this there `err` can become non-nil in 
```
info, err = w.Filesystem.Lstat(path)
```

This was present before this PR however the fs walker already does a Lstat. I assumed this has a point that I just don't see, but if there isn't, I can remove it and which will clean this section up a bit (tests pass without this additional LStat).

imsodin:
@calmh @AudriusButkevicius Any chance you can look at this again? #4616 turned out to be hairier than I first thought and building on top of this PR (in particular on having access to new and old file infos in the model) would be my course of choice  - however I would prefer the (non-nitpicky part of) review of this PR to be over before using it for further work.

calmh:
There's a lot to wrap my mind around and I'm not sure I understand it all yet, but from an architecture/high-level/non-nitpicky point of view it looks fine to me.

AudriusButkevicius:
I have to go through it in anger once I am next to a desktop.

AudriusButkevicius:
So architecture wise what is being done here makes sense, and definately improves things.
I guess I still have comments around the implementation itself, as in, it's still complicated enough that it's hard to grasp weather there are bugs or not, so I'd still expect comments asking to restructure things etc.

If you want to have those comments before we progressing, let me know, as I didn't dig deep yet.

imsodin:
Getting a "this is at least roughly the way to go" is good enough for me. I am actually in not too much a hurry, I just had some line of thoughts that all got entangled in this PR, so I wanted to know where we stand on it.

imsodin:
@AudriusButkevicius @calmh Gentle reminder that this still needs review. I was able to reduce the main part of the scanner logic, making it quite a bit more readable (that's after [calmh's review bookmark](https://github.com/syncthing/syncthing/pull/4584#discussion_r158608884) ;) ). If you don't have time now that's obviously fine, I will just want to make sure it is not forgotten.

calmh:
I'm back from being away skiing for a week and will dive into it again. There's always a higher threshold for reviewing stuff where I actually need to *think* and not just look and react...

AudriusButkevicius:
So I think this is ok, in terms of what it does as it addresses a problem, but I really start thinking that scan should append everything to some sort of on disk database, which we then have a sort method for which deals with all of these things.

Perhaps that's a issue for another day.

imsodin:
I don't really understand what "these things" are in "sort method for which deals with all of these things" - I assume you are thinking about renames? Other than that I don't see a reason to store anything on disk, as memory usage is restricted by batch size (except for big files, but that's another issue).

Also beside the issue I address in this PR, I also have a branch on top of this, with solutions to two more issues (send-only out-of-sync for metadata only changes and ignoring/unignoring files properly). Admittedly part of it could be done without this PR, but not all of it. And I intend to do displaying scan failure in GUI and retry failed dirs on top of this eventually as well. So while I believe this PR has merit enough on its own, there is also more.

lib/scanner/walk.go
```diff
@@ -70,16 +72,28 @@ type Config struct {
...
 
-func Walk(ctx context.Context, cfg Config) chan protocol.FileInfo {
+type fsWalkResult struct {
+	path string
+	info fs.FileInfo
+	err  error
+}
+
+type ScanResult struct {
+	New *protocol.FileInfo
```
calmh: Having pointers to FileInfos here makes my spider sense tingle that we're doing something clever with updating them and having another pointer somewhere else that sees the change. I don't think that's the case right? If not I think we should go with just the structs here. There's an argument to be made that the struct is growing and that copying it everwhere is not super efficient, but it's not huge and we do that everywhere else...

Same comment on the chan types where you send FileInfos.

lib/scanner/walk.go
```diff
@@ -238,48 +273,177 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
+		case fsChan <- fsWalkResult{
+			path: path,
+			info: info,
+			err:  nil,
+		}:
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+
+		// under no circumstances shall we descend into a symlink
```
calmh: I'm 100% sure this is mythical and can be removed

lib/scanner/walk.go
AudriusButkevicius: The idea is that delete is new = nil, old = something and new discovery is vica versa.

lib/scanner/walk.go
imsodin: Let me recheck this (sorry, I am a bit cloudy atm): This case already existed pre-PR. Descending into symlinked dirs should already be prevented in `fs.Walk`, so I agree this check is redundant. I understood it as extra security for a messy topic (currently I am thinking about the platform dependent behaviour of `IsDir` or the whole reparse point BS) and making it explicit, that we skip symlinks. Should this be removed (possibly adding a comment about skipping symlinks)?

lib/scanner/walk.go
imsodin: Actually, I did it a little bit different: `New` is never `nil`, but always a valid file info just as before (i.e. for a deleted file, the deleted flag is set, version increased, blocks wiped, ...).  Only if no file existed previously `Old` is `nil`. So I could make these structs instead of pointers and have a "default" struct in that case and `Old` is identifiable as "didn't exist before" by having an empty version.

imsodin:
@calmh I unpointerised the file infos and introduced an ugly-named [(f FileInfo) NonExistent() bool](https://github.com/syncthing/syncthing/pull/4584/files#diff-7ee4177f0c9c90ed45b403044de4d983R120) function to check whether a file info is just the default instance and thus representing "no file" instead of nil testing.

I would love to have this in v0.14.45, because there were two reports about non-empty dir removals and they might be due to the issues that are fixed in this PR.

lib/scanner/walk.go
```diff
@@ -238,48 +273,177 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
+			New: f.DeletedCopy(w.ShortID),
+			Old: f,
+		}:
+		case <-done:
+		}
 	}
 }
 
-func (w *walker) walkRegular(ctx context.Context, relPath string, info fs.FileInfo, fchan chan protocol.FileInfo) error {
+func (w *walker) checkIgnored(f protocol.FileInfo, finishedChan chan<- ScanResult, done <-chan struct{}) bool {
```
calmh: `checkIgnoredAndInvalidate` for symmetry with above?

lib/protocol/bep_extensions.go
```diff
@@ -117,13 +117,34 @@ func (f FileInfo) WinsConflict(other FileInfo) bool {
 	return f.Version.Compare(other.Version) == ConcurrentGreater
 }
 
+func (f FileInfo) NonExistent() bool {
```
calmh: I'd prefer this to be named something else for clarity, maybe `IsZero` or `IsEmpty`. Saying `NonExistent` feels like it might return false if the FileInfo represents a file that doesn't exist on disk or something like that.

AudriusButkevicius: Also, why not just compare it to protocol.FileInfo{}? We could add an Equal method can compare with that?

calmh: An `Equals()` would be fairly involved what with all the slices and stuff, and we don't actually need one so far. I think the *implementation* here is fine, but the naming is misleading.

calmh:
Ready to merge from my POV but validating the resulting commit message sucks from a phone and I’m not sure if the details are up to date. Imsodin, merge with something reasonable?

imsodin:
Sure, will do. One question regarding the use of prefixes: scanner and model are definitely the most affected packages, but db and protocol are also changed a bit. Does an "all" (or just "lib") tag make more sense in such a case?

To stay within 80 chars I will probably call the commit just "Refactor scanning" and give some detail in the body of the commit message.

calmh:
Yeah go with “all” as it affects things all over. But it’s subjective.

PR #4591, opened 2017-12-12
cmd/syncthing: Fix /rest/system/browse for folder path completion (fixes #4590)

### LINKED ISSUE #4590: Path completion in new folder dialog does not work anymore
Previously directories matching the current input were displayed in a dropdown box below the entry field when creating a new folder in the web UI. This does no longer work in general (it does e.g. work when entering just "/").

calmh:
Two issues since the filesystem migration;

- filepath.Base() doesn't do what the code expected when the path ends
  in a slash, and we make sure the path ends in a slash.

- the return should be fully qualified, not just the last component.

With this change it works as before, and passes the new test for it.


imsodin:
@st-review lgtm

st-review:

calmh:
@st-review merge

st-review:
:ok_hand: Merged as a4147d90191bedd5d69415eafb08c7c83caea926. Thanks, @calmh!

PR #4594, opened 2017-12-12
lib/model: Don't leak fd when truncate fails (fixes #4593)

### LINKED ISSUE #4593: Fd leak when temp file cannot be truncated
If the truncate call fails after we've opened a temp file we will return with an error, leaving the temp file fd open.

calmh:
Also attempt to handle this nicer by ignoring the truncate failure when
it doesn't matter, and recover by deleting the temp file when it does.

### DIFF
--- a/lib/model/sharedpullerstate.go
+++ b/lib/model/sharedpullerstate.go
@@ -167,8 +167,8 @@ func (s *sharedPullerState) tempFile() (io.WriterAt, error) {
 				// corner case when the old file is larger than the new one
 				// and we can't just overwrite blocks and let the old data
 				// linger at the end. In this case we attempt a delete of
-				// the file and hope for best luck next time, when we should
-				// come around with s.reused == 0.
+				// the file and hope for better luck next time, when we
+				// should come around with s.reused == 0.
 
 				fd.Close()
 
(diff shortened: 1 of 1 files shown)


lib/model/sharedpullerstate.go
```diff
@@ -158,8 +158,27 @@ func (s *sharedPullerState) tempFile() (io.WriterAt, error) {
...
+				// should come around with s.reused == 0.
+
+				fd.Close()
+
+				if err := s.fs.Remove(s.tempName); err != nil {
+					l.Debugln("failed to remove temporary file:", err)
+				}
+
+				s.failLocked("dst truncate", err)
+				return nil, err
```
AudriusButkevicius: This is confusing, it seems like the err from the if would shadow this, but it probably doesn't

imsodin: Maybe name the error in the "removal if" something like `remErr` to make it obvious.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as c7522063b35c3e0656d335375d4f19dbeb9a82b3. Thanks, @calmh!

Issue #4586, opened 2017-12-10
Versioner file tagging is inconsistent


AudriusButkevicius:
Simple uses `file.ModTime()`, staggered uses `time.Now()`.

calmh:
I think this is intentional. Simple just keeps a number of versions, and names them based on when the file was modified. Staggered tracks when the versions were created so it can keep x amount of them in a certain time interval etc.

AudriusButkevicius:
Sure, but for purposes of restoring, it's confusing and not obvious that/why it differs.

calmh:
So for clarity for future readers (because I had to go looking) we now always use the existing file modtime for the timestamp in the version name.

calmh:
This was changed in v1.2.0, so that the timestamp in the *name* is when we archived the file; the mtime is unchanged.

The trashcan versioner doesn’t alter the name, so instead it *does* change the mtime to when we archived the file. This inconsistency will persist.

PR #4602, opened 2017-12-16
cmd/syncthing: UI for version restoration (fixes #2599)

### LINKED ISSUE #2599: GUI for .stversion restoration (feature)
I've come across the versioning function of Syncthing. I use it to sync the files to a backup storage.
However restoration form the .stversion folder is a mess. You have to rename all the files to remove the timestamp in the name. It would be great to have a GUI for this function.
## USECASE:

I had the problem, when i was working with my colleague on a school project.
I accidentally deleted a folder inclusive its subfolders. Through inotify, they were also immediately deleted on his side too. Luckily he had the "Simple Versioning" enabled. So we could copy all the files back from the .stversion folder. But it was HARD work to manually rename all the files back to their original name and to sort the multiple versions out.

It would have been only four or five clicks with a GUI. Just give the order to revert the files (copy the files back to the Sync folder and rename them), which have a timestamp between 13:10 and 13:20.

And at least, I had to ask my Colleague to copy the files back, because they were in his .stversion folder. I couldn't have done anything from my side.

I just thought it would be a nice feature in usability, to have such a GUI revert function.
## IMPLEMENTATION:

It would just require a list of all the files in .stversion folder to be transfered.
After that the list only has to be filtered for the files, the user wants to restore,
based on the timestamp and send back.
At least a single command could be emited to fullfill the copy and rename process of the selected files.
## GUI:

I propose to insert a "Revert" Button next to the "Rescan" and "Edit" buttons of the folders.
Of course only if versioning is enabled in any of the synced devices.

This opens a Window like following Picture:
gui_proposal

AudriusButkevicius:
Fetches all stuff in one go, otherwise we couldn't do filtering and stuff.
Can't for the life of me get the templated translation stuff to work.

image
image
image
image
image


calmh:
Woah

uok:
🎅 is on 🔥

AudriusButkevicius:
Don't understand the mac failure, and don't feel like debugging it by committing on the branch.

lib/model/model_test.go
```diff
@@ -2870,6 +2872,215 @@ func TestIssue4475(t *testing.T) {
...
+	}
+
+	for name := range expectedVersions {
+		t.Errorf("not found expected %s", name)
+	}
+
+	// Restoring non existing folder fails.
+	_, err = m.RestoreFolderVersions("does not exist", nil)
+	if err == nil {
+		t.Error(err)
```
imsodin: Logging only nil on error is a bit obscure - "Should return error on non existing folder"?

lib/model/model.go
```diff
@@ -2376,6 +2374,138 @@ func (m *Model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
+		// either does not exist, or is actually a file.
+		if info, err := filesystem.Lstat(file); err == nil && !info.IsRegular() {
+			errors[file] = "cannot replace a non-file"
+			continue
+		} else if err != nil && !fs.IsNotExist(err) {
+			errors[file] = err.Error()
+			continue
+		} else {
+
+		}
```
imsodin: Nothing in the else and generally, shouldn't it be either if...else or continue but no else?

gui/default/syncthing/folder/restoreVersionsMassActions.html
```diff
@@ -0,0 +1,11 @@
+<div class="dropdown">
+  <button class="btn btn-default btn-xs dropdown-toggle" type="button" data-toggle="dropdown">
+    <span translate>Mass actions</span>
```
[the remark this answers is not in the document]
AudriusButkevicius: disagree in this case, it's not a dialog title, it's dropdown item.

lib/model/model.go
```diff
@@ -2376,6 +2374,138 @@ func (m *Model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
+		return nil, err
+	}
+
+	return files, nil
+}
+
+func (m *Model) RestoreFolderVersions(folder string, versions map[string]int64) (map[string]string, error) {
+	fcfg, ok := m.cfg.Folder(folder)
+	if !ok {
+		return nil, errFolderMissing
```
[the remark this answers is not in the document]
AudriusButkevicius: I guess if you screwed something up big time you'd might want to pause and restore.

gui/default/syncthing/folder/restoreVersionsMassActions.html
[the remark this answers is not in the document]
AudriusButkevicius: we could do with a better name though, as it's not obvious that is works on sub-items.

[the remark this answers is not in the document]
imsodin: "Select recursively"? Then "Do not restore all" should become something like "Deselect all", the other two options already fit (except that I would use plural "versions").

lib/model/model.go
```diff
@@ -2376,6 +2374,136 @@ func (m *Model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
+	}
+
+	loc, err := time.LoadLocation("Local")
+	if err != nil {
+		return nil, err
+	}
+
+	files := make(map[string][]interface{})
```
calmh: Why not `map[string][]FileVersion` ?

lib/model/model.go
```diff
@@ -2376,6 +2374,136 @@ func (m *Model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
+		}
+
+		filesystem.MkdirAll(filepath.Dir(target), 0755)
+		if err == nil {
+			err = osutil.Copy(filesystem, source, target)
+		}
+
+		if err != nil {
+			errors[target] = err.Error()
+			delete(restore, target)
```
calmh: This delete looks like a no-op and confuses me

lib/model/model.go
```diff
@@ -65,6 +65,12 @@ type Availability struct {
 	FromTemporary bool              `json:"fromTemporary"`
 }
 
+type FileVersion struct {
+	VersionTime int64 `json:"versionTime"`
```
calmh: We typically send times as `time.Time` over JSON. Why not here?

lib/model/model.go
AudriusButkevicius: Gui does a good job not to depend on lib/model, this would have to changed if we want to return something from the model package there.

lib/model/model.go
AudriusButkevicius: We'd have to parse that there and back here requiring intermediate structs, and the API is based around timestamps.

calmh: This is just a matter of `foo.versionTime = new Date(foo.versionTime)` on the Javascript side, and it should work automatically on the way back.

lib/model/model.go
calmh: Ugh, yeah. Hmm. Would it be super ugly to make this struct a `versioner.Version` or something? Otherwise at least a comment on what's going on here would be nice...

AudriusButkevicius:
Addressed comments.
Templating seems to work now, probably just needed a refresh or something.

AudriusButkevicius:
@calmh when will you have time to try this out?

calmh:
So mostly this looks good. One technical problem, one nit. Technical problem might be because of my previous complaints - there is a time zone issue in the file tag handling.

screen shot 2017-12-29 at 21 16 44

```
$ ls -la .stversions/stindex/
total 32
drwxr-xr-x   6 jb  staff   192 Dec 29 21:14 ./
drwxr-xr-x  18 jb  staff   576 Dec 29 21:14 ../
-rw-r--r--   1 jb  staff  2293 Dec 29 21:10 dumpsize~20171229-211038.go
-rw-r--r--   1 jb  staff  2334 Dec 29 21:10 dump~20171229-211038.go
-rw-r--r--   1 jb  staff   857 Dec 29 21:10 main~20171229-211038.go
-rw-r--r--   1 jb  staff  1083 Dec 29 21:10 util~20171229-211038.go
```

The file name it could not restore according to the screenshot is one hour off from the one actually on disk. I'm currently at UTC+1 which is probably not a coincidence.

Nit: The font for directories is super light and all but unreadable when selected. (This probably looks differently on non-Mac and/or non-HiDPI.) This could be because we only ship a standard and light font weight (or something), I think we use/used the light font in headers. It's also sucky that the whole dialog scrolls instead of just the file list, but that's a problem shared with every other dialog we have so I'm not going to judge too harshly in this specific case.

screen shot 2017-12-29 at 21 19 34

AudriusButkevicius:
Timezones are on you to blame, as it worked fine while I used a sane metric which starts at 1970.

AudriusButkevicius:
Can you identify where it gets skewed?

calmh:
Not specifically, but it *says* "21:10" in all the little dropdowns in the UI, and only fails at the restore stage when looking for "20:10". So at a guess, in the Javscript to Go conversion. Honestly I'm not sure how that even happens. I could imagine javascript side sending "21:10" and it getting interpreted as UTC on the Go side before becoming local time, but that's the wrong error direction. Or maybe it's just the formatter for the filename tag that uses UTC when it should use local time. 🤷‍♂️

AudriusButkevicius:
Try now.

calmh:
New error, complaining about the nonexistence of the target, which does not make sense to me.

screen shot 2017-12-30 at 10 13 43

```
jb@kvar:~/s/g/s/s/test $ ls -l s2/stcli/
jb@kvar:~/s/g/s/s/test $ ls -l s2/.stversions/stcli/
total 128
-rw-r--r--  1 jb  staff  1068 Dec 29 21:10 LICENSE~20171229-211038
-rw-r--r--  1 jb  staff  2771 Dec 29 21:10 client~20171229-211038.go
-rw-r--r--  1 jb  staff  4501 Dec 29 21:10 cmd_devices~20171229-211038.go
-rw-r--r--  1 jb  staff  1563 Dec 29 21:10 cmd_errors~20171229-211038.go
-rw-r--r--  1 jb  staff  9319 Dec 29 21:10 cmd_folders~20171229-211038.go
-rw-r--r--  1 jb  staff  1838 Dec 29 21:10 cmd_general~20171229-211038.go
-rw-r--r--  1 jb  staff  2906 Dec 29 21:10 cmd_gui~20171229-211038.go
-rw-r--r--  1 jb  staff  5230 Dec 29 21:10 cmd_options~20171229-211038.go
-rw-r--r--  1 jb  staff  1597 Dec 29 21:10 cmd_report~20171229-211038.go
-rw-r--r--  1 jb  staff  1092 Dec 29 21:10 labels~20171229-211038.go
-rw-r--r--  1 jb  staff  1329 Dec 29 21:10 main~20171229-211038.go
-rw-r--r--  1 jb  staff  3713 Dec 29 21:10 utils~20171229-211038.go
jb@kvar:~/s/g/s/s/test $ 
```

AudriusButkevicius:
It's complaining that the dir does not exist. I though that's covered, but let me try.

calmh:
The dir does exist, I ls:d it above. Unless it created it after complaining.

AudriusButkevicius:
I know it works for me, but I have all the dirs in place. Let me add a test case to try and understand this.

AudriusButkevicius:
Try now.

AudriusButkevicius:
@st-review rebuild

AudriusButkevicius:
@calmh any other comments?

calmh:
It works now. But there is a GUI bug:

screen shot 2017-12-30 at 19 46 38

The modal doesn't scroll any further up, because that's the end of the page, leaving most of the date selector off screen.

Did you sort out whatever it was with the translation strings or that's moot?

AudriusButkevicius:
Yeah translation worked. I can force it to open up I guess.

calmh:
And of course, now #4586 actually matters.

AudriusButkevicius:
This drops up now.

This doesn't work with trashcan versioner, as that one doesn't tag filenames, yet I guess it's fine.

calmh:
Clearly you don't get off the hook that easily

screen shot 2017-12-30 at 20 20 12

That's also unscrollable

calmh:
I also think this might merit button ("Restore Versions") placement at the outmost folder level, i.e. beside "Edit", not inside the edit dialog. I would not go looking inside the edit dialog when I wanted to restore a file.

AudriusButkevicius:
I give up. If you have a single file, you might not want to filter by date I guess.

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -196,6 +196,9 @@
     <button type="button" class="btn btn-default btn-sm" id="editIgnoresButton" ng-click="editingExisting ? editIgnores() : editIgnoresOnAddingFolder()" ng-disabled="folderEditor.$invalid">
       <span class="fa fa-eye-slash"></span>&nbsp;<span translate>Ignore Patterns</span>
     </button>
+    <button type="button" class="btn btn-default btn-sm" ng-click="restoreVersions.show()" ng-if="editingExisting && currentFolder.versioning && currentFolder.versioning.type">
+      <span class="fa fa-calendar"></span>&nbsp;<span translate>Restore Versions</span>
```
calmh: I'd go with `fa-undo` here instead of the calendar?

calmh:
`min-height: 75%` or something on the containing modal? If that works and doesn't look like complete ass I think it's a solution, otherwise I guess it's good enough as it is

calmh:
> This doesn't work with trashcan versioner

We should make that work, but it's a valid later step

AudriusButkevicius:
So even if I make the modal 75% or something (which is not trivial because we have an angular directive), I can't seem to find a way to fill the empty space.

AudriusButkevicius:
> We should make that work, but it's a valid later step
I'm not sure how you'd make it work given it doesn't actually have a version...

AudriusButkevicius:
@st-review rebuild

AudriusButkevicius:
@calmh bum0p

calmh:
Did you try the min-height thing?

AudriusButkevicius:
Yeah, I can't find an easy way to do it given out <modal> directive hides most of the wrapper div's.
Even if I make the modal 75% bigger, I can't find a way to push/expand the element that holds the tree.

AudriusButkevicius:
This is how many heights I have to add, and the awesome end result:

image

calmh:
@st-review merge

cmd/syncthing: Add UI for version restoration (fixes #2599)

st-review:
@calmh: Build status is `failure` -- refusing to merge.

calmh:
yay a small window of passing tests

quick

merge!

calmh:
@st-review merge

cmd/syncthing: Add UI for version restoration (fixes #2599)

st-review:

calmh:
the fuck?

calmh:
it's not pending

calmh:
@st-review merge

cmd/syncthing: Add UI for version restoration (fixes #2599)

Skip-check: mac

st-review:
@calmh: There's already a merge pending for this PR.

calmh:
Right.

st-review:
@calmh: Merge failed:

```
$ git fetch -f origin refs/pull/4602/head:pr-4602
From https://github.com/syncthing/syncthing
 * [new ref]           refs/pull/4602/head -> pr-4602
From https://github.com/syncthing/syncthing
   a9f0659f..b0e2050c  master     -> orig/master
   c7f136c2..b0e2050c  master     -> origin/master
$ git reset --hard
HEAD is now at c7f136c2 lib/upnp: Each service is it's own NAT device
$ git checkout master
Already on 'master'
Your branch is behind 'origin/master' by 1 commit, and can be fast-forwarded.
HEAD is now at b0e2050c cmd/syncthing: UI for version restoration (fixes #2599) (#4602)
$ git clean -fxd
Skipping repository syncthing/syncthing
$ git merge --squash --no-commit pr-4602
$ git commit -F -
On branch master
nothing to commit, working tree clean

```

generalmanager:
@AudriusButkevicius This is incredible, thank you so much for all the work you did on this! 
@calmh I agree on always using the archive time. This way it's consistent and we can use a short and meaningful label like "Filter by archival date" / "Filter by replacement date" which should make sense even to the less technical users.

generalmanager:
This might already be implemented, but if so I couldn't find it in the screenshots:
It would be sweet if one could choose a different restore path than the original one to review differences before merging or if the original path was e.g. a mounted NFS which disappeared.

AudriusButkevicius:
Please elaborate. I don't understand what is being asked.

generalmanager:
Oh sure, I just wanted to keep it as short and conciese as possible, as I tend to overcomplicate things. Seems like I failed again tough.

Can the files only be restored into the parent-folder of the .stversions folder or can the folder to which the files are restored be chosen freely?

AudriusButkevicius:
It restores into the parent. Restoring into a separate folder makes no sense (as it's not a restore just a copy to a random place in that case).

generalmanager:
Restoring to the parent is obviously the main use case. But imo the magic is the renaming while keeping the folder structure intact, so beeing able to restore to another directory could be useful in certain usecases (I'll use the syncthing UI terminology to avoid confusion.):

- It can't be restored to the folder defined in the "Folder Path" setting because this was on some mounted directory which isn't available at the time of restoring e.g. because it's a network drive or an external drive which broke and has to be replaced. 

This might sound like an obscure scenario, because by default .stversions folder is inside the folder path set in the config. But as syncthing allows archiving configuring the "Versions Path", conflicts like this might be more common than one would assume.

- A big folder structure was deleted at the same time, but only a subset should be restored, e.g. because the accidental deletion wasn't noticed for some time. So at the time of restoring some files in the folder path are newer than in the .stversions folder and should not be overwritten. There are also more complicated scenarios, but what I'm mainly getting at is that one would want to manually do a diff and precisely check for changes in content, ACLs and so on.

Does that make more sense to you?

calmh:
I think it's quite common for backup-like things to offer restoration to a new, separate directory. And we're sort of acting like a backup-like thing here, so yes it makes sense.

But that's a new feature ticket, not a discussion on this PR. This implementation is fine as is for a first implementation.

generalmanager:
Sure, I'll open a new ticket. I wrote here because I couldn't tell if it was already implemented so I wanted to ask first. Just wanted to say thanks again to @AudriusButkevicius for spending so much time on this huge PR.

PR #4603, opened 2017-12-17
 lib/connections, lib/config: Bandwidth throttling per remote device (fixes #4516)

### LINKED ISSUE #4516: Bandwidth throttling per remote device
We use syncthing in many controlled situations, where we could use bandwidth throttling per remote device, beside the already available global throttle setting.

It would be quite useful to have something like "maxRecvKbps" and "maxSendKbps" _per remote device_ in the advanced configuration, beside as a global option for the local device.

Thank you for giving us syncthing, it is a wonderful tool.

qepasa:
### Purpose

Hello everyone. As described in the issue. This fix enables read/write bandwidth throttling per remote device alongside current global limiting. It can be set via device tab in advanced settings.

### Testing

Testing was done using 2 remote devices. Tested scenarios include adding and removing devices and various configurations of r/w rates:
device1: 300/300, device2: unlimited/unlimited, global: unlimited/unlimited
device1: 300/300, device2: unlimited/unlimited, global: 150/150
device1: 300/300, device2: 50/50, global: 150/150
device1: unlimited/unlimited, device2: 50/50, global: 150/150
device1: unlimited/unlimited, device2: unlimited/unlimited, global: 700/700

Of course these tests may not be comprehensive. I'll be thankful for any additional suggestions or scenarios.


lib/connections/limiter.go
```diff
@@ -99,50 +188,66 @@ func (lim *limiter) String() string {
...
+	limiter  *limiter
+	isLAN    bool
+	remoteID protocol.DeviceID
 }
 
 func (w *limitedWriter) Write(buf []byte) (int, error) {
 	if !w.isLAN || w.limiter.limitsLAN.get() {
-		take(w.limiter.write, len(buf))
+		deviceLimiter, ok := w.limiter.deviceWriteLimiter[w.remoteID]
+		if ok == false {
```
AudriusButkevicius: !ok

lib/connections/limiter.go
```diff
@@ -12,53 +12,111 @@ import (
...
-	return &limitedWriter{writer: w, limiter: lim, isLAN: isLAN}
+// Create new maps with new limiters
+func (lim *limiter) rebuildMap(to config.Configuration) {
+	lim.deviceWriteLimiter = make(deviceLimiters)
+	lim.deviceReadLimiter = make(deviceLimiters)
+	for _, v := range to.Devices {
+		lim.deviceWriteLimiter[v.DeviceID] = rate.NewLimiter(rate.Inf, limiterBurstSize)
+		lim.deviceReadLimiter[v.DeviceID] = rate.NewLimiter(rate.Inf, limiterBurstSize)
+	}
+	l.Debugln("Rebuild finished")
```
AudriusButkevicius: This has no context, rebuild of what, where, why

lib/connections/limiter.go
```diff
@@ -12,53 +12,111 @@ import (
...
 	cfg.Subscribe(l)
-	prev := config.Configuration{Options: config.OptionsConfiguration{MaxRecvKbps: -1, MaxSendKbps: -1}}
+	prev := config.Configuration{Options: config.OptionsConfiguration{MaxRecvKbps: -1, MaxSendKbps: -1}, Devices: devices}
 	l.CommitConfiguration(prev, cfg.RawCopy())
 	return l
 }
 
-func (lim *limiter) newReadLimiter(r io.Reader, isLAN bool) io.Reader {
-	return &limitedReader{reader: r, limiter: lim, isLAN: isLAN}
+func getInitialDevicesConfiguration(cfgCopy config.Configuration) config.DeviceConfigurationList {
```
AudriusButkevicius: This is only called once, I'd say that it's fine to wedge in the loop inline in the constructor.

lib/connections/limiter.go
```diff
@@ -12,53 +12,111 @@ import (
...
+		myID:               deviceID,
+	}
+
+	// Keep read/write limiters for every connected device
+	devices := getInitialDevicesConfiguration(cfg.RawCopy())
+	for _, v := range devices {
+		l.deviceReadLimiter[v.DeviceID] = rate.NewLimiter(rate.Inf, limiterBurstSize)
+		l.deviceWriteLimiter[v.DeviceID] = rate.NewLimiter(rate.Inf, limiterBurstSize)
 	}
+
```
AudriusButkevicius: Could just call rebuildMap here?

lib/connections/limiter.go
```diff
@@ -12,53 +12,111 @@ import (
...
+		lim.deviceWriteLimiter[v.DeviceID] = rate.NewLimiter(rate.Inf, limiterBurstSize)
+		lim.deviceReadLimiter[v.DeviceID] = rate.NewLimiter(rate.Inf, limiterBurstSize)
+	}
+	l.Debugln("Rebuild finished")
+}
+
+// Compare read/write limits in configurations
+func (lim *limiter) checkDeviceLimits(from, to config.Configuration) bool {
+	for i := range from.Devices {
+		if from.Devices[i].DeviceID.Compare(to.Devices[i].DeviceID) != 0 {
```
AudriusButkevicius: You can just use equality for device ID's they should be comparable just fine.

lib/connections/limiter.go
```diff
@@ -73,22 +131,53 @@ func (lim *limiter) CommitConfiguration(from, to config.Configuration) bool {
...
 	recvLimitStr := "is unlimited"
 	if to.Options.MaxSendKbps > 0 {
 		sendLimitStr = fmt.Sprintf("limit is %d KiB/s", to.Options.MaxSendKbps)
 	}
+
 	if to.Options.MaxRecvKbps > 0 {
 		recvLimitStr = fmt.Sprintf("limit is %d KiB/s", to.Options.MaxRecvKbps)
 	}
-	l.Infof("Send rate %s, receive rate %s", sendLimitStr, recvLimitStr)
+	l.Infof("Global send rate %s, receive rate %s", sendLimitStr, recvLimitStr)
```
AudriusButkevicius: Overall or Total. Global usually talks about locations.

lib/connections/limiter.go
```diff
@@ -99,50 +188,66 @@ func (lim *limiter) String() string {
...
 }
 
 func (r *limitedReader) Read(buf []byte) (int, error) {
 	n, err := r.reader.Read(buf)
 	if !r.isLAN || r.limiter.limitsLAN.get() {
-		take(r.limiter.read, n)
+		deviceLimiter, ok := r.limiter.deviceReadLimiter[r.remoteID]
+		if ok == false {
+			l.Debugln("deviceLimiter was not in the map")
+			deviceLimiter = rate.NewLimiter(rate.Inf, limiterBurstSize)
```
AudriusButkevicius: How can this happen?

lib/connections/limiter.go
```diff
@@ -12,53 +12,111 @@ import (
 	"sync/atomic"
 
 	"github.com/syncthing/syncthing/lib/config"
+	"github.com/syncthing/syncthing/lib/protocol"
 	"golang.org/x/net/context"
 	"golang.org/x/time/rate"
 )
 
+type deviceLimiters map[protocol.DeviceID]*rate.Limiter
```
AudriusButkevicius: This probably needs to be sync.Map (atomic map) as you have not implemented any locking, and there are definately multiple routines poking at it. Also, the atomic map should only be initialized once, because if you re-assign the map on the limiter struct, you still need a lock.

lib/connections/limiter.go
qepasa: It happens after removing remote device. I'm not sure about limitedWriter but I added it just to be safe.

lib/connections/limiter.go
```diff
@@ -73,22 +146,55 @@ func (lim *limiter) CommitConfiguration(from, to config.Configuration) bool {
...
 	lim.limitsLAN.set(to.Options.LimitBandwidthInLan)
 
+	// Set limits for devices
+	for _, v := range to.Devices {
+		if v.DeviceID == lim.myID {
+			// This limiter was created for local device. Should skip this device
+			continue
+		}
+
+		readLimiter, _ := lim.deviceReadLimiters.Load(v.DeviceID)
```
AudriusButkevicius: You should not throw away the ok, in case the value is not there.
Also, you are not locking here, so you might be accessing deviceReadLimiters as the map is being swapped out.

This is also a reason why rebuildMap should potentially do it without swapping the maps.

lib/connections/limiter.go
```diff
@@ -12,53 +12,126 @@ import (
...
+		}
+	}
+
+	// assign new maps
+	lim.mu.Lock()
+	defer lim.mu.Unlock()
+	lim.deviceWriteLimiters = deviceWriteLimiters
+	lim.deviceReadLimiters = deviceReadLimiters
+
+	l.Debugln("Rebuild of device limiters map finished")
```
AudriusButkevicius: Wouldn't it be better for rebuild to happen in place, this way getting rid of the lock that we have to get on every read and write.

lib/connections/limiter.go
```diff
@@ -12,53 +12,126 @@ import (
...
 }
 
-func (lim *limiter) newWriteLimiter(w io.Writer, isLAN bool) io.Writer {
-	return &limitedWriter{writer: w, limiter: lim, isLAN: isLAN}
+// Compare read/write limits in configurations
+func (lim *limiter) checkDeviceLimits(from, to config.Configuration) bool {
+	for i := range from.Devices {
+		if from.Devices[i].DeviceID != to.Devices[i].DeviceID {
```
AudriusButkevicius: This should blow up if a device is removed, as len(to) < len(from).

lib/connections/limiter.go
```diff
@@ -12,53 +12,126 @@ import (
...
-		from.Options.LimitBandwidthInLan == to.Options.LimitBandwidthInLan {
+		from.Options.LimitBandwidthInLan == to.Options.LimitBandwidthInLan &&
+		lim.checkDeviceLimits(from, to) {
 		return true
 	}
 
+	// A device has been added or removed
+	if len(from.Devices) != len(to.Devices) {
+		lim.rebuildMap(to)
+	}
```
AudriusButkevicius: Also, this is unnecessery if you handle it properly in checkDeviceLimits

lib/connections/limiter.go
AudriusButkevicius: You should probably write a test case for this logic.
Test case for adding
Test case for removing 
Test case for adding and removing in the same config commit cycle.

lib/connections/limiter.go
```diff
@@ -12,53 +12,126 @@ import (
...
 }
 
-func (lim *limiter) newWriteLimiter(w io.Writer, isLAN bool) io.Writer {
-	return &limitedWriter{writer: w, limiter: lim, isLAN: isLAN}
+// Compare read/write limits in configurations
+func (lim *limiter) checkDeviceLimits(from, to config.Configuration) bool {
+	for i := range from.Devices {
+		if from.Devices[i].DeviceID != to.Devices[i].DeviceID {
+			// Something has changed in device configuration
+			lim.rebuildMap(to)
```
AudriusButkevicius: To be honest, this function as it states should only do checking, and not rebuilding (as a side-effect), and rebuilding and other cruft should happen in the CommitConfiguration.

lib/connections/limiter.go
```diff
@@ -73,22 +146,55 @@ func (lim *limiter) CommitConfiguration(from, to config.Configuration) bool {
...
+		sendLimitStr := "is unlimited"
+		recvLimitStr := "is unlimited"
+		if v.MaxSendKbps > 0 {
+			sendLimitStr = fmt.Sprintf("limit is %d KiB/s", v.MaxSendKbps)
+		}
+
+		if v.MaxRecvKbps > 0 {
+			recvLimitStr = fmt.Sprintf("limit is %d KiB/s", v.MaxRecvKbps)
+		}
+		l.Infof("Device %s: send rate %s, receive rate %s", v.DeviceID, sendLimitStr, recvLimitStr)
```
AudriusButkevicius: This will always print stuff, even if someones rate hasn't changed since before.
It will further always mess around with the rate limiters, regardless if the value changed or not.

imsodin:
Now you are using both locks and `sync.Map`, why?

The `rebuildMap` calls in several places seem overly complicated to me: Couldn't you just iterate over devices from old and new configuration in `CommitConfiguration` and delete/add/adjust limiters as necessary? When initializing you can then just call `CommitConfiguration` with empty maps as well.

lib/connections/limiter.go
```diff
@@ -99,50 +205,74 @@ func (lim *limiter) String() string {
...
 func (r *limitedReader) Read(buf []byte) (int, error) {
 	n, err := r.reader.Read(buf)
 	if !r.isLAN || r.limiter.limitsLAN.get() {
-		take(r.limiter.read, n)
+
+		// in case rebuildMap was called
+		r.limiter.mu.Lock()
+		deviceLimiter, ok := r.limiter.deviceReadLimiters.Load(r.remoteID)
+		r.limiter.mu.Unlock()
+		if !ok {
```
AudriusButkevicius: To be honest, if !ok, I'd panic (or do nothing), as it should clearly be there.
Now it's like a bandaid for a problem we can't explain.

I'd do a if deviceLimiter != nil check in take

lib/connections/limiter.go
```diff
@@ -99,50 +205,74 @@ func (lim *limiter) String() string {
...
+		take(w.limiter.write, deviceLimiter.(*rate.Limiter), len(buf))
 	}
 	return w.writer.Write(buf)
 }
 
 // take is a utility function to consume tokens from a rate.Limiter. No call
 // to WaitN can be larger than the limiter burst size so we split it up into
 // several calls when necessary.
-func take(l *rate.Limiter, tokens int) {
+func take(l, deviceLimiter *rate.Limiter, tokens int) {
```
AudriusButkevicius: the l could do with a better name now, and the doc string should be updated.

lib/connections/limiter.go
```diff
@@ -12,53 +12,126 @@ import (
...
-func (lim *limiter) newReadLimiter(r io.Reader, isLAN bool) io.Reader {
-	return &limitedReader{reader: r, limiter: lim, isLAN: isLAN}
+// Create new maps with new limiters
+func (lim *limiter) rebuildMap(to config.Configuration) {
+	deviceWriteLimiters := new(sync.Map)
+	deviceReadLimiters := new(sync.Map)
+
+	// copy *limiter in case remote device is still connected, when remote device is added we create new read/write limiters
+	for _, v := range to.Devices {
+		if readLimiter, ok := lim.deviceReadLimiters.Load(v.DeviceID); ok {
```
AudriusButkevicius: What if there are two rebuildMap calls in parallel due to frequent config save or something like that, this might be might-way through a swap too.

qepasa:
@AudriusButkevicius You're right. I assumed (which was pretty stupid) that only one routine at the time visits CommitConfiguration. I guess the right way will be to rewrite it so that "rebuild" happens without swapping map instances and definitely implement mentioned tests and fix other issues pointed by you. Thanks for feedback!

@imsodin My intention was to use lock in order to ensure atomic assignment in rebuildMap.

AudriusButkevicius:
@qepasa thanks for working on this.

lib/connections/limiter.go
```diff
@@ -12,53 +12,126 @@ import (
...
 // as appropriate.
 type limiter struct {
-	write     *rate.Limiter
-	read      *rate.Limiter
-	limitsLAN atomicBool
+	write               *rate.Limiter
+	read                *rate.Limiter
+	limitsLAN           atomicBool
+	deviceReadLimiters  *sync.Map
+	deviceWriteLimiters *sync.Map
```
calmh: `sync.Map` is Go 1.9 and we're currently targeting 1.8. Although we *could* bump our requirements I think there should be little enough churn on these that we can actually use regular maps and locking...

qepasa:
Hello everyone. Sorry for delay with coding. I tried to keep your suggestions in mind. If there is anything you would like to be done differently I'll make sure to implement it faster this time.

lib/connections/limiter.go
```diff
@@ -12,53 +12,197 @@ import (
...
+		}
+	}
+
+	l.Debugln("Rebuild of device limiters map finished")
 }
 
-func (lim *limiter) newWriteLimiter(w io.Writer, isLAN bool) io.Writer {
-	return &limitedWriter{writer: w, limiter: lim, isLAN: isLAN}
+// Compare read/write limits in configurations
+func (lim *limiter) checkDeviceLimits(from, to config.Configuration) bool {
```
AudriusButkevicius: Again, doesn't need to live on lim.
Also the function name doesn't explain well what this does. Should probably be called deviceLimitsChanged() and invert what it returns.

lib/connections/limiter.go
```diff
@@ -12,53 +12,197 @@ import (
...
+	devices := cfg.RawCopy().Devices
+	for _, value := range devices {
+		value.MaxRecvKbps = -1
+		value.MaxSendKbps = -1
+
+		// Keep read/write limiters for every connected device
+		l.deviceWriteLimiters[value.DeviceID] = rate.NewLimiter(rate.Inf, limiterBurstSize)
+		l.deviceReadLimiters[value.DeviceID] = rate.NewLimiter(rate.Inf, limiterBurstSize)
+	}
+	prev := config.Configuration{Options: config.OptionsConfiguration{MaxRecvKbps: -1, MaxSendKbps: -1}, Devices: devices}
```
imsodin: I believe the entire above block isn't necessary, as CommitConfiguration will add all missing devices anyway, so https://github.com/syncthing/syncthing/pull/4603/files#diff-da45d4f40dfa16a9aae760dd650fe79aL35 should be sufficient.

lib/connections/limiter.go
```diff
@@ -12,53 +12,197 @@ import (
...
+	} else {
+		writeLimiter.SetLimit(1024 * rate.Limit(v.MaxSendKbps))
+	}
+}
+
+// This function handles removing, adding and updating of device limiters.
+// Pass pointer to avoid copying. Pointer already points to copy of configuration
+// so we don't have to worry about modifying real config.
+func (lim *limiter) processDevicesConfiguration(from, to *config.Configuration) {
+
```
imsodin: empty

lib/connections/limiter.go
```diff
@@ -12,53 +12,197 @@ import (
...
-func (lim *limiter) newWriteLimiter(w io.Writer, isLAN bool) io.Writer {
-	return &limitedWriter{writer: w, limiter: lim, isLAN: isLAN}
+// Compare read/write limits in configurations
+func (lim *limiter) checkDeviceLimits(from, to config.Configuration) bool {
+	if len(from.Devices) != len(to.Devices) {
+		return false
+	}
+	// len(from.Devices) == len(to.Devices) so we can do range from.Devices
+	for i := range from.Devices {
+		if from.Devices[i].DeviceID != to.Devices[i].DeviceID {
```
AudriusButkevicius: This is a nit-pick, but we don't really need to rely on the device order here. It's possible that rates aren't changed, just the order gets changed.

lib/connections/limiter.go
```diff
@@ -39,97 +40,167 @@ func newLimiter(deviceID protocol.DeviceID, cfg *config.Wrapper) *limiter {
...
+func (lim *limiter) processDevicesConfiguration(from, to *config.Configuration) {
 
-	// copy *limiter in case remote device is still connected, when remote device is added we create new read/write limiters
+	devicesToRemove := make(map[protocol.DeviceID]bool)
+	// Get all devices that were previously in map, these are candidates for removal
+	for _, v := range from.Devices {
+		if v.DeviceID != to.MyID {
+			devicesToRemove[v.DeviceID] = true
+		}
+	}
```
imsodin: Could be `map[protocol.DeviceID]struct{}` and then later instead of `= false` do `delete(devicesToRemove, v.DeviceID)`. Then you can just iterate over the final map in the end without checking a bool.

lib/connections/limiter.go
```diff
@@ -12,53 +12,197 @@ import (
...
+		writeLimiter.SetLimit(1024 * rate.Limit(v.MaxSendKbps))
+	}
+}
+
+// This function handles removing, adding and updating of device limiters.
+// Pass pointer to avoid copying. Pointer already points to copy of configuration
+// so we don't have to worry about modifying real config.
+func (lim *limiter) processDevicesConfiguration(from, to *config.Configuration) {
+
+	devicesToRemove := make(map[protocol.DeviceID]bool)
```
AudriusButkevicius: This could be `map[DeviceID]struct{}`, and `devicesToRemove[x] = false` is essentially `delete(devicesToRemove, x)`

Or better would be:
```
for dev in to.Devices {
   seen[dev] = struct{}{}
}

for dev in from.Devices {
   if  _, ok := seen[dev]; !ok {
      lim.deleteLimiters(dev)
   }
}
```

lib/connections/limiter.go
```diff
@@ -12,53 +12,197 @@ import (
...
+	myID                protocol.DeviceID
+	mu                  *sync.Mutex
+	deviceMapMutex      *sync.RWMutex
 }
 
 const limiterBurstSize = 4 * 128 << 10
 
-func newLimiter(cfg *config.Wrapper) *limiter {
+func newLimiter(deviceID protocol.DeviceID, cfg *config.Wrapper) *limiter {
+
```
imsodin: empty

lib/connections/limiter.go
AudriusButkevicius: There should be some sort of DeviceMap or something like that utility function.

lib/connections/limiter.go
```diff
@@ -73,22 +217,25 @@ func (lim *limiter) CommitConfiguration(from, to config.Configuration) bool {
 
 	lim.limitsLAN.set(to.Options.LimitBandwidthInLan)
 
+	// Delete, add or update limiters for devices
+	lim.processDevicesConfiguration(&from, &to)
```
AudriusButkevicius: Why is this a pointer?

lib/connections/limiter.go
```diff
@@ -12,53 +12,197 @@ import (
...
+		}
+
+		writeLimiter, okW := lim.getDeviceWriteLimiter(v.DeviceID)
+		if okW {
+			if devicesToRemove[v.DeviceID] {
+				// Device has not been removed or added, we should've
+				// already marked it as false in devicesToRemove
+				panic("broken symmetry in device read/write limiters")
+			}
+		}
```
AudriusButkevicius: This is too complicated. The only way these things get set are under a lock, so it's very unlikely they get out of sync.
Also, perhaps it's better to get getDevice{Read,Write}Limiter always return a new limiter even if one does not exists (and add it to the map).

I am thinking about a case where a new device gets added, and there is a race between CommitConfiguration here, and us making a connection. It's probably always better wrap every connection in a limiter (and create one if it doesn't exist) and then adjust the rates later.

lib/connections/limiter.go
```diff
@@ -12,53 +12,197 @@ import (
...
+	lim.mu.Lock()
+	defer lim.mu.Unlock()
+
 	if from.Options.MaxRecvKbps == to.Options.MaxRecvKbps &&
 		from.Options.MaxSendKbps == to.Options.MaxSendKbps &&
-		from.Options.LimitBandwidthInLan == to.Options.LimitBandwidthInLan {
+		from.Options.LimitBandwidthInLan == to.Options.LimitBandwidthInLan &&
+		lim.checkDeviceLimits(from, to) {
 		return true
 	}
```
imsodin: I think this global check is not very useful as global limits are reset even if just a device changed and whether device change is checked again later on. So I would either not check at all and just reset everything or check the different parameters separately.

lib/connections/limiter.go
```diff
@@ -12,53 +12,197 @@ import (
...
+			panic("broken symmetry in device read/write limiters")
+		}
+
+		// limiters for this device are created so we can store previous rates for logging
+		previousReadLimit := readLimiter.Limit()
+		previousWriteLimit := writeLimiter.Limit()
+
+		l.Debugf("okR: %t, okW: %t, prevWriteLim: %d, prevReadLim: %d", okR, okW, previousReadLimit, previousWriteLimit)
+
+		lim.setLimitsForDevice(v, readLimiter, writeLimiter)
```
AudriusButkevicius: Is it not better not to do this if the limits are the same as previously?
As in, move this after the if?

lib/connections/limiter.go
imsodin: Yes, that should be moved from model to config: https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L2569
That also makes it easier to remove limiters, just treat it like folders in models CommitConfiguration.

imsodin:
Argh this github reviews... Probably better look at the comments in the diff, this comment view is fractured badly.

> Hello everyone. Sorry for delay with coding. I tried to keep your suggestions in mind. If there is anything you would like to be done differently I'll make sure to implement it faster this time.

Even if you weren't occupied with Christmas holidays and family reunions there is no reason at all to apologize for ~1week (and also if longer) followup times after a round of review. Reviewers will nag (or close) if it comes to the point of being too quiet ;)

lib/connections/limiter.go
```diff
@@ -155,34 +123,35 @@ func (lim *limiter) processDevicesConfiguration(from, to *config.Configuration)
...
+	for k := range fromMap {
+		if _, ok := toMap[k]; !ok {
+			return true
 		}
 		// Read/write limits were changed for this device
-		if from.Devices[i].MaxSendKbps != to.Devices[i].MaxSendKbps ||
-			from.Devices[i].MaxRecvKbps != to.Devices[i].MaxRecvKbps {
-			return false
+		if fromMap[k].MaxSendKbps != toMap[k].MaxSendKbps ||
+			fromMap[k].MaxRecvKbps != toMap[k].MaxRecvKbps {
```
imsodin: No manual line breaks (I was/am a fan of them too, but it's not the Syncthing/Go style ;) ).

lib/connections/limiter.go
```diff
@@ -12,44 +12,161 @@ import (
...
+		}
+
+		// limiters for this device are created so we can store previous rates for logging
+		previousReadLimit := readLimiter.Limit()
+		previousWriteLimit := writeLimiter.Limit()
+
+		// Nothing about this device has changed. Start processing next device
+		if okR && okW &&
+			rate.Limit(v.MaxRecvKbps)*1024 == previousReadLimit &&
+			rate.Limit(v.MaxSendKbps)*1024 == previousWriteLimit {
```
imsodin: line breaks

lib/connections/limiter.go
```diff
@@ -12,44 +12,161 @@ import (
...
 
 func (lim *limiter) VerifyConfiguration(from, to config.Configuration) error {
 	return nil
 }
 
 func (lim *limiter) CommitConfiguration(from, to config.Configuration) bool {
+	// to ensure atomic update of configuration
+	lim.mu.Lock()
+	defer lim.mu.Unlock()
+	if deviceLimitsChanged(from, to) {
```
imsodin: I would remove this check. You already do the same checks in `processDevicesConfiguration` and skip any devices which don't change.

lib/connections/limiter.go
```diff
@@ -12,44 +12,161 @@ import (
...
+		writeLimiter.SetLimit(1024 * rate.Limit(v.MaxSendKbps))
+	}
 }
 
-func (lim *limiter) newWriteLimiter(w io.Writer, isLAN bool) io.Writer {
-	return &limitedWriter{writer: w, limiter: lim, isLAN: isLAN}
+// This function handles removing, adding and updating of device limiters.
+// Pass pointer to avoid copying. Pointer already points to copy of configuration
+// so we don't have to worry about modifying real config.
+func (lim *limiter) processDevicesConfiguration(from, to *config.Configuration) {
```
imsodin: Doesn't need to pass configs as pointers.

lib/config/config.go
```diff
@@ -764,3 +764,13 @@ func cleanSymlinks(filesystem fs.Filesystem, dir string) {
 		return nil
 	})
 }
+
+// mapDeviceConfigs returns a map of device ID to device configuration for the given
+// slice of folder configurations.
+func MapDeviceConfigs(devices []DeviceConfiguration) map[protocol.DeviceID]DeviceConfiguration {
```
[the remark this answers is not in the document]
AudriusButkevicius: Agree. And we do this in a few other places (in the model I think), so we should use this instead.

AudriusButkevicius:
My previous comments on always creating limiters and always wrapping each device connection in two limiters still apply

qepasa:
@AudriusButkevicius 
I'd like to make a point against it. 
We could loose control over deleting devices. For example after removing a device there is some communications between devices which results in reading after removing limiters for said device (output from debug in limitedReader.Read). If we were to create a limiter when performing a get on deviceLimitersMap, this limiter would be left in the map despite removing a device. 
Right now there are checks in take method that pretty much result in the same outcome as creating new  limiters with unlimited rates. Of course if that is not a compelling argument in your opinion I'll change it according to your suggestion.

AudriusButkevicius:
People have finite amount of devices, and a limiter is not taking up 100MB of RAM.
I'd rather have limiters left in ram than require hacks/checks in limiterReader.Read/Write which we can't even explain why they are happening (implying that there already is some sort of race you are not covering)

qepasa:
@AudriusButkevicius 
Okay, I understand. It that case what about creating new limiter (if not present) but not inserting it to map in the getLimiters method?

AudriusButkevicius:
And why would this be better?

qepasa:
It would get rid of checks in take method, make it possible to still use rw locking instead of regular lock and avoid leaving limiters for devices which are no longer paired. On the other hand it's still kinda hacky workaround.

AudriusButkevicius:
Actually, I think this should be implemented differently all together.

Currently all the connections are wrapper in a limiter which is then wrapped in either limitedReader or limitedWriter which for some reason reach back into the limiter to get the limiter for a given device.

Why can't we construct limitedReader/limitedWriter for each connection once, give it the global limiter, and give it the per device limiter, and not have to reach into the map?

The map can still be kept for rate adjustments, but I don't see why we need to reach into the map on every read/write.

qepasa:
It seems logical. I think that your suggestion with creating and adding to map in getDevice would make sense in that case. I'll try to do it this way then.

lib/connections/limiter.go
```diff
@@ -12,44 +12,130 @@ import (
...
 }
 
 func (lim *limiter) VerifyConfiguration(from, to config.Configuration) error {
 	return nil
 }
 
 func (lim *limiter) CommitConfiguration(from, to config.Configuration) bool {
+	// to ensure atomic update of configuration
+	lim.mu.Lock()
+	defer lim.mu.Unlock()
```
imsodin: Is this necessary? The map mutex should be protection enough, right?

lib/connections/limiter.go
```diff
@@ -97,52 +182,68 @@ func (lim *limiter) String() string {
 	return "connections.limiter"
 }
 
+func (lim *limiter) newReadLimiter(remoteID protocol.DeviceID, r io.Reader, isLAN bool) io.Reader {
```
imsodin: While you're changing this: Wouldn't `newLimitedReader` be a more appropriate name? (same for the writer below)

lib/connections/limiter.go
```diff
@@ -12,44 +12,130 @@ import (
...
-func (lim *limiter) newWriteLimiter(w io.Writer, isLAN bool) io.Writer {
-	return &limitedWriter{writer: w, limiter: lim, isLAN: isLAN}
+// This function handles removing, adding and updating of device limiters.
+// Pass pointer to avoid copying. Pointer already points to copy of configuration
+// so we don't have to worry about modifying real config.
+func (lim *limiter) processDevicesConfiguration(from, to config.Configuration) {
+	seen := make(map[protocol.DeviceID]struct{})
+
+	// Mark devices which should not be removed, create new limiters if needed and assign new limiter rate
+	for _, v := range to.Devices {
```
imsodin: What does "v" stand for? We usually use "dev".

imsodin:
Besides nitpicks this looks good to me and I believe @AudriusButkevicius request has been included as well.

lib/connections/limiter.go
```diff
@@ -162,3 +263,37 @@ func (b *atomicBool) set(v bool) {
 func (b *atomicBool) get() bool {
 	return atomic.LoadInt32((*int32)(b)) != 0
 }
+
+// Utility functions for atomic operations on device limiters map
+func (lim *limiter) getOrCreateDeviceWriteLimiter(deviceID protocol.DeviceID) (*rate.Limiter, bool) {
```
imsodin: Personally I would drop these getOrCreate functions and do what is necessary directly where called. This is just in two places, and in one place you already know that no device limiter exists yet. However that's a style question and I am quite the contrary of an authority on style, so don't act on my suggestion unless somebody else (or you yourself) agree with it.

AudriusButkevicius:
Time for this on sunday hopefully.

lib/connections/limiter.go
```diff
@@ -78,17 +163,17 @@ func (lim *limiter) CommitConfiguration(from, to config.Configuration) bool {
 	if to.Options.MaxSendKbps > 0 {
 		sendLimitStr = fmt.Sprintf("limit is %d KiB/s", to.Options.MaxSendKbps)
 	}
+
```
imsodin: separation doesn't seem helpful here

lib/connections/limiter.go
```diff
@@ -97,52 +182,68 @@ func (lim *limiter) String() string {
...
 }
 
-// take is a utility function to consume tokens from a rate.Limiter. No call
-// to WaitN can be larger than the limiter burst size so we split it up into
+// take is a utility function to consume tokens from a overall rate.Limiter and deviceLimiter.
+// No call to WaitN can be larger than the limiter burst size so we split it up into
 // several calls when necessary.
-func take(l *rate.Limiter, tokens int) {
+func take(l, deviceLimiter *rate.Limiter, tokens int) {
+
```
imsodin: empty line

lib/connections/limiter.go
```diff
@@ -78,17 +163,17 @@ func (lim *limiter) CommitConfiguration(from, to config.Configuration) bool {
...
 	}
-	l.Infof("Send rate %s, receive rate %s", sendLimitStr, recvLimitStr)
+	l.Infof("Overall send rate %s, receive rate %s", sendLimitStr, recvLimitStr)
 
 	if to.Options.LimitBandwidthInLan {
 		l.Infoln("Rate limits apply to LAN connections")
 	} else {
 		l.Infoln("Rate limits do not apply to LAN connections")
 	}
-
```
imsodin: We usually separate return with an empty line, so I'd not remove it here.

lib/connections/limiter_test.go
```diff
@@ -0,0 +1,202 @@
...
+	dev4Conf = config.NewDeviceConfiguration(device4, "device4")
+
+	dev2Conf.MaxRecvKbps = rand.Int() % 100000
+	dev2Conf.MaxSendKbps = rand.Int() % 100000
+
+	waiter, _ := cfg.SetDevices([]config.DeviceConfiguration{dev1Conf, dev2Conf, dev3Conf, dev4Conf})
+	waiter.Wait()
+}
+
+func checkActualAndExpected(actualR, actualW, expectedR, expectedW map[protocol.DeviceID]*rate.Limiter, t *testing.T) {
```
AudriusButkevicius: add t.Helper() here, and pass t as the first argument.

lib/connections/limiter.go
```diff
@@ -162,3 +263,37 @@ func (b *atomicBool) set(v bool) {
...
+
+	if !ok {
+		limiter = rate.NewLimiter(rate.Inf, limiterBurstSize)
+		lim.deviceReadLimiters[deviceID] = limiter
+	}
+
+	return limiter, ok
+}
+
+func (lim *limiter) deleteDeviceLimiters(deviceID protocol.DeviceID) {
```
AudriusButkevicius: This is literally used in one place, just embed this code in that place and be done with it.

lib/connections/limiter.go
```diff
@@ -162,3 +263,37 @@ func (b *atomicBool) set(v bool) {
...
+
+	if !ok {
+		limiter = rate.NewLimiter(rate.Inf, limiterBurstSize)
+		lim.deviceWriteLimiters[deviceID] = limiter
+	}
+
+	return limiter, ok
+}
+
+func (lim *limiter) getOrCreateDeviceReadLimiter(deviceID protocol.DeviceID) (*rate.Limiter, bool) {
```
AudriusButkevicius: The name is terrible.
Just getReadLimiter() or even readLimiter(), and the fact that it creates it underneath is an irrelevant detail that doesn't need to be exposed as part of the function name.

lib/connections/limiter.go
```diff
@@ -162,3 +263,37 @@ func (b *atomicBool) set(v bool) {
...
+	lim.deviceMapMutex.Lock()
+	defer lim.deviceMapMutex.Unlock()
+	limiter, ok := lim.deviceWriteLimiters[deviceID]
+
+	if !ok {
+		limiter = rate.NewLimiter(rate.Inf, limiterBurstSize)
+		lim.deviceWriteLimiters[deviceID] = limiter
+	}
+
+	return limiter, ok
```
AudriusButkevicius: I don't think this boolean is needed, all you care about when deciding to print or not print the message about rates is whether the rates are different from what they were before.

lib/connections/limiter.go
AudriusButkevicius: Agree

lib/connections/limiter.go
AudriusButkevicius: Agree

lib/connections/limiter.go
AudriusButkevicius: agree

lib/connections/limiter.go
```diff
@@ -12,44 +12,130 @@ import (
...
+
+	// Delete remote devices which were removed in new configuration
+	for _, v := range from.Devices {
+		if _, ok := seen[v.DeviceID]; !ok {
+			l.Debugf("deviceID: %s should be removed", v.DeviceID)
+			lim.deleteDeviceLimiters(v.DeviceID)
+		}
+	}
+
+	l.Debugln("Rebuild of device limiters map finished")
```
AudriusButkevicius: This is a lie, as I don't think we're rebuilding stuff here.

lib/connections/limiter.go
```diff
@@ -12,44 +12,130 @@ import (
...
+		}
+		if v.MaxRecvKbps <= 0 {
+			currentWriteLimit = rate.Inf
+		}
+
+		// Nothing about this device has changed. Start processing next device
+		if okR && okW && currentReadLimit == previousReadLimit && currentWriteLimit == previousWriteLimit {
+			continue
+		}
+		setLimitsForDevice(v, readLimiter, writeLimiter)
```
AudriusButkevicius: Would it make sense to have
```
func (limiter) setLimits(device, readRate int, writeRate int) {
}
```

that then reaches into the maps, gets the limiters and what not.

lib/connections/limiter.go
AudriusButkevicius: I can imagine stuff getting out of sync a bit if two saves happen in parallel, but to be honest, I am not worried about a case like that.

qepasa: That's why I added mutex at the beginning of CommitConfiguration call

lib/connections/limiter.go
qepasa: I guess when many routines try to save configs at the same time it could cause some trouble when handling device limiters. I.e. first we remove device X then add the same device X and then update of device X rates. Routines could get scheduled so that remove stops right before the loop where deleting is done. Then add would process the device and next config would update it's rates. After removing is resumed we're left with no limiter for X despite it was set it in the most recent config.

qepasa:
@AudriusButkevicius Right now UI portion looks like that: https://imgur.com/a/l238P
I guess it automatically created those fields based on device config json. If you'd like to see it done differently just let me know.

lib/connections/limiter.go
imsodin: Right. To me it seems redundant to have two locks protecting the limiter. Couldn't you just have a single lock that you acquire during `CommitConfiguration` and `newLimitedReader/Writer` (and then obviously not acquiring it in getReadLimiter)?

lib/connections/limiter.go
```diff
@@ -12,44 +12,136 @@ import (
...
 	prev := config.Configuration{Options: config.OptionsConfiguration{MaxRecvKbps: -1, MaxSendKbps: -1}}
+
 	l.CommitConfiguration(prev, cfg.RawCopy())
 	return l
 }
 
-func (lim *limiter) newReadLimiter(r io.Reader, isLAN bool) io.Reader {
-	return &limitedReader{reader: r, limiter: lim, isLAN: isLAN}
+// This function sets limiters according to corresponding DeviceConfiguration
+func (lim *limiter) setLimits(device config.DeviceConfiguration) {
```
AudriusButkevicius: There is an awful lot of rate conversion happening here and in processDeviceConfigurations.
Perhaps this setLimits function could return a boolean stating whether or not the rates have changed, and then processDeviceConfigurations just needs to print stuff without having to do conversions by itself.

qepasa:
I added suggested fixes. Thank you guys for patience with this pull request and for quick reviews.

lib/connections/limiter.go
```diff
@@ -82,48 +87,28 @@ func (lim *limiter) processDevicesConfiguration(from, to config.Configuration) {
...
 		// Nothing about this device has changed. Start processing next device
-		if currentReadLimit == previousReadLimit && currentWriteLimit == previousWriteLimit {
-			continue
-		}
-		lim.setLimits(dev)
-
-		readLimitStr := "is unlimited"
-		if dev.MaxRecvKbps > 0 {
-			readLimitStr = fmt.Sprintf("limit is %d KiB/s", dev.MaxRecvKbps)
+		if limitsChanged := lim.setLimits(dev); limitsChanged {
```
imsodin: This can just be `if lim.setLimits(dev) {`.

lib/connections/limiter.go
```diff
@@ -97,52 +173,71 @@ func (lim *limiter) String() string {
...
+	defer lim.mu.Unlock()
+	deviceLimiter := lim.getReadLimiter(remoteID)
+	return &limitedReader{reader: r, limiter: lim, deviceLimiter: deviceLimiter, isLAN: isLAN}
+}
+
+func (lim *limiter) newLimitedWriter(remoteID protocol.DeviceID, w io.Writer, isLAN bool) io.Writer {
+	lim.mu.Lock()
+	defer lim.mu.Unlock()
+	deviceLimiter := lim.getWriteLimiter(remoteID)
+	return &limitedWriter{writer: w, limiter: lim, deviceLimiter: deviceLimiter, isLAN: isLAN}
```
AudriusButkevicius: The fact that processblabla does locking outside the function, and this does inside, feels wrong.

lib/connections/limiter.go
```diff
@@ -12,44 +12,121 @@ import (
...
+	// Mark devices which should not be removed, create new limiters if needed and assign new limiter rate
+	for _, dev := range to.Devices {
+		if dev.DeviceID == to.MyID {
+			// This limiter was created for local device. Should skip this device
+			continue
+		}
+		seen[dev.DeviceID] = struct{}{}
+
+		// Nothing about this device has changed. Start processing next device
+		if limitsChanged := lim.setLimits(dev); limitsChanged {
```
AudriusButkevicius: The variable assignment is pointless, you can straight do if setLimits(dev) {

lib/connections/limiter_test.go
```diff
@@ -0,0 +1,204 @@
...
+		addedDevice: rate.NewLimiter(rate.Limit(addDevConf.MaxSendKbps*1024), limiterBurstSize),
+	}
+	actualR := lim.deviceReadLimiters
+	actualW := lim.deviceWriteLimiters
+
+	checkActualAndExpected(t, actualR, actualW, expectedR, expectedW)
+}
+
+func initConfig(t *testing.T) {
+	t.Helper()
```
AudriusButkevicius: Don't need t.helper as you are not asserting anything here

lib/connections/limiter.go
imsodin: We are starting to contradict each other once again x-): My reasoning to suggest this was that the locking now happens in `CommitConfiguration` and in these `newLimited...` functions, which are all "top-level", i.e. called "externally" (not all outside of the package, but outside of the `limiter` code). Also locking the entirety of `CommitConfiguration` was considered necessary to disallow concurrent change and thus with a single lock the locking cannot happen on `get...Limiter` anymore (single lock is not a requirement obviously, I just consider it cleaner).

AudriusButkevicius: Fine, but why can't we lock lim.mu before calling the newblabla twice in the connection service, this way we lock the same way in both  places. Also a function that needs locking before called, we should add the locked word into the name like we do in the model.

imsodin: Now I see.  
@qepasa So locking should happen [here](https://github.com/syncthing/syncthing/pull/4603/files/6967960f5032d4fa89843731f945bfd8f4daceef#diff-d31f92d68068bfede5c6b6b01e1b558bR268) instead and a whole bunch of functions needs the `Locked` suffix: `setLimits`, `new...`, `get...`, `processD...`.

lib/connections/limiter_test.go
```diff
@@ -169,8 +169,7 @@ func TestAddAndRemove(t *testing.T) {
 	checkActualAndExpected(t, actualR, actualW, expectedR, expectedW)
 }
 
-func initConfig(t *testing.T) {
-	t.Helper()
+func initConfig() {
```
AudriusButkevicius: This can just return a config, rather than setting some global that leaks between the tests.

AudriusButkevicius:
Implementation wise looks ok to me, needs to resolve merge conflicts and pass the tests (which I am not sure if they are failing due to anything this PR is doing). 
Also perhaps needs UI to expose this in device editor.

imsodin:
> Also perhaps needs UI to expose this in device editor.

I agree that this is probably a nice enough functionality to get a place there. If @qepasa wants to do it in this PR, cool, but otherwise I think it's fine to leave it in advanced settings and open an enhancement issue for someone else (or qepasa) to do that eventually.

lib/connections/limiter.go
```diff
@@ -12,44 +12,121 @@ import (
 	"sync/atomic"
 
 	"github.com/syncthing/syncthing/lib/config"
+	"github.com/syncthing/syncthing/lib/protocol"
 	"golang.org/x/net/context"
 	"golang.org/x/time/rate"
+	"sync"
```
calmh: This should probably be our `lib/sync`

lib/connections/limiter.go
```diff
@@ -12,44 +12,121 @@ import (
...
 type limiter struct {
-	write     *rate.Limiter
-	read      *rate.Limiter
-	limitsLAN atomicBool
+	write               *rate.Limiter
+	read                *rate.Limiter
+	limitsLAN           atomicBool
+	deviceReadLimiters  map[protocol.DeviceID]*rate.Limiter
+	deviceWriteLimiters map[protocol.DeviceID]*rate.Limiter
+	mu                  *sync.Mutex
```
calmh: Just `sync.Mutex`, no need for a pointer (regardless of which sync package)

lib/connections/limiter.go
```diff
@@ -12,44 +12,121 @@ import (
...
 
 const limiterBurstSize = 4 * 128 << 10
 
 func newLimiter(cfg *config.Wrapper) *limiter {
 	l := &limiter{
-		write: rate.NewLimiter(rate.Inf, limiterBurstSize),
-		read:  rate.NewLimiter(rate.Inf, limiterBurstSize),
+		write:               rate.NewLimiter(rate.Inf, limiterBurstSize),
+		read:                rate.NewLimiter(rate.Inf, limiterBurstSize),
+		mu:                  &sync.Mutex{},
```
calmh: Adjust here to nothing at all (standard lib `sync`) or `sync.NewMutex()` (our `sync`)

lib/connections/limiter.go
```diff
@@ -12,44 +12,121 @@ import (
...
+	readLimiter.SetLimit(currentReadLimit)
+	writeLimiter.SetLimit(currentWriteLimit)
+
+	return true
 }
 
-func (lim *limiter) newWriteLimiter(w io.Writer, isLAN bool) io.Writer {
-	return &limitedWriter{writer: w, limiter: lim, isLAN: isLAN}
+// This function handles removing, adding and updating of device limiters.
+// Pass pointer to avoid copying. Pointer already points to copy of configuration
```
calmh: Not clear what the pointer comments are about. The configuration parameters are not pointers.

lib/connections/service.go
```diff
@@ -265,8 +265,10 @@ next:
 		// keep up with config changes to the rate and whether or not LAN
 		// connections are limited.
 		isLAN := s.isLAN(c.RemoteAddr())
-		wr := s.limiter.newWriteLimiter(c, isLAN)
-		rd := s.limiter.newReadLimiter(c, isLAN)
+		s.limiter.mu.Lock()
```
calmh: This is unusual, we do don't want to take locks on mutexes inside other objects and then call their `...Locked` functions. Wrap in a method.

`rd, wr := s.limiter.getLimiters(remoteID, c, isLAN)`

imsodin:
> [...] all of this juggling of mutexes and maps and stuff does not seem worth it to me compared to just creating the limiter with the appropriate settings at connection time. Yeah then we can't change the per device rate limit on the fly, but this will almost never happen and we anyway disconnect peers most of the time when we change folders or devices. [...]

@calmh Isn't this dropping of connection one of the long-standing issues nobody wants to tackle because it is so entangled? What I mean is that not needing a "complete reset" and thus at least not increasing the entanglement, seems to already provide quite a bit of merit.

calmh:
LGTM apart from above mutex nits, and that the parameters to `take` could have better names (`l` being `globalLimiter` or something maybe).

Has anyone tested that this actually works?

lib/connections/limiter.go
```diff
@@ -173,6 +169,14 @@ func (lim *limiter) String() string {
 	return "connections.limiter"
 }
 
+func (lim *limiter) getLimiters(remoteID protocol.DeviceID, c internalConn, isLAN bool) (io.Reader, io.Writer) {
+	lim.mu.Lock()
+	defer lim.mu.Unlock()
```
AudriusButkevicius: no point using here defer as the scope is small, as this just prevents potential inlining.

lib/connections/limiter.go
```diff
@@ -97,53 +169,76 @@ func (lim *limiter) String() string {
...
+}
+
+func (lim *limiter) newLimitedReaderLocked(remoteID protocol.DeviceID, r io.Reader, isLAN bool) io.Reader {
+	deviceLimiter := lim.getReadLimiterLocked(remoteID)
+	return &limitedReader{reader: r, limiter: lim, deviceLimiter: deviceLimiter, isLAN: isLAN}
+}
+
+func (lim *limiter) newLimitedWriterLocked(remoteID protocol.DeviceID, w io.Writer, isLAN bool) io.Writer {
+	deviceLimiter := lim.getWriteLimiterLocked(remoteID)
+	return &limitedWriter{writer: w, limiter: lim, deviceLimiter: deviceLimiter, isLAN: isLAN}
```
AudriusButkevicius: Merge these two two line functions into the callee to reduce the mental stack depth.

qepasa:
I can try to make UI for this but I would prefer it to be a new issue.

AudriusButkevicius:
@imsodin @calmh pls re-review and lets put this to bed.

The person has put an immense amount of effort for dealing with my (our?) comments so far, and we just left him in silence. I am happy with this.

calmh:
I have nits to pick, but i'll pick them myself in a followup at some point

PR #4604, opened 2017-12-17
gui: Add log viewer (fixes #2644)

### LINKED ISSUE #2644: Actual debug UI
We have the REST parts for setting debug levels and accessing log data, but no UI for it. We need something that allows setting the various debug facilities, view and download the log data. It should probably be on a page by itself.

AudriusButkevicius:
Just because there are a ton of people struggling to set env vars.
Perhaps this should live in advanced settings, and perhaps we should have a button to view the log.

image

### DIFF
--- a/gui/default/syncthing/core/logViewerModalView.html
+++ b/gui/default/syncthing/core/logViewerModalView.html
@@ -1,0 +1,35 @@
+<modal id="logViewer" icon="book" status="default" heading="{{'Logs' | translate}}" large="yes" closeable="yes">
+  <div class="modal-body">
+    <ul class="nav nav-tabs">
+      <li class="active"><a data-toggle="tab" href="#log-viewer-log" translate>Log</a></li>
+      <li><a data-toggle="tab" href="#log-viewer-facilities" translate>Debugging Facilities</a></li>
+    </ul>
+    <div class="tab-content">
+
+      <div id="log-viewer-log" class="tab-pane in active">
+        <label translate ng-if="logging.logEntries.length == 0">Loading...</label>
+        <textarea class="form-control" rows="20" ng-if="logging.logEntries.length != 0" readonly style="font-family: Consolas; font-size: 11px; overflow: auto;">{{ logging.content() }}</textarea>
+      </div>
+
+      <div id="log-viewer-facilities" class="tab-pane">
+        <label>Available debug logging facilities:</label>
+        <table class="table table-condensed table-striped">
+          <tbody>
+            <tr ng-repeat="(name, data) in logging.facilities">
+              <td>
+                <input type="checkbox" ng-model="data.enabled" ng-change="logging.onFacilityChange(name)" ng-disabled="data.enabled == null"> <span>{{ name }}</span>
+              </td>
+              <td>{{ data.description }}</td>
+            </tr>
+          </tbody>
+        </table>
+      </div>
+
+    </div>
+  </div>
+  <div class="modal-footer">
+    <button type="button" class="btn btn-default btn-sm" data-dismiss="modal">
+      <span class="fa fa-times"></span>&nbsp;<span translate>Close</span>
+    </button>
+  </div>
+</modal>
--- a/gui/default/syncthing/core/logViewerModalView.html
+++ b/gui/default/syncthing/core/logViewerModalView.html
@@ -8,7 +8,8 @@
 
       <div id="log-viewer-log" class="tab-pane in active">
         <label translate ng-if="logging.logEntries.length == 0">Loading...</label>
-        <textarea class="form-control" rows="20" ng-if="logging.logEntries.length != 0" readonly style="font-family: Consolas; font-size: 11px; overflow: auto;">{{ logging.content() }}</textarea>
+        <textarea ng-focus="logging.paused = true" ng-blur="logging.paused = false" id="logViewerText" class="form-control" rows="20" ng-if="logging.logEntries.length != 0" readonly style="font-family: Consolas; font-size: 11px; overflow: auto;">{{ logging.content() }}</textarea>
+        <p translate class="help-block" ng-style="{'visibility': logging.paused ? 'visible' : 'hidden'}">Log tailing paused. Click here to continue</p>
       </div>
 
       <div id="log-viewer-facilities" class="tab-pane">
--- a/gui/default/syncthing/core/logViewerModalView.html
+++ b/gui/default/syncthing/core/logViewerModalView.html
@@ -9,7 +9,7 @@
       <div id="log-viewer-log" class="tab-pane in active">
         <label translate ng-if="logging.logEntries.length == 0">Loading...</label>
         <textarea ng-focus="logging.paused = true" ng-blur="logging.paused = false" id="logViewerText" class="form-control" rows="20" ng-if="logging.logEntries.length != 0" readonly style="font-family: Consolas; font-size: 11px; overflow: auto;">{{ logging.content() }}</textarea>
-        <p translate class="help-block" ng-style="{'visibility': logging.paused ? 'visible' : 'hidden'}">Log tailing paused. Click here to continue</p>
+        <p translate class="help-block" ng-style="{'visibility': logging.paused ? 'visible' : 'hidden'}">Log tailing paused. Click here to continue.</p>
       </div>
 
       <div id="log-viewer-facilities" class="tab-pane">
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1091,6 +1091,19 @@ angular.module('syncthing.core')
         };
 
         $scope.editSettings = function () {
+            // Get logging debug options
+            $http.get(urlbase + '/system/debug').success(function (data) {
+                var debug = {};
+                data.enabled = data.enabled || [];
+                $.each(data.facilities, function(key, value) {
+                    debug[key] = {
+                        description: value,
+                        enabled: data.enabled.indexOf(key) > -1
+                    }
+                })
+                $scope.debug = debug;
+            }).error($scope.emitHTTPError);
+
             // Make a working copy
             $scope.tmpOptions = angular.copy($scope.config.options);
             $scope.tmpOptions.deviceName = $scope.thisDevice().name;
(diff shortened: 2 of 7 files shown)


canton7:
Awesome! A big +1 to having a button which shows the log -- that's going to be far easier for users to grok than "figure out where, if anywhere, your system is logging to".

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1091,6 +1091,19 @@ angular.module('syncthing.core')
         };
 
         $scope.editSettings = function () {
+            // Get logging debug options
+            $http.get(urlbase + '/system/debug').success(function (data) {
+                var debug = {};
+                data.enabled = data.enabled || [];
+                $.each(data.facilities, function(key, value) {
```
imsodin: Is this the same as `data.facilities.forEach`? Just asking because `$.each` isn't used in syncthingController.js anywhere else.

lib/fs/debug.go
```diff
@@ -14,9 +14,9 @@ import (
 )
 
 var (
-	l = logger.DefaultLogger.NewFacility("filesystem", "Filesystem access")
+	l = logger.DefaultLogger.NewFacility("fs", "Filesystem access")
```
imsodin: This needs to be changed in the help text in main too (and subsequently in docs).

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: Same thing. forEach I think is less cross browsery.

lib/fs/debug.go
AudriusButkevicius: Help text in main is generated from available facilities.

AudriusButkevicius:
Now displays logs, and "tails" them by polling every 2s.

image
image
image

imsodin:
The log view keeps re-writing tags from the beginning (cmdline confirms it isn't actually restarting):

screenshot_2017-12-19_09-37-46

Also the trailing is very jumpy and doesn't actually follow the last line, but that may be related to the above problem. I think (de-)activating trailing should be done with a button that toggles it (while still also disabling when clicking into the text area).

What is the intention of the log view? If it is meant as the main source when asking users for logs, 250 lines might not be enough and a button to copy or save the log would be neat. To not increase memory consumption by raising the limit, maybe there could be an additional debug config option to keep all the logs.

AudriusButkevicius:
It's rewriting logs if you haven't built syncthing, as there is a bug in the since parsing logic. I guess when we tell people to look at the log, they atleast can glimpse at the last 250 lines.

You can open the log and keep it open for a while I guess, but I suspect this would only be used to peek at the last few lines.

imsodin:
I have rebuilt before and now with clean & assets for good measure, the problem with repeating logs and trailing not working remains (nothing in the JS console).

AudriusButkevicius:
Do the http calls have a since query string?

Whats the issue with trailing apart from you not liking the way you enable disable it?

AudriusButkevicius:
Also, the message is a red herring. Essentially, it stops tailing when the field gains focus, and the message comes up as 'click on me' just for the user to lose focus on the textarea (clicking the text doesn't actually do anything).

imsodin:
I am confused: In JS console it looks like the `last` variable is correct, but in gui.go the since query string is empty. When I try from command line with the example for RFC3339 given in Go's docs like this:
```
 curl -X GET -H "X-API-Key: *" "localhost:8384/rest/system/log?since=2006-01-02T15:04:05Z07:00'
```
it fails to parse it (custom debug log after time.Parse):
```
[CBBHG] 2017/12/19 12:18:32.604678 gui.go:961: INFO: getSystemlog q.Get("since"): 2006-01-02T15:04:05Z07:00 err: parsing time "2006-01-02T15:04:05Z07:00": extra text: 07:00
```

The "test case" 
```golang
	t, err := time.Parse(time.RFC3339, "2006-01-02T15:04:05Z07:00")
	fmt.Println(t, err)
```
also results in
```
0001-01-01 00:00:00 +0000 UTC parsing time "2006-01-02T15:04:05Z07:00": extra text: 07:00.
```
What is wrong here?

AudriusButkevicius:
Strange that it can generate timestamps with that signature yet not parse them, anyway, that's definately a bug.

AudriusButkevicius:
I guess I could replace the text with "Click anywhere outside of the text box to continue tailing the log"

calmh:
"Z07:00" is definitely a broken time zone specifier, so wherever that is generated should be fixed.

imsodin:
> Strange that it can generate timestamps with that signature yet not parse them, anyway, that's definately a bug.

Nah, it's me non-obvious format notation combined with me not RTFM (thoroughly). The time zone is either `Z` or `+-xx:xx`, so `Z07:00` is invalid.

That's actually a pretty annoying format for html queries, because `+` gets converted to ` ` -> error again. Therefore we need to either convert to UTC or encode the timezone `+` as `%2B`. Also the

canton7:
Well, anything you put in a query string should be automatically escaped by whatever library you're using to do the query, so that shouldn't be an issue...

imsodin:
How does the library know whether you mean a literal + or not? Curl and angularjs (I am assume `$http.get` is angularjs) don't do it.

Also I would propose to change the log rest functions such that a problem like this is a bit more obvious:

```diff
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -955,20 +955,36 @@ func (s *apiService) postSystemErrorClear(w http.ResponseWriter, r *http.Request
 
 func (s *apiService) getSystemLog(w http.ResponseWriter, r *http.Request) {
 	q := r.URL.Query()
-	since, err := time.Parse(time.RFC3339, q.Get("since"))
-	l.Debugln(err)
+	var t time.Time
+	if since := q.Get("since"); since != "" {
+		var err error
+		if t, err = time.Parse(time.RFC3339, since); err != nil {
+			l.Debugln(err)
+			http.Error(w, err.Error(), 400)
+			return
+		}
+	}
+
 	sendJSON(w, map[string][]logger.Line{
-		"messages": s.systemLog.Since(since),
+		"messages": s.systemLog.Since(t),
 	})
 }
 
 func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
 	q := r.URL.Query()
-	since, err := time.Parse(time.RFC3339, q.Get("since"))
-	l.Debugln(err)
+	var t time.Time
+	if since := q.Get("since"); since != "" {
+		var err error
+		if t, err = time.Parse(time.RFC3339, since); err != nil {
+			l.Debugln(err)
+			http.Error(w, err.Error(), 400)
+			return
+		}
+	}
+
 	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
 
-	for _, line := range s.systemLog.Since(since) {
+	for _, line := range s.systemLog.Since(t) {
 		fmt.Fprintf(w, "%s: %s\n", line.When.Format(time.RFC3339), line.Message)
 	}
 }
```

canton7:
Apparently you should use encodeURIComponent with $http.get.

> How does the library know whether you mean a literal + or not?

You pass in some values. They're url encoded. The server url decodes them, and gets your values. If you pass in a `+`, that's urlencoded to `%2B` and the server decodes it to `+`. If you pass in a ` `, that's urlencoded to `+` and the server decodes it to ` `.

imsodin:
That's it - using `encodeURIComponent` on `last` the log display (including trailing) works smoothly here.

gui/default/syncthing/core/logViewerModalView.html
```diff
@@ -0,0 +1,36 @@
...
+    <ul class="nav nav-tabs">
+      <li class="active"><a data-toggle="tab" href="#log-viewer-log" translate>Log</a></li>
+      <li><a data-toggle="tab" href="#log-viewer-facilities" translate>Debugging Facilities</a></li>
+    </ul>
+    <div class="tab-content">
+
+      <div id="log-viewer-log" class="tab-pane in active">
+        <label translate ng-if="logging.logEntries.length == 0">Loading...</label>
+        <textarea ng-focus="logging.paused = true" ng-blur="logging.paused = false" id="logViewerText" class="form-control" rows="20" ng-if="logging.logEntries.length != 0" readonly style="font-family: Consolas; font-size: 11px; overflow: auto;">{{ logging.content() }}</textarea>
+        <p translate class="help-block" ng-style="{'visibility': logging.paused ? 'visible' : 'hidden'}">Log tailing paused. Click here to continue</p>
```
imsodin: The click on me as text seems fine and better than the new "Click anywhere..." proposal: Otherwise someone might click on the background and complain when the log modal disappears :P 
I would add a dot (or three) at the end though.

imsodin:
@calmh The mac agent is missing - is it a planned "revision outage" or something unforeseen? More to the point: Will it be up soon(ish) or should we start skipping that test to merge stuff?

Apart from the very nitpicky comment about punctuation, this is lgtm.

gui/default/syncthing/core/logViewerModalView.html
AudriusButkevicius: Three dots to existing text?

imsodin: To be consistent (first sentence gets a final dot): "Log tailing paused. Click here to continue."  
On second thought, I don't like three dots ("continue..."). The idea was, that it signifies that something (trailing) will continue when you click, but that seems unnecessary and might be a problem for translation.

calmh:
Mac builder is down because it’s at my home, internet there is apparently down, and I’m currently in Dubai and unable to troubleshoot. I’ll be back in a couple of days and will fix it then.

imsodin:
Translation strings need to be updated.

AudriusButkevicius:
@calmh does that before every release anyway

calmh:
@st-review lgtm

st-review:

calmh:
I think the viewer deserves to be full screen and not a modal but that's a valid future improvement and this is great as is.

AudriusButkevicius:
@imsodin needs to leave his mark.

imsodin:
@st-review lgtm

st-review:
:ok_hand: Merged as c58b383b6df4c718a6ea24f3de6699f9b77d5582. Thanks, @AudriusButkevicius!

PR #4607, opened 2017-12-19
lib/model: Check for invalid filenames after ignore patterns

imsodin:
If files with invalid characters on windows are ignored, they should not cause errors when pulling. Came up [in the forum](https://forum.syncthing.net/t/ignoring-osx-icon-r-files-on-windows/11067/6?u=imsodin).

The first commit contains only the test to verify that it actually reproduces this problem.

### DIFF
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/d4l3k/messagediff"
 	"github.com/syncthing/syncthing/lib/config"
 	"github.com/syncthing/syncthing/lib/db"
+	"github.com/syncthing/syncthing/lib/events"
 	"github.com/syncthing/syncthing/lib/fs"
 	"github.com/syncthing/syncthing/lib/ignore"
 	"github.com/syncthing/syncthing/lib/protocol"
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
Isn't process directly supposed to fail in this case? It seems thats not handled anymore?

imsodin:
@AudriusButkevicius I don't understand what isn't handled anymore here? Deleted files with invalid filenames are added to process directly, but that was also the case before this PR. The only difference is, that now ignored invalid files don't create an error, but get succeed (get ignored/invalidated like a normal ignored file).

AudriusButkevicius:
Right, sorry missed the is not deleted versus deleted bit. This looks ok to me.

calmh:
Can you test the case when we are supposed to delete a file with an invalid name?

I think the test that you removed was broken and not as intended - I remember the intention being "complain about invalid files, unless they are deleted in which case don't process them". I suspect we will currently process and complain about not being able to delete files, which we should not as they couldn't possibly exist.

If I'm wrong on this and the delete failure is silent, LGTM.

calmh:
@st-review lgtm with the above caveat

st-review:

imsodin:
@calmh The behaviour of this PR at the time of your comment was the same as before: Deleted files with invalid names are processed as if the names were valid. This would result in errors due to removal/renaming failing. However this combination can never happen in the puller: We cannot already have a file with invalid filenames and in the db, deleted files that we don't already have are not considered as needed. Therefore I moved the invalid name case before the delete case and added a test case that is dependent on this db behaviour.

calmh:
Cool

imsodin:
Then @AudriusButkevicius needs to leave his mark :)

AudriusButkevicius:
I am not comfortable with the test messing in a common playground. Can we setup a temp dir for this?

imsodin:
Sure, done.

The same should be done for a few existing tests. Maybe it would even be "cleanest" to not have a common playground at all and run every test in its own temporary dir generating any needed files. But that's not for this PR.

AudriusButkevicius:
@st-review lgtm

st-review:
:ok_hand: Merged as 841205dbfe2c819ef432bc8c253598cabebd78a0. Thanks, @imsodin!

PR #4622, opened 2017-12-29
lib/fs: Handle deduplicated files on NTFS (fixes #1845)

### LINKED ISSUE #1845: Files on deduplicated NTFS volumes are ignored by scanner
When adding a folder on a volume with NTFS Deduplication enabled, deduped files are incorrectly skipped by the scanner. Files that are still pending deduplication are processed as expected. Folders are also processed as normal.

A bit of digging indicates that os.Lstat on a deduped file has a FileMode containing ModeSymLink. Looking at the Read function in [symlink_windows.go](https://github.com/syncthing/syncthing/blob/master/internal/symlinks/symlink_windows.go#L104), the type of reparse point is available in reparseData.reparseTag, which should allow for correctly handling the condition. 

A quick reference to the tag values is available in [smbfsctl.h](https://github.com/torvalds/linux/blob/master/fs/cifs/smbfsctl.h#L100). Of interest are:
- IO_REPARSE_TAG_DEDUP = 0x80000013
- IO_REPARSE_TAG_SYMLINK = 0xA000000C
- IO_REPARSE_TAG_MOUNT_POINT = 0xA0000003 (may be related to #1830)
- More info about [Reparse Points](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365511%28v=vs.85%29.aspx) and supported operations

For reference, it is possible to query a reparse point's tag and data with <code>fsutil reparsepoint query PATH_TO_FILE</code>. This even works across SMB shares with UNC paths.

calmh:
These files always have the symlink bit set, because they are reparse points. Nonetheless they are not symlinks, and Lstat reports a size for them. We use this fact to disambiguate, and hope fervently that nothing else matches this description so it comes back to bite us...

### Testing

@niphlod tested:

screen shot 2017-12-29 at 20 42 40


lib/fs/fsfileinfo_windows.go
```diff
@@ -0,0 +1,34 @@
...
+
+package fs
+
+import (
+	"os"
+)
+
+func (e fsFileInfo) Mode() FileMode {
+	m := e.FileInfo.Mode()
+	if e.Size() > 0 {
```
AudriusButkevicius: Checking if symlink bit it set first is cheaper, only then check for size.

lib/fs/fsfileinfo_windows.go
```diff
@@ -0,0 +1,34 @@
...
+		return true
+	}
+	return e.FileInfo.Mode().IsRegular()
+}
+
+func (e fsFileInfo) IsSymlink() bool {
+	if e.Size() > 0 {
+		return false
+	}
+	return e.FileInfo.Mode()&os.ModeSymlink == os.ModeSymlink
```
AudriusButkevicius: Do you even need this? Can't we just call e.Mode() instead of e.FileInfo.Mode()?

lib/fs/fsfileinfo_windows.go
```diff
@@ -0,0 +1,34 @@
...
+		m &^= os.ModeSymlink
+	}
+	return FileMode(m)
+}
+
+func (e fsFileInfo) IsRegular() bool {
+	if e.Size() > 0 {
+		return true
+	}
+	return e.FileInfo.Mode().IsRegular()
```
AudriusButkevicius: Do you even need this? Can't we just call e.Mode() instead of e.FileInfo.Mode()?

lib/fs/fsfileinfo_windows.go
```diff
@@ -0,0 +1,38 @@
...
+		// makes them non-regular.
+		m := e.FileInfo.Mode() &^ os.ModeSymlink
+		return m.IsRegular()
+	}
+	return e.FileInfo.Mode().IsRegular()
+}
+
+func (e fsFileInfo) IsSymlink() bool {
+	// N.B. this uses *our* Mode() which appropriately masks this bit.
+	return e.Mode()&FileMode(os.ModeSymlink) != 0
```
AudriusButkevicius: Just ModeSymlink, which is already defined in fs.go.

calmh: It actually isn't but maybe it should be

calmh:
There, now I actually read your comments Audrius.

lib/fs/fsfileinfo_windows.go
```diff
@@ -0,0 +1,38 @@
...
+func (e fsFileInfo) IsRegular() bool {
+	if e.FileInfo.Mode()&os.ModeSymlink != 0 && e.Size() > 0 {
+		// Take out the symlink bit and see if it's a regular *now*... This
+		// is a bit of extra carefulness against things that might have the
+		// symlink bit set, have a size, yet have some other bit set that
+		// makes them non-regular.
+		m := e.FileInfo.Mode() &^ os.ModeSymlink
+		return m.IsRegular()
+	}
+	return e.FileInfo.Mode().IsRegular()
```
AudriusButkevicius: Perhaps:
```
m := e.FileInfo.Mode()
if m&.ModeSymlink && .. .{
    m ^= ModeSymlink
}

return m.IsRegular()

calmh:
Good to go

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as a9f0659f2f4bf910f82b652fd27a864074ec7ab8. Thanks, @calmh!

PR #4629, opened 2017-12-31
WIP: Lower priority

calmh:
No idea if this works yet


AudriusButkevicius:
This should definately live in osutil.

cmd/syncthing/ioprio_linux.go
```diff
@@ -0,0 +1,31 @@
...
+const (
+	IOPRIO_CLASS_RT ioprioClass = iota + 1
+	IOPRIO_CLASS_BE
+	IOPRIO_CLASS_IDLE
+)
+
+const (
+	IOPRIO_WHO_PROCESS = iota + 1
+	IOPRIO_WHO_PGRP
+	IOPRIO_WHO_USER
```
AudriusButkevicius: Correct go way would be IoprioWhoProcess

lib/config/optionsconfiguration.go
```diff
@@ -143,6 +143,7 @@ type OptionsConfiguration struct {
 	KCPSendWindowSize       int                     `xml:"kcpSendWindowSize" json:"kcpSendWindowSize" default:"128"`
 	KCPReceiveWindowSize    int                     `xml:"kcpReceiveWindowSize" json:"kcpReceiveWindowSize" default:"128"`
 	DefaultFolderPath       string                  `xml:"defaultFolderPath" json:"defaultFolderPath" default:"~"`
+	SetLowPriority          bool                    `xml:"setLowPriority" json:"setLowPriority" default:"true"`
```
AudriusButkevicius: I'd make this an int, 0 means normal, 1 means BE towards the end, 2 means IDLE etc.

cmd/syncthing/ioprio_linux.go
calmh: syscall constants are typically not mangled from their original though, neither in the syscall package nor by us historically

lib/config/optionsconfiguration.go
calmh: Meh, do not want. I think we should just provide a reasonable default and let the user do the setup themselves if they care more than that. The only reason for a config at all is to have an off switch to avoid screwing with the potential user settings...

cmd/syncthing/ioprio_linux.go
AudriusButkevicius: fine.

calmh:
So this does not in fact work well on any system. The APIs affect *threads* and not processes, and we have no thread control nor the ability to run this from all threads. Windows complicates matters further by warning against deadlocks when low prio threads hold locks that high prio threads need. As an example, doing what I do in this PR on Linux, where niceness could in theory affect the whole process, gives me:

```
jb@acro:~ $ ps axms
...
  502 27317 0000000000000000                -                -                - -    pts/4      0:03 ./syncthing -home
  502     - 0000000000000000 fffffffe3bfa3a00 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:02 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff SNl+ -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
  502     - 0000000000000000 0000000000000000 0000000000000000 fffffffe7fc1feff Sl+  -          0:00 -
```

That's one thread with the `N` flag and a bunch of them without.

PR #4630, opened 2018-01-01
all: Disable a few flaky tests

calmh:
### Purpose

Gnnh

### DIFF
--- a/lib/events/events_test.go
+++ b/lib/events/events_test.go
@@ -212,6 +212,8 @@ func TestSubscriptionIDs(t *testing.T) {
 }
 
 func TestBufferedSub(t *testing.T) {
+	t.Skip("too flaky //jb")
+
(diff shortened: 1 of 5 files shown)


imsodin:
I guess your no fan of disabling those on slow build platforms only? They only ever fail on that one.

calmh:
It's apparently like whackamole anyway

calmh:
But no, not really. Tests depending on how fast the builder is is not great, regardless if just one of the builders is below the threshold at the moment. As all the fs tests use the same testScenario rig I guess all of them are about equally susceptible to this?

calmh:
fwiw the Go 1.10 test caching will make this issue moot as the tests won't be rerun nearly as often, assuming we make them compatible

calmh:
ok fuckit

PR #4632, opened 2018-01-01
lib/fs: harmonize CreateSymlink definitions (fixes #4567)

### LINKED ISSUE #4567: Arguments to fs.CreateSymlink are mixed up
In the interface and everywhere else, it is name then target, except for basicfs on unix, where it is the other way around:

```
lib/fs/errorfs.go:func (fs *errorFilesystem) CreateSymlink(name, target string) error                     { return fs.err }
lib/fs/basicfs_windows.go:func (BasicFilesystem) CreateSymlink(path, target string) error {
lib/fs/logfs.go:func (fs *logFilesystem) CreateSymlink(name, target string) error {
lib/fs/filesystem.go:   CreateSymlink(name, target string) error
lib/fs/basicfs_unix.go:func (f *BasicFilesystem) CreateSymlink(target, name string) error {
```

I guess the basicfs on unix version is correct, as it is consistent with the os library.

ceh:
See #4567 for rationale.

### DIFF
--- a/lib/fs/basicfs_windows.go
+++ b/lib/fs/basicfs_windows.go
@@ -28,7 +28,7 @@ func (BasicFilesystem) ReadSymlink(path string) (string, error) {
 	return "", errNotSupported
 }
 
-func (BasicFilesystem) CreateSymlink(path, target string) error {
+func (BasicFilesystem) CreateSymlink(target, name string) error {
 	return errNotSupported
 }
 
--- a/lib/fs/errorfs.go
+++ b/lib/fs/errorfs.go
@@ -20,7 +20,7 @@ type errorFilesystem struct {
 func (fs *errorFilesystem) Chmod(name string, mode FileMode) error                      { return fs.err }
 func (fs *errorFilesystem) Chtimes(name string, atime time.Time, mtime time.Time) error { return fs.err }
 func (fs *errorFilesystem) Create(name string) (File, error)                            { return nil, fs.err }
-func (fs *errorFilesystem) CreateSymlink(name, target string) error                     { return fs.err }
+func (fs *errorFilesystem) CreateSymlink(target, name string) error                     { return fs.err }
 func (fs *errorFilesystem) DirNames(name string) ([]string, error)                      { return nil, fs.err }
 func (fs *errorFilesystem) Lstat(name string) (FileInfo, error)                         { return nil, fs.err }
 func (fs *errorFilesystem) Mkdir(name string, perm FileMode) error                      { return fs.err }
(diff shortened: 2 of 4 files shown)


lib/fs/basicfs_unix.go
```diff
@@ -14,7 +14,7 @@ func (BasicFilesystem) SymlinksSupported() bool {
 	return true
 }
 
-func (f *BasicFilesystem) CreateSymlink(target, name string) error {
+func (f *BasicFilesystem) CreateSymlink(name, target string) error {
```
AudriusButkevicius: So I am not sure, if this is right, as this swaps arguments to all other things.

lkwg82: at least some tests failing

imsodin: Target (as in what the symlink is pointing to) must be first to do it the same way as `os`.

ceh: Sorry about that, my bad (pushed and created a PR which wasn't the latest). Updated!

AudriusButkevicius:
@st-review lgtm

st-review:

calmh:
Erh, no. This works, but it's wrong. Tracing what actually happens, starting with the one place it is used in anger:

`rwfolder.go#741:  return f.fs.CreateSymlink(file.SymlinkTarget, path)`

That clearly has the `target` as the first parameter and the `name` as the second. Then everything is named the opposite in the filesystem layer. At the end we reach

```
basicfs_unix.go#22:  return os.Symlink(name, target)
```

but

```
// Symlink creates newname as a symbolic link to oldname.
// If there is an error, it will be of type *LinkError.
func Symlink(oldname, newname string) error {
```

so the first parameter is what we call `target` and the second is `name` and it's switched again, and works.

I'd harmonize this by fixing the order of the arguments in os.Symlink so our CreateSymlink is correct, and then switching the arguments to CreateSymlink in rwfolder and the tests.

imsodin:
As the order in our code isn't consistent, why use the opposite order than in os package? That seems like asking for confusion. We can keep all code that uses this function and just unify all our function definitions to be `Symlink(target, name string)`.

calmh:
Yeah do that instead, just as long as it’s both consistent and works. :)

ceh:
PTAL @calmh

calmh:
Will do. Not in front of a computer, but tomorrow!

imsodin:
@st-review lgtm

st-review:

AudriusButkevicius:
@st-review lgtm

st-review:
:ok_hand: Merged as bbc178ccc4caff764d49f5b079cacf253c744b72. Thanks, @ceh!

Issue #2760, opened 2016-02-05
Max simultaneous scans?


entonio:
Is there a way to limit the number of concurrent scans? I have a NAS that works reasonably when processing a single folder, but performance drops exponentially when syncthing tries to rescan more than one at the same time, which is normal given disk thrashing (I suppose flash storage wouldn't suffer from this issue). Is there any way to tell syncthing (on Mac) to rescan only one folder at a time, rather than to try them all at once? If not, could that be implemented?

AudriusButkevicius:
There is no way, apart from managing rescans yourself, or making sure that the scan ranges somehow don't overlap.

That could be implemented, but needs someone who is interested in implementing this.

entonio:
Thanks! (I'd have added the tag myself, but only the owners can do it, I believe.)

Is there any way to avoid rescan on re/starting syncthing? With that, I could manage the rescans myself, but without it I can't because they start automatically (a number of those folders are 'write only', being mirrors of remote read-only folders, so properly speaking they only need the initial scan).

AudriusButkevicius:
No, a scan on startup is mandatory, as we need to reconcile the state while we were offline.

entonio:
I understand that rationale, but from the moment one can disable normal rescans via the 0 interval, then Syncthing is not monitoring the folder anyway - anything can happen to it and Syncthing won't know. From that perspective, if the user disabled rescanning it, then it should apply to startup also. There's always manual scanning available if need be.
Picture 10 TB folders which are only used as mirrors ('write only') and having to rescan them with every restart for no gain. The only useful scan is on creation, because the user may have initialized the folder with a copy of the remote contents, butt all subsequent ones are of no use (unless some problem comes up, but for that there's manual rescan). I do realise you didn't set out to write Syncthing for the purpose of mirroring, but it has turned out as an excellent choice for it.

sisu4syncthing:
@entonio 
For really big (trillions of files) folders (which almost never change in a week), I have set the rescan interval to 36000 seconds, which are 10 hours. 
So maybe you can add a zero and let the folder rescan every 100 hours?

entonio:
[...]

PR #4635, opened 2018-01-01
PoC: single global folder scanner (as opt-in)

lkwg82:
### Purpose

Limit parallel scanning and just run one folder scanner at a time.

proof-of-concept implementation for #2760 

### Testing

not yet, except manual and printing

### Documentation

TBD

-----
Please discuss!


lib/model/folderscanner.go
```diff
@@ -50,10 +50,32 @@ func (f *folderScanner) Scan(subdirs []string) error {
...
+}
+
+func newFoldeScannerFactory(hasSingleGlobalFolderScanner bool) folderScannerFactory {
+	return folderScannerFactory{
+		hasSingleGlobalFolderScanner: hasSingleGlobalFolderScanner,
+		globalScanRequests:           make(chan rescanRequest),
+	}
+}
+
+func (fsf *folderScannerFactory) CreateOrGetSingleGlobalOrNewChannel() chan rescanRequest {
```
AudriusButkevicius: Srsly?

https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition

Why do we even need a factory.
I'd use a semaphore, and store how many concurrent scans we allow.

AudriusButkevicius: Also, I am not sure this does what you expect. I suspect this causes any random folder to be scanned.

lkwg82: LOL, needed to read and ROFL.
I dont play often with semaphores ... thx for the hint.

Still laughing ... GREAT!

lkwg82: Why should it scan randomly?

AudriusButkevicius: Because every folder is reading and sending to the same channel

lkwg82: aaah ... I see ... in the folder impl's. ok. 

```
case req := <-f.scan.now:
			req.err <- f.scanSubdirs(req.subdirs)
``` 

Silly me. :|

lkwg82:
Broken as of now.

calmh:
I'd probably look into making `scanner.parallelHasher` a persistent, global thing instead. Export it, make it take a filesystem and output channel as well as the thing to scan on the inbox channel, and have a wrapper function Scan() that essentially

```
func (ph* parallelHasher) Scan(fs fs, in protocol.FileInfo) (protocol.FileInfo, error) {
  rc := make chan(something)
  ph.inbox <- request{...}
  result := <-rc
  return result.info, result.err
}
```

then the total parallelity is the number of workers on the parallel hasher.

lkwg82:
I'm currently writing tests and miss the understanding of how the `Serve()` of a folder is invoked. I guess with the help of supervisor or am I conpletly wrong?

imsodin:
You are right: https://godoc.org/github.com/thejerf/suture
The folder services are added to the model supervisor in `startFolderLocked`.

AudriusButkevicius:
You are really getting to the point where there is no way to get this merged. In between dinner and shower time I had this implemented locally, and it took around 12 lines in a single file (untested). You are now in the thousands with the java paradigms and dependencies you are pulling in.

lkwg82:
> You are really getting to the point where there is no way to get this merged.

Partly it is my learning to write clean code with tests in an idiomatic way. So after growing the codebase I will shrink it again. (http://blog.thecodewhisperer.com/permalink/when-to-write-learning-tests)

> In between dinner and shower time I had this implemented locally, and it took around 12 lines in a single file (untested).

Nice. I admire you ability and your understanding of the code base. 
But ... 

> untested 
How you even know it runs?

Try to write a test - first!

> You are now in the thousands with the java paradigms and dependencies you are pulling in.

I dont rely on the `stretchr/assert` libs, just shorter test-validations. I can easily remove it.

But I assume that there could be better separation of concern. No need for parameters just passed through.

I try to get next year to FOSDEM to drink beer and talk to you in person. I'm eager to meet you. When you are in Berlin or around, let me know!
 

> All you need is an atomic counter and sleeping while that breaches some maximum

This is my idea, borrowed from you. Now I'm marching through the code. Find the best way.

Btw, thx for your early feedback. Really appreciate that.

AudriusButkevicius:
My version:

```diff
diff --git a/lib/config/optionsconfiguration.go b/lib/config/optionsconfiguration.go
index 74e1df4..e96f92f 100644
--- a/lib/config/optionsconfiguration.go
+++ b/lib/config/optionsconfiguration.go
@@ -143,6 +143,7 @@ type OptionsConfiguration struct {
        KCPSendWindowSize       int                     `xml:"kcpSendWindowSize" json:"kcpSendWindowSize" default:"128"`
        KCPReceiveWindowSize    int                     `xml:"kcpReceiveWindowSize" json:"kcpReceiveWindowSize" default:"128"`
        DefaultFolderPath       string                  `xml:"defaultFolderPath" json:"defaultFolderPath" default:"~"`
+       ParallelScans           int                     `xml:"parallelScans" json:"parallelScans" default:"0"`

        DeprecatedUPnPEnabled        bool     `xml:"upnpEnabled,omitempty" json:"-"`
        DeprecatedUPnPLeaseM         int      `xml:"upnpLeaseMinutes,omitempty" json:"-"`
diff --git a/lib/model/model.go b/lib/model/model.go
index 6b3026b..01d280a 100644
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -19,6 +19,7 @@ import (
        "runtime"
        "sort"
        "strings"
+       "sync/atomic"
        "time"

        "github.com/syncthing/syncthing/lib/config"
@@ -38,7 +39,10 @@ import (
        "github.com/thejerf/suture"
 )

-var locationLocal *time.Location
+var (
+       scansActive   int32
+       locationLocal *time.Location
+)

 func init() {
        var err error
@@ -1908,6 +1912,25 @@ func (m *Model) ScanFolderSubdirs(folder string, subs []string) error {
 }

 func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, subDirs []string) error {
+       parallelScans := m.cfg.Options().ParallelScans
+       if parallelScans > 0 {
+               for {
+                       active := atomic.LoadInt32(&scansActive)
+                       if int(active) < parallelScans {
+                               if atomic.CompareAndSwapInt32(&scansActive, active, active+1) {
+                                       defer atomic.AddInt32(&scansActive, -1)
+                                       break
+                               }
+                       }
+                       select {
+                       case <-ctx.Done():
+                               return ctx.Err()
+                       case <-time.After(100 * time.Millisecond):
+                       }
+               }
+
+       }
+
        m.fmut.RLock()
        if err := m.checkFolderRunningLocked(folder); err != nil {
                m.fmut.RUnlock()
```

lkwg82:
this is incomplete:

we have a parallel
- ... hashing of files per folder
- ... walk of subdirs 

you gate the walk of subdirs, and leaves out the parallel hashing.

I'm not quite sure how to combine these in a good way.

And I dont like the waiting, why not use a semaphore?

AudriusButkevicius:
The original ticket talks about limit number of folders we are scanning, this is what I've implemented.
Reason for not using a semaphore is because semaphore size cannot be adjusted on the fly, atleast in x/sync package.

Number of concurrent files we hash in a scan is already handled by FolderConfiguration.Hashers config parameter.
Walking of subdirs is not parallel either.

lkwg82:
aah, you try to solve  deadlocking, understand

AudriusButkevicius:
I am trying to solve the ticket you are referencing. What are you trying to solve?

lkwg82:
I'd like to have some tests for parts of the system. Also I try to solve the 'single scanner' issue.

lkwg82:
@st-review rebuild

lkwg82:
> Walking of subdirs is not parallel either.

got it :(

lkwg82:
but ... yes there is always sth to say against ^g^ ...

you miss restricting number of hashers in when `single`-option is activated, so a big directory tree and big files still would slow down spinning disks. Agree? (ask in case I missed sth else)

AudriusButkevicius:
As I said before, I don't think there should be a single option to start with. It should be adjustable.
Hashers is already adjustable, so I don't think we need to do anything there.

I stopped following the diff as it's still 7k and smells like Java.

AudriusButkevicius:
When I say smells like Java, I mean tests being in their own directory.
In go, the standard for tests is mymodule.go holds the code, mymodule_test.go holds the tests.

If I want to run a tests for my library, having to cd into test/features directory is very non-Go.

Also, there is a lot of stutterig:

https://blog.golang.org/package-names

see Naming package contents

scanner.NewScannerLimiter can just be scanner.NewLimiter etc

lkwg82:
> tests being in their own directory

What about 'almost' black-box tests for a feature? Given a selected option with an expected behaviour. In my experience it makes a lot of sense to separate from unit/integration tests.

> smells like Java

Aaarg. That sounds too dogmatic. 

Ok, naming is difficult, but I agree can be shorter although I'm in favor for longer names. (this is not a problem at all)

What is the problem with Java?? I started my coding life with pascal/perl and php. So I know there are different ways, but I've seen similarities accross languages in the area of design.

When you write 'Java' I'm going to read 'the devil' :). No it is not, it is just very verbose. It seems you were hurt by some java guys. Sry! 
You understand bitshifting and I understand testing and design. Let's come together!

I will throw away 99% of my changes when I see it is a bit more testable and not less readable.
As we say in Germany there are some 'skeletons in the closet' ... the 'model.go' is a really big mess. 

There is too less separation of concern. Hard to test. So bug-prone when changing.

AudriusButkevicius:
So I am not arguing that there are tons of skeletons in the closets.
I am however arguing that some of the changes you are making do not conform with Go guidelines or the way go developers would expect to find things.

lkwg82:
What is the guideline in Go for spec tests? Mix it with unit and it? Hey come on there can be an advantage of having it at a different place, or?

AudriusButkevicius:
Well you can argue there can be an advantage to using snake case and what not, yet that's not what we are using, nor is this what the rest of the Go community (or go the language itself) is using for tests.

Testing is explained here:

https://golang.org/pkg/testing/

lkwg82:
> snake case 

I'll fix if I mess around with different styles.

The link for Testing is not more than writing unit tests. No word about spec tests/bdd or any other. 

But lets say when we have a impl we can agree on, we discuss tests. Ok?

calmh:
I haven't followed all of this, but if you're looking for black box tests in Go you do that by putting the test file in `package foo_test` while in the same directory as `package foo`. It becomes a separate package as usual and needs to import `foo` to access it, while still being part of the package `foo` tests.

lkwg82:
@calmh so I'll move the blackbox tests for syncthing itself to `cmd/syncthing/syncthing_test`, correct?

AudriusButkevicius:
No, tests for a package still live in foo_test.go file alongside foo.go. It can declare `package foo_test` which is different from `package foo`, hence can't access the internal (lowercase) functions, and becomes black box tests.

lkwg82:
I think I didn't make my point clear enough. 

I dont want to mix unit and integration tests. I'd like to have a test for a feature we add. So we can run all feature tests and see, if they still in place.
Even in the case the feature is distributed over many files. And in the case the implementation has changed.

Maybe there is a better solution to this problem I'm not aware of.

AudriusButkevicius:
Integration tests in our case involve running a compiled syncthing binary and poking at it from the API etc.

What you are writing are still unit tests, even though the unit makes use of multiple packages, yet it still starts at some top level package that can be unit tested.

lkwg82:
mhm, ok, I understand and agree. Not willing to write browser automation. 

Where to put this test? `model_test.go`?

AudriusButkevicius:
I don't know, I haven't checked the diff for the reasons I've explained before.

lkwg82:
See: https://github.com/syncthing/syncthing/pull/4635/files#diff-acdcb1f6e6c58b5c3608134d5c6aa1c6R33-R37

AudriusButkevicius:
Sorry, but I can't answer as I think the diff you have there does too much for what it's trying to achieve.

imsodin:
I haven't looked at it in detail either, as it changes very often. Also due to shifting code between files there is no diff for parts - that makes it harder to review. I would delay doing that until code is stable (or even do it in a separate PR afterwards). From a quick glance it looks like you add an option to limit parallel hashing globally to a single instance and enforce it at folder level. As we allow adjusting parallel hasher per folder, it seems just consistent to also allow configuring a number of global parallel hashers.

AudriusButkevicius:
To be honest, I have no clue what this is trying to achieve. The original ticket asks for a way to limit number of folders are scanned, this has a boolean (rather than a number) and for some reason goes tapping into hashers.

lib/config/folderconfiguration.go
```diff
@@ -275,3 +275,7 @@ func (l FolderDeviceConfigurationList) Len() int {
 func (f *FolderConfiguration) CheckFreeSpace() (err error) {
 	return checkFreeSpace(f.MinDiskFree, f.Filesystem())
 }
+
+func (f *FolderConfiguration) SetFilesystem(fs fs.Filesystem) {
```
calmh: This method should not exist imho, there is no reason for someone from the outside to be able to manipulate the cache structure

lib/fs/basicfs.go
```diff
@@ -28,6 +28,9 @@ type BasicFilesystem struct {
 	root string
 }
 
+func NewBasicFilesystem(root string) *BasicFilesystem {
```
calmh: This should just be a rename of the existing constructor if it needs to be exported

lib/model/model.go
```diff
@@ -162,6 +164,8 @@ func NewModel(cfg *config.Wrapper, id protocol.DeviceID, clientName, clientVersi
 		remotePausedFolders: make(map[protocol.DeviceID][]string),
 		fmut:                sync.NewRWMutex(),
 		pmut:                sync.NewRWMutex(),
+
+		scannerLimiter: scanner.NewScannerLimiter(cfg.Options().SingleGlobalScanner),
```
calmh: Lets not do the boolean parameter. We don't have other types that takes a boolean for whether they do their thing or not. If needed, we `if` here and set the attribute to some no-op variant of the interface here.

lib/scanner/walk.go
```diff
@@ -559,3 +562,54 @@ type noCurrentFiler struct{}
...
+type ScannerLimiter interface {
+	Aquire(ctx context.Context, d ...string)
+	Release(d ...string)
+}
+
+func NewScannerLimiter(single bool) ScannerLimiter {
+	if single {
+		l.Infoln("single global folderScanner limit ")
+		return &singleGlobalScannerLimiter{
+			sem: semaphore.New(1),
```
calmh: This whole thing is equivalent to a regular `sync.Mutex` or a single-buffered `chan struct{}`. I do not think we should pull in that dependency. It's also racy and not great (I happened to glance at it) so I think we should just use a channel if we need to do more than `1` here.

lkwg82:
> I'm also slightly worried about the premise for this whole thing.

Before I adjust the code ... can we consider your worry and discuss this first?

lib/config/folderconfiguration.go
lkwg82: How would you test, that the limiter is in place? As we dont have separation between concerns I can not replacing scanning with a dummy, so I need to check some other places. I could not find another "good" place.

This unveils the model is hard to test.

lkwg82:
@calmh I'll try to catch some numbers.

calmh:
So there are several triggers causing scans:

 1. periodic of the full folder, based on a timer
 2. on request for a subset of files or directories, based on API request or the puller having noticed something out of date and asked for a scan
 3. on request from the filesystem watcher when it saw something change

I don't think we should block scans for cases #2 or #3 based on this. (#3 might be borderline, but blocking the scan defeats the purpose of the filesystem watcher and the better action is to turn it off.)

As such I think you should hook in where the scheduled scans are triggered:

- https://github.com/syncthing/syncthing/blob/master/lib/model/rwfolder.go#L203
- https://github.com/syncthing/syncthing/blob/master/lib/model/rofolder.go#L51

In both of those cases I suggest grabbing a "scan token" from the semaphore, doing the thing, and returning it. Grabbing the token should be non blocking, with the scan timer pushed forward some interval (30s? `15 + rand.Intn(45) * time.Second`?) on failure - again, so that we don't block the other actions happening in that loop.

Testing wise this is tricky, as you've identified. I wouldn't spend too much energy on it. If it's conceptually simple we can live with a manual or integration test, and adding thousands of lines of test code is also a maintenance burden, likely outweighing the advantage of having the tests.

lkwg82:
thx for the explanation and code direction.

This PoC is for me heavy learning  and I enjoy it! 

> As for personal coding style, I suggest viewing it as an exercise to implement this in as few lines as possible with as small a diff as possible. :)

Accepted! :)

imsodin:
To address the blocking operation problem, couldn't the hasher limitation also be moved in the opposite direction: to the parallel hasher? That still means delays but at least every folder has an equal chance to get a file scanned, so one big folder scanning won't block everything else.  
Taking calmh's scenarios 2 and 3 out of the limitation certainly solves it, but it feels like defeating the point of this PR, as these aren't necessarily small scans (e.g. hitting rescan in the web UI).

AudriusButkevicius:
I think we need a new state, like ScanQueued to display in the UI and to be honest, just block the scans.

lkwg82:
I just did some "benchmarking":

A spining disk, with  6 shares each a copy of 191436 files in 21962 directories a total size of ~5.54 GiB.

the parallel startup and the sequential startup (from starting until last 'initial scan' was completed) roughly the same 5minutes. I did >5times for each configuration.

the underlying filesystem is zfs.

So from what I see there will be no significant benefit to implement this.

lkwg82:
just checked

AudriusButkevicius:
It's potentially because you mess around with numHashers. Also, on zfs you technically read the same data 6 times so it's not a fair benchmark, and using ext4 or windows might have better results. Even if this doesn't make things faster, it potentially reduces threshing, which might give better UX.

calmh:
Yeah… ZFS is a smart filesystem. Even if you don’t have RAM enough for the file data to become cashed, it explicitly recognizes that there are five (or whatever) parallell sequential file reads and does readahead per file to avoid unnecessary seeking.

This is one of the reasons I haven’t cared much about this historically - servers are smart (so it doesn’t matter there), modern consumer devices have SSDs (so it doesn’t matter there), old consumer devices are slow anyway (this is where it matters, but the target audience is diminishing every day).

lkwg82:
I noticed I'm not doing it correctly, but this is the setup I have. So I'm not interested anymore.

PR #4637, opened 2018-01-02
gui: Prevent error without completion and nicer wrapping (fixes #4636)

### LINKED ISSUE #4636: List of files needed on remote is not wrapped at word boundaries
The automatically inserted line breaks in the table looks rather ugly:

screenshot_2018-01-02_12-06-16

Also if you are in a hurry clicking on the out of sync link, you can trigger a JS error because the completion object may not yet be initialised.

imsodin:
Old compared to new word breaking:

wrap

The "fix" for undefined completions just disables the early return, meaning there will be a few needless rest calls, but I think that's ok (it's anyway not likely that this scenario is triggered unconsciously).

@calmh As this is minor, I think you can wait for the autoaccept panic issue to be fixed before releasing a new rc.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -2054,7 +2054,7 @@ angular.module('syncthing.core')
             resetRemoteNeed();
             $scope.remoteNeedDevice = device;
             $scope.deviceFolders(device).forEach(function(folder) {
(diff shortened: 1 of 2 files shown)


calmh:
@st-review merge

st-review:
:ok_hand: Merged as d87287c0d0319f30025cdc91824d9d5d58fdd40e. Thanks, @imsodin!

PR #4639, opened 2018-01-02
lib/model: Fix panic when auto-accepting existing paused folder (fixes #4636)

### LINKED ISSUE #4636: List of files needed on remote is not wrapped at word boundaries

AudriusButkevicius:
This no longer pokes at model internals, and only touches the config.
As a result, model handles this in CommitConfiguration, which restarts
the folders if things change, which repopulate m.folderDevice, m.deviceFolder
and other interal mappings.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1096,16 +1096,17 @@ func (m *Model) handleIntroductions(introducerCfg config.DeviceConfiguration, cm
 	foldersDevices := make(folderDeviceSet)
 
 	for _, folder := range cm.Folders {
-		// We don't have this folder, skip.
-		if _, ok := m.folderDevices[folder.ID]; !ok {
-			continue
-		}
-
 		// Adds devices which we do not have, but the introducer has
 		// for the folders that we have in common. Also, shares folders
 		// with devices that we have in common, yet are currently not sharing
 		// the folder.
-	nextDevice:
+
+		fcfg, ok := m.cfg.Folder(folder.ID)
+		if !ok {
+			// Don't have this folder, carry on.
+			continue
+		}
+
 		for _, device := range folder.Devices {
 			// No need to share with self.
 			if device.ID == m.id {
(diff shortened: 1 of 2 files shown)


lib/model/model.go
```diff
@@ -1118,22 +1119,29 @@ func (m *Model) handleIntroductions(introducerCfg config.DeviceConfiguration, cm
...
-				if er == folder.ID {
-					// We already share the folder with this device, so
-					// nothing to do.
-					continue nextDevice
+			} else {
+				for _, dev := range fcfg.DeviceIDs() {
+					if dev == device.ID {
+						// We already share the folder with this device, so
+						// nothing to do.
+						continue
```
imsodin: This should probably continue the outer for loop.

lib/model/model_test.go
```diff
@@ -1218,6 +1225,113 @@ func TestAutoAcceptFallsBackToID(t *testing.T) {
 	}
 }
 
+func TestAutoAcceptPausedWhenFolderConfigChanged(t *testing.T) {
+	// Existing folder
+	id := srand.String(8)
+	idOther := srand.String(8) // To check that path does not get changed.
+	defer os.RemoveAll(filepath.Join("testdata", id))
```
imsodin: The `idOther` needs removing too. Same for existing `TestAutoAcceptExistingFolder` test. My testdata is littered with these random dirs ;)  
Or run all tests in temporary dirs anyway.

AudriusButkevicius: Uh yeah, good stop.

calmh:
Neat

calmh:
@st-review merge

st-review:
:ok_hand: Merged as 53f4dfe83cb57d2cad1618edeb5df65b8ebab1ac. Thanks, @AudriusButkevicius!

PR #4633, opened 2018-01-01
build: adds flag for debugging the binary with delve

lkwg82:
…ry with delve

### Purpose

enables debugging with delve, see documentation [PR/366 in docs](https://github.com/syncthing/docs/pull/366)

`go run build.go -debugBinary build` (omits -ldflags)

PR #4642, opened 2018-01-02
lib/scanner: little bit de-obfuscating the code ;)

lkwg82:
### Purpose

readability

### Testing

rerun tests


lib/scanner/blocks.go
```diff
@@ -87,14 +94,14 @@ func Blocks(ctx context.Context, r io.Reader, blocksize int, sizehint int64, cou
...
-			WeakHash: whf.Sum32(),
+			WeakHash: hashFun.Sum32(),
 		}
 
 		blocks = append(blocks, b)
 		offset += n
 
 		hf.Reset()
-		whf.Reset()
+		hashFun.Reset()
```
AudriusButkevicius: hashFun and hf mean the same thing, yet are different things.
I'd call them strongHash and weakHash

lkwg82: yes, I had the same idea

lib/scanner/blocks.go
```diff
@@ -24,48 +24,64 @@ type Counter interface {
...
+}
+
+// Blocks returns the blockwise hash of the reader.
+func Blocks(ctx context.Context, reader io.Reader, blocksize int, sizehint int64, counter Counter, useWeakHashes bool) ([]protocol.BlockInfo, error) {
+	if counter == nil {
+		counter = &NoopCounter{}
+	}
+
+	strongHash := sha256.New()
+	hashLength := strongHash.Size()
```
AudriusButkevicius: Size() clearly indicates that it's a size, rather than a length?

lib/scanner/blocks.go
```diff
@@ -24,48 +24,64 @@ type Counter interface {
...
+
+	writer, weakHash := createWriterAndWeakHashFun(strongHash, useWeakHashes)
```
AudriusButkevicius: Given function signatures in go start with func rather than fun, let's call it func here too.

lib/scanner/blocks.go
```diff
@@ -24,48 +24,64 @@ type Counter interface {
...
+	blocks, hashes, lreader := prepareReader(sizehint, reader, blocksize, hashLength)
+
 	// A 32k buffer is used for copying into the hash function.
 	buf := make([]byte, 32<<10)
 
 	var offset int64
-	lr := io.LimitReader(r, int64(blocksize)).(*io.LimitedReader)
+	var thisHash []byte
```
AudriusButkevicius: Don't see this used

lkwg82: see (already rename hashLength)
```go
                hashes = strongHash.Sum(hashes)
		thisHash, hashes = hashes[:hashSize], hashes[hashSize:]

		b := protocol.BlockInfo{
			Size:     int32(n),
			Offset:   offset,
			Hash:     thisHash,
			WeakHash: weakHash.Sum32(),
		}
```

AudriusButkevicius: Nvm, needed expanding.

lib/scanner/blocks.go
```diff
@@ -215,3 +229,7 @@ func (noopHash) Size() int                 { return 0 }
 func (noopHash) Reset()                    {}
 func (noopHash) Sum([]byte) []byte         { return nil }
 func (noopHash) Write([]byte) (int, error) { return 0, nil }
+
+type NoopCounter struct{}
```
AudriusButkevicius: can be private (lower case name)

lib/scanner/blocks.go
lkwg82: what do you mean?

lib/scanner/blocks.go
lkwg82: no, replaced `nil` counters with this `NoopCounter `

lkwg82: so we push nil to the boundaries

lib/scanner/blocks.go
AudriusButkevicius: Didn't see as I didn't expand the diff, this is ok.

AudriusButkevicius:
@st-review lgtm

st-review:

calmh:
I'm sorry, I don't see this as an improvement. I could quibble about specific naming and type issues etc but the main point is that I see it as busywork that increases churn (thus lowering recognition and potentially adding bugs) while not actually improving things to a significant degree. I understand (and share) the impulse to pick something apart when figuring out how it works, but most such picking-apart do not need to be committed.

Refactoring is fine in general, but this does not carry its weight in my opinion.

AudriusButkevicius:
So I disagree, I think this makes things better. Yeah the createThisAndThat func is a bit java'y, and explicitly passing a NoopScanner when someone can easily still pass in a nil and break things adds no value, but other changes make it better.

lkwg82:
@calmh I'm a bit surprised, that you just close it. I agree on the potential-risk perspective and really appreciate that.

Let me explain why I did this and what the rationale is behind:

I'm working on PR #4635 (PoC: single global folder scanner (as opt-in)) and read code on the way.
For the `single folder scanner` it was necessary to have #4633 and this PR and more to come. But these supplementary changes can live on there own.

So I read code and see here and there some unreadable lines/functions. So what I do as practicing software craftsman ... clean and refactor on my way through (the jungle).

As I learned some time ago, small PRs are more likely to get merged than bigger ones. So I split features from simple refactorings. So this is one of these refactorings.

So the improvements of this refactoring is enable changeability of the code by make it readable and unterstandable. So I expanded the names of the variables, shortened the function from ~80 to ~60 lines and extracted sub-scopes.

> I see it as busywork that increases churn.

That has a negative mood. Maybe this is just an expression of stress. Nobody is immune.

As a opensource project we should encourage collaboration by make our work accessable, so to make it readable. And this was not readable. Same with the `model.go` which is a big monster and is getting bigger and bigger. 

Besides that:

- What should the process with refactorings? (I dont like doing features with cleanups mixed.)
- I interpret your mention of `potentially adding bugs` as fear and distrust in the tests we have. Should we start here?
- What about `clean code` in here? I stopped by the docker code it is readable very well. That gives us more qualitive PRs.

I'm pretty sure I missed something. Hope it makes some sense in your eyes ;).

>

AudriusButkevicius:
I agree with you, your PR makes things more readable, hence is adding value and should he merged.

calmh:
I wasn't going to argue too much about this, but since I've said A I'll have to say B.

Please try to take this as criticism of the PR in a constructive manner, although it is sort of borderline as we are really discussing minor issues here to a greater degree than they deserve. That's counterproductive by itself, but it's my fault it's required in this case and I need to explain what I'm talking about so take it with my pre-apology. :/

## General points

- Refactorings and cleanups are good.
- Changes to make things prettier, where prettier is just personal preference, is not.

A reasonable refactoring might improve structure and make difficult things easier by breaking out and naming blocks of code. However, local variable names in Go are (shorter rather than long)[https://github.com/golang/go/wiki/CodeReviewComments#variable-names] and there is (maybe unfortunately) a certain vernacular. `i`, `r` and `w` are often counters, readers and writers for example. If the variable is used only for a few lines all within a screen, `reader` is no clearer than `r`.

Your PR here strikes me as extracting easy things and giving them ambiguous names, while also giving longer names to things that don't need longer names. And you can disagree about the names, but then we're back to making something prettier for personal preference.

(Some of the names in this code are not great; the hf/mhf/whf thing isn't optimal, I'll admit, and can improve.)

## The Actual Nitpicks

### NoopCounter transform

To me this doesn't add clarity as it's a fairly well established pattern to pass nil for optional interfaces. You might argue that it's not obvious what the nil is, and that's true but also affects things like the boolean in the same call. Since this is a constructor I'd consider it reasonable to convert it to functional style - WithCounter(...) etc, which would indeed make it clearer.

#### Potential bug

Passing nil is now a panic while it used to be valid. I'm hoping you grepped and fixed all the current callers for that. There's still a nonzero risk of someone passing a nil in the future, as it's not an uncommon pattern. Maybe that will be covered by a test, maybe not.

*If* we really want to avoid the nil check in the loop, I'd drop it in at the absolute top of the function and leave the call sites alone.

```go
if counter == nil {
    counter = NoopCounter{}
}
```

### createWriterAndWeakHash

This is an ambiguous name for a simple code block, that tells me almost nothing about what it does. I'm also usually sceptical of functions that contains a single if based only on a boolean parameter as the decision is already made at a higher level - and the if should be there. An improvement on the original here in my opinion could for example be this, without extracting a function:

```go
	var weakHf hash.Hash32 = noopHash{}
	var multiHf io.Writer = hf
	if useWeakHashes {
	  // Use an actual weak hash function, make make the multiHf
	  // write to both hash functions. 
		weakHf = adler32.New()
		multiHf = io.MultiWriter(hf, weakHf)
	}
```

### prepareReader

Again, this name tells me nothing about what actually happens in the function, and what the function does is mostly about allocation. The function also has the problem of potentially doing too much, with three complex return values - if it's preparing a reader, why return a list of blocks and stuff.

Naming is difficult, but if the point of the exercise is to improve naming I think we need to put some more work into this one.

If anything this function is more about preallocating block and hash areas than preparing readers and should be just that, but then we're just breaking out three simple, commented lines into a function again.

Those comments and allocations directly in Block() also provides some context for one of the few "tricky" aspects of the larger function (the hash and block area preallocation/usage), and moving them out into an anonymously named function makes it potentially harder to understand what happens when you reach the "// Carve out a hash-sized chunk of "hashes" to store the hash for this block." part.

The last block-sized limited reader returned by the function has, imho, nothing to do in that function, as it's something that gets reset and reused on every iteration later. It's more a property of the for loop than anything else.

A lot of arguments against small things, many of which would probably pass without comment as a part of larger block of new code, but when applied to existing old code it makes it not-an-improvement in my mind.

### r/reader lr/lreader

No clarity improvement because an `r` is a reader by default in a function like this. If anything is unusual and might warrant expansion it's the "l", maybe `limitedR`.

lkwg82:
Thx. for your explanations and almost all I understand and support after peeking and poking with this particular code.

> variable names  ... shorter rather than long

This collides with long lines of code and hence introduces complexity like of math proofs, where you need to lookup all  `a1`, `t`, `b2` ...  again and again.

My conclusion is that this refactoring needs a more global perspective as it also may needs restructuring some of the environment.

lkwg82:
@calmh Took your arguments into account and reverted almost all and did little other things I came accross.

lib/scanner/blocks.go
AudriusButkevicius: Still can be private.

AudriusButkevicius:
Apart from the noop thing being public, looks good to me.

lib/scanner/blocks.go
imsodin: I don't really see the point having a type for just one nil check?

imsodin: Actually, ignore my comment above, but Audrius' one about private/not exported still is unaddressed.

lkwg82:
will take a look into later

imsodin:
@st-review lgtm

st-review:

AudriusButkevicius:
@st-review lgtm

lib/scanner: Refactoring

st-review:
:ok_hand: Merged as 89a021609b8eb5fdc74a83a5094342719133c6f5. Thanks, @lkwg82!

PR #4650, opened 2018-01-05
gui: New rest endpoint to get errors when web UI is opened

imsodin:
Since #4340 pulls aren't happening every 10s anymore and may be delayed up to 1h.
This means that no folder error event reaches the web UI for a long time, thus no
failed items will show up for a long time. Now pulling errors are populated when the
web UI is opened.

One thing I noticed: A whole lot of functions on the folders, queue, folderstate, ... are exported but never used anywhere outside of the model package. Is this legacy or hopeful planning that all this stuff could be completely separated from model one day? :D

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -56,6 +56,7 @@ const (
 	defaultEventMask   = events.AllEvents &^ events.LocalChangeDetected &^ events.RemoteChangeDetected
 	diskEventMask      = events.LocalChangeDetected | events.RemoteChangeDetected
 	eventSubBufferSize = 1000
+	everything         = 1 << 16
 )
 
 type apiService struct {
@@ -111,7 +112,7 @@ type modelIntf interface {
 	RemoteSequence(folder string) (int64, bool)
 	State(folder string) (string, time.Time, error)
 	UsageReportingStats(version int, preview bool) map[string]interface{}
-	PullErrors(folder string) ([]model.FileError, error)
+	PullErrors(folder string, page, perpage int) ([]model.FileError, error)
 }
 
 type configIntf interface {
@@ -680,11 +681,22 @@ func (s *apiService) getDBCompletion(w http.ResponseWriter, r *http.Request) {
 func (s *apiService) getDBStatus(w http.ResponseWriter, r *http.Request) {
 	qs := r.URL.Query()
 	folder := qs.Get("folder")
-	sendJSON(w, folderSummary(s.cfg, s.model, folder))
+	if sum, err := folderSummary(s.cfg, s.model, folder); err != nil {
+		http.Error(w, err.Error(), http.StatusNotFound)
+	} else {
+		sendJSON(w, sum)
+	}
 }
 
-func folderSummary(cfg configIntf, m modelIntf, folder string) map[string]interface{} {
+func folderSummary(cfg configIntf, m modelIntf, folder string) (map[string]interface{}, error) {
 	var res = make(map[string]interface{})
+
+	pullErrors, err := m.PullErrors(folder, 1, everything)
+	if err != nil && err != model.ErrFolderPaused {
+		// Stats from the db can still be obtained if the folder is just paused
+		return nil, err
+	}
+	res["pullErrors"] = len(pullErrors)
 
 	res["invalid"] = "" // Deprecated, retains external API for now
 
@@ -699,7 +711,6 @@ func folderSummary(cfg configIntf, m modelIntf, folder string) map[string]interf
 
 	res["inSyncFiles"], res["inSyncBytes"] = global.Files-need.Files, global.Bytes-need.Bytes
 
-	var err error
 	res["state"], res["stateChanged"], err = m.State(folder)
 	if err != nil {
 		res["error"] = err.Error()
@@ -720,7 +731,7 @@ func folderSummary(cfg configIntf, m modelIntf, folder string) map[string]interf
 		}
 	}
 
-	return res
+	return res, nil
 }
 
 func (s *apiService) postDBOverride(w http.ResponseWriter, r *http.Request) {
@@ -736,7 +747,7 @@ func getPagingParams(qs url.Values) (int, int) {
 	}
 	perpage, err := strconv.Atoi(qs.Get("perpage"))
 	if err != nil || perpage < 1 {
-		perpage = 1 << 16
+		perpage = everything
 	}
 	return page, perpage
 }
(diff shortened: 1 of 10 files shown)


AudriusButkevicius:
Aren't the errors cleared after syncing, so this will still disappear?

imsodin:
They are cleared when syncing starts, which is fine, as when syncing stops, a folder error event is dispatched (if there are any errors). This PR only adds new information on UI start, after that everything should work fine (except I am missing something).

AudriusButkevicius:
Perhaps instead of clearing we should store them indefinately unconditionally? If there are no errors after pull, we set errors to empty array, and in the UI refresh errors on folder state change

imsodin:
I don't really know what you mean by indefinitely, but yeah, why not keep them until sync finished. This way if syncing takes a long time, you can still view the errors from before. Maybe it should be separate from this PR though, as it changes behaviour while this just fixes a (minor) regression.

AudriusButkevicius:
Well this already changes fundamentally how errors are delivered. I'd even suggest to ignore the event in JS and rely purely on this call.

imsodin:
Currently it doesn't change anything, it just uses this new rest endpoint on startup to populate the failed list. After that it still relies fully on the error events. What is the advantage of querying this info from JS instead of reaction to the event?  

Another option would be to add the number of failed items to the folder summary and never store failed items in JS. Only if the user clicks on the link, the list is queried with this new rest endpoint.

calmh:
@st-review lgtm (but not tested)

st-review:

lib/model/rwfolder.go
```diff
@@ -1811,13 +1811,20 @@ func (f *sendReceiveFolder) deleteDir(dir string, ignores *ignore.Matcher, scanC
 	return err
 }
 
-// A []fileError is sent as part of an event and will be JSON serialized.
-type fileError struct {
+func (f *sendReceiveFolder) PullErrors() []FileError {
+	if state, _, _ := f.getState(); state == FolderSyncing {
+		return nil
```
AudriusButkevicius: Why do we care about the state here?

imsodin: If the folder is syncing, the error list has just been cleared, is building up but not complete. Once it is, it will be sent on the folder error event. It's kind of similar as in the UI, while sync is in progress, there are no errors. But as you commented before, that premise is probably not ideal. Maybe you can comment on the last paragraph in https://github.com/syncthing/syncthing/pull/4650#issuecomment-355743842 , that would be the way I would go forward to change how failed items work (I didn't really understand your earlier proposition except for clearing errors after pulling).

AudriusButkevicius:
What I am essentially saying is perhaps we should get rid of the event usage in the GUI and only retrieve errors via PullErrors, which we query from UI on state changes. We don't clear errors on start of synx and just swap old errors with new after sync.

imsodin:
Agreed. What do you think about on top of that never remember errors, only error numbers, and query errors only when the user clicks on the failed item link? That would mean handling it like out of sync: Keep the number of failed items in the folder summary and actually query failed only if the user clicks on the link (and query is then paged too).

AudriusButkevicius:
Yeah, that makes sense. The errors could also be paginated.

cmd/syncthing/gui.go
```diff
@@ -111,6 +112,7 @@ type modelIntf interface {
 	RemoteSequence(folder string) (int64, bool)
 	State(folder string) (string, time.Time, error)
 	UsageReportingStats(version int, preview bool) map[string]interface{}
+	PullErrors(folder string, page, perpage int) ([]model.FileError, error)
```
AudriusButkevicius: Pagination should be in the GUI package, not the model I think.

lib/model/model.go
```diff
@@ -2226,25 +2226,13 @@ func (m *Model) State(folder string) (string, time.Time, error) {
 	return state.String(), changed, err
 }
 
-func (m *Model) PullErrors(folder string, page, perpage int) ([]FileError, error) {
+func (m *Model) PullErrors(folder string) ([]FileError, error) {
 	m.fmut.RLock()
+	defer m.fmut.RUnlock()
 	if err := m.checkFolderRunningLocked(folder); err != nil {
```
AudriusButkevicius: Just thinking out loud. I suspect after calling this functions we usually reach into the runner, so perhaps this function should return the runner or error (paused/not exists).

gui/default/syncthing/core/syncthingController.js
```diff
@@ -670,12 +647,12 @@ angular.module('syncthing.core')
             refreshNeed($scope.neededFolder);
         };
 
-        $scope.failedPageChanged = function (page) {
-            $scope.failedCurrentPage = page;
-        };
-
-        $scope.failedChangePageSize = function (perpage) {
-            $scope.failedPageSize = perpage;
+        $scope.refreshFailed = function (page, perpage) {
```
AudriusButkevicius: Is the capitalization if perpage consistent in JS and rest of Go?

lib/model/model.go
imsodin: Sometimes the runner is used afterwards, but also a few other folder map values, and at least once it is used without needing the runner afterwards. It seems more straightforward to me if it is a status check, as opposed to a lookup that also does some extended status checking. And I guess this isn't really relevant performance wise (?).

gui/default/syncthing/core/syncthingController.js
imsodin: Yes, it's everywhere like this. Maybe it is like this because the url parameter is lowercase.

lib/model/model.go
AudriusButkevicius: Sure, just checking.

imsodin:
This is ready from my side - what's the state of review?

cmd/syncthing/gui.go
```diff
@@ -1349,6 +1361,32 @@ func (s *apiService) postFolderVersionsRestore(w http.ResponseWriter, r *http.Re
 	sendJSON(w, ferr)
 }
 
+func (s *apiService) getPullErrors(w http.ResponseWriter, r *http.Request) {
+	qs := r.URL.Query()
+	folder := qs.Get("folder")
+	page, perpage := getPagingParams(qs)
+
+	if errors, err := s.model.PullErrors(folder); err != nil {
+		http.Error(w, err.Error(), http.StatusNotFound)
```
AudriusButkevicius: Just return here, and de-dent the else.

lib/model/rwfolder.go
```diff
@@ -1811,13 +1811,17 @@ func (f *sendReceiveFolder) deleteDir(dir string, ignores *ignore.Matcher, scanC
 	return err
 }
 
-// A []fileError is sent as part of an event and will be JSON serialized.
-type fileError struct {
+func (f *sendReceiveFolder) PullErrors() []FileError {
+	return f.currentErrors()
```
AudriusButkevicius: Just rename f.currentErrors to PullErrors()

AudriusButkevicius:
Apart from the few nitpicks, looks good.
Did you test the JS part?

imsodin:
I have a test folder with some failed items in it, and the web UI and failed item display do what I expect without any errors in the console.

AudriusButkevicius:
@st-review lgtm

st-review:

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as fecb21cdb119332dcb39c72f08cf2f5f10e7692c. Thanks, @imsodin!

PR #4664, opened 2018-01-11
Substitute '\' with '/' in paths in External File Versionining (fixes #4560)

### LINKED ISSUE #4560: Windows - External File Versioning Command path requires \\ instead of \
### What happened

On Windows, an external versioning command path requires \\\\ (two backslashes) instead of single \\.
This sounds a bit unnatural and the instruction in [the doc page](https://docs.syncthing.net/users/versioning.html#example-for-windows) doesn't work.

### Any steps to reproduce the problem.

For example, if the command is set like this:

iss
then I get this error log:
`[VQZ3H] 2017/11/30 00:25:31.724302 rwfolder.go:1675: INFO: Puller (folder "\"sttest\" (vu3ga-eo2y9)", file "test.txt"): delete file: exec: "C:UserstestusrDesktopext.bat": file does not exist`

OK the path isn't recognized correctly with single \ for some reason.

I found two workarounds:
- use \\\\ instead of \\
- use / (this is uncommon for windows)

### Version Information

Syncthing Version: v0.14.40
OS Version: Windows 10 x64
Browser Version: Edge

Kropekk:
### Purpose

Substitute '\\' with '/' in paths in External File Versionining (fixes #4560). Spaces are handled correctly. When the variable is read in back-end (Go file), the '\\' is already parsed, hence substitution is done in front-end.

Edit: I was wrong, we can see unparsed input in .go file as well
### Testing

I've just changed path separator from Windows-style to Linux-style. Both separators work well on Windows, so there is no need to test.

### Documentation

Previously documentation didn't work, now it will.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1581,6 +1581,7 @@ angular.module('syncthing.core')
                 $scope.currentFolder.staggeredMaxAge = 365;
             }
             $scope.currentFolder.externalCommand = $scope.currentFolder.externalCommand || "";
+            $scope.currentFolder.externalCommand = $scope.currentFolder.externalCommand.split(String.fromCharCode(92)).join(String.fromCharCode(47));
 
             $scope.editFolderModal();
         };
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1581,7 +1581,6 @@ angular.module('syncthing.core')
                 $scope.currentFolder.staggeredMaxAge = 365;
             }
             $scope.currentFolder.externalCommand = $scope.currentFolder.externalCommand || "";
-            $scope.currentFolder.externalCommand = $scope.currentFolder.externalCommand.split(String.fromCharCode(92)).join(String.fromCharCode(47));
 
             $scope.editFolderModal();
         };
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1666,7 +1666,7 @@ angular.module('syncthing.core')
                 folderCfg.versioning = {
                     'Type': 'external',
                     'Params': {
-                        'command': '' + folderCfg.externalCommand.replace(/\\(?! )/g, '/')
+                        'command': '' + folderCfg
                     }
                 };
                 delete folderCfg.externalFileVersioning;
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1581,7 +1581,6 @@ angular.module('syncthing.core')
                 $scope.currentFolder.staggeredMaxAge = 365;
             }
             $scope.currentFolder.externalCommand = $scope.currentFolder.externalCommand || "";
-
             $scope.editFolderModal();
         };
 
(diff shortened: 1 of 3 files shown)


calmh:
What problem does this resolve?

calmh:
This will have the side effect of showing forward slashes the next time the config is opened, which might be unexpected for the Windows user. I'd expect the backend to just not interpret backslashes as escapes on Windows.

AudriusButkevicius:
No, we need this to fix this in go or wherever it's broken.

AudriusButkevicius:
So I think the approach is wrong, as it still breaks under things like `c:\program files\foo.exe` as it would quote it half way.

Kropekk:
Paths with spaces should be quoted. If not, then it's impossible to distinguish between arguments and remainder of a path. I think we've already discussed it on IRC channel.

AudriusButkevicius:
Does this look correct to you?

https://play.golang.org/p/nwCeebIkNoN

AudriusButkevicius:
What I am saying is that I don't think we can do this for the user, we need to add a help message asking users to quote binary paths if they contain spaces.

AudriusButkevicius:
Have you verified that single quotes do not work?

Kropekk:
Yes, I did. More info may be found here: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_02

AudriusButkevicius:
Sorry, that doesn't explain why single quotes cannot be used?

```
	x := `'c:\program files\foo.exe' -user=x -pass=bar "something with spaces"`
	a, b := shellquote.Split(x)
	for i, v := range a {
		fmt.Println(i, v)
	}
	fmt.Println(b)
```

returns

```
0 c:\program files\foo.exe
1 -user=x
2 -pass=bar
3 something with spaces
<nil>
```

which to me says that single quotes are ok?

Kropekk:
I think that yesterday I had a counterexample, but I cannot recall it now. Today single quotes seem to be working as well :)

I wonder if we should warn the user about quotes in path. Specifically:
	If the path contains double quote and no single quotes, then it's okay.
	If the path contains single quote and no double quotes, then the whole path should surrounded by double-quotation marks.
	If the path contains both single and double quotes, then the whole path should be surrounded by double-quotation marks and double-quotation marks inside path should be escaped with backslash.
Is it enough to mention it in docs?

AudriusButkevicius:
I don't think we can do any reasonable quoting for the user, as I've already shown in my example.

AudriusButkevicius:
All we can do, is warn them with the help message.

Kropekk:
What do you mean? In which example? The one with path containing space? User is already notified that paths like these should be quoted.

AudriusButkevicius:
Right, hence I don't understand why we need go change.

Kropekk:
Because of the issue that I try to resolve? Right now c:\path\to\file.bat doesn't work.

AudriusButkevicius:
Right, that I think we can fix.
Just before `shellquote.Split` we should do:
```
if runtime.GOOS == "windows" {
    command = strings.Replace(command, `\`, `\\`, -1)
}
```

or something like that.

AudriusButkevicius:
I've raised https://github.com/kballard/go-shellquote/issues/8 in the meanwhile

AudriusButkevicius:
Can you clean up the siff and not remove random lines?

AudriusButkevicius:
Apart from the unclean diff and the build errors, this looks good.

Kropekk:
Sorry for the delay. I hope that now my pull request looks better.

Kropekk:
Sorry for the delay. I hope that my pull request looks better now.

calmh:
@AudriusButkevicius force merging stuff that fails the author check breaks the build. Please do not.

AudriusButkevicius:
Uhh, it just said it's red as someone requested changes.

calmh:
It will have looked like my PR I just opened looks and have had a large red merge button

PR #4666, opened 2018-01-12
Test and fix added for cases where the the root folder path is linked to via …

PrototypeNM1:
…a symlink for *nix and Windows.

Signed-off-by: Nicholas Rishel <PrototypeNM1@users.noreply.github.com>

### Purpose

Requested by @AudriusButkevicius in #4353 to demonstrate symlinks to the root sync directory failing on Windows. Opened the pull request to get feedback and direction, not expecting the code to be merged as-is.

### Testing

Open Windows command line as administrator for symlink creation privileges (admin redundant if you're running W10 in developer mode). Run `go run build.go test` and look for the error `Expected no error when root folder path is provided via a symlink: ...` If there is a way to run the walk.go tests more directly please let me know.

### Note

I only got as far as verifying the issue in the walk.go code, but I'm guessing this isn't the primary problem. Just as a test I added an !info.IsSymlink() check at https://github.com/syncthing/syncthing/blob/14df5211af7c4a6a3b0defd427173939fc78cebd/lib/scanner/walk.go#L489 (not committed) which successfully silenced the error but did not allow files to sync.

### DIFF
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -32,10 +32,10 @@ func newBasicFilesystem(root string) *BasicFilesystem {
 	root = cleanRoot(root)
 
 	// Check if root is a symlink to a directory
-	// filepath.EvalSymlinks does not support windows long filenames as of Go 1.9.2
+	// NOTE: filepath.EvalSymlinks does not support Windows long filenames as of Go 1.9.2
 	// thus symlink traversal needs to occur before adding long filename support
-	if symTarget, err := checkRootSymlink(root); err == nil {
-		root = cleanRoot(symTarget)
+	if target, err := filepath.EvalSymlinks(root); err == nil {
+		root = cleanRoot(target)
 	}
 
 	// Attempt to enable long filename support on Windows. We may still not
@@ -80,35 +80,6 @@ func cleanRoot(root string) string {
 	}
 
 	return root
-}
-
-func checkRootSymlink(root string) (string, error) {
-	// Check if root is a symlink
-	info, infoErr := underlyingLstat(root)
-	if infoErr != nil {
-		return root, infoErr
-	}
-	if info.Mode()&os.ModeSymlink == 0 {
-		return root, errors.New("Not a symlink")
-	}
-
-	// Find target of the symlink
-	// NOTE: does not support Windows long filenames as of Go 1.9.2
-	target, targetErr := filepath.EvalSymlinks(root)
-	if targetErr != nil {
-		return root, targetErr
-	}
-
-	// Verify the target is a directory
-	targetInfo, targetInfoErr := underlyingLstat(target)
-	if targetInfoErr != nil {
-		return root, targetInfoErr
-	}
-	if targetInfo.IsDir() {
-		// Successfully found target directory
-		return target, nil
-	}
-	return root, errors.New("Symlink is not a directory")
 }
 
 // rooted expands the relative path to the full path that is then used with os
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -29,6 +29,33 @@ type BasicFilesystem struct {
 }
 
 func newBasicFilesystem(root string) *BasicFilesystem {
+	root = cleanRoot(root)
+
+	// Check if root is a symlink to a directory
+	// filepath.EvalSymlinks does not support windows long filenames as of Go 1.9.2
+	// thus symlink traversal needs to occur before adding long filename support
+	if symTarget, err := checkRootSymlink(root); err == nil {
+		root = cleanRoot(symTarget)
+	}
+
+	// Attempt to enable long filename support on Windows. We may still not
+	// have an absolute path here if the previous steps failed.
+	if runtime.GOOS == "windows" {
+		if filepath.IsAbs(root) && !strings.HasPrefix(root, `\\`) {
+			root = `\\?\` + root
+		}
+		// If we're not on Windows, we want the path to end with a slash to
+		// penetrate symlinks. On Windows, paths must not end with a slash.
+	} else if root[len(root)-1] != filepath.Separator {
+		root = root + string(filepath.Separator)
+	}
+
+	return &BasicFilesystem{
+		root: root,
+	}
+}
+
+func cleanRoot(root string) string {
 	// The reason it's done like this:
 	// C:          ->  C:\            ->  C:\        (issue that this is trying to fix)
 	// C:\somedir  ->  C:\somedir\    ->  C:\somedir
@@ -52,21 +79,36 @@ func newBasicFilesystem(root string) *BasicFilesystem {
 		}
 	}
 
-	// Attempt to enable long filename support on Windows. We may still not
-	// have an absolute path here if the previous steps failed.
-	if runtime.GOOS == "windows" {
-		if filepath.IsAbs(root) && !strings.HasPrefix(root, `\\`) {
-			root = `\\?\` + root
-		}
-		// If we're not on Windows, we want the path to end with a slash to
-		// penetrate symlinks. On Windows, paths must not end with a slash.
-	} else if root[len(root)-1] != filepath.Separator {
-		root = root + string(filepath.Separator)
+	return root
+}
+
+func checkRootSymlink(root string) (string, error) {
+	// Check if root is a symlink
+	info, infoErr := underlyingLstat(root)
+	if infoErr != nil {
+		return root, infoErr
+	}
+	if info.Mode()&os.ModeSymlink == 0 {
+		return root, errors.New("Not a symlink")
 	}
 
-	return &BasicFilesystem{
-		root: root,
+	// Find target of the symlink
+	// NOTE: does not support Windows long filenames as of Go 1.9.2
+	target, targetErr := filepath.EvalSymlinks(root)
+	if targetErr != nil {
+		return root, targetErr
 	}
+
+	// Verify the target is a directory
+	targetInfo, targetInfoErr := underlyingLstat(target)
+	if targetInfoErr != nil {
+		return root, targetInfoErr
+	}
+	if targetInfo.IsDir() {
+		// Successfully found target directory
+		return target, nil
+	}
+	return root, errors.New("Symlink is not a directory")
 }
 
 // rooted expands the relative path to the full path that is then used with os
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -29,33 +29,6 @@ type BasicFilesystem struct {
 }
 
 func newBasicFilesystem(root string) *BasicFilesystem {
-	root = cleanRoot(root)
-
-	// Check if root is a symlink to a directory
-	// filepath.EvalSymlinks does not support windows long filenames as of Go 1.9.2
-	// thus symlink traversal needs to occur before adding long filename support
-	if symTarget, err := checkRootSymlink(root); err == nil {
-		root = cleanRoot(symTarget)
-	}
-
-	// Attempt to enable long filename support on Windows. We may still not
-	// have an absolute path here if the previous steps failed.
-	if runtime.GOOS == "windows" {
-		if filepath.IsAbs(root) && !strings.HasPrefix(root, `\\`) {
-			root = `\\?\` + root
-		}
-		// If we're not on Windows, we want the path to end with a slash to
-		// penetrate symlinks. On Windows, paths must not end with a slash.
-	} else if root[len(root)-1] != filepath.Separator {
-		root = root + string(filepath.Separator)
-	}
-
-	return &BasicFilesystem{
-		root: root,
-	}
-}
-
-func cleanRoot(root string) string {
 	// The reason it's done like this:
 	// C:          ->  C:\            ->  C:\        (issue that this is trying to fix)
 	// C:\somedir  ->  C:\somedir\    ->  C:\somedir
@@ -79,36 +52,21 @@ func cleanRoot(root string) string {
 		}
 	}
 
-	return root
-}
-
-func checkRootSymlink(root string) (string, error) {
-	// Check if root is a symlink
-	info, infoErr := underlyingLstat(root)
-	if infoErr != nil {
-		return root, infoErr
-	}
-	if info.Mode()&os.ModeSymlink == 0 {
-		return root, errors.New("Not a symlink")
-	}
-
-	// Find target of the symlink
-	// NOTE: does not support Windows long filenames as of Go 1.9.2
-	target, targetErr := filepath.EvalSymlinks(root)
-	if targetErr != nil {
-		return root, targetErr
+	// Attempt to enable long filename support on Windows. We may still not
+	// have an absolute path here if the previous steps failed.
+	if runtime.GOOS == "windows" {
+		if filepath.IsAbs(root) && !strings.HasPrefix(root, `\\`) {
+			root = `\\?\` + root
+		}
+		// If we're not on Windows, we want the path to end with a slash to
+		// penetrate symlinks. On Windows, paths must not end with a slash.
+	} else if root[len(root)-1] != filepath.Separator {
+		root = root + string(filepath.Separator)
 	}
 
-	// Verify the target is a directory
-	targetInfo, targetInfoErr := underlyingLstat(target)
-	if targetInfoErr != nil {
-		return root, targetInfoErr
+	return &BasicFilesystem{
+		root: root,
 	}
-	if targetInfo.IsDir() {
-		// Successfully found target directory
-		return target, nil
-	}
-	return root, errors.New("Symlink is not a directory")
 }
 
 // rooted expands the relative path to the full path that is then used with os
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -32,10 +32,10 @@ func newBasicFilesystem(root string) *BasicFilesystem {
 	root = cleanRoot(root)
 
 	// Check if root is a symlink to a directory
-	// NOTE: filepath.EvalSymlinks does not support Windows long filenames as of Go 1.9.2
+	// filepath.EvalSymlinks does not support windows long filenames as of Go 1.9.2
 	// thus symlink traversal needs to occur before adding long filename support
-	if target, err := filepath.EvalSymlinks(root); err == nil {
-		root = cleanRoot(target)
+	if symTarget, err := checkRootSymlink(root); err == nil {
+		root = cleanRoot(symTarget)
 	}
 
 	// Attempt to enable long filename support on Windows. We may still not
@@ -80,6 +80,35 @@ func cleanRoot(root string) string {
 	}
 
 	return root
+}
+
+func checkRootSymlink(root string) (string, error) {
+	// Check if root is a symlink
+	info, infoErr := underlyingLstat(root)
+	if infoErr != nil {
+		return root, infoErr
+	}
+	if info.Mode()&os.ModeSymlink == 0 {
+		return root, errors.New("Not a symlink")
+	}
+
+	// Find target of the symlink
+	// NOTE: does not support Windows long filenames as of Go 1.9.2
+	target, targetErr := filepath.EvalSymlinks(root)
+	if targetErr != nil {
+		return root, targetErr
+	}
+
+	// Verify the target is a directory
+	targetInfo, targetInfoErr := underlyingLstat(target)
+	if targetInfoErr != nil {
+		return root, targetInfoErr
+	}
+	if targetInfo.IsDir() {
+		// Successfully found target directory
+		return target, nil
+	}
+	return root, errors.New("Symlink is not a directory")
 }
 
 // rooted expands the relative path to the full path that is then used with os
--- a/lib/fs/walkfs.go
+++ b/lib/fs/walkfs.go
@@ -79,10 +79,9 @@ func (f *walkFilesystem) walk(path string, info FileInfo, walkFn WalkFunc) error
 // and directories are filtered by walkFn. The files are walked in lexical
 // order, which makes the output deterministic but means that for very
 // large directories Walk can be inefficient.
-// Walk follows root symbolic links (in all cases on Windows, with trailing
-// '/' on *nix).
+// Walk does not follow symbolic links.
 func (f *walkFilesystem) Walk(root string, walkFn WalkFunc) error {
-	info, err := f.Stat(root)
+	info, err := f.Lstat(root)
 	if err != nil {
 		return walkFn(root, nil, err)
 	}
--- a/lib/fs/walkfs.go
+++ b/lib/fs/walkfs.go
@@ -46,7 +46,7 @@ func (f *walkFilesystem) walk(path string, info FileInfo, walkFn WalkFunc) error
 		return err
 	}
 
-	if !info.IsDir() {
+	if !info.IsDir() && path != "." {
 		return nil
 	}
 
--- a/lib/fs/walkfs.go
+++ b/lib/fs/walkfs.go
@@ -79,9 +79,10 @@ func (f *walkFilesystem) walk(path string, info FileInfo, walkFn WalkFunc) error
 // and directories are filtered by walkFn. The files are walked in lexical
 // order, which makes the output deterministic but means that for very
 // large directories Walk can be inefficient.
-// Walk does not follow symbolic links.
+// Walk follows root symbolic links (in all cases on Windows, with trailing
+// '/' on *nix).
 func (f *walkFilesystem) Walk(root string, walkFn WalkFunc) error {
-	info, err := f.Lstat(root)
+	info, err := f.Stat(root)
 	if err != nil {
 		return walkFn(root, nil, err)
 	}
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -2071,7 +2071,7 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 		})
 
 		if iterError != nil {
-			l.Debugln("Stopping scan of folder %s due to: %s", folderCfg.Description(), err)
+			l.Debugln("Stopping scan of folder %s due to: %s", folderCfg.Description(), iterError)
 			return iterError
 		}
 	}
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1970,6 +1970,10 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 		UseWeakHashes:         weakhash.Enabled,
 	})
 
+	if err := runner.CheckHealth(); err != nil {
+		return err
+	}
+
 	batch := make([]protocol.FileInfo, 0, maxBatchSizeFiles)
 	batchSizeBytes := 0
 	changes := 0
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1954,7 +1954,7 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 
 	runner.setState(FolderScanning)
 
-	fchan, err := scanner.Walk(ctx, scanner.Config{
+	fchan := scanner.Walk(ctx, scanner.Config{
 		Folder:                folderCfg.ID,
 		Subs:                  subDirs,
 		Matcher:               ignores,
@@ -1969,17 +1969,6 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 		ProgressTickIntervalS: folderCfg.ScanProgressIntervalS,
 		UseWeakHashes:         weakhash.Enabled,
 	})
-
-	if err != nil {
-		// The error we get here is likely an OS level error, which might not be
-		// as readable as our health check errors. Check if we can get a health
-		// check error first, and use that if it's available.
-		if ferr := runner.CheckHealth(); ferr != nil {
-			err = ferr
-		}
-		runner.setError(err)
-		return err
-	}
 
 	batch := make([]protocol.FileInfo, 0, maxBatchSizeFiles)
 	batchSizeBytes := 0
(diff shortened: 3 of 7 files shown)


imsodin:
Thanks for the reproducer, the issue is pretty simple: #4545 fixes `FolderConfiguration.CheckPath` but the scanner package does not have access to that and thus uses it's own `checkDir` function which fails on symlinks. As we test the folder path in the model before scanning, I think we can simply drop this check in the scanner.

calmh:
Yes

imsodin:
You can combine the two tests into a single test and just skip in case symlinks don't work on windows similar to how it's done here: https://github.com/syncthing/syncthing/blob/445c4edeca8f70bec8ae02a17db234ed06e88d62/lib/config/config_test.go#L464

Also I just noticed your comment that you already circumvented this problem, but it still doesn't start to sync. I don't immediately see a similar problem nearby. Anything in the log?

PrototypeNM1:
@imsodin Nothing stood out in the logs. Could you check 3c6ea0f, I wasn't anticipating the test failure for *nix given that symlink to root works perfectly fine in Ubuntu.

lib/scanner/walk_test.go
```diff
@@ -366,6 +366,43 @@ func TestWalkSymlinkWindows(t *testing.T) {
...
+
+	link := "link"
+
+	os.Remove(link)
+	defer os.Remove(link)
+
+	if err := osutil.DebugSymlinkForTestsOnly("testdata/dir1", link); err != nil && runtime.GOOS == "windows" {
+		// Probably we require permissions we don't have.
+		t.Skip("Need admin permissions or developer mode to run symlink test on Windows: " + err.Error())
+	}
```
AudriusButkevicius: What if error non nil, but we not on Windows?

lib/scanner/walk_test.go
```diff
@@ -366,6 +366,43 @@ func TestWalkSymlinkWindows(t *testing.T) {
 	}
 }
 
+func TestWalkRootSymlink(t *testing.T) {
+
+	// Create a folder with a symlink to it
+
+	link := "link"
```
AudriusButkevicius: Let's do this in a temp directory.

lib/scanner/walk_test.go
```diff
@@ -366,6 +366,43 @@ func TestWalkSymlinkWindows(t *testing.T) {
 	}
 }
 
+func TestWalkRootSymlink(t *testing.T) {
+
+	// Create a folder with a symlink to it
```
AudriusButkevicius: Too many blanks lines everywhere, esp around comments from both sides.

lib/scanner/walk_test.go
PrototypeNM1: I was following suite with the above referenced https://github.com/syncthing/syncthing/blob/445c4edeca8f70bec8ae02a17db234ed06e88d62/lib/config/config_test.go#L464. The only reason to check for the error is because of Window's unique behavior of requiring admin privileges or developer mode to create symlinks and this test is more a way to notify that the testing environment hasn't been set up correctly. Checking for errors on other platforms is testing an external dependency.

AudriusButkevicius: Still, if this fails on non-windows, the test probably will fail anyway, so we should abort early.

lib/scanner/walk_test.go
PrototypeNM1: This function was mostly a copy/paste of the TestWalkSymlink[Windows/Unix]() functions with modifications but mostly retained style/spacing. I left the pull request open to edits from maintainers so it would probably be easiest for you to shore up styling.

lib/scanner/walk_test.go
PrototypeNM1: Might be worth revisiting the TestWalkSymlink\Windows/Unix\]() functions in that case too.

PrototypeNM1:
I'm a bit confused as to how scanner fits into Syncthing as a whole. When I added the test for number of files in the directory it introduced a test failure on Mac and Linux even though the most recent Linux release detects the files in a symlinked root directory perfectly fine.

lib/scanner/walk_test.go
AudriusButkevicius: So don't take this as personal offence or something like that. It's more than likely that the existing tests code has issues. I am just pointing out things I see.

lib/scanner/walk.go
```diff
@@ -486,10 +485,6 @@ func (w *walker) checkDir() error {
 		return err
 	}
 
-	if !info.IsDir() {
-		return errors.New(w.Filesystem.URI() + ": not a directory")
-	}
```
imsodin: You can remove the entire `checkDir` function.

lib/scanner/walk_test.go
```diff
@@ -366,6 +366,43 @@ func TestWalkSymlinkWindows(t *testing.T) {
...
+		// Probably we require permissions we don't have.
+		t.Skip("Need admin permissions or developer mode to run symlink test on Windows: " + err.Error())
+	}
+
+	// Scan it
+
+	fchan, err := Walk(context.TODO(), Config{
+		Filesystem: fs.NewFilesystem(fs.FilesystemTypeBasic, link),
+		BlockSize:  128 * 1024,
+	})
```
imsodin: The tests fail because you are missing `Hashers`. But you can use `walkDir` anyway, which takes care of such things and you don't need to loop over fchan later on:
```
files, err := walkDir(fs.NewFilesystem(fs.FilesystemTypeBasic, link), ".")
```

lib/scanner/walk_test.go
PrototypeNM1: Actually, it looks like Go 1.4 added support for Windows symlinks - https://github.com/golang/go/commit/cf521ce64f50c4f300294d6b649f34eca87bb8a3. Scrolling through the commit it looks like they are also testing for long path names. I'll redactor that add a test.

lib/scanner/walk_test.go
PrototypeNM1: No worries, not taken personally. :) The question should have asked is whether it would be helpful if I modified those tests while looking a this file? The reason I haven't is because it's out of scope of this pull request and I'm a bit uncomfortable modifying tests that are otherwise known to work correctly.

AudriusButkevicius: I think do what you are doing here, and if you want to make other cleanups, file a separate PR.
This way if there is some disagreement or discussions around the refactoring, that doesn't get in the way of getting this merged.

PrototypeNM1:
@imsodin Could you check that the failing tests can be safely removed given the prior go ahead to remove checkDir()?

The current status is that the sync folder's .stfolder is created successfully, but the UI indicates there are no files/folders in the sync folder.

imsodin:
Sure, you can remove `TestWalkError`.

Logs with `STTRACE=scanner,model` should show where the remaining problem lies.

lib/scanner/walk_test.go
```diff
@@ -366,6 +367,36 @@ func TestWalkSymlinkWindows(t *testing.T) {
...
+	// Create a folder with a symlink in it
+	tmp, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmp)
+
+	link := tmp + "/link"
+	dest, _ := filepath.Abs("testdata/dir1")
+	if err := os.Symlink(dest, link); err != nil {
```
imsodin: Why don't you use `DebugSymlinkForTestsOnly` anymore?

AudriusButkevicius: Apart from this, the PR looks good to me.

PrototypeNM1: That's a mistake, I had misremembered @calmh's open issue for Go https://github.com/golang/go/issues/22874 as a Syncthing issue that Go now supported Windows symlinks.

PrototypeNM1: Using `os.Symlink` was definitively a bug and shows that the issue could still be in scanner. `os.Symlink` was triggering the `t.Skip` case, switching back to `DebugSymlinkForTestsOnly` resurfaces that scanner is not detecting the files.

@AudriusButkevicius is there a reason an err on creating symlinks on Windows shouldn't trigger a failure? I know it's that way in config_test.go's `TestFolderCheckPath` and prior walk_test.go tests. Is this an issue for testing on older versions of Windows?

lib/scanner/walk_test.go
```diff
@@ -121,27 +122,6 @@ func TestWalk(t *testing.T) {
...
-
-	_, err = Walk(context.TODO(), Config{
-		Filesystem: fs.NewFilesystem(fs.FilesystemTypeBasic, "testdata/bar"),
-		BlockSize:  128 * 1024,
-	})
-
-	if err == nil {
-		t.Error("no error from non-directory")
-	}
-}
```
AudriusButkevicius: What does this test fail with?
It's scary if this starts passing now. I would still expect an error at some point.

AudriusButkevicius:
Sorry to throw you around, but I now think that we should keep checkDir given that test now fails. We should probably just adjust checkDir to succeed on symlinks at root.

AudriusButkevicius:
Perhaps you are right, I haven't checked.

imsodin:
I have checked: `CheckHealth` happens before `Walk` and that includes checking the path. Also besides that I think the check is pointless, I consider it bad to implement a duplicate `FolderConfiguration.CheckPath` in the scanner package. This can lead to exactly this situation, where something (here: symlinks) is fixed in one place but forgotten in the other. We should get to a consensus on this, otherwise it's pretty annoying for @PrototypeNM1 if he gets conflicting advice.

AudriusButkevicius:
Fine, sorry @PrototypeNM1, @imsodin is right, lets get rid of that and the test :(

PrototypeNM1:
@AudriusButkevicius no worries.

Running with STTRACE=scanner I get the following log:
[7GOKR] 2018/01/14 18:33:22.989656 walk.go:101: DEBUG: Walk [] 131072 &{0xc042141230 [] [] false <nil> d41d8cd98f00b204e9800998ecf8427e 0xc04218a840 0xc042076160 0xc042056728}
[7GOKR] 2018/01/14 18:33:22.990626 walk.go:164: DEBUG: Walk progress done default [] 131072 &{0xc042141230 [] [] false <nil> d41d8cd98f00b204e9800998ecf8427e 0xc04218a840 0xc042076160 0xc042056728}

PrototypeNM1:
Commenting out https://github.com/PrototypeNM1/syncthing/blob/fecb21cdb119332dcb39c72f08cf2f5f10e7692c/lib/fs/walkfs.go#L49-L51 "fixes" the issue. So detecting root here or in the single calling function `Walk()` and checking if it is a symlink to a directory should be sufficient to fix the issue. That said, it looks like there is no guarantee that `Walk` is called with a root directory given it's usage in walk.go https://github.com/PrototypeNM1/syncthing/blob/fecb21cdb119332dcb39c72f08cf2f5f10e7692c/lib/scanner/walk.go#L115-L121.

AudriusButkevicius:
I was proposing that we add a check if it's a symlink, and allow that, yet as @imsodin pointed out, we should probably revert that revert and get rid of checkDir all together.

PrototypeNM1:
@AudriusButkevicius `checkDir` is removed, the code I linked to is a directory check in the recursive walk through the folder. Sorry, I made things confusing by winding back the reverted code commits and force pushing to github (hence the change isn't showing in this conversation).

@imsodin my last message outlines the spot in code which is preventing traversal into a root directory symlink, at first glance I'm not sure how to fix the problem as it looks like `Walk` is not guaranteed to be called from the root directory given the test code (unless I'm mis-reading the test code).

lib/scanner/walk.go
```diff
@@ -101,10 +100,6 @@ type walker struct {
 func (w *walker) walk(ctx context.Context) (chan protocol.FileInfo, error) {
```
imsodin: This (and `Walk`) can now drop the `error` return value (all returns just pass nil).

imsodin:
Ok, so this looks like on windows `IsDir` on `os.FileInfo` of a link to a directory returns false, hence the walk fs does not descend into it, but on Linux and Mac it returns true. That seems like a Go bug/inconsistency and I seem to remember both @calmh and @AudriusButkevicius battling with something similar recently - do you know anything about this?

PrototypeNM1:
@imsodin I'm also guessing that https://github.com/PrototypeNM1/syncthing/blob/a9fd2d90f5fa1d879872c60529cf226a422b2067/lib/scanner/walk.go#L261-L269 will probably become a bug for Windows if the behavior for /lib/fs/walkfs.go changes. So both will probably need to be modified in parallel.

AudriusButkevicius:
@imsodin it's a os inconsistency, windows symlinks are typed.

imsodin:
Great :) In walkfs.go it is specified that `Walk` does not follow symbolic links:
```
// Walk does not follow symbolic links.
func (f *walkFilesystem) Walk(root string, walkFn WalkFunc) error {
```
But evidently (as the test on linux/mac passes), this is only true on Windows. So do we want the filesystem `Walk` function to descend into symlinks, so it is up to the caller and the given `walkFn` to skip symlinks, or should it really not descend into symlinks?

AudriusButkevicius:
It should not descend period, on all platforms. On windows it shouldn't even see them.

PrototypeNM1:
The reason's working on Linux right now is because walkFn is returning early for root directories at https://github.com/PrototypeNM1/syncthing/blob/a9fd2d90f5fa1d879872c60529cf226a422b2067/lib/scanner/walk.go#L221-L223. Again though I'm not sure how to verify if root is actually root of the sync folder. If it is though would the workaround be to re-implement `IsDir` for `FileInfo`?

AudriusButkevicius:
Right, but that return just continues iteration. Root being a symlink should be ok.

imsodin:
The remaining problem is entirely in the filesystem package at https://github.com/syncthing/syncthing/blob/master/lib/fs/walkfs.go#L49 . Well more generally it is a problem in all places where `IsDir` is used alone, as its return value is inconsistent between windows and linux/mac. We should either define at interface level what `IsDir` is supposed to return (i.e. should it return true or false for a symlink pointing to a directory) and fix it in the filesystem implementation or we always need to check both `IsDir` and `IsSymlink` in the walkFilesystem implementation.

AudriusButkevicius:
pinging @calmh for opinion.

calmh:
I'm at a tablet only so not giving a relevant opinion at the moment but will. I'll just throw out there that one relevant difference between Windows and Unix is that on Unix "symlink" will stat as a symlink while "symlink/" stats at the directory on the other side. On windows both stat as symlinks. We make sure to add a trailing slash to the folder path, which gives us the right behavior automatically on Unix but not windows.

PrototypeNM1:
I implemented the `Stat` in `fs.Walk` based solution for your consideration.

AudriusButkevicius:
This looks fine to me, but I'll leave for others to verify my judgement.

AudriusButkevicius:
@st-review lgtm

st-review:

imsodin:
@st-review don't

This allows descending into arbitrary symlinks at will - I proposed (dangerous) BS. See this updated test:
```diff
--- a/lib/scanner/walk_test.go
+++ b/lib/scanner/walk_test.go
@@ -270,30 +270,24 @@ func TestWalkSymlinkUnix(t *testing.T) {
 	defer os.RemoveAll("_symlinks")
 
 	os.Mkdir("_symlinks", 0755)
-	os.Symlink("destination", "_symlinks/link")
+	os.Symlink("../testdata", "_symlinks/link")
 
-	// Scan it
+	for _, path := range []string{".", "link"} {
+		// Scan it
 
-	fchan := Walk(context.TODO(), Config{
-		Filesystem: fs.NewFilesystem(fs.FilesystemTypeBasic, "_symlinks"),
-		BlockSize:  128 * 1024,
-	})
+		files, _ := walkDir(fs.NewFilesystem(fs.FilesystemTypeBasic, "_symlinks"), path)
 
-	var files []protocol.FileInfo
-	for f := range fchan {
-		files = append(files, f)
-	}
-
-	// Verify that we got one symlink and with the correct attributes
+		// Verify that we got one symlink and with the correct attributes
 
-	if len(files) != 1 {
-		t.Errorf("expected 1 symlink, not %d", len(files))
-	}
-	if len(files[0].Blocks) != 0 {
-		t.Errorf("expected zero blocks for symlink, not %d", len(files[0].Blocks))
-	}
-	if files[0].SymlinkTarget != "destination" {
-		t.Errorf("expected symlink to have target destination, not %q", files[0].SymlinkTarget)
+		if len(files) != 1 {
+			t.Errorf("expected 1 symlink, not %d", len(files))
+		}
+		if len(files[0].Blocks) != 0 {
+			t.Errorf("expected zero blocks for symlink, not %d", len(files[0].Blocks))
+		}
+		if files[0].SymlinkTarget != "../testdata" {
+			t.Errorf("expected symlink to have target destination, not %q", files[0].SymlinkTarget)
+		}
 	}
 }
 ```

Another idea to solve this: We could resolve symlinks in the root path when creating the filesystem. We already try to make it absolute.

**Comment by st-review** [comment]
@imsodin: Preventing merge for the time being. Push a new revision to reset!

**Comment by PrototypeNM1** [comment]
@imsodin Should I copy/paste the new test over or could you push your changes to my root_symlink_test branch?

**Comment by imsodin** [comment]
Easiest for me would be if you could just `git apply` the diff. Also this does not cover your test, I would not remove that.

**Comment by PrototypeNM1** [comment]
@imsodin @AudriusButkevicius Could you review e6f84e6?

File: lib/fs/basicfs.go [hunk 9]
```diff
@@ -52,21 +79,36 @@ func newBasicFilesystem(root string) *BasicFilesystem {
...
+}
+
+func checkRootSymlink(root string) (string, error) {
+	// Check if root is a symlink
+	info, infoErr := underlyingLstat(root)
+	if infoErr != nil {
+		return root, infoErr
+	}
+	if info.Mode()&os.ModeSymlink == 0 {
+		return root, errors.New("Not a symlink")
```
imsodin: As later on we do not actually act on an error here, why do you do this symlink check (and later on the directory check)?

lib/fs/basicfs.go
```diff
@@ -29,6 +29,33 @@ type BasicFilesystem struct {
 }
 
 func newBasicFilesystem(root string) *BasicFilesystem {
+	root = cleanRoot(root)
```
imsodin: Does the symlink check potentially choke on a non-cleaned path? Otherwise this could just be moved after the symlink check to avoid calling it twice.

lib/fs/basicfs.go
PrototypeNM1: It's probably safe to skip this check, using error on `EvalSymlink` as indication that this isn't a symlink instead. The latter directory check is also likely unnecessary since if it is a symlink and not a directory, than it will return an error later in path checking. I'll go ahead and slim this code.

lib/fs/basicfs.go
PrototypeNM1: I'm not familiar enough with the code base to answer this with confidence, but my suspicion is that it could for the same reasons that make the root cleaning necessary. Is the purpose of cleaning the root directory for Syncthing's internal assumptions for path handling or to ensure that calls to Go's filepath succeeds?

PrototypeNM1: The relevant commit for the code moved to cleanRoot is https://github.com/syncthing/syncthing/commit/3d8b4a42b72de26f6e15fdc2b2a7feeb116f9577

@AudriusButkevicius Could you add context for the path cleaning code and whether this 2 step check makes sense.

imsodin: In hindsight I am actually pretty sure you're right, at least for the tilde expansion. The other two `cleanRoot` changes (absolutification and the C:\ stuff) may or may not be necessary, but they are cheap, so unless Audrius can says with certainty that they are not necessary, lets just be save and use it twice as you do already.

AudriusButkevicius: We should do it twice, in case eval symlinks returns a relative path.

PrototypeNM1:
The current test failures seem to derive from a very minor bug in Go's `EvaluateSymlink` Windows implementation: paths which consistent of only the drive letter - e.g. c:\ - returns without error. This breaks the tests when the drive letter is capitalized, probably in their path cleanup code.

imsodin:
I had a look at the actual filepath implementation and it is not the "normal filepath cleanup" (`Clean`), but only `EvalSymlinks` that does path "normalization" on windows, which capitalises the drive letter. I believe this does not matter and having consistent drive letter case is actually a nice side-effect. So changing the failing tests to upper-case drive letters should be fine, but I would like to hear what @AudriusButkevicius thinks first.

AudriusButkevicius:
What do I think on what matter?

imsodin:
Using `filepath.EvalSymlinks` on the fs root being unproblematic. It converts e.g. `e:` to `E:`.

AudriusButkevicius:
That shouldn't matter, yet I fail to understand why we are evaling symlink there?

PrototypeNM1:
@AudriusButkevicius Mostly because I mistakenly thought that EvalSymlinks would error when the path did not contain a symlink, hence I removed the symlink checks seen in `checkRootSymlink` in [this commit](https://github.com/syncthing/syncthing/pull/4666/commits/e6f84e68fcc9415f74f2e56086142a1754b54eda#diff-1f813992a41e8e15fc531c77ee2c0f76R85) and instead just used the result of `EvalSymlinks`  if there is no error.

I think skipping the checks is cleaner and unlikely to have noticeable impact on performance. If having the checks is preferable let me know.

imsodin:
> That shouldn't matter, yet I fail to understand why we are evaling symlink there?

@AudriusButkevicius That is by now the essence of this PR, the rest are just various fixes unearthed during debugging: If the root path is not symlink evaluated (well strictly speaking only the last component of it), walking will abort because it does not descend into symlinks. To prevent that, this resolves symlinks in the root path internally (as opposed to adding hacks to the walk function to treat the root directory of a filesystem specially).

AudriusButkevicius:
But when we walk, we already walk behind the link, so wether there is a link at the root or not, it should not matter, as that is user controlled. We only care about symlinks in our walled garden.

imsodin:
Yes, we don't care and that's why it is ok to use `EvalSymlinks`. It is a problem because if `/foo/bar` is the root of the folder and also is a symlink to `/foo/linked` then `fs.Walk(".")` will abort because `fs.Lstat(".")` returns the fileinfo of the symlink, which we do not descend into, as in our walled garden, we do not trust symlinked directories.

AudriusButkevicius:
We should not complain about symlinks at the root, just deeper down. I don't think EvalSymlinks adds any value here.

imsodin:
That means changing `fs.Walk` at https://github.com/syncthing/syncthing/blob/master/lib/fs/walkfs.go#L84 to something like

```
info, err = f.LStat(root)
if root == "." && info.IsSymlink() {
    info, err = f.Stat(root)
}
```
That's what I called a hack earlier and I would prefer `EvalSymlink` on filesystem creation over this. Would you prefer this or do you have a more elegant idea to allow symlinked root?

AudriusButkevicius:
I still think this is a hack, why can't se just Lstat, and if it's a symlink and path == ".", let it through?

imsodin:
It can't just be passed through in that case, because `walk` needs a valid non-symlink directory info. The check could be moved to the `walk` function, such that on descending into every directory (calling `walk` on it) do an (unnecessary for subpaths) check for `path == "."`. I consider this even worse then the hack I wrote above and I would prefer just resolving symlinks in the fs root path over both (but I think I mentioned that before).

AudriusButkevicius:
What info do we actually need about a directory other than it's existence? Especially on the root... We don't even sync permissions on the root, so why do we need valid stat info for the root?

If we scan a set of subdirectories, rather than root, the path will never be ".".

imsodin:
> If we scan a set of subdirectories, rather than root, the path will never be ".".

Every time the rescan timer fires or anyone hits the rescan button in the web UI a scan on "." is triggered.

Honestly, I personally don't see the problem with evaluating symlinks on roots (in part because it would forestall the entire discussion we are having). A filesystem walk by definition calls a walk-function on every valid encountered item, one of its argument being the file info. What our scanner doesn't need of these infos should not be a reason to omit them, at least not if there is another way. And no matter what we actually do after, adding a `== "."`  condition into every invocation of `walk` (called on every dir encountered) is hacky to me.

Honestly, what I am saying, is if you disagree, that evaluating symlinks on root solves the problem while not causing any problems, please provide an alternative solution or a concrete problem with the existing solution.

AudriusButkevicius:
I am not saying that evaluating symlinks is causing issues. I am saying it causes a side effect for users, whereby errors suddently might look like they are coming from a different place than where the folder is.

Is your concern the fact that we're adding an extra if string equals that will have to be evaluated on every item of the walk? I suspect performance impact of that is close to non-existant. We could also modify walk implementation for "x" not to walk "x" itself, yet that's semantics change which is ugly.

imsodin:
> I am saying it causes a side effect for users, whereby errors suddently might look like they are coming from a different place than where the folder is.

Thanks, I didn't get that before.

@PrototypeNM1 so to conclude that lengthy discussions. lets try adding `&& path != "."` to https://github.com/syncthing/syncthing/blob/master/lib/fs/walkfs.go#L49 instead of root manipulation.

PrototypeNM1:
Is that Mac test error a red herring?

imsodin:
@st-review lgtm

st-review:
@imsodin: Build status is `failure` -- refusing to merge.

lib/model/model.go
```diff
@@ -1970,17 +1970,6 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
...
-	if err != nil {
-		// The error we get here is likely an OS level error, which might not be
-		// as readable as our health check errors. Check if we can get a health
-		// check error first, and use that if it's available.
-		if ferr := runner.CheckHealth(); ferr != nil {
-			err = ferr
-		}
-		runner.setError(err)
-		return err
-	}
```
AudriusButkevicius: Should we not keep this given checkDir is going away?

imsodin: There could be another 
```golang
 	if err := runner.CheckHealth(); err != nil {
 		return err
 	}
```
here to abort a little bit earlier in case the scanner has scanned some stuff before the folder health problem arose.

AudriusButkevicius: Scanning doesn't matter much, as there are periodic checks as the results arrive after the scan before stuff goes into the db.

imsodin: I should have been more explicit, this was what I meant: There was a health check before starting the scanner and this removed block of code is just polishing the error from now removed `checkDir`, so the only benefit of keeping a health check there is, that we may return slightly earlier instead of before sending to db or fchan was closed.

AudriusButkevicius: Sure, yet I think there is still value to have one of the checks, as it seems we're axe'ing both.

lib/fs/walkfs.go
```diff
@@ -46,7 +46,7 @@ func (f *walkFilesystem) walk(path string, info FileInfo, walkFn WalkFunc) error
 		return err
 	}
 
-	if !info.IsDir() {
+	if !info.IsDir() && path != "." {
```
AudriusButkevicius: Why aren't we doing it in the scanner?
This changes semantics of what os.Walk would do, which I am not a fan off.

AudriusButkevicius:
This is like the worst PR syncthing ever had, we've disagreed, argued and changed direction like ten times by now and poor @PrototypeNM1 had to deal with our nonsense for nearly two weeks now. Doing and undoing random changes that we come up with.

AudriusButkevicius:
@PrototypeNM1 so me and @imsodin had a chat offline, to stop throwing you around with our disagreement, and agreed that these are the things left to do:

1. Keep the runner.CheckHealth() check before the walk
2. You need to add yourself to AUTHORS/NICKS (does that still exist?) file

PrototypeNM1:
No worries on the meandering, most of my recent dev work is by myself so this has been a fun aside. 😆 

To clarify on point 1, do you mean in [model.go](https://github.com/syncthing/syncthing/pull/4666/files/b9d36126c123bdfe1c68c913a0d3b7a2009ea0e6#diff-1df9805b2381b6288b220226f82d5dc9) after `scanner.Walk`?

AudriusButkevicius:
Yes, there is no longer an error case, but we should do a health check just in case.

AudriusButkevicius:
@st-review lgtm

st-review:
@AudriusButkevicius: Merge failed:

```
$ git fetch -f origin refs/pull/4666/head:pr-4666
From https://github.com/syncthing/syncthing
 * [new ref]         refs/pull/4666/head -> pr-4666
From https://github.com/syncthing/syncthing
   93b5180..885e3f1  master     -> orig/master
$ git reset --hard
HEAD is now at 885e3f1 lib/ignores: Update lines even if patterns didn't change (fixes #4689)
$ git checkout master
Already on 'master'
HEAD is now at 885e3f1 lib/ignores: Update lines even if patterns didn't change (fixes #4689)
$ git clean -fxd
$ git merge --squash --no-commit pr-4666
Updating 885e3f1..a9d370d
Fast-forward
 AUTHORS                        |   1 +
 lib/fs/walkfs.go               |   2 +-
 lib/model/model.go             |  13 +---
 lib/scanner/infinitefs_test.go |   4 ++
 lib/scanner/walk.go            |  28 ++-------
 lib/scanner/walk_test.go       | 137 ++++++++++++++++-------------------------
 6 files changed, 65 insertions(+), 120 deletions(-)
$ git commit -F -
Commit message subject:

    Test added for cases where the the root folder path is linked to via a symlink for *nix and Windows.

match this pattern:

    ^[\w/,\. ]+: \w

```

AudriusButkevicius:
@st-review lgtm

lib/scanner: Support walking a symlink root (ref #4353)

st-review:
:ok_hand: Merged as a505231774ff11a40640efd601249d627cf3a351. Thanks, @PrototypeNM1!

Catfriend1:
Hi, according to the issue and logs in https://github.com/syncthing/syncthing-android/issues/1032 I suspect this to cause the showstopper for syncthing android we currently experience. There, directory root of a syncthing folder is scanned twice. Now, it is no longer possible syncing the internal phone memory which worked ok with syncthing 14.43 or lower.

PR #4669, opened 2018-01-13
gui: Count deleted items for remote out of sync items display (fixes #4668)

### LINKED ISSUE #4668: Remote device out of sync items shows "0 items, ~0 B"
This was reported in the forum: https://forum.syncthing.net/t/sync-stuck-at-95-0b/11150

The status shows no needed items and when you click the link, it stays at "loading data...". The problem is, that needed deleted files are not taken into account.

imsodin:
See #4668 

In addition I noticed that the completion sent in the summary service do not include the "needDeletes" parameter. This is not directly needed to fix the linked issue, but I still fixed it.

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -665,14 +665,17 @@ func (s *apiService) getDBCompletion(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	comp := s.model.Completion(device, folder)
-	sendJSON(w, map[string]interface{}{
+	sendJSON(w, jsonCompletion(s.model.Completion(device, folder)))
+}
+
+func jsonCompletion(comp model.FolderCompletion) map[string]interface{} {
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
Are deletes actually included in the summary?

imsodin:
It's not part of the folder summary. This information is available at the db/completion endpoint and through folder completion events. In the GUI in both cases the data is simply assigned to `$scope.completion[device][folder]`, so the data should be identical - which it wasn't before. Let me make that explicit by adding a separate function for that.

calmh:
@st-review lgtm

st-review:

AudriusButkevicius:
@st-review lgtm

st-review:
:ok_hand: Merged as f1c73999bee8903d0b71b200047f889dc07ef82c. Thanks, @imsodin!

PR #4671, opened 2018-01-14
lib/model: Microoptimization of unifySubs and blockDiff

imsodin:
Recently unnecessary allocs came up in some comments and I noticed some in the code. This "optimization" is cosmetic and the benchmarks are definitely over the top, but as I have already written them out, why not keep it. Also I personally consider the rewritten `unifySubs` simpler.

```
BenchmarkUnifySubs-4      200000              7947 ns/op            1008 B/op        36 allocs/op
BenchmarkUnifySubs-4      500000              3143 ns/op             256 B/op         8 allocs/op

BenchmarkDiff-4           300000              5334 ns/op            1920 B/op        27 allocs/op
BenchmarkDiff-4           300000              5013 ns/op            2592 B/op        24 allocs/op
```

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -2771,48 +2771,29 @@ func readOffsetIntoBuf(fs fs.Filesystem, file string, offset int64, buf []byte) 
 // The exists function is expected to return true for all known paths
 // (excluding "" and ".")
 func unifySubs(dirs []string, exists func(dir string) bool) []string {
-	subs := trimUntilParentKnown(dirs, exists)
-	sort.Strings(subs)
-	return simplifySortedPaths(subs)
-}
-
-func trimUntilParentKnown(dirs []string, exists func(dir string) bool) []string {
-	var subs []string
-	for _, sub := range dirs {
-		for sub != "" && !fs.IsInternal(sub) {
-			sub = filepath.Clean(sub)
-			parent := filepath.Dir(sub)
-			if parent == "." || exists(parent) {
-				break
-			}
-			sub = parent
-			if sub == "." || sub == string(filepath.Separator) {
-				// Shortcut. We are going to scan the full folder, so we can
-				// just return an empty list of subs at this point.
-				return nil
-			}
-		}
-		if sub == "" {
-			return nil
-		}
-		subs = append(subs, sub)
+	if len(dirs) == 0 {
+		return nil
 	}
-	return subs
-}
-
-func simplifySortedPaths(subs []string) []string {
-	var cleaned []string
-next:
-	for _, sub := range subs {
-		for _, existing := range cleaned {
-			if sub == existing || strings.HasPrefix(sub, existing+string(fs.PathSeparator)) {
-				continue next
-			}
-		}
-		cleaned = append(cleaned, sub)
+	sort.Strings(dirs)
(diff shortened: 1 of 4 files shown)


lib/model/model.go
```diff
@@ -2771,48 +2771,29 @@ func readOffsetIntoBuf(fs fs.Filesystem, file string, offset int64, buf []byte)
...
+	unified := dirs[:0]
+	prev := "./" // Anything that can't be parent of a clean path
+	for _, dir := range dirs {
+		dir = filepath.Clean(dir)
+		if dir == prev || strings.HasPrefix(dir, prev+string(fs.PathSeparator)) {
+			continue
 		}
-		cleaned = append(cleaned, sub)
+		parent := filepath.Dir(dir)
+		for !(exists(parent) || parent == "." || parent == string(fs.PathSeparator)) {
```
AudriusButkevicius: Invert this and drop the parenthesis.

AudriusButkevicius:
@st-review lgtm

st-review:

lib/model/model.go
```diff
@@ -2771,48 +2771,29 @@ func readOffsetIntoBuf(fs fs.Filesystem, file string, offset int64, buf []byte)
...
+	unified := dirs[:0]
+	prev := "./" // Anything that can't be parent of a clean path
+	for _, dir := range dirs {
+		dir = filepath.Clean(dir)
+		if dir == prev || strings.HasPrefix(dir, prev+string(fs.PathSeparator)) {
+			continue
 		}
-		cleaned = append(cleaned, sub)
+		parent := filepath.Dir(dir)
+		for !exists(parent) && parent != "." && parent != string(fs.PathSeparator) {
```
AudriusButkevicius: I'd move the exists check last, as that's the most expensive one.

lib/model/model.go
```diff
@@ -2771,48 +2771,29 @@ func readOffsetIntoBuf(fs fs.Filesystem, file string, offset int64, buf []byte)
...
+		if dir == prev || strings.HasPrefix(dir, prev+string(fs.PathSeparator)) {
+			continue
 		}
-		cleaned = append(cleaned, sub)
+		parent := filepath.Dir(dir)
+		for parent != "." && parent != string(fs.PathSeparator) && !exists(parent) {
+			dir = parent
+			parent = filepath.Dir(dir)
+		}
+		unified = append(unified, dir)
```
calmh: This is subtle and scary stuff as we're here modifying the backing array for dirs which we are also iterating over. It works because we are always looking at the same element we would be overwriting or one further along in the array - but it at the very least deserves a comment noting that this happens as it's liable to be a disaster if this gets refactored in some way. For example so that a given path gets split into two that both get appended or something.

imsodin: I see that now, I didn't even think of it as dangerous because using the same backing array was my initial thought for doing this. Instead of writing a comment, I rewrote the loop to explicitly modify the `dirs` slice given as argument. How does that look?

calmh:
@st-review lgtm

st-review:
:ok_hand: Merged as bd63fd73b149ca41122f901f9d9fa9655942b42c. Thanks, @imsodin!

PR #4672, opened 2018-01-15
lib/upgrade: Always return latest version, even if older than current (fixes #4654)

### LINKED ISSUE #4654: Upgrade system shows an error on RCs in some cases
```
jb@unu:~ $ syncthing 
[monitor] 22:43:40 INFO: Starting syncthing
[SYNO4] 22:43:40 INFO: syncthing v0.14.43-rc.1+8-gdb03562d "Dysprosium Dragonfly" (go1.9.2 darwin-amd64) jb@unu.kastelo.net 2018-01-05 18:11:09 UTC
...
[SYNO4] 22:43:42 INFO: Automatic upgrade is always enabled for candidate releases.
...
[SYNO4] 22:43:43 INFO: Automatic upgrade: couldn't find a release to download
```

That last error shouldn't happen.

screen shot 2018-01-07 at 22 46 00

That 500 error shouldn't happen either.

There is at the time of writing no newer version, which is what the upgrade check should return.

calmh:
The only special check remaining is the one to prefer a minor upgrade over a major one.

### Purpose

Fix the case where RCs error out because they are set not to upgrade to RCs and the only other available releases are older and filtered out.

### Testing

The current latest release is 0.14.43. I tested compiling this as 0.14.42, 0.14.43-rc.1, 0.14.43, 0.14.43-rc.1, 0.14.44 and verified that the upgrade happened for 0.14.42, 0.14.43-rc.1 and no error thrown (or erroneous downgrades) on the others.

### DIFF
--- a/lib/upgrade/upgrade_supported.go
+++ b/lib/upgrade/upgrade_supported.go
@@ -134,34 +134,25 @@ func SelectLatestRelease(rels []Release, current string, upgradeToPreReleases bo
 
 	var selected Release
 	for _, rel := range rels {
-		switch CompareVersions(rel.Tag, current) {
-		case Older, MajorOlder:
-			// This is older than what we're already running
-			continue
-
-		case MajorNewer:
+		if CompareVersions(rel.Tag, current) == MajorNewer {
 			// We've found a new major version. That's fine, but if we've
 			// already found a minor upgrade that is acceptable we should go
 			// with that one first and then revisit in the future.
 			if selected.Tag != "" && CompareVersions(selected.Tag, current) == Newer {
 				return selected, nil
 			}
-			// else it may be viable, do the needful below
-
-		default:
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as fcc6a677a54e143596efc31494edfcba5ac1eb1c. Thanks, @calmh!

imsodin:
Shouldn't we upgrade to new RCs when running a RC even if upgrades are set to stable releases? E.g. having upgraded to 0.14.42-rc1 and then switching back to stable upgrades only, it should still install 0.14.42-rcX until it reaches 0.14.42, after which 0.14.43-rcX are not installed.

calmh:
That might actually be the better behavior, yes.

PR #4673, opened 2018-01-15
cmd/syncthing, lib/sync: Don't do deadlock detection when STDEADLOCKTIMEOUT=0 (fixes #4644)

### LINKED ISSUE #4644: Impossible to run non-release builds without deadlock detectors
```
	if t := os.Getenv("STDEADLOCKTIMEOUT"); len(t) > 0 {
		it, err := strconv.Atoi(t)
		if err == nil {
			m.StartDeadlockDetector(time.Duration(it) * time.Second)
		}
	} else if !IsRelease || IsBeta {
		m.StartDeadlockDetector(20 * time.Minute)
	}
```
and
```
	useDeadlock = os.Getenv("STDEADLOCKTIMEOUT") != ""
)

func init() {
	l.SetDebug("sync", strings.Contains(os.Getenv("STTRACE"), "sync") || os.Getenv("STTRACE") == "all")

	if n, err := strconv.Atoi(os.Getenv("STLOCKTHRESHOLD")); err == nil {
		threshold = time.Duration(n) * time.Millisecond
	}
	if n, err := strconv.Atoi(os.Getenv("STDEADLOCKTIMEOUT")); err == nil {
		deadlock.Opts.DeadlockTimeout = time.Duration(n) * time.Second
	}
	l.Debugf("Enabling lock logging at %v threshold", threshold)
}
```

makes it impossible to run without deadlock detector, as if you set `STDEADLOCKTIMEOUT` to something that's not a number, you still end up with `useDeadlock=true`

calmh:
Allows setting STDEADLOCKTIMEOUT =0 (or any integer <= 0) to disable the deadlock detection entirely.

### Testing

I ran it with `STTRACE=sync STDEADLOCKTIMEOUT=<various>` and the output seemed like what I expected.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -770,10 +770,9 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 
 	m := model.NewModel(cfg, myID, "syncthing", Version, ldb, protectedFiles)
 
-	if t := os.Getenv("STDEADLOCKTIMEOUT"); len(t) > 0 {
-		it, err := strconv.Atoi(t)
-		if err == nil {
-			m.StartDeadlockDetector(time.Duration(it) * time.Second)
+	if t := os.Getenv("STDEADLOCKTIMEOUT"); t != "" {
+		if secs, _ := strconv.Atoi(t); secs > 0 {
+			m.StartDeadlockDetector(time.Duration(secs) * time.Second)
 		}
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 838c182b5b1665473d68fc0670a8030a2268b2e4. Thanks, @calmh!

PR #4674, opened 2018-01-15
cmd/syncthing: Release candidates always upgrade to newer release candidates (ref #4672)

calmh:
### Purpose

Have RCs auto upgrade to newer RCs regardless of preference. Once back on a real release the users preference is honored as usual.

### Testing

None, but it's just a logical-OR...

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -515,7 +515,12 @@ func debugFacilities() string {
 func checkUpgrade() upgrade.Release {
 	cfg, _ := loadOrDefaultConfig()
 	opts := cfg.Options()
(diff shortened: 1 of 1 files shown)


imsodin:
Possible problem: E.g. current release 43-rc.1, config set to stable updates only. 43 comes out, but Syncthing isn't running -> no upgrade. 44-rc.1 comes out and Syncthing is started up -> upgrade from 43-rc.1 to 44-rc.1 despite selecting stable only.

calmh:
Yeah that's undesirable. And fixing it is more invasive than I want at the moment. I think it's fine to wait for the next full release in case the user has decided that's what they want to do so I retract this.

PR #4675, opened 2018-01-15
cmd/syncthing, lib/config, lib/osutil: Lower process priority (fixes #4628)

### LINKED ISSUE #4628: Set background priority / niceness
Per #3212 it is desirable to set the process priority down by default. I think we should do that as well, with an off switch.

- New config option: `options.setLowPriority` which defaults to `true`
- When this option is true, at startup, we:
  - On Unixes, set our niceness to 9.
  - On Windows, set priority class PROCESS_MODE_BACKGROUND_BEGIN

It would be nice to also set I/O priority on Linux, but honestly the `ioprio_set` API is annoying to use from Go and I'm not going there at this point.

@canton7 any concerns here for you? You wanted to know what the outcome became, and you already set something by default?

We might want to take care to not *raise* our priority as part of this, though, by checking if the original priority is default or not. But on the other hand I think that on at least Unixes it's not possible for the process to raise its own priority so this comes automatically.

calmh:
### Purpose

Be nicer to the rest of the system by default.

Adds a new option `setLowPriority` which is true by default. If set to false we do nothing to change our priority on startup, like before.

### Testing

I've tested this on Mac, Linux and FreeBSD and see the desired result on all of them: all threads in the final syncthing process (not the monitor) become niceness level 9.

I have not inspected the result on Windows... I don't have a Windows VM on my laptop right now, I'd really appreciate if someone ran the resulting binary and checked if it did what it's supposed to and didn't, like, crash or something...

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -916,8 +916,9 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 	cleanConfigDirectory()
 
 	if cfg.Options().SetLowPriority {
-		// This is best effort and we ignore failures.
-		osutil.SetLowPriority()
+		if err := osutil.SetLowPriority(); err != nil {
+			l.Warnln("Lowering process priority:", err)
+		}
 	}
 
 	code := <-stop
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -917,7 +917,7 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 
 	if cfg.Options().SetLowPriority {
 		if err := osutil.SetLowPriority(); err != nil {
-			l.Warnln("Lowering process priority:", err)
+			l.Warnln("Failed to lower process priority:", err)
 		}
 	}
 
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -915,6 +915,11 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 
 	cleanConfigDirectory()
 
+	if cfg.Options().SetLowPriority {
+		// This is best effort and we ignore failures.
+		osutil.SetLowPriority()
+	}
+
 	code := <-stop
 
 	mainService.Stop()
--- a/lib/osutil/lowprio_linux.go
+++ b/lib/osutil/lowprio_linux.go
@@ -1,4 +1,4 @@
-// Copyright (C) 2017 The Syncthing Authors.
+// Copyright (C) 2018 The Syncthing Authors.
 //
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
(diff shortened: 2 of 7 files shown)


lib/osutil/lowprio_linux.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// SetLowPriority lowers the process CPU scheduling priority, and possibly
+// I/O priority depending on the platform and OS.
+func SetLowPriority() {
+	// Move ourselves to a new process group so that we can use the process
+	// group variants of Setpriority etc to affect all of our threads in one
+	// go. If this fails, bail, so that we don't affect things we shouldn't.
+	if err := syscall.Setpgid(0, 0); err != nil {
+		return
```
AudriusButkevicius: sync.Once this?

cmd/syncthing/main.go
```diff
@@ -915,6 +915,11 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 
 	cleanConfigDirectory()
 
+	if cfg.Options().SetLowPriority {
+		// This is best effort and we ignore failures.
+		osutil.SetLowPriority()
```
AudriusButkevicius: We could atleast log failures.

AudriusButkevicius:
Would it make sense to make this like a value you can adjust? Default being like 3, so people can go up and down? Perhaps someone is running syncthing as a service.

cmd/syncthing/main.go
calmh: Agreed

lib/osutil/lowprio_linux.go
calmh: I don't see the point, it's just called once from main and we'd need to remember the error in another package variable to be able to return it again etc

calmh:
I'd really not want to add too many knobs to this because then we'll end up having to expose all the scheduler classes for iopriority, the various names of the priority classes in windows or the bitmasks, etc. I think we should just do something sane by default, and if someone wants something better they can disable our stuff by config and do whatever they please at launch time.

lib/osutil/lowprio_linux.go
AudriusButkevicius: Yeah, I was just pointing out in case other Go users use it. Also, with errors being returned, sync.Once makes more sense (as I assume doing pgrp twice will fail).

canton7:
Worth logging failures?

calmh:
Logging failures added... As warning level no less. Is that too aggressive? Maybe have it like that for the RC and see what drops in?

lib/osutil/lowprio_windows.go
```diff
@@ -0,0 +1,43 @@
...
+// SetLowPriority lowers the process CPU scheduling priority, and possibly
+// I/O priority depending on the platform and OS.
+func SetLowPriority() error {
+	handle, err := syscall.GetCurrentProcess()
+	if err != nil {
+		return errors.Wrap(err, "get process handler")
+	}
+	defer syscall.CloseHandle(handle)
+
+	_, _, err := syscall.Syscall(uintptr(setPriorityClass), uintptr(handle), belowNormalPriorityClass, 0, 0)
```
AudriusButkevicius: Why is the ioprio return value handling different from windows?

canton7:
(A note on copyright: in my understanding, the date in the copyright is the *earliest* year at which you assert copyright over the work, not the latest year. It says "I wrote this code in 2017, so anything which appears since then and isn't written by me is a derivative work". If you update the copyright year to 2018, you're more or less relinquishing your claim to have written the work in 2017, so if another copy pops up copyrighted 2017, your claim to hold copyright looks a little shakey...).

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as c554ffccc9c8fd54c34dc1b1ff01226fed79c31e. Thanks, @calmh!

lib/osutil/lowprio_windows.go
calmh: It's not, any more, I think

PR #4677, opened 2018-01-17
lib/osutil: Disable setting priority on Windows (fixes #4676)

### LINKED ISSUE #4676: Panic: fatal error: unexpected signal during runtime execution
I have 3x computers synced using syncthing:
Windows 7 32bit - work, installed manually
Windows 10 64bit - home, installed via just-install
Void Linux 64bit - server, run via docker linuxserver/syncthing

On startup, work computer produces the attached panic log. This panic happens every start I start syncthing and I felt reporting the issue was better than re-installing the application in the hopes it would resolve but I'm willing to try.

Yesterday I added a number of sqlite databases (baikal, selfoss, etc) from the 'server' install, but my Windows 10 install gives me no issues.

Attached a redacted panic log.

### Version Information

Syncthing Version: v0.14.44-rc.1
OS Version: Windows 7 32bit

[panic-20180117-091638.log](https://github.com/syncthing/syncthing/files/1638379/panic-20180117-091638.log)

calmh:
Presumably fixing the crash, leaving us to improve this calmly.

### DIFF
--- a/lib/osutil/lowprio_windows.go
+++ b/lib/osutil/lowprio_windows.go
@@ -6,42 +6,8 @@
 
 package osutil
 
-import (
-	"syscall"
-
-	"github.com/pkg/errors"
-)
-
-var (
-	kernel32, _         = syscall.LoadLibrary("kernel32.dll")
-	setPriorityClass, _ = syscall.GetProcAddress(kernel32, "SetPriorityClass")
-)
-
-const (
-	// https://msdn.microsoft.com/en-us/library/windows/desktop/ms686219(v=vs.85).aspx
-	aboveNormalPriorityClass   = 0x00008000
-	belowNormalPriorityClass   = 0x00004000
-	highPriorityClass          = 0x00000080
-	idlePriorityClass          = 0x00000040
-	normalPriorityClass        = 0x00000020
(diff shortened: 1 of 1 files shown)


calmh:
@st-review merge

st-review:
:ok_hand: Merged as 582539a1e60ac325cea1cc801552f9a4390e4e13. Thanks, @calmh!

AudriusButkevicius:
I guess this is so urgent as you want a new RC?

calmh:
Yeah printing one currently

calmh:
Can try to get a fixed version into an rc3, but lets not crash in the meantime

AudriusButkevicius:
I hopefully can test this, yet not until friday I suspect.

calmh:
I can see a couple of things that might go wrong. I'm blindly assuming that kernel32.dll is available (and we do that in other places) and that SetPriorityClass is found. That may not be the case. The process handle in the crash backtrace is 0xffffffff which is odd, yet that is one call where we do check the error. And of course I might have misunderstood something else.

AudriusButkevicius:
I suspect it crashes due to an exception/fault/whatever it's called raised during the syscall, and the exception is raised because something is not found or not valid.

PR #4682, opened 2018-01-17
lib/osutil: Don't attempt to reduce our niceness level (fixes #4681)

### LINKED ISSUE #4681: "Failed to lower process priority: set niceness: permission denied" when already niced
```
$ nice -15 ./syncthing -home test
...
[4DVPI] 13:45:31 WARNING: Failed to lower process priority: set niceness: permission denied
```

calmh:
### Purpose

Avoid complaining to the user about permission denied when this happens.

### DIFF
--- a/lib/osutil/lowprio_linux.go
+++ b/lib/osutil/lowprio_linux.go
@@ -42,30 +42,33 @@ func ioprioSet(class ioprioClass, value int) error {
 // SetLowPriority lowers the process CPU scheduling priority, and possibly
 // I/O priority depending on the platform and OS.
 func SetLowPriority() error {
-	const wantNiceLevel = 9
+	// Process zero is "self", niceness value 9 is something between 0
+	// (default) and 19 (worst priority).
+	const (
+		pidSelf       = 0
+		wantNiceLevel = 9
+	)
 
 	// Move ourselves to a new process group so that we can use the process
 	// group variants of Setpriority etc to affect all of our threads in one
 	// go. If this fails, bail, so that we don't affect things we shouldn't.
 	// If we are already the leader of our own process group, do nothing.
-	if pgid, err := syscall.Getpgid(0); err != nil {
+	if pgid, err := syscall.Getpgid(pidSelf); err != nil {
 		// This error really shouldn't happen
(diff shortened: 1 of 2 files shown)


calmh:
@st-review don't

st-review:

lib/osutil/lowprio_linux.go
```diff
@@ -42,23 +42,33 @@ func ioprioSet(class ioprioClass, value int) error {
...
 // I/O priority depending on the platform and OS.
 func SetLowPriority() error {
+	// Process zero is "self", niceness value 9 is something between 0
+	// (default) and 19 (worst priority).
+	const (
+		pidSelf       = 0
+		wantNiceLevel = 9
+	)
+
+	if cur, _ := syscall.Getpriority(syscall.PRIO_PROCESS, pidSelf); cur >= wantNiceLevel {
```
AudriusButkevicius: Why are we not handling the error?

calmh: I'm being lazy and assuming that if there is an error I'm not to get a positive integer back

calmh:
Gnnh

```
   C library/kernel differences
       Within  the  kernel,  nice  values  are actually represented using the range 40..1 (since negative numbers are
       error codes) and these are the values employed by the setpriority() and getpriority() system calls.  The glibc
       wrapper  functions for these system calls handle the translations between the user-land and kernel representa‐
       tions of the nice value according to the formula unice = 20 - knice.  (Thus, the kernel's 40..1  range  corre‐
       sponds to the range -20..19 as seen by user space.)
```

lib/osutil/lowprio_linux.go
imsodin: Wouldn't it make sense to abort if the niceness is anything but the default value? Line in if the user sets niceness manually, it's nicer (haha) to not mess with it at all.

calmh:
I think this works properly now

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 3af4164c8bd89c4f9d881c664e9bcb4b4586e875. Thanks, @calmh!

PR #4685, opened 2018-01-18
lib/db: Do not modify underlying array of argument

imsodin:
Previously the fileinfo list was filtered in place, now a copy is used. This list is used in other functions than this, so it should not be modified. I noticed this problem while writing and debugging a new test, when repeat fileinfo entries showd in the index updates. I cannot reproduce this anymore and I do not know whether it can occur in practice, but modifying in place is anyway the wrong thing to do in this case.

### DIFF
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -126,7 +126,7 @@ func (s *FileSet) Update(device protocol.DeviceID, fs []protocol.FileInfo) {
 	l.Debugf("%s Update(%v, [%d])", s.folder, device, len(fs))
 
 	// do not modify fs in place, it is still used in outer scope
-	fs = append([]protocol.FileInfo{}, fs...)
+	fs = append([]protocol.FileInfo(nil), fs...)
 
 	normalizeFilenames(fs)
 
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -124,6 +124,10 @@ func (s *FileSet) Drop(device protocol.DeviceID) {
 
 func (s *FileSet) Update(device protocol.DeviceID, fs []protocol.FileInfo) {
 	l.Debugf("%s Update(%v, [%d])", s.folder, device, len(fs))
+
+	// do not modify fs in place, it is still used in outer scope
+	fs = append([]protocol.FileInfo{}, fs...)
+
 	normalizeFilenames(fs)
 
 	s.updateMutex.Lock()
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
@st-review merge

st-review:

imsodin:
@AudriusButkevicius The test that failed was relying on the underlying fs being modified in place, as it omitted the sequence numbers in the expected results. Also the normalization should not happen in place either, so I moved copying the input array to the beginning of the function.

lib/db/set.go
```diff
@@ -124,6 +124,10 @@ func (s *FileSet) Drop(device protocol.DeviceID) {
 
 func (s *FileSet) Update(device protocol.DeviceID, fs []protocol.FileInfo) {
 	l.Debugf("%s Update(%v, [%d])", s.folder, device, len(fs))
+
+	// do not modify fs in place, it is still used in outer scope
+	fs = append([]protocol.FileInfo{}, fs...)
```
AudriusButkevicius: This can just be fs := fs no?

imsodin: I just tested it (with  something like `newFs := fs` due to "no new variable") and that still uses the same backing array. I assumed this was the case before due to this [official wiki entry on slice copying](https://github.com/golang/go/wiki/SliceTricks#copy).

AudriusButkevicius: I think you can just do `append(nil, fs...)` in that case.

imsodin: Go apparently still wants a typed nil: `append([]protocol.FileInfo(nil), fs...)`

lib/db/set.go
```diff
@@ -124,6 +124,10 @@ func (s *FileSet) Drop(device protocol.DeviceID) {
 
 func (s *FileSet) Update(device protocol.DeviceID, fs []protocol.FileInfo) {
 	l.Debugf("%s Update(%v, [%d])", s.folder, device, len(fs))
+
+	// do not modify fs in place, it is still used in outer scope
+	fs = append([]protocol.FileInfo(nil), fs...)
```
AudriusButkevicius: And why doesn't an untyped nil not work?

AudriusButkevicius: I guess this works with pointer slices.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as fae2ca84585b3bdea21409943e69bdea70a9adc9. Thanks, @imsodin!

PR #4692, opened 2018-01-20
 lib/db: Optimize memory usage for huge files, refactor updating and fix benchs

imsodin:
This started in a discussion about memory use spiking when scanning is finished here: https://forum.syncthing.net/t/out-of-memory-crash-while-scanning/11170/13
The I thought I saw unnecessary block copies abound, but it turned out I was wrong. However in the process I saw other potential memory optimizations and some refactor potential:

 - The function `(m *BlockMap) Add` is only used in test and there it can be replaced by `Update` -> remove
 - Do not store existing files to be discarded in a list -> move discarding into `(m *BlockMap) Update`.  
This might also be achievable while keeping `Discard` separate, but I quite like having a single `Update` function as they are never used alone and it's the equivalent function to `Update` on `FileSet`.
 - Check whether the block map batch exceeds the maximum size after every block instead of after every file (i.e. for one large file, the transaction will be split into several batches).
 - Discard block data early when removing an old file from the block map.

### Testing

I fixed the previously broken benchmarks (unnoticed due to build tags). There is no consistent change visible between master and this PR - so this may save some memory and should not impact performance (much).

### DIFF
--- a/lib/db/benchmark_test.go
+++ b/lib/db/benchmark_test.go
@@ -16,17 +16,18 @@ import (
 	"testing"
 
 	"github.com/syncthing/syncthing/lib/db"
+	"github.com/syncthing/syncthing/lib/fs"
 	"github.com/syncthing/syncthing/lib/protocol"
 )
 
 var files, oneFile, firstHalf, secondHalf []protocol.FileInfo
-var fs *db.FileSet
+var s *db.FileSet
 
 func init() {
 	for i := 0; i < 1000; i++ {
 		files = append(files, protocol.FileInfo{
 			Name:    fmt.Sprintf("file%d", i),
-			Version: protocol.Vector{{ID: myID, Value: 1000}},
+			Version: protocol.Vector{[]protocol.Counter{{ID: myID, Value: 1000}}},
 			Blocks:  genBlocks(i),
 		})
 	}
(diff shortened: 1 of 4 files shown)


calmh:
How about a memory benchmark? (Using `go test -bench . -benchmem` or the t.ReportAllocs() or whatever the method is called, I forget.)

imsodin:
I increase benchtime to get better results. Allocs/op seem to correlate with benchtime and vary slightly by run (shouldn't they be deterministic?). If there is a trend, then weirdly enough it looks like overall the PR uses slightly more allocs/op.
 
before:
```
BenchmarkReplaceAll-4                         10	2572511544 ns/op	253095833 B/op	  606398 allocs/op
BenchmarkUpdateOneChanged-4     	  300000	    120480 ns/op	   87763 B/op	     548 allocs/op
BenchmarkUpdateOneUnchanged-4   	  200000	    119976 ns/op	   87698 B/op	     549 allocs/op
BenchmarkUpdateOneFile-4        	  200000	    215976 ns/op	   34519 B/op	     232 allocs/op
```

after:
```
BenchmarkReplaceAll-4           	      20	2405571755 ns/op	268937934 B/op	  636290 allocs/op
BenchmarkUpdateOneChanged-4     	  200000	    142700 ns/op	  108597 B/op	     558 allocs/op
BenchmarkUpdateOneUnchanged-4   	  200000	    131637 ns/op	  108556 B/op	     559 allocs/op
BenchmarkUpdateOneFile-4        	  200000	    232549 ns/op	   36151 B/op	     267 allocs/op
```

I must admit I didn't look at what the benchmarks do besides what was needed to get the to run. Probably there is none for the single big file scenario.

calmh:
So not to be annoying, but the benchmark changes (to the extent they're statistically relevant) all became worse?

imsodin:
Not annoying at all, I kind of purposely forgot about this PR :P  
I wouldn't call the bench changes statistically relevant at all - most of them are clearly not relevant and in those which may be, I observed weird dependencies on runtime parameters. I didn't see enough potential gain to go through the work of either writing "better" benchmarks (whatever that would be) or do a actual statistics on benchmark results.

So I close this and will do a new PR with the benchmarks fixes and maybe at some point I'll come back on some of the changes here.

PR #4698, opened 2018-01-24
lib/config: Reject empty folder IDs

calmh:
### Purpose

Empty folder IDs aren't cromulent and will cause a crash on startup. With this change we also reject such folder IDs when deserializing the config, meaning we will not accept it if it comes in via the API for example.

### Testing

Unit.

### DIFF
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -872,7 +872,7 @@ func TestIssue4219(t *testing.T) {
 
 func TestInvalidDeviceIDRejected(t *testing.T) {
 	// This test verifies that we properly reject invalid device IDs when
-	// deserializing a JSON config from the API.
+	// deserializing a JSON config.
 
 	cases := []struct {
 		id string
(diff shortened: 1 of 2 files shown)


lib/config/config_test.go
```diff
@@ -916,12 +916,54 @@ func TestInvalidDeviceIDRejected(t *testing.T) {
...
+		devs[0] = dev0
+
+		invalidJSON, err := json.Marshal(cfg)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		_, err = ReadJSON(bytes.NewReader(invalidJSON), device1)
+		if tc.ok && err != nil {
+			if err != nil {
```
imsodin: Redundant inner if.

imsodin:
@st-review lgtm

st-review:

calmh:
@st-review merge

st-review:
:ok_hand: Merged as e0931e201e35e71729768062d086849cf8d87090. Thanks, @calmh!

PR #4699, opened 2018-01-25
Missing datalist for parameter DefaultFolderPath

Gregger83:
Add missing datalist for parameter DefaultFolderPath in SettingsForm

### Purpose

When setting the folder path on the "editFolder"-Dialog there is a list of subdirs below the current directory.

When setting the DefaultFolderPath in the settings dialog, this list of subdirs is missing.

### DIFF
--- a/gui/default/syncthing/settings/settingsModalView.html
+++ b/gui/default/syncthing/settings/settingsModalView.html
@@ -82,6 +82,9 @@
           <div class="form-group">
             <label translate for="urVersion">Default Folder Path</label>
             <input id="DefaultFolderPath" class="form-control" type="text" ng-model="tmpOptions.defaultFolderPath"/>
+            <datalist id="directory-list">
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
This doesn't do what you think it should, as this is not actually fetching the available subdirs.

AudriusButkevicius:
directoryList gets populated when currentFolder.path changes.

```
        $scope.$watch('currentFolder.path', function (newvalue) {
            if (!newvalue) {
                return;
            }
            $scope.currentFolder.path = expandTilde(newvalue);
            $http.get(urlbase + '/system/browse', {
                params: { current: newvalue }
            }).success(function (data) {
                $scope.directoryList = data;
        });
```

This is not changing currentFolder.path.

I suspect this should start working if you also add a watch for `tmpOptions.defaultFolderPath`

imsodin:
Timing out...

PR #4702, opened 2018-01-25
lib/db: Update global counts on invalidation (fixes #4701)

### LINKED ISSUE #4701: Global is different from local state when ignoring files
Syncthing :

add TestDir,
save (No need to share or any special option)

Open file manager into the directory TestDir
create :
- NewDir
- file.txt
- file.pdf
copy file.* into NewDir

Syncthing :
Open TestDir,
check display to see global = local state : 4 files 1 folder

Edit share
Click button Ignore Patterns
enter :

```
(?i)!*.pdf
/*

```
Save,
save.

Check global state : 4/1; local 2/0 Vs reality 4/1

Open file manager into the directory TestDir/NewDir
delete file.txt

Syncthing :
rescan
Check global state : 4/1; local 2/0 Vs reality 3/1

Quite confusing as there is no share with other device already enabled; global should equal local and/or reality. At least, global shouldn’t be greater than reality. 
Thanks ! 
BD

### Version Information
Syncthing Version: v0.14.43, Linux (ARM)
OS Version: Linux Qnap

imsodin:
### Purpose

Fix #4701

This only fixes calculations in the future. So any existing disparities will stay until the metadata is next recomputed. @calmh Do you think it is worth/necessary to add a config transition to trigger a metadata recalculation or is it fine to just wait until one happens "naturally"?

### Testing

New unit test.

### DIFF
--- a/lib/db/leveldb_transactions.go
+++ b/lib/db/leveldb_transactions.go
@@ -168,7 +168,7 @@ insert:
 	if insertedAt == 0 {
 		// We just inserted a new newest version. Fixup the global size
 		// calculation.
-		if !file.Version.Equal(oldFile.Version) {
+		if !file.Version.Equal(oldFile.Version) || file.Invalid != oldFile.Invalid {
 			meta.addFile(globalDeviceID, file)
 			if hasOldFile {
 				// We have the old file that was removed at the head of the list.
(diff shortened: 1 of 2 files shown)


calmh:
@st-review merge

st-review:
:ok_hand: Merged as 364f61bda61dba63885a0e8b1c796e20f48c473a. Thanks, @imsodin!

PR #4704, opened 2018-01-25
gui: Display rescan button when out of sync and remove deprecated folder state

imsodin:
It should be possible to trigger a scan on an out of sync folder, i.e. in case a problem was resolved manually on disk. I reversed the condition from displaying rescan for some specific folder states to all states except unknown and paused.

The invalid parameter on the folder summary is deprecated as per comment in gui.go -> removed.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -439,7 +439,7 @@
                   <button type="button" class="btn btn-default btn-sm" ng-click="restoreVersions.show(folder.id)" ng-if="folder.versioning.type">
                     <span class="fa fa-undo"></span>&nbsp;<span translate>Versions</span>
                   </button>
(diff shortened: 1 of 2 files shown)


calmh:
> all states except unknown and paused

Do we really want to allow rescan while syncing? That won't actually *work*, will it?

imsodin:
It will schedule a blocking scan, i.e. the scan will happen after syncing finished - but that wasn't a conscious choice on my part, simply a mistake. State "scanning" was allowed too, which we don't want either I would say. I went back to explicitly allowing states, which is a bit less prone to such stupid mistakes.

AudriusButkevicius:
Lgtm

calmh:
@st-review merge

st-review:
:ok_hand: Merged as ae760798e1d53c3a89dfb52dceccd534c475a397. Thanks, @imsodin!

PR #4707, opened 2018-01-27
all: Implement variable block size

calmh:
### Purpose

More efficient on large files. That point works as expected, handling a large file (I tested with 64 GB) now works with a handful megs of ram instead of hundreds.

### Testing

Passes the usual tests plus integration, haven't run it in production yet.


AudriusButkevicius:
So I would just cut a 0.15, and don't care about compatibility.
If all nodes agree how the block size is calculated, then this doesn't need to be stored anywhere and sent as part of the fileinfos, which I think is a nice win.

calmh:
Just a side not so we don't forget, since db1dc998 (in prep for 1.0) 0.15 would no longer be considered a major so that would need to be wound back first.

calmh:
Side note two, the havoc might not be as big as I thought. Older versions actually ask for the block with size and offset as announced, and write it as such, so should be able to swallow index entries with larger block size. We "just" kill the block diff mechanism between old and new versions. Needs testing, clearly.

imsodin:
What about adding a little hysteresis to changing the block size to avoid changing it too often. Without it hypothetical tiny shifts up and down in size can cause a full retransmit every time. I am not sure whether this scenario would happen at all in the wild, but adding a hysteresis to "prevent" it regardless seems trivial. I.e. on creation of a file, use the optimal block size, but when a file already exists, only change the block size once a threshold is reached (e.g. 2^2 or two steps).

calmh:
This was *almost* compatible with the current code. Things work just fine except on one specific but important case: when a device without this change has synced a file with large blocks, and then is going to sync another change to the same file. This goes through the copier which has a buffer that is statically 128 KiB-sized and it panics.

What I actually propose to do is this:

- Add this under a config flag `UseLargeBlocks` (per folder), which defaults to off. Users can turn this on, with the understanding (documented somewhere) that it'll panic 0.14.44 and earlier devices.

- At some point in the future (when the user base on 0.14.44 is <5% or something), change the flag default value to true and migrate all existing configs to true.

- At some further distant point in the future (when the user base on 0.14.44 is <1% or something), remove the flag entirely making this the only behavior.

I'm still thinking about the hysteresis point, haven't convinced myself one way or the other yet. I suspect it'll add complexity for a rather unusual situation.

AudriusButkevicius:
I say we cut 0.15, and get on with our life

calmh:
One note on this. While it has the desired effect on memory utilization while scanning and uses less RAM for block lists and stuff while syncing... it also has the opposite effect while syncing. With the default setting of 64 puller routines and a large file we will allocate 64 * 16 MiB = 1 GiB of block buffers while pulling, plus GC overhead and other related stuff probably, on (roughly) both sides.

This is user tweakable as the number of pullers can be reduced. But maybe we'd like to be smarter about it to begin with as I'm sure there is no performance advantage of firing requests for a full GiB of data at the time. Maybe just reducing the default number of pullers to 2 or so makes sense. The point of many pullers is to keep lots of data in flight for large files, which the block size would now do instead. For small files we're limited on the copier routines instead...

AudriusButkevicius:
I think we should use pullers as a multiplier, and do it based on the blocksize, so we do (8mb/block size)*pullersMultiplier

calmh:
But the number of pullers is set at the start of a puller iteration, and the block size is per file.

AudriusButkevicius:
Well that obviously needs to be reworked. Essentially have an infinite pool of pullers, yet use no mire than 8mb worth of them per file * multiplier (float?). Also, this ties nicely into copier as to how much memory it's using.

calmh:
I’m not sure that needs to go into this same PR, for complexity reasons, but yeah. I could imagine not having fixed pullers, but having the copier launch short lived pullers as one goroutine per block. It could use a semaphore / rate limiter thing to launch enough pullers to have x MB outstanding requests at any given time.

lib/scanner/blocks.go
```diff
@@ -16,6 +16,24 @@ import (
...
+var blockSizes = []int{
+	128 << KiB, // up to 250 MiB
+	256 << KiB, // up to 500 MiB
+	512 << KiB, // up to 1000 MiB
+	1 << MiB,   // up to 2000 MiB
+	2 << MiB,   // up to 4000 MiB
+	4 << MiB,   // up to 8000 MiB
+	8 << MiB,   // up to 16000 MiB
+	16 << MiB,  // to infinity and beyond
+}
```
imsodin: A list of blocksizes is defined both here and in protocol without reference to the min and max blocksizes defined in protocol. In principle this list choice is arbitrary (e.g. rounding at 1MiB), so I think we should only be doing it in one place and refer to it in other places. And maybe to be super cautious, generate the min/max-blocksizes from this list.

imsodin:
> I'm still thinking about the hysteresis point, haven't convinced myself one way or the other yet. I suspect it'll add complexity for a rather unusual situation.

I don't think it will add much complexity, something like the following should be all it takes. And preventing a full rescan of a big file until it is clear that it is really necessary seems worth it to me, even if is a unlikely scenario.
```diff
--- a/lib/scanner/blocks.go
+++ b/lib/scanner/blocks.go
@@ -125,8 +125,18 @@ func Blocks(ctx context.Context, r io.Reader, blocksize int, sizehint int64, cou
 	return blocks, nil
 }
 
-func BlockSize(fileSize int64) int {
+func BlockSize(fileSize int64, oldBlockSize int) int {
 	const DesiredBlocks = 2000
+	// Adjust block size only if new block size differs by two steps
+	if ratio := oldBlockSize * DesiredBlocks / fileSize; ratio == 0 {
+		// new filesize is bigger
+		if fileSize/(oldBlockSize*DesiredBlocks) < 2 {
+			return oldBlocksize
+		}
+	} else if ratio < 4 {
+		// Different limit as the unchanged interval is [limit/2:limit]
+		return oldBlocksize
+	}
 
 	blockSize := 0
 	for _, blockSize = range blockSizes {
```

calmh:
I'm convinced, I'll add some hysteresis to it.

calmh:
This is pretty much good to go from my point of view. It needs some squashing and and tying to the corresponding issue, and I'm going to write an article on the usage and pitfalls of the advanced config that enables it... Then in half a year or so make it the default I think.

lib/scanner/walk.go
```diff
@@ -285,7 +285,25 @@ func (w *walker) walkRegular(ctx context.Context, relPath string, info fs.FileIn
 		curMode |= 0111
 	}
 
+	blockSize := protocol.MinBlockSize
+	if w.UseLargeBlocks {
+		blockSize = protocol.BlockSize(info.Size())
+	}
+
 	cf, ok := w.CurrentFiler.CurrentFile(relPath)
+	if ok {
```
imsodin: Also requires `w.UseLargeBlocks`.

AudriusButkevicius:
So how does this work. Someone announces blocks of larger size, old clients request smaller sizes, hashes mismatch? Feels like a breaking change regardless?

Also, why do we need to send the block size around? Why can't everyone just compute it based on size?

calmh:
Old (as well as new) clients all deal with reality in terms of blocks. The only places where the block size was hard coded was when scanning and the buffers used for copying / pulling. So old clients accept info about large blocks, pull large blocks, copy large blocks, and everything works out mostly fine. The caveat is that if a change happens on an old client they will rescan with small blocks and other devices that have large blocks won’t recognize the hashes for reuse and instead download the whole thing.

Apart from the panic that I fixed and that is now in .46, backwards compatibility is 100% except for unnecessary full downloads happening on block size mismatch.

We need the block size around because it might differ from the otherwise computed size. With the hysteresis added we don’t immediately change the block size when we go 1 KiB over the limit, for example. Also, files present since before large blocks existed.

AudriusButkevicius:
Can't we compute the block size by fileSize/numberOfBlocks? Not sure why we need it in the fileInfo.

imsodin:
How would that work in terms of protocol specs?

Btw build failures are "silent" merge conflicts with master.

calmh:
We need it internally in the because we have the truncated version without blocks. At the protocol level it may not be necessary, but the same could be said for the file size which is also derivable from the block list. I think it’s nice to have.

calmh:
I'm on a plane pretty much all of tomorrow and will try to use that time to write some docs and spec updates.

calmh:
I think this is good to go.

AudriusButkevicius:
Doesn't blockMap some handling of this to indicate block size?

AudriusButkevicius:
Also, checking temp file before pull, old file before pull, for block reuse, how is block difference handled there?

imsodin:
@AudriusButkevicius 

> Doesn't blockMap some handling of this to indicate block size?

There shouldn't be an issue as blocks with differing size won't have the same hash, so any match will have the same size as the requested one.

> Also, checking temp file before pull, old file before pull, for block reuse, how is block difference handled there?

I think there is nothing to handle, as both actually compute hashes with the desired (same) block size.

AudriusButkevicius:
> There shouldn't be an issue as blocks with differing size won't have the same hash, so any match will have the same size as the requested one.

There can be in theory false positive lookups of matching hashes yet different block sizes

calmh:
We already deal with 128000 different block sizes today. If you guys are now going to argue that sha256 isn’t collision resistant then we can all stop doing this and go home because then nothing works.

AudriusButkevicius:
Whatever, let's see the world burn, I am pretty sure we overlooked something.

calmh:
Oh I'm sure it's going to break spectacularly somehow. :) But it shouldn't be due to hash collisions

calmh:
aaaah it's merged! 😰

AudriusButkevicius:
It will be because we forgot to pass matching block size somewhere, when comparing something.

PR #4708, opened 2018-01-27
lib/db: Fix benchmarks

imsodin:
They didn't run before, due to being out-of-date compared to the rest of the project (they're hidden behind a build flag, so they don't surface often I guess).


calmh:
I don't see the need for that build tag either, whatever the original reason might have been. Lets get rid of it so this doesn't happen again. Otherwise lgtm of course

calmh:
@st-review rebuild

calmh:
@st-review merge

st-review:
:ok_hand: Merged as 14a5561e439edcdb56e3eb8b63a676ed05ea6c63. Thanks, @imsodin!

PR #4712, opened 2018-01-27
lib/fs: Introduce walkfs debug facility

imsodin:
This adds a new debug facility "walkfs" which does the same thing as "fs", but it also applies to operations when walking a filesystem. I added this unconditionally on another PR for debugging, but it doesn't really belong there, so lets discuss/decide here (original comments: https://github.com/syncthing/syncthing/pull/4584/files#r156240587).

### DIFF
--- a/lib/fs/debug.go
+++ b/lib/fs/debug.go
@@ -18,5 +18,11 @@ var (
 )
 
 func init() {
-	l.SetDebug("fs", strings.Contains(os.Getenv("STTRACE"), "fs") || os.Getenv("STTRACE") == "all")
+	if strings.Contains(os.Getenv("STTRACE"), "walkfs") || os.Getenv("STTRACE") == "all" {
+		l.SetDebug("fs", true)
+		logger.DefaultLogger.NewFacility("walkfs", "Filesystem access while walking")
+		l.SetDebug("walkfs", true)
+		return
+	}
+	l.SetDebug("fs", strings.Contains(os.Getenv("STTRACE"), "fs"))
 }
--- a/lib/fs/debug.go
+++ b/lib/fs/debug.go
@@ -19,10 +19,11 @@ var (
 
 func init() {
 	logger.DefaultLogger.NewFacility("walkfs", "Filesystem access while walking")
-	if strings.Contains(os.Getenv("STTRACE"), "walkfs") || os.Getenv("STTRACE") == "all" {
-		l.SetDebug("fs", true)
+	switch {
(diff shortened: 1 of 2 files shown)


lib/fs/debug.go
```diff
@@ -18,5 +18,11 @@ var (
 )
 
 func init() {
-	l.SetDebug("fs", strings.Contains(os.Getenv("STTRACE"), "fs") || os.Getenv("STTRACE") == "all")
+	if strings.Contains(os.Getenv("STTRACE"), "walkfs") || os.Getenv("STTRACE") == "all" {
+		l.SetDebug("fs", true)
+		logger.DefaultLogger.NewFacility("walkfs", "Filesystem access while walking")
```
calmh: Why not just make this a regular facility thing and not conditional on it being set to begin with?

lib/fs/debug.go
```diff
@@ -18,5 +18,11 @@ var (
...
 
 func init() {
-	l.SetDebug("fs", strings.Contains(os.Getenv("STTRACE"), "fs") || os.Getenv("STTRACE") == "all")
+	logger.DefaultLogger.NewFacility("walkfs", "Filesystem access while walking")
+	if strings.Contains(os.Getenv("STTRACE"), "walkfs") || os.Getenv("STTRACE") == "all" {
+		l.SetDebug("fs", true)
+		l.SetDebug("walkfs", true)
+		return
+	}
+	l.SetDebug("fs", strings.Contains(os.Getenv("STTRACE"), "fs"))
```
calmh: Sorry to be annoying, but this is still too much and removes the `all` variant from `fs`. How about:

```go
sttrace := os.Getenv("STTRACE")
l.SetDebug("fs", strings.Contains(sttrace, "fs") || sttrace == "all") // also matches walkfs
l.SetDebug("walkfs", strings.Contains(sttrace, "walkfs") || sttrace == "all")
```

imsodin: How so? For "all" both "fs" and "walkfs" are set. For "walkfs" additionally "fs" is set, because only setting "walkfs" does not have any effect.

calmh: In your current code, in the `all` case, you set both to true in the `if`. Then you set `fs` to false after the `if` because the all check is missing there. Not repeating it is cleaner.

imsodin: Honestly, I am (still) a bit confused, as I return from the body of the first `if`, so it implicitly behind a check. I changed it to a `switch`, is this cleaner?

Issue #4552, opened 2017-11-26
UI for filesystem watcher


calmh:
Now that we have the watcher code in place, let’s add the UI. There’s a couple of different parts to that:

- enabling it to begin with
- showing its current status
- showing a reasonable error when it doesn’t work

The enabling part is fairly simple. I think we’ve already mostly agreed on a checkbox, and when ticked it will also automatically increase the rescan interval to something high. This happens in the config UI, so the user can change the interval if they want, and in the code it’s still the same interval mechanism as usual. When the watcher is disabled we suggest the usual default interval of 60 seconds. One thing to note here might be a documentation link or comment about why rescans may still be necessary. As far as I know, permission only changes are not picked up by the watcher? Mtime only changes? May need some investigation.

The folder status thing on the left should have a row for the fs watcher status. For most things we don’t show a row unless the value diverges from the default. Nonetheless I think the fs watcher row should always be present and probably replace the rescan interval row. Having it there with “Watcher: On (rescans at 7200s)” or whatever even in the default case gives peace of mind that it’s actually running.  

When it’s enabled but not running we should show an error indication similar to when a folder stopped. (XXX: and change the scan interval...? Not a UI question)

If there are cases when it’s “half running” we need to figure that out. For example if it’s running fine and I copy in a directory into the folder and now it exceeds the system max number of open watch frobbits. 

We should not, at this time, automatically enable watches for existing folders. We may want to have it on by default for newly created folders?

uok:
Thanks for working on this feature!
My suggestion is to have setting to enable either "interval" OR "watcher" to not confuse the user. In the UI it could be displayed as "rescan interval: 60s" vs. "rescan interval: automatic". I don't like the idea of enabling it for existing folders, but for new folders it makes sense (maybe take it out of advanced settings? not everybody has a bullish cpu or changes files that often)

imsodin:
The one thing that isn't addressed: What will we call it? We kind of delayed this before. Given "watch" is already in the config names and it is actually descriptive (it watches out for changed files), I think we should keep that. In the UI we could add/focus on "change" instead of "FS" as it is probably easier to grasp what it does to the user.

Permissions are picked up by all platforms except mac. It may be possible to do there too, but the native event names aren't conclusive and I didn't dive deeper into it. I don't know about mtime only changes. In any case having no rescan intervals at all isn't a good idea in general, as watcher relies on various different backends which may have deficiencies (and for some of those, I heard that they do).

[...]

PR #4713, opened 2018-01-27
all: Display fs watcher status and retry starting it (ref #4552)

imsodin:
### Purpose

Get an actual status notification about the fs watcher working in web UI and don't just run away if setting the watcher up fails, retry every minute. The warning is only logged if it the first error after starting the watcher.  
This is a first step towards fully integrating the change watcher into the UI, and afterwards enabling it by default. This more or less fixes how we "call" this feature, so if this should be different, this is the time to complain.. I went for "Watch For Changes" which is more of a short description of what it does, than an actual name. I personally prefer it that way, "proper" names are annoying :)

### Testing

Still pass and runs.

### Screenshots

The labels are outdated, it is not <strike>"Watching For Changes"</strike>, but "Watch For Changes" now.

When everything is fine:  
screenshot_2018-01-27_18-01-36
When watcher failed:  
screenshot_2018-01-27_18-01-10


lib/model/folder.go
```diff
@@ -8,12 +8,17 @@ package model
...
+	"errors"
 	"time"
 
 	"github.com/syncthing/syncthing/lib/config"
+	"github.com/syncthing/syncthing/lib/ignore"
+	"github.com/syncthing/syncthing/lib/sync"
 	"github.com/syncthing/syncthing/lib/watchaggregator"
 )
 
+var errWatchNotStarted error = errors.New("Filesystem change watcher has not been started yet")
```
calmh: The error itself should just be a sentence fragment without upper case, with whatever does the actual logging in the end providing the context.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

`errors.New("not started")`

lib/model/folder.go
```diff
@@ -127,24 +136,60 @@ func (f *folder) scanTimerFired() {
...
 	f.model.fmut.RUnlock()
-	eventChan, err := f.Filesystem().Watch(".", ignores, ctx, f.IgnorePerms)
-	if err != nil {
-		l.Warnf("Failed to start filesystem watcher for folder %s: %v", f.Description(), err)
-	} else {
-		f.watchChan = make(chan []string)
-		f.watchCancel = cancel
-		watchaggregator.Aggregate(eventChan, f.watchChan, f.FolderConfiguration, f.model.cfg, ctx)
-		l.Infoln("Started filesystem watcher for folder", f.Description())
+	f.watchChan = make(chan []string)
```
calmh: When I look at it now I'm a bit suspicious that setting these without locking might be a data race. I'll try to prove it one way or the other...

calmh: I can't tickle the race detector to complain about it, so I guess it's fine

imsodin: `startWatch` is only ever to be called synchronously from the main for loop in the folder's `Serve` - that's why `scheduleWatchRestart` exists and probably deserves a few comments.

lib/model/folder.go
imsodin: That comes up in almost every review - I seem to be unable to fix it in my ... brain. I just wanted to add `grep -rP '(errors.New|fmt.Errorf)\("[A-Z]' lib/ cmd/` as a crutch to my local "build script", and there's like a lot of pre-existing capitalized error strings (not from me though xD ).

imsodin:
> I don't think it looks good to have the same refresh icon for both status lines, even if the intention is to indicate that they both relate to each other.

I agree, the design is not nice. Originally I intended to merge the two, i.e. have just one row called "Rescan", with possible value "periodic: *interval*, watching for changes (failed)" but that's way too long. I think as the two functions are about the same thing, some kind of visual linking is necessary. Having just one row may still work with using symbols (a clock\ for periodic and an eye for watcher) and having a description as a tooltip. That would fit in with how we display the global/local state.

What do you think about that or any other ideas how to represent scanning states nicely?

calmh:
I think that idea is fine, or possibly as an alternative two rows where the second has no icon and is just indented one icon-space to show that it belongs with the one above. But it might not make sense visually, would have to see it to know.

imsodin:
Whitespace is somehow messed up and tooltip is not done (translation nightmare), but in principle it would look something like this (I kind of like it):

screenshot_2018-01-28_12-12-26

Maybe the string should just be "error, retrying" instead of "failed to setup, retrying" or something completely different.

calmh:
LGTM

AudriusButkevicius:
lgtm too I guess.

imsodin:
Is there a way to use translation strings in JS (syncthingController.js)?  
Or the broader problem: I want a conditional tooltip like  
"Full rescans every {{interval}} and watching for changes"  
"Full rescans every {{interval}} and failed watching for changes (retrying every minute): {{err}}"  
"Full rescans every {{interval}} and disabled watching for changes"  
One way would be to create a function that generates this string, for which I would need translation in JS. Other than that I don't see how to achieve that short of having a top-level ng-if for every possible case.

calmh:
I'm sure there is a way to translate the strings from within JS, but some of the tooling expects all of the strings to be present in the HTML when generating lang-en.json. So I suggest just ng-if:ing it and having the strings you mention as three alternatives in the HTML. Maybe there is some sort of magic to have them as three spans and just pull in the one you want by ID into the tooltip or something, although I'm not sure.

imsodin:
I made little adjustments, it now looks like this:

screenshot_2018-01-28_14-00-31
screenshot_2018-01-28_14-00-48

One problem: `{{ 'string' | translate }}` type translations needed for tooltips don't allow for data placeholders in there (at least I wouldn't know how), so the translation strings are split up.

bugith:
Please help to translate : does the latest tooltip means "Disabled (periodic scanning AND watching for changes)" or "Disabled periodic scanning, now watching for changes)" ?
Thank you

imsodin:
@bugith The split up tooltip strings didn't make it into the translation, so they may not make a lot of sense right now. I will fix this asap - thanks for pointing to the problem! This PR is unrelease and the translation strings are only in lang-en.json, not in the files of the other languages, so this shouldn't be on transifex yet, right? Just wondering how this caught your attention :)

bugith:
I just sometimes have look in transifex and PR's to see if there is some untranslated strings for upcoming releases, just to make sure I don't forget anything ;-)
For this ~issue~ I guess these are strings 67 92 187 210 & 219
Have a good day man !

bugith:
And yes, you're right : these tooltip texts are not in TF. No worry, there are other, albeit some are there (e.g. local/global details). Although, your screenshots above were great help for me to do my best possible job ASAP (i.e. not waiting for the release to discover I missed strings and sometimes loose my hairs searching in the GUI where is that damned string I forgot, or in the best case adjust my blind translation)

PR #4732, opened 2018-02-04
vendor: Change vendoring mechanism to dep, update dependencies (fixes #4727)

### LINKED ISSUE #4727: Use dep
I've tried it, and it now actually works for our use cases. The day to day usage is simple (just edit import paths as usual and run `dep ensure` and you're done). It's supposed to be the future thing for Go package management.

calmh:
### Purpose

Align with the future of the Go community


calmh:
Blocked on https://github.com/golang/dep/issues/1625, dep not removing symlinks which cause snapcraft to fail.

PR #4733, opened 2018-02-04
dockerfile: Add Dockerfile

calmh:
This adds a multi stage build Dockerfile. The end state is equivalent to
the current syncthing/docker repository (which will be retired).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ COPY . .
 ENV CGO_ENABLED=0
 ENV BUILD_HOST=syncthing.net
 ENV BUILD_USER=docker
-RUN go run build.go build syncthing
+RUN rm -f syncthing && go run build.go build syncthing
 
 FROM alpine
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+FROM golang:1.9 AS builder
+
+WORKDIR /go/src/github.com/syncthing/syncthing
+COPY . .
+
+ENV CGO_ENABLED=0
+ENV BUILD_HOST=syncthing.net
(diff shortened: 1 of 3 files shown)


lkwg82:
consider basic tests on the image (see https://blog.lgohlke.de/docker/tdd/devops/2017/02/23/pragmatic-approach-to-add-tests-for-docker-files.html)

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM golang:1.9 AS builder
+
+WORKDIR /go/src/github.com/syncthing/syncthing
+COPY . .
+
+ENV CGO_ENABLED=0
+ENV BUILD_HOST=syncthing.net
```
rubenbe: BUILD_HOST and BUILD_USER don't work in my environment, the resulting syncthing binary announces itself as built by unknown-user:
` sudo docker run -it --rm --entrypoint syncthing newsyncthing -version
syncthing v0.14.44-rc.2+24-g1aabe427-docker "Dysprosium Dragonfly" (go1.9.3 linux-amd64) unknown-user@8459f1a59c00 2018-02-04 14:27:14 UTC`

calmh: Only way I can imagine this failing is if the binary was already in place and up to date so the build.go script doesn't rebuild it... adding an rm to the Dockerfile to counter that scenario

rubenbe: OK, facepalm, I'm not awake. You can discard my comment. I was on the wrong docker branch.

calmh: Oh right, good, I pushed an early version to the main repo to test an automated build, yes. Will remove to avoid confusion :D

PR #4735, opened 2018-02-04
lib/fs: Further unflake watch tests

calmh:
### Purpose

Always explicitly allow the directory create event that we might get.


calmh:
@st-review merge

st-review:
@calmh: Merge failed:

```
Nothing to merge, as far as I can tell.
```

PR #4741, opened 2018-02-07
Remove KCP (fixes #4737)

### LINKED ISSUE #4737: Remove KCP
The KCP implementation didn't live up to our expectations and should be removed. We will aim to add TCP hole punching when the required support in standard library is available, and monitor for other suitable technologies in the meantime.

calmh:
### Purpose

Less KCP.

### Testing

Uhm.


calmh:
@canton7 @Nutomic @kozec This outright removes fields from the config. We'll still accept (and ignore) them when posted, but they'll be missing in the JSON we produce. Problem?

AudriusButkevicius:
We should make sure we don't complain about unknown addresses.

calmh:
Ah, good point

Nutomic:
The only fields that change are DefaultKCPPort and DefaultStunServers, right? I've never used them so it will be fine.

canton7:
Nope, doesn't affect me.

AudriusButkevicius:
So having checked lib/connections, this probably needs a new error type errDeprecated in getDialerFactory that does not cause a l.Info.

AudriusButkevicius:
Otherwise this looks good.

PR #4747, opened 2018-02-09
lib/protocol: Invalid files should always lose

imsodin:


AudriusButkevicius:
How does this solve the 'my files ain't downloading' issue?

imsodin:
This doesn't solve any issue I am aware of, but it prevents an invalid file with concurrent version to be put ahead of a valid file in the global list: https://github.com/syncthing/syncthing/blob/master/lib/db/leveldb_transactions.go#L154

imsodin:
@AudriusButkevicius Regarding your last comment: Is this PR good to go or is there something to discuss?

AudriusButkevicius:
lgtm

PR #4748, opened 2018-02-09
Dynamic number of pullers

calmh:
This makes the number of pullers vary with the amount of outstanding data instead of being a fixed number. It works, as far as I can tell from running integration tests.

View with ?w=1 or suffer the consequences.

A number of tests were tightly coupled to there being a channel for pull requests, which there isn't any longer. Need to consider what they were testing and if there's still a reasonable way to do that... 

Could also retain the previous architecture and just have the copier routines block when the amount would be exceeded. We could then hardcode it to something like 128 or 256 routines as they are essentially free when they're not doing anything.


AudriusButkevicius:
Looks good, apart from loosing tests :(

imsodin:
Instead of doing the limiting in the copier, couldn't you just send to the pullChan as before and do the limiting there?

calmh:
Here's a less invasive version... I think it's Ok, going to stress it a bit though.

calmh:
No, your idea is better. I’ll fix that.

lib/model/rwfolder.go
```diff
@@ -342,14 +339,12 @@ func (f *sendReceiveFolder) pullerIteration(ignores *ignore.Matcher, ignoresChan
...
-		go func() {
-			// pullerRoutine finishes when pullChan is closed
-			f.pullerRoutine(pullChan, finisherChan, requestLimiter)
-			pullWg.Done()
-		}()
-	}
+	pullWg.Add(1)
+	go func() {
+		// pullerRoutine finishes when pullChan is closed
+		f.pullerRoutine(pullChan, finisherChan)
```
imsodin: I'd just pass `pullWg` to pullerroutine instead of creating a new waitgroup inside it and having this "artificial" wait-group for a single puller routine.

lib/model/rwfolder.go
```diff
@@ -1367,76 +1360,99 @@ func verifyBuffer(buf []byte, block protocol.BlockInfo) ([]byte, error) {
...
 
-			// Verify that the received block matches the desired hash, if not
-			// try pulling it from another device.
-			_, lastError = verifyBuffer(buf, state.block)
-			if lastError != nil {
-				l.Debugln("request:", f.folderID, state.file.Name, state.block.Offset, state.block.Size, "hash mismatch")
-				continue
-			}
+func (f *sendReceiveFolder) pullBlock(state pullBlockState, out chan<- *sharedPullerState, requestLimiter *byteSemaphore) {
+	// Get an fd to the temporary file. Technically we don't need it until
```
AudriusButkevicius: Don't see the limiter used in this function

AudriusButkevicius:
Apart from Simon's comment, lgtm

calmh:
Build is fail because I upgraded the image to 1.10rc2 and it doesn't work. Taking it down to 1.10rc1 again for the RC...

lib/model/rwfolder.go
calmh: I disagree. It looks redundant, but doing it this way retains the behavior that the function call is synchronous, like all the other ones around it, which is easier to reason about.

imsodin: I don't understand the part about synchronous behaviour: As it is inside a `go func() {...}()` construct I'd think it's asynchronous anyway and there wouldn't be any different behaviour with `go f.pullerRoutine(pullChan, finisherChan, pullWg)`?

calmh: Maybe my wording was incorrect. What I mean is that I think it's cleaner that, when you launch a goroutine, it's clear what will happen when it exits and how we know that it's exited. Passing a waitgroup and expecting the function to decrement it when it exits is a bit unusual - and we're giving the function stuff to do that is not part of its main concern. There's no need to make the function specific to being called in a goroutine, it can just do it's thing and the caller gets to manage state.

imsodin:
Looks good to me too, just one question: How do we handle the configuration change? I'd say pullers is obscure enough to just drop it with a note in the release note and be done with it - I guess @Nutomic @canton7 @kozec don't have anything specific for pullers in their wrappers (?). Should `maxOutstandingBytes` become configurable and thus replace `pullers` as advanced "knob"? To me it seems to be unnecessary, but maybe some users do rely on it (we don't have any usage reporting on that unfortunately).

calmh:
I’m hoping we can just drop `pullers` (docs need updating though). Preferably I’d avoid adding the outstanding data amount as a config, we already have too many inscrutable configs...

calmh:
And, of course, I've now run into a legitimate situation where this might need to be tweaked down from the default, so a config it is.

imsodin:
Maybe the config variable should be called `PullerMaxPendingKiB` to make it more explicit what it is. And there should probably be a config transition that converts non-standard puller numbers to the equivalent in the new config.
Otherwise this seems good to go, or is there still something WIP/POC-ish that I didn't notice?

calmh:
Yeah, updated according to suggestions. Not WIP or POC any more as far as I'm concerned.

calmh:
We’ll accept but ignore it, but it won’t be present in the config we return over the REST API.

PR #4750, opened 2018-02-10
 lib: Handle metadata changes for send-only folders (fixes #4616, fixes #4627)

### LINKED ISSUE #4616: Send only folder out of sync due to ignored items
Something is buggy here. Two machines, A and B. A is receive only and has ignore patterns:

screen shot 2017-12-27 at 21 55 00

B is newly set up and has no ignore patterns. I browse a few files on B, creating .DS_Store files. Now, on A:

screen shot 2017-12-27 at 21 55 47

screen shot 2017-12-27 at 21 55 54

Hitting "override changes" deletes the ignored .DS_Store files...

v0.14.42, gut feeling says #4460 is to blame somehow.

### LINKED ISSUE #4627: A file deleted from all nodes may exist in the "out of sync" list
When a file is deleted from the slave first, and then from the read-only master. It disappeared and later existed again in the "out of sync" list on the master. I feel it should be excluded permanently in the list.

My steps to reproduce:
- Create 4 files 0, 1, 2, 3 on the master node. Set the folder in the master to be "send only"
- Do an initial synchronization
- Delete 2, 3 on the slave
- Rescan on the slave and synchronize
- Delete 1, 2 on the master
- Pause and resume on the master (optional)
- The "out of sync" list shows only 3
- Later (I tried 3 times)
  - In 2 attempts, the final list showed after a pause and a resume
  - In 1 attempt, I paused and resumed several times before seeing the count of "out of sync" list to be 2
- Now we have 0, 3 on the master and 0 on the slave
- Eventually the "out of sync" list shows 2 and 3. Since 2 is already deleted from the master and the slave, I think it shouldn't appear here

I tried to restart Syncthing or the operation system, but the problem persisted

Screenshots from the master:

untitled

untitled2

### Version Information

Syncthing Version: v0.14.42
OS Version: 

Master: Windows7 (64-bit)  Slave: Android 7.1
I also encountered a similar issue with Master: Android 4.4    Slave: Android 7.1

imsodin:
As promised, this is the followup to #4584, fixing two more bugs. There are two main changes in behaviour:

The first one is simple: Send-only folders gain a "puller", which is triggered the same way as for the send-receive folder. It checks needed files for ignored files and files, that only differ by internal metadata.  

The second one is about unignoring files: Before this PR when a file is unignored, during scanning its version is updated. I.e. the unignored, potentially vastly outdated compared to the global version file is propagated to all other devices. The new behaviour is, that such a file is moved into conflict with the current global file. This is achieved by removing all counters except our own (which was increased) in the version. This obviously has the potential to create a lot of conflict files, but I consider this vastly preferable to silently overwriting the current global state with the previously ignored file.


calmh:
I haven't read this yet but it sounds good.

Can it also fix the case where two devices are initialized with the same set of files and one is send only? Currently it'll end up about 50% out of sync, but the send only side could accept changes that only affect the version vector (I.e., not file data or metadata) and it would just work.

imsodin:
Yes it can - I guess this is the main usage scenario next to "both devices delete a file" described in the issue.

The current implementation has one downside I didn't mention above: All out-of-sync files in a send-only folder are looked up in the database whenever an index update comes in. This could be improved, if on index update only the "new" files are looked up, but afaik there is no way to distinguish these from "old" ones. I don't think this is very bad though compared to what it gains.

One more thing: The condition determining whether two files are equal is aweful:
```
curFile.Deleted != intf.IsDeleted() || (!curFile.Deleted && (file.Size != curFile.Size || file.Type != curFile.Type || (!f.ignorePermissions(file) && file.Permissions != curFile.Permissions) || !file.ModTime().Equal(curFile.ModTime()) || !blocksEqual(file.Blocks, curFile.Blocks)))
```
Similar conditions exist in walk.go. I was considering consolidating all of these to the protocol package in the form of a `func (f FileInfo) IsEqual(other FileInfo) bool` function - does that sound reasonable?

lib/db/leveldb_dbinstance.go
```diff
@@ -351,12 +351,6 @@ func (db *Instance) withNeed(folder, device []byte, truncate bool, needAllInvali
 			if bytes.Equal(v.Device, device) {
 				have = true
 				haveFileVersion = v
-				// We need invalid files regardless of version when
```
calmh: What's the thing behind the need invalid stuff that we just added now no longer making sense? Just a couple of sentences of background for my edification?

(It's a minus diff and I was confused when it was added so I'm not complaining, it's just that I'm now confused again.)

lib/model/folder.go
```diff
@@ -238,3 +251,25 @@ func (f *folder) setError(err error) {
 
 	f.stateTracker.setError(err)
 }
+
+// Helper function to check whether either the ignorePerm flag has been
+// set on the local host or the FlagNoPermBits has been set on the file/dir
+// which is being pulled.
+func (f *folder) ignorePermissions(file protocol.FileInfo) bool {
+	return f.IgnorePerms || file.NoPermissions
```
calmh: Micro-nit: I'm not sure this actually merits a helper function. The code is shorter and clearer than the comment.

lib/model/requests_test.go
```diff
@@ -286,17 +284,159 @@ func TestRequestVersioningSymlinkAttack(t *testing.T) {
...
+	timeout = time.NewTimer(5 * time.Second)
+	select {
+	case <-timeout.C:
+		t.Fatalf("timed out before index was received")
+	case <-done:
+		return
 	}
+}
+
+func TestPullInvalidIgnoredSO(t *testing.T) {
```
calmh: Nano-nit: put these above the helper function, so the helper function makes sense when encountered. (I had to cmd-f to see if it was actually called).

lib/model/requests_test.go
```diff
@@ -286,17 +284,159 @@ func TestRequestVersioningSymlinkAttack(t *testing.T) {
 	}
 }
 
-func setupModelWithConnection() (*Model, *fakeConnection, string) {
-	tmpFolder, err := ioutil.TempDir(".", "_request-")
-	if err != nil {
-		panic("Failed to create temporary testing dir")
+// This test checks that (un-)ignored/invalid/deleted files are treated as expected.
+func pullInvalidIgnored(t *testing.T, ft config.FolderType) {
```
calmh: t.Helper() somewhere here

lib/model/rofolder.go
```diff
@@ -70,3 +75,62 @@ func (f *sendOnlyFolder) String() string {
...
+	folderFiles.WithNeed(protocol.LocalDeviceID, func(intf db.FileIntf) bool {
+		if len(batch) == maxBatchSizeFiles || batchSizeBytes > maxBatchSizeBytes {
+			f.model.updateLocalsFromPulling(f.folderID, batch)
+			batch = batch[:0]
+			batchSizeBytes = 0
+		}
+
+		if ignores.ShouldIgnore(intf.FileName()) {
+			file := intf.(protocol.FileInfo)
+			file.Invalidate(f.model.id.Short())
```
calmh: (We have this and `f.model.shortID` sprinkled both here and in rwfolder so not blaming you, but in both cases I think we should just get a `myID` in the constructor and remember it instead of reaching into the model... Doesn't need to be fixed in this PR, just mentioning it so it's out there and maybe someone remembers to clean it up)

lib/model/rofolder.go
```diff
@@ -70,3 +75,62 @@ func (f *sendOnlyFolder) String() string {
...
+		curFile, ok := f.model.CurrentFolderFile(f.folderID, intf.FileName())
+		if !ok {
+			if intf.IsDeleted() {
+				panic("Should never get a deleted file as needed when we don't have it")
+			}
+			return true
+		}
+
+		file := intf.(protocol.FileInfo)
+		if curFile.Deleted != intf.IsDeleted() || (!curFile.Deleted && (file.Size != curFile.Size || file.Type != curFile.Type || (!f.ignorePermissions(file) && file.Permissions != curFile.Permissions) || !file.ModTime().Equal(curFile.ModTime()) || !blocksEqual(file.Blocks, curFile.Blocks))) {
```
calmh: *This*, on the other hand, *does* merit a helper function!

lib/db/leveldb_dbinstance.go
imsodin: Before the invalidation of ignored files was introduced, `WithNeed` always returned ignored files, as it's version wasn't updated. Their version was updated to keep "in-sync" with the global version. Therefore these files would not get pulled unless the global version gets updated. The removed option returned all invalid files and was triggered when ignore patterns change. With the new mechanism of setting the version to conflict with the global state when unignoring, this is no longer needed.

lib/model/rofolder.go
imsodin: Yep. I proposed in my last comment to move this check and the similar checks in walk.go to the protocol package into a `func (f FileInfo) IsEqual(other FileInfo) bool` function - does that sound like the right place?

lib/model/rofolder.go
imsodin: Done anyway (the `f.model.id.Short()` thing is likely entirely my product). I also took the opportunity to structure the folder members by function.

lib/model/rofolder.go
imsodin: @calmh I went ahead with it, [see this diff](https://github.com/syncthing/syncthing/pull/4750/files/f3409d3a4d05aa5577b8eef638bf8fe6b2ec7727..d9adcd584ffc983a2acfd607b88cb2c23452112f).

lib/protocol/bep_extensions.go
```diff
@@ -126,6 +127,72 @@ func (f FileInfo) IsEmpty() bool {
...
+//  - invalid flag
+//  - permissions, unless they are ignored
+// A file is not "equal", if it has different
+//  - modification time
+//  - size
+//  - blocks, unless there are no blocks to compare (scanning)
+// A symlink is not "equal", if it has different
+//  - target
+// A directory does not have anything specific to check.
+func (f FileInfo) IsEqual(other FileInfo, ignorePerms bool, noBlocks bool) bool {
```
calmh: Params should either both be `ignore...` or `no...`, imho

lib/protocol/bep_extensions.go
```diff
@@ -126,6 +127,72 @@ func (f FileInfo) IsEmpty() bool {
...
+	if f.IsEmpty() || other.IsEmpty() {
+		return f.IsEmpty() == other.IsEmpty()
```
calmh: With this definition, two fileinfos can be "equal" if they both have no version, even if everything else is different. That's odd and might bite us in the future.

lib/protocol/bep_extensions.go
```diff
@@ -126,6 +127,72 @@ func (f FileInfo) IsEmpty() bool {
...
+	}
+
+	// characteristics common to all types
+	if f.Type != other.Type || f.Deleted != other.Deleted || f.Invalid != other.Invalid || (!(ignorePerms || f.NoPermissions || other.NoPermissions) && !PermsEqual(f.Permissions, other.Permissions)) {
```
calmh: Break this up a bit, especially the permissions juggling.

Also, we're considering fileinfos equal even if the name differs?

lib/protocol/bep_extensions.go
```diff
@@ -126,6 +127,72 @@ func (f FileInfo) IsEmpty() bool {
...
+	switch f.Type {
+	case FileInfoTypeFile:
+		return f.Size == other.Size && f.ModTime().Equal(other.ModTime()) && (noBlocks || BlocksEqual(f.Blocks, other.Blocks))
+	case FileInfoTypeSymlink:
+		return f.SymlinkTarget == other.SymlinkTarget
+	case FileInfoTypeDirectory:
+		return true
+	}
+
+	// Should the deprecates symlinks type be handled somehow?
```
calmh: no

lib/protocol/bep_extensions.go
```diff
@@ -126,6 +127,72 @@ func (f FileInfo) IsEmpty() bool {
...
+// Permissions (config) and blocks (scanning) can be excluded from the comparison.
+// Any file info is not "equal", if it has different
+//  - type
+//  - deleted flag
+//  - invalid flag
+//  - permissions, unless they are ignored
+// A file is not "equal", if it has different
+//  - modification time
+//  - size
+//  - blocks, unless there are no blocks to compare (scanning)
```
calmh: Why don't we care about the version vector? I'm generally suspicious of this definition of "IsEqual". IMHO, "IsEqual" means that every attribute on the struct is exactly identical and nothing else. Possibly we can get away with `ignorePerms` and `ignoreBlocks`, but nothing else should be skipped unless we explicitly say so...

lib/protocol/bep_extensions.go
```diff
@@ -126,6 +127,72 @@ func (f FileInfo) IsEmpty() bool {
...
+		return a&0600 == b&0600
+	default:
+		// All bits count
+		return a&0777 == b&0777
+	}
+}
+
+// BlocksEqual returns whether two slices of blocks are exactly the same hash
+// and index pair wise.
+func BlocksEqual(src, tgt []BlockInfo) bool {
```
calmh: This used to live in `scanner`, then someone refactored that to a private function in `model` for some reason, and now it's in `protocol`. We should stop moving things around...

lib/protocol/bep_extensions.go
imsodin: I should have stuck with the original name `IsEquivalent` or maybe `Unchanged`. Version (and Sequence, potentially others) is purposely excluded. The job of this function is to identify file infos that represent the same underlying files (currently used for scanning and send-only resolution). The conditions have been take over from walk.go.

lib/protocol/bep_extensions.go
imsodin: Yeah - I just needed it here and I certainly wont dip into model from protocol (can't anyway because circular) and as I thus had to move it, protocol felt like the right place (it's operating on protocol.BlockInfo after all).

calmh:
Good stuff.

Issue #4756, opened 2018-02-13
v0.14.45-rc eats files


calmh:
As reported in https://forum.syncthing.net/t/files-deleted-after-update-to-v0-14-45-rc-2/11282

Potentially multiple issues, almost surely related to #4584.

Off the top of my head I can see one issue with directories that have been renamed case-only on case insensitive systems: whereas previously we would see the files just fine, we would now lockstep-walk the database vs the filesystem, potentially note that the directory we expected didn't happen, and mark everything below as deleted. That is, we see `foo/` when walking and conclude that`Foo/**` which is in the database has been deleted. This propagates to some other device where the deletes happen, and it's a self fulfilling prophesy.

JonasT:
Is there a mechanism to automatically revert release candidates on affected machines as well? It might be worth putting one in: I just randomly saw odd messages and thought "huh I might try the older version to fix it", and if I hadn't looked at it, that version would have run forever even though you guys seem to already be aware that it's broken. (and deleted whatever in the long run)

calmh:
The switch we can throw is to release a newer version. I don't think we want a central kill- or downgrade-switch.

PR #4757, opened 2018-02-13
all: Revert simultaneously walk fs and db on scan

imsodin:
This reverts commit 6d3f9d5154b4c6b39f8083255cc9ed36b5c51146 / #4584.

In the presence of both case sensitive and insensitive systems this is seriously flawed, see #4756 and https://forum.syncthing.net/t/files-deleted-after-update-to-v0-14-45-rc-2/11282/19. Admittedly there is now evidence for that, but circumstances and Occam's razor point to it.

This undos the fixes for #2571 and #4573.


calmh:
Yeah, hold this horse for just a moment.

imsodin:
Sure, sure, just didn't think and expected merge conflicts to happen and thus wanted to get that work done - wasn't any work though :P

calmh:
I have nothing smarter to offer in the short term

Issue #4758, opened 2018-02-13
Suture services should not survive panics


calmh:
We use `suture.Service`s in a lot of places and I like the interface, and that we get restarts on error returns etc. I'm not fond of the fact that it transparently swallows panics.

For example, a `panic("bug: foo")` in the database layer that we expect should be loud and fatal currently goes silently unnoticed if it's called from a folder runner, as that is a suture.Service and it just gets automatically restarted. Apart from the obvious issues, this leaves us with deadlocks if the panic happened inside a lock that wasn't defer-unlocked. 

I suggest we add a `PanicsPanic` options to the package, and see if we can get it accepted upstream. If not, I can live with a fork (it's a very stable package, seldom updated).

PR #4761, opened 2018-02-14
lib/db: Don't panic on negative counts

calmh:
So, negative counts should never happen and hence the original idea to
panic. However, this sucks as the panic will happen in a folder runner,
be automatically swallowed by suture, and the runner gets restarted but
now we are in a bad state. (Related: #4758)

At the time of writing the global list is somewhat in flux (we've
changed how ignored files are handled, invalid bits, etc.) and I think
that can cause unusual conditions here. Hence just fixing up the numbers
instead until the next full recount interval.


AudriusButkevicius:
Shouldn't we just null out the database and recheck in that case?

Issue #4762, opened 2018-02-14
Failes to create folder root directory on windows


wwwutz:
( referenced in https://forum.syncthing.net/t/error-when-sync-dest-to-nonexistant-root-directory-on-windows/11277/9 )

Call to `os.MkdirAll( '\\?\F:\BOOM', permissions )` with `\\?\` UNC prefix fails. This happens, when `F:\BOOM` does not exist.

code lines: 

https://github.com/syncthing/syncthing/blob/68c1b2dd472f18c3c37bb188fdb106df55dea927/lib/fs/basicfs_windows.go#L78

### Version Information

Syncthing Version: v0.14.45-rc.2+1-gb57c9b6a-dirty
OS Version: Windows 7

PR #4763, opened 2018-02-14
lib/fs: Fix MkdirAll failure due to \\?\

imsodin:
https://forum.syncthing.net/t/error-when-sync-dest-to-nonexistant-root-directory-on-windows/11277/18  
Main purpose of this PR for now is to provide diffs/builds for testing to @wwutz that actually compile on windows (anyone else interested is obviously also welcome to test).

To reproduce run the following while `F:\FOO` does not exist:

```golang
package main

import (
	"fmt"

	"github.com/syncthing/syncthing/lib/fs"
)

func main() {
	base := `F:\`
	targ := `FOO\BAR`
	testFs := fs.NewFilesystem(fs.FilesystemTypeBasic, base)
	if err := testFs.MkdirAll(targ, 777); err != nil {
		fmt.Printf("Base: %v, target: %v, err: %v", base, targ, err)
	}
}
```
Result:  
`Base: F:\, target: FOO\BAR, err: mkdir \\?: The filename, directory name, or volume label syntax is incorrect.`

@calmh Am I right assuming that a unit test can't successfully run ```os.MkdirAll(`C:\FOO\BAR\`)``` on the build server?

### DIFF
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -154,6 +154,19 @@ func (f *BasicFilesystem) Mkdir(name string, perm FileMode) error {
 	return os.Mkdir(name, os.FileMode(perm))
 }
 
+// MkdirAll creates a directory named path, along with any necessary parents,
+// and returns nil, or else returns an error.
+// The permission bits perm are used for all directories that MkdirAll creates.
+// If path is already a directory, MkdirAll does nothing and returns nil.
+func (f *BasicFilesystem) MkdirAll(path string, perm FileMode) error {
+	path, err := f.rooted(path)
+	if err != nil {
+		return err
+	}
+
+	return f.mkdirAll(path, os.FileMode(perm))
+}
+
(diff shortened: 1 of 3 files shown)


wwwutz:
My firsts tests went smooth. I added `D:\ABC1`, `D:\ABC2\ABC` and `D:\ABC3\ABC\ABC` and all folders got created. Now it would also be possible to directly sync to a thumbdrive.

wwwutz:
When we're not able to write to `C:`, maybe it is possible to map a local folder to a drive letter on the test system ? Like mentioned in https://superuser.com/questions/644684/mapping-drive-letters-to-local-folders

calmh:
I'm not sure what the permissions do on the windows build server exactly, but regardless I don't want our tests to start probing stuff at c: (or z: or whatever) on other people's systems.

calmh:
Wasn't there a bug filed for this somewhere? Feel free to un-WIP and merge

wwwutz:
Bug filed was #4762 , so this is fixed.

PR #4765, opened 2018-02-14
lib/model: More precise deletion detection (fixes #2571, fixes #4573)

### LINKED ISSUE #2571: Replacing a directory with a symlink or vice versa is buggy

### LINKED ISSUE #4573: Inaccessible files during scan are marked as deleted

imsodin:
This fixes the same issues as #4584 but without the combined db/fs scan. The basic idea is to only mark files as deleted, if we really are sure they are deleted. I.e. only if the lstat error is `IsNotExist` or a parent is a link/file.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -2053,30 +2053,23 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 				// still here.
 				// Simply stating it wont do as there are tons of corner
 				// cases (e.g. parent dir->simlink, missing permissions)
-				if osutil.IsDeleted(mtimefs, f.Name) {
-					nf := protocol.FileInfo{
-						Name:       f.Name,
-						Type:       f.Type,
-						Size:       0,
-						ModifiedS:  f.ModifiedS,
-						ModifiedNs: f.ModifiedNs,
-						ModifiedBy: m.id.Short(),
-						Deleted:    true,
-						Version:    f.Version.Update(m.shortID),
-					}
-					// We do not want to override the global version
-					// with the deleted file. Keeping only our local
-					// counter makes sure we are in conflict with any
-					// other existing versions, which will be resolved
-					// by the normal pulling mechanisms.
-					if f.IsInvalid() {
-						nf.Version.DropOthers(m.shortID)
-					}
-
-					batch = append(batch, nf)
-					batchSizeBytes += nf.ProtoSize()
-					changes++
+				if !osutil.IsDeleted(mtimefs, f.Name) {
+					return true
 				}
+				nf := protocol.FileInfo{
+					Name:       f.Name,
+					Type:       f.Type,
+					Size:       0,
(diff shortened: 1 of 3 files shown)


lib/model/model.go
```diff
@@ -2066,6 +2064,14 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 						Deleted:    true,
 						Version:    f.Version.Update(m.shortID),
 					}
+					// We do not want to override the global version
```
calmh: Move the comment inside the if, I was super confused about why we would not want to override the global file until I read on and realized it was for a corner case.

It's also confusing that we are inside an `!f.IsInvalid()` case and yet it may be invalid, but I guess that's Ok...

imsodin: I dropped the if entirely from this PR. That is the change of behaviour introduced in #4750 and thus should be done there (didn't realize it when splitting this PR from the other one).

imsodin:
I just run this PR together with the send-only metadata PR on my system and it ran smoothly.

Is this (still) good to go?

PR #4767, opened 2018-02-16
WIP: Ask remote peers to rescan

AudriusButkevicius:
Not sure if I am doing the invalidation right, yet you get the idea of what I am trying to do.


AudriusButkevicius:
@calmh comments on the generic idea?

calmh:
I'm not sure I like a protocol message for this. As far as I'm concerned this is a bug on the sender side - we're specifically asking for a block with a certain hash (or, I forget, maybe just implicitly because it claimed it had that hash to begin with) and it's giving us something else. It should not be our responsibility, as the receiver, to tell it what to do about that.

We could / should instead verify the hash as we're sending and trigger a rescan as appropriate. Of course we don't want to do that all the time as it's expensive. Potential ways to mitigate;

 - If we have a weak hash, use that. We just need a CRC check here, really...
 - Keep track of block hashes requested per peer device, in something like a size limited LRU list. Verify the hash of the block we're sending if it's a repeat request (and then mark it as verified if it actually does check out so we don't do it for requests three, four, ...).

AudriusButkevicius:
Well my argument is that receiver already does that, so why not use that as a mechanism for signaling?

calmh:
Because it adds odd cruft to the protocol that doesn’t belong there.

AudriusButkevicius:
Well in that case the last option sounds most reasonable.

AudriusButkevicius:
So do we agree with the last one as the measure?
I guess we could check scan speed and if it's below let's say 20MB/s, not do that?

AudriusButkevicius:
To be honest, my heart breaks thinking about all the phones and RPIs doing that, versus just one lousy protocol message.

calmh:
If you think that's too painful, why not choose the middle one (check the block the second time the same peer asks for it - this should never happen, really).

AudriusButkevicius:
I don't think this is always a bug, it's also interactions with other apps.

AudriusButkevicius:
Do you agree or disagree that it's not always a bug?

calmh:
I’m not sure it 100% matters, but since the protocol exchange today is

A: I have a block with hash 1234
B: Give me the block with hash 1234
A: Here’s a banana
B: ...

I think it’s a bug on A side. I think it’s reasonable to expect A to not send something other than what it said it had, and what we asked for. 

That other apps can fool Syncthing by changing the contents and retaining the timestamp is sort of beside the point. It can be resolved by A not being lazy and double checking its data. It can be optimized so this is not hideously expensive.

AudriusButkevicius:
Ok

AudriusButkevicius:
So I was thinking about this, but any LRU idea seems it's going to be a hit and miss with many small files situation.

Bloom filter would probably work okay'ish but getting the size right is hard, as at some point you'll just start flagging everything.

What about this.

We make weak hash always be there (in the scans atleast, we can still skip it on rolling local file to reduce transfers).
We send weak hashes part of the request.
We check that what we've read matches weak hash they given us, if it doesn't we go on a venture of looking stuff up in our own database, seeing if those weak hashes match, if they don't doing a strong hash comparison, rescans and what not.

Alternative would be to check if the weak hash sent is non-zero, and only do this checking in that case, but it's somewhat half-assed as then we sort of don't have a problem anymore and have it at the same time, based on the current moon phase.

calmh:
Yeah I agree on the weak hash thing. Make it mandatory for scans. Check it on transmission. I don't think we should do anything fancy on failure though - return an error code to the requestor, tag the file as invalid in the database (this will also fire an index update at some point, so we stop getting requests for it), fire a rescan (which should pick up that it's invalid in the db and do a full hash).

AudriusButkevicius:
Have a look at this once you have time.

Well tested as usual.

AudriusButkevicius:
I have no idea how to fix the serialization test, I probably need to downgrade something but it's not obvious what.

AudriusButkevicius:
Or was that fixed by hand?

AudriusButkevicius:
@imsodin have a look

lib/model/rwfolder.go
```diff
@@ -1274,17 +1274,8 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
...
-						if hash != nil {
-							l.Debugf("Finder block mismatch in %s:%s:%d expected %q got %q", folder, path, index, block.Hash, hash)
-							err = f.model.finder.Fix(folder, path, index, block.Hash, hash)
-							if err != nil {
-								l.Warnln("finder fix:", err)
-							}
-						} else {
-							l.Debugln("Finder failed to verify buffer", err)
-						}
+					if err := verifyBuffer(buf, block); err != nil {
```
imsodin: Should we trigger a rescan of the file here?

lib/model/model.go
```diff
@@ -1372,9 +1372,64 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset
...
+		} else {
+			l.Debugf("%v recheckFile: %s: %q / %q: blocks do not match", m, deviceID, folder, name, err)
+		}
+	} else {
+		l.Debugf("%v recheckFile: %s: %q / %q: error while hashing file: %s", m, deviceID, folder, name, err)
+	}
+
+	// The file probably did not get rescanned, as mtimes haven't changed.
+	// Invalidate it, and trigger another scan.
+	cf.Invalidate(m.shortID)
```
imsodin: I don't get where mtime comes into play. Neither do I understand the need to invalidate (that's maybe linked to not getting the mtime bit).  
In general wouldn't it be simpler and use less resources to rescan the file directly instead of explicitly hashing after checking the block hash. The current flow hashes a changed file twice.

lib/model/rwfolder.go
AudriusButkevicius: We are mid way through a pull, so if the stuff doesn't match, scanning fill pick it up next I suspect.

This block map fixing stuff was to deal with me fking up the first version and storing half of the hash in the map, so I am just removing the 2-3 year old code that was fixing the damage.

imsodin: Yeah right, I didn't get the context right. And thanks for the map fix info, I was sort of wondering about that but didn't see the point of the removed code, so kept silent :)

lib/model/model.go
AudriusButkevicius: The idea here is as follows:
1. We get what other side expects to get as part of the request.
2. We validate that what we've read matches the weak hash, if it does, nothing happens.
3. If it doesn't match, we validate using SHA256, if it matches it's all good.
4. If it still doesn't match, look up our own file in the database, and compare hashes for that block. If the our hashes don't match request hashes, it means the requestor is asking for something different, so we do nothing.


You are right that we don't need to hash here, as we know that what we've read doesn't match the hash the user has asked us for, and we also know that our hash matches his, which means the file is somehow changed, so we can skip the HashFile and go straight to invalidate and rescan.

The idea behind the mtime is the fact that if something modifies files in place, or artificially adjusts mtimes, we could miss some changes, hence we double check what we're reading.

imsodin: Ok, I get the invalidation now. One little nick there: In almost all cases it will result in an index update sent with the invalid file info directly followed by an update valid file info. Only in the rare case of a file that has unchanged size and mtime, but changed content, will the invalidation have any effect, in all others it is useless overhead. You could pass the file info from `Request` and check whether it is equivalent to the current file in the database (as in `(f FileInfo) IsEquivalent`)  and invalidate if it is and scan otherwise.

AudriusButkevicius: Request does not have a FileInfo, it just has folder, name, offset, size, hash, weakHash.
Essentially the goal is to force rescan a file.

AudriusButkevicius:
This will have a negative effect that the files sourced from devices that did not weak hash will now be hit with constant SHA256 checks for every request.

lib/model/model.go
imsodin: I don't know whether force rescanning is a good idea. The only real case where a  file is has changed content without changed size&mtime I can think of is bitrot. In that case we'd rather not scan and send the data.

As to what I meant with checking whether the file is changed: Within `Request` the file is stated and within `recheckFile` the current file info is obtained from the db. That's all the information needed to check whether the two are "equivalent".

imsodin:
Please not 2. Other options:  
 3. Add database transition where any file without weak hashes is rehashed fully when changed or just weak hashed if unchanged. Clean but cumbersome.  
 4. Don't ever use SHA256 for validation, just assume everything is fine as before if there are no weak hashes from either side. Makes this best effort, but generally cheaper.

AudriusButkevicius:
4 is essentially 1.

@calmh ?

calmh:
I’ll check this out tomorrow. No review time this evening.

calmh:
I'm not sure why we need to pass the weak hash in the request. As far as I understand it, we only need to check that the hash of the data we are going to send matches the hash we have on file locally, and by extension matches the weak hash we have on file locally. Adding it to the protocol requires us to nail down what the weak hash is exactly and require everyone to use it etc.

AudriusButkevicius:
Sending weak hash avoids db lookup in the best case, where as what you are proposing always hits the database.

imsodin:
> Adding it to the protocol requires us to nail down what the weak hash is exactly and require everyone to use it etc.

Isn't hash specification in general a missing piece of the BEP? Block hashes (not weak, regular ones) are extensively exchanged, but as far as I can see the algorithm isn't specified anywhere. Since a while there is also the `weak_hash` field in the block info, whose meaning isn't touched on at all. It might be a good idea to think about this in terms of the protocol, with or without including weak hashes in requests.

Some thoughts: SHA256 for hashes should probably just be made explicit in the specs (I feel like that must already be in there and I just missed it). For weak hashes explain their intended use (identification of shifted blocks and cheaper verification) and add the used algo in the cluster config. Thus an implementation can decide whether it can/should care about sent weak hashes or disregard them. Without this information weak hashes are guesswork unless you know the implementation of the connected device. Sure, this is largely theoretical due to no other mature implementations, but then again a goal of the BEP is to keep the option of alternative implementations open.

calmh:
I was sure there was some wording around the hash in the BEP spec, but there isn't. Perhaps it was lost in the transition to protobuf. What I remember it saying was that the hash algo was implied by the hash length and that 32 bytes forever implied it was a SHA256 hash. At this point we could probably just mandate that SHA256 is the one and only hash for this protocol version.

So far the weak hash has really only had local significance so we haven't had need to specify it. If we're going to depend on it we need to specify it, and then not change it. It doesn't appear to be standardized how our current adler32 does rolling... https://stackoverflow.com/questions/9699315/differences-in-calculation-of-adler32-rolling-checksum-python

AudriusButkevicius:
Pr approved then?

calmh:
Hold your horses.

AudriusButkevicius:
What am I waiting for? Counter PR or do you want me to do something with this?

calmh:
I want to read the code, to begin with. So far I just noted a concept I didn't like and got hung up on that. I still don't like it, but there may not be a better solution. Yes, I'm being a bit slow here, sorry, but that's life at the moment...

AudriusButkevicius:
Ah right, no rush, I just thought you had a better idea on how to solve this.

lib/model/model.go
```diff
@@ -1353,7 +1353,7 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset
 			return protocol.ErrNoSuchFile
 		}
 
-		if err := readOffsetIntoBuf(folderFs, tempFn, offset, buf); err == nil {
+		if err := readOffsetIntoBuf(folderFs, tempFn, offset, buf); err == nil && scanner.Validate(buf, hash, weakHash) {
```
calmh: That's a fairly expensive operation to hide at the tail end of an if condition. Maybe make it its own thing below the reading?

lib/model/model.go
```diff
@@ -1372,9 +1372,56 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset
...
+	// Update the index and tell others
+	// The file will temporarily become invalid, which is ok as the content is messed up.
+	m.updateLocalsFromScanning(folder, []protocol.FileInfo{cf})
+
+	if err := m.ScanFolderSubdirs(folder, []string{name}); err != nil {
+		l.Debugf("%v recheckFile: %s: %q / %q rescan: %s", m, deviceID, folder, name, err)
+	} else {
+		l.Debugf("%v recheckFile: %s: %q / %q", m, deviceID, folder, name)
+	}
+
```
calmh: line

lib/scanner/blocks.go
```diff
@@ -107,6 +108,29 @@ func Blocks(ctx context.Context, r io.Reader, blocksize int, sizehint int64, cou
...
+		if _, err := io.Copy(whf, rd); err == nil && whf.Sum32() == weakHash {
+			return true
+		}
+		// Copy error or mismatch, go to next algo.
+		rd.Seek(0, io.SeekStart)
+	}
+
+	if len(hash) > 0 {
+		hf := sha256.New()
+		if _, err := io.Copy(hf, rd); err == nil {
```
calmh: I don't think it's even theoretically possible that a copy from a bytes.Reader to the hasher can fail, but no harm no foul in that case

AudriusButkevicius:
@calmh so I have no idea why the proto tests are failing. Did we hand modify the generated code? Or do I need a special version of protoc?

calmh:
Uh no immediate idea, no magic should be required afaik. On mobile, but will take a look at the test later.

calmh:
6809d38c At a guess, you might need newer gogoproto / protoc, but the script should use the vendored gogoproto which is recent...

calmh:
I think your local stuff may just be old. I grabbed your branch, `go run build.go proto`, tests pass for me.

AudriusButkevicius:
I did a local go generate

AudriusButkevicius:
Right, it seems build.go updated gogoproto or something, as I can't repro running go generate locally anymore.

AudriusButkevicius:
Huh, still fails.

calmh:
Yeah I guess you need to update your gogoproto compiler or protoc or something.

calmh:
I pushed a commit to your branch. From the diff it looks like I have newer protoc stuff and I guess that does not come from vendoring

lib/scanner/blockqueue.go
```diff
@@ -106,7 +104,7 @@ func (ph *parallelHasher) hashFiles(ctx context.Context) {
 				panic("Bug. Asked to hash a directory or a deleted file.")
 			}
 
-			blocks, err := HashFile(ctx, ph.fs, f.Name, f.BlockSize(), ph.counter, ph.useWeakHashes)
```
calmh: Maybe remove this parameter altogether

AudriusButkevicius: It's still used in some cases I think

AudriusButkevicius:
The autodetection et all is still useful as it still can be disabled when pulling.

calmh:
So the autodetection is about scanning speed, and with this in we should always scan with both the real and weak hash, no? For pulling we have the min-percentage-changed threshold which I guess should remain, and can be set >100 to disable the check for weak hashes

AudriusButkevicius:
Yeah, sure, we can just keep the percentage.

AudriusButkevicius:
So the hashFile(useWeakHash) parameter is used in cpuBench which is then reported back to UR.
If we remove the parameter, we end up with sha256perf meaning sha256+adler32

lib/model/model.go
```diff
@@ -1382,9 +1382,54 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset
 		return protocol.ErrGeneric
 	}
 
+	if !scanner.Validate(buf, hash, weakHash) {
+		m.recheckFile(deviceID, folderFs, folder, name, int(offset)/len(buf), hash)
```
imsodin: Return with `ErrNoSuchFile` here.

lib/model/model.go
```diff
@@ -1382,9 +1382,54 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset
...
+	// Seems to want a different version of the file, whatever.
+	if !bytes.Equal(block.Hash, hash) {
+		l.Debugf("%v recheckFile: %s: %q / %q i=%d: hash mismatch %x != %x", m, deviceID, folder, name, blockIndex, block.Hash, hash)
+		return
+	}
+
+	// The hashes provided part of the request match what we expect to find according
+	// to what we have in the database, yet the content we've read off the filesystem doesn't
+	// Something is fishy, invalidate the file and rescan it.
+	cf.Invalidate(m.shortID)
```
imsodin: My concern from an outdated comment is still applies:  

> I don't know whether force rescanning is a good idea. The only real case where a file has changed content without changed size&mtime I can think of is bitrot. In that case we'd rather not scan and send the data.

The only benefit of invalidation that I see is that the other device won't try to pull from us again until we update. That might even be bad, if other blocks are still fine.

AudriusButkevicius: We can't force a rescan if the mtime has not changed for example, so invalidation allows us to force rescan.

imsodin: I am aware it is necessary for force rescanning, my concern is exactly about force rescanning. It feels inconsistent, as such changes will only be detected and scanned if they are requested from another device. And I still don't know about another typical case of modified file with mtime&size unchanged except bitrot.

AudriusButkevicius:
So I'll leave hash performance as it is, and leave the flag in there, as I don't see much harm in reporting two numbers rather than one, just for the same of having a flag.

What is left to do here?

calmh:
A test? I think the reques_tests.go thing has a couple of tests that runs the request path, something in there maybe could be used for inspiration

AudriusButkevicius:
Added a crude test

AudriusButkevicius:
So, what's up with this.

PR #4770, opened 2018-02-21
cmd/strelaypoolsrv: Return better error codes and messages

calmh:
The current 500 "test failed" looks and sounds like a problem in the relay pool server, while it actually indicates a problem on the announcing side. Instead use 400 "connection test failed" to indicate that the request was bad and what was the test.

### DIFF
--- a/cmd/strelaypoolsrv/main.go
+++ b/cmd/strelaypoolsrv/main.go
@@ -90,6 +90,10 @@ var (
 	evictionTimers  = make(map[string]*time.Timer)
 )
 
+const (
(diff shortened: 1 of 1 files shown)


calmh:
Yeah so we had `github.com/zillode/notify` whereas it's actually called `github.com/Zillode/notify` on Github, and dep / vgo / possibly other things complain about the mismatch so I fixed it... But then of course git, teamcity, Go, and our build process all have various issues with case only renames on case insensitive filesystems. Go figure...

AudriusButkevicius:
This can still be improved, as in we could report the ip announced versus IP seen.

PR #4772, opened 2018-02-24
all: Fix code typos

HairyFotr:


lib/osutil/symlink_windows.go
```diff
@@ -26,7 +26,7 @@ func DebugSymlinkForTestsOnly(oldname, newname string) error {
 	// '/' does not work in link's content
 	oldname = filepath.FromSlash(oldname)
 
-	// need the exact location of the oldname when its relative to determine if its a directory
```
calmh: this fix is incorrect

PR #4775, opened 2018-02-25
cmd/syncthing: Reset delta indexes after the latest index changes

calmh:
This was necessary for me to get everything to play nice after imsodins latest changes, and I think we might as well throw it in there as part of this release.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -720,7 +720,6 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 
 	dbFile := locations[locDatabase]
 	ldb, err := db.Open(dbFile)
-
 	if err != nil {
 		l.Fatalln("Cannot open database:", err, "- Is another copy of Syncthing already running?")
 	}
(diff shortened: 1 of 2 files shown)


calmh:
I think mostly it'll happen under noninteractive circumstances so I don't want to make too much noise about it. A log entry in main when it does the dropping?

AudriusButkevicius:
How about we drop delta indexes every new version?

imsodin:
An info level log entry explaining what's happening sounds reasonable too, I was even thinking only about the changelog :) 

> How about we drop delta indexes every new version?
No fan of that here - seems to heavy for something that mostly won't have any effect. If there is an actual problem there is always `-reset-deltas` and if there is something specific, we can still trigger it on update.

AudriusButkevicius:
How is that heavy? We seem to fiddle with various stuff affecting these nearly every release.

The support burden of things fucking up is much more heavy than sending some bytes over the wire. It's not that we need to rescan or anything.

imsodin:
Right, it is not generally resource heavy, but e.g. on non-highend smartphones it certainly will be noticeably slow. And in principle I don't see how our fiddling should affect the index (apparently it does anyway), as handling updates differently should not affect index sequences, thus it seems wrong to me to just go in with a hammer pre-emptively.

AudriusButkevicius:
It sucks on the phone if it happens on every startup like it used to, but one-off every new version which is once every 2 weeks, that doesn't sound too bad for me, as especially half the people don't even know how they are running syncthing and can't call command line switches (esp on mobile).
 
Anyways, I am ok with whatever you guys decide.

calmh:
"Every new release" is annoying as we don't really know on startup whether we are on the same release as last time or not. That said, clearing this out about once a month (which is the current release frequency) does seem reasonable. The mechanism that could do that is the db metadata recalculation thing which already happens once a month....

AudriusButkevicius:
We could stop abusing noop config versions and just have a mechanism for storing the version of syncthing in the database, and then if that changes, call some callback.

cmd/syncthing/main.go
```diff
@@ -746,12 +745,30 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
...
+	miscDB := db.NewNamespacedKV(ldb, string(db.KeyTypeMiscData))
+	prevVersion, _ := miscDB.String("prevVersion")
+
+	// Strip away prerelease/beta stuff and just compare the release
+	// numbers. 0.14.44 to 0.14.45-banana is an upgrade, 0.14.45-banana to
+	// 0.14.45-pineapple is not.
+
+	prevParts := strings.Split(prevVersion, "-")
+	curParts := strings.Split(Version, "-")
+	if prevParts[0] != curParts[0] {
```
imsodin: Does that work? The first time running the miscDB does not contain `prevVersion` and thus `prevParts` should be `nil`. Also maybe use the code from the upgrade package here.

calmh: Yeah, it works. Split always returns at least one string, which may be empty, so the first run will be considered an upgrade which is desired. The stuff in lib/upgrade is overkill, more cumbersome to when we just need the string comparison, and unexported.

imsodin:
So still good to merge from me.  

I think in general it would be handy if new commits would deactivate previous review approvals, which apparently is even an option offered by Github: https://stackoverflow.com/a/44747392/3864852

calmh:
It is, but it invalidates my fairly common "lgtm, just fix this typo/naming/other nitpick" pattern :)

imsodin:
I think that wasn't possible in the times of the bot (i.e. lgtm were invalidated), so that pattern shouldn't yet be too hard to untrain, right ;)  
I'd be in favour of changing it, as it fixes a relatively common pattern for me, when a requested trivial fix (with approval) turns into something (not so) slightly different than requested -  what about the approval then?

calmh:
Okay, tweaked it.

AudriusButkevicius:
hmm, no longer allows modifying subject now.

PR #4779, opened 2018-02-25
lib/connections: Fix relay connections when two devices use the same relay (fixes #4778)

### LINKED ISSUE #4778: devices connected to same relay never reconnect ?
referencing https://forum.syncthing.net/t/infrequent-irregular-disconnected-devices/11303/36

bug happens when two of your devices connect to the same relay, as the next dial time is decided per address, which happens to be the same for two devices connected to the same relay.

AudriusButkevicius:


AudriusButkevicius:
Don't merge, this doesn't work.

AudriusButkevicius:
This should work.

PR #4781, opened 2018-03-01
gui: In remote need use index and auto-expand if only one folder (fixes #4759)

### LINKED ISSUE #4759: List of out of sync items from remote device is not displayed
There is a bug in the "out of sync items" modal for remote devices. Only the header and folder name are shown although the network log shows that the JSON data is fetched. This happens if the folder name has a space in it and the generated CSS id in data-target (responsible for toggling the list) becomes invalid.

imsodin:
If there is multiple folders, it is beneficial to have an overview over them at the beginning and it should be obvious that they are expandable. In the case of a single folder, that's less so, therefore it will be expanded by default now.


calmh:
Mac builder is drunk

PR #4789, opened 2018-03-07
all: Transition to using fs watcher by default (fixes #4552)

### LINKED ISSUE #4552: UI for filesystem watcher

imsodin:
### Purpose

Add the watcher functionality to the edit folder modal, make it the default for new folders and provide a notification to enable it for existing folders. As suggested by @AudriusButkevicius I structured the foler modal with tabs like the settings.

The new default for new folders is watcher enabled and hourly full scans. For existing folders, if the user chooses so in the web UI, watcher is enable and full scans done at 60x current interval capped at once per day.

### Testing

Manual poking around.

### Screenshots

screenshot_2018-03-07_11-23-33

screenshot_2018-03-07_11-23-20

screenshot_2018-03-07_11-23-09

screenshot_2018-03-07_11-21-49

### Documentation

https://github.com/syncthing/docs/pull/379


uok:
Tabs are a great idea! I think it would also help users make use of the restore feature to enable versioning by default

imsodin:
The issues mentioned in the original PR text were cause by a dirty testing environment and lazy copy&pasting - it's now working as expected.

> I think it would also help users make use of the restore feature to enable versioning by default

That's unrelated to this PR - if you want to discuss it, the forum is the right place.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -62,8 +62,9 @@ angular.module('syncthing.core')
         $scope.folderDefaults = {
             selectedDevices: {},
             type: "readwrite",
-            rescanIntervalS: 60,
+            rescanIntervalS: 3600,
             fsWatcherDelayS: 10,
+            fsWatcherEnabled: true,
```
AudriusButkevicius: You need an equivalent change in go to have this enabled for introduced folders etc which does not involve js

gui/default/syncthing/core/notifications.html
```diff
@@ -41,3 +41,31 @@ <h3 class="panel-title"><span class="fa fa-flash"></span>&nbsp;<span translate>A
...
+      <p><a href="https://docs.syncthing.net/users/syncing.html#scanning"><span class="fa fa-fw fa-book"></span>&nbsp;<span translate>Learn more</span></a></p>
+      <p>
+        <span translate>Do you want to enable watching for changes for all your folders?</span><br/>
+        <span translate>Additionally the full rescan interval will be increased (times 60, i.e. new default of 1h). You can also configure it manually for every folder later after choosing No.</span>
+      </p>
+      <p translate translate-value-syncthing-inotify="syncthing-inotify">Warning: If you are using an external watcher like {%syncthingInotify%}, you should make sure it is deactivated.</p>
+    </div>
+    <div class="panel-footer clearfix">
+      <div class="pull-right">
+        <button type="button" class="btn btn-primary btn-sm" ng-click="activateAllFsWatchers(); dismiss()">
```
AudriusButkevicius: I suspect this might not work due to angular scoping rules.

AudriusButkevicius:
We could also fold ignores modal into this tabbed one.

gui/default/syncthing/core/notifications.html
imsodin: What do you suspect not to work? It does what I expect it to.

AudriusButkevicius: You did test it, and it did work, right?

imsodin: Correct, as in the notification vanished and fs watcher was enabled and rescan interval adjusted.

AudriusButkevicius: Right, ok.

gui/default/syncthing/core/syncthingController.js
imsodin: done

imsodin:
> We could also fold ignores modal into this tabbed one.

Agreed and already done, but I would like to have this in a separate commit, as its quite a bit of (unrelated) additional diff.

uok:
Are there any downsides to activating the folder watcher? If yes, this should also be mentioned in the yes/now notification.

imsodin:
@canton7 You already have some transition mechanism in place, right? Therefore you probably don't want the notification about enabling the watcher displayed - so you'll need to remove "fsWatcherNotification" from `unackedNotificationID` on upgrade, is that ok?

@kozec @sieren @Martchus Pinging you in case your wrappers do anything about filesystem notifications.

> Are there any downsides to activating the folder watcher? If yes, this should also be mentioned in the yes/now notification.

None are known to me (there likely are scenarios where a watcher isn't wanted, but that's not really a downside). I assumed by stating that the watcher is the new default, it is sufficiently clear that it is recommended for use, isn't it?

canton7:
> @canton7 You already have some transition mechanism in place, right? Therefore you probably don't want the notification about enabling the watcher displayed - so you'll need to remove "fsWatcherNotification" from unackedNotificationID on upgrade, is that ok?

My strategy is simple: if the user enables filesystem watching in Syncthing, I disable mine. So I'm happy to leave your work here untouched: as users switch to using your filesystem watcher, mine will get out of the way.

In future I'll need some strategy to move users over to yours as I deprecate mine, but that's in the dim and distant future.

imsodin:
By accident I discovered a very strange race: Since enabling the watcher by default in go I get the following weird race in model *only if fswatcher debug facility is activated*:
```
WARNING: DATA RACE
Read at 0x00c42caac7f0 by goroutine 331:
  reflect.typedmemmove()
      /media/ext4_data/Linux/source/go/src/runtime/mbarrier.go:259 +0x0
  reflect.packEface()
      /media/ext4_data/Linux/source/go/src/reflect/value.go:112 +0x10b
  reflect.valueInterface()
      /media/ext4_data/Linux/source/go/src/reflect/value.go:961 +0x170
  reflect.Value.Interface()
      /media/ext4_data/Linux/source/go/src/reflect/value.go:931 +0x51
  fmt.(*pp).printValue()
      /media/ext4_data/Linux/source/go/src/fmt/print.go:692 +0x3994
  fmt.(*pp).printValue()
      /media/ext4_data/Linux/source/go/src/fmt/print.go:846 +0x26da
  fmt.(*pp).printArg()
      /media/ext4_data/Linux/source/go/src/fmt/print.go:682 +0x19e
  fmt.(*pp).doPrintln()
      /media/ext4_data/Linux/source/go/src/fmt/print.go:1136 +0x6e
  fmt.Sprintln()
      /media/ext4_data/Linux/source/go/src/fmt/print.go:264 +0x5f
  github.com/syncthing/syncthing/lib/logger.(*logger).debugln()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/logger/logger.go:114 +0xa8
  github.com/syncthing/syncthing/lib/logger.(*facilityLogger).Debugln()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/logger/logger.go:284 +0xc6
  github.com/syncthing/syncthing/lib/fs.(*logFilesystem).Watch()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/fs/logfs.go:133 +0x515
  github.com/syncthing/syncthing/lib/model.(*folder).startWatchAsync()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/model/folder.go:212 +0x291

Previous write at 0x00c42caac7f0 by goroutine 392:
  github.com/syncthing/syncthing/lib/ignore.(*Matcher).parseLocked()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/ignore/ignore.go:159 +0x1e7
  github.com/syncthing/syncthing/lib/ignore.(*Matcher).Load()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/ignore/ignore.go:137 +0x2b5
  github.com/syncthing/syncthing/lib/model.(*Model).internalScanFolderSubdirs()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/model/model.go:1942 +0x5d5
  github.com/syncthing/syncthing/lib/model.(*folder).scanSubdirs()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/model/folder.go:131 +0xcb
  github.com/syncthing/syncthing/lib/model.(*folder).scanTimerFired()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/model/folder.go:143 +0x6e
  github.com/syncthing/syncthing/lib/model.(*sendReceiveFolder).Serve()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/model/rwfolder.go:201 +0xae6
  github.com/syncthing/syncthing/vendor/github.com/thejerf/suture.(*Supervisor).runService.func1()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/vendor/github.com/thejerf/suture/supervisor.go:525 +0x70

Goroutine 331 (running) created at:
  github.com/syncthing/syncthing/lib/model.(*folder).startWatch()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/model/folder.go:202 +0x2c1
  github.com/syncthing/syncthing/lib/model.(*sendReceiveFolder).Serve()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/model/rwfolder.go:156 +0xebb
  github.com/syncthing/syncthing/vendor/github.com/thejerf/suture.(*Supervisor).runService.func1()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/vendor/github.com/thejerf/suture/supervisor.go:525 +0x70

Goroutine 392 (running) created at:
  github.com/syncthing/syncthing/vendor/github.com/thejerf/suture.(*Supervisor).runService()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/vendor/github.com/thejerf/suture/supervisor.go:515 +0x68
  github.com/syncthing/syncthing/vendor/github.com/thejerf/suture.(*Supervisor).Serve()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/vendor/github.com/thejerf/suture/supervisor.go:413 +0xda6
==================
```
This seems completely bogus - any ideas as to why and/or how to get rid of it?

calmh:
Not bogus, but not super serious. The logfs is doing bad stuff here as it’s passing the *ignore.Matcher directly to Debugln(), which traverses the whole ignore matcher struct using reflect without holding any locks, which is racy. That log line in logfs.go#133 should be fixed.

gui/default/assets/lang/lang-en.json
```diff
@@ -49,6 +50,7 @@
    "Connection Error": "Connection Error",
    "Connection Type": "Connection Type",
    "Connections": "Connections",
+   "Contiuously watching for changes is now available within Syncthing. This will detect changes on disk and issue a scan on only the modified paths. The benefits are that changes are propagated quicker and that less full scans are required.": "Contiuously watching for changes is now available within Syncthing. This will detect changes on disk and issue a scan on only the modified paths. The benefits are that changes are propagated quicker and that less full scans are required.",
```
calmh: Spelling, Continuously. And I think s/less/fewer/

calmh:
Any chance gets to the rescan interval in th folder editor dialog when checking/unchecking the watcher? I think we discussed setting it to 60/3600 in response to the checkbox, and letting the user edit thereafter?

imsodin:
@calmh The interval is now set to 60/3600 on toggling the watcher checkbox.  
@AudriusButkevicius Ping regarding your review.

I think it would make sense to have this in the next release, now that the watcher status is visible in the "main folder panes". Any showstoppers on this?

calmh:
I'm actually going to go negatory on that and say that .46 will do well as just a bugfix release. We can introduce new regressions in .47 instead, including the fallout from turning on fswatcher by default.

AudriusButkevicius:
I'm happy with this.

It seems @calmh is too, if not, we can always revert.

PR #4798, opened 2018-03-10
cmd/syncthing, lib/db: Be nicer about dropping deltas on upgrade

calmh:
### Purpose

When dropping delta index IDs due to upgrade, only drop our local one.
Previously, when dropping all of them, we would trigger a full send in
both directions on first connect after upgrade. Then the other side
would upgrade, doing the same thing. Net effect is full index data gets
sent twice in both directions.

With this change we just drop our local ID, meaning we will send our
full index on first connect after upgrade. When the other side upgrades,
they will do the same. This is a bit less cruel.

### Testing

unit

### DIFF
--- a/lib/db/leveldb_test.go
+++ b/lib/db/leveldb_test.go
@@ -9,6 +9,8 @@ package db
 import (
 	"bytes"
 	"testing"
+
+	protocol "github.com/syncthing/syncthing/lib/protocol"
 )
 
 func TestDeviceKey(t *testing.T) {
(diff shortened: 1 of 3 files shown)


lib/db/leveldb_test.go
```diff
@@ -9,6 +9,8 @@ package db
 import (
 	"bytes"
 	"testing"
+
+	protocol "github.com/syncthing/syncthing/lib/protocol"
```
imsodin: What's the "protocol" specifier for?

calmh:
Good question, I don’t know why that was added. Goimports does that when the package name differs from the last path component, but that’s not the case here. I’ll remove it.

PR #4800, opened 2018-03-10
 lib/scanner, lib/fs: Don't create file infos with abs paths (fixes #4799)

### LINKED ISSUE #4799: Scan on absolute path creates incorrect absolute duplicate file infos
### Reproducer

Run scan through rest API on a "virtually" absolute path that has no changed children, e.g. in my case folder root at `/home/simon/test-sync` and
```
curl -X POST -H "X-API-KEY: myapikey" "localhost:8385/rest/db/scan?folder=3vbey-elvwr&sub=/foo"
```

Result:  
State counter increases and scanner debug log shows the problem:
```
[BXAEQ] 2018/03/10 11:09:21.303458 walk.go:101: DEBUG: Walk [/foo] 131072 Matcher/[nonreach* **/nonreach* nonreach*/** **/nonreach*/**]@0xc4202522d0
[BXAEQ] 2018/03/10 11:09:21.303837 walk.go:316: DEBUG: to hash: /foo File{Name:"/foo", Sequence:0, Permissions:0664, ModTime:2017-12-27 23:10:28.912229075 +0100 CET, Version:{[{BXAEQNX 1}]}, Length:741, Deleted:false, Invalid:false, NoPermissions:true, Blocks:[]}
[BXAEQ] 2018/03/10 11:09:21.303876 walk.go:186: DEBUG: real to hash: /foo
[BXAEQ] 2018/03/10 11:09:21.304035 walk.go:164: DEBUG: Walk progress done 3vbey-elvwr [/foo] 131072 Matcher/[nonreach* **/nonreach* nonreach*/** **/nonreach*/**]@0xc4202522d0
```

Expected:  
No change at all.

### Proposed fix
  
According to [this comment](https://github.com/syncthing/syncthing/blob/master/lib/fs/basicfs.go#L108) the absolute path is valid, so it shouldn't throw an error, but scan normally removing the leading path separator.

imsodin:
This moves part of `(*BasicFilesystem) rooted` to `canonicalize` which is then also used in `walkFilesystem` on the input argument.

Added a unit test on the scanner package walking an absolute path.


lib/fs/walkfs.go
```diff
@@ -38,7 +38,13 @@ func NewWalkFilesystem(next Filesystem) Filesystem {
 
 // walk recursively descends path, calling walkFn.
 func (f *walkFilesystem) walk(path string, info FileInfo, walkFn WalkFunc) error {
-	err := walkFn(path, info, nil)
+	var err error
+	path, err = canonicalize(path)
```
calmh: You can just `path, err :=` here (and in the other use of canonicalize above) but I don't particularly care either way.

imsodin:
@calmh Did that and added a "database cleanup" checking for files starting with a path separator to remove any cruft we accumulated. If any such files are found, the folder metadata is dropped as well.

lib/config/config.go
```diff
@@ -374,6 +377,11 @@ func (cfg *Configuration) clean() error {
 	return nil
 }
 
+func convertV27V28(cfg *Configuration) {
```
calmh: I have a counter suggestion. There is now a generic metadata kv-store area in the database, look in main for where it checks for the upgrade detection. We could have an integer there which is the database schema version or something and avoid doing this based on the config version. Pro: it avoids the "hack" on what is not really a config migration. Con: it's one more magic integer that causes stuff to happen. Discuss.

imsodin: I am in favor of one (hopefully) more magic integer with some semantic connection to the stuff it causes to happen instead of one number used for anything.

lib/fs/basicfs.go
```diff
@@ -88,38 +88,22 @@ func (f *BasicFilesystem) rooted(rel string) (string, error) {
 		expectedPrefix += pathSep
 	}
 
-	// The relative path should be clean from internal dotdots and similar
-	// funkyness.
-	rel = filepath.FromSlash(rel)
-	if filepath.Clean(rel) != rel {
```
imsodin: @AudriusButkevicius  Is there a specific reason to be so strict about path arguments being "clean" in the fs package?  
I implemented `fs.Canonicalize` as the "central path-cleanup-function" e.g. to be used for `model.unifySubs` i.e. on external, potentially non-clean paths. Therefore I removed this check that `filepath.Clean` must have no effect.

calmh: Without this check, what happens with paths like `foo/../../../bar`?

imsodin: Becomes `../../bar` and gets rejected, tested here
https://github.com/syncthing/syncthing/pull/4800/files#diff-fdedb2fe92a10afee3525356a2e710aeR75
and here
https://github.com/syncthing/syncthing/pull/4800/files#diff-aa4a1cf97526cef7590736f637040b8fR385

calmh: So are you saying the check is a no-op, and we might as well keep it?

imsodin: No, I am saying that if we keep the check, then something like `fs.Canonicalize(./foo)` or `fs.Lstat("./foo")` will return an error. What I want and already do in the new `Canonicalize` function is to replace the check with just `file = filepath.Clean(file)`, i.e. for the example `fs.Canonicalize(./foo) == "foo"`.

calmh: Right I might not have read well enough. As long as the clean happens I don’t think it’s critical that it was already clean when entering the function.

lib/db/leveldb_dbinstance.go
```diff
@@ -83,6 +84,31 @@ func newDBInstance(db *leveldb.DB, location string) *Instance {
 	return i
 }
 
+// Update does transitions to the current db version if necessary
+func (db *Instance) Update(cfgVersion int) {
```
calmh: I'd like something a little bit more descriptive here as Update can mean many things for a database. `UpdateSchema`? `UpdateFormat`?

cmd/syncthing/main.go
```diff
@@ -770,21 +770,8 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
...
-		for folderID, folderCfg := range folders {
-			changed += ldb.AddInvalidToGlobal([]byte(folderID), protocol.LocalDeviceID[:])
-			for _, deviceCfg := range folderCfg.Devices {
-				changed += ldb.AddInvalidToGlobal([]byte(folderID), deviceCfg.DeviceID[:])
-			}
-		}
-		l.Infof("Database update: Added %d ignored files to the global list", changed)
-	}
+	// Potential database transitions
+	ldb.Update(cfg.RawCopy().OriginalVersion)
```
calmh: I'd just have a `const databaseSchemaVersion = 1` at the top of main.go and use that here. We can apply all the old transformations (again) when going 0 -> 1 and then bump it as appropriate in the future?

lib/db/leveldb_dbinstance.go
```diff
@@ -648,6 +677,46 @@ func (db *Instance) AddInvalidToGlobal(folder, device []byte) int {
...
+
+	dbi := t.NewIterator(util.BytesPrefix([]byte{KeyTypeDevice}), nil)
+	defer dbi.Release()
+
+	found := make(map[string]struct{})
+
+	for dbi.Next() {
+		folder := db.deviceKeyFolder(dbi.Key())
+		device := db.deviceKeyDevice(dbi.Key())
+		var f FileInfoTruncated
```
calmh: I think you can just use `db.deviceKeyName(dbi.Key())` instead of unmarshalling at all here.

lib/fs/basicfs_test.go
```diff
@@ -365,16 +365,16 @@ func TestRooted(t *testing.T) {
 		{"baz/foo/", "/bar/baz", "baz/foo/bar/baz", true},
 
 		// Not escape attempts, but oddly formatted relative paths. Disallowed.
```
calmh: Comment fail, these are apparently no longer disallowed

lib/fs/basicfs_test.go
```diff
@@ -365,16 +365,16 @@ func TestRooted(t *testing.T) {
...
-		{"baz/foo", "./bar/baz", "", false},
-		{"baz/foo", "bar/../baz", "", false},
-		{"baz/foo", "/bar/../baz", "", false},
-		{"baz/foo", "./bar/../baz", "", false},
-		{"baz/foo", "bar/../baz", "", false},
-		{"baz/foo", "/bar/../baz", "", false},
-		{"baz/foo", "./bar/../baz", "", false},
+		{"foo", "", "foo/", true},
+		{"foo", "/", "foo/", true},
+		{"foo", "/..", "foo/", true},
```
calmh: This is oddness, "foo" + "/.." is an escape attempt, and should be disallowed,

lib/fs/basicfs_test.go
```diff
@@ -365,16 +365,16 @@ func TestRooted(t *testing.T) {
...
-		{"baz/foo", "./bar", "", false},
-		{"foo", "./bar/baz", "", false},
-		{"baz/foo", "./bar/baz", "", false},
-		{"baz/foo", "bar/../baz", "", false},
-		{"baz/foo", "/bar/../baz", "", false},
-		{"baz/foo", "./bar/../baz", "", false},
-		{"baz/foo", "bar/../baz", "", false},
-		{"baz/foo", "/bar/../baz", "", false},
-		{"baz/foo", "./bar/../baz", "", false},
+		{"foo", "", "foo/", true},
```
calmh: Canonicalize now returns stuff ending in slashes?

lib/fs/basicfs_test.go
```diff
@@ -426,12 +423,12 @@ func TestRooted(t *testing.T) {
 			{`\\?\c:\`, `foo`, `\\?\c:\foo`, true},
 			{`c:\`, `\foo`, `c:\foo`, true},
 			{`\\?\c:\`, `\foo`, `\\?\c:\foo`, true},
-			{`c:\`, `\\foo`, ``, false},
-			{`c:\`, ``, ``, false},
-			{`c:\`, `\`, ``, false},
-			{`\\?\c:\`, `\\foo`, ``, false},
-			{`\\?\c:\`, ``, ``, false},
-			{`\\?\c:\`, `\`, ``, false},
+			{`c:\`, `\\foo`, `c:\foo`, true},
```
calmh: This should not be allowed I think, it should fall under the double Windows pathsep rule (UNC path) and be dropped as it were?

lib/fs/filesystem.go
```diff
@@ -198,3 +198,39 @@ func IsInternal(file string) bool {
 	}
 	return false
 }
+
+// Canonicalize checks that the file path is valid and in "canonical" form:
```
calmh: It does not check this, possibly it *makes* it so

lib/fs/basicfs_test.go
imsodin: `rooted` does and already did before this PR for the root path of the fs (and only the root). That's a side effect of needing to fulfill the prefix test (`if !strings.HasPrefix(joined, expectedPrefix)`). What I changed is that `"/"` and `""` are accepted as the same as `"."`. Paths like `/foo` were already allowed, so it seems just consistent to allow this as well (or disallow "/foo").

lib/fs/basicfs_test.go
imsodin: According to how `filepath.Clean` works this is not an escape attempt:
```
4. Eliminate .. elements that begin a rooted path:
   that is, replace "/.." by "/" at the beginning of a path,
   assuming Separator is '/'.
```

cmd/syncthing/main.go
imsodin: Makes sense, I'll do that. There's just one thing: The global invalid transition isn't idempotent. At the time there were no global invalid entries, so the transition doesn't check whether there is already an entry present, it just inserts - that was a bad (fair-weather) idea... So i'll have to fix that too.

imsodin:
@calmh 

> I'm still suspicious. It changes a bunch of returns of Rooted from failed to success, in some cases with suspicious results imho. This is essential to our security and I don't get a warm fuzzy feeling about changing it

I introduced two changes which caused all these fail->success: Non-clean (as in `filepath.Clean`) paths are now allowed as arguments and `"/"` & `""` are now equivalent to `"."` instead of being disallowed. The former is for convenience and the latter for consistency (`"/foo"` was ok and treated like `"foo"`).

For all the "old" database transitions which now are redone, I collected them in one single database iteration for efficiency. The global invalid transition assumed no invalid files in global present which isn't the case anymore. However I don't see any reason to not just use `updateGlobal` instead of adapted (duplicated) code like before (yes I know, I wrote that code...).

lib/fs/filesystem_test.go
```diff
@@ -41,3 +41,68 @@ func TestIsInternal(t *testing.T) {
...
+
+func TestCanonicalize(t *testing.T) {
+	type testcase struct {
+		path     string
+		expected string
+		ok       bool
+	}
+	cases := []testcase{
+		// Valid cases
+		{"/bar", "bar", true},
```
calmh: You have the same test case a bunch of times in this table, here and below. I guess it comes from search/replace on another table? Just remove duplicates and it's cool.

lib/db/leveldb_dbinstance.go
```diff
@@ -525,21 +540,39 @@ func (db *Instance) checkGlobals(folder []byte, meta *metadataTracker) {
...
+	ignAdded := 0
+	meta := newMetadataTracker() // dummy metadata tracker
+
 	for dbi.Next() {
+		folder := db.deviceKeyFolder(dbi.Key())
+		device := db.deviceKeyDevice(dbi.Key())
+		name := string(db.deviceKeyName(dbi.Key()))
+
+		// Remove files with absolute path (see #4799)
+		if strings.HasPrefix(name, string(fs.PathSeparator)) {
```
calmh: Pretty sure we have unix style paths in the database regardless of OS, so we should look for just `'/'` here and not `fs.PathSeparator`.

PR #4804, opened 2018-03-12
lib/fs, lib/model: Make tests caching compatible (fixes #4749)

### LINKED ISSUE #4749: Make all tests Go 1.10 caching compatible
Go 1.10 uses test caching, where tests are not rerun unless the relevant code or environment changes. This is awesome. Many of our tests already get cached, but many others do not. Mostly this is because we do filesystem operations and the end state is not identical with the start state, according to the test cacher. Most cases can probably be fixed by just properly cleaning up after ourselves.

The packages that don't cached at the moment are `lib/fs`, `lib/model`, and `cmd/syncthing`.

imsodin:
The problems preventing caching to work were not removed temporary files, temporary files placed outside of  `testdata` and files modified during tests. I also used temp names for wrapper files - caching doesn't actually care about that, but I assumed it would and it is cleaner anyway.


calmh:
omg ❤️ ❤️

calmh:
Small windows fail otherwise awesome

imsodin:
@calmh When removing the line 
```
defaultConfig = config.Wrap("/tmp/test", _defaultConfig)
```
in `init()` I was wondering why that ever worked on windows. Is there something like `/tmp` on Windows?

calmh:
There is a c:\tmp on the build server. I’m not sure if created by this or from something else.

imsodin:
I now put these files into `testdata/tmp` on windows, as I don't really want to create a "root level" directory just for testing.

imsodin:
@calmh The integration test `TestConflictsDefault` seems to be flaky, it switched between fail and success a few times without a change already.

calmh:
Yeah, I noticed, something is bad there. Haven't been exercising the integration tests that much lately, just enabled them recently. (They don't block merges either at the moment.)

imsodin:
Ups, didn't notice that the merge frame wasn't red, just looked at the red cross. I think it's great to have those, they helped me discover quite a few pitfalls when doing the fs/db scan PR (even though they didn't find the capital issue :P ). So thanks a lot for reviving them (a while ago already) and integrating them into the CI!

PR #4811, opened 2018-03-13
lib/ignore, lib/scanner: Catch included items below ignored ones

imsodin:
### Purpose

Including files with the ignore patterns is really complicated, as due to skipping ignored parent directories the at least conceptually simple "first rule wins" policy doesn't hold anymore. Even worse, this is inconsistent between scanning and pulling: An included filed below an ignored directory is not scanned but it is pulled, e.g. with
```
!foo
*
```
changes to `subdir/foo` are not picked up, but remote changes to `subdir/foo` are synced.

This proposal changes the way files are traversed while scanning: If there are any include patterns, ignored directories aren't skipped anymore (but obviously still ignored).

Pro:
 - Simpler ignore rules
Con:
 - Potentially a lot more data to traverse/stat.

Having the fs watcher mitigates the con somewhat.

### Testing

One new test in the scanner package that 

### Screenshots

If this is a GUI change, include screenshots of the change. If not, please
feel free to just delete this section.

### Documentation

Examples need to be updated, rules don't mention the parent dir problem.


lib/scanner/walk_test.go
```diff
@@ -513,12 +504,39 @@ func TestIssue4799(t *testing.T) {
...
-	files, err := walkDir(fs, "/foo")
+	files := walkDir(fs, "/foo")
+	if len(files) != 1 || files[0].Name != "foo" {
+		t.Error(`Received unexpected file infos when walking "/foo"`, files)
+	}
+}
+
+func TestRecurseUnignore(t *testing.T) {
+	stignore := `
+	!/dir1/cfile
```
calmh: A test with the non rooted file pattern would be good too

imsodin:
> Are there ignore patterns that are valid and working today that will mean something different when this goes in?

It's possible to use this particular behaviour for something like e.g. this
```
!includeInSomeDirs
!/dir1/
/dir1/*
!/dir2/
/dir2/*
*
```
but I don't remember ever seeing it and it feels contrived - but given that it is possible, chances are someone does use it.

lib/fs/basicfs_watch_test.go
```diff
@@ -117,14 +143,25 @@ func TestWatchOutside(t *testing.T) {
...
 			}
 			cancel()
 		}()
 		fs.watchLoop(".", testDirAbs, backendChan, outChan, fakeMatcher{}, ctx)
 	}()
 
 	backendChan <- fakeEventInfo(filepath.Join(filepath.Dir(testDirAbs), "outside"))
+
+	select {
+	case <-time.NewTimer(10 * time.Second).C:
```
AudriusButkevicius: `time.After`

lib/scanner/walk.go
```diff
@@ -277,6 +244,67 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
+			l.Debugln("ignored (patterns):", path)
+			if !w.Matcher.SkipIgnoredDirs() {
+				return nil
+			}
+			return skip
+		}
+		if ignoredParent != "" {
+			// Add parent directories of the current, not ignored path if needed
+			if rel := strings.TrimPrefix(path, ignoredParent+string(fs.PathSeparator)); rel != path {
+				for _, name := range strings.Split(filepath.Base(rel), string(fs.PathSeparator)) {
```
AudriusButkevicius: There is filepath.Parts or something like that.

lib/scanner/walk.go
```diff
@@ -277,6 +244,67 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
+					if err = handleItem(ignoredParent); err != nil {
+						return err
+					}
+					ignoredParent = filepath.Join(ignoredParent, name)
+				}
+			}
+			ignoredParent = ""
+		}
+
+		return handleItem(path)
```
AudriusButkevicius: Some comment what this whole new change is doing would be nice.
For now it's a bit of black magic, given the nested functional dynamic programming here,

lib/ignore/ignore.go
```diff
@@ -282,6 +292,10 @@ func (m *Matcher) ShouldIgnore(filename string) bool {
 	return false
 }
 
+func (m *Matcher) SkipIgnoredDirs() bool {
```
AudriusButkevicius: I guess this is a nice optimization to have, yet I'd drop it in favor of having less code.

@calmh ?

imsodin: It's more than just optimization: E.g. the fs watcher (well notify) fails entirely if it encounters files it can't watch due to permissions, which can be worked around with ignores. I know that's far from optimal, but it is how it is. And I depend on this workaround :)

lib/scanner/walk.go
imsodin: The only ones I know are `filepath.Split`, but it only splits the base/parent dir from the filename and `filepath.SplitList`, which is for "patha:pathb:pathc" kind of lists.

AudriusButkevicius: Wow, I am amazed there isn't anything.

lib/scanner/walk.go
```diff
@@ -277,6 +244,67 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
+			return skip
+		}
+		if ignoredParent != "" {
+			// Add parent directories of the current, not ignored path if needed
+			if rel := strings.TrimPrefix(path, ignoredParent+string(fs.PathSeparator)); rel != path {
+				for _, name := range strings.Split(filepath.Base(rel), string(fs.PathSeparator)) {
+					if err = handleItem(ignoredParent); err != nil {
+						return err
+					}
+					ignoredParent = filepath.Join(ignoredParent, name)
```
AudriusButkevicius: Like, how does this work. 

If I am inside a parent that is ignored (`a/b/c`), trim the prefix and get relative path to ignored parent (if I am `a/b/c/d` I get d), yet `path` is always the full path (`a/b/c/d`), so why are we comparing these?

Then we do some split on the filepath.Base, filepath base returns the last element after the last path separator, so what's the point of the split?

The we call ourselves with the parent?
Why can't this be handled by the walk itself, why do we have to help walk?

I am reading this but this makes no sense to me.

imsodin: I can totally relate to your reaction, now that I reread the code with a little distance... I shuffled the code around and added comments. Hopefully together with the following explanation of "why do it in the first place" it is more clear now.

We need to help walk in the following scenario: Unignored path below an ignored path, e.g. `foo/bar/baz` with
```
!baz
/foo
```
Going towards `foo/bar/baz`, walk tells us to handle `foo` and `foo/bar`, but we don't do anything with them, as they are ignored. Then we come upon `foo/bar/baz` and realize, that we need to add it to the database. However we also need to add the parent directories, as we ignored them before.

lib/scanner/walk.go
```diff
@@ -242,40 +274,50 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
+			return handleItem(path)
 		}
 
-		return err
+		// The previously ignored parent directories of the current, not
+		// ignored path need to be handled as well.
+		if err = handleItem(ignoredParent); err != nil {
+			return err
+		}
+		for _, name := range strings.Split(rel, string(fs.PathSeparator)) {
```
AudriusButkevicius: Why did the filepath.Base disappear?

imsodin: Because you didn't like it :)  
Before I used it to excluded the current path and handle it at the end of the function. Now I  handle it in this loop and at the end of the function there is just a `return nil` as no error occurred before.

lib/scanner/walk.go
```diff
@@ -242,40 +274,50 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
+			ignoredParent = filepath.Join(ignoredParent, name)
+			if err = handleItem(ignoredParent); err != nil {
+				return err
+			}
+		}
+
+		return nil
```
AudriusButkevicius: This is now clear to me.

lib/scanner/walk.go
```diff
@@ -242,40 +274,50 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
...
+		if w.Matcher.Match(path).IsIgnored() {
+			l.Debugln("ignored (patterns):", path)
+			if w.Matcher.SkipIgnoredDirs() {
+				return skip
 			}
-			if info.IsDir() {
-				// under no circumstances shall we descend into a symlink
-				return fs.SkipDir
+			// If the parent wasn't ignored already, set this path as the "highest" ignored parent
+			if ignoredParent == "" || !strings.HasPrefix(path, ignoredParent+string(fs.PathSeparator)) {
```
AudriusButkevicius: Shouldn't we only set this if path is a directory?

lib/scanner/walk.go
imsodin: I indeed lost the line that prevents this when simplifying the code: After hitting `e`, `ignoredParent` must be reset of course (because the parent isn't ignored anymore).

lib/scanner/walk.go
imsodin: No, the "upmost" parent is kept, because at this point we don't yet know whether there is an unignored child. And if there is, we need to handle all parents.

lib/scanner/walk.go
imsodin: Actually I didn't read your example carefully enough, because I jumped at my mistake: `e` and all the others are not ignored or included, so why should they trigger handling?

lib/scanner/walk.go
AudriusButkevicius: You could do it the other way around, which would make it less magical.

Keep building up ignoredParent up to the unignored item, then once you do hit an unignored item, walk ignoredParent backwards up to the point where ignoredParent is not actually ignored.

Also the directory check thing still prevails.

lib/scanner/walk.go
AudriusButkevicius: Well ignoring works by patterns, and my gut says that `/a/b/c/d` expands to something like `/a/b/c/d/**` pattern that then matches everything under that.

Your tests don't handle anything deeper than 1 level it seems.

AudriusButkevicius:
Also, I'm thinking if there is also a way to move all of this cruft to the ignore package.

Ignore package could return ShouldIgnore(pat) IgnoreResult, and IgnoreResult could be an enum:

1. Ignore
2. IgnoreTransient (child of ignored)
3. NotIgnore

lib/scanner/walk.go
imsodin: This doesn't work, we need to keep track of the upmost/shortest ignored directory. Consider the patterns
```
!not
/foo
```
and files `foo/aaa/ignored` and `foo/not`. `foo` is ignored and becomes `ignoredParent`, then `foo/aaa` becomes `ignoredParent`. `foo/aaa/ignored` is just ignored, nothing else happens. Then `foo/not` is not ignored, but `ignoredParents` is `foo/aaa` so `foo` is not handled.

lib/scanner/walk.go
imsodin: I extended the tests to cover multiple levels.

AudriusButkevicius:
@imsodin I think you think this is ok now?

imsodin:
@AudriusButkevicius If you're convinced by my replies to your points then yes - except for the test failure, which is fixed in #4846, as it's totally unrelated to the ignoring issue solved here.

AudriusButkevicius:
I'll try to propose an alternative version.

AudriusButkevicius:
```diff
diff --git a/lib/scanner/walk.go b/lib/scanner/walk.go
index e5c9231..c390d4b 100644
--- a/lib/scanner/walk.go
+++ b/lib/scanner/walk.go
@@ -285,40 +285,32 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
                                return skip
                        }
                        // If the parent wasn't ignored already, set this path as the "highest" ignored parent
-                       if info.IsDir() && (ignoredParent == "" || !strings.HasPrefix(path, ignoredParent+string(fs.PathSeparator))) {
+                       if info.IsDir() {
                                ignoredParent = path
                        }
                        return nil
                }

-               if ignoredParent == "" {
-                       // parent isn't ignored, nothing special
-                       return handleItem(path)
-               }
-
-               // Part of current path below the ignored (potential) parent
-               rel := strings.TrimPrefix(path, ignoredParent+string(fs.PathSeparator))
-
-               // ignored path isn't actually a parent of the current path
-               if rel == path {
+               if ignoredParent == "" || filepath.Dir(path) != ignoredParent {
+                       // parent isn't ignored, or we're out of the ignored parent, nothing special
                        ignoredParent = ""
                        return handleItem(path)
                }

-               // The previously ignored parent directories of the current, not
-               // ignored path need to be handled as well.
-               if err = handleItem(ignoredParent); err != nil {
-                       return err
-               }
-               for _, name := range strings.Split(rel, string(fs.PathSeparator)) {
-                       ignoredParent = filepath.Join(ignoredParent, name)
-                       if err = handleItem(ignoredParent); err != nil {
-                               return err
+               // For each part upto this path, if it's ignored, handle it (unignore it)
+               soFar := ""
+               parts := strings.Split(path, string(fs.PathSeparator))
+               parts = parts[:len(parts)-1]
+               for _, part := range parts {
+                       soFar = filepath.Join(soFar, part)
+                       if w.Matcher.Match(soFar).IsIgnored() {
+                               if err = handleItem(soFar); err != nil {
+                                       return err
+                               }
                        }
                }
-               ignoredParent = ""
-
-               return nil
+               // Handle the actual item
+               return handleItem(path)
        }
 }
```

AudriusButkevicius:
All of these tho have the weakness of calling handleItem for ignored parents multiple times.

imsodin:
> All of these tho have the weakness of calling handleItem for ignored parents multiple times.

Exactly?!  
That might be harmless in terms of the db/indexes, but at least duplicate events will be generated. And I wouldn't want to count on that when there is a way to prevent it in the first place.

AudriusButkevicius:
And what's the way to prevent that? Because I don't think your code prevents that.

AudriusButkevicius:
If you have

/a/b/c ignored
/a/b/c/d/e/f unignored
/a/b/c/d/e/g unignored

I suspect your code will still unginore d twice, d/e twice.

imsodin:
I don't think it will: `ignoredParent` is "/a/b/c". Then the unignored "/a/b/c/d/e/f" is encountered and it's parents are handled and `ignoredParent = ""`. Then when encountering "/a/b/c/d/e/g" `ignoredParent` is empty, no parents are handled. Even if there are other files in between that are ignored, they are at the same or higher "level" (path components length). The test covers a similar scenario: https://github.com/syncthing/syncthing/pull/4811/files#diff-35882400d2fb360084c22317e3766b16R579

AudriusButkevicius:
So my version passes your test case.

imsodin:
@AudriusButkevicius And now?

@calmh There's a committer email address problem again. I guess @pramodhkp hasn't set a git email address, so github adds a default one?  
```
authors_test.go:47: Missing author: 1507241+pramodhkp@users.noreply.github.com
```

AudriusButkevicius:
Ok, now mine fails.

calmh:
OK, this thing with GitHub changing the author email at merge time sucks. The commit had a perfectly cromulent email before merging...

lib/scanner/walk.go
```diff
@@ -197,7 +199,43 @@ func (w *walker) walk(ctx context.Context) chan protocol.FileInfo {
 }
 
 func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protocol.FileInfo) fs.WalkFunc {
+	var skip error // nil
 	now := time.Now()
+	ignoredParent := ""
+
+	handleItem := func(path string) error {
```
calmh: Can this be a method instead of a closure? I see it uses skip, but this could be handled by the caller or be a parameter. It would reduce the mental stack depth to understand what's going on here

lib/scanner/walk.go
calmh: Yeah so I'm also a bit skeptical here. Consider, patterns:

```
!ignore/d1/d2/foo
ignore
```

when we reach `ignore` it will get ignored and set as the parent, remembered. Then we continue walking and at some point reach the unignored `ignore/d1/d2/foo`. We'll start the loop to look at parents, `ignore/d1` to begin with. This is a directory, so it will hit the `walkDir` in the closure above and start walking recursively, at some point reaching `ignore/d1/d2/foo` again, ... ?

lib/scanner/walk.go
imsodin: It requires lots of parameters - is it worth it?

```diff
diff --git a/lib/scanner/walk.go b/lib/scanner/walk.go
index fccb47b1b..3603ddcd0 100644
--- a/lib/scanner/walk.go
+++ b/lib/scanner/walk.go
@@ -203,39 +203,6 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
 	now := time.Now()
 	ignoredParent := ""
 
-	handleItem := func(path string) error {
-		info, err := w.Filesystem.Lstat(path)
-		// An error here would be weird as we've already gotten to this point, but act on it nonetheless
-		if err != nil {
-			return skip
-		}
-
-		path, shouldSkip := w.normalizePath(path, info)
-		if shouldSkip {
-			return skip
-		}
-
-		switch {
-		case info.IsSymlink():
-			if err := w.walkSymlink(ctx, path, dchan); err != nil {
-				return err
-			}
-			if info.IsDir() {
-				// under no circumstances shall we descend into a symlink
-				return fs.SkipDir
-			}
-			return nil
-
-		case info.IsDir():
-			err = w.walkDir(ctx, path, info, dchan)
-
-		case info.IsRegular():
-			err = w.walkRegular(ctx, path, info, fchan)
-		}
-
-		return err
-	}
-
 	return func(path string, info fs.FileInfo, err error) error {
 		select {
 		case <-ctx.Done():
@@ -294,7 +261,7 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
 
 		if ignoredParent == "" {
 			// parent isn't ignored, nothing special
-			return handleItem(path)
+			return w.handleItem(path, ctx, fchan, dchan, skip)
 		}
 
 		// Part of current path below the ignored (potential) parent
@@ -303,17 +270,17 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
 		// ignored path isn't actually a parent of the current path
 		if rel == path {
 			ignoredParent = ""
-			return handleItem(path)
+			return w.handleItem(path, ctx, fchan, dchan, skip)
 		}
 
 		// The previously ignored parent directories of the current, not
 		// ignored path need to be handled as well.
-		if err = handleItem(ignoredParent); err != nil {
+		if err = w.handleItem(ignoredParent, ctx, fchan, dchan, skip); err != nil {
 			return err
 		}
 		for _, name := range strings.Split(rel, string(fs.PathSeparator)) {
 			ignoredParent = filepath.Join(ignoredParent, name)
-			if err = handleItem(ignoredParent); err != nil {
+			if err = w.handleItem(ignoredParent, ctx, fchan, dchan, skip); err != nil {
 				return err
 			}
 		}
@@ -323,6 +290,39 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
 	}
 }
 
+func (w *walker) handleItem(path string, ctx context.Context, fchan, dchan chan protocol.FileInfo, skip error) error {
+	info, err := w.Filesystem.Lstat(path)
+	// An error here would be weird as we've already gotten to this point, but act on it nonetheless
+	if err != nil {
+		return skip
+	}
+
+	path, shouldSkip := w.normalizePath(path, info)
+	if shouldSkip {
+		return skip
+	}
+
+	switch {
+	case info.IsSymlink():
+		if err := w.walkSymlink(ctx, path, dchan); err != nil {
+			return err
+		}
+		if info.IsDir() {
+			// under no circumstances shall we descend into a symlink
+			return fs.SkipDir
+		}
+		return nil
+
+	case info.IsDir():
+		err = w.walkDir(ctx, path, info, dchan)
+
+	case info.IsRegular():
+		err = w.walkRegular(ctx, path, info, fchan)
+	}
+
+	return err
+}
+
 func (w *walker) walkRegular(ctx context.Context, relPath string, info fs.FileInfo, fchan chan protocol.FileInfo) error {
 	curMode := uint32(info.Mode())
 	if runtime.GOOS == "windows" && osutil.IsWindowsExecutable(relPath) {
```

imsodin:
I added an earlier `info.IsSymlink() && info.IsDir()` check for ignored items - this does not mean that it was possible to recurse into symlinks before (I checked). Walking the filesystem and `handleItem` both already have precautions against that. The change just aborts a little earlier to prevent pointless code execution: https://github.com/syncthing/syncthing/pull/4811/commits/601f2f791b7cf0eb11d8fe3e0ad2c6c24104f22e#diff-7a7f4905b4d4073af61d4719cc9aff0eL284

imsodin:
@AudriusButkevicius Just saw your edit:

> Yet I wonder what happens with:
>
> [ignored section]/[unignored section]/[ignored section]/[unignored section]

Works just fine: https://github.com/syncthing/syncthing/pull/4811/commits/cec756154bbe579291f635fd868d6cf1a06af2c2

imsodin:
@AudriusButkevicius @calmh lgtm?

AudriusButkevicius:
So lgtm, but I'd like a second blessing from @calmh , as this is complicated enough.

imsodin:
Totally forgot this is still open - @calmh's blessing (or not) seems to be missing ;)

lib/scanner/walk.go
```diff
@@ -197,7 +199,10 @@ func (w *walker) walk(ctx context.Context) chan protocol.FileInfo {
 }
 
 func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protocol.FileInfo) fs.WalkFunc {
+	var skip error // nil
```
calmh: Why hoist this? It looks like it gets set to nil anyway at function entry, but it becomes part of the closure and I'm afraid I'm missing some cleverness?

imsodin: No cleverness, just a leftover from changing `handleItem` from closure to method.

PR #4816, opened 2018-03-15
cmd/syncthing: Improve local host check (fixes #4815)

### LINKED ISSUE #4815: Local host header check is too restrictive
When doing the host check, syncthing calls addressIsLocalHost(), which appears to be too restrictive.  There are valid localhost addresses that are failing this test.  For example, 127.5.6.7.  In fact 127.0.0.0/8 is defined as being local to the machine.

My use case is this:

To avoid more and more obscure local port forwarding when I SSH to a remote box (e.g. ssh -L 28384:localhost:8384), I put specific localhost entries in /etc/hosts, such as
    127.5.6.7 myremotehost.lo 
and then I can put this in my ~/.ssh/config:
    Host myremotehost.com
        LocalForward myremotehost.lo:8384 localhost:8384
and then navigate to
    http://myremotehost.lo:8384

This breaks because addressIsLocalHost() doesn't think 127.5.6.7 is a valid local host, and I get the Host Check Error.

I can disable the check, but that weakens security.  It would be better to have the function fixed.

I'm not a Go programmer.  I can see in GitHub where the code is called, and where the test function is.  Perhaps it's a standard library function.  I tried googling and searching the golang API site, but I can't find it.  Sorry I can't be more help.

Syncthing Version: v0.14.32
OS Version: Debian 7
Browser Version: Firefox 58.0.2 (Ubuntu)

calmh:
This does a less restrictive (and more correct) check on the IP address, and also allows the fully qualified "localhost.".

Covered by unit test.


AudriusButkevicius:
Why `localhost.` tho?

calmh:
Because it's a valid way to refer to localhost and we might run into it in the host header and maybe bind address. It's just something I forgot the first time this was added.

PR #4829, opened 2018-03-20
lib/nat: Fix clearAddresses/notify deadlock.

grahammiln:
### Purpose

Fix a deadlock when removing a NAT port mapping.

`clearAddresses` write locks the struct and then calls `notify`. `notify` in turn tries to obtain a read lock on the same mutex. The result was a deadlock. This change unlocks the struct before calling `notify`.

### Testing

I am not sure how to automate the testing of this change.

### Authorship

Graham Miln (grahammiln) <graham.miln@miln.eu>

### DIFF
--- a/lib/nat/structs.go
+++ b/lib/nat/structs.go
@@ -53,11 +53,11 @@ func (m *Mapping) clearAddresses() {
 		removed = append(removed, addr)
 		delete(m.extAddresses, id)
 	}
+	m.expires = time.Time{}
+	m.mut.Unlock()
 	if len(removed) > 0 {
 		m.notify(nil, removed)
 	}
-	m.expires = time.Time{}
-	m.mut.Unlock()
 }
 
 func (m *Mapping) notify(added, removed []Address) {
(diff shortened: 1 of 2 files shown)


grahammiln:
Thank you for the suggested testing approach. I have added a unit test to exercise the previously deadlocking code.

grahammiln:
Is there anything more you need me to do before this can be merged?

AudriusButkevicius:
Can you add yourself to authors?

calmh:
Let me do that.

AudriusButkevicius:
So squash merge with admin privs ignoring the authors check?

calmh:
No. Hold.

calmh:
Actual commit author doesn't match the contact given:

   Missing author: graham.miln@dssw.co.uk

Which is why it's still red, and one reason to not force it. @grahammiln add that address too, or you want to fix the commits?

grahammiln:
I would prefer the author line below, as this work is not related to dssw.co.uk:
Graham Miln (grahammiln) graham.miln@miln.eu

calmh:
I sorted it out for now. Thanks!

PR #4836, opened 2018-03-23
lib/upgrade: Handle version containing spaces

imsodin:
### Purpose

The release name of the android hotfix is `v0.14.45 (+dev)`, resulting in patch version 0 (as conversion to string of `45 (` fails). This is somewhat relevant because of the symlink compatibility check here: https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L922  

### Testing

Test comparing  `v0.14.45 (+dev)` with `v0.14.13`.

### DIFF
--- a/lib/upgrade/upgrade_common.go
+++ b/lib/upgrade/upgrade_common.go
@@ -199,6 +199,7 @@ func versionParts(v string) ([]int, []interface{}) {
 		v = v[1:]
 	}
 	parts := strings.SplitN(v, "+", 2)
+	parts = strings.SplitN(parts[0], " ", 2)
 	parts = strings.SplitN(parts[0], "-", 2)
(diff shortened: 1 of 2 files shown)


calmh:
Our versions are mostly semver, and semver doesn’t have spaces in it.

imsodin:
Sure, but as this example demonstrates we don't have control over what versions are actually used for a sizable part of users. Another approach to "solve" this problem would be to enforce semver in build.go.

imsodin:
This is based on a wrong premise, see https://github.com/syncthing/usage-reporting/pull/15#issuecomment-375711653:
> the F-Droid version is "v0.14.45+1-gbf165d68"

calmh:
Right, that version should be handled perfectly fine. (But maybe not by the usage reporting thing)

Issue #4823, opened 2018-03-18
Filter the global change log UI


xgdgsc:
### Version Information

Syncthing Version: v0.14.45
OS Version: Windows 10
Browser Version: chrome  65.0.3325.162

# Request
I noticed previous discussions on showing recent changes for folder or global like https://github.com/syncthing/syncthing/issues/2924  https://github.com/syncthing/syncthing/pull/3694 . But the current global change log displays a lot of `ignored`  which isn' t useful for me to see what actually happend. How to filter this?

And recent change for a folder would also be very useful, just need to extend the current `last changed file` to a clickable expandable list of files.

Catfriend1:
> forum.syncthing.net please ?

AudriusButkevicius:
What do you mean by ignored?

Last changed works differently, so extending that means a rewrite, you can use the global change list for that.

xgdgsc:
I mean sometimes the full list of global recent changed files would include `ignored` files. I want to filter these out and only see the `modified` `deleted` `added` files.

AudriusButkevicius:
This should not happen, as you sure the files are ignored?

xgdgsc:
They are ignored correctly.

imsodin:
Yes, it does happen - I introduced that when changing the way ignores work, that was probably a mistake: https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L1798  
I mean it can be helpful to see that a change was ignored, but this happens whenever the global version of an ignored file changes. With lots of ignored files, it drowns other stuff out. I think we shouldn't log a disk change event in case of ignored files (but eventually implement an option to display ignored files, not sure whether there is a ticket for that already).

AudriusButkevicius:
Yeah, we should just skip invalid updates, as they don't signify anything meaningful in terms of local state changes.

calmh:
I think this was addressed by #4837 (but that PR tagged the wrong issue)

PR #4837, opened 2018-03-25
lib/model: Dont log ignored files (fixes #4832)

### LINKED ISSUE #4832: sha256 could use faster implementation where available
[github.com/minio/sha256-simd](github.com/minio/sha256-simd) has SIMD instruction using sha256 that is much faster than Go 1.10 builtin:

mstenber@poro ~/go/go-tfhfs/fs>go test -bench '^BenchmarkShaBase256'
goos: darwin
goarch: amd64
pkg: github.com/fingon/go-tfhfs/fs
BenchmarkShaBase256/builtin-50-12         	 3000000	       392 ns/op 127.44 MB/s
BenchmarkShaBase256/builtin-1000-12       	  200000	      6117 ns/op 163.47 MB/s
BenchmarkShaBase256/builtin-20000-12      	   10000	    108678 ns/op 184.03 MB/s

BenchmarkShaBase256/simd-50-12            	 5000000	       295 ns/op 169.17 MB/s
BenchmarkShaBase256/simd-1000-12          	  300000	      3858 ns/op 259.15 MB/s
BenchmarkShaBase256/simd-20000-12         	   20000	     74677 ns/op 267.82 MB/s

builtin = crypto/sha256
simd = the above link.

applicable in all testcases (50 bytes to 20 kb) on my machine.

AudriusButkevicius:


lib/model/model.go
```diff
@@ -1795,8 +1795,8 @@ func (m *Model) diskChangeDetected(folderCfg config.FolderConfiguration, files [
 		case file.IsDeleted():
 			action = "deleted"
 
-		case file.Invalid:
-			action = "ignored" // invalidated seems not very user friendly
+		case file.IsInvalid():
```
imsodin: Should be moved above `IsDeleted`, as a deletion may also be ignored.

AudriusButkevicius:
Addressed.

uok:
this fixes #4823 and not #4832

PR #4839, opened 2018-03-26
Add icon & file info to syncthing.exe

harrisonhjones:
### Purpose

This change fixes/addresses #3668 by adding an Icon to the Windows `syncthing.exe` binary.

Based on https://github.com/syncthing/syncthing/pull/4645

### Testing

- Ran `go run build.go` - success
- Verified the` binary was created at `bin/syncthing.exe` - success
- Verified the binary had the Syncthing icon - success
- Verified the binary had the given file properties by right clicking on it, clicking the "properties" menu item, and then the "details" tab. - success

Note, I tested this build on a Windows machine.

### Screenshots

image

### Documentation

I don't believe this requires a documentation change.

### Authorship

- Name: Harrison Jones (or Harrison H. Jones if middle initials are allowed)
- Email: harrisonhjones@gmail.com and/or harrison@harrisonhjones.com


AudriusButkevicius:
I get when running build
```
exec: "goversioninfo": executable file not found in %PATH%
exit status 1
```

It should not abort if it's not present

build.go
```diff
@@ -621,6 +630,55 @@ func buildSnap(target target) {
...
+	file.Close()
+	defer func() {
+		if err := os.Remove(jsonPath); err != nil {
+			log.Printf("Warning: unable to remove generated %s: %v. Please remove it manually.", jsonPath, err)
+		}
+	}()
+
+	sysoPath := filepath.Join(dir, "cmd", "syncthing", "resource.syso")
+
+	runPrint("goversioninfo", "-o", sysoPath)
```
AudriusButkevicius: Probably want to replace with runError

harrisonhjones: Updated to use `runError` and also had `shouldBuildSyso return an error (could probably just rename to `buildSyso`) if it failed to build the syso.

**Testing**

1. Removed `goversioninfo` from my system and re-ran `go run build.go`.
   - Build succeeded with a logged warning
1. Added `goversioninfo` back to my system and re-ran `go run build.go`
   - Build succeeded without a warning and the `syncthing.exe` was encoded properly with the icon and build information.

AudriusButkevicius:
@st-jenkins rebuild

AudriusButkevicius:
I think the integration test failures are a red herring, hence merged this.

AudriusButkevicius:
We need to: 
```
go get github.com/josephspurrier/goversioninfo
go install github.com/josephspurrier/goversioninfo/cmd/goversioninfo
```

on the build servers, yet I think I no longer have access to those.

AudriusButkevicius:
Also, the icon is too high res (or perhaps we should include a few as I think Windows supports that), but that's a separate issue

AudriusButkevicius:
Seems that even if I remove the layers it ends up pixelated.

harrisonhjones:
I generated the icon with multiple sizes using an .ico plugin for Paint.Net. I'm happy to regenerate it I just don't quite understand the issue

AudriusButkevicius:
It seems that no matter what I do the icon remains pixelated. I suspect it's because of number of colors allowed in the smaller icons.

canton7:
This is the one I use: lots of resolutions, smart scaling, bit of manual tweaking. https://github.com/canton7/SyncTrayzor/blob/master/src/SyncTrayzor/Icons/default.ico

harrisonhjones:
@canton7 it looks a bit clipped at the bottom (in my browser at least) but if it's better than the one I contributed let's us it! (assuming you are ok with that)

AudriusButkevicius:
I'll try yours tomorrow.

calmh:
About the tool, this needs to be added to the build agent docker image we use. I've done so, new image should be rolling out during the afternoon. (https://github.com/kastelo/dockers/tree/master/teamcity-agent)

AudriusButkevicius:
Go get just fetches the code right, I think this needs a go install too?

calmh:
Nah it installs the binary as well

calmh:
And, uh, I’m a dummy because this obviously should happen on the windows builder which has nothing to do with the docker image. I’ll sort it out.

calmh:
Yeah but to be relevant for us it needs to happen on the box where we do windows builds, which is the windows builder

PR #4842, opened 2018-03-27
lib/scanner, lib/model: Actually assign version when un-ignoring (fixes #4841)

### LINKED ISSUE #4841: Un-ignoring a file that doesn't exist locally results in deletion on remote devices
Reported by cosas in the forum: https://forum.syncthing.net/t/adding-a-ignore-dir-item-in-local-device-deletes-dir-on-remote-and-shows-95-0b

Steps to reproduce:  
 1. Create ignore patterns
```
/foo
```
on device A.
2. Create file `foo/bar` on device B and make sure it is scanned and "synced" to A (creating the invalid file info).
3. Change ignore patterns on A to
```
!bar
/foo
```

Result:
`foo/bar` is deleted on B

Expected result:
`foo/bar` is created on A

imsodin:
### Purpose

See #4841. While `func (v Vector) DropOthers(id ShortID) Vector` does modify the underlying slice in place if possible, it doesn't always and is thus not to be used to modify the Vector in place (as are all the `Vector` methods). This resulted in files that should be updated as conflict to override the global state, which in case of non-existing files leads to data loss across the cluster.

@calmh Would you prefer PRs against the release branch for stuff like this or always to master?

### Testing

Unit tests on un-ignoring in the model when no file exists and in the scanner when there is a file. Also tested manually with the reproducer given in the issue.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -2076,7 +2076,7 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 				// other existing versions, which will be resolved
 				// by the normal pulling mechanisms.
 				if f.IsInvalid() {
-					nf.Version.DropOthers(m.shortID)
+					nf.Version = nf.Version.DropOthers(m.shortID)
 				}
 
 				batch = append(batch, nf)
--- a/lib/model/requests_test.go
+++ b/lib/model/requests_test.go
@@ -425,6 +425,45 @@ func pullInvalidIgnored(t *testing.T, ft config.FolderType) {
 	}
 }
 
+func TestIssue4841(t *testing.T) {
+	m, fc, tmpDir := setupModelWithConnection()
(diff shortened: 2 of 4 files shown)


imsodin:
@calmh The test failures are due to new files in `lib/scanner/testdata` introduced in #4811, but that PR isn't merged yet. Somehow it still manages to pollute the working directory of this build. Sounds like black magic...

calmh:
uh

calmh:
Fault appears to lie with the private gopath building thing we have, which is supposed to clean out files that aren't present in the original directory, but apparently doesn't? I manually wiped the temporary gopath. Once we get vgo and gopath-less builds that stuff should go away...

calmh:
I did some hasty changes to not do the tmp gopath dance on the build server, we'll see if I broke something else in the process.

PR #4844, opened 2018-03-27
lib/osutil: Use unix lowprio implementation on Android

AudriusButkevicius:
Not sure if the tags are right, but I hope :)


calmh:
Note that our build process doesn't build anything with GOOS=android, so the best you'll know is that you didn't fuck up the linux build

AudriusButkevicius:
I am sure it's ok, as the space is an or comma is an and.

PR #4846, opened 2018-03-28
lib/fs: Don't panic when watching a folder with symlinked root

imsodin:
### Purpose

This was uncovered by accident in #4811: https://build.syncthing.net/viewLog.html?buildId=18757&buildTypeId=Syncthing_BuildLinuxCross  
But it's sufficiently different to get it's own PR/review. Essentially the notify backend always resolves symlinks in paths, while `basicFilesystem` does not. This makes path returned by the notify library look like they were outside of the folder root. In addition `unrooted` strips the leading path separator even if the path isn't below root, which the previous check didn't account for.

### Testing

One test to check the symlinked root scenario and one directly on `unrootedChecked` for the path separator issue.

### DIFF
--- a/lib/fs/basicfs_watch_test.go
+++ b/lib/fs/basicfs_watch_test.go
@@ -123,7 +123,7 @@ func TestWatchOutside(t *testing.T) {
 			}
 			cancel()
 		}()
-		fs.watchLoop(".", testDirAbs, backendChan, outChan, fakeMatcher{}, ctx)
+		fs.watchLoop(".", backendChan, outChan, fakeMatcher{}, ctx)
 	}()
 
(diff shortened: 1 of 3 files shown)


lib/fs/basicfs_watch_test.go
```diff
@@ -208,6 +208,54 @@ func TestWatchErrorLinuxInterpretation(t *testing.T) {
 	}
 }
 
+func TestWatchSymlinkedRoot(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Involves symlinks")
+	}
```
AudriusButkevicius: There is a better way to check this somewhere.

imsodin: i know bit it's in osutil and that imports fs so it's cyclic. I didn't consider it worth working around that.

PR #4847, opened 2018-03-28
cmd/syncthing/gui_statics.go: Set Content-Type header regardless of asset location

fuzzybear3965:
### Purpose

mimetypes for assets aren't always submitted (for example, if not using compiled-in assets).

### Testing

I used Chrome developer tools to ensure that assets were loaded properly when the value of the envvar `STGUIASSETS` was non-empty at runtime. Before this fix, the browser refused to load the assets due to identical mimetype errors triggered for each asset. After this fix, all assets are loaded successfully.


AudriusButkevicius:
Lgtm, I'll leave for @calmh to merge as he deals with author files.

calmh:
Thanks!

fuzzybear3965:
Haha, sick! I'm happy to help.

PR #4851, opened 2018-03-31
Non-existent function call in setInterval

fuzzybear3965:
### Purpose

Somewhere in the history (I don't know how to find when), this function was removed. So, this setInterval is just wasting cycles.


AudriusButkevicius:
It exists on line 723

fuzzybear3965:
Oh, man.

PR #4866, opened 2018-04-08
cmd/strelaypoolsrv: Move metric scraping to the server itself.

AudriusButkevicius:
Just like, close your eyes, imagine a nice beach, approve, and merge.

Things this does: 

1. Relaypoolsrv scrapes stats of relays by itself, every 30s by default using a magnificent pointer swap technique
2. Relaypoolsrv exports these part of /endpoint
3. Relaypoolsrv exports these (and it's own stats) as prometheus metrics on /metrics
4. Relaypoolsrv stores known relays in a file and tries to load it on startup. To make sure restarts (while I was developing) don't leave the pool empty.
5. Relays themselves export build info.


cmd/strelaypoolsrv/stats.go
```diff
@@ -0,0 +1,213 @@
...
+				stats: fetchStats(rel),
+			}
+			wg.Done()
+		}(rel)
+	}
+
+	wg.Wait()
+	close(results)
+
+	mut.Lock()
```
calmh: We'll block here for at minimum five seconds (the timeout, as there will always be someone just not responding), during which time we will not respond to requests for /endpoint? We get roughly 5/s requests so it's not a huge amount, but still means we are unavailable-ish for five seconds out of every 30

calmh:
I'm guessing this is tested by deploying, as it looks like we've started serving out more data from the relays server :D

calmh:
Oh, and I see the dashboardery. Nice.

cmd/strelaypoolsrv/stats.go
AudriusButkevicius: results is buffered, and wg.Wait() already returned, so we only go through what's in the buffered channel.

calmh: Oh, I missed the buffering.

calmh:
Sounds good to me but I'm not going to pretend to have read it all. I'll approve this one on trust.

PR #4871, opened 2018-04-09
lib/model: Prevent warning on request in paused folder (fixes #4870)

### LINKED ISSUE #4870: Don't show a warning when a file from a paused folder is requested
> ```
> 2018-04-03 19:02:20: Request from SomePeer for file 2011/201108 Coups de pagaies sur Dordogne/20110804_134450-FV-p8040297.jpg in unshared folder “xxx5x-xxx2x”.
> ```
> The peer is member in the folder share settings

> I think this is a race, essentially a folder gets paused as a request from a remote peer comes in. Paused folders are not shared with anyone.
> 
> This shouldn’t happen as the connection gets broken, but I suspect there is a window between folder tear down and actual disconnection where this can happen.

https://forum.syncthing.net/t/some-requests-from-peer-in-unshared-folder/11451
https://forum.syncthing.net/t/request-from-for-file-in-unshared-folder/11468

imsodin:
See #4870 for purpose. 

I didn't check that the race between connection drop and incoming request checks out, I just added the appropriate check for a paused folder to `Request`. In there I also stumbled over some pre-fs cruft and removed that as well. No specific testing done.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1309,23 +1309,33 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset 
 		return protocol.ErrInvalid
 	}
 
-	if !m.folderSharedWith(folder, deviceID) {
-		l.Warnf("Request from %s for file %s in unshared folder %q", deviceID, name, folder)
-		return protocol.ErrNoSuchFile
+	// Make sure the path is valid and in canonical form
+	var err error
+	if name, err = fs.Canonicalize(name); err != nil {
+		l.Debugf("Request from %s in paused folder %q for invalid filename %s", deviceID, folder, name)
+		return protocol.ErrInvalid
 	}
-	if deviceID != protocol.LocalDeviceID {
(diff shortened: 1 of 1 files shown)


lib/model/model.go
```diff
@@ -1920,11 +1930,6 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 			break
 		}
 
-		// We test each path by joining with "root". What we join with is
-		// not relevant, we just want the dotdot escape detection here. For
-		// historical reasons we may get paths that end in a slash. We
-		// remove that first to allow the rootedJoinedPath to pass.
-		sub = strings.TrimRight(sub, string(fs.PathSeparator))
```
AudriusButkevicius: Why removing this?

imsodin: `routedJoinedPath` doesn't exist anymore and scanner/fs doesn't care about trailing path separators in the prefixes/subs (also I believe in the sub unification before calling `internalScanFolderSubdirs` we already format paths).

PR #4874, opened 2018-04-11
script: Don't base64 encode the assets

calmh:
We did this to minimize source size and make it compile faster. Nowadays byte slice literals compile as fast as anything, and the source isn't checked in so it doesn't matter how big it is.

Not using base64 avoids a decode step at startup and makes the binary about 400k smaller.

### DIFF
--- a/script/genassets.go
+++ b/script/genassets.go
@@ -11,8 +11,8 @@ package main
 import (
 	"bytes"
 	"compress/gzip"
-	"encoding/base64"
 	"flag"
+	"fmt"
 	"go/format"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


imsodin:
Effects sound good, but I don't pretend to understand the change.

calmh:
Prior to this change, the genassets script generates entries like this for each asset that is compiled in:

```go
assets["black/assets/css/theme.css"], _ = base64.StdEncoding.DecodeString("H4sIAAAAAAAA/6xX3...
```

post the change, this happens instead:

```go
assets["black/assets/css/theme.css"] = []byte{0x1f, 0x8b, 0x8, ...
```

PR #4878, opened 2018-04-12
 lib/fs: Fix watcher panic due to casing on windows (fixes #4877) 

### LINKED ISSUE #4877: Panic loop with fswatcher enabled in syncthing v0.14.47-rc.1
Windows 10 x86_64

Syncthing gets stuck constantly restarting itself in a panic loop and it appears to be related to a particular folder on a non-OS drive.  When I manually disabled the fswatcher for that folder, syncthing appears to start working normally.

Here is an excerpt from the logs:
```
[XPQFK] 09:23:00 INFO: Completed initial scan of readwrite folder "Curse" (2g7rh-swpeu)
Panic at 2018-04-12T09:23:00-05:00
panic: bug: Notify backend is processing a change outside of the watched path: \\?\d:\minecraft\curse\Minecraft\Instances\Undiscovered\config\ruins_config\Gilly\GillyDCG2.tml
```

Full log (ip address and hostnames are obfuscated):

[panic-20180412-092300-redact.zip](https://github.com/syncthing/syncthing/files/1903640/panic-20180412-092300-redact.zip)

imsodin:
### Purpose

`filepath.EvalSymlinks` nicely converts the drive letter to uppercase, while `filepath.Clean` does not. Hence if a lowercase letter was used as folder path, the watcher would panic. Now I make sure that all absolute paths in the watcher code are symlink-evaluated.

### Testing

Added reproducer for #4877 

If anyone wants to merge, the second last commit message is the right one...

### DIFF
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -68,15 +68,8 @@ func adjustRoot(root string) string {
 	// Attempt to enable long filename support on Windows. We may still not
 	// have an absolute path here if the previous steps failed.
 	if runtime.GOOS == "windows" {
-		// The uppercasing is important, as filepath.EvalSymlinks converts
-		// to uppercase and folder escape checks depend on consistent
-		// case. See https://github.com/syncthing/syncthing/issues/4877
-		if strings.HasPrefix(root, `\\?\`) {
-			if vol := filepath.VolumeName(root[4:]); len(vol) == 2 {
-				root = `\\?\` + strings.ToUpper(vol) + root[6:]
-			}
-		} else if vol := filepath.VolumeName(root); len(vol) == 2 {
-			root = `\\?\` + strings.ToUpper(vol) + root[2:]
+		if filepath.IsAbs(root) && !strings.HasPrefix(root, `\\`) {
+			root = `\\?\` + root
 		}
 		return root
 	}
@@ -95,8 +88,20 @@ func adjustRoot(root string) string {
 // directory, this returns an error, to prevent accessing files that are not in the
 // shared directory.
 func (f *BasicFilesystem) rooted(rel string) (string, error) {
+	return rooted(rel, f.root)
+}
+
+// rootedSymlinkEvaluated does the same as rooted, but the returned path will not
+// contain any symlinks.  package. If the relative path somehow causes the final
+// path to escape the root directory, this returns an error, to prevent accessing
+// files that are not in the shared directory.
+func (f *BasicFilesystem) rootedSymlinkEvaluated(rel string) (string, error) {
+	return rooted(rel, f.rootSymlinkEvaluated)
+}
+
+func rooted(rel, root string) (string, error) {
 	// The root must not be empty.
-	if f.root == "" {
+	if root == "" {
 		return "", ErrInvalidFilename
 	}
 
@@ -104,7 +109,7 @@ func (f *BasicFilesystem) rooted(rel string) (string, error) {
 
 	// The expected prefix for the resulting path is the root, with a path
 	// separator at the end.
-	expectedPrefix := filepath.FromSlash(f.root)
+	expectedPrefix := filepath.FromSlash(root)
 	if !strings.HasSuffix(expectedPrefix, pathSep) {
 		expectedPrefix += pathSep
 	}
@@ -118,7 +123,7 @@ func (f *BasicFilesystem) rooted(rel string) (string, error) {
 	// The supposedly correct path is the one filepath.Join will return, as
 	// it does cleaning and so on. Check that one first to make sure no
 	// obvious escape attempts have been made.
-	joined := filepath.Join(f.root, rel)
+	joined := filepath.Join(root, rel)
 	if rel == "." && !strings.HasSuffix(joined, pathSep) {
 		joined += pathSep
 	}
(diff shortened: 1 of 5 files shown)


lib/fs/basicfs.go
```diff
@@ -88,16 +88,29 @@ func adjustRoot(root string) string {
 // directory, this returns an error, to prevent accessing files that are not in the
 // shared directory.
 func (f *BasicFilesystem) rooted(rel string) (string, error) {
+	return rooted(rel, f.root)
+}
+
+// rootedSymlinkEvaluated does the same as rooted, but the returned path
+// will not contain any symlinks.
+// package. If the relative path somehow causes the final path to escape the root
```
calmh: comment breakage

imsodin:
I assumed that on disk and what os/filepath reports is always the same - no user choice. We could always enforce upper case drive letters in `BasicFilesystem` the same way `filepath.EvalSymlink` does. It also avoids adding yet another "anti-symlink" function.

```diff
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -68,8 +68,11 @@ func adjustRoot(root string) string {
        // Attempt to enable long filename support on Windows. We may still not
        // have an absolute path here if the previous steps failed.
        if runtime.GOOS == "windows" {
-               if filepath.IsAbs(root) && !strings.HasPrefix(root, `\\`) {
-                       root = `\\?\` + root
+               if vol := filepath.VolumeName(root); len(vol) == 2 {
+                       // The uppercasing is important, as filepath.EvalSymlinks converts
+                       // to uppercase and folder escape checks depend on consistent
+                       // case. See https://github.com/syncthing/syncthing/issues/4877
+                       root = `\\?\` + strings.ToUpper(vol) + root[2:]
                }
                return root
        }
```


Also is a much smaller diff, it just looks pretty incon

imsodin:
I cleaned the commit chaos up and pushed both possible solutions:

 1. Separate `rootedSymlinkEvaluated` to get an absolute path based on `f.rootSymlinkEvaluated`:  
https://github.com/syncthing/syncthing/pull/4878/commits/456e24ae124ab68cf9c2e6d537f5f1a3a6e93df7

 2. Convert all windows volume letters to uppercase in the filesystem package:  
https://github.com/syncthing/syncthing/pull/4878/files/cdfb011bb03b0040a4842b6a6d96384a177c3438..1b6bfaff345dea0dca854fa7fca2366d24c724af

I don't see any gotchas with either option. I think 2. is the neater solution, but 1. is more "localized", i.e. less potential to break something totally unrelated that isn't on the radar.

@calmh The authors check fails because @wwwutz has wwwutz@gmail.com in the authors file and wwwutz@googlemail.com in commit.

wwwutz:
Oops, sorry for the email probs. I'll change it to gmail.com.

imsodin:
Thanks for noticing. So `filepath.EvalSymlinks` does even more magic - I get more and more scared of the greater case independence change.  

My current hypothesis given the `toNorm` function and syscalls involvement (only) in the windows specific part of the filepath package is, that all `os` and `filepath` functions except `filepath.EvalSymlinks` preserve the case of the path as given in the arguments. What exactly is done on windows is mysterious to me, it's pretty convoluted: https://golang.org/src/path/filepath/symlink_windows.go#L174

This means it has to be Option 1: It add the option to create absolute paths using the root that passed through `filepath.EvalSymlinks`. So whatever magic it does, it is consistent magic.

calmh:
Re the @wwwutz email, I wonder wtf happened there. The commit specifically had the gmail address in it, and that was rejected by the check, and I added that, and the check passed, and I merged. I guess GitHub must have replaced the email address as part of the squash merge? Very odd. I'll sort it out.

wwwutz:
Re my email @calmh sorry for thr OT noise, but googlemail.com was set up in my github profile for a couple of years. I never changed it coz it never was a problem. I changed it to gmail just after this issue came up today. In my profile! So wherever you find googlemail.com it came from some github-profile-magic. Which should not happen again. I'm looping. 8)

imsodin:
So the latest update also tests that differences in case in the root path (not only volume letter) is handled gracefully. From my side the current state is good to merge.

imsodin:
@AudriusButkevicius @calmh Do you have some time to review this? It is v0.14.47-rc.1 material.

AudriusButkevicius:
lgtm

calmh:
I'll pop an rc-2 tomorrow.

Side note @imsodin: I noticed that your commit messages often have a leading space (` lib/fs: Whatever`) that is almost invisible in the merge UI here and looks  bad afterwards if it slips through. Can you stop doing that? :D

imsodin:
> Side note @imsodin: I noticed that your commit messages often have a leading space (lib/fs: Whatever) that is almost invisible in the merge UI here and looks bad afterwards if it slips through. Can you stop doing that? :D

Ahem, I'll try, but where are these spaces? I seem to have a very stronk blind spot for them, I can't find any in this branch or anywhere else.

calmh:
Eh, good question - I don't see them when looking at your branch. Nonetheless GitHub adds them before and after in the subject in the merge dialog... Maybe it's a browser issue on my side! This is the one I missed, better at the end than at the beginning. :)

screen shot 2018-04-17 at 08 07 55

PR #4880, opened 2018-04-14
cmd/syncthing/gui.go: remove spurious <nil> debug for time.Parse()

wwwutz:
### Purpose

Remove spurious `gui.go:985:  DEBUG:  <nil>` on successful call `time.Parse()` in  `getSystemLog` and `getSystemLogTxt`.

### Testing

Turn on gui debug


calmh:
Thanks!

PR #4884, opened 2018-04-17
lib/model: Increase the default pull limit (fixes #4883)

### LINKED ISSUE #4883: Default defaultPullerPendingKiB size need to increased to 16M?
Now we have a maximum block size of 16M, but the default pull defaultPullerPendingKiB size is 8M, this will lead to the default configuration can not be pulled 16M file block

`
defaultPullerPendingKiB = 8192 // must be larger than block size
...
...
bytes := int(state.block.Size)
requestLimiter.take(bytes)
`

calmh:
Bumping the limit to 2 * the max block size (16 MiB) is a slight
increase compared to previously. Nonetheless I think it's good to allow
us to queue one request and have one on the way in, or conversely have
one large block on the way in and be able to ask for smaller blocks from
others at the same time.

### DIFF
--- a/lib/model/rwfolder.go
+++ b/lib/model/rwfolder.go
@@ -81,7 +81,7 @@ const (
 const (
 	defaultCopiers          = 2
 	defaultPullerPause      = 60 * time.Second
-	defaultPullerPendingKiB = 8192 // must be larger than block size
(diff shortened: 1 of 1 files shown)


xjtdy888:
I think our pull limit should be a global limit, which can effectively control the use of memory.when multiple folder are pulled at the same time

PR #4886, opened 2018-04-17
lib/osutil: Fix TraversesSymlink with symlinked fs root on windows (fixes #4875)

### LINKED ISSUE #4875: Zeroed local/global "state" (local files untouched) and remote file deletion with root Windows symlink
When specifying a sync directory as a symlink to another drive and using the filesystem watcher I receive the following error on Windows:

Failed to start filesystem watcher for folder "Documents" (mqanr-z452q): GetFileAttributesEx /S:\Documents: The filename, directory name, or volume label syntax is incorrect.

The path provided in Syncthing's config is "~/Documents", I suspect the / is not being stripped given GetFileAttributesEx **/S:\...** I verified the ~ is not related to the issue by also specifying the absolute path and verifying the error persists.

Also triggered by this is that after a seemingly successful scan of the directory and it's contents (e.g. the number of files and folders look correct) the files and folders drop to 0, and remote devices empty their drives resulting is data loss on remote devices (note the local sync folders are unaffected).

### Version Information

Syncthing Version: v0.14.46
OS Version: Windows 10

imsodin:
It's windows specific - I'll explain if the build server confirms it, nothing to do for anyone here (yet) :)

### DIFF
--- a/lib/osutil/traversessymlink.go
+++ b/lib/osutil/traversessymlink.go
@@ -32,29 +32,21 @@ func (e NotADirectoryError) Error() string {
 	return fmt.Sprintf("not a directory: %s", e.path)
 }
 
-// TraversesSymlink returns an error if base and any path component of name up to and
-// including filepath.Join(base, name) traverses a symlink.
-// Base and name must both be clean and name must be relative to base.
+// TraversesSymlink returns an error if any path component of name (including name
+// itself) traverses a symlink.
 func TraversesSymlink(filesystem fs.Filesystem, name string) error {
-	base := "."
-	path := base
-	info, err := filesystem.Lstat(path)
+	var err error
+	name, err = fs.Canonicalize(name)
 	if err != nil {
 		return err
 	}
-	if !info.IsDir() {
-		return &NotADirectoryError{
-			path: base,
-		}
-	}
 
 	if name == "." {
-		// The result of calling TraversesSymlink("some/where", filepath.Dir("foo"))
 		return nil
 	}
 
-	parts := strings.Split(name, string(fs.PathSeparator))
-	for _, part := range parts {
+	var path string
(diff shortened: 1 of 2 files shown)


imsodin:
In #4765 (v0.14.46) I started using `osutil.TraversesSymlink` (through `osutil.IsDeleted`) to detect deleted files. This is fine except that `osutil.TraversesSymlink` wrongly also checks the folder root. It uses `IsDir`, which on linux is true for a symlink to a dir, but false on windows. So if you change the folder root to be a symlink on windows, the entire folder gets wiped in the db. So this was nicely hidden by working on linux as expected and there being hardly any windows&symlink tests - proving once again that the combination is evil.

This is in principle no rc material, but given that it has potential for data loss, I'd propose to still add it to v0.14.47 @AudriusButkevicius @calmh

AudriusButkevicius:
There not being tests does not prove the combination is evil, it proves we are shit developers.

lib/osutil/traversessymlink.go
```diff
@@ -32,29 +32,21 @@ func (e NotADirectoryError) Error() string {
...
 	}
-	if !info.IsDir() {
-		return &NotADirectoryError{
-			path: base,
-		}
-	}
 
 	if name == "." {
-		// The result of calling TraversesSymlink("some/where", filepath.Dir("foo"))
 		return nil
```
AudriusButkevicius: Leave this comment?

calmh:
LGTM. So with this in mind, maybe we can make an attempt to fix the following symlink related tests that are skipped on Windows (maybe not *all* of them symlink related, but clearly many):

screen shot 2018-04-17 at 19 28 03

imsodin:
Hooks seem to be broken and manually starting builds runs on the old commits - so fuck tests and merge? It was just a comment.

AudriusButkevicius:
Do it

PR #4888, opened 2018-04-17
lib/model: Allow limiting number of concurrent scans (fixes #2760)

### LINKED ISSUE #2760: Max simultaneous scans?

AudriusButkevicius:
Tested by adding a sleep statement prior the scan starting 🗡 

image

Request to control the order/priority by number of LOL.txt files inside the folder coming in 3... 2... 1...


imsodin:
No LOLs, but still some potentially annoying thoughts: Waiting at that stage blocks all folder activity. It would be nicer if pulling and stuff would still work, i.e. doing the global scan limit check in the main folder loop (which might complicate other things I am not thinking about). Even more bike shedding: Multiple incoming scans (REST, timer, fs watcher) while waiting could be merged to prevent potential redundant scans back to back.

A completely different alternative I was contemplating is to limit hashers globally and assign hashing slots to active scans. I.e. still scan everything in parallel but hopefully speed things up by preventing disk hog due to lots of parallel reading.

AudriusButkevicius:
I guess you are right, yet at the same time I struggle to imagine how this would be implemented. We'd have to somehow queue the scans back to back, the state management becomes a stack, as in we're in a queued scan, hey we're in a pull now, hey we're back in a queued scan etc.

calmh:
It's simple, which is attractive. But I share imsodins concern. When I think about this, I think I'd hook into the loop in rwfolder/rofolder. When it's time to for a scheduled scan it attempts to grab one of the available scan tokens (non-blockingly) and proceeds with the scan if possible, otherwise postpone the scan or something. This still allows pulls to happen.

Similarly, forced scans (from the API or because we discover the need while pulling) shouldn't be forced to wait either - they could attempt to grab a token in the background and return it when they're done. If they didn't manage to get one at all it will be returned immediately when it's granted. If they do manage to get one it locks out other scheduled scans for the duration.

st-review:
:robot: *beep boop*

calmh:
Reopennfoe testing

AudriusButkevicius:
Testing what?

calmh:
Whether it works, in practice. As per forum.

calmh:
I had to spell the code a little differently as the scanning is now in a place where there is no global config object, but it's equivalent.

calmh:
yolo

AudriusButkevicius:
Missing a release note?

calmh:
GitHub is confused. The milestone this is on is now 0.14.55 as 0.14.54 slided in inbetween.

PR #4899, opened 2018-04-22
lib/config, lib/model: Include paused folders in cluster config (fixes #4897)

### LINKED ISSUE #4897: Pausing on introducer removes device on other devices
Didn't see this myself, but sounds serious enough to investigate:

```
23:39:57 <_aD> hmm, if an introducer pauses all folders that it shares with a device X,
               other devices that have been introduced remove that device X
```

imsodin:
### Purpose

Currently cluster configs are built based on the folders in the model internals. Paused folders are removed from the model, so they don't get into the cluster config. Thus to another device it looks like the folder was removed/unshared. This in itself is already incorrect. If this happens to all folders shared with one device, this device is removed on the remote if we are marked as introducer (#4897).
 
This delegates the creation of cluster configs to the config package and only adds the
index IDs and sequence numbers in the model. It also ensures that all lists of devices and
folders in the config are sorted by ID.

### Testing

Added a test to check that the cluster config contains paused folders.

### DIFF
--- a/lib/config/wrapper.go
+++ b/lib/config/wrapper.go
@@ -267,6 +267,13 @@ func (w *Wrapper) Folders() map[string]FolderConfiguration {
 	return w.folderMap
 }
 
+// FolderList returns a slice of devices.
+func (w *Wrapper) FolderList() []FolderConfiguration {
+	w.mut.Lock()
+	defer w.mut.Unlock()
+	return append([]FolderConfiguration(nil), w.cfg.Folders...)
+}
+
 // SetFolder adds a new folder to the configuration, or overwrites an existing
 // folder with the same ID.
 func (w *Wrapper) SetFolder(fld FolderConfiguration) (Waiter, error) {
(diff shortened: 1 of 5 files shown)


lib/config/wrapper.go
```diff
@@ -445,3 +445,39 @@ func (w *Wrapper) MyName() string {
 func (w *Wrapper) CheckHomeFreeSpace() error {
 	return checkFreeSpace(w.Options().MinHomeDiskFree, fs.NewFilesystem(fs.FilesystemTypeBasic, filepath.Dir(w.ConfigPath())))
 }
+
+// GenerateClusterConfig returns a protocol.ClusterConfig for the given peer device
+// with all the relevant configuration information. It still needs to be completed
+// with database information.
+func (w *Wrapper) GenerateClusterConfig(device protocol.DeviceID) protocol.ClusterConfig {
```
AudriusButkevicius: This should not live in the config.

imsodin: As all it does is translate the config accessing unimported infos I considered it a good fit. The other option would be to add something like `FolderSlice() []FolderConfiguration { return w.cfg.Folders }` to `Wrapper`.

AudriusButkevicius: Config is about config, ClusterConfig is a protocol level thing. If you want, have a helper in the model or something.

PR #4902, opened 2018-04-24
gui: localize numbers (fixes #4896)

### LINKED ISSUE #4896: Feature request: Thousand separator
In the device view, under local status, Syncthing shows the total amount of files, folders and the size of the data, like this:

image

With this many files and folders, it becomes pretty hard to get an idea of their number (is it hundreds of thousands or tens of thousands?). A thousand separator would help: `312'441` (EU) or `312,441` (US).

BTW the screenshot shows a real scenario, the host is a Raspberry Pi 3! With that much data on an external HDD, it still runs perfectly smooth. Syncthing never ceases to impress me.

wweich:
### Purpose
Localize Numbers, to add a thousands separator and set the decimal separator to the users locale.

The `toLocaleString` function uses the OS's locale. I didn't try to get the locale from the language selection, as it is not fine enough, as e.g. Switzerland has different chars for thousand / decimal separator than Germany, but both use the German translations (probably).

### Testing
Manually tested in Firefox and Chrome on Windows 7, Linux and Android, and in IE11 on Windows 7. (with German locale)

### Screenshots
Before:
grafik

After:
grafik

### DIFF
--- a/gui/default/syncthing/core/binaryFilter.js
+++ b/gui/default/syncthing/core/binaryFilter.js
@@ -6,16 +6,16 @@ angular.module('syncthing.core')
             }
             if (input > 1024 * 1024 * 1024) {
                 input /= 1024 * 1024 * 1024;
-                return input.toLocaleString(undefined, { maximumFractionDigits: 2}) + ' Gi';
+                return input.toLocaleString(undefined, {maximumFractionDigits: 2}) + ' Gi';
             }
             if (input > 1024 * 1024) {
                 input /= 1024 * 1024;
-                return input.toLocaleString(undefined, { maximumFractionDigits: 2}) + ' Mi';
+                return input.toLocaleString(undefined, {maximumFractionDigits: 2}) + ' Mi';
             }
             if (input > 1024) {
                 input /= 1024;
-                return input.toLocaleString(undefined, { maximumFractionDigits: 2}) + ' Ki';
+                return input.toLocaleString(undefined, {maximumFractionDigits: 2}) + ' Ki';
             }
-            return Math.round(input).toLocaleString(undefined, { maximumFractionDigits: 2}) + ' ';
+            return Math.round(input).toLocaleString(undefined, {maximumFractionDigits: 2}) + ' ';
         };
     });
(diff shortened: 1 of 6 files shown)


gui/default/syncthing/core/binaryFilter.js
```diff
@@ -6,16 +6,16 @@ angular.module('syncthing.core')
             }
             if (input > 1024 * 1024 * 1024) {
                 input /= 1024 * 1024 * 1024;
-                return input.toFixed(decimals(input, 2)) + ' Gi';
+                return input.toLocaleString(undefined, { maximumFractionDigits: 2}) + ' Gi';
             }
```
AudriusButkevicius: Space after {

AudriusButkevicius: Also, can decimal die now?

wweich:
The `decimals` function is now only used in the naturalFilter, which is currently used nowhere. So both could die now.

calmh:
Lost the limit on decimals for the cpu percentage, right? 0.886% isn’t horrible (though overly specific), but 68.234946% or something would be pretty ugly.

AudriusButkevicius:
Lets kill both dead things then

calmh:
The old system handled significant digits, so for percentages it would be three significant digits (0.123, 1.23, 12.3, 123), but I'm not going to fight about it.

wweich:
`toLocaleString` also has an `maximumSignificantDigits` option we could use. I could add another parameter to the `localeNumber`-Filter to set how the number (first parameter) should be used.

imsodin:
@wweich I would very much welcome that. I just tested something else and my status line was truncated:
screenshot_2018-04-26_14-27-02

If there is a way to use significant digits, that's perfect. If not I'd rather have no decimal digits at all.

wweich:
`maximumSignificantDigits` wouldn't help here, as it only cuts decimal digits. "normal" digits below significats get replace with zero.
So in your situation, setting `maximumSignificantDigits` to `3` would result in 221,000 files, 22,400 folders.

PR #4906, opened 2018-04-26
lib/db, lib/model: Add sequence->deviceKey to db for sending indexes

imsodin:
### Purpose

After a scan (full hashing - I botched up my database on the laptop, not Syncthing's fault) I was once again astonished how long it takes to "sync" the folder. Syncing in this case means resolving conflicts which are actually identical. The operation was CPU limited and looking at the profile the problem is obvious: Whenever the puller iteration sends a batch of file infos to the db, index updates are sent. Sending an index update means walking and unmarshaling the entire db for that folder (sorting by sequence isn't even in the profile, that's probably negligible in comparison).

So instead of doing that I propose to add a new section to the database to store filenames by sequence. For speed and simplicity I store the actual  full `deviceKey` instead of the filename. The additional 9 bytes per file seem worth it. In total this means 22 + filename bytes per file increased database size. For our 95% user with a (huge) mean file length of 200 this would amount to ~50MB additional db size.

### Testing

Added a test to make sure that `WithHaveSequence` does return files ordered by sequence. Also ran integration tests on it. Had it running on my laptop and it behaved well. Didn't do a full rehash to see whether the syncing/conflict resolution phase was actually significantly faster, but I don't doubt it is.


calmh:
I expect this is a very nice win, and I'm going to run with it for a little while.

While I'm doing that you can think about the other half of the equation. :) For every incoming index update we schedule a pull. The pull looks through the database for things that have changed...

calmh:
Effect is roughly as expected. I tested it with my two "main" devices, starting up with `-reset-deltas` and waiting until both sides were in sync and idle. The device I took the measurements on is the faster one (particularly, having the DB on SSD) so it spends quite some time waiting for the other side. Here's standard 0.14.47:

screen shot 2018-04-27 at 08 44 44

CPU usage is obvious, the network scale is positive=in, negative=out. The long wait time in the middle is waiting for the other side to sort the index, then it comes in at the end. It takes about 800 seconds to complete and settle down. With your change:

screen shot 2018-04-27 at 08 45 49

I kept the timescale same so they're comparable.

lib/db/leveldb_dbinstance.go
```diff
@@ -88,13 +89,21 @@ func (db *Instance) UpdateSchema() {
 	miscDB := NewNamespacedKV(db, string(KeyTypeMiscData))
 	prevVersion, _ := miscDB.Int64("dbVersion")
 
+	if prevVersion == dbVersion {
```
calmh: `>=` so we avoid downgrades? (Until we for some reason need downgrades and handle them correctly.)

lib/db/leveldb_dbinstance.go
```diff
@@ -171,7 +205,26 @@ func (db *Instance) withHave(folder, device, prefix []byte, truncate bool, fn It
...
+			return
+		}
+	}
+}
+
+func (db *Instance) withHaveSequence(folder []byte, startSeq int64, fn Iterator) {
+	t := db.newReadOnlyTransaction()
+	defer t.close()
+
+	dbi := t.NewIterator(util.BytesPrefix(db.sequenceKey(folder, 0)[:keyPrefixLen+keyFolderLen]), nil)
```
calmh: We could generate the proper range here directly instead of seeking, it ought to be from `sequenceKey(folder, startSet)` to `sequenceKey(folder, maxInt64)` where `maxInt64 = 1<<63-1`?

lib/db/leveldb_dbinstance.go
```diff
@@ -606,6 +659,23 @@ func (db *Instance) updateSchema0to1() {
...
+	t := db.newReadWriteTransaction()
+	defer t.close()
+
+	var sk []byte
+	var dk []byte
+
+	for _, folderStr := range db.ListFolders() {
+		folder := []byte(folderStr)
+		db.withHave(folder, protocol.LocalDeviceID[:], nil, true, func(f FileIntf) bool {
+			t.Put(db.sequenceKeyInto(sk, folder, f.SequenceNo()), db.deviceKeyInto(dk[:cap(dk)], folder, protocol.LocalDeviceID[:], []byte(f.FileName())))
```
calmh: This is quite compact. It wouldn't kill you to make the key creation business a couple of lines of their own.

lib/db/set.go
```diff
@@ -157,6 +157,8 @@ func (s *FileSet) Update(device protocol.DeviceID, fs []protocol.FileInfo) {
 		}
 		s.blockmap.Discard(discards)
 		s.blockmap.Update(updates)
+		s.db.removeSequences([]byte(s.folder), discards)
```
calmh: This is neat. I'd been thinking about something similar before but didn't know where to hook in the update code. 👍

lib/db/leveldb_dbinstance.go
```diff
@@ -144,6 +153,31 @@ func (db *Instance) updateFiles(folder, device []byte, fs []protocol.FileInfo, m
 	}
 }
 
+func (db *Instance) addSequences(folder []byte, fs []protocol.FileInfo) {
+	t := db.newReadWriteTransaction()
+	defer t.close()
+
+	var sk []byte
+	var dk []byte
```
AudriusButkevicius: Don't these need a make()? The code allocates a new slice if the original one is too small, but that doesn't affect the existing slice, just creates a new one, so I suspect this allocates on every iteration?

imsodin: Good catch. I only use the `...Into` functions as arguments instead of assigning. That's going to change due to https://github.com/syncthing/syncthing/pull/4906#discussion_r184605927 and then I think it should be fine.

lib/db/set.go
imsodin: Honestly my main thought throughout doing this was: Where is the catch? This is too simple to not having been done before.

lib/db/leveldb_dbinstance.go
AudriusButkevicius: You can just assign the return value back to sk in the loop, and pass in sk to tge db op.

AudriusButkevicius: Also, it's interesting if we don't have something like this already elsewhere.

lib/db/set.go
AudriusButkevicius: Sorry, I am curious now, what's the complicated bit here? Finding where to do removals?

lib/db/leveldb_dbinstance.go
imsodin: That's what I do since the review commit, see some lines below. And good instincts again, I do the same mistake in the migration code...

lib/db/set.go
calmh: I always imagined doing it in the something-update function in the middle of the database layer, where it would then need to load the previous version to find the old sequence etc. This is cleaner.

imsodin:
@calmh 

> While I'm doing that you can think about the other half of the equation. :) For every incoming index update we schedule a pull. The pull looks through the database for things that have changed...

Idea: Do the entire "need-business" on update and use it to manage new filename+device->nothing data indicating what we need. Then a call to `WithNeed` iterates over that, i.e. only over the actually needed file infos.

lib/db/leveldb_dbinstance.go
imsodin: And yes, there was quite a lot of potential for key reuse - I addressed it in https://github.com/syncthing/syncthing/pull/4906/commits/ab9dc3834a50ef5b675ac35aaa79ee5c3772614e

imsodin:
@AudriusButkevicius Sorry for dismissing your review, I made the mistake of not realizing that (unprefixed) db iteration keeps iterating also over other prefixes. I am surprised that it didn't trigger any consistent test failures (I discovered it due to a racy out of bounds read when binary decoding a value for a debug message o.O).

AudriusButkevicius:
So I'll let calmh put the nail in this coffin

calmh:
Yeah this appears to have broken the integration tests. It may just be because something is faster now than it was before and the test thing isn't keeping up, but I need to investigate.

calmh:
Yeah integration tests are not the fault of this change. 👍

PR #4908, opened 2018-05-01
Update Dockerfile to incorporate exposed ports

thunderstorm99:
added EXPOSE to Dockerfile. this way these ports will show up in docker front-ends like cockpit.

### Purpose

When including the EXPOSE and VOLUME functions, a front-end for docker will know which ports and what folders this container will use, therefore allowing users to more easily adjust these settings. 

### Testing

I re-built the container from scratch and no issues were found. To test it for yourself just paste in the one line and re-build the container with docker build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,8 @@ RUN rm -f syncthing && go run build.go build syncthing
 
 FROM alpine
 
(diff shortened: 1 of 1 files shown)


calmh:
Thanks!

thunderstorm99:
Well thank you.

PR #4931, opened 2018-05-08
gui: Enable proper asset caching

AudriusButkevicius:
Angular upgrade breaks everything spectacularly

### DIFF
--- a/cmd/syncthing/gui_statics.go
+++ b/cmd/syncthing/gui_statics.go
@@ -25,7 +25,7 @@ import (
 
 type staticsServer struct {
 	assetDir        string
-	assets          map[string]auto.Asset
+	assets          map[string][]byte
 	availableThemes []string
 
 	mut   sync.RWMutex
@@ -102,7 +102,7 @@ func (s *staticsServer) serveAsset(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Check for a compiled in asset for the current theme.
-	asset, ok := s.assets[theme+"/"+file]
+	bs, ok := s.assets[theme+"/"+file]
 	if !ok {
 		// Check for an overridden default asset.
 		if s.assetDir != "" {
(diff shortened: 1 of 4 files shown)


cmd/syncthing/gui_statics.go
```diff
@@ -115,14 +118,32 @@ func (s *staticsServer) serveAsset(w http.ResponseWriter, r *http.Request) {
...
+		asset, ok = s.assets[config.DefaultTheme+"/"+file]
 		if !ok {
 			http.NotFound(w, r)
 			return
 		}
 	}
 
+	etag := fmt.Sprintf("%d", asset.Modified.Unix())
+
+	w.Header().Set("Last-Modified", asset.Modified.Format(http.TimeFormat))
```
calmh: I'd skip this whole `auto.Asset` and `asset.Modified` business and use our `main.BuildStamp` as the modification time for all assets...?

AudriusButkevicius: But this is nice :(

calmh: I mean yeah it looks proper and all, pats on the back for that. But the end result is the same with less code. :)

AudriusButkevicius: I'll have to path up relaypoolsrv somehow tho, so not sure how much less code it will be.

calmh: Fine.

calmh: Must be `asset.Modified.UTC().Format(http.TimeFormat)` though or the time stamp might be a lie since the format hard codes `GMT`

calmh: Or maybe better add the `.UTC()` to when it parses the unixtime at startup.

calmh: I have a distinct memory we already had all this code at one point and already had the bug with the time zone format and the if-modified-since either always failing or always succeeding depending on which side of the greenwich line you were on... Due to the time.Parse() handling zones correctly and the `.Format(http.TimeFormat)` just being an ugly hack really

AudriusButkevicius: The format indicates that it should expect a timezone no?

calmh: https://play.golang.org/p/Did5Tn_OFDQ

calmh: Yep

AudriusButkevicius:
Updated as per comments, I think.

AudriusButkevicius:
This potentially could have been part of the RC as angular change fucks everything spectacularly.

calmh:
Yep. I’ve bent the rules slightly and released rc3 with this and other recent bug fixes.

Issue #4925, opened 2018-05-07
Deleted files within deleted directories aren't detected with FS watcher


FloGa:
Sometimes, the FsWatcher recognizes a directory as being deleted, but not the contents of this directory. I cannot reliably reproduce this.

Suppose I am in a shared directory, label "default". As a minimal example, I create a new folder and put a single file in it:

    mkdir newfolder
    touch newfolder/newfile

Wait ten seconds, until FsWatcher triggers a sync. The log shows correctly:

    [HU2GR] 00:01:14 VERBOSE: Folder "default" is now scanning
    [HU2GR] 00:01:14 VERBOSE: Local change detected in folder "default": modified dir newfolder
    [HU2GR] 00:01:14 VERBOSE: Local change detected in folder "default": modified file newfolder/newfile
    [HU2GR] 00:01:14 VERBOSE: Folder "default" is now idle
    [HU2GR] 00:01:14 VERBOSE: Folder "default" is now syncing
    [HU2GR] 00:01:17 VERBOSE: Folder "default" is now idle
    [HU2GR] 00:01:18 VERBOSE: Summary for folder "default" is map[localDirectories:2737 globalFiles:8590 inSyncFiles:8590 localBytes:404203037 pullErrors:0 inSyncBytes:404203037 localDeleted:546018 needSymlinks:0 needDirectories:0 needFiles:0 needDeletes:0 globalDirectories:2737 globalDeleted:882565 localFiles:8590 globalSymlinks:0 globalBytes:404203037 localSymlinks:0 needBytes:0 sequence:5339784 version:5339784 state:idle]
    [HU2GR] 00:01:33 VERBOSE: Completion for folder "default" on device ENFQI... is 99.99996833274659%
    [HU2GR] 00:01:52 VERBOSE: Device ENFQI... sent an index update for "default" with 2 items
    [HU2GR] 00:01:59 VERBOSE: Completion for folder "default" on device ENFQI... is 100%

(`ENFQI` being another computer under my control.)

If I delete the directory, like so:

    rm -r newfolder

Then sometimes only the directory will be recognized as deleted:

[...]

PR #4936, opened 2018-05-10
lib/osutil: Add test for IsDeleted (ref #4925)

imsodin:
Trying to reproduce #4925.

### DIFF
--- a/lib/osutil/osutil_test.go
+++ b/lib/osutil/osutil_test.go
@@ -226,8 +226,10 @@ func TestIsDeleted(t *testing.T) {
 		{"dir/del", true},
 		{"dir/del/del", true},
 		{"del/del/del", true},
-		{"inacc", false},
-		{"inacc/file", false},
+	}
+	if runtime.GOOS != "windows" {
+		// Doesn't support 0000 dir permissions
+		cases = append(cases, tc{"inacc", false}, tc{"inacc/file", false})
 	}
 
 	testFs := fs.NewFilesystem(fs.FilesystemTypeBasic, "testdata")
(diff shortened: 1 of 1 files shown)


imsodin:
@calmh Is the linux build run as root and thus not failing to stat `inacc/file` where `inacc` has `000` perms?

imsodin:
@AudriusButkevicius Yes, apart from the minor issue mentioned above (that scenario is now just skipped on the linux build server).

calmh:
No, no root, but I don't think 000 will prevent a stat (didn't look at the code)

imsodin:
@calmh Any ideas why it doesn't prevent it on the build server (it does on my pc)?

imsodin:
I (too hastily) updated the test to just skip that part if it fails. See the ["old test"](https://github.com/syncthing/syncthing/pull/4936/files/da08b34979d5680df6f24f9845c24dcd346289f7#diff-bc0d4ee330436348056e2922eb8f814aR252) and resulting [test failure](https://build.syncthing.net/viewLog.html?buildId=20589&buildTypeId=Syncthing_BuildLinuxCross&tab=buildLog&_focus=1082&state=1082#_state=1082)

calmh:
Right, so the Linux build agent *does* in fact run as root in it's litter docker world. Sorry about that misinformation. I think that's pretty broken but probably the alternative is a mess of permission problems instead. Maybe we can run with your later root-avoiding version of the test and be happy that it succeeds as non-root on for example mac instead. :/

imsodin:
Totally. It will also work whenever anyone runs tests locally.

calmh:
At least you're not here yet: https://github.com/kubernetes/kubernetes/commit/eace2d08d0b3d5b493eefec9f8532ef2980617ac

PR #4938, opened 2018-05-10
lib/watchaggregator: Prevent race on config update

imsodin:
It never triggered the race detection before, now it did for once. The folder id never changes, so store that on creation.

### DIFF
--- a/lib/watchaggregator/aggregator.go
+++ b/lib/watchaggregator/aggregator.go
@@ -98,7 +98,8 @@ func (dir *eventDir) eventType() fs.EventType {
 }
 
 type aggregator struct {
-	// Needed to prevent race due to asynchronous, exported CommitConfiguration
+	// folderID never changes and is accessed in CommitConfiguration, which
+	// asynchronously updates folderCfg -> can't use folderCfg.ID (racy)
 	folderID        string
 	folderCfg       config.FolderConfiguration
 	folderCfgUpdate chan config.FolderConfiguration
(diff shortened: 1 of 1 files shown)


imsodin:
It's a bit obscure: Several calls to `CommitConfiguration` can lead to a race on `a.folderCfg.ID`.

```
WARNING: DATA RACE
Read at 0x00c422eb9c80 by goroutine 320:
  github.com/syncthing/syncthing/lib/watchaggregator.(*aggregator).CommitConfiguration()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/watchaggregator/aggregator.go:393 +0x127
  github.com/syncthing/syncthing/lib/config.(*Wrapper).notifyListener()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/config/wrapper.go:184 +0x193
  github.com/syncthing/syncthing/lib/config.(*Wrapper).notifyListeners.func1()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/config/wrapper.go:175 +0x178

Previous write at 0x00c422eb9c80 by goroutine 312:
  github.com/syncthing/syncthing/lib/watchaggregator.(*aggregator).mainLoop()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/watchaggregator/aggregator.go:408 +0x7cd
```

imsodin:
I can't follow. `aggregator.CommitConfiguration` is asynchronous, so there is nothing preventing multiple calls to it right after each other and that is racy. I think this is something to be fixed within the watchaggregator package regardless of whether Syncthing does or does not do such calls.

AudriusButkevicius:
Lgtm yet could explain better in the comment that the race happens because folderCfg variable gets adjustes in the main loop hence two calls to commitconfig can cause a data race.

imsodin:
@AudriusButkevicius Does the new comment make sense?

PR #4939, opened 2018-05-10
lib/model: Refactor {ro,rw}folder serve loop into just folder

calmh:
The actual pull method (which is really the only thing that differs between them) is now an interface member which gets overridden by the subclass.

"Subclass?!" Well, this is dynamic dispatch with overriding, I guess.

### Purpose

Yeah, about that. I'm going to implement the damn recv only folder. As part of that I need to clean shit up. I hope to try to do that in manageable steps so it doesn't turn into an impossible-to-review ball of mud diff. No promises, though.

In the end I want something like the `folder` type that has all the scaffolding and looping and stuff, a `puller` member that implements the pull (like this PR) and something similar on the scan side. Possibly a scan results filter of some sort. Then the three folder types built out of these components.

### Testing

It compiles, and it passes the integration tests...

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -44,7 +44,7 @@ type folder struct {
 }
 
 type puller interface {
-	pull() (success bool)
+	pull() bool // true when successfull and should not be retried
 }
 
 func newFolder(model *Model, cfg config.FolderConfiguration) folder {
@@ -107,13 +107,18 @@ func (f *folder) Serve() {
 			}
 
 		case <-pullFailTimer.C:
-			if !f.puller.pull() {
-				// Pulling failed, try again later.
-				pullFailTimer.Reset(pause)
-				// Back off from retrying to pull with an upper limit.
-				if pause < 60*f.basePause() {
-					pause *= 2
-				}
+			if f.puller.pull() {
+				// We're good. Don't schedule another fail pull and reset
+				// the pause interval.
+				pause = f.basePause()
+				continue
+			}
+
+			// Pulling failed, try again later.
+			pullFailTimer.Reset(pause)
+			// Back off from retrying to pull with an upper limit.
+			if pause < 60*f.basePause() {
+				pause *= 2
 			}
 
 		case <-initialCompleted:
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -39,6 +39,12 @@ type folder struct {
 	restartWatchChan chan struct{}
 	watchErr         error
 	watchErrMut      sync.Mutex
+
+	puller puller
+}
+
+type puller interface {
+	pull() (success bool)
 }
 
 func newFolder(model *Model, cfg config.FolderConfiguration) folder {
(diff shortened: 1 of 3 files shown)


lib/model/folder.go
```diff
@@ -64,6 +70,83 @@ func newFolder(model *Model, cfg config.FolderConfiguration) folder {
...
+				pullFailTimer.Reset(pause)
+			}
+
+		case <-pullFailTimer.C:
+			if !f.puller.pull() {
+				// Pulling failed, try again later.
+				pullFailTimer.Reset(pause)
+				// Back off from retrying to pull with an upper limit.
+				if pause < 60*f.basePause() {
+					pause *= 2
```
AudriusButkevicius: Out of this pr context, but the fact we never reset this is worrying.

lib/model/folder.go
```diff
@@ -39,6 +39,12 @@ type folder struct {
 	restartWatchChan chan struct{}
 	watchErr         error
 	watchErrMut      sync.Mutex
+
+	puller puller
+}
+
+type puller interface {
+	pull() (success bool)
```
AudriusButkevicius: Dont need the braces, can just be pull() bool?

lib/model/rwfolder.go
```diff
@@ -135,121 +134,29 @@ func newSendReceiveFolder(model *Model, cfg config.FolderConfiguration, ver vers
...
-		// make sure to not block index receiving.
-	}
-}
-
 func (f *sendReceiveFolder) String() string {
 	return fmt.Sprintf("sendReceiveFolder/%s@%p", f.folderID, f)
 }
 
-func (f *sendReceiveFolder) pull(prevIgnoreHash string) (curIgnoreHash string, success bool) {
+func (f *sendReceiveFolder) pull() (success bool) {
```
AudriusButkevicius: Pull() bool

lib/model/rwfolder.go
```diff
@@ -94,9 +94,10 @@ type dbUpdateJob struct {
 type sendReceiveFolder struct {
 	folder
 
-	fs        fs.Filesystem
-	versioner versioner.Versioner
-	pause     time.Duration
+	prevIgnoreHash string
+	fs             fs.Filesystem
+	versioner      versioner.Versioner
+	pause          time.Duration
```
AudriusButkevicius: Not used anymore?

lib/model/folder.go
```diff
@@ -39,6 +39,12 @@ type folder struct {
 	restartWatchChan chan struct{}
 	watchErr         error
 	watchErrMut      sync.Mutex
+
+	puller puller
```
AudriusButkevicius: Still feels that sendRcv et all should embed folder, and then derived can just implement their own pull (and base folder panics).

calmh: Yeah that's the intention. But embedded structs can't see their embedder to delegate functionality to them, so we need some hook in the other direction, either a func or interface value of some kind.

lib/model/folder.go
calmh: lol

lib/model/folder.go
AudriusButkevicius: Oh, that sucks, I though embedded.Foo() would call parent.Foo().

AudriusButkevicius:
Lgtm, tests  are slooooow

PR #4940, opened 2018-05-10
lib/db: Fix prefixed walks (fixes #4925)

### LINKED ISSUE #4925: Deleted files within deleted directories aren't detected with FS watcher

imsodin:
### Purpose

Currently walking the db with a prefix is broken in the presence of a file, that has the same prefix, as the entire walk is aborted when such a file is encountered.

### Testing

Added new unit test.

### DIFF
--- a/lib/db/benchmark_test.go
+++ b/lib/db/benchmark_test.go
@@ -19,9 +19,13 @@ import (
 )
 
 var files, oneFile, firstHalf, secondHalf []protocol.FileInfo
-var s *db.FileSet
+var benchS *db.FileSet
 
-func init() {
+func lazyInitBenchFileSet() {
+	if benchS != nil {
+		return
+	}
+
 	for i := 0; i < 1000; i++ {
 		files = append(files, protocol.FileInfo{
 			Name:    fmt.Sprintf("file%d", i),
@@ -36,9 +40,9 @@ func init() {
 	oneFile = firstHalf[middle-1 : middle]
 
 	ldb, _ := tempDB()
-	s = db.NewFileSet("test)", fs.NewFilesystem(fs.FilesystemTypeBasic, "."), ldb)
-	replace(s, remoteDevice0, files)
-	replace(s, protocol.LocalDeviceID, firstHalf)
+	benchS = db.NewFileSet("test)", fs.NewFilesystem(fs.FilesystemTypeBasic, "."), ldb)
+	replace(benchS, remoteDevice0, files)
+	replace(benchS, protocol.LocalDeviceID, firstHalf)
 }
 
 func tempDB() (*db.Instance, string) {
@@ -70,16 +74,19 @@ func BenchmarkReplaceAll(b *testing.B) {
 }
 
 func BenchmarkUpdateOneChanged(b *testing.B) {
+	lazyInitBenchFileSet()
+
 	changed := make([]protocol.FileInfo, 1)
 	changed[0] = oneFile[0]
 	changed[0].Version = changed[0].Version.Update(myID)
 	changed[0].Blocks = genBlocks(len(changed[0].Blocks))
 
+	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
 		if i%1 == 0 {
-			s.Update(protocol.LocalDeviceID, changed)
+			benchS.Update(protocol.LocalDeviceID, changed)
 		} else {
-			s.Update(protocol.LocalDeviceID, oneFile)
+			benchS.Update(protocol.LocalDeviceID, oneFile)
 		}
 	}
 
@@ -87,17 +94,23 @@ func BenchmarkUpdateOneChanged(b *testing.B) {
 }
 
 func BenchmarkUpdateOneUnchanged(b *testing.B) {
+	lazyInitBenchFileSet()
+
+	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		s.Update(protocol.LocalDeviceID, oneFile)
+		benchS.Update(protocol.LocalDeviceID, oneFile)
 	}
 
 	b.ReportAllocs()
 }
 
 func BenchmarkNeedHalf(b *testing.B) {
+	lazyInitBenchFileSet()
+
+	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
 		count := 0
-		s.WithNeed(protocol.LocalDeviceID, func(fi db.FileIntf) bool {
+		benchS.WithNeed(protocol.LocalDeviceID, func(fi db.FileIntf) bool {
 			count++
 			return true
 		})
@@ -110,9 +123,12 @@ func BenchmarkNeedHalf(b *testing.B) {
 }
 
 func BenchmarkHave(b *testing.B) {
+	lazyInitBenchFileSet()
+
+	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
 		count := 0
-		s.WithHave(protocol.LocalDeviceID, func(fi db.FileIntf) bool {
+		benchS.WithHave(protocol.LocalDeviceID, func(fi db.FileIntf) bool {
 			count++
 			return true
 		})
@@ -125,9 +141,12 @@ func BenchmarkHave(b *testing.B) {
 }
 
 func BenchmarkGlobal(b *testing.B) {
+	lazyInitBenchFileSet()
+
+	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
 		count := 0
-		s.WithGlobal(func(fi db.FileIntf) bool {
+		benchS.WithGlobal(func(fi db.FileIntf) bool {
 			count++
 			return true
 		})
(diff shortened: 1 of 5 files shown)


lib/db/set_test.go
```diff
@@ -892,6 +892,57 @@ func TestWithHaveSequence(t *testing.T) {
...
+
+	i = 0
+	s.WithPrefixedGlobalTruncated("dir/", func(fi db.FileIntf) bool {
+		i++
+		return true
+	})
+	if i != 1 {
+		t.Errorf(`Expected 1, got %v global items below "dir/"`, i)
+	}
+}
```
AudriusButkevicius: Just out of curiosity, was this failing previously on second time iterating? If yes, that would mean the iterators are somehow stateful? If not, I am confused why you do the same thing twice.

lib/db/leveldb_dbinstance.go
```diff
@@ -923,6 +903,19 @@ func unmarshalTrunc(bs []byte, truncate bool) (FileIntf, error) {
...
 }
 
+func unmarshalVersionList(data []byte) (VersionList, bool) {
+	var vl VersionList
+	if err := vl.Unmarshal(data); err != nil {
+		l.Debugln("unmarshal error:", err)
+		return VersionList{}, false
+	}
+	if len(vl.Versions) == 0 {
+		l.Debugln("empty version list")
```
AudriusButkevicius: I can't see how these debug logs are useful, they would appear thousands and thousands of times in a db walk, amd include no indication which file they are talking about etc. I'd just remove them.

lib/db/leveldb_dbinstance.go
```diff
@@ -186,20 +186,27 @@ func (db *Instance) removeSequences(folder []byte, fs []protocol.FileInfo) {
...
-	slashedPrefix := prefix
-	if !bytes.HasSuffix(prefix, []byte{'/'}) {
-		slashedPrefix = append(slashedPrefix, '/')
-	}
-
 	for dbi.Next() {
 		name := db.deviceKeyName(dbi.Key())
-		if len(prefix) > 0 && !bytes.Equal(name, prefix) && !bytes.HasPrefix(name, slashedPrefix) {
+		if len(prefix) > 0 && !bytes.HasPrefix(name, prefix) {
 			return
```
AudriusButkevicius: So this no longer walks the directory itself? I think that's not how it should work, as I think walk returns the directory.

lib/db/leveldb_dbinstance.go
```diff
@@ -306,75 +319,53 @@ func (db *Instance) getGlobal(folder, file []byte, truncate bool) (FileIntf, boo
 		return nil, false
 	}
 
-	var vl VersionList
-	err = vl.Unmarshal(bs)
-	if err == leveldb.ErrNotFound {
-		return nil, false
-	}
```
AudriusButkevicius: I wonder why this error check is here. Seems like an unprobable error to be returned by Unmarshal?

lib/db/leveldb_dbinstance.go
calmh: It's the same territory as the error above, it's useless-ish but it's never supposed to happen

lib/db/leveldb_dbinstance.go
calmh: I think this handled at the top.

imsodin:
@calmh Exactly because the previous implementation didn't handle it correctly/there were no tests specifying the desired behaviour, I just implemented the behaviour I am used to from rsync. I can easily change it to include `dir` regardless of whether `dir` or `dir/` is given as prefix. Should I?

calmh:
Yes, I think so. I don't offhand know whether we always/never/sometimes pass the slash, but I don't think we ever use the prefixed iteration to mean anything else than "everything" (no prefix) or "this directory and children" (any prefix). So accepting both and behaving the same in both cases seems sane

lib/db/leveldb_dbinstance.go
AudriusButkevicius: Then change it to a panic?

imsodin:
Just changed the behaviour to always iterate over the prefix, and thus discovered a small bug in `model.GlobalDirectoryTree` and a negligence in a model test: An index with only files, not parent directory is used. As far as I know this should never happen in reality (i.e. parents must exist). I see two options:

1. When updating files in the database, always check that parent directories exist.
2. As before, just ignore missing parent directories when walking the db (e.g. if prefix is `foo/bar` and `foo/bar` is not in the db, but `foo/bar/baz` is, just iterate over `foo/bar/baz`).

Opinions?

imsodin:
I just went with previous behaviour. The `lib/db/benchmark_test.go` change is unrelated, it just floods logs with db debugging enabled (as in MBs). The [change in model.go](https://github.com/syncthing/syncthing/pull/4940/files#diff-1df9805b2381b6288b220226f82d5dc9L2406) is a fix to keep existing tests passing (i.e. the previous condition was useless, but for reasons unknown tests still passed).

lib/db/benchmark_test.go
```diff
@@ -36,9 +40,9 @@ func init() {
 	oneFile = firstHalf[middle-1 : middle]
 
 	ldb, _ := tempDB()
-	s = db.NewFileSet("test)", fs.NewFilesystem(fs.FilesystemTypeBasic, "."), ldb)
-	replace(s, remoteDevice0, files)
-	replace(s, protocol.LocalDeviceID, firstHalf)
+	benchS = db.NewFileSet("test)", fs.NewFilesystem(fs.FilesystemTypeBasic, "."), ldb)
```
AudriusButkevicius: Bracket in string

imsodin: Yeah, that's common to all (most?) lib/db tests - no idea why, I just kept it when c&p-ing.

calmh: I think it's just a mistake that's been perpetuated. I've cleaned it out of the other test file in another pr

PR #4942, opened 2018-05-11
all: Serialize folder types to new names

calmh:
It's been a year and a half since we started accepting the new names.
It's time we start producing them.

@AudriusButkevicius @canton7 @kozec will this break all the wrappers? It changes the folder type returned by the /rest/system/config call and the string on disk. If so should we hold off a little while or should we just say that we will *never* change the names and scrap this?

We still accept the old type in incoming configs, and should probably continue to do so forever.

The reason for the change is that the future trio `readwrite`, `readonly`, `recvonly` [makes no sense whatsoever](https://thedailywtf.com/articles/What_Is_Truth_0x3f_) (which is why Heiko introduced the new names way back).

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -274,7 +274,7 @@
               <div class="panel-progress" ng-show="folderStatus(folder) == 'scanning' && scanProgress[folder.id] != undefined" ng-attr-style="width: {{scanPercentage(folder.id)}}%"></div>
               <h4 class="panel-title">
                 <div class="panel-icon hidden-xs">
-                  <span class="fa fa-fw" ng-class="[folder.type == 'readonly' ? 'fa-lock' : 'fa-folder']"></span>
+                  <span class="fa fa-fw" ng-class="[folder.type == 'sendonly' ? 'fa-lock' : 'fa-folder']"></span>
                 </div>
                 <div class="panel-status pull-right text-{{folderClass(folder)}}" ng-switch="folderStatus(folder)">
                   <span ng-switch-when="paused"><span class="hidden-xs" translate>Paused</span><span class="visible-xs">&#9724;</span></span>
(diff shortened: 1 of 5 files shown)


AudriusButkevicius:
I can control it in android by fixing it as we upgrade syncthing version. I think synctrayzor shouldn't care as it does not change display based on that. Gtk will most likely show bad stuff in the UI. There is also tue OS X wrapper and the generic c++ tray icon that might or might not be affected.

I say fuck it and lets go with this.

gui/default/index.html
```diff
@@ -358,11 +358,11 @@ <h4 class="panel-title">
...
                       </td>
                     </tr>
-                    <tr ng-if="folder.type != 'readwrite'">
+                    <tr ng-if="folder.type != 'sendreceive'">
                       <th><span class="fa fa-fw fa-lock"></span>&nbsp;<span translate>Folder Type</span></th>
                       <td class="text-right">
-                        <span ng-if="folder.type == 'readonly'" translate>Send Only</span>
-                        <span ng-if="folder.type != 'readonly'">{{ folder.type.charAt(0).toUpperCase() + folder.type.slice(1) }}</span>
+                        <span ng-if="folder.type == 'sendonly'" translate>Send Only</span>
+                        <span ng-if="folder.type == 'sendreceive'" translate>Send & Receive</span>
```
uok: you can drop this line as it will never show because <tr> above is conditional, which I think is best solution to save vertical space if folder runs in normal mode. No need to output "send & receive"

canton7:
I stopped trying to deserialize that attribute after you broke it two years ago, so I'm good!

imsodin:
Also pinging @Martchus @xor-gate

xor-gate:
I'm not using the folder type in the syncthing-macosx wrapper currently. I only use the label in the tray menu.

Catfriend1:
Which milestone is this expected to go in? 14.48?

AudriusButkevicius:
49

calmh:
Ok sounds like no big deal. The deserializing of both kinds is already in place since ages so not visible in the diff. I'll clean up the rest today.

Martchus:
Thanks for pinging. I don't think my tray application will do anything stupid if you change that. 

BTW: Don't forget to update the documentation for the REST-API as well.

PR #4946, opened 2018-05-12
gui: Change rescan interval only if fs watcher was actually toggled (fixes #4944)

### LINKED ISSUE #4944: Edit Folder UI displays default "Full Rescan Interval (s)" value instead of the actual one
Scenario:
 - Set "Full Rescan Interval (s)" to something other than default 3600s (watcher enabled), for example to 9000s (2 h 30 min)
 - Save the change and confirm it is written properly in config.xml
 - Restart syncthing
 - Open UI, review folder info
 - Folder info shows rescan at 2h 30min correctly
 - Open Edit Folder dialog
 - Switch to "Advanced" tab
 * Actual result: "Full Rescan Interval (s)" shows 3600
 * Expected result: "Full Rescan Interval (s)" shows 9000 as saved in config.xml

Clicking "Save" at this point will save 3600 to the config, even though the user has never entered such value.

Reproduction isn't 100%, and my impression is that the more folders devices configured, and the slower everything is, the more likely the issue is to occur.
The occurrence conditions are somewhat similar to the issue 4915 (stignore is accidentaly overwritten by web UI on slow machines). Although the two probably aren't closely related.

### Version Information

Syncthing Version: v0.14.47
OS Version: Windows 10 x64
Browser Version: SyncTrayzor's UI

imsodin:
See #4944 for problem description. Tested by manual fooling around.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1506,13 +1506,13 @@ angular.module('syncthing.core')
             $scope.currentFolder.path = pathJoin($scope.config.options.defaultFolderPath, newvalue);
         });
 
-        $scope.$watch('currentFolder.fsWatcherEnabled', function (newvalue) {
-            if (newvalue) {
+        $scope.fsWatcherToggled = function () {
+            if ($scope.currentFolder.fsWatcherEnabled) {
                 $scope.currentFolder.rescanIntervalS = 3600;
             } else {
                 $scope.currentFolder.rescanIntervalS = 60;
             }
-        });
+        };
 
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
Why does this fix it? Seems to do the same ?

imsodin:
@AudriusButkevicius It does what the previous one should have done. But watching the variable hat the side-effect, that it was also toggled when e.g. modifying a dir with fs watcher enabled, then editing one with fs watcher disabled (without ever touching the checkbox). Now the user needs to actually click the checkbox to trigger the interval change.

PR #4950, opened 2018-05-12
all: Use Executable from os instead of osext (fixes #4900)

### LINKED ISSUE #4900: replace deprecated `github.com/kardianos/osext` with standard library's `os`
According to comments on an upstream issue (https://github.com/kardianos/osext/issues/21), this go package should not be used anymore (at least not on go 1.8+), except when compatibility with older go versions is desired. Since this isn't the case with `syncthing`, upstream recommended replacing usage of `osext.Executable()` with `os.Executable()`, which is available via the go standard library since go 1.8.

The only usage of this - now seemingly deprecated - package and function is in this file, according to a quick search: `syncthing/lib/upgrade/upgrade_common.go`. So, replacing those two function invocations should be trivial (although there are caveats, as mentioned in the [docs](https://golang.org/pkg/os/#Executable)).

imsodin:
`osext.Executable()` used `os.Executable()` in the background since go1.8 and I don't think anyone will build new versions on <=go1.7, so this is essentially changes nothing.


calmh:
While I suspect this is fine, can you double check that the stdlib os.Executable supports all our platforms (or that kardinanos doesn't use another implementation on any platform) and test this by hand at least once?

Breaking auto upgrades is painful to recover from.

imsodin:
@calmh Everything osext supported is also supported by os and I just performed an update from 0.14.46 with this commit cherry-picked to latest rc version.

imsodin:
Just linux - I don't have any other platforms available.

calmh:
Code is pretty clear, it already just calls os.Executable.

PR #4951, opened 2018-05-14
lib/model: Also wait for ItemFinished in TestDeregister...

imsodin:
The deregistering tests have historically not been flaky, however since about 11. May, `TestDeregisterOnFailInCopy` fails in about 50% of cases on the Windows build server only:  
https://build.syncthing.net/project.html?tab=testDetails&testNameId=-1067371058420929398&projectId=Syncthing&buildTypeId=Syncthing_BuildWindows  
It fails after an arbitrary sleep period of 100ms to let the finisher do its job. <strike>Instead of</strike> In addition to this sleep period, it is now polling 1s for the `ItemFinished` event.

### Testing

Should be run repeatedly on the windows build server just to get a (statistically insignificant) sample instead of a coin toss.

### DIFF
--- a/lib/model/folder_sendrecv_test.go
+++ b/lib/model/folder_sendrecv_test.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/syncthing/syncthing/lib/config"
 	"github.com/syncthing/syncthing/lib/db"
+	"github.com/syncthing/syncthing/lib/events"
 	"github.com/syncthing/syncthing/lib/fs"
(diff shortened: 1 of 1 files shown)


imsodin:
So a whooping sample of [three consecutive runs on windows did not produce any test failures](https://build.syncthing.net/project.html?projectId=Syncthing&testNameId=-1067371058420929398&tab=testDetails) - whether that lasts with this addition I can't say, but I don't see any downsides - so lgtm from me.

PR #4958, opened 2018-05-17
 lib/db: Add prefix to track locally needed files 

imsodin:
### Purpose

> While I'm doing that you can think about the other half of the equation. :) For every incoming index update we schedule a pull. The pull looks through the database for things that have changed...
https://github.com/syncthing/syncthing/pull/4906#issuecomment-384872057 

It took me a while longer than it took Jakob to test it, but here it is.

The needKey->nil bucket tracks locally needed files and is update whenever files are updated. Thus traversing local needs does no longer need to go over the entire global bucket detecting what we actually need, it just walks this new bucket. For remote devices I don't do the same for less complexity and smaller db size, as lookups of remote needs are less frequent - that works just like before.

Some of the refactoring is due to an earlier version of the new test that worked without the jsons db and thus needed access to parts of `readWriteTransaction.updateGlobal`. In my opinion the refactor is still worthwhile.

The l -> fl/vl renaming is to not override the logger variable.

### Testing

I added a test for the db migration from schema 0 to 3.

Edit:  
I didn't the database schema change in the local flags PR. Just ignore that aspect in the review, depending on timeline I will adjust this PR accordingly or do a separate PR for that shuffling.


calmh:
Sounds awesome; will take some pondering to review properly as it's important, so bear with me. :)

lib/db/leveldb_transactions.go
```diff
@@ -85,107 +86,89 @@ func (t readWriteTransaction) insertFile(fk, folder, device []byte, file protoco
...
+		l.Debugln("update global; same version, global unchanged")
+		return false
+	}
 
-				if i == 0 {
-					// Keep the current newest file around so we can subtract it from
-					// the metadata if we replace it.
-					oldFile, hasOldFile = t.getFile(folder, fl.Versions[0].Device, name)
-				}
+	if insertedAt == 0 {
```
calmh: One thing to take into account here that I had to fix up in `updateGlobal` for receive only, and that I *think* can be an issue already without the receive only stuff...

If we update a local file to a *lower* version than what is currently in there, what is the head of the version list may change without the insert being at index zero. For receive only folders this happens when the user does a "revert", but I think it could happen already today when a file gets unignored as we then tweak the version. Whether the new version becomes the head of the list or not depends on conflict resolution, and if it does *not* win the resolution it'll be inserted somewhere in the middle.

So, basically, if the version list is `{F1, F2, F3}` where F1 is our local file and the need is nil (we have the latest version), an update after unignore (or the upcoming revert operation) can make it `{F2, F1', F3}` or `{F2, F3, F1'}`, where `F1'` is the version-tweaked unignored file.

We need to make F2 the needed version then, despite the insert happening somewhere further down the list.

After receive only folders it becomes worse, although that's not your problem at the moment. :) The head of the list can then be a version we need, which gets withdrawn due to a change happening on that receive only device, and potentially another version from another device then becomes promoted to the head of the list and becomes the *new* needed file.

imsodin: Definitely true, `removedAt == 0` needs to be handled as well - thanks for catching that.
Doesn't that mean that global meta counting is broken already in this scenario?

calmh: Yeah, I think so. I have it fixed but it's on top of a couple of other things, and of course it interferes with this PR.

AudriusButkevicius:
Again, db stuff magic, 3 or 4 years later I still don't understand it well, so I'll leave it for you guys.

calmh:
@imsodin 👍 please do

PR #4962, opened 2018-05-22
all: Fix FS watcher restarting and web UI indication (fixes #4923)

### LINKED ISSUE #4923: fsWatcher is not reinstated after a syncthing folder reappeared on the filesystem
Reproducer
1. Create a syncthing folder on the filesystem with the file watcher enabled as it is default now.
2. Stop Syncthing
3. move the folder away (alternatively unmount the device representing the folder)
4. Start Syncthing
5. Open Web UI
6. Error message appears (folder marker missing)
7. Move the folder back in the original place (or mount the device thats representing the folder)
8. Syncthing continues working on that folder (shows "current" instead of "error" again)
9. Web UI now displays the notice "file watcher could not be installed"

Workaround solution
Restart Syncthing, file watcher and folder is okay then.

Syncthing Version: v0.14.47
OS Version: Windows 10 x64
Browser Version: Chrome 63

imsodin:
The main change is actually initializing `restartWatchChan` to make it restart when recovering from an error.

When testing I realized two more related web UI issues:  
1. When a folder is stopped, the scanning tooltip still indicates that the watcher has an error and is trying to start again - which it is not.
2. The watcher error is propagated to the UI in the folder summary, which is not triggered when the watcher error is changed. The only way I see to fix this is introducing a new event for that. A dedicated event seems to be overkill, but no other event fits. If someone thinks of a simpler way to get that info to the web UI, please say so.


cmd/syncthing/summaryservice.go
```diff
@@ -105,14 +105,14 @@ func (c *folderSummaryService) listenForUpdates() {
...
 
+					c.foldersMut.Lock()
 					select {
 					case c.immediate <- folder:
-						c.foldersMut.Lock()
 						delete(c.folders, folder)
-						c.foldersMut.Unlock()
-
 					default:
+						c.folders[folder] = struct{}{}
```
AudriusButkevicius: Why this change?

lib/model/folder.go
```diff
@@ -338,6 +351,18 @@ func (f *folder) startWatchAsync(ctx context.Context, ignores *ignore.Matcher) {
...
+					data["from"] = prevErr.Error()
+				}
+				if err != nil {
+					data["to"] = err.Error()
+				}
+				events.Default.Log(events.FolderWatcherStateChanged, data)
+			}
 			if err != nil {
 				if prevErr == errWatchNotStarted {
 					l.Warnf("Failed to start filesystem watcher for folder %s: %v", f.Description(), err)
```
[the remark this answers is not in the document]
imsodin: That also just a warning? I don't know of an existing way to display something like a warning but only when active.

cmd/syncthing/summaryservice.go
imsodin: Because I had to change a few lines up and this caught my eye as quite wrong: If you want to send a folder summary immediately, but can't, it doesn't make sense to not send one at all.

lib/model/folder.go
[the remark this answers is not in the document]
AudriusButkevicius: The GUI insecure stuff does it. Essentially conditional elements visible based on mode[folder].watchErr, or just remove rhe warning or word it that it's temporary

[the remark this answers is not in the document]
imsodin: Now I see, I was looking at the wrong message. Removing the warning entirely seems bad, as the failure is then only visible when the folder is expanded. And if the rescan interval is high, that means the folder won't pick up changes. Having a top banner dedicated to those errors is very clear, but also potentially overkill. I created a draft (definitely needs "design" polish), with an explanation text and then a list of all folder<->error pairs (here just one):
screenshot_2018-05-25_10-57-25
Comments?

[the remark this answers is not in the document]
AudriusButkevicius: Fa watcher is not a very user understandable name, but looks ok.

gui/default/index.html
```diff
@@ -257,6 +257,33 @@ <h3 class="panel-title">
       </div>
     </div>
 
+    <!-- Panel: FS watcher errors -->
+
+    <div ng-if="sizeOf(fsWatcherErrorMap()) > 0" class="row">
```
AudriusButkevicius: you can assign the return value of that on the div temporarily I think, so there wouldn't be 2 calls to this.

imsodin:
The new undismissable notification instead of warnings looks like this:

screenshot_2018-05-28_23-39-26

Personally I am not terribly happy about this, design wise - both how it actually looks and how much place it takes up. On the other hand there needs to be some kind of notification without expanding folders.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -240,6 +240,13 @@ angular.module('syncthing.core')
             }
         });
 
+        $scope.$on(Events.FOLDER_WATCHER_STATE_CHANGED, function (event, arg) {
```
AudriusButkevicius: What if the event rolled off? How often is this re-emit? Why not have this part of some folder completion status call or some shit like that?

AudriusButkevicius: Perhaps part of folder summary service?

imsodin: It already is part of that and updated when incoming as well. Unfortunately a fs watcher error does not trigger a new folder summary and the only way it can trigger is by emitting an event. And if there is an event anyway, I thought I might as well use it in the UI. This trigger could however be removed without big drawbacks (folder summaries are probably a bit slower generally and might even get delayed by a minute (?) worst case).

gui/default/index.html
imsodin: It looks like that doesn't work because ng-init is only evaluated once on startup, then never again.

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: Syncthing stores last 1k or so events. If you have a lot of tiny files being downloaded, FileStarted, FileFinished will force your event to roll off and it will never be visible.

Why can't your watcher emit an event, but that even be picked up by the summary service and the summary service sets the error on the summary, and then you check if the error is in the summary.

imsodin: That happens too, `scope.model` is the folder summary and contains the fs watcher error. This event handling is just to speed things up in case the folder summary is delayed. I'll just remove it.

AudriusButkevicius: It's never delayed by more than 2 or so seconds, Adding an extra event because of that seems like overkill?

imsodin: It's removed since 4 days ;)

lib/events/events.go
```diff
@@ -44,6 +44,7 @@ const (
 	FolderScanProgress
 	FolderPaused
 	FolderResumed
+	FolderWatcherErrorStateChanged
```
AudriusButkevicius: Can we call this FolderWatchStateChanged, just in case there are more than one stage in the future, and we could reuse the event.

imsodin: Agreed.  
And I am not complaining, just smiling slightly: https://github.com/syncthing/syncthing/pull/4962#discussion_r190689729

AudriusButkevicius: ha.

I have to find things to pick on, otherwise what's the purpose of the review?

AudriusButkevicius:
Ah, the title of the dialog sucks, people don't know what FS is.

imsodin:
@AudriusButkevicius I wrote it out - good to merge? (I want to get rid of some open PRs :) )

lib/model/folder.go
```diff
@@ -78,9 +79,10 @@ func newFolder(model *Model, cfg config.FolderConfiguration) folder {
 
 		pullScheduled: make(chan struct{}, 1), // This needs to be 1-buffered so that we queue a pull if we're busy when it comes.
 
-		watchCancel: func() {},
-		watchErr:    errWatchNotStarted,
-		watchErrMut: sync.NewMutex(),
+		restartWatchChan: make(chan struct{}, 1),
```
AudriusButkevicius: Where is this channel used?

imsodin: https://github.com/imsodin/syncthing/blob/aa470b5303f6774cd4eaad7d0f71efb51c4f97d2/lib/model/folder.go#L166

This is like *the* main fix of this PR, the rest is UI cosmetics :)

AudriusButkevicius: It's not used in this pr tho, so I am puzzled.

imsodin: The problem was, that before `restartWatchChan` never was initialized, so always was `nil` -> the watch never restarted after a failure at any point, even though everything was in place already.

imsodin:
@calmh Given the RC tomorrow, I'd like to merge - fine?

PR #4970, opened 2018-05-24
lib/scanner: Copy execute bits from previous version on Windows (fixes #4969)

### LINKED ISSUE #4969: Syncthing overrides linux execute file mode when synching from Windows
### THE ISSUE
Syncthing erases the execute file mode from script files after syncing the file from Windows, making users always have to remember to chmod the file whenever it's changed in the Windows machine.

### REPRO STEPS
* Set up syncthing between a Linux machine and a Windows machine.
* On Linux, create a script file and chmod it to enable execute.
* Wait for the file to be sync'ed on Windows, and then edit the file's contents.
* When the file edits are sync'ed to the Linux machine, observe the file mode.

###EXPECTED
 The execute file mode remains.

###ACTUAL
The execute file mode is erased.

AudriusButkevicius:


imsodin:
Seems right. In addition I think that what you did for windows should also be done when ignore permission is set on the folder. I.e. don't update the permissions from disk, but keep the pre-existing permissions - it seems counter-intuitive to use permissions when they should be ignored.

AudriusButkevicius:
They just don't get applied if ignore permissions is set. i think its ok to send them across.

imsodin:
It still feels inconsistent that the same instance ignores permissions on incoming updates, but changes permissions on "local updates"/scanning. If I enable an ignore permission setting, I expect this device to not interact with permissions.
Imagine a cluster with two unix and one windows device, windows ignores permissions, the other two don't. Lets say a unix device creates a file with 600 permissions. That's propagated fine everywhere with correct permission bits on the second unix device. If that file later gets modified on windows, it will get 666 permissions. If windows would realize, that it's ignore permissions and keep previous permission bits, it would still successfully update the file and the 600 permission would survive.

AudriusButkevicius:
Why would it get 666? Only if the file is executable on windows based on ext it would get that, even then, it would set the permissions and then set haspermissioms to false, which on the other end would not do a chmod I expect?

PR #4971, opened 2018-05-25
gui: Default theme (widescreen)

uok:
Additional default theme that uses 90% of screen width on large resolutions.

### DIFF
--- a/gui/default (widescreen)/assets/css/theme.css
+++ b/gui/default (widescreen)/assets/css/theme.css
@@ -1,0 +1,11 @@
+/**
+
+Default theme (widescreen)
+
+**/
+
+@media (min-width: 1200px) {
+  .container {
+    width: 90%;
+  }
+}


AudriusButkevicius:
Pic?

wweich:
Why not change the `container` to `container-fluid` in the default theme. `container-fluid` is always full width, with the same responsiveness.

uok:
@AudriusButkevicius do you mean a screenshot? I don't have a widescreen monitor so a screenshot from theme widescreen would look exactly like theme regular
@wweich The idea was to have the default theme (centered gui with margins left/right) and widescreen theme (90% width). There is no need for editing html.

AudriusButkevicius:
Why don't we use the fluid stuff?

uok:
@wweich thanks for the screenshots and the info, I update the media query.
@AudriusButkevicius Then we would need to change the index.html file as well because right now you can only change colors/layouts of existing elements (CSS classes and HTML tags). In the future then somebody needs to update it every time a new element is added to the default index.html - I'm not sure this is worth the work.

AudriusButkevicius:
As far as I understand the fluid stuff could be part of the default ui, as it should not change much.

PR #4973, opened 2018-05-28
gui: Restrict shown decimals and restrict size of header columns

imsodin:
This is more or less a continuation of the number localization in  #4902. Now it's no longer about localization but a reasonable amount of digits. The new "policy" is that only show three significant digits if there are decimals and maximally two decimals. I.e. don't change something like 24378273, and display 14.2323 as 14.2, 4.2323 as 4.23 and 0.21312 as 0.21. For percentages I even went to just 2 significant digits, because percentages just feel like two digit information to me (because 0-99 (please ignore 0-9)). That saves space and doesn't hurt my sensitive eyes with insignificant digits :P  

This still wasn't enough to prevent ellipsis on big total states, so I additionally gave some of the white space from the header column to the data column.

screenshot_2018-05-28_21-23-38
screenshot_2018-05-28_21-25-38

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -360,8 +360,3 @@ ul.three-columns li, ul.two-columns li {
 .fancytree-ext-table {
     width: 100% !important;
 }
-
-[id^="folder-"] th,
-[id^="device-"] th {
-    width: 200px;
-}
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -360,3 +360,8 @@ ul.three-columns li, ul.two-columns li {
 .fancytree-ext-table {
     width: 100% !important;
 }
+
+[id^="folder-"] th,
+[id^="device-"] th {
+    width: 200px;
+}
--- a/gui/default/syncthing/core/localeNumberFilter.js
+++ b/gui/default/syncthing/core/localeNumberFilter.js
@@ -1,9 +1,6 @@
 angular.module('syncthing.core')
     .filter('localeNumber', function () {
-        return function (input, decimals) {
-            if (typeof(decimals) !== 'undefined') {
-                return input.toLocaleString(undefined, {maximumFractionDigits: decimals});
-            }
+        return function (input) {
             return input.toLocaleString();
         };
     });
--- a/gui/default/syncthing/core/percentFilter.js
+++ b/gui/default/syncthing/core/percentFilter.js
@@ -1,0 +1,15 @@
+angular.module('syncthing.core')
+    .filter('percent', function () {
+        return function (input) {
+            // Prevent 0.00%
+            if (input === undefined || input < 0.01) {
+                return 0 + '%';
+            }
+            // Hard limit at two decimals
+            if (input === undefined || input < 0.1) {
+                return input.toLocaleString(undefined, {maximumFractionDigits: 2}) + '%';
+            }
+            // "Soft" limit at two significant digits (e.g. 1.2%, not 1.27%)
+            return input.toLocaleString(undefined, {maximumSignificantDigits: 2}) + '%';
+        };
+    });
--- a/gui/default/syncthing/core/percentFilter.js
+++ b/gui/default/syncthing/core/percentFilter.js
@@ -6,7 +6,7 @@ angular.module('syncthing.core')
                 return 0 + '%';
             }
             // Hard limit at two decimals
-            if (input === undefined || input < 0.1) {
+            if (input < 0.1) {
                 return input.toLocaleString(undefined, {maximumFractionDigits: 2}) + '%';
             }
             // "Soft" limit at two significant digits (e.g. 1.2%, not 1.27%)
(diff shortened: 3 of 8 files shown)


wweich:
Is the percent filter build-in?

Setting maximumSignificantDigits to 3 means, that 100GiB and more will have no decimals and that even non-decimal-digits will be "removed":
- between 1000 and 1009 -> 1000
- between 1010 and 1019 -> 1010
- between 1020 and 1023 -> 1020

imsodin:
Nope, my "creation" - but I apparently forgot to add it - thanks for catching that.

That's why I don't just globally set maximumSignificantDigits to 3, only if there are maximally 3 digits.. However you just brought me to notice that for the binary filter numbers 1000-1023 may exist, which would indeed get rounded wrongly.

wweich:
> For percentages I even went to just 2 significant digits

Setting `maximumSignificantDigits` to `2` for percent will still produce something like `0.00000000023` as the `2` and `3` are the two significant numbers of `0.0000000002312`.

> Can angular-modules/filters share scopes somehow? Would be nice to share code between binary and metric filter.

You could (like the old `decimals` function, that I removed) add a global function to `app.js` with 2 parameters (the number and a bool for binary or not) and just does the necessary divisions to get the input below 1024 or 1000, so the filter just formats the output.

imsodin:
@wweich Please give me a grace period until you actually see that function :) I am aware of that particular problem and have addressed it. I do *not* just set maximumSignificantDigits to 2, I cap it (like everything) to two decimals max regardless.

Thanks for the app.js tip.

gui/default/assets/css/overrides.css
```diff
@@ -360,3 +360,8 @@ ul.three-columns li, ul.two-columns li {
 .fancytree-ext-table {
     width: 100% !important;
 }
+
+[id^="folder-"] th,
+[id^="device-"] th {
+    width: 200px;
+}
```
AudriusButkevicius: Why is this needed?

gui/default/syncthing/core/localeNumberFilter.js
```diff
@@ -1,9 +1,6 @@
...
-        return function (input, decimals) {
-            if (typeof(decimals) !== 'undefined') {
-                return input.toLocaleString(undefined, {maximumFractionDigits: decimals});
-            }
+        return function (input) {
             return input.toLocaleString();
         };
-    });
\ No newline at end of file
+    });
```
AudriusButkevicius: Can't we just call `{{ x.toLocaleString() }}` ?
This doesn't seem to do much?

gui/default/syncthing/core/percentFilter.js
```diff
@@ -0,0 +1,15 @@
...
+    .filter('percent', function () {
+        return function (input) {
+            // Prevent 0.00%
+            if (input === undefined || input < 0.01) {
+                return 0 + '%';
+            }
+            // Hard limit at two decimals
+            if (input === undefined || input < 0.1) {
+                return input.toLocaleString(undefined, {maximumFractionDigits: 2}) + '%';
+            }
```
AudriusButkevicius: the `input === undefined` makes no sense

gui/default/syncthing/core/percentFilter.js
```diff
@@ -0,0 +1,15 @@
...
+            // "Soft" limit at two significant digits (e.g. 1.2%, not 1.27%)
+            return input.toLocaleString(undefined, {maximumSignificantDigits: 2}) + '%';
```
AudriusButkevicius: So what happens with 100%?

gui/default/assets/css/overrides.css
imsodin: Even with the space gained by not showing some decimals, I still got truncated global states and at the same time there is unused white space on the left. With this, the white space is reduced, resulting in not truncated data.

gui/default/syncthing/core/percentFilter.js
imsodin: Nothing, it is already rounded.

gui/default/syncthing/core/localeNumberFilter.js
imsodin: It doesn't, but it is chained with `| alwaysNumber` which isn't compatible with `.toLocaleString`(?)

gui/default/assets/css/overrides.css
AudriusButkevicius: Where is the whitespace on the left coming from? Padding/margin?
Also, perhaps min-width then?

imsodin: The entire thing is a total can of worms, I was googling in dispair, because I didn't want to believe that it really is this bad: The table is fixed, because you can't set any size limits on auto, i.e. it exceeds its parent divs boundaries. min/max-width doesn't work on tables/td/th. Fixed with width set is the best solution I could find.

AudriusButkevicius: I hope this doesn't screw up mobile.

imsodin: Possibly. I don't know and didn't test.

AudriusButkevicius: If you own one, you could just navigate to the UI to your machine from your mobile.

imsodin: @AudriusButkevicius I removed this. After experimenting some more it became clear, that a good solution isn't simple and it's not directly related to the number mangling in this PR.

PR #4974, opened 2018-05-28
lib/config, lib/model: Handle shared with information in config (fixes #4870)

### LINKED ISSUE #4870: Don't show a warning when a file from a paused folder is requested

imsodin:
Original/main purpose is to actually fix #4870. In the previous PR I tried to deduce whether the folder is paused from the information within the model. That clearly fails, because paused folders aren't in the model. 

Paused folders also caused problems elsewhere in the past (cluster config for one). `folderDevices` and `deviceFolders` is fully redundant and additionally they track information about which folder is shared with whom -> that's actually config business (again, paused folders). So I moved that functionality to wrapper.go and adapted all affected sections in model.go (sometimes adding the distinction between a unshared vs just paused folder).

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -829,11 +829,10 @@ func (m *Model) handleIndex(deviceID protocol.DeviceID, folder string, fs []prot
 
 	l.Debugf("%v (in): %s / %q: %d files", op, deviceID, folder, len(fs))
 
-	if !m.cfg.IsSharedWith(folder, deviceID) {
+	if cfg, ok := m.cfg.Folder(folder); !ok || !cfg.SharedWith(deviceID) {
 		l.Infof("%v for unexpected folder ID %q sent from device %q; ensure that the folder exists and that this device is selected under \"Share With\" in the folder configuration.", op, folder, deviceID)
 		return
-	}
-	if cfg, ok := m.cfg.Folder(folder); ok && cfg.Paused {
+	} else if cfg.Paused {
 		l.Debugf("%v for paused folder (ID %q) sent from device %q.", op, folder, deviceID)
 		return
 	}
@@ -910,7 +909,7 @@ func (m *Model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
 	m.fmut.Lock()
 	var paused []string
 	for _, folder := range cm.Folders {
-		if !m.cfg.IsSharedWith(folder.ID, deviceID) {
+		if cfg, ok := m.cfg.Folder(folder.ID); !ok || !cfg.SharedWith(deviceID) {
 			if m.cfg.IgnoredFolder(folder.ID) {
 				l.Infof("Ignoring folder %s from device %s since we are configured to", folder.Description(), deviceID)
 				continue
@@ -922,14 +921,10 @@ func (m *Model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
 			})
 			l.Infof("Unexpected folder %s sent from device %q; ensure that the folder exists and that this device is selected under \"Share With\" in the folder configuration.", folder.Description(), deviceID)
 			continue
-		}
-
-		if folder.Paused {
+		} else if folder.Paused {
 			paused = append(paused, folder.ID)
 			continue
-		}
-
-		if cfg, ok := m.cfg.Folder(folder.ID); ok && cfg.Paused {
+		} else if cfg.Paused {
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


lib/model/model.go
```diff
@@ -1136,53 +1104,57 @@ func (m *Model) handleIntroductions(introducerCfg config.DeviceConfiguration, cm
...
+func (m *Model) handleDeintroductions(introducerCfg config.DeviceConfiguration, cm protocol.ClusterConfig, foldersDevices config.FolderDeviceSet) bool {
 	changed := false
-	foldersIntroducedByOthers := make(folderDeviceSet)
+	devicesNotIntroduced := make(map[protocol.DeviceID]struct{})
 
+	folders := m.cfg.FolderList()
 	// Check if we should unshare some folders, if the introducer has unshared them.
-	for _, folderCfg := range m.cfg.Folders() {
-		folderChanged := false
+	for i := range folders {
```
AudriusButkevicius: Just `for _, folderCfg := range folders` ?

lib/model/model_test.go
```diff
@@ -1286,7 +1286,8 @@ func TestAutoAcceptPrefersLabel(t *testing.T) {
 			},
 		},
 	})
-	if fcfg, ok := wcfg.Folder(id); !ok || !m.folderSharedWith(id, device1) || !strings.HasSuffix(fcfg.Path, label) {
+	if fcfg, ok := wcfg.Folder(id); !ok || !wcfg.IsSharedWith(id, device1) || !strings.HasSuffix(fcfg.Path, label) {
+
```
AudriusButkevicius: new blank line

lib/model/model.go
AudriusButkevicius: Actually, you are right, but I'd just modify folders[i] to cause less confusion

lib/model/model.go
```diff
@@ -935,28 +909,23 @@ func (m *Model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
...
 				"device":      deviceID.String(),
 			})
 			l.Infof("Unexpected folder %s sent from device %q; ensure that the folder exists and that this device is selected under \"Share With\" in the folder configuration.", folder.Description(), deviceID)
 			continue
+		} else if folder.Paused {
+			paused = append(paused, folder.ID)
+			continue
+		} else if cfg.Paused {
+			continue
 		}
```
AudriusButkevicius: All of these end up with continues, can we break them into separate ifs, so that we don't have to think about the previous if checking if next one will be hit.

lib/model/model.go
```diff
@@ -1106,15 +1074,10 @@ func (m *Model) handleIntroductions(introducerCfg config.DeviceConfiguration, cm
 			if _, ok := m.cfg.Devices()[device.ID]; !ok {
 				// The device is currently unknown. Add it to the config.
 				m.introduceDevice(device, introducerCfg)
-				changed = true
```
AudriusButkevicius: This was lost somehow?

imsodin: Not lost, removed. It is superfluous as will already be set to true here:
https://github.com/syncthing/syncthing/pull/4974/files/dfa19b40a8c03563a50dd7278f6b5114768e35aa#diff-1df9805b2381b6288b220226f82d5dc9R1090

AudriusButkevicius: That doesn't show anything meaningful.

AudriusButkevicius: ah it's at the bottom, ok.

imsodin: I seriously need to stop linking code and start copying it - these github-diff-links are so broken.

imsodin:
@calmh Would you like to review this? If not I'd like to merge :)

@AudriusButkevicius Could you re-approve please?

calmh:
I glanced through it, I'll take this one on faith.

Issue #4956, opened 2018-05-16
Windows/Linux/Solaris cluster problematic names with directories and duplication


whorfin:
I had a syncthing directory containing many directories, this particular branch being problematic:
`books/0purchased/`
   [many files, subdirectories, etc under here]

I have a cluster of Mac, Windows, Linux RHEL 7 machines, and one Solaris [OmniOS CE] box.
After wondering why synchronization was taking so long, I noticed that some of the linux/solaris machines were also maintaining a `0PURCH~1` directory, which was a sort-of duplicate of `0purchased`.
Windows and Mac clients only ever say the `0purchased` directory
A back-and-forth sync-storm caused changes in `0purchased` to bounce back and forth through the `0PURCH~1` copy, as far as the non-unixy systems saw it.

On the Windows box, I renamed the `0purchased` directory to `_purchased`.
This resulted in `_purchased` being added to all the unixy systems [with no further weird named subdir copies], but `0purchased` began to be pushed back to the Windows box, presumably from the `0PURCH~1` copy.

The whole experience was spooky, and raises the question as to whether normalization for windows-style names [the `blah~1` thing smells a lot like 8.3 SFN naming nonsense] is missing some cases...

Cheers

### Version Information

Syncthing Version: v0.14.47
OS Version: Windows 10 / RHEL 7.2 [linux] / OmniOSCE-r151022 [Solaris]

calmh:
I have no idea how the synthesized 8.3 name popped up in there, but it's correct that we do not attempt to normalize this and figure out what the "actual" name is.

AudriusButkevicius:
There is already a ticket for 8.3 name resolution.
I think we get those names from the watcher.

calmh:
That's kind of bad though? It would be good to know under what circumstances that happen, because it fucks things up pretty royally.

AudriusButkevicius:
My guess is that the kernel bubbles up whatever path the underlying apllication has written to. So if your midnight commander used 8.3 path, this is what you get notified about. There are some win32 api calls to resolve these paths I think.

This is just a theory, yet this ticket is no way different from the existing 8.3 ticket, so I don't see why this should stay open.

AudriusButkevicius:
This doesn't fuck things any worse than the case difference nonsense.

calmh:
[...]

PR #4975, opened 2018-05-28
lib/fs: Resolve 8.3 filenames from watcher (ref #3800)

imsodin:
On windows, check if paths coming from the FS watcher are 8.3 ones and convert to "normal", similar to how @canton7 did it for synctrayzor: https://github.com/syncthing/syncthing/issues/4956#issuecomment-389580448

### DIFF
--- a/lib/fs/basicfs_windows.go
+++ b/lib/fs/basicfs_windows.go
@@ -159,13 +159,19 @@ func (f *BasicFilesystem) resolveWin83(absPath string) string {
 	if in, err := syscall.UTF16FromString(absPath); err != nil {
 		out := make([]uint16, 4*len(absPath)) // *2 for UTF16 and *2 to double path length
 		if n, err := syscall.GetLongPathName(&in[0], &out[0], uint32(len(out))); err == nil {
+			l.Infoln("first getLong, n:", n, "len(out):", len(out))
 			if n <= uint32(len(out)) {
 				return syscall.UTF16ToString(out[:n])
 			}
 			out = make([]uint16, n)
 			if _, err = syscall.GetLongPathName(&in[0], &out[0], n); err == nil {
+				l.Infoln("second getLong, n:", n, "len(out):", len(out))
 				return syscall.UTF16ToString(out)
+			} else {
+				l.Infoln("second getLong, err:", err)
 			}
+		} else {
+			l.Infoln("first getLong, err:", err)
 		}
 	}
 	// Failed getting the long path. Return the part of the path which is
--- a/lib/fs/basicfs_windows.go
+++ b/lib/fs/basicfs_windows.go
@@ -173,6 +173,8 @@ func (f *BasicFilesystem) resolveWin83(absPath string) string {
 		} else {
 			l.Infoln("first getLong, err:", err)
 		}
+	} else {
+		l.Infoln("utf16 from string, err:", err)
 	}
 	// Failed getting the long path. Return the part of the path which is
 	// already a long path.
--- a/lib/fs/basicfs_windows.go
+++ b/lib/fs/basicfs_windows.go
@@ -150,3 +150,31 @@ func (f *BasicFilesystem) Roots() ([]string, error) {
 
 	return drives, nil
 }
+
+func (f *BasicFilesystem) resolveWin83(absPath string) string {
+	if !strings.Contains(absPath, '~') {
+		return absPath
+	}
+	if in, err := windows.UTF16FromString(absPath); err != nil {
+		out = make([]uint16, 2*len(p))
+		n := uint32(len(b))
+		for {
+			n, err = windows.GetLongPathName(&in[0], &out[0], n)
+			if err != nil {
+				break
+			}
+			if n <= uint32(len(out)) {
+				return windows.UTF16ToString(out[:n])
+			}
+			out = make([]uint16, n)
+		}
+	}
+	// Failed getting the long path. Return the part of the path which is
+	// already a long path.
+	for absPath = filepath.Dir(absPath); strings.HasPrefix(absPath, f.rootSymlinkEvaluated); absPath = filepath.Dir(absPath) {
+		if !strings.Contains(absPath, '~') {
+			return absPath
+		}
+	}
+	return f.rootSymlinkEvaluated
+}
--- a/lib/fs/basicfs_windows.go
+++ b/lib/fs/basicfs_windows.go
@@ -179,12 +179,8 @@ func (f *BasicFilesystem) resolveWin83(absPath string) string {
 }
 
 func isMaybeWin83(absPath string) bool {
-	parts := strings.Split(absPath, "~")
-	// Return true unless splitting occurs due to "~syncthing~"
-	for i := 1; i < len(parts); i += 2 {
-		if parts[i] != WindowsTempPrefix[1:len(WindowsTempPrefix)-1] {
-			return true
-		}
+	if !strings.Contains(absPath, "~") {
+		return false
 	}
-	return false
+	return strings.Contains(strings.TrimPrefix(absPath, WindowsTempPrefix), "~")
 }
--- a/lib/fs/basicfs_windows.go
+++ b/lib/fs/basicfs_windows.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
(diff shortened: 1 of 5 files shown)


lib/fs/basicfs_windows.go
```diff
@@ -150,3 +150,31 @@ func (f *BasicFilesystem) Roots() ([]string, error) {
...
+
+func (f *BasicFilesystem) resolveWin83(absPath string) string {
+	if !strings.Contains(absPath, '~') {
+		return absPath
+	}
+	if in, err := windows.UTF16FromString(absPath); err != nil {
+		out = make([]uint16, 2*len(p))
+		n := uint32(len(b))
+		for {
+			n, err = windows.GetLongPathName(&in[0], &out[0], n)
```
AudriusButkevicius: Where is this coming from?

imsodin: This is a left-over from a zealous fit in which I tried to replace syscall with x/sys wherever possible (due to that weird freeze policy that I admittedly don't understand fully). Then later I decided it was likely a bad idea and definitely not something to be mixed into this 8.3 PR. I can still post a PR with the replacement if you consider it (potentially) worthwhile.

AudriusButkevicius:
Can we have a test to prove that the function works?

AudriusButkevicius:
I assume this will also fix cases where the old inotify sends us 8.3 paths?

imsodin:
No it doesn't. I consider sending BS paths over the rest API as shooting your own foot and thus not our/Syncthing's concern.

I didn't write a test because of two reasons: 1. The windows API call requires the file to exist, so you'd actually have to create files such that you can access them with 8.3. 2. I don't know how 8.3 names work and don't intend to learn it. My google-fu wasn't good enough to turn up an example list. If someone could enlighten me whether there is anything special to consider in point 1 and point me to some examples, I'd be happy to create a test.

canton7:
From the [GetLongPathName docs](https://msdn.microsoft.com/en-us/library/windows/desktop/aa364980(v=vs.85).aspx):

> On many file systems, a short file name contains a tilde (~) character. However, not all file systems follow this convention. Therefore, do not assume that you can skip calling GetLongPathName if the path does not contain a tilde (~) character.

imsodin:
@canton7 I chose to ignore that, because of laziness and a nice excuse in the form of [your implementation](https://github.com/canton7/SyncTrayzor/blob/9f8d4382f057f81fd0337003920b38955076d7ce/src/SyncTrayzor/Services/FileWatcher.cs#L237) ;)
```
if (pathExists && path.Contains("~"))
    path = this.GetLongPathName(path);
```

Are short paths without a tilde really a concern or is this just FUD?

lib/fs/basicfs_windows.go
```diff
@@ -150,3 +151,31 @@ func (f *BasicFilesystem) Roots() ([]string, error) {
 
 	return drives, nil
 }
+
+func (f *BasicFilesystem) resolveWin83(absPath string) string {
+	if !strings.Contains(absPath, "~") {
+		return absPath
+	}
+	if in, err := syscall.UTF16FromString(absPath); err != nil {
+		out := make([]uint16, 2*len(absPath))
```
canton7: Worth allocating more space? Due to the test on L156, we know that we've got an 8.3 path and we're going to end up with a long path. Since `out` is a UTF-16 buffer, it has 2 bytes per character. Allocating `2*len(absPath)` means that `out` is the same length as `in`, i.e. you'll have to reallocate if `out` is even 1 char longer than `in`. Since we know that `out` will definitely be longer than `in`, this means we'll always reallocate. Perhaps this should be `2*len(in)`?

lib/fs/basicfs_windows.go
```diff
@@ -150,3 +151,31 @@ func (f *BasicFilesystem) Roots() ([]string, error) {
...
+		n := uint32(len(out))
+		for {
+			n, err = syscall.GetLongPathName(&in[0], &out[0], n)
```
canton7: Nit: `n` has two meanings: it starts off being the size of the buffer, and then becomes the return value from `GetLongPathName`.

canton7: Why not use `uint32(len(out))` instead of `n` as the last param?

canton7:
> @canton7 I chose to ignore that, because of laziness and a nice excuse in the form of your implementation ;)

> if (pathExists && path.Contains("~"))
?     path = this.GetLongPathName(path);

> Are short paths without a tilde really a concern or is this just FUD?

Yeah, I'm guilty of the same shortcut. Just spotted that in the docs, thought I'd call it out in case anyone cared. I don't have a better suggestion 😄

lib/fs/basicfs_windows.go
```diff
@@ -150,3 +151,31 @@ func (f *BasicFilesystem) Roots() ([]string, error) {
...
+			if err != nil {
+				break
+			}
+			if n <= uint32(len(out)) {
+				return syscall.UTF16ToString(out[:n])
+			}
+			out = make([]uint16, n)
```
canton7: The returned `n` is:

> The size of the buffer lpszLongPath points to, in TCHARs.

Apparently `TCHAR` can mean `char` or `wchar_t` depending on whether your project is compiled as unicode or not. I'm not sure what Go does here, but it's possible that it might actually want a buffer with `2*n` *bytes*. 

If it does, this combined with the loop will throw this function into an infinite loop. I think one of those test things might be a good idea here...

lib/fs/basicfs_windows.go
canton7: According to this docs, the last param is:

> The size of the buffer lpszLongPath points to, in TCHARs.

Apparently TCHAR can mean char or wchar_t depending on whether your project is compiled as unicode or not. I'm not sure what Go does here, but it's possible you're telling `GetLongPathName` that it's got double the amount of space available that it actually does.

imsodin:
Resurfacing relevant comment from @canton7: 

> According to this docs, the last param is:
> 
>     The size of the buffer lpszLongPath points to, in TCHARs.
> 
> Apparently TCHAR can mean char or wchar_t depending on whether your project is compiled as unicode or not. I'm not sure what Go does here, but it's possible you're telling GetLongPathName that it's got double the amount of space available that it actually does.
> 

No idea - that makes a test mandatory instead of a good idea. Should it work if I just create `LFDataTool` and run the function on `LFDATA~1` (from referenced test)?

lib/fs/basicfs_windows.go
```diff
@@ -150,3 +151,38 @@ func (f *BasicFilesystem) Roots() ([]string, error) {
 
 	return drives, nil
 }
+
+func (f *BasicFilesystem) resolveWin83(absPath string) string {
+	if !strings.Contains(absPath, "~") {
+		return absPath
+	}
```
AudriusButkevicius: Perhaps we should check for caps and length? Our own filenames contain ~ on wondows.

imsodin: I made a check that filters out "false positives" due to `~syncthing~`. If you know of a simpler or more precise criterion based on caps/length for that 8.3 stuff, please let me know.

lib/fs/basicfs_windows.go
```diff
@@ -150,3 +151,40 @@ func (f *BasicFilesystem) Roots() ([]string, error) {
...
+func isMaybeWin83(absPath string) bool {
+	parts := strings.Split(absPath, "~")
+	// Return true unless splitting occurs due to "~syncthing~"
+	for i := 1; i < len(parts); i += 2 {
+		if parts[i] != WindowsTempPrefix[1:len(WindowsTempPrefix)-1] {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
imsodin: > Is it not easier to check length and squiggle being second to last char?

I don't know what that means concretely. I changed it to explicitly check for the `~syncthing~`, I think now it should be clear/robust (?).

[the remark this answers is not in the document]
AudriusButkevicius: Actually, having read the 8.3 path spec, it's more complicated than I thought, so this is probably gudenuf.

AudriusButkevicius:
Good to go?

imsodin:
I think so. Maybe @canton7 wants to have another look?

canton7:
LGTM, I think.

PR #4978, opened 2018-06-01
lib/config, lib/model: Limit size of puller queue (fixes #4976)

### LINKED ISSUE #4976: Excessive RAM usage when syncing large folders
When syncing we build a queue of the files to sync, then sort it according to the desired criteria (size, time, random, etc.). For millions of files this can use a lot of RAM. In some cases we don't really care too much about the exact order and it would be better to use less RAM. For this, we could have a limit on the queue size. When doing a pull and iterating over the database / need list, we stop iterating once the limit is reached. The pull proceeds as normal. When complete, we do another pass and grab more needed files, repeat until done.

calmh:
**Not necessarily for immediate merge given the discussion on the issue, but for consideration at least in the meantime.**

### Purpose

This adds a new config PullerMaxQueueLength that sets the maximum number of items we handle in one puller iteration. If set to <= 0 it defaults to 1000. Point of this is to limit the amount of RAM used for the queue. Downside is that renames that fall accross batches will not be correctly detected... Feels more and more like renames should be flagged somehow on the sending side.

### Testing

Integration tests (50k files and one single file) still pass, and use a but less RAM on the receiving side in the 50k files case.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -79,9 +79,10 @@ const (
 )
 
 const (
-	defaultCopiers          = 2
-	defaultPullerPause      = 60 * time.Second
-	defaultPullerPendingKiB = 2 * protocol.MaxBlockSize / 1024
+	defaultCopiers              = 2
+	defaultPullerPause          = 60 * time.Second
+	defaultPullerPendingKiB     = 2 * protocol.MaxBlockSize / 1024
+	defaultPullerMaxQueueLength = 1000
 
 	maxPullerIterations = 3
 )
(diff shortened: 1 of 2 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -291,6 +301,15 @@ func (f *sendReceiveFolder) pullerIteration(ignores *ignore.Matcher, ignoresChan
...
 					f.queue.Push(file.Name, file.Size, file.ModTime())
 					changed++
+
+					if changed >= maxQueueLen {
+						// We're done for now; set the finished flag to
+						// false so the caller knows we did not do
+						// everything we could, and stop iterating.
+						finished = false
+						return false
+					}
```
imsodin: Shouldn't this block be outside of the switch? `processDirectly` could get huge too.

PR #4982, opened 2018-06-01
lib/model: Wait for folder to stop (fixes #4981)

### LINKED ISSUE #4981: Race on folder restart
The following unrelated test failure due to a nil-deref recently happened:

```
[TestIssue4357] [Test Output]
2018/06/01 13:31:01.338709 model.go:2630: INFO: Adding folder "default" (default)
2018/06/01 13:31:01.338709 set.go:73: INFO: No stored folder metadata for "default": recalculating
2018/06/01 13:31:01.339691 model.go:198: INFO: Ready to synchronize "default" (default) (sendreceive)
2018/06/01 13:31:01.341635 model.go:394: INFO: Paused folder "default" (default)
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x80 pc=0x84cf42]

goroutine 936 [running]:
github.com/syncthing/syncthing/lib/ignore.(*Matcher).SkipIgnoredDirs(0x0, 0x100)
	C:/BuildAgent/work/src/github.com/syncthing/syncthing/lib/ignore/ignore.go:305 +0x32
github.com/syncthing/syncthing/lib/fs.(*BasicFilesystem).Watch(0xc0499c3ae0, 0xa67d35, 0x1, 0xaec620, 0x0, 0xaf00e0, 0xc045f13640, 0xc045c6c700, 0xc045c6c780, 0xc045f13658, ...)
	C:/BuildAgent/work/src/github.com/syncthing/syncthing/lib/fs/basicfs_watch.go:40 +0x10a
github.com/syncthing/syncthing/lib/model.(*folder).startWatchAsync(0xc045bfadc0, 0xaf00e0, 0xc045f13640, 0x0)
	C:/BuildAgent/work/src/github.com/syncthing/syncthing/lib/model/folder.go:337 +0x213
created by github.com/syncthing/syncthing/lib/model.(*folder).startWatch
	C:/BuildAgent/work/src/github.com/syncthing/syncthing/lib/model/folder.go:327 +0x17e
```

This happened on a folder restart, when the model already removed the ignore patterns but the folder loop was still running. This is possible due to `select` choosing a random option, not necessarily the canceled context. This could also lead to parallel pulling and scanning, if a folder is restarted during pulling, because the pull operation can keep running (processing deletes) after the folder was stopped.

I'll propose a mechanism to wait for the folder to "clean up" in the model shortly.

imsodin:
### Purpose

Make sure the folder actually stopped before removing all the assets in the model and continuing to prevent races on folder restart.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -37,6 +37,7 @@ type folder struct {
 	scanNow             chan rescanRequest
 	scanDelay           chan time.Duration
 	initialScanFinished chan struct{}
+	stopped             chan struct{}
 
 	pullScheduled chan struct{}
 
@@ -81,6 +82,7 @@ func newFolder(model *Model, cfg config.FolderConfiguration) folder {
 		watchCancel: func() {},
 		watchErr:    errWatchNotStarted,
 		watchErrMut: sync.NewMutex(),
+		stopped:     make(chan struct{}),
 	}
 }
 
@@ -91,6 +93,7 @@ func (f *folder) Serve() {
 	defer func() {
 		f.scanTimer.Stop()
 		f.setState(FolderIdle)
+		close(f.stopped)
 	}()
 
 	pause := f.basePause()
@@ -225,6 +228,10 @@ func (f *folder) Stop() {
 	f.cancel()
 }
 
+func (f *folder) Stopped() <-chan struct{} {
+	return f.stopped
+}
+
 // CheckHealth checks the folder for common errors, updates the folder state
 // and returns the current folder error, or nil if the folder is healthy.
 func (f *folder) CheckHealth() error {
@@ -320,7 +327,11 @@ func (f *folder) restartWatch() {
 func (f *folder) startWatch() {
 	ctx, cancel := context.WithCancel(f.ctx)
 	f.model.fmut.RLock()
-	ignores := f.model.folderIgnores[f.folderID]
+	ignores, ok := f.model.folderIgnores[f.folderID]
+	if !ok {
+		l.Debugf("Failed to get ignore patterns for folder %v while starting the filesystem watcher -> aborting, folder will be stopped soon (racy)", f.Description())
+		return
+	}
 	f.model.fmut.RUnlock()
 	f.watchChan = make(chan []string)
 	f.watchCancel = cancel
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -67,6 +67,7 @@ type service interface {
 	CheckHealth() error
 	PullErrors() []FileError
 	WatchError() error
+	Stopped() <-chan struct{}
 
 	getState() (folderState, time.Time, error)
 	setState(state folderState)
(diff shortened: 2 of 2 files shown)


lib/model/folder.go
```diff
@@ -225,6 +228,10 @@ func (f *folder) Stop() {
 	f.cancel()
 }
 
+func (f *folder) Stopped() <-chan struct{} {
+	return f.stopped
+}
```
AudriusButkevicius: I *think*, Stop() could wait for f.stopped to be closed, at which point m.Remove(id) would block until Stop() returns.

lib/model/folder.go
```diff
@@ -320,7 +327,11 @@ func (f *folder) restartWatch() {
 func (f *folder) startWatch() {
 	ctx, cancel := context.WithCancel(f.ctx)
 	f.model.fmut.RLock()
-	ignores := f.model.folderIgnores[f.folderID]
+	ignores, ok := f.model.folderIgnores[f.folderID]
+	if !ok {
+		l.Debugf("Failed to get ignore patterns for folder %v while starting the filesystem watcher -> aborting, folder will be stopped soon (racy)", f.Description())
+		return
+	}
```
AudriusButkevicius: Surely, while the parent context is not cancelled, ignores should be there, as the ignores get removed after the folder is stopped (with the <-f.Stopped() stuff).

lib/model/folder.go
AudriusButkevicius: There is RemoveAndWait in 2.0.3 version of suture.

lib/model/model.go
```diff
@@ -353,10 +353,13 @@ func (m *Model) RemoveFolder(cfg config.FolderConfiguration) {
...
 func (m *Model) tearDownFolderLocked(folder string) {
-	// Stop the services running for this folder
+	// Stop the services running for this folder and wait for them to finish
+	// stopping to prevent races on restart.
+	m.fmut.Unlock()
 	for _, id := range m.folderRunnerTokens[folder] {
-		m.Remove(id)
+		m.RemoveAndWait(id, 0)
 	}
+	m.fmut.Lock()
```
AudriusButkevicius: Do we know that nothing in the stopping code needs a lock?

imsodin: Yes, tests time out without it. I mean it probably doesn't always, but it might (e.g. finishing pull).

lib/model/folder.go
```diff
@@ -327,11 +324,7 @@ func (f *folder) restartWatch() {
 func (f *folder) startWatch() {
 	ctx, cancel := context.WithCancel(f.ctx)
 	f.model.fmut.RLock()
-	ignores, ok := f.model.folderIgnores[f.folderID]
-	if !ok {
-		l.Debugf("Failed to get ignore patterns for folder %v while starting the filesystem watcher -> aborting, folder will be stopped soon (racy)", f.Description())
-		return
-	}
+	ignores := f.model.folderIgnores[f.folderID]
```
AudriusButkevicius: We should however add a small delay to the timer, so that if both the context is cancelled, and the timer is fired, the ctx.Done() would win.

imsodin: I don't think it's necessary: Due to `RemoveAndWait` `folderIgnores` won't be removed until this is finished, so it can't be `nil`.

lib/model/model.go
AudriusButkevicius: What I mean is that if something within the folder.Stop code tries to access the model and get fmut, we'd dead lock.

The only reason we need a lock here is to prevent a race on the token storage map, we could do the stopping outside of the lock I guess.

lib/model/folder.go
AudriusButkevicius: Right.

lib/model/model.go
imsodin: I do the stopping outside of the lock. I even got the tokens out of the lock, which was wrong (corrected now).

PR #4986, opened 2018-06-03
gui: Add tabs in device editor

qepasa:
### Purpose

Hello everyone. This is long due update to GUI aimed at exposing bandwidth throttling (should have done that long ago, sorry!). I created it following @imsodin suggestion from a few weeks back on IRC. In case you would like to see other things exposed in each tab I'll happily add them.

### Testing

Not much. Clicked through device editor and checked most of its existing functionalities and the new one.

### Screenshots

https://imgur.com/a/EEHGyTY

### DIFF
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -1,123 +1,168 @@
 <modal id="editDevice" status="default" icon="{{editingExisting ? 'fas fa-pencil-alt' : 'fas fa-desktop'}}" heading="{{editingExisting ? 'Edit Device' : 'Add Device' | translate}}" large="yes" closeable="yes">
-  <div class="modal-body">
-    <form role="form" name="deviceEditor">
-      <div class="form-group" ng-class="{'has-error': deviceEditor.deviceID.$invalid && deviceEditor.deviceID.$dirty}" ng-init="loadFormIntoScope(deviceEditor)">
-        <label translate for="deviceID">Device ID</label>
-        <div ng-if="!editingExisting">
-          <input name="deviceID" id="deviceID" class="form-control text-monospace" type="text" ng-model="currentDevice.deviceID" required="" valid-deviceid list="discovery-list" aria-required="true"/>
-          <datalist id="discovery-list">
-            <option ng-repeat="id in discovery" value="{{id}}" />
-          </datalist>
-          <p class="help-block" ng-if="discovery">
-            <span translate>You can also select one of these nearby devices:</span>
-            <ul>
-              <li ng-repeat="id in discovery"><a ng-click="currentDevice.deviceID = id">{{id}}</a></li>
-            </ul>
-          </p>
-          <p class="help-block">
-            <span translate ng-if="deviceEditor.deviceID.$valid || deviceEditor.deviceID.$pristine">The device ID to enter here can be found in the "Actions > Show ID" dialog on the other device. Spaces and dashes are optional (ignored).</span>
-            <span translate ng-show="deviceEditor.deviceID.$valid || deviceEditor.deviceID.$pristine">When adding a new device, keep in mind that this device must be added on the other side too.</span>
-            <span translate ng-if="deviceEditor.deviceID.$error.required && deviceEditor.deviceID.$dirty">The device ID cannot be blank.</span>
-            <span translate ng-if="deviceEditor.deviceID.$error.validDeviceid && deviceEditor.deviceID.$dirty">The entered device ID does not look valid. It should be a 52 or 56 character string consisting of letters and numbers, with spaces and dashes being optional.</span>
-            <span translate ng-if="deviceEditor.deviceID.$error.unique && deviceEditor.deviceID.$dirty">A device with that ID is already added.</span>
-          </p>
-        </div>
-        <div ng-if="editingExisting" class="well well-sm text-monospace" select-on-click>{{currentDevice.deviceID}}</div>
+<div class="modal-body">
+<form role="form" name="deviceEditor">
+  <ul class="nav nav-tabs" ng-init="loadFormIntoScope(folderEditor)">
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
The screenshots don't look right, errors from wrong tab are visible.

AudriusButkevicius:
Also, I would put compression in advanced.

AudriusButkevicius:
And potentially even addresses.

qepasa:
You're right, I overlooked those errors when I was copying from folder modal. Here is link to screenshots with suggested updates: https://imgur.com/a/hFfJoGG

imsodin:
Maybe add a note on transfer limits about what to enter to not limit at all (is it 0?).

AudriusButkevicius:
also the formatting for the HTML looks wrong, wrong indentation, etc.

qepasa:
Yes, it's 0. I added a note and a tooltip. It looks like this: 
device tab note

I know I screwed up formatting. I foolishly hit reformat file in goland and could not get html formatting settings to indent it properly afterwards.
It looks more or less how it should now although I could have overlooked something.

uok:
A few suggestions - please
- use already existing strings and labels from the general settings modal like "Incoming Rate Limit (KiB/s)" so the translations can be reused.
- keep wording consistent, e.g. use "device rate limits" and not "device bandwidth limits"
- extra text hint about "0" is not needed because
- for red hints just copy texts from general settings "The rate limit must be a non-negative number (0: no limit)"

qepasa:
Thanks for feedback. I added suggested changes. It looks like this now:
gui
gui2

imsodin:
Generally looks good to me, just one (debatable) point. I'd consider compression more "advanced" (as in less used) than address - personally I'd put both into advanced or address into general.

uok:
I think leaving "address" in section general would nudge (new) users to think about setting up direct connections instead of using relay/discovery servers.

AudriusButkevicius:
So this PR itself is ok.

There are plenty of nice things we could expose given we now have the space, but that's a story for a different PR.

As to leaving address in general, people have no clue what dynamic means, and there is already a warning icon if connected not directly in the UI.

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -1,98 +1,134 @@
...
+                <div class="col-md-6">
+                  <div ng-class="{'has-error': deviceEditor.maxSendKbps.$invalid && deviceEditor.maxSendKbps.$dirty}">
+                    <div class="row">
+                      <span class="col-md-8" translate>Outgoing Rate Limit (KiB/s)</span>
+                      <div class="col-md-4">
+                        <input name="maxSendKbps" id="maxSendKbps" class="form-control" type="number" ng-model="currentDevice.maxSendKbps" required="" aria-required="true" min="0"/>
+                      </div>
+                    </div>
+                    <p class="help-block" ng-if="!deviceEditor.maxSendKbps.$valid && deviceEditor.maxSendKbps.$dirty" translate>
+                      The rate limit must be a non-negative number (0: no limit) </p>
```
AudriusButkevicius: The way this is broken over multiple lines makes me sad, please adjust to either `<p>foo</b>` or
```
<p>
   Foo
</p>
```

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -127,8 +126,7 @@
                       <input name="maxRecvKbps" id="maxRecvKbps" class="form-control" type="number" ng-model="currentDevice.maxRecvKbps" required="" aria-required="true" min="0"/>
                     </div>
                   </div>
-                  <p class="help-block" ng-if="!deviceEditor.maxRecvKbps.$valid && deviceEditor.maxRecvKbps.$dirty" translate>
-                    The rate limit must be a non-negative number (0: no limit) </p>
+                  <p class="help-block" ng-if="!deviceEditor.maxRecvKbps.$valid && deviceEditor.maxRecvKbps.$dirty" translate>The rate limit must be a non-negative number (0: no limit) </p>
```
uok: there are spaces before `</p>`- I don't know if this is a problem for translation strings.

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -116,8 +116,7 @@
                         <input name="maxSendKbps" id="maxSendKbps" class="form-control" type="number" ng-model="currentDevice.maxSendKbps" required="" aria-required="true" min="0"/>
                       </div>
                     </div>
-                    <p class="help-block" ng-if="!deviceEditor.maxSendKbps.$valid && deviceEditor.maxSendKbps.$dirty" translate>
-                      The rate limit must be a non-negative number (0: no limit) </p>
+                    <p class="help-block" ng-if="!deviceEditor.maxSendKbps.$valid && deviceEditor.maxSendKbps.$dirty" translate>The rate limit must be a non-negative number (0: no limit) </p>
```
uok: there is a space before `</p>`- I don't know if this is a problem for translation strings.

AudriusButkevicius: I had the same comment, make sure its identical to the code in the other place.

qepasa: There are a few such spaces before closing tags in current device modal but not in folder modal. I will get rid of all just to be safe.

qepasa: Well to be exact there are no such spaces in code from master. I must have added them accidentally when I was fooling around with formatting. Anyway, they are gone.

AudriusButkevicius:
This seems to affect one other string (middle) I did not expect:

image

It was double space at one place before, now it's removed, but that means everyone will have to retranslate, so we should change it back.

AudriusButkevicius:
Jesus fucking christ, the CI builds are now taking 20 minutes. Is the code base being transpiled to C++ before it gets compiled or something?

PR #4989, opened 2018-06-05
gui: Hide "allowed networks" if empty

uok:
### Purpose

Hide line for "allowed networks" (advanced config) in device panel if it is empty.

### Testing

yep

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -653,7 +653,7 @@
                       <th><span class="fas fa-fw fa-exclamation-triangle text-danger"></span>&nbsp;<span translate>Connection Type</span></th>
                       <td class="text-right">{{connections[deviceCfg.deviceID].type}}</td>
                     </tr>
-                    <tr ng-if="deviceCfg.allowedNetworks">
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Hold on. Empty string is false'ish, no? Why is this needed then? This also breaks for null/undefined values.

uok:
I was expecting if advanced config is not (never was in my case) set that the line isn't displayed.  That's why I created this PR. I'm open to suggestions for possible values to check for

calmh:
FWIW, I have the problem too:

screen shot 2018-06-05 at 19 53 50

I think it stems from the fact that this is an array, and an empty array in Javascript is trueish... I can't vouch for the fix; it might work because Angular stringifies the empty array to the empty string.

AudriusButkevicius:
It seems stringification is done afterwards, so you probably need what we had there already && arr.length > 1

uok:
I checked config.xml: entry "<allowedNetwork>...</allowedNetwork>` only exists if it is set. So in my case (and probably 99% of all users) there is no entry at all.
A few lines down in index.html I found a similar check for folders array so I updated the PR to use this method.

PR #4992, opened 2018-06-06
gui: fix pagination location

wwwutz:
WIP

prevent buttons from moving on dynamic rendering

### Purpose

When trying to go through the list of unsynced files in the GUI , the pagination buttons for next-page and previous-page are moving left or right due to dynamic rendering. This makes it almost impossible to keep the mouse pointer or finger press location at the same position. On the last page the buttons might even move up, so the last click might end up outside of the window closing it.

In the original layout the pagination buttons were rendered using `&lsaquo;` and `&rsaquo;` [&lsaquo;] and [&rsaquo;]) making the buttons smaller and harder to point at.

This has been changed to font-awesome symbols `fa-long-arrow-alt-left` and `fa-long-arrow-alt-right` as this is consistent with other ST UI buttons.

### Testing

Force a couple of hundred files to change, start up the web UI and open the Out-of-Sync items window in Folders view. Move the mouse pointer to the long-left-arrow icon (next page) and keep clicking. The window should change contents and stay on the last page without moving the mouse pointer.

### Screenshots

Before (pagination at the bottom)

st-old-pagination

Last page ( the mouse pointer is now pointing outside the window so it would close)

st-old-last-page

After this pull request ( buttons more visible and they are fixed )

st-fix-pagination

### Documentation

There are no screenshots in the docs, so no need to update anything there.

### ToDo

I didn't find a way to prevent modifying the vendor files to get the same result via .css. I see this is a problem, but I think making the UI more usable is worth it.


wweich:
Instead of messing arround with vendor files and putting the right arrow on the left (which is very unintuitive IMHO), you can just use CSS:
```
.pagination > li {
    min-width: 3.4em;
    display: inline-block;
    text-align: center;
}

.pagination > li > a {
    width: 100%;
}
```

It will look like this:
bildschirmfoto vom 2018-06-06 um 18 35 19
bildschirmfoto vom 2018-06-06 um 18 31 49
bildschirmfoto vom 2018-06-06 um 18 31 34

The next button will almost exactly stay at the same position.

_On small screens, the pagination will not look more messy than with the default style._

Also putting the pagination (only) on top is not very user friendly IMHO.

wwwutz:
If we get the button position get fixed by plain .css, that's ok with me. 

My initial trouble with the current design is that I have to scroll down to get to the next page. And on the last page, when you're unlucky, you hit the outside.

You see this weird design decision often in forum-UIs or discussion boards. I don't understand why I"m forced to scroll or move my mouse/finger etc when a simple click/touch would be enough.

So I would like to see the buttons on the top and on the bottom. Which is OK for long lists, so you don't have to either scroll down or up.

What about the "arrows" (or little weird dots on HiDPI displays). Would it be possible to fix them in .css, too ? They are hidden in the `$templateCache` magic.

AudriusButkevicius:
> My initial trouble with the current design is that I have to scroll down to get to the next page. And on the last page, when you're unlucky, you hit the outside.

You can access the last page from the page selector, no need to slam the next arrow repeatedly?

Also, the scroll down penalty is one-off, and having to scroll down to see the first N items because the top is crammed with controls and barely any lines are visible is equally bad, so I am not sure I buy the "move pagination to the top" argument.
Also, it's alienating as hell as that is not what people are used to. 
Amazon et all have much better UX people that design their UIs, and they don't seem to have pagination buttons at the top? Nor am I a fan of swapping the arrow locations, as this different to what everyone else does.

I guess I'd be happy with a CSS hacking to make the buttons not move around when changing pages, and change the arrow icons, but the rest seems like it makes it worse.

As for whether it's possible to change the icons by not changing the vendor files, I am pretty sure it should be, as it's "cascading" style sheets after all.

wweich:
The buttons are markup so we'd need to figure out CSS hacks to replace them. As they don't have special CSS classes, they are indistinguishable from the other buttons. Setting the first and last won't work as the 10,25,50 uses the same CSS classes.

The pagination addon has a template option thought, so we can set a template containing other arrow chars when initializing it, instead of editing the vendor file.

alpadev:
I wanted to jump in because I love your project and just looked for some posibilities to contribute back to it.

A maybe not so hacky way to make these pagination steps of equal size:
```css
@supports (display: inline-grid) {
    @media (min-width: 564px) {
        dir-pagination-controls .pagination {
            display: inline-grid;
            grid-auto-columns: 1fr;
            grid-auto-flow: column;
        }

        dir-pagination-controls .pagination a {
            display: block;
            float: none;
            text-align: center;
        }
    }
}
```

st-review:
:robot: *beep boop*

PR #4996, opened 2018-06-08
lib/protocol: Test for IsEquivalent

calmh:
### Purpose

Specify the current behavior of the function, as there was no test already that I could find.

### Testing

Indeed.


calmh:
Yep.

PR #4999, opened 2018-06-10
cmd/syncthing: Add tray icon for windows

AudriusButkevicius:
This is a bit of a moon shot, and I haven't yet figured out how to fully polish this.

Essentially when launching via explorer, it starts a tray icon, in all other cases it starts headless.

image

Also, it's compiled headless by default, this avoids the console popping up when launched from explorer.
This does not break stdin/out wiring, and I can still see log output if I launch from cygwin.

However if I launch from cmd, it seems to start a headless process, so I have no console attached it seems.

Perhaps we should revert the compiled headless, and accept the console popping up briefly before we wire the tray icon.

Also, if the tray icon fails, we can keep the console visible,

CC @canton7


AudriusButkevicius:
Also, perhaps I'll add Toast support, so we can pop up a message as it starts up, as otherwise the tray icon might not even show up in the task bar (goes straight into the minimized section)

AudriusButkevicius:
Also, we could use the full shebang https://github.com/getlantern/systray, yet that's a bit grim, as it unpacks dlls at runtime, dlopens them, etc, etc.
Also, seems to use CGO for Darwin and Linux, so not sure we could do all the cross-compilation into dragonfly and stuff.

canton7:
What did you want my input on?

AudriusButkevicius:
@canton7 mostly the win32 api stuff and -H windowsgui stuff.

AudriusButkevicius:
Also, any ideas if this will break anything in synctrayzor if we go with windowsgui?

canton7:
Right, so compiling as a windows gui application means that windows will never launch it as a console application, and you'll never get stdout. Cmd will always detcha it. Although you can choose to separately launch a console window if you want. Not sure what cygwin is doing there. 

I'm not sure off the top of my head whether it will break things which try to run it as a child process: I don't think it does, but i'm on holiday this week and can't check. If so, it will also break things like nssm. 

As long as the default behaviour is to launch the browser when started, I don't think this will confuse people too much. You can always show a "it's in the tray" message there.

AudriusButkevicius:
Perhaps I'll leave the console and merge it for the RC for people to try out.

AudriusButkevicius:
Gud enouf I think:

image

cmd/syncthing/main.go
```diff
@@ -592,7 +602,7 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 	// We want any logging it does to go through our log system.
 	mainService := suture.New("main", suture.Spec{
 		Log: func(line string) {
-			l.Debugln(line)
+			l.Infoln(line)
```
AudriusButkevicius: I think it's pointless to have this in debug, disagree?

imsodin: Currently agree, because unsolved #4758.

cmd/syncthing/main.go
```diff
@@ -201,6 +203,9 @@ are mostly useful for developers. Use with care.
                    "h", "m" and "s" abbreviations for hours minutes and seconds.
                    Valid values are like "720h", "30s", etc.
 
+ STNOTRAY          Prevents syncthing from hiding the console and starting up
+                   with a tray icon. (Windows only)
```
imsodin: Does this mean the default is to hide console? For double clicking an exe that seems fine, for someone running it from cmd this seems annoying. I am just asking because my current windows understanding is power user == syncthing exe and normal user == synctrayzor. So I am wondering whether such a default (if it is the case) is a disservice to power user. Just a thought, I am no windows user.

cmd/syncthing/main.go
```diff
@@ -220,8 +225,11 @@ The following are valid values for the STTRACE variable:
 // Environment options
 var (
 	noUpgradeFromEnv = os.Getenv("STNOUPGRADE") != ""
-	innerProcess     = os.Getenv("STNORESTART") != "" || os.Getenv("STMONITORED") != ""
+	isMonitored      = os.Getenv("STMONITORED") != ""
```
imsodin: What's the reason to add this?

AudriusButkevicius: This is to distinguish between our monitor, and synctrayzor as a monitor, because synctrayzor might not put logs in a logfile like our monitor does, hence no `Open log` menu item.

AudriusButkevicius:
To be honest, now it's quite far from that package.
I am not a license guru.
I guess we remove the copyright, and put his copyright in, given we're already MIT licensed?

imsodin:
Neither am I, I am wondering about that crap all the time. I think we are MPL, which is compatible with MIT, but not the same. So from the literal MIT text I'd say we need to place a license.MIT or something in there plus his copyright one along ours. But I have never seen any open source project that did this, so my idea is probably wrong. As MPL and MIT are both very permissive and you say it's pretty different, not doing anything but referencing the original code might be just fine.

If there is any licensing guru reading this, feel free to chime in (as long as you don't tell us to do a lot of work :P ).

AudriusButkevicius:
Sorry, I thought we are MIT, but the license has changed so many times that I forgot which one of the fucking acronyms we use these days.

Anyways, so should I just copy the MIT header in there, leave his copyright in that file or what?

cmd/syncthing/main.go
AudriusButkevicius: To be honest, I am thinking if we even need this. This is to force launch without a tray from explorer, but that means that you have to set the env var first, which probably means setting it OS wide.

calmh:
I think retain the license file and copyright is the clean and easy thing to do. No need to pollute their library with our MPL if we can avoid it.

AudriusButkevicius:
It's not really a fork, it's more like ripping out the part we care about.
The original implementation spawns a separate process to maintain the tray icon on osx and others, and communicates over a socket, which is OTT for what we want to do.

AudriusButkevicius:
I'll raise an issue to let the guy know that I've bastardized his code. At this point it's nothing like his code, so it would be very hard to contribute anything back.

calmh:
Ah

AudriusButkevicius:
Good to go/

xjtdy888:
Tooltip is not display when the mouse over to the tray icon.
qq20180626-195533

AudriusButkevicius:
So it seems we can either have balloon notifications or tooltips (if we drop to older version of the API), not both.

The API for this is a complete mess:

https://docs.microsoft.com/en-us/windows/desktop/api/shellapi/ns-shellapi-_notifyicondataa

Note the double `szTip` definition.

If someone cares about this strongly in the future, I welcome to be my guest and solve the API nonsense.

AudriusButkevicius:
Even read the Qt code, can't for the love of me work it out:

https://github.com/qt/qtbase/blob/a37dd93defd91b79fb6730d0ff0515a66a0d3972/src/plugins/platforms/windows/qwindowssystemtrayicon.cpp#L344

xjtdy888:
After windows explorer.exe crash and restart,  tray icon disappear, we need handle TaskbarCreated mesage  and  restore it.

https://stackoverflow.com/questions/32044732/how-to-restore-a-missing-notification-icon-when-explorer-exe-is-restarted

cmd/syncthing/main.go
```diff
@@ -643,6 +653,69 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 	l.Infoln(LongVersion)
 	l.Infoln("My ID:", myID)
 
+	if runtime.GOOS == "windows" && ((startTray && !noTray) || runtimeOptions.forceTray) {
```
calmh: This thing should be broken out into a function somewhere.

AudriusButkevicius:
I think we leave this to stir for now. I want to reimplement the tray icon business how it's done in Qt, so that we handle all the edge cases as they were pointed out by @xjtdy888.

Obviously, that's a fair chunk of work, so not sure when this will happen.

Catfriend1:
Quite a while ago when being on c++ and pure winApi... I recall to readd the systray icon after an explorer restart by sending NIF_ADD messages to the tray periodically with a constant ID so the icon does created, recreated and one is only there. NIF_MODIFY didn't readd the icon and could only modify an existing one thats shown on the tray. When I'm on the screen next time I'll look it up in my old projects.

st-review:
:robot: *beep boop*

PR #5001, opened 2018-06-12
[dockerfile] Fix 'permission denied'

wy-z:
Signed-off-by: weiyang <weiyang.ones@gmail.com>

### Purpose

https://github.com/syncthing/syncthing/blob/master/Dockerfile#L15
`VOLUME` will mount new volume and change the owner of `/var/syncthing` to `root`
```
This Dockerfile results in an image that causes docker run to create a new mount point at /myvol and copy the greeting file into the newly created volume.
```
https://docs.docker.com/engine/reference/builder/#volume

image

### Testing

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,20 +10,19 @@ RUN rm -f syncthing && go run build.go build syncthing
 
 FROM alpine
 
-EXPOSE 8384 22000 21027/udp
-
-VOLUME ["/var/syncthing"]
-
 RUN apk add --no-cache ca-certificates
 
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
You're supposed to mount the volume yourself, as far as I understand by using -v flag.

wy-z:
Yeah, but I think this is a bug~

calmh:
I think the failure here is a feature. If it did not fail, you could set up Syncthing and things would appear to work, only to be blown away on the next restart due to your `--rm`. Using a volume avoids this.

wy-z:
good~

PR #5003, opened 2018-06-13
lib/model: Don't panic when rechecking file (fixes #5002)

### LINKED ISSUE #5002: v0.14.48 panic: runtime error: index out of range
Syncthing Version: syncthing v0.14.48 "Dysprosium Dragonfly" (go1.10.2 linux-386) teamcity@build.syncthing.net 2018-05-14 06:53:06 UTC
OS Version: Slackware 14.2 32bit

Panics at irregular intervals; sometimes as soon as ~40s after startup, sometimes several hours.  Always the message is "panic: runtime error: index out of range".  
Presumably my machine acquired v0.14.48 within twelve hours of its release, which would put the first panic 24hrs or more since release.  

From the first log file:
[7JAPU] 16:16:31 INFO: Folder DABingeNeedRetire isn't making progress. Pausing puller for 1m0s.
Panic at 2018-06-08T16:17:16-08:00
panic: runtime error: index out of range

goroutine 247254 [running]:
[....]

Switching back to v0.14.47 (And turning off auto update in the config) fixes this problem.

calmh:
### Purpose

Obvs

### Testing

Unit

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1402,7 +1402,7 @@ func (m *Model) recheckFile(deviceID protocol.DeviceID, folderFs fs.Filesystem, 
 		return
 	}
 
-	if blockIndex > len(cf.Blocks) {
+	if blockIndex >= len(cf.Blocks) {
 		l.Debugf("%v recheckFile: %s: %q / %q i=%d: block index too far", m, deviceID, folder, name, blockIndex)
 		return
 	}
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -3611,7 +3611,7 @@ func TestIssue4903(t *testing.T) {
 }
 
 func TestIssue5002(t *testing.T) {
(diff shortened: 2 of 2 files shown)


AudriusButkevicius:
Surprised it took so long for this to come out.

lamixer:
Is there any workaround for this problem while we await 14.49 (and in my case, for Fedora package to push that version)? 
Can I do anything to make it so that an index is not currently equal to the number of blocks?

AudriusButkevicius:
A rescan should help or a modification/touch of the file in question

lamixer:
Thanks, AudriusB ... I touched some files I thought it could be and Syncthing did seem to run for a few seconds (better than before, less than one second). I don't want to waste your time, so I'll just say I eagerly await the next release -- I guess I can just drop the binary in overwriting the packaged version and that should probably not cause any serious problems!

AudriusButkevicius:
You can touch -r. Or just upgrade to rc

PR #5008, opened 2018-06-15
lib/db: Remove updated invalid files from need bucket (fixes #5007)

### LINKED ISSUE #5007: Ignored files are still considered needed
Originally came up in the forum: https://forum.syncthing.net/t/did-something-recently-changed-in-ignores-handling/11714/6

What happens is that invalid/ignored files are not added at the top of the global list, even if they are of the newest version. That's fine, but the code that keeps the need bucket up to date does assume that only changes at the beginning of the list are relevant, which isn't true in this case.

imsodin:
### Purpose

Fixes #5007, Also in the puller updating invalid files wasn't considered a change, so in the logs it looks like everything works fine in the puller, while it doesn't.

An additional problem is, that for RC users that had an update to ignored files while using the newest rc the need bucket is out of sync with reality. That requires either manual db backup/rest by the user or a db transition just for DBs exactly at version 4 (?). I can do that, just not today anymore.

### Testing

Unit test.

### DIFF
--- a/lib/db/leveldb_transactions.go
+++ b/lib/db/leveldb_transactions.go
@@ -97,13 +97,17 @@ func (t readWriteTransaction) updateGlobal(gk, folder, device []byte, file proto
 		return false
 	}
 
+	name := []byte(file.Name)
+
 	if removedAt != 0 && insertedAt != 0 {
+		if bytes.Equal(device, protocol.LocalDeviceID[:]) && file.IsInvalid() && file.Version.Equal(fl.Versions[0].Version) {
+			l.Debugf("local need delete; folder=%q, name=%q", folder, name)
+			t.Delete(t.db.needKey(folder, name))
+		}
 		l.Debugf(`new global for "%v" after update: %v`, file.Name, fl)
 		t.Put(gk, mustMarshal(&fl))
 		return true
 	}
-
-	name := []byte(file.Name)
 
 	// Remove the old global from the global size counter
 	var oldGlobalFV FileVersion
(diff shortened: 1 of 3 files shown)


calmh:
> db backup/rest by the user or a db transition just

how about treating this like our metadata and rebuilding every 30 days, in which case a manual workaround is available when this is reported (STRECHECKDBEVERY=0 ...)

lib/db/set_test.go
```diff
@@ -997,6 +997,31 @@ func TestMoveGlobalBack(t *testing.T) {
 	}
 }
 
+func TestIssue5007(t *testing.T) {
```
calmh: Add a description here of what this test actually tests for, please. I've realized it's super annoying in the future to run across an anonymous numbered test case.

lib/db/leveldb_transactions.go
```diff
@@ -97,14 +97,18 @@ func (t readWriteTransaction) updateGlobal(gk, folder, device []byte, file proto
 		return false
 	}
 
+	name := []byte(file.Name)
+
 	if removedAt != 0 && insertedAt != 0 {
+		if bytes.Equal(device, protocol.LocalDeviceID[:]) && file.IsInvalid() && file.Version.Equal(fl.Versions[0].Version) {
```
calmh: I guess we could skip the `IsInvalid` here and just delete regardless?

imsodin:
> how about treating this like our metadata and rebuilding every 30 days, in which case a manual workaround is available when this is reported (STRECHECKDBEVERY=0 ...)

I'd rather fix it up front for everyone potentially affected and with the db transition mechanism this is super easy. We can still add that reset on top of it, but I'd also wait with that until there is a reason (i.e.  mysterious problems).

PR #5011, opened 2018-06-17
cmd/syncthing: Browser is not refreshed after UI file update

xjtdy888:
If-Modified-Since condition is incorrect, causing HTTP to always return StatusNotModified


AudriusButkevicius:
The second addition clearly causes confusion, what we should do is probably
```
if modified.Equals(t) || modified.Before(t) {
   return NotModified
}
``

xjtdy888:
Before I did not correctly fork the warehouse, I will resubmit the pull request

PR #5014, opened 2018-06-17
lib/model: More auto accept tests

AudriusButkevicius:
Trying to repro the folder already started.

### DIFF
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -68,6 +68,10 @@ func init() {
 				DeviceID:          device1,
 				AutoAcceptFolders: true,
 			},
+			{
+				DeviceID:          device2,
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
The second commit adds a test that does not check correctness, just checks for panics.

calmh:
Maybe comment to that effect on the tests, otherwise I guess merge at will?

calmh:
The flakiness is strong with this one... (Due to no fault of this PR itself as far as I can tell)

calmh:
Going to do a new RC with the other fixes before looking closer

AudriusButkevicius:
@imsodin, @calmh  wanna do the honours?

PR #5015, opened 2018-06-17
gui: Fix add device (fixes #5012)

### LINKED ISSUE #5012: Cannot add new devices
Since #4986 I cannot save a new device in the GUI. The save button never becomes enabled. To reproduce,

- "Add new device"
- Enter a valid device ID
- Observe the validation error going away
- The save button is still disabled

Needs fixing for .49

qepasa:
### Purpose

As tittle suggests, fixes the issue with adding new device. Moreover I discovered a bug with entering rational numbers as limiter values hence the pattern and changed error message.

### Testing

Adding new device with and without editing limiter values and editing existing device limiter values.

### DIFF
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -113,20 +113,20 @@
                     <div class="row">
                       <span class="col-md-8" translate>Outgoing Rate Limit (KiB/s)</span>
                       <div class="col-md-4">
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
I think leave the strings as they are, as this will be yet another string to translate.

imsodin:
Not only is it an additional translation string, I also don't know how common integer is as a known for non-techie people.

calmh:
Thanks!

PR #5016, opened 2018-06-18
cmd/syncthing: Browser is not refreshed after UI file update

xjtdy888:
If-Modified-Since condition is incorrect, causing HTTP to always return StatusNotModified


AudriusButkevicius:
I'll leave for @calmh to merge.

PR #5022, opened 2018-06-20
cmd/syncthing, lib/db: Abort execution if db version is too high (fixes #4994)

### LINKED ISSUE #4994: Database schema migrations should handle up/down/up-grades
Lets assume I'm on 0.14.45. I upgrade to 0.14.48 and two database migrations are applied. But 0.14.48 has some sort of bug that I can't tolerate and I roll back to 0.14.45. Things are fine.

Then 0.14.49 comes out and I upgrade. The migrations will be applied again, but there's already a bunch of new indexes and stuff in place. Some of those keys will probably be overwritten correctly, some will remain from the previous migration run and be totally wrong at this point. Chaos ensues.

We should wipe the key space for what we're creating in a migration, before creating it.

---

Not all migrations will be backwards compatible and then the downgrade step itself will fail. That's a different issue, and probably fine.

imsodin:
### Purpose

Abort updating the database if it has a newer version than the program itself knows. Then abort launching Syncthing, telling the user about the issue with advice to either upgrade Syncthing or reset the database.

### Testing

Check that updating the db does fail with the correct error if the stored dbVersion is too high.


lib/db/leveldb_dbinstance_updateschema.go
```diff
@@ -15,12 +16,18 @@ import (
 
 const dbVersion = 4
 
-func (db *Instance) updateSchema() {
+var ErrDatabaseDowngrade = errors.New("the version of the db is more recent than Syncthing and thus not compatible")
```
calmh: Just a matter of opinion, but I'd prefer something more terse here and have the explanatory stuff all in the log message. Something like `database version from the future` or `unknown database version` maybe?

cmd/syncthing/main.go
```diff
@@ -701,7 +701,9 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 
 	dbFile := locations[locDatabase]
 	ldb, err := db.Open(dbFile)
-	if err != nil {
+	if err == db.ErrDatabaseDowngrade {
+		l.Fatalln("Cannot open database:", err, "- Did you downgrade Syncthing? If you can't upgrade to the most recent version, you need to reset the database (-reset-db) which will cause a full rehash of all your files")
```
AudriusButkevicius: Wat? We can open the database... We should also mention that any out of sync files will end up being conflicts.

lib/db/leveldb_dbinstance_updateschema.go
AudriusButkevicius: Agree

calmh:
> There are too many ways this can go wrong.

I can agree with that assessment. It just hurts a bit to add a case, that might not be super unusual in the wild for all we know, where we essentially tell the user "haha, you're fucked, time to wipe everything and start over".

In the distant past we did db migrations by copying and converting to a new db name. Without doing that whole dance we could zip a backup of the database prior to migrations. That leaves a path open to go back with a working, although outdated, database.

If nothing else I think we should probably change the database version field to be the Syncthing version required to open it. Then the error message can be `The index database is in a format that requires Syncthing v0.14.59, this syncthing v0.14.50 cannot open it.`

cmd/syncthing/main.go
imsodin: Good point. Question: Will all of them become conflicts or just those, that were never updated locally? E.g. if a file before was {ourID: 10, otherID: 10}, after reset it will be {ourID: 1}, this wouldn't be a conflict. If I am not mixing this up, a less dramatic formulation like "out of sync files may end up being conflicts" is appropriate?

AudriusButkevicius:
> "haha, you're fucked, time to wipe everything and start over".

We've done that with upgrades in the past, no biggie?
Also, people seem to use this as a daily stress reliever anyway, as we've been told by that guy that did large blocks + db reset.

imsodin:
I am not keen to write and maintain up-, down- and re-update mechanisms. I believe the gain is pretty small, because the chance that it will be used is small: A user only ever needs to downgrade if we really fuck up. Then there must be a db migration in between the rollback and used version, which in the most likely case will be just one version apart. The downgrade mechanism would then only be of use if there is a db migration in exactly this version increment and the two are compatible.

As to creating copies/zips on db update: We can do that. It will certainly annoy some people by "filling up" their `~/.config` with db backups. A warning on update informing about the creation of a db zip seems a bit aggressive, info logging would hardly be noticed.

PR #5023, opened 2018-06-21
gui: split listen addresses input field (fixes #4518)

### LINKED ISSUE #4518: Better way of handling enabling/disabling of default listen addresses
The current way of replacing the magic `default` string with something large and incomprehensible is, frankly, crap. This becomes more urgent as it's the only sane way to disable KCP. One suggestion:

screen shot 2017-11-17 at 09 51 33

qepasa:
### Purpose

Making listen addresses more accessible and expandable. Variation on issue description given the fact that KCP is no longer supported.

### Testing

Tested on default options (frontend converts combination of 'default' from and into corresponding relay and tcp listen addresses) and other non-default combinations

### Screenshots

image of connectons

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -58,6 +58,10 @@ ul+h5 {
     border-top: none;
 }
 
+.panel-body .form-group {
+    margin-bottom: 0;
+}
+
 .logo {
     margin: 0;
     padding: 0;
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -373,8 +373,19 @@ angular.module('syncthing.core')
         function updateLocalConfig(config) {
             var hasConfig = !isEmptyObject($scope.config);
 
+            var defaultTcpAddress = 'tcp://0.0.0.0:22000';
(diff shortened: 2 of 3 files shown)


AudriusButkevicius:
I think it's not as trivial as you think.
Doing filtering based on prefixes might not be as trivial as you think. 
There might be tcp4/6, relay, dynamic+http dynamic+https etc.

We should probably rethink how this is displayed in the UI, perhaps as a checkbox "default addresses", which then somehow shows which addresses would be injected by default etc, yet allow additional addresses somehow.

I guess a good continuation of this would be to add support disabling dialing via relays, disable dialing via tcp yet while still supporting listening, etc.

AudriusButkevicius:
Perhaps the "Default addresses" checkbox could have a tooltip, explaining what it would add to each of the address lines or something like that.

Not sure, we need a guy with UX skills for this, sadly I am not one of them.

Catfriend1:
Good start, @qepasa
Now the prefixes are known this could get a first step into the direction of improving the ui. Later, it could get further improved to handle future prefixes more flexible so changes to prefixes will not generate a have-to-remember to also change the split part constants.

st-review:
:robot: *beep boop*

PR #5026, opened 2018-06-23
lib/ignore: Add (?s) prefix to ignore patterns

imsodin:
### Purpose

Paths matching a pattern prefixed with `(?s)` will be skipped, even if there are
include patterns. This is useful e.g. if there are inaccessible files below a
path that would cause fs watcher errors or to speed up the scan.

Also I renamed `resultIncluded` to `resultIgnored` for clarity as discussed before.

### Testing

Adapted and extended the existing include/exclude/skip test.

### Documentation

Needs an update to https://docs.syncthing.net/users/ignoring.html

### DIFF
--- a/lib/fs/basicfs_watch.go
+++ b/lib/fs/basicfs_watch.go
@@ -37,14 +37,11 @@ func (f *BasicFilesystem) Watch(name string, ignore Matcher, ctx context.Context
 		eventMask |= permEventMask
 	}
 
-	if ignore.SkipIgnoredDirs() {
-		absShouldIgnore := func(absPath string) bool {
-			return ignore.ShouldIgnore(f.unrootedChecked(absPath))
-		}
-		err = notify.WatchWithFilter(filepath.Join(absName, "..."), backendChan, absShouldIgnore, eventMask)
-	} else {
-		err = notify.Watch(filepath.Join(absName, "..."), backendChan, eventMask)
+	watchShouldIgnore := func(absPath string) bool {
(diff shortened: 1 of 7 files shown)


AudriusButkevicius:
I don't like this idea.

I can understand the theoretical problem with the scans, but to be honest, stat calls are cheap so I don't think we should care.

I don't understand the watcher issue.

imsodin:
The watcher aborts setting up the recursive watchpoint when it encounters any errors along the way, e.g. a dir it has no read permission set. In a concrete example I have a shared folder with lots of application data. It also contains source and chroots, which (mostly) don't want to sync, except for one dir:

```
!/source/my-meta-packages
/source/*
/chroot
```

Before that was fine, because it did not recurse into chroot. With the new method it does and fails. Also it uses up a lot of unnecessary file descriptors.

In this specific example I could easily just move the unignored directory out of "source", but I like to think that Syncthing with ignores adapts to the users structure, not the other way around.

I generally agree about the efficiency, that was more of a side-thought, the above is the main issue. And we do get reports every now and then of people with really slow devices where already stating the entire tree takes a long time - once long than the before default 60s scan interval.

AudriusButkevicius:
Can we not make the watcher not fail in these cases?
Also, I guess it can be more clever about what it's watching inside the include dirs?

imsodin:
Yes and no. It is probably possible to handle errors gracefully and do the include/ignore with recursion in notify, but it would certainly be quite a bit of work and would make our fork deviate much more from upstream - that's not an option for me at the moment.

PR #5027, opened 2018-06-24
all: Add receive only folder type

calmh:
Adds the fabled receive only folder type. This is at the stage where it works for me, it needs some more polish and testing, but I'd like some early feedback on the mechanisms involved. There's a block comment on the receiveOnlyFolder type that describes the high level of how it works. It's a good entry point to reviewing.

I tweaked the iconography a little for symmetry.

screen shot 2018-06-24 at 10 11 49


calmh:
I chose a different name for revert because it's a different thing than override. Admittedly "override" could mean in either direction, but I think it's clearer to separate it to avoid misunderstandings. It also makes it clear from the outside what's going to happen in response to a `POST /rest/system/{override,revert}` instead of having to know what the folder type is to deduce the effect.

So the thing about the "synthetic" device ID... When a file has been created locally and we want to revert that, we need to delete it. The way to get the puller to delete something is to have a delete entry that is newer than what we have locally. Absent another way to solve it we need to fake an entry that becomes the best version, and we can't use the LocalDeviceID because that would mean it goes into the index as if we already had the delete. Hence the synthetic device ID that we use to fake the entry that we then need to sync.

AudriusButkevicius:
> So the thing about the "synthetic" device ID... When a file has been created locally and we want to revert that, we need to delete it. The way to get the puller to delete something is to have a delete entry that is newer than what we have locally. Absent another way to solve it we need to fake an entry that becomes the best version, and we can't use the LocalDeviceID because that would mean it goes into the index as if we already had the delete. Hence the synthetic device ID that we use to fake the entry that we then need to sync.

This feels like a hack. Things that should be deleted or revert, should just be deleted or reverted. I don't see why they would have to go in the index at all.

AudriusButkevicius:
> I chose a different name for revert because it's a different thing than override. Admittedly "override" could mean in either direction, but I think it's clearer to separate it to avoid misunderstandings. It also makes it clear from the outside what's going to happen in response to a POST /rest/system/{override,revert} instead of having to know what the folder type is to deduce the effect.

I also disagree with this. I'd rather document what override does in each folders case, rather than duplicate a ton of code and pollute the folder interface.

calmh:
Well, it is a bit of a hack to be sure. But consider a changed file on a receive only device. It's there, and it's changed, so it should be scanned and the result should go in the index. This drives things like conflict resolution and making sure we have a correctish version number if the folder type is changed from receive only to send-receive. Same goes for newly created files. It's there, so we scan it, and we can do conflict resolution if it also appears on another device that then announces it to us.

As the index is sort of append-only (per file) I think it would be a worse hack to retroactively remove things from our index and manually remove it from disk. It's also fairly nice that all disk updates are done by the puller, because we're sure that things like versioning and conflict resolution is always handled the same way.

But perhaps there is a better way to handle deletes, specifically. Maybe we could have another local flag, `LocalFlagShouldDelete`, that we set on our local index entry. These files magically get put on the need list and the puller interprets the flag as equivalent to there being an overriding global entry with the delete bit set.

calmh:
What's the ton of duplication?

lib/protocol/bep_extensions.go
```diff
@@ -145,7 +157,15 @@ func (f FileInfo) IsEmpty() bool {
 	return f.Version.Counters == nil
 }
 
-// IsEquivalent checks that the two file infos represent the same actual file content,
+func (f FileInfo) IsEquivalent(other FileInfo) bool {
+	return f.isEquivalent(other, false, false, 0)
+}
+
+func (f FileInfo) IsEquivalentOptional(other FileInfo, ignorePerms bool, ignoreBlocks bool, ignoreFlags uint32) bool {
```
imsodin: What is the reason for this function? Or differently: Is there ever a situation where a different flag than local change needs to be ignored or where the local change flag must not be ignored?

imsodin:
I think sending files with an empty version makes other devices think we are out of sync (for globally existing files). An alternative approach would be to add local changes to the index, but filter them out when sending.

As to the synthetic device: I think it's a nice hack, but I am not sure whether it is really worth it/necessary. Making sure that versioning is happening can also be achieved by sharing the file deletion code between the puller and reverting. I don't see how conflict resolution comes in with deleting files that only exist locally. Then the file info wouldn't need to be removed from index, it could just be made a "normal" (no local change flag) deleted file info.

Regarding `Revert` and `Override`: I was actually thinking that it would make sense to have both options on both "semi-directional" folder types, potentially even by file (that would be close to the mentioned manual folder). It seems totally plausible that you want to allow a certain change locally on a send-only folder or send something on a receive-only one. Therefore I agree with keeping two distinct names and rest endpoints.

lib/protocol/bep_extensions.go
calmh: I added the parameterless `IsEquivalent` because that's often the one you want, and not having to pass false, false, etc to get the default behavior is nice. The `...Optional` is the old one with all the parameters for when you need that.

Having the `ignoreFlags` as a parameter is necessary. The local change flag should not be ignored in equality comparisons unless you are doing the comparison in a receive only folder. The flow of flag bits is that the scanner gets a set of bits that it's supposed to set on newly discovered/changed files, and this set of bits also goes into the ignore set. This set of bits is zero during the whole process for send-receive and send-only folders and is `LocalFlagReceiveOnlyChanged` for receive-only folders.

calmh:
> I think sending files with an empty version makes other devices think we are out of sync (for globally existing files).

That's intentional. It indicates to other devices that the receive only device does not have the expected contents. Making the version empty means their change makes *them* out of sync, instead of making *us* out of sync.

> the synthetic device

Yeah since this is just for deleting, let me try doing the deletes directly in the revert code instead and see if it becomes messy. It might be fine and then we can drop that point.

uok:
Thanks for working in this feature! Since there are many (similar) folder icons now, I suggest to use a different icon for "folder id", e.g. `info-circle`

xjtdy888:
Can the readonly folder do not  file scanning?  folder scanning is a waste, in addition to block reused.
We only need to received device index and synchronize.

AudriusButkevicius:
No, we also need to offer an option to revert modified files, and to know that they are changed we need to scan. We get block reuse for free, so I don't see why we should not do that.

calmh:
Disabling scanning is an option on any folder, you can do that today but with the consequences Audrius mention.

lib/db/set.go
```diff
@@ -37,8 +37,12 @@ type FileSet struct {
 type FileIntf interface {
 	FileSize() int64
 	FileName() string
+	FileLocalFlags() uint32
```
AudriusButkevicius: It's already a file interface, it's clear that local flags belong to a file. If it conflicts with the field then do a java and call it GetLocalFlags or LocalFlagsSet

lib/db/leveldb_transactions.go
```diff
@@ -218,7 +218,7 @@ func (t readWriteTransaction) removeFromGlobal(gk, folder, device, file []byte,
 		if f, ok := t.getFile(folder, fl.Versions[0].Device, file); ok {
 			// A failure to get the file here is surprising and our
 			// global size data will be incorrect until a restart...
-			meta.addFile(globalDeviceID, f)
+			meta.addFile(protocol.GlobalDeviceID, f)
```
AudriusButkevicius: This whole rename could be in a separate pr, but not a big deal I guess.

lib/db/set.go
calmh: That's not what we did with the two attributes above it, `FileSize` and `FileName`.

calmh:
Just did a merge with master, the previous revision was approved so I'm going to go ahead and merge this now that master is aimed for .50, and then followup with a name cleanup PR for the file intf

PR #5029, opened 2018-06-24
lib/model: Release both locks when waiting for services to stop (fixes #5028)

imsodin:
### Purpose

Fixing #5028 as followup to #4982.

### Testing

None due to the racy nature of the issue.

PR #5030, opened 2018-06-24
lib/model, lib/weakhash: Abort pulling quicker on folder stop (ref #5028)

imsodin:
### Purpose

@xjtdy888 originally reported a deadlock when restarting a folder (changing config) while pulling in #5028. Further reports in the [PR fixing that](https://github.com/syncthing/syncthing/pull/5029) showed that stopping a folder can take a long time due to transmissions in progress. Therefore I am adding abort points at suitable places to speed stopping up. I split `pullerRoutine` up to be able to return early without having to care about the routine cleanup at every return.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -206,8 +206,13 @@ func (f *folder) Scan(subdirs []string) error {
 		subdirs: subdirs,
 		err:     make(chan error),
 	}
-	f.scanNow <- req
-	return <-req.err
+
+	select {
+	case f.scanNow <- req:
+		return <-req.err
+	case <-f.ctx.Done():
+		return nil
+	}
 }
 
 func (f *folder) Reschedule() {
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -211,7 +211,7 @@ func (f *folder) Scan(subdirs []string) error {
 	case f.scanNow <- req:
 		return <-req.err
 	case <-f.ctx.Done():
-		return nil
+		return f.ctx.Err()
 	}
 }
 
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -256,7 +256,7 @@ func (f *sendReceiveFolder) pullerIteration(ignores *ignore.Matcher, ignoresChan
 		doneWg.Done()
 	}()
 
-	changed, fileDeletions, dirDeletions := f.processNeeded(ignores, dbUpdateChan, copyChan, finisherChan, scanChan)
+	changed, fileDeletions, dirDeletions, err := f.processNeeded(ignores, dbUpdateChan, copyChan, finisherChan, scanChan)
 
 	// Signal copy and puller routines that we are done with the in data for
 	// this iteration. Wait for them to finish.
@@ -265,13 +265,14 @@ func (f *sendReceiveFolder) pullerIteration(ignores *ignore.Matcher, ignoresChan
 	close(pullChan)
 	pullWg.Wait()
 
-	// Signal the finisher chan that there will be no more input.
+	// Signal the finisher chan that there will be no more input and wait
+	// for it to finish.
 	close(finisherChan)
-
-	// Wait for the finisherChan to finish.
 	doneWg.Wait()
 
-	f.processDeletions(ignores, fileDeletions, dirDeletions, dbUpdateChan, scanChan)
+	if err == nil {
+		f.processDeletions(ignores, fileDeletions, dirDeletions, dbUpdateChan, scanChan)
+	}
 
 	// Wait for db updates and scan scheduling to complete
 	close(dbUpdateChan)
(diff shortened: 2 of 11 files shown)


imsodin:
Now protocol's `Connection.Request` also is stoppable by context. The transmission will continue, but as there is a connection drop on folder restart that's anyway going to be stopped.

It actually occurred that an alternative approach that wouldn't require any changes to the protocol package would be to drop connections before stopping the folder service. I don't see any problems with that - do you @AudriusButkevicius @calmh ?

AudriusButkevicius:
As long as nothing else bursts in flames I guess its ok

lib/model/folder_sendrecv.go
```diff
@@ -1109,7 +1156,14 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
...
 		}
 
+	blocks:
 		for _, block := range state.blocks {
+			select {
+			case <-f.ctx.Done():
+				break blocks
+			default:
+			}
+
```
AudriusButkevicius: Isn't this breakout imply we're done with success or something along those lines?

AudriusButkevicius: Should we state.fail(f.ctx.Err()) ?

imsodin: It shouldn't be necessary because the finisher only closes and "commits" a file if all blocks are accounted for, which isn't the case if we abort early here. However I agree, it's probably a good idea to make it explicit.

AudriusButkevicius: I guess you might get "Failed items" which is not ideal, but I am somehow afraid this will under some circumstances assume we're done downloading and rename the file to place.

AudriusButkevicius: The folder won't disappear from the UI, nor will the events about failed items in the buffers.

imsodin: Failed items are queried from a dedicated rest endpoint and are stored in `sendReceiveFolder.errors`, so they do get reset. Even if pulling on folder stop takes a long time, the folder error events would only be issued at the end of one puller iteration, but it will abort before that when the context is cancelled.

imsodin:
@AudriusButkevicius @calmh Where do we stand on this?

Together with PR #5029 it is v0.14.48 RC material, which is due to be released in less than a week. #5029 is entirely unproblematic. For this PR I don't expect any problems either, but that assessment isn't worth much. I'd propose including it in v0.18.48 nonetheless, as the alternative would be to revert https://github.com/syncthing/syncthing/commit/b7234785f8c31b4de1553b04877be07ec13d81fe and thus reintroduce a potential nil deref. Admittedly, that was present for a long time before and probably never bit anyone due to suture recovery.

calmh:
Why is this for 48? is #5028 introduced in that version?

calmh:
Oh I see, it goes back to #4982. So I'd missed that, which means we need a new 49 RC, which makes that release slip a while.

lib/model/folder.go
```diff
@@ -206,8 +206,13 @@ func (f *folder) Scan(subdirs []string) error {
...
 		err:     make(chan error),
 	}
-	f.scanNow <- req
-	return <-req.err
+
+	select {
+	case f.scanNow <- req:
+		return <-req.err
+	case <-f.ctx.Done():
+		return nil
```
calmh: return f.ctx.Error(), no? We have no reason to believe the scan succeeded.

lib/model/folder_sendrecv.go
```diff
@@ -311,6 +350,12 @@ func (f *sendReceiveFolder) pullerIteration(ignores *ignore.Matcher, ignoresChan
 		return true
 	})
 
+	select {
+	case <-f.ctx.Done():
+		return -1, nil, nil
```
calmh: Not sure why the `-1, nil, nil` return in these cases. Probably it doesn't matter because whoever is looking at the return is also going to exit, but if we've done 39 changes already I see no reason to lie about it?

imsodin: `-1, nil, nil` was just my way of returning an error - returning `changed, fileDeletions, dirDeletions` feels like lying too (we didn't necessarily change all of them). I think the clean thing to do here is add an error return value.

lib/model/folder_sendrecv.go
```diff
@@ -352,7 +353,7 @@ func (f *sendReceiveFolder) processNeeded(ignores *ignore.Matcher, dbUpdateChan
 
 	select {
 	case <-f.ctx.Done():
-		return -1, nil, nil
+		return 0, nil, nil, f.ctx.Err()
```
AudriusButkevicius: why 0?, why not just return the things we've processed?

AudriusButkevicius: @calmh?

AudriusButkevicius:
We can also now address this by upgrading suture

calmh:
I'll pull in the latest suture. With that and the other fix, are we good enough that I can drop a new 49 RC, with this pushed to 50?

AudriusButkevicius:
Yeah, it atleast should not deadlock.

lib/model/folder_sendrecv.go
calmh: So normally that's of course the case, but that's for a function that does one thing or fails. In this case I think we should legitimately return "I did 59 things, these are the results, *and then I failed* so here's also an error".

calmh:
OK coming up

imsodin:
With Audrius' suture fix it won't deadlock, but the changes in 49 won't then actually solve issue 4981 (which is in the changelog), because the global suture timeout of 10s will cancel `RemoveAndWait` before the puller actually stopped if pulling is going on and taking a long time.

AudriusButkevicius:
No it won't. That's what I fixed in suture. It will correctly block until the service is down, and no longer indefinitely even after the service stopped. The 10s timeout no longer has any influence.

imsodin:
Yes, right. It will still hit the following branch:
https://github.com/AudriusButkevicius/suture/blob/daf204b831cd75ed3c45db07368087e7ad9eed2a/supervisor.go#L561
```golang
			case <-s.getAfterChan(s.timeout):
                            s.LogBadStop(s, namedService.Service, namedService.name)
```
but that just results in a unnecessary warning that service failed to terminate timely, but it won't affect `WaitAndRemove` (which I wrongly thought it would). Thanks!

calmh:
(good to merge from my pov once you guys are happy)

PR #5035, opened 2018-06-26
angular.js: restore license

aviau:
It was removed in the last vendor update.

As requested by the MIT license:

> The above copyright notice and this permission notice shall be included in
> all copies or substantial portions of the Software.

### DIFF
--- a/gui/default/vendor/angular/angular-sanitize.js
+++ b/gui/default/vendor/angular/angular-sanitize.js
@@ -2,6 +2,25 @@
  * @license AngularJS v1.3.20
  * (c) 2010-2014 Google, Inc. http://angularjs.org
  * License: MIT
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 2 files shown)


aviau:
Anything I need to do? Looks like there is a failing build.

imsodin:
@aviau That's not a problem with your changes. Either it's related to the current snap problems or your branch name just happened to be the first one to exceed some weird limit:

```
Step 7/8: build snaps (Command Line) (2s)
[22:45:26][Step 7/8] Starting: /opt/buildagent/syncthing-2-temp/agentTmp/custom_script7789799283595263934
[22:45:26][Step 7/8] in directory: /opt/buildagent/syncthing-2-work/src/github.com/syncthing/syncthing
[22:45:28][Step 7/8] Issues while validating None: The 'version' property does not match the required schema: '0.14.49-rc.2+11-g223b508-angular-license' is too long (maximum length is 32)
```

@calmh Do you know what that limit is?

aviau:
hah right. Looks like my branch name is too long :o)

calmh:
That's a snapcraft limitation, it says 32 characters.

That said - "restore" implies something was removed; these files are exactly as distributed by the Angular project. What's the worry here, that someone might not understand the "License: MIT" in the files? If so, I think the right avenue is to take it up with Angular, as this change will be reverted any time we upgrade anything.

aviau:
> "restore" implies something was removed

Some were removed during an update in 3b5e1fa0fcd9bc6f1e29b8625b3c4756345c9f94.

>  these files are exactly as distributed by the Angular project

Yes that is true, I don't know why they don't include the full license text here. Probably just because they don't care. It is their software so they can do whatever they want. However that does not mean that Syncthing can distribute Angular without respecting the license terms.

Realistically Google is never going to sue you, however I cannot say "Google isn't going to sue us, it is fine!" when I upload something in the Debian archive.

> What's the worry here, that someone might not understand the "License: MIT" in the files

I am not the author of the license so I can't really tell exactly why naming the license or linking to it is not enough, but the license is very clear that it has to be copied verbatim:

> The above copyright notice and this permission notice shall be included in
> all copies or substantial portions of the Software.


>  Other option would be to just drop a license.txt here with the MIT license.

Yes, that works fine too, I'll open a PR.

PR #5041, opened 2018-06-30
Add configurable UID and GID to Docker container

nvllsvm:
### Purpose
Allows for configuring the UID and GID Syncthing runs as in the container. Uses [su-exec](https://github.com/ncopa/su-exec) from the Alpine repos to accomplish this. Addition of su-exec results in <2MB increase in image size.

### Testing
I've been using the [same approach](https://github.com/nvllsvm/docker-syncthing/blob/master/entrypoint.sh#L4-L5) in my Syncthing image for sometime on systems with varying UIDs and GIDs without issue.

I performed the following tests, each with a freshly created folder on the host mounted as volume /var/syncthing:

- started container without environment variables overridden. Verified configuration files written to the host folder had UID 1000 and GID 1000.
- started container with PUID=123 and PGID=456. Verified configuration files written to the host folder had UID 123 and GID 456.

### Documentation
Included in this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,15 +18,15 @@ RUN apk add --no-cache ca-certificates
 
 COPY --from=builder /go/src/github.com/syncthing/syncthing/syncthing /bin/syncthing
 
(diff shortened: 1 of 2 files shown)


calmh:
Neat

PR #5042, opened 2018-07-01
vendor: Add missing file to github.com/syncthing/notify

imsodin:
Followup to #5032 where I forgot to git add this file.


AudriusButkevicius:
This is just a dep update?

imsodin:
Yes, the linked PR is a dep update, where a new file was introduced, which I missed (just used `git commit -a`). That's the file in this PR.

AudriusButkevicius:
Just vendor: Update x then?

Issue #4961, opened 2018-05-21
Getting tons of "The system cannot find the path specified." in the log


vladipus:
Getting just tons of
`[RQWCQ] 20:35:47 INFO: Puller (folder "My Folder" (*****-*****), file "folder\path"): dir mkdir: CreateFile C:\sync\folder\path folder\path: The system cannot find the path specified.`

 - what happened,
Deleted a folder.
 - what you expected to happen instead, and
Re-create the missing directory structure or not sync the deleted folder at all.
 - any steps to reproduce the problem.
Use Syncthing in a real multi-user environment.

### Version Information

Syncthing Version: v0.14.47
OS Version: Windows 10

vladipus:
This error is pursuing me for ages in all sorts of Syncthing versions, on all sorts of folders I'm using at home and at work. Nothing helps, including the rescans. Recreating the directory structure by hand is too cumbersome.

calmh:
> any steps to reproduce the problem.
> Use Syncthing in a real multi-user environment.

While flippancy is nice and all and understandable in frustration, we have >150k users and this is the first report of this exact nature. Some steps would be good.

vladipus:
Great respect for the software, no doubt, but this really drives me nuts.

We are not doing anything particular, actually. We just regularly delete and add tons of different files and folders. As a programmer, to date, I can't understand why Syncthing won't re-create a directory tree in a `mkdir -p` way, if it needs to.

vladipus:
For example, I just deleted a Music folder and see:
`
Music\iTunes\Album Artwork\Cache\5C0ADFAFE000EA85\11\08\11 | The system cannot find the path specified.
-- | --


`

canton7:
> If it cannot find the path specified, why doesn't it just CREATE it??? 

Please help us figure that out 🙂 

Reading [this](https://github.com/syncthing/syncthing/blob/6802505dda3451804a8790b00c13161a782bf5f0/lib/fs/basicfs_windows.go#L36), it looks like it fails if the folder's parent's parent doesn't exist? It looks like MkdirAll has now been fixed upstream, but not yet released.

vladipus:
Right now I'm doing he same thing by hand with:
user@DESKTOP-USRPNKK MSYS /z/FFZ-RW/Branding
$ mkdir -p "Music/iTunes/Album Artwork/Cache/5C0ADFAFE000EA85/11/08/11"

[...]

PR #5048, opened 2018-07-04
 lib/model: Also handle missing parent dir non-regular items 

imsodin:
### Purpose

This is an improvemnt of PR #4493 and related to (and maybe fixing) #4961 and #4475. Maybe fixing, because there is no clear reproducer for that problem.

The previous PR added a mechanism to resurrect missing parent directories, if there is a valid child file to be pulled. The same mechanism does not exist for dirs and symlinks, even though a missing parent can happen for those items as well. Therefore this PR extends the resurrection to all types of pulled items.

In addition I moved the `IsDeleted` branch while iterating over `processDirectly` to the existing `IsDeleted` branch in the `WithNeed` iteration. This saves one pointless assigmenet and `IsDeleted` query. Also 

### Testing

New request test to check that the parent of a directory is resurrected.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -288,6 +288,9 @@ func (f *sendReceiveFolder) processNeeded(ignores *ignore.Matcher, dbUpdateChan 
 
 	changed := 0
 	var processDirectly []protocol.FileInfo
+	var dirDeletions []protocol.FileInfo
+	fileDeletions := map[string]protocol.FileInfo{}
+	buckets := map[string][]protocol.FileInfo{}
 
 	// Iterate the list of items that we need and sort them into piles.
 	// Regular files to pull goes into the file queue, everything else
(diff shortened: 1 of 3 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -640,6 +610,40 @@ func (f *sendReceiveFolder) handleDir(file protocol.FileInfo, dbUpdateChan chan<
 	}
 }
 
+// checkParent verifies that the thing we are handling lives inside a directory,
+// and not a symlink or regular file. It also resurrects missing parent dirs.
+func (f *sendReceiveFolder) checkParent(file string, scanChan chan<- string) bool {
+	parent := filepath.Dir(file)
+
+	if err := osutil.TraversesSymlink(f.fs, parent); err != nil {
+		f.newError("traverses q", file, err)
```
calmh: "traverses q"? (I get that this is probably a pre-existing error message, but wat does it mean

AudriusButkevicius: Was probably q for queue

PR #5049, opened 2018-07-05
cmd/stcli: Add config command and output pretty JSON

xor-gate:
### Purpose

The `stcli` now knows a new trick to get and prettyprint the config in JSON format and print it pretty on the console. This very handy for development and remote debugging.

### Testing

It seems the stcli tool has no tests (manual only). The addition can be manually tested as follows:

```
$ stcli --apikey <apikey> config
{
    "devices": [
        {
            "addresses": [
                "dynamic"
            ],
            "allowedNetworks": [],
            "autoAcceptFolders": false,
            "certName": "",
            "compression": "metadata",
.......
}
```

### Documentation

I have searched through the docs but it seems the `stcli` tool is undocumented. So I have no documentation to add.

### DIFF
--- a/cmd/stcli/cmd_general.go
+++ b/cmd/stcli/cmd_general.go
@@ -3,9 +3,9 @@
 package main
 
 import (
-	"os"
 	"encoding/json"
 	"fmt"
+	"os"
 
 	"github.com/AudriusButkevicius/cli"
 )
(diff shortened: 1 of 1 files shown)


calmh:
Yeah that tool is untested, undocumented and out of date. Efforts to make it otherwise are certainly appreciated! I'll add you to the author list. The other error is wanting a `gofmt` of the code, and I'd like to recommend a two space JSON indent like we use in other places but I don't feel strongly about it.

calmh:
Oh, and, generally, branch in your own fork rather than the main repo please, to avoid too much litter. 👍 The exception being if you want a build server build of your branch but not for to be a PR, but this is uncommon.

xor-gate:
Seems fine for me I will fix it. I have read the contribution guidelines about branches but this was not 100% clear for me so we should also add it there we must use forks with branches instead of littering the main repos with branches (here https://docs.syncthing.net/dev/contributing.html#branches).

calmh:
Hm, yes. Most people don't have any other choice at the beginning, and then the habit is formed. ;) It also results in double builds on the server.

PR #5054, opened 2018-07-09
lib/db: Catch unignored/conflicting files as needed (fixes #5053)

### LINKED ISSUE #5053: Unignored file not considered needed
@AudriusButkevicius @calmh This is yet another issue introduced in https://github.com/syncthing/syncthing/pull/4958 and thus 0.14.49 material - I am really sorry about that disaster and slow piece-by-piece resolution. It also needs another db transition while there is already one on master, that's not in release - I will file separate PRs for master and release branches to address that. Regarding release schedules: May it be reasonable to delay 0.14.50, but already release 0.14.51-rc.1 as normal?

Originally came up in https://forum.syncthing.net/t/did-something-recently-changed-in-ignores-handling/11714/35

Steps to reproduce:

1. Share a folder between devices A and B.
2. Modify file F on A, wait for sync.
3. Modify file F on B, wait for sync.
4. Ignore F on A.
5. Edit F on A and B.
6. Unignore F on A.

Expected: Create and sync conflicts.

Actual: Nothing is synced and no conflicts are created.

imsodin:
### Purpose

Fixes #5053 by doing the check, if the globally updated file needs to be added/removed from the need bucket, before updating the meta counter. Thus it happens on every update, not only if the first item in the global list is affected by the update. An update can be relevant to what's locally needed without happening at the front of the global list, if it involves conflicting files.

This also requires a recalculation of the need bucket. As the local flags change was introduced in dbVersion 5, but is not yet introduced in the release branch, dbVersion 5 needs to be skipped. The local flags change now happens to 7 (for master), while the need bucket reset happens to 6 (for release). Thus even if someone is running master, the transition will happen correctly. 

The diff looks bigger than it is, it's just moving the need bucket part of the code ahead.

### Testing

New unit test.

### DIFF
--- a/lib/db/leveldb_transactions.go
+++ b/lib/db/leveldb_transactions.go
@@ -99,11 +99,31 @@ func (t readWriteTransaction) updateGlobal(gk, folder, device []byte, file proto
 
 	name := []byte(file.Name)
 
-	if removedAt != 0 && insertedAt != 0 {
-		if bytes.Equal(device, protocol.LocalDeviceID[:]) && file.Version.Equal(fl.Versions[0].Version) {
-			l.Debugf("local need delete; folder=%q, name=%q", folder, name)
-			t.Delete(t.db.needKey(folder, name))
+	var newGlobal protocol.FileInfo
+	if insertedAt == 0 {
+		// Inserted a new newest version
+		newGlobal = file
+	} else if new, ok := t.getFile(folder, fl.Versions[0].Device, name); ok {
+		// The previous second version is now the first
+		newGlobal = new
+	} else {
+		panic("This file must exist in the db")
+	}
+
+	// Fixup the list of files we need.
+	nk := t.db.needKey(folder, name)
+	hasNeeded, _ := t.db.Has(nk, nil)
+	if localFV, haveLocalFV := fl.Get(protocol.LocalDeviceID[:]); need(newGlobal, haveLocalFV, localFV.Version) {
+		if !hasNeeded {
+			l.Debugf("local need insert; folder=%q, name=%q", folder, name)
+			t.Put(nk, nil)
 		}
+	} else if hasNeeded {
+		l.Debugf("local need delete; folder=%q, name=%q", folder, name)
+		t.Delete(nk)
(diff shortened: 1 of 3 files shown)


lib/db/leveldb_transactions.go
```diff
@@ -149,6 +123,29 @@ func (t readWriteTransaction) updateGlobal(gk, folder, device []byte, file proto
...
+		oldGlobalFV = removedFV
+	} else if len(fl.Versions) > 1 {
+		// The previous newest version is now at index 1
+		oldGlobalFV = fl.Versions[1]
+	}
+	if oldFile, ok := t.getFile(folder, oldGlobalFV.Device, name); ok {
+		// A failure to get the file here is surprising and our
+		// global size data will be incorrect until a restart...
+		meta.removeFile(globalDeviceID, oldFile)
+	}
```
AudriusButkevicius: Else panic to weed out the bugs?

lib/db/leveldb_transactions.go
```diff
@@ -149,6 +123,29 @@ func (t readWriteTransaction) updateGlobal(gk, folder, device []byte, file proto
 		t.Delete(nk)
 	}
 
+	if removedAt != 0 && insertedAt != 0 {
+		l.Debugf(`new global for "%v" after update: %v`, file.Name, fl)
+		t.Put(gk, mustMarshal(&fl))
```
AudriusButkevicius: Why is the delete gone?

lib/db/leveldb_transactions.go
imsodin: I agree, but lets do that in a different commit to separate the potential ensuing madness from this fix :)

lib/db/leveldb_transactions.go
imsodin: All need bucket operations, including the edge case that was handled at this point before, are now handled in the block that is above (https://github.com/imsodin/syncthing/blob/6e2b90e7701b1ec86a3c4afb758d0afe9d674532/lib/db/leveldb_transactions.go#L113)

PR #5055, opened 2018-07-09
lib/db: Catch unignored/conflicting files as needed (fixes #5053)

### LINKED ISSUE #5053: Unignored file not considered needed

imsodin:
Fixes #5053 on the release branch, skipping db version 5 as it is already used on master. See #5054 (PR to master) for more info.

### DIFF
--- a/lib/db/leveldb_dbinstance_updateschema.go
+++ b/lib/db/leveldb_dbinstance_updateschema.go
@@ -13,7 +13,7 @@ import (
 	"github.com/syndtr/goleveldb/leveldb/util"
 )
 
-const dbVersion = 6
+const dbVersion = 5
 
 func (db *Instance) updateSchema() {
 	miscDB := NewNamespacedKV(db, string(KeyTypeMiscData))
@@ -34,9 +34,9 @@ func (db *Instance) updateSchema() {
 	if prevVersion < 3 {
 		db.updateSchema2to3()
 	}
-	// This update fixes problems existing in versions 3 to 5
-	if prevVersion < 6 && prevVersion > 2 {
-		db.updateSchema3to4()
+	// This update fixes problems existing in versions 3 and 4
+	if prevVersion == 3 || prevVersion == 4 {
+		db.updateSchemato5()
 	}
 
 	miscDB.PutInt64("dbVersion", dbVersion)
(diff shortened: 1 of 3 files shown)


lib/db/leveldb_dbinstance_updateschema.go
```diff
@@ -34,8 +34,8 @@ func (db *Instance) updateSchema() {
 	if prevVersion < 3 {
 		db.updateSchema2to3()
 	}
-	// This update fixes a problem that only exists in dbVersion 3.
-	if prevVersion == 3 {
+	// This update fixes problems existing in versions 3 to 5
+	if prevVersion < 6 && prevVersion > 2 {
 		db.updateSchema3to4()
```
calmh: The stuff that is version five on master then needs to become version seven? This feels weird. Can we not assume that noone runs master, increase master to six, and let this be five so the sequence makes sense for users going 48 -> 49 -> 50 (released versions)?

lib/db/leveldb_dbinstance_updateschema.go
```diff
@@ -34,9 +34,9 @@ func (db *Instance) updateSchema() {
 	if prevVersion < 3 {
 		db.updateSchema2to3()
 	}
-	// This update fixes a problem that only exists in dbVersion 3.
-	if prevVersion == 3 {
-		db.updateSchema3to4()
+	// This update fixes problems existing in versions 3 and 4
+	if prevVersion == 3 || prevVersion == 4 {
+		db.updateSchemato5()
```
canton7: "Schemato"? Is that like a tomato?

calmh:
Negatory on this, as I noted on #5059. Lets fix up master so that we are happy, and try again with an RC from there, including whatever features and bugfixes has gone in in the meantime. The 49 RC slips as necessary to do this.

PR #5056, opened 2018-07-09
lib/fs: Catch size-preserving changes on windows (fixes #5050)

### LINKED ISSUE #5050: Filesystem watcher doesn't notify changes on Windows when the file size remains the same
On Windows, if a file is modified but the file size remains the same, the file system watcher doesn't detect the change. Syncthing won't sync the changes to the file until the next folder scan.

I can reproduce this issue by opening a text file with Notepad, changing one of the characters (leaving the length of the file the same) and saving the file. With the `NOTIFY_DEBUG` environment variable enabled, the Syncthing logs don't show a change having been detected (I would expect to see 'dispatching notify.FileActionModified').

I suspect this issue can be fixed by including [`FILE_NOTIFY_CHANGE_LAST_WRITE`](https://docs.microsoft.com/en-us/windows/desktop/api/WinBase/nf-winbase-readdirectorychangesw#FILE_NOTIFY_CHANGE_LAST_WRITE) in the filter passed to `ReadDirectoryChanges` in [`watcher_readdcw.go`](https://github.com/syncthing/notify/blob/master/watcher_readdcw.go). This will generate an event when the last-write time of a file is changed.

### Version Information

Syncthing Version: v0.14.48
OS Version: Windows 10 64 bit OS Build 17134.112
Browser Version: N/A

imsodin:
Fixes #5050

### DIFF
--- a/lib/fs/basicfs_watch_eventtypes_readdcw.go
+++ b/lib/fs/basicfs_watch_eventtypes_readdcw.go
@@ -11,7 +11,7 @@ package fs
 import "github.com/syncthing/notify"
 
 const (
-	subEventMask  = notify.FileNotifyChangeFileName | notify.FileNotifyChangeDirName | notify.FileNotifyChangeSize | notify.FileNotifyChangeCreation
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Needs a dep update?

imsodin:
Nope, that's our own filesystem watcher code.

PR #5059, opened 2018-07-09
lib/model: Revert rechecking for now

AudriusButkevicius:
Reverts it for .49, while the FlagsLocal business goes in and fixes the conflict issues.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1355,53 +1355,7 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset 
 		return protocol.ErrGeneric
 	}
 
-	if !scanner.Validate(buf, hash, weakHash) {
-		m.recheckFile(deviceID, folderFs, folder, name, int(offset)/len(buf), hash)
-		return protocol.ErrNoSuchFile
-	}
-
 	return nil
-}
-
-func (m *Model) recheckFile(deviceID protocol.DeviceID, folderFs fs.Filesystem, folder, name string, blockIndex int, hash []byte) {
-	cf, ok := m.CurrentFolderFile(folder, name)
-	if !ok {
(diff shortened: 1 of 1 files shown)


lib/model/model.go
```diff
@@ -1355,55 +1355,9 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset
 		return protocol.ErrGeneric
 	}
 
-	if !scanner.Validate(buf, hash, weakHash) {
-		m.recheckFile(deviceID, folderFs, folder, name, int(offset)/len(buf), hash)
-		return protocol.ErrNoSuchFile
-	}
```
imsodin: I'd just remove `m.rcheckFile`, but keep validating and returning an error if it fails.

calmh:
So apart from the merge conflicts which I'm sure are understandable this doesn't build when applied directly to the release branch. 

```
# github.com/syncthing/syncthing/lib/model
lib/model/model.go:10:2: imported and not used: "bytes"
lib/model/model_test.go:3632:3: m.recheckFile undefined (type *Model has no field or method recheckFile)
lib/model/model_test.go:3633:3: m.recheckFile undefined (type *Model has no field or method recheckFile)
lib/model/model_test.go:3634:3: m.recheckFile undefined (type *Model has no field or method recheckFile)
```

I can trivially fix that, but it's not enormously confidence inspiring...

calmh:
@AudriusButkevicius Are we ok with this as it exists on `master` at the moment, or do we require a fix of some kind there? (That is, can I release a .49-rc from `master` right now, or should I wait?)

AudriusButkevicius:
It should be ok in master, atleast so @imsodin claims, as in your local flags stuff should fix it, yet I never understood what the issue was, as I thought invalid files should not cause conflicts

PR #5066, opened 2018-07-14
lib/model: Record error for unavailable files

imsodin:
### Purpose

Currently unavailable files show up as out-of-sync in the UI without any progress
or explanation. Now these files will be added to the failed items. The changed
count isn't inreased, as there is no point in retrying those files until a change
in connection happens. At that point, a pull will be scheduled anyway.

### Testing

None.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -66,6 +66,7 @@ var (
 	errDirHasToBeScanned   = errors.New("directory contains unexpected files, scheduling scan")
 	errDirHasIgnored       = errors.New("directory contains ignored files (see ignore documentation for (?d) prefix)")
 	errDirNotEmpty         = errors.New("directory is not empty")
+	errNotAvailable        = errors.New("no peers that have this file are connected. Will be retried on next connection.")
 )
 
 const (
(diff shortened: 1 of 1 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -353,7 +354,7 @@ func (f *sendReceiveFolder) processNeeded(ignores *ignore.Matcher, dbUpdateChan
 					return true
 				}
 			}
-			l.Debugln(f, "Needed file is unavailable", file)
+			f.newError("checking availability", file.Name, errNotAvailable)
```
AudriusButkevicius: Perhaps just `pull` rather than check availability.

lib/model/folder_sendrecv.go
```diff
@@ -66,6 +66,7 @@ var (
 	errDirHasToBeScanned   = errors.New("directory contains unexpected files, scheduling scan")
 	errDirHasIgnored       = errors.New("directory contains ignored files (see ignore documentation for (?d) prefix)")
 	errDirNotEmpty         = errors.New("directory is not empty")
+	errNotAvailable        = errors.New("no peers that have this file are connected. Will be retried on next connection.")
```
AudriusButkevicius: Very long, The crux here is that `no connected device has this version of the file`

AudriusButkevicius: Or even "latest version of the file"

lib/model/folder_sendrecv.go
```diff
@@ -353,7 +353,7 @@ func (f *sendReceiveFolder) processNeeded(ignores *ignore.Matcher, dbUpdateChan
 					return true
 				}
 			}
-			f.newError("processing needed", file.Name, errNotAvailable)
+			f.newError("pull", file.Name, errNotAvailable)
```
AudriusButkevicius: We could also add file.ModifiedBy to the error message.

PR #5068, opened 2018-07-16
cmd/strelaypoolsrv: Prevent scraped metrics moving backwards

AudriusButkevicius:
Comment explains it.
Not tested, as usual.

### DIFF
--- a/cmd/strelaypoolsrv/stats.go
+++ b/cmd/strelaypoolsrv/stats.go
@@ -44,6 +44,8 @@ var (
 	relayGlobalRate         = makeGauge("relay_global_rate", "Global rate applied on the whole relay", "relay")
 	relayBuildInfo          = makeGauge("relay_build_info", "Build information about a relay", "relay", "go_version", "go_os", "go_arch")
 	relayLocationInfo       = makeGauge("relay_location_info", "Location information about a relay", "relay", "city", "country", "continent")
+
+	lastStats = make(map[string]stats)
 )
 
 func makeGauge(name string, help string, labels ...string) *prometheus.GaugeVec {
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
@calmh

cmd/strelaypoolsrv/stats.go
```diff
@@ -210,4 +218,34 @@ func deleteMetrics(host string) {
...
+	new.Options.GlobalRate = mergeValue(new.Options.GlobalRate, old.Options.GlobalRate)
+	return new
+}
+
+func mergeValue(new, old int) int {
+	if new >= old || old == 0 || new == 0 {
+		return new
+	}
+	delta := (float64(old-new) * 100) / float64(new)
+	if delta > 1.0 {
```
imsodin: Does this intentionally return `new` if `old` is 1% *larger* than `new` or is my brain failing? And what's the 1% thing anyway? From the comment on `mergeStats` I'd expect a simple `if new < old { return old } return new`.

AudriusButkevicius: We still need to capture actual rate resets, if I restart the relay server, the number has to go down.

imsodin: Right. and the sign business, i.e. `old-new` vs `new-old`? To me it looks like `new` is returned for `old > 1.01 * new` and `old` in all other cases - which would be wrong.

calmh:
```
if new > old {
  return new
}
if new > 0.99 * old {
  return old // slight backwards movement
}
return new // counter reset
}
```
?

AudriusButkevicius: image

imsodin:
```
In [1]: def merge(new, old):
   ...:     delta = (float(old-new)*100) / float(new)
   ...:     if delta > 1.0:
   ...:         return new
   ...:     return old
   ...: 

In [2]: merge(1001, 1000)
Out[2]: 1000
```

AudriusButkevicius: :(

calmh:
This ought to be easy to add a test for...

AudriusButkevicius:
done

Issue #5063, opened 2018-07-11
panic: cannot start already running folder


JulienPalard:
Tried to add a third shared folder between my phone and my laptop. Initiated the share from my phone, Syncthing paniced on my laptop, and now synthings panics while starting.

### Version Information

- syncthing v0.14.48 "Dysprosium Dragonfly" (go1.10.2 linux-amd64) teamcity@build.syncthing.net 2018-05-14 06:53:06 UTC
- On Debian buster

### Error logs

```
[...]
17:47:03 INFO: Adding folder "Caméra" (nexus_6p-photos)
Panic at 2018-07-11T17:47:03+02:00
panic: cannot start already running folder nexus_6p-photos
[...]
```

AudriusButkevicius:
Check your config, you probably have two folders with the same id.

JulienPalard:
Hi @AudriusButkevicius and thanks for your prompt answer, greatly appreciated!

I though id were auto-generated, I just checked android-side and (if I get it right) I have 4 shares, which ids are:

- nexus_6p-photos
- default
- mv3sz-cniao
- 5z1xy-bg5iv

I don't know how to check the ids on my computer as the app won't start from now, but the app started to crash when I added a directory from my phone, and I have only two devices : my laptop and my phone.

I do not see an id conflict, please re-open.

(I'm new to syncthing, I don't expect a software to panic on an error you can make me fix by telling me a single simple line, please re-open the issue. It may not be an issue as you understand it, but it's at least an UX issue that needs to be fixed, I mean, my mother won't be able to use a software that crash and don't tell how to recover on simple errors, she won't be able to write english on a forum or whatever).

Bests

AudriusButkevicius:
The discussion on this working out the issue should happen on the forum, so this will not be reopened until we work out the actual cause so there is a concrete thing to fix.

JulienPalard:
Sorry to report issue.

AudriusButkevicius:
Every potential bug should first be debugged and understood in the forum, unless you know exactly what the cause is and what has to be fixed.

calmh:
I disagree; a panic is a bug, and we should have an open ticket on it regardless of not yet understanding the root cause.

AudriusButkevicius:
Sure, lets leave this to rot eternally then. 

For all that I know the user might have edited the config by hand and caused this. Is it still a bug in that case? 

That's why this needs a conversation on the forum first, as this just talks about consequences, not the details how this happened.

[...]

PR #5069, opened 2018-07-16
lib/config: Copy inputs and copy config outputs

AudriusButkevicius:
This is a guess as to what could be the cause for #5063.

In the last debug logs the relevant part was:

```
[GXVUP] 22:26:43 INFO: Existing folder default false
[GXVUP] 22:26:43 INFO: Existing folder gdrxc-frazc false
[GXVUP] 22:26:43 INFO: Existing folder jjnhx-nhqzq false
[GXVUP] 22:26:43 INFO: Missing folder ewa2m-zcpje

// CommitConfiguration in the model coming in:

[GXVUP] 22:26:43 INFO: from folder map[
default:...
ewa2m-zcpje:...
gdrxc-frazc:...


[GXVUP] 22:26:43 INFO: to folder map[
ewa2m-zcpje:...
gdrxc-frazc:...
jjnhx-nhqzq:...
default:...

[GXVUP] 22:26:43 INFO: Adding folder "test1" (jjnhx-nhqzq)
```

So `ewa2m-zcpje` was introduced, but it somehow overwrote an existing folders spot in the config.

### DIFF
--- a/lib/config/wrapper.go
+++ b/lib/config/wrapper.go
@@ -265,7 +265,7 @@ func (w *Wrapper) Folders() map[string]FolderConfiguration {
 func (w *Wrapper) FolderList() []FolderConfiguration {
 	w.mut.Lock()
 	defer w.mut.Unlock()
-	return append(nil, w.cfg.Copy().Folders...)
+	return w.cfg.Copy().Folders
 }
 
 // SetFolder adds a new folder to the configuration, or overwrites an existing
--- a/lib/config/wrapper.go
+++ b/lib/config/wrapper.go
@@ -73,7 +73,7 @@ type Wrapper struct {
 // disk.
 func Wrap(path string, cfg Configuration) *Wrapper {
 	w := &Wrapper{
-		cfg:  cfg,
+		cfg:  cfg.Copy(),
 		path: path,
 		mut:  sync.NewMutex(),
 	}
@@ -142,7 +142,7 @@ func (w *Wrapper) RawCopy() Configuration {
 func (w *Wrapper) Replace(cfg Configuration) (Waiter, error) {
 	w.mut.Lock()
 	defer w.mut.Unlock()
-	return w.replaceLocked(cfg)
+	return w.replaceLocked(cfg.Copy())
 }
 
 func (w *Wrapper) replaceLocked(to Configuration) (Waiter, error) {
@@ -154,7 +154,7 @@ func (w *Wrapper) replaceLocked(to Configuration) (Waiter, error) {
 
 	for _, sub := range w.subs {
 		l.Debugln(sub, "verifying configuration")
-		if err := sub.VerifyConfiguration(from, to); err != nil {
+		if err := sub.VerifyConfiguration(from.Copy(), to.Copy()); err != nil {
 			l.Debugln(sub, "rejected config:", err)
 			return noopWaiter{}, err
 		}
@@ -187,15 +187,14 @@ func (w *Wrapper) notifyListener(sub Committer, from, to Configuration) {
 	}
 }
 
-// Devices returns a map of devices. Device structures should not be changed,
-// other than for the purpose of updating via SetDevice().
+// Devices returns a map of devices.
 func (w *Wrapper) Devices() map[protocol.DeviceID]DeviceConfiguration {
 	w.mut.Lock()
 	defer w.mut.Unlock()
 	if w.deviceMap == nil {
 		w.deviceMap = make(map[protocol.DeviceID]DeviceConfiguration, len(w.cfg.Devices))
 		for _, dev := range w.cfg.Devices {
-			w.deviceMap[dev.DeviceID] = dev
+			w.deviceMap[dev.DeviceID] = dev.Copy()
 		}
 	}
 	return w.deviceMap
@@ -213,13 +212,13 @@ func (w *Wrapper) SetDevices(devs []DeviceConfiguration) (Waiter, error) {
 		replaced = false
 		for newIndex := range newCfg.Devices {
 			if newCfg.Devices[newIndex].DeviceID == devs[oldIndex].DeviceID {
-				newCfg.Devices[newIndex] = devs[oldIndex]
+				newCfg.Devices[newIndex] = devs[oldIndex].Copy()
 				replaced = true
 				break
 			}
 		}
 		if !replaced {
-			newCfg.Devices = append(newCfg.Devices, devs[oldIndex])
+			newCfg.Devices = append(newCfg.Devices, devs[oldIndex].Copy())
 		}
 	}
 
@@ -238,19 +237,14 @@ func (w *Wrapper) RemoveDevice(id protocol.DeviceID) (Waiter, error) {
 	defer w.mut.Unlock()
 
 	newCfg := w.cfg.Copy()
-	removed := false
 	for i := range newCfg.Devices {
 		if newCfg.Devices[i].DeviceID == id {
 			newCfg.Devices = append(newCfg.Devices[:i], newCfg.Devices[i+1:]...)
-			removed = true
-			break
+			return w.replaceLocked(newCfg)
 		}
 	}
-	if !removed {
-		return noopWaiter{}, nil
-	}
 
-	return w.replaceLocked(newCfg)
+	return noopWaiter{}, nil
 }
 
 // Folders returns a map of folders. Folder structures should not be changed,
@@ -261,7 +255,7 @@ func (w *Wrapper) Folders() map[string]FolderConfiguration {
 	if w.folderMap == nil {
 		w.folderMap = make(map[string]FolderConfiguration, len(w.cfg.Folders))
 		for _, fld := range w.cfg.Folders {
-			w.folderMap[fld.ID] = fld
+			w.folderMap[fld.ID] = fld.Copy()
 		}
 	}
 	return w.folderMap
@@ -271,7 +265,7 @@ func (w *Wrapper) Folders() map[string]FolderConfiguration {
 func (w *Wrapper) FolderList() []FolderConfiguration {
 	w.mut.Lock()
 	defer w.mut.Unlock()
-	return append([]FolderConfiguration(nil), w.cfg.Folders...)
+	return append([]FolderConfiguration(nil), w.cfg.Copy().Folders...)
 }
 
 // SetFolder adds a new folder to the configuration, or overwrites an existing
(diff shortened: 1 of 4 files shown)


imsodin:
Do you know/have an idea where the problem occurs specifically? The diff looks a bit like a shotgun approach at first glance.

lib/config/wrapper.go
```diff
@@ -271,7 +265,7 @@ func (w *Wrapper) Folders() map[string]FolderConfiguration {
 func (w *Wrapper) FolderList() []FolderConfiguration {
 	w.mut.Lock()
 	defer w.mut.Unlock()
-	return append([]FolderConfiguration(nil), w.cfg.Folders...)
+	return append([]FolderConfiguration(nil), w.cfg.Copy().Folders...)
```
imsodin: Appending to a nil slice already ensures we return a copy.

lib/config/wrapper.go
```diff
@@ -281,17 +275,15 @@ func (w *Wrapper) SetFolder(fld FolderConfiguration) (Waiter, error) {
...
 
 	newCfg := w.cfg.Copy()
-	replaced := false
+
 	for i := range newCfg.Folders {
 		if newCfg.Folders[i].ID == fld.ID {
-			newCfg.Folders[i] = fld
-			replaced = true
-			break
+			newCfg.Folders[i] = fld.Copy()
```
imsodin: Same about copying arguments.

lib/config/wrapper.go
```diff
@@ -300,31 +292,31 @@ func (w *Wrapper) SetFolder(fld FolderConfiguration) (Waiter, error) {
...
+	return w.cfg.Options.Copy()
 }
 
 // SetOptions replaces the current options configuration object.
 func (w *Wrapper) SetOptions(opts OptionsConfiguration) (Waiter, error) {
 	w.mut.Lock()
 	defer w.mut.Unlock()
 	newCfg := w.cfg.Copy()
-	newCfg.Options = opts
+	newCfg.Options = opts.Copy()
```
imsodin: Same about copying arguments.

lib/config/wrapper.go
```diff
@@ -300,31 +292,31 @@ func (w *Wrapper) SetFolder(fld FolderConfiguration) (Waiter, error) {
...
+	return w.cfg.GUI.Copy()
 }
 
 // SetGUI replaces the current GUI configuration object.
 func (w *Wrapper) SetGUI(gui GUIConfiguration) (Waiter, error) {
 	w.mut.Lock()
 	defer w.mut.Unlock()
 	newCfg := w.cfg.Copy()
-	newCfg.GUI = gui
+	newCfg.GUI = gui.Copy()
```
imsodin: Same about copying arguments.

lib/config/wrapper.go
```diff
@@ -372,14 +364,17 @@ func (w *Wrapper) Folder(id string) (FolderConfiguration, bool) {
...
+			return folder.Copy(), true
 		}
 	}
 	return FolderConfiguration{}, false
 }
 
 // Save writes the configuration to disk, and generates a ConfigSaved event.
 func (w *Wrapper) Save() error {
+	w.mut.Lock()
+	defer w.mut.Unlock()
```
imsodin: Wait, shouldn't this already have been racy before?

lib/config/wrapper.go
```diff
@@ -154,7 +154,7 @@ func (w *Wrapper) replaceLocked(to Configuration) (Waiter, error) {
 
 	for _, sub := range w.subs {
 		l.Debugln(sub, "verifying configuration")
-		if err := sub.VerifyConfiguration(from, to); err != nil {
+		if err := sub.VerifyConfiguration(from.Copy(), to.Copy()); err != nil {
```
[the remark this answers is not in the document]
AudriusButkevicius: "should". Better safe than sorry.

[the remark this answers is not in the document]
AudriusButkevicius: I suspect this actually causes the bug, as somewhere the verification builds some alternative representation of the folder slice to check some existance and screws something up.

lib/config/wrapper.go
AudriusButkevicius: It does not ensure the slices within the folder structs are copied correctly.

lib/config/wrapper.go
AudriusButkevicius: It should have, but just fixing what I found.

lib/config/wrapper.go
```diff
@@ -213,13 +212,13 @@ func (w *Wrapper) SetDevices(devs []DeviceConfiguration) (Waiter, error) {
 		replaced = false
 		for newIndex := range newCfg.Devices {
 			if newCfg.Devices[newIndex].DeviceID == devs[oldIndex].DeviceID {
-				newCfg.Devices[newIndex] = devs[oldIndex]
+				newCfg.Devices[newIndex] = devs[oldIndex].Copy()
```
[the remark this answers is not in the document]
AudriusButkevicius: I kind of agree, but the fact there is a bug somewhere here implies something's fucky, so I'd rather be overly cautious and make sure we do the right thing, in case the caller somehow manages to fiddle with this.

lib/config/wrapper.go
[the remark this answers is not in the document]
imsodin: There is just a single implementation of `Committer.VerifyConfiguration` that does not just return nil - the `apiService`, which calls `to.GUI.Address()`. Modifying the config in a verification function would simply be wrong, no gray area there. However there's not much to argue against "better safe than sorry",

lib/config/wrapper.go
imsodin: Oh right, good one. My money is on this as error source :)

lib/config/wrapper.go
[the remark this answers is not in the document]
imsodin: Ok, I think we needn't argue much, as we are actually on the same page - your just looking at it from the perspective of squashing a nasty bug, while I am in the easy situation of looking at it "academically" for review. I don't think this is a bottleneck even in setups that reach to the boundaries of scalability of Sycnthing.

lib/config/wrapper.go
imsodin: Not complaining, just wondering.

lib/config/wrapper.go
```diff
@@ -281,17 +275,15 @@ func (w *Wrapper) SetFolder(fld FolderConfiguration) (Waiter, error) {
...
 		if newCfg.Folders[i].ID == fld.ID {
-			newCfg.Folders[i] = fld
-			replaced = true
-			break
+			newCfg.Folders[i] = fld.Copy()
+			return w.replaceLocked(newCfg)
 		}
 	}
-	if !replaced {
-		newCfg.Folders = append(w.cfg.Folders, fld)
```
AudriusButkevicius: So after a bunch of bisecting, this seems to be an issue, essentially appending new folders to a copy from the old map, yet I don't understand why this could be an issue, as all of this happens under a lock.

calmh: I'm just shooting from the hip here, but could there be an issue when two folders are added in rapid successions, such as from the same cluster config. That is, if the folder list is initially A, B and we append to A, B, C but this doesn't get propagated and the next folder added gets appended to A, B as A, B, D - The "C" folder is now mysteriously gone and bad things happen...

calmh:
Yeah I'm not super happy about adding a bunch of copying everywhere because we're not sure if or when the config is changed. That seems like a problem we should solve at the root...

AudriusButkevicius:
So the only thing is perhaps sensible to revert is copying inputs.
Other than that I think other changes make sense.

AudriusButkevicius:
Can we quantify what we want unwinded?

imsodin:
Personally, I'd temporarily undo everything except the "fixes" and improvements and copying when returning config elements, then ask @pitlochry again for a test. But obviously I don't know what @calmh has in mind ;)

AudriusButkevicius:
So I know that there is only a single line that fixes the issue, yet I think copying in most cases is the sensible thing todo, because FolderMap() should never return references to the orignal folders, nor should we risk folks modifying things in VerifyConfiguration, nor should we rely on folks not modifying slices after they did SetFolder etc. With copies the behaviour is more consistent.

imsodin:
> So I know that there is only a single line that fixes the issue

I missed that one, would you mind pointing it out explicitly (again?) to me?

AudriusButkevicius:
See last comments.

lib/config/wrapper.go
imsodin: I don't have a concrete scenario, Both auto accepting happening in `ClusterConfig` and `CommitConfiguration` happen without lock on `fmut`. And `CommitConfiguration` is called in a goroutine, i.e. the wrapper lock won't be held while the model is running it. So in principle it might be possible for `CommitConfiguration` removing/adding folders at the same time as `ClusterConfig` does.

calmh:
I don't understand how the proposed diff could possibly fix the problem. In model.CommitConfiguration it's looking at "from" and "to" and drawing the conclusion that "from" lacks a folder that's clearly present from before, at device startup even. This means that the Folders slice in "from" is broken, somehow. I don't see anywhere we modify a Folders slice without copying it first.

The one thing that worries me a little is notifyListeners:

https://github.com/syncthing/syncthing/blob/b83c5b32bfbe8662b174d48604786a56f7401f79/lib/config/wrapper.go#L170-L180

The copying of "from" and "to" happens outside of the lock. Yet, I don't see how "from" or the Folders slice inside it could ever be modified, even if we return from replaceLocked and those copies happen later...

AudriusButkevicius:
So I did reproduce it. I'll try running the tests with race.

calmh:
So I see now that you already found and fixed that, but I didn't see that part of the diff and thought you were talking about the `fld.Copy()` that you also added. So I agree - that line is the problem, but no need to copy the folderconfig once more.

AudriusButkevicius:
So I think other changes don't do damage? We should also probably copy from and to prior to passing things to the new routines.

calmh:
I think the other changes are voodoo and muddy the waters. The point of the copies is not to prevent modifications by the caller, but to give them something that doesn't change under their feet if there's a second config commit while they are looking at the first.

I agree we should copy before firing the routines though.

imsodin:
> [...] The point of the copies is not to prevent modifications by the caller, but to give them something that doesn't change under their feet  [...]

Wouldn't that include quite a few of Audrius added copies, like https://github.com/syncthing/syncthing/pull/5069/files#diff-5d899bb376d9455c8659d63b8b210d53R191 ?

calmh:
I started marking up all the changes with yes/no IMHO, but it boils down to "yes, copy when we're returning a slice or map of something that the caller can reasonably expect to be the owner of" and no in the other cases.

AudriusButkevicius:
Anything else left here?

lib/config/wrapper.go
```diff
@@ -142,7 +142,7 @@ func (w *Wrapper) RawCopy() Configuration {
 func (w *Wrapper) Replace(cfg Configuration) (Waiter, error) {
 	w.mut.Lock()
 	defer w.mut.Unlock()
-	return w.replaceLocked(cfg)
+	return w.replaceLocked(cfg.Copy())
```
imsodin: I think this copy and the one in `replaceLocked` are redundant.

AudriusButkevicius: Not really, as `to` is not copied when stored on `w.cfg`, so the called still can fiddle with original slices passed in `wrapper.Replace(cfg)`

calmh:
Practically this is all fine, philosophically I think it's dirty to add copies everywhere to protect against some theoretical fiddling coder boogeyman. It's not like we usually worry about this whenever we pass a slice of something to somewhere. But it solves the bug (which is not due to too few calls to Copy()...) and I'll leave the potential cleanup as a battle for another day.

PR #5072, opened 2018-07-19
[docker] Create a dedicated syncthing user

bebehei:
### Purpose

A dedicated user is necessary to create relative references via `~/<folder>` or `$HOME/<folder>`. Having the syncthing process just running under a unprivileged UID/GID, will remove the home folder relation and therefore will result in nonexistent shares after update.

Fixes the changes introduced in #5041

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,11 +21,30 @@ COPY --from=builder /go/src/github.com/syncthing/syncthing/syncthing /bin/syncth
 RUN apk add --no-cache su-exec
 
 ENV STNOUPGRADE=1
+ENV PUSR=syncthing
 ENV PUID=1000
+ENV PGRP=syncthing
 ENV PGID=1000
 
 HEALTHCHECK --interval=1m --timeout=10s \
   CMD nc -z localhost 8384 || exit 1
 
-ENTRYPOINT chown $PUID:$PGID /var/syncthing \
-    && su-exec $PUID:$PGID /bin/syncthing -home /var/syncthing/config -gui-address 0.0.0.0:8384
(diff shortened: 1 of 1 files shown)


calmh:
I think this is fine, although I don’t understand the `&& true` at the end? However I think a simpler approach would be to set the HOME end var explicitly, unless suexec will stomp on it.

bebehei:
Setting the `HOME` variable could work. I haven't tested it yet. But I actually wouldn't go for it. IIRC, the `HOME` variable is considered on some shells to be read only.

While reading up the [POSIX docs](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_03), I wouldn't consider this option anymore:

> HOME:
> 
> The system shall initialize this variable at the time of login to be a pathname of the user's home directory.

Combining the current approach with a manually set `HOME` variable would work, but in the specs, there is a user necessary.

---

The `&& true` is actually just a habit of mine. It's a void command, which allows you to append commands later, without polluting the diff of the previous line (adding a ` \` at the end of the line).

So, when you append a command as the last command, you can automatically see it in the git log, that it's only a single command added:


```patch
 RUN true \
  && apk add --no-cache ca-certificates \
+ && pip install otherpackage \
  && true
```

Instead of:

```patch
-RUN apk add --no-cache ca-certificates
+RUN apk add --no-cache ca-certificates \
+ && pip install otherpackage
```

I agree, this is rather esoteric, but it aids your eye to see, where the changes are.

Shall I remove it?

calmh:
No, it's fine as is. Thanks for the explanations.

PR #5081, opened 2018-07-26
github: Add sections to issue template

ProactiveServices:
Add a section on using the correct issue tracker
Add a section on database corruption


ProactiveServices:
If it's agreed that adding these sections is helpful and/or will be read by anyone using the tracker...

calmh:
Sadly, I suspect the "offenders" will not read this part. Nonetheless it's helpful, thanks.

PR #5085, opened 2018-07-29
gui: Use one instead of on to have callbacks fire one

AudriusButkevicius:
Every time we open a dialog we register a new callback, but never deregister the old one. 
This uses one time callbacks in places where I think it makes sense (such as paths where the callback will be re-registered on modal being shown).


calmh:
"... once", and I saw it literally while the finger clicked the merge button 🤦‍♂️

PR #5093, opened 2018-08-01
build: Build and set GOPATH before generating assets

imsodin:
From the v0.14.49 announcement thread: https://forum.syncthing.net/t/syncthing-v0-14-49/11875/6

Since `go` took over generating the assets in https://github.com/syncthing/syncthing/commit/ef5ca0c218cfd072860d2b06d63c3d36ddc63813 the mechanism to build and set a temporary `GOPATH` is broken, as assets are generated before `GOPATH` is set. This PR inverts the order of that.

### DIFF
--- a/build.go
+++ b/build.go
@@ -211,14 +211,14 @@ func main() {
 		if err != nil {
 			log.Fatal(err)
 		}
+		os.Setenv("GOPATH", gopath)
+		log.Println("GOPATH is", gopath)
 		if !noBuildGopath {
-			lazyRebuildAssets()
 			if err := buildGOPATH(gopath); err != nil {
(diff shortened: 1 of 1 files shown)


calmh:
Did you check that it works?

imsodin:
I did the change on the 0.14.49 tar that I extracted outside of my GOPATH without any GOPATH set, and it worked.

wwwutz:
Yep, fixes the issue. Thanks.

wwwutz:
oops, I should have tested better!

Now I get a `404 page not found` so it builds, but the assets are missing.

any hints how to debug ?

Issue #5092, opened 2018-08-01
File watcher misses deletions if file is moved across shared folders


Catfriend1:
Reproducer
I created the file mentioned on the screenshot below under path "D:/Dokumentation/Zutrittskontrolle" (share level is "D:/Dokumentation").
- File showed correctly as added during ~ 10 seconds.
I moved the file to "D:/Installationsdateien" (top level of another share). Both shared folders reside on the same disk.
- It showed correctly as added during ~ 10 seconds.
I expected the file to show up as "deleted" in "D:/Dokumentation" shared folder and this did not happen.
I waited for 3 minutes to make sure the file watcher has enough time but nothing happened.

Rescan interval of both shares is configures 3600 seconds. I hit the "Rescan now" button and the deletion notice appeared almost immediately. (compare to screenshot's top line)

Went to the syncthing menu and looked through the log, no clue where this bug came from as during testing with the reproducer no new log line came up. Restarted syncthing and again did the reproducer. Problem persists.

Screenshot of recent changes popup dialog
![image](https://user-images.githubusercontent.com/16361913/43511274-480600b4-9578-11e8-9a66-2e4b381c3c62.png)

### Version Information
Syncthing Version: v0.14.49
OS Version: Windows 10 x64
Browser Version: Chrome 63.0.3239 x64

Catfriend1:
I again tried to do the reproducer, but now I only get "modified" instead of "file added". I deleted the "test file" beforehand but somehow it didn't get totally removed from the index? This time, I had more log facilities on:

![image](https://user-images.githubusercontent.com/16361913/43511606-186b284c-9579-11e8-9fd8-098b2c5c4544.png)


> 2018-08-01 10:50:43 aggregator/"Dokumentation" (cdaf2-v5ayb): No old fs events
2018-08-01 10:50:43 aggregator/"Dokumentation" (cdaf2-v5ayb): Resetting notifyTimer to 10s
2018-08-01 10:50:51 aggregator/"Installationsquellen" (n4q4z-ohpfz): No tracked events, waiting for new event.
2018-08-01 10:50:53 aggregator/"Dokumentation" (cdaf2-v5ayb): No old fs events
2018-08-01 10:50:53 aggregator/"Dokumentation" (cdaf2-v5ayb): Resetting notifyTimer to 10s
2018-08-01 10:51:03 aggregator/"Dokumentation" (cdaf2-v5ayb): No old fs events
[...]

PR #5094, opened 2018-08-02
lib/watchaggregator: Don't delay mixed events only

imsodin:
This is a follow-up to #4955, where the notification of deletions was made faster.  In the logs from #5092 (unrelated issue) I realized a small mistake in the implementation: Currently only "pure" deletions are returned early, if there is no other types of events. There's no point in delaying mixed events either, if no non-remove events exist.

I wrote a small test to check that there is no delay in this scenario. And I also fixed a minor bug in the output in case of a timing test failure.

### DIFF
--- a/lib/watchaggregator/aggregator.go
+++ b/lib/watchaggregator/aggregator.go
@@ -308,8 +308,8 @@ func (a *aggregator) actOnTimer(out chan<- []string) {
 	}
 	oldEvents := make(map[string]*aggregatedEvent, c)
 	a.popOldEventsTo(oldEvents, a.root, ".", time.Now(), true)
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
What is mixed in that case?

Catfriend1:
Wow @imsodin, as I wrote on the issue it works correctly for me, but if you can make it even better ... :+1:

imsodin:
@AudriusButkevicius Mixed means both remove and non-remove have been recorded at the same path. In the logs I mentioned this happened because we receive a remove event for `parent/removed` and at the same time a modification event for `parent`, which is a non-remove event.

imsodin:
lgtm?

AudriusButkevicius:
This is all foreign to me, so looks ok :)

PR #5097, opened 2018-08-04
lib/model: Check availability later to catch renames

imsodin:
### Purpose

I was thinking about the keep pulling and deleting stuff when disk is full and noticed that currently the availabilty check happens too early: If a file is unavailable, it's not pushed to the queue meaning deletes can't be matched against it -> potential renames (don't need availability) won't be detected.

### Testing

None, passes existing tests. This is a small optimization not a bug and I am lazy right now -> no additional unit test.


imsodin:
@calmh A heads up due to the merge: In the other PR you voiced potential reservations against this.

PR #5099, opened 2018-08-06
lib/model: Don't stop folder if out of disk space (fixes #2370)

### LINKED ISSUE #2370: Scanning and deletes should be processed even when above the free space limit
Not sure if it is a bug or not, but it's better to write it down:
- I am running into the free space limit and I get these messages: `[TYZZZ] INFO: Skipping folder dokumente scan due to folder error: insufficient free space`
- Due to the limit currently my sync stucks at 59 % (at that box where the limit has been hit)
- I have deleted the big file, that is being pulled on another node
- Syncing stays at 59 %

---

After restarting syncthing it recognizes that there is not more file to pull; that's fine. But because of the free space limit it keeps stopping that particular folder, so it does not delete the file, hence it keeps stuck.

imsodin:
### Purpose

This removes the out of disk space check from CheckHealth. The disk space is now
only checked if there are files to pull, in which case pulling those files is
stopped, but everything else (dirs, links, deletes) keeps running -> can recover
disk space through pulling.

### Testing

No new unit test - not sure how to go about creating one.


imsodin:
I had a look at the history and the only time this passage was modified was when it was introduced in the delta index PR:
https://github.com/syncthing/syncthing/pull/3427/files#diff-feb9fde15f8dd977fe9b279a36a447a2

There is no reference to that change in the PR comments and the code comment makes it sound like this is "just an optimization".

lib/model/folder_sendrecv.go
```diff
@@ -413,19 +403,23 @@ func (f *sendReceiveFolder) processNeeded(ignores *ignore.Matcher, dbUpdateChan
...
-		f.queue.SortSmallestFirst()
-	case config.OrderLargestFirst:
-		f.queue.SortLargestFirst()
-	case config.OrderOldestFirst:
-		f.queue.SortOldestFirst()
-	case config.OrderNewestFirst:
-		f.queue.SortNewestFirst()
+	// No need to order if we aren't going to pull anything.
+	spaceErr := f.CheckFreeSpace()
+	if spaceErr != nil {
```
AudriusButkevicius: `==` ?

calmh:
So you're right that that's fine I think. However I think it's excessive to check for free space for every block we pull and every block list we copy. Note that we have an extra check in handleFile:

https://github.com/syncthing/syncthing/blob/165417c46235b0b1787599980532dd8d23e01390/lib/model/folder_sendrecv.go#L1006-L1012

There we check specifically that there is enough space to handle the file, regardless of what the free space check has said before. I think this would be a good place to move the normal free space check. 

Essentially, if we are supposed to do a free space check, do so here. If we do and that passes, also check that the file actually fits (like it does now).

imsodin:
Wasn't aware of that check, that definitely obsoletes the pull/copy ones. However I think this check as it is now does not respect the disk free requirement, it just looks at the absolute free disk space.

Looking at this PR again the current implementation is definitely flawed: It's pretty useless to check once at the beginning of the loop and then reuse the result of that check. I think the right thing to do are checks at the specific point where disk space is required, like the one mentioned in `handleFile`.

imsodin:
@calmh @AudriusButkevicius Please review again, I changed it significantly :  
It still removes the disk space check from the folder health check, but instead of checking free disk space once, the check now happens in places where new disk space is actually needed, i.e. before pulling and when copying to rename. I replaced the existing free disk check before pulling with `FolderConfiguration.CheckAvailableSpace`, because it did not take into account `FolderConfiguration.MinDiskFree`.

Issue #5043, opened 2018-07-01
root on symlinked path causes panic when using "Watch for changes"


larroy:
I think the logic here has problems when the path is symlinked:

https://github.com/syncthing/syncthing/blob/master/lib/fs/basicfs_watch.go#L113

I have a "storage" folder symlinked to mnt:  /home/piotr/storage -> /mnt/storage/storage/

This causes panic when using the notify backend

panic: bug: Notify backend is processing a change outside of the filesystem root: root==/home/piotr/storage/synced/oneplus5/, rootSymEval==/home/piotr/storage/synced/oneplus5/, path==/mnt/storage/storage/synced/oneplus5/.stfolder

### Version Information

Syncthing Version: v0.14.48
OS Version: Linux 64  Ubuntu 16.04 (installed from apt)
Browser Version: (if applicable, for GUI issues)

imsodin:
Can you run the following script on your system and report the results, i.e. install Go and run `go run test.go`

```golang
package main

import (
	"fmt"
	"path/filepath"
)

func main() {
	fmt.Println(filepath.EvalSymlinks("/home/piotr/storage/synced/oneplus5/"))
}
```

In any case we need to act on, not just ignore an error when determining the root with symlinks evaluated: https://github.com/syncthing/syncthing/blob/master/lib/fs/basicfs.go#L56  
Either panic already there (probably a bad idea, though I don't get why it should ever fail), or disable watching to prevent panics.

larroy:
```
/mnt/storage/storage/synced/oneplus5 <nil>
```

larroy:
Why the assertion in the first place? what are we trying to accomplish in unrootedChecked ?

imsodin:
@larroy 

Why asserting/panicing is a good question - it makes life easy for me as I don't need to think about how to handle a case that should never happen. A panic is ideal for that, as one line of code guarantees that the issue will come to my attention if it ever arises. Obviously, it's not quite user-friendly. All of that is tangential to your actual case though.

What needs figuring out is where the evaluation of symlinks in your filesystem instance failed, i.e. why there is this conflict:

panic: bug: Notify backend is processing a change outside of the filesystem root: root==/home/piotr/storage/synced/oneplus5/, **rootSymEval==/home/piotr/storage/synced/oneplus5/**, path==/mnt/storage/storage/synced/oneplus5/.stfolder

fmt.Println(filepath.EvalSymlinks("/home/piotr/storage/synced/oneplus5/"))
[...]

PR #5104, opened 2018-08-07
lib/model, lib/protocol: Fix error handling when terminating a connection

imsodin:
### Purpose

In the protocol's readerloop do actually set the err variable instead of
returning directly, such that it is correctly passed to c.close in the deferred
statement.
In the model do a nil-check before calling err.Error().

### Testing

Simple unit test that doesn't panic anymore.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1270,10 +1270,15 @@ func (m *Model) Closed(conn protocol.Connection, err error) {
 	m.pmut.Unlock()
 
 	l.Infof("Connection to %s at %s closed: %v", device, conn.Name(), err)
-	events.Default.Log(events.DeviceDisconnected, map[string]string{
-		"id":    device.String(),
-		"error": err.Error(),
-	})
+
+	data := map[string]string{
+		"id": device.String(),
+	}
+	if err != nil {
+		data["error"] = err.Error()
+	}
+	events.Default.Log(events.DeviceDisconnected, data)
+
 	close(closed)
 }
 
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -340,11 +340,12 @@ func (c *rawConnection) readerLoop() (err error) {
 	for {
 		select {
 		case <-c.closed:
-			err = errClosed
+			err = ErrClosed
 			return
 		default:
 		}
 
+		var msg message
 		msg, err = c.readMessage()
 		if err == errUnknownMessage {
 			// Unknown message types are skipped, for future extensibility.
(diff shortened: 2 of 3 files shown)


lib/protocol/protocol.go
```diff
@@ -340,11 +340,12 @@ func (c *rawConnection) readerLoop() (err error) {
 	for {
 		select {
 		case <-c.closed:
-			return ErrClosed
```
calmh: As far as I understand, these changes should be an ugly no-op?

lib/model/model.go
```diff
@@ -1270,10 +1270,15 @@ func (m *Model) Closed(conn protocol.Connection, err error) {
 	m.pmut.Unlock()
 
 	l.Infof("Connection to %s at %s closed: %v", device, conn.Name(), err)
-	events.Default.Log(events.DeviceDisconnected, map[string]string{
-		"id":    device.String(),
-		"error": err.Error(),
-	})
+
+	data := map[string]string{
```
calmh: It should be safe to assume that we never close with a nil error - if the error was nil, the connection would not have been closed. I'd even throw in an `if err == nil { panic("bug:...") }` to that effect. Clearly you've *seen things* I haven't, though?

imsodin:
@calmh Yes, I have seen ghosts aka this nil deref https://github.com/syncthing/syncthing/issues/5043#issuecomment-411097106 which points to https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L1275. The return value `err` is declared in the function signature and used it in the defer statement, so I assume if we return directly that error doesn't get assigned to `err` and thus the defer statement uses a nil error. On second thoughts, that can't be because then every connection close should have resulted in a nil deref.

So actually I am at a loss as to how the referenced nil deref every happened...

lib/model/model.go
AudriusButkevicius: Agree. The bug is clearly in the protocol package.

imsodin:
I now believe my changes in this PR are BS from A to Z. In the original report `model.Closed` was called with a nil error, but I don't understand at all how. It also looks like that was a double panic, something really strange is going on.
```
github.com/syncthing/syncthing/lib/model.(*Model).Closed(0xc420211320, 0x7f3481835e60, 0xc4224d3710, 0x0, 0x0)
```

Also some further weirdness, that probably explains the discrepancy between my example above (direct defer) vs the code in protocol.go (defer with func):

```golang
package main

import (
	"fmt"
)

func deferStuff() (err error) {
	defer func() {
		fmt.Println("defer func", err)
	}()
	defer fmt.Println("defer direct", err)
	return fmt.Errorf("this is an error")
}

func main() {
	fmt.Println(deferStuff())
}
```

```
defer direct <nil>
defer this is an error
this is an error
```

calmh:
If we panicked in the protocol code while `err` was still `nil`, the defer would be called with `c.close(nil)`. I think this is an unexpected side effect of the panic, but probably doesn't matter much as we're already going down due to the panic... (which should be fixed, of course)

PR #5105, opened 2018-08-08
lib/fs: Evaluate root when watching not on fs creation (fixes #5043)

### LINKED ISSUE #5043: root on symlinked path causes panic when using "Watch for changes"

imsodin:
### Purpose

Currently the folder root is evaluated (to resolve symlinks) at fs creation. At that point the directory may not exist or `EvalSymlinks` may fail for other reasons. The current implementation just ignores that case by assuming there are no symlinks. That's dangerous because if there are, it will lead to panics on watching. Also the only place where this evaluated root is needed is for watching. Therefore the evaluation of the root path now happens when `fs.Watch` is called. This call will then just fail if `filepath.EvalSymlinks` returns an error, making it transparent to the user, leading to retries and preventing the panic.

### Testing

Existing ones still pass.

### DIFF
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -25,8 +25,7 @@ var (
 // The BasicFilesystem implements all aspects by delegating to package os.
 // All paths are relative to the root and cannot (should not) escape the root directory.
 type BasicFilesystem struct {
-	root                 string
-	rootSymlinkEvaluated string
+	root string
 }
 
 func newBasicFilesystem(root string) *BasicFilesystem {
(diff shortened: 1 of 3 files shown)


lib/fs/basicfs.go
```diff
@@ -53,34 +52,17 @@ func newBasicFilesystem(root string) *BasicFilesystem {
...
-			root = `\\?\` + root
-		}
-		return root
-	}
-
-	// If we're not on Windows, we want the path to end with a slash to
-	// penetrate symlinks. On Windows, paths must not end with a slash.
-	if root[len(root)-1] != filepath.Separator {
+		root = longFilenameSupport(root)
+	} else if root[len(root)-1] != filepath.Separator {
```
calmh: I'd do a real strings.HasSuffix here instead, avoids crashing on the empty string etc

imsodin: Done. I also adapted the windows fs creation test for unixes to check this and other basic scenarios. Apparently the tilde and absolutification meddling above already changed "" to "/", so it wouldn't panic even before.

imsodin:
It does not pass reliable on master, it just seems to be happening a whole lot right now. See e.g. here: https://build.syncthing.net/viewLog.html?buildId=24835&tab=buildResultsDiv&buildTypeId=Syncthing_BuildWindows#testNameId8662697608653222476

I initially didn't want to fix #5110 immediately (that's why I created the issue as a reminder), but these recurrent failures changed my mind - PR incoming. Hopefully it will actually fix things...

Issue #5078, opened 2018-07-25
panic: nil pointer dereference 


scienmind:
Syncthing performed an autoupgrade from .48 to .49, then run for about 2 hours.
At some point (to speed up important transfers) I paused all folders, and then unpaused the folder of interest.
After a while I clicked "Resume All", to get ST back to normal state.

Few seconds later, ST crashed.

Log: https://privatebin.net/?3b2250f7aede7ed9#BnRnau5xAI3kkf8otln/ojGyyaG9jRnTILidgO1HykU=  

### Version Information

Syncthing Version: v0.14.49
OS Version: Windows 10 x64
Run by SyncTrayzor v.1.1.21

calmh:
I think this is a race condition with a device connecting at the same time we're in the middle unpausing a folder - which is likely to happen as we reconnect devices as part of the unpausing.

PR #5106, opened 2018-08-08
lib/model: Catch racy nil deref in ClusterConfig

imsodin:
### Purpose

Next try at fixing the following panic (from https://github.com/syncthing/syncthing/issues/5043#issuecomment-411097106):

```
Panic at 2018-08-07T10:24:04-05:00
panic: runtime error: invalid memory address or nil pointer dereference
    panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x91f1cc]

goroutine 577 [running]:
github.com/syncthing/syncthing/lib/model.(*Model).Closed(0xc420211320, 0x7f3481835e60, 0xc4224d3710, 0x0, 0x0)
    /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/model/model.go:1262 +0x50c
github.com/syncthing/syncthing/lib/protocol.(*rawConnection).close.func1()
    /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/protocol/protocol.go:833 +0x206
sync.(*Once).Do(0xc420730148, 0xc420efb120)
    /media/ext4_data/Linux/source/go/src/sync/once.go:44 +0xbe
github.com/syncthing/syncthing/lib/protocol.(*rawConnection).close(0xc4207300c0, 0x0, 0x0)
    /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/protocol/protocol.go:820 +0x6b
github.com/syncthing/syncthing/lib/protocol.(*rawConnection).readerLoop.func1(0xc4207300c0, 0xc420efbfd0)
    /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/protocol/protocol.go:335 +0x41
panic(0xa75380, 0x1117ca0)
    /media/ext4_data/Linux/source/go/src/runtime/panic.go:502 +0x229
github.com/syncthing/syncthing/lib/db.(*FileSet).IndexID(0x0, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xc422e5c501)
    /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/db/set.go:263 +0x5e
github.com/syncthing/syncthing/lib/model.(*Model).ClusterConfig(0xc420211320, 0x4b91a04c26d701c3, 0xeb45ab20275c2e1e, 0x81bc97fd845af037, 0x4df64d3c0e0a8595, 0xc420ea8500, 0x4, 0x4)
    /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/model/model.go:949 +0xda6
github.com/syncthing/syncthing/lib/protocol.(*rawConnection).readerLoop(0xc4207300c0, 0x0, 0x0)
    /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/protocol/protocol.go:361 +0xed0
created by github.com/syncthing/syncthing/lib/protocol.(*rawConnection).Start
    /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/protocol/protocol.go:225 +0x43
```

Line numbers refer to a PR - the following
```
github.com/syncthing/syncthing/lib/model.(*Model).ClusterConfig(0xc420211320, 0x4b91a04c26d701c3, 0xeb45ab20275c2e1e, 0x81bc97fd845af037, 0x4df64d3c0e0a8595, 0xc420ea8500, 0x4, 0x4)
    /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/model/model.go:949 +0xda6
```
points to https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L962.

My theory: `ClusterConfig` happens before `CommitConfiguration` finishes -> config is present, but the respective entry in `m.folderFiles` is still missing. Even if that theory is wrong in this case(I would expect the call to `fs.IndexID` to panic, but here `s.db.getIndexID` within `fs.IndexID` does) it could happen and it doesn't hurt to be safe. Continuing is fine, because when the folder is started, the connection is dropped -> new cluster config.

### Testing

None as this is racy.

EDIT: Actually I am really not on top of everything at the moment: The panic above is the same as in #5078.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -953,12 +953,17 @@ func (m *Model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
 		if cfg.Paused {
 			continue
 		}
+		fs, ok := m.folderFiles[folder.ID]
+		if !ok {
+			// Shouldn't happen because !cfg.Paused, but might happen
+			// if the folder is about to be unpaused, but not yet.
+			continue
+		}
 
 		if !folder.DisableTempIndexes {
 			tempIndexFolders = append(tempIndexFolders, folder.ID)
 		}
 
-		fs := m.folderFiles[folder.ID]
 		myIndexID := fs.IndexID(protocol.LocalDeviceID)
 		mySequence := fs.Sequence(protocol.LocalDeviceID)
 		var startSequence int64


imsodin:
@calmh What's your stance on whether this fixes #5078? I am confused by the trace panicking within the `FileSet` instead of when calling its method. On the other hand the second last call points exactly at this potential race/nil deref, so I am hopeful it is the same thing. Maybe close and hope it is fixed and reopen in case it gets reported again?

imsodin:
Ah right, I keep forgetting that method calls on typed nil pointers are allowed in Go... Thanks for the reminder :)

calmh:
It helps to keep in mind that the first parameter to functions in the backtrace is the receiver, if it's a method. So we can see the `db.(*FileSet).IndexID(0x0, ...` call and that the receiver is nil. And then consider what happens in the method.

PR #5111, opened 2018-08-11
lib/model: Run receive-only tests in temporary dirs (ref #5105)

imsodin:
Whether or not it actually helps in any way is unsure, but I believe it is anyway good practice so why not try it out.


calmh:
This is probably fine (although I wouldn't involve `fs`) but it sweeps the root cause under the rug and I want to understand that first.

imsodin:
Using fs is just convenient, otherwise you'd have to litter `filepath.Join(dir, "foo/bar")` all over the place.

As to understanding the root cause first: The unrelated test errors on my PRs are just slowly getting on my nerves. Having to press run manually here and there is fine, doing it repeatedly without success is not so cool. I totally understand that you want to know the root case (well I want to know it too, but I don't want to search it :P ).

calmh:
I'm searching for it currently

PR #5112, opened 2018-08-11
lib/model: Don't run watcher on recvonly tests (fixes #5110)

### LINKED ISSUE #5110: Run folder tests in temporary directories
`TestRecvOnlyRevertNeeds` (among others) is extremely flaky, especially on the windows builder.
https://build.syncthing.net/project.html?tab=testDetails&testNameId=8662697608653222476&projectId=Syncthing&buildTypeId=Syncthing_BuildWindows

The error is
```
[TestRecvOnlyRevertNeeds] folder_recvonly_test.go:206: mkdir _recvonly: Access is denied.
```
which happens in a utility function generating files for testing in 	`_recvonly/knownDir`. Multiple tests use that dir, so the failures may be due to interference between tests. If every test used its own temporary dir, this could be ruled out as a cause.

calmh:
Alternate solution

### DIFF
--- a/lib/model/folder_recvonly_test.go
+++ b/lib/model/folder_recvonly_test.go
@@ -22,17 +22,35 @@ func TestRecvOnlyRevertDeletes(t *testing.T) {
 	// Make sure that we delete extraneous files and directories when we hit
 	// Revert.
 
-	os.RemoveAll("_recvonly")
-	defer os.RemoveAll("_recvonly")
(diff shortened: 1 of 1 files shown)


imsodin:
Maybe we should disable fs watching on all model tests. The problem also exists in many send-receive tests, just that we already do use separate test folders, so it doesn't show up as failures (but the test dirs will probably stay in place, which isn't problematic, but ugly).

PR #5114, opened 2018-08-11
lib/db: Use indirect file keys (fixes #5047)

### LINKED ISSUE #5047: Reduce database size by FileInfo indirection
Currently we store each FileInfo per the device that announces it - one for ourselves and once for each peer device. When the devices in a cluster are in sync the FileInfo will be identical for all of them. If there are many devices (think hundreds or thousands) this can become a large multiplier on the database size.

We could instead use an intermediary index and a reference counter. That is, instead of todays

```
(KeyTypeDevice, device, folder, filename) -> marshalled FileInfo
```

we could do

```
(KeyTypeDevice, device, folder, filename) -> hash of marshalled FileInfo
(KeyTypeHashedFileInfo, hash of marshalled FileInfo) -> (refcounter, marshalled FileInfo)
```

When doing inserts we hash the marshalled FileInfo and store that in place of the marshalled FileInfo. We then look up the item corresponding to that hash and either write `(1, marshalled FileInfo)` or bump the counter if it already exists.

When overwriting a FileInfo we need to do the lookup and decrease the counter for the old value, potentially deleting the actual FileInfo when the counter reaches zero.

(We only *delete* FileInfos when dropping a folder, which can be special cased.)

calmh:
### Purpose

Instead of direcly marshalling the FileInfo for every (device,folder,file) key we instead go via a layer of indireciton that is the SHA256 of the marshalled FileInfo. We don't track usage via refcounts or similar, opting instead for a periodic GC step. This GC uses a Cuckoo filter to keep track of keys that are (probably) in use, keeping the memory usage bounded even for large sets of files. A small number of entries may be left behind due to the probabilistic nature of the filter, but this is on the order of less than a tenth of a percent for the largest file sets, less for everyone else.

### FAQ

- Why not refcounting? This is less work to track and less likely to fuck up.
- Why hash of marshalled FileInfo and not the relevant subset of attributes directly? This is more provably correct and doesn't require a database migration if the relevant set of fields changes.
- Why the Cuckoo filter? Because keeping the full set of used file hashes in RAM might exhaust RAM, and doing it with an extra accounting table in the database seems slow and annoying. This feels like a reasonable tradeoff.
- Why, at all? Much smaller DB when there are lots of remote devices. The amount of code was quite small.

### Testing

Minor new test to verify that GC happens.

### DIFF
--- a/lib/db/leveldb.go
+++ b/lib/db/leveldb.go
@@ -14,19 +14,20 @@ import (
 )
 
 const (
-	KeyTypeDevice = iota
-	KeyTypeGlobal
-	KeyTypeBlock
-	KeyTypeDeviceStatistic
-	KeyTypeFolderStatistic
-	KeyTypeVirtualMtime
-	KeyTypeFolderIdx
-	KeyTypeDeviceIdx
-	KeyTypeIndexID
-	KeyTypeFolderMeta
-	KeyTypeMiscData
-	KeyTypeSequence
-	KeyTypeNeed
+	KeyTypeDevice          = 0
+	KeyTypeGlobal          = 1
+	KeyTypeBlock           = 2
+	KeyTypeDeviceStatistic = 3
+	KeyTypeFolderStatistic = 4
+	KeyTypeVirtualMtime    = 5
+	KeyTypeFolderIdx       = 6
+	KeyTypeDeviceIdx       = 7
+	KeyTypeIndexID         = 8
+	KeyTypeFolderMeta      = 9
+	KeyTypeMiscData        = 10
+	KeyTypeSequence        = 11
+	KeyTypeNeed            = 12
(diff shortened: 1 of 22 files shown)


calmh:
Actually no, this is conceptually flawed. bbl, maybe.

imsodin:
Any hints as to the flaw in the concept? I am frequently entertaining this idea as well (though based on version not hash as key), thus I am curious :)

calmh:
Every file info is unique because the sequence is set by each device, so there is no reuse with this scheme. We would need to handle sequence separately outside of the fileinfo, or do a light version and just indirect the block list, both ideas are more complicated than what I implemented.

AudriusButkevicius:
Sounds like a shit show.

PR #5119, opened 2018-08-13
all: Don't let Suture capture panics (fixes #4758)

### LINKED ISSUE #4758: Suture services should not survive panics

calmh:
Fork with new option. Default behavior stays unchanged, because I'm lazy and don't want to rewrite all the tests in the forked package.


AudriusButkevicius:
I think this should go back upstream?

calmh:
It's counter to the philosophy of capturing all the things, so it's not welcome upstream.  I have an old issue on it there.

AudriusButkevicius:
It didn't look like it was a hard no on the old issue, more like how about this, you saying sounds good and closing it.

calmh:
Maybe I misremember. I’ll file a PR.

PR #5120, opened 2018-08-13
Add heartbeats to stdiscosrv

### LINKED ISSUE #5117: cmd/stdiscosrv: Not enough traffic breaks replication between discovery servers
### Environment:

Two discovery servers, one running v0.14.46@armhf/Raspbian, the other running on v0.14.49@x86_64/Centos7. I occasionally see in the logs:
```
stdiscosrv[11974]: Replication read size: read tcp 172.31.3.1:19200->192.168.4.1:50316: i/o timeout
```
or
```
stdiscosrv[12505]: Replication write: write tcp 192.168.4.2:40726->192.168.8.1:19200: write:broken pipe
stdiscosrv[12505]: Supervisor main: main: Failed service 'replicationSender("192.168.8.1:19200")' (1.000000 failures of 5.000000), restarting: true, error: "{replicationSender(\"192.168.8.1:19200\") replicationSender(\"192.168.8.1:19200\")} returned unexpectedly", stacktrace: [unknown stack trace]
```
After closer investigation, it seem that the culprit is this line: https://github.com/syncthing/syncthing/blob/2c9d96375bb5e3a8dcf7840412462182ce0b9433/cmd/stdiscosrv/replication.go#L248-L252

Namely, unless there is enough traffic on the wire, the replication connection will be torn down. What is more, the is no reconnect on the other side: https://github.com/syncthing/syncthing/blob/2c9d96375bb5e3a8dcf7840412462182ce0b9433/cmd/stdiscosrv/replication.go#L105-L110

So basically, if there is not enough traffic in the replication servers, the replication will not work.

I think the solution would be to not to set a read timeout in https://github.com/syncthing/syncthing/blob/2c9d96375bb5e3a8dcf7840412462182ce0b9433/cmd/stdiscosrv/replication.go#L245 - i.e. wait indefinitely (preferable), or handle reconnects on the sending side (more work), or implement heartbeats in the replication protocol (quite a bit of work).

Let me know if this sounds reasonable. I can prepare a patch.

### Version Information
```
stdiscosrv v0.14.46-ds1 (go1.10rc2 linux-arm)
stdiscosrv v0.14.49 (go1.9.4 linux-amd64)
```

vespian:
### Purpose

This PR fixes #5117 - it adds keepalive support to stdiscosrv

This change should be backwards-compatible, i.e. older versions of stdiscosrv should ignore the heartbeats, though it has not been tested.

### Testing

Manually tested the replication keepalives by setting up a replicated setup of two stdiscosrv processes and testing if the connection is not being torn down after 60 seconds as it used to.

### DIFF
--- a/cmd/stdiscosrv/replication.go
+++ b/cmd/stdiscosrv/replication.go
@@ -18,6 +18,9 @@ import (
 	"github.com/syncthing/syncthing/lib/protocol"
 )
 
+const replicationReadTimeout = time.Minute
+const replicationHeartbeatInterval = time.Second * 30
+
 type replicator interface {
 	send(key string, addrs []DatabaseAddress, seen int64)
 }
@@ -79,10 +82,22 @@ func (s *replicationSender) Serve() {
 		return
 	}
 
+	heartBeatTicker := time.NewTicker(replicationHeartbeatInterval)
+	defer heartBeatTicker.Stop()
+
 	// Send records.
 	buf := make([]byte, 1024)
 	for {
 		select {
+		case <-heartBeatTicker.C:
+			if len(s.outbox) > 0 {
+				// No need to send heartbeats if there are events/prevrious
+				// heartbeats to send, they will keep the connection alive.
+				continue
+			}
+			// Empty replication message is the heartbeat:
+			s.outbox <- ReplicationRecord{}
+
 		case rec := <-s.outbox:
 			// Buffer must hold record plus four bytes for size
 			size := rec.Size()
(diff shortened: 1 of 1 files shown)


calmh:
I think old version will actually do a database put with an empty key, not sure if that errors out somehow. If it doesn't, I think this is fine as it doesn't cause any real issues. If it does, the replication message could be for a valid but dummy device ID (protocol.LocalDeviceID for example).

Regardless, losing a message if the connection goes down is not something I worry about. These are not unique, precious works of art. It'll get refreshed in 30 minutes.

calmh:
Apart from that, LGTM

PR #5123, opened 2018-08-14
all: Limit requests from remotes

imsodin:
### Purpose

I think this came up in the forum more or less recently, but I don't know the context anymore. Anyhow currently there is no limit on simultaneous requests, which means there is potentially lots of concurrent disk access and theoretically unlimited RAM usage. I reused `byteSemaphore` from pulling (moving it to the util package) to restrict active requests to `2 * MaxBlockSize` (magic).

### Testing

None.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -15,7 +15,6 @@ import (
 	"runtime"
 	"sort"
 	"strings"
-	stdsync "sync"
 	"time"
 
 	"github.com/syncthing/syncthing/lib/config"
@@ -28,6 +27,7 @@ import (
 	"github.com/syncthing/syncthing/lib/scanner"
 	"github.com/syncthing/syncthing/lib/sha256"
 	"github.com/syncthing/syncthing/lib/sync"
+	"github.com/syncthing/syncthing/lib/util"
 	"github.com/syncthing/syncthing/lib/versioner"
 	"github.com/syncthing/syncthing/lib/weakhash"
 )
@@ -1289,7 +1289,7 @@ func verifyBuffer(buf []byte, block protocol.BlockInfo) error {
 }
 
 func (f *sendReceiveFolder) pullerRoutine(in <-chan pullBlockState, out chan<- *sharedPullerState) {
-	requestLimiter := newByteSemaphore(f.PullerMaxPendingKiB * 1024)
+	requestLimiter := util.NewByteSemaphore(f.PullerMaxPendingKiB * 1024)
 	wg := sync.NewWaitGroup()
 
 	for state := range in {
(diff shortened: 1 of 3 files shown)


imsodin:
@AudriusButkevicius Not according to this comment:

```golang
			// Requests are handled asynchronously
			go c.handleRequest(*msg)
```
https://github.com/syncthing/syncthing/blob/master/lib/protocol/protocol.go#L395

calmh:
I think this limiting should happen model-side and not protocol-side...

imsodin:
That was my first intention too, but then I saw that the buffer is created and released in the protocol, therefore I thought it make more sense to limit around that. Otherwise the bytes are released to the limiter before they are actually released from the buffer (i.e. it's released for the entire time of actually sending the data to the remote).

AudriusButkevicius:
Also, I think this value should be adjustable, together with the receive buffer.

AudriusButkevicius:
The fact the value is not released to the pool is ok, as it was not released before anyway.

imsodin:
@AudriusButkevicius 

What exactly do you want to have adjustable - the amount of bytes that can be concurrently requested by a remote and what else?

And you don't consider it a problem, that when limiting on the model side of things, the limit is not enforced while sending the data, only while reading it? Meaning disk IO will be limited, but RAM usage not - at least not in the scenario where the time required to read the data is smaller than the time required to send it to the remote.

calmh:
I think the only thing we *need* to limit is RAM usage, so this needs to happen where the buffer is allocated. Given that the buffers are block-sized regardless of the request size (I seem to remember? no checking now) I think we can do something simple here. It's also the case that handling multiple requests from one connection doesn't buy us much performance - rather the opposite. I think we could something as drastic as just processing at most 4 requests in parallel per connection and it would be fine.

We need to *send* lots of requests because we want lots of data in flight over potentially seconds of network latency. That doesn't mean we need to try to process them in parallel on the sending side.

AudriusButkevicius:
Yeah, I guess having the byte pool allocate 10000 buffers who then all block on the bytes semaphore is not ideal.

calmh:
Another stray thought here... Having the protocol package `go` the request messages might not be the right thing. Removing that and having the model do the needful to do I/O in the background after applying limiting etc might be better.

xjtdy888:
I think we should consider limiting the number of requests per connection, and more distributed data requests, 
I agree that it should not be on the protocol, 
We have another issue related to this,
In Pull, model.Availability returns a list of candidate devices. Assuming that only one device is currently connected, all Pulls will request data from this connection. In the meantime, if we connect more devices and have candidates Qualification, but we have no chance to request data from it, all the pressure is concentrated on the device that was connected at the beginning

imsodin:
See #5216 for the followup PR that is a prototype which takes the approach suggested in the comments above.

PR #5126, opened 2018-08-15
lib/model: Always release the lock

imsodin:
This is a regression from #4974 and thus affecting v0.14.49. Discovered thanks to this report:
https://forum.syncthing.net/t/panic-deadlock-detected-at-fmut/11976  
The regression being present in v0.14.49 makes this no v0.14.50 RC material. However as it is pretty bad, I'd propose to make an exception and include it anyway.

Not that it makes it any better, but the reason for including the config stuff in the `fmut` lock was probably the edge case I now handled by checking that `m.folderCfgs[folder]` is really present.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1308,8 +1308,6 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset 
 		return protocol.ErrInvalid
 	}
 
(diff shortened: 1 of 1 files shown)


calmh:
Lovely. We have the time, so might as well drop this into a new .50 RC yes.

PR #5128, opened 2018-08-15
lib/model: Small fixes to test convenience functions

imsodin:
While writing a test (not this PR) I noticed a problem in tests addressed in this PR:  Temporary files were deleted in setup functions, instead of at the end of tests:  
 - `setupModelWithConnectionManual`:  
https://github.com/syncthing/syncthing/pull/5128/files#diff-5943690a8b9407e04ad15ac110b5d9c0L626  
 - `newState`:  
https://github.com/syncthing/syncthing/pull/5128/files#diff-9accdb1b2ab06368238d103a1f87b3c6L159).

Also `newState` added folders, but didn't start them: https://github.com/syncthing/syncthing/pull/5128/files#diff-9accdb1b2ab06368238d103a1f87b3c6R163

And `createTmpWrapper` needlessly returned it's path, which is redundant with `wrapper.ConfigPath()`.

I am aware the diff is horrible, but the only non-cosmetical changes are those I mentioned above.

### DIFF
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -129,12 +129,11 @@ func TestMain(m *testing.M) {
 		panic(err)
 	}
 
-	var wrapperPath string
-	defaultCfgWrapper, wrapperPath = createTmpWrapper(defaultCfg)
+	defaultCfgWrapper = createTmpWrapper(defaultCfg)
 
 	exitCode := m.Run()
 
-	os.Remove(wrapperPath)
+	os.Remove(defaultCfgWrapper.ConfigPath())
 	defaultFs.Remove(tmpName)
 	defaultFs.RemoveAll(config.DefaultMarkerName)
 	defaultFs.RemoveAll(tmpLocation)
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
So it's all in tests and they still all pass which I guess means this is good to go, yet I am suspicious of adding the StartFolder yet not having to remove it anywhere. I also know there were some paused folder tests which might have relied on the model being with non-started folders, yet they still seem to pass.

AudriusButkevicius:
I guess double starts would cause a panic.

imsodin:
@AudriusButkevicius That's the thing: Paused folders are both stopped and removed in the model. It's actually a bit useless that we have separate `AddFolder` and `StartFolder` functions, one is never called without the other. That's actually something that should be changed at some point too.

AudriusButkevicius:
I am pretty sure paused folders used to be added but not started, as we used to check their existance in the model.
I guess we now check against config rather than the model, so perhaps this is where the disconnect happened.

Also, I think ClusterConfig sends not started folders too

imsodin:
No, that was actually a point of disagreement between us when you added paused folders: I thought they should be added but not started and you argued for removing them entirely. Checking config instead of model came much later. And yes, `ClusterConfig` does send paused devices by going over devices from the config. Anyway, I will wait a moment to give @calmh a chance to object to this if he wants to, then I'll merge (I have a followup PR waiting).

PR #5129, opened 2018-08-16
lib/model: Fix regressions detecting deletes/ignores (fixes #5125, fixes #5127)

### LINKED ISSUE #5125: symlinks being removed with impunity
symlinks created on a Linux maching are being removed by a Windows machine. Syncthing used to warn it cannot sync it, but for a while now this is no longer true.

The links are within the repository, so are supported (or at least used to be). Even I forget to run as admin in the Windows machine, I do not want it to delete files!

I would be happy run a special debug mode,
thanks

### LINKED ISSUE #5127: Parent directories of unignored files keep being included and immediately ignored again
All credits to cosas for being tenacious and providing perfect logs, until I finally understood the issue: https://forum.syncthing.net/t/did-something-recently-changed-in-ignores-handling-part-2/11979

In #4811 the scanner was made to include items below ignored parent directories, but I didn't adjust the model side of things. In the model the db is walked and to detect deleted and newly ignored files. Ignored parents of included items must be filtered out as well.

Reproducer (sloppy, unit test already exists):

1. Add ignores like this
```
!foo
*
```
2. Create a file `/bar/foo`.

Result: On every scan, `bar` features as modified in the recent changes list.  

Expected: `bar` should feature once in that list, then a scan shouldn't do anything.

imsodin:
Ready for review - windows only test was verified to work and fix is now in place.  
<strike>**Do not merge**. This does not yet contain the fix for #5125, only the unit test that should reproduce it (<strike>i.e. if the new windows test passes, that's a failure</strike> it fails :) ).</strike>

### Purpose

When recurse unignoring (#4811), not only the scanner side of things should have been adapted, the model side of things (detecting ignores and deletes) as well.  
When adding more fine grained control over invalid files via local flags (#4952), the existing `IsIgnored` condition isn't enough anymore to filter files before checking whether they were deleted.

### Testing

Unit tests for both issues.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -2090,24 +2090,6 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 	batch.reset()
 	var toIgnore []db.FileInfoTruncated
 	ignoredParent := ""
-	ignore := func(f db.FileInfoTruncated) {
-		l.Debugln("marking file as ignored", f)
-		nf := f.ConvertToIgnoredFileInfo(m.id.Short())
-		batch.append(nf)
-		changes++
-	}
-	batchToIgnore := func() error {
-		for _, f := range toIgnore {
-			ignore(f)
-
-			if err := batch.checkFlush(); err != nil {
-				return err
-			}
-		}
-		toIgnore = toIgnore[:0]
-		ignoredParent = ""
-		return nil
-	}
 	pathSep := string(fs.PathSeparator)
 	for _, sub := range subDirs {
 		var iterError error
@@ -2121,10 +2103,18 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 			}
 
 			if ignoredParent != "" && !strings.HasPrefix(f.Name, ignoredParent+pathSep) {
-				if err := batchToIgnore(); err != nil {
-					iterError = err
-					return false
+				for _, f := range toIgnore {
+					l.Debugln("marking file as ignored", f)
+					nf := f.ConvertToIgnoredFileInfo(m.id.Short())
+					batch.append(nf)
+					changes++
+					if err := batch.checkFlush(); err != nil {
+						iterError = err
+						return false
+					}
 				}
+				toIgnore = toIgnore[:0]
+				ignoredParent = ""
 			}
 
 			switch ignored := ignores.Match(f.Name).IsIgnored(); {
@@ -2141,7 +2131,10 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
 					return true
 				}
 
-				ignore(f)
+				l.Debugln("marking file as ignored", f)
+				nf := f.ConvertToIgnoredFileInfo(m.id.Short())
+				batch.append(nf)
+				changes++
 
 			case f.IsIgnored() && !ignored:
 				// Successfully scanned items are already un-ignored during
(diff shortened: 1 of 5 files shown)


AudriusButkevicius:
The colsure functions are turning into terrible incomprehensible spagetti code :(

imsodin:
@AudriusButkevicius Does this make it more readable and less spaghetti-like?

I am aware the whole thing is far from the ideal of simple and elegant. The elegant solution was combining walking the disk and database and that failed due to the case-sensitivity issue.

dschouten:
Is this going to be reviewed and accepted soon? Syncthing on Windows is severely affected when connected to Linux machines without this fix ... symlinks disappearing all the time is a big pain.  Thanks!

calmh:
It will, but the release you'll see it in is 0.14.51 on the first Tuesday of October (release candidate a bit earlier) so the review time here isn't super critical.

lib/model/model.go
```diff
@@ -1738,32 +1738,33 @@ func sendIndexes(conn protocol.Connection, folder string, fs *db.FileSet, ignore
...
+			if err = conn.Index(folder, fs); err != nil {
+				return err
+			}
+			l.Debugln(debugMsg(msg[0]))
+			initial = false
+		} else {
+			if err = conn.IndexUpdate(folder, fs); err != nil {
+				return err
 			}
+			l.Debugln(debugMsg(msg[0]))
```
calmh: `msg[1]`? Actually, just skip the debug logging, it serves not much purpose and isn't worth the complication of the `msg` stuff imho

lib/model/model.go
```diff
@@ -1742,22 +1742,17 @@ func sendIndexTo(prevSequence int64, conn protocol.Connection, folder string, fs
...
 			initial = false
-		} else {
-			if err = conn.IndexUpdate(folder, fs); err != nil {
+			if err = conn.Index(folder, fs); err != nil {
 				return err
 			}
-			l.Debugln(debugMsg(msg[0]))
+			return nil
+		}
+		if err = conn.IndexUpdate(folder, fs); err != nil {
```
calmh: actually, `return conn.Index…` should suffice for both of these

calmh: or wait, `err` is shadowed? is that intentional? maybe I need to read this again

calmh: It looks to me like the error is properly handled in the return path, so it should not be necessary to closure over it....

imsodin: The return thing was a stupid oversight and the error declaration is in place for `WithHaveSequence` later on - I moved that back to make that more clear.

lib/model/model.go
```diff
@@ -1738,32 +1738,22 @@ func sendIndexes(conn protocol.Connection, folder string, fs *db.FileSet, ignore
...
 	deviceID := conn.ID()
-	batch := make([]protocol.FileInfo, 0, maxBatchSizeFiles)
-	batchSizeBytes := 0
 	initial := prevSequence == 0
-	var err error
-	var f protocol.FileInfo
-	debugMsg := func(t string) string {
-		return fmt.Sprintf("Sending indexes for %s to %s at %s: %d files (<%d bytes) (%s)", folder, deviceID, conn, len(batch), batchSizeBytes, t)
+	batch := newFileInfoBatch(nil)
+	batch.flushFn = func(fs []protocol.FileInfo) error {
```
AudriusButkevicius: Why not pass this in the "constructor"?

Issue #837, opened 2014-10-13
Ignore some files by default?


gunwald:
I would suggest to introduce a global ignore list, that has some default entries. For example:
   *.git/* # ignore git repros 
   *.sparkleshare/* # ignore git sparkleshare 
   *.~lock.* # ignore LibreOffice lock files
   *.owncloudsync* # ignore owncloud

It is annoying to add to each folder on each node every time all files you want to be ignored.

Zillode:
I think we should reconsider this. We should ignore operating system files that are known to cause conflicts as soon as users start using Syncthing. My current list of such files consists of:

```
".DS_Store", // OSX (https://github.com/syncthing/syncthing/issues/826)
".Spotlight-V100", // OSX
"Thumbs.db", // Windows
"lost+found", // Linux (https://discourse.syncthing.net/t/un-ignore-subfolders-files-of-ignored-folder/1460/10 , https://github.com/syncthing/syncthing/issues/1090)
```
I suggest to list these as the default .stignore content.

AudriusButkevicius:
`
".DS_Store",
"Thumbs.db",
"._*",
".Spotlight-V100",
".Trashes",
"desktop.ini",
`

@Zillode's list make sense,
I think we should have a default set of OS specific things (rather than .svn et al) @calmh ?

facastagnini:
I build my exclusion list using the expressions relevant to my OS from here:

http://support.code42.com/CrashPlan/Latest/Troubleshooting/What_Is_Not_Backing_Up

I think is a comprehensive list.

generalmanager:
I was going to propose my own list, but the CrashPlan one is way more comprehensive. Nice find @facastagnini !

facastagnini:
You are welcome!

I started using Syncthing and Crashplan pretty much at the same time, so
when I was building a common ignore list for both tools I stumble upon that.

Since we are linking that list I feel it is appropriated to thanks them for
sharing that comprehensive ignore list, and give them some credit saying
that Crashplan is an awesome backup tool that complements my Syncthing
setup.

flungo:
I am looking to implement this, if it hasn't already been started. Not sure if it would be best to have this hard coded into the binary, or if to create the global ignore list suggested populated by default with the system generated files. The global ignore list would probably be best located in the user config directory (alongside the config.xml).

I think as a minimum the list should include:

    .DS_Store
    .DS_Store?
    ._*
    .Spotlight-V100
    .Trashes
    .Trash-*
[...]

PR #5132, opened 2018-08-18
lib/ignore Have a nice treeview for handling ignores - Next Gen Ignores

ibizaman:
### Purpose

This is PR attempts to address the requirements of issue #2491 by following the rules outlined in [this comment](https://github.com/syncthing/syncthing/issues/2491#issuecomment-359889677).

This PR is in progress. I posted it [as requested](https://github.com/syncthing/syncthing/issues/2491#issuecomment-413694849). What's notably missing is advanced testing of part of it, see [this comment](https://github.com/syncthing/syncthing/issues/2491#issuecomment-413598000).

Please see commit messages for info on the implementation.

### Testing

Unfortunately testing needs to be manual.

What would be greatly needed to gain confidence in this PR is a way to unit test the new js functions. The main scary one is the function that parses the .stignore to populate the basic view. It doesn't work right now correctly. It is a pain to test manually though.

I'm less woried about the UI part - the update of the UI when a user clicks - as there is not that much going on on that front. I'm not saying it's flawless, just that it is possible to exhaustively test the behavior with manual testing.

### Screenshots

Clicking on the Ignore Patterns shows 2 tabs, the Basic one is selected by default:
image

This shows the advanced tab:
image

Opening the folder asynchronously loads the contents by getting the global state from syncthing
image

Selectively syncing a few files:
image

Corresponding advanced rules:
image

Another configuration:
image

And corresponding advanced rules:
image

Extreme case 1, sync all:
image

Corresponding pattern is empty:
image

Extreme case 2, ignore all:
image

Corresponding patterns:
image

### Documentation

I guess we should discuss documentation later when the implementation is more solid? Although I'm totally fine to tackle this now if requested.

Disclaimer: first PR here, I don't want to waste anyone's time so feel free to just skim over it and tell me what could help you review this more easily.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1569,16 +1569,16 @@ angular.module('syncthing.core')
             });
         };
 
-	$scope.syncedFiles = {
-	    parseable: false,
-	    all: false,
-	    files: [],
-	    reset: function() {
-		$scope.syncedFiles.parseable = false;
-		$scope.syncedFiles.all = false;
-		$scope.syncedFiles.files = [];
-	    }
-	};
+        $scope.syncedFiles = {
+            parseable: false,
+            all: false,
+            files: [],
+            reset: function() {
+                $scope.syncedFiles.parseable = false;
+                $scope.syncedFiles.all = false;
+                $scope.syncedFiles.files = [];
+            }
+        };
 
         $scope.editFolder = function (folderCfg) {
             $scope.editingExisting = true;
@@ -1626,70 +1626,70 @@ angular.module('syncthing.core')
 
             function enableBasicRules(enable) {
                 if (enable) {
-		    console.log('enabling basic rules');
+                    console.log('enabling basic rules');
                     $('#folder-ignores #folderIgnoresTree').removeAttr('disabled');
                     //$('#folder-ignores #folderIgnoresTree').show();
                 } else {
-		    console.log('disabling basic rules');
+                    console.log('disabling basic rules');
                     $('#folder-ignores #folderIgnoresTree').attr('disabled', 'disabled');
                     //$('#folder-ignores #folderIgnoresTree').hide();
                 }
             }
 
             function writeAdvancedRules() {
-		console.log('writeAdvancedRules');
-		if ($scope.syncedFiles.all) {
+                console.log('writeAdvancedRules');
+                if ($scope.syncedFiles.all) {
                     $('#folder-ignores textarea').val('');
-		    return;
-		}
+                    return;
+                }
 
                 var rules = ['*'];
-		$scope.syncedFiles.files.forEach(function (file) {
-		    console.log('file', file);
-		    var fileSplit = file.split('/');
-		    var fileDir = fileSplit.slice(0, fileSplit.length - 1).join('/');
-		    rules.push('!/' + file);
-		    if (fileDir === '') {
-			rules.push('*');
-		    } else {
-			rules.push('/' + fileDir + '/*');
-		    }
-		});
+                $scope.syncedFiles.files.forEach(function (file) {
+                    console.log('file', file);
+                    var fileSplit = file.split('/');
+                    var fileDir = fileSplit.slice(0, fileSplit.length - 1).join('/');
+                    rules.push('!/' + file);
+                    if (fileDir === '') {
+                        rules.push('*');
+                    } else {
+                        rules.push('/' + fileDir + '/*');
+                    }
+                });
 
-		rules = rules.sort(function(a, b) {
-		    var a_toSort = a;
-		    if (a_toSort.startsWith('!')) {
-			a_toSort = a_toSort.slice(1, a_toSort.length).concat('1');
-		    } else {
-			a_toSort = a_toSort.concat('2');
-		    }
+                rules = rules.sort(function(a, b) {
+                    var a_toSort = a;
+                    if (a_toSort.startsWith('!')) {
+                        a_toSort = a_toSort.slice(1, a_toSort.length).concat('1');
+                    } else {
+                        a_toSort = a_toSort.concat('2');
+                    }
 
-		    var b_toSort = b;
-		    if (b_toSort.startsWith('!')) {
-			b_toSort = b_toSort.slice(1, b_toSort.length).concat('1');
-		    } else {
-			b_toSort = b_toSort.concat('2');
-		    }
+                    var b_toSort = b;
+                    if (b_toSort.startsWith('!')) {
+                        b_toSort = b_toSort.slice(1, b_toSort.length).concat('1');
+                    } else {
+                        b_toSort = b_toSort.concat('2');
+                    }
 
-		    return a_toSort <= b_toSort;
-		}).filter(function(value, index, self) {
-		    return self.indexOf(value) === index
-		});
+                    return a_toSort <= b_toSort;
+                }).filter(function(value, index, self) {
+                    return self.indexOf(value) === index
+                });
 
-		console.log('writeAdvancedRules - rules:', rules);
-		$('#folder-ignores textarea').val(rules.join('\n'));
+                console.log('writeAdvancedRules - rules:', rules);
+                $('#folder-ignores textarea').val(rules.join('\n'));
             }
 
             function parseAdvancedRules() {
-		/*
+                /*
                  * parse:
                  *
-		 *    !/c/c
-		 *    !/c/a
-		 *    /c/*
-		 *    !/b
-		 *    !/a
-		 *    *
+                 *    !/c/c
+                 *    !/c/a
+                 *    /c/*
+                 *    !/b
+                 *    !/a
+                 *    *
                  *
                  * into object with:
                  *
@@ -1699,60 +1699,60 @@ angular.module('syncthing.core')
                  *     files: ['/c/c', '/c/a', '/b', '/a'],
                  *    }
                  *
-		 */
-		console.log('parseAdvancedRules');
-		var rules = $('#folder-ignores textarea').val().split('\n');
+                 */
+                console.log('parseAdvancedRules');
+                var rules = $('#folder-ignores textarea').val().split('\n');
                 $scope.syncedFiles.reset();
-		$scope.syncedFiles.parseable = true;
-		$scope.syncedFiles.all = true;
+                $scope.syncedFiles.parseable = true;
+                $scope.syncedFiles.all = true;
 
                 var prevRule = null;
                 for (const rule of rules) {
-		    console.log('rule', rule, 'prevRule', prevRule);
+                    console.log('rule', rule, 'prevRule', prevRule);
                     if (rule === '') {
                         continue;
                     } else {
-			$scope.syncedFiles.all = false;
-		    }
+                        $scope.syncedFiles.all = false;
+                    }
 
-		    // TODO: handles cases with multiple prevRule and one rule (multiple files in same dir)
-		    if (rule === '*') {
-			continue;
-		    } else if (prevRule === null) {
-			if (!rule.startsWith('!')) {
-			    console.log('not start with !')
+                    // TODO: handles cases with multiple prevRule and one rule (multiple files in same dir)
+                    if (rule === '*') {
+                        continue;
+                    } else if (prevRule === null) {
+                        if (!rule.startsWith('!')) {
+                            console.log('not start with !')
                             $scope.syncedFiles.parseable = false;
-			    console.log('parseAdvancedRules - syncedFiles:', $scope.syncedFiles);
+                            console.log('parseAdvancedRules - syncedFiles:', $scope.syncedFiles);
                             return;
                         } else {
-			    prevRule = rule;
-			}
+                            prevRule = rule;
+                        }
                     } else {
-			var prevRuleSplit = prevRule.split('/');
-			//var prevRulePath = ([prevRuleSplit[0].slice(1, prevRuleSplit[0].length)].concat(prevRuleSplit.slice(1, prevRuleSplit.length - 1))).join('/');
-			// Remove leading ! and basename
-			var prevRulePath = prevRuleSplit.slice(0, prevRuleSplit.length - 1).join('/');
-			// Remove leading !
-			var rulePathWithBasename = prevRule.slice(1, prevRule.length);
-			// Remove last *
-			var ruleSplit = rule.split('/');
-			var rulePath = ruleSplit.slice(0, prevRuleSplit.length - 1).join('/');
-			console.log('prevRulePath', prevRulePath, 'rulePath', rulePath)
+                        var prevRuleSplit = prevRule.split('/');
+                        //var prevRulePath = ([prevRuleSplit[0].slice(1, prevRuleSplit[0].length)].concat(prevRuleSplit.slice(1, prevRuleSplit.length - 1))).join('/');
+                        // Remove leading ! and basename
+                        var prevRulePath = prevRuleSplit.slice(0, prevRuleSplit.length - 1).join('/');
+                        // Remove leading !
+                        var rulePathWithBasename = prevRule.slice(1, prevRule.length);
+                        // Remove last *
+                        var ruleSplit = rule.split('/');
+                        var rulePath = ruleSplit.slice(0, prevRuleSplit.length - 1).join('/');
+                        console.log('prevRulePath', prevRulePath, 'rulePath', rulePath)
 
-			if (prevRulePath !== rulePath) {
-			    console.log('prevRulePath !== rulePath')
+                        if (prevRulePath !== rulePath) {
+                            console.log('prevRulePath !== rulePath')
                             $scope.syncedFiles.parseable = false;
-			    console.log('parseAdvancedRules - syncedFiles:', $scope.syncedFiles);
+                            console.log('parseAdvancedRules - syncedFiles:', $scope.syncedFiles);
                             return;
-			} else {
-			    console.log(rulePathWithBasename)
-			    $scope.syncedFiles.files.push(rulePathWithBasename);
-			    prevRule = null;
-			}
-		    }
+                        } else {
+                            console.log(rulePathWithBasename)
+                            $scope.syncedFiles.files.push(rulePathWithBasename);
+                            prevRule = null;
+                        }
+                    }
                 }
 
-		console.log('parseAdvancedRules - syncedFiles:', $scope.syncedFiles);
+                console.log('parseAdvancedRules - syncedFiles:', $scope.syncedFiles);
             }
 
             enableBasicRules(false);
@@ -1783,54 +1783,54 @@ angular.module('syncthing.core')
                 }
             });
 
-	    function updateBasicRulesFromSyncedFiles() {
-		console.log('updateBasicRulesFromSyncedFiles');
-		enableBasicRules($scope.syncedFiles.parseable);
+            function updateBasicRulesFromSyncedFiles() {
+                console.log('updateBasicRulesFromSyncedFiles');
+                enableBasicRules($scope.syncedFiles.parseable);
 
-		if ($scope.syncedFiles.all === true) {
-		    $('#folder-ignores #folderIgnoresTree').fancytree('getTree').rootNode.children[0].setSelected(true);
-		} else {
-		    $('#folder-ignores #folderIgnoresTree').fancytree('getTree').visit(function(node) {
-			if ($scope.syncedFiles.files.includes(node.key)) {
-			    node.setSelected(true);
-			}
-		    });
-		}
-	    }
+                if ($scope.syncedFiles.all === true) {
+                    $('#folder-ignores #folderIgnoresTree').fancytree('getTree').rootNode.children[0].setSelected(true);
+                } else {
+                    $('#folder-ignores #folderIgnoresTree').fancytree('getTree').visit(function(node) {
+                        if ($scope.syncedFiles.files.includes(node.key)) {
+                            node.setSelected(true);
+                        }
+                    });
+                }
+            }
 
-	    function updateSyncedFilesFromBasicRules() {
-		console.log('updateSyncedFilesFromBasicRules');
-		$scope.syncedFiles.reset();
-		$scope.syncedFiles.parseable = true;
+            function updateSyncedFilesFromBasicRules() {
+                console.log('updateSyncedFilesFromBasicRules');
+                $scope.syncedFiles.reset();
+                $scope.syncedFiles.parseable = true;
 
-		$('#folder-ignores #folderIgnoresTree').fancytree('getTree').visit(function(node) {
-		    if (node.isSelected()) {
-			if (node.key === '') {
-			    // If we include the whole tree, no need
-			    // for any rule so we break now
-			    $scope.syncedFiles.all = true;
-			    return false;
-			} else if (!node.parent.isSelected()) {
-			    $scope.syncedFiles.files.push(node.key);
-			}
-		    }
-		    return true;
-		});
+                $('#folder-ignores #folderIgnoresTree').fancytree('getTree').visit(function(node) {
+                    if (node.isSelected()) {
+                        if (node.key === '') {
+                            // If we include the whole tree, no need
+                            // for any rule so we break now
+                            $scope.syncedFiles.all = true;
+                            return false;
+                        } else if (!node.parent.isSelected()) {
+                            $scope.syncedFiles.files.push(node.key);
+                        }
+                    }
+                    return true;
+                });
 
-	    }
+            }
 
             // When switching to the Basic tab, parse the current ignore
             // rules and either disable the tree if the rules are too
             // complicated or pre-check the currently not-ignored files
             // and folders.
             $('#folder-ignores .nav-tabs a').on('shown.bs.tab', function(event) {
-		console.log('change tab', event);
+                console.log('change tab', event);
                 if (event.target.getAttribute('href') === '#folder-ignores-basic') {
-		    parseAdvancedRules();
-		    updateBasicRulesFromSyncedFiles();
+                    parseAdvancedRules();
+                    updateBasicRulesFromSyncedFiles();
                 } else {
-		    updateSyncedFilesFromBasicRules();
-		    writeAdvancedRules();
+                    updateSyncedFilesFromBasicRules();
+                    writeAdvancedRules();
                 }
             });
 
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1615,11 +1615,41 @@ angular.module('syncthing.core')
 
             $('#folder-ignores textarea').val($translate.instant("Loading..."));
             $('#folder-ignores textarea').attr('disabled', 'disabled');
+            $('#folder-ignores #folderIgnoresTree').attr('disabled', 'disabled');
             $http.get(urlbase + '/db/ignores?folder=' + encodeURIComponent($scope.currentFolder.id))
                 .success(function (data) {
                     $scope.currentFolder.ignores = data.ignore || [];
                     $('#folder-ignores textarea').val($scope.currentFolder.ignores.join('\n'));
+
+                    $('#folder-ignores #folderIgnoresTree').fancytree({
+                        source: [
+                            {title: $scope.currentFolder.id, key: '', lazy: true}
+                        ],
+                        checkbox: true,
+                        clickFolderMode: 2, // Expand, no-select, when click on folder
+                        selectMode: 0,      // Select one node only
+                        tabindex: "0",      // Tree control can be reached using TAB keys
+                        lazyLoad: function(event, data) {
+                            var dfd = $.Deferred();
+                            data.result = dfd.promise();
+                            console.log("data", data);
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
So first problem I see is that your includes are per file, so for example adding a new file to the folder might not sync it. Also, the code partially assumes old ignores behaviour, where including some specific file requires excluding files in the parent directory (which I think is no longer needed).

This could probably be fixed by essentially including everything under a directory once it has been included (which the ignores handle already, so you don't need to include individual files).

I also imagine people wanting an inverse mode, where people exclude things rather than include.

Also, you should clean up and format the code before someone digs into it, as it's hard to follow with indentation being all over the shop.

scienmind:
IMO Audrius is right, and the inverse would make more sense (and we're talking about _ignores_ after all).
That'd have 2 consequences:

* selecting a directory/file would exclude it (some UI clue would be appropriate, maybe dim the selected folders, or add some other "ignored emblem")
* any new files/folders outside of "excluded tree" will be synced, and not ignored

Another thing to consider is how "Basic" would respond to manual (aka advanced) ignore patterns. For example, if someone ignores files/folders by their partial name.
In that case one would also probably expect the basic view not to "overwrite" the manual ignores accidentally.

P.S.: If **not** going with the inverse (i.e. selecting what to sync, not what to ignore), it should probably be titled "Selective Sync" then, or something in that spirit.

ibizaman:
I'm fine going with either exclude or include. I agree it makes more sense to exclude a folder/file that is checked. We could have both but I'd pick one for this PR, to keep things simple.

> * selecting a directory/file would exclude it (some UI clue would be appropriate, maybe dim the selected folders, or add some other "ignored emblem")
> * any new files/folders outside of "excluded tree" will be synced, and not ignored

This is IMO a very simple set of rule, with no surprising implicit behavior.

> Also, the code partially assumes old ignores behaviour, where including some specific file requires excluding files in the parent directory (which I think is no longer needed).

True I didn't check the ignore rules since quite some time now. I checked https://docs.syncthing.net/users/ignoring.html but don't see what you meant.

> Another thing to consider is how "Basic" would respond to manual (aka advanced) ignore patterns. For example, if someone ignores files/folders by their partial name.
> In that case one would also probably expect the basic view not to "overwrite" the manual ignores accidentally.

This looks quite complicate to handle. The basic view should indeed avoid overwriting part of the .stignore.

AudriusButkevicius:
Re overwriting, my initial idea was to split ignores into two, but to be honest, your approach is much simpler, as it results in no go code changes. You could wrap the generated ignores in a comment block asking users not to mess around.

ibizaman:
I like the comment block idea.

I guess the block should be at the top then, to avoid the user adding a rule above it that would change the behavior of a rule in the comment block. I'm thinking about the following special case from https://docs.syncthing.net/users/ignoring.html#patterns:

> Note that files in ignored directories can not be re-included this way. This is due to the fact that Syncthing stops scanning when it reaches an ignored directory, so doesn’t know what files it might contain.

calmh:
That part you quote is incorrect nowadays, should be removed.

calmh:
(I've fixed the docs)

ibizaman:
Do we agree on this?

The basic tree view will populate a comment block in the advanced (text) view.

# Comment block:
  - I have no definitive idea for the format of the comment block, I'm proposing one to begin discussion
  ```
  ###############################
  # Syncthing basic rules block #
  ###############################
  # This code block is handled automatically by Syncthing.
  # Any manual change here will be overwritten when switching to the Basic tab.
  rule1
  rule2
  # End of Syncthing basic rules block
  ```
  - The parser will search for a comment block with the following regex rules:
    - The block of comments begins by a line of `#`, followed by a line beginning by `#` and containing the words `basic` (ignoring case), followed by a line of `#`.
    - The footer - the last line - is a line beginning by `#` and containing the word `end` and `basic` (ignoring case).
  - When writing the basic rules in the comment block, the parser will erase the whole block of code and re-insert the header and footer as well as the rules.
    - If no comment block found, the parser will insert one at the end.

# Treeview UI

- There's an editable checkbox next to each folder and file.
- Checking a folder also checks automatically all folders and files underneath.
- Visually make it obvious that checked files are excluded.

# Treeview -> Text generation
In pseudocode:
```python
for all nodes:
    if node is not checked:
        do nothing
    if node.parent is checked:
        do nothing
    add node.fullPath to rules
```

# Text -> Treeview
In pseudocode:
```python
for all rules:
    parse rule to full path of node
    if corresponding node is already checked:
        do nothing
    if parent of corresponding node is already checked:
        do nothing
    check corresponding node
```

AudriusButkevicius:
@calmh what do you think?

I think we should not put the automatic generated part of the file in the advanced section, and extract it oout before populating the advanced section.

I also, we might have to put the text sepcified ignores before the tree ones, as ignoring an extension should probably happening before including a directory with everything in it.

As I said before I feel we will have to have two modes of the tree, one which is check to include, and one which is check to exclude, as these are generally two different behaviours I expect users to want.

As for checking nodes, it sounds ok, we might want  to display them in the ui differently, for example checking a directory might want to check and disable all children, and it might want to tri-state check the parent indicating that some (if not all) children are checked.

AudriusButkevicius:
Also, thanks on working on this, this is great progress.

xjtdy888:
Strongly Agree @AudriusButkevicius 
> As I said before I feel we will have to have two modes of the tree, one which is check to include, and one which is check to exclude, as these are generally two different behaviours I expect users to want.

calmh:
I agree with @canton7 above about not having two modes, instead having the default set on the folder root and being able to override it at lower levels.

Also I'm not sure we need to support both tree-view and advanced rules at the same time. I think we should default to just a tree view, the actual text-form _rules_ are not visible anywhere in the UI, the `.stignore` file contains just the rules and maybe a top level comment about them being generated by the ignores editor.

For those who are not well served by this, they can click on `Use advanced editor` and get essentially exactly what we have today. Going in the other direction should of course be possibly, but would be accompanied by a warning `Only rules understood by the standard ignore editor will be retained` and the conversion is best effort. Other advanced rules just vanish when you hop into standard mode.

canton7:
> Also I'm not sure we need to support both tree-view and advanced rules at the same time. I think we should default to just a tree view, the actual text-form rules are not visible anywhere in the UI

I think it's nice to at least show the rules, even if we show them in a non-editable way. It's a good starting point if someone wants to move to tweaking them in an advanced way, and will help with the inevitable support requests: being able to say "paste the contents of \<this bit of the UI\>" is easier than saying "expand your ignores tree -- all of it -- and screenshot the whole thing in multiple parts".

canton7:
I think, unfortunately, we also need to go back to the topic of rules which are included by default on various platforms, like `(?d).DS_Store` on Mac, and the same for Thumbs.db on Windows. Not showing these in the tree view would also be a good thing, I think.

ibizaman:
@AudriusButkevicius 

> I think we should not put the automatic generated part of the file in the advanced section, and extract it oout before populating the advanced section.

100% agreed. To confirm, the comment block would only be added/replace when clicking on the save button when populating the `.stignore` file.

> I also [think] we might have to put the text sepcified ignores before the tree ones, as ignoring an extension should probably happening before including a directory with everything in it.

I'm confused as to why the location of the rules is necessary.  When I quoted the docs in [this comment](https://github.com/syncthing/syncthing/pull/5132#issuecomment-414716312) @calmh answered that this was not applicable anymore.

I agree though we should be consistent and always place them either at the top or bottom, not randomly all over the place.

> [...] checking a directory might want to check and disable all children, and it might want to tri-state check the parent indicating that some (if not all) children are checked.

That's actually already handled by the treeview plugin.
ignores - video of tri-state

Although we'll probably need to re-visit this if we go with the 3 modes.
---

@canton7 I like the 3 modes. I don't think it makes it too complicated for non-technical users although we need to come up with a way to make the effect of inherit mode obvious. The last example accomplishes that when nodes inherit from the exclude rule already.

> I think, unfortunately, we also need to go back to the topic of rules which are included by default on various platforms, like (?d).DS_Store on Mac, and the same for Thumbs.db on Windows. Not showing these in the tree view would also be a good thing, I think.

In the current proposal we don't show them in the treeview because only the rules in the Syncthing maintained comment block are parsed from the text form to the treeview form.

@calmh IMO the following two sentences contradict themselves:

> [...] I think we should default to just a tree view, the actual text-form rules are not visible anywhere in the UI, [...]

> For those who are not well served by this, they can click on Use advanced editor and get essentially exactly what we have today. No tree editor, just a big text box. [...]

Furthermore, for me this is what the two tabs achieve already: you can switch from the treeview only mode to the text-form rules only mode. You don't see both at the same time.

> Going in the other direction should of course be possible, but would be accompanied by a warning Only rules understood by the standard ignore editor will be retained and the conversion is best effort. Advanced rules that don't make sense to the tree editor just vanish when you hop into standard mode.

Actually this is also the goal and what proposed above. We even proposed how to make it more graceful by allowing user-defined rules (supposedly advanced rules) and Syncthing generated rules from the treeview to co-exist thanks to the comment block.

imsodin:
> I think it's nice to at least show the rules, even if we show them in a non-editable way. It's a good starting point if someone wants to move to tweaking them in an advanced way, and will help with the inevitable support requests: being able to say "paste the contents of <this bit of the UI>" is easier than saying "expand your ignores tree -- all of it -- and screenshot the whole thing in multiple parts".

I agree this is a nice to have, but I would certainly hide/collapse it by default. More UI elements means more user confusion. Also I'd consider this secondary, if it already is in this PR then that's nice, but I don't think it's a need to have.

> I think, unfortunately, we also need to go back to the topic of rules which are included by default on various platforms, like (?d).DS_Store on Mac, and the same for Thumbs.db on Windows. Not showing these in the tree view would also be a good thing, I think.

Good point. Again I'd say this is a very good thing to have, but I would not decide the fate of this PR on whether this makes it in or not. Mostly because otherwise we doom it to a death by feature creep. If we go with separating tree-view and patterns then I would add a "simple pattern" input field below the tree view where such default ignores can go and the use can add other simple rules (e.g. `*.bak`).

> I agree with @canton7 above about not having two modes, instead having the default set on the folder root and being able to override it at lower levels.

I am not a big fan of this. Tristate buttons are complicated and in my view the main point of this change is to make it simpler for users who are overwhelmed by the current system. If a user wants to leverage the full power of the ignore system, they should use the advanced version.

ibizaman:
Btw, there's some discussion mainly about the inclusion or not of advanced rules in the UI. I have no opinion on what's the best for the end user. I made this PR and the proposal backwards compatible (by including the advanced section) to not remove features. But if we think it's better to remove it that's totally fine for me.

ibizaman:
@calmh got it that makes sense.

If we want only the treeview, what about:

We provide realtime visual feedback in the treeview when the user is creating a new rule.
Nodes rejected are greyed out, nodes included are "normal".

The rules could be a dynamic list of UI text box, with a `+` button to add a new text box. With maybe a second text box to comment the rule (for the user to know what that's for).

We probably don't need a checkbox in the treeview if instead, when a user clicks on a node, a menu with three buttons appears next to it: `always include`, `always exclude` and `reset` (not sure about the naming). Then although we still have some sort of tri-state, it's hopefully more obvious for the end user what actions he can take and what impact those have. Especially with the auto-greying of nodes.

ibizaman:
So I've been thinking lately about what having a realtime visual feedback in the treeview implies. For this to work we need the frontend to have a rule engine that matches the server's.

Either we re-create one in javascript in the frontend, either we expose an endpoint that in some way returns what files are ignored.

I would rather do the second one as it's more maintainable because there's only one rule engine, but then I don't see how to make the roundtrip to the server efficient (sending the user defined rules then returning the list of ignored files).

AudriusButkevicius:
Why is this needed?

ibizaman:
@AudriusButkevicius oh I understood that we wanted to merge advanced and basic view. If we're okay to only have a "basic" treeview, then that makes it much easier for sure!

AudriusButkevicius:
To be honest, I am now lost on what we are doing here. I proposed to merge the lists and do hacky parsin, but @calmh says that not having visibility of your advanced patterns affecting the tree is bad, at which point I can't see a solution to this anymore, as advanced patterns with double star, or extension whitelist means that the whole tree has to be loaded ahead of time, which does not scale.

AudriusButkevicius:
@calmh what we are we actually doing here? What do you expect this to look like.

calmh:
I expect there to be two ways to build your ignore patterns. One way is clickety click in the tree. If that's not good enough you can switch to advanced mode. Advanced mode is essentially what we have today: a text box. No tree.

When switching from tree to advanced we can trivially populate the text box. When switching from advanced to tree we can either do best effort conversion or just drop all the existing patterns, both are fine with me.

I do **not** want there to be *both* a tree and some other patterns, where you might select "foo" in the tree and put "!foo" in the advanced box and the result is 🤷‍♀️

AudriusButkevicius:
Why is it not ok not to have feedback from advanced patterns in the tree? Just because both end up in the same file (we could hide tree generated content in the edit box, or split it into a separate file)?

I think you will most of the time use both anyway, and I don't think people will expect changes to advanced section to affect the tree.

calmh:
OK I get the thing about wanting to ignore file patterns and extensions. What I don't like is the ambiguity of having the tree control say one thing and the advanced patterns say something else. Suddenly you need to know our implementation to understand what's going to happen. Prohibiting conflicts isn't trivial either.

Let's say checkbox means ignore.

```
--Tree--

[ ] /
    [X] foo
    [ ] bar

--Advanced--

![fb][oa][or]
```

It's schrödingers ignore pattern.

Tstassin:
I see three main reasons to use ignore patterns :

1. selective sync of folders (Dropbox like)
2. cluster filtering of weird/temp/special **common** files/folders (`.DS_Store`, `@eaDir`,...)
3. cluster filtering of special and specific files/folders (`temp`, `fullRes`)

This PR will provide us a nice interface to do `1.`, but I still want to combine this with my `2.` and `3.` settings.

I would go with the last suggestion of @calmh : first take account of selected folders from the tree (_reducing the set of folders to process_), then filter **this selection** with the advanced patterns.
Maybe include checkboxes for common ignored filetypes (`2.`).
A strong comment and UX design will reflect the way ignores are processed : 

selective folder sync graphical interface `(1.)`
**then** (_optional_) ignore common filetype checkbox(es) `(2.)`
**then** (_optional_)  filter the subset with advanced ignore patterns `(3.)`

By activating "advanced ignore patterns" you get a modal explaining how they work (order of processing) with 1. and 2. and an "debug at your own risk" text ?

AudriusButkevicius:
@calmh

> Or are you saying the tree checkboxes take effect first, and any extra patterns just affect the stuff that was included in that pass? That could also be reasonable. But then we're not just talking GUI but actual ignore mechanism tweaks.

This is what I am talking about. And I am not suggesting we show advanced patterns in the tree either. They are advanced, and people adding them must know what they are doing and read the warning above that changes in the advanced mode are not reflected in the tree amd get on with their lives.

grahamPegNetwork:
Awesome discussion here @calmh @AudriusButkevicius and everybody else. 
My two cents:
The core of the feature I see coming is on par with other sync tools: Selective sync. The fact that we are talking about using the ignore system to handle it probably doesn't change the end result. As a user I want to choose a folder and select which folders get synced to my device, that's about it. 
I'm also a fan of the comments about advanced rules which power users can use to filter out junk like thumbs.db and would love to see that as a supplement to the basic/tree view selection.

Looking forward to using these features! I have syncthing running on several servers/computers/phones and would consider building from source to take advantage of this feature when it's ready for more general testing.

Thanks for your contributions!

AudriusButkevicius:
@calmh can we put this to bed once and for all?

calmh:
I'm not sure what that means, practically. As is, this PR does not implement the proposal above. (I think there is also some technical issue with it; when I tried it now, saving the ignore patterns from the tree editor leaves the UI greyed out (modal background) and the patterns unsaved.)

AudriusButkevicius:
That means of me convincing you to give up the idea for advanced patterns needing to affect the tree state.

uok:
I think having a tree mode that allows users to select which files they want to include (or exclude) is a big improvement for Syncthing GUI. A separate section are advanced rules for power users.
Other ideas: 
- a toggle that switches between simple (tree) and advanced mode, so it is clear to user which is active
- a button that generates advanced rules from tree view so users can redefine them manually in advanced mode

grahamPegNetwork:
> @calmh can we put this to bed once and for all?

Another couple cents from me on this topic. IMO it doesn't have to be perfect, just well thought out. If I was roughly quoting from "Getting Real" (the group behind basecamp) a lot of software dev is about launching something that works fast, getting input from the community on it, and improving on it as needed. Maybe people want this feature to manage ignores, maybe others want it to manage selective sync. (I'm one of the latter)

Most of all, details aside, I'd love to see this feature period, in any form. If we can get it rolled out to my phone/laptop/servers we can get more feedback, more issues, maybe even complaints.

If this PR is providing even half of the functionality users expect and isn't breaking anything major, ship out the best you can as fast as possible and improve from there. Don't get me wrong, there's good discussion here, but this PR was also created nearly 3 months ago. I totally appreciate the volunteer time put into this project and the discussion here and can't say much as a non-contributor...but it seems like this is or will be at a "good enough" state to be pushed out. If there are issues they shouldn't break anything from the code I've seen, and they can always be improved on. Heck, I'll probably make some minor ones in the future as well.

Apologies if anything sounded whiny/demanding. You're all doing a great job, I'd totally love to see it actually go out though.

AudriusButkevicius:
Well the PR (or attempts to, as there are still rough corners) implements a tree, which is what the users want. What it doesn't do, is what you deem necessery, is reflect advanced patterns in the tree. Yet users agree that advanced patterns are advanced and not having them reflect in the tree is ok. If you agree to that, then it's somewhat clear what the implementation should look like.

calmh:
I tried to run the PR now, but I didn't get a tree, just got the usual text box and a lot of JS errors, so can't give an opinion on what is currently implemented.

However, I don't expect advanced patterns to be reflected in the tree. I was suggesting the exact opposite - that one either use the tree *or* advanced patterns. With some mix of the two the results are bound to be confusing.

But -- since the impression is that I'm blocking this, and I don't really have an idea of what the PR proposes exactly, and I've obviously also been a bit misunderstood, please do the thing without me. :) There are enough opinions here anyway and I'm sure the result will be reasonable and an improvement on the current situation regardless. If it turns out confusing for whatever reason, we can iterate.

ibizaman:
Thanks all for your feedback. Just to be clear, this PR is in WIP. I was advised rightfully to create a PR even if not everything is working as it's better to had feedback sooner. That being said @calmh you should see a tree so there's probably a bug somewhere when I updated to latest master.

There are a lot of great ideas and I'd like to recap the various options. I see a few big themes.

# Options
## One vs Two tabs
### Two tabs - restrictive
The PR's original goal, as it is in the commits, is to provide two tabs: a "basic" tree view and an "advanced" text view where the user can write any rule.

Behavior:
When switching from tree view to text tab, the text would be replaced by what has been selected in the tree view.
When switching in the opposite direction, the text is parsed and the tree view populated as stated in the text rules.

The tree view would only be displayed if the text view contains simple rules that are parseable in the tree view. If not, it's disabled.

### Two tabs - dedicated section
Very similar to the above `Two tabs - restrictive`, the difference is that there's a dedicated section in the text rules that are used by the the tree view. The tree view would only populate and parse that section. The section would be delimited by comments.

Two variants were proposed:
- we show the tree view section of the text rules in the UI
- we don't
In either case, the generated .stignore will have that section.

### Combined - only tree
We do not show the advanced section at all and the user can only select files/folders to ignore through the treeview.

### Combined - tree and advanced
We show the tree view and the advanced section underneath.

Variants:
- the advanced section is a textbox like currently
- the advanced section is a dynamic list of inputs

## Behavior of tree view
Checking a file/folder can either mean include or exclude it or both with a tristate.

## User feedback of advanced rules
### No feedback
The interaction between tree view selections and advanced rules is not shown anywhere and the user needs to know what's going on.

### Feedback in the tree
??? There are some good ideas here but not concrete enough IMO to be valid options to consider without more refinement.

## Default rules
How do we show default rules? Do we show default rules?
They are not shown currently.

# Conclusion?
If I read this thread correctly, we're tending to agree on the following for v1:
- `One vs Two tabs`? `Two tabs - dedicated section`
- `Behavior of tree view`? `Exclude`
- `User feedback of advanced rules?` `No feedback`
- `Default rules?` `Not shown`

AudriusButkevicius:
Just to clarify, the blockiness comes from this sentence: 

> that one either use the tree or advanced patterns. With some mix of the two the results are bound to be confusing.

There is no way not to make it not confusing without sacrificing existing functionality that people expect to have. So we just have to accept that advanced cases (picking something in the tree and igoring jpegs) will be confusing (or not obvious), and there is no reasonable way around it. We'll plaster warnings in the ui.

AudriusButkevicius:
I think the restrictive mode makes no sense for the same reason as explained in my previous comment. 

I think the tree should have two modes, namely it should have an option to act as an include or as an exclude, because users will either want to sync that one folder only, or all but that one folder.

Advanced patterns should also have two modes, mode deciding if they will be applied before tree or after the tree.

I think everyone agrees that it should be two tabs with warnings plastered in the advanced view explaining that things there will not show up in the tree and might have undesired effects.

I think we leave out default rules all together as it's a separate problem.

ibizaman:
> restrictive mode makes no sense

Agreed, was just putting in there for completion.



> Advanced patterns should also have two modes, mode deciding if they will be applied before tree or after the tree.

Maybe split the advanced textbox in two?

> I think the tree should have two modes, namely it should have an option to act as an include or as an exclude

I think we can do with only one mode, specifically with the following convention in the treeview:
- a checked checkbox means include every children and new children under node, unless some children are explicitly unchecked
- an unchecked checkbox means ignore every children and new children under node, unless some children are explicitly checked

I'm referencing https://github.com/syncthing/syncthing/pull/5132#issuecomment-415704821 as @canton7 gave good examples. I'm doing similar examples with only two modes:

`[x] for include`, `[ ] for exclude`

By default, there are no advanced rules so everything is included:
```
[x] Root
   [x] A
      [x] A.1
         [x] A.1.a
         [x] A.1.b
      [x] A.2
         [x] A.2.a
         [x] A.2.b
```

Unchecking A, you get:
```
[x] Root
   [ ] A
      [ ] A.1
         [ ] A.1.a
         [ ] A.1.b
      [ ] A.2
         [ ] A.2.a
         [ ] A.2.b
```
Behavior here is if a new folder `B` is added as a sibling of `A` it gets included. If a folder `A.3` is added as a child of `A` it is ignored.

Checking A.1 and A.2.a yields:
```
[x] Root
   [ ] A
      [x] A.1
         [x] A.1.a
         [x] A.1.b
      [ ] A.2
         [x] A.2.a
         [ ] A.2.b
```
Behavior here is same as above, if `A.1.c` is added as a child of `A.1` it is now included. If `A.2.c` is added as a child of `A.2` it is still ignored like above.

The "other mode" would be is the user unchecks the root node, then everything is ignored by default:
```
[ ] Root
   [ ] A
      [ ] A.1
         [ ] A.1.a
         [ ] A.1.b
      [ ] A.2
         [ ] A.2.a
         [ ] A.2.b
```
Behavior here is any new child of root is ignored by default.
For reference this would produce a `.stignore` containing only `*`.

One can then only sync `A.1` by checking it:
```
[ ] Root
   [ ] A
      [x] A.1
         [x] A.1.a
         [x] A.1.b
      [ ] A.2
         [ ] A.2.a
         [ ] A.2.b
```
Behavior here is if a new folder `B` is added as a sibling of `A` it gets ignored. If a folder `A.1.c` is added as a child of `A.1` it is included.

grahamPegNetwork:
> I think the restrictive mode makes no sense for the same reason as explained in my previous comment.
> 
> I think the tree should have two modes, namely it should have an option to act as an include or as an exclude, because users will either want to sync that one folder only, or all but that one folder. (Infer that from the last line being a * I guess?)
> 
> Advanced patterns should also have two modes, mode deciding if they will be applied before tree or after the tree.
> 
> I think everyone agrees that it should be two tabs with warnings plastered in the advanced view explaining that things there will not show up in the tree and might have undesired effects.
> 
> I think we leave out default rules all together as it's a separate problem.

This all sounds good. I agree on the last part, default exclude rules do belong on their own issue. It appears that #837 already begins to address this.

AudriusButkevicius:
> Maybe split the advanced textbox in two? With the option to "fold" them to avoid cluttering the UI?

Not sure what fold means. In essence all we need is to decide if they get processed before or after.

> Single mode tree

I agree that what you've proposed works, yet the fact you had to write a paragraph to explain it raises concerns of the obviousness of how it works. Perhaps we would just need extensive documentation. Also there will be confusion as to what is actually selected if the selections propagate automatically, I think once I select a folder, all folders underneath that should have disabled checkboxes. Also, I suspect there will be cases where a user wants to sync all files in a folder but not subdirectories, but perhaps we don't care about that on the first iteration.

grahamPegNetwork:
In my experience with other selective sync software, checking or unchecking a folder would also recursively perform the same operation on every file/folder below it by default which seems most intuitive. If you then wish to update sub-folder rules you can manually. I believe ownCloud/nextCloud/Seafile use a model similar to this though I haven't touched them in ages.

ibizaman:
> [...], yet the fact you had to write a paragraph to explain it raises concerns of the obviousness of how it works. Perhaps we would just need extensive documentation.

@AudriusButkevicius I get your point but I don't think that's fair 😛 here I was being extensive to show that we can have both include and exclude modes with a simple bi-state checkbox. The trick being having a checkbox on the root node.

Like you propose, we could and should IMO show examples in the documentation but in the UI it could be enough to say something like:

"A file or folder is included in the sync if it is checked and excluded from the sync if it is not. Every new file or folder will inherit their checked state from their parent folder."

> Also there will be confusion as to what is actually selected if the selections propagate automatically, I think once I select a folder, all folders underneath that should have disabled checkboxes.

I actually wouldn't do this, I would go with what @grahamPegNetwork proposes:

> checking or unchecking a folder would also recursively perform the same operation on every file/folder below it

This way you can selectively sync subfolders.

> Also, I suspect there will be cases where a user wants to sync all files in a folder but not subdirectories, but perhaps we don't care about that on the first iteration.

I agree we shouldn't do this for v1 as anyway one can still do this through advanced rules anyway.

AudriusButkevicius:
> I actually wouldn't do this, I would go with what @grahamPegNetwork proposes:

But this needs fairly special handling as you don't want to add a new pattern line for every subfolder, so you have to recursively check that every child is all checked. 

How would these be represented as patterns?
1. Root is checked, some deep folder is unchecked
2. Root is unchecked, some deep folder is unchecked
3. Root checked, some deel folder is unchecked, some even deeper folder is checked again

I am starting to think that it might be hellish to parse the patterns to work out the state of the tree, also pattern order is important.

grahamPegNetwork:
Good day all, how are we looking for the merge? I have some friends interested in syncthing, wondering how far we are from having this sort of feature rolled out. Hope everybody had a good Christmas/New Year!
Also tagging #2491 for easier reference to this issue.

darkermatter:
This is how it's done in Carbon Copy Cloner. Considering the software has been around since ~~the beginning of time~~ 2002, I imagine [Mike Bombich](https://bombich.com) has put the interface through lot's of iterations.


image

ibizaman:
Hey, sorry for not answering for so long.

On my side, the biggest reason this PR is stalling is lack of time to invest in it. There's a bunch of things going on in my life right now that takes all my time.

I'm actually not afraid of the parsing bit. For having thought about that initially when I created the PR, there are indeed a few cases that needs to be taken care of but I think we can make it work. Also from a user perspective, I think we can make it intuitive and as others point it out, other software did it already.

I also want to reassure core developers of Syncthing that I think I can make the code this PR adds readable and maintainable, as big as it needs to get. I know there are some quirks right now but it's by no means definitive.

What misses the most though is javascript unit tests. I would have a much easier time developing the parser with those.

I think the best first step for this PR is to not think about the parsing. I'd like to show what I want as a UI first. So people can checkout the PR and we can agree on that as it's IMO the most important thing to nail down correctly.

Then I can switch to the second part which will be the parsing.

Unfortunately, as I said in the beginning I don't have time for it right now. It's in my top priority though, if that can help on the wait.

ismay:
I can only speak for myself of course, but thanks for the update and take all the time you need! Didn't want to rush you or anything, was just curious about the current state 😊

st-review:
:robot: *beep boop*

breadteam:
Hey folks - was there any progress on implementing a feature like this? I'm rather new to syncthing and I'm certainly not any kind of developer. I am, however, someone who would really, really like to see this functionality built into Syncthing.

Is there another pull request for this functionality that was alive during 2019?

totti4ever:
> Hey folks - was there any progress on implementing a feature like this? I'm rather new to syncthing and I'm certainly not any kind of developer. I am, however, someone who would really, really like to see this functionality built into Syncthing.
> 
> Is there another pull request for this functionality that was alive during 2019?

I second that question :-)
It would just be very helpful for using syncthing at my wife's phone to sync our common pictures

imsodin:
If there are no updates here on in the issue it addresses (https://github.com/syncthing/syncthing/issues/2491) then there has not been any progress - such questions are thus pointless and just use up time of people involved in Syncthing.

PR #5137, opened 2018-08-20
lib/versioner: Trash can versioner should tag files

AudriusButkevicius:
Fixes something


calmh:
So, eh, no it shouldn't? This is counter to the point of the trashcan versioner.

Issue #5142, opened 2018-08-24
"Support bundle" download


calmh:
Some other products have a concept of a "support bundle" that can be easily created/downloaded by the user, and which contains "all" useful information a support person might need to debug what's going on. In the Syncthing case, this could take the form of an API method (and menu option in the GUI) that results in a zip file download. In this zip file we would have

- A text file with basic build info (version / platform)
- The current configuration
- The last few log lines, whatever we have cached in memory
- Any panic files we can find in the config directory
- The contents of the usage reporting data (this would include things like the benchmark and folder sizes which is good to know)
- Maybe, a heap and CPU profile (just a couple of seconds for CPU in that case)
- …?

PR #5145, opened 2018-08-25
"Support bundle" download #5142

bahadir60:
### Purpose
To add a button, "SupportBundle", on GUI that downloads some info for a support people. Issue number #5142 
### Testing
None.

### Screenshots
SupportBundle button added under Actions section.

button_supportbundle

Clicking SupportBundle button makes two things. First one is open a window to saveas zip file and the second is save same zip file under the Syncthing base folder under the created support_bundle folder.

save_window

Zip file named as support_bundle + timestamp
saved_zip

Under the saved zip there are 4 files. (The files will be added for further needs)
saved_files

## Authorship
git config --global user.name "Bahadir Yilmaz"
git config --global user.email "bahadiryilmaz@yahoo.com"


lib/logger/logger.go
```diff
@@ -339,6 +340,25 @@ func (r *recorder) Since(t time.Time) []Line {
 	return nil
 }
 
+func (r *recorder) All() []Line {
```
calmh: I think this method is not really necessary as it's equivalent to `Since(time.Time{})` (i.e., everything since the zero time).

(Removing it here and from the logger interface will also make the build pass, as it currently fails due to the test version of the recorder not fulfilling the updated interface.)

cmd/syncthing/gui.go
```diff
@@ -1023,6 +1026,45 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *apiService) getSupportBundle(w http.ResponseWriter, r *http.Request) {
```
calmh: I think that in this method we can probably avoid creating all the temp files and instead buffer them in memory as byte slices or `bytes.Buffer` objects. The zip creator can use `bytes.Buffer`s to read from.

That also simplifies error handling and avoids issues with not being able to create temp files and whatnot.

cmd/syncthing/support_bundle.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"archive/zip"
+	"io"
+	"log"
+	"os"
+)
+
+// ZipFiles compresses one or many files into a single zip archive file
+func ZipFiles(filename *os.File, files []string) error {
```
calmh: That is, if this took a slice of `struct { Name string, Data io.Reader }` then the Data can be a `bytes.Buffer` and the rest of the zip code is roughly the same, minus open/close stuff.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1047,6 +1047,96 @@ angular.module('syncthing.core')
             return deviceCfg.deviceID.substr(0, 6);
         };
 
+        $scope.downloadFile = function (httpPath) {
```
calmh: I'm not sure what's going on here because I didn't read it. :) Can we not serve the zip file with the appropriate `Content-Disposition` header and let the browser be responsible for saving the file under the correct name etc?

cmd/syncthing/support_bundle.go
AudriusButkevicius: Also, I think you could zip to a io.Writer (http.ResponseWriter) avoiding creating a files on disk.

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: Yeah this is an overkill, you just need `<a href="/system/support" target="_blank>`

cmd/syncthing/gui.go
```diff
@@ -1023,6 +1026,45 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	createJSONFile("config.json", jsonConfig)
+
+	jsonLog, _ := json.MarshalIndent(s.systemLog.All(), "", "	")
+	createJSONFile("log.json", jsonLog)
+
+	jsonError, _ := json.MarshalIndent(s.guiErrors.All(), "", "	")
+	createJSONFile("error.json", jsonError)
+
+	heapRuntime := fmt.Sprintf("syncthing-heap-%s-%s-%s-%s.pprof", runtime.GOOS, runtime.GOARCH, Version, time.Now().Format("150405")) // hhmmss
+	createJSONFile("heapRuntime.txt", []byte(heapRuntime))
```
AudriusButkevicius: Not sure what this is. The ticket talks about capturing a heap profile from the runtime, explaining where the memory is allocated. There is probably some way to get a dump summary of the heap too.

uok:
This looks like a good feature for easier support. My suggestion would be to call menu item "Debug" or "Diagnostics" and the output file "debug_syncthing.....zip"

bahadir60:
@uok thanks your comment, i get the name from the ticket, may be your suggestion is more suitable, however support_bundle is also common name.

gui/default/syncthing/core/syncthingController.js
bahadir60: @calmh @AudriusButkevicius 
I made some changes on code to get Content-Disposition header appropriately. Apart from that, I tried to cut off the unnecessary parts. Can you review my code?

        $scope.getSupportBundle = function (){
            $http.get(urlbase + '/system/support', { responseType: 'arraybuffer' })
             .success( function (data, status, headers) {
                var header = headers()["content-disposition"]
                var filename = header.split(';')[1].trim().split('=')[1];
                filename.replace(/"/g, '');
                var contentType = headers['content-type']
                // Get the blob url
                    var url = window.URL || window.webkitURL || window.mozURL || window.msURL;
                    if(url)
                    {
                        var a = document.createElement('a');
                        var blob = new Blob([data], { type: contentType });
                        var url = url.createObjectURL(blob);
                        a.setAttribute('href', url);
                        a.setAttribute("download", filename);
                        var event = document.createEvent('MouseEvents');
                        event.initMouseEvent('click', true, true, window, 1, 0, 0, 0, 0, false, false, false, false, 0, null);
                        a.dispatchEvent(event);
                    };
             });
            };

bahadir60: @AudriusButkevicius thx for your quick reply. I try to use <a href="/system/support" target="_blank> but i get '404 page not found.' error. 
code

AudriusButkevicius: Because the url is prefixed with /rest/ etc, I just gave this as an ideological example.

bahadir60: @AudriusButkevicius sorry to misunderstanding 'literally' word. I'll try it.

AudriusButkevicius: Yeah, my bad, literally meant "all you need to do is just tell the browser to open the url".

bahadir60:
I made some changes base on comments and ticket issues.

It can be max 8 files in the Zip. GUI error shown in picture as "error.json" and "panic*.log" files are optional, if they are generated by the system, included in zip.

zipinclude

**Ticket issues:**
_-A text file with basic build info (version / platform):_  **As a json file can be change to text file**
_- The current configuration:_ **Json file. Apikey, User and Password set redacted.** 
configredacted

_-The last few log lines, whatever we have cached in memory:_ **Json file.**
_-Any panic files we can find in the config directory_:  **panic-timestamp.log**
_-The contents of the usage reporting data (this would include things like the benchmark and folder sizes which is good to know)_: **I am not sure my code covers all this. I am waiting comments on this issue.**
_-Maybe, a heap and CPU profile (just a couple of seconds for CPU in that case)_: **CPU profile duration set to 4 seconds. It can be change base on comments.**

-_Thinking to added some codes ASAP:_ Besides the clicking the Support Bundle button opens a new window to save the zipped file to the desired location, saves the relevant file to the local computer root folder. However it is not good mind to save all files in back up. So **I will implement code that makes not to store over 10 zip file in local computer.(fifo)**
(@AudriusButkevicius i am trying to write "href" for button. But not achieved yet.)

Nutomic:
"Support Bundle" sounds to me like it contains help documents or a manual. I think "Debug Bundle" is definitely clearer.

AudriusButkevicius:
Just a debug dump?

bahadir60:
Debug or debugging words may confuse to `GUI.Debugging`. `Support` gives me attention that it is for support personnel, instead of client. Also, there are many examples to naming `Support Bundle` (or `Support Bundles` if it is selectable ) same as what we are doing here.

EserGozcu:
@bahadir60  it does not have to be necessarily for the support team. I agree that Debug Bundle is a little bit confusing but perhaps we can find something more relevant.  Checking various applications right now: I see JIRA uses as "Create Support zip", another product we use have "Support Bundle", VMware uses "Export Support Bundle", Jenkins says "support-bundle" . Maybe it is better to stick to something close to those ones?  @AudriusButkevicius  ?

AudriusButkevicius:
Sure, if that is what is used elsewhere. I don't like the word bundle, as it sounds like something you'd buy as a service (support service bundle for your aws account or something like that).

cmd/syncthing/gui.go
```diff
@@ -1023,6 +1026,81 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	buf  bytes.Buffer
+}
+
+func (s *apiService) getSupportBundle(w http.ResponseWriter, r *http.Request) {
+	var arrayFile []fileByte
+
+	jsonConfig, _ := json.MarshalIndent(getRedactedConfig(s), "", "  ")
+	arrayFile = addBufferFile(arrayFile, "config.json", jsonConfig)
+
+	jsonLog, _ := json.MarshalIndent(s.systemLog.Since(time.Time{}), "", "  ")
```
calmh: I don't think we should drop in the log as JSON; a simple text format would be better and more readable. Basically, for each entry in the returned log lines, `fmt.Fprintf(buf, "%s: %s\n", entry.When, entry.Message)` (roughly) to just print it to a buffer and add that as the `log.txt` or something.

cmd/syncthing/gui.go
```diff
@@ -1023,6 +1026,81 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+
+	jsonError, _ := json.MarshalIndent(s.guiErrors.Since(time.Time{}), "", "  ")
+	if string(jsonError) != "null" {
+		arrayFile = addBufferFile(arrayFile, "error.json", jsonError)
+	}
+
+	panicFiles, ok := checkPanicFile(baseDirs["config"])
+	if ok {
+		for _, f := range panicFiles {
+			panicF, _ := ioutil.ReadFile(f)
```
calmh: Don't skip error handling here. If we get an error back the `panicF` will be null and bad things might happen. Just `if err != nil { continue }` would do.

cmd/syncthing/gui.go
```diff
@@ -1023,6 +1026,81 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	var arrayFile []fileByte
+
+	jsonConfig, _ := json.MarshalIndent(getRedactedConfig(s), "", "  ")
+	arrayFile = addBufferFile(arrayFile, "config.json", jsonConfig)
+
+	jsonLog, _ := json.MarshalIndent(s.systemLog.Since(time.Time{}), "", "  ")
+	arrayFile = addBufferFile(arrayFile, "log.json", jsonLog)
+
+	jsonError, _ := json.MarshalIndent(s.guiErrors.Since(time.Time{}), "", "  ")
+	if string(jsonError) != "null" {
```
calmh: I'd suggest

```
if errs := s.guiErrors.Since(time.Time{}); len(errs) > 0 {
  // do marshalling etc
}
```

instead of comparing to the "null" string here.

cmd/syncthing/gui.go
```diff
@@ -1023,6 +1026,81 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	usageReportingData, _ := json.MarshalIndent(getusageReportingData(s, r), "", "  ")
+	arrayFile = addBufferFile(arrayFile, "usageReportingData.json", usageReportingData)
+
+	rawConf := s.cfg.RawCopy()
+	debugCurrent := rawConf.GUI.Debugging
+	if !debugCurrent {
+		debugUndebug(s, rawConf, true)
+	}
+
+	heapPprof := getPprof(s, "/rest/debug/heapprof")
```
calmh: I wouldn't grab these from the API. Instead, look at what the corresponding API methods do and just copy the code. It's just one to three lines each to create the profiles directly into a buffer, avoiding the API calls and undebug etc. stuff.

cmd/syncthing/support_bundle.go
```diff
@@ -0,0 +1,143 @@
...
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/syncthing/syncthing/lib/config"
+)
+
+// AddBufferFile adds new json to fileByte(name, bytes.Buffer)
+func addBufferFile(arrayFile []fileByte, name string, json []byte) []fileByte {
```
calmh: Instead of using this function, it's more idiomatic Go to just

```
arrayFile = append(arrayFile, fileEntry{name: name, data: data})
```

where `fileEntry` is something like

```
type fileEntry struct {
  name string
  data []byte
}
```

I don't think keeping a buffer directly in the `fileByte` (as it's called now) here adds much value. Instead, the buffer handling can be done as necessary by the creator / user of these structs.

cmd/syncthing/support_bundle.go
```diff
@@ -0,0 +1,143 @@
...
+	singleFile.buf.Write(json)
+	arrayFile = append(arrayFile, singleFile)
+	return arrayFile
+}
+
+// GetRedactedConfig redacting some parts of config
+func getRedactedConfig(s *apiService) config.Configuration {
+	rawConf := s.cfg.RawCopy()
+	rawConf.GUI.APIKey = "REDACTED"
+	rawConf.GUI.Password = "REDACTED"
```
calmh: Only do this if the User/Password value are set, to preserve the fact that no authentication is set when it isn't.

```
guiCfg := s.cfg.GUI()
if guiCfg.Password != "" {
   guiCfg.Password = "REDACTED"
}
...
```

cmd/syncthing/support_bundle.go
```diff
@@ -0,0 +1,143 @@
...
+	rawConf.GUI.User = "REDACTED"
+	return rawConf
+}
+
+// GetusageReportingData collects needed data for support personnel
+func getusageReportingData(s *apiService, r *http.Request) map[string]interface{} {
```
calmh: What does this do that we can't just use the return value of `reportData()` directly?

bahadir60: Thanks ur comments. I will implement them.

bahadir60:
@calmh I made some changes based on your comments. Also I added `EraseSupportBundleZip()` function that cleans old zip files over 10 in the local computer back-up. But, cleaning can be change to based on time and added to `main.go - cleanConfigDirectory()` function.

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,96 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	arrayFile = append(arrayFile, fileEntry{name: "log.txt", data: buff.Bytes()})
+
+	// Errors as a JSON
+	if errs := s.guiErrors.Since(time.Time{}); len(errs) > 0 {
+		jsonError, _ := json.MarshalIndent(errs, "", "  ")
+		arrayFile = append(arrayFile, fileEntry{name: "errors.json", data: jsonError})
+	}
+
+	// Errors as a JSON
+	panicFiles, ok := checkPanicFile(baseDirs["config"])
```
calmh: Comment here is cut/paste from above. Also I would suggest skipping the comma-ok return from checkPanicFile. In fact, I'd skip the entire function and

```
if files, err := filepath.Glob(filepath.Join(basedirs["config"], "panic*")); err == nil {
  for _, f := range files {
    ...
```

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,96 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+		"codename":    Codename,
+		"longVersion": LongVersion,
+		"os":          runtime.GOOS,
+		"arch":        runtime.GOARCH,
+	}, "", "  ")
+	arrayFile = append(arrayFile, fileEntry{name: "versionPlatform.json", data: versionPlatform})
+
+	// Report Data as a JSON
+	version := usageReportVersion
+	if val, _ := strconv.Atoi(r.URL.Query().Get("version")); val > 0 {
```
calmh: I think skip this. On the regular usage reporting endpoint we need to be able to select usage reporting version, but here's that's not relevant. You can just pass the `usageReportVersion` constant directly to `reportData` to get the latest version.

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,96 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	usageReportingData, _ := json.MarshalIndent(reportData(s.cfg, s.model, s.connectionsService, version, true), "", "  ")
+	arrayFile = append(arrayFile, fileEntry{name: "usageReportingData.json", data: usageReportingData})
+
+	// Heap and CPU Proofs as a pprof extension
+	var heapBuffer, cpuBuffer bytes.Buffer
+	filename := fmt.Sprintf("syncthing-heap-%s-%s-%s-%s.pprof", runtime.GOOS, runtime.GOARCH, Version, time.Now().Format("150405")) // hhmmss
+	runtime.GC()
+	pprof.WriteHeapProfile(&heapBuffer)
```
calmh: 👍

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,96 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	arrayFile = append(arrayFile, fileEntry{name: filename, data: heapBuffer.Bytes()})
+
+	const duration = 4 * time.Second
+	filename = fmt.Sprintf("syncthing-cpu-%s-%s-%s-%s.pprof", runtime.GOOS, runtime.GOARCH, Version, time.Now().Format("150405")) // hhmmss
+	pprof.StartCPUProfile(&cpuBuffer)
+	time.Sleep(duration)
+	pprof.StopCPUProfile()
+	arrayFile = append(arrayFile, fileEntry{name: filename, data: cpuBuffer.Bytes()})
+
+	path := filepath.Join(baseDirs["config"], "support_bundle")
```
calmh: I think having these directly in the config dir is more in line with how we do everything else. Call them `support-bundle-2006-01-02T15.04.05.zip` and we're good.

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,96 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	os.MkdirAll(path, 0700)
+
+	output := time.Now().Format("2006-01-02T15.04.05") + ".zip"
+	newfilepath := filepath.Join(path, output)
+	newfile, _ := os.Create(newfilepath)
```
calmh: Must check error.

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,96 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	defer newfile.Close()
+
+	// Add files to zip
+	zipBuffer(newfile, arrayFile)
+
+	// Delete old zips over 10 (fifo)
+	eraseSupportBundleZip(path)
+
+	from, _ := os.Open(newfilepath)
```
calmh: You're doing something odd here, I think maybe serving the file twice. `http.ServeFile` serves directly from the filesystem so no need to open first. (You do however need to close the writing side, above.) Alternately we could do this in memory - have `zipBuffer` zip *into* another buffer, write those to file and also write the same bytes to the request.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1047,6 +1047,28 @@ angular.module('syncthing.core')
             return deviceCfg.deviceID.substr(0, 6);
         };
 
+        $scope.getSupportBundle = function (){
+            $http.get(urlbase + '/system/support', {responseType: 'arraybuffer'})
```
calmh: I think this thing is not required at all. A simple `<a href="...">` pointing directly to the `/system/support` endpoint should cause the browser to save the file, as you are setting the content-disposition header server-side. No javascript required. :)

bahadir60: Because of using `<a href="...">` and access without a CSRF token, i changed the path `/rest/system/support` to '/rest/debug/support`. Like `cpuprof` and `heapprof`, I think `support bundle` should work when`debugging enabled`. @calmh ? @AudriusButkevicius ?

cmd/syncthing/gui.go
bahadir60: @calmh Thanks your comment. Yes, I recognized that the file served twice. File opening and `io.Copy(w, from)` part is unnecessary. But, sorry that I didn't understand `(You do however need to close the writing side, above.)` part of ur comment. I appreciate if u explain it.

calmh: If you write the zip file to an in memory bytes.Buffer this is a non-issue. But if you, like now, write it to an actual file on disk and then read it back, you have the problem that until the writing side has `Close()`d the file, all contents might not be written and hence cannot be read yet. For zip files this is definitely the case because the close call is what writes the directory at the end of the file. Before calling close, the zip file is not a zip file.

gui/default/syncthing/core/syncthingController.js
calmh: I think not. Even with redacted passwords the support bundle contain too much private information to serve without auth. The existing debug stuff are just profiles, which don't contain anything sensitive.

I think linking directly should work anyway, because we use HTTP Basic auth for the login, with a session cookie, which the browser should supply by itself.

cmd/syncthing/gui.go
bahadir60: Ohh I got it, thanks for explanation.

gui/default/syncthing/core/syncthingController.js
bahadir60: I'm not sure understand u precisely. If I'm wrong please correct me. 
`handler` first goes to `csrfMiddleware(...)` function that use `/rest` as a prefix. In function we have 3 options. If begin with prefix `/rest` needs a api key, or if `rest/debug` needs neither a api key or a CSRF token. The third option is not begin with prefix `/rest` which needs a CSRF token, while don't have CSRF token http.Cookie() get a new one. 

So;
1. if `Support Bundle` is begin with prefix '/rest' and not '/rest/debug' we should serve a api key.
2. if begin with '/rest/debug' we should use debugging or new handleFunc like `getRestMux.HandleFunc("/rest/debug/support", s.getSupportBundle)` which don't need apikey, CSRF or debugging. **It is not a option as we agree.**
3. if **not** begin with prefix `/rest` needs a CSRF token. This time adding to `mux.Handle(...)` or `mux.HandleFunc(..)` like `s.getQR` is enough.

calmh: OK OK, I was confused, sorry. The endpoints under `/rest/debug` require authentication as much as anything else, but indeed don't have the CSRF requirement. So yes, you are right, the best place for this is under `/rest/debug/support.zip` or something like that, and will require enabling debugging to get. I think that's an acceptable extra step.

bahadir60: Thanks, I will push my code ASAP.

bahadir60:
Base on @calmh  and @AudriusButkevicius' comments, I made some changes. Thanks to them.

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1028,93 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	if errs := s.guiErrors.Since(time.Time{}); len(errs) > 0 {
+		jsonError, _ := json.MarshalIndent(errs, "", "  ")
+		arrayFile = append(arrayFile, fileEntry{name: "errors.json", data: jsonError})
+	}
+
+	// Errors as a JSON
+	if panicFiles, err := filepath.Glob(filepath.Join(baseDirs["config"], "panic*")); err != nil {
+		for _, f := range panicFiles {
+			panicF, err := ioutil.ReadFile(f)
+			if err != nil {
```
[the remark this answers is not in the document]
bahadir60: It is my bad, thanks

gui/default/index.html
```diff
@@ -81,6 +81,8 @@
             <li class="divider" aria-hidden="true"></li>
             <li><a href="" ng-click="advanced()"><span class="fas fa-fw fa-cogs"></span>&nbsp;<span translate>Advanced</span></a></li>
             <li><a href="" ng-click="logging.show()"><span class="far fa-fw fa-file-alt"></span>&nbsp;<span translate>Logs</span></a></li>
+            <li class="divider" aria-hidden="true"></li>
```
calmh: One more thing. These items should only be shown when debugging is enabled, otherwise the link will anyway not work. Also not sure why the font size adjustment here?

AudriusButkevicius:
Once this is "settled" I'd like to do another round of reviews too.

calmh:
You can have the baton

gui/default/index.html
bahadir60: Font size is just for icon, may be adjustment in JS is appropriate, I will check it.

bahadir60:
if I m not wrong button needs a simple JS code hide it when debugging is not enabled. I will write it ASAP

calmh:
You can `ng-show` the tags based on some useful criteria

bahadir60:
@calmh Base on your suggestion I made changes. Thanks.

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,87 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+type fileEntry struct {
+	name string
+	data []byte
+}
+
+func (s *apiService) getSupportBundle(w http.ResponseWriter, r *http.Request) {
+	var arrayFile []fileEntry
```
AudriusButkevicius: It's not an array technically, it's a slice, so either just call it files, or perhaps use a map, given it's a name to bytes type of struct.

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,87 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+
+	// Redacted configuration as a JSON
+	jsonConfig, _ := json.MarshalIndent(getRedactedConfig(s), "", "  ")
```
AudriusButkevicius: Why not check for errors?

cmd/syncthing/support_bundle.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"archive/zip"
+	"io"
+
+	"github.com/syncthing/syncthing/lib/config"
+)
+
+// GetRedactedConfig redacting some parts of config
+func getRedactedConfig(s *apiService) config.Configuration {
```
AudriusButkevicius: The definition of what's redacted is a bit wishy washy.
For example, if I don't want to reveal the devices I want to be connected to, perhaps we should replace device IDs with mock ones, and replace the content in the logs?

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,87 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	arrayFile = append(arrayFile, fileEntry{name: "config.json", data: jsonConfig})
+
+	// Log as a text
+	var buflog bytes.Buffer
+	for _, line := range s.systemLog.Since(time.Time{}) {
+		fmt.Fprintf(&buflog, "%s: %s\n", line.When.Format(time.RFC3339), line.Message)
```
AudriusButkevicius: Again, perhaps we should do some annonimising by regexping IPs etc?

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,87 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	// Errors as a JSON
+	if errs := s.guiErrors.Since(time.Time{}); len(errs) > 0 {
+		jsonError, _ := json.MarshalIndent(errs, "", "  ")
+		arrayFile = append(arrayFile, fileEntry{name: "errors.json", data: jsonError})
+	}
+
+	// Errors as a JSON
+	if panicFiles, err := filepath.Glob(filepath.Join(baseDirs["config"], "panic*")); err == nil {
+		for _, f := range panicFiles {
+			if panicF, err := ioutil.ReadFile(f); err == nil {
```
AudriusButkevicius: what does the F signify at the end? Why not panicFile, or just fileContent or content?

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,87 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	// Log as a text
+	var buflog bytes.Buffer
+	for _, line := range s.systemLog.Since(time.Time{}) {
+		fmt.Fprintf(&buflog, "%s: %s\n", line.When.Format(time.RFC3339), line.Message)
+	}
+	arrayFile = append(arrayFile, fileEntry{name: "log.txt", data: buflog.Bytes()})
+
+	// Errors as a JSON
+	if errs := s.guiErrors.Since(time.Time{}); len(errs) > 0 {
+		jsonError, _ := json.MarshalIndent(errs, "", "  ")
```
AudriusButkevicius: again, error checks

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,87 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	if panicFiles, err := filepath.Glob(filepath.Join(baseDirs["config"], "panic*")); err == nil {
+		for _, f := range panicFiles {
+			if panicF, err := ioutil.ReadFile(f); err == nil {
+				arrayFile = append(arrayFile, fileEntry{name: filepath.Base(f), data: panicF})
+			}
+		}
+	}
+
+	// Version and platform information as a JSON
+	versionPlatform, _ := json.MarshalIndent(map[string]string{
```
AudriusButkevicius: error checks?

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,87 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+		"version":     Version,
+		"codename":    Codename,
+		"longVersion": LongVersion,
+		"os":          runtime.GOOS,
+		"arch":        runtime.GOARCH,
+	}, "", "  ")
+	arrayFile = append(arrayFile, fileEntry{name: "versionPlatform.json", data: versionPlatform})
+
+	// Report Data as a JSON
+	usageReportingData, _ := json.MarshalIndent(reportData(s.cfg, s.model, s.connectionsService, usageReportVersion, true), "", "  ")
```
AudriusButkevicius: error checks?

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,87 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+
+	const duration = 4 * time.Second
+	filename = fmt.Sprintf("syncthing-cpu-%s-%s-%s-%s.pprof", runtime.GOOS, runtime.GOARCH, Version, time.Now().Format("150405")) // hhmmss
+	pprof.StartCPUProfile(&cpuBuffer)
+	time.Sleep(duration)
+	pprof.StopCPUProfile()
+	arrayFile = append(arrayFile, fileEntry{name: filename, data: cpuBuffer.Bytes()})
+
+	// Add files to buffer zip
+	var newfileBuffer bytes.Buffer
```
AudriusButkevicius: newFileBuffer

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,87 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	zipBuffer(&newfileBuffer, arrayFile)
+
+	output := fmt.Sprintf("support-bundle-%s.zip", time.Now().Format("2006-01-02T15.04.05"))
+	newfilepath := filepath.Join(baseDirs["config"], output)
+
+	// Write buffer zip to local zip file (back up)
+	err := ioutil.WriteFile(newfilepath, newfileBuffer.Bytes(), 0500)
+	if err != nil {
+		l.Infof("Support Bundle zip could not be created locally: %s", newfilepath)
+	}
```
AudriusButkevicius: The info might as well be a warning? Also, I think we should just serve the file out of memory, whats the point of serving it off the disk from a place where nobody will ever find it?

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,87 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	pprof.StartCPUProfile(&cpuBuffer)
+	time.Sleep(duration)
+	pprof.StopCPUProfile()
+	arrayFile = append(arrayFile, fileEntry{name: filename, data: cpuBuffer.Bytes()})
+
+	// Add files to buffer zip
+	var newfileBuffer bytes.Buffer
+	zipBuffer(&newfileBuffer, arrayFile)
+
+	output := fmt.Sprintf("support-bundle-%s.zip", time.Now().Format("2006-01-02T15.04.05"))
```
AudriusButkevicius: This is not really output, this is just a file name.

cmd/syncthing/support_bundle.go
```diff
@@ -0,0 +1,50 @@
...
+	return rawConf
+}
+
+// ZipBuffer creates a zip and adds fileEntry(name, []byte]) array to zip
+func zipBuffer(writer io.Writer, files []fileEntry) {
+	zipWriter := zip.NewWriter(writer)
+	defer func() {
+		err := zipWriter.Close()
+		if err != nil {
+			l.Infoln("Zipwriter not close:", err)
```
AudriusButkevicius: This probably would return an error, instead of doing logging itself, then you only have to log once, and return upon receiving an error.

gui/default/index.html
```diff
@@ -81,6 +81,8 @@
             <li class="divider" aria-hidden="true"></li>
             <li><a href="" ng-click="advanced()"><span class="fas fa-fw fa-cogs"></span>&nbsp;<span translate>Advanced</span></a></li>
             <li><a href="" ng-click="logging.show()"><span class="far fa-fw fa-file-alt"></span>&nbsp;<span translate>Logs</span></a></li>
+            <li class="divider" aria-hidden="true" ng-show="config.gui.debugging"></li>
+            <li><a href="/rest/debug/support" target="_blank" ng-show="config.gui.debugging"><span class="fa fa-user-md"></span>&nbsp;<span translate>Support Bundle</span></li>
```
AudriusButkevicius: Out of curiosity, how does one enable debugging? I don't think it's available in the UI?

EserGozcu: I think it refers to  Actions --> Advanced --> GUI --> Debugging

AudriusButkevicius: I never noticed it, it's not obvious what it does, so perhaps we should add a message explaining what it enables (if it's not already there, have not checked).

cmd/syncthing/support_bundle.go
EserGozcu: Would not it kill the purpose of this feature request? if we mock the Device ID or anonymise the IP addresses at the logs here, it will be hard to troubleshoot an ongoing issue, which in this case we want this Support Bundle as helpful as possible?

AudriusButkevicius: I don't think we need identities to debug things?

gui/default/index.html
bahadir60: @AudriusButkevicius I'm not sure are you asking this, just for information.
[Syncthing Docs](https://docs.syncthing.net/users/profiling.html) profiling part explains how to enable `Debugging` and what is for.

AudriusButkevicius: Essentially, all I was saying is that I never noticed there being a debug flag for GUI available in the UI.

cmd/syncthing/support_bundle.go
EserGozcu: loud thinking here; What about instead mocking the DeviceID, just showing the first 7 digits of it like it is seen at the UI?  If we mock whole DeviceID, it will be difficult to pinpoint the issue or correlate this with the DeviceID we have at the GUI. Imagine there are 100 Devices there..

AudriusButkevicius: Sure, yet I suspect it will be easier to replace it with a integer based dummy than be able to serialise only first few characters.

cmd/syncthing/gui.go
EserGozcu: Can we keep the IP as it is since it is seen at GUI, for the consistency ?

AudriusButkevicius: I am not saying we have to anything at all, I am happy to send ips as they are, but I am saying that it might lack adoption if we don't annonimise stuff, as we have plenty of paranoid people using this.

@calmh?

cmd/syncthing/support_bundle.go
bahadir60: I didn't get any error when I tried the code, but your suggestion is more suitable. I will change it

cmd/syncthing/gui.go
bahadir60: yeap, I use filename for heapprof and cpuprof but I agree that output is not appropriate, going to change...

cmd/syncthing/gui.go
bahadir60: yeap going to change

cmd/syncthing/gui.go
bahadir60: as your estimation F for file,I will use panicFile.

gui/default/index.html
calmh: I'm not sure where we would expose that. Currently the process is that I say "can you take a profile, here's the process" and link to the doc above which explains it. I envision the new process would be "can you grab a support bundle, here's the process" and a link to that document which would gain a new section.... Or something

cmd/syncthing/gui.go
bahadir60: As @calmh suggestion I will implement redaction for device ID and IPs

cmd/syncthing/gui.go
bahadir60: I'm not sure warning showing on UI about that,  in my opinion it is not more important thing but like config old or Syncthing old version, making a backup in local computer will be useful.

cmd/syncthing/gui.go
AudriusButkevicius: Lets do that in a separate pr tho.

EserGozcu: @calmh @AudriusButkevicius  What about the relevant lines at the existing Logs? Does this mean that we need to do same for the  existing Log implementation to cover the security usecase? I can see MyID at GUI -->  Actions --> Logs    If we change this at the current implementation, supportBundle does not have to go through logs and do mocking there since it will already be anonymous.

cmd/syncthing/gui.go
AudriusButkevicius: Also, capitalisation of newfilepath

AudriusButkevicius: Also in case of error this would produce an empty file with no indication to the user that it failed, or why...

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,87 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	// Write buffer zip to local zip file (back up)
+	err := ioutil.WriteFile(newfilepath, newfileBuffer.Bytes(), 0500)
+	if err != nil {
+		l.Infof("Support Bundle zip could not be created locally: %s", newfilepath)
+	}
+
+	// Serve the buffer zip to client for download
+	w.Header().Set("Content-Type", "application/zip")
+	w.Header().Set("Content-Disposition", "attachment; filename="+output)
+	io.Copy(w, strings.NewReader(newfileBuffer.String()))
```
AudriusButkevicius: Do you really need to funnel this via a string. Can't you seek to start and reuse the buffer or something along those lines?

cmd/syncthing/support_bundle.go
AudriusButkevicius: That doesn't mean it can't happen.

cmd/syncthing/gui.go
bahadir60: I got the point. Thanks for suggestion, I change code snippet to `io.Copy(w, &zipFileBuffer)`.

cmd/syncthing/gui.go
AudriusButkevicius: People are concious of whats in the log as they see it when they copy it, they have no idea whats in the magic zip file they were asked to upload, unless they explicitly go in and check.

AudriusButkevicius: I think that log viewer could be used for other purposes than just pasting logs for support. You could actually read the logs to debug something locally hence need to see ids and ips.

bahadir60:
Based on comments I made changes. Thanks all.

cmd/syncthing/gui.go
calmh: Yeah, no redaction at the deeper levels. Only when creating support bundle. I'd probably do it regex based, IP numbers and device IDs are quite distinctive and can be detected that way.

I very much suggest a separate function to redact any given log line, and a unit test that runs a number of different log lines through it.

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,93 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	data []byte
+}
+
+func (s *apiService) getSupportBundle(w http.ResponseWriter, r *http.Request) {
+	var files []fileEntry
+
+	// Redacted configuration as a JSON
+	if jsonConfig, err := json.MarshalIndent(getRedactedConfig(s), "", "  "); err == nil {
+		files = append(files, fileEntry{name: "config.json", data: jsonConfig})
+	}
```
AudriusButkevicius: Should we not do something in case of error, rather than skipping it completely? Warn or something atleast?

cmd/syncthing/gui.go
```diff
@@ -1024,6 +1027,93 @@ func (s *apiService) getSystemLogTxt(w http.ResponseWriter, r *http.Request) {
...
+	filename = fmt.Sprintf("syncthing-cpu-%s-%s-%s-%s.pprof", runtime.GOOS, runtime.GOARCH, Version, time.Now().Format("150405")) // hhmmss
+	pprof.StartCPUProfile(&cpuBuffer)
+	time.Sleep(duration)
+	pprof.StopCPUProfile()
+	files = append(files, fileEntry{name: filename, data: cpuBuffer.Bytes()})
+
+	// Add buffer files to buffer zip
+	var zipFilesBuffer bytes.Buffer
+	if err := zipBuffer(&zipFilesBuffer, files); err != nil {
+		l.Warnln("Support Bundle zip not created: ", err)
```
AudriusButkevicius: "Failed to create support bundle zip:"

AudriusButkevicius: Also, no point creating zip file etc if ti failed? Just return a 500 and be done with it?

cmd/syncthing/support_bundle.go
```diff
@@ -0,0 +1,52 @@
...
+		zipFile, err := zipWriter.Create(file.name)
+		if err != nil {
+			l.Infoln("Zipwriter not create filename:", err)
+		}
+		_, err = zipFile.Write(file.data)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
AudriusButkevicius: replace with `return zipWriter.Close()`, and make the defer simply double close in case of success (which should be noop)

cmd/syncthing/gui.go
bahadir60: I added error checks for all json marshalling, I'm not sure needed or not. Any suggestions?

calmh:
I made some style tweaks to your branch

Issue #5144, opened 2018-08-25
Tests fail on Go 1.11 / Windows


calmh:
```
[12:03:11]	[Step 4/5] TestFolderCheckPath
[12:03:11]	[TestFolderCheckPath] [Test Output]
2018/08/25 13:01:59.058460 folderconfiguration.go:99: INFO: bug: uncached filesystem call (should only happen in tests)
2018/08/25 13:01:59.058460 folderconfiguration.go:99: INFO: bug: uncached filesystem call (should only happen in tests)
2018/08/25 13:01:59.058460 folderconfiguration.go:99: INFO: bug: uncached filesystem call (should only happen in tests)
2018/08/25 13:01:59.058460 folderconfiguration.go:99: INFO: bug: uncached filesystem call (should only happen in tests)
2018/08/25 13:01:59.059176 folderconfiguration.go:99: INFO: bug: uncached filesystem call (should only happen in tests)
2018/08/25 13:01:59.059176 folderconfiguration.go:99: INFO: bug: uncached filesystem call (should only happen in tests)
    config_test.go:480: running with symlink check
    config_test.go:500: unexpected error in case link: folder path missing != %!s(<nil>)
```

```
[12:03:11]	[Step 4/5] TestWalkRootSymlink
[12:03:11]	[TestWalkRootSymlink] [Test Output]
    walk_test.go:336: expected two files, not 0
```

Looks symlink related, as usual.

AudriusButkevicius:
Only the walk one fails for me locally, yet I am still on the rc

calmh:
Difference could be because build server runs as local admin and you do not, or vice versa, or they differ in enablement of symlinks/developer mode perhaps.

AudriusButkevicius:
No, I did make sure both run properly and are not skipped.

calmh:
Yet fails it does `¯\_(ツ)_/¯`

AudriusButkevicius:
Nvm, my test was wrong, as I had a file symlink.

AudriusButkevicius:
https://github.com/golang/go/issues/27225

calmh:
Right. But then we do the "rooted". This (new) `fs` test should pass, imho:

```
func TestWindowsFolderRoot(t *testing.T) {
	fs := BasicFilesystem{root: `c:\test`}
	rooted, err := fs.rooted(".")
	if err != nil {
		t.Fatal(err)
	}
	if rooted != `c:\test` {
[...]

PR #5146, opened 2018-08-26
 lib/fs: Don't append path sep to root in rooted (ref #5144) 

imsodin:
### Purpose

By @calmh in #5144 

> We take great care to make sure a folder root doesn't end in slash on Windows...

> Right. But then we do the "rooted". This (new) fs test should pass, imho:
> 
> ```
> func TestWindowsFolderRoot(t *testing.T) {
> 	fs := BasicFilesystem{root: `c:\test`}
> 	rooted, err := fs.rooted(".")
> 	if err != nil {
> 		t.Fatal(err)
> 	}
> 	if rooted != `c:\test` {
> 		t.Error("Incorrect root:", rooted)
> 	}
> }
> ```
> 
> But it does not:
> 
> ```
> basicfs_windows_test.go:101: Incorrect root: c:\test\
> ```
> 
> So we undo the efforts of our root normalization stuff from earlier...
> 

This treats the case specially when the  path given to `rooted` is equivalent to `.` (folder root), in that it just returns `BasicFilesystem.root` without modifying it, thus leaving the precautions taken in `newBasicFilesystem` untouched. 

### Testing

Needed to adjust and add some test cases to `TestRooted` and added the test proposed by Jakob above.


calmh:
I don't think that's the right precaution. I think the right one is "paths on windows should never end in slash" and if we do that in rooted() we might not need to also do it in the fs constructor...

imsodin:
@calmh Then I think we need to talk about all the path separator precautions in place, and which is really the desired behaviour. From the current implementation: 

1. On Windows drive letters as folder root get a trailing path set: `C:` -> `C:\`
2. On Windows folder roots do not have a trailing path sep.
3. On all other platforms roots must have a trailing path sep (to penetrate symlinks).
4. Rooted paths other than `.` never have a trailing path sep.
5. `.` when rooted gets a trailing path sep.

Now apparently the no trailing path separator on windows ever seems to be a valid concern. I do not see what's the reason for 3. - how do trailing path separators help penetrate symlinks? And why would that even be desired? Generally the "canonical" form of paths is Go is without a trailing path separator (which is also pretty usual in other places as far as I know).

So unless there is a good reason for 3, I'd say we just go with no trailing path separator in the resulting paths (steps in between might still require it due to `HasPrefix` checks, but that need not influence the final result)

calmh:
No 3 is desired on Unixes because it handles folder roots which are symlinks because foo/ will stat as directory when foo stats as symlink. Maybe that’s historical and not needed any more if we specifically allow the folder root to stat as symlink. If so we can probably simplify to always stripping trailing slashes.

calmh:
`[Step 5/5] open C:\BuildAgent\temp\buildTmp\go-build518408782\b001\exe\a.out.exe: Operation did not complete successfully because the file contains a virus or potentially unwanted software.` 

😞

calmh:
I had a frank discussion with Windows Defender and I think we managed to agree that the build agent artifacts are not to be considered threats henceforth.

This LGTM - or rather, it looks like we're removing some safety checks, but the tests that verify them still pass so I'm OK with that. I think my new temporary test should just be folded into the list of paths that we already checkin the other test you tweaked, last among the "windows special items" that already checked we handled the `c:\` root correctly.

But that's a nit. 👍 from me

imsodin:
@calmh 

Thanks for taking care of that devil (well generally taking care of that wind... stuff :) ).  

The removed safety checks were redundant since the introduction of `Canonicalize`.  

I addressed the nit and rebased, as the original commits and commit messages aren't relevant anymore.

PR #5147, opened 2018-08-26
all: Skip fewer tests due to symlinks on windows (fixes #4887)

### LINKED ISSUE #4887: Don't skip tests related to symlinks on windows
As per https://github.com/syncthing/syncthing/pull/4886#issuecomment-382076308:

> [...] So with this in mind, maybe we can make an attempt to fix the following symlink related tests that are skipped on Windows (maybe not all of them symlink related, but clearly many):
> 
> screen shot 2018-04-17 at 19 28 03

`osutil.DebugSymlinkForTestsOnly` is the magic that makes it possible. For fs tests we'd need to split the osutil package to prevent cyclic imports.

imsodin:
### Purpose

This only concerns tests, see #4887.  
The only test that could be "un-skipped" was `TestWatchSymlinkedRoot`, all other skipped tests concern scanning or pulling, which we do not support (regardless of what `os` supports). All the other changes are either due to `DebugSymlinkForTestsOnly` being moved to the `fs` package (due to cyclic imports) or handling test cancellation in that function too (thus prohibiting it's use outside of testing). I left the hack in  `DebugSymlinkForTestsOnly` as a fallback if `os.CreateSymlink` still doesn't work (e.g. on old Go or Windows versions).

### Testing

Still pass on linux, pass on windows too, except for the known go1.11 failures which are being addressed in another PR.

### DIFF
--- a/lib/fs/debug_symlink.go
+++ b/lib/fs/debug_symlink.go
@@ -6,16 +6,20 @@
 
 // +build !windows
 
-package osutil
+package fs
 
 import (
 	"os"
+	"testing"
 )
 
 // DebugSymlinkForTestsOnly is not and should not be used in Syncthing code,
 // hence the cumbersome name to make it obvious if this ever leaks. Its
 // reason for existence is the Windows version, which allows creating
 // symlinks when non-elevated.
-func DebugSymlinkForTestsOnly(oldname, newname string) error {
-	return os.Symlink(oldname, newname)
+func DebugSymlinkForTestsOnly(t *testing.T, oldname, newname string) {
+	t.Helper()
+	if err := os.Symlink(oldname, newname); err != nil {
+		t.Fatal(err)
+	}
 }
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/d4l3k/messagediff"
 	"github.com/syncthing/syncthing/lib/fs"
-	"github.com/syncthing/syncthing/lib/osutil"
 	"github.com/syncthing/syncthing/lib/protocol"
 )
 
@@ -429,6 +428,8 @@ func TestFolderCheckPath(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	fs.DebugSymlinkForTestsOnly(t, filepath.Join(n, "dir"), filepath.Join(n, "link"))
+
 	testcases := []struct {
 		path string
 		err  error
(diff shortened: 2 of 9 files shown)


lib/fs/debug_symlink.go
```diff
@@ -6,16 +6,20 @@
 
 // +build !windows
 
-package osutil
+package fs
 
 import (
 	"os"
+	"testing"
```
calmh: Nope nope nope. :) This links the actual binary with the testing package.

```
jb@kvin:~/D/syncthing-macos-amd64-v0.14.50-rc.2+27-g28b0dfee $ ./syncthing --help
Usage:
  syncthing [options]

Options:
  -audit                          Write events to audit file
  -auditfile=""                   Specify audit file (use "-" for stdout, "--" for stderr)
  -browser-only                   Open GUI in browser
  -device-id                      Show the device ID
  -generate=""                    Generate key and config in specified dir, then exit
  -gui-address=""                 Override GUI address (e.g. "http://192.0.2.42:8443")
  -gui-apikey=""                  Override GUI API key
  -home=""                        Set configuration directory
  -logfile="-"                    Log file name (still always logs to stdout). Cannot be used together with -no-restart/STNORESTART environment variable.
  -logflags="2"                   Select information in log line prefix (see below)
  -no-browser                     Do not start browser
  -no-restart                     Disable monitor process, managed restarts and log file writing
  -paths                          Show configuration paths
  -paused                         Start with all devices and folders paused
  -reset-database                 Reset the database, forcing a full rescan and resync
  -reset-deltas                   Reset delta index IDs, forcing a full index exchange
  -test.bench=""                  run only benchmarks matching `regexp`
  -test.benchmem                  print memory allocations for benchmarks
  -test.benchtime="1s"            run each benchmark for duration `d`
  -test.blockprofile=""           write a goroutine blocking profile to `file`
  -test.blockprofilerate="1"      set blocking profile `rate` (see runtime.SetBlockProfileRate)
  -test.count="1"                 run tests and benchmarks `n` times
  -test.coverprofile=""           write a coverage profile to `file`
  -test.cpu=""                    comma-separated `list` of cpu counts to run each test with
  -test.cpuprofile=""             write a cpu profile to `file`
  -test.failfast                  do not start new tests after the first test failure
  -test.list=""                   list tests, examples, and benchmarks matching `regexp` then exit
  -test.memprofile=""             write an allocation profile to `file`
  -test.memprofilerate="0"        set memory allocation profiling `rate` (see runtime.MemProfileRate)
 ... and so on
```

imsodin:
I don't have time right now to redo this. Thanks @calmh for noticing the stupidity :)

PR #5169, opened 2018-09-07
ldap auth mode

cyberb:
### Purpose

Allows gui to authenticate against ldap using simple bind approach.

Additional config section:

    <authMode>ldap</authMode>
    <ldapServer>localhost</ldapServer>
    <ldapPort>389</ldapPort>
    <ldapBindDn>cn=%s,ou=users,dc=syncloud,dc=org</ldapBindDn>

Existing user/password authentication is preserved and can be used as before if authMode is not present or not equals to ldap.

### Testing

Few mising unit tests are added for existing auth.
No unit tests for ldap, although integration tests are available in Syncloud project:
https://github.com/syncloud/syncthing/blob/master/integration/verify.py#L99

## Authorship

Boris Rybalkin (ribalkin@gmail.com)

### DIFF
--- a/cmd/syncthing/gui_auth.go
+++ b/cmd/syncthing/gui_auth.go
@@ -12,12 +12,15 @@ import (
 	"net/http"
 	"strings"
 	"time"
+	"fmt"
+	"log"
 
 	"github.com/syncthing/syncthing/lib/config"
 	"github.com/syncthing/syncthing/lib/events"
 	"github.com/syncthing/syncthing/lib/rand"
 	"github.com/syncthing/syncthing/lib/sync"
 	"golang.org/x/crypto/bcrypt"
+	"gopkg.in/ldap.v2"
 )
 
 var (
@@ -77,42 +80,21 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
 			return
 		}
 
-		// Check if the username is correct, assuming it was sent as UTF-8
-		username := string(fields[0])
-		if username == cfg.User {
-			goto usernameOK
-		}
+        authOk := false
+        username := string(fields[0])
+        password := string(fields[1])
+        if cfg.IsAuthModeLdap() {
+            authOk = AuthLdap(username, password, cfg.LdapServer, cfg.LdapPort, cfg.LdapBindDn)
+        } else {
+            authOk, username = AuthSimple(username, password, cfg.User, cfg.Password)
+        }
 
-		// ... check it again, converting it from assumed ISO-8859-1 to UTF-8
-		username = string(iso88591ToUTF8(fields[0]))
-		if username == cfg.User {
-			goto usernameOK
-		}
+        if !authOk {
+            emitLoginAttempt(false, username)
+            error()
+            return
+        }
 
-		// Neither of the possible interpretations match the configured username
-		emitLoginAttempt(false, username)
-		error()
-		return
-
-	usernameOK:
-		// Check password as given (assumes UTF-8 encoding)
-		password := fields[1]
-		if err := bcrypt.CompareHashAndPassword([]byte(cfg.Password), password); err == nil {
-			goto passwordOK
-		}
-
-		// ... check it again, converting it from assumed ISO-8859-1 to UTF-8
-		password = iso88591ToUTF8(password)
-		if err := bcrypt.CompareHashAndPassword([]byte(cfg.Password), password); err == nil {
-			goto passwordOK
-		}
-
-		// Neither of the attempts to verify the password checked out
-		emitLoginAttempt(false, username)
-		error()
-		return
-
-	passwordOK:
 		sessionid := rand.String(32)
 		sessionsMut.Lock()
 		sessions[sessionid] = true
@@ -126,6 +108,45 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
 		emitLoginAttempt(true, username)
 		next.ServeHTTP(w, r)
 	})
+}
+
+func AuthSimple(username string, password string, configUser string, configPassword string) (bool, string) {
+
+    if username != configUser {
+        username = string(iso88591ToUTF8([]byte(username)))
+        if username != configUser {
+            return false, ""
+        }
+    }
+
+    configPasswordBytes := []byte(configPassword)
+    passwordBytes := []byte(password)
+    if err := bcrypt.CompareHashAndPassword(configPasswordBytes, passwordBytes); err != nil {
+        passwordBytes = iso88591ToUTF8(passwordBytes)
+        if err := bcrypt.CompareHashAndPassword(configPasswordBytes, passwordBytes); err != nil {
+            return false, ""
+        }
+    }
+
+    return true, username
+}
+
+func AuthLdap(username string, password string, server string, port int, bindDn string) bool {
+    l, err := ldap.Dial("tcp", fmt.Sprintf( "%s:%d", server, port))
+    if err != nil {
+        log.Println(err)
+        return false
+    }
+
+    defer l.Close()
+
+    err = l.Bind(fmt.Sprintf(bindDn, username), password)
+    if err != nil {
+        log.Println(err)
+        return false
+    }
+
+    return true
 }
 
 // Convert an ISO-8859-1 encoded byte string to UTF-8. Works by the
--- a/cmd/syncthing/gui_auth_test.go
+++ b/cmd/syncthing/gui_auth_test.go
@@ -1,0 +1,43 @@
+// Copyright (C) 2014 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package main
+
+import (
+	"testing"
+	"golang.org/x/crypto/bcrypt"
+
+    "github.com/syncthing/syncthing/cmd/syncthing"
+
+)
+
+func TestSimpleAuthOK(t *testing.T) {
+    passwordHashBytes, _ := bcrypt.GenerateFromPassword([]byte("pass"), 14)
+    ok, username := main.AuthSimple("user", "pass", "user", string(passwordHashBytes))
+    if !ok {
+        t.Fatalf("should pass auth")
+    }
+    if username != "user" {
+        t.Fatalf(username)
+    }
+}
+
+func TestSimpleAuthUsernameFail(t *testing.T) {
+    passwordHashBytes, _ := bcrypt.GenerateFromPassword([]byte("pass"), 14)
+    ok, _ := main.AuthSimple("userWRONG", "pass", "user", string(passwordHashBytes))
+    if ok {
+        t.Fatalf("should fail auth")
+    }
+}
+
+func TestSimpleAuthPasswordFail(t *testing.T) {
+    passwordHashBytes, _ := bcrypt.GenerateFromPassword([]byte("passWRONG"), 14)
+    ok, _ := main.AuthSimple("user", "pass", "user", string(passwordHashBytes))
+    if ok {
+        t.Fatalf("should fail auth")
+    }
+}
+
--- a/lib/config/guiconfiguration.go
+++ b/lib/config/guiconfiguration.go
@@ -17,6 +17,10 @@ type GUIConfiguration struct {
 	RawAddress                string `xml:"address" json:"address" default:"127.0.0.1:8384"`
 	User                      string `xml:"user,omitempty" json:"user"`
 	Password                  string `xml:"password,omitempty" json:"password"`
+	AuthMode                  string `xml:"authMode,omitempty" json:"authMode"`
+	LdapServer                string `xml:"ldapServer,omitempty" json:"ldapServer"`
+	LdapPort                  int    `xml:"ldapPort,omitempty" json:"ldapPort"`
+	LdapBindDn                string `xml:"ldapBindDn,omitempty" json:"ldapBindDn"`
 	RawUseTLS                 bool   `xml:"tls,attr" json:"useTLS"`
 	APIKey                    string `xml:"apikey,omitempty" json:"apiKey"`
 	InsecureAdminAccess       bool   `xml:"insecureAdminAccess,omitempty" json:"insecureAdminAccess"`
@@ -24,6 +28,18 @@ type GUIConfiguration struct {
 	Debugging                 bool   `xml:"debugging,attr" json:"debugging"`
 	InsecureSkipHostCheck     bool   `xml:"insecureSkipHostcheck,omitempty" json:"insecureSkipHostcheck"`
 	InsecureAllowFrameLoading bool   `xml:"insecureAllowFrameLoading,omitempty" json:"insecureAllowFrameLoading"`
+}
+
+func (c GUIConfiguration) IsAuthEnabled() bool {
+    return c.IsAuthModeSimple() || c.IsAuthModeLdap()
+}
+
+func (c GUIConfiguration) IsAuthModeLdap() bool {
+    return c.AuthMode == "ldap"
+}
+
+func (c GUIConfiguration) IsAuthModeSimple() bool {
+    return (len(c.User) > 0 && len(c.Password) > 0)
 }
 
 func (c GUIConfiguration) Address() string {
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -343,7 +343,7 @@ func (s *apiService) Serve() {
 	handler = withDetailsMiddleware(s.id, handler)
 
 	// Wrap everything in basic auth, if user/password is set.
-	if len(guiCfg.User) > 0 && len(guiCfg.Password) > 0 {
+	if guiCfg.IsAuthEnabled() {
 		handler = basicAuthAndSessionMiddleware("sessionid-"+s.id.String()[:5], guiCfg, handler)
 	}
 


AudriusButkevicius:
So in principle this looks okay'ish to me other tham formatting issues etc. I would however like to propose for discussion (to see if people agree or disagree) hiding all the ldap settings behind an environment variable rather than config, this way we don't bleed ldap supports existance to the config which would probably be very rarely used anyway.nas in setting STLDAPSERVER=foo.com:1234 STLDAPSUBJECT=foo and checking if these env vars are set switching to ldap auth.

cmd/syncthing/gui_auth.go
```diff
@@ -128,6 +110,45 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
 	})
 }
 
+func AuthSimple(username string, password string, configUser string, configPassword string) (bool, string) {
+
+    if username != configUser {
+        username = string(iso88591ToUTF8([]byte(username)))
```
calmh: This "might be UTF-8, might be ISO-8859-1" business with the username and password is a property of HTTP basic auth and affects auth over LDAP as much as built in auth. Hence it should be cleared up on some higher level.

lib/config/guiconfiguration.go
```diff
@@ -26,6 +30,18 @@ type GUIConfiguration struct {
 	InsecureAllowFrameLoading bool   `xml:"insecureAllowFrameLoading,omitempty" json:"insecureAllowFrameLoading"`
 }
 
+func (c GUIConfiguration) IsAuthEnabled() bool {
+    return c.IsAuthModeSimple() || c.IsAuthModeLdap()
+}
+
+func (c GUIConfiguration) IsAuthModeLdap() bool {
+    return c.AuthMode == "ldap"
```
calmh: AuthMode here should be an enum of sorts and not a plain string I think.

lib/config/guiconfiguration.go
```diff
@@ -12,15 +12,28 @@ import (
 	"strings"
 )
 
+const (
+	AuthModeSimple = "simple"
+	AuthModeLDAP   = "ldap"
+)
```
AudriusButkevicius: These consts probably should be their own types and enums, like FolderType et al.

cmd/syncthing/gui_auth.go
```diff
@@ -110,43 +116,62 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
+		log.Println(err)
+		return false
+	}
 
-    err = l.Bind(fmt.Sprintf(bindDn, username), password)
-    if err != nil {
-        log.Println(err)
-        return false
-    }
+	if cfg.IsLDAPTLSModeStartTSL() {
```
AudriusButkevicius: Instead of this IsModeX IsModeY I'd do
```
switch cfg.GetTLSMode() {
   case TlsModeFoo:
        connection = foo.connect()
   case TlsModeBar:
         connection = barr.dial() 
```

then you only need one accessor on the config object.

lib/config/guiconfiguration.go
```diff
@@ -12,15 +12,28 @@ import (
...
 	Password                  string `xml:"password,omitempty" json:"password"`
 	AuthMode                  string `xml:"authMode,omitempty" json:"authMode"`
-	LdapServer                string `xml:"ldapServer,omitempty" json:"ldapServer"`
-	LdapPort                  int    `xml:"ldapPort,omitempty" json:"ldapPort"`
-	LdapBindDn                string `xml:"ldapBindDn,omitempty" json:"ldapBindDn"`
+	LDAPServer                string `xml:"LDAPServer,omitempty" json:"LDAPServer"`
+	LDAPPort                  int    `xml:"LDAPPort,omitempty" json:"LDAPPort"`
+	LDAPBindDn                string `xml:"LDAPBindDn,omitempty" json:"LDAPBindDn"`
+	LDAPTLSMode               string `xml:"LDAPTLSMode,omitempty" json:"LDAPTLSMode"`
+	LDAPInsecureSkipVerify    bool   `xml:"LDAPInsecureSkipVerify,attr" json:"LDAPInsecureSkipVerify" default:"false"`
```
AudriusButkevicius: Again, discussion point before we make any changes.

If we are starting to add LDAP support to syncthing, perhaps we should have a separate LDAPConfiguration section in the config, and then GUIAuthMode == LDAP uses the LDAPConfiguration, and then once we add some other non-GUI feature that uses LDAP, we don't have to add yet another LDAP configuration to some other part of the config.


/cc @calmh

cmd/syncthing/gui_auth_test.go
```diff
@@ -16,26 +16,24 @@ import (
...
     passwordHashBytes, _ := bcrypt.GenerateFromPassword([]byte("pass"), 14)
-    ok, username := main.AuthSimple("user", "pass", "user", string(passwordHashBytes))
+    ok := main.AuthSimple("user", "pass", "user", string(passwordHashBytes))
     if !ok {
         t.Fatalf("should pass auth")
     }
-    if username != "user" {
-        t.Fatalf(username)
-    }
+
```
AudriusButkevicius: Spurious new line

cmd/syncthing/gui_auth.go
```diff
@@ -110,43 +116,62 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
+
+func AuthSimple(username string, password string, configUser string, configPassword string) bool {
+
+	if username != configUser {
+		return false
+	}
+
+	configPasswordBytes := []byte(configPassword)
+	passwordBytes := []byte(password)
+	if err := bcrypt.CompareHashAndPassword(configPasswordBytes, passwordBytes); err != nil {
```
AudriusButkevicius: Can just be `return CompareAndHasPassword(...) == nil`

cmd/syncthing/gui_auth.go
```diff
@@ -110,43 +116,62 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
+func auth(username string, password string, cfg config.GUIConfiguration) bool {
+	if cfg.IsAuthModeLDAP() {
+		return AuthLDAP(username, password, cfg)
+	} else {
+		return AuthSimple(username, password, cfg.User, cfg.Password)
+	}
+}
+
+func AuthSimple(username string, password string, configUser string, configPassword string) bool {
+
```
AudriusButkevicius: blank line

cmd/syncthing/gui_auth.go
```diff
@@ -110,43 +116,62 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
-        }
-    }
-
-    return true, username
+func auth(username string, password string, cfg config.GUIConfiguration) bool {
+	if cfg.IsAuthModeLDAP() {
+		return AuthLDAP(username, password, cfg)
+	} else {
+		return AuthSimple(username, password, cfg.User, cfg.Password)
+	}
```
AudriusButkevicius: Given auth is private (lowercase), no point for auth methods to be public.

cmd/syncthing/gui_auth.go
```diff
@@ -110,43 +116,62 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
 }
 
-func AuthLdap(username string, password string, server string, port int, bindDn string) bool {
-    l, err := ldap.Dial("tcp", fmt.Sprintf( "%s:%d", server, port))
-    if err != nil {
-        log.Println(err)
-        return false
-    }
+func AuthLDAP(username string, password string, cfg config.GUIConfiguration) bool {
+
```
AudriusButkevicius: blank line

cmd/syncthing/gui_auth.go
```diff
@@ -110,43 +116,62 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
+	var err error
+	if cfg.IsLDAPTLSModeTSL() {
+		connection, err = ldap.DialTLS("tcp", address, &tls.Config{InsecureSkipVerify: cfg.LDAPInsecureSkipVerify})
+	} else {
+		connection, err = ldap.Dial("tcp", address)
+	}
 
-    defer l.Close()
+	if err != nil {
+		log.Println(err)
```
AudriusButkevicius: We don't use log package, we have `l.{Debug/Info/Warn}ln` for this type of stuff.

cyberb:
1. Renamed all the ldap/Ldap to LDAP
2. Added LDAPTLSMode: nontls/tls/starttls (did not test tls/starttls, based on examples from: https://github.com/go-ldap/ldap/blob/master/ldap_test.go)
3. Not sure how to properly do enums in go, so added few consts
4. Not sure how to properly handle dependencies so added ldap lib into vendor libs (to make TeamCity happy)
5. Not sure how to run tests so removed test import cycle (to make TeamCity happy)

vendor/gopkg.in/ldap.v2/LICENSE
```diff
@@ -0,0 +1,22 @@
+The MIT License (MIT)
```
AudriusButkevicius: I'd prefer that we used the github hosted version of the package. Given we use vendoring, and go modules just around the corner, using gopkg.in makes little sense.

lib/config/guiconfiguration.go
calmh: Yeah, I agree with giving this thing its own config struct/section, and perhaps combining server+port into just an address, we don't typically separate the parts in other places of the config.

cmd/syncthing/gui_auth.go
```diff
@@ -128,6 +116,64 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
 
+func auth(username string, password string, cfg config.GUIConfiguration) bool {
+	if cfg.IsAuthModeLDAP() {
+		return authLDAP(username, password, cfg)
+	} else {
+		return authSimple(username, password, cfg.User, cfg.Password)
+	}
+}
+
+func authSimple(username string, password string, configUser string, configPassword string) bool {
```
calmh: It's not strictly required for this PR, but the auth functions that have the same signature (or could have) could be types that implement an interface, with something that creates the right one based on the config. It's a bit boilerplate-ish and can be easily added after the fact so skip that for now if you think it's easier.

Not doing any further style nitpicking as Audrius is already all over it. :)

vendor/gopkg.in/ldap.v2/LICENSE
calmh: This thing also imports `gopkg.in/asn1-ber.v1` so while I agree in principle I suspect the fight is lost on this package

lib/config/guiconfiguration.go
AudriusButkevicius: I think you pick one or the other?

calmh: Maybe. Or we fall back to trying static if LDAP is down? Can we distinguish "down" from "denied" based on the bind result?

cyberb: Technically Dial is up/down check and Bind is allowed/denied check and they are separate.

calmh: Cool then it's easy to separate. So what's the desired outcome here? Assuming LDAP is configured,

- LDAP up, bind denied => denied
- LDAP down, static auth configured => follow static auth
- LDAP down, no static auth => denied (I guess? or allowed?)

cyberb: I personally prefer no fallback. If i set auth mode to be ldap and it does not work i would like know right away as an admin, while fallback would waste my time.
I would say:
1. Auth mode is set to ldap do only ldap
2. Auth mode is set to static do only static
3. Auth mode not set do static if login and password set (config backward compatibility for now)

cyberb: I am working on enums as suggested. Static is better, agree. Also extracting ldap config section. Will do another push soon.
1. Can you show me where the docs are?
2. Also I am still not sure how do you do deps. I had to do go get ./... but your TeamCity seems to not do that, so I had to include libs in vendor dir. Is this ok?

calmh: Docs are at docs.syncthing.net, github.com/syncthing/docs. Dependencies need to be vendored, yes. We use gvt (https://github.com/FiloSottile/gvt). Essentially, `gvt fetch gopkg.in/whatever` and it'll add it to `vendor/` and update the manifest.

cyberb:
1. Extracted enums (ldap configuration, auth type)
2. This time I used gvt fetch gopkg.in/ldap.v2

cyberb:
Let me check correct logger

lib/config/ldaptlsmode.go
```diff
@@ -0,0 +1,46 @@
...
+
+const (
+	LDAPTLSModeNoTLS LDAPTLSMode = iota // default is notls
+	LDAPTLSModeTLS
+	LDAPTLSModeStartTLS
+)
+
+func (t LDAPTLSMode) String() string {
+	switch t {
+	case LDAPTLSModeNoTLS:
```
AudriusButkevicius: Perhaps this should be plain not notls?

cmd/syncthing/gui_auth.go
```diff
@@ -128,6 +116,60 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
+func auth(username string, password string, cfg config.GUIConfiguration) bool {
+	if cfg.AuthMode == config.AuthModeLDAP {
+		return authLDAP(username, password, cfg.LDAPAuth)
+	} else {
+		return authStatic(username, password, cfg.User, cfg.Password)
+	}
+}
+
+func authStatic(username string, password string, configUser string, configPassword string) bool {
+
```
AudriusButkevicius: Blank line

lib/config/authmode.go
```diff
@@ -0,0 +1,41 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package config
+
+type AuthMode int
+
+const (
+	AuthModeStatic AuthMode = iota // default is static
+	AuthModeLDAP
```
AudriusButkevicius: I would add a None mode and stop checking for username and password from being set, and do a migration in the config setting auth mode.

This would also need fixes in the UI.

cmd/syncthing/gui_auth.go
```diff
@@ -128,6 +116,60 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
+	}
+
+	configPasswordBytes := []byte(configPassword)
+	passwordBytes := []byte(password)
+	return bcrypt.CompareHashAndPassword(configPasswordBytes, passwordBytes) == nil
+}
+
+func authLDAP(username string, password string, cfg config.LDAPConfiguration) bool {
+
+	address := cfg.LDAPAddress
```
AudriusButkevicius: Blank line

cmd/syncthing/gui_auth.go
```diff
@@ -128,6 +116,60 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
+
+func authStatic(username string, password string, configUser string, configPassword string) bool {
+
+	if username != configUser {
+		return false
+	}
+
+	configPasswordBytes := []byte(configPassword)
+	passwordBytes := []byte(password)
+	return bcrypt.CompareHashAndPassword(configPasswordBytes, passwordBytes) == nil
```
AudriusButkevicius: This might have been this way, but this just hit me. 

Perhaps move the username check after the hash check to prevent timing attacks on username probing.

cmd/syncthing/gui_auth.go
```diff
@@ -128,6 +116,60 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
+	var connection *ldap.Conn
+	var err error
+	if cfg.LDAPTLSMode == config.LDAPTLSModeTLS {
+		connection, err = ldap.DialTLS("tcp", address, &tls.Config{InsecureSkipVerify: cfg.LDAPInsecureSkipVerify})
+	} else {
+		connection, err = ldap.Dial("tcp", address)
+	}
+
+	if err != nil {
+		log.Println(err)
```
AudriusButkevicius: Please remove all usages of the log package and use the `l` logger instance.

lib/config/authmode.go
AudriusButkevicius: @calmh wdyt?

lib/config/guiconfiguration.go
```diff
@@ -13,17 +13,27 @@ import (
...
-	Theme                     string `xml:"theme" json:"theme" default:"default"`
-	Debugging                 bool   `xml:"debugging,attr" json:"debugging"`
-	InsecureSkipHostCheck     bool   `xml:"insecureSkipHostcheck,omitempty" json:"insecureSkipHostcheck"`
-	InsecureAllowFrameLoading bool   `xml:"insecureAllowFrameLoading,omitempty" json:"insecureAllowFrameLoading"`
+	Enabled                   bool              `xml:"enabled,attr" json:"enabled" default:"true"`
+	RawAddress                string            `xml:"address" json:"address" default:"127.0.0.1:8384"`
+	User                      string            `xml:"user,omitempty" json:"user"`
+	Password                  string            `xml:"password,omitempty" json:"password"`
+	AuthMode                  AuthMode          `xml:"authMode,omitempty" json:"authMode"`
+	LDAPAuth                  LDAPConfiguration `xml:"ldapAuth,omitempty" json:"ldapAuth"`
```
AudriusButkevicius: I would move LDAPConfigurarion as LDAPConfiguration to Config instead of GUIConfiguration.

cmd/syncthing/gui_auth.go
cyberb: It is actually:
Plain:
  dial
Tls:
  dialTls
StartTls:
  dial
  startTls

I can do it like this if you want, just dial will be repeated :)

lib/config/authmode.go
calmh: That would seem clean but if that were the default it would mean existing things that just set username and password would result in no auth which is bad.

cmd/syncthing/gui_auth.go
calmh: Currently both username and password is checked, and failure on either results in a short random delay before returning an error.

lib/config/config.go
```diff
@@ -125,17 +125,18 @@ func ReadJSON(r io.Reader, myID protocol.DeviceID) (Configuration, error) {
...
-	PendingDevices []ObservedDevice      `xml:"pendingDevice" json:"pendingDevices"`
-	XMLName        xml.Name              `xml:"configuration" json:"-"`
-
-	MyID            protocol.DeviceID `xml:"-" json:"-"` // Provided by the instantiator.
-	OriginalVersion int               `xml:"-" json:"-"` // The version we read from disk, before any conversion
+	Version           int                   `xml:"version,attr" json:"version"`
+	Folders           []FolderConfiguration `xml:"folder" json:"folders"`
+	Devices           []DeviceConfiguration `xml:"device" json:"devices"`
+	GUI               GUIConfiguration      `xml:"gui" json:"gui"`
+	LDAPConfiguration LDAPConfiguration     `xml:"ldapConfiguration" json:"ldapConfiguration"`
```
calmh: Drop the `Configuration` part of the JSON/XML names I think.

AudriusButkevicius: Also, potentially for the member name itself? GUIConfiguration is called GUI.

calmh: Aye

cmd/syncthing/gui_auth.go
```diff
@@ -128,6 +115,58 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
+	var connection *ldap.Conn
+	var err error
+	if cfg.LDAPTLSMode == config.LDAPTLSModeTLS {
+		connection, err = ldap.DialTLS("tcp", address, &tls.Config{InsecureSkipVerify: cfg.LDAPInsecureSkipVerify})
+	} else {
+		connection, err = ldap.Dial("tcp", address)
+	}
+
+	if err != nil {
+		l.Warnln(err)
```
AudriusButkevicius: If this is `connection reset by peer` error or alike, perhaps it would make sense adding context to it, something like
```l.Warnln("ldap.Dial", err)```

File: cmd/syncthing/gui_auth.go [hunk 23]
```diff
@@ -128,6 +115,58 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
+		return false
+	}
+
+	if cfg.LDAPTLSMode == config.LDAPTLSModeStartTLS {
+		err = connection.StartTLS(&tls.Config{InsecureSkipVerify: cfg.LDAPInsecureSkipVerify})
+		if err != nil {
+			l.Warnln(err)
```
AudriusButkevicius: ditto about context

cmd/syncthing/gui_auth.go
```diff
@@ -128,6 +115,58 @@ func basicAuthAndSessionMiddleware(cookieName string, cfg config.GUIConfiguratio
...
+			return false
+		}
+	}
+
+	defer connection.Close()
+
+	err = connection.Bind(fmt.Sprintf(cfg.LDAPBindDn, username), password)
+	if err != nil {
+		l.Warnln(err)
```
AudriusButkevicius: ditto about context

lib/config/guiconfiguration.go
```diff
@@ -13,17 +13,26 @@ import (
...
+	InsecureSkipHostCheck     bool              `xml:"insecureSkipHostcheck,omitempty" json:"insecureSkipHostcheck"`
+	InsecureAllowFrameLoading bool              `xml:"insecureAllowFrameLoading,omitempty" json:"insecureAllowFrameLoading"`
+}
+
+func (c GUIConfiguration) IsAuthEnabled() bool {
+	return c.IsAuthModeStatic() || c.AuthMode == AuthModeLDAP
+}
+
+func (c GUIConfiguration) IsAuthModeStatic() bool {
+	return (len(c.User) > 0 && len(c.Password) > 0)
```
AudriusButkevicius: no need for the outer braces

lib/config/ldaptlsmode.go
```diff
@@ -0,0 +1,46 @@
...
+const (
+	LDAPTLSModePlain LDAPTLSMode = iota // default is notls
+	LDAPTLSModeTLS
+	LDAPTLSModeStartTLS
+)
+
+func (t LDAPTLSMode) String() string {
+	switch t {
+	case LDAPTLSModePlain:
+		return "plain"
```
AudriusButkevicius: Now that I think about it, it's not plain tls, it's simply not tls at all, so perhaps this should not be LDAPTLSMode, but LDAPTransport or LDAPProtocol or something alike. Or the mode should be none, (as in non-tls).

Sorry for distracting you here earlier.

AudriusButkevicius:
Minus the nitpicks, this looks good to me.

lib/config/ldaptlsmode.go
cyberb: Lets do LDAPTransport with LDAPTransportPlain, LDAPTransportTLS and LDAPTransportStartTLS?

cmd/syncthing/gui_auth.go
```diff
@@ -124,34 +124,30 @@ func auth(username string, password string, guiCfg config.GUIConfiguration, ldap
 }
 
 func authStatic(username string, password string, configUser string, configPassword string) bool {
-	if username != configUser {
-		return false
-	}
-
 	configPasswordBytes := []byte(configPassword)
 	passwordBytes := []byte(password)
-	return bcrypt.CompareHashAndPassword(configPasswordBytes, passwordBytes) == nil
```
calmh: Nice

lib/config/ldapconfiguration.go
```diff
@@ -7,10 +7,10 @@
 package config
 
 type LDAPConfiguration struct {
-	LDAPAddress            string      `xml:"address,omitempty" json:"addresd"`
-	LDAPBindDn             string      `xml:"bindDn,omitempty" json:"bindDn"`
-	LDAPTLSMode            LDAPTLSMode `xml:"tlsMode,omitempty" json:"tlsMode"`
-	LDAPInsecureSkipVerify bool        `xml:"insecureSkipVerify,omitempty" json:"insecureSkipVerify" default:"false"`
+	LDAPAddress            string        `xml:"address,omitempty" json:"addresd"`
```
calmh: So, given we are now inside an LDAP struct we can skip the LDAP prefix here as well. And capitalize DN in BindDN.

cyberb:
Do you have any more comments?

AudriusButkevicius:
Oh, and thanks for your patience.

PR #5176, opened 2018-09-10
lib/fs: Case insensitive conversion to rel path on windows

imsodin:
https://forum.syncthing.net/t/panic-bug-notify-backend-is-processing-a-change-outside-of-the-filesystem-root/12133

Essentially there seems to be a problem with drive letter casing on windows again, but the test checking that is useless. This is currently just an attempt to make that test work and reproduce the problem, not yet a fix in itself.

### DIFF
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -90,10 +90,6 @@ func (f *BasicFilesystem) unrooted(path string) string {
 	return rel(path, f.root)
 }
 
-func rel(path, prefix string) string {
-	return strings.TrimPrefix(strings.TrimPrefix(path, prefix), string(PathSeparator))
-}
-
 func (f *BasicFilesystem) Chmod(name string, mode FileMode) error {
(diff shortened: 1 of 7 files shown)


imsodin:
As the test didn't reproduce the issue, but already two users independently reported it on the forum, I went the easy way and converted the paths to be compared to unicode lowercase with the existing method. My current working thesis, is that the backend (notify package) does not check/canonicalise paths received from the OS in any way, and the OS is not consistent in how it reports casing (that's likely, as it isn't consistent regarding the use of win83 names either). If this hypothesis were true, this was broken in earlier Syncthing version as well, i.e. no regression.

calmh:
This *looks* reasonable to me. Is there some sort of test coverage on this or are we shooting blindly?

calmh:
It was a genuine question, not to imply there weren't any tests. :) However I added a pass with unrootedChecked as well to see that it does not panic on these cases...

PR #5181, opened 2018-09-11
dockerfile: Improve UID/GID handling (fixes #5180)

### LINKED ISSUE #5180: Docker image fails when PGID set to existing group
The Docker image for 0.14.50 fails to start when the `PGID` environment variable is set to `100`

```
$ docker run --net=host -e PUID=1000 -e PGID=100 syncthing/syncthing
addgroup: gid '100' in use
```

### Version Information

Syncthing Version: v0.14.50
OS Version: Arch Linux, macOS High Sierra

calmh:
### Purpose

This removes the user and group juggling, which would fail when given
for example a PGID that already existed as the "syncthing" group could
then not be created with that PGID. It's not reasonable to expect the
user to know which group/user names/IDs are already present in the
Docker image.

Instead we now just launch under the specified IDs, while manually
setting the HOME env var to give us a home directory - the only thing we
needed the user entry for anyway.

Also updates to Go 1.11 and building without upgrades instead of
disabling by env var.

### Testing

I've ran it manually under existing and non-existing IDs, and it works for me; previously it didn't.

```
$ docker run -it --rm -e PUID=2 -e PGID=2 syncthing
$ docker run -it --rm -e PUID=12345 -e PGID=12345 syncthing
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.10 AS builder
+FROM golang:1.11 AS builder
 
 WORKDIR /go/src/github.com/syncthing/syncthing
 COPY . .
@@ -6,7 +6,7 @@ COPY . .
 ENV CGO_ENABLED=0
 ENV BUILD_HOST=syncthing.net
 ENV BUILD_USER=docker
-RUN rm -f syncthing && go run build.go build syncthing
+RUN rm -f syncthing && go run build.go -no-upgrade build syncthing
 
 FROM alpine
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -18,27 +18,15 @@ RUN apk add --no-cache ca-certificates su-exec
...
-          -h /var/syncthing \
-          -G "${PGRP}" \
-          -u "${PUID}" \
-          "${PUSR}" \
-    ) \
- && chown "${PUSR}:${PGRP}" /var/syncthing \
- && su-exec "${PUSR}:${PGRP}" \
+ENTRYPOINT \
+  chown "${PUID}:${PGID}" /var/syncthing \
+  && su-exec "${PUID}:${PGID}" \
```
AudriusButkevicius: Who creates them if they don't exist?
The user can choose to specify whatever he wants.

calmh: No-one creates them. They don't need to be created, that's the point of the change.

AudriusButkevicius: I didn't read the text, but I didn't think this was possible.

PR #5191, opened 2018-09-13
gui: Don't fail if pendingFolders is undefined (fixes #5190)

### LINKED ISSUE #5190: Javascript error in connection with new device notification
I just added a new device on .50-rc.2 and noticed the javascript error line at the top. The browser console shows:

```
Error: $scope.devices[i].pendingFolders is undefined
$scope.syncthingStatus@http://localhost:8384/syncthing/core/syncthingController.js:976:17
$parseFunctionCall@http://localhost:8384/vendor/angular/angular.js:12493:15
expressionInputWatch@http://localhost:8384/vendor/angular/angular.js:12902:31
$digest@http://localhost:8384/vendor/angular/angular.js:14423:34
$apply@http://localhost:8384/vendor/angular/angular.js:14694:13
scheduleApplyAsync/applyAsyncId<@http://localhost:8384/vendor/angular/angular.js:14984:11
completeOutstandingRequest@http://localhost:8384/vendor/angular/angular.js:4959:7
Browser/self.defer/timeoutId<@http://localhost:8384/vendor/angular/angular.js:5347:7
```

I haven't investigated any further.

It happens after pressing the save button when adding the new device.

imsodin:
#5190 

Manual testing.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -973,7 +973,7 @@ angular.module('syncthing.core')
                         deviceCount--;
                         break;
                 }
-                pendingFolders += $scope.devices[i].pendingFolders.length;
+                pendingFolders += $scope.sizeOf($scope.devices[i].pendingFolders);
             }
 
             // enumerate notifications


calmh:
That function looks optimized for maps/objects and the variable looks like a slice server-side?

AudriusButkevicius:
It should also never be null/undefined as we should be initialising everything to nil slices. We should fix it in go.

imsodin:
@AudriusButkevicius Can you fix this issue? I just proposed this PR because it was a really quick/simple fix when I debugged the reason for the JS error.

AudriusButkevicius:
Will do when I have time.

PR #5195, opened 2018-09-14
all: Check files on disk/in db when deleting/renaming (fixes #5194)

### LINKED ISSUE #5194: Unscanned/conflicting files are overwritten/removed in niche cases
If a file is changed on disk, but hasn't been scanned yet and it gets deleted or overwritten when renaming during pulling, the changed file gets overwritten. Instead it the operation should be aborted and the file on disk scheduled for scanning.  
If a file is in conflict with the one that is pulled, and it is overwritten in a rename operation, no conflict copy is created.  

Neither of these scenarios is very likely especially since watching for changes is available.

For the exact scenarios, refer to the unit tests in the PR that will soon be linked with this issue.

imsodin:
### Purpose

#5194 

The most relevant changes are those to `deleteFile`, `renameFile` and in the new `checkToBeDeleted` functions. The rest is just restructuring and fixes for existing tests.

### Testing

Three new unit tests.

### DIFF
--- a/lib/protocol/bep_extensions.go
+++ b/lib/protocol/bep_extensions.go
@@ -13,6 +13,7 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/syncthing/syncthing/lib/fs"
 	"github.com/syncthing/syncthing/lib/rand"
 )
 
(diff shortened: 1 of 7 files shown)


lib/protocol/bep_extensions.go
```diff
@@ -212,6 +213,26 @@ func (f FileInfo) isEquivalent(other FileInfo, ignorePerms bool, ignoreBlocks bo
 	return false
 }
 
+func FromFSFileInfo(fi fs.FileInfo, name string, filesystem fs.Filesystem) FileInfo {
```
calmh: I haven't reviewed anything yet, just happened to scroll to the bottom and saw this - i don't think we want to add a dependency on `fs` from `protocol`, so probably better to have this helper where it's used... or maybe scanner?

imsodin: Scanner probably makes sense - I actually wanted to use it there but didn't for diff size reasons :)

Issue #5198, opened 2018-09-15
DB refactor for cleaner design, enabling case insensitivity


calmh:
Currently our database "design" is roughly as follows:

- There is a `leveldb.DB` at the bottom. There's only one instance of this, and it just shuffles byte slice values to/from byte slice keys and iterates over them.
- There is our `db.Instance` which wraps the `leveldb.DB`. It contains all the logic for updating global sets, managing keys, updating indexes, and also keeps state for these things. There can be only one.
- There is the `db.FileSet` which is folder specific, has a pointer to the `db.Instance` and provides some high level operations and keeping state per folder. There is exactly one instance per folder.

I want to implement case insensitivity. In practice that mostly affects the key generation (i.e., how we map a given folder / device / file to a byte slice key in the database) and not the logic. Doing it in the current design means passing around a `caseInsensitive` bool everywhere though the `db.Instance` and sprinkling `if` cases over the key generation.  It would be cleaner if the key generation was a type / layer of its own. Making things case insensitive then mostly involves wrapping the key generator in a case folder and not changing any of the logic.

This supposed key generation layer belongs somewhere between the `db.Instance` and the `leveldb.DB`. But it depends on the indexes and state that are currently kept in the `db.Instance`. Usage wise, the case insensitivity is a per-folder thing, so it would be nice if the `db.FileSet` could just say "give me a new case insensitive db instance".

As a refactor, I'm imagining something like the following:

- At the bottom, the `leveldb.DB` as usual.
- Above this, a new thin `db.instance`. Its purpose is mainly to keep the device and folder indexes, which must be shared by all db users and is tied closely to the database. This is one-to-one with the `leveldb.DB`.
- Above this, a new `db.logic` (or whatever) type which is mostly everything that is `db.Instance` today. It has references to the `db.instance` and a new `db.keyer` interface type which does the key generation. By itself it doesn't keep any state that isn't per-folder (I think probably no state at all).
- At the top, the `db.FileSet`. Instead of just keeping a reference to today's `db.Instance` it will (upon creation) create a new `db.logic` with an appropriate `db.keyer` and reference to the `db.instance`.

There are also a few extra database-using types, like the `BlockMap`, `NamespacedKV`, `MTimeFS`. These should require only minor changes, wrapping the thin `db.instance` and maybe or maybe not taking a `keyer` (depending on whether they need its services).

I plan on doing this without changing any logic or introducing any functionality as step one. It will be probably be unreviewable anyway but one can try...

PR #5199, opened 2018-09-15
lib/db: Refactor key handling (ref #5198)

calmh:
This breaks out the key generation stuff into a separate type. It's cleaner on its own, and it prepares for future stuff.

### DIFF
--- a/lib/db/keyer.go
+++ b/lib/db/keyer.go
@@ -6,7 +6,35 @@
 
 package db
 
-import "encoding/binary"
+import (
+	"encoding/binary"
+)
+
+const (
+	keyPrefixLen   = 1
+	keyFolderLen   = 4 // indexed
+	keyDeviceLen   = 4 // indexed
+	keySequenceLen = 8
+	keyHashLen     = 32
+
+	maxInt64 int64 = 1<<63 - 1
+)
+
+const (
+	KeyTypeDevice          = 0
+	KeyTypeGlobal          = 1
+	KeyTypeBlock           = 2
+	KeyTypeDeviceStatistic = 3
+	KeyTypeFolderStatistic = 4
+	KeyTypeVirtualMtime    = 5
+	KeyTypeFolderIdx       = 6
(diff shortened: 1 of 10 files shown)


lib/db/keyer.go
```diff
@@ -0,0 +1,174 @@
...
+
+	// Folder metadata
+	GenerateFolderMetaKey(key, folder []byte) folderMetaKey
+}
+
+// defaultKeyer implements our key scheme. In needs folder and device
+// indexes.
+type defaultKeyer struct {
+	folderIdx *smallIndex
+	deviceIdx *smallIndex
```
AudriusButkevicius: Not sure if these are used anywhere outside of this new type, so perhaps the type should initialise these by itself?

calmh: So in principle you're right, at the moment. But in the future I want there to be (possibly) more than one keyer, and they must share the indexes. Hence the indexes must be injected and not created by the keyer itself.

lib/db/keyer.go
```diff
@@ -0,0 +1,202 @@
...
+
+	// index IDs
+	GenerateIndexIDKey(key, device, folder []byte) indexIDKey
+	DeviceFromIndexIDKey(key []byte) ([]byte, bool)
+
+	// Mtimes
+	GenerateMtimesKey(key, folder []byte) mtimesKey
+
+	// Folder metadata
+	GenerateFolderMetaKey(key, folder []byte) folderMetaKey
```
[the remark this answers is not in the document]
imsodin: Regarding methods: Totally makes sense, thanks for the explanation.

Regarding the constants (not relevant for this PR): The usage in cmd/stindex looks bad anyway (it reimplements some of the db code) and should instead be something like `db.String(key, value []byte) in my opinion. And the usage in lib/db is restricted to `...Statistics` types. In general I consider these implementation details that should not be accessed from outside (e.g. `db.NewNamespacedKV(ldb, db.KeyTypeDevice)` should not be possible, modifications would be a recipe for disaster).

PR #5200, opened 2018-09-16
all: Adjust windows perms in fs package

imsodin:
### Purpose

Currently `IsWindowsExecutable` is used in various places and `maskModePerm` too, but only in scanner. It is not respected in model/puller. This moves all these windows specific permission modifications to the filesystem package. such that `fs.FileInfo.Mode()` already returns the desired permissions.

### Testing

Nothing new.


imsodin:
@AudriusButkevicius I had to fix something else, because I didn't realize first that `FileMode` is a superset to permissions, and the commit thus dismissed your review.

PR #5201, opened 2018-09-16
lib/model: Extend shortcutFile

imsodin:
### Purpose

The function `shortcutFile` is used in just one place and handles just part of the "shortcutting" a file. Now it handles all of it, removing some code from the already long `handleFile` function, fixing `newError` being called twice on the same error and needing one db lookup less.

### Testing

Nothing new.


lib/model/folder_sendrecv.go
```diff
@@ -979,33 +979,7 @@ func (f *sendReceiveFolder) handleFile(file protocol.FileInfo, copyChan chan<- c
...
-			"action": "metadata",
-		})
-
-		if err != nil {
-			f.newError("shortcut", file.Name, err)
-		} else {
-			dbUpdateChan <- dbUpdateJob{file, dbUpdateShortcutFile}
-		}
-
-		return
```
calmh: !!!

PR #5202, opened 2018-09-16
lib/scanner: Centralise protocol.FileInfo creation

imsodin:
Mostly just refactoring to have less redundancy. Also there is a small fix to `scanner.CreateFileInfo`: Before it did set the `Permissions` and time fields of `FileInfo` for symlinks, which we do not actually want to record.

### DIFF
--- a/lib/scanner/walk.go
+++ b/lib/scanner/walk.go
@@ -330,16 +330,11 @@ func (w *walker) walkRegular(ctx context.Context, relPath string, info fs.FileIn
 	}
 
 	f, _ := CreateFileInfo(info, relPath, nil)
-	f = w.updateFileInfo(f, curFile)
+	f = w.updateFileInfo(f, curFile, hasCurFile)
 	f.NoPermissions = w.IgnorePerms
 	f.RawBlockSize = int32(blockSize)
 
 	if hasCurFile {
-		if runtime.GOOS == "windows" {
-			// If we have an existing index entry, copy the executable bits
-			// from there.
-			f.Permissions |= (curFile.Permissions & 0111)
-		}
 		if curFile.IsEquivalentOptional(f, w.IgnorePerms, true, w.LocalFlags) {
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


lib/scanner/walk.go
```diff
@@ -336,21 +329,17 @@ func (w *walker) walkRegular(ctx context.Context, relPath string, info fs.FileIn
...
-		RawBlockSize:  int32(blockSize),
-		LocalFlags:    w.LocalFlags,
-	}
+	f, _ := CreateFileInfo(info, relPath, nil)
+	f = w.updateFileInfo(f, curFile)
+	f.NoPermissions = w.IgnorePerms
+	f.RawBlockSize = int32(blockSize)
 
 	if hasCurFile {
+		if runtime.GOOS == "windows" {
```
calmh: could/should this happen in w.updateFileInfo?

imsodin: I tried to keep that function simple. This would require passing `hasCurFile` and conditions on that, the file type and goos == windows. Personally I'd rather not do that, but it doesn't really matter.

calmh: It already has curFile and the job to update the new based on the old, so it feels to me it would fit there.

lib/scanner/walk.go
```diff
@@ -515,8 +510,13 @@ func (w *walker) normalizePath(path string, info fs.FileInfo) (normPath string,
 	return normPath, false
 }
 
-// updateFileInfo updates walker specific members of protocol.FileInfo that do not depend on type
-func (w *walker) updateFileInfo(file, curFile protocol.FileInfo) protocol.FileInfo {
+// updateFileInfo updates walker specific members of protocol.FileInfo
+func (w *walker) updateFileInfo(file, curFile protocol.FileInfo, hasCurFile bool) protocol.FileInfo {
```
calmh: No need to pass the `hasCurFile` really. If we didn't have it, the permissions on the passed `curFile` would be zero and the `|` would be a no-op. But it doesn't matter greatly either way.

PR #5203, opened 2018-09-16
lib/model: Treat failed rename like del&update

imsodin:
This removes even more functionality from `renameFile` which already exists in `deleteFile`. It does so by proceeding with updating and deleting the files separately, if an error occurs during renaming (i.e. the error is now returned, not handled in `renameFile`).


imsodin:
I guess this flew under the radar?

lib/model/folder_sendrecv.go
```diff
@@ -900,36 +900,26 @@ func (f *sendReceiveFolder) renameFile(cur, source, target protocol.FileInfo, db
...
-		if err = f.performFinish(nil, target, curTarget, true, tempName, dbUpdateChan, scanChan); err != nil {
-			return
-		}
-	} else {
-		// We failed the rename so we have a source file that we still need to
-		// get rid of. Attempt to delete it instead so that we make *some*
-		// progress. The target is unhandled.
+	if err = f.performFinish(nil, target, curTarget, true, tempName, dbUpdateChan, scanChan); err != nil {
+		return err
+	}
```
AudriusButkevicius: This is a bit sketchy. The old file is gone the new file is there, yet chmod et al fails and db is not updated. As a minimum I suspect this will result in a conflict? What are the other failure scenarios?

AudriusButkevicius: In the old case perform finish failing would imply success which I guess is also wrong.

imsodin: In the old case failure was not really handled at all. Now there shouldn't be any conflicts: If rename failed, nothing happened. If performFinish fails it is the same as with regular pulls: A temp file remains. As the del and create are then handled separately, the temp file can even be reused (even if most likely the same error will happen again on performFinish). The deletion isn't a problem either, it will just notice that the file is already gone and send the appropriate db update.

AudriusButkevicius: A right, I forgot what performFinish does.

PR #5210, opened 2018-09-18
cmd/syncthing: Listen on UNIX socket (fixes #3616)

### LINKED ISSUE #3616: Disable GUI and make API available on unix socket
Context: running syncthing as a service on a headless server.

The GUI should not be started as it is not used at all. Syncthing is controlled through `systemd` and its configuration is managed with `ansible`. Every user account has its own instance of syncthing handling its files. All was great until I wanted to enable syncthing-inotify.

```
Sep 21 06:36:24 staging syncthing-inotify[529]: [WARNING] Cannot connect to Syncthing: Get https://localhost:8384/rest/404: dial tcp [::1]:8384: getsockopt: connection refused
```

It turns out services communicate with syncthing using the same channel as user's browsers. This means I cannot disable the GUI or I cannot enable any additional tool that talks to syncthing. I thought no problem, let's enable the GUI and mitigate the issue by only listening on a unix socket, which I can restrict access to through filesystem permissions. Alas, the GUI is hardcoded to use a tcp port.

Thus the feature request:
1. allow dissociating API and GUI channels. Or at the very least start a stripped down version with just the API available.
2. allow `unix://`-type address for GUI.

I believe those would be great steps towards allowing syncthing to work flawlessly in a traditional daemon style.

calmh:
This adds the ability to listen on (only) a UNIX socket. To use this functionality it's required to use the $STGUIADDRESS or -gui-address overrides. Specifically:

```
$ syncthing -gui-address unix:///var/run/syncthing.sock
```

If the scheme is `unix:`, plain HTTP will be spoken. If the scheme is `unixs:`, HTTPS will be spoken.

The socket will be created with permissions from umask. Anything else is the caller's responsibility.

### Testing

curl can load stuff:

```
$ curl -H X-API-Key:... --unix-socket /tmp/sock http://localhost/rest/system/status
{
  "alloc": 11990248,
  "connectionServiceStatus": {
    "dynamic+https://relays.syncthing.net/endpoint": {
      "lanAddresses": [
...
```

On Windows I'm sure hilarious things will happen. Haven't tested.

### Documentation

This should be added to the man page.


AudriusButkevicius:
Seems half-assed as you cannot configure it in the actual config...

calmh:
Possibly. But half assed might do it for the .1% who actually need this feature. It's also something that's not legitimately possible to tweak from within the UI with any expectation of being able to reconnect to the UI again...

calmh:
Maybe we should disable in-GUI editing of the GUI address whenever we are running with an override, as it has no effect. On the other hand that prevents fixing a broken GUI config by starting with an override and editing in the GUI. Maybe just a note in there that the settings are currently used.

calmh:
screenshot 2018-09-18 at 22 13 05

gui/default/syncthing/settings/settingsModalView.html
```diff
@@ -111,6 +111,10 @@
         <div id="settings-gui" class="tab-pane">
           <div class="form-group"  ng-class="{'has-error': settingsEditor.Address.$invalid && settingsEditor.Address.$dirty}">
             <label translate for="Address">GUI Listen Address</label>&emsp;<a href="https://docs.syncthing.net/users/guilisten.html" target="_blank"><span class="fas fa-question-circle"></span>&nbsp;<span translate>Help</span></a>
+            <p class="text-warning" ng-show="system.guiAddressOverridden">
+                <span class="fas fa-exclamation-triangle"></span>
+                <span translate>The GUI address is overridden by startup options. Changes here will not take effect.</span>
```
imsodin: Maybe something like "Changes will not take effect as long as these options are in effect."?

Ferroin: Agreed, it should be clear that the changes will be saved, and will take effect if the GUI address override is removed.

AudriusButkevicius:
If I choose to modify it and not be able to reconnect to the ui why is that a problem? Why do I have to use command line switched for that?

calmh:
Under what circumstances would that be a useful thing to do?

lib/config/guiconfiguration.go
```diff
@@ -31,6 +31,10 @@ func (c GUIConfiguration) IsAuthEnabled() bool {
 	return c.AuthMode == AuthModeLDAP || (len(c.User) > 0 && len(c.Password) > 0)
 }
 
+func (c GUIConfiguration) IsOverridden() bool {
+	return os.Getenv("STGUIADDRESS") != ""
+}
+
```
AudriusButkevicius: Is it not overriden if specified on the command line?

calmh:
OK, how about that. It gives us a config element `addressType` that can be set to `unix`. It can even be set via the GUI, it's just not really visible: if the listen address is set to an absolute file path, it will try to use a UNIX socket at that path.

### Testing

I successfully failed to connect to the GUI after changing the listen address to a UNIX socket.

Ferroin:
Also, just finished checking on one of my systems (I proxy Syncthing's WebUI through nginx already, so it's literally just a one-line change for me outside of Syncthing itself), and it appears to be working correctly.

calmh:
Actually, we don't even need the address type. If the configured address starts with a slash, it's a socket path. If it doesn't, it's the usual kind of address.

AudriusButkevicius:
Exactly. You use useTls as you did before. If net.ParseHost or whatever fails, assume unix socket (as you might want cwd relative sockets too).

calmh:
Not supported. We could, of course, but then the GUI must accept `172.0.0.1:8384/gui` as a valid unix socket, and I'm pretty sure that's not what the (confused, admittedly) user intended.

Ferroin:
Also, it's pretty atypical to use relative paths for UNIX sockets.  If you're using an absolute path, you always know exactly what socket you're connecting to, but you have no such guarantee when using a relative path.

AudriusButkevicius:
I was once deploying multiple postgres dbs on the same box and they were all fronted by a connection poller, so the poller would use a relative directory to the startup script to look for socket and postgres would do the same to create the socket, so a single script starts both and they all containes within the same directory.

Sure containers is the answer but that was not an option.

I imagine nginx + syncthing per user type of scenario.

Ferroin:
> I imagine nginx + syncthing per user type of scenario.

I can think of three ways to do this without using relative paths at all, none of which are difficult to implement (keep in mind, nginx has a sanely working include syntax for it's config files, and you only need one line referencing the socket), and all of which avoid being dependent on being started in a specific directory.

PR #5212, opened 2018-09-18
lib/db: Refactor: use a Lowlevel type underneath Instance

calmh:
This adds a thin type that holds the state associated with the
leveldb.DB, leaving the huge Instance type more or less stateless. Also
moves some keying stuff into the DB package so that other packages need
not know the keying specifics.

(This does not, yet, fix the cmd/stindex program, in order to keep the
diff size down. Hence the keying constants are still exported.)

No functional changes here.


lib/db/lowlevel.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+const (
+	dbMaxOpenFiles = 100
+	dbWriteBuffer  = 4 << 20
+)
+
+// Lowlevel is the lowest level database interface. It has a very simple
+// purpose: hold the actual *leveldb.DB database, and the in-memory state
+// that belong to that database. In the same way that a single on disk
```
imsodin: belongs

lib/db/leveldb_dbinstance.go
```diff
@@ -10,72 +10,41 @@ import (
...
 func OpenMemory() *Instance {
-	db, _ := leveldb.Open(storage.NewMemStorage(), nil)
-	ldb, _ := newDBInstance(db, "<memory>")
-	return ldb
+	db, _ := NewInstance(OpenMemoryLowlevel())
+	return db
 }
 
-func newDBInstance(db *leveldb.DB, location string) (*Instance, error) {
+func NewInstance(ll *Lowlevel) (*Instance, error) {
```
imsodin: Why is this exported now?

lib/db/set.go
```diff
@@ -324,10 +324,7 @@ func DropFolder(db *Instance, folder string) {
 	db.dropFolder([]byte(folder))
 	db.dropMtimes([]byte(folder))
 	db.dropFolderMeta([]byte(folder))
-	bm := &BlockMap{
-		db:     db,
-		folder: db.folderIdx.ID([]byte(folder)),
-	}
+	bm := NewBlockMap(db.Lowlevel, db.folderIdx.ID([]byte(folder)))
```
imsodin: Just `NewBlockMap(db.Lowlevel)` and then do the `folderIdx` stuff there, as it belongs to `Lowlevel`.

calmh:
So instead of getting into the discussion about why the Lowlevel was exported, I did some more work to better illustrate where this was heading. Sorry. The commit message needs some rewriting and this may be a bit "WIP" at the moment (plane hacking happened).

In principle, what goes on now is that the Lowlevel is the thing we expose as the database handle for outside the `db` package. The old `Instance` (now `instance`) contains the logic but is not exported any more.

This required some things to change - some methods on `Instance` became package level functions instead (dropping a folder, dropping delta indexes). 

I simplified the delta index dropping to always drop the entire index instead of local/remote separately. That was mostly out of lazyness - doing it the old way requires a `keyer` to properly extract the device ID for an entry. Might still want to undo this, I'm not 100% decided.

PR #5215, opened 2018-09-21
all: Display errors while scanning in web UI (fixes #4480)

### LINKED ISSUE #4480: Should show indication of local files/directories that can't be scanned for whatever reason
First of all: sorry if I missed any similar tickets, surprisingly I couldn't find any.

While debugging a syncthing-setup I noticed that lots of files were counted but not transmitted because they were created by root and unreadable by the user.
They were very much listable by the user tough, so I was pretty amazed that we don't already notify the user of such and similar problematic cases such as symlinks. 

I'm pretty sure the best way to signal this would be to change the folder status to "out of sync" with the same visual representation but a new status similar to "unsupported type or file/directory unreadable". I'd also like to change the color of the "downloading"status to something less alarming and reserve red as a clear warning color for this unfortunate state. In addition files/folders with such problems should be listed on top of the other out of sync files and the whole line should be permanently red, to underscore that the user has to do something about it.

Now there is the rare case of people who knowingly add such directories and still want to sync the other stuff apart from unreadable or symlinked files/folders. The ideal solution would be to keep a list of ignored paths per shared folder, which don't trigger this warning with a nice "ignore this path in the future" button next to the warning in the out of sync dialog.

But I'd be perfectly happy to have a general "don't show me such warnings for this shared folder" boolean for now, even though it doesn't cover the case of shared folders with known problematic paths, where other unsyncable files are unknowingly added later on.
In this case one will never be notified of the new problems, because these kinds of errors are completely ignored already.

imsodin:
### Purpose

As the discussed in #4480 there is currently a discrepancy between syncing errors, which are displayed in the UI, and scanning errors, which are silently dropped. In this PR I do not change how scanning works (e.g. no additional scans due to errors), but I do track errors that do occur and make them visible through folder summary (number) and a rest endpoint like `folder/pullerrors` for the actual errors.

As tracking these errors is clearly the folder's responsibility, I needed to move the scan code from `Model` to `folder`. In my opinion that's anyway where it belongs. I did not attempt to actually remove the overlap between `Model` and `folder` - not accessing `f.model` all over the folder code is a task for another day.

Discussion point:  
I added a new folder state "Scan Failed" which takes effect if the folder is not out of sync, but had failures on scan. Alternatively in this state the folder could just stay up-to-date or also use the out of sync state. Staying up to date would be the status quo, and make sense as in it is not know, whether the file we couldn't scan changed, i.e. we are up to date from a cluster perspective. Setting it to out of sync just means we count items we can't scan as out of sync, until we can and know for sure whether it is changed or not.

### Testing

Manual

### Screenshots

screenshot_2018-09-21_11-06-22
screenshot_2018-09-21_11-06-52

### Documentation

New rest endpoint to be documented


gui/default/syncthing/core/syncthingController.js
```diff
@@ -651,13 +652,27 @@ angular.module('syncthing.core')
...
+                return 0;
+            }
+            switch ($scope.failedType) {
+            case 'pull':
+                return $scope.model[$scope.failed.folder].pullErrors;
+            case 'scan':
+                return $scope.model[$scope.failed.folder].scanErrors;
+            }
+            return 0;
+        }
```
AudriusButkevicius: This feels like a hack, just have two separate structs to store the stuff you want to store and separate functions for actions (they can be curried I guess)

gui/default/syncthing/transfer/failedFilesModalView.html
```diff
@@ -1,11 +1,14 @@
-<modal id="failed" status="warning" icon="fas fa-exclamation-circle" heading="{{'Failed Items' | translate}}" large="yes" closeable="yes">
+<modal id="failed" status="warning" icon="fas fa-exclamation-circle" heading="Failed Items" large="yes" closeable="yes">
```
AudriusButkevicius: Why removed the translate?

lib/model/folder_sendonly.go
```diff
@@ -30,7 +31,7 @@ func newSendOnlyFolder(model *Model, cfg config.FolderConfiguration, _ versioner
 	return f
 }
 
-func (f *sendOnlyFolder) PullErrors() []FileError {
+func (f *sendOnlyFolder) PullErrors() []scanner.FileError {
```
AudriusButkevicius: This feels weird, PullErrors() comming from the scanner package.
Perhaps have protocol.FileError or somewhere where it makes more sense.

lib/model/folder_sendrecv.go
```diff
@@ -204,10 +204,10 @@ func (f *sendReceiveFolder) pull() bool {
 			// we're not making it. Probably there are write
 			// errors preventing us. Flag this with a warning and
 			// wait a bit longer before retrying.
-			if folderErrors := f.PullErrors(); len(folderErrors) > 0 {
+			if folderPullErrors := f.PullErrors(); len(folderPullErrors) > 0 {
 				events.Default.Log(events.FolderErrors, map[string]interface{}{
 					"folder": f.folderID,
-					"errors": folderErrors,
+					"errors": folderPullErrors,
```
AudriusButkevicius: This is not nice, folders errors which is generic returns pull errors only.
Perhaps we should enrich FileError to define error type, rename APIs to talk about folder/errors rather than folder/pullErrors and have a single consolidated UI with a single consolidated set of errors for both pull and download, which gets rid of all the switch if hacks in the UI and simplifies it there.

lib/connections/limiter.go
```diff
@@ -134,36 +134,38 @@ func (lim *limiter) CommitConfiguration(from, to config.Configuration) bool {
...
-	if to.Options.LimitBandwidthInLan {
-		l.Infoln("Rate limits apply to LAN connections")
-	} else {
-		l.Infoln("Rate limits do not apply to LAN connections")
+	if limited {
+		if to.Options.LimitBandwidthInLan {
+			l.Infoln("Rate limits apply to LAN connections")
+		} else {
+			l.Infoln("Rate limits do not apply to LAN connections")
+		}
```
AudriusButkevicius: I don't disagree with the change, but should be a separate PR

lib/model/folder.go
```diff
@@ -262,14 +271,231 @@ func (f *folder) getHealthError() error {
 }
 
 func (f *folder) scanSubdirs(subDirs []string) error {
-	if err := f.model.internalScanFolderSubdirs(f.ctx, f.folderID, subDirs, f.localFlags); err != nil {
-		// Potentially sets the error twice, once in the scanner just
-		// by doing a check, and once here, if the error returned is
-		// the same one as returned by CheckHealth, though
-		// duplicate set is handled by setError.
+	if err := f.CheckHealth(); err != nil {
```
AudriusButkevicius: This change is very confusing, I guess that's just unwinding a single stack level? Or is it doing more as the diff is huge, hence it's hard to tell.

lib/connections/limiter.go
imsodin: Agree, that somehow mistakenly slipped into this PR. Filed a separate one: #5242

lib/model/folder.go
imsodin: It moves the code from `Model.internalScanFolderSubdirs` (https://github.com/syncthing/syncthing/pull/5215/files#diff-1df9805b2381b6288b220226f82d5dc9L1968) into `folder.scanSubdirs`. There are no functional changes, but a few things had to be adjusted to the new underlying type. I could rebase the PR making a first commit, where I just copy the function over (won't build) and a second commit, where I make it work - would that help?

imsodin:
Having just a single type of file error is a similar question to what I brought up in the original post about folder state. The more I think about it, the better I like it: From a dev perspective, whether the error occurs at scanning or pulling is a clear distinction (however there are even links there too), but from a user perspective it really doesn't matter. Something didn't work with regards to a file, the error explains what. If the error isn't clear enough, then that's not a problem of not separating scan/pull, but making the error (context) more descriptive. In terms of rest compatibility: Nothing will change, just more errors will be reported. I wouldn't even add a new field, but even if we didn't, that wouldn't break backwards compatibility.

So unless @calmh has objections, I would go ahead and integrate the "new scan errors" into the existing `FolderErrors` API.

calmh:
mnee7

imsodin:
I didn't remember that the rest endpoint was inconsistently called `/rest/folder/pullerrors` (events is `FileErrors`). So I added `/rest/folder/fileerrors` and "marked" the pull one as deprecated. Also needs adjustments in the docs.

AudriusButkevicius:
I think just drop the endpoint all together, I'd bet my socks that there are very few users of this, if any. (@canton7 @kozec)

imsodin:
This completely slipped from my radar. Given it's been almost a month since @canton7 and @kozec were pinged I think it's save to assume they are not using it.  
@calmh Do you have an opinion on whether to drop or keep `/rest/folder/pullerrors`?

imsodin:
So I dropped `/rest/folder/pullerrors`. I'd like to merge this before the next RC, it's lying around for pretty long already. I will take care of the docs once merged.

@calmh In my opinion dropping the endpoint needs to be mentioned in the changelog. Should I create a dummy issue like "Replace REST Api endpoint `/rest/folder/pullerrors` with `/rest/folder/fileerrors` or is a "manual comment" in the release notes more better?

cmd/syncthing/gui.go
```diff
@@ -276,7 +276,7 @@ func (s *apiService) Serve() {
 	getRestMux.HandleFunc("/rest/db/status", s.getDBStatus)                      // folder
 	getRestMux.HandleFunc("/rest/db/browse", s.getDBBrowse)                      // folder [prefix] [dirsonly] [levels]
 	getRestMux.HandleFunc("/rest/folder/versions", s.getFolderVersions)          // folder
-	getRestMux.HandleFunc("/rest/folder/pullerrors", s.getPullErrors)            // folder
+	getRestMux.HandleFunc("/rest/folder/fileerrors", s.getFileErrors)            // folder
```
calmh: As we're anyway cleaning up and clarifying, why not just `/rest/folder/errors`? The intention is to cover all sorts of errors that happen in any phase of the sync operation right?

calmh:
RC time is the 13th so there's plenty of time for this to make it.

cmd/syncthing/gui.go
imsodin: Yes, it covers errors at any point during scanning or syncing associated with a file (item). Errors "on the folder as a whole" (e.g. watcher error, home disk full) are not reported on this endpoint. That was my intention of leaving "file" in there, but trying to represent this tiny distinction in the name is possibly more confusing than helpful.

calmh:
(build fail)

lib/model/folder.go
```diff
@@ -430,3 +680,73 @@ func (f *folder) basePause() time.Duration {
...
+		}
+		filtered = append(filtered, fe)
+	}
+	f.scanErrors = filtered
+}
+
+func (f *folder) Errors() []FileError {
+	f.scanErrorsMut.Lock()
+	defer f.scanErrorsMut.Unlock()
+	return f.scanErrors
```
calmh: This isn't great because the backing array can get modified later by another goroutine. Needs to return a copy to be goroutine safe.

lib/scanner/walk.go
```diff
@@ -226,6 +239,8 @@ func (w *walker) walkAndHashFiles(ctx context.Context, fchan, dchan chan protoco
 
 		if !utf8.ValidString(path) {
 			l.Warnf("File name %q is not in UTF8 encoding; skipping.", path)
+			// Purposely do not send a ScanResult with an error, as this can never
+			// be recovered, as this invalid path can never be succesfully scanned.
```
calmh: Not clear on the reasoning here. Isn't this similar to any other scan error like permission denied - something we flag and the user can fix?

acolomb:
> So I dropped `/rest/folder/pullerrors`. I'd like to merge this before the next RC, it's lying around for pretty long already. I will take care of the docs once merged.

Seems that never happened @imsodin? :wink:  I'm currently trying to fill in all blanks in the API docs, PR incoming soon.

PR #5216, opened 2018-09-21
lib/model, lib/protocol: Handle request concurrency in model

imsodin:
This is a successor of #5123 where it was suggested, that the concurrency of request (and thus later the limiting) should not be handled in the protocol package, but the model. This is a prototype of doing this by letting the model call a `connection.Response` function instead of directly returning the data from disk.

This PR builds, but tests are not adapted yet. The limiting isn't implemented either, but there are comments showing where it would happen. I'd like to know whether this approach seems sensible before carrying on.


lib/model/model.go
```diff
@@ -1323,6 +1325,32 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset
...
+
+	go m.request(conn, res, folderCfg, deviceID, folder, name, size, offset, hash, weakHash, fromTemporary)
+}
+
+func (m *Model) requestPreparation(folderFs fs.Filesystem, folderIgnores *ignore.Matcher, deviceID protocol.DeviceID, folder, name string, size int32, offset int64, fromTemporary bool) error {
+	if offset < 0 {
+		return protocol.ErrInvalid
+	}
+
+	if cfg, ok := m.cfg.Folder(folder); !ok || !cfg.SharedWith(deviceID) {
```
calmh: The calling method already has a folderCfg that it could give us, instead of us going digging in the config again.

lib/model/model.go
```diff
@@ -1334,26 +1362,38 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset
...
 		return protocol.ErrNoSuchFile
 	}
 
+	return nil
+}
+
+func (m *Model) request(conn protocol.Connection, res protocol.RequestResult, folderCfg config.FolderConfiguration, deviceID protocol.DeviceID, folder, name string, size int32, offset int64, hash []byte, weakHash uint32, fromTemporary bool) {
+	defer func() {
+		conn.Response(res)
+		<-res.Done
```
calmh: I don't know what you're trying to do with Done here.

Apart from that, the approach seems sensible. The other way that strikes me is that the request method could return a future of sorts (`struct { Data <-chan []byte, Error <-chan error }` perhaps) that the protocol layer would launch a goroutine to wait on and do the actual response message creation, handling the request ID etc. It's not super clean, imho, that the model layer needs to know about and handle this.

imsodin: I think Done is a future. It uses a mechanism that was already used in the protocol package, that notifies the creator of the object (here 'protocol.RequestResult`) when it is finished handling the contained data (i.e. the taken bytes can be given back to the semaphore).

Together with the actual limiting that I added, does this make sense now?

imsodin:
> This does allocations on every request, which supposedly was a big pain point in the past...

I can use some kind of pool again, but the way pooling was done currently was "bad" according to multiple posts I read about Go's pool: It shouldn't be used with variable buffer sizes. Anyway, that's not what I intend to discuss with this PR, it's just an outline for how the request concurrency should work. Lets discuss and agree on that first.

The callback aka `connection.Response` is there to remove the necessity to spawn goroutines in protocol, due to @calmh's comment (https://github.com/syncthing/syncthing/pull/5123#issuecomment-414053692):

> Another stray thought here... Having the protocol package go the request messages might not be the right thing. Removing that and having the model do the needful to do I/O in the background after applying limiting etc might be better.

Both ways are doable with limiting in the model and I don't see any clear advantages of one way over the other.

imsodin:
@AudriusButkevicius I don't know and I don't really understand the different opinions here, but letting `Model.Request` return the data directly and spawning goroutines in the protocol does make the diff much smaller: https://github.com/imsodin/syncthing/commit/7d3e443afec5d1cff842c42853ee1b741dc85c53  
Given I still don't see any hard arguments either way, I'd be in favor of that. What do you think about the approach in the commit above (just to check before I turn around yet another PR, allocations not yet considered)?

AudriusButkevicius:
I still think its sucky... We started passing buf from protocol to model because protocol knew when the buffer is no longer is needed, now we are passing a done chan to the protocol to tell us the same thing so that we could give back to the limiter in the model. Both if these things should live in the same layeras both track something related to the payload size. Request should return a response, and protocol should close the response/done the context when done sending stuff down the socket, returning stuff to the byte pool and limiter (potentially both now living in the model).

imsodin:
@AudriusButkevicius I don't know and I don't really understand the different opinions here, but letting `Model.Request` return the data directly and spawning goroutines in the protocol does make the diff much smaller: https://github.com/imsodin/syncthing/commit/7d3e443afec5d1cff842c42853ee1b741dc85c53  
Given I still don't see any hard arguments either way, I'd be in favor of that. What do you think about the approach in the commit above (just to check before I turn around yet another PR, allocations not yet considered)?

imsodin:
> We started passing buf from protocol to model because protocol knew when the buffer is no longer is needed, now we are passing a done chan to the protocol to tell us the same thing so that we could give back to the limiter in the model. Both if these things should live in the same layeras both track something related to the payload size. 

That's why it all lives in the model now. The problem is, you can't allocate a buffer without access to the limiter. So if you want to create the buffer in protocol as before, you need a limiter in protocol. That was my first PR which was shut down, because limiter is supposed to live in model.

So I guess the main question to answer is: Where should buffer allocation and thus limiting happen - in protocol or model?

imsodin:
I am confused: Thant sounds exactly like what I did here: https://github.com/imsodin/syncthing/commit/7d3e443afec5d1cff842c42853ee1b741dc85c53: Limiter in the model, protocol signals that it's done with the buffer by closing a channel (I can make that a callback/context/whatever, channel just seemed the simplest option). How is that different from what you imagine (except for the missing pool, but as said, that's still to be done).

AudriusButkevicius:
This is ok as an idea, I don't like the implementation returning a channel, and there not being a pool.
It should return a response/context which should be closable, doneable or something like that.

imsodin:
@AudriusButkevicius I added a `protocol.RequestResult` type which is doneable and does its cleanup (limiter and pool) itself. Does that look better?

The pool consists of one pool per block size because apparently `sync.Pool` shouldn't be used with objects of varying sizes: https://github.com/golang/go/issues/23199

lib/model/folder_sendrecv.go
```diff
@@ -1158,7 +1157,8 @@ func (f *sendReceiveFolder) shortcutFile(file, curFile protocol.FileInfo, dbUpda
 // copierRoutine reads copierStates until the in channel closes and performs
 // the relevant copies when possible, or passes it to the puller routine.
 func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan chan<- pullBlockState, out chan<- *sharedPullerState) {
-	buf := make([]byte, protocol.MinBlockSize)
+	buf := protocol.GetBuf(protocol.MinBlockSize)
+	defer protocol.PutBuf(buf)
```
AudriusButkevicius: We sure buf is not captured at the point of defer?

lib/model/model.go
```diff
@@ -55,6 +55,10 @@ const (
 	maxBatchSizeFiles = 1000       // Either way, don't include more files than this
 )
 
+// maxRequestBytes is the maximal amount of bytes that may be requested by
+// remotes at the same time.
+var maxRequestBytes = 2 * protocol.MaxBlockSize
```
AudriusButkevicius: Should be a config option perhaps?

lib/protocol/bufferpool.go
```diff
@@ -4,27 +4,67 @@ package protocol
 
 import "sync"
 
+// Global pool to get buffers from. Requires Blocksizes to be initialised,
+// therefore it is initialized in the same init() as BlockSizes
+var thePool bufferPool
+
+// GetBuf returns a byte slice with the requested size from the global pool
+func GetBuf(size int) []byte {
```
AudriusButkevicius: I'd have a `var protocol.BufferPool = &bufferPool{}` that implements some interface, and use protocol.BufferPool.GetBuffer() etc.

Also, we could just have a separate pool in the model package and separate pool in protocol package (moving the pool implementation to utils or something like that), so that we don't have this ugly reaching into the protocol package. (Or move the cruft into its own package?)

imsodin: I was thinking about the scope of pools, and came up with no good reason to restrict it: The more one pool is used, the higher the chances to save some allocations.

I originally implemented it in the util package, but then I can't base the bucketing directly on block sizes. I can of course redefine them there.

lib/model/folder_sendrecv.go
imsodin: It's just used to transfer data from one file to another. After the routine, nothing happens with it anymore.

AudriusButkevicius: My point is that there the buf too small switcheroo with a new buf, and I suspect the old buf gets put back twice, and new one is left dangling.

lib/protocol/bufferpool.go
AudriusButkevicius: @calmh ?

lib/model/folder_sendrecv.go
AudriusButkevicius: Probably, I'd try it in the playground to see qhat actually happens.

imsodin: I did try before committing :)  
As far as I understand all arguments to the deferred function are evaluated when the defer statement is executed. So putting the function inside an anonymous function prevents that evaluation.

lib/model/model.go
```diff
@@ -1339,37 +1357,58 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, offset
...
 	}
 
+	// Restrict parallel requests by connection/device
+
+	m.pmut.RLock()
+	limiter := m.connRequestLimiters[deviceID]
+	m.pmut.RUnlock()
+
+	if limiter != nil {
+		limiter.take(int(size))
```
xjtdy888: We have passed the limiter to the NewRequestResult, why not take the consumption in the NewRequestResult.
And we ```r.limiter.give(len(r.data))``` is also already inside RequestResult

AudriusButkevicius: I've asked not to do it in the request as it feels hidden away. Ideally we would not give back in the response too, but I don't see how to do it other than adding a generic onDone callbaxk registration.

imsodin:
@calmh Please have a look at the change I made to `byteSemaphore`. It now doesn't panic anymore, if a buffer bigger than the max size is taken/given back. Instead it just treats it like a request for the max size, thus limiting it to one such request at the time: https://github.com/syncthing/syncthing/pull/5216/commits/87706229769baf4192303462e98e4c6c22996aa6#diff-2c1ceee49dfbbba5797511c2f21dcb0b

@AudriusButkevicius The test would probably check that with a sufficiently small limit, the second request won't return until the first one does? The problem is, there is just a "racy" distinction between two requests returned in parallel and two request after each other. I'd somehow have to abstract the part of `Request` after acquiring the bytes from the limiter and mock that for the test. That just seems very involved. Do you have a better idea? Or an idea, how what I described can be done easily - afaik you can't replace member functions at runtime.

calmh:
I approve of that change. In fact, it enables something I've wanted recently - puller max pending < max block size, for the cases where we want a low number and won't be using 2 meg blocks.

AudriusButkevicius:
I would set the limit to X, make two requests of X, expect second one to return only after first one finishes.

imsodin:
@AudriusButkevicius Thanks for insisting, it was actually super easy: I can (should be able to) control the return time of the second call using the `Done` method of the result of the first...

In the process I realized that it's actually pretty ugly to return a non-nil result when there is an error - so I changed that as well.

lib/model/model.go
```diff
@@ -1319,23 +1319,25 @@ func (m *Model) closeLocked(device protocol.DeviceID) {
 	closeRawConn(conn)
 }
 
-type RequestResult struct {
+// Implements protocol.RequestResult
+type requestResult struct {
```
AudriusButkevicius: Call it response?

lib/model/model.go
```diff
@@ -1319,23 +1319,25 @@ func (m *Model) closeLocked(device protocol.DeviceID) {
...
 }
 
-func (r *RequestResult) Data() []byte {
+func (r *requestResult) Data() []byte {
 	return r.data
 }
 
-func (r *RequestResult) Done() {
+// Returns the byte slice back to the pool and releases the bytes to the limiter.
+func (r *requestResult) Done() {
```
AudriusButkevicius: Call close?

I am taking this from python requests library, where request returns a response and response needs to be closed to release resources.

lib/model/model.go
```diff
@@ -1430,18 +1439,23 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, size in
 	if info, err := folderFs.Lstat(name); err != nil || !info.IsRegular() {
 		// Reject reads for anything that doesn't exist or is something
 		// other than a regular file.
-		return res, protocol.ErrNoSuchFile
+		l.Debugln(folderFs.URI())
```
AudriusButkevicius: Wut?

lib/model/model.go
```diff
@@ -1407,7 +1408,14 @@ func (m *Model) Request(deviceID protocol.DeviceID, folder, name string, size in
...
 	}
 
-	res := NewRequestResult(int(size), limiter)
+	// The requestResult releases the bytes to the limiter when its Done method is called.
+	res := newRequestResult(int(size), limiter)
+	defer func() {
+		// Done it ourselves if it isn't returned due to an error
+		if err != nil {
+			res.Done()
+		}
```
AudriusButkevicius: I'd make done/close be a noop if called multiple times, as tracking who called it where is hard.

imsodin: Done - doesn't really influence this part though (?).

imsodin:
> I trust you've actually run this a little bit, maybe even with the race detector, and it doesn't like crash or sync the wrong data?

I have only run it on a pretty small test setup, done some transfers and now also validated that the transferred data is actually the same (don't see how it would even be achievable to do a transfer of wrong data without any errors). I haven't run it on my bigger production folders though.

lib/model/model.go
```diff
@@ -1316,19 +1319,43 @@ func (m *Model) closeLocked(device protocol.DeviceID) {
...
+
+// Returns the byte slice back to the pool and releases the bytes to the limiter.
+func (r *requestResponse) Close() {
+	select {
+	case <-r.closed:
+		return
+	default:
+	}
+	protocol.BufferPool.Put(r.data)
+	close(r.closed)
```
AudriusButkevicius: I'd move this in the default case, as this can be racy.

imsodin: Good catch...  
How will moving it into the default clause make it not racy though?

AudriusButkevicius: it would not make it not racy I guess, just makes it less likely

AudriusButkevicius:
You could just us an atomic and a compare and swap, instead of a lock + channel.

If you stay with a lock then just having a boolean instead of a channel is easier.
Or you could replace the whole thing with a sync.Once

lib/model/model_test.go
```diff
@@ -3948,3 +3946,46 @@ func (c *alwaysChanged) Seen(fs fs.Filesystem, name string) bool {
 func (c *alwaysChanged) Changed() bool {
 	return true
 }
+
+func TestRequestLimit(t *testing.T) {
```
calmh: I don't understand how this test works. Explain?

imsodin: ... Yes true, this is racy :|  
The idea is that the returned object from `Request` must be closed before a subsequent call to `Request` can return. So I check that before closing, the second `Request` is still blocking, and after closing it returns. However as you say, that may equally be satisfied by a sufficiently long execution time of `Request`.

I could introduce a delay after the second `Request`. It's still racy, but maybe good enough? At least even the current version of the test without a delay does fail without the changes in this PR on my system. And the test failure/flakyness modes are false negatives, not false positives, which is a bit less annoying.

imsodin: @calmh What do you think about the changed test?  
I added a delay of 10 times the duration of the first request, which under normal circumstances should be enough to make sure the second request has time enough to terminate (in general the second request for the same file should even be faster).

calmh:
LGTM from a quick reading, merge unless @AudriusButkevicius has something unresolved?

lib/model/model.go
```diff
@@ -1619,6 +1680,11 @@ func (m *Model) AddConnection(conn connections.Connection, hello protocol.HelloR
 	m.conn[deviceID] = conn
 	m.closed[deviceID] = make(chan struct{})
 	m.deviceDownloads[deviceID] = newDeviceDownloadState()
+	if device.MaxRequestKiB == 0 {
+		m.connRequestLimiters[deviceID] = newByteSemaphore(2 * protocol.MaxBlockSize)
```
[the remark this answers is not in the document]
imsodin: I wouldn't know why, the configurable is `MaxRequestKiB`, this is just a default when set to 0.  
I just realized I am not handling the case where `MaxRequestKiB < 0`: I guess that could mean don't limit at all - does that make sense?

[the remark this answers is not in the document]
AudriusButkevicius: I guess, as a side effect allows us to disable this whole thing, which is nice.

AudriusButkevicius:
You have my blessing child

PR #5217, opened 2018-09-21
Suggest setting up auth on initial startup (fixes #4703)

### LINKED ISSUE #4703: Suggest setting up auth on initial startup

Relates to #3357.

bahadir60:
### Purpose

Suggest user for setting `User` and `Password` in GUI Authentication, if not set. 
Suggest setting up auth on initial startup (fixes #4703)

### Testing
None

### Screenshots

If `User` and `Password` not set, initial startup this information panel shows up.

screen shot 2018-09-21 at 20 01 06

Clicking: - `OK` closes panel. And panel will shows up again 1 weeks later in loading (if user and password not set).
               - `Settings` opens settings for setting relevant boxes.
After setting user and password, panel will shows off. Except cleaning user and password, panel never shows up again.

Localstroge key `authWarningDismissed` holds `true` and `false` and `authWarningDismissed` holds expiration time when it set clicking `OK` button. 

### Documentation

Not needed.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -50,6 +50,7 @@ angular.module('syncthing.core')
         $scope.themes = [];
         $scope.globalChangeEvents = {};
         $scope.metricRates = false;
+        $scope.authWarningDismissed = true;
         $scope.folderPathErrors = {};
         $scope.currentFolder = {};
         resetRemoteNeed();
@@ -57,6 +58,12 @@ angular.module('syncthing.core')
         try {
             $scope.metricRates = (window.localStorage["metricRates"] == "true");
         } catch (exception) { }
+
+        try {
+            if(typeof window.localStorage["authWarningDismissed"] != 'undefined'){
+                $scope.authWarningDismissed = (window.localStorage["authWarningDismissed"] == "true");
+            }
+          } catch (exception) { }
 
         $scope.folderDefaults = {
             selectedDevices: {},
@@ -395,6 +402,9 @@ angular.module('syncthing.core')
                 && guiCfg.authMode !== 'ldap'
                 && !guiCfg.insecureAdminAccess;
 
+            $scope.noAuth = ($scope.authWarningDismissed || checkAuthDate())
+                && (!guiCfg.user || !guiCfg.password);
+
             if (!hasConfig) {
                 $scope.$emit('ConfigLoaded');
             }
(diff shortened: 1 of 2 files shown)


imsodin:
I don't think this is disabled at all, i.e. it will display whenever you open the web UI or refresh. That's just plain annoying, this needs to be permanently dismissable. Also you shouldn't use `guiCfg` within some function, but the appropriate object in `$scope`. And then I'd have a few comments on the actual text, but lets sort the "functional" stuff first.

bahadir60:
Definitely yes, I rearrange function. Thanks for comment.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -395,11 +395,80 @@ angular.module('syncthing.core')
...
+
             if (!hasConfig) {
                 $scope.$emit('ConfigLoaded');
             }
         }
 
+        function initAuthorization(guiCfg) {
+            var userPwd = !guiCfg.user || !guiCfg.password;
+
+            if (userPwd === true){
```
calmh: There is logic oddness here. If we don't have `initNotification` set, we set it to one. Then, if it's set to one (which it is), we set it to zero and return `userPwd` (which we've already confirmed is `true`). I think something else is intended?

I would expect a local storage variable `authenticationWarningDismissed` or something like that. If it's set we don't show the warning. When the user dismisses the dialog we set it to one. Nothing more needed, really, I think.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -395,11 +395,80 @@ angular.module('syncthing.core')
...
+                localStorage.removeItem(key);
+                localStorage.removeItem(key + '_expiresTime');
+            } catch(e) {
+                console.log('removeStorage: Error removing key ['+ key + '] from localStorage: ' + JSON.stringify(e) );
+                return false;
+            }
+            return true;
+        }
+
+        function getStorage(key) {
```
calmh: I think these timed storage things are overkill. We have other places in here that gets and sets local storage variables, essentially just

```
try {
  window.localStorage[key] = val;
} catch (exception) { }
```

and

```
try {
  val = window.localStorage[key];
} catch (exception) { }
```

bahadir60: Thanx, I made some changes base on your comments.

st-review:
:robot: *beep boop*

PR #5218, opened 2018-09-21
Switch "Upgrade To" to style used in every translated file

KeyWeeUsr:
Note: This is really a stupid PR, but the visual appearance just makes me frown.

It's quite a good way to notify users about the update, but optically it doesn't look right to me:

image
 
"To" itself with the upcase is really annoying in there to my eyes. I tried to look up even how to remove the `v` from `v0.14.50` but without success.

image

additionally highlighting the version is also better for optically separating it, so that it doesn't appear like one word or "Tov"/"Tor" and similar stuff.

image

If this change is welcome, please point me to something that would help me removing "v" from the version. Most likely even adding `<b>` tag to the version string alone could be a quick fix instead of rewriting the .JSON translations. Unless the version string is used anywhere else as a condition or something.

### DIFF
--- a/gui/default/assets/lang/lang-en.json
+++ b/gui/default/assets/lang/lang-en.json
@@ -334,7 +334,7 @@
    "Up to Date": "Up to Date",
    "Updated": "Updated",
    "Upgrade": "Upgrade",
-   "Upgrade To {%version%}": "Upgrade To {{version}}",
+   "Upgrade To {%version%}": "Upgrade to {{version}}",
    "Upgrading": "Upgrading",
    "Upload Rate": "Upload Rate",
    "Uptime": "Uptime",


calmh:
Currently it follows title case as everywhere else in the UI. Nothing wrong with the "v" either imho.

KeyWeeUsr:
Hmmm, so the translations have the wrong case in the Upgrade button? Danish/Czech/German/Russian/... are all "Upgrade to <version>" instead of "Upgrade To". Seems ugly to me as the "Upgrade to ..." is a sentence  unlike e.g. "Minimum Free Disk Space" in the settings, but that's just my taste.

Also, UK translation is "Upgrade to", that's why I made a PR, because it seemed as a bug in the translation too. Anyway, I do see it's rather a dumb thing to pick for "fixing", so feel free to close if you think the text is ok. :)

calmh:
Other languages are not English and can do their own thing. Nonetheless the place to fix it is in the html source, which is also the translation key and will break all existing translations of this. So although I’m leaning towards you being right, I think the pain of fixing it exceeds the utility, sorry.

PR #5222, opened 2018-09-25
DEBUG ONLY: add some debug info

imsodin:
This PR just exists to create debug builds to investigate https://forum.syncthing.net/t/panic-runtime-error/12215.


AudriusButkevicius:
close?

Issue #597, opened 2014-08-29
identical files are marked as "Out of sync"


seidler2547:
I have a PC and a NAS, and I wanted to sync a larger directory from PC to NAS. To speed up the initial sync, I rsync'ed the directory to the NAS. I set the PC to readonly (Master Repository). I added both directories to both nodes. Syncthing keeps complaining about ~90% "Out of Sync" on the master node.
 When I open the "Out of Sync Items", it shows Directories as "Update" and other files as "Sync with a downward arrow".

Just to make sure, here is one of the files that are marked as "out of sync":
#### PC:

```
[stefan@abraham Personal]$ ls -ln 'Church/Publicity/Organisation/Rotas/Rotas 2007/Service Rota July_Aug_2007.xls'
-rw-r--r-- 1 1000 100 19456 Jul  3  2007 Church/Publicity/Organisation/Rotas/Rotas 2007/Service Rota July_Aug_2007.xls
[stefan@abraham Personal]$ sha512sum 'Church/Publicity/Organisation/Rotas/Rotas 2007/Service Rota July_Aug_2007.xls'
cce7bb372c767ff4641351003d4740ecec45a845cca787881f603f479e92ee66cb7801bd7b2183f34167ea8ce6169f75fb42fc2ed541069e9ae2fecef7a475ce  Church/Publicity/Organisation/Rotas/Rotas 2007/Service Rota July_Aug_2007.xls
```
#### NAS:

```
root@marsboard:/media/data/Backups/BTSync/abraham.Personal# ls -ln 'Church/Publicity/Organisation/Rotas/Rotas 2007/Service Rota July_Aug_2007.xls'
-rw-r--r-- 1 1000 100 19456 Jul  3  2007 Church/Publicity/Organisation/Rotas/Rotas 2007/Service Rota July_Aug_2007.xls
root@marsboard:/media/data/Backups/BTSync/abraham.Personal# sha512sum 'Church/Publicity/Organisation/Rotas/Rotas 2007/Service Rota July_Aug_2007.xls'
cce7bb372c767ff4641351003d4740ecec45a845cca787881f603f479e92ee66cb7801bd7b2183f34167ea8ce6169f75fb42fc2ed541069e9ae2fecef7a475ce  Church/Publicity/Organisation/Rotas/Rotas 2007/Service Rota July_Aug_2007.xls
```

I tried restarting, removing and re-adding the repos, but nothing helps.

janlimpens:
[...]

PR #5230, opened 2018-09-26
lib/db, lib/model: Resolve identical recv only items (fixes #5130)

### LINKED ISSUE #5130: "Revert Local Changes" button stays when the file causing it is deleted
This is v0.14.50 RC material, but it's rather minor. (@calmh)

Reproducer:

1. Create receive only folder.
2. Create any file, wait for scan such that the reversion button appears.
3. Delete the file, wait for scan such that the local state goes down to zero.

Result: "Revert Local Changes" button still visible.
Expected: "Revert Local Changes" button disappears, as the is no real difference.

imsodin:
### Purpose

For send-only folders we "pull" files and merge them with the global ones, if they do not actually differ (i.e. just version differs). For receive-only folders, we can't do this by "pulling", as the files in question aren't needed. Therefore updates from scanning are compared to the global state and if they match, they are merged.

The test I wrote to check that it works as intended uncovered another problem: Invalid files with higher versions than valid ones "shadow" those in the global list. Therefore I changed the ordering to always move invalid files to the back of the global list. 

### Testing

I added a test checking that creation&deletion and modification&undo operations result in no receive only file infos. For the ordering of the global list with invalid infos I added a simple scenario that illustrates the problem.


calmh:
> problem: Invalid files with higher versions than valid ones "shadow" those in the global list.

I though this was a feature? Honestly I struggle to think of the original reason right now, but if this is not the case I'm not sure we need invalid files in the global list at all? We didn't use to have them there, then added them... Something about ignore handling I'm sure.

imsodin:
The feature is exactly, that ignored/invalid files are present in the global list. This allows to discern them from files that the remote doesn't have (the latter should be accounted). The position of invalid files is irrelevant for this purpose: If the version of an entry for a particular device is greater or equal to the global version (first one in the list), then that device doesn't need the file.

AudriusButkevicius:
database stuff
image

calmh:
This feels like it might address #597 at the same time?

lib/model/model.go
```diff
@@ -2042,14 +2042,38 @@ func (m *Model) internalScanFolderSubdirs(ctx context.Context, folder string, su
...
+				case gf.IsEquivalentOptional(fs[i], false, false, protocol.FlagLocalReceiveOnly):
+					// What we have locally is equivalent to the global file.
+					fs[i].Version = fs[i].Version.Merge(gf.Version)
+					fallthrough
+				case fs[i].IsDeleted() && gf.IsReceiveOnlyChanged():
+					// Our item is deleted and the global item is our own
+					// receive only file. We can't delete file infos, so
+					// we just pretend it is a normal deleted file (nobody
+					// cares about that).
+					fs[i].LocalFlags &^= protocol.FlagLocalReceiveOnly
```
calmh: Wait.

1. There is a normal file. It's in sync everywhere.
2. We, a receive only device, change the file. We set the receive only changed bit and update the version. This is the global item now in our view.
3. We delete the file.
4. ???

imsodin: That's what the change in global list is about: https://github.com/syncthing/syncthing/pull/5230/files#diff-69b15624219bb5333982e6cf1df4ef78L50  
Meaning in your point 2, "This is the global item now in our view." is no longer true, as it's invalid. If there is any non-invalid file, that's global before ours. This switch branch can only happen if we are the only ones having the file (otherwise `gf.IsReceiveOnlyChanged` is false).

calmh: Right, yes, OK. And the GreaterOrEqual in the need loop should mean that it doesn't become needed anyway.

I merge conflicted you by doing a couple of renames, will see if I can merge that for you.

PR #5237, opened 2018-10-02
gui: migrate to fork-awesome (fixes #5236)

### LINKED ISSUE #5236: Consider moving to Fork-Awesome or another free font
Hello!

It has come to my attention that Syncthing has moved from Font-Awesome v4 to Font-Awesome v5.

While I am sure that the intention was simply to update to the latest and greatest, it comes at the cost of making Syncthing impossible to build from source.

As of v5, FontAwesome has switched to a proprietary build system and it is no longer possible to build the icons. Of couse, this was done silently and I am sure you were not aware of this when migrating. 

We have been in contact with Font-Awesome developpers (see #[902981](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=902981#76)) and they have responded clearly:
 > the first question to get out of the way is: “Will Font Awesome open source their build tool?”
>
> And the answer is no for the time being.

This makes it impossible, for example, for Font-Awesome v5 to be included in Debian.

The community has already reacted to this, and the Fork-Awesome project was created, which is a fork of Font-Awesome v4, back when the build system was still available:
 - https://forkawesome.github.io/Fork-Awesome/
 - https://github.com/ForkAwesome/Fork-Awesome

I have gotten in touch with the guys at Fork-Awesome and they have been open consider a v5 compatibility layer, which they have already started working on, making it much easier to migrate to Fork-Awesome for Font-Awesome v5 users. See the following discussion: https://github.com/ForkAwesome/Fork-Awesome/issues/112

I have tested this compatibility layer with Syncthing and have gotten positive results. However, if Syncthing starts using icons that are not included in Fork-Awesome, then there is nothing that a compatibility layer will be able to do.

Would Syncthing consider migrating to Fork-Awesome?

aviau:
### Purpose

Move to a free icon set, [fork-awesome](https://forkawesome.github.io/Fork-Awesome/), as awesome v5 uses a proprietary build system. (See #5236)

### Testing

I ran Syncthing and saw that the icons are were present. I felt happy and accomplished.

### Screenshots

screenshot from 2018-10-02 17-48-50

If I ever forgot an icon, it is just a matter of adapting it to the fork-awesome name found [here](https://forkawesome.github.io/Fork-Awesome/icons/).

I only had to rename the refresh icons since fork-awesome supports most of the v5 names when you include the compat css layer.


AudriusButkevicius:
Let's get another +1 from @calmh, @imsodin or @uok

uok:
Sorry, strong no from here. 
Font Awesome has more than 1300 free icons and they work excellent with CSS only (no javascript needed). I don't understand what this PR tries to achieve or improve. Also wrong syntax was used (fork uses different prefix). In my opinion not worth the time.

AudriusButkevicius:
The original ticket explains what the issue is. The license of v5 is propriotary and incompatible.

uok:
Once again, no! The creator of this ticket did not even take the time to go through all html files (only index.html) :disappointed: and used compat.css because too lazy to replace code with correct fork syntax (no  fas-, far- and fab-prefix there). If this is merged it will break gui.

aviau:
> and used compat.css because too lazy to replace code with correct fork syntax

This compat file was written explicitly for Syncthing. What is the problem with that?

> If this is merged it will break gui.

What? Did you see the screenshot?

imsodin:
Whether it is better to use the compat layer or do a "full migration" to fork-awesome is probably discusseable, but if this PR works (I haven't seen any concrete point about breakage and I didn't check) I'd go with it. On the merits that it solves an actual problem and "someone" (@aviau isn't unknown to syncthing and calling him lazy is almost ironic) has done it.

calmh:
Yeah I don't follow the angst here either. Certainly it's possible some detail somewhere was overlooked, but it doesn't look like a total breakage at least.

uok:
It would have been great to know about this stuff 6 months ago when I migrated to v5, could have saved me quite some time going through all icons and new syntax. I don't do dev work on a daily basis so I think I will not be contributing any longer.

canton7:
It looks like this only started kicking off on Debian a few months ago, and a conclusion was only reached a few days ago. There was no way for you to know 😃

I think this PR does the right thing given the situation, though, for the sake of Debian users.

PR #5239, opened 2018-10-03
lib/model, cmd/syncthing: Wait for folder restarts to complete (fixes #5233)

### LINKED ISSUE #5233: Folder restart can leave several folder instances running
Consider this short, scary snippet from a v0.14.50 instance, after sharing/unsharing a folder a couple of times;

```
2018-10-02 14:55:57 progress emitter: registering fldr 192.168.11.215\2018-05-26\f.csv
2018-10-02 14:55:57 sendreceive/fldr@0x13fc7a20 parent not missing 192.168.11.106\2018-05-26\14.13.31.zip
2018-10-02 14:55:57 sendreceive/fldr@0x13fc7a20 taking shortcut on 192.168.11.106\2018-05-26\14.13.31.zip
2018-10-02 14:55:57 sendreceive/fldr@0x136749a0 filesystem notification rescan
2018-10-02 14:55:57 sharedPullerState fldr 192.168.11.215\2018-05-26\f.csv copyNeeded -> 0
2018-10-02 14:55:57 sendreceive/fldr@0x1425c160 parent not missing 192.168.11.219\2018-05-25\14.38.48.csv
2018-10-02 14:55:57 sendreceive/fldr@0x13675340 filesystem notification rescan
2018-10-02 14:55:57 sendreceive/fldr@0x1425c160 closing 192.168.11.107\2018-05-26\14.38.47.zip
```

The three things `sendreceive/fldr@0x13fc7a2`, `sendreceive/fldr@0x136749a0` and `sendreceive/fldr@0x1425c160` should definitely not be seen at the same time here.

calmh:
This is the somewhat ugly - but on the other hand clear - fix for what
is really a somewhat thorny issue. To avoid zombie folder runners a new
mutex is introduced that protects the RestartFolder operation. I hate
adding more mutexes but the alternatives I can think of are worse.

The other part of it is that the POST /rest/system/config operation now
waits for the config commit to complete. The point of this is that until
the commit has completed we should not accept another config commit. If
we did, we could end up with two separate RestartFolders queued in the
background. While they are both correct, and will run without
interfering with each other, we can't guarantee the order in which they
will run. Thus it could happen that the newer config got committed
first, and the older config commited after that, leaving us with the
wrong config running.

### Testing

The RestartFolder issue gets a new unit test. Before adding the mutex this would fail with something like

```
--- FAIL: TestFolderRestartZombies (1.01s)
    model_test.go:3890: Number of running routines increased from 13 to 1204
```

indicating a bunch of zombies were around.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -94,6 +94,7 @@ type Model struct {
 	clientName    string
 	clientVersion string
 
+	fmut               sync.RWMutex                                           // protects the below
 	folderCfgs         map[string]config.FolderConfiguration                  // folder -> cfg
 	folderFiles        map[string]*db.FileSet                                 // folder -> files
 	deviceStatRefs     map[protocol.DeviceID]*stats.DeviceStatisticsReference // deviceID -> statsRef
@@ -101,14 +102,15 @@ type Model struct {
 	folderRunners      map[string]service                                     // folder -> puller or scanner
 	folderRunnerTokens map[string][]suture.ServiceToken                       // folder -> tokens for puller or scanner
 	folderStatRefs     map[string]*stats.FolderStatisticsReference            // folder -> statsRef
-	fmut               sync.RWMutex                                           // protects the above
 
+	pmut                sync.RWMutex // protects the below
 	conn                map[protocol.DeviceID]connections.Connection
 	closed              map[protocol.DeviceID]chan struct{}
 	helloMessages       map[protocol.DeviceID]protocol.HelloResult
 	deviceDownloads     map[protocol.DeviceID]*deviceDownloadState
 	remotePausedFolders map[protocol.DeviceID][]string // deviceID -> folders
-	pmut                sync.RWMutex                   // protects the above
+
+	restartMut sync.Mutex // protects folder restarts
 }
 
 type folderFactory func(*Model, config.FolderConfiguration, versioner.Versioner, fs.Filesystem) service
@@ -162,6 +164,7 @@ func NewModel(cfg *config.Wrapper, id protocol.DeviceID, clientName, clientVersi
 		remotePausedFolders: make(map[protocol.DeviceID][]string),
 		fmut:                sync.NewRWMutex(),
 		pmut:                sync.NewRWMutex(),
+		restartMut:          sync.NewRWMutex(),
 	}
 	if cfg.Options().ProgressUpdateIntervalS > -1 {
 		go m.progressEmitter.Serve()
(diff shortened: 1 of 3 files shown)


lib/model/model.go
```diff
@@ -94,21 +94,23 @@ type Model struct {
 	clientName    string
 	clientVersion string
 
+	fmut               sync.RWMutex                                           // protects the below
```
calmh: (Placement changed because Go best practices is that the mutex lives on top of the things it protects...)

calmh:
Test is flaky because other tests don't take care to stop their own goroutines before returning, needs some haxing.

lib/model/model.go
```diff
@@ -94,21 +94,24 @@ type Model struct {
...
 
+	pmut                sync.RWMutex // protects the below
 	conn                map[protocol.DeviceID]connections.Connection
 	closed              map[protocol.DeviceID]chan struct{}
 	helloMessages       map[protocol.DeviceID]protocol.HelloResult
 	deviceDownloads     map[protocol.DeviceID]*deviceDownloadState
 	remotePausedFolders map[protocol.DeviceID][]string // deviceID -> folders
-	pmut                sync.RWMutex                   // protects the above
+
+	restartMut     sync.Mutex // protects folder restarts
```
AudriusButkevicius: Should we do `map[string]sync.Mutex` here? Imagine removal of a device that has all folders sshared, all of the folders would potentially need to be restarted, which means scans/transfers would have to be cancelled, which could take a fair amount of time...

calmh: Yeah I was considering it, but didn't want to complicate matters... But I can do that and see how it looks.

calmh:
I think that was quite clean. It's intentional that we never remove anything from the restart mutex map, as it should explicitly cover multiple restarts of the same folder ID

imsodin:
> It's intentional that we never remove anything from the restart mutex map, as it should explicitly cover multiple restarts of the same folder ID

What about removing folders?

calmh:
What about it? :) We'll have an extra mutex costing a few bytes lying around, that's all.

imsodin:
Not much about it. I was just thinking there is for sure some user that has a server that runs endlessly and adds and removes folders (lots of them of course) with a script hourly and then after half a year or so will come asking about memory leakage, together with a few other completely unrelated symptoms :P

calmh:
When they do, they'll be reporting the problem on an old version so we can't troubleshoot. Need to try it on the current version! Half a year later, `goto 10`. ;)

PR #5240, opened 2018-10-04
gui: Sort folders and devices on status (#fixes 5044)

### LINKED ISSUE #5044: Sort folders/devices on status
Currently folders and devices are always sorted on label/ID. When there are a lot of them, being able to sort based on sync status (and then label/ID as secondary) would be useful.

bahadir60:
### Purpose
Sort `folders` and `devices` on `status` instead of names. (fixes #5044).

Folders sorted by 7 and devices sorted by 6 different status.

#### Select status sorting:  **Action > Advanced > Options**
screen

#### Before `status sort` (namely it is `default`):
screen shot 2018-10-06 at 22 39 32

#### After `status sort`:
screen shot 2018-10-06 at 22 39 53

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1514,6 +1514,54 @@ angular.module('syncthing.core')
             });
         };
 
+        $scope.statusOtherDevices = function () {
+            var unknownArray = [], unusedArray = [], pausedArray = [], insyncArray = [], disconnectedArray = [], defaultArray =[];
+
+            // loop through all devices
+            var devices = $scope.otherDevices();
+            var deviceLength = devices.length;
+            for (var i = 0; i < deviceLength; i++) {
+                var status = $scope.deviceStatus({
+                    deviceID:$scope.devices[i].deviceID
+                });
+
+                switch (status){
+                    case 'unused':
+                        unusedArray.push(devices[i]);
+                        break;
+
+                    case 'unknown':
+                        unknownArray.push(devices[i]);
+                        break;
+
+                    case 'paused':
+                        pausedArray.push(devices[i]);
+                        break;
+
+                    case 'insync' || 'syncing':
+                        insyncArray.push(devices[i]);
+                        break;
+
+                    case 'disconnected':
+                        disconnectedArray.push(devices[i]);
+                        break;
+
+                    default:
+                        defaultArray.push(devices[i]);
+                }
+            };
+
+            return unknownArray.concat(insyncArray, disconnectedArray, pausedArray, unusedArray, defaultArray);
+        };
+
+        $scope.guiShowOtherDevices = function () {
+            var statusSort = $scope.config.options.sortFolderDeviceByStatus;
+            if (statusSort){
+                return $scope.statusOtherDevices();
+            }
+            return $scope.otherDevices();
+        };
+
         $scope.thisDevice = function () {
             return $scope.thisDeviceIn($scope.devices);
         }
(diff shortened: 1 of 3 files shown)


calmh:
Code looks OK to me. However, I don't think we want to switch to this as our one and only method - I strongly dislike when things move around, for example. Some people want this, but it needs to be an option, somehow.

Other GUI options we have (like, language and rate units) are local to the browser and saved in browser local storage. We should probably do the same here, whatever the UI ends up looking like.

uok:
@bahadir60, thanks for working on this

thoughts and ideas:
+ option to change (alphabetically or status) or disable sorting to prevent wierd moving because average user only has 2 folders (see [statistics](https://data.syncthing.net/))
+ if device connects and is in sync, folder is "syncthing" for 1-2 seconds, so it would jump. maybe update sorting only with intervals?
+ status "unknown" should be first, I think
+ there is also status "scanning", I don't know if relevant

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1559,6 +1603,54 @@ angular.module('syncthing.core')
...
+                }
+            };
+
+            idleArray = folderList(idleArray)
+            pausedArray = folderList(pausedArray)
+            syncingArray = folderList(syncingArray)
+            unknownArray = folderList(unknownArray)
+            stoppedArray = folderList(stoppedArray)
+            unsharedArray = folderList(unsharedArray)
+            defaultArray = folderList(defaultArray)
```
wweich: Wouldn't it be better to concat the Arrays first and invoke `folderList()`only once?

bahadir60: if we concat arrays and than call `folderList` we turn back default sorting base on `id` so `name`.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1508,6 +1508,50 @@ angular.module('syncthing.core')
             });
         };
 
+        $scope.guiShowOtherDevices = function () {
+            var insyncArray, unusedArray, pausedArray, syncingArray, unknownArray, disconnectedArray, defaultArray = [];
```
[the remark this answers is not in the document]
bahadir60: Yeap it is possible but this time I will create varibles inside the case. Empty array is not a big problem IMHO.

bahadir60:
@calmh may be we add a checkbox under the `Actions > Advance > GUI` or `Actions > Advance > Options` to change `default` to `status` sorting. And as you said we can store the checkbox value on localstorage. What u think about it?

bahadir60:
@uok thanks you comment. I will add button or checkbox for selection. The default is sort by `name` as usual. Scanning and syncing status are in same array (code line 1623), IMHO, it will not jump. But after testing we can change order of course. Thanx.

bahadir60:
I put a checkbox under the `Advanced Configuration > Options` and based on comments changed the order of status arrays.

uok:
If it's not too much trouble could you change option label from `Folder Device Sort By Status` to `Sort Folder Device By Status` or `Sort Folders Devices By Status`? It's a little clearer this way, thanks!

bahadir60:
@uok thanx, naming changed based on ur comment.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1508,6 +1510,60 @@ angular.module('syncthing.core')
...
+                    default:
+                        defaultArray.push(devices[i]);
+                }
+            };
+
+            return unknownArray.concat(insyncArray, disconnectedArray, pausedArray, unusedArray, defaultArray);
+        };
+
+        $scope.guiShowOtherDevices = function () {
+            if($scope.statusSort == 'undefined'){
```
calmh: I don't understand what's going on in here, maybe I missed something in a previous discussion. If `statusSort` is `undefined` (which is better checked as just `$scope.statusSort === undefined` without the string conversion, btw), we return a future-scheduled function that checks again the value of `statusSort` (which we already determined is undefined), and ... what? :)

bahadir60: I honestly did not remember why I coded like that, it makes double check may be. Of course, I changed it. Thanks.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1514,6 +1514,54 @@ angular.module('syncthing.core')
...
         };
 
+        $scope.statusOtherDevices = function () {
+            var unknownArray = [], unusedArray = [], pausedArray = [], insyncArray = [], disconnectedArray = [], defaultArray =[];
+
+            // loop through all devices
+            var devices = $scope.otherDevices();
+            var deviceLength = devices.length;
+            for (var i = 0; i < deviceLength; i++) {
+                var status = $scope.deviceStatus({
```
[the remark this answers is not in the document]
bahadir60: Thanks, It is really cool. But I think we have a problem about not defined array (status). if don't have `byStatus['insync']`array return statement gives error. This time I loop in `byStatus` object.
```
var devices = $scope.otherDevices();
var byStatus = {};
for (var i = 0; i < devices.length; i++) {
  var status = $scope.deviceStatus(devices[i]);
  if (!byStatus[status]) {
    // needs init
    byStatus[status] = [];
  }
  byStatus[status].push(devices[i]);
}

var result = [];
Object.entries(byStatus).forEach(entry => {
// let key = entry[0];
  let value = entry[1];
result.concat(value);
});
return result;
```
However, this time the desired order can not be obtained which we made it in return statement before. I should rearrange it by key but need again loops, switch etc.

or
I can write a another function like;
```
var devices = $scope.otherDevices();
var byStatus = {};
for (var i = 0; i < devices.length; i++) {
  var status = $scope.deviceStatus(devices[i]);
  if (!byStatus[status]) {
    // needs init
    byStatus[status] = [];
  }
  byStatus[status].push(devices[i]);
}
// desired order keys
keys = ['insync', 'diconnected', 'unknown',...] 
var result = [];
for (var i=0; i < keys.lenght; i++){
   if (byStatus. hasOwnProperty(key[i]){
        result.concat(byStatus[key[i]])
    }
}
return result;

I am not sure which one is best.

st-review:
:robot: *beep boop*

PR #5260, opened 2018-10-11
lib/db: Properly remove FileInfos when dropping folder

calmh:
This doesn't cause any catastrophic issues other than FileInfos
lingering in the database forever... But the prefix here was wrong
(including a nil device ID) resulting in this cleanup being a no-op.


calmh:
Good catch

PR #5263, opened 2018-10-11
vendor: Update github.com/syncthing/notify

imsodin:
Lets see whether the upstream fix for the darwin cgo stuff works.


calmh:
What is this a fix for?

imsodin:
Since your fix works, it purely cosmetic to bring us in line with upstream: Before it was either `refZero` or `nil`, now it is `C.kCFAllocatorDefault`, which apparently does the right thing in both cases (i.e. notify compiles on 1.11 and 1.10 without build tags). See https://github.com/rjeczalik/notify/pull/162

PR #5268, opened 2018-10-12
lib/config: Disable folder free disk check when configured (fixes #5267)

### LINKED ISSUE #5267: Disabling the minimum disk free space check on folders doesn't work
Even though the `minDiskFree` config option is set to zero, which should disable it, the check still happens and aborts pulling if there isn't any free disk space (which can always be the case for SSHFS): https://forum.syncthing.net/t/insufficient-space-error-from-v0-14-51-with-sshfs/12276

imsodin:
#5267
https://forum.syncthing.net/t/insufficient-space-error-from-v0-14-51-with-sshfs/12276

This only changes things in the corner-case where free space is smaller than the space required for the pull operation (`usage.Free - req <= 0`).


DocMarty84:
It works great thanks!

PR #5271, opened 2018-10-14
all: Event.Data: interface{} -> map[string]interface{} (fixes #5270)

### LINKED ISSUE #5270: Syncthing panics with 'interface conversion: interface ...'
I've noticed that Syncthing 0.14.51 crashes on my Arch Linux system with the following error:   
```
panic: interface conversion: interface {} is map[string]string, not map[string]interface {}
```
Full error log: http://paste.opensuse.org/96753389

Here's how I could reproduce the issue:
1. Copy https://github.com/Martchus/syncthingtray/blob/master/connector/testfiles/testconfig/config.xml to an empty directory
2. Start `syncthing -gui-address=http://localhost:4001 -gui-apikey=syncthingtestinstance -home=/dir/with/config -no-browser -verbose`

Further notes:
* The problem only occurs if device `MMGUI6U-WUEZQCP-XZZ6VYB-LCT4TVC-ER2HAVX-QYT6X7D-S6ZSG2B-323KLQ7` is actually reachable in my local network.
* This device in my local network has actually a different device ID than stated in the configuration file. That 'misconfiguration' comes from the fact that https://github.com/Martchus/syncthingtray/blob/master/connector/testfiles/testconfig/config.xml is really just intended for testing Syncthing Tray's code and no actual connection is supposed to happen.
* I have another Syncthing instance running at the same time which is configured correctly and it can connect to the mentioned device in my local network.

Comments from @imsodin via IRC:
> [23:50] \<imsodin\> Martchus: The log is pretty clear, looks like a straight forward bug. what's not straight forward is, that it is present since an eternity and you seem to be the first to observe it!?
> [23:50] \<imsodin\> could you file it on github please
> [00:06] \<Martchus\> to me 'panic: interface conversion: interface {} is map[string]string, not map[string]interface {}' mean litterally nothing - likely some (de)serialization code - but it doesn't point to a file or line of code
> [00:13] \<imsodin\> the next block contains the info about which line in the code this originates.
> [00:13] \<imsodin\> then I grep for "DeviceRejected" and see that the type in which it is logged is different from the type in the assertion -> boom
> [00:14] \<imsodin\> i can reliably reproduce by starting a fresh instance with -verbose and by adding this device on a remote device. as soon as the "invitation" reaches here, it panics
> [00:15] \<imsodin\> i suspect this did not surface due to suture handling the panics before

imsodin:
### Purpose

Instead of just fixing the one issue (see https://github.com/syncthing/syncthing/issues/5270#issuecomment-429669744), I removed one level of interfaces/ambiguity. To be really consequent, all events should be structs, thus preventing any potential for false type assertions, but that's also very cumbersome. I like `map[string]interface{}` better than `interface` for `event.Data`, but if you don't (or just can't stand this ugly diff), just tell me and the fix will become a one line diff.

### Testing

Manual testing: Doesn't panic anymore on rejected folder with `-verbose` and verbose logging still works as far as I observed.

### DIFF
--- a/lib/model/progressemitter.go
+++ b/lib/model/progressemitter.go
@@ -91,12 +91,12 @@ func (t *ProgressEmitter) Serve() {
 
 func (t *ProgressEmitter) sendDownloadProgressEvent() {
 	// registry lock already held
-	output := make(map[string]map[string]*pullerProgress)
+	output := make(map[string]interface{})
 	for _, puller := range t.registry {
 		if output[puller.folder] == nil {
 			output[puller.folder] = make(map[string]*pullerProgress)
 		}
-		output[puller.folder][puller.file.Name] = puller.Progress()
+		output[puller.folder].(map[string]*pullerProgress)[puller.file.Name] = puller.Progress()
 	}
 	events.Default.Log(events.DownloadProgress, output)
 	l.Debugf("progress emitter: emitting %#v", output)
--- a/cmd/syncthing/auditservice_test.go
+++ b/cmd/syncthing/auditservice_test.go
@@ -20,13 +20,13 @@ func TestAuditService(t *testing.T) {
 	service := newAuditService(buf)
 
 	// Event sent before start, will not be logged
-	events.Default.Log(events.ConfigSaved, "the first event")
+	events.Default.Log(events.ConfigSaved, map[string]interface{}{"": "the first event"})
 
 	go service.Serve()
 	service.WaitForStart()
 
 	// Event that should end up in the audit log
-	events.Default.Log(events.ConfigSaved, "the second event")
+	events.Default.Log(events.ConfigSaved, map[string]interface{}{"": "the second event"})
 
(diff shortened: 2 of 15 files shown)


lib/model/progressemitter.go
```diff
@@ -91,12 +91,13 @@ func (t *ProgressEmitter) Serve() {
 
 func (t *ProgressEmitter) sendDownloadProgressEvent() {
 	// registry lock already held
-	output := make(map[string]map[string]*pullerProgress)
+	output := make(map[string]interface{})
+	// output := make(map[string])
```
Martchus: leftover?

Martchus:
Thanks for fixing the issue!

> just tell me and the fix will become a one line diff

As non-Go programmer I don't have an opinion here. But independently from that I would say it makes sense to split this into 2 commits (Fix issue ..., Polish/refactor ...).

PR #5277, opened 2018-10-18
build: Add desktop files and icons to .deb (fixes #3439)

### LINKED ISSUE #3439: Ship .desktop files

imsodin:
### Purpose

Integrate Syncthing better with DEs (make it appear in menus) and allow for easier autostart on login with a DE, i.e. no more explaining of `systemd --user`.

### Testing

None so far, waiting for the build artifacts.


canton7:
As I naive user, I wouldn't know the difference between "Syncthing Daemon" and "Syncthing Web UI"

imsodin:
@canton7 I know what you mean, but I don't see any simple way of having a single entry that does "start service if not running already, otherwise just open the web UI". Except of course ugly hacks like `Exec=/bin/sh -c "syncthing || syncthing -browser-only"` (which assumes any failure is due to syncthing already running).

imsodin:
> At the very least, I'd rename "Syncthing Daemon" to "Start Syncthing" or something.

Done.

> What happens if someone clicks on "Syncthing Daemon" twice?

The second time it fails, but the user wont even notice. Which I think is ok, as the result is still that Syncthing is running.

> If we merged the two, users wouldn't be able to add the daemon to their gnome autostart thigny without the browser opening up, right? That would suck :(

That's actually a good point: Before the last commit a browser window would open for the syncthing-start.desktop (previously daemon), which I now disabled.

> Am I right assuming that DEs like gnome/unity/kde allow you to set a .destkop file to autostart?

Yes, that's right. I actually hoped the readme makes this clear (?) ;)

aviau:
> Yes, that's right. I actually hoped the readme makes this clear (?) ;)

It does!

PR #5281, opened 2018-10-21
lib: chmod -x on progressemitter.go and errors.go

aviau:
These files were probably not meant to be executable?

Let me know if I am mistaken.

Cheers,


AudriusButkevicius:
The test failure unrelated, but might point to a bug

PR #5283, opened 2018-10-21
lib: fmt config/config_test.go

aviau:
This file wasnt gofmt-ed, it was annoying to exclude it of my diffs :smile:


AudriusButkevicius:
It is formatted, with a newer version of gofmt that changed how the format looks.

Issue #1347, opened 2015-02-11
Directory auto-complete is case-sensitive


uok:
auto-complete for directories is a really helpful feature, please make it case-insensitive

```
c:\A
c:\B
c:\C
```

typing "c:\b" results in

expected
`c:\B`

calmh:
(file system specific though)

dzarda:
Maybe I'll have a look at this during the weekend, just to get my head around it.

dzarda:
The matching currently works using [Glob](http://golang.org/pkg/path/filepath/#Glob). This is the culprit, since this is always case-sensitive. It would be rewritten.

Also, is a condition `runtime.GOOS == "windows"` a good enough approximation of "file system specificity" ? I mean, how can I _easily_ and _quickly_ detect if a path is on a case-insensitive partition?

calmh:
You probably can't, but the system type is a good approximation, and probably an even better approximation of what behavior the user expects (i.e. a Linux user is probably not shocked if she gets case sensitive completion on a FAT filesystem; her shell will do the same).

sciurius:
The best test is to create a dummy file FooBar.txt and check for the existence of foobar.txt .

calmh:
A folder may be on a read only filesystem (if the `.stfolder` file has been created manually), or may span filesystems for that matter. But yes, we could do a probe. I've considered it as well for things like whether we should use IgnorePermissions or not etc.

edsonmedina:
Shouldn't a fix for this be done in `Glob`'s repository instead?

bennofs:
shoudn't auto-complete be always case-insensitive? Even if I am on Linux, I usually don't have files that differ only in case so it still can be convenient to only type lowercase to match an uppercase file. Since this is only auto-completion, I think that should be fine as long as we sort exact matches before "fuzzy case" ones?

AudriusButkevicius:
Agree, but this needs a custom implementation of globbing.

bennofs:
OK, i'll try to impment this then

PR #5289, opened 2018-10-25
Make directory auto-complete case insensitive

bennofs:
### Purpose

This implements #1347 making directory auto-completion case insensitive.

### Testing

Added tests to `gui_test.go`.

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -45,7 +45,6 @@ import (
 	"github.com/syncthing/syncthing/lib/versioner"
 	"github.com/vitrun/qart/qr"
 	"golang.org/x/crypto/bcrypt"
-	"unicode"
 )
 
 var (
@@ -1543,45 +1542,29 @@ func (s *apiService) getSystemBrowse(w http.ResponseWriter, r *http.Request) {
 	sendJSON(w, browseFiles(current, fsType))
 }
 
+type matchKind int
+
+const (
+	MatchExact matchKind = iota
+	MatchCaseIns
+	NoMatch
+)
+
 type match struct {
-	path            string
-	caseDifferences int
+	path string
+	kind matchKind
 }
 
-func prefixCaseDifferences(s, prefix string) int {
+func checkPrefixMatch(s, prefix string) matchKind {
 	if strings.HasPrefix(s, prefix) {
-		return 0
+		return MatchExact
 	}
 
-	sRunes := []rune(s)
-	prefixRunes := []rune(prefix)
-
-	diff := 0
-	for i := 0; i < len(sRunes) && i < len(prefixRunes); i += 1 {
-		if sRunes[i] == prefixRunes[i] {
-			continue
-		}
-
-		lower := sRunes[i]
-		upper := prefixRunes[i]
-		if lower > upper {
-			lower = upper
-		}
-
-		r := unicode.SimpleFold(lower)
-		for lower < r && r < upper {
-			r = unicode.SimpleFold(r)
-		}
-
-		// not a prefix at all
-		if r != upper {
-			return -1
-		}
-
-		diff += 1
+	if strings.HasPrefix(strings.ToLower(s), strings.ToLower(prefix)) {
+		return MatchCaseIns
 	}
 
-	return diff
+	return NoMatch
 }
 
 func browseFiles(current string, fsType fs.FilesystemType) []string {
@@ -1615,27 +1598,26 @@ func browseFiles(current string, fsType fs.FilesystemType) []string {
 
 	for _, subdirectory := range subdirectories {
 		info, err := fs.Stat(subdirectory)
-		if !(err == nil && info.IsDir()) {
+		if err != nil || !info.IsDir() {
 			continue
 		}
 
-		m := match{
-			path:            filepath.Join(searchDir, subdirectory) + pathSeparator,
-			caseDifferences: prefixCaseDifferences(subdirectory, searchFile),
-		}
-
-		if m.caseDifferences >= 0 {
-			matches = append(matches, m)
+		matchKind := checkPrefixMatch(subdirectory, searchFile)
+		if matchKind != NoMatch {
+			matches = append(matches, match{
+				path: filepath.Join(searchDir, subdirectory) + pathSeparator,
+				kind: matchKind,
+			})
 		}
 	}
 
-	// sort by number of differences with pattern
+	// sort exact matches before case-insensitive ones
 	// ties are resolved by alphabetical order
 	sort.Slice(matches, func(i, j int) bool {
-		if matches[i].caseDifferences == matches[j].caseDifferences {
+		if matches[i].kind == matches[j].kind {
 			return strings.ToLower(matches[i].path) < strings.ToLower(matches[j].path)
 		}
-		return matches[i].caseDifferences < matches[j].caseDifferences
+		return matches[i].kind == MatchExact
 	})
 
 	ret := make([]string, 0, len(matches))
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -1570,7 +1570,7 @@ func browseFiles(current string, fsType fs.FilesystemType) []string {
 	subdirectories, _ := fs.DirNames("")
 
 	type match struct {
-		path string
+		path             string
 		case_differences int
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/syncthing/gui.go
```diff
@@ -1542,6 +1543,18 @@ func (s *apiService) getSystemBrowse(w http.ResponseWriter, r *http.Request) {
 	sendJSON(w, browseFiles(current, fsType))
 }
 
+func makeSearchGlobPattern(searchFile string) string {
+	p := ""
```
AudriusButkevicius: Use string builder

cmd/syncthing/gui.go
```diff
@@ -1567,15 +1580,31 @@ func browseFiles(current string, fsType fs.FilesystemType) []string {
...
 		info, err := fs.Stat(subdirectory)
 		if err == nil && info.IsDir() {
+			idx := len(ret)
 			ret = append(ret, filepath.Join(searchDir, subdirectory)+pathSeparator)
+			matchcount = append(matchcount, 0);
+			for i := 0; i < len(subdirectory) && i < len(searchFile); i += 1 {
+				if subdirectory[i] == searchFile[i] {
+					matchcount[idx] += 1
+				}
+			}
```
AudriusButkevicius: Whats the intention here?
Struggling to understand what this does.

cmd/syncthing/gui.go
```diff
@@ -1567,15 +1580,31 @@ func browseFiles(current string, fsType fs.FilesystemType) []string {
...
 		}
 	}
+	// sort by number of matches with pattern, largest number of matches first
+	// ties are resolved by alphabetical order
+	sort.Slice(ret, func(i, j int) bool {
+		if matchcount[i] == matchcount[j] {
+			return strings.ToLower(ret[i]) < strings.ToLower(ret[j])
+		}
+		return matchcount[i] > matchcount[j]
+	})
```
AudriusButkevicius: It's strange that ret is sorted based on some other array. Is it not easier to use a map or a slice of structs where it has the match and the count?

cmd/syncthing/gui_test.go
```diff
@@ -1002,13 +1006,15 @@ func TestBrowse(t *testing.T) {
...
+		{tmpDir + pathSep, []string{dirPath, mixedCaseDirPath}},
 		// Globbing is automatic based on prefix.
 		{tmpDir + pathSep + "d", []string{dirPath}},
 		{tmpDir + pathSep + "di", []string{dirPath}},
 		{tmpDir + pathSep + "dir", []string{dirPath}},
 		{tmpDir + pathSep + "f", nil},
 		{tmpDir + pathSep + "q", nil},
+		// Globbing is case-insensitve
+		{tmpDir + pathSep + "mixed", []string{mixedCaseDirPath}},
 	}
```
AudriusButkevicius: A test for the magical counting would be nice

AudriusButkevicius: We have a fake filesystem implementation which you might be able to use here.

cmd/syncthing/gui.go
AudriusButkevicius: Should you not break on the first mismatch?

cmd/syncthing/gui.go
bennofs: hmm, with a map or a slice of structs I then need allocate another array at the end for the return value. Is this preferred?

bennofs:
addressed all the feedback now

cmd/syncthing/gui.go
```diff
@@ -1567,15 +1567,51 @@ func browseFiles(current string, fsType fs.FilesystemType) []string {
 
 	fs := fs.NewFilesystem(fsType, searchDir)
 
-	subdirectories, _ := fs.Glob(searchFile + "*")
+	subdirectories, _ := fs.DirNames("")
```
AudriusButkevicius: Use "."

cmd/syncthing/gui.go
```diff
@@ -1567,15 +1567,51 @@ func browseFiles(current string, fsType fs.FilesystemType) []string {
...
 
-	ret := make([]string, 0, len(subdirectories))
+	type match struct {
+		path             string
+		case_differences int
+	}
```
AudriusButkevicius: Don't make inline structs in functions. Put it at file scope. Also, no snake case in go.

cmd/syncthing/gui.go
```diff
@@ -1567,15 +1567,51 @@ func browseFiles(current string, fsType fs.FilesystemType) []string {
...
+
+			continue next_subdir
+		}
+
 		info, err := fs.Stat(subdirectory)
 		if err == nil && info.IsDir() {
-			ret = append(ret, filepath.Join(searchDir, subdirectory)+pathSeparator)
+			m.path = filepath.Join(searchDir, subdirectory) + pathSeparator
+			matches = append(matches, m)
 		}
```
AudriusButkevicius: Why do we even bother doing the path comparison dance if its not a dir?

cmd/syncthing/gui.go
```diff
@@ -1567,15 +1567,51 @@ func browseFiles(current string, fsType fs.FilesystemType) []string {
...
+
+next_subdir:
 	for _, subdirectory := range subdirectories {
+		var m match
+		for i := 0; i < len(subdirectory) && i < len(searchFile); i += 1 {
+			if subdirectory[i] == searchFile[i] {
+				continue
+			}
+
+			if strings.EqualFold(string(subdirectory[i]), string(searchFile[i])) {
```
AudriusButkevicius: This can be funky with utf8 where a character is more than one byte, perhaps you should extract runes using utf8 package, and do unicode.ToLower(rune) etc.

cmd/syncthing/gui.go
```diff
@@ -1567,15 +1567,51 @@ func browseFiles(current string, fsType fs.FilesystemType) []string {
...
 		}
 	}
+
+	// sort by number of differences with pattern
+	// ties are resolved by alphabetical order
+	sort.Slice(matches, func(i, j int) bool {
+		if matches[i].case_differences == matches[j].case_differences {
+			return strings.ToLower(matches[i].path) < strings.ToLower(matches[j].path)
+		}
+		return matches[i].case_differences < matches[j].case_differences
```
AudriusButkevicius: This can just be:

return caseDiffB > caseDiffA || lower(a) < lower(b)

cmd/syncthing/gui.go
```diff
@@ -1567,15 +1567,51 @@ func browseFiles(current string, fsType fs.FilesystemType) []string {
...
 
-	ret := make([]string, 0, len(subdirectories))
+	type match struct {
+		path             string
+		case_differences int
+	}
+
+	matches := make([]match, 0, len(subdirectories))
+
+next_subdir:
```
AudriusButkevicius: camelCase I think

cmd/syncthing/gui.go
bennofs: Isn't a stat call costly? I die it this way to avoid unnecessary statt calls

calmh: Not really, in the context of something triggered essentially manually via a network API call.

cmd/syncthing/gui.go
bennofs: I believe that would be incorrect. Example: `caseDiffB = 0, caseDiffA = 1` should always return `false` even if `lower(a) < lower(b)`

bennofs:
refactored the changes according to suggestions

cmd/syncthing/gui.go
```diff
@@ -1567,14 +1609,38 @@ func browseFiles(current string, fsType fs.FilesystemType) []string {
...
+			continue
+		}
+
+		m := match{
+			path:            filepath.Join(searchDir, subdirectory) + pathSeparator,
+			caseDifferences: prefixCaseDifferences(subdirectory, searchFile),
 		}
+
+		if m.caseDifferences >= 0 {
+			matches = append(matches, m)
```
AudriusButkevicius: You could store the prefifDiff on the stack, and if that is not negative, only then create match in the append statement.

AudriusButkevicius:
There is also some magic to prioritise items that match the casing more than others.

djhenk:
This might be a dumb question, but why needs this feature a custom search algorithm? Couldn't we just be a bit more clever by expanding the search pattern to include both upper and lower case? Instead of searching for `"name*"`,  search for `"[nN][aA][mM][eE]*"`.

AudriusButkevicius:
That was the initial approach but its not obvious if that handles unicode

bennofs:
if you want I can change the code to not do the "how many characters are of correct case" check and instead only track whether the match required case-insensitive matching or not (and then sort case-sensitive matches before the insensitive ones). This could reduce the complexity a bit.

AudriusButkevicius:
I think its ok as it is, its a nice touch, yet I want @imsodin or @calmh to dig into it and confirm that they agree with the approach and wether my utf-8 nonsense is even needed.

bennofs:
implemented the suggested simplifications

AudriusButkevicius:
Agree with @imsodin's suggestions, otherwise looks ok (including dropping matchkind type)

bennofs:
thanks for all the suggestions, the code is much simpler now

PR #5290, opened 2018-10-25
cmd/syncthing, lib/relay: Fixes regarding stopping of services

imsodin:
While running  #5216 with race detector on my laptop ("in production") I got three races:

```
==================
WARNING: DATA RACE
Read at 0x00c002377ab0 by goroutine 338:
  github.com/syncthing/syncthing/lib/relay/client.(*staticClient).Stop()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/relay/client/static.go:172 +0x47
  github.com/syncthing/syncthing/lib/relay/client.(*dynamicClient).Stop()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/relay/client/dynamic.go:128 +0x109
  github.com/syncthing/syncthing/lib/connections.(*relayListener).Stop()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/connections/relay_listen.go:124 +0xac
  github.com/syncthing/syncthing/vendor/github.com/thejerf/suture.(*Supervisor).stopSupervisor.func1()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/vendor/github.com/thejerf/suture/supervisor.go:590 +0x3e
Previous write at 0x00c002377ab0 by goroutine 65:
  [failed to restore the stack]
Goroutine 338 (running) created at:
  github.com/syncthing/syncthing/vendor/github.com/thejerf/suture.(*Supervisor).stopSupervisor()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/vendor/github.com/thejerf/suture/supervisor.go:589 +0x2d0
  github.com/syncthing/syncthing/vendor/github.com/thejerf/suture.(*Supervisor).Serve()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/vendor/github.com/thejerf/suture/supervisor.go:434 +0x108b
  github.com/syncthing/syncthing/vendor/github.com/thejerf/suture.(*Supervisor).runService.func1()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/vendor/github.com/thejerf/suture/supervisor.go:540 +0x6c
Goroutine 65 (running) created at:
  github.com/syncthing/syncthing/lib/connections.(*relayListener).Serve()
      /media/ext4_data/Coding/go/src/github.com/syncthing/syncthing/lib/connections/relay_listen.go:57 +0x1f3
[...]

PR #5293, opened 2018-10-26
cmd/syncthing, lib/relay: Fixes regarding stopping of services

imsodin:
Github doesn't let me reopen #5290, so this fixes the same races, but (hopefully) doesn't introduce stupidities this time.

### DIFF
--- a/cmd/syncthing/usage_report.go
+++ b/cmd/syncthing/usage_report.go
@@ -17,6 +17,7 @@ import (
 	"runtime"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/syncthing/syncthing/lib/config"
@@ -326,6 +327,9 @@ type usageReportingService struct {
 	connectionsService *connections.Service
 	forceRun           chan struct{}
 	stop               chan struct{}
+	stopped            chan struct{}
+	stopOnce           sync.Once
+	stopMut            sync.RWMutex
 }
 
 func newUsageReportingService(cfg *config.Wrapper, model *model.Model, connectionsService *connections.Service) *usageReportingService {
@@ -335,7 +339,9 @@ func newUsageReportingService(cfg *config.Wrapper, model *model.Model, connectio
 		connectionsService: connectionsService,
 		forceRun:           make(chan struct{}),
 		stop:               make(chan struct{}),
+		stopped:            make(chan struct{}),
 	}
+	close(svc.stopped) // Not yet running, dont block on Stop()
 	cfg.Subscribe(svc)
 	return svc
 }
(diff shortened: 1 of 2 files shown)


cmd/syncthing/usage_report.go
```diff
@@ -335,7 +339,9 @@ func newUsageReportingService(cfg *config.Wrapper, model *model.Model, connectio
 		connectionsService: connectionsService,
 		forceRun:           make(chan struct{}),
 		stop:               make(chan struct{}),
+		stopped:            make(chan struct{}),
 	}
+	close(svc.stopped) // Not yet running, dont block on Stop()
```
AudriusButkevicius: Reads on nil channel return immediately

cmd/syncthing/usage_report.go
```diff
@@ -387,14 +398,19 @@ func (s *usageReportingService) VerifyConfiguration(from, to config.Configuratio
...
+		}
 	}
 	return true
 }
 
 func (s *usageReportingService) Stop() {
-	close(s.stop)
-	close(s.forceRun)
+	s.stopMut.RLock()
+	s.stopOnce.Do(func() { close(s.stop) })
```
AudriusButkevicius: Nil the channel after closing, and close if not nil (under lock).

cmd/syncthing/usage_report.go
imsodin: No, they block: https://play.golang.org/p/WWsV4OHivFj

calmh:
This feels overengineered, tbh. Why would stop be called more than once, requring the `Once`? We should have a correct pattern for this once and for all the we use everywhere...

cmd/syncthing/usage_report.go
```diff
@@ -387,14 +398,19 @@ func (s *usageReportingService) VerifyConfiguration(from, to config.Configuratio
 
 func (s *usageReportingService) CommitConfiguration(from, to config.Configuration) bool {
 	if from.Options.URAccepted != to.Options.URAccepted || from.Options.URUniqueID != to.Options.URUniqueID || from.Options.URURL != to.Options.URURL {
-		s.forceRun <- struct{}{}
+		select {
+		case s.forceRun <- struct{}{}:
+		case <-s.stop:
```
AudriusButkevicius: Stop is read under a lock elsewhere

imsodin:
Agreed. The `Once`s were more or less a reaction to me getting confused by the running pattern of the services (some through supervisor, some manually), so I just wanted to do it "safe". However apparently until know `Stop` was never called more than once, so there's indeed no reason to assume it might.  
Now this is just an ugly bunch of locks required to make it non-racy (I assume the race occurs when `Stop` is called immediately after `Serve`).

calmh:
We should look into improving that wholesale though, yes

PR #5307, opened 2018-11-03
gui: 'Select/Deselect All' tick box when adding new folder/device or editing folder/device(#fixes 4000)

bahadir60:
### Purpose

To allow **selecting or if all selected let deselecting all tick boxes** (toggle) when adding new folder (`Add Folder`), new device (`Add Remote Device`) or using `Edit` button on folder (under 'Folders') and device (under 'Remote Devices').

### Screenshots

- **`Select All`**

screen shot 2018-11-04 at 15 51 55

- **`Deselect All`**

screen shot 2018-11-04 at 15 52 01

### Documentation

No need.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1371,6 +1371,20 @@ angular.module('syncthing.core')
             $('#editDevice').modal();
         };
 
+        $scope.selectAllFolders = function() {
+            Object.entries($scope.folders).forEach(entry =>{
+                let id = entry[1].id;
+                $scope.currentDevice.selectedFolders[id] = true;
+            });
+        };
+
+        $scope.deSelectAllFolders = function() {
+            Object.entries($scope.folders).forEach(entry =>{
+                let id = entry[1].id;
+                $scope.currentDevice.selectedFolders[id] = false;
+            });
+        };
+
         $scope.addDevice = function (deviceID, name) {
             return $http.get(urlbase + '/system/discovery')
                 .success(function (registry) {
@@ -1692,6 +1706,20 @@ angular.module('syncthing.core')
                 });
 
             $scope.editFolderModal();
+        };
+
+        $scope.selectAllDevices = function() {
+            var devices = $scope.otherDevices();
+            for (var i = 0; i < devices.length; i++){
+                $scope.currentFolder.selectedDevices[devices[i].deviceID] = true;
+            }
+        };
+
+        $scope.deSelectAllDevices = function() {
+            var devices = $scope.otherDevices();
+            for (var i = 0; i < devices.length; i++){
+                $scope.currentFolder.selectedDevices[devices[i].deviceID] = false;
+            }
         };
 
         $scope.addFolder = function () {
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -69,7 +69,8 @@
                 <label translate for="folders">Share Folders With Device</label>
                 <p class="help-block">
                   <span translate>Select the folders to share with this device.</span>&emsp;
-                  <small><a ng-click="selectAllFolders()" translate>Select All</a> - <a ng-click="deSelectAllFolders()" translate>Deselect All</a></small>
+                  <small><a href="#" ng-click="selectAllFolders()" translate>Select All</a>&emsp;
+                    <a href="#" ng-click="deSelectAllFolders()" translate>Deselect All</a></small>
                 </p>
                 <div class="row">
                   <div class="col-md-4" ng-repeat="folder in folderList()">
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -67,7 +67,10 @@
             <div class="col-md-12">
               <div class="form-group">
                 <label translate for="folders">Share Folders With Device</label>
-                <p translate class="help-block">Select the folders to share with this device.</p>
+                <p class="help-block">
+                  <span translate>Select the folders to share with this device.</span>&emsp;
+                  <small><a ng-click="selectAllFolders()" translate>Select All</a> - <a ng-click="deSelectAllFolders()" translate>Deselect All</a></small>
+                </p>
                 <div class="row">
                   <div class="col-md-4" ng-repeat="folder in folderList()">
                     <div class="checkbox">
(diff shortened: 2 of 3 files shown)


imsodin:
Just talking about UI, not code: In my opinion this option must not look like regular devices. Neither do I see this function as a checkbox, it should be a button (it toggles selecting all or none devices, it's not a device itself). Also I don't like elements that automatically change it's function in general. In this case it doesn't make sense anyway, because both select and deselect need to be enabled if just one device is selected or unselected respectively.  
Therefore if I had to come up with a solution, I'd have two buttons clearly separated from the list of devices for selecting or deselecting all. Of course it could be one button that does something like "toggle de-/select all" which just switches between all or none, but the naming is confusing. And I'd place them above or below the list of devices. That in particular is just a proposition from someone who doesn't really know UI, but I am pretty set on not mixing this element with devices.

AudriusButkevicius:
Agree, I'd just have a thing that switches between select all deselect all somewhere next to the title text.

EserGozcu:
Agreed otherwise "Select All"  looks like another device name

uok:
Please keep in mind that average Joe only has [2-3 devices](https://data.syncthing.net/) in this list. Also how often do you share one folder with all devices, so I'm not sure if this is a GUI improvement? If you really think this is helpful I would only add this as small hyperlink after "... folder with. <TOGGLE LINK HERE>" Alternatively you can also set tab focus on first check box and go wild alternating tab key and space bar.

bahadir60:
Im going to try rewrite without function thanks for comments.

bahadir60:
@imsodin Can you explain your suggestion a bit more. I'm going to explain my idea, which part do you think unnecessary?  `clickOneFolder()` is for compare and understands which button should be active or deactive and also understands ticked on folders. `selectAll` and `deSellectAll` functions makes functionality to buttons. Why I'm using new js? When I tested code, I clicked continuously and randomly, and some point the scope stopped updating itself. [](https://github.com/angular/angular.js/wiki/Understanding-Scopes) suggest to use object base scope instead of primitive scope. I tried it and worked.

imsodin:
Something like this (unchecked code!):
```JS
        $scope.selectedDevices = function () {
            return $scope.selectedEntries($scope.currentFolder.selectedDevices);
        };

        $scope.selectedFolders = function () {
            return $scope.selectedEntries($scope.currentDevice.selectedFolders);
        };

        $scope.selectedEntries = function (obj) {
            sel = [];
            for (var id in obj) {
                if (obj[k]) {
                    sel.push(id)
                }
            }
            return sel;
        };
```
and then the condition is e.g.
```
selectedDevices().length === otherDevices().length
```
Or alternatively the `clickOneDevice` function could in-/decrement a counter (e.g. `currentFolder.selectedFolderCount`), so that you don't need the new functions above, the above condition is then just
```
currentFolder.selectedFolderCount === otherDevices().length
```

All just ideas, the main point is that the current change set is pretty big and there's quite a lot of loops and conditions and I believe they aren't necessary.

Design wise I like @uok's idea - in any case it should be as non-intrusive as possible. I wouldn't use color for buttons, that should be reserved for buttons, that we really need to pull attention to (especially the red one - it does nothing particularly dangerous or alarming).

bahadir60:
@imsodin thanks for detailed explanation, I took advantage of your comments. I want to make UI be attractive however I understand that it is unnecessary features of using buttons which are enabled/disabled and colored.
@uok  thanx as your suggest I use your code snippet.

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -42,7 +42,10 @@
           </div>
           <div class="form-group">
             <label translate for="devices">Share With Devices</label>
-            <p translate class="help-block">Select the devices to share this folder with.</p>
+            <p class="help-block">
+              <span translate>Select the folders to share with this device.</span>&emsp;
```
uok: this is the wrong text!

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -67,7 +67,10 @@
             <div class="col-md-12">
               <div class="form-group">
                 <label translate for="folders">Share Folders With Device</label>
-                <p translate class="help-block">Select the folders to share with this device.</p>
+                <p class="help-block">
+                  <span translate>Select the folders to share with this device.</span>&emsp;
+                  <small><a href="#" ng-click="selectAllFolders()" translate>Select All</a> - <a href="#" ng-click="deSelectAllFolders()" translate>Deselect All</a></small>
```
uok: I'm not sure if you need the `href="#"`part, maybe you can test this.

uok:
@bahadir60 ok, see code comments

gui/default/syncthing/folder/editFolderModalView.html
bahadir60: My bad, thanks

gui/default/syncthing/device/editDeviceModalView.html
bahadir60: Actually no needed. I am going to test it. Thanx

calmh: It's actually nice to have `href="#"` because it means you get the "hand" mouse cursor over the links. Otherwise we get just the normal text selection cursor that doesn't indicate an activity.

bahadir60: Yeah, good point, I didn't think like that, thanks

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1371,6 +1371,20 @@ angular.module('syncthing.core')
...
+                $scope.currentDevice.selectedFolders[id] = true;
+            });
+        };
+
+        $scope.deSelectAllFolders = function() {
+            Object.entries($scope.folders).forEach(entry =>{
+                let id = entry[1].id;
+                $scope.currentDevice.selectedFolders[id] = false;
+            });
+        };
```
AudriusButkevicius: This could have been setAllFolderSelectionStatus(bool)

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1694,6 +1708,20 @@ angular.module('syncthing.core')
...
+                $scope.currentFolder.selectedDevices[devices[i].deviceID] = true;
+            }
+        };
+
+        $scope.deSelectAllDevices = function() {
+            var devices = $scope.otherDevices();
+            for (var i = 0; i < devices.length; i++){
+                $scope.currentFolder.selectedDevices[devices[i].deviceID] = false;
+            }
+        };
```
AudriusButkevicius: Ditto about having a single function.

PR #5314, opened 2018-11-09
fix socket path and do not run syncthing as root

peterhoeg:
### Purpose

The documentation currently mentions /var/run/st.sock as a suggested location
for the socket file which has a couple of problems:

 - /var/run has been deprecated in favour of /run for a long time (both still
   work as most distributions set up a symlink)
 - when running as a system instance as a given user, the user will most likely
   not have permissions to write directly to /run

So we change the suggested socket path to /run/syncthing/st.sock and have the
systemd units take care of creating the directory. Further to #3616.

Secondly, it is possible to tell systemd to enable a unit for all users in one go which
will configure *user* services for every user:
systemctl --global enable syncthing.service

The problem with that is that root or other system users will then also try to
spawn a syncthing instance on login which is probably not what you want.

Instead only enable the *user* service for normal users.

### Testing

I have been using the 2nd commit for months. The first commit doesn't change any existing behaviour and thus cannot break anything.

### Documentation

Man page is updated accordingly.

### DIFF
--- a/etc/linux-systemd/system/syncthing@.service
+++ b/etc/linux-systemd/system/syncthing@.service
@@ -9,6 +9,7 @@ ExecStart=/usr/bin/syncthing -no-browser -no-restart -logflags=0
 Restart=on-failure
 SuccessExitStatus=3 4
 RestartForceExitStatus=3 4
+RuntimeDirectory=syncthing
 
 [Install]
 WantedBy=multi-user.target
--- a/etc/linux-systemd/user/syncthing.service
+++ b/etc/linux-systemd/user/syncthing.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=Syncthing - Open Source Continuous File Synchronization
+ConditionUser=!@system
 Documentation=man:syncthing(1)
 
 [Service]
--- a/etc/linux-systemd/user/syncthing.service
+++ b/etc/linux-systemd/user/syncthing.service
@@ -7,6 +7,7 @@ ExecStart=/usr/bin/syncthing -no-browser -no-restart -logflags=0
 Restart=on-failure
 SuccessExitStatus=3 4
 RestartForceExitStatus=3 4
+RuntimeDirectory=syncthing
 
 [Install]
 WantedBy=default.target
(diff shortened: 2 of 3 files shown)


calmh:
Systemd change sounds good.

Man pages come from the docs repository and are only auto generated here. 

The only info I find about `/run` when googling is that it's the Linux "early in boot" variant of `/var/run`, for things starting prior to `/var` being available. I don't think that really applies to us. Regardless, it's just an example path, perhaps it would be better to add a note to the page saying to use `/var/run`, `/run` or another appropriate directory depending on the OS and distribution. (People on BSD, Solaris, etc will be surprised at a suggestion of `/run`, while people like you who know the distinction will know to substitute `/run` instead of `/var/run` when appropriate.)

etc/linux-systemd/user/syncthing.service
```diff
@@ -1,12 +1,14 @@
 [Unit]
 Description=Syncthing - Open Source Continuous File Synchronization
+ConditionUser=!@system
```
imsodin: I agree running a user service as a system user doesn't make much sense, but I still don't like this condition much. The lesser reason is that I think it's superfluous (I don't think anyone does that) and more importantly, ccording to the man-page it silently skips the script, it doesn't fail it, i.e. there is no obvious feedback to the user that it failed.

etc/linux-systemd/system/syncthing@.service
```diff
@@ -9,6 +9,7 @@ ExecStart=/usr/bin/syncthing -no-browser -no-restart -logflags=0
 Restart=on-failure
 SuccessExitStatus=3 4
 RestartForceExitStatus=3 4
+RuntimeDirectory=syncthing
```
imsodin: As far as I can see we never use the runtime directory, unless the user manually creates a unix socket. Meaning in 99.9% this is just an empty directory. And even for those using the socket, it seems to be common practice to just place a single file directly into `run` according to what I see there on my system.

etc/linux-systemd/user/syncthing.service
calmh: What does "system user" mean here? I (and others) often run Syncthing as a low-uid user without a password and no login rights. This is perfectly fine.

st-review:
:robot: *beep boop*

PR #5321, opened 2018-11-12
cmd/syncthing: Improve logged device information on startup

imsodin:


calmh:
Why not?

imsodin:
Because to me it seems weird to print the connection information to ourselves, which is anyway always the default `[dynamic]`. And we already print our own ID during startup before this.

AudriusButkevicius:
I agree with this

calmh:
Fair enough on the addresses, but the name is sometimes useful. I'd rather leave it in.

imsodin:
I didn't expect so much discussion, I just wanted to get rid of this from my git workspace :)

Next try: Adding the name to the earlier "My ID" message, skipping the local device and not skipping devices without a name (IDs and connections are still relevant).

calmh:
It's a trivial change, so we can discuss it forever. You should have just snuck it in among something more complex.

For example, I wonder if there isn't someone out there parsing the "My ID" line at startup to get the device ID...

cmd/syncthing/main.go
```diff
@@ -828,9 +828,11 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 		pprof.StartCPUProfile(f)
 	}
 
+	myDev, _ := cfg.Device(myID)
+	l.Infof(`My device is "%v"`, myDev.Name)
```
calmh:
```suggestion
	l.Infof(`My name is "%v"`, myDev.Name)
```

imsodin: Fancy :)

calmh:
I fail to find more things to complain about or discuss here.

calmh:
Oh! Except the commit subject is now wrong of course. Caveat merger.

PR #5322, opened 2018-11-12
lib/model: Don't compare permissions if IgnorePerms is true

imsodin:
The fixed problem is likely the cause of https://forum.syncthing.net/t/0-14-52-file-deletion-issues/12436. "Likely" because I didn't reproduce (try to) myself, but the mechanism is clear: Local device ignores permissions but the remote does not, then different permissions let the consistency check fail when deleting the file and scheduls it for scanning. Scanning won't detect a change, because it does correctly ignore permissions.


AudriusButkevicius:
Test to catch this in the future maybe?

calmh:
Fast-merging because I want in the RC

PR #5326, opened 2018-11-13
lib: Introduce fs.IsParent (fixes #5324)

### LINKED ISSUE #5324: Incorrect warning when --home and path on same directory level
ST incorrectly opens a warning window, when config and shared folder have the same string prefix.

How to reproduce

1. `syncthing --home=/tmp/ABCCONFIG`
2. add folder `/tmp/ABC`
3. enjoy warning

### Version Information

Syncthing Version: v0.14.52, Linux (64 bit)
OS Version: LFS Linux

Source hint: 
`lib/model/model.go:273 warnAboutOverwritingProtectedFiles()`

imsodin:
https://github.com/syncthing/syncthing/issues/5324#issuecomment-438179326 by @calmh :
> Path prefixes are Hard™ :)

Make it harder to mess prefixes up by replacing the `strings.HasPrefix(..., ...+pathSeparator)` dance with `fs.IsParent(..., ...)`.

### DIFF
--- a/lib/fs/filesystem.go
+++ b/lib/fs/filesystem.go
@@ -199,12 +199,11 @@ func NewFilesystem(fsType FilesystemType, uri string) Filesystem {
 func IsInternal(file string) bool {
 	// fs cannot import config, so we hard code .stfolder here (config.DefaultMarkerName)
 	internals := []string{".stfolder", ".stignore", ".stversions"}
-	pathSep := string(PathSeparator)
 	for _, internal := range internals {
 		if file == internal {
 			return true
 		}
-		if strings.HasPrefix(file, internal+pathSep) {
+		if IsParent(file, internal) {
 			return true
 		}
 	}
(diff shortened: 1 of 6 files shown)


calmh:
There's the corner cases of `IsParent("/foo/bar", "/foo/bar")` and `IsParent("/foo/bar/", "/foo/bar/")` (and the other combos of trailing slashes).

Is a directory the parent of itself? I guess not. Yet for warnings like the one in the bug, we care if the directory is the same *or* a parent...

imsodin:
I'd say we should check for that case explicitly, as e.g. already done here: https://github.com/syncthing/syncthing/pull/5326/files#diff-0803d07e626e3aac840cbd7587d44bc5L738  
Does that seem reasonable?

calmh:
Yes, except I guess the equal check needs to be insensitive to trailing path seps?

imsodin:
We do (try to) ensure that paths do not end in a slash everywhere, so it shouldn't be necessary. Should I still do it just to be save?

calmh:
If we didn't before we can probably skip it now as well

Issue #5320, opened 2018-11-12
gui: Display rate limit


uok:
Currently a rate limit (global or per device) is nowhere displayed in the GUI. To find out which device is rate limited you need to edit one device after the other to check the settings. 
Here are some ideas how to display it:

Text note
![limit-text](https://user-images.githubusercontent.com/9367272/48348332-5d94c280-e681-11e8-879a-3873e9b4df2b.png)
Icon
![limit-icon](https://user-images.githubusercontent.com/9367272/48348342-61284980-e681-11e8-8941-b3e40db82979.png)
Text line
![limit-line](https://user-images.githubusercontent.com/9367272/48348346-64233a00-e681-11e8-8179-fb563e35adce.png)
Text info
![limit-muted](https://user-images.githubusercontent.com/9367272/48401520-b0c25000-e729-11e8-9b1c-0e6a49995ca1.png)

If no rate limit is set the icon/texts are hidden. 
My preference is icon version

calmh:
Haven’t looked yet, just to note that there is a precedent in the comment on ignore patterns reducing the local state. We might want to make the two things look roughly the same.

AudriusButkevicius:
Agree, we should just have the grey text beneath displaying the limits, both global (on "self" device) and per device.

uok:
Text info
![limit-muted](https://user-images.githubusercontent.com/9367272/48401296-15c97600-e729-11e8-8513-a9f759d75d3c.png)

uok:
ok, I added grey text version

AudriusButkevicius:
The Capalised Sentences look weird, but otherwise looks ok.

uok:
I copied the text from settings. If the writing is changed I guess it needs to be translated again? 
We could just go with a general "Rate limit: 123 KiB/s" in both cases, as it should be clear what it means.

AudriusButkevicius:
I guess if we are reusing an existing string is has to stay this way. Anyways, I'll let someone else chip in.

calmh:
Existing translations are good enough imho

calmh:
Woops

clintmoyer:
I submitted a change that details the rate limit in a tooltip.

https://github.com/syncthing/syncthing/pull/5327

uok:
@clintmoyer thanks for contributing
Please implement as it was agreed in this issue (grey text and wording). Tooltip is not needed.
Thanks!

uok:
I will work on PR

[...]

PR #5327, opened 2018-11-13
Add rate limit as tooltip

clintmoyer:
Shows the rate limit for global settings or per device as a tooltip when you hover over the corresponding real-time rate.

https://github.com/syncthing/syncthing/issues/5320

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -548,7 +548,7 @@
                   <tr>
                     <th><span class="fas fa-fw fa-cloud-download-alt"></span>&nbsp;<span translate>Download Rate</span></th>
                     <td class="text-right">
-                      <a href="#" class="toggler" ng-click="toggleUnits()">
+                      <a href="#" class="toggler" ng-click="toggleUnits()" tooltip data-original-title="Incoming Limit:{{config.options.maxRecvKbps}}KiB/s">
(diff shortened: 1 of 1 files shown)


gui/default/index.html
```diff
@@ -661,7 +661,7 @@ <h4 class="panel-title">
                     <tr ng-if="connections[deviceCfg.deviceID].connected">
                       <th><span class="fas fa-fw fa-cloud-upload-alt"></span>&nbsp;<span translate>Upload Rate</span></th>
                       <td class="text-right">
-                        <a href="#" class="toggler" ng-click="toggleUnits()">
+                        <a href="#" class="toggler" ng-click="toggleUnits()" tooltip data-original-title="Outgoing Limit:{{deviceCfg.maxRecvKbps}}KiB/s">
```
uok: you mixed up Send and Recv here

clintmoyer:
I added an AngularJS-Translate filter to the tooltip, it now translates per the browser language. I set the string to match our existing implementation [here](https://github.com/syncthing/syncthing/blob/v0.14.52/gui/default/assets/lang/lang-en.json#L146). Do we want the units to toggle? Right now the settings has it hard-coded to KiB/s.

calmh:
So this was superseded by #5328, sorry for the lack of coordination there

PR #5328, opened 2018-11-14
gui: Display rate limit (fixes #5320)

### LINKED ISSUE #5320: gui: Display rate limit

uok:
Display rate limit for own device and connected devices

limit-show

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -552,8 +552,13 @@
                         <span ng-if="!metricRates">{{connectionsTotal.inbps | binary}}B/s</span>
                         <span ng-if="metricRates">{{connectionsTotal.inbps*8 | metric}}bps</span>
                         ({{connectionsTotal.inBytesTotal | binary}}B)
+                        <small ng-if="config.options.maxRecvKbps > 0"><br/>
+                          <i class="text-muted"><span translate>Limit</span>:
+                            <span ng-if="!metricRates">{{config.options.maxRecvKbps*1024 | binary}}B/s</span>
+                            <span ng-if="metricRates">{{config.options.maxRecvKbps*1024*8 | metric}}bps</span>
+                          </i>
+                        </small>
                       </a>
-                      <small ng-if="config.options.maxRecvKbps > 0"><br/><i class="text-muted"><span translate>Limit</span>: {{config.options.maxRecvKbps}} KiB/s</i></small>
                     </td>
                   </tr>
                   <tr>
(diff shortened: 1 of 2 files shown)


uok:
Ok, this works nice!
How do I get a new string into the GUI and transifex for translation?

gui/default/index.html
```diff
@@ -551,8 +551,9 @@ <h4 class="panel-title">
                       <a href="#" class="toggler" ng-click="toggleUnits()">
                         <span ng-if="!metricRates">{{connectionsTotal.inbps | binary}}B/s</span>
                         <span ng-if="metricRates">{{connectionsTotal.inbps*8 | metric}}bps</span>
-                        ({{connectionsTotal.inBytesTotal | binary}}B)</span>
```
imsodin: Good find. These superfluous spans are missing in other places, e.g. the closing one for 
```
                                                <span tooltip data-original-title="{{localStateTotal.files | alwaysNumber | localeNumber}} {{'files' | translate}}, {{ localStateTotal.directories | alwaysNumber | localeNumber}} {{'directories' | translate}}, ~{{ localStateTotal.bytes | binary}}B">
```
in line 571 (or thereabouts).

gui/default/index.html
```diff
@@ -551,8 +551,9 @@ <h4 class="panel-title">
...
+                        ({{connectionsTotal.inBytesTotal | binary}}B)
                       </a>
+                      <small ng-if="config.options.maxRecvKbps > 0"><br/><i class="text-muted"><span translate>Limit</span>: {{config.options.maxRecvKbps}} KiB/s</i></small>
```
imsodin: I am pretty sure `maxRecvKbps` are actually bits, not bytes. And this should follow the rate limit unit, i.e. change unit on toggle as well.

gui/default/index.html
uok: fixed!

gui/default/index.html
uok: I entered "123" in the settings (it says KiB/s there) and "123" is displayed in the GUI line I added.

calmh:
Chill until next Wednesday, when magic scripts run and it happens automagically.

Or, look into `go run build.go translate` which updates json files based on the HTML source etc

uok:
ok, then this PR is good to go from my side.

uok:
good to go
Is commit squashing still needed?

calmh:
Squashing happens at merge 👍

PR #5337, opened 2018-11-22
all: Display list of locally changed items in UI (fixes #5336)

### LINKED ISSUE #5336: Show list of locally changed files for receive only folders
Currently one needs to press the revert button blindly - there's no indication what will get reverted. For example this came up in a case, where the revert button was always showing: https://forum.syncthing.net/t/revert-local-changes/12457  
This could be done similar to the out-of-sync/failed items lists.

imsodin:
### Purpose

#5336

### Testing

Manual testing in a test setup with a single file (I don't use receive only).

### Screenshots

I did not put much thought into the visual representation, feel free to suggest improvements.

screenshot_2018-11-22_09-18-36
screenshot_2018-11-22_09-18-57

### Documentation

Document `/rest/db/localchanged`.

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -90,6 +90,7 @@ type modelIntf interface {
 	Revert(folder string)
 	NeedFolderFiles(folder string, page, perpage int) ([]db.FileInfoTruncated, []db.FileInfoTruncated, []db.FileInfoTruncated)
 	RemoteNeedFolderFiles(device protocol.DeviceID, folder string, page, perpage int) ([]db.FileInfoTruncated, error)
+	LocalChangedFiles(folder string, page, perpage int) []db.FileInfoTruncated
 	NeedSize(folder string) db.Counts
 	ConnectionStats() map[string]interface{}
 	DeviceStatistics() map[string]stats.DeviceStatistics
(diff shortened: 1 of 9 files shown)


AudriusButkevicius:
I would call the endpoint local invalid files, and then we could use this to show ignored files and modified locally separately?

imsodin:
What is the reasoning for that?  
Semantically items ignored by patterns and items not propagated due to folder type have nothing in common. That both are considered invalid internally is in my opinion an implementation Derek
detail. I totally see the value of listing ignored items, but I think that's a separate issue.

AudriusButkevicius:
People asked for it before, so I want to avoid yet another API endpoint in the future. You can have a magical filter variable to filter what you are after I guess.

imsodin:
> I want to avoid yet another API endpoint in the future

That's the part I don't get - what is the problem with an additional endpoint? Wasn't it you that argued we should have proper endpoints for some config operations instead of one catch-all get/set the entire config endpoint? For me this is similar: Either get a big list of unrelated files with some magic (I love magic, but I try to avoid it when programming xD ) or lists of files for specific purposes.

acolomb:
Guess you just never came back to document the API endpoint @imsodin, huh? :wink:

I'm currently trying to fill in all blanks in the API docs.  But this one is a bit hard because even the model implementation does not have any comment on the method...

imsodin:
I guess that's a method where I hoped the function signature would speak for itself: It returns a list of locally changed items in a folder that doesn't share local changes with remotes (receive-only, receive-encrypted), with pagination.

PR #5343, opened 2018-12-02
gui: css: use lining nums for Raleway font

ix5:
The "old-style" numerals have ascenders and descenders that break the horizontal text alignment.
Nice for fancy magazine text, not so nice for matter-of-fact information-dense interfaces like syncthing web.
Solution: Use lining numerals via css font property.

### Purpose

Improve readability for web interface with Raleway font.

### Testing

View the change in the web interface.

### Screenshots

Old-style numerals(note the "9" and "3" in particular): 
syncthing-oldstyle
Lining numerals:
syncthing-lining

### Documentation
---

## Authorship
ix5

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -15,6 +15,7 @@ body {
 h1, h2, h3, h4, h5 {
     font-family: "Raleway", "Helvetica Neue", Helvetica, Arial, sans-serif;
     line-height: 1.25;
+    font-variant-numeric: lining-nums;
 }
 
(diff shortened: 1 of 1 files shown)


calmh:
This is a matter of opinion so without an objective right or wrong. That font and those numerals are only used in the headers, not in the tables that show most of the numerical data. As such I think this is harmless, and part of the "look" of Syncthing so I quite like it. It feels a little more whimsical and human than the straight numerals. So, thanks for the idea and attention, but I'm going to decline the change.

PR #5350, opened 2018-12-05
[WIP] Systemd units and ufw rules for relay and discovery

cromefire:
### Purpose

fixes #5115 

### Testing

 - [ ] stdiscosrv
 - [x] strelaysrv

### Documentation

It changes relay and discovery service configuration and execution, may need docs. 

## Authorship

Cromefire_ <tim.langhorst@outlook.de>
desbma <?>

PR #5351, opened 2018-12-06
Add hardening options to syncthing systemd services (fixes #5286)

### LINKED ISSUE #5286: Security: systemd service hardening
Feature request / request for comments:

systemd provides various options to increase security and control to the kernel and various filesystem paths:
https://gist.github.com/ageis/f5595e59b1cddb1513d1b425a323db04

I have used the following set in various other service / unit files:

```
      PrivateTmp=true
      ProtectSystem=strict
      # ProtectHome=true (probably would block syncthing from creating new folders et al)
      PrivateDevices=true
      ProtectKernelTunables=true
      ProtectKernelModules=true
      ProtectControlGroups=true
      NoNewPrivileges=true
```

strelaysrv already includes some of these options:
https://github.com/syncthing/syncthing/blob/7f55fbbe8415e9ff30db1aec02563590fa068650/cmd/strelaysrv/etc/linux-systemd/strelaysrv.service#L11-L14

syncthing does not include any at the moment:
https://github.com/syncthing/syncthing/blob/7f55fbbe8415e9ff30db1aec02563590fa068650/etc/linux-systemd/user/syncthing.service
https://github.com/syncthing/syncthing/blob/7f55fbbe8415e9ff30db1aec02563590fa068650/etc/linux-systemd/system/syncthing%40.service

There is an open PR for discosrv that includes similar options as strelaysrv:
https://github.com/syncthing/discosrv/pull/19/files
(stale PR, should be closed - at least in that repo)

---------

I would be interested in your thoughts or guidelines on improving service security by adding some of these options. Do you see a benefit? Would you recommend them - or even add them to the template service file included in the deb packages?

### Version Information

OS Version: Ubuntu / Debian / Linux

desbma:
This fixes #5286 by adding security options to Syncthing systemd service files.

See the related pull request for the relay service options, for the rationale on the choice of options : https://github.com/syncthing/syncthing/pull/5350#pullrequestreview-182004704

In summary :
* we want to reduce the scope of what the executable can do, because syncthing will likely run on public facing IP, so is a target for attack
* we want to restrict capabilities, but not overdo it because the service will not run as root anyway
* we can not be as restrictive as with the relay service because its uses, and particularly the paths it will read/write to are not predictible

Compared to the relay hardening options:
* we use `ProtectSystem=full` instead of the more restrictive `ReadOnlyPaths=/  ReadWritePaths=/var/lib/syncthing-relaysrv` because we don't know where syncthing may need to write
* we can not use `ProtectHome=true`, because syncthing may need to read/write files in `/home`
* we can not use `PrivateDevices=true` either, because otherwise systemd throws this error:
```
syncthing.service: Failed to drop capabilities: Operation not permitted
syncthing.service: Failed at step CAPABILITIES spawning /usr/bin/syncthing: Operation not permitted
```
I am not 100% sure, but I presume this is because the systemd user process does not have the required capabilities to setup the `/dev` mount namespace.

I have tested it in the following configurations:
* `syncthing.service`: Ubuntu 18.04, Syncthing 0.14.53
* `syncthing@.service`: Arch Linux, Syncthing 0.14.52

### DIFF
--- a/etc/linux-systemd/system/syncthing@.service
+++ b/etc/linux-systemd/system/syncthing@.service
@@ -10,5 +10,12 @@ Restart=on-failure
 SuccessExitStatus=3 4
 RestartForceExitStatus=3 4
 
+# Hardening
+ProtectSystem=full
+PrivateTmp=true
+SystemCallArchitectures=native
+MemoryDenyWriteExecute=true
+NoNewPrivileges=true
+
 [Install]
 WantedBy=multi-user.target
--- a/etc/linux-systemd/user/syncthing.service
+++ b/etc/linux-systemd/user/syncthing.service
@@ -8,5 +8,12 @@ Restart=on-failure
 SuccessExitStatus=3 4
 RestartForceExitStatus=3 4
 
+# Hardening
+ProtectSystem=full
+PrivateTmp=true
+SystemCallArchitectures=native
+MemoryDenyWriteExecute=true
+NoNewPrivileges=true
+
 [Install]
 WantedBy=default.target


calmh:
Neat

PR #5366, opened 2018-12-11
lib/model: Attempt to unflake Deregister tests (fixes #5362)

### LINKED ISSUE #5362: lib/model: TestDeregister* tests are bad
screen shot 2018-12-10 at 18 49 07

calmh:
### Purpose

Set up the event subscription earlier, try to be less dependent on timeouts.

### Testing

I can't make it fail, but lets have the slow ass builders run it a few times...


AudriusButkevicius:
However, still seems to fail :(

calmh:
Yeah wonder if the test is fine and the thing under test bad ...

calmh:
I suspect it's fine, just that the events takes >1 s to come when the builder VM, and the machine it's running on, is loaded. Added some debugging so we can possibly see that in the future.

PR #5367, opened 2018-12-11
build.go: avoid rebuilding cgo.go

Mic92:
### Purpose

Fix permission problem when building syncthing from a system-wide toolchain.

Also see https://github.com/golang/go/issues/24674
and https://github.com/NixOS/nixpkgs/issues/51825

### Testing

build is wip

### Documentation

not required

## Authorship

ok

### DIFF
--- a/build.go
+++ b/build.go
@@ -450,7 +450,7 @@ func build(target target, tags []string) {
 
 	rmr(target.BinaryName())
 
-	args := []string{"build", "-i", "-v"}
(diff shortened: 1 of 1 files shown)


calmh:
As far as I can tell this boils down to an issue in how the Go compiler and runtime is built and packaged... However, the `-i` to `go build` is pretty much a no-op nowadays so if this improves things for you, sure, why not.

PR #5379, opened 2018-12-16
lib/config: Add encrypted string type, encrypt API key (fixes #5378)

### LINKED ISSUE #5378: Protect secrets in the config
For #5376 we need to store a secret (the bind password) in the config. Storing this in plain text is no good. Hence we need a way to safely store secrets in the configuration. When having that, we might as well apply it to the API key, until such time the API key system is revamped.

calmh:
### Purpose

This protects the API key against prying eyes, and more generally enables storing secrets in the configuration.

When a config entry is set to be an encrypted string it is sealed in a NaCL secretbox with the SHA256 of our private key as the secret. This means the secrecy of the secrets in the config is equivalent with the secrecy of the private key.

Config values can be set in plain text in the config as usual, and are sent in plain text over the JSON API. This is necessary because the GUI needs to be able to show the API key. When the config is saved, the value becomes encrypted. This is indicated by getting a large base64 blob in an "encrypted" attribute. That is,

    <apikey>foo</apikey>

becomes

    <apikey encrypted="..."></apikey>

thus indicating for both humans and machines what's going on. For what this actually looks like, see testdata/v29.xml.

### Testing

There are a couple of test on the encryption and serialization, including that we can still load plaintext values.

### Documentation

I'll update the config docs to mention this prior to release.


calmh:
(Maybe hold off on this while we consider if we still want the API key readable after all. Going to need the rest of it for the other thing, though.)

AudriusButkevicius:
Not a fan of the API key change. If I can read the config, I can probay read the private key.

calmh:
Right so this is of limited utility for the API key and there are consumers of it in its current state... So closing. The rest of the code will make a reappearance in another PR though

PR #5381, opened 2018-12-16
gui: Make sync status accessible by screen readers (fixes #2697)

### LINKED ISSUE #2697: Folder status "box" character not readable by screen readers / not color blindness friendly
When the UI is zoomed to a high enough level the folder status text, e.g. "Up to Date", is replaced by a "box" character. This is not pronounceable by some screen readers, such as [NVDA](http://www.nvaccess.org/). I have very little HTML/CSS experience but I assume there are methods that can be employed so that screen readers are still given text to read out in this circumstance.

uok:
Make sync status accessible for screen readers and colorblind users

- [x] add icons instead of block symbol
- [x] add aria labels


madduck:
I can't really compile at the moment, as I am travelling without my gear, but I looked at the diff and it all makes sense. Thanks a lot! Looking forward to this merge.

PR #5383, opened 2018-12-17
lib/rc: Fix hangups when Syncthing process ends unexpectedly

imsodin:
### Purpose

If the syncthing process exits before startup integration tests are hanging indefinitely as `rc.Process.AwaitStartup` is blocked and also panics somewhere else in the code don't propagate to the test (don't know exactly why, it's not the meta lock issue). So I replaced the lock/cond mechanism with channels and monitor the Syncthing process instead of only checking it when `rc.Process.Stop` is called.

Result: Fast abortion and indication of panics in integration tests.

### Testing

Ran integration tests with "artificial panics".

### DIFF
--- a/lib/rc/rc.go
+++ b/lib/rc/rc.go
@@ -22,7 +22,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strconv"
-	stdsync "sync"
 	"time"
 
 	"github.com/syncthing/syncthing/lib/config"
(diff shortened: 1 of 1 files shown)


calmh:
tl;dr right now but the reason panics don't propagate is that these things run syncthing as a subprocess... it panicking just means a log entry and an exit. But there is some code in the "stop and check" path that stops the process, where it looks in the output for race detector or panic output.

PR #5384, opened 2018-12-17
all: Become a Go module (fixes #5148)

### LINKED ISSUE #5148: Consider switching to Go modules
Now that Go 1.11 has landed we have Go modules support in the standard tooling, [with forward compatibility](https://golang.org/doc/go1.11#modules) (what we do now will work in the future). Our current preferred vendoring tool [has been discontinued](https://twitter.com/FiloSottile/status/1033216762454503424). I propose that we switch internally to using modules, while retaining the vendor directory for pre-1.11 (or in-GOPATH) builds. Here's a short fictive FAQ on how I envision that.

### What's the migration process?

Essentially `GO111MODULE=on go build ./...`. The new mode kicks in, creates a `go.mod` file based on our existing vendor manifest. We delete delete that manifest. The vendor directory remains.

The build script today attempts to set up a temporary GOPATH and build there if called with the source outside of GOPATH. We disable this when the compiler is 1.11+.

(In the future I want to clean up our package paths a bit as well, but realistically after Go 1.12.)

### What do I need to do as a developer?

Nothing, urgently. When building within GOPATH Go 1.11 behaves exactly like previous versions, and previous versions will continue working as well because the dependencies are still vendored.

### What's the point then?

When building outside of GOPATH (with Go 1.11) the new behavior kicks in and everything still just works. This is neat for build servers and stuff as well.

Cleaner dependency/vendor updates.

### How do I update a dependency?

This is when you need to actually interact with the module system. You do it like this;

1. Enable the module system. Do this by using Go 1.11 and not having the source in GOPATH, or by setting `GO111MODULE=on`.
2. `go get` your updated dependency from within the `syncthing`  source. The `go mod` file is updated appropriately.
3. `go mod vendor` to update the vendored dependencies.

### How do I hack on a dependency without pushing it etc?

This is a bit rough. You can do it like you would today (be in GOPATH, remove it from the vendor dir), or you can add a `replace` directive to `go.mod` that points a certain package towards a local path on your disk.

### How does this work with my favorite IDE?

Consult your IDE vendor.

calmh:
### Purpose

Use the new hotness. Build cleanly outside `GOPATH`. Use modern tools for vendoring.

The changes in more detail:

- Our dependencies are now declared in go.mod & go.sum. Managing these is done using `go get`, `go get -u`, `go mod tidy`, `go mod vendor` while in the modules mode (`GO111MODULE=on` or outside `GOPATH`).

- The vendor directory is still populated, using `go mod vendor`. This means there is no difference in building stuff for those still using `GOPATH` etc. In the long run, Go 1.12+, we can drop the vendor dir.

- The non-Go dependencies we have for `go generate` (protobuf schemas), and the XDR builder, aren't easily accessible as modules. Instead I hacked the build script to clone and check out the stuff we need, if it isn't already present, when we run `go run build.go proto`. This isn't done very often by most developers, so I think it's fine.

- I removed the stuff in build.go for juggling GOPATH and building outside of it. If you're outside GOPATH today and on a modern compiler, the module support will kick in instead.

### Testing

It still builds for me.


calmh:
So there's a hilarious amount of protocol failures due to updating the protobuf package, so that's interesting and merits investigation...

calmh:
This is not user facing, and should really only affect those juggling dependencies or wanting to work without gopath. Given that, is anyone strongly *opposed* to this change, @AudriusButkevicius @imsodin ?

AudriusButkevicius:
This is probably the future, so I am ok with that. I am not planning to review the diff as I presume it's a massive noop change.

calmh:
Welcome to the future. Now to fix all the unexpected build failures...

PR #5393, opened 2018-12-18
test: Update configs

imsodin:
Running integration tests leaves you with changed checked in config files. This commits the state at the end of a successful run. There is one change (versioning) that looks like it's more than just adding the default value for a new config item - I'll comment in the code.

### DIFF
--- a/test/h2/config.xml
+++ b/test/h2/config.xml
@@ -1,11 +1,11 @@
 <configuration version="28">
-    <folder id="default" label="" path="s2" type="readwrite" rescanIntervalS="15" fsWatcherEnabled="false" fsWatcherDelayS="10" ignorePerms="false" autoNormalize="true">
+    <folder id="default" label="" path="s2" type="sendreceive" rescanIntervalS="15" fsWatcherEnabled="false" fsWatcherDelayS="10" ignorePerms="false" autoNormalize="true">
         <filesystemType>basic</filesystemType>
         <device id="I6KAH76-66SLLLB-5PFXSOA-UFJCDZC-YAOMLEK-CP2GB32-BV5RQST-3PSROAU" introducedBy=""></device>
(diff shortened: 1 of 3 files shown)


test/h2/config.xml
```diff
@@ -1,11 +1,11 @@
 <configuration version="28">
-    <folder id="default" label="" path="s2" type="readwrite" rescanIntervalS="15" fsWatcherEnabled="false" fsWatcherDelayS="10" ignorePerms="false" autoNormalize="true">
+    <folder id="default" label="" path="s2" type="sendreceive" rescanIntervalS="15" fsWatcherEnabled="false" fsWatcherDelayS="10" ignorePerms="false" autoNormalize="true">
         <filesystemType>basic</filesystemType>
         <device id="I6KAH76-66SLLLB-5PFXSOA-UFJCDZC-YAOMLEK-CP2GB32-BV5RQST-3PSROAU" introducedBy=""></device>
         <device id="MRIW7OK-NETT3M4-N6SBWME-N25O76W-YJKVXPH-FUMQJ3S-P57B74J-GBITBAC" introducedBy=""></device>
         <device id="373HSRP-QLPNLIE-JYKZVQF-P4PKZ63-R2ZE6K3-YD442U2-JHBGBQG-WWXAHAU" introducedBy=""></device>
         <minDiskFree unit="%">1</minDiskFree>
-        <versioning></versioning>
+        <versioning type="staggered"></versioning>
```
imsodin: Not quite sure why this change is here. Tests run fine with it though.

calmh: There are integration tests that change the versioning, I think

calmh: When I want to upgrade the configs I usually check out the clean version from the repo then just manually `syncthing -home h1`, ^C, once per dir.

imsodin: Does that mean the versioning integration test should be "fixed" to revert this change on exit? Or is it just expected that these configs are changed after integration tests?

Issue #5340, opened 2018-11-29
Incorrectly out of sync, sequence number oddness.


calmh:
It seems there might be an occasional issue with the index sending. I noticed this in my home setup. Two devices, "kvin" and "pseudo"; kvin thinks pseudo is out of sync missing one file, pseudo thinks it has everything, but the file is missing.

kvin side:

<img width="1164" alt="screen shot 2018-11-29 at 09 15 05" src="https://user-images.githubusercontent.com/125426/49208702-c20d8c80-f3b8-11e8-87ed-22de4bab7431.png">

<img width="1174" alt="screen shot 2018-11-29 at 09 15 13" src="https://user-images.githubusercontent.com/125426/49208713-c9349a80-f3b8-11e8-8601-b03af85b3f9f.png">

(note 2702 files locally, pseudo missing one pdf)

pseudo side:

<img width="1175" alt="screen shot 2018-11-29 at 09 15 20" src="https://user-images.githubusercontent.com/125426/49208739-dbaed400-f3b8-11e8-8894-f1bb9e264a3c.png">

(2701 files locally)

Checking the file via rest shows it present on kvin but not pseudo, which matches the above:

```
jb@kvin:~ $ curl ... 'http://localhost:8384/rest/db/file?folder=documents&file=Kastelo/Ekonomi/2018+Bokföring/INV04871726_A00552008_11252018.pdf'
{
  "availability": null,
  "global": {
    "deleted": false,
    "ignored": false,
    "invalid": false,
    "localFlags": 0,
    "modified": "2018-11-26T07:03:54+01:00",
    "modifiedBy": "SYNO4VL",
    "mustRescan": false,
    "name": "Kastelo/Ekonomi/2018 Bokföring/INV04871726_A00552008_11252018.pdf",
    "noPermissions": false,
    "numBlocks": 1,
    "permissions": "0644",
    "sequence": 902,
    "size": 71947,
    "type": 0,
    "version": [
      "SYNO4VL:1"
    ]
  },
  "local": {
    "deleted": false,
    "ignored": false,
    "invalid": false,
    "localFlags": 0,
    "modified": "2018-11-26T07:03:54+01:00",
    "modifiedBy": "SYNO4VL",
    "mustRescan": false,
    "name": "Kastelo/Ekonomi/2018 Bokföring/INV04871726_A00552008_11252018.pdf",
    "noPermissions": false,
    "numBlocks": 1,
    "permissions": "0644",
    "sequence": 902,
[...]

PR #5394, opened 2018-12-18
lib/db: Update item by item and strict sequence increase

imsodin:
### Purpose

I don't have a smoking gun, that it's related to #5340, but it's motivated by it:  
Currently updates happens by passing multiple slices of fileinfos around to several functions that update the db in various places. This has recently be significantly improved by @calmh, but I still don't like it in principle: It uses more memory and I think it's also more fragile (e.g. when process is interrupted mid-update). Another related issue is how sequences are stored and updated (also see https://forum.syncthing.net/t/question-about-metadata-buckets-and-sequence/12579). This PR does not change how they are saved (lacyness), but the update now happens in only one place and in a dedicated function for consistency. <strike>I also added a panic to prevent the sequence from ever decreasing. Maybe that panic should be hidden behind `shouldDebug`.</strike> Undid that, see next post.

Also I recommend looking at it commit by commit. The first is purely cosmetic to make further diffs smaller, the latter contain the main changes explained above. It might even make sense to commit them separately to master in the end.

### Testing

Unit and integration testing, no production testing. <strike>One unit test that depends on stored db data fails, because the db data is erroneous (we didn't care about sequences in testing back then). I can fix that if this change meets favorable opinions, it's just cumbersome, so I skipped it for now.</strike> See below.

### DIFF
--- a/lib/db/blockmap.go
+++ b/lib/db/blockmap.go
@@ -55,12 +55,24 @@ func (m *BlockMap) Add(files []protocol.FileInfo) error {
 }
 
 // Update block map state, removing any deleted or invalid files.
-func (m *BlockMap) Update(files []protocol.FileInfo) error {
+func (m *BlockMap) Update(files []protocol.FileInfo) {
+	fn, finished := m.updateFn()
+	for _, f := range files {
+		fn(f, nil, false)
+	}
+	finished()
+}
+
+func (m *BlockMap) updateFn() (func(nf protocol.FileInfo, ef FileIntf, efOk bool), func()) {
 	batch := new(leveldb.Batch)
 	buf := make([]byte, 4)
 	var key []byte
-	for _, file := range files {
+	return func(file protocol.FileInfo, ef FileIntf, efOk bool) {
 		m.checkFlush(batch)
+
+		if efOk {
+			m.discard(ef.(protocol.FileInfo), key, batch)
+		}
 
 		switch {
 		case file.IsDirectory():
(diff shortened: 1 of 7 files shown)


imsodin:
I removed the panic that checked for monotonically increasing sequences. That's only true for local updates and updates from remotes, that (correctly) support delta indexing. It is not the case e.g. when recalculating metadata from database.

calmh:
So first part is fine. Second does subtle stuff with the sequence numbering that I'm not sure I understand (instead of the clear and mutex-locked `nextSeq` we now read the sequence once, increment *that* a bunch of times and update the stored sequence number ... somewhere else?). Third part introduces a bunch of complex closures with very abbreviated names. Maybe it makes sense, maybe it doesn't, I just don't follow it at the moment and the closure stuff increases complexity above my pain threshold...

calmh:
>  I just want to do all db transactions for a single fileinfo at once instead of passing the list of fileinfos to multiple functions who do operations one after the other.

This might be a good thing. I think the essential part, then, is to make sure that all changes for a given FileInfo are committed in the same "transaction" (batch write), so that at an iterator doesn't see a partial result if it happens to be looking at the time.

I'm still not convinced about the closures. The `m.updateFn` is used exactly once; why not inline it and skip the closure?

I have a different thing in a branch where I add *another* mutex to the database layer - a `localDeviceMutex` which is an RWMutex held for all operations related to the local device. That is, the Update() method will grab it with Lock()/Unlock() if device == protocol.LocalDeviceID and the iterators and readers will grab it with RLock()/RUnlock() when looking at local files and the global set. I'm not sure if this is elegant or desperate. The intention is to ensure a consistent view of local files and stuff like the sequence index...

imsodin:
Superseded by https://github.com/syncthing/syncthing/pull/5441

PR #5396, opened 2018-12-20
Allow setting the go command in build.go

rumpelsepp:
### Purpose

The `go` command is hardcoded in the "build system". Let's make this configurable, since my debian has `go 1.10` by default and I installed `go1.11` as `go1.11` in my home directory.

### Testing

This command works.

```
$ go run build.go -gocmd=go1.11.1
```

Also, I updated the dependencies, since there was an error:

```
go: verifying github.com/prometheus/procfs@v0.0.0-20171226183907-b15cd069a834: checksum mismatch
        downloaded: h1:jQVavQNtg1+b1/lxzhVA1JS3NYnCTo4b5Mo6z1N6apo=
        go.sum:     h1:HRxr4uZnx/S86wVQsfXcKhadpzdceXn2qCzCtagcI6w=
exit status 1
```

Feel free to drop the dependency update.

### Documentation

```
rumpelsepp@kronos ~/P/j/syncthing (master)> go run build.go -help
Usage of /tmp/go-build426115166/b001/exe/build:
  -debug-binary
        Create unoptimized binary to use with delve, set -gcflags='-N -l' and omit -ldflags
  -goarch string
        GOARCH (default "amd64")
  -gocmd go
        Specify go command
  -goos string
        GOOS (default "linux")
  -installsuffix string
        Install suffix, optional
  -no-upgrade
        Disable upgrade functionality
  -pkgdir go build
        Set -pkgdir parameter for go build
  -race
        Use race detector
  -tags string
        Extra tags, space separated
  -version string
        Set compiled in version string (default "v0.14.55-rc.2+5-gf4e42934-dirty")
exit status 2
```


calmh:
Windows failure isn't your fault, I need to repair the builder apparently.

I'm skeptical to the dep update; it's large, at the moment unnecessary, and should be accompanied by a vendor update. The checksum error you're seeing is a Go bug, you need 1.11.4 as 1.11.1-3 did the wrong thing for module zips with symlinks in them.

(I generally build with odd Go version by doing `PATH=/usr/local/go1.10/bin:$PATH ./build.sh` for example, obviating this change.)

rumpelsepp:
Redefining `PATH` seems odd to me. You get a `go1.11.X` binary by using `go get golang.org/dl/go1.11.1` or so. Since `build.go` has `go` hard coded, this doesn't work either.

I removed the dependency update.

PR #5404, opened 2018-12-24
lib/config, cmd/syncthing: Improve LDAP authentication (fixes #5376)

### LINKED ISSUE #5376: Improve LDAP authentication
It's nice that we now have LDAP authentication. Now that we have it, and in the spirit of taking the whole arm when offered the hand, there are a couple of improvements we could make. To make it more enterprise, though that wasn't the target of the initial implementation. ;)

Currently the mode of operation is LDAP bind by pattern. That is, the user enters "jb", we have a pattern "CN=%s,OU=users,DC=kastelo,DC=net", we combine the two and attempt a bind. This mode of operation is perfect against a small OpenLDAP where 1) every user in the system should have Syncthing access and 2) the name in the DN (CN=... above) is what the user enters. Both of these two assumptions are typically false in larger setups or in AD.

A more complete implementation would (at least as an optional mode of operation) have:

1) a bind user, which has permissions to search the directory
2) a search pattern where we plug in the username the user entered and match it against logon name (AD terminology) or email
3) a group membership restriction, which in its simplest incarnation can be coded into the user search pattern above as an `&& memberOf = ...` type of thing.
4) and then do the bind against the found user, like we do today.

All in all this is about five more lines of code, excl the extra config option (I just implemented it in another product).

calmh:
This adds an extra mode to the LDAP setup, useful for authentication against AD and enforcing group memberships.

As this requires storing a password in the config, we add support for encrypted secrets in the config. The password is encrypted using the hash of our secret key as the key, making the password not readable for anyone not possessing our secret key.

Additionally the advanced config editor learns to edit the LDAP configuration.

screenshot 2018-12-24 at 11 48 03


calmh:
The GUI stuff is in `advanced`, where it gets its own section since it's a separate non-gui section in the actual config.

The point of the encryption is that sharing your config.xml, accidentally or intentionally, does not leak the password. It is not intended as protection against someone who has free access to the config dir. Adding a separate file adds a bunch of new moving parts to config saving / loading which I don't see a reason for.

imsodin:
> The GUI stuff is in `advanced`, where it gets its own section since it's a separate non-gui section in the actual config.

Didn't realize that was already the case before this PR - does it have any application outside of GUI? Well doesn't matter anymore, hardly warrants a config transition (if at all).

So this is basically lgtm from my side, but I'd rather have someone else do the review honors, because I didn't really review it.

cmd/syncthing/gui_auth.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/syncthing/syncthing/lib/rand"
 	"github.com/syncthing/syncthing/lib/sync"
 	"golang.org/x/crypto/bcrypt"
-	"gopkg.in/ldap.v2"
+	ldap "gopkg.in/ldap.v2"
```
imsodin: Is this due to the module changes?

calmh: newer goimports style

st-review:
:robot: *beep boop*

PR #5405, opened 2018-12-25
gui: Enable large blocks by default, add to config dialog

calmh:
### Purpose

It's been ~9 months since we introduced large blocks support; the odds of a cluster being mixed between 0.14.45- and 0.14.56+ should be quite small. This enables it for newly created folders.

Also moves a couple of `</div>` that were out of balance and reindents accordingly, so try a white space insensitive diff when reviewing...

screenshot 2018-12-25 at 14 39 26


Catfriend1:
I've already read multiple threads on the forum about enabling large blocks causing trouble. Assuming I've got synchthing versions >=0.14.48 in place (most are 0.14.54's) and I don't clear the database but post-enable large blocks on all nodes and all existing folders:
- will files get rehashed automatically?
- will files get transferred even if they were the same already?
- are problems expected that are already known? 
So in summary I'd like to know if it can be enabled later after a folder had been created in the past or if it should only get enabled for folders newly created.

wweich:
The large blocks setting enables variable block size. Files won't get rehashed, so files which would get higher block sizes are kept in DB with small block sizes.
You can enable it for existing folders. Only new / changed files will get variable block sizes. But if it doesn't get enabled on all nodes at the same time, and there are changes to a file on two devices (one with large blocks, one without) you probably end up transferring the whole file, as not a single block matches.

calmh:
Which problems are these?

Catfriend1:
Thanks for the explanation. I'm currently thinking about enabling it on my nodes and making large blocks also default after a user notice question on android telling that this setting if turned on needs to be consistent across all nodes.

imsodin:
In my opinion we could just enable it for all folders, also existing once. If we want to be cautious, do it with the usual banner in the web UI where the user needs to confirm or deny switching and with a warning about the incompatibility with <0.14.45. However I'd say this is unlikely enough to just enable it without notice. Just my 5 cents, I don't care much either way (I do like the feature though :) ).

Catfriend1:
I basically agree with @imsodin 
Trying out large Blocks the last days, I noticed the already mentioned aspect about files getting retransferred and/or conflicts getting created for files bigger than 256 MByte when one side gets a database reset (I did this on one device to resolve other problems remaining in the database from the far past) and found the creation of those conflicts "duplicates" of the same original file a little bit nasty to clean up. Would there be any way to implement a "hint" for a syncthing instance, that starts a fresh database, to notify other nodes about that so they handle block sizes correctly without creating conflicts?
I guess the following happened:
- Folder F on Device A and B was in sync without large blocks.
- I've enabled large blocks on both devices simultaneously. > Folder still in sync.
- For another misbehaving folder G, I reset the database on device B and kept the folder F config on B including the large blocks setting.
- Rehash on B took place, large blocks were put into the database (folder F).
- Device A and B exchanged index for folder F.
- Device B got "small legacy blocks" from A for folder F while it had (itself) large blocks from the rehash.
- Sync-conflict-XXXXX files got created and needed to be cleant up manually.

AudriusButkevicius:
so wazzup wit dis?

calmh:
I'm awaiting either acceptance or actionable review comments. I don't think just enabling it everywhere is the right thing at this point.

imsodin:
What's the intention of adding it to the folder edit dialog? As I remember the only reason for making it configurable is that it isn't backward compatible - other than that there is no reason not to enable it in general. Or more general question: What would be the next steps to "streamline" large blocks?

calmh:
My thinking behind adding it to the dialog was that since it's a backwards-incompatible thing there should be an easy opt out.

With default-on most new folders would get it set. The trouble with flipping it on everywhere is that there'll be issues on existing folders shared between v1.0.1 and v1.0.2 (if 1.0.2 flips it on automatically). 
Especially things like NAS:es (and Android sometimes) tend to version lag quite a while, leaving even people who *want* the latest with mixed version clusters. Screwing up their existing stuff is unkind.

After a while with default-on it's easier to set it to always on, and remove the option completely. Certainly some will anyway be taken by surprise, but it'll be fewer.

imsodin:
Thanks for the explanation!

I'd add a help symbol/link to https://docs.syncthing.net/advanced/folder-uselargeblocks.html, potentially together with some sort of pointer about recommending to activate and compatibility.

Also on the screenshot "Full Rescan Interval" is bold, while it shouldn't be (subentry of "Scanning"), but I can't see the reason in the html (maybe the screenshot is outdated?).

AudriusButkevicius:
I say, cut a 2.0.0 and break everyone. We were rolling in that 0.x stuff way too long, time to roll forward.

calmh:
Yeah but you say that for every change ever and it still ain't happenin'

@imsodin I'll check it out & sort it

alex2108:
I think the problem with sync conflicts that @Catfriend1 mentioned should be somehow handled, otherwise there will be a lot of complaints about conflicts after merging this, resetting database or syncing files with rsync and adding the folder afterwards is not that uncommon I think.

Another thought: could this even happen when the folder always had large blocks on all devices? Hysteresis could lead to mismatch of the block size.

calmh:
This is specifically for new folders, so I don’t so that being a common failure case.

Catfriend1:
What will happen if a user shares an existing and already scanned folder from node A to B while A hasn't large blocks enabled and B sees this a new folder in the config and automatically turns large blocks on?

calmh:
You'll get a mismatch in the settings.

What we could do, maybe, is make large blocks "infectious". Send a boolean in the initial negotiation, if we see it we enable large blocks regardless. It doesn't help much with earlier versions, but it would help in future conversions.

Also keep in mind that the effect of mismatching settings is just a less efficient sync and some wasted bandwidth. Horror stories about sync conflicts are about special case scenarios like setting up a new cluster with devices that already have matching data but mismatching settings.

Catfriend1:
I've also thought of proposing the infectious approach when nodes see each other next time connecting those days. If possible, that might help a lot during migration and can be dropped later in a future version.

calmh:
While convenient though, that would make it extremely frustrating and almost impossible to turn off large blocks, should one want to do that.

Catfriend1:
Should we enable large blocks for new folders on Android also?
What will happen if a user shares a folder from Syncthing 1.0.0 (noLargeBlocks) to an Android instance running Syncthing 1.0.1 and we suggest enabling the blocks, is there any way for the other side on Android to know that a shared folder has largeBlocks enabled or disabled to propose the right setting state to the user when adding?

PR #5415, opened 2018-12-29
Improve messages when free space is running out

ziotom78:
### Purpose

Use a better error message when disk space is running low. Without this PR, the user gets the error `error on folder XXX: 0.578253 % < 1 %`, which is not helpful. With this PR, the error becomes

    error on folder XXX: not enough free space, 0.578253 % < 1 %

### Testing

I successfully checked that all the tests ran by `go test` pass.

### Documentation

N/A

## Authorship

Maurizio Tomasi <ziotom78@gmail.com>

### DIFF
--- a/lib/config/size.go
+++ b/lib/config/size.go
@@ -85,10 +85,10 @@ func checkFreeSpace(req Size, usage fs.Usage) error {
 	if req.Percentage() {
 		freePct := (float64(usage.Free) / float64(usage.Total)) * 100
 		if freePct < val {
-			return fmt.Errorf("%f %% < %v", freePct, req)
+			return fmt.Errorf("not enough free space, %f %% < %v", freePct, req)
 		}
 	} else if float64(usage.Free) < val {
-		return fmt.Errorf("%v < %v", usage.Free, req)
+		return fmt.Errorf("not enough free space, %v < %v", usage.Free, req)
 	}
 
 	return nil


imsodin:
Did you see the error in a real log? The actual error messages are 

```golang
	return fmt.Errorf("insufficient space in %v %v", fs.Type(), fs.URI())
```
https://github.com/syncthing/syncthing/blob/master/lib/config/folderconfiguration.go#L287

```golang
	return fmt.Errorf("insufficient space on home disk (%v): %v", path, err)
```
https://github.com/syncthing/syncthing/blob/master/lib/config/wrapper.go#L506

Where `err` is the error you changed. As far as I see the final error message is now redundant.

ziotom78:
> Did you see the error in a real log?

Yes, I did. Yesterday my Samsung tablet refused to synchronize with my laptop, showing this error message and nothing else. This is the reason that prompted me to dig into the code and issue this PR.

imsodin:
Is your tablet on a version before 0.14.52? If not I fail to see how that error message has happened...

ziotom78:
> Is your tablet on a version before 0.14.52

Yep. Sorry, I am running 0.14.51, which is the latest version available on the Google Play Store as today. Do you mean that the problem was already fixed in master? How should I updated Syncthing, then?

Sturmlicht:
Updating sync Android requirs a new release to be waited for as it says updates are disabled by the admin/maintainer. I've solved my issue in this by switching over to syncthing-fork which offers syncthing 14.54 atm.

PR #5423, opened 2019-01-02
lib/model: Improve filesystem operations during tests (fixes #5422)

### LINKED ISSUE #5422: TestIssue3028 flaky on windows
It fails often on windows (never on linux and mac): https://build.syncthing.net/project.html?tab=testDetails&testNameId=-7613971120170926101&projectId=Syncthing&buildTypeId=Syncthing_BuildWindows

The reason seems to be leftover temporary dirs. This is just an example, exact output varies a lot:

```
09:20:44 INFO: No stored folder metadata for "default": recalculating
09:20:44 INFO: Ready to synchronize "default" (default) (sendreceive)
09:20:44 INFO: Scanner (folder default, file "3UYFCZEL"): scan: open \\?\C:\BuildAgent\work\src\github.com\syncthing\syncthing\lib\model\testdata\3UYFCZEL: Access is denied.
09:20:44 INFO: Scanner (folder default, file "62p3oC3P"): scan: open \\?\C:\BuildAgent\work\src\github.com\syncthing\syncthing\lib\model\testdata\62p3oC3P: Access is denied.
09:20:44 INFO: Scanner (folder default, file "7NU4PaYp"): scan: open \\?\C:\BuildAgent\work\src\github.com\syncthing\syncthing\lib\model\testdata\7NU4PaYp: Access is denied.
09:20:44 INFO: Scanner (folder default, file "96UXk9Uw"): scan: open \\?\C:\BuildAgent\work\src\github.com\syncthing\syncthing\lib\model\testdata\96UXk9Uw: Access is denied.
09:20:44 INFO: Scanner (folder default, file "H2HDpfkw"): scan: open \\?\C:\BuildAgent\work\src\github.com\syncthing\syncthing\lib\model\testdata\H2HDpfkw: Access is denied.
09:20:44 INFO: Scanner (folder default, file "hcZWkAhz"): scan: open \\?\C:\BuildAgent\work\src\github.com\syncthing\syncthing\lib\model\testdata\hcZWkAhz: Access is denied.
09:20:44 INFO: Scanner (folder default, file "icRdaFGV"): scan: open \\?\C:\BuildAgent\work\src\github.com\syncthing\syncthing\lib\model\testdata\icRdaFGV: Access is denied.
09:20:44 INFO: Scanner (folder default, file "r3PVuJYN"): scan: open \\?\C:\BuildAgent\work\src\github.com\syncthing\syncthing\lib\model\testdata\r3PVuJYN: Access is denied.
09:20:44 INFO: Scanner (folder default, file "TvHSHohQ"): scan: open \\?\C:\BuildAgent\work\src\github.com\syncthing\syncthing\lib\model\testdata\TvHSHohQ: Access is denied.
09:20:44 INFO: Scanner (folder default, file "xzUzuwNh"): scan: open \\?\C:\BuildAgent\work\src\github.com\syncthing\syncthing\lib\model\testdata\xzUzuwNh: Access is denied.
09:20:44 INFO: Completed initial scan of sendreceive folder "default" (default)
    model_test.go:2399: Incorrect local accounting; got 12 deleted files, expected 2
    model_test.go:2402: Incorrect global accounting; got 12 deleted files, expected 2
```

imsodin:
### Purpose

Introduces MustFilesystem that panics on errors and should be used for operatins
during testing which must never fail.  
Create temporary directories outside of testdata. It shouldn't matter when tests run sequentially, but if for some reason there is still leftovers, other tests will fail, so why not be save (and theoretically work concurrently).

### Testing

None yet and I can't be sure it really fixes #5422 (but I hope so), so lets see what CI has to say about it.


AudriusButkevicius:
This is a bit insane, if defer for some removal fails (because the test failed or was skipped) the test will blow up

imsodin:
> This is a bit insane, if defer for some removal fails (because the test failed or was skipped) the test will blow up

I realised that panicking is quite insane and there is no point of making this a filesystem. The idea was that it could be used in all tests regardless of package, but to achieve that a separate utility package would be more suitable than putting.

lib/model/testos_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"os"
+	"time"
+)
+
+// fatal is the required common interface between *testing.B and *testing.T
+type fatal interface {
+	Fatal(...interface{})
+}
+
+type fatalOs struct {
```
AudriusButkevicius: call it testOs that takes testing.T?

imsodin: It should also work with benchmarks, that's why I added the `fatal` interface. And I don't like naming the variable (`testOS`) the same way as the type (`fatalOs`). Does that make sense or would you use other names (or the same)?

PR #5429, opened 2019-01-04
lib/logger: Strip control characters from log output (fixes #5428)

### LINKED ISSUE #5428: Beeps randomly on Windows
Even when the console is hidden, it can emit beeps if a bell character is printed. We might print such if they are part of a filename.

calmh:
### Purpose

Don't randomly beep

### Testing

Bask in the glory of my unit test

### DIFF
--- a/lib/logger/logger.go
+++ b/lib/logger/logger.go
@@ -79,7 +79,7 @@ func New() Logger {
 	return newLogger(os.Stdout)
 }
 
-func newLogger(w io.Writer) Logger {
+func newLogger(w io.Writer) *logger {
 	res := &logger{
 		facilities: make(map[string]string),
 		debug:      make(map[string]struct{}),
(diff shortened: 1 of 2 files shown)


imsodin:
Isn't this going to be super slow when enabling any kind of debug logging? Or am I overestimating this (python-loops are bad might be resonating in my brain).

calmh:
Even our trace logs are few and far between by modern CPU standards I think... But I can bench it to be sure.

I wonder if there is some bug around the level stuff in Go <1.10, but we haven't changed this in ages and it used to work (but not be tested...)

calmh:
It is a bit slower, about 2x, slightly faster with the fast path (below benchmarks) and I still don't understand why it breaks.

```
BenchmarkLog-8             	  300000	      3830 ns/op	   0.26 MB/s	     624 B/op	       8 allocs/op
BenchmarkLogNoStripper-8   	  500000	      2861 ns/op	   0.35 MB/s	     352 B/op	       6 allocs/op
```

Skipping this for now as there is oddness and limited utility. I though it would be a trivial cost-free win.

PR #5433, opened 2019-01-05
vendor: Update minio/sha256-simd

AudriusButkevicius:
On my Ryzen 7.

Before:
```
[2ERMY] 00:06:04 INFO: Single thread SHA256 performance is 501 MB/s using minio/sha256-simd (444 MB/s using crypto/sha256).
```
After:
```
[2ERMY] 00:09:50 INFO: Single thread SHA256 performance is 1888 MB/s using minio/sha256-simd (444 MB/s using crypto/sha256).
```


go.sum
```diff
@@ -41,6 +41,8 @@ github.com/matttproud/golang_protobuf_extensions v1.0.1 h1:4hp9jkHxhMHkqkrB3Ix0j
 github.com/matttproud/golang_protobuf_extensions v1.0.1/go.mod h1:D8He9yQNgCq6Z5Ld7szi9bcBfOoFv/3dc6xSMkL2PC0=
 github.com/minio/sha256-simd v0.0.0-20171213220625-ad98a36ba0da h1:tazA5y1hWYJO8VSYbU36yBhXeIvruLXMUKu6WBtcJck=
 github.com/minio/sha256-simd v0.0.0-20171213220625-ad98a36ba0da/go.mod h1:2FMWW+8GMoPweT6+pI63m9YE3Lmw4J71hV56Chs1E/U=
+github.com/minio/sha256-simd v0.0.0-20190104231041-e529fa194128 h1:hEDK0Zao06IGlO1ada0FLT2g3KEot2vCqFp8gdvJqzM=
+github.com/minio/sha256-simd v0.0.0-20190104231041-e529fa194128/go.mod h1:2FMWW+8GMoPweT6+pI63m9YE3Lmw4J71hV56Chs1E/U=
```
imsodin: The entries above should have probably been removed.

calmh: Yeah... `go mod tidy` does this, it'll go away at some point. Initially `go.sum` was intended to be append-only, but that clearly doesn't scale to eternity.

calmh: Also I guess the commit prefix in the future should be something like `go.mod:` or `dependencies:`

PR #5440, opened 2019-01-09
all: Send Close BEP msg on intentional disconnect

imsodin:
### Purpose

This avoids waiting until next ping and timeout until the connection is actually
closed both by notifying the peer of the disconnect and by immediately closing
the local end of the connection after that. As a nice side effect, info level
logging about dropped connections now have the actual reason in it, not a generic
timeout error which looks like a real problem with the connection.

Inspired by the logs containing tons of misleading lines about connection failures posted in https://forum.syncthing.net/t/syncthing-reconnect-disconnect-cycles-frequent/12675

### Testing

Manual testing on test setup by un-/pausing folders.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -413,12 +413,12 @@ func (m *Model) RestartFolder(from, to config.FolderConfiguration) {
 		errMsg = "restarting"
 	}
 
-	m.tearDownFolderLocked(from, fmt.Errorf(errMsg+" folder %v", to.Description()))
+	m.tearDownFolderLocked(from, fmt.Errorf("%v folder %v", errMsg, to.Description()))
 	if !to.Paused {
 		m.addFolderLocked(to)
 		m.startFolderLocked(to.ID)
 	}
-	l.Infof(infoMsg+" folder %v (%v)", to.Description(), to.Type)
+	l.Infof("%v folder %v (%v)", infoMsg, to.Description(), to.Type)
 }
 
 func (m *Model) UsageReportingStats(version int, preview bool) map[string]interface{} {
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -8,11 +8,9 @@ package model
 
 import (
 	"bytes"
-	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"net"
 	"path/filepath"
 	"reflect"
@@ -128,6 +126,9 @@ var (
 	errFolderNotRunning  = errors.New("folder is not running")
 	errFolderMissing     = errors.New("no such folder")
 	errNetworkNotAllowed = errors.New("network not allowed")
+	// errors about why a connection is closed
+	errIgnoredFolderRemoved = errors.New("folder no longer ignored")
+	errReplacingConnection  = errors.New("replacing connection")
 )
 
 // NewModel creates and starts a new model. The model starts in read-only mode,
@@ -226,7 +227,7 @@ func (m *Model) startFolderLocked(folder string) config.FolderType {
 
 	// Close connections to affected devices
 	for _, id := range cfg.DeviceIDs() {
-		m.closeLocked(id)
+		m.closeLocked(id, fmt.Errorf("started folder %v", cfg.Description()))
 	}
 
 	v, ok := fs.Sequence(protocol.LocalDeviceID), true
@@ -339,7 +340,7 @@ func (m *Model) RemoveFolder(cfg config.FolderConfiguration) {
 	// Delete syncthing specific files
 	cfg.Filesystem().RemoveAll(config.DefaultMarkerName)
 
-	m.tearDownFolderLocked(cfg)
+	m.tearDownFolderLocked(cfg, fmt.Errorf("removing folder %v", cfg.Description()))
 	// Remove it from the database
 	db.DropFolder(m.db, cfg.ID)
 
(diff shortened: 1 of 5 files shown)


lib/model/model.go
```diff
@@ -398,14 +399,26 @@ func (m *Model) RestartFolder(from, to config.FolderConfiguration) {
...
+		errMsg = "pausing"
+	case from.Paused:
+		infoMsg = "Unpaused"
+		errMsg = "unpausing"
+	default:
+		infoMsg = "Restarted"
+		errMsg = "restarting"
+	}
+
+	m.tearDownFolderLocked(from, fmt.Errorf(errMsg+" folder %v", to.Description()))
```
calmh: Would be neater with `%s` than mixed `+` and format strings imho

lib/model/model.go
```diff
@@ -398,14 +399,26 @@ func (m *Model) RestartFolder(from, to config.FolderConfiguration) {
...
+	if !to.Paused {
 		m.addFolderLocked(to)
-		folderType := m.startFolderLocked(to.ID)
-		l.Infoln("Restarted folder", to.Description(), fmt.Sprintf("(%s)", folderType))
+		m.startFolderLocked(to.ID)
 	}
+	l.Infof(infoMsg+" folder %v (%v)", to.Description(), to.Type)
```
calmh: likewise

lib/protocol/protocol.go
```diff
@@ -801,22 +809,42 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
+		done := make(chan struct{})
+		c.sendClose <- asyncMessage{&Close{err.Error()}, done}
+		<-done
+
+		go c.commonClose(err)
+	})
+}
+
+// close is called if there is an unexpected error during normal operation.
 func (c *rawConnection) close(err error) {
```
calmh: I'm not super fond of the having both `close()` and `Close()`, the potential for confusion is enormous... I'm slightly confused even now. 😕 <- look, confused emoji

lib/protocol/protocol.go
```diff
@@ -801,22 +809,42 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
+// Close is called when the connection is regularely closed and thus the Close
+// BEP message is sent before terminating the actual connection. The error
+// argument specifies the reason for closing the connection.
+func (c *rawConnection) Close(err error) {
+	c.once.Do(func() {
+		done := make(chan struct{})
+		c.sendClose <- asyncMessage{&Close{err.Error()}, done}
+		<-done
+
+		go c.commonClose(err)
```
calmh: Some sort of comment on why we `go` this? I have a vague feeling it's because this might be called from a protocol callback and there are potential deadlock issues?

lib/model/model.go
imsodin: Not sure where my brain was at :\

lib/protocol/protocol.go
imsodin: Is `internalClose` suitable? It's a bit cryptic, but so is `errorClose` when `Close` takes an error as an argument as well and I couldn't think of anything better.

lib/protocol/protocol.go
imsodin: Clear like this?

calmh:
This could probably take care of #3379 somehow as well

lib/model/model.go
```diff
@@ -398,14 +399,26 @@ func (m *Model) RestartFolder(from, to config.FolderConfiguration) {
...
+	case to.Paused:
+		infoMsg = "Paused"
+		errMsg = "pausing"
+	case from.Paused:
+		infoMsg = "Unpaused"
+		errMsg = "unpausing"
+	default:
+		infoMsg = "Restarted"
+		errMsg = "restarting"
+	}
```
AudriusButkevicius: This fancy shortcutting switch behaviour takes a while to process, was pretty sure at one point this was an unfinished pr.

lib/model/model.go
```diff
@@ -128,6 +126,9 @@ var (
 	errFolderNotRunning  = errors.New("folder is not running")
 	errFolderMissing     = errors.New("no such folder")
 	errNetworkNotAllowed = errors.New("network not allowed")
+	// errors about why a connection is closed
+	errIgnoredFolderRemoved = errors.New("folder no longer ignored")
+	errReplacingConnection  = errors.New("replacing connection")
```
AudriusButkevicius: I thought (back in the day), the protocol had error codes to signify the reason in a cross implementation compatible way. Is that no longer the case?

lib/protocol/protocol.go
AudriusButkevicius: Why wouldn't we want an internal close here? Is it valid to close the connection multiple times? I assume that would panic on closing c.closed a second time. At which point the sync.once can move into the commonClose and commonClose can become internalClose?

lib/model/model.go
imsodin: Nope, human readable string according to the docsé

> Close
>
> [...]
>
> Fields
> 
> The reason field contains a human readable description of the error condition.
>

lib/protocol/protocol.go
imsodin: You lost me there. However you made me look at it again and I found a way this can deadlock and solving that indeed gets rid of `commonClose` (but introduces a second once). PR incoming

lib/model/model.go
AudriusButkevicius: Ah, it's response error code.

PR #5441, opened 2019-01-09
lib/db: Do all update operations on a single item at once

imsodin:
### Purpose

This is a successor to https://github.com/syncthing/syncthing/pull/5394 and has the same goal, but with less moving parts. By integrating `BlockMap` into `instance`/`keyer` all operations can now be done in `instance.updateLocalFiles` sequentially. Meaning the batch will only be flushed to db in a state, where all "update operations" for the contained items are done.

### Testing

None.

### DIFF
--- a/lib/db/blockmap.go
+++ b/lib/db/blockmap.go
@@ -11,125 +11,14 @@ import (
 	"fmt"
 
 	"github.com/syncthing/syncthing/lib/osutil"
-	"github.com/syncthing/syncthing/lib/protocol"
 
-	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/util"
 )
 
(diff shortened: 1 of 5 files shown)


AudriusButkevicius:
database stuff
image

imsodin:
As usual when doing something bigger, I made quite few not directly related changes. To make this easier to review I split it in two commits, the first being various polishing and optimization, the second being the actual change.

imsodin:
I think this still has merit, but I acknowledge that it is a bit of a monster - I ignored that due to hoping to fix #5340 with it. Now that this is out of the way, would you rather have me feed you this PR in easier to digest pieces or do you prefer to slay (and potential roast) the monster in one big effort? ;)

calmh:
Easy pieces if you don't mind

PR #5442, opened 2019-01-09
lib/protocol: Fix potential deadlock when closing connection (ref #5440)

imsodin:
### Purpose

While waiting for the Close BEP message to be sent, sending something else (that was randomly chosen in a switch over the Close) may fail and thus cause `internalClose` to be called, while the once is active -> deadlock.

This fixes this scenario by having separate onces for sending the close message and actually closing.

Should I revert https://github.com/syncthing/syncthing/commit/24ffd8be99bf21ade538fde04f1ff2bded06c53f and do this PR on top or is it fine as a followup like this?

### Testing

Don't see a way to trigger the described scenario reliably.

### DIFF
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -171,11 +171,12 @@ type rawConnection struct {
 	nextID    int32
 	nextIDMut sync.Mutex
 
-	outbox      chan asyncMessage
-	sendClose   chan asyncMessage
-	closed      chan struct{}
-	once        sync.Once
-	compression Compression
+	outbox        chan asyncMessage
+	closed        chan struct{}
+	closeOnce     sync.Once
+	sendCloseOnce sync.Once
+	writerExited  chan struct{}
+	compression   Compression
 }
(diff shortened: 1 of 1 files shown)


lib/protocol/protocol.go
```diff
@@ -813,41 +808,38 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
+		select {
+		case <-done:
+		case <-c.closed:
+		}
 	})
+
+	// No more sends are necessary, therefore closing the underlying
+	// connection can happen at the same time as the internal cleanup.
+	// And this prevents a potential deadlock due to calling c.receiver.Closed
+	go c.internalClose(err)
```
AudriusButkevicius: This comment is kind of funky now given there is only two closes rather than 3? Or am I just misunderstanding what it means.

imsodin: The comment should not depend on how many closes there is. Does this change make it clearer?

PR #5446, opened 2019-01-11
Multiarch for Docker

### LINKED ISSUE #5436: MultiArch for Docker
I'd like to feature-request Docker image releases of the same [binary release variety](https://github.com/syncthing/syncthing/releases). There will soon be a stable [docker manifest command](https://docs.docker.com/edge/engine/reference/commandline/manifest/) that allows the creation of image lists that allow automatic distribution of platform and architecture specific image variants. [All official docker images now use manifests](https://blog.docker.com/2017/09/docker-official-images-now-multi-platform/) and I think that would be a great addition to syncthing, which currently only serves the default linux/amd64 variant. I know that in #2591 linuxserver mentioned their release of such variants, but an official release is always more appealing.

Dargmuesli:
### Purpose

This refactors the Dockerfile into a multiarch.Dockerfile, that is provided along with build hook scripts to push a manifest from DockerHub to DockerHub. Fixes #5436.

### Testing

My fork builds successfully on DockerHub. See [cloud](https://cloud.docker.com/repository/registry-1.docker.io/dargmuesli/syncthing/builds/4f7112fe-0482-42d9-b367-b1c8414f7c13) (I'm not sure if someone else can) or [hub](https://hub.docker.com/r/dargmuesli/syncthing)

### Documentation

There are no documentation changes necessary as Docker manifests make it possible to pull the correct image automatically and thus remove the architecture (tag) selection step.

### DIFF
--- a/docker/hooks/build
+++ b/docker/hooks/build
@@ -12,24 +12,24 @@ AUTHTOKEN=$(cat ~/.dockercfg | jq '.["https://index.docker.io/v1/"] .auth')
 echo "{\"auths\":{\"https://index.docker.io/v1/\":{\"auth\": $AUTHTOKEN}},\"experimental\": \"enabled\"}" > ~/.docker/config.json
 ### -------------------- ###
 
-MULTIARCH_DOCKERFILE_PATH=$(echo "$DOCKERFILE_PATH" | sed "s/Dockerfile/multiarch.Dockerfile/")
+MULTIARCH_DOCKERFILE_PATH="${DOCKERFILE_PATH/Dockerfile/multiarch.Dockerfile}"
 
 # register qemu modules
 docker run --rm --privileged multiarch/qemu-user-static:register
 
 # remove MULTIARCH line and build default image
 sed "/__MULTIARCH_/d" "$MULTIARCH_DOCKERFILE_PATH" > "$DOCKERFILE_PATH"
-docker build --build-arg BUILD_GOOS="linux" --build-arg BUILD_GOARCH="amd64" --build-arg SERVE_U="amd64" --build-arg SERVE_R="alpine" --build-arg QEMUARCH="amd64" -f "$DOCKERFILE_PATH" -t $(echo $IMAGE_NAME | sed "s/:latest/:amd64-latest/") .
+docker build --build-arg BUILD_GOOS="linux" --build-arg BUILD_GOARCH="amd64" --build-arg SERVE_U="amd64" --build-arg SERVE_R="alpine" --build-arg QEMUARCH="amd64" -f "$DOCKERFILE_PATH" -t "${IMAGE_NAME/:latest/:amd64-latest}" .
 
 # enable MULTIARCH line and build multiarch images
 sed "s/__MULTIARCH_//g" "$MULTIARCH_DOCKERFILE_PATH" > "$DOCKERFILE_PATH"
 curl -L "https://github.com/multiarch/qemu-user-static/releases/download/v3.1.0-2/x86_64_qemu-i386-static.tar.gz" | tar xz
-docker build --build-arg BUILD_GOOS="linux" --build-arg BUILD_GOARCH="386" --build-arg SERVE_U="i386" --build-arg SERVE_R="alpine" --build-arg QEMUARCH="i386" -f "$DOCKERFILE_PATH" -t $(echo $IMAGE_NAME | sed "s/:latest/:i386-latest/") .
+docker build --build-arg BUILD_GOOS="linux" --build-arg BUILD_GOARCH="386" --build-arg SERVE_U="i386" --build-arg SERVE_R="alpine" --build-arg QEMUARCH="i386" -f "$DOCKERFILE_PATH" -t "${IMAGE_NAME/:latest/:i386-latest}" .
 curl -L "https://github.com/multiarch/qemu-user-static/releases/download/v3.1.0-2/x86_64_qemu-arm-static.tar.gz" | tar xz
-docker build --build-arg BUILD_GOOS="linux" --build-arg BUILD_GOARCH="arm" --build-arg SERVE_U="arm32v6" --build-arg SERVE_R="alpine" --build-arg QEMUARCH="arm" -f "$DOCKERFILE_PATH" -t $(echo $IMAGE_NAME | sed "s/:latest/:arm-latest/") .
+docker build --build-arg BUILD_GOOS="linux" --build-arg BUILD_GOARCH="arm" --build-arg SERVE_U="arm32v6" --build-arg SERVE_R="alpine" --build-arg QEMUARCH="arm" -f "$DOCKERFILE_PATH" -t "${IMAGE_NAME/:latest/:arm-latest}" .
 curl -L "https://github.com/multiarch/qemu-user-static/releases/download/v3.1.0-2/x86_64_qemu-aarch64-static.tar.gz" | tar xz
-docker build --build-arg BUILD_GOOS="linux" --build-arg BUILD_GOARCH="arm64" --build-arg SERVE_U="arm64v8" --build-arg SERVE_R="alpine" --build-arg QEMUARCH="aarch64" -f "$DOCKERFILE_PATH" -t $(echo $IMAGE_NAME | sed "s/:latest/:aarch64-latest/") .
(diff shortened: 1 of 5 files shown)


AudriusButkevicius:
Sorry, why is qemu involved? Not sure why it's needed?

Dargmuesli:
That's because DockerHub runs amd64 machines. No problem for Go, sure, but required for a build of & from an image of a different architecture.

I forgot to mention here that this is currently configured for the "latest" tag only, not for versioned tags. That's to do when this is confirmed to work fine for you too.

AudriusButkevicius:
Ok, not sure I understand how this works in this case.
Sure, build machines are running on amd64, so how does apt-get and all other stuff get executed on exotic arches if docker build runs on 

I don't see anything invoking qemu.
Also, what is __MULTIARCH_COPY  ?

Dargmuesli:
No problem. This is achieved by registering qemu via the docker run command that is executed before the builds. I can send you a link about this later.
Either the whole __MULTIARCH_COPY line is deleted (for amd64, that doesn't need qemu) or the COPY prefix is removed to enable the qemu enabling command.

Both is handled in the build hook.

docker/hooks/push
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# push individual tags
+docker push $(echo $IMAGE_NAME | sed "s/:latest/:i386-latest/")
```
calmh: As you're using bash all of these `echo|sed` tricks can be replaced by `${IMAGE_NAME/:latest/:i386-latest}` variable expansions

Dargmuesli: New, to me. Thanks for the hint, will try!

Dargmuesli:
Guys, we haven't even reached the meta state yet.

what-if-we-put-docker-inside-of-docker-dope-smoking-36186256

Regarding Qemu in the release: yep, I'll test it without. Good point.
Testing is more or less the build result on DockerHub.

Dargmuesli:
I've commited the requested changes, so you can set up your automated build on DockerHub with versioning and dev-tags as you wish. Keep in mind that the Dockerfile location is `docker/Dockerfile` now :)

docker/hooks/build
```diff
@@ -0,0 +1,35 @@
...
+
+# remove MULTIARCH line and build default image
+sed "/__MULTIARCH_/d" "$MULTIARCH_DOCKERFILE_PATH" > "$DOCKERFILE_PATH"
+docker build --build-arg BUILD_GOOS="linux" --build-arg BUILD_GOARCH="amd64" --build-arg SERVE_U="amd64" --build-arg SERVE_R="alpine" --build-arg QEMUARCH="amd64" -f "$DOCKERFILE_PATH" -t "${IMAGE_NAME/:latest/:amd64-latest}" .
+
+# enable MULTIARCH line and build multiarch images
+sed "s/__MULTIARCH_//g" "$MULTIARCH_DOCKERFILE_PATH" > "$DOCKERFILE_PATH"
+curl -L "https://github.com/multiarch/qemu-user-static/releases/download/v3.1.0-2/x86_64_qemu-i386-static.tar.gz" | tar xz
+docker build --build-arg BUILD_GOOS="linux" --build-arg BUILD_GOARCH="386" --build-arg SERVE_U="i386" --build-arg SERVE_R="alpine" --build-arg QEMUARCH="i386" -f "$DOCKERFILE_PATH" -t "${IMAGE_NAME/:latest/:i386-latest}" .
+curl -L "https://github.com/multiarch/qemu-user-static/releases/download/v3.1.0-2/x86_64_qemu-arm-static.tar.gz" | tar xz
```
calmh: For some sort of maintainability, I'd really like to see this as a loop over archs with the eqmu version broken out into a variable. As is this is a wall of magic, which we will need to get into to change stuff in the future...

docker/hooks/push
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# push individual tags
+docker push "${IMAGE_NAME/:latest/:i386-latest}"
```
calmh: This would need to support the actual tag being built, rather than latest, yeah?

docker/multiarch.Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+### Serve ###
+FROM $SERVE_U/$SERVE_R
+
+ARG QEMUARCH=amd64
+ENV PUID=1000 PGID=1000
+EXPOSE 8384 22000 21027/udp
+VOLUME ["/var/syncthing"]
+
+__MULTIARCH_COPY qemu-${QEMUARCH}-static /usr/bin/
+RUN apk add --update --no-cache ca-certificates su-exec
```
calmh: I read some articles and now understand (I think) that at this point we will have an `apk` binary for arm (or whatever) and use the qemu here via a binfmt_misc handler that was  registered earlier to run it. Also, to run the subsequent `rm`... This could really use a commit of some kind here, because it's totally non-obvious unless you're familiar with the background on multiarch docker images.

docker/hooks/build
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+# navigate from Dockerfile folder to project root
+cd ..
+
+### upgrade to Docker 18 ###
```
calmh: Do we want to predicate this hackishness on something we can detect from the environment? We're running on some opaque vm here that we have on control over or insight into and that might change at any point, right?

Dargmuesli:
Ya, this is currently pretty hacky, because it seems like the Docker guys & girls didn't have support for multiple architectures in mind when it was initially created.

docker/hooks/build
Dargmuesli: Yes, I thought about this too, but I weren't able to come up with a nice way to store the build argument combination for easy looping as I'm not incredibly experienced with bash. Could you do implement this?

docker/hooks/push
Dargmuesli: Correct. I don't know how your current DockerHub build setup looks like, so I left it vanilla.

docker/multiarch.Dockerfile
Dargmuesli: With "commit" do you mean that we should add a comment explaining it there?

docker/hooks/build
Dargmuesli: What exactly do you mean with "something we can detect from the environment"?

I have no idea why DockerHub runs Docker 17, maybe because of some LTS strategy...

I mean we could apply syncthing for admission to the [Library's official images](https://github.com/docker-library/official-images). Then syncthing would be built on dedicated machines without Qemu-hacks, but at the cost of quick releases. Afaik one needs to send a PR to update an official image, but I could be wrong about that.

docker/hooks/build
calmh: I admit I'm a bit damaged and treating bash as a real language only goes so far, but something like

```
archs=( 386:i386:i386 arm:arm32v6:arm ... )
for arch in "${archs[@]}" ; done
  parts=( ${arch//:/ } )
  goarch=${parts[0]}
  serve_u=${parts[1]}
  qemuarch=${parts[2]}
  docker build ...
done
```

calmh:
Yeah, to be super clear I'm not accusing you creating a horror, it seems like the horror is mostly by design. But it is a bit of a horror. I'd have expected something like a manifest file declaring a list of supported architectures, and docker hub doing whatever qemu magic it needs around the relevant docker build commands...

calmh:
Our docker hub build rules look like this today:

screen shot 2019-01-14 at 19 03 18

That is, tagged releases go into their own tag name and "latest", and "master" goes into "dev" for those wanting bleeding edge:

screen shot 2019-01-14 at 19 04 29

etc...

Dargmuesli:
Yup, I'll take care of that soon

Dargmuesli:
Where did you screenshot those build settings? On https://cloud.docker.com/repository/docker/syncthing/syncthing/builds/edit?

Dargmuesli:
Where did you screenshot those build settings? On https://cloud.docker.com/repository/docker/syncthing/syncthing/builds/edit?

calmh:
The URL for me is https://cloud.docker.com/repository/docker/syncthing/syncthing/hubbuilds but I guess approximately, yes?

Dargmuesli:
Funny, this URL doesn't exist for my project. That's maybe caused by Docker's recent merge of their web platforms.
I've implemented the requested changes. This should be usable now :)

Dargmuesli:
Neeevver mind, something went wrong with tagging. I'll look into it.

PR #5452, opened 2019-01-14
all: Update protobuf package 1.0.0 -> 1.2.0

calmh:
Also adds a few file global options to keep the generated code similar to what we already had.

### Purpose

Newness. I had a memory that there was some complication in this upgrade that I wanted to sort out, but it does not appear such now.

### Testing

Not much beyond the units, but the generated code doesn't appear to differ significantly apart from a bunch of new `XXX_...` methods that are apparently mandatory. We should be good.


AudriusButkevicius:
Xxx could he disabled before.

calmh:
There are some xxx fields that can be disabled (the options I added), the methods are new from upstream (golang/protobuf) and apparently not disableable.

PR #5457, opened 2019-01-14
cmd/syncthing, gui: Settings dialog upgrades/reporting for candidates (fixes #4216)

### LINKED ISSUE #4216: "Automatic upgrades" web GUI option in pre-release versions is misleading
Usually I stick to the release channel and upgrade manually. I wanted to take advantage of a recent bug fix in a pre-release version.

I set the Options to permit release candidates and after upgrading I wanted to set automatic upgrades back to "disabled". The web UI lets me do this, asked for a Syncthing restart, then the automatic upgrades are forced back on, as per pre-release behaviour. I assumed something had caused the config save to fail and tried to diagnose the "problem" :-)

It would be helpful if the Options->Automatic upgrades option showed "Stable and release candidates" and locked the menu to make it clear that the option cannot be changed, perhaps with a tooltip or a brief explanation below.

calmh:
### Purpose

This adds booleans to the /system/version response to advice the GUI whether the running version is a candidate release or not. (We could parse it from the version string, but why duplicate the logic.)

Additionally the settings dialog locks down the upgrade and usage reporting options on candidate releases. This matches the current behavior, it just makes it obvious what actually *can* be chosen.

The usage reporting settings locks down 1) when on a candidate version 2) when on a non-candidate version and selecting candidate upgrades. The latter isn't strictly necessary but it's what was already happening, and it also illustrates that the choice will be taken from you if you do upgrade to a candidate so I think this is fine.

### Testing

Manual

### Screenshots

On candidate:

screen shot 2019-01-14 at 18 09 05

screen shot 2019-01-14 at 18 09 36

On release:

screen shot 2019-01-14 at 18 10 24

### DIFF
--- a/gui/default/syncthing/settings/settingsModalView.html
+++ b/gui/default/syncthing/settings/settingsModalView.html
@@ -70,7 +70,7 @@
             <div class="col-md-6">
               <div class="form-group">
                 <label translate for="urVersion">Anonymous Usage Reporting</label> (<a href="" translate data-toggle="modal" data-target="#urPreview">Preview</a>)
-                <div ng-if="tmpOptions.upgrades != 'candidate'">
+                <div ng-if="tmpOptions.upgrades != 'candidate' && !version.isCandidate">
                   <select class="form-control" id="urVersion" ng-model="tmpOptions._urAcceptedStr">
                     <option ng-repeat="n in urVersions()" value="{{n}}">{{'Version' | translate}} {{n}}</option>
                     <!-- 1 does not exist, as we did not support incremental formats back then. -->
@@ -78,7 +78,7 @@
                     <option value="-1" translate>Disabled</option>
                   </select>
                 </div>
-                <p class="help-block" ng-if="tmpOptions.upgrades == 'candidate'">
+                <p class="help-block" ng-if="tmpOptions.upgrades == 'candidate' || version.isCandidate"">
                   <span translate>Usage reporting is always enabled for candidate releases.</span>
                 </p>
               </div>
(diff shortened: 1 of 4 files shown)


gui/default/syncthing/settings/settingsModalView.html
```diff
@@ -70,15 +70,15 @@
             <div class="col-md-6">
               <div class="form-group">
                 <label translate for="urVersion">Anonymous Usage Reporting</label> (<a href="" translate data-toggle="modal" data-target="#urPreview">Preview</a>)
-                <div ng-if="tmpOptions.upgrades != 'candidate'">
+                <div ng-if="tmpOptions.upgrades != 'candidate' && !version.isCandidate">
```
AudriusButkevicius: Whats the point of the upgrades check in this case?

gui/default/syncthing/settings/settingsModalView.html
```diff
@@ -70,15 +70,15 @@
...
                   <select class="form-control" id="urVersion" ng-model="tmpOptions._urAcceptedStr">
                     <option ng-repeat="n in urVersions()" value="{{n}}">{{'Version' | translate}} {{n}}</option>
                     <!-- 1 does not exist, as we did not support incremental formats back then. -->
                     <option value="0" translate>Undecided (will prompt)</option>
                     <option value="-1" translate>Disabled</option>
                   </select>
                 </div>
-                <p class="help-block" ng-if="tmpOptions.upgrades == 'candidate'">
+                <p class="help-block" ng-if="tmpOptions.upgrades == 'candidate' || version.isCandidate"">
```
AudriusButkevicius: Ditto

gui/default/syncthing/settings/settingsModalView.html
calmh: This shows the "usage reporting always on" variant (no choice, and corresponding help text) when upgrades to candidates is selected. That's not a change.

gui/default/syncthing/settings/settingsModalView.html
calmh: Ditto

PR #5462, opened 2019-01-18
lib/db: Fix iterating sequence index (fixes #5340)

### LINKED ISSUE #5340: Incorrectly out of sync, sequence number oddness.

calmh:
### Purpose

There was a problem in iterating the sequence index that could result
in missing updates. The issue is that while the index was (correctly)
iterated in a snapshot, the actual file infos were read dirty outside of
the snapshot. This fixes this by doing the reads inside the snapshot,
and also updates a couple of other places that did the same thing more
or less harmfully (I didn't investigate).

To avoid similar issues in the future I did some renaming of the
getFile* methods - the ones in a transaction are just getFile, while the
ones directly on the database are variants of getFileDirty to highlight
what's going on.

### Testing

There's a new unit test. Unfortunately it rather needs to run longer than is acceptable in the short test suite. Without the change, the test would fail after a few seconds. When run with `STTRACE=db` it would also trigger a `panic: sequence index corruption`.

### DIFF
--- a/lib/db/instance.go
+++ b/lib/db/instance.go
@@ -107,7 +107,7 @@ func (db *instance) withHave(folder, device, prefix []byte, truncate bool, fn It
 			prefix = append(prefix, '/')
 		}
 
-		if f, ok := db.getFileTrunc(db.keyer.GenerateDeviceFileKey(nil, folder, device, unslashedPrefix), true); ok && !fn(f) {
+		if f, ok := db.getFileDirtyTrunc(db.keyer.GenerateDeviceFileKey(nil, folder, device, unslashedPrefix), true); ok && !fn(f) {
 			return
 		}
 	}
@@ -124,11 +124,7 @@ func (db *instance) withHave(folder, device, prefix []byte, truncate bool, fn It
 			return
 		}
 
-		// The iterator function may keep a reference to the unmarshalled
-		// struct, which in turn references the buffer it was unmarshalled
-		// from. dbi.Value() just returns an internal slice that it reuses, so
-		// we need to copy it.
-		f, err := unmarshalTrunc(append([]byte{}, dbi.Value()...), truncate)
+		f, err := unmarshalTrunc(dbi.Value(), truncate)
 		if err != nil {
 			l.Debugln("unmarshal error:", err)
 			continue
@@ -147,7 +143,7 @@ func (db *instance) withHaveSequence(folder []byte, startSeq int64, fn Iterator)
 	defer dbi.Release()
 
 	for dbi.Next() {
-		f, ok := db.getFile(dbi.Value())
+		f, ok := t.getFileByKey(dbi.Value())
 		if !ok {
 			l.Debugln("missing file for sequence number", db.keyer.SequenceFromSequenceKey(dbi.Key()))
 			continue
@@ -209,14 +205,18 @@ func (db *instance) withAllFolderTruncated(folder []byte, fn func(device []byte,
 	}
 }
 
-func (db *instance) getFile(key []byte) (protocol.FileInfo, bool) {
-	if f, ok := db.getFileTrunc(key, false); ok {
+func (db *instance) getFileDirty(folder, device, file []byte) (protocol.FileInfo, bool) {
+	return db.getFileDirtyByKey(db.keyer.GenerateDeviceFileKey(nil, folder, device, file))
+}
+
+func (db *instance) getFileDirtyByKey(key []byte) (protocol.FileInfo, bool) {
+	if f, ok := db.getFileDirtyTrunc(key, false); ok {
 		return f.(protocol.FileInfo), true
 	}
 	return protocol.FileInfo{}, false
 }
 
-func (db *instance) getFileTrunc(key []byte, trunc bool) (FileIntf, bool) {
+func (db *instance) getFileDirtyTrunc(key []byte, trunc bool) (FileIntf, bool) {
 	bs, err := db.Get(key, nil)
 	if err == leveldb.ErrNotFound {
 		return nil, false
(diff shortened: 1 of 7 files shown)


lib/db/instance.go
```diff
@@ -124,11 +124,7 @@ func (db *instance) withHave(folder, device, prefix []byte, truncate bool, fn It
 			return
 		}
 
-		// The iterator function may keep a reference to the unmarshalled
-		// struct, which in turn references the buffer it was unmarshalled
-		// from. dbi.Value() just returns an internal slice that it reuses, so
-		// we need to copy it.
-		f, err := unmarshalTrunc(append([]byte{}, dbi.Value()...), truncate)
```
calmh: This talk was about the old xdr implementation. The protobuf unmarshaller keeps no references.

lib/db/instance.go
```diff
@@ -297,7 +297,7 @@ func (db *instance) withGlobal(folder, prefix []byte, truncate bool, fn Iterator
 
 		fk = db.keyer.GenerateDeviceFileKey(fk, folder, vl.Versions[0].Device, name)
 
-		f, ok := db.getFileTrunc(fk, truncate)
+		f, ok := t.getFileTrunc(fk, truncate)
```
AudriusButkevicius: From consistency perspective, shouldn't the root check at the top of this function happen part of the transaction?

calmh: I considered it, but since it just reads one value it's equivalent anyway and taking the snapshot looked like overhead. However I see now that `getGlobal` actually takes a snapshot of its own, so it would be cleaner to do that... I'll tweak.

imsodin: Same for `withHave`. As far as I understood, taking the snapshot is no or negligible overhead, because `db.Get` internally takes a snapshot too. So first doing `db.Get` then in most cases starting a transaction and thus taking a seconds snapshot seems like more overhead.

lib/db/instance.go
```diff
@@ -209,14 +205,18 @@ func (db *instance) withAllFolderTruncated(folder []byte, fn func(device []byte,
...
+
+func (db *instance) getFileDirtyByKey(key []byte) (protocol.FileInfo, bool) {
+	if f, ok := db.getFileDirtyTrunc(key, false); ok {
 		return f.(protocol.FileInfo), true
 	}
 	return protocol.FileInfo{}, false
 }
 
-func (db *instance) getFileTrunc(key []byte, trunc bool) (FileIntf, bool) {
+func (db *instance) getFileDirtyTrunc(key []byte, trunc bool) (FileIntf, bool) {
```
imsodin: There are now only calls to `db.getFileDirty(...` left, so this function and `getFileDirtyByKey` could be removed.

calmh:
Seeing as we might need a meme then

calmh:
I was considering doing just that. On the one hand we've had this bug since 0.14.48, on the other hand it's an annoying, confidence-reducing issue that we should get out the fix for sooner rather than later, and it's >2w left of the RC time...

imsodin:
Dammit, when rebasing #5441 I think I just discovered the exact same problem you solved here but with `getGlobal` instead. Let me do a minimal PR with just that fixed, I am not yet sure whether or not it requires another RC

PR #5464, opened 2019-01-18
lib/db: Prevent some allocations by using DeviceID instead of []bytes

imsodin:
First of a few PRs doing some kind of polish/optimization/refactor on the database.  
This is micro-polish/-optimization based on the fact that protocol.DeviceIDFromBytes` needs to allocate a `[]byte`, while getting `[]byte` from `protocol.DeviceID` is cheap (and short `[:]`).

### DIFF
--- a/lib/db/db_test.go
+++ b/lib/db/db_test.go
@@ -185,7 +185,7 @@ func TestUpdate0to3(t *testing.T) {
 		haveUpdate0to3[remoteDevice1][0].Name: haveUpdate0to3[remoteDevice1][0],
 		haveUpdate0to3[remoteDevice0][2].Name: haveUpdate0to3[remoteDevice0][2],
 	}
-	db.withNeed(folder, protocol.LocalDeviceID[:], false, func(fi FileIntf) bool {
+	db.withNeed(folder, protocol.LocalDeviceID, false, func(fi FileIntf) bool {
 		e, ok := need[fi.FileName()]
 		if !ok {
 			t.Error("Got unexpected needed file:", fi.FileName())
--- a/lib/db/db_test.go
+++ b/lib/db/db_test.go
@@ -185,7 +185,7 @@ func TestUpdate0to3(t *testing.T) {
 		haveUpdate0to3[remoteDevice1][0].Name: haveUpdate0to3[remoteDevice1][0],
 		haveUpdate0to3[remoteDevice0][2].Name: haveUpdate0to3[remoteDevice0][2],
 	}
-	db.withNeed(folder, protocol.LocalDeviceID, false, func(fi FileIntf) bool {
+	db.withNeed(folder, protocol.LocalDeviceID[:], false, func(fi FileIntf) bool {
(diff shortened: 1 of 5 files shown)


calmh:
I'm not convinced. Once you have a `[]byte` it's cheap to pass around, a `DeviceID` is a fair bit larger. The current "boundary" where `Set` talks in real types like `DeviceID` and the database layer talks about byte slices is fairly clear, and this muddies it.

Looking at this though, I wouldn't be surprised if the deep level `l.Debugln`s are quite expensive, executed thousands of times, and too voluminous to ever be turned on. The `DeviceIDFromBytes` there will run regardless of whether debugging is enabled. Maybe we should remove those.

Of course, a benchmark could prove this is a good/bad/irrelevant change. :)

imsodin:
> Once you have a `[]byte` it's cheap to pass around, a `DeviceID` is a fair bit larger.

How so? `DeviceID` is just a `[DeviceIDLength]byte`, i.e. of the same size.

calmh:
You're right, they are about the same size

imsodin:
Argh, I spent some time benchmarking just to make a really basic realizations: Arrays are passed by value, i.e. cause allocation -> you were still right: passing slices is more efficient. That's why this PR doesn't change much in the benchmarks, but commenting out all `l.Debug` lines does. `DeviceID.String` isn't particularly cheap either, so passing around `DeviceID`s is definitely not the way to go. I for one like those debug logs, they may be a flood but they also are the only real source of information for anything db related and helped me several times. I see two ways to avoid all those allocations when not debugging:

 - Hide the expensive `l.Debug` calls behind `shouldDebug`.
 - At the set - instance boundary convert `DeviceID` to `device []byte` and `devStr string` and pass around both.

Opinions?

imsodin:
I'll stop trying - I can't find any change in benchmarks, even with a simple change like this which in my opinion should just prevent one allocation per loop execution:

```diff
--- a/lib/db/instance.go
+++ b/lib/db/instance.go
@@ -325,6 +325,7 @@ func (db *instance) withNeed(folder, device []byte, truncate bool, fn Iterator)
 
        var fk []byte
        devStr := protocol.DeviceIDFromBytes(device).String()
+       var needDevice protocol.DeviceID
        for dbi.Next() {
                vl, ok := unmarshalVersionList(dbi.Value())
                if !ok {
@@ -341,7 +342,7 @@ func (db *instance) withNeed(folder, device []byte, truncate bool, fn Iterator)
 
                name := db.keyer.NameFromGlobalVersionKey(dbi.Key())
                needVersion := vl.Versions[0].Version
-               needDevice := protocol.DeviceIDFromBytes(vl.Versions[0].Device)
+               copy(needDevice[:], vl.Versions[0].Device)
 
                for i := range vl.Versions {
                        if !vl.Versions[i].Version.Equal(needVersion) {

```

Either the go compiler is super smart at optimization or I just don't get it (I'd wager on the latter).

PR #5465, opened 2019-01-18
lib/db: Deduplicate comparing old and new items

imsodin:
Add a utility function `unchanged` that determines whether two `FileIntf` are the same from a database perspective (i.e. whether or not the update needs to happen).

### DIFF
--- a/lib/db/instance.go
+++ b/lib/db/instance.go
@@ -48,7 +48,7 @@ func (db *instance) updateFiles(folder, device []byte, fs []protocol.FileInfo, m
 			err = ef.Unmarshal(bs)
 		}
 
-		if unchanged(f, ef, err == nil) {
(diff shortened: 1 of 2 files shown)


lib/db/instance.go
```diff
@@ -48,9 +48,7 @@ func (db *instance) updateFiles(folder, device []byte, fs []protocol.FileInfo, m
 			err = ef.Unmarshal(bs)
 		}
 
-		// Local flags or the invalid bit might change without the version
-		// being bumped. The IsInvalid() method handles both.
-		if err == nil && ef.Version.Equal(f.Version) && ef.IsInvalid() == f.IsInvalid() {
+		if unchanged(f, ef, err == nil) {
```
calmh: Passing the bool is weird. `err == nil && dbEquivalent(f, ef)` scans better imho

PR #5468, opened 2019-01-19
lib/db: Deduplicate getFile* code

imsodin:


imsodin:
> Also, I might even remove get stuff from db instance and always force all code to start a transaction.

I think of transitions as an internal thing of `instance`, meaning `FileSet` code shouldn't touch it (which is also the only place the db get function is used outside of tests).

AudriusButkevicius:
Defer screws up inlining optimisations, so any function that has a defer in it cannot be inlined.

calmh:
I think any lack of inlining is dwarfed by the fact that these things often become I/O operations or, if not that, anyway contain decompression, checksumming, unmarshalling and stuff.

PR #5470, opened 2019-01-19
lib/db: Consistent use of buffers

imsodin:
Second to last PR :)

In some places we do use/share byte buffers for keys, in some we don't. In this PR the use is made consistent such that buffers are used for all "normal" operations. Normal in the sense that they aren't part of some kind of "special" operation, but happen often (every loop of update, drop, ...).

### DIFF
--- a/lib/db/instance.go
+++ b/lib/db/instance.go
@@ -53,7 +53,7 @@ func (db *instance) updateFiles(folder, device []byte, fs []protocol.FileInfo, m
 		t.insertFile(dk, folder, device, f)
 
 		gk = db.keyer.GenerateGlobalVersionKey(gk, folder, name)
-		t.updateGlobal(gk, keyBuf, folder, device, f, meta)
+		keyBuf, _ = t.updateGlobal(gk, keyBuf, folder, device, f, meta)
 
 		// Write out and reuse the batch every few records, to avoid the batch
 		// growing too large and thus allocating unnecessarily much memory.
--- a/lib/db/instance.go
+++ b/lib/db/instance.go
@@ -34,7 +34,7 @@ func (db *instance) updateFiles(folder, device []byte, fs []protocol.FileInfo, m
 	t := db.newReadWriteTransaction()
 	defer t.close()
 
-	var dk, gk []byte
+	var dk, gk, keyBuf []byte
 	devID := protocol.DeviceIDFromBytes(device)
 	for _, f := range fs {
 		name := []byte(f.Name)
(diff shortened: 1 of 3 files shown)


lib/db/instance.go
```diff
@@ -53,7 +53,7 @@ func (db *instance) updateFiles(folder, device []byte, fs []protocol.FileInfo, m
 		t.insertFile(dk, folder, device, f)
 
 		gk = db.keyer.GenerateGlobalVersionKey(gk, folder, name)
-		t.updateGlobal(gk, folder, device, f, meta)
+		t.updateGlobal(gk, keyBuf, folder, device, f, meta)
```
calmh: `buf, _ =  ...`

lib/db/instance.go
```diff
@@ -34,7 +34,7 @@ func (db *instance) updateFiles(folder, device []byte, fs []protocol.FileInfo, m
 	t := db.newReadWriteTransaction()
 	defer t.close()
 
-	var dk, gk []byte
+	var dk, gk, keyBuf []byte
```
imsodin: The same thing could be achieved with just two buffers and the naming isn't as succinct as possible (keyBuf vs just buf), but this choice will become more clear in the followup PR (with blockBuf). Having three buffers also is a bit safer, as then `dk` really only is used as device-file-key, not for other purposes too.

lib/db/transactions.go
```diff
@@ -65,25 +64,25 @@ func (t readOnlyTransaction) getFileTrunc(key []byte, trunc bool) (FileIntf, boo
 	return f, true
 }
 
-func (t readOnlyTransaction) getGlobalInto(gk, dk, folder, file []byte, truncate bool) ([]byte, []byte, FileIntf, bool) {
-	gk = t.db.keyer.GenerateGlobalVersionKey(gk, folder, file)
+func (t readOnlyTransaction) getGlobal(buf, folder, file []byte, truncate bool) ([]byte, FileIntf, bool) {
```
calmh: So `buf` in the calling function is fine, but calling it `buf` in here makes no sense. Maybe as the parameter name it's ok, because it's a random buffer of sorts and not actually the key "on the way in" - but then the assign on the next line should be `key :=` and that's what we use in the rest of the function and return. Other things like `t.Get(buf, nil)` looks like we're getting stuff into the buf or whatever.

lib/db/instance.go
calmh: I think having several is fine. Less risk that we reuse it half way in some function call and then expect it to be unchanged on the next line.

lib/db/transactions.go
imsodin: Actually, what about calling the parameter `key` instead of `buf`? That is unspecific (not gk, dk, nk, ...) and doesn't have the potential for the "fill-this-buffer" confusion.

calmh: I think `keyBuf` is the better one then. It indicates a buffer that will hold a key. `key` looks like the caller should pass some kind of actual key there.

PR #5471, opened 2019-01-20
lib/db: Various polish

imsodin:
This is mostly stuff I forgot in the recent PRs or for the final reduction of #5441's diff size.

naming: buf -> keyBuf
dedup: use getFileTrunc
manually inline insertFile

### DIFF
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -163,12 +163,12 @@ func (s *FileSet) Update(device protocol.DeviceID, fs []protocol.FileInfo) {
 	fs = fs[:0]
 	folder := []byte(s.folder)
 	for _, nf := range oldFs {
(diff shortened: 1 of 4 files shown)


lib/db/set.go
```diff
@@ -163,12 +163,12 @@ func (s *FileSet) Update(device protocol.DeviceID, fs []protocol.FileInfo) {
 	fs = fs[:0]
 	folder := []byte(s.folder)
 	for _, nf := range oldFs {
-		ef, ok := s.db.getFileDirty(folder, device[:], []byte(osutil.NormalizedFilename(nf.Name)))
+		ef, ok := s.db.getFileDirty(folder, device[:], []byte(nf.Name))
```
imsodin: Normalization already happens earlier in `Update` via `normalizeFilenames(fs)`.

PR #5475, opened 2019-01-22
build: Further DockerHub Script Optimization (#5459)

Dargmuesli:
~ Outsources configuration data to .csv
~ Improves comments
~ Checks Docker version before upgrade
~ Makes array variables readable
~ Creates a manifest with the correct architecture for individual images
~ Fixes the multiarch Dockerfile

### DIFF
--- a/docker/hooks/archs.csv
+++ b/docker/hooks/archs.csv
@@ -1,0 +1,7 @@
+BUILD_GOOS:BUILD_GOARCH:SERVE_U:SERVE_R:QEMUARCH:VARIANT
+linux:386:i386:alpine:i386:
+linux:amd64:amd64:alpine:amd64:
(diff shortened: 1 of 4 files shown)


AudriusButkevicius:
Please don't self-merge PRs, everything that goes into thia repo needs reviews. 

For all that I know you might be adding a backdoor to the containers and not wanting anyone to see it.

AudriusButkevicius:
Nevermind, this is a PR into a branch, which is even more weird, but harmless I guess.

Dargmuesli:
Yep, @calmh added me as collaborator when I requested creation of the docker branch on which you were supposed to be able to do your own DockerHub config testing, but I had no push permissions on that branch. Sorry though, I thought one PR would be enough and then discovered problems that made me self-merge another one.

PR #5478, opened 2019-01-23
vendor: rm -rf vendor

calmh:
This removes our vendored dependencies. They provide no value for our own build or development processes and are a bit of a drag to carry around. For our source releases, the build job can accomplish the same thing by a "go mod vendor" to recreate the vendor dir (from the cryptographically verified dependencies).

Other "downstreams" could then either use those source releases, or use module builds.


calmh:
I cheated and added a "go mod vendor" (using modern Go) just ahead of the build-with-old-Go build step. But I'm not sure if it's valuable...

PR #5479, opened 2019-01-24
all: Copy owner/group from parent (fixes #5445)

### LINKED ISSUE #5445: User/group ownership following parent dir
Syncing ownership is an oft requested feature (#1329). It's also complex to do it right, which is why that issue languishes without action.

One common use case is to synchronize user home directories, or similar structures where a single *folder* might contain multiple directory hierarchies of files, each with a specific user/group owner. A simple way to handle such a case is to enable copying of user/group from the parent directory. That is, if we are creating `foo/bar/baz`, at the final stage where we set the permissions etc we look up the ownership of `foo/bar` and apply those to `foo/bar/baz`. This means that assuming `/home/jb/` is already in place and we are syncing `/home/`, anything under `/home/jb` will get the right ownership.

This scheme becomes just a checkbox similar to "ignore permissions"; no protocol changes are required. There is no need to have the same uid/gid across systems. The setting is device local only.

Creating a new user in this scheme requires doing it on both sides, to get the ownership info propagated. I'm OK with that. If this is missed, one can simply do a `chown -R` on the destination side, and the ownership will be preserved from that point onwards.

The situation on Windows might be more complex, but on the other hand the whole ownership issue is less important to begin with as ACLs are inherited and contain much of the same information.

On linux the process will need `CAP_FOWNER` (I think) which is something we should document that the user needs to sort out.

calmh:
### Purpose

This adds a folder option "CopyParentOwner" which, when set, makes
Syncthing attempt to retain the owner/group information when syncing
files. Specifically, at the finisher stage we look at the parent dir to
get owner/group and then attempt a Lchown call on the temp file. For
this to succeed Syncthing must be running with the appropriate
permissions. On Linux this is CAP_FOWNER, which can be granted by the
service manager on startup or set on the binary in the filesystem. Other
operating systems do other things, but often it's not required to run as
full "root". On Windows this patch does nothing - ownership works
differently there and is generally less of a deal, as permissions are
inherited as ACLs anywah.

### Testing

There are unit tests on the Lchown functionality, which requires the
above permissions to run. There is also a unit test on the folder which
uses the fake filesystem and hence does not need special permissions to
run.

### Documentation

This really needs a doc article on the config. I'll write one.


lib/fs/fileinfo_unix.go
```diff
@@ -0,0 +1,25 @@
...
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// +build !windows
+
+package fs
+
+import "syscall"
```
imsodin: It hardly matters here, but shouldn't syscall generally be avoided in favor of golang.org/x/sys/... packages?

lib/model/folder_sendrecv.go
```diff
@@ -17,6 +16,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
```
imsodin: Error context traces - nice concept :)

There's quite a few places where it could be used as well, where we currently use a something custom like `fmt.Errorf("%s: %s", context, err)` - but that's unrelated to this PR (maybe it's the spark for it's usage elsewhere :) ).

lib/fs/fileinfo_unix.go
calmh: Not in this case, because we need to match on what is actually returned. :)

lib/model/folder_sendrecv.go
calmh: It is, and I've wanted to introduce it. I took the step to do so here because without it we would just get a `Lstat: foo: permission denied` or something which would be a lot of tricky debugging to figure out what the fuck is going on, while seeing the `copy owner from parent: Lstat: foo: ...` is a much better place to start from when figuring out why the stat (or whatever) fails.

calmh:
I'd forgotten that symlinks even has owners... Will dig into it

calmh:
I added handling for symlinks and enough support in fakefs to be able to test it. Fakefs now "supports" symlinks to the extent that they can be created and inspected, but it doesn't support following them in Open() or when traversing a directory etc but that was Hard 🤷‍♂️ and I couldn't be bothered.

calmh:
CopyOwnershipFromParent SGTM

lib/fs/fakefs.go
```diff
@@ -126,10 +126,19 @@ func newFakeFilesystem(root string) *fakefs {
...
+	fakeEntryTypeFile fakeEntryType = iota
+	fakeEntryTypeDir
+	fakeEntryTypeSymlink
+)
+
 // fakeEntry is an entry (file or directory) in the fake filesystem
 type fakeEntry struct {
 	name     string
-	isdir    bool
+	typ      fakeEntryType
```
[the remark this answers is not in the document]
calmh: as this is only used for the fakery, I think it's valid to retain the fake- prefix?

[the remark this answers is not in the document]
AudriusButkevicius: I am referring to typ member rather than the type name

PR #5487, opened 2019-01-28
all: Use new reflect based CLI

AudriusButkevicius:
This now uses my black magic package that does reflection based CLI construction which hopefully will stay up to date for some things (and others can be added easily).

Supports reading stuff from stdin which allows to batch commands in a single post.

Next step would be to do tag based field validation so stuff can be validated before posting and not relying on syncthing itself to tell us if stuff is ok.

Can be easily extended to modify file instead of rest endpoint I guess.

Does two other things:
Moves location stuff from cmd/syncthing into lib/location
Moves metadata (build et all) stuff from cmd/syncthing into lib/meta


AudriusButkevicius:
Actually, moving the meta stuff seems not to work :(

calmh:
If you move the variables out of the main package you need to adjust the -X parameters to the linker to have the full package path, something like `-X github.com/syncthing/syncthing/lib/meta.Version=...`.

AudriusButkevicius:
Well, to some definition of revives. It will be harder touse but will not go out of date.

lib/locations/locations.go
```diff
@@ -0,0 +1,157 @@
...
+	return nil
+}
+
+// defaultConfigDir returns the default configuration directory, as figured
+// out by various the environment variables present on each platform, or dies
+// trying.
+func defaultConfigDir() string {
+	switch runtime.GOOS {
+	case "windows":
+		if p := os.Getenv("LocationalAppData"); p != "" {
```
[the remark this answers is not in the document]
calmh: Wait, why are these things changing? I was assuming this was a pure move/rename thing, not with changes in it... Ah, this is a botched search/replace maybe?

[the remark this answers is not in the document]
AudriusButkevicius: Indeed.

cmd/syncthing/gui.go
```diff
@@ -1092,16 +1094,16 @@ func (s *apiService) getSupportBundle(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Archived log (default on Windows)
-	if logFile, err := ioutil.ReadFile(locations[locLogFile]); err == nil {
+	if logFile, err := ioutil.ReadFile(locations.GetLocation(locations.LogFileLocation)); err == nil {
```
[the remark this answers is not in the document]
AudriusButkevicius: Agree

AudriusButkevicius:
So most of the black magic comes from https://github.com/AudriusButkevicius/recli which has a explanation in the readme what it does.

Take it (stcli) for a spin and it will all make sense, all config mutations are done via 'stcli modify ...'  command group.

imsodin:
I did a very limited spin:

```
> ./bin/stcli modify version get
28
> ./bin/stcli modify version set 42
> ./bin/stcli modify version get
42
```

Seems to work, but what is it? And worse, I probably just changed something in my production config o.O

```
> grep -i version ~/.config/syncthing/config.xml
<configuration version="42">
[...]
```

Apparently I did :D  
Luckily the tool also provides me with the means to revert the effects of my curiosity easily. And if I had actually read to the end of the output of `./bin/stcli` I would have known it.

Looks very impressive!

Two things I noticed:

 - I'd rename the command `modify` to `config`. It really is about the config and querying or changing it.

 - It's a bit confusing that what you regularly know as commands (`list`, `add`) is mixed with the contents of a structure, e.g. folder names. I'd expect a command like `select` that takes one argument corresponding to a value of the output of `list` instead of making every folder name a command. Or somehow separate the "normal commands" from "content commands".

AudriusButkevicius:
For collections, I agree that adding a second level of indirection is probably the right way.

I still want to have a `dump` to dump the whole config that happens to coincide with other config keys, otherwies it becomes:

```
stcli config get options get something dump
```

rather than
```
stcli config options something dump
```

lib/locations/locations.go
```diff
@@ -0,0 +1,157 @@
+// Copyright (C) 2019 The Syncthing Authors.
```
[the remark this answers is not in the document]
AudriusButkevicius: Not really, exposes enums, setters etc, which did not exist before.

AudriusButkevicius:
I think this is ready for a proper shake.

cmd/stcli/main.go
```diff
@@ -1,63 +1,190 @@
...
 		},
-		cli.BoolFlag{
-			Name:   "insecure, i",
-			Usage:  "Do not verify SSL certificate",
-			EnvVar: "STINSECURE",
+	}
+
+	// Do not print usage of these flags, and ignore errors as this can't understand plenty of things
+	flags.Usage = func() {}
+	_ = flags.Parse(os.Args)
```
imsodin: `os.Args[1:]` (currently reads info from config when giving an invalid argument to `-gui-address`), apparently command name must not be called to `Parse`.

cmd/stcli/main.go
```diff
@@ -1,63 +1,190 @@
...
+		guiCfg = cfg.GUI()
+	}
+
+	if guiCfg.Address() == "" {
+		log.Fatalln("Could not find GUI Address")
+	}
+
+	if guiCfg.APIKey == "" {
+		log.Fatalln("Could not find GUI API key")
+	}
```
imsodin: The error messages should reflect that you need to specify both `-gui-address` and `-api-key` or none - that's not obvious.

AudriusButkevicius: Added an extra message at the top.
If something is missing in the config, the more specific error is more useful.

AudriusButkevicius:
This is now awesome or what

calmh:
Exploring the CLI blindly (and it is a bit awesome) sometimes runs into that full `reflect` experience:

```
jb@kvar:~/d/g/s/syncthing $ ./bin/stcli config devices add
panic: reflect: call of reflect.Value.Elem on slice Value

goroutine 1 [running]:
reflect.Value.Elem(0x449a5e0, 0xc0003ca4e0, 0x197, 0x0, 0x0, 0x0)
	/usr/local/go/src/reflect/value.go:805 +0x1bd
```

AudriusButkevicius:
fixes in a few mins

lib/config/deviceconfiguration.go
```diff
@@ -36,6 +37,7 @@ func NewDeviceConfiguration(id protocol.DeviceID, name string) DeviceConfigurati
 		DeviceID: id,
 		Name:     name,
 	}
+	util.SetDefaults(&d)
```
golangcibot: Error return value of `util.SetDefaults` is not checked (from `errcheck`)

lib/config/folderconfiguration.go
```diff
@@ -69,18 +69,13 @@ type FolderDeviceConfiguration struct {
...
-		MaxConflicts:     -1,
-		FilesystemType:   fsType,
-		Path:             path,
+		ID:             id,
+		Label:          label,
+		Devices:        []FolderDeviceConfiguration{{DeviceID: myID}},
+		FilesystemType: fsType,
+		Path:           path,
 	}
+	util.SetDefaults(&f)
```
golangcibot: Error return value of `util.SetDefaults` is not checked (from `errcheck`)

lib/config/size_test.go
```diff
@@ -6,7 +6,26 @@
...
+	"github.com/syncthing/syncthing/lib/util"
+)
+
+type TestStruct struct {
+	Size Size `default:"10%"`
+}
+
+func TestSizeDefaults(t *testing.T) {
+	x := &TestStruct{}
+	util.SetDefaults(x)
```
golangcibot: Error return value of `util.SetDefaults` is not checked (from `errcheck`)

lib/locations/locations.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/syncthing/syncthing/lib/fs"
+)
+
+type LocationEnum string
+
+// Use strings as keys to make printout and serialization of the locations map
+// more meaningful.
+const (
+	ConfigFile    LocationEnum = "config"
```
golangcibot: only the first constant in this group has an explicit type (from `staticcheck`)

lib/locations/locations.go
```diff
@@ -0,0 +1,157 @@
...
+	PanicLog                   = "panicLog"
+	AuditLog                   = "auditLog"
+	GUIAssets                  = "GUIAssets"
+	DefFolder                  = "defFolder"
+)
+
+type BaseDirEnum string
+
+const (
+	ConfigBaseDir BaseDirEnum = "config"
```
golangcibot: only the first constant in this group has an explicit type (from `staticcheck`)

AudriusButkevicius:
god this bot is terrible.

AudriusButkevicius:
Do you want me to address all of those? Including propagating the utils.SetDefault errors, or should those just be panics?

calmh:
I don't think we've ever checked the return of SetDefaults, so it should probably not even have an error return.

calmh:
That's the opposite of what I meant. :) Actually looking at the errors in SetDefault, it can only error when it can't parse the default given in the struct tag. That's clear panic territory, and it even already has a panic in it. So SetDefaults() shouldn't return an error and your original code is fine, it's SetDefault that should adjust.

AudriusButkevicius:
totally awesome now

AudriusButkevicius:
image

calmh:
I don't suggest doing anything like it in *this* PR, but *if* there was a nice and working cli I'd suggest linking it into the main binary.

```
$ syncthing serve       # starts the daemon, like today
$ syncthing             # alias for "syncthing serve", does the same thing as today
$ synching folder list  # triggers the stcli stuff
... etc
```

AudriusButkevicius:
I think let's get the code in, and worry about the shuffle later. 

I guess there is no risk in the cli package changes as that was a dumpster fire anyway, hence the review should focus on breaking existing stuff.

And yes. Good spot with the var expansion, that needs to be fixed.

AudriusButkevicius:
image

imsodin:
Just one minor thing about locations: Setting home with a trailing slash (which is the default expansion of bash) like this:

`./bin/stcli --home config-testing/ show version`

Results in the config file path `config-testing//config.xml`. The input path needs to be sanitized (e.g. `filepath.Clean`), either in stcli, or in lib/locations on `SetBaseDir` or in `expandLocations`.

Apart from that it looks good, but apart from main I hardly reviewed the stcli stuff at all.

AudriusButkevicius:
why does the path matter?

imsodin:
It doesn't, but as I found it by accident already now, it's easy to fix and I'd rather not have to think about whether this double slash is relevant when I debug something in the future, it would be nice if it were fixed right away.

calmh:
Imma accept this but oh boy should it ever have been a series of smaller PRs...

AudriusButkevicius:
Ye, don't dispute that. Sadly the other PR changes came after the main changes.

calmh:
New CLI is actually not bad ❤️

AudriusButkevicius:
We should open a ticket for whatever rest you envision being embedded somewhere.

Issue #5305, opened 2018-11-02
Please build Syncthing with GOARM


drizzt:
Hi,
I see that syncthing is built, for arm, without specify the [GOARM variable](https://github.com/golang/go/wiki/GoArm).

I tested on my NAS (Synology DS416j) and I see a big difference from the official build (linux-arm) and a custom build, built with GOARM=7.

Since the time to build Syncthing is very low I suggest you to change build.go and the release script in order to build for arm5 (GOARM=5), arm6 (GOARM=6) and arm7 (GOARM=7) as different builds

calmh:
Our binaries are built with GOARM=5. The upgrade system looks for a binary matching GOOS-GOARCH, and these do not differ between GOARM variants, so we can only easily provide one binary version. We used to do it differently, but it leads to all kinds of interesting issues, especially for people who actually build the binary *on* ARM, and we stopped doing that (#1051, #1218). I'm content with providing just a single lowest-common-denominator binary for 32 bit arm for now.

thePanz:
Where is the GOARM=5 defined?

calmh:
It's an environment variable. Set it before building.

PR #5491, opened 2019-01-31
WIP: Add GOARM handling

thePanz:
### Purpose

Allow building ARM specific binaries given by specifying the `goarm` parameter

Looks like the GOARM configuration is not specified nowhere in the code, this changes keeps the current behavior while allowing the GOARM parameter to be configured and used at build time.

Related issues:
- #5305 

The reason for this MR: I'd like to directly run Syncthing on my RaspberryPI without compiling from sources on that (very limited) hardware.

### DIFF
--- a/build.go
+++ b/build.go
@@ -35,6 +35,7 @@ import (
 
 var (
 	versionRe     = regexp.MustCompile(`-[0-9]{1,3}-g[0-9a-f]{5,10}`)
+	goarm         string
 	goarch        string
 	goos          string
 	noupgrade     bool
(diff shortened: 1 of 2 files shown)


calmh:
This, unfortunately, doesn't help on its own. The upgrade system doesn't know to look for the right kind of binary as it only looks at OS and arch. One could imagine extending the upgrade system, and tagging all binaries with what GOARM they were built with. This fails when building directly on an ARM system, as you will get some default GOARM value from the compiler but we don't know what it is when unspecified. That could be worked around by guessing (but there are cases where the guess will be wrong) or requiring the user to specify (onerous) and all in all it becomes complicated - too complicated for what it's worth.

So as is, we've decided to let this be a factor of the environment and build with some maximally compatible default for our own binaries. If those binaries are not good enough you will unfortunately need to build from source, or use some distribution that builds suitable packages.

thePanz:
Thanks for your reply Jackob.
Do you have examples on cross-compiling in Go? could be interesting to have
such documentation in the readme section too

Thank you again! :)

calmh:
Yes, https://docs.syncthing.net/dev/building.html in addition to above

PR #5494, opened 2019-01-31
cmd/syncthing: Pass SIGTERM on in monitor (fixes #5493)

### LINKED ISSUE #5493: Monitor bug: TERM signaling not passed to syncthing process
### Details
... as further detailed in https://forum.syncthing.net/t/cli-shutdown/12798
 
"When intercepting TERM it kills the main syncthing process, instead of signalling it to stop" [[2](https://forum.syncthing.net/t/cli-shutdown/12798/2)]

### Version Information
Tested with
Syncthing Version: Syncthing v1.0.0 "Erbium Earthworm" (Macports build)
OS Version: ProductName:	Mac OS X ProductVersion:	10.11.6  BuildVersion:	15G22010

imsodin:
Tested manually.

### DIFF
--- a/cmd/syncthing/monitor.go
+++ b/cmd/syncthing/monitor.go
@@ -127,7 +127,7 @@ func monitorMain(runtimeOptions RuntimeOptions) {
 		select {
 		case s := <-stopSign:
 			l.Infof("Signal %d received; exiting", s)
-			cmd.Process.Kill()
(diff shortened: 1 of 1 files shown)


calmh:
Could have passed on the same signal we got, maybe, but this is probably equally fine.

PR #5498, opened 2019-02-01
cmd/syncthing: fix strings.HasPrefix args order

Quasilyte:
Signed-off-by: Iskander Sharipov <quasilyte@gmail.com>


Quasilyte:
I found it while testing new static analysis check. :)
Part of my process is to send fixed to the upstream.

PR #5500, opened 2019-02-02
all: Bunch of more linter fixes

calmh:
This is so boring


calmh:
There's no need there's just a shitload of underscores

PR #5501, opened 2019-02-02
all: Even more boring linter fixes

calmh:
More of same, except there was an actual error checking bug in this one. :)

### DIFF
--- a/lib/osutil/atomic.go
+++ b/lib/osutil/atomic.go
@@ -81,7 +81,7 @@ func (w *AtomicWriter) Close() error {
 	}
 
 	// Try to not leave temp file around, but ignore error.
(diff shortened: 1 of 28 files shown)


AudriusButkevicius:
This is seriously non-sense. If you expect people to write code like this, I suspect we won't get anything done most of the time.

calmh:
It would be nonsense if there weren't also actual bugs sprinkled among these things.

calmh:
Honestly we can probably revert the underscores crap later. We'll get linter warnings on new code, but that's kinda fine as it's just a note that the error checking requires a though.

calmh:
I removed the underscoryness again

PR #5509, opened 2019-02-05
lib/model: Improve TestIssue5063

imsodin:
This lets the test time out after 10s instead of triggering the global timeout and panicking when failing. It also does some shuffling around of cleanup and waitgroup to prevent `directory not empty` errors from `os.RemoveAll` on openbsd.

### DIFF
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -17,6 +17,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"runtime/pprof"
 	"strconv"
 	"strings"
 	"sync"
@@ -1122,8 +1123,9 @@ func TestIssue5063(t *testing.T) {
 	wcfg, m := newState(defaultAutoAcceptCfg)
 	defer testOs.Remove(wcfg.ConfigPath())
 
-	addAndVerify := func(wg *sync.WaitGroup) {
-		id := srand.String(8)
+	wg := sync.WaitGroup{}
+
+	addAndVerify := func(id string) {
 		m.ClusterConfig(device1, protocol.ClusterConfig{
 			Folders: []protocol.Folder{
 				{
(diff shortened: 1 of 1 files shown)


lib/model/model_test.go
```diff
@@ -1132,20 +1134,37 @@ func TestIssue5063(t *testing.T) {
...
+	ids := make([]string, reps)
+	for i := 0; i < reps; i++ {
 		wg.Add(1)
-		go addAndVerify(wg)
+		ids[i] = srand.String(8)
+		go addAndVerify(ids[i])
 	}
+	defer func() {
+		for _, id := range ids {
+			testOs.RemoveAll(id)
```
golangcibot: Error return value of `testOs.RemoveAll` is not checked (from `errcheck`)

lib/model/model_test.go
```diff
@@ -1132,20 +1134,37 @@ func TestIssue5063(t *testing.T) {
...
-	wg.Wait()
+	finished := make(chan struct{})
+	go func() {
+		wg.Wait()
+		close(finished)
+	}()
+	select {
+	case <-finished:
+	case <-time.After(10 * time.Second):
+		pprof.Lookup("goroutine").WriteTo(os.Stdout, 1)
```
golangcibot: Error return value of `(*runtime/pprof.Profile).WriteTo` is not checked (from `errcheck`)

calmh:
While you have it fresh in mind, can you add some commentary to the function head about what the hell the test is testing? I’ve realized that numbered test cases are a horror to look for after the fact and must always have a description for what they’re about.... which should have been obvious perhaps.

AudriusButkevicius:
I think this was some service modifying config that was passed to wrapper.replace(), which as a result would corrupt the config, because we did not copy, which would then cause a just introduced device/folder to go missing.

imsodin:
I don't really know myself - I debugged issues that weren't related to what the test really wants to reproduce. I quickly scanned the issue/PR and added my "findings", which aren't that enlightening I fear.

AudriusButkevicius:
Loving the new bot.

imsodin:
Well, the red cross unsettles me a bit, but then again it found something relevant here :)

AudriusButkevicius:
I think the religious `_ =` is a bit stupid. My IDE already highlights it, so it's usually a concious decision from my side.
Anyways, until the redness goes away, I can't merge.

imsodin:
Ah wait, I thought GolangCI passing is not required for merge as announced in the forum? @calmh

calmh:
Looks to me like it just wants one approving review

calmh:
Now that I approved it I have the normal green merge button

calmh:
screenshot 2019-02-05 at 23 15 53

calmh:
And I agree, don't do the `_ =` dance unless there is specific reason to highlight that we're skipping to check an error return for whatever reason.

PR #5511, opened 2019-02-05
lib/model: Scan conflicts after creation

imsodin:
The functional change is to scan for newly created conflicts. With watching for changes enabled this isn't much of an improvement (but doesn't hurt either), without it means those files will be propagated much faster - we already know about them anyway.

`TestRequestRemoteRenameChanged` was flaky, as it depended on a race between the remote and local change. If the local change comes out as the winner (global file), which happened in most cases, no conflict was created and the test passed. If the remote wins the local change is moved for conflict and the file overwritten with the remote -> test fails. Now the test is adjusted such that always the remote version wins and it is tested that both the remote version and the local version (as conflict) are sent to the remote and exist with the correct data locally 

The `Debugln` change is unrelated: It showed try 0, 1, 2, now it's the more intuitive 1, 2, 3.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -196,7 +196,7 @@ func (f *sendReceiveFolder) pull() bool {
 
 		changed := f.pullerIteration(ignores, folderFiles, scanChan)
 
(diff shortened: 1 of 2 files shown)


imsodin:
Any comments on this?

PR #5513, opened 2019-02-06
lib/model: Use functions to generate config

imsodin:
Use functions (methods on `config.Wrapper`, test utility functions, ...) to handle configuration - it's less error prone (and less redundant).

### DIFF
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -54,7 +54,7 @@ func init() {
 	defaultFolderConfig = testFolderConfig("testdata")
 
 	defaultCfgWrapper = createTmpWrapper(config.New(myID))
-	defaultCfgWrapper.SetDevice(config.DeviceConfiguration{DeviceID: device1})
+	defaultCfgWrapper.SetDevice(config.NewDeviceConfiguration(device1, "device1"))
 	defaultCfgWrapper.SetFolder(defaultFolderConfig)
 	opts := defaultCfgWrapper.Options()
 	opts.KeepTemporariesH = 1
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -51,24 +51,17 @@ func init() {
 
 	defaultFs = fs.NewFilesystem(fs.FilesystemTypeBasic, "testdata")
 
-	defaultFolderConfig = config.NewFolderConfiguration(myID, "default", "default", fs.FilesystemTypeBasic, "testdata")
-	defaultFolderConfig.Devices = []config.FolderDeviceConfiguration{
-		{DeviceID: myID},
-		{DeviceID: device1},
-	}
-	defaultFolderConfig.FSWatcherEnabled = false
-	defaultCfg = config.Configuration{
-		Version: config.CurrentVersion,
-		Folders: []config.FolderConfiguration{defaultFolderConfig},
-		Devices: []config.DeviceConfiguration{
-			config.NewDeviceConfiguration(myID, "myID"),
-			config.NewDeviceConfiguration(device1, "device1"),
-		},
-		Options: config.OptionsConfiguration{
-			// Don't remove temporaries directly on startup
-			KeepTemporariesH: 1,
-		},
-	}
+	defaultFolderConfig = testFolderConfig("testdata")
+
+	defaultCfgWrapper = createTmpWrapper(config.New(myID))
+	defaultCfgWrapper.SetDevice(config.DeviceConfiguration{DeviceID: device1})
+	defaultCfgWrapper.SetFolder(defaultFolderConfig)
+	opts := defaultCfgWrapper.Options()
+	opts.KeepTemporariesH = 1
+	defaultCfgWrapper.SetOptions(opts)
+
+	defaultCfg = defaultCfgWrapper.RawCopy()
+
 	defaultAutoAcceptCfg = config.Configuration{
 		Devices: []config.DeviceConfiguration{
 			{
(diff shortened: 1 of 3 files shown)


lib/model/model_test.go
```diff
@@ -51,24 +51,17 @@ func init() {
...
-		},
-		Options: config.OptionsConfiguration{
-			// Don't remove temporaries directly on startup
-			KeepTemporariesH: 1,
-		},
-	}
+	defaultFolderConfig = testFolderConfig("testdata")
+
+	defaultCfgWrapper = createTmpWrapper(config.New(myID))
+	defaultCfgWrapper.SetDevice(config.DeviceConfiguration{DeviceID: device1})
```
golangcibot: Error return value of `defaultCfgWrapper.SetDevice` is not checked (from `errcheck`)

lib/model/model_test.go
```diff
@@ -51,24 +51,17 @@ func init() {
...
+	defaultCfgWrapper.SetFolder(defaultFolderConfig)
```
golangcibot: Error return value of `defaultCfgWrapper.SetFolder` is not checked (from `errcheck`)

lib/model/model_test.go
```diff
@@ -51,24 +51,17 @@ func init() {
...
+	opts := defaultCfgWrapper.Options()
+	opts.KeepTemporariesH = 1
+	defaultCfgWrapper.SetOptions(opts)
```
golangcibot: Error return value of `defaultCfgWrapper.SetOptions` is not checked (from `errcheck`)

lib/model/requests_test.go
```diff
@@ -330,15 +330,11 @@ func pullInvalidIgnored(t *testing.T, ft config.FolderType) {
...
-	cfg.Folders[0].Devices = []config.FolderDeviceConfiguration{
-		{DeviceID: myID},
-		{DeviceID: device1},
-	}
-	cfg.Folders[0].Type = ft
-	m, fc, w := setupModelWithConnectionManual(cfg)
+	w := createTmpWrapper(defaultCfgWrapper.RawCopy())
+	fcfg, tmpDir := testFolderConfigTmp()
+	fcfg.Type = ft
+	w.SetFolder(fcfg)
```
golangcibot: Error return value of `w.SetFolder` is not checked (from `errcheck`)

lib/model/requests_test.go
```diff
@@ -723,25 +719,30 @@ func TestRequestSymlinkWindows(t *testing.T) {
...
+	cfg := config.NewFolderConfiguration(myID, "default", "default", fs.FilesystemTypeBasic, path)
+	cfg.FSWatcherEnabled = false
+	cfg.Devices = append(cfg.Devices, config.FolderDeviceConfiguration{DeviceID: device1})
+	return cfg
+}
 
+func setupModelWithConnection() (*Model, *fakeConnection, string, *config.Wrapper) {
+	w := createTmpWrapper(defaultCfgWrapper.RawCopy())
+	fcfg, tmpDir := testFolderConfigTmp()
+	w.SetFolder(fcfg)
```
golangcibot: Error return value of `w.SetFolder` is not checked (from `errcheck`)

imsodin:
They do indeed, and one of them is:

```
    folder_recvonly_test.go:176: insufficient space on home disk (/dev): 0.000000 % < 1 %
```

Has that any truth to it?

lib/model/model_test.go
```diff
@@ -51,24 +51,17 @@ func init() {
...
-		},
-		Options: config.OptionsConfiguration{
-			// Don't remove temporaries directly on startup
-			KeepTemporariesH: 1,
-		},
-	}
+	defaultFolderConfig = testFolderConfig("testdata")
+
+	defaultCfgWrapper = createTmpWrapper(config.New(myID))
+	defaultCfgWrapper.SetDevice(config.NewDeviceConfiguration(device1, "device1"))
```
golangcibot: Error return value of `defaultCfgWrapper.SetDevice` is not checked (from `errcheck`)

calmh:
Ehm. /dev indeed has zero bytes available, because it's a virtual device filesystem (like devfs on linux). That has nothing to do with the user homes... But we have a couple of places that uses the config wrapper with `/dev/null` as the config location...

calmh:
Yeah, it fails the same way on my desktop Mac, and this "fixes" it:

```diff
diff --git a/lib/model/folder_recvonly_test.go b/lib/model/folder_recvonly_test.go
index fb27715eb..ae9368c21 100644
--- a/lib/model/folder_recvonly_test.go
+++ b/lib/model/folder_recvonly_test.go
@@ -346,7 +346,7 @@ func setupROFolder() *Model {
        cfg := defaultCfg.Copy()
        cfg.Folders = append(cfg.Folders, fcfg)

-       wrp := config.Wrap("/dev/null", cfg)
+       wrp := config.Wrap("/tmp/foo", cfg)

        db := db.OpenMemory()
        m := NewModel(wrp, myID, "syncthing", "dev", db, nil)
```

It doesn't immediately make sense to me that the location of the config affects the scan call, so I think this is an indicator of a bug somewhere else?

imsodin:
> It doesn't immediately make sense to me that the location of the config affects the scan call, so I think this is an indicator of a bug somewhere else?

Why not? As a result of a scan, we will likely need to update the db, which means there should be enough space to do that.

calmh:
Maybe so. Does that come from the config location? Why did it start failing now? Anyway, the error is legit - we can't expect /dev to have free space.

imsodin:
Of course, and I fixed it by using `createTmpWrapper`, which apparently puts it in `/tmp` on everything and `./test-tmp` on windows`. It came up now, as we used manually created configs before, which were missing the home dir free space requirement. This is now set to the default value. We do use in-memory dbs, so the space requirement isn't so important, but we might write to the config. Admittedly that won't need much space, but it doesn't hurt either (and if ever some tests uses a on-disk db, it's good to have that check in place).

imsodin:
lgtm?

PR #5514, opened 2019-02-06
lib/versioner: Restore for all versioners, cross-device support

### LINKED ISSUE #1634: Use copy instead of rename for version folder on other file systems
I've set the version folder to a different device for staged file versioning, now I'm getting the following errors:

```
Puller: final: rename /home/hashworks/filename /mnt/WDGREEN2TB/Backups/Cloud/filename: invalid cross-device link
```

Syncthing should use soft links here.

### LINKED ISSUE #4586: Versioner file tagging is inconsistent

### LINKED ISSUE #4631: In-GUI restoration should be possible from trashcan versioner
Currently (#4602) it's not because it doesn't follow the filename tagging scheme. I propose that,

- GetFolderVersions, when seeing a file without a tag, uses the filename as is and the modtime as the time

- RestoreFolderVersions, when being asked to restore a file with a given tag and not finding it, checks to see if there is a file without the tag (and only that file?). If so, use it and don't care what the tag says. (Avoids trying to match modtimes with varying precision.)

The modtimes in the trashcan versioner are the time of archiving, so already make sense as is. We just need to pick them up and allow restoration of the files. There will never be more than one version of a file, but the date selection can still make sense to restore any files changed yesterday and yesterday only, for example.

### LINKED ISSUE #5338: Version restore not possible with custom versioning path set
Syncthing cannot find previous versions when using _Staggered File Versioning_ using the option to define a folder path for the versioning. 
While syncthing correctly saves versions in the defined folder, when clicking "Versions", it display "No Data.", making it impossible to restore previous versions.

### Version Information

Syncthing Version: v0.14.52, Linux (ARM)
OS Version: Raspbian GNU/Linux, 9 (stretch)
Browser Version: Firefox Dev Ed, 64.0b12 (64-bit)

### LINKED ISSUE #5419: File versioning across drives
I set Syncthing running on Windows to do Staggered File Versioning from D:\Synced\Music to H:\Syncthing History\Music, and it fails with "The system cannot move the file to a different disk drive." displayed in the Web GUI's Could not Sync dialog.

I expected it to perform the versioning across the two drives without hassle.

To reproduce..
 1. Create a synced folder on one drive (let's call it Drive A)
 2. Configure Syncthing on a Windows host to do Staggered File Versioning, with the Versions Path set to a different drive (let's call this one Drive B)
 3. Delete a file or folder from the synced folder on a different computer
 4. Syncthing on the Windows host will attempt to copy the files, but fail with "The system cannot move the file to a different disk drive."

### Version Information

Syncthing Version: v0.14.54
OS Version: Windows 10 Pro for Workstations, build 1809
Browser Version: N/A

AudriusButkevicius:
### Purpose

Fixes #4631
Fixes #4586
Fixes #1634
Fixes #5338
Fixes #5419

This also:
~~1. Forces all versioners to use the same tagging, meaning that trash can versioner will start adding timestamps (cleanup is mtime based so it's ok)~~
2. Removes Rename function which seems not to be used anywhere important anymore, and TryRename is now RenameOrCopy
3. Versioning (and ton of other stuff) should now happen across filesystems when possible.
4. Adds versioning support for untagged and old file.txt~2014... tagged files. (There was no other way, in order to support trashcan versioner)

### Testing

Mended existing tests, now they pass sanely, added extra tests for RenameAndCopy (manually tested cross device scenario)


lib/model/model.go
```diff
@@ -2439,69 +2393,22 @@ func (m *Model) RestoreFolderVersions(folder string, versions map[string]time.Ti
...
-			continue
+		if err := ver.Restore(file, version); err != nil {
+			restoreErrors[file] = err.Error()
 		}
 	}
 
 	// Trigger scan
 	if !fcfg.FSWatcherEnabled {
-		m.ScanFolder(folder)
+		go m.ScanFolder(folder)
```
golangcibot: Error return value of `m.ScanFolder` is not checked (from `errcheck`)

AudriusButkevicius:
Another thing that would be possible, is to write a migration that migrates all files in versioning directories to the new tagging system removing a bunch of complicated logic and normalising stuff. That migration could also remove the external path madness in staggered versioner, and stop folders if that is not possible (no symlinks on windows etc).

I would really be in favor of this rather than having to support the legacy.

The external directory stuff is not so bad given we now have cross filesystem "renames", but I still think it should be a symlink setup by user, rather than a text field we provide.

lib/versioner/trashcan.go
```diff
@@ -44,49 +42,7 @@ func NewTrashcan(folderID string, fs fs.Filesystem, params map[string]string) Ve
...
-		return err
-	}
-
-	// Set the mtime to the time the file was deleted. This is used by the
-	// cleanout routine. If this fails things won't work optimally but there's
-	// not much we can do about it so we ignore the error.
-	t.fs.Chtimes(archivedPath, time.Now(), time.Now())
-
-	return nil
+	return archiveFile(t.fs, t.fs, ".stversions", filePath)
```
calmh: I'm sympathetical to the desire to haromonize, but *not* renaming files is the only point of the trashcan versiioner. This change makes it equivalent to a simple versioner that never throws away files, which is a different thing that we already have.

AudriusButkevicius: Sure, but then version restoration will always be able to pick up random files in the version directory, which I guess I don't mind.

AudriusButkevicius:
Also, I am now backtracking on killing the custom location for versions. I think perhaps we should even expand it and add support for it everywhere together with ability to specify filesystem type so people can version to s3 if they want to.

lib/versioner/trashcan.go
calmh: I don't see that as a problem either. If it's in the versioning dir, it's a version. If it's untagged we assume it's the only one and that it dates from its modtime.

calmh:
Sure

AudriusButkevicius:
Ready

AudriusButkevicius:
image

build.go
```diff
@@ -338,7 +338,7 @@ func test(pkgs ...string) {
 
 	useRace := runtime.GOARCH == "amd64"
 	switch runtime.GOOS {
-	case "darwin", "linux", "freebsd", "windows":
+	case "darwin", "linux", "freebsd": // , "windows": # See https://github.com/golang/go/issues/27089
```
imsodin: What's this? Seems unrelated

lib/osutil/osutil.go
```diff
@@ -22,37 +22,78 @@ import (
...
+			}
+
+			commonBytes := make([]byte, 0, smaller)
+
+			for i := 0; i < smaller; i++ {
+				if srcURI[i] != dstURI[i] {
+					break
+				}
+				commonBytes = append(commonBytes, srcURI[i])
+			}
```
imsodin: Just something I noticed: `commonBytes` must not represent a valid path , e.g. `/real/fake-first` and `/real/fake-second` results in `/real/fake`. Any operation on this filesystem should fail, so it probably will still do the right thing (not do anything). And bytes might be faster than a solution using `filepath.Split` or `strings.Split(path, fs.PathSeparator)` - or not.

build.go
AudriusButkevicius: it is unrelated, but I can't run tests locally, I guess I can move this into a separate PR

lib/fs/util.go
```diff
@@ -89,3 +90,63 @@ func IsParent(path, parent string) bool {
...
+
+	// Note, we treat string as []byte, not []rune as is often
+	// done in Go. (And sep as byte, not rune). This is because
+	// most/all supported OS' treat paths as string of non-zero
+	// bytes. A filename may be displayed as a sequence of Unicode
+	// runes (typically encoded as UTF-8) but paths are
+	// not required to be valid UTF-8 or in any normalized form
+	// (e.g. "é" (U+00C9) and "é" (U+0065,U+0301) are different
+	// file names.
+	c := []byte(path.Clean(paths[0]))
```
imsodin: Use `filepath.Clean` (also below) to use `os.PathSeparator` not always `/` (and fiilepath is already imported).

lib/fs/util_test.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	if runtime.GOOS == "windows" {
+		test(`c:\Audrius\Downloads`, `c:\Audrius\Docs`, `c:\Audrius\`)
+		test(`c:\Audrius\Downloads`, `C:\Audrius\Docs`, ``) // Case differences :(
+		test(`c:\Audrius-a\Downloads`, `c:\Audrius-b\Docs`, `c:\`)
+	} else {
+		test(`/Audrius/Downloads`, `/Audrius/Docs`, `/Audrius/`)
+		test(`/Audrius\Downloads`, `/audrius\Docs`, ``)
+		test(`/Audrius-a/Downloads`, `/Audrius-b/Docs`, ``)
```
imsodin: Why is the common path empty? `` and `.` signify the current working dir, while the common dir is `/`.

lib/fs/util.go
```diff
@@ -89,3 +90,63 @@ func IsParent(path, parent string) bool {
...
+	for i := len(c) - 1; i >= 0; i-- {
+		if c[i] == PathSeparator {
+			c = c[:i]
+			break
+		}
+	}
+
+	if len(c) > 0 {
+		c = append(c, PathSeparator)
+	}
```
imsodin: Don't add a trailing path separator - in general don't handle them manually, use `filepath.Clean` to keep paths all over the place consistent.

lib/fs/util.go
```diff
@@ -89,3 +90,63 @@ func IsParent(path, parent string) bool {
...
+				c = c[:i]
+				break
+			}
+		}
+	}
+
+	// Remove trailing non-separator characters and the final separator
+	for i := len(c) - 1; i >= 0; i-- {
+		if c[i] == PathSeparator {
+			c = c[:i]
```
imsodin: This also removes the path separator of root. Use `c = c[:i+1]` and `filepath.Clean` on the result (see below) - that will also remove trailing path separators.

lib/fs/util.go
AudriusButkevicius: Sorry, this comment is a bit unclear. I guess you are saying just to use filepath rather than path?

lib/fs/util_test.go
AudriusButkevicius: Fair point.

lib/fs/util.go
imsodin: Yes. The package `filepath` (well `path/filepath`) is a system-specific variant of `path`, i.e. `path` always expects slash as path separator, while `filepath` uses `os.PathSeparator`.

AudriusButkevicius: Man, this common path problem is much harder than I expected. relative vs abs paths, "" vs "/" vs "." etc. Edge case galore.

lib/fs/util.go
```diff
@@ -91,62 +90,42 @@ func IsParent(path, parent string) bool {
...
-		c = append(c, PathSeparator)
+	if len(common) == 0 {
+		if runtime.GOOS != "windows" && filepath.IsAbs(first) && filepath.IsAbs(second) {
+			return "/"
+		}
+		return ""
 	}
 
-	return string(c)
+	result := filepath.Clean(strings.Join(common, string(PathSeparator)))
```
imsodin: `filepath.Join(common...)`

lib/fs/util.go
```diff
@@ -91,62 +90,42 @@ func IsParent(path, parent string) bool {
...
+
+	if runtime.GOOS == "windows" {
+		if len(result) == 3 && strings.HasSuffix(result, ":.") {
+			// filepath.Clean("C:\") return "C:.", fix that up.
```
imsodin: Seriously? That's horrible.

lib/fs/util.go
```diff
@@ -91,62 +90,42 @@ func IsParent(path, parent string) bool {
...
+	count := len(firstParts)
+	if len(secondParts) < len(firstParts) {
+		count = len(secondParts)
 	}
 
-	// Remove trailing non-separator characters and the final separator
-	for i := len(c) - 1; i >= 0; i-- {
-		if c[i] == PathSeparator {
-			c = c[:i]
+	common := make([]string, 0, count)
```
imsodin: Instead of creating `common` you could just track the number of equal component and then `common == firstParts[:i]`.

AudriusButkevicius:
@imsodin ready for further probes

lib/fs/util.go
```diff
@@ -89,3 +89,49 @@ func IsParent(path, parent string) bool {
...
+func CommonPrefix(first, second string) string {
+	if filepath.IsAbs(first) != filepath.IsAbs(second) {
+		// Whatever
+		return ""
+	}
+
+	firstParts := strings.Split(filepath.Clean(first), string(PathSeparator))
+	secondParts := strings.Split(filepath.Clean(second), string(PathSeparator))
+
+	isAbs := filepath.IsAbs(first) && filepath.IsAbs(second)
```
imsodin: Just so I have something to remark: After the first check above you wouldn't have to check both :P

lib/fs/util.go
```diff
@@ -89,3 +89,49 @@ func IsParent(path, parent string) bool {
...
+
+	common := make([]string, 0, count)
+	for i := 0; i < count; i++ {
+		if firstParts[i] != secondParts[i] {
+			break
+		}
+		common = append(common, firstParts[i])
+	}
+
+	if runtime.GOOS == "windows" && isAbs &&
```
imsodin: No line breaks please xD (also below).  
Also I'd personally wrap the whole thing into `if isAbs {` to make it obvious it's about handling weird absolute path root edge cases (maybe not so edgy, but weird). However less indentation is also nice, so can be either way.

lib/fs/util_test.go
```diff
@@ -0,0 +1,48 @@
...
+		t.Helper()
+		res := CommonPrefix(first, second)
+		if res != expect {
+			t.Errorf("Expected %s got %s", expect, res)
+		}
+	}
+
+	if runtime.GOOS == "windows" {
+		test(`c:\Audrius\Downloads`, `c:\Audrius\Docs`, `c:\Audrius`)
+		test(`c:\Audrius\Downloads`, `C:\Audrius\Docs`, ``) // Case differences :(
```
imsodin: I vaguely remember us handling drive letter case somehow in `basicFilesystem`, but not sure - more a note to self than an actionable review item.

lib/fs/util_test.go
```diff
@@ -0,0 +1,48 @@
...
+		test(`C:\Audrius-a\Downloads`, `C:\Audrius-b\Docs`, `C:\`)
+		test(`\\?\C:\Audrius-a\Downloads`, `\\?\C:\Audrius-b\Docs`, `\\?\C:\`)
+		test(`\\?\C:\Audrius\Downloads`, `\\?\C:\Audrius\Docs`, `\\?\C:\Audrius`)
+		test(`Audrius-a\Downloads`, `Audrius-b\Docs`, ``)
+		test(`Audrius\Downloads`, `Audrius\Docs`, `Audrius`)
+		test(`c:\Audrius\Downloads`, `Audrius\Docs`, ``)
+		test(`c:\`, `c:\`, `c:\`)
+		test(`\\?\c:\`, `\\?\c:\`, `\\?\c:\`)
+		test(`Audrius`, `Audrius`, `Audrius`)
+		test(`.`, `.`, `.`)
```
imsodin: Do the last two cases outside of the runtime if.

lib/fs/util.go
AudriusButkevicius: Why not? I want to know if both are non absolute. On unix it checks if first rune is / so this is pretty cheapo.

lib/fs/util.go
AudriusButkevicius: No cuz there is a magical append.

AudriusButkevicius: Or I could, but that confuses me as to what that does, is the backing array shared or not, etc? If not, it's still a copy...

lib/fs/util.go
imsodin: You already check whether the absolutemess of both paths is the same before, so checking just one path gives the same result

AudriusButkevicius: Single instruction versus "why on earth"

imsodin: Yep, and I shouldn't have wrote it in the first place, this discussion I started just wastes time over nothing.

AudriusButkevicius:
Buuump

AudriusButkevicius:
The small nitpicks can be addresses, but I want a bigger review first.

AudriusButkevicius:
When can I have a review?

AudriusButkevicius:
I think I addressed the points.

imsodin:
> It looks like on creation the filename format is "free" (`fileTagger` function argument) but on restoration it is not (`TagFilename` is being used) - that seems conflicting.

I'd like clarification on this: It looks like the `tagger fileTagger` is always `TagFilename` except for `Trashcan` - so far so good. However `TagFilename` (respectively `Un`-) is hardcoded in `retrieveVersions` and `archiveFile`.

AudriusButkevicius:
In `retriveVersions` we do
```
	taggedFilename := TagFilename(filePath, tag)
	oldTaggedFilename := filePath + tag
	untaggedFileName := filePath
```

```for _, candidate := range []string{taggedFilename, oldTaggedFilename, untaggedFileName} {```

So we try all combinations of possible file name.

I don't see where we are hard-coding anything in archive file. There is a hardcoded untag, but it's because untag knows how to untag all variations of the path I believe.

**Comment by imsodin** [comment]
So the used options are just tagging or no tagging (and some legacy tagging for restoring), right? Then my confusion stems from `type fileTagger func(string, string) string` looking like it allows almost anything, when it's essentially just a switch. I'd replace the `tagger` argument with a `notag bool` argument.

In general lgtm - while I obviously still can't claim to have thought it through entirely.

**Comment by AudriusButkevicius** [comment]
I am not sure if I understand your question.
If you have specific questions, can you please point me to the exact lines you have questions about, as now I just look at some function and it's not obvious what you are confused about.

Bare in mind this was written 3 months ago, so I myself don't exactly recall how this works.

File: lib/versioner/util.go [hunk 16]
```diff
@@ -38,11 +57,215 @@ func UntagFilename(path string) (string, string) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return files, nil
+}
+
+type fileTagger func(string, string) string
+
+func archiveFile(srcFs, dstFs fs.Filesystem, filePath string, tagger fileTagger) error {
```
imsodin: I propose to replace `tagger fileTagger` with `noTag bool` (and remove `type fileTagger` in general. Follow up at https://github.com/syncthing/syncthing/pull/5514/files#diff-607726c1924296ce5c4bb168265bf986R175

lib/versioner/util.go
```diff
@@ -38,11 +57,215 @@ func UntagFilename(path string) (string, string) {
...
+
+	file := filepath.Base(filePath)
+	inFolderPath := filepath.Dir(filePath)
+
+	err = dstFs.MkdirAll(inFolderPath, 0755)
+	if err != nil && !fs.IsExist(err) {
+		return err
+	}
+
+	ver := tagger(file, info.ModTime().Format(TimeFormat))
```
imsodin: Followup to comment on https://github.com/syncthing/syncthing/pull/5514/files#diff-607726c1924296ce5c4bb168265bf986R138 :
Replace with something like 
```
ver := file
if !noTag { ver = TagFilename(...) }
```

lib/versioner/util.go
AudriusButkevicius: I don't see how this makes it more flexible?

imsodin: It makes it less flexible, which is exactly what I intend to fix what I was originally confused about: Archiving is flexible (it uses `fileTagger`) and restoring is not (it uses `Un`-/`TagFilename` which understands "normal" tag or no tag). So I propose to reduce the flexibility to just tag or no tag instead of `fileTagger`.

AudriusButkevicius: I dunno, I kind of disagree with you.

imsodin: Ok fine with me.

AudriusButkevicius:
Merging for now, if @calmh has concerns, we can address them in a new PR.

PR #5520, opened 2019-02-10
ui: Localize items counts

mattrobenolt:
There were a few places that I noticed where counts weren't getting comma separated in the UI.

### Testing

Frankly, I have not tested this, just fixed through observation of what's being done elsewhere.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -376,7 +376,7 @@
                     <tr ng-if="model[folder.id].needTotalItems > 0">
                       <th><span class="fas fa-fw fa-cloud-download-alt"></span>&nbsp;<span translate>Out of Sync Items</span></th>
                       <td class="text-right">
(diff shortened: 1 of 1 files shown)


calmh:
Is the localization for other things done in `binary` etc? If so, I guess we could do the corresponding in `alwaysNumber` and skip a separate pipe?

mattrobenolt:
Yeah, from a quick glace, others are `binary` and some are `percent` as a filter.

mattrobenolt:
I think I know what you're asking. `binary` gets run through `unitPrefix` as seen here: https://github.com/syncthing/syncthing/blob/master/gui/default/syncthing/app.js#L204

Which internally does `toLocaleString` which is the same as what the `localeNumberFilter` does.

mattrobenolt:
From my very little knowledge about syncthing source code, I don't _think_ you can easily get rid of `localeNumber` and just do the logic inside of `alwaysNumber`. There are cases where the output of `alwaysNumber` get piped into something else, such as `percent`. And if we made `alwaysNumber` always localized into a string, that'd break anything expecting the output of that to be a number.

calmh:
Got it, thanks for clarifications.

PR #5521, opened 2019-02-11
etc: Add keywords to .desktop files (fixes #5365)

### LINKED ISSUE #5365: .desktop files: desktop-entry-lacks-keywords-entry
Hello!

It could be a good idea to add a Keywords entry to the .desktop files.

The Debian package linter complains about this:
 - https://lintian.debian.org/tags/desktop-entry-lacks-keywords-entry.html

This is defined in the spec here:
 - https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html

It helps users find the app faster.

Cheers,

nekr0z:
### Purpose

Make .deb builds more pleasing to the Debian package linter.

### Testing

Frankly, none.

### DIFF
--- a/etc/linux-desktop/syncthing-start.desktop
+++ b/etc/linux-desktop/syncthing-start.desktop
@@ -6,4 +6,5 @@ Exec=/usr/bin/syncthing -no-browser
 Icon=syncthing
 Terminal=false
 Type=Application
+Keywords=synchronization;daemon;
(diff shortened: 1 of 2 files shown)


imsodin:
Thanks!

nekr0z:
Glad to be of use! Thanks for Syncthing!

PR #5525, opened 2019-02-12
lib/config, lib/model: Use path from locations to check disk space for db

imsodin:
Currently we check disk space at the path passed to `config.Wrapper`, which works for `cmd/syncthing` as the database is at the same place, but it's not right in general. Now the check is done at `locations.Get(locations.Database)`.

### DIFF
--- a/lib/config/size.go
+++ b/lib/config/size.go
@@ -78,7 +78,7 @@ func (s *Size) ParseDefault(str string) error {
 	return err
 }
 
-func checkFreeSpace(req Size, usage fs.Usage) error {
+func CheckFreeSpace(req Size, usage fs.Usage) error {
 	val := req.BaseValue()
 	if val <= 0 {
 		return nil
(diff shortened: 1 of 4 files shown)


lib/config/size.go
```diff
@@ -78,7 +78,7 @@ func (s *Size) ParseDefault(str string) error {
 	return err
 }
 
-func checkFreeSpace(req Size, usage fs.Usage) error {
+func CheckFreeSpace(req Size, usage fs.Usage) error {
```
AudriusButkevicius: Move to osutil or fsutil or something?

AudriusButkevicius: I guess size would have to move too

imsodin: Indeed. I was thinking about moving the entire size thing to `fs`, as `Usage` lives there and juggling bytes seems more appropriate for `fs` than `config`, but didn't in favor of not moving code around. Do you consider `fs` as suitable?

AudriusButkevicius: It's a bit weird as size is this auxilary struct that does not represent anything in fs in general.

PR #5528, opened 2019-02-12
lib/model: Various model test fixes and polish

imsodin:
Quite a few  `defer m.Stop()` statements were missing. And cleanup of temporary directories caused test failures on openbsd, so just don't fail the test (compromise: temp stuff rarely persisting vs flaky test).

### DIFF
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -620,8 +620,6 @@ func TestDeviceRename(t *testing.T) {
 }
 
 func TestClusterConfig(t *testing.T) {
-	testOs := &fatalOs{t}
-
 	cfg := config.New(device1)
 	cfg.Devices = []config.DeviceConfiguration{
 		{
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
Why drop fatalOs in some cases but not the others?

lib/model/model_test.go
```diff
@@ -1329,22 +1312,22 @@ func TestAutoAcceptNewFolderPremutationsNoPanic(t *testing.T) {
 					m.Stop()
 					testOs.RemoveAll(id)
 					testOs.RemoveAll(label)
+					os.Remove(wcfg.ConfigPath())
```
AudriusButkevicius: TempOS still needed above?

imsodin: Yes, from my understanding of this test it is, as all loop iterations use the same id and label, thus failed removals will cause unwanted conflicts.

Issue #4085, opened 2017-04-06
Refactor out cmd/syncthing


calmh:
Currently cmd/syncthing contains a shitload of stuff. This makes sense as is, but complicates reuse. What reuse? Possibilities:

 - libsyncthing.so which is actually loadable from C land and contains all the required functionality.
 - An actual GUI application written as above or in Go directly (using Qt bindings for example)
 - The current CLI/web application.

Ideally cmd/syncthing would be the 100 or so lines requires to parse environment variables and command line switches, maybe load the config, and then `app := syncthing.NewApp(&cfg); app.Run()` or something like that.

Martchus:
Note only a few adjustments are required to make the `libsyncthing.so` use-case *basically* work:
* adjustments in Syncthing's repo itself: https://github.com/Martchus/syncthing/tree/libsyncthing
* C++ interface and CMake integration for building Syncthing as library (contains the commands one need to invoke to build Syncthing as library): https://github.com/Martchus/syncthingtray/tree/master/libsyncthing

(I didn't propose a PR yet because I want to see whether this actually leads to something before. Also my Go knowledge isn't the best so the code quality is likely not the best, too.)

One thing which is missing for the `libsyncthing.so` use-case is the ability to stop all ongoing background tasks (goroutines). This is required to restart the Syncthing instance running 'inside' the application. So when refactoring, it would be nice if that requirement would be considered.

AudriusButkevicius:
We should weed out the goroutines that did not shut down nicely after a stop

Martchus:
Those should not be hard to find. In my tests I just added a sleep() at the end and stopped the debugger. I guess all threads it shows (beside the main thread) are goroutines. Except one which is the garbage collector. At least I assume it also lives in its own thread.

Btw, not sure how to keep that beast under control. `cgo` apparently does some magic to start the gc on the first C-to-Go function call. Not sure when it stops the gc again. Maybe when all goroutines terminate? But I'm just asking out of interest. I don't think this really matters in practice as long as the gc doesn't cause crashes or unwanted behavior.

calmh:
The go runtime never terminates threads. You can count the number of goroutines though with something like runtime.NumGoRoutines() (which should be 1 if the main goroutine is the only one still alive and that’s where you’re doing the call).

calmh:
In addition to the already mentioned shutdown/restart issue we should also remove calls to log.Fatal and similar (these should instead be a return from the top level function with an error of some kind).

[...]

PR #5529, opened 2019-02-12
cmd/syncthing, lib/api: Separate api/gui into own package (ref #4085)

imsodin:
Factor out all code related to the gui/rest API into a separate package. I tried to do as little change as possible, i.e. `lib/gui` is huge and could probably be broken up further in another step. As little as possible means `main.go` contains relevant changes (obviously) and in `lib/gui` additional function parameters and an interface were added as well as some variables exported.

Tested manually.


lib/gui/gui_test.go
```diff
@@ -74,7 +75,8 @@ func TestStopAfterBrokenConfig(t *testing.T) {
 	}
 	w := config.Wrap("/dev/null", cfg)
 
-	srv := newAPIService(protocol.LocalDeviceID, w, "../../test/h1/https-cert.pem", "../../test/h1/https-key.pem", "", nil, nil, nil, nil, nil, nil, nil, nil)
+	locations.SetBaseDir(locations.ConfigBaseDir, "../../test/h1")
```
golangcibot: Error return value of `locations.SetBaseDir` is not checked (from `errcheck`)

lib/gui/gui_test.go
```diff
@@ -484,8 +484,8 @@ func startHTTP(cfg *mockedConfig) (string, error) {
 	addrChan := make(chan string)
 
 	// Instantiate the API service
-	svc := newAPIService(protocol.LocalDeviceID, cfg, httpsCertFile, httpsKeyFile, assetDir, model,
-		eventSub, diskEventSub, discoverer, connections, errorLog, systemLog, cpu)
+	locations.SetBaseDir(locations.ConfigBaseDir, "../../test/h1")
```
golangcibot: Error return value of `locations.SetBaseDir` is not checked (from `errcheck`)

lib/gui/gui_test.go
```diff
@@ -946,10 +946,11 @@ func TestEventMasks(t *testing.T) {
 	cfg := new(mockedConfig)
 	defSub := new(mockedEventSub)
 	diskSub := new(mockedEventSub)
-	svc := newAPIService(protocol.LocalDeviceID, cfg, "", "", "", nil, defSub, diskSub, nil, nil, nil, nil, nil)
+	locations.SetBaseDir(locations.ConfigBaseDir, "../../test/h1")
```
golangcibot: Error return value of `locations.SetBaseDir` is not checked (from `errcheck`)

lib/gui/gui_test.go
imsodin: Maybe `locations.SetBaseDir` should panic when called with an invalid argument? @AudriusButkevicius

calmh:
Didn’t read yet but one of the things that has annoyed me for years is that we talk about “GUI” when we really mean “api”. I think this should probably be lib/api as that is what it provides to 90%. The GUI stuff is mostly in the static serving and small magic things like cookies and meta.js

lib/api/api_test.go
```diff
@@ -74,7 +75,8 @@ func TestStopAfterBrokenConfig(t *testing.T) {
 	}
 	w := config.Wrap("/dev/null", cfg)
 
-	srv := newAPIService(protocol.LocalDeviceID, w, "../../test/h1/https-cert.pem", "../../test/h1/https-key.pem", "", nil, nil, nil, nil, nil, nil, nil, nil)
+	locations.SetBaseDir(locations.ConfigBaseDir, "../../test/h1")
```
golangcibot: Error return value of `locations.SetBaseDir` is not checked (from `errcheck`)

lib/api/api_test.go
```diff
@@ -484,8 +484,8 @@ func startHTTP(cfg *mockedConfig) (string, error) {
 	addrChan := make(chan string)
 
 	// Instantiate the API service
-	svc := newAPIService(protocol.LocalDeviceID, cfg, httpsCertFile, httpsKeyFile, assetDir, model,
-		eventSub, diskEventSub, discoverer, connections, errorLog, systemLog, cpu)
+	locations.SetBaseDir(locations.ConfigBaseDir, "../../test/h1")
```
golangcibot: Error return value of `locations.SetBaseDir` is not checked (from `errcheck`)

lib/api/api_test.go
```diff
@@ -946,10 +946,11 @@ func TestEventMasks(t *testing.T) {
 	cfg := new(mockedConfig)
 	defSub := new(mockedEventSub)
 	diskSub := new(mockedEventSub)
-	svc := newAPIService(protocol.LocalDeviceID, cfg, "", "", "", nil, defSub, diskSub, nil, nil, nil, nil, nil)
+	locations.SetBaseDir(locations.ConfigBaseDir, "../../test/h1")
```
golangcibot: Error return value of `locations.SetBaseDir` is not checked (from `errcheck`)

cmd/syncthing/main.go
```diff
@@ -631,7 +649,7 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 		}()
 	}
 
-	perf := cpuBench(3, 150*time.Millisecond, true)
+	perf := api.CpuBench(3, 150*time.Millisecond, true)
```
AudriusButkevicius: Seems like the wrong this place for this. Scanner maybe?

cmd/syncthing/main.go
```diff
@@ -795,8 +813,8 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 
 	if opts := cfg.Options(); build.IsCandidate {
 		l.Infoln("Anonymous usage reporting is always enabled for candidate releases.")
-		if opts.URAccepted != usageReportVersion {
-			opts.URAccepted = usageReportVersion
+		if opts.URAccepted != api.UsageReportVersion {
+			opts.URAccepted = api.UsageReportVersion
```
AudriusButkevicius: Move to lib/ur?

cmd/syncthing/main.go
```diff
@@ -810,7 +828,7 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 		cfg.Save()
 	}
 
-	usageReportingSvc := newUsageReportingService(cfg, m, connectionsService)
+	usageReportingSvc := api.NewUsageReportingService(cfg, m, connectionsService, noUpgradeFromEnv)
```
AudriusButkevicius: Lib/ur?

cmd/syncthing/main.go
```diff
@@ -1005,14 +1023,14 @@ func setupGUI(mainService *suture.Supervisor, cfg *config.Wrapper, m *model.Mode
 	cpu := newCPUService()
 	mainService.Add(cpu)
 
-	api := newAPIService(myID, cfg, locations.Get(locations.HTTPSCertFile), locations.Get(locations.HTTPSKeyFile), runtimeOptions.assetDir, m, defaultSub, diskSub, discoverer, connectionsService, errors, systemLog, cpu)
-	cfg.Subscribe(api)
-	mainService.Add(api)
+	apiSvc := api.NewAPIService(myID, cfg, runtimeOptions.assetDir, tlsDefaultCommonName, m, defaultSub, diskSub, discoverer, connectionsService, errors, systemLog, cpu, &exiter{}, noUpgradeFromEnv)
+	cfg.Subscribe(apiSvc)
```
AudriusButkevicius: Api should probably do this itself, it has the config.

lib/api/api.go
```diff
@@ -147,35 +150,54 @@ type rater interface {
...
+	Shutdown()
+}
+
+type ApiService interface {
+	suture.Service
+	config.Committer
+	WaitForStart()
+}
+
+func NewAPIService(id protocol.DeviceID, cfg configIntf, assetDir, tlsDefaultCommonName string, m modelIntf, defaultSub, diskSub events.BufferedSubscription, discoverer discover.CachingMux, connectionsService connectionsIntf, errors, systemLog logger.Recorder, cpu rater, main mainIntf, noUpgrade bool) ApiService {
```
AudriusButkevicius: Just api.New and api.Service?

lib/api/api_test.go
AudriusButkevicius: This leaks state between tests?

cmd/syncthing/main.go
```diff
@@ -264,6 +264,24 @@ func parseCommandLineOptions() RuntimeOptions {
 	return options
 }
 
+type exiter struct{}
```
AudriusButkevicius: Control or Controller or App? Also, interface that lives somewhere common oppose to oppurtunistic interface matching?

lib/api/api_test.go
```diff
@@ -27,11 +27,23 @@ import (
...
 	"github.com/syncthing/syncthing/lib/fs"
+	"github.com/syncthing/syncthing/lib/locations"
 	"github.com/syncthing/syncthing/lib/protocol"
 	"github.com/syncthing/syncthing/lib/sync"
 	"github.com/thejerf/suture"
 )
 
+func TestMain(m *testing.M) {
+	orig := locations.GetBaseDir(locations.ConfigBaseDir)
+	locations.SetBaseDir(locations.ConfigBaseDir, "testdata/config")
```
golangcibot: Error return value of `locations.SetBaseDir` is not checked (from `errcheck`)

lib/api/api_test.go
```diff
@@ -27,11 +27,23 @@ import (
...
+
+	exitCode := m.Run()
+
+	locations.SetBaseDir(locations.ConfigBaseDir, orig)
```
golangcibot: Error return value of `locations.SetBaseDir` is not checked (from `errcheck`)

cmd/syncthing/main.go
imsodin: It's also in lib/ur now. The memory and cpu stuff could be moved to their own package, but having them in usage reporting seems sane enough to me.

imsodin:
In the last two commit happened exactly the thing why I didn't want to further split up lib/api in this PR/commit: I had to do lots of changes. I could undo the last two commits and you first review just splitting out lib/api from cmd/syncthing as a huge block, then I'll PR again further splitting up of lib/api. Or you review this largish diff - just let me know which you prefer.

lib/api/debug.go
```diff
@@ -0,0 +1,26 @@
...
+var (
+	l = logger.DefaultLogger.NewFacility("http", "REST API")
+)
+
+func shouldDebugHTTP() bool {
+	return l.ShouldDebug("http")
+}
+
+func init() {
+	l.SetDebug("http", strings.Contains(os.Getenv("STTRACE"), "http") || os.Getenv("STTRACE") == "all")
```
AudriusButkevicius: Rename to API? 🙃 🤔 
No? Too brave?

cmd/syncthing/main.go
```diff
@@ -264,6 +266,23 @@ func parseCommandLineOptions() RuntimeOptions {
...
+	stop <- exitRestarting
+}
+
+func (_ *controller) Shutdown() {
+	l.Infoln("Shutting down")
+	stop <- exitSuccess
+}
+
+func (_ *controller) ExitUpgrading() {
+	stop <- exitUpgrading
```
AudriusButkevicius: Not related and please feel free to ignore, but could also log something

cmd/syncthing/main.go
```diff
@@ -783,20 +798,12 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
...
-	}
-
 	// Candidate builds always run with usage reporting.
 
 	if opts := cfg.Options(); build.IsCandidate {
 		l.Infoln("Anonymous usage reporting is always enabled for candidate releases.")
-		if opts.URAccepted != usageReportVersion {
-			opts.URAccepted = usageReportVersion
+		if opts.URAccepted != ur.UsageReportVersion {
+			opts.URAccepted = ur.UsageReportVersion
```
AudriusButkevicius: ur.Version?
Also, perhaps ur -> reporting?

cmd/syncthing/main.go
```diff
@@ -1079,7 +1090,8 @@ func standbyMonitor() {
 			// things a moment to stabilize.
 			time.Sleep(restartDelay)
 
-			restart()
+			l.Infoln("Restarting")
+			stop <- exitRestarting
```
AudriusButkevicius: Again, not related and can be ignored but I think we should init a global controller in init() {} or var, and then use that everywhere oppose to relying on some number of global channels in the main package and some magical struct in other cases.

lib/api/api.go
```diff
@@ -1253,27 +1137,27 @@ func (s *apiService) postDBIgnores(w http.ResponseWriter, r *http.Request) {
 	s.getDBIgnores(w, r)
 }
 
-func (s *apiService) getIndexEvents(w http.ResponseWriter, r *http.Request) {
-	s.fss.gotEventRequest()
+func (s *service) getIndexEvents(w http.ResponseWriter, r *http.Request) {
+	s.fss.GotEventRequest()
```
AudriusButkevicius: Again, unrelated, feel free to ignore, but this feels like a callback hence should be "On<Something>" rather than just "Something()"

lib/foldersummary/folder_summary.go
```diff
@@ -4,24 +4,38 @@
...
 // The folderSummaryService adds summary information events (FolderSummary and
 // FolderCompletion) into the event stream at certain intervals.
-type folderSummaryService struct {
+type service struct {
 	*suture.Supervisor
 
-	cfg       configIntf
-	model     modelIntf
+	cfg       ur.ConfigIntf
+	model     ur.ModelIntf
```
AudriusButkevicius: Having to reach out into UR is GRIMMMMMM

lib/foldersummary/folder_summary.go
```diff
@@ -210,19 +298,13 @@ func (c *folderSummaryService) sendSummary(folder string) {
 
 		// Get completion percentage of this folder for the
 		// remote device.
-		comp := jsonCompletion(c.model.Completion(devCfg.DeviceID, folder))
+		comp := model.JsonCompletion(c.model.Completion(devCfg.DeviceID, folder))
```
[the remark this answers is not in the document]
imsodin: It's needed in both foldersummary und ur, so I thought model was the common denominator.

lib/ur/usage_report.go
```diff
@@ -453,3 +474,62 @@ func cpuBenchOnce(duration time.Duration, useWeakHash bool, bs []byte) float64 {
...
+	GlobalSize(folder string) db.Counts
+	LocalSize(folder string) db.Counts
+	ReceiveOnlyChangedSize(folder string) db.Counts
+	CurrentSequence(folder string) (int64, bool)
+	RemoteSequence(folder string) (int64, bool)
+	State(folder string) (string, time.Time, error)
+	UsageReportingStats(version int, preview bool) map[string]interface{}
+	FolderErrors(folder string) ([]model.FileError, error)
+	WatchError(folder string) error
+}
```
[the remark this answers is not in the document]
imsodin: A less hackish alternative is to make model, wrapper, connections, ... interfaces. E.g. instead of using `*config.Wrapper` all over, use `config.Wrapper` which is an interface and in production is "backed by `*config.wrapper`. That however means those interfaces need to be complete and thus the mocked objects need some more methods.

cmd/syncthing/main.go
```diff
@@ -810,9 +817,21 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
...
+	setupGUI(mainService, cfg, m, defaultSub, diskSub, cachedDiscovery, connectionsService, usageReportingSvc, errors, systemLog, runtimeOptions)
+
+	myDev, _ := cfg.Device(myID)
+	l.Infof(`My name is "%v"`, myDev.Name)
+	for _, device := range cfg.Devices() {
+		if device.DeviceID != myID {
+			l.Infof(`Device %s is "%v" at %v`, device.DeviceID, device.Name, device.Addresses)
+		}
+	}
+
```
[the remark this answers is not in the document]
imsodin: I wanted to keep the log output of "gui is here" followed by "this is my device name and everyone elses" together. The usage reporting log is now before that block, instead of after, because I need usage reporting to pass to the api service

lib/api/debug.go
AudriusButkevicius: Dis hidden dunno why

lib/api/api.go
imsodin: Get the callback bit, but not the "On" vs "Something()" part.

cmd/syncthing/main.go
imsodin: ur.Version? Yes.  
reporting seems somewhat too generic, so I'd rather go with usagereporting, which seems a bit long. Not sure about this.

lib/api/api.go
AudriusButkevicius: Actually, this should probably just be called Invalidate()

imsodin: That I understand even less. The summary service records the time that `GotEventRequest` was last called as the time when an event was last requested - how does that invalidate anything?

cmd/syncthing/main.go
imsodin: Something like this? https://github.com/syncthing/syncthing/pull/5532

cmd/syncthing/main.go
AudriusButkevicius: Let's keep it as it is I guess.

lib/api/api.go
AudriusButkevicius: Yeah, you are right.
GotEventRequest() is like saying "what I got" rather than saying I call your "OnEventRequest()" when I get something. Essentially what I am saying is that GotEventRequest() feels like an accessor to some boolean flag.

lib/gui/gui_test.go
AudriusButkevicius: Yeah, sure

lib/ur/usage_report.go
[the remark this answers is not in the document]
imsodin: Currently there are `modelIntf`, `configIntf` and `connectionsIntf` in gui.go. I see three options to handle those with split packages:

1. Create them as `ModelIntf`, ... in their respective packages (model, ...).

2. In their respective packages replace the currently exported struct used as pointers in other packages (e.g. `*config.Wrapper`) with interfaces and internalize the struct.

3. The current "solution" in this PR: Move all these interfaces to a package that is imported, but doesn't import other consumers of these interfaces.

4. Move all interfaces to a dedicated `lib/interfaces` package.

1. and 4. are very similar and in my opinion defining the interfaces in their "package of origin" is much more sensible. I still think 2. is the "correct and clean" solution, but it's too much work. 3. is simple but ugly. I'd like to go ahead with 1., but would be fine with 2. as well and as always I am open to be convinced of anything else as well.

@AudriusButkevicius ?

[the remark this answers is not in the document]
AudriusButkevicius: I don't think we can define things in their location of origin as I though the purpose of the interfaces was to solve circular deps. If that is not the case then we can just hide the implementation by lowercasing it and always use the interface which I would be ok with. Yet as I said I was under the impression we can't do that.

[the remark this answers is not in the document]
imsodin: We can, these interface are purely for testing. They just allow starting services with mocked models, configs, ... instead of the real thing.

>  If that is not the case then we can just hide the implementation by lowercasing it and always use the interface which I would be ok with

Oh no, I was hoping you wouldn't like that approach xD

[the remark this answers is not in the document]
AudriusButkevicius: Perhaps worthy a separate PR

calmh:
So... I didn't read yet, but where are we with this? There are some outstanding comments. This is one of those that's going to be perpetually in merge conflict state if we don't get it done...

AudriusButkevicius:
I think it depends on one of the other prs to simplify

imsodin:
Yup, #5548 needs to get in, then I will resolve conflicts and I think all of Audrius current comments are addressed. Except for the brave one, which I just rediscovered now: https://github.com/syncthing/syncthing/pull/5529#discussion_r256591131

lib/api/debug.go
imsodin: Why not, we can still enable the new "api" facility when encountering "http" to stay backwards compatible. @calmh?

calmh: No stated compatibility on debug facilities, rename is fine by me.

lib/foldersummary/folder_summary.go
calmh: mmm this seems like too small packages perhaps? But I agree, this dependency edge is ugly.

imsodin: That's got a point, should `lib/foldersummary` go to `lib/ur`? It also a report of sorts. Will just mean that `ur.Service` will have to be come `ur.UsageReportingService` to prevent ambiguities with `foldersummary.Service` which then becomes `ur.FolderSummaryService`.

imsodin: Or `lib/model`? I know moving stuff to `lib/model` is the opposite way of what we usually do, but thematically it seems right to have it with all the rest of the folder code (moving folders to lib/folder might also be viable at some point).

imsodin:
This is ready for (another round of) review from my side.

imsodin:
Where do we stand with this?

AudriusButkevicius:
Need to look at this on a desktop

lib/model/folder_summary.go
```diff
@@ -36,7 +46,7 @@ type folderSummaryService struct {
 	lastEventReqMut sync.Mutex
 }
 
-func newFolderSummaryService(cfg config.Wrapper, m model.Model) *folderSummaryService {
+func NewFolderSummaryService(cfg config.Wrapper, m Model, id protocol.DeviceID) FolderSummaryService {
```
AudriusButkevicius: id not used?

imsodin: Good catch, it's worse: It should be assigned to the `folderSummary.id`, as that is used. That's something I'd expect a linter to catch (unused function param).

calmh:
giphy

I never read the diff but I'm totally on board with the concept and hope you didn't break anything :p

PR #5537, opened 2019-02-14
all: Get rid of fatal logging

imsodin:
### Purpose

A package/library has no business calling `os.Exit` ever, but that's what `logger.Fatal...` does. Solution: Panic on errors that just must not happen or bubble errors up until they reach `cmd/syncthing` (or whoever calls) and let them decide on what's to do.

### Testing

Tests pass and test setup starts and GUI reacts to input.

### DIFF
--- a/cmd/syncthing/gui.go
+++ b/cmd/syncthing/gui.go
@@ -78,9 +78,8 @@ type apiService struct {
 	stop               chan struct{} // signals intentional stop
 	configChanged      chan struct{} // signals intentional listener close due to config change
 	started            chan string   // signals startup complete by sending the listener address, for testing only
-	startedOnce        chan struct{} // the service has started successfully at least once
-	firstStartupFailed chan struct{} // the service failed to start successfully even once
-	firstStartupErr    error
+	startedOnce        chan struct{} // the service has started at least once
+	startupErr         error
 	cpu                rater
 
 	guiErrors logger.Recorder
@@ -168,7 +167,6 @@ func newAPIService(id protocol.DeviceID, cfg configIntf, httpsCertFile, httpsKey
 		stop:               make(chan struct{}),
 		configChanged:      make(chan struct{}),
 		startedOnce:        make(chan struct{}),
-		firstStartupFailed: make(chan struct{}),
 		guiErrors:          errors,
 		systemLog:          systemLog,
 		cpu:                cpu,
@@ -178,12 +176,8 @@ func newAPIService(id protocol.DeviceID, cfg configIntf, httpsCertFile, httpsKey
 }
 
 func (s *apiService) WaitForStart() error {
-	select {
-	case <-s.startedOnce:
-	case <-s.firstStartupFailed:
-		return s.firstStartupErr
-	}
-	return nil
+	<-s.startedOnce
+	return s.startupErr
 }
 
 func (s *apiService) getListener(guiCfg config.GUIConfiguration) (net.Listener, error) {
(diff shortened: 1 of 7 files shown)


cmd/syncthing/gui.go
```diff
@@ -410,6 +425,16 @@ func (s *apiService) Serve() {
...
 }
 
+// Complete implements suture.IsCompletable, which signifies to the supervisor
+// whether to stop restarting the service.
+func (s *apiService) Complete() bool {
+	select {
+	case <-s.firstStartupFailed:
+		return true
+	}
+	return false
```
golangcibot: unreachable code (from `govet`)

cmd/syncthing/gui.go
```diff
@@ -410,6 +425,16 @@ func (s *apiService) Serve() {
 	}
 }
 
+// Complete implements suture.IsCompletable, which signifies to the supervisor
+// whether to stop restarting the service.
+func (s *apiService) Complete() bool {
+	select {
```
golangcibot: should use a simple channel send/receive instead of select with a single case (from `gosimple`)

cmd/syncthing/main.go
```diff
@@ -921,33 +961,39 @@ func loadOrDefaultConfig() (*config.Wrapper, error) {
...
 	} else if err != nil {
-		l.Fatalln("Failed to load config:", err)
+		return nil, errors.Wrap(err, "Failed to load config")
 	}
 
 	if cfg.RawCopy().OriginalVersion != config.CurrentVersion {
 		err = archiveAndSaveConfig(cfg)
 		if err != nil {
-			l.Fatalln("Config archive:", err)
+			return nil, errors.Wrap(err, "Config archive")
```
AudriusButkevicius: Wraps convention is to use lowercase (everywhere), because if its wrapping multiple times you end up with a lot of weird capitalisation. Furthermore most of these will go into l.Warn which is already capitalised.

lib/config/folderconfiguration.go
```diff
@@ -106,7 +106,7 @@ func (f FolderConfiguration) Versioner() versioner.Versioner {
 	}
 	versionerFactory, ok := versioner.Factories[f.Versioning.Type]
 	if !ok {
-		l.Fatalf("Requested versioning type %q that does not exist", f.Versioning.Type)
+		panic(fmt.Sprintf("Requested versioning type %q that does not exist", f.Versioning.Type))
```
AudriusButkevicius: This was not printing stacktraces, now it will, which is a bit grim.

imsodin: I thought this could never happen anyway, thus panic. I could also just return an error and log a warning in the model and continue without a versioner.

cmd/syncthing/gui.go
```diff
@@ -236,14 +248,17 @@ func (s *apiService) Serve() {
 			// We let this be a loud user-visible warning as it may be the only
 			// indication they get that the GUI won't be available.
 			l.Warnln("Starting API/GUI:", err)
-			return
+
+		case <-s.firstStartupFailed:
```
AudriusButkevicius: How does this work? Why can't startedOnce handle this?

imsodin: `startedOnce` only gets close after a successful start including all setup and logging. I am not sure whether that's relevant or if we could just close it at the beginning of `Serve`, in which case `firstStartupFailed` is indeed redundant. I just went with the smallest change possible.  
Having a closer look I'd say all of the mux setup should probably be in `newApiService` anyway.

PR #5539, opened 2019-02-14
build: Add option to get test coverage

imsodin:
To get the whole codecov thing actually running someone with permissions to modify the build process on the team city CI. Just a step running tests with the new `-coverage` option and then upload the report to codecov (see https://github.com/codecov/example-go#travis-ci). Probably we should wait for the end of Jakobs holidays :)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,5 @@
+coverage:
+  range: "40...100"
+
+ignore:
+  - "**.pb.go"
--- a/build.go
+++ b/build.go
@@ -48,6 +48,7 @@ var (
 	pkgdir           string
 	cc               string
 	debugBinary      bool
+	coverage         bool
 	timeout          = "120s"
 	gogoProtoVersion = "v1.2.0"
 )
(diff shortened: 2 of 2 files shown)


AudriusButkevicius:
So I've added the coverage job. If you log in to codecov with your github creds you should be able to see the repo.
For some reason codecov thinks the builds are not finished hence does not publish coverage onto the PRs.

imsodin:
To me it looks like it's confused about the commit. Afaik there is some magic with new branches and merging happening on teamcity. Probably the coverage thing has to happen on a "non-magic" branch or commit/branch info must be specified explicitly to the codecov script (probably the latter). On phone right now so can't check.

AudriusButkevicius:
Fixed that, had one report where it worked out the root correctly, yet it still thinks something is pending. Perhaps Teamcity or github needs to call codecov?

imsodin:
Looks like it almost works now (it didn't pick up the `.codecov.yml` but I hope that happens on next master commit). The bots post are still missing as Syncthing doesn't have to codecov app activated and I don't seem to be able to do that, probably the following applies (even though it shouldn't in my opinion):  

> Tip: If an app requires organization-level access, then only an organization owner can purchase, install, or cancel the app, and manage app billing for the organization. If the app doesn't require organization-level access, then a repository administrator can install and uninstall the app.

AudriusButkevicius:
I don't see why the app needs access to the org, the damn bot just makes a comment.

PR #5543, opened 2019-02-16
build, script: Localize .desktop files (fixes #5535)

### LINKED ISSUE #5535: .desktop files lack localization
[Specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html) advises that `Name`, `GenericName`, `Comment` and `Keywords` entries in `.desktop` file's `[Desktop Entry]` section are _localestring_. This calls for adding language-specific lines, i.e.:
```
GenericName[en]=File synchronization
GenericName[sv]=Filsynkronisering
GenericName[ru]=Синхронизация файлов
```
Naturally, this all should be utilizing the current transifex process.

nekr0z:
Pick up lines that support translations from `.desktop` files and feed them to Transifex. When translations appear, put them back into corresponding `.desktop` files.

Tested manually.

### DIFF
--- a/build.go
+++ b/build.go
@@ -762,6 +762,7 @@ func translate() {
 func transifex() {
 	os.Chdir("gui/default/assets/lang")
 	runPrint(goCmd, "run", "../../../../script/transifexdl.go")
+	runPrint(goCmd, "run", "../../../../script/update-desktop-translations.go", "../../../../etc/linux-desktop/", "./")
 }
 
 func ldflags() string {
--- a/script/translate.go
+++ b/script/translate.go
@@ -129,19 +129,19 @@ func parseDesktop(filename string) {
 		}
 
 		switch value := strings.SplitN(line, "=", 2); value[0] {
-			case "Name", "GenericName", "Comment":
-				if len(value[1]) == 0 {
+		case "Name", "GenericName", "Comment":
+			if len(value[1]) == 0 {
+				continue
+			}
+			translation(value[1])
+		case "Keywords":
+			words := strings.Split(value[1], ";")
+			for _, word := range words {
+				if len(word) == 0 {
 					continue
 				}
-				translation(value[1])
-			case "Keywords":
-				words := strings.Split(value[1], ";")
-				for _, word := range words {
-					if len(word) == 0 {
-						continue
-					}
-					translation(word)
-				}
+				translation(word)
+			}
 		}
 	}
 }
--- a/script/translate.go
+++ b/script/translate.go
@@ -10,6 +10,7 @@ package main
 
 import (
 	"encoding/json"
+	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
@@ -21,6 +22,10 @@ import (
 
 var trans = make(map[string]string)
 var attrRe = regexp.MustCompile(`\{\{'([^']+)'\s+\|\s+translate\}\}`)
+
+// .desktop entries that can be localestrings
+var groupRe = regexp.MustCompile(`^\[Desktop Entry\]$`)
+var locRe = regexp.MustCompile(`^(Name|GenericName|Comment|Keywords)=.*\S*.*`)
 
 // exceptions to the untranslated text warning
 var noStringRe = regexp.MustCompile(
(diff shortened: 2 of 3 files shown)


AudriusButkevicius:
Aren't the desktop files just ini files? Perhaps they can be handled more easily?

nekr0z:
> Aren't the desktop files just ini files? Perhaps they can be handled more easily?

Sure, some ini interface like `github.com/go-ini/ini`, `github.com/claudetech/ini` or `github.com/drbig/simpleini`could result in tidier and (supposedly) a bit more future-proof implementation. I did'n go that road out of fear of introducing another dependency, but if any of those packages (or yet another package providing such an interface) look safe to you and other maintainers, I can rewrite to utilize that.

AudriusButkevicius:
I'd trust using a ini library more than getting regexes right. @imsodin wdyt?

nekr0z:
For what it's worth, I improved readability on regexes and tightened one of them up a bit.

By the way, I don't think we could get rid of regexes alltogether, since the whole `key[language]` thing seems to be `.desktop`-specific and not implemented in any ini-library anyway.

imsodin:
Unfortunately, I don't have a clear opinion: Having a library for a simple problem and non-standard ini vs cumbersome review of "self-made" implementation. From a very quick skim of the code which looked very organized, I'd tend to self-made implementation, then again I don't really have the time to review this. I know it' really harsh, but translating desktop files has probably my lowest possible personal priorisation. TLDR: I don't really care, but I'd probably review a shorter diff much earlier.

nekr0z:
Thanks, @imsodin!

I totally understand this is really a low-priority issue. And of course, this PR can wait until one of maintainers has time and inspiration to deal with it. In the meantime, I will insert some comments where it makes sence and will try to improve code readability more if I can.

I'll look further into ini libraries, but from what I've already researched there's not really that much to be gained in terms of code simplification by using them. The problem is, `ini` itself is a very loose-defined format, and the spec for our `.desktop` files is very strict and well laid out compared to general `ini` standarts (or lack thereof). Libraries tend to focus mostly on old MS Windows conventions of what is and is not allowed in `ini` files, and making a library follow `.desktop` file spec may require jumping through another set of hoops.

AudriusButkevicius:
Did you try a toml parser?

imsodin:
For reference: There is https://pyxdg.readthedocs.io/en/latest/desktopentry.html which is an official implementation in python.

nekr0z:
@AudriusButkevicius I did consider TOML, albeit briefly, but decided against it. Maybe I should review it more closely, thanks for suggestion.

There are things about TOML that I really don't like in regard to this issue here:
1. TOML spec insists strings are to be double-quoted, i.e.:
  `Name = "Syncthing Web UI"`,
  and writes output accordingly, which is simply not valid for a .desktop file. The correct syntax per .desktop spec is
  `Name=Syncthing Web UI`,
  and while spaces around `=` can be lived with (as spec requires they be ignored), `"` in value is a meaningful character and should be parsed verbatim.
2. Their parsers write out to the file according to their specs. There would be jumping through hoops, testing, some more testing and even more testing and jumping to make sure that we adhere to the spec that explicitly tells us to _preserve verbatim whatever we don't know how to read or are not sure how to properly parse_.

@imsodin Thanks a lot, I will smoke this one in as best as I can (although it being in python won't help at all).

imsodin:
> @imsodin Thanks a lot, I will smoke this one in as best as I can (although it being in python won't help at all).

I don't know what the general position is, but as this is about "file generation" not building I wouldn't rule out python just because it isn't go.

AudriusButkevicius:
> For reference: There is https://pyxdg.readthedocs.io/en/latest/desktopentry.html which is an official implementation in python.

As if some existing format wasn't good enough.
God, I hate people.

nekr0z:
> I wouldn't rule out python just because it isn't go.

A build dependency of **python**? For such a minor issue? You can't possibly be serious.

Researching the matter further and further I found myself starting to contemplate writing my own library, which scared me quite a bit. Sweet Lady of Discord, had I imagined how deep this rabbit hole goes, I wouldn't poke at this issue with a three-yard pole!

imsodin:
> Sweet Lady of Discord, had I imagined how deep this rabbit hole goes, I wouldn't poke at this issue with a three-yard pole!

I totally get that and I don't want you to get scared off from contributing by this ... whatever it is ;)

> A build dependency of python? For such a minor issue? You can't possibly be serious.

My point is that it isn't a build dependency, it's a "deployment" (?) dependency. What I mean is it only needs to be run when modifying .desktop files or by CI, and a python dep on a linux CI is trivial. Or more earnestly, not really fitting go-ini library, own library and existing dedicated python-library are probably equally annoying.  
In my opinion just leave this as is and don't put anymore time in this rabbit hole (there's plenty other ones :) ), someone (I) will eventually come around to reviewing it.

nekr0z:
> In my opinion just leave this as is and don't put anymore time in this rabbit hole (there's plenty other ones :) ), someone (I) will eventually come around to reviewing it.

Thank you for that!

To clarify: I'm not a developer, not even much of a programmer; for me this contribution is more about learning than anything else. I'm glad to be of help, I would be extremely glad (and proud of myself) if this or any other PR i can possibly open gets merged. But basically what I do here is exploit the Syncthing project maintainers to facilitate my own growth as a programmer and a contributor by way of comments, suggestions, code review, ridicule of my efforts, or any other feedback I can get from you - and I can't stress enough how thankful I am for _any_ feedback, positive or otherwise. I totally understand that your objective here is not to help me learn, but to maintain an important project many people rely on. What I am doing is _by definition_ just wasting your time, basically. I do hope that the time you are wasting on me is somewhat justified by usefulness of my contribution (current and possibly future) to the project, but it's up to you to decide how much of your time it's worth. I trust all of you maintainers to manage your time to the benefit of Syncthing as a whole.
TL;DR: Whatever you do, in terms of priorities Synchting Project comes first, my PRs come last. ;)

calmh:
I think this is too many moving parts for the resulting benefit, especially the ini parser and code for writing those desktop files. There is another approach that I'm fleshing out, where we instead use the Go templating language, put some of the "logic" (format knowledge) in `.desktop.tpl` files, and have the translation system just able to understand that we can use template files. I think that will be less code, and the next time we run into an oddball file format that needs translation we already have all the machinery in place.

st-review:
:robot: *beep boop*

Issue #4271, opened 2017-07-25
Make use of filesystems that can handle copies efficiently. 


lemmi:
## Context

Copy-on-Write filesystems like btrfs, aswell as some network filesystems (smb, nfs, probably more, s3 has apis aswell,) can make copying more efficient with

1. only referencing data instead of copying (deduplication)
2. "Server side copies" (no need to roundtrip the data to the client and back to the server).

The first case will usually result in instant copies since it's only a metadata operation. The filesystem will take care that changes made to one copy won't affect the other. These are not to be confused with hardlinks!
The second case will benefit those that run syncthing on a network storage. The fileserver can be instructed to do the copy operation locally. Additionally the server can make use of the first case aswell.

Making use of deduplication (first case) has the potential to make file versioning very cheap in the sense that only new data will be written to disk. This is a common scenario if only metadata in image or music files is changed. Usually these types of files locate the data at the beginning of a file, then append the metadata (Exif, ID3 tags,...).

## Why?

I use staggered versioning and recently freed up about 100G of disk space by making use of this capabilities. There are tools like [duperemove][dupe] that can identify common datablocks and then rewrite the metadata of files to share common blocks. However this process involves scanning and hashing all files in question, comparing them and applying the metadata transformation afterwards. In my case this took several hours.

To make this work, almost every filesystem capable of such functionality needs to be instructed to do so.
Since syncthing already knows which blocks are shared by files and when files are copied, I'd like to make use of this knowledge right away.

## How?

For this to work seamlessly, it needs to be integrated in some form with the new Filesystem Interface #4228. This is currently WIP and I'd like to arrange for a compatible solution.
 
One option is to add another method that will mimic the syscall [copy_file_range(2)][CopyFileRange] to the Filesystem interface. But this will require every backend to actually implement this Method.

Another possibility is to have convenience function test for the capabilities and handle the fallback:

Add an interface filesystem backends should implement if they are capable of handling efficient copies:
```go
type FileRangeCopier interface {
    CopyFileRange(src File, srcoff int64, dst File, dstoff int64, length int) (n int, err error)
}
```
Then provide the function outside the Filesystem interface:
```go
package fs
// ...
func CopyFileRange(fs Filesystem, src File, srcoff int64, dst File, dstoff int64, length int64) (n int, err error) {
    // backend can do fast copy for us and do we want it?
    if frc, ok := fs.(FileRangeCopier); ok && cfg.Options.UseCopyFileRange {
[...]

PR #5544, opened 2019-02-16
lib/fs: Add fs.CopyRange

AudriusButkevicius:
Add in-kernel copy function that can be used with cow filesystems.

Ref #4271


lib/fs/util.go
```diff
@@ -15,7 +15,11 @@ import (
 	"strings"
 )
 
-var errNoHome = errors.New("no home directory found - set $HOME (or the platform equivalent)")
+var (
+	errNoHome = errors.New("no home directory found - set $HOME (or the platform equivalent)")
+	// Apparently BTRFS wants ioctl, XFS/NFS wants copy_file_range, so make this adjustable
+	copyOptimisations = getCopyOptimisations()
```
golangcibot: `copyOptimisations` is unused (from `deadcode`)

AudriusButkevicius:
This is now ready for review, this just adds the functionality for now, it's not used anywhere.

lib/fs/basicfs_copy_range_generic.go
```diff
@@ -0,0 +1,86 @@
...
+	if n, err := src.Seek(oldSrcOffset, io.SeekStart); err != nil {
+		return err
+	} else if n != oldSrcOffset {
+		return io.ErrUnexpectedEOF
+	}
+
+	if n, err := dst.Seek(oldDstOffset, io.SeekStart); err != nil {
+		return err
+	} else if n != oldDstOffset {
+		return io.ErrUnexpectedEOF
```
imsodin: Why unexpected EOF?

lib/fs/basicfs_copy_range.go
```diff
@@ -0,0 +1,106 @@
...
+	name string
+	impl copyRangeFuncGeneric
+}
+
+func registerCopyRangeImplementation(impl copyRangeImplementation) {
+	mut.Lock()
+	defer mut.Unlock()
+
+	for _, implName := range implementationOrder {
+		if implName == impl.name {
```
imsodin: I'd do an early return on `util.StringIndex(implementationOrder, impl.Name) == -1` instead.

AudriusButkevicius:
Yeah, bruteforcing is not ideal, and to be honest, in the simple case where this is most relevant, namely block finder, we read the data into userspace, verify the hash, and then write it.

Now, we read the data into userspace, verify it, in the best case succeed with a single syscall, but probably not, which then drops down to multiple syscalls and then the generic implementation that replaces a single write call with 4 x seek + 2 x stat + 1 x read + 1 x write etc.

We could "probe" each filesystem wether it supports it, but I so don't want to do this, and it can still fail badly when crossing symlinks and simillar garbage, requiring to fallback.

I guess in the most common cases this won't matter, in the block finder case, a few extra syscalls won't be that bad. @calmh?

Btw, this has a few bugs that are fixed in the ignore pr, which I need pull in here, so don't merge yet.

AudriusButkevicius:
```
root@you:~/st/lib/fs# STFSTESTPATH=/drives/btrfs/ go test -v -test.run=TestCopyRange
=== RUN   TestCopyRange
=== RUN   TestCopyRange/ioctl
=== RUN   TestCopyRange/ioctl/0_4_4_4_5_1_true
=== RUN   TestCopyRange/ioctl/0_4_0_0_5_1_true
=== RUN   TestCopyRange/ioctl/1_4_4_4_5_1_true
=== RUN   TestCopyRange/ioctl/0_0_4_4_4_1_true
=== RUN   TestCopyRange/ioctl/1_0_4_4_4_1_true
=== RUN   TestCopyRange/ioctl/0_8_4_4_9_1_true
=== RUN   TestCopyRange/ioctl/0_0_0_0_0_10_false
=== RUN   TestCopyRange/copy_file_range
=== RUN   TestCopyRange/copy_file_range/0_4_4_4_5_1_true
=== RUN   TestCopyRange/copy_file_range/0_4_0_0_5_1_true
=== RUN   TestCopyRange/copy_file_range/1_4_4_4_5_1_true
=== RUN   TestCopyRange/copy_file_range/0_0_4_4_4_1_true
=== RUN   TestCopyRange/copy_file_range/1_0_4_4_4_1_true
=== RUN   TestCopyRange/copy_file_range/0_8_4_4_9_1_true
=== RUN   TestCopyRange/copy_file_range/0_0_0_0_0_10_false
=== RUN   TestCopyRange/sendfile
=== RUN   TestCopyRange/sendfile/0_4_4_4_5_1_true
=== RUN   TestCopyRange/sendfile/0_4_0_0_5_1_true
=== RUN   TestCopyRange/sendfile/1_4_4_4_5_1_true
=== RUN   TestCopyRange/sendfile/0_0_4_4_4_1_true
=== RUN   TestCopyRange/sendfile/1_0_4_4_4_1_true
=== RUN   TestCopyRange/sendfile/0_8_4_4_9_1_true
=== RUN   TestCopyRange/sendfile/0_0_0_0_0_10_false
=== RUN   TestCopyRange/generic
=== RUN   TestCopyRange/generic/0_4_4_4_5_1_true
=== RUN   TestCopyRange/generic/0_4_0_0_5_1_true
=== RUN   TestCopyRange/generic/1_4_4_4_5_1_true
=== RUN   TestCopyRange/generic/0_0_4_4_4_1_true
=== RUN   TestCopyRange/generic/1_0_4_4_4_1_true
=== RUN   TestCopyRange/generic/0_8_4_4_9_1_true
=== RUN   TestCopyRange/generic/0_0_0_0_0_10_false
--- PASS: TestCopyRange (1.11s)
    --- PASS: TestCopyRange/ioctl (0.32s)
        --- PASS: TestCopyRange/ioctl/0_4_4_4_5_1_true (0.05s)
        --- PASS: TestCopyRange/ioctl/0_4_0_0_5_1_true (0.04s)
        --- PASS: TestCopyRange/ioctl/1_4_4_4_5_1_true (0.05s)
        --- PASS: TestCopyRange/ioctl/0_0_4_4_4_1_true (0.05s)
        --- PASS: TestCopyRange/ioctl/1_0_4_4_4_1_true (0.05s)
        --- PASS: TestCopyRange/ioctl/0_8_4_4_9_1_true (0.05s)
        --- PASS: TestCopyRange/ioctl/0_0_0_0_0_10_false (0.03s)
    --- PASS: TestCopyRange/copy_file_range (0.30s)
        --- PASS: TestCopyRange/copy_file_range/0_4_4_4_5_1_true (0.04s)
        --- PASS: TestCopyRange/copy_file_range/0_4_0_0_5_1_true (0.04s)
        --- PASS: TestCopyRange/copy_file_range/1_4_4_4_5_1_true (0.04s)
        --- PASS: TestCopyRange/copy_file_range/0_0_4_4_4_1_true (0.05s)
        --- PASS: TestCopyRange/copy_file_range/1_0_4_4_4_1_true (0.05s)
        --- PASS: TestCopyRange/copy_file_range/0_8_4_4_9_1_true (0.05s)
        --- PASS: TestCopyRange/copy_file_range/0_0_0_0_0_10_false (0.03s)
    --- PASS: TestCopyRange/sendfile (0.24s)
        --- PASS: TestCopyRange/sendfile/0_4_4_4_5_1_true (0.03s)
        --- PASS: TestCopyRange/sendfile/0_4_0_0_5_1_true (0.03s)
        --- PASS: TestCopyRange/sendfile/1_4_4_4_5_1_true (0.04s)
        --- PASS: TestCopyRange/sendfile/0_0_4_4_4_1_true (0.03s)
        --- PASS: TestCopyRange/sendfile/1_0_4_4_4_1_true (0.04s)
        --- PASS: TestCopyRange/sendfile/0_8_4_4_9_1_true (0.03s)
        --- PASS: TestCopyRange/sendfile/0_0_0_0_0_10_false (0.03s)
    --- PASS: TestCopyRange/generic (0.25s)
        --- PASS: TestCopyRange/generic/0_4_4_4_5_1_true (0.04s)
        --- PASS: TestCopyRange/generic/0_4_0_0_5_1_true (0.03s)
        --- PASS: TestCopyRange/generic/1_4_4_4_5_1_true (0.04s)
        --- PASS: TestCopyRange/generic/0_0_4_4_4_1_true (0.04s)
        --- PASS: TestCopyRange/generic/1_0_4_4_4_1_true (0.04s)
        --- PASS: TestCopyRange/generic/0_8_4_4_9_1_true (0.04s)
        --- PASS: TestCopyRange/generic/0_0_0_0_0_10_false (0.03s)
PASS
ok      github.com/syncthing/syncthing/lib/fs   1.115s
```

lib/fs/basicfs_copy_range_test.go
```diff
@@ -0,0 +1,238 @@
...
+					} else if fi.Size() != int64(testCase.expectedDstSize) {
+						t.Errorf("expected %d size, got %d", testCase.expectedDstSize, fi.Size())
+					}
+				})
+			}
+		})
+	}
+}
+
+func writeFull(w io.Writer, buf []byte) error {
```
calmh: Disk writes are always full writes so this utility function is unnecessary

AudriusButkevicius:
I think sendfile still has merit in the case of "RenameOrCopy" for versioning, given we now support versioning across devices.
I can imagine a single sendfile across devices being more efficient than multiple reads and writes.
Does this have a significant enough impact in terms of wall clock time for users, I don't know.

Also, sendfile matters for filesystems like nfs and cifs, where sendfile is a server side copy, potentially saving a roundtrip

For that reason, I'd leave sendfile where it is. Sure, in block finder it will be some sort of "false" optimisation as we have the data in userspace anyway, but I still think it's useful in terms of preserving a sane api (ala not have fs.CopyRange and fs.Sendfile etc).

Also, copy_file_range resorts to a kernel space copy in case the filesystem is now copy on write anyway.
There are flags you could give it which explicitly tells it to fail if it can't reuse blocks, but they aren't documented and I've only seen them in blog posts, so don't feel super comfortable using them.

Ferroin:
> Does this have a significant enough impact in terms of wall clock time for users, I don't know.

For at least Linux users, it should show an improvement for single files proportionate to the total amount of data being transferred, though it's usually negligible for files smaller than at least a few MB in size, and for really small files (dozens of kB in size), it can actually give _worse_ performance than read+write in a loop (there's more setup involved in sendfile, but once it's set up it's faster than read+write)

> Sure, in block finder it will be some sort of "false" optimisation as we have the data in userspace anyway, but I still think it's useful in terms of preserving a sane api (ala not have fs.CopyRange and fs.Sendfile etc).

Actually, it might be marginally faster here on Linux in some cases too, at least for the largest blocks.  If the data being written out is still in the page cache, then sendfile _should_ be faster for blocks past a certain size (I don't know what that size is though) because it will just reuse the data that's in the page cache instead of having to go out to disk for it (and in doing so will also avoid having to copy the data from userspace to kernel space).

AudriusButkevicius:
I retract my comments about sendfile being optimised for NFS and CIFS.
Having read the implementation in the kernel, it uses splice which essentially just uses pipes to move data between two files, so it's a in-kernel memory copy, but that's about it.

So the only argument for having is for large file copies when versioning files between different block devices (which is still a valid case to have it for).

calmh:
I didn’t think of pure copies made outside of the puller, so yeah that’s valid.

AudriusButkevicius:
Ok, so it seems we agree on keeping it around, I assume we are also ok to keep it in the block finder?

calmh:
Sounds reasonable that if we have the capability we should use it. Unless it causes overwhelming ugliness or pain for some reasons.

AudriusButkevicius:
Sure, just want to make sure we are all aware that it might come with a downside of being slower than a userspace copy in the fallback case due to all the "create pipe, wire them together, register with the io poller, etc" to copy 128k when no block reuse is possible.

AudriusButkevicius:
Also, no idea how this behaves on solaris on freebsd... Let's wait for maintains to scream I guess.

calmh:
Given these are just optimizations we could build tag them just for Linux. It seems odd to me that it would be slower to do it the optimized way than for us to do it in user space, but I don’t see it mattering too much either way. The value as i see it is in CoW/extent reuse, the rest is meh

imsodin:
Just a though: Make it a new filesystem, that is just basicFilesystem with CoW. I think users which use CoW filesystems should be knowledgeable enough to expect them to choose that optimized fs.

AudriusButkevicius:
I personally think this is fine as it is, if people start complaining, we can revisit.
I just wanted people to be aware of the edge cases.

Sendfile (for large cross filesystem copies) is posix compliant, so it should be there on Solaris/BSD.
It's there on Darwin but only for socket to file, not file to file.

AudriusButkevicius:
Anyways, I think this is good as it is, as this does not use the actual function anywhere.
We can argue about how it will be used iin the next PR once this lands.

calmh:
I think the fs layer is too low, it doesn’t really provide things like copying primitives. 

I’m more worried that due to bugs or race conditions we read and hash one set of data and then copy another, causing irreparable corruption.

AudriusButkevicius:
Again, the use of this is not in this PR, we can fist fight the details of the implementation in the next PR.

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/gogo/protobuf v1.2.0
 	github.com/golang/groupcache v0.0.0-20171101203131-84a468cf14b4
 	github.com/golang/snappy v0.0.0-20170215233205-553a64147049 // indirect
+	github.com/golang/sys v0.0.0-20190215142949-d0b11bdaac8a
```
calmh: Where is this from anyway? I guess this might be the physical location of `golang.org/x/sys` but it should not show up here as a dependency I think?

lib/fs/basicfs_copy_range_sendfile.go
```diff
@@ -0,0 +1,74 @@
...
+
+	for size > 0 {
+		// From the MAN page:
+		//
+		// If offset is not NULL, then it points to a variable holding the file offset from which sendfile() will start
+		// reading data from in_fd. When sendfile() returns, this variable will be set to the offset of the byte
+		// following the last byte that was read. If offset is not NULL, then sendfile() does not modify the current
+		// file offset of in_fd; otherwise the current file offset is adjusted to reflect the number of bytes read from
+		// in_fd.
+		n, err := syscall.Sendfile(int(dst.Fd()), int(src.Fd()), &srcOffset, int(size))
```
calmh: You're not doing anything with srcOffset here. Shouldn't that be incremented similar to how you seek for dstOffset?

lib/fs/basicfs_copy_range_copyfilerange.go
```diff
@@ -0,0 +1,44 @@
...
+		// 	The file offset of fd_in is not changed, but off_in is adjusted
+		// appropriately.
+		n, err := unix.CopyFileRange(int(src.Fd()), &srcOffset, int(dst.Fd()), &dstOffset, int(size), 0)
+		if n == 0 && err == nil {
+			return io.ErrUnexpectedEOF
+		}
+		if err != nil && err != syscall.EAGAIN {
+			return err
+		}
+		dstOffset += int64(n)
```
calmh: > The file offset of fd_in is not changed, but off_in is adjusted appropriately

I guess you take this to mean that srcOffset is automatically increased, even in the case of error, but dstOffset is not? Are you sure about that, on both accounts?

lib/fs/basicfs_copy_range.go
```diff
@@ -0,0 +1,93 @@
...
+
+// CopyRange tries to use the most optimal way to copy data between two files.
+// Takes size bytes at offset srcOffset from the source file, and copies the data to destination file at offset
+// dstOffset. If required, adjusts the size of the destination file to fit that much data.
+//
+// On Linux/BSD it tries to use ioctl and copy_file_range system calls, which if the underlying filesystem supports it
+// tries referencing existing data in the source file, instead of making a copy and taking up additional space.
+//
+// If that is not possible, the data will be copied using an in-kernel copy (copy_file_range fallback, sendfile),
+// oppose to user space copy, if those system calls are available and supported for the source and target in question.
```
calmh: Why not fall back to a regular copy in the end? I get that there's no code to use this yet, but we'd have to do the fallback there instead otherwise.

lib/fs/basicfs_copy_range_sendfile.go
AudriusButkevicius: from the comment above
```
                // If offset is not NULL, then it points to a variable holding the file offset from which sendfile() will start
		// reading data from in_fd. When sendfile() returns, this variable will be set to the offset of the byte
		// following the last byte that was read.
```

lib/fs/basicfs_copy_range_copyfilerange.go
AudriusButkevicius: 🤷‍♂  useful content in the man page:
```
The following semantics apply for off_in, and similar statements
       apply to off_out:
```

how simillar exactly? :suspect:

lib/fs/basicfs_copy_range.go
AudriusButkevicius: I thought this is exactly what we did not want to do, and fail if these fancy things don't work and just write the verified block from ram, oppose to reading it again and writing it back, triggering a race?

Versioner et-al can catch this failure and do the copy themselves, block finder can just write stuff it has in ram.

In practice, I don't think this will ever fail, unless windows/solaris/exotics.

lib/fs/basicfs_copy_range_copyfilerange.go
AudriusButkevicius: Apparently you are right, assertion of the dstOffset after write changing verifies that.

I can't really write a test for it, as, I can't force unix.CopyRange to return prematurely to catch writes happening in the wrong place.

go.mod
AudriusButkevicius: Idk how that happened, re-gogetting fixed it.

AudriusButkevicius:
Good to go? I want to start arguing about the use of this.

calmh:
Hold your horses. This is potentially the most dangerous PR ever in terms of not fucking up people's data and there is zero margin for doing the wrong thing in any corner case ever so I'm going to be all over this with a microscope for days before any merging happens.

calmh:
You can still do that though. Create it as a PR to be merged with this branch and we can argue about it as much as we like in the meantime.

AudriusButkevicius:
This specific PR is harmless in itself as stuff is not used anywhere, but yeah, I get the sentiment, just want to explore the use of this, as I suspect it might change what this PR itself looks like.

calmh:
I sympathise, but we're probably not going to re-test and re-review if we merge this, so might as well do the extra legwork now. But especially if the "use it" part of the code might change the stuff here, maybe just add it to the same PR? The different parts of it are relatively self contained, I don't think it makes it much harder to review really.

lib/fs/basicfs_copy_range_ioctl.go
```diff
@@ -0,0 +1,54 @@
...
+};
+*/
+type fileCloneRange struct {
+	srcFd     int64
+	srcOffset uint64
+	srcLength uint64
+	dstOffset uint64
+}
+
+func copyRangeIoctl(src, dst basicFile, srcOffset, dstOffset, size int64) error {
```
calmh: One thing to note (somewhere) about this one is that the write must be disk block aligned in both offset and size. Luckily, we generally deal with blocks that are larger than the disk block size so this is mostly fine. But it means we can't use this to copy an entire file, unless that file happens to be an even number of blocks long. For example, this fails pretty much all the tests:

```
diff --git a/lib/fs/basicfs_copy_range_test.go b/lib/fs/basicfs_copy_range_test.go
index a75c2655..f23d6390 100644
--- a/lib/fs/basicfs_copy_range_test.go
+++ b/lib/fs/basicfs_copy_range_test.go
@@ -18,7 +18,7 @@ import (
 )

 var (
-       generationSize  int64 = 4 << 20
+       generationSize  int64 = 4 << 20 + 1
        defaultCopySize int64 = 1 << 20
```

calmh:
Thinking about this a little, shouldn't this be split out into two separate exported functions as there are two separate optimizations we're aiming for?

- In the regular block copier (puller), we have just read and hashed a block. We want to write it to the destination. A possible optimization is re-using the existing on disk block for better space utilization and less write I/O. Only `ioctl_ficlonerange` provides this for sure. The alternative `copy_file_range` *might* provide it but might just be another read+write. `sendfile` is just a read+write combo which is going to be strictly worse than just writing the block we already have.

- When doing file copies (versioner) we don't really care and can use either method. Except, `ioctl_ficlonerange` requires extra care for handling the last block as I understand it.

Are there cases where `ioctl_ficlonerange` does good stuff and `copy_file_range` does not? They same to be available from the same kernel version. The latter at least seems to handle odd size reads/writes for us. 

So would two separate operations be reasonable?

1) Something based on `copy_file_range` only, used for when we want block reuse. This fails, we just do the regular write.
2) Something that does `copy_file_range` + `sendfile` + regular `io.Copy` fallback order, for when we just want to shuffle data from A to B

AudriusButkevicius:
I am somewhat starting to agree with you.

>When doing file copies (versioner) we don't really care and can use either method. Except, ioctl_ficlonerange requires extra care for handling the last block as I understand it.

ioctl_ficlonerange manpage says:
```
If src_length is zero, the ioctl reflinks to the end of the source file.
```
Implying that this is handled. I suspect the filesystem still reserves a full block even if it's not full, in which case reflinking is still ok.


>  They same to be available from the same kernel version.

That is not true, copy_file_range was added in 4.5, BTRFS_IOC_CLONE_RANGE existed way before that in 2.6.29 (released March 2009 according to https://github.com/torvalds/linux/commit/c5c9cd4d1b827fe545ed2a945e91e3a6909f3886#diff-11eb66b8278b195d1d754021c8283a3dR1025), but it was btrfs specific and was "generalised" in 4.5 which then became FI_CLONERANGE.

It seems copy_file_range does take flags, which I've checked btrfs and xfs support, essentially allowing to say reflink only, otherwise fail, but sadly it hasn't yet been pushed through all the way up: https://github.com/torvalds/linux/blob/master/fs/read_write.c#L1582


I propse the following:
For whole file copy, ioctl, copy_file_range, sendfile, io.Copy, first one to succeed wins.
For block finder:
1. ioctl, if it succeeds, re-verify by reading back to avoid races?
2. copy_file_range, if it succeeds, re-verify by reading back to avoid races?
3. fd.Write(buf)


We could also add support for https://manpages.debian.org/testing/manpages-dev/ioctl_ficlone.2.en.html
which is full file clone support that has been there for a while and potentially might be supported by other filesystems.

calmh:
So about the zero length thing, that would need special handling at the call site. We do block by block so we’d need to know that the last block it the last block, or we just skip the optimization when the length isn’t an even multiple of 128k...

About the version introduced, yeah, it was just the _name_ ioctl_ficlone that was added in 4.5. That said, old btrfs are scary buggy (like eat your data for lunch buggy), hopefully newer is better, so we even care about people pre-4.5 for a simple optimization? I’d prefer to optimize for implementation clarity and copy-file-range wins there...

AudriusButkevicius:
> So about the zero length thing, that would need special handling at the call site. We do block by block so we’d need to know that the last block is the last block, or we just skip the optimization when the length isn’t an even multiple of 128k...

We can still call ioctl, it will just fail, and we'll fallback to other things.

> About the version introduced, yeah, it was just the name ioctl_ficlone that was added in 4.5. That said, old btrfs are scary buggy (like eat your data for lunch buggy), hopefully newer is better, so do we even care about people pre-4.5 for a simple optimization? I’d prefer to optimize for implementation clarity and copy-file-range wins there...

I guess 4.5 is fairly "bleeding edge" so I suspect very little people will pick this up. On the other hand I agree, if you are running btrfs, you are pretty much a maniac anyway, and might as well move to the bleeding kernel to support this.



Also, I just noticed that Darwin/BSD has sys_clonefileat which we might want to add support for, yet I'll have no means of testing this.

AudriusButkevicius:
So what do we want to do with this?
io_ficlonerange seems like it's a thing, so completely not supporting it seems weird.
cp --reflink uses that to this day, and xfs_io (http://man7.org/linux/man-pages/man8/xfs_io.8.html) declares the two operations as separate calls.

AudriusButkevicius:
We should also probably support FI_CLONE for full file copies

AudriusButkevicius:
Bhuumphh

calmh:
Well, I’m old and conservative and scared of change. This does all kind of stuff that scares me in order to optimize for, at best, a minority. I suggest the following to begin with:

- Pick of one these implementations, whichever feels most mainstream or useful. We ship that one. When the data corruption issues hit we can at least skip the “uh, wonder which of the 37 bruteforced syscalls was the problem” step. 

- Hide the magic behind an env var (feature flag). Flag not set, usual code path only. 

- Pick a champion user who wants this, runs it in production for a month, and reports on the awesomeness of it. 

Then we add more implementations and phase out the feature flag...

Ferroin:
If you need someone to help test this in production usage, I can cover that for at least the case of it running on top of BTRFS on Linux, though I don't know which of the paths that might exercise.

AudriusButkevicius:
Sure, I can do adhoc tests too, this is more about a potential sacrifice of your data to bad code, and running it in prod.

Ferroin:
@AudriusButkevicius I'm talking about running it in production type usage (not exactly rigorous production usage, but well within normal desktop user usage), I just mentioned that it would only cover testing on BTRFS on Linux because that's the only setup I have where this code would get exercised at all (I could probably get it tested on XFS too, but that would still leave at least ZFS uncovered, and possibly other filesystems).

As far as the risk of data loss, that wouldn't really be an issue for me, all the data I manage with Syncthing is backed up through other means as well.

st-review:
:robot: *beep boop*

AudriusButkevicius:
Nah

Atera:
Hmm, since I was one of the ones who mentioned really having a lot of use for this, I would also be happy to test. I also run BTRFS on Debian Linux, but might be able to test on ZFS on Debian as well before the month is over. If you give me instructions on how to test and report back I would be happy to oblige.

st-review:
:robot: *beep boop*

Atera:
Mmm, perhaps I missed it elsewhere, but was this code/work from last year incorporated into Syncthing already?

AudriusButkevicius:
It was not.

Atera:
Has any testing been done, or do you need any testing done? I hadn't said anything till now since it seemed things were on the right track some months ago, but it is a feature that would save me terabytes of disk on my servers, so I do care a fair bit about it especially if the snags at the moment are just testing and reporting which I could help with. Thanks again for writing the relevant code, Audrius.

AudriusButkevicius:
So now CopyRange takes a method for the type of copy method to use.
I intend to add a folder parameter that specifies the method to use, defaulting to standard that just does io.CopyN.

Again, this is just a PR that adds methods, it does not use them anywhere yet.

Some comments
1. Seems only ioctl is quirky and requires offsets to be fs block size aligned. I can't think of where we'd use offsets that are not fs block size aligned (weak hash stuff maybe, but we can probably skip using this that all together for that, as it's the same file)
2. Sizes do not need to be fs block size aligned, meaning "last block issue" is not an actual issue as last block would start on a nice offset anyway.
2. Added ioctl optimisation for full file copies.

Tests still pass on XFS and BTRFS:
```
root@you:~/st/lib/fs# STFSTESTPATH=/mnt/xfs go test -v -test.run=TestCopyRange
--- PASS: TestCopyRange (1.95s)
    --- PASS: TestCopyRange/all (0.38s)
        --- PASS: TestCopyRange/all/append_to_end (0.04s)
        --- PASS: TestCopyRange/all/append_to_end,_offsets_at_start (0.04s)
        --- PASS: TestCopyRange/all/overwrite_part_of_destination_region (0.04s)
        --- PASS: TestCopyRange/all/overwrite_all_of_destination (0.03s)
        --- PASS: TestCopyRange/all/overwrite_part_of_destination (0.03s)
        --- PASS: TestCopyRange/all/destination_gets_expanded_as_it_is_being_written_to (0.05s)
        --- PASS: TestCopyRange/all/source_file_too_small (0.05s)
        --- PASS: TestCopyRange/all/not_block_aligned_write (0.03s)
        --- PASS: TestCopyRange/all/last_block (0.02s)
        --- PASS: TestCopyRange/all/whole_file_copy_block_aligned (0.02s)
        --- PASS: TestCopyRange/all/whole_file_copy_not_block_aligned (0.03s)
    --- PASS: TestCopyRange/copy_file_range (0.37s)
        --- PASS: TestCopyRange/copy_file_range/append_to_end (0.03s)
        --- PASS: TestCopyRange/copy_file_range/append_to_end,_offsets_at_start (0.03s)
        --- PASS: TestCopyRange/copy_file_range/overwrite_part_of_destination_region (0.04s)
        --- PASS: TestCopyRange/copy_file_range/overwrite_all_of_destination (0.03s)
        --- PASS: TestCopyRange/copy_file_range/overwrite_part_of_destination (0.05s)
        --- PASS: TestCopyRange/copy_file_range/destination_gets_expanded_as_it_is_being_written_to (0.06s)
        --- PASS: TestCopyRange/copy_file_range/source_file_too_small (0.03s)
        --- PASS: TestCopyRange/copy_file_range/not_block_aligned_write (0.02s)
        --- PASS: TestCopyRange/copy_file_range/last_block (0.02s)
        --- PASS: TestCopyRange/copy_file_range/whole_file_copy_block_aligned (0.02s)
        --- PASS: TestCopyRange/copy_file_range/whole_file_copy_not_block_aligned (0.02s)
    --- PASS: TestCopyRange/ioctl (0.29s)
        --- SKIP: TestCopyRange/ioctl/append_to_end (0.03s)
        --- SKIP: TestCopyRange/ioctl/append_to_end,_offsets_at_start (0.03s)
        --- SKIP: TestCopyRange/ioctl/overwrite_part_of_destination_region (0.03s)
        --- SKIP: TestCopyRange/ioctl/overwrite_all_of_destination (0.03s)
        --- SKIP: TestCopyRange/ioctl/overwrite_part_of_destination (0.03s)
        --- SKIP: TestCopyRange/ioctl/destination_gets_expanded_as_it_is_being_written_to (0.03s)
        --- PASS: TestCopyRange/ioctl/source_file_too_small (0.03s)
        --- SKIP: TestCopyRange/ioctl/not_block_aligned_write (0.02s)
        --- SKIP: TestCopyRange/ioctl/last_block (0.02s)
        --- SKIP: TestCopyRange/ioctl/whole_file_copy_block_aligned (0.02s)
        --- SKIP: TestCopyRange/ioctl/whole_file_copy_not_block_aligned (0.02s)
    --- PASS: TestCopyRange/sendfile (0.47s)
        --- PASS: TestCopyRange/sendfile/append_to_end (0.05s)
        --- PASS: TestCopyRange/sendfile/append_to_end,_offsets_at_start (0.04s)
        --- PASS: TestCopyRange/sendfile/overwrite_part_of_destination_region (0.04s)
        --- PASS: TestCopyRange/sendfile/overwrite_all_of_destination (0.03s)
        --- PASS: TestCopyRange/sendfile/overwrite_part_of_destination (0.03s)
        --- PASS: TestCopyRange/sendfile/destination_gets_expanded_as_it_is_being_written_to (0.05s)
        --- PASS: TestCopyRange/sendfile/source_file_too_small (0.04s)
        --- PASS: TestCopyRange/sendfile/not_block_aligned_write (0.06s)
        --- PASS: TestCopyRange/sendfile/last_block (0.04s)
        --- PASS: TestCopyRange/sendfile/whole_file_copy_block_aligned (0.05s)
        --- PASS: TestCopyRange/sendfile/whole_file_copy_not_block_aligned (0.03s)
    --- PASS: TestCopyRange/standard (0.45s)
        --- PASS: TestCopyRange/standard/append_to_end (0.05s)
        --- PASS: TestCopyRange/standard/append_to_end,_offsets_at_start (0.07s)
        --- PASS: TestCopyRange/standard/overwrite_part_of_destination_region (0.05s)
        --- PASS: TestCopyRange/standard/overwrite_all_of_destination (0.04s)
        --- PASS: TestCopyRange/standard/overwrite_part_of_destination (0.04s)
        --- PASS: TestCopyRange/standard/destination_gets_expanded_as_it_is_being_written_to (0.06s)
        --- PASS: TestCopyRange/standard/source_file_too_small (0.05s)
        --- PASS: TestCopyRange/standard/not_block_aligned_write (0.03s)
        --- PASS: TestCopyRange/standard/last_block (0.02s)
        --- PASS: TestCopyRange/standard/whole_file_copy_block_aligned (0.03s)
        --- PASS: TestCopyRange/standard/whole_file_copy_not_block_aligned (0.02s)
PASS
ok      github.com/syncthing/syncthing/lib/fs   1.966s
```

btrfs
```
root@you:~/st/lib/fs# STFSTESTPATH=/mnt/btrfs go test -v -test.run=TestCopyRange
--- PASS: TestCopyRange (3.67s)
    --- PASS: TestCopyRange/standard (0.38s)
        --- PASS: TestCopyRange/standard/append_to_end (0.05s)
        --- PASS: TestCopyRange/standard/append_to_end,_offsets_at_start (0.04s)
        --- PASS: TestCopyRange/standard/overwrite_part_of_destination_region (0.04s)
        --- PASS: TestCopyRange/standard/overwrite_all_of_destination (0.04s)
        --- PASS: TestCopyRange/standard/overwrite_part_of_destination (0.04s)
        --- PASS: TestCopyRange/standard/destination_gets_expanded_as_it_is_being_written_to (0.05s)
        --- PASS: TestCopyRange/standard/source_file_too_small (0.04s)
        --- PASS: TestCopyRange/standard/not_block_aligned_write (0.03s)
        --- PASS: TestCopyRange/standard/last_block (0.02s)
        --- PASS: TestCopyRange/standard/whole_file_copy_block_aligned (0.03s)
        --- PASS: TestCopyRange/standard/whole_file_copy_not_block_aligned (0.03s)
    --- PASS: TestCopyRange/all (0.82s)
        --- PASS: TestCopyRange/all/append_to_end (0.05s)
        --- PASS: TestCopyRange/all/append_to_end,_offsets_at_start (0.05s)
        --- PASS: TestCopyRange/all/overwrite_part_of_destination_region (0.05s)
        --- PASS: TestCopyRange/all/overwrite_all_of_destination (0.12s)
        --- PASS: TestCopyRange/all/overwrite_part_of_destination (0.10s)
        --- PASS: TestCopyRange/all/destination_gets_expanded_as_it_is_being_written_to (0.11s)
        --- PASS: TestCopyRange/all/source_file_too_small (0.12s)
        --- PASS: TestCopyRange/all/not_block_aligned_write (0.06s)
        --- PASS: TestCopyRange/all/last_block (0.04s)
        --- PASS: TestCopyRange/all/whole_file_copy_block_aligned (0.06s)
        --- PASS: TestCopyRange/all/whole_file_copy_not_block_aligned (0.06s)
    --- PASS: TestCopyRange/copy_file_range (1.02s)
        --- PASS: TestCopyRange/copy_file_range/append_to_end (0.13s)
        --- PASS: TestCopyRange/copy_file_range/append_to_end,_offsets_at_start (0.10s)
        --- PASS: TestCopyRange/copy_file_range/overwrite_part_of_destination_region (0.06s)
        --- PASS: TestCopyRange/copy_file_range/overwrite_all_of_destination (0.07s)
        --- PASS: TestCopyRange/copy_file_range/overwrite_part_of_destination (0.10s)
        --- PASS: TestCopyRange/copy_file_range/destination_gets_expanded_as_it_is_being_written_to (0.11s)
        --- PASS: TestCopyRange/copy_file_range/source_file_too_small (0.18s)
        --- PASS: TestCopyRange/copy_file_range/not_block_aligned_write (0.08s)
        --- PASS: TestCopyRange/copy_file_range/last_block (0.04s)
        --- PASS: TestCopyRange/copy_file_range/whole_file_copy_block_aligned (0.08s)
        --- PASS: TestCopyRange/copy_file_range/whole_file_copy_not_block_aligned (0.06s)
    --- PASS: TestCopyRange/ioctl (0.69s)
        --- PASS: TestCopyRange/ioctl/append_to_end (0.09s)
        --- PASS: TestCopyRange/ioctl/append_to_end,_offsets_at_start (0.09s)
        --- PASS: TestCopyRange/ioctl/overwrite_part_of_destination_region (0.07s)
        --- PASS: TestCopyRange/ioctl/overwrite_all_of_destination (0.10s)
        --- PASS: TestCopyRange/ioctl/overwrite_part_of_destination (0.10s)
        --- PASS: TestCopyRange/ioctl/destination_gets_expanded_as_it_is_being_written_to (0.08s)
        --- PASS: TestCopyRange/ioctl/source_file_too_small (0.03s)
        --- PASS: TestCopyRange/ioctl/not_block_aligned_write (0.02s)
        --- PASS: TestCopyRange/ioctl/last_block (0.02s)
        --- PASS: TestCopyRange/ioctl/whole_file_copy_block_aligned (0.03s)
        --- PASS: TestCopyRange/ioctl/whole_file_copy_not_block_aligned (0.05s)
    --- PASS: TestCopyRange/sendfile (0.75s)
        --- PASS: TestCopyRange/sendfile/append_to_end (0.08s)
        --- PASS: TestCopyRange/sendfile/append_to_end,_offsets_at_start (0.09s)
        --- PASS: TestCopyRange/sendfile/overwrite_part_of_destination_region (0.07s)
        --- PASS: TestCopyRange/sendfile/overwrite_all_of_destination (0.06s)
        --- PASS: TestCopyRange/sendfile/overwrite_part_of_destination (0.07s)
        --- PASS: TestCopyRange/sendfile/destination_gets_expanded_as_it_is_being_written_to (0.09s)
        --- PASS: TestCopyRange/sendfile/source_file_too_small (0.07s)
        --- PASS: TestCopyRange/sendfile/not_block_aligned_write (0.06s)
        --- PASS: TestCopyRange/sendfile/last_block (0.04s)
        --- PASS: TestCopyRange/sendfile/whole_file_copy_block_aligned (0.05s)
        --- PASS: TestCopyRange/sendfile/whole_file_copy_not_block_aligned (0.07s)
PASS
ok      github.com/syncthing/syncthing/lib/fs   3.689s
```

AudriusButkevicius:
@calmh @imsodin bhuuump

imsodin:
There's some comments talking about which options are chosen when, however it doesn't look like anything like that is done yet. Is the final idea that there's no configuration and Syncthing will just use whatever is most efficient and works? And for a first step now there's activation by e.g. env vars to make it testable?

AudriusButkevicius:
There is literally nothing at the moment that would use this.
This is dead code for now.  We can bike shed how this gets used in the next PR.

imsodin:
What's the point of separating it? Feels weird to merge dead code into master.

AudriusButkevicius:
That was @calmh's request, roughly 1.5 years ago, to simplify the review process and make it less scary.

AudriusButkevicius:
Actually, checking the comments, the opposite was requested, so I'll open another PR adding usage of this.

calmh:
Superseded by #6746? Reopen if I missed some subtlety

Issue #5480, opened 2019-01-24
Docker: Exposed Volumes


Dargmuesli:
The [README-Docker.md](https://github.com/syncthing/syncthing/blob/master/README-Docker.md) references two exposed volumes. But the Dockerfile does not expose the first (`/var/syncthing/config`). Shouldn't the README get corrected?

calmh:
Probably. Only /var/syncthing is a volume by default, although the user can of course map as they wish

Dargmuesli:
But if they do thay receive a `chmod` error. Certificates cannot be generated, ... I just tried and wondered for an hour why it isn't working. Volumes inside default volumes seem not to work.

jordanpotter:
I attempted to create a data volume for `/var/syncthing/config` and came across the same issue. Ubuntu 18.10 running Docker 18.09.2

calmh:
I don't think this has anything to do with volumes as such. If you *do* use extra modules you need to make sure Syncthing has write permissions on them. If it doesn't, and cannot create certificates in its config dir, then it obviously cannot work.

PR #5545, opened 2019-02-17
docs: Correct Docker README (#5480)

Dargmuesli:
Plus some formatting.
If you rather wish to add a `VOLUME` tag to the Dockerfile, feel free to close this alternative solution.

### DIFF
--- a/README-Docker.md
+++ b/README-Docker.md
@@ -1,39 +1,30 @@
 # Docker Container for Syncthing
 
 Use the Dockerfile in this repo, or pull the `syncthing/syncthing` image
-from Docker Hub. Use volumes to have the synchronized files available on the
-host.
+from Docker Hub.
 
-The exposed volumes are by default:
-
-    /var/syncthing/config   - the configuration and index directory into the Container
-    /var/syncthing          - the default sync folder into the Container
-
-You can add more folders and map them as you prefer.
+Use the `/var/syncthing` volume to have the synchronized files available on the
+host. You can add more folders and map them as you prefer.
 
 Note that Syncthing runs as UID 1000 and GID 1000 by default. These may be
 altered with the ``PUID`` and ``PGID`` environment variables.
 
-Example usage:
+## Example Usage
 
 ```
 $ docker pull syncthing/syncthing
 $ docker run -p 8384:8384 -p 22000:22000 \
-    -v /wherever/st-cfg:/var/syncthing/config \
     -v /wherever/st-sync:/var/syncthing \
     syncthing/syncthing:latest
 ```
 
-Note that local device discovery will not work with the above command resulting
-in poor local transfer rates if local device addresses are not manually
-configured.
+Note that local device discovery will not work with the above command, resulting in poor local transfer rates if local device addresses are not manually configured.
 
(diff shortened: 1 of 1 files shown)


jordanpotter:
Thoughts on actually making `/var/syncthing/config` a mount point in the `Dockerfile` instead? :thinking:

Dargmuesli:
I'm not sure if volumes inside volumes work without issues. I suspect that that they don't, leading to #5480.

AudriusButkevicius:
How is your comment relevant tto this PR tho?

Dargmuesli:
Which comment are you referring to?

jordanpotter:
I think he was referring to my comment? If so, I was asking about an alternative solution also related to https://github.com/syncthing/syncthing/issues/5480

@Dargmuesli really good point. My thinking was the following:

As a consumer of the Docker image and a user of data volumes semi-religiously, I would want to create a data volume for `/var/syncthing/config` to store all the configuration files. I'm not sure why I would want a data volume for `/var/syncthing`, which also stores the default sync folder.

Dargmuesli:
True. I just `exec`ed into my container and found that `/var/syncthing/` only contains the `config` folder. Can someone confirm that `/var/syncthing/` really doesn't contain anything else?

jordanpotter:
Looks like it also contains the default sync folder. From inside a fresh container:
```
# find /var/syncthing/
/var/syncthing/
/var/syncthing/config
/var/syncthing/config/https-cert.pem
/var/syncthing/config/https-key.pem
/var/syncthing/config/key.pem
/var/syncthing/config/index-v0.14.0.db
/var/syncthing/config/index-v0.14.0.db/MANIFEST-000017
/var/syncthing/config/index-v0.14.0.db/LOCK
/var/syncthing/config/index-v0.14.0.db/CURRENT
/var/syncthing/config/index-v0.14.0.db/000015.ldb
/var/syncthing/config/index-v0.14.0.db/000016.log
/var/syncthing/config/index-v0.14.0.db/000014.ldb
/var/syncthing/config/index-v0.14.0.db/LOG
/var/syncthing/config/config.xml
/var/syncthing/config/cert.pem
/var/syncthing/Sync
/var/syncthing/Sync/.stfolder
```

jordanpotter:
What if the mount points were instead `/var/syncthing/config` and `/var/syncthing/Sync`? That way the configuration could be in a separate volume from the data.

AudriusButkevicius:
And break every existing user in the world.

AudriusButkevicius:
Why can't you just mend the config in the same mount point?
Anyways, this is severly off topic and not related to this PR so should be moved elsewhere (forum ideally).

jordanpotter:
> And break every existing user in the world.

That's not completely accurate. The official Docker `syncthing/syncthing` repo has tags available for this reason.

> Why can't you just mend the config in the same mount point?

I feel the spirit of Separation of Concerns would say config and data should be separate. But I 100% see your point.

> Anyways, this is severly off topic and not related to this PR so should be moved elsewhere (forum ideally).

Sounds good! I'll start a conversation there. Thank you :+1:

PR #5546, opened 2019-02-18
lib/fs: Rename fsFile* to basicFile*

imsodin:
The code for `fsFileInfo` was separated over two files without apparent reason. Also it is `basicFilesystem` specific, as well as `fsFile`, so `basicFile` seems more approriate (as well as the "basic_" filename prefix.


AudriusButkevicius:
I wanted to go the other way and rename basicFs to localFs or something like that and then have localFile

imsodin:
Yeah that sounds even better, but personally I wouldn't consider the potential (unlikely) problems of changing a public element as a bigger downside compared to the naming. Anyway I think this is a separate issue, this PR just makes the current situation consistent without renaming anything important. However if you want to tackle that right away, this PR can obviously be closed.

AudriusButkevicius:
I know exactly the gif I am looking for but I can't find it:

(Kid running and calling for his dad)

@calmh ?

imsodin:
Actually what I meant to say is that I am fine with whatever you want to roll with, it's your baby. The only thing that bugs me and needs fixing is the coexistence of `fsfileinfo_*.go` and `fileinfo_*.go` with methods on the same type. When merging them I noticed the `fs` vs `basic` discrepancy and "fixed" that too on the go, but that's just an afterthought.

PR #5556, opened 2019-02-23
lib/scanner: Use standard adler32 when we don't need rolling

AudriusButkevicius:
Seems the rolling adler32 implementation is super slow when executed on large blocks, even tho I can't explain why. Use standard adler when we don't need to roll
```
BenchmarkFind1MFile-16    	     	     	     	     100	  18739828 ns/op	  55.95 MB/s	  398924 B/op	      20 allocs/op
BenchmarkBlock/adler32-131072/#00-16     	     	     200	   9678923 ns/op	1083.36 MB/s	 2654929 B/op	     163 allocs/op
BenchmarkBlock/bozo32-131072/#00-16      	     	      20	  73577340 ns/op	 142.51 MB/s	 2654928 B/op	     163 allocs/op
BenchmarkBlock/buzhash32-131072/#00-16   	     	      20	  61456805 ns/op	 170.62 MB/s	 2654928 B/op	     163 allocs/op
BenchmarkBlock/buzhash64-131072/#00-16   	     	      20	  61651235 ns/op	 170.08 MB/s	 2654928 B/op	     163 allocs/op
BenchmarkBlock/vanilla-adler32-131072/#00-16         	     300	   4357369 ns/op	2406.44 MB/s	 2654929 B/op	     163 allocs/op
BenchmarkBlock/adler32-16777216/#00-16               	       2	 545514500 ns/op	  19.22 MB/s	   65624 B/op	       5 allocs/op
BenchmarkBlock/bozo32-16777216/#00-16                	       1	4683325500 ns/op	   2.24 MB/s	51970144 B/op	      24 allocs/op
BenchmarkBlock/buzhash32-16777216/#00-16             	       1	3880333000 ns/op	   2.70 MB/s	51970144 B/op	      24 allocs/op
BenchmarkBlock/buzhash64-16777216/#00-16             	       1	3878978300 ns/op	   2.70 MB/s	51970144 B/op	      24 allocs/op
BenchmarkBlock/vanilla-adler32-16777216/#00-16       	     300	   4011820 ns/op	2613.72 MB/s	   65624 B/op	       5 allocs/op
BenchmarkRoll/adler32-131072/#00-16                  	    2000	    949645 ns/op	 138.02 MB/s	       0 B/op	       0 allocs/op
BenchmarkRoll/bozo32-131072/#00-16                   	    2000	    796951 ns/op	 164.47 MB/s	       0 B/op	       0 allocs/op
BenchmarkRoll/buzhash32-131072/#00-16                	    2000	    860292 ns/op	 152.36 MB/s	       0 B/op	       0 allocs/op
BenchmarkRoll/buzhash64-131072/#00-16                	    2000	    882446 ns/op	 148.53 MB/s	       0 B/op	       0 allocs/op
BenchmarkRoll/adler32-16777216/#00-16                	      10	 121974010 ns/op	 137.55 MB/s	       0 B/op	       0 allocs/op
BenchmarkRoll/bozo32-16777216/#00-16                 	      10	 106611300 ns/op	 157.37 MB/s	       0 B/op	       0 allocs/op
BenchmarkRoll/buzhash32-16777216/#00-16              	      10	 116323520 ns/op	 144.23 MB/s	       0 B/op	       0 allocs/op
BenchmarkRoll/buzhash64-16777216/#00-16              	      10	 114303920 ns/op	 146.78 MB/s	       0 B/op	       0 allocs/op
```

CC @chmduquesne

### DIFF
--- a/lib/scanner/blocks.go
+++ b/lib/scanner/blocks.go
@@ -10,9 +10,9 @@ import (
 	"bytes"
 	"context"
 	"hash"
+	"hash/adler32"
 	"io"
 
-	"github.com/chmduquesne/rollinghash/adler32"
 	"github.com/syncthing/syncthing/lib/protocol"
 	"github.com/syncthing/syncthing/lib/sha256"
 )
--- a/lib/weakhash/benchmark_test.go
+++ b/lib/weakhash/benchmark_test.go
@@ -71,12 +71,12 @@ func BenchmarkBlock(b *testing.B) {
 
 	sizes := []int64{128 << 10, 16 << 20}
 
-	buf := make([]byte, 10 << 20)
+	buf := make([]byte, 10<<20)
 	rand.Read(buf)
 
 	for _, testSize := range sizes {
 		for _, test := range tests {
-			b.Run(test.name + "-" + fmt.Sprint(testSize), func(bb *testing.B) {
+			b.Run(test.name+"-"+fmt.Sprint(testSize), func(bb *testing.B) {
 				bb.Run("", func(bbb *testing.B) {
 					bbb.ResetTimer()
 					for i := 0; i < bbb.N; i++ {
@@ -95,7 +95,6 @@ func BenchmarkBlock(b *testing.B) {
 							test.hash.Reset()
 						}
 					}
-
 
 					bbb.SetBytes(int64(len(buf)))
 					bbb.ReportAllocs()
@@ -129,7 +128,7 @@ func BenchmarkRoll(b *testing.B) {
 
 	for _, testSize := range sizes {
 		for _, test := range tests {
-			b.Run(test.name + "-" + fmt.Sprint(testSize), func(bb *testing.B) {
+			b.Run(test.name+"-"+fmt.Sprint(testSize), func(bb *testing.B) {
 				bb.Run("", func(bbb *testing.B) {
 					data := make([]byte, testSize)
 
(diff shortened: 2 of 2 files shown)


calmh:
The reason it's slow is Write() delegates to the original adler32 by first appending to new data to the internal window and then calling Reset()/Write(). With our 32k copy buffer that means we're hashing the same data lots of times, first the first 32k, then 64k, then 96k, etc... A larger copy buffer would avoid this. But this is fine too.

imsodin:
> The reason it's slow is Write() delegates to the original adler32 by first appending to new data to the internal window and then calling Reset()/Write(). With our 32k copy buffer that means we're hashing the same data lots of times, first the first 32k, then 64k, then 96k, etc... A larger copy buffer would avoid this. But this is fine too.

Does that mean the problem might still exist in the weak hash finder where we do need to roll and can be fixed with different copy buffers? Or is it already done correctly there?

AudriusButkevicius:
I think rolling does not do a copy, you do the copy once to init the buffer and then the roll just fiddles with bytes in that. Yeah, but benchmarks show it's not that quick.

calmh:
Rolling is slow by itself, but doesn't suffer from this issue so it's fine as is.

chmduquesne:
Rolling definitely does not do a copy, otherwise this lib would be pretty useless :-)

PR #5558, opened 2019-02-26
lib/tlsutil: Enable TLS 1.3 when available, on test builds (fixes #5065)

### LINKED ISSUE #5065: Add support for TLS1.3
TLS1.3 support should land in golang [1.11 or 1.12](https://github.com/golang/go/issues/9671) and would improve connection security/latency.

Check the following blog posts to get a grasp of its benefits:
https://www.cloudflare.com/learning-resources/tls-1-3/
https://kinsta.com/blog/tls-1-3/

calmh:
### Purpose

This enables TLS 1.3 negotiation on Go 1.12 by setting the GODEBUG
variable. For now, this just gets enabled on test versions (those with a
dash in the version number).

Users wishing to enable this on production builds can set GODEBUG
manually.

The string representation of connections now includes the TLS version
and cipher suite. This becomes part of the log output on connections.
That is, when talking to an old client:

    Established secure connection .../TLS1.2-TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256

and now potentially:

    Established secure connection .../TLS1.3-TLS_AES_128_GCM_SHA256

(The cipher suite was there previously in the log output, but not the
TLS version.)

I also added this info as a new Crypto() method on the connection, and
propagate this out to the API and GUI, where it can be seen in the
connection address hover (although with bad word wrapping sometimes).

screen shot 2019-02-26 at 07 31 24

### Testing

I ran it manually and got the expected negotiations.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -709,7 +709,7 @@
                     <tr>
                       <th><span class="fas fa-fw fa-link"></span>&nbsp<span translate>Address</span></th>
                       <td ng-if="connections[deviceCfg.deviceID].connected" class="text-right">
-                        <span tooltip data-original-title="{{ connections[deviceCfg.deviceID].type.indexOf('Relay') > -1 ? '' : connections[deviceCfg.deviceID].type }}">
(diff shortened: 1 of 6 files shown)


AudriusButkevicius:
Yet not sure why this technical stuff matters to the user.

PR #5563, opened 2019-02-27
lib/model: Use errors.Wrap for pull errors

imsodin:
The original intention was to remove the meaningless `finisher` context, prompted by a report with this log line:

```
[...] 10:10:15 INFO: Puller (folder "..." (...), file "..."): finisher: pull: no such file
```

Then realized, that it's not the only place with non-ideal contexts and that there is `errors.Wrap`, so I switched to using that and in the process removed/added/changed some contexts.

And while at it I most likely found a bug that overwrites conflicts. I didn't have time to write a unit test to reproduce it, but I am pretty sure about it, so I added comments at the relevant places.

### DIFF
--- a/lib/model/sharedpullerstate.go
+++ b/lib/model/sharedpullerstate.go
@@ -7,10 +7,11 @@
 package model
 
 import (
-	"fmt"
 	"io"
 	"path/filepath"
 	"time"
+
+	"github.com/pkg/errors"
 
 	"github.com/syncthing/syncthing/lib/fs"
 	"github.com/syncthing/syncthing/lib/protocol"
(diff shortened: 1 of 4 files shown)


AudriusButkevicius:
Needs more time to think about it in front of a larger screen.

lib/model/sharedpullerstate.go
```diff
@@ -176,7 +177,7 @@ func (s *sharedPullerState) tempFile() (io.WriterAt, error) {
 					l.Debugln("failed to remove temporary file:", remErr)
 				}
 
-				s.failLocked("dst truncate", err)
+				s.failLocked(err)
```
AudriusButkevicius: Not wrapped

imsodin: What would the context be? I expect the error message from `fd.Truncate` to say that it tried to truncate and I struggle to come up with a succinct explanation of what exactly the corner case is that resulted in this error being emitted ("setting size of sparse file with reused blocks").

imsodin:
> So this looks ok, other than the comments added which are addressed in the other PR.

You're right, I should have opened an issue instead -> will do.

lib/model/sharedpullerstate.go
AudriusButkevicius: Well if it's something like permission denied, we won't know what operation it's for.

imsodin: These functions always return an error like `operation: error`, so having `dst truncate:

PR #5565, opened 2019-02-27
Change remote device table column proportions

### LINKED ISSUE #4531: GUI folder/device status rows should handle longer values nicer
Hi,
in the WebGui, can we have the full view on IPV6 adresses?

They are unreadable because of the "..." at the end of each adress.

Steps to reproduce:

Open the WebGUI and open Remote Devices to see some IPs

 - what you expected to happen instead

The whole IP Adress, so I can compare or textcopy them.

Versions:
Syncthing Version: v0.14.40, 64Bit on Mac/Linux/Windows
OS Version: Windows 7 / Mac OSX Sierra
Browser Version: FireFox / Safari 

Screenshots:
bildschirmfoto 2017-11-19 um 21 19 43

bildschirmfoto 2017-11-19 um 21 30 30

otbutz:
### Purpose

Fixes #4531

### Screenshots

Large:
grafik

Medium:
grafik

Small:
grafik

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -124,6 +124,18 @@ table.table-condensed td.no-overflow-ellipse {
     white-space: normal;
 }
 
+table.table-auto {
+	table-layout: auto;
+}
+
+table.table-auto th {
+	width: 1px;
+}
+
+table.table-auto td {
+    max-width: 0px;
+}
+
 .folder-advanced {
     padding: 1rem;
     margin-bottom: 15px;
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -337,7 +337,7 @@
             </button>
             <div id="folder-{{$index}}" class="panel-collapse collapse">
               <div class="panel-body">
-                <table class="table table-condensed table-striped">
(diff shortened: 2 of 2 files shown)


calmh:
You're making the label column rather narrow. What happens in translations with longer labels? It looks like the IPv6 address flows nicely under the label, but it also happens to already be on a lower line - what happens if it's on the same line as the label?

otbutz:
grafik

Better?

The table-layout is now `auto` and the `th` column scales based on its content using `width: 1px;`. The `td` claims only the remaining space with `max-width: 0px;`

https://github.com/otbutz/syncthing/tree/fit-to-content

otbutz:
@calmh should I open a second pull request for my first commit? I think that it's fix for a regression in the current master branch: https://github.com/syncthing/syncthing/pull/5565/commits/321e3c0b9e1bee573dc3b37e9aa8998907245f62

otbutz:
I haven't tested it but it seems reasonable to use it for them as well.

otbutz:
@calmh Give it a spin ;)

calmh:
👌

PR #5567, opened 2019-02-27
Dir sym replace confl

imsodin:
The first commit is https://github.com/syncthing/syncthing/pull/5563, please have a look at that first and/or ignore it when reviewing this (both PRs change the same code and I don't feel like solving conflicts).

### Purpose

fix 5566

### Testing

New unit tests for symlinks and dirs.


lib/model/folder_sendrecv_test.go
```diff
@@ -863,3 +882,74 @@ func TestCopyOwner(t *testing.T) {
...
+
+	if confls := existingConflicts(name, ffs); len(confls) != 1 {
+		t.Fatal("Expected one conflict, got", len(confls))
+	} else if scan := <-scanChan; confls[0] != scan {
+		t.Fatal("Expected request to scan", confls[0], "got", scan)
+	}
+}
+
+// TestSRConflictReplaceFileByLink checks that a conflict is created when an existing file
+// is replaced with a directory and versions are conflicting
```
AudriusButkevicius: link

AudriusButkevicius:
There is so many changes in this PR which makes it unclear where the fix to the explained problem is.

lib/model/folder_sendrecv.go
```diff
@@ -1880,8 +1909,35 @@ func (f *sendReceiveFolder) deleteDir(dir string, ignores *ignore.Matcher, scanC
 	return err
 }
 
+// itemChanged returns true if the given disk file differs from the information
+// in the database and schedules that file for scanning
+func (f *sendReceiveFolder) itemChanged(stat fs.FileInfo, item protocol.FileInfo, hasItem bool, scanChan chan<- string) (changed bool, err error) {
```
AudriusButkevicius: I think a lot of stuff that is passed into this, could just be geneerated by this function itself, so you'd just need to pass in a name or protocol.FileInfo (to avoid the db access).

I wonder how often we'd end up with double stats because of this. If alot, perhaps not worth it.

lib/model/sharedpullerstate.go
```diff
@@ -176,7 +177,7 @@ func (s *sharedPullerState) tempFile() (io.WriterAt, error) {
 					l.Debugln("failed to remove temporary file:", remErr)
 				}
 
-				s.failLocked("dst truncate", err)
+				s.failLocked(err)
```
AudriusButkevicius: Why no wrap?

lib/model/folder_sendrecv.go
```diff
@@ -789,15 +840,20 @@ func (f *sendReceiveFolder) deleteFile(file protocol.FileInfo, scanChan chan<- s
...
-		// we have resolved the conflict.
-		file.Version = file.Version.Merge(cur.Version)
-		err = osutil.InWritableDir(func(name string) error {
-			return f.moveForConflict(name, file.ModifiedBy.String(), scanChan)
-		}, f.fs, file.Name)
-	} else if f.versioner != nil && !cur.IsSymlink() {
+		// There is a conflict here, which shouldn't happen as deletions
+		// always lose. Merge the version vector of the file we have
+		// locally and commit it to db to resolve the conflict.
+		// of deleting.
```
AudriusButkevicius: extra dot in the sentence.

lib/model/folder_sendrecv.go
imsodin: `itemChanged` is usually called when an item is there aka after stat returned a nil error. If stat returns an `IsNotExist` error, this is a special case and doesn't need the full rescan treatment, as a change will always win over a deletion in a conflict. To me it seems semantically simpler/clearer to first check and act on the stat result and if necessary, call `itemChanged`. I believe the correct thing would happen, if `itemChanged` returned false (file did not change) if there is no file on disk, but I wouldn't expect it to do that given it's name.

imsodin:
@AudriusButkevicius Where are we with this?

lib/model/folder_sendrecv.go
```diff
@@ -798,15 +840,19 @@ func (f *sendReceiveFolder) deleteFile(file protocol.FileInfo, scanChan chan<- s
 		return dbUpdateJob{}, err
 	}
 
+	if cur.IsDirectory() {
+		return dbUpdateJob{}, errUnexpectedDirOnFileDel
+	}
```
AudriusButkevicius: Why doesn't this trigger a scan?

lib/model/folder_sendrecv.go
```diff
@@ -567,17 +567,39 @@ func (f *sendReceiveFolder) handleDir(file protocol.FileInfo, dbUpdateChan chan<
...
+			return
+		} else if changed {
+			l.Debugln("item changed on disk compared to db; not replacing with dir:", file.Name)
+			scanChan <- curFile.Name
+			f.newPullError(file.Name, errModified)
+			return
+		}
+
+		// Remove it to replace with the dir.
+		if !curFile.IsSymlink() && f.inConflict(curFile.Version, file.Version) {
```
AudriusButkevicius: Shouldn't this check info.IsSymlink()?

AudriusButkevicius: I guess it has not changed, so we can trust the db?

AudriusButkevicius: Also the not a dir not a symlimk is spread over two ifs making it had to follow the logic.

imsodin: The two checks aren't really related: If `IsDir`, the item on disk doesn't need to be checked/conflicted/deleted/archived, it just gets updated. If `!IsDir`, we need to check `!IsSymlink` due to conflicts (only then). Would a (redundant) `curFile.IsDirectory()` make it more readable?

lib/model/folder_sendrecv.go
imsodin: Is the added comment clear?

lib/model/folder_sendrecv.go
AudriusButkevicius: I think adding a second directory check before moving for conflict would make it clearer. Also, I am starting to wonder how this works, the local conflict is always the loosing conflict which is not what I would have expected.

lib/model/folder_sendrecv.go
AudriusButkevicius: Sure, but why are we not scheduling a scan in this case?

lib/model/folder_sendrecv.go
imsodin: That's admittedly a bit magical: In the db layer files are ordered by the usual conflict mechanism, and are thus only needed locally if our side of the conflict is loosing. That's why we don't need to check which item wins in the puller - if it wasn't the local one it wouldn't come up in the puller.

lib/model/folder_sendrecv.go
imsodin: So my comment wasn't clear :)  
Just above we checked that what is on disk (a dir) is the same as what is in the db, so a scan would not change anything about the situation. What is wrong is that the remote says: "Delete this file" and we see "We don't have a file, it's a dir". Maybe we could do some magic at this point to resolve the situation, but really I don't think this case should ever happen. However I can't prove it, so I want to make it a visible error, but not a panic, as that's really user-unfriendly. If I get proven wrong by reports, we can think about what magic is needed to resolve the situation.

AudriusButkevicius:
Feel free to merge.

PR #5574, opened 2019-03-04
lib/scanner: Consistently use CreateFileInfo and remove outdated comment

imsodin:
The removed comment simply isn't true and hasn't been in a while. I can't think of why I didn't use `CreateFileInfo` for symlinks, now I do.  
No change in functionality. The added `NoPermissions` line in `CreateFileInfo` does not have any effect, as permissions are never considered for symlinks, but it's still the correct thing to do.


AudriusButkevicius:
Please check the failures, if unrelated, feel free to merge.

PR #5576, opened 2019-03-05
lib/model: Optimize dbUpdaterRoutine

imsodin:
Saving and iterating the files twice is totally unnecessary. That I used `fileInfoBatch` is tangential - it's just convenient.

Sorry (well not so much actually) for the flood of low-impact PRs, it's a consequence of me trying to stop carrying around unrelated changes in feature branches that may eventually come to light (probably not).


lib/model/folder_sendrecv.go
```diff
@@ -1678,26 +1650,35 @@ loop:
...
+			if !job.file.IsInvalid() && job.jobType&(dbUpdateHandleFile|dbUpdateDeleteFile) != 0 {
+				found = true
+				lastFile = job.file
 			}
 
+			job.file.Sequence = 0
+
+			batch.append(job.file)
+
+			batch.flushIfFull()
```
golangcibot: Error return value of `batch.flushIfFull` is not checked (from `errcheck`)

PR #5581, opened 2019-03-07
lib/model: Use temp dir from osutils in tests

imsodin:
When adding the code that's removed in this PR I wasn't aware that `osutil.TempFile` with an empty dir argument creates the file in a platform specific temporary directory. And `os.RemovAll("/tmp")` was a very smart move (not that it would have worked, but still).


AudriusButkevicius:
Not even sure how this slipped through ...

AudriusButkevicius:
Again, on phone, can't merge with failed statuses.

PR #5585, opened 2019-03-08
lib/model: Introduce waitForState test utility

imsodin:
Not much to say, it's just as boring as it looks like.

### DIFF
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -1768,6 +1768,23 @@ func TestIgnores(t *testing.T) {
 	changeIgnores(t, m, []string{})
 }
 
+func waitForState(t *testing.T, m *model, status string) {
+	t.Helper()
+	timeout := time.Now().Add(2 * time.Second)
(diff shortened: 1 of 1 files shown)


lib/model/model_test.go
```diff
@@ -1768,6 +1768,23 @@ func TestIgnores(t *testing.T) {
...
+			return
+		}
+		if err != nil && err.Error() == status {
+			return
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+	t.Fatalf("Timed out waiting for status: %s, current status: %v", status, err)
+}
+
```
AudriusButkevicius: Perhaps should take a folder name to be more generally useful?
Also, perhaps all of these should start going into some sort of model_test_utils.go or something

imsodin: Agreed about the separate file for utilities, but that's a potentially deep rabbithole so I'd rather not do it in right now in this PR.

PR #5586, opened 2019-03-09
lib/model: Introduce must test utility

imsodin:
As there's already `fatalOs.must`, why not change that to `must(fatal, func() error)` to save a few constructs like

```
if err := fn(...); err != nil {
        t.Fatal(err)
}
```

However now a few days after that thought I am no longer convinced that this is really any better:

```diff
-	if err := ioutil.WriteFile("testdata/testrm2", []byte("Hello"), 0644); err != nil {
-		t.Fatal(err)
-	}
+	must(t, func() error { return ioutil.WriteFile("testdata/testrm2", []byte("Hello"), 0644) })
```

Since variadic return values don't exist, it's anyway just a partial patch.

Before sending the change to oblivion I'd like to hear another opinion.

### DIFF
--- a/lib/model/folder_recvonly_test.go
+++ b/lib/model/folder_recvonly_test.go
@@ -35,15 +35,9 @@ func TestRecvOnlyRevertDeletes(t *testing.T) {
 	testOs.MkdirAll("_recvonly/.stfolder", 0755)
 	testOs.MkdirAll("_recvonly/ignDir", 0755)
 	testOs.MkdirAll("_recvonly/unknownDir", 0755)
-	if err := ioutil.WriteFile("_recvonly/ignDir/ignFile", []byte("hello\n"), 0644); err != nil {
-		t.Fatal(err)
-	}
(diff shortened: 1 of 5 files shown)


AudriusButkevicius:
What's the point of the function wrapper?
Why not just `must(t, err)` ?

imsodin:
No reason.

AudriusButkevicius:
Generally this is considered an anti-pattern, so I'll see what @calmh thinks.

imsodin:
Then lets just drop it, that's nothing to occupy one more person with - I'd rather have him look at your versioning PR, my gui/api/... refactor or code coverage or whatever before this. Thanks for your time!

AudriusButkevicius:
I don't think it might be that big of a deal in the tests however.

calmh:
I have no strong feelings either way, it reads roughly the same to me regardless. The fatal/panic-on-error function is definitely an anti pattern in other code, but tests already have the fatal so ... whatever

PR #5587, opened 2019-03-09
lib/model: Run recvonly tests in temporary dirs

imsodin:
Receive only folder tests should also run in separate, temporary directories and clean up after themselves (wrapper files were not removed before).

### DIFF
--- a/lib/model/folder_recvonly_test.go
+++ b/lib/model/folder_recvonly_test.go
@@ -11,6 +11,7 @@ import (
 	"context"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


lib/model/folder_recvonly_test.go
```diff
@@ -336,23 +342,17 @@ func setupKnownFiles(t *testing.T, data []byte) []protocol.FileInfo {
...
-	fcfg.RescanIntervalS = 86400
-
-	cfg := defaultCfg.Copy()
-	cfg.Folders = append(cfg.Folders, fcfg)
-
-	wrp := createTmpWrapper(cfg)
-
-	db := db.OpenMemory()
-	m := newModel(wrp, myID, "syncthing", "dev", db, nil)
+	w.SetFolder(fcfg)
```
golangcibot: Error return value of `w.SetFolder` is not checked (from `errcheck`)

imsodin:
Reminder ping

Issue #5346, opened 2018-12-05
Systemd synthing unit starts too early


Rocco83:
Currently, systemd unit must start after network.target. https://github.com/syncthing/syncthing/blob/7203ccb73d110eb91ff9000726591d731cfa44ad/etc/linux-systemd/system/syncthing%40.service#L4 

This is sometimes too early, e.g. in case of sssd (ldap, ...) users.

IMO, unit must be started after multi-user.target.

### Version Information

Syncthing Version: v0.14.53
OS Version: Debian 9
Browser Version: N/A

calmh:
Sounds reasonable.

calmh:
The fix was incorrect and causes a circular dependency (multi-user.target wants syncthing, syncthing must start after multi-user.target). If the problem remains, it must be fixed in a different manner.

calmh:
Reopened by #5676 / e2470c8bc89ae7ce96aa71401f3127795893474f

Rocco83:
Why after multi-user?
What's the purpose?
Or at least a valid use case :-)

calmh:
I'm completely confused by your followup questions , as it seems you are questioning your own premise. Thinking more about it however I think it's fine as is. We are started as part of multiuser. If there are odd things that need to start  before Syncthing they can declare a wanted-by or you can edit to Syncthing unit to add a dependency.

Rocco83:
ooops, sorry, too late in the evening and too fast typing.

I have checked https://www.freedesktop.org/software/systemd/man/systemd.special.html
Another possible way is to depend on:
nss-lookup.target <- lookup for host/service
nss-user-lookup.target <- lookup for user
network.tarkget <- lookup for network
remote-fs.target <- lookup for remote filesystem (nfs, to say)
local-fs.target <- lookup for local filesystem (/home, to say -- some corner case are existing like btrfs not starting "by time" for very large filesystems)

Again, it is a possibility...

mir06:
Thanks @Rocco83 for this hint. I also use sssd with an LDAP backend. Just added `After=network.target nss-user-lookup.target` to `/etc/systemd/system/syncthing@.service` and syncthing is now starting fine for all users.

grandchild:
I think the correct dependency is `After=network-online.target`. This fixes the startup for me on Ubuntu18. The `network-online` target is described here.
Pull request is here: #6328

grandchild:
[...]

PR #5593, opened 2019-03-11
PR for Systemd unit should declare after=multiuser.target #5346

georgespatton:
Should fix the issue:

Systemd unit should declare after=multiuser.target #5346


imsodin:
Is multi-user after network?

nekr0z:
> Is multi-user after network?

Nope. At least not on all systems. And definitely not by default in systemd.

Although `network.target` is only useful as a shutdown dependency (see [here](https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/) for details), we probably need it as such.

TL;DR: `After=network.target multi-user.target` is a way to go if you ask me.

georgespatton:
Updated the PR with both targets.

PR #5597, opened 2019-03-11
cmd: Add check for newer config file and an option to override it (fixes #4921)

### LINKED ISSUE #4921: Syncthing should have guard against running old versions with new .config/* files
Problem:
I have latest release of Syncthing (v0.14.47) and during development & debugging I accidentally run older version of Syncthing v0.14.(4 or 17) and it was successfully started in background because config paths are the same (defaults).
Now, my _config.xml_ becomes a mix of old-style and new-style options but still has version=28 and who knows what can happen to other files if version will be even older(e.g. db format incompatibility).

For example:
```diff
 <configuration version="28"> # both files

-        <minDiskFree unit="%">1</minDiskFree>
+        <minDiskFreePct>0</minDiskFreePct>
-        <pullerMaxPendingKiB>0</pullerMaxPendingKiB>
+        <pullers>0</pullers>
```

What's expected? One of this:
- it should fail to start early with appropriate message unless some special option is used like `--allow-downgrade` (or whatever)
- or at least start in read-only mode for everything

### Version Information

Syncthing Version: Any
OS Version: Any

nekr0z:
### Testing

Compiled and tested manually on linux-amd64, works as expected.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -990,7 +990,7 @@ func loadConfigAtStartup(forceNewConfig bool) (config.Wrapper, error) {
 
 	if cfg.RawCopy().OriginalVersion != config.CurrentVersion {
 		if cfg.RawCopy().OriginalVersion > config.CurrentVersion && !forceNewConfig {
-			return nil, errors.New("Existing config version is newer than expected. Use -force-new-config to override.")
+			return nil, errors.New("Config file version (" + strconv.Itoa(cfg.RawCopy().OriginalVersion) + ") is newer than supported version (" + strconv.Itoa(config.CurrentVersion) + "). If this is expected, use -force-new-config to override.")
 		}
 		err = archiveAndSaveConfig(cfg)
 		if err != nil {
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -163,35 +163,35 @@ var (
 )
 
 type RuntimeOptions struct {
-	confDir        string
-	resetDatabase  bool
-	resetDeltaIdxs bool
-	showVersion    bool
-	showPaths      bool
-	showDeviceId   bool
-	doUpgrade      bool
-	doUpgradeCheck bool
-	upgradeTo      string
-	noBrowser      bool
-	browserOnly    bool
-	hideConsole    bool
-	logFile        string
-	auditEnabled   bool
-	auditFile      string
-	verbose        bool
-	paused         bool
-	unpaused       bool
-	guiAddress     string
-	guiAPIKey      string
-	generateDir    string
-	noRestart      bool
-	profiler       string
-	assetDir       string
-	cpuProfile     bool
-	stRestarting   bool
-	logFlags       int
-	showHelp       bool
-	allowNewConfig bool
+	confDir          string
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
The error message should print the versions and suggest to use the option only of this is expected.

nekr0z:
Fair point, will do.

пн, 11 мар. 2019 г. в 20:44, Audrius Butkevicius <notifications@github.com>:

nekr0z:
Done.

cmd/syncthing/main.go
```diff
@@ -987,6 +989,9 @@ func loadConfigAtStartup() (config.Wrapper, error) {
 	}
 
 	if cfg.RawCopy().OriginalVersion != config.CurrentVersion {
+		if cfg.RawCopy().OriginalVersion > config.CurrentVersion && !forceNewConfig {
+			return nil, errors.New("Config file version (" + strconv.Itoa(cfg.RawCopy().OriginalVersion) + ") is newer than supported version (" + strconv.Itoa(config.CurrentVersion) + "). If this is expected, use -force-new-config to override.")
```
AudriusButkevicius: `fmt.Errorf("bla bla %d bla", number)`

Also, the flag is not forcing anything, but allowing, so perhaps -allow-newer-config

nekr0z:
> I wonder about refusing future configs via api like explained in the ticket, but I guess that's not very often that it would happen, as most users get, modify, post back.

Should be doable, but lib/config is a little bit beyond my speaking go just yet. :(

AudriusButkevicius:
I think -allow-newer-config would be the more correct term, as -allow-new-config is a bit ambigious, for example, if we already have a database, but no config in some distant future we might refuse to start, and option like that might be used to allow creation of a new config.

Again, I know I am annoying and it's a nit-pick, but I just want to be extra clear what the option means.

cmd/syncthing/main.go
```diff
@@ -244,6 +245,7 @@ func parseCommandLineOptions() RuntimeOptions {
 	flag.BoolVar(&options.unpaused, "unpaused", false, "Start with all devices and folders unpaused")
 	flag.StringVar(&options.logFile, "logfile", options.logFile, "Log file name (still always logs to stdout). Cannot be used together with -no-restart/STNORESTART environment variable.")
 	flag.StringVar(&options.auditFile, "auditfile", options.auditFile, "Specify audit file (use \"-\" for stdout, \"--\" for stderr)")
+	flag.BoolVar(&options.allowNewConfig, "allow-new-config", false, "Allow loading newer version of config file")
```
AudriusButkevicius: Perhaps we should drop the word "file", as I don't think we refer to it being a file explicitly anywhere else. 
For all that we know the config might become a directory.

AudriusButkevicius:
Sorry, given you are new, I tend to look at PRs with more details and be annoying and picky :(

nekr0z:
> Sorry, given you are new, I tend to look at PRs with more details and be annoying and picky :(

Not only is it OK, but thank you for that! As a developer, I'm only starting to learn how to do things, and your input helps me a lot. As a user, the tighter code review the better.

PR #5600, opened 2019-03-12
lib/config: Set UseLargeBlocks to true by default (fixes #5599)

### LINKED ISSUE #5599: Default folder in v1.1.0 isn’t created with useLargeBlocks = true
https://forum.syncthing.net/t/default-folder-in-v1-1-0-isnt-created-with-uselargeblocks-true/13003

imsodin:
Fixes #5599, followup to #5405.


calmh:
It's unclear to me what this would accomplish. In the issue (forum link) it talks about the API. If a config comes in through the API it's expected to be complete. I'm pretty sure our setdefaults stuff doesn't take effect on that path. Maybe they could, but it seems tricky.

imsodin:
As I understood it, the thread in the forum talks about the default folder, which is generated in main.go through `NewFolderconfiguration` where defaults do apply.

calmh:
Right if this is for the default folder then I understand.

PR #5605, opened 2019-03-19
gui: Add licensing information to aboutModalView (fixes #1223)

### LINKED ISSUE #1223: mention license in about dialog
the user don't sees that Syncthing is free software inside the webinterface.

for me that is very important!

nekr0z:
### Testing

Compiled, ran, works. (linux-amd64)

### DIFF
--- a/gui/default/syncthing/core/aboutModalView.html
+++ b/gui/default/syncthing/core/aboutModalView.html
@@ -9,6 +9,8 @@
     </h1>
     <hr />
 
+    <p><span translate>Syncthing is Free and Open Source Software licensed as</span> <a href="https://github.com/syncthing/syncthing/blob/master/LICENSE">MPL v2.0</a>.</p>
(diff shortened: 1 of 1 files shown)


gui/default/syncthing/core/aboutModalView.html
```diff
@@ -9,6 +9,8 @@ <h1 class="text-center">
     </h1>
     <hr />
 
+    <p><span translate>Syncthing is Free and Open Source Software licensed as</span> <a href="https://github.com/syncthing/syncthing/blob/master/LICENSE">MPL v2.0</a>.</p>
```
calmh: I don't think we can break a translated sentence like this; other languages might need to put the "MPL 2.0" string somewhere in the middle.

nekr0z: Good point.

PR #5608, opened 2019-03-20
lib/model: Update db on scan/pull in folder

imsodin:
Updating the db after pulling or scanning is done in model, while all the actual work is done in folder -> move it there. This also avoids having to pass the model update functions as arguments to folder methods.


imsodin:
Reminding ping.

PR #5611, opened 2019-03-21
lib/fs: Fallback EvalSymlinks method on windows (fixes #5609)

### LINKED ISSUE #5609: Filesystem watching failed when parent folder is not listable
### The problem

On windows, if any of the ancestor folders of a monitored folder is not listable (i.e., the "List Folder" permission is unset), the file system watcher will fail. In the verbose log there are these messages:

```
[VK627] 15:17:59 VERBOSE: FolderWatchStateChanged events.Event{SubscriptionID:26, GlobalID:26, Time:time.Time{wall:0xbf1ce1fdcebe16cc, ext:1632000601, loc:(*time.Location)(0x1463440)}, Type:33554432, Data:map[string]interface {}{"folder":"maa2q-yxpwp", "to":"Access is denied."}}
[VK627] 15:18:05 VERBOSE: Summary for folder "maa2q-yxpwp" is map[errors:0 globalBytes:0 globalDeleted:0 globalDirectories:0 globalFiles:0 globalSymlinks:0 globalTotalItems:0 inSyncBytes:0 inSyncFiles:0 localBytes:0 localDeleted:0 localDirectories:0 localFiles:0 localSymlinks:0 localTotalItems:0 needBytes:0 needDeletes:0 needDirectories:0 needFiles:0 needSymlinks:0 needTotalItems:0 pullErrors:0 sequence:0 state:idle version:0 watchError:Access is denied.]
```

The "Watch Error: Access is denied" message will finally appear on the GUI webpage.

I am working in a multi-user workstation where fine-grained ACLs are configured by the administrator. It's common to have a folder structure like "C:\secret_agents\007" where I own the "007" folder but cannot list the parent folder, "secret_agent". 
Older versions of Syncthing (<= 0.14.49) worked fine with such ACL settings but the above error starts to appear when I update syncthing to newer versions (>= 1.0).

### Environments

- Syncthing Version: v1.1.0
- OS Version: Windows 10 and 7
- File system: NTFS

### Steps to reproduce

1. Create a folder `C:\parent\child` and configure Syncthing to monitor it.
2. In the access control list (ACL) of folder `C:\parent` , unset the "List Folder" permission and keep the "Traverse Folder" permission allowed.
3. Enable all permissions for folder `C:\parent\child`.
4. Restart Syncthing.

gdlmx:
### Purpose

This PR fixes #5609 by circumventing the "access denied" error of `filepath.EvalSymlinks`. In the current implementation of `EvalSymlinks`, it makes a [`FindFirstFile` system call](https://github.com/golang/go/blob/44d3bb998ca00e49d9e0138954287af206b614bf/src/path/filepath/symlink_windows.go#L34) for every ancestor directories. Therefore it will fail if If the user lacks access permission to any of those ancestors.

This PR introduces a fallback method to get the ["normalized"](https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-getfinalpathnamebyhandlew#remarks) path from a more suitable WinAPI call, `GetFinalPathNameByHandle` which tries to query the norm path from the file system directly without visiting any ancestor directory. Therefore it can circumvent the above problem.

By definition, the norm path doesn't contain any symlink or short 8.3 file names. The case of the path components should also be correct. 

### Testing

A unit test function `TestGetFinalPath` is composed to test normal windows paths as long as a path to a special folder structure with ACLs that will trigger the issue [#5609](https://github.com/syncthing/syncthing/issues/5609#issuecomment-475275310). This test finishes successfully on Windows 10 (build 1809).

## Authorship

The PR is authored by Mingxuan Lin

### DIFF
--- a/lib/fs/basicfs_windows.go
+++ b/lib/fs/basicfs_windows.go
@@ -218,7 +218,33 @@ func evalSymlinks(in string) (string, error) {
 		out, err = filepath.EvalSymlinks(in[4:])
 	}
 	if err != nil {
-		return "", err
+		isSymlink, err_i := isUnderSymlink(in)
+		if err_i != nil || isSymlink {
+			return "", err
+		}
+		// Recover from the prior error from EvalSymlinks
+		// if the `in` path doesn't contain any symlink
+		out = in
 	}
 	return longFilenameSupport(out), nil
+}
+
+func isUnderSymlink(in string) (bool, error) {
+	separator := string(os.PathSeparator)
+	abspath := ""
+	for i, sec := range strings.Split(in, separator) {
+		if i == 0 {
+			abspath = sec
+			continue
+		}
+		abspath += separator + sec
+		fi, err := os.Lstat(abspath)
+		if err != nil {
+			return false, err
+		}
+		if fi.Mode()&os.ModeSymlink != 0 {
+			return true, nil
+		}
+	}
+	return false, nil
 }
--- a/lib/fs/basicfs_windows.go
+++ b/lib/fs/basicfs_windows.go
@@ -211,30 +211,6 @@ func isMaybeWin83(absPath string) bool {
 	return strings.Contains(strings.TrimPrefix(filepath.Base(absPath), WindowsTempPrefix), "~")
 }
 
-func evalSymlinks(in string) (string, error) {
-	out, err := filepath.EvalSymlinks(in)
-	if err != nil && strings.HasPrefix(in, `\\?\`) {
-		// Try again without the `\\?\` prefix
-		out, err = filepath.EvalSymlinks(in[4:])
-	}
-	if err != nil {
-		// Try to get a normalized path from Win-API
-		var err1 error
-		out, err1 = getFinalPathName(in)
-		if err1 != nil {
-			return "", err // return the prior error
-		}
-		// Trim UNC prefix, equivalent to
-		// https://github.com/golang/go/blob/2396101e0590cb7d77556924249c26af0ccd9eff/src/os/file_windows.go#L470
-		if strings.HasPrefix(out, `\\?\UNC\`) {
-			out = `\` + out[7:] // path like \\server\share\...
-		} else {
-			out = strings.TrimPrefix(out, `\\?\`)
-		}
-	}
-	return longFilenameSupport(out), nil
-}
-
 func getFinalPathName(in string) (string, error) {
 	// Return the normalized path
 	// Wrap the call to GetFinalPathNameByHandleW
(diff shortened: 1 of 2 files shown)


imsodin:
This now looks like it is essentially what `filepath.EvalSymlinks` does on windows if their "normal operation" encounters an error (`evalSymlinksUsingGetFinalPathNameByHandle`), except that the filepath implementation then passes every path component of the result of `GetFinalPathName` to `FindFirstFile`. The comment on that function says it does so to get the "correct case". However your test cases implies that `GetFinalPathName` already returns the correct case. So this could be reported to upstream and eventually fixed there. And in the meantime for easier understanding/maintenance copy code ad verbatim (where possible) from https://golang.org/src/path/filepath/symlink_windows.go and https://golang.org/src/internal/syscall/windows/syscall_windows.go.

Also does this method `GetFinalPathName` have any drawbacks (e.g. unreliable, not available on all supported windowses) or would this be a cheap/preliminary fix for the case sensitivity issues we have (instead of weird behaviour, we could check the "real"/preexisting name and throw an error if the pulled file info has different case).

gdlmx:
The API method `GetFinalPathName` does "more" things than `EvalSymlinks`:
- It also resolves [network drive letter](https://support.microsoft.com/en-us/help/4026635/windows-map-a-network-drive). For example, if drive `H:\` is a network drive mounted to `\\192.168.1.2\Users\xunlei`. It will also expand `H:` so that 
```go
GetFinalPathName(`H:\Desktop`) == `\\?\UNC\192.168.1.2\Users\tom\Desktop`
```
- It always returns a [fully qualified path](https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file#fully-qualified-vs-relative-paths) even when the input is a relative path. That is, it implies `GetFullPathName` and `GetLongPathName`.

```go
// Current dir at `H:\`
GetFinalPathName(`.\Desktop`) == `\\?\UNC\192.168.1.2\Users\tom\Desktop`
```
In this PR, I use  `GetFinalPathName` as a *fallback method* simply because of those *extra* things it does.
Therefore, I don't think we can propose `GetFinalPathName` to upstream as a fix to `EvalSymlinks` because of the change of behavior. Instead, we can propose to add it to `syscall` to make it easier to use.

The most noticeable drawback/limitation is that Windows XP doesn't support `GetFinalPathName`. Maybe that's why Go doesn't have it in syscall yet. It's good to know that Go version 1.11 [drops support for XP](https://github.com/golang/go/wiki/Windows).

gdlmx:
@imsodin We can copy [codes of the private functions](https://github.com/golang/go/blob/d227a0811b76791fad04eeba35cf2794a719d610/src/os/file_windows.go#L430) `normaliseLinkPath` and `openSymlink` in the `os` package, if you prefer this way. `normaliseLinkPath` is basically equivalent to my `getFinalPathName` function in this PR.

imsodin:
@gdlmx Now that https://github.com/syncthing/syncthing/pull/5734 has landed it's fine to merge this. However some code regarding watch paths/windows roots changed in the meantime, such that this PR needs to be adjusted before merge. Can you take care of it or should I do it?

imsodin:
@calmh The characteristics of `GetFinalPathName` seem consistent to what's expected from `filepath.EvalSymlinks` (casing, ...), so in my opinion it should be save to use in `evalSymlink`. Alternatively if we want to stick to the default behaviour, it could be moved to `watchPaths` to apply to watching operations only. What do you think?

calmh:
My concerns were for the correct operation of "your" areas (i.e. the watcher), so if you're happy with that then I'm happy. Merge away.

PR #5622, opened 2019-03-27
lib/model: Optimize puller for meta only changes

imsodin:
### Purpose

Pulling can be quite slow and consume lots of memory. While working on reactivating #4980 I noticed some of the intermediate storage of file infos to process is not necessary or can be avoided. There are two main changes:

1. Remove `processDirectly` by really processing those items directly.  
Concerns dirs&symlinks only, should be beneficial in all scenarios. I also remove the ordering by component count, as that is a noop due to elements in the db being sorted lexicographically, which implies parents first as well.

2. Check for metadata only file changes before pushing items to queue.  
This is aimed mostly at initial sync with preexisting data or big deletions: A huge list of out-of-sync items, but they are all equal. In this case we save all the queue storage and sorting and one database lookup. For non-shortcuteable files, this costs one more database lookup.

### Testing

Existing unit and integration tests pass.


AudriusButkevicius:
This wins close to nothing, but the change in itself is not wrong.

PR #5623, opened 2019-03-27
lib/model: Refactor progressEmitter to de-/activate by config (fixes #4613)

### LINKED ISSUE #4613: Progress emitter doesn't start from config change if disabled at startup
Just a todo for something I ran into while looking at something else. The `Serve()` method is only run if the interval is positive when the model starts, yet it may become positive later due to a config commit and the progress emitter looks like it's made to handle this. Fix is trivial, but I want to write a test at the same time.

imsodin:
The main purpose is to make the progress emitter actually en-/disable when `ProgressUpdateIntervalS` is changed between negative and positive. This requires to make some kind of cleanup on disabling, that deregisters all already sent progress updates on connected remotes. And the way puller states and connections were saved (separating folder id and file name by `//` instead of a double map), and copies created on every send (5s by default) seemed overly complicated, so I changed that using maps. There's probably more potential for simplification, but I didn't want to deviate much from how it worked before.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -1219,9 +1219,7 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
 			continue
 		}
 
-		if f.model.progressEmitter != nil {
-			f.model.progressEmitter.Register(state.sharedPullerState)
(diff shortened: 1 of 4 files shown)


imsodin:
Reminding ping.

PR #5626, opened 2019-03-27
lib/model: Missing queue.Done and reset between pulls (fixes #5332)

### LINKED ISSUE #5332: Duplicate Files Names in Out of Sync List
One of my shared folders shows one item out of sync on the Syncthing web interface. When I click through to see the list of out of sync items it shows 7 files with the same name. They are all syncing and showing the exact same level of progress and they increment at the same time. On a second identical backup system there are 3 of the same file name shown.

There is only one copy of this file on the source computer with this name that should be syncing. The path shown for all 7 or all 3 are identical.

Everything appears to be working as expected, this appears to be a display issue that I caught because the file is large and takes a long time to sync.

In case it provides a hint the backup system with 7 names in the list is on the LAN while the one with 3 names is remote and syncing at about half the rate.

All systems are running 0.14.52 on Linux. The source running Fedora 29, the two backup computers are running Ubuntu 18.04.1 LTS on Odroid-HC2 computers.

syncthingduplicateitems

syncthingduplicateitems2

imsodin:
I am not actually certain it fixes #5332 for lack of reproducer, but it's likely the cause: When a file cannot be pulled, it isn't removed from the queue and readded on the next try. And the queue is never reset, so that stays there "forever" (as long as the folder is constantly running). I fixed that instance and added `jobQueue.Reset` that removes everything from queue that's invoked when the iteration is finished - in case the folder is stopped (or I missed something else). And I noticed a missing `changed++` on rename, which is largely cosmetic (as in can't think of an adverse effect right now).

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -291,6 +291,8 @@ func (f *sendReceiveFolder) pullerIteration(scanChan chan<- string) int {
 }
 
 func (f *sendReceiveFolder) processNeeded(dbUpdateChan chan<- dbUpdateJob, copyChan chan<- copyBlocksState, scanChan chan<- string) (int, map[string]protocol.FileInfo, []protocol.FileInfo, error) {
+	defer f.queue.Reset()
+
 	changed := 0
 	var processDirectly []protocol.FileInfo
 	var dirDeletions []protocol.FileInfo
@@ -478,6 +480,7 @@ nextFile:
 				// Remove the pending deletion (as we performed it by renaming)
 				delete(fileDeletions, candidate.Name)
 
+				changed++
 				f.queue.Done(fileName)
 				continue nextFile
 			}
@@ -493,6 +496,7 @@ nextFile:
 			}
 		}
 		f.newPullError(fileName, errNotAvailable)
+		f.queue.Done(fileName)
 	}
 
 	return changed, fileDeletions, dirDeletions, nil
--- a/lib/model/queue.go
+++ b/lib/model/queue.go
@@ -110,6 +110,13 @@ func (q *jobQueue) Shuffle() {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -291,6 +291,8 @@ func (f *sendReceiveFolder) pullerIteration(scanChan chan<- string) int {
 }
 
 func (f *sendReceiveFolder) processNeeded(dbUpdateChan chan<- dbUpdateJob, copyChan chan<- copyBlocksState, scanChan chan<- string) (int, map[string]protocol.FileInfo, []protocol.FileInfo, error) {
+	defer f.queue.Reset()
```
AndrewSav: This seems to be cause problems now, as I'm explaining [here](https://github.com/syncthing/syncthing/issues/6219#issuecomment-570759947), as `/rest/db/need` call relies on the queue to be up-to-date all the time, not only why processed by `processNeeded`. Is there a reason we need "defer" here at all, and not just plain `f.queue.Reset` so that it's done before the processing and not after? This way the queue will be correct most of the time.  @imsodin @calmh

imsodin: what you describe is the intended behaviour and i don't see a problem: progress can only be shown while something is being processed. at the point queue.Reset is called there's anyway nothing in queue anymore.

AndrewSav: @imsodin processNeeded finishes much earlier that the pullers finished their job.

imsodin: Right, by default the last 2 file are indeed affected by this. i think a fix should be straight forward (just reset after the finisher routine stops). I'll have a look tomorrow.

AndrewSav: Sounds good to me. Thank you for looking into this.

PR #5635, opened 2019-04-03
docker: Create entrypoint script (fixes #5631)

### LINKED ISSUE #5631: Dockerfile ENTRYPOINT doesn't pass along extra arguments from "docker run"
When I'm running syncthing from the Docker image, I would often like to pass it additional command line arguments. My expectation, based on other "application" images I've used (pdal is one example), is that extra arguments to `docker run` would be passed on to `syncthing` in the container.

To reproduce:

```
docker run --pull syncthing/syncthing -help
```

Expected behavior: help text is printed, as with `syncthing -help`.

Actual behavior: `syncthing` starts like nothing was passed:

```
% docker run syncthing/syncthing -help
[monitor] 01:23:27 INFO: Starting syncthing
[start] 01:23:27 INFO: Generating ECDSA key and certificate for syncthing...
[BRLLT] 01:23:27 INFO: syncthing v1.1.0 "Erbium Earthworm" (go1.11.6 linux-amd64) docker@syncthing.net 2019-02-25 18:25:08 UTC
[BRLLT] 01:23:27 INFO: My ID: BRLLTIP-EZSBBHZ-OZJLDQH-SPQWQGD-VT7UQWK-MFFNRSM-5G72AVM-XOEXYA7
[BRLLT] 01:23:28 INFO: Single thread SHA256 performance is 380 MB/s using crypto/sha256 (377 MB/s using minio/sha256-simd).
[... truncated]
```

Docker image version: 61de83be0df4

- - -

I believe this can be addressed through some combination of using the _exec form_ of `ENTRYPOINT` and, optionally, `CMD` for "default" arguments that would be thrown away if the user provides any. See https://docs.docker.com/engine/reference/builder/#entrypoint

tomjakubowski:
### Purpose

Using the [preferred "exec" form](https://docs.docker.com/engine/reference/builder/#entrypoint) of the ENTRYPOINT directive allows the user to pass additional command line arguments to the `syncthing` command when using `docker run` or `docker create`.

### Testing

I built a new image and tagged it:

```
% docker build . -t sthing
```

And then ran the image with no options to check that it behaved like before:

```
% docker run sthing
...
[Q3FAA] 21:25:45 INFO: My name is "70dd5dd00c6f"
^C[monitor] 21:25:56 INFO: Signal 2 received; exiting
[Q3FAA] 21:25:56 INFO: Shutting down
[Q3FAA] 21:25:56 INFO: Exiting
```

It was especially important that pressing "CTRL+C" in the terminal continued to deliver the interrupt signal to syncthing. (The `exec` in the entrypoint script takes care of that.)

Then, ran with `-help`, and saw the help text:

```
% docker run sthing -help
Usage:
  syncthing [options]

Options:
...
```

Finally, with a bogus option to check that exit codes were passed on:

```
% docker run sthing -bogus >/dev/null
% echo $?
2
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,16 +17,11 @@ VOLUME ["/var/syncthing"]
 RUN apk add --no-cache ca-certificates su-exec
 
 COPY --from=builder /src/syncthing /bin/syncthing
+COPY --from=builder /src/script/docker-entrypoint.sh /bin/entrypoint.sh
 
 ENV PUID=1000 PGID=1000
 
 HEALTHCHECK --interval=1m --timeout=10s \
   CMD nc -z localhost 8384 || exit 1
 
-ENTRYPOINT \
-  chown "${PUID}:${PGID}" /var/syncthing \
-  && su-exec "${PUID}:${PGID}" \
-     env HOME=/var/syncthing \
-     /bin/syncthing \
-       -home /var/syncthing/config \
-       -gui-address 0.0.0.0:8384
+ENTRYPOINT ["/bin/entrypoint.sh", "-home", "/var/syncthing/config", "-gui-address", "0.0.0.0:8384"]
--- a/script/docker-entrypoint.sh
+++ b/script/docker-entrypoint.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+set -eu
+
+chown "${PUID}:${PGID}" /var/syncthing \
+  && exec su-exec "${PUID}:${PGID}" \
+     env HOME=/var/syncthing \
+     /bin/syncthing \
+     -home /var/syncthing/config \
+     -gui-address http://0.0.0.0:8443/ \
+     "$@"
--- a/script/docker-entrypoint.sh
+++ b/script/docker-entrypoint.sh
@@ -5,7 +5,4 @@ set -eu
 chown "${PUID}:${PGID}" /var/syncthing \
   && exec su-exec "${PUID}:${PGID}" \
      env HOME=/var/syncthing \
-     /bin/syncthing \
-     -home /var/syncthing/config \
-     -gui-address http://0.0.0.0:8443/ \
-     "$@"
+     /bin/syncthing "$@"


AudriusButkevicius:
I think (and this is outdated memory) you can combine the two?

Exec for the binary name, entry point for arguments?
Because I think now users will need to specify the binary name (script name in this case) if they want to run with custom arguments.

calmh:
Extra params are appended to the ENTRYPOINT so this is fine in our case. (We will correctly handle `-home foo -home bar` as well and use `bar`, if  someone wants to override that for example.)

AudriusButkevicius:
Right, but I think the same is achieved with exec + entrypoint and does not require double arguments.

calmh:
But I don't think that's desirable anyhow. Lets say we did

```
ENTRYPOINT the-script-that-runs-syncthing
CMD ["-home", ... etc]
```

then any parameters given would replace the ones in CMD. But, that means that if I do `docker run syncthing -paused` I no longer get the correct -home or a working GUI (because no -gui-address). So appending args seems the only correct behavior to me?

AudriusButkevicius:
Ok, I guess if $HOME was setup properly this would not be necessery.

AudriusButkevicius:
Ok, fine.

calmh:
Kinda... But we've (I've) painted us into a corner by having an unusual `-home` location (not a dot-dir), to make overriding the place with a volume nicer.

AudriusButkevicius:
You can still pass home in the exec script. I just think its nicer to so `docker run syncthing -help` than `docker run syncthing entrypoint.sh -help`, which can still be achieves by exec + entrypoint, even if we provide double home and double gui-address

calmh:
But the proposed solution enables `docker run syncthing -help` perfectly? It's in the example in the PR description even.

tomjakubowski:
Sorry, I'm new to the review feature on GitHub and can't figure out how to reply directly to the comments you've left.

> Maybe tuck it in as script/docker-entrypoint.sh? (It being called just /bin/entrypoint.sh in the docker image is fine though.)

taken care of

re: the entrypoint: I'm not sure now why I duplicated the arguments between the `ENTRYPOINT` directive and the entrypoint script itself, which I think may have been an error. It would probably suffice to leave them only in the entrypoint script, and let the user override the entrypoint if they would like to avoid passing those arguments for whatever reason.

calmh:
I'd actually prefer them only in the Dockerfile, as that's closer to where one might look to see what the default parameters are and it's where they connect to the actual path used for the volume etc. So yeah; drop those from the entry point script, and then it looks good to me.

PR #5643, opened 2019-04-10
lib/model: Pause pull for at least as long as failed pull took (fixes #5641)

### LINKED ISSUE #5641: Syncthing gets stuck in a high CPU "Failed Files" state - Possibly a symlink issue?
Syncthing Version: v1.1.1
OS Version: Windows 7 & OSX 10.9 & Ubuntu 18.04

# What I did

I've set up Syncthing on three machines - my desktop and my laptop that I need to sync files between, and a dedicated AWS EC2 instance for backups and so that stuff still get synced if both of my computers aren't on at the same time. (They usually aren't.)

First I set it up on the AWS instance, then I set it up on my Macbook. I set things synchronizing and after a couple of hiccups I had a couple folders duplicated on the dedicated server. (Per hiccups: Syncing slows down massively when the GUI is open?? I mean, I get why this could happen, but some kind of notice in the GUI would have been really helpful, I spent like an hour troubleshooting why network would be having problems that could explain the slow transfer rate when in fact it all had to do with the GUI.)

Then I set it up on Windows. I have been using Dropbox up to now, and the reason I'm setting up Syncthing is because Dropbox is ending support for my Macbook's OS. This means that all the files I need to sync were already being synchronized by Dropbox. I shared the smaller of the two folders I've been trying out with Syncthing between all the devices and everything was great. Syncthing recognized right away that everything was already synchronized. Then I shared the larger one and things were fine at first, but then my Macbook and the AWS instance both started getting confused.

# What happened next

Except for when I pause the larger folder being shared, both my Macbook and the AWS instance are reporting very high CPU usage. The GUI shows the folder in a "Failed Files" state. Syncthing seemed to be repeatedly scanning the folder. I did _not_ have this problem when sharing the folder from my Macbook with the AWS instance, this issue only occurred after adding my Windows machine to the bunch.

# The log

Here is a portion of the very noisy log from syncthing on OSX. The log's frequent mention of symlinks is the reason why I suspect the issue to be connected to symlinks.

```
[NDTXR] 23:08:59 INFO: Puller (folder "Projects" (mxwe3-xw9bp), file "PyDwarf/df/libs/SDL_ttf.framework/Headers/SDL_ttf.h"): checking parent dirs: traverses symlink: PyDwarf/df/libs/SDL_ttf.framework/Headers
[NDTXR] 23:08:59 INFO: Puller (folder "Projects" (mxwe3-xw9bp), file "PyDwarf/df/libs/SDL_ttf.framework/Resources/Info.plist"): checking parent dirs: traverses symlink: PyDwarf/df/libs/SDL_ttf.framework/Resources
[NDTXR] 23:08:59 INFO: Puller (folder "Projects" (mxwe3-xw9bp), file "PyDwarf/df/libs/SDL_ttf.framework/Versions/Current/Headers/SDL_ttf.h"): checking parent dirs: traverses symlink: PyDwarf/df/libs/SDL_ttf.framework/Versions/Current
[NDTXR] 23:08:59 INFO: Puller (folder "Projects" (mxwe3-xw9bp), file "PyDwarf/df/libs/SDL_ttf.framework/Versions/Current/Resources/Info.plist"): checking parent dirs: traverses symlink: PyDwarf/df/libs/SDL_ttf.framework/Versions/Current
[NDTXR] 23:08:59 INFO: Puller (folder "Projects" (mxwe3-xw9bp), file "PyDwarf/df/libs/SDL.framework/Versions/Current/Headers/SDL_syswm.h"): checking parent dirs: traverses symlink: PyDwarf/df/libs/SDL.framework/Versions/Current
[NDTXR] 23:08:59 INFO: Puller (folder "Projects" (mxwe3-xw9bp), file "rust/hello_world/target/debug/hello_world.dSYM/Contents/Resources/DWARF/hello_world-149ec8b9da1f89ba"): checking parent dirs: traverses symlink: rust/hello_world/target/debug/hello_world.dSYM
[NDTXR] 23:08:59 INFO: Puller (folder "Projects" (mxwe3-xw9bp), file "rust/hello_world/target/debug/hello_world.dSYM/Contents/Info.plist"): checking parent dirs: traverses symlink: rust/hello_world/target/debug/hello_world.dSYM
```

imsodin:
https://github.com/syncthing/syncthing/issues/5641#issuecomment-481227444

> Potentially the high cpu usage is due to a discrepancy between the pauses between successive failed tries and how long one try takes: If it takes e.g. 10min, you get 10min activity, 1min pause, 10min activity, 2min pause, 10min activity, 4min pause, ... - so essentially lots of activity. I'd propose to time the try and set the pause to at least that time (capped at 1h). If anything happens (scan, remote changes) that triggers a pull anyway, so you shouldn't "miss out" on a resolution like this. Does this sound reasonable?

The change means, that if e.g. a pull took 30min and failed, the pause will be immediately 32min instead of just 1min. Or if it took longer than an hour, the pause will immediately be 1h (well 64min), but never more than that.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -132,6 +132,23 @@ func (f *folder) Serve() {
 
 	initialCompleted := f.initialScanFinished
 
+	pull := func() {
+		startTime := time.Now()
+		if f.puller.pull() {
+			// We're good. Don't schedule another pull and reset
+			// the pause interval.
+			pause = f.basePause()
+			return
+		}
+		// Pulling failed, try again later.
+		delay := pause + time.Since(startTime)
+		l.Infof("Folder %v isn't making sync progress - retrying in %v.", f.Description(), delay)
+		pullFailTimer.Reset(delay)
+		if pause < 60*f.basePause() {
+			pause *= 2
+		}
+	}
+
 	for {
 		select {
 		case <-f.ctx.Done():
@@ -143,24 +160,10 @@ func (f *folder) Serve() {
 			case <-pullFailTimer.C:
 			default:
 			}
-
-			startTime := time.Now()
-			if !f.puller.pull() {
-				// Pulling failed, try again later.
-				pause = f.resetPullFailTimer(pullFailTimer, pause, time.Since(startTime))
-			}
+			pull()
 
 		case <-pullFailTimer.C:
-			startTime := time.Now()
-			if f.puller.pull() {
-				// We're good. Don't schedule another fail pull and reset
-				// the pause interval.
-				pause = f.basePause()
-				continue
-			}
-
-			// Pulling failed, try again later.
-			pause = f.resetPullFailTimer(pullFailTimer, pause, time.Since(startTime))
+			pull()
 
 		case <-initialCompleted:
 			// Initial scan has completed, we should do a pull
@@ -693,22 +696,6 @@ func (f *folder) basePause() time.Duration {
 		return defaultPullerPause
 	}
 	return time.Duration(f.PullerPauseS) * time.Second
-}
-
-// resetPullFailTimer reschedules pull retrying with increasingly longer pauses
-// between tries (with lower limit based on pull duration and fix upper limit).
-func (f *folder) resetPullFailTimer(timer *time.Timer, pause, pullDur time.Duration) time.Duration {
-	max := 60 * f.basePause()
-	// Pause at least as long as the time the previous (failed) pull took
-	for pullDur > pause && pause < max {
-		pause *= 2
-	}
-	l.Infof("Folder %v isn't making sync progress - retrying in %v.", f.Description(), pause)
-	timer.Reset(pause)
-	if pause > max {
-		return pause
-	}
-	return 2 * pause
 }
 
 func (f *folder) String() string {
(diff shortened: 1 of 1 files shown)


lib/model/folder.go
```diff
@@ -698,6 +695,22 @@ func (f *folder) basePause() time.Duration {
 	return time.Duration(f.PullerPauseS) * time.Second
 }
 
+// resetPullFailTimer reschedules pull retrying with increasingly longer pauses
+// between tries (with lower limit based on pull duration and fix upper limit).
+func (f *folder) resetPullFailTimer(timer *time.Timer, pause, pullDur time.Duration) time.Duration {
+	max := 60 * f.basePause()
+	// Pause at least as long as the time the previous (failed) pull took
+	for pullDur > pause && pause < max {
+		pause *= 2
```
calmh: This method is confusing. There are multiple opportunities to return a pause > max and in some cases we multiply the pause by four in one pass... 

I think it would be simpler to keep the easy sequence on `pause` (1, 2, 4 etc minutes up to max) and always sleep `pullDur + pause`?

imsodin:
@calmh Seems reasonable and indeed a lot simpler -> done. To prevent duplication and as `select` doesn't support `fallthrough`, I put that code in a function. Whether as a function variable (now) or as a method passing `pause` in and out seems pretty insignificant to me - do you have any preferences?

PR #5644, opened 2019-04-10
lib/fs: Add test that symlinks are skipped on walk

imsodin:
Just want to use CI to check a hunch.


imsodin:
My hunch was (luckily) wrong, at least a little test coverage came out of it :)

Issue #4170, opened 2017-05-23
panic: bug: ClusterConfig called on closed or nonexistent connection 


eyeson:
I have had two different Syncthing instances crash on Windows 10. One is running v0.14.28, one v.0.14.29-rc1.
Both are using SyncTrayzor (V1.1.16).

I restarted one, only to have it crash ~5 mins later. I captured the Synctrazor log from that one and have attached it.

I will endeavour to provide any other information required.

Forum post for reference: [https://forum.syncthing.net/t/syncthing-crashing/9845](url)

[SyncTrayzor.txt](https://github.com/syncthing/syncthing/files/1021902/SyncTrayzor.txt)

calmh:
For clarity:

```
[5NSF2] 09:17:06 INFO: Failed to exchange Hello messages with K2ECF46 (95.97.92.221:22067): read tcp 192.168.1.50:61924->95.97.92.221:22067: i/o timeout 
entWatcher: Received 0 events 
panic: bug: ClusterConfig called on closed or nonexistent connection 
goroutine 4242 [running]: 
github.com/syncthing/syncthing/lib/model.(*Model).ClusterConfig(0xc042058280, 0xdbf211def3228856, 0xc92a31f0bf8e42e1, 0xd87e29914adce15e, 0x87221d62cccbc3d9, 0xc042631f00, 0x1, 0x1) 
	c:/jenkins/workspace/syncthing-release-windows/src/github.com/syncthing/syncthing/lib/model/model.go:789 +0x21c3 
github.com/syncthing/syncthing/lib/protocol.(*nativeModel).ClusterConfig(0xc042b91fe0, 0xdbf211def3228856, 0xc92a31f0bf8e42e1, 0xd87e29914adce15e, 0x87221d62cccbc3d9, 0xc042631f00, 0x1, 0x1) 
	<autogenerated>:74 +0x95 
github.com/syncthing/syncthing/lib/protocol.(*rawConnection).readerLoop(0xc0457315c0, 0x0, 0x0) 
	c:/jenkins/workspace/syncthing-release-windows/src/github.com/syncthing/syncthing/lib/protocol/protocol.go:306 +0x11de 
created by github.com/syncthing/syncthing/lib/protocol.(*rawConnection).Start 
	c:/jenkins/workspace/syncthing-release-windows/src/github.com/syncthing/syncthing/lib/protocol/protocol.go:171 +0x4a 
goroutine 1 [chan receive, 70 minutes]: 
[...]

PR #5645, opened 2019-04-11
lib/protocol: Send close msg on receive error

imsodin:
There are three related changes:
 - Send a close message with a reason when an error occurs with incoming messages.
 - Wait for reader and writer loops to exit before signalling to the model that the connection was removed. This could in principle be the reason for the cluster-config-panic (https://github.com/syncthing/syncthing/issues/4170), if the cluster config comes in just before the connection is closed and parsing it takes longer than closing the connection.
 - Don't send close messages via the writer loop, but wait for that to stop and then send the close message directly. This has the advantage that if there are outstanding send operations, the close message doesn't have to "compete" with those to get sent.

### DIFF
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -184,8 +184,9 @@ type rawConnection struct {
 
 	outbox        chan asyncMessage
 	closed        chan struct{}
+	readerStopped chan struct{}
+	writerStopped chan struct{}
 	closeOnce     sync.Once
-	sendCloseOnce sync.Once
 	compression   Compression
 }
 
@@ -220,15 +221,17 @@ func NewConnection(deviceID DeviceID, reader io.Reader, writer io.Writer, receiv
 	cw := &countingWriter{Writer: writer}
 
 	c := rawConnection{
-		id:          deviceID,
-		name:        name,
-		receiver:    nativeModel{receiver},
-		cr:          cr,
-		cw:          cw,
-		awaiting:    make(map[int32]chan asyncResult),
-		outbox:      make(chan asyncMessage),
-		closed:      make(chan struct{}),
-		compression: compress,
+		id:            deviceID,
+		name:          name,
+		receiver:      nativeModel{receiver},
+		cr:            cr,
+		cw:            cw,
+		awaiting:      make(map[int32]chan asyncResult),
+		outbox:        make(chan asyncMessage),
+		closed:        make(chan struct{}),
+		readerStopped: make(chan struct{}),
+		writerStopped: make(chan struct{}),
+		compression:   compress,
 	}
 
 	return wireFormatConnection{&c}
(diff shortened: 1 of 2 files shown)


lib/protocol/protocol.go
```diff
@@ -813,44 +819,50 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
+		c.stopAndCleanup(err)
+
 		done := make(chan struct{})
-		c.send(&Close{err.Error()}, done)
-		select {
-		case <-done:
-		case <-c.closed:
-		}
-	})
+		c.writeMessage(asyncMessage{&Close{err.Error()}, done})
```
golangcibot: Error return value of `c.writeMessage` is not checked (from `errcheck`)

lib/protocol/protocol.go
```diff
@@ -813,44 +819,50 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
-			}
-		}
-		c.awaitingMut.Unlock()
+		c.stopAndCleanup(err)
 
 		c.receiver.Closed(c, err)
 	})
 }
 
+// stopAndCleanup must only be hold from within c.closeOnce
```
calmh: "only be called"?

lib/protocol/protocol.go
```diff
@@ -813,44 +819,50 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
-
-		c.awaitingMut.Lock()
-		for i, ch := range c.awaiting {
-			if ch != nil {
-				close(ch)
-				delete(c.awaiting, i)
-			}
-		}
-		c.awaitingMut.Unlock()
+		c.stopAndCleanup(err)
```
calmh: This will deadlock, because internalClose is called from the writer loop and stopAndCleanup will wait for the writer loop to exit... I'm not sure that's the only deadlock possibility, I didn't look closely, but I'm suspicious in general :)

lib/protocol/protocol.go
```diff
@@ -813,42 +817,52 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
 		}
-		c.awaitingMut.Unlock()
+	}
+	c.awaitingMut.Unlock()
 
-		c.receiver.Closed(c, err)
-	})
+	if send {
+		done := make(chan struct{})
+		c.writeMessage(asyncMessage{&Close{err.Error()}, done})
```
golangcibot: Error return value of `c.writeMessage` is not checked (from `errcheck`)

lib/protocol/protocol.go
imsodin: Right, I assumed (...) that once `sync.Once.Do` was invoked, any subsequent calls will just skip it, but documentation says your suspicion is right, it blocks all subsequent calls until it returns. I didn't see an obvious way to fix this with `Once`, so I did it with known means of a channel and a lock.

imsodin:
Closed (for now), as parts were cherry-picked to #5657. I'll readdress this when that is done.

PR #5646, opened 2019-04-11
lib/protocol: Don't send any messages before cluster config

imsodin:
This is rather pedantic protocol level race, but who knows, maybe someone got bitten by it anyway at some point:  
Cluster-config must be the first message sent, but there is a tiny window in `AddConnection` between releasing `pmut` and sending the message, in which the puller (which can be up and running hard) can send a request on that connection. If that happens, the remote side will close the connection (silently for now, resulting in a timeout eventually - see https://github.com/syncthing/syncthing/pull/5645).

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1766,6 +1766,7 @@ func (m *model) GetHello(id protocol.DeviceID) protocol.HelloIntf {
 // AddConnection adds a new peer connection to the model. An initial index will
 // be sent to the connected peer, thereafter index updates whenever the local
 // folder changes.
+// Must not be called concurrently.
 func (m *model) AddConnection(conn connections.Connection, hello protocol.HelloResult) {
 	deviceID := conn.ID()
 	device, ok := m.cfg.Device(deviceID)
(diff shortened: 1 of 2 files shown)


calmh:
Also maybe #4170?

lib/model/model.go
```diff
@@ -1820,10 +1825,7 @@ func (m *model) AddConnection(conn connections.Connection, hello protocol.HelloR
 	l.Infof(`Device %s client is "%s %s" named "%s" at %s`, deviceID, hello.ClientName, hello.ClientVersion, hello.DeviceName, conn)
 
 	conn.Start()
-	m.pmut.Unlock()
 
-	// Acquires fmut, so has to be done outside of pmut.
-	cm := m.generateClusterConfig(deviceID)
 	conn.ClusterConfig(cm)
```
calmh: Only thing that makes me a little but uneasy about this is that ClusterConfig() is a potentially blocking call... Although, at this point, the writer loop should be up since Start() and be reading the outbox channel that ClusterConfig() needs to send on, so hopefully it's fine.

imsodin: Good point, in the worst case it blocks `pmut` for as long as the timeout of the underlying connection. Alternative idea: First check and drop old connection holding `pmut`, then release it and already start and send cluster config on the new connection before acquiring `pmut` again and adding the connection to the model. This relies on `AddConnection` not being called concurrently, but that's already the case today. I'll add a comment to that effect, if we want to enfore/ensure it, we'd need to add another lock.

imsodin:
DO NOT MERGE

There really should be an option for a PR author to do something akin of a review with request for changes. I'll follow up with info on why in an edit.

imsodin:
I added `model.connReady` (`map[protocol.DeviceID]struct{}`) which is set once the cluster-config was sent. Before that `requestGlobal` will abort and `Availability` will not include that device.

AudriusButkevicius:
Can we do something more sensible and make this a property of the connection (or even logic in the connection, where trying to send any other message than cluster config returns errNotReady) rather than wedging more stuff on the model?

imsodin:
@AudriusButkevicius I changed it to let all messages to send wait until the cluster-config was sent in `lib/protocol`.

lib/protocol/protocol.go
```diff
@@ -322,9 +324,20 @@ func (c *rawConnection) Request(folder string, name string, offset int64, size i
...
 }
 
-// ClusterConfig send the cluster configuration message to the peer and returns any error
+// ClusterConfig sends the cluster configuration message to the peer.
+// It must be called just once (as per BEP).
 func (c *rawConnection) ClusterConfig(config ClusterConfig) {
-	c.send(&config, nil)
+	select {
+	case <-c.sentClusterConfig:
+		return
```
AudriusButkevicius: Why this? I think it might be fine to send multiple of these in the future.

imsodin: Not unless the BEP gets changed:

> A Cluster Config message MUST be the first post authentication message sent on a BEP connection. Additional Cluster Config messages MUST NOT be sent after the initial exchange.

Also the code as it is would not work at all if multiple calls to `ClusterConfig` were allowed, as it sends outside of the send loop and sends must never happen concurrently. Of course this can be done differently, it's just a bit more complicated and in my opinion unnecessary due to BEP.

AudriusButkevicius: Fine

PR #5648, opened 2019-04-12
lib/model: Fix block index calculation for recheckFile (fixes #5649)

### LINKED ISSUE #5649: Checking block content on unexpected request results doesn't work for the last block
Detected in https://forum.syncthing.net/t/pull-no-such-file-error/13139/22

This is a super corner case:

1. Program allocates space and thus sets modtime.
2. Syncthing scans.
3. Program does the actual change on disk, that only affects the last block of the file.
4. Peer requests that last block.

Result: Block index too high error.
Expected: File gets force rescanned.

imsodin:
This is the result of https://forum.syncthing.net/t/pull-no-such-file-error/13139/22

Currently the block index is calculated from requested offset and size, but that's not correct for the last block, which usually is smaller than the file's blocksize.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1576,7 +1576,7 @@ func (m *model) Request(deviceID protocol.DeviceID, folder, name string, size in
 	}
 
 	if !scanner.Validate(res.data, hash, weakHash) {
-		m.recheckFile(deviceID, folderFs, folder, name, int(offset)/int(size), hash)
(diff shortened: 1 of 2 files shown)


calmh:
bug ticket on this plz

imsodin:
@calmh done: #5649. Title is longish, but I really couldn't come up with anything shorter that still contains some information (as opposed to "file doesn't sync in corner case") :)

PR #5657, opened 2019-04-16
lib/model, lib/protocol: Wait for reader/writer loops on close (fixes #4170)

### LINKED ISSUE #4170: panic: bug: ClusterConfig called on closed or nonexistent connection

imsodin:
### Purpose

Fix the infamous cluster config panics #4170.

### Testing

None, as the panic isn't reliably reproducible, but it am confident xD


AudriusButkevicius:
Wait group might be more simple?

AudriusButkevicius:
I wonder if there is no deadlocks here.

Reader blocks on delivering a message, which tries to acquire a pmut/fmut in the model, and the conn.Close() happens under pmut/fmut which is blocked on the wg.wait returning, which is blocked on reader loop exiting.

imsodin:
That's indeed a possibility. Solution: Don't hold the locks in model while closing.

And because I couldn't help myself I cleaned up a little bit of legacy code in the touched areas.

lib/model/model.go
```diff
@@ -1412,21 +1407,21 @@ func (m *model) Closed(conn protocol.Connection, err error) {
...
 		// There is no connection to close
 		return
 	}
-
 	conn.Close(err)
 }
 
+func (m *model) closeConnToDevices(fcfg config.FolderConfiguration, err error) {
+	for _, dev := range fcfg.DeviceIDs() {
+		m.close(dev, err)
```
AudriusButkevicius: This acquires pmut repeatedly.

lib/model/model.go
```diff
@@ -1412,21 +1407,21 @@ func (m *model) Closed(conn protocol.Connection, err error) {
...
+	m.pmut.Unlock()
 	if !ok {
 		// There is no connection to close
 		return
 	}
-
 	conn.Close(err)
 }
 
+func (m *model) closeConnToDevices(fcfg config.FolderConfiguration, err error) {
```
AudriusButkevicius: This could just take []DeviceID as that is what DeviceIDs() returns.

PR #5671, opened 2019-04-28
cmd, lib/db: Actually close goleveldb (fixes #5505)

### LINKED ISSUE #5505: Leaking goroutines through model and leveldb
While running a model test with `-count 1000` I ran into OOM. Looking at a stacktrace I found these culprits:

```
    340 github.com/syncthing/syncthing/lib/model.(*ProgressEmitter).Serve
    340 github.com/syndtr/goleveldb/leveldb.(*DB).compactionError
    340 github.com/syndtr/goleveldb/leveldb.(*DB).mCompaction
    340 github.com/syndtr/goleveldb/leveldb.(*DB).mpoolDrain
    340 github.com/syndtr/goleveldb/leveldb.(*DB).tCompaction
    340 github.com/syndtr/goleveldb/leveldb/util.(*BufferPool).drain
```

The progress emitter is straight forward: It should be added to the supervisor not manually run in a goroutine. The leveldb stuff is because we do not close the the database, both in tests and in cmd/syncthing's `main`. In `main` the database is manipulated before passing it to the model, but not concurrently with the model. So there's two options

1. Close the db on `model.Close`: Nice because we don't need to think about it in all the tests.
2. Close the db where it is created, meaning in cmd/syncthing's `main` and every single test that uses a db. The "cleaner" approach because the same function that opens it also closes it but it's tedious.

Opinions? @AudriusButkevicius @calmh

imsodin:
We currently don't close the database, just call `os.Exit`. At the same time we do not keep track of whatever is currently accessing the database (model, api, ...), which is fine in many cases, but not for a few random `panic`s we have in our code and `goleveldb.DB.NewIterator` (https://github.com/syndtr/goleveldb/issues/235, other functions just report `goleveldb.ErrClosed`). Therefore I had to add some bubblewrap to prevent panics after close.

This does not ensure that we are in a consistent state on close, but it at least explicitly closes the database instead of crashing it. Doing that would probably require actually keeping track of everything the model does and adding `model.Close` that ensures everything is terminated before it returns - I'd say that would be quite a lot of work.

### DIFF
--- a/lib/db/instance.go
+++ b/lib/db/instance.go
@@ -506,6 +506,9 @@ func (db *instance) getIndexID(device, folder []byte) protocol.IndexID {
 func (db *instance) setIndexID(device, folder []byte, id protocol.IndexID) {
 	bs, _ := id.Marshal() // marshalling can't fail
 	if err := db.Put(db.keyer.GenerateIndexIDKey(nil, device, folder), bs, nil); err != nil {
+		if err == leveldb.ErrClosed {
+			return
+		}
 		panic("storing index ID: " + err.Error())
 	}
 }
--- a/lib/db/instance.go
+++ b/lib/db/instance.go
@@ -505,10 +505,7 @@ func (db *instance) getIndexID(device, folder []byte) protocol.IndexID {
 
 func (db *instance) setIndexID(device, folder []byte, id protocol.IndexID) {
 	bs, _ := id.Marshal() // marshalling can't fail
-	if err := db.Put(db.keyer.GenerateIndexIDKey(nil, device, folder), bs, nil); err != nil {
-		if err == leveldb.ErrClosed {
-			return
-		}
+	if err := db.Put(db.keyer.GenerateIndexIDKey(nil, device, folder), bs, nil); err != nil && err != leveldb.ErrClosed {
 		panic("storing index ID: " + err.Error())
 	}
 }
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -9,12 +9,15 @@ package db
 import (
 	"os"
 	"strings"
+	"sync"
 	"sync/atomic"
 
 	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/errors"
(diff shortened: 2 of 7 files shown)


lib/db/lowlevel.go
```diff
@@ -34,6 +37,8 @@ type Lowlevel struct {
 	location  string
 	folderIdx *smallIndex
 	deviceIdx *smallIndex
+	closeMut  *sync.RWMutex
```
calmh: this could as well be a boolean given that we only look at it from within the mutex?

lib/db/instance.go
```diff
@@ -506,6 +506,9 @@ func (db *instance) getIndexID(device, folder []byte) protocol.IndexID {
 func (db *instance) setIndexID(device, folder []byte, id protocol.IndexID) {
 	bs, _ := id.Marshal() // marshalling can't fail
 	if err := db.Put(db.keyer.GenerateIndexIDKey(nil, device, folder), bs, nil); err != nil {
+		if err == leveldb.ErrClosed {
+			return
+		}
```
AudriusButkevicius: Why is it needed only here?

imsodin: Because in other places our error checking amounts to something like
```
	if err != nil {
		l.Debugln("surprise error:", err)
		return nil
	}
```
which I consider an adequate reaction to being closed and doesn't need more specific checks (and it doesn't result in a panic).

lib/db/lowlevel.go
AudriusButkevicius: Presumably having a channel here would be easier? Single member variable?

imsodin: "Unprotected" channel is still racy:  

1. `NewIterator` steps past the closed check.  
2. `Close` closes the channel.  

Result: `DB.Close` must happen after `DB.NewIterator`, but may happen before.

lib/db/instance.go
calmh: There's also stuff like db writes where we just straight up panic on any error

imsodin: Now all panics in lib/db are either removed or have a check for close errors (or are not related to goleveldb in the first place).

PR #5675, opened 2019-04-29
lib/protocol: Don't close asyncMessage.done on success (fixes #5674)

### LINKED ISSUE #5674: panic: close of closed channel
Hi,

I'm using syncthing from git (amongst others on other nodes).  I'm getting:

panic: close of closed channel

on a moderately reproducible basis.  I did a git bisect between:

`3bea59b0d9eb9168619811e822ddf421db5b811b`

and

`0ca1f26ff8c01bf7d93ffdb4d375db32863ec215`

The commit that introduced these panics was:

`04b927104fe2983875016fe565c352ee9517036a`

The subsequent changes made after this perhaps make the panic more difficult to trigger, but reliable enough for a git bisection.

I'm running the github version on Linux x86_64:
Linux Kernel: 4.19.30
Distro: Slackware current
glibc: 2.29
Go: 1.12.4
syncthing v1.1.2-rc.1+23-g0ca1f26f "Erbium Earthworm" (go1.12.4 linux-amd64)

Syncthing cluster: 6 devices, including android, virtualised win10, and macosx.  All other systems are running release versions.

Steps to reproduce the problem.
1) Compile syncthing.
2) Start syncthing.
3) Connect to cluster (already up).
4) Kill syncthing on Mac or Android, then restart.
5) Panic appears on Linux client.

Sorry if bug reports should only be against releases.

Thanks for syncthing.  A very lovely bit of coding.

===Rich

imsodin:
Regression introduced in #5646 as identified by @rahrah in #5674, thus luckily not yet released.

### DIFF
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -641,9 +641,9 @@ func (c *rawConnection) handleResponse(resp Response) {
 	c.awaitingMut.Unlock()
 }
 
-func (c *rawConnection) send(msg message, done chan struct{}) bool {
+func (c *rawConnection) send(msg message, done chan struct{}) (sent bool) {
 	defer func() {
-		if done != nil {
+		if !sent && done != nil {
 			close(done)
 		}
 	}()


lib/protocol/protocol.go
```diff
@@ -641,9 +641,9 @@ func (c *rawConnection) handleResponse(resp Response) {
 	c.awaitingMut.Unlock()
 }
 
-func (c *rawConnection) send(msg message, done chan struct{}) bool {
+func (c *rawConnection) send(msg message, done chan struct{}) (sent bool) {
 	defer func() {
-		if done != nil {
+		if !sent && done != nil {
```
calmh: Why not do the if & close just before the "return true" below, it's just one place and avoids the defer entirely?

PR #5679, opened 2019-05-01
lib/protocol: Fix yet another deadlock (fixes #5678)

### LINKED ISSUE #5678: Connection closing deadlocks on send error
This is a regression introduced in #5442, which ironically fixed a deadlock by introducing a new one. To reproduce run `TestReconnectSenderDuringTransfer` from the integration tests, it will time out (possibly others too).

imsodin:
Calling `internalClose` in `writerLoop` calls `wg.Wait` which waits for the deferred `wg.Done` also in `writerLoop`: face -> wall

### DIFF
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -392,22 +392,22 @@ func (c *rawConnection) readerLoop() error {
 	}()
 
 	state := stateInitial
-	var errMsg messageWithError
+	var msgWithErr messageWithError
 	for {
 		select {
-		case errMsg = <-inbox:
+		case msgWithErr = <-inbox:
 		case <-c.closed:
 			return ErrClosed
 		}
-		if errMsg.err != nil {
-			if errMsg.err == errUnknownMessage {
+		if msgWithErr.err != nil {
+			if msgWithErr.err == errUnknownMessage {
 				// Unknown message types are skipped, for future extensibility.
 				continue
 			}
-			return errMsg.err
+			return msgWithErr.err
 		}
 
-		switch msg := errMsg.msg.(type) {
+		switch msg := msgWithErr.msg.(type) {
 		case *ClusterConfig:
 			l.Debugln("read ClusterConfig message")
 			if state != stateInitial {
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -245,7 +245,10 @@ func (c *rawConnection) Start() {
 		err := c.readerLoop()
 		c.internalClose(err)
 	}()
-	go c.writerLoop()
+	go func() {
+		err := c.writerLoop()
+		c.internalClose(err)
+	}()
 	go c.pingSender()
 	go c.pingReceiver()
 }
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
The same issue in ping receiver

imsodin:
Yeah right. And also `c.readMessage` is a blocking call, so that shouldn't be done under wait-group supervision either (see https://github.com/syncthing/syncthing/pull/5679/commits/ba05c0e3bbffe3fced25d5f803337b85749b8b5c).

lib/protocol/protocol.go
```diff
@@ -363,25 +370,44 @@ func (c *rawConnection) ping() bool {
 	return c.send(&Ping{}, nil)
 }
 
-func (c *rawConnection) readerLoop() (err error) {
-	defer c.wg.Done()
+type errorMessage struct {
```
AudriusButkevicius: readResponse? This feels like an actual error message not the conjunction of two types independently.

lib/protocol/protocol.go
```diff
@@ -363,25 +370,44 @@ func (c *rawConnection) ping() bool {
...
+		}
+	}()
+
 	state := stateInitial
+	var errMsg errorMessage
 	for {
-		if c.Closed() {
+		select {
+		case errMsg = <-inbox:
+		case <-c.closed:
```
AudriusButkevicius: Shouldn't we just wait for readMessage to return ErrClosed or whatever, to make sure it exits?
Also, who closes inbox?

imsodin: That's exactly what we can't do: If e.g. the model closes the connection, `readMessage` won't return until `internalClose`/`wg.Wait` returns. Ideally `io.Reader` would have an equivalent with channels, but unfortunately it doesn't.
And nobody closes `inbox` currently, I can make the `readMessage` goroutine close it, but it's irrelevant as the receiver already terminated at that point.

lib/protocol/protocol.go
imsodin: Ou, I see how that's confusing now (I went `asyncMessage` -> `errorMessage` in my head). `readResponse` also is a bit ambiguous, as there is a `Response` message that gets read. Maybe `messageWithError`?

AudriusButkevicius: Yeah

imsodin:
> LGTM (modulo naming as pointed out) although significant changes in the protocol package are scary. :)

Totally agree and the bandwaggon of followup PRs that I am producing kinda hardens the point further. I still intend to do even more (related) changes after this o.O  
I hope you have the patience (and some confidence) left, I truly believe them to be worth it.

lib/protocol/protocol.go
```diff
@@ -363,25 +370,44 @@ func (c *rawConnection) ping() bool {
...
 		}
-
-		msg, err := c.readMessage(fourByteBuf)
-		if err == errUnknownMessage {
-			// Unknown message types are skipped, for future extensibility.
-			continue
-		}
-		if err != nil {
-			return err
+		if errMsg.err != nil {
```
AudriusButkevicius: Can we also rename this variable to msgWithErr or something?

PR #5680, opened 2019-05-01
lib/model, lib/osutil: Reduce Lstats when pulling

imsodin:
### Purpose

Currently for every pulled item, we stat every single path component (below root). That's clearly redundant and as we iterate in alphabetic order, it's easy to optimize. To use this ordering, all `checkParent` calls need to be done in one place, therefore I had to refactor `processNeeded`.


lib/osutil/traversessymlink.go
```diff
@@ -35,19 +35,21 @@ func (e NotADirectoryError) Error() string {
...
-	if err != nil {
-		return err
-	}
+	return TraversesSymlinkSegment(filesystem, name, ".")
+}
 
+// TraversesSymlinkSegment returns an error if any path component of name below
+// parent (including name itself) traverses a symlink. Expects
+// `fs.IsParent(name, parent) == true`, otherwise behaviour is undefined.
+func TraversesSymlinkSegment(filesystem fs.Filesystem, name, parent string) error {
```
calmh: This IsParent requirement doesn't sound like it would hold for the dot we're passing in, i.e. `IsParent("whatever", ".") => false` I would expect? Is `filepath.Rel` cool with this?
 
(Also out parameter order in IsParent is the opposite of filepath.Rel which is at the very least confusing... But not your fault or a thing in this PR :)

Pavel8419:
привет всем
--
Отправлено из Mail.ru для Android четверг, 02 мая 2019г., 08:29 +03:00 от Jakob Borg  notifications@github.com :

lib/model/folder_sendrecv.go
```diff
@@ -316,67 +317,79 @@ func (f *sendReceiveFolder) processNeeded(dbUpdateChan chan<- dbUpdateJob, copyC
...
 					// Put files into buckets per first hash
-					key := string(df.Blocks[0].Hash)
-					buckets[key] = append(buckets[key], df)
+					key := string(curFile.Blocks[0].Hash)
+					buckets[key] = append(buckets[key], curFile)
+					return true
 				}
+				f.deleteFileWithCurrent(file, curFile, hasCurFile, dbUpdateChan, scanChan)
+				return true
 			}
```
AudriusButkevicius: No changed++?

AudriusButkevicius:
In principle looks ok, apart from I think feeling that changed++ is missing in a few places.

Also, it feels like this is one of those "lets refactor everything" while introducing a small optimisation.

imsodin:
Ups.

AudriusButkevicius:
I do see however that there was a few fixes that fixed brokeness in the original code, and I think this is still manageable in terms of size so perhaps it's ok to have this PR.

I guess all I am saying I'd have more confidence to just yolo approve if the change was just that change, without the extras.

This is not blame, as I do this myself, just an observation.

AudriusButkevicius:
Nvm

lib/osutil/traversessymlink.go
imsodin: `IsParent` is aware that "." is special, there's even a test case confirming it.

imsodin:
> I do see however that there was a few fixes that fixed brokeness in the original code, and I think this is still manageable in terms of size so perhaps it's ok to have this PR.

Uhm, it's been a while since I originally coded this, so I might have just forgotten, but as far as I see/remember in my opinion this PR just optimises and makes some things more readable - I am not aware of anyfixed brokenness. Could you enlighten me?

lib/osutil/traversessymlink.go
calmh: Make a note of that in the comment please, otherwise it does look like we're saying you cannot do what we specifically do a couple of lines higher up.

AudriusButkevicius:
This adds a few changed++ and newPullError in a few places, so presumably this also squashes a few bugs.

lib/osutil/traversessymlink_test.go
```diff
@@ -70,6 +70,59 @@ func TestTraversesSymlink(t *testing.T) {
...
 
+func TestTraversesSymlinkSegment(t *testing.T) {
+	tmpDir, err := ioutil.TempDir("", "Syncthing-TestTraversesSymlinkSegment-")
+	if err != nil {
+		panic("Failed to create temporary testing dir")
+	}
+	defer os.RemoveAll(tmpDir)
+
+	fs := fs.NewFilesystem(fs.FilesystemTypeBasic, tmpDir)
+	fs.MkdirAll("a/b/c", 0755)
```
golangcibot: Error return value of `fs.MkdirAll` is not checked (from `errcheck`)

imsodin:
> Prove me wrong.

Cannot do. Added tests, proved you right and fixed it - thanks!

lib/model/folder_sendrecv.go
```diff
@@ -317,72 +318,89 @@ func (f *sendReceiveFolder) processNeeded(dbUpdateChan chan<- dbUpdateJob, copyC
...
-		case file.IsDeleted():
-			if file.IsDirectory() {
-				// Perform directory deletions at the end, as we may have
-				// files to delete inside them before we get to that point.
-				dirDeletions = append(dirDeletions, file)
-			} else {
-				fileDeletions[file.Name] = file
-				df, ok := f.fset.Get(protocol.LocalDeviceID, file.Name)
+		parent := filepath.Dir(file.Name)
+		if err := f.checkParent(parent, lastValidParent, scanChan); err != nil {
```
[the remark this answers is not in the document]
AudriusButkevicius: Yeah this is not good. Reduces lstats and sanity checks.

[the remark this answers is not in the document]
imsodin: > Yeah this is not good. Reduces lstats and sanity checks.

Sorry, I don't get what this is referring to specifically.

The file deletions issue is now fixed by moving handling deletes ahead of `checkParent`. 

The time gap is indeed potentially very large. It has already been potentially big before (think big file on slow connection). So if we want to care about that, I'd add a check in `performFinish` just before replacing the old file.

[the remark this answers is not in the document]
imsodin: What the heck is happening with the comment order?!

calmh:
GitHub has serious time travel issues with comments tonight

lib/model/folder_sendrecv.go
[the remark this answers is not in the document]
AudriusButkevicius: So the gap was not that big, as we've checked the parent as we've handled the file I believe, and then immediately opened the file (which made sure that the fd stayed in the same place).
If the underling directory moved, potentially the move to place would have failed.

imsodin:
Ok, so now this uses the optimization for dirs and links only, files are checked after popping from queue as before - now I guess it comes more down to whether you consider the "marginal, accompanying" worth it.

imsodin:
Actually lets not mix stuff, it might be a bit pedantic but the recent reverts made me lean more towards as small commits as possible. I made separate PRs: #5699 #5700

calmh:
So, this is not zero cost in terms of added code + complexity; what is the real world gain in a benchmark of some sort? (The `ManyFiles` integration test might be one such, that has previously shown impacts on stat/sync/similar stuff)

imsodin:
The many files benchmark (reduced to 10000 files to stay within 10min) does not show any significant difference. Therefore I am closing this - the parts split off into #5699 in my opinion have merit regardless of benchmarking (as they don't increase complexity in my opinion).

AudriusButkevicius:
If you comment out the fsync, what happens then?

PR #5682, opened 2019-05-02
lib/model: Let fakeConnection call Model.Closed on close

imsodin:
### Purpose

By making the fake connection callback into model through `Closed` we get closer to reality and add more coverage. It also uncovered one weird piece of code, by letting model panic when calling `Closed` on an already removed connection (which apparently doesn't happen in reality, but still is bad).  
I split this PR off another change that requires it (to be PRed). 

### Testing

Existing tests pass. One test that explicitly called `Closed` needed to be adjusted to now having that called automatically.


imsodin:
Unfortunately that test depends on the connection not getting closed (it not failing on other systems is just a race: test is faster than the connection being closed), so I can't enable it unconditionally.

lib/model/model_test.go
```diff
@@ -328,13 +328,21 @@ type fakeConnection struct {
 	model                    *model
 	indexFn                  func(string, []protocol.FileInfo)
 	requestFn                func(folder, name string, offset int64, size int, hash []byte, fromTemporary bool) ([]byte, error)
+	closeFn                  func(error)
 	mut                      sync.Mutex
 }
 
-func (f *fakeConnection) Close(_ error) {
+var fakeCloseErr = fmt.Errorf("fakeConnection was closed")
```
golangcibot: `fakeCloseErr` is unused (from `deadcode`)

lib/model/model_test.go
```diff
@@ -1139,6 +1145,16 @@ func TestIssue5063(t *testing.T) {
 	wcfg, m := newState(defaultAutoAcceptCfg)
 	defer os.Remove(wcfg.ConfigPath())
 
+	m.pmut.Lock()
+	for id, c := range m.conn {
+		conn := c.(*fakeConnection)
+		conn.mut.Lock()
+		conn.closeFn = func(_ error) {}
+		m.conn[id] = conn
```
AudriusButkevicius: You don't need this, as it's a pointer anyway.

lib/model/model_test.go
```diff
@@ -2660,20 +2676,6 @@ func TestSharedWithClearedOnDisconnect(t *testing.T) {
...
-	if _, ok := m.helloMessages[device2]; !ok {
-		t.Error("hello missing early")
-	}
-
-	if _, ok := m.deviceDownloads[device2]; !ok {
-		t.Error("downloads missing early")
-	}
-
-	m.Closed(conn2, fmt.Errorf("foo"))
-
```
AudriusButkevicius: Why is this being removed?

imsodin: That's now called automatically by the fake connection on Close (some lines above).

AudriusButkevicius: Ah right

PR #5684, opened 2019-05-02
lib/model: Consistent use of locks

imsodin:
In most places we already follow these practice, I adjusted the rest. This should be a trivial change that at best reduces time waited for locks and at worst just makes code more consistent, but as it is about locks please still have an actual good look at it.

Whenever possible:

- Release locks early:  
If possible just get whatever you need and then release before doing the processing.
- Defer unlocks:  
Just in case of early returns and panics

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -363,8 +363,8 @@ func (m *model) AddFolder(cfg config.FolderConfiguration) {
 	}
 
 	m.fmut.Lock()
+	defer m.fmut.Unlock()
 	m.addFolderLocked(cfg)
-	m.fmut.Unlock()
 }
 
 func (m *model) addFolderLocked(cfg config.FolderConfiguration) {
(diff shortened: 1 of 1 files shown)


calmh:
It's like juggling chainsaws but it looks OK to me

calmh:
(Merge at will or if Audrius wants to lend an extra pair of eyes)

lib/model/model.go
```diff
@@ -585,6 +585,8 @@ func (info ConnectionInfo) MarshalJSON() ([]byte, error) {
 func (m *model) ConnectionStats() map[string]interface{} {
 	m.fmut.RLock()
 	m.pmut.RLock()
+	defer m.pmut.RUnlock()
+	defer m.fmut.RUnlock()
```
AudriusButkevicius: This unlocks them in the wrong order?

lib/model/model.go
```diff
@@ -363,8 +363,8 @@ func (m *model) AddFolder(cfg config.FolderConfiguration) {
 	}
 
 	m.fmut.Lock()
+	defer m.fmut.Unlock()
 	m.addFolderLocked(cfg)
-	m.fmut.Unlock()
```
AudriusButkevicius: Defer prevents inlining, so I am not sure why we thing this is better...

lib/model/model.go
imsodin: How does unlock order matter?

lib/model/model.go
imsodin: I deferred it because of the function call, which incidentally also isn't inlineable: https://github.com/golang/go/wiki/CompilerOptimizations#function-inlining

PR #5685, opened 2019-05-02
 lib/protocol: Fix deadlock on sending close message 

imsodin:
Please bear with me, unfortunately it's necessary, due to the deadlock reproduced in this PR: Writing is also a blocking call, so by waiting on `writerLoop`, we can get into a deadlock. To summarize what I want to achieve (not only in this PR, in all those related PRs):

1. After calling `receiver.Closed`, more calls to `receiver` must not happen:  
Reason for the whole wait/channel business on close.
2. Send close messages to remotes to propagate close reasons and speed closing up.
3. Do not try to send a close message after an error occurred while sending (not much chance for it to succeed).
4. Do not block on sending the final close message (current timeout is 1s, maybe a bit short?).

This passes both unit and integration tests.

### DIFF
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -186,8 +186,7 @@ type rawConnection struct {
 	outbox            chan asyncMessage
 	closed            chan struct{}
 	closeOnce         sync.Once
-	sendCloseOnce     sync.Once
-	wg                sync.WaitGroup
+	readerStopped     chan struct{}
 	compression       Compression
 }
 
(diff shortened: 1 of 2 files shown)


calmh:
So, this is like the fourth deadlock fix after the initial don't-send-messages-before-clusterconfig fix, and the changes are large... I'm feeling it might be time to step back and re-evaluate that fix and the whole patch series since then. There was a bug, certainly, but not this deadlockiness that we have now in a very critical place in the code...

imsodin:
I get what your saying - and it's even a bit worse in that the change starting all of this was https://github.com/syncthing/syncthing/commit/24ffd8be99bf21ade538fde04f1ff2bded06c53f. Luckily it was quite straight forward to revert all the relevant commits and the end result is "working" (unit tests pass): https://github.com/syncthing/syncthing/compare/master...imsodin:protocolRevert

There's some unrelated or non-critical changes between the revert and this PR, I'll try to apply them soon, so we get a clean diff. Then we can either work/review from there or just commit the revert and go on from there eventually.

PR #5688, opened 2019-05-03
lib/protocol: Revert unreleased changes related to closing connections

imsodin:
As brought up by @calmh in the most recent PR (#5685) in a string of PRs related to handling closing of protocol connections, the way I introduced these changes wasn't particularly trust inspiring and the affected code is critical.

There were three involved changes, with follow-up fixes each:

1. Propagating closing of connection in model to protocol: #5440 -> #5442 
2. Sending cluster configs first: #5646 -> #5675 
3. Do not call the model after calling `*model.Closed` (fixes cluster config panic #4170): #5657 -> #5679 (unmerged: #5685)

My proposal for the next RC is to revert these changes (0229f04) and redo one unrelated change (337ae59). This brings us back to a state, that has been in place for a long time. There are known problems, but we were able to live with them for a long time, so we can again. I do intend to tackle this problems again for the next RC cycle. If you agree, I'll PR both commits separately, as I believe it makes for a cleaner history.

### DIFF
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -182,13 +182,11 @@ type rawConnection struct {
 	nextID    int32
 	nextIDMut sync.Mutex
 
-	sentClusterConfig chan struct{}
-	outbox            chan asyncMessage
-	closed            chan struct{}
-	closeOnce         sync.Once
-	sendCloseOnce     sync.Once
-	wg                sync.WaitGroup
-	compression       Compression
+	outbox        chan asyncMessage
+	closed        chan struct{}
+	closeOnce     sync.Once
+	sendCloseOnce sync.Once
+	compression   Compression
 }
 
 type asyncResult struct {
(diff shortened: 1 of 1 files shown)


calmh:
You're reverting quite far back now... The ones I would unwind are 9f358eca, ec7c88ca, 19b51c9b, 5da41f75, 04b92710 which is + 9f358eca and - 24ffd8be - c8741185 as those two are already included in releases.

imsodin:
Do you really want to revert 9f358ec too? That one has nothing to do with anything else, that's just code refactoring/hardening.

calmh:
I was thinking it was on top of the other changes, but if it's entirely standalone then no, no need to revert that

imsodin:
Ok, one more question: With the recent extension of tests (https://github.com/syncthing/syncthing/pull/5682) fully reverting https://github.com/syncthing/syncthing/commit/5da41f75fab1b8cf07907789c738c59c960ace4b doesn't pass model tests anymore. What about keeping the `model.go` part of that commit?

calmh:
Use good judgement, no need to revert for revertings sake, just to back out the hole we were digging in protocol, to dig a better one later :)

PR #5690, opened 2019-05-06
lib/model: Remove individual pull errors on retry (fixes #5659)

### LINKED ISSUE #5659: Reset individual errors whenever processing the item
See https://forum.syncthing.net/t/cosmetic-problem-pull-connection-closed/13177
Ping @imsodin

imsodin:
This removes transient pull errors when a second iteration is run, instead of keeping them until all potentially very long sync operations are done.  

Fixes #5659

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -310,6 +310,7 @@ func (f *sendReceiveFolder) processNeeded(dbUpdateChan chan<- dbUpdateJob, copyC
 		}
 
 		if f.IgnoreDelete && intf.IsDeleted() {
+			f.resetPullError(intf.FileName())
 			l.Debugln(f, "ignore file deletion (config)", intf.FileName())
 			return true
 		}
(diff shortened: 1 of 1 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -497,21 +505,22 @@ func (f *sendReceiveFolder) processDeletions(fileDeletions map[string]protocol.F
...
-	for i := range dirDeletions {
+	for _, dir := range dirDeletions {
 		select {
 		case <-f.ctx.Done():
 			return
 		default:
 		}
 
-		dir := dirDeletions[len(dirDeletions)-i-1]
+		f.resetPullError(dir.Name)
```
AudriusButkevicius: Now we try to delete `a\b` before `a\b\c`. Seems wrong?

imsodin: Thanks!  
I think I should take a break after fixing the two known messes I already caused. I am producing so much BS at the moment.

calmh: Ah, I missed this behavior change in the review as well :/

PR #5691, opened 2019-05-06
lib/model: Fix regression deleting directories on pull (ref #5690)

imsodin:


lib/model/folder_sendrecv.go
```diff
@@ -513,14 +513,15 @@ func (f *sendReceiveFolder) processDeletions(fileDeletions map[string]protocol.F
...
-	for _, dir := range dirDeletions {
+	// Process in reverse order to delete depth first
+	for i := range dirDeletions {
 		select {
 		case <-f.ctx.Done():
 			return
 		default:
 		}
 
-		f.resetPullError(dir.Name)
```
calmh: Don't you still want to reset the pull error for the dir though?

PR #5692, opened 2019-05-07
Added "Links" directory consisting of helpful URLs/Links + Experimental Electron Wrapper

jebeld17:
### Purpose

I propose adding './syncthing/links' to the Synchthing Master branch for use as a sort-of "bookmarks" list for helpful URLs to end-users. .url files within this directory could be integrated into Syncthing as Desktop, App Drawer, Start Menu, etc. shortcuts on general public end-user machines.

### Testing

This addition is not very technical and contains no scripts. Simply an added directory. 

### Documentation

(View changes below for proper documentation)

### EDIT - 05/07/2019 at 12:04PM EST

I've edited this Pull Request to continue having the changes made listed above, but now an Experimental Electron Wrapper within another new directory as well. This, in theory, could replace any .URL links entirely with a built-in AIO solution.

## Authorship

👍I agree to putting my name and contact information into the program.
Jeb Stuart Eldridge
jebeld17@gmail.com


calmh:
Thanks for the idea! However the url file in the source repo on its own doesn’t do much good, and it’s a bit platform specific, installation specific, and inscrutable. How would this be interpreted as suggested?

jebeld17:
> Thanks for the idea! However the url file in the source repo on its own doesn’t do much good, and it’s a bit platform specific, installation specific, and inscrutable. How would this be interpreted as suggested?

I am not really that great with coding - unfortunately - but I am getting there, so please excuse my sloppiness. ;)
Anyways, this is more of a proof-of-concept and wireframe for what could be a great addition. Everything here - whether it's a .URL link or an Electron app - can be scripted somehow. 
For example, the "Syncthing Web Console.url" file is currently set to the default "http://127.0.0.1:8384/" URL, but that file can be automatically edited to change if the user applies a configuration change manually.

AudriusButkevicius:
I don't think this belongs here. This is the repo that holds the source code, even the docs live in a separate repo. Yes, this has some systemd bits and what not, but that is for the purpose to make it easy for package maintainers.

As it stands there is no package for windows, so all of these things in some directory in the source tree don't do anything.

If you want to have some bits of stuff to use and to give people to look at, sure, but keep that in a separate repo.

Last thing I want is hundreds of megabytes of chromium/electron runtime lying around doing nothing in the source code repo making all git operations slow.

calmh:
Yeah, I’m not opposed to an electron wrapper, but not here and not like this. Sorry.

PR #5699, opened 2019-05-09
lib/model: Refactor file deletions when pulling

imsodin:
Tiny improvement split from #5680:  
Immediately delete symlinks and files, that are not candidates for renames. And reuse the file info from database if possible.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -333,17 +333,23 @@ func (f *sendReceiveFolder) processNeeded(dbUpdateChan chan<- dbUpdateJob, copyC
 				// Perform directory deletions at the end, as we may have
 				// files to delete inside them before we get to that point.
 				dirDeletions = append(dirDeletions, file)
+			} else if file.IsSymlink() {
+				f.resetPullError(file.Name)
+				f.deleteFile(file, dbUpdateChan, scanChan)
 			} else {
-				fileDeletions[file.Name] = file
 				df, ok := f.fset.Get(protocol.LocalDeviceID, file.Name)
 				// Local file can be already deleted, but with a lower version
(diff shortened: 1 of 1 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -333,17 +333,23 @@ func (f *sendReceiveFolder) processNeeded(dbUpdateChan chan<- dbUpdateJob, copyC
...
 				// WithNeed, furthermore, the file can simply be of the wrong
 				// type if we haven't yet managed to pull it.
 				if ok && !df.IsDeleted() && !df.IsSymlink() && !df.IsDirectory() && !df.IsInvalid() {
+					fileDeletions[file.Name] = file
 					// Put files into buckets per first hash
 					key := string(df.Blocks[0].Hash)
 					buckets[key] = append(buckets[key], df)
+				} else {
+					f.resetPullError(file.Name)
+					f.deleteFileWithCurrent(file, df, ok, dbUpdateChan, scanChan)
```
AudriusButkevicius: Almost feels like the delete should reset the error internally.

AudriusButkevicius: Also, this channel passing fiesta is getting out of hand, this whole thing is begging for a rewrite.

imsodin: To be consistent then all other ones should also go into e.g. `shortcutFile`, `handleDir` and so on. I opted for having all the error resets in the "top-level" functions (`processNeeded`, `processDeletions`). Honestly, I don't think it matters much.

And I most certainly don't care about channel passing here (and not much in general - or maybe, maybe not...).

imsodin: done

PR #5700, opened 2019-05-09
lib/model: Increment changed centrally when pulling

imsodin:
Tiny improvement split from #5680:  
Increment the `changed` counter in one place, just because it's simpler and less likely to be forgotten. Thus any checks which should not result in a retry should be done before that point (see comment).

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -315,7 +315,16 @@ func (f *sendReceiveFolder) processNeeded(dbUpdateChan chan<- dbUpdateJob, copyC
 			return true
 		}
 
+		if runtime.GOOS == "windows" && fs.WindowsInvalidFilename(intf.FileName()) {
+			f.newPullError(intf.FileName(), fs.ErrInvalidFilename)
+			return true
+		}
+
 		file := intf.(protocol.FileInfo)
+
+		// Do all checks that should not result in the pull being retried
+		// above this point.
+		changed++
 
 		switch {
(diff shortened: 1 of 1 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -315,18 +315,23 @@ func (f *sendReceiveFolder) processNeeded(dbUpdateChan chan<- dbUpdateJob, copyC
 			return true
 		}
 
+		if runtime.GOOS == "windows" && fs.WindowsInvalidFilename(intf.FileName()) {
```
calmh: But now we can no longer escape complaints and errors on invalid file names by ignoring them

lib/model/folder_sendrecv.go
```diff
@@ -315,18 +315,23 @@ func (f *sendReceiveFolder) processNeeded(dbUpdateChan chan<- dbUpdateJob, copyC
...
+			f.newPullError(intf.FileName(), fs.ErrInvalidFilename)
+			return true
+		}
+
 		file := intf.(protocol.FileInfo)
 
+		// Do all checks that should not result in the pull being retried
+		// above this point.
+		changed++
```
calmh: Previously we did not increment changed when a file had no availability. It seems to me this change might result in eternal retries of unavailable files (because devices are offline, for example).

imsodin:
Yeah right this is wrong - and even if not, it would have been change for changes sake which is bad in itself.

Issue #5695, opened 2019-05-08
Panic on Windows 10 with filesystem watching


FomalhautWeisszwerg:
## Problem

Syncthing v1.1.2 that runs on Windows 10 cause panic during synchronizing files with the other node.

### Node 0
OS: Debian GNU/Linux
Syncthing version: v1.1.2

### Node 1
OS: Windows 10 Pro (version 10.0.18362 N/A Build 18362)
Syncthing version: v1.1.2

## Steps to reproduce the problem

1. Run Syncthing on the Node0 (Debian) with command `syncthing 2>&1`

2. Run Syncthing on the Node1 (Windows10) with command `syncthing`

3. Syncthing on the Node1 always causes panic.

## Log

The file [panic-20190508-223555.log](https://github.com/syncthing/syncthing/files/3157632/panic-20190508-223555.log) was recorded on the Node1 (Windows10).

Do you have any ideas?

calmh:
Hello from the future!

<img width="716" alt="Screenshot 2019-05-08 at 17 39 47" src="https://user-images.githubusercontent.com/125426/57388368-4f4edb80-71b8-11e9-91a7-129b4d1d7603.png">

calmh:
I’ve dropped a v1.1.3 release to fix the panic. Filesystem watching of whole disks still won’t work properly though (#5578, same as in v1.1.1).

calmh:
Wow the time warp on the comments here...

uok:
a glitch in the Github matrix... If you see a Windows 10 agent... RUN

PR #5701, opened 2019-05-09
lib/fs: Remove \\?\ for drive letters when watching (fixes #5578)

### LINKED ISSUE #5578: Filesystem watcher error when watching disk root directory on Windows

imsodin:
Predecessor: #5633 reverted in https://github.com/syncthing/syncthing/commit/a04fcfe7494a6e3ede091534597539b77dc05a52 due to https://github.com/syncthing/syncthing/issues/5695

The problem was, that the prefix was removed only on the watched path, but not on the path that events were compared against.

### DIFF
--- a/lib/fs/basicfs_watch_test.go
+++ b/lib/fs/basicfs_watch_test.go
@@ -149,6 +149,53 @@ func TestWatchRename(t *testing.T) {
 	testScenario(t, name, testCase, expectedEvents, allowedEvents, fakeMatcher{})
 }
 
+// TestWatchWinRoot checks that a watch at a drive letter does not panic due to
+// out of root event on every event.
+// https://github.com/syncthing/syncthing/issues/5695
+func TestWatchWinRoot(t *testing.T) {
+	if runtime.GOOS != "windows" {
+		t.Skip("Windows specific test")
+	}
+
+	outChan := make(chan Event)
+	backendChan := make(chan notify.EventInfo, backendBuffer)
+
+	ctx, cancel := context.WithCancel(context.Background())
+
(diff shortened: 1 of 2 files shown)


lib/fs/basicfs_watch_test.go
```diff
@@ -149,6 +149,53 @@ func TestWatchRename(t *testing.T) {
...
+		t.Skip("Windows specific test")
+	}
+
+	outChan := make(chan Event)
+	backendChan := make(chan notify.EventInfo, backendBuffer)
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	// testFs is Filesystem, but we need BasicFilesystem here
+	root := `D:\`
```
AudriusButkevicius: Why not c:\ which should always be there?

calmh: I assumed this does not matter as there is no d: on the build server

imsodin: It doesn't, there is no real filesystem access, the watching is faked, but c: would probably have been more "normal", not sure exactly what my thoughts on this choice were.

PR #5702, opened 2019-05-09
lib/ur: Implement crash (panic) reporting (fixes #959)

### LINKED ISSUE #959: Feature request: Implement posibility for automatic error log sending to dev. team
Following this project from the sideline (by watching this git) I see a lot of sending back and forth error logs. Why not implement an option (like many others do), where error logs are sent automatically to a central repository? This could be both backend and frontend. This option should be an option so its possible to 'not to participate'. This is not an uncommon feature as a lot of software gives the user the opportunity to participate in making the software better/more stable. The user should at any point be able to see what have been sent. Then when reporting bugs here at Github, only a/multiple reference numbers is enough to post with a description of the situation.

calmh:
### Purpose

This implements a simple crash reporting method. It piggybacks on the panic log files created by the monitor process, picking these up and uploading them from the usage reporting routine.

Internally this is known as "panic uploading" but the user facing term is "crash reporting" as this is more mainstream.

This becomes a new usage reporting version. I've added strings to the GUI to the effect that usage reporting includes crash reporting.

The method and server side is trivial; panic logs are uploaded under their SHA256 hash, with a HEAD request first to avoid sending the entire log multiple times. Actually doing something useful with this data is up to the operator of the crash receiver. (Also making sure that the crash reports don't exhaust available disk, etc.)

A new config value points to the crash receiver base URL, which defaults to "https://crash.syncthing.net/newcrash" (following the pattern of "https://data.syncthing.net/newdata" for usage reports, but allowing us to separate the service as required).

### Testing

I created a panic log manually and noted that it got uploaded to the server, which I've set up on data.syncthing.net.

### Caveats

- It won't work when running without the monitor, because there will be no panic files. I don't think this is a problem, we just want to be able to get some sort of representative sample and indicator when a new problem is introduced into the wild...

- It'll send a bunch of HEAD requests for old, known, reports until they are cleaned out by the directory cleaner thing. I think the volume will be manageable.

- Syncthing needs to be able to stay running at least long enough for the usage reporting thing to kick in (30 minutes). We won't get reports of complete disasters.


lib/ur/usage_report.go
```diff
@@ -66,7 +72,23 @@ func New(cfg config.Wrapper, m model.Model, connectionsService connections.Servi
...
+		go func() {
+			uploadPanicLogs(s.cfg.Options().CRURL, locations.GetBaseDir(locations.ConfigBaseDir))
+			close(uploadDone)
+		}()
+	} else {
+		close(uploadDone)
+	}
+
+	// Do the data report, and also wait for uploads to complete (potentially)
+	err := s.reportData(urVersion, false)
```
imsodin: This returns usage reporting data, not an error.

calmh:
> I don't think uploading crashes should hook into ReportData

You're right, this is a mistake. I read it as a verb phrase ("report the data!") not as a getter ("get report data").

> hooking directly into the panic detection in monitor

I considered this, even started by doing that, but it complicates asking the user for permission

calmh:
I removed the waiting as well. When I initially did this I envisioned it running directly on startup, potentially then running often if there is a restart loop. In practice it'll run 30 minutes after startup and then once a day, so no real need to wait between requests. This is simpler.

lib/ur/usage_report.go
```diff
@@ -66,7 +72,8 @@ func New(cfg config.Wrapper, m model.Model, connectionsService connections.Servi
 // ReportData returns the data to be sent in a usage report with the currently
```
calmh: Note that this comment was a lie, unfortunately.

canton7:
> It won't work when running without the monitor, because there will be no panic files

That means you're unlikely to get crash reports from Windows users, so it's less likely you'll see Windows-only issues?

calmh:
Is there a reasonable way to improve that?

The wire interface is trivial though. Synctrazor et al can easily implement it, if desired, once deployed.

lib/ur/usage_report.go
AudriusButkevicius: Why?

lib/ur/usage_report.go
```diff
@@ -33,7 +34,12 @@ import (
 // Current version number of the usage report, for acceptance purposes. If
 // fields are added or changed this integer must be incremented so that users
 // are prompted for acceptance of the new report.
-const Version = 3
+const Version = 4
```
AudriusButkevicius: I know it's easy to piggyback of this, but the "diff" thing allowing to diff between the Ur versions will show no diffs, causing utter confusion.

I guess it's sad to duplicate all of the code, but I feel we should give some sort of indication there, or have a separate prompt. This also could be enabled by default with some warning on startup?

AudriusButkevicius:
Also, I don't remember how it works, but don't panics always roll into main() anyway?

AudriusButkevicius:
Also, there is sentry software which i recommended a while a go which deals with this, charts, similar panic deduplication, etc.

calmh:
> Why can't we recover somewhere in syncthingMain?

Panics in other goroutines do not "bubble" into main. There needs to be a deferred recover on the call stack, and main is not in the call stack for those routines. 

> I know it's easy to piggyback of this, but the "diff" thing allowing to diff between the Ur versions will show no diffs, causing utter confusion.

I know, it's not optimal. I added the text that usage reporting includes crash reporting; users who remember & care should hopefully notice the difference, those that don't might be slightly confused and then "whatever", which I think is suboptimal but acceptable.

> have a separate prompt. This also could be enabled by default with some warning on startup?

More prompts are not great, imho. Enabled by default (even with a warning) will probably get us lynched.

> Also, there is sentry software which i recommended a while a go which deals with this, charts, similar panic deduplication, etc.

I'm sure there is. Personally I don't care about that level of statistical analysis at this point, and it can be retrofitted if we do care. I definitely don't want to deal with tying all Syncthing binaries in the wild to a specific cloud provider whose privacy policy we will need to depend on, track and defend towards our users, etc.

This, for me, is very much a case of something being better than nothing. Having this, and taking a look for panics from the RC before release, might have saved us from the snafu the other day. That's all I ask.

AudriusButkevicius:
Sentry is self hosted so not sure why privacy policy stuff matters? Do we have a privacy policy which we promise to users?

We are literally reinventing a wheel here.

calmh:
Self hosted is good. Sentry, specifically, means a new VM with postgresql, redis, configuration and stuff. That's acceptable but a bit more heavyweight than I'd prefer. Maybe it adds functionality that's worth it? 🤷‍♂ 

Using it is more annoying, imho, as capturing a panic requires [wrapping at the call site](https://docs.sentry.io/clients/go/usage/#capturing-panics). This is invasive and something we're liable to forget, lots of places it needs adding but finding most of them is greppable...

So for just getting a handful of panic backtraces this seems like a strictly worse solution to me.

Is there other stuff that makes it worthwile? Can it effectively replace our entire usage reporting and do better reporting than we do ourselves, for example? I see you can send "events" with arbitrary tagged data in it, but not sure if that would result in something similar.

calmh:
I'm open to being convinced. You're welcome to spin up a sentry server and make a proof of concept of integrating it. This all seems like a lot of work to me though, work that I'm not really super interested in taking on for the eternal future, as opposed to the PR above that I wrapped up in about an hour.

AudriusButkevicius:
I'll take that challange, just need to find some time.
Do we have any infra that has docker on it?

calmh:
Not with the resources free, but we can set up a new VM. I can do that and ansible it up to common+docker state if you like.

calmh:
sentry.syncthing.net, the usual access applies

calmh:
@AudriusButkevicius How about a combined approach with this plus sentry?

- We do *something* like what is in this PR. Maybe do the reporting from the monitor process as well, maybe separate out crash reporting permission from usage reporting permission and default crash reporting to on, etc, we can sweat the details.
- The server side tool accepts the text format report like now, and stores it under a SHA256 of the report entire, like now. It can make this available as well under for example `https://crash.syncthing.net/report/$thesha`.
- Another server side tool parses these reports and shoots them into Sentry. We have all the version info, timestamps, etc we need in the log, in a format so we know what to expect. We can load load source code etc at this stage from git, checking out the appropriate version.
- In addition to roughly the same stuff you injected into sentry, it can also inject a link to the full report URL as above.

I think this could be good because

1. Simpler implementation Syncthing-side and better control over how we accept the reports server side.
2. Easier to integrate with the wrappers. They need to either create a panic file like we expect it and Syncthing will take care of reporting it, or just report in the same format to our endpoint. That "format" is just our stdout+stderr so no real logic required.
3. We get a nice aggregatable UI and technical raw format crash reports for the greybeards
4. Cake is consumed, yet remains?

AudriusButkevicius:
The wrappers would then need to do the magical "caputre first N lines and last N lines" stuff to make it work, but yeah, this sounds reasonable.

I guess my only concern with the current implementation is that it won't capture crash on startup stuff, as the logic is not in the monitor.

calmh:
Which is valid, and fixable. It's really just a matter of running the reporter thread (also?) from the monitor, or at least do one iteration of it when the monitor (re)starts syncthing to pick up anything present at that time.

I've tried to find data on what other projects do in terms of opting in / out of crash reporting, to benchmark what is the common thing. I've found it hard to google or get concrete data on this though. :/

AudriusButkevicius:
Hashicorp uses https://github.com/mitchellh/panicwrap

AudriusButkevicius:
So sentry is now setup on https://sentry.syncthing.net.

Docker compose that runs the circus is in onpremise dir in my home directory.
Email is still broken, and requires changing the mailgun key in docker-compose.yml, and doing a docker-compose restart.

We can add crash.syncthing.net as a cname to sentry.syncthing.net, run your magical service in a docker container part of that compose thing, and change the Caddyfile in the onprem dir to setup forwarding for that cname to your service.

cmd/stcrashreceiver/stcrashreceiver.go
```diff
@@ -0,0 +1,116 @@
...
+	case http.MethodPut:
+		r.servePut(base, w, req)
+	default:
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+	}
+}
+
+// serveHead responds to HEAD requests by checking if the named report
+// already exists in the system.
+func (r *crashReceiver) serveHead(base string, w http.ResponseWriter, req *http.Request) {
```
golangcibot: `(*crashReceiver).serveHead` - `req` is unused (from `unparam`)

cmd/stcrashreceiver/stcrashreceiver.go
```diff
@@ -0,0 +1,116 @@
...
+		return
+	}
+	defer dst.Close() // errors shmerrors what am I going to do about it
+
+	// Read at most maxRequestSize into it. At this point we don't really
+	// care about errors any more, whether they are network or I/O errors.
+	// Consider it best effort.
+	log.Println("Receiving report", base)
+	lr := io.LimitReader(req.Body, maxRequestSize)
+	io.Copy(dst, lr)
```
golangcibot: Error return value of `io.Copy` is not checked (from `errcheck`)

cmd/stcrashreceiver/stcrashreceiver.go
```diff
@@ -0,0 +1,132 @@
...
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+
+	// Read at most maxRequestSize of report data.
+	log.Println("Receiving report", base)
+	lr := io.LimitReader(req.Body, maxRequestSize)
+	bs, err := ioutil.ReadAll(lr)
+	if err != nil {
+		log.Printf("Reading report: %v", base, err)
```
golangcibot: printf: Printf call needs 1 arg but has 2 args (from `govet`)

calmh:
I've moved reporting to the monitor, added sentry proxying to the crash receiver. Going to go and steal your source code loader too @AudriusButkevicius and add it. Also fixed up sentry so that it now listens on 443 and can send email

AudriusButkevicius:
> Also fixed up sentry so that it now listens on 443
Caddy did that, and got the certs for us?

AudriusButkevicius:
I don't see the source code loader.
I also think this needs a one off notification on startup box like we had for fs notifications.

calmh:
Here comes the notifications to begin with. As discussed on the forum, the migration does `crash reporting enabled = usage reporting enabled || global discovery enabled` and then presents one of these:

Screen Shot 2019-06-10 at 10 22 57


Screen Shot 2019-06-10 at 10 25 56

calmh:
I was hoping to get this out in 1.2.0 but time is short so I think it will need to slide to the next release, which properly will make that 1.3.0 instead.

calmh:
wrong button

calmh:
btw https://github.com/syncthing/docs/pull/451 is the developer side of crash reporting, not the user facing thing i plan to write

AudriusButkevicius:
Release is tomorrow, hence why won't make it?

cmd/stcrashreceiver/sentry.go
```diff
@@ -0,0 +1,160 @@
...
+	// it. So we copy the values from the packet to the client first...
+	cli.SetRelease(pkt.Release)
+	cli.SetEnvironment(pkt.Environment)
+
+	_, errC := cli.Capture(pkt, nil)
+	return <-errC
+}
+
+func parseReport(path string, report []byte) (*raven.Packet, error) {
+	parts := bytes.SplitN(report, []byte("\n"), 2)
```
imsodin: I couldn't find where the version is written to the data being sent on the cmd/syncthing side of things, but there's plenty of newlines in the panic report.

cmd/syncthing/monitor.go
```diff
@@ -430,3 +441,39 @@ func childEnv() []string {
 	env = append(env, "STMONITORED=yes")
 	return env
 }
+
+// maybeReportPanics tries to figure out if crash reporting is on or off,
+// and reports any panics it can find if it's enabled. We spend at most
+// panicUploadMaxWait uploading panics...
+func maybeReportPanics() {
+	// Try to get a config to see if/where panics should be reported.
+	cfg, err := loadOrDefaultConfig()
```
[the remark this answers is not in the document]
calmh: Yes. Why not? :) Typically this will mean something on first startup only, where we will use the same defaults that will be written out when the real process starts.

cmd/stcrashreceiver/sentry.go
calmh: It's in the first line

cmd/syncthing/monitor.go
```diff
@@ -33,6 +34,8 @@ const (
 	loopThreshold         = 60 * time.Second
 	logFileAutoCloseDelay = 5 * time.Second
 	logFileMaxOpenTime    = time.Minute
+	panicUploadMaxWait    = 2 * time.Minute
```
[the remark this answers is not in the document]
calmh: EDGE network, large panic? Not sure, I just picked a number out of thin air. We could push it down a bit, maybe 30s?

calmh:
@AudriusButkevicius if it makes it, it makes it. I might skip the source code loader at this stage because I want to get that part right, and it doesn't go into the release anyhow.

cmd/stcrashreceiver/sentry.go
imsodin: Yeah, I am sure it's there in plain sight, but all I find is how the `filterLogLines` looks for `Panic ` and then keeps everything from that line on. I can't find anything that retains the info log line with the version info, that is before that \*shrug\*

cmd/syncthing/crash_reporting.go
```diff
@@ -0,0 +1,152 @@
...
+	filtered := data[:0]
+	matched := false
+	for _, line := range bytes.Split(data, []byte("\n")) {
+		switch {
+		case !matched && bytes.HasPrefix(line, []byte("Panic ")):
+			// This begins the panic trace, set the matched flag and append.
+			matched = true
+			fallthrough
+		case len(filtered) == 0 || matched:
+			// This is the first line or inside the panic trace.
```
calmh: @imsodin This case will be taken when `len(filtered)==0` (we have not yet accepted any lines), which ends with appending the line to the `filtered` set. So we grab the first line here. That line has the known format and informs the receiver of version etc.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -2426,4 +2426,14 @@ angular.module('syncthing.core')
...
+
+        $scope.enableCrashReporting = function () {
+            $scope.config.options.crashReportingEnabled = true;
+            $scope.saveConfig();
+        };
+
+        $scope.disableCrashReporting = function () {
+            $scope.config.options.crashReportingEnabled = false;
+            $scope.saveConfig();
+        };
```
AudriusButkevicius: `setCrashReportingEnabled(bool)`?

PR #5707, opened 2019-05-10
lib/model, lib/testutils: Test closing a connection on folder restart

imsodin:
The added test would have caught a deadlock that in the end resulted in the revert of all these changes (https://github.com/syncthing/syncthing/pull/5688).

lib/testutil is currently only used in model, but I wrote some more tests reproducing e.g. #4170 in lib/protocol which will also use this code. I intentionally put it into lib/testutil not lib/util, as this is code that should never be included in production code. And there's certainly more test helpers useful in many packages, e.g. lib/model/testos_test.go which could go there in the future.

### DIFF
--- a/lib/model/fakeconns_test.go
+++ b/lib/model/fakeconns_test.go
@@ -13,6 +13,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/syncthing/syncthing/lib/connections"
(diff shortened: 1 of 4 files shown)


AudriusButkevicius:
I think I'll disable coverage, as it varies badly based on which go routines take what path during the test.

imsodin:
Reminder

Issue #5706, opened 2019-05-10
Spurious test failures on Windows


calmh:
I don't always run tests on Windows, but when I do they fail.

```
C:\Users\Jakob Borg\source\repos\syncthing>go version
go version go1.12 windows/amd64

C:\Users\Jakob Borg\source\repos\syncthing>go run build.go test
ok      github.com/syncthing/syncthing/lib/api  6.613s
ok      github.com/syncthing/syncthing/lib/auto 0.334s
ok      github.com/syncthing/syncthing/lib/beacon       0.386s
ok      github.com/syncthing/syncthing/lib/build        0.253s
ok      github.com/syncthing/syncthing/lib/config       0.447s
ok      github.com/syncthing/syncthing/lib/connections  1.320s
ok      github.com/syncthing/syncthing/lib/db   1.355s
ok      github.com/syncthing/syncthing/lib/dialer       0.411s [no tests to run]
ok      github.com/syncthing/syncthing/lib/discover     0.702s
ok      github.com/syncthing/syncthing/lib/events       5.805s
--- FAIL: TestResolveWindows83 (0.00s)
    basicfs_windows_test.go:62: Resolving for 8.3 names of "\\?\C:\Users\JAKOBB~1\AppData\Local\Temp\476953126\LFDATA~1" resulted in "\\?\C:\Users\Jakob Borg\AppData\Local\Temp\476953126\LFDataTool", expected "\\?\C:\Users\JAKOBB~1\AppData\Local\Temp\476953126\LFDataTool"
    basicfs_windows_test.go:65: Resolving for 8.3 names of "\\?\C:\Users\JAKOBB~1\AppData\Local\Temp\476953126\foo\LFDATA~1" resulted in "\\?\C:\Users\JAKOBB~1\AppData\Local\Temp\476953126", expected "\\?\C:\Users\JAKOBB~1\AppData\Local\Temp\476953126\foo"
    basicfs_windows_test.go:68: Resolving for 8.3 names of "\\?\C:\Users\JAKOBB~1\AppData\Local\Temp\476953126\foo\bar\baz" resulted in "\\?\C:\Users\JAKOBB~1\AppData\Local\Temp\476953126", expected "\\?\C:\Users\JAKOBB~1\AppData\Local\Temp\476953126\foo\bar\baz"
--- FAIL: TestIsWindows83 (0.00s)
[...]

PR #5709, opened 2019-05-10
lib/fs: Consider win83 for root path as well when watching (ref #5706)

imsodin:
When checking paths from watches we resolve win83 stuff on the full path including root, but we never checked root itself. Funnily we haven't been bitten by it in production, potentially because of a side-effect of `filepath.EvalSymlinks`. Now we go the save way and treat both the root and paths from watches the same way. Hopefully fixes part of #5706

### DIFF
--- a/lib/fs/basicfs_unix.go
+++ b/lib/fs/basicfs_unix.go
@@ -76,3 +76,19 @@ func rel(path, prefix string) string {
 }
 
 var evalSymlinks = filepath.EvalSymlinks
+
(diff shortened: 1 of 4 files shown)


AudriusButkevicius:
How does this work if the path we are notified is 83, like it usually is?

imsodin:
The path we are notified off is already passed through `resolveWin83`, the root it is compared against wasn't, but now is.

PR #5711, opened 2019-05-11
test: Report time per MiB on transfer benchs

imsodin:
This is just a bit of polish for the integration transfer benchmarks:  
50000 files took more than 10min on my machine, and 10min is the internal timeout of the supervisor stuff (lib/rc).  
Also transfer rates are so low, that `%.1f`with MiB wasn't particularly useful. And I now print time per MiB instead of total time, to get rid of variation due to slightly varying total MiBs transferred.

### DIFF
--- a/test/transfer-bench_test.go
+++ b/test/transfer-bench_test.go
@@ -16,7 +16,7 @@ import (
 )
 
 func TestBenchmarkTransferManyFiles(t *testing.T) {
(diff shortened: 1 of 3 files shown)


calmh:
Fine by me. The default amount is just something that takes a reasonable time (~30-60s, iirc) on my macbook...

PR #5717, opened 2019-05-12
Remove superfluous check for IndexID in remote ClusterConfig.

acolomb:
### Purpose

Fix an obvious logic error in the model when handling `ClusterConfig` messages.  Checking for `IndexID == 0` inside a block where `IndexID != 0` was a precondition.  This should now correctly clear local index data when the remote side announces an IndexID of zero.

### Testing

Builds okay with this change, and the test suite fails locally for reasons I think are unrelated to this change (same FAILs on master).  I don't know enough about the protocol internals to imagine a real-world test scenario.  The logic error seems clear enough though, causing the intended handling code not to be called.


calmh:
Sounds reasonable.

@AudriusButkevicius I blocked the codecov app for now... Because 😬

PR #5719, opened 2019-05-12
build: Disable codecov comments/statuses

imsodin:
> I just revoked its permissions for syncthing/syncthing as the comments were … less than illuminating. (No comment on the rest.)  
https://forum.syncthing.net/t/code-coverage-integration/11848/7

So we can add the permissions again but still aren't troubled by less than enlightening comments/statuses.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,5 +1,10 @@
+comment: off
+
 coverage:
   range: "40...100"
+  status:
+    project: off
+    patch: off
 
 ignore:
   - "**.pb.go"


AudriusButkevicius:
We can do it for the org in their UI.

imsodin:
I can't find anything, can you do it (or point me at the right direction)?

AudriusButkevicius:
There is a global yaml editor for the whole org, which explains how to add bot names etc. Look at org settings.

imsodin:
Ok, found it (probably): https://codecov.io/account/gh/syncthing/yaml  
However only team members may adjust it.

AudriusButkevicius:
I think I gave you admin, even tho it was supposed to pickup stuff from gh permissions.

imsodin:
Now access worked, lets hope it does what it's supposed to.

PR #5721, opened 2019-05-13
Wedge in sentry everywhere

AudriusButkevicius:
This is a massive find replace.

It wraps suture in our own package which makes sure that each service started is wrapped in a panic capture.

Furthermore, as it's unavoidable `go func() {}()` becomes `sentry.Go(func() {})`


lib/sentry/sourcecodeloader.go
```diff
@@ -0,0 +1,125 @@
...
+		data, n := loader.Load(filename, line, context)
+		if data != nil {
+			return data, n
+		}
+	}
+	return nil, 0
+}
+
+type githubSourceCodeLoader struct {
+	mut   sync.Mutex
```
golangcibot: `mut` is unused (from `structcheck`)

AudriusButkevicius:
Rough preview:

image

AudriusButkevicius:
This is not finished, but open for discussion, before I waste more time on this.

lib/sentry/sourcecodeloader.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, loader := range l.loaders {
+		data, n := loader.Load(filename, line, context)
+		if data != nil {
+			return data, n
+		}
+	}
+	return nil, 0
+}
+
+type githubSourceCodeLoader struct {
```
calmh: What is all this stuff ... ?

lib/sentry/sentry.go
```diff
@@ -0,0 +1,145 @@
...
+				logger.DefaultLogger.Warnln("Logged panic. Event ID: " + eventID)
+			}
+		}
+
+		panic(err)
+	}
+}
+
+// This is lobotomised from the raven package.
+func enrichExtra(extra raven.Extra, err error) {
```
calmh: Can we grab the log buffer here and have it show up in a reasonable manner in the UI? Seeing what we were doing just before we crashed is generally very useful.

lib/sentry/transport.go
```diff
@@ -0,0 +1,25 @@
...
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package sentry
+
+import (
+	"github.com/getsentry/raven-go"
+	"github.com/pkg/errors"
+)
+
+var ErrSentryDisabled = errors.New("sentry disabled via config")
```
calmh: I think all the raven stuff becomes no-ops if the DSN is empty?

calmh:
I don't mind all the wrapping so, to me, the questions is just a matter of which end result is more desirable. I haven't yet ever logged in to the Sentry interface so I don't know its capabilities, so that remains to be seen. From what I gather from the above we don't get our log data, which to me is a significant downside. Hopefully this is easy to fix. The other question is how the data is managed in the UI. It looks nicely categorized based on tags and the actual backtrace. The only advantage plain text files on disk might have is the ability to free text grep them for stuff (like, a given service being present in the backtrace).

imsodin:
I agree that "full-text-searchability" of all reports would definitely be nice. Docs and google aren't entirely conclusive on what exactly is searchable: https://docs.sentry.io/workflow/search https://github.com/getsentry/sentry/issues/3248

Other nice things could be that there are a few points in the code where we debug log essentially "this cannot/shouldn't happen" and then get on with our lives, which could be reported to sentry.

With regard to wrapping suture: It would also be possible and arguably cleaner/simpler to hook into the already existing recovery mechanism there: https://github.com/thejerf/suture/blob/master/supervisor.go#L591 Given the successful `PassTroughPanic` addition I think a panic function hook could be accepted upstream. Just an idea.

AudriusButkevicius:
I was thinking about the feature addition in suture adding a custom defer func to be called in the service routine, but I think it would get refused on the basis of its interaction with existing failure capture mechanism. If that defered recovers the panic suture would have no way of knowing the cause of the failure and passthroughpanic would not work.

I guess your defer can repanic but thats ugly.

lib/sentry/transport.go
AudriusButkevicius: Sadly not, it just does nothing (does not unset the DSN):

https://github.com/getsentry/raven-go/blob/master/client.go#L488

lib/sentry/sentry.go
AudriusButkevicius: Done it differently.

lib/sentry/sourcecodeloader.go
AudriusButkevicius: I think we might not need this once we have the github integration up and running, but for now we implement a custom code loader which allows us to load code around the location of the crash, so that we'd get nice stack traces.

AudriusButkevicius:
Added support for log lines by abusing breadcrumbs:
image

Also added support for pushing all routine stacks:

image

image

AudriusButkevicius:
Also package (dep) information, but that doesn't seem to be printed in the UI (visible in "raw" view).

AudriusButkevicius:
Found a way to make the deps show up: 
image

lib/sentry/sourcecodeloader.go
calmh: It feels ... bizarre, that the code that is crashing should go looking for its own source on GitHub. This very much feels like something that should happen server side based on the backtrace...

lib/config/config.go
```diff
@@ -65,6 +65,8 @@ var (
 	DefaultDiscoveryServers = append(DefaultDiscoveryServersV4, DefaultDiscoveryServersV6...)
 	// DefaultTheme is the default and fallback theme for the web UI.
 	DefaultTheme = "default"
+	// DefaultSentyDSN is the default DSN used for sentry reporting
+	DefaultSentryDSN = "http://badeb4c1239b4b7281b2a37a7ca17560@sentry.syncthing.net:9000/3"
```
calmh: We should have this be just `https://sentry.syncthing.net/push` or something like that, and add the magic ID in the Caddy proxying. We might need to rotate the ID/key (it seems like you can't *set* this explicitly in the Sentry UI, and we might change the server, for example) and we don't want the constant compiled everywhere then...

lib/sentry/transport.go
calmh: That's fascinating, given https://docs.sentry.io/clients/go/config/

```
SetDSN

The DSN tells the SDK where to send the events. This option is always required
and the Sentry Go SDKs can pick this up from an environment variable (SENTRY_DSN).
When set to an empty string, SDK won’t send any events to Sentry, and all Capture*
methods will effectively act as no-ops.
```

lib/config/config.go
AudriusButkevicius: You can't do that. The client does explicit parsing of the user password stuff and uses it for header generation

lib/sentry/sourcecodeloader.go
AudriusButkevicius: I wonder if we could package all source code as part of the build, and use that.

It seems the github integration doesn't do source code lookup from gitbub.

lib/config/config.go
AudriusButkevicius: Or I guess we could embed some garbage in the user password and set it at the reverse proxy. Perhaps that would work.

AudriusButkevicius:
I suspect we could also do most of this in the monitor process (parse the logs) if we felt strongly about it, not needing all the wrapping stuff.

lib/config/config.go
calmh: Mmm we need to do *something* anyway if we use this, as we can't really deploy it tied to a key that we can't easily regenerate or move to another machine etc

calmh:
Mmm... Remember it needs to tie in to user permission somehow.

Not a judgement on your work, but it seems like really weak sauce in a Go panic reporting package that you need to collect backtraces and source yourself via various hacks to add to the report yourself...

AudriusButkevicius:
Well the source code is not available in the binary, so you have to do something. It's sad that you can't just wire sentry to rip it out of a repo on git.

Yes, raven client could be better, so could go be allowing to access the stack without having to print it to a string and then parse it back. I also wish go had a better way to catch panics.

AudriusButkevicius:
So what are peoples thoughts on this?

lib/sentry/sourcecodeloader.go
imsodin: Source code view is nice and all, but wouldn't the info in the normal trace (info on function/arguments and file/lineno) be enough?

imsodin:
I am not part of the entire hosting/setup effort, so the decision whether this is worth the effort isn't really mine. I don't mind the wrapping, I have some comments for review, but nothing "fundamental" (I'll hold off on those until the basic decision happened). From my simple stand-point, I am happy with this if you are fine with getting and keeping this running.

calmh:
Here's how I see it. There are some pros with this solution:

- Multiple identical panics are aggregated.
- Panics are nicely tagged with metadata.
- Sentry has some automatic redaction mechanisms that could be used to remove IP addresses etc from incoming data, I think.

There are some cons:

- It won't report anything that isn't properly wrapped. This includes, at least:
  - Stuff we (or new contributors) forget to wrap properly
  - Goroutines in dependencies (such as the db)
  - `fatal error` and such that are not panics and thus not recoverable
- It won't report if it *can't* at the time of panic. For example, if the connection service panics when the WiFi interface disconnects we'll never hear about it because the device will be offline at the moment it needs to report the panic.
- There's a lot of code to maintain, that might require updates for new Go versions (backtrace parsing).
- There's a not-insignificant amount of infrastructure to maintain & keep up to date.

As a matter of opinion, I think the GitHub loading stuff is bizarre and dangerous (might block, when we want to restart after a crash). If we can get source available server side by some after the fact enrichment, great. If we can't, I'm quite used to reading just the traceback.

Speaking of, I don't see the full traceback anywhere on the events in Sentry? You added that it should report all goroutines, yeah?

calmh:
> I don't see the full traceback anywhere on the events in Sentry

Ok, so it's there, you just have to 1) select the routine in the dropdown and 2) see the rest of the backtrace for that routine below. Frankly, that's a terrible way to look at a backtrace, and completely impossible for narrowing down a deadlock between hundreds of routines. To me this is a blocker compared to just looking at a text backtrace.

Screen Shot 2019-05-17 at 08 16 27

Screen Shot 2019-05-17 at 08 17 17

AudriusButkevicius:
You have the routine state in the dropdown, so you can look at the ones are in semacquire. We can choose not to send the surrounding code for the routines or we can package the code with the binary so there is no network access. 

I don't think running a single vm with a few containers wired via a a single docker conpose file classifies as a lot of intrastructure to run. We could even install docker on an existing box and run that stuff there.

We could move the panic log parsing to the monitor and report stuff from there, etc.

With your suggested method, if stuff is crashing on startup, we'd never see it in general, so not sure how much value there is in that.

Anyways, your opinion seems pretty conclusive, so I'll back off, even if I disagree with parts of it.

calmh:
> You have the routine state in the dropdown, so you can look at the ones are in semacquire.

Yeah, but come on. This is the only point I'll fight you on because it's important, the rest are just various tradeoffs and matters of taste. In a typical deadlock case there are hundreds of routines in semacquire. The only way, for me, to debug that is to feed it to `pp` which brings it down to a few tens of unique stacks. Then for each of those start annotating which locks they may hold and what they want. It's cumbersome and time consuming but it works to clear up the backtrace into something useful. There's *no* way to do that while juggling hundreds of identical dropdown entries - it's like trying to draw a map of the country but you're only allowed to look at one square centimeter at a time.

If the full blob of backtrace was available as a field or .txt file download from Sentry I'd be OK with that. Use the in GUI overview when that's good, dive into the original when needed. Can we get that? If we can, then we can start looking at whether to maybe capture from the monitor, maybe upload existing old traces as well, etc.

AudriusButkevicius:
So you can only attach 10kb of text, there is disagreeing docs, saying you might be able to do 250k.

https://docs.sentry.io/accounts/quotas/#attributes-limits
https://docs.sentry.io/development/sdk-dev/data-handling/

I don't think we can send files to sentry so either it would be some uncomfortable hack or we'd have to bolt something on the side.

I understand the pp argument, but how often do we have panics where it's a deadlock (the detector doesn't even run in releases), mostly it's just someone fatfingering something.

calmh:
It's not *just* deadlocks, although that's a killer case. It's also just things like "was the puller doing something when this panic happened?" which with a text backtrace can be answered immediately by `Cmd-f pull` and without it requires manually clicking an eyeballing hundreds of dropdown entries. It's just a stupid interface design, to have as the *only* option.

PR #5724, opened 2019-05-14
lib/model: Consistently cleanup model in tests

imsodin:
Consistently clean up behind tests using `model`. This removes leftover testing directories, goroutine leaks between tests and might also reduce coverage variability.


lib/model/model_test.go
```diff
@@ -3249,12 +3138,9 @@ func TestIssue5002(t *testing.T) {
...
 func TestParentOfUnignored(t *testing.T) {
-	wcfg, m := newState(defaultCfg)
-	defer func() {
-		m.Stop()
-		defaultFolderConfig.Filesystem().Remove(".stignore")
-		os.Remove(wcfg.ConfigPath())
-	}()
+	m := newState(defaultCfg)
+	defer stopModel(m)
+	defer defaultFolderConfig.Filesystem().Remove(".stignore")
```
golangcibot: Error return value of `(github.com/syncthing/syncthing/lib/fs.Filesystem).Remove` is not checked (from `errcheck`)

lib/model/testutils_test.go
```diff
@@ -114,6 +115,21 @@ func setupModel(w config.Wrapper) *model {
...
 }
 
+func newModel(cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersion string, ldb *db.Lowlevel, protectedFiles []string) *model {
+	return NewModel(cfg, id, clientName, clientVersion, ldb, protectedFiles).(*model)
+}
+
+func stopModel(m *model) {
+	m.Stop()
+	m.db.Close()
+	os.Remove(m.cfg.ConfigPath())
```
calmh: This is a slightly surprising action by something called just `stopModel`, but if it's only called from contexts where that makes sense it's OK by me under the circumstances

imsodin: Yeah, it actually just stops and cleans up everything that's used when a model is running during tests. Would `stop`, `cleanupModel` or `cleanupTestModel` reflect that better?

calmh: cleanupSomething would be better I think yes. But it’s not a big deal, given that we always start the model with a tmpdir config. Do whatever and merge. :)

PR #5725, opened 2019-05-14
lib/connections, lib/model: Check addr before hello exchange

imsodin:
There's no point and arguably even bad to exchange hello messages and do all kinds of stuff when the connection comes from a blacklisted address -> check that first.

### DIFF
--- a/lib/connections/connections_test.go
+++ b/lib/connections/connections_test.go
@@ -104,7 +104,7 @@ func TestAllowedNetworks(t *testing.T) {
 	}
 
 	for _, tc := range cases {
-		res := IsAllowedNetwork(tc.host, tc.allowed)
+		res := isAllowedNetwork(tc.host, tc.allowed)
 		if res != tc.ok {
 			t.Errorf("allowedNetwork(%q, %q) == %v, want %v", tc.host, tc.allowed, res, tc.ok)
 		}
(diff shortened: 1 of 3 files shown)


calmh:
I'm not sure I agree. The device is known and accepted; we trust it. It's trying to connect from some place we don't want it to. It seems reasonable and polite to say "hi, I'm syncthing, I know you, but I'm not going to talk to you at the moment".

imsodin:
I get that, but only the "hi, I'm devicename on syncthing version xyz" is actually happening, then we just go silent. I'd say that's worse than not sending a hello at all, thus not creating any useless work at all. If we'd actually send a message saying "I know you, but I don't like your current address" that might have some use, but currently we potentially block the single loop handling connections for up to 20s (timeout).

calmh:
Clearly we should close the connection with an error. If we don't do that it's definitely a bug, but I think a different one with a different solution than you're proposing

imsodin:
Ok fine, we'll have to agree to disagree on the fundamental point

What do you mean by closing the connection with an error - a BEP close message or is there the possibility to close the underlying connection such that it isn't just dropped and the other side is notified?

imsodin:
I should have just rtfm. From BEP:

> [...] even in the case that a connection is rejected a Hello message must be sent before the connection is terminated.

I still think the check should happen in `lib/model` and the connection status reporting implemented added in #5727 should also apply here - I'll open another PR for that.

PR #5727, opened 2019-05-16
lib/api, lib/connections, gui: Show connection error for disconnected devices (fixes #3345)

### LINKED ISSUE #3345: Show last connection error per discovered address
After #3340, we should track and show the error we got when trying to connect to each configured or discovered address. This could be "time out", "connection refused" etc and will be somewhat raw and technical but still help people figure out what's going on.

calmh:
### Purpose

Aids in understanding why a device is not connecting.

This adds functionality in the connetions service to track the last error per address. That is in turn exposed in the /rest/system/status API method, as that is also where we already show the listener status from the connection service.

The GUI uses this info where it lists addresses, showing errors (if any) in red underneath each address. The time of the last connection attempt is in a tooltip, because there really isn't space to put it somewhere better. :(

I also slightly refactored the existing status method on the connection service to have a better name and return typed information, including an `ok` bool when the `error` is missing, to be clearer and friendlier.

### Testing

I've only tested this manually. Automated testing of this seems annoying. :/

### Screenshots

Screenshot 2019-05-16 at 09 43 33

Screenshot 2019-05-16 at 09 48 02

### Documentation

The new data in system/status should be added to the example/docs.

```
jb@kvar:~ $ curl -ksH X-api-key:... https://localhost:8384/rest/system/status
{
  "alloc": 44296776,
  "connectionServiceStatus": {
    "tcp://:22000": {
      "ok": true,
      "lanAddresses": [
        "tcp://:22000"
      ],
      "wanAddresses": [
        "tcp://:22000"
      ]
    }
  },
...
  "lastDialStatus": {
    "tcp://10.20.30.40": {
      "when": "2019-05-16T07:41:23Z",
      "error": "dial tcp 10.20.30.40:22000: i/o timeout"
    },
    "tcp://172.16.33.3:22000": {
      "when": "2019-05-16T07:40:43Z",
      "ok": true
    },
    "tcp://83.233.120.221:22000": {
      "when": "2019-05-16T07:41:13Z",
      "error": "dial tcp 83.233.120.221:22000: connect: connection refused"
    }
  },
...
}
```

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -90,8 +90,20 @@ var tlsVersionNames = map[uint16]string{
 // dialers. Successful connections are handed to the model.
 type Service interface {
 	suture.Service
-	Status() map[string]interface{}
+	ListenerStatus() map[string]ListenerStatusEntry
+	ConnectionStatus() map[string]ConnectionStatusEntry
 	NATType() string
+}
+
+type ListenerStatusEntry struct {
+	Error        string   `json:"error,omitempty"`
+	LANAddresses []string `json:"lanAddresses"`
+	WANAddresses []string `json:"wanAddresses"`
+}
+
+type ConnectionStatusEntry struct {
+	When  time.Time `json:"when"`
+	Error string    `json:"error,omitempty"`
 }
 
 type service struct {
@@ -112,6 +124,9 @@ type service struct {
 	listeners          map[string]genericListener
 	listenerTokens     map[string]suture.ServiceToken
 	listenerSupervisor *suture.Supervisor
+
+	connectionStatusMut sync.RWMutex
+	connectionStatus    map[string]ConnectionStatusEntry // address -> latest error/status
 }
 
 func NewService(cfg config.Wrapper, myID protocol.DeviceID, mdl Model, tlsCfg *tls.Config, discoverer discover.Finder,
@@ -151,6 +166,9 @@ func NewService(cfg config.Wrapper, myID protocol.DeviceID, mdl Model, tlsCfg *t
 			FailureBackoff:    600 * time.Second,
 			PassThroughPanics: true,
 		}),
+
+		connectionStatusMut: sync.NewRWMutex(),
+		connectionStatus:    make(map[string]ConnectionStatusEntry),
 	}
 	cfg.Subscribe(service)
 
@@ -378,18 +396,23 @@ func (s *service) connect() {
 
 				uri, err := url.Parse(addr)
 				if err != nil {
+					s.setConnectionStatus(addr, err)
 					l.Infof("Parsing dialer address %s: %v", addr, err)
 					continue
 				}
 
 				if len(deviceCfg.AllowedNetworks) > 0 {
 					if !IsAllowedNetwork(uri.Host, deviceCfg.AllowedNetworks) {
+						s.setConnectionStatus(addr, errors.New("network disallowed"))
 						l.Debugln("Network for", uri, "is disallowed")
 						continue
 					}
 				}
 
 				dialerFactory, err := getDialerFactory(cfg, uri)
+				if err != nil {
+					s.setConnectionStatus(addr, err)
+				}
 				switch err {
 				case nil:
 					// all good
@@ -424,6 +447,7 @@ func (s *service) connect() {
 				}
 
 				dialTargets = append(dialTargets, dialTarget{
+					addr:     addr,
 					dialer:   dialer,
 					priority: priority,
 					deviceID: deviceID,
@@ -431,7 +455,7 @@ func (s *service) connect() {
 				})
 			}
 
-			conn, ok := dialParallel(deviceCfg.DeviceID, dialTargets)
+			conn, ok := s.dialParallel(deviceCfg.DeviceID, dialTargets)
 			if ok {
 				s.conns <- conn
 			}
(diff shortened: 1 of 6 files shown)


calmh:
*Maybe* we should strip the errors from everything up to and including the last colon, so that `dial tcp x.y.z.w:a: foo: connection refused` becomes just `connection refused`. Then we could probably fit the time on the same line.

AudriusButkevicius:
The ok, err paradigm is a terrible idea. It leaks from a decision go made into our apis. Presence or absence of error is more than good enough to indicate wether a error happened or not, I see zero value in the ok being there.

calmh:
I really wasn't thinking at all of the Go `err/ok` thing, I just thought that for a casual reader who reads the API output instead of the documentation (like I would do),

```
"lastDialStatus": {
    "tcp://172.16.33.3:22000": {
      "when": "2019-05-16T07:40:43Z"
    }
}
```

says effectively nothing. It just says that something happened at that time to that address, but not what. I can't know to expect an `error` if I don't see it. 

But this is just API aesthetics, it doesn't affect the surrounding functionality at all so 🤷‍♂

imsodin:
This is great!

As proposed I would move the full error message to the tooltip and display `*time*: *last part of err*` - that's more user friendly and less redundant (IP being already listed above).

And regarding API aesthetics: I'd also consider `error: null` (or whatever variant) as cleaner than two values `error` and `ok`, which also (dis-)appear. I'd consider logic that acts on the existence of `error`, without checking the value, as quite fragile. Whether or not an example in docs missing `error` is enough for this being a breaking change, I am not sure, but it doesn't seem practically relevant.

calmh:
Screenshot 2019-05-16 at 13 28 36

calmh:
> I'd consider logic that acts on the existence of error, without checking the value, as quite fragile.

Just for the record, if the consuming side is Go and has a ```
type whatevs struct {
   Error string `json:"error"`
}
```

that's not going to be able to deserialize the `error: null` that we send now. Java, JavaScript, C#, Python probably don't have this specific issue though so I expect most of our integrations should survive... Hopefully. It's what the RC is for, right?

File: lib/connections/service.go [hunk 1]
```diff
@@ -667,9 +665,8 @@ func (s *service) ListenerStatus() map[string]ListenerStatusEntry {
 
 		err := listener.Error()
 		if err != nil {
-			status.Error = err.Error()
-		} else {
-			status.OK = true
+			errStr := err.Error()
+			status.Error = &errStr
```
imsodin: Could you please explain the reason for the string pointer, I can't figure it out.

calmh: To be able to produce `"error": null` in the JSON

imsodin: Hmm, yeah I see, should have guessed that...  
And grepping to see whether we already use something like this in other places pointed me at `events.Error` being used in pulling, for just that purpose: https://github.com/syncthing/syncthing/blob/master/lib/events/events.go#L485 (weird place for that function though, and orry for the bikeshedding).

uok:
May I suggest "Error: connection refused (13:28:21)" as format? Thanks!

calmh:
I'll give you the parenthesis as it puts the error in front, but I'll pass on the "Error:" as I think it's clear from the contect that it's an error, and it would introduce all of the translation complexity, including sentence structure issues around where to place the error message and time plus the fact that it would end up a pseudo-translated mess. Keeping it as a technical string works, I think.

calmh:
Screen Shot 2019-05-16 at 16 02 15

lib/connections/service.go
```diff
@@ -90,10 +90,22 @@ var tlsVersionNames = map[uint16]string{
...
 
+type ListenerStatusEntry struct {
+	Error        *string  `json:"error"`
+	LANAddresses []string `json:"lanAddresses"`
+	WANAddresses []string `json:"wanAddresses"`
+}
+
+type ConnectionStatusEntry struct {
+	When  time.Time `json:"when"`
+	Error *string   `json:"error"`
```
AudriusButkevicius: I think this can be a normal string with omitempty?

lib/connections/service.go
```diff
@@ -633,24 +657,47 @@ func (s *service) ExternalAddresses() []string {
...
-		status["wanAddresses"] = urlsToStrings(listener.WANAddresses())
+		status.LANAddresses = urlsToStrings(listener.LANAddresses())
+		status.WANAddresses = urlsToStrings(listener.WANAddresses())
 
 		result[addr] = status
 	}
 	s.listenersMut.RUnlock()
 	return result
+
+}
```
AudriusButkevicius: Blank

AudriusButkevicius:
My omitempty suggestion still valid:
https://play.golang.org/p/qYiqDeKj4lH

imsodin:
That would bring us back to the concern that Jakob stated in https://github.com/syncthing/syncthing/pull/5727#issuecomment-492969249 that 

```
"lastDialStatus": {
    "tcp://172.16.33.3:22000": {
      "when": "2019-05-16T07:40:43Z"
    }
}
```
doesn't say much, as opposed to the current, not omitempty, 
```
"lastDialStatus": {
    "tcp://172.16.33.3:22000": {
      "when": "2019-05-16T07:40:43Z",
      "error": "null",
    }
}
```
or the originally proposed
```
"lastDialStatus": {
    "tcp://172.16.33.3:22000": {
      "when": "2019-05-16T07:40:43Z",
      "ok": "true",
    }
}
```

I really think we are spending too much time on relatively pointless aesthetics here (me included). From my side it's lgtm with any of these options.

PR #5732, opened 2019-05-18
lib/protocol: Test for Close on blocking send deadlock (ref #5442)

imsodin:
This is a test that reproduces the deadlock scenario described in #5442.

### DIFF
--- a/lib/protocol/protocol_test.go
+++ b/lib/protocol/protocol_test.go
@@ -11,11 +11,13 @@ import (
 	"io"
 	"io/ioutil"
 	"runtime"
-	"strings"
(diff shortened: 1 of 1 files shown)


imsodin:
Should I do one big PR regarding the protocol changes I'd like to do with commits that should be individually merged? I can see that all these small, seemingly pointless PRs are low-prio, but they are requirements for fixing the cluster config panics which is somewhat important ((not really all requirements, but separate and would cause conflicts).

AudriusButkevicius:
Let's have them one at a time. Scouts promise I'll look at them.

PR #5733, opened 2019-05-19
lib/model: Close connections when model is stopped

imsodin:
The idea is that we notify connected remotes that were stopping instead of just dropping the connection.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -180,6 +180,7 @@ var (
 	// errors about why a connection is closed
 	errIgnoredFolderRemoved = errors.New("folder no longer ignored")
 	errReplacingConnection  = errors.New("replacing connection")
+	errStopped              = errors.New("Syncthing is being stopped")
 )
 
 // NewModel creates and starts a new model. The model starts in read-only mode,
@@ -223,6 +224,25 @@ func NewModel(cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersio
 	cfg.Subscribe(m)
 
 	return m
+}
+
+func (m *model) Stop() {
+	m.Supervisor.Stop()
+	devs := m.cfg.Devices()
+	ids := make([]protocol.DeviceID, 0, len(devs))
+	for id := range devs {
+		ids = append(ids, id)
+	}
+	m.pmut.RLock()
+	closed := make([]chan struct{}, 0, len(m.closed))
+	for _, c := range m.closed {
+		closed = append(closed, c)
+	}
+	m.pmut.RUnlock()
+	m.closeConns(ids, errStopped)
+	for _, c := range closed {
+		<-c
+	}
 }
 
 // StartDeadlockDetector starts a deadlock detector on the models locks which
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -868,6 +868,7 @@ func TestIssue5063(t *testing.T) {
 		conn.mut.Lock()
 		conn.closeFn = func(_ error) {}
 		conn.mut.Unlock()
+		defer m.Closed(c, errStopped) // to unblock m.Stop()
 	}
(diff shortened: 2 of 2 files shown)


lib/model/model_test.go
```diff
@@ -3267,12 +3268,7 @@ func TestSanitizePath(t *testing.T) {
 func TestConnCloseOnRestart(t *testing.T) {
 	w, fcfg := tmpDefaultWrapper()
 	m := setupModel(w)
-	defer func() {
-		m.Stop()
-		m.db.Close()
-		os.RemoveAll(fcfg.Filesystem().URI())
-		os.Remove(w.ConfigPath())
-	}()
+	defer cleanupModelAndRemoveDir(m, fcfg.Filesystem().URI())
```
AudriusButkevicius: Not relevant to this, but we could just get all the folders from the config (that exits on the mode) and remove them, without having to pass a URI.

lib/model/model.go
```diff
@@ -225,6 +226,25 @@ func NewModel(cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersio
...
+	for id := range devs {
+		ids = append(ids, id)
+	}
+	m.pmut.RLock()
+	closed := make([]chan struct{}, 0, len(m.closed))
+	for _, c := range m.closed {
+		closed = append(closed, c)
+	}
+	m.pmut.RUnlock()
+	m.closeConns(ids, errStopped)
```
AudriusButkevicius: I am not sure what the strategy here is, but we use error codes to signal stuff, so perhaps it would be worth expanding on that.

lib/model/model_test.go
```diff
@@ -868,6 +868,7 @@ func TestIssue5063(t *testing.T) {
 		conn.mut.Lock()
 		conn.closeFn = func(_ error) {}
 		conn.mut.Unlock()
+		defer m.Closed(c, errStopped) // to unblock m.Stop()
```
AudriusButkevicius: Who blocks Stop() ?

lib/model/model_test.go
imsodin: Yes, in some cases. But some of the tests are running on `./testdata` which is committed to git, so there it shouldn't remove anything.

lib/model/model.go
imsodin: What strategy? I just explicitly close all connections with an error that essentially says "Bye, not talking to you anymore because I am shutting down".

AudriusButkevicius: Yeah, but closing connections takes an error, then error types are mapped to error codes in the protocol package which are sent to the other side. (Or that was how it was when I first implemented it)

We are now closing with various types of errors the protocol does know what error code to send for, and they are all probably end up as generic errors with a message. Are there intentions to add separate protocol close codes for this or not?

lib/model/model_test.go
imsodin: The line two above is a noop, therefore `model.Close` will deadlock ans it never receives the `Closed` call. However I now don't understand why I added this `closeFn` business in the first place.

lib/model/model.go
imsodin: I think you are mixing this up with request responses. Close BEP messages just have a human readable message as field: https://docs.syncthing.net/specs/bep-v1.html#close

AudriusButkevicius: You are right.

AudriusButkevicius:
Hmm, sudden test failures: 

panic: bug: ClusterConfig called on closed or nonexistent connection

Wonder if it's related.

imsodin:
Potentially. However the fix for that is incoming (hopefully, can't reproduce locally with these tests, but I do have a new one that causes this panic).

AudriusButkevicius:
Perhaps it's just the buildservers running out of ram making this race more prominent.

PR #5734, opened 2019-05-19
lib/fs, lib/model: Add error channel to Watch to avoid panics (fixes #5697)

### LINKED ISSUE #5697: Handle fatal watch errors gracefully
Syncthing panics if it receives events outside of the folder root. I did that on the incorrect assumption that the notify library does in some form canonicalize all involved paths. However they only do that for internal housekeeping, paths passed to and received from the underlying filesystem watching interfaces are not touched. They even mention that some implementations change the path in certain situations. I have no intention to find out what all the underlying implementations do. So the sane thing is to report a verbose error if an out-of-root path is encountered and potentially stop watching. And then we can try to find out what the special case is and mitigate that in `filesystem.Watch` in a next release, while the user still has a working Syncthing (as opposed to a panicking one).

imsodin:
This adds the plumbing necessary to pass errors from a successfully set up watch to whoever receives the events and uses this to pass the outside root error on instead of panicking on it.

The handling on the model/folder side is pretty simple right now and works the same way as watch setup errors: Retry watching every minute and do a full rescan. The rescan is done through the rescan timer, such that we do at most one full rescan per minute, which is the default for folders without watching.

### DIFF
--- a/lib/fs/basicfs_watch.go
+++ b/lib/fs/basicfs_watch.go
@@ -11,7 +11,6 @@ package fs
 import (
 	"context"
 	"errors"
-	"fmt"
 
 	"github.com/syncthing/notify"
 )
@@ -118,14 +117,4 @@ func (f *BasicFilesystem) eventType(notifyType notify.Event) EventType {
 		return Remove
 	}
 	return NonRemove
-}
-
-type ErrWatchEventOutsideRoot struct{ msg string }
-
-func (e *ErrWatchEventOutsideRoot) Error() string {
-	return e.msg
-}
-
-func (f *BasicFilesystem) newErrWatchEventOutsideRoot(absPath, root string) *ErrWatchEventOutsideRoot {
-	return &ErrWatchEventOutsideRoot{fmt.Sprintf("Watching for changes encountered an event outside of the filesystem root: f.root==%v, root==%v, path==%v. This should never happen, please report this message to forum.syncthing.net.", f.root, root, absPath)}
 }
--- a/lib/fs/basicfs_watch.go
+++ b/lib/fs/basicfs_watch.go
@@ -11,6 +11,7 @@ package fs
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/syncthing/notify"
 )
@@ -20,10 +21,10 @@ import (
 // Not meant to be changed, but must be changeable for tests
 var backendBuffer = 500
 
-func (f *BasicFilesystem) Watch(name string, ignore Matcher, ctx context.Context, ignorePerms bool) (<-chan Event, error) {
+func (f *BasicFilesystem) Watch(name string, ignore Matcher, ctx context.Context, ignorePerms bool) (<-chan Event, <-chan error, error) {
 	watchPath, root, err := f.watchPaths(name)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	outChan := make(chan Event)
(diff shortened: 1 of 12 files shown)


lib/fs/basicfs_watch.go
```diff
@@ -74,7 +79,18 @@ func (f *BasicFilesystem) watchLoop(name, evalRoot string, backendChan chan noti
...
+			relPath, err := f.unrootedChecked(ev.Path(), evalRoot)
+			if err != nil {
+				select {
+				case errChan <- err:
+					l.Debugln(f.Type(), f.URI(), "Watch: Sending error", err)
+				case <-ctx.Done():
+				}
+				notify.Stop(backendChan)
+				l.Debugln(f.Type(), f.URI(), "Watch: Stopped due to", err)
+				return
```
AudriusButkevicius: Who closes outChan?
I don't see anyone closing it anywhere.

lib/model/folder.go
```diff
@@ -611,53 +608,87 @@ func (f *folder) startWatch() {
...
-			defer f.watchMut.Unlock()
-			watchaggregator.Aggregate(eventChan, f.watchChan, f.FolderConfiguration, f.model.cfg, ctx)
+			watchaggregator.Aggregate(eventChan, f.watchChan, f.FolderConfiguration, f.model.cfg, aggrCtx)
 			l.Debugln("Started filesystem watcher for folder", f.Description())
-			return
+		case err = <-errChan:
+			f.setWatchError(err)
+			// This error was previously a panic and should never occur, so generate
+			// a warning, but don't do it repetitively.
+			if !warnedOutside {
```
AudriusButkevicius: This is forever false.

lib/model/folder.go
```diff
@@ -611,53 +608,87 @@ func (f *folder) startWatch() {
...
-				if err != nil {
-					data["to"] = err.Error()
-				}
-				events.Default.Log(events.FolderWatchStateChanged, data)
+			if f.watchErr != nil && f.watchErr != errWatchNotStarted {
+				// We encountered an error before -> schedule rescan
+				// to catch potentially missed changes.
+				// We do this at most once per minute which is the
+				// default rescan time without watcher.
+				f.Delay(0)
```
AudriusButkevicius: This cruft could be in setWatchError

lib/fs/basicfs_watch.go
imsodin: That's correct and actually intended, as the context is the stopping/unblocking mechanism. This allows the receiver to use whatever it gets from the channel without checking for `nil`(or getting/checking the second bool "closedness" value). Is there some pitfall I am not seeing with not closing channels? I only ever know about the (somewhat obvious) principle of only sender should close a channel, but not that channels should be closed.

AudriusButkevicius: Isn't the aggregator just
```
go func() {
   for x := range chan {
   }
}()
```
?
That doesn't need to check for closedness and does not leak a routine (it probably doesn't anyway as channels are closed when gced, and where there are no writers left)

imsodin: Well but something happens inside the loop, and if `x` is not a real event but either `nil` or a default struct behaviour is not defined. And as in almost everywhere it's:

```
go func() {
   for {
      var x
      select {
         case x = <-chan:
         case <-ctx.Done():
            return
      }
   }
   ...
}()
```

AudriusButkevicius: Fine

PR #5735, opened 2019-05-20
build: Include cross-package test coverage

imsodin:
Our currently calculated coverage is quite a bit lower than it actually is, as go by default only considers coverage of the currently tested package even if many packages are given (as in our case with `lib/...`). On my system this change doesn't work, I reported it upstream at https://github.com/golang/go/issues/30374#issuecomment-493425265. If this passes it proofes something is broken on my side and we can use it, otherwise it works as another data point for the upstream issue.

### DIFF
--- a/build.go
+++ b/build.go
@@ -348,7 +348,7 @@ func test(pkgs ...string) {
 	}
 
 	if coverage {
-		args = append(args, "-covermode", "atomic", "-coverprofile", "coverage.txt")
+		args = append(args, "-covermode", "atomic", "-coverprofile", "coverage.txt", "-coverpkg", strings.Join(pkgs, ","))
 	}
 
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
I assume thisis still a WIP as it's clearly failing?

imsodin:
Yup, as in PR type. The upstream report shifted to https://github.com/golang/go/issues/32150 and has been reproduced, apparently it's fixed on tip and a backport to 1.12 will be investigated.

imsodin:
Go1.12.6 with the fix is here. And with the project wide coverage, there are ~3k more lines tracked and still 3% more lines covered overall (cosmetic, but still). I also hope it will reduce flakyness of the coverage, lets see about that.

PR #5673, opened 2019-04-29
lib: Add container types with disk overflow

imsodin:
This is a new take at the PR #4980: I simplified it by dropping all the dynamic memory usage limiting and replaced it with a configurable limit per object. I'd still prefer a global limit based on available system memory, but I get that's too many moving parts/too much complexity. One option would be to set the default value to something like a percent of the system memory instead of  the current 16MB.

PR #5737, opened 2019-05-20
lib/connections: Add QUIC protocol support

AudriusButkevicius:
Reuses a lot of KCP beef.
Works in my limited testing: make the listen loop select{} on the device that sits on the internet, force both to listen QUIC only.

I think when I implemented KCP I used a different router that was less strict so I could test better.

I don't have two NATs to test this properly, but definately looks more reliable than KCP was.


AudriusButkevicius:
Needs dial stream open timeouts

calmh:
It's not completely broken at least ❤️ 

Testing/review to follow :)

Screen Shot 2019-05-21 at 07 47 49

calmh:
STUN is defective though

```
[MKYEN] 07:58:05 INFO: quic://:22000 resolved external address quic://44.123.73.11:37845 (via stun.voxgratia.org:3478)
[MKYEN] 07:58:23 INFO: quic://:22000 resolved external address quic://13.28.215.245:2718 (via stun.voxgratia.org:3478)
[MKYEN] 07:58:41 INFO: quic://:22000 resolved external address quic://54.50.222.179:974 (via stun.voxgratia.org:3478)
[MKYEN] 07:58:59 INFO: quic://:22000 resolved external address quic://122.184.175.142:20392 (via stun.voxgratia.org:3478)
[MKYEN] 07:59:17 INFO: quic://:22000 resolved external address quic://23.194.99.11:18176 (via stun.voxgratia.org:3478)
[MKYEN] 07:59:35 INFO: quic://:22000 resolved external address quic://67.152.218.138:22137 (via stun.voxgratia.org:3478)
...
```

calmh:
Also, I have disabled NAT traversal in config on this specific box, for reasons. Yet is STUNs. It should not.

AudriusButkevicius:
Interesting. Why would your router force a new port on a packet every 18s?
Seems more like a defective router?

calmh:
It most certainly does not. It also does not force a completely random public IP on me every 18 s, I promise :)

AudriusButkevicius:
My money is on a sketchy stun server in that case.

It should pick a different one on restart hopefully.

calmh:
Looks like. Wireshark agrees with Syncthing, that it returns basically random garbage, addresses are random in each response. However, at least on this computer it *always* picks precisely this server, on every startup... Maybe they respond with garbage to unknown queriers? Maybe it's a bug? Maybe we need to handle this and blacklist a server that responds with random external IPs...

```
...
Frame 10: 134 bytes on wire (1072 bits), 134 bytes captured (1072 bits) on interface 0
    Interface id: 0 (en0)
        Interface name: en0
        Interface description: Wi-Fi
    Encapsulation type: Ethernet (1)
    Arrival Time: May 21, 2019 10:59:34.001566000 CEST
    [Time shift for this packet: 0.000000000 seconds]
    Epoch Time: 1558429174.001566000 seconds
    [Time delta from previous captured frame: 0.096064000 seconds]
    [Time delta from previous displayed frame: 0.096064000 seconds]
    [Time since reference or first frame: 36.495842000 seconds]
    Frame Number: 10
    Frame Length: 134 bytes (1072 bits)
    Capture Length: 134 bytes (1072 bits)
    [Frame is marked: False]
    [Frame is ignored: False]
    [Protocols in frame: eth:ethertype:ip:udp:stun]
Ethernet II, Src: JuniperN_40:16:90 (28:8a:1c:40:16:90), Dst: Apple_62:d0:1e (8c:85:90:62:d0:1e)
    Destination: Apple_62:d0:1e (8c:85:90:62:d0:1e)
        Address: Apple_62:d0:1e (8c:85:90:62:d0:1e)
        .... ..0. .... .... .... .... = LG bit: Globally unique address (factory default)
        .... ...0 .... .... .... .... = IG bit: Individual address (unicast)
    Source: JuniperN_40:16:90 (28:8a:1c:40:16:90)
        Address: JuniperN_40:16:90 (28:8a:1c:40:16:90)
        .... ..0. .... .... .... .... = LG bit: Globally unique address (factory default)
        .... ...0 .... .... .... .... = IG bit: Individual address (unicast)
    Type: IPv4 (0x0800)
Internet Protocol Version 4, Src: 216.93.246.18, Dst: 172.16.33.35
    0100 .... = Version: 4
    .... 0101 = Header Length: 20 bytes (5)
    Differentiated Services Field: 0x00 (DSCP: CS0, ECN: Not-ECT)
        0000 00.. = Differentiated Services Codepoint: Default (0)
        .... ..00 = Explicit Congestion Notification: Not ECN-Capable Transport (0)
    Total Length: 120
    Identification: 0x0000 (0)
    Flags: 0x4000, Don't fragment
        0... .... .... .... = Reserved bit: Not set
        .1.. .... .... .... = Don't fragment: Set
        ..0. .... .... .... = More fragments: Not set
        ...0 0000 0000 0000 = Fragment offset: 0
    Time to live: 54
    Protocol: UDP (17)
    Header checksum: 0xa8d1 [validation disabled]
    [Header checksum status: Unverified]
    Source: 216.93.246.18
    Destination: 172.16.33.35
User Datagram Protocol, Src Port: 3478, Dst Port: 22000
    Source Port: 3478
    Destination Port: 22000
    Length: 100
    Checksum: 0xae0f [unverified]
    [Checksum Status: Unverified]
    [Stream index: 0]
    [Timestamps]
        [Time since first frame: 36.495842000 seconds]
        [Time since previous frame: 0.096064000 seconds]
Session Traversal Utilities for NAT
    [Request In: 9]
    [Time: 0.096064000 seconds]
    Message Type: 0x0101 (Binding Success Response)
        .... ...1 ...0 .... = Message Class: 0x10 Success Response (2)
        ..00 000. 000. 0001 = Message Method: 0x0001 Binding (0x001)
        ..0. .... .... .... = Message Method Assignment: IETF Review (0x0)
    Message Length: 72
    Message Cookie: 2112a442
    Message Transaction ID: 3c2ae4dcebe338ad882781c2
    Attributes
        MAPPED-ADDRESS: 65.26.180.201:53442
            Attribute Type: MAPPED-ADDRESS (0x0001)
                0... .... .... .... = Attribute Type Comprehension: Required (0x0)
                .0.. .... .... .... = Attribute Type Assignment: IETF Review (0x0)
            Attribute Length: 8
            Reserved: 00
            Protocol Family: IPv4 (0x01)
            Port: 53442
            IP: 65.26.180.201
        SOURCE_ADDRESS (Deprecated): 67.83.249.83:15908
            Attribute Type: SOURCE_ADDRESS (0x0004)
                0... .... .... .... = Attribute Type Comprehension: Required (0x0)
                .0.. .... .... .... = Attribute Type Assignment: IETF Review (0x0)
            Attribute Length: 8
            Reserved: 00
            Protocol Family: IPv4 (0x01)
            Port: 15908
            IP: 67.83.249.83
        CHANGED_ADDRESS (Deprecated): 36.77.233.34:2242
            Attribute Type: CHANGED_ADDRESS (0x0005)
                0... .... .... .... = Attribute Type Comprehension: Required (0x0)
                .0.. .... .... .... = Attribute Type Assignment: IETF Review (0x0)
            Attribute Length: 8
            Reserved: 00
            Protocol Family: IPv4 (0x01)
            Port: 2242
            IP: 36.77.233.34
        XOR-MAPPED-ADDRESS: 67.112.22.185:52830
            Attribute Type: XOR-MAPPED-ADDRESS (0x8020)
                1... .... .... .... = Attribute Type Comprehension: Optional (0x1)
                .0.. .... .... .... = Attribute Type Assignment: IETF Review (0x0)
            Attribute Length: 8
            Reserved: 00
            Protocol Family: IPv4 (0x01)
            Port (XOR-d): ef4c
            [Port: 52830]
            IP (XOR-d): 6262b2fb
            [IP: 67.112.22.185]
        SOFTWARE
            Attribute Type: SOFTWARE (0x8022)
                1... .... .... .... = Attribute Type Comprehension: Optional (0x1)
                .0.. .... .... .... = Attribute Type Assignment: IETF Review (0x0)
            Attribute Length: 20
            Software: Vovida.org 0.98-CPC
```

lib/connections/quic_misc.go
```diff
@@ -0,0 +1,141 @@
...
+	quicConfig         = &quic.Config{
+		ConnectionIDLength: 4,
+		KeepAlive:          true,
+	}
+)
+
+// Track connections on which we are listening on, to increase the probability of accessing a connection that
+// has a NAT port mapping. This also makes our outgoing port stable and same as incoming port which should allow
+// better probability of punching through.
+func getListeningConn() net.PacketConn {
```
calmh: I think this stuff needs handling for quic4/quic6, lest we have a listener `quic4://...` and try to use that socket to dial an IPv6 address, or vice versa, which will fail.

lib/connections/quic_listen.go
```diff
@@ -0,0 +1,325 @@
...
+	defer stunConn.Close()
+	defer quicConn.Close()
+	defer deregisterConn(quicConn)
+	defer filterConn.Close()
+	defer packetConn.Close()
+
+	l.Infof("QUIC listener (%v) starting", quicConn.LocalAddr())
+	defer l.Infof("QUIC listener (%v) shutting down", quicConn.LocalAddr())
+
+	go t.stunRenewal(stunConn)
```
calmh: Only if NAT traversal is enabled, otherwise no STUN

lib/config/wrapper.go
```diff
@@ -105,6 +107,27 @@ type wrapper struct {
...
+			addresses = append(addresses, DefaultStunServers...)
+		default:
+			addresses = append(addresses, addr)
+		}
+	}
+
+	addresses = util.UniqueStrings(addresses)
+
+	// Shuffle
+	rand.Shuffle(len(addresses), func(i, j int) {
```
calmh: This will be deterministic, mostly, as we don't seed the default random generator afaik. Maybe create a new `Rand` based on our `secureSource` and shuffle on that. Overkill, but at least [guaranteed to be random](https://xkcd.com/221/).

lib/config/config.go
```diff
@@ -39,13 +39,16 @@ const (
 var (
 	// DefaultTCPPort defines default TCP port used if the URI does not specify one, for example tcp://0.0.0.0
 	DefaultTCPPort = 22000
+	// DefaultQUICPort defines default QUIC port used if the URI does not specify one, for example quic://0.0.0.0
+	DefaultQUICPort = 22442
```
calmh: I really feel we should use 22000 here as well as it's sort of mentally ingrained as the Syncthing port. I forget what the reason was to do something else? (But I remember we did talk about this before.)

calmh:
On the positive side it connected immediately with NAT on both sides. One side was behind my oddity Juniper firewall/NAT, the other side tethered and thus NAT:ed behind my iPhone/4G. 👍

AudriusButkevicius:
This now forces a requirement on Go 1.12 due to quic library.
Addressed review comments, testing stays limited, as by standard.

calmh:
That's not really optimal, lots of people building using 1.11 still (including us for Android, but that's easily fixed). We could hide the quic stuff behind a build tag, but then things would error when attempting to listen on it and it would be noisy... Not 100% sure on what's best here yet.

calmh:
```
panic: interface conversion: *pfilter.FilteredConn is not net.Conn: missing method Read

goroutine 106 [running]:
github.com/syncthing/syncthing/lib/connections.(*quicListener).Serve(0xc000158000)
        /Users/jb/dev/github.com/syncthing/syncthing/lib/connections/quic_listen.go:74 +0x41b
github.com/thejerf/suture.(*Supervisor).runService.func1(0xc0001c6690, 0xc000000000, 0x54b4728, 0xc000158000)
        /Users/jb/go/pkg/mod/github.com/thejerf/suture@v3.0.2+incompatible/supervisor.go:600 +0x47
created by github.com/thejerf/suture.(*Supervisor).runService
        /Users/jb/go/pkg/mod/github.com/thejerf/suture@v3.0.2+incompatible/supervisor.go:588 +0x5b
```

calmh:
I made the Registry take net.PacketConns instead and removed the assertion, but maybe we want the registry to be more general (if we ever implement tcp reuse), in which case a wrapper would be nicer. Leaving that for your consideration, testing with my hacked version for the moment :)

AudriusButkevicius:
The purpose of the registry was tcp reuse. But that will be net.Addr or even string at which point the registry will become interface{}.

I guess we can just make it disabled on go 1.11.
Sad that there are no build tags for go versions.

calmh:
Oh, there are. `//+build go1.12` is what we need here I expect.

By the way STUN issue is more general than a single flaky server. I suspect there's some version of it we don't understand or something like that (didn't look yet);

```
[NRNJ6] 06:03:56 INFO: quic://:22000 detected NAT type: Full cone NAT
[NRNJ6] 06:03:56 INFO: quic://:22000 resolved external address quic://34.243.42.177:62808 (via stun.ekiga.net:3478)
[NRNJ6] 06:04:14 INFO: quic://:22000 resolved external address quic://69.59.119.98:3980 (via stun.ekiga.net:3478)
[NRNJ6] 06:04:32 INFO: quic://:22000 resolved external address quic://108.250.203.150:53145 (via stun.ekiga.net:3478)
[NRNJ6] 06:04:50 INFO: quic://:22000 resolved external address quic://86.223.254.137:61378 (via stun.ekiga.net:3478)
[NRNJ6] 06:04:50 INFO: quic://:22000 resolved external address quic://69.223.147.21:27114 (via stun.counterpath.net:3478)
[NRNJ6] 06:05:08 INFO: quic://:22000 resolved external address quic://38.21.190.191:55078 (via stun.counterpath.net:3478)
[NRNJ6] 06:05:44 INFO: quic://:22000 resolved external address quic://105.68.213.194:14291 (via stun.counterpath.net:3478)
[NRNJ6] 06:05:44 INFO: quic://:22000 detected NAT type: Not behind a NAT
[NRNJ6] 06:05:44 INFO: quic://:22000 resolved external address quic://176.9.4.205:22000 (via stun.schlund.de:3478)
```

AudriusButkevicius:
I mean if wireshark doesn't understand it, I am not sure who does.

calmh:
Mmm. Nonetheless, other STUN clients manage to get sane responses from these servers. I just tried Syncthing plus stunclient from the stuntman package. The requests look different (Syncthing sends FINGERPRINT and SOFTWARE attributes, stunclient sends just a CHANGE_REQUEST attribute), Syncthing gets a gibberish response and stunclient gets a correct one. The bug might not be on our side, but then perhaps we are sending too fancy attributes or something...

AudriusButkevicius:
I think we send keepalives periodically, where the response to that is not well defined.

calmh:
Setting the software to the empty string makes it work.

```
diff --git a/lib/connections/quic_listen.go b/lib/connections/quic_listen.go
index 11eef9661..6dc13af28 100644
--- a/lib/connections/quic_listen.go
+++ b/lib/connections/quic_listen.go
@@ -194,7 +194,7 @@ func (t *quicListener) NATType() string {

 func (t *quicListener) stunRenewal(listener net.PacketConn) {
        client := stun.NewClientWithConnection(listener)
-       client.SetSoftwareName("syncthing")
+       client.SetSoftwareName("")

        var natType stun.NATType
        var extAddr *stun.Host
```

We still send a software attribute, but zero length.  There's charitable and uncharitable interpretations on what the server might be doing here, but at the very least an unexpected software attribute seems to trip some of them up... Ideally the stun package might learn to not send the attribute when the software is the empty string, but it doesn't seem critical.

calmh:
There is now also `stun.syncthing.net` (the discovery servers).

AudriusButkevicius:
The server could choose not to reply instead of sending garbage if it was private. I also think there is some crude implementation of auth in stun.

But yeah, running our own stun would be best, given we already run discovery which is essentially stun over tcp. The only problem is that every device will send a packet every 18s which might be a lot.

Also, it either requires coordination of two discovery servers or two ips per discovery server.

calmh:
Oh, shit, I didn't notice that part about deploying a stun server...

calmh:
```
$ go test -v -run LargeFile8G -tags integration\ benchmark

TCP:
2019/05/22 13:54:00 Result: 3.64e+05 KiB/s synced

QUIC:
2019/05/22 13:59:45 Result: 4.46e+04 KiB/s synced
```

This is totally fine by me under the circumstances 👍

AudriusButkevicius:
So I've read somewhere that UDP mappings on routers live for around 20s, hence the 18s keepalive. In practice, if there are writes happening (dialing for example), we wouldn't need to do it so often.

I guess we could create a lastWriteTrackingPacketConn{} wrapper, but I wonder how likely are we going to be using that connection. I suspect in most cases it will stay at "garbage every 18s".

We could also probe the router by increasing that timeout every loop until we hit an address change, and then going back one increment.

Or we could do both.

Can we not have two ips per vm on digitalocean?

I am genuinely interested about the proportions of nat types on the internet, so it's my pet peeve to have that. I saw this information as a reward for doing this work and work exposing it in UR.

AudriusButkevicius:
So I've read somewhere that UDP mappings on routers live for around 20s, hence the 18s keepalive. In practice, if there are writes happening (dialing for example), we wouldn't need to do it so often.

I guess we could create a lastWriteTrackingPacketConn{} wrapper, but I wonder how likely are we going to be using that connection. I suspect in most cases it will stay at "garbage every 18s".

We could also probe the router by increasing that timeout every loop until we hit an address change, and then going back one increment. (Another interesting metric to collect)

Or we could do both.

Can we not have two ips per vm on digitalocean?

I am genuinely interested about the proportions of nat types on the internet, so it's my pet peeve to have that. I saw this information as a reward for doing this work and work exposing it in UR.

Catfriend1:
Just a quick info: I'm building on go 1.12.1 for Android and it works fine.

calmh:
You can't get two "straight" IPs on a droplet, but you can assign floating IPs which get NAT:ed in towards the droplet. I think the stuntman server supports this, I'll look into it.

I was off by a factor thousand so I think the traffic should be manageable today. Roughly 200k clients (according to the discovery server) * 128 bytes (roughly, per discovery, according to wireshark) / 18 seconds per discovery * 86400 seconds per day ~= 114 GiB of STUN per day ~= 3.4 TiB of STUN per month. It's not nothing but if fits within the quotas we have today.

The interval should be configurable though.

We can do the probing experiments but let's keep it simple for now. One exception that I think might be worth doing; when we detect "no NAT" we immediately bump the interval to ten minutes or so... Another one which is probably quite rare would be static NAT (address changes but port doesn't), which won't need keepalives and can also use the higher interval...

AudriusButkevicius:
Actually reading around it seems most things have default nat timeout of 2 minutes for UDP.
It would be interesting to see what others are doing here.

Also, our keepalive seems to to be a wrong type of message that gets a response.

calmh:
So increasing the default interval to a minute is already a significant reduction in traffic and feels better.

The IP situation on DigitalOcean is a bit annoying. The floating IP thing is fine, but doesn't help because the stun server needs two separate physical interfaces to listen on, because it needs to listen on both port 3478 and 3479 on both the primary and alternate address. I'll set this up with a relay VM to forward the alternate traffic into the VM-private network, which *is* a separate interface from the public one.

AudriusButkevicius:
You could also setup forwarding between the discovery servers perhaps somehow so we don't need a side vm.

calmh:
Yeah. I was hoping to keep it simple enough that I don't need to deploy IP tunnels or VPNs and stuff, but I'm having some issues with the DO private networking at the moment. Having their support look into it. One way or the other we'll sort it out I'm sure.

calmh:
It looks like the networking at DO might not fly for this purpose, at least not without overwhelming complexity (forwarding, all of the IP tunneling stuff, UDP encapsulation, NAT in both directions, source routing, etc) which I'm not excited about understanding and maintaining in a year or five...

AWS is more powerful, but charges for all the outgoing traffic. That's about 100 bytes for every mapping request, charged at $0.089 per GiB. That's on the order of $238 monthly for the 18s case, which is out of scope. The 60s case is still ~$70 today and will only increase, so I'm not excited about that either really.

We could of course just use the public server "pool" but ...

calmh:
OK, strike all of the above. I got it to work fine with just DO floating IP:s, so we're good. We should still reduce the default interval to a minute or so I think.

```
jb@kvin:~ $ stunclient --mode full --verbosity 2 stun.syncthing.net
Resolved stun.syncthing.net to 198.211.120.59:0
config.fBehaviorTest = true
config.fFilteringTest = true
config.timeoutSeconds = 0
config.uMaxAttempts = 0
config.addrServer = 198.211.120.59:3478
socketconfig.addrLocal = 0.0.0.0:0
Sending message to 198.211.120.59:3478
Got response (68 bytes) from 198.211.120.59:3478 on interface 172.16.33.3:57504
Other address is 188.166.128.84:3479

Preparing message for filtering test #2 (ChangeRequest=AA)
Sending message to 198.211.120.59:3478
Got response (68 bytes) from 188.166.128.84:3479 on interface 172.16.33.3:57504
Preparing message for behavior test #2 (destination=AP)
Sending message to 188.166.128.84:3478
Got response (68 bytes) from 188.166.128.84:3478 on interface 172.16.33.3:57504
Binding test: success
Local address: 172.16.33.3:57504
Mapped address: 83.233.120.221:2583
Behavior test: success
Nat behavior: Endpoint Independent Mapping
Filtering test: success
Nat filtering: Endpoint Independent Filtering
```

AudriusButkevicius:
So I added the thing that decreases the keepalives from 3 minutes to less and less until it finds a stable interval.

The stun code is a complete cluster fuck. I don't know how I don't end up with such a fucking mess in other languages. I guess having goto just makes me abuse it.

Also, it does not do keepalives if writes were happening on the connection (we don't check the address tho, so for all that we know local network stuff could prevent us from doing keepalives).
If we removed it, it would probably be less of a clusterfuck.

AudriusButkevicius:
Oh, and this has a few of my original "let's change some random utility function behaviour everywhere while we are at it".

lib/connections/quic_listen.go
```diff
@@ -0,0 +1,408 @@
...
+		return "unknown"
+	}
+	return v.String()
+}
+
+func (t *quicListener) stunRenewal(listener net.PacketConn, getLastWriteTime func() time.Time) {
+	for {
+	disabled:
+		if t.stunDisabled() {
+			time.Sleep(time.Second)
```
calmh: Should check for the listener stopping here

lib/connections/quic_listen.go
```diff
@@ -0,0 +1,408 @@
...
+
+		l.Debugf("Starting stun for %s", t.uri)
+
+		t.nat.Store(stun.NATUnknown)
+		t.mut.Lock()
+		t.address = nil
+		t.mut.Unlock()
+
+		for _, addr := range t.cfg.StunServers() {
+			if !t.runStunForServer(listener, addr, getLastWriteTime) {
```
calmh: Please add some commentary here about `runStunForServer` not actually returning unless the server fails, as that part is quite deep in the call stack and it took me a little while to figure out that this doesn't buggily run one stun transaction against every server in sequence

lib/connections/quic_misc.go
```diff
@@ -0,0 +1,124 @@
...
+	if jIsUnspecified != iIsUnspecified {
+		return len(iLocalAddr.Network()) <= len(jLocalAddr.Network())
+	}
+
+	return iIsUnspecified
+}
+
+type writeTrackingPacketConn struct {
+	net.PacketConn
+	lastWrite atomic.Value
```
calmh: I'm not sure of the specifics of `atomic.Value` but the atomics stuff generally prefer being first in the struct for ensured 64 bit alignedness on 32 bit archs etc.

Maybe given this is done on every packet write we should have it be a regular int64 and store the unix nanos, to avoid atomic:ing a whole struct

lib/rand/random.go
```diff
@@ -73,3 +74,14 @@ func SeedFromBytes(bs []byte) int64 {
 	// uint64s and XOR them together.
 	return int64(binary.BigEndian.Uint64(s[0:]) ^ binary.BigEndian.Uint64(s[8:]))
 }
+
+// Shuffle the order of elements
+func Shuffle(slice interface{}) {
+	rv := reflect.ValueOf(slice)
+	swap := reflect.Swapper(slice)
+	length := rv.Len()
+	if length < 2 {
```
calmh: Might as well do this before all the reflection

lib/discover/cache.go
```diff
@@ -124,7 +124,7 @@ func (m *cachingMux) Lookup(deviceID protocol.DeviceID) (addresses []string, err
 	l.Debugln("lookup results for", deviceID)
 	l.Debugln("  addresses: ", addresses)
 
-	addresses = util.UniqueStrings(addresses)
+	addresses = util.UniqueTrimmedStrings(addresses)
```
calmh: We used to sort these now we don't. I'm not sure that matters other than aesthetically (they are shown in that order in the UI) but I see no reason to change it in this PR

lib/rand/random.go
AudriusButkevicius: It's an interface, can't do that.

calmh: Oh, right

lib/config/wrapper.go
```diff
@@ -105,6 +107,28 @@ type wrapper struct {
 	requiresRestart uint32 // an atomic bool
 }
 
+func (w *wrapper) StunServers() []string {
+	var addresses []string
+	for _, addr := range w.cfg.Options.StunServers {
+		switch addr {
+		case "default":
+			defaultPrimaryAddresses := DefaultPrimaryStunServers
+			rand.Shuffle(defaultPrimaryAddresses)
```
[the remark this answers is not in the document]
AudriusButkevicius: I guess the intetion was not to shuffle the original, and I just forgot that we need to copy.

AudriusButkevicius:
I've addressed the review comments, I'll try to get the stun into a separate package, but it seems fairly interleaved with everything.

calmh:
test fail on the address sorting, otherwise LGTM shipit

calmh:
Oh and I didn't mean to suggest a separate stun *package*, just a type in the connections package

AudriusButkevicius:
Actually, it looks quite nice when all the cruft is hidden away in a separate package, stay tuned.

AudriusButkevicius:
Have another look and take it for another test please.

calmh:
tl;dr yet, but it seems to work at least in my setup

lib/connections/config.go
```diff
@@ -8,5 +8,6 @@ package connections
 
 const (
 	tcpPriority   = 10
+	quicPriority  = 100
```
golangcibot: `quicPriority` is unused (from `deadcode`)

lib/api/mocked_config_test.go
```diff
@@ -122,6 +122,10 @@ func (m *mockedConfig) GlobalDiscoveryServers() []string {
 	return nil
 }
 
+func (m *mockedConfig) StunServers() []string {
```
golangcibot: receiver name m should be consistent with previous receiver name c for mockedConfig (from `golint`)

lib/config/config.go
```diff
@@ -65,6 +68,28 @@ var (
 	DefaultDiscoveryServers = append(DefaultDiscoveryServersV4, DefaultDiscoveryServersV6...)
 	// DefaultTheme is the default and fallback theme for the web UI.
 	DefaultTheme = "default"
+	// Default stun servers should be substituted when the configuration
```
golangcibot: comment on exported var `DefaultPrimaryStunServers` should be of the form `DefaultPrimaryStunServers ...` (from `golint`)

lib/config/optionsconfiguration.go
```diff
@@ -81,9 +84,13 @@ func (orig OptionsConfiguration) RequiresRestartOnly() OptionsConfiguration {
...
 		if len(v) > 0 && v != "true" {
-			panic(fmt.Sprintf(`unexpected tag value: %s. expected untagged or "true"`, v))
+			panic(fmt.Sprintf(`unexpected tag value: %s. Expected untagged or "true"`, v))
 		}
 		return v != "true"
 	})
 	return copy
 }
+
+func (opts OptionsConfiguration) IsStunDisabled() bool {
```
golangcibot: receiver name opts should be consistent with previous receiver name orig for OptionsConfiguration (from `golint`)

lib/stun/filter.go
```diff
@@ -0,0 +1,66 @@
...
+	id := string(out[8:20])
+	f.mut.Lock()
+	f.ids[id] = time.Now().Add(time.Minute)
+	f.reap()
+	f.mut.Unlock()
+}
+
+func (f *stunFilter) ClaimIncoming(in []byte, addr net.Addr) bool {
+	if f.isStunPayload(in) {
+		id := string(in[8:20])
```
golangcibot: SA6001: m[string(key)] would be more efficient than k := string(key); m[k] (from `staticcheck`)

lib/stun/stun.go
```diff
@@ -0,0 +1,311 @@
...
+		if nextSleep < minSleep {
+			nextSleep = minSleep
+		}
+	tryLater:
+		sleepFor := nextSleep
+		now := time.Now()
+
+		nextKeepalive := lastWrite.Add(sleepFor)
+		if nextKeepalive.After(now) {
+			sleepFor = nextKeepalive.Sub(time.Now())
```
golangcibot: S1024: should use time.Until instead of t.Sub(time.Now()) (from `gosimple`)

lib/stun/stun.go
```diff
@@ -0,0 +1,311 @@
...
+
+		if s.cfg.Options().IsStunDisabled() {
+			// Disabled, give up
+			l.Debugf("%s disabled, aborting stun ", s)
+			return false
+		}
+
+		// Check if any writes happened while we were sleeping, if they did, sleep again
+		lastWrite = s.writeTrackingPacketConn.getLastWrite()
+		if gap := time.Now().Sub(lastWrite); gap < nextSleep {
```
golangcibot: S1012: should use `time.Since` instead of `time.Now().Sub` (from `gosimple`)

lib/connections/config.go
AudriusButkevicius: Lies, get Go1.12 then speak.

lib/util/utils.go
```diff
@@ -111,21 +110,23 @@ func CopyMatchingTag(from interface{}, to interface{}, tag string, shouldCopy fu
 	}
 }
 
-// UniqueStrings returns a list on unique strings, trimming and sorting them
-// at the same time.
-func UniqueStrings(ss []string) []string {
-	var m = make(map[string]bool, len(ss))
-	for _, s := range ss {
-		m[strings.Trim(s, " ")] = true
+// UniqueTrimmedStrings returns a list on unique strings, trimming at the same time.
```
imsodin: I hope you caught all cases that relied on the sorting (and I guess there aren't many) ;)

lib/connections/quic_listen.go
```diff
@@ -0,0 +1,236 @@
...
+		// Accept blocks forever, give it 10s to do it's thing.
+		ok := make(chan struct{})
+		go func() {
+			select {
+			case <-ok:
+				return
+			case <-time.After(10 * time.Second):
+				l.Debugln("timed out waiting for AcceptStream on", session.RemoteAddr())
+				_ = session.Close()
+			}
```
imsodin: Maybe add `t.stop` to this select just to prevent a potential 10s wait.

lib/connections/quic_listen.go
```diff
@@ -0,0 +1,236 @@
...
+		t.err = err
+		t.mut.Unlock()
+		l.Infoln("Listen (BEP/quic):", err)
+		return
+	}
+
+	defer func() { _ = packetConn.Close() }()
+	defer func() { _ = conn.Close() }()
+	defer svc.Stop()
+	defer registry.Unregister(t.uri.Scheme, conn)
```
imsodin: At least parts of those should probably go above the if-block above to terminated the started stun service on that early return. And purely on personal taste I'd rather put consecutive defers into a single deferred function.

lib/connections/quic_listen.go
```diff
@@ -0,0 +1,236 @@
...
+		case <-t.stop:
+			if err == nil {
+				_ = session.Close()
+			}
+			return
+		default:
+		}
+		if err != nil {
+			if err, ok := err.(net.Error); !ok || !err.Timeout() {
+				l.Warnln("Listen (BEP/quic): Accepting connection:", err)
```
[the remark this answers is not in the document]
AudriusButkevicius: There shouldn't be errors on accept.

lib/util/utils.go
AudriusButkevicius: Yes, that's the idea. Only versioners were depending on it.

lib/connections/quic_misc_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (mockPacketConn) SetReadDeadline(t time.Time) error {
+	panic("implement me")
+}
+
+func (mockPacketConn) SetWriteDeadline(t time.Time) error {
+	panic("implement me")
+}
+
+func TestPacketConnUnspecifiedVsSpecified(t *testing.T) {
```
calmh: This set of tests is tedious and not necessarily exhaustive, the whole file could be a table driven 

```
cases := []struct{
  netA string
  addrA string
  netB string
  addrB string
}{
  // a and b must be distinct, a is always the lesser one
  { "tcp", ":1235", "tcp", "127.0.0.1:1234" }, // unspecified is lesser
  ...
}

for _,tc := range cases {
  a, b := ... // resolve addresses, wrap in mockPacketConns
  if !packetConnLess(a, b) { fail }
  if packetConnLess(b, a) { fail }
  if packetConnLess(a, a) || packetConnLess(b, b) { fail }
}
```

lib/stun/stun.go
```diff
@@ -0,0 +1,310 @@
...
+
+func (s *Service) isCurrentNATTypePunchable() bool {
+	return s.natType == NATNone || s.natType == NATPortRestricted || s.natType == NATRestricted || s.natType == NATFull
+}
+
+func areDifferent(first, second *Host) bool {
+	if (first == nil) != (second == nil) {
+		return true
+	}
+	if first != nil {
```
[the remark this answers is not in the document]
AudriusButkevicius: if their nil'ness is not equal, we exit above, the fact that first is not nil means second is not nil.

lib/connections/registry/registry.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/syncthing/syncthing/lib/sync"
+)
+
+var (
+	mut       = sync.NewMutex()
+	available = make(map[string][]interface{})
+)
+
+func Register(scheme string, item interface{}) {
```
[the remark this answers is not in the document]
AudriusButkevicius: In the tcp punchthrough we need to reach into the registry from the dialer package, so this has to be generally global.

I'll do the var Default = New() thing.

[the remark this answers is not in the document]
AudriusButkevicius: Also, the lessThan func will vary based on what's wedged into the registry.

lib/stun/stun.go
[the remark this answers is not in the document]
calmh: Yeah that piece of logic already had me squinting a bit :)

lib/connections/registry/registry.go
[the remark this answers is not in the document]
calmh: Ah, you mean for any given registry, with an existing set of whatevers, I might want to vary the less func at get-time? Ok then. Otherwise I think it would be neater to pass it in the constructor when the registry is created.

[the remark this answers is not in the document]
AudriusButkevicius: Registry is literally a global map of interface{}.
Get could return a []interface{} and then it's down to the caller to sort it I guess if you prefer it that way.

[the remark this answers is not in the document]
AudriusButkevicius: We could otherwise perhaps do a type switch in the registry and hardcode a few sorting functions.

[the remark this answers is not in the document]
calmh: Riiight, so in the end each `scheme` will probably have it's own less func? I guess then it feels to me like each scheme should be a separate registry, with the less func registered there? 

```
tcpRegistry = NewRegistry("tcp", tcpLessFunc)
```

I guess of this really needs to be *global* we could have a RegistryRegistry (cue ...BeanFactoryAdapterInterface jokes).

```
registry.Global.NewSubRegistry("tcp", tcpLessFunc)
registry.Global.Add("tcp", foo)
_ = registry.Global.Get("tcp") // uses tcpLessFunc
```

I don't know if I'm making things better or worse, just thinking out loud...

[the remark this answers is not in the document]
AudriusButkevicius: I don't really see why this is such a big deal to be honest, and why the current thing feels wrong?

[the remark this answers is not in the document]
AudriusButkevicius: I guess we can have a QuicRegistry and a TCPRegistry etc, but that's just more globals.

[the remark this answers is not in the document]
calmh: It's ok, especially if it becomes a type and not just a weird singleton package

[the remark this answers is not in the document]
AudriusButkevicius: It's a weird singleton package, but because it needs to be, because otherwise how will dialer package reach into it?

It's ok means it's ok as it is, or it's ok and  we should split it into two globals?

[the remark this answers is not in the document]
calmh: It's ok as a weird `var Default = New()` global. At least so that it's visible where the state lives. I'm tempted to say "Add a couple of tests for it" because that would make it very annoying to have it as just package magic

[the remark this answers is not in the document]
AudriusButkevicius: It's already a var Default = New() with tests in the latest commit.

calmh:
I have a feeling we're spawning too many packages, too. Imho we could have both registry and dialer as just types in the connections package, avoid exposing as many globals, just have a few variables in the connections package to hold connections-global state

AudriusButkevicius:
dialer package deals with proxy stuff, that's an existing package, hence not sure why this is mentioned.
The only new packages is stun and registry.

Stun could live in connections, but I feel doing the type aliasing and not exposing any of the external package deps from within our stun is nice.

It's nicer to do stun.NATUnknown than just NATUnknown (Because the type happens to live in the connections package based on your suggestion).

The registry thing needs to be it's own package as dialer is used by connections (and others), hence dialer cannot import connections, and putting all of this registry stuff into dialer seems wrong. I guess we could rename dialer to sockets, and then add listen functionality there which deals with the registry stuff, but even then I am not sure that's possible as the quic stuff needs access to the registry in the connections package.

calmh:
> The registry thing needs to be it's own package as dialer is used by connections (and others), hence dialer cannot import connections, and putting all of this registry stuff into dialer seems wrong.

Yes. My point was that `dialer`, `stun` and `registry` packages could have been incorporated directly into a larger `connections` package as there are interdependencies, hiding more details from the outside world and avoiding the small one-type-only packages. The nat type etc is then `natUnknown` or `stunNATUnknown` because it presumably doesn't even need exporting.

But this is just musing. I'm not proposing this as a change for this PR.

AudriusButkevicius:
Is there anything that still needs to happen in this PR?

calmh:
awesomeness

AudriusButkevicius:
I am also sure this will break stuff in amazing ways, but we won't know that until it's out in the wild.

imsodin:
Thanks a lot for this!

While we are talking about amazingly breaking stuff, would you care to have another look at https://github.com/syncthing/syncthing/pull/5673 :)

AudriusButkevicius:
I'll do once at home.

Lillecarl:
@calmh Not related to the issue, per se. But i couldn't help but mention: DO redundancy is suboptimal. https://www.digitalocean.com/products/ https://www.digitalocean.com/community/questions/is-there-a-difference-between-droplet-storage-and-block-storage

They're storing your data locally on the hypervisor node, you already know what that means for an application that doesn't have an active failover. (If i recall correctly this is also how SmartOS does it too though).

Though as mentioned, there are other STUN servers to use if yours fail. It's a tradeoff.

AudriusButkevicius:
This has been out for quite a few days, and relay usage has not dropped by much, so it's not clear this works :(

imsodin:
It does work to some degree at least, maybe not as universally as hoped for: 60% of reports are on v1.2.0 and the global transport shares are at 18% for relay and 3% for quic.

AudriusButkevicius:
It also seems a lot of the nats are port restricted

olifre:
Doesn't this require TLS 1.3 which requires users to opt-in manually for stable builds? 
This might explain why the number of users is so low.

AudriusButkevicius:
I think they use qtls which is their own library

olifre:
Thanks for the clarification. Indeed, it's unclear to me in which situations QUIC should work successfully - for me, it fails with `context deadline exceeded` between a client behind a `Port restricted NAT` (classic home router) and the other client behind a `Symmetric UDP firewall` (linux firewall, host with global IP). Other tools such as `tinc` punch through fine in this case. 
Let me know if this is a situation in which QUIC should work and if you want to debug this, then we can of course take this to a new issue :wink:.

AudriusButkevicius:
https://www.frozenmountain.com/developers/blog/webrtc-nat-traversal-methods-a-case-for-embedded-turn implies STUn does not work and you require TURN (relays).

I've seen a more detailed chart somewhere else on the internet.

Reading tinc docs:
> Regardless of how you set up the tinc daemons to connect to each other, VPN traffic is always (if possible) sent directly to the destination, without going through intermediate hops.

The if possible sounds interesting, not clear if it's sent directly if not possible.

olifre:
> https://www.frozenmountain.com/developers/blog/webrtc-nat-traversal-methods-a-case-for-embedded-turn implies STUn does not work and you require TURN (relays).

`Symmetric UDP firewall` != `Symmetric NAT`. In my case, one end is "only firewalled". It has a global IPv4 address, and no port mangling is done, but UDP traffic is blocked, unless an entry is generated in the connection tracking table of the firewall by hole-punching (i.e. a classical router firewall, but without NAT) => `Symmetric UDP firewall`. 
Maybe the code does not handle this non-NAT case correctly? 

The debug output I get in the logs from Syncthing, however, correctly indicates the external IP addresses of both ends (i.e. in one case the global IP, in the other case the external IP of the home router), so I do not think TURN is needed. 

> The if possible sounds interesting, not clear if it's sent directly if not possible.

They attempt direct connections continuously, and probe the maximum UDP package size if a connection can be established with tiny packages. If it does not work, intermediate hops are used (i.e. automatic meshing: Other members of the VPN automatically become relays). Note that `tinc` also does not need STUN: Information about external IP addresses is exchanged between the nodes, if possible (of course that ideally means one node must be publicly reachable, which then automagically becomes the exchange point). In short, it's a layer2/layer3 meshing system with integrated and reliably working hole punching and relaying.

olifre:
>  Maybe the code does not handle this non-NAT case correctly?

My knowledge of Go is zero, but maybe it's as easy as adding `NATSymmetricUDPFirewall` to the big OR in `isCurrentNATTypePunchable()`? I don't see why it should not be punchable - `NATSymmetricUDPFirewall` changes neither IP nor port, there's no `NAT` (the constant name is rather misleading, RFC 3489 also does not prefix that with "NAT"), it's just firewalled :wink:.

AudriusButkevicius:
Add it to the list, compile it, and see if it works.
I'd like to get to the bottom of this, but this should really continue on the forum, potentially with debug logs for connections and stun package.

olifre:
> I'd like to get to the bottom of this, but this should really continue on the forum, not an old closed PR, potentially with debug logs for connections and stun package.

Here you go:
https://forum.syncthing.net/t/issues-with-quic-transport/13590
I added some other findings of different issues. Let me know in the forums if anything else is needed / unclear.

PR #5741, opened 2019-05-25
lib/protocol: Don't send anything else before cluster config

imsodin:
As per BEP cluster config should be the first message sent, which is now enforced in protocol package.

### DIFF
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -182,11 +182,12 @@ type rawConnection struct {
 	nextID    int32
 	nextIDMut sync.Mutex
 
-	outbox        chan asyncMessage
-	closed        chan struct{}
-	closeOnce     sync.Once
-	sendCloseOnce sync.Once
-	compression   Compression
+	outbox           chan asyncMessage
+	clusterConfigBox chan *ClusterConfig
+	closed           chan struct{}
+	closeOnce        sync.Once
+	sendCloseOnce    sync.Once
+	compression      Compression
 }
 
 type asyncResult struct {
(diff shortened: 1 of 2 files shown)


lib/protocol/protocol.go
```diff
@@ -182,11 +182,12 @@ type rawConnection struct {
 	nextID    int32
 	nextIDMut sync.Mutex
 
-	outbox        chan asyncMessage
-	closed        chan struct{}
-	closeOnce     sync.Once
-	sendCloseOnce sync.Once
-	compression   Compression
+	outbox           chan asyncMessage
+	clusterConfigBox chan *ClusterConfig
```
AudriusButkevicius: why pointer?

lib/protocol/protocol.go
```diff
@@ -640,6 +644,16 @@ func (c *rawConnection) send(msg message, done chan struct{}) bool {
...
+	select {
+	case cc := <-c.clusterConfigBox:
+		err := c.writeMessage(cc)
+		if err != nil {
+			c.internalClose(err)
+			return
+		}
+	case <-c.closed:
+		return
+	}
```
AudriusButkevicius: What about hello etc?

imsodin: Currently they are done with a function not bound to a connection instance. However I do have a WIP branch (that I totally forgot about) where I want to change that (to do proper error handling in that part too), so it definitely is a concern. Let's put this PR on hold until I figured that out. Thanks!

AudriusButkevicius: Actually, I think Hello happens before we wrap the connection in a protocol thing, or not?

imsodin: Exactly, and if there is any error we don't send a close message, even after exchange hellos, which isn't optimal (we have long timeouts). So I want to include hellos and do send close messages. So cluster config wont be the first thing sent anymore, it will be hello -> CC -> anything, i.e. this PR isn't sufficient and might just complicate further PRs. As there is no urgent need for this fix, I want to do these two changes together (or at least in a compatible way).

AudriusButkevicius: We should not do that, as the connection would have to writed to the model to send hello's presumably, or expose some magical method.

The idea is that hello happens so early that there is no way we could accidentially share data.

lib/protocol/protocol.go
imsodin: It needs to implement the `message` interface which only the pointer does.

imsodin:
[Audrius' review comment](https://github.com/syncthing/syncthing/pull/5741#discussion_r287568526) and [Jakobs clarification on BEP](https://forum.syncthing.net/t/bep-and-sending-close-msg-before-cluster-config/13302) cleared up a few things to me. There's now just one more little change to the protocol package that I want to do and it's independent from this -> reopened.

AudriusButkevicius:
So I think the change is ok, but I don't understand how this helps to fix anything.
I believe we have checks on the other side to assert this behaviour, so I am not sure what this fixes.

imsodin:
Yes, there are checks on the other send when receiving, but they result in the connection being dropped, which is obviously not so nice. As the connection has to be started before a cluster config is sent, it is in principle possible that e.g. requests can be sent before the cluster config. That's what this PR prevents.

PR #5742, opened 2019-05-25
lib/protocol: Don't call receiver after calling Closed (fixes #4170)

### LINKED ISSUE #4170: panic: bug: ClusterConfig called on closed or nonexistent connection

imsodin:
We have to make sure, that the receiver/model isn't called after its `Closed` method is called. Due to the blocking nature of the underlying reader, we can't wait for reading to finish. Therefore reading is done in a separate routine and we wait for the routine handling the read message to terminate.

### DIFF
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -182,11 +182,14 @@ type rawConnection struct {
 	nextID    int32
 	nextIDMut sync.Mutex
 
-	outbox        chan asyncMessage
-	closed        chan struct{}
-	closeOnce     sync.Once
-	sendCloseOnce sync.Once
-	compression   Compression
+	inbox               chan message
+	outbox              chan asyncMessage
+	clusterConfigBox    chan *ClusterConfig
+	receiverLoopStopped chan struct{}
+	closed              chan struct{}
+	closeOnce           sync.Once
+	sendCloseOnce       sync.Once
+	compression         Compression
 }
 
 type asyncResult struct {
@@ -220,15 +223,18 @@ func NewConnection(deviceID DeviceID, reader io.Reader, writer io.Writer, receiv
 	cw := &countingWriter{Writer: writer}
 
 	c := rawConnection{
-		id:          deviceID,
-		name:        name,
-		receiver:    nativeModel{receiver},
-		cr:          cr,
-		cw:          cw,
-		awaiting:    make(map[int32]chan asyncResult),
-		outbox:      make(chan asyncMessage),
-		closed:      make(chan struct{}),
-		compression: compress,
+		id:                  deviceID,
+		name:                name,
+		receiver:            nativeModel{receiver},
+		cr:                  cr,
+		cw:                  cw,
+		awaiting:            make(map[int32]chan asyncResult),
+		inbox:               make(chan message),
+		outbox:              make(chan asyncMessage),
+		clusterConfigBox:    make(chan *ClusterConfig),
+		receiverLoopStopped: make(chan struct{}),
+		closed:              make(chan struct{}),
+		compression:         compress,
 	}
 
 	return wireFormatConnection{&c}
@@ -237,8 +243,9 @@ func NewConnection(deviceID DeviceID, reader io.Reader, writer io.Writer, receiv
 // Start creates the goroutines for sending and receiving of messages. It must
 // be called exactly once after creating a connection.
 func (c *rawConnection) Start() {
+	go c.readerLoop()
 	go func() {
-		err := c.readerLoop()
+		err := c.receiverLoop()
 		c.internalClose(err)
 	}()
 	go c.writerLoop()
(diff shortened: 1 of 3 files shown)


lib/protocol/protocol.go
```diff
@@ -847,6 +866,8 @@ func (c *rawConnection) internalClose(err error) {
 		}
 		c.awaitingMut.Unlock()
 
+		<-c.receiverLoopStopped
```
golangcibot: Using the variable on range scope `receiverLoopStopped` in function literal (from `scopelint`)

lib/protocol/protocol_test.go
```diff
@@ -125,6 +125,53 @@ func TestCloseOnBlockingSend(t *testing.T) {
...
+	br, bw := io.Pipe()
+
+	c0 := NewConnection(c0ID, ar, bw, m0, "c0", CompressNever).(wireFormatConnection).Connection.(*rawConnection)
+	c0.Start()
+	c1 := NewConnection(c1ID, br, aw, m1, "c1", CompressNever)
+	c1.Start()
+	c0.ClusterConfig(ClusterConfig{})
+	c1.ClusterConfig(ClusterConfig{})
+
+	c1.Index("default", nil)
```
golangcibot: Error return value of `c1.Index` is not checked (from `errcheck`)

lib/protocol/protocol.go
```diff
@@ -182,11 +182,14 @@ type rawConnection struct {
...
 	nextIDMut sync.Mutex
 
-	outbox        chan asyncMessage
-	closed        chan struct{}
-	closeOnce     sync.Once
-	sendCloseOnce sync.Once
-	compression   Compression
+	inbox               chan message
+	outbox              chan asyncMessage
+	clusterConfigBox    chan *ClusterConfig
```
AudriusButkevicius: Not used?

lib/protocol/protocol.go
```diff
@@ -348,25 +355,37 @@ func (c *rawConnection) ping() bool {
...
-			// Unknown message types are skipped, for future extensibility.
-			continue
-		}
-		if err != nil {
-			return err
-		}
+	}
+}
 
+func (c *rawConnection) receiverLoop() (err error) {
```
AudriusButkevicius: dispatcherLoop?

lib/protocol/protocol.go
```diff
@@ -237,8 +243,9 @@ func NewConnection(deviceID DeviceID, reader io.Reader, writer io.Writer, receiv
 // Start creates the goroutines for sending and receiving of messages. It must
 // be called exactly once after creating a connection.
 func (c *rawConnection) Start() {
+	go c.readerLoop()
 	go func() {
-		err := c.readerLoop()
+		err := c.receiverLoop()
 		c.internalClose(err)
```
AudriusButkevicius: I guess this will result in a double close, but it's a sync.Once so it's ok.
The alternative would be to call internalClose with within the receiver, and not call internalClose on <-c.close.
But it doesn't matter much I guess.

lib/protocol/protocol_test.go
```diff
@@ -125,6 +125,53 @@ func TestCloseOnBlockingSend(t *testing.T) {
...
+	select {
+	case <-c0.closed:
+	case <-time.After(time.Second):
+		t.Fatal("timed out before c0.closed was closed")
+	}
+
+	select {
+	case <-m0.closedCh:
+		t.Errorf("receiver.Closed called before receiver.Index")
+	default:
```
AudriusButkevicius: This is somewhat racey between the internalClose doing it's thing and actually getting to this point to check.
For all that we know interalClose hasn't even run yet, hence this channel is not closed.

lib/protocol/protocol.go
AudriusButkevicius: I am surprised it's complaining about this but not any other variable.

lib/protocol/protocol.go
imsodin: >  But it doesn't matter much I guess.

That. And that I don't want to touch this code unless I have to and it's tested (and I will touch it :) ).

lib/protocol/protocol.go
imsodin: I don't understand what it says at all xD

AudriusButkevicius: It's saying that the function refers to some variable that is defined outside of it's scope.
ala 
for i := range {
    go func() {
        i   // omg you are using i
    }()
}

lib/protocol/protocol_test.go
imsodin: It should be, but the chain of events is indeed a bit obscure:
1. Sending index, which blocks the `receiver-/dispatcherloop`.
2. Calling `internalClose` in a goroutine: Closes `c0.closed`, then blocks on waiting for the `receiver-/dispatcherloop` to terminate.
3. Check that `c0` is closed and `m0` did not yet receive a call to `Closed`.
4. Unblock sending index, which unblocks and then lets the `receiver-/dispatcherloop` terminate.
5. Check that now `m0` got `Closed`.

lib/protocol/protocol.go
```diff
@@ -847,6 +864,8 @@ func (c *rawConnection) internalClose(err error) {
 		}
 		c.awaitingMut.Unlock()
 
+		<-c.dispatcherLoopStopped
```
golangcibot: Using the variable on range scope `dispatcherLoopStopped` in function literal (from `scopelint`)

imsodin:
That's a bummer - looks like there's another mechanism that can trigger that panic: https://build.syncthing.net/viewLog.html?buildId=39591&buildTypeId=Syncthing_BuildLinuxCross
I'd like to hope it's just the 5063 test being broken, but that's probably wishful thinking.

lib/protocol/protocol.go
imsodin: But it isn't defined outside of it's scope?!

AudriusButkevicius:
It seems that after the last PR the test is consistently failing now.

imsodin:
> It seems that after the last PR the test is consistently failing now.

I found out why, that "on-the-side-change" in that PR, where I reverted the special close handling was necessary. Adding some milliseconds delay to the test I was able to reproduce it. The panic is irrelevant in any case. And the test seems useless too, do you know what it is supposed to do?

AudriusButkevicius:
The test is not useless, essentially there was a config corruption race in autoaccept and that is what it's testing.
I am not sure why it blows up with a different panic now, but it should be fixed.

AudriusButkevicius:
needs a rebase I guess

PR #5746, opened 2019-05-27
lib/protocol: Prioritize close msg and add close timeout

imsodin:
Currently sending the close message has the same priority as sending any other message, so might block for a long time (ping timeout, i.e. 90s), same if the underlying connection is not working. If we knowingly close the connection (folder restart or whatever) there's no point in waiting that long. I added `CloseTimeout` variable to set how long we do wait, and set it to 10s, which is simply magic, so feel free to propose another.

### DIFF
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -184,6 +184,7 @@ type rawConnection struct {
 
 	inbox                 chan message
 	outbox                chan asyncMessage
+	closeBox              chan asyncMessage
 	clusterConfigBox      chan *ClusterConfig
 	dispatcherLoopStopped chan struct{}
 	closed                chan struct{}
@@ -218,6 +219,11 @@ const (
 	ReceiveTimeout = 300 * time.Second
 )
 
+// CloseTimeout is the longest we'll wait when trying to send the close
+// message before just closing the connection.
+// Should not be modified in production code, just for testing.
+var CloseTimeout = 10 * time.Second
+
 func NewConnection(deviceID DeviceID, reader io.Reader, writer io.Writer, receiver Model, name string, compress Compression) Connection {
 	cr := &countingReader{Reader: reader}
 	cw := &countingWriter{Writer: writer}
@@ -231,6 +237,7 @@ func NewConnection(deviceID DeviceID, reader io.Reader, writer io.Writer, receiv
 		awaiting:              make(map[int32]chan asyncResult),
 		inbox:                 make(chan message),
 		outbox:                make(chan asyncMessage),
+		closeBox:              make(chan asyncMessage),
 		clusterConfigBox:      make(chan *ClusterConfig),
 		dispatcherLoopStopped: make(chan struct{}),
 		closed:                make(chan struct{}),
@@ -683,6 +690,11 @@ func (c *rawConnection) writerLoop() {
 				return
 			}
 
+		case hm := <-c.closeBox:
+			_ = c.writeMessage(hm.msg)
+			close(hm.done)
+			return
+
 		case <-c.closed:
 			return
 		}
@@ -850,17 +862,20 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
 func (c *rawConnection) Close(err error) {
 	c.sendCloseOnce.Do(func() {
 		done := make(chan struct{})
-		c.send(&Close{err.Error()}, done)
+		timeout := time.NewTimer(CloseTimeout)
 		select {
-		case <-done:
+		case c.closeBox <- asyncMessage{&Close{err.Error()}, done}:
+			select {
+			case <-done:
+			case <-timeout.C:
+			case <-c.closed:
+			}
+		case <-timeout.C:
 		case <-c.closed:
 		}
 	})
 
-	// No more sends are necessary, therefore further steps to close the
-	// connection outside of this package can proceed immediately.
-	// And this prevents a potential deadlock due to calling c.receiver.Closed
-	go c.internalClose(err)
+	c.internalClose(err)
 }
 
 // internalClose is called if there is an unexpected error during normal operation.
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -675,6 +675,10 @@ func (c *rawConnection) writerLoop() {
 			c.internalClose(err)
 			return
 		}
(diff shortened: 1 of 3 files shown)


lib/protocol/protocol.go
```diff
@@ -850,17 +862,20 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
 func (c *rawConnection) Close(err error) {
 	c.sendCloseOnce.Do(func() {
 		done := make(chan struct{})
-		c.send(&Close{err.Error()}, done)
+		timeout := time.NewTimer(CloseTimeout)
```
golangcibot: Using the variable on range scope `time` in function literal (from `scopelint`)

lib/protocol/protocol.go
```diff
@@ -850,17 +862,20 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
 		select {
-		case <-done:
+		case c.closeBox <- asyncMessage{&Close{err.Error()}, done}:
```
golangcibot: Using the variable on range scope `closeBox` in function literal (from `scopelint`)

lib/protocol/protocol.go
```diff
@@ -850,17 +862,20 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
+			select {
+			case <-done:
+			case <-timeout.C:
```
golangcibot: Using the variable on range scope `C` in function literal (from `scopelint`)

lib/protocol/protocol.go
```diff
@@ -850,17 +862,20 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
+			case <-c.closed:
```
golangcibot: Using the variable on range scope `closed` in function literal (from `scopelint`)

lib/protocol/protocol.go
```diff
@@ -850,17 +862,20 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
+			}
+		case <-timeout.C:
```
golangcibot: Using the variable on range scope `C` in function literal (from `scopelint`)

calmh:
scopelint seems completely broken. I'll disable that at some point.

lib/protocol/protocol.go
```diff
@@ -850,17 +862,20 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
 		case <-c.closed:
 		}
 	})
 
-	// No more sends are necessary, therefore further steps to close the
-	// connection outside of this package can proceed immediately.
-	// And this prevents a potential deadlock due to calling c.receiver.Closed
-	go c.internalClose(err)
+	c.internalClose(err)
```
calmh: Now you're scaring me again by removing anti deadlock stuff without comment...

calmh:
Note also that we already do some write deadline contortions on close in other places. You might want to take that into account / just do the same thing. The reason for that is that `(*tls.Conn).Close()` is not just a close, it writes a TLS alert and then closes the connection. It's essentially the same thing, so should probably be handled the same.

lib/protocol/protocol.go
imsodin: Don't be scared, there's tests now :)

But your right, I should have commented on it: The test added in https://github.com/syncthing/syncthing/pull/5707 checks this deadlock scenario and it is obsolete since closing connections doesn't happen under `model.pmut` lock anymore.

imsodin:
Right, forgot about that. There we only wait 250ms which seems a bit short for going through the entire proto layer, doesn't it?

calmh:
Well att that stage we're essentially just waiting for the write syscall which is either instantaneous (the message fit in the connection's outgoing tcp buffer) or not (the buffer is full so we've been blocked for a long time)...

imsodin:
On the protocol side of things we potentially wait for the current write operation to finish (e.g. request response) before sending the close message, including stuff like marshalling and compression. Maybe I am overcautious and that still fits easily in 250ms even on a low-powered device?

calmh:
No I agree and we are not in that kind of rush anyway. A few seconds seems fine to wait for this.

AudriusButkevicius:
What about close before cluster config?

lib/protocol/protocol.go
```diff
@@ -850,17 +866,20 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
 func (c *rawConnection) Close(err error) {
 	c.sendCloseOnce.Do(func() {
 		done := make(chan struct{})
-		c.send(&Close{err.Error()}, done)
+		timeout := time.NewTimer(CloseTimeout)
```
golangcibot: Using the variable on range scope `time` in function literal (from `scopelint`)

lib/protocol/protocol.go
```diff
@@ -850,17 +866,20 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
 		select {
-		case <-done:
+		case c.closeBox <- asyncMessage{&Close{err.Error()}, done}:
```
golangcibot: Using the variable on range scope `closeBox` in function literal (from `scopelint`)

lib/protocol/protocol.go
```diff
@@ -850,17 +866,20 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
+			select {
+			case <-done:
+			case <-timeout.C:
```
golangcibot: Using the variable on range scope `C` in function literal (from `scopelint`)

lib/protocol/protocol.go
```diff
@@ -850,17 +866,20 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
+			case <-c.closed:
```
golangcibot: Using the variable on range scope `closed` in function literal (from `scopelint`)

lib/protocol/protocol.go
```diff
@@ -850,17 +866,20 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
+			}
+		case <-timeout.C:
```
golangcibot: Using the variable on range scope `C` in function literal (from `scopelint`)

AudriusButkevicius:
So lgtm, but I'll let @calmh do the honours

PR #5749, opened 2019-05-29
Create SECURITY.md

pahakalle:
### Purpose

Github added the Security tab, so putting the vulnerability reporting info there would be good


canton7:
Seems weird to start at a second-level heading, but no matter.

PR #5753, opened 2019-05-29
lib/connections: Try TCP punchthrough

AudriusButkevicius:
This is hopefully a PR with only the relevant changes, if I haven't failed to git.


lib/dialer/debug.go
```diff
@@ -0,0 +1,23 @@
...
+	"os"
+	"strings"
+
+	"github.com/syncthing/syncthing/lib/logger"
+)
+
+var (
+	l = logger.DefaultLogger.NewFacility("dialer", "Dialing connections")
+	// To run before init() of other files that log on init.
+	_ = func() error {
```
golangcibot: init$1 - result 0 (error) is always nil (from `unparam`)

lib/dialer/internal.go
```diff
@@ -160,3 +160,40 @@ func (a fallbackAddr) Network() string {
...
+	if host, _, err := net.SplitHostPort(jAddr.String()); err == nil {
+		jIsUnspecified = host == "" || net.ParseIP(host).IsUnspecified()
+	}
+
+	if jIsUnspecified != iIsUnspecified {
+		return len(iAddr.Network()) <= len(jAddr.Network())
+	}
+
+	return iIsUnspecified
+}
```
AudriusButkevicius: This is wrong, as I know from the quic stuff. Also could be refactored and unified.

lib/dialer/debug.go
AudriusButkevicius: It's done for a reason explained in the comment above.

lib/dialer/internal.go
```diff
@@ -160,3 +161,24 @@ func (a fallbackAddr) Network() string {
...
 }
+
+// Sort available addresses, preferring unspecified alddress.
+func tcpAddrLess(i interface{}, j interface{}) bool {
+	return util.AddressUnspecifiedLess(i.(*net.TCPAddr), j.(*net.TCPAddr))
+}
+
+func dialContextReusePort(ctx context.Context, network, addr string) (net.Conn, error) {
+	dialer := net.Dialer{
+		Control: ReusePortControl,
```
golangcibot: undeclared name: `ReusePortControl` (from `typecheck`)

lib/dialer/internal.go
```diff
@@ -160,3 +161,24 @@ func (a fallbackAddr) Network() string {
...
+	}
+	localAddrInterface := registry.Get(network, tcpAddrLess)
+	if localAddrInterface != nil {
+		dialer.LocalAddr = localAddrInterface.(*net.TCPAddr)
+	}
+
+	conn, err := dialer.DialContext(ctx, network, addr)
+	if dialer.LocalAddr != nil {
+		l.Debugf("dialContextReusePort dial %s %s via %s: %v", network, addr, dialer.LocalAddr, err)
```
golangcibot: undeclared name: `l` (from `typecheck`)

lib/connections/tcp_listen.go
```diff
@@ -167,6 +179,14 @@ func (t *tcpListener) WANAddresses() []*url.URL {
 		}
 	}
 	t.mut.RUnlock()
+
+	// If we support ReusePort, add an unspecified zero port address, which will be resolved by the discovery server
+	// in hopes that TCP punch through works.
+	if dialer.SupportsReusePort {
```
golangcibot: SupportsReusePort not declared by package dialer (from `typecheck`)

lib/connections/tcp_listen.go
```diff
@@ -56,15 +58,22 @@ func (t *tcpListener) Serve() {
 		return
 	}
 
-	listener, err := net.ListenTCP(t.uri.Scheme, tcaddr)
+	lc := net.ListenConfig{
+		Control: dialer.ReusePortControl,
```
golangcibot: ReusePortControl not declared by package dialer (from `typecheck`)

lib/dialer/debug.go
golangcibot: init$1 - result 0 (error) is always nil (from `unparam`)

AudriusButkevicius:
Ready for review maybe? Given 1.2.0 is out, and quic has not broken the world, let's try to do it again.

lib/dialer/internal.go
```diff
@@ -123,12 +123,13 @@ func getDialer(forward proxy.Dialer) proxy.Dialer {
...
-type timeoutDirectDialer struct {
-	timeout time.Duration
+type contextDirectDialer struct {
+	ctx context.Context
 }
 
-func (d *timeoutDirectDialer) Dial(network, addr string) (net.Conn, error) {
-	return net.DialTimeout(network, addr, d.timeout)
+func (d *contextDirectDialer) Dial(network, addr string) (net.Conn, error) {
+	dialer := net.Dialer{}
```
imsodin: You could embed `net.Dialer` into `contextDirectDialer`.

lib/connections/tcp_dial.go
```diff
@@ -33,7 +34,9 @@ type tcpDialer struct {
 func (d *tcpDialer) Dial(id protocol.DeviceID, uri *url.URL) (internalConn, error) {
 	uri = fixupPort(uri, config.DefaultTCPPort)
 
-	conn, err := dialer.DialTimeout(uri.Scheme, uri.Host, 10*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	conn, err := dialer.DialContextReusePort(ctx, uri.Scheme, uri.Host)
+	cancel()
```
imsodin: `dialer` is used below, so defer?

lib/dialer/control_unix.go
```diff
@@ -0,0 +1,51 @@
...
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// +build !solaris,!windows
+
+package dialer
+
+import (
+	"syscall"
+
+	"golang.org/x/sys/unix"
```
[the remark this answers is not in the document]
AudriusButkevicius: Sure, I will convert most to use unix, but net.Listen.Control passes a syscall.RawConn.

lib/util/utils.go
```diff
@@ -175,6 +176,22 @@ func Address(network, host string) string {
 	return u.String()
 }
 
+func AddressUnspecifiedLess(a, b net.Addr) bool {
```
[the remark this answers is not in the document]
AudriusButkevicius: gl?

lib/connections/tcp_listen.go
```diff
@@ -49,12 +51,19 @@ func (t *tcpListener) serve(stop chan struct{}) error {
 		return err
 	}
 
-	listener, err := net.ListenTCP(t.uri.Scheme, tcaddr)
+	lc := net.ListenConfig{
+		Control: dialer.ReusePortControl,
+	}
+
```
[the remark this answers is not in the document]
AudriusButkevicius: I find it more readable this way.

lib/connections/tcp_dial.go
AudriusButkevicius: dialer is a package name in this case, so cancel just releases resources associated with the context constructed above post dial

lib/connections/tcp_listen.go
```diff
@@ -72,9 +81,12 @@ func (t *tcpListener) serve(stop chan struct{}) error {
 	acceptFailures := 0
 	const maxAcceptFailures = 10
 
+	// :(, but what can you do.
```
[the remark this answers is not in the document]
AudriusButkevicius: net.Listener does not have SetDeadline. We used to net.ListenTCP with returned a TcpListener, now we do ListenConfig.Listen which returns Listener.

lib/util/utils.go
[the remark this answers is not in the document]
imsodin: Yeah, I checked midsentence whether what I'm writing makes sense and it didn't, but i forgot to remove that half-sentence.

lib/dialer/public.go
```diff
@@ -89,11 +52,48 @@ func SetTrafficClass(conn net.Conn, class int) error {
...
+		var timeout time.Duration
+		if deadline, ok := ctx.Deadline(); ok {
+			timeout = deadline.Sub(deadline)
+		}
+		if conn, err := dialer.DialContext(ctx, network, addr); noFallback || err == nil {
+			return conn, err
+		}
+		// If the deadline was set, reset it again for the next dial attempt.
+		if timeout != 0 {
+			ctx, _ = context.WithTimeout(ctx, timeout)
```
golangcibot: lostcancel: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (from `govet`)

lib/osutil/ping.go
```diff
@@ -18,7 +19,8 @@ import (
 // tcp.
 func TCPPing(address string) (time.Duration, error) {
 	start := time.Now()
-	conn, err := dialer.DialTimeout("tcp", address, time.Second)
+	ctx, _ := context.WithTimeout(context.Background(), time.Second)
```
golangcibot: lostcancel: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (from `govet`)

lib/relay/client/methods.go
```diff
@@ -21,7 +22,8 @@ func GetInvitationFromRelay(uri *url.URL, id syncthingprotocol.DeviceID, certs [
 		return protocol.SessionInvitation{}, fmt.Errorf("Unsupported relay scheme: %v", uri.Scheme)
 	}
 
-	rconn, err := dialer.DialTimeout("tcp", uri.Host, timeout)
+	ctx, _ := context.WithTimeout(context.Background(), timeout)
```
golangcibot: lostcancel: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (from `govet`)

lib/dialer/public.go
```diff
@@ -89,11 +52,48 @@ func SetTrafficClass(conn net.Conn, class int) error {
...
-
-	case dialerConn:
-		return SetTrafficClass(conn.Conn, class)
-
 	default:
 		return fmt.Errorf("unknown connection type %T", conn)
 	}
 }
+
+func dialContextWithFallback(fallback proxy.ContextDialer, ctx context.Context, network, addr string) (net.Conn, error) {
```
golangcibot: context.Context should be the first parameter of a function (from `golint`)

AudriusButkevicius:
I've addressed the review comments and unmarked some comments as done as it's not for me to judge when it's done.

lib/dialer/control_unix.go
```diff
@@ -0,0 +1,51 @@
...
+		l.Debugln("SO_REUSEPORT not supported")
+	case err != nil:
+		l.Debugln("Unknown error when determining SO_REUSEPORT support", err)
+	default:
+		l.Debugln("SO_REUSEPORT supported")
+		SupportsReusePort = true
+	}
+}
+
+func ReusePortControl(network, address string, c syscall.RawConn) error {
```
imsodin: `network` and `address` could be `_` to indicate that their unused.

lib/dialer/public.go
```diff
@@ -89,11 +52,50 @@ func SetTrafficClass(conn net.Conn, class int) error {
...
+			defer cancel()
+		}
+	}
+	return fallback.DialContext(ctx, network, addr)
+}
+
+// DialContext tries dialing via proxy if a proxy is configured, and falls back to
+// a direct connection if no proxy is defined, or connecting via proxy fails.
+// If the context has a timeout, the timeout might be applied twice.
+func DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
```
imsodin: The "Context" in the function name seems unnecessary to me.

lib/relay/client/dynamic.go
```diff
@@ -150,15 +151,26 @@ type dynamicAnnouncement struct {
...
+				latency = time.Hour
+			}
+			results <- urlWithLatency{irelay, latency}
+			wg.Done()
+		}(relay)
+	}
 
-		id := int(latency/time.Millisecond) / 50
+	wg.Wait()
+	close(results)
```
imsodin: I think this doesn't work: The loop below won't read anything from results, as range aborts if the looped over channel is closed. Personally I wouldn't buffer the channel and put the two lines above into a goroutine.

AudriusButkevicius: https://play.golang.org/p/4cGK8dy18Dh

lib/dialer/public.go
AudriusButkevicius: It's kind of a mimic from net.DialContext, or net.Dialer.DialContext?

lib/relay/client/dynamic.go
imsodin: Ah right, so the close kind of gets appended to the buffered elements in the channel. I guess that makes sense. Thanks for teaching me :)

AudriusButkevicius:
Editing on the mobile on a train, yay.
Anyways, addressed comments.

Time for @calmh?

lib/relay/client/dynamic.go
```diff
@@ -150,15 +151,26 @@ type dynamicAnnouncement struct {
...
+	wg := sync.NewWaitGroup()
+	wg.Add(len(input))
+	results := make(chan urlWithLatency, len(input))
 	for _, relay := range input {
-		latency, err := osutil.GetLatencyForURL(relay)
-		if err != nil {
-			latency = time.Hour
-		}
+		go func(irelay string) {
+			latency, err := osutil.GetLatencyForURL(irelay)
```
calmh: I remember saying this somewhere at some point but maybe in another PR: I don't think the latency measurement will be correct or relevant when we do hundreds of connects in parallel. Maybe if we are limited to a random subset or relays, or do them in batches, or just take it easy and do it sequentially like now.

lib/dialer/public.go
```diff
@@ -89,11 +52,50 @@ func SetTrafficClass(conn net.Conn, class int) error {
...
-	case dialerConn:
-		return SetTrafficClass(conn.Conn, class)
-
 	default:
 		return fmt.Errorf("unknown connection type %T", conn)
 	}
 }
+
+func dialContextWithFallback(ctx context.Context, fallback proxy.ContextDialer, network, addr string) (net.Conn, error) {
+	dialer := proxy.FromEnvironment().(proxy.ContextDialer)
```
calmh: Given these functions can return an error I'd prefer if we were paranoid about the type assertions and returned an error in case of failure instead of panicking. Also goes for the one to *net.TCPAddr further below in the file.

lib/dialer/debug.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"os"
+	"strings"
+
+	"github.com/syncthing/syncthing/lib/logger"
+)
+
+var (
+	l = logger.DefaultLogger.NewFacility("dialer", "Dialing connections")
+	// To run before init() of other files that log on init.
```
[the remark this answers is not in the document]
calmh: Yeah this is really ugly and a huge red flag that we need to fix logging...

lib/dialer/public.go
```diff
@@ -89,11 +55,57 @@ func SetTrafficClass(conn net.Conn, class int) error {
...
+// fails. If the context has a timeout, the timeout might be applied twice.
+func DialContextReusePort(ctx context.Context, network, addr string) (net.Conn, error) {
+	dialer := &net.Dialer{
+		Control: ReusePortControl,
+	}
+	localAddrInterface := registry.Get(network, tcpAddrLess)
+	if localAddrInterface != nil {
+		if addr, ok := localAddrInterface.(*net.TCPAddr); !ok {
+			return nil, errUnexpectedInterfaceType
+		} else {
```
golangcibot: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary) (from `golint`)

AudriusButkevicius: You are fucking stupid.

imsodin:
The old go version check (go1.10.7) fails due to `ListenConfig` and `Dialer.Control` that have been introduced in go1.11: https://golang.org/doc/go1.11#net. go1.13 will be out soon and by using modules we already have a kind of "soft requirement" for go1.11, so I think we should be fine bumping the old go version check to go1.11.

lib/dialer/public.go
calmh: Hey, it's right although you may not agree with the style choice

AudriusButkevicius: I struggle to understand where the suggested style choice would be better.

AudriusButkevicius:
Addressed comments, take it for a spin.

AudriusButkevicius:
good to go?

AudriusButkevicius:
bhump

lib/dialer/control_unix.go
```diff
@@ -0,0 +1,51 @@
+// Copyright (C) 2019 The Syncthing Authors.
```
golangcibot: /usr/local/go/src/bytes/reader.go:98:18: DecodeRune not declared by package utf8 (from `typecheck`)

calmh:
giphy

calmh:
So... The reuse by itself seems fine, as in I get a connection between the two listening ports when there's nothing untoward between them. But I'm not sure how this works in the presence of NAT and absense of something STUN-like to inform us about the external TCP port?

I mean, given A -> NAT -> B (A behind NAT, B on public IP for simplicity) and both using default ports we'll get a connection attempt A -> B looking something like `A-internal-IP:22000 -> B-public-IP:22000` which at the NAT gateway becomes `A-public-IP:52389 -> B-public-IP:22000`. Assuming this connection doesn't work (stateful firewall on B) and we need the punchthrough, the connection in the other direction will be `B-public-IP:22000 -> A-public-IP:22000` which doesn't open the door for the incoming connection from A, nor will it reach A as it's not the reverse of the NAT:ed connection.

For this thing to work we'd need to know the outside port number that A gets and be able to announce that?

AudriusButkevicius:
That is why discovery uses the same connection to report it's IP, where the NAT port gets captured.

lib/util/utils_test.go
```diff
@@ -226,6 +226,51 @@ func TestCopyMatching(t *testing.T) {
...
+		netB  string
+		addrB string
+	}{
+		// B is assumed the winner.
+		{"tcp", "127.0.0.1:1234", "tcp", ":1235"},
+		{"tcp", "127.0.0.1:1234", "tcp", "0.0.0.0:1235"},
+		{"tcp4", "0.0.0.0:1234", "tcp", "0.0.0.0:1235"}, // tcp4 on the first one
+	}
+
+	for i, testCase := range cases {
```
golangcibot: unnecessary leading newline (from `whitespace`)

calmh:
I've deployed new discovery servers and run this build locally, but I'm not seeing fully appropriate magic.

```
jb@kvin:~ % stfinddevice SYNO4VL-...
discovery-v6.syncthing.net (91.566µs):
  lookups not supported
discovery-v4.syncthing.net (182.793µs):
  lookups not supported
discovery.syncthing.net (1.113722845s):
  address: quic://83.233.124.224:22000
  address: tcp://83.233.124.224:0
  address: tcp://83.233.124.224:22001
discovery.syncthing.net (1.129136223s):
  address: quic://83.233.124.224:22000
  address: tcp://83.233.124.224:0
  address: tcp://83.233.124.224:22001
```

There is the `:0`, but not a mapped port. The firewall shows it getting mapped to something ephemeral:




Can you get this part to work?

AudriusButkevicius:
It should have not recorded 0 as the discovered port.
I don't think I have a way to test this either, this is all hypothetical.

st-review:
:robot: *beep boop*

AudriusButkevicius:
I'll get to this, some day.

AudriusButkevicius:
So the announce part seems to work for me.
Added req.RemoteAddr to the typical log:
```
380704bb7b4d7c03 POST / 86.29.10.111:22000
```
And log the announcement addresses before and after the "fixup"
```
Announce: 2ERMYXS-ZKTQICW-AWCYDBJ-ZDSV4JA-5PYZHDV-6Q4COSC-3POAZJO-UDAYFAS got: [tcp://0.0.0.0:0 tcp://0.0.0.0:22000 quic://0.0.0.0:22000] made: [tcp://86.29.10.111:22000 tcp://86.29.10.111:22000 quic://86.29.10.111:22000] 
```

I suspect the problem you've seen is because nginx/caddy fronting the proxy does not forward X-Forwarded-Port.

AudriusButkevicius:
How do we take this forward?

calmh:
Some kind of reasonable testing I guess. Have you run this and seen it work? If so I'll take it for another spin and we can debug any issues. Last time I had the feeling it was all theoretical and hence I wasn't sure whether to debug my network or just assume that things didn't work as intended...

AudriusButkevicius:
Well the only thing I tested is that the source address connecting to discovery is port 22000, and that discovery correctly substitutes port 0 with that port (on a local discovery server). I have not tested actual punch through capabilities as I don't have two devices behind different domestic routers.

AudriusButkevicius:
I think the best way forward with this would be to cherry-pick and deploy the discovery changes and adding support for X-Forwarded-Port.

This should have no effect.


Then run with this PR for a week to verify it doesn't break anything existing in your setup, and then just merge it on good will and see if anything comes from the RC?

AudriusButkevicius:
@calmh @imsodin bhuump. What are we doing with this?

cmd/stdiscosrv/apisrv.go
```diff
@@ -384,27 +391,34 @@ func fixupAddresses(remote net.IP, addresses []string) []string {
 			continue
 		}
 
-		if host == "" || ip.IsUnspecified() {
-			// Replace the unspecified IP with the request source.
+		if remote != nil {
+			if host == "" || ip.IsUnspecified() {
```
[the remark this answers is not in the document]
AudriusButkevicius: There is a block after the nested if which would have to repeat `remote != nil` then.

AudriusButkevicius:
Soo, approve and elts go? :D

imsodin:
Only @calmh can deploy, so the ball is in his court now, isn't it?  
I might also just run a discovery server at home for testing, I just need to overcome the laziness to set the ports and clients up for that...

calmh:
I actually did deploy the discovery servers from this a while back (still running), but I can do so again. I'm generally not super excited about merging code that the person who wrote it hasn't been able to test at all for whether it actually does what it's supposed to, but if there's confidence that it doesn't sabotage existing connection setup and a plan for actual post-merge testing and reporting back, sure.

calmh:
So looking at this it seems we want `X-Forwarded-Port` to be the source port of the client, while it's generally meant to be the (original) server port? I'll set it up with the client port

calmh:
Well some cursory testing at home seems to indicate it doesn't in fact fuck anything up beyond repair, so that's good 👍

AudriusButkevicius:
You are right, I should probably change it to X-Client-Port or something.

I'll also try to come with with something that atleast tries to show it working.

calmh:
I would try it with one end tethered behind mobile, assuming the other is behind a "regular" router. But this still requires full cone (I think? or maybe just deterministic port mappings) so might not work in that setup even if everything is great on our side.

calmh:
Anyway it's deployed server side, though we can redeploy with a different header name later

PR #5756, opened 2019-05-30
gui: Split device list in folder sharing options by shared / unshared

acolomb:
### Purpose

This is part of (and a prerequisite for) the idea outlined in [this forum discussion](https://forum.syncthing.net/t/manual-intervention-for-introducer-feature/12212) to list suggested devices that could improve cluster connectivity.  So far, this PR only separates the "already shared with" (declared as category 1 in the discussion) devices from the unshared (category 3) devices in two sections, each of which is hidden when empty.  As a side-effect, that makes it easier to see what changes have been made in the checkmarks before saving.

### Testing

Folder editing has been verified for a cluster of two remote devices, every combination of shared or unshared.  Adding a new folder looks as expected, with only the second section visible.  Adding a pending folder (sent from a peer) does so as well, with the inviting device checked.

### Screenshots

Screenshot_2019-05-30 devtest Syncthing

### Documentation

TODO!  Should check for screenshots and texts in Usage guide.


uok:
According to the [usage stats](https://data.syncthing.net/) the average user has 3 devices. In the sharing settings a device can either be shared (= checked box) or not shared (= unchecked box) so I in my opinion splitting the 2-3 entries into two sections is not helpful, even confusing for usability.

acolomb:
@uok Thanks for your consideration.  As mentioned above, this is just a stop-gap for having an additional category of "suggested" devices (see #5758).

In fact, these categories look quite crowded if each has only one or two entries.  We could maybe collapse them if the number of entries is small.  On the other hand, the "average" cluster with only three devices will probably have only one section because everything is shared with everyone.
> According to the [usage stats](https://data.syncthing.net/) the average user has 3 devices. In the sharing settings a device can either be shared (= checked box) or not shared (= unchecked box) so I in my opinion splitting the 2-3 entries into two sections is not helpful, even confusing for usability.

calmh:
I think the change is good technically, but I’m confused by the wording. Why add "initially"? And I think "unrelated" (and by extension *related*) is not a term we’ve used before - unshared devices or non-shared devices, perhaps?

acolomb:
@calmh the wording is owed to the [proposal on the forum](https://forum.syncthing.net/t/manual-intervention-for-introducer-feature/12212/5?u=acolomb) (prototype in #5758), where a third category is introduced.  In contrast to the devices listed there under "suggested", these "unrelated" devices do not yet have access to the folder as far as we know, not even indirectly.  Those two words could still be discussed, but for #5758, some terminology is needed for the distinction.  Sorry for not keeping this one clearly self-contained.  I expected to finish the other PR quicker.

For this PR alone, I agree that the wording is a little confusing since there is no distinction to be made.  Does this stand a chance to get merged independently?  If so, I'll gladly change the wording if we have a better suggestion.  How about this:

> Share With Other Devices
> -----
> Select other devices to share this folder with.

imsodin:
What about Jakobs proposal "unshared"? Then the potential third category could be "indirectly shared" with an explanation like "Select to directly connect and share with.". To me "other" feels like a filler word without much meaning.

acolomb:
@imsodin I like that suggestion, keeping the heading a "current state" description, which sets it apart from the explanation below concerning what the user can do with the checkboxes.

I also replaced the word "other" by "additional", to make it clear that this will extend the circle of devices with sharing access.  New preview:

> Unshared Devices
> -----
> Select additional devices to share this folder with.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1712,25 +1724,30 @@ angular.module('syncthing.core')
             $scope.editFolderModal();
         };
 
-        $scope.selectAllDevices = function () {
-            var devices = $scope.otherDevices();
+        $scope.selectAllSharedDevices = function (state = true) {
```
calmh: I think default parameter values are one of the ES6 features we haven't adopted yet for compatibility reasons?

AudriusButkevicius:
What's happening with this? It seems it needs a small change to land?

acolomb:
I don't know much about ECMAScript, so I was waiting for the maintainers' call on whether using default values would be okay.

calmh:
Please take my comment as confirmation that it isn't.

acolomb:
I switched to just making the argument mandatory and providing true / false whenever the functions are called.  This way we still only need a single function to select or deselect, avoiding boilerplate code duplication.

Could do the same for the functions in the "edit device" sharing dialog. But that's not necessarily material for this PR.

AudriusButkevicius:
So @calmh kind of approved this, but I find this to be worse UX. 

It's super confusing that you have two sections that do the same thing... Yet they are separate, giving some illusion that they are doing something else with these devices.

Am I missing some long-game idea that will get this to all make sense?

acolomb:
The long-term idea is #5758, where a third category will be added for "suggested" devices that are known to have the same folder through others' ClusterConfig messages, but not yet added locally.  I'll try to work on that again the next four weeks, as time and two kids permit.

I'm very much open to other suggestions for the layout / interaction. Somehow being able to see what the current status is and what changes are about to be applied is a Good Thing, though IMHO. Kind of like diffing in Git before committing :-) Of course that only makes sense for little more complex clusters.

AudriusButkevicius:
Is that true?

https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L2176

We only advertise folders that the other side is supposed to see, for privacy reasons, and we should not send information about other folders we are not sharing with a device.

Back to the change, for small clusters (2-3 devices), this change adds tons of confusion. 

The state of a folder is binary, it's either shared, or not, I don't see any value into breaking this into 4 states, as if anything it adds confusion and takes up screen space.

imsodin:
He is going after the information added a bit further down: https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L2194 That includes all devices we share the folder with, regardless of whether the device we are sending the CM to already shares the folder with these devices or not.

I am generally fine with this change: It only uses vertical space, which is cheap. And to me it doesn't introduce another category, it displays the current state: If I change stuff it's still clear what the previous state was. Whether it adds more confusion than clarity I am really not suitable to judge due to too much exposure to Syncthing.

acolomb:
@AudriusButkevicius #5758 is actually a working prototype, so feel free to try it out and see what info is available for your cluster, might be useful.  The main thing holding it back is that the info is stored rather inefficiently within the XML config.  Need to port that over to using the database.

acolomb:
As for the possible confusion: Initially, when opening the dialog, all items in the upper category are checked, and none in the lower section.  So until the user starts to change things, there are only two states, grouped into two categories.

AudriusButkevicius:
Just an FYI, I am neutral about this change, so if others are happy, let's get it merged.

acolomb:
I certainly wouldn't mind ;-)

acolomb:
Conflict is caused by #6073. I'm working on merging the general idea from that change into this PR.

acolomb:
I hope that merging current master into this PR branch was the right way to go about it.  If you like I can do a rebase and force-push instead.

It will be squashed before merging anyway I suppose?  I do like to keep all commit messages in the history for documentation purposes, but that's up to you maintainers :-)

uok:
I still think this is adding confusion to the UX. How often do you really edit sharing? And why is a simple list with "checkmark - shared" and "no checkmark - not shared" not enough? 

The split view does not really help the user I think. Also as I mentioned earlier the majority of users only has a handful of devices (like in your screenshot). Why not just add the unshared/unrelated devices to the list with simple checkboxes instead splitting them?

acolomb:
I don't think that distinction would be worth the trouble. Both maintenance and support wise.

Please again notice that the average Joe user with thre

acolomb:
I don't think that distinction would be worth the trouble. Both maintenance and support wise.

Please again notice that any empty category is automatically hidden. So suppose the average Joe user has only three devices. They will probably share folders with all of them (my guess, don't know if the statistics can verify this claim). Once that state is reached, it will look exactly like it does now, one category with all boxes checked. Seeing two categories shows the user Joe that he "forgot" one checkmark in order to reach the supposed goal of sharing with all devices. 

For power users who deliberately share a folder to only a subset of many devices, this makes it easier to distinguish. Just today I was in that situation again, with private and work related folders being shared to different subsets obviously and the list becoming not easy to grasp at first sight.

acolomb:
Ping.

Let's get this merged at last before the next round of inevitable merge conflicts come up.

Thanks.

acolomb:
Thank you @imsodin.

PR #5757, opened 2019-05-30
lib/model: Refactor sending indexes as suture service

imsodin:
This PR makes sure sending indexes is gracefully terminated when the model is stopped: Currently the goroutine is just killed on `os.Exit` if polling, which shouldn't be problematic. However when introducing more state like the disk overflow stuff, making sure we properly shut down and clean up behind ourselves becomes important.  
Edit: This might not have any relevance to the disk overflow stuff at all, but I still think it's a good thing to do.

And it removes quite a few function arguments, which I think is nice too.


AudriusButkevicius:
We should probably move it to a separate file.

Also, just a thought, but it seems we're repeating the same service pattern again and again.
Perhaps we should have somekind of `util.AsSutureService(func(stop <-chan struct{})) suture.Service` thing, and then migrate everything to that.

AudriusButkevicius:
Actually that's a bad idea. What I am after is essentially an abstract class, which can have it's own member variables, but need to implement `run(chan)` thing.

I don't think you can have that in Go, so I guess repetition it is then.

AudriusButkevicius:
https://medium.com/@adrianwit/abstract-class-reinvented-with-go-4a7326525034
but god that's horrible, you have to have a new for everything.

lib/model/model.go
```diff
@@ -1912,12 +1939,18 @@ func sendIndexes(conn protocol.Connection, folder string, fs *db.FileSet, prevSe
 		// currently in the database, wait for the local index to update. The
 		// local index may update for other folders than the one we are
 		// sending for.
-		if fs.Sequence(protocol.LocalDeviceID) <= prevSequence {
-			sub.Poll(time.Minute)
+		if s.fset.Sequence(protocol.LocalDeviceID) <= s.prevSequence {
+			select {
+			case <-s.stop:
+				return
+			case <-evChan:
```
calmh: This'll burn a bit more CPU as you'll now loop and create a one minute timer on every event, eventually having potentially hundreds of thousands of timers running in the background. I don't think we need the timer any more either given there's the select on the `stop` in parallel. If we do, a `time.Ticker` declared once outside the loop should do.

imsodin:
I assume the timeout on sending indexes is there in case of missed events.

AudriusButkevicius:
I prefer this option to the other one.

AudriusButkevicius:
I preferred storing the tokens :( why do we not want it?

imsodin:
> My comment about the timer requires resolving, and I'd prefer an even more explicit comment on Complete() 

Done. Is the comment explicit?

> We'll also accumulate tokens for every connection until the folder is stopped...

I removed storing the token and added the connection closed channel to the index sender. It's not really crucial that the index sender terminates before the folder is finished being removed.

imsodin:
> I preferred storing the tokens :( why do we not want it?

lib/model/model.go
```diff
@@ -1964,8 +1969,10 @@ func (s *indexSender) Stop() {
 	<-s.stopped
 }
 
-// Complete always returns true, as Serve only terminates when a connection is
-// closed/has failed, in which case retrying doesn't help.
+// Complete returning true means that the suture supervisor will not restart
+// this service if it wasn't stopped. Here it always returns true, as indexSender
+// only terminates when a connection is closed/has failed, in which case
+// retrying doesn't help.
```
AudriusButkevicius: We should explain that there is a hidden interface satisfaction check in suture for this.

The other option would be for the folder to be supervisor where all of its cruft gets run (versiomer, index senders et al), so we tear down the whole supervisor.

Also, it's a great opportunity to move this thing into a separate file.

calmh:
Right, so my comment about the tokens was more about the fact that we'll end up with x amount of "services" that are still managed by suture but not running (because completed) and with us keeping track of the token for a long time. That seems suboptimal and was one reason why I preferred not involving suture, really. But not tracking the token at all means we leak the service (I think?!) so is probably not an option.

imsodin:
> But not tracking the token at all means we leak the service (I think?!) so is probably not an option.

I don't think so: It's still internally registered with model as supervisor, so if that gets stopped so does the index sender service. And if the connection get's closed the index sender stops and due to `Complete` it doesn't get restarted but removed by suture: https://godoc.org/github.com/thejerf/suture#IsCompletable

I'll update the comment.

calmh:
Oh, right, *is removed* is the key here. Fine.

calmh:
Maybe add a short note about that besides `Add()` as well, as it's a legitimate question and none of us will remember this in six months

AudriusButkevicius:
Y no separate file?

AudriusButkevicius:
The sender is part of connection is a bit moot no? As its a sender per folder per connection?

imsodin:
> The sender is part of connection is a bit moot no? As its a sender per folder per connection?

No and yes - a folder can be running for a long time, while connections come and go. Therefore we would accumulate tokens as Jakob mentioned before.

> Y no separate file?

Can do, just seems like a mightily small file. Also personally I'd rather separate code changes from moving code without changes when possible.

calmh:
Yeah, diff makes sense like this. If you want to separate it, a separate commit for that would be good.

AudriusButkevicius:
Any line we can move out of that monster file should be moved out.

imsodin:
I get that sentiment. To that effect we could move a whole bunch of non-model types/functions (the map..., fileinfobatches, ...) to e.g. utils.go -> other PR.

PR #5758, opened 2019-05-30
[WIP] Collect and suggest devices for missing cluster links

acolomb:
### Purpose

This is an early prototype of the "suggested devices" feature discussed [on the forum](https://forum.syncthing.net/t/manual-intervention-for-introducer-feature/12212).  The PR includes:

* The backend changes for collecting info about device candidates to suggest, from the ClusterConfig messages.  Currently stored inside the config, because it is easy to experiment and debug.  This needs to be moved to the database though, since on a large cluster this dumps a lot of data to the XML and it touches the config for every received ClusterConfig message with additional candidates.  The general structure of collected data could be reused, though.

* PR #5756 as a prerequisite for splitting the GUI's folder sharing device list in groups.

* Another section (hidden if empty) in the GUI's Edit Folder dialog - Sharing tab.  It lists both known devices that have only an indirect link to this folder (category 2 from the forum discussion), as well as completely unknown devices (category 4) that would have been added if the Introducer feature was used.  The latter are included as clickable device IDs, with an appended alias hint *only if all introducers agree on the name*.

### Known Issues ###

* Actually adding unknown devices does not work correctly.  The folder and device editing dialogs share a `editingExisting` flag, which obviously needs to be separate when launching one modal dialog from the other.  Pre-filling some form fields is also non-functional, see FIXMEs in the code.

* The new section is currently only used when editing a folder, not when adding a pending folder shared from another device.  In theory, there could already be candidate devices, so showing them in that incarnation of the dialog would make sense.

* This should not be released to the userbase at large because of the inefficient storage of candidate devices in the configuration.  Large clusters will experience horrible performance!

* Noticed while testing: The ClusterConfig message currently includes any remote device (B) which the sender (A) "shares" the folder with.  This should possibly be limited to "mutually shared" links, where B also shares with A.  Once the device suggestions are merged, this extra data could make it easier to trick people into sharing data with the wrong device maybe?

* Collected candidate devices are not being cleaned up yet.  Need to decide on a strategy for keeping these records clean.

* Some smaller issues and questions marked `FIXME` in the code. RFC.

### Testing

All devices are listed under the expected category in a cluster with five devices.  Selecting the suggested devices works, except for unknown devices.  Verified the config entries both in the XML as well as over the JSON API.

### Screenshots

Screenshot_2019-05-31 devtest Syncthing

### Documentation

TODO: Once the data storage is agreed upon, any additions need to go to the API docs.  Also need a new section explaining the concept of "suggested devices" in the Usage guide.  Note that this is not a major addition to the inner workings of Syncthing, just exposing data that was already available.


lib/config/wrapper.go
```diff
@@ -549,3 +551,21 @@ func (w *wrapper) AddOrUpdatePendingFolder(id, label string, device protocol.Dev
 
 	panic("bug: adding pending folder for non-existing device")
 }
+
+func (w *wrapper) AddOrUpdateFolderCandidateDevice(folder string,
+	device, introducer protocol.DeviceID, name, certName string, addresses []string) {
+	defer w.Save()
```
golangcibot: Error return value of `w.Save` is not checked (from `errcheck`)

lib/config/observed.go
```diff
@@ -22,5 +22,55 @@ type ObservedDevice struct {
...
+	// Sort addresses into a map for deduplication
+	addressMap := make(map[string]struct{}, len(d.Addresses))
+	for _, s := range d.Addresses {
+		addressMap[s] = struct{}{}
+	}
+	for _, s := range addresses {
+		addressMap[s] = struct{}{}
+	}
+	d.Addresses = make([]string, 0, len(addressMap))
+	for a, _ := range addressMap {
```
golangcibot: should omit 2nd value from range; this loop is equivalent to `for a := range ...` (from `golint`)

lib/config/wrapper.go
acolomb: Copied verbatim from `AddOrUpdatePendingDevice()` and others.  Sorry I'm new to Go and have no idea how that should be checked in conjunction with `defer`.

This will not live in the config anyway, long-term.

lib/config/folderconfiguration.go
```diff
@@ -284,3 +288,33 @@ func (f *FolderConfiguration) CheckAvailableSpace(req int64) error {
...
 }
+
+func (f *FolderConfiguration) AddOrUpdateCandidateDevice(device, introducer protocol.DeviceID, name, certName string, addresses []string) {
+
+	//FIXME locking already handled in config wrapper?!?
+
+	for i := range f.CandidateDevices {
+		cDev := &f.CandidateDevices[i]
+		if cDev.ID == device {
+			if cDev.CertName != certName {
```
golangcibot: SA9003: empty branch (from `staticcheck`)

imsodin:
What's the status of this PR?

acolomb:
I haven't found much time to work on this. Two kids keeping me busy ;-)

Plus the unexpected backlash on #5756 (for such a small change) did not exactly boost my motivation. I'd like to have some reassurance that people will actually welcome this addition and I'm heading in the right direction. Given that the idea has been up for discussion [on the forum](https://forum.syncthing.net/t/manual-intervention-for-introducer-feature/12212) for over a year now.

Technically only the collected data needs to be stored in the database instead of the config. Lack of experience with schema-less databases being the main hurdle there. Plus some cleanup code to dump outdated entries.

AudriusButkevicius:
If you are not motivated, then we should revert #5756, as without the epiphany of the idea, I think it's actually a detrimental change for small set of devices.

imsodin:
I can help with the db if needed, that shouldn't be a showstopper.

acolomb:
@AudriusButkevicius seeing that one finally merged today was actually a very positive sign ;-)

st-review:
:robot: *beep boop*

acolomb:
Just pushed two commits to keep this prototype work up-to-date with master.  Just in case anyone wants to try it out.

I'm still trying to get my head around how to move this stuff from config to the database, but time is very limited.  Kindly asking for patience.  Once there are some meaningful commits toward a better implementation, I will open a new draft PR.

acolomb:
> I can help with the db if needed, that shouldn't be a showstopper.

Thanks for the offer, @imsodin. I hope it still stands?

I've looked at the DB storage code now and then, and finally managed to make it build with my custom protobuf messages to hold the information about candidate devices. Defined some additional types in `lib/db/structs.proto`, hitting a show-stopping error about my `google/protobuf/timestamp.proto` usage several times. But that seems to work now. Added new `NamespacedKV` types for pending and candidate devices.

So where to go from here?

1. I would need an API endpoint to retrieve the data within the GUI. `GET /rest/db/candidates` was my [suggestion](https://forum.syncthing.net/t/manual-intervention-for-introducer-feature/12212/39?u=acolomb), and I think it would be implemented in `lib/api/api.go`. Looking at that, there are also (undocumented) `GET /rest/folder/*` endpoints, so `GET /rest/folder/candidates` may be better?

2. Similar requests seem to be delegated to the model, which actually retrieves the data from the DB. I'm looking at the `folderFiles` and `deviceStatRefs` members for reference on how to access it.

   * So would a new method in `model.go` be the right place to implement this?

   * Should I extend the model by some mapping member akin to the two mentioned? Or would a new function suffice that directly searches the database?

   * What type would that function return? Currently I'm reusing (and extending) the types from `config/observed.go`. When candidate / pending devices info gets moved out of the config, where would be the right place to define such types?

3. Writing the info to the database seems obvious, in the model's `ClusterConfig` function. I'll just replace the config entry writing code with writing to the database.

4. On the DB layer, does this functionality (together with moving the PendingDevice / PendingFolder out of the config) warrant a new source file, like is done with `lib/db/set.go` and `lib/stats/*.go`? Or would the model just stupidly read and write entries in the `NamespacedKV` object directly? If the former is preferred, where should this all live?

5. GUI is a totally different beast, so I'll focus on wiring up the REST API first.

Please excuse if these questions seem trivial. I find it hard to wrap my head around the overall architecture with all these layers involved and could use some mentoring.

Just in case it makes a difference to the answers: My plan is to restructure / move the pending devices / folders stuff first "for practice", then add the candidate devices on top of that, since they are similar in nature IMHO. The former goal is based on @AudriusButkevicius' [comment](https://forum.syncthing.net/t/manual-intervention-for-introducer-feature/12212/27?u=acolomb) that storing the pending stuff in the config instead of the DB was a mistake in the first place.

imsodin:
> Thanks for the offer, @imsodin. I hope it still stands?

Sure does.

Can you push your experimental changes somewhere, regardless of how unfinished/broken it is. It makes understanding of what you do and giving useful answers easier.

1. `/rest/folder` is about local shares, that are already preasent. This is about meta information on other devices/connections. In general I am not too fond of `/rest/db`, I don't think db is something meaningful externally (unless it's really about db stats). I'd probably put it in to `/rest/system`, some endpoints there seem related the most.

2.  
 - Definitely model. Everything goes to (for better or rather worse).  
 - I don't think members are needed. If it isn't looked up often (I don't think it is) and the logic is sufficiently small, just a function in model retrieving the info from db seems like the way to go. If it needs e.g. caching in memory or gets large/messy, I'd put that in the db if it makes sense (related to 4.).
 - The type will need defining in api (or db if a more complex implementation is needed and done there). The api cannot import model, so defining on the model just means you'll need to redefine on the api. And the api will be the only consumer anyway I think.

4. Again, depends on the complexity of the entire thing. If it becomes largish or complex, then it's definitely nice if you can separate it from the model (so that in model you just need to call functions). If that's needed a separate source file is likely a good idea, but again that depends on what it looks like. I'd expect it to land in the db package.

The plan of first moving the pending stuff from config to db sounds very reasonable. And don't apologize for asking questions, and definitely for them being "trivial" - anything gets trivial if you stare long enough at it, and so by corollary nothing is initially trivial (I fondly remember one math prof that was honest enough during a proof to state "And now, if we stare long enough at this line, the next step is evident: ...").

acolomb:
Thanks for your taking time to answer thoroughly.

> Can you push your experimental changes somewhere, regardless of how unfinished/broken it is. It makes understanding of what you do and giving useful answers easier.

Will do when I had the next time slice available for staring and pushing to a WIP branch on my fork.

>  1. `/rest/folder` is about local shares, that are already preasent. This is about meta information on other devices/connections. In general I am not too fond of `/rest/db`, I don't think db is something meaningful externally (unless it's really about db stats). I'd probably put it in to `/rest/system`, some endpoints there seem related the most.

The candidate devices are definitely information about already shared folders, in the sense of "whom else you might want to share this with". For pending devices you are right and I'm considering `/rest/system/connections/pending` or just adding a sibling element (pending) to the `/rest/system/connections` JSON response.

For pending folders, it is not clear at all, though I agree that `/rest/db/` is not the right place. Either we create a totally new endpoint e.g. `/rest/system/pendingfolders` or it could actually be integrated in the `/rest/system/connections` response as well, for the origin device that is offering the folder. That would be harder to aggregate into a list though for API users and also requires some rather inefficient database accesses, as they would be keyed by folder ID.

>   4. Again, depends on the complexity of the entire thing. If it becomes largish or complex, then it's definitely nice if you can separate it from the model (so that in model you just need to call functions). If that's needed a separate source file is likely a good idea, but again that depends on what it looks like. I'd expect it to land in the db package.

So my gut feeling says the helper functions will live in `lib/db/observed.go`, which is basically a translation of my current extensions (in this PR) to `lib/config/observed.go` but using the new protobuf types. Obviously the model does not import the API package, so the API types should be split off like was done for `lib/stats`, which the model does import.

The information of interest will be structured already in the DB, so it seems odd to mirror it with another API structure type. The `jsonFileInfoTrunc` / `db.FileInfoTruncated` construct seems to do a similar protobuf -> JSON translation so that may be a way to go as well. Not quite as nice as the automated XML / Go / JSON translations done in `lib/config` though.

The next steps are becoming more an more evident, your prof was right ;-)

acolomb:
One more generic question: I don't see what benefit the `NamespacedKV` construct brings to this use case. It eases handling of database keys of the form:

    <KeyType byte>             +  "string"  +  "field name"
    # for example:
    4(KeyTypeFolderStatistic)  +  folderID  +  "lastFileName"

If we want to store a protobuf message structure as [suggested](https://forum.syncthing.net/t/manual-intervention-for-introducer-feature/12212/36?u=acolomb) by @calmh, the last part (field names) is already defined through that mechanism and the key is just:

    <KeyType byte>             +  "string"
    14(KeyTypePendingFolder)   +  folderID

So would it still be a good idea to go through `NamespacedKV`? My data (DB values) will be structured deeper than the folder stats for example, where there are just three different fields of primitive type.

I'm leaning toward just implementing the DB key generation myself, independent of the `NamespacedKV` and stick with protobuf as value structure.

imsodin:
`NamespacedKV` is a utility wrapper that handles key generation, such that it's usable outside of `lib/db` (keys aren't exposed). You'd have to extend it to have a `Put` method to directly store `[]byte`, which isn't a big deal. In the end you probably don't need to use it, as you need a type on which you implement some kind of `AddPendingFolder`/`RemovePendingFolder`/... methods. So I'd  expect some new type like `db.ConnectionInfo` (not necessarily a good name) that can then use the db and keyer to store/retrieve it's data in a prefix.

acolomb:
A first draft of the REST API for the pending folders and devices lives in my fork of the docs repository:

https://github.com/acolomb/syncthing-docs/tree/rest-api-for-pending-stuff

Thought it would be a good place since that will need documentation anyway. Should I open a draft PR for the docs repo as a new place for the discussion or shall we keep it here?

Kindly asking for comments on the API design. Candidate devices (which this PR is actually about) will be modeled on top of that.

imsodin:
> Thought it would be a good place since that will need documentation anyway. Should I open a draft PR for the docs repo as a new place for the discussion or shall we keep it here?

Please do, so one can comment directly on the relevant sections.

acolomb:
Done at https://github.com/syncthing/docs/pull/498. Comments welcome.

acolomb:
> Can you push your experimental changes somewhere, regardless of how unfinished/broken it is. It makes understanding of what you do and giving useful answers easier.

I've started pushing some draft work to #6443 for only the pending folders / devices stuff.  Let's continue the related discussion there.

PR #5763, opened 2019-06-06
all: Remove "large blocks" config

calmh:
We now always use large / variable blocks.

This also needs a doc update, specific mention in the release notes and, potentially, a notice dialog on first startup? I personally think we *don't* need to do an alert for it, as it's not that big of a deal and it's not particularly actionable by the user anyway.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -78,7 +78,6 @@ angular.module('syncthing.core')
             externalCommand: "",
             autoNormalize: true,
             path: "",
-            useLargeBlocks: true,
         };
 
         $scope.localStateTotal = {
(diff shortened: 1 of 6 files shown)


imsodin:
> I personally think we don't need to do an alert for it, as it's not that big of a deal and it's not particularly actionable by the user anyway.

I agree, but I call it that there will be at least one report of a user with debian stable on their NAS and something recent on the laptop which will be bitten by the large block panic. I still agree because it wouldn't help that person much if the UI informs them after the fact (upgrading).

calmh:
Yes there will probably be a nonzero amount of pain, your scenario is likely. :) But it's been over a year since we released a version that could handle large blocks so I think it's legitimate.

calmh:
I wonder if it would be good to retain the flag in the json representation of the config, ignored on input and always `true` on output?

AudriusButkevicius:
Ping @canton7 @kozec @Catfriend1 for awareness, but otherwise yoloooo

Catfriend1:
Thanks will look into it;)

canton7:
I haven't been keeping on top of this discussion, but I don't *think* this affects me? Please tell me if I'm wrong!

Catfriend1:
If the option is completely gone in config.xml and json, the Android wrapper will need https://github.com/syncthing/syncthing-android/pull/1270/files to be reverted as soon as it's built upon the core release containing this change.

PR #5769, opened 2019-06-09
lib/versioner: Use mtime for version cleanup (fixes #5765)

### LINKED ISSUE #5765: Verioner cleanup looks at the wrong time
Should look at mtime of when the file was archived, not mtime of the file.
To be honest, we should probably swap the two to start with, for easier search by users.

AudriusButkevicius:


calmh:
I didn't read too closely yet, but this just fixes the bug without switching the times around, right? Do we want to do that or not? If so, I think now is the time...

lib/versioner/staggered_test.go
```diff
@@ -70,7 +70,19 @@ func TestStaggeredVersioningVersionCount(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		versionsWithMtime = append(versionsWithMtime, versionWithMtime{
+			name:  file,
+			mtime: versionTime,
+		})
+	}
+
+	rem := v.toRemove(versionsWithMtime, now)
```
calmh: This test should be rewritten. There is no need to run the versioner at all to just call toRemove, and it's now quite confusing that we extract the mtime from the time when we explicitly don't act like that on the info in the tag. (Sorry for pushing back on this; it's not your fault the test is what it is, but this makes it worse.)

AudriusButkevicius:
I see appeal in that as it's easy for users to find when files were versioned by doing a find, but I guess they can do a find based on mtime too.

I hacked the test and wasn't sure we'll settle on this version so didn't spend much time.
I'll fix the tests I guess.

AudriusButkevicius:
Please go through it once more, hopefully this lands tonight.

lib/versioner/simple.go
```diff
@@ -73,12 +72,14 @@ func (v Simple) Archive(filePath string) error {
 	// Use all the found filenames. "~" sorts after "." so all old pattern
 	// files will be deleted before any new, which is as it should be.
 	versions := util.UniqueTrimmedStrings(append(oldVersions, newVersions...))
-	sort.Strings(versions)
 
-	if len(versions) > v.keep {
-		for _, toRemove := range versions[:len(versions)-v.keep] {
+	// Versions with mtimes
```
calmh: Captain obvious :)

lib/versioner/simple.go
```diff
@@ -73,12 +72,14 @@ func (v Simple) Archive(filePath string) error {
...
+	versionsWithMtimes := versionsToVersionsWithMtime(v.versionsFs, versions)
+
+	if len(versionsWithMtimes) > v.keep {
+		for _, toRemove := range versionsWithMtimes[:len(versionsWithMtimes)-v.keep] {
```
calmh: The list of versions must be properly sorted here, which they are not (they are sorted by name or whatever glob happened to return rather than mtime). I think `versionsToVersionsWithMtime` should do this sorting so the caller doesn't need to remember.

AudriusButkevicius:
I won't be able to address this until late evening so please wrap it up.

calmh:
Will do so, hopefully without fucking it up 👍

calmh:
Haha, I fucked it up. Apparently it's not that simple. *looking closer*

AudriusButkevicius:
There are two call sites to expire, one doesn't sort hence why I pushed it down I think.

calmh:
Yah... This staggered thing... So I think we'll get to live with some double sorting for safety's sake.

PR #5776, opened 2019-06-10
lib/fs: Enhance mtimefs, use everywhere

AudriusButkevicius:
Use the wrapped filesystem in most places.
This fixes the two isolated test cases I've added, no idea if it propagates to the right thing in the puller and does not cause "modified but not rescanned" chain.

I am not sure where .save was called from before, as I think puller did not use this filesystem?

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -107,10 +107,10 @@ type sendReceiveFolder struct {
 	pullErrorsMut sync.Mutex
 }
 
-func newSendReceiveFolder(model *model, fset *db.FileSet, ignores *ignore.Matcher, cfg config.FolderConfiguration, ver versioner.Versioner, fs fs.Filesystem) service {
+func newSendReceiveFolder(model *model, fset *db.FileSet, ignores *ignore.Matcher, cfg config.FolderConfiguration, ver versioner.Versioner) service {
 	f := &sendReceiveFolder{
 		folder:        newFolder(model, fset, ignores, cfg),
-		fs:            fs,
+		fs:            fset.MtimeFS(),
 		versioner:     ver,
 		queue:         newJobQueue(),
 		pullErrorsMut: sync.NewMutex(),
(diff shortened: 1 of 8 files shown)


imsodin:
> I am not sure where .save was called from before, as I think puller did not use this filesystem?

The filesystem passed to the folder factories already was an mtimefs.

AudriusButkevicius:
https://github.com/syncthing/syncthing/pull/5776/files#diff-cf783533b0c2c4fe7c3a325925917a56R113
begs to disagree

AudriusButkevicius:
Nvm, we call mtimefs higher up, so you are right, this makes it less ambigious atleast.

lib/model/folder_sendrecv.go
```diff
@@ -107,10 +107,10 @@ type sendReceiveFolder struct {
 	pullErrorsMut sync.Mutex
 }
 
-func newSendReceiveFolder(model *model, fset *db.FileSet, ignores *ignore.Matcher, cfg config.FolderConfiguration, ver versioner.Versioner, fs fs.Filesystem) service {
+func newSendReceiveFolder(model *model, fset *db.FileSet, ignores *ignore.Matcher, cfg config.FolderConfiguration, ver versioner.Versioner) service {
 	f := &sendReceiveFolder{
 		folder:        newFolder(model, fset, ignores, cfg),
-		fs:            fs,
```
calmh: I don't think this change is an improvement. Apparently we don't do that today, but being able to create a folder instance pointing at a filesystem of your choice seems a good idea, while being able to get a filesystem from the database layer is more of an ugly hack due to the need for an mtimefs. This type & constructor doesn't need to know about that hack.

AudriusButkevicius: Sure, again, please undo the changes and merge if you want this to land in the RC as I won't have time.

calmh: Yeps!

AudriusButkevicius:
Can't approve

calmh:
:)

calmh:
Gonna make an issue for this to resolve, too

PR #5779, opened 2019-06-11
cmd/stcrashreceiver: Add source code loader

calmh:
One complication is that in our production builds the paths that hit the loader look like "/opt/tcagent/syncthing-1-work/174e136266f8a219/lib/db/lowlevel.go" which doesn't contain our package path nor a clear marker for where the "root" directory is. I opted instead to look for the /lib/ or /cmd/ component which is always present somewhere in our packages.

It appears to work:

Screen Shot 2019-06-11 at 11 22 07

### DIFF
--- a/cmd/stcrashreceiver/sentry.go
+++ b/cmd/stcrashreceiver/sentry.go
@@ -111,9 +111,10 @@ func parseReport(path string, report []byte) (*raven.Packet, error) {
 	}
 
 	pkt := &raven.Packet{
-		Message:  string(subjectLine),
-		Platform: "go",
-		Release:  version.tag,
+		Message:     string(subjectLine),
+		Platform:    "go",
+		Release:     version.tag,
+		Environment: version.environment(),
 		Tags: raven.Tags{
 			raven.Tag{Key: "version", Value: version.version},
 			raven.Tag{Key: "tag", Value: version.tag},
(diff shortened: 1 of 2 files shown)


cmd/stcrashreceiver/sentry.go
```diff
@@ -80,17 +86,25 @@ func parseReport(path string, report []byte) (*raven.Packet, error) {
 		return nil, err
 	}
 
+	// Lock the source code loader to the version we are processing here.
+	if version.commit != "" {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

calmh: 🙉

cmd/stcrashreceiver/sentry.go
```diff
@@ -80,30 +86,38 @@ func parseReport(path string, report []byte) (*raven.Packet, error) {
...
 		if gr.First {
 			trace.Frames = make([]*raven.StacktraceFrame, len(gr.Stack.Calls))
 			for i, sc := range gr.Stack.Calls {
-				trace.Frames[len(trace.Frames)-1-i] = &raven.StacktraceFrame{
-					Function: sc.Func.Name(),
-					Module:   sc.Func.PkgName(),
-					Filename: sc.SrcPath,
-					Lineno:   sc.Line,
-				}
+				trace.Frames[len(trace.Frames)-1-i] = raven.NewStacktraceFrame(0, sc.Func.Name(), sc.SrcPath, sc.Line, 3, nil)
```
AudriusButkevicius: Why just first? Should we send all and mark which one crashed like in my impl?

calmh: I guess we could. One of the things that turned me off sentry was how that turned the backtrace piece of the ui into an unreadable mess of identical dropdown choices, so I'm kinda cool with having just the culprit goroutine there and the rest in the raw.

AudriusButkevicius: It does preselect the culprit tho. We could run it against that collapser thingy and only send different ones?

PR #5782, opened 2019-06-11
cmd/syncthing, lib/db: Exit/close db faster (fixes #5781)

### LINKED ISSUE #5781: Shutdown takes too long, triggering fmut deadlock panic
When the Syncthing is shut down and thus the database closed, iterations may still be happening (e.g. receiving indexes). This is not safe according to the goleveldb documentation. And in general it isn't nice if Syncthing just doesn't shut down for a long while after the user asked it to do so. And this triggers the deadlock detector in RCs. 

Sentry Issue: [SYNCTHING-G](https://sentry.syncthing.net/syncthing/syncthing/issues/29/?referrer=github_integration)

imsodin:
This adds a 10s timeout on closing the db and additionally cancels active
db iterators and waits for them to terminate before closing the db.

@calmh I hope the amount of introduced complications passes below the bar - I am actually quite happy with it, but I suspect mine is also a bit high :)

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -39,6 +39,7 @@ type Lowlevel struct {
 	deviceIdx *smallIndex
 	closed    bool
 	closeMut  *sync.RWMutex
+	iterWG    sync.WaitGroup
 }
 
 // Open attempts to open the database at the given location, and runs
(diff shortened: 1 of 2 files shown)


lib/db/lowlevel.go
```diff
@@ -109,32 +110,45 @@ func (db *Lowlevel) Delete(key []byte, wo *opt.WriteOptions) error {
...
 
 func (db *Lowlevel) Close() {
-	db.closeMut.Lock()
-	defer db.closeMut.Unlock()
+	db.closeMut.RLock()
 	if db.closed {
+		db.closeMut.RUnlock()
 		return
 	}
 	db.closed = true
```
calmh: writing with rlock

lib/db/lowlevel.go
```diff
@@ -109,32 +110,45 @@ func (db *Lowlevel) Delete(key []byte, wo *opt.WriteOptions) error {
 }
 
 func (db *Lowlevel) NewIterator(slice *util.Range, ro *opt.ReadOptions) iterator.Iterator {
+	return db.newIterator(func() iterator.Iterator { return db.DB.NewIterator(slice, ro) })
+}
+
+func (db *Lowlevel) newIterator(fn func() iterator.Iterator) iterator.Iterator {
```
calmh: comment here on why it takes a constructor function (and maybe call it something like that) instead of just an iterator, as it took me a while to figure out what was going on here and why :)

imsodin:
> Do we know what happens, in all places, when an iterator stops iterating prematurely? It strikes me that this isn't guaranteed not to cause consistency issues (we were iterating to count something, iterator returns, we store the counter before db shuts down)...

Good point, I don't. I now added checks to `Put` and `Write` on `Lowlevel` which won't allow writes when `Lowlevel.closed` is true. Thus potential inconsistent values due to aborted iteration can't be written to database.

AudriusButkevicius:
Closed could be a chan removing the locking.

imsodin:
You'd still need the locking, as you'd have a race between closing closed -> closing db and checking closed -> doing something with the db.

AudriusButkevicius:
As per comment in the other thread, we should understand why it's slow rather than YOLO 10mins whatevs lets corrupt the db.

Perhaps it's flushing out the write buffers and disks are slow or something.

imsodin:
For context, did you notice https://sentry.syncthing.net/syncthing/syncthing/issues/29/activity/?query=is%3Aunresolved ?

imsodin:
Seems like we got to an agreement in https://github.com/syncthing/syncthing/issues/5781 - anything left to discuss? Maybe the time limit for closing.

AudriusButkevicius:
Lgtm, the close locks are sometimes held during the operation itself which is weird, but other than that, lgtm

imsodin:
AFAIK it may be weird, but it is necessary, to prevent a race as stated above: 

> You'd still need the locking, as you'd have a race between closing closed -> closing db and checking closed -> doing something with the db.

If you disagree and I got this wrong, then that's an important part of this PR and must be addressed. So could you please elaborate on the weirdness?

PR #5783, opened 2019-06-11
lib/model: Release pmut before logging event in Closed

imsodin:
This is due to https://sentry.syncthing.net/syncthing/syncthing/issues/26, where the routine holding pmut is executing this:

```
	events.Default.Log(events.DeviceDisconnected, map[string]string{
		"id":    device.String(),
		"error": err.Error(),
	})
```

While there's no (good) reason for this to block, there's also no reason to keep holding pmut for that.

I think this is tiny enough to be issueless :)

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1406,9 +1406,9 @@ func (m *model) Closed(conn protocol.Connection, err error) {
 	device := conn.ID()
 
 	m.pmut.Lock()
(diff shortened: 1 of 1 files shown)


calmh:
This one is tiny, but it's also just one of probably a thousand places where we generate events while holding a lock. I think we should hold off on fixing this until we get more data about whether it's necessary, and in that case make a bigger overhaul.

PR #5802, opened 2019-06-17
lib/relay: Initialise stop channel (fixes #5801)

### LINKED ISSUE #5801: panic: close of nil channel in relay client stop
1.2.0-rc.1

```

  File "C:/BuildAgent/work/174e136266f8a219/lib/relay/client/dynamic.go", line 124, in Stop
    close(c.stop)
  File "C:/BuildAgent/work/174e136266f8a219/lib/connections/relay_listen.go", line 124, in Stop
    t.client.Stop()
  File "C:/BuildAgent/work/pkg/mod/github.com/thejerf/suture@v3.0.2+incompatible/supervisor.go", line 651, in func1

panic: close of nil channel
```

https://crash.syncthing.net/report/f9/b0/f9b0fd4fbea87f1291d8559b2980ecbc98ae47e3b88b9c2320cc8865fa2f2c71

//@calmh

imsodin:
The stop channel is defined within `Serve`, so if `Stop` is called before that it closes a nil channel -> define at creation time. I left the (re-)definition in `Serve`, as despite the naming this isn't a suture service, but something custom, that may be restarted. I looked into cleaning this up, but it would be a major change so I abandoned that.

Testing: None.

### DIFF
--- a/lib/relay/client/dynamic.go
+++ b/lib/relay/client/dynamic.go
@@ -43,7 +43,8 @@ func newDynamicClient(uri *url.URL, certs []tls.Certificate, invitations chan pr
 		closeInvitationsOnFinish: closeInvitationsOnFinish,
 		timeout:                  timeout,
 
-		mut: sync.NewRWMutex(),
+		mut:  sync.NewRWMutex(),
+		stop: make(chan struct{}),
 	}
 }
 


calmh:
Sorry to be a neggo but since this just trades the crash bug for a silent bug, and the crash being rare, I think we can hold off and fix it properly at some point?

AudriusButkevicius:
Yeah this goes back to the "we need a template for suture services"

PR #5805, opened 2019-06-17
all: Refactor cmd/syncthing creating lib/syncthing (ref #4085)

imsodin:
### Purpose

Just in case anyone is bored ( :P ) this is a stab at a first step for refactoring cmd/syncthing. It does not tackle any of the hard problems (logging), it just adds an interface to start the "Syncthing application" and separates things done in cmd (env vars, loading config, ...) and in the application itself.

And this separation is one point of discussion, because it's pretty clear in some cases, but pretty arbitrary in other. One "underlying motivation" for decisions was, to limit interactions with filesystem or OS in general on the lib side and keep the interface simple. I marked items which I could just as well see in the other category with (?). A probably non-exhaustive list:

cmd:
 - Loading config from file
 - Parsing command line options and env vars.
 - Handling signals
 - (?) CPU profiling
 - Standby monitoring
 - Auto upgrades
 - Starting browser

lib:
 - (?) Maximizing open file limits
 - (?) Loading keys/certs  
FS interaction vs having a `tls.Certificate` in the interface
 - (?) Http profiler interface
 - (?) Http profiler interface
 - (?) Loading, updating, resetting delta indexes, ... for db  
Again FS interaction vs having a `Lowlevel` in the interface

The a bit unusual commit series is to make rebasing easier and diffs cleaner. The first three commits are mostly code shuffling, the latter commits changes to make things really work.

### Testing

It runs - haven't tested options/env vars etc, this is for later if/when this matures.

### DIFF
--- a/lib/syncthing/syncthing.go
+++ b/lib/syncthing/syncthing.go
@@ -1,0 +1,103 @@
+// Copyright (C) 2014 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package syncthing
+
+import (
+	"io"
+	"sync"
+
+	"github.com/syncthing/syncthing/lib/config"
+	"github.com/syncthing/syncthing/lib/db"
+	"github.com/syncthing/syncthing/lib/protocol"
+
+	"github.com/thejerf/suture"
+)
+
+const (
+	bepProtocolName      = "bep/1.0"
+	tlsDefaultCommonName = "syncthing"
+	maxSystemErrors      = 5
+	initialSystemLog     = 10
+	maxSystemLog         = 250
+)
+
+type ExitStatus int
+
+const (
+	ExitSuccess ExitStatus = 0
+	ExitError   ExitStatus = 1
+	ExitRestart ExitStatus = 3
+)
+
+type Options struct {
+	AssetDir         string
+	AuditWriter      io.Writer
+	DeadlockTimeoutS int
+	NoUpgrade        bool
+	ProfilerURL      string
+	ResetDeltaIdxs   bool
+	Verbose          bool
+}
+
+type App struct {
+	myID        protocol.DeviceID
+	mainService *suture.Supervisor
+	ll          *db.Lowlevel
+	opts        Options
+	cfg         config.Wrapper
+	exitStatus  ExitStatus
+	startOnce   sync.Once
+	stop        chan struct{}
+	stopped     chan struct{}
+}
+
+func New(cfg config.Wrapper, opts Options) *App {
+	return &App{
+		opts:    opts,
+		cfg:     cfg,
+		stop:    make(chan struct{}),
+		stopped: make(chan struct{}),
+	}
+}
+
+func (a *App) Run() ExitStatus {
+	a.Start()
+	return a.Wait()
+}
+
+func (a *App) Start() {
+	a.startOnce.Do(func() {
+		if err := a.startup(); err != nil {
+			close(a.stop)
+			a.exitStatus = ExitError
+			close(a.stopped)
+			return
+		}
+		go a.run()
+	})
+}
+
+func (a *App) Wait() ExitStatus {
+	<-a.stopped
+	return a.exitStatus
+}
+
+func (a *App) Stop(stopReason ExitStatus) ExitStatus {
+	select {
+	case <-a.stopped:
+	case <-a.stop:
+	default:
+		close(a.stop)
+	}
+	<-a.stopped
+	// If there was already an exit status set internally, ignore the reason
+	// given to Stop.
+	if a.exitStatus == ExitSuccess {
+		a.exitStatus = stopReason
+	}
+	return a.exitStatus
+}
--- a/lib/syncthing/syncthing.go
+++ b/lib/syncthing/syncthing.go
@@ -7,12 +7,33 @@
 package syncthing
 
 import (
+	"crypto/tls"
+	"fmt"
 	"io"
+	"net/http"
+	"os"
+	"runtime"
+	"strconv"
+	"strings"
 	"sync"
+	"time"
 
+	"github.com/syncthing/syncthing/lib/api"
+	"github.com/syncthing/syncthing/lib/build"
 	"github.com/syncthing/syncthing/lib/config"
+	"github.com/syncthing/syncthing/lib/connections"
 	"github.com/syncthing/syncthing/lib/db"
+	"github.com/syncthing/syncthing/lib/discover"
+	"github.com/syncthing/syncthing/lib/events"
+	"github.com/syncthing/syncthing/lib/locations"
+	"github.com/syncthing/syncthing/lib/logger"
+	"github.com/syncthing/syncthing/lib/model"
+	"github.com/syncthing/syncthing/lib/osutil"
 	"github.com/syncthing/syncthing/lib/protocol"
+	"github.com/syncthing/syncthing/lib/rand"
+	"github.com/syncthing/syncthing/lib/sha256"
+	"github.com/syncthing/syncthing/lib/tlsutil"
+	"github.com/syncthing/syncthing/lib/ur"
 
 	"github.com/thejerf/suture"
 )
@@ -81,9 +102,7 @@ func (a *App) Start() {
 	})
 }
 
-func syncthingMain(runtimeOptions RuntimeOptions) {
-	setupSignalHandling()
-
+func (a *App) startup() error {
 	// Create a main service manager. We'll add things to this as we go along.
 	// We want any logging it does to go through our log system.
 	mainService := suture.New("main", suture.Spec{
@@ -159,12 +178,6 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 		"home": locations.GetBaseDir(locations.ConfigBaseDir),
 		"myID": myID.String(),
 	})
-
-	cfg, err := loadConfigAtStartup(runtimeOptions.allowNewerConfig)
-	if err != nil {
-		l.Warnln("Failed to initialize config:", err)
-		os.Exit(exitError)
-	}
 
 	if err := checkShortIDs(cfg); err != nil {
 		l.Warnln("Short device IDs are in conflict. Unlucky!\n  Regenerate the device ID of one of the following:\n  ", err)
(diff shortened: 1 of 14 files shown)


lib/syncthing/syncthing.go
```diff
@@ -0,0 +1,505 @@
...
+			a.exitStatus = ExitError
+			a.err = err
+			close(a.stopped)
+			return
+		}
+		go a.run()
+	})
+}
+
+func (a *App) startup() error {
```
golangcibot: cyclomatic complexity 36 of func `(*App).startup` is high (> 30) (from `gocyclo`)

lib/syncthing/syncthing.go
```diff
@@ -0,0 +1,505 @@
...
+type ExitStatus int
+
+const (
+	ExitSuccess ExitStatus = 0
+	ExitError   ExitStatus = 1
+	ExitRestart ExitStatus = 3
+	ExitUpgrade ExitStatus = 4
+)
+
+type Options struct {
```
golangcibot: struct of size 72 bytes could be of size 64 bytes:
```
struct{
	AssetDir        	string,
	AuditWriter     	io.Writer,
	ProfilerURL     	string,
	DeadlockTimeoutS	int,
	NoUpgrade       	bool,
	ResetDeltaIdxs  	bool,
	Verbose         	bool,
}
``` (from `maligned`)

File: lib/syncthing/syncthing.go [hunk 3]
```diff
@@ -0,0 +1,505 @@
...
+	// Event subscription for the API; must start early to catch the early
+	// events. The LocalChangeDetected event might overwhelm the event
+	// receiver in some situations so we will not subscribe to it here.
+	defaultSub := events.NewBufferedSubscription(events.Default.Subscribe(api.DefaultEventMask), api.EventSubBufferSize)
+	diskSub := events.NewBufferedSubscription(events.Default.Subscribe(api.DiskEventMask), api.EventSubBufferSize)
+
+	// Attempt to increase the limit on number of open files to the maximum
+	// allowed, in case we have many peers. We don't really care enough to
+	// report the error if there is one.
+	osutil.MaximizeOpenFileLimit()
```
golangcibot: Error return value of `osutil.MaximizeOpenFileLimit` is not checked (from `errcheck`)

lib/syncthing/syncthing.go
```diff
@@ -0,0 +1,505 @@
...
+	if a.opts.DeadlockTimeoutS > 0 {
+		m.StartDeadlockDetector(time.Duration(a.opts.DeadlockTimeoutS) * time.Second)
+	} else if !build.IsRelease || build.IsBeta {
+		m.StartDeadlockDetector(20 * time.Minute)
+	}
+
+	// Add and start folders
+	for _, folderCfg := range a.cfg.Folders() {
+		if folderCfg.Paused {
+			folderCfg.CreateRoot()
```
golangcibot: Error return value of `folderCfg.CreateRoot` is not checked (from `errcheck`)

lib/syncthing/syncthing.go
```diff
@@ -0,0 +1,505 @@
...
+		}
+	}
+
+	// Candidate builds always run with usage reporting.
+
+	if opts := a.cfg.Options(); build.IsCandidate {
+		l.Infoln("Anonymous usage reporting is always enabled for candidate releases.")
+		if opts.URAccepted != ur.Version {
+			opts.URAccepted = ur.Version
+			a.cfg.SetOptions(opts)
```
golangcibot: Error return value of `a.cfg.SetOptions` is not checked (from `errcheck`)

lib/syncthing/syncthing.go
```diff
@@ -0,0 +1,505 @@
...
+			a.cfg.Save()
```
golangcibot: Error return value of `a.cfg.Save` is not checked (from `errcheck`)

lib/syncthing/syncthing.go
```diff
@@ -0,0 +1,505 @@
...
+			// Unique ID will be set and config saved below if necessary.
+		}
+	}
+
+	// If we are going to do usage reporting, ensure we have a valid unique ID.
+	if opts := a.cfg.Options(); opts.URAccepted > 0 && opts.URUniqueID == "" {
+		opts.URUniqueID = rand.String(8)
+		a.cfg.SetOptions(opts)
```
golangcibot: Error return value of `a.cfg.SetOptions` is not checked (from `errcheck`)

lib/syncthing/syncthing.go
```diff
@@ -0,0 +1,505 @@
...
+		a.cfg.Save()
```
golangcibot: Error return value of `a.cfg.Save` is not checked (from `errcheck`)

lib/syncthing/syncthing.go
```diff
@@ -0,0 +1,505 @@
...
+	select {
+	case <-a.stopped:
+	case <-a.stop:
+	default:
+		close(a.stop)
+	}
+	<-a.stopped
+	// If there was already an exit status set internally, ignore the reason
+	// given to Stop.
+	if a.exitStatus == ExitSuccess {
```
AudriusButkevicius: The comment seems the opposite of the implementation.

AudriusButkevicius:
In principle I like the change, but as you say, this has some things left out.

lib/syncthing/syncthing.go
imsodin: Yes, the comment is misleading, probably even incorrect: `ExitSuccess` is the zero value of  `exitStatus` (nobody set an error), so set to `stopReason`.

imsodin:
> In principle I like the change, but as you say, this has some things left out.

Whether this is a desired direction in general is what I wanted to check, so I am pretty happy about that feedback - thanks :)

And I think it would also make sense on it's own (without changing logger etc.), as an incremental step. Doing all at once "optimally" seems like sure way to make sure it never happens. What needs doing is hashing out the cmd vs lib scopes, so if you have any opinions on those, I'd be happy to hear them.

AudriusButkevicius:
So what do we want to do with this?
I like the changes, I think lib/syncthing should not have any interactions with the filesystem itself, and have the main bootstrap the lib.

I don't mind doing this incrementally, but I would prefer if there was commitment to drive this home.
If we can't drive this home, I'd rather live in the current world oppose to some half-way house.

calmh:
I disagree, I think this is a good and worthwhile change on its own by just separating startup and monitor and stuff from the meat of the application. Shipit.

imsodin:
> So what do we want to do with this?

I mostly did it for what Jakob mentioned: Separating the command line/monitor stuff from the actual "Syncthing stuff" and proper termination (e.g. don't use os.Exit anywhere). Most of this could also be done without lib/syncthing, but it seemed related enough to also make a step towards the library goal. However I don't know the specific requirements for that except for what was stated in related discussions, which is proper termination, "clean" interface and logging.

Essentially I hoped I'd get input what would be required for this library to be useful, e.g. for compilation as a C library, which I remember @calmh having done experimentally and a user on the forum doing (unfortunately I don't remember the context and can't find it) or for android (storage access), where @AudriusButkevicius looked into quite extensively at some point.

AudriusButkevicius:
Sorry I haven't had time to contribute to the conversations brewing on the PRs as I was out most of the weekend. Will try to find some time soon.

imsodin:
For funsies I tried creating bindings with gomobile. That doesn't work with modules, so needs to be worked around with vendoring and GO111MODULE=off. Then binding lib/syncthing works fine, but we also need config and db. That doesn't work due to whatever incompatibilities, but can be worked around by creating some dummy wrapper package that only contains functions generate a config/db. So I was wondering whether it made sense to add such utility functions to this package (lib/syncthing), that allow generating config object from file descriptor or even text as string and creating a database from a location. Potentially the same for the certificate (haven't tested that yet).

As there is no abstraction on the db side of things, I am actually in favor of reading the database from disk in lib/syncthing and removing it from the interface. And maybe instead of passing the config object, pass a file descriptor to the config in the interface? Basically my idea is to only put "low-level" types into the interface.

AudriusButkevicius:
I think the DB should be an interface, which java world can implement.
Config should have a config.FromContent(string baa) I think.

imsodin:
Ahm thanks for the approval :)  
May I still do something about the discussed interface points? In my opinion the current spirit of versioning would require that changing the lib/service interface is a minor version bump, so I'd like this PR to be somewhat polished (I know it will probably still change, but at least I'd like to not know for sure that it will change when merging :D ).

AudriusButkevicius:
I think lets get the bulk of it out of the way and iterate on it later.

imsodin:
See #5830 for db interface.

And I totally forgot to check #4085: It was @Martchus who has done quite a bit of work on a c++ interface. How does this PR fit in with what you did?

Martchus:
Good question. I'll have to look into this PR to give an answer.

But I haven't done that much - just the minimum to be able to start Syncthing as library with a callback for log messages. And I incorporated the build of Syncthing as a library within CMake.

The only thing which was really missing was the ability to stop the beast again and to prevent it from terminating the entire application on a fatal error. Let's see whether this PR implements these points.

Martchus:
@imsodin Oh, and of course thanks for taking the effort.

AudriusButkevicius:
:shipit:

Martchus:
Seems like handling CLI arguments and creating key and certificate is still done in `cmd/syncthing`. But that things I actually need, too. At least to for the beginning it would be quite convenient to simply pass the configuration via CLI arguments. So I still needed the hacky approach I took last time which is removing all `os.Exit` calls from `syncthingMain` and let it return the exit code instead.

imsodin:
What about now with the utility functions for config, db and certs?

I don't understand the command line part: Why would you need that when you call it directly in C(++)? The options can be passed with `syncthing.Options`, which should be implementable in C, right? Or is there functionality in cmd that you need as well when invoking it as a library? In that case I can probably change that too.

AudriusButkevicius:
Lets iterate on continuation in a new PR.

AudriusButkevicius:
Merge if this still works pls.

Martchus:
I haven't seen the utilities yet. Those should do it.

Passing the options as CLI arguments was only for my own convenience to get started. Passing a string is simple and I don't need to parse anything myself or provide GUI elements for this.

AudriusButkevicius:
I think this is good. If you plan to review this, please give a shout, otherwise I'll click merge next time I stumble upon this.

calmh:
Doit

PR #5806, opened 2019-06-17
lib: Add util.Service as suture.Service template (fixes #5801)

### LINKED ISSUE #5801: panic: close of nil channel in relay client stop

imsodin:
> Yeah this goes back to the "we need a template for suture services"
https://github.com/syncthing/syncthing/pull/5802#issuecomment-502695914

I created such a template and put it into place wherever there already was a `stopped` channel.

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -334,12 +334,6 @@ func (s *service) connect(stop chan struct{}) {
 	var sleep time.Duration
 
 	for {
-		select {
-		case <-stop:
-			return
-		default:
-		}
-
 		cfg := s.cfg.RawCopy()
 
 		bestDialerPrio := 1<<31 - 1 // worse prio won't build on 32 bit
@@ -477,11 +471,16 @@ func (s *service) connect(stop chan struct{}) {
 
 		if initialRampup < sleep {
 			l.Debugln("initial rampup; sleep", initialRampup, "and update to", initialRampup*2)
-			time.Sleep(initialRampup)
+			sleep = initialRampup
 			initialRampup *= 2
 		} else {
 			l.Debugln("sleep until next dial", sleep)
-			time.Sleep(sleep)
+		}
+
+		select {
+		case <-time.After(sleep):
+		case <-stop:
+			return
 		}
 	}
 }
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -184,16 +184,22 @@ func NewService(cfg config.Wrapper, myID protocol.DeviceID, mdl Model, tlsCfg *t
 	// the common handling regardless of whether the connection was
 	// incoming or outgoing.
 
-	service.Add(serviceFunc(service.connect))
-	service.Add(serviceFunc(service.handle))
+	service.Add(util.AsService(service.connect))
+	service.Add(util.AsService(service.handle))
 	service.Add(service.listenerSupervisor)
 
 	return service
 }
 
-func (s *service) handle() {
-next:
-	for c := range s.conns {
+func (s *service) handle(stop chan struct{}) {
+	var c internalConn
+	for {
+		select {
+		case <-stop:
+			return
+		case c = <-s.conns:
+		}
+
 		cs := c.ConnectionState()
 
 		// We should have negotiated the next level protocol "bep/1.0" as part
@@ -298,7 +304,7 @@ next:
 			// config. Warn instead of Info.
 			l.Warnf("Bad certificate from %s at %s: %v", remoteID, c, err)
 			c.Close()
-			continue next
+			continue
 		}
 
 		// Wrap the connection in rate limiters. The limiter itself will
@@ -313,11 +319,11 @@ next:
 		l.Infof("Established secure connection to %s at %s", remoteID, c)
 
 		s.model.AddConnection(modelConn, hello)
-		continue next
+		continue
 	}
 }
 
-func (s *service) connect() {
+func (s *service) connect(stop chan struct{}) {
 	nextDial := make(map[string]time.Time)
 
 	// Used as delay for the first few connection attempts, increases
@@ -328,6 +334,12 @@ func (s *service) connect() {
 	var sleep time.Duration
 
 	for {
+		select {
+		case <-stop:
+			return
+		default:
+		}
+
 		cfg := s.cfg.RawCopy()
 
 		bestDialerPrio := 1<<31 - 1 // worse prio won't build on 32 bit
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -27,6 +27,7 @@ import (
 	"github.com/syncthing/syncthing/lib/scanner"
 	"github.com/syncthing/syncthing/lib/stats"
 	"github.com/syncthing/syncthing/lib/sync"
+	"github.com/syncthing/syncthing/lib/util"
 	"github.com/syncthing/syncthing/lib/watchaggregator"
 )
 
@@ -36,6 +37,7 @@ var scanLimiter = newByteSemaphore(0)
 var errWatchNotStarted = errors.New("not started")
 
 type folder struct {
+	*util.Service
 	stateTracker
 	config.FolderConfiguration
 	*stats.FolderStatisticsReference
@@ -54,7 +56,6 @@ type folder struct {
 	scanNow             chan rescanRequest
 	scanDelay           chan time.Duration
 	initialScanFinished chan struct{}
-	stopped             chan struct{}
 	scanErrors          []FileError
 	scanErrorsMut       sync.Mutex
 
@@ -98,7 +99,6 @@ func newFolder(model *model, fset *db.FileSet, ignores *ignore.Matcher, cfg conf
 		scanNow:             make(chan rescanRequest),
 		scanDelay:           make(chan time.Duration),
 		initialScanFinished: make(chan struct{}),
-		stopped:             make(chan struct{}),
 		scanErrorsMut:       sync.NewMutex(),
 
 		pullScheduled: make(chan struct{}, 1), // This needs to be 1-buffered so that we queue a pull if we're busy when it comes.
@@ -109,7 +109,7 @@ func newFolder(model *model, fset *db.FileSet, ignores *ignore.Matcher, cfg conf
 	}
 }
 
-func (f *folder) Serve() {
+func (f *folder) serve() {
 	atomic.AddInt32(&f.model.foldersRunning, 1)
 	defer atomic.AddInt32(&f.model.foldersRunning, -1)
 
@@ -119,7 +119,6 @@ func (f *folder) Serve() {
 	defer func() {
 		f.scanTimer.Stop()
 		f.setState(FolderIdle)
-		close(f.stopped)
 	}()
 
 	pause := f.basePause()
@@ -254,9 +253,8 @@ func (f *folder) Delay(next time.Duration) {
 	f.scanDelay <- next
 }
 
-func (f *folder) Stop() {
+func (f *folder) stop() {
 	f.cancel()
-	<-f.stopped
 }
 
 // CheckHealth checks the folder for common errors, updates the folder state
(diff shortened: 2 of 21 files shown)


lib/relay/client/client.go
```diff
@@ -39,3 +42,62 @@ func NewClient(uri *url.URL, certs []tls.Certificate, invitations chan protocol.
...
+
+type commonClient struct {
+	*util.Service
+
+	invitations              chan protocol.SessionInvitation
+	closeInvitationsOnFinish bool
+
+	stop chan struct{}
+
+	conn *tls.Conn
```
golangcibot: `conn` is unused (from `structcheck`)

AudriusButkevicius:
I was imagining it differently, it essentially `util.AsSuture(func (closed chan struct{}) service.State) suture.Service`, where state can be IsComplete I guess. Idk. Or no return value from the runner func.

AudriusButkevicius:
Essentially you just have a function that runs for the duration of the channel being open.
There could be an alternative implementation that returns an error, which could be used for listener services etc.

imsodin:
Who closes the channel? As it is I don't get what you mean, maybe some pseudocode for a "sample implementation".

AudriusButkevicius:
The wrapper implements stop and start, the function which we wrap just gets a channel as an argument which the wrapper closes on call to stop. On the phone so pseudo code is hard.

imsodin:
At the very least this needs a second channel, that the caller closes after the first channel closed, so that the service thingy knows stopping has finished. At that point as I understand your proposal we are essentially back to my PR (except that I return util.Service, which admittedly might be just suture.Service and the struct type can be internalized, but that seems like a minor detail).

AudriusButkevicius:
Well it knows when the function returned, it can have a second channel internally.

imsodin:
Ah yes, I think I am finally getting now how it should work, sorry for the slowness :)  
However the more I seem to understand the less it differs from my PR: The only difference I see is that on close, in your version a close channel is closed, in my version a close function is called. I chose the latter, as it is more flexible. In some cases cleanup needs to be done regardless of whether serve is running or not and in other places (folder) the closing mechanism isn't a channel, but a context.

AudriusButkevicius:
The problem with your proposal is that you have to manage the lifetime of the channel yourself, which is essentially what blew up in the face in the relay service, so I'd like to have a thing that literally deals with
`func(close chan struct{}) error`

imsodin:
Is this (close to) what you want?

AudriusButkevicius:
Yeah this is more like it.

I guess we could have a second implementation that returns
```
type ServiceWithError interface {
  suture.Service
  Error() error
}
```

That wraps `func(chan) error`

imsodin:
Sure we could, that smells a bit like feature creep to me though ;)

lib/relay/client/static.go
```diff
@@ -332,13 +241,17 @@ func performHandshakeAndValidation(conn *tls.Conn, uri *url.URL) error {
 	return nil
 }
 
-func messageReader(conn net.Conn, messages chan<- interface{}, errors chan<- error) {
+func messageReader(conn net.Conn, messages chan<- interface{}, errors chan<- error, stop chan struct{}) {
```
golangcibot: `conn` can be `io.Reader` (from `interfacer`)

lib/relay/client/static.go
```diff
@@ -137,52 +99,28 @@ func (c *staticClient) Serve() {
 
 			case protocol.RelayFull:
 				l.Infof("Disconnected from relay %s due to it becoming full.", c.uri)
-				c.setError(fmt.Errorf("Relay full"))
-				c.disconnect()
+				return fmt.Errorf("Relay full")
```
golangcibot: ST1005: error strings should not be capitalized (from `stylecheck`)

lib/relay/client/static.go
```diff
@@ -137,52 +99,28 @@ func (c *staticClient) Serve() {
...
 
-		case <-c.stop:
+		case <-stop:
 			l.Debugln(c, "stopping")
-			c.setError(nil)
-			c.disconnect()
+			return nil
 
-		// We always exit via this branch of the select, to make sure the
-		// the reader routine exits.
```
imsodin: I now pass the closed channel to that routine to remove this tedious requirement.

imsodin:
@AudriusButkevicius Added that option and adapted relay clients to use it (which required a bit of a refactor, but made things simpler in my opinion).

lib/connections/service.go
```diff
@@ -313,11 +319,11 @@ next:
 		l.Infof("Established secure connection to %s at %s", remoteID, c)
 
 		s.model.AddConnection(modelConn, hello)
-		continue next
+		continue
 	}
 }
 
-func (s *service) connect() {
+func (s *service) connect(stop chan struct{}) {
```
golangcibot: cyclomatic complexity 34 of func `(*service).connect` is high (> 30) (from `gocyclo`)

lib/connections/service.go
```diff
@@ -328,6 +334,12 @@ func (s *service) connect() {
 	var sleep time.Duration
 
 	for {
+		select {
+		case <-stop:
+			return
+		default:
+		}
```
AudriusButkevicius: There is a massive sleep at the moment, the stop should be plumbed in there.

lib/model/folder.go
```diff
@@ -256,7 +256,7 @@ func (f *folder) Delay(next time.Duration) {
 
 func (f *folder) Stop() {
 	f.cancel()
-	<-f.stopped
+	f.Service.Stop()
```
AudriusButkevicius: Shouldn't we just do defer f.cancel() in serve?

lib/model/folder.go
```diff
@@ -119,7 +120,6 @@ func (f *folder) Serve() {
 	defer func() {
 		f.scanTimer.Stop()
 		f.setState(FolderIdle)
-		close(f.stopped)
```
AudriusButkevicius: f.cancel() here?

lib/relay/client/dynamic.go
```diff
@@ -14,45 +14,29 @@ import (
...
 type dynamicClient struct {
-	pooladdr                 *url.URL
-	certs                    []tls.Certificate
-	invitations              chan protocol.SessionInvitation
-	closeInvitationsOnFinish bool
-	timeout                  time.Duration
-
-	mut    sync.RWMutex
-	err    error
+	*commonClient
```
AudriusButkevicius: does this need to be a pointer?

lib/relay/client/dynamic.go
```diff
@@ -115,24 +93,23 @@ func (c *dynamicClient) Serve() {
...
 
 func (c *dynamicClient) Stop() {
+	c.ServiceWithError.Stop()
 	c.mut.RLock()
 	defer c.mut.RUnlock()
-	close(c.stop)
-	if c.client == nil {
-		return
+	if c.client != nil {
+		c.client.Stop()
```
AudriusButkevicius: Make the service defer client stop?

lib/stun/stun.go
```diff
@@ -100,24 +103,25 @@ func New(cfg config.Wrapper, subscriber Subscriber, conn net.PacketConn) (*Servi
...
-		stop:    make(chan struct{}),
-	}, otherDataConn
+	}
+	s.Service = util.AsService(s.serve)
+	return s, otherDataConn
 }
 
 func (s *Service) Stop() {
-	close(s.stop)
+	s.Service.Stop()
```
AudriusButkevicius: Again, can the stunConn be closed in serve?

lib/ur/usage_report.go
```diff
@@ -422,23 +411,16 @@ func (s *Service) VerifyConfiguration(from, to config.Configuration) error {
...
 	if from.Options.URAccepted != to.Options.URAccepted || from.Options.URUniqueID != to.Options.URUniqueID || from.Options.URURL != to.Options.URURL {
-		s.stopMut.RLock()
 		select {
 		case s.forceRun <- struct{}{}:
-		case <-s.stop:
+		default:
+			// s.forceRun is one buffered, so even though nothing
+			// was sent, a run will still happen after this point.
 		}
-		s.stopMut.RUnlock()
```
AudriusButkevicius: There is a change in behaviour here. This used to block until the forceRun completed potentially, or atleast was accepted.
That might no longer be the case. I suspect it's ok, but suggest you verify if it really is.

lib/util/utils.go
```diff
@@ -170,3 +174,69 @@ func Address(network, host string) string {
...
+	s.err = err
+	s.mut.Unlock()
+}
+
+func (s *service) Stop() {
+	s.mut.Lock()
+	s.stopping = true
+	s.mut.Unlock()
+	close(s.stop)
+	<-s.stopped
```
AudriusButkevicius: I think these services can never be restarted, as s.stop is only initialised once?
Also it feels wrong touching s.stop and s.stopped outside of the lock, as they are clearly initialised under the lock, so I feel this is racey.

lib/stun/stun.go
imsodin: In this case it looks like no, as the `stunConn` is established in `New`, not in `serve`.

lib/model/folder.go
imsodin: The cancelled context is also used in functions within serve, e.g. puller, so that would be too late.

lib/ur/usage_report.go
imsodin: The relevant select above just has three options: stop, forceRun (resets timer) and timer fired (send a usage report). So the only way there could be a significant difference is if at the time of this being called, a report is being sent. Then this would have blocked until that is finished, now it immediately unblocks. The result is the same: As soon as the usage report goes through, another one happens. And on the side that changes the config I can't see any reason why someone should want to wait on the previous (not the scheduled) usage report being finished.

lib/util/utils.go
imsodin: They indeed can't be restarted after `Stop`, that's by design as mandated by suture:

> [...] A supervisor will call .Stop() only once. Thus, it may be as destructive as it likes to get the service to stop.
> 
> Once Stop has been called on a Service, the Service SHOULD NOT be reused in any other supervisor! [...]

I added `stopping` after a race came up in testing - in hindsight that's an ugly bolt-on. I revised the stop/stopped-locking-machinism, I hope it's clearer now (and should be race free).

imsodin:
Regarding name of services:  
The object returned by `util.AsService` is always embedded into another object, which itself can/should implement `fmt.Stringer`. I think it's always a good thing for any non-trivial types to implement that, otherwise all it's components are accessed when being logged, which could be racy. In a few places we just "convert a function to a service", in which case there was neither a name before nor now, I could either add another wrapper (`ServiceWithName`) or add a name parameter to `AsService`, which will then be somewhat redundant, as in most places the value passed will just be `.String()`.  
Any opinions on which is the better variant?

calmh:
Do nothing about the naming, sounds like status quo is fine

lib/model/folder.go
AudriusButkevicius: I guess we could in the run do
```
go func()
  select {
   case <-stop:
     f.cancel()
  }
}()
```

Yet I don't feel strongly about this.

AudriusButkevicius:
https://media.giphy.com/media/GGMUpbtiNX4C4/giphy.gif

PR #5808, opened 2019-06-18
lib/versioner: Fix staggered version cleanup and all versioner restoration (fixes #5807)

### LINKED ISSUE #5807: Staggered ver: WARNING: versioner: lstat … no such file or directory
### version info
Syncthing: 1.2.0-rc.1, Linux (64 bit)
OS: Debian/sid
Browser: Firefox/Nightly

### Problem
Since first RC (first error was logged at 1.2.0-rc.1 update) every restart gives lots of
```
[ZILLL] 2019/06/18 14:13:21.144061 DEBUG: Versioner: Expiring versions [{kkk/IP 2015.pdf {0 63600387389 0x164dc20}}]
[ZILLL] 2019/06/18 14:13:21.144064 DEBUG: Versioner: File over maximum age -> delete  kkk/IP 2015.pdf
[ZILLL] 2019/06/18 14:13:21.144072 WARNING: versioner: lstat /path/syncthing/data/qwz/.stversions/kkk/IP 2015.pdf: no such file or directory
```
for various directories containing Staggered versions. 

I have not switched versioning method. Trying to restore gives
```
Restore Versions
Some items could not be restored:
teszt/read.me 	version not found
```

The files exist, as
```
  File: /path/syncthing/data/wmm/.stversions/kkk/IP 2015~20180802-100304.pdf
  Size: 48092           Blocks: 96         IO Block: 4096   regular file
Device: fd0ch/64780d    Inode: 270277611   Links: 1
Access: (0640/-rw-r-----)  Uid: ( 1000/    grin)   Gid: ( 1000/    grin)
Access: 2016-06-01 16:16:29.000000000 +0200
Modify: 2016-06-01 16:16:29.000000000 +0200
Change: 2018-08-02 10:03:04.284463709 +0200
 Birth: -
```

### More details
...at https://forum.syncthing.net/t/warning-versioner-lstat-no-such-file-or-directory/13390/3

AudriusButkevicius:
Idk, wrote the code, tests pass, did not test.

The problem is that version restoration now happens based on mtime of the file not the name, so restoring requires looking up all versions.

### DIFF
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -2852,9 +2852,7 @@ func TestVersionRestore(t *testing.T) {
 	defer cleanupModel(m)
 	m.ScanFolder("default")
 
-	sentinel, err := time.ParseInLocation(versioner.TimeFormat, "20200101-010101", time.Local)
-	must(t, err)
-	sentinelTag := sentinel.Format(versioner.TimeFormat)
+	sentinel, err := time.ParseInLocation(versioner.TimeFormat, "20180101-010101", time.Local)
 
 	for _, file := range []string{
 		// Versions directory
@@ -2866,7 +2864,6 @@ func TestVersionRestore(t *testing.T) {
 		".stversions/dir/file~20171210-040406.txt",
 		".stversions/very/very/deep/one~20171210-040406.txt", // lives deep down, no directory exists.
 		".stversions/dir/existing~20171210-040406.txt",       // exists, should expect to be archived.
-		".stversions/dir/file.txt~20171210-040405",           // old tag format, supported
 		".stversions/dir/cat",                                // untagged which was used by trashcan, supported
 
 		// "file.txt" will be restored
@@ -2897,7 +2894,7 @@ func TestVersionRestore(t *testing.T) {
 		"file.txt":               1,
 		"existing":               1,
 		"something":              1,
-		"dir/file.txt":           4,
+		"dir/file.txt":           3,
 		"dir/existing.txt":       1,
 		"very/very/deep/one.txt": 1,
 		"dir/cat":                1,
@@ -2942,6 +2939,8 @@ func TestVersionRestore(t *testing.T) {
 		"very/very/deep/one.txt": makeTime("20171210-040406"),
 	}
 
+	beforeRestore := time.Now().Truncate(time.Second)
+
 	ferr, err := m.RestoreFolderVersions("default", restore)
 	must(t, err)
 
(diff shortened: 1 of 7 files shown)


AudriusButkevicius:
Idk, I am contemplating on inverting this whole "mtime is the version time" stuff. We supposedly act as good guys and store the mtime in the filename, but in practice never use it.

AudriusButkevicius:
Perhaps we back the original PR out in the RC, and try again later?
Or do we wish to get it right in the next RC?

AudriusButkevicius:
Or do you wish to fix this, change behaviour and roll forward?

uok:
Personally I prefer "keep original mtime, add deletion timestamp to filename" mode. I find it easier in daily life (e.g. just remove timestamp and move file back) but I understand that each Syncthing user has different view on that.

calmh:
I don’t care, so to me the less complex solution is the better one. Also I didn’t read any code (on mobile) but multiple globs just to find the thing named by a given name + time combo sounds horrible. Perhaps we can glob once and keep passing around the entries as {real name, friendly name, mtime, deltime} structs or something. 

Also I think we can forget about old style patterns nowadays.

lib/model/model_test.go
```diff
@@ -2852,9 +2852,7 @@ func TestVersionRestore(t *testing.T) {
 	defer cleanupModel(m)
 	m.ScanFolder("default")
 
-	sentinel, err := time.ParseInLocation(versioner.TimeFormat, "20200101-010101", time.Local)
-	must(t, err)
-	sentinelTag := sentinel.Format(versioner.TimeFormat)
+	sentinel, err := time.ParseInLocation(versioner.TimeFormat, "20180101-010101", time.Local)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

lib/versioner/util.go
```diff
@@ -203,28 +213,24 @@ func restoreFile(src, dst fs.Filesystem, filePath string, versionTime time.Time,
...
-	for _, candidate := range []string{taggedFilename, oldTaggedFilename, untaggedFileName} {
-		if info, err := src.Lstat(candidate); fs.IsNotExist(err) || !info.IsRegular() {
-			continue
-		} else if err != nil {
-			// All other errors are fatal
-			return err
-		} else if candidate == untaggedFileName && !info.ModTime().Truncate(time.Second).Equal(versionTime) {
-			// No error, and untagged file, but mtime does not match, skip
-			continue
+	if info, err := src.Lstat(taggedFilePath); err == nil && info.IsRegular() {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

AudriusButkevicius:
So this now puts the version back in the filename for easy lookup. Still does hackery for trashcan via mtime.

lib/model/model_test.go
calmh: This bot complaint is legit, err is never read

calmh:
Sounds reasonable, will review at a computer

PR #5815, opened 2019-06-25
gui, lib/model: Fix download progress accounting (fixes #5811)

### LINKED ISSUE #5811: Web interface: download progress of folders do not get updated
### Version Information

Syncthing: v1.1.4, macOS (64 bit)
macOS: 10.14.5 (18F132)
Browsers:
- Safari: 12.1.1 (14607.2.6.1.1)
- Firefox: 67.0.3 (64-Bit)
- Chrome: 75.0.3770.100 (Official Build) (64-Bit)

### Expected Behavior

- Download progress of remote folders get updated automatically

### Actual Behavior

- Download progress of remote folders do not get updated and are stuck to size and percentage values of when the web interface was loaded.
- Local status like download/upload rate do get updated
- A `DownloadProgress` object is logged periodically in the browser's console
- No errors are logged in the browser or syncthing
- Same behavior on all browsers listed above
- Progress is displayed correctly when the users refreshed the web interface by reloading

imsodin:
### Purpose

As usual with the web UI this wasn't as straight forward than I though, but this time the reason was in the model: Folder summaries are not sent on download progress events, but they do include information from download progress events. So if one wants to account download progress in the web UI, there's no way to know which download progress events are already accounted for in the last folder summary and which are not. Solution: Don't account download progress in folder summary. Anyone who wants the fine grained updates can do that subscribing to folder summaries and download progress events.

And I noticed the syncing status without a percentage at the end of a sync. In my opinion it's better to say that we are syncing at 100% (deleting stuff) than just syncing with no info, as the latter looks like there's still lots of work to do.

### Testing

Manual testing.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -325,7 +325,7 @@
                   <span ng-switch-when="idle"><span class="hidden-xs" translate>Up to Date</span><span class="visible-xs" aria-label="{{'Up to Date' | translate}}"><i class="fas fa-fw fa-check"></i></span></span>
                   <span ng-switch-when="syncing">
                     <span class="hidden-xs" translate>Syncing</span>
-                    <span ng-show="syncRemaining(folder.id)">({{syncPercentage(folder.id) | percent}}, {{syncRemaining(folder.id) | binary}}B)</span>
+                    <span>({{syncPercentage(folder.id) | percent}}, {{model[folder.id].needBytes | binary}}B)</span>
                   </span>
                   <span ng-switch-when="outofsync"><span class="hidden-xs" translate>Out of Sync</span><span class="visible-xs" aria-label="{{'Out of Sync' | translate}}"><i class="fas fa-fw fa-exclamation-circle"></i></span></span>
                   <span ng-switch-when="faileditems"><span class="hidden-xs" translate>Failed Items</span><span class="visible-xs" aria-label="{{'Failed Items' | translate}}"><i class="fas fa-fw fa-exclamation-circle"></i></span></span>
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -816,22 +816,6 @@ angular.module('syncthing.core')
             return Math.floor(pct);
         };
 
-        $scope.syncRemaining = function (folder) {
-            // Remaining sync bytes
-            if (typeof $scope.model[folder] === 'undefined') {
-                return 0;
-            }
-            if ($scope.model[folder].globalBytes === 0) {
-                return 0;
-            }
-
-            var bytes = $scope.model[folder].globalBytes - $scope.model[folder].inSyncBytes;
-            if (isNaN(bytes) || bytes < 0) {
-                return 0;
-            }
-            return bytes;
-        };
-
         $scope.scanPercentage = function (folder) {
             if (!$scope.scanProgress[folder]) {
                 return undefined;
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -791,7 +791,8 @@ func (m *model) ReceiveOnlyChangedSize(folder string) db.Counts {
 	return db.Counts{}
 }
 
-// NeedSize returns the number and total size of currently needed files.
+// NeedSize returns the number of currently needed files and their total size
+// minus the amount that has already been downloaded.
 func (m *model) NeedSize(folder string) db.Counts {
 	m.fmut.RLock()
 	rf, ok := m.folderFiles[folder]
--- a/lib/model/folder_summary.go
+++ b/lib/model/folder_summary.go
@@ -157,65 +157,68 @@ func (c *folderSummaryService) listenForUpdates() {
 
 		select {
 		case ev := <-sub.C():
-			if ev.Type == events.DeviceConnected {
-				// When a device connects we schedule a refresh of all
-				// folders shared with that device.
-
-				data := ev.Data.(map[string]string)
-				deviceID, _ := protocol.DeviceIDFromString(data["id"])
-
-				c.foldersMut.Lock()
-			nextFolder:
-				for _, folder := range c.cfg.Folders() {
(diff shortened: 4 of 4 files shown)


gui/default/syncthing/core/syncthingController.js
```diff
@@ -808,11 +826,17 @@ angular.module('syncthing.core')
             if (typeof $scope.model[folder] === 'undefined') {
                 return 100;
             }
-            if ($scope.model[folder].globalBytes === 0) {
+            var global = $scope.model[folder].globalBytes;
```
AudriusButkevicius: Global is a keyword

lib/model/model.go
```diff
@@ -809,7 +809,6 @@ func (m *model) NeedSize(folder string) db.Counts {
 			return true
 		})
 	}
-	result.Bytes -= m.progressEmitter.BytesCompleted(folder)
```
AudriusButkevicius: Why is this going away?

imsodin:
Yeah, let me condense the PR description to only contain the relevant bits:

Goal: Update sync percentage on download progress events.

Approach: Add download progress to the synced bytes in JS-land.

Problem: API adds download progress to the synced bytes in folder summary, but does not emit new folder summaries on download progress events. Thus JS-land may account download progress twice and has no way of detecting that.

Solution: Do not add download progress in folder summary, thus aligning the information in folder summaries with what triggers them. Info on download progress can still be obtained from download progress events.

Actually, that didn't get any shorter :|

lib/model/model.go
imsodin: This method is only used in folder summaries, where it is problematic (see https://github.com/syncthing/syncthing/pull/5815#issuecomment-505416623) and per its docstring:

> // NeedSize returns the number and total size of currently needed files.

AudriusButkevicius: I think this should stay, as it changes what the API returns for no good reason.

imsodin: > no good reason

The reason is to fix #5811 as I explained - I'd expect a bit more of an argument to judge that "no good reason".

The other option I see would be to add another method to the model API: `DownloadedBytes`, which cold then be added to folder summaries, which could then be used in the UI to do the accounting - but that's what I call adding an API method for no good reason...

I will happily consider concrete suggestions.

AudriusButkevicius: Right, I was under the assumption this number was not used in the UI, so you are removing this to avoid double deductions?
What happens when a file download is finished, are we guaranteed a new summary event? 
Because files disappear from scope.progress, and percentage drops back down?

calmh: I don’t think there are any guarantees, and there are timing aspects. There will likely be a period of time between the download progress deregistering and a folder update. I suggest we accept that the percentage drops back momentarily, or perhaps alter the GUI so that only folder update messages are allowed to reduce the percentage. The latter might be complicated...

imsodin: I didn't think of the progress being dropped, thanks for pointing it out. Dropping percentages is ugly indeed, I'd go so far to say that I'd rather not reflect partial file downloads in the progress (like it is today) than having that. Weirdly I tested quite a lot manually and never saw this happening (I saw it stay at 100% for some time rather often though). I'll have a look whether I can get it to work without timing problems (or show that it already works :) ).

imsodin:
DO NOT MERGE

I thought this was a minimal change and thus pushed and then tested -> it crashes.

imsodin:
Now it works and while the change is actually tiny (schedule folder summary on download progress) the diff grew unfortunately - I recommend looking just at the second commit (the first is copy-pasting code into a separate function).

The reason I changed my mind is that there already is batching and backup from folder summaries if they take too much time. So just scheduling folder summaries on download progress is simple and doesn't have any of the discussed problems and doesn't change any API.

gui/default/index.html
```diff
@@ -325,7 +325,7 @@ <h4 class="panel-title">
                   <span ng-switch-when="idle"><span class="hidden-xs" translate>Up to Date</span><span class="visible-xs" aria-label="{{'Up to Date' | translate}}"><i class="fas fa-fw fa-check"></i></span></span>
                   <span ng-switch-when="syncing">
                     <span class="hidden-xs" translate>Syncing</span>
-                    <span ng-show="syncRemaining(folder.id)">({{syncPercentage(folder.id) | percent}}, {{syncRemaining(folder.id) | binary}}B)</span>
```
calmh: This is some sort of behavior change, we now always show the span, presumably with "100%, 0 b" in the in sync case?

calmh: Or potentially like `undefined %, NaN b` in some other cases :D

imsodin: > And I noticed the syncing status without a percentage at the end of a sync. In my opinion it's better to say that we are syncing at 100% (deleting stuff) than just syncing with no info, as the latter looks like there's still lots of work to do.

And is there any chance no folder summary was sent when pulling? That's the only reason needbytes may be undefined

calmh: Deleting stuff should be at 95%, I think… I guess the status you’re seeing might be that we’ve got the latest folder summary but not yet the state change?

I guess things could  be undefined when starting up? The existing function checked for it, at least.

imsodin: That's one option, the other is that we pulled everything, so nothing is needed anymore, but the puller is still processing deletions, that it does at the end, and thus the state is still syncing

The 95% thing concerns remote devices only.

Sure they could, but when starting folder status is unknown, then scanning and only then possibly pulling. I can't imagine no summary to have been sent at that point. And index.html is full of unchecked references to model

calmh: My feeling was that since the undefined condition was handled previously, perhaps it was for a reason. But perhaps that reason is obsolete or a misunderstanding.

Personally I don’t think "Syncing, 100%, 0 b" while we are still working is in any way more informative than just "Syncing", but it’s a minor point 🤷‍♂

PR #5819, opened 2019-06-26
lib/model: Correct/unify check if item changed

imsodin:
The previous `itemChanged` already scheduled scanning, but the same was done on the caller side in most places (which isn't problematic, but useless).  I made that behaviour explicit by calling it `scanIfItemChanged`. And it now returns `errModified` if it was changed in case the caller wants to know about that, shortening the code in many places.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -577,13 +577,9 @@ func (f *sendReceiveFolder) handleDir(file protocol.FileInfo, dbUpdateChan chan<
 	case err == nil && !info.IsDir():
 		// Check that it is what we have in the database.
 		curFile, hasCurFile := f.model.CurrentFolderFile(f.folderID, file.Name)
-		if changed, err := f.itemChanged(info, curFile, hasCurFile, scanChan); err != nil {
+		if err := f.scanIfItemChanged(info, curFile, hasCurFile, scanChan); err != nil {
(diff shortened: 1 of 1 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -1514,12 +1506,10 @@ func (f *sendReceiveFolder) performFinish(file, curFile protocol.FileInfo, hasCu
 		// There is an old file or directory already in place. We need to
 		// handle that.
 
-		if changed, err := f.itemChanged(stat, curFile, hasCurFile, scanChan); err != nil {
-			return err
-		} else if changed {
-			l.Debugln("file changed on disk compared to db; not finishing:", file.Name)
-			scanChan <- curFile.Name
-			return errModified
+		if err := f.scanIfItemChanged(info, curFile, hasCurFile, scanChan); err != nil {
```
golangcibot: undeclared name: `info` (from `typecheck`)

lib/model/folder_sendrecv.go
```diff
@@ -1514,12 +1506,10 @@ func (f *sendReceiveFolder) performFinish(file, curFile protocol.FileInfo, hasCu
...
+			err = errors.Wrap(err, "handling file")
+			f.newPullError(file.Name, err)
+			return
```
golangcibot: wrong number of return values (want 1, got 0) (from `typecheck`)

PR #5829, opened 2019-07-03
lib/model, lib/protocol: Track closing connections (fixes #5828)

### LINKED ISSUE #5828: panic: Index update for not running folder
```

  File "C:/BuildAgent/work/174e136266f8a219/lib/model/model.go", line 997, in handleIndex
    panic(fmt.Sprintf("%v for not running folder %q", op, folder))
  File "C:/BuildAgent/work/174e136266f8a219/lib/model/model.go", line 964, in IndexUpdate
    m.handleIndex(deviceID, folder, fs, true)
  File "C:/BuildAgent/work/174e136266f8a219/lib/protocol/nativemodel_windows.go", line 25, in IndexUpdate
    m.Model.IndexUpdate(deviceID, folder, files)
  File "C:/BuildAgent/work/174e136266f8a219/lib/protocol/protocol.go", line 567, in handleIndexUpdate
    c.receiver.IndexUpdate(c.id, im.Folder, im.Files)
  File "C:/BuildAgent/work/174e136266f8a219/lib/protocol/protocol.go", line 429, in dispatcherLoop
    c.handleIndexUpdate(*msg)
  File "C:/BuildAgent/work/174e136266f8a219/lib/protocol/protocol.go", line 255, in func1
    err := c.dispatcherLoop()

panic: Index update for not running folder "qkqer-stxet"
```
https://crash.syncthing.net/report/66/c9/66c93e0353714dea390830d3caa769bf3424832f41673c10bb74d251f1611229  
Sentry Issue: [SYNCTHING-1H](https://sentry.syncthing.net/syncthing/syncthing/issues/65/?referrer=github_integration)

I believe this is due to a race: We close the connection when stopping a folder, but do not wait for the connection to actually be closed before removing the folder elements from the model. So there is a small window where incoming indexes (or other things) can go through after that point. PR incoming.

imsodin:
### Purpose

This change has a few connected elements:

1. Stopping a folder now waits on the closed connections to stop before removing the folder elements from the model (this is the actual fix of #5828).
2. The utility function closing connections now returns the closed channels of these connections, such that the caller can wait on them if required (convenience and doing both things (closing and getting closed channels) under the same lock).
3. Speed up calls to `protocol.Close` and sending close messages in the protocol:  
Close now spawns a goroutine for both sending the close message and closing, while before it spawned the goroutine just for closing after sending the message. Thus the caller can quicker go on with whatever else it needs to do and if it needs to wait, it already now has to wait for receiving the `Closed` callback.  
The second element is a mechanism to stop sending messages on close: Right now the close message competes with any other message to get sent. With the additional `sendStop` chan all messages waiting to be sent abort when a close message is to be sent.

### Testing

None specific to this issue, just spun up the test system and restarted a folder. I don't have an idea how to unit test this race consistently.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -411,6 +411,8 @@ func (m *model) tearDownFolderLocked(cfg config.FolderConfiguration, err error) 
 		m.RemoveAndWait(id, 0)
 	}
 
+	// Wait for connections to stop to ensure that no more calls to methods
+	// expecting this folder to exist happen (e.g. .IndexUpdate).
 	w.Wait()
 
 	m.fmut.Lock()
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -221,10 +221,8 @@ func (m *model) Stop() {
 	for id := range devs {
 		ids = append(ids, id)
 	}
-	closed := m.closeConns(ids, errStopped)
-	for _, c := range closed {
-		<-c
-	}
+	w := m.closeConns(ids, errStopped)
+	w.Wait()
 }
 
 // StartDeadlockDetector starts a deadlock detector on the models locks which
@@ -407,15 +405,13 @@ func (m *model) tearDownFolderLocked(cfg config.FolderConfiguration, err error) 
 	// Close connections to affected devices
 	// Must happen before stopping the folder service to abort ongoing
 	// transmissions and thus allow timely service termination.
-	closed := m.closeConns(cfg.DeviceIDs(), err)
+	w := m.closeConns(cfg.DeviceIDs(), err)
 
 	for _, id := range tokens {
 		m.RemoveAndWait(id, 0)
 	}
 
-	for _, c := range closed {
-		<-c
-	}
+	w.Wait()
 
 	m.fmut.Lock()
 
@@ -1432,10 +1428,9 @@ func (m *model) Closed(conn protocol.Connection, err error) {
 	close(closed)
 }
 
-// closeConns will close the underlying connection for given devices and
-// return an array of channels that will be closed once the connections are
-// finished closing.
-func (m *model) closeConns(devs []protocol.DeviceID, err error) []chan struct{} {
+// closeConns will close the underlying connection for given devices and return
+// a waiter that will return once all the connections are finished closing.
+func (m *model) closeConns(devs []protocol.DeviceID, err error) config.Waiter {
 	conns := make([]connections.Connection, 0, len(devs))
 	closed := make([]chan struct{}, 0, len(devs))
 	m.pmut.Lock()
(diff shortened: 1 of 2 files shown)


lib/protocol/protocol.go
```diff
@@ -867,26 +870,29 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
+		c.sendCloseOnce.Do(func() {
+			close(c.sendStop)
+			done := make(chan struct{})
 			select {
-			case <-done:
+			case c.closeBox <- asyncMessage{&Close{err.Error()}, done}:
+				select {
+				case <-done:
+				case <-timeout.C:
+				case <-c.closed:
```
AudriusButkevicius: This is confusing. We await for something but there are no calls after this, why do we care on waiting for this subselect?

imsodin: We are waiting for `done`/the send to go through (or timeout or connection being closed under our feat) before calling `c.internalClose` further down. While it doesn't make a difference we could also return on the `c.closed` cases (`c.internalClose` is a noop if `c.closed` is closed).

lib/model/model.go
```diff
@@ -1434,23 +1432,36 @@ func (m *model) Closed(conn protocol.Connection, err error) {
...
 		if conn, ok := m.conn[dev]; ok {
 			conns = append(conns, conn)
+			closed = append(closed, m.closed[dev])
 		}
 	}
 	m.pmut.Unlock()
 	for _, conn := range conns {
 		conn.Close(err)
 	}
+	return closed
```
AudriusButkevicius: Not a fan of this []chan approach. We always wait for all of them, perhaps return a Waiter() or a wait group?

imsodin: Went with a `Waiter`, definitely neater that way, thanks.

imsodin:
Not closing connections for reconfiguration would indeed be quite nice, but it would introduce similar timing issues, where you'd have to make sure after re-sending the cluster config that no more calls regarding this folder come in before removing the folder from the model.

calmh:
Didn't read this yet, just noting that

> Stopping a folder now waits on the closed connections to stop before removing the folder elements from the model

this can take minutes, unless there are deadlines wired to it, given block transfers in progress etc. I already noticed in a different context that pausing a device does not immediately disconnect a device nowadays, it can take a while and the device shows as "paused" while still connected... Not sure if related, really, or I'm just talking shit but I'm carefully skeptical about more "wait for x before y" things when the wait might be long and appear to the user like stuff hung or nothing's happening in the meantime.

imsodin:
Closing the connection waits on two things: Sending a close msg, capped at 10s, and the "dispatcher" loop to terminate. The latter calls methods on the model, and that's exactly whats needed to prevent the panic. Calls in that loop that may take a moment are index updates and requests, both of which should take about the same time as what we are already waiting on: Folder to stop (pulling in progress can mean waiting for multiple outstanding requests). So in the end the worst case is a 10s wait.

An alternative is to just silently drop calls to methods for a non-existing folder. That may be fine, but it would be a clear deviation from current principles.

lib/protocol/protocol.go
```diff
@@ -867,26 +870,29 @@ func (c *rawConnection) shouldCompressMessage(msg message) bool {
...
 // argument specifies the reason for closing the connection.
 func (c *rawConnection) Close(err error) {
-	c.sendCloseOnce.Do(func() {
-		done := make(chan struct{})
-		timeout := time.NewTimer(CloseTimeout)
-		select {
-		case c.closeBox <- asyncMessage{&Close{err.Error()}, done}:
+	// Close might be called from a method that is called from within
+	// dispatcherLoop, resulting in a deadlock.
+	// The sending above must happen before spawning the routine, to prevent
```
calmh: I don't understand this comment. Which sending above?

lib/protocol/protocol.go
```diff
@@ -662,12 +664,13 @@ func (c *rawConnection) send(msg message, done chan struct{}) bool {
 	select {
 	case c.outbox <- asyncMessage{msg, done}:
 		return true
+	case <-c.sendStop:
```
calmh: Slightly confusing naming - we're not sending a stop, we're stopping the sender? `preventSends`? I dunno. Maybe leave as is.

lib/protocol/protocol.go
imsodin: Thanks so much - I just didn't read that comment at all and it is explaining why I can't do what I did in this PR before. It's nice that I write comments (often after some nudging), but I really also need to read them...

PR #5830, opened 2019-07-03
lib/db, cmd: Add abstraction level to database (ref #4085)

imsodin:
> I think the DB should be an interface, which java world can implement.
https://github.com/syncthing/syncthing/pull/5805#issuecomment-507815753

I propose this interface as an intermediate step to the lib/syncthing PR (#5805) (from [backend.go](https://github.com/syncthing/syncthing/pull/5830/files#diff-ae0519820d68bd6a365eaf4ddcb6c2e6)): 

```golang
var (
	ErrNotFound = errors.New("given key not found in db")
	ErrClosed   = errors.New("db backend was closed")
)

// Backend is the interface that a database needs to implement to be used with
// Syncthing.
// It is not guaranteed that after calling Close no more calls are made to other
// methods. All returned errors due to methods being called after Close must be
// ErrClosed.
// Errors due to a given key not being found in the database, must be ErrNotFound.
type Backend interface {
	Get(key []byte) ([]byte, error)
	Has(key []byte) (bool, error)
	Put(key, val []byte) error
	Delete(key []byte) error
	NewBatch() Batch
	NewIterator(slice *util.Range) (iterator.Iterator, error)
	GetSnapshot() (Snapshot, error)
	Close()
}

// Batch is a container to check in several changes at once to the database to
// ensure a semantically consistent state.
//
// CheckFlush should check the size of the Batch write and call Flush in
// appropriate intervals (e.g. when the Batch exceeds a certain size).
//
// Flush commits the contents of the Batch to db and calls Reset on it on success.
type Batch interface {
	CheckFlush() error
	Delete(key []byte)
	Flush() error
	Len() int
	Put(key, value []byte)
	Reset()
}

// Snapshot provides a consistent view of the database at the time of creation.
// It must be released after use by calling Release.
type Snapshot interface {
	Get(key []byte) ([]byte, error)
	Has(key []byte) (bool, error)
	NewIterator(*util.Range) (iterator.Iterator, error)
	Release()
}
```

The separate `ErrorChecker` interface is due to read only transactions, which shouldn't have the full `Backend` to prevent writing.

Most of lowlevel.go moved to goleveldb.go as an implementation of `Backend` (gitk nicely diffs both lowlevel.go and goleveldb.go against the old lowlevel.go, github apparently doesn't recognize/care about the similarity). The rest of the changes are attributed to adding `error` to various methods in the interface to get rid of the hack types for closed dbs (`closedIter` and `closedSnap`).


lib/db/goleveldb.go
```diff
@@ -0,0 +1,289 @@
...
+	"github.com/syndtr/goleveldb/leveldb/util"
+)
+
+const (
+	goleveldbMaxOpenFiles = 100
+	goleveldbWriteBuffer  = 16 << 20
+	goleveldbFlushBatch   = goleveldbWriteBuffer / 4 // Some leeway for any leveldb in-memory optimizations
+)
+
+type goleveldb struct {
```
golangcibot: struct of size 40 bytes could be of size 32 bytes:
```
struct{
	DB      	*github.com/syncthing/syncthing/vendor/github.com/syndtr/goleveldb/leveldb.DB,
	closeMut	*sync.RWMutex,
	iterWG  	sync.WaitGroup,
	closed  	bool,
}
``` (from `maligned`)

File: lib/db/namespaced.go [hunk 2]
```diff
@@ -36,28 +36,32 @@ func NewNamespacedKV(db *Lowlevel, prefix string) *NamespacedKV {
...
+	batch.Flush()
 }
 
 // PutInt64 stores a new int64. Any existing value (even if of another type)
 // is overwritten.
 func (n *NamespacedKV) PutInt64(key string, val int64) {
 	var valBs [8]byte
 	binary.BigEndian.PutUint64(valBs[:], uint64(val))
-	n.db.Put(n.prefixedKey(key), valBs[:], nil)
+	n.db.Put(n.prefixedKey(key), valBs[:])
```
golangcibot: Error return value of `n.db.Put` is not checked (from `errcheck`)

lib/db/namespaced.go
```diff
@@ -69,14 +73,14 @@ func (n *NamespacedKV) Int64(key string) (int64, bool) {
 // type) is overwritten.
 func (n *NamespacedKV) PutTime(key string, val time.Time) {
 	valBs, _ := val.MarshalBinary() // never returns an error
-	n.db.Put(n.prefixedKey(key), valBs, nil)
+	n.db.Put(n.prefixedKey(key), valBs)
```
golangcibot: Error return value of `n.db.Put` is not checked (from `errcheck`)

lib/db/namespaced.go
```diff
@@ -87,13 +91,13 @@ func (n NamespacedKV) Time(key string) (time.Time, bool) {
 // PutString stores a new string. Any existing value (even if of another type)
 // is overwritten.
 func (n *NamespacedKV) PutString(key, val string) {
-	n.db.Put(n.prefixedKey(key), []byte(val), nil)
+	n.db.Put(n.prefixedKey(key), []byte(val))
```
golangcibot: Error return value of `n.db.Put` is not checked (from `errcheck`)

lib/db/namespaced.go
```diff
@@ -139,7 +143,7 @@ func (n NamespacedKV) Bool(key string) (bool, bool) {
 // Delete deletes the specified key. It is allowed to delete a nonexistent
 // key.
 func (n NamespacedKV) Delete(key string) {
-	n.db.Delete(n.prefixedKey(key), nil)
+	n.db.Delete(n.prefixedKey(key))
```
golangcibot: Error return value of `n.db.Delete` is not checked (from `errcheck`)

lib/db/set.go
```diff
@@ -292,10 +292,14 @@ func DropFolder(ll *Lowlevel, folder string) {
...
-	dbi := db.NewIterator(util.BytesPrefix([]byte{KeyTypeIndexID}), nil)
+	dbi, err := db.NewIterator(util.BytesPrefix([]byte{KeyTypeIndexID}))
+	if err != nil {
+		l.Debugln("NewIterator:", err)
+		return
+	}
 	defer dbi.Release()
 	for dbi.Next() {
-		db.Delete(dbi.Key(), nil)
+		db.Delete(dbi.Key())
```
golangcibot: Error return value of `db.Delete` is not checked (from `errcheck`)

lib/db/smallindex.go
```diff
@@ -82,7 +84,7 @@ func (i *smallIndex) ID(val []byte) uint32 {
 	key := make([]byte, len(i.prefix)+8) // prefix plus uint32 id
 	copy(key, i.prefix)
 	binary.BigEndian.PutUint32(key[len(i.prefix):], id)
-	i.db.Put(key, val, nil)
+	i.db.Put(key, val)
```
golangcibot: Error return value of `i.db.Put` is not checked (from `errcheck`)

lib/db/smallindex.go
```diff
@@ -115,7 +117,7 @@ func (i *smallIndex) Delete(val []byte) {
 		// Put an empty value into the database. This indicates that the
 		// entry does not exist any more and prevents the ID from being
 		// reused in the future.
-		i.db.Put(key, []byte{}, nil)
+		i.db.Put(key, []byte{})
```
golangcibot: Error return value of `i.db.Put` is not checked (from `errcheck`)

AudriusButkevicius:
Actually rethinking my own words, I wonder if its likely that Java will implement its own database.

What was the problem with gomobile in general? Does this help solve it? Perhaps turning LowLevel into an interface solves this too and we rid of one of the abstractions?

lib/db/lowlevel.go
```diff
@@ -99,208 +104,638 @@ func (db *Lowlevel) Committed() int64 {
...
-	case errors.IsCorrupted(err):
-		return true
+	vl, ok := unmarshalVersionList(bs)
+	if !ok {
+		return nil
+	}
 
-	case strings.Contains(err.Error(), "corrupted"):
-		return true
+	var devices []protocol.DeviceID
```
golangcibot: Consider preallocating `devices` (from `prealloc`)

lib/db/lowlevel.go
```diff
@@ -99,208 +104,638 @@ func (db *Lowlevel) Committed() int64 {
...
+		l.Debugln("newReadWriteTransaction:", err)
+		return
+	}
+	defer t.close()
+
+	for _, key := range [][]byte{
+		// Remove all items related to the given folder from the device->file bucket
+		db.keyer.GenerateDeviceFileKey(nil, folder, nil, nil).WithoutNameAndDevice(),
+		// Remove all sequences related to the folder
+		db.keyer.GenerateSequenceKey(nil, []byte(folder), 0).WithoutSequence(),
```
golangcibot: unnecessary conversion (from `unconvert`)

imsodin:
The distinction between `Lowlevel` and `instance` is indeed not necessary - I removed that.

Lowlevel holds some state which has nothing to do with the db backend, so that wouldn't make sense to include in that interface. In general this is not directly relevant to the gomobile problem, it's just that with a user of the library could either choose to use our existing implementation of the `Backend` interface through goleveldb or implement it itself. Having `*LowLevel` in the library interface seems pretty obscure and any user definitely needs to just use our implementation. So in essence in my opinion it's either something like the interface in this PR or then we can just put `dbPath string` into the interface.

lib/db/goleveldb.go
```diff
@@ -0,0 +1,289 @@
...
+	*leveldb.DB
+	closed   bool
+	closeMut *sync.RWMutex
+	iterWG   sync.WaitGroup
+}
+
+// OpenGoleveldb attempts to open the database at the given location, and runs
+// recovery on it if opening fails. Worst case, if recovery is not possible,
+// the database is erased and created from scratch.
+func OpenGoleveldb(location string) (Backend, error) {
```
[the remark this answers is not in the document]
AudriusButkevicius: Also, why does this need to be public? For util binaries?

lib/db/goleveldb.go
```diff
@@ -0,0 +1,289 @@
...
+	defer db.closeMut.RUnlock()
+	if db.closed {
+		return leveldb.ErrClosed
+	}
+	return db.DB.Delete(key, nil)
+}
+
+func (db *goleveldb) IsClosed(err error) bool {
+	return err == leveldb.ErrClosed
+}
```
AudriusButkevicius: I dislike this. net package returns net.Error that has IsTimeout() etc, we should have db.Error,

or db.ErrNotFound (oppose to some checker magic).

lib/db/lowlevel.go
```diff
@@ -99,208 +104,638 @@ func (db *Lowlevel) Committed() int64 {
...
-	return db.DB.Delete(key, wo)
 }
 
-func (db *Lowlevel) NewIterator(slice *util.Range, ro *opt.ReadOptions) iterator.Iterator {
-	return db.newIterator(func() iterator.Iterator { return db.DB.NewIterator(slice, ro) })
+func (db *Lowlevel) withHave(folder, device, prefix []byte, truncate bool, fn Iterator) {
+	t, err := db.newReadOnlyTransaction()
+	if err != nil {
+		l.Debugln("newReadOnlyTransaction:", err)
+		return
```
AudriusButkevicius: We should not be swallowing these silently.
Throughout the file everywhere.

lib/db/lowlevel.go
```diff
@@ -99,208 +104,638 @@ func (db *Lowlevel) Committed() int64 {
...
-	defer db.closeMut.RUnlock()
-	if db.closed {
-		return leveldb.ErrClosed
+// updateLocalFiles adds fileinfos to the db, and updates the global versionlist,
+// metadata, sequence and blockmap buckets.
+func (db *Lowlevel) updateLocalFiles(folder []byte, fs []protocol.FileInfo, meta *metadataTracker) {
+	t, err := db.newReadWriteTransaction()
+	if err != nil {
+		l.Debugln("newReadWriteTransaction:", err)
+		return
```
AudriusButkevicius: bad stuff to hide this

lib/db/lowlevel.go
```diff
@@ -99,208 +104,638 @@ func (db *Lowlevel) Committed() int64 {
...
+	for dbi.Next() {
+		name := db.keyer.NameFromDeviceFileKey(dbi.Key())
+		if len(prefix) > 0 && !bytes.HasPrefix(name, prefix) {
+			return
+		}
+
+		f, err := unmarshalTrunc(dbi.Value(), truncate)
+		if err != nil {
+			l.Debugln("unmarshal error:", err)
+			continue
```
AudriusButkevicius: again, bad stuff

lib/db/lowlevel.go
```diff
@@ -99,208 +104,638 @@ func (db *Lowlevel) Committed() int64 {
...
-	defer it.db.closeMut.RUnlock()
-	if it.db.closed {
-		return false
-	}
-	return fn()
+// unchanged checks if two files are the same and thus don't need to be updated.
+// Local flags or the invalid bit might change without the version
+// being bumped. The IsInvalid() method handles both.
+func unchanged(nf, ef FileIntf) bool {
+	return ef.FileVersion().Equal(nf.FileVersion()) && ef.IsInvalid() == nf.IsInvalid()
```
AudriusButkevicius: The diff of this file is unreadable.
Idk, perhaps make it over several commits or something to make it possible to understand what's yours and what's just moving.

The whole "assume no error" looks like a disasster to me.

lib/db/namespaced.go
```diff
@@ -36,28 +36,32 @@ func NewNamespacedKV(db *Lowlevel, prefix string) *NamespacedKV {
 
 // Reset removes all entries in this namespace.
 func (n *NamespacedKV) Reset() {
-	it := n.db.NewIterator(util.BytesPrefix(n.prefix), nil)
+	it, err := n.db.NewIterator(util.BytesPrefix(n.prefix))
+	if err != nil {
+		l.Debugln("NewIterator:", err)
+		return
```
AudriusButkevicius: BAAAAD

lib/db/namespaced.go
AudriusButkevicius: Let's fix this, we're making a library, failing silently is bad.

lib/db/schemaupdater.go
```diff
@@ -91,10 +91,18 @@ func (db *schemaUpdater) updateSchema() error {
 }
 
 func (db *schemaUpdater) updateSchema0to1() {
-	t := db.newReadWriteTransaction()
+	t, err := db.newReadWriteTransaction()
+	if err != nil {
+		l.Debugln("newReadWriteTransaction:", err)
+		return
```
AudriusButkevicius: BAAD

lib/db/smallindex.go
```diff
@@ -42,7 +41,10 @@ func newSmallIndex(db *leveldb.DB, prefix []byte) *smallIndex {
 // load iterates over the prefix space in the database and populates the in
 // memory maps.
 func (i *smallIndex) load() {
-	it := i.db.NewIterator(util.BytesPrefix(i.prefix), nil)
+	it, err := i.db.NewIterator(util.BytesPrefix(i.prefix))
+	if err != nil {
+		return
```
AudriusButkevicius: bad

lib/db/transactions.go
```diff
@@ -83,18 +88,22 @@ func (t readOnlyTransaction) getGlobal(keyBuf, folder, file []byte, truncate boo
...
+	return readWriteTransaction{
+		readOnlyTransaction: t,
+		Batch:               db.NewBatch(),
+	}, nil
 }
 
 func (t readWriteTransaction) close() {
-	t.flush()
+	_ = t.Flush()
 	t.readOnlyTransaction.close()
```
AudriusButkevicius: This could return an error, and tell us that it fucked up and didn't do anything so we can let the user know?

lib/model/testutils_test.go
```diff
@@ -116,7 +116,7 @@ func setupModel(w config.Wrapper) *model {
 	return m
 }
 
-func newModel(cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersion string, ldb *db.Lowlevel, protectedFiles []string) *model {
+func newModel(cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersion string, ldb *db.Instance, protectedFiles []string) *model {
```
golangcibot: `newModel` - `clientName` always receives `"syncthing"` (from `unparam`)

cmd/stindex/idxck.go
```diff
@@ -31,7 +31,7 @@ type sequenceKey struct {
 	sequence uint64
 }
 
-func idxck(ldb *db.Lowlevel) (success bool) {
+func idxck(ldb *db.Instance) (success bool) {
```
golangcibot: cyclomatic complexity 40 of func `idxck` is high (> 30) (from `gocyclo`)

imsodin:
@AudriusButkevicius I reworked the commits to be more readable and incorporated some of your feedback. I have not done anything about the error handling. I agree that it is/was abysmal and needs improvement. I just first need to think about what we actually want to do: As any error that's not not found/closed is probably fatal, at the moment we either ignore it or panic. I am unsure what the most useful thing would be (on a high level): Shut syncthing down? Display a warning, archive the db and start over on a new one?

lib/db/lowlevel.go
```diff
@@ -7,88 +7,93 @@
...
+}
+
+// Batch is a container to check in several changes at once to the database to
+// ensure a semantically consistent state.
+//
+// CheckFlush should check the size of the Batch write and call Flush in
+// appropriate intervals (e.g. when the Batch exceeds a certain size).
+//
+// Flush commits the contents of the Batch to db and calls Reset on it on success.
+type Batch interface {
```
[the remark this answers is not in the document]
imsodin: Im not much of a fan of the "er" suffix, it seems pointless to me. It's the general type that defines how a batch has to behave, and implementations are then `mysqlBatch`, `leveldbBatch` and so on. If any suffix I'd like "Intf" much better. Just imagine "Backender", "Modeller", "Snapshotter", ... urgh

lib/db/lowlevel.go
```diff
@@ -7,88 +7,93 @@
...
+	CheckFlush() error
```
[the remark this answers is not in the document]
imsodin: My reasoning is, that it can depend on the db when a good size to flush is reached. Letting the db flush on `Put` if necessary would invalidate the purpose of `Batch`: Collecting changes together that should be done atomically (semantically). I.e. `CheckFlush` is caller saying to the db, that it may flush at this point if the db thinks it's a good idea, and `Flush` says do it now (likely because operation ended and nothing more is being added to Batch).

I could create an `flushBatch` type used by `Interface` that takes care of flushing based on size regardless of the underlying `Backend`, then `CheckFlush` could be removed from the interface.

lib/db/lowlevel.go
```diff
@@ -7,88 +7,93 @@
...
+// It is not guaranteed that after calling Close no more calls are made to other
+// methods (see ErrorChecker.IsClosed).
+type Backend interface {
+	ErrorChecker
+	Get(key []byte) ([]byte, error)
+	Has(key []byte) (bool, error)
+	Put(key, val []byte) error
+	Delete(key []byte) error
+	NewBatch() Batch
+	NewIterator(slice *util.Range) (iterator.Iterator, error)
```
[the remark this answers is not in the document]
imsodin: I thought as this are packages separate from the meat of the golevel implementation it could be fine to use them, but sure makes sense to have the full interface/types here. As it is plainly at least inspired by syndtr/goleveldb, I probably need to attribute that somehow? Maybe a comment at the top referencing to `LICENSE-syndtr` which contains the copyright (BSD 2-clause, which is fine with MPL I think). Or am I overthinking this?

imsodin:
I have looked a bit into it, and bubbling all errors up to the exported methods in lib/db alone is already huge. Then there is a few places in model where those methods are used: Some in synchronous services (e.g. folders) and some asynchronously (e.g. remote indexes). Any solution that wants to "react moderately" to db errors (e.g. restart folder or connection) would need some kind of error tracking to not retry indefinitely and user facing reporting - and even then there's some danger that some effect of a db failure has been misjudged and in worst case even messes up remotes. Therefore I'd tend to just shut down. It should hardly ever happen and whatever controls Syncthing should already take care of reporting the problem and throttling restarts. And if the db is toast startup will fail, if it isn't it is the cleanest possible restart (index exchange etc). It would also have the added benefit that we don't need to bubble errors all the way up: The database could just get an interface to a "Syncthing controller" (like the rest API already has) and trigger exiting on error (after doing whatever error reporting/printing that is suitable).  
What are your thoughts on this?

AudriusButkevicius:
So I do think we should bubble the errors all the way up if we assume this behaves like a library. I don't know how we can do this but I am not a fan of the inverse flow control where we give everyone a "shoot myself" switch.

@calmh what do you think about the error handling issue in general?

calmh:
I didn't read this at all yet because I've been on a business trip this week and have had no time/energy. What are we trying to do, make it easier to switch out the lowest level database? Is that ever going to be a thing?

But if it were, I do think we should properly propagate errors upwards, to the boundary of the db package. It's not the db package's responsibility to decide what an error means.

Then at some point in the stack the reality becomes that we cannot operate without the database is working order so we're going to crash when an error is returned. I think it's fine to move that decision as high up as possible though, potentially into the model layer. There could, at that point, exist a type crashingDatabase that simply wraps the database layers, removing error returns by intercepting them and shutting down when an error happens...

imsodin:
@calmh This PR is very low priority, it's just a consequence of the idea to have an interface in the proposed lib/syncthing interface for the database. Yes, in principle such that an alternative lowest level database could be used - pretty hypothetically. I think it's a nice change, but it's also a huge change that's currently just change for changes sake, so I am close to letting it drop :) 

The entire error discussion mostly sparked because the bot started screaming at all the places I ignored errors and errors are ignored already today, which isn't a nice state to be in. Meaning having actual error handling, by bubbling up or whatever, is something to do, regardless of whether in this PR or not.

PR #5834, opened 2019-07-07
lib/model: Remove pointless error that watch hasn't started (fixes #5833)

### LINKED ISSUE #5833: Paused folder shows filesystem watcher error
Hello,
coming from here:
https://forum.syncthing.net/t/filesystem-watcher-errors-on-paused-folder/13467

Sadly, I couldn't reproduce it myself today, but hopefully someone here can, or find the part of code :-)

 - what happened,
When PAUSING a folder, a filesystem watcher error comes up, telling the user "folder is paused".

 - what you expected to happen instead, and
Nothing but a real error. Folder is paused is not an error IMHO.

 - any steps to reproduce the problem.
Well.... it happes from time to time, not on every try. 
Press the PAUSE button from a folder. Maybe some more often as usual. 
I'm sorry, it just happens when I don't think about it, so I cannot tell you how to click that button :-)

### Version Information

Syncthing Version: v1.4.1
OS Version: MacOSX Sierra

imsodin:
The first thing that happens when a folder is started is also starting the filesystem watching, so there will be a meaningful or no error almost immediately. There's no reason to have an artificial "watch not yet started" error. While at it I also simplified some related, tangential stuff and fixed one minor bug (comment in diff).

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -8,7 +8,6 @@ package model
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"math/rand"
 	"path/filepath"
@@ -32,8 +31,6 @@ import (
 
 // scanLimiter limits the number of concurrent scans. A limit of zero means no limit.
 var scanLimiter = newByteSemaphore(0)
-
-var errWatchNotStarted = errors.New("not started")
 
 type folder struct {
 	stateTracker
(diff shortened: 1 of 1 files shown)


lib/model/folder.go
```diff
@@ -641,7 +627,6 @@ func (f *folder) monitorWatch(ctx context.Context) {
 				if _, ok := err.(*fs.ErrWatchEventOutsideRoot); ok {
 					l.Warnln(err)
 					warnedOutside = true
-					return
```
imsodin: Returning here is clearly wrong: We still need to stop the watch, and retry (otherwise what's the point of `warnedOutside`). The bug is still minor as it only comes up when the watcher reports some event outside of the folder.

calmh:
This also means a stopped watcher has nil error, and doesn't emit a change event when watching starts (the first time, at least)?

imsodin:
Yes, a stopped watcher indeed has a nil error. A watcher is only stopped when the folder is or the watcher is disabled by config, so I think that's correct. And yes it does not emit an event if it is started without an error. Do you think it's necessary to introduce states for "running" and "not running", which could be both ignored by the UI to prevent the incorrect notification? With this PRs those states would just be `nil` errors, so all good.

PR #5846, opened 2019-07-10
Minify JavaScript and CSS files prior to building Syncthing

### LINKED ISSUE #5432: Minify Syncthing's JavaScript and CSS files
I ran my Syncthing site through [GTmetrix](https://gtmetrix.com/) earlier today and noticed that the site's loading time could greatly improved by minifying Syncthing's JavaScript and CSS. It should be fairly easy to implement this feature request because there are numerous tools out there that do it for you automatically, such as this one: https://www.minifier.org/

Here's the report from GTmetrix with all the relevant details: https://gtmetrix.com/reports/syncthing.datahoarder.xyz/csYS04j6

whalehub:
### Purpose

Fixes #5432.

### Testing

I've been running a minified build of Syncthing for the past six months without noticing any issues.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM golang:1.12 AS builder
 WORKDIR /src
 COPY . .
 
-RUN apk update && apk add nodejs nodejs-npm
+RUN apk update && apk add findutils nodejs nodejs-npm
 RUN npm install -g --unsafe-perm csso-cli uglify-js
 RUN find / -type f -name '*.js' | while read -r line; do uglifyjs "$line" -c -o "$line"; done
 RUN find / -type f -name '*.css' | while read -r line; do csso "$line" -o "$line"; done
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
We're serving this via localhost, whats the purpose of destroying readability and stacktraces for saving a few kilobytes which are not even sent over network most of the time?

Ferroin:
@AudriusButkevicius Minified JavaScript, CSS, and HTML are usually parsed marginally faster by the browser, so it can still have an impact even over the loopback interface.

That said, I personally don't think that the at most tens of milliseconds this would save over the loopback interface is worth the overhead of complicating the build and complicating debugging.

whalehub:
@AudriusButkevicius According to gtmetrix.org, minifying the Javascript and CSS files cuts down the size of the page from over 700KB+ to just over 400KB which speeds up its loading time. Also, many people use Syncthing on a VPS or dedicated server, hence the popularity of the Docker image.

AudriusButkevicius:
It's a load once and cached until the next release, accessed mostly by one user. Screwing up stacktraces, debugability, adding build complexity to save 300kb once per release is too high of a price to pay. We're not google where doing this optimization actually has impact on the world.

PR #5847, opened 2019-07-10
lib/versioner: Replace all placeholders in external command (fixes #5849)

### LINKED ISSUE #5849: Cannot concatinate external versioner template variables
See https://github.com/syncthing/syncthing/pull/5847

xduugu:
### Old title: lib/versioner: Add placeholder to provide the absolute file path to external commands

This commit adds support for a new placeholder, %FILE_PATH_FULL%, to the
command of the external versioner. The placeholder will be replaced by
the absolute path of the file that should be deleted.

### Purpose

I want syncthing to move deleted file versions to the [FreeDesktop.org Trash](https://specifications.freedesktop.org/trash-spec/trashspec-1.0.html). There is already a set of scripts, [trash-cli](https://github.com/andreafrancia/trash-cli), that can be used to interact with the trash on command line, but it requires the absolute path of the file that should be moved to trash. If the existing placeholders were used, a wrapper script would be required, because constructs like `"%FOLDER_PATH%/%FILE_PATH%"` are not supported. By using the new placeholder, no wrapper script is necessary:
`/usr/bin/trash-put -- %FILE_PATH_FULL%`

### Testing

The command `/usr/bin/trash-put -- %FILE_PATH_FULL%` was configured as external versioner command for a folder and several files were deleted from this folder on a linked device.

### Documentation

The documentation will be added as soon as the name of the placeholder is confirmed.

### DIFF
--- a/lib/versioner/external.go
+++ b/lib/versioner/external.go
@@ -10,7 +10,6 @@ import (
 	"errors"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"runtime"
 	"strings"
 	"time"
@@ -75,7 +74,6 @@ func (v External) Archive(filePath string) error {
 		"%FOLDER_FILESYSTEM%": v.filesystem.Type().String(),
 		"%FOLDER_PATH%":       v.filesystem.URI(),
 		"%FILE_PATH%":         filePath,
-		"%FILE_PATH_FULL%":    filepath.Join(v.filesystem.URI(), filePath),
 	}
 
 	for i, word := range words {
--- a/lib/versioner/external.go
+++ b/lib/versioner/external.go
@@ -78,7 +78,7 @@ func (v External) Archive(filePath string) error {
 
 	for i, word := range words {
 		for key, val := range context {
-			word = strings.ReplaceAll(word, key, val)
+			word = strings.Replace(word, key, val, -1)
 		}
 
 		words[i] = word
--- a/lib/versioner/external.go
+++ b/lib/versioner/external.go
@@ -77,9 +77,11 @@ func (v External) Archive(filePath string) error {
 	}
 
 	for i, word := range words {
-		if replacement, ok := context[word]; ok {
-			words[i] = replacement
+		for key, val := range context {
+			word = strings.ReplaceAll(word, key, val)
 		}
+
+		words[i] = word
 	}
 
 	cmd := exec.Command(words[0], words[1:]...)
--- a/lib/versioner/external.go
+++ b/lib/versioner/external.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"time"
@@ -74,6 +75,7 @@ func (v External) Archive(filePath string) error {
 		"%FOLDER_FILESYSTEM%": v.filesystem.Type().String(),
 		"%FOLDER_PATH%":       v.filesystem.URI(),
 		"%FILE_PATH%":         filePath,
+		"%FILE_PATH_FULL%":    filepath.Join(v.filesystem.URI(), filePath),
 	}
 
 	for i, word := range words {


AudriusButkevicius:
Why are the slash joined constructs not supported? I thought they were?

xduugu:
The command is split into words which are delimited by whitespace characters (https://github.com/kballard/go-shellquote/blob/master/unquote.go#L17) and each word that exists as key in the `context` map is replace by the map value. `"%FOLDER_PATH%/%FILE_PATH%"` is not a valid key, so it does not get replaced.

AudriusButkevicius:
We should change the context to do a string replace in every word, or even the command line as a whole

calmh:
Potentially useful, at least as an example (given that we've chosen different markers): https://golang.org/pkg/os/#Expand

xduugu:
I would use a regular expression replacement to not break existing commands (https://golang.org/pkg/regexp/#Regexp.ReplaceAllStringFunc). Should I re-use this pull requests or open a new one?

AudriusButkevicius:
We could migrate everyone from old format to new format once and use os.Expand

calmh:
I think a migration just to change the marker character is unnecessarily disruptive. Personally I think the bug is that `"%FOLDER_PATH%/%FILE_PATH%"` doesn't work, but I also don't think it's a problem to just do the proposed addition as-is (and *then* fix the parsing as above).

AudriusButkevicius:
So I disagree (yet you already yanked my right to object by approving).
I think we should fix handling of "%FOLDER_PATH%/%FILE_PATH%", we can literally string replace that in the raw string before we break it down into tokens.

That's much simpler, and doesn't polute the variable space with duplicate stuff (which you have to compose)

calmh:
No I specifically left the ball in your court by not merging, just indicating that I'm fine with doing so. But then we need a bug ticket on the issue and a solution to that.

This PR has the advantage that the pollution is minimal, and the user suffering from the issue already took steps to resolve it. :)

AudriusButkevicius:
I don't think this changes the API in any way (other than improving it), so why does that need a issue?

calmh:
It needs *something* as if we close this the problem otherwise just disappears in cyberspace with no record of it bothering anyone or needing solving.

calmh:
Probably would have been easier to just fix it than talk about it

AudriusButkevicius:
I think this is a better fix:
```diff
diff --git a/lib/versioner/external.go b/lib/versioner/external.go
index d2276fe0..b26bf081 100644
--- a/lib/versioner/external.go
+++ b/lib/versioner/external.go
@@ -77,9 +77,10 @@ func (v External) Archive(filePath string) error {
        }

        for i, word := range words {
-               if replacement, ok := context[word]; ok {
-                       words[i] = replacement
+               for variable, value := range context {
+                       word = strings.Replace(word, variable, value, -1)
                }
+               words[i] = word
        }

        cmd := exec.Command(words[0], words[1:]...)
```

xduugu:
This fix looks fine for #5849. I haven't recognized this bug, because I thought it is supposed to work like this. :smile:

But still I think it could be worth to add a variable for the full path, since I expect this to be the common use case. At least, I cannot think of a use case that only needs the folder path or the relative file path.

AudriusButkevicius:
Feel free to adjust your PR to do somelike like the suggested implementation does.
I want to make sure you get the credit.

The variables are separate because they were from the start, there was just arg[1] arg[2] back in the day, with the custom filesystem abstraction it made sense to allow to receive the filesystem type, hence we moved to templated command line to retrofix the old calling convention.

I don't think having two variables is wrong in any way, as the script you are calling might be "folder aware", knowing where the files are.

xduugu:
I have adjusted the pull request (and unfortunately messed it up a little bit) to only fix the bug. Apparently, I have used a function for replacement which is only available since Go 1.12. Which versions of Go are currently supported by syncthing? I could not find a list.

AudriusButkevicius:
Yeah, we try not to break backwards compatability for no reason, Debian for example does not yet have Go 1.12, so this would break their builds. Just use replace with -1.

PR #5852, opened 2019-07-12
lib: Add mtime window when comparing files

imsodin:
### Purpose

This is a product of https://forum.syncthing.net/t/ignore-permissions-different-mtimes-on-android-sdcard/13342 - I am not aware of an issue for this, please point me to it if it already exists.

This adds the folder config option "ModTimeWindowS" and uses this in the functions to compare file infos, which in turn are used when scanning or checking files on disk when pulling. When a folder is `prepare`d I use https://github.com/shirou/gopsutil to get the fs type and set the window to 2s (if not already at 2s or bigger). It's quite likely that this will fail on android, so currently it's likely on the android wrappers to set the correct window. If there's a better option to get the fs type or otherwise decide on whether to use a window or not, than using a secondary feature of this package, please let me know.

### Testing

Once the general concept is agreed on I'd like to write tests for `IsEquivalent` with the new parameter (currently I just set the window to 0 everywhere in tests).

### DIFF
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -10,6 +10,9 @@ import (
 	"errors"
 	"fmt"
 	"runtime"
+	"strings"
+
+	"github.com/shirou/gopsutil/disk"
 
 	"github.com/syncthing/syncthing/lib/fs"
 	"github.com/syncthing/syncthing/lib/protocol"
@@ -53,6 +56,7 @@ type FolderConfiguration struct {
 	WeakHashThresholdPct    int                         `xml:"weakHashThresholdPct" json:"weakHashThresholdPct"` // Use weak hash if more than X percent of the file has changed. Set to -1 to always use weak hash.
 	MarkerName              string                      `xml:"markerName" json:"markerName"`
 	CopyOwnershipFromParent bool                        `xml:"copyOwnershipFromParent" json:"copyOwnershipFromParent"`
+	ModTimeWindowS          int                         `xml:"modTimeWindowS" json:"modTimeWindowS"`
 
 	cachedFilesystem fs.Filesystem
 
@@ -234,6 +238,14 @@ func (f *FolderConfiguration) prepare() {
 
 	if f.MarkerName == "" {
 		f.MarkerName = DefaultMarkerName
+	}
+
+	if f.ModTimeWindowS < 2 && f.Path != "" {
+		if usage, err := disk.Usage(f.Filesystem().URI()); err == nil {
+			if strings.HasPrefix(strings.ToLower(usage.Fstype), "fat") {
+				f.ModTimeWindowS = 2
+			}
+		}
 	}
 }
 
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -241,10 +241,10 @@ func (f *FolderConfiguration) prepare() {
 	}
 
 	if f.ModTimeWindowS < 2 && f.Path != "" {
-		if usage, err := disk.Usage(f.Filesystem().URI()); err == nil {
-			if strings.HasPrefix(strings.ToLower(usage.Fstype), "fat") {
-				f.ModTimeWindowS = 2
-			}
+		if usage, err := disk.Usage(f.Filesystem().URI()); err != nil {
+			l.Debugln("Error determining FS type:", err)
+		} else if strings.HasPrefix(strings.ToLower(usage.Fstype), "fat") {
+			f.ModTimeWindowS = 2
 		}
 	}
 }
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -242,11 +242,15 @@ func (f *FolderConfiguration) prepare() {
 		f.MarkerName = DefaultMarkerName
 	}
 
-	if runtime.GOOS == "android" && f.RawModTimeWindowS <= 0 {
-		f.cachedModTimeWindowS = 2
-	} else if f.RawModTimeWindowS < 0 {
+	switch {
+	case f.RawModTimeWindowS == 0:
+		if runtime.GOOS == "android" {
+			f.cachedModTimeWindowS = 2
+		}
+		fallthrough
+	case f.RawModTimeWindowS < 0:
 		f.cachedModTimeWindowS = 0
-	} else {
+	case f.RawModTimeWindowS > 0:
 		f.cachedModTimeWindowS = f.RawModTimeWindowS
 	}
 }
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -244,16 +244,21 @@ func (f *FolderConfiguration) prepare() {
 	}
 
 	switch {
-	case f.RawModTimeWindowS < 0:
-		f.cachedModTimeWindowS = 0
 	case f.RawModTimeWindowS > 0:
 		f.cachedModTimeWindowS = f.RawModTimeWindowS
 	case runtime.GOOS == "android":
-		if usage, err := disk.Usage(f.Filesystem().URI()); err != nil || strings.Contains(strings.ToLower(usage.Fstype), "fat") {
+		usage, err := disk.Usage(f.Filesystem().URI())
+		if err != nil {
+			l.Debugf("Error detecting FS at %v on android, setting mtime window to 2s: %v", f.Path, err)
 			f.cachedModTimeWindowS = 2
+			break
 		}
-	default:
-		f.cachedModTimeWindowS = 0
+		if strings.Contains(strings.ToLower(usage.Fstype), "fat") {
+			l.Debugf("Detecting FS at %v on android, found %v, thus setting mtime window to 2s", f.Path, usage.Fstype)
+			f.cachedModTimeWindowS = 2
+			break
+		}
+		l.Debugf("Detecting FS at %v on android, found %v, thus leaving mtime window at 0", f.Path, usage.Fstype)
 	}
 }
 
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -246,6 +246,7 @@ func (f *FolderConfiguration) prepare() {
 	case f.RawModTimeWindowS == 0:
 		if runtime.GOOS == "android" {
 			f.cachedModTimeWindowS = 2
+			break
 		}
 		fallthrough
 	case f.RawModTimeWindowS < 0:
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/shirou/gopsutil/disk"
 
@@ -58,8 +59,8 @@ type FolderConfiguration struct {
 	CopyOwnershipFromParent bool                        `xml:"copyOwnershipFromParent" json:"copyOwnershipFromParent"`
 	RawModTimeWindowS       int                         `xml:"modTimeWindowS" json:"modTimeWindowS"`
 
-	cachedFilesystem     fs.Filesystem
-	cachedModTimeWindowS int
+	cachedFilesystem    fs.Filesystem
+	cachedModTimeWindow time.Duration
 
 	DeprecatedReadOnly       bool    `xml:"ro,attr,omitempty" json:"-"`
 	DeprecatedMinDiskFreePct float64 `xml:"minDiskFreePct,omitempty" json:"-"`
@@ -116,8 +117,8 @@ func (f FolderConfiguration) Versioner() versioner.Versioner {
 	return versionerFactory(f.ID, f.Filesystem(), f.Versioning.Params)
 }
 
-func (f FolderConfiguration) ModTimeWindowS() int {
-	return f.cachedModTimeWindowS
+func (f FolderConfiguration) ModTimeWindow() time.Duration {
+	return f.cachedModTimeWindow
 }
 
 func (f *FolderConfiguration) CreateMarker() error {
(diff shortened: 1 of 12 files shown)


lib/config/folderconfiguration.go
```diff
@@ -235,6 +239,14 @@ func (f *FolderConfiguration) prepare() {
 	if f.MarkerName == "" {
 		f.MarkerName = DefaultMarkerName
 	}
+
+	if f.ModTimeWindowS < 2 && f.Path != "" {
+		if usage, err := disk.Usage(f.Filesystem().URI()); err != nil {
```
AudriusButkevicius: I wonder if this does the right thing, as on android GOOS=android, not GOOS=linux

lib/config/folderconfiguration.go
```diff
@@ -235,6 +239,14 @@ func (f *FolderConfiguration) prepare() {
 	if f.MarkerName == "" {
 		f.MarkerName = DefaultMarkerName
 	}
+
+	if f.ModTimeWindowS < 2 && f.Path != "" {
```
AudriusButkevicius: Should this be ModTimeWindowS <= 0 ?
1s is still valid, for ext2 et al.

lib/protocol/bep_extensions.go
```diff
@@ -213,6 +213,17 @@ func (f FileInfo) isEquivalent(other FileInfo, ignorePerms bool, ignoreBlocks bo
...
 
+func ModTimeEqual(a, b time.Time, modTimeWindowS int) bool {
+	if a == b {
+		return true
+	}
+	diff := a.Sub(b)
+	if diff < 0 {
+		diff *= -1
+	}
+	return diff < time.Duration(modTimeWindowS)*time.Second
```
AudriusButkevicius: Should this be <=? Or we assume a single step change is large enough?
Not implying it should be that way, just opening up a discussion.

lib/config/folderconfiguration.go
imsodin: Then we'd need that case for ext2. Right now my intention is to force at least 2s window if we detect a fat filesystem. If the window is already >=2s, there's no need to check.

lib/config/folderconfiguration.go
imsodin: I think `GOOS == linux` is true on android, so it should at least try to do something, however as I wrote I suspect it wont work due restrictions on android, so the wrapper would still need to set this config itself.

lib/protocol/bep_extensions.go
imsodin: I'd expect it to be, i.e. I would think that anything that happens in [0,2) to get a timestamp 0 and in [2, 4) to get 2 and so on. However that's just hopeful thinking on my part, I have no clue what FAT/android caching does in reality.

lib/config/folderconfiguration.go
AudriusButkevicius: We should verify if that is the case, I suspect it will work, as it works via a simple stat call, which we already do.

lib/model/model_test.go
```diff
@@ -3303,3 +3303,55 @@ func TestConnCloseOnRestart(t *testing.T) {
 		t.Fatal("Timed out before connection was closed")
 	}
 }
+
+func TestModTimeWindow(t *testing.T) {
+	w, fcfg := tmpDefaultWrapper()
+	tfs := fcfg.Filesystem()
+	fcfg.ModTimeWindowS = 2
+	w.SetFolder(fcfg)
```
golangcibot: Error return value of `w.SetFolder` is not checked (from `errcheck`)

Catfriend1:
Feel free to ping me if you need a test APK with some setting automatically applied when the folder is created on fat32 fs. I'll take a look at Android docs on if the wrapper can easily detect fs type :-).

lib/config/folderconfiguration.go
imsodin: Ok, got that.

However I just "remembered" something else: Shouldn't we leave it at 0 regardless of FS (fat or anything else) for all OSes except android? AFAIK the problem isn't the precision, the mtimefs takes care of that, but the weird mtime caching thingy of android that resets on reboot, right?

AudriusButkevicius: I think we should check for GOOS=android, check the path of the folder, if it's in /storage/0 or something like that, and currently stored value is zero (negative would mean do not guess), we return 2 instead of zero.

AudriusButkevicius: And yes, you are right that mtime fs fixes that on other platforms, but I think it's ok to have this as a setting, in case we want to drop mtime fs at some point.

lib/model/model_test.go
```diff
@@ -3303,3 +3303,55 @@ func TestConnCloseOnRestart(t *testing.T) {
 		t.Fatal("Timed out before connection was closed")
 	}
 }
+
+func TestModTimeWindow(t *testing.T) {
+	w, fcfg := tmpDefaultWrapper()
+	tfs := fcfg.Filesystem()
+	fcfg.ModTimeWindowS() = 2
```
golangcibot: cannot assign to fcfg.ModTimeWindowS() (value of type int) (from `typecheck`)

lib/config/folderconfiguration.go
imsodin: So /storage/0 definitely isn't it. For me it's e.g. `/storage/CB88-16EA` for external SD. And internal SD (`/storage/emulated` for me) is some android hackery remount thingy, the base FS isn't FAT, no idea how it behaves though and whether the 2s windows is necessary. So I propose to just put the 2s window an android entirely - fix a dirty problem with a dirty fix.

lib/config/folderconfiguration.go
```diff
@@ -235,6 +241,14 @@ func (f *FolderConfiguration) prepare() {
 	if f.MarkerName == "" {
 		f.MarkerName = DefaultMarkerName
 	}
+
+	if runtime.GOOS == "android" && f.RawModTimeWindowS <= 0 {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

lib/config/folderconfiguration.go
```diff
@@ -235,6 +241,14 @@ func (f *FolderConfiguration) prepare() {
...
+		f.cachedModTimeWindowS = 2
+	} else if f.RawModTimeWindowS < 0 {
+		f.cachedModTimeWindowS = 0
```
AudriusButkevicius: I think this should be above android check, and we should still allow folks on android set it to -1 which means "don't touch it"

lib/config/folderconfiguration.go
Catfriend1: Hi, putting it entirely is okay for me at a first glance. It would be more exact if we call mount | grep fat (or bufferedreader string compares if grep isnt available) and then only set this Attribut forcefully through the wrapper if a folder path points to sdcard. Source : https://stackoverflow.com/questions/27227198/how-do-i-check-the-type-of-sdcards-filesystem
Even more better, we could do it in java (Environment.getExternalStorageDirectory()) and compare that to the folder path for deciding 0 or 2 secs.

lib/config/folderconfiguration.go
AudriusButkevicius: This is always zero because of the fallthrough...

AudriusButkevicius: Do we even need this branch? It will be zero by default, then android check can be folded into the switch.
I also think we should still check if its fat, as we pull in the lib and not use it now.

imsodin:
I'd like to revisit the meaning of the possible values. Audrius proposal was:

 - <0: Set to 0 unconditionally.
 - 0: Do internal magic to detect the right window (which usually is 0).
 - >0: Just use that value.

I think it's a bit weird that you need to set it to <0 to force 0, and 0 means auto. Another option would be to make it a string field with the following option:
 - empty or "auto": Do internal magic.
 - unsigned integer: Use as number of seconds.
 - parsable by time.ParseDuration: Use the value.
That's a bit more effort to parse the config, but it's more intuitive to use in my opinion.

What do you think about these or other options?

AudriusButkevicius:
I think we already have magical "negative means disabled" in a few places, so this is not new I believe.

I am somewhat against a string as this is never anything else than an integer (we could do milliseconds I guess)

imsodin:
Ok, agreed. We'll need to wait for https://github.com/shirou/gopsutil/pull/720 to be merged for CI to pass. Also we could replace github.com/calmh/du with this library (covers more OSes) if we depend on it anyway.

AudriusButkevicius:
So that seems to have landed

imsodin:
Meant to write released, not merged. At least I didn't find out how to force a commit hash instead of release with gomodules.

AudriusButkevicius:
Github.com/foo@hash I think.

Catfriend1:
Did some more tests, I discovered one more problem I'm not sure if it's with this PR or another bug:
After my steps explained above ...
- Add some new files to the folder on PC.
- Wait for them to sync to the phone. After "up2date".
- phone:  Restart Android OS.
- phone: Revert button shows up again, but only for the newly added files.
- phone: Hit revert button -> Idle, up2date, Everything ok.
- phone: Restart Android OS.
- phone: All files are still okay, NO revert button.

Catfriend1:
Took a second test phone and started the same steps over with v1.1.4 (without any patch).
After getting in sync on the phone's receiveOnly side folder, I need to revert. Then I restart the phone. Then I need to revert and again and again.

Conclusion: The PR #5852 solves a bit of the problem. I cannot test if it solves it completely because I don't know if it was correct compiling "v1.2.0+PR_5852". Maybe I missed some other commits from syncthing/master this PR relies on and that's causing the "one-time have to revert" glitch mentioned above?!?

Catfriend1:
I've made a new test APK containing the current Syncthing's master from 2019-07-21 (commit #2b622d0774549a7e37cd42f4bb5206ae96a54cda) and this PR #5859 .
Link: https://drive.google.com/open?id=1TUjA_dQRvCYKzSe70LPS7H1kDKJSMvZE

I observe the same behaviour as described above with the v1.2.0+PR5852 test APK.

Short summary:
- I get in sync on the phone.
- I reboot Android OS
image


- I see the revert button
image


- I click the revert button
image


- Then, the revert button doesn't show again.
- I reboot the phone.
- Everything still fine, no revert button.

imsodin:
Thanks for the thorough debugging, good to see it works - mostly... Could you enable model and db debug facilities on android with the scenario where you sync from remote, reboot and then get the incorrect revert button again (ideally with less test files, the log will be huge anyway).

I updated gopsutil with the `@hash` method, an actual release should happen towards the end of the month (i.e. in time for the next RC).

AudriusButkevicius:
I suspect it could be mtimes on directories.

Catfriend1:
- Chose 1 file to sync in the test folder (unzip.exe).
- Let it sync and get "up2date" on the phone.
Log until then:
[2019-07-21_19-51-39_syncthing_afterSync.zip](https://github.com/syncthing/syncthing/files/3414889/2019-07-21_19-51-39_syncthing_afterSync.zip)

- Restarted phone.
- Revert button showed up after app start for this exact one file.
image

- Hit the revert button.
- Exited app.
Log until then (beginning at app start after reboot and ending after "revert" took place and everything showed up2date)
[2019-07-21_19-54-29_syncthing_afterReboot.zip](https://github.com/syncthing/syncthing/files/3414890/2019-07-21_19-54-29_syncthing_afterReboot.zip)

imsodin:
After pulling:
```
[LJFUB] DEBUG: insert (local); folder="09m32-57cr9" File{Name:"unzip.exe", Sequence:5192, Permissions:0755, ModTime:2017-10-27 15:39:46 +0000 UTC, Version:{[{SKMUSOR 3}]}, Length:167936, Deleted:false, Invalid:false, LocalFlags:0x0, NoPermissions:false, BlockSize:131072, Blocks:[Block{0/131072/2480423031/87a5cb9d1926c9b94d9e30aab780446bf3d2df0a3d307f189712abab5db129db} Block{131072/36864/2842166102/618e7fca86658cc3cc188c8e8fd3a1af08218f28b0c6894f3dd4af185968bdd3}]}
```
2017-10-27 15:39:46
After reboot and scan:
```
[LJFUB] DEBUG: insert (local); folder="09m32-57cr9" File{Name:"unzip.exe", Sequence:5193, Permissions:0770, ModTime:2019-07-21 17:52:50 +0000 UTC, Version:{[{LJFUBFJ 1} {SKMUSOR 3}]}, Length:167936, Deleted:false, Invalid:false, LocalFlags:0x8, NoPermissions:true, BlockSize:131072, Blocks:[Block{0/131072/2480423031/87a5cb9d1926c9b94d9e30aab780446bf3d2df0a3d307f189712abab5db129db} Block{131072/36864/2842166102/618e7fca86658cc3cc188c8e8fd3a1af08218f28b0c6894f3dd4af185968bdd3}]}
```
2019-07-21 17:52:50

So something did in fact update the mtime, that's nothing like the FAT mtime window issue addressed in this PR. As it's thus unrelated to this PR, lets take it to the forum to narrow it further down.

Catfriend1:
Good find, I also wondered about that changed mtime looking at the log yesterday. I'll check if there are other apps causing this maybe also stock Android apps as this is a test device having just syncthing installed.

lib/protocol/bep_extensions.go
```diff
@@ -213,6 +213,17 @@ func (f FileInfo) isEquivalent(other FileInfo, ignorePerms bool, ignoreBlocks bo
 	return false
 }
 
+func ModTimeEqual(a, b time.Time, modTimeWindowS int) bool {
+	if a == b {
```
calmh: This should be `a.Equal(b)` to correctly handle time zones and monotonic counters and whatnot

lib/config/folderconfiguration.go
```diff
@@ -111,6 +116,10 @@ func (f FolderConfiguration) Versioner() versioner.Versioner {
 	return versionerFactory(f.ID, f.Filesystem(), f.Versioning.Params)
 }
 
+func (f FolderConfiguration) ModTimeWindowS() int {
```
calmh: I think this function should be `ModTimeWindow() time.Duration` (and hence also the call sequence all the way down to `ModTimeEqual`) instead of `...S int`. 

The actual config values are typically seconds to avoid billions in the xml file, but given this is a function anyway...

lib/config/folderconfiguration.go
```diff
@@ -233,6 +243,24 @@ func (f *FolderConfiguration) prepare() {
...
+			l.Debugf("Error detecting FS at %v on android, setting mtime window to 2s: %v", f.Path, err)
+			f.cachedModTimeWindow = 2 * time.Second
+			break
+		}
+		if strings.Contains(strings.ToLower(usage.Fstype), "fat") {
+			l.Debugf("Detecting FS at %v on android, found %v, thus setting mtime window to 2s", f.Path, usage.Fstype)
+			f.cachedModTimeWindow = 2 * time.Second
+			break
+		}
+		l.Debugf("Detecting FS at %v on android, found %v, thus leaving mtime window at 0", f.Path, usage.Fstype)
```
Catfriend1: Hmm... as I reported on the forum, usage.Fstype == "" on Android 8.0 (Moto X Play lux)
Any ideas why that happens @imsodin ? Or does the detection probably need root privileges?
Just tell me if I can help here testing another go code snippet :)

imsodin: I would have expexted that err being nil and fsttpe being "" is impossible. that's something to ask about upstream. for us it means we need to check both error and fstype not being an empty string for now.

PR #3370, opened 2016-07-01
lib/config: Remove "Invalid" attribute (fixes #2471)

calmh:
### Purpose

 - Duplicate folder IDs is now fatal during startup

 - Invalid folder flags in the ClusterConfig is fatal for the connection (this will go away soon with the proto changes, as we won't have any unknown flags any more then)

 - Empty path is a folder error reported at runtime by the regular health check things

### Testing

There are a couple of new unit tests.

Issue #2471, opened 2015-11-14
CheckFolderHealth needs to check for duplicate (invalid) folder IDs and missing Path


calmh:
We set the non-saved config attribute `Invalid` to "duplicate folder id" or "no directory configured" for those two errors when loading the config. We propagate this to the UI so the folder *looks* stopped and shows the error in question. But we never even look at the `Invalid` attribute anywhere else in the code, so the folder is in fact running as usual and doing scans and whatnot.

We need to have `Invalid` actually invalidate a folder, and do it in a way that doesn't get revived by `CheckFolderHealth`. It should be used only for nonrecoverable errors - config error, or potentially the new checksum mismatch error.

Alternatively we could just refuse to load the config when it's broken in this manner, have `CheckFolderHealth` handle the missing path case (it probably already does) and remove the `Invalid` thing altogether?

AudriusButkevicius:
We should have checkfolderhealth check the two cases when its broken

calmh:
Yeah

calmh:
Remind me why we need the "invalid" thing as opposed to just erroring the folder at runtime? Not having a path seems like the same case as the path not existing, so trivially handled by the folder health thing.

Duplicate IDs is worse. That should never ever happen so I'm tempted to just crash at config loading time.

AudriusButkevicius:
Sure, yet crashing doesn't help removing.

calmh:
Current GUI can't handle folders with duplicate IDs anyway so there is no way to remove them even if shown. We used to change the ID on one or both offending folders to allow the GUI to work with them, but that's also not correctable except by hand editing the config file so I'm not convinced of the gain.

AudriusButkevicius:
Ok lets crash then

calmh:
Yes, I think that's appropriate. As one more nail in that coffin, imagine there is a duplicate folder ID (but different paths). The index contains data. The user resolves the situation by selecting the "other" folder compared to what was in the index. Boom, all files that no longer exist in the new folder path are blown away on other devices.

PR #5854, opened 2019-07-13
lib/config: Error on empty folder path (fixes #5853)

### LINKED ISSUE #5853: Inconsistent handling of empty folder path
When creating a folder cfg, we just ignore an empty folder path and do not create a filesystem (which then leads to panics when we try to access it). On adding the folder in the model we do panic in case of an empty folder path.

What I think we should do is add a check to `Configuration.clean` for folder path being empty and throw an error in case it is (just like for ID).

https://sentry.syncthing.net/share/issue/220ffe41505240b2880ac862654283c0/
https://sentry.syncthing.net/share/issue/00728f6e8d0548a09b03124995132218/

imsodin:
### Purpose

Throw an error while cleaning/checking the config if a folder path is empty, as that's simply not valid. Currently we just ignore/mask it by not creating a filesystem and hoping nobody will ever do it, but panic reporting says it happens (#5853). 

### Testing

There's a test explicitly checking that an empty folder path is accepted and blame points at this PR/issue: https://github.com/syncthing/syncthing/issues/2471 https://github.com/syncthing/syncthing/pull/3370 Allowing the folder path to be empty seems to be a tangential change there and it talks about "folders being invalid". I don't even know what that concept was, it certainly doesn't exist anymore so I don't think this contains any reason to block this change in behaviour (disallowing empty folder path in config).

### DIFF
--- a/lib/config/config.go
+++ b/lib/config/config.go
@@ -10,6 +10,7 @@ package config
 import (
 	"encoding/json"
 	"encoding/xml"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -89,6 +90,11 @@ var (
 		"stun.voxgratia.org:3478",
 		"stun.xten.com:3478",
 	}
+)
+
+var (
+	errFolderIDEmpty   = errors.New("folder with empty ID in configuration")
+	errFolderPathEmpty = errors.New("folder with empty path in configuration")
 )
 
 func New(myID protocol.DeviceID) Configuration {
@@ -273,7 +279,11 @@ func (cfg *Configuration) clean() error {
 		folder.prepare()
 
 		if folder.ID == "" {
-			return fmt.Errorf("folder with empty ID in configuration")
+			return errFolderIDEmpty
+		}
+
+		if folder.Path == "" {
+			return errFolderPathEmpty
 		}
 
 		if _, ok := existingFolders[folder.ID]; ok {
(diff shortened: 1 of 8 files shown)


lib/config/config.go
```diff
@@ -273,7 +279,11 @@ func (cfg *Configuration) clean() error {
 		folder.prepare()
 
 		if folder.ID == "" {
-			return fmt.Errorf("folder with empty ID in configuration")
+			return errFolderIDEmpty
+		}
+
+		if folder.Path == "" {
+			return errFolderPathEmpty
```
calmh:
```suggestion
			return fmt.Errorf("folder %q: %v", folder.ID, errFolderPathEmpty)
```

lib/config/config.go
```diff
@@ -91,6 +92,11 @@ var (
 	}
 )
 
+var (
+	errFolderIDEmpty   = errors.New("folder with empty ID in configuration")
```
calmh:
```suggestion
	errFolderIDEmpty   = errors.New("folder has empty ID")
```

lib/config/config.go
```diff
@@ -91,6 +92,11 @@ var (
...
+	errFolderPathEmpty = errors.New("folder with empty path in configuration")
```
calmh:
```suggestion
	errFolderPathEmpty = errors.New("folder has empty path")
```

PR #5857, opened 2019-07-16
Add -umask option to set umask (for non-Windows platforms)

DanielO:
### Purpose

Allow setting of umask directly which simplifies startup scripts.

### Testing

Light testing on a FreeBSD server, verified new files are created with the expected permissions.

### Documentation

https://github.com/syncthing/docs/pull/460

## Authorship

based on the commit metadata.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -180,7 +180,7 @@ type RuntimeOptions struct {
 	logFlags         int
 	showHelp         bool
 	allowNewerConfig bool
-	umask		 string
+	umask            string
 }
 
 func defaultRuntimeOptions() RuntimeOptions {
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
Again, you can set this in the shell, so I think this is a needless addition

AudriusButkevicius:
It's cheap and easy to add these options, but incredibly hard to remove, so I am personally against this. Lets see what others say.

DanielO:
I found it quite difficult to work out how to get the umask set correctly in the init script due to the multiple levels of escaping required.

I don't necessarily disagree with you though but the end result is this lets me reliably set the umask which is critical to having permissions set correctly.

AudriusButkevicius:
Sure, but you could make the same argument about environment variables, and we should add -env, because of multiple levels of escaping. Multiple levels of escaping is sort of not our problem and us having to work around it seems wrong.

Again, lets see what others have to say.

calmh:
I agree that while there is nothing technically
wrong with the implementation here this is not something we want to add. The umask is better set in a small shell script or as an option in the service manager, whichever such is used. It would also be odd to add something that doesn’t make sense on Windows and is silently ignored there.

PR #5858, opened 2019-07-16
lib/syncthing: Refactor to use util.AsService

imsodin:
This is a an unrelated side-effect to working on un-globalifying events. The only difference except using `util.AsService` is that I now subscribe to events in the constructor thus the whole `WaitForStart` business is not required anymore, as the service already listens for events after construction. And Unsubscribe happens in `Stop`, not deferred on `serve`, as suture might restart the service (serve being called multiple times).

### DIFF
--- a/lib/syncthing/auditservice.go
+++ b/lib/syncthing/auditservice.go
@@ -20,16 +20,14 @@ import (
 // event per line, to the specified writer.
 type auditService struct {
 	suture.Service
-	w       io.Writer // audit destination
-	sub     *events.Subscription
-	stopped chan struct{} // signals stop complete
+	w   io.Writer // audit destination
+	sub *events.Subscription
 }
 
 func newAuditService(w io.Writer) *auditService {
 	s := &auditService{
-		w:       w,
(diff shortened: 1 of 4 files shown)


lib/syncthing/auditservice.go
```diff
@@ -10,56 +10,50 @@ import (
...
 // event per line, to the specified writer.
 type auditService struct {
-	w       io.Writer     // audit destination
-	stop    chan struct{} // signals time to stop
-	started chan struct{} // signals startup complete
+	suture.Service
+	w       io.Writer // audit destination
+	sub     *events.Subscription
 	stopped chan struct{} // signals stop complete
 }
```
AudriusButkevicius: I don't think we need the stopped anymore?

AudriusButkevicius: Or perhaps we can scrap it?

imsodin: We can. I just accepted the comment on `WaitForStop` that we need it for testing, but we don't actually (`Stop` now only returns once it actually stopped).

PR #5862, opened 2019-07-16
cmd/syncthing: Ensure myID is set by making it local (fixes #5859)

### LINKED ISSUE #5859: Duplicate local device (1.2.1-rc.1)
The local device appears as a duplicate in the GUI:

Screenshot 2019-07-16 at 22 01 13

The hostname on the second entry is "wrong" because that's what the computer in question thinks it's named at the moment. That also indicates that these are two *different* device entries, not just a display bug.

There is indeed a device entry in the config, with a *blank* device ID:

Screenshot 2019-07-16 at 22 01 46

It is *not* present in the config on disk however.

imsodin:
The global `myID` variable was never set. To prevent that from happening, I removed the global variable and pass `myID` around where necessary and call with `protocol.DeviceID{}` where the ID doesn't matter (upgrade checks and such) to make the behaviour explicit.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -60,8 +60,6 @@ const (
 	maxSystemLog         = 250
 )
 
-var myID protocol.DeviceID
-
 const (
 	usage      = "syncthing [options]"
 	extraUsage = `
@@ -333,13 +331,12 @@ func main() {
 			os.Exit(exitError)
 		}
 
-		myID = protocol.NewDeviceID(cert.Certificate[0])
-		fmt.Println(myID)
+		fmt.Println(protocol.NewDeviceID(cert.Certificate[0]))
 		return
 	}
 
 	if options.browserOnly {
-		if err := openGUI(); err != nil {
+		if err := openGUI(protocol.EmptyDeviceID); err != nil {
 			l.Warnln("Failed to open web UI:", err)
 			os.Exit(exitError)
 		}
@@ -396,8 +393,8 @@ func main() {
 	}
 }
 
-func openGUI() error {
-	cfg, err := loadOrDefaultConfig()
+func openGUI(myID protocol.DeviceID) error {
+	cfg, err := loadOrDefaultConfig(myID)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


calmh:
I did the same thing, except I spelled `protocol.DeviceID{}` as `protocol.EmptyDeviceID` as I think that's cleaner

calmh:
The other API problem has disappeared somewhere along the way, so this can be a `fixes` I think

imsodin:
I also added a check to detect and remove devices with empty ID from config to clean up affected devices.

lib/config/config_test.go
```diff
@@ -1100,6 +1100,32 @@ func TestDeviceConfigObservedNotNil(t *testing.T) {
...
+		Folders: []FolderConfiguration{
+			{
+				ID:      "foo",
+				Path:    "testdata",
+				Devices: []FolderDeviceConfiguration{{}},
+			},
+		},
+	}
+
+	cfg.clean()
```
golangcibot: Error return value of `cfg.clean` is not checked (from `errcheck`)

calmh:
Also on the list of devices per folder?

imsodin:
That already happens in `FolderConfiguration.prepare` as it checks that all folder devices actually exist. That's why I moved the device list check ahead, so that first invalid devices get filtered.

cmd/syncthing/main.go
```diff
@@ -421,31 +418,28 @@ func generate(generateDir string) error {
 		return err
 	}
 
+	var myID protocol.DeviceID
 	certFile, keyFile := filepath.Join(dir, "cert.pem"), filepath.Join(dir, "key.pem")
 	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
 	if err == nil {
 		l.Warnln("Key exists; will not overwrite.")
+		myID = protocol.NewDeviceID(cert.Certificate[0])
 		l.Infoln("Device ID:", protocol.NewDeviceID(cert.Certificate[0]))
```
AudriusButkevicius: Use myID?

AudriusButkevicius: Or actually move it out of the if

PR #5863, opened 2019-07-16
lib: Ensure timely service termination (fixes #5860)

### LINKED ISSUE #5860: Takes a long time to shut down on Ctrl-C (1.2.1-rc.1)
```
MRIW7] 22:14:47 INFO: Detected 2 NAT services
[monitor] 22:14:48 INFO: Signal 2 received; exiting
[MRIW7] 22:14:48 INFO: TCP listener (127.0.0.1:22002) shutting down
2019/07/16 22:14:58 Supervisor discover.cachingMux: discover.cachingMux: Service IPv6 local failed to terminate in a timely manner
2019/07/16 22:14:58 Supervisor local: local: Service multicastBeacon failed to terminate in a timely manner
[MRIW7] 22:14:58 INFO: connections.Service: Service c.S.listenerSupervisor failed to terminate in a timely manner
[MRIW7] 22:14:58 INFO: c.S.listenerSupervisor: Service dynamic+https://relays.syncthing.net/endpoint failed to terminate in a timely manner
[MRIW7] 22:14:58 INFO: Exiting
```

v1.2.1-rc.1

imsodin:
Now instead of various services not terminating at all or only after like forever, everything shuts down more or less immediately. Beware this is late-night-iterating between fixing something, getting another trace, fixing the next hangup, ... - some of my approaches taken are probably on the crude/heavy-handed side, but they seem to be working, so it's at least a basis to improve on.

### DIFF
--- a/lib/beacon/broadcast.go
+++ b/lib/beacon/broadcast.go
@@ -11,8 +11,9 @@ import (
 	"net"
 	"time"
 
-	"github.com/syncthing/syncthing/lib/sync"
 	"github.com/thejerf/suture"
+
+	"github.com/syncthing/syncthing/lib/util"
 )
 
 type Broadcast struct {
@@ -44,16 +45,16 @@ func NewBroadcast(port int) *Broadcast {
 	}
 
 	b.br = &broadcastReader{
-		port:    port,
-		outbox:  b.outbox,
-		connMut: sync.NewMutex(),
+		port:   port,
+		outbox: b.outbox,
 	}
+	b.br.ServiceWithError = util.AsServiceWithError(b.br.serve)
 	b.Add(b.br)
 	b.bw = &broadcastWriter{
-		port:    port,
-		inbox:   b.inbox,
-		connMut: sync.NewMutex(),
+		port:  port,
+		inbox: b.inbox,
 	}
+	b.bw.ServiceWithError = util.AsServiceWithError(b.bw.serve)
 	b.Add(b.bw)
 
 	return b
@@ -76,34 +77,34 @@ func (b *Broadcast) Error() error {
 }
 
 type broadcastWriter struct {
-	port    int
-	inbox   chan []byte
-	conn    *net.UDPConn
-	connMut sync.Mutex
-	errorHolder
+	util.ServiceWithError
+	port  int
+	inbox chan []byte
 }
 
-func (w *broadcastWriter) Serve() {
+func (w *broadcastWriter) serve(stop chan struct{}) error {
 	l.Debugln(w, "starting")
 	defer l.Debugln(w, "stopping")
 
 	conn, err := net.ListenUDP("udp4", nil)
 	if err != nil {
 		l.Debugln(err)
-		w.setError(err)
-		return
+		return err
 	}
 	defer conn.Close()
 
-	w.connMut.Lock()
-	w.conn = conn
-	w.connMut.Unlock()
+	for {
+		var bs []byte
+		select {
+		case bs = <-w.inbox:
+		case <-stop:
+			return nil
+		}
 
-	for bs := range w.inbox {
 		addrs, err := net.InterfaceAddrs()
 		if err != nil {
 			l.Debugln(err)
-			w.setError(err)
+			w.SetError(err)
 			continue
 		}
 
@@ -134,14 +135,13 @@ func (w *broadcastWriter) Serve() {
 				// Write timeouts should not happen. We treat it as a fatal
 				// error on the socket.
 				l.Debugln(err)
-				w.setError(err)
-				return
+				return err
 			}
 
 			if err != nil {
 				// Some other error that we don't expect. Debug and continue.
 				l.Debugln(err)
-				w.setError(err)
+				w.SetError(err)
 				continue
 			}
 
@@ -150,17 +150,11 @@ func (w *broadcastWriter) Serve() {
 		}
 
 		if success > 0 {
-			w.setError(nil)
+			w.SetError(nil)
 		}
 	}
-}
 
-func (w *broadcastWriter) Stop() {
-	w.connMut.Lock()
-	if w.conn != nil {
-		w.conn.Close()
-	}
-	w.connMut.Unlock()
+	return nil
 }
 
 func (w *broadcastWriter) String() string {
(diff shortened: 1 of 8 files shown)


lib/beacon/broadcast.go
```diff
@@ -150,77 +150,74 @@ func (w *broadcastWriter) Serve() {
...
 	}
-}
 
-func (w *broadcastWriter) Stop() {
-	w.connMut.Lock()
-	if w.conn != nil {
-		w.conn.Close()
-	}
-	w.connMut.Unlock()
+	return nil
```
golangcibot: unreachable: unreachable code (from `govet`)

lib/beacon/broadcast.go
```diff
@@ -150,77 +150,74 @@ func (w *broadcastWriter) Serve() {
...
 
-}
-
-func (r *broadcastReader) Stop() {
-	r.connMut.Lock()
-	if r.conn != nil {
-		r.conn.Close()
-	}
-	r.connMut.Unlock()
+	return nil
```
golangcibot: unreachable: unreachable code (from `govet`)

lib/beacon/multicast.go
```diff
@@ -142,54 +138,49 @@ func (w *multicastWriter) serve(stop chan struct{}) {
...
 		if success > 0 {
-			w.setError(nil)
-		} else {
-			l.Debugln(err)
-			w.setError(err)
+			w.SetError(nil)
 		}
 	}
+
+	return nil
```
golangcibot: unreachable: unreachable code (from `govet`)

lib/beacon/multicast.go
```diff
@@ -206,30 +197,50 @@ func (r *multicastReader) serve(stop chan struct{}) {
...
-			continue
-		}
-		l.Debugf("recv %d bytes from %s", n, addr)
+		go func() {
+			for {
+				n, _, addr, err := pconn.ReadFrom(bs)
+				if err != nil {
+					l.Debugln(err)
+					r.SetError(err)
+					select {
```
golangcibot: S1000: should use a simple channel send/receive instead of `select` with a single case (from `gosimple`)

lib/beacon/broadcast.go
```diff
@@ -150,77 +150,74 @@ func (w *broadcastWriter) Serve() {
...
-			return
-		}
-
-		r.setError(nil)
 
-		l.Debugf("recv %d bytes from %s", n, addr)
+	for {
+		go func() {
+			for {
+				n, addr, err := conn.ReadFrom(bs)
```
calmh: This is a race against the copy out of bs in the loop below.

I think it would be cleaner to not have this goroutine and break the loop by closing the socket as previously? If having a separate stop method doesn't fit in the new architecture, this could maybe be as simple as launching a `go func() { <-stop; conn.Close() }` at the beginning?

lib/nat/registry.go
```diff
@@ -19,34 +19,33 @@ func Register(provider DiscoverFunc) {
...
 	c := make(chan Device)
-	done := make(chan struct{})
-
 	for _, discoverFunc := range providers {
 		go func(f DiscoverFunc) {
 			for _, dev := range f(renewal, timeout) {
-				c <- dev
+				select {
+				case c <- dev:
+				case <-stop:
```
calmh: I'd add an explicit return to avoid having to understand how long time the `f(...)` call might take for each device while exiting

lib/relay/client/dynamic.go
```diff
@@ -147,18 +149,32 @@ type dynamicAnnouncement struct {
...
 
+	wg := sync.NewWaitGroup()
+	wg.Add(len(input))
+	results := make(chan urlWithLatency, len(input))
 	for _, relay := range input {
-		latency, err := osutil.GetLatencyForURL(relay)
-		if err != nil {
-			latency = time.Hour
-		}
+		go func(irelay string) {
```
calmh: Doing all of these at once might overwhelm the connection and skew the results. I think we are not that much in a hurry. (Although checking stop after each is nice.)

lib/nat/registry.go
AudriusButkevicius: This will leave hanging routines tho, we should push the stop down and pass in a context. The wg.Done should be moved up too.

lib/util/utils.go
```diff
@@ -187,6 +187,7 @@ func AsService(fn func(stop chan struct{})) suture.Service {
 type ServiceWithError interface {
 	suture.Service
 	Error() error
+	SetError(error)
 }
```
AudriusButkevicius: Starting to realize util.Service should have passed in a context oppose to a channel.

lib/relay/client/dynamic.go
imsodin: Reverted the goroutines, is still fast enough in my tests.

lib/nat/registry.go
imsodin: The changes were anyway wrong (my assumption of number of elements in `nats` was incorrect). I reverted everything except the stop-select part.

lib/beacon/broadcast.go
imsodin: Done

lib/beacon/multicast.go
```diff
@@ -224,12 +236,12 @@ func (r *multicastReader) serve(stop chan struct{}) {
...
 		select {
 		case r.outbox <- recv{c, addr}:
-		case <-stop:
-			return
 		default:
 			l.Debugln("dropping message")
 		}
 	}
+
+	return nil
```
golangcibot: unreachable: unreachable code (from `govet`)

lib/util/utils.go
imsodin: That would definitely be useful in places, but not happening in this PR (just due to time).

imsodin:
Just as a head up: I should have addressed all the comments, so this is ready again from my side.

lib/beacon/broadcast.go
```diff
@@ -150,77 +158,62 @@ func (w *broadcastWriter) Serve() {
...
+	}()
 
 	bs := make([]byte, 65536)
 	for {
 		n, addr, err := conn.ReadFrom(bs)
 		if err != nil {
 			l.Debugln(err)
-			r.setError(err)
-			return
+			return err
```
AudriusButkevicius: Will the error be cleared post stop? As in, if a stop causes error, what happens?

lib/beacon/multicast.go
```diff
@@ -161,35 +163,41 @@ func (w *multicastWriter) String() string {
...
+		return err
 	}
+	done := make(chan struct{})
+	defer close(done)
+	go func() {
+		select {
+		case <-stop:
+		case <-done:
+		}
+		l.Infoln("mutlicastreader close conn")
```
AudriusButkevicius: Hmm

lib/beacon/broadcast.go
```diff
@@ -150,77 +158,62 @@ func (w *broadcastWriter) Serve() {
...
-	r.connMut.Unlock()
+	done := make(chan struct{})
+	defer close(done)
+	go func() {
+		select {
+		case <-stop:
+		case <-done:
+		}
+		conn.Close()
+	}()
```
AudriusButkevicius: Just defer conn.Close?

lib/beacon/multicast.go
```diff
@@ -78,29 +78,35 @@ func (m *Multicast) Error() error {
...
 	}
+	done := make(chan struct{})
+	defer close(done)
+	go func() {
+		select {
+		case <-stop:
+		case <-done:
+		}
+		conn.Close()
+	}()
```
AudriusButkevicius: Ditto defer conn.Close()?

lib/beacon/multicast.go
AudriusButkevicius: Ditto

lib/beacon/broadcast.go
imsodin: No it isn't, meaning if there's an error set at the time it is stopped, that error remains. I don't think it matters though, does it?

AudriusButkevicius:
Other than that random log message, lgtm

lib/nat/registry.go
```diff
@@ -19,34 +19,42 @@ func Register(provider DiscoverFunc) {
...
+			case <-stop:
+				return
+			}
 		}
-		close(done)
 	}()
 
 	wg.Wait()
 	close(c)
-	<-done
```
calmh: This makes a race, we don't know that the goroutine above is done modifying `nats` before returning it

lib/util/utils.go
calmh: It's a fairly straghtforward refactor to do separately.

PR #5874, opened 2019-07-19
lib/model: Close conn on dev pause (fixes #5873)

### LINKED ISSUE #5873: Pausing device no longer works
I paused this thing way back and it just keeps on syncing. New connections are prevented but existing ones are not terminated.

Screenshot 2019-07-19 at 12 07 44

imsodin:
Not much to say: Problem happened in #5829 for no good reason. Only upside: There's now a unit test.

### DIFF
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/syncthing/syncthing/lib/config"
 	"github.com/syncthing/syncthing/lib/db"
+	"github.com/syncthing/syncthing/lib/events"
 	"github.com/syncthing/syncthing/lib/fs"
 	"github.com/syncthing/syncthing/lib/ignore"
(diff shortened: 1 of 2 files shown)


lib/model/model_test.go
```diff
@@ -3303,3 +3304,31 @@ func TestConnCloseOnRestart(t *testing.T) {
...
+	m, _, fcfg := setupModelWithConnection()
+	defer cleanupModelAndRemoveDir(m, fcfg.Filesystem().URI())
+
+	m.pmut.RLock()
+	closed := m.closed[device1]
+	m.pmut.RUnlock()
+
+	dev := m.cfg.Devices()[device1]
+	dev.Paused = true
+	m.cfg.SetDevice(dev)
```
golangcibot: Error return value of `m.cfg.SetDevice` is not checked (from `errcheck`)

Issue #5298, opened 2018-10-30
Consider a less racy config system


calmh:
We've had some bugs lately that are caused by races in config handling. As it looks today we have

- a global config "wrapper" object that is created at startup and always holds the current version of the config
- config change notifications that some things subscribe to and get old/new config values when it changes.

Some services use just the wrapper, some use just the configs from the notifications, some use both. Services can, depending on how they manage this, be looking at different versions of the config during a change event. For example, when a device is being removed it's possible that one service hasn't processed the change event yet and sees the device as existing, while another service looks directly at the wrapper and sees it as gone.

Looking at the current, global, config is inherently racy.

I think we should have something more transactional, where a config change goes through something like several phases of notifications:

- `ConfigWillChange(old, new)` - services should consider whether they are affected and if so cease processing in preparation
- `ApplyConfigChange(old, new)` - services act accordingly and apply the new config, starting up again

The guarantee we need here is that all affected services cease processing based on the old config before any service starts acting on the new. The config wrapper must ensure to order config changes properly if several happen concurrently. Noone should be looking at or retaining the global config object.

(We already have these two phases, we just call them different names and don't have the guarantees above.)

AudriusButkevicius:
There are other races.
A lot of things rely on cfg.Raw(), which is not atomic and might tramp changes of others. 
We should have a generation number in the raw, which cfg.Replace() checks that it matches the existiing version, if it does, updates the config and increments, otherwise aborts returning an error.

Alternative would be
```
cfg.Mutate(func(raw) {
  do stuff
})
```

imsodin:
While trying to harden the effect of config changes to the model along with connections, I quickly realized I was going into the direction of the problem stated in this issue. I now have a prototype of a mixture of the mechanisms described above and wanted to get some feedback before going ahead:

Basically anyone that subscribes to the config wrapper gets a `config.Changer` object, which implements all the current config modification methods (`SetFolder` and so on). That object allows for exactly one change to the config - any further changes will fail. On `CommitConfiguration` besides the old and new config object the subscribers also get a new `config.Changer` object.

And somewhat orthogonally to the above I also intend to remove all the "reading" methods on wrapper, i.e. subscribers need to store any config that they need access to.

It does nothing about ceasing operation when a config changes is announced and about to happen. And it doesn't try to order concurrent changes. I initially tried to version configs and on concurrent changes try to determine whether the changes conflict - however that's nightmarish.

calmh:
[...]

PR #5881, opened 2019-07-22
lib/model: Remove incorrect/useless panics

imsodin:
This PR removes the following two panics:
https://sentry.syncthing.net/share/issue/36dca98f1c154311b1e868aaa07e4ccc/  
https://sentry.syncthing.net/share/issue/0203181ba1674e1ba4f1d54929eee492/  

The one in folder_sendonly is simply useless, as it doesn't represent an unrecoverable/dangerous state nor does it provide useful info to find out why it happens. Without throwing a panic that item might be incorrectly shown as out-of-sync in the UI, but that's better than a panic, as the user could report that to us and we might be able to find out why it happens.

In model we previously did check the wrapper when starting a folder. That's incorrect, as starting a model is triggered by the wrapper, but at the point that we actually do it the wrapper might already be updated again. Simply ignoring the wrapper and going ahead means we proceed starting the folder and shortly after will stop it regularly again, which is totally fine.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -248,10 +248,8 @@ func (m *model) StartFolder(folder string) {
 
 // Need to hold lock on m.fmut when calling this.
 func (m *model) startFolderLocked(cfg config.FolderConfiguration) {
-	if err := m.checkFolderRunningLocked(cfg.ID); err == errFolderMissing {
-		l.Warnln("Cannot start nonexistent folder", cfg.Description())
-		panic("cannot start nonexistent folder")
-	} else if err == nil {
+	_, ok := m.folderRunners[cfg.ID]
+	if ok {
 		l.Warnln("Cannot start already running folder", cfg.Description())
 		panic("cannot start already running folder")
 	}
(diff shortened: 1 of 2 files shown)


lib/model/model.go
```diff
@@ -248,10 +248,8 @@ func (m *model) StartFolder(folder string) {
 
 // Need to hold lock on m.fmut when calling this.
 func (m *model) startFolderLocked(cfg config.FolderConfiguration) {
-	if err := m.checkFolderRunningLocked(cfg.ID); err == errFolderMissing {
-		l.Warnln("Cannot start nonexistent folder", cfg.Description())
-		panic("cannot start nonexistent folder")
-	} else if err == nil {
+	_, ok := m.folderRunners[cfg.ID]
+	if ok {
```
AudriusButkevicius: I think this will panic anyway, as we believe the folder is no longer in the config, so it's some sort of race.

imsodin: It shouldn't. We never rely on `m.cfg` throughout this function. As I wrote it is a race, but an expected/as design one (see https://github.com/syncthing/syncthing/issues/5298) as in first the folder was added/changed and before the model could follow through with that, it was already removed again in the config. That is not a problem for the model though, it will simply stop the folder once it gets notified about the secondary config change.

PR #5883, opened 2019-07-22
lib/model: Create new file-set after stopping folder (fixes #5882)

### LINKED ISSUE #5882: Folder Meta data inconsistent
https://github.com/syncthing/syncthing/blob/484fa0592e3579ed3ca1447a9cb87663c8ac49b4/lib/model/model.go#L468
```
if !to.Paused {
		// Creating the fileset can take a long time (metadata calculation)
		// so we do it outside of the lock.
		fset = db.NewFileSet(to.ID, to.Filesystem(), m.db)
	}
```

When we are scanning a folder with a large number of files, if we edit the folder property and trigger the folder restart, since the DB (maybe FileSet.Update) is being updated, the new NewFileSet here uses a new meta data object(load from db), At the same time, the unfinished GO thread writes the meta(Data lost, covered by new meta object). At this point our index exists, but the meta information is lost, which seems to be missing the file, and the file will never be scanned.

Reproduce:
1.  Add a very large number of directories
2.  In the scanning & indexing, modify the folder attributes such as the pull order.
3.  Maybe step 2 is repeated multiple times, each time a few seconds apart
4.  You can see that the number of files in the directory is inconsistent with the actual, and the lost files will never be synchronized.

imsodin:
It's fine to temporarily let go of fmut, we do it also while tearing down the folder and possibly other places. There's a restart mutex held during the entire process.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -461,18 +461,16 @@ func (m *model) RestartFolder(from, to config.FolderConfiguration) {
 		errMsg = "restarting"
 	}
 
-	var fset *db.FileSet
-	if !to.Paused {
-		// Creating the fileset can take a long time (metadata calculation)
-		// so we do it outside of the lock.
-		fset = db.NewFileSet(to.ID, to.Filesystem(), m.db)
-	}
-
 	m.fmut.Lock()
 	defer m.fmut.Unlock()
 
 	m.tearDownFolderLocked(from, fmt.Errorf("%v folder %v", errMsg, to.Description()))
 	if !to.Paused {
+		// Creating the fileset can take a long time (metadata calculation)
+		// so we do it outside of the lock.
+		m.fmut.Unlock()
+		fset := db.NewFileSet(to.ID, to.Filesystem(), m.db)
+		m.fmut.Lock()
 		m.addFolderLocked(to, fset)
 		m.startFolderLocked(to)
 	}


xjtdy888:
If there is no lock, what happens to the restartFolder call multiple times, possibly generating multiple fileset instances and addingFolderLocked, and losing the previous fileset instance?

imsodin:
> There's a restart mutex held during the entire process.

https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L447

xjtdy888:
my servers has been upgraded to the RC version, there is 600G+, I really don't want to rescan the hash. I hope the patch code can help me update the database automatically. I am not sure if it is correct.
@imsodin 
```go
func (db *schemaUpdater) updateSchema7to8() {
	t := db.newReadWriteTransaction()
	defer t.close()

	var sk []byte
	var dk []byte
	var mk []byte
	for _, folderStr := range db.ListFolders() {
		folder := []byte(folderStr)
		meta := newMetadataTracker()

		if err := meta.fromDB(db.instance, []byte(folder)); err != nil {
			continue
		}

		seqs := cuckoo.New()

		seqbuf := make([]byte, binary.MaxVarintLen64)
		db.withHave(folder, protocol.LocalDeviceID[:], nil, false, func(f FileIntf) bool {
			n := binary.PutVarint(seqbuf, f.SequenceNo())

			if !seqs.Lookup(seqbuf[:n]) {
				seqs.Insert(seqbuf[:n])
				return true
			}

			fi := f.(protocol.FileInfo)
			sk = db.keyer.GenerateSequenceKey(sk, folder, fi.Sequence)
			dk = db.keyer.GenerateDeviceFileKey(dk, folder, protocol.LocalDeviceID[:], []byte(f.FileName()))
			t.Delete(sk)

			fi.Sequence = meta.nextLocalSeq()
			sk = db.keyer.GenerateSequenceKey(sk, folder, fi.Sequence)

			t.Put(dk, mustMarshal(&fi))
			t.Put(sk, dk)

			t.checkFlush()
			return true
		})

		mk := db.keyer.GenerateFolderMetaKey(mk, folder)
		t.Delete(mk)

	}
}
```

AudriusButkevicius:
Isn't this now what the original code looked like?

imsodin:
The fmut unlock/lock around newfileset is the relevant difference to the original

AudriusButkevicius:
Wasn't it done before the fmut was acquired as provided in your link regarding the restart lock?

AudriusButkevicius:
It seems the change is already on master?

imsodin:
Yeah, but it is done before `m.tearDownFolderLocked`, which is wrong and lead to #5882. Fixing that issue is what this PR is about.

PR #5884, opened 2019-07-22
lib/model, lib/nat: More service termination speedup

imsodin:
Some more service termination speedup from what I noticed on my production system.


imsodin:
It shortens shutdown by about 2s on my system when stopping shortly after startup. Agree about context.

AudriusButkevicius:
anyways, it's approved, merge at will.

PR #5886, opened 2019-07-23
all: Remove global events.Default (ref #4085)

imsodin:
### Purpose

Most of this PR is just creating and passing around an instance of `events.Logger`. An actual change is that `cmd/syncthing` no longer has access to events, thus it queries the rest api for auto-updating. Another that the config wrapper does not emit events anymore, that needs to be done on the calling side (as a config wrapper is used in contexts without event logging happening). 

### Testing

Test setup auto-upgrades. The few things I did in the UI behaved fine.


cmd/syncthing/main.go
```diff
@@ -511,16 +512,15 @@ func performUpgrade(release upgrade.Release) {
 	}
 }
 
-func upgradeViaRest() error {
-	cfg, _ := loadOrDefaultConfig(protocol.EmptyDeviceID)
-	u, err := url.Parse(cfg.GUI().URL())
+func restGet(guiCfg config.GUIConfiguration, suffix string, timeoutS int) (*http.Response, error) {
```
golangcibot: `restGet` - `suffix` is unused (from `unparam`)

lib/connections/service.go
```diff
@@ -130,7 +131,7 @@ type service struct {
 }
 
 func NewService(cfg config.Wrapper, myID protocol.DeviceID, mdl Model, tlsCfg *tls.Config, discoverer discover.Finder,
-	bepProtocolName string, tlsDefaultCommonName string) *service {
+	bepProtocolName string, tlsDefaultCommonName string, evLogger *events.Logger) *service {
```
golangcibot: exported func NewService returns unexported type *connections.service, which can be annoying to use (from `golint`)

lib/watchaggregator/aggregator.go
```diff
@@ -125,19 +125,19 @@ func newAggregator(folderCfg config.FolderConfiguration, ctx context.Context) *a
 	return a
 }
 
-func Aggregate(in <-chan fs.Event, out chan<- []string, folderCfg config.FolderConfiguration, cfg config.Wrapper, ctx context.Context) {
+func Aggregate(in <-chan fs.Event, out chan<- []string, folderCfg config.FolderConfiguration, cfg config.Wrapper, evLogger *events.Logger, ctx context.Context) {
```
golangcibot: context.Context should be the first parameter of a function (from `golint`)

calmh:
> it queries the rest api for auto-updating

Sorry, this is bizarre. :) I'm all for avoiding globals, as far as is convenient at least, but why not construct the main event queue in cmd/syncthing and pass it downwards?

cmd/syncthing/main.go
```diff
@@ -790,7 +800,9 @@ func autoUpgrade(cfg config.Wrapper, app *syncthing.App) {
 
 		rel, err := upgrade.LatestRelease(opts.ReleasesURL, build.Version, opts.UpgradeToPreReleases)
 		if err == upgrade.ErrUpgradeUnsupported {
-			events.Default.Unsubscribe(sub)
+			if sub != nil {
```
golangcibot: nilness: impossible condition: nil != nil (from `govet`)

cmd/syncthing/main.go
```diff
@@ -814,7 +826,9 @@ func autoUpgrade(cfg config.Wrapper, app *syncthing.App) {
 			timer.Reset(checkInterval)
 			continue
 		}
-		events.Default.Unsubscribe(sub)
+		if sub != nil {
```
golangcibot: nilness: impossible condition: nil != nil (from `govet`)

imsodin:
> Sorry, this is bizarre. :) I'm all for avoiding globals, as far as is convenient at least, but why not construct the main event queue in cmd/syncthing and pass it downwards?

The idea was to keep `lib/events` "internal", i.e. a consumer of `lib/syncthing` doesn't need to deal with it. I committed an alternative, such that one still does not have to deal with events, but can get the event logger from a started `syncthing.App` to use it (which `cmd/syncthing does to avoid the rest detour).

lib/syncthing/syncthing.go
```diff
@@ -376,6 +391,17 @@ func (a *App) run() {
 	close(a.stopped)
 }
 
+func (a *App) EventLogger() (*events.Logger, error) {
```
AudriusButkevicius: GetEventLogger?
Also, feels like something that should be injected from cmd, oppose to retrievable from the lib, then app should not even be responsible for running it, which then could be injected into the autoupgrade path.

imsodin: Sure, definitely the straight forward thing to do. An argument for retrieving is that some other consumer of `lib/syncthing`, which might not care at all about events (e.g. Martchus' syncthingtray), does not have to deal with it, but is forced to when injected. Of course it isn't much that they'd need to do (define the event logger type, call yet another utility function and pass the object to the app creator). What do you think about that - should I still change to injection?

And just in case: We are pretty inconsistent with using the `Get` prefix (or not using it) - I could easily find a few examples for both. Personally I like it better without, but I don't care about it, I'll change if needed (see above).

imsodin:
I don't think there's many places where we need to create a "dummy event service". At least I'd consider the risk of forgetting to define an event logger in production code, which then "silently works", but does nothing, to be worse than having to deal with events in tests.

lib/syncthing/syncthing.go
AudriusButkevicius: You still reach out into the app to grab it for upgrades, which means either upgrades need to move into lib/syncthing, or event creation needs to move out.

Re naming, not that I think about it, it's os.Hostname() not os.GetHostname(), so probably I am just biast having written a lot of Java.

imsodin: I don't understand how using events in cmd/syncthing requires upgrades to move into lib/syncthing or event creation to move out of lib/syncthing. In my view upgrades are a thing of the released Syncthing binary, nothing the library needs to know about (well it has an exit signal for upgrading because of the upgrade rest endpoint). And one criterion was that the library should ideally not interact with the FS. I am not much worried about moving events creation out of the library, my only argument against it is convenience, I'd just like to understand the need for doing so.

calmh: For me it's a question of general principles, dependency injection / inversion of control. The purpose of lib/syncthing is not to be a creator of event systems, so it is best to leave it to the caller. That way the caller can pass an event implementation that does what they like, including one that is suitable for testing.

lib/syncthing/utils.go
```diff
@@ -124,3 +125,7 @@ func copyFile(src, dst string) error {
 func OpenGoleveldb(path string) (*db.Lowlevel, error) {
 	return db.Open(path)
 }
+
+func NewEventsLogger() *events.Logger {
```
AudriusButkevicius: Why not just events.NewLogger(), Trying to keep the facade minimal?

lib/api/api_test.go
```diff
@@ -36,10 +36,15 @@ import (
...
 var (
-	confDir = filepath.Join("testdata", "config")
-	token   = filepath.Join(confDir, "csrftokens.txt")
+	confDir      = filepath.Join("testdata", "config")
+	token        = filepath.Join(confDir, "csrftokens.txt")
+	testEvLogger = events.NewLogger()
 )
 
+func init() {
+	go testEvLogger.Serve()
```
AudriusButkevicius: An opportunity for events.NoopLogger that doesn't need a serve? Or do these tests rely on that working?

lib/discover/global_test.go
```diff
@@ -15,10 +15,17 @@ import (
...
 
+	"github.com/syncthing/syncthing/lib/events"
 	"github.com/syncthing/syncthing/lib/protocol"
 	"github.com/syncthing/syncthing/lib/tlsutil"
 )
 
+var testEvLogger = events.NewLogger()
+
+func init() {
+	go testEvLogger.Serve()
```
AudriusButkevicius: Ditto about the noop logger opportunity?

lib/model/model.go
```diff
@@ -2613,6 +2615,13 @@ func (m *model) checkDeviceFolderConnectedLocked(device protocol.DeviceID, folde
 	return nil
 }
 
+func (m *model) saveConfig() {
+	if err := m.cfg.Save(); err != nil {
```
AudriusButkevicius: However bad this sounds, my gut says we should make .Save() require a logger, or NewWrapper require a logger.
It's way too easy to just m.cfg.Save().

I think the config is boot strapped high enought that we should be able to inject the logger into the wrapper.

lib/syncthing/utils.go
imsodin: My principle was that you can do anything you need to run `syncthing.App` from `lib/syncthing`, i.e. you don't need to bind any other packages (which might e.g. not be go bindable).

lib/api/api_test.go
imsodin: As far as I see the tests don't rely on them working, but they start an api service which interacts with it. The types in events are concrete, not interfaces, so this would require new interfaces and two noop-types - I don't think it's worth the effort.

lib/events/events.go
```diff
@@ -489,3 +498,31 @@ func Error(err error) *string {
...
 	return &str
 }
+
+type noopLogger struct{}
+
+func NewNoopLogger() Logger {
+	return &noopLogger{}
+}
+
+func (_ *noopLogger) Serve() {}
```
golangcibot: receiver name should not be an underscore, omit the name if it is unused (from `golint`)

lib/events/events.go
```diff
@@ -489,3 +498,31 @@ func Error(err error) *string {
...
+
+func (_ *noopLogger) Stop() {}
```
golangcibot: receiver name should not be an underscore, omit the name if it is unused (from `golint`)

lib/events/events.go
```diff
@@ -489,3 +498,31 @@ func Error(err error) *string {
...
+
+func (_ *noopLogger) Log(t EventType, data interface{}) {}
```
golangcibot: receiver name should not be an underscore, omit the name if it is unused (from `golint`)

lib/watchaggregator/aggregator.go
```diff
@@ -125,19 +125,19 @@ func newAggregator(folderCfg config.FolderConfiguration, ctx context.Context) *a
 	return a
 }
 
-func Aggregate(in <-chan fs.Event, out chan<- []string, folderCfg config.FolderConfiguration, cfg config.Wrapper, ctx context.Context) {
+func Aggregate(in <-chan fs.Event, out chan<- []string, folderCfg config.FolderConfiguration, cfg config.Wrapper, evLogger events.Logger, ctx context.Context) {
```
golangcibot: context.Context should be the first parameter of a function (from `golint`)

lib/syncthing/utils.go
calmh: This is a tangent but I don't think we should aim for having a C-compatible interface to lib/syncthing. That forces un-Go-like choices (like passing in interfaces for dependencies). The people who want a C binding can create a small Go wrapper to do the required stuff.

imsodin: Why is passing interfaces un-Go-like? Honest question, as I thought that interfaces as function parameters are preferable to type, as it allows simpler testing by mocking (or if applicable replacing the type with a custom one).

And in this and earlier `lib/syncthing` PRs I don't remember having added any new interfaces for library purposes , just utility functions (the new event interface is due to the requested noop-logger). I believe there is something to be said for a library consumer not needing to bind multiple packages/libraries, if they don't use their functionality (events, database, config), but need the objects for the library initialisation. That's the purpose of `lib/syncthing/util.go`.

imsodin:
Any chance for this to be merged before the new RC? Just asking in case review (approval) is just on stand-by :)

lib/syncthing/utils.go
calmh: Sorry no, I mean the exact opposite. Having the constructor accept interfaces for things like the logger would be very Go-ish, but not easily consumed by something like C I think.

calmh:
Hmm I just tagged rc.1, but I'll take a look at this this evening. I haven't yet read/reviewed at all really. Sorry.

imsodin:
No worries, it's not like it's high prio (if Audrius tcp-punching PR would be a contender for your review time, I'd advocate for that :) ). It's just a largish refactor so I'd have liked to put it in the RC such that it gets tested in advance of any potential further anti-global/pro-library refactors (like assets, logs, connection registry, ...).

lib/connections/service.go
```diff
@@ -90,6 +90,7 @@ var tlsVersionNames = map[uint16]string{
 // dialers. Successful connections are handed to the model.
 type Service interface {
 	suture.Service
+	discover.AddressLister
```
calmh: Hm?

lib/syncthing/syncthing.go
```diff
@@ -119,12 +130,14 @@ func (a *App) startup() error {
 	})
 	a.mainService.ServeBackground()
 
+	a.mainService.Add(a.evLogger)
```
calmh: I'm very skeptical of this. If New() takes an event logger it should assume it's already up and running, same as any other constructor taking an event logger.

lib/syncthing/syncthing.go
```diff
@@ -376,6 +389,17 @@ func (a *App) run() {
 	close(a.stopped)
 }
 
+func (a *App) EventLogger() (events.Logger, error) {
```
calmh: As above. We took the event logger as a parameter so I don't think it's our responsibility to provide it to anyone else.

lib/syncthing/utils.go
```diff
@@ -124,3 +125,7 @@ func copyFile(src, dst string) error {
 func OpenGoleveldb(path string) (*db.Lowlevel, error) {
 	return db.Open(path)
 }
+
+func NewEventsLogger() events.Logger {
```
calmh: Not sure this adds value?

lib/events/events.go
```diff
@@ -409,12 +422,18 @@ func (s *Subscription) Poll(timeout time.Duration) (Event, error) {
...
 }
 
-func (s *Subscription) C() <-chan Event {
+func (s *subscription) C() <-chan Event {
 	return s.events
 }
 
+func (s *subscription) Unsubscribe() {
+	s.toUnsubscribe <- s
+	for range s.events {
```
calmh: This is probably fine, yet it tickles my spider sense for possible deadlocks given events are processed by the same loop that handles the unsubscribe and the unsubscribe channels is unbuffered...

lib/events/events.go
```diff
@@ -199,13 +204,20 @@ func UnmarshalEventType(s string) EventType {
...
 
-type Logger struct {
-	subs                []*Subscription
+type Logger interface {
+	suture.Service
+	Log(t EventType, data interface{})
+	Subscribe(mask EventType) Subscription
+}
+
+type evLogger struct {
```
calmh: Since we are in package events, I think this should be just `type logger`?

lib/connections/service.go
imsodin: This is entirely unrelated change due to complying with the bot. See explanation in the separate PR, to not pollute the diff: #5947

lib/syncthing/syncthing.go
imsodin: That's cruft from when the logger wasn't passed as argument into `App` - will remove.

lib/syncthing/syncthing.go
imsodin: I see, that's indeed just consequent if we expose the event logging separately, which we do. I need to mentally drop (or much de-prioritize) wanting to keep the lib/syncthing interface as small and simple as possible :)

lib/events/events.go
imsodin: That conflicts with the `logger` package. I could alternatively import that under a different name, not sure what would be appropriate.

calmh: The logger package is only referred to in debug.go so I don't think that's a problem.

(Bonus idea to seed an idea for a future PR: how about merging the logger and the event system, having only events but having some of them formatted for log output on stdout?!)

lib/events/events.go
imsodin: I was under the incorrect assumption that previously `Unsubscribe` blocked until the logger was actually done unsubscribing and wanted to not change behaviour. So for the same reason (and your spider sense) I am removing it again - thanks.

lib/syncthing/syncthing.go
```diff
@@ -44,6 +45,11 @@ const (
 	maxSystemLog         = 250
 )
 
+var (
+	errNotStarted     = errors.New("the App was not yet started")
```
golangcibot: `errNotStarted` is unused (from `deadcode`)

lib/syncthing/syncthing.go
```diff
@@ -44,6 +45,11 @@ const (
...
+	errAlreadyStopped = errors.New("the App was already stopped")
```
golangcibot: `errAlreadyStopped` is unused (from `deadcode`)

lib/syncthing/syncthing.go
```diff
@@ -8,6 +8,7 @@ package syncthing
 
 import (
 	"crypto/tls"
+	"errors"
```
golangcibot: "errors" imported but not used (from `typecheck`)

lib/events/events.go
imsodin: Go is too strict for that:

```
lib/events/debug.go:13:2: logger redeclared in this block
        previous declaration at lib/events/events.go:213:6
```

I'll use `liblogger` for the package.

(Bonus reaction: \*lalalalalala\* mustn't listen (for now (...))

lib/events/events.go
```diff
@@ -355,32 +370,30 @@ func (l *Logger) Subscribe(mask EventType) *Subscription {
...
 }
 
-func (l *Logger) Unsubscribe(s *Subscription) {
-	l.funcs <- func() {
-		dl.Debugln("unsubscribe")
-		for i, ss := range l.subs {
-			if s == ss {
-				last := len(l.subs) - 1
+func (l *logger) unsubscribe(s *subscription) {
+	dl.Debugln("unsubscribe")
```
AudriusButkevicius: Fairly useless msg?

lib/events/events.go
```diff
@@ -219,32 +231,31 @@ type Event struct {
...
 )
 
-func NewLogger() *Logger {
-	l := &Logger{
-		timeout: time.NewTimer(time.Second),
-		events:  make(chan Event, BufferSize),
-		funcs:   make(chan func()),
-		stop:    make(chan struct{}),
+func NewLogger() Logger {
+	l := &logger{
```
AudriusButkevicius: util.SutureService?

lib/events/events.go
```diff
@@ -489,3 +506,31 @@ func Error(err error) *string {
 	str := err.Error()
 	return &str
 }
+
+type noopLogger struct{}
+
+func NewNoopLogger() Logger {
```
AudriusButkevicius: Just `const NoopLogger Logger = &noopLogger{}` ? No function call needed

lib/syncthing/auditservice.go
```diff
@@ -20,14 +20,16 @@ import (
...
 
-func newAuditService(w io.Writer) *auditService {
+func newAuditService(w io.Writer, evLogger events.Logger) *auditService {
 	s := &auditService{
-		w:   w,
-		sub: events.Default.Subscribe(events.AllEvents),
+		w:        w,
+		sub:      evLogger.Subscribe(events.AllEvents),
+		evLogger: evLogger,
 	}
```
AudriusButkevicius: Where is the logger being used?

lib/syncthing/syncthing.go
```diff
@@ -100,6 +104,7 @@ func (a *App) Run() ExitStatus {
 // e.g. the API is ready for use.
 func (a *App) Start() {
 	a.startOnce.Do(func() {
+		defer close(a.started)
```
AudriusButkevicius: Who uses this?

lib/syncthing/verboseservice.go
```diff
@@ -19,12 +19,14 @@ import (
...
+	sub      events.Subscription
+	evLogger events.Logger
 }
 
-func newVerboseService() *verboseService {
+func newVerboseService(evLogger events.Logger) *verboseService {
 	s := &verboseService{
-		sub: events.Default.Subscribe(events.AllEvents),
+		sub:      evLogger.Subscribe(events.AllEvents),
+		evLogger: evLogger,
```
AudriusButkevicius: Again, who uses this?

lib/watchaggregator/aggregator_test.go
```diff
@@ -277,8 +284,15 @@ func compareBatchToExpectedDirect(t *testing.T, batch []string, expectedPaths []
...
 }
 
-func testScenario(t *testing.T, name string, testCase func(c chan<- fs.Event), expectedBatches []expectedBatch) {
+func testScenario(t *testing.T, name string, testCase func(c chan<- fs.Event), expectedBatches []expectedBatch, evLogger events.Logger) {
 	t.Helper()
+
+	if evLogger == nil {
+		evLogger = events.NewLogger()
+		go evLogger.Serve()
+		defer evLogger.Stop()
```
AudriusButkevicius: Noop logger?

lib/events/events.go
imsodin: Unfortunately `noopLogger` isn't a constant. And making it variable seems bad to me.

AudriusButkevicius: I think it's ok?  Go does that with http.DefaultHttpClient etc.

imsodin:
This PR is a mixture of two closely related, but separate changes:

 - Remove global `events.Default`
 - Introduce noop-logger and for that make `events.Logger` an interface

Would you prefer to split the two or to just follow through as is?

AudriusButkevicius:
I think its ok

lib/events/events.go
calmh: Interesting. But in that case yes, I'd import it with another name as it's just in the debug.go

lib/events/events.go
calmh: Yes. (Or `type noopLogger int` and then `const NoopLogger Logger = noopLogger(0)` but I think that's uglier. :)

PR #5892, opened 2019-07-27
cmd/stcrashreceiver: Store and serve compressed reports

calmh:
This changes the on disk format for new raw reports to be gzip
compressed. Also adds the ability to serve these reports in plain text,
to insulate web browsers from the change (previously we just served the
raw reports from disk using Caddy).

### DIFF
--- a/cmd/stcrashreceiver/stcrashreceiver.go
+++ b/cmd/stcrashreceiver/stcrashreceiver.go
@@ -54,12 +54,12 @@ func (r *crashReceiver) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	// The final path component should be a SHA256 hash in hex, so 64 hex
 	// characters. We don't care about case on the request but use lower
 	// case internally.
-	base := strings.ToLower(path.Base(req.URL.Path))
-	if len(base) != 64 {
+	reportID := strings.ToLower(path.Base(req.URL.Path))
+	if len(reportID) != 64 {
 		http.Error(w, "Bad request", http.StatusBadRequest)
 		return
 	}
-	for _, c := range base {
+	for _, c := range reportID {
 		if c >= 'a' && c <= 'f' {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


cmd/stcrashreceiver/stcrashreceiver.go
```diff
@@ -110,8 +142,14 @@ func (r *crashReceiver) servePut(base string, w http.ResponseWriter, req *http.R
 		return
 	}
 
-	// Create an output file
-	err = ioutil.WriteFile(fullPath, bs, 0644)
+	// Compress the report for storage
+	buf := new(bytes.Buffer)
+	gw := gzip.NewWriter(buf)
+	gw.Write(bs)
```
golangcibot: Error return value of `gw.Write` is not checked (from `errcheck`)

calmh: I know. It can't fail.

cmd/stcrashreceiver/stcrashreceiver.go
```diff
@@ -122,7 +160,7 @@ func (r *crashReceiver) servePut(base string, w http.ResponseWriter, req *http.R
 	if r.dsn != "" {
 		go func() {
 			// There's no need for the client to have to wait for this part.
-			if err := sendReport(r.dsn, path, bs); err != nil {
+			if err := sendReport(r.dsn, base, bs); err != nil {
```
AudriusButkevicius: I can't see it in the PR if this has an effect, but why is this changing?

cmd/stcrashreceiver/stcrashreceiver.go
```diff
@@ -78,14 +82,42 @@ func (r *crashReceiver) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+	fullPath := filepath.Join(r.dir, path)
+	if fd, err := os.Open(fullPath + ".gz"); err == nil {
+		// There is a compressed report
+		defer fd.Close()
+		gr, err := gzip.NewReader(fd)
+		if err != nil {
+			http.Error(w, "Internal server error", http.StatusInternalServerError)
+			return
+		}
+		_, _ = io.Copy(w, gr) // best effort
```
AudriusButkevicius: In theory you could just wedge content-encoding: gzip and return the zipped bytes, but yeah, this shouldn't matter much.

cmd/stcrashreceiver/stcrashreceiver.go
```diff
@@ -78,14 +82,42 @@ func (r *crashReceiver) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 	path := filepath.Join(r.dirFor(base), base)
-	if _, err := os.Lstat(path); err != nil {
-		http.Error(w, "Not found", http.StatusNotFound)
+	fullPath := filepath.Join(r.dir, path)
+	if _, err := os.Lstat(fullPath + ".gz"); err == nil {
+		// There is a compressed report
+		return
 	}
-	// 200 OK
+	if _, err := os.Lstat(fullPath); err == nil {
```
imsodin: It looks like `path` vs `fullpath` is a change in behaviour unrelated to compression. As far as I understand before just `path` was used, but `fullPath` was used to store reports, so it shouldn't have worked/worked just by accident. Now `path` doesn't seem to be used at all except for creating `fullPath`, so could be removed. I feel like I must be missing something here.

calmh: It didn't work before, unless the dir happened to be `"."` which it was when I was testing. (On the other hand it didn't matter much as it only affected the HEAD request to see if the report was already uploaded, and when uploaded it was also marked as done on the client side.)

cmd/stcrashreceiver/stcrashreceiver.go
calmh: This is because previously it used the path to construct the url like `$baseURL/aa/bb/aabbccdd...` now it just does `$baseURL/aabbccdd` without the leading directories, as that part is anyway handled magically in the HTTP server.

cmd/stcrashreceiver/stcrashreceiver.go
calmh: Yes possibly, after checking if the client supports it etc but I figured this code is almost never executed to begin with (very few viewers, likely just me) and I couldn't be bothered. :)

cmd/stcrashreceiver/stcrashreceiver.go
imsodin: Ok, thanks for the clarification. You could drop `path` here and in `serveGet` then and construct `fullPath` in a single call to `filepath.Join` to prevent anyone (me) searching for the place where `path` is used :)

calmh:
I improved path handling, and removed the stuff for plain text reports (I've gzipped the ones alread y present)

cmd/stcrashreceiver/stcrashreceiver.go
```diff
@@ -68,40 +70,58 @@ func (r *crashReceiver) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 	case http.MethodPut:
-		r.servePut(base, w, req)
+		r.servePut(reportID, fullPath, w, req)
 	default:
 		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
 	}
 }
 
+// serveGet responds to GET requests by serving the uncompressed report.
+func (r *crashReceiver) serveGet(fullPath string, w http.ResponseWriter, req *http.Request) {
```
golangcibot: `(*crashReceiver).serveGet` - `req` is unused (from `unparam`)

PR #5894, opened 2019-07-27
lib/db: Use transaction when flushing batch

calmh:
We use batches when we care about correctness, and try to get the right
write ordering in those batches. But there's nothing to say we don't
shut down (or crash) halfway through writing a batch and end up with
(for example) a FileInfo but no need index, or vice versa (I don't
remember the order but both are bad).

This should make sure than as long as we put all the relevant things in
one batch, that batch should be written atomically.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -256,8 +256,38 @@ func (b *batch) checkFlush() {
 }
 
 func (b *batch) flush() {
-	if err := b.db.Write(b.Batch, nil); err != nil && err != leveldb.ErrClosed {
+	if b.Batch.Len() == 0 {
+		// We're good here.
+		return
+	}
+
+	if b.Batch.Len() == 1 {
+		// There can be no need for a transaction for a single operation
+		if err := b.db.Write(b.Batch, nil); err != nil && err != leveldb.ErrClosed {
+			panic(err)
+		}
+		return
+	}
+
+	// Create an atomic update for the batch
+
+	tran, err := b.db.OpenTransaction()
+	if err == leveldb.ErrClosed {
+		return
+	} else if err != nil {
+		panic(err.Error())
+	}
+
+	if err := tran.Write(b.Batch, nil); err == leveldb.ErrClosed {
+		return
+	} else if err != nil {
 		panic(err)
+	}
+
+	if err := tran.Commit(); err == leveldb.ErrClosed {
+		return
+	} else if err != nil {
+		panic(err.Error())
 	}
 }
 
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -237,14 +237,36 @@ func leveldbIsCorrupted(err error) bool {
 
 type batch struct {
 	*leveldb.Batch
-	db *Lowlevel
+	w writer
+}
+
+func newBatch(w writer) *batch {
+	return &batch{
+		Batch: new(leveldb.Batch),
+		w:     w,
+	}
 }
 
 func (db *Lowlevel) newBatch() *batch {
-	return &batch{
-		Batch: new(leveldb.Batch),
-		db:    db,
-	}
+	return newBatch(db)
+}
+
+// Put implements the writeonly interface on top of the batch
+func (b *batch) Put(key []byte, val []byte, _ *opt.WriteOptions) error {
+	b.Batch.Put(key, val)
+	return nil
+}
+
+// Delete implements the writeonly interface on top of the batch
+func (b *batch) Delete(key []byte, _ *opt.WriteOptions) error {
+	b.Batch.Delete(key)
+	return nil
+}
+
+// Write implements the writeonly interface on top of the batch
+func (b *batch) Write(batch *leveldb.Batch, _ *opt.WriteOptions) error {
+	batch.Replay(b.Batch)
+	return nil
 }
 
 // checkFlush flushes and resets the batch if its size exceeds dbFlushBatch.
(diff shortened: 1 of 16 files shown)


lib/db/lowlevel.go
```diff
@@ -256,9 +256,24 @@ func (b *batch) checkFlush() {
 }
 
 func (b *batch) flush() {
-	if err := b.db.Write(b.Batch, nil); err != nil && err != leveldb.ErrClosed {
+	tran, err := b.db.OpenTransaction()
```
AudriusButkevicius: Should this return an error instead of a panic?

imsodin:
Reading this

>  Transaction is expensive and can overwhelm compaction, especially if transaction size is small. Use with caution. 

makes me a bit vary because when batch sizes last came up in a PR, their size (the decision on their size) was pretty arbitrary. I am thus unsure whether this warning applies to our frequency and size of batches. On `Write` it says that it can internally further aggregate batches, which might or might not be relevant performance wise for us.

In general I understand both batches and transactions to be atomic. How are transactions more resilient against crashes while writing?

lib/db/lowlevel.go
imsodin: Yes and no. It would definitely be nice if we actually tracked errors, but the current state is that we either panic or ignore errors all over the db package. Returning an error here would just mean panicking or ignoring one level up. Changing this would require a whole lot more changes.

calmh:
Writing a batch is just a sequence of writes, hopefully in close temporal vicinity of each other but no other guarantees as far as I can understand.

https://godoc.org/github.com/syndtr/goleveldb/leveldb#DB.Write

The transaction is atomic so either the whole thing goes through or it does not:

https://godoc.org/github.com/syndtr/goleveldb/leveldb#DB.OpenTransaction

I think our transactions can be infinitely small, but no larger than a batch... Maybe lots of small ones are more problematic than lots of small writes. But I'm also pretty sure that *not* using transactions is a correctness problem.

(And db performance is currently not great on large setups, so we need to look at compaction anyway :/)

calmh:
I'm halfway considering a rip and replace to some other database engine with some cleanup along the way, there are many many things I'd like fixed. But that's a significant amount of work (and danger) that I don't have time for. Anyone up for a one month temp gig haxing on this full time? (50% serious)

AudriusButkevicius:
Whats the goal?

AudriusButkevicius:
And no, I don't have a month of free time, just trying to understand the problem.

calmh:
That's a larger discussion, I may take it up on the forum. :) Actual transaction safety was one of the subgoals, and I think this may still be a step in that direction.

imsodin:
I wanted to quickly run the lib/db benchmarks on this and encountered problems on this branch, but not on master. I do not yet have any clue as to what's the problem, it might very well be with the benchs.

master:  
```
goos: linux
goarch: amd64
pkg: github.com/syncthing/syncthing/lib/db
   49714	     24467 ns/op	    3829 B/op	      81 allocs/op
BenchmarkUpdateOneChanged-4         	     460	   2322446 ns/op	 1109234 B/op	    1160 allocs/op
BenchmarkUpdate100Changed-4         	      21	  75467268 ns/op	36264362 B/op	   38816 allocs/op
BenchmarkUpdate100ChangedRemote-4   	      82	  13066305 ns/op	14547166 B/op	   32727 allocs/op
BenchmarkUpdateOneUnchanged-4       	   13899	     85008 ns/op	   87400 B/op	     542 allocs/op
BenchmarkNeedHalf-4                 	      13	  82102816 ns/op	80587163 B/op	  396469 allocs/op
      15	  77662195 ns/op	80896395 B/op	  397821 allocs/op
BenchmarkHave-4                     	      54	  20432381 ns/op	20461270 B/op	  132556 allocs/op
BenchmarkGlobal-4                   	      12	 106965133 ns/op	121607262 B/op	  533548 allocs/op
BenchmarkNeedHalfTruncated-4        	      48	  27744046 ns/op	19590588 B/op	   16218 allocs/op
BenchmarkHaveTruncated-4            	     300	   4025080 ns/op	  338951 B/op	    3327 allocs/op
BenchmarkGlobalTruncated-4          	      38	  40405629 ns/op	40958542 B/op	   24081 allocs/op
   11628	    109399 ns/op	   33350 B/op	     174 allocs/op
PASS
ok  	github.com/syncthing/syncthing/lib/db	26.634s
```

trans:  
```
goos: linux
goarch: amd64
pkg: github.com/syncthing/syncthing/lib/db
   51510	     23602 ns/op	    3816 B/op	      81 allocs/op
BenchmarkUpdateOneChanged-4         	     100	  32934969 ns/op	 6757039 B/op	    2397 allocs/op
BenchmarkUpdate100Changed-4         	      12	 104308806 ns/op	46925422 B/op	   45429 allocs/op
BenchmarkUpdate100ChangedRemote-4   	      69	  43654639 ns/op	31990579 B/op	   38390 allocs/op
BenchmarkUpdateOneUnchanged-4       	   13768	     91596 ns/op	   87400 B/op	     542 allocs/op
--- FAIL: BenchmarkNeedHalf
    benchmark_test.go:141: wrong length 600 != 500
      15	  75892030 ns/op	83143661 B/op	  399002 allocs/op
BenchmarkHave-4                     	      48	  21590372 ns/op	20461270 B/op	  132556 allocs/op
BenchmarkGlobal-4                   	      10	 110024536 ns/op	122154725 B/op	  533610 allocs/op
--- FAIL: BenchmarkNeedHalfTruncated
    benchmark_test.go:219: wrong length 600 != 500
BenchmarkHaveTruncated-4            	     308	   3937576 ns/op	  338951 B/op	    3327 allocs/op
BenchmarkGlobalTruncated-4          	      37	  38676087 ns/op	41669463 B/op	   24147 allocs/op
       5	 204250338 ns/op	26984136 B/op	     876 allocs/op
FAIL
exit status 1
FAIL	github.com/syncthing/syncthing/lib/db	29.734s
FAIL
```

calmh:
That's interesting, and also doesn't happen on my computer...

calmh:
I wonder if there is somehow a delay in applying the transaction and there is some bad interaction with the read snapshot of the next replace (in the two `replace(); replace` calls in the init/test). We should hold off and investigate. Maybe I'm holding the database wrong.

calmh:
Just for fun I'm going to try and make `readWriteTransaction` a wrapper around an actual database transaction and see what breaks. It'll be a somewhat larger change.

imsodin:
I can't get it reliably anymore either. It did happen more than once, but unfortunately I didn't find out how to reproduce it reliably. I'll propose a change to remove the global fileset, to make sure there's no weird interactions between tests.

calmh:
Full transaction support is a quite invasive change. We have a bunch of things that look read-only but are not (such as looking up a folder ID, which might cause the allocation of a new one) where we need to pass around a `writer` interface that can be a transaction or not, depending. And transactions can't be nested, which they are currently, so the sane way to do things will be to create transactions of the appropriate type at the top level (wherever that ends up being) and pass them downwards *all* the way.

imsodin:
It fails quite rarely, but by now I got a failure on both master and this branch, so it's unrelated. Quite weird that it failed on my first try...

Also it didn't fail on https://github.com/syncthing/syncthing/pull/5902 (yet?)

PR #5905, opened 2019-07-28
lib/fs: Use gopsutils for disk usage

imsodin:
We depend on gopsutil anyway and it adds support for a few more OSes (BSDs and solaris).


calmh:
Maybe without the crashes :)

PR #5906, opened 2019-07-28
lib/config: Handle empty Fstype for mtime-window

imsodin:
https://github.com/syncthing/syncthing/pull/5852#discussion_r307451542

Also having numerous different branches just for debug logging gets silly.

### DIFF
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -248,18 +248,12 @@ func (f *FolderConfiguration) prepare() {
 	case f.RawModTimeWindowS > 0:
 		f.cachedModTimeWindow = time.Duration(f.RawModTimeWindowS) * time.Second
 	case runtime.GOOS == "android":
-		usage, err := disk.Usage(f.Filesystem().URI())
-		if err != nil {
-			l.Debugf("Error detecting FS at %v on android, setting mtime window to 2s: %v", f.Path, err)
+		if usage, err := disk.Usage(f.Filesystem().URI()); err != nil || usage.Fstype == "" || strings.Contains(strings.ToLower(usage.Fstype), "fat") {
 			f.cachedModTimeWindow = 2 * time.Second
-			break
+			l.Debugf(`Detecting FS at %v on android: Setting mtime window to 2s: err == %v, usage.Fstype == "%v"`, f.Path, err, usage.Fstype)
+		} else {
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
I guess we can't do path based checks?

Catfriend1:
Just another idea, as I saw calmh's getenv tune PR those days:
Could we do a string.beginswith check and take the string from env var likee this:
Mtime = 2 if path beginswith (getenv(intsdpath));

Then I could get the internal storage path in the wrapper code by asking android os for the exact path on the "current" phone the app is running on and set it to the envvar "intsdpath" before launching libsyncthing.so executable. This would give us 100% accuracy and the SAF code to detect the int sd path is already there in official (and forks) java code.

AudriusButkevicius:
If you know the path is internal storage, in the wrapper you can already create the folder with the flag set in the config, I don't see much point dealing with env var magic.

AudriusButkevicius:
Also, I'd change this condition to:

`fsType.startswith("fat") || android && fsType == ""`, so that we still handle fat on other platforms.

Catfriend1:
Ok let's set modtimewindowS from the wrapper :-). I'm already drafting that.

imsodin:
doesn't mtimefs take care of fat? unless we have reason to expect similar brokenness as on android I'd just keep everything behind the android guard.

AudriusButkevicius:
Fair argument I guess.

Catfriend1:
Just for my time&test planning: When will this be expected to land in an RC/Release? 1.2.1 or 1.2.2?

calmh:
@Catfriend1 ^

Catfriend1:
Hi, 

I've just made a test build with 1.2.1-rc.5+mtime+this_pr commits and it works very well. One thing I noticed is, this lines comes up for whatever storage (internal, external) and sets mtime window to 2secs.

> 20:45:00 [start] DEBUG: Detecting FS at /storage/emulated/0/testfolder on android: Setting mtime window to 2s: err == <nil>, usage.Fstype == ""

But it's no big deal. We could set it from the wrapper if anyone needs hypergood accuracy but as that would require starting a shell with mount and parsing the Output from java I'm not doing it now - we would have needed that in two places (much overhead): config upgrade "on app start" and folder creation activity's code.

PR #5911, opened 2019-07-29
cmd/stcrashreceiver: Group issues with varying traces

imsodin:
All these "out of memory" issues are just in the way of recognizing anything relevant and multiply all the time, as their stacktraces are all over the place. I added a list of strings to check the panic messages against, and if any match, use that string as an identifier for event/issue grouping instead of the stacktrace (https://docs.sentry.io/data-management/rollups/?platform=go#custom-grouping).

### DIFF
--- a/cmd/stcrashreceiver/sentry.go
+++ b/cmd/stcrashreceiver/sentry.go
@@ -29,6 +29,14 @@ func init() {
 var (
 	clients    = make(map[string]*raven.Client)
 	clientsMut sync.Mutex
+
+	// The stacktraces accompanying these exceptions vary, thus creating
+	// tons of separate issues for the same issue. Match on these messages
+	// instead.
+	messagesToGroup = []string{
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
I guess this just like the relaypoolsrv is not autodeployed, so needs upgrading.

calmh:
Fwiw I’ve been adding filters in sentry to simply reject all these messages. But there are apparently lots of different fatal errors, and our users experience all of them...

imsodin:
I didn't find any pointer to that functionality - I don't care about stats so this filtering in sentry is good enough for me.

Another thought: Would it be worthwhile to do filtering on the receiving side, such that the traces don't get stored (disk space)?

calmh:
Yeah, it would save some space. As is they are stored uselessly for the retention period, but 🤷‍♂

imsodin:
I take that as a "who cares" ;)

PR #5912, opened 2019-07-30
all: Remove potentially problematic errors from panics (fixes #5839)

### LINKED ISSUE #5839: Prevent leaking log info into stack traces
Despite taking care to scrub log info from crash reports there are some places where it leaks. Specifically the panic line itself, which in some cases include folder labels or file names. These should be removed. Useful info can be logged before the actual panic, available for humans looking on the same machine.

Wrong:

```
panic(fmt.Sprintf("weird error on folder %v", folder))
```

Better:

```
l.Infoln("Weird: folder: %v", folder)
panic("weird error")
```

Borderline:

```
panic(err)
```

(consider what `err` might be)

imsodin:
I replaced all panics containing an error in any form, except for serialization and db stuff.

The first commit is entirely dedicated to shared puller states, where a comment claims that it needs to duplicate `inWritableDir`, but I fail to see how. They do use `fs.ModePerm` and do not change permissions if they are ignored, but the fromer is something that should also happen in all other places and the latter is in my opinion wrong: We should still try to follow "leave no trace", but not complain if it fails.

And `lib/locations` does not use logger, so I used plain `fmt.Println` to emit the error before panicking for simplicity. Is that fine or might the crash parsing choke on that?

### DIFF
--- a/cmd/syncthing/heapprof.go
+++ b/cmd/syncthing/heapprof.go
@@ -23,11 +23,15 @@ func init() {
 			rate = i
 		}
 		l.Debugln("Starting heap profiling")
-		go saveHeapProfiles(rate)
+		go func() {
+			err := saveHeapProfiles(rate) // Only returns on error
+			l.Warnln("Heap profiler failed:", err)
+			panic("Heap profiler failed")
+		}()
 	}
 }
 
-func saveHeapProfiles(rate int) {
+func saveHeapProfiles(rate int) error {
 	runtime.MemProfileRate = rate
 	var memstats, prevMemstats runtime.MemStats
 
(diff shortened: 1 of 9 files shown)


cmd/syncthing/heapprof.go
```diff
@@ -38,26 +42,27 @@ func saveHeapProfiles(rate int) {
...
-				panic(err)
+				return err
 			}
 
 			prevMemstats = memstats
 		}
 
 		time.Sleep(250 * time.Millisecond)
 	}
+	return nil
```
golangcibot: unreachable: unreachable code (from `govet`)

lib/model/folder_sendrecv.go
```diff
@@ -1078,7 +1078,7 @@ func (f *sendReceiveFolder) handleFile(file protocol.FileInfo, copyChan chan<- c
 			// Otherwise, discard the file ourselves in order for the
 			// sharedpuller not to panic when it fails to exclusively create a
 			// file which already exists
-			inWritableDir(f.fs.Remove, f.fs, tempName)
+			f.inWritableDir(f.fs.Remove, tempName)
```
golangcibot: Error return value of `f.inWritableDir` is not checked (from `errcheck`)

Issue #5915, opened 2019-07-31
error


kmfiter:
2019-07-31 12:51:11 c.S.listenerSupervisor: Failed service 'dynamic+https://relays.syncthing.net/endpoint' (2.795523 failures of 2.000000), restarting: false, error: "{dynamic+https://relays.syncthing.net/endpoint dynamic+https://relays.syncthing.net/endpoint} returned unexpectedly", stacktrace: [unknown stack trace]

AudriusButkevicius:
Please check the forum, this was answered before.

kmfiter:
2019-07-31 16:01:17 My ID: H3BOB2M-CMASA6D-S4WHVY6-UDWY5XK-MXUN5TY-QQ6U4KW-CI67Z5U-GPXBFQE
2019-07-31 16:01:18 Single thread SHA256 performance is 9.7 MB/s using crypto/sha256 (9.6 MB/s using minio/sha256-simd).
2019-07-31 16:01:20 Hashing performance is 8.82 MB/s
2019-07-31 16:01:20 Starting deadlock detector with 20m0s timeout
2019-07-31 16:01:20 Ready to synchronize "test" (lhkkf-vyefq) (sendreceive)
2019-07-31 16:01:20 Overall send rate is unlimited, receive rate is unlimited
2019-07-31 16:01:20 Using discovery server https://discovery.syncthing.net/v2/?noannounce&id=LYXKCHX-VI3NYZR-ALCJBHF-WMZYSPK-QG6QJA3-MPFYMSO-U56GTUK-NA2MIAW
2019-07-31 16:01:20 Using discovery server https://discovery-v4.syncthing.net/v2/?nolookup&id=LYXKCHX-VI3NYZR-ALCJBHF-WMZYSPK-QG6QJA3-MPFYMSO-U56GTUK-NA2MIAW
2019-07-31 16:01:20 Using discovery server https://discovery-v6.syncthing.net/v2/?nolookup&id=LYXKCHX-VI3NYZR-ALCJBHF-WMZYSPK-QG6QJA3-MPFYMSO-U56GTUK-NA2MIAW
2019-07-31 16:01:20 Anonymous usage reporting is always enabled for candidate releases.
2019-07-31 16:01:20 ...
2019-07-31 16:01:20 QUIC listener ([::]:22000) starting
2019-07-31 16:01:20 TCP listener ([::]:22000) starting
[...]

PR #5917, opened 2019-07-31
lib/connections: Log errors in relay clients

imsodin:
Currently a fatal relay listening error can go undetected (well maybe debug logging in lib/relay, but even if), and the only indication the user is presented with is the suture supervisor logging like in https://github.com/syncthing/syncthing/issues/5915:

```
2019-07-31 16:01:30 c.S.listenerSupervisor: Failed service 'dynamic+https://relays.syncthing.net/endpoint' (2.745697 failures of 2.000000), restarting: false, error: "{dynamic+https://relays.syncthing.net/endpoint dynamic+https://relays.syncthing.net/endpoint} returned unexpectedly", stacktrace: [unknown stack trace]
```

This adds logging of an error at `Info` level, and "demotes" the existing logging on relay client creation error from `Warning` to `Info` level. It is just a failing relay, not really something to notify the user with a big banner.

And I took the opportunity to do some housekeeping too.

### DIFF
--- a/lib/connections/relay_listen.go
+++ b/lib/connections/relay_listen.go
@@ -16,6 +16,7 @@ import (
 	"github.com/syncthing/syncthing/lib/dialer"
 	"github.com/syncthing/syncthing/lib/nat"
 	"github.com/syncthing/syncthing/lib/relay/client"
+	"github.com/syncthing/syncthing/lib/util"
 )
 
 func init() {
@@ -26,6 +27,7 @@ func init() {
 }
 
 type relayListener struct {
+	util.ServiceWithError
 	onAddressesChangedNotifier
 
 	uri     *url.URL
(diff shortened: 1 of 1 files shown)


lib/connections/relay_listen.go
```diff
@@ -34,30 +36,22 @@ type relayListener struct {
...
 	clnt, err := client.NewClient(t.uri, t.tlsCfg.Certificates, nil, 10*time.Second)
 	invitations := clnt.Invitations()
 	if err != nil {
-		t.mut.Lock()
-		t.err = err
-		t.mut.Unlock()
-		l.Warnln("Listen (BEP/relay):", err)
-		return
+		l.Infoln("Listen (BEP/relay):", err)
+		return err
```
AudriusButkevicius: Hmm, I suspect clnt might be null and the failure is a panic on calling invitations?

imsodin: Wow, yes that's another glaring issue in plain sight, and it still evaded me - fixed :)

PR #5919, opened 2019-08-01
Update my username and email address

carstenhag:
### Purpose
Just want to update my username and email. Not sure if this change is even needed, as it perhaps will get automatically updated?

## Authorship

based on the commit metadata.


carstenhag:
Thanks. So the mail can't be removed? I know that it won't dissappear from the git logs, but at least I want it gone on the source code.

calmh:
Well. The file exists to map commit emails to human beings and there's an automatic check that enforces this; but on the other hand you just have a readme update so I can exclude that commit from the check and remove your entry from the authors file if you prefer.

carstenhag:
That would be great. Thanks for doing this for me :)

PR #5922, opened 2019-08-02
gui: Parse strings before copying options object (fixes #5824)

### LINKED ISSUE #5823: Usage Reporting prompt displays immediately
When starting a new syncthing instance the Usage Report prompt is immediately displayed, rather than waiting for urInitialDelayS in the config file. Tested on v1.1.4.

### LINKED ISSUE #5824: Sync Protocol Listen Address in Actions behaves badly
When changing the Sync Protocol Listen Address, and clicking "Save" in the browser GUI, a prompt opens confirming if the user fishes to discard the changes. Whichever option is chosen, the changes are applied.

If  anything other than `default` the behaviour is the same, but the console also outputs errors:
```
INFO: Listen (BEP/tcp): listen tcp :22000: bind: Only one usage of each socket address (protocol/network address/port) is normally permitted.
INFO: c.S.listenerSupervisor: Failed service 'tcp://:22000' (1.000000 failures of 2.000000), restarting: true, error: "{tcp://:22000 tcp://:22000} returned unexpectedly", stacktrace: [unknown stack trace]
```

Semi-relatedly: if the specification includes an IP address or port that cannot be bound, syncthing is aware of this but it's only reported in the console. Also if the "tcp://" prefix is missing, the input is saved but syncthing fails to listen. A GUI warning for these user errors might be helpful.

All tested with v1.1.4.

imsodin:
The check if config has unsaved changes compares `tmpOptions` to `config.options`, but only the latter got the updated values from `_listenAddressesStr` into `listenAddresses`. Tested manually.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1279,6 +1279,13 @@ angular.module('syncthing.core')
                     $scope.protocolChanged = true;
                 }
 
+                // Parse strings to arrays before copying over
+                ['listenAddresses', 'globalAnnounceServers'].forEach(function (key) {
+                    $scope.tmpOptions[key] = $scope.tmpOptions["_" + key + "Str"].split(/[ ,]+/).map(function (x) {
+                        return x.trim();
+                    });
+                });
+
                 // Apply new settings locally
                 $scope.thisDeviceIn($scope.tmpDevices).name = $scope.tmpOptions.deviceName;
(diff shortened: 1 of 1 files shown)


calmh:
Does this fix the spurious "discard?" dialog when changing listen addresses? I think I filed an issue on that (or maybe I just thought I should)

calmh:
Maybe that's what you're aiming for but #5823 doesn't look like the right issue?

imsodin:
Yes, I meant to reference #5824, which is the issue you described.

PR #5936, opened 2019-08-07
Fix typo in Polish translation

megapctr:
### Purpose

There was a typo

### Testing

I don't think this change requires testing


AudriusButkevicius:
Check the read me at the root of the directory.

hacker131:
LGTM

PR #5942, opened 2019-08-11
lib/tlsutil: Remove hardcoded curve preferences (fixes #5940)

### LINKED ISSUE #5940: TLS: Support x25519, reconsider elliptic curve priority in handshake
This issue is related to a post I created in the forums some time ago: https://forum.syncthing.net/t/android-and-tls1-2/12931/38

Foreword note: This feature request only talks about curves supported in the ECDHE handshake. This is **not** related to the elliptic curve used in the ECDSA certificate.

### Background and reasoning

With TLS 1.3 right around the corner - it will become on by default in Go 1.13 - it's time to reconsider some elliptic curves in the handshake.

Syncthing currently configures Go to support 3 curves, preferring them in the following order:

`CurvePreferences: []tls.CurveID{tls.CurveP521, tls.CurveP384, tls.CurveP256}` ([here](https://github.com/syncthing/syncthing/blob/f24676ba5ab294d3bd250479718564eb404b93b6/lib/tlsutil/tlsutil.go#L92))

Starting with TLS 1.3, clients connecting to syncthing - browsers, syncthing clients - will need to _guess_ the curves supported by the server. _(This is because the ECDHE handshake is done immediatly in the ClientHello)_. Right now, **many** [modern] applications are preferring `x25519`. I've already given examples in the forum post and I can give additonal info's if requested.

For example, in my tests Firefox guessed curves `x25519` (also known as `Curve25519`) and `P-256` as backup in the TLS 1.3 `key_share`. Chrome **only** guessed `x25519` in my test setup. If the server doesn't support at least one of the guessed groups/curves, the Server will send a `HelloRetryRequest` and the client will need to try again, this time choosing a server supported curve.

If Syncthing does not support `x25519`, this will likely cause `HelloRetryRequest`'s, which lengthen the TLS handshake and are totally unnecessary/avoidable by supporting x25519.

_Supplement: I just verified that enabling TLS 1.3 in Syncthing (via GODEBUG) and then opening the web gui with Chrome definetly caused repeated  `HelloRetryRequest`, because Chrome is convinced to use x25519._

From a security point of view, `x25519` is supposedly the "best" elliptic curve (note the quotation marks), because it was openly standardized (NIST curves are based on partly unknown values) and because it's implementation is said to be less affected by timing issues or other implementation errors. 

Additionally, it's supposedly faster compared to the NIST curves, but that is probably assuming no hardware acceleration is present.

### Go support

Major question is obviously, does Go support `x25519`? Browsing the source code of Go, it seems that support for this curve in TLS was added in Go 1.8 (see [here](https://github.com/golang/go/commit/8a11cb318f5c4c15b67ffe9ff9b2fa772bf1dd85)). So if Syncthing is fine supporting only Go 1.8 and newer, adding in `x25519` should cause _no (new) issues or compatibility problems_. Clients that do not know this curve will simply fall back to one of the three NIST curves.

### Curve priority

The last point of my feature request concerns the priority of the elliptic curves, because these affect what is used in most enviroments.

This is the current default curve preference, found in the latest Go source code:

`var defaultCurvePreferences = []CurveID{X25519, CurveP256, CurveP384, CurveP521}`

Syncthing currently misses x25519 and has the other curves exactly the other way around. There's also this comment in the syncthing source code:

> // We want the longer curves at the front, because that's more
		// secure (so the web tells me, don't ask me to explain the
		// details).

I, personally, disagree with this. I support curve priority order of the Go team - x25519 and P-256 are both very secure, together widely supported curves. The other two are (mostly) useless junk that hardly adds in any security benefit. The only real difference is the bit-level-security (number of operations required for a brute-force). Assuming you can break P-256, it's not because of the 128 bit security level, but because of some other weakness that will likely affect P-384 and P-521 too (because they were likely generated using the same unknown constants).

Having x25519 the curve at the front (= the preferred) would be my favored choice, because of the reasons above. And if a 128-bit-security curve is preferred, there's no real reason to place a 256-bit-curve (P-521) right behind it - bit-level-security is not the important factor here and if it were, an attacker could just break x25519. Therefore I totally favor the curve priority order that the Go team has choosen as default.

I understand that this is only my personal, half-informed opinion, but at least reconsider the current curve priority.

### Summary (TL;DR)

1. Add support for `Curve25519/x25519`, preferably putting it at the front.
2. Reorder the priority of the other three NIST curves.

I can _probably_ do a Pull Request with the needed changes in the next few days, altough I've zero experience in Go (changes should be minimal).

### Version Information

TLS tests performed with Syncthing Version: v1.2.1 (both with TLS 1.2 and TLS 1.3)
TLS tests performed with OS Version: Windows 10 v1903
TLS tests performed with Browser Version: Firefox 68.0.1, Chrome 76.0.3809.100

calmh:
They are arguable outdated and we are better off trusting the standard
library than trying to keep up with it ourselves.

### DIFF
--- a/lib/tlsutil/tlsutil.go
+++ b/lib/tlsutil/tlsutil.go
@@ -86,10 +86,6 @@ func SecureDefault() *tls.Config {
 	return &tls.Config{
 		// TLS 1.2 is the minimum we accept
 		MinVersion: tls.VersionTLS12,
-		// We want the longer curves at the front, because that's more
-		// secure (so the web tells me, don't ask me to explain the
-		// details).
-		CurvePreferences: []tls.CurveID{tls.CurveP521, tls.CurveP384, tls.CurveP256},
 		// The cipher suite lists built above. These are ignored in TLS 1.3.
 		CipherSuites: cs,
 		// We've put some thought into this choice and would like it to


GermanCoding:
Wonderful solution, thanks! Will verify it soon but I guess this should solve the compatibility issues with Chrome's TLS 1.3.

PR #5950, opened 2019-08-13
lib/discover: Don't leak goroutines

imsodin:
Run methods as services instead of just kicking off a goroutine that will never stop.

### DIFF
--- a/lib/beacon/beacon.go
+++ b/lib/beacon/beacon.go
@@ -7,9 +7,13 @@
 package beacon
 
 import (
+	"fmt"
 	"net"
+	"time"
 
 	"github.com/thejerf/suture"
+
+	"github.com/syncthing/syncthing/lib/util"
 )
 
 type recv struct {
@@ -19,7 +23,84 @@ type recv struct {
 
 type Interface interface {
 	suture.Service
+	fmt.Stringer
 	Send(data []byte)
 	Recv() ([]byte, net.Addr)
 	Error() error
+}
+
+type cast struct {
+	*suture.Supervisor
+	name    string
+	reader  util.ServiceWithError
+	writer  util.ServiceWithError
+	outbox  chan recv
+	inbox   chan []byte
+	stopped chan struct{}
+}
+
+// newCast creates a base object for multi- or broadcasting. Afterwards the
+// caller needs to set reader and writer with the addReader and addWriter
+// methods to get a functional implementation of Interface.
+func newCast(name string) *cast {
+	return &cast{
+		Supervisor: suture.New(name, suture.Spec{
+			// Don't retry too frenetically: an error to open a socket or
+			// whatever is usually something that is either permanent or takes
+			// a while to get solved...
+			FailureThreshold: 2,
+			FailureBackoff:   60 * time.Second,
+			// Only log restarts in debug mode.
+			Log: func(line string) {
+				l.Debugln(line)
+			},
+			PassThroughPanics: true,
+		}),
+		name:    name,
+		inbox:   make(chan []byte),
+		outbox:  make(chan recv, 16),
+		stopped: make(chan struct{}),
+	}
+}
+
+func (c *cast) addReader(r util.ServiceWithError) {
+	c.reader = r
+	c.Add(r)
+}
+
+func (c *cast) addWriter(w util.ServiceWithError) {
+	c.writer = w
+	c.Add(w)
+}
+
+func (c *cast) Stop() {
+	c.Supervisor.Stop()
+	close(c.stopped)
+}
+
+func (c *cast) String() string {
+	return fmt.Sprintf("%s@%p", c.name, c)
+}
+
+func (c *cast) Send(data []byte) {
+	select {
+	case c.inbox <- data:
+	case <-c.stopped:
+	}
+}
+
+func (c *cast) Recv() ([]byte, net.Addr) {
+	select {
+	case recv := <-c.outbox:
+		return recv.data, recv.src
+	case <-c.stopped:
+	}
+	return nil, nil
+}
+
+func (c *cast) Error() error {
+	if err := c.reader.Error(); err != nil {
+		return err
+	}
+	return c.writer.Error()
 }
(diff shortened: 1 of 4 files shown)


lib/discover/local.go
```diff
@@ -154,13 +145,22 @@ func (c *localClient) sendLocalAnnouncements() {
...
+func (c *localClient) recvAnnouncements(stop chan struct{}) {
+	b := c.beacon
 	warnedAbout := make(map[string]bool)
 	for {
+		select {
+		case <-stop:
+			return
+		default:
+		}
+
```
AudriusButkevicius: I think this will still leak as Recv blocks forever

imsodin: Not anymore

lib/beacon/beacon.go
```diff
@@ -19,7 +23,84 @@ type recv struct {
...
+	*suture.Supervisor
+	name    string
+	reader  util.ServiceWithError
+	writer  util.ServiceWithError
+	outbox  chan recv
+	inbox   chan []byte
+	stopped chan struct{}
+}
+
+// newCast creates a base object for multi- or braodcasting. Afterwards the
```
golangcibot: `braodcasting` is a misspelling of `broadcasting` (from `misspell`)
```suggestion
// newCast creates a base object for multi- or broadcasting. Afterwards the
```

lib/discover/local.go
calmh: I didn't follow all the breadcrumbs but stopping the service should at some point close the socket which unblocks Recv and results in shutdown, I hope?

lib/beacon/broadcast.go
```diff
@@ -7,84 +7,26 @@
...
-	b.inbox <- data
+func NewBroadcast(port int) Interface {
+	c := newCast("broadcastBeacon")
+	c.addReader(util.AsServiceWithError(func(stop chan struct{}) error {
+		return writeBroadcasts(c.inbox, port, stop)
+	}))
+	c.addWriter(util.AsServiceWithError(func(stop chan struct{}) error {
+		return readBroadcasts(c.outbox, port, stop)
+	}))
+	return c
```
AudriusButkevicius: This could just take a function, and the addreader/writer wraps it in a service and passes in the right channel, and potentially a port (newCast could take a port)

lib/beacon/broadcast.go
```diff
@@ -149,35 +90,25 @@ func (w *broadcastWriter) serve(stop chan struct{}) error {
 			if err != nil {
 				// Some other error that we don't expect. Debug and continue.
 				l.Debugln(err)
-				w.SetError(err)
```
AudriusButkevicius: This changes behaviour?

AudriusButkevicius: Or I guess this compliments the success condition change below.

lib/beacon/multicast.go
```diff
@@ -152,33 +95,23 @@ func (w *multicastWriter) serve(stop chan struct{}) error {
...
 		}
 	}
 }
 
-func (w *multicastWriter) String() string {
-	return fmt.Sprintf("multicastWriter@%p", w)
-}
+func readMulticasts(outbox chan<- recv, addr string, stop chan struct{}) error {
+	l.Debugln("starting readMulticasts")
+	defer l.Debugln("stopping readMulticasts")
```
AudriusButkevicius: The cast could log this cruft I guess? It has the name

lib/discover/local.go
AudriusButkevicius: Recv reads of a channel, so connection closing won't do much.

lib/beacon/broadcast.go
imsodin: Exactly: Before we set a transient error for every loop iteration. However we didn't do it consistently: If a new error comes up, that gets set, but if something succeeds, the error doesn't get set. I don't see the value of exposing this intermediate errors. Now we just expose an error if none of the tries succeed (inverse condition to removing the error again previously).

lib/discover/local.go
imsodin: Recv now unblocks when stopped: https://github.com/syncthing/syncthing/pull/5950/files/f66c15d355b2b55f60b706e397839172c18c232b#diff-9564247335505ac56ab73a0d2b2310a9R92

lib/beacon/broadcast.go
imsodin: I did not unify this part as the function signatures vary a lot (different channels and port integer vs addr string) - passing around `util.ServiceWithErrors` avoids those problems.

lib/beacon/multicast.go
imsodin: Not really, as it's just the supervisor, it doesn't invoke these methods itself.

lib/beacon/beacon.go
```diff
@@ -19,7 +23,84 @@ type recv struct {
...
+}
+
+func (c *cast) addReader(r util.ServiceWithError) {
+	c.reader = r
+	c.Add(r)
+}
+
+func (c *cast) addWriter(w util.ServiceWithError) {
+	c.writer = w
+	c.Add(w)
```
AudriusButkevicius: This could just be:
```
func (c *cast) addWriter(svc func(stop chan struct{}) error) {
    c.writer = util.ServiceWithError(func(stop chan struct) error {
         l.Debug("Starting " + c.name + " writer")
         err := svc(stop)
         l.Debug("Stopped " + c.name + " writer", err)
         return err
    })
    c.Add(c.writer)
}
```

then on both services you just do:
```
        c.addWriter(func(stop chan struct{}) error {
		return readBroadcasts(c.outbox, port, stop)
	})
```


Also, note that reader and writer are the other way around?

lib/beacon/broadcast.go
```diff
@@ -7,84 +7,26 @@
...
-}
-
-func (b *Broadcast) Send(data []byte) {
-	b.inbox <- data
+func NewBroadcast(port int) Interface {
+	c := newCast("broadcastBeacon")
+	c.addReader(util.AsServiceWithError(func(stop chan struct{}) error {
+		return writeBroadcasts(c.inbox, port, stop)
+	}))
+	c.addWriter(util.AsServiceWithError(func(stop chan struct{}) error {
```
AudriusButkevicius: Side note, not related.
AsServiceWithError could use one stackframe above as the service name.

imsodin:
Could I get a re-approval on this please?

PR #5954, opened 2019-08-15
lib/model: Optimise locking around conn-close and puller states

imsodin:
There is no deadlock here, but if a write operation takes really long, that can block `model.pmut` (see https://forum.syncthing.net/t/new-panic-v1-2-2-rc2/13647/2). This PR breaks this "lock-linking" in two places:

1. Separate locks for `sharedPullerState` and the contained file-descriptor. Thus really long writes do not block simple, but often used methods like `.Updated` or `Progress`.

2. Release `pmut` before calling `progressemitter.temporaryIndexUnsubscribe`.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1417,12 +1417,11 @@ func (m *model) Closed(conn protocol.Connection, err error) {
 	device := conn.ID()
 
 	m.pmut.Lock()
-	defer m.pmut.Unlock()
 	conn, ok := m.conn[device]
 	if !ok {
+		m.pmut.Unlock()
 		return
 	}
-	m.progressEmitter.temporaryIndexUnsubscribe(conn)
 	delete(m.conn, device)
 	delete(m.connRequestLimiters, device)
 	delete(m.helloMessages, device)
@@ -1430,6 +1429,9 @@ func (m *model) Closed(conn protocol.Connection, err error) {
 	delete(m.remotePausedFolders, device)
 	closed := m.closed[device]
 	delete(m.closed, device)
+	m.pmut.Unlock()
+
+	m.progressEmitter.temporaryIndexUnsubscribe(conn)
 
 	l.Infof("Connection to %s at %s closed: %v", device, conn.Name(), err)
 	events.Default.Log(events.DeviceDisconnected, map[string]string{
--- a/lib/model/sharedpullerstate.go
+++ b/lib/model/sharedpullerstate.go
@@ -64,7 +64,6 @@ type pullerProgress struct {
 
 // lockedWriterAt adds a lock to protect from closing the fd at the same time as writing.
 // WriteAt() is goroutine safe by itself, but not against for example Close().
-// I.e. one must acquire the (write) lock before closing fd.
 type lockedWriterAt struct {
 	mut sync.RWMutex
 	fd  fs.File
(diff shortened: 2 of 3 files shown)


lib/model/model.go
```diff
@@ -1417,19 +1417,21 @@ func (m *model) Closed(conn protocol.Connection, err error) {
...
 	delete(m.conn, device)
 	delete(m.connRequestLimiters, device)
 	delete(m.helloMessages, device)
 	delete(m.deviceDownloads, device)
 	delete(m.remotePausedFolders, device)
 	closed := m.closed[device]
 	delete(m.closed, device)
+	m.pmut.Unlock()
+
+	m.progressEmitter.temporaryIndexUnsubscribe(conn)
```
AudriusButkevicius: Doesn't this reach into the model to do its thing?

imsodin: No.

lib/model/sharedpullerstate.go
```diff
@@ -62,17 +62,17 @@ type pullerProgress struct {
...
+	mut sync.Mutex
+	fd  fs.File
 }
 
 func (w lockedWriterAt) WriteAt(p []byte, off int64) (n int, err error) {
-	(*w.mut).Lock()
-	defer (*w.mut).Unlock()
-	return w.wr.WriteAt(p, off)
+	w.mut.Lock()
+	defer w.mut.Unlock()
```
AudriusButkevicius: This is a value receiver, this locking does fuck all

lib/model/sharedpullerstate.go
```diff
@@ -265,18 +265,20 @@ func (s *sharedPullerState) finalClose() (bool, error) {
...
 
-		if err := s.fd.Close(); err != nil && s.err == nil {
+		if err := s.writer.fd.Close(); err != nil && s.err == nil {
 			// This is our error as we weren't errored before.
 			s.err = err
 		}
-		s.fd = nil
+		s.writer.mut.Unlock()
+		s.writer = nil
 	}
```
AudriusButkevicius: Why are we changing shit here? The only reason locking exists on the write is to prevent interleaved writes, otherwise no locking is needed?

AudriusButkevicius:
The close happens in the finisher, once it's in the finisher, no writes should happen?

calmh:
That's my feeling as well so I'm not sure what we're protecting against (some sort of situation when we fail a pull?). But we should either do it correctly or not at all. 

90de5659ea8ad3241315a171c6bd2b05c40019ff

imsodin:
I think I know what it protects against: The finisher gets invoked for every pulled block and goes through if either there is nothing to do left, or an error occurred. However an error having occurred in one block-pulling does not guarantee that no other block-pulling is still going on. Does that make sense?

calmh:
It does as described, I was too lazy to see if this is what actually happens in the code :)

calmh:
For that matter it could be an RWMutex with the WriteAt being the "R" part since those can be concurrent with each other, just not with Close.

lib/model/sharedpullerstate.go
```diff
@@ -266,6 +267,8 @@ func (s *sharedPullerState) finalClose() (bool, error) {
...
 
 	if s.writer != nil {
+		// The lock must be acquired before closing the fd.
+		s.writer.mut.Lock()
```
calmh: `lockedWriterAt.SyncClose() error` ?

imsodin:
So I went back to acquiring the lock when closing, keeping a separate lock for writing/fd, as it's completely unrelated to protecting the other stats in the puller state and could still lead to locking if writing takes ages (which isn't our fault, but still can become our problem). And I changed the lock to RW as suggested for concurrent writing (which probably doesn't change anything as IO is still synchronous?).

lib/model/sharedpullerstate.go
```diff
@@ -62,17 +62,28 @@ type pullerProgress struct {
...
+	fd  fs.File
 }
 
-func (w lockedWriterAt) WriteAt(p []byte, off int64) (n int, err error) {
-	(*w.mut).Lock()
-	defer (*w.mut).Unlock()
-	return w.wr.WriteAt(p, off)
+func (w *lockedWriterAt) WriteAt(p []byte, off int64) (n int, err error) {
+	w.mut.RLock()
+	defer w.mut.RUnlock()
```
AudriusButkevicius: defo deserves a comment

imsodin: Are these helpful?

PR #5958, opened 2019-08-15
Playground: GitHub CI

calmh:
This is just to see if GitHub picks this up in a branch and does something interesting with it or not

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,20 @@
+name: Go
+on: [push]
+jobs:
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Surprisingly this seems to have worked ok?

calmh:
It did, yes. I haven't thought through the implications but being able to do tests & CI on windows & mac:s managed by others isn't unattractive. The downside is that the windows side runs Go 1.12.4 at the moment...

PR #5960, opened 2019-08-16
cmd/stupgrades: Generate appropriate upgrade data (fixes #5924)

### LINKED ISSUE #5924: Upgrade server JSON generator
Currently the JSON served from upgrade.syncthing.net is simply the GitHub API response cached locally. This has disadvantages: it includes a shit-ton of attributes and is very large, hence we limit it to the last five releases. The last five releases are currently all RC:s, so old stable versions currently can't upgrade to anything newer stable.

We should have a tool that grabs the API data from GitHub, filters it through our structs (to remove attributes we don't understand and hence don't need to retain) and perhaps whittles it down to the last RC and the last stable. This should be a smaller response that is valid in more cases.

The end result is a JSON file that we can serve to clients (with gzip compression).

calmh:
This is a tiny tool to grab the GitHub releases info and generate a
more concise version of it. The conciseness comes from two aspects:

- We select only the latest stable and pre. There is no need to offer
  upgrades to versions that are older than the latest. (There might be, in
  the future, when we hit 2.0. We can revisit this at that time.)

- We use our structs to deserialize and reserialize the data. This means
  we remove all attributes that we don't understand and hence don't
  require.

All in all the new response is about 10% the size of the previous one and
avoids the issue where we only serve a bunch of release candidates and
no stable.

### DIFF
--- a/cmd/stupgrades/main.go
+++ b/cmd/stupgrades/main.go
@@ -30,7 +30,9 @@ func main() {
 	sort.Sort(upgrade.SortByRelease(rels))
 	rels = filterForLatest(rels)
 
-	json.NewEncoder(os.Stdout).Encode(rels)
+	if err := json.NewEncoder(os.Stdout).Encode(rels); err != nil {
+		os.Exit(1)
+	}
 }
 
 // filterForLatest returns the latest stable and prerelease only. If the
--- a/cmd/stupgrades/main.go
+++ b/cmd/stupgrades/main.go
@@ -1,0 +1,53 @@
+// Copyright (C) 2019 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"os"
+	"sort"
+
+	"github.com/syncthing/syncthing/lib/upgrade"
+)
+
+const defaultURL = "https://api.github.com/repos/syncthing/syncthing/releases?per_page=25"
+
+func main() {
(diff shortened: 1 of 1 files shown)


cmd/stupgrades/main.go
```diff
@@ -0,0 +1,53 @@
...
+	rels := upgrade.FetchLatestReleases(*url, "")
+	if rels == nil {
+		// An error was already logged
+		os.Exit(1)
+	}
+
+	sort.Sort(upgrade.SortByRelease(rels))
+	rels = filterForLatest(rels)
+
+	json.NewEncoder(os.Stdout).Encode(rels)
```
golangcibot: Error return value of `(*encoding/json.Encoder).Encode` is not checked (from `errcheck`)

AudriusButkevicius:
How exactly will this be used? Web server will invoke binary for outout?

cmd/stupgrades/main.go
```diff
@@ -0,0 +1,55 @@
...
```
golangcibot: Error return value of `(*encoding/json.Encoder).Encode` is not checked (from `errcheck`)

calmh:
This thing runs quarterly in cron on the server:

```diff
diff --git a/roles/www/files/fetch-releases.sh b/roles/www/files/fetch-releases.sh
index f9c8eb4..43b6b24 100755
--- a/roles/www/files/fetch-releases.sh
+++ b/roles/www/files/fetch-releases.sh
@@ -4,7 +4,7 @@ set -euo pipefail
 pushd /var/www/upgrades.syncthing.net
 rm -f meta.json.new

-curl --fail -s -o meta.json.new https://api.github.com/repos/syncthing/syncthing/releases\?per_page=5
+/opt/local/bin/stupgrades > meta.json.new
 mv meta.json.new meta.json

 gzip -c < meta.json > meta.json.gz.new
```

PR #5961, opened 2019-08-16
connection logging: replace "connect to" with "connect with", to better reflect reality

marco-m:
### Purpose

This PR attemps to reduce misleading interpretation of logging messages of the type "connected to X", while the same code path could also refer to a "connection from X".

Following this [forum discussion](https://forum.syncthing.net/t/were-is-a-spurious-node-rjr2g2a-coming-from/13649/6), this PR replaces the **`to`** in log messages of type "connecting to", "connected to", "connection to" with a **`with`**.

NOTE: in certain cases, I might have replaced a correct **`to`** with a misleading **`with`**, so please pay attention to the various occurrences :-)

### Testing

Changes are only in literal strings, no testing performed.

### Documentation

No documentation necessary.

## Authorship

based on the commit metadata.

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -227,7 +227,7 @@ func (s *service) handle(stop chan struct{}) {
 		// though, especially in the presence of NAT hairpinning, multiple
 		// clients between the same NAT gateway, and global discovery.
 		if remoteID == s.myID {
-			l.Infof("Connected to myself (%s) at %s - should not happen", remoteID, c)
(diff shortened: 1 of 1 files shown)


marco-m:
> Imho many of these do not make sense. “Connected with myself” and “secure connection with”, for example.

Do you mean that I should leave them as-is or that I should remove the log invocation completely ?

marco-m:
Closing due to no activity

PR #5967, opened 2019-08-19
lib/db: Use different defaults for larger databases (fixes #5966)

### LINKED ISSUE #5966: Database performance: improve situation with huge databases
A very large database can today result in almost continuous compaction and resulting very bad performance.

calmh:
This introduces a better set of defaults for large databases. I've
experimentally determined that it results in much better throughput in a
couple of scenarios with large databases, but I can't give any
guarantees the values are always optimal. They're probably no worse than
the defaults though.

Tested in production (the values, not this specific code).

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -284,12 +284,13 @@ func dbIsLarge(location string) bool {
 	}
 
 	marker := filepath.Join(location, dbLargeMarkerFile)
-	if size > dbLargeThreshold {
-		// Database is large.
+	if !large && size > dbLargeThreshold {
+		// Database is large, there is no marker.
 		_ = ioutil.WriteFile(marker, []byte("Say, that's a large database you have there!\n"), 0644)
 		large = true
-	} else if large && size < dbLargeThreshold {
-		// Database has shrunk
+	} else if large && size < dbLargeThreshold/2 {
+		// Database was large but has shrunk to less than half the
+		// threshold.
 		_ = os.Remove(marker)
 		large = false
 	}
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -7,7 +7,9 @@
 package db
 
 import (
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
@@ -30,7 +32,8 @@ const (
 	// have dbMaxOpenFiles of them we will need to start thrashing fd:s.
 	// Switching to large database settings causes larger files to be used
 	// when compaticting, reducing the number.
-	largeDBThreshold = dbMaxOpenFiles * (2 << MiB)
+	dbLargeThreshold  = dbMaxOpenFiles * (2 << MiB)
+	dbLargeMarkerFile = "LARGE"
 
 	KiB = 10
 	MiB = 20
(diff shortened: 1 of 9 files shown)


calmh:
Changing these parameters doesn't cause a conversion of the existing data, it just causes files created in the next compaction round etc. to be different. But regardless of that one can of course imagine a database that is precisely on the edge between the two stages and cycles "randomly" between them on each restart.

My first attempt at this was actually to grab database statistics (leveldb.DBStats) and store those in a JSON file, but it felt a bit heavy handed to maintain. How about I create a marker file, `LARGE` or something inside the DB, and use that for hysteresis? (I'm 99% sure random files in the DB directory are fine and left alone, but I'll double check.) Worst case the marker file can be `${dbname}.large` (i.e., `index-v0.14.0.db.large`) or something...

imsodin:
I was thinking about en entry in the misc db part, but that creates a hen-egg problem :) Marker file seems reasonable. However maybe it's also too many moving parts for a very unlikely case - feel free to not do anything about it.

AudriusButkevicius:
Whats the problem with miscdb? Surely if it grew big, its ok to roll with the big settings even if post compaction it drops below the threshold as it has a tendency to grow? I think miscdb is perfect for this, if it's not there, make a decision, and stick to it. Surely for small dbs the large settings don't have that much of an impact so perhaps we just default to large by default?

I guess I am more concerned when folders get removed

AudriusButkevicius:
Rereading my comment I seem to disagree with both of my own statements in the same comment... :man_shrugging:

calmh:
The problem with miscdb is that to check the value we need to open the db, but we need to know this prior to opening the db.

I'm not sure flipflopping is a problem, but at the very least it might be odd and certainly something the leveldb dude doesn't test extensively for. Picking one set of parameters and using it for as long as feasible is likely better.

AudriusButkevicius:
Is that your thinking or is that what you were actually told?

I think the most sane approach in this case is flipflopping.

calmh:
Completely my own thinking, no official data. I was to begin with somewhat suspicious that it would even accept a database with a mixture of block sizes, table sizes, etc, but that seems to be fine. That said - the marker stuff is literally about three lines of code in `dbIsLarge` -- is it really a big deal?

AudriusButkevicius:
I'd go with a dbprofile file that holds the string that's the profile, in case there is a medium, ir in case we want to support forcing a certain profile. E.g user replacing "small-auto" with "small" to force the small profile, or something like that. Or a json dump of the params (woth an extra mode param declaring whether its auto or not) in the file, so that it's tweakable by the user.

calmh:
I guess at that point we might as well stick it in the config instead?

AudriusButkevicius:
:man_shrugging: idk, maybe? You still need something for auto vs leave it as it is, as I guess we want the experience to be good by default.

calmh:
In that case I'll make it `auto`/`small`/`large` with `auto` (the default) having the risk of flipflopping in rare circumstances and the other two being pinning in either direction.

calmh:
Type and constant duplication is because we currently don't have any dependency edges `db` <-> `config` and I don't think either direction makes sense to have

lib/db/lowlevel.go
```diff
@@ -49,18 +64,52 @@ type Lowlevel struct {
...
+	large := dbIsLarge(location, tuning)
+	opts := optsFor(large)
+	return open(location, opts)
+}
+
+// optsFor returns the database options to use when opening a database. The
+// large bool indicates whether we think the database should use the large
+// mode parameters. Settings can be overridden by debug environment
+// variables.
+func optsFor(large bool) *opt.Options {
```
AudriusButkevicius: Feels like this should take the tuning, and call dbIsLarge in case of auto.

PR #5972, opened 2019-08-21
lib/connections: Upgrade QUIC package, use contexts for timeout

calmh:
### Purpose

Use the new hotness.

### Testing

QUIC still works for me with this change.


lib/connections/quic_listen.go
```diff
@@ -105,53 +109,33 @@ func (t *quicListener) serve(stop chan struct{}) error {
 	l.Infof("QUIC listener (%v) starting", packetConn.LocalAddr())
 	defer l.Infof("QUIC listener (%v) shutting down", packetConn.LocalAddr())
 
-	// Accept is forever, so handle stops externally.
-	go func() {
-		select {
-		case <-stop:
-			_ = listener.Close()
```
AudriusButkevicius: This is now nowhere to be found.

lib/connections/quic_listen.go
```diff
@@ -105,53 +109,33 @@ func (t *quicListener) serve(stop chan struct{}) error {
...
-
 		select {
 		case <-stop:
-			if err == nil {
-				_ = session.Close()
-			}
 			return nil
 		default:
 		}
+
```
AudriusButkevicius: Don't need the select thing above as the parent context would be cancelled? Potentially the for loop should check for ctx.Done()'ness?

calmh: We could check the ctx instead of the stop channel if we like, but we need to break the loop somehow so I think the select is still required as-is.

calmh: Oh now I see what you mean

calmh: Or, well, we could `for ctx.Err() == nil {` instead; do you think that's cleaner?

AudriusButkevicius: Idk, my point is the context is the token of termination, not the channel so we should use that.

calmh:
I removed the timeout on accept -- it'll stop when the context closes and there's no need for us to loop on it otherwise I think.

lib/connections/quic_listen.go
```diff
@@ -101,57 +105,32 @@ func (t *quicListener) serve(stop chan struct{}) error {
...
-				_ = session.Close()
-			}
+		case <-ctx.Done():
 			return nil
 		default:
 		}
+
+		session, err := listener.Accept(ctx)
 		if err != nil {
 			if err, ok := err.(net.Error); !ok || !err.Timeout() {
```
imsodin: Also don't warn for `err == context.Canceled`. And in principle that could be used to break the loop instead of the select above (but I think the explicit check at the beginning like now is cleaner).

calmh: Fixd. I'm keeping the select because there is no guarantee that the QUIC package won't do a `return fmt.Errorf("accept: %v", err)` itself tomorrow and then the `==` will fail. (Right now it works.) (Longer term we should use `xerrors.Is`...)

PR #5973, opened 2019-08-21
lib/util: Add caller info to service (ref #5932)

imsodin:
If something goes wrong with a service (e.g. #5932) there's no information about what created the service. This adds this information by retrieving the first stack frame outside of the `lib/util` package. Currently it is used to implement `fmt.Stringer` and printed if `Stop` is called more than once.

### DIFF
--- a/lib/util/utils.go
+++ b/lib/util/utils.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/url"
 	"reflect"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -178,7 +179,7 @@ func Address(network, host string) string {
 // AsService wraps the given function to implement suture.Service by calling
 // that function on serve and closing the passed channel when Stop is called.
 func AsService(fn func(stop chan struct{})) suture.Service {
-	return AsServiceWithError(func(stop chan struct{}) error {
+	return asServiceWithError(func(stop chan struct{}) error {
 		fn(stop)
 		return nil
 	})
@@ -186,6 +187,7 @@ func AsService(fn func(stop chan struct{})) suture.Service {
 
 type ServiceWithError interface {
 	suture.Service
+	fmt.Stringer
 	Error() error
 	SetError(error)
 }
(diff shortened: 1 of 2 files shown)


lib/util/utils.go
```diff
@@ -178,22 +179,37 @@ func Address(network, host string) string {
...
+	return asServiceWithError(fn)
+}
+
+// caller retrieves information about the creator of the service, i.e. the stack
+// two levels up from itself.
+func caller() string {
+	pc := make([]uintptr, 1)
+	_ = runtime.Callers(4, pc)
+	f, _ := runtime.CallersFrames(pc).Next()
+	return f.Function
```
AudriusButkevicius: This might need a few frames I think, as we have utilities that turn things into services, so the actual service creation is two frames up.

imsodin: I wrote a test to make sure it is actually 4 frames up, and that passes. If there's utility stuff on top of `lib/util`, we can't really know about that in lib/util, so we'd have to pass that information as an argument at which point we can just add `name string` to the arguments (but that would be tedious).

AudriusButkevicius: Why would adding a name be tedious?

calmh: I also think this is not tedious. :) `AsService(name string, fn func(stop chan struct{}))` seems perfectly cromulent

imsodin: Yeah, nothing tedious about the function itself, just tedious to change all invocations (though hopefully in most cases it will be simple as the caller hopefully implements `fmt.Stringer`). I'll maybe consider that along with a change to `context` instead of the stop channel.

calmh:
All that said, none of this is going to help at all with the backtrace in the linked issue. The name we're talking about is just used by suture, not the Go backtrace printer.

imsodin:
Not in general in a backtrace, but for that specific issue the added custom panic instead of closing a closed channel should do it: https://github.com/syncthing/syncthing/pull/5973/files#diff-5a0f9444d3e15d1089040dcf82adb577R257

calmh:
Right yeah ok, in the double stop case or other suture-printed problems yeah.

AudriusButkevicius:
In principle this change is no worse than the existing situation, so we should merge this, unless we plan to start giving things names.

PR #5975, opened 2019-08-22
lib/model: Hold locks as shortly as possible

imsodin:
After Moisie reported some deadlock false positives (https://forum.syncthing.net/t/web-based-syncthing-not-showing-info/13529/37) I had another look at all the `fmut`/`pmut` locks in model. I found some more cases where a lock could be replaced by a read-lock and calling functions could be moved outside of the lock. The bigger change was getting rid of the `add-/start-/tearDownFolderLocked` functions respectively move the locking inside those functions. We already did release the lock inside two of those functions, so there's arguably no point in doing so and now I limited the locking even further.

Unit and integration tests passed.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -240,17 +240,15 @@ func (m *model) StartDeadlockDetector(timeout time.Duration) {
 
 // StartFolder constructs the folder service and starts it.
 func (m *model) StartFolder(folder string) {
-	m.fmut.Lock()
-	defer m.fmut.Unlock()
-	folderCfg := m.folderCfgs[folder]
-	m.startFolderLocked(folderCfg)
+	var tokens []suture.ServiceToken
 
-	l.Infof("Ready to synchronize %s (%s)", folderCfg.Description(), folderCfg.Type)
-}
+	m.fmut.RLock()
+	cfg := m.folderCfgs[folder]
+	fset := m.folderFiles[folder]
+	ignores := m.folderIgnores[folder]
+	_, ok := m.folderRunners[folder]
+	m.fmut.RUnlock()
 
-// Need to hold lock on m.fmut when calling this.
-func (m *model) startFolderLocked(cfg config.FolderConfiguration) {
-	_, ok := m.folderRunners[cfg.ID]
 	if ok {
 		l.Warnln("Cannot start already running folder", cfg.Description())
 		panic("cannot start already running folder")
@@ -260,10 +258,6 @@ func (m *model) startFolderLocked(cfg config.FolderConfiguration) {
 	if !ok {
 		panic(fmt.Sprintf("unknown folder type 0x%x", cfg.Type))
 	}
-
-	folder := cfg.ID
-
-	fset := m.folderFiles[folder]
 
 	// Find any devices for which we hold the index in the db, but the folder
 	// is not shared, and drop it.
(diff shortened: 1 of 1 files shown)


lib/model/model.go
```diff
@@ -1907,15 +1894,18 @@ func (m *model) DownloadProgress(device protocol.DeviceID, folder string, update
 }
 
 func (m *model) deviceStatRef(deviceID protocol.DeviceID) *stats.DeviceStatisticsReference {
-	m.fmut.Lock()
-	defer m.fmut.Unlock()
-
-	if sr, ok := m.deviceStatRefs[deviceID]; ok {
+	m.fmut.RLock()
```
calmh: This is wrong as concurrent calls might both conclude there is no ref and create a new one, clobbering each other. The original is correct.

lib/model/model.go
```diff
@@ -1823,21 +1809,22 @@ func (m *model) AddConnection(conn connections.Connection, hello protocol.HelloR
 		return
 	}
 
-	m.pmut.Lock()
-	if oldConn, ok := m.conn[deviceID]; ok {
+	m.pmut.RLock()
```
calmh: I think both the new and old code is incorrect here. If there is an old connection we will attempt to close it. But since we don't hold the lock over the close, and don't remove the connection from `m.conn` we can have two concurrent calls both seeing the old conn and closing it, then setting the new conn with the other call clobbering it...

lib/model/model.go
```diff
@@ -369,41 +363,33 @@ func (m *model) AddFolder(cfg config.FolderConfiguration) {
...
-// Need to hold lock on m.fmut when calling this.
-func (m *model) tearDownFolderLocked(cfg config.FolderConfiguration, err error) {
+func (m *model) tearDownFolder(cfg config.FolderConfiguration, err error) {
 	// Stop the services running for this folder and wait for them to finish
 	// stopping to prevent races on restart.
+	m.fmut.RLock()
 	tokens := m.folderRunnerTokens[cfg.ID]
-
-	m.fmut.Unlock()
+	m.fmut.RUnlock()
```
calmh: Here also I think both old and new is broken. Some rethinking required. :/

lib/model/model.go
imsodin: In principle yes, in practice this should be prevented by `AddConnection` being called from a single loop in `lib/service`, thus this cannot be called concurrently.

imsodin:
I should have mentioned this in the PR description: I considered `fmut`/`pmut` solely as protections, not as gate-keepers for correct execution order/prevent unwanted parallelism and tried to figure out whether or not such gate-keeping becomes necessary by my changes.  
E.g. `tearDownFolder` is only called from `CommitConfiguration`, which I assumed to not be called concurrently, but that's not the case currently (#5298). However I think the right thing to do then is to protect that from being called concurrently, because that's definitely bad (we want the latest changes to take effect). Simplest thing would be just a commitConfiguration mutex (thought that wouldn't guarantee ordering).

lib/model/model.go
imsodin: Right. My beaf with this is that an informational call can block everything else due to taking that lock. In https://github.com/syncthing/syncthing/pull/5975/commits/df1ffe610049957195cc9ae877399b3bc9379288 I moved the creation of `deviceStatRefs` into `CommitConfiguration` such that we only require read-locks here.

calmh:
I don't trust this PR enough to merge it. Some things are clearly fine, but other things aren't (clearly fine) and we have no testing in place to verify in either direction. We can assume that some things aren't called concurrently but that's something that will come back and bite us in the ass. I'd rather assume that things *are* called concurrently and have the current locking in place to assure correctness. At least that's the intention, although as we can see in some places it still isn't fully correct today...

calmh:
I think this is dead as-is and has been superseded in the uncontroversial parts. Hence closing, yell if I'm wrong.

PR #5987, opened 2019-08-30
lib/model: Correctly handle manual rescan with ignore error (fixes #5985)

### LINKED ISSUE #5985: Log spam and high CPU use if #include path is wrong in .stignore
I have first reported this problem in the SyncTrayzor repository (https://github.com/canton7/SyncTrayzor/issues/526), as I use Syncthing mostly on Windows, but I am reporting it here too, as I believe that this is a general issue.

The problem is that if #include in .stignore is set to a non-existent file path, Syncthing continuously spams the log file with the following entries.

```
[***] **:**:** WARNING: Error on folder "***" (***-***): loading ignores: failed to load include file ***: open \\?\***: The system cannot find the path specified.
[***] **:**:** INFO: Cleared error on folder "***" (***-***)
```

Because of the log spam, Syncthing uses a lot of CPU which bogs down the whole system. If SyncTrayzor is open, it will actually use 100% of CPU and make everything unresponsive due to the console log display in SyncTrayzor itself. Without SyncTrayzor, it is just Syncthing.exe using ~50% of CPU in a mid-range 4 core system. Fortunately, the system is still responsive enough to operate, but this may not be the case on a slower device. Also, the folder status in the Web GUI keeps changing rapidly with no end to it. The only way to fix the issue seems to be to manually edit the `.stignore` file and fix the incorrect path there which as a result stops the log spam. Unfortunately, the Web GUI does not allow to edit the incorrect path directly, as the "Ignored Patterns" window is grayed out with a `Failed to load ignored patterns.` message.

Would it be possible to simply log the problem once so that the user knows of the incorrect path, and not spam the log with the error messages? Right now, even a simple spellling mistake in the #include path can cause this issue. I have not tested this in Android, but if the same thing happens there too, then stopping the log spam there will be major pain, as you would need either a 3rd party file manager or a terminal application to even be able to see and modify/delete the hidden .stignore files.

### Version Information

Syncthing Version: v1.2.1
OS Version: Windows 7 Enterprise x64

calmh:
Assume a folder error was set due to bad ignores on the latest scan.
Previously, doing a manual rescan would result in:

1. Clearing the folder error, which schedules (immediately) an fs
   watcher restart

2. Attempting to load the ignores, which fails, so we set a folder
   error and bail.

3. Now the fs watcher restarts, as scheduled, so we trigger a scan.
   Goto 1.

This change fixes this by not clearing the error until the error is
actually cleared, that is, if both the health check and ignore loading
succeeds.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -171,20 +171,23 @@ func (f *folder) serve(_ chan struct{}) {
 			}
 
 		case <-f.scanTimer.C:
-			l.Debugln(f, "Scanning subdirectories")
+			l.Debugln(f, "Scanning due to timer")
 			f.scanTimerFired()
 
 		case req := <-f.scanNow:
+			l.Debugln(f, "Scanning due to request")
 			req.err <- f.scanSubdirs(req.subdirs)
 
 		case next := <-f.scanDelay:
+			l.Debugln(f, "Delaying scan")
 			f.scanTimer.Reset(next)
 
 		case fsEvents := <-f.watchChan:
-			l.Debugln(f, "filesystem notification rescan")
+			l.Debugln(f, "Scan due to watcher")
 			f.scanSubdirs(fsEvents)
 
 		case <-f.restartWatchChan:
+			l.Debugln(f, "Restart watcher")
 			f.restartWatch()
 		}
 	}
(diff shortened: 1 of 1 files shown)


calmh:
Yeah I was suspicious about that one as well, I also have a feeling that I didn't understand the "restart watcher" path in terms of where it gets torn down. I'll look into this.

PR #5988, opened 2019-09-01
Added a help-block string to the 'File Versioning' tab

ralienpp:
### Purpose

Provide a bit more information to users when they create new folders, to clarify that "File versioning" can be related to file deletion, not just overwriting. 

A broader context is available on the forum: https://forum.syncthing.net/t/missing-syncthing-directory-missing-data-on-devices/13684/9?u=ralienpp

### Testing

None, I looked at the code of the page and added this string, in a way analogous to other tabs.

### DIFF
--- a/gui/default/syncthing/folder/editFolderModalView.html
+++ b/gui/default/syncthing/folder/editFolderModalView.html
@@ -65,6 +65,9 @@
         <div id="folder-versioning" class="tab-pane">
           <div class="form-group">
             <label translate>File Versioning</label>&emsp;<a href="https://docs.syncthing.net/users/versioning.html" target="_blank"><span class="fas fa-question-circle"></span>&nbsp;<span translate>Help</span></a>
+            <p class="help-block">
+              <span translate>What to do if files are replaced or deleted, either locally, or on remote devices.</span>&emsp;
+            </p>
             <select class="form-control" ng-model="currentFolder.fileVersioningSelector">
               <option value="none" translate>No File Versioning</option>
               <option value="trashcan" translate>Trash Can File Versioning</option>
--- a/gui/default/assets/lang/lang-en.json
+++ b/gui/default/assets/lang/lang-en.json
@@ -373,6 +373,7 @@
    "Watch for Changes": "Watch for Changes",
    "Watching for Changes": "Watching for Changes",
    "Watching for changes discovers most changes without periodic scanning.": "Watching for changes discovers most changes without periodic scanning.",
+   "What to do if files are replaced or deleted, either locally, or on remote devices.": "What to do if files are replaced or deleted, either locally, or on remote devices.",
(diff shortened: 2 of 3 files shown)


gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -65,6 +65,9 @@
         <div id="folder-versioning" class="tab-pane">
           <div class="form-group">
             <label translate>File Versioning</label>&emsp;<a href="https://docs.syncthing.net/users/versioning.html" target="_blank"><span class="fas fa-question-circle"></span>&nbsp;<span translate>Help</span></a>
+            <p class="help-block">
+              <span translate>What to do if files are replaced or deleted, either locally, or on remote devices.</span>&emsp;
```
imsodin: This is wrong, as Syncthing can't version local deletes. Syncthing only reacts to changes the user makes locally, i.e. it sees the change after it happened. In case of a deletion, it just notices that there is nothing on disk where previously there was a file - so the data is gone already. Remote deletions on the other hand still exist locally, so instead of just deleting the files, Syncthing can use them for versioning.

AudriusButkevicius: Also this sentence triggers my grammatical spider senses. Not sure what the goal of the sentence is.

Catfriend1: Why not take a short form of the explanation from the docs?
> Syncthing supports archiving the old version of a file when it is deleted or replaced with a newer version from the cluster. Click to see available versioning strategies. File versioning is configured per folder, on a per-device basis.

ralienpp: Thank you for the input and the corrections. The explanation from the docs is better, but I think it would make sense, especially for the Android client, to make it shorter.

The goal of the sentence is to make users aware of the relationship between versioning and deletion.

How about a slightly more concise wording:

- I removed "per folder", because we're already looking at a folder-specific dialog, thus it is of reason to expect that the settings are folder-specific
- "Click to see..." is removed to shorten the text, but I don't insist on adding it

> Syncthing supports archiving the old version of a file when it is deleted or replaced with a newer version from the cluster. File versioning is configured on a per-device basis.

image

imsodin: I'd completely remove the "per-device" sentence - not even sure what it is trying to say. Then I'd replace "the cluster" with "remote devices", as it's constent with the rest of the UI and easier to understand. And starting with "Syncthing supports" in a Syncthing dialog feels redundant, I'd rather write something like "Choose a strategy to archive old versions of a files when they are deleted or replaced with a newer version from remote devices."

AudriusButkevicius:
Hi,

Do you have any intentions of addressing the review comments, or should we close this?

Thanks.
A

st-review:
:robot: *beep boop*

PR #5992, opened 2019-09-03
lib/api: Skip an IPv6 specific test inside Docker (fixes #5991)

### LINKED ISSUE #5991: lib/api: TestHostCheck never finishes with Go 1.13rc2 under Docker
`go test -v -race -short ./lib/api`

Go 1.12:

```
=== RUN   TestHostCheck
22:17:37 INFO: GUI and API listening on 127.0.0.1:49708
22:17:37 INFO: Access the GUI via the following URL: http://127.0.0.1:0/
22:17:37 INFO: GUI and API listening on 127.0.0.1:49713
22:17:37 INFO: Access the GUI via the following URL: http://127.0.0.1:0/
22:17:37 INFO: GUI and API listening on [::]:49715
22:17:37 INFO: Access the GUI via the following URL: http://127.0.0.1:0/
22:17:37 INFO: GUI and API listening on [::1]:49717
22:17:37 INFO: Access the GUI via the following URL: http://[::1]:0/
--- PASS: TestHostCheck (0.21s)
```

Go 1.13rc2:

```
=== RUN   TestHostCheck
20:15:27 INFO: GUI and API listening on 127.0.0.1:34355
20:15:27 INFO: Access the GUI via the following URL: http://127.0.0.1:0/
20:15:27 INFO: GUI and API listening on 127.0.0.1:42867
20:15:27 INFO: Access the GUI via the following URL: http://127.0.0.1:0/
20:15:27 INFO: GUI and API listening on [::]:36807
20:15:27 INFO: Access the GUI via the following URL: http://127.0.0.1:0/
(panics after x minutes)
```

Seems to *only* be under Docker, so probably some sort of IPv6 badness there that triggers it...

calmh:
Isn't this just the most beautiful thing you've ever seen?

### Testing

```
docker run -it --rm -v $(pwd):/foo -w /foo -e GOPROXY=direct \
  golang:1.13rc2 go test -v -race -short ./lib/... ./cmd/...
```

### DIFF
--- a/lib/api/api_test.go
+++ b/lib/api/api_test.go
@@ -18,6 +18,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"testing"
@@ -826,6 +827,11 @@ func TestHostCheck(t *testing.T) {
 
 	// This should all work over IPv6 as well
 
+	if runningInContainer() {
+		// Working IPv6 in Docker can't be taken for granted.
+		return
+	}
+
 	cfg = new(mockedConfig)
 	cfg.gui.RawAddress = "[::1]:0"
 	baseURL, err = startHTTP(cfg)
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Why don't we check that the ipv6 stuff is supported instead?

AudriusButkevicius:
Checking half of those things does not mean docker, it just means cgroups or lxc, which could just be systemd.

calmh:
That's not what my Googling told me, but if you have a better test in mind please don't hesitate to show it :)

imsodin:
An ugly, lazy method would be to introduce a `STTESTSSKIPIPV6` env var.

PR #5996, opened 2019-09-04
lib/connections: Improve write rate limiting (fixes #5138)

### LINKED ISSUE #5138: Rate limit can be very bursty
Filing this to see if anyone thinks it's a problem...

We do rate limiting by delaying read/write operations until the amount of data in the read/write is allowed by the desired rate. The layer above this deals in messages. So at least for writes, a message will go out as a single write. Messages can be large, especially with large blocks. So, for example, a 100 KiB/s rate limit sending a 4 GiB file (4 MiB blocks) will result in one message every 40 seconds. That's 4 MiB sent infinitely fast, and then a long wait.

If the purpose of the rate limit is to avoid choking an ADSL link to avoid dropping VoIP packets (for example) this doesn't accomplish the desired effect at all.

If the purpose is just to limit the amount of data sent during any given hour or so, it's fine.

Solving this would involve making the LimitedReader/Writer smarter about the amount of data in relation to the limit and potentially chopping up the operations into smaller reads/writes.

calmh:
This splits large writes into smaller ones when using a rate limit,
making them into a legitimate trickle rather than large bursts with a
long time inbetween.

### Testing

Unit test to make sure the end result is correct and I didn't fumble the offsets or whatnot.

### DIFF
--- a/lib/connections/limiter.go
+++ b/lib/connections/limiter.go
@@ -232,7 +232,7 @@ type limitedReader struct {
 
 func (r *limitedReader) Read(buf []byte) (int, error) {
 	n, err := r.reader.Read(buf)
-	if !r.isLAN || r.limitsLAN.get() {
+	if !r.isLAN || r.limitsLAN.get() && !isUnlimited(r.waiter) {
 		take(r.waiter, n)
 	}
 	return n, err
@@ -247,7 +247,7 @@ type limitedWriter struct {
 }
 
 func (w *limitedWriter) Write(buf []byte) (int, error) {
-	if w.isLAN && !w.limitsLAN.get() {
+	if w.isLAN && !w.limitsLAN.get() || isUnlimited(w.waiter) {
 		// Fast path
 		return w.writer.Write(buf)
 	}
@@ -292,6 +292,14 @@ func take(waiter waiter, tokens int) {
 			tokens = 0
 		}
 	}
+}
+
+func isUnlimited(waiter waiter) bool {
+	rl, ok := waiter.(*rate.Limiter)
+	if !ok {
+		return false
+	}
+	return rl.Limit() == rate.Inf
 }
 
 type atomicBool int32
--- a/lib/connections/limiter.go
+++ b/lib/connections/limiter.go
@@ -32,6 +32,7 @@ type limiter struct {
 type waiter interface {
 	// This is the rate limiting operation
 	WaitN(ctx context.Context, n int) error
+	Limit() rate.Limit
 }
 
 const (
@@ -303,11 +304,7 @@ func take(waiter waiter, tokens int) {
 }
 
 func isUnlimited(waiter waiter) bool {
-	rl, ok := waiter.(*rate.Limiter)
-	if !ok {
-		return false
-	}
-	return rl.Limit() == rate.Inf
+	return waiter.Limit() == rate.Inf
 }
 
 type atomicBool int32
@@ -336,4 +333,14 @@ func (tw totalWaiter) WaitN(ctx context.Context, n int) error {
 		}
 	}
 	return nil
+}
+
+func (tw totalWaiter) Limit() rate.Limit {
+	min := rate.Inf
+	for _, w := range tw {
+		if l := w.Limit(); l < min {
+			min = l
+		}
+	}
+	return min
 }
--- a/lib/connections/limiter.go
+++ b/lib/connections/limiter.go
@@ -287,24 +287,29 @@ func (w waiterHolder) unlimited() bool {
 	return w.waiter.Limit() == rate.Inf
 }
 
-// take is a utility function to consume tokens from a overall rate.Limiter and deviceLimiter.
-// No call to WaitN can be larger than the limiter burst size so we split it up into
-// several calls when necessary.
+// take is a utility function to consume tokens, because no call to WaitN
+// must be larger than the limiter burst size or it will hang.
 func (w waiterHolder) take(tokens int) {
+	// For writes we already split the buffer into smaller operations so those
+	// will always end up in the fast path below. For reads, however, we don't
+	// control the size of the incoming buffer and don't split the calls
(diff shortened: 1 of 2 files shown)


lib/connections/limiter.go
```diff
@@ -244,10 +247,28 @@ type limitedWriter struct {
 }
 
 func (w *limitedWriter) Write(buf []byte) (int, error) {
-	if !w.isLAN || w.limitsLAN.get() {
-		take(w.waiter, len(buf))
+	if w.isLAN && !w.limitsLAN.get() || isUnlimited(w.waiter) {
```
AudriusButkevicius: The operator order is not obvious, can you use ( ) to make it more obvious?

lib/connections/limiter.go
```diff
@@ -228,13 +231,18 @@ type limitedReader struct {
...
+		return r.reader.Read(buf)
 	}
+	n, err := r.reader.Read(buf)
+	take(r.waiter, n)
 	return n, err
 }
 
+func (r *limitedReader) fastpath() bool {
+	return r.isLAN && !r.limitsLAN.get() || isUnlimited(r.waiter)
+}
```
AudriusButkevicius: This forces me to check the spec for operator precedence, please use ()

calmh:
The looping over waiters to see if they are all unlimited, for the fast path, might look inefficient. But prior to this we'd anyway loop over them all to take tokens from them even when unlimited, so it's an improvement.

lib/connections/limiter.go
```diff
@@ -228,13 +232,18 @@ type limitedReader struct {
...
 
 func (r *limitedReader) Read(buf []byte) (int, error) {
-	n, err := r.reader.Read(buf)
-	if !r.isLAN || r.limitsLAN.get() {
-		take(r.waiter, n)
+	if r.fastpath() {
+		return r.reader.Read(buf)
 	}
+	n, err := r.reader.Read(buf)
+	take(r.waiter, n)
```
AudriusButkevicius: If !fastpath { take }

lib/connections/limiter.go
```diff
@@ -243,11 +252,32 @@ type limitedWriter struct {
 	isLAN     bool
 }
 
+func (w *limitedWriter) fastpath() bool {
+	return w.isLAN && !w.limitsLAN.get() || isUnlimited(w.waiter)
+}
```
AudriusButkevicius: ( )

lib/connections/limiter.go
```diff
@@ -221,53 +229,82 @@ func getRateLimiter(m map[protocol.DeviceID]*rate.Limiter, deviceID protocol.Dev
...
 // take is a utility function to consume tokens from a overall rate.Limiter and deviceLimiter.
 // No call to WaitN can be larger than the limiter burst size so we split it up into
 // several calls when necessary.
-func take(waiter waiter, tokens int) {
+func (w waiterHolder) take(tokens int) {
 	if tokens < limiterBurstSize {
 		// This is the by far more common case so we get it out of the way
 		// early.
-		waiter.WaitN(context.TODO(), tokens)
+		w.waiter.WaitN(context.TODO(), tokens)
```
golangcibot: Error return value of `w.waiter.WaitN` is not checked (from `errcheck`)

lib/connections/limiter.go
```diff
@@ -221,53 +229,82 @@ func getRateLimiter(m map[protocol.DeviceID]*rate.Limiter, deviceID protocol.Dev
...
 		return
 	}
 
 	for tokens > 0 {
 		// Consume limiterBurstSize tokens at a time until we're done.
 		if tokens > limiterBurstSize {
-			waiter.WaitN(context.TODO(), limiterBurstSize)
+			w.waiter.WaitN(context.TODO(), limiterBurstSize)
```
golangcibot: Error return value of `w.waiter.WaitN` is not checked (from `errcheck`)

lib/connections/limiter.go
```diff
@@ -221,53 +229,82 @@ func getRateLimiter(m map[protocol.DeviceID]*rate.Limiter, deviceID protocol.Dev
...
 			tokens -= limiterBurstSize
 		} else {
-			waiter.WaitN(context.TODO(), tokens)
+			w.waiter.WaitN(context.TODO(), tokens)
```
golangcibot: Error return value of `w.waiter.WaitN` is not checked (from `errcheck`)

lib/connections/limiter.go
```diff
@@ -280,33 +271,45 @@ func (w *limitedWriter) Write(buf []byte) (int, error) {
...
 		// This is the by far more common case so we get it out of the way
 		// early.
-		waiter.WaitN(context.TODO(), tokens)
+		w.waiter.WaitN(context.TODO(), tokens)
 		return
 	}
 
 	for tokens > 0 {
 		// Consume limiterBurstSize tokens at a time until we're done.
 		if tokens > limiterBurstSize {
```
calmh: This stuff will now never happen for the writer, but it's *possible* for a read() to return more data than the limiter burst so we still need this looping function

AudriusButkevicius: Worth changing the comment?

lib/connections/limiter_test.go
```diff
@@ -185,6 +189,92 @@ func TestAddAndRemove(t *testing.T) {
...
+
+	// Verify there were a single write and that the end result is identical.
+	if cw.writeCount != 1 {
+		t.Error("expected just the one write")
+	}
+	if !bytes.Equal(src, dst.Bytes()) {
+		t.Error("results should be equal")
+	}
+}
+
```
AudriusButkevicius: Should we have a test with two writers, picking the lower limit of the two?

calmh: I added a test. Number of writes isn't decided by the rate though but by the max write size constant. (Changed the (irrelevant) rate of the limiters to avoid this confusion.)

PR #5997, opened 2019-09-04
cmd/*, lib/build: Set correct LongVersion (fixes #5993)

### LINKED ISSUE #5993: Build information missing in discovery and relay server binaries
Since v1.2.2, the discovery and relay server releases are missing the `build user, build host` and `build date` information. This causes the following awkward output on startup:

```
Started Syncthing Discovery Server.
stdiscosrv  (go1.12.9 linux-amd64) @ 1970-01-01 00:00:00 UTC
```
and

```
Started Syncthing Relay Server.
main.go:149: strelaysrv  (go1.12.9 linux-amd64) @ 1970-01-01 00:00:00 UTC
```

The builder info is just an @ and the build date is the UNIX timestamp of zero.

Older releases (v1.0.1) print correct data here, for example `stdiscosrv v1.0.1 (go1.11.5 linux-amd64) teamcity@build.syncthing.net 2019-01-18 10:34:18 UTC`.

The main syncthing binary is also fine and prints everything correctly, for example:

`syncthing v1.2.2 "Fermium Flea" (go1.12.9 windows-amd64) teamcity@build.syncthing.net 2019-08-15 13:51:09 UTC`

As far as I've diagnosed the issue apart from the output on startup everything works OK.

I haven't looked at any other `/cmd` binaries apart from the three mentioned above, so I don't know if other executables are affected.

### Version Information

Syncthing Version: v1.2.2 (syncthing, discovery, relay)
OS Version: Bug discovered/verified on Linux amd64.

EDIT: Just for the sake of completeness: I forgot to mention that the `version` information is missing too, there's just a whitespace where "v1.2.2" should have been.

calmh:
The relay and discosrv didn't use the new lib/build package, now they
do. Conversely the lib/build package wasn't aware there might be other
users and hard coded the program name - now it's adjustable.

(Alternatively we could provide a slightly shorter LongVersion and have
the provider add the name, but that might result in different formats
used and it required "syncthing" be hardcoded in places like lib/api.
I think having that as the default and letting someone fix it up early
in main is OK.)

I also added `-version` flags to those programs.


lib/build/build.go
```diff
@@ -36,6 +36,9 @@ var (
 	// Set by Go build tags
 	Tags []string
 
+	// Set by the caller
+	program = "syncthing"
```
imsodin: I guess using `Syncthing` for the proper "brand name" would be too scary (as there might be places where `syncthing` is expected/hardcoded).

cmd/stcli/main.go
```diff
@@ -29,6 +28,8 @@ import (
 )
 
 func main() {
+	build.SetProgram("stcli")
+
```
AudriusButkevicius: Have a constant, use it here and in app.Name?

lib/build/build.go
calmh: Yeah... Breaking change, reserved for 2.0!

calmh:
Yes... My first attempt was actually to add it to the build script, but that didn't work out as we do `go install ./cmd/...`, but if we had an explicit loop it might be better. Let me check that out.

imsodin:
> This way you can't forget to set it.

Then there must be some check in `lib/build` that the name is not empty. Otherwise someone building directly must know about the -X argument to not silently get an empty program name.

calmh:
They already get a whole bunch of `unknown-dev unknown@unknown 1970-01-01` crap in the version if they do that. But I think defaulting to `syncthing` is fine. If the other ldflags are already in place there is no difference, if they're building some other random tool and it incorrectly calls itself syncthing 🤷‍♂

AudriusButkevicius:
I think unknown-prog is ok too?

calmh:
I wouldn't be surprised if builders like Debian have adopted our ldflags but not our script, and it's unnecessary to break them as part of this change...

imsodin:
I am pretty sure they do. I think we should stay as compatible with the canonical go build process (just using `go build`) as possible. I.e. our build script takes care of all the stuff and is recommended to use, but running the appropriate `go` commands by yourself (assets and stuff) should be as painless as possible.

calmh:
I agree. This now doesn't change that situation;

```
jb@kvin:~/d/s/syncthing $ go build ./cmd/syncthing/
jb@kvin:~/d/s/syncthing $ ./syncthing -version
syncthing unknown-dev "Fermium Flea" (go1.13 darwin-amd64) unknown@unknown 1970-01-01 00:00:00 UTC
jb@kvin:~/d/s/syncthing $
```

imsodin:
I would be in favor of the previous solution (setting the name in the main functions of the respective targets) instead of introducing another command line option required for building, as then it's on those who know the package to not forget (i.e. us, the maintainers), not downstream packagers or random users building themselves. Just my 2 cents, nothing I am hard pressed on ;)

AudriusButkevicius:
Good to merge from my perspective, I'll leave to the owner to decide if it's actually done.

PR #5998, opened 2019-09-05
lib/api: Refactor to run tests in parallel

calmh:
This is an experiment in testing, based on the advise to always call
t.Parallel() at the start of every test. Doing so makes tests run in
parallel, which is usually faster, but also exposes package level state
and potential race conditions better.

To support this I had to redesign the CSRF manager to not be package
global, which was indeed an improvement. And tests run five times faster
now.

### DIFF
--- a/lib/api/api_csrf.go
+++ b/lib/api/api_csrf.go
@@ -13,83 +13,103 @@ import (
 	"os"
 	"strings"
 
-	"github.com/syncthing/syncthing/lib/config"
-	"github.com/syncthing/syncthing/lib/locations"
 	"github.com/syncthing/syncthing/lib/osutil"
 	"github.com/syncthing/syncthing/lib/rand"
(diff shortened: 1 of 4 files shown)


lib/api/api_csrf.go
```diff
@@ -13,131 +13,158 @@ import (
...
-		// Verify the CSRF token
-		token := r.Header.Get("X-CSRF-Token-" + unique)
-		if !validCsrfToken(token) {
-			http.Error(w, "CSRF Error", 403)
-			return
-		}
+	// Verify the CSRF token
+	token := r.Header.Get("X-CSRF-Token-" + m.unique)
+	if !m.validToken(token) {
+		http.Error(w, "CSRF Error", 403)
```
golangcibot: ST1013: should use constant http.StatusForbidden instead of numeric literal 403 (from `stylecheck`)

PR #6013, opened 2019-09-15
lib/fs: Check events against both the user and eval root

imsodin:
This PR is a reaction to the following report:  
https://forum.syncthing.net/t/bug-report/13777  
```
2019-09-06 22:51:11: Watching for changes encountered an event outside of the filesystem root: f.root==\?\C:\Etc\Dev, root==\?\f:\dev, path==\?\c:\etc\dev. This should never happen, please report this message to forum.syncthing.net
```

I have no clue why and how it's possible that given the "real" path (no junctions) the windows watcher sometimes returns events for the "linked" path and the "real" path. So I extended the existing path juggling around watches to check against both the folder root as specified by the user and the folder root after applying "canonicalization" steps on it (`os.EvalSymlinks` and syscall stuff).

### DIFF
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -345,6 +345,6 @@ func (e *ErrWatchEventOutsideRoot) Error() string {
 	return e.msg
 }
 
-func (f *BasicFilesystem) newErrWatchEventOutsideRoot(absPath, root string) *ErrWatchEventOutsideRoot {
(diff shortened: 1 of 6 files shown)


AudriusButkevicius:
A test for this specific change would be nice.

calmh:
Especially given this was apparently not even compiled on Windows ;)

imsodin:
There's already tests in place that check a basic watch works, which includes this change. I added one minimal test for `unrootedChecked` when given two roots paths. I don't want to try to touch any actual windows junction or so stuff.

PR #6028, opened 2019-09-23
gui, lib/model: Add new state FolderPreparingSync (fixes #6027)

### LINKED ISSUE #6027: Folder shows as "Syncing" when there is no data transfer
This issue is a companion to https://forum.syncthing.net/t/i-want-to-get-rid-of-the-syncing-status/13793/18. Basically every so often a user gets confused by Syncthing saying it's syncing, while they made sure manually that the data is already the same. The proposed solution is an additional state before "Syncing", while Syncthing is traversing the database to make a list of stuff it needs etc. Details in the forum or PR (incoming).

imsodin:
### Purpose

See #6027:

> This issue is a companion to https://forum.syncthing.net/t/i-want-to-get-rid-of-the-syncing-status/13793/18. Basically every so often a user gets confused by Syncthing saying it's syncing, while they made sure manually that the data is already the same. The proposed solution is an additional state before "Syncing", while Syncthing is traversing the database to make a list of stuff it needs etc. Details in the forum or PR (incoming).

### Testing

Copied a lot of testfiles and saw it stay in the new state for a short moment before going back to idle.

### Screenshots

Screenshot_2019-09-23_15-04-26

(Unrelated: Looks like pulling is faster than updating the global state in the web UI, which rises to 23 during/after pulling - however getting this screenshot required sub-second timing, so it's definitely not an issue.)

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -802,7 +802,7 @@ angular.module('syncthing.core')
             if (status == 'paused') {
                 return 'default';
             }
-            if (status === 'syncing' || status === 'preparing-sync' || status === 'scanning') {
+            if (status === 'syncing' || status === 'sync-preparing' || status === 'scanning') {
                 return 'primary';
             }
             if (status === 'unknown') {
@@ -961,7 +961,7 @@ angular.module('syncthing.core')
             for (var i = 0; i < folderListCache.length; i++) {
                 var status = $scope.folderStatus(folderListCache[i]);
                 switch (status) {
-                    case 'preparing-sync':
+                    case 'sync-preparing':
                     case 'syncing':
                         syncCount++;
                         break;
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -802,7 +802,7 @@ angular.module('syncthing.core')
             if (status == 'paused') {
                 return 'default';
             }
-            if (status === 'syncing' || status === 'scanning') {
+            if (status === 'syncing' || status === 'preparing-sync' || status === 'scanning') {
                 return 'primary';
             }
             if (status === 'unknown') {
@@ -961,6 +961,7 @@ angular.module('syncthing.core')
             for (var i = 0; i < folderListCache.length; i++) {
                 var status = $scope.folderStatus(folderListCache[i]);
                 switch (status) {
+                    case 'preparing-sync':
                     case 'syncing':
                         syncCount++;
                         break;
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -195,7 +195,7 @@ func (f *sendReceiveFolder) pull() bool {
 
 		// Needs to be set on every loop, as the puller might have set
 		// it to FolderSyncing during the last iteration.
-		f.setState(FolderPreparingSync)
+		f.setState(FolderSyncPreparing)
 
 		changed := f.pullerIteration(scanChan)
 
@@ -1392,19 +1392,13 @@ func (f *sendReceiveFolder) pullerRoutine(in <-chan pullBlockState, out chan<- *
 	requestLimiter := newByteSemaphore(f.PullerMaxPendingKiB * 1024)
 	wg := sync.NewWaitGroup()
 
-	// To set the folder state to syncing when the first block is started.
-	active := false
-
 	for state := range in {
 		if state.failed() != nil {
 			out <- state.sharedPullerState
 			continue
 		}
 
-		if !active {
-			f.setState(FolderSyncing)
-			active = true
-		}
+		f.setState(FolderSyncing) // Does nothing if already FolderSyncing
 
 		// The requestLimiter limits how many pending block requests we have
 		// ongoing at any given time, based on the size of the blocks
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -176,7 +176,6 @@ func (f *sendReceiveFolder) pull() bool {
 
 	l.Debugf("%v pulling", f)
 
-	f.setState(FolderSyncing)
 	f.clearPullErrors()
 
 	scanChan := make(chan string)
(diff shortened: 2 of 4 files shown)


gui/default/syncthing/core/syncthingController.js
```diff
@@ -802,7 +802,7 @@ angular.module('syncthing.core')
             if (status == 'paused') {
                 return 'default';
             }
-            if (status === 'syncing' || status === 'scanning') {
+            if (status === 'syncing' || status === 'preparing-sync' || status === 'scanning') {
```
AudriusButkevicius: Why not just preparing?

lib/model/folder_sendrecv.go
```diff
@@ -1389,12 +1392,20 @@ func (f *sendReceiveFolder) pullerRoutine(in <-chan pullBlockState, out chan<- *
...
+	active := false
+
 	for state := range in {
 		if state.failed() != nil {
 			out <- state.sharedPullerState
 			continue
 		}
 
+		if !active {
+			f.setState(FolderSyncing)
```
AudriusButkevicius: Make setState noop if there is no transition?

gui/default/syncthing/core/syncthingController.js
imsodin: I considered that too generic, there might as well be a "preparing-scan" at some point.

lib/model/folder_sendrecv.go
AudriusButkevicius: Its done once per file, makes it less ugly to read

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: Perhaps them preparing-to-sync? Preparing-sync sounds like sync is a thing, ala preparing-dinner

lib/model/folder_sendrecv.go
imsodin: Once per block, but the general point still stands: readability over untested, perceived, probably negligible performance ;)

imsodin:
Yes that's correct. My initial motivation for this change was a preseeded folder, i.e. there's lots of "needed" files but they can all be fast-tracked somehow. That can take a long time and no syncing occurs. In my opinion copying blocks locally is already covered ok in the sync modal with the progress bar indications.

https://forum.syncthing.net/t/i-want-to-get-rid-of-the-syncing-status/13793/19
> Do you think it is possible to have any kind of progress indication? e.g. 10 of 100 blocks/database entries checked = display “Preparing (10%)”
> My wish is to avoid a new state that stays there for a long(er) time without (again) showing that Syncthing is doing black magic important stuff under the hood

What do you think about including the amount of bytes to sync in the "preparing" state (same as in syncing)? I removed it because I don't think the number is very insightful (at the beginning it includes both to be fast-tracked files and such that actually need copying/pulling), however it still gives some kind of progress indication (the number will shrink with fast-tracked files) as opposed to nothing at all.

gui/default/syncthing/core/syncthingController.js
uok: I'm for "Preparing" or "Preparing to sync" (small "s" like "Waiting to scan")

uok:
I agree that the number of bytes is not helpful. 
Let's go without any progress for now and see how it works in the wild.

gui/default/syncthing/core/syncthingController.js
imsodin: I went for `sync-preparing` as I discovered a precedent for that in `scan-waiting` (and I really don't think the variable name matters that much).

And on the UI side I'd like to keep the connection to syncing - it's not just preparing anything, it's preparing a sync run. And regarding capitalization: Currently "Waiting to scan" is the outlier, everything else is in title case. Feel free to start a forum discussion if you'd like to change that (though probably not, as it requires retranslation for a minor detail).

lib/model/folder_sendrecv.go
AudriusButkevicius: Can we not make the folder state lockless with atomics?

AudriusButkevicius:
I am a bit lost what is it that actually takes time here then? We sync file by file, so I assume the wait also happens file by file. Any one file that needs to be synced throws this whole "preparing" concept to the bin, and goes back to the old "oh it days syncing but not doing anything" problem. Is this actually making it generally better? I feel its making it better in a rare edge case.

imsodin:
I am not after "oh it days syncing but not doing anything", I am after "oh it's syncing but it shouldn't, I know my data is already in sync".

And first we go over all files, not doing any copying/pulling, only fast tracking and queuing. That's the phase I am targeting. And I am targeting it, as I did experience really long sync phases when resetting a folder (with data in place/synced) and I believe quite a few reports being confused about syncing when there shouldn't be any on the forum were in this state too - I can't prove it though. As it's a simple enough change, I figured it was worth the shot.

AudriusButkevicius:
That stage is just a bunch of db lookups and array comparisons, how long can that take? I think we walk the db on startup, so presumably less than startup?

lib/model/folder_sendrecv.go
imsodin: Not really, at least we wouldn't get rid of the lock, as it also protects two other `folderState` members, which must be changed atomically together. Making just `current` atomic would allow to take the shortcut return in `setState` without acquiring the lock, but that would be a bug waiting to happen, when someone will assign to `current` while holding the locking without using `atomic`.

imsodin:
Lgtm?

AudriusButkevicius:
Idk, I think this is still half baked, as it will still do syncing with no network activity for large temp files.

But yolo, whatever, it's probably marginally better.

imsodin:
I can change it that it switches between the new "preparing" state and syncing state whenever the puller routine goes idle/active (with e.g. a 1s delay to prevent rapid switching). I considered that but thought it might have more potential to confuse than help, but I am not sure.

AudriusButkevicius:
If @calmh thinks this is ok and useful, lets roll with it.

calmh:
I think it's useful but I haven't reviewed it any since you've been going at it.

PR #6037, opened 2019-09-30
lib/model: Adjust blocks reported in usage reporting (fixes #5995)

### LINKED ISSUE #5995: Usage-reported transfer stats are meaningless
The reported transfer stats use the unit of "blocks" which made sense at the time. Now, however, a block can be any size at all between 128k and 16M, and are slowly getting bigger on average as this is rolled out. Hence the graph is mysteriously shrinking:

Screen Shot 2019-09-04 at 09 32 40

The ratio between the values *might* have some validity (but I'm not 100% sure), but in general for this to be useful we need to switch to byte counts instead.

AudriusButkevicius:


lib/model/folder_sendrecv.go
```diff
@@ -1560,14 +1560,14 @@ func (f *sendReceiveFolder) finisherRoutine(in <-chan *sharedPullerState, dbUpda
 				f.newPullError(state.file.Name, err)
 			} else {
 				blockStatsMut.Lock()
-				blockStats["total"] += state.reused + state.copyTotal + state.pullTotal
-				blockStats["reused"] += state.reused
-				blockStats["pulled"] += state.pullTotal
+				blockStats["total"] += (state.reused + state.copyTotal + state.pullTotal) * (state.file.BlockSize() / protocol.MinBlockSize)
```
calmh: The multiplier seems like a candidate for a variable

Issue #6038, opened 2019-10-01
Unignoring files only changed on this device after deleting propagates deletion to others


imsodin:
The following problem has come up a few times recently in the forum. The scenario is:

1. Device A creates file F, it gets synced to device B.
2. Device A ignores F and deletes it (e.g. to free up space) - device B still has the valid file.
3. Device A un-ignores F.

Observed: F gets deleted on B.

Expected: F gets synced to A.

We already have a mechanism which achieves the expected behaviour, by dropping all counters in the version from other devices when un-ignoring. So if F was ever modified by another device than A (the one where the ignore-delete-unignore cycle happened), it works, but it fails when there's no such other counters.

imsodin:
Fixed in https://github.com/syncthing/syncthing/commit/a2a14c84245771952b9b1c6cb3eabb207da42186 / https://github.com/syncthing/syncthing/pull/6039 (missed the # before the bug number).

PR #6039, opened 2019-10-01
lib/model: Set empty version when unignoring deleted files (fixes 6038)

imsodin:
### Purpose

Drop the entire version instead of just counters from others when un-ignoring a deleted item. This makes the behaviour consistent, in that files deleted locally do not get deleted on the entire cluster regardless of whether it was ever modified on any other device than the local one.

### Testing

I added a new test for the scenario described in https://github.com/syncthing/syncthing/issues/6038 and had to adjust two existing tests, which explicitly checked for only the local counter being preserved in the vector. They now check that the vector is empty.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -512,12 +512,11 @@ func (f *folder) scanSubdirs(subDirs []string) error {
 					LocalFlags: f.localFlags,
 				}
 				// We do not want to override the global version
-				// with the deleted file. Keeping only our local
-				// counter makes sure we are in conflict with any
-				// other existing versions, which will be resolved
-				// by the normal pulling mechanisms.
+				// with the deleted file. Setting to an empty
+				// version makes sure the file gets in sync on
+				// the following pull.
 				if file.ShouldConflict() {
-					nf.Version = nf.Version.DropOthers(f.shortID)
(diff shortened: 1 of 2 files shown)


calmh:
However... There was a point in the previous behavior, in that a local counter with the others zeroed causes a conflict. This is essential if we have changed the file while it was ignored, but not deleted it. We want that to result in a conflict copy.

With this change (no counters) our file becomes an ancestor of everything. Sure, if we've deleted the file then it's a "null" ancestor and everything else will override it as intended. But if we still have the file and have modified it, it will simply get stomped on.

imsodin:
That's not the case, the change only concerns deletions: If we still have a file locally, we still only drop other counters, resulting in a conflict (or update if there's no other counters). That's covered by `TestPullInvalidIgnored` (specifically the `ignExisting` test case). The difference to deletions is, that it is correct (i.e. no data loss) that there is a conflict, if the file changed locally and on any remote, and a simple update if it only ever changed locally, while I consider it incorrect, and data loss, that a file that only ever changed locally and was deleted while ignored, is deleted on the entire cluster upon un-ignoring.

calmh:
Right, I didn't read properly, sorry.

PR #6047, opened 2019-10-01
lib/syncthing: Prevent hangup on error during startup (fixes #6043)

### LINKED ISSUE #6043: Cannot exit after failed startup
If syncthing encounters a fatal error during startup (for example it cannot listen on the API/GUI port) it should exit. Currently it does not, and also it cannot be made to exit after this point, ignoring Ctrl-C. It requires a hard kill to get rid of the process.

imsodin:
There was already some discussion in IRC, for context:

> 15:32:21 - imsodin: The current mechanism is built around the idea of the whole thing as being a "one-shot" operation. I.e. you start it and then wait for an error, therefore no cleaning up is being done and onces used. And that's mostly due to racyness, as e.g. the api can call `Stop` before `startup` returns. However's first at stopping/producing an error gets to do that. My approach at fixing it would have been to close `stopped` at creation, and re-create it after successful startup.
> 15:45:11 - calmh: we can think about it. i first thought it was an emergency, but it's at least a bit cornercasey so no panic to fix
> 15:47:56 - calmh: however i'm thinking that API wise i'm not super excited about things that start routines in the background and may or may not have done its thing when it returns
> 15:48:22 - calmh: i'd probably rather have seen a synchronous Run() method that blocks (essentially ending in a mainSupervisor.Serve() call)
> 15:48:39 - calmh: and then being able to call Stop() on that thing to make Run() return
> 15:49:16 - calmh: with an API promise that we call Run() and Stop() exactly once each
> 15:49:31 - calmh: obviating the need for sync.Once stuff etc

I then wrote a test to reproduce it and while at it some other behaviour. Besides fixing issues found by the test (error reporting), I added
 - the proposed promise that `Start` and `Run` should be called once only to the api, removing the need for `startOnce`. 
 - an error return value to `Start`.
 - removed a useless check in the `stopOnce`, which is always true (default value is `ExitSuccess` and `a.exitStatus` is only changed within the condition).

I believe `stopOnce` is still required, because even if `syncthing.App.Stop` is promised to be called once only, the app can still be closed from the rest api (and potentially in the future internal failure) - which needs to be race-protected.

And in `cmd/syncthing` I still use a `Start`/`Wait` combi instead of `Run`, as we need to know when the API is running before starting the browser.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -647,7 +647,9 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 		}
 	}
 
-	app.Start()
+	if err := app.Start(); err != nil {
+		os.Exit(syncthing.ExitError)
+	}
 
 	cleanConfigDirectory()
 
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -648,7 +648,7 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 	}
 
 	if err := app.Start(); err != nil {
-		os.Exit(syncthing.ExitError)
+		os.Exit(int(syncthing.ExitError))
 	}
 
 	cleanConfigDirectory()
--- a/lib/syncthing/syncthing.go
+++ b/lib/syncthing/syncthing.go
@@ -92,17 +92,9 @@ func New(cfg config.Wrapper, ll *db.Lowlevel, evLogger events.Logger, cert tls.C
 	return a
 }
 
-// Run does the same as Start, but then does not return until the app stops. It
-// is equivalent to calling Start and then Wait.
-// Either Run or Start may be called once only.
-func (a *App) Run() ExitStatus {
-	a.Start()
-	return a.Wait()
-}
-
 // Start executes the app and returns once all the startup operations are done,
 // e.g. the API is ready for use.
-// Either Run or Start may be called once only.
+// Must be called once only.
 func (a *App) Start() error {
 	if err := a.startup(); err != nil {
 		a.stopWithErr(ExitError, err)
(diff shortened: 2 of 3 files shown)


cmd/syncthing/main.go
```diff
@@ -647,7 +647,9 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 		}
 	}
 
-	app.Start()
+	if err := app.Start(); err != nil {
+		os.Exit(syncthing.ExitError)
```
golangcibot: cannot use syncthing.ExitError (constant 1 of type syncthing.ExitStatus) as int value in argument to os.Exit (from `typecheck`)

lib/syncthing/syncthing.go
```diff
@@ -358,6 +360,8 @@ func (a *App) startup() error {
 }
 
 func (a *App) run() {
+	a.stopped = make(chan struct{})
```
calmh: This is a race, it happens concurrently with Wait()

imsodin:
Completely forgot about this PR - should have gone into the RC. This is ready from my side.

lib/syncthing/syncthing.go
```diff
@@ -89,25 +88,29 @@ func New(cfg config.Wrapper, ll *db.Lowlevel, evLogger events.Logger, cert tls.C
...
 	}
+	close(a.stopped) // Hasn't been started, so shouldn't block on Wait.
+	return a
 }
 
-// Run does the same as start, but then does not return until the app stops. It
+// Run does the same as Start, but then does not return until the app stops. It
 // is equivalent to calling Start and then Wait.
+// Either Run or Start may be called once only.
 func (a *App) Run() ExitStatus {
```
calmh: Run() is unused, apart from a test that just checks that it does roughly what it says on the tin. I think we could do without it?

PR #6053, opened 2019-10-03
lib/db: Always use small db settings on 32 bit archs

calmh:
### Purpose

Avoid apparent issues with large db settings on 32 bit archs.

(This is probably a goleveldb bug but I don't have the time/inclination to troubleshoot it.)

### Testing

⁉️


uok:
Thanks! 
If other users come here, for reference see https://forum.syncthing.net/t/many-small-ldb-files-with-database-tuning/13842

PR #6054, opened 2019-10-03
all, lib/logger: Refactor SetDebug calls

lukaslihotzki:
There are a lot of similar SetDebug calls which can be integrated into NewFacility. This pull request should not change behavior. I tested it with some, but not all facilities.

### DIFF
--- a/lib/api/debug.go
+++ b/lib/api/debug.go
@@ -7,9 +7,6 @@
 package api
 
 import (
-	"os"
-	"strings"
-
 	"github.com/syncthing/syncthing/lib/logger"
 )
 
@@ -24,5 +21,7 @@ func shouldDebugHTTP() bool {
 func init() {
 	// The debug facility was originally named "http", changed in:
 	// https://github.com/syncthing/syncthing/pull/5548
-	l.SetDebug("api", strings.Contains(os.Getenv("STTRACE"), "api") || strings.Contains(os.Getenv("STTRACE"), "http") || os.Getenv("STTRACE") == "all")
+	if l.IsTraced("http") {
+		l.SetDebug("api", true)
+	}
 }
--- a/lib/fs/debug.go
+++ b/lib/fs/debug.go
@@ -7,9 +7,6 @@
 package fs
 
 import (
-	"os"
-	"strings"
-
 	"github.com/syncthing/syncthing/lib/logger"
 )
 
@@ -19,11 +16,7 @@ var (
 
 func init() {
 	logger.DefaultLogger.NewFacility("walkfs", "Filesystem access while walking")
-	switch {
-	case strings.Contains(os.Getenv("STTRACE"), "walkfs") || os.Getenv("STTRACE") == "all":
-		l.SetDebug("walkfs", true)
-		fallthrough
-	case strings.Contains(os.Getenv("STTRACE"), "fs"):
+	if logger.DefaultLogger.ShouldDebug("walkfs") {
 		l.SetDebug("fs", true)
 	}
 }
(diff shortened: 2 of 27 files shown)


lib/fs/debug.go
```diff
@@ -19,11 +16,7 @@ var (
 
 func init() {
 	logger.DefaultLogger.NewFacility("walkfs", "Filesystem access while walking")
-	switch {
-	case strings.Contains(os.Getenv("STTRACE"), "walkfs") || os.Getenv("STTRACE") == "all":
-		l.SetDebug("walkfs", true)
-		fallthrough
-	case strings.Contains(os.Getenv("STTRACE"), "fs"):
+	if logger.DefaultLogger.ShouldDebug("walkfs") {
```
AudriusButkevicius: Just setdebug(... ShouldDebug())

AudriusButkevicius: Also, do we need this? Because "fs" is in "walkfs"

lib/sync/debug.go
```diff
@@ -23,13 +22,11 @@ var (
 	// We make an exception in this package and have an actual "if debug { ...
 	// }" variable, as it may be rather performance critical and does
 	// nonstandard things (from a debug logging PoV).
-	debug       = strings.Contains(os.Getenv("STTRACE"), "sync") || os.Getenv("STTRACE") == "all"
+	debug       = logger.DefaultLogger.IsTraced("sync")
```
AudriusButkevicius: l.ShouldDebug()?

lib/api/debug.go
```diff
@@ -24,5 +21,7 @@ func shouldDebugHTTP() bool {
 func init() {
 	// The debug facility was originally named "http", changed in:
 	// https://github.com/syncthing/syncthing/pull/5548
-	l.SetDebug("api", strings.Contains(os.Getenv("STTRACE"), "api") || strings.Contains(os.Getenv("STTRACE"), "http") || os.Getenv("STTRACE") == "all")
+	if l.IsTraced("http") {
```
AudriusButkevicius: I think we can remove this these days.

lib/logger/logger.go
```diff
@@ -50,6 +50,7 @@ type Logger interface {
 	Warnf(format string, vals ...interface{})
 	ShouldDebug(facility string) bool
 	SetDebug(facility string, enabled bool)
+	IsTraced(facility string) bool
```
AudriusButkevicius: Based on the comments I think we could get rid of this

lib/api/debug.go
imsodin: I would keep it: The change isn't that old and it doesn't really hurt.

lib/fs/debug.go
lukaslihotzki: `l.SetDebug("fs", logger.DefaultLogger.ShouldDebug("walkfs"))` would disable `fs` tracing if `walkfs` is not traced.

Relying on "fs" being in "walkfs" feels like a hack. Also, [postSystemDebug](https://github.com/syncthing/syncthing/blob/master/lib/api/api.go#L615) can set just `walkfs` without setting `fs`. Sadly, this code is run just once before postSystemDebug.

lukaslihotzki:
Somehow the last commit failed CI. I don't know why. Everything seems fine locally.

imsodin:
That's the coverage runner that somehow flakes out on TestCSRFRequired every now and then, nothing to worry about.

lib/api/debug.go
AudriusButkevicius: But removing it does not cause damage

calmh:
This is a rare beauty, having a -10x line ratio. Good stuff.

Screen Shot 2019-10-03 at 20 09 04

PR #6061, opened 2019-10-04
gui: better info for stalled and lengthy scans (fixes #5627)

### LINKED ISSUE #5627: Stalled scans need better UI representation
Currently, a stalled scan keeps increasing Remaining Time on a folder, which leads to ridiculous amounts displayed: 

Should be `Unknown (stalled)` instead.

Also, for scans not completely stalled but really too slow `> 1 year` should suffice. Suggested in [this forum thread](https://forum.syncthing.net/t/remaining-time-ui/13086/5).

nekr0z:
Fixes #5627 

### Testing

Not really.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,8 @@ require (
 	github.com/maruel/panicparse v1.3.0
 	github.com/mattn/go-isatty v0.0.9
 	github.com/minio/sha256-simd v0.1.1
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+	github.com/modern-go/reflect2 v1.0.1 // indirect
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
I'd drop it to 1 month cap rather than 1 year, otherwise it looks ok to me.

I don't think anyone will use software if it takes more than a month to be ready.

nekr0z:
> I'd drop it to 1 month cap

I was basing on what @calmh said on the forum thread, and he mentioned a year. If a month is more sutable, I'm all for it.

AudriusButkevicius:
@calmh ?

calmh:
Arbitrary time interval is arbitrary, no big deal.

PR #6074, opened 2019-10-09
lib/fs: add case-insensitive fakefs

nekr0z:
### Purpose

Yep, "a little step for a man" in the direction of #2739.

### DIFF
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -570,10 +570,14 @@ func (fs *fakefs) URI() string {
 }
 
 func (fs *fakefs) SameFile(fi1, fi2 FileInfo) bool {
+	var ok bool
 	if fs.insens {
-		return UnicodeLowercase(fi1.Name()) == UnicodeLowercase(fi2.Name())
+		ok = (UnicodeLowercase(fi1.Name()) == UnicodeLowercase(fi2.Name()))
+	} else {
+		ok = (fi1.Name() == fi2.Name())
 	}
-	return fi1.Name() == fi2.Name()
+
+	return ok && fi1.ModTime() == fi2.ModTime()
 }
 
 // fakeFile is the representation of an open file. We don't care if it's
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -507,9 +507,8 @@ func (fs *fakefs) Rename(oldname, newname string) error {
 	dst, ok := p1.children[newKey]
 	if ok {
 		if fs.insens && newKey == oldKey {
-			// case-only in-place rename
-			entry.name = filepath.Base(newname)
-			return nil
+			// case-only in-place rename (gives os.ErrExist on OSX)
+			return os.ErrExist
 		}
 
 		if dst.entryType == fakeEntryTypeDir {
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -580,7 +580,7 @@ func (fs *fakefs) SameFile(fi1, fi2 FileInfo) bool {
 		ok = fi1.Name() == fi2.Name()
 	}
 
-	return ok && fi1.ModTime() == fi2.ModTime() && fi1.Mode() == fi2.Mode() && fi1.IsDir() == fi2.IsDir() && fi1.IsRegular() == fi2.IsRegular() && fi1.IsSymlink() == fi2.IsSymlink() && fi1.Owner() == fi2.Owner() && fi1.Group() == fi2.Group()
+	return ok && fi1.ModTime().Equal(fi2.ModTime()) && fi1.Mode() == fi2.Mode() && fi1.IsDir() == fi2.IsDir() && fi1.IsRegular() == fi2.IsRegular() && fi1.IsSymlink() == fi2.IsSymlink() && fi1.Owner() == fi2.Owner() && fi1.Group() == fi2.Group()
 }
 
 // fakeFile is the representation of an open file. We don't care if it's
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -171,7 +171,6 @@ func (fs *fakefs) entryForName(name string) *fakeEntry {
 		if entry.entryType != fakeEntryTypeDir {
 			return nil
 		}
-
 		var ok bool
 		entry, ok = entry.children[comp]
 		if !ok {
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -507,8 +507,9 @@ func (fs *fakefs) Rename(oldname, newname string) error {
 	dst, ok := p1.children[newKey]
 	if ok {
 		if fs.insens && newKey == oldKey {
-			// case-only in-place rename (gives os.ErrExist on OSX)
-			return os.ErrExist
+			// case-only in-place rename
+			entry.name = filepath.Base(newname)
+			return nil
 		}
 
 		if dst.entryType == fakeEntryTypeDir {
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -258,6 +258,9 @@ func (fs *fakefs) Create(name string) (File, error) {
 	if err != nil {
 		return nil, err
 	}
+	if fs.insens {
+		return &fakeFile{fakeEntry: entry, presentedName: filepath.Base(name)}, nil
+	}
 	return &fakeFile{fakeEntry: entry}, nil
 }
 
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -227,6 +227,9 @@ func (fs *fakefs) create(name string) (*fakeEntry, error) {
 		entry.size = 0
 		entry.mtime = time.Now()
 		entry.mode = 0666
+		if fs.insens {
+			entry.name = filepath.Base(name)
+		}
 		return entry, nil
 	}
 
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -314,15 +314,14 @@ func (fs *fakefs) Mkdir(name string, perm FileMode) error {
 	entry := fs.entryForName(dir)
 	key := base
 
-	if fs.insens {
-		key = UnicodeLowercase(key)
-	}
-
 	if entry == nil {
 		return os.ErrNotExist
 	}
 	if entry.entryType != fakeEntryTypeDir {
 		return os.ErrExist
+	}
+	if fs.insens {
+		key = UnicodeLowercase(key)
 	}
 	if _, ok := entry.children[key]; ok {
 		return os.ErrExist
@@ -398,16 +397,15 @@ func (fs *fakefs) OpenFile(name string, flags int, mode FileMode) (File, error) 
 	entry := fs.entryForName(dir)
 	key := base
 
-	if fs.insens {
-		key = UnicodeLowercase(key)
-	}
-
 	if entry == nil {
 		return nil, os.ErrNotExist
 	} else if entry.entryType != fakeEntryTypeDir {
 		return nil, errors.New("not a directory")
 	}
 
+	if fs.insens {
+		key = UnicodeLowercase(key)
+	}
 	if flags&os.O_EXCL != 0 {
 		if _, ok := entry.children[key]; ok {
 			return nil, os.ErrExist
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -227,9 +227,6 @@ func (fs *fakefs) create(name string) (*fakeEntry, error) {
 		entry.size = 0
 		entry.mtime = time.Now()
 		entry.mode = 0666
-		if fs.insens {
-			//entry.name = filepath.Base(name)
-		}
 		return entry, nil
 	}
 
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -511,18 +511,17 @@ func (fs *fakefs) Rename(oldname, newname string) error {
 		if dst.entryType == fakeEntryTypeDir {
 			return errors.New("is a directory")
 		}
-
-		if fs.insens && newKey == oldKey {
-			// case-only in-place rename
-			entry.name = filepath.Base(newname)
-			return nil
-		}
 	}
 
 	p1.children[newKey] = entry
 	entry.name = filepath.Base(newname)
 
-	delete(p0.children, oldKey)
+	// don't want to remove the entry for in-place case-only rename
+	// on an insensitive FS
+	if !(p0 == p1 && oldKey == newKey && fs.insens) {
+		delete(p0.children, oldKey)
+	}
+
 	return nil
 }
 
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -506,8 +506,12 @@ func (fs *fakefs) Rename(oldname, newname string) error {
 
 	dst, ok := p1.children[newKey]
 	if ok {
-		// in-place case-only directory renames will fail on insensitive systems
-		// they do fail on OS X, so that's expected
+		if fs.insens && newKey == oldKey {
+			// case-only in-place rename
+			entry.name = filepath.Base(newname)
+			return nil
+		}
+
 		if dst.entryType == fakeEntryTypeDir {
 			return errors.New("is a directory")
 		}
@@ -516,11 +520,7 @@ func (fs *fakefs) Rename(oldname, newname string) error {
 	p1.children[newKey] = entry
 	entry.name = filepath.Base(newname)
 
-	// don't want to remove the entry for in-place case-only rename
-	// on an insensitive FS
-	if !(p0 == p1 && oldKey == newKey && fs.insens) {
-		delete(p0.children, oldKey)
-	}
+	delete(p0.children, oldKey)
 
 	return nil
 }
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -578,9 +578,9 @@ func (fs *fakefs) SameFile(fi1, fi2 FileInfo) bool {
 	// where ModTime is not that precise
 	var ok bool
 	if fs.insens {
-		ok = (UnicodeLowercase(fi1.Name()) == UnicodeLowercase(fi2.Name()))
+		ok = UnicodeLowercase(fi1.Name()) == UnicodeLowercase(fi2.Name())
 	} else {
-		ok = (fi1.Name() == fi2.Name())
+		ok = fi1.Name() == fi2.Name()
 	}
 
 	return ok && fi1.ModTime() == fi2.ModTime() && fi1.Mode() == fi2.Mode() && fi1.IsDir() == fi2.IsDir() && fi1.IsRegular() == fi2.IsRegular() && fi1.IsSymlink() == fi2.IsSymlink() && fi1.Owner() == fi2.Owner() && fi1.Group() == fi2.Group()
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -570,6 +570,10 @@ func (fs *fakefs) URI() string {
 }
 
 func (fs *fakefs) SameFile(fi1, fi2 FileInfo) bool {
+	// BUG: real systems base file sameness on path, inodes, etc
+	// we try our best, but FileInfo just doesn't have enough data
+	// so there be false positives, especially on Windows
+	// where ModTime is not that precise
 	var ok bool
 	if fs.insens {
 		ok = (UnicodeLowercase(fi1.Name()) == UnicodeLowercase(fi2.Name()))
@@ -577,7 +581,7 @@ func (fs *fakefs) SameFile(fi1, fi2 FileInfo) bool {
 		ok = (fi1.Name() == fi2.Name())
 	}
 
-	return ok && fi1.ModTime() == fi2.ModTime()
+	return ok && fi1.ModTime() == fi2.ModTime() && fi1.Mode() == fi2.Mode() && fi1.IsDir() == fi2.IsDir() && fi1.IsRegular() == fi2.IsRegular() && fi1.IsSymlink() == fi2.IsSymlink() && fi1.Owner() == fi2.Owner() && fi1.Group() == fi2.Group()
 }
 
 // fakeFile is the representation of an open file. We don't care if it's
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -468,7 +468,7 @@ func (fs *fakefs) RemoveAll(name string) error {
 
 	entry := fs.entryForName(filepath.Dir(name))
 	if entry == nil {
-		return os.ErrNotExist
+		return nil // all tested real systems exibit this behaviour
 	}
 
 	// RemoveAll is easy when the file system uses garbage collection under
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -504,6 +504,7 @@ func (fs *fakefs) Rename(oldname, newname string) error {
 	dst, ok := p1.children[newKey]
 	if ok {
 		if fs.insens && newKey == oldKey {
+			// case-only in-place rename
 			entry.name = filepath.Base(newname)
 			return nil
 		}
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -506,14 +506,16 @@ func (fs *fakefs) Rename(oldname, newname string) error {
 
 	dst, ok := p1.children[newKey]
 	if ok {
+		// in-place case-only directory renames will fail on insensitive systems
+		// they do fail on OS X, so that's expected
+		if dst.entryType == fakeEntryTypeDir {
+			return errors.New("is a directory")
+		}
+
 		if fs.insens && newKey == oldKey {
 			// case-only in-place rename
 			entry.name = filepath.Base(newname)
 			return nil
-		}
-
-		if dst.entryType == fakeEntryTypeDir {
-			return errors.New("is a directory")
 		}
 	}
 
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -228,7 +228,7 @@ func (fs *fakefs) create(name string) (*fakeEntry, error) {
 		entry.mtime = time.Now()
 		entry.mode = 0666
 		if fs.insens {
-			entry.name = filepath.Base(name)
+			//entry.name = filepath.Base(name)
 		}
 		return entry, nil
 	}
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -155,6 +155,9 @@ type fakeEntry struct {
 
 func (fs *fakefs) entryForName(name string) *fakeEntry {
 	// bug: lookup doesn't work through symlinks.
+	if fs.insens {
+		name = UnicodeLowercase(name)
+	}
 
 	name = filepath.ToSlash(name)
 	if name == "." || name == "/" {
@@ -170,11 +173,7 @@ func (fs *fakefs) entryForName(name string) *fakeEntry {
 		}
 
 		var ok bool
-		if fs.insens {
-			entry, ok = entry.children[UnicodeLowercase(comp)]
-		} else {
-			entry, ok = entry.children[comp]
-		}
+		entry, ok = entry.children[comp]
 		if !ok {
 			return nil
 		}
--- a/lib/fs/fakefs_test.go
+++ b/lib/fs/fakefs_test.go
@@ -728,7 +728,8 @@ func TestFakeFSSameFile(t *testing.T) {
 		want bool
 	}{
 		{f1: "Bar", f2: "Baz", want: false},
-		{f1: "Bar", f2: "/Foo/Bar", want: true},
+		{f1: "Bar", f2: "/Foo/Bar", want: false},
+		{"Bar", "Bar", true},
 	}
 
 	for _, test := range testCases {
@@ -754,7 +755,8 @@ func testFakeFSSameFileInsens(t *testing.T, fs Filesystem) {
 		want bool
 	}{
 		{f1: "bAr", f2: "baZ", want: false},
-		{f1: "baR", f2: "/fOO/bAr", want: true},
+		{f1: "baR", f2: "/fOO/bAr", want: false},
+		{"baz", "BAZ", true},
 	}
 
 	for _, test := range testCases {
@@ -787,10 +789,14 @@ func testFakeFSCreateInsens(t *testing.T, fs Filesystem) {
 		t.Fatal(err)
 	}
 
+	defer fd1.Close()
+
 	fd2, err := fs.Create("fOo")
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	defer fd2.Close()
 
 	if fd2.Name() != "fOo" {
 		t.Errorf("name of created file \"fOo\" is %s", fd2.Name())
@@ -799,9 +805,6 @@ func testFakeFSCreateInsens(t *testing.T, fs Filesystem) {
 	if fd1.Name() != "fOo" {
 		t.Errorf("name of the file created as \"FOO\" is %s", fd1.Name())
 	}
-
-	fd1.Close()
-	fd2.Close()
 }
 
 func cleanup(fs Filesystem) error {
--- a/lib/fs/fakefs_test.go
+++ b/lib/fs/fakefs_test.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"reflect"
 	"runtime"
 	"sort"
@@ -201,21 +202,24 @@ func TestFakeFSCaseSensitive(t *testing.T) {
 	}
 
 	if runtime.GOOS == "linux" {
-		if err := os.Mkdir("test-tmp", 0755); err != nil {
-			t.Fatalf("could not create temporary dir for testing")
+		testDir, err := ioutil.TempDir("", "")
+		if err != nil {
+			t.Fatalf("could not create temporary dir for testing: %s", err)
 		}
 
-		if _, err := os.Create("test-tmp/.stfolder"); err != nil {
-			t.Fatalf("could not create .stfolder")
+		if fd, err := os.Create(filepath.Join(testDir, ".stfolder")); err != nil {
+			t.Fatalf("could not create .stfolder: %s", err)
+		} else {
+			fd.Close()
 		}
 
 		defer func() {
-			if err := os.RemoveAll("test-tmp"); err != nil {
-				t.Fatalf("could not remove test-tmp directory")
+			if err := os.RemoveAll(testDir); err != nil {
+				t.Fatalf("could not remove test-tmp directory: %s", err)
 			}
 		}()
 
-		filesystems = append(filesystems, testFS{runtime.GOOS, newBasicFilesystem("test-tmp")})
+		filesystems = append(filesystems, testFS{runtime.GOOS, newBasicFilesystem(testDir)})
 	}
 
 	runTests(t, tests, filesystems)
@@ -242,21 +246,24 @@ func TestFakeFSCaseInsensitive(t *testing.T) {
 	}
 
 	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
-		if err := os.Mkdir("test-tmp", 0755); err != nil {
-			t.Fatalf("could not create temporary dir for testing")
+		testDir, err := ioutil.TempDir("", "")
+		if err != nil {
+			t.Fatalf("could not create temporary dir for testing: %s", err)
 		}
 
-		if _, err := os.Create("test-tmp/.stfolder"); err != nil {
-			t.Fatalf("could not create .stfolder")
+		if fd, err := os.Create(filepath.Join(testDir, ".stfolder")); err != nil {
+			t.Fatalf("could not create .stfolder: %s", err)
+		} else {
+			fd.Close()
 		}
 
 		defer func() {
-			if err := os.RemoveAll("test-tmp"); err != nil {
-				t.Fatalf("could not remove test-tmp directory")
+			if err := os.RemoveAll(testDir); err != nil {
+				t.Fatalf("could not remove test-tmp directory: %s", err)
 			}
 		}()
 
-		filesystems = append(filesystems, testFS{runtime.GOOS, newBasicFilesystem("test-tmp")})
+		filesystems = append(filesystems, testFS{runtime.GOOS, newBasicFilesystem(testDir)})
 	}
 
 	runTests(t, tests, filesystems)
--- a/lib/fs/fakefs_test.go
+++ b/lib/fs/fakefs_test.go
@@ -191,6 +191,7 @@ type test struct {
 
 func TestFakeFSCaseSensitive(t *testing.T) {
 	var tests = []test{
+		{"OpenFile", testFakeFSOpenFile},
 		{"RemoveAll", testFakeFSRemoveAll},
 		{"Remove", testFakeFSRemove},
 		{"SameFile", testFakeFSSameFile},
@@ -614,9 +615,7 @@ func testFakeFSMkdirInsens(t *testing.T, fs Filesystem) {
 	}
 }
 
-func TestFakeFSOpenFile(t *testing.T) {
-	fs := newFakeFilesystem("/openf")
-
+func testFakeFSOpenFile(t *testing.T, fs Filesystem) {
 	if _, err := fs.OpenFile("foobar", os.O_RDONLY, 0664); err == nil {
 		t.Errorf("got no error opening a non-existing file")
 	}
--- a/lib/fs/fakefs_test.go
+++ b/lib/fs/fakefs_test.go
@@ -17,7 +17,6 @@ import (
 	"runtime"
 	"sort"
 	"testing"
-	"time"
 )
 
 func TestFakeFS(t *testing.T) {
@@ -800,9 +799,6 @@ func testFakeFSSameFile(t *testing.T, fs Filesystem) {
 			t.Fatalf("Could not create %s: %s", filename, err)
 		} else {
 			fd.Close()
-			if runtime.GOOS == "windows" {
-				time.Sleep(1 * time.Millisecond)
-			}
 		}
 	}
 
--- a/lib/fs/fakefs_test.go
+++ b/lib/fs/fakefs_test.go
@@ -199,6 +199,7 @@ func TestFakeFSCaseSensitive(t *testing.T) {
 		{"Mkdir", testFakeFSMkdir},
 		{"SameFile", testFakeFSSameFile},
 		{"DirNames", testFakeFSDirNames},
+		{"FileName", testFakeFSFileName},
 	}
 	var filesystems = []testFS{
 		{"fakefs", newFakeFilesystem("/foo")},
@@ -438,17 +439,15 @@ func testFakeFSStatInsens(t *testing.T, fs Filesystem) {
 	assertDir(t, fs, "/foo", []string{"aaa"})
 }
 
-func TestFakeFSFileName(t *testing.T) {
+func testFakeFSFileName(t *testing.T, fs Filesystem) {
 	var testCases = []struct {
-		fs     string
 		create string
 		open   string
 	}{
-		{"/foo", "bar", "bar"},
+		{"bar", "bar"},
 	}
 
 	for _, testCase := range testCases {
-		fs := newFakeFilesystem(testCase.fs)
 		if _, err := fs.Create(testCase.create); err != nil {
 			t.Fatal(err)
 		}
--- a/lib/fs/fakefs_test.go
+++ b/lib/fs/fakefs_test.go
@@ -205,7 +205,7 @@ func TestFakeFSCaseSensitive(t *testing.T) {
 
 	if runtime.GOOS == "linux" {
 		testDir := createTestDir(t)
-		defer removeTestDir(t)
+		defer removeTestDir(t, testDir)
 		filesystems = append(filesystems, testFS{runtime.GOOS, newBasicFilesystem(testDir)})
 	}
 
@@ -241,7 +241,7 @@ func TestFakeFSCaseInsensitive(t *testing.T) {
 
 	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
 		testDir := createTestDir(t)
-		defer removeTestDir(t)
+		defer removeTestDir(t, testDir)
 		filesystems = append(filesystems, testFS{runtime.GOOS, newBasicFilesystem(testDir)})
 	}
 
@@ -265,7 +265,7 @@ func createTestDir(t *testing.T) string {
 	return testDir
 }
 
-func removeTestDir(t *testing.T) {
+func removeTestDir(t *testing.T, testDir string) {
 	t.Helper()
 
 	if err := os.RemoveAll(testDir); err != nil {
--- a/lib/fs/fakefs_test.go
+++ b/lib/fs/fakefs_test.go
@@ -575,7 +575,7 @@ func testFakeFSMkdirInsens(t *testing.T, fs Filesystem) {
 func TestFakeFSOpenFile(t *testing.T) {
 	fs := newFakeFilesystem("/openf")
 
-	if _, err := fs.OpenFile("foobar", os.O_RDONLY, 0664); err == nil {
+	if fd, err := fs.OpenFile("foobar", os.O_RDONLY, 0664); err == nil {
 		t.Errorf("got no error opening a non-existing file")
 	}
 
@@ -597,25 +597,35 @@ func TestFakeFSOpenFile(t *testing.T) {
 }
 
 func testFakeFSOpenFileInsens(t *testing.T, fs Filesystem) {
-	if _, err := fs.OpenFile("FooBar", os.O_RDONLY, 0664); err == nil {
+	fd, err := fs.OpenFile("FooBar", os.O_RDONLY, 0664)
+	if err == nil {
 		t.Errorf("got no error opening a non-existing file")
 	}
+	fd.Close()
 
-	if _, err := fs.OpenFile("fOObar", os.O_RDWR|os.O_CREATE, 0664); err != nil {
+	fd, err = fs.OpenFile("fOObar", os.O_RDWR|os.O_CREATE, 0664)
+	if err != nil {
 		t.Fatal(err)
 	}
+	fd.Close()
 
-	if _, err := fs.OpenFile("fOoBaR", os.O_RDWR|os.O_CREATE|os.O_EXCL, 0664); err == nil {
+	fd, err = fs.OpenFile("fOoBaR", os.O_RDWR|os.O_CREATE|os.O_EXCL, 0664)
+	if err == nil {
 		t.Errorf("created an existing file while told not to")
 	}
+	fd.Close()
 
-	if _, err := fs.OpenFile("FoObAr", os.O_RDWR|os.O_CREATE, 0664); err != nil {
+	fd, err = fs.OpenFile("FoObAr", os.O_RDWR|os.O_CREATE, 0664)
+	if err != nil {
 		t.Fatal(err)
 	}
+	fd.Close()
 
-	if _, err := fs.OpenFile("FOOBAR", os.O_RDWR, 0664); err != nil {
+	fd, err = fs.OpenFile("FOOBAR", os.O_RDWR, 0664)
+	if err != nil {
 		t.Fatal(err)
 	}
+	fd.Close()
 }
 
 func TestFakeFSRemoveAll(t *testing.T) {
--- a/lib/fs/fakefs_test.go
+++ b/lib/fs/fakefs_test.go
@@ -809,8 +809,8 @@ func testFakeFSSameFile(t *testing.T, fs Filesystem) {
 		f2   string
 		want bool
 	}{
-		{f1: "Bar", f2: "Baz", want: false},
-		{f1: "Bar", f2: "/Foo/Bar", want: false},
+		{"Bar", "Baz", false},
+		{"Bar", "/Foo/Bar", false},
 		{"Bar", "Bar", true},
 	}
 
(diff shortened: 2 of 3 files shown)


lib/fs/fakefs.go
```diff
@@ -162,8 +168,19 @@ func (fs *fakefs) entryForName(name string) *fakeEntry {
...
 		var ok bool
-		entry, ok = entry.children[comp]
+
+		if fs.insens {
+			for _, child := range entry.children {
+				if strings.EqualFold(child.name, comp) {
+					entry = child
+					ok = true
+				}
+			}
```
AudriusButkevicius: Why not just store stuff lowercased in the first place?

nekr0z: Because a) we'd rather `Stat` returned the filename the way it is capitalized, the way it does on the systems in question, and b) lowercase conversion and case folding are surprisingly different in Unicode.

AudriusButkevicius: Sure, but only the map keys need to be lower, the entries themselves could have capitalized name, and unicode.ToLower is a thing.

calmh: Even `strings.ToLower` specifically says "all Unicode letters mapped to their lower case" which ought to be correct enough. The thing with Stat is that it specifically returns the name you passed to it, not whatever is stored at the filesystem level. Otherwise our lives would be easier.

nekr0z: OK, let me get priorities straight so I have a better idea of what we're trying to achieve.

Please correct me if I'm wrong here, but I somehow got these ideas for basis:
1. we have an issue where we're reliably losing user data
1. `fakefs` is used for testing and doesn't really do anything in production code
3. cycling through the `map` the way I did it is slow, and slow is bad, we want things fast
4. for tests we want to reproduce actual systems' behaviour as accurately as possible, so that we don't later get into the situation where our code passes the test with flying colors but still fails in production

According to the project goals, not losing data is out first and topmost priority. Thus, when choosing between speed/developing convenience (i.e. test speed) and keeping user's data safe we are bound to choose the latter.

I've updated the PR to have real-world-issue example in test. And [here](https://play.golang.org/p/lqKzKotS5G2) is a code snippet that shows there is effectively no way to make this test pass if we use `strings.ToLower`. That's kind of what I meant with my first comment.

I guess what I'm saying is: it's better to make sure `4` above is achieved, even if it means sacrifising `3` (especially in light of `2`). Otherwise, further down the line trying to achieve `1` is a Sisyphean attempt, because we're leaving a gaping loophole at the very first step.

calmh: Sure. Fakefs is a limited simulation of reality for specific purposes; it's performance is not completely irrelevant but almost. That said, if the aim is to use this to work towards a solution where Syncthing works well on case insensitive filesystems, it should simulate at least the relevant aspects of those systems that we care about. One of the very specific things that case insensitive file systems (or at least the Go APIs on top of them) do is lie about the case of things when you stat them, always returning the name you asked for rather than what is actually stored on disk.

So I don't particularly care how fast or slow your map lookup is. It doesn't matter. But returning an entry with the name `Foo` when the stat call was for `fOO` is something that doesn't happen in reality and is something that I think would be worth doing the same in our simulation. It's the typical example of something that could make our tests pass with flying colors yet break completely in the real world. That's all. :)

nekr0z: Point taken. Will work on `stat` part, too (`strings.ToLowerCase` wouldn't help there much, either). Consider this PR a WIP for now. :)

nekr0z: Done. Stat behaves the same way `os.Stat` does on OS X (not tested on Windows, don't have it at home).

AudriusButkevicius: My point still stands, I think you could have the map keys lowercased so you don't need to iterate over the keys

nekr0z: Problem is, I couldn't. `strings.ToLower("ΣΊΣΥΦΟΣ")` is NOT equal to  `strings.ToLower("Σίσυφος")`, and these strings do point to the same file on a case-insensitive FS.

calmh:
```go
func unicodeFoldLower(s string) string {
	rs := []rune(s)
	for i := range rs {
		for r := unicode.SimpleFold(rs[i]); r > rs[i]; r = unicode.SimpleFold(r) {
			rs[i] = r
		}
	}
	return string(rs)
}
```
?

nekr0z: Wow. Just... wow.

My attempt on this func is 10% slower, has twice the line count, and tends to prefer uppercase.

Is it a good idea to make this exported and put into /lib/util (seeing as how it will obviously be reused for case-insensitive indexing)?

nekr0z: Reworked that and added tests for all I could think of.

imsodin: What about the existing `UnicodeLowercase` at https://github.com/syncthing/syncthing/blob/master/lib/fs/folding.go#L13 ? That uses unicode's `ToLower`/`ToUpper`.

lib/fs/fakefs.go
```diff
@@ -163,9 +170,19 @@ func (fs *fakefs) entryForName(name string) *fakeEntry {
...
-		if !ok {
-			return nil
+
+		if fs.insens {
+			entry, ok = entry.children[unicodeFoldLower(comp)]
+			if !ok {
+				return nil
+			}
+
+			entry.name = comp
```
AudriusButkevicius: We might need to copy entry? Because this can change the value after it has been returned.

lib/fs/fakefs.go
nekr0z: Right, I've missed it, thanks for pointing out.

lib/fs/fakefs.go
nekr0z: I can't see where this happens (other than in places where there are already tests to prove things work as expected).

What I did find while looking for those, however, is DirNames that supposedly returns names for all the files in a directory. There are no tests for it, so I missed it initially. Do we use it? Because if we do and if we rely on it, I'd better revert back to my initial approach: I can't fathom a way to make such a function to return file names in their original capitalization if our index is folded :(

AudriusButkevicius:
Dir names would work if you used entry.name oppose to the map key.

nekr0z:
> Dir names would work if you used entry.name oppose to the map key.

They're actually working as expected, because that's exactly what I did without realizing it. :) Tests added to demonstrate and confirm. \

lib/fs/fakefs.go
nekr0z: Ok, judging by the tests it looks to me that everything does work as expected now. Ready for review.

AudriusButkevicius: You don't have a test for this.

The test would be create aaa, stat AAA (you will get AAA) store the result, stat again AaA, check stored AAA, it would become AaA.

nekr0z: I don't see why it would behave the way you described given how code is written, but I've just added a test to be sure it behaves as intended ;)

calmh: The stat result is intentionally a copy, so won't reflect later changes:

https://github.com/syncthing/syncthing/blob/28ad9ba434c49e2b06964a2dbb4022ba470524e5/lib/fs/fakefs.go#L737-L740

but it looks like you're modifying the original, this changing the file name "on disk" by stat:ing it?

AudriusButkevicius: Yeah you'd have to restat the first file handle.

nekr0z: OK, I officially admit I'm too dumb for this. Reading and re-reading the code most carefully what you both are saying makes perfect sense, and this new version of the test should most definitely fail, because, as demonstrated by another test, DirNames does show what's actually in the directory.

Problem is, the test passes. :/

If you have time for this when you come back, I'd really appreciate you showing me where things go wrong. Otherwise, don't bother: wasting too much of maintainers' time is not something I'm aiming to do here.

lib/fs/fakefs.go
```diff
@@ -221,6 +237,10 @@ func (fs *fakefs) create(name string) (*fakeEntry, error) {
 		return entry, nil
 	}
 
+	if fs.insens {
+		name = UnicodeLowercase(name)
```
imsodin: This means the name of the entry will also be lowercase. Don't we want the name on creation to be as given in the argument and just the map key to be lowercase?

lib/fs/fakefs_test.go
```diff
@@ -172,3 +176,181 @@ func TestFakeFSRead(t *testing.T) {
...
+		if _, err := fs.Create("/" + filename); err != nil {
+			t.Fatalf("Could not create %s: %s", filename, err)
+		}
+	}
+
+	assertDir(t, fs, "/", filenames)
+}
+
+func assertDir(t *testing.T, fs *fakefs, directory string, filenames []string) {
+	t.Helper()
```
imsodin: This also isn't a helper.

lib/fs/fakefs.go
imsodin: What exactly do you think should fail?  
Unless I misunderstood something (totally possible :) ) I do see the issue pointed out in this conversation and a few others (see rest of review comments), so I generally agree a failure would be nice/expected. However the nature of the issues (using the map keys or argument as names in the returned values) ensures always consistent (while incorrect) results thus tests don't fail. I don't find a test that triggers the particular failure pointed out here. For that you'd e.g. create `Foo`, stat `foo` and then check that what's "on disk" (in fakefs) is still `Foo` (by manually traversing the fake-entry-maps). However that's not even the case currently as on creation the name is already folded.

As I can't comment on `DirNames` (no change): You need to adapt it to not return the map keys as names, but the actual unfolded names of the map entries.

lib/fs/fakefs.go
```diff
@@ -310,6 +334,10 @@ func (fs *fakefs) Mkdir(name string, perm FileMode) error {
 }
 
 func (fs *fakefs) MkdirAll(name string, perm FileMode) error {
+	if fs.insens {
+		name = UnicodeLowercase(name)
```
imsodin: Same here (maybe more, I'll stop here ;) ).

lib/fs/fakefs.go
```diff
@@ -286,6 +306,10 @@ func (fs *fakefs) Mkdir(name string, perm FileMode) error {
 	fs.mut.Lock()
 	defer fs.mut.Unlock()
 
+	if fs.insens {
+		name = UnicodeLowercase(name)
```
imsodin: Same issue here, this will create an entry with a lowercase name.

nekr0z: @imsodin thank you so much! I think I finally grok my blunder, or at least I'm on the path to do it. Will have to take several hours of excessive caffeine and nicotine intake to get it all straight (not likely to happen on business days), but I'll get there. Thanks again!

imsodin: That doesn't sound like much fun ;) From what I saw during my review I'd say you just need to go over the methods and ensure the following three things:

1. Create fake-entries with the name given as a parameter (not folded).
2. Once a fake-entry is created, no operation may change it's name (except for replacing it with a new one/renaming).
3. When stating change the name on a copy of the entry (should already be copied now as pointed out by Jakob) to be equal to the name given as a parameter.

I believe that should be doable without (overly much) intoxication :D

nekr0z: Yep, I got the general steps already. What I meant by "to get _it all_ straight" was not the code itself (I think I'll straighten it up by tomorrow evening), but some misaligned (in my head) conceptions about how things work, so that I start speaking better Go, so that my future PRs don't result in your three wasting the whole lot of your time trying to talk sense to me. ;-)

nekr0z: OK, I think it's good now. I fixed a couple of tests and wrote some new ones along the way, so that at least everything I've changed is covered. I might have missed something, though, so please do critisize! ;)

lib/fs/fakefs.go
```diff
@@ -163,7 +169,12 @@ func (fs *fakefs) entryForName(name string) *fakeEntry {
 			return nil
 		}
 		var ok bool
-		entry, ok = entry.children[comp]
+
```
imsodin: Empty line bit out of place.

lib/fs/fakefs.go
```diff
@@ -220,7 +231,6 @@ func (fs *fakefs) create(name string) (*fakeEntry, error) {
 		entry.mode = 0666
 		return entry, nil
 	}
-
```
imsodin: Seems weird to remove this empty line while adding two further down around `if fs.insens`. There's a few more emptly lines added/removed which look potentially spurious. I usually just go with what the existing code looks like (in here it seems to be a bit inconsistent, with some places with quite a few empty lines and others almost none - I tend to add them if I feel there's a sensible break (and I'd consider a returned branch as such)).

lib/fs/fakefs.go
```diff
@@ -429,12 +467,20 @@ func (fs *fakefs) Rename(oldname, newname string) error {
 	fs.mut.Lock()
 	defer fs.mut.Unlock()
 
+	oldBase := filepath.Base(oldname)
+	newBase := filepath.Base(newname)
```
imsodin: Maybe `oldKey` and `newKey` to make their purpose even more explicit.

lib/fs/fakefs.go
nekr0z: Oh, regarding whitelines I just tend to do what `wsl` linter tells me to while leaving the existing code be, so feel free to nitpick. If you feel like there are ways to impove code readability, I should most definitely listen and do what's right.

lib/fs/fakefs.go
nekr0z: Fair point, will adjust.

lib/fs/fakefs.go
nekr0z: Yep. Fixed.

lib/fs/fakefs.go
```diff
@@ -279,13 +295,18 @@ func (fs *fakefs) Lstat(name string) (FileInfo, error) {
 	if entry == nil {
 		return nil, os.ErrNotExist
 	}
-	return &fakeFileInfo{*entry}, nil
+
+	info := &fakeFileInfo{*entry}
+	if fs.insens {
+		info.name = filepath.Base(name)
```
AudriusButkevicius: We didn't do the base stuff before. Why do we need it but just in this special case?

lib/fs/fakefs.go
```diff
@@ -299,7 +320,12 @@ func (fs *fakefs) Mkdir(name string, perm FileMode) error {
 		return os.ErrExist
 	}
 
-	entry.children[base] = &fakeEntry{
+	key := base
+	if fs.insens {
+		key = UnicodeLowercase(key)
+	}
```
AudriusButkevicius: This breaks "already exists" for case insensitive cases. Suggest you add a test.

lib/fs/fakefs.go
nekr0z: Because we need `Lstat` to return the name exactly as it was asked. For case-sensitive filesystems it's the only way that can happen anyway, but for case-insensitive ones we explicitly want returned `fakeFileInfo.name` to be whatever was asked for. Could be an `if` here, but since case-sensitive will work anyway, I opted to spare an unnecessary condition. Can add it if you think that would be better.

lib/fs/fakefs.go
```diff
@@ -376,6 +411,10 @@ func (fs *fakefs) OpenFile(name string, flags int, mode FileMode) (File, error)
 		mtime: time.Now(),
 	}
 
+	if fs.insens {
+		base = UnicodeLowercase(base)
+	}
```
AudriusButkevicius: The exists check is not taken into account

lib/fs/fakefs.go
```diff
@@ -429,12 +468,20 @@ func (fs *fakefs) Rename(oldname, newname string) error {
 	fs.mut.Lock()
 	defer fs.mut.Unlock()
 
+	oldKey := filepath.Base(oldname)
```
AudriusButkevicius: RemoveAll above needs case folding. Please add a test.

lib/fs/fakefs.go
nekr0z: You're right. Wrote a test, will work on it.

lib/fs/fakefs.go
nekr0z: Oh, thanks for pointing it out. There also appears to be a lovely race-condition in this function, so a thorough test is needed anyway. Will work on it.

lib/fs/fakefs.go
nekr0z: Done. Thanks for pointing out!

lib/fs/fakefs.go
```diff
@@ -444,13 +499,15 @@ func (fs *fakefs) Rename(oldname, newname string) error {
 		return os.ErrNotExist
 	}
 
-	dst, ok := p1.children[filepath.Base(newname)]
+	dst, ok := p1.children[newKey]
 	if ok && dst.entryType == fakeEntryTypeDir {
```
calmh: I think this results in error `is a directory` when doing a case-only rename of an existing directory?

nekr0z: Yep, wrote a failing test, expect commits later today.

nekr0z: Actually, here you go, done.

AudriusButkevicius:
SameFile not addressed.
Rename lowercases whole path, RemoveAll does not and just lowercases the last component, Remove doesn't either.

Sorry, but this is a very inconsistent implementation. If you did lowercasing of the whole path as the first thing in every function I'd have much confidence this actually works.

nekr0z:
Rename just lowercases the last component, as it should, relying on entryForName to get through the path (because we don't really want to rename half the path on the way, do we?)

Remove I've missed, and it is not likely to work, will test and fix it, thank you. Same goes for SameFile, but I'm genuinely confused with this one: is it really supposed to return `true` whenever two files have the same name, even if they are in different directories, have different length, different contents, etc? Because that's what it seems to be doing...

nekr0z:
> If you did lowercasing of the whole path as the first thing in every function I'd have much confidence this actually works.

If won't work for *every* function. Let's see:
* entryForName - you're welcome
* Chmod - no need to touch at all
* Lchown - same
* Chtimes - same
* create - we *do* want the entry created with given name, not the folded one
* Create - no need to touch
* CreateSymlink - same
* DirNames - same
* Lstat - we want to return what was asked
* Mkdir - same as create, we want the created name not folded
* MkdirAll - same
* Open - need to return what was asked for, same as Lstat
* OpenFile - same as create and Mkdir
* ReadSymlink - no need to touch
* Remove - you're welcome
* RemoveAll -you're welcome
* Rename - same way as create, Mkdir and OpenFile, need to preserve unfolded name
* Stat - no need to touch
* SameFile - you're welcome?
other methods for fakefs are really stubs, nothing to be done there.

For fakeFile, Name and Stat are taken care of, the rest have nothing to do with naming.

AudriusButkevicius:
After more thorough review I think what you are doing is required, but the following really confuses me:

For example, create is incredibly confusing, because you try to get the file as is, which abuses the fact that entryForName folds the case.
If that does not exist, you get a dir, which again, non-transparently folds the name internally via entryForName, yet for the actual child name, you fold the name externally, so the lowercasing happen in 3 different stacks which is incredibly hard to follow mentally, for a reviewer.



Issues that I see:
* create - in case the file already exists returns a file with different case than asked for I believe, Create does not handle that.
* Rename - I am not sure why we have `if fs.insens && newKey == oldKey` key equality check? Also, feels that `is directory` check should happen before us fiddling with some struct referenced by some map?

AudriusButkevicius:
I guess it would be interesting to run the same tests you wrote against a real case insensitive filesystem to validate that the assumptions made in the tests actually matches the reality.

nekr0z:
> the lowercasing happen in 3 different stacks which is incredibly hard to follow mentally, for a reviewer.

I totally get your frustration, and wish I could have done things easier and more straightforward, but either I'm too dumb for that, or those insensitive filesystems are a real pain.


> * create - in case the file already exists returns a file with different case than asked for I believe, Create does not handle that.

Awesome catch, thank you! Tested and fixed.


> * Rename - I am not sure why we have `if fs.insens && newKey == oldKey` key equality check?

Case-only in-place renames (i.e. `foobar -> FooBar`). Added a comment there.


> * Also, feels that `is directory` check should happen before us fiddling with some struct referenced by some map?

Actually, no. Renaming a file onto another file is not a crime and should be done; the check is there to  make sure the user didn't rename a file onto an existing directory, destroying all its contents. In case of an insensitive filesystem this check would trigger an error if the user tried a (totally legit) case-only in-place rename on a directory (`foo -> FOO` where `foo` is a directory to start with), so it is to be done *after* we accounted for the case-only in-place occasion.


> it would be interesting to run the same tests you wrote against a real case insensitive filesystem to validate that the assumptions made in the tests actually matches the reality.

Indeed. I need to figure out a way do to it, other than installing Go on one of my wife's MacBooks (which she won't appreciate).

nekr0z:
To clarify the `create` kerfuffle (and please tell me if my logic is flawed):

`create` basically covers two usecases:
* "creating" a file that already exists, in which case it dumps the old contents and puts a fresh and new file over it. For an insensitive FS all we need to do is make sure we create with non-folded filename as it was given, so that's what's done (now - thanks to you)
* creating a new file where there wasn't one. For an insensitive FS all we need to do is make sure fakeentry.name is non-folded as given, but the key in parent directory's children map is folded - and that's really the only thing we need folding for.

The rest of operations, such as finding the proper parent to create a child for, checking if the child exists and is not a directory, etc., we rely on entryForName to handle. The fact that entryForName does that by folding everything is really tangential and doesn't affect create's logic.

lib/fs/fakefs.go
```diff
@@ -218,6 +227,9 @@ func (fs *fakefs) create(name string) (*fakeEntry, error) {
 		entry.size = 0
 		entry.mtime = time.Now()
 		entry.mode = 0666
+		if fs.insens {
+			entry.name = filepath.Base(name)
+		}
```
AudriusButkevicius: You have an issue again, where the name leaks between two create calls. Call create with foo, store ref, call create with Foo, I expect name of the first ref to change.

nekr0z: It does indeed change, see the updated test. Or do you expect it to remain `foo`?

AudriusButkevicius: I do not expect filenames of things we already have descriptors for to change just because someone else opened the file with a different case.

imsodin:
> > it would be interesting to run the same tests you wrote against a real case insensitive filesystem to validate that the assumptions made in the tests actually matches the reality.
> 
> Indeed. I need to figure out a way do to it, other than installing Go on one of my wife's MacBooks (which she won't appreciate).

It's a bit of an annoyance, but you could move your tests into non-test functions that take the fs as an argument. Then in the actual test function you can run that test once with the fake fs and once with a real fs if the test is run on windows - CI will tell you what whether it's happy with it then.

lib/fs/fakefs.go
imsodin: I think we really need to run those tests on a real fs on windows (or mac) to validate the behaviour of an actual case-insensitive fs: see https://github.com/syncthing/syncthing/pull/6074#issuecomment-545316396.

AudriusButkevicius:
I can run the tests on my windows machine once I am at home.

lib/fs/fakefs.go
nekr0z: > I do not expect filenames of things we already have descriptors for to change just because someone else opened the file with a different case.

In this case someone didn't just open the file, but actually successfully created a new file it that place. I agree, we need to test the real behaviour to have an idea what to really shoot for.

AudriusButkevicius:
I also suspect there might be differences between osx ntfs and fat32...

nekr0z:
Soooo, here it becomes ugly! :-)

Looks like OSX doesn't allow to case-only rename files - at all. And our Windows build system butchers Greek filenames to the point of failure (although your general run-of-the-mill Windows 10 machine supposedly allows Greek letters in filenames).

nekr0z:
I think I'll throw out the Greek part for now (we test the folding rules elsewhere anyway, so that's covered) and concentrate on bringing the behaviour to lowest common denominator.

AudriusButkevicius:
There is a better way to run parameterized tests, see my open PR with CopyRange

nekr0z:
> There is a better way to run parameterized tests, see my open PR with CopyRange

Thank you! Learning more every hour :)

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,640 @@ func TestFakeFSRead(t *testing.T) {
...
+func testFakeFSStatInsens(t *testing.T, fs Filesystem) {
+	if err := fs.Mkdir("/foo", 0755); err != nil {
+		t.Fatal(err)
+	}
+
+	if fd, err := fs.Create("/Foo/aaa"); err != nil {
+		t.Fatal(err)
+	}
+
+	fd.Close()
```
golangcibot: undeclared name: `fd` (from `typecheck`)

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,640 @@ func TestFakeFSRead(t *testing.T) {
...
+		t.Errorf("want %s, got %s", filenames, got)
+	}
+}
+
+func testFakeFSStatInsens(t *testing.T, fs Filesystem) {
+	if err := fs.Mkdir("/foo", 0755); err != nil {
+		t.Fatal(err)
+	}
+
+	if fd, err := fs.Create("/Foo/aaa"); err != nil {
```
golangcibot: fd declared but not used (from `typecheck`)

lib/fs/fakefs_test.go
```diff
@@ -790,6 +793,9 @@ func testFakeFSCreateInsens(t *testing.T, fs Filesystem) {
 	if fd1.Name() != "fOo" {
 		t.Errorf("name of the file created as \"FOO\" is %s", fd1.Name())
 	}
+
+	fd1.Close()
```
AudriusButkevicius: Should just defer acter open?

nekr0z: Yep. And after every Create, apparently. Even after `_, err := Create`. I'm starting to hate windows.

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,664 @@ func TestFakeFSRead(t *testing.T) {
...
+
+	if err := fs.Mkdir("/FOO", 0755); err == nil {
+		t.Errorf("got no error while creating existing directory")
+	}
+}
+
+func TestFakeFSOpenFile(t *testing.T) {
+	fs := newFakeFilesystem("/openf")
+
+	if fd, err := fs.OpenFile("foobar", os.O_RDONLY, 0664); err == nil {
```
golangcibot: fd declared but not used (from `typecheck`)

nekr0z:
I don't get why Windows build still fails - some CI issue? All should work now.

AudriusButkevicius:
I guess you are abusing CI to run tests on Windows?

Reason I ask is I'd like to unsub from notifications, as there is one per commit you push, but I do want to review the result once it's done, so I'd need mentioning to pick it up again.

nekr0z:
> I guess you are abusing CI to run tests on Windows?

I did, but now it is all done, so you're welcome to review.

AudriusButkevicius:
this fixes tests for me on windows. 
the problem was that .stfolder descriptor was not closed post creation.
```diff
diff --git a/lib/fs/fakefs_test.go b/lib/fs/fakefs_test.go
index 153bd711..f6f7c610 100644
--- a/lib/fs/fakefs_test.go
+++ b/lib/fs/fakefs_test.go
@@ -13,6 +13,7 @@ import (
        "io/ioutil"
        "os"
        "path"
+       "path/filepath"
        "reflect"
        "runtime"
        "sort"
@@ -241,21 +242,24 @@ func TestFakeFSCaseInsensitive(t *testing.T) {
        }

        if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
-               if err := os.Mkdir("test-tmp", 0755); err != nil {
-                       t.Fatalf("could not create temporary dir for testing")
+               testDir, err := ioutil.TempDir("", "")
+               if err != nil {
+                       t.Fatalf("could not create temporary dir for testing: %s", err)
                }

-               if _, err := os.Create("test-tmp/.stfolder"); err != nil {
+               if fd, err := os.Create(filepath.Join(testDir, ".stfolder")); err != nil {
                        t.Fatalf("could not create .stfolder")
+               } else {
+                       _ = fd.Close()
                }

                defer func() {
-                       if err := os.RemoveAll("test-tmp"); err != nil {
-                               t.Fatalf("could not remove test-tmp directory")
+                       if err := os.RemoveAll(testDir); err != nil {
+                               t.Fatalf("could not remove test directory: %s", err)
                        }
                }()

-               filesystems = append(filesystems, testFS{runtime.GOOS, newBasicFilesystem("test-tmp")})
+               filesystems = append(filesystems, testFS{runtime.GOOS, newBasicFilesystem(testDir)})
        }

        runTests(t, tests, filesystems)
```

lib/fs/fakefs.go
```diff
@@ -444,13 +504,25 @@ func (fs *fakefs) Rename(oldname, newname string) error {
...
-	dst, ok := p1.children[filepath.Base(newname)]
-	if ok && dst.entryType == fakeEntryTypeDir {
-		return errors.New("is a directory")
+	dst, ok := p1.children[newKey]
+	if ok {
+		// in-place case-only directory renames will fail on insensitive systems
+		// they do fail on OS X, so that's expected
+		if dst.entryType == fakeEntryTypeDir {
+			return errors.New("is a directory")
+		}
```
AudriusButkevicius: Shouldn't this be gated by fs.insens?
Presumably this should succeed on linux with case sensitivity?

lib/fs/fakefs.go
```diff
@@ -500,18 +572,30 @@ func (fs *fakefs) URI() string {
...
-	return fi1.Name() == fi2.Name()
+	// BUG: real systems base file sameness on path, inodes, etc
+	// we try our best, but FileInfo just doesn't have enough data
+	// so there be false positives, especially on Windows
+	// where ModTime is not that precise
+	var ok bool
+	if fs.insens {
+		ok = (UnicodeLowercase(fi1.Name()) == UnicodeLowercase(fi2.Name()))
+	} else {
+		ok = (fi1.Name() == fi2.Name())
```
AudriusButkevicius: No need for the () around the equality

lib/fs/fakefs.go
```diff
@@ -444,13 +504,25 @@ func (fs *fakefs) Rename(oldname, newname string) error {
...
+		// they do fail on OS X, so that's expected
+		if dst.entryType == fakeEntryTypeDir {
+			return errors.New("is a directory")
+		}
+
+		if fs.insens && newKey == oldKey {
+			// case-only in-place rename
+			entry.name = filepath.Base(newname)
+			return nil
+		}
```
AudriusButkevicius: This is sort of the same as the stuff that is happening outside of the if ok condition, we could just make the delete conditional on `fs.insens && p0 == p1 && oldKey == newKey`

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+		{"Remove", testFakeFSRemove},
+		{"SameFile", testFakeFSSameFile},
+	}
+	var filesystems = []testFS{
+		{"fakefs", newFakeFilesystem("/foo")},
+	}
+
+	if runtime.GOOS == "linux" {
+		if err := os.Mkdir("test-tmp", 0755); err != nil {
+			t.Fatalf("could not create temporary dir for testing")
```
AudriusButkevicius: add err to format string (everywhere, it seems missing in most places)

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+		{"RemoveAll", testFakeFSRemoveAll},
+		{"Remove", testFakeFSRemove},
+		{"SameFile", testFakeFSSameFile},
+	}
+	var filesystems = []testFS{
+		{"fakefs", newFakeFilesystem("/foo")},
+	}
+
+	if runtime.GOOS == "linux" {
+		if err := os.Mkdir("test-tmp", 0755); err != nil {
```
AudriusButkevicius: See the code I posted, we should use that to run tests in a temp dir rather than some specific dir in our local checkout (which then if fails makes git checkout dirty)

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("could not create file: %s", err)
+	}
+
+	_, err = fd1.Write(bs1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	fd1.Close()
```
AudriusButkevicius: would leave file unclosed if Write fails, use defer after create.

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+	}
+
+	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
+		if err := os.Mkdir("test-tmp", 0755); err != nil {
+			t.Fatalf("could not create temporary dir for testing")
+		}
+
+		if _, err := os.Create("test-tmp/.stfolder"); err != nil {
+			t.Fatalf("could not create .stfolder")
+		}
```
AudriusButkevicius: missing a close.

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+
+	if path.Clean(directory) == "/" {
+		filenames = append(filenames, ".stfolder")
+	}
+	sort.Strings(filenames)
+	sort.Strings(got)
+
+	if !reflect.DeepEqual(got, filenames) {
+		t.Errorf("want %s, got %s", filenames, got)
+	}
```
AudriusButkevicius: Don't do this, just compare the length, and if the length is the same, write a loop to compare each item.

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+
+	fd1, err := fs.Open("/FOO/AAA")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if info, err = fd1.Stat(); err != nil {
+		t.Fatal(err)
+	}
```
AudriusButkevicius: This doesn't check anything.

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+
+	fd2, err := fs.Open("Foo/aAa")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if _, err = fd2.Stat(); err != nil {
+		t.Fatal(err)
+	}
```
AudriusButkevicius: Again, what's the expected failure case here?

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+
+	if info.Name() != "AAA" {
+		t.Errorf("want AAA, got %s", info.Name())
+	}
```
AudriusButkevicius: This is like a random info from 10 lines above. 

I think the test we care about is
```
fd1 = open(x)
fd2 = open(X)
assert(fd1.Name(), "x")
assert(fd2.Name(), "X")
assert(fd1.Stat().Name(), "x")
assert(fd2.Stat().Name(), "X")
```

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+
+	fd1.Close()
+	fd2.Close()
```
AudriusButkevicius: Again, leaking stuff in case of errors, use defer.

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+		{"/foo", "bar", "bar"},
+	}
+
+	for _, testCase := range testCases {
+		fs := newFakeFilesystem(testCase.fs)
+		if _, err := fs.Create(testCase.create); err != nil {
+			t.Fatal(err)
+		}
+
+		fd, err := fs.Open(testCase.open)
```
AudriusButkevicius: fd never closed

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+		fs     string
+		create string
+		open   string
+	}{
+		{"/foo", "bar", "bar"},
+	}
+
+	for _, testCase := range testCases {
+		fs := newFakeFilesystem(testCase.fs)
+		if _, err := fs.Create(testCase.create); err != nil {
```
AudriusButkevicius: leaking fd

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+	assertDir(t, fs, "/foo", []string{"aaa"})
+}
+
+func TestFakeFSFileName(t *testing.T) {
+	var testCases = []struct {
+		fs     string
+		create string
+		open   string
+	}{
+		{"/foo", "bar", "bar"},
```
AudriusButkevicius: Should we test for a failure case too?

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	fd, err := fs.Create("/foo/bar/baz/qux")
+	if err != nil {
+		t.Fatal(err)
+	}
+	fd.Close()
+
+	if err := fs.Rename("/foo/bar/baz/qux", "/foo/baz/bar/qux"); err == nil {
```
AudriusButkevicius: the name could be a bit more different, it was not obvious that they were different because the difference is a single character in a 11 character directory path.

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+
+	for _, dir := range dirs {
+		assertDir(t, fs, dir.dir, dir.files)
+	}
+
+	// the next rename can be done on Windows, and maybe elsewhere, but not on OS X
+	// so we're shooting for the lowest common denominator
+	if _, ok := fs.(*fakefs); ok || runtime.GOOS == "darwin" {
+		if err := fs.Rename("/foo/bar/BAZ", "/FOO/BAR/bAz"); err == nil {
+			t.Errorf("In-place case-only directory renames fail on OS X, should fail here, too: %s", err)
```
AudriusButkevicius: I guess this is not an expected failure, so should be "not expected to fail here"?

AudriusButkevicius: Should we have a mode to emulate this in fakefs?

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+		t.Errorf("created an existing file while told not to")
+	}
+
+	if _, err := fs.OpenFile("foobar", os.O_RDWR|os.O_CREATE, 0664); err != nil {
+		t.Fatal(err)
+	}
+
+	if _, err := fs.OpenFile("foobar", os.O_RDWR, 0664); err != nil {
+		t.Fatal(err)
+	}
```
AudriusButkevicius: Tons of leaks in this function

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+}
+
+func testFakeFSOpenFileInsens(t *testing.T, fs Filesystem) {
+	fd, err := fs.OpenFile("FooBar", os.O_RDONLY, 0664)
+	if err == nil {
+		t.Errorf("got no error opening a non-existing file")
```
AudriusButkevicius: just fatal I think, to abort the test so you can scrap half the ifs in this test.

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+		t.Errorf("real systems don't return error here")
+	}
+}
+
+func testFakeFSRemove(t *testing.T, fs Filesystem) {
+	if err := fs.Mkdir("/Foo", 0755); err != nil {
+		t.Fatal(err)
+	}
+
+	if _, err := fs.Create("/Foo/Bar"); err != nil {
```
AudriusButkevicius: leaks

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	fd, err := fs.Create("/Foo/Bar")
+	if err != nil {
+		t.Fatal(err)
+	}
+	fd.Close()
+
+	if err := fs.Remove("/FOO"); err == nil || err == os.ErrNotExist {
```
AudriusButkevicius: should check for not empty error instead?

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+		t.SkipNow()
+	}
+
+	if err := fs.Mkdir("/Foo", 0755); err != nil {
+		t.Fatal(err)
+	}
+
+	filenames := []string{"Bar", "Baz", "/Foo/Bar"}
+	for _, filename := range filenames {
+		if _, err := fs.Create(filename); err != nil {
```
AudriusButkevicius: leaks

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+	}
+
+	testCases := []struct {
+		f1   string
+		f2   string
+		want bool
+	}{
+		{f1: "Bar", f2: "Baz", want: false},
+		{f1: "Bar", f2: "/Foo/Bar", want: false},
+		{"Bar", "Bar", true},
```
AudriusButkevicius: inconsistent struct init, either annotate member names, or don't

lib/fs/fakefs_test.go
```diff
@@ -172,3 +178,722 @@ func TestFakeFSRead(t *testing.T) {
...
+		fd.Close()
+	}
+
+	testCases := []struct {
+		f1   string
+		f2   string
+		want bool
+	}{
+		{f1: "bAr", f2: "baZ", want: false},
+		{"baz", "BAZ", true},
```
AudriusButkevicius: ditto about member naming.

AudriusButkevicius:
beware there is like 9 hidden comments.

lib/fs/fakefs_test.go
AudriusButkevicius: I think there is a test that does something like this below, minus the stat stuff, which could be added there, which then questions the value of this test.

nekr0z:
Thank you! I'll have a look into all this as soon as I have spare time (tomorrow, hopefully), but it looks like we're nearing the finish line here, finally.

lib/fs/fakefs.go
nekr0z: Nope. On sensitive filesystem this check is always requires, because there is no such thing as an in-place rename. `foo -> Foo` is always a rename to a new file, not the same one, and if `Foo` is an existing directory, we want to abort this, hence the check.

lib/fs/fakefs_test.go
nekr0z: Yep, Open can be omitted, but it basically checks that fd.Stat doesn't spoil the name as well as fs.Stat.

lib/fs/fakefs_test.go
nekr0z: Open and fd.Stat same file with different filename. Don't want fd1.Name() to change.

lib/fs/fakefs_test.go
nekr0z: fs.Stat and fd.Stat are different beasts, so testing here explicitly. Will analyze more, but my hunch is that this particular kerfuffle isn't tested elsewhere.

lib/fs/fakefs_test.go
nekr0z: All we test here is that Open gives fd the desired name no matter what the real filename is. Don't really see what a failure case would be.

lib/fs/fakefs_test.go
nekr0z: I'd actually love it if other maintainers would say something about this particular issue. Thing is: in-place case-only directory rename fails on OS X and doesn't fail on Windows. Judging by other tests, this is actually the only place Win and OSX behaviors differ.

My hunch is we have fakefs simulate OS X in this case, i.e. fail, as this is the lowest common. However, I can see how it would be beneficial to have both cases covered. Question is: do we really need two fakefs insensitive modes that would only differ in this particular behavior?

If we don't, then this error message is correct and what's expected.

lib/fs/fakefs_test.go
nekr0z: got it

lib/fs/fakefs_test.go
nekr0z: fixed

lib/fs/fakefs_test.go
nekr0z: fixed

lib/fs/fakefs_test.go
nekr0z: fixed

lib/fs/fakefs_test.go
nekr0z: got it

lib/fs/fakefs_test.go
nekr0z: yep, got it

lib/fs/fakefs_test.go
AudriusButkevicius: But seems the name is the same here.

nekr0z: Here - yes, since this is the test for case-sensitive filesystems (a redundant one, perhaps).

lib/fs/fakefs_test.go
calmh: This is fascinating and I had no idea. Frankly I think this is a Go standard library bug and nothing else. The `rename` **system call** on macOS succeeds perfectly fine in `rename("test", "TEST")`, as shown by tracing `mv` on my system. However, *Go* just `stat`s both the old and new name and then returns "already exists" without even trying the rename.

AudriusButkevicius:
Should I review this, or is there comments still to be addressed?

lib/fs/fakefs_test.go
AudriusButkevicius: Raised it as https://github.com/golang/go/issues/35222

Yet it seems this should only be an issue for directories, and not files 🤷‍♂

nekr0z: And only on `darwin`, actually. On windows Go behaves and doesn't throw this error.

nekr0z:
> Should I review this, or is there comments still to be addressed?

I think I did all there was to do, so please review if you have time.

lib/fs/fakefs.go
```diff
@@ -289,17 +312,23 @@ func (fs *fakefs) Mkdir(name string, perm FileMode) error {
 	dir := filepath.Dir(name)
 	base := filepath.Base(name)
 	entry := fs.entryForName(dir)
+	key := base
+
+	if fs.insens {
+		key = UnicodeLowercase(key)
+	}
+
```
AudriusButkevicius: this could be moved closer to where the key is used

lib/fs/fakefs.go
```diff
@@ -344,28 +378,38 @@ func (fs *fakefs) Open(name string) (File, error) {
...
+	defer fs.mut.Unlock()
+
 	dir := filepath.Dir(name)
 	base := filepath.Base(name)
 	entry := fs.entryForName(dir)
+	key := base
+
+	if fs.insens {
+		key = UnicodeLowercase(key)
+	}
```
AudriusButkevicius: again, could move key folding closer to where it's used.

lib/fs/fakefs_test.go
```diff
@@ -172,3 +176,728 @@ func TestFakeFSRead(t *testing.T) {
...
+}
+
+type test struct {
+	name string
+	impl func(t *testing.T, fs Filesystem)
+}
+
+func TestFakeFSCaseSensitive(t *testing.T) {
+	var tests = []test{
+		{"read", testFakeFSRead},
```
AudriusButkevicius: casing

lib/fs/fakefs_test.go
```diff
@@ -172,3 +176,728 @@ func TestFakeFSRead(t *testing.T) {
...
+		defer removeTestDir(t, testDir)
+		filesystems = append(filesystems, testFS{runtime.GOOS, newBasicFilesystem(testDir)})
+	}
+
+	runTests(t, tests, filesystems)
+}
+
+func TestFakeFSCaseInsensitive(t *testing.T) {
+	var tests = []test{
+		{"read", testFakeFSRead},
```
AudriusButkevicius: casing

lib/fs/fakefs_test.go
```diff
@@ -172,3 +176,728 @@ func TestFakeFSRead(t *testing.T) {
...
+		{"OpenFile", testFakeFSOpenFile},
+		{"RemoveAll", testFakeFSRemoveAll},
+		{"Remove", testFakeFSRemove},
+		{"Mkdir", testFakeFSMkdir},
+		{"SameFile", testFakeFSSameFile},
+		{"DirNames", testDirNames},
+		{"FileName", testFakeFSFileName},
+		{"generalInsens", testFakeFSCaseInsensitive},
```
AudriusButkevicius: casing

lib/fs/fakefs_test.go
```diff
@@ -172,3 +176,728 @@ func TestFakeFSRead(t *testing.T) {
...
+		{"Mkdir", testFakeFSMkdir},
+		{"SameFile", testFakeFSSameFile},
+		{"DirNames", testDirNames},
+		{"FileName", testFakeFSFileName},
+	}
+	var filesystems = []testFS{
+		{"fakefs", newFakeFilesystem("/foo")},
+	}
+
+	if runtime.GOOS == "linux" {
```
AudriusButkevicius: Instead of doing this, we should instead validate whether the filesystem in testDir is case sensitive or not, same like I did in the Go PR https://github.com/golang/go/pull/35298

lib/fs/fakefs_test.go
```diff
@@ -172,3 +176,728 @@ func TestFakeFSRead(t *testing.T) {
...
+		{"SameFileInsens", testFakeFSSameFileInsens},
+		{"CreateInsens", testFakeFSCreateInsens},
+		{"FileNameInsens", testFakeFSFileNameInsens},
+	}
+
+	var filesystems = []testFS{
+		{"fakefs", newFakeFilesystem("/foobar?insens=true")},
+	}
+
+	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
```
AudriusButkevicius: ditto about checks: https://github.com/golang/go/pull/35298

lib/fs/fakefs_test.go
```diff
@@ -172,3 +176,728 @@ func TestFakeFSRead(t *testing.T) {
...
+
+	if err := fs.Remove("/FoO"); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func testFakeFSSameFile(t *testing.T, fs Filesystem) {
+	if runtime.GOOS == "windows" {
+		// windows time in not precise enough
+		t.SkipNow()
```
AudriusButkevicius: Should do t.Skip("windows time in not precise enough")

AudriusButkevicius: But hold on, this is not a real filesystem so the time can be as accurate as we want. On a real filesystem our modtime checks are not used, as we call os.SameFile, so not sure why this fails for fakefs on windows?

nekr0z: I don't have a Windows box to test on, but I suppose it has something to do with time resolution beind worse on Windows than it is on other OSes in Go.  If I remember correctly, it used to be as bad as 15.6 ms before Go 1.7, then was supposedly fixed but still is nowhere near Unix-likes. I guess I could put a sleep into this test for completeness' sake.

lib/fs/fakefs.go
nekr0z: Solved.

lib/fs/fakefs.go
nekr0z: Still need to circumvent this check for non-destructive in-place rename, so left it here. Tell me if it still looks wrong to you.

lib/fs/fakefs_test.go
```diff
@@ -796,6 +800,9 @@ func testFakeFSSameFile(t *testing.T, fs Filesystem) {
 			t.Fatalf("Could not create %s: %s", filename, err)
 		} else {
 			fd.Close()
+			if runtime.GOOS == "windows" {
+				time.Sleep(1 * time.Millisecond)
+			}
```
AudriusButkevicius: We should not need this, I'd like to see this fails and repro it.

nekr0z: This does fail without this sleep on our build server. I'll remove it to demostrate.

nekr0z:
@AudriusButkevicius see? I'm not sure why Mac and Coverage suddenly failed (looks unrelated), but Windows build fails this very test for the very reason I've been mentioning.

nekr0z:
I bet [this](https://play.golang.org/p/tidQqmIfjHv) fails on Windows, but have no machine to check.

AudriusButkevicius:
Time should not be compared using, it should be done using .Equals

nekr0z:
> Time should not be compared using, it should be done using .Equals

My bad. Fixed.

nekr0z:
But that still doesn't help Windows pass this test (and we seem to have Mac build server broken).

nekr0z:
Thanks so very much to all three of you who helped make it happen!

calmh:
Thank *you* for patience and diligence in working through this thing. Hope to see it result in future improvements! :)

PR #6078, opened 2019-10-12
all: Update certificate lifetimes (fixes #6036)

### LINKED ISSUE #6036: gui: Generate certificates accepted by iOS 13 / macOS Catalina
Some new requirements have been added in new OSes. One thing we will run afoul of:

> TLS server certificates must have a validity period of 825 days or fewer (as expressed in the NotBefore and NotAfter fields of the certificate). (https://support.apple.com/en-us/HT210176)

We generate GUI certs with very long validity time. We'd need to cut this down to two years, and then also be prepared to re-generate the cert before it expires.

(I haven't yet tested whether the "fuck it, I trust this cert anyway" dialog for self-signed certs maybe overrides this requirement too.)

calmh:
This adds a certificate lifetime parameter to our certificate generation
and hard codes it to twenty years in some uninteresting places. In the
main binary there are a couple of constants but it results in twenty
years for the device certificate and 820 days for the HTTPS one. 820 is
less than the 825 maximum Apple allows nowadays.

This also means we must be prepared for certificates to expire, so I add
some handling for that and generate a new certificate when needed. For
our own certificates (common name Syncthing) we regenerate a month ahead
of time. For other certificates we leave well enough alone.

### DIFF
--- a/cmd/stdiscosrv/main.go
+++ b/cmd/stdiscosrv/main.go
@@ -100,7 +100,7 @@ func main() {
 	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
 	if err != nil {
 		log.Println("Failed to load keypair. Generating one, this might take a while...")
-		cert, err = tlsutil.NewCertificate(certFile, keyFile, "stdiscosrv")
+		cert, err = tlsutil.NewCertificate(certFile, keyFile, "stdiscosrv", 20*365)
 		if err != nil {
 			log.Fatalln("Failed to generate X509 key pair:", err)
 		}
--- a/cmd/strelaypoolsrv/main.go
+++ b/cmd/strelaypoolsrv/main.go
@@ -633,7 +633,7 @@ func createTestCertificate() tls.Certificate {
 	}
 
 	certFile, keyFile := filepath.Join(tmpDir, "cert.pem"), filepath.Join(tmpDir, "key.pem")
-	cert, err := tlsutil.NewCertificate(certFile, keyFile, "relaypoolsrv")
+	cert, err := tlsutil.NewCertificate(certFile, keyFile, "relaypoolsrv", 20*365)
 	if err != nil {
 		log.Fatalln("Failed to create test X509 key pair:", err)
 	}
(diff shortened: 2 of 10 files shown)


AudriusButkevicius:
Should we regen the http certs for all users as a migration anticipating it becoming an issue with mac os soon?

calmh:
Yeah, why not. We only need to regen certificates issued after July 1, that’s the cutoff for the newer stricter rules.

calmh:
I realize people can still hit this by running on Linux and surfing from mac, but there's also damage in dropping certificates that people have already added to their browsers' trust store so I think limiting this check to mac makes sense.

GermanCoding:
I would check if the cert is self-signed in addition to checking CN == "syncthing". I'm usually deploying custom CA's for local devices in networks I work with and I don't care that much about CN's. When I generated a custom certificate for syncthing's HTTPS I almost named it "syncthing", since that's a fitting description for the cert. If I had done that, this change would **overwrite** my custom CA-signed certificates with self-signed ones (at some point).

The number of people with custom certs named syncthing is probably less than one in a million, but a safety check here would still be nice. Or make it clear that CN = "syncthing" is reserved and that certs with that name can get overwritten.

calmh:
Right, there is that, and also I had forgotten how our HTTPS certs are really generated. We generally use the hostname for common name for HTTPS certs, but we don't add any DNS names or IP addresses, which any real HTTPS cert should have, so check for that.

(This might be something need to fix as well in the future for these to be accepted, but that's a future tweak. At that time we might also generate a long lived CA for our HTTPS certificates and sign using that, so that if the user imports it even future regenerated certificates might be accepted.)

AudriusButkevicius:
You still accept leafs, not the CA so not sure how that would help.

calmh:
If I import the CA the leaves will be accepted. But that discussion is for another issue/PR really.

AudriusButkevicius:
Sure, but thats not an option when the warning pops up in the browser.

calmh:
Regardless that's not what this PR is about...

PR #6080, opened 2019-10-12
Disable automatic upgrades via command line argument (fixes #6079)

### LINKED ISSUE #6079: Allow the user to disable automatic upgrades via a command line argument
Disabling automatic updates via the command line would be a nice to have feature. I know it would directly benefit the usage of syncthing that my setup requires.

Thanks!

scavet64:
### Purpose

The purpose of this PR is to provide the user with an additional method for disabling the automatic upgrade of syncthing. This is an alternative to the environment variable method in the case that it is not possible to use or it is not desired to use the variable. A similar alternative exists for the `STNORESTART` environment variable.

### Testing

Testing has been preform by building and running the executable with and without the new command line argument. I verified that the change has done what I have expected by verifying the syncthing logs upon launching the executable

### Documentation

The required document changes would be to update this page to include the new command line argument and some details behind it: https://docs.syncthing.net/users/syncthing.html


calmh:
There is legacy badness around which things are command line options or environment variables or both. There is some sense in harmonizing this to also add a command line switch. However, I don't think we need it and I'd rather have fewer switches than more. So unless there is a good case where this is *required* (and I can't think of one) I think we will pass on this one.

PR #6082, opened 2019-10-13
lib/connections: Rate limit quic accept loop (fixes #6081)

### LINKED ISSUE #6081: lib/connections: Quic error loop too tight
Since i upgraded to SyncThing 1.3.0 (running on Windows 7 SP1 x64), the following warning appears continuously about 100 times A SECOND which quickly fills up my log and then crashes SyncThing completely: 

"WARNING: Listen (BEP/quic): Accepting connection: read udp 0.0.0.0:22000: wsarecvfrom: The connection has been broken due to keep-alive activity detecting a failure while the operation was in progress."

It does not start immediately, sometimes after three minutes, sometimes after a period of hours, but then, when it starts, it floods the command window, floods the log and finally kills SyncThing.

AudriusButkevicius:
Produced completely in github web editor, let's see if it will blend.

### DIFF
--- a/lib/connections/quic_listen.go
+++ b/lib/connections/quic_listen.go
@@ -110,7 +110,7 @@ func (t *quicListener) serve(stop chan struct{}) error {
 
 	l.Infof("QUIC listener (%v) starting", packetConn.LocalAddr())
 	defer l.Infof("QUIC listener (%v) shutting down", packetConn.LocalAddr())
-	
+
 	acceptFailures := 0
 	const maxAcceptFailures = 10
 
@@ -126,7 +126,7 @@ func (t *quicListener) serve(stop chan struct{}) error {
 			return nil
 		} else if err != nil {
 			l.Infoln("Listen (BEP/quic): Accepting connection:", err)
-			
+		
 			acceptFailures++
 			if acceptFailures > maxAcceptFailures {
 				// Return to restart the listener, because something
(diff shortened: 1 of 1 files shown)


lib/connections/quic_listen.go
```diff
@@ -122,8 +125,22 @@ func (t *quicListener) serve(stop chan struct{}) error {
...
+			
+			acceptFailures++
+			if acceptFailures > maxAcceptFailures {
+				// Return to restart the listener, because something
+				// seems permanently damaged.
+				return err
+			}
+
+			// Slightly increased delay for each failure.
+			time.Sleep(time.Duration(acceptFailures) * time.Second)
```
golangcibot: undeclared name: `time` (from `typecheck`)

imsodin:
Is there anything that makes accept failure special, i.e. likely that repeated tries with short backoffs lead to success. Otherwise I'd just return from serve and adjust the supervisor to allow a few more tries.

In any case I'd log individual failures at info-level and only once the threshold is reached at warning-level to avoid repeated, likely identical warnings/UI banners.

calmh:
I agree we should bail after multiple failures, so we get a new socket and maybe a better start in life. Probably best not to do it on every failure though as there are a lot of moving parts in starting up the quic service...

AudriusButkevicius:
So I'm away till weekend, mobile only, feel free to bastardize this.

Issue #5907, opened 2019-07-28
Database correctness: bubble errors instead of panics


calmh:
Because error handling.

PR #6107, opened 2019-10-25
Implement database abstraction, error checking (ref #5907)

calmh:
So here's a slog of a PR ...

### Purpose

This PR does two things, because one lead to the other:

- Move the leveldb specific stuff into a small "backend" package that
defines a backend interface and the leveldb implementation. This allows,
potentially, in the future, switching the db implementation so another
KV store should we wish to do so.

- Add proper error handling all along the way. The db and backend
packages are now errcheck clean. However, I drew the line at modifying
the FileSet API in order to keep this manageable and not continue
refactoring all of the rest of Syncthing. As such, the FileSet methods
still panic on database errors, except for the "database is closed"
error which is instead handled by silently returning as quickly as
possible, with the assumption that we're anyway "on the way out".

### Testing

I've been running with this for a few days and it passes at least some integration testing. That doesn't constitute heavy testing exactly and database changes are always scary. That said I think this already improves things as there was potential for, for example, silently failing to save a new folder ID and yet continuing on with life... That won't happen now.

### "Pluggable backend"

That's not only a fantasy. Part of the purpose of this PR was to test another KV implementation, and I'm in fact running this by just dropping in a badger implementation in backend with no other changes and this seems fine so far. Whether we want to go with that or not is absolutely not decided but I like the ability to experiment.


cmd/stindex/idxck.go
```diff
@@ -31,7 +33,7 @@ type sequenceKey struct {
 	sequence uint64
 }
 
-func idxck(ldb *db.Lowlevel) (success bool) {
+func idxck(ldb backend.Backend) (success bool) {
```
golangcibot: cyclomatic complexity 40 of func `idxck` is high (> 30) (from `gocyclo`)

lib/db/schemaupdater.go
```diff
@@ -65,36 +64,58 @@ func (db *schemaUpdater) updateSchema() error {
...
+	if err := miscDB.PutString("dbMinSyncthingVersion", dbMinSyncthingVersion); err != nil {
+		return err
+	}
 
 	return nil
 }
 
-func (db *schemaUpdater) updateSchema0to1() {
-	t := db.newReadWriteTransaction()
+func (db *schemaUpdater) updateSchema0to1() error {
```
golangcibot: Function 'updateSchema0to1' has too many statements (58 > 40) (from `funlen`)

cmd/stindex/dumpsize.go
```diff
@@ -37,11 +39,14 @@ func (h *ElementHeap) Pop() interface{} {
 	return x
 }
 
-func dumpsize(ldb *db.Lowlevel) {
+func dumpsize(ldb backend.Backend) {
```
golangcibot: Function 'dumpsize' has too many statements (42 > 40) (from `funlen`)

lib/db/backend/leveldb_open.go
```diff
@@ -0,0 +1,226 @@
...
+// OpenMemory returns a new Lowlevel referencing an in-memory database.
+func OpenLevelDBMemory() Backend {
+	ldb, _ := leveldb.Open(storage.NewMemStorage(), nil)
+	return &leveldbBackend{ldb: ldb}
+}
+
+// optsFor returns the database options to use when opening a database with
+// the given location and tuning. Settings can be overridden by debug
+// environment variables.
+func optsFor(location string, tuning Tuning) *opt.Options {
```
golangcibot: Function 'optsFor' is too long (62 > 60) (from `funlen`)

cmd/stindex/dump.go
```diff
@@ -13,11 +13,15 @@ import (
 	"time"
 
 	"github.com/syncthing/syncthing/lib/db"
+	"github.com/syncthing/syncthing/lib/db/backend"
 	"github.com/syncthing/syncthing/lib/protocol"
 )
 
-func dump(ldb *db.Lowlevel) {
-	it := ldb.NewIterator(nil, nil)
+func dump(ldb backend.Backend) {
```
golangcibot: Function 'dump' has too many statements (49 > 40) (from `funlen`)

lib/db/instance.go
```diff
@@ -51,42 +58,69 @@ func (db *instance) updateRemoteFiles(folder, device []byte, fs []protocol.FileI
...
 	}
+
+	return t.commit()
 }
 
 // updateLocalFiles adds fileinfos to the db, and updates the global versionlist,
 // metadata, sequence and blockmap buckets.
-func (db *instance) updateLocalFiles(folder []byte, fs []protocol.FileInfo, meta *metadataTracker) {
-	t := db.newReadWriteTransaction()
+func (db *instance) updateLocalFiles(folder []byte, fs []protocol.FileInfo, meta *metadataTracker) error {
```
golangcibot: Function 'updateLocalFiles' has too many statements (58 > 40) (from `funlen`)

lib/db/instance.go
```diff
@@ -177,25 +263,37 @@ func (db *instance) withHaveSequence(folder []byte, startSeq int64, fn Iterator)
...
-			return
+			return nil
 		}
 	}
+	return dbi.Error()
 }
 
-func (db *instance) withAllFolderTruncated(folder []byte, fn func(device []byte, f FileInfoTruncated) bool) {
-	t := db.newReadWriteTransaction()
+func (db *instance) withAllFolderTruncated(folder []byte, fn func(device []byte, f FileInfoTruncated) bool) error {
```
golangcibot: Function 'withAllFolderTruncated' is too long (62 > 60) (from `funlen`)

lib/db/instance.go
```diff
@@ -205,42 +303,61 @@ func (db *instance) withAllFolderTruncated(folder []byte, fn func(device []byte,
...
 	defer t.close()
-	_, f, ok := t.getGlobal(nil, folder, file, truncate)
-	return f, ok
+	_, f, ok, err := t.getGlobal(nil, folder, file, truncate)
+	return f, ok, err
 }
 
-func (db *instance) withGlobal(folder, prefix []byte, truncate bool, fn Iterator) {
-	t := db.newReadOnlyTransaction()
+func (db *instance) withGlobal(folder, prefix []byte, truncate bool, fn Iterator) error {
```
golangcibot: Function 'withGlobal' is too long (61 > 60) (from `funlen`)

lib/db/instance.go
```diff
@@ -312,19 +447,28 @@ func (db *instance) availability(folder, file []byte) []protocol.DeviceID {
 		devices = append(devices, n)
 	}
 
-	return devices
+	return devices, nil
 }
 
-func (db *instance) withNeed(folder, device []byte, truncate bool, fn Iterator) {
+func (db *instance) withNeed(folder, device []byte, truncate bool, fn Iterator) error {
```
golangcibot: Function 'withNeed' has too many statements (45 > 40) (from `funlen`)

lib/db/instance.go
```diff
@@ -372,81 +522,168 @@ func (db *instance) withNeed(folder, device []byte, truncate bool, fn Iterator)
...
+		if err := t.deleteKeyPrefix(key.WithoutHashAndName()); err != nil {
+			return err
+		}
 	}
+	return t.commit()
 }
 
-func (db *instance) checkGlobals(folder []byte, meta *metadataTracker) {
-	t := db.newReadWriteTransaction()
+func (db *instance) checkGlobals(folder []byte, meta *metadataTracker) error {
```
golangcibot: Function 'checkGlobals' is too long (61 > 60) (from `funlen`)

lib/db/transactions.go
```diff
@@ -7,111 +7,152 @@
...
-	t.flush()
+	t.WriteTransaction.Release()
 	t.readOnlyTransaction.close()
 }
 
 // updateGlobal adds this device+version to the version list for the given
 // file. If the device is already present in the list, the version is updated.
 // If the file does not have an entry in the global list, it is created.
-func (t readWriteTransaction) updateGlobal(gk, keyBuf, folder, device []byte, file protocol.FileInfo, meta *metadataTracker) ([]byte, bool) {
+func (t readWriteTransaction) updateGlobal(gk, keyBuf, folder, device []byte, file protocol.FileInfo, meta *metadataTracker) ([]byte, bool, error) {
```
golangcibot: Function 'updateGlobal' is too long (86 > 60) (from `funlen`)

lib/db/transactions.go
```diff
@@ -202,71 +273,98 @@ func need(global FileIntf, haveLocal bool, localVersion protocol.Vector) bool {
 // removeFromGlobal removes the device from the global version list for the
 // given file. If the version list is empty after this, the file entry is
 // removed entirely.
-func (t readWriteTransaction) removeFromGlobal(gk, keyBuf, folder, device []byte, file []byte, meta *metadataTracker) []byte {
+func (t readWriteTransaction) removeFromGlobal(gk, keyBuf, folder, device []byte, file []byte, meta *metadataTracker) ([]byte, error) {
```
golangcibot: Function 'removeFromGlobal' has too many statements (44 > 40) (from `funlen`)

cmd/stindex/dump.go
AudriusButkevicius: Shut it

calmh: I told the bot to lighten up a bit

lib/db/backend/backend.go
```diff
@@ -0,0 +1,116 @@
...
+	ReadTransaction
+	Writer
+	Commit() error
+}
+
+type Iterator interface {
+	Next() bool
+	Key() []byte
+	Value() []byte
+	Error() error
```
imsodin: How do I realize there is an error while iterating - presumably I'd need to call `Error` every time after `Next`? Looking at code sometimes that's done, sometimes not.  
I'd expect to have an error return value on `Next`, and then no error on `Value`/`Key` if that succeeds, as my expectation would be that calling `Next` already does look those values up. Otherwise I'd expect an error return value on all of them. That's a bit cumbersome probably, but most consistent with other methods and least surprising in my opinion.

lib/db/instance.go
```diff
@@ -372,81 +522,168 @@ func (db *instance) withNeed(folder, device []byte, truncate bool, fn Iterator)
...
 			if !fn(gf) {
-				return
+				return nil
 			}
 
 			// This file is handled, no need to look further in the version list
 			break
 		}
 	}
+	return dbi.Error()
```
AudriusButkevicius: Same question as @imsodin, the relation between .Next() and Error() is not clear.

We should declare how it works in the interface, and make sure leveldb guarantees that behaviour.

lib/db/transactions.go
```diff
@@ -7,111 +7,152 @@
...
 }
 
-func (t readOnlyTransaction) getFileTrunc(key []byte, trunc bool) (FileIntf, bool) {
-	bs, err := t.Get(key, nil)
-	if err == leveldb.ErrNotFound {
-		return nil, false
+func (t readOnlyTransaction) getFileTrunc(key []byte, trunc bool) (FileIntf, bool, error) {
+	bs, err := t.Get(key)
+	if backend.IsNotFound(err) {
+		return nil, false, nil
```
AudriusButkevicius: Why not return that?

Feels like "does this file exist, no, I'll overwrite it" would make it terrible for us.

lib/db/transactions.go
```diff
@@ -7,111 +7,152 @@
...
+	var err error
+	keyBuf, err = t.keyer.GenerateGlobalVersionKey(keyBuf, folder, file)
+	if err != nil {
+		return nil, nil, false, err
+	}
 
-	bs, err := t.Get(keyBuf, nil)
+	bs, err := t.Get(keyBuf)
+	if backend.IsNotFound(err) {
+		return keyBuf, nil, false, nil
```
AudriusButkevicius: ditto

lib/db/transactions.go
```diff
@@ -7,111 +7,152 @@
...
+		WriteTransaction: tran,
+		readOnlyTransaction: readOnlyTransaction{
+			ReadTransaction: tran,
+			keyer:           db.keyer,
+		},
+	}, nil
+}
+
+func (t readWriteTransaction) commit() error {
+	defer t.readOnlyTransaction.close()
```
AudriusButkevicius: funny defer

lib/db/transactions.go
AudriusButkevicius: Also, why are we using database as some sort of termination signal? Shouldn't we close the db once the App{} returns?

lib/fs/mtimefs.go
```diff
@@ -13,8 +13,8 @@ import (
 // The database is where we store the virtual mtimes
 type database interface {
 	Bytes(key string) (data []byte, ok bool)
-	PutBytes(key string, data []byte)
-	Delete(key string)
+	PutBytes(key string, data []byte) error
```
AudriusButkevicius: Bytes() above could also return an error, oppose to "it's not there".

lib/db/backend/backend.go
calmh: The underlying details vary... The idea with this interface is that you check it after the loop, that is if there's a read error then Next() will return false to stop the iteration and Error() will return why (as opposed to nil if the iteration just reached the end). Maybe I've missed to do that somewhere.

We could have Next() return an error directly instead, but then you need to compare that to some sentinel and the code becomes more cumbersome as you may want access to that error after the iteration so it cannot just become a `for Next() == nil` thing.

That Key() does not return an error is because there is no possibility for it to error once iteration to a key has succeeded. Value() *can* fail in at least the badger implementation because values are not always prefetched. So it's some abstraction leakage if you will.

lib/db/backend/leveldb_test.go
```diff
@@ -0,0 +1,45 @@
...
+	// Deleting a non-existent key is not an error
+
+	db := OpenLevelDBMemory()
+	defer db.Close()
+
+	err := db.Delete([]byte("a"))
+	if err != nil {
+		t.Error(err)
+	}
+}
```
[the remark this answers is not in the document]
calmh: That's correct, and once there is more than one database implementation we need to run them against both. (Badger doesn't work like this test requires, by default, which is why it exists.) Maybe at that point we factor out the tests and pass in an implementation.

lib/db/backend/leveldb_backend.go
```diff
@@ -0,0 +1,169 @@
...
+	if t.batch == nil {
+		t.batch = new(leveldb.Batch)
+	}
+	t.batch.Put(key, val)
+	return wrapLeveldbErr(t.checkFlush())
+}
+
+func (t *leveldbTransaction) Commit() error {
+	defer t.rel.Release()
+	defer t.leveldbSnapshot.Release()
```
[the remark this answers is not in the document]
calmh: This will result in a synchronous i/o most likely so I think inlining of tiny functions is a teensy consideration here. But sure it could be rewritten with a temp variable.

lib/db/set.go
```diff
@@ -145,73 +166,110 @@ func (s *FileSet) Update(device protocol.DeviceID, fs []protocol.FileInfo) {
...
+	}
 }
 
 func (s *FileSet) Get(device protocol.DeviceID, file string) (protocol.FileInfo, bool) {
-	f, ok := s.db.getFileDirty([]byte(s.folder), device[:], []byte(osutil.NormalizedFilename(file)))
+	f, ok, err := s.db.getFileDirty([]byte(s.folder), device[:], []byte(osutil.NormalizedFilename(file)))
+	if backend.IsClosed(err) {
+		return protocol.FileInfo{}, false
+	} else if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
calmh: These are just the same behavior we already had, to make this PR clearer by avoiding to refactoring all the rest of Syncthing to handle spurious db errors.

lib/db/transactions.go
calmh: This is just what we already did

lib/db/backend/backend.go
imsodin: Got it. It seems non-ideal to put the onus on the consumer to remember that after `.Next() == False` one needs to check `.Error()` to be sure the iteration aborted successfully at the end. Maybe an alternative approach would be to let `Release` return the error, thus it's at least guaranteed the consumer gets into contact with an error.

imsodin: My point is simply that it (distinction end reached and aborted due to error) seems very easy to miss and unfortunately I can't think of any simple/clean way to make it more in your face. So probably just what Audrius wrote somewhere: Add some comment to the interface specifying the behaviour (`Next` returns false on error and when reaching end, and `Error` needs to be called afterwards to detect if an error occurred). In principle this can be inferred from the interface, but I believe it's unlikely to be immediately obvious in practice.

calmh: Gotcha. I'm doing a revision soonish to handle the feedback and also some things that came out of the badger tests (where a transaction cannot be committed/discarded with active iterators, making the defers sometimes impractical).

AudriusButkevicius:
What are we doing with this? Do you plan to finish this, or are you abandoning it?

calmh:
Dude. Chill. I'm finishing it.

lib/db/backend/leveldb_test.go
```diff
@@ -0,0 +1,15 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package backend
+
+import "testing"
+
+func TestLevelDBBackendBehavior(t *testing.T) {
+	testBackendBehavior(t, func() Backend {
```
golangcibot: unlambda: replace `func() Backend {
	return OpenLevelDBMemory()
}` with `OpenLevelDBMemory` (from `gocritic`)

golangcibot: unlambda: replace `func() Backend {
	return OpenLevelDBMemory()
}` with `OpenLevelDBMemory` (from `gocritic`)

calmh:
I adressed some of the review comments. Two things I deliberately haven't: panic()ing in the top level methods, and returning errors from NamespacedKV. The reason for that is that both of those result in cascading changes everywhere else -- refactoring the pulling loop to be able to shut down on database error, etc. etc. I think this is large enough as-is, and the behavior w.r.t. when we panic or not is the same as it was previously. We should do that refactor, but I'm afraid that if I do  it as part of this PR we'll never get there.

lib/db/transactions.go
calmh: The bool here is specifically "does it exist" which is why we handle IsNotFound separately. Previously we just interpreted *any* error as "not found".

lib/db/transactions.go
calmh: I changed some of the defers, but I don't think

```
defer b()
return a()
```

is any worse than

```
err := a()
b()
return err
```

really?

calmh: I guess we can reorder in this case though without harm

lib/db/backend/backend.go
```diff
@@ -0,0 +1,116 @@
...
+	Close() error
+}
+
+type Tuning int
+
+const (
+	// N.b. these constants must match those in lib/config.Tuning!
+	TuningAuto Tuning = iota
+	TuningSmall
+	TuningLarge
```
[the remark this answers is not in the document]
calmh: Nope :)

lib/db/instance.go
```diff
@@ -177,25 +263,37 @@ func (db *instance) withHaveSequence(folder []byte, startSeq int64, fn Iterator)
...
 	defer dbi.Release()
 
 	var gk, keyBuf []byte
 	for dbi.Next() {
 		device, ok := db.keyer.DeviceFromDeviceFileKey(dbi.Key())
 		if !ok {
 			// Not having the device in the index is bad. Clear it.
-			t.Delete(dbi.Key())
-			t.checkFlush()
+			if err := t.Delete(dbi.Key()); err != nil {
```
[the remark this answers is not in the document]
calmh: release/close don't return errors. They are deferred to ensure cleanup in case we return early due to some other error. The method then ends with `return commit()` which does pass the final error.

lib/db/set.go
```diff
@@ -285,21 +371,47 @@ func (s *FileSet) ListDevices() []protocol.DeviceID {
...
+	}
 
 	// Also clean out the folder ID mapping.
-	db.folderIdx.Delete([]byte(folder))
+	if err := db.folderIdx.Delete([]byte(folder)); backend.IsClosed(err) {
+		return
+	} else if err != nil {
+		panic(err)
```
AudriusButkevicius: In theory all of this should be in a "transaction"?

AudriusButkevicius:
Also unresolved the namespacedkv Bytes() comments as it seems they are not addressed, nor commented on.

In general, I think the reviewer should be resolving the comments once the discussion concludes or he agrees with the code changes provided. If you address a comment just drop a note as addressed so the reviewer knows its done.

Now I have to go double check every resolved comment to see if you left any feedback, if you did anything about it or not, as it's resolves with no message of what was done, did you ignore it, change code, or what.

lib/db/backend/leveldb_backend.go
```diff
@@ -0,0 +1,165 @@
...
+	}, nil
+}
+
+func (b *leveldbBackend) NewWriteTransaction() (WriteTransaction, error) {
+	snap, err := b.ldb.GetSnapshot()
+	if err != nil {
+		return nil, wrapLeveldbErr(err)
+	}
+	return &leveldbTransaction{
+		leveldbSnapshot: leveldbSnapshot{
```
imsodin: Could reuse `NewReadTransaction` instead.

lib/db/backend/leveldb_backend.go
```diff
@@ -0,0 +1,165 @@
...
+type leveldbTransaction struct {
+	leveldbSnapshot
+	ldb   *leveldb.DB
+	batch *leveldb.Batch
+	rel   *releaser
+}
+
+func (t *leveldbTransaction) Delete(key []byte) error {
+	t.batch.Delete(key)
+	return wrapLeveldbErr(t.checkFlush())
```
imsodin: That reintroduces the problem with non-atomic db updates: Previously `checkFlush` was only called when all db operations that need to go through (or not) together were added to the batch. The only options I see are to either expose `checkFlush`, guaranteeing that the db is only written in those cases, or add a batch-like part to the interface.

lib/db/backend/leveldb_open.go
```diff
@@ -0,0 +1,226 @@
+// Copyright (C) 2019 The Syncthing Authors.
```
imsodin: Local git considers leveldb_open.go as a copy of lowlevel.go (52%), thus maybe leave the copyright year at 2018.

lib/syncthing/utils.go
```diff
@@ -124,5 +125,9 @@ func copyFile(src, dst string) error {
 }
 
 func OpenGoleveldb(path string, tuning config.Tuning) (*db.Lowlevel, error) {
-	return db.Open(path, db.Tuning(tuning))
+	ldb, err := backend.Open(path, backend.Tuning(tuning))
+	if err != nil {
+		return nil, err
+	}
+	return db.NewLowlevel(ldb), nil
```
imsodin: I'd like to use this change to make the interface of `lib/syncthing` deal with `backend.Backend` instead of `*db.Lowlevel`. Or do you want to defer that to a separate PR.

lib/db/util_test.go
```diff
@@ -54,7 +58,7 @@ func openJSONS(file string) (*leveldb.DB, error) {
 			return nil, err
 		}
 
-		db.Put(row["k"], row["v"], nil)
+		_ = db.Put(row["k"], row["v"])
```
imsodin: Why don't you return the error?

lib/db/transactions.go
```diff
@@ -7,111 +7,152 @@
...
 
-func (t readOnlyTransaction) getFileByKey(key []byte) (protocol.FileInfo, bool) {
-	if f, ok := t.getFileTrunc(key, false); ok {
-		return f.(protocol.FileInfo), true
+func (t readOnlyTransaction) getFileByKey(key []byte) (protocol.FileInfo, bool, error) {
+	f, ok, err := t.getFileTrunc(key, false)
+	if err != nil {
+		return protocol.FileInfo{}, false, err
+	}
+	if !ok {
```
imsodin: Could drop this and use `err != nil || !ok` above

lib/db/transactions.go
```diff
@@ -7,111 +7,152 @@
...
+		return nil, nil, false, err
 	}
-
-	return keyBuf, nil, false
+	fi, ok, err := t.getFileTrunc(keyBuf, truncate)
+	if !ok {
+		return keyBuf, nil, false, nil
+	}
+	if err != nil {
+		return nil, nil, false, err
```
imsodin: Can still return `keyBuf` and thus also merge the two ifs. Sure we're probably panicking thus making reuse not really an issue, but it also doesn't hurt.

lib/db/transactions.go
```diff
@@ -121,24 +162,33 @@ func (t readWriteTransaction) updateGlobal(gk, keyBuf, folder, device []byte, fi
...
-		} else {
-			// This file must exist in the db, so this must be caused
-			// by the db being closed - bail out.
+		keyBuf, err = t.keyer.GenerateDeviceFileKey(keyBuf, folder, fl.Versions[0].Device, name)
+		if err != nil {
+			return nil, false, err
+		}
+		new, ok, err := t.getFileByKey(keyBuf)
+		if err != nil {
+			return nil, false, err
```
imsodin: Same here: I don't see a reason to not return `keyBuf` and thus can combine with if below.

And I'd return `keyBuf` instead of nil on error also in all the other instances. I'll stop commenting to have less conversations to hide if fixed/dismissed.

lib/db/set.go
```diff
@@ -285,21 +371,47 @@ func (s *FileSet) ListDevices() []protocol.DeviceID {
 // database.
 func DropFolder(ll *Lowlevel, folder string) {
 	db := newInstance(ll)
-	db.dropFolder([]byte(folder))
-	db.dropMtimes([]byte(folder))
-	db.dropFolderMeta([]byte(folder))
```
imsodin:
```
for f := range []func([]byte) error{db.dropFolder, db.dropMtimes, db.dropFolderMeta, db.folderIdx.Delete} {
    if err := f([]byte(folder)); ...
```

lib/db/set.go
```diff
@@ -285,21 +371,47 @@ func (s *FileSet) ListDevices() []protocol.DeviceID {
...
 
 // DropDeltaIndexIDs removes all delta index IDs from the database.
 // This will cause a full index transmission on the next connection.
 func DropDeltaIndexIDs(db *Lowlevel) {
-	dbi := db.NewIterator(util.BytesPrefix([]byte{KeyTypeIndexID}), nil)
+	dbi, err := db.NewPrefixIterator([]byte{KeyTypeIndexID})
+	if backend.IsClosed(err) {
+		return
+	} else if err != nil {
+		return
```
imsodin: panic?

lib/db/backend/leveldb_backend.go
calmh: This is somewhat of an issue. Frankly one of the motivations behind this work is to move to something that provides real transactions (whether that's still goleveldb or something else) and I'm not convinces the current guarantees are what you think. Having multiple writes go out as a batch doesn't make it atomic, it just means there are multiple non-atomic writes happening in close temporal succession... I agree though that this change makes it strictly less atomic, whatever the state before.

For what it's worth the Badger interface is the same. Badger can at any point in a transaction return ErrTransactionTooBig at which point the thing to do is to commit it, create a new one, and continue onwards...

lib/syncthing/utils.go
calmh: I'm not sure we want to expose backend outside the db package. Can we explore later?

lib/db/transactions.go
calmh: I really think that in this error case there is no continuation possible and it's cleaner to return all the nils to indicate that all bets are off at this point. Returning the keybuf, so far as it indicates anything at all, indicates that just this little thing didn't work out quite right but please continue calling more db functions, which isn't really the thing...

lib/db/set.go
calmh: I think that's a lot less readable than just having each call on its own line :)

calmh:
@AudriusButkevicius yeah sorry about the resolvingness, that was the wrong thing to do.

As for namedspacedkv I've tried to answer that a couple of times. :) I'm deliberately avoiding to change the external API of the db package to return errors where we currently don't, as that rabbit whole goes *deep*...

lib/db/set.go
calmh: In principle... In practice that looks very infections once I start pulling the threads, almost the same as removing the top-level write operations completely. Let me see what I can do.

lib/db/backend/leveldb_backend.go
imsodin: Ok, maybe I shouldn't have brought up "atomic". However this clearly makes it worse, as before we were vulnerable to an inconsistent db while "there are multiple non-atomic writes happening in close temporal succession", now we are vulnerable during the entire time an implementation thinks it doesn't yet need to flush a transaction.

> We don't create large transactions though, except in certain "internal" loops like when deleting a folder, as all other updates from the outside come in as small batches anyway.

Possibly, but that means we depend on implementation details, namely that model-side batching creates batches small enough that the backend implementation commits it as a whole.

And I don't know what the properties of a "real transactions" are - could you quickly explain what exactly you want there?

lib/db/transactions.go
imsodin: My point is that I believe the distinction (whether returning `nil` or `keyBuf`) really doesn't matter at all when `err != nil` and I consider a second, separate if branche noise that reduces readability. I don't think anyone would even look at `keyBuf` when encountering an error, even less go like "lets do something drastic if err != nil, unless keyBuf != nil, then try to recover somehow, because apparently they want me to keep using that buffer".

lib/db/backend/leveldb_backend.go
calmh: What I want is for things to be atomically committed once we call commit, all or nothing. Currently there's no guarantee -- in fact, when there *is* a problem it would probably be quite unlikely for it to happen precisely *between* two batches. I mean, if there's an I/O error or the disk becomes full it will probably happen when we are in fact writing, like in the middle of a batch, not when we are *not* writing. :)

But I'm still thinking of the best way to handle this. We could add some sort of Checkpoint() method which is advisory that if a transaction / batch needs to be broken, this would be a good time to do it... (I.e., that current checkFlush)

lib/db/transactions.go
calmh: OK, fair

lib/db/set.go
calmh: Actually you are right, I was just looking at the minus diff not the new code.

lib/db/backend/leveldb_backend.go
imsodin: > What I want is for things to be atomically committed once we call commit, all or nothing. [...]

Cool, I very much want that too - that was my only theory, but I dismissed it as too unrealistic :D 

Well, OOM, power-out, ... are still possibilities.

And how to do it: I was thinking about adding a simple `*Batch` type that contains keys to put/delete and a `Apply(b *Batch)` to the writer interface with the requirement that the writer must flush either before or after all the operations in batch.

lib/db/transactions.go
imsodin: Good point too - might have asked Audrius for a third opinion before going through, but maybe it's just as well to stop bikeshedding at this point (sorry about that ;) ).

AudriusButkevicius:
I think the namespaced change should be a very shallow rabbit hole, but we can always fix it later.

lib/db/backend/leveldb_backend.go
calmh: I don't want more types, as Batch on top of Transaction on top of DB doesn't make sense on its own, imho. But I added the Checkpoint thing to be roughly equivalent with the already existing checkFlush. The behavior is sort of unspecified ("maybe it commits, maybe it doesn't, it might affect your reads") but that's what we already had and what we get from leveldb...

imsodin: With leveldb I don't understand why the read ambiguity is needed. A leveldb snapshot is just the state of a the db at a specific point in time - we can't modify or don't release/resnapshot when flushing the buffer associated with the snapshot making up a read-write-transaction. The only unspecified behaviour I get is that we can't require "checkpointing" to only happen on calling `Checkpoint`, but just tell the implementation should checkpoint then if `Checkpoint` is called regularly. If we do require this ambiguity about future reads on `Checkpoint`, if anyway possible `Checkpoint` should get an additional return value that signifies whether the transaction has been broken thus (further reads can change) or not. 

Another thing that should be stated explicitly: I assume `Commit` is final (i.e. no more writes can/should happen after calling it)?

calmh: You're right, I was mixing up things. There is a snapshot so there is no ambiguity. There is also no need for such an ambiguity in the future either as we can always have a read-snap + write-tran.

calmh:
@imsodin About Tuning, I'm not sure how to handle that. At the moment I pretended it was a general thing that other potential databases could support. But if the leveldb comes from the backend package, that's where the type needs to live... Having it at the `db` level would be an import cycle. I guess the leveldb backend could be a package of its own? That would leave the backend package pretty much just a couple of interfaces and make it harder to do things like reuse tests.

imsodin:
My proposal may be more skirting than solving the issue, but I'd still like it better than the current state: Remove the "pseude-backend-agnostic" `backend.Open` and `backend.OpenMemory` and readd `db.Open` and `db.OpenMemory` that directly call the specific `db.OpenLevelDB`/`db.OpenLevelMemory`.

calmh:
My point in having the `backend` *package* to begin with is to make sure that there is no dependency from the larger `db` package directly to an implementation -- it can only come via the `Backend` interface, and the `backend` package itself is quite light as it's just the low level wrapper types. I don't think removing the `backend` package and moving all the types up one level improves the situation really, or I'm not understanding what you're suggesting.

calmh:
That was my intention... I think from a user perspective we should only have kind of database at any given point. We might need to have an internal migration step if we do change the backend, but that should be essentially invisible from the outside...

PR #6109, opened 2019-10-26
gui: Show a warning when authentication is not configured but listening on localhost (fixes #4703)

### LINKED ISSUE #4703: Suggest setting up auth on initial startup

e-sch:
### Purpose

This shows a panel to warn the user that the UI can be accessed by other users on the same machine. It will only show when listening on localhost and remote access is not possible as there is already a warning for that.

The panel can be dismissed and it will not show up again across reloads and restarts.

I wasn't sure whether I should use the panel-warning or panel-info CSS class, can someone make a suggestion?

### Testing
- Set `STGUIADDRESS` to `127.0.0.1:8384` and disable auth -> "Warning!" panel appears
- Set `STGUIADDRESS` to `0.0.0.0:8384` and disable auth -> "Danger!" panel appears (can not be dismissed)
- Click `Do not show again` and reload page / restart Syncthing -> warning panel does not appear
- Set `STGUIADDRESS` to `127.0.0.1:8384` and setup auth -> warning panel does not appear

### Screenshots
2019-10-27-000928_1187x618_scrot


lib/config/config_test.go
```diff
@@ -68,6 +68,7 @@ func TestDefaultValues(t *testing.T) {
 		OverwriteRemoteDevNames: false,
 		TempIndexMinBlocks:      10,
 		UnackedNotificationIDs:  []string{},
+		DismissedPanels:         []string{},
```
imsodin: I think you could just use `UnackedNotificationIDs` for this. It's admittedly not it's original purpose, but I'd personally still rather stretch its scope than add a new config field just for this panel.

gui/default/index.html
```diff
@@ -122,6 +122,39 @@ <h3 class="panel-title">
...
+              <div class="panel-icon">
+                <span class="fas fa-exclamation-circle"></span>
+              </div>
+              <span translate>Warning!</span>
+            </h3>
+          </div>
+          <div class="panel-body">
+            <p>
+            <span translate>If this is a shared computer, other users might be able to access the admin interface.</span>
+            <span translate>Please set a GUI Authentication User and Password in the Settings dialog.</span>
```
imsodin: I wouldn't capitalize user, password and settings. And I'd word it something like "You can set a user and password in the GUI tab of the settings dialog to restrict access".

gui/default/index.html
```diff
@@ -122,6 +122,39 @@ <h3 class="panel-title">
...
+            <h3 class="panel-title">
+              <div class="panel-icon">
+                <span class="fas fa-exclamation-circle"></span>
+              </div>
+              <span translate>Warning!</span>
+            </h3>
+          </div>
+          <div class="panel-body">
+            <p>
+            <span translate>If this is a shared computer, other users might be able to access the admin interface.</span>
```
imsodin: Is this intentionally conditional? Otherwise just "other users can access".

lib/config/config_test.go
AudriusButkevicius: Or use the same mechanism that UnackedNotificationIDs  uses to show a one-off modal like we do for other stuff, just change the display condition to have the additional constraints on addresses and other stuff.

st-review:
:robot: *beep boop*

PR #6114, opened 2019-10-30
Fix removal of a marker when there are still folders referencing it (fixes #6106)

### LINKED ISSUE #6106: .stfolder is deleted on folder removal even if the same folder is still being shared in other places
1. Open the GUI of a clean Syncthing installation. The default folder list should look like this.

image

2. Add a new folder pointing to the exactly same path as the "Default Folder".

image

3. Remove the newly created folder using the GUI, and "rescan" the "Default Folder".

image

The following error message is shown,

`Error on folder "Default Folder" (default): folder marker missing`

and the folder is not being synced anymore.

I think that it would be useful if Syncthing did not remove `.stfolder ` in a case where the same folder is still being shared in other places.

One can ask, why would you share the same folder multiple times? I personally have encountered this error when trying to share a folder from Android, and then realising that I would rather share a folder from a different device which would point at the same path. Thus, I accepted the newly shared folder, and then removed the one created previously in Android, as it was no longer needed. The result of the operation was as above.

This was just a simple mistake of mine, but I can also easily imagine valid reasons for sharing the same folder multiple times, such as having different ignore patterns for each of them, etc. In addition to that, to actually fix the problem, one needs to manually recreate the missing `.stfolder` which, depending on the operating system, may not be straightforward.

### Version Information

Syncthing Version: v1.3.0

ruslanye:
### Purpose

I modified RemoveFolder to fix #6106 
After the changes, if someone tries to remove a folder, path of which is still referenced, the `.stfolder` won't be removed.

### Testing

I checked changes manually. I'm not sure if unit tests are needed.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -389,8 +389,17 @@ func (m *model) RemoveFolder(cfg config.FolderConfiguration) {
 	m.fmut.Lock()
 	defer m.fmut.Unlock()
 
-	// Delete syncthing specific files
-	cfg.Filesystem().RemoveAll(config.DefaultMarkerName)
+	isPathUnique := true
+	for folderID, folderCfg := range m.folderCfgs {
+		if folderID != cfg.ID && folderCfg.Path == cfg.Path {
+			isPathUnique = false
+			break
+		}
+	}
+	if isPathUnique {
+		// Delete syncthing specific files
+		cfg.Filesystem().RemoveAll(config.DefaultMarkerName)
+	}
 
 	m.tearDownFolderLocked(cfg, fmt.Errorf("removing folder %v", cfg.Description()))
 	// Remove it from the database


lib/model/model.go
```diff
@@ -389,8 +389,17 @@ func (m *model) RemoveFolder(cfg config.FolderConfiguration) {
...
+	isPathUnique := true
+	for folderID, folderCfg := range m.folderCfgs {
+		if folderID != cfg.ID && folderCfg.Path == cfg.Path {
+			isPathUnique = false
+			break
+		}
+	}
+	if isPathUnique {
+		// Delete syncthing specific files
+		cfg.Filesystem().RemoveAll(config.DefaultMarkerName)
```
golangcibot: Error return value of `(github.com/syncthing/syncthing/lib/fs.Filesystem).RemoveAll` is not checked (from `errcheck`)

PR #6124, opened 2019-11-03
gui, lib/api: Adds support for prefers-color-scheme (fixes #6115)

### LINKED ISSUE #6115: Change Web UI based on prefers-color-scheme
### What happens

The Web UI has the same theme regardless of the preferred color scheme the user has set on their browser.

### What I expected to happen

The Web UI is dark when the [`prefers-color-scheme` property](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) is set to `dark` or the default one when it is `light` or `no-preference`.

### Version Information

Syncthing Version: v1.3.0
OS Version: Linux Mint 18.1
Browser Version: Firefox 70.0 64 bit

mx-psi:
### Purpose

Fixes #6115. In particular:

- moves the default theme into a new "light" theme,
- serves assets from other themes when the path starts with `theme-assets` and
- on the default theme, load the correct style depending on `prefers-color-scheme`.

### Testing

*Instructions on how to activate dark mode on Firefox for testing are available [here](https://stackoverflow.com/a/56757527).*

- Open web UI with `prefers-color-scheme` set to `light` or `no-preference` → it has the light theme
- Open web UI with `prefers-color-scheme` set to `dark` → it has the dark theme
- Open web UI in a browser without support for `prefers-color-scheme` → it has the light theme
- Set "Light" theme → the web UI has the current default theme (regardless of `prefers-color-scheme`)
- Set a custom theme → the web UI has the custom theme (regardless of `prefers-color-scheme` or whether it is supported)

### DIFF
--- a/gui/default/assets/css/theme.css
+++ b/gui/default/assets/css/theme.css
@@ -7,5 +7,5 @@
 
 */
 
-@import "/dark/assets/css/theme.css" (prefers-color-scheme: dark);
-@import "/light/assets/css/theme.css" (prefers-color-scheme: light), (prefers-color-scheme: no-preference);
+@import "/theme-assets/dark/assets/css/theme.css" (prefers-color-scheme: dark);
+@import "/theme-assets/light/assets/css/theme.css" (prefers-color-scheme: light), (prefers-color-scheme: no-preference);
--- a/gui/default/assets/css/theme.css
+++ b/gui/default/assets/css/theme.css
@@ -7,5 +7,5 @@
 
 */
 
-@import "/theme-assets/dark/assets/css/theme.css" (prefers-color-scheme: dark);
+@import "/theme-assets/dark/assets/css/theme.css" screen and (prefers-color-scheme: dark);
 @import "/theme-assets/light/assets/css/theme.css" (prefers-color-scheme: light), (prefers-color-scheme: no-preference);
--- a/gui/default/assets/css/theme.css
+++ b/gui/default/assets/css/theme.css
@@ -7,29 +7,5 @@
 
 */
 
-.panel-progress {
-    background: #3498db;
-}
-
-.identicon rect {
-    fill: #333;
-}
-
-.panel-warning .identicon rect {
-    fill: #fff;
-}
-
-.li-column {
-    background-color: rgb(236, 240, 241);
-    border-radius: 3px;
-}
-
-.panel-heading:hover, .panel-heading:focus {
-    text-decoration: none;
-}
-
-.fancytree-ext-filter-hide tr.fancytree-submatch span.fancytree-title,
-.fancytree-ext-filter-hide span.fancytree-node.fancytree-submatch span.fancytree-title {
-    color: black !important;
-    font-weight: lighter !important;
-}
+@import "/dark/assets/css/theme.css" (prefers-color-scheme: dark);
+@import "/light/assets/css/theme.css" (prefers-color-scheme: light), (prefers-color-scheme: no-preference);
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -389,6 +389,7 @@ angular.module('syncthing.core')
             });
 
             refreshNoAuthWarning();
+            setDefaultTheme();
 
             if (!hasConfig) {
                 $scope.$emit('ConfigLoaded');
@@ -649,6 +650,17 @@ angular.module('syncthing.core')
             $scope.remoteNeed = {};
             $scope.remoteNeedFolders = [];
             $scope.remoteNeedDevice = undefined;
+        }
+
+        function setDefaultTheme(){
+          // If theme is default and there is no support for prefers-color-scheme
+          if ($scope.config.gui.theme === "default" && window.matchMedia('(prefers-color-scheme: dark)').media === 'not all') {
+            document.documentElement.style.display = 'none';
+            document.head.insertAdjacentHTML(
+              'beforeend',
+              '<link rel="stylesheet" href="/theme-assets/light/assets/css/theme.css" onload="document.documentElement.style.display = \'\'">'
+            );
+          }
         }
 
         function saveIgnores(ignores, cb) {
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -652,15 +652,21 @@ angular.module('syncthing.core')
             $scope.remoteNeedDevice = undefined;
         }
 
-        function setDefaultTheme(){
-          // If theme is default and there is no support for prefers-color-scheme
-          if ($scope.config.gui.theme === "default" && window.matchMedia('(prefers-color-scheme: dark)').media === 'not all') {
-            document.documentElement.style.display = 'none';
-            document.head.insertAdjacentHTML(
-              'beforeend',
-              '<link rel="stylesheet" href="/theme-assets/light/assets/css/theme.css" onload="document.documentElement.style.display = \'\'">'
-            );
-          }
+
+        function setDefaultTheme() {
+            if (!document.getElementById("fallback-theme-css")){
+
+                // check if no support for prefers-color-scheme
+                var colorSchemeNotSupported = typeof window.matchMedia === "undefined" || window.matchMedia('(prefers-color-scheme: dark)').media === 'not all';
+
+                if ($scope.config.gui.theme === "default" && colorSchemeNotSupported) {
+                    document.documentElement.style.display = 'none';
+                    document.head.insertAdjacentHTML(
+                      'beforeend',
+                      '<link id="fallback-theme-css" rel="stylesheet" href="/theme-assets/light/assets/css/theme.css" onload="document.documentElement.style.display = \'\'">'
+                    );
+                }
+            }
         }
 
         function saveIgnores(ignores, cb) {
--- a/lib/api/api_statics.go
+++ b/lib/api/api_statics.go
@@ -23,6 +23,8 @@ import (
 	"github.com/syncthing/syncthing/lib/sync"
 )
 
+const themePrefix = "theme-assets/"
+
 type staticsServer struct {
 	assetDir        string
 	assets          map[string][]byte
@@ -89,7 +91,6 @@ func (s *staticsServer) serveAsset(w http.ResponseWriter, r *http.Request) {
 	s.mut.RUnlock()
 
 	// If path starts with special prefix, get theme and file from path
-	const themePrefix = "theme-assets/"
 	if strings.HasPrefix(file, themePrefix) {
 		path := file[len(themePrefix):]
 		i := strings.IndexRune(path, '/')
--- a/lib/api/api_statics.go
+++ b/lib/api/api_statics.go
@@ -88,6 +88,21 @@ func (s *staticsServer) serveAsset(w http.ResponseWriter, r *http.Request) {
 	theme := s.theme
 	s.mut.RUnlock()
 
+	// If path starts with special prefix, get theme and file from path
+	const themePrefix = "theme-assets/"
+	if strings.HasPrefix(file, themePrefix) {
+		path := file[len(themePrefix):]
+		i := strings.IndexRune(path, '/')
+
+		if i == -1 {
+			http.NotFound(w, r)
+			return
+		}
+
+		theme = path[:i]
+		file = path[i+1:]
+	}
+
 	// Check for an override for the current theme.
 	if s.assetDir != "" {
 		p := filepath.Join(s.assetDir, theme, filepath.FromSlash(file))
@@ -120,13 +135,8 @@ func (s *staticsServer) serveAsset(w http.ResponseWriter, r *http.Request) {
 		// Check for a compiled in default asset.
 		bs, ok = s.assets[config.DefaultTheme+"/"+file]
 		if !ok {
-
-			// Check for a compiled outside of current theme
-			bs, ok = s.assets[file]
-			if !ok {
-				http.NotFound(w, r)
-				return
-			}
+			http.NotFound(w, r)
+			return
 		}
 	}
 
--- a/lib/api/api_statics.go
+++ b/lib/api/api_statics.go
@@ -120,8 +120,13 @@ func (s *staticsServer) serveAsset(w http.ResponseWriter, r *http.Request) {
 		// Check for a compiled in default asset.
 		bs, ok = s.assets[config.DefaultTheme+"/"+file]
 		if !ok {
-			http.NotFound(w, r)
-			return
+
+			// Check for a compiled outside of current theme
+			bs, ok = s.assets[file]
+			if !ok {
+				http.NotFound(w, r)
+				return
+			}
 		}
 	}
 
--- a/gui/light/assets/css/theme.css
+++ b/gui/light/assets/css/theme.css
@@ -1,0 +1,35 @@
+/*
+// Copyright (C) 2016 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+*/
+
+.panel-progress {
+    background: #3498db;
+}
+
+.identicon rect {
+    fill: #333;
+}
+
+.panel-warning .identicon rect {
+    fill: #fff;
+}
+
+.li-column {
+    background-color: rgb(236, 240, 241);
+    border-radius: 3px;
+}
+
+.panel-heading:hover, .panel-heading:focus {
+    text-decoration: none;
+}
+
+.fancytree-ext-filter-hide tr.fancytree-submatch span.fancytree-title,
+.fancytree-ext-filter-hide span.fancytree-node.fancytree-submatch span.fancytree-title {
+    color: black !important;
+    font-weight: lighter !important;
+}


gui/default/assets/css/theme.css
```diff
@@ -7,29 +7,5 @@
...
-.panel-heading:hover, .panel-heading:focus {
-    text-decoration: none;
-}
-
-.fancytree-ext-filter-hide tr.fancytree-submatch span.fancytree-title,
-.fancytree-ext-filter-hide span.fancytree-node.fancytree-submatch span.fancytree-title {
-    color: black !important;
-    font-weight: lighter !important;
-}
+@import "/dark/assets/css/theme.css" (prefers-color-scheme: dark);
```
AudriusButkevicius: I don't like this path, this should either be be assets/<theme>/<stuff>
or 
themes/<theme>/assets/<stuff>
or 
/assets/css/theme.css?theme=dark (this is probably not very cache friendly).
or something like that. 

Random stuff rooted feels sad.

mx-psi: I am not sure if I understand what path you want in the first two cases.
For example, in the second option, if the path is not rooted, as in:
```css
@import "theme-assets/dark/assets/css/theme.css" (...);
```
the browser would request `/assets/css/theme-assets/dark/assets/css/theme.css` to the server, so I guess I would have to exclude any path that has `theme-assets` in it to handle it. Is that what you want?

AudriusButkevicius: Sorry, what I mean is that I do not want localhost:8483/dark/foo/bar, as /dark is at the root of webserver, does not describe that is it, or what it represents.

AudriusButkevicius: We could have 
/theme-specific-assets/dark/css/theme.css

mx-psi: Okay, I see now what you meant. Thank you for clarifying!

mx-psi:
The path for getting an asset for a different theme is now `theme-assets/<theme>/<stuff>`.

Ferroin:
This needs to fall back correctly if the media query is not supported.

As per https://caniuse.com/#feat=prefers-color-scheme, it's currently not supported by:

* Internet Explorer (will never have support)
* Edge, prior to the Chromium-based redesign that's supposedly coming out any day now.
* iOS Safari version 12.4 and older.
* Opera Mini (will never have support)
* Opera Mobile
* UC Browser for Android
* Samsung Internet
* An assortment of other low-usage browsers

Unless I'm mistaken, all the queries this is making right now will evaluate to false on such browsers.

I'm pretty sure that handling this is going to require dropping to JS, though at that point it makes more sense to just use JS to set a cookie (or header, or whatever else) to pass the info back to the server component.

lib/api/api_statics.go
```diff
@@ -88,6 +88,21 @@ func (s *staticsServer) serveAsset(w http.ResponseWriter, r *http.Request) {
 	theme := s.theme
 	s.mut.RUnlock()
 
+	// If path starts with special prefix, get theme and file from path
+	const themePrefix = "theme-assets/"
```
AudriusButkevicius: Should declare at module scope.

calmh:
I haven't tested it - does it work? On both browsers with *and without* (I don't see any acknowledgement of @Ferroin s concerns above) support for this magic?

mx-psi:
@Ferroin is right, it does not fall back correctly in a browser without support. `theme.css` from the light theme changes very little so it is difficult to notice as of now, but this would become a problem with any future changes to the light theme if older browsers still need to be supported.

A possible solution with the current code is to import the light theme unconditionally first and rely on CSS order, as in:
```css
@import "/theme-assets/light/assets/css/theme.css";
@import "/theme-assets/dark/assets/css/theme.css" (prefers-color-scheme: dark);
```
This works but care should be taken in the future to avoid any rules on the light theme having more specificity than the ones in the dark theme.

If that is not a feasible option, I see no other way but to use JS (for example as [web.dev suggests here](https://web.dev/prefers-color-scheme/#finding-out-if-dark-mode-is-supported-by-the-browser)). I am not familiar with Angular so I would probably need some help to implement that.

AudriusButkevicius:
Can't we do this in js? In that case we would be able to check the absence of value.
Also, we probably don't want to use dark theme if someone tries to print the page as described in the post.

mx-psi:
I have used the article's code to include support for older browsers and tested it with an older Firefox version. As I said, I am not familiar with Angular so please check that I haven't messed up anything.

I have also done some testing with regards to printing and it looks like in recent versions of Firefox and Chromium it works just fine (maybe they ignore the color scheme when printing?).

tomasz1986:
I am not a Syncthing developer, so you can ignore my advice, but honestly this looks very hacky. The whole `prefers-color-scheme` is a part of the Media Queries Level 5, which at the moment is still just at the stage of being called "Editor’s Draft, 16 August 2019".

https://drafts.csswg.org/mediaqueries-5

> Status of this document
>
>This is a public copy of the editors’ draft. It is provided for discussion only and may change at any moment. Its publication here does not imply endorsement of its contents by W3C. Don’t cite this document other than as work in progress.

The popular Web browsers only started to support it a few months ago, so everything released prior to that has no support for the feature at all. If you want to implement support for such an experimental feature, I believe that it would be a good idea to do it in the most unobtrusive way possible, with no risk of affecting the older browsers.

I personally think that the method of just overriding the light theme values mentioned above

```
@import "/theme-assets/light/assets/css/theme.css";
@import "/theme-assets/dark/assets/css/theme.css" (prefers-color-scheme: dark);
```

is much safer, as the unsupported media query will be simply ignored by the old browsers, with no need for a JavaScript bandaid.

mx-psi:
Even if the official document is still technically a draft, I think the feature is mature enough for Syncthing to support it. [Roughly three quarters of users](https://caniuse.com/#feat=prefers-color-scheme) could benefit from it right now and the feature seems stable: the issues at [WebKit](https://bugs.webkit.org/show_bug.cgi?id=190499), [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=889087)  and [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1494034) have been closed for months and the existing [CSSWG Drafts issues](https://github.com/w3c/csswg-drafts/issues?q=is%3Aissue+prefers-color-scheme+is%3Aopen+sort%3Aupdated-desc) do not discuss any changes to the syntax or the expected behavior of the feature.

That being said, I also favor the method that does not use Javascript as @tomasz1986 suggests, since I find it simpler to maintain and update, but of course I am not a Syncthing developer either.

AudriusButkevicius:
Actually thinking about this further, I think we might be able to do this using css hackery.  Is there a way to do something like "if you support this property include this file otherwise don't"? Or browsers that do not support it just defaults to matching the selector?

tomasz1986:
> Even if the official document is still technically a draft, I think the feature is mature enough for Syncthing to support it. [Roughly three quarters of users](https://caniuse.com/#feat=prefers-color-scheme) could benefit from it right now and the feature seems stable: the issues at [WebKit](https://bugs.webkit.org/show_bug.cgi?id=190499), [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=889087) and [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1494034) have been closed for months and the existing [CSSWG Drafts issues](https://github.com/w3c/csswg-drafts/issues?q=is%3Aissue+prefers-color-scheme+is%3Aopen+sort%3Aupdated-desc) do not discuss any changes to the syntax or the expected behavior of the feature.

I personally would not believe any browser usage statistics, as they are heavily skewed by adblockers, which usually also block analytics, and adblocker usage in some countries is as high as 40-50%. Also, Android users are often stuck with a very old version of WebView, which Syncthing uses to display the GUI, regardless of the actual web browsers installed. In addition to that, IE11 support is a must if you care about any enterprise or government systems.

Nevertheless, my point was just to be very careful when implementing such a feature, not to not implement it all :wink:.

> Actually thinking about this further, I think we might be able to do this using css hackery. Is there a way to do something like "if you support this property include this file otherwise don't"? Or browsers that do not support it just defaults to matching the selector?

The above-mentioned CSS should do it. The feature shall be ignored in the browsers which have no support for it. If you want to exclude printing too, then the following code will probably suffice.

```
@import "/theme-assets/light/assets/css/theme.css";
@import "/theme-assets/dark/assets/css/theme.css" screen and (prefers-color-scheme: dark);
```

This should limit the automatic dark theme application only to screens (i.e. exclude printers and screen readers), and only when the `prefers-color-scheme` rule is supported and set to `dark`. Before the actual implementation though, it would be good to test this at least in IE11 and older Android.

tomasz1986:
I have actually just tested the changes both in Chromium, Firefox, and IE11.

1. As predicted, IE11 simply ignores the conditional media query and uses the light theme as default.

2. Firefox applies the dark theme correctly, but switches automatically to black text and white background for printing (with no option to keep the colours?).

3. Chromium, on the other hand, also applies the dark theme correctly, but when it comes to printing, it has an option to print "background graphics".

image

Now, using just `@import "/theme-assets/dark/assets/css/theme.css" (prefers-color-scheme: dark);` results in printing a dark-themed page (and probably eating a lot of ink). To prevent this, `@import "/theme-assets/dark/assets/css/theme.css" screen and (prefers-color-scheme: dark);` is required.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -651,6 +652,17 @@ angular.module('syncthing.core')
...
         }
 
+        function setDefaultTheme(){
+          // If theme is default and there is no support for prefers-color-scheme
+          if ($scope.config.gui.theme === "default" && window.matchMedia('(prefers-color-scheme: dark)').media === 'not all') {
+            document.documentElement.style.display = 'none';
+            document.head.insertAdjacentHTML(
+              'beforeend',
+              '<link rel="stylesheet" href="/theme-assets/light/assets/css/theme.css" onload="document.documentElement.style.display = \'\'">'
+            );
```
AudriusButkevicius: Why are we setting something to display none? Also, on a browser that does not support this, we'd be repeatedly injecting the same element on every config update (even if nothing has changed).

gui/default/syncthing/core/syncthingController.js
```diff
@@ -651,6 +652,17 @@ angular.module('syncthing.core')
             $scope.remoteNeedDevice = undefined;
         }
 
+        function setDefaultTheme(){
```
AudriusButkevicius: Space after closing brace

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: Also, I guess this ends up being a third css file we load if there is no support?
Or does it not load the other two?
If it does not, cannot we just match on 'not all' in css?

AudriusButkevicius: Or use a not(selector) ?

gui/default/syncthing/core/syncthingController.js
mx-psi: Do you use any JS linter/formatter? You suggest using `go fmt` and `go lint` in the contributing guidelines but there is no similar tool suggested for JS.

gui/default/syncthing/core/syncthingController.js
mx-psi: > Why are we setting something to display none?

This is suggested by [web.dev](https://web.dev/prefers-color-scheme/#loading-strategy): "to avoid a [flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content), I hide the content of the page until light.css has loaded."

> we'd be repeatedly injecting the same element on every config update

my bad, I only tested loading a new theme and in that case the page is reloaded.

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: No, but it's inconsistent with the rest.

gui/default/syncthing/core/syncthingController.js
mx-psi: > If it does not, cannot we just match on 'not all' in css? Or use a not(selector) ?

I tried not using JS and changing the light theme condition to `not all and (prefers-color-scheme: dark)`. It works correctly on Firefox 70 but it doesn't on Firefox 52 (a browser without support for `prefers-color-scheme`).

AudriusButkevicius: I would expect 'not all' to be a value matcher for the pref?

AudriusButkevicius: ok tried it myself, it seems you can't do it without JS.
 It seems that not all browsers support `mediaMatch`, so we should check if that exists on the window.

tomasz1986: > > If it does not, cannot we just match on 'not all' in css? Or use a not(selector) ?
> 
> I tried not using JS and changing the light theme condition to `not all and (prefers-color-scheme: dark)`. It works correctly on Firefox 70 but it doesn't on Firefox 52 (a browser without support for `prefers-color-scheme`).

`not all` means `not screen`, `not print`, and `not speech`, so it is basically the same as not having the media query at all. It has nothing to do with support for the specific feature, and of course will be completely ignored by any browser.

I am no Web development expert, but this honestly looks very confusing. Why not just apply the light theme always, unconditionally, just with CSS and drop the JavaScript altogether? This way you can be 100% sure that everything will work correctly in old browsers. The code will be much simpler too, with no additional JavaScript hackery.

AudriusButkevicius: Because someone will add some rule (all td's are white) to the light theme, which will break dark theme, because both files are loaded, and dark theme doesn't explicitly state that all td's are dark, and inherit the colour from the body.

tomasz1986: But is this not the case right now (without this commit, that is)?

The original default theme was light, and the other themes have been built on top of it. Is this wrong? Do you want to completely separate the light theme from the rest? The problem is that by doing so, there will be no fallback, which means good bye to progressive enhancement.

On a side note, if that is the intention, this change will likely break any users' custom themes which until now could safely rely on the fact that that the default theme was light.

AudriusButkevicius: That is not the case now, as all the default stuff comes from bootstrap itself.
The dark theme just overrides what bootstrap does, which we can assume is stable and we can cover the ground.

Our light theme on the other hand, is customisation to bootstrap, which for now does not have a dodgy td: white, but it might, at which point that will override the dark td inherited from body when setting body colour in bootstrap.

tomasz1986:
This has not been mentioned before, but this change also affect any users' custom themes, which until now could safely rely on the fact that the default theme was light. With this change, the light theme specific CSS will no longer be applied to them, which may cause the custom themes no to display correctly.

If the modification is inevitable, would it not be a good idea to at least update the documentation to explain the new behaviour?

gui/default/syncthing/core/syncthingController.js
tomasz1986: Oh, you are right. I wrongly assumed that the `theme.css` from the `default` theme was applied to all the other themes, while this does not seem to be the case. I am sorry for the confusion.

I still do not believe that using JavaScript to detect browser features in order to provide a "fallback" for CSS is the right way. If something goes wrong with the JavaScript hackery here (which is very likely e.g. in a quirky older browser), then neither the light nor the dark theme will be applied at all. This may not be a big deal at the moment, as the two themes have a very limited set of styles right now, but may become a problem if they are expanded in the future.

mx-psi: > `not all` means `not screen`, `not print`, and `not speech`, so it is basically the same as not having the media query at all. It has nothing to do with support for the specific feature, and of course will be completely ignored by any browser.

That media query can be confusing but it doesn't work that way. `not all and (prefers-color-scheme: dark)` is parsed as `not (all and (prefers-color-scheme: dark))`, that is, the negation of `prefers-color-scheme: dark`. 

`all` is included because one can't negate `prefers-color-scheme` directly; as [MDN says](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#Inverting_a_query's_meaning) : "The not keyword can't be used to negate an individual feature query, only an entire media query". See also [this SO answer](https://stackoverflow.com/a/36494725/3414720).

This query works on recent browsers (and it is the one used in the last revision of [Thunderbird's CSS](https://hg.mozilla.org/comm-central/file/tip/mail/themes/windows/mail/multimessageview.css#l23) to set a dark theme, see line 23), but it didn't work on Firefox 52 when I tried. You are welcome to suggest other possible media queries to see if we can come up with something compatible with older browsers, but sadly I don't think there is a pure CSS solution.

tomasz1986: I have updated myself on the media queries specification (https://www.w3.org/TR/css3-mediaqueries and https://www.w3.org/TR/mediaqueries-4), and yes, you are right about the `not` keyword, which is applied to the whole declaration following it. The whole `all` is not required though. By the official specification, you could even do a simple negation like this:

```
@import "/theme-assets/dark/assets/css/theme.css" screen and (prefers-color-scheme: dark);
@import "/theme-assets/light/assets/css/theme.css" not screen and (prefers-color-scheme: dark);
```

The problem is that this does not work in real life. While the syntax may work correctly in newer browsers, IE11 completely ignores it, despite the fact that theoretically the browser should support CSS3 media queries.

With this in mind, in my opinion, the only way to do the whole thing with CSS only is to apply one of the themes by default, and then override the values with the other one (i.e. use the CSS cascade). This would indeed require to maintain the two themes simultaneously. They could be even merged into a single file. If the Syncthing developers do not like this, then indeed a JavaScript band-aid may be the only possible method.

Pavel8419:
7
--
Отправлено из Mail.ru для Android пятница, 08 ноября 2019г., 16:49 +03:00 от Audrius Butkevicius  notifications@github.com :

AudriusButkevicius:
Looks good. I have a follow up PR fixing cache invalidation sucking badly.

AudriusButkevicius:
Thanks for the PR

mx-psi:
Great, thank you for your assistance and your patience!

tomasz1986:
I wrote about it above in a comment, but it was apparently ignored.

When using dark theme and printing the page, no theme is used for the printed page. Is this the desirable behaviour?

You need to have `print` included in the rule in order to use the light theme for it.

```
@import "/theme-assets/light/assets/css/theme.css" print, (prefers-color-scheme: light), (prefers-color-scheme: no-preference);
```

AudriusButkevicius:
I think so.

tomasz1986:
Why not use the light theme?

Now, when `prefers-color-scheme: light` or `prefers-color-scheme: no-preference` is detected by the browser, it is also applied to the printed page. With `prefers-color-scheme: dark` detected, no theme is applied to the printed page.

This means that the printed page will look different depending on the OS/browser color scheme, even though the theme selected in Syncthing is just "default" in both cases.

AudriusButkevicius:
Sure, file a PR fixing it, but I think this is so moot that nobody actually cares. The page is not made for printing, it doesn't make sense to do that.

tomasz1986:
Yes, that is true, and I am personally not interested in printing the page myself too, but you were the one to mention print in the first place, hence my question.

wweich:
This doesn't work with reverse proxy, as the URL `/theme-assets/dark/assets/css/theme.css` doesn't exist.
That's why all other (internal) links in Syncthing are relative, not root-relative.
As links in CSS are relative to the CSS file, it should probably be
```css
@import "../../theme-assets/dark/assets/css/theme.css" screen and (prefers-color-scheme: dark);
```
The link in the JS fallback will probably work, if you just remove the leading `/`.

PR #6125, opened 2019-11-04
Fix bufferpool puts (ref #4976)

calmh:
There was a logic error in Put() which made us put all large blocks into
segment zero, where we subsequently did not look for them.

I also added a lowest threshold, as we otherwise allocate a 128KiB
buffer when we need 24 bytes for a header and such.

After:

```
2019/11/04 18:21:49 Result: Wall time: 2.665497ms / MiB
2019/11/04 18:21:49 Result: 3.84e+05 KiB/s synced
2019/11/04 18:21:49 Receiver: Utime: 4.506592ms / MiB
2019/11/04 18:21:49 Receiver: Stime: 1.126973ms / MiB
2019/11/04 18:21:49 Receiver: MaxRSS: 321632 KiB
2019/11/04 18:21:49 Sender: Utime: 4.163059ms / MiB
2019/11/04 18:21:49 Sender: Stime: 963.312µs / MiB
2019/11/04 18:21:49 Sender: MaxRSS: 113696 KiB
```

Before:

```
2019/11/04 17:16:03 Result: Wall time: 2.77946ms / MiB
2019/11/04 17:16:03 Result: 3.68e+05 KiB/s synced
2019/11/04 17:16:03 Receiver: Utime: 4.580026ms / MiB
2019/11/04 17:16:03 Receiver: Stime: 1.283715ms / MiB
2019/11/04 17:16:03 Receiver: MaxRSS: 644452 KiB
2019/11/04 17:16:03 Sender: Utime: 4.62121ms / MiB
2019/11/04 17:16:03 Sender: Stime: 1.075979ms / MiB
2019/11/04 17:16:03 Sender: MaxRSS: 1790704 KiB
```

### DIFF
--- a/lib/protocol/bufferpool.go
+++ b/lib/protocol/bufferpool.go
@@ -30,7 +30,7 @@ func newBufferPool() bufferPool {
 func (p *bufferPool) Get(size int) []byte {
 	// Too big, isn't pooled
 	if size > MaxBlockSize {
-		atomic.AddInt64(&p.misses, 1)
+		atomic.AddInt64(&p.skips, 1)
 		return make([]byte, size)
 	}
 
--- a/lib/protocol/bufferpool.go
+++ b/lib/protocol/bufferpool.go
@@ -22,17 +22,10 @@ func (p *bufferPool) Get(size int) []byte {
 		return make([]byte, size)
 	}
 
-	// Find the segment containing a slice guaranteed to be >= size
-	var i, blockSize int
-	for i, blockSize = range BlockSizes {
-		if size <= blockSize {
-			break
-		}
-	}
-
-	var bs []byte
 	// Try the fitting and all bigger pools
-	for j := i; j < len(BlockSizes); j++ {
+	var bs []byte
+	bkt := getBucketForSize(size)
+	for j := bkt; j < len(BlockSizes); j++ {
 		if intf := p.pools[j].Get(); intf != nil {
 			bs = *intf.(*[]byte)
 			return bs[:size]
@@ -45,7 +38,7 @@ func (p *bufferPool) Get(size int) []byte {
 	if size < MinBlockSize/64 {
 		return make([]byte, size)
 	}
-	return make([]byte, BlockSizes[i])[:size]
+	return make([]byte, BlockSizes[bkt])[:size]
 }
 
 // Put makes the given byte slice available again in the global pool
@@ -56,12 +49,9 @@ func (p *bufferPool) Put(bs []byte) {
 	if c > MaxBlockSize*1.5 || c < MinBlockSize {
 		return
 	}
-	for i, blockSize := range BlockSizes {
-		if c <= blockSize {
-			p.pools[i].Put(&bs)
-			return
-		}
-	}
+
+	bkt := putBucketForSize(c)
+	p.pools[bkt].Put(&bs)
 }
 
 // Upgrade grows the buffer to the requested size, while attempting to reuse
@@ -76,4 +66,30 @@ func (p *bufferPool) Upgrade(bs []byte, size int) []byte {
 	// buffer.
 	p.Put(bs)
 	return p.Get(size)
+}
+
+// getBucketForSize returns the bucket where we should *get* a slice of
+// a certain size. Each bucket is guaranteed to hold slices that are *at
+// least* the block size for that bucket, so if the block size is larger
+// than our size we are good.
+func getBucketForSize(size int) int {
+	for i, blockSize := range BlockSizes {
+		if size <= blockSize {
+			return i
+		}
+	}
+	return len(BlockSizes) - 1
+}
+
+// putBucketForSize returns the bucket where we should *put* a slice of a
+// certain size. Each bucket is guaranteed to hold slices that are *at
+// least* the block size for that bucket, so we find the highest bucket that
+// is smaller or equal to our size.
+func putBucketForSize(size int) int {
+	for i := len(BlockSizes) - 1; i >= 0; i-- {
+		if size >= BlockSizes[i] {
+			return i
+		}
+	}
+	return -1
 }
--- a/lib/protocol/bufferpool.go
+++ b/lib/protocol/bufferpool.go
@@ -35,12 +35,11 @@ func (p *bufferPool) Get(size int) []byte {
 	}
 
 	// Try the fitting and all bigger pools
-	var bs []byte
 	bkt := getBucketForLen(size)
 	for j := bkt; j < len(BlockSizes); j++ {
 		if intf := p.pools[j].Get(); intf != nil {
-			bs = *intf.(*[]byte)
 			atomic.AddInt64(&p.hits[j], 1)
+			bs := *intf.(*[]byte)
 			return bs[:size]
 		}
 	}
--- a/lib/protocol/bufferpool.go
+++ b/lib/protocol/bufferpool.go
@@ -36,7 +36,7 @@ func (p *bufferPool) Get(size int) []byte {
 
 	// Try the fitting and all bigger pools
 	var bs []byte
-	bkt := getBucketForSize(size)
+	bkt := getBucketForLen(size)
 	for j := bkt; j < len(BlockSizes); j++ {
 		if intf := p.pools[j].Get(); intf != nil {
 			bs = *intf.(*[]byte)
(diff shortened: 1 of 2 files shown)


calmh:
This is still broken, I'mma add some tests

AudriusButkevicius:
Why are we not using it for small slices? There shouldn't be any downsides. 

I guess for things like headers we should just have a "static" slice, not to churn trash.

calmh:
The difference is only in when we don't have a block available in any bucket. Previously we'd allocate a 128 KiB block and return that. Now, this is a block we can maybe reuse and that's good. But we've also just allocated 128 KiB when we just needed 24 bytes or whatever which comes from a completely different allocator in Go... So... :shrug:

And yes, I guess we should. But this is about making the buffer pool actually do buffer pooly things and not just allocate bytes forever...

calmh:
```
2019/11/04 18:48:51 Verifying...
2019/11/04 18:48:57 Result: Wall time: 2.626817ms / MiB
2019/11/04 18:48:57 Result: 3.9e+05 KiB/s synced
2019/11/04 18:48:57 Receiver: Utime: 4.599636ms / MiB
2019/11/04 18:48:57 Receiver: Stime: 1.166088ms / MiB
2019/11/04 18:48:57 Receiver: MaxRSS: 324292 KiB
2019/11/04 18:48:57 Sender: Utime: 4.278935ms / MiB
2019/11/04 18:48:57 Sender: Stime: 978.709µs / MiB
2019/11/04 18:48:57 Sender: MaxRSS: 89528 KiB
```

lib/protocol/bufferpool_test.go
```diff
@@ -0,0 +1,41 @@
...
+		// Up to the next blocksize - 1 we still put in the same bucket, but
+		// we look for these blocks in the next bucket where we know they
+		// can be found.
+		{size: 2*MinBlockSize - 1, put: 0, get: 1},
+
+		// Next blocksize at the border...
+		{size: 2 * MinBlockSize, put: 1, get: 1},
+
+		// ... and past it.
+		{size: 2*MinBlockSize + 1, put: 1, get: 2},
```
AudriusButkevicius: I don't get this. We should always allocate bucket sized slices, and we should always get and return them to the same pool.

AudriusButkevicius:
Different allocator area? First time I've heard that there is more than a single area.

lib/protocol/bufferpool_test.go
calmh: Granted the cap should be an even block size when we put it, but the size we are asking for will be random and the two needs to be handled separately anyway. Plus in this way we survive if someone puts a random buffer. 🤷‍♂

calmh:
Small blocks come from arenas containing lots of equally sized blocks, allocated in chunks. Large blocks are rounded up to page size and handled individually. So the cost for allocating a (say) 24 byte block is often near zero as we already have a page full of such blocks and just need to pick one, while a large allocation is seldom free.

lib/protocol/bufferpool_test.go
AudriusButkevicius: We readjust the size on reading anyway, so we should put stuff based on cap, if it does match a bucket, panic.

calmh:
Made it a bit less chill and more panicky so we'll see if it's misused.

lib/protocol/bufferpool_test.go
```diff
@@ -0,0 +1,106 @@
...
+	const routines = 100
+	const runtime = time.Second
+
+	bp := newBufferPool()
+	t0 := time.Now()
+
+	var wg sync.WaitGroup
+	for i := 0; i < routines; i++ {
+		wg.Add(1)
+		go func() {
```
golangcibot: SA2002: the goroutine calls T.Fatal, which must be called in the same goroutine as the test (from `staticcheck`)

lib/protocol/bufferpool_test.go
```diff
@@ -0,0 +1,116 @@
...
```
golangcibot: SA2002: the goroutine calls T.Fatal, which must be called in the same goroutine as the test (from `staticcheck`)

calmh:
PTAL

lib/protocol/bufferpool_test.go
```diff
@@ -0,0 +1,125 @@
...
+	var wg sync.WaitGroup
+	fail := make(chan struct{}, routines)
+	for i := 0; i < routines; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for time.Since(t0) < runtime {
+				blocks := make([][]byte, 10)
+				for i := range blocks {
+					want := rand.Intn(1.5 * MaxBlockSize)
```
AudriusButkevicius: This 1.5 is mysterious

lib/protocol/bufferpool_test.go
```diff
@@ -0,0 +1,125 @@
...
+			if res != tc.bkt {
+				t.Errorf("block of size %d should put into bucket %d, not %d", tc.size, tc.bkt, res)
+			}
+		}
+	}
+}
+
+func TestStressBufferPool(t *testing.T) {
+	const routines = 10
+	const runtime = 2 * time.Second
```
AudriusButkevicius: T.short = t.skip()

lib/protocol/bufferpool.go
```diff
@@ -2,56 +2,72 @@
...
 	}
-	// All pools are empty, must allocate.
-	return make([]byte, BlockSizes[i])[:size]
+
+	atomic.AddInt64(&p.misses, 1)
+
+	// All pools are empty, must allocate. For very small slices where we
+	// didn't have a block to reuse, just allocate a small slice instead of
+	// a large one. We won't be able to reuse it, but avoid some overhead.
+	if size < MinBlockSize/64 {
```
imsodin: Is this the mentioned "arena-limit"? If so, a comment about that would make the number less magic.

lib/protocol/bufferpool.go
```diff
@@ -67,3 +83,30 @@ func (p *bufferPool) Upgrade(bs []byte, size int) []byte {
...
+		}
+	}
+
+	panic(fmt.Sprintf("bug: tried to get impossible block size %d", size))
+}
+
+// putBucketForSize returns the bucket where we should put a slice of a
+// certain size. Each bucket is guaranteed to hold slices that are precisely
+// the block size for that bucket, so we just find the matching one.
+func putBucketForSize(cap int) int {
```
imsodin: `putBucketForCap`?

lib/protocol/bufferpool.go
calmh: It isn't, it's just something I picked out of thin air. I don't know what the specific limit is and it's anyway subject to change. If this is a contentious point I can remove the optimization...

lib/protocol/bufferpool_test.go
```diff
@@ -0,0 +1,125 @@
...
+		{size: MinBlockSize, bkt: 0},
+		{size: MinBlockSize + 1, panics: true},
+		{size: 2 * MinBlockSize, bkt: 1},
+		{size: MaxBlockSize, bkt: len(BlockSizes) - 1},
+		{size: MaxBlockSize + 1, panics: true},
+	}
+
+	for _, tc := range cases {
+		if tc.panics {
+			shouldPanic(t, func() { putBucketForSize(tc.size) })
```
golangcibot: undeclared name: `putBucketForSize` (from `typecheck`)

lib/protocol/bufferpool_test.go
```diff
@@ -0,0 +1,125 @@
...
+		{size: 2*MinBlockSize - 1, bkt: 1},
+		{size: 2 * MinBlockSize, bkt: 1},
+		{size: 2*MinBlockSize + 1, bkt: 2},
+		{size: MaxBlockSize, bkt: len(BlockSizes) - 1},
+		{size: MaxBlockSize + 1, panics: true},
+	}
+
+	for _, tc := range cases {
+		if tc.panics {
+			shouldPanic(t, func() { getBucketForSize(tc.size) })
```
golangcibot: undeclared name: `getBucketForSize` (from `typecheck`)

lib/protocol/bufferpool.go
imsodin: Ok fine, my point was just about mentioning it in comments, in case it was based on some arena thingy. Whether or not to have the optimization I don't have anything to base an opinion on, so go with whatever your gut says ;)

AudriusButkevicius:
Baaah. squash message. We should have a hook that prevents quickfire merges with bad names.

PR #6134, opened 2019-11-07
lib/logger: Implement suppression (fixes #6132)

### LINKED ISSUE #6132: Dynamic log supression
Sometimes we log a shitload of data - when we're stuck retrying failing sync for lots of items, for example, or when there's a bug and we get stuck retrying something like socket accepts thousands of times a second. We do want to log most of this, but at a reasonable rate. Some requirements I have on the logging:

- Failed items must be logged. At least once, and with some indication that failures are ongoing. I don't need to see every time everything fails, but a grep for something that has failed must show *something*. 

- Repeated log entries are fine if the volume is low. For example, if I manually pause and resume a folder I want log output for that in real time, even if I do it twice in succession.

- We should not log repeated log entries infinitely.

I propose that we add some sort of suppression to the logging layer. Something like the following would work for me.

1. Assume we keep track on the number of log entries produced per minute, over at least the last three to five minutes.

2. Once every minute we rotate the buckets and make the decision on whether suppression is on or off based on comparing the rate to some threshold. If the rate is low enough to not cause problems (say, <100 messages/minute on average over the last five minutes) we don't do any kind of suppression, if it's higher we enable suppression. There is a log entry when we enable or disable suppression.

2. Assume also that we keep track of the hashes of log messages we've printed (sans timestamp etc), in the same one minute buckets.

3. When we're in suppression and a log entry matches an existing hash we don't log it and just bump a counter somewhere instead.

4. Once a minute we log the number of suppressed messages and clear the counter.

The result would be that under normal circumstances there isn't any suppression. Once we end up in a state with 100k failed items we'd log them once, trip the threshold, next time just say "100k previously seen log messages suppressed since $last_minute" and that'd be that.

The cost here would be keeping hashes of the last few minutes log entries. In normal operation that's roughly zero. When there's a shitload of entries we will incur som memory cost and may want a limit. If we limit ourselves to say 1M log messages tracked that might be maybe 32 megabytes of hashes and the same again in map overhead and whatnot. I think that's an OK price to pay. There could of course be an off switch.

As a special case we could always do a comparison to exactly the last message printed and not repeat that, instead saying "last message repeated 48736 times" once every minute or so. This just for when we get stuck in some buggy accept loop.

calmh:
With this change, certain identical log messages are filtered out,
reducing log volume when there are large numbers of repeats.

The mechanism is as described in #6132.

Oh and I fixed the historical licensing of the logging package.


lib/logger/suppression.go
```diff
@@ -0,0 +1,171 @@
...
+func (l *suppressingLogger) Output(level int, message string) error {
+	l.mut.Lock()
+	defer l.mut.Unlock()
+
+	hash := messageHash(message)
+	l.buckets[0].seen[hash] = struct{}{}
+
+	if l.suppressSameAsLast(hash) {
+		// Message is the same as the last message
+		return nil
```
imsodin: Does this address any concrete case? I couldn't think of any. And in general I'd rather keep repeat messages if suppression is not active/volume is low. I believe that we should not hide them, but correct them where they occur when someone notices (e.g. the repeat warnings in quic recently).

TLDR: I wouldn't suppress anything unless active/high volume.

lib/logger/logger.go
```diff
@@ -33,6 +36,12 @@ const (
 	DebugFlags   = log.Ltime | log.Ldate | log.Lmicroseconds | log.Lshortfile
 )
 
+const (
+	suppressionBuckets  = 5
+	supressionThreshold = 100
```
imsodin: double-p-typo

lib/logger/suppression.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	if l.active && l.haveSeen(hash) {
+		l.buckets[0].suppressed++
+		return nil
+	}
+
+	l.buckets[0].count++
+	return l.next.Output(level+1, message)
```
imsodin: Why +1?

calmh:
I removed the last message suppression as it's not strictly necessary.

Using utils.AsService would shave maybe a line of code, but isn't possible because it would cause an import loop (util uses sync, sync uses logger)...

I'm guessing both you misunderstood the `level` parameter to be about the log level, which it's not. In this case it's the number of levels to go up to find the source line of the log caller. I think that makes the +1 and the zero make sense, holler if you think it doesn't.

calmh:
Are you all certain we need an off switch for this? I mean, it's our decision as an application what to output as log messages, and we have several other places where we for example just output a warning once even though the thing might happen/fail repeatedly.

As for this being attacking it from the wrong end (@imsodin) -- what do you suggest? The main scenario when this would kick in, except for bugs, is when there's a large amount of repeated pull failures. Perhaps we should keep state for each such failure and just log it the first time?

imsodin:
> I'm guessing both you misunderstood the level parameter to be about the log level, [...]

Exactly, thanks for the explanation.

> Are you all certain we need an off switch for this?

Not certain no, quite the opposite: I believe this will make debugging easier by removing noise, but I am not certain that I won't ever be stock enough while debugging to suspect that there's some signal in that suppressed noise. So feel free to discount my view as the FUD it is xD

As to the wrong end: My main sentiment is that this is too much complexity and resources for a limited problem. Otherwise what you said: filter/optimize logging where we know there's lots of logging or fix the bug if logging gets unreasonably verbose. And for the latter case have an unsophisticated fuse that curbs logging when unreasonable (if the first 100MB of log don't contain enough info to track the bug, the next 10GB probably wont either).

For pulling I propose https://github.com/syncthing/syncthing/pull/6149

AudriusButkevicius:
I'd still like a off switch for this, even if its an env var.

I don't mind this change per-se, I think it will prevent a few ugly cases we had recently.

calmh:
I agree filtering log commands in the puller (and so on) is better, and we should do that as you've proposed. I still think this has a place as a "fuse", and given that I think we can't do it much less complex or it will either be useless or eat valuable log data...

imsodin:
What about just counting while not active, and starting to hash only once the activity threshold has been exceeded?

calmh:
Good suggestion!

calmh:
So, I think this works, but do we need it? We have fixed the puller output, which is the large logging culprit in the normal case. We have added rotation, which avoids murdering Windows folks' systems by default. We can still get stuck in logging loops due to bugs, but the rotation will act as a fuse for that and for people with systemd/syslog setups that's unfortunate but their log systems can do deduplication as required... I think we're good without this for now.

PR #6135, opened 2019-11-07
lib/model: Add folders on start in model

imsodin:
Nobody outside of model has any business meddling directly with folders and starting folders together with the model also seems more fitting (and it has an almost neutral/slightly negative line count :) ).

Testing: Folders still start on startup.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -72,9 +72,6 @@ type Model interface {
 
 	connections.Model
 
-	AddFolder(cfg config.FolderConfiguration)
-	RestartFolder(from, to config.FolderConfiguration)
-	StartFolder(folder string)
 	ResetFolder(folder string)
 	DelayScan(folder string, next time.Duration)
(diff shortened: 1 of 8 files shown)


lib/model/model.go
```diff
@@ -217,6 +214,19 @@ func NewModel(cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersio
 	return m
 }
 
+func (m *model) Serve() {
+	// Add and start folders
+	for _, folderCfg := range m.cfg.Folders() {
+		if folderCfg.Paused {
+			folderCfg.CreateRoot()
```
golangcibot: Error return value of `folderCfg.CreateRoot` is not checked (from `errcheck`)

AudriusButkevicius:
So I'd argue that folder adding and removal should all be dome external, to lift logic out of model to keep it slimmer.

PR #6141, opened 2019-11-08
lib/model: Simplify pull error/retry logic (fixes #6139)

### LINKED ISSUE #6139: Some sync errors get cleared despite them still being valid
When there were no changes recorded in a pull iteration, we reset all errors. However for some errors which are not recoverable through other sync actions (invalid filenames, file not available, maybe more) we do not increment the changes counter, to prevent another iteration. The errors however are still valid and should not be reset.

This was discovered in https://forum.syncthing.net/t/14-out-of-sync-items/14048

imsodin:
Change counter zero does not mean there weren't any errors -> resetting all errors is incorrect. So instead of the global error reset on "success" errors are now reset every iteration (this makes sure we catch transient errors, which we didn't before). I.e. no more error resetting for individual items (which is anyway pretty fragile as it's easy to forget it). Same for incrementing `changed`: It's now done globally in the beginning instead of in various places, and decremented in the one place we don't do anything ever and thus don't want to retry (windows invalid filenames).

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -176,8 +176,6 @@ func (f *sendReceiveFolder) pull() bool {
 
 	l.Debugf("%v pulling", f)
 
-	f.clearPullErrors()
-
 	scanChan := make(chan string)
 	go f.pullScannerRoutine(scanChan)
 
@@ -185,6 +183,8 @@ func (f *sendReceiveFolder) pull() bool {
 		close(scanChan)
 		f.setState(FolderIdle)
 	}()
+
+	changed := 0
 
 	for tries := 0; tries < maxPullerIterations; tries++ {
 		select {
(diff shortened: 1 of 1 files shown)


calmh:
LGTM apart from breaking tests, which could be the tests being wrong but requires investigation.

imsodin:
I believe it's just flakyness (Mac and win pass) - I'll have a closer look when back at a computer (Sunday or Monday).

calmh:
👍 

CopierFinder failure looks systematic even on reruns, and isn't one that is known to be flaky previously I think.

imsodin:
I wanted to have the fast failure from master and after merging that it now did pass. And I can't reproduce on my pc. I don't understand how those failures can happen: The two functions that failed (`TestWeakHash` and `TestCopierFinder`) both should receive something on the pull chan but time out waiting for that. The common pattern is

```
go f.copierRoutine(..., pullChan, ...)
f.handleFile
<-pullChan
```

That points at a race between `copierRoutine` "starting up" and `handleFile` passing the file to it. However `handleFile` just sends it on the copier chan waiting for it to be read, it doesn't care if copierRoutine isn't ready yet.

PR #6144, opened 2019-11-10
Cleanup of things we can improve when requiring Go 1.12

calmh:
Home dir, exit code. Also dropping the TLS 1.3 thing since that's enabled by default in Go 1.13 that we normally build with.

### DIFF
--- a/cmd/syncthing/monitor.go
+++ b/cmd/syncthing/monitor.go
@@ -151,17 +151,14 @@ func monitorMain(runtimeOptions RuntimeOptions) {
 				// Successful exit indicates an intentional shutdown
 				return
 			} else if exiterr, ok := err.(*exec.ExitError); ok {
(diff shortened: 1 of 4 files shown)


cmd/syncthing/monitor.go
```diff
@@ -151,17 +151,15 @@ func monitorMain(runtimeOptions RuntimeOptions) {
...
-					switch status.ExitStatus() {
-					case syncthing.ExitUpgrade.AsInt():
-						// Restart the monitor process to release the .old
-						// binary as part of the upgrade process.
-						l.Infoln("Restarting monitor...")
-						if err = restartMonitor(args); err != nil {
-							l.Warnln("Restart:", err)
-						}
-						return
+				switch exiterr.ExitCode() {
```
golangcibot: singleCaseSwitch: should rewrite switch statement to if statement (from `gocritic`)

PR #6149, opened 2019-11-11
lib/model: Don't info log repeat pull errors

imsodin:
This reduces the verbosity of the puller via the following:

 - Don't drop the connection before stopping the folder. Instead make sure that an active connection doesn't prevent stopping the folder -> no connection lost errors.

 - Don't log errors for items due to the folder being stopped: That's not a file level problem (and is inconsistent, we also don't log an error for all remaining files to be synced that aren't due to folder stopping).

 - Only print errors if they are new/changed compared to the last pull iteration. For that instead of just removing errors before a pull iteration, track new errors in a new map and reset the old errors after the pull iteration. To offset for the lack of duplicate error reporting, I added a log line after every pull about how many files couldn't be synced/failed.

### DIFF
--- a/lib/model/fakeconns_test.go
+++ b/lib/model/fakeconns_test.go
@@ -33,7 +33,7 @@ type fakeConnection struct {
 	folder                   string
 	model                    *model
 	indexFn                  func(string, []protocol.FileInfo)
-	requestFn                func(folder, name string, offset int64, size int, hash []byte, fromTemporary bool) ([]byte, error)
+	requestFn                func(folder, name string, offset int64, size int, hash []byte, fromTemporary bool, ctx context.Context) ([]byte, error)
 	closeFn                  func(error)
 	mut                      sync.Mutex
 }
@@ -82,11 +82,11 @@ func (f *fakeConnection) IndexUpdate(folder string, fs []protocol.FileInfo) erro
 	return nil
 }
 
-func (f *fakeConnection) Request(folder, name string, offset int64, size int, hash []byte, weakHash uint32, fromTemporary bool) ([]byte, error) {
+func (f *fakeConnection) Request(folder, name string, offset int64, size int, hash []byte, weakHash uint32, fromTemporary bool, ctx context.Context) ([]byte, error) {
 	f.mut.Lock()
 	defer f.mut.Unlock()
 	if f.requestFn != nil {
-		return f.requestFn(folder, name, offset, size, hash, fromTemporary)
+		return f.requestFn(folder, name, offset, size, hash, fromTemporary, ctx)
 	}
 	return f.fileData[name], nil
 }
--- a/lib/model/fakeconns_test.go
+++ b/lib/model/fakeconns_test.go
@@ -82,7 +82,7 @@ func (f *fakeConnection) IndexUpdate(folder string, fs []protocol.FileInfo) erro
 	return nil
 }
 
-func (f *fakeConnection) Request(folder, name string, offset int64, size int, hash []byte, weakHash uint32, fromTemporary bool, ctx context.Context) ([]byte, error) {
+func (f *fakeConnection) Request(ctx context.Context, folder, name string, offset int64, size int, hash []byte, weakHash uint32, fromTemporary bool) ([]byte, error) {
 	f.mut.Lock()
 	defer f.mut.Unlock()
 	if f.requestFn != nil {
--- a/lib/model/fakeconns_test.go
+++ b/lib/model/fakeconns_test.go
@@ -33,7 +33,7 @@ type fakeConnection struct {
 	folder                   string
 	model                    *model
 	indexFn                  func(string, []protocol.FileInfo)
-	requestFn                func(folder, name string, offset int64, size int, hash []byte, fromTemporary bool, ctx context.Context) ([]byte, error)
+	requestFn                func(ctx context.Context, folder, name string, offset int64, size int, hash []byte, fromTemporary bool) ([]byte, error)
 	closeFn                  func(error)
 	mut                      sync.Mutex
 }
@@ -86,7 +86,7 @@ func (f *fakeConnection) Request(ctx context.Context, folder, name string, offse
 	f.mut.Lock()
 	defer f.mut.Unlock()
 	if f.requestFn != nil {
-		return f.requestFn(folder, name, offset, size, hash, fromTemporary, ctx)
+		return f.requestFn(ctx, folder, name, offset, size, hash, fromTemporary)
 	}
 	return f.fileData[name], nil
 }
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -170,7 +170,7 @@ func (f *sendReceiveFolder) pull() bool {
 		}
 	}()
 	if err := f.ignores.Load(".stignore"); err != nil && !fs.IsNotExist(err) {
-		err = fmt.Errorf("loading ignores: %v", err)
+		err = errors.Wrap(err, "loading ignores")
 		f.setError(err)
 		return false
 	}
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -104,7 +104,8 @@ type sendReceiveFolder struct {
 
 	queue *jobQueue
 
-	pullErrors    map[string]string // path -> error string
+	pullErrors    map[string]string // errors for most recent/current iteration
+	oldPullErrors map[string]string // errors from previous iterations for log filtering only
 	pullErrorsMut sync.Mutex
 }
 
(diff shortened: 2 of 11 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -104,7 +104,8 @@ type sendReceiveFolder struct {
 
 	queue *jobQueue
 
-	pullErrors    map[string]string // path -> error string
+	pullErrors    map[string]string // errors for most recent/current iteration
+	oldPullErrors map[string]string // errors from previous iterations for log filtering only
```
AudriusButkevicius: Is it useful to compare the actual message? If it failed before, it failed before...

Also wonder if it was not useful to invert the storage, store error -> files, or store error pointers not to bloat with repeated strings. Hopefully each pull batch is small enough, but the fact everything is stored in memory and can all fail on dc is a bit grim.

lib/model/folder_sendrecv.go
```diff
@@ -1458,7 +1466,17 @@ func (f *sendReceiveFolder) pullBlock(state pullBlockState, out chan<- *sharedPu
...
+		done := make(chan struct{})
+		go func() {
+			buf, lastError = f.model.requestGlobal(selected.ID, f.folderID, state.file.Name, state.block.Offset, int(state.block.Size), state.block.Hash, state.block.WeakHash, selected.FromTemporary)
+			close(done)
+		}()
+		select {
+		case <-done:
+		case <-f.ctx.Done():
+			state.fail(errors.Wrap(f.ctx.Err(), "folder stopped"))
+			return
```
AudriusButkevicius: This actually changes behaviour by exiting the loop early, not passing stuff to finisher to unregister for progress emission etc.

Also, what does this change have to do with the spirit of this PR?

lib/model/folder_sendrecv.go
```diff
@@ -1458,7 +1466,17 @@ func (f *sendReceiveFolder) pullBlock(state pullBlockState, out chan<- *sharedPu
 		// leastBusy can select another device when someone else asks.
 		activity.using(selected)
 		var buf []byte
-		buf, lastError = f.model.requestGlobal(selected.ID, f.folderID, state.file.Name, state.block.Offset, int(state.block.Size), state.block.Hash, state.block.WeakHash, selected.FromTemporary)
+		done := make(chan struct{})
+		go func() {
+			buf, lastError = f.model.requestGlobal(selected.ID, f.folderID, state.file.Name, state.block.Offset, int(state.block.Size), state.block.Hash, state.block.WeakHash, selected.FromTemporary)
```
AudriusButkevicius: Push the context down?

You could even push it down part of asyncMessage which does not dispatch it if the context is closed, and the response channel.

lib/model/model.go
```diff
@@ -427,18 +427,13 @@ func (m *model) tearDownFolderLocked(cfg config.FolderConfiguration, err error)
...
-	w := m.closeConns(cfg.DeviceIDs(), err)
-
 	for _, id := range tokens {
 		m.RemoveAndWait(id, 0)
 	}
 
 	// Wait for connections to stop to ensure that no more calls to methods
 	// expecting this folder to exist happen (e.g. .IndexUpdate).
-	w.Wait()
+	m.closeConns(cfg.DeviceIDs(), err).Wait()
```
AudriusButkevicius: Will this not result in a bunch of panics and errors for non-existing folders? We have a bunch of panics in sentry for Index messages for folders that do not exist. If anything, this will make it worse.

lib/model/folder_sendrecv.go
imsodin: I think so. If at some earlier point the error was a "no available device", then it becomes available and there's a permission error, the user should know about the newer one.

We need the path as key to check for existing errors. I could change it to store a error instead of the string, maybe that stops a string allocation. However I don't think it will impact memory as the original error goes out of scope, right_

lib/model/model.go
imsodin: Preventing that is the whole point of waiting here before removing the map items protected by `fmut` below. Folders (as in the services) have nothing to do with these panics, just what the model stores about them. Changing the order of stopping folder service and closing conns is unrelated to these panics.

lib/model/folder_sendrecv.go
AudriusButkevicius: If its referenced by the map the fact it goes out of scope does not matter, if its deviced once on the module level.

lib/model/model.go
AudriusButkevicius: So why was it done this way before?

lib/model/folder_sendrecv.go
calmh: If you want to be specific about the error I suggest storing a weak hash (like the 64 bit fnv I used in the log suppression idea) of the error instead of the error itself. Should save quite a lot compared so storing lots of potentially long strings.

lib/model/model.go
imsodin: As there was nothing to terminate the request, waiting for a folder to stop could take a really long time (e.g. timed out connection). By closing first ourselves it was made sure that request would fail. However that's imo a hack (mine btw), when the correct thing is to abort such potentially long-time operation based on the context.

lib/model/folder_sendrecv.go
imsodin: We anyway need to store one copy of the string for the API/UI. To not store duplicate error strings during a pull iteration, I'll immediately drop the error in the old map when encountering it again. So no additional memory usage.

lib/protocol/protocol.go
```diff
@@ -274,13 +276,13 @@ func (c *rawConnection) IndexUpdate(folder string, idx []FileInfo) error {
...
 		Files:  idx,
-	}, nil)
+	}, nil, context.TODO())
 	c.idxMut.Unlock()
 	return nil
 }
 
 // Request returns the bytes for the specified block after fetching them from the connected peer.
-func (c *rawConnection) Request(folder string, name string, offset int64, size int, hash []byte, weakHash uint32, fromTemporary bool) ([]byte, error) {
+func (c *rawConnection) Request(folder string, name string, offset int64, size int, hash []byte, weakHash uint32, fromTemporary bool, ctx context.Context) ([]byte, error) {
```
golangcibot: context.Context should be the first parameter of a function (from `golint`)

lib/protocol/protocol.go
```diff
@@ -639,12 +641,13 @@ func (c *rawConnection) handleResponse(resp Response) {
 	c.awaitingMut.Unlock()
 }
 
-func (c *rawConnection) send(msg message, done chan struct{}) bool {
+func (c *rawConnection) send(msg message, done chan struct{}, ctx context.Context) bool {
```
golangcibot: context.Context should be the first parameter of a function (from `golint`)

lib/protocol/wireformat.go
```diff
@@ -34,7 +35,7 @@ func (c wireFormatConnection) IndexUpdate(folder string, fs []FileInfo) error {
 	return c.Connection.IndexUpdate(folder, myFs)
 }
 
-func (c wireFormatConnection) Request(folder, name string, offset int64, size int, hash []byte, weakHash uint32, fromTemporary bool) ([]byte, error) {
+func (c wireFormatConnection) Request(folder string, name string, offset int64, size int, hash []byte, weakHash uint32, fromTemporary bool, ctx context.Context) ([]byte, error) {
```
golangcibot: context.Context should be the first parameter of a function (from `golint`)

lib/model/folder_sendrecv.go
calmh: 👌

lib/model/fakeconns_test.go
```diff
@@ -33,7 +33,7 @@ type fakeConnection struct {
 	folder                   string
 	model                    *model
 	indexFn                  func(string, []protocol.FileInfo)
-	requestFn                func(folder, name string, offset int64, size int, hash []byte, fromTemporary bool) ([]byte, error)
+	requestFn                func(folder, name string, offset int64, size int, hash []byte, fromTemporary bool, ctx context.Context) ([]byte, error)
```
AudriusButkevicius: convention is to have this as the first argument.

AudriusButkevicius: also, this could reuse protocol.BufferPool 🙃

lib/protocol/protocol.go
```diff
@@ -639,12 +641,13 @@ func (c *rawConnection) handleResponse(resp Response) {
...
 	select {
-	case c.outbox <- asyncMessage{msg, done}:
+	case c.outbox <- asyncMessage{msg, done, ctx}:
```
AudriusButkevicius: I don't see any change in outbox consumer?

lib/model/fakeconns_test.go
imsodin: Don't get it.

Also should have addressed all the comments.

lib/model/folder_sendrecv.go
```diff
@@ -1757,6 +1765,11 @@ func (f *sendReceiveFolder) moveForConflict(name, lastModBy string, scanChan cha
 }
 
 func (f *sendReceiveFolder) newPullError(path string, err error) {
+	if errors.Cause(err) == f.ctx.Err() {
```
AudriusButkevicius: I wonder if we don't do something like `fmt.Errorf("foo: %q", otherErr)`, which would kill Cause from working.
Did you check all paths for things that check the context?

imsodin: Good point, we did - now it's errors.Wrap everywhere.

imsodin:
An actual approval review is still needed to merge :)

imsodin:
Meh, conflict... sorry, need yet another approval

PR #6151, opened 2019-11-12
lib/ignore: Keep skipping ignored dirs for rooted patterns

imsodin:
### Purpose

This came up a few times, most recently here: https://forum.syncthing.net/t/how-to-correctly-exclude-all-files-and-folders-except-folder-abc-and-def/14062  
For most include patterns we need to traverse everything to catch them, but for rooted patterns without wildcards in all but the last part of the path we don't actually.

### Testing

To test this I needed to refactor to have a function that creates patterns from lines/strings instead of files/readers.


lib/ignore/ignore.go
```diff
@@ -358,87 +389,90 @@ func loadParseIncludeFile(filesystem fs.Filesystem, file string, cd ChangeDetect
...
-				line = line[4:]
-			} else if strings.HasPrefix(line, "(?d)") && !seenPrefix[2] {
-				seenPrefix[2] = true
-				pattern.result |= resultDeletable
-				line = line[4:]
-			} else {
-				break
-			}
+	for {
+		if strings.HasPrefix(line, "!") && !seenPrefix[0] {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

lib/ignore/ignore.go
```diff
@@ -30,6 +30,14 @@ const (
 	resultFoldCase          = 1 << iota
 )
 
+var defaultResult Result = resultInclude
+
+func init() {
+	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
```
golangcibot: string `darwin` has 3 occurrences, make it a constant (from `goconst`)

lib/ignore/ignore.go
```diff
@@ -30,6 +30,14 @@ const (
...
+		defaultResult |= resultFoldCase
+	}
```
AudriusButkevicius: Should have a comment why this is done

lib/ignore/ignore_test.go
```diff
@@ -1098,3 +1098,37 @@ func TestPartialIncludeLine(t *testing.T) {
...
+		{`!/t*t`, true},
+		{`!/parent/test`, true},
+		{`!/parent/t[eih]t`, true},
+		{`!/parent/t*t`, true},
+		{`!/pa\[esd\]rent/t*t`, true},
+		{`!/pa*nt/test`, false},
+		{`!/pa[sdf]nt/t[eih]t`, false},
+		{`!/lowest/pa[sdf]nt/test`, false},
+		{`!/lo*st/parent/test`, false},
+		{`/pa*nt/test`, true},
```
AudriusButkevicius: I am not sure why some cases have the outcome that they have.

I'd also like to see some tests that cover multiple patterns proving the scans don't dvelve down into ignored dirs.

lib/ignore/ignore.go
AudriusButkevicius: I guess this is just shifting code from a function.

lib/ignore/ignore_test.go
AudriusButkevicius: After digging into the code I think I now understand why it is the way it is, but I'd still like a wider test, for example `!/**.mp3`, last component would not be checked, yet double star allows you to punch through all levels.

imsodin: Thanks so much. I missed both that case and another complication (I knew there were nasty pitfalls here...):  
When a user adds the pattern `!/foo*` we internally add `!/foo*` and `!/foo*/**`. Under my current rules this results in not skipping ignored dirs, which is unnecessary. However if a user enters `!/foo*/` or `!/foo*/**`, ignored dirs should indeed not be skipped. However the two cases are indistinguishable when looking at a single `Pattern`.

So either this gets more complex. Or we just optimize for rooted paths without any wildcards anywhere.

imsodin: I found a way to do it with little complexity. Maybe a bit hacky, but what isn't when it comes to ignores. I also added more tests, include one in lib/scanner to check that we really don't delve.

lib/ignore/ignore.go
```diff
@@ -50,6 +58,34 @@ func (p Pattern) String() string {
...
 }
 
+func (p Pattern) allowsSkippingIgnoredDirs() bool {
+	if p.result.IsIgnored() {
+		return true
+	}
+	if p.pattern[0] != '/' {
+		return false
+	}
+	// Double asteriks everywhere in the path except at the end is bad
```
golangcibot: `asteriks` is a misspelling of `asterisk` (from `misspell`)
```suggestion

```

lib/ignore/ignore_test.go
```diff
@@ -1098,3 +1098,55 @@ func TestPartialIncludeLine(t *testing.T) {
...
+	!/bar
+	*
+	`
+	if err := pats.Parse(bytes.NewBufferString(stignore), ".stignore"); err != nil {
+		t.Fatal(err)
+	}
+	if !pats.SkipIgnoredDirs() {
+		t.Error("SkipIgnoredDirs should be true")
+	}
+
```
golangcibot: unnecessary trailing newline (from `whitespace`)

lib/ignore/ignore.go
```diff
@@ -203,11 +208,17 @@ func (m *Matcher) parseLocked(r io.Reader, file string) error {
 	}
 
 	m.skipIgnoredDirs = true
+	var previous string
 	for _, p := range patterns {
+		// We automatically add patterns with a /** suffix - no need to check them
+		if l := len(p.pattern); l > 3 && p.pattern[:len(p.pattern)-3] == previous {
+			continue
+		}
```
AudriusButkevicius: This optimization adds confusion and has questionable performance gains.

imsodin: The comment is confusing: This is not an optimisation, this is needed for the case outlined in https://github.com/syncthing/syncthing/pull/6151#discussion_r346215210:

A user entering a path like `!/foo/b*/**` means we should descend into ignored dirs (to catch e.g. `foo/bar/baz`). Contrary a user entering `!/foo/b*` doesn't require us to descend, however we internally add an additional pattern `!/foo/b*/**` to catch everything below. And here because of the previous pattern there's no need to descend into ignored dirs.

lib/ignore/ignore.go
```diff
@@ -50,6 +58,34 @@ func (p Pattern) String() string {
...
+		return false
+	}
+	// Any wildcards (*, [...]) anywhere except for the last path component are bad
+	comps := strings.Split(tocheck, "/")
+	for _, comp := range comps[:len(comps)-1] {
+		if strings.Contains(comp, "*") {
+			return false
+		}
+		// need to check for escapes
+		for i := strings.Index(comp, "["); i != -1; i = strings.Index(comp[i+1:], "[") {
```
calmh: There are also `?` wildcards and `{bracketed,alternatives}`, should we check for those?

Also escapes are not processed on Windows, so `\[foo]` there is a rooted path with a wildcard expression in it...

imsodin: I really shouldn't handle this globbing stuff manually...  
I switched to using `glob.QuoteMeta` to check for any "special stuff". That has the limitation that it doesn't work with escaped brackets `\[`. So this optimization will not apply to any include patterns with such in them.

AudriusButkevicius:
I'd like to have a better look at this in front of a bigger screen

lib/ignore/ignore.go
```diff
@@ -52,6 +60,27 @@ func (p Pattern) String() string {
...
 
+func (p Pattern) allowsSkippingIgnoredDirs() bool {
+	if p.result.IsIgnored() {
+		return true
+	}
+	if p.pattern[0] != '/' {
+		return false
+	}
+	// Double asterisk everywhere in the path except at the end is bad
+	tocheck := strings.TrimSuffix(p.pattern[1:], "**")
```
AudriusButkevicius: why do we [1:] it here?

lib/ignore/ignore.go
```diff
@@ -52,6 +60,27 @@ func (p Pattern) String() string {
...
+	if strings.Contains(tocheck, "**") {
+		return false
+	}
+	// Any wildcards (*, [...]) anywhere except for the last path component are bad
```
AudriusButkevicius: I think `**` is also a valid wildcard in the library.

I think all you need to do is:
1. Get the string without the last component
2. Check that quoted string matches original string

lib/ignore/ignore.go
imsodin: No good reason anymore - remnant from iterating over all path components.

lib/ignore/ignore.go
imsodin: The comment shouldn't list wildcards anymore.  
`**` is different from other wildcards, in that it also can't appear in the last component, except as a suffix (e.g. `!/foo/b**r` needs to descend to catch `foo/boo/far`), while all other wildcards are ok in the last component.

PR #6153, opened 2019-11-13
Decrease padding on the panel and modal bodies.

tgxn:
This allows better viewing when on a condensed screen, and reduces screen real estate slightly.

### Purpose

On smaller screens the padding of 15px is too much in these panels and modals. I have reduced them accordingly.

### Testing

Built using docker.

### Screenshots

New Panels:
image

Old Panels:
image

New Modal:
image

Old Modal:
image

### Documentation

I added a comment on the CSS I added, but none really required.

## Authorship

me, domenic@tgxn.net

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -247,6 +247,18 @@ a.toggler:hover {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


calmh:
Thanks!

PR #6154, opened 2019-11-13
lib/model: Remove unneccessary panic handling indexes

imsodin:
Initially this was just the first comment, which drops index updates if the folder isn't present in model instead of panicking (as that can happen when config changes happen). However that lead to a test panic. Turns out we do weird stuff in tests, like adding a folder but not starting it - we never do that in production. In production though we sometimes call the `...Locked` methods holding locks over both adding and starting, sometimes the "wrapper-functions" (`addFolder`). That really just introduces needless potential for races. So I removed all of them in favor of a single `addFolder` method, which takes the lock once and also doesn't need to release it in between (by dropping connections in the beginning).

To accommodate for tests that need a special ignores (change detection), I made ignores a parameter to `addFolderWithIgnores`, which isn't used in production (`addFolder` just calls `addFolderWithIgnores`  with the "normal" ignores).

To review look at the first and last commit - the middle one is just copy-pasting to make the diff of the third look nicer.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -224,7 +224,7 @@ func (m *model) Serve() {
 			folderCfg.CreateRoot()
 			continue
 		}
-		m.addFolder(folderCfg)
+		m.newFolder(folderCfg)
 	}
 	m.Supervisor.Serve()
 }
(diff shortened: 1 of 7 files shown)


lib/model/model.go
```diff
@@ -1016,23 +1016,18 @@ func (m *model) handleIndex(deviceID protocol.DeviceID, folder string, fs []prot
...
 	m.fmut.RUnlock()
 
-	if !existing {
-		l.Warnf("%v for nonexistent folder %q", op, folder)
-		panic("handling index for nonexistent folder")
+	if !ok {
+		// This can happen due to a config change that isn't yet
+		// reflected in the model
+		l.Debugf("%v for folder %q missing in model", op, folder)
+		return
```
calmh: Here's a question though... What are those circumstances exactly? Can it be that we're not up and running yet so we miss an index update and then process later ones, thus permanently lack index info? Or is this just when we're tearing things down, or what?

imsodin:
Actually the windows failure is legitimate: Some folder tests apparently actually require folders to be added but not started. Looks like that's done to test operations that would usually be done by the folder service by doing those operations manually. So seems like a legitimate issue and I'll reintroduce the add/start separation, but keep the locking changes.

lib/model/model.go
imsodin: Race on folder restart (e.g. unpause): After removing the folder from model and closing the connection, you need to acquire fmut to readd the folder in which time the connection can already be established and index messages incoming. However looking at this again I think we should be able to do all of it (removing, adding and starting) while continuously holding fmut - thus preventing this race.

The race on adding a folder however I believe we can't prevent (easily): Folder is added in cfg, some unrelated change closes the connection, which then gets readded before the folder is added in the model. The only way I see to prevent this is to newer use the wrapper directly in the model (specifically in `generateClusterConfig`), but copy everything to model during `CommitConfiguration`. Thus we'd always have a consistent state within model. Here it means even if the connection comes in before we add the folder, we don't send the info about the new folder in the CC, thus won't receive index updates preemptively.

lib/model/model.go
```diff
@@ -468,16 +464,41 @@ func (m *model) restartFolder(from, to config.FolderConfiguration) {
...
 	}
 	l.Infof("%v folder %v (%v)", infoMsg, to.Description(), to.Type)
 }
 
+// Adds and starts a newly added folder (for unpaused/changed folders, user restartFolder)
+func (m *model) newFolder(cfg config.FolderConfiguration) {
+	l.Infoln("Adding new folder", cfg.Description())
+
+	// Close connections to affected devices
+	m.closeConns(cfg.DeviceIDs(), fmt.Errorf("started folder %v", cfg.Description()))
```
AudriusButkevicius: Should we do the closing afterwards, given metadata calculation can take some time, and we might reconnect and start getting index updates while that calculation is happening?

imsodin:
I'll close this for now: For two reasons: Adding/Starting folders is a mess in tests and this PR makes it worse. And I'd like to actually fix the racyness of config changes/connection handling/folder adding. I'll schedule a task for myself to drop this panic before the next RC if I don't follow through with this.

PR #6155, opened 2019-11-14
lib/model: Prevent cleanup-race in testing (ref #6152)

imsodin:
This is a followup to #6152: While preventing leaked goroutines is good, just closing all the channels results in races. So I added a wait group where necessary. And I found another leak in the progress emitter tests.


imsodin:
Ok, finally seems to be flake- and raceless.

PR #6156, opened 2019-11-14
lib/model: Keep fmut locked while adding/starting/restarting folders

imsodin:
This is basically #6154 but as unobtrusive as possible: Don't remove panics/functions and don't touch tests. I just refactors to make locking more consistent:

 - Don't have `...Locked` functions that drop the lock halfway through -> move closing connections out of `startFolderLocked` and split `tearDownFolderLocked` into `stopFolder` and `removeFolderLocked`.

 - Keep the lock continuously when restarting an existing or adding a new folder (add `newFolder` for the latter).

What I want to achieve is reduce complexity/margin for mistakes around locks and prevent that some other operation acquiring fmut comes in between adding-starting or removing-readding folders.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -224,8 +224,7 @@ func (m *model) Serve() {
 			folderCfg.CreateRoot()
 			continue
 		}
-		m.addFolder(folderCfg)
-		m.startFolder(folderCfg.ID)
+		m.newFolder(folderCfg)
 	}
 	m.Supervisor.Serve()
 }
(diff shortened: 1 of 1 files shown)


lib/model/model.go
```diff
@@ -483,22 +488,40 @@ func (m *model) restartFolder(from, to config.FolderConfiguration) {
...
 	defer m.fmut.Unlock()
 
-	m.tearDownFolderLocked(from, fmt.Errorf("%v folder %v", errMsg, to.Description()))
+	m.removeFolderLocked(from)
 	if !to.Paused {
-		// Creating the fileset can take a long time (metadata calculation)
-		// so we do it outside of the lock.
-		m.fmut.Unlock()
-		fset := db.NewFileSet(to.ID, to.Filesystem(), m.db)
-		m.fmut.Lock()
```
AudriusButkevicius: Could we use newFolder here? Also logic wise, if both from and to are paused, we'd drop connections for no good reason inside of stopFolder.

lib/model/model.go
```diff
@@ -288,11 +291,6 @@ func (m *model) startFolderLocked(cfg config.FolderConfiguration) {
 		}
 	}
 
-	// Close connections to affected devices
-	m.fmut.Unlock()
-	m.closeConns(cfg.DeviceIDs(), fmt.Errorf("started folder %v", cfg.Description()))
-	m.fmut.Lock()
-
 	v, ok := fset.Sequence(protocol.LocalDeviceID), true
```
AudriusButkevicius: the ok is always true (not that it's related, but caught me off guard)

calmh: This is bizarre, must be some sort of temporary refactoring damage left in by mistake

PR #6163, opened 2019-11-18
Fix misspellings

lu4p:
### Purpose
Fix misspellings in the .go files by running misspell -w


calmh:
Thanks, but no thanks. Spelling only corrections to comments are in my opinion more churn than value.

PR #6166, opened 2019-11-19
lib: Replace done channel with contexts in and add names to util services

imsodin:
### Purpose

This is the straighforward, tedious change to make the util services take an identifying string as parameter and use contexts instead of a close channel (see https://github.com/syncthing/syncthing/pull/5973#discussion_r316135960 and https://github.com/syncthing/syncthing/pull/5863#discussion_r304234063). This hasn't been utilized yet except for `model.folder`, where the context from the service is used instead of the existing, newly created context, and quic_listen, where the done channel was "converted" to a context. Also I needed to add a few `.String` methods.

More PRs making use of the contexts to follow.

### Testing

Still pass.

### DIFF
--- a/lib/beacon/beacon.go
+++ b/lib/beacon/beacon.go
@@ -80,7 +80,7 @@ func (c *cast) createService(svc func(context.Context) error, suffix string) uti
 		err := svc(ctx)
 		l.Debugf("Stopped %v %v: %v", c.name, suffix, err)
 		return err
-	}, c.String())
+	}, fmt.Sprintf("%s/%s", c, suffix))
 }
 
 func (c *cast) Stop() {
--- a/lib/beacon/beacon.go
+++ b/lib/beacon/beacon.go
@@ -7,6 +7,7 @@
 package beacon
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"time"
@@ -63,23 +64,23 @@ func newCast(name string) *cast {
 	}
 }
 
-func (c *cast) addReader(svc func(chan struct{}) error) {
+func (c *cast) addReader(svc func(context.Context) error) {
 	c.reader = c.createService(svc, "reader")
 	c.Add(c.reader)
 }
 
-func (c *cast) addWriter(svc func(stop chan struct{}) error) {
+func (c *cast) addWriter(svc func(ctx context.Context) error) {
 	c.writer = c.createService(svc, "writer")
 	c.Add(c.writer)
 }
 
-func (c *cast) createService(svc func(chan struct{}) error, suffix string) util.ServiceWithError {
-	return util.AsServiceWithError(func(stop chan struct{}) error {
+func (c *cast) createService(svc func(context.Context) error, suffix string) util.ServiceWithError {
+	return util.AsServiceWithError(func(ctx context.Context) error {
 		l.Debugln("Starting", c.name, suffix)
-		err := svc(stop)
+		err := svc(ctx)
 		l.Debugf("Stopped %v %v: %v", c.name, suffix, err)
 		return err
-	})
+	}, c.String())
 }
 
 func (c *cast) Stop() {
--- a/lib/beacon/broadcast.go
+++ b/lib/beacon/broadcast.go
@@ -29,13 +29,10 @@ func writeBroadcasts(ctx context.Context, inbox <-chan []byte, port int) error {
 		l.Debugln(err)
 		return err
 	}
-	done := make(chan struct{})
-	defer close(done)
+	doneCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	go func() {
-		select {
-		case <-ctx.Done():
-		case <-done:
-		}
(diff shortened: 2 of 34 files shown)


lib/relay/client/static.go
```diff
@@ -241,7 +242,7 @@ func performHandshakeAndValidation(conn *tls.Conn, uri *url.URL) error {
 	return nil
 }
 
-func messageReader(conn net.Conn, messages chan<- interface{}, errors chan<- error, stop chan struct{}) {
+func messageReader(ctx context.Context, conn net.Conn, messages chan<- interface{}, errors chan<- error) {
```
golangcibot: `conn` can be `io.Reader` (from `interfacer`)

lib/beacon/broadcast.go
```diff
@@ -109,7 +110,7 @@ func readBroadcasts(outbox chan<- recv, port int, stop chan struct{}) error {
 	defer close(done)
```
calmh: `done` should be a child context I think, or potentially same context if there is no other path out the function than via the passed context (I didn't check).

I see there are other places with the same thing, not going to mention them, this might be one of those things you were considering for a followup PR?

lib/connections/service.go
```diff
@@ -185,18 +186,18 @@ func NewService(cfg config.Wrapper, myID protocol.DeviceID, mdl Model, tlsCfg *t
 	// the common handling regardless of whether the connection was
 	// incoming or outgoing.
 
-	service.Add(util.AsService(service.connect))
-	service.Add(util.AsService(service.handle))
+	service.Add(util.AsService(service.connect, service.String()))
```
calmh: So in a lot of places where we create "sub"-services I think just passing on the name of the parent service isn't appropriate. These should be `fmt.Sprintf("%s/connector", service)` and so on to differentiate or the name will be meaningless?

imsodin:
Sorry for inflicting this diff on you... And thanks (and kudos) for catching the names issue despite the monstrosity.

lib/connections/service.go
```diff
@@ -185,18 +186,18 @@ func NewService(cfg config.Wrapper, myID protocol.DeviceID, mdl Model, tlsCfg *t
...
+	service.Add(util.AsService(service.handle, service.String()))
```
AudriusButkevicius: Should be different names?

lib/discover/local.go
```diff
@@ -81,9 +82,9 @@ func NewLocal(id protocol.DeviceID, addr string, addrList AddressLister, evLogge
 		c.beacon = beacon.NewMulticast(addr)
 	}
 	c.Add(c.beacon)
-	c.Add(util.AsService(c.recvAnnouncements))
+	c.Add(util.AsService(c.recvAnnouncements, c.String()))
 
-	c.Add(util.AsService(c.sendLocalAnnouncements))
+	c.Add(util.AsService(c.sendLocalAnnouncements, c.String()))
```
AudriusButkevicius: Different names?

lib/beacon/beacon.go
```diff
@@ -63,23 +64,23 @@ func newCast(name string) *cast {
...
+func (c *cast) createService(svc func(context.Context) error, suffix string) util.ServiceWithError {
+	return util.AsServiceWithError(func(ctx context.Context) error {
 		l.Debugln("Starting", c.name, suffix)
-		err := svc(stop)
+		err := svc(ctx)
 		l.Debugf("Stopped %v %v: %v", c.name, suffix, err)
 		return err
-	})
+	}, c.String())
 }
```
AudriusButkevicius: Add suffix?

lib/model/folder_summary.go
```diff
@@ -63,8 +64,8 @@ func NewFolderSummaryService(cfg config.Wrapper, m Model, id protocol.DeviceID,
 		lastEventReqMut: sync.NewMutex(),
 	}
 
-	service.Add(util.AsService(service.listenForUpdates))
-	service.Add(util.AsService(service.calculateSummaries))
+	service.Add(util.AsService(service.listenForUpdates, service.String()))
+	service.Add(util.AsService(service.calculateSummaries, service.String()))
```
AudriusButkevicius: Different names?

lib/relay/client/client.go
```diff
@@ -81,3 +82,7 @@ func (c *commonClient) Invitations() chan protocol.SessionInvitation {
 	defer c.mut.RUnlock()
 	return c.invitations
 }
+
+func (c *commonClient) String() string {
+	return fmt.Sprintf("commonClient/@%p", c)
+}
```
AudriusButkevicius: We probably want this on the "derived" implementations?

imsodin:
The context introduced here only transfer signals from stopped services - that's not really an error. E.g. a listener we want an error if the connection dies, but if it cleanly shuts down, I consider a nil error correct. If we did that, we'd need checks higher up (wherever we use the error) to filter out `context.Cancelled` errors.

PR #6168, opened 2019-11-20
gui/default: Allow to degrade 'Automatic upgrades' option to 'No upgrades' (fixes #6044)

### LINKED ISSUE #6044: Clicking "Save" in options says "You have unsaved changes" when changing automatic upgrade setting
This happens to three virtual machines running centOS 7 and syncthing 1.3.0 .
In the options, selecting **specifically** "No upgrades" when "Stable releases and release candidates" is selected and in effect, then saving causes the "You have unsaved changes" popup to appear, making it impossible to switch directly from "Stable and candidates" to "No upgrades".
It's as if clicking outside the window or "Close".
To select it, I had to switch first from "Stable and candidates" to "Stable only", save, then go back and select "No updates", then save.
Not a software-breaking bug but worth mentionning.

AkaZecik:
### Purpose

Fix for the issue #6044, where one couldn't change `Automatic upgrades` option from `Stable releases and release candidates` to `No upgrades` without changing it to `Stable releases only` first.

### Testing

I did only a manual testing. I had to force the option 'No upgrades' to appear, because the cloned version of the repository automatically hides it, unlike the official release version. I simply modified the condition to always be `true`.

The new configuration got saved, based on the console logs from the `GET /rest/system/config` call. However, restarting Syncthing somehow causes the settings to get reverted and it happens only if I leave it at `No upgrades` before restarting.

I suppose it's because the cloned repo is a release candidate and not a 'stable' version, but I haven't digged dipper into that.


gui/default/syncthing/settings/settingsModalView.html
```diff
@@ -327,7 +327,7 @@
     <button type="button" class="btn btn-primary btn-sm" ng-click="saveSettings()">
       <span class="fas fa-check"></span>&nbsp;<span translate>Save</span>
     </button>
-    <button type="button" class="btn btn-default btn-sm" data-dismiss="modal">
+    <button type="button" class="btn btn-default btn-sm" ng-click="closeSettings()">
       <span class="fas fa-times"></span>&nbsp;<span translate>Close</span>
```
AudriusButkevicius: Clicking out of the modal also closes it, which would now no longer prompt users.

PR #6176, opened 2019-11-22
lib/model, lib/protocol: Add contexts sending indexes and download-progress

imsodin:
Not much interesting here, just a continuation of context on requests (https://github.com/syncthing/syncthing/pull/6149) and context on services (https://github.com/syncthing/syncthing/pull/6166).

### DIFF
--- a/lib/protocol/protocol_test.go
+++ b/lib/protocol/protocol_test.go
@@ -75,10 +75,12 @@ func TestClose(t *testing.T) {
 		t.Error("Ping should not return true")
 	}
 
-	c0.Index("default", nil)
-	c0.Index("default", nil)
+	ctx := context.Background()
 
-	if _, err := c0.Request(context.Background(), "default", "foo", 0, 0, nil, 0, false); err == nil {
+	c0.Index(ctx, "default", nil)
+	c0.Index(ctx, "default", nil)
+
+	if _, err := c0.Request(ctx, "default", "foo", 0, 0, nil, 0, false); err == nil {
 		t.Error("Request should return an error")
 	}
 }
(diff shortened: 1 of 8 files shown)


lib/protocol/protocol_test.go
```diff
@@ -75,10 +75,12 @@ func TestClose(t *testing.T) {
 		t.Error("Ping should not return true")
 	}
 
-	c0.Index("default", nil)
-	c0.Index("default", nil)
+	ctx := context.Background()
 
-	if _, err := c0.Request(context.Background(), "default", "foo", 0, 0, nil, 0, false); err == nil {
+	c0.Index(ctx, "default", nil)
```
golangcibot: Error return value of `c0.Index` is not checked (from `errcheck`)

lib/protocol/protocol_test.go
```diff
@@ -75,10 +75,12 @@ func TestClose(t *testing.T) {
...
+	c0.Index(ctx, "default", nil)
```
golangcibot: Error return value of `c0.Index` is not checked (from `errcheck`)

lib/protocol/protocol_test.go
```diff
@@ -152,7 +154,7 @@ func TestCloseRace(t *testing.T) {
 	c0.ClusterConfig(ClusterConfig{})
 	c1.ClusterConfig(ClusterConfig{})
 
-	c1.Index("default", nil)
+	c1.Index(context.Background(), "default", nil)
```
golangcibot: Error return value of `c1.Index` is not checked (from `errcheck`)

calmh:
Lot of passing contexts to things that don't act on them at all? Or there's some deeper purpose? (And, I think, a couple of context.Background() that should probably be context.TODO()?)

imsodin:
protocol.send uses it. Admittedly there's no specific use case for this change, but adding a context to `Request` I needed to put quite a few `context.TODO`s. And I like striking off easy todos.

As to new Backgrounds vs TODOs: I put Backgrounds in `ProgressEmitter.clearLocked` as it's called from `CommitConfiguration`, i.e. asynchronously, so there's nothing to cancel on. and in tests as I don't see any need to replace them with real contexts.

PR #6181, opened 2019-11-23
lib: Wrap errors with errors.Wrap instead of fmt.Errorf

imsodin:
This is prompted by concern over wrapped errors when filtering errors in https://github.com/syncthing/syncthing/pull/6177. And it's purely a grep/sed effort. We may still be dropping errors in places that could/should be wrapped.


calmh:
Then when all of this is done, migrate to xerrors.Errorf() and friends :D

PR #6186, opened 2019-11-25
lib/versioner: Reduce surface area

calmh:
This is a refactor while I was anyway rooting around in the versioner.
Instead of exporting every possible implementation and the factory and
letting the caller do whatever, this now encapsulates all that and
exposes a New() that takes a config.VersioningConfiguration.

Given that and that we don't know (from the outside) how a versioner
works or what state it keeps, we now just construct it once per folder
and keep it around. Previously it was recreated for each restore
request.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -328,8 +328,10 @@ func (m *model) startFolderLocked(cfg config.FolderConfiguration) {
 	_ = ffs.Hide(".stversions")
 	_ = ffs.Hide(".stignore")
 
+	var ver versioner.Versioner
 	if cfg.Versioning.Type != "" {
-		ver, err := versioner.New(ffs, cfg.Versioning)
+		var err error
+		ver, err = versioner.New(ffs, cfg.Versioning)
 		if err != nil {
 			panic(fmt.Errorf("creating versioner: %v", err))
 		}
@@ -340,12 +342,12 @@ func (m *model) startFolderLocked(cfg config.FolderConfiguration) {
 			token := m.Add(service)
 			m.folderRunnerTokens[folder] = append(m.folderRunnerTokens[folder], token)
 		}
-		m.folderVersioners[folder] = ver
 	}
+	m.folderVersioners[folder] = ver
 
 	ignores := m.folderIgnores[folder]
 
-	p := folderFactory(m, fset, ignores, cfg, m.folderVersioners[folder], ffs, m.evLogger)
+	p := folderFactory(m, fset, ignores, cfg, ver, ffs, m.evLogger)
 
 	m.folderRunners[folder] = p
 
@@ -2455,8 +2457,11 @@ func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly 
 
 func (m *model) GetFolderVersions(folder string) (map[string][]versioner.FileVersion, error) {
 	m.fmut.RLock()
-	ver := m.folderVersioners[folder]
+	ver, ok := m.folderVersioners[folder]
 	m.fmut.RUnlock()
+	if !ok {
+		return nil, errFolderMissing
+	}
 	if ver == nil {
 		return nil, errNoVersioner
 	}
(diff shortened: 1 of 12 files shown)


lib/versioner/external.go
```diff
@@ -21,22 +21,22 @@ import (
...
 }
 
-type External struct {
+type external struct {
 	command    string
 	filesystem fs.Filesystem
 }
 
-func NewExternal(folderID string, filesystem fs.Filesystem, params map[string]string) Versioner {
+func newExternal(folderID string, filesystem fs.Filesystem, params map[string]string) Versioner {
```
golangcibot: `newExternal` - `folderID` is unused (from `unparam`)

lib/versioner/simple.go
```diff
@@ -15,22 +15,22 @@ import (
...
 
-type Simple struct {
+type simple struct {
 	keep       int
 	folderFs   fs.Filesystem
 	versionsFs fs.Filesystem
 }
 
-func NewSimple(folderID string, folderFs fs.Filesystem, params map[string]string) Versioner {
+func newSimple(folderID string, folderFs fs.Filesystem, params map[string]string) Versioner {
```
golangcibot: `newSimple` - `folderID` is unused (from `unparam`)

lib/versioner/staggered.go
```diff
@@ -22,26 +22,26 @@ import (
...
 	versionsFs    fs.Filesystem
-	interval      [4]Interval
+	interval      [4]interval
 	mutex         sync.Mutex
 
 	testCleanDone chan struct{}
 }
 
-func NewStaggered(folderID string, folderFs fs.Filesystem, params map[string]string) Versioner {
+func newStaggered(folderID string, folderFs fs.Filesystem, params map[string]string) Versioner {
```
golangcibot: `newStaggered` - `folderID` is unused (from `unparam`)

lib/versioner/trashcan.go
```diff
@@ -20,21 +20,21 @@ import (
...
-type Trashcan struct {
+type trashcan struct {
 	suture.Service
 	folderFs     fs.Filesystem
 	versionsFs   fs.Filesystem
 	cleanoutDays int
 }
 
-func NewTrashcan(folderID string, folderFs fs.Filesystem, params map[string]string) Versioner {
+func newTrashcan(folderID string, folderFs fs.Filesystem, params map[string]string) Versioner {
```
golangcibot: `newTrashcan` - `folderID` is unused (from `unparam`)

AudriusButkevicius:
Baah, more cruft in the model, why can't we just cache it like we do with filesystem?

calmh:
The model is the place where that caching is supposed to happen, imho. That it also does too many other things is unfortunate. Constructing things in the config is never right imho, especially not when we're also looking at stuff like "transactionally" replacing the entire config...

lib/model/model.go
```diff
@@ -2444,14 +2454,11 @@ func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
 
 func (m *model) GetFolderVersions(folder string) (map[string][]versioner.FileVersion, error) {
-	fcfg, ok := m.cfg.Folder(folder)
-	if !ok {
-		return nil, errFolderMissing
-	}
-
-	ver := fcfg.Versioner()
+	m.fmut.RLock()
+	ver := m.folderVersioners[folder]
```
imsodin: I'd keep returning `errFolderMissing` if `!ok` (using `ver, ok := ...`) - it's a tiny bit more accurate.

AudriusButkevicius:
Does this deal with versioner reconfiguration and restarting it?

imsodin:
> Does this deal with versioner reconfiguration and restarting it?

Versioner config is part of folder config - so a change triggers a folder restart, which reconstructs the versioner.

calmh:
Also we were already handling starting and stopping the versioner in the model... We were just assuming that restores didn't require the versioner to be running.

PR #6187, opened 2019-11-25
lib/connections: Dialer code deduplication

imsodin:
I have a change brewing towards https://github.com/syncthing/syncthing/issues/5298 and that diff is going to be hideously huge. This PR is part of making it a bit smaller by separating whatever I can.

### DIFF
--- a/lib/connections/relay_dial.go
+++ b/lib/connections/relay_dial.go
@@ -24,8 +24,7 @@ func init() {
 }
 
 type relayDialer struct {
-	cfg    config.Wrapper
-	tlsCfg *tls.Config
+	commonDialer
 }
 
 func (d *relayDialer) Dial(id protocol.DeviceID, uri *url.URL) (internalConn, error) {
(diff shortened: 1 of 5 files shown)


lib/connections/relay_dial.go
```diff
@@ -66,17 +65,14 @@ func (d *relayDialer) Dial(id protocol.DeviceID, uri *url.URL) (internalConn, er
...
 type relayDialerFactory struct{}
 
-func (relayDialerFactory) New(cfg config.Wrapper, tlsCfg *tls.Config) genericDialer {
-	return &relayDialer{
-		cfg:    cfg,
-		tlsCfg: tlsCfg,
-	}
+func (relayDialerFactory) New(opts config.OptionsConfiguration, tlsCfg *tls.Config) genericDialer {
+	return &relayDialer{commonDialer{
+		trafficClass:      opts.TrafficClass,
```
AudriusButkevicius: This makes it not pick up config changes, as it gets a one off static copy of the config.

imsodin: A dialer doesn't need to pick up, they get constructed on every connect loop iteration (timed).

PR #6188, opened 2019-11-25
lib/config, lib/connections: Remove ListenAddresses hack

imsodin:
This is also mainly to reduce an upcoming diff. However removing that wrap-hack in `connections.service` is in my opinion anyway worthwhile.

### DIFF
--- a/lib/config/migrations.go
+++ b/lib/config/migrations.go
@@ -248,9 +248,9 @@ func migrateToConfigV14(cfg *Configuration) {
 	hasDefault := false
 	for _, raddr := range cfg.Options.DeprecatedRelayServers {
 		if raddr == "dynamic+https://relays.syncthing.net/endpoint" {
-			for i, addr := range cfg.Options.ListenAddresses {
+			for i, addr := range cfg.Options.RawListenAddresses {
 				if addr == "tcp://0.0.0.0:22000" {
-					cfg.Options.ListenAddresses[i] = "default"
+					cfg.Options.RawListenAddresses[i] = "default"
 					hasDefault = true
 					break
 				}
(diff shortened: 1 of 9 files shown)


lib/config/migrations.go
```diff
@@ -248,9 +248,9 @@ func migrateToConfigV14(cfg *Configuration) {
 	hasDefault := false
 	for _, raddr := range cfg.Options.DeprecatedRelayServers {
 		if raddr == "dynamic+https://relays.syncthing.net/endpoint" {
-			for i, addr := range cfg.Options.ListenAddresses {
+			for i, addr := range cfg.Options.RawListenAddresses {
 				if addr == "tcp://0.0.0.0:22000" {
-					cfg.Options.ListenAddresses[i] = "default"
+					cfg.Options.RawListenAddresses[i] = "default"
```
golangcibot: string `default` has 4 occurrences, make it a constant (from `goconst`)

AudriusButkevicius:
Seems like a noop change, which potentially makes things worse by pushing the logic deeper. What's the driver for this?

imsodin:
```diff
	for _, addr := range config.Wrap("", to, s.evLogger).ListenAddresses() {
	for _, addr := range to.Options.ListenAddresses() {
```

That's the motivation. Alternatively I can undo the `Raw...` stuff and put `ListenAddresses` on `config.Configuration`. I just thought it belongs to `OptionsConfiguration` thematically.

AudriusButkevicius:
I guess we should scrap cfg.ListenAddresses then, and just keep cfg.Options().ListenAddresses(), but I guess we should do it for the rest of the stuff too, as now it's inconsistent

imsodin:
> Also, GUI address does not need same treatment?

It is already `GUIConfiguration.RawAddress` - or did you mean something else?

AudriusButkevicius:
I guess I was questioning whether its on the wrapper or on GUI object, being lazy to check myself.

imsodin:
Already on GUI object ;)

PR #6191, opened 2019-11-26
Bump Syncthing version in GitHub issue template

Otiel:
Since Syncthing current version is v1.3.1, the current proposal is that the GitHub issue template shows `v1.x.y` instead of `v0.x.y`.


AudriusButkevicius:
Thanks

PR #6193, opened 2019-11-26
Ensure uPnP http requests have trailing \r\n

tmm1:
The http request used for upnp discovery was malformed.

See https://play.golang.org/p/-S2WDBwK29m and notice the request ends in `\r\n` instead of `\r\n\r\n`


AudriusButkevicius:
Thanks.

PR #6197, opened 2019-11-27
Don't create empty ".stignore" files (fixes #6190)

### LINKED ISSUE #6190: Do not create .stignore if there are no ignore patterns
### TLDR;

Syncthing should not create empty `.stignore` files as it clutters the folder contents.

### Prerequisites

1. Choose a folder that doesn't have any Ignore patterns.
2. Make sure there's no existing `.stignore` file or delete it.

### Steps to reproduce

1. Go to the Syncthing GUI, and click "Edit" the selected folder.
2. Hit "Save" (without changing anything).

### Actual results

A new, empty `.stignore` file is created.

### Expected results

When hitting "Save", if there are no Ignore patterns specified, Syncthing should not create a `.stignore` file, or should delete it if it already exists.

### Version Information

Syncthing Version: v1.3.1

orangefoil:
### Purpose

This will
1. Prevent creation of a new .stignore file if there are no ignore patterns (fixes #6190).
2. Delete an existing .stignore if all ignore patterns are removed.

### Testing

An automated test case is included with this pull request.  
If desired the changes can be tested manually as follows:
1. Go to the Syncthing GUI.
2. Edit a folder that doesn't have any ignore patterns.
3. Click "save" without making any changes. 
4. Observe that no .stignore file was created.
5. Edit the folder again and add an ignore pattern.
6. Observe that a .stignore file was created.
7. Edit the folder yet again and remove all ignore patterns.
8. Observe that the .stignore file was deleted.

### A note on software design
I'm undecided if the new `DeleteIgnores()` function in `lib/ignore/ignore.go` adds unnecessary complexity and whether it's function body should be inlined in `lib/model/model.go` instead.
Should you consider merging this PR I'd kindly request some feedback on this.

### DIFF
--- a/lib/api/testdata/config/csrftokens.txt
+++ b/lib/api/testdata/config/csrftokens.txt
@@ -1,2 +1,0 @@
-aMkpEUzSEfniyWxHaiVCf4jrRYDSckWo
-DqQ3MrasdMPT52WgRPKgiZA6J2jkiAgt
--- a/lib/api/testdata/config/csrftokens.txt
+++ b/lib/api/testdata/config/csrftokens.txt
@@ -1,0 +1,2 @@
+aMkpEUzSEfniyWxHaiVCf4jrRYDSckWo
+DqQ3MrasdMPT52WgRPKgiZA6J2jkiAgt
--- a/lib/ignore/ignore.go
+++ b/lib/ignore/ignore.go
@@ -564,10 +564,11 @@ func WriteIgnores(filesystem fs.Filesystem, path string, content []string) error
 
 // DeleteIgnores should be called instead of WriteIgnores if parameter "content" is empty
 func DeleteIgnores(filesystem fs.Filesystem, path string) error {
-	if osutil.IsDeleted(filesystem, path) {
+	err := filesystem.Remove(path)
+	if fs.IsNotExist(err) {
 		return nil
 	}
-	return filesystem.Remove(path)
+	return err
 }
 
 type modtimeCheckerKey struct {
--- a/lib/ignore/ignore.go
+++ b/lib/ignore/ignore.go
@@ -562,6 +562,14 @@ func WriteIgnores(filesystem fs.Filesystem, path string, content []string) error
 	return nil
 }
 
+// DeleteIgnores should be called instead of WriteIgnores if parameter "content" is empty
+func DeleteIgnores(filesystem fs.Filesystem, path string) error {
+	if osutil.IsDeleted(filesystem, path) {
+		return nil
+	}
+	return filesystem.Remove(path)
+}
+
 type modtimeCheckerKey struct {
 	fs   fs.Filesystem
 	name string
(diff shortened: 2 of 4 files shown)


calmh:
Thanks! Frankly I think this should happen inside `WriteIgnores`. If the content length (after stripping whitespace?) is zero, it just does an fs.Remove directly. Instead of checking existence beforehand, checking the error for fs.IsNotExist() and translating that to nil is safer.

AudriusButkevicius:
I think we should not yank the file if the user created it, and should simply do nothing if its empty.

calmh:
That seems overly complex (and potentially race prone) in that we need to act based on not only the passed content but whether the file already exists or not, as we *should* then delete if it exists and is non-empty but no do anything if it exists and is precisely empty? 

I think that if the ignore editor in the GUI is used the file is "ours" and we can do what we want. Removing it when the patterns are removed seems the proper thing to me regardless of the previous state.

AudriusButkevicius:
Well they can't, as we'll nuke it every time someone saves folders settings in the UI.

calmh:
Right. So if they want to do that they get to add a comment to the file saying `// This file intentionally left blank`. I don't see a huge issue here.

AudriusButkevicius:
And whats the big issue with simply not creating file if not necessary?
It still covers the original complaint as is less parasitic

calmh:
This is deep in bikeshedding territory imho, but honestly I'm not sure what you're proposing. What I'm saying:

- When saving, is there content to save?
  - yes: save it
  - no: delete the file (doesn't matter if it already exists or not)

What you're proposing, as I understand it:

- When saving, is there content to save?
  - yes: save it
  - no: look at the existing file
     - empty or nonexistent: do nothing
     - not empty: write an empty file? or delete it?

My dislike in this is the (admittedly trivial) race condition between "look at the existing file" and "act on that information" plus the requirement to having to do that to begin with instead of the easy solution to always act the same...

AudriusButkevicius:
Ok, then I think we should scrap the ignores.DeleteIgnores as it does the check then do bahviour, and just do `fs.Remove(path)` check if it's not non-exists, and propagate that up.

imsodin:
@orangefoil : My interpretation of the lengthy discussion:

 - Everything should happen in `WriteIgnore`.

 - Just call `err := fs.Remove` and then return a nil error if `IsNotExist(err)` instead of checking for existence before.

I hope you bear with us ;)

orangefoil:
Thank you all for your feedback and patience!

My reasoning for creating the `DeleteIgnores` function was that it might be confusing and counter-intuitive that a function called `WriteIgnores` does not always write ignores, but delete a file in some cases.  
However I agree that having everything happen in `WriteIgnores` and leaving `lib/model/model.go` untouched is probably less convoluted.

Right now I don't quite understand why the "Check Old Go Version" check is failing and what the generated output is trying to tell me.

calmh:
Thanks!

PR #6198, opened 2019-11-28
cmd/syncthing: Implement log rotation (fixes #6104)

### LINKED ISSUE #6104: WIndows: Syncthing log file filling up primary drive.

Log grows to 69 GB I can not open the file to read it. Too large to see what is being written in there. 

### Please do post actual bug reports and feature requests.

Syncthing log file grows to GBs in size, this morning it was 69GB in size. Had a couple system do this to me now where the log file fills up like that. It fills the main drive, the main drive gets full and Syncthing stops syncing due to space constraints. 
Last time I updated the exe and rebooted the system and it ran fine for a while, about 2 weeks. 
This is the second time in 2 days that the log file has filled up like this on my server. 

### Version Information

Syncthing Version: v1.3.0
Windows 7 Pos and Windows Server 2008 R2 Datacenter
Syncthing is running as a service.

calmh:
Since we've taken upon ourselves to create a log file by default on Windows, this adds proper management of that log file. There are two new options:

```
-log-max-old-files="3"    Number of old files to keep (zero to keep only current).
-log-max-size="10485760"  Maximum size of any file (zero to disable log rotation).
```

The default values result in four files (syncthing.log, synchting.0.log, ..., syncthing.3.log) each up to 10 MiB in size. To not use log rotation at all, the user can say --log-max-size=0.

### DIFF
--- a/cmd/syncthing/monitor.go
+++ b/cmd/syncthing/monitor.go
@@ -9,10 +9,12 @@ package main
 import (
 	"bufio"
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"os/signal"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"syscall"
@@ -48,7 +50,15 @@ func monitorMain(runtimeOptions RuntimeOptions) {
 
 	logFile := runtimeOptions.logFile
 	if logFile != "-" {
-		var fileDst io.Writer = newAutoclosedFile(logFile, logFileAutoCloseDelay, logFileMaxOpenTime)
+		var fileDst io.Writer
+		if runtimeOptions.logMaxSize > 0 {
+			open := func(name string) (io.WriteCloser, error) {
+				return newAutoclosedFile(name, logFileAutoCloseDelay, logFileMaxOpenTime), nil
+			}
+			fileDst = newRotatedFile(logFile, open, int64(runtimeOptions.logMaxSize), runtimeOptions.logMaxFiles, time.Second)
+		} else {
+			fileDst = newAutoclosedFile(logFile, logFileAutoCloseDelay, logFileMaxOpenTime)
+		}
 
 		if runtime.GOOS == "windows" {
 			// Translate line breaks to Windows standard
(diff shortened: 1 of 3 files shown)


cmd/syncthing/monitor_test.go
```diff
@@ -7,13 +7,128 @@
...
+	// new bytes don't fit.
+	if _, err := rf.Write(testData); err != nil {
+		t.Fatal(err)
+	}
+	checkSize(t, logName, len(testData))
+	checkSize(t, logName+".0", len(testData))
+	checkNotExist(t, logName+".1")
+
+	// Write another byte. That should fit without causing an extra rotate.
+	rf.Write([]byte{42})
```
golangcibot: Error return value of `rf.Write` is not checked (from `errcheck`)

xjtdy888:
Do we consider gz-compressing archived log files?

cmd/syncthing/monitor.go
```diff
@@ -317,6 +326,87 @@ func restartMonitorWindows(args []string) error {
...
+}
+
+func (r *rotatedFile) rotate() {
+	// The files are named "name", "name.0", "name.1", ...
+	// "name.(r.maxFiles-1)". Increase the numbers on the
+	// suffixed ones.
+	for i := r.maxFiles - 1; i > 0; i-- {
+		from := fmt.Sprintf("%s.%d", r.name, i-1)
+		to := fmt.Sprintf("%s.%d", r.name, i)
+		err := os.Rename(from, to)
```
AudriusButkevicius: I am not sure if "rename on top" works on all platforms tho?

calmh: Test pass on Windows, which is the only one that I would expect trouble from 🤷‍♂

calmh:
No gzip here, this is mostly for Windows users who wouldn't know what to do with it. In fact, that's an argument for not doing this with the file extension either, so let me fix that too...

xjtdy888:
Previously we always cleared the log every time we started f.opened.IsZero() flags |= os.O_TRUNC
Now we have log rotate, maybe appending or doing a log rotation would be better

calmh:
Yes. This in fact does a log rotation on startup. Means we might rotate away logs quicker than expected but I think that's fine. Previously we'd have just kept *no* old logs.

cmd/syncthing/monitor.go
```diff
@@ -317,6 +327,94 @@ func restartMonitorWindows(args []string) error {
...
+}
+
+func (r *rotatedFile) Write(bs []byte) (int, error) {
+	// Do the rotate check at most once a minStatInterval once any given file is
+	// open. Limits the number of stat() calls when there's a lot of logging,
+	// and limits the number of warnings if we're stuck unable to rotate for
+	// whatever reason.
+	if r.current != nil && time.Since(r.lastStat) >= r.minStatInterval {
+		info, err := os.Lstat(r.name)
+		if err != nil && !os.IsNotExist(err) {
```
AudriusButkevicius: We are the ones writing to this file, why can't we just track the number of bytes we've written so far?

calmh: We could do that, especially given we start from zero, so yeah, let me fix.

PR #6201, opened 2019-11-28
gui: Sort versions by date in restore dropdown

rschoon:
### Purpose

In the GUI, the versions are not sorted in the restore modal, appearing in whatever random order the API is providing them as.  This is especially annoying to look through if you have a lot of versions for a file.  This PR forces the versions to be sorted by date in reverse order (so that after the current version, the next version is the latest version before that, and so forth).

### Testing

I've tested it with staggered versioning in chrome and firefox, and you can see the change in drop down behavior in the screenshot below.

### Screenshots

Before:
syncthing-gui-sort-before

After:
syncthing-gui-sort-after

### Documentation

I didn't actually see anything about that part of the UI in the docs.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -2056,6 +2056,9 @@ angular.module('syncthing.core')
                                     value.modTime = new Date(value.modTime);
                                     value.versionTime = new Date(value.versionTime);
                                 });
+                                values.sort(function (a, b) {
+                                    return b.versionTime - a.versionTime;
+                                });
                             });
                             if (closed) return;
                             $scope.restoreVersions.versions = data;


AudriusButkevicius:
I wonder why these come out unsorted from the backend. Perhaps we should fix that as well?

rschoon:
I didn't really dig into what the backend was doing because it looks like it gets to the specific versioner pretty quickly, but I'm pretty certain it's just returning in filesystem order (ls -fl shows the actual files in .stversions is in the same order as the before screenshot) which is of course somewhat arbitrary.

AudriusButkevicius:
Hmm, that's what I was expecting. The date we embed in the filename sorts lexically, so why would ls show output out of order like in the ui?

PR #6203, opened 2019-11-29
all: Propagate errors from NamespacedKV

calmh:
As foretold by the prophecy, "once the database refactor is merged, then
shall appear a request to propagate errors from the store known
throughout the land as the NamedspacedKV, and it shall be good".

### DIFF
--- a/lib/stats/folder.go
+++ b/lib/stats/folder.go
@@ -51,8 +51,6 @@ func (s *FolderStatisticsReference) GetLastFile() (LastFile, error) {
 	deleted, _, err := s.ns.Bool("lastFileDeleted")
 	if err != nil {
 		return LastFile{}, err
-	} else if !ok {
-		return LastFile{}, nil
 	}
 	return LastFile{
 		At:       at,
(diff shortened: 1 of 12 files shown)


lib/stats/folder.go
```diff
@@ -35,45 +35,71 @@ func NewFolderStatisticsReference(ldb *db.Lowlevel, folder string) *FolderStatis
...
+	file, ok, err := s.ns.String("lastFileName")
+	if err != nil {
+		return LastFile{}, err
+	} else if !ok {
+		return LastFile{}, nil
+	}
+	deleted, _, err := s.ns.Bool("lastFileDeleted")
+	if err != nil {
+		return LastFile{}, err
+	} else if !ok {
```
imsodin: This check is redundant, unless the `ok` shouldn't be omitted in the last assignment above (also if it was purposely omitted, maybe add a comment - I don't understand why it might be).

calmh: That's a copy-paste-o

PR #6204, opened 2019-11-30
lib/db: Remove *instance by making everything *Lowlevel

imsodin:
The distinction between `instance` and `Lowlevel` is now pretty much moot. On every entry point to the db package where we get a `Lowlevel,` we pass that to `newInstance`. Just adding `keyer` to `Lowlevel` was literally everything that's needed to change to get rid of `*instance` (along with s/instance/Lowlevel/g and mv instance.go lowlevel.go).


lib/db/lowlevel.go
```diff
@@ -19,18 +23,835 @@ type Lowlevel struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	vl, ok := unmarshalVersionList(bs)
+	if !ok {
+		return nil, nil
+	}
+
+	var devices []protocol.DeviceID
```
golangcibot: Consider preallocating `devices` (from `prealloc`)

imsodin:
I didn't even check Github's diff... I expected that it detects the similarity. Indeed the code is just moved and `s/instance/Lowlevel/g`.

PR #6207, opened 2019-12-03
lib/osutil: Increase maxfiles on macOS properly (Fixes #6206)

### LINKED ISSUE #6206: Soft limit for maxfiles is NOT increased to the hard one on macOS
### Expected behaviour
During the startup, the soft limit for the number of concurrently open files, a.k.a. `maxfiles` in `macOS`,  is increased up the hard one.
 
### Actual behaviour
During the startup, the soft limit for the number of concurrently open files, a.k.a. `maxfiles` in `macOS`, is not increased up to the hard one. 

During syncing of large directories, there are many "too many open files errors" on the `stdout`.

### Steps to reproduce
1. Set the soft limit for `maxfiles` to an extremely small number (e.g. 30) and the hard one to `unlimited`
```console
sudo launchctl limit maxfiles 30 unlimited
``` 

2. Start `syncthing`

### Version Information
Syncthing Version: **syncthing v1.3.1 "Fermium Flea" (go1.13.4 darwin-amd64)**
OS Version: **macOS 10.15.1 Catalina**

pbrit:

### DIFF
--- a/lib/osutil/rlimit_unix.go
+++ b/lib/osutil/rlimit_unix.go
@@ -8,7 +8,14 @@
 
 package osutil
 
-import "syscall"
+import (
+	"runtime"
+	"syscall"
+)
+
+const (
+	darwinOpenMax = 10240
+)
 
 // MaximizeOpenFileLimit tries to set the resource limit RLIMIT_NOFILE (number
 // of open file descriptors) to the max (hard limit), if the current (soft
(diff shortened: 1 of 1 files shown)


lib/osutil/rlimit_unix.go
```diff
@@ -8,7 +8,14 @@
...
 package osutil
 
-import "syscall"
+import (
+	"runtime"
+	"syscall"
+)
+
+const (
+	OPEN_MAX = 10240
```
golangcibot: don't use ALL_CAPS in Go names; use CamelCase (from `golint`)

calmh:
Thanks for the issue and the fix!

Issue #6209, opened 2019-12-03
github.com/syncthing/syncthing/lib/protocol tests running out of memory on 32bit arches


decathorpe:
With the changes between syncthing 1.3.1 and 1.3.2, running the tests for `github.com/syncthing/syncthing/lib/protocol` on 32 bit architectures (i686 and armv7hl) now runs out of system memory:

on armv7hl:

```
runtime: out of memory: cannot allocate 20922368-byte block (2805399552 in use)
fatal error: out of memory
```

on i686:

```
runtime: out of memory: cannot allocate 23289856-byte block (3761274880 in use)
fatal error: out of memory
```

Can I ignore this issue on 32 bit architectures?

Syncthing Version: v1.3.2
OS Version: fedora (32)

imsodin:
There have been changes to that part of the code in 1.3.2 and allocating ~20GiB doesn't seem right. Do you get stacktraces/test info about when/where this happens?

decathorpe:
Yes, sorry, I should have included that:

```
runtime: out of memory: cannot allocate 19210240-byte block (3765501952 in use)
fatal error: out of memory
runtime stack:
runtime.throw(0x56a0c555, 0xd)
	/usr/lib/golang/src/runtime/panic.go:774 +0x70
runtime.largeAlloc(0x1251f7c, 0xbe0101, 0xfa9c3d5)
	/usr/lib/golang/src/runtime/malloc.go:1140 +0x11a
runtime.mallocgc.func1()
	/usr/lib/golang/src/runtime/malloc.go:1033 +0x3a
runtime.systemstack(0x524fd700)
	/usr/lib/golang/src/runtime/asm_386.s:399 +0x62
runtime.mstart()
	/usr/lib/golang/src/runtime/proc.go:1146
goroutine 36 [running]:
runtime.systemstack_switch()
	/usr/lib/golang/src/runtime/asm_386.s:360 fp=0x58c7ee7c sp=0x58c7ee78 pc=0x566a08a0
runtime.mallocgc(0x1251f7c, 0x56b9b180, 0x56818001, 0x5668a9c5)
	/usr/lib/golang/src/runtime/malloc.go:1032 +0x6ff fp=0x58c7eed0 sp=0x58c7ee7c pc=0x566503ff
runtime.makeslice(0x56b9b180, 0x1251f7c, 0x1251f7c, 0x1c0)
[...]

PR #6210, opened 2019-12-03
lib/protocol: Decrease runtime/mem usage of bufferpool stresstest (ref #6209)

imsodin:
`TestStressBufferPool` uses lots of memory. To see what it might use in a worst case scenario of continuous incoming requests, I extended the test time until the max memory usage didn't increase anymore, which happened at 20s/~4GiB. I think that's not unreasonable, i.e. nothing needs to change in production code.

However the test as it stands also often uses 3GiB on my system, which seems a bit excessive. I changed it to run just as long as required (in 100ms steps) to hit all code paths (which was previously checked after 2s). For me this usually takes ~0.4GiB).

### DIFF
--- a/lib/protocol/bufferpool_test.go
+++ b/lib/protocol/bufferpool_test.go
@@ -3,7 +3,7 @@
 package protocol
 
 import (
-	"sync"
+	"context"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


lib/protocol/bufferpool_test.go
```diff
@@ -70,18 +70,44 @@ func TestStressBufferPool(t *testing.T) {
...
-	const runtime = 2 * time.Second
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
 
 	bp := newBufferPool()
-	t0 := time.Now()
 
-	var wg sync.WaitGroup
+	checkDone := func() bool {
+		if bp.puts == 0 || bp.skips == 0 || bp.misses == 0 {
```
calmh: Reading these while the test is running seems like a race condition

imsodin: I thought that was ok because they are atomics. Anyway if, as you noted in the issue, it isn't uncommon for non-short test to use that much memory, this PR can just be closed.

PR #6212, opened 2019-12-04
lib/model, lib/protocol: Use error handling to avoid panic on non-started folder (fixes #6174)

### LINKED ISSUE #6174: Panic loop on startup - handling index for nonexistent folder -  v1.3.2-rc.2
Syncthing auto-updated itself earlier this morning and entered a panic loop during startup on one of my Ubuntu 18.04 LTS hosts.  I have several other Ubuntu 18.04 LTS hosts that all appear to have updated and restarted without issue.

I've attached a sanitized copy of the crash log and the relevant error line in the log is:

```
[A6CDR] 08:34:48 WARNING: Index for nonexistent folder "default"
Panic at 2019-11-21T08:34:48-06:00
panic: handling index for nonexistent folder
```

Syncthing Version: v1.3.2-rc.2
OS Version: Ubuntu 18.04 LTS
Browser Version: n/a

I'll delay attempting to recover for a while if you'd like me to try troubleshooting further.

calmh:
This adds error returns to model methods called by the protocol layer. Returning an error will cause the connection to be torn down as the message couldn't be handled. Using this to signal that a folder isn't
currently available will then cause a reconnection a few moments later, when it'll hopefully work better.

Tested manually by running with STRECHECKDBEVERY=0 on a nontrivially sized setup. This panics reliably before this patch, but just causes a disconnect/reconnect now.

```
[SYNO4] 08:30:57 INFO: Index for nonexistent folder "ourlp-uzomp"
[SYNO4] 08:30:57 INFO: Connection to ... at ... closed: receiver error: no such folder
...
[SYNO4] 08:30:58 INFO: Ready to synchronize "foto/2019" (ourlp-uzomp) (sendreceive)
[SYNO4] 08:30:59 INFO: Completed initial scan of sendreceive folder "foto/2019" (ourlp-uzomp)
...
[SYNO4] 08:31:55 INFO: Established secure connection to (same device)
```

@imsodin This was the least invasive way to fix this I could think of, and I think it might have future benefits like providing a way out when we get a database error for example. If you have better ideas I'm all ears.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1039,17 +1039,17 @@ func (m *model) RemoteNeedFolderFiles(device protocol.DeviceID, folder string, p
 
 // Index is called when a new device is connected and we receive their full index.
 // Implements the protocol.Model interface.
-func (m *model) Index(deviceID protocol.DeviceID, folder string, fs []protocol.FileInfo) {
-	m.handleIndex(deviceID, folder, fs, false)
+func (m *model) Index(deviceID protocol.DeviceID, folder string, fs []protocol.FileInfo) error {
+	return m.handleIndex(deviceID, folder, fs, false)
 }
 
 // IndexUpdate is called for incremental updates to connected devices' indexes.
 // Implements the protocol.Model interface.
-func (m *model) IndexUpdate(deviceID protocol.DeviceID, folder string, fs []protocol.FileInfo) {
-	m.handleIndex(deviceID, folder, fs, true)
+func (m *model) IndexUpdate(deviceID protocol.DeviceID, folder string, fs []protocol.FileInfo) error {
+	return m.handleIndex(deviceID, folder, fs, true)
 }
 
-func (m *model) handleIndex(deviceID protocol.DeviceID, folder string, fs []protocol.FileInfo, update bool) {
+func (m *model) handleIndex(deviceID protocol.DeviceID, folder string, fs []protocol.FileInfo, update bool) error {
 	op := "Index"
 	if update {
 		op += " update"
@@ -1059,10 +1059,10 @@ func (m *model) handleIndex(deviceID protocol.DeviceID, folder string, fs []prot
 
 	if cfg, ok := m.cfg.Folder(folder); !ok || !cfg.SharedWith(deviceID) {
 		l.Infof("%v for unexpected folder ID %q sent from device %q; ensure that the folder exists and that this device is selected under \"Share With\" in the folder configuration.", op, folder, deviceID)
-		return
+		return errFolderMissing
 	} else if cfg.Paused {
 		l.Debugf("%v for paused folder (ID %q) sent from device %q.", op, folder, deviceID)
-		return
+		return errFolderNotRunning
 	}
 
 	m.fmut.RLock()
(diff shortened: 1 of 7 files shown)


lib/protocol/protocol.go
```diff
@@ -388,7 +388,9 @@ func (c *rawConnection) dispatcherLoop() (err error) {
 			if state != stateInitial {
 				return fmt.Errorf("protocol error: cluster config message in state %d", state)
 			}
-			c.receiver.ClusterConfig(c.id, *msg)
+			if err := c.receiver.ClusterConfig(c.id, *msg); err != nil {
+				return fmt.Errorf("receiver error: %v", err)
```
imsodin: Please use errors.Wrap (don't think it matters here, but still). Same below

lib/model/model.go
```diff
@@ -1071,17 +1071,17 @@ func (m *model) handleIndex(deviceID protocol.DeviceID, folder string, fs []prot
...
-		panic("handling index for nonexistent folder")
+		l.Infof("%v for nonexistent folder %q", op, folder)
+		return errFolderMissing
 	}
 
 	if running {
 		defer runner.SchedulePull()
 	} else if update {
 		// Runner may legitimately not be set if this is the "cleanup" Index
 		// message at startup.
```
imsodin: Unrelated: I believe this comment (and the entire if running block) is incorrect: I don't know what "cleanup" Index it is referring to and no matter what it is, a runner must always be set up if `existing == true`.

calmh: I think this is legacy. Back in the day we did `model.Index(someFolder, someDevice, nil)` to drop index data for that device/folder in main.

calmh:
> wouldn't close the connection on request errors

We don't, we convert that error into a response message with the error in it.

lib/protocol/protocol.go
calmh: Wrap'd

lib/model/model.go
calmh: I don't see any such call so I removed that check and simplified

calmh: ... but of course we do so all the time from tests...

imsodin:
Ah crap. The test are failing for the same reason I had some troubles a while ago too: They do add folders but not start them in places. That can't happen in production, but probably is intentional in those tests to test folder functions without actually running a folder service.

calmh:
I resimplified in the other direction a bit... Instead of returning an error on IndexUpdate for non-running, just let it slide. All we need to do is process the update into the database, and we don't *need* the folder to be running for that, we just need it to exist.

lib/model/model.go
```diff
@@ -1059,10 +1059,10 @@ func (m *model) handleIndex(deviceID protocol.DeviceID, folder string, fs []prot
 
 	if cfg, ok := m.cfg.Folder(folder); !ok || !cfg.SharedWith(deviceID) {
 		l.Infof("%v for unexpected folder ID %q sent from device %q; ensure that the folder exists and that this device is selected under \"Share With\" in the folder configuration.", op, folder, deviceID)
-		return
+		return errors.Wrap(errFolderMissing, folder)
```
imsodin: Just a nitpicky opinion, feel free to dismiss silently: The `folder` addition seems unnecessary to me: The caller knows about the folder and thus could wrap the error with that additional info if it thinks it's useful for whatever context the error is used in. Generally I think errors should be wrapped with information pertinent to the error that wouldn't otherwise be available at the receiver of the error.

calmh: I can see the merit of that, yet it seems to me like the parameter is often included in the error, compare `open: some-file-name: not found` returned from os.Open. I'm going to ignore your complaint for the moment and meditate on it for a future "harmonize all the errors" pass :)

PR #6224, opened 2019-12-11
lib/syncthing: Expose backend instead of lowlevel

imsodin:
Instead of having the concrete `db.Lowlevel` type in Syncthing's libraries interface, this proposes to expose the `backend.Backend` interface. That's something a library consumer can actually implement themselves if they don't want to use "our" db backend.

### DIFF
--- a/lib/syncthing/utils.go
+++ b/lib/syncthing/utils.go
@@ -16,7 +16,6 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/syncthing/syncthing/lib/config"
-	"github.com/syncthing/syncthing/lib/db"
 	"github.com/syncthing/syncthing/lib/db/backend"
 	"github.com/syncthing/syncthing/lib/events"
 	"github.com/syncthing/syncthing/lib/fs"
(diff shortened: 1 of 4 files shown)


lib/syncthing/utils.go
```diff
@@ -124,10 +123,6 @@ func copyFile(src, dst string) error {
 	return nil
 }
 
-func OpenGoleveldb(path string, tuning config.Tuning) (*db.Lowlevel, error) {
-	ldb, err := backend.Open(path, backend.Tuning(tuning))
-	if err != nil {
-		return nil, err
-	}
-	return db.NewLowlevel(ldb), nil
+func OpenDBBackend(path string, tuning config.Tuning) (backend.Backend, error) {
```
calmh: This function doesn't add much value. We could do the cast in main...

imsodin: The intention is that all parameters of `syncthing.New` can be created from this package alone, i.e. no need to import `lib/db/backend`. And I am not sure anymore, but I think it might even be a bit more than just convenience: I believe creating reverse bindings for a package puts some restrictions on its exported functions, but not on dependent packages. And afaik I made sure lib/syncthing conforms to those, but not any others.

calmh: Fair 'nuff

PR #6225, opened 2019-12-11
all: Remove config reading through wrapper

imsodin:
### Purpose

This is the large PR I alluded to a few times in the past month: It is a step towards https://github.com/syncthing/syncthing/issues/5298 in that it removes all the methods on `config.Wrapper` to directly access the config. Anyone needing access to the config now needs to subscribe and use the config passed on `CommitConfiguration.

The diff is horrible, as almost anything interacts with config. So I'd suggest to look only at the wrapper changes (loading and subscribing now also returns a config). And if you feel adventurous at the model, as that's as usual where the real party takes place. However in general I guess there's first "higher-level" stuff to be discussed.

### Testing

In addition to tests, I played around with folders on my test setup and had this (without a few master merges) running for about a week in production on my laptop without problems.


lib/syncthing/utils.go
```diff
@@ -42,71 +42,71 @@ func LoadOrGenerateCertificate(certFile, keyFile string) (tls.Certificate, error
...
-	if cfg.RawCopy().OriginalVersion != config.CurrentVersion {
-		if cfg.RawCopy().OriginalVersion == config.CurrentVersion+1101 {
+	if cfg.OriginalVersion != config.CurrentVersion {
+		if cfg.OriginalVersion == config.CurrentVersion+1101 {
 			l.Infof("Now, THAT's what we call a config from the future! Don't worry. As long as you hit that wire with the connecting hook at precisely eighty-eight miles per hour the instant the lightning strikes the tower... everything will be fine.")
 		}
-		if cfg.RawCopy().OriginalVersion > config.CurrentVersion && !allowNewerConfig {
-			return nil, fmt.Errorf("config file version (%d) is newer than supported version (%d). If this is expected, use -allow-newer-config to override.", cfg.RawCopy().OriginalVersion, config.CurrentVersion)
+		if cfg.OriginalVersion > config.CurrentVersion && !allowNewerConfig {
+			return nil, config.Configuration{}, fmt.Errorf("config file version (%d) is newer than supported version (%d). If this is expected, use -allow-newer-config to override.", cfg.OriginalVersion, config.CurrentVersion)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

lib/connections/service.go
```diff
@@ -587,6 +602,15 @@ func (s *service) CommitConfiguration(from, to config.Configuration) bool {
 	}
 
 	s.listenersMut.Lock()
+	// If TrafficClass changes, stop all listeners to get them restarted
+	if from.Options.TrafficClass != to.Options.TrafficClass {
+		for addr := range s.listeners {
+			l.Debugln("Stopping listener", addr)
+			s.listenerSupervisor.Remove(s.listenerTokens[addr])
```
golangcibot: Error return value of `s.listenerSupervisor.Remove` is not checked (from `errcheck`)

lib/model/model.go
```diff
@@ -1946,15 +1961,20 @@ func (m *model) AddConnection(conn connections.Connection, hello protocol.HelloR
...
-	if (device.Name == "" || m.cfg.Options().OverwriteRemoteDevNames) && hello.DeviceName != "" {
+	m.fmut.RLock()
+	overwriteRemoteDevNames := m.options.OverwriteRemoteDevNames
+	m.fmut.RUnlock()
+	if (device.Name == "" || overwriteRemoteDevNames) && hello.DeviceName != "" {
 		device.Name = hello.DeviceName
-		m.cfg.SetDevice(device)
-		m.cfg.Save()
+		w, _ := m.cfgw.SetDevice(device)
+		m.cfgw.Save()
```
golangcibot: Error return value of `m.cfgw.Save` is not checked (from `errcheck`)

lib/model/model_test.go
```diff
@@ -118,7 +118,9 @@ func createTmpWrapper(cfg config.Configuration) config.Wrapper {
 }
 
 func newState(cfg config.Configuration) *model {
-	wcfg := createTmpWrapper(cfg)
+	initCfg := config.New(myID)
+	wcfg := createTmpWrapper(initCfg)
+	wcfg.Replace(cfg)
```
golangcibot: Error return value of `wcfg.Replace` is not checked (from `errcheck`)

lib/model/model_test.go
```diff
@@ -291,26 +293,22 @@ func BenchmarkRequestInSingleFile(b *testing.B) {
...
-	rawCfg := config.New(device1)
-	rawCfg.Devices = []config.DeviceConfiguration{
-		{
-			DeviceID: device1,
-		},
-	}
-	cfg := config.Wrap("testdata/tmpconfig.xml", rawCfg, events.NoopLogger)
+	rawCfg := config.New(myID)
+	cfg := createTmpWrapper(rawCfg)
+	cfg.SetDevice(config.DeviceConfiguration{DeviceID: device1})
```
golangcibot: Error return value of `cfg.SetDevice` is not checked (from `errcheck`)

lib/model/model_test.go
```diff
@@ -321,61 +319,80 @@ func TestDeviceRename(t *testing.T) {
...
 		t.Errorf("Device name not overwritten")
 	}
 }
 
 func TestClusterConfig(t *testing.T) {
-	cfg := config.New(device1)
-	cfg.Devices = []config.DeviceConfiguration{
+	cfg := config.New(myID)
+	wrapper := createTmpWrapper(cfg)
+	wrapper.SetDevices([]config.DeviceConfiguration{
```
golangcibot: Error return value of `wrapper.SetDevices` is not checked (from `errcheck`)

lib/model/model_test.go
```diff
@@ -321,61 +319,80 @@ func TestDeviceRename(t *testing.T) {
...
 			DeviceID:   device1,
 			Introducer: true,
 		},
 		{
 			DeviceID: device2,
 		},
-	}
-	cfg.Folders = []config.FolderConfiguration{
+	})
+	wrapper.SetFolders([]config.FolderConfiguration{
```
golangcibot: Error return value of `wrapper.SetFolders` is not checked (from `errcheck`)

lib/model/model_test.go
```diff
@@ -3402,11 +3574,11 @@ func TestDevicePause(t *testing.T) {
 
 	m.pmut.RLock()
 	closed := m.closed[device1]
+	dev := m.deviceCfgs[device1]
 	m.pmut.RUnlock()
 
-	dev := m.cfg.Devices()[device1]
 	dev.Paused = true
-	m.cfg.SetDevice(dev)
+	m.cfgw.SetDevice(dev)
```
golangcibot: Error return value of `m.cfgw.SetDevice` is not checked (from `errcheck`)

lib/model/testutils_test.go
```diff
@@ -37,14 +37,13 @@ func init() {
...
-	opts.KeepTemporariesH = 1
-	_, _ = defaultCfgWrapper.SetOptions(opts)
+	defaultCfg = config.New(myID)
+	defaultCfgWrapper = createTmpWrapper(defaultCfg)
 
-	defaultCfg = defaultCfgWrapper.RawCopy()
+	defaultCfg.Devices = append(defaultCfg.Devices, config.NewDeviceConfiguration(device1, "device1"))
+	defaultCfg.Folders = append(defaultCfg.Folders, defaultFolderConfig)
+	defaultCfg.Options.KeepTemporariesH = 1
+	defaultCfgWrapper.Replace(defaultCfg)
```
golangcibot: Error return value of `defaultCfgWrapper.Replace` is not checked (from `errcheck`)

lib/syncthing/syncthing.go
```diff
@@ -300,36 +301,36 @@ func (a *App) startup() error {
...
-	if opts := a.cfg.Options(); build.IsCandidate {
+	if build.IsCandidate {
 		l.Infoln("Anonymous usage reporting is always enabled for candidate releases.")
-		if opts.URAccepted != ur.Version {
-			opts.URAccepted = ur.Version
-			a.cfg.SetOptions(opts)
-			a.cfg.Save()
+		if a.cfg.Options.URAccepted != ur.Version {
+			a.cfg.Options.URAccepted = ur.Version
+			a.cfgw.SetOptions(a.cfg.Options)
```
golangcibot: Error return value of `a.cfgw.SetOptions` is not checked (from `errcheck`)

lib/syncthing/syncthing.go
```diff
@@ -300,36 +301,36 @@ func (a *App) startup() error {
...
+			a.cfgw.Save()
```
golangcibot: Error return value of `a.cfgw.Save` is not checked (from `errcheck`)

lib/syncthing/syncthing.go
```diff
@@ -300,36 +301,36 @@ func (a *App) startup() error {
...
 	}
 
 	// If we are going to do usage reporting, ensure we have a valid unique ID.
-	if opts := a.cfg.Options(); opts.URAccepted > 0 && opts.URUniqueID == "" {
-		opts.URUniqueID = rand.String(8)
-		a.cfg.SetOptions(opts)
-		a.cfg.Save()
+	if a.cfg.Options.URAccepted > 0 && a.cfg.Options.URUniqueID == "" {
+		a.cfg.Options.URUniqueID = rand.String(8)
+		a.cfgw.SetOptions(a.cfg.Options)
```
golangcibot: Error return value of `a.cfgw.SetOptions` is not checked (from `errcheck`)

lib/syncthing/syncthing.go
```diff
@@ -300,36 +301,36 @@ func (a *App) startup() error {
...
+		a.cfgw.Save()
```
golangcibot: Error return value of `a.cfgw.Save` is not checked (from `errcheck`)

cmd/syncthing/main.go
```diff
@@ -623,18 +622,18 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
...
 		l.Infoln("Automatic upgrade is always enabled for candidate releases.")
-		if opts := cfg.Options(); opts.AutoUpgradeIntervalH == 0 || opts.AutoUpgradeIntervalH > 24 {
-			opts.AutoUpgradeIntervalH = 12
+		if cfg.Options.AutoUpgradeIntervalH == 0 || cfg.Options.AutoUpgradeIntervalH > 24 {
+			cfg.Options.AutoUpgradeIntervalH = 12
 			// Set the option into the config as well, as the auto upgrade
 			// loop expects to read a valid interval from there.
-			cfg.SetOptions(opts)
-			cfg.Save()
+			cfgw.SetOptions(cfg.Options)
```
golangcibot: Error return value of `cfgw.SetOptions` is not checked (from `errcheck`)

cmd/syncthing/main.go
```diff
@@ -623,18 +622,18 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
...
+			cfgw.Save()
```
golangcibot: Error return value of `cfgw.Save` is not checked (from `errcheck`)

lib/model/model.go
```diff
@@ -2714,20 +2753,21 @@ func (m *model) checkDeviceFolderConnectedLocked(device protocol.DeviceID, folde
...
-// slice of folder configurations.
-func mapFolders(folders []config.FolderConfiguration) map[string]config.FolderConfiguration {
-	m := make(map[string]config.FolderConfiguration, len(folders))
-	for _, cfg := range folders {
-		m[cfg.ID] = cfg
+func (m *model) folderFilesystems() map[string]fs.Filesystem {
+	m.fmut.RLock()
+	defer m.fmut.RUnlock()
+	folderFilesystems := make(map[string]fs.Filesystem, len(m.folderCfgs))
+	for folderID, folderCfg := range m.folderCfgs {
```
golangcibot: unnecessary leading newline (from `whitespace`)

lib/ur/usage_report.go
```diff
@@ -375,32 +380,32 @@ func (s *Service) sendUsageReport() error {
 			DialContext: dialer.DialContext,
 			Proxy:       http.ProxyFromEnvironment,
 			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: s.cfg.Options().URPostInsecurely,
+				InsecureSkipVerify: s.cfg.Options.URPostInsecurely,
```
golangcibot: G402: TLS InsecureSkipVerify may be true. (from `gosec`)

cmd/syncthing/main.go
```diff
@@ -685,15 +684,15 @@ func setupSignalHandling(app *syncthing.App) {
 	}()
 }
 
-func loadOrDefaultConfig(myID protocol.DeviceID, evLogger events.Logger) (config.Wrapper, error) {
+func loadOrDefaultConfig(myID protocol.DeviceID, evLogger events.Logger) (config.Wrapper, config.Configuration, error) {
```
golangcibot: loadOrDefaultConfig - result 0 (github.com/syncthing/syncthing/lib/config.Wrapper) is never used (from `unparam`)

lib/model/model_test.go
```diff
@@ -1312,7 +1440,10 @@ func TestAutoAcceptPausedWhenFolderConfigChanged(t *testing.T) {
 	})
 	m.generateClusterConfig(device1)
 
-	if fcfg, ok := m.cfg.Folder(id); !ok {
+	m.fmut.RLock()
+	fcfg, ok = m.folderCfgs[id]
+	m.fmut.RUnlock()
+	if !ok {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

lib/model/model_test.go
```diff
@@ -1369,7 +1503,10 @@ func TestAutoAcceptPausedWhenFolderConfigNotChanged(t *testing.T) {
...
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

lib/model/model.go
```diff
@@ -1789,10 +1802,7 @@ func (m *model) GetIgnores(folder string) ([]string, []string, error) {
 	m.fmut.RUnlock()
 
 	if !cfgOk {
-		cfg, cfgOk = m.cfg.Folders()[folder]
-		if !cfgOk {
-			return nil, nil, fmt.Errorf("Folder %s does not exist", folder)
-		}
+		return nil, nil, fmt.Errorf("Folder %s does not exist", folder)
```
golangcibot: ST1005: error strings should not be capitalized (from `stylecheck`)

lib/syncthing/syncthing.go
```diff
@@ -300,36 +301,36 @@ func (a *App) startup() error {
...
 	// GUI
 
-	if err := a.setupGUI(m, defaultSub, diskSub, cachedDiscovery, connectionsService, usageReportingSvc, errors, systemLog); err != nil {
+	if err := a.setupGUI(a.cfg.GUI, m, defaultSub, diskSub, cachedDiscovery, connectionsService, usageReportingSvc, errors, systemLog); err != nil {
 		l.Warnln("Failed starting API:", err)
 		return err
 	}
 
-	myDev, _ := a.cfg.Device(a.myID)
+	myDev, _ := a.cfg.DeviceMap[a.myID]
```
golangcibot: S1005: should write `myDev := a.cfg.DeviceMap[a.myID]` instead of `myDev, _ := a.cfg.DeviceMap[a.myID]` (from `gosimple`)

lib/ur/usage_report.go
```diff
@@ -375,32 +380,32 @@ func (s *Service) sendUsageReport() error {
...
 			},
 		},
 	}
-	_, err := client.Post(s.cfg.Options().URURL, "application/json", &b)
+	_, err := client.Post(s.cfg.Options.URURL, "application/json", &b)
```
golangcibot: response body must be closed (from `bodyclose`)

AudriusButkevicius:
So I see the PR but I struggle to understand how this takes us in the right direction.
What is this addressing and how?

calmh:
I think it intends to take us into a less racy world, by having the config be static at all times except when explicitly replaced at certain checkpoints... Which gives me two thoughts regarding the PR;

- There are a couple of `cfg, _, _ := wrapper.Subscribe(someService); ...; someService.CommitConfiguration(..., cfg)` where we first subscribe someone who is interested in config changes and then manually call CommitConfiguration once to kick it off. This is in theory racy as an *actual* configuration commit could happen between those two things happening.

- I see room for a configuration based service wrapper here. There is the `AsService()` which takes and manages a `func(context.Context)`. We could have an alternate `func(context.Context, Configuration)` which gets passed the current configuration at startup time. When the config changes the context is cancelled and the next run uses the newer configuration. In this way at least some of services need less state and handling of config changes on their own...

AudriusButkevicius:
Is that the reason we have most races? I was under the assumption that most races happen because of config changes in quick succession, where the components observe different versions of the config (because nobody calls waiter.Wait()) and logic between them disagrees. I am not sure this solves that (I guess this makes it easier to solve that). 

You still have the "I read config, changed it, and wrote it back tramping someone elses changes" problem, which I suspect is partial cause of the issues we are seeing.



I don't think there is anything here that prevents


I think a fair chunk of components don't really care about having in a consistent state, they care about the latest value, for which the wrapper removes a lot of boiler plate for. Things like NatTimeout, nobody cares if its stale or out of sync or whatever.

imsodin:
Re Audrius: I don't know about most, but it is at least a race in model. We know have replaced some panics with reasonable error handling, but all those connection before folder mixups were exactly because we mixed config gotten through CommitConfiguration and querying the wrapper directly.  
And as you already wrote: This does not address the quick-configs in succession and tramping others changes.

As to the boilerplate and components not caring about consistent config state: It would be possible to keep the wrapper options, but "advise" to not use them if you need a consistent state internally or with other components. However what this means is not really defined and it seems fragile to me. Having a bit of boilerplate seemed like an ok price to pay to remove that distinction. And with Jakobs idea about the service wrapper including config changes quite a of that boilerplate should be removable.

Re Jakobs first point:  
I can readd `Wrapper.RawCopy` with a notice that it should only be used for initial setup. That would also remove the need to return the config itself from `config.Load` and so on.

calmh:
I was thinking maybe an option would be to have `Subscribe()` always call `CommitConfiguration()` with the current config to start things off. The subscriber would get the call with identical `from` and `to` configs. It could compare `to` with whatever is running/cached currently to decide if it needs to act or not. (Probably it's always best for such services to look at a cached copy vs `to` rather than depending on `from` being what is currently running...)

lib/config/wrapper.go
```diff
@@ -496,3 +404,63 @@ func (w *wrapper) AddOrUpdatePendingFolder(id, label string, device protocol.Dev
...
+
+		cfg := <-l.configChan
+		for {
+			select {
+			case <-parentCtx.Done():
+				return
+			default:
+			}
+
+			ctx, cancel := context.WithCancel(parentCtx)
```
golangcibot: lostcancel: the cancel function is not used on all paths (possible context leak) (from `govet`)

lib/config/wrapper.go
```diff
@@ -496,3 +404,63 @@ func (w *wrapper) AddOrUpdatePendingFolder(id, label string, device protocol.Dev
...
+			done := make(chan struct{})
+			go func() {
+				fn(ctx, cfg)
+				close(done)
+			}()
+
+			select {
+			case cfg = <-l.configChan:
+			case <-done:
+				return
```
golangcibot: lostcancel: this return statement may be reached without using the cancel var defined on line 451 (from `govet`)

lib/connections/limiter.go
```diff
@@ -110,33 +110,36 @@ func (lim *limiter) processDevicesConfigurationLocked(from, to config.Configurat
...
 	defer lim.mu.Unlock()
 
 	// Delete, add or update limiters for devices
-	lim.processDevicesConfigurationLocked(from, to)
+	lim.processDevicesConfigurationLocked(to)
 
-	if from.Options.MaxRecvKbps == to.Options.MaxRecvKbps &&
-		from.Options.MaxSendKbps == to.Options.MaxSendKbps &&
-		from.Options.LimitBandwidthInLan == to.Options.LimitBandwidthInLan {
+	if lim.firstConf && lim.opts.MaxRecvKbps == to.Options.MaxRecvKbps &&
```
calmh: `firstConf` is badly named -- it's false for the first config. I stared a long time at this `if` wondering why you would ignore the first conf change only.

lib/api/api.go
```diff
@@ -814,7 +839,7 @@ func (s *service) getDBFile(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *service) getSystemConfig(w http.ResponseWriter, r *http.Request) {
-	sendJSON(w, s.cfg.RawCopy())
+	sendJSON(w, s.cfg)
```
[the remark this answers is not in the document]
imsodin: We only start/stop the server over config changes relevant to the UI, not in general, so I think it's necessary paranoia - so thanks for noticing :)

lib/config/wrapper.go
```diff
@@ -496,3 +404,67 @@ func (w *wrapper) AddOrUpdatePendingFolder(id, label string, device protocol.Dev
...
+
+	cfg := <-s.configChan
+	for {
+		select {
+		case <-s.ctx.Done():
+			return
+		default:
+		}
+
+		ctx, cancel := context.WithCancel(s.ctx)
```
golangcibot: lostcancel: the cancel function is not used on all paths (possible context leak) (from `govet`)

lib/config/wrapper.go
```diff
@@ -496,3 +404,67 @@ func (w *wrapper) AddOrUpdatePendingFolder(id, label string, device protocol.Dev
...
+		done := make(chan struct{})
+		go func() {
+			fn(ctx, cfg)
+			close(done)
+		}()
+
+		select {
+		case cfg = <-s.configChan:
+		case <-done:
+			return
```
golangcibot: lostcancel: this return statement may be reached without using the cancel var defined on line 440 (from `govet`)

imsodin:
Is this something we want to pursue? It's a huge diff and by itself doesn't remove the racyness of config changes, therefore it is a real question of whether it is worth doing the reviewing and having the risk of followup-PRs to fix stuff. The alternative being either leaving the config mostly as it is or coming up with a full "non-racy" rework.  
In my personal (obviously biased) opinion I think it's more realistic to come to that goal in incremental steps than in one large step, thus I would like to pursue this PR. However I'd completely understand if it is deemed not worth the effort at this point (e.g. due to other ongoing work), I just want to know before bringing the branch up-to-date.

st-review:
:robot: *beep boop*

PR #6233, opened 2019-12-19
logger: Fix a typo in a comment

newer027:
Fix for a small typo in a comment.


calmh:
Hi and thanks for the suggestion! Comment/spelling-only pull requests add more churn than they are worth in my opinion, but we're always open for more substantial contributions.

PR #6241, opened 2019-12-27
gui: Missing line break (fixes #6240)

### LINKED ISSUE #6240: GUI: ignore patterns path is one row too low
1) edit folder
2) go to tab 'Ignore Patterns'

actual:
Path is one row too low. It look like it's supposed to be one row up.
Syncthing ignore patterns next line

expected:
Path is one row above current place. 

Version Browser: Mozilla Firefox 71.0 (64-Bit)
Version Syncthing: v1.3.2
Version OS: Windows 8.1 Pro N

imsodin:
For reasons unknown no separation is added here between the last line of the tab body and the separating line to the footer. And wrapping in `<p>` didn't do it either.  
For a screenshot of how it is now see #6240, with this PR well there's a line break before the horizontal line.

### DIFF
--- a/gui/default/syncthing/folder/editFolderModalView.html
+++ b/gui/default/syncthing/folder/editFolderModalView.html
@@ -162,9 +162,8 @@
             <dd><span translate>Comment, when used at the start of a line</span></dd>
           </dl>
           <hr />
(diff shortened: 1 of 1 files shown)


wweich:
The `div`s between the `hr` and the new `br` are set to `pull-left`, which means `float: left;`, so they are outside the normal text flow, which results in the wrong line behavior.

Why are these `div`s floated in the first place?

Removing the `class="pull-left"` should fix the problem (the right way).

PR #6250, opened 2020-01-04
lib/db: Add Badger backend

calmh:
I'm not proposing this as a production change, just putting it out there for discussion. Our leveldb implementation has sometimes worried me in terms of lack of transaction support and in some cases less than responsive maintainership. I wanted to try Badger and see what happens, and I've been running with it myself for a couple of months now. There are no performance advantages that I've been able to measure, on the contrary it's a little bit slower and uses more memory. It has real transactions though.


AudriusButkevicius:
Is this something that should be looked at?

calmh:
I dunno, probably not. I was hoping to present a list of advantages to using badger, but I can't see any obvious ones except the transactionality which is mostly theoretical and the possible more-people-behind-it project stability thing. Doesn't feel like there's a pressing need to change database backends really.

imsodin:
Except maybe for the steady stream of crashes reported to sentry originating in goleveldb or golang code executed by goleveldb.

calmh:
Right. So, if anything, what I think we could do is add something like precisely what this is - just the code as an option behind a magic env var. Interested parties, of which I'm one, can run with this var set to see how it behaves and if there is bad stuff happening. At some point, we can ask for testing from other people, make a decision...

calmh:
Keeping this one warm and merged... 😬

lib/db/backend/badger_backend.go
```diff
@@ -0,0 +1,377 @@
...
+}
+
+func (b *badgerBackend) Put(key, val []byte) error {
+	select {
+	case <-b.closed:
+		return errClosed{}
+	default:
+	}
+
+	txn := b.bdb.NewTransaction(true)
```
imsodin: Are you not using `NewWriteTransaction` due to the additional read transaction created there? `Put` is probably mostly fast enough, but in principle you should also use the `closeWG` stuff here.

lib/db/lowlevel.go
```diff
@@ -311,6 +311,8 @@ func (db *Lowlevel) dropDeviceFolder(device, folder []byte, meta *metadataTracke
 	if err != nil {
 		return err
 	}
+	defer dbi.Release()
```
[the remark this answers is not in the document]
calmh: Yeah, it could. Badger panics when you commit a transaction with an open iterator, leveldb lets it slide, hence the change here since it was discovered here.

lib/db/backend/badger_backend.go
```diff
@@ -0,0 +1,377 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package backend
+
+import (
+	"bytes"
+	"sync"
+
+	badger "github.com/dgraph-io/badger/v2"
```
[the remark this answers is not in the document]
calmh: goimports thing. 🤷‍♂

lib/db/backend/badger_backend.go
```diff
@@ -0,0 +1,377 @@
...
+)
+
+const (
+	checkpointFlushMinSize = 1 << 20
+)
+
+func OpenBadger(path string) (Backend, error) {
+	opts := badger.DefaultOptions(path + ".badger") // temporary path safety
+	opts = opts.WithMaxCacheSize(64 << MiB).WithCompactL0OnClose(false)
+	opts.Logger = nil
```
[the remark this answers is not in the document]
calmh: IIRC it logged some stats and stuff to stdout by default. Until proven otherwise there's nothing I want to hear from the db library that it doesn't tell me via an `error` return so I dropped it. :)

lib/db/backend/badger_backend.go
calmh: Good call on the closeWG

lib/db/backend/badger_backend.go
```diff
@@ -0,0 +1,377 @@
...
+	select {
+	case <-b.closed:
+		return errClosed{}
+	default:
+	}
+
+	// XXX: check if this is appropriate or if we also need Flatten and
+	// whatnot. Also it apparently returns an error if the GC didn't result
+	// in anything being freed...
+	_ = b.bdb.RunValueLogGC(0.5)
```
AudriusButkevicius: Is the return error typed in a way we can check it? Dropping this on the floor sucks.

lib/db/backend/badger_backend.go
```diff
@@ -0,0 +1,377 @@
...
+	})
+}
+
+func (t *badgerTransaction) Put(key, val []byte) error {
+	t.size += len(key) + len(val)
+	kc := make([]byte, len(key))
+	copy(kc, key)
+	vc := make([]byte, len(val))
+	copy(vc, val)
+	return t.transactionRetried(func(txn *badger.Txn) error {
```
AudriusButkevicius: Why do we need these copies?

lib/db/backend/badger_backend.go
```diff
@@ -0,0 +1,377 @@
...
+		return txn.Set(kc, vc)
+	})
+}
+
+// transactionRetried performs the given operation in the current
+// transaction, with commit and retry if Badger says the transaction has
+// grown too large.
+func (t *badgerTransaction) transactionRetried(fn func(*badger.Txn) error) error {
+	if err := fn(t.txn); err == badger.ErrTxnTooBig {
+		if err := t.txn.Commit(); err != nil {
```
AudriusButkevicius: This sort of kills batch atomicity. Can we guide the batch sizes if we know what the max txn size is instead of splitting it?

lib/db/backend/badger_backend.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+}
+
+func (i *badgerIterator) Key() []byte {
+	return i.it.Item().Key()
+}
+
+func (i *badgerIterator) Value() []byte {
+	val, err := i.it.Item().ValueCopy(nil)
+	if err != nil {
```
AudriusButkevicius: Change the db api to reduce garbage generatwd while we're at it?

lib/db/backend/badger_backend.go
```diff
@@ -0,0 +1,377 @@
...
+		panic("unexpected iteration failure: " + err.Error())
+	}
```
AudriusButkevicius: Store the error on the transaction and return in Error() instead of panicing?

lib/db/backend/badger_backend.go
calmh: txn.Set is documented to keep references to the given slices and forbids modification of them until the transaction is committed, which is not the interface we promise in Backend I thing.

lib/db/backend/badger_backend.go
calmh: Yeah this is a wart. I don't think it kicks in until the transaction is many, many megabytes large (way after our normal batch sizes), but it's not documented at all. I'll do some testing.

lib/db/backend/badger_backend.go
calmh: Maybe when/if we deprecate leveldb? It seems premature and excessive to me right now

calmh:
I think this is ready for review. From a user point of view this is triggered by setting `USE_BADGER=1`. Upon opening a database we check if 1) it doesn't exist and 2) the other kind does exist, and if so attempt a copy and then move the old one out of place. This allows for quite seamless switching from one to the other by just twiddling the environment variable.

lib/db/backend/badger_backend.go
imsodin: I don't think we say anything about the values in either `Set` nor `Get` in the interface comments, but I also do think we treat it that way in code. So it should be added to the "backend specs" that after returning the given/returned values should be safe for modifying.

lib/db/backend/badger_backend.go
imsodin: This currently checks transaction size on checkpointing only though - we should also do it on every write/del with a higher limit like on leveldb.

lib/db/backend/badger_backend.go
```diff
@@ -0,0 +1,464 @@
...
+	b.closedMut.RLock()
+	defer b.closedMut.RUnlock()
+	if b.closed {
+		return nil, errClosed{}
+	}
+
+	rel, err := newReleaser(b.closeWG)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
calmh: Yes, and thanks for for pointing out I had forgotten how the closeWG works! Code cleaner now.

lib/db/backend/badger_backend.go
```diff
@@ -0,0 +1,464 @@
...
+	opts := badger.DefaultOptions("").WithInMemory(true)
+	opts.Logger = nil
+	bdb, err := badger.Open(opts)
+	if err != nil {
+		panic(err)
+	}
+	return &badgerBackend{
+		bdb:     bdb,
+		closeWG: &closeWaitGroup{},
+	}
```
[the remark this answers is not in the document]
calmh: ✅

lib/locations/locations.go
```diff
@@ -80,7 +80,20 @@ func GetBaseDir(baseDir BaseDirEnum) string {
...
+	BadgerDirname  = "indexdb.badger"
+)
+
+var databaseDirname = LevelDBDirname
+
+func init() {
+	if os.Getenv("USE_BADGER") != "" {
+		// XXX: HACK
+		databaseDirname = BadgerDirname
+		locationTemplates[Database] = "${data}/" + BadgerDirname
```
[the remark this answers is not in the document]
calmh: It actually didn't, but I'm now not sure why it didn't and it was a bit dirty. Improving.

lib/db/backend/badger_backend.go
calmh: The one thing that comes to mind is that we in many places go to some length to reuse the same key buffer repeatedly in a loop. For values I don't think we do the same, but still the interface for now is clearly that a database operation doesn't take ownership of the passed slices like the Badger operations do.

Then again that key operation is now meaningless with this interface so we could simplify our code elsewhere by always allocating and achieving the same thing... But I don't want to make this change bigger than necessary and Nobody(tm) will be running this code on day one.

lib/db/backend/badger_backend.go
calmh: Well, in LevelDB this is a thing because the "transaction" is an in-memory batch and we don't want to let it grow too much. In Badger it's an actual transaction on disk as far as I can tell, so we don't actually have something in memory that we must limit.

If we call Checkpoint() we're pretty much guaranteed to never hit the transaction-too-large-auto-commit and if we don't, then I don't think it matters?

lib/db/backend/badger_backend.go
```diff
@@ -214,22 +176,33 @@ func (b *badgerBackend) Delete(key []byte) error {
...
+	// Basically, the RunValueLogGC will pick some promising thing to
+	// garbage collect at random and return nil if it improved the
+	// situation, then return ErrNoRewrite when there is nothing more to GC.
+	// The 0.5 is the discard ratio, for which the method docs say they
+	// "recommend setting discardRatio to 0.5, thus indicating that a file
+	// be rewritten if half the space can be discarded".
+	var err error
+	for err == nil {
+		err = b.bdb.RunValueLogGC(0.5)
+	}
```
imsodin, line 195: I am not expecting this to fail, but if badger ever does get into a weird state where it returns `nil` forever, this is going to be a silent complete hangup. Maybe worth adding a timer into the mix and info log every 10 min (or whatever magic number comes to mind) that compaction is still ongoing - that would then at least point at what the problem is. I wouldn't do a hard limit on time, there's some seriously slow devices out there in "normal use".

lib/db/backend/badger_backend.go
```diff
@@ -0,0 +1,443 @@
...
+	rel2, err := newReleaser(b.closeWG)
+	if err != nil {
+		rel1.Release()
+		return nil, err
+	}
+
+	// We use two transactions here to preserve the property that our
+	// leveldb wrapper has, that writes in a transaction are completely
+	// invisible until it's committed, even inside that same transaction.
```
AudriusButkevicius, line 85: Sad life.

lib/db/backend/badger_backend.go
```diff
@@ -0,0 +1,443 @@
...
+	// Basically, the RunValueLogGC will pick some promising thing to
+	// garbage collect at random and return nil if it improved the
+	// situation, then return ErrNoRewrite when there is nothing more to GC.
+	// The 0.5 is the discard ratio, for which the method docs say they
+	// "recommend setting discardRatio to 0.5, thus indicating that a file
+	// be rewritten if half the space can be discarded".
+	var err error
+	t0 := time.Now()
+	for err == nil {
+		if time.Since(t0) > time.Hour {
```
AudriusButkevicius, line 196: Hour?
That's a long time.

I guess we rarely call this so it should not matter?

Also, this would never be printed if it just hangs.

calmh: Yeah... I mean, it can probably hang if the disk hangs, but that goes for all the other database calls as well and is not something we generally worry about. This is more for the really unexpected case of badger changing their api and this looping forever for whatever unforeseen reason. (Or compaction just honestly not able to keep up with our writes for a long time.) Compaction currently just runs after schema updates and gc runs.

lib/db/backend/badger_backend.go
calmh: Well if this becomes a thing in the future we can change our data model to adapt to having dirty reads in transactions instead. (Which is more common, really.)

PR #6261, opened 2020-01-09
lib/events, lib/model: Unflake test and prevent deadlock on event uns…

imsodin:
A test recently flaked: https://build.syncthing.net/viewLog.html?buildId=52372&buildTypeId=Syncthing_BuildLinuxCross. I increased the timeout and instead of querying the model every 10ms (acquiring locks etc etc) to get the folder state listen to events. Also if the event logger is stopped calls to unsubscribe deadlock -> propagate context to subscriptions for to escape that.

### DIFF
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -1549,21 +1549,28 @@ func TestEmptyIgnores(t *testing.T) {
 	}
 }
 
-func waitForState(t *testing.T, m *model, folder, status string) {
+func waitForState(t *testing.T, sub events.Subscription, folder, expected string) {
 	t.Helper()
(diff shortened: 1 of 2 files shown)


lib/model/model_test.go
```diff
@@ -1549,21 +1549,28 @@ func TestEmptyIgnores(t *testing.T) {
 	}
 }
 
-func waitForState(t *testing.T, m *model, folder, status string) {
+func waitForState(t *testing.T, sub events.Subscription, folder, expected string) {
```
golangcibot: `waitForState` - `folder` always receives `"default"` (from `unparam`)

Issue #5207, opened 2018-09-17
vague arror message: "folder marker missing"


nikoss:
The error "folder marker missing" is kind of vague.
Looking around, it means that the " .stfolder " file is missing.

So, why doesn't the error message say so?

> the file ".stfolder" is missing from the folder

nikoss:
Well, then, maybe a link to a specific part of the documentation would be a better option.

Because I had this problem a couple of times in the past... I did googled it... and, always, my (wrong?) response was to recreate the missing file... 

Maybe the links provided by Google said something about not, just, creating the missing file, and I missed it. Or, maybe, it was mentioned in the 50th comment, and I didn't even read it.

Zcehtro:
I've had this problem as well.
Some users have access to hidden files and folders, see the .stfolder marker on the network drive and remove it because they don't know what it is.

Instead of simply getting the error message when this is the case, it would be useful to be pointed to the proper documentation in order to correctly address the issue and resolve it.

Also, usually the user can solve this, but maybe there could be a feature where the WEBUI or GUI offers a dialog box to let the user decide how to fix it but the application handles the fixing procedure? ie. .stfolder marker is missing, so the UI says "If you have removed it, would you like to recreate it? Doing so will ignore all changes and perform an initial folder merge once again." or something similar according to the situation; basically user is always accountable.

This would speed up many accidental messups solve it with a few clicks.

calmh:
> basically user is always accountable

Nonetheless, suggesting a one click option that will delete everything everywhere in a common case (folder not mounted) is unfriendly. 

I’m all for somehow linking to relevant documentation however. Someone just needs to write it.

sohanlalpareek:
https://m.facebook.com/Pareeksonu-544817519015836/?ref_component=mbasic_bookmark&ref_page=%2Fwap%2Fhome.php&refid=7Leave a comment

RichardMyersGV:
I feel that Zechtro's suggestion is not unfriendly at all, but very useful and very friendly.  The "one button" fix clearly explains the implications and allows choices.  Perhaps the "common case (folder not mounted)" can be suggested as a possible problem as well, along with the implications ("delete everything everywhere") of clicking the fix.

Ferroin:
@calmh @RichardMyersGV Might I suggest an alternative approach?

If we have Syncthing actually look at the folder, we can have it behave a bit more sensibly here.  I can think of four cases that are pretty easy to differentiate that matter here:

[...]

PR #6270, opened 2020-01-16
lib/config: Add some info to the folder marker missing (ref #5207)

imsodin:
What should really happen is good documentation on the folder marker, why it's there and what it's disappearance, and then put a link to that into the error message. However I don't have the time for that and don't particularly feel like it either, especially after the insistent, and in my totally personal, potentially entirely unjustified view, quite entitled demand for a "solution. Nevertheless this has been on my todo list so I'd like to at least do a partial solution (and hey, maybe it actually happens for once that someone else than a maintainer steps up to write that particular bit of documentation that is so important (I agree)).

### DIFF
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -23,7 +23,7 @@ import (
 var (
 	ErrPathNotDirectory = errors.New("folder path not a directory")
 	ErrPathMissing      = errors.New("folder path missing")
-	ErrMarkerMissing    = errors.New("folder marker missing")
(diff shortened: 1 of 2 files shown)


imsodin:
Should have followed the PR template and posted screenshots (being honest, I didn't test before): :D 

Screenshot_2020-01-16_15-33-03

Btw: It's supposed to look like that, I just didn't know which folder on my test setup had that issue (but I remember that at least one did), so I unpaused all :)

calmh:
Neat, great that your setup is totally broken and has a large variety of errors to provide :D

PR #6272, opened 2020-01-16
lib/model: Return paused summary instead of error on paused folders

imsodin:
See https://forum.syncthing.net/t/rest-db-status-returns-404-on-startup/14329

If the api starts while the folder is not yet running or generally when querying `/rest/db/status` for a an about to be started folder, we return a 404. Now it is treated the same as a paused folder, i.e. send a valid folder summary. However the comment was wrong: We can't get any info from db for a paused folder, nevertheless it's important to get valid entries with zero values, otherwise the UI blows up.

### DIFF
--- a/lib/model/folder_summary.go
+++ b/lib/model/folder_summary.go
@@ -78,8 +78,8 @@ func (c *folderSummaryService) Summary(folder string) (map[string]interface{}, e
 	var res = make(map[string]interface{})
 
 	errors, err := c.model.FolderErrors(folder)
-	if err != nil && err != ErrFolderPaused {
-		// Stats from the db can still be obtained if the folder is just paused
+	if err != nil && err != ErrFolderPaused && err != errFolderNotRunning {
+		// Stats from the db can still be obtained if the folder is just paused/being started
 		return nil, err
 	}
 	res["errors"] = len(errors)


calmh:
This will leave some other attributes in odd states though, like the "State" field which will become entirely blank because there is no runner and

 https://github.com/syncthing/syncthing/blob/e782bab9fc4e0e7b63b487b5a1014359399a5f1e/lib/model/model.go#L2304-L2316

imsodin:
That's the same thing that happens today when the folder is paused. I agree it's not ideal, but it was the most minimal and consistent (aka leave the API behaviour mostly untouched) fix I came up with. One alternative I thought about was setting the `state` field to paused: That's inconsistent with it's previous use, where it is about the state a running folder is in. Also is it `state`, or `error` and we don't have a `stateChanged` time in either case. Another introducing another boolean field `running` and when the folder is not running, send a summary with just that field set to false. That however breaks our web UI by not sending zero values for the local/global item fields. It's definitely an API change and would require tweaking the UI or defining a minimal empty summary, that gets sent along even if `!running`. (Or even more radical: Don't remove paused folders from the model, only stop them, then the folder summary service still has access to db and can correctly report the state at the time of pausing).

calmh:
Oh, right, already happens... Huh. This is not ideal imho, but a separate issue then.

Issue #4924, opened 2018-05-06
Move index db to $XDG_DATA_HOME/syncthing/


ChristianTacke:
### Feature Request

Syncthing already uses `$XDG_CONFIG_HOME/syncthing/` for its main configuration.

Could you consider moving the indexdb data to `$XDG_DATA_HOME/syncthing/`?
This clearly is "data" and the XDG spec says to put "user-specific data files" there.

Why? indexdb can grow large and people can consider to put them on a specific partition for example or exclude them from daily backup and only incude them in weeklys, etc.

### Version Information

Syncthing Version: 0.14.37+ds1-1~bpo9+1
OS Version: Debian/stable

calmh:
We could, and it might even have made sense from the beginning, but at this point I'm not sure it's worth it. It would make Syncthing behave differently on XDG platforms compared to everywhere else, and also open the question for what even is an "XDG platform" to begin with.

ChristianTacke:
Well, glib has an answer to the "XDG Platform" question for the data dir: https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-get-user-data-dir
Maybe go has something alike?

You could even do something like this:
- If `$XDG_DATA_HOME` is set, let's use it.
- if `$HOME/.local/share/` exists, use it.
- Keep the old behaviour. (it's probably not an "XDG Platform")

calmh:
I could see something like those three bullet points work. I'd add a zero:th one before, "if the database exists in its current location, continue using that" because suddenly moving potentially gigabytes of data would be unexpected.

Apart from that I like the explicitness of depending on the variable being set or the directory existing. I personally think `~/.local/share` is an abomination, but if the user already has it they are presumably fine with it.

ChristianTacke:
That zero'th item sounds fine to me. As you already stated, the alternative would have been to move the indexdb. If both directories are on the same FS, this would be cheap, but we can't know.

Well, `.local/share` sounds a bit strange, but some tools use `.local/lib` for architecture specific data. Having seen that, it starts to make sense.

ChristianTacke:
Thinking about it, I like your zero'th item: It will calm everybody, who wants to keep stuff in one place.

canton7:
Is it worth letting users control the location of the database using a config/env var?

ChristianTacke:
That's the real question here: Should this be done?

- @calmh at least seems to be not totally against this.
[...]

PR #6278, opened 2020-01-19
cmd/syncthing: Always use monitor process (fixes #4774, fixes #5786)

### LINKED ISSUE #4774: Doesn't react to Ctrl-C when run in a subshell with -no-restart (Linux)
When run in a subshell, on Linux, i.e.

```
jb@acro:~ $ cat run.sh 
#!/bin/bash

syncthing -no-browser -no-restart

jb@acro:~ $ ./run.sh 
```

Syncthing won't get and exit due to a Ctrl-C. If the last line is `exec syncthing ...` it's all good as then there is no intermediate shell.

Another workaround is to *not* use `-no-restart`, as the monitor process will then handle this correctly, getting the Ctrl-C and passing the signal to its child. And honestly this is the expected way to run when running "interactively" anyway.

This is due to the process group juggling in c554ffcc, 91210cbb, c2119c9e. The controlling terminal sends the signal to the foreground process group, which in this case is just bash (as we move away into our own process group). 

BSD, Mac, etc don't suffer from this as they have sane thread implementations where we don't need to do tricks with the process group.

A potential "fix" could be to have the monitor process do this instead, by using runtime.LockOSThread, setting the niceness for the current thread, then syscall.ForkExec to spawn the real syncthing from that thread, inheriting the priority. I'm not very fond of moving stuff into the monitor.

Or we could just remove the hack and tell Linux users to set the niceness themselves. Or maybe just let this be a known issue forever.

### LINKED ISSUE #5786: Consider always running the monitor process
The monitor process handles multiple different concerns:
- Restarts, in that the main syncthing process can exit with a certain exit code and get restarted
- Crashes, in that the main syncthing process will get restarted
- Crash reporting, both storing the crash from above and sending it to the mothership when enabled.

If the restart behavior is undesired the monitor process can be disabled by env var or startup switch.

However, we *could* just always use the monitor process for crash tracking and when given one of the flags as above exit afterwards, letting whoever is monitoring us see the restart happening.

imsodin:
### Purpose

Syncthing is now always run through a monitor process. Reasons found in #5786 and it also fixes #4774 as predicted there :)

This will lead to a >20% increase in reports to sentry (according to data.syncthing.net and assuming most apt/arch/fedora/... use some kind of service manager) and an initial spike of old reports, just in case that needs some precautionary measures.

I also took the opportunity to add logging to the desktop entry (also tested to work).

### Testing

I manually tested the following that without `-no-restart` it restarts on ctrl-c in subshell, on restart in UI and on auto-update and it does not restart on shutdown in UI, and with `-no-restart` it doesn't restart in any of the aforementioned cases.

### DIFF
--- a/etc/linux-desktop/syncthing-start.desktop
+++ b/etc/linux-desktop/syncthing-start.desktop
@@ -2,7 +2,7 @@
 Name=Start Syncthing
 GenericName=File synchronization
 Comment=Starts the main syncthing process in the background.
-Exec=/usr/bin/syncthing -no-browser
+Exec=/usr/bin/syncthing -no-browser -logfile=~/.config/syncthing/syncthing.log
 Icon=syncthing
 Terminal=false
 Type=Application
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -47,6 +47,7 @@ import (
 const (
 	tlsDefaultCommonName   = "syncthing"
 	deviceCertLifetimeDays = 20 * 365
+	sigTerm                = syscall.Signal(15)
 )
 
 const (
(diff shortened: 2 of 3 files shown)


etc/linux-desktop/syncthing-start.desktop
```diff
@@ -2,7 +2,7 @@
 Name=Start Syncthing
 GenericName=File synchronization
 Comment=Starts the main syncthing process in the background.
-Exec=/usr/bin/syncthing -no-browser
+Exec=/usr/bin/syncthing -no-browser -logfile=~/.config/syncthing/syncthing.log
```
calmh: You're going to unleash the fucking *fury* of the XDG aficionados here, be prepared to defend yourself :D

calmh:
There's apparently a merge conflict with master which is why the builds didn't happen

calmh:
No, that was something else, looks like the build server was just confused...

etc/linux-desktop/syncthing-start.desktop
imsodin: I know (and I even have more affections than aversion towards XDG myself (I definitely prefer it a thousand times to programs cluttering my home with `.hereIAm` dirs at the other end of the extremes), so I might eventually get around to fixing https://github.com/syncthing/syncthing/issues/4924, including log location) :)

calmh:
I guess you had this cooking for a while and rebased it, commit inherited original time of august 2019, build server thought "oldcrap, not gonna 🤷‍♀" and that was that

imsodin:
I need to cover my tracks better xD  
I made a very half-assed version of this and then got bored by it then. And now when I felt like finally doing it remembered that something existed and rebase-amended it - note to self: Just checkout old stuff into a new branch :D

calmh:
Note that as Audrius pointed out some time you possibly just broke force stopping Syncthing on Android.

Issue #5162, opened 2018-09-04
Create Unix dotfiles as hidden files in Windows


kakra:
I'm syncing between Windows and Linux hosts, and I'd prefer if SyncThing would mark newly created dotfiles as hidden automatically in Windows. If the user later decides to unhide the files, that's okay. The hidden attribute should only be applied upon creation of a file.

Maybe SyncThing could have an option like "hide-on-creation" which I could fill with patterns. It could default to a list like:

```
.*
desktop.ini
```

See also another report #4935 about storing Windows attributes in xattr data to preserve them across different file system architectures.

calmh:
I think this would match how we try to handle execute bits from Windows files etc, so seems reasonable. (The dotfiles part; I'm less convinced about the configurable patterns thing. And the xattrs things is a totally different issue.)

p0larstern:
I would like to work on this issue if this is still open, but I am new to this community and I may need help.

calmh:
Per the discussion on the pull request this is not a desirable feature. Hence, based on feedback, I've decided we should not implement this.

PR #6280, opened 2020-01-20
Sync file attributes (#5162)

AkaZecik:
### Purpose

It's a proposed change for the feature request #5162. I based my work on [Sync file attributes on Windows OS.](https://github.com/syncthing/syncthing/pull/5154)

This code currently synchronizes only the hidden attribute.

I extended the filesystem with two functions: `GetAttributes` and `SetAttributes`. I have seen the discussion regarding the naming conventions, but I decided to use simple names, as changing them to something won't take time.

`SetAttributes` takes two bit-masks, one of them determines which bits should be changed and the other one determines their new values. This function modifies only the attributes present in the former bit-mask, the rest remains unchanged.

Both functions have significance mainly on Windows. On Unix, `GetAttributes` will report whether the file is hidden, but won't do anything when trying to make a file hidden (explanation in a comment).

I also kept the `Hide` and `Unhide` functions, but with different definitions, because it required little to no changes in the rest of the code.

I followed the suggestion that bits for consecutive attributes should be added incrementally and should be defined in the protocol. Unfortunately, `Protocol Buffers` don't support numeric constants or bit fields, so I placed them in `protocol.go` as `uint32`, similarly to `FlagLocal` right above.

I will continue the discussion on forum: https://forum.syncthing.net/t/enforcing-hidden-file-attribute-in-the-windows-world

### Testing

I haven't written tests yet. Also, I've had a hard time setting things up on Windows, so I haven't tested it manually. I would appreciate it if someone would find some spare time to check it out.

### Documentation

Documentation requires changes in the protocol definition here: https://github.com/syncthing/docs/blob/master/specs/bep-v1.rst. It should include information about the new `attributes` field and perhaps the definition for its bits.


lib/fs/basicfs_unix.go
```diff
@@ -52,6 +53,22 @@ func (f *BasicFilesystem) Hide(name string) error {
 	return err
 }
 
+func (f *BasicFilesystem) GetAttributes(name string) (uint32, error) {
+	var f_attrs uint32 = 0
```
golangcibot: don't use underscores in Go names; var `f_attrs` should be `fAttrs` (from `golint`)

lib/fs/basicfs_unix.go
```diff
@@ -52,6 +53,22 @@ func (f *BasicFilesystem) Hide(name string) error {
...
+	// for now, only the 'hidden' attribute is supported
+	if name[0:1] == "." {
+		f_attrs &= protocol.FileAttributeHidden
```
golangcibot: don't use underscores in Go names; var `f_attrs` should be `fAttrs` (from `golint`)

lib/fs/basicfs_unix.go
```diff
@@ -52,6 +53,22 @@ func (f *BasicFilesystem) Hide(name string) error {
...
+	}
+	return f_attrs, nil
+}
+
+func (f *BasicFilesystem) SetAttributes(name string, f_attrs uint32, mask uint32) error {
```
golangcibot: don't use underscores in Go names; method parameter f_attrs should be fAttrs (from `golint`)

lib/fs/errorfs.go
```diff
@@ -7,43 +7,45 @@
...
+func (fs *errorFilesystem) Remove(name string) error                                     { return fs.err }
+func (fs *errorFilesystem) RemoveAll(name string) error                                  { return fs.err }
+func (fs *errorFilesystem) Rename(oldname, newname string) error                         { return fs.err }
+func (fs *errorFilesystem) Stat(name string) (FileInfo, error)                           { return nil, fs.err }
+func (fs *errorFilesystem) SymlinksSupported() bool                                      { return false }
+func (fs *errorFilesystem) Walk(root string, walkFn WalkFunc) error                      { return fs.err }
+func (fs *errorFilesystem) Unhide(name string) error                                     { return fs.err }
+func (fs *errorFilesystem) Hide(name string) error                                       { return fs.err }
+func (fs *errorFilesystem) GetAttributes(name string) (uint32, error)                    { return 0, fs.err }
+func (fs *errorFilesystem) SetAttributes(name string, f_attrs uint32, mask uint32) error { return fs.err }
```
golangcibot: don't use underscores in Go names; method parameter f_attrs should be fAttrs (from `golint`)

lib/fs/fakefs.go
```diff
@@ -547,6 +547,17 @@ func (fs *fakefs) Unhide(name string) error {
 	return nil
 }
 
+
+func (fs *fakefs) GetAttributes(name string) (uint32, error) {
+    // noop for now
+    return 0, nil
+}
+
+func (fs *fakefs) SetAttributes(name string, f_attrs uint32, mask uint32) error {
```
golangcibot: don't use underscores in Go names; method parameter f_attrs should be fAttrs (from `golint`)

lib/fs/filesystem.go
```diff
@@ -42,6 +42,8 @@ type Filesystem interface {
 	Watch(path string, ignore Matcher, ctx context.Context, ignorePerms bool) (<-chan Event, <-chan error, error)
 	Hide(name string) error
 	Unhide(name string) error
+	GetAttributes(name string) (uint32, error)
+	SetAttributes(name string, f_attrs uint32, mask uint32) error
```
golangcibot: don't use underscores in Go names; interface method parameter f_attrs should be fAttrs (from `golint`)

calmh:
I wonder how many actual attributes are relevant to Syncthing in the long run, and hence whether it would be better to just run with a `bool hidden` in the protocol for now? Not super excited about the attributes-and-mask API for Filesystem, it feels rather too low level and implementation defined...

AkaZecik:
@calmh I applied your suggestions. Let me know if it is a good place to go from and what else you would want to see implemented.

lib/protocol/bep.proto
```diff
@@ -122,6 +122,7 @@ message FileInfo {
     bool deleted        = 6;
     bool invalid        = 7 [(gogoproto.customname) = "RawInvalid"];
     bool no_permissions = 8;
+    bool hidden         = 18;
```
AudriusButkevicius: Why 18?

lib/fs/filesystem.go
```diff
@@ -42,6 +42,7 @@ type Filesystem interface {
 	Watch(path string, ignore Matcher, ctx context.Context, ignorePerms bool) (<-chan Event, <-chan error, error)
 	Hide(name string) error
 	Unhide(name string) error
+	IsHidden(name string) (bool, error)
```
AudriusButkevicius: I think this should not be a property of the filesystem but a property of the FileInfo.

AudriusButkevicius: Same like a file has a name, size, permissions, it has a hidden property.

lib/fs/basicfs_unix.go
```diff
@@ -52,6 +52,15 @@ func (f *BasicFilesystem) Hide(name string) error {
 	return err
 }
 
+func (f *BasicFilesystem) IsHidden(name string) (bool, error) {
+	fileInfo, err := f.Stat(name)
+	if err != nil {
+		return false, err
+	}
+	name = fileInfo.Name()
+	return name[0:1] == ".", nil
```
AudriusButkevicius: name[0] == '.'

lib/db/structs.proto
```diff
@@ -45,6 +45,7 @@ message FileInfoTruncated {
     bool deleted        = 6;
     bool invalid        = 7 [(gogoproto.customname) = "RawInvalid"];
     bool no_permissions = 8;
+    bool hidden         = 18;
```
AudriusButkevicius: Why 18

lib/fs/basicfs_unix.go
AkaZecik: I thought that name[0:1] wouldn't panic if name was an empty string. I must have misstyped something before, because now I checked and it panics. I will change it to yours, thanks.

lib/protocol/bep.proto
AkaZecik: I didn't notice that 14 and 15 were available. I can change that to either. I've read that field numbers below 16 take only one byte so it might actually be a good idea.

lib/fs/basicfs_unix.go
imsodin: name[0] panics too for an empty string. And you also need to check for `name == "."`, which is not hidden.

lib/fs/filesystem.go
imsodin: Yep. Add it to `FileIntf` and then the types that fulfill that (protocol fileinfo and db truncated file info).

AkaZecik: From what I see `BasicFileInfo` doesn't provide any direct access to methods which could allow to query for attributes. In that case, `IsHidden` would need to take an argument with some filesystem. Other solution would be to extend `BasicFileInfo` with a field 'hidden' and everywhere it is constructed we would need to query for this attribute; from what I found it happens only when returning from `(BasicFileSystem).Lstat()`, `(basicFile).Lstat` and `(BasicFileSystem).Stat()`.

Is any of those options what you meant?

AkaZecik: Oh, it seems you meant a different `FileInfo`... :)

lib/protocol/bep.proto
calmh: The fields are now ordered to create a reasonable struct definition, which blurs the allocation order. But I think that for this reason we used lower numbered fields first, and had blocks + seldom used ones at 16+ for this reason. I think hidden works fine at 18 because it's almost never going to be set and even then just becomes two bytes.

lib/fs/filesystem.go
calmh: Seems like it would fit among the other stat results like size and permissions etc, yes.

lib/protocol/bep_extensions.go
```diff
@@ -213,6 +218,10 @@ func (f FileInfo) isEquivalent(other FileInfo, modTimeWindow time.Duration, igno
 		return false
 	}
 
+	if f.Hidden != other.Hidden {
```
imsodin: I haven't thought it through, but this might be problematic: A file that's hidden on windows but doesn't start with a dot will never be hidden on linux. That might just undo any hidden-ness on windows or even result in some nasty pull-scan loop.

imsodin: I think we need to take a step back and look at what exactly should be achieved here:

As is a file named 'foo' hidden on windows, that is pulled by a linux machine will be created, and then once scanning will see it's not hidden (no dot), and thus revert the hidden status. We can circumvent that by adding a check for windows in this line to prevent scanning on other systems to revert the hidden status. However that still leaves the problem that if the linux machine changes the file, it will then become unhidden again. That could be mitigated by adding yet another field to ignore the hidden status. Or make the hidden field an uint instead of bool, allowing for a third "undecided" state. However that feels like too much complexity for the gain and a "best effort" solution might make more sense: Create files with dot as hidden on windows, pick up the hidden attribute and send it as an optional parameter, that gets ignored (and dropped) by anyone but windowses and on windowses, preserve the hidden status when pulling a file without a hidden attribute.

calmh:
This is problematic in that it is an attribut that is not representable on Unix. The original ask is easier: dotfiles should get the hidden flag on Windows. That's just a last step in the puller. The rest requires consideration...

tomasz1986:
> This code currently synchronizes only the hidden attribute.

I know that this is currently only about one attribute, but because of the title mentioning "just" file attributes, I just want to ask not to sync the other attributes, such as encrypted (which would be a disaster) or compressed. The latter is very situational, i.e. I may want to compress my files on a fast desktop PC, but may not want to do that on a slow Intel Atom laptop.

calmh:
I think this is probably misguided as-is, for the reasons imsodin outlines.

PR #6283, opened 2020-01-21
lib/db: Deduplicate block lists in database (fixes #5898)

### LINKED ISSUE #5898: Database performance: reduce duplication
The database becomes quite large in large setups and this reduces performance, due to writes and compaction. Currently we keep a copy of each file info for each peer. This is simple but wasteful, especially for the block list. There are attributes we need to remember per device (the invalid bits, for example) but it should be enough to keep one copy of the block list, keyed by folder/file name/version.

calmh:
### Purpose

This moves the block list in the database out from being just a field on
the FileInfo to being an object of its own. When putting a FileInfo we
marshal the block list separately and store it keyed by the sha256 of
the marshalled block list. When getting, if we are not doing a
"truncated" get, we do an extra read and unmarshal for the block list.

Old block lists are cleared out by a periodic GC sweep. The alternative
would be to use refcounting, but:

- There is a larger risk of getting that wrong and either dropping a
  block list in error or keeping them around forever.

- It's tricky with our current database, as we don't have dirty reads.
  This means that if we update two FileInfos with identical block lists in
  the same transaction we can't just do read/modify/write for the ref
  counters as we wouldn't see our own first update. See above about
  tracking this and risks about getting it wrong.

GC uses a bloom filter for keys to avoid heavy RAM usage. GC can't run
concurrently with FileInfo updates so there is a new lock around those
operation at the lowlevel.

The end result is a much more compact database, especially for setups
with many peers where files get duplicated many times.

This is per-key-class stats for a large database I'm currently working
with, under the current schema:

```
 0x00:  9138161 items, 870876 KB keys + 7397482 KB data, 95 B +  809 B avg, 1637651 B max
 0x01:   185656 items,  10388 KB keys + 1790909 KB data, 55 B + 9646 B avg,  924525 B max
 0x02:   916890 items,  84795 KB keys +    3667 KB data, 92 B +    4 B avg,     192 B max
 0x03:      384 items,     27 KB keys +       5 KB data, 72 B +   15 B avg,      87 B max
 0x04:     1109 items,     17 KB keys +      17 KB data, 15 B +   15 B avg,      69 B max
 0x06:      383 items,      3 KB keys +       0 KB data,  9 B +    2 B avg,      18 B max
 0x07:      510 items,      4 KB keys +      12 KB data,  9 B +   24 B avg,      41 B max
 0x08:     1349 items,     12 KB keys +      10 KB data,  9 B +    8 B avg,      17 B max
 0x09:      194 items,      0 KB keys +     123 KB data,  5 B +  634 B avg,   11484 B max
 0x0a:        3 items,      0 KB keys +       0 KB data, 14 B +    7 B avg,      30 B max
 0x0b:   181836 items,   2363 KB keys +   10694 KB data, 13 B +   58 B avg,     173 B max
 Total 10426475 items, 968490 KB keys + 9202925 KB data.
```

Note 7.4 GB of data in class 00, total size 9.2 GB. After running the
migration we get this instead:

```
 0x00:  9138161 items, 870876 KB keys + 2611392 KB data, 95 B +  285 B avg,    4788 B max
 0x01:   185656 items,  10388 KB keys + 1790909 KB data, 55 B + 9646 B avg,  924525 B max
 0x02:   916890 items,  84795 KB keys +    3667 KB data, 92 B +    4 B avg,     192 B max
 0x03:      384 items,     27 KB keys +       5 KB data, 72 B +   15 B avg,      87 B max
 0x04:     1109 items,     17 KB keys +      17 KB data, 15 B +   15 B avg,      69 B max
 0x06:      383 items,      3 KB keys +       0 KB data,  9 B +    2 B avg,      18 B max
 0x07:      510 items,      4 KB keys +      12 KB data,  9 B +   24 B avg,      41 B max
 0x09:      194 items,      0 KB keys +     123 KB data,  5 B +  634 B avg,   11484 B max
 0x0a:        3 items,      0 KB keys +       0 KB data, 14 B +   17 B avg,      51 B max
 0x0b:   181836 items,   2363 KB keys +   10694 KB data, 13 B +   58 B avg,     173 B max
 0x0d:    44282 items,   1461 KB keys +   61081 KB data, 33 B + 1379 B avg, 1637399 B max
 Total 10469408 items, 969939 KB keys + 4477905 KB data.
```

Class 00 is now down to 2.6 GB, with just 61 MB added in class 0d.

There will be some additional reads in some cases which theoretically
hurts performance, but this will be more than compensated for by smaller
writes and better compaction.

On my own home setup which just has three devices and a handful of
folders the difference is smaller in absolute numbers of course, but
still less than half the old size:

```
 0x00:  297122 items,  20894 KB keys + 306860 KB data, 70 B + 1032 B avg, 103237 B max
 0x01:  115299 items,   7738 KB keys +  17542 KB data, 67 B +  152 B avg,    419 B max
 0x02: 1430537 items, 121223 KB keys +   5722 KB data, 84 B +    4 B avg,    253 B max
 ...
 Total 1947412 items, 151268 KB keys + 337485 KB data.
```

to:

```
 0x00:  297122 items,  20894 KB keys +  37038 KB data, 70 B +  124 B avg,    520 B max
 0x01:  115299 items,   7738 KB keys +  17542 KB data, 67 B +  152 B avg,    419 B max
 0x02: 1430537 items, 121223 KB keys +   5722 KB data, 84 B +    4 B avg,    253 B max
 ...
 0x0d:   18041 items,    595 KB keys +  71964 KB data, 33 B + 3988 B avg, 101109 B max
 Total 1965447 items, 151863 KB keys + 139628 KB data.
```

### Testing

I'm running this at home now. Stay tuned...

### One more thing

With this, the old `TruncatedFileInfo` and `FileIntf` are essentially useless. But they are infectious and the diff to remove them is large (I started on it). I'm intentionally leaving that for a separate refactor-only commit in the future.


lib/db/lowlevel.go
```diff
@@ -824,16 +861,96 @@ func (db *Lowlevel) dropPrefix(prefix []byte) error {
...
+	it, err = db.NewPrefixIterator([]byte{KeyTypeBlockList})
+	if err != nil {
+		return err
+	}
+	matched := 0
+	for it.Next() {
+		if filter.Test(it.Key()) {
+			matched++
+		} else {
+			t.Delete(it.Key())
```
golangcibot: Error return value of `t.Delete` is not checked (from `errcheck`)

lib/db/schemaupdater.go
```diff
@@ -107,10 +99,11 @@ func (db *schemaUpdater) updateSchema() error {
 		return err
 	}
 
-	return nil
+	l.Infoln("Compacting database after migration...")
+	return db.Compact()
 }
 
-func (db *schemaUpdater) updateSchema0to1() error {
+func (db *schemaUpdater) updateSchema0to1(_ int) error {
```
imsodin: Nit: Given the new argument the previous version in the name isn't really correct anymore.

lib/db/transactions.go
```diff
@@ -58,13 +59,50 @@ func (t readOnlyTransaction) getFileTrunc(key []byte, trunc bool) (FileIntf, boo
...
+	bs, err := t.Get(fi.BlockListKey)
+	if err != nil {
+		return err
+	}
+	var bl BlockList
+	if err := bl.Unmarshal(bs); err != nil {
+		return err
+	}
+	fi.Blocks = bl.Blocks
+	fi.BlockListKey = nil
```
imsodin: What the motivation of "niling" `Blocks` and `BlockListKey` in the appropriate place instead of having `protocol.FileInfo` with only the former and another type to be stored with only the latter, preventing any misshaps like forgetting a switch?

cmd/stindex/accounting.go
```diff
@@ -0,0 +1,58 @@
...
+		if s > max[t] {
+			max[t] = s
+		}
+	}
+
+	tw := tabwriter.NewWriter(os.Stdout, 1, 1, 1, ' ', tabwriter.AlignRight)
+	toti, totds, totks := 0, 0, 0
+	for t := range ksizes {
+		if ksizes[t] > 0 {
+			// yes metric kilobytes 🤘
```
imsodin:
```
        ,--.!,
     __/   -*-
   ,d08b.  '|`
   0088MM     
   `9MMP'
UNI

```

lib/db/lowlevel.go
```diff
@@ -44,6 +69,9 @@ func (db *Lowlevel) ListFolders() []string {
 // updateRemoteFiles adds a list of fileinfos to the database and updates the
 // global versionlist and metadata.
 func (db *Lowlevel) updateRemoteFiles(folder, device []byte, fs []protocol.FileInfo, meta *metadataTracker) error {
+	db.gcMut.RLock()
```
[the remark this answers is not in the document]
calmh: Good idea, I’ll try that. Technically not all write transactions are incompatible with the gc, but it shouldn’t be running long and often enough for that to be an issue so safer is better.

lib/db/transactions.go
calmh: I was thinking this was simpler and avoids having to copy a bunch of fields to a new type and keep that in sync. As noted I want to remove the *truncated type completely at some point...

calmh: An alternative could be to have a database-only message type that embeds a fileinfo and has a custom marshaller that nils out the blocks and such. Don’t know if that’s a thing with protobuf, can check.

lib/db/structs.proto
```diff
@@ -40,13 +40,24 @@ message FileInfoTruncated {
     int32                 block_size     = 13 [(gogoproto.customname) = "RawBlockSize"];
 
     // see bep.proto
-    uint32 local_flags = 1000;
+    uint32   local_flags    = 1000;
+    // bytes block_list_key = 1001;
```
AudriusButkevicius: Undo diff?

lib/db/structs.proto
```diff
@@ -40,13 +40,24 @@ message FileInfoTruncated {
...
     bool invalid        = 7 [(gogoproto.customname) = "RawInvalid"];
     bool no_permissions = 8;
 }
 
+// BlockList is the structure used to store block lists
+message BlockList {
+    repeated protocol.BlockInfo Blocks = 1 [(gogoproto.nullable) = false];
+}
+
+// BlockListKeyOnly is used to unly unmarshal the block list key from a FileInfo
```
AudriusButkevicius: Unly

lib/model/model.go
```diff
@@ -1967,7 +1968,8 @@ func (s *indexSender) sendIndexTo(ctx context.Context) error {
 		if f.IsReceiveOnlyChanged() {
 			f.Version = protocol.Vector{}
 		}
-		f.LocalFlags = 0 // never sent externally
+		f.LocalFlags = 0     // never sent externally
+		f.BlockListKey = nil // never sent externally
```
AudriusButkevicius: I wonder if we should just send this, and potentially reuse it for blockfinder/renames by having yet another indirection index?

Also, allows to do equivalence checks without having to load the blocklist, allowing to minimise db throughput in some cases.

calmh: We could do that, I've many times thought about having a whole-file hash available for quicker comparisons. If so, it should probably be stricter specified than just "sha256 of whatever the protobuf marshaller happened to produce", something like sha256 of all the hashes and sizes sequentially or such. And we could produce it once and for all at scan time, so it wouldn't matter if it was slightly more expensive to produce...

calmh: I made this a real field that is the sha of all the block hashes. Using it to optimize other things is a future thing.

lib/db/lowlevel.go
```diff
@@ -9,9 +9,23 @@ package db
...
+
+const (
+	// We set the bloom filter capacity to handle 100k individual block lists
+	// with a false positive probability of 1% for the first pass. Once we know
+	// how many block lists we have we will use that number instead, if it's
+	// more than 100k. For fewer than 100k block lists we will just get better
+	// false positive rate instead.
+	blockGCBloomCapacity          = 100000
+	blockGCBloomFalsePositiveRate = 0.01 // 1%
+	blockGCInterval               = 13 * time.Hour
```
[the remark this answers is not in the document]
calmh: The GC thing is a full walk of the database, so can take potentially minutes during which a lock is held. It's not important that it happens immediately so I don't want to block startup (or even sync stuff that would otherwise happen at startup) for it. I picked 13 hours as something that felt often enough yet hopefully not impacting and doesn't divide 24 hours so that it should slide to different times of day. That avoids it always happening during peak hours for example, if you have peak hours.

If we were to make this configurable I think it should be an env var as it's not even interesting enough to be an advanced config for almost all users, I think...

go.mod
```diff
@@ -35,11 +35,14 @@ require (
...
 	github.com/sasha-s/go-deadlock v0.2.0
 	github.com/shirou/gopsutil v0.0.0-20190714054239-47ef3260b6bf
+	github.com/spaolacci/murmur3 v1.1.0 // indirect
 	github.com/syncthing/notify v0.0.0-20190709140112-69c7a957d3e2
 	github.com/syndtr/goleveldb v1.0.1-0.20190923125748-758128399b1d
 	github.com/thejerf/suture v3.0.2+incompatible
 	github.com/urfave/cli v1.22.2
 	github.com/vitrun/qart v0.0.0-20160531060029-bf64b92db6b0
+	github.com/willf/bitset v1.1.10 // indirect
+	github.com/willf/bloom v2.0.3+incompatible
```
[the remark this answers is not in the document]
calmh: I looked at it now and it's an awkward API for us, it's designed to create filters that are serialized to a buffer and then saved to the database, and can be queried after the fact. I think it's fine to bring in a package for the purpose.

lib/db/lowlevel.go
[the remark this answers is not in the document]
calmh: So thinking more about it I'm a bit wary of this. There's nothing that currently says it's forbidden to open a new `readWriteTransaction` while already in one, but if we add an `RLock` to it that will be a deadlocking operation if it happens concurrently with GC. As it is, taking the `RLock` explicitly only at the top of the `update...Files` and drop operations is clear this won't be a problem.

lib/db/schemaupdater.go
calmh: Mmm. Only added to support the weirdness in the to-v5 migrations... But I guess since I'm anyway touching the function signatures I might as well remove the "from" part in the name.

lib/db/structs.proto
calmh: ✅

lib/db/structs.proto
calmh: ✔️

lib/db/schemaupdater.go
```diff
@@ -68,35 +69,26 @@ func (db *schemaUpdater) updateSchema() error {
...
-		}
+	type migration struct {
+		schemaVersion int64
+		migration     func(prevVersion int) error
 	}
-	if prevVersion < 6 {
-		if err := db.updateSchema5to6(); err != nil {
-			return err
-		}
+	var migrations = []migration{
```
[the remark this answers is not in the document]
calmh: Yeah... Wasn't sure which was the prettier.

cmd/stindex/accounting.go
calmh: Makes it easier to mentally scale down further to gigabytes etc :)

lib/db/transactions.go
calmh: Since the block list anyhow requires special handling (marshalling separately, checking if it's already stored, ...) I don't this would help not forgetting. As is it is hidden away in t.putFile and t.unmarshalTrunc which is where we've historically done the similar thing.

calmh:
💥

Issue #6226, opened 2019-12-12
Synchting using much memory (again)


Helium314:
I just noticed that syncthing was using 1.2 GiB memory on my Raspberry Pi 3 (1 GiB available, + ~1 GiB swap on SD card), which slowed down things a lot.
This happened while uploading 2 files (ca 150 MiB each) to the only other remote device that was online. It looks a bit similar to #6160, but after upgrading to 1.3.2 this did not happen again despite transferring larger files several times.

After restarting syncthing the transfers continued normally, with memory usage staying below 200 MiB... so unfortunately I can't reproduce this issue.

Anyway I got a [memory profile](https://github.com/syncthing/syncthing/files/3957038/syncthing-heap-linux-arm-v1.3.2-171438.pprof.gz) and a [screenshot of the web GUI](https://user-images.githubusercontent.com/43007630/70734327-5ae52080-1d0c-11ea-969b-dd479311eee5.png) from the problem, I hope this helps.
Database tuning is set to small, I don't remember touching any other "advanced" option.

AudriusButkevicius:
It did not happen once you upgraded, and the upgrade included the fix, so it's somewhat expected as it's fixed?

Which version is the profile from, and is it during the period of high memory usage or just some random period?

I feel we should close this and move it to the forum until it's clear what exactly we are talking about here.

Helium314:
#6160 did not happen again since the upgrade. I think. Because #6160 was easily reproducible, while I noticed this problem only once. Still it looks very similar in the effects I see: high memory usage while uploading.

This happened just today on 1.3.2 for 32 bit ARM Linux. The profile was taken during high memory usage, maybe a few minutes after the screenshot (slow Raspi takes time to react).

imsodin:
The profile accounts ~500MB most of which is used by quic (with packet handling using 270MB) - which seems like too much. Looks like it's unrelated to requests/syncing. Without a reproducer it's probably not possible to debug this.

AudriusButkevicius:
We should probably raise it with them indicating which objects are causing this

calmh:
Haven't looked at anything yet, but keep in mind we're lagging a couple of QUIC versions since they've dropped Go 1.12 compatibility and we haven't.

Helium314:
[...]

PR #6286, opened 2020-01-21
build: Update quic keeping go1.12 compat (ref #6226)

imsodin:
Yeiii go modules :)  

quic-go requires go1.13, but the only reason is crypto/ed25519, which is the same as golang.org/x/crypto/ed25519 and that's available on <go1.13 too. Even if go1.14 will be out soon and we might drop go1.12, I feel it's useless to throw compatibility out of the window without need.

QUIC got some updates in the code that might be relevant to heap mem accumulation recorded in #6226 and maybe even fix it.


calmh:
woah.gif

Although it looks like 1.12 doesn't work that may, maybe buggy module support, so you're probably fucked anyway...

calmh:
Go 1.14 is soon here and then we can drop 1.12 with a clear conscience.

PR #6288, opened 2020-01-22
[WIP] View included files

nitrajka:
After a long semester, I finally found some time to finish the code, whether you decide to merge it or not.

Reference to #6108  and issue https://github.com/syncthing/syncthing/issues/4729 .

@AudriusButkevicius 
I reprogrammed the route to return the array-like structure as you suggested in the mentioned PR. I also did the AngularJS part. However, I simplified it a little bit so I don't spend a lifetime programming it.
It looks like this now:

Screen Shot 2020-01-22 at 16 05 05
Screen Shot 2020-01-22 at 16 05 16
The blue directories are clickable. "Back" is not present in root directory.
Screen Shot 2020-01-22 at 16 05 25
You can go back to root directory by clicking on "Back".

Currently the route returns this kind of result:

```
[
  {
    "name": ".DS_Store",
    "isDirectory": false,
    "children": []
  },
  {
    "name": "file-0",
    "isDirectory": false,
    "children": []
  },
  {
    "name": "file-1",
    "isDirectory": false,
    "children": []
  },
  {
    "name": "file-2",
    "isDirectory": false,
    "children": []
  },
  {
    "name": "file-3",
    "isDirectory": false,
    "children": []
  },
  {
    "name": "file-4",
    "isDirectory": false,
    "children": []
  },
  {
    "name": "file-5",
    "isDirectory": false,
    "children": []
  },
  {
    "name": "neh",
    "isDirectory": true,
    "children": []
  },
  {
    "name": "neh2",
    "isDirectory": true,
    "children": []
  },
  {
    "name": "neh3",
    "isDirectory": true,
    "children": []
  },
  {
    "name": "turaa.jpg",
    "isDirectory": false,
    "children": []
  }
]
```

There are still things I would like to ask about before I decide to finish them somehow.

1. There is a test suite in lib/model/model_test.go which says, that route "some/directory" doesn't have to come at all, but "some/directory/file" comes.
https://github.com/nitrajka/syncthing/blob/view-included-files/lib/model/model_test.go#L1943
 In this case, I have to create the whole path from one route. I didn't know that when programming the back end. Should I fix it?

2. If you would consider merging it, I should add the documentation. That is the purpose of syncthing/docs repository, right? So, what do you think?

3. As I am creating the route when nesting in the newly programmed file browser, I use the linux-based path delimeter: 

https://github.com/nitrajka/syncthing/blob/view-included-files/gui/default/syncthing/folder/folderFilesModalView.html#L9

That may be a problem for windows based file systems. Do you have any hint on how should I fix it?

4. Should I add translations on https://www.transifex.com/projects/p/syncthing/ ?

5. The original route returned a structure where a file had some metadata. My version doesn't have any metadata. Is that a problem?

6. Every time I run compiled syncthing I use "STNOUPGRADE=1" flag. But we want the upgraded version to have file browser as well. Do we?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,8 @@ require (
 	github.com/maruel/panicparse v1.3.0
 	github.com/mattn/go-isatty v0.0.11
 	github.com/minio/sha256-simd v0.1.1
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+	github.com/modern-go/reflect2 v1.0.1 // indirect
 	github.com/onsi/ginkgo v1.9.0 // indirect
 	github.com/onsi/gomega v1.6.0 // indirect
 	github.com/oschwald/geoip2-golang v1.4.0
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -206,6 +206,11 @@ table.table-auto td {
     margin-bottom: 0.25em;
 }
 
+#folderFiles .modal-body {
+    height: 300px;
+    overflow-y: scroll;
+}
+
 button.panel-heading {
     display: block;
     position: relative;
@@ -414,4 +419,12 @@ ul.three-columns li, ul.two-columns li {
     .btn:not(.panel-heading) + .btn:not(.panel-heading) {
         margin-bottom: 1rem;
     }
+
+    #folderFiles .modal-header {
+        min-height: 74px;
+    }
+}
+
+#folderFiles .modal-header {
+    min-height: 64px;
 }
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -495,7 +495,7 @@
                       <th><span class="fas fa-fw fa-copy"></span>&nbsp;<span translate>Shared files</span></th>
                       <td class="text-right">
                         <span>
-                          <a href="" data-toggle="modal" data-target="#folderFiles" ng-click="setCurrentFolder(folder.id)">show</a>
+                          <a href="" data-toggle="modal" data-target="#folderFiles" ng-click="setCurrentFolderAndGetContents(folder.id)">show</a>
                         </span>
                       </td>
                     </tr>
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -495,7 +495,8 @@
                       <th><span class="fas fa-fw fa-copy"></span>&nbsp;<span translate>Shared files</span></th>
                       <td class="text-right">
                         <span>
-                          <a href="" data-toggle="modal" data-target="#folderFiles" ng-click="setCurrentFolderAndGetContents(folder.id)">show</a>
+                          <a href="#" data-toggle="modal" data-target="#folderFiles" ng-click="setCurrentFolderAndGetContents(folder.id, '')">show</a>
+                        <a href="#" data-toggle="modal" data-target="#folderFiles" ng-click="setCurrentFolderAndGetContents(folder.id, 'neh2')">show neh2</a>
                         </span>
                       </td>
                     </tr>
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -492,7 +492,7 @@
                       </td>
                     </tr>
                     <tr>
-                      <th><span class="fas fa-fw fa-copy"></span>&nbsp;<span translate>Shared files</span></th>
+                      <th><span class="fas fa-fw fa-copy"></span>&nbsp;<span translate>Included files</span></th>
                       <td class="text-right">
                         <span>
                           <a href="#" data-toggle="modal" data-target="#folderFiles" ng-click="setCurrentFolderAndGetContents(folder.id, '')">show</a>
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -491,6 +491,14 @@
                         </span>
                       </td>
                     </tr>
+                    <tr>
+                      <th><span class="fas fa-fw fa-copy"></span>&nbsp;<span translate>Shared files</span></th>
+                      <td class="text-right">
+                        <span>
+                          <a href="" data-toggle="modal" data-target="#folderFiles" ng-click="setCurrentFolder(folder.id)">show</a>
+                        </span>
+                      </td>
+                    </tr>
                   </tbody>
                 </table>
               </div>
@@ -832,6 +840,7 @@
   <ng-include src="'syncthing/folder/editFolderModalView.html'"></ng-include>
   <ng-include src="'syncthing/folder/restoreVersionsModalView.html'"></ng-include>
   <ng-include src="'syncthing/folder/restoreVersionsConfirmation.html'"></ng-include>
+  <ng-include src="'syncthing/folder/folderFilesModalView.html'"></ng-include>
   <ng-include src="'syncthing/settings/settingsModalView.html'"></ng-include>
   <ng-include src="'syncthing/settings/advancedSettingsModalView.html'"></ng-include>
   <ng-include src="'syncthing/settings/discardChangesConfirmation.html'"></ng-include>
@@ -885,6 +894,7 @@
   <script type="text/javascript" src="syncthing/core/tooltipDirective.js"></script>
   <script type="text/javascript" src="syncthing/core/uncamelFilter.js"></script>
   <script type="text/javascript" src="syncthing/core/uniqueFolderDirective.js"></script>
+  <script type="text/javascript" src="syncthing/core/treeViewDirective.js"></script>
   <script type="text/javascript" src="syncthing/core/validDeviceidDirective.js"></script>
   <script type="text/javascript" src="assets/lang/valid-langs.js"></script>
   <script type="text/javascript" src="assets/lang/prettyprint.js"></script>
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -496,7 +496,6 @@
                       <td class="text-right">
                         <span>
                           <a href="#" data-toggle="modal" data-target="#folderFiles" ng-click="setCurrentFolderAndGetContents(folder.id, '')">show</a>
-                        <a href="#" data-toggle="modal" data-target="#folderFiles" ng-click="setCurrentFolderAndGetContents(folder.id, 'neh2')">show neh2</a>
                         </span>
                       </td>
                     </tr>
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1098,8 +1098,7 @@ angular.module('syncthing.core')
                 $scope.parentFolderPrefix = null;
             }
 
-            let prefixArg = (prefix !== "") ? "&prefix=" + prefix.substr(1) : "";
-            $http.get(urlbase + '/db/browse?folder=' + id + "&levels=0" + prefixArg).success(function (data) {
+            $http.get(urlbase + '/db/browse?folder=' + id + "&levels=0&prefix=" + prefix.substr(1)).success(function (data) {
                 $scope.currentFolder.content = data;
             }).error($scope.emitHTTPError);
         };
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1085,11 +1085,9 @@ angular.module('syncthing.core')
         };
 
         $scope.setCurrentFolderAndGetContents = function (id) {
-            console.log("going to fetch data from controller")
             $scope.currentFolder = angular.copy($scope.folders[id]);
-            $http.get(urlbase + '/db/browse?folder=' + id + "&levels=0").success(function (data) {
+            $http.get(urlbase + '/db/browse?folder=' + id + "&levels=3").success(function (data) {
                 $scope.currentFolder.content = data;
-                console.log("Data: ", data)
             }).error($scope.emitHTTPError);
         };
 
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1085,9 +1085,11 @@ angular.module('syncthing.core')
         };
 
         $scope.setCurrentFolderAndGetContents = function (id) {
+            console.log("going to fetch data from controller")
             $scope.currentFolder = angular.copy($scope.folders[id]);
-            $http.get(urlbase + '/rest/db/browse').success(function (data) {
+            $http.get(urlbase + '/db/browse?folder=' + id + "&levels=0").success(function (data) {
                 $scope.currentFolder.content = data;
+                console.log("Data: ", data)
             }).error($scope.emitHTTPError);
         };
 
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -53,8 +53,8 @@ angular.module('syncthing.core')
         $scope.metricRates = false;
         $scope.folderPathErrors = {};
         $scope.currentFolder = {};
-        $scope.currentPath = '';
-        $scope.backPath = '';
+        $scope.currentFolderPrefix = '';
+        $scope.parentFolderPrefix = null;
 
         resetRemoteNeed();
 
@@ -1088,25 +1088,17 @@ angular.module('syncthing.core')
         };
 
         $scope.setCurrentFolderAndGetContents = function (id, prefix) {
-            console.log($scope.folders);
-            console.log("id", id, "prefix", prefix);
-
-            $scope.currentPath = prefix;
+            $scope.currentFolderPrefix = prefix;
             $scope.currentFolder = angular.copy($scope.folders[id]);
 
             let position = prefix.lastIndexOf("/");
             if (position !== -1) {
-                $scope.backPath = prefix.substr(0, position);
+                $scope.parentFolderPrefix = prefix.substr(0, position);
+            } else {
+                $scope.parentFolderPrefix = null;
             }
 
-            $scope.currentFolder.topLevel = true;
-            let prefixArg = "";
-
-            if (prefix !== "") {
-                $scope.currentFolder.topLevel = false;
-                prefixArg = "&prefix=" + prefix.substr(1);
-            }
-
+            let prefixArg = (prefix !== "") ? "&prefix=" + prefix.substr(1) : "";
             $http.get(urlbase + '/db/browse?folder=' + id + "&levels=0" + prefixArg).success(function (data) {
                 $scope.currentFolder.content = data;
             }).error($scope.emitHTTPError);
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -53,6 +53,9 @@ angular.module('syncthing.core')
         $scope.metricRates = false;
         $scope.folderPathErrors = {};
         $scope.currentFolder = {};
+        $scope.currentPath = '';
+        $scope.backPath = '';
+
         resetRemoteNeed();
 
         try {
@@ -1085,9 +1088,24 @@ angular.module('syncthing.core')
         };
 
         $scope.setCurrentFolderAndGetContents = function (id, prefix) {
+            console.log($scope.folders);
+            console.log("id", id, "prefix", prefix);
+
+            $scope.currentPath = prefix;
             $scope.currentFolder = angular.copy($scope.folders[id]);
-            $scope.currentFolder.topLevel = (prefix === "");
-            var prefixArg = (prefix !== "") ? ("&prefix=" + prefix) : "";
+
+            let position = prefix.lastIndexOf("/");
+            if (position !== -1) {
+                $scope.backPath = prefix.substr(0, position);
+            }
+
+            $scope.currentFolder.topLevel = true;
+            let prefixArg = "";
+
+            if (prefix !== "") {
+                $scope.currentFolder.topLevel = false;
+                prefixArg = "&prefix=" + prefix.substr(1);
+            }
 
             $http.get(urlbase + '/db/browse?folder=' + id + "&levels=0" + prefixArg).success(function (data) {
                 $scope.currentFolder.content = data;
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1084,6 +1084,13 @@ angular.module('syncthing.core')
             return deviceID.substr(0, 6);
         };
 
+        $scope.setCurrentFolderAndGetContents = function (id) {
+            $scope.currentFolder = angular.copy($scope.folders[id]);
+            $http.get(urlbase + '/rest/db/browse').success(function (data) {
+                $scope.currentFolder.content = data;
+            }).error($scope.emitHTTPError);
+        };
+
         $scope.findDevice = function (deviceID) {
             var matches = $scope.devices.filter(function (n) {
                 return n.deviceID === deviceID;
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1084,9 +1084,12 @@ angular.module('syncthing.core')
             return deviceID.substr(0, 6);
         };
 
-        $scope.setCurrentFolderAndGetContents = function (id) {
+        $scope.setCurrentFolderAndGetContents = function (id, prefix) {
             $scope.currentFolder = angular.copy($scope.folders[id]);
-            $http.get(urlbase + '/db/browse?folder=' + id + "&levels=3").success(function (data) {
+            $scope.currentFolder.topLevel = (prefix === "");
+            var prefixArg = (prefix !== "") ? ("&prefix=" + prefix) : "";
+
+            $http.get(urlbase + '/db/browse?folder=' + id + "&levels=0" + prefixArg).success(function (data) {
                 $scope.currentFolder.content = data;
             }).error($scope.emitHTTPError);
         };
--- a/gui/default/syncthing/core/treeViewDirective.js
+++ b/gui/default/syncthing/core/treeViewDirective.js
@@ -1,4 +1,23 @@
 angular.module('syncthing.core')
-    .directive('treeElement', function ($compile) {
-        //todo: handle directive
-})
+    .directive('treeElement', function () {
+        return {
+            restrict: 'EA',
+            scope: {
+                isDirectory: "=",
+                name: "@",
+                folderId: "@"
+            },
+            template: `
+                <div ng-if="isDirectory">
+                    <a href="#" ng-click="setCurrentFolderAndGetContents(folderId, name)">
+                        <span style="margin:10px;" class="fas fa-folder"></span>
+                        <span>{{name}}</span>
+                    </a>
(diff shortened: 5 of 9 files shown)


AudriusButkevicius:
In principle as functionality, seems to work, but I think we need to have a wider discussion how we expect to deliver this and how we expect the UI to look.

lib/model/model.go
```diff
@@ -69,6 +69,12 @@ type Availability struct {
 	FromTemporary bool              `json:"fromTemporary"`
 }
 
+type DirectoryTree struct {
```
AudriusButkevicius: DirectoryTree.IsDirectory = false is somewhat a misleading name.

As DirectoryTree is sometimes a file.

FolderItem or just Item?

lib/model/model.go
```diff
@@ -69,6 +69,12 @@ type Availability struct {
...
+	Name        string           `json:"name"`
+	IsDirectory bool             `json:"isDirectory"`
```
AudriusButkevicius: Could use protocol.FileInfoType enum value here?
This way we don't need boolean flags for each, and you'd know about symlinks etc.

I think it should just serialise straight to a string.

lib/model/model.go
```diff
@@ -2432,42 +2441,33 @@ func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
-					newdir := make(map[string]interface{})
-					last[path] = newdir
-					last = newdir
-				} else {
-					last = directory.(map[string]interface{})
-				}
-			}
+		fullName := strings.Split(f.Name, sep)
+		name := fullName[len(fullName)-1]
+		path := fullName[:len(fullName)-1]
```
AudriusButkevicius: I'd rather we used filepath.Dir/Base, as those functions were made for this purpose.

lib/model/model.go
```diff
@@ -2432,42 +2441,33 @@ func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
+
+		tree := DirectoryTree{Name: name, IsDirectory: f.IsDirectory(), Children: []*DirectoryTree{}}
```
AudriusButkevicius: no point allocating the children slice if it's not a directory.

lib/model/model.go
```diff
@@ -2432,42 +2441,33 @@ func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
+		for !reflect.DeepEqual(path, currentPrefix) && len(currentPrefix) != 0 {
```
AudriusButkevicius: I'd be more comfortable doing string prefix comparison (we hust have to make sure not to match `foo/bar` against `foo/barbarian`)

lib/model/model.go
```diff
@@ -2432,42 +2441,33 @@ func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
+			breadcrumbs = breadcrumbs[:len(breadcrumbs)-1]
+			currentPrefix = currentPrefix[:len(currentPrefix)-1]
 		}
 
-		if !dirsonly && base != "" {
-			last[base] = []interface{}{
-				f.ModTime(), f.FileSize(),
-			}
+		if !dirsonly || f.IsDirectory() {
+			breadcrumbs[len(breadcrumbs)-1].Children = append(breadcrumbs[len(breadcrumbs)-1].Children, &tree)
```
AudriusButkevicius: I wonder if it's not easier to track "parent", and update it as we traverse deeper or traverse out.
I know array indexing is fast, but the code is a lot less readable.

lib/model/model.go
```diff
@@ -2406,22 +2412,25 @@ func (m *model) RemoteSequence(folder string) (int64, bool) {
...
 
 	if prefix != "" && !strings.HasSuffix(prefix, sep) {
 		prefix = prefix + sep
 	}
 
+	output := DirectoryTree{Children: []*DirectoryTree{}}
+	breadcrumbs := []*DirectoryTree{&output}
+
+	var currentPrefix []string
 	files.WithPrefixedGlobalTruncated(prefix, func(fi db.FileIntf) bool {
```
AudriusButkevicius: This also does not return what you expect.

This returns all files advertised by all devices, some which we might not have, or some which we might ignored by us.

The original purpose of this function was to be able to make a browsable tree for users to see all files available across all devices, select which folders they want to sync by selecting them in the tree, and based on that adjusting the ignores file.


So I guess the question is, what exactly are we trying to answer with this feature. All files available, local files available? Local not ignored files? All not ignored files?

lib/model/model.go
```diff
@@ -2406,22 +2412,25 @@ func (m *model) RemoteSequence(folder string) (int64, bool) {
...
 
-	output := make(map[string]interface{})
 	sep := string(filepath.Separator)
 	prefix = osutil.NativeFilename(prefix)
 
 	if prefix != "" && !strings.HasSuffix(prefix, sep) {
 		prefix = prefix + sep
 	}
 
+	output := DirectoryTree{Children: []*DirectoryTree{}}
```
AudriusButkevicius: rename this to root?

lib/model/model_test.go
```diff
@@ -2035,13 +2003,64 @@ func TestGlobalDirectorySelfFixing(t *testing.T) {
...
+						}),
+					}),
+				}),
+			}),
+		}),
+	}
+
+	if mm(actual) != mm(expected) {
+		t.Errorf("Does not match:\n%s\n%s", mm(actual), mm(expected))
+	}
```
AudriusButkevicius: I suspect this should stop working, as I think your current implementation breaks this "feature"

AudriusButkevicius:
Sorry, answers to your questions which I should have addressed beforehand:
1. I am not sure how much of an issue it is. Perhaps we should not try to fix this, and crash the application instead, as it uncovers inconsistencies in the database. @calmh the context is: "when iterating foo/bar/baz, we somehow not encounter one of the parents while iterating the index". I think it's fair to crash in that case?
2. I think we should add/update documentation about the new API endpoint, the feature itself I think is fairly self explanatory, but if you feel it's worth mentioning, then by all means, file a PR to the docs repo.
3. That's fine, internal data model uses / anyway, and I think we do normalization everywhere.
4. These would be picked up automatically, as long as the html tags have "translate" tag on them, or js stuff is piped via `| translate`
5. I think to make this as versitile and as useful for other purposes than the one at hand, we should include additional metadata.
6. That env var is only used at runtime not compile time, so it has no effect if you set it locally. I have it in my .bash_profile to disable it for good when developing.

gui/default/syncthing/folder/folderFilesModalView.html
```diff
@@ -0,0 +1,27 @@
+<modal id="folderFiles" status="default" icon="fas fa-cog" heading="{{'Included files' | translate}}" large="yes" closeable="yes">
+    <div class="modal-body">
+        <table class="table table-condensed table-striped table-auto">
+            <tbody>
+            <tr ng-repeat="file in currentFolder.content">
```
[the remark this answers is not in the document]
imsodin: Either that or just reuse the pageable stuff from out-of-sync et al modals.

lib/model/model.go
imsodin: That, definitely. Whenever you need to manipulate a path manually (not through `path/filepath`), a comment explaining what happens is probably a good idea.

lib/model/model.go
nitrajka: Do you have some reasonable argument why is matching prefixes better? I understand, why is string comparison better than comparing slices of strings. But why prefix matching?

In case of string matching, I had to care about delimiters which here and there caused a spaghetti code. I realized, that path won't be longer than some small constant, therefore I decided to use slice instead of strings. IMHO, the code is better without caring about delimiters... Is that really a problem?

lib/model/model.go
nitrajka: I didn't think those two behave as I needed in my algorithm. Now I see they do. :)
However, I still need `Dir()` to return an empty string instead of `.`. Guess I just `if(Dir()==".") {path = "" }` ?

Please, consider the comment below about `path` and `name` being slices instead of integers.

lib/model/model.go
nitrajka: Good point. Will rename.

gui/default/syncthing/folder/folderFilesModalView.html
```diff
@@ -0,0 +1,27 @@
...
+                <td tree-element
+                    is-directory="file.isDirectory"
+                    name="{{file.name}}"
+                    show="setCurrentFolderAndGetContents(currentFolder.id, currentFolderPrefix + '/' + file.name)"
+                >
```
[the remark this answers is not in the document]
nitrajka: Sorry, didn't notice that one. You are right.

lib/model/model.go
nitrajka: Yes, makes sense.

lib/model/model.go
AudriusButkevicius: Essentially, the thing I don't like is reflect. I guess I meant equality comparison but said prefix comparison.
I am ok to compare string slices, but just not using reflect.

lib/model/model.go
AudriusButkevicius: Not sure about them being integers, but doing the if suggested sounds ok to me.

lib/model/model.go
nitrajka: Well yes, I didn't expect that 😀I just returned a structure which came in a way of paths in strings. (Didn't include files if `dirsOnly` was set to True) I programmed it to make the existing tests pass. I thought to make the tests pass while changing the resulting structure would be enough.
I thought that this resulting structure should contain local, not ignored files (I forgot to not include the ignored ones) from one shared device's directory (for example `default` or other, not all of them). 
The piece of code where we sent all of the files available across all devices to `WithPrefixedGlobalTruncated` is already done? (I didn't detect such piece of code.) Or should It be done as well?

lib/model/model_test.go
nitrajka: Yes, this test function (`TestGlobalDirectorySelfFixing`) is failing. wasn't sure about the implementation in those test cases, therefore I rather asked first.

lib/model/model.go
AudriusButkevicius: WithPrefixedGlobal is the function that gives you all the files available from all devices. It does not take local ignoring into account however.

imsodin: I think you can keep your changes and just create two wrapper functions `GlobalDirectoryTree` and `LocalDirectoryTree`, which under the hood both do the same thing but use `files.WithPrefixedGlobalTruncated` or `files.WithPrefixedHaveTruncated` (with local device ID) respectively. That could even be extended to a `RemoteDirectoryTree(...remoteID...)` which returns the files as seen by the remote devices (probably unnecessary though).

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1084,6 +1087,23 @@ angular.module('syncthing.core')
...
+            $scope.currentFolder = angular.copy($scope.folders[id]);
+
+            let position = prefix.lastIndexOf("/");
+            if (position !== -1) {
+                $scope.parentFolderPrefix = prefix.substr(0, position);
+            } else {
+                $scope.parentFolderPrefix = null;
+            }
+
+            let prefixArg = (prefix !== "") ? "&prefix=" + prefix.substr(1) : "";
```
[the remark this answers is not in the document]
nitrajka: It does.

gui/default/syncthing/folder/folderFilesModalView.html
```diff
@@ -0,0 +1,27 @@
+<modal id="folderFiles" status="default" icon="fas fa-cog" heading="{{'Included files' | translate}}" large="yes" closeable="yes">
+    <div class="modal-body">
```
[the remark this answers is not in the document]
nitrajka: Breadcrumbs are quite a good idea. I'll see what I can do.

Did you mean to have `..` instead of the back button?
> users are used to that
Did you mean users of command line? 😀 IMHO, not all of the users know command line, but all of them surely know a file browser. All of the file browsers have a back button. (At least I think so, not sure.)
Or did you just mean to put it on the top of the panel instead of the bottom?

gui/default/syncthing/folder/folderFilesModalView.html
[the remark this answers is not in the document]
nitrajka: Makes sense.

lib/model/model.go
AudriusButkevicius: Actually checking which files which device has might be a nice feature.

gui/default/index.html
```diff
@@ -491,6 +491,14 @@ <h4 class="panel-title">
...
                       </td>
                     </tr>
+                    <tr>
+                      <th><span class="fas fa-fw fa-copy"></span>&nbsp;<span translate>Included files</span></th>
+                      <td class="text-right">
+                        <span>
+                          <a href="#" data-toggle="modal" data-target="#folderFiles" ng-click="setCurrentFolderAndGetContents(folder.id, '')">show</a>
+                        </span>
+                      </td>
+                    </tr>
```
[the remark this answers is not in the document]
nitrajka: Sorry, I don't know where is "a bit of text that appears under filecounts when ignores are setup". Could you please somehow navigate me? Or screenshot?

According to what you wrote in one of the comments
>to see all files available across all devices, select which folders they want to sync by selecting them in the tree, and based on that adjusting the ignores file

it quite makes sense to display all files (but symlinks, I noticed, syncthing doesn't sync those) and indicate their status, maybe metadata. But okay, I'll start small.

lib/model/model.go
nitrajka: It doesn't serialise to a string 😞 
Screen Shot 2020-01-27 at 18 18 58
But I can still use `if fileType === 0 {} else if fileType === 1...` when rendering, right?

lib/model/model.go
nitrajka: That is true. However, this:

```
var children []*FolderItem = nil
if f.FileType() == 1 {
	children = make([]*FolderItem, 0)
}
tree := FolderItem{Name: name, FileType: f.FileType(), Children: children}
``` 
is much longer and less readable that this:
tree := FolderItem{Name: name, FileType: f.FileType(), Children:[]*FolderItem{}}

Isn't the capacity in the second case allocated to 0 anyway? Just asking. I will use the `if` if you think it would be better.

lib/model/model.go
nitrajka: Ok, I understand.
In my new solution, I still use a slice, but when comparing `currentPrefix` with the prefix of a given `FolderItem`, I create a string from `currentPrefix`. 

```
for strings.Join(currentPrefix, sep) != path && len(currentPrefix) != 0 {
	breadcrumbs = breadcrumbs[:len(breadcrumbs)-1]
	currentPrefix = currentPrefix[:len(currentPrefix)-1]
}
```
It is still in O(n^2) time (as before with reflect), we do not use reflect and does not have to care about delimiters.

lib/model/model.go
nitrajka: In my opinion, that would be the same amount of work, just in a different place in the code. But if you insist, I can try to implement it as you suggest.

lib/model/model.go
nitrajka: Found out, that no `if` needed. (used tests and debugger as well to test it)

lib/model/model.go
imsodin: On the need endpoints where file infos are returned, `FileInfoType.String` is used for serialization. Maybe for programmatic purposes it's even easier to deal with ints?

nitrajka: Should I add an icon for `FileInfoTypeDeprecatedSymlinkFile` (2) and `FileInfoTypeDeprecatedSymlinkDirectory` (3) as well? Even if they are deprecated?

nitrajka: As for me, we can have ints. It is more error-prone to compare strings than ints.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1084,6 +1087,23 @@ angular.module('syncthing.core')
...
+        $scope.setCurrentFolderAndGetContents = function (id, prefix) {
+            $scope.currentFolderPrefix = prefix;
+            $scope.currentFolder = angular.copy($scope.folders[id]);
+
+            let position = prefix.lastIndexOf("/");
+            if (position !== -1) {
+                $scope.parentFolderPrefix = prefix.substr(0, position);
+            } else {
+                $scope.parentFolderPrefix = null;
+            }
```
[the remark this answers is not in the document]
nitrajka: I am not one either. But I understand your point.

lib/model/model.go
imsodin: Nono, ignore deprecated stuff for new features.

gui/default/index.html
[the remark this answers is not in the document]
AudriusButkevicius: image
this is the block of text I am referring to. Or perhaps make the numbers themselves clickable?

gui/default/syncthing/folder/folderFilesModalView.html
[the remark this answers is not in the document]
AudriusButkevicius: I don't really care if it's back arrow as a list item, or .., or a back button above the item list, but usually these controls are at the top, not at the bottom.

gui/default/index.html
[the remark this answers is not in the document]
imsodin: That seems quite neat, thus there are already three areas to click for global, local and ignored file trees.

lib/model/model.go
AudriusButkevicius: Int's are not nice for public APIs, so let's use strings.

I suspect you won't find those two deprecated anywhere anymore, but for carefulness sake, we should just return them as symlinks.

lib/model/model.go
AudriusButkevicius: I for some reason think you will need it, as the old code had it. 
I suspect it only comes up when you try to get the details of a file at the root, as the filesystem abstraction uses relative path resolution internally, which leads to "what is x/z/y relative to the filesystem abstraction root x/z" which returns y, which then Dir() returns "."

lib/model/model.go
AudriusButkevicius: Array with the size of zero is not the same as no array, as there is still array struct allocated with size and capacity.

In theory, you can just leave it as nil in all cases because when you do:

`foo.Children = append(foo.Children, newChild)`

if foo.Children is nil, the append would allocate a new array for you.

gui/default/syncthing/folder/folderFilesModalView.html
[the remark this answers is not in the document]
AudriusButkevicius: I think pagination is very unusual for a "directory browser" kind of thing. I think a container scrolling inside the modal is better.

gui/default/syncthing/folder/folderFilesModalView.html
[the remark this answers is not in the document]
nitrajka: Ah yes, now I understand.

nitrajka:
@AudriusButkevicius 
I am sorry for the resolved comments. I am used to resolving them (from a private project) when I think we (me and the reviewer) agreed on something. Usually, in that case, I write a comment like "Makes sense" or "You are right. I will reimplement that". The reason why I mark them as resolved is to keep my todo list with the PR obvious. The resolved ones are already locally implemented and I am testing them. The not resolved ones are yet to be done. I am sorry, one more time for the confusion. I won't resolve them manually (or at all) for Syncthing project from now on.

gui/default/syncthing/folder/folderFilesModalView.html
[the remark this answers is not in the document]
nitrajka: Screen Shot 2020-01-28 at 10 10 32

What do you say to this? I added smoe css to `gui/default/assets/css/overrides.css`

lib/model/model.go
nitrajka: Okay, I'll add support for those two as well. And the thing about the public api and ints, I agree. The values should be self-explanatory for people who use just public api. I'll try to serialize them to string.

lib/model/model.go
nitrajka: In the example with `x/z/y`, the `Base()` returns `y` (as you wrote) and `Dir()` should return `x/z`. Am I missing something?

> strings.Split() will have to be replaced with a loop
> if it ends up being super ugly we can live with this

IMHO, I think ugly code will be less readable. Why should we replace `strings.Split()` with custom code if the function does everything we want for us on one line?

lib/model/model.go
nitrajka: Ah so. Now I understand. I'll leave `nill` there.

lib/model/model.go
nitrajka: Hmm, I see, that file which maps strings to ints is generated.

lib/model/model.go
nitrajka: So you intend to compare it characterwise?
Yes, this looks more wasteless, however, I do not like the idea, where I should care about delimiters. IMHO, it is less readable. 
Anyway, isn't that a bit overoptimization? The file depth will really be small.

lib/model/model.go
nitrajka: Okay, the idea with these 2 lines you wrote, seems more reasonable to me than tracking a parent. I don't have to rework the algorithm from scratch. I just split the oneliner.

lib/model/model.go
```diff
@@ -2432,42 +2441,36 @@ func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
-					last = newdir
-				} else {
-					last = directory.(map[string]interface{})
-				}
-			}
+		name := filepath.Base(f.Name)
+		path := filepath.Dir(f.Name)
+
+		var children []*FolderItem = nil
+		if f.FileType() == 1 {
```
imsodin:
```suggestion
		if f.FileType() == protocol.FileInfoTypeDirectory {
```

lib/model/model.go
imsodin: Splitting this line as proposed by Audrius would definitely help. And I just spent quite a bit of time figuring out how the entire loop works (if I did figure it out). Is this correct: `currentPrefix` contains the path components of the latest encountered directory and `breadcrumbs` the tree nodes aka `FolderItem`s corresponding to those path components? If that's correct, comments stating something like that when declaring those variables would be a good idea.

lib/model/model.go
AudriusButkevicius: IsDir()

gui/default/syncthing/core/treeViewDirective.js
```diff
@@ -0,0 +1,27 @@
...
+                    <span style="margin:10px;" class="fas fa-file"></span>
+                    <span>{{name}}</span>
+                </div>
+                <div ng-if="fileType === 1">
+                    <a href="#" ng-click="show()">
+                        <span style="margin:10px;" class="fas fa-folder"></span>
+                        <span>{{name}}</span>
+                    </a>
+                </div>
+                <div ng-if="fileType === 4">
```
AudriusButkevicius: As discussed, this should be strings, integers for enums for public APIs is a big pain.

gui/default/syncthing/folder/folderFilesModalView.html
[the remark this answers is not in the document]
AudriusButkevicius: I personally think it's better. 
I'd reduce the amount of vertical space the list items take up, as it seems we have 2x the font size of gaps between the lines of text. Perhaps it's the icons that are causing this.

lib/model/model.go
```diff
@@ -2445,29 +2445,32 @@ func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
+		name := filepath.Base(f.Name)
+		path := filepath.Dir(f.Name)
 
-		tree := DirectoryTree{Name: name, IsDirectory: f.IsDirectory(), Children: []*DirectoryTree{}}
-		for !reflect.DeepEqual(path, currentPrefix) && len(currentPrefix) != 0 {
+		var children []*FolderItem = nil
+		if f.FileType() == 1 {
+			children = make([]*FolderItem, 0)
+		}
+		tree := FolderItem{Name: name, FileType: f.FileType(), Children: children}
```
AudriusButkevicius: Could do f.FileType().String().

Also, as I suggested, children can always be nil (simply omitted from declaration all together), due to how append works..

lib/model/model.go
```diff
@@ -2445,29 +2445,32 @@ func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
 			currentPrefix = currentPrefix[:len(currentPrefix)-1]
 		}
 
-		if !dirsonly || f.IsDirectory() {
+		if !dirsonly || f.IsDirectory() && !f.IsIgnored() {
 			breadcrumbs[len(breadcrumbs)-1].Children = append(breadcrumbs[len(breadcrumbs)-1].Children, &tree)
 		}
 
-		if f.IsDirectory() {
+		if f.IsDirectory() && !f.IsIgnored() {
```
AudriusButkevicius: Should f.IsIgnored() be a flag that is part of FolderItem which is visible but potentially greyed out in the tree view?
If not, perhaps this should an additional parameter on the rest API call? Seems like this will become a generally useful endpoint.

gui/default/assets/css/overrides.css
```diff
@@ -414,4 +419,12 @@ ul.three-columns li, ul.two-columns li {
...
     }
+
+    #folderFiles .modal-header {
+        min-height: 74px;
+    }
 }
+
+#folderFiles .modal-header {
+    min-height: 64px;
+}
```
AudriusButkevicius: Can you give a rough explanation why we're adding these? Presumably the header would be as tall as the content needs, if that's the case, why do we need a min?

gui/default/assets/css/overrides.css
```diff
@@ -206,6 +206,11 @@ table.table-auto td {
     margin-bottom: 0.25em;
 }
 
+#folderFiles .modal-body {
+    height: 300px;
```
AudriusButkevicius: There must be a better way of doing this. 

Perhaps we can do something like height 90% (or whatever the margins from the top are to make top and bottom gaps equal)

go.mod
```diff
@@ -26,6 +26,8 @@ require (
 	github.com/maruel/panicparse v1.3.0
 	github.com/mattn/go-isatty v0.0.11
 	github.com/minio/sha256-simd v0.1.1
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+	github.com/modern-go/reflect2 v1.0.1 // indirect
```
AudriusButkevicius: I think these are no longer needed?

lib/model/model.go
nitrajka: @imsodin Yes, you figured it out correctly. The comments make sense.

gui/default/syncthing/folder/folderFilesModalView.html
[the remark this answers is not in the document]
ibizaman: What do you think of re-using fancytree that is already vendored here?

st-review:
:robot: *beep boop*

PR #6290, opened 2020-01-23
lib/config, lib/model: Limit concurrent pulls (fixes #5914)

### LINKED ISSUE #5914: Limit folder parallelization, or otherwise figure out & solve performance for many folders
Assume two devices, 200 GIB of random data (`fakefs` on both sides), one side scanned and idle and the other one empty.

With one folder * 200 GiB it syncs in reasonable time (~30-60 minutes).

With 1000 folders * 200 MiB it takes unreasonable time (24 hours or more).

We should resolve this, either by figuring out what the problem is exactly or, as a quick fix, add a sync limiter similar to the scan limiter so that we only sync maybe 10 folders or so in parallel by default.

calmh:
### Purpose

Adds a new folder state "Waiting to Sync" in the same vein as the existing "Waiting to Scan". This vastly improves performances in the rare cases when there are lots and lots of folders operating.

I also changed the default zero value to mean "limit to num-cpu-cores" instead of "no limit".

### Testing

I've run it a bunch and looked in the GUI...


gui/default/index.html
```diff
@@ -313,7 +313,7 @@ <h4 class="panel-title">
                   <span ng-switch-when="paused"><span class="hidden-xs" translate>Paused</span><span class="visible-xs" aria-label="{{'Paused' | translate}}"><i class="fas fa-fw fa-pause"></i></span></span>
                   <span ng-switch-when="unknown"><span class="hidden-xs" translate>Unknown</span><span class="visible-xs" aria-label="{{'Unknown' | translate}}"><i class="fas fa-fw fa-question-circle"></i></span></span>
                   <span ng-switch-when="unshared"><span class="hidden-xs" translate>Unshared</span><span class="visible-xs" aria-label="{{'Unshared' | translate}}"><i class="fas fa-fw fa-unlink"></i></span></span>
-                  <span ng-switch-when="scan-waiting"><span class="hidden-xs" translate>Waiting to scan</span><span class="visible-xs" aria-label="{{'Waiting to scan' | translate}}"><i class="fas fa-fw fa-hourglass-half"></i></span></span>
+                  <span ng-switch-when="scan-waiting"><span class="hidden-xs" translate>Waiting to Scan</span><span class="visible-xs" aria-label="{{'Waiting to Scan' | translate}}"><i class="fas fa-fw fa-hourglass-half"></i></span></span>
```
AudriusButkevicius: Why this change?
Extra translators, and scan is not a name or a place so that capitalisation would naturally make sense?

gui/default/index.html
```diff
@@ -324,6 +324,10 @@ <h4 class="panel-title">
                   </span>
                   <span ng-switch-when="idle"><span class="hidden-xs" translate>Up to Date</span><span class="visible-xs" aria-label="{{'Up to Date' | translate}}"><i class="fas fa-fw fa-check"></i></span></span>
                   <span ng-switch-when="localadditions"><span class="hidden-xs" translate>Local Additions</span><span class="visible-xs" aria-label="{{'Local Additions' | translate}}"><i class="fas fa-fw fa-check"></i></span></span>
+                  <span ng-switch-when="sync-waiting">
+                    <span class="hidden-xs" translate>Waiting to Sync</span>
+                    <span class="visible-xs" aria-label="{{'Waiting to Sync' | translate}}"><i class="fas fa-fw fa-hourglass-half"></i></span>
```
AudriusButkevicius: sync I guess?234

lib/config/config_test.go
```diff
@@ -86,8 +86,13 @@ func TestDefaultValues(t *testing.T) {
 
 func TestDeviceConfig(t *testing.T) {
 	for i := OldestHandledVersion; i <= CurrentVersion; i++ {
+		cfgFile := fmt.Sprintf("testdata/v%d.xml", i)
+		if _, err := os.Stat(cfgFile); os.IsNotExist(err) {
+			continue
+		}
```
AudriusButkevicius: We should also get rid of the noop files that we've produced over the years.

lib/config/optionsconfiguration.go
```diff
@@ -152,3 +154,20 @@ func (opts OptionsConfiguration) GlobalDiscoveryServers() []string {
 	}
 	return util.UniqueTrimmedStrings(servers)
 }
+
+func (opts OptionsConfiguration) EffectiveMaxConcurrentFolders() int {
+	// If a value is set, trust that.
+	if opts.MaxConcurrentFolders > 0 {
+		return opts.MaxConcurrentFolders
+	}
```
AudriusButkevicius: I guess you want people to set it to 999 for "unlimited"?

test/folders.sh
```diff
@@ -0,0 +1,11 @@
...
+
+for ((id=0;id<200;id++)); do
+cat <<EOT
+    <folder id="$id" label="" path="$id?maxsize=1000&amp;seed=$id" type="sendreceive" rescanIntervalS="3600" fsWatcherEnabled="false" fsWatcherDelayS="10" ignorePerms="false" autoNormalize="true">
+        <filesystemType>fake</filesystemType>
+        <device id="I6KAH76-66SLLLB-5PFXSOA-UFJCDZC-YAOMLEK-CP2GB32-BV5RQST-3PSROAU" introducedBy=""></device>
+        <device id="MRIW7OK-NETT3M4-N6SBWME-N25O76W-YJKVXPH-FUMQJ3S-P57B74J-GBITBAC" introducedBy=""></device>
+    </folder>
+EOT
+done
```
AudriusButkevicius: What's this?

lib/config/optionsconfiguration.go
```diff
@@ -152,3 +154,20 @@ func (opts OptionsConfiguration) GlobalDiscoveryServers() []string {
...
+	// Otherwise default to the number of CPU cores in the system as a rough
+	// approximation of system powerfullness.
+	if n := runtime.GOMAXPROCS(-1); n > 0 {
+		return n
+	}
+	// We should never get here to begin with, but since we're here let's
+	// use some sort of reasonable compromise between the old "no limit" and
+	// getting nothing done... (Median number of folders out there at time
+	// of writing is two, 95-percentile at 12 folders.)
+	return 4 // https://xkcd.com/221/
```
AudriusButkevicius: Perhaps worth adding a notice about this new behaviour? Or you reckon change log entry is enough?

Essentially, if I had a setup where I have 10 folders, each on a different (or simply much more capable) io subsystem, this might lead to a "wtf why is it slow" moment.

gui/default/index.html
imsodin: Consistency -  "Title Casing" is used throughout the web UI. I left it alone when adding "Preparing to Sync" due to re-translation, but I think that's not going to be an issue.

calmh: This

lib/config/optionsconfiguration.go
calmh: I was intending -1 to mean unlimited as usual, but I see know that I didn't implement that.

lib/config/optionsconfiguration.go
calmh: I think change log is enough.

(And this specific case (`return 4`) might as well have been a panic, I don't think gomaxprocs ever returns <= 0, although I guess you mean the larger behavior change of defaulting to gomaxprocs.)

lib/config/optionsconfiguration.go
```diff
@@ -152,3 +154,20 @@ func (opts OptionsConfiguration) GlobalDiscoveryServers() []string {
 	}
 	return util.UniqueTrimmedStrings(servers)
 }
+
+func (opts OptionsConfiguration) EffectiveMaxConcurrentFolders() int {
```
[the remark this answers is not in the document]
calmh: ✅

test/folders.sh
calmh: Tiny script to generate lots of folder entries for testing; I had to write it several times so though I might as well check it in 🤷‍♂

lib/config/optionsconfiguration.go
calmh: ✅ fixed

lib/config/config_test.go
calmh: ✅

gui/default/index.html
calmh: Title case as noted

lib/model/folder.go
```diff
@@ -279,6 +280,16 @@ func (f *folder) getHealthError() error {
 	return nil
 }
 
+func (f *folder) pull() bool {
+	f.setState(FolderSyncWaiting)
```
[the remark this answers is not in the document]
calmh: ✅

lib/config/optionsconfiguration.go
```diff
@@ -52,7 +53,7 @@ type OptionsConfiguration struct {
 	TrafficClass            int      `xml:"trafficClass" json:"trafficClass"`
 	DefaultFolderPath       string   `xml:"defaultFolderPath" json:"defaultFolderPath" default:"~"`
 	SetLowPriority          bool     `xml:"setLowPriority" json:"setLowPriority" default:"true"`
-	MaxConcurrentScans      int      `xml:"maxConcurrentScans" json:"maxConcurrentScans"`
+	RawMaxConcurrentFolders int      `xml:"maxConcurrentFolders" json:"maxConcurrentFolders"`
```
AudriusButkevicius: MaxFolderConcurrency? Current one reads badly and doesn't really explain what it does. Feels like a limit of how many folders I can have.

calmh:
Valid point, I fix

PR #6292, opened 2020-01-24
gui: Prevent text overflow in file lists (fixes #6268)

### LINKED ISSUE #6268: Out of sync panel layout overflows
Sometimes Out of sync List overflows.

98b14f0f16c568817eff122ff38a830a087377ab

Looks like the first filename is too long

v1.3.2, Linux (64 bit)
Chrome 79.0.3945.117 (64 бит)

tomasz1986:
### Purpose

This change forces breakage of long file paths and names, so that the text does not overflow the available display area, which is the case in many places in the GUI right now, such as "Recent Changes" or "Out of Sync Items". See the commit for details.

### Testing

I have tested in Chromium 79.0.3945.130 and IE 11.973.17763.0, but the `word-break: break-all` CSS rule itself is supported in all browsers, starting from at least IE 6.

### Screenshots

Text cut and date not visible before:

image

Text wrapped and date visible after:

image

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -136,6 +136,11 @@ table.table-auto td {
     max-width: 0px;
 }
 
+/* Wrap long file paths to prevent text overflow. See issue #6268. */
+.file-path {
+    word-break: break-all;
+}
+
 .folder-advanced {
     padding: 1rem;
     margin-bottom: 15px;
--- a/gui/default/syncthing/device/globalChangesModalView.html
+++ b/gui/default/syncthing/device/globalChangesModalView.html
@@ -20,7 +20,7 @@
             <td>{{changeEvent.data.action}}</td>
             <td>{{changeEvent.data.type}}</td>
             <td class="no-overflow-ellipse">{{folderLabel(changeEvent.data.folder)}}</td>
-            <td class="no-overflow-ellipse">{{changeEvent.data.path}}</td>
+            <td class="file-path no-overflow-ellipse">{{changeEvent.data.path}}</td>
             <td class="no-overflow-ellipse">{{changeEvent.time | date:"yyyy-MM-dd HH:mm:ss"}}</td>
           </tr>
         </tbody>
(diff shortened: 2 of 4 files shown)


imsodin:
The following is just a question, if it's not feasible feel free to ignore:  
The screenshot in #6286 shows no breaks at path-separators. Some googling suggests that many browser will break at path separators with `break-word`, which would suffice in that example and in general would be much nicer. As you say this isn't enough if an individual path component is too long. Is there any way to make it break at path separators and only break at arbitrary points if necessary?

tomasz1986:
Thank you for the comment.

I have just done a few more tests, and also read some things, and it seems that `break-word` is deprecated in the newest CSS specifications.

>For compatibility with legacy content, the word-break property also supports a deprecated break-word keyword. When specified, this has the same effect as word-break: normal and overflow-wrap: anywhere, regardless of the actual value of the overflow-wrap property.
>
>source: https://www.w3.org/TR/css-text-3/#propdef-word-break

However, in Chromium both `break-word` and `break-all` seem to behave exactly the same, i.e. the text is cur wherever necessary, with no respect for the slashes.

The biggest problem is that Internet Explorer does not support `break-word` at all, only `break-all`, so it seems that `break-all` is the only viable solution. I have updated the commit with this information.

imsodin:
Sounds reasonable.

There's also quite a few other modals where paths are featured, e.g. failedFilesModalView.html,  localChangedFilesModalView.html, ... They probably should all get this class too.

tomasz1986:
Yes, there are likely more places where such full paths are used. I had not had enough time to check all of them, so I just decided to fix the two very visible ones first.

I should be able to have a look at them later this week, and then update the commit to cover all the cases.

AudriusButkevicius:
Should we roll with this?

tomasz1986:
Please wait a day or two. There is at least one more place, where the CSS should be applied. For other places, they are already affected by `word-break: break-all` cascaded from the `body`, and I am not sure whether we should apply the property twice (just in case), or not.

tomasz1986:
You can roll with this, I believe.

I have only added the the class to the lines, which actually need it. I have left alone those that have `word-break: break-all` already applied from the `body`.

PR #6295, opened 2020-01-26
lib/db: Configurable block GC time

calmh:
Also retain the interval over restarts by storing last GC time in the
database. This to make sure that GC eventually happens even if the
interval is configured to a long time (say, a month).

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -9,6 +9,7 @@ package db
 import (
 	"bytes"
 	"encoding/binary"
+	"os"
 	"time"
 
 	"github.com/syncthing/syncthing/lib/db/backend"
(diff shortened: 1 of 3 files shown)


lib/db/lowlevel.go
```diff
@@ -473,10 +483,33 @@ func (db *Lowlevel) gcRunner() {
...
+// zero if it should happen directly.
+func (db *Lowlevel) timeUntil(key string, every time.Duration) time.Duration {
+	miscDB := NewMiscDataNamespace(db)
+	lastTime, _, _ := miscDB.Int64(key) // error wilfully ignored
+	nextTime := time.Unix(lastTime, 0).Add(every)
+	sleepTime := time.Until(nextTime)
+	if sleepTime < 0 {
+		sleepTime = 0
+	}
+	l.Infoln("sleep time is", sleepTime)
```
AudriusButkevicius: Debug?

calmh: Ugh

Issue #6284, opened 2020-01-21
Database: (optionally) garbage collect deleted files


calmh:
So, large databases again. Here's stats for a block-deduplicated yet still large database. There are a couple of custom classes:

```
% stindex -mode account index-v0.14.0.db
 0x00:  2206187 items, 182629 KB keys + 1463447 KB data,  82 B +  663 B avg,    2573 B max
 0x01:   185656 items,  10388 KB keys + 1790909 KB data,  55 B + 9646 B avg,  924525 B max
 0x02:   916890 items,  84795 KB keys +    3667 KB data,  92 B +    4 B avg,     192 B max
 0x03:      384 items,     27 KB keys +       5 KB data,  72 B +   15 B avg,      87 B max
 0x04:     1109 items,     17 KB keys +      17 KB data,  15 B +   15 B avg,      69 B max
 0x06:      383 items,      3 KB keys +       0 KB data,   9 B +    2 B avg,      18 B max
 0x07:      510 items,      4 KB keys +      12 KB data,   9 B +   24 B avg,      41 B max
 0x09:      194 items,      0 KB keys +     123 KB data,   5 B +  634 B avg,   11484 B max
 0x0a:        3 items,      0 KB keys +       0 KB data,  14 B +   20 B avg,      58 B max
 0x0b:   181836 items,   2363 KB keys +   10694 KB data,  13 B +   58 B avg,     173 B max
 0x0d:    44282 items,   1461 KB keys +   59899 KB data,  33 B + 1352 B avg, 1606431 B max
 0xf0:   206952 items,  23084 KB keys +   31301 KB data, 111 B +  151 B avg,     340 B max
 0xf1:  6725022 items, 665162 KB keys + 1116642 KB data,  98 B +  166 B avg,    4788 B max
 Total 10469408 items, 969939 KB keys + 4476724 KB data.
```

The two interesting ones here are f0 (deleted files < 90 days old) and f1 (deleted files > 90 days old). These are actually class 00 that I've just broken out for accounting. Compared to 00 (normal, live files) the old deleted files are almost half of the total volume (1.1 GB out of 1.4+1.1 GB) and a vast majority of the items (6.7 M out of 6.7+2.2 M). The latter especially is a lot of iteration that wouldn't need to happen, when we iterate. Oh, and volume wise it's the vast majority of keys as well.

[...]

PR #6299, opened 2020-01-29
lib/db: Garbage collect deleted items

calmh:
### Purpose

This is in relation #6284 in order to clean out old deleted files. It's not beautiful in any way whatsoever, and there may be things I didn't consider, so consider this a suggestion/wip.

This new GC step simply iterates and deleted items that are 1) deleted 2) not in the need list 3) older than a threshold. We drop our file entry, every one elses file entries, and the global entry. Poof, gone! If the other side runs with the same set up then they'll do the same and that's the end of it. If they don't, the  entry will get reintroduced next time we do a full index transfer and the cycle repeats... 

One thing not handled: metadata accounting adjustment. Maybe we need to, maybe we don't (deleted files don't play too large a role there), I haven't considered it closely.

As for why bother, the linked issue shows a database where deleted items are the vast majority of items. I don't have any reason to believe this is atypical for how it looks after being up and running for a while.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -521,7 +521,7 @@ func (db *Lowlevel) gcRunner() {
 				l.Warnln("Deleted file GC failed:", err)
 			}
 			db.recordTime(deletedGCTimeKey)
-			blocksTimer.Reset(db.timeUntil(deletedGCTimeKey, deletedGCInterval))
+			deletedTimer.Reset(db.timeUntil(deletedGCTimeKey, deletedGCInterval))
 		}
 	}
 }
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -28,13 +28,28 @@ const (
 	blockGCBloomFalsePositiveRate = 0.01 // 1%
 	blockGCDefaultInterval        = 13 * time.Hour
 	blockGCTimeKey                = "lastBlockGCTime"
+
+	deletedGCDefaultAge      = 180 * 24 * time.Hour
+	deletedGCMinAge          = 7 * 24 * time.Hour
+	deletedGCDefaultInterval = 37 * time.Hour
+	deletedGCTimeKey         = "lastDeletedGCTime"
 )
 
-var blockGCInterval = blockGCDefaultInterval
+var (
+	blockGCInterval   = blockGCDefaultInterval
+	deletedGCAge      = deletedGCDefaultAge
+	deletedGCInterval = deletedGCDefaultInterval
+)
 
 func init() {
 	if dur, err := time.ParseDuration(os.Getenv("STGCBLOCKSEVERY")); err == nil {
 		blockGCInterval = dur
+	}
+	if dur, err := time.ParseDuration(os.Getenv("STGCDELETEDAGE")); err == nil {
+		deletedGCAge = dur
+	}
+	if dur, err := time.ParseDuration(os.Getenv("STGCDELETEDEVERY")); err == nil {
+		deletedGCInterval = dur
 	}
 }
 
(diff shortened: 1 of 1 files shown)


calmh:
I intended to file this as a draft PR but I didn't and I don't see a way to convert it now...

lib/db/lowlevel.go
```diff
@@ -590,6 +627,122 @@ func (db *Lowlevel) gcBlocks() error {
...
+	}
+	defer t.Release()
+
+	cutoff := time.Now().Add(-deletedGCAge)
+
+	it, err := db.NewPrefixIterator([]byte{KeyTypeSequence})
+	if err != nil {
+		return err
+	}
+	for it.Next() {
```
calmh: I'm iterating the sequence list because my initial thought was that I could stop once a too-new deleted item is reached. However there may be deleted items with times in the future because broken clocks and then we'd never gc, and I didn't change it to some other iteration afterwards...

lib/db/lowlevel.go
```diff
@@ -590,6 +627,123 @@ func (db *Lowlevel) gcBlocks() error {
...
+		gvBs, err := t.Get(gvKey)
+		if err != nil {
+			return err
+		}
+		var vl VersionList
+		if err := vl.Unmarshal(gvBs); err != nil {
+			return err
+		}
+
+		// We can clean this entry.
```
imsodin: Shouldn't we check that noone else needs this file either (i.e. it's deleted for everyone)? I guess it's mostly theoretical, but in principle a device could be offline since before the file was deleted for such a long time that GC kicks in, and whenever it comes online, it will propagate it's file instead of delete it.

AudriusButkevicius: Yes

lib/db/lowlevel.go
```diff
@@ -590,6 +627,123 @@ func (db *Lowlevel) gcBlocks() error {
...
+
+		dkey := deviceFileKey(it.Value())
+		f, ok, err := t.getFileTrunc(dkey, true)
+		if err != nil {
+			return err
+		}
+		if !ok {
+			continue
+		}
+		if !f.IsDeleted() || f.ModTime().After(cutoff) {
```
AudriusButkevicius: I think this modtime is meaningless, as simply deleting an old file would gc it immediately? 
Or are we giving it an artificial mtime of now once we notice it's deleted?

Relying on users clock feels very fragile. Could we look back lets say up to half of the sequence numbers available or something like that?

calmh: So, I *thought* we set the modtime to the time of deletion, but indeed we do not. That makes this whole thing a bit fraught. Even limiting to a fraction of the sequence space would do no good as we can always have the "rm -rf *" situation and then we should retain all the items even though they are all deletes. Back to the drawing board I guess. I could add a deletedTime attribute and let this kick in in six months, I could have the GC sweep keep entries for (date, max sequence) so we can look up a date six months in the past and only scan up to that sequence number ...

lib/db/lowlevel.go
calmh: Yes, I even did this to begin with. It seems reasonable to do so. In practice, in *complicated* setups, it seems there is always this device that's still in the config even though it was decommisioned six months ago and who has no modern index entries... You could argue that this means this feature does not work and that's OK and I have no excellent counter argument to that.

AudriusButkevicius: Well surely they can just remove it?
We also have last seen for the device which we could use.

calmh: Yep. Solutions abound. This is the smaller problem I think, the other one is more serious.

lib/db/lowlevel.go
AudriusButkevicius: I think setting an artificial mtime is fine, but then just cross your fingers and hope that the clocks on all devices are ok, as this would become what other devices would decide to gc on.

PR #6303, opened 2020-01-31
lib/model: Add global request limiter (fixes #6302)

### LINKED ISSUE #6302: Avoid thundering herd issue by global request limiter
We have a per-connection request limiter today to limit the number of concurrent requests we handle for any given connection. However, there's no global equivalent. This is a problem in large setups. For example, if I have 250 connected devices and announce a change to a new large file, I'll immediately get 250 (num clients) * 16 MiB (block size) * 2 (defaultPullerPendingKib) requests, requiring 8 GiB to buffer. Then we need to marshal, TLS encrypt and send these, getting us easily up 32 GiB or so of memory usage. Rate limits and such will only make this worse as we will hold on to the memory for a long time while we slowly send out the responses.

We should have some sort of global request limiter that limits the number of in flight requests we process, in total, at any given moment.

calmh:
### Purpose

This adds a new config with the simple and concise name `maxConcurrentIncomingRequestKiB`. This limits how many bytes we have "in the air" in the form of response data being read and processed.

After some testing I think that not having this limiter is seldom a great idea and thus I propose a default value of 256 MiB for this new setting.

I also refactored the folder IO limiter to be a model/folder attribute instead of a package global.

### Testing

I set up one instance with a fakefs with a few gigs in it, and 200 other empty fakefs instances. The amount of data in the fakefs wasn't huge, files averaging 1 GiB so block sizes of either 512 KiB or 1 MiB -- far from the worst case.

Prior to this patch it uses ~ 11 GiB in steady state:

Screenshot 2020-01-31 at 13 37 28

With the 256 meg setting we get ~ 1 GiB:

Screenshot 2020-01-31 at 13 47 44

Note that the upload rate didn't really degrade either. This is obviously heavily CPU limited, with 201 instances running full out on my little desktop...

I suspect that in most cases it's not a huge win to have more things in flight than this, and we should strive to behave responsibly on normal computers, hence the default. But it's of course tunable if the extreme setups there want to experiment.


lib/config/optionsconfiguration.go
```diff
@@ -175,3 +177,26 @@ func (opts OptionsConfiguration) MaxFolderConcurrency() int {
 	// of writing is two, 95-percentile at 12 folders.)
 	return 4 // https://xkcd.com/221/
 }
+
+func (opts OptionsConfiguration) MaxConcurrentIncomingRequestKiB() int {
```
AudriusButkevicius: In practice this will never be less than megabytes, because  minAllowed is MaxBlockSize * 2 which is 32mb, so this should be MiB.

calmh: We have a bunch of similar values already though and they are all in KiB

lib/model/model.go
```diff
@@ -177,38 +183,39 @@ func NewModel(cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersio
...
+		folderRunnerTokens:   make(map[string][]suture.ServiceToken),
+		folderVersioners:     make(map[string]versioner.Versioner),
+		conn:                 make(map[protocol.DeviceID]connections.Connection),
+		connRequestLimiters:  make(map[protocol.DeviceID]*byteSemaphore),
+		closed:               make(map[protocol.DeviceID]chan struct{}),
+		helloMessages:        make(map[protocol.DeviceID]protocol.HelloResult),
+		deviceDownloads:      make(map[protocol.DeviceID]*deviceDownloadState),
+		remotePausedFolders:  make(map[protocol.DeviceID][]string),
+		fmut:                 sync.NewRWMutex(),
+		pmut:                 sync.NewRWMutex(),
```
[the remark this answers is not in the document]
calmh: Fixd in followup

PR #6309, opened 2020-02-05
cmd/syncthing, lib/locations: Separate data and config dirs (fixes #4924)

### LINKED ISSUE #4924: Move index db to $XDG_DATA_HOME/syncthing/

imsodin:
### Purpose

Doesn't change anything for existing users. Neither does it change anything for windows and mac users. For linux users, if there is no db at the "usual place" and there's a relevant xdg env var or the directory `~/.local/share/syncthing` in place, put db and logs there. Also there's two new flags `-config` and `-data` accompanying `-home` to set separate dirs for config/keys and db/logs.

### Testing

My test system picked the db up at a new location with command line flags.

### DIFF
--- a/build.go
+++ b/build.go
@@ -364,12 +364,12 @@ func test(pkgs ...string) {
 
 	args := []string{"test", "-short", "-timeout", timeout, "-tags", "purego"}
 
-	if runtime.GOARCH == "amd64" {
-		switch runtime.GOOS {
-		case "darwin", "linux", "freebsd": // , "windows": # See https://github.com/golang/go/issues/27089
-			args = append(args, "-race")
-		}
-	}
+	// if runtime.GOARCH == "amd64" {
+	// 	switch runtime.GOOS {
+	// 	case "darwin", "linux", "freebsd": // , "windows": # See https://github.com/golang/go/issues/27089
+	// 		args = append(args, "-race")
+	// 	}
+	// }
 
 	if coverage {
 		args = append(args, "-covermode", "atomic", "-coverprofile", "coverage.txt", "-coverpkg", strings.Join(pkgs, ","))
--- a/build.go
+++ b/build.go
@@ -364,12 +364,12 @@ func test(pkgs ...string) {
 
 	args := []string{"test", "-short", "-timeout", timeout, "-tags", "purego"}
 
-	// if runtime.GOARCH == "amd64" {
-	// 	switch runtime.GOOS {
-	// 	case "darwin", "linux", "freebsd": // , "windows": # See https://github.com/golang/go/issues/27089
-	// 		args = append(args, "-race")
-	// 	}
-	// }
+	if runtime.GOARCH == "amd64" {
+		switch runtime.GOOS {
+		case "darwin", "linux", "freebsd": // , "windows": # See https://github.com/golang/go/issues/27089
+			args = append(args, "-race")
+		}
+	}
 
 	if coverage {
 		args = append(args, "-covermode", "atomic", "-coverprofile", "coverage.txt", "-coverpkg", strings.Join(pkgs, ","))
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -284,16 +284,24 @@ func main() {
 		osutil.HideConsole()
 	}
 
-	// Not set as default above because the string can be really long.
+	// Not set as default above because the strings can be really long.
 	var err error
-	if options.homeDir != "" {
+	homeSet := options.homeDir != ""
+	confSet := options.confDir != ""
+	dataSet := options.dataDir != ""
+	switch {
+	case dataSet != confSet:
+		err = errors.New("Either both or none of -conf and -data must be given, use -home to set both at once.")
+	case homeSet && dataSet:
+		err = errors.New("Option -home must not be used together with -conf and -data.")
+	case homeSet:
 		if err = setLocation(options.homeDir, locations.ConfigBaseDir); err == nil {
 			err = setLocation(options.homeDir, locations.DataBaseDir)
 		}
-	} else if options.confDir != "" {
-		err = setLocation(options.confDir, locations.ConfigBaseDir)
-	} else if options.dataDir != "" {
-		err = setLocation(options.dataDir, locations.DataBaseDir)
+	case dataSet:
+		if err = setLocation(options.confDir, locations.ConfigBaseDir); err == nil {
+			err = setLocation(options.dataDir, locations.DataBaseDir)
+		}
 	}
 	if err != nil {
 		l.Warnln(err)
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -291,9 +291,9 @@ func main() {
 	dataSet := options.dataDir != ""
 	switch {
 	case dataSet != confSet:
-		err = errors.New("Either both or none of -conf and -data must be given, use -home to set both at once.")
+		err = errors.New("either both or none of -conf and -data must be given, use -home to set both at once")
 	case homeSet && dataSet:
-		err = errors.New("Option -home must not be used together with -conf and -data.")
+		err = errors.New("-home must not be used together with -conf and -data")
 	case homeSet:
 		if err = setLocation(options.homeDir, locations.ConfigBaseDir); err == nil {
 			err = setLocation(options.homeDir, locations.DataBaseDir)
@@ -304,7 +304,7 @@ func main() {
 		}
 	}
 	if err != nil {
-		l.Warnln(err)
+		l.Warnln("Command line options:", err)
 		os.Exit(syncthing.ExitError.AsInt())
 	}
 
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -65,6 +65,12 @@ I.e. to prefix each log line with date and time, set -logflags=3 (1 + 2 from
 above). The value 0 is used to disable all of the above. The default is to
 show time only (2).
 
+Logging always happens to the command line (stdout) and optionally to the
+file at the path specified by -logfile=path. In addition to an explicity path,
+the special values "default" and "-" may be used. The former logs to
+DATADIR/syncthing.log (see -data-dir), which is the default on windows, and the
+latter only to stdout, no file, which is the default anywhere else.
+
 
 Development Settings
 --------------------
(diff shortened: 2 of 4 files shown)


cmd/syncthing/main.go
```diff
@@ -269,20 +284,20 @@ func main() {
...
-				l.Warnln("Failed to make options path absolute:", err)
-				os.Exit(syncthing.ExitError.AsInt())
-			}
-		}
-		if err := locations.SetBaseDir(locations.ConfigBaseDir, options.confDir); err != nil {
-			l.Warnln(err)
-			os.Exit(syncthing.ExitError.AsInt())
+	// Not set as default above because the string can be really long.
+	var err error
+	if options.homeDir != "" {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

cmd/syncthing/main.go
```diff
@@ -269,20 +284,20 @@ func main() {
...
+		if err = setLocation(options.homeDir, locations.ConfigBaseDir); err == nil {
+			err = setLocation(options.homeDir, locations.DataBaseDir)
 		}
+	} else if options.confDir != "" {
+		err = setLocation(options.confDir, locations.ConfigBaseDir)
+	} else if options.dataDir != "" {
```
calmh: Only check -data if I didn't give -config or -home? It feels to me these should take effect in order, first -home and then -config/-data independently

lib/locations/locations.go
```diff
@@ -123,23 +131,40 @@ func defaultConfigDir() string {
...
+				if _, err := os.Lstat(xdgData); err != nil {
+					return config
+				}
+			} else {
+				xdgData = strings.SplitN(xdgDatas, ":", 2)[0]
+			}
+		}
+		// Only use XDG location if no database exists at "normal" location
+		if _, err := os.Lstat(filepath.Join(home, "syncthing", databaseDirname)); os.IsNotExist(err) {
+			return xdgData
```
calmh: Returns empty if there is no data dir yet and the XDG stuff is unset

lib/locations/locations.go
```diff
@@ -123,23 +131,40 @@ func defaultConfigDir() string {
...
+func defaultDataDir(home, config string) string {
+	switch runtime.GOOS {
+	case "windows", "darwin":
+		return config
+
+	default:
+		xdgData := os.Getenv("XDG_DATA_HOME")
+		if xdgData == "" {
+			if xdgDatas := os.Getenv("XDG_DATA_DIRS"); xdgDatas == "" {
+				xdgData = filepath.Join(home, ".local/share/syncthing")
```
calmh: Logic is convoluted, I'd rather linearize this a bit

```
// first check for the existing traditional location
if _, err := os.Lstat(traditional location); err == nil {
  // ...
}

// go hunting for an xdg dir
xdgData := os.Getenv("XDG_DATA_HOME")
if xdgData == "" {
  // split and grab xdg_data_dirs
}
if xdgData != "" {
  // return the xdg dir we found
}

try for the existence of the default XDG dir and return that if it exists

return traditional place
```

cmd/syncthing/main.go
```diff
@@ -269,21 +284,29 @@ func main() {
...
-				os.Exit(syncthing.ExitError.AsInt())
-			}
+	// Not set as default above because the strings can be really long.
+	var err error
+	homeSet := options.homeDir != ""
+	confSet := options.confDir != ""
+	dataSet := options.dataDir != ""
+	switch {
+	case dataSet != confSet:
+		err = errors.New("Either both or none of -conf and -data must be given, use -home to set both at once.")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

cmd/syncthing/main.go
```diff
@@ -269,21 +284,29 @@ func main() {
...
+	case homeSet && dataSet:
+		err = errors.New("Option -home must not be used together with -conf and -data.")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

lib/locations/locations.go
```diff
@@ -123,23 +131,42 @@ func defaultConfigDir() string {
...
+		return config
+
+	default:
+		// If a database exists at the "normal" location, use that anyway.
+		if _, err := os.Lstat(filepath.Join(home, "syncthing", databaseDirname)); err == nil {
+			return config
+		}
+		// Always use this env var, as it's explicitly set by the user
+		if xdgHome := os.Getenv("XDG_DATA_HOME"); xdgHome != "" {
+			return xdgHome
```
calmh: So `XDG_DATA_HOME` here will be something like `~/.local/share` which is not where we should store the database, we should use a `syncthing` subdir? (Same for xdgDefault stuff below I think)

cmd/syncthing/main.go
calmh: I think these could be regular error-style strings and context added in the `Warnln("Command line options:", err)`

etc/linux-desktop/syncthing-start.desktop
```diff
@@ -2,7 +2,7 @@
 Name=Start Syncthing
 GenericName=File synchronization
 Comment=Starts the main syncthing process in the background.
-Exec=/usr/bin/syncthing -no-browser -logfile=~/.config/syncthing/syncthing.log
+Exec=/usr/bin/syncthing -no-browser -logfile=""
```
calmh: This is a really surprising way to say "logfile in the standard location" :/

lib/locations/locations.go
```diff
@@ -123,23 +131,42 @@ func defaultConfigDir() string {
...
+// config directory but might be something else.
+func defaultDataDir(home, config string) string {
+	switch runtime.GOOS {
+	case "windows", "darwin":
+		return config
+
+	default:
+		// If a database exists at the "normal" location, use that anyway.
+		if _, err := os.Lstat(filepath.Join(home, "syncthing", databaseDirname)); err == nil {
+			return config
```
calmh: This is weird though as we stat one location and return another. What if I have one instance running with the database in the default location and start another with `-home ~/asdasfd` (I do this all the time)

imsodin: It's not just weird, it's plain wrong. It should `filepath.Join(config, databaseDirname)` in the stat.

And `-home` overrules default locations, i.e. it doesn't matter what happens in this function in that case.

Also let me do 's/home/userHome/' - it's annoying, but it's definitely less error prone than calling a variable `home` that's not connected with `-home`.

etc/linux-desktop/syncthing-start.desktop
imsodin: Yes it is. Unfortunately go doesn't allow string flags with a default value for when the flag is used, but no argument given, so that's that. The alternative would be to add `-logfile-default`, but that's ugly too.

lib/locations/locations.go
calmh: Yes please

etc/linux-desktop/syncthing-start.desktop
calmh: I think it's fair to say that this is undiscoverable to the point of being useless, really. Do we need to have the "logfile, but I'm not going to say where, just put it at the default" functionality on Unix?

That whole "the strings are too long so we default  to blank and then fix it up later" might be causing more trouble than it's worth, too.

imsodin: I believe it's worth it. For example in case of DE autostarts through this desktop file, which I think is a very likely first contact point for a "non-technical" user when trying to setup something to run at start. I'd add another magic value `default` alongside `-` instead of an empty string. And maybe add a short explanation in the extended usage about logging. Does that sound more useful/less obscure?

calmh: Yeah, I think so.

cmd/syncthing/main.go
```diff
@@ -195,11 +203,13 @@ func defaultRuntimeOptions() RuntimeOptions {
...
-	if runtime.GOOS != "windows" {
-		// On non-Windows, we explicitly default to "-" which means stdout. On
-		// Windows, the blank options.logFile will later be replaced with the
-		// default path, unless the user has manually specified "-" or
-		// something else.
+	// On non-Windows, we explicitly default to "-" which means stdout. On
+	// Windows, the "default" options.logFile will later be replaced with the
+	// default path, unless the user has manually specified "-" or
+	// something else.
+	if runtime.GOOS == "windows" {
```
golangcibot: string `windows` has 4 occurrences, make it a constant (from `goconst`)

cmd/syncthing/main.go
```diff
@@ -65,6 +65,12 @@ I.e. to prefix each log line with date and time, set -logflags=3 (1 + 2 from
 above). The value 0 is used to disable all of the above. The default is to
 show time only (2).
 
+Logging always happens to the command line (stdout) and optionally to the
+file at the path specified by -logfile=path. In addition to an explicity path,
```
golangcibot: `explicity` is a misspelling of `explicitly` (from `misspell`)
```suggestion

```

cmd/syncthing/main.go
```diff
@@ -65,6 +65,12 @@ I.e. to prefix each log line with date and time, set -logflags=3 (1 + 2 from
...
+the special values "default" and "-" may be used. The former logs to
+DATADIR/syncthing.log (see -data-dir), which is the default on windows, and the
```
calmh:
```suggestion
DATADIR/syncthing.log (see -data-dir), which is the default on Windows, and the
```

build.go
```diff
@@ -364,12 +364,12 @@ func test(pkgs ...string) {
 
 	args := []string{"test", "-short", "-timeout", timeout, "-tags", "purego"}
 
-	if runtime.GOARCH == "amd64" {
-		switch runtime.GOOS {
-		case "darwin", "linux", "freebsd": // , "windows": # See https://github.com/golang/go/issues/27089
-			args = append(args, "-race")
-		}
-	}
+	// if runtime.GOARCH == "amd64" {
```
calmh: assuming mistake

calmh:
So this is, I think, a 1.5.0 type change. Do we want to get that in there now, or do we want to stabilize a little with a less nice and relaxed 1.4.1 bugfix release in the midterm?

PR #6310, opened 2020-02-06
lib/config, gui: Set unix socket permissions for GUI listen address (fixes #5979)

### LINKED ISSUE #5979: UNIX socket permission bits
I tried to use a UNIX socket with nginx, but the default permission bits for the socket are 644:

```
$ ls -l syncthing.sock 
srwxr-xr-x 1 syncthing syncthing 0 авг 27 15:32 syncthing.sock
```

Nginx uses `www-data` user, and that's why it can't use this UNIX socket and prints "Permission denied".

I can change these permission bits using umask 0007 to make it work, but this solution is not suitable in all cases.

It would be nice to have an option that controls socket permission bits (and maybe even a group).

kroppt:
### Purpose

This pull request solves #5979.
The problem is that a user may wish a unix socket to have specific permissions.
We can solve this by adding a configuration option for unix socket permissions.

### Testing

Navigate under the web UI to Actions > Settings > GUI and change the GUI Listen Address line to a local file. That will be created as a web socket. Check the permissions of this file.

### Screenshots

screen

### Documentation

https://github.com/syncthing/docs/pull/492

### Tasklist
- [X] Add basic implementation showcasing solution efficacy
- [x] Add configuration option
- [x] Add UI for configuration option
- [x] Use configuration option in code
- [x] Update man pages
- [X] Add documentation for new option
- [X] Finish PR information

### DIFF
--- a/gui/default/syncthing/settings/settingsModalView.html
+++ b/gui/default/syncthing/settings/settingsModalView.html
@@ -172,6 +172,8 @@
               </div>
             </div>
             <div class="col-md-6">
+              <label translate>UNIX Permissions</label>
+              <input id="UnixPerms" class="form-control" type="text" ng-model="tmpGUI.unixPerms" />
             </div>
           </div>
         </div>
--- a/gui/default/syncthing/settings/settingsModalView.html
+++ b/gui/default/syncthing/settings/settingsModalView.html
@@ -172,7 +172,7 @@
               </div>
             </div>
             <div class="col-md-6">
-              <div ng-if="tmpGUI.address.startsWith('/')" class="form-group" ng-class="{'has-error': settingsEditor.UnixSocketPermissions.$invalid && settingsEditor.UnixSocketPermissions.$dirty}">
+              <div ng-if="tmpGUI.address.startsWith('/') || tmpGUI.address.startsWith('unix://') || tmpGUI.address.startsWith('unixs://')" class="form-group" ng-class="{'has-error': settingsEditor.UnixSocketPermissions.$invalid && settingsEditor.UnixSocketPermissions.$dirty}">
                 <label translate>UNIX Permissions</label>
                 <input id="UnixSocketPermissions" name="UnixSocketPermissions" class="form-control" type="text" ng-model="tmpGUI.unixSocketPermissions" ng-pattern="/^0?[0-7]{0,3}$/" />
                 <p class="help-block" ng-show="settingsEditor.UnixSocketPermissions.$invalid" translate>
--- a/gui/default/syncthing/settings/settingsModalView.html
+++ b/gui/default/syncthing/settings/settingsModalView.html
@@ -172,8 +172,13 @@
               </div>
             </div>
             <div class="col-md-6">
-              <label translate>UNIX Permissions</label>
-              <input id="UnixPerms" class="form-control" type="text" ng-model="tmpGUI.unixPerms" />
+              <div class="form-group" ng-class="{'has-error': settingsEditor.UnixPerms.$invalid && settingsEditor.UnixPerms.$dirty}">
+                <label translate>UNIX Permissions</label>
+                <input id="UnixPerms" name="UnixPerms" class="form-control" type="text" ng-model="tmpGUI.unixPerms" ng-pattern="/^0?[0-7]{0,3}$/" />
+                <p class="help-block" ng-show="settingsEditor.UnixPerms.$invalid" translate>
+                  Enter up to three octal digits.
+                </p>
+              </div>
             </div>
           </div>
         </div>
--- a/gui/default/syncthing/settings/settingsModalView.html
+++ b/gui/default/syncthing/settings/settingsModalView.html
@@ -172,10 +172,10 @@
               </div>
             </div>
             <div class="col-md-6">
-              <div ng-if="tmpGUI.address.startsWith('/')" class="form-group" ng-class="{'has-error': settingsEditor.unixSocketPermissions.$invalid && settingsEditor.unixSocketPermissions.$dirty}">
+              <div ng-if="tmpGUI.address.startsWith('/')" class="form-group" ng-class="{'has-error': settingsEditor.UnixSocketPermissions.$invalid && settingsEditor.UnixSocketPermissions.$dirty}">
                 <label translate>UNIX Permissions</label>
-                <input id="unixSocketPermissions" name="unixSocketPermissions" class="form-control" type="text" ng-model="tmpGUI.unixSocketPermissions" ng-pattern="/^0?[0-7]{0,3}$/" />
-                <p class="help-block" ng-show="settingsEditor.unixSocketPermissions.$invalid" translate>
+                <input id="UnixSocketPermissions" name="UnixSocketPermissions" class="form-control" type="text" ng-model="tmpGUI.unixSocketPermissions" ng-pattern="/^0?[0-7]{0,3}$/" />
+                <p class="help-block" ng-show="settingsEditor.UnixSocketPermissions.$invalid" translate>
                   Enter up to three octal digits.
                 </p>
               </div>
--- a/gui/default/syncthing/settings/settingsModalView.html
+++ b/gui/default/syncthing/settings/settingsModalView.html
@@ -172,7 +172,7 @@
               </div>
             </div>
             <div class="col-md-6">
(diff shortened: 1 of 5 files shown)


AudriusButkevicius:
This doesn't add configuration, it forces the socket to be with insecure permissions with no way to switch that off, so I don't think this will fly.

AudriusButkevicius:
Nvm, it's a wip with a checklist of things todo, should read the rest of the content before opening my mouth :man_facepalming:

gui/default/syncthing/settings/settingsModalView.html
```diff
@@ -172,6 +172,13 @@
               </div>
             </div>
             <div class="col-md-6">
+              <div class="form-group" ng-class="{'has-error': settingsEditor.UnixPerms.$invalid && settingsEditor.UnixPerms.$dirty}">
```
AudriusButkevicius: `ng-if="tmpGUI.address.startswith('unix://')"` ?

calmh:
This is an exceedingly niche thing to have in the config gui.

AudriusButkevicius:
Agree hence the conditional hide suggestion.

gui/default/syncthing/settings/settingsModalView.html
kroppt: Ya, thanks. In this case, though, only '/' passes address validation.

lib/config/guiconfiguration.go
```diff
@@ -59,6 +61,14 @@ func (c GUIConfiguration) Address() string {
 	return c.RawAddress
 }
 
+func (c GUIConfiguration) Permissions() os.FileMode {
```
AudriusButkevicius: The function name permissions is too generic. If we decide to add ldap permissioning to the gui, this name being taken would be grim.

gui/default/syncthing/settings/settingsModalView.html
```diff
@@ -172,6 +172,13 @@
               </div>
             </div>
             <div class="col-md-6">
+              <div ng-if="tmpGUI.address.startsWith('/')" class="form-group" ng-class="{'has-error': settingsEditor.UnixPerms.$invalid && settingsEditor.UnixPerms.$dirty}">
```
AudriusButkevicius: unix:// and unixs:// are also valid prefixes.

lib/config/guiconfiguration.go
```diff
@@ -9,12 +9,14 @@ package config
...
 	"net/url"
 	"os"
+	"strconv"
 	"strings"
 )
 
 type GUIConfiguration struct {
 	Enabled                   bool     `xml:"enabled,attr" json:"enabled" default:"true"`
 	RawAddress                string   `xml:"address" json:"address" default:"127.0.0.1:8384"`
+	UnixPermissions           string   `xml:"unixPerms,omitempty" json:"unixPerms"`
```
AudriusButkevicius: Please keep the config and json and code variable names the same, preferably verbose.

calmh: I'd even go all the way to `unixSocketPermissions` to avoid any confusion about where these unix permissions might apply.

lib/config/guiconfiguration.go
kroppt: Done.

lib/config/guiconfiguration.go
kroppt: Done.

gui/default/syncthing/settings/settingsModalView.html
kroppt: I can't get that to work, it tries using TCP. Testing through the UI also won't work, because the Regex that verifies the address does not allow such a prefix, so you cannot save the page. In the code, it only checks for '/' to determine whether it's a UNIX socket (unless the environment variable is used, in which case it resolves then checks for the `unix` prefix).

In order to comply, I would need to alter how the code understands `Address`. If you want me to do that, let me know and I will do it.

AudriusButkevicius: I would add the code to check for prefix, even if it's not possible, so that when the regexp is modified, it's not overlooked.

Also, what happens if you put unix:// in the config by hand and then open the UI? Does it get removed? If not, then perhaps we should adjust the regexp.

Also, you suggest that it tries unix:// as TCP, yet we have code that explicitly handles unix://, so I am surprised it does not work.

kroppt: Fair point about adding it.
If it is using config settings for address, then it only checks `/`.
If it is using environment variables, it checks for `unix` in the scheme.

gui/default/syncthing/settings/settingsModalView.html
```diff
@@ -172,6 +172,13 @@
               </div>
             </div>
             <div class="col-md-6">
+              <div ng-if="tmpGUI.address.startsWith('/') || tmpGUI.address.startsWith('unix://') || tmpGUI.address.startsWith('unixs://')" class="form-group" ng-class="{'has-error': settingsEditor.UnixSocketPermissions.$invalid && settingsEditor.UnixSocketPermissions.$dirty}">
```
imsodin: (entirely optional) That's long enough that it could warrant a `isUnix(address)` util function.

gui/default/syncthing/settings/settingsModalView.html
```diff
@@ -172,6 +172,13 @@
...
+                <label translate>UNIX Permissions</label>
```
imsodin: I would have expected the `Unix` kind of capitalisation", but I am not sure.

lib/api/api.go
```diff
@@ -187,6 +187,16 @@ func (s *service) getListener(guiCfg config.GUIConfiguration) (net.Listener, err
 		return nil, err
 	}
 
+	if guiCfg.Network() == "unix" && guiCfg.SocketPermissions() != 0 {
+		// We should error if this fails under the assumption that these permissions are
+		// required for operation. If the configuration option isn't correctly formatted,
+		// then we will ignore it.
```
imsodin: I don't see anything about formatting here, I guess that happens within `.SocketPermissions()`? In that case move the second sentence there (or just drop it).

lib/config/guiconfiguration.go
```diff
@@ -9,12 +9,14 @@ package config
...
 	"net/url"
 	"os"
+	"strconv"
 	"strings"
 )
 
 type GUIConfiguration struct {
 	Enabled                   bool     `xml:"enabled,attr" json:"enabled" default:"true"`
 	RawAddress                string   `xml:"address" json:"address" default:"127.0.0.1:8384"`
+	UnixSocketPermissions     string   `xml:"unixSocketPermissions,omitempty" json:"unixSocketPermissions"`
```
imsodin: The usual pattern for fields that shouldn't be accessed directly is prepending `Raw`, d.h. field is `RawUnixSocketPermissions` and the method below `UnixSocketPermissions`.

man/syncthing-config.5
```diff
@@ -607,6 +607,10 @@ If the address is an absolute path it is interpreted as the path to a UNIX socke
 (Added in v0.14.52.)
 .UNINDENT
 .TP
+.B unixSocketPermissions
```
imsodin: That's autogenerated from the documentation repo, and I think you already have a PR line up for that - i.e. no need to modify the manpage here.

gui/default/syncthing/settings/settingsModalView.html
kroppt: Done.

gui/default/syncthing/settings/settingsModalView.html
kroppt: I basically followed how it showed up already in documentation (except in the code). Let me know if you want it changed anywhere.

lib/api/api.go
kroppt: Done.

man/syncthing-config.5
kroppt: Done.

lib/config/guiconfiguration.go
kroppt: Done.

AudriusButkevicius:
Thanks for your time and patience.

kroppt:
Not sure if it was forgotten, but the docs PR hasn't been reviewed / merged.

PR #6318, opened 2020-02-11
lib/fs, lib/model: Rewrite RecvOnly tests

calmh:
During some other work I discovered these tests weren't great, so I've
rewritten them to be a little better. The real changes here are:

- Don't play games with not starting the folder and such, and don't
  construct a fake folder instance -- just use the one the model has. The
  folder starts and scans but the folder contents are empty at this point
  so that's fine.

- Use a fakefs instead of a temp dir.

- To support the above, implement a fakefs option `?content=true` to
  make the fakefs actually retain written content. Use sparingly,
  obviously, but it means the fakefs can usually be used instead of an
  on disk real directory.

### DIFF
--- a/lib/fs/fakefs_test.go
+++ b/lib/fs/fakefs_test.go
@@ -896,6 +896,28 @@ func testFakeFSCreateInsens(t *testing.T, fs Filesystem) {
 	assertDir(t, fs, "/", []string{"FOO"})
 }
 
+func TestReadWriteContent(t *testing.T) {
+	fs := newFakeFilesystem("foo?content=true")
+	fd, err := fs.Create("file")
+	if err != nil {
+		t.Fatal()
+	}
+
+	fd.Write([]byte("foo"))
+	fd.WriteAt([]byte("bar"), 5)
+	expected := []byte("foo\x00\x00bar")
+
+	buf := make([]byte, 1024)
+	n, err := fd.ReadAt(buf, 1) // note offset one byte
+	if n != len(expected)-1 {
+		t.Fatal("wrong number of bytes read")
+	}
+	if !bytes.Equal(buf[:n], expected[1:]) {
+		fmt.Printf("%d %q\n", n, buf[:n])
+		t.Error("wrong data in file")
+	}
+}
+
 func cleanup(fs Filesystem) error {
 	filenames, _ := fs.DirNames("/")
 	for _, filename := range filenames {
--- a/lib/fs/fakefs_test.go
+++ b/lib/fs/fakefs_test.go
@@ -911,7 +911,7 @@ func TestReadWriteContent(t *testing.T) {
 	}
 	expected := []byte("foo\x00\x00bar")
 
-	buf := make([]byte, 1024)
+	buf := make([]byte, len(expected)-1)
 	n, err := fd.ReadAt(buf, 1) // note offset one byte
 	if err != nil {
 		t.Fatal(err)
--- a/lib/fs/fakefs_test.go
+++ b/lib/fs/fakefs_test.go
@@ -900,15 +900,22 @@ func TestReadWriteContent(t *testing.T) {
 	fs := newFakeFilesystem("foo?content=true")
 	fd, err := fs.Create("file")
 	if err != nil {
-		t.Fatal()
+		t.Fatal(err)
 	}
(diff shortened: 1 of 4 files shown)


lib/fs/fakefs_test.go
```diff
@@ -896,6 +896,28 @@ func testFakeFSCreateInsens(t *testing.T, fs Filesystem) {
...
 }
 
+func TestReadWriteContent(t *testing.T) {
+	fs := newFakeFilesystem("foo?content=true")
+	fd, err := fs.Create("file")
+	if err != nil {
+		t.Fatal()
+	}
+
+	fd.Write([]byte("foo"))
```
golangcibot: Error return value of `fd.Write` is not checked (from `errcheck`)

lib/fs/fakefs_test.go
```diff
@@ -896,6 +896,28 @@ func testFakeFSCreateInsens(t *testing.T, fs Filesystem) {
...
+	fd.WriteAt([]byte("bar"), 5)
```
golangcibot: Error return value of `fd.WriteAt` is not checked (from `errcheck`)

lib/model/folder_recvonly_test.go
```diff
@@ -260,11 +247,11 @@ func TestRecvOnlyUndoChanges(t *testing.T) {
 
 	// Remove the file again and undo the modification
 
-	testOs.Remove(file)
-	must(t, ioutil.WriteFile(filepath.Join(ffs.URI(), "knownDir/knownFile"), oldData, 0644))
-	folderFs.Chtimes("knownDir/knownFile", knownFiles[1].ModTime(), knownFiles[1].ModTime())
+	ffs.Remove(file)
```
golangcibot: Error return value of `ffs.Remove` is not checked (from `errcheck`)

lib/model/folder_recvonly_test.go
```diff
@@ -260,11 +247,11 @@ func TestRecvOnlyUndoChanges(t *testing.T) {
...
+	must(t, writeFile(ffs, "knownDir/knownFile", oldData, 0644))
+	ffs.Chtimes("knownDir/knownFile", knownFiles[1].ModTime(), knownFiles[1].ModTime())
```
golangcibot: Error return value of `ffs.Chtimes` is not checked (from `errcheck`)

lib/fs/fakefs_test.go
```diff
@@ -896,6 +896,28 @@ func testFakeFSCreateInsens(t *testing.T, fs Filesystem) {
...
+	expected := []byte("foo\x00\x00bar")
+
+	buf := make([]byte, 1024)
+	n, err := fd.ReadAt(buf, 1) // note offset one byte
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

lib/model/folder_recvonly_test.go
```diff
@@ -310,30 +297,38 @@ func setupKnownFiles(t *testing.T, ffs fs.Filesystem, data []byte) []protocol.Fi
...
-		},
-	}
-	m.fmut.RUnlock()
+	defer m.fmut.RUnlock()
+	f := m.folderRunners["ro"].(*receiveOnlyFolder)
 
 	return m, f
 }
+
+func writeFile(fs fs.Filesystem, filename string, data []byte, perm fs.FileMode) error {
```
golangcibot: `writeFile` - `perm` always receives `0644` (`420`) (from `unparam`)

imsodin:
Thanks! Those folder-not-running-just-adding shenanigans caused me some headaches before too.

calmh:
There is still another problem I'm bashing my head against now which is that there is a race on the `ctx` field in `Folder`... It's set  by `serve()` and read by `Scan()` which can run concurrently. Refactoring to pass `ctx` as a parameter instead to all functions and having a separate `done` channel is doable but the diff is large...

imsodin:
How is that possible? If I understand it correctly `Scan` first waits on `<-f.initialScanFinished`, which unblocks only after `f.ctx` is set in `serve`.

calmh:
You are right, it isn't possible, I was tricked by another problem I introduced myself.

PR #6239, opened 2019-12-27
all: Transactionalize db.FileSet (fixes #5952)

imsodin:
When pulling we go once over all needed files in the db, and then at various later points do lookups in the db. In the meantime changes may happen (e.g. that's the only explanation I have for #5952). This PR moves db access through `db.FileSet` to `db.Snapshot` which can be obtained from `db.FileSet`.  
The huge diff on lowlevel.go and transaction.go is just copying over the methods that created a transaction at the beginning to actually be methods on transactions themselves.

Maybe it would be worth keeping utility functions on `db.FileSet` that allow access without taking a snapshot. I would definitely make especially some testing code shorter, then again I think it has some merit being force to think about snapshots (potential concurrent db changes) when dealing with the db.

PR #6320, opened 2020-02-11
lib/db: Don't panic getting a snapshot when db is closed (fixes #6319)

### LINKED ISSUE #6319: panic: database is closed when creating snapshot
Sentry Issue: [SYNCTHING-Y7](https://sentry.syncthing.net/syncthing/syncthing/issues/986/?referrer=github_integration)

```

  File "C:/BuildAgent/work/174e136266f8a219/lib/db/set.go", line 199, in Snapshot
    panic(err)
  File "C:/BuildAgent/work/174e136266f8a219/lib/model/folder.go", line 359, in scanSubdirs
    snap := f.fset.Snapshot()
  File "C:/BuildAgent/work/174e136266f8a219/lib/model/folder.go", line 583, in scanTimerFired
    err := f.scanSubdirs(nil)
  File "C:/BuildAgent/work/174e136266f8a219/lib/model/folder.go", line 173, in serve
    f.scanTimerFired()
  File "C:/BuildAgent/work/174e136266f8a219/lib/util/utils.go", line 183, in func1
    fn(ctx)
  File "C:/BuildAgent/work/174e136266f8a219/lib/util/utils.go", line 247, in Serve
    err = s.serve(s.ctx)
  File "C:/BuildAgent/work/pkg/mod/github.com/thejerf/suture@v3.0.2+incompatible/supervisor.go", line 600, in func1

panic: database is closed
```

imsodin:
Regression introduced in #6239, thus RC material.

### DIFF
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -196,6 +196,9 @@ type Snapshot struct {
 func (s *FileSet) Snapshot() *Snapshot {
 	t, err := s.db.newReadOnlyTransaction()
 	if err != nil {
+		if backend.IsClosed(err) {
+			return nil
+		}
 		panic(err)
 	}
 	return &Snapshot{


lib/db/set.go
```diff
@@ -196,6 +196,9 @@ type Snapshot struct {
 func (s *FileSet) Snapshot() *Snapshot {
 	t, err := s.db.newReadOnlyTransaction()
 	if err != nil {
+		if backend.IsClosed(err) {
+			return nil
```
calmh: I expect we're just going to panic one later when we call a method on the nil snapshot instead?

calmh:
Nice rabbit hole you dug there, with the database closing stuff :D

lib/db/set.go
imsodin: No, we don't close the db until the model is stopped and the snapshot is only used there. It's the same as with `NewFileSet` (that rabbit hole wasn't dug by me for once as far as I remember :D ).

calmh: Well, if this method is called at all then surely its return value will be used immediately thereafter?

imsodin: Oh damn, so the panic shouldn't happen in the first place. I retract my statement and go back to thinking.

calmh: And *cough* #5671 *cough*

calmh:
Regardless, we need to move towards having error returns with clean exit paths for all database operations, so fixing this is necessary in the grand scheme of things.

imsodin:
Agreed. And maybe this might not even be strictly RC material: The only way I can explain the panic (it's a single one so far) is that the model/folder didn't shut down in time so `syncthing.App` went ahead with the shutdown anyway, causing the panic. Same thing can happen with the file-set, which was already in 1.3.4. So we could just live with it for now and let it fall under https://github.com/syncthing/syncthing/issues/5907.

calmh:
Hmm yes. If it's "just" a stray panic during exceptionally slow shutdown then I'm mostly OK with it. Lets let the RC simmer for a while and see how it looks.

PR #6325, opened 2020-02-11
lib/model: Consistent error return values for folder methods on model

imsodin:
~~#6272 makes sure not running folders don't result in a 404 on the rest api, however the method  `Model.DBSnapshot` introduced in #6239 doesn't look at why a folder isn't running, it just returns a folder missing error if it's not running. So Synctrayzor still presents an error message due to a 404: https://forum.syncthing.net/t/rest-db-status-returns-404-on-startup/14329/19~~  
Looks like #6239 isn't easily compatible with not returning a 404. At least it isn't doable with the fixes in this PR. The PR in itself is still useful.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -2320,10 +2320,11 @@ func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly 
 
 func (m *model) GetFolderVersions(folder string) (map[string][]versioner.FileVersion, error) {
 	m.fmut.RLock()
-	ver, ok := m.folderVersioners[folder]
+	err := m.checkFolderRunningLocked(folder)
+	ver := m.folderVersioners[folder]
 	m.fmut.RUnlock()
-	if !ok {
-		return nil, errFolderMissing
+	if err != nil {
+		return nil, err
 	}
 	if ver == nil {
 		return nil, errNoVersioner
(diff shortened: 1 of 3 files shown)


lib/model/model.go
```diff
@@ -840,10 +840,11 @@ func (m *model) Completion(device protocol.DeviceID, folder string) FolderComple
...
 func (m *model) DBSnapshot(folder string) (*db.Snapshot, error) {
 	m.fmut.RLock()
-	rf, ok := m.folderFiles[folder]
+	err := m.checkFolderRunningLocked(folder)
+	rf := m.folderFiles[folder]
 	m.fmut.RUnlock()
-	if !ok {
-		return nil, errFolderMissing
+	if err != nil {
+		return nil, err
```
calmh: The desired outcome here is that we return `errFolderNotRunning` (from checkFolderRunning) instead of the current `errFolderMissing`?

calmh: It feels like this thing might benefit from an actual API test

lib/model/model_test.go
```diff
@@ -3479,3 +3479,35 @@ func TestNewLimitedRequestResponse(t *testing.T) {
 		t.Error("Bytes weren't returned in a timely fashion")
 	}
 }
+
+func TestFolderAPIErrors(t *testing.T) {
+	wcfg, fcfg := tmpDefaultWrapper()
+	fcfg.Paused = true
+	wcfg.SetFolder(fcfg)
```
golangcibot: Error return value of `wcfg.SetFolder` is not checked (from `errcheck`)

lib/model/model.go
imsodin: Misunderstood your comment. The test I added is still useful in my opinion, but does nothing for the rest api. Which is indeed still returning a 404, and that needs a different workaround. I'll file a separate PR for that.

calmh:
I think `errFolderNotRunning` should probably go away, too. It's the "folder isn't in the model, but it's in the config, and it's not paused" sort of fallback error that should only happen due to race conditions. If we checked `m.folderCfgs` instead of the global config for the folder we should never see this inconsistency I think.

PR #6326, opened 2020-02-11
lib/model: Return empty summary on paused folders (ref #6272)

imsodin:
This now actually addresses the 404 on the rest api that makes trouble for SyncTrayzor (https://forum.syncthing.net/t/rest-db-status-returns-404-on-startup/14329). The behaviour is, that we return an empty folder summary in that case. I think that's not a good API and instead the folder summary service should have access to the db for paused folders as well (e.g. global and local state are still useful info then), but that's a much bigger change. This just makes sure the behaviour doesn't change.

I didn't add a rest api error, as it can't use the model package. However I added a test in the model that makes sure that no error is returned when getting a folder summary for a paused folder.

### DIFF
--- a/lib/model/folder_summary.go
+++ b/lib/model/folder_summary.go
@@ -15,6 +15,7 @@ import (
 	"github.com/thejerf/suture"
 
 	"github.com/syncthing/syncthing/lib/config"
+	"github.com/syncthing/syncthing/lib/db"
 	"github.com/syncthing/syncthing/lib/events"
 	"github.com/syncthing/syncthing/lib/protocol"
 	"github.com/syncthing/syncthing/lib/sync"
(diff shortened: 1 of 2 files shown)


lib/model/model_test.go
```diff
@@ -3479,3 +3479,16 @@ func TestNewLimitedRequestResponse(t *testing.T) {
 		t.Error("Bytes weren't returned in a timely fashion")
 	}
 }
+
+func TestSummaryPausedNoError(t *testing.T) {
+	wcfg, fcfg := tmpDefaultWrapper()
+	fcfg.Paused = true
+	wcfg.SetFolder(fcfg)
```
golangcibot: Error return value of `wcfg.SetFolder` is not checked (from `errcheck`)

lib/model/folder_summary.go
```diff
@@ -77,28 +78,36 @@ func (c *folderSummaryService) String() string {
...
-	if err != nil {
-		return nil, err
-	}
-
+	var local, global, need, ro db.Counts
+	var ourSeq, remoteSeq int64
 	errors, err := c.model.FolderErrors(folder)
+	if err == nil {
+		var snap *db.Snapshot
+		if snap, err = c.model.DBSnapshot(folder); err == nil {
```
calmh: release dat snap

calmh: Also, DBSnapshot still returns `errFolderMissing` for non-started folders does it not?

AudriusButkevicius: The snap is released if its acquired.

calmh: Now it is, yes, but it wasn't when I commented. :)

calmh:
Needs merge

PR #6328, opened 2020-02-11
etc: Use After=nework-online in systemd service

grandchild:
Also add "Wants=network-online.target"
https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

### Purpose

Syncthing often fails to start on Ubuntu18.04 because the network is not really up.

### Testing

I have edited the service file like in the PR and it does start later, after the network is up.

### DIFF
--- a/etc/linux-systemd/system/syncthing@.service
+++ b/etc/linux-systemd/system/syncthing@.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=Syncthing - Open Source Continuous File Synchronization for %I
 Documentation=man:syncthing(1)
-After=network.target
+After=network-online.target
+Wants=network-online.target
 
 [Service]
 User=%i


imsodin:
Syncthing does not require that the system is online to run.

calmh:
> Syncthing often fails to start on Ubuntu18.04 because the network is not really up.

What does such a failure to start look like? I also don't think we require any particular network connectivity, or shouldn't at least. I guess if you've configured the UI to bind to a specific IP address and that IP address doesn't exist yet this could be an issue, but that seems like a very local condition...

grandchild:
This is embarassing. It fails with
```
Feb 12 10:11:16 berl34793786a systemd[946]: syncthing@<user>.service: Failed to determine user credentials: Connection refused
Feb 12 10:11:16 berl34793786a systemd[946]: syncthing@<user>.service: Failed at step USER spawning /usr/bin/syncthing: Connection refused
```
which is probably because of sssd user management, or something. I will see if including `After=nss-user-lookup.target` fixes the issue, and think about fixing this or  opening a new PR, if it's something that is generalizable. Ordering after `nss-user-lookup.target` shouldn't break anything for users that don't used it even if that target isn't available.

Very sorry for the noise.

PR #6329, opened 2020-02-12
lib: Faster termination on exit (ref #6319)

imsodin:
> There's a steady trickle of "panic: database closed" in sentry...
https://github.com/syncthing/syncthing/issues/6319#issuecomment-585049339

https://sentry.syncthing.net/syncthing/syncthing/?environment=Candidate&query=panic%3A+database+is+closed

I had a look at the goroutines running at the time of the panic. They are all due to slow shutdown.  
In some scanning was blocking, where there's a logical place to check the context.  
One (maybe 2, don't remember) were during pulling. Here I don't think there's much we can do in the way of checking contexts. It looks like it just fired up pullers until the limit is reached, and those are then spending time on waiting for model locks or doing io, i.e. a context check would be too late.  
Then there is one that hangs in the connect loop. There are a few places where we loop over device or addresses doing something with them without checking the context in between.  
Then there's one where everything is still running, even the context aren't cancelled there. That just looks like proper freeze where I can make no sense of it at all. It's a single occurrence so I am fine with that.

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -360,6 +360,12 @@ func (s *service) connect(ctx context.Context) {
 		var seen []string
 
 		for _, deviceCfg := range cfg.Devices {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+
 			deviceID := deviceCfg.DeviceID
 			if deviceID == s.myID {
 				continue
@@ -378,6 +384,11 @@ func (s *service) connect(ctx context.Context) {
 
 			var addrs []string
 			for _, addr := range deviceCfg.Addresses {
+				select {
+				case <-ctx.Done():
+					return
+				default:
+				}
 				if addr == "dynamic" {
 					if s.discoverer != nil {
 						if t, err := s.discoverer.Lookup(deviceID); err == nil {
@@ -396,6 +407,11 @@ func (s *service) connect(ctx context.Context) {
 			dialTargets := make([]dialTarget, 0)
 
 			for _, addr := range addrs {
+				select {
+				case <-ctx.Done():
+					return
+				default:
+				}
 				// Use a special key that is more than just the address, as you might have two devices connected to the same relay
 				nextDialKey := deviceID.String() + "/" + addr
 				seen = append(seen, nextDialKey)
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -384,14 +384,9 @@ func (s *service) connect(ctx context.Context) {
 
 			var addrs []string
 			for _, addr := range deviceCfg.Addresses {
-				select {
-				case <-ctx.Done():
-					return
-				default:
-				}
 				if addr == "dynamic" {
 					if s.discoverer != nil {
-						if t, err := s.discoverer.Lookup(deviceID); err == nil {
+						if t, err := s.discoverer.Lookup(ctx, deviceID); err == nil {
 							addrs = append(addrs, t...)
 						}
 					}
@@ -407,11 +402,6 @@ func (s *service) connect(ctx context.Context) {
 			dialTargets := make([]dialTarget, 0)
 
 			for _, addr := range addrs {
-				select {
-				case <-ctx.Done():
-					return
-				default:
-				}
 				// Use a special key that is more than just the address, as you might have two devices connected to the same relay
 				nextDialKey := deviceID.String() + "/" + addr
 				seen = append(seen, nextDialKey)
--- a/lib/discover/cache_test.go
+++ b/lib/discover/cache_test.go
@@ -7,6 +7,7 @@
 package discover
 
 import (
+	"context"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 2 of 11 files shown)


lib/connections/service.go
```diff
@@ -378,6 +384,11 @@ func (s *service) connect(ctx context.Context) {
 
 			var addrs []string
 			for _, addr := range deviceCfg.Addresses {
+				select {
+				case <-ctx.Done():
+					return
+				default:
+				}
```
AudriusButkevicius: Lookup should probably take a context instead.

lib/connections/service.go
```diff
@@ -396,6 +407,11 @@ func (s *service) connect(ctx context.Context) {
 			dialTargets := make([]dialTarget, 0)
 
 			for _, addr := range addrs {
+				select {
+				case <-ctx.Done():
+					return
+				default:
+				}
 				// Use a special key that is more than just the address, as you might have two devices connected to the same relay
```
AudriusButkevicius: I don't think this loop does anything lengthy, the dial is the thing that takes time.

lib/model/folder.go
```diff
@@ -449,6 +449,12 @@ func (f *folder) scanSubdirs(subDirs []string) error {
 		return err
 	}
 
+	select {
+	case <-f.ctx.Done():
+		return f.ctx.Err()
+	default:
```
AudriusButkevicius: There exists check like this in the snap.WithPrefix..., does this one actually help?

lib/model/folder_sendrecv.go
```diff
@@ -1200,6 +1200,13 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
 	}()
 
 	for state := range in {
+		select {
+		case <-f.ctx.Done():
+			state.fail(errors.Wrap(f.ctx.Err(), "folder stopped"))
+			out <- state.sharedPullerState
+		default:
+		}
```
AudriusButkevicius: Again, the only lengthy procedure seems weak hashing, which already takes a context? Do you have evidence this helps?

calmh:
I don't think this is the cause for the "database failed to stop" stuff though. I can reproduce that by just starting syncthing with default config and Ctrl-C:ing it. Looking at the routines (adding a panic after the failed to stop message) I don't see anything that should have an open database transaction. I suspect we might just be leaking something in the close waitgroup.

lib/connections/service.go
imsodin: I simply didn't check what it does exactly, I just added it based on the loop internals being lengthy (as in lines of code lengthy) and thus wanting to be safe.

lib/model/folder.go
imsodin: Trying to spell out the reason why I added this all arguments dissolved and they only remaining reason is that a panic on sentry happened at this point, which is not an argument at all. Rechecking the goroutines shows the problem is with lib/connections -> removing.

lib/discover/cache_test.go
```diff
@@ -96,7 +99,7 @@ func TestCacheSlowLookup(t *testing.T) {
 	// Start a lookup, which will take at least a second
 
 	t0 := time.Now()
-	go c.Lookup(protocol.LocalDeviceID)
+	go c.Lookup(context.Background(), protocol.LocalDeviceID)
```
golangcibot: Error return value of `c.Lookup` is not checked (from `errcheck`)

imsodin:
I contextified the `discover.FinderService.Lookup` function, which on another look at the goroutines in the reported panics is the main cause of delayed shutdown together with looping over all devices and calling `model.Connection` in lib/connections.

> I don't think this is the cause for the "database failed to stop" stuff though. I can reproduce that by just starting syncthing with default config and Ctrl-C:ing it. Looking at the routines (adding a panic after the failed to stop message) I don't see anything that should have an open database transaction. I suspect we might just be leaking something in the close waitgroup.

It definitely doesn't cover all cases (I mentioned one about pulling above), maybe not yours either. In the reported panics (e.g. https://crash.syncthing.net/report/2193b8026662ebea725e4b73f1147e98e07e3f986f0075704213c8eda5ec4bd8) it looks like suture fails to shutdown the connection service, and thus also other services stay running, which then panic ( in the linked example it's the index sender, in others a folder). That's what I am trying to address here.

lib/discover/global.go
```diff
@@ -403,3 +403,28 @@ func (e *errorHolder) Error() error {
...
+
+func (c *contextClient) Get(ctx context.Context, url string) (*http.Response, error) {
+	// For <go1.13 compatibility, use the following commented line once that
+	// isn't required anymore.
+	// req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Cancel = ctx.Done()
```
golangcibot: SA1019: req.Cancel is deprecated: Set the Request's context with NewRequestWithContext instead. If a Request's Cancel field and context are both set, it is undefined whether Cancel is respected.  (from `staticcheck`)

calmh:
I found two other places that would prevent database shutdown - new GC runner, and the metadata recalculator. The latter one runs on first startup and explains why my shutdown doesn't work then.

PR #6333, opened 2020-02-12
lib/db: Use flags from arg not LocalFlags() updating meta

imsodin:
We store metadata in single flag buckets. On adding files to meta we do it right and just use the flags passed as arguments, on removing we do however use the localflags from the file. That's what is fixed in the first commit (single line). The second commit renames `flags` to `flag` in places where it should by just a single bit and adds a check to `countsPtr` that it is indeed the case.

Unfortunately (or not) I think this has no real-world implications, as we only ever have one local flag set on a file. If I am wrong on that, this might actually fix inconsistent local states.

### DIFF
--- a/lib/db/meta.go
+++ b/lib/db/meta.go
@@ -28,8 +28,8 @@ type metadataTracker struct {
 }
 
 type metaKey struct {
-	dev   protocol.DeviceID
-	flags uint32
+	dev  protocol.DeviceID
+	flag uint32
 }
 
 func newMetadataTracker() *metadataTracker {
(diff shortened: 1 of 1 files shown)


lib/db/meta.go
```diff
@@ -196,8 +200,8 @@ func (m *metadataTracker) removeFile(dev protocol.DeviceID, f FileIntf) {
 	}
 }
 
-func (m *metadataTracker) removeFileLocked(dev protocol.DeviceID, flags uint32, f FileIntf) {
-	cp := m.countsPtr(dev, f.FileLocalFlags())
```
calmh: I’m not sure about this change. Like, if we have a recv only folder we might add a bunch of files with the local change bit. Later we change the folder type, the files are going to be reprocessed. Should we not remove them with the same bits they were added?

imsodin: Yes and we do by calling `removeFileLocked` with all the bits of `f.FileLocalFlags` separately, see https://github.com/syncthing/syncthing/blob/713ea31f9bf807c8f91d334f8c8758db14ce36a8/lib/db/meta.go#L197. Same when adding files https://github.com/syncthing/syncthing/blob/713ea31f9bf807c8f91d334f8c8758db14ce36a8/lib/db/meta.go#L143, where the locked variant already does use the `flags` argument: https://github.com/syncthing/syncthing/blob/713ea31f9bf807c8f91d334f8c8758db14ce36a8/lib/db/meta.go#L165

Currently `flags` is entirely unused in `removeFileLocked`.

PR #6336, opened 2020-02-13
lib/db: Recover sequence number and metadata on startup (fixes #6335)

### LINKED ISSUE #6335: Crash or hard shutdown can case database inconsistency, out of sync
Database updates are tracked by sequence number. When Syncthing starts after a crash there is a risk that the sequence number becomes out of sync and old sequence numbers are reused. This results in missed updates to peers and devices mysteriously out of sync.

calmh:
### Purpose

If we crashed after writing new file entries but before updating metadata in the database the sequence number and metadata will be wrong. This fixes that.

### Testing

I have an elegant stress test for this, but it's too large to fit in the margins of this PR. I'm going to improve that a bit more (still bug hunting) and then commit it separately.

Testing it separately is annoying as we need to reach in and mess up the metadata which is tricky. :/

### DIFF
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -146,6 +146,7 @@ func (s *FileSet) verifyLocalSequence() (ok bool) {
 
 	curSeq := s.meta.Sequence(protocol.LocalDeviceID)
 	snap := s.Snapshot()
+	ok = true
 	snap.WithHaveSequence(curSeq, func(fi FileIntf) bool {
 		if fi.SequenceNo() > curSeq {
 			// Out of sync, we will recalculate the folder.
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -148,7 +148,7 @@ func (s *FileSet) verifyLocalSequence() bool {
 
 	snap := s.Snapshot()
 	ok := true
-	snap.WithHaveSequence(curSeq, func(fi FileIntf) bool {
+	snap.WithHaveSequence(curSeq+1, func(fi FileIntf) bool {
 		ok = false // we got something, which we should not have
 		return false
 	})
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -146,14 +146,6 @@ func (s *FileSet) verifyLocalSequence() bool {
 
 	curSeq := s.meta.Sequence(protocol.LocalDeviceID)
 
-	first, err := s.db.keyer.GenerateSequenceKey(nil, folder, curSeq)
-	it, err := s.db.NewRangeIterator(first, nil)
-	if err != nil {
-		return false
-	}
-	defer it.Release()
-	return !it.Next()
-
 	snap := s.Snapshot()
 	ok := true
 	snap.WithHaveSequence(curSeq, func(fi FileIntf) bool {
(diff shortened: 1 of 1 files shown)


lib/db/set.go
```diff
@@ -123,6 +131,35 @@ func (s *FileSet) recalcCounts() error {
...
+	//
+	// Note that we can have the same thing happen for remote devices but
+	// there it's not a problem -- we'll simply advertise a lower sequence
+	// number than we've actually seen and receive some duplicate updates
+	// and then be in sync again.
+
+	curSeq := s.meta.Sequence(protocol.LocalDeviceID)
+	snap := s.Snapshot()
+	ok = true
+	snap.WithHaveSequence(curSeq, func(fi FileIntf) bool {
```
imsodin: To not have to unmarshal all those fileinfo, you could directly access the db with an iterator starting at curSeq, something like this:

```
	first, err := s.db.keyer.GenerateSequenceKey(nil, folder, curSeq)
	it, err := s.db.NewRangeIterator(first, nil)
	return !it.Next()
```

lib/db/set.go
```diff
@@ -123,6 +131,35 @@ func (s *FileSet) recalcCounts() error {
...
+	// not yet written new metadata to disk.
+	//
+	// Note that we can have the same thing happen for remote devices but
+	// there it's not a problem -- we'll simply advertise a lower sequence
+	// number than we've actually seen and receive some duplicate updates
+	// and then be in sync again.
+
+	curSeq := s.meta.Sequence(protocol.LocalDeviceID)
+	snap := s.Snapshot()
+	ok = true
```
imsodin: You could just drop the named return argument, in which case that couldn't have gone forgotten before

lib/db/set.go
calmh: Yeah... That's kind of lowlevel though, and we anyway only ever iterate one iteration at most.

imsodin: Right, I mistakenly assumed `WithHaveSequence` iterates starting from the beginning, which it obviously doesn't. Sorry for the noise.

lib/db/set.go
```diff
@@ -80,25 +80,33 @@ func NewFileSet(folder string, fs fs.Filesystem, db *Lowlevel) *FileSet {
...
 		l.Infof("Stored folder metadata for %q is %v old; recalculating", folder, age)
-		if err := s.recalcCounts(); backend.IsClosed(err) {
-			return nil
-		} else if err != nil {
-			panic(err)
-		}
+		goto recalcMeta
+	}
+
+	return &s
```
imsodin: This just made me blink, then think about the goto, think again, and then get it. Personally I'd rather use a function variable defined above like

```
recalc := func() *FileSet {
	if err := s.recalcMeta(); backend.IsClosed(err) {
		return nil
	} else if err != nil {
		panic(err)
	}
	return &s
}
```

and replace the gotos with calling and returning that.

AudriusButkevicius:
Orthogonal question. Should metadata updates happen in the same "transaction" as the entries landing? Or is that too gnarly to push through?

calmh:
We should, it's doable, and there is another bug on precisely that, so hang tight a moment...

calmh:
Actually that can be a separate one

PR #6337, opened 2020-02-13
lib/db: Keep metadata better in sync (ref #6335)

calmh:
This adds metadata updates to the same write batch as the underlying
file change. The odds of a metadata update going missing is greatly
reduced.

Bonus change: actually commit the transaction in recalcMeta.

This now makes the previous recovery thing not kick in during my stress testing...

### DIFF
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -117,6 +117,8 @@ func (s *FileSet) recalcMeta() error {
 	if err != nil {
 		return err
 	}
+	defer t.close()
+
 	var deviceID protocol.DeviceID
 	err = t.withAllFolderTruncated([]byte(s.folder), func(device []byte, f FileInfoTruncated) bool {
 		copy(deviceID[:], device)
@@ -128,7 +130,10 @@ func (s *FileSet) recalcMeta() error {
 	}
 
 	s.meta.SetCreated()
-	return s.meta.toDB(s.db, []byte(s.folder))
+	if err := s.meta.toDB(t, []byte(s.folder)); err != nil {
+		return err
+	}
+	return t.Commit()
 }
 
 // Verify the local sequence number from actual sequence entries. Returns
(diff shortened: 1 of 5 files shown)


imsodin:
That's a bit wasteful, as in a batch 1000 items (the amount we use in model) we do 1000 writes to the db on the meta key, but only the last one is actually effective (at least that's how I understand goleveldb's `Batch`, as a "stupid" list of db operations to be performed sequentially).

calmh:
Yes that's stupid, this is better.

calmh:
This does in fact trigger the recalc about as often as we did previously. Would it be over engineering it to add a "pre-flush" hook to Checkpoint() so that we could say

```
t.Checkpoint(func (t readWriteTransaction) error {
  return meta.toDB(t, ...)
})
```

and have that only run on actual checkpoints?

imsodin:
I like the interface :)  
However it does indeed smell a bit like over-engineering, or better fells like more complexity than gain (I don't consider having to recalc metadata after a crash a particularly high cost).

@AudriusButkevicius Wanna give a comment on your favorite topic: database-stuff™  ;)

calmh:
The cost is actually fairly high, on the order of 45 minutes to an hour in some setups. It just hopefully shouldn't kick in that often, and there is also a cost to doing a thousand unnecessary writes.

lib/db/set.go
```diff
@@ -202,12 +220,6 @@ func (s *FileSet) Update(device protocol.DeviceID, fs []protocol.FileInfo) {
 	s.updateMutex.Lock()
 	defer s.updateMutex.Unlock()
 
-	defer func() {
-		if err := s.meta.toDB(s.db, []byte(s.folder)); err != nil && !backend.IsClosed(err) {
-			panic(err)
-		}
-	}()
```
AudriusButkevicius: Why is this going away?

lib/db/set.go
```diff
@@ -184,7 +189,20 @@ func (s *FileSet) Drop(device protocol.DeviceID) {
...
+		panic(err)
+	}
+	defer t.close()
+
+	if err := s.meta.toDB(t, []byte(s.folder)); backend.IsClosed(err) {
+		return
+	} else if err != nil {
+		panic(err)
+	}
+	if err := t.Commit(); backend.IsClosed(err) {
```
AudriusButkevicius: This is to protect against writing our half the keys on shutdown? Feels like metadata might be better of as a single key, then we would not need trasactions.

calmh: It's so that toDB can take a transaction and be part of the same batch as the files it accounts, instead of just being hammered as a separate write later.

lib/db/set.go
calmh: Because it's done inside the updateLocal/updateRemote functions as part of their transaction.

lib/db/set.go
calmh: It is a single key, fwiw.

AudriusButkevicius: But in this specific case the transaction is not used for anything else, so why can't we just do straight writes?

calmh: Because toDB takes a transaction, as it must. Or we could duplicate it into toDBWithoutTransaction, but ... we don't drop devices that often.

PR #6339, opened 2020-02-13
lib/upnp: Exit quicker

calmh:
During NAT discovery we block for 10s (NatTimeoutS) before returning.
This is mostly noticeable when Ctrl-C:ing a Syncthing directly after
startup as we wait for those ten seconds before shutting down. This
makes it check the context a little bit more frequently.

### DIFF
--- a/lib/upnp/upnp.go
+++ b/lib/upnp/upnp.go
@@ -175,9 +175,10 @@ USER-AGENT: syncthing/1.0
 	ctx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
-	// Listen for responses until a timeout is reached
-loop:
+	// Listen for responses until a timeout is reached or the context is
+	// cancelled
 	resp := make([]byte, 65536)
+loop:
 	for {
 		if err := socket.SetDeadline(time.Now().Add(250 * time.Millisecond)); err != nil {
 			l.Infoln("UPnP socket:", err)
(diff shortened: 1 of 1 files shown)


lib/upnp/upnp.go
```diff
@@ -178,15 +172,29 @@ USER-AGENT: syncthing/1.0
...
 	l.Debugln("Listening for UPnP response for device type", deviceType, "on", intf.Name)
 
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
 	// Listen for responses until a timeout is reached
+loop:
 	for {
 		resp := make([]byte, 65536)
+		if err := socket.SetDeadline(time.Now().Add(time.Second)); err != nil {
```
AudriusButkevicius: 100ms maybe? 1s is still annoying.

I assume there is no api with a context?

calmh: Yeah not for UDP sockets I'm afraid. :( 250ms?

lib/upnp/upnp.go
```diff
@@ -160,12 +160,6 @@ USER-AGENT: syncthing/1.0
...
 
-	err = socket.SetDeadline(time.Now().Add(timeout))
-	if err != nil {
-		l.Debugln("UPnP discovery: setting socket deadline:", err)
-		return
-	}
-
 	l.Debugln("Sending search request for device type", deviceType, "on", intf.Name)
 
 	_, err = socket.WriteTo(search, ssdp)
```
imsodin: What about this, it now has no timeout at all?

calmh: I don't think UDP writes are ever blocking really

PR #6341, opened 2020-02-13
lib/model: Stop summary sender faster (ref #6319)

calmh:
One of the causes of "panic: database is closed" is that we try to send
summaries after it's been closed. Calculating summaries can take a long
time and if we have a lot of folders it's not unreasonable to think
that we might be stopped in this loop, so prepare to bail here.

### DIFF
--- a/lib/model/folder_summary.go
+++ b/lib/model/folder_summary.go
@@ -275,7 +275,7 @@ func (c *folderSummaryService) calculateSummaries(ctx context.Context) {
 					return
 				default:
 				}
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Could go deeper into Completion upto https://github.com/syncthing/syncthing/blob/a596e5e2f0dcd6356f59dbedc793026c9f576756/lib/model/model.go#L796

imsodin:
No need, unless it's urgent. There's a PR incoming getting rid of that db traversing.

PR #6342, opened 2020-02-13
lib/db: Don't call backend.Commit twice (ref #6337)

imsodin:
In `recalcMeta` we first call `t.withAllFolderTruncated` and then `t.Commit`, while it isn't allowed to call anything except `Release` after `Commit`. For some reasons that doesn't blow up in tests. It did in a very weird way (segfault deep in leveldb) in a local feature branch, took me a while to find the cause (leveldb should return an error, not segfault, but that's hardly worth spending effort on).

Also from a conceptual standpoint I believe it's wrong to call `t.Commit` or `t.Release` anywhere else than the place where the transaction was created.

### DIFF
--- a/lib/db/transactions.go
+++ b/lib/db/transactions.go
@@ -755,10 +755,7 @@ func (t *readWriteTransaction) withAllFolderTruncated(folder []byte, fn func(dev
 			return nil
 		}
 	}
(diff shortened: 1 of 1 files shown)


calmh:
frack... This explains why there was no commit in recalcMeta and I was confused about it. Ah well.

PR #6344, opened 2020-02-15
lib: Let DB errors take the App down gracefully (fixes #5907)

### LINKED ISSUE #5907: Database correctness: bubble errors instead of panics

imsodin:
Taking a stab at getting rid of all the panics in the db (#5907). As is in the nature of the issue, the diff is huge. The central mechanism is that `syncthing.App` passes an interface to `model.Model` to indicate a fatal failure has happened and Syncthing needs to stop (that more or less already exists for lib/api). Then the model passes a db specific variant of that on to `db.FileSet`, which is expected to call it whenever a fatal error occurs. Methods on `FileSet` that do already return something and other functions in `db`, now also return an error. These errors are interpreted by callers as "the shit hit the fan: We will be shut down soon, so lets bail right now".


lib/db/set_test.go
```diff
@@ -1496,9 +1496,9 @@ func TestIgnoreAfterReceiveOnly(t *testing.T) {
 
 	s.Update(protocol.LocalDeviceID, fs)
 
-	snap := s.Snapshot()
+	snap := snapshot(t, s)
 	defer snap.Release()
-	if f, ok := snap.Get(protocol.LocalDeviceID, file); !ok {
+	if f, ok := get(t, snap, protocol.LocalDeviceID, file); !ok {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

lib/model/folder_sendrecv.go
```diff
@@ -1839,7 +1888,9 @@ func (f *sendReceiveFolder) deleteDirOnDisk(dir string, snap *db.Snapshot, scanC
 			toBeDeleted = append(toBeDeleted, fullDirFile)
 		} else if f.ignores != nil && f.ignores.Match(fullDirFile).IsIgnored() {
 			hasIgnored = true
-		} else if cf, ok := snap.Get(protocol.LocalDeviceID, fullDirFile); !ok || cf.IsDeleted() || cf.IsInvalid() {
+		} else if cf, ok, err := snap.Get(protocol.LocalDeviceID, fullDirFile); err != nil {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

lib/model/folder_sendrecv_test.go
```diff
@@ -144,12 +144,12 @@ func TestHandleFile(t *testing.T) {
...
 	requiredFile.Blocks = blocks[1:]
 
-	m, f := setupSendReceiveFolder(existingFile)
+	m, f := setupSendReceiveFolder(t, existingFile)
 	defer cleanupSRFolder(f, m)
 
 	copyChan := make(chan copyBlocksState, 1)
 
-	f.handleFile(requiredFile, f.fset.Snapshot(), copyChan)
+	f.handleFile(requiredFile, snapshot(t, f.fset), copyChan)
```
golangcibot: Error return value of `f.handleFile` is not checked (from `errcheck`)

lib/model/folder_sendrecv_test.go
```diff
@@ -195,7 +195,7 @@ func TestHandleFileWithTemp(t *testing.T) {
...
```
golangcibot: Error return value of `f.handleFile` is not checked (from `errcheck`)

lib/model/folder_sendrecv_test.go
```diff
@@ -250,7 +250,7 @@ func TestCopierFinder(t *testing.T) {
 	go f.copierRoutine(copyChan, pullChan, finisherChan)
 	defer close(copyChan)
 
-	f.handleFile(requiredFile, f.fset.Snapshot(), copyChan)
+	f.handleFile(requiredFile, snapshot(t, f.fset), copyChan)
```
golangcibot: Error return value of `f.handleFile` is not checked (from `errcheck`)

lib/model/folder_sendrecv_test.go
```diff
@@ -382,7 +382,7 @@ func TestWeakHash(t *testing.T) {
 
 	// Test 1 - no weak hashing, file gets fully repulled (`expectBlocks` pulls).
 	fo.WeakHashThresholdPct = 101
-	fo.handleFile(desiredFile, fo.fset.Snapshot(), copyChan)
+	fo.handleFile(desiredFile, snapshot(t, fo.fset), copyChan)
```
golangcibot: Error return value of `fo.handleFile` is not checked (from `errcheck`)

lib/model/folder_sendrecv_test.go
```diff
@@ -411,7 +411,7 @@ func TestWeakHash(t *testing.T) {
 
 	// Test 2 - using weak hash, expectPulls blocks pulled.
 	fo.WeakHashThresholdPct = -1
-	fo.handleFile(desiredFile, fo.fset.Snapshot(), copyChan)
+	fo.handleFile(desiredFile, snapshot(t, fo.fset), copyChan)
```
golangcibot: Error return value of `fo.handleFile` is not checked (from `errcheck`)

lib/model/folder_sendrecv_test.go
```diff
@@ -995,7 +995,7 @@ func TestDeleteBehindSymlink(t *testing.T) {
 	fi.Version = fi.Version.Update(device1.Short())
 	scanChan := make(chan string, 1)
 	dbUpdateChan := make(chan dbUpdateJob, 1)
-	f.deleteFile(fi, f.fset.Snapshot(), dbUpdateChan, scanChan)
+	f.deleteFile(fi, snapshot(t, f.fset), dbUpdateChan, scanChan)
```
golangcibot: Error return value of `f.deleteFile` is not checked (from `errcheck`)

lib/model/model.go
```diff
@@ -1600,30 +1642,34 @@ func newLimitedRequestResponse(size int, limiters ...*byteSemaphore) *requestRes
 	return res
 }
 
-func (m *model) recheckFile(deviceID protocol.DeviceID, folderFs fs.Filesystem, folder, name string, size int32, offset int64, hash []byte) {
-	cf, ok := m.CurrentFolderFile(folder, name)
+func (m *model) recheckFile(deviceID protocol.DeviceID, folderFs fs.Filesystem, folder, name string, size int32, offset int64, hash []byte) error {
```
golangcibot: `(*model).recheckFile` - `folderFs` is unused (from `unparam`)

lib/model/testutils_test.go
```diff
@@ -111,19 +111,19 @@ func setupModelWithConnectionFromWrapper(w config.Wrapper) (*model, *fakeConnect
...
+	m := newModel(t, w, myID, "syncthing", "dev", db, nil)
 	m.ServeBackground()
 
 	m.ScanFolders()
 
 	return m
 }
 
-func newModel(cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersion string, ldb *db.Lowlevel, protectedFiles []string) *model {
+func newModel(t testing.TB, cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersion string, ldb *db.Lowlevel, protectedFiles []string) *model {
```
golangcibot: `newModel` - `clientName` always receives `"syncthing"` (from `unparam`)

lib/model/testutils_test.go
```diff
@@ -217,3 +217,39 @@ func dbSnapshot(t *testing.T, m Model, folder string) *db.Snapshot {
...
+type testFailer struct {
+	t testing.TB
+}
+
+func (f *testFailer) Fail() {
+	f.t.Helper()
+	f.t.Error("Model failed")
+}
+
+func newFileSet(t testing.TB, folder string, fs fs.Filesystem, ldb *db.Lowlevel) *db.FileSet {
```
golangcibot: `newFileSet` - `folder` always receives `"default"` (from `unparam`)

lib/model/testutils_test.go
```diff
@@ -217,3 +217,39 @@ func dbSnapshot(t *testing.T, m Model, folder string) *db.Snapshot {
...
+func snapshot(t testing.TB, fs *db.FileSet) *db.Snapshot {
+	t.Helper()
+	snap, err := fs.Snapshot()
+	if err != nil {
+		t.Fatal(err)
+	}
+	return snap
+}
+
+func currentFolderFile(t testing.TB, m Model, folder, file string) (protocol.FileInfo, bool) {
```
golangcibot: `currentFolderFile` - `folder` always receives `"default"` (from `unparam`)

calmh:
I didn’t read this at all except the description, on mobile and such, but I’m not confined this is the right way of doing things. I’d expect an error return of the top level Run() or similar due to an error bubbling and that would wake everything down.

imsodin:
Lets hash it out at the "description level" before investing time in the code:  

> I’d expect an error return of the top level Run() or similar due to an error bubbling and that would wake everything down.

What is the "top level Run()"? The top level `App.run` just spawns suture service and then waits for anyone to call `App.Stop` (signals via cmd/syncthing or the rest api) and then tells suture to stop and closes the db.

What  I changed is that model can also call `App.Stop`, causing `App.Wait` to return with `ExitError`. The model logs the shutdown reason with the error message from the database.  
The interface between `db.FileSet` and model is to reduce the change to it's methods. That could be dropped without changing the overall mechanism much, it would just be some tedious work and an even larger diff.

calmh:
Right, so we have a bunch of services running more or less independently. These services can run into a (retryable) error, which they will then return and get restarted for. This seems appropriate for problems that prevents the service from fulfilling its purpose, yet doesn't necessarily affect the whole app -- discovery listener failed to read a packet, connection service couldn't listen on a configured port (borderline), folder doesn't exist on disk (handled slightly differently, but conceptually the same).

Then we have stuff where retrying will not help -- database errors. If we get a failure reading it will most likely not work to try again, and a failure writing means we can't safely continue at all, etc. My first idea was to somehow signal the error severity and let some top level thing (although which?) handle it -- `return util.FatalError{err}` and something higher up `if errors.is(err, util.FatalError) { boom }`. But what does `boom` consist of? Can we expect to shut down gracefully? What does "gracefully" mean when we got an I/O error from the database?

Currently I'm thinking error handling should be at the service level, and it will have three options:

- Do something itself (log and continue for example for minor stuff)
- Return the error (we hope for better luck after new initialization, e.g. reopen the socket)
- Panic the error (we cannot continue, and neither should anyone else because we are in a bad state)

This is similar to what we already do, but the panicking should of course not happen in the database package, but for example in the folder runner. We might still want some sort of classification there, for example ErrClosed from the database should probably be a return and not a panic. Something that just reads from the database like the index sender should probably not ever panic, instead just return and hope for better luck.

imsodin:
If we panic in the end, what's the point of doing error handling in `lib/db` then instead of just panicking then and there as we do now? My entire motivation was that panicking isn't considered a nice behaviour for a library in general and as it happens in a service, i.e.  goroutine, it's not even recoverable (not sure about that). So I wanted to exit and return an error instead.

As to who is the top-level thing that decides what to do: Whoever uses the library, for us `cmd/syncthing`. And what it does depends on the use-case of the library (e.g. how it can make the user aware of the fatal error or maybe there's circumstances where rebuilding the db automatically is an appropriate choice). In our case I'd just somehow preserve a bit of state marking the db as corrupt and abort saying so when Syncthing is started.

calmh:
Panicking in a library is clearly broken design. That said, I'm not sure the distinction of what is a "library" or not is totally clear in our case. :) It's obviously not `lib/db`, because it can't really know the severity of a failure and the appropriate action to take. We might want to retry some read errors forever, just noting to the log that something failed. But if we've just synced a file and then get an I/O error back when committing to the db, what can we do? I'm not convinced it's sensible to try a soft shutdown, waiting for all other operations to complete, attempting to close the database, and so on. In that scenario `lib/model` starts to feel more like "the application" than "a library" and might have the power to take down the program when it runs into unfixable badness.

calmh:
I'm not saying this must be done using panic though. It could equally well be done by returning an error of a kind that indicates that we must exit. It's just that our suture infrastructure doesn't really support that at the moment.

imsodin:
I am circumventing that limitation of the suture infra by passing a "control object" from the creator of the top-level suture services to places where exit conditions are present (like it's already the case for `lib/api`). What is your view on that?

As to the different natures of errors: I'd do the distinction between possibly recoverable db errors and conditions where we mustn't use the error anymore inside `lib/db` by marking the db as broken, most likely by putting a file into the db directory. Then on the outside we can distinguish between errors that might be recoverable by a full restart and those where there's no point in trying again.

And when I talk of library I am currently thinking of golang bindings for `lib/syncthing` - which already works (Martchus uses it) but with limitations (https://github.com/syncthing/syncthing/issues/4085). I also dabbled in making bindings for our filesystem abstractions, and that should work too (needs a bit work due to unsupported types in the interface). If that works, the "only" thing required to make Android useful again is a nice SAF java implementation adapted to support our filesystem interface. Though maybe google will budge on requiring SAF and make all of this moot - one can hope :)

AudriusButkevicius:
This is now quite far behind. What are we doing with this? Was any consensus on what needs to be done achieved, or is this not something we want to have?

imsodin:
I still want something like it. I need to make a minimal example of my proposed mechanism to discuss and get consensus - doing the legwork to create an insensibly huge diff is the wrong way around.

calmh:
I'll open a thread on the forum for some high level brainstorming on this

imsodin:
For reference because I had a hard time finding it lately: https://forum.syncthing.net/t/shutdown-on-error/14908

PR #6349, opened 2020-02-18
lib: Modify FileInfos consistently (ref #6321)

imsodin:
I unsuccessfully tried to reproduce the scenario https://github.com/syncthing/syncthing/issues/6321#issuecomment-587548420. That's due to the GC not kicking in, even if noone actually has the blocks. It's enough for an ignored file to have the `BlocksHash`.

Nevertheless I believe this is a good change: It firstly makes sure that whenever `Blocks` is set to nil, so is `BlocksHash`. That means blocks can get GCed when everyone has ignored files, while it can't at the moment. I also put in a check in the db that panics if this isn't true. That could be changed to just fix it on the fly, but I'd rather know about it when it happens and fix it. And while checking for other places where the same mistake might happen, I ensured that wherever file infos are modified, that's happening through the same functions (i.e. consistently).

### DIFF
--- a/lib/db/structs.go
+++ b/lib/db/structs.go
@@ -129,27 +129,36 @@ func (f FileInfoTruncated) FileModifiedBy() protocol.ShortID {
 }
 
 func (f FileInfoTruncated) ConvertToIgnoredFileInfo(by protocol.ShortID) protocol.FileInfo {
-	return protocol.FileInfo{
-		Name:         f.Name,
-		Type:         f.Type,
-		ModifiedS:    f.ModifiedS,
-		ModifiedNs:   f.ModifiedNs,
-		ModifiedBy:   by,
-		Version:      f.Version,
-		RawBlockSize: f.RawBlockSize,
-		LocalFlags:   protocol.FlagLocalIgnored,
-	}
+	file := f.copyToFileInfo()
+	file.SetIgnored(by)
+	return file
 }
 
-func (f FileInfoTruncated) ConvertToDeletedFileInfo(by protocol.ShortID, localFlags uint32) protocol.FileInfo {
+func (f FileInfoTruncated) ConvertToDeletedFileInfo(by protocol.ShortID) protocol.FileInfo {
+	file := f.copyToFileInfo()
+	file.SetDeleted(by)
+	return file
+}
+
+// copyToFileInfo just copies all members of FileInfoTruncated to protocol.FileInfo
+func (f FileInfoTruncated) copyToFileInfo() protocol.FileInfo {
 	return protocol.FileInfo{
-		Name:       f.Name,
-		Type:       f.Type,
-		ModifiedS:  time.Now().Unix(),
-		ModifiedBy: by,
-		Deleted:    true,
-		Version:    f.Version.Update(by),
-		LocalFlags: localFlags,
+		Name:          f.Name,
+		Size:          f.Size,
+		ModifiedS:     f.ModifiedS,
+		ModifiedBy:    f.ModifiedBy,
+		Version:       f.Version,
+		Sequence:      f.Sequence,
+		SymlinkTarget: f.SymlinkTarget,
+		BlocksHash:    f.BlocksHash,
+		Type:          f.Type,
+		Permissions:   f.Permissions,
+		ModifiedNs:    f.ModifiedNs,
+		RawBlockSize:  f.RawBlockSize,
+		LocalFlags:    f.LocalFlags,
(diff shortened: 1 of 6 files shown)


lib/db/structs.go
```diff
@@ -129,27 +129,36 @@ func (f FileInfoTruncated) FileModifiedBy() protocol.ShortID {
...
+func (f FileInfoTruncated) ConvertToDeletedFileInfo(by protocol.ShortID) protocol.FileInfo {
+	file := f.copyToFileInfo()
+	file.SetDeleted(by)
+	return file
 }
 
-func (f FileInfoTruncated) ConvertToDeletedFileInfo(by protocol.ShortID, localFlags uint32) protocol.FileInfo {
+// copyToFileInfo just copies all members of FileInfoTruncated to protocol.FileInfo
+func (f FileInfoTruncated) copyToFileInfo() protocol.FileInfo {
 	return protocol.FileInfo{
```
calmh: As a slower but potentially more robust way,

```
bs, _ :=  f.Marshal()
var f protocol.FileInfo
f.Unmarshal(bs)
```

will pick up new fields automatically... 🤷‍♂

lib/model/folder_recvonly.go
```diff
@@ -104,14 +104,8 @@ func (f *receiveOnlyFolder) Revert() {
...
 
-			fi = protocol.FileInfo{
-				Name:       fi.Name,
-				Type:       fi.Type,
-				ModifiedS:  time.Now().Unix(),
-				ModifiedBy: f.shortID,
-				Deleted:    true,
-				Version:    protocol.Vector{}, // if this file ever resurfaces anywhere we want our delete to be strictly older
-			}
+			fi = fi.SetDeleted(f.shortID)
```
calmh: SetDeleted doesn't have a return value?

lib/model/folder_recvonly.go
```diff
@@ -155,7 +149,7 @@ func (f *receiveOnlyFolder) Revert() {
 		})
 	}
 	if len(batch) > 0 {
-		f.updateLocalsFromScanning(batch)
+		f.updateLoclsFromScanning(batch)
```
calmh: wat

lib/protocol/bep_extensions.go
```diff
@@ -266,24 +266,37 @@ func BlocksEqual(a, b []BlockInfo) bool {
...
 
 func (f *FileInfo) SetMustRescan(by ShortID) {
-	f.LocalFlags = FlagLocalMustRescan
-	f.ModifiedBy = by
-	f.Blocks = nil
-	f.Sequence = 0
+	f.setLocalFlags(by, FlagLocalMustRescan)
 }
 
 func (f *FileInfo) SetIgnored(by ShortID) {
```
calmh: `ShortID` is capitalized in all of these, old sin, but shouldn't be...

lib/protocol/bep_extensions.go
```diff
@@ -266,24 +266,37 @@ func BlocksEqual(a, b []BlockInfo) bool {
...
+	f.ModifiedS = time.Now().Unix()
+	f.setNoContent()
+}
+
+func (f *FileInfo) setLocalFlags(by ShortID, flags uint32) {
+	f.RawInvalid = false
+	f.LocalFlags = flags
 	f.ModifiedBy = by
-	f.Blocks = nil
 	f.Sequence = 0
```
calmh: Not clear to me why this one sets sequence to zero but not the others...

lib/db/structs.go
imsodin: If lots of stuff gets deleted or ignored, this function gets called for every item. So can be a lot.

lib/protocol/bep_extensions.go
imsodin: Why? It's used abundantly outside of lib/protocol.

lib/protocol/bep_extensions.go
imsodin: Me neither, didn't want to change existing behaviour though. I don't think it matters, as for local it gets a new sequence number when adding to db and for remotes sequence isn't used (and I don't see any reason to modify a fileinfo for a remote before adding to db). However I also don't see any point in resetting it.

I'll prune it, call out if you can think of any reason to keep it or just think "better don't touch it".

lib/protocol/bep_extensions.go
calmh: braino

Issue #6304, opened 2020-02-03
Syncing issues, database missing sequence entries


midnightdim:
I recently installed and configured Syncthing to sync files across multiple devices. It seemed to work on some folders, but then I ran into an issue or maybe even 2 different issues.

There's one folder X shared from device A (Home PC) to device B (Brix).
The current issue is:
1. Device B shows (in syncthing UI) that it's "Up to Date" with device A.
2. Device B shows local state different from global state for the folder X. Both device A and device B show the same global state (which is expected).
3. Device A shows "syncing" with device B, but nothing really happens.
4. Device A shows some "Out of Sync Items" - I randomly checked 4-5 files on device B - they exist there. But the total folder sizes are different on device A and device B (the latter is smaller).

Device A:
![изображение](https://user-images.githubusercontent.com/6467552/73632212-a4f3de00-468d-11ea-982e-61bc24038bca.png)
![изображение](https://user-images.githubusercontent.com/6467552/73632244-b4732700-468d-11ea-9423-5d98c01af8a1.png)

Device B:
![изображение](https://user-images.githubusercontent.com/6467552/73632260-c2c14300-468d-11ea-8a61-ae536969d743.png)
![изображение](https://user-images.githubusercontent.com/6467552/73632275-ce146e80-468d-11ea-9e41-c76a373087d5.png)

--

There's another folder Y also shared from device A to device B.
The problem is:
1. The total size of this folder on the file system is different on A and B. On device B the size is smaller. [UPDATE: this was incorrect - see my comment below]
2. In syncthing UI device A and device B show different number of files for this folder (and different total size) - that's unexpected, and each of them shows that local state is equal to global state. Also, the total size and files count shown on device B are bigger than what device A shows - this is unexpected since all files came to B from A and also wrong (see point 1).

Device A:
![изображение](https://user-images.githubusercontent.com/6467552/73632164-9279a480-468d-11ea-8d84-430b14ed4cdc.png)

Device B:
![изображение](https://user-images.githubusercontent.com/6467552/73632144-85f54c00-468d-11ea-9e84-47f4aa4ddbb8.png)

--

In the logs I don't see any issues. Tried restarting syncthing on both ends, even reinstalled it on device A, the problems are still there.

Version Information:

Device A:
syncthing v1.3.3 "Fermium Flea" (go1.13.4 windows-amd64) teamcity@build.syncthing.net 2019-12-08 12:31:26 UTC
OS Version: Windows 7

Device B:
[...]

PR #6350, opened 2020-02-19
lib/db: Schema update to repair sequence index (ref #6304)

imsodin:
While #6336 and #6337 take care of preventing new sequence index problems, they do not clean up existing messes. This adds (extends) the schema transitions to check for and fix missing and duplicate sequence entries. I had to refactor the metadata loading/rechecking functions, such that they aren't methods of `FileSet` anymore.

The PR as it is is intended to go into a new 1.4.0 RC, as it addresses a problem that's partially fixed there. If we ship it in 1.4.0 the min Syncthing version needs to be modified.

### DIFF
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -431,7 +431,7 @@ func (db *schemaUpdater) updateSchema6to7(_ int) error {
 
 func (db *schemaUpdater) updateSchemato9(prev int) error {
 	// Loads and rewrites all files with blocks, to deduplicate block lists.
-	// Checks for files with the same sequence number
+	// Checks for missing or incorrect sequence entries and rewrites those.
 
 	t, err := db.newReadWriteTransaction()
 	if err != nil {
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -463,12 +463,11 @@ func (db *schemaUpdater) updateSchemato9(prev int) error {
 				return err
 			}
 			switch dk, err := t.Get(sk); {
-			case backend.IsNotFound(err):
-				if err := t.Put(sk, it.Key()); err != nil {
+			case err != nil:
+				if !backend.IsNotFound(err) {
 					return err
 				}
-			case err != nil:
-				return err
+				fallthrough
 			case !bytes.Equal(it.Key(), dk):
 				folderStr := string(folder)
 				meta, ok := metas[folderStr]
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -7,9 +7,11 @@
 package db
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
 
+	"github.com/syncthing/syncthing/lib/db/backend"
 	"github.com/syncthing/syncthing/lib/protocol"
 )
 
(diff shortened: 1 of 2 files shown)


lib/db/schemaupdater.go
```diff
@@ -421,24 +426,63 @@ func (db *schemaUpdater) updateSchema6to7(_ int) error {
...
 			return err
 		}
+		folder, ok := t.keyer.FolderFromDeviceFileKey(it.Key())
+		if !ok {
+			return errFolderIdxMissing
+		}
+		if sk, err = t.keyer.GenerateSequenceKey(sk, folder, fi.Sequence); err != nil {
+			return err
+		}
+		if dk, err := t.Get(sk); backend.IsNotFound(err) {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

AudriusButkevicius:
We should also add an easier way to check the consistency of the database.

lib/db/schemaupdater.go
```diff
@@ -421,24 +429,69 @@ func (db *schemaUpdater) updateSchema6to7(_ int) error {
...
+			case err != nil:
+				if !backend.IsNotFound(err) {
+					return err
+				}
+				fallthrough
+			case !bytes.Equal(it.Key(), dk):
+				folderStr := string(folder)
+				meta, ok := metas[folderStr]
+				if !ok {
+					meta = loadMetadataTracker(db.Lowlevel, folderStr)
```
calmh: We might start refreshing metadata here in the middle of a migration. That's probably not disastrous, but weird...

lib/db/schemaupdater.go
```diff
@@ -421,24 +429,69 @@ func (db *schemaUpdater) updateSchema6to7(_ int) error {
...
+					metas[folderStr] = meta
+				}
+				fi.Sequence = meta.nextLocalSeq()
+				if sk, err = t.keyer.GenerateSequenceKey(sk, folder, fi.Sequence); err != nil {
+					return err
+				}
+				if err := t.Put(sk, it.Key()); err != nil {
+					return err
+				}
+				if err := t.putFile(it.Key(), fi); err != nil {
```
calmh: Actually this thing is broken. We unmarshal to a raw fileinfo, which will have a block hash but not filled in blocks, then use putFile which expects a fully formed FileInfo with blocks filled in. The combo is bad and will result in brokenness. Either we need to actually load the blocks as well using unmarshalTrunc or do a raw marshal+put as well.

calmh: At least now when putFile reacts badly to `blocks == nil && blocksHash != nil`. If we want it to be able to save raw FileInfos then it must let that past without complains, including not trying to nil out blocksHash just because the blocks are nil.

imsodin: Oh crap, nice find. Makes the panics almost benign compared to wrecking dbs with the latest, luckily unshipped change of niling blockshashes.

I agree ignoring blockhash consistency at such a low level makes most sense.

And I need to stop any future urge to use a panic to ensure expected behavior. Either it's important enough to test or add something like "non-panic sentry events" or just behave sanely in any case...

PR #6354, opened 2020-02-22
lib/db: Also fix non-nil BlocksHash in transition (fixes #6353)

### LINKED ISSUE #6353: panic: Blocks is nil, but BlocksHash is not
Culprit is #6350: The check&panic when putting a file to db triggers for all old db entries, because the fix in #6350 only targets new entries. 

Sentry Issue: [SYNCTHING-100](https://sentry.syncthing.net/syncthing/syncthing/issues/1043/?referrer=github_integration)

```

  File "C:/BuildAgent/work/174e136266f8a219/lib/db/transactions.go", line 473, in putFile
    panic("Blocks is nil, but BlocksHash is not")
  File "C:/BuildAgent/work/174e136266f8a219/lib/db/schemaupdater.go", line 485, in updateSchemato9
    if err := t.putFile(it.Key(), fi); err != nil {
  File "C:/BuildAgent/work/174e136266f8a219/lib/db/schemaupdater.go", line 97, in updateSchema
    if err := m.migration(int(prevVersion)); err != nil {
  File "C:/BuildAgent/work/174e136266f8a219/lib/db/schemaupdater.go", line 52, in UpdateSchema
    return updater.updateSchema()
  File "C:/BuildAgent/work/174e136266f8a219/lib/syncthing/syncthing.go", line 185, in startup
    if err := db.UpdateSchema(a.ll); err != nil {
  File "C:/BuildAgent/work/174e136266f8a219/lib/syncthing/syncthing.go", line 106, in Start
    if err := a.startup(); err != nil {
  File "C:/BuildAgent/work/174e136266f8a219/cmd/syncthing/main.go", line 641, in syncthingMain
    if err := app.Start(); err != nil {
  File "C:/BuildAgent/work/174e136266f8a219/cmd/syncthing/main.go", line 381, in main
    syncthingMain(options)

panic: Blocks is nil, but BlocksHash is not
```

imsodin:
Extend the transition to clean up any entries with `Blocks == nil && BlocksHash != nil` to prevent panics.

### DIFF
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -26,8 +26,9 @@ import (
 //   7: v0.14.53
 //   8: v1.4.0
 //   9: v1.4.0
+//   10: v1.4.0
 const (
-	dbVersion             = 9
+	dbVersion             = 10
 	dbMinSyncthingVersion = "v1.4.0"
 )
 
@@ -88,7 +89,7 @@ func (db *schemaUpdater) updateSchema() error {
 		{5, db.updateSchemaTo5},
 		{6, db.updateSchema5to6},
 		{7, db.updateSchema6to7},
-		{9, db.updateSchemato9},
+		{10, db.updateSchemato10},
 	}
 
 	for _, m := range migrations {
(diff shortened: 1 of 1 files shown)


lib/db/schemaupdater.go
```diff
@@ -482,21 +486,20 @@ func (db *schemaUpdater) updateSchemato9(prev int) error {
...
 		}
-		if prev == 8 {
-			// The transition to 8 already did the changes below.
-			continue
-		}
-		if fi.Blocks == nil {
-			continue
+		// 1.4.0 RCs had a bug where we'd reset the blocks, but
+		// not the blocks-hash (version 10).
+		if fi.Blocks == nil && fi.BlocksHash != nil {
```
calmh: Isn't this every file if prev >= 8?

imsodin: In theory I'd still like to not just swallow incorrect blocks hashes, but in practice I agree: https://github.com/syncthing/syncthing/pull/6355

And as far as I understand the transition to 8 didn't change `BlocksHash` if `Blocks == nil` and before that `BlocksHash` didn't exist, so only files which were ignored/otherwise invalidated or deleted after upgrading to a 1.4.0 release candidate should be affected. The amount of panic events on sentry seems to suggest otherwise though.

calmh: You’re raw unmarshalling the file info, so every file with blocks will have a block hash and nil block list

Issue #5930, opened 2019-08-06
Remove /rest/db/browse


andreymal:
As @calmh [said](https://github.com/syncthing/syncthing/issues/5845#issuecomment-510424104),

>Syncthing, however, is not a file manager.

As @AudriusButkevicius [said](https://github.com/syncthing/syncthing/issues/5845#issuecomment-510445357),

>I do agree we should remove /browse however

The `/rest/db/browse` and `/rest/system/browse` endpoints are file manager features; if Syncthing is not a file manager, these endpoints should be removed.

I made this issue so you don’t forget to carry out your plans in next major version.

calmh:
/system/browse is in use by the GUI, and not for file manager reasons. /db/browse is not in use, though any use would be non-file-manager in nature as it does not in fact browse files but our in-database data. As it's dead code (I don't know of any uses of it) I think we can consider removing it.

AudriusButkevicius:
There were a few PRs starting to use it, like a browser showing which files are ignored etc, a PR building checkboxes for ignores etc, so there is some use, yet the current api is super ugly and hard to work with, so we should either hide it until its needed or redo the api if we want to keep it.

andreymal:
Since v2.0.0-beta.2 has been released with breaking changes, a friendly reminder that Syncting is not a file manager and these two endpoints should be removed

calmh:
Neither of these have anything to do with being a file manager, imho. /system/browse is in use by the GUI. /db/browse is not in use as far as I know and could be refactored, but it's an essential feature for things like a potential tree selector for ignore patterns. I don't see an obvious reason to remove either, 2.0 or not.

andreymal:
>a potential tree selector for ignore patterns.

This is literally a file manager.

In this case, please reopen 5845 because the search feature will be useful for that potential tree selector

andreymal:
> my project about selective ignores

Another example of Syncthing acting as a real file manager

> an ability to look inside the database and **find** a stuff available in the "cloud"

Yeah this is exactly what #5845 is about (I'm still sad)

PR #6358, opened 2020-02-24
Add tree view showing what files are ignored

ibizaman:
### Prelude

This PR is a second stab at #2491, after the failure of #5132. This PR is not ready to be merged yet but it should be close to providing a useful feature that can be merged even though it doesn't fulfill all requirements in the issue. Think of it as a way to slowly chip away at #2491. The requirements I tried to fulfill here are:

* Must be understandable by someone who understands the interface in use by DropBox, OneDrive, etc.
* Must be a simple tree view with checkboxes
  * Root of tree (=folder root) checked by default
* Power user access to flexible, underlying patterns

Feel free to give your opinion on this, even if it means changing a bunch of stuff or simply refusing to merge. I don't mind at all as that would mean we're progressing on this issue by knowing what we _don't_ want.

### Purpose

This PR introduces a tree view in the edit folder modal in the ignores tab. The tree view shows the whole global directory, not only what's local to the device/folder. Each node of the tree has a checkbox that is checked unless that file or directory is ignored. The sync is one way: updating the patterns changing what checkbox is checked but changing a checkbox has no effect. Doing this I was able to keep the PR concise.

The corresponding folder state shows that indeed, some files are ignored:

When updating the ignored patterns and clicking Save, the modal is not closed but instead the nodes in the tree view are refreshed. The goal is to see relatively quickly the impact of changing the ignore rules. It's kind of like having a REPL loop. I modified the behavior of the Save button to _not_ close the modal, otherwise refreshing the tree doesn't even make sense. I don't really like modifying the Save button behavior but I don't know what would be better.

Here is for example what happens if I modify the pattern match `documents` above and click on save:

When clicking on a node, the bottom text is updated with information pertaining to that node. It's very crude and probably useless for now but I wanted to show this is possible.

`documents` is selected:

`pictures` is selected:

Changing the checkbox state on a node has no effect at all. This is most certainly bad UX and I'm happy to change that to something that makes more sense.

The tree is lazy loaded using fancytree's lazyLoad feature. I added a new api endpoint that merges the `db/browse` and `db/file` while only keeping a few fields in the returned json. I felt it made more sense to do that in go than in javascript. Note that #5930 talks about removing the `db/browse` api endpoint and this PR goes a bit against the claim that Syncthing is not a file browser.

### Testing

Manual. I set up two Syncthing instances and shared a folder between them:
```
STNOUPGRADE=true STGUIASSETS=~/src/syncthing/gui ./bin/syncthing -no-browser
STNOUPGRADE=true STGUIASSETS=~/src/syncthing/gui ./bin/syncthing -no-browser -home ~/.config/syncthing2 -gui-address "http://127.0.0.1:8385"
```

I then made some updates to the ignored patterns.

### Documentation

I was waiting a consensus on this first.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1771,8 +1771,82 @@ angular.module('syncthing.core')
                     $scope.emitHTTPError(err);
                 });
 
+            $scope.currentFolderPath = $scope.currentFolder.path;
+            // See fancytree demo with all options:
+            // http://wwwendt.de/tech/fancytree/demo/index.html
+            $scope.tree = $("#folder-ignores-tree").fancytree({
+                extensions: [],
+                quicksearch: true,  // select next node by spelling the name
+                source: [{
+                    "title": $scope.currentFolderPath,
+                    "key": "",
+                    folder: true,
+                    lazy: true,
+                    unselectable: true,
+                    selected: true,
+                }],
+                checkbox: true,
+                clickFolderMode: 4, // one click activates, double click expands
+                selectMode: 2,      // selecting a folder does not select its children elements
+                tabindex: "0",      // the whole tree is focused with the TAB key
+                beforeActivate: function(event, data) {
+                    var node = data.node;
+                    if (!data.node.isActive()) {
+                        $("#folder-ignores-nodestatus").
+                            html($scope.currentFolderPath + '/' + data.node.key + ":<br /><ul>" +
+                                 "<li>type:" + (data.node.folder ? "directory" : "file") + "</li>" +
+                                 "<li>ignored: " + (data.node.selected ? "false" : "true") + "</li>" +
+                                 "</ul>");
+                    } else {
+                        $("#folder-ignores-nodestatus").html($translate.instant("Click on a node to see more details about it."));
+                    }
+                },
+                lazyLoad: function(event, data) {
+                    var dfd = $.Deferred();
+                    data.result = dfd.promise();
+                    $http.get(urlbase + '/folder/state?folder=' + encodeURIComponent($scope.currentFolder.id) + '&prefix=' + encodeURIComponent(data.node.key))
+                        .success(function (tree) {
+                            var result = $.map(tree, function(obj, title) {
+                                var key = data.node.key ? data.node.key + '/' + title : title;
+                                var selected = !obj.ignored;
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
See https://github.com/syncthing/syncthing/pull/6288 for the changes to the Go API which potentially compliments this. If we can use this instead of the new endpoint, I'd rather go for that.
Your current implementation does a db lookup for every file, where you could just use a pattern matcher instead

I think this feature is generally useful with a few caveats which are debatable:
1. The checkboxes should not be clickable or there should be no checkboxes at all and potentially just show what's left?
2. I would not faf with the save button, and instead do a debounce on text input, essentially once text input is modified you start a timer (show spinner instead of a tree or something), if the timer expires without any new input, you call the apis, if there was additional input, you reset the timer back to 2s or so, and keep showing the spinner. But this would need some sort of new API endpoint where you post patterns (before they are saved), and query the tree that way.
3. In your `Here is for example what happens if I modify the pattern match documents above and click on save:` I am not sure why all nodes are selected.
4. The details as they stand now are not super useful, I guess it would be useful if we showed additional data such as size, who last modified, etc.

imsodin:
To me this seems to miss the target audience: "Power users" who want patterns probably don't care much about the tree - it's nice to have but not a significant help. I'd expect all other users getting just as frustrated having a tree now with nice checkboxes and all, but not being able to use it.

If I remember in the previous PR a separation of an tree and advanced mode was brought up: Either a checkable tree view with filename only patterns (i.e. no path matching) or just patterns as they exist now. That's a direction I'd expect to fulfill needs of both sides. And I am not quite sure how your PR goes into a direction (obviously doesn't need to be this concrete one) that would cover a wide range of requirements. What I mean is that the proposed tree to visualize existing patterns feels nice, and I am not opposed to it in general, but I don't understand how it is a step towards #2459. It feels more like #6288 just for ignore patterns (and maybe that's it, merging the two creating a good visualization of global/local/ignored/remote items would be an awesome feature).

AudriusButkevicius:
I guess you are right, that this is somewhat not what was asked, but I think having a preview is quite nice feature, even if it does not take us to the right direction.

Power users will probably ignore it and carry on as they do, and novice users can atleast see what their uneducated guesses are resulting in.

Obviously, ideally we would just have advanced and basic modes and carry on with our life.

st-review:
:robot: *beep boop*

PR #6359, opened 2020-02-24
cmd/syncthing: Do not truncate/rotate logs at start

imsodin:
Currently we are truncating a file when opening it. And we aren't opening a file when creating a rotated logger, instead rotating and opening on first write. Now we don't truncate anymore and rotate only if the log exceeds the limit.

See also https://forum.syncthing.net/t/rotating-logfiles-on-every-start/14542

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -230,7 +230,7 @@ func parseCommandLineOptions() RuntimeOptions {
 	flag.BoolVar(&options.Verbose, "verbose", false, "Print verbose log output")
 	flag.BoolVar(&options.paused, "paused", false, "Start with all devices and folders paused")
 	flag.BoolVar(&options.unpaused, "unpaused", false, "Start with all devices and folders unpaused")
-	flag.StringVar(&options.logFile, "logfile", options.logFile, "Log file name (still always logs to stdout).")
+	flag.StringVar(&options.logFile, "logfile", options.logFile, "Log file name (still always logs to stdout). This is best effort, i.e. Syncthing will keep running if encountering errors writing to this file.")
 	flag.IntVar(&options.logMaxSize, "log-max-size", options.logMaxSize, "Maximum size of any file (zero to disable log rotation).")
 	flag.IntVar(&options.logMaxFiles, "log-max-old-files", options.logMaxFiles, "Number of old files to keep (zero to keep only current).")
 	flag.StringVar(&options.auditFile, "auditfile", options.auditFile, "Specify audit file (use \"-\" for stdout, \"--\" for stderr)")
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -230,7 +230,7 @@ func parseCommandLineOptions() RuntimeOptions {
 	flag.BoolVar(&options.Verbose, "verbose", false, "Print verbose log output")
 	flag.BoolVar(&options.paused, "paused", false, "Start with all devices and folders paused")
 	flag.BoolVar(&options.unpaused, "unpaused", false, "Start with all devices and folders unpaused")
-	flag.StringVar(&options.logFile, "logfile", options.logFile, "Log file name (still always logs to stdout). This is best effort, i.e. Syncthing will keep running if encountering errors writing to this file.")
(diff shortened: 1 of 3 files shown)


calmh:
This doesn't take the current size when opening into account, the file might already be large when we open it.

cmd/syncthing/monitor.go
```diff
@@ -160,6 +166,11 @@ func monitorMain(runtimeOptions RuntimeOptions) {
 			cmd.Process.Signal(sigHup)
 			err = <-exit
 
+		case err = <-logErrChan:
+			l.Warnln("Logger error, stopping Syncthing:", err)
+			cmd.Process.Signal(sigTerm)
```
golangcibot: Error return value of `cmd.Process.Signal` is not checked (from `errcheck`)

imsodin:
Thanks. That means opening a rotated logfile can now produce an error. There's no error handling for logging so far, errors writing logs are just ignored. I now handle both the new error when opening and write errors: They are considered fatal and take the entire app down (restarting after error by default). I believe logging is important enough to warrant that behaviour.

calmh:
Hmm. This is from the monitor so we can't expect a restart unless we're running under a service manager... I'm not really sure I consider logging that fundamental, it feels more like a secondary concern to be honest.

imsodin:
I don't actually shut down the monitor on logging error, I stop the monitored Syncthing process. After that the usual process kicks in, meaning it restarts if `err != nil` and `-no-restart` was given. However you're right, both reporting the error about logging, then normal shutdown and eventually restart will just trigger the error again. I'll revert that, but propose to keep exiting when encountering an error to open the log file on start, as the user is likely to notice that - do you agree or should we just ignore that as well and keep logging to stdout?

calmh:
Frankly this (the error handling) is complicated and not in the spirit of the original PR subject... To resolve the unnecessary rotate I'd suggest having our `create()`/`open()` return `io.Writer, int64, error` with the middle thing being the size at open, determined by Stat() on the fd after opening. (And the flag thing you did.)

I think having the log file open being fatal on first open (start) is fine. If an error happens later I think it should just be a stderr print and carry on...

imsodin:
> Frankly this (the error handling) is complicated and not in the spirit of the original PR subject...

It came into my focus due to needing to know the filesize at start, which can result in an error. Before no filesystem interaction happened until first log write and thus no error could occur.  

I now removed error handling, except for printing an info line if the error mentioned above occurs at startup, i.e. behaviour not regarding truncation/rotation is unchanged. I'll propose a follow up PR with limited (startup) error handling.

cmd/syncthing/main.go
```diff
@@ -230,7 +230,7 @@ func parseCommandLineOptions() RuntimeOptions {
 	flag.BoolVar(&options.Verbose, "verbose", false, "Print verbose log output")
 	flag.BoolVar(&options.paused, "paused", false, "Start with all devices and folders paused")
 	flag.BoolVar(&options.unpaused, "unpaused", false, "Start with all devices and folders unpaused")
-	flag.StringVar(&options.logFile, "logfile", options.logFile, "Log file name (still always logs to stdout).")
+	flag.StringVar(&options.logFile, "logfile", options.logFile, "Log file name (still always logs to stdout). This is best effort, i.e. Syncthing will keep running if encountering errors writing to this file.")
```
calmh: I don't think we need to be this verbose really...

Screenshot 2020-02-26 at 07 43 14

cmd/syncthing/monitor.go
```diff
@@ -55,28 +55,32 @@ func monitorMain(runtimeOptions RuntimeOptions) {
...
 		}
-
-		if runtime.GOOS == "windows" {
-			// Translate line breaks to Windows standard
-			fileDst = osutil.ReplacingWriter{
-				Writer: fileDst,
-				From:   '\n',
-				To:     []byte{'\r', '\n'},
+		if err != nil {
+			l.Infoln(`Failed to setup logging to file "%s", proceeding with logging to stdout only: %v`, logFile, err)
```
calmh: This should be a printf call... And might use %q instead of the quotes, although that will double escape things on Windows so maybe not. And maybe actually Warnf. And the final error will typically include the filename anyway. :)

PR #6362, opened 2020-02-25
lib/connections: Refactor status for testing (ref #6361)

imsodin:
This probably overkill but I just felt the urge to write a test, even if that test comes too late.


calmh:
❤️

PR #6363, opened 2020-02-25
lib/dialer: Bring back address faking connection (fixes #6289)

### LINKED ISSUE #6289: "general SOCKS server failure" since syncthing 1.3.3
https://forum.syncthing.net/t/general-socks-server-failure-since-syncthing-1-3-3/14375

Syncthing Version: v1.3.3
OS Version: Archlinux

Since I upgraded syncthing to 1.3.3 on archlinux, it is unable to connect to my cntlm socks proxy running locally. The log says repeatedly:
```
INFO: Listen (BEP/relay): joining session: socks connect tcp localhost:8010->127.0.0.1:443: unknown error general SOCKS server failure
```
I have configured socks for syncthing via systemctl edit syncthing@user
```
[Service]
Environment=all_proxy='socks5://localhost:8010' ALL_PROXY_NO_FALLBACK=1 http_proxy= https_proxy= GOMAXPROCS=1
```
Downgrading syncthing to 1.3.2 works as a workaround.
Logs of both versions with STTRACE=dialer:
https://gist.github.com/SanskritFritz/3e13c89f00f4f7b857b705a58dcd3d43
https://gist.github.com/SanskritFritz/719eccbd21e29f54a9df844e67f2196f

AudriusButkevicius:
Seems it was required.


lib/dialer/public.go
```diff
@@ -85,6 +89,9 @@ func dialContextWithFallback(ctx context.Context, fallback proxy.ContextDialer,
 		proxyConn, proxyErr = dialer.DialContext(ctx, network, addr)
 		l.Debugf("Dialing proxy result %s %s: %v %v", network, addr, proxyConn, proxyErr)
 		close(proxyDone)
+		proxyConn = dialerConn{
```
calmh: Needs same wrapping also in the noFallback case above

calmh: BTW why is this fallback dialer so complicated? The old fallback dialer was straightforward, and had the property of generally not leaking the client address when the proxy works...

AudriusButkevicius: It used the same wrapper and I think it was even more complicated, as we had hacks for constructing our own proxy etc. This one does parallel dials, where the previous was doing timeout/2 and giving that to each attempt.

calmh: Yeah the wrapper is fine, I was just thinking compared to the previous "try one, then the other". Parallel sounds good but I'm not sure that we in general want to always have a direct connection attempt when the user specifically told us to use the proxy.

AudriusButkevicius: The no fallback case it handled above as you said, which does not do parallel dials.

calmh: That's not the default though. And it's not what this PR is about anyway, but I consider it a negative change that we always do direct connections when behind `all_proxy`, which we didn't use to do before.

imsodin: Fair - lets undo that in a follow-up PR then.

As to wrapping the fallback: That's a direct connection, so doesn't need wrapping, right? It wasn't before: https://github.com/syncthing/syncthing/pull/6177/files#diff-9f500d7732c3c394ce9e70b584df7dabL73

calmh: I'm not saying wrap the fallback, I'm saying the no-fallback proxy-only dial on line 77 needs a wrap as well.

imsodin: ... yeah sorry for that one - and I once considered reading my forte

AudriusButkevicius: I don't think dual dial is a big deal. Go does it for dual stack. This is the non no-fallback case, in which case the user is ok for us to dial directly, also, we still wait for proxy to timeout before we use the non-proxy one, so the behaviour is the same, but just faster.

calmh: I don't think dual stack is the same thing at all. Proxy users will generally expect the proxy to be used -- see for example all of the browsers, none of which implement fallback by default. Chrome used to but changed to not do direct fallback.

If we have this behavior I think we should change the default on the fallback option, in effect adding a `PROXY_FALLBACK` instead that must be set to activate this behavior.

AudriusButkevicius: So this would be breaking everyone. Chrome is an actively used application, syncthing is mostly setup and forget. 

I don't think we should go for less connective defaults, as syncthings biggest problems (other than db voes) is not being able to connect. Hence relays, hence stun (which doesn't seem to do much), hence default behaviour of fallback.

lib/dialer/public.go
```diff
@@ -85,6 +92,9 @@ func dialContextWithFallback(ctx context.Context, fallback proxy.ContextDialer,
...
```
calmh: This needs to happen before we close the channel or it's a race

PR #6367, opened 2020-02-26
lib: Repair sequence inconsistencies

imsodin:
The activation by env var is intended to see if it fixes the problems in https://forum.syncthing.net/t/panic-non-increasing-sequence/14570. If it does, this should probably be done as a db transition (potentially keeping the env var for future debugging).

I tested this on my laptop ("production environment"), finding nothing to fix (and crashing nothing too...).


lib/db/lowlevel.go
```diff
@@ -604,6 +604,91 @@ func (db *Lowlevel) gcBlocks() error {
...
+					return err
+				}
+				continue
+			}
+			if seq := t.keyer.SequenceFromSequenceKey(dbi.Key()); seq == fi.SequenceNo() {
+				continue
+			}
+
+			// Mismatch: Bump sequence and rewrite.
+			fixed += 1
```
golangcibot: should replace `fixed += 1` with `fixed++` (from `golint`)

calmh:
Isn't this precisely the migration we have in 1.4.0-rc.7?

imsodin:
No, it's slightly different: There we iterate over the file infos, and check there's a corresponding sequence entry. Here we iterate over sequence entries, and check that the corresponding fileinfo has a matching file info. Those wold not be picked up by the former check.

calmh:
Right... So maybe the checker should walk both, to both find missing/wrong sequence entries and old sequence entries to be removed?

imsodin:
Maybe: I don't yet see a connection between the failure mechanism fixed in 1.4.0 and what I clean up in this PR: Sudden shutdown causing an incorrect current sequence, thus sequences entries getting overwritten would not leave behind sequence entries pointing at a file with a different sequence number. Quite the opposite, sequence entries are then missing.  
That's why I wanted to know whether this is even an issue for https://forum.syncthing.net/t/panic-non-increasing-sequence/14570. And then hopefully also find out what caused it.

calmh:
👍 

FWIW, idxck does this check also today

AudriusButkevicius:
So I have a PR brewing that adds a web ui for idxchk

calmh:
Do we really want or need that? It's a debugging tool that I'd hate to expose to end users, nor think they should need to use or care...

imsodin:
It's confirmed to problem of inconsistent sequence number between key and fileinfo exists: https://forum.syncthing.net/t/panic-non-increasing-sequence/14570/28

I still don't have any idea how it happened though. What do you think about adding a db transition to fix the existing inconsistencies, and then always doing the checks in https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L1988 on RCs?

AudriusButkevicius:
People have questions when things don't work, we point them to download links which is extra faf compared to just opening some modal in web ui to see if there are any inconsistencies.

But if you say we don't want it, then we don't want it, but my thinking was same as tracing toggles in the ui.

calmh:
Well the checker exists and has proven useful in debugging so I'm not saying you're wrong there. 

But I would hate for it to become a thing where it's user culture that whenever something seems odd you might need to run an index-fsck because Syncthing is some buggy crap that corrupts its database all the time and needs user intervention to debug and fix it.

I do not ever want "did you check the database for corruption" to be a reasonable first question someone might ask on the forum when something seems odd. If we're at that  point we might as well scrap the database layer, switch to badger, try again and hope for the best. The database shouldn't be something the user needs to maintain or ideally even know about...

AudriusButkevicius:
Sure, but at this point we don't know where the problem lies, so I understand the ambition to fix it, but in the meanwhile we need to tell people how to recover, and for that we need people to run things.

imsodin:
There's still a chance that the remaining cases are past relicts, and don't occur anymore. I know, wishful thinking, but anyway can we please have the debugging tool discussion somewhere else and focus on the problem at hand here:

> It's confirmed that the problem of inconsistent sequence numbers between key and fileinfo exists: https://forum.syncthing.net/t/panic-non-increasing-sequence/14570/28
> 
> I still don't have any idea how it happened though. What do you think about adding a db transition to fix the existing inconsistencies, and then always doing the checks in https://github.com/syncthing/syncthing/blob/master/lib/model/model.go#L1988 on RCs?

calmh:
Right! So, I'm pro having something to fix this, and running it on upgrade for historical setups. If we can somehow detect and record the fact that this happens, we could run the check/fix similar to the "metadata out of sync, recalculating" thing that happens since recently.

calmh:
Looking at an old, large database I see a lot of `duplicate sequence entry` and `sequence mismatch` errors from idxck. So yeah, we need this somewhere as a migration or periodic check...

imsodin:
Ok, so definitely a one time migration for these problems. Ideally we'd then also make the debug checks for sequence consistency when sending indexes non-optional and if detecting a problem, create a support bundle and db copy, and a warning message telling the user to report the issue, then run the sequence repairing again. That could also be extended to all the panicking in the db (minus the repairing): Save some flag that marks the db as tainted. On subsequent startup also create a db copy and a warning message.

I'll see that I can turn this PR into a db migration in the next days, for the rest I probably don't have time at the moment.

lib/db/schemaupdater.go
```diff
@@ -514,3 +511,10 @@ func (db *schemaUpdater) updateSchemato9(prev int) error {
 
 	return t.Commit()
 }
+
+// updateSchemato10 makes sure the sequence numbers in the sequence keys match
+// those in the corresponding file entries.
+func (db *schemaUpdater) updateSchemato10(_ int) error {
```
golangcibot: func `(*schemaUpdater).updateSchemato10` is unused (from `unused`)

imsodin:
I believe throwing a warning and repairing the db is a sane behaviour. If these inconsistencies are just are relict of the past, it will not happen, otherwise the verbosity will bring in reports soon. And zipping the db probably wont do any good anyway, it will just show what the warning already says.  
So ready to review from my side.

lib/db/lowlevel.go
```diff
@@ -604,6 +604,83 @@ func (db *Lowlevel) gcBlocks() error {
 	return db.Compact()
 }
 
+// repairSequence makes sure the sequence numbers in the sequence keys match
+// those in the corresponding file entries. It returns the amount of fixed entries.
+func (db *Lowlevel) repairSequence(folders []string) (int, error) {
+	t, err := db.newReadWriteTransaction()
```
calmh: I think this should hold the gc lock as well. Probably won't cause issues now, but it rewrites FileInfos and gets called at random times so in principle they are in the same territory.

lib/db/lowlevel.go
```diff
@@ -604,6 +604,83 @@ func (db *Lowlevel) gcBlocks() error {
...
+		}
+
+		if meta != nil {
+			if err := meta.toDB(t, folder); err != nil {
+				return 0, err
+			}
+		}
+	}
+
+	return fixed, nil
```
calmh: No commit so this does nothing, also no check of the iterator error

lib/db/lowlevel.go
imsodin: Done.

lib/db/db_test.go
```diff
@@ -249,6 +250,166 @@ func TestUpdate0to3(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	defer trans.close()
+
+	addFile := func(f protocol.FileInfo, seq int64) {
+		dk, err := trans.keyer.GenerateDeviceFileKey(nil, folder, id[:], []byte(f.Name))
+		if err != nil {
+			t.Fatal(err)
+		}
+		trans.putFile(dk, f)
```
golangcibot: Error return value of `trans.putFile` is not checked (from `errcheck`)

lib/db/schemaupdater.go
```diff
@@ -520,13 +467,21 @@ func (db *schemaUpdater) updateSchemato9(prev int) error {
...
 	db.recordTime(indirectGCTimeKey)
 
 	return t.Commit()
 }
+
+// updateSchemato10 makes sure the sequence numbers in the sequence keys match
+// those in the corresponding file entries.
+func (db *schemaUpdater) updateSchemato10(_ int) error {
+	db.gcMut.RLock()
+	defer db.gcMut.RUnlock()
```
calmh: FWIW the schema updates are for sure guaranteed to run before we start any gc, so this is unnecessary. I think we need to (mentally at least) always be sure that schema updates run before anything else starts touching the database.

imsodin:
While writing a test I realized that it doesn't really make sense to think of the db transition to 9 (checking every file entry has a corresponding sequence entry) as a separate thing from checking that sequence entries point at the right thing. Therefore I reverted that transition and moved those checks into the new function to repair sequences. Then did all of that in a new transition to 10. That means part the sequence repairing is done on upgrading to 1.4.0, and the full variant later on 1.4.1. That's clearly not ideal but in my opinion ok, as the part done in 1.4.0 addresses the fixed problem if metadata isn't written to disk on quick shutdown, while the full thing repairs defects of unknown origins.

lib/db/schemaupdater.go
```diff
@@ -20,21 +18,14 @@ import (
...
 const (
-	dbVersion             = 9
-	dbMinSyncthingVersion = "v1.4.0"
-)
-
-var (
-	errFolderIdxMissing = errors.New("folder db index missing")
-	errDeviceIdxMissing = errors.New("device db index missing")
+	dbVersion             = 10
+	dbMinSyncthingVersion = "v1.4.1"
```
calmh: Acksherly no need to bump the min version

lib/db/set.go
```diff
@@ -531,6 +531,12 @@ func (s *FileSet) ListDevices() []protocol.DeviceID {
 	return s.meta.devices()
 }
 
+func (s *FileSet) RepairSequence() (int, error) {
+	s.db.gcMut.RLock()
```
calmh: Should hold `s.updateMutex` here as well

lib/db/lowlevel.go
```diff
@@ -621,6 +621,110 @@ func (db *Lowlevel) gcIndirect() error {
...
+				return 0, err
+			}
+		}
+		if err := t.Checkpoint(func() error {
+			return meta.toDB(t, folder)
+		}); err != nil {
+			return 0, err
+		}
+	}
+
```
calmh: Check `it.Error()`

lib/db/lowlevel.go
```diff
@@ -621,6 +621,110 @@ func (db *Lowlevel) gcIndirect() error {
...
+				continue
+			}
+		}
+		// Either the file is missing or has a different sequence number
+		fixed++
+		if err := t.Delete(it.Key()); err != nil {
+			return 0, err
+		}
+	}
+
```
calmh: `it.Release()`, check `it.Error()`

lib/db/schemaupdater.go
calmh: Eh, looking closer on this I realized db schema versions don't actually work like I thought they worked. Indeed with this bump to 10 it won't be possible to open the database with 1.4.0 any more, even though the on disk format is in fact the same. We should have something like a compatibility version different from the db version, or actually compare the min db version to our running version to determine if we're ok to proceed... Possibly not for this PR though.

lib/db/lowlevel.go
AudriusButkevicius: Need to release first?

calmh: Technically, currently, no. *Badger* requires releasing snapshots/iterators before committing the transaction they were created in so I've adopted the pattern...

calmh: Oh, checking it.Error could be done either before or after it.Release with the same effect AFAIK

lib/db/schemaupdater.go
calmh: Thinking one step further I think this should maybe not be a "migration" at all as it doesn't migrate, rather it should be run with the metadata refresh (periodically and when crash is detected) and triggered as it is from the index sender?

lib/db/lowlevel.go
AudriusButkevicius: welll if we return early, the release does not happen?

calmh: There is a deferred release, and releasing twice is fine, it's just committing without release that is not guaranteed to be fine.

imsodin: And since the latest commit all should be fine in this regard for this PR, right?

calmh: yep

lib/db/schemaupdater.go
imsodin: Missed that comment:  
I wouldn't do it periodically, as it is rather heavy (iterates over all local file and sequence entries and unmarshalling every file on the way), but only triggered by crash and index sender. I'd still keep the migration, even though it's a misnomer, as otherwise users will get the warning in the UI for a problem that we know has existed and is mitigated with the migration.

calmh: If it were just a misnomer I'd be fine with it, but it forces a compatibility change and prevents 1.4.0 from opening the db. I think the GUI issue is similar to out of date metadata which we handle by the periodic metadata recalculation which is equally heavy? We could bump the key that stores that recalculation date, causing a recalc on upgrade without a migration.

imsodin: Ah right, the `dbMinSyncthingVersion` implementation doesn't actually do what it should as you wrote above. That would however be an easy fix, so I might just do that in this PR?

It is pretty much exactly the same as the metadata thing (well twice as heavy due to double pass). I agree it's doable, I am just hesitant to add periodic mitigation for something that isn't needed as far as we know today (there are no known sequence error modes and the index sender should pick up any errors which we aren't aware of) - even if the performance impact is reasonable.

calmh: Well there are no known failure modes to get the metadata out of sync either, yet we do that periodically, and similarly we transmit the whole index on upgrade for some historical reason... I think we should run this check when recovering from a crash, and then either on upgrade (every upgrade) or periodically (STRECHECKDBEVERY). Potentially the "recheck db every ..." should go away and just run on upgrades that too.

calmh: I'd rather not complicate the migration check, especially not in a 1.4.1 release. It's easier to make the non-migration be a non-migration :)

imsodin: done - periodically, because it keeps historic unknown weirdness together xD

calmh:
'cept not building 'n such

imsodin:
Yeah, pushing after committing apparently isn't optional.

Issue #6372, opened 2020-02-27
Reduce database size by optimizing version list storage


calmh:
It's somewhat rare in the grand scheme of things, but there are setups with hundreds of devices in a cluster. In some cases this results in files with hundreds of entries in the version vector. This is fine and expected. In our global version list we keep for each file an array of `(version vector, device, invalid)`. In the case where there are many devices and a long version vector we get that vector repeated for every device and this global entry can grow to hundreds of kilobytes for a single file entry.

Thing is though, in the steady state all these hundreds of version vectors are actually the same. So if we grouped the version list by version vector, with each item being `(version vector, (device, invalid)...)` we'd cut that by a factor hundreds.

In one database I'm looking at, the change would be from

```
 0x00: 1169574 items, 70491 KB keys +  5506325 KB data, 60 B +   4707 B avg,   7634 B max
 0x01:   52634 items,  3004 KB keys +  5412346 KB data, 57 B + 102829 B avg, 517727 B max
 0x02:  259848 items, 23082 KB keys +     1039 KB data, 88 B +      4 B avg,    256 B max
...
 Total 1581101 items, 98761 KB keys + 10935349 KB data.
```

to

```
 0x00: 1169574 items, 70491 KB keys + 5506325 KB data, 60 B +  4707 B avg,  7634 B max
 0x01:   52634 items,  3004 KB keys +  818664 KB data, 57 B + 15553 B avg, 56103 B max
 0x02:  259848 items, 23082 KB keys +    1039 KB data, 88 B +     4 B avg,   256 B max
...
 Total 1581101 items, 98761 KB keys + 6341667 KB data.
```

or about a 40% reduction in database size. Smaller setups aren't going to gain that, but it won't hurt either.

The actual file entries still contain all the copies of the version vectors, so class 0x00 is large. Theoretically we could do the same there as with the block lists and bring that down too, but that would actually hurt small setups with the extra indirection. Although we could just skip the indirection (storing it like we do now) if the version list is smaller than *x*, and have `getFile` handle both the version list being already in place or there being a hash pointer to it.

imsodin:
[...]

PR #6373, opened 2020-02-27
lib/db: Slightly improve indirection (ref #6372)

calmh:
### Purpose

I was working on indirecting version vectors, and that resulted in some
refactoring and improving the existing block indirection stuff. We may
or may not end up doing the version vector indirection, but I think
these changes are reasonable anyhow and will simplify the diff
significantly if we do go there. The main points are:

- A bunch of renaming to make the indirection and GC not about "blocks"
  but about "indirection".

- Adding a cutoff so that we don't actually indirect for small block
  lists. This gets us better performance when handling small files as it
  cuts out the indirection for quite small loss in space efficiency.

- Being paranoid and always recalculating the hash on put. This costs
  some CPU, but the consequences if a buggy or malicious implementation
  silently substituted the block list by lying about the hash would be bad.

**This may or may not be RC material depending on how we think about the change to indirecting small block lists and letting the old env var into the wild...**

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -19,22 +19,30 @@ import (
 )
 
 const (
-	// We set the bloom filter capacity to handle 100k individual block lists
-	// with a false positive probability of 1% for the first pass. Once we know
-	// how many block lists we have we will use that number instead, if it's
-	// more than 100k. For fewer than 100k block lists we will just get better
-	// false positive rate instead.
-	blockGCBloomCapacity          = 100000
-	blockGCBloomFalsePositiveRate = 0.01 // 1%
-	blockGCDefaultInterval        = 13 * time.Hour
-	blockGCTimeKey                = "lastBlockGCTime"
+	// We set the bloom filter capacity to handle 100k individual items with
+	// a false positive probability of 1% for the first pass. Once we know
+	// how many items we have we will use that number instead, if it's more
+	// than 100k. For fewer than 100k items we will just get better false
+	// positive rate instead.
+	indirectGCBloomCapacity          = 100000
+	indirectGCBloomFalsePositiveRate = 0.01 // 1%
+	indirectGCDefaultInterval        = 13 * time.Hour
+	indirectGCTimeKey                = "lastIndirectGCTime"
+
+	// Use indirection for the block list when it exceeds this many entries
+	blocksIndirectionCutoff = 3
 )
 
-var blockGCInterval = blockGCDefaultInterval
+var indirectGCInterval = indirectGCDefaultInterval
 
 func init() {
+	// deprecated
 	if dur, err := time.ParseDuration(os.Getenv("STGCBLOCKSEVERY")); err == nil {
-		blockGCInterval = dur
+		indirectGCInterval = dur
+	}
+	// current
+	if dur, err := time.ParseDuration(os.Getenv("STGCINDIRECTEVERY")); err == nil {
+		indirectGCInterval = dur
 	}
 }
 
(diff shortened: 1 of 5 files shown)


lib/db/transactions.go
```diff
@@ -453,26 +456,28 @@ func (t readWriteTransaction) close() {
...
 
-func (t readWriteTransaction) putFile(key []byte, fi protocol.FileInfo) error {
-	if fi.Blocks != nil {
-		if fi.BlocksHash == nil {
-			fi.BlocksHash = protocol.BlocksHash(fi.Blocks)
-		}
-		blocksKey := t.keyer.GenerateBlockListKey(nil, fi.BlocksHash)
-		if _, err := t.Get(blocksKey); backend.IsNotFound(err) {
+func (t readWriteTransaction) putFile(fkey []byte, fi protocol.FileInfo) error {
+	var bkey []byte
```
imsodin: Why is this declared out here? Isn't (re-) used outside of the if scope below as far as I can see.

lib/db/transactions.go
```diff
@@ -723,15 +728,12 @@ func (t *readWriteTransaction) withAllFolderTruncated(folder []byte, fn func(dev
...
 			continue
 		}
-		var f FileInfoTruncated
-		// The iterator function may keep a reference to the unmarshalled
-		// struct, which in turn references the buffer it was unmarshalled
-		// from. dbi.Value() just returns an internal slice that it reuses, so
-		// we need to copy it.
-		err := f.Unmarshal(append([]byte{}, dbi.Value()...))
+
+		intf, err := t.unmarshalTrunc(dbi.Value(), true)
```
imsodin: Isn't the commend above valid (anymore)? `unmarshalTrunc` doesn't do any buffer copying.

lib/db/transactions.go
```diff
@@ -453,26 +456,28 @@ func (t readWriteTransaction) close() {
...
-			if err := t.Put(blocksKey, blocksBs); err != nil {
+			if err := t.Put(bkey, blocksBs); err != nil {
 				return err
 			}
 		} else if err != nil {
 			return err
 		}
+		fi.Blocks = nil
+	} else {
+		fi.BlocksHash = nil
```
AudriusButkevicius: I don't like this. I think this is a generally useful piece of information to have. Perhaps we should check if the block list is empty when loading instead to decide if we need to take an indirection or not.

lib/db/lowlevel.go
```diff
@@ -19,22 +19,30 @@ import (
...
+
+	// Use indirection for the block list when it exceeds this many entries
+	blocksIndirectionCutoff = 3
 )
 
-var blockGCInterval = blockGCDefaultInterval
+var indirectGCInterval = indirectGCDefaultInterval
 
 func init() {
+	// deprecated
```
AudriusButkevicius: Lost the plot a bit, but is this actually released already?

lib/db/transactions.go
```diff
@@ -453,26 +456,28 @@ func (t readWriteTransaction) close() {
...
-	if fi.Blocks != nil {
-		if fi.BlocksHash == nil {
-			fi.BlocksHash = protocol.BlocksHash(fi.Blocks)
-		}
-		blocksKey := t.keyer.GenerateBlockListKey(nil, fi.BlocksHash)
-		if _, err := t.Get(blocksKey); backend.IsNotFound(err) {
+func (t readWriteTransaction) putFile(fkey []byte, fi protocol.FileInfo) error {
+	var bkey []byte
+
+	if len(fi.Blocks) > blocksIndirectionCutoff {
```
AudriusButkevicius: Does idxchk need handling of this?

lib/db/transactions.go
AudriusButkevicius: Also, avoid adding to bloom filter if it has a blockhash but also a blocklist in that case?

calmh: We could. You mean that we should always calculate, set and send the hash, even when we don't need it internally?

lib/db/lowlevel.go
calmh: It's in the RC. So if we delay the 1.4.0 release and put this into another RC this can go away, otherwise it kinda needs to be there.

lib/db/transactions.go
calmh: Not really. It would, if we did your change to always have the hash, but as is now the idxck checks the hash if it's present and otherwise not.

calmh:
Changed to always set the hash regardless

lib/db/transactions.go
imsodin: I sort of agree, even though the review comment below suggests the opposite: I feel like `BlocksHash` should either be a db internal implementation detail or always be there if exposed towards protocol.

lib/db/transactions.go
calmh: It was in my original which also indirected the version vector... I figured it could stay there for clarity, but it has no practical reason now, no.

lib/db/transactions.go
calmh: It wasn't valid at all since we switched to protobuf, which doesn't keep references (or we'd be fucked in so many places...).

This change in general isn't strictly needed right now, but I think all unmarshalling should use the unmarshalTrunc function. (Again, when indirecting version vectors that also affects the truncated version.)

AudriusButkevicius:
I'd say delay it, or release the rc with the env var changed so we're not releasing legacy straight away.

PR #6385, opened 2020-02-28
cmd/syncthing: Do auto-upgrade before startup (fixes #6384)

### LINKED ISSUE #6384: Consider "synchronous" upgrades of release candidates
Currently we fire off the auto upgrade routine in the background on startup, and it maybe downloads and upgrade a little later, and then a restart is scheduled a minute later. This is usually fine, but sometimes we fuck up in release candidates and they crash. If that crash is early & persistent then the auto upgrade won't kick in and the user is stuck.

Maybe, potentially, we should consider doing the upgrade check early and before doing other stuff.

imsodin:
From https://github.com/syncthing/syncthing/issues/6384:

> This is usually fine, but sometimes we fuck up in release candidates and they crash. If that crash is early & persistent then the auto upgrade won't kick in and the user is stuck.
> 
> Maybe, potentially, we should consider doing the upgrade check early and before doing other stuff.

Of course it will never happen again that I produce a panic-inducing PR (ahem...), but nevertheless I believe that we definitely should do that.

I tested `-upgrade-check` `-upgrade` and with and without `STNOUPGRADE=1`: It all did the expected.

I made `checkUpgradeExitOnErr` a function variable instead of a global function, as I believe no functions except the `main...` ones should ever call `os.Exit`. If it is considered ugly, I'd rather go for the little bit of duplication of explicitly handling the errors in the two upgrade ifs below that.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -358,15 +358,31 @@ func main() {
 		return
 	}
 
+	checkUpgradeExitOnErr := func() upgrade.Release {
+		release, err := checkUpgrade()
+		if err != nil {
+			l.Warnln("Upgrade:", err)
+			if _, ok := err.(errNoUpgrade); ok {
+				os.Exit(syncthing.ExitNoUpgradeAvailable.AsInt())
+			}
+			os.Exit(syncthing.ExitError.AsInt())
+		}
+		return release
+	}
+
 	if options.doUpgradeCheck {
-		checkUpgrade()
+		_ = checkUpgradeExitOnErr()
 		return
 	}
 
 	if options.doUpgrade {
-		release := checkUpgrade()
-		performUpgrade(release)
-		return
+		release := checkUpgradeExitOnErr()
+		if err := performUpgrade(release); err != nil {
+			l.Warnln("Upgrade:", err)
+			os.Exit(syncthing.ExitError.AsInt())
+		}
+		l.Infof("Upgraded to %q", release.Tag)
+		os.Exit(syncthing.ExitUpgrade.AsInt())
 	}
 
 	if options.resetDatabase {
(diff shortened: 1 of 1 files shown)


cmd/syncthing/main.go
```diff
@@ -568,6 +588,46 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
...
+	// unless we are in a build where it's disabled or the STNOUPGRADE
+	// environment variable is set.
+
+	if build.IsCandidate && !upgrade.DisabledByCompilation && !runtimeOptions.NoUpgrade {
+		l.Infoln("Automatic upgrade is always enabled for candidate releases.")
+		if opts := cfg.Options(); opts.AutoUpgradeIntervalH == 0 || opts.AutoUpgradeIntervalH > 24 {
+			opts.AutoUpgradeIntervalH = 12
+			// Set the option into the config as well, as the auto upgrade
+			// loop expects to read a valid interval from there.
+			cfg.SetOptions(opts)
```
golangcibot: Error return value of `cfg.SetOptions` is not checked (from `errcheck`)

cmd/syncthing/main.go
```diff
@@ -568,6 +588,46 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
...
+			cfg.Save()
```
golangcibot: Error return value of `cfg.Save` is not checked (from `errcheck`)

cmd/syncthing/main.go
```diff
@@ -358,15 +360,31 @@ func main() {
 		return
 	}
 
+	checkUpgradeExitOnErr := func() upgrade.Release {
```
calmh: I think this should be a normal function despite the given rationale. Closures have a significant mental cost, at least for me, which is not necessary in this case

calmh: Or maybe just inline this and have a function `exitCodeForUpgrade(error) int` that does the error checking and returns the appropriate exit code, as that's the only real logic, and do the exit inline in main

cmd/syncthing/main.go
```diff
@@ -568,6 +588,46 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
...
+			if err == nil {
+				if err = performUpgradeDirect(release); err == nil {
+					l.Infof("Upgraded to %q, exiting now.", release.Tag)
+					os.Exit(syncthing.ExitUpgrade.AsInt())
+				}
+			}
+			if err != upgrade.ErrUpgradeUnsupported {
+				shouldAutoUpgrade = true
+			} else if err != nil {
+				if _, ok := err.(errNoUpgrade); !ok {
```
calmh: This branch will only be taken if err == ErrUpgradeUnsupported, or my double negation logic parser is bad (in which case this also needs simplification I think, as I've stared at these if clauses now for like half a minute)

cmd/syncthing/main.go
```diff
@@ -614,35 +674,14 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
...
-		} else {
-			go autoUpgrade(cfg, app, evLogger)
-		}
-	}
-
 	if err := app.Start(); err != nil {
 		os.Exit(syncthing.ExitError.AsInt())
 	}
 
+	if shouldAutoUpgrade {
```
calmh: I think this is too far removed from where we set the variable. Better to start that routine directly and not have to carry the `shouldAutoUpgrade` state along?

imsodin: Can't be done directly, as the initial upgrade should happen before initialising db and app. Moved it up to right after `App` is created.

cmd/syncthing/main.go
imsodin: Thanks for the catch. And I refactored it, should be somewhat less convoluted now (still quite a few ifs though).

cmd/syncthing/main.go
```diff
@@ -810,6 +807,17 @@ func standbyMonitor(app *syncthing.App) {
 	}
 }
 
+func shouldUpgrade(cfg config.Wrapper, runtimeOptions RuntimeOptions) bool {
+	if opts := cfg.Options(); opts.AutoUpgradeIntervalH < 0 {
```
calmh: We could bail early here if `upgrade.DisabledByCompilation` and then not have to handle `ErrUpgradeUnsupported` in the auto upgrade case?

AudriusButkevicius:
I'll have a look in front of a larger screen

cmd/syncthing/main.go
```diff
@@ -568,6 +582,45 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
...
+		release, err := checkUpgrade()
+		if err == nil {
+			if err = performUpgradeDirect(release); err == nil {
+				l.Infof("Upgraded to %q, exiting now.", release.Tag)
+				os.Exit(syncthing.ExitUpgrade.AsInt())
+			}
+		}
+		// Log the error if relevant.
+		if err != nil {
+			if _, ok := err.(errNoUpgrade); !ok {
```
AudriusButkevicius: Should this not be `err.(*errNoUpgrade)` ?  As err is effectively an interface?
Did you verify this works?

imsodin: I checked that it works. `errNoUpgrade` is returned, not a pointer. An interface variable need not be a pointer, it just has to implement the interface as a value or pointer. And usually methods are implemented with pointer receivers, but it is customary to use value receivers for custom errors (I guess because you never change them after creating).

AudriusButkevicius: as I said, lgtm otherwise

PR #6387, opened 2020-02-28
build: fix syso creation (fixes #6386)

### LINKED ISSUE #6386: Windows build resource.syso creation fails since v1.4.0.rcX
From https://forum.syncthing.net/t/windows-build-resource-syso-creation-fails-since-v1-4-0-rcx/14588

Hi there,

I've just tried to build syncthing.exe on Windows with

    [INFO] go version
    go version go1.13.4 windows/amd64
    [INFO] go get github.com/josephspurrier/goversioninfo
    go: finding github.com/josephspurrier/goversioninfo latest
    [INFO] go install github.com/josephspurrier/goversioninfo/cmd/goversioninfo
    [INFO] go mod download
    [INFO] go run build.go -no-upgrade -goos windows -goarch amd64 build syncthing
    Warning: Windows binaries will not have file information encoded: failed to create C:\Temp\git\SYNCTH~1\cmd\syncthing\resource.syso: exit status 2

... and it fails with "status 2". The file "resource.syso" doesn't get created but I get a "syncthing.exe" without a resource manifest that can be started properly. I did checkout v1.4.0-rc.1 to test this.

When checking out at v1.3.4 (release) the build fully succeeds and the exe icon (indicating the resource.syso was successfully compiled) appears in Windows Explorer. No error message so far during build.

I've narrowed it down to the commit to blame for this behaviour:
https://github.com/syncthing/syncthing/commit/b4064e07dce0268c8d228ed507ae3cf2b4c8b432

When I do:

    git checkout v1.4.0-rc.1 (or v1.4.0-rc.9)
    git revert b4064e07dce0268c8d228ed507ae3cf2b4c8b432
    go run build.go -no-upgrade -goos windows -goarch amd64 build syncthing

It builds fine with the exe resources included and icon shows up.

image|546x453 

Should I raise the bug to GitHub?

Kind regards
Catfriend1

#

> nekr0z
> 
> Which is strange, because the field i’m looking at is “24”, so it should be parseable as int.
> 
> OK, I see the issue, will fix. Still, do open an issue.

nekr0z:
### Purpose

Pass `int`s instead of `string`s where `goversioninfo` wants them.

### Testing

Tested manually, works as expected.

### DIFF
--- a/build.go
+++ b/build.go
@@ -631,14 +631,25 @@ func buildSnap(target target) {
 
 func shouldBuildSyso(dir string) (string, error) {
 	type M map[string]interface{}
-	major, minor, patch, build := semanticVersion()
+	var vStr [4]string
+	vStr[0], vStr[1], vStr[2], vStr[3] = semanticVersion()
+
+	var vInt [4]int
+	for i, v := range vStr {
+		ver, err := strconv.Atoi(v)
+		if err != nil {
+			return "", err
+		}
+		vInt[i] = ver
+	}
+
 	bs, err := json.Marshal(M{
 		"FixedFileInfo": M{
 			"FileVersion": M{
-				"Major": major,
-				"Minor": minor,
-				"Patch": patch,
-				"Build": build,
+				"Major": vInt[0],
+				"Minor": vInt[1],
+				"Patch": vInt[2],
+				"Build": vInt[3],
 			},
 		},
 		"StringFileInfo": M{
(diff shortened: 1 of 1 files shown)


nekr0z:
> Ugh. Sorry about that. Maybe semanticVersion should return ints instead?

Yep, you're right, it's not used anywhere else. I should have checked. Will rewrite in a minute.

PR #6390, opened 2020-02-29
lib/db: Prevent GC concurrently with migration (fixes #6389)

### LINKED ISSUE #6389: Database migration can lead to missing blocks error
There is a problem in the latest database migration, where 

1. the block GC can run concurrently with the migration, and
2. the block GC doesn't properly use a snapshot for all its operations.

If the database is large then the block GC routine will take time and read FileInfos (none of which have block hashes yet) while the migration runs. The migration creates block list entries in the meantime. Then the block GC starts iterating block lists, without a snapshot, so it sees the new entries and deletes them because the FileInfos didn't refer to them when it saw those.

The end result is erroneously deleted block lists.

calmh:
Also, holy fuck... 😬

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -488,8 +488,18 @@ func (db *Lowlevel) dropPrefix(prefix []byte) error {
 }
 
 func (db *Lowlevel) gcRunner() {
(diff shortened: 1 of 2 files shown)


calmh:
There. Final (!?) cause for missing block lists nailed, maybe.

PR #6391, opened 2020-03-01
all: Tweak error creation

calmh:
- In the few places where we wrap errors, use the new Go 1.13 "%w"
  construction instead of %s or %v.

- Where we create errors with constant strings, consistently use
  errors.New and not fmt.Errorf.

- Remove capitalization from errors in the few places where we had that.

### DIFF
--- a/lib/rc/rc.go
+++ b/lib/rc/rc.go
@@ -232,7 +232,7 @@ func (p *Process) Events(since int) ([]Event, error) {
 	dec.UseNumber()
 	err = dec.Decode(&evs)
 	if err != nil {
-		return nil, fmt.Errorf("events: %s in %q", err, bs)
+		return nil, fmt.Errorf("events: %w in %q", err, bs)
 	}
 	return evs, err
 }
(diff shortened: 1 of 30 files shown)


lib/rc/rc.go
```diff
@@ -232,7 +232,7 @@ func (p *Process) Events(since int) ([]Event, error) {
 	dec.UseNumber()
 	err = dec.Decode(&evs)
 	if err != nil {
-		return nil, fmt.Errorf("Events: %s in %q", err, bs)
+		return nil, fmt.Errorf("events: %s in %q", err, bs)
```
AudriusButkevicius: %w?

calmh: Requires turning the error around, %w only works as the last verb. This is internal so I didn't bother.

AudriusButkevicius: "only works with last verb" lol wat.

calmh: It used to be:

> If the last argument is an error and the format string ends with ": %w", the returned error implements an Unwrap method returning it. (https://godoc.org/golang.org/x/xerrors#Errorf)

apparently that was relaxed in the final implementation

Issue #6249, opened 2020-01-04
Remove unnecessary RAM/CPU stats from GUI


gabriel-fallen:
As long as (big) files synchronization (especially initial one) is IO-intensive task I'd like to see IO (hard drive) utilization in the Web UI along the CPU and memory lines.

---

Subject edited from initial "add I/O stats" to "remove existing crap", adding note here as the initial post is otherwise rather confusing // @calmh

calmh:
I’m very much in favor of doing the opposite and removing CPU/RAM from the UI instead. Every operating system has utilities to report on these things and there is no real reason we should be duplicating it I think?

AudriusButkevicius:
Sort of agree, also, it's usually misreported.

IO in general doesn't even have an API for most OSes

tomasz1986:
> I’m very much in favor of doing the opposite and removing CPU/RAM from the UI instead. Every operating system has utilities to report on these things and there is no real reason we should be duplicating it I think?

Being able to see how much CPU and RAM is used by Syncthing does come in handy when administering it remotely through a Web browser though.

gabriel-fallen:
@calmh removing all the system stats altogether is an option too.

On the other hand being very well aware of `htop` and using it on a regular basis I still found it handy that Syncthing shows its own resource consumption right in the Web UI and I somewhat missed the IO stats.

PR #6393, opened 2020-03-03
remove CPU & RAM measurements from syncthing GUI

jesselucas:
### Purpose

Remove CPU & RAM measurements from GUI to address issue #6249

### Testing

Test UI in browser

### Screenshots

Screen Shot 2020-03-03 at 3 54 36 PM

## Authorship

based on the commit metadata.


calmh:
I think we should also remove the CPU measurement code from the backend and report that stat as zero in the API. Whether that constitutes and API change that should trigger a 1.5 release depends on how loud people yell... I personally don't care much. We should add a note to the relevant REST API docs though documenting that this happened.

The RAM usage stuff goes into the usage report so should be retained as-is I think, and so can stay in the API as well...

man/syncthing-faq.7
```diff
@@ -128,12 +128,6 @@ firewall\-setup to enable direct connections.
 Second, if one of the devices is a very low powered machine (a Raspberry Pi,
 or a phone, or a NAS, or similar) you are likely constrained by the CPU on
 that device. See the next question for reasons Syncthing likes a faster CPU.
-You can verify this by looking at the CPU utilization in the GUI. If it is
```
calmh: The man pages are auto generated from the syncthing/docs repo, so the change should happen there instead.

lib/api/api.go
```diff
@@ -81,7 +81,6 @@ type service struct {
 	fss                  model.FolderSummaryService
 	urService            *ur.Service
 	systemConfigMut      sync.Mutex // serializes posts to /rest/system/config
-	cpu                  Rater
```
calmh: Is this Rater interface used anywhere else? Otherwise kill it

jesselucas: Good catch!

Issue #5957, opened 2019-08-15
"Local discovery" broadcast breaks mobile/cellular connection


eugenesan:
"Local discovery" broadcast breaks mobile/cellular connection of at least one Cellular operator (Visible @ USA).

It seems the cellular network or provider-specific services in the OS are "allergic" to the broadcast. Changing the port number did not help.

I would suggest limiting local discovery functionality for use on WiFi only.

Maybe the daemon should have a runtime/command-line option to override the feature configured in permanent config so the wrapper/GUI will have the option to detect the condition and pass it to the daemon.

Syncthing Version: v1.2.0
OS Version: Android 9.0
CellOperator: Visible@USA
Browser Version: Android GUI

AudriusButkevicius:
First of all, wrong repo, second of all, you can disable local discovery from the web ui.
Thirdly, your provider should get their shit together.

eugenesan:
First, thank you for such a prompt response.
Second, why is it a wrong repo?
Currently, the daemon supports either enabling the feature or disabling it. On mobile platforms, where type the connection might change on the fly, regardless of the "shittiness" of the ISP, differentiating between types of connections would be a useful addition. Why send local discovery broadcast when not on local network?

P.S. I agree, the provider is at fault for failing with UDP broadcasts.

AudriusButkevicius:
Because syncthing itself has no concept of different types of connections, so if this was due to be fixed, it would have to be fixed in the android wrapper to toggle local discovery based on the connection type.

Yet, this is a network provider issue, and I'd be against adding random workaround into an application to work around some one bad provider.

There is no harm in sending broadcasts all the time, after all, they are broadcasts, looking for things regardless where you are connected to.

Perhaps there is an equally bad provider out there, where people connected to the same tower happen to be in the same LAN with broadcasts enabled 🤷‍♂

PR #6404, opened 2020-03-13
lib/beacon: only send *cast to flagged interfaces

Hello71:
### Purpose

saves some traffic (potential mobile wakeups), may help with #5957.

### Testing

built and run on Linux, seems to work correctly (broadcasts still sent on Ethernet interfaces, not sent on WireGuard interfaces)

### Documentation

I don't think this change requires documentation beyond a changelog message.

### DIFF
--- a/lib/beacon/multicast.go
+++ b/lib/beacon/multicast.go
@@ -67,6 +67,10 @@ func writeMulticasts(ctx context.Context, inbox <-chan []byte, addr string) erro
 
 		success := 0
 		for _, intf := range intfs {
+			if intf.Flags&net.FlagMulticast == 0 {
+				continue
+			}
+
 			wcm.IfIndex = intf.Index
 			pconn.SetWriteDeadline(time.Now().Add(time.Second))
 			_, err = pconn.WriteTo(bs, wcm, gaddr)
--- a/lib/beacon/broadcast.go
+++ b/lib/beacon/broadcast.go
@@ -44,17 +44,29 @@ func writeBroadcasts(ctx context.Context, inbox <-chan []byte, port int) error {
 			return nil
 		}
 
-		addrs, err := net.InterfaceAddrs()
+		intfs, err := net.Interfaces()
 		if err != nil {
 			l.Debugln(err)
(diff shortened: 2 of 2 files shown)


Hello71:
the v4 version is a little bit pyramidy, might be worth refactoring it to use continue.

lib/beacon/multicast.go
```diff
@@ -67,24 +67,26 @@ func writeMulticasts(ctx context.Context, inbox <-chan []byte, addr string) erro
...
-
-			success++
-
-			select {
-			case <-doneCtx.Done():
-				return nil
-			default:
+			if intf.Flags & net.FlagMulticast != 0 {
+				wcm.IfIndex = intf.Index
+				pconn.SetWriteDeadline(time.Now().Add(time.Second))
```
golangcibot: Error return value of `pconn.SetWriteDeadline` is not checked (from `errcheck`)

lib/beacon/multicast.go
```diff
@@ -67,24 +67,26 @@ func writeMulticasts(ctx context.Context, inbox <-chan []byte, addr string) erro
...
+				_, err = pconn.WriteTo(bs, wcm, gaddr)
+				pconn.SetWriteDeadline(time.Time{})
```
golangcibot: Error return value of `pconn.SetWriteDeadline` is not checked (from `errcheck`)

calmh:
Doesn't build so I question the testing, but looks good in principle.

Hello71:
ah, messed up my operators when I was trying to appease the CI warnings. the lesson is to always test your trivial refactorings...

lib/beacon/multicast.go
```diff
@@ -67,6 +67,10 @@ func writeMulticasts(ctx context.Context, inbox <-chan []byte, addr string) erro
 
 		success := 0
 		for _, intf := range intfs {
+			if intf.Flags & net.FlagMulticast == 0 {
```
calmh: These need gofmt, otherwise reasonable

Issue #5380, opened 2018-12-16
gui: Display folder/device name in modal


uok:
It would be great if the folder/device modal also displays the name of selected folder/device in the modal header. The modal overlays the folder/device list and also e.g. if you switch to another tab, you have no clue which folder you are editing.

Suggested changes for modal headers
Edit folder -> Edit folder: {foldername}
Edit device -> Edit device: {devicename}
Restore Versions -> Restore Versions: {foldername}

maybe this could also be helpful for adding new folder/device once name is entered
Add folder -> Add folder: {newfoldername}
Add device -> Add device: {newdevicename}

Should be quite easy to implement

PR #6405, opened 2020-03-13
gui: Display device name in modal

Shidory:
Edit device name in modal


calmh:
Hi!

Thanks for the contribution! This is a good start, but we should also have the same for edited folders and not just edited devices. As for the formatting, I'd suggest not adding the colon in the translated string, instead just adding the device in `()` parenthesis when it's there. Also you might want to use the `deviceName()` function which grabs the name when there is one and otherwise shows the short device ID.

Shidory:
Oh, thanks. I'll see to it.

uok:
Thanks, this fixes #5380

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -1,4 +1,4 @@
-<modal id="editDevice" status="default" icon="{{editingExisting ? 'fas fa-pencil-alt' : 'fas fa-desktop'}}" heading="{{editingExisting ? 'Edit Device' : 'Add Device' | translate}}" large="yes" closeable="yes">
+<modal id="editDevice" status="default" icon="{{editingExisting ? 'fas fa-pencil-alt' : 'fas fa-desktop'}}" heading="{{editingExisting ? 'Edit Device : ' : 'Add Device' | translate}} {{currentDevice.name}}" large="yes" closeable="yes">
```
calmh:
```suggestion
<modal id="editDevice" status="default" icon="{{editingExisting ? 'fas fa-pencil-alt' : 'fas fa-desktop'}}" heading="{{editingExisting ? 'Edit Device' : 'Add Device' | translate}} {{currentDevice.name}}" large="yes" closeable="yes">
```

PR #6406, opened 2020-03-13
gui: name in dialog box

keevBush:
display name on modal restore version

PR #6409, opened 2020-03-15
Certfile keyfile generated

keevBush:
- Display error when failed load keypair
- Generate keypair when cert.pem and key.pem files exist both

### DIFF
--- a/cmd/stdiscosrv/main.go
+++ b/cmd/stdiscosrv/main.go
@@ -67,6 +67,13 @@ var (
 	debug = false
 )
 
+func fileExists(filename string) bool {
+	_, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
 func main() {
 	var listen string
 	var dir string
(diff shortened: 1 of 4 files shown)


keevBush:
Last work to display Folder name into a restoreversion modal is in this pull request

imsodin:
Please keep the two separated, i.e. dialog naming in #6406 and the key errors/generation in this PR.

cmd/stdiscosrv/main.go
```diff
@@ -99,11 +106,15 @@ func main() {
 
 	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
 	if err != nil {
-		log.Println("Failed to load keypair. Generating one, this might take a while...")
-		cert, err = tlsutil.NewCertificate(certFile, keyFile, "stdiscosrv", 20*365)
-		if err != nil {
-			log.Fatalln("Failed to generate X509 key pair:", err)
+		log.Println("Failed to load keypair because:", err)
+		if !fileExists(certFile) && !fileExists(keyFile) {
```
calmh: I'd go with just `if os.IsNotExist(err)` here; that should be sufficient

keevBush: for cert.pem or for key.pem or for both files?

calmh: It's ok to just check the returned error from `LoadX509KeyPair`. If it's `os.IsNotExist` then we know that either the key or the cert didn't exist, and we can generate new ones.

keevBush: I fix that

gui/default/syncthing/folder/restoreVersionsModalView.html
```diff
@@ -1,4 +1,4 @@
-<modal id="restoreVersions" status="default" icon="fas fa-undo" heading="{{'Restore Versions' | translate}}" large="yes" closeable="yes">
```
calmh: gui changes should still not be in this PR

keevBush: It's on another branch. "dialog-titre"

PR #6411, opened 2020-03-15
lib/db: Don't ignore failures unmarshaling version lists

imsodin:
Doing other db work I noticed that `VersionList` is the only type where unmarshalling errors and invalid contents (empty) is just accepted and not acted on (neither fixed nor an error returned). For anything else, we just return the error. Looking at the history I don't see anything that justifies a special treatment for `VersionList`. Checkglobal now deletes invalid global entries and everywhere else they are just unmarshalled and an error is returned if invalid.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -380,9 +380,11 @@ func (db *Lowlevel) checkGlobals(folder []byte, meta *metadataTracker) error {
 
 	var dk []byte
 	for dbi.Next() {
-		vl, ok := unmarshalVersionList(dbi.Value())
-		if !ok {
-			continue
+		var vl VersionList
+		if err := vl.Unmarshal(dbi.Value()); err != nil || len(vl.Versions) == 0 {
+			if err := t.Delete(dbi.Key()); err != nil {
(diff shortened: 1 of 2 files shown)


calmh:
There's a subtle difference in that before in these places we skipped/continued for empty version lists, whereas now we just continue processing. I didn't look to see if the behavior is the same (I guess we'll typically loop over the versions, which is still a no-op), and I guess we generally shouldn't have any empty version lists either...

calmh:
Right, it'll panic here and there if there are empty lists. Which there shouldn't be. But then we'll see it in sentry. So, Ok, I guess.

imsodin:
> Right, it'll panic here and there if there are empty lists. Which there shouldn't be. But then we'll see it in sentry. So, Ok, I guess.

Sorry, should have made that explicit: It does panic and I considered that ok, because it must never happen and if it does, it will have other consequences (db inconsistencies) and I'd rather know about them to have a chance of fixing it. Maybe it would be cleaner to check for empty version lists and return an error, instead of panicking due to an index out of range (the end result will still be a panic).

PR #6402, opened 2020-03-10
lib/db: Extend set test with second remote

imsodin:
Unit Testing Only

I have a PR brewing with significant db changes (well finished, but sort of in "self-review") and as part of that I extend test coverage. In the large catch-all unit test checking global/local/remote/needed/... I added a second remote into the fray. Also I set sequences on all file infos, not just the local ones.

Issue #5899, opened 2019-07-28
Database performance: reduce iteration


calmh:
We do essentially full table scans to figure out stuff like remove device completion status and some other stuff. This sucks and should be fixed so that we update that info on database changes.

This means database updates will be more expensive but removes issues like having to have the folder summary service and folders looking "unknown" for a long time when the GUI first loads.

PR #6413, opened 2020-03-15
lib/db, lib/model: Keep need stats in metadata (ref #5899)

imsodin:
This is the "main thing" from which I separated #6411 and #6412 (still in this PR as the first two commits). I also kept the rest as three commits for now, as they address separate issues and thus makes reviewing potentially easier - they should still be merged together in the end.

### Purpose

The purpose of this PR is to get rid of the frequent db access due to folder summaries/API completion queries. These require the global list and unmarshalling lots of stuff. I added a new metadata flag for needed files. Thus when a file is added to the db, that flag is updated and the summary/db query can just directly access those values from memory. Only if the actual list of needed files is required (i.e. user opens the list in UI) the db needs to be accessed.

### Testing

I extended tests in #6402 to cover more ground on "needs". And I now extended all these tests to look at both the metadata and db when checking needs.

I have also done some manual testing a while ago and before merging want to run this on my main setup.


calmh:
Zomg

lib/db/schemaupdater.go
```diff
@@ -563,6 +563,12 @@ func (db *schemaUpdater) updateSchemato10(_ int) error {
 	for _, folderStr := range db.ListFolders() {
 		folder := []byte(folderStr)
 
+		// Update the existing metadata if it can be read, otherwise
+		// it will get recreated later in normal operations.
+		meta := newMetadataTracker()
+		metaErr := meta.fromDB(db.Lowlevel, []byte(folder))
+		meta.resetAll(protocol.GlobalDeviceID) // Is rewritten in updateGlobal
```
imsodin: I believe this is wrong: Updating global will do nothing except updating the `Deleted` field of in global list entries, as the versions won't change (thus no change in need or global counts). I probably need to not touch global  and recalc needs only. And I definitely want to write a test for the transition :)

calmh:
Does this transform every file `put` into n `get`s, one for each device in the setup? So if I have a thousand devices, every update becomes a thousand reads plus, potentially, a thousand puts of updated metadata?

imsodin:
Database access stays the same, max two: The new and old global file (and usually only one, as the inserted one is already the global one in most cases). What changes is that now the `VersionList`, that is anyway loaded, gets checked for every connected device and the metadata updated. So when the global version gets higher, there will be need checks and metadata bumps for every device, but that's in-memory only.

But good point, need to check benchmarks, didn't do that yet.

lib/db/schemaupdater.go
imsodin: It is good (wrote the test) - a few lines above I empty the global list (because it wouldn't do anything otherwise, not even update the deleted field as the version doesn't change).

imsodin:
I added a benchmark for updating with lots of remotes: There's 10 additional remotes with half of them on the old version, the other half on the new version. Then I update another device (local or a remote) between those two sets of files (as in existing benchs). I don't see any relevant change running it for 10s. While I expected no significant change (as db operations should be much costlier than a single loop over all devices with a few comparisons and writes to memory), they seem suspiciously close:

master:
```
BenchmarkUpdate100Changed10Remotes
BenchmarkUpdate100Changed10Remotes-4         	14577836	      1601 ns/op	     448 B/op	      16 allocs/op
BenchmarkUpdate100ChangedRemote10Remotes
BenchmarkUpdate100ChangedRemote10Remotes-4   	14450607	      1723 ns/op	     448 B/op	      15 allocs/op
```

PR:
```
BenchmarkUpdate100Changed10Remotes
BenchmarkUpdate100Changed10Remotes-4         	14166939	      1595 ns/op	     448 B/op	      16 allocs/op
BenchmarkUpdate100ChangedRemote10Remotes
BenchmarkUpdate100ChangedRemote10Remotes-4   	15173292	      1593 ns/op	     448 B/op	      15 allocs/op
```

A benchmark for getting need stats before and now feels pretty pointless: Now it's looking up a struct in memory, before iterating/unmarshalling the entire global prefix of the db doing lots of checks.

calmh:
> A benchmark for getting need stats before and now feels pretty pointless

but that's the one that shows the awesomeness 😿

imsodin:
> but that's the one that shows the awesomeness crying_cat_face

Brace for the awesomeness:

master:

```
BenchmarkUpdate100Changed10Remotes
BenchmarkUpdate100Changed10Remotes-4         	 7592064	      1578 ns/op	     448 B/op	      16 allocs/op
BenchmarkUpdate100ChangedRemote10Remotes
BenchmarkUpdate100ChangedRemote10Remotes-4   	 6654831	      2070 ns/op	     448 B/op	      15 allocs/op
BenchmarkNeedCount
BenchmarkNeedCount-4                         	    2136	   5576769 ns/op	  905676 B/op	   14089 allocs/op
```

PR:

```
BenchmarkUpdate100Changed10Remotes
BenchmarkUpdate100Changed10Remotes-4         	 7089142	      1775 ns/op	     448 B/op	      16 allocs/op
BenchmarkUpdate100ChangedRemote10Remotes
BenchmarkUpdate100ChangedRemote10Remotes-4   	 6603822	      1782 ns/op	     448 B/op	      15 allocs/op
BenchmarkNeedCount
BenchmarkNeedCount-4                         	255457330	        47.7 ns/op	       0 B/op	       0 allocs/op
```

lib/db/schemaupdater.go
```diff
@@ -154,9 +156,13 @@ func (db *schemaUpdater) updateSchema0to1(_ int) error {
 			if err != nil {
 				return err
 			}
+			// Purposely pass nil file name to remove from global list,
+			// but don't touch meta and needs
 			buf, err = t.removeFromGlobal(gk, buf, folder, device, nil, nil)
 			if err != nil {
-				return err
+				if err != errEntryFromGlobalMissing {
```
calmh, line 163: &&

imsodin:
Most of the global update stuff is optimisations. I thought about in a first step doing a version that just stupidly checks everything and doesn't take any shortcuts - I can look into that again.

AudriusButkevicius:
I guess this just needs a lonely evening and a large crate of wine?

imsodin:
I could make smaller, functional but on their own useless PRs against a feature branch to make it shorter evenings with smaller creates (or equally sized ones and funnier evenings). Essentially the first few commits here, plus hopefully i can split the last of the first ones up further.

lib/model/model.go
```diff
@@ -787,52 +787,35 @@ func (m *model) Completion(device protocol.DeviceID, folder string) FolderComple
...
 
-	l.Debugf("%v Completion(%s, %q): %f (%d / %d = %f)", m, device, folder, completionPct, need, tot, needRatio)
+	l.Debugf("%v Completion(%s, %q): %f (%d / %d = %f)", m, device, folder, completionPct, need.Bytes, tot, needRatio)
 
 	return FolderCompletion{
 		CompletionPct: completionPct,
-		NeedBytes:     need,
-		NeedItems:     items,
+		NeedBytes:     need.Bytes,
+		NeedItems:     need.Files + need.Directories + need.Files,
```
calmh, line 816: Need twice the files?

lib/db/schemaupdater.go
```diff
@@ -484,3 +498,66 @@ func (db *schemaUpdater) updateSchemato9(prev int) error {
 
 	return t.Commit()
 }
+
+func (db *schemaUpdater) updateSchemato10(_ int) error {
```
calmh, line 502: I don't fully understand why we need a migration as opposed to just doing a metadata recalc, which we could force if there is no entry for the need bit for example?

imsodin: This PR adds a `Deleted` flag to the global version, for which a db transition is required. However recalculating needs is indeed not necessary anymore now that we recalc on upgrade. I removed that, which means I redid the transition to be more efficient (unmarshal global lists only once). That also triggered a deficiency in my test data: Didn't set the `prevVersion` to 9 and didn't set any `Sequence` on the file infos -> adjusted generation func and regenerated test db.

lib/db/schemaupdater.go
```diff
@@ -22,9 +22,10 @@ import (
 //   6: v0.14.50
 //   7: v0.14.53
 //   8-9: v1.4.0
+//   10: v1.5.0
 const (
-	dbVersion             = 9
-	dbMinSyncthingVersion = "v1.4.0"
+	dbVersion             = 10
+	dbMinSyncthingVersion = "v1.5.0"
```
calmh, line 28:
```suggestion
	dbMinSyncthingVersion = "v1.6.0"
```

lib/db/schemaupdater.go
```diff
@@ -22,9 +22,10 @@ import (
 //   6: v0.14.50
 //   7: v0.14.53
 //   8-9: v1.4.0
+//   10: v1.5.0
```
calmh, line 25:
```suggestion
//   10: v1.6.0
```

calmh:
Rerunning with `STRECHECKDBEVERY=1s` to force metadata recalc brought it back in line...

imsodin:
The migration does not trigger a db check/meta recalc itself, but requires at least a meta recalc. Which on upgrade will happen, as schema upgrades come before the version comparison, leading to the check and recalc part. However that should only cause incorrect metadata, not out of sync - looking into that.

imsodin:
That's it, could reproduce before (d1b13b7), is working correctly now:  
Running a dev build, the check/recalc on upgrade did not happen. This check recalculates the amount of needed files, which should be zero, but isn't due to the migration. When you click on the out of sync items list, it is empty, because the actually stored list of files that are out of sync is correct. Then recalculating metadata the two get in sync again.  
Now the migration triggers a meta recalc regardless of upgrade or not, by deleting the metadata in db.

AudriusButkevicius:
Is v1.6.0 rc going out tomorrow, given sudden influx of reviews?

calmh:
That’s generally the plan yes

AudriusButkevicius:
Ok, I'll try to wrap up the reorder stuff shortly, given its close.

PR #6419, opened 2020-03-18
keypair generate: check if keypair doestn't exist for genrate one

keevBush:
We check if error is File not found before generate a keypair

### DIFF
--- a/cmd/stdiscosrv/main.go
+++ b/cmd/stdiscosrv/main.go
@@ -105,6 +105,8 @@ func main() {
 			if err != nil {
 				log.Fatalln("Failed to generate X509 key pair:", err)
 			}
+		} else if err != nil {
+			log.Fatalln("Failed to load keypair:", err)
 		}
 	}
 
--- a/cmd/stdiscosrv/main.go
+++ b/cmd/stdiscosrv/main.go
@@ -102,20 +102,7 @@ func main() {
 		log.Println("Failed to load keypair. Generating one, this might take a while...")
 		cert, err = tlsutil.NewCertificate(certFile, keyFile, "stdiscosrv", 20*365)
 		if err != nil {
-			if strings.Contains(err.Error(), os.ErrPermission.Error()) {
-				log.Fatalln("Failed to generate : permission not allowed")
-			} else if strings.Contains(err.Error(), "no such file or directory") {
-				log.Fatalln("Failed to generate : Path doesn't exist")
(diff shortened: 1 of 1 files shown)


cmd/stdiscosrv/main.go
```diff
@@ -99,10 +99,12 @@ func main() {
 
 	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
 	if err != nil {
-		log.Println("Failed to load keypair. Generating one, this might take a while...")
-		cert, err = tlsutil.NewCertificate(certFile, keyFile, "stdiscosrv", 20*365)
-		if err != nil {
-			log.Fatalln("Failed to generate X509 key pair:", err)
+		if os.IsNotExist(err) {
```
calmh: We still need to handle the case when there is an error but it's not "doesn't exist". I suggest 

```
if os.IsNotExist(err) {
  ... what you did
} else if err != nil {
  log.Fatalln("Failed to load keypair:", err
}
```

cmd/stdiscosrv/main.go
```diff
@@ -99,10 +99,14 @@ func main() {
...
+			log.Println("Failed to load keypair. Generating one, this might take a while...")
+			cert, err = tlsutil.NewCertificate(certFile, keyFile, "stdiscosrv", 20*365)
+			if err != nil {
+				log.Fatalln("Failed to generate X509 key pair:", err)
+			}
+		} else if err != nil {
```
calmh: 👍 now we don't need the outer `if err != nil` either

keevBush: That error is for if we have another error with regenrate a keypair

calmh: Right. But `os.IsNotExist(nil)` is legal, and becomes false. So consider whether we need the extra `if err != nil` before it.

calmh:
I think my meaning got lost. Please consider each branch and what will happen if the operation fails (`err != nil`). Currently an error to generate the certificate will go unreported and we'll crash when `cert` is nil. Something like this should be clean and simple:

```
if os.IsNotExist(err) {
  // there is no certificate, we should create one, set same err variable
  cert, err = tlsutil.NewCertificate(certFile, keyFile, "stdiscosrv", 20*365)
}
if err != nil {
  // The error from LoadX509KeyPair was not "does not exist", or we got an error from NewCertificate
  log.Fatal(...)
}
```

calmh:
I fixed this up to illustrate the minimal fix I was aiming for

Issue #6427, opened 2020-03-19
Panic due to non-existing entry in a global list


imsodin:
`readWriteTransaction.removeFromGlobal` has one path where an unexpected error is handled badly. As in the global list will then be inconsistent. `checkGlobal` then cleans that up later. This surfaced because of the `withNeed` refactor, which does not ignore invalid global list entries anymore.

I intentionally do not link to sentry, as the error message added in the refactor contains filenames (will fix that too of course).

PR #6429, opened 2020-03-19
lib/db, lib/syncthing: Repair db once on upgrade (ref #6425, #6427)

imsodin:
Both #6425 and #6427 require a metadata (respectively sequence) recalc. This adds an exported method to `db.Lowlevel` to trigger that, which is then done on startup if an upgrade with a previous version <1.4.2 is detected.

I moved the code for the recalculation from set.go to lowlevel.go and made them methods of `*Lowlevel` (which before got passed as arg). This is entirely copy&paste, except that the previous `recalc` fixture now is now a method called `getMetaAndCheck`.

I manually tested running a binary from this PR on a db from 1.4.0.


lib/syncthing/syncthing.go
```diff
@@ -224,9 +225,15 @@ func (a *App) startup() error {
 
 	prevParts := strings.Split(prevVersion, "-")
 	curParts := strings.Split(build.Version, "-")
-	if prevParts[0] != curParts[0] {
+	if rel := upgrade.CompareVersions(prevParts[0], curParts[0]); rel != upgrade.Equal {
 		if prevVersion != "" {
 			l.Infoln("Detected upgrade from", prevVersion, "to", build.Version)
+			// Check and repair metadata and sequences once immediately
```
calmh: Why not just do checkrepair on every upgrade for now?

imsodin: Doing it on all upgrades including RCs feels like a lot of churn (e.g. considering the 11 RCs last month). And we need it to happen now on v1.4.1-rc.2. So maybe do it on stable updates and v1.4.1-rc.2 explicitly could be a middle way. Or am overthinking and for RC users it's fine to do it on every upgrade?

calmh: Hmm. I didn't consider that we want it to happen between-rc:s this time. However, I wonder if the correct thing here isn't to actually run this on every upgrade, even rc-to-rc. The criteria then would be to split on `+` and just string-equals compare the thing before the plus (if any).

calmh: I think the full index transfer is a lot heavier than the checks we do, so if anything that would be the part I think we can remove nowadays...

PR #6433, opened 2020-03-19
lib/model: Don't mark deleted local only as locally changed (fixes #6432)

### LINKED ISSUE #6432: Deleted file that existed locally only reported as locally changed
A file that only ever existed locally, but is now deleted, keeps showing up as a locally changed item. That has no value, as for all practical intents and purposes it does not exist. See https://forum.syncthing.net/t/how-to-accept-local-changes/14657/47?u=imsodin and posts leading up to it (thanks @xarx00).

imsodin:
> A file that only ever existed locally, but is now deleted, keeps showing up as a locally changed item. That has no value, as for all practical intents and purposes it does not exist.

Like with similar special conditions on send only folder, this adds a stage to pulling that checks for this situation and fixes it by removing the local flag.

### DIFF
--- a/lib/db/structs.go
+++ b/lib/db/structs.go
@@ -129,19 +129,27 @@ func (f FileInfoTruncated) FileModifiedBy() protocol.ShortID {
 }
 
 func (f FileInfoTruncated) ConvertToIgnoredFileInfo(by protocol.ShortID) protocol.FileInfo {
-	file := f.CopyToFileInfo()
+	file := f.copyToFileInfo()
 	file.SetIgnored(by)
 	return file
 }
 
 func (f FileInfoTruncated) ConvertToDeletedFileInfo(by protocol.ShortID) protocol.FileInfo {
-	file := f.CopyToFileInfo()
+	file := f.copyToFileInfo()
 	file.SetDeleted(by)
 	return file
 }
 
+// ConvertDeletedToFileInfo converts a deleted truncated file info to a regular file info
+func (f FileInfoTruncated) ConvertDeletedToFileInfo() protocol.FileInfo {
+	if !f.Deleted {
+		panic("ConvertDeletedToFileInfo must only be called on deleted items")
+	}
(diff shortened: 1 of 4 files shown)


lib/db/structs.go
```diff
@@ -129,19 +129,19 @@ func (f FileInfoTruncated) FileModifiedBy() protocol.ShortID {
...
 func (f FileInfoTruncated) ConvertToDeletedFileInfo(by protocol.ShortID) protocol.FileInfo {
-	file := f.copyToFileInfo()
+	file := f.CopyToFileInfo()
 	file.SetDeleted(by)
 	return file
 }
 
 // copyToFileInfo just copies all members of FileInfoTruncated to protocol.FileInfo
-func (f FileInfoTruncated) copyToFileInfo() protocol.FileInfo {
+func (f FileInfoTruncated) CopyToFileInfo() protocol.FileInfo {
```
calmh: Somewhat skeptical of exposing this method given the latest change to putFile and the fact that the resulting FileInfo can't be easily or safely passed back into the database layer. The Deleted/Ignored cases above are the exceptions where this is safe (as those shouldn't have block lists).

imsodin: I added a new method that converts deleted, truncated file infos to protocol file infos and panics if the file isn't actually deleted.

imsodin:
lgtm?

calmh:
A test for this situation would be cool...

imsodin:
I wrote the test and it passes on master - in hindsight obviously: Deleted files that only we had are handled by scanning (already tested),  and if the remote deletes a file we already deleted we pull that deletion thus removing the locally changed item. I'll file the test separately.

calmh:
So test revealed there was no bug, or bug is something other than we though?

imsodin:
Initially I thought it was "no bug", but right now I am writing another test as I have suspicion for "https://github.com/syncthing/syncthing/pull/6433#issuecomment-604987278" :)

PR #6434, opened 2020-03-20
lib/db: Don't whack blocks when putting truncated file

calmh:
As of the latest database checker we are again putting files without
blocks. I'm not 100% convinced that's a great idea, but we also do it
for ignored files apparently so it looks like we probably should support
it.

### DIFF
--- a/lib/db/db_test.go
+++ b/lib/db/db_test.go
@@ -261,11 +261,11 @@ func TestRepairSequence(t *testing.T) {
 	short := protocol.LocalDeviceID.Short()
 
 	files := []protocol.FileInfo{
-		{Name: "fine"},
-		{Name: "duplicate"},
-		{Name: "missing"},
-		{Name: "overwriting"},
-		{Name: "inconsistent"},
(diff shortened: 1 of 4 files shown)


calmh:
Actually no. Accepting random fileinfos with a hash is a problem and we shouldn't do it.

calmh:
On the other hand, non-deleted/invalid without blocks are already dropped at the protocol level.

imsodin:
Oh dammit, I am still not on top of the blockhashes...

The changes look good. If you'd rather not allow it, an alternative would be to use `procotol.checkFileInfoConsistency` in `putFile` and panic on error - then my mistakes will at least properly blow up early.

calmh:
I think this is a fine compromise

calmh:
A proper repair for this isn't super trivial and I'm hoping it won't be widespread enough yet that we need one :/

imsodin:
Latest number of rc.2's is 1400, but it only concerns people with messed up sequence entries, so I hope were good...

PR #6441, opened 2020-03-22
cmd/syncthing: Properly handle STNORESTART=1 (fixes #6440)

### LINKED ISSUE #6440: Doesn't run monitor process when started with STNORESTART=1
In #6278 we intended to always run the monitor process, and this happens with the `--no-restart` flag, but not when the environment variable is set.

calmh:
Makes it truly equivalent to -no-restart, and also updates the option
descriptions to be more truthful.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -108,10 +108,7 @@ are mostly useful for developers. Use with care.
  STLOCKTHRESHOLD   Used for debugging internal deadlocks; sets debug
                    sensitivity.  Use only under direction of a developer.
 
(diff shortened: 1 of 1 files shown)


Catfriend1:
Does that mean STNORESTART no longer exists and if required STMONITORED has to be used instead?

calmh:
STNORESTART exists and retains its meaning, it’s just that that meaning wasn’t “don’t launch another process” but “don’t manage restarts internally”.

STMONITORED is how Syncthing knows whether it’s living inside the monitor process or not. It’s not a documented feature, but it’s a hack if you really need to prevent an extra process from being created. Almost nobody should do that. Android might be a valid exception.

PR #6444, opened 2020-03-22
Auth on initial startup

### LINKED ISSUE #4703: Suggest setting up auth on initial startup

Yithis:
### Purpose

Add warning about setting up auth on initial startup (fixes #4703). My solution is inspired by one proposed in #5217.

### Testing

Building and checking if everything is working properly.

### Screenshots

image


gui/default/syncthing/core/syncthingController.js
```diff
@@ -51,13 +51,21 @@ angular.module('syncthing.core')
...
         $scope.folderPathErrors = {};
         $scope.currentFolder = {};
         resetRemoteNeed();
 
         try {
             $scope.metricRates = (window.localStorage["metricRates"] == "true");
         } catch (exception) { }
+        try {
+            if(typeof window.localStorage["authWarningDismissed"] != 'undefined'){
+                $scope.authWarningDismissed = (window.localStorage["authWarningDismissed"] == "true");
```
AudriusButkevicius: no need for `()`

gui/default/syncthing/core/syncthingController.js
```diff
@@ -51,13 +51,21 @@ angular.module('syncthing.core')
         $scope.themes = [];
         $scope.globalChangeEvents = {};
         $scope.metricRates = false;
+        $scope.authWarningDismissed = true;
```
AudriusButkevicius: Why does this start as true? If will always start off as dismissed, because the localstorage key will be undefined.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -449,6 +457,9 @@ angular.module('syncthing.core')
                 && (!guiCfg.user || !guiCfg.password)
                 && guiCfg.authMode !== 'ldap'
                 && !guiCfg.insecureAdminAccess;
+
+            $scope.noAuth = ($scope.authWarningDismissed || checkAuthDate())
```
AudriusButkevicius: Should this not be `!$scope.authWarningDismissed`

gui/default/syncthing/core/syncthingController.js
```diff
@@ -682,6 +693,33 @@ angular.module('syncthing.core')
...
+                }
+
+                var expire = Number(window.localStorage["authWarningDismissedTime"]);
+                var now = new Date().getTime();
+
+                if(now > expire){
+                    return true;
+                } else {
+                    return false;
+                }
```
AudriusButkevicius: `return now > expire;`

gui/default/syncthing/core/syncthingController.js
```diff
@@ -682,6 +693,33 @@ angular.module('syncthing.core')
             });
         };
 
+        $scope.authenticationDontWant = function () {
```
AudriusButkevicius: dismissAuthenticationSetupSuggestion

gui/default/syncthing/core/syncthingController.js
AudriusButkevicius: Also, this value is never set to true, it's only set to false.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -682,6 +693,33 @@ angular.module('syncthing.core')
...
+            try {
+                window.localStorage["authWarningDismissed"] = false;
```
AudriusButkevicius: Why false?

gui/default/syncthing/core/syncthingController.js
```diff
@@ -682,6 +693,33 @@ angular.module('syncthing.core')
...
+
+                var expire = Number(window.localStorage["authWarningDismissedTime"]);
+                var now = new Date().getTime();
+
+                if(now > expire){
+                    return true;
+                } else {
+                    return false;
+                }
+            } catch (exception) { }
```
AudriusButkevicius: Returns undefined in case of exception

gui/default/syncthing/core/syncthingController.js
```diff
@@ -682,6 +693,29 @@ angular.module('syncthing.core')
             });
         };
 
+        $scope.dismissAuthenticationSetupSuggestion = function () {
+            try {
+                window.localStorage["authWarningDismissed"] = true;
+                var expire = 7*24*60*60*1000;  // Miliseconds (7 days * 24 hours * 60 mins * 60 secs * 1000 miliseconds)
```
imsodin: I most definitely don't want to get that banner every week on my multi-user systems with only one user - I wouldn't expire at all.

Yithis: I agree that such reminder can be annoying. So should I rewrite my PR in way that this notification will be using `config.options.unackedNotificationIDs` and will work as others?

imsodin: In my opinion yes.

Yithis:
Since I messed up with updating my branch I will recreate this PR to make it more readable

PR #6448, opened 2020-03-23
all: Pass db intervals as args not env vars

imsodin:
Lets cmd/syncthing do the env var parsing and the db intervals are then passed on to `db.NewLowlevel` instead of having package global vars. I opted for adding `NewLowlevelDefault` and using that in tests instead of using `NewLowlevel(db, 0, 0)` in tests, but can't really say anymore why exactly - I don't care either way. 

The only relevant parts of the diff are:  
https://github.com/syncthing/syncthing/pull/6448/files#diff-90be85495f283f39a39308cf8ffb765c
https://github.com/syncthing/syncthing/pull/6448/files#diff-e4dc421e8157ba4e8ba69c47a592dd10
https://github.com/syncthing/syncthing/pull/6448/files#diff-913aac63c6b88a6efc318e601657c57a

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -71,9 +71,9 @@ func NewLowlevel(backend backend.Backend, opts ...Option) *Lowlevel {
 
 type Option func(*Lowlevel)
 
-// RecheckIntervalOpt sets the time interval in between metadata recalculations
+// WithRecheckInterval sets the time interval in between metadata recalculations
 // and consistency checks.
-func RecheckIntervalOpt(dur time.Duration) Option {
+func WithRecheckInterval(dur time.Duration) Option {
 	return func(db *Lowlevel) {
 		if dur != 0 {
 			db.recheckInterval = dur
@@ -81,8 +81,8 @@ func RecheckIntervalOpt(dur time.Duration) Option {
 	}
 }
 
-// IndirectGCIntervalOpt sets the time interval in between GC runs.
-func IndirectGCIntervalOpt(dur time.Duration) Option {
+// WithIndirectGCInterval sets the time interval in between GC runs.
+func WithIndirectGCInterval(dur time.Duration) Option {
 	return func(db *Lowlevel) {
 		if dur != 0 {
 			db.indirectGCInterval = dur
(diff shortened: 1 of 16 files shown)


calmh:
What about functional options like ignore.New?

lib/db/lowlevel.go
```diff
@@ -41,41 +41,55 @@ const (
...
 	db.keyer = newDefaultKeyer(db.folderIdx, db.deviceIdx)
 	go db.gcRunner()
 	return db
 }
 
+type Option func(*Lowlevel)
+
+// RecheckIntervalOpt sets the time interval in between metadata recalculations
+// and consistency checks.
+func RecheckIntervalOpt(dur time.Duration) Option {
```
calmh: Typically called WithRecheckInterval for better sentence flow when calling

lib/db/lowlevel.go
```diff
@@ -48,27 +41,55 @@ func init() {
...
 	return db
 }
 
+type Option func(*Lowlevel)
+
+// WithRecheckInterval sets the time interval in between metadata recalculations
+// and consistency checks.
+func WithRecheckInterval(dur time.Duration) Option {
+	return func(db *Lowlevel) {
+		if dur != 0 {
```
calmh: So about that, I would do `STRECHECKDBEVERY=0 syncthing` to force a check *now*. This breaks that, I'd have to say `STRECHECKDBEVERY=1s` instead. Which is maybe fine... At least for the GC interval zero isn't a great choice. (We could say that zero means no GC, but it's also intuitive enough to say `STGCINDIRECTEVERY=100000d` or so...)

calmh: There's also negative numbers which we might want to handle or reject

imsodin: I handled negative as meaning default too. The reasoning for 0 meaning default is that I don't want to check the options in `lib/syncthing` to decide whether to use the `With` options - just send the zero value and get the default. And I think "properly" adding disabling is overkill, as I don't see a use-case outside of debugging (and there the workaround with huge intervals is fine).

PR #6449, opened 2020-03-24
lib/model: Unset local flag on deleted files (fixes #6436)

### LINKED ISSUE #6436: Revert Local Changes red button does not work correctly
- what happened,
When a Syncthing folder contains locally-changed-items, the red "Revert Local Changes" button is shown. When this button is pressed, it seemingly does nothing. When it is pressed for the second time, the local changes are reverted.

Screenshot: https://ibb.co/Nx3VkqY

 - what you expected to happen instead
The red button should work on the first press.

In fact, it might be just a GUI refreshing problem. The first press may actually work, but the second press of the button might be needed to refresh the GUI. However, it may be a different problem, because when I reloaded the GUI at the beginning and then pressed the red button (once), it worked correctly.

### Version Information

Syncthing Version: v1.4.0
OS Version: Synology NAS DS416play (Linux x64)
Browser Version: Firefox 68.6.0esr

imsodin:
We previously didn't remove the local flag for files that only exist locally, resulting in those disappearing from local/global state, but still existing as locally changed items (thus override button and number of locally changed items remains in UI).

I also added two info log lines to the revert and override methods. Currently there's no clear indication of these operations happening in the logs. Given those are not regular operations, but ones triggered by the user, I believe that is appropriate verbosity.

### DIFF
--- a/lib/model/folder_sendonly.go
+++ b/lib/model/folder_sendonly.go
@@ -97,7 +97,7 @@ func (f *sendOnlyFolder) pull() bool {
 }
 
 func (f *sendOnlyFolder) Override() {
-	l.Infof("Overriding global stat on folder %v", f.Description)
+	l.Infof("Overriding global state on folder %v", f.Description)
 
 	f.setState(FolderScanning)
 	batch := make([]protocol.FileInfo, 0, maxBatchSizeFiles)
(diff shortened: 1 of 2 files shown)


lib/model/folder_sendonly.go
```diff
@@ -97,6 +97,8 @@ func (f *sendOnlyFolder) pull() bool {
 }
 
 func (f *sendOnlyFolder) Override() {
+	l.Infof("Overriding global stat on folder %v", f.Description)
```
AudriusButkevicius: Stat

AudriusButkevicius:
The flag is confusing, it's more like ModifiedLocally. Anyways, db stuff, leave it to someone else.

PR #6451, opened 2020-03-24
lib/api: Set ServerName on LDAPS connections (fixes #6450)

### LINKED ISSUE #6450: LDAP auth doesn't handle LDAPS with certificate validation
With the following config:

```
<ldap>
    <address>example.com:636</address>
    <bindDN>%s@ad.example.com</bindDN>
    <transport>tls</transport>
</ldap>
```

and an appropriate certificate for example.com on the host in question, Syncthing errors auth attempts with the following:

```
[QZV4N] 12:37:18 WARNING: LDAP Dial: LDAP Result Code 200 "Network Error": tls:
   either ServerName or InsecureSkipVerify must be specified in the tls.Config
```

calmh:
tls.Dial needs it for certificate verification.

### DIFF
--- a/lib/api/api_auth.go
+++ b/lib/api/api_auth.go
@@ -11,6 +11,7 @@ import (
 	"crypto/tls"
 	"encoding/base64"
 	"fmt"
+	"net"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


calmh:
Tested manually with correct and incorrect certificates, with and without insecureSkipVerify, it works as it's supposed to.

PR #6453, opened 2020-03-24
go.mod: Update quic-go to 0.14.4

imsodin:
As we don't support go1.12 anymore, lets upgrade quic-go. Unfortunately not to the most recent version, as they broke 1.13 by using embedded interfaces (https://github.com/lucas-clemente/quic-go/commit/c88a69034d6b6d288a01578f5256e90aca95428a). For a library they are a bit too bleeding-edge-no-matter-what in my opinion...

I think @calmh mentioned something about needing to adapt to new API on upgrade. It builds for me and due to the lack of release notes I didn't look further into it.


AudriusButkevicius:
What is 1.14.4? Latest release they list is 0.14?

calmh:
There was some api change around contexts and timeouts, iirc, but maybe it was in newer versions. If it builds it builds.

imsodin:
Oh no, I must have gotten infected with the 1-0-blindness - too much exposure on the forum.

Issue #6150, opened 2019-11-11
Switch data.s.n and relays.s.n to OpenStreetMap instead of Google Maps


calmh:
For great justice.

https://switch2osm.org/the-basics/
https://leafletjs.com/examples/quick-start/

AudriusButkevicius:
all the bubble heatmap stuff would have to be re-done which I suspect is hard.

PR #6454, opened 2020-03-24
ursrv: remove google map and add leaflet heatmap

keevBush:
I remove google map for leaflet and leafletheatmap.
I added the data locally to test directly on the HTML template file.

### DIFF
--- a/cmd/ursrv/static/index.html
+++ b/cmd/ursrv/static/index.html
@@ -215,7 +215,6 @@ found in the LICENSE file.
     var locations = [];
     {{range $location, $weight := .locations}}
     locations.push({lat:{{- $location.Latitude -}},lng:{{- $location.Longitude -}},count:{{- $weight -}}});
-    // locations.push({location: new google.maps.LatLng({{- $location.Latitude -}}, {{- $location.Longitude -}}), weight: {{- $weight -}}});
     {{- end}}
 
     function drawHeatMap() {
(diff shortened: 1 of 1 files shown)


calmh:
I like it! Can you clean out the old google maps stuff (should be no references to it left)?

Two things about the actual heatmap though. First I think it's better if it starts zoomed out to the whole world, something like this:

Screenshot 2020-03-25 at 09 51 08

Secondly, I think the actual heat levels need some tweaking. We have a few hotspots, but much of the usage is pretty much invisible because there are lots of "weight 1" points that are just a very very faint blue tone... It would be nice if it were a bit more visible even for small usage sites

cmd/ursrv/static/index.html
```diff
@@ -211,47 +216,48 @@
...
+      };
+      
+      var baseLayer = L.tileLayer(
+        'http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',{
```
uok: https ?

uok:
Thanks, this fixes https://github.com/syncthing/syncthing/issues/6150

cmd/ursrv/static/index.html
keevBush: For second proposition can I have a json file for my test ???

keevBush:
Can I have a json file with actual data for simply test

calmh:
Sure let me sort that out

calmh:
https://data.syncthing.net/locations.json

calmh:
You can use that in the page instead of baking it into the template if that's easier / cleaner.

AudriusButkevicius:
Very nice!

PR #6459, opened 2020-03-26
strelaypoolsrv: remove google map and add leaflet map

keevBush:
Remove google map and add Leaflet for relays.syncthing.net

### DIFF
--- a/cmd/strelaypoolsrv/gui/index.html
+++ b/cmd/strelaypoolsrv/gui/index.html
@@ -9,6 +9,7 @@
     <meta name="author" content=""/>
 
     <title>Relay stats</title>
+    <link rel="stylesheet" href="//use.fontawesome.com/releases/v5.0.13/css/all.css"/>
     <link href="//maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.min.css" rel="stylesheet"/>
     <link rel="stylesheet" href="//use.fontawesome.com/releases/v5.0.13/css/all.css"/>
     <link rel="stylesheet" href="//unpkg.com/leaflet@1.6.0/dist/leaflet.css"
@@ -48,7 +49,7 @@
         <p>Please wait while we gather data</p>
       </div>
       <div>
-        <div ng-show="relays === undefined" class="ng-hide">
+        <div ng-show="relays !== undefined" class="ng-hide">
           <p>
             Currently {{ relays.length }} relays online ({{ totals.goMaxProcs }} cores in total).
           </p>
--- a/cmd/strelaypoolsrv/gui/index.html
+++ b/cmd/strelaypoolsrv/gui/index.html
@@ -191,7 +191,7 @@
 
 
     <script type="text/javascript" src="//code.jquery.com/jquery-2.1.4.min.js"></script>
-    <script type="text/javascript" src="http://cdnjs.cloudflare.com/ajax/libs/angular.js/1.5.8/angular.min.js"></script>
+    <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/angular.js/1.5.8/angular.min.js"></script>
     <script type="text/javascript" src="//maxcdn.bootstrapcdn.com/bootstrap/3.3.5/js/bootstrap.min.js"></script>
   </body>
 
@@ -239,7 +239,6 @@
       attribution: 'Leaflet',
       maxZoom: 17
     }).addTo($scope.map);
-    // $scope.mapBounds = new google.maps.LatLngBounds();
     $scope.tooltipTemplate = $('#infoTemplate').html();
     $scope.usedLocations = {};
     $scope.sortType = 'stats.numActiveSessions';
--- a/cmd/strelaypoolsrv/gui/index.html
+++ b/cmd/strelaypoolsrv/gui/index.html
@@ -9,8 +9,14 @@
     <meta name="author" content=""/>
 
     <title>Relay stats</title>
-    <link href="//maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.min.css" rel="stylesheet"/>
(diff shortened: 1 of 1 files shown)


cmd/strelaypoolsrv/gui/index.html
```diff
@@ -9,8 +9,14 @@
     <meta name="author" content=""/>
 
     <title>Relay stats</title>
-    <link href="//maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.min.css" rel="stylesheet"/>
-    <link rel="stylesheet" href="//use.fontawesome.com/releases/v5.0.13/css/all.css"/>
+    <link href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.min.css" rel="stylesheet"/>
+    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.0.13/css/all.css"/>
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.6.0/dist/leaflet.css"
```
AudriusButkevicius: You should probably keep the hrefs without protocol, aka `//....`, then you don't have browser errors when accessing the page via http oppose to https, etc.

cmd/strelaypoolsrv/gui/index.html
```diff
@@ -38,11 +44,11 @@
     <div class="container">
       <h1>Relay Pool Data</h1>
       <div ng-if="relays === undefined" class="text-center">
-        <img src="//cdnjs.cloudflare.com/ajax/libs/galleriffic/2.0.1/css/loader.gif" alt=""/>
+        <img src="https://cdnjs.cloudflare.com/ajax/libs/galleriffic/2.0.1/css/loader.gif" alt=""/>
         <p>Please wait while we gather data</p>
       </div>
       <div>
-        <div ng-show="relays !== undefined" class="ng-hide">
+        <div ng-show="relays === undefined" class="ng-hide">
```
calmh: pretty sure this change is incorrect

cmd/strelaypoolsrv/gui/index.html
```diff
@@ -185,9 +191,8 @@ <h1>Relay Pool Data</h1>
 
 
     <script type="text/javascript" src="//code.jquery.com/jquery-2.1.4.min.js"></script>
-    <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/angular.js/1.5.8/angular.min.js"></script>
+    <script type="text/javascript" src="http://cdnjs.cloudflare.com/ajax/libs/angular.js/1.5.8/angular.min.js"></script>
```
calmh: no `http`, either `//...` as it was to keep whatever the browser is using, or `https` if we need to hardcode something for some reason

cmd/strelaypoolsrv/gui/index.html
```diff
@@ -228,11 +233,13 @@ <h1>Relay Pool Data</h1>
...
-      mapTypeId: google.maps.MapTypeId.ROADMAP
-    });
-    $scope.mapBounds = new google.maps.LatLngBounds();
+    $scope.map = L.map('map').setView([40.90296, 1.90925], 2);
+    L.tileLayer('http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
+    {
+      attribution: 'Leaflet',
+      maxZoom: 17
+    }).addTo($scope.map);
+    // $scope.mapBounds = new google.maps.LatLngBounds();
```
calmh: remove

cmd/strelaypoolsrv/gui/index.html
keevBush: Yes it me I change that when I load the map

cmd/strelaypoolsrv/gui/index.html
```diff
@@ -38,7 +45,7 @@
     <div class="container">
       <h1>Relay Pool Data</h1>
       <div ng-if="relays === undefined" class="text-center">
-        <img src="//cdnjs.cloudflare.com/ajax/libs/galleriffic/2.0.1/css/loader.gif" alt=""/>
+        <img src="https://cdnjs.cloudflare.com/ajax/libs/galleriffic/2.0.1/css/loader.gif" alt=""/>
```
AudriusButkevicius: needs https removed

calmh: Does it though? I think the root cause here is that it's difficult to test these pages without a fully fledged strelaypoolserv/ursrv setup running, so they're opened as files and then the `//...` url's don't work. I see no harm in hardcoding https for all of these

keevBush: Yes I have the same problem. For resolve there I had https for all of there

calmh:
In principle lgtm, but there is some weirdness with the circles. They are not centered on the markers:

Screen Shot 2020-03-28 at 11 47 52
Screen Shot 2020-03-28 at 11 48 07

AudriusButkevicius:
While at it, we might want to change the circles to log scale.

AudriusButkevicius:
Also, the ng-mouseover stuff that puts the relay to focus when you hover over it in the list below probably does not work anymore.

AudriusButkevicius:
There is also stuff like
```
      $scope.map.fitBounds($scope.mapBounds);
      if ($scope.relays.length == 1) {
        $scope.map.setZoom(13);
      }
```

which I am not sure if it's supported in the new map thing?

AudriusButkevicius:
I don't even see $scope.mapBounds existing anymore.

keevBush:
I look that problem is leaflet. More youb zoom more precision on the map is revelated. Is Leaflet

keevBush:
Pin update his precision when You zoom

AudriusButkevicius:
There is still the missing functionality of the relay being "moused over and focused" when you hover over the row for the relay in the data table below.

AudriusButkevicius:
It seems the circles are in the wrong place after the adjustments that prevent stuff from overlapping:
image

PR #6460, opened 2020-03-26
lib/model: Do Revert/Override synchronously

imsodin:
Writing a test for https://github.com/syncthing/syncthing/pull/6433 I realised `Override` and `Revert` happen asynchronously, i.e. you can't depend on any order of folder operations in conjunction with those. In addition they both set folder state and update the db, so as far as I see they mustn't run asynchronously.

I added a general purpose `syncChan` - I could also add two special purpose chans for overriding and reverting instead if this is "too magic".

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -57,6 +57,7 @@ type folder struct {
 	scanErrorsMut       sync.Mutex
 
 	pullScheduled chan struct{}
+	syncChan      chan chan struct{}
 
 	watchCancel      context.CancelFunc
 	watchChan        chan []string
@@ -96,6 +97,7 @@ func newFolder(model *model, fset *db.FileSet, ignores *ignore.Matcher, cfg conf
 		scanErrorsMut:       sync.NewMutex(),
 
 		pullScheduled: make(chan struct{}, 1), // This needs to be 1-buffered so that we queue a pull if we're busy when it comes.
+		syncChan:      make(chan chan struct{}),
 
 		watchCancel:      func() {},
 		restartWatchChan: make(chan struct{}, 1),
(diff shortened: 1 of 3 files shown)


calmh:
Would it make sense to have these operations be called from the folder loop, and use the same mechanism as we do for requesting a scan for example?

imsodin:
As these are methods on the folder "subtypes", we can't just call `f.override` from `folder`, but we'd have to register a interface for both `Revert` and `Override` like `puller`, which feels like a worse solution than this.

calmh:
The methods could be part of a (more general) `puller` interface. Having a single interface for the three low level implementation methods seems fine, and in the loop we could handle it actor-style with a `chan func()` that the exported methods use to run things synchronously on the loop...

imsodin:
With actor-style an interface isn't even necessary - does this look more like what you had in mind?

lib/model/folder.go
```diff
@@ -172,9 +174,9 @@ func (f *folder) serve(ctx context.Context) {
 			l.Debugln(f, "Scanning due to timer")
 			f.scanTimerFired()
 
-		case req := <-f.scanNow:
-			l.Debugln(f, "Scanning due to request")
-			req.err <- f.scanSubdirs(req.subdirs)
+		case req := <-f.doInSyncChan:
+			l.Debugln(f, "Running something due to request")
```
calmh: :D

lib/model/folder.go
```diff
@@ -224,13 +226,19 @@ func (f *folder) Jobs(_, _ int) ([]string, []string, int) {
...
-		err:     make(chan error),
+	return f.doInSync(func() error { return f.scanSubdirs(subdirs) })
+}
+
+// doInSync allows to run functions synchronously in folder.serve from exported,
+// asynchronously called methods.
+func (f *folder) doInSync(fn func() error) error {
+	req := syncRequest{
+		fn:  fn,
+		err: make(chan error),
```
calmh:
```suggestion
		err: make(chan error, 1),
```

lest we deadlock the loop if we take the ctx.Done() branch below

calmh:
I guess it's a matter of taste if it's cleaner to have `folder` reach into each implementation via an interface, or have the implementation reach into the sync loop guts of `folder`... This way is fine to me as well

Issue #6410, opened 2020-03-15
Wrong 30-days-interval in staggered versioning


mv1005:
### Problem
When using staggered versioning, syncthing deletes versions withing the 30-days limit that are NOT closer than 1 day.

### Steps to reproduce
- Setup a test folder with staggered versioning, leave default settings
- Create dummy version files in the .stversions-folder named `testfile~YYYYMMDD-000000` one for each day reaching back 30 days in the past starting from now
- Restart syncthing to trigger version cleanup
- Syncthing is cleaning versions which are older than 6 days, leaving one file per week

### Expected behaviour
Syncthing leaves all versions untouched, since documentation says:
> For the first 30 days, the most recent version is kept every day.

### Version Information
Syncthing Version: v1.3.4
OS Version: Ubuntu 19.10

I have investigated the situation by looking at the code in `lib/versioner/staggered.go`. There is one digit missing in the number of seconds specifying the 30-days-interval.

```
interval: [4]interval{
    {30, 3600},       // first hour -> 30 sec between versions
    {3600, 86400},    // next day -> 1 h between versions
    {86400, 592000},  // next 30 days -> 1 day between versions  // This line is causing the issue
    {604800, maxAge}, // next year -> 1 week between versions
},
```

The number saying `592000` actually is a time period of 6.8 days only instead of the intended 30 days. The correct number would be `2592000` (Note the leading 2 which is missing).

Furthermore, the tests implemented in `lib/versioner/staggered_test.go` do not cover the issue above.

I have fixed the interval number and updated the tests and could provide a patch for that.

calmh:
Please do!

mv1005:
[0001-lib-versioner-Fixed-staggered-intervals.patch.tar.gz](https://github.com/syncthing/syncthing/files/4335289/0001-lib-versioner-Fixed-staggered-intervals.patch.tar.gz)

mv1005:
Sorry, closed by accident.

uok:
@mv1005 nice catch!
you can use the built-in code editor from Github to create a pull request and submit your patch

mv1005:
What is the best way to contribute code without using the built-in editor? I could fork the projetc and file a pull request, but is there some more lightweight way?

AudriusButkevicius:
Sadly not. I think that's the only two ways to create PRs.

uok:
I normally use [Github Desktop](https://desktop.github.com) but since this patch is just a few lines I'd use the web editor.
[...]

PR #6462, opened 2020-03-27
lib/versioner: Extended tests of intervals

mv1005:
This extends the tests for intervals of the staggered versioner to cover
all intervals.

This relates to github issue #6410.

### Purpose

This is an addition to #6410. If the tests cover all intervals, wrong intervals would be detected regardless of their notation.

### Testing

Run tests in `lib/versioner/staggered_test.go`, should be green. Modify any interval, tests should be red.

### DIFF
--- a/lib/versioner/staggered_test.go
+++ b/lib/versioner/staggered_test.go
@@ -29,31 +29,70 @@ func TestStaggeredVersioningVersionCount(t *testing.T) {
 	versionsWithMtime := []string{
 		// 14:00:00 is "now"
 		"test~20160415-140000", // 0 seconds ago
+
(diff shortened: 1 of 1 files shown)


mv1005:
Well, this depends on the actual intervals to be fixed. This PR won't build until #6461 is merged.

calmh:
Of course. Failure now is the point. 👌 

(builds not happening automatically is because the commit is from february and the build server considers the branch "stale")

mv1005:
So I should just rebase this PR once #6461 has been merged? I am still new to the github workflow...

calmh:
No need. I can trigger the builds, they will happen with a merge to master, and we can merge this afterwards.

PR #6463, opened 2020-03-27
lib/model: Add test for recv-only changes undone by remote

imsodin:
This is the test I wrote for a PR (#6433) that turned out to fix something that doesn't need fixing - as one can see the scenario passes on master too. The scenario is, that I change files on a RO folder and then later the remote changes the files the same way. Normal pulling already makes sure that we end up in a situation without local changes.

In addition the receive only tests work on a "ro" folder, however the used config also has the standard folder, which makes reading debug logs a pain -> removed the default folder.

I'll file another scenario that does requires a fix/improvement separately.


imsodin:
@calmh: The build failure on windows is consistent and TC correctly reports which test is failing, but says build log is empty (it isn't, I can find it in the downloaded full build log). I found https://blog.jetbrains.com/teamcity/2019/08/building-go-programs-in-teamcity/ and indeed we don't do any json stuff in the build config. However before I start fiddling with that (on a copy), I'd like to check if this is intentional.

calmh:
We do use the JSON output as that is what the teamcity test runner consumes... I guess it means the "build log is empty" for that specific test, if that is what you're looking at?

imsodin:
Right, I missed the `GOFLAGS` env var. However something is clearly not working that was before: When I expand the failed test or the entire build log on https://build.syncthing.net/buildConfiguration/Syncthing_BuildWindows/59501?, I do not find any information on the failed test (`TestRecvOnlyRemoteUndoChanges`), but when I download the full build log the test failure/output is there.

imsodin:
Yeah, but the test is failing on windows, which I didn't yet have time to investigate. Plus the buildserver hides/misses/... the output of the failed test and I didn't figure out why it might be doing it.

imsodin:
Can someone please let me know whether they also experience the problem on TC that output from failed tests isn't displayed (not just in this build job, every failed job recently - and the info is there in the downloadable raw log). Just to be sure that I am not looking at a local problem, once I have time to look into it further.

calmh:
Yeah it's the same for me, TC is being very unhelpful here.

lib/model/folder_recvonly_test.go
```diff
@@ -326,6 +326,86 @@ func TestRecvOnlyDeletedRemoteDrop(t *testing.T) {
...
+	}
+	size = receiveOnlyChangedSize(t, m, "ro")
+	if size.Files+size.Directories > 0 {
+		t.Fatalf("ROChanged: expected nothing: %+v", size)
+	}
+
+	// Create a file and modify another
+
+	const file = "foo"
+	const knownFile = "knownDir/knownFile"
```
AudriusButkevicius, line 373: It does not like this on Windows, hence fails. 
If I change it to:
```
knownFile = "knownDir\\knownFile"
```
it passes.

AudriusButkevicius, line 373:
```suggestion
	knownFile := path.Join("knownDir", "knownFile")
```

calmh:
So the fact that this fails on Windows is indicative of a bug?

```
=== RUN   TestRecvOnlyRemoteUndoChanges
11:33:49 INFO: No stored folder metadata for "ro"; recalculating
11:33:49 INFO: Ready to synchronize "ro" (ro) (receiveonly)
11:33:49 INFO: Completed initial scan of receiveonly folder "ro" (ro)
TestRecvOnlyRemoteUndoChanges: folder_recvonly_test.go:406: Receive only: expected all zero: {Files:1 Directories:0 Symlinks:0 Deleted:0 Bytes:4 Sequence:0 DeviceID:[255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255] LocalFlags:8}
--- FAIL: TestRecvOnlyRemoteUndoChanges (0.01s)
```

AudriusButkevicius:
Perhaps my commit needs undoing because it supposedly fixed windows? 🗡️

imsodin:
You're call in https://github.com/syncthing/syncthing/pull/6463#discussion_r417630125 was good, but the fix not: `path` is slash only, `path/filepath` is the cross-compatible library.

calmh (approved):

PR #6465, opened 2020-03-27
build: Add reviewdog golangci-lint runner

calmh:
### Purpose

The golangci-lint service is going away, this is a replacement.


calmh:
So this *wasn't* a replacement when run as a github action, because when it's run from a PR it doesn't get the secret github token for our repo (sensibly) and can't report status checks here. We can probably run it on the build server though.

PR #6469, opened 2020-03-29
go.mod: Update jackpal/gateway dependency (fixes #5288)

### LINKED ISSUE #5288: Errors running "ip" and "route" as snap package
On my Ubuntu machine where syncthing is running (from the snap) I was running `snappy-debug.security scanlog` (a tool snap packagers use to identify confinement problems) for another program, and noticed it complain about syncthing. Full log below.

It looks like the syncthing application is trying to call out to the /bin/ip and /sbin/route commands. These are not bundled in the snap, and confinement prevents them from running. You may need to bundle them via a `stage-packages` stanza in the snapcraft.yaml. `ip` appears to originate in the `iproute2` package and `route` is in the net-tools package. Hope that helps.

Syncthing Version:  v0.14.51, Linux (64 bit)
OS Version: Ubuntu 18.04

```
= AppArmor =
Time: Oct 24 19:50:59
Log: apparmor="DENIED" operation="exec" profile="snap.syncthing.syncthing" name="/sbin/route" pid=18198 comm="syncthing" requested_mask="x" denied_mask="x" fsuid=1000 ouid=0
File: /sbin/route (exec)
Suggestions:
* adjust snap to ship 'route'
* adjust program to use relative paths if the snap already ships 'route'

= AppArmor =
Time: Oct 24 19:50:59
Log: apparmor="DENIED" operation="exec" profile="snap.syncthing.syncthing" name="/bin/ip" pid=18199 comm="syncthing" requested_mask="x" denied_mask="x" fsuid=1000 ouid=0
File: /bin/ip (exec)
Suggestions:
* adjust snap to ship 'ip'
* adjust program to use relative paths if the snap already ships 'ip'
```

albertodonato:
### Purpose

Thsi updates the `github.com/jackpal/gateway` to a version that doesn't use `/sbin/route` and `/sbin/ip`, indirectly fixing #5288.

### Testing

Tests still pass, I also tested a local snap build, denials for those binaries are gone.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/gobwas/glob v0.2.3
 	github.com/gogo/protobuf v1.3.1
 	github.com/golang/groupcache v0.0.0-20190702054246-869f871628b6
-	github.com/jackpal/gateway v1.0.5
+	github.com/jackpal/gateway v1.0.6
 	github.com/jackpal/go-nat-pmp v1.0.2
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
 	github.com/kr/pretty v0.2.0 // indirect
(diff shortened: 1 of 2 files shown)


imsodin:
Thanks.

I've filed https://github.com/jackpal/gateway/issues/22 - lets see whether we can get a release soon to use that instead of the commit. Otherwise lgtm.

imsodin:
Upstream just created a new dot release, please update the go.mod accordingly - thanks.

albertodonato:
Updated dependency to 1.0.6.

PR #6473, opened 2020-03-29
lib/protocol: Remove unused channel Connection.preventSends

greatroar:
### Purpose

Removes a channel from protocol.Connection to which no messages were ever sent.


AudriusButkevicius:
ty

PR #6475, opened 2020-03-29
lib/protocol: Faster Luhn algorithm and better testing

greatroar:
### Purpose

The previous implementation of the Luhn algorithm was very generic; its tests didn't cover the actual alphabet for device IDs. This new specialized version is shorter and has tests that cover the actual base-32 Luhn used. It's also almost 4× faster:

```
name         old time/op    new time/op     delta
Luhnify-8      1.00µs ± 1%     0.28µs ± 4%   -72.38%  (p=0.000 n=9+10)
Unluhnify-8     992ns ± 2%      274ns ± 1%   -72.39%  (p=0.000 n=10+9)
```

The unused luhnValidate function is gone.

### DIFF
--- a/lib/protocol/deviceid.go
+++ b/lib/protocol/deviceid.go
@@ -165,7 +165,7 @@ func luhnify(s string) (string, error) {
 	for i := 0; i < 4; i++ {
 		p := s[i*13 : (i+1)*13]
 		copy(res[i*(13+1):], p)
-		l, err := luhnBase32.generate(p)
+		l, err := luhn32(p)
 		if err != nil {
 			return "", err
 		}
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
So I think our luhn is not actually luhn, and it has an implementation bug, so we probably need to retain that.

greatroar:
I spotted the comment. This PR retains the current behavior, broken or not, and the test asserts that (try running it on master).

calmh:
This may well be better, but there is zero performance requirements on this code and the existing one has worked untouched for years. As such this feels more like churn than an improvement.

calmh:
In general though I'd prefer fixes for things that need fixing rather than rewrite of things are stable.

PR #6477, opened 2020-03-30
lib/util: Fix potential data race

greatroar:
This fixes a potential data race in lib/util.service. That type's Serve method sets a field, stopped, while holding a lock. Its Stop method, however, gets the field's value without synchronizing. The solution is to get the variable while holding the lock but only use it after unlocking.


AudriusButkevicius:
Thank you!

PR #6478, opened 2020-03-30
gui: Number of days must be number flashes red then disappears

### LINKED ISSUE #3808: gui: Number of days must be number flashes red then disappears
# What happened

When adding a new directory (6th one, all others were synced already), using trashcan versioning, I entered "150" without the quotes for the number of days to keep the files in trash. When I pressed the continue or create button, a number of things flashed up as red but the dialog closed so quickly I could just hardly read "Number of days must be a number ...".

Then on the other computer, with exactly the same browser version and linux version, I agreed to add the new shared folder, and filled in the same entries, including "150" for the number of days, and got the same unexpected behaviour.

# What I expected to happen

Either 150 is a number, and the dialog closes and all is well, or my input is invalid and the dialog does not close, leaving all the red areas and text to show me what to change.

### Version Information

Syncthing Version: v0.14.14 Linux (64 bit) "Dysprosium Dragonfly"
OS Version: Linux 3.16.0-4-amd64 #1 SMP Debian 3.16.36-1+deb8u2 (2016-10-19) x86_64 GNU/Linux
Browser Version: Firefox 45.4.0

Shidory:
Fix the message that appears in red and quickly disappears(fixes #3808)

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1847,7 +1847,9 @@ angular.module('syncthing.core')
                     }
                 };
                 delete folderCfg.trashcanFileVersioning;
-                
+                $('#editFolder').on('hidden.bs.modal', function (e) {
+                    delete folderCfg.trashcanClean;
+                })
             } else if (folderCfg.fileVersioningSelector === "simple") {
                 folderCfg.versioning = {
                     'Type': 'simple',
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1847,7 +1847,7 @@ angular.module('syncthing.core')
                     }
                 };
                 delete folderCfg.trashcanFileVersioning;
-                delete folderCfg.trashcanClean;
+                
             } else if (folderCfg.fileVersioningSelector === "simple") {
                 folderCfg.versioning = {
                     'Type': 'simple',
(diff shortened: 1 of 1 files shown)


gui/default/syncthing/core/syncthingController.js
```diff
@@ -1847,7 +1847,7 @@ angular.module('syncthing.core')
                     }
                 };
                 delete folderCfg.trashcanFileVersioning;
-                delete folderCfg.trashcanClean;
```
calmh: Deleting this attribute is proper, because it's a temporary one just used by the config editor and shouldn't be in the config that we post to the backend.

We could make the folder editor use a different temporary object, or maybe delete this after the modal has closed but before posting, or maybe use [setPristine](https://docs.angularjs.org/api/ng/type/form.FormController#$setPristine) to disable the validations while the dialog is closing.

Shidory: All right, I'll do it again some other way.

calmh:
This might work but has some issues... Looking at it and now understanding the problem, I have another idea. What if instead of

https://github.com/syncthing/syncthing/blob/3164341cf9cedfc31831f8bac8355581453251f7/gui/default/syncthing/core/syncthingController.js#L1828

we did

```
var folderCfg = angular.copy($scope.currentFolder);
```

to get our own copy of the config object. Then we can modify it as much as we like while the dialog is closing, because it does not see our changes.

Shidory:
For the setting and the device it's not a problem.

PR #6482, opened 2020-03-31
lib/scanner: Fix HashFile benchmark

greatroar:
Prevents the following error when running the benchmark with plain go test -bench=.:

```
--- FAIL: BenchmarkHashFile
    walk_test.go:558: open /_random.data: no such file or directory
FAIL
exit status 1
FAIL	github.com/syncthing/syncthing/lib/scanner	0.115s
```

### DIFF
--- a/lib/scanner/walk_test.go
+++ b/lib/scanner/walk_test.go
@@ -554,7 +554,7 @@ func BenchmarkHashFile(b *testing.B) {
 	b.ResetTimer()
 
 	for i := 0; i < b.N; i++ {
-		if _, err := HashFile(context.TODO(), fs.NewFilesystem(fs.FilesystemTypeBasic, ""), testdataName, protocol.MinBlockSize, nil, true); err != nil {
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Actually good point.

greatroar:
I'm not sure what NewFilesystem's "URI" is supposed to mean, because it's not documented anywhere. lib/api.browseFiles,

https://github.com/syncthing/syncthing/blob/9c67d57c281d97d666e1e8782a81690b9e278a54/lib/api/api.go#L1537

seems to rely on the interpretation of an empty URI as the global filesystem root.

imsodin:
That doesn't actually rely on global FS root:

```golang
	if current == "" {
		filesystem := fs.NewFilesystem(fsType, "")
		if roots, err := filesystem.Roots(); err == nil {
			return roots
		}
		return nil
	}
```

The `.Roots()` call returns the filesystem roots ('/' on unix, `C:\`, `D:\` etc. on windows) regardless of the filesystem URI.

greatroar:
Right, so it passes "" just to pass some value. I guess in that case it doesn't matter what interpretation "" has; NewFilesystem could translate it to ".". WDYT?

greatroar:
Sorry, I didn't see #6483. Closing in favor of that PR.

PR #6488, opened 2020-04-02
lib/api: Add LDAP search filters (fixes #5376)

### LINKED ISSUE #5376: Improve LDAP authentication

calmh:
## Purpose

This adds the functionality to run a user search with a filter for LDAP
authentication. The search is done after successful bind, as the binding
user. The typical use case is to limit authentication to users who are
member of a group or under a certain OU. For example, to only match
users in the "Syncthing" group in otherwise default Active Directory
set up for example.com:

    <searchBaseDN>CN=Users,DC=example,DC=com</searchBaseDN>
    <searchFilter>(&amp;(sAMAccountName=%s)(memberOf=CN=Syncthing,CN=Users,DC=example,DC=com))</searchFilter>

The search filter is an "and" of two criteria (with the ampersand being
XML quoted),

- "(sAMAccountName=%s)" matches the user logging in
- "(memberOf=CN=Syncthing,CN=Users,DC=example,DC=com)" matches members
  of the group in question.

Authentication will only proceed if the search filter matches precisely
one user.

## Testing

WFM

### DIFF
--- a/lib/api/api_auth.go
+++ b/lib/api/api_auth.go
@@ -172,7 +172,7 @@ func authLDAP(username string, password string, cfg config.LDAPConfiguration) bo
 	}
 
 	if cfg.SearchFilter == "" || cfg.SearchBaseDN == "" {
-		l.Warnln("LDAP Configuration: both searchFilter and searchBaseDN must be set, or neither.")
+		l.Warnln("LDAP configuration: both searchFilter and searchBaseDN must be set, or neither.")
 		return false
 	}
 
--- a/lib/api/api_auth.go
+++ b/lib/api/api_auth.go
@@ -171,13 +171,14 @@ func authLDAP(username string, password string, cfg config.LDAPConfiguration) bo
 		return true
 	}
 
+	if cfg.SearchFilter == "" || cfg.SearchBaseDN == "" {
+		l.Warnln("LDAP Configuration: both searchFilter and searchBaseDN must be set, or neither.")
+		return false
+	}
+
 	// If a search filter or search base is set we do an LDAP search for the
 	// user. If this matches precisely one user then we are good to go. The
-	// search filter uses the same %s interpolation as the bind DN. In
-	// practice it is required to set both the search base DN and the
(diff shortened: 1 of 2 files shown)


lib/api/api_auth.go
```diff
@@ -166,6 +166,31 @@ func authLDAP(username string, password string, cfg config.LDAPConfiguration) bo
 		return false
 	}
 
+	if cfg.SearchFilter != "" {
```
AudriusButkevicius, line 169: Should check that base is not empty as well as otherwise it will make no sense afaik

calmh: I changed this slightly but the new logic is the same - if either is set we run the code. It'll fail when just one is set, but I'd rather we fail always if just the filter is set than let authentication succeed and the filter just *look* active

calmh: I guess we could check for it specifically and fail the auth with a specific error pointing out the config problem...

PR #6490, opened 2020-04-02
lib/db: Custom Bloom filter implementation

greatroar:
Sorry for being a bit late to the party (given #6486), but here's a reimplementation of the Bloom filter code without external dependencies. For large Bloom filters, it attains much higher speeds, because it (1) doesn't have to compute modulos, and (2) doesn't hash its input, which is already a hash. That also skips all the trouble with murmurhash.

Tests and benchmarks are included, including a version that works with the current state of the code. benchstat results:

    name                  old time/op    new time/op     delta
    BloomfilterAdd1e5-8     22.4µs ± 1%     22.6µs ± 1%    +1.13%  (p=0.002 n=8+9)
    BloomfilterAdd1e6-8      216µs ± 2%      293µs ± 1%   +35.83%  (p=0.000 n=10+9)
    BloomfilterAdd1e7-8      3.84s ± 0%      1.07s ± 1%   -72.01%  (p=0.000 n=8+9)
    BloomfilterTest1e5-8     305ns ± 4%       45ns ± 1%   -85.15%  (p=0.000 n=10+10)
    BloomfilterTest1e6-8     338ns ± 1%       46ns ± 0%   -86.30%  (p=0.000 n=8+8)
    BloomfilterTest1e7-8     350ns ± 4%      116ns ± 2%   -66.86%  (p=0.000 n=10+8)
    
    name                  old speed      new speed       delta
    BloomfilterAdd1e5-8    143GB/s ± 2%    141GB/s ± 1%    -0.88%  (p=0.019 n=9+9)
    BloomfilterAdd1e6-8    148GB/s ± 2%    109GB/s ± 1%   -26.38%  (p=0.000 n=10+9)
    BloomfilterAdd1e7-8   83.3MB/s ± 0%  297.7MB/s ± 1%  +257.29%  (p=0.000 n=8+9)
    BloomfilterTest1e5-8   105MB/s ± 4%    707MB/s ± 1%  +573.33%  (p=0.000 n=10+10)
    BloomfilterTest1e6-8  94.5MB/s ± 1%  690.2MB/s ± 0%  +630.13%  (p=0.000 n=8+8)
    BloomfilterTest1e7-8  91.3MB/s ± 4%  272.8MB/s ± 8%  +198.74%  (p=0.000 n=10+9)


calmh:
Sweet. This looks like a neat and fast implementation, and I assume you wrote it custom for the purpose given the copyright. Thing is though, after looking at it for a few minutes and being only familiar with the high level details of how a bloom filter works, I can't claim to understand every bitshift in it. Given that and that open source is generally free as in a puppy (adopt it and maintain for life), I'm not 100% sure I want to become a maintainer for a bespoke bloom filter implementation.

How about a github.com/greatroar/bloom package, with a real person copyright, license, readme, and releases, and then a PR to use that as a dependency? You might say that's precisely the same thing, and to some extent it is, but it delineates maintenance responsibilities slightly differently.

greatroar:
I'll see what I can do.

PR #6494, opened 2020-04-04
cmd/strelay*: Various tweaks

calmh:
See individual commits. In principle:

- Make the tester return an error instead of a boolean, and pass this back to the client for visibility.
- Improve the client to show better errors about what's going on
- Fix incorrect handling of the "incorrect response code" error on the pool server
- Fix incorrect status code accounting on the pool server
- Add configurable queue length and concurrency for the test routines on the server

(merge as I rebase, I guess, unless there are a lots of changes in the meantime)

### DIFF
--- a/cmd/strelaypoolsrv/main.go
+++ b/cmd/strelaypoolsrv/main.go
@@ -334,10 +334,10 @@ func mimeTypeForFile(file string) string {
 func handleRequest(w http.ResponseWriter, r *http.Request) {
 	timer := prometheus.NewTimer(apiRequestsSeconds.WithLabelValues(r.Method))
 
-	lw := NewLoggingResponseWriter(w)
-
+	w = NewLoggingResponseWriter(w)
 	defer func() {
 		timer.ObserveDuration()
+		lw := w.(*loggingResponseWriter)
 		apiRequestsTotal.WithLabelValues(r.Method, strconv.Itoa(lw.statusCode)).Inc()
 	}()
 
@@ -397,7 +397,7 @@ func handlePostRequest(w http.ResponseWriter, r *http.Request) {
 		if debug {
 			log.Println("Failed to parse payload")
 		}
-		http.Error(w, err.Error(), 500)
+		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
@@ -406,7 +406,7 @@ func handlePostRequest(w http.ResponseWriter, r *http.Request) {
 		if debug {
 			log.Println("Failed to parse URI", newRelay.URL)
 		}
-		http.Error(w, err.Error(), 500)
+		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


cmd/strelaypoolsrv/main.go
```diff
@@ -335,6 +345,7 @@ func handleRequest(w http.ResponseWriter, r *http.Request) {
 	timer := prometheus.NewTimer(apiRequestsSeconds.WithLabelValues(r.Method))
 
 	lw := NewLoggingResponseWriter(w)
+	w = lw
```
AudriusButkevicius, line 348: This is weird? Get rid of one of them?

cmd/strelaysrv/pool.go
```diff
@@ -28,38 +33,62 @@ func poolHandler(pool string, uri *url.URL, mapping mapping) {
...
+		case http.StatusOK:
 			var x struct {
 				EvictionIn time.Duration `json:"evictionIn"`
 			}
-			err := json.NewDecoder(resp.Body).Decode(&x)
-			if err == nil {
+			if err := json.NewDecoder(resp.Body).Decode(&x); err == nil {
 				rejoin := x.EvictionIn - (x.EvictionIn / 5)
-				log.Println("Joined", pool, "rejoining in", rejoin)
+				log.Printf("Joined pool %s, rejoining in %v", pool, rejoin)
```
AudriusButkevicius, line 56: Doesn't this need \n?

cmd/strelaysrv/pool.go
```diff
@@ -28,38 +33,62 @@ func poolHandler(pool string, uri *url.URL, mapping mapping) {
...
+			continue
+
+		case httpStatusEnhanceYourCalm:
+			log.Printf("Failed to join %v: under load (rate limiting)", pool)
+			time.Sleep(time.Minute)
+			continue
+
+		case http.StatusUnauthorized:
+			log.Printf("Failed to join %v: IP address not matching external address", pool)
+			log.Println("Aborting")
```
AudriusButkevicius, line 82: Os.Exit perhaps?

cmd/strelaypoolsrv/main.go
calmh: We can't get rid of `w` as it's a parameter and will always be in scope, so the risk is that we will use it anyway. We can get rid of `lw` and instead use a type assertion in the defer, do you think that's cleaner?

cmd/strelaysrv/pool.go
calmh: `log.Print*` calls don't because they always end the line... I've never decided on whether it's better to include the `\n` always or never due to that :/

cmd/strelaysrv/pool.go
calmh: Well the relay works, it just can't register with the pool. Which might be secondary to it's relayness... So no, I think this should be as is (and as it was)

cmd/strelaypoolsrv/main.go
calmh: Fixed-up to do that, it was cleaner I think.

PR #6496, opened 2020-04-04
cmd/strelaypoolsrv: Fix race condition in caching

greatroar:
Successful LRU cache lookups modify the cache's recency list, so RWMutex.RLock isn't enough protection.

Secondarily, multiple concurrent lookups with the same key should not lead to multiple rate limiters being created, so release the lock only when presence of the key has been ascertained.

### DIFF
--- a/cmd/strelaypoolsrv/main.go
+++ b/cmd/strelaypoolsrv/main.go
@@ -115,10 +115,10 @@ var (
 	proto           string
 	statsRefresh    = time.Minute / 2
 
-	getMut      = sync.NewRWMutex()
+	getMut      = sync.NewMutex()
 	getLRUCache *lru.Cache
 
-	postMut      = sync.NewRWMutex()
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
To be honest, it's probably just easier to `lock, defer unlock`

greatroar:
Easier, yes, but rate.Limit.Allow, is a pretty long method that does a few potentially expensive things. On the other hand, it is O(1) when externally synchronized. Your call, of course.

AudriusButkevicius:
I think this is fine.

AudriusButkevicius:
I'll leave for some other maintainer to merge, just to get a second pair of eyes over this.

greatroar:
I've force-pushed a new version with the same code but an updated commit message; it contained some typos and a thinko (successful lookups cause a change in the structure, not unsuccessful ones).

PR #6498, opened 2020-04-05
Custom login fix

MpoyiMmg:
### Purpose

New HTML page with login form
New backend API request to process the login and hand out a cookie or  bearer token (see https://github.com/syncthing/syncthing/blob/master/lib/api/api_auth.go and the api code where that gets called)
Redirect unauthenticated requests to the login form (in the auth middleware probably)
We need a "logout" method to invalidate the session and send the user back to the login screen

### Testing

I tested the authentication with the new login interface.  To access the login form, click on action then login.

### Screenshots
Capture d'écran Deepin_zone de sélection _20200404221341

## Authorship

Gael MUTANDA
mpoyigaelmmg@gmail.com

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -68,7 +68,6 @@
           </a>
           <ul class="dropdown-menu">
             <li><a href="" ng-click="showSettings()"><span class="fas fa-fw fa-cog"></span>&nbsp;<span translate>Settings</span></a></li>
-            <li><a href="" ng-click="logout()"><span class="fas fa-fw fa-sign-out"></span>&nbsp;<span translate>Logout</span></a></li>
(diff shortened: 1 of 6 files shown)


calmh:
Looks like a good start.

MpoyiMmg:
Thank you! Are there any improvements to be made?

Le jeu. 16 avr. 2020 à 08:24, Jakob Borg <notifications@github.com> a
écrit :

tomasz1986:
> I tested the authentication with the new login interface. To access the login form, click on action then login.

Does the new code work in all Web browsers, including IE11 and mobile (Android, iOS) browsers?

MpoyiMmg:
By the way, there was a mistake with that sweater request! I fixed all the
problems and everything works fine now and I have to redo the sweater
request with what works

Issue #6457, opened 2020-03-26
Panic loop after upgrade to 1.4.1-rc3


eiszfuchs:
I have tried to find a solution to this problem, but I couldn't find anything that matched my situation, within these issues nor with a Google search.
I noticed that my folders went out of sync a few days back. I thought something crashed again, as it did from time to time. I upgraded to 1.4.0 when the latest stable came out, but that's when the panics started. I pulled the latest Release Candidate instead, as it supposedly fixed the issue with rebuilding the database, so Syncthing started again just fine.

I then noticed another round of panics, just as the re-scans kicked in. One of the folders made it panic, so I paused all folders and unpaused them until I got to the folder that caused the panics.

This is what Syncthing does, over and over, piling up log files:

```
[P4WYR] 13:14:35 INFO: Connection to XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX at xxx.xxx.xxx.xxx:59442-xxx.xxx.xxx.xxx:22000/tcp-client/TLS1.3-TLS_AES_128_GCM_SHA256 closed: unpausing folder "FOLDER_NAME" (xvicV-6grJj)
[P4WYR] 13:15:54 INFO: Device XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX folder "FOLDER_NAME" (xvicV-6grJj) has a new index ID (0xAC706B84FF20D287)
Panic at 2020-03-26T13:15:57+01:00
panic: device present in global list but missing as device/fileinfo entry
```

I couldn't find anything that could explain the error. The folder ID is the same on both devices. Folders haven't changed and are still correct on both ends.

### Version Information

Syncthing Version: v1.4.1-rc3
OS Version: Arch Linux on Raspberry Pi

imsodin:
In principle it's possible that the db repair on upgrade never happened for you if it failed the first time around (will file a PR fixing that). Can you please create a copy of your database (by default `~/.config/syncthing/index-v0.14.0.db`), then run Syncthing again with `STTRACE=db STRECHECKDBEVERY=1s` collecting the (possibly massive) log.

eiszfuchs:
Okay, I'm assuming this does a database check. I performed these commands:
```
cp -R index-v0.14.0.db index-v0.14.0.db-copy
[...]

PR #6502, opened 2020-04-06
lib/db: Don't get blocklists on drop and missing continue (ref #6457)

imsodin:
This fixes the panic in #6457, however not the underlying problem.  
After quite extensive debugging with @eiszfuchs (thanks) the problem there became clear: While the check on upgrade makes sure there is a file info for ever global list entry, it does not unmarshal that file info. On dropping a file we do, and as part of unmarshalling we also look for the block list. That lookup then failed and caused the file not found error (which mislead me for a long time, as the file itself is there, just not the blocks).

Now we only get the truncated file info when dropping a file, which is enough and thus avoids the panic. The file info is still without a block list, so a problem will likely come up at some other point. And I have no clue what might cause it in the first place. In the debugged instance no RCs prior to 1.4.1-rc.3 were ever installed (and known bugs causing missing blocklists were only present in RCs).

The missing (now added) continue was luckily harmless: Without it the key was deleted twice, which has no consequence.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -406,6 +406,7 @@ func (db *Lowlevel) checkGlobals(folder []byte, meta *metadataTracker) error {
 			if err := t.Delete(dbi.Key()); err != nil {
 				return err
 			}
+			continue
 		}
 
 		// Check the global version list for consistency. An issue in previous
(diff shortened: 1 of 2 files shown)


imsodin:
Wait merging please

imsodin:
Sorry for for dismissing again - went on a little hunt for other places were we needlessly get the full file info.

I'll probably in a further step remove `getFileByKey`, such that everyone needs to use `getFileTrunc` - however that involves a small (but potentially impactful) change to error handling so I'd rather put that into a separate PR.

PR #6507, opened 2020-04-06
cmd/strelaypoolsrv: Simplify LRU usage

greatroar:
Follow-up to #6496. I felt bad about the complicated locking, so I rewrote the method. It now has one Unlock at the same level of nesting as the Lock.

There is a functional change here: when a new rate limiter is constructed, it is used in the very same call. In the old implementation, clients got their first request for free. This has been the case since the rate limiting was introduced in 3cacb48f3cdf8c8e3d4c331ad737df69a1aaea65, but I assume it was an oversight.

### DIFF
--- a/cmd/strelaypoolsrv/main.go
+++ b/cmd/strelaypoolsrv/main.go
@@ -589,19 +589,15 @@ func limit(addr string, cache *lru.Cache, lock sync.Mutex, intv time.Duration, b
 	}
 
 	lock.Lock()
-	bkt, ok := cache.Get(addr)
-	if ok {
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Feels weird that the on is on the type assertion not on the cache return value?

AudriusButkevicius:
Also, if its not ok, we still call .Allow on a nil I believe?

calmh:
No, it's fine. If the type assertion fails it's either nil or "something else" (shouldn't happen), in which case it's assigned a new limiter.

PR #6509, opened 2020-04-07
lib/{fs,scanner}: gofmt from Go 1.14

calmh:
For passing correctness check


calmh:
Yeah I'm rolling back the Go 1.14 upgrade until we sort that out. This looks compatible with 1.13 though so we can leave this is (when the checks pass).

PR #6495, opened 2020-04-04
lib/osutil: Don't remove before rename on Windows (ref #6493)

calmh:
This was needed in ancient times but not currently.

(To be borne out by unit test on Windows builder)

Issue #6493, opened 2020-04-04
panic: runtime error: invalid memory address or nil pointer dereference


tomasz1986:
Today, right after clicking the "Pause All" button in the Web GUI, Syncthing suddenly stopped working. The process was still running in the background, but the GUI suddenly became unavailable.

A panic log was automatically created in the Syncthing home folder, which I have attached below. In addition, the `config.xml` file was reset, while the old one was renamed to `.syncthing.tmp.001460216`.

I have renamed the file back to `config.xml` and Syncthing seems to be working fine again.

[panic.log](https://github.com/syncthing/syncthing/files/4430903/panic.log)

### Version Information

Syncthing Version: v1.4.0
OS Version: Windows 10 Enterprise 2019 LTSC x64

calmh:
Actual panic is a "classic" failed handling of database close: 

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x198 pc=0xaef5d0]

goroutine 252049 [running]:
github.com/syndtr/goleveldb/leveldb.(*DB).isClosed(...)
	C:/BuildAgent/work/pkg/mod/github.com/syndtr/goleveldb@v1.0.1-0.20190923125748-758128399b1d/leveldb/db_state.go:230
github.com/syndtr/goleveldb/leveldb.(*DB).ok(...)
	C:/BuildAgent/work/pkg/mod/github.com/syndtr/goleveldb@v1.0.1-0.20190923125748-758128399b1d/leveldb/db_state.go:235
github.com/syndtr/goleveldb/leveldb.(*Snapshot).Get(0xc01d4bb800, 0xc00f1e0190, 0x48, 0x48, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	C:/BuildAgent/work/pkg/mod/github.com/syndtr/goleveldb@v1.0.1-0.20190923125748-758128399b1d/leveldb/db_snapshot.go:103 +0x60
github.com/syncthing/syncthing/lib/db/backend.leveldbSnapshot.Get(0xc01d4bb800, 0xc01cd4e450, 0xc00f1e0190, 0x48, 0x48, 0xc00011db80, 0x10000c007d08630, 0x0, 0x48, 0xc00f1e0190)
	C:/BuildAgent/work/174e136266f8a219/lib/db/backend/leveldb_backend.go:117 +0x69
[...]

PR #6510, opened 2020-04-07
lib/osutil: Fix "atomic" rename on Windows (ref #6495, ref #6493)

calmh:
So, in a funny plot twist, it turns out that WriteFile in Go 1.13
doesn't actually set the read only bit on Windows when called with
permissions 0444 so my test was broken. With an improved test it turns
out that Rename does not, in fact, overwrite a read-only file on
Windows. This adds a fix for that.

(Rename might get improved in Go 1.15...)

### DIFF
--- a/lib/osutil/atomic.go
+++ b/lib/osutil/atomic.go
@@ -9,6 +9,7 @@ package osutil
 import (
 	"errors"
 	"path/filepath"
+	"runtime"
 
 	"github.com/syncthing/syncthing/lib/fs"
 )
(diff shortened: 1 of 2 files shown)


lib/osutil/atomic.go
```diff
@@ -92,7 +93,15 @@ func (w *AtomicWriter) Close() error {
 		return err
 	}
 
-	if err := w.fs.Rename(w.next.Name(), w.path); err != nil {
+	err := w.fs.Rename(w.next.Name(), w.path)
+	if runtime.GOOS == "windows" && fs.IsPermission(err) {
+		// On Windows, we might not be allowed to rename over the file
+		// because it's read-only. Get us some write permissions and try
+		// again.
+		_ = w.fs.Chmod(w.path, 0644)
```
calmh, line 101: I'm not gonna try to do a stat and defer an undo of this. We're supposed to overwrite and destroy this file, with a file that we *have* write permissions on, and if it didn't help to set the old file to writeable then who knows what's even happening any more

PR #6514, opened 2020-04-07
cmd/syncthing, lib/db: Store upgrade info to throttle queries (fixes #6513)

### LINKED ISSUE #6513: Upgrade check at start is very aggressive
Since the new upgrade check (#6385) we're getting a significant flood of upgrade requests. The problem is that some clients are stuck in restart loops, for various reasons, and hammer the upgrade server. We can of course offload this in various ways (CDNs and whatnot) but it still seems like an oversight that we should handle better...

Server side rate limiting at the HTTPS level isn't enormously useful as the main cost is in the TLS handshake to begin with.

imsodin:
### Purpose

Make sure we check for updates max once per minute, and attempt upgrades to the same version max once per hour. For that we store the last times we checked for and attempted an upgrade plus which version we attempted to upgrade to. The reason for the distinction is that we want to reasonably quickly pick up new releases (e.g. if we push a quick followup-release to something broken),  put not attempt repeatedly to check for upgrades or to try to upgrade to the same version.

In cmd/syncthing, we handle a db backend, not a lowlevel. However that's actually enough for `db.NamespacedKV`, I just needed to replace lowlevel with backend there.

### Testing

Manually tested that upgrading still works. Then put garbage into the release url and the first time an error was printed, the second time not (check was skipped as I started within a minute).

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -148,15 +148,15 @@ var (
 	innerProcess    = os.Getenv("STMONITORED") != ""
 	noDefaultFolder = os.Getenv("STNODEFAULTFOLDER") != ""
 
-	errConcurrentUpgrade    = errors.New("upgrade prevented by other running Syncthing instance")
-	errTooEarlyUpgradeCheck = fmt.Errorf("last upgrade check happened less than %v ago, skipping", upgradeCheckInterval)
-	errTooEarlyUpgrade      = fmt.Errorf("last upgrade happened less than %v ago, skipping", upgradeInterval)
-
-	upgradeCheckInterval = time.Minute
-	upgradeInterval      = time.Hour
+	upgradeCheckInterval = 5 * time.Minute
+	upgradeRetryInterval = time.Hour
 	upgradeCheckKey      = "lastUpgradeCheck"
 	upgradeTimeKey       = "lastUpgradeTime"
 	upgradeVersionKey    = "lastUpgradeVersion"
+
+	errConcurrentUpgrade    = errors.New("upgrade prevented by other running Syncthing instance")
+	errTooEarlyUpgradeCheck = fmt.Errorf("last upgrade check happened less than %v ago, skipping", upgradeCheckInterval)
+	errTooEarlyUpgrade      = fmt.Errorf("last upgrade happened less than %v ago, skipping", upgradeRetryInterval)
 )
 
 type RuntimeOptions struct {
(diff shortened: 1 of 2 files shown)


cmd/syncthing/main.go
```diff
@@ -907,6 +901,32 @@ func autoUpgrade(cfg config.Wrapper, app *syncthing.App, evLogger events.Logger)
 	}
 }
 
+func initialAutoUpgrade(misc *db.NamespacedKV) (upgrade.Release, error) {
+	if last, ok, err := misc.Time(upgradeCheckKey); err != nil {
+		return upgrade.Release{}, err
```
calmh, line 906: I think we should be more accepting of the database throwing an error, here and below. There being an unmarshal error on the time shouldn't be reason to block an upgrade, for example? We just avoid the upgrade if everything checks out and the times say we should avoid it, otherwise we try, imho.

cmd/syncthing/main.go
```diff
@@ -907,6 +901,32 @@ func autoUpgrade(cfg config.Wrapper, app *syncthing.App, evLogger events.Logger)
...
+		// Only check time if we try to upgrade to the same release.
+		if lastTime, ok, err := misc.Time(upgradeTimeKey); err != nil {
+			return upgrade.Release{}, err
+		} else if ok && time.Since(lastTime) < upgradeInterval {
+			return upgrade.Release{}, errTooEarlyUpgrade
+		}
+	}
+	_ = misc.PutString(upgradeVersionKey, release.Tag)
+	_ = misc.PutTime(upgradeTimeKey, time.Now())
+	return release, upgrade.To(release)
```
calmh, line 927: This is a sneakily hidden "oh, and also do the actual upgrade" at the end of the function :) I think this function would be better off as a `checkEarlyUpgrade(misc) (upgradeTo Release, shouldUpgrade bool)` and then the actual upgrade step outside if it returns a release + true?

cmd/syncthing/main.go
```diff
@@ -147,7 +148,15 @@ var (
 	innerProcess    = os.Getenv("STMONITORED") != ""
 	noDefaultFolder = os.Getenv("STNODEFAULTFOLDER") != ""
 
-	errConcurrentUpgrade = errors.New("upgrade prevented by other running Syncthing instance")
+	errConcurrentUpgrade    = errors.New("upgrade prevented by other running Syncthing instance")
+	errTooEarlyUpgradeCheck = fmt.Errorf("last upgrade check happened less than %v ago, skipping", upgradeCheckInterval)
+	errTooEarlyUpgrade      = fmt.Errorf("last upgrade happened less than %v ago, skipping", upgradeInterval)
+
+	upgradeCheckInterval = time.Minute
```
calmh, line 155:
```suggestion
	upgradeCheckInterval = 5 * time.Minute
```

cmd/syncthing/main.go
```diff
@@ -147,7 +148,15 @@ var (
...
+	upgradeInterval      = time.Hour
```
calmh, line 156: `upgradeRetryInterval`?

PR #6518, opened 2020-04-08
lib/db: Make database GC a service, stop on Stop()

calmh:
This makes the GC runner a service that will stop fairly quickly when
told to.

As a bonus, STTRACE=app will print the service tree on the way out,
including any errors they've flagged.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -8,12 +8,15 @@ package db
 
 import (
 	"bytes"
+	"context"
 	"encoding/binary"
 	"time"
 
 	"github.com/syncthing/syncthing/lib/db/backend"
 	"github.com/syncthing/syncthing/lib/protocol"
 	"github.com/syncthing/syncthing/lib/sync"
+	"github.com/syncthing/syncthing/lib/util"
+	"github.com/thejerf/suture"
 	"github.com/willf/bloom"
 )
 
@@ -40,24 +43,30 @@ const (
 // database can only be opened once, there should be only one Lowlevel for
 // any given backend.
 type Lowlevel struct {
+	*suture.Supervisor
 	backend.Backend
 	folderIdx          *smallIndex
 	deviceIdx          *smallIndex
 	keyer              keyer
 	gcMut              sync.RWMutex
 	gcKeyCount         int
-	gcStop             chan struct{}
 	indirectGCInterval time.Duration
 	recheckInterval    time.Duration
 }
 
 func NewLowlevel(backend backend.Backend, opts ...Option) *Lowlevel {
 	db := &Lowlevel{
+		Supervisor: suture.New("db.Lowlevel", suture.Spec{
+			// Only log restarts in debug mode.
+			Log: func(line string) {
+				l.Debugln(line)
+			},
+			PassThroughPanics: true,
+		}),
 		Backend:            backend,
 		folderIdx:          newSmallIndex(backend, []byte{KeyTypeFolderIdx}),
 		deviceIdx:          newSmallIndex(backend, []byte{KeyTypeDeviceIdx}),
 		gcMut:              sync.NewRWMutex(),
-		gcStop:             make(chan struct{}),
 		indirectGCInterval: indirectGCDefaultInterval,
 		recheckInterval:    recheckDefaultInterval,
 	}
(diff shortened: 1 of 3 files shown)


lib/syncthing/syncthing.go
```diff
@@ -475,3 +482,37 @@ func (e *controller) Shutdown() {
...
+		if sub, ok := svc.(supervisor); ok {
+			printServiceTree(w, sub, level+1)
+		} else {
+			printService(w, svc, level+1)
+		}
+	}
+}
+
+func printService(w io.Writer, svc interface{}, level int) {
+	type errorer interface{ Error() error }
```
imsodin, line 506: Isn't this just `error`?

lib/db/lowlevel.go
```diff
@@ -573,6 +577,21 @@ func (db *Lowlevel) gcIndirect() error {
 	// Indirection GC needs to run when there are no modifications to the
 	// FileInfos or indirected items.
 
+	// checkCtx is a helper that periodically checks the ctx for Done() and
+	// if so returns an error.
```
imsodin, line 581: Is the "ctx-check-throttling" actually useful? I'd expect receiving from `ctx.Done()` to be fairly cheap and we do quite a few operations in between. I haven't been shy throwing context checks into database iterations, so if it is relevant that should probably be changed too.

lib/syncthing/syncthing.go
```diff
@@ -475,3 +482,37 @@ func (e *controller) Shutdown() {
 func (e *controller) ExitUpgrading() {
 	e.Stop(ExitUpgrade)
 }
+
+type supervisor interface{ Services() []suture.Service }
```
imsodin, line 486: Why do you need that type instead of using `*suture.Supervisor` directly?

lib/db/lowlevel.go
calmh: I’ll check, hopefully you’re right.

lib/syncthing/syncthing.go
calmh: We typically wrap the supervisor so the concrete type we’re looking at at any given moment isn’t a *suture.Supervisor. The top level is but not when we’ve dug into the tree.

lib/syncthing/syncthing.go
calmh: Nope. :)

lib/db/lowlevel.go
calmh: It's about 5% of a bloom filter lookup, so not *completely* negligible in a tight loop, but in the middle of a database iteration that might (worst case) involve I/O I think you're right and it is negligible. I'll remove the complexities.

lib/syncthing/syncthing.go
imsodin: Brainz was on ice apparently xD

lib/syncthing/syncthing.go
```diff
@@ -475,3 +482,37 @@ func (e *controller) Shutdown() {
...
+
+func printService(w io.Writer, svc interface{}, level int) {
+	type errorer interface{ Error() error }
+
+	t := "-"
+	if _, ok := svc.(supervisor); ok {
+		t = "+"
+	}
+	fmt.Fprintln(w, strings.Repeat("  ", level), t, svc)
+	if es, ok := svc.(errorer); ok {
```
AudriusButkevicius, line 513: You can just do: `foo.(interface { Error() error })` to avoid the type def.

calmh: Yeah but is that any clearer? I dunno 🤷‍♂

AudriusButkevicius: Probably, than a typedef in the function body.

imsodin: I don't care either way - if I needed to come up with any rationale I'd probably go with Audrius variant as it only appears once and thus when reading you don't need to search for the definition of `errorer`. That's however super contrived, it's pretty clear what's going on anyway

One might even throw in a third option if one wanted to bikeshed hard: Define the interface in lib/util or somewhere:
```
 $ grep -r "func() error" lib/ | wc -l
28
```
I hope it's clear I am not actually suggesting doing that for this PR though :)

PR #6523, opened 2020-04-12
improve device status

chadsmiley:
The status icon was not being determined correctly. Here are my findings:

1. The current device was always being included and it status is "disconnected"
2. The syncthingStatus function was passing the device id instead of the device object like other places.
3. Device status for disconnected was not valid, might be because of #1 finding
4. Folders that had been paused where not included
5. device status of "unused" had a dash at the end, seems unusual

Changes
1. Added check for devices of the current device, so it would be ignored
2. Passed the device object to syncthingStatus
3. Added disconnected status for devices
4. Added paused status for folders
5. Changed the syncthingStatus for unused status to unused-, so there would be no impact on the rest of the system.

Testing-
I pause and disconnected both devices and folders. Also caused folders to error and caused normal syncing.

My first Syncthing change so thanks for your patience.  If there is any question or changes please let me know.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -928,6 +928,9 @@ angular.module('syncthing.core')
         $scope.deviceStatus = function (deviceCfg) {
             var status = '';
 
+            if( deviceCfg.deviceID === this.myID){
+                return '';
+            }
             if ($scope.deviceFolders(deviceCfg).length === 0) {
                 status = 'unused-';
             }
@@ -987,6 +990,9 @@ angular.module('syncthing.core')
                     case 'syncing':
                         syncCount++;
                         break;
+                    case 'paused':
+                        pauseCount++;
+                        break;
                     case 'stopped':
                     case 'unknown':
                     case 'outofsync':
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
So I think having an icon in case of errors makes sense, having a notice icon in case of disconencted devices does not make sense, as you might expect them to be disconnected.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -928,6 +928,9 @@ angular.module('syncthing.core')
         $scope.deviceStatus = function (deviceCfg) {
             var status = '';
 
+            if( deviceCfg.deviceID === this.myID){
+                return '';
+            }
```
AudriusButkevicius, line 933: What case is this trying to address?

chadsmiley: The current device always shows as discounted. Maybe this is why disconnected was not included.

chadsmiley:
I can see why you would say that about disconnected.  A phone that is on mobile data, or a laptop that is turned off as it travels.  Would a person rather know that there is a potential problem rather than nothing.

I respect your decision, and you have more experience with the Syncthing than I do. If you would like me to adjust anything please let me know.

AudriusButkevicius:
I think there are plenty of cases where having a device disconnected is considered ok. People have setups where they work on one or the other machine, and expect it to be disconnected.

Your idea does not apply generally, it applies to a subset of users, so for the cases where it does not apply there will be confusion.

chadsmiley:
No changes are needed then.  Based on the comment in the code the only time the pause icon should be used is when all devices are paused.

PR #6529, opened 2020-04-13
lib/db: Refactor getting global lists

imsodin:
Removes some duplication around getting and unmarshaling global lists/`VersionList`.

No change in behaviour except for an old transition which will no longer ignore unmarshal errors (as it doesn't handle it itself).

### DIFF
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -391,7 +391,7 @@ func (db *schemaUpdater) updateSchema6to7(_ int) error {
 		err := t.withNeedLocal(folder, false, func(f FileIntf) bool {
 			name := []byte(f.FileName())
 			global := f.(protocol.FileInfo)
-			fl, err := t.getGlobalVL(gk, folder, name)
+			fl, err := t.getGlobalVersions(gk, folder, name)
 			if backend.IsNotFound(err) {
 				// If there is no global list, we hardly need it.
 				key, err := t.keyer.GenerateNeedFileKey(nk, folder, name)
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -391,12 +391,8 @@ func (db *schemaUpdater) updateSchema6to7(_ int) error {
 		err := t.withNeedLocal(folder, false, func(f FileIntf) bool {
 			name := []byte(f.FileName())
 			global := f.(protocol.FileInfo)
-			gk, delErr = db.keyer.GenerateGlobalVersionKey(gk, folder, name)
-			if delErr != nil {
-				return false
-			}
-			svl, err := t.Get(gk)
-			if err != nil {
+			fl, err := t.getGlobalVL(gk, folder, name)
+			if backend.IsNotFound(err) {
 				// If there is no global list, we hardly need it.
 				key, err := t.keyer.GenerateNeedFileKey(nk, folder, name)
 				if err != nil {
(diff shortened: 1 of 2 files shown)


lib/db/schemaupdater.go
```diff
@@ -391,12 +391,8 @@ func (db *schemaUpdater) updateSchema6to7(_ int) error {
 		err := t.withNeedLocal(folder, false, func(f FileIntf) bool {
 			name := []byte(f.FileName())
 			global := f.(protocol.FileInfo)
-			gk, delErr = db.keyer.GenerateGlobalVersionKey(gk, folder, name)
-			if delErr != nil {
-				return false
-			}
-			svl, err := t.Get(gk)
-			if err != nil {
+			fl, err := t.getGlobalVL(gk, folder, name)
```
AudriusButkevicius, line 394: VL -> Versions?

It's not that long to require an abberviation, and makes it less cryptic.

imsodin:
lgtm?

imsodin:
ping

lib/db/transactions.go
```diff
@@ -473,11 +475,8 @@ func (t readWriteTransaction) putFile(fkey []byte, fi protocol.FileInfo, truncat
 func (t readWriteTransaction) updateGlobal(gk, keyBuf, folder, device []byte, file protocol.FileInfo, meta *metadataTracker) ([]byte, bool, error) {
 	l.Debugf("update global; folder=%q device=%v file=%q version=%v invalid=%v", folder, protocol.DeviceIDFromBytes(device), file.Name, file.Version, file.IsInvalid())
 
-	var fl VersionList
-	svl, err := t.Get(gk)
-	if err == nil {
-		_ = fl.Unmarshal(svl) // Ignore error, continue with empty fl
-	} else if !backend.IsNotFound(err) {
+	fl, err := t.getGlobalVersionsByKey(gk)
```
AudriusButkevicius, line 478: This changes behaviour, we no longer ignore the error.

lib/db/schemaupdater.go
```diff
@@ -405,13 +401,9 @@ func (db *schemaUpdater) updateSchema6to7(_ int) error {
...
-			}
-			var fl VersionList
-			err = fl.Unmarshal(svl)
-			if err != nil {
-				// This can't happen, but it's ignored everywhere else too,
-				// so lets not act on it.
-				return true
+			} else if err != nil {
+				delErr = err
+				return false
```
AudriusButkevicius, line 406: Again, different behaviour in case of marshal error?

lib/db/transactions.go
```diff
@@ -112,23 +112,34 @@ func (t readOnlyTransaction) fillFileInfo(fi *protocol.FileInfo) error {
...
+
+	return vl, nil
+}
+
+func (t readOnlyTransaction) getGlobal(keyBuf, folder, file []byte, truncate bool) ([]byte, FileIntf, bool, error) {
+	vl, err := t.getGlobalVersions(keyBuf, folder, file)
+	if backend.IsNotFound(err) {
+		return nil, nil, false, nil
+	} else if err != nil {
 		return nil, nil, false, err
```
AudriusButkevicius, line 143: used to return keybuf in some cases, not sure if relevant.

lib/db/transactions.go
```diff
@@ -112,23 +112,34 @@ func (t readOnlyTransaction) fillFileInfo(fi *protocol.FileInfo) error {
...
+		return VersionList{}, err
+	}
+
+	return vl, nil
+}
+
+func (t readOnlyTransaction) getGlobal(keyBuf, folder, file []byte, truncate bool) ([]byte, FileIntf, bool, error) {
+	vl, err := t.getGlobalVersions(keyBuf, folder, file)
+	if backend.IsNotFound(err) {
+		return nil, nil, false, nil
```
AudriusButkevicius, line 141: used to return keybuf, not sure if relevant tho?

lib/db/transactions.go
imsodin: Yeah, right. However that's actually the "new correct behaviour" since https://github.com/syncthing/syncthing/pull/6411 where I apparently missed this.

AudriusButkevicius:
What about the last comment? Is that still kosher?

imsodin:
Originally I considered it a desirable change for consistency, however db transition are actually the kind of place to keep legacy stuff and just not touch it at all. will revert soon.TM

lib/db/schemaupdater.go
imsodin: change undone

PR #6533, opened 2020-04-14
lib/model: Harden sanitizePath

greatroar:
sanitizePath in lib/model compiles the same RE every time it is called. Moving the RE to a global is not only faster, but also catches errors in the RE earlier.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -14,11 +14,11 @@ import (
 	"net"
 	"path/filepath"
 	"reflect"
-	"regexp"
 	"runtime"
 	"strings"
 	stdsync "sync"
 	"time"
+	"unicode"
 
 	"github.com/pkg/errors"
 	"github.com/thejerf/suture"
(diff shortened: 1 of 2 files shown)


calmh:
With the thing inside the function it's essentially a constant and its scope is clear. With it as a package global it can in principle be used and mutated anywhere in the (very large) package...

greatroar:
The package being large is a different problem altogether :)

Now that think about it, does this function sanitize the path enough? What comes in is a label sent by a remote device, but I don't see a check whether that's even UTF-8. E.g.,

    sanitizePath("\xff") => "\xff"

Perhaps it would be better to enumerate what is allowed than try to filter out what is not.

calmh:
Increasing the safety is certainly welcome. The comment describes the things considered invalid and why.

greatroar:
I rewrote sanitizePath and added a fuzz test.

lib/model/model.go
```diff
@@ -2709,6 +2712,20 @@ func (m *syncMutexMap) Get(key string) sync.Mutex {
 // whitespace is collapsed to a single space. Additionally, whitespace at
 // either end is removed.
 func sanitizePath(path string) string {
-	invalid := regexp.MustCompile(`([[:cntrl:]]|[<>:"'/\\|?*\n\r\t \[\]\{\};:!@$%&^#])+`)
-	return strings.TrimSpace(invalid.ReplaceAllString(path, " "))
+	buf := bytes.NewBuffer(make([]byte, 0, len(path)))
```
calmh, line 2715: I think this might be slightly more idiomatic/efficient as a strings.Builder?

greatroar: Good point. See updated PR.

PR #6535, opened 2020-04-14
lib/model: Hide dot files when syncing to a Windows system

jesselucas:
### Purpose
When syncing a dotfile (files matching .*) to a Windows system the file will remain hidden. This is enabled by default and controlled by a folder level configuration hideDotFiles (that defaults to true). This change only affects Windows. Addressing issue: https://github.com/syncthing/syncthing/issues/5162

### Testing
Unit test, lib/model/sharedpullerstate_windows_test.go. GUI change is in the folder's edit, advanced modal section on Windows systems only.

### Screenshots
hide_dot_files_win


imsodin:
Thanks for this contribution.

One thing I am not certain about: Should this be configurable? And if yes, should it be in the folder modal? I think most windows user will just be confused about what files starting with a dot are and why they should be special/hidden.

AudriusButkevicius:
I too think this should probably just happen, or on by default and an advanced config option for those who really don't want it, because I am not sure I agree with the original statement that all of these files should be hidden, just because of how we handle executables.

calmh:
I think on by default, advanced option to disable. Windows-only users won't have any of these files and won't care either way. Mixed-OS users syncing dotfiles might reasonably expect dotfiles to be hidden, at least I think so but guessing users' intentions accurately is of course impossible.

tomasz1986:
Please be very, very careful with this one. A lot of software in Windows actually does create and use dot files. Windows itself actually comes with at least a few very notable folders starting with a .dot. They are the .NET Framework related folders, which are by no means meant to be hidden.

Just on my currently used system, some of the software that creates .dot files is Font Forge, Radicale, T-Clock, VirtualBox, Chromium, Avidemux, GIMP, Krita, and all the Mozilla related stuff (Thunderbird, SeaMonkey, Tor Browser).

One big thing with hidden files and folders is that you cannot operate on them from CMD without removing the attribute, i.e. `del .hiddenFile` or `rd .hiddenFolder` will not work. Hiding the files will break all batch scripts that assume their visibility.

I would strongly suggest to have this option disabled by default. If not, then at least please consider the .NET Framework related files and folders, because having them suddenly hidden is not what a Windows user expects.

jesselucas:
Thanks for the feedback! I've changed the option to be disabled by default. If this seems acceptable to everyone I'll update the docs.

AudriusButkevicius:
I'm not sure however if I agree. 

Perhaps we could sync the hidden bit and on windows actually check if stuff is hidden, and on linux set it for dot files. Sure, if you sync dotnet via win -> linux -> win you'd probably be in for a surprise, but I'd rather make it a flag you have to switch off on Linux than the other way around?

imsodin:
Given the additional behaviour/meaning windows imparts on hiding stuff described by @tomasz1986 I am not sure it's a good idea to get involved with it at all. Or put differently: The semantics of hiding seems significantly different on linux and windows, so syncing it might not be desirable at all.

tomasz1986:
I am only speaking from a perspective of a lifelong Windows user, but to put it simply,  we do not expect our files to be hidden just because their filename has a dot in front of it :smirk:.

Apart from .NET, some software uses .dot files with no apparent reason, like the aforementioned T-Clock, a tiny Windows-only program, which has a file called `.reset_portable.cmd` that by no means is meant to be hidden.

I have absolutely no problems with an option to hide such files, but having this enabled by default only because the files are being synced between Windows and Linux machines would be unexpected, at least for me, especially if "Linux" includes Android as well.

jesselucas:
How's everyone feeling about this being disabled by default or is this better left alone and close this for now?

calmh:
OK then, closing this as undesirable. Whaddayaknow.

tomasz1986:
I am very sorry if my worries were the reason for closing. I just do not feel that this is safe.

However, you can easily just use a batch script like

```
for /f delims^= %%a in ('dir /ad-h /b /s .*') do (
	attrib +h "%%a"
)
```

and set it to run every hour or so with Task Scheduler to hide all .dot folders in Windows if you really need this feature.

PR #6536, opened 2020-04-14
gui, lib/config: Add GUI user and password notification (fixes #4703)

### LINKED ISSUE #4703: Suggest setting up auth on initial startup

Yithis:
### Purpose

Recreation of #6444 PR. Add warning about setting up auth on initial startup (fixes #4703).

### Testing

Building and checking if everything is working properly.

### Screenshot

Screenshot from 2020-04-20 01-54-46

### DIFF
--- a/gui/default/syncthing/core/notifications.html
+++ b/gui/default/syncthing/core/notifications.html
@@ -126,6 +126,9 @@
       <p>
       <span translate>Username/Password has not been set for the GUI authentication. Please consider setting it up.</span>
       </p>
+      <p>
+      <span translate>If you want to prevent other users on this computer to access Syncthing and through it your files, consider setting up authentication.</span>
+      </p>
     </div>
     <div class="panel-footer">
       <button type="button" class="btn btn-sm btn-default pull-right" ng-click="showSettings()">
--- a/gui/default/syncthing/core/notifications.html
+++ b/gui/default/syncthing/core/notifications.html
@@ -123,8 +123,8 @@
       <h3 class="panel-title"><span class="fas fa-bolt"></span>&nbsp;<span translate>GUI Authentication: Set User and Password</span></h3>
     </div>
     <div class="panel-body">
-      <p translate>Username/Password has not been set for GUI authentication.</p>
-      <p translate>If you want to prevent other users on this computer to access Syncthing and through it your files, consider setting up authentication.</p>
+      <p><span translate>Username/Password has not been set for GUI authentication.</span></p>
+      <p><span translate>If you want to prevent other users on this computer to access Syncthing and through it your files, consider setting up authentication.</span></p>
     </div>
     <div class="panel-footer">
       <button type="button" class="btn btn-sm btn-default pull-right" ng-click="showSettings()">
--- a/gui/default/syncthing/core/notifications.html
+++ b/gui/default/syncthing/core/notifications.html
@@ -123,8 +123,9 @@
       <h3 class="panel-title"><span class="fas fa-bolt"></span>&nbsp;<span translate>GUI Authentication: Set User and Password</span></h3>
     </div>
     <div class="panel-body">
-      <p><span translate>Username/Password has not been set for GUI authentication.</span></p>
-      <p><span translate>If you want to prevent other users on this computer to access Syncthing and through it your files, consider setting up authentication.</span></p>
+      <p>
+      <span translate>Username/Password has not been set for the GUI authentication. Please consider setting it up.</span>
+      </p>
     </div>
     <div class="panel-footer">
       <button type="button" class="btn btn-sm btn-default pull-right" ng-click="showSettings()">
--- a/gui/default/syncthing/core/notifications.html
+++ b/gui/default/syncthing/core/notifications.html
@@ -124,10 +124,10 @@
     </div>
     <div class="panel-body">
       <p>
-      <span translate>Username/Password has not been set for the GUI authentication. Please consider setting it up.</span>
+        <span translate>Username/Password has not been set for the GUI authentication. Please consider setting it up.</span>
       </p>
       <p>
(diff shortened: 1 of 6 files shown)


Yithis:
@imsodin What do you think about this solution? There is one test about overridden vales that is not passing and I'm not sure if I should change test or something more in my code?

imsodin:
That test is a bit annoying in that you either need the new notification id to the expected options in the test or update the test data in overriden.xml with the new config version (31).

Yithis:
I fixed things you mentioned. Is everything correct now?

Yithis:
@imsodin, what do you think?

gui/default/syncthing/core/syncthingController.js
```diff
@@ -448,6 +448,10 @@ angular.module('syncthing.core')
                 && (!guiCfg.user || !guiCfg.password)
                 && guiCfg.authMode !== 'ldap'
                 && !guiCfg.insecureAdminAccess;
+
+            if (guiCfg.user && guiCfg.password) {
+            	dismissNotification('authenticationUserAndPassword');
```
imsodin, line 453: Indentation

gui/default/syncthing/core/notifications.html
```diff
@@ -116,3 +116,25 @@ <h3 class="panel-title"><span class="fas fa-bolt"></span>&ensp;<span translate>A
...
 </notification>
+
+<notification id="authenticationUserAndPassword">
+  <div class="panel panel-success">
+    <div class="panel-heading">
+      <h3 class="panel-title"><span class="fas fa-bolt"></span>&nbsp;<span translate>GUI Authentication: Set User and Password</span></h3>
+    </div>
+    <div class="panel-body">
+      <p>
+      <span translate>Username/Password has not been set for the GUI authentication. Please consider setting it up.</span>
```
imsodin, line 127:
```suggestion
      <span translate>Username/Password has not been set for GUI authentication. Please consider setting it up.</span>
```
Feels more natural to me. And generally I'd add a short explanation of why you want to set up authentication - there's good reasons for both choices. Something like e.g. "If you want to prevent other users on this computer to access Syncthing and through it your files, consider setting up authentication." (which is maybe not a good sentence, but hopefully illustrates what I mean).

gui/default/syncthing/core/syncthingController.js
Yithis: Done

gui/default/syncthing/core/notifications.html
Yithis: Done. I used your sentence, since I don't have any better idea how to illustrate this

Yithis:
I fixed things you mentioned and added a screenshot to first comment.

gui/default/syncthing/core/notifications.html
```diff
@@ -116,3 +116,28 @@ <h3 class="panel-title"><span class="fas fa-bolt"></span>&ensp;<span translate>A
...
```
imsodin, line 127: Indentation (not commenting on more instances).

lib/config/config.go
```diff
@@ -107,6 +107,8 @@ func New(myID protocol.DeviceID) Configuration {
 	util.SetDefaults(&cfg.Options)
 	util.SetDefaults(&cfg.GUI)
 
+	cfg.Options.UnackedNotificationIDs = []string{"authenticationUserAndPassword"}
```
imsodin, line 110: I think this fits best in `.prepare`.

lib/config/config.go
```diff
@@ -417,7 +419,7 @@ nextPendingDevice:
 		cfg.Options.AlwaysLocalNets = []string{}
 	}
 	if cfg.Options.UnackedNotificationIDs == nil {
-		cfg.Options.UnackedNotificationIDs = []string{}
+		cfg.Options.UnackedNotificationIDs = []string{"authenticationUserAndPassword"}
```
imsodin, line 422: This is not the right place to add this, as `clean` is called on every config change and we don't want to reintroduce it every time. What should happen as part of clean though is to check if user&password are set and if yes, remove `"authenticationUserAndPassword"` if present.

Yithis: Fixed as suggested

lib/config/config.go
Yithis: Fixed as suggested

gui/default/syncthing/core/notifications.html
Yithis: Fixed, hope I found them all

lib/config/config.go
imsodin: I really don't know what I was thinking when suggesting this: It means the notification is readded on every change to the folder - that's bad. I am sorry for the back-and-forth...

Yithis: No problem, will it be ok if I will go back to the version when it is added in `.New`?

imsodin: Sure. And just to avoid a misunderstanding: The other change you made in that commit in `clean` was good.

lib/config/config.go
```diff
@@ -418,6 +420,12 @@ nextPendingDevice:
 	}
 	if cfg.Options.UnackedNotificationIDs == nil {
 		cfg.Options.UnackedNotificationIDs = []string{}
+	} else if cfg.GUI.User != "" && cfg.GUI.Password != "" {
+		for i, key := range cfg.Options.UnackedNotificationIDs {
+			if key == "authenticationUserAndPassword" {
+				cfg.Options.UnackedNotificationIDs = append(cfg.Options.UnackedNotificationIDs[:i], cfg.Options.UnackedNotificationIDs[i+1:]...)
```
AudriusButkevicius, line 426: add a break?
Alternatively, you might need to decrement i

Yithis: Fixed, thank you

Yithis:
Is there anything more to change?

PR #6537, opened 2020-04-15
lib/db: Switch to faster blobloom Bloom filter pkg

greatroar:
Follow-up to #6490. I've reimplemented my Bloom filter code to produce a [separate library](https://github.com/greatroar/blobloom), that is, in my benchmarks, between 4 and 14x faster than willf/bloom and never allocates when doing insertions or lookups. Perhaps more importantly (we're taking microseconds here), it allows easier control over the memory use. Finally, the problematic murmur3 package is no longer required.

Currently, the maximum size of the Bloom filter is hardwired to 32MiB, which means that some 2.7M elements can be inserted before its false positive rate starts rising above 1%. For a desired capacity below that number, it will use less memory.


calmh:
There are some devices out there with > 50 million files in them. I'm not sure how well those setups work in practice, but I'm sure they use a lot of RAM. Is there a technical reason for the fairly low ceiling, or could the filter simply accept larger sizes?

greatroar:
I have off-by-one error in the exponent: 27 million, not 2.7 million.

But anyway, the ceiling is there as an example. If raised to half a GiB, the filter would support 500M elements at an FPR of 2%. The technical cap on the filter size is 512 * 2^32 bits, or 256GiB.

lib/db/lowlevel.go
```diff
@@ -27,7 +27,8 @@ const (
 	// than 100k. For fewer than 100k items we will just get better false
 	// positive rate instead.
 	indirectGCBloomCapacity          = 100000
-	indirectGCBloomFalsePositiveRate = 0.01 // 1%
+	indirectGCBloomFalsePositiveRate = 0.01    // 1%
+	indirectGCBloomMaxMem            = 8 << 22 // Use at most 32MiB memory
```
calmh, line 31:
```suggestion
	indirectGCBloomMaxMem            = 32 << 20 // Use at most 32MiB memory, which covers our desired FP rate at 27 M items
```

greatroar:
Oops, I forgot to build for 32-bit systems. BRB.

greatroar:
All green now. My tests all succeed when run with GOARCH=386, too.

lib/db/lowlevel.go
```diff
@@ -27,7 +27,8 @@ const (
 	// than 100k. For fewer than 100k items we will just get better false
 	// positive rate instead.
 	indirectGCBloomCapacity          = 100000
-	indirectGCBloomFalsePositiveRate = 0.01 // 1%
+	indirectGCBloomFalsePositiveRate = 0.01     // 1%
+	indirectGCBloomMaxMem            = 32 << 20 // Use at most 32MiB memory, which covers our desired FP rate at 27 M items
```
imsodin, line 31: I like if mem related variables mention their unit (e.g. here `...MaxMemBytes` or just `...MaxBytes`). Too much potential for confusion, e.g. further down where it needs to be converted to bits (nothing against the usage here, it seems obvious and clear here and now, but in general it isn't necessarily).

greatroar: Done.

PR #6538, opened 2020-04-15
build, lib/build: Build faster, auto detect program name

calmh:
This changes the build script to build all the things in one go
invocation, instead of one invocation per cmd. This is a lot faster
because it means more things get compiled concurrently. It's especially
a lot faster when things *don't* need to be rebuilt, possibly because it
only needs to build the dependency map and such once instead of once per
binary.

In order for this to work we need to be able to pass the same ldflags to
all the binaries. This means we can't set the program name with an
ldflag; instead I made it auto detect using os.Executable().

When it needs to rebuild everything (go clean -cache):

    ( ./old-build -gocmd go1.14.2 build all 2> /dev/null; )  65.82s user 11.28s system 574% cpu 13.409 total
    ( ./new-build -gocmd go1.14.2 build all 2> /dev/null; )  63.26s user 7.12s system 1220% cpu 5.766 total

On a subsequent run (nothing to build, just link the binaries):

    ( ./old-build -gocmd go1.14.2 build all 2> /dev/null; )  26.58s user 7.53s system 582% cpu 5.853 total
    ( ./new-build -gocmd go1.14.2 build all 2> /dev/null; )  18.66s user 2.45s system 1090% cpu 1.935 total

### DIFF
--- a/build.go
+++ b/build.go
@@ -25,7 +25,6 @@ import (
 	"os"
 	"os/exec"
 	"os/user"
-	"path"
 	"path/filepath"
 	"regexp"
 	"runtime"
@@ -405,12 +404,9 @@ func install(target target, tags []string) {
 		defer shouldCleanupSyso(sysoPath)
 	}
 
-	for _, pkg := range target.buildPkgs {
-		args := []string{"install", "-v"}
-		args = appendParameters(args, tags, pkg)
-
-		runPrint(goCmd, args...)
-	}
+	args := []string{"install", "-v"}
+	args = appendParameters(args, tags, target.buildPkgs...)
+	runPrint(goCmd, args...)
 }
 
 func build(target target, tags []string) {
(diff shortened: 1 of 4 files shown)


AudriusButkevicius:
I'd rather then start hardcoding program names in each binary, then do this stuff. People can name their stuff whatever they want, last thing I want is `lol-ma-look-at-me.exe 1.4.2` showing up in random places.

imsodin:
This needs some kind of notification in the changelog for packagers (e.g. @aviau, don't know who the arch etc people are ), I'll try to think of it when it's time (or maybe solve that with a way to persist changelogs other than github releases, if it's not too much work I believe that would have merit).

calmh:
I don't think it does. Yes, it's no longer required to set the `Program` variable in the linker options, but there is no harm in doing so (it won't error) and everything will work as it should.

imsodin:
Ah cool - I wrongly assumed that it will complain as `build.Program` doesn't exist anymore. Sorry for the noise.

PR #6543, opened 2020-04-16
Default ignores patterns

Shidory:
syncthing default ignores


lib/api/api.go
```diff
@@ -1173,6 +1174,30 @@ func (s *service) getDBIgnores(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+func (s *service) getIgnoresFromDeletedFolder(w http.ResponseWriter, r *http.Request) {
```
calmh, line 1177:
```suggestion
func (s *service) getIgnoresRaw(w http.ResponseWriter, r *http.Request) {
```

lib/api/api.go
```diff
@@ -1173,6 +1174,30 @@ func (s *service) getDBIgnores(w http.ResponseWriter, r *http.Request) {
...
+	qs := r.URL.Query()
+
+	path := qs.Get("path")
+	finalPath := path + "/.stignore"
+	if _, err := os.Stat(finalPath); !os.IsNotExist(err) {
```
calmh, line 1182: Don't `stat()` the file beforehand; even if this succeeds, the file may be gone by the time we get to `open()`. Instead just try to open and handle the error by returning from the handler.

lib/api/api.go
```diff
@@ -1173,6 +1174,30 @@ func (s *service) getDBIgnores(w http.ResponseWriter, r *http.Request) {
...
+
+		file, err := os.Open(finalPath)
+		if err != nil {
+			log.Fatal(err)
```
calmh, line 1186: Let's not crash here. :) Instead, something like

```
if err != nil {
  http.Error(w, "Not found", http.StatusNotFound)
  return
}
```

lib/api/api.go
```diff
@@ -1173,6 +1174,30 @@ func (s *service) getDBIgnores(w http.ResponseWriter, r *http.Request) {
...
+		}
+		defer file.Close()
+
+		b, err := ioutil.ReadAll(file)
```
calmh, line 1190: In fact, you can use `ioutil.ReadFile` to get all the file bytes directly without having to open+read+close. We typically call a byte slice `bs`.

lib/api/api.go
```diff
@@ -1173,6 +1174,30 @@ func (s *service) getDBIgnores(w http.ResponseWriter, r *http.Request) {
...
+
+		finalData := string(b[:len(b)])
```
calmh, line 1192: Simpler as just `string(bs)` without the length stuff, and you can do it directly in the map, so:

```
sendJSON(w, map[string]string{
  "ignore": string(bs),
})
```

lib/api/api.go
```diff
@@ -247,6 +247,7 @@ func (s *service) serve(ctx context.Context) {
 	getRestMux.HandleFunc("/rest/db/completion", s.getDBCompletion)              // device folder
 	getRestMux.HandleFunc("/rest/db/file", s.getDBFile)                          // folder file
 	getRestMux.HandleFunc("/rest/db/ignores", s.getDBIgnores)                    // folder
+	getRestMux.HandleFunc("/rest/db/path", s.getIgnoresFromDeletedFolder)        // folder
```
calmh, line 250: This should not be a `/db/` method I think. Maybe `/rest/system/raw-ignore`?

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -1,4 +1,4 @@
-<modal id="editDevice" status="default" icon="{{editingExisting ? 'fas fa-pencil-alt' : 'fas fa-desktop'}}" heading="{{editingExisting ? 'Edit Device' : 'Add Device' | translate}} {{currentDevice.name}}" large="yes" closeable="yes">
+<modal id="editDevice" status="default" icon="{{editingExisting ? 'fas fa-pencil-alt' : 'fas fa-desktop'}}" heading="{{editingExisting ? 'Edit Device : ' : 'Add Device' | translate}} {{currentDevice.name}}" large="yes" closeable="yes">
```
calmh, line 1: Looks like a change from another branch? remove

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1789,13 +1791,30 @@ angular.module('syncthing.core')
         };
 
         $scope.addFolder = function () {
+            $('#folderLabel').change(function() {
+                var path = $('#folderPath').val();
+                $http.get(urlbase + '/db/path?path=' + encodeURIComponent(path))
+                .success(function (data) {
+                    if(data.ignore){
+                        $('#folder-ignores textarea').val(data.ignore);
```
calmh, line 1799: Remove the disabled attr here as well? We've loaded the ignore patterns, so we might as well let the user edit them before creating the folder

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1768,6 +1769,7 @@ angular.module('syncthing.core')
                 })
                 .error(function (err) {
                     $('#folder-ignores textarea').val($translate.instant("Failed to load ignore patterns."));
+                    $('#folder-ignores textarea').removeAttr('disabled');
```
calmh, line 1772: Probably should not make this change. It's part of the other issue (editing ignore patterns that fail to load...)

Shidory: We've also already worked on the editing ignore patterns that fail to load

calmh: I know, but lets make that a separate change & PR

gui/default/syncthing/core/syncthingController.js
Shidory: Already at this level the user can modify the ignores.

gui/default/syncthing/core/syncthingController.js
Shidory: On the same branch or on another ?

lib/api/api.go
Shidory: can you clarify what you mean by that?

gui/default/syncthing/core/syncthingController.js
calmh: Separate issues => separate branches, separate PRs

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1705,6 +1705,7 @@ angular.module('syncthing.core')
         };
 
         $scope.editFolder = function (folderCfg) {
+            console.log($scope.config);
```
calmh, line 1708: remove

lib/api/api.go
```diff
@@ -1173,6 +1173,7 @@ func (s *service) getDBIgnores(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+
```
calmh, line 1176: remove

gui/default/syncthing/device/editDeviceModalView.html
calmh: not addressed

lib/api/api.go
calmh: You can remove the whole outer `if ... os.Stat()` thing. Either the file exists, and then you don't need it, or it does exist, and then you get the same error from `ioutil.ReadFile`. If you want to "mask" the does-not-exist error as a successful empty return, then do the `os.IsNotExist()` check on *that* error.

gui/default/syncthing/core/syncthingController.js
calmh: But we set the `disabled` attributes sometimes (when there is an error loading the folder), and we reuse the same modal for adding a folder here, so I think we still need to remove the `disabled` attribute, even if we think it's probably not set.

lib/api/api.go
```diff
@@ -1173,6 +1174,21 @@ func (s *service) getDBIgnores(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+func (s *service) getIgnoresRaw(w http.ResponseWriter, r *http.Request) {
+	qs := r.URL.Query()
+
+	path := qs.Get("raw-ignore")
+	finalPath := path + "/.stignore"
+	bs, err := ioutil.ReadFile(finalPath)
```
calmh, line 1182: Let's call the query param `path` and use filepath.Join:

```suggestion
	folderPath := qs.Get("path")
	ignoresPath := filepath.Join(folderPath, ".stignore")
	bs, err := ioutil.ReadFile(ignoresPath)
```

lib/api/api.go
```diff
@@ -247,6 +247,7 @@ func (s *service) serve(ctx context.Context) {
 	getRestMux.HandleFunc("/rest/db/completion", s.getDBCompletion)              // device folder
 	getRestMux.HandleFunc("/rest/db/file", s.getDBFile)                          // folder file
 	getRestMux.HandleFunc("/rest/db/ignores", s.getDBIgnores)                    // folder
+	getRestMux.HandleFunc("/rest/system/raw-ignore", s.getIgnoresRaw) 		     // folder
```
calmh, line 250:
```suggestion
	getRestMux.HandleFunc("/rest/db/ignores/raw", s.getIgnoresRaw) 		     // path
```

calmh: The ignores strictly speaking don't live in the `db`, but they probably should and we can keep the API endpoint there for the time being... sorry for changing my mind on this.

AudriusButkevicius:
Can someone give a brief intro how this is supposed to work?

Shidory:
We are supposed to give the user the possibility to create a default ignore pattenrs that would apply when creating a new folder

Shidory:
And if the folder had already been created and deleted, give the possibility to retrieve the ignorance that was applied to it.

lib/api/api.go
```diff
@@ -1173,6 +1174,21 @@ func (s *service) getDBIgnores(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+func (s *service) getIgnoresRaw(w http.ResponseWriter, r *http.Request) {
+	qs := r.URL.Query()
+
+	folderPath := qs.Get("raw-ignore")
```
calmh, line 1180:
```suggestion
	folderPath := qs.Get("path")
```

Again, lets call this `path`, please.

calmh:
@Shidory just needs a gofmt to pass that check as well.

@AudriusButkevicius There is a new field for default ignore patterns in the global settings. When creating a new folder we first try to load an existing .stignore (to avoid clobbering) and otherwise fill with the default ignore patterns. Which are empty by default, which matches the current behavior.

AudriusButkevicius:
Wouldn't it be better to do it in go?

Now anyone using the API/cli/rest will have to mimic the same behaviour.

calmh:
Do what in go? We already set all folder options including initial ignore patterns via the api (GUI).

AudriusButkevicius:
Ah yes, we no longer have any "defaults" in Go, so I guess this approach makes sense, even if it's sad.

calmh:
Well, I mean, technically it would be totally possible to apply the defaults from the config if the GUI sends a folder create with an empty set of ignore patterns, or maybe a magic flag. But I think that would be a much more surprising behavior and probably trickier to implement, as the physical folder creation is far separated from the config POST. And raises questions like should the patterns be applied when a new folder is just added to the config and syncthing restarted? That would seem to be better served then by a folder flag "use default ignores" and have them be always applied from the config instead of the .stignore. Would that be better?

AudriusButkevicius:
Not sure, it's probably good as it is, but I think all of the logic for folder creation/defaults being in the UI is a bad state to be in.

I know it's already an existing state, so I was just questioning whether we should keep piling on top of that.

In an ideal world you'd call `folder/defaults` which gives you the prepopulated default values, you tweak them in the UI and post them back I guess.

imsodin:
there is an issue about configurable folder defaults. i think a first step towards that would be exposing something like /rest/folder/default/config and then we could also have  /rest/folder/default/ignores. seems unrelated to this pr though

calmh:
Right. So in the interest of keeping things sane, we postpone this issue until a system for actual folder defaults exist, at which point we can revisit how to handle the default ignore patterns specifically.

PR #6544, opened 2020-04-16
lib/model: Delay watch setup on errors (fixes #5731)

### LINKED ISSUE #5731: 100% CPU every minute on watch setup retry
What happened: syncthing, for some folders, get stuck at 100% CPU for hours (state is `idle`)
What you expected to happen instead: CPU drops after a few minutes (when state is `idle`)
Any steps to reproduce the problem:
- set inotify limit too low
- use FSwatcher

The main issue is that there's no way to know the inotify limits to use. For instance:
- `sysctl.conf` has `fs.inotify.max_user_watches = 204800`
- directory (root folder) has `379320` directories (but only `103917` directories not ignored) and `1970011` files (but only `41029` files not ignored)

Increasing the inotify limit to 1M+ fixes the issue.

## Questions/thoughts/ideas

- does the inotify limit need to be higher than all directories (ignored + not ignored)?
- is there a way to list all files/directories to be watched? This way, we could prevent this issue by determining a correct value before increasing it with random values (here `204800` as stated in the docs doesn't help). Or perhaps just give the information in the error: `inotify limit was hit, XXX directories are to be watched (see ...)`.
- using `strace`, I've seen a lot of calls to `inotify_rm_watch`. I don't understand why? Is it because I've reached the inotify limit? After increasing the limit, there's no more calls.
- hitting the inotify limit should be reported as `INFO` or `WARNING` in the logs. I've searched a lot before understanding why my files were not in sync, enabling verbose helped a lot, because the message was clear: inotify issue (+ link to how to deal with the issue).
- I've used `sysctl -p` to update the limit right now, after editing `/etc/sysctl.conf`, perhaps it would be useful to update your docs for this very command, I think it's better as it uses the real configuration file (`/etc/sysctl.conf`), so no bad surprises on reboot.

## Thanks

Thank you for this tool, used alongside of Unison for other purposes, it's really great!

### Version Information

Syncthing Version: v1.1.3
OS Version: Debian stable (Stretch), using https://apt.syncthing.net/ repository

imsodin:
Delay trying to setup watches up to an hour on errors. If setting up is successful, the interval is reset if the watch lasted for longer than the retry interval, otherwise the retry interval is also increased and watching retried.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -677,6 +677,8 @@ func (f *folder) monitorWatch(ctx context.Context) {
 	var eventChan <-chan fs.Event
 	var errChan <-chan error
 	warnedOutside := false
+	var lastWatch time.Time
+	pause := time.Minute
 	for {
 		select {
 		case <-failTimer.C:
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Doesn't the error message say "a minute" and need to be updated?

PR #6546, opened 2020-04-17
Custom login fix

MpoyiMmg:
### Purpose

New HTML page with login form
New backend API request to process the login and hand out a cookie or bearer token (see https://github.com/syncthing/syncthing/blob/master/lib/api/api_auth.go and the api code where that gets called)
Redirect unauthenticated requests to the login form (in the auth middleware probably)
We need a "logout" method to invalidate the session and send the user back to the login screen

### Screenshots
Capture d'écran Deepin_zone de sélection _20200417162352
Capture d'écran Deepin_zone de sélection _20200417164640

## Authorship

Gael MUTANDA
mpoyigaelmmg@gmail.com


MpoyiMmg:
I made a new sweater request with the modifications I had to make compared to the previous one.

calmh:
I made some fixes to the backend login/logout and pushed to your branch.

So there are some things here which don't work as expected;

- When using the wrong username/password the error message appears correctly. But with repeated failures it goes away and no error is shown.
- The logout link looks wrong and doesn't appear to call the logout method.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -14,10 +14,63 @@ angular.module('syncthing.core')
...
+        function loginSuccessfull(){
             setInterval($scope.refresh, 10000);
             Events.start();
         }
 
+        $scope.login = function () {
+            var username = $scope.currentFolder.username
+            var password = $scope.currentFolder.password
+
+            $http.post("rest/login?username="+username+"&password="+password).success(
```
calmh, line 44: `urlbase` + `/rest/login...`

and some kind of URL quoting of the username/password.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -14,10 +14,63 @@ angular.module('syncthing.core')
...
+                    loginSuccessfull();
+
+                }).error(function(){
+                    
+                    checkLoginAndStart();
+
+                    $('#error').html(`<div class="alert alert-danger" role="alert" id="error">
+                                            <p style="color: #fff">Error! your credentials are not correct!\n Retry please...</p>
+                                        </div>`);
+                })
```
calmh, line 61: Again. Use some of the regular angular constructs.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -14,10 +14,63 @@ angular.module('syncthing.core')
...
+        }
+
+        function loginSuccessfull(){
             setInterval($scope.refresh, 10000);
             Events.start();
         }
 
+        $scope.login = function () {
+            var username = $scope.currentFolder.username
+            var password = $scope.currentFolder.password
```
calmh, line 42: Username and password have nothing to do with folders. This should be a `$scope.login` perhaps, and we should also make sure to not store the password here once login has been attempted.

lib/api/api_auth.go
```diff
@@ -42,6 +42,74 @@ type authAndSessionMiddleware struct {
 	evLogger   events.Logger
 }
 
+func basicAuthAndSessionMiddleware(cookieName string, guiCfg config.GUIConfiguration, ldapCfg config.LDAPConfiguration, next http.Handler, evLogger events.Logger) http.Handler {
```
calmh, line 45: This code was added back by mistake, it should not be in use, please remove again.

calmh:
This is starting to look reasonable. There are some things that might need a wider discussion;

@syncthing/maintainers Doing login in-page obviously requires loading the page without auth. For that purpose we now allow or should allow the following things to be loaded without auth:

- All the things that are not `/rest/...`: this includes all the GUI assets and such, but also our virtual metadata.js that includes our device ID. This is used by the HTTP interceptor to know where to look for CSRF cookies and stuff like that, and all that loads prior to login now. Is this a problem, you think? We leak the device ID to anyone who can talk to the GUI, but on the other hand the device ID is already available by connecting to the sync port which is usually more available than the GUI port...
- `/rest/svc/lang` because that tells the GUI what language to use. Currently it's not handled so this pops a basic auth login anyway right now. Now, *most* of the things we have under `/rest/svc` are not sensitive (getting languages, interpreting a device ID, getting a random string) but of course this is also where the usage report previous lives...

So how ugly exceptions do we want to make here? I can hack in an exception to allow `/rest/svc/lang` specifically without auth, or we could say that `/rest/svc` is outside of auth and move the report endpoint... Or we could invent a `/restNoAuth` and move/copy the `lang` and `metadata.js` things there... Or maybe remove `metadata.js` and refactor the login request to return the values with a successful response and then patch up the http interceptor somehow...

AudriusButkevicius:
I think we should make rest/svc no auth, move usage reports elsewhere.

I am still not sure I am a fan of the ajax login stuff. Seems like checking the cookies, if its not there redirecting to /login with a simple post form seems more simple? Then we probably just need the lang endpoint?

calmh:
To propose a design and revisit, then, as it will require reorganizing assets and slightly the rest endpoint.

PR #6547, opened 2020-04-17
Leaflet map fix definitive: Remove circle on map

keevBush:
I removed circle on the map relays stats


AudriusButkevicius:
Why?

keevBush:
radius change every time with Leaflet and @calmh had made a suggestion to remove it.

calmh:
needs cleanup of random whitespace (blank lines) added here and there, otherwise lgtm. I don't know what the problem is with the circles, but they don't add much value imho

AudriusButkevicius:
They could be made to add value, we just need to adjust the size and make them logarithmic scale.

calmh:
At that point it might as well be a heatmap? I mean, that's what we're simulating right?

AudriusButkevicius:
Sure, but I guess it would have to be logarithmic as well.
The idea is that we want to see where most bandwidth is being used up.

calmh:
Okay, so we leave the broken stuff in place until someone feels like fixing it.

Issue #5224, opened 2018-09-25
Accept new connections in place of old ones when appropriate


calmh:
When a connection dies we sometimes do not pick up on this until a few minutes later due to timeouts kicking in. In the meantime, if the other side tries to connect it is refused as a duplicate connection. The relevant code has this comment:

```
// We should not already be connected to the other party. TODO: This
// could use some better handling. If the old connection is dead but
// hasn't timed out yet we may want to drop *that* connection and keep
// this one. But in case we are two devices connecting to each other
// in parallel we don't want to do that or we end up with no
// connections still established...
```

We should implement the mentioned "better handling".

It's probably as easy as just dropping the existing connection if it's older than some threshold, otherwise dropping the new one.

(I'm sure there are still implementation complexities as we need to disconnect and wait for that to percolate everywhere before we can add the new connection etc.)

calmh:
Relates to #5219

PR #6548, opened 2020-04-18
Accept new connections in place of old ones when appropriate #5224

MikolajTwarog:
As suggested I added replacing connection when the current one is older than 5 seconds. To do that I added new field in Statistics structure.

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -46,8 +46,9 @@ var (
 )
 
 const (
-	perDeviceWarningIntv = 15 * time.Minute
-	tlsHandshakeTimeout  = 10 * time.Second
+	perDeviceWarningIntv    = 15 * time.Minute
+	tlsHandshakeTimeout     = 10 * time.Second
+	minConnectionReplaceAge = 10 * time.Second
 )
 
 // From go/src/crypto/tls/cipher_suites.go
@@ -276,7 +277,7 @@ func (s *service) handle(ctx context.Context) {
 		ct, connected := s.model.Connection(remoteID)
 
 		// Lower priority is better, just like nice etc.
-		if connected && (ct.Priority() > c.priority || time.Since(ct.Statistics().StartedAt) > 10*time.Second) {
+		if connected && (ct.Priority() > c.priority || time.Since(ct.Statistics().StartedAt) > minConnectionReplaceAge) {
 			l.Debugf("Switching connections %s (existing: %s new: %s)", remoteID, ct, c)
 		} else if connected {
 			// We should not already be connected to the other party. TODO: This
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
I am not sure this is how I imagine this to work. My personal thought on how I would handle this would be to add a mechanism to actively check the liveness of the connection, and only drop it if we deem it to be "dead". In this specific case we might drop a TCP connection over a UDP one which has very latient packets.

calmh:
I would settle for accepting a new connection if the old connection-to-be-replaced is at 10s old or something like that. That would add as a tie breaker in simultaneous connections, so we don't end up with dropping both of them, and it would handle replacing dead connections that only the other side has realized is dead.

In theory we might replace a connection with something worse (tcp -> quic), but that would only happen if the other side doesn't have the tcp connection any more and the quic connection is the attempt that goes through. We'll continue trying higher priority protocols so if tcp works we'll get back to it in time.

AudriusButkevicius:
Fine I guess. Do we really want the start time to live on the stats part?

MikolajTwarog:
I thought it would be better than adding new method to Connection.

MikolajTwarog:
Should I change anything more?

AudriusButkevicius:
Idk, it's probably fine, just want @calmh to confirm.

imsodin:
I think this is a good step.  
Ideally considering the huge rat-tail caused by a connection change, I'd issue a ping immediately on an incoming connection attempt. and if that fails drop the old connection and add this connection (and do this ping close dance outside of the `handle` loop and pass it back in if appropriate). If you feel like trying something like that (no guarantees it actually works out), that would of course be great, but I am also fine with the mechanism in this PR.

MikolajTwarog:
Right now I don't have enough time to try this approach. Maybe it would be good to create a new issue about this, so I can work on in some time later.

AudriusButkevicius:
I'll leave for @calmh to merge I guess.

lib/connections/service.go
```diff
@@ -276,7 +276,7 @@ func (s *service) handle(ctx context.Context) {
 		ct, connected := s.model.Connection(remoteID)
 
 		// Lower priority is better, just like nice etc.
-		if connected && ct.Priority() > c.priority {
+		if connected && (ct.Priority() > c.priority || time.Since(ct.Statistics().StartedAt) > 10*time.Second) {
```
calmh, line 279: This would be a good place for a named constant, `minConnectionReplaceAge`?

MikolajTwarog: I think you are right. I added it.

calmh:
Does it work though? :) I mean, it looks like it should, but did/can we test this somehow?

MikolajTwarog:
I was testing it using my two devices. I was switching WiFi on my second device. Every time I tried it new connection was accepted instead of the old one. It was not the case before my changes.

GermanCoding:
By the way there was a TODO in the code (https://github.com/syncthing/syncthing/pull/6548/files#diff-d31f92d68068bfede5c6b6b01e1b558bL282-L287) that still exists altough it was presumably fixed by this PR.

PR #6550, opened 2020-04-19
Fix page numbering

### LINKED ISSUE #6408: Page numbering on "Out  of Sync" pop-up
The "number of items to display" option at the foot of the pop-up page is (thankfully!) persistent, but so is the page number.  When the pop-up page is closed _(either with the "Close" button or by clicking on originating page "behind" the pop-up page)_, the page number that was being displayed is also persistent, so that on re-opening the pop-up _(which always displays page 1 of the refreshed list)_, the page number displayed is wrong - should be 1, of course.

**A very, VERY minor bug in a superb package!**

May I be a nuisance and - as it's related to the same part of SyncThing - mention a preference?   It would be good if the page selector could be at the top of the pop-up as well as it would save a lot of scrolling when 50 items are being displayed.

### Version Information

Syncthing Version: v1.3.4 (Linux 64-bit)
OS Version: Windows 7 (for SyncThing monitoring)
Browser Version: Firefox 74 (64-bit)

Yithis:
### Purpose

Add fields to save amount of elements per page and page number (fixes #6408).

### Testing

Local test if everything is compiling and does it solve issue (only for failed and changed items)

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -685,6 +685,8 @@ angular.module('syncthing.core')
         };
 
         $scope.refreshFailed = function (page, perpage) {
+            $scope.neededCurrentPage = page;
+            $scope.neededPageSize = perpage;
             if (!$scope.failed || !$scope.failed.folder) {
                 return;
             }
@@ -696,6 +698,8 @@ angular.module('syncthing.core')
         };
 
         $scope.refreshRemoteNeed = function (folder, page, perpage) {
+            $scope.neededCurrentPage = page;
+            $scope.neededPageSize = perpage;
             if (!$scope.remoteNeedDevice) {
                 return;
             }
@@ -711,6 +715,8 @@ angular.module('syncthing.core')
         };
 
         $scope.refreshLocalChanged = function (page, perpage) {
+            $scope.neededCurrentPage = page;
+            $scope.neededPageSize = perpage;
             if (!$scope.localChangedFolder) {
                 return;
             }
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -42,9 +42,10 @@ angular.module('syncthing.core')
         $scope.folderStats = {};
         $scope.progress = {};
         $scope.version = {};
-        $scope.needed = {};
-        $scope.neededCurrentPage = 1;
-        $scope.neededPageSize = 10;
+        $scope.needed = {
+            page: 1,
+            perpage: 10
+        };
         $scope.neededFolder = '';
         $scope.failed = {};
         $scope.localChanged = {};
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
So the issue is the fact that pagination widgets reuse the same numbers between different dialogs?
Is it not easier to to reset everything to initial state before showing the modal?

imsodin:
I didn't have time to check yet, but the change here is essentially a revert of https://github.com/syncthing/syncthing/pull/6452/files and I could reproduce #6408 before that PR (in fact that PR was an attempt at fixing it, that turned out to be ineffective). So I doubt this PR actually achieves what it hopes to, but I hope I am wrong (hopefully will check soon).

Yithis:
I think it is better to save settings at least for same widget from user experience's point of view. What do you think about creating 8 fields to save those settings?

Yithis:
> I didn't have time to check yet, but the change here is essentially a revert of https://github.com/syncthing/syncthing/pull/6452/files and I could reproduce #6408 before that PR (in fact that PR was an attempt at fixing it, that turned out to be ineffective). So I doubt this PR actually achieves what it hopes to, but I hope I am wrong (hopefully will check soon).

I think it is not a pure revert, since now I'm using those settings in a different way, but I can be wrong.

AudriusButkevicius:
Sorry, but adding 8 fields is mad. We should just reset the fields on dialog show and get on with our life. Also, I am not sure I still understand the bug we are fixing or why it happens.

imsodin:
> I think it is not a pure revert, since now I'm using those settings in a different way, but I can be wrong.

Sorry my statement came out way more dismissive than I intended. I did not yet have time to take this for a spin and look closer, I should reserve my comments for after then. A question though:

So the thing you probably changed is one of all those places you explicitly set the page number/size. Do you have any clue why this doesn't work if you explicitly set the object/dict members?

> Also, I am not sure I still understand the bug we are fixing or why it happens.

Neither do I, to me it's one of those inexplicable angular nightmares. Reproducing it is simple enough though (at least on my system, which doesn't mean much for JS...).

Yithis:
Oh sorry, it looks like I wrongly read your PR. First of all you missed a `;` at the end of dict declaration. But even when I add it, it looks like it is not properly defined.
Second of all you are calling all function starting with "show" with fixed parameters, for example `showNeed(1, 10)`. I think there should be values from dict. 

What do you think about such solution. I will add 5 values at the initialization of dictionary. 4 of them will be storing page number and the last one will be for the amount of items per page, only last value will be common for all widgets.

Yithis:
After more investigation I think I understand this issue much better. The problem was that dict `needed` was deleted after closing pop-up window. Solution I wrote in my last commit is solving issue for neededFilesModalView, but I think it is occuring in all widgets (I checked it and it is true for failedFilesModalView). After adding similar changes to those which I proposed (functions should be called with saved values, not fixed `(1, 10)`), this should be solved also for other views. But before creating more changes I wanted to know your opinion? Also there is a question where should those values be saved? I don't think that for example using values from dict `needed` for failed files is proper solution.

AudriusButkevicius:
We should definately not save the values. It makes no sense to go to page 3000 next time you open the dialog. There might not be any items in page 3000

st-review:
:robot: *beep boop*

PR #6553, opened 2020-04-20
lib/db: Upgrade to Blobloom v0.1.1

greatroar:
I was hoping you'd wait just an hour longer to merge #6537, because I just changed the Blobloom API (methods and Config fields renamed). Here's an update to v0.1.1, which is faster still and is now under the Apache 2.0 license instead of MPL.


greatroar:
I'll try not to do this again. The core API of Blobloom that Syncthing uses is now stable.

PR #6570, opened 2020-04-24
cmd/syncthing: Don't try to upgrade from unknown-dev

greatroar:
### Purpose

When syncthing is built without the build.go machinery, its version is unknown-dev. The autoupgrader considers that older than 1.x and replaces the binary. This patch fixes that behavior.

### Testing

Tested by building syncthing by a plain "go build" in cmd/syncthing, then running it with ./syncthing.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -196,7 +196,7 @@ func defaultRuntimeOptions() RuntimeOptions {
 	options := RuntimeOptions{
 		Options: syncthing.Options{
 			AssetDir:    os.Getenv("STGUIASSETS"),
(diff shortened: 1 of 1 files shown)


calmh:
It's not obvious to me under which circumstances this change is desirable. In the other direction, it  strikes me that if we ever made a mistake and released an incorrectly tagged version everyone would be stuck on that forever.

AudriusButkevicius:
Yeah, for local development, I just have STNOUPGRADE=1 in my bashrc

tomasz1986:
Regardless of the decision to implement this or not, there is a typo in the commit.

```
unknnown-dev
```

PR #6574, opened 2020-04-26
all: Reorder sequences for better rename detection

AudriusButkevicius:
Maybe works idk?


lib/model/folder.go
```diff
@@ -439,6 +440,7 @@ func (f *folder) scanSubdirs(subDirs []string) error {
 	}()
 
 	f.clearScanErrors(subDirs)
+	alreadyDeleted := make(map[string]struct{})
```
AudriusButkevicius, line 443: To be honest, I am not sure if this is even needed. I guess there is no harm in handling the delete twice?

imsodin: It's not necessary because the delete detection is done on a new snapshot, i.e. already with all the changes from the scan including the newly added deletes, thus no repeat handling.

calmh:
Reorder what for whatnow? :) Maybe a short summary of what this does?

AudriusButkevicius:
This creates a BlocksHash -> []filename mapping and upon discovery of a "new" file (non existent or previously deleted) looks up all other files according to the blockshash.

For each of the files, it checks if the file exists in the database, and does not exist on the filesystem.

If it finds one of these, compares blocks, sizes, etc, if that matches, it records the fact that the file is deleted immediately after the appearance of the new file (in terms of sequence), rather than in the "WithHave" walk that happens after we finish processing stuff coming back from the scanner.

This also changes the fileInfoBatch stuff not to count deleted files towards the "fullness" of the batch, so that we make sure that this create + delete falls into the same batch when dispatching.

AudriusButkevicius:
Further thinking about this, we could expose `UniqueToken() []byte` for fileinfos, on unix it could be inodes, on windows there is some thing simillar, and on other filesystems we could be BlocksHash.

Given this is internal to the device it does not matter what we use, but I guess inodes would give us more certainty.

lib/model/folder.go
```diff
@@ -450,6 +451,49 @@ func (f *folder) scanSubdirs(subDirs []string) error {
 
 		batch.append(res.File)
 		changes++
+
+		if res.Previous != nil && !res.Previous.IsDeleted() {
+			// Not a "new" file
```
imsodin, line 456: Why not check for deleted identical files for "non-new" files? I might just do `mv foo bar` where something already existed at `bar`.

AudriusButkevicius: I was thinking about it, and there is no good reason to have this check, but I was concerned that there is no way to distinguish between `mv foo bar` and `touch bar`, and I am not sure if having a extra db lookup and potentially a few stats would be of concern.

I guess there is no good way to catch cases like:
```
mv b a
mv c b
```

as check of B for A would return the fact that B exists, but B is a completely different file C,
and given leveldb does not support dirty reads, I don't think there is a way to fix this.

AudriusButkevicius:
Right, this is good for review as far as I am concerned.

calmh:
Sounds like a good change, now that I understand what it does.

lib/model/folder_sendrecv.go
```diff
@@ -1177,6 +1177,16 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
...
+	folderFilesystems := make(map[string]fs.Filesystem)
+	// Hope that it's usually in the same folder, so start with that one.
+	folders := []string{f.folderID}
+	for folder, cfg := range f.model.cfg.Folders() {
+		folderFilesystems[folder] = cfg.Filesystem()
+		if folder != f.folderID {
+			folders = append(folders, folder)
+		}
+	}
+
```
AudriusButkevicius, line 1189: This chunk not related, but we only need to do it once, so I moved it out of the loop.

lib/db/keyer.go
```diff
@@ -62,6 +62,9 @@ const (
 
 	// KeyTypeBlockList <block list hash> = BlockList
 	KeyTypeBlockList = 13
+
+	// KeyTypeBlockMapList <int32 folder ID> <block list hash> <file name> = <nothing>
+	KeyTypeBlockMapList = 14
```
imsodin, line 67: I'd have expected BlockListMap or BlockListHashMap (or both without Map).

lib/db/set.go
```diff
@@ -380,6 +382,27 @@ func (s *Snapshot) RemoteNeedFolderFiles(device protocol.DeviceID, page, perpage
...
+	}
+
+	for iter.Next() {
+		file := string(s.t.keyer.NameFromBlockMapListKey(iter.Key()))
+		if !iterator(file) {
+			break
+		}
+	}
+
+	iter.Release()
```
imsodin, line 402: I'd defer it. Not necessary here but is just safer e.g. in case an early return gets added later.

lib/model/folder.go
```diff
@@ -450,6 +451,51 @@ func (f *folder) scanSubdirs(subDirs []string) error {
...
+				return false
+			default:
+			}
+
+			cf, ok := snap.Get(protocol.LocalDeviceID, path)
+			if !ok {
+				return true
+			}
+			if !bytes.Equal(cf.BlocksHash, res.File.BlocksHash) {
+				l.Debugf("Mismatching block map list hashes: got %x expected %x", cf.BlocksHash, res.File.BlocksHash)
```
imsodin, line 467: That's should never happen though, right? I'd drop this check or move it to the db package and then correct the inconsistency (remove key) and log at a higher level.

lib/model/folder.go
```diff
@@ -450,6 +451,51 @@ func (f *folder) scanSubdirs(subDirs []string) error {
...
+				return true
+			}
+			if cf.IsDeleted() || cf.IsInvalid() || cf.IsIgnored() || cf.IsUnsupported() || cf.IsDirectory() || cf.IsSymlink() {
+				l.Debugf("Found something of unexpected type in block map list: %s", cf)
+				return true
+			}
+			if f.ignores.Match(cf.Name).IsIgnored() {
```
imsodin, line 474: This check isn't necessary, the scanner completely ignores ignored items (what a sentence), i.e. doesn't return them.

lib/model/model.go
```diff
@@ -2657,11 +2658,14 @@ func newFileInfoBatch(fn func([]protocol.FileInfo) error) *fileInfoBatch {
 
 func (b *fileInfoBatch) append(f protocol.FileInfo) {
 	b.infos = append(b.infos, f)
-	b.size += f.ProtoSize()
+	if !f.IsDeleted() {
```
imsodin, line 2661: This means no flushing on delete detection and thus huge batches when a lot is deleted at once. And I don't think it's necessary anyway: If you just don't call `flushIfFull` between adding the newly added file and checking/adding a corresponding deleted file you are fine too.

lib/db/keyer.go
AudriusButkevicius: So I am open to naming improvements.

The thing that has various hash -> something mappings is called BlockMap, now it  it can also map blocklist hashes to files, so perhaps BlockMapBlockList? 

😱

lib/model/folder.go
AudriusButkevicius: This potentially means we're somehow out of sync. 
I don't think I can control that easily in the db package, as I'd have to retrieve every key I find to verify this.

lib/model/folder.go
AudriusButkevicius: Sure, but this is a random filename that we got from the blocklist. It could have been ignored after it was added to it?

lib/model/model.go
AudriusButkevicius: The index sender uses this and it doesn't know when to stop.

Actually I don't think this actually does what I intended.

You could still add something that is not deleted, that drags it past the full threshold forces it to flush, and then the delete is the first thing in the next batch which could have been used to detect a rename.

We probably need to change the api to something like:
```
if batch.wouldFillBatch(fi) {
   batch.flush()
}
batch.add(fi)
```

then when full, if "next" item is not a delete, I say I would be full, and force that to the next batch.
if "next" item is a delete, I check that previous item is not a delete, if it's not a delete, I accept the last delete, if previous item is a delete, I say I am already full.

lib/model/folder.go
imsodin: Right of course, I messed up the `cf`s and `res`'.

lib/model/folder.go
imsodin: Right, actually now that you pointed it out: The first thing you do here is get the actual fileinfo - why don't you do that in the db code? Then you also don't need to introduce the new string iterator.

AudriusButkevicius: That's a fair suggestion, I could also deal with flag filtering and "fixing up" there.

lib/db/keyer.go
imsodin: Ups, yeah my names are wrong, these keys aren't directly related to `BlocksList`, just the hash of it. So I assume your `List` refers to the list of filenames? As the `BlockMap` is about individual blocks, I'd definitely use `BlocksHash` in the name to distinguish. Maybe `KeyTypeBlocksHashFiles`?

lib/model/folder.go
AudriusButkevicius: Actually, this is all in a read-only transaction, so I don't think I can do any fixing up there.

lib/db/keyer.go
AudriusButkevicius: I think it's better now.

lib/db/set.go
AudriusButkevicius: Addressed.

lib/model/model.go
AudriusButkevicius: I think I addressed this neatly.

lib/model/model_test.go
```diff
@@ -3554,3 +3555,166 @@ func TestFolderAPIErrors(t *testing.T) {
...
+
+	m.ScanFolders()
+
+	// Reach info the model, grab the fileset.
+
+	m.fmut.RLock()
+	fset := m.folderFiles["default"]
+	m.fmut.RUnlock()
+
+	snap := fset.Snapshot()
```
imsodin, line 3655:
```suggestion
	count := 0
	snap := dbSnapshot(t, m, "default")
```

lib/model/model_test.go
```diff
@@ -3554,3 +3555,166 @@ func TestFolderAPIErrors(t *testing.T) {
...
+	must(t, ffs.Rename("three", "new-three"))
+
+	// Change type
+	must(t, ffs.Remove("four"))
+	must(t, ffs.Mkdir("four", 0644))
+
+	m.ScanFolders()
+
+	// Check we're left with 2 of the 5
+	snap = fset.Snapshot()
```
imsodin, line 3692:
```suggestion
	snap = dbSnapshot(t, m, "default")
```

lib/model/model_test.go
```diff
@@ -3554,3 +3555,166 @@ func TestFolderAPIErrors(t *testing.T) {
...
+	m.fmut.RLock()
+	fset := m.folderFiles["default"]
+	m.fmut.RUnlock()
+
+	count := 0
+	snap := fset.Snapshot()
+	snap.WithHave(protocol.LocalDeviceID, func(i db.FileIntf) bool {
+		count++
+		return true
+	})
```
imsodin, line 3585:
```suggestion
	count := 0
	snap := dbSnapshot(t, m, "default")
```

lib/model/model_test.go
```diff
@@ -3554,3 +3555,166 @@ func TestFolderAPIErrors(t *testing.T) {
...
+	}
+	// Rename
+	must(t, ffs.Rename("3", "17"))
+	must(t, ffs.Rename("16", "4"))
+
+	// Scan
+	m.ScanFolders()
+
+	// Verify sequence of a appearing is followed by c disappearing.
+	snap = fset.Snapshot()
```
imsodin, line 3608:
```suggestion
	snap = dbSnapshot(t, m, "default")
```

lib/db/set.go
```diff
@@ -380,6 +381,44 @@ func (s *Snapshot) RemoteNeedFolderFiles(device protocol.DeviceID, page, perpage
 	return files
 }
 
+func (s *Snapshot) WithBlocksHash(hash []byte, iterator Iterator) error {
+	key, err := s.t.keyer.GenerateBlockListMapKey(nil, []byte(s.folder), hash, nil)
```
imsodin, line 385: It would be more consistent with all the other iterators if the implementation lived in `transactions.go` and has the usual ```Debugf(`%s WithBlocksHash ...``` logging here.

lib/db/lowlevel.go
```diff
@@ -331,6 +350,14 @@ func (db *Lowlevel) dropFolder(folder []byte) error {
 		return err
 	}
 
+	k5, err := db.keyer.GenerateBlockListMapKey(nil, folder, nil, nil)
```
imsodin, line 353: Is there a reason to not reuse byte slices in this function (except that it's already like that now)? Not asking to have it changed, just curious because I don't see it and if there is a reason, I'd like to know in case I ever touch this :)

lib/db/set.go
```diff
@@ -380,6 +381,44 @@ func (s *Snapshot) RemoteNeedFolderFiles(device protocol.DeviceID, page, perpage
...
+			continue
+		}
+
+		if !bytes.Equal(fi.BlocksHash, hash) {
+			// Perhaps should be at least info?
+			l.Debugf("Mismatching block map list hashes: got %x expected %x", fi.BlocksHash, hash)
+			continue
+		}
+
+		if fi.IsDeleted() || fi.IsInvalid() || fi.IsIgnored() || fi.IsUnsupported() || fi.IsDirectory() || fi.IsSymlink() {
```
imsodin, line 409: `IsIgnored` and `IsUnsupported` are already covered by `IsInvalid`.

lib/db/set.go
```diff
@@ -380,6 +381,44 @@ func (s *Snapshot) RemoteNeedFolderFiles(device protocol.DeviceID, page, perpage
...
+
+	for iter.Next() {
+		file := string(s.t.keyer.NameFromBlockListMapKey(iter.Key()))
+		fi, ok := s.Get(protocol.LocalDeviceID, file)
+		if !ok {
+			continue
+		}
+
+		if !bytes.Equal(fi.BlocksHash, hash) {
+			// Perhaps should be at least info?
```
imsodin, line 404: I'd say either we are confident this doesn't happen, then leave it at debug (or don't even check), or make it a warning to get reports asap (and maybe even create a repair mechanism like with sequences, but 🤮).

lib/db/set.go
AudriusButkevicius: There is nothing of the snapshot package in that file:

```
/cygdrive/c/Gohome/src/github.com/syncthing/syncthing (renames) $ cat lib/db/transactions.go | grep Snapshot | wc -l
0
```

Seems like not the right place to land that?

AudriusButkevicius: I guess you suggest to split it out. Right, let me do that.

lib/db/lowlevel.go
AudriusButkevicius: it's not something that is done often, so I guess it does not matter, but I've changed it never the less.

lib/db/transactions.go
```diff
@@ -290,6 +291,49 @@ func (t *readOnlyTransaction) withGlobal(folder, prefix []byte, truncate bool, f
...
+	}
+	defer iter.Release()
+
+	for iter.Next() {
+		file := string(t.keyer.NameFromBlockListMapKey(iter.Key()))
+		f, ok, err := t.getFile(folder, protocol.LocalDeviceID[:], []byte(osutil.NormalizedFilename(file)))
+		if backend.IsClosed(err) {
+			return err
+		} else if err != nil {
+			panic(err)
```
imsodin, line 312:
```suggestion
```

Errors are already filtered in the calling func in set.go.

AudriusButkevicius:
I guess this might need a migration to populate stuff?

lib/db/schemaupdater.go
```diff
@@ -484,3 +486,41 @@ func (db *schemaUpdater) updateSchemato9(prev int) error {
 
 	return t.Commit()
 }
+
+func (db *schemaUpdater) updateSchemaTo10(prev int) error {
```
imsodin, line 490:
```suggestion
func (db *schemaUpdater) updateSchemaTo10(_ int) error {
```

lib/db/schemaupdater.go
```diff
@@ -484,3 +486,41 @@ func (db *schemaUpdater) updateSchemato9(prev int) error {
...
+				return true
+			}
+
+			name := []byte(f.FileName())
+			dk, putErr = db.keyer.GenerateBlockListMapKey(dk, folder, f.BlocksHash, name)
+			if putErr != nil {
+				return false
+			}
+
+			putErr = t.Put(dk, nil)
```
imsodin, line 515: Checkpointing after this.

lib/model/folder_sendrecv.go
```diff
@@ -454,7 +454,7 @@ nextFile:
 
 		// Check our list of files to be removed for a match, in which case
 		// we can just do a rename instead.
-		key := string(fi.Blocks[0].Hash)
+		key := string(fi.BlocksHash)
 		for i, candidate := range buckets[key] {
 			if candidate.BlocksEqual(fi) {
```
calmh, line 459: This `if` just checks the same thing we used as the key (the `BlocksHash`) so we could drop it

lib/model/folder.go
```diff
@@ -450,6 +450,41 @@ func (f *folder) scanSubdirs(subDirs []string) error {
...
+			}
+
+			nf := fi
+			nf.SetDeleted(f.shortID)
+			nf.LocalFlags = f.localFlags
+			if fi.ShouldConflict() {
+				// We do not want to override the global version with
+				// the deleted file. Setting to an empty version makes
+				// sure the file gets in sync on the following pull.
+				nf.Version = protocol.Vector{}
```
calmh, line 482: What's going to happen here is that the other side will reject our delete as whatever it has is already better... So we could skip this stuff and just `if.ShouldConflict() { /* fuckit */ return true }` somewhere above?

lib/model/folder_sendrecv.go
AudriusButkevicius: Even when trying to match a delete, we still check block by block. I don't think there is that much harm in checking this.

lib/model/folder.go
AudriusButkevicius: So this is just a copy of the code below, I guess if you think this should be changed this should be a separate PR?

calmh: The purposes are different. In the existing case we send out the null-versioned delete to indicate that 1) we don't have the file 2) our info is out of date and should be ignored. The effect will be that we immediately pull the file... Here, we're looking to optimise the order of index updates, but this index update is pointless. I dunno. I'm not really sure when this would happen, the renamed-from file needs to have the Unsupported, Ignored or ReceiveOnlyChanged flags... I guess if we uningnore a file at precisely the same time we also rename it?

I guess the comment could say something about this being weird and unlikely but just matches the behavior of somewhere else...

lib/model/folder_sendrecv.go
calmh: BlocksEqual just checks the hash if there is one, but yeah it doesn't do any harm as is.

lib/model/folder.go
imsodin: Good point: Essentially this rename detection doesn't have any effect on a receive-only folders, i.e. this entire new block could be skipped `if f.localFlags == FlagLocalReceiveOnly`. Having the `ShouldConflict` check just for safety anyway seems sensible. And similar to the ignores check etc. above I'd also not handle it here but just `return true`.

lib/model/folder_sendrecv.go
AudriusButkevicius: Ah yes! I thought it still compared the blocks, I'll fix that.

lib/model/folder.go
AudriusButkevicius: Can someone give me pseudo-code what this should look like?

It feels like I'm in the united nations headquarters parking lot with all the flags around me.

imsodin: I was thinking of something like this (with the new function not being a required change, just makes it easiert o take about):

Move this entire block into a function, lets say `detectRenames`. Then have

```
	if f.localFlags != FlagLocalReceiveOnly {
		f.detectRenames(allTheArguments)
	}
```

In that function drop your current `ShouldConflict` condition, and add a new one between `ignores.Match` and `BlocksEqual`:

```
f fi.ShouldConflict() {
     // This should not happen, but even if it does: Conflicting changes are sent with a zero version vector, i.e. won't ever be needed/synced by remotes thus can't participate in renames.
    return true
}
```

AudriusButkevicius: What's the point of the return value if its not used when using detectRenames?

imsodin: Pseudo-diff (fake indentation) without the `detectRenames` which is just niceness (and maybe caused the confusion), hopefully it makes clear what I mean:

```diff
--- lib/model/folder.go	2020-05-11 14:39:42.028608307 +0200
+++ fordiff.txt	2020-05-11 14:39:11.701112734 +0200
@@ -451,6 +451,7 @@
 		batch.append(res.File)
 		changes++
 
+		if f.localFlags != protocol.FlagLocalReceiveOnly {
 		snap.WithBlocksHash(res.File.BlocksHash, func(ifi db.FileIntf) bool {
 			fi := ifi.(protocol.FileInfo)
 
@@ -460,6 +461,10 @@
 			default:
 			}
 
+			if fi.ShouldConflict() {
+				return true
+			}
+
 			if f.ignores.Match(fi.Name).IsIgnored() {
 				return true
 			}
@@ -475,16 +480,11 @@
 			nf := fi
 			nf.SetDeleted(f.shortID)
 			nf.LocalFlags = f.localFlags
-			if fi.ShouldConflict() {
-				// We do not want to override the global version with
-				// the deleted file. Setting to an empty version makes
-				// sure the file gets in sync on the following pull.
-				nf.Version = protocol.Vector{}
-			}
 			batch.append(nf)
 			changes++
 			return false
 		})
+		}
 	}
 
 	if err := batch.flush(); err != nil {
```

PR #6587, opened 2020-04-28
lib/model: Add support for different puller block ordering

AudriusButkevicius:
Change the default to use "standard" ordering which effectively:
1. Works our which devices "need" the file, where need means they don't have it and we share the folder with them.
2. Works out which devices can get the file directly (not through us), which means they bi-directionally share (as per their ClusterConfig) the folder with someone who "has" the file as per snap.Available.
3. Sort all of these devices that can get the file directly (including us)
4. Work out our index in that list list of devices
5. Slice the blocks into len(devices) chunks.
6. Put the chunk at our index at the front, and shuffle the rest of the chunks.

This should improve the initial spread of the file, and also ease up the IO, as we fetch stuff in continous chunks, producing linear IO.


imsodin:
I hope we have relevant benchmarks, I want to run them on some old spinner to witness the awesomeness :D

Regarding the config: I see the potential value in niches but question the weighting of that versus having to maintain, document and mostly support such a thing. Users do find knobs and eventually (mis)use them for anything regardless of whether it's related at all or not. Then again i do want syncthing to cater to "powerusers". Maybe as a middleground we could move all the stuff we'd rather not be touched (i.e. what's only in the advanced settings in gui) into a subgroup with a scary sounding name ("advanced" is too mild, "experimental"maybe though not really accurate).

I'll have a look at the code tomorrow once in front of a pc again.

AudriusButkevicius:
I don't think there is anything to benchmark here, as I don't expect things to be faster, I simply expect less brutal io on spinning drives.

I agree that the config is growing big and potentially we should split it into stuff people are expected to tweak and people should mostly leave alone, but I think the UI sort of does that for us.

If anything, I don't think this sort of change should be in this PR.

imsodin:
> If anything, I don't think this sort of change should be in this PR.

Totally agree.

I didn't mean to say this requires benchs, i am convinced it does help. i still hope I'll find the time to create a bench that can show the speedup, just because

imsodin:
For the linear io: We still do individual requests and writes for blocks, not chunks of blocks. Due to parallel block pulls, there's no guarantee that this happens in order after all. Do you expect the fs to be able to profit from writes to contiguous blocks happening closer together time-wise, but not necessarily in order? Otherwise we'd probably also have to reorganise copiers and pullers to handle chunks of blocks for this to have any effect (request luckily already allow for arbitrary data chunks not bound by blocks).

Generally reviewing this there's quite a few maps, slices and loops to follow. Things are nicely structured and named, so it's very doable to get reasonable quick what is actually going on, but I am wondering whether a simpler solution couldn't achieve the same goals. As far as I understand most of the complexity is to ensure all participating devices which can get the file directly start with a different block range - correct? For linear io wouldn't it suffice to just use a fixed desired chunk length (or len(blocks)/len(allDevices) where that's bigger). To ensure different starting points it should suffice to just take all known devices including us unfiltered, and take our position in the ordered list as our starting point with the fixed chunk length. If the file is too small, i.e. there's not enough blocks, just start chunking at block number `ourDevicePosition * len(blocks in file) / len(devices)`. The only drawback to your solution I see is that you get the longest possible contiguous chunks, while with mine you get "long enough" (where I admittedly don't know what "long enough" is, maybe `MaxBlockSize` (16MiB)).

AudriusButkevicius:
I think the important bit here is for spinning disks not having to move the spindle too far.
Most OS'es have write caching, so even if we pull blocks relatively out of order, they are still nearby and the OS will reorder the writes when flushing stuff to the underlying disk.
The current implementation randomly shuffles all the blocks, so you do not benefit much from caching, and end up thrashing the drives most of the time.

If I understand your solution it's similar to mine, but minus the knowing who is connected and who has what?

imsodin:
> If I understand your solution it's similar to mine, but minus the knowing who is connected and who has what?

Exactly, trying to cut down on complexity/length and folder-model-interactions.

AudriusButkevicius:
So you split the file into numDevicesIamSharingTheFolder with, pick chunk based on my id ordering, and shuffle the rest?

Sure, that works, and is less code too, if that is what we want

imsodin:
If there's no other benefit to more code, I'd always choose less. However I might just be missing (the relevance of) the advantage of your solution.

AudriusButkevicius:
There probably isn't much. I guess we could run simulations and compare 🧑‍🔬

lib/model/blockpullreorderer.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+type randomOrderBlockPullReorderer struct{}
+
+func (randomOrderBlockPullReorderer) Reorder(blocks []protocol.BlockInfo) []protocol.BlockInfo {
+	rand.Shuffle(blocks)
+	return blocks
+}
+
+func newBlockPullReorderer(order config.BlockPullOrder, id protocol.DeviceID, otherDevices []protocol.DeviceID) blockPullReorderer {
```
imsodin, line 33: I'd move that to directly below the `blockPullReorder` interface.

AudriusButkevicius: Move what exactly?

imsodin: This entire function. Now it's interface declaration, then the first two interface implementations, then the interface constructor, then another implementation.

AudriusButkevicius: Still not sure I get it.

We have a factory method that constructs one of the implementations that satisfy the interface.
One of the implementations is slightly more complicated and has it's own constructor.

You are saying you'd fold the implementations constructor into the factory method?

imsodin: Instead of 

```
type blockPullReorderer interface {
	Reorder(blocks []protocol.BlockInfo) []protocol.BlockInfo
}

type inOrderBlockPullReorderer struct{}

func (inOrderBlockPullReorderer) Reorder(blocks []protocol.BlockInfo) []protocol.BlockInfo {
	return blocks
}

type randomOrderBlockPullReorderer struct{}

func (randomOrderBlockPullReorderer) Reorder(blocks []protocol.BlockInfo) []protocol.BlockInfo {
	rand.Shuffle(blocks)
	return blocks
}

func newBlockPullReorderer(order config.BlockPullOrder, id protocol.DeviceID, otherDevices []protocol.DeviceID) blockPullReorderer {
...
}

 func newStandardBlockPullReorderer(id protocol.DeviceID, otherDevices []protocol.DeviceID) *standardBlockPullReorderer {
...
}


type standardBlockPullReorderer struct {
	myIndex int
	count   int
	shuffle func(interface{}) // Used for test
}
```

I'd order it like

```
type blockPullReorderer interface {
	Reorder(blocks []protocol.BlockInfo) []protocol.BlockInfo
}
func newBlockPullReorderer(order config.BlockPullOrder, id protocol.DeviceID, otherDevices []protocol.DeviceID) blockPullReorderer {
...
}

type inOrderBlockPullReorderer struct{}

func (inOrderBlockPullReorderer) Reorder(blocks []protocol.BlockInfo) []protocol.BlockInfo {
	return blocks
}

type randomOrderBlockPullReorderer struct{}

func (randomOrderBlockPullReorderer) Reorder(blocks []protocol.BlockInfo) []protocol.BlockInfo {
	rand.Shuffle(blocks)
	return blocks
}


 func newStandardBlockPullReorderer(id protocol.DeviceID, otherDevices []protocol.DeviceID) *standardBlockPullReorderer {
...
}


type standardBlockPullReorderer struct {
	myIndex int
	count   int
	shuffle func(interface{}) // Used for test
}
```

AudriusButkevicius: Ah right, this is function ordering.

AudriusButkevicius:
@calmh can revert if he's not happy, merging just to make it to the last train.

calmh:
It’s fine. I’d already tagged 1.6.0-rc.1 but I can move the tag.

calmh:
It’s not the last train though. There are regular trains all the time. :)

calmh:
It would be great with some docs on what these options are too

PR #6588, opened 2020-04-29
lib/model: Add option to disable fsync

AudriusButkevicius:
I guess it's somewhat against the spirit, but I don't think there is any harm in making it configurable.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -1581,17 +1581,17 @@ func (f *sendReceiveFolder) dbUpdaterRoutine(dbUpdateChan <-chan dbUpdateJob) {
 		// sync directories
 		for dir := range changedDirs {
 			delete(changedDirs, dir)
-			fd, err := f.fs.Open(dir)
-			if err != nil {
-				l.Debugf("fsync %q failed: %v", dir, err)
-				continue
-			}
 			if !f.FolderConfiguration.DisableFsync {
+				fd, err := f.fs.Open(dir)
+				if err != nil {
+					l.Debugf("fsync %q failed: %v", dir, err)
+					continue
+				}
 				if err := fd.Sync(); err != nil {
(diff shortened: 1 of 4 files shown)


tomasz1986:
If possible, please add at least basic information to the Docs on this. Disabling fsync by inexperienced (yet curious) users can easily lead to massive data loss, especially on battery-powered devices.

imsodin:
Was there any discussion on this i missed? if not what's the motivation behind the change? this would be something I'd definitely like to hide better

AudriusButkevicius:
There were prior complaints about slowness when syncthing small files, where the files were cache files that can be regenerated, but take some time to produce. For that sort of scenario running without fsync might be perfectly acceptable.

Effectively I think it's a useful shotgun to have in the tool box.

lib/model/folder_sendrecv.go
```diff
@@ -1586,8 +1586,10 @@ func (f *sendReceiveFolder) dbUpdaterRoutine(dbUpdateChan <-chan dbUpdateJob) {
 				l.Debugf("fsync %q failed: %v", dir, err)
 				continue
 			}
-			if err := fd.Sync(); err != nil {
-				l.Debugf("fsync %q failed: %v", dir, err)
+			if !f.FolderConfiguration.DisableFsync {
```
calmh, line 1589: The entire flushfn shoudl go away if we're not syncing, and ideally we don't need to track changedDir either (but the cost isn't enormous so it might not be worth sprinkling yet more ifs to accomplish it, I didn't check)

Catfriend1:
I think being able to disable fsync might be useful in some cases. I've also had that in the past where I asked me why syncing was slow on spinning drives with a lot of small files involved. Wished I could set it on my receive only's back then.

Catfriend1:
Should we tag this with "api"? I'm asking because I always put new config fields in the config model of the android wrapper so if the user changes the setting, e.g. through the Web UI, it is persisted across app restarts.

PR #6589, opened 2020-04-29
lib/model: Progress emitter network operations dont need to be blocking

AudriusButkevicius:
If the network is slow, this could block while holding a lock, and block other components that use progress emitter (Completion for the UI for example)

### DIFF
--- a/lib/model/progressemitter.go
+++ b/lib/model/progressemitter.go
@@ -149,7 +149,7 @@ func (t *ProgressEmitter) sendDownloadProgressMessagesLocked(ctx context.Context
 			updates := state.update(folder, activePullers)
 
 			if len(updates) > 0 {
-				conn.DownloadProgress(ctx, folder, updates)
+				go conn.DownloadProgress(ctx, folder, updates)
 			}
 		}
 	}
--- a/lib/model/progressemitter.go
+++ b/lib/model/progressemitter.go
@@ -37,6 +37,16 @@ type ProgressEmitter struct {
 	timer *time.Timer
 }
 
+type progressUpdate struct {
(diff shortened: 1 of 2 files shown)


lib/model/progressemitter.go
```diff
@@ -149,7 +149,7 @@ func (t *ProgressEmitter) sendDownloadProgressMessagesLocked(ctx context.Context
 			updates := state.update(folder, activePullers)
 
 			if len(updates) > 0 {
-				conn.DownloadProgress(ctx, folder, updates)
+				go conn.DownloadProgress(ctx, folder, updates)
```
calmh, line 152: capturing iteration variable in `folder`.

AudriusButkevicius:
What about if we dispatch them into their own routines, but for non-delete things we wedge in a 5s timeout. I guess if we're struggling to dispatch a few integers so badly, we might as well drop it on the floor.

We can't drop the deletes on the floor as we'd infinitely show those blocks as available, but we don't really care about the order of those?

AudriusButkevicius:
I mean or we could just hang without the lock, this would stop the progress emitter from functioning, but that's probably fine?

calmh:
I mostly like the "queue, but forget after a while" thing...

AudriusButkevicius:
So I think the current solution is fairly simple and elegant.

calmh:
Right, so this just blocks the progress update sending, while allowing it to otherwise accumulate more changes and respond to whatever needs responding to? LGTM if I didn't misunderstand

AudriusButkevicius:
Well this stops reporting stuff if sends are blocking. Effectively turning it off...

PR #6590, opened 2020-04-29
lib/connections: React to listeners going up and down faster

AudriusButkevicius:
So the first problem is that there is a race between the listeners starting and the global announce starting.

It seems global announce tries to send a notification straight away, at which point none of the listeners might be up yet, so it sends nothing, and backs off for 5 mins.

Second problem is that listeners starting/stopping do not actually notify of the addresses changing, which means the announcer does not know it needs to do something and, and we end up waiting the whole announceErrorRetryInterval (5mins) after startup before sending the first real announcement.

### DIFF
--- a/lib/connections/structs.go
+++ b/lib/connections/structs.go
@@ -174,6 +174,12 @@ type listenerFactory interface {
 	Valid(config.Configuration) error
 }
 
+type ListenerAddresses struct {
+	URI          *url.URL
+	WANAddresses []*url.URL
+	LANAddresses []*url.URL
+}
+
 type genericListener interface {
 	Serve()
 	Stop()
@@ -188,7 +194,7 @@ type genericListener interface {
 	WANAddresses() []*url.URL
 	LANAddresses() []*url.URL
 	Error() error
-	OnAddressesChanged(func(genericListener))
+	OnAddressesChanged(func(ListenerAddresses))
 	String() string
 	Factory() listenerFactory
 	NATType() string
@@ -203,14 +209,28 @@ type Model interface {
 }
 
 type onAddressesChangedNotifier struct {
-	callbacks []func(genericListener)
+	callbacks []func(ListenerAddresses)
 }
 
-func (o *onAddressesChangedNotifier) OnAddressesChanged(callback func(genericListener)) {
+func (o *onAddressesChangedNotifier) OnAddressesChanged(callback func(ListenerAddresses)) {
 	o.callbacks = append(o.callbacks, callback)
 }
 
 func (o *onAddressesChangedNotifier) notifyAddressesChanged(l genericListener) {
+	o.notifyAddresses(ListenerAddresses{
+		URI:          l.URI(),
+		WANAddresses: l.WANAddresses(),
+		LANAddresses: l.LANAddresses(),
+	})
+}
+
+func (o *onAddressesChangedNotifier) clearAddresses(uri *url.URL) {
+	o.notifyAddresses(ListenerAddresses{
+		URI: uri,
+	})
+}
+
+func (o *onAddressesChangedNotifier) notifyAddresses(l ListenerAddresses) {
 	for _, callback := range o.callbacks {
 		callback(l)
 	}
--- a/lib/connections/structs.go
+++ b/lib/connections/structs.go
@@ -224,9 +224,9 @@ func (o *onAddressesChangedNotifier) notifyAddressesChanged(l genericListener) {
 	})
 }
 
-func (o *onAddressesChangedNotifier) clearAddresses(uri *url.URL) {
+func (o *onAddressesChangedNotifier) clearAddresses(l genericListener) {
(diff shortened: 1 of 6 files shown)


imsodin:
as you have now added a way to trigger an announcement on change/start, is the initial timer based trigger even necessary anymore?

Plus is there any delay/aggregation? otherwise announce on change may increase discovery load significantly, right?

AudriusButkevicius:
It internally debounces the announcements by 2s for things changing rapidly, presumably for cases like startup.

But thinking about this further, flopping of the listener could cause it DoS discovery.

AudriusButkevicius:
Actually I am not sure how this should be handled anymore.

Flip-flopping nat mapping/stun thing would cause announcements to be sent ignoring all advice Retry-After and what not, but a flip-flopping listener would not because it does not emit any events.

Perhaps we increase the startup delay to like 15s (because I think it takes 10s for us to decide which relay we are on) and call it a day, and hope that everything started in time?

Alternative would be to keep this code, but perhaps bump the debounce time to like 30, so that if there is a listener flip flopping, hopefully it would do that sub-30s, and effectively be permanently debounced and not send anything to the discovery server?

@calmh?

calmh:
Didn't yet read the code with the intent to understand, but my feeling would be that that announcer should react and track instantaneously changes to the listeners, yet have a debounce interval so that it doesn't announce a change until it's been stable for x seconds (or, perhaps, been y seconds since last announce, where y >> x)...

AudriusButkevicius:
So I guess it' sort of reflects what you expect, but please check the code and let me know if something is off.

imsodin:
Changes lookg good.

Regarding announcement timing (not changed by this PR): I am not sure what the timer currently does, as we reset the timer without checking if it's expired already. Which the go docs say you shouldn't do. If that's only because of the return value being inherently racy and it just restarts the timer at 2s regardless, that racyness is not an issue as we don't care about the return value. However it means we defer announcements indefinitely if something is flip-flopping in under 2s (which shouldn't happen, but...) - I'd say there should also be an upper limit after which an announcement is sent no matter what.

```
	for {
		select {
		case <-eventSub.C():
			// Defer announcement by 2 seconds, essentially debouncing
			// if we have a stream of events incoming in quick succession.
			timer.Reset(2 * time.Second)
		case <-timer.C:
			c.sendAnnouncement(ctx, timer)
		case <-ctx.Done():
			return
		}
	}
```

AudriusButkevicius:
Added a simple counter to track that.

lib/discover/global.go
```diff
@@ -197,20 +197,33 @@ func (c *globalClient) serve(ctx context.Context) {
...
+				// if we have a stream of events incoming in quick succession.
+				timer.Reset(2 * time.Second)
+			} else if timerResetCount == 10 {
+				// Yet only do it if we haven't had to reset 10 times in a row, so if
+				// something is flip-flopping within 2 seconds, we don't end up in a permanent reset loop.
+				l.Warnf("Detected a flip-flopping listener")
+				c.setError(errors.New("flip flopping listener"))
+				// Incrementing the count above 10 will prevent us from warning or setting the error again
+				// It will also suppress event based resets until we've had a proper round after announceErrorRetryInterval
+				timer.Reset(announceErrorRetryInterval)
```
imsodin, line 222: I'd still do a single announce before this. E.g. if flipflopping occurs right from the start, legitimate announcements don't make it out until the whole error retry interval has passed.

AudriusButkevicius: Yeah, that's 5 minutes, it's not a tragedy if we're flip-flopping.

lib/connections/structs.go
```diff
@@ -203,14 +209,28 @@ type Model interface {
...
 
 func (o *onAddressesChangedNotifier) notifyAddressesChanged(l genericListener) {
+	o.notifyAddresses(ListenerAddresses{
+		URI:          l.URI(),
+		WANAddresses: l.WANAddresses(),
+		LANAddresses: l.LANAddresses(),
+	})
+}
+
+func (o *onAddressesChangedNotifier) clearAddresses(uri *url.URL) {
```
calmh, line 227: I'd have this one take a `genericListener` as well and call `URI()` on it just for symmetry with `notifyAddressesChanged` and making sure we clear the same URI we've announced

lib/discover/global.go
```diff
@@ -197,20 +197,33 @@ func (c *globalClient) serve(ctx context.Context) {
...
+	timerResetCount := 0
+
 	for {
 		select {
 		case <-eventSub.C():
-			// Defer announcement by 2 seconds, essentially debouncing
-			// if we have a stream of events incoming in quick succession.
-			timer.Reset(2 * time.Second)
-
+			if timerResetCount < 10 {
```
calmh, line 211: Take the opportunity to declare constants for the magic numbers and times?

calmh: Also the same instance of `10` further down and the actual delays and stuff?

PR #6591, opened 2020-04-29
lib/connections: Update quic

AudriusButkevicius:
I have not checked what exactly landed in the package, but it must be better, right?


AudriusButkevicius:
I don't think that freebsd issue is caused by me. I am a bit at a loss of wtf is happenin.

imsodin:
we can't upgrade without forking as they use go1.14 novelties ( inconsequential ones: interface compositing). i am considering asking for "stable" releases upstream and if getting negative response, creating quic-go-stable.

AudriusButkevicius:
Aren't we on 1.14 anyway?

calmh:
We aren't, because the initial 1.14 release was a bit rocky and now is not the time to switch (with an RC out).

But regardless, our policy is that it should work to build syncthing at latest-1, so Go 1.13 currently.

AudriusButkevicius:
Given there is an explicit 1.14 builder, I believe the rest of the build infra is still on 1.13?
If that is true, then the build passed all the checks, and I am not sure why we can't upgrade?

imsodin:
We can indeed, they recently changed their minds, jipiiii: [make quic-go build with both Go 1.13 and 1.14 ](https://github.com/lucas-clemente/quic-go/commit/3ac2fb3161d0b4bdd304771d55972fdc5b9d2f19).

lib/connections/quic_dial.go
```diff
@@ -80,7 +80,7 @@ func (d *quicDialer) Dial(ctx context.Context, _ protocol.DeviceID, uri *url.URL
 	stream, err := session.OpenStreamSync(ctx)
 	if err != nil {
 		// It's ok to close these, this does not close the underlying packetConn.
-		_ = session.Close()
+		_ = session.CloseWithError(0, err.Error())
```
imsodin, line 83: Probably doesn't matter, but I'd expect `1` for the generic BEP error (or just any generic error)? Same below.

PR #6594, opened 2020-05-02
lib/model: Merge add and start folder funcs and related refactor

imsodin:
Folders are never added and started separately except in tests and the "unlocked" functions aren't used at all except in tests, thus I merged the two into a single function. And I renamed `removeFolderLocked` to `deleteFolderEntriesLocked`, as it isn't the locked equivalent to `removeFolder`. These changes do not change functionality, they just make things more easier to understand and thus more robust (currently one might wonder about the relevance of the distinction of adding and starting a folder, when there isn't any).

For tests that need to inject a custom ignore option and previously did this by adding a folder, then changing ignores and the starting the folder, I just added a variant of `addAndStartFolder` where ignores are passed as an argument.

For send-receive tests that rely on the folder service not actually running and manually constructed a `sendReceiveFolder` object, I changed that to start a model with the desired folder as usual, and then stop the suture services and do any additional changes to the folder object necessary to get tests running. This keeps the setup/settings closer to what is actually happening in production.

And it trims a few lines of code and adds/fixes some test timeouts (which I noticed because I forgot to to reset the folder ctx at first on the send-receive tests, which caused everything to hang).

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -275,22 +275,22 @@ func (m *model) StartDeadlockDetector(timeout time.Duration) {
 	detector.Watch("pmut", m.pmut)
 }
 
-// startFolder constructs the folder service and starts it.
-func (m *model) startFolder(folder string) {
-	m.fmut.RLock()
-	folderCfg := m.folderCfgs[folder]
-	m.fmut.RUnlock()
+// Need to hold lock on m.fmut when calling this.
+func (m *model) addAndStartFolderLocked(cfg config.FolderConfiguration, fset *db.FileSet) {
+	ignores := ignore.New(cfg.Filesystem(), ignore.WithCache(m.cacheIgnoredFiles))
+	if err := ignores.Load(".stignore"); err != nil && !fs.IsNotExist(err) {
+		l.Warnln("Loading ignores:", err)
+	}
 
-	// Close connections to affected devices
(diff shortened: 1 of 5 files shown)


lib/model/model.go
```diff
@@ -474,7 +441,7 @@ func (m *model) stopFolder(cfg config.FolderConfiguration, err error) {
 }
 
 // Need to hold lock on m.fmut when calling this.
-func (m *model) removeFolderLocked(cfg config.FolderConfiguration) {
+func (m *model) deleteFolderEntriesLocked(cfg config.FolderConfiguration) {
```
AudriusButkevicius, line 444: Why the rename?

imsodin: > And I renamed `removeFolderLocked` to `deleteFolderEntriesLocked`, as it isn't the locked equivalent to `removeFolder`.

I consider the name misleading, given `removeFolder` is meant to actually completely remove the folder (as in it's no longer in config and we drop all we know from db), while this is used e.g. when pausing a folder as well.

AudriusButkevicius: Stop? Cleanup? Shutdown?

imsodin: `cleanupFolderLocked` should work. Stop is taken and Shutdown not really accurate: It just removes the entries for said folder in all of the maps protected by fmut. That's why I thought a name representing that would be good. So maybe `deleteFolderMapEntriesLocked`? I am fine with either.

PR #6598, opened 2020-05-04
Handle ignores

Shidory:
Purpose being to soft migrate to `$markerName/ignores.txt` when operations are made in the GUI, while retaining compatibility with existing setups.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -12,6 +12,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"os"
 	"path/filepath"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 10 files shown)


lib/model/model.go
```diff
@@ -1725,10 +1736,41 @@ func (m *model) SetIgnores(folder string, content []string) error {
 		return err
 	}
 
-	if err := ignore.WriteIgnores(cfg.Filesystem(), ".stignore", content); err != nil {
```
calmh, line 1728: We should really make sure we are writing the same file we are claiming to write in the GUI... Need to think about how to best do that...

Shidory: Before writing at this level, there are different conditions that check if we are actually writing in the file we are trying to write in Gui.

st-review:
:robot: *beep boop*

Issue #5541, opened 2019-02-15
Emit "added" instead of "modified" event when a deleted files is created again


Tstassin:
[Following this discussion in the forum](https://forum.syncthing.net/t/removed-and-then-re-added-subfolder-triggers-deleted-and-then-modified-events/12857)

When a file or folder is deleted, it is marked as `deleted: true` in the database. An event with `data.action: 'deleted'` is also emitted.

When this file or folder is created again OR a file or folder with same name and path is created, it is marked as `deleted: false` in the database and an event with `data.action: 'modified'` is emitted. This is **counterintuitive**, emitting `data.action: 'added' `seems more consistent.

Unfortunately, I'm not good at coding with Go, I will not be able to make the PR.

![capture d ecran 2019-02-15 a 09 01 13](https://user-images.githubusercontent.com/10681663/52858036-502c2000-3129-11e9-91eb-bbc61d425ced.png)

### Version Information

Syncthing Version: v1.0.1
OS Version: Mac OS X / Ubuntu

AudriusButkevicius:
I wouldn't hold your breath on someone fixing this, it's minor and affects pretty much no-one.

calmh:
Actually, I change my mind. The current actions are the following:

- `updated` which refers to all updates of content on disk (added, changed, mkdir)
- `metadata` which is a special for metadata only update (permissions or modtime of files)
- `deleted` when we delete something on disk.

You could argue there should be an `added` but there's never been one and there is no provision for one in the code. So I think this is not a bug, just how we intend things to work.

calmh:
I change my mind again. :) The above is correct for sync actions, but for local change events we do have `added`, and don't correctly emit it after a delete. That is a bug.

imsodin:
The change events are generated by the same function for both local and sync/remote changes at https://github.com/syncthing/syncthing/blob/master/lib/model/folder.go#L811. And at that point we have information on the previous state, i.e. if the file was newly created or modified except for the version vector, which is used for the `added` action type. So AFAIK this is working as expected.

Tstassin:
So according to you the following scenario is correct :

Create directory `dir/`

> Event : added

Create file `dir/document.pdf`

> Event : added

Delete file `dir/document.pdf`

> Event : deleted

One year later, for some other purpose, somebody else creates another `dir/document.pdf`, with another content and purpose the event is still :

> Event : modified

[...]

PR #6605, opened 2020-05-05
lib/protocol: Avoid data loss on database wipe by higher version numbers (fixes #3876)

### LINKED ISSUE #3876: Removing and re-adding a folder may cause data loss
Removing and re-adding a folder causes a reset of the index for that folder and a rescan. Consider the following scenario, with two devices A and B:

- A and B are in sync
- A and B are disconnected
- A changes a couple of files, version vector gets bumped, let's say {A:42, B:36}
- A removes and re-adds the folder
- A rescans and assigns new version vectors with {A:1} to everything
- A and B connect
- B's older file will have a version vector like {A:40, B: 36}
- B's older files overwrite A's newer files (it's not a conflict)

Maybe not a super common scenario, but a folder can be removed by mistake and re-added, or because it's moved on disk, or for many other reasons.

Idea: should the short ID in the version vector not be based on our device ID, but instead a random ID assigned to the folder, e.g., our delta index ID?

calmh:
This makes version vector values clock based instead of just incremented
from zero. The effect is that a vector that is created from scratch
(after database reset) will have a higher value for the local device
than what it could have been previously, causing a conflict. That is, if
we are A and we had

    {A: 42, B: 12}

in the old scheme, a reset and rescan would give us

    {A: 1}

which is a strict ancestor of the older file (this might be wrong). With
the new scheme we would instead have

    {A: someClockTime, b: otherClockTime}

and the new version after reset would become

    {A: someClockTime+delta}

which is in conflict with the previous entry (better).
In case the clocks are wrong (current time is less than the value in the
vector) we fall back to just simple increment like today.

This scheme is ineffective if we suffer a database reset while at the
same time setting the clock back far into the past. It's however no
worse than what we already do.

This loses the ability to emit the "added" event, as we can't look for
the magic 1 entry any more. That event was however already broken
(#5541).

Another place where we infer meaning from the vector itself is in
receive only folders, but there only criteria is that the vector is only
one item long and only includes ourselves, which remain the case with
this change.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,9 +40,9 @@ require (
 	github.com/thejerf/suture v3.0.2+incompatible
 	github.com/urfave/cli v1.22.2
 	github.com/vitrun/qart v0.0.0-20160531060029-bf64b92db6b0
-	golang.org/x/crypto v0.0.0-20200423211502-4bdfaf469ed5
+	golang.org/x/crypto v0.0.0-20200221231518-2aa609cf4a9d
 	golang.org/x/net v0.0.0-20190827160401-ba9fcec4b297
-	golang.org/x/sys v0.0.0-20200223170610-d5e6a3e2c0ae
+	golang.org/x/sys v0.0.0-20191224085550-c709ea063b76
 	golang.org/x/text v0.3.2
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
--- a/go.mod
+++ b/go.mod
@@ -40,9 +40,9 @@ require (
 	github.com/thejerf/suture v3.0.2+incompatible
 	github.com/urfave/cli v1.22.2
 	github.com/vitrun/qart v0.0.0-20160531060029-bf64b92db6b0
-	golang.org/x/crypto v0.0.0-20200221231518-2aa609cf4a9d
+	golang.org/x/crypto v0.0.0-20200423211502-4bdfaf469ed5
 	golang.org/x/net v0.0.0-20190827160401-ba9fcec4b297
-	golang.org/x/sys v0.0.0-20191224085550-c709ea063b76
+	golang.org/x/sys v0.0.0-20200223170610-d5e6a3e2c0ae
 	golang.org/x/text v0.3.2
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -904,19 +904,8 @@ func (f *folder) emitDiskChangeEvents(fs []protocol.FileInfo, typeOfEvent events
 		objType := "file"
 		action := "modified"
 
-		switch {
-		case file.IsDeleted():
+		if file.IsDeleted() {
 			action = "deleted"
-
-		// If our local vector is version 1 AND it is the only version
-		// vector so far seen for this file then it is a new file.  Else if
-		// it is > 1 it's not new, and if it is 1 but another shortId
-		// version vector exists then it is new for us but created elsewhere
-		// so the file is still not new but modified by us. Only if it is
-		// truly new do we change this to 'added', else we leave it as
-		// 'modified'.
-		case len(file.Version.Counters) == 1 && file.Version.Counters[0].Value == 1:
-			action = "added"
 		}
 
 		if file.IsSymlink() {
--- a/lib/protocol/vector.go
+++ b/lib/protocol/vector.go
@@ -2,6 +2,8 @@
 
 package protocol
 
+import "time"
+
 // The Vector type represents a version vector. The zero value is a usable
 // version vector. The vector has slice semantics and some operations on it
 // are "append-like" in that they may return the same vector modified, or v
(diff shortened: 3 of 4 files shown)


AudriusButkevicius:
Made a comment about conflict handling. I think we don't do what we used to do, and we will create conflicts even for identical files.

calmh:
No, it really doesn't, at least not under good circumstances.

Some file changed multiple times on both sides, then touched, dropped db, restarted on one side;

```
jb@kvin:..thing/syncthing/test % touch s1/syncthing
jb@kvin:..thing/syncthing/test % rm -rf h1/index-v0.14.0.db
jb@kvin:..thing/syncthing/test % STTRACE=scanner,model ../bin/syncthing -home h1
[start] 2020/05/05 14:30:27.659300 main.go:588: INFO: syncthing v1.5.0+36-ga31b412c-dirty-clocks "Fermium Flea" (go1.14.2 darwin-amd64) jb@kvin.kastelo.net 2020-05-05 10:59:05 UTC
...
[I6KAH] 2020/05/05 14:30:29.173467 walk.go:372: DEBUG: to hash: syncthing File{Name:"syncthing", Sequence:0, Permissions:0755, ModTime:2020-05-05 14:30:22.998198 +0200 CEST, Version:{[{I6KAH76 1588681829}]}, Length:24464600, Deleted:false, Invalid:false, LocalFlags:0x0, NoPermissions:false, BlockSize:131072, Blocks:[]}
[I6KAH] 2020/05/05 14:30:29.173496 walk.go:240: DEBUG: ignored (internal): .stfolder
[I6KAH] 2020/05/05 14:30:29.173509 walk.go:193: DEBUG: real to hash: syncthing
[I6KAH] 2020/05/05 14:30:29.173516 folder.go:628: INFO: Completed initial scan of sendreceive folder ¯\_(ツ)_/¯ Räksmörgås 动作 Адрес
...
[I6KAH] 2020/05/05 14:30:29.194056 service.go:326: INFO: Established secure connection to MRIW7OK-NETT3M4-N6SBWME-N25O76W-YJKVXPH-FUMQJ3S-P57B74J-GBITBAC at 127.0.0.1:60878-127.0.0.1:22002/tcp-client/TLS1.3-TLS_AES_128_GCM_SHA256
...
[I6KAH] 2020/05/05 14:30:29.241282 folder.go:628: INFO: Completed initial scan of sendreceive folder default
[I6KAH] 2020/05/05 14:30:29.241303 folder.go:243: DEBUG: sendreceive/default@0xc000196000 next rescan in 11.293150492s
[I6KAH] 2020/05/05 14:30:29.241490 model.go:1969: DEBUG: indexSender@0xc000212070 for default to  at 127.0.0.1:60878-127.0.0.1:22002/tcp-client/TLS1.3-TLS_AES_128_GCM_SHA256: Sending 1 files (<9799 bytes)
[I6KAH] 2020/05/05 14:30:29.255417 model.go:943: DEBUG: Index update (in): MRIW7OK-NETT3M4-N6SBWME-N25O76W-YJKVXPH-FUMQJ3S-P57B74J-GBITBAC / "default": 1 files
[I6KAH] 2020/05/05 14:30:29.255933 folder_sendrecv.go:161: DEBUG: sendreceive/default@0xc000196000 pulling
[I6KAH] 2020/05/05 14:30:29.255985 folder_sendrecv.go:233: DEBUG: sendreceive/default@0xc000196000 copiers: 1 pullerPendingKiB: 32768
[I6KAH] 2020/05/05 14:30:29.256172 folder_sendrecv.go:1140: DEBUG: sendreceive/default@0xc000196000 taking shortcut on syncthing
[I6KAH] 2020/05/05 14:30:29.267717 folder_sendrecv.go:186: DEBUG: sendreceive/default@0xc000196000 changed 1 on try 1
[I6KAH] 2020/05/05 14:30:29.267736 folder_sendrecv.go:233: DEBUG: sendreceive/default@0xc000196000 copiers: 1 pullerPendingKiB: 32768
[I6KAH] 2020/05/05 14:30:29.267806 folder_sendrecv.go:186: DEBUG: sendreceive/default@0xc000196000 changed 0 on try 2
[I6KAH] 2020/05/05 14:30:29.267884 progressemitter.go:296: DEBUG: progress emitter: bytes completed for default: 0
...
```

no conflict files

AudriusButkevicius:
I was looking at `ShouldConflict()` which I believe interacts with this in some way, hence the confusion. I am not sure if that still does what it needs.

Going back to the original claim:
```
func (f *sendReceiveFolder) inConflict(current, replacement protocol.Vector) bool {
	if current.Concurrent(replacement) {
		// Obvious case
		return true
	}
	if replacement.Counter(f.shortID) > current.Counter(f.shortID) {
		// The replacement file contains a higher version for ourselves than
		// what we have. This isn't supposed to be possible, since it's only
		// we who can increment that counter. We take it as a sign that
		// something is wrong (our index may have been corrupted or removed)
		// and flag it as a conflict.
		return true
	}
	return false
}
```

This implies that if we see our clock go backwards, we automatically go into conflict, implying the original issue should not be possible to start with?

calmh:
Right, that's a thing and it looks promising. It fails in the case where we created a file and someone else deleted it:

- We create: `{A: 1}`
- They delete: `{A: 1, B: 1}`
- We forget
- We recreate: `{A: 1}`

this is an ancestor of the deleted file, nothing strange and no conflict, we just apply the delete again.

(Something else weird happens when the above code actually works and does it's thing, I just tested... It leaves the newly discovered file, it doesn't get synced to the other side, and both consider themselves up to date... I need to research that.)

AudriusButkevicius:
Presumably the case you suggested could be addressed by changing > to >= and adding `&& replacement != current`

calmh:
I'm not sure why that would be, the `current={A:1}, replacement={A:1, B:1}` case is totally legit however you slice it. It's the same as if we had discovered the file and then later someone deleted it, it's just that the delete entry is already laying about in one of the remote fileinfo sets...

AudriusButkevicius:
I guess we should close the "action" type ticket, as it's not something we will fix by the looks of it.
Perhaps we should change the "modified" to "updated"

PR #6611, opened 2020-05-07
script: Store assets as strings

greatroar:
This changes the static asset storage in lib/api and cmd/strelaypoolsrv from []byte to string to prevent a copy of the 1.6MB of asset data at startup (verified by inspecting the assembler output, go tool compile -S gui.files.go). Since syncthing runs two processes, 3.2MB of RAM is saved.

This requires the HTTP server code to use WriteString instead of Write. I've checked the stdlib implementations of http.ResponseWriter and they implement io.StringWriter, so no performance should be lost there.

The only functional change is that the Content-Length is no longer set when sending static assets to a browser that doesn't understand gzip (using io.Copy was just the easiest option).

### DIFF
--- a/cmd/strelaypoolsrv/main.go
+++ b/cmd/strelaypoolsrv/main.go
@@ -1,11 +1,8 @@
 // Copyright (C) 2015 Audrius Butkevicius and Contributors (see the CONTRIBUTORS file).
 
-//go:generate go run ../../script/genassets.go gui >auto/gui.go
-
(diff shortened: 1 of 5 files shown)


calmh:
Sounds reasonable. api_test.go needs fixup. I'm going to see if I can hack the build server to not have defective and useless build logs.

greatroar:
Fixed api_test.go.

Issue #6519, opened 2020-04-09
Option to auto accept folders from introduced devices


Confuset:
Hi everyone,
this is a feature request for an option that enables the "Auto Accept" feature by default for all newly introduced devices.

Motivation:
In my setup I do have a central server that basically connects all devices. The idea of that server is to allow for mutual online/offline devices to sync my data.  Occasionally I add new devices to that network, but in order to fully function I would have to connect to that central server and enable "Auto Accept" for that new device in order to be able to sync new datafolders across.
If I could set that option on that server once, would be awesome and resolve my issue :-)

This could be improved by a simple "Auto Accept Default Option"-Option :-D

Cheers and stay healthy,
Confuset

AudriusButkevicius:
My guy says no, as I think auto accept is a hack to start with which we never should have implemented. 

People should either set up remote management, or setup their own automation that does the management. 

We will never be able to cover every scenario people come up with, so we should stay lean let people work out their own problems.

Confuset:
Well, the question is how am I supposed to add that feature by any means? I would have to react on that "Device added" event somehow and then fiddle with the config files to actually set "Auto Accept" for that new device. I can't use ssh here, because I can not ssh into that device and than open up the UI to set the option…

My main motivation is the attack surface. If I can set this option on the device itself, I can do that from a given trusted device. No remote access, no management what so ever, is needed here. And the code for that should just be suuper simple… As it is only a default option, which currently defaults to false…which than could be defaulted to true…

???

AudriusButkevicius:
Yes, you can read the config every 30 seconds and set the device to auto accept. 
Heck, you can probably even do it with cron, cli/jq and 20 lines of bash.

Your motivation is you can't do it yourself, or it requires work from your side.
My motivation is tit's a dangerous bolt on feature that solves a problem specific to you.

A few months down the line, we'll have another user asking for some feature very specific to his circumstances, or worse, an extension to some feature we didn't want to have in the first place.

Confuset:
Ok. Lets just turn this around. Would you accept a PR for this?

AudriusButkevicius:
It's not just up to me.

I'd probably be against as I haven't seen strong argument for why this should be available to everyone, as it seems fairly niche to you  (yet ironically I mostly to blame for the introducer and the autoaccept features landing...)

@calmh @imsodin ?

calmh:
[...]

PR #6612, opened 2020-05-07
Auto accept folders default value #6519

MikolajTwarog:
I added new checkbox in advanced options, that changes the default value, that "Auto accept folders" takes when a new device is introduced.


AudriusButkevicius:
I am not sure this is a good way forward for this. 

We probably want defaults for devices and defaults for folders, as this options struct will explode from random stuff.

MikolajTwarog:
You don't want this feature at all or do you want some other implementation?

AudriusButkevicius:
Personally, I don't want it at all, but @calmh suggested it's a sensible continuation.

If we do do this, then I definately don't feel like device defaults should live in options, as that's already a massive struct begging to be split up.

MikolajTwarog:
So you want two new tabs in options, where you can set all the default values for devices and folders?

imsodin:
That seems reasonable to me. In config it can literally be a `FolderConfiguration` object, in the UI I'd reuse the folder edit dialogue as far as it makes sense. The main point is that it should also not have the advanced options that are omitted in the folder edit dialogue.

MikolajTwarog:
I can work on that.

AudriusButkevicius:
We probably want a new section in the root config called defaults, which then have one of each, yet I am not sure reusing the same structs will cut it, as there is no way to say there is no default I guess.

imsodin:
Do we need "no default"? I'd say there's always a "default default". Parts of that currently exists in the UI code and other parts as annotations on the go structs.

MikolajTwarog:
I added the structs. Is this ok so far?

AudriusButkevicius:
Did you try reusing the device/folder structs, or does that not make much sense?

MikolajTwarog:
I think that they have too much stuff, that we don't want defaults for, like ID or label.

AudriusButkevicius:
Sounds fair.
In that case, I'd atleast try to populate the defaults from a default folder config using reflection, what I mean is:

````
from := FolderConfiguration{}
util.SetDefaults(&from)
to := FolderConfigurationDefaults{}
util.CopyMatchingFields(&from, &to)
```
So that we don't need to maintain defaults in two places, as it's bound to go out of sync if not done programatically.

MikolajTwarog:
Should the defaults tab be in advanced or settings?

AudriusButkevicius:
Probably advanced, if at all. People who care about this can probably go tweak the xml.

imsodin:
> Should the defaults tab be in advanced or settings?

Regardless of how it's done in the end (I tend to agree with Audrius), as a first step I would not do anything in the web UI and focus on getting the go/config side of things working. The advanced settings part is generated automatically anyway, which gives you the most basic controls in UI without effort.

MikolajTwarog:
Is there anything else needed?

MikolajTwarog:
I don't understand why the TestNewSaveLoad test in config doesn't work.

AudriusButkevicius:
There is a function that we use that initializes empty slices and maps, I suspect you not applying that function on the config after loading.

AudriusButkevicius:
Check `Configuration.clean()`, namely:
```
if cfg.Folders[i].Versioning.Params == nil {
	cfg.Folders[i].Versioning.Params = map[string]string{}
}
```

MikolajTwarog:
Thanks, it was the problem.

MikolajTwarog:
I also don't know why the model test doesn't work on Build (Linux) and Coverage (Syncthing).

imsodin:
Currently the problem of duplicated defaults still applies (https://github.com/syncthing/syncthing/pull/6612#issuecomment-625918837). Wouldn't a simple solution without reflect be to move the user configurable members to the default type, and embed that into the exiting one. Something like

```
type FolderConfiguration struct {
	DefaultFolderConfiguration

	ID                      string                      `xml:"id,attr" json:"id"`
	Label                   string                      `xml:"label,attr" json:"label" restart:"false"`
[...]
}
type DefaultFolderConfiguration struct {
	RescanIntervalS int                     `xml:"rescanIntervalS,attr" json:"rescanIntervalS" default:"3600"`
	Type            FolderType              `xml:"type,attr" json:"type"`
	Versioning      VersioningConfiguration `xml:"versioning" json:"versioning"`
	Order           PullOrder               `xml:"order" json:"order"`
	MinDiskFree     Size                    `xml:"minDiskFree" json:"minDiskFree" default:"1%"`
}
```

AudriusButkevicius:
Sounds good.

st-review:
:robot: *beep boop*

PR #6616, opened 2020-05-08
Removed health check from Dockerfile (fixes #6604)

### LINKED ISSUE #6604: Docker healthcheck fails when run in host network
When syncthing is run in host network configuration as per guidance [here](https://github.com/syncthing/syncthing/blob/master/README-Docker.md#discovery), the container health check fails, however everything seems to work fine.

#### Configuration:
`STGUIADDRESS=` set as environment variable.
Container is behind a reverse proxy which overwrites host header to `localhost`.

Verified by running `nc -z localhost 8384 || exit 1` inside container. Same works fine when run on host.

shaaraddalvi:
### Purpose

When syncthing container is running in host network mode, the health check fails. However, the services work fine. On dashboard tools like Portainer, it gives a wrong impression that the services aren't responding.

### Testing

Validated that in host network mode the docker container functions properly.

## Authorship

based on the commit metadata.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,8 +22,5 @@ COPY --from=builder /src/script/docker-entrypoint.sh /bin/entrypoint.sh
 
 ENV PUID=1000 PGID=1000 HOME=/var/syncthing
 
-HEALTHCHECK --interval=1m --timeout=10s \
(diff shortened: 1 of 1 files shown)


calmh:
Thanks!

PR #6619, opened 2020-05-08
Clean up after days

MarioMajila:
Clean up of old versions of files versioned with the simple file versioning after days


lib/versioner/simple.go
```diff
@@ -63,6 +72,76 @@ func (v simple) Archive(filePath string) error {
...
+		if err != nil {
+			return err
+		}
+
+		if info.IsDir() && !info.IsSymlink() {
+			dirTracker.addDir(path)
+			return nil
+		}
+		
+		versionTime, err := time.ParseInLocation(TimeFormat, extractTag(path), time.Local)
```
calmh, line 123: Check the error return; if there is an error we should just skip the file (`return nil`)

lib/versioner/simple.go
```diff
@@ -19,22 +23,27 @@ func init() {
...
+	suture.Service
 	keep       int
+	cleanOutDays int
 	folderFs   fs.Filesystem
 	versionsFs fs.Filesystem
 }
 
 func newSimple(folderFs fs.Filesystem, params map[string]string) Versioner {
 	keep, err := strconv.Atoi(params["keep"])
+	cleanOutDays, _ := strconv.Atoi(params["cleanoutDays"])
```
calmh, line 35: This should happen after the error checking that belongs to the line above

PR #6621, opened 2020-05-08
lib/db: Fix non-truncated fileinfo loading

AudriusButkevicius:


calmh:
Nice. I think the only place this is actually used is the recv only revert thing, explaining it going rather unnoticed :)

PR #6623, opened 2020-05-09
Scheduled rate limits

### LINKED ISSUE #6130: Scheduled rate limits
This has come up a lot in the forum but there's no issue for it here that I can find. We've evolved the rate limits a couple of times and this is something that many/most other rate limiting things have. There is a workaround in using something to periodically reconfigure via the API, but this is not realistically something most users can accomplish.

For reference, here's how this looks in Transmission:

Screen Shot 2019-11-06 at 07 33 22

There's a general limit that can be set or not, and an optional other set of limits that can kick in at certain times. This is both quite flexible and not rocketsciency in UI or implementation. I think we should consider having it.

Yithis:
### Purpose

Fixes #6130 

### Screenshots

Screenshot from 2020-05-31 11-04-19
Screenshot from 2020-05-31 11-03-55


Yithis:
At first I created a new fields in the gui and a prototype of the timer. I'm aware that I should add tests and create some way to kill this additional process. Do you have any idea what would be best way to kill this gorutine in case of settings change?

uok:
Looks nice, I'm sure many people will find this helpful! 
I suggest to combine the hour/minutes field into one "from" and one "to" field and move them to left column. Also you could hide the fields if "Enable Scheduled Rates" is unchecked.

Yithis:
Converting those fields into one  isn't that simple due to conversion from html types into go types. But I will try to hide them when this option isn't enable. Still biggest issue is to how to shutdown this additional gorutine when changing limiter. @calmh May you give me a hand with it?

uok:
There is an HTML input type "time" that can be used (which degradeds to input type "text" if unsupported) and it also has max/min attributes.
Ok, I did not know that Go cannot split at the ":" to convert it

calmh:
So about the timer, here's what I would do. On the `limiter` add a `time.Timer` field. In `CommitConfiguration` something roughly like this:

1. if the timer is already non-nil, something was already scheduled. `close(lim.timer.C)` and set `lim.timer = nil`.
2. if we are supposed to schedule a change:
	1. calculate `waitTime` for how long to the next change
	2. `lim.timer = time.NewTimer(waitTime)`
	3. `go lim.timerLoop(lim.timer)`

The `timerLoop` looks something like this:

```
func (lim *limiter) timerLoop(timer *time.Timer) {
  for range timer.C {
    // it's time for a change.
    // Do the appropriate change, under mutex.
    // Figure out the next waitTime
    timer.Reset(waitTime)
  }
}
```

The `for range timer.C` will stop when we `close(lim.timer.C)` on next config change. Otherwise the loop will self-serve via resetting its own timer. It can't look at `lim.timer` which might change during a config change, just use the timer given as parameter.

Yithis:
I applied suggested changes. You can't close receive-only channel, that's why I used `if !t.Stop() {
	<-t.C
}`.

lib/connections/limiter.go
```diff
@@ -124,6 +127,87 @@ func (lim *limiter) VerifyConfiguration(from, to config.Configuration) error {
...
+	begin := time.Date(now.Year(), now.Month(), now.Day(), beginHour, beginMinute, 0, 0, now.Location())
+
+	tillBegin := begin.Sub(now)
+
+	var waitTime time.Duration
+
+	if tillBegin >= 0 {
+		waitTime = tillBegin
+	} else {
+		endHour := to.Options.ScheduledRatesToHour
```
imsodin, line 144: The usual style is to use early returns, e.g. here:

```suggestion
	if tillBegin >= 0 {
		return tillBegin
	}
	endHour := to.Options.ScheduledRatesToHour
```

lib/connections/limiter.go
```diff
@@ -124,6 +127,87 @@ func (lim *limiter) VerifyConfiguration(from, to config.Configuration) error {
...
+		} else {
+			begin = begin.AddDate(0, 0, 1)
+			waitTime = begin.Sub(now)
+		}
+	}
+
+	return waitTime
+}
+
+func (lim *limiter) timerLoop(timer *time.Timer, to config.Configuration) {
```
imsodin, line 162: I wouldn't pass the entire configuration, just the 4 values required.

lib/connections/limiter.go
```diff
@@ -124,6 +127,87 @@ func (lim *limiter) VerifyConfiguration(from, to config.Configuration) error {
...
 }
 
+func nextChange(to config.Configuration) time.Duration {
+	now := time.Now()
+
+	beginHour := to.Options.ScheduledRatesFromHour
+	beginMinute := to.Options.ScheduledRatesFromMinute
+	begin := time.Date(now.Year(), now.Month(), now.Day(), beginHour, beginMinute, 0, 0, now.Location())
+
+	tillBegin := begin.Sub(now)
```
imsodin, line 137: This entire time dance happens a few times, worth making a small utility function like e.g. `timeDiff(now time.Time, day, hour, minute int) time.Duration`

lib/connections/limiter.go
```diff
@@ -172,6 +256,20 @@ func (lim *limiter) CommitConfiguration(from, to config.Configuration) bool {
...
+		lim.timer = nil
+	}
+
+	if to.Options.ScheduledRatesEnabled {
+
+		waitTime := nextChange(to)
+		lim.timer = time.NewTimer(waitTime)
+		go lim.timerLoop(lim.timer, to)
+	}
+
```
imsodin, line 272: That should go to the beginning of the function and the rest of it skipped if scheduling is activated, as the limits are then set in the timer.

lib/connections/limiter.go
```diff
@@ -124,6 +127,87 @@ func (lim *limiter) VerifyConfiguration(from, to config.Configuration) error {
...
+		}
+
+		if sendLimitValue <= 0 {
+			lim.write.SetLimit(rate.Inf)
+		} else {
+			lim.write.SetLimit(1024 * rate.Limit(sendLimitValue))
+			sendLimitStr = fmt.Sprintf("limit is %d KiB/s", sendLimitValue)
+		}
+
+		l.Infof("Overall send rate %s, receive rate %s", sendLimitStr, recvLimitStr)
```
imsodin, line 204: Setting the limits duplicates the existing code in `CommitConfiguration`, should probably become a utility function.

Yithis:
I changed GUI and code like suggested

lib/connections/limiter.go
Yithis: Done

lib/connections/limiter.go
Yithis: Done

lib/connections/limiter.go
Yithis: Done

lib/connections/limiter.go
Yithis: I didn't fix this one, in this function you are using 8 values (4 for time and 4 for rates), so I'm not sure it would be good change

lib/connections/limiter.go
Yithis: Done

lib/connections/limiter.go
```diff
@@ -134,37 +222,30 @@ func (lim *limiter) CommitConfiguration(from, to config.Configuration) bool {
 
 	if from.Options.MaxRecvKbps == to.Options.MaxRecvKbps &&
 		from.Options.MaxSendKbps == to.Options.MaxSendKbps &&
-		from.Options.LimitBandwidthInLan == to.Options.LimitBandwidthInLan {
+		from.Options.LimitBandwidthInLan == to.Options.LimitBandwidthInLan &&
+		lim.timer == nil && !to.Options.ScheduledRatesEnabled {
```
imsodin, line 226: This does not return early, if scheduled rate limits are enabled both in from and to and no limits changed, but it could, right? I'd have expected this:

```suggestion
		from.Options.ScheduledRatesEnabled == to.Options.ScheduledRatesEnabled {
```

Am I missing something?

lib/connections/limiter.go
imsodin: Just `config.Options` then?

lib/connections/limiter.go
Yithis: I didn't want to return early in any case that timer should exist, since in this case we have to check 2 next values for scheduled limits and 4 next values for time changing.
```suggestion
		from.Options.ScheduledRatesEnabled == to.Options.ScheduledRatesEnabled &&
		from.Options.ScheduledRatesFromHour == to.Options.ScheduledRatesFromHour &&
		from.Options.ScheduledRatesFromMinute == to.Options.ScheduledRatesFromMinute &&
		from.Options.ScheduledRatesToHour == to.Options.ScheduledRatesToHour && 
		from.Options.ScheduledRatesToMinute == to.Options.ScheduledRatesToMinute &&
		from.Options.MaxScheduledSendKbps == to.Options.MaxScheduledSendKbps &&
		from.Options.MaxScheduledRecvKbps == to.Options.MaxScheduledRecvKbps {
```

imsodin: Ah that's what I missed - thanks for reminding me :)

lib/connections/limiter.go
```diff
@@ -124,6 +127,91 @@ func (lim *limiter) VerifyConfiguration(from, to config.Configuration) error {
...
+		beginHour := to.Options.ScheduledRatesFromHour
+		beginMinute := to.Options.ScheduledRatesFromMinute
+		begin := time.Date(now.Year(), now.Month(), beginDay, beginHour, beginMinute, 0, 0, now.Location())
+
+		endHour := to.Options.ScheduledRatesToHour
+		endMinute := to.Options.ScheduledRatesToMinute
+		endDay := correspondingEndDay(beginDay, beginHour, beginMinute, endHour, endMinute)
+		end := time.Date(now.Year(), now.Month(), endDay, endHour, endMinute, 0, 0, now.Location())
+
+		if now.Sub(begin) >= 0 && end.Sub(now) >= 0 {
```
imsodin, line 204: This time handling duplicates that in `nextChange`: In `nextChange` you already implicitly determine whether the change is from scheduled to normal or the other way around. You could return that as a bool and use it to decide to which limit to change to. Thus no time handling necessary at all here in `timerLoop`.

Yithis: Changed as suggested

lib/connections/limiter.go
Yithis: Sound good, changed

Yithis:
Should I do anything else? Or everything is fine already?

AudriusButkevicius:
So I have not been following the PR, but just my two cents. The schedules should probably go into new schedules struct.

We should probably create a new type that can be serialized and deserialized from which supports hh:mm format, so we don't need many variables per timeslice.

Yithis:
I agree that this would be a perfect solution, but is it good to change this many things at once? In case of any problems it would be harder to find mistake. I thought that adding such struct would be a good next PR.

AudriusButkevicius:
I personally would not be ok landing a PR that adds so many fields to the config.

Yithis:
Implementing such structure would change amount of fields in the config from 6 to 4, but I understand. If one more person will give feedback that such change would be good in this PR I will implement this, since I would like to have a direct path where this PR should go.

calmh:
I just glanced at this and thought the same. You might look at the free space percentage thing which is a similar custom config type, in that case holding a value and a unit, in this case maybe a time structure to begin with and then a schedule structure that binds two of those together. I didn't look if it's already the case but we can be sure that the next requirement will be scheduling things between time A and time B on Tuesdays, Wednesdays and Sundays only. It'd be good if there were a schedule structure then that we can just add a days field to.

AudriusButkevicius:
If we land this, people will start using this, so any change to this will then need a migration, which means we'll have to keep the old fields forever.

In theory the field could support `09:30-16:30` text as a timerange, reducing the number of fields.

Also, I have not looked at the code, but it's not clear to me from the ui what's the precedence of the limits. Global always applied, and then scheduled takes over when in the schedule?

calmh:
I would settle for something like a

```
type schedule struct {
  starts scheduleTime
  ends scheduleTime
}

type scheduleTime {
  hour int
  minute int 
}
```

and the corresponding obvious serializations

```
<schedule>
  <starts hour=12 minute=0 />
  <ends hour=23 minute=30 />
</schedule>
```

I would expect the scheduled limits to *replace* the global / standard limit when the schedule takes effect. (That is, it can be either higher or lower and the effect will be the expected one...)

AudriusButkevicius:
What if you want different days to start at different times 😱

calmh:
Then you get to justify and implement that.

AudriusButkevicius:
I am more concerned in making a choice of config structures that prevents implementing this in the future without a massive migration/deprecation hoop.

calmh:
Well having some like above and then adding for example the `days` field is migrationless. (Assuming we take the absence of a days field to mean "no restriction on days").

AudriusButkevicius:
I think we should probably have an array of start+end time pairs and currently only assume there will always be one which will have a "every day" meaning for now. Then when we decide to do multiple days, we have a few ways to go around it, we could expand into NumberOfDays elements per day.

```
type Configuration struct {
     ...
     Schedules struct {
           RateLimitSchedule RateLimitSchedule
     }
     ...
}


type RateLimitSchedule {
     MaxRecvKbps int
     MaxSendKbps
     Schedule Schedule
}

type Schedule struct {
    Entries []ScheduleEntry
}

type ScheduleEntry struct {
     Start HourMinute
     End HourMinutte
}
```

There is then also the `I want it at 300kb/s on Tuesdays but 700kb/s on Wednesdays` case, which would need []RateLimitSchedule instead of RateLimitSchedule...

Basically, this is why I was even reluctant to do anything about this ticket all together.

As soon as we add any of this functionality, people will come after us with pitch forks for not supporting advanced enough scenarios.

These things are easily controller via the CLI, and people should just use cron or something that is not shit scheduling stuff and has a ui complicated enough to manage this.

uok:
I think the main use case for scheduled rate limit is e.g. for an office during business hours or e.g. at home at night when the user is home. 

The easiest to use schedulers for rate limiting I have seen over the years are built into torrent clients. The GUI has a large table with day of week and one cell for each hour of the day. The user can then "draw" when the rate limit is active or not. This should suffice for 99% of use cases and the config would only need to store an array for each day (or one big array for the week)

[Examples](https://www.google.com/search?q=torrent+scheduler&tbm=isch) (Google Image Search)

AudriusButkevicius:
Sure, effectively I want to make sure that whatever config structures we pick, we can support that in the future.
Even the uTorrent example is tri-state, high speed, low speed, off.

imsodin:
I think there's a bit too much worrying about potential future hoops and obsolete stuff here - what's the likeliness that anyone takes the time to come up with a reasonable PR to do such intricate scheduling? None I'll wager. I personally would rather have a potential migration in the future than complex data structures in config that we don't handle (or at least realistically intend to handle in the future).

AudriusButkevicius:
I'd personally be against merging the current implementation, which adds 7 fields to an already large existing struct.

As a minimum I think it should go into it's own Config->Schedules->RateLimitSchedule struct.
I think I'd still prefer to have an HourMinute type for startTime, endTime.

The rest of what I am proposing is to just turning Foo into []Foo and Bar into []Bar, and use x[0] handling in a few places, which sure, we can skip, but it's not actually rocket science.

calmh:
But it's also unnecessary and the XML config has the sometimes nice property that a `<foo>...</foo>` will marshal/unmarshal the same into a `[]Foo` or a `Foo` field, so we don't need to jump into that hoop right now. I agree with imsodin -- simpler is better -- and with Audrius -- wrap all the times and stuff in appropriate types.

AudriusButkevicius:
Sure, sounds good as long as we can make the XML work later without having to carry the `Deprectated_` stuff.

uok:
following up on my idea for the weekly scheduler: do we even need the minute granularity?

0 ... rate limit is off
1 ... rate limit is on

so a business week 9-17 would look like this in XML
```
M 000000000111111111000000
T 000000000111111111000000
W 000000000111111111000000
T 000000000111111111000000
F 000000000111111111000000
S 000000000000000000000000
S 000000000000000000000000
```
The scheduler reads this config and simply keeps all days+hours that have a scheduled "1" in the memory. The "0" are not needed

examples 
Monday, 11:46 -> there is a "1" for hour "11" in the config -> rate limit on
Saturday, 20:53 -> there is nothing in the memory config -> rate limit off

imsodin:
Please don't use Yithis' PR to discuss potential future extensions but keep that on the forum and eventually a new issue if it considered a desirable enhancement (unless of course Yithis expressed an intention to work on such an enhancement now, which I am not aware of - please correct me if wrong). That side-tracks the completion of the efforts in this PR.
 
My understanding is that for this PR, a better structure/serialisation for the rate limiting config options is required. Something like what calmh brought up in https://github.com/syncthing/syncthing/pull/6623#issuecomment-636693263 and has been discussed in the follow-up comments. @Yithis I hope a path forward for the config is clear, otherwise please ask for clarification.

uok:
ok, sorry - I will continue in forum (which is not working at the moment)

lib/config/config.go
```diff
@@ -244,6 +244,10 @@ func (cfg *Configuration) prepare(myID protocol.DeviceID) error {
 
 	cfg.MyID = myID
 
+	if len(cfg.Options.ScheduledRates.Entries) > 1 {
+		cfg.Options.ScheduledRates.Entries = cfg.Options.ScheduledRates.Entries[1:2]
+	}
```
Yithis, line 249: Without those lines there is problem that to load a configuration, you have to create a new one and then read from XML, so this loaded configuration will have default field of ScheduledRates.Entries at the first place and loaded at the second place.

Yithis:
Does current solutions sounds good for now and in terms of future updates?

AudriusButkevicius:
My personal suggestion was to move all of this schedule stuff out of options in general.

Create a new `Schedules` struct at the root of the config (not on Options like it is now), and create 
 RatesSchedule struct that lives on the Schedules struct. We can then latter add FolderSchedules or whatever if ever needed.

Also, I think you might not need an explicit enabled variable, as in theory if start time = end time, that implies it's disabled, also if the rates are 0 (infinite), that implies it's disabled? (Or is zero suddenly "stop all traffic thing"?)

Yithis:
> Create a new `Schedules` struct at the root of the config (not on Options like it is now), and create
> RatesSchedule struct that lives on the Schedules struct. We can then latter add FolderSchedules or whatever if ever needed.

Ok, I will move this structure.
 
> Also, I think you might not need an explicit enabled variable, as in theory if start time = end time, that implies it's disabled, also if the rates are 0 (infinite), that implies it's disabled? (Or is zero suddenly "stop all traffic thing"?)

I agree it is something you can do, but I think that with a more advanced usage you will have to anyway add this field. Also it is more readable from a code point of view. Also it is more consistent to stay with 0 implies infinite limit. But if you really want I can change it too.

AudriusButkevicius:
I am not sure what 0 mean, I am bringing this up for discussion as to what feels right.
I think you can get away without an enable flag forever, because absence of entries or absence of time gap implies it's not enabled.

Yithis:
Does current version look better for you, @AudriusButkevicius?

AudriusButkevicius:
So a few comments:
1. I think the config should have generic schedule holder struct called Schedules, not ScheduleRates, and then Schedules structs has RatesSchedule struct.
2. MaxRecvKbps/MaxSendKbps should be lifted out of ScheduleEntry into RatesSchedule
3. RatesSchedule can have a single ScheduleEntry for now, we can convert it to an array when we need to, without breaking XML compatibility.
4. Remove StartDay/EndDay for now, as we're not handling that now anyway.
5. We could rename ScheduleEntry to TimeFrame or something, because once you list the Kbps stuff out of it, that name would more correctly represent what it is.

Effectively:
```
Config {
   Schedules {
      RatesSchedule {
          MaxSendKbps int
          MaxRecvKbps int
          Time TimeFrame 
       }
   }
}
```
We can later convert RatesSchedule into []RatesSchedule and
Time into []Time if needed.

imsodin:
What's the motivation for putting this at the config top level? Sure the options object is vast, but moving single items out of it into the top-level just moves that problem one level up if followed through to the end. To reduce the size of options you'd had to create config groups like "connection" instead, but that's not related to this PR. I think schedules should stay in options, I don't want a cluttered top-level config and cluttered options.

AudriusButkevicius:
Let's say hypothetically, we add FolderPauseSchedule DevicePauseSchedule, RemoteDeviceRateSchedule etc (these would be slices), would you still want to have this in Options?

It's the same argument that was given when adding pending devices pending folders, or LDAP stuff.

Options has become a catch all for everything, and if it's more than a single parameter, it probably should not live in Options.

lib/config/time.go
```diff
@@ -12,78 +12,86 @@ import (
...
-	StartDay    int `json:"startDay" xml:"startDay"`
 	EndHour     int `json:"endHour" xml:"endHour"`
 	EndMinute   int `json:"endMinute" xml:"endMinute"`
-	EndDay      int `json:"endDay" xml:"endDay"`
-	MaxSendKbps int `json:"maxSendKbps" xml:"maxSendKbps"`
-	MaxRecvKbps int `json:"maxRecvKbps" xml:"maxRecvKbps"`
 }
 
-func ParseSchedule(s string) (Schedule, error) {
+func ParseSchedules(s string) (Schedules, error) {
```
AudriusButkevicius, line 32: Not really sure if this is needed anymore.

I think the special parsing is only needed for TimeFrame (00:00-00:00, or perhaps break it into 2 x TimeMinute which parses just 00:00) but to be honest, given this now lives in it's own struct, perhaps it's ok to leave it as it is without any special parsing.

Also, I don't think we need any defaults here, all zeroes here are sensible defaults.

lib/config/time.go
```diff
@@ -12,78 +12,86 @@ import (
...
-	return fmt.Sprintf("%d:%d-%d:%d %d %d", s.Entries[0].StartHour, s.Entries[0].StartMinute,
-		s.Entries[0].EndHour, s.Entries[0].EndMinute, s.Entries[0].MaxSendKbps, s.Entries[0].MaxRecvKbps)
+func (s Schedules) String() string {
+	return fmt.Sprintf("%d:%d-%d:%d %d %d", s.RatesSchedule.Time.StartHour, s.RatesSchedule.Time.StartMinute,
+		s.RatesSchedule.Time.EndHour, s.RatesSchedule.Time.EndMinute, s.RatesSchedule.MaxSendKbps, s.RatesSchedule.MaxRecvKbps)
 }
 
-func (s Schedule) IsEnabled() bool {
-	return len(s.Entries) > 0
+func (s Schedules) IsEnabled() bool {
```
AudriusButkevicius, line 88: Assume this is a container for all other future schedules we might implement, in which case it does not make sense to have IsEnabled() on this. If you need it, it should probably be on RatesSchedule

imsodin:
> Let's say hypothetically, we add FolderPauseSchedule DevicePauseSchedule, RemoteDeviceRateSchedule etc (these would be slices), would you still want to have this in Options?

I get that point, I guess makes sense. I was still operating too much under the impression that's never going to happen.

AudriusButkevicius:
I am too, but it doesn't cost much to be ready for it.

lib/config/time.go
Yithis: Changed

lib/config/time.go
Yithis: Changed

Yithis:
This one test is failing, because I deleted code from previous commits. In my opinion it is not a issue and it should pass.

st-review:
:robot: *beep boop*

jtagcat:
What's the status on this? Is it stuck behind a PR test (not shown on closed PRs), and nothing else?

PR #6624, opened 2020-05-09
cmd/strelaypoolsrv, lib/api: Factor out static asset serving

greatroar:
Much of the asset handling code shared by these modules now lives in lib/api/assets, with the following modifications.

* strelaypoolsrv had its mtime checking reversed: it reported StatusNotModified when modified.Add(time.Second).After(t).
* Generated ETags were "bare" decimal numbers, but [RFC 7232 §2.3](https://tools.ietf.org/html/rfc7232#section-2.3) requires them to be quoted strings. Since this changes invalidates all old ETags, I changed them to hex for compactness while I was at it.
* lib/api added charset=utf-8 for textual MIME types, but strelaypoolsrv  didn't. Both now do.


imsodin:
Just an idea, haven't thought it through: Would it make sense to have this new code in `lib/api/assets` and `lib/auto` in a single package?

greatroar:
I would have merged the two, but cmd/strelaypoolsrv has different assets (actually only one) than syncthing.

AudriusButkevicius:
Perhaps this should be lib/assets and lib/auto. lib/api is a syncthing specific package for it's rest api, so it feels weird that we reach into that in the relay server.

greatroar:
Your call. The Go module system doesn't care what the path is :)

imsodin:
So `lib/api/assets` -> `lib/assets` and `lib/auto` -> `lib/api/auto`? (We're really good at bikeshedding :P )

AudriusButkevicius:
Yeah, sounds good.

greatroar:
Ok, did that.

PR #6626, opened 2020-05-09
lib/db, lib/model: Add FileSet.FolderIsNew

imsodin:
Improves the "new folder detection"-heuristics by using the database's folder id index instead of looking for file entries.

This was prompted by https://forum.syncthing.net/t/stop-syncthing-from-auto-creating-syncd-folders-that-dont-exist/14920.

### DIFF
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -23,10 +23,11 @@ import (
 )
 
 type FileSet struct {
-	folder string
-	fs     fs.Filesystem
-	db     *Lowlevel
-	meta   *metadataTracker
+	folder      string
+	folderIsNew bool
+	fs          fs.Filesystem
+	db          *Lowlevel
+	meta        *metadataTracker
 
 	updateMutex sync.Mutex // protects database updates and the corresponding metadata changes
 }
@@ -64,11 +65,16 @@ type Iterator func(f FileIntf) bool
 func NewFileSet(folder string, fs fs.Filesystem, db *Lowlevel) *FileSet {
 	return &FileSet{
 		folder:      folder,
+		folderIsNew: !db.folderIdx.HasID([]byte(folder)),
 		fs:          fs,
 		db:          db,
 		meta:        db.loadMetadataTracker(folder),
 		updateMutex: sync.NewMutex(),
 	}
+}
+
+func (s *FileSet) FolderIsNew() bool {
+	return s.folderIsNew
 }
 
 func (s *FileSet) Drop(device protocol.DeviceID) {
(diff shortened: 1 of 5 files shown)


AudriusButkevicius:
Will this actually work correctly? Because it gets created as soon as it's used, so even if we set this variable to true, pausing unpausing the folder now makes this false.

imsodin:
That's the point: Once we created the folder path and marker (first time the folder is started) we never want to do it again.

I previously relied on `db.loadMetadataTracker` to create an entry in the folder index, I now do that explicitly before and explain in a comment why.

imsodin:
> I should write a test though.

Added a test, fails on master and passes here :)

calmh:
This strikes me as potentially annoying as if the first mkdirs fail for any reason (like wrong permissions) they will never be retried, instead the user need to manually sort through “folder path missing” and dire warnings about “folder marker missing” after that. 

How about a flag to lock it down once a first scan has succeeded?

calmh:
Also there should be an issue describing the problem this solves.

imsodin:
The "issue" it solves is what I inferred from the linked forum thread:
1. Syncthing folder created on a mount root, nothing in there.
2. Then later Syncthing is started while the mount isn't online.
Result: Syncthing creates the root folder and `.stfolder` in the underlying fs, mount no longer mountable.

I'd say that's partly user fault, but thought this is a neat and easy/quick solution to prevent the problem. However admittedly not being able to create the marker first time around and then not retrying is the same kind of problem: In principle user error, but still annoying and should be prevented. So this exchanges one niche problem with another, net gain zeroish. And I am not interested enough in this to actually add a new db flag or something for it -> closing.

Catfriend1:
Could we please list this as an issue, so there's hope someone who feels attracted will do it in the future?

> Syncthing folder created on a mount root, nothing in there.
Then later Syncthing is started while the mount isn't online.
Result: Syncthing creates the root folder and .stfolder in the underlying fs, mount no longer mountable.
(...)
How about a flag to lock it down once a first scan has succeeded?

AudriusButkevicius:
Do what exactly? I think that current handling of this is the more sane version than what was proposed here.

I don't really see a way this can be fixed without introducing a different set of annoyances.

imsodin:
We could in the db track whether the folder marker was successfully created and then no longer recreate it (which is equivalent to "when first scan happened", which calmh proposed above). As opposed to trying again and again until any other file was scanned. That seems like a better behaviour, but also like too much effort for a preventable niche problem. I think in principle it's a legitimate enhancement issue, however to do this you need to know the model and db, two of the most convoluted things about syncthing, so I'd wager it won't ever happen.

AudriusButkevicius:
Perhaps we should fail folder creation if we can't create the marker straight away, database magic for this purpose is magic.

PR #6633, opened 2020-05-12
Add 'RestartSec=5' to Systemd service (system)

NinoM4ster:
### Purpose
Might help avoiding the 'Start request repeated too quickly.' error.

### Testing
Since the error doesn't happen very often, it's hard to test it (and I just applied it on my server anyways).
But it shouldn't be harmful, should it?

### Documentation
Maybe mention that it might take 5 seconds to restart if failed.

### DIFF
--- a/etc/linux-systemd/system/syncthing@.service
+++ b/etc/linux-systemd/system/syncthing@.service
@@ -7,6 +7,7 @@ After=network.target
 User=%i
 ExecStart=/usr/bin/syncthing -no-browser -no-restart -logflags=0
 Restart=on-failure
+RestartSec=5
 SuccessExitStatus=3 4
 RestartForceExitStatus=3 4
 


calmh:
Also add it to the system unit?

NinoM4ster:
I tried adding to both without creating two fork requests but couldn't figure it out, so no, I only added it on /system/syncthing@.service

calmh:
Nonetheless it should be done on both. :) You can add another commit to this branch and it will show up in the same pull request.

NinoM4ster:
will do, in a few hours.

thanks for the info though, I'm still not used to GitHub.

NinoM4ster:
well, it actually did not. unless I really didn't see the right button.

https://github.com/syncthing/syncthing/pull/6634

calmh:
Thanks, I merged it manually.

PR #6635, opened 2020-05-12
lib/db: Fix updating/removing from global (ref #6413)

imsodin:
This PR is a weird one: It corrects mistakes in #6413, but those mistakes are "benign". Meaning I only discovered them because I needed to change exactly that code for the global version list refactor and it looked off, however the both the old and corrected code do the same thing in the end. The corrected one is just less convoluted and/or actually takes the shortcuts intended (e.g. the return in line 809).

I suggest to look at it as separate commits: The complete diff is rather ugly, but the first four commits are straightforward (just moving stuff around). The last one isn't entirely straight forward, but it's still much clearer what's going on than in the full diff.

This is sort of a grey area in terms of RC: The PR only changes stuff that's RC specific, but there's also no real bug, in principle just "improvements".

### DIFF
--- a/lib/db/transactions.go
+++ b/lib/db/transactions.go
@@ -763,17 +763,50 @@ func (t readWriteTransaction) removeFromGlobal(gk, keyBuf, folder, device []byte
 		return nil, err
 	}
 
+	if len(fl.Versions) == 0 {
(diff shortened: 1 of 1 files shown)


calmh:
I think it's safe to say I don't understand these parts of the code as well as I should any more.

PR #6376, opened 2020-02-27
Add indirection for large version vectors.

calmh:
### Purpose

Add block list-like indirection for version vectors with more than ten members. 

Example difference in an extreme (but real world) case:

```
 0x00: 1169574 items, 70491 KB keys +  5506325 KB data, 60 B +   4707 B avg,   7634 B max
 0x01:   52634 items,  3004 KB keys +  5412346 KB data, 57 B + 102829 B avg, 517727 B max
 0x02:  259848 items, 23082 KB keys +     1039 KB data, 88 B +      4 B avg,    256 B max
 ...
 0x0d:   44773 items,  1477 KB keys +    12328 KB data, 33 B +    275 B avg,  99123 B max
 Total 1581101 items, 98761 KB keys + 10935349 KB data.
```

to:

```
 0x00: 1169574 items,  70491 KB keys +  219351 KB data, 60 B +    187 B avg,    655 B max
 0x01:   52634 items,   3004 KB keys + 5412346 KB data, 57 B + 102829 B avg, 517727 B max
 0x02:  259848 items,  23082 KB keys +    1039 KB data, 88 B +      4 B avg,    256 B max
 ...
 0x0d:   44772 items,   1477 KB keys +   12328 KB data, 33 B +    275 B avg,  99123 B max
 0x0e:   40817 items,   1346 KB keys +  234449 KB data, 33 B +   5743 B avg,   7336 B max
 Total 1621918 items, 100108 KB keys + 5882824 KB data.
```

PR #6638, opened 2020-05-13
lib/db: Refactor to use global list by version (ref #6372)

imsodin:
WIP notice: This branch sits on top of https://github.com/syncthing/syncthing/pull/6376, i.e. that should be merged first and the relevant changes start with 317932a.

The changes in this PR were already partly discussed in the comments following https://github.com/syncthing/syncthing/issues/6372#issuecomment-625461989.

Goal
---

Group the global list of files by version, instead of having one flat list for all devices. This removes lots of duplicate `protocol.Vector`s. 

I strongly recommend looking at this commit by commit, or at least in these "two stages":

"Stage 1":
---

In all commits except the last one (39f3e91) I there's no change to the underlying proto storage. These changes serve to separate logic of handling the storage types in structs.go from the higher-level db changes in transactions.go and other changes necessary for the last commit, that can stand on their own (i.e. every commit is functional).

e2a48f5:  
I need access to `WinsConflict` with `FileIntf` - it's currently defined on `protocol.FileInfo`.

"Stage 2":
---

Actually change the proto storage. It is now a ordered list of versions, containing two lists of devices:  one with devices that have an valid entry and one for those with invalid ones. This means the first entry may not be the global one, if it only contains invalid devices (and at least one valid one exists). Thus to get the global version, one must use `.GetGlobal`, not just get the first version. To ensure that I used our usual `Raw` prefix to make it `.RawVersions`.


lib/db/structs.go
```diff
@@ -196,98 +195,297 @@ func (vl VersionList) String() string {
...
+	if invDevice {
+		vl.RawVersions[i].InvalidDevices = popDeviceAt(vl.RawVersions[i].InvalidDevices, j)
+	} else {
+		vl.RawVersions[i].Devices = popDeviceAt(vl.RawVersions[i].Devices, j)
 	}
-	return vl, FileVersion{}, -1
+	// If the last valid device of the previous global was removed above,
+	// the next entry is now the global entry (unless all entries are invalid).
+	if len(vl.RawVersions[i].Devices) == 0 && globalPos == i {
+		return vl, vl.RawVersions[i], true, globalPos == vl.findGlobal(), nil
```
calmh, line 297: There's literally no way `globalPos == vl.findGlobal()` can be true here, as we've already established there are no valid devices in that version? (I'm not sure if that means we can just simplify a little or if there's a logic error in the thinking somewhere else...)

lib/db/structs.go
```diff
@@ -196,98 +195,297 @@ func (vl VersionList) String() string {
...
-			// to determine the winner.)
-			//
-			// A surprise missing file entry here is counted as a win for us.
-			if of, ok, err := t.getFile(folder, vl.Versions[i].Device, []byte(file.Name)); err != nil {
-				return vl, removedFV, removedAt, i, err
-			} else if !ok || file.WinsConflict(of) {
-				vl = vl.insertAt(i, nv)
-				return vl, removedFV, removedAt, i, nil
-			}
+	for ; i < len(vl.RawVersions); i++ {
```
calmh, line 244: Shouldn't this loop be a `func (vl *VersionList) insert(...)`?

lib/db/lowlevel.go
```diff
@@ -499,6 +492,30 @@ func (db *Lowlevel) checkGlobals(folder []byte, meta *metadataTracker) error {
...
+		}
+		f, ok, err := t.getFileTrunc(dk, true)
+		if err != nil {
+			return vl, false, err
+		}
+		if !ok {
+			changed = changed || ok
+			continue
+		}
+		vl, _, _, _, _, _, _, err = vl.update(folder, device, f, t)
```
calmh, line 511: this is literally a caterpillar emoji 🐛 

that's probably a red flag of some kind :D

lib/db/lowlevel.go
```diff
@@ -499,6 +492,30 @@ func (db *Lowlevel) checkGlobals(folder []byte, meta *metadataTracker) error {
...
+		dk, err = t.keyer.GenerateDeviceFileKey(dk, folder, device, name)
+		if err != nil {
+			return vl, false, err
+		}
+		f, ok, err := t.getFileTrunc(dk, true)
+		if err != nil {
+			return vl, false, err
+		}
+		if !ok {
+			changed = changed || ok
```
calmh, line 508: wat (`ok == false` here)

lib/db/schemaupdater.go
```diff
@@ -629,5 +669,298 @@ func (db *schemaUpdater) updateSchemaTo12(_ int) error {
...
+
+func (db *schemaUpdater) rewriteGlobals(t readWriteTransaction) error {
+	it, err := t.NewPrefixIterator([]byte{KeyTypeGlobal})
+	if err != nil {
+		return err
+	}
+	for it.Next() {
+		var vl VersionListDeprecated
+		if err := vl.Unmarshal(it.Value()); err != nil {
+			return err
```
calmh, line 687: should we consider that we might crash after committing some new version lists but before completing the migration, so that we should skip unmarshal errors on the assumption that they might already be rewritten?

lib/db/structs.go
imsodin: A version with no valid devices can be the global version, if there's no valid devices at all. That's what I tried to explain in the comment above. If we were sure there's still a valid device, we could just return `true` here.

lib/db/schemaupdater.go
imsodin: Adressed by unmarshalling in the new format on error and `continue` if that succeeds.

lib/db/structs.go
imsodin: Done.

lib/db/lowlevel.go
imsodin: A green caterpillar though, so it's fine?! :P 

We need all the return values in `updateGlobal`. We could call `getGlobal` before and after there to shorten the caterpillar by 4 elements, but that would be redundant (as it's exactly what `update` does internally).

lib/db/schemaupdater.go
```diff
@@ -629,5 +669,304 @@ func (db *schemaUpdater) updateSchemaTo12(_ int) error {
...
+				break
+			}
+			oldPos++
+			if lastVersion.Equal(fv.Version) {
+				newVl.RawVersions[newPos].Devices = append(newVl.RawVersions[newPos].Devices, fv.Device)
+				continue
+			}
+			newVl.RawVersions = append(newVl.RawVersions, newFileVersion(fv.Device, fv.Version, false, fv.Deleted))
+			lastVersion = fv.Version
+			newPos++
```
calmh, line 716: `newPos = len(newVl.RawVersions)` before the append avoids the odd initialization to -1 and the risk of somehow getting out of sync in the ++

lib/db/schemaupdater.go
```diff
@@ -629,5 +669,304 @@ func (db *schemaUpdater) updateSchemaTo12(_ int) error {
...
 
+	if err := db.rewriteGlobals(t); err != nil {
+		return err
+	}
+
 	return t.Commit()
 }
+
+func (db *schemaUpdater) rewriteGlobals(t readWriteTransaction) error {
+	it, err := t.NewPrefixIterator([]byte{KeyTypeGlobal})
```
calmh, line 680: defer an extra it.Release

lib/db/schemaupdater.go
```diff
@@ -629,5 +669,304 @@ func (db *schemaUpdater) updateSchemaTo12(_ int) error {
...
+			}
+			continue
+		}
+		if len(newVl.RawVersions) == 0 {
+			fv := vl.Versions[oldPos]
+			newVl.RawVersions = []FileVersion{newFileVersion(fv.Device, fv.Version, true, fv.Deleted)}
+		}
+		newPos = 0
+	outer:
+		for _, fv := range vl.Versions[oldPos:] {
```
calmh, line 730: I don't really understand what happens here or why. What's the reason we can't just trust the order the versions are currently in? Are the ordering requirements in the new format different than in the old?

lib/db/schemaupdater.go
```diff
@@ -629,5 +669,304 @@ func (db *schemaUpdater) updateSchemaTo12(_ int) error {
...
+			}
+			return err
+		}
+		if len(vl.Versions) == 0 {
+			if err := t.Delete(it.Key()); err != nil {
+				return err
+			}
+		}
+
+		var newVl VersionList
```
calmh, line 701: This thing might do well as a `func convertVersionList(vl DeprecatedVersionList) VersionList` (with a readOnlyTransaction as well I guess as required)...?

lib/db/schemaupdater.go
imsodin: In the old format it's like two sorted lists appended to each other, first consisting of valid ones, second of invalid ones (e.g. something like `v:1, v:2, v:4, inv:1, inv:3, inv:4`). We handled the valid half above, now it's the invalid half. For which we need to check if a version already exists from the valid ones -> add it there, or if it's a new one, we need to get the correct position (e.g. in the given example, the position of `inv:3` isn't clear).

calmh: But do we really need the fileinfo to determine the ordering? The conflict winner wasn't completely known with the previous order either, so could we not let that still be the case?

That would mean the migration is a simple aggregation on version (keep an index along the way) followed by a sort of the versions on version vector only

imsodin: We did (and still do) determine the conflict winner on insert to the global list, i.e. when versions are one of the two concurent values, get the files and call `WinsConflict` on them. Otherwise the puller would get the wrong file.

calmh: I'm feeling dense here. Given we did/do that, the entries should already be in the right order in the old version list? The part I'm not getting is why we need more information than the old version list in order to determine the order in the new version list.

imsodin: Yeah right, I densely didn't consider that if `b.IsInvalid()` then `protocol.WinsConflict(a, b) == true`, and we do know whether a and b are (in-)valid - will simplify.  
If of interest: The perceived, incorrect problem was that the old version list consisted of two lists ordered by version (one for valid, one for invalid devices), while the new one is a single ordered list. With an example: Lets say the old list is `[{valid, A:2}, {valid, A:1}, {invalid, B:1}] `, now where does the third version go? It's not clear from the ordering itself, as it's a conflict, but it is considering it is invalid...

lib/db/schemaupdater.go
imsodin: Done.

lib/db/schemaupdater.go
```diff
@@ -715,7 +715,7 @@ func (db *schemaUpdater) rewriteGlobals(t readWriteTransaction) error {
 	return it.Error()
 }
 
-func convertVersionList(vl VersionListDeprecated, gk []byte, t readOnlyTransaction) (VersionList, error) {
+func convertVersionList(vl VersionListDeprecated, gk []byte) (VersionList, error) {
```
calmh, line 718: gk unusued?

lib/db/transactions.go
```diff
@@ -319,20 +341,14 @@ func (t *readOnlyTransaction) withGlobal(folder, prefix []byte, truncate bool, f
 		if err := vl.Unmarshal(dbi.Value()); err != nil {
 			return err
 		}
+		l.Infoln("withGlobal", string(name))
```
calmh, line 344: debugging

lib/db/schemaupdater.go
```diff
@@ -629,5 +670,275 @@ func (db *schemaUpdater) updateSchemaTo12(_ int) error {
 		return err
 	}
 
+	if err := db.rewriteGlobals(t); err != nil {
```
calmh, line 673: I think we should make this a version 13 migration. Even though 12 is not out, people might be running master (I was...) and the effect then is that we skip this part of the migration now and checkGlobals deletes all the globals because they don't unmarshal correctly and then weirdness ensues.

imsodin: Done. And now doing the same - well, running ahead of master :)

imsodin:
> Yes?

Almost: I took another look myself and found some stuff I think deserves a comment, one rebase artefact (needless `if`) and style choices that made sense in earlier revisions but not any more (`popVersion` et al) - all polishing only luckily.

lib/db/structs.go
```diff
@@ -365,6 +370,32 @@ func (vl VersionList) findDevice(device []byte) (bool, int, int, bool) {
...
+func (vl VersionList) versionIndex(version protocol.Vector) int {
+	for i, v := range vl.RawVersions {
+		if version.Equal(v.Version) {
+			return i
+		}
+	}
+	return -1
+}
+
+func (vl VersionList) popVersionAt(i int) VersionList {
```
calmh, line 391: This (and maybe the others) feel like they should be pointer methods, as it modifies the receiver. I didn't check if that's against expectations anywhere (I think not) but it otherwise looks like a value method which it isn't because the receiver is actually invalid after the pop.

imsodin: I honestly just blindly followed how it previously worked, and it was always in the style of `(vl VersionList) update(...) VersionList`, i.e. value receiver and return the value. Changing to pointer receiver seems sensible, but then there's no point in returning the receiver - or am I missing something here?

lib/db/structs.go
```diff
@@ -192,102 +191,321 @@ func (c Counts) Equal(o Counts) bool {
 	return c.Files == o.Files && c.Directories == o.Directories && c.Symlinks == o.Symlinks && c.Deleted == o.Deleted && c.Bytes == o.Bytes
 }
 
-func (vl VersionList) String() string {
+func (vl *VersionList) String() string {
```
calmh, line 194:
```suggestion
func (vl VersionList) String() string {
```

PR #6639, opened 2020-05-13
lib/db: Dont add symlinks to blocks map (fixes #6637)

### LINKED ISSUE #6637: Error: Found something of unexpected type in block list map
#### Does your log mention database corruption?
Maybe (see log)

#### Info
 - Arch Linux x86_64, Syncthing v1.6.0-rc.1

#### What happened
 - after updating the version of syncthing the following error / warning was shown:
```
May 13 17:06:25 moritz syncthing[530]: [MP6JR] WARNING: Found something of unexpected type in block list map: Symlink{Name:"qq0s0464.default-nightly/lock", Type:SYMLINK, Sequence:2266, Version:{[{MP6JR34 1589311222}]}, Deleted:false, Invalid:false, LocalFlags:0x0, NoPermissions:true, SymlinkTarget:"127.0.0.1:+56071"}
May 13 17:08:06 moritz syncthing[530]: [MP6JR] WARNING: Non-increasing sequence detected: Checking and repairing the db...
May 13 17:08:06 moritz syncthing[530]: [MP6JR] INFO: Repaired 1 sequence entries in database
```

#### Expected
no error / warning

#### How to reproduce

I don't know...

imsodin:
The source of the problem is in lowlevel.go

I also have a variant for the release branch - it's a bit annoying due to the db schema numbers.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -201,7 +201,7 @@ func (db *Lowlevel) updateLocalFiles(folder []byte, fs []protocol.FileInfo, meta
 		blocksHashSame := ok && bytes.Equal(ef.BlocksHash, f.BlocksHash)
 
 		if ok {
-			if !ef.IsDirectory() && !ef.IsDeleted() && !ef.IsInvalid() {
+			if len(ef.Blocks) != 0 {
 				for _, block := range ef.Blocks {
 					keyBuf, err = db.keyer.GenerateBlockMapKey(keyBuf, folder, block.Hash, name)
 					if err != nil {
@@ -262,7 +262,7 @@ func (db *Lowlevel) updateLocalFiles(folder []byte, fs []protocol.FileInfo, meta
 		}
 		l.Debugf("adding sequence; folder=%q sequence=%v %v", folder, f.Sequence, f.Name)
 
-		if !f.IsDirectory() && !f.IsDeleted() && !f.IsInvalid() {
+		if len(f.Blocks) != 0 {
 			for i, block := range f.Blocks {
 				binary.BigEndian.PutUint32(blockBuf, uint32(i))
 				keyBuf, err = db.keyer.GenerateBlockMapKey(keyBuf, folder, block.Hash, name)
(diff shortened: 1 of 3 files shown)


lib/db/schemaupdater.go
```diff
@@ -590,7 +591,7 @@ func (db *schemaUpdater) updateSchemaTo11(_ int) error {
 		var putErr error
 		err := t.withHave(folder, protocol.LocalDeviceID[:], nil, true, func(fi FileIntf) bool {
 			f := fi.(FileInfoTruncated)
-			if f.IsDirectory() || f.IsDeleted() || f.IsInvalid() || f.BlocksHash == nil {
```
calmh, line 593: This should have already been caught by the BlocksHash==nil? Maybe a len(BlocksHash)==0 instead which is less fragile?

imsodin: Yes this is indeed not an actual problem, just changed it for completeness.

imsodin:
Actually we could also live quite well without the transition and without silencing: The new check in folder.go prevents the warning too. The affected RC users will just have a bit of garbage in their db that will stay there, but it shouldn't hurt. Does that make any sense, and if yes, should I do that (i.e. remove the transition).

lib/db/schemaupdater.go
calmh: Also a len(Blocks)==0 for good measure

calmh:
Yes remove the migration

lib/db/schemaupdater.go
imsodin: Missed this: It's a `FileInfoTruncated` - no blocks there.

imsodin:
@AudriusButkevicius Could you have a look and do the honours?

lib/db/lowlevel.go
```diff
@@ -201,7 +201,7 @@ func (db *Lowlevel) updateLocalFiles(folder []byte, fs []protocol.FileInfo, meta
 		blocksHashSame := ok && bytes.Equal(ef.BlocksHash, f.BlocksHash)
 
 		if ok {
-			if !ef.IsDirectory() && !ef.IsDeleted() && !ef.IsInvalid() {
+			if len(ef.Blocks) != 0 {
```
AudriusButkevicius, line 204: Can't something invalid (like a locally modified receive only file or something) not have blocks?

calmh: It's not supposed to, but keeping the IsInvalid() check might make sense

AudriusButkevicius:
On the phone, but I assume the blocks get unset and redirected deeper than updatelocal?

imsodin:
Yes.

PR #6645, opened 2020-05-14
Dir junctions

xarx00:
### Purpose

Fix for an issue found in https://github.com/syncthing/syncthing/pull/6606:

When directory junctions are synchronized from Windows to Synology NAS, the Syncthing Folder at Windows must have Ignore Permissions checked. Otherwise, the directories at NAS corresponding to the directory junctions do not have the Traverse Folders permission set. And their content is not synchronized due to permission problems.

This pull request fixes this issue.

### Testing

Testing performed in live environment by creating a directory junction and checking that it has synchronized with correct privileges.

## Authorship

based on the commit metadata.


calmh:
Right. Comment please that this sets the execute bits which are required unix-side and otherwise not present. Just the `0111` is a bit magic.

lib/fs/lstat_windows.go
```diff
@@ -57,7 +57,7 @@ type dirJunctFileInfo struct {
 }
 
 func (fi *dirJunctFileInfo) Mode() os.FileMode {
-	return fi.FileInfo.Mode() ^ os.ModeSymlink | os.ModeDir
+	return fi.FileInfo.Mode() ^ os.ModeSymlink | os.ModeDir | 0111
```
calmh, line 60:
```suggestion
        // Simulate a directory and not a symlink; also set the execute
        // bits so the directory can be traversed Unix-side.
	return fi.FileInfo.Mode() ^ os.ModeSymlink | os.ModeDir | 0111
```

xarx00:
In fact, there are many more problems when syncing from Windows to NAS that are NOT related to directory junctions. I'll open a discussion thread for that in the forum.

PR #6647, opened 2020-05-14
lib/model: Fix checking children when trying to delete a dir (fixes #6646)

### LINKED ISSUE #6646: Misleading error message when to be deleted dir contains receive-only changes
https://forum.syncthing.net/t/problems-when-syncing-from-windows-to-synology-nas/14996

When a directory deleted on a remote contains receive-only files, the following error message is given:

> directory has been deleted on a remote device but contains unexpected files, scheduling scan

Receive-only files aren't unexpected and rescanning won't help.

imsodin:
https://forum.syncthing.net/t/problems-when-syncing-from-windows-to-synology-nas/14996

When a directory deleted on a remote contains receive-only files, rescanning won't help and the current error message wont direct the user at the actual problem (locally changed files).

Also the scan chan cannot be nil, that might be true in tests, but not everywhere else (we send to it without checking for nil-ness in various other places). And if it is nil in tests, that should be changed instead of having weird code paths like this.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -61,17 +61,18 @@ type copyBlocksState struct {
 const retainBits = fs.ModeSetgid | fs.ModeSetuid | fs.ModeSticky
 
 var (
-	activity                  = newDeviceActivity()
-	errNoDevice               = errors.New("peers who had this file went away, or the file has changed while syncing. will retry later")
-	errDirPrefix              = "directory has been deleted on a remote device but "
-	errDirHasToBeScanned      = errors.New(errDirPrefix + "contains unexpected files, scheduling scan")
-	errDirHasIgnored          = errors.New(errDirPrefix + "contains ignored files (see ignore documentation for (?d) prefix)")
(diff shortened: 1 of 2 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -1825,14 +1824,13 @@ func (f *sendReceiveFolder) deleteDirOnDisk(dir string, snap *db.Snapshot, scanC
 		} else if f.ignores != nil && f.ignores.Match(fullDirFile).IsIgnored() {
 			hasIgnored = true
 		} else if cf, ok := snap.Get(protocol.LocalDeviceID, fullDirFile); !ok || cf.IsDeleted() || cf.IsInvalid() {
+			if f.Type == config.FolderTypeReceiveOnly && cf.IsReceiveOnlyChanged() {
+				hasReceiveOnlyChanged = true
+			}
```
AudriusButkevicius, line 1829: doesn't this need an else?

imsodin:
While I am addressing one deficiency of that code, I might as well fix it right:  
Just checking `IsInvalid()` makes no sense at this point. What we should check (and now do) is if the file on disk is the same as in db, and if not rescan.

PR #6648, opened 2020-05-14
docker: Ensure correct ownership of Syncthing base and config directory

jensrudolf:
This commit changes the docker entrypoint script to change the ownership
of the Syncthing home dir recursively. This allows e.g. the config dir
to be bind-mounted from docker host.

### Purpose
The purpose of the change is to allow the syncthing config dir (e.g. /var/syncthing/config) to be bind-mounted from the host. A scenario, I use at my company is to have /var/syncthing as a persistent named volume that contains the database and bind mount the config directory from the host. This allows e.g. to deploy the configuration with Ansible. However, to allow Syncthing to read, write and chmod the config files, the ownership change in the entrypoint script must be applied recursively to the syncthing home dir.

### Testing
Tested this with bind mount of host dir /etc/syncthing into /var/syncthing/config

### DIFF
--- a/lib/fs/lstat_windows.go
+++ b/lib/fs/lstat_windows.go
@@ -57,7 +57,9 @@ type dirJunctFileInfo struct {
 }
 
 func (fi *dirJunctFileInfo) Mode() os.FileMode {
-	return fi.FileInfo.Mode() ^ os.ModeSymlink | os.ModeDir
+	// Simulate a directory and not a symlink; also set the execute
+	// bits so the directory can be traversed Unix-side.
+	return fi.FileInfo.Mode() ^ os.ModeSymlink | os.ModeDir | 0111
 }
 
 func (fi *dirJunctFileInfo) IsDir() bool {
--- a/script/docker-entrypoint.sh
+++ b/script/docker-entrypoint.sh
@@ -2,6 +2,6 @@
 
 set -eu
 
-chown "${PUID}:${PGID}" "${HOME}" \
+chown -R "${PUID}:${PGID}" "${HOME}" \
   && exec su-exec "${PUID}:${PGID}" \
      env HOME="$HOME" "$@"
--- a/script/docker-entrypoint.sh
+++ b/script/docker-entrypoint.sh
@@ -2,6 +2,8 @@
 
 set -eu
 
-chown -R "${PUID}:${PGID}" "/var/syncthing/config" || true \
-   && exec su-exec "${PUID}:${PGID}" \
+mkdir -p "/var/syncthing/config" \
+  && chown "${PUID}:${PGID}" "/var/syncthing" \
+  && chown -R "${PUID}:${PGID}" "/var/syncthing/config" \
+  && exec su-exec "${PUID}:${PGID}" \
      env HOME="$HOME" "$@"
--- a/script/docker-entrypoint.sh
+++ b/script/docker-entrypoint.sh
@@ -2,6 +2,6 @@
 
 set -eu
 
-chown -R "${PUID}:${PGID}" "${HOME}" \
-  && exec su-exec "${PUID}:${PGID}" \
+chown -R "${PUID}:${PGID}" "/var/syncthing/config" || true \
+   && exec su-exec "${PUID}:${PGID}" \
      env HOME="$HOME" "$@"


AudriusButkevicius:
Why can't ansible fix the permissions?

jensrudolf:
Ansible (or any other mechanism on the host side) would have to find out about the host mapping of the Syncthing UID used in the container. E.g. What host UID is mapped to the UID 1000 (or whatever) used in the container. While I think this is technically possible, it would add unnecessary complexity to the deployment. You actually don't know the UID before the container is even created. This proposed change to solve this from within the entrypoint script is rather simple in my view.
Furthermore, looking at the entrypoint scripts of other container (e.g. Redis), you'll find them using a similar approach (using the 'find' command, though).

calmh:
UIDs are UIDs, no? 1000 in the container will be 1000 on the host, and you can either trust us to have this be 1000 forever or you can set it yourself using the PUID/PGID variables to whatever you want it to be.

AudriusButkevicius:
Agree, it would be strange if I mounted some directory into a container and it messed up the permissions.

jensrudolf:
Hm, from my observation, syncthing requires a certain access mode applied to its config dir and config files (keys and xml). On startup, it changes it to 0700 (at least the directory). Therefore, it has to be the owner of the config dir, otherwise the startup would fail. At least this is my observation. Wouldn't it be in the responsibility of the entrypoint script, which is launching syncthing, to ensure that the required ownership is correctly set at least for the config dir and its config files?
I agree, that chown'ing recursively everything down from /var/syncthing might be an overkill, but I would like to ensure correct ownership of the config dir and files from within the container.

AudriusButkevicius:
We don't tweak the permissions of the config explicitly, we probably inherit 700 from the umask when we modify it.

jensrudolf:
OK, agreed. But I think that requires an additional check for presence of the config dir as it is not part of the image and only exists in case one mounts it from the outside.

calmh:
Yeah I think it's fair for the container to take more explicit ownership of the config/database dir. I'd go so far as to say that I think the intention is for the /var/syncthing directory in the container to actually hold just the config & database, and for every synced folder to be mapped in as other volumes, really.

jensrudolf:
OK, I'd like to make some modifications then, to make this more sane and really only touch config and database.

jensrudolf:
I modified the `chown` command to change only the ownership of `/var/syncthing/config` and its contents. This directory is currently hard-coded in the Dockerfile entrypoint statement and is set to contain the database and config (using the `-home` switch). What do you think?

AudriusButkevicius:
This now ignores environment variables passed to the container.

jensrudolf:
Well, actually I am wondering, why the entrypoint statement uses `/var/syncthing/config` hardcoded but then chowned $HOME, which could be set to any other directory irrelevant to Syncthing, which would still use /var/syncthing/config. Do I understand you correct, that $HOME should set the Syncthing config dir via `-home` switch and not the default sharing directory (`/var/syncthing`)?

AudriusButkevicius:
I have no idea. I think it's probably an oversight? But at the same time I can see how things would break for people who already rely on this.

calmh:
We create a volume at `/var/syncthing`. This is mainly to hold Syncthing's config and database, but we also need a default for synced folders. Hence we set `$HOME` to `/var/syncthing`, and make sure we have write permissions there. We use `/var/syncthing/config` explicitly for the config/database dir because it would otherwise default to `/var/syncthing/.config/syncthing` which is ugly and unexpected.

The currently suggested change is not great.

```
% docker build -t st .
...
Successfully built d1cdd3942ac9
Successfully tagged st:latest
jb@kvin:../syncthing/syncthing % docker run -it --rm st
chown: /var/syncthing/config: No such file or directory
12:50:13 WARNING: Failure on home directory: mkdir /var/syncthing/config: permission denied
```

jensrudolf:
Please help me understand things right in order to propose a proper change. This is how I understand the current image setup:
- $HOME is supposed to be the default directory for shares. It defaults to `/var/syncthing` but can be adapted through the env var at container creation time to point to somewhere else. In order to add shares from within syncthing, this directory should be writetable for $UID (syncthing) but syncthing will startup properly without write permission to $HOME and the user might choose other (writable) directories to create shares.
- `/var/syncthing` (hardcoded) is intended to be used as volume but is not necessarily the same as $HOME
- `/var/syncthing/config` (hardcoded) is used as location for database and config but is not nessarily the same as $HOME/config
- In order to startup properly, the following requirements must be met:
  - be able to create `/var/syncthing/config` in case it does not exists, which is the default condition in the image. Therefore, $UID needs write access to `/var/syncthing`
  - in case `/var/syncthing/config` already exists (that would be our use case as we mount it from the host), $UID must be the owner to be able to set the correct permissions on config xml, certs and database

I would like to ensure that both of the above mentioned requirements are met from withing the container (through the entrypoint script) and not from outside as syncthings requirements to file permissions might change in future and that would require all users to implement changes in the deployment mechanisms.

Please correct me, if I made a mistake in any of my assumptions.

calmh:
I think your assumptions sound right, but if it turns out I’m wrong it’s probably because I’ve had a couple of glasses of wine and read this on my phone. 

Error with custom home looks like a container bug. I think the container should ensure usable ownership on /var/syncthing and in general treat that place like it owns it. HOME may be set wherever but then it’s the users responsibility to see that it’s usable by Syncthing as required. 

I don’t remember what the original issue was. But it seems like the container should ensure /var/syncthing and /var/syncthing/config (if it exists) are writable by the current uid.

jensrudolf:
Thanks for your feedback. I made another change to my pull request in order implement the expected behavior. My last change ensures existence of `/var/syncthing/config` as well as the required ownership of $UID and $GID for `/var/syncthing`, `/var/syncthing/config` and its contents.

If $HOME is set to a custom directory, the user has to ensure its proper permissions.

I think the last change is more conservative than my first attempt and should not mess up with existing setups. It is now possible to mount in the config directory from the docker host without setting the required permissions from outside. Besides, it fixes the startup problem when $HOME is set to a custom directory.
What do you think?

imsodin:
I don't think we should mix `$HOME` and hard-coded `/var/syncthing`. We set `$HOME` to `/var/syncthing` in the dockerfile so I'd use `$HOME` everywhere.

Unrelated: I think all the docker stuff should be in `docker` (or `etc/docker`, doesn't fit entirely though with what's there now). Having multiple files In the root dir and this one file in `script/` feels messy.

AudriusButkevicius:
> Unrelated: I think all the docker stuff should be in docker (or etc/docker, doesn't fit entirely though with what's there now). Having multiple files In the root dir and this one file in script/ feels messy.

I don't think that works with dockerhub.

st-review:
:robot: *beep boop*

PR #6649, opened 2020-05-15
Fixed Docker health check for host networks (fixes #6604)

### LINKED ISSUE #6604: Docker healthcheck fails when run in host network

shaaraddalvi:
Fixed the health check instead of removing it for host networks as per @thedanbob's suggestions. Verified that the health check works for both host as well as non-host network configurations.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,5 +22,8 @@ COPY --from=builder /src/script/docker-entrypoint.sh /bin/entrypoint.sh
 
 ENV PUID=1000 PGID=1000 HOME=/var/syncthing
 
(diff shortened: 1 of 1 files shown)


shaaraddalvi:
Thanks for quick turnaround @calmh!

Issue #6650, opened 2020-05-15
panic: sequence index corruption


mohe2015:
### Does your log mention database corruption?
yes

I don't know but I don't think this has something to do with hardware corruption.

### Include required information

Please be sure to include at least:

 - Arch Linux, Syncthing v1.6.0-rc.2+11-gf05ccd775, f05ccd775abf1a51a39b0363d72fe7942815e01b

Info: I'm also https://forum.syncthing.net/t/up-to-date-but-not-syncing/14998 and I also reported https://github.com/syncthing/syncthing/issues/6637

### what happened
Some time after I checked e-mails with the latest Thunderbird release (I don't know when exactly) syncthing paniced but only at the laptop side not at the raspberry pi 3b. I think it also had the problem outlined in my forum post before but I'm not sure any more.

Logs: https://gist.github.com/mohe2015/126aa0fb81fe4a633fd829e1417ba1bc
More data can be provided on request. I won't reset the database for now (systemctls automatic restarts failed with the same panic) so I could also provide that if that can help.

### what you expected to happen instead
obviously not crashing

### any steps to reproduce the problem.
don't know, it was a fresh database with multiple gigabytes of data on two devices one of which is a laptop with an ssd and the other one is a raspberry pi 3b with a 64gb usb flash drive.

It could be possible that backing up .thunderbird may exhibit the problem.

calmh:
The panic/crash is a debug feature caused by the debug tracing you've enabled. Without that it would try to just heal the inconsistency. That said, there is an inconsistency and there shouldn't be...

imsodin:
Wow. Thank you!

I don't yet have a clue why it's happening, but what is happening is that in a single batch of local file updates, we get the same file twice. We then twice remove the original file from before the update, and add the first and second file.

I do believe it's a bug that we get the same file twice in one batch, but I also think we should handle that case. Because for remote devices I think that can happen legitimately, so we should be able handle it in db.

AudriusButkevicius:
So the same file twice could be done by rename detection, as it makes the file as deleted twice most likely.

calmh:
Sounds plausible, the file is 0 bytes long so likely match for any other created 0 byte long file...

I guess the problem is our non-dirty-read transactions. We create a new item and sequence entry, dropping the old sequence entry. Then we get the next instance, create a new item and sequence entry but can't find the intermediate sequence entry we created.

imsodin:
[...]

PR #6651, opened 2020-05-15
lib/db: Filter repeat files in one update (ref #6650)

imsodin:
### Purpose

From the code:
```
	// If one file info is present multiple times, only keep the last.
	// Updating the same file multiple times is problematic, because the
	// previous updates won't yet be represented in the db when we update it
	// again. Additionally even if that problem was taken care of, it would
	// be pointless because we remove the previously added file info again
	// right away.
```

### Testing

Added a unit test updating the same file twice in one update, then checking that it's not present twice in the sequence bucket afterwards (it is now).

### DIFF
--- a/lib/db/db_test.go
+++ b/lib/db/db_test.go
@@ -568,14 +568,16 @@ func TestDropDuplicates(t *testing.T) {
 		{[]int{0, 1, 1, 1, 1}, []int{0, 1}},
 		{[]int{0, 0, 0, 1}, []int{0, 1}},
 		{[]int{0, 1, 2, 3}, []int{0, 1, 2, 3}},
-		{[]int{0, 1, 2, 3, 0, 1, 2, 3}, []int{0, 1, 2, 3}},
+		{[]int{3, 2, 1, 0, 0, 1, 2, 3}, []int{0, 1, 2, 3}},
 		{[]int{0, 1, 1, 3, 0, 1, 0, 1, 2, 3}, []int{0, 1, 2, 3}},
 	}
 
 	for tci, tc := range tcs {
 		inp := make([]protocol.FileInfo, len(tc.in))
+		expSeq := make(map[string]int)
 		for i, j := range tc.in {
-			inp[i] = protocol.FileInfo{Name: names[j]}
+			inp[i] = protocol.FileInfo{Name: names[j], Sequence: int64(i)}
+			expSeq[names[j]] = i
 		}
 		outp := normalizeFilenamesAndDropDuplicates(inp)
 		if len(outp) != len(tc.out) {
@@ -585,6 +587,9 @@ func TestDropDuplicates(t *testing.T) {
 		for i, f := range outp {
 			if exp := names[tc.out[i]]; exp != f.Name {
 				t.Errorf("tc %v: Got file %v at pos %v, expected %v", tci, f.Name, i, exp)
+			}
+			if exp := int64(expSeq[outp[i].Name]); exp != f.Sequence {
+				t.Errorf("tc %v: Got sequence %v at pos %v, expected %v", tci, f.Sequence, i, exp)
 			}
 		}
 	}
--- a/lib/db/db_test.go
+++ b/lib/db/db_test.go
@@ -553,3 +553,39 @@ func TestUpdateTo10(t *testing.T) {
 		t.Error("vl.Versions[1] not deleted for c")
 	}
 }
+
+func TestDropDuplicates(t *testing.T) {
+	names := []string{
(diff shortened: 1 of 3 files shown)


lib/db/set.go
```diff
@@ -470,10 +476,19 @@ func DropDeltaIndexIDs(db *Lowlevel) {
...
-func normalizeFilenames(fs []protocol.FileInfo) {
-	for i := range fs {
-		fs[i].Name = osutil.NormalizedFilename(fs[i].Name)
+func normalizeFilenamesAndDropDuplicates(fs []protocol.FileInfo) []protocol.FileInfo {
+	positions := make(map[string]int, len(fs))
+	newFs := fs[:0]
+	for _, f := range fs {
+		norm := osutil.NormalizedFilename(f.Name)
+		if i, ok := positions[norm]; ok {
+			newFs = append(newFs[:i], newFs[i+1:]...)
```
calmh, line 485: Problem: this changes the positions of everything after i, so the map is now wrong for those items. Maybe replace the item to be dropped with a sentinel value, and then filter these in a latter stage?

calmh: (And try to make a couple of nasty test cases for just this function?)

calmh:
Pretty sure this still suffers from the same or similar problem. Highly recommend a test case like (simpligied) `{"dup", "foo", "dup", "bar", "dup", "baz", "dup"} => {"foo", "bar", "baz"}` and similar.

lib/db/db_test.go
```diff
@@ -553,3 +553,39 @@ func TestUpdateTo10(t *testing.T) {
...
+	}
+	tcs := []struct{ in, out []int }{
+		{[]int{0}, []int{0}},
+		{[]int{0, 1}, []int{0, 1}},
+		{[]int{0, 1, 0, 1}, []int{0, 1}},
+		{[]int{0, 1, 1, 1, 1}, []int{0, 1}},
+		{[]int{0, 0, 0, 1}, []int{0, 1}},
+		{[]int{0, 1, 2, 3}, []int{0, 1, 2, 3}},
+		{[]int{0, 1, 2, 3, 0, 1, 2, 3}, []int{0, 1, 2, 3}},
+		{[]int{0, 1, 1, 3, 0, 1, 0, 1, 2, 3}, []int{0, 1, 3, 2}},
```
calmh, line 572:
```suggestion
		{[]int{3, 2, 1, 0, 0, 1, 2, 3}, []int{0, 1, 2, 3}},
		{[]int{0, 1, 1, 3, 0, 1, 0, 1, 2, 3}, []int{0, 1, 2, 3}},
```

calmh: Ideally we'd also want to check that it's the correct instance of the FileInfo that's kept, not just any one of the equivalents...

AudriusButkevicius:
Just a note, this makes the whole "what should we put in the db in case of rename" argument moot, as that is likely to get tramped.

calmh:
Only to the extent that that was already the case, I think?

calmh:
I mean yeah, we might drop your early delete notification, but only if we have the same delete in the same batch so it should be roughly equivalent

AudriusButkevicius:
I am not sure if this change is good though, I'd rather panic, than drop something that I wanted to end up in a database. Adding something has various side-effects to the blockmap and what not, so feels we're opening a door to even more inconsistency.

imsodin:
Actually I think neither is the original reported issue caused by the new rename detection nor is this PR a problem for rename detection: We flush the batch and acquire a new db snapshot between scanning/detecting renames and walking the db to detect deletions. 

And we can't panic in general: For remotes it's totally legitimate to get a file multiple times at once, if fast changes after each other are batched together.

Adding and removing and readding a file shouldn't have side-effects. I am not saying it doesn't, but I can't think of any and if there are I'd expect problems regardless of this change or not. An alternative would be to track the last added file in the actual update code, and then use that as the "previous" file instead of getting it from db.

AudriusButkevicius:
The rename detection could mark the same file as deleted twice, just because the hash matches...

imsodin:
Ah, right, like that it can indeed. But that's not a problem for the ordering, right: The deletion is just going to be close to one of the multiple files it matches.

AudriusButkevicius:
Not one, potentially multiple.

calmh:
I don’t see the problem here. Assuming this didn’t cause the sequencing bug the we solve by dropping the earlier of the two updates, that update would still have been invisible and with no effect since it was overridden before the batch was committed. This patch should have no visible effect on the state of the database, except for not fucking up the sequences.

PR #6656, opened 2020-05-16
lib/model: Partial revert of rename fix (fixes #6653)

### LINKED ISSUE #6653: panic: nil pointer dereference in leveldb.(*DB).isClosed()
No

Please be sure to include at least:

 - Syncthing https://github.com/syncthing/syncthing/commit/5c9dc4c8831566d1ff2ad8cf4047fbd02aff8867 (master)

# What I did
- reset everything on this side, reset database on other side, tried adding folders for this side

# What happened
Panic. https://gist.github.com/mohe2015/2db012669030f3b7c8a8f28f8ed0f445

# Steps to reproduce
I don't know I think the current master is just broken.

calmh:
We can't just drop the snap because it's in use elsewhere. This should
be equally functional.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -460,13 +460,8 @@ func (f *folder) scanSubdirs(subDirs []string) error {
 			continue
 		}
 
-		if batch.full() {
-			if err := batch.flush(); err != nil {
-				return err
-			}
-			snap.Release()
-			snap = f.fset.Snapshot()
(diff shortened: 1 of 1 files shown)


calmh:
Fast-tracking as we have an RC out that can't scan.

AudriusButkevicius:
Thanks for fixing this.

PR #6661, opened 2020-05-17
lib/assets: Allow assets to remain uncompressed

greatroar:
All compiled-in assets are currently compressed, but six of them become larger because of this, so the compression/decompression step is a waste of time and space: two WOFF fonts, three PNG images and a tiny JavaScript file.

This PR adds support for uncompressed assets. The generator script decides on the format depending on whether compression makes it bigger or smaller. Also, the compression level is set to BestCompression. The binary gets 4kB smaller on Linux/amd64 as a result.


lib/assets/assets.go
```diff
@@ -53,19 +53,28 @@ func Serve(w http.ResponseWriter, r *http.Request, asset Asset) {
...
 		var gr *gzip.Reader
-		gr, _ = gzip.NewReader(strings.NewReader(asset.ContentGz))
+		gr, _ = gzip.NewReader(strings.NewReader(asset.Content))
 		io.Copy(w, gr)
 		gr.Close()
 	}
 }
 
+func gzipped(content string) bool {
+	const gzmagic = "\x1f\x8b" // RFC 1952, §2.3.1.
```
AudriusButkevicius, line 74: We should just have a flag/enum for the compression used, if any.

calmh: At which point we can also add an `length int` and serve that as `Content-Length` to clients not speaking gzip.

calmh:
> The binary gets 4kB smaller on Linux/amd64 as a result.

That's almost 0.02% percent of an improvement!

greatroar:
Yeah, this is a nano-optimization in terms of space. Skipping the decompression in the browser and giving it a proper Content-Length are more important than the space saving, but those effects are much harder to measure.

Btw., I've toyed around with [go-zopfli](https://github.com/foobaz/go-zopfli) and it can shave off a few percent.

lib/assets/assets.go
greatroar: That would be nicer, but the code for it is much more complicated. The generated code and the client in lib/assets need some common type to communicate bool+string.

calmh: The `Asset` struct?

greatroar: Yes. Turned out to be easier than I thought, see updated code.

greatroar:
Looks like the failing tests are due to go generate not being run. They work fine on my machine.

lib/assets/assets.go
```diff
@@ -53,14 +55,18 @@ func Serve(w http.ResponseWriter, r *http.Request, asset Asset) {
...
+
+	switch {
+	case !asset.Gzipped:
+		io.WriteString(w, asset.Content)
+	case strings.Contains(r.Header.Get("Accept-Encoding"), "gzip"):
 		header.Set("Content-Encoding", "gzip")
-		header.Set("Content-Length", strconv.Itoa(len(asset.ContentGz)))
-		io.WriteString(w, asset.ContentGz)
-	} else {
+		io.WriteString(w, asset.Content)
```
AudriusButkevicius, line 65: Content-length gone?

calmh: It's above

calmh:
Test failures seem genuine though

```
# github.com/syncthing/syncthing/lib/api [github.com/syncthing/syncthing/lib/api.test]
lib/api/api_test.go:167:3: cannot use map[string]string literal (type map[string]string) as type map[string]assets.Asset in field value
FAIL	github.com/syncthing/syncthing/lib/api [build failed]

--- FAIL: TestServeGzip (0.00s)
    assets_test.go:86: unexpected content "\x1f\x8b\b\x00\x00\x00\x00\x00\x00\xff\xb2\xc9(\xc9ͱ\xf3H\xcd\xc9\xc9\xd7Q(\xcf/\xcaIQ\xb4\xd1\a\x8b\x01\x02\x00\x00\xff\xff/\xa3B'\x1a\x00\x00\x00"
FAIL
```

lib/assets/assets.go
calmh: Wait. We're now sending the length of the uncompressed asset with the compressed data? does that work?

greatroar: No. Sorry, I was mistaken. It turns out [Content-Length is the encoded length](https://stackoverflow.com/a/3819303).

lib/assets/assets.go
```diff
@@ -53,14 +54,18 @@ func Serve(w http.ResponseWriter, r *http.Request, asset Asset) {
...
+	default:
 		// gunzip for browsers that don't want gzip.
 		var gr *gzip.Reader
-		gr, _ = gzip.NewReader(strings.NewReader(asset.ContentGz))
+		gr, _ = gzip.NewReader(strings.NewReader(asset.Content))
```
calmh, line 68: wrong content-length here?

greatroar: Yes. Should be fixed now, and I added an assertion to the test.

PR #6673, opened 2020-05-21
lib: Add encryption token and folder type

imsodin:
This is a PR for https://github.com/syncthing/syncthing/pull/6214, not master.

### Purpose

This adds a token to the cluster config to indicate "encryptionness": A token is added to a device, either when you encrypt for that device, or for yourself, if you are encrypted. The other side checks that it gets the tokens it expects and drops the connection otherwise, to avoid creating multiple sets of encrypted data with different passwords.

There's also a new folder type for encrypted folders, which behaves like a send-receive folder, except that on detecting changes, it nukes those on disk and sets an empty version vector on it's own file, such that it then pulls in the global file again.

What is still missing:

 - Any user facing info (except for errors/logging): Indicate when a folder should be encrypted, when there's a password mismatch.
 - Folder introduction: Both manual in web UI and auto-accept should correctly suggest/set encrypted folder type.
 - Folder marker has to be a directory to store the encryption token. Could be an error, could be converted on the fly for `.stfolder` (what about custom markers then?).

 Companion doc PR: https://github.com/syncthing/docs/pull/519

### Testing

None whatsoever yet.


imsodin:
This is ready for review now.

From the missing items above, I haven't done anything about indicating encryption in the UI yet. Password mismatches are signalled with warnings, so that's already in place.

calmh:
There was just one thing that strikes me, that this goes to some length to ensure it can write inside the folder marker. While I think it's safe to assume that an encrypted folder is not used also for normal files, and won't have a custom marker. We create a bunch of directories at the top level with magic names for the encrypted files. So we might as well drop the token in a specific name at the top level, or forbid changing the folder marker, or even use a non-standard marker name always just to indicate it's not a normal folder... Just thinking out loud

lib/model/model.go
```diff
@@ -999,6 +1038,36 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
 			continue
 		}
 
+		ccDevice, hasDevice := ccDevices[folder.ID]
+		ccDeviceUs, hasDeviceUs := ccDevices[folder.ID]
```
calmh, line 1042: missing an `us`?

lib/model/model.go
```diff
@@ -1132,6 +1200,82 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
...
+
+	if !(isEncDev || isEncUs) {
+		return errEncNotEncryptedUs
+	}
+
+	if isEncDev {
+		pwToken := protocol.PasswordToken(fcfg.ID, folderDevice.EncryptionPassword)
+		match := false
+		if hasTokenUs {
+			match = !bytes.Equal(pwToken, ccDeviceUs.EncPwToken)
```
calmh, line 1232: Why is this not-equals?

imsodin:
Writing tests for this, I wondered whether the current auto-accept policy needs changing: I now reject new folder which are encrypted. Reason being there's two choices there: Add PW to decrypt the folder, or add the folder as encrypted. However it might be useful to set a device to auto-accept and then automatically add the offered encrypted folder as an encrypted folder. So essentially if you expect a device to offer encrypted folders you want to decrypt, you shouldn't tick auto-accept.  
So I am leaning towards auto-accepting new, encrypted folders instead of rejecting them and requiring manual intervention - opinions?

imsodin:
I changed the auto-accept policy to accept new folders which are encrypted automatically as a locally encrypted folder.

Writing a test for that, I discovered I returned from ClusterConfig while holding a lock... Thus I refactored that anyway way too long function such that locking happens on an outer scope, thus preventing such a mistake (or at least making it super obvious, as lock and unlock are right next to each other now).

lib/model/model.go
```diff
@@ -334,8 +342,17 @@ func (m *model) addAndStartFolderLockedWithIgnores(cfg config.FolderConfiguratio
 
 	ffs := fset.MtimeFS()
 
+	if cfg.Type == config.FolderTypeEncrypted {
+		if encToken, err := fs.ReadFile(ffs, encTokenPath(cfg)); err == nil {
```
calmh, line 346: Can we make this a structured file (json?) and include the folder ID in it?

```
{  "folderID": "foo", "token": "..." }
```

Currently it's a weakness that if everything comes crashing down and I need to offline-decrypt the files I need to know both the password and the folder ID, but the folder ID isn't stored anywhere and is probably randomly generated.

imsodin:
The token is now json encoded with the id.

I now also completed the previously half-done (if generous) mechanism to detect and undo changes on an encrypted folder: The scanner got a new flag `DetectChangesOnly`: If true, it doesn't hash files just returns the fileinfo without blocks when detecting a change. Thus we can mark those changes as invalid and get back valid, encrypted data. In hindsight maybe `NoBlocks` would be a better name for this option (?).

This also means that the encrypted dir extension isn't strictly necessary anymore (it was in place to prevent any scanning when there wasn't a folder type). However I believe it's still a nice touch to make it super obvious on all top-level elements that this is an encrypted folder.

AudriusButkevicius:
Just wanted to say, I don't plan to review the individual PRs to the feature branch, as it seems there is a lot of happening, but I'd still like to go through the final monster PR once you guys think it's done done.

If you think something needs discussion and requires my attention. just summon me with an `@`

lib/protocol/bep.proto
```diff
@@ -74,6 +74,7 @@ message Device {
     bool            introducer                 = 7;
     uint64          index_id                   = 8 [(gogoproto.customname) = "IndexID", (gogoproto.customtype) = "IndexID", (gogoproto.nullable) = false];
     bool            skip_introduction_removals = 9;
+    bytes           enc_pw_token               = 10;
```
calmh, line 77: I think we should go with verbose names in the protocol spec, so `encryption_password_token`?

PR #6678, opened 2020-05-25
lib/versioner: Create versioning directory recursively (fixes #6565)

### LINKED ISSUE #6565: Create versioning directory recursively (creating parents as necessary)
This is the scene: 

I set a sync folder which enabled File Versioning and set the **versions path** to `/sdcard/Syncthing control/Desktop/Markdown` , but I didn't create this path manually, and when syncing, there would be conflict, when I check the logs, it says: 

```
Puller (folder "/sdcard/Documents/Markdown" (lziuh-l9bdv), item "test.pdf"): delete file: mkdir /sdcard/Syncthing control/Desktop/Markdown: no such file or directory
```

It seems the Syncthing tries to create the versioning folder. Because it's parent folder is not created manually, and it just used "mkdir" but not "mkdir -p", it can't create the parent folder of the versioning. 

So I suggest the solution is to use `mkdir -p` to replace the `mkdir` in unix like system when creating versioning folder.

Shidory:
If I can get some feedback on the test it will help me a lot as I am still learning how to do it with Go.


lib/versioner/staggered_test.go
```diff
@@ -75,3 +76,23 @@ func parseTime(in string) time.Time {
...
+
+	tmpDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	folderFs := fs.NewFilesystem(fs.FilesystemTypeBasic, tmpDir)
+
+	versioner := newStaggered(folderFs, map[string]string{
+		"fsType": "",
```
imsodin, line 91: You don't need to add the empty string to the map. If `"fsType"` is missing, that's the same as an empty string.

PR #6684, opened 2020-05-26
lib/config, lib/model: Commit auto-accepted folders all at once

imsodin:
Currently for every auto-accepted folder in a cluster config, we change that folder in the config and wait for that change to happen. Now instead of committing to config, `handleAutoAccepts` returns the changed config object and all of the changes are committed at once in `ClusterConfig`.

### DIFF
--- a/lib/api/mocked_config_test.go
+++ b/lib/api/mocked_config_test.go
@@ -102,6 +102,10 @@ func (c *mockedConfig) SetFolder(fld config.FolderConfiguration) (config.Waiter,
 	return noopWaiter{}, nil
 }
 
+func (c *mockedConfig) SetFolders(folders []config.FolderConfiguration) (config.Waiter, error) {
+	return noopWaiter{}, nil
+}
+
 func (c *mockedConfig) Device(id protocol.DeviceID) (config.DeviceConfiguration, bool) {
 	return config.DeviceConfiguration{}, false
 }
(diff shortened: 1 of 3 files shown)


imsodin:
@AudriusButkevicius Your auto accept racyness test is pretty good: The previous change apparently made config racyness much worse for cluster configs. There was a window between copying the entire config and modifying it, in which another cluster config could copy the same config. If both introduce a new folder, only one will "survive". Thus I added `Wrapper.SetFolders` which replaces/adds multiple folders at once without any need to copy the config.  
While doing that, I realized the `wrapper.folderMap` caching is boguous: We return the same object without copying on every `.Folders` call. The initial copy only protects the underlying config from changes, however the first caller of `.Folders` can still change the underlying map, influencing what is returned on a second call. Thus I removed `.folderMap` and copy on every call to `.Folders`.

AudriusButkevicius:
I think there is still a race of two clusterconfig messages touching the same folder though...

Effectively, ages ago I was talking about:

cfg.Mutate(func(*cfg) {}) being the only sensible way that permits serializing access to the config. If we don't serialize it, there will always be races due to read modify write, as none of those are serial.

imsodin:
Oh I am not saying there's no racyness anymoe now, I am saying this is now about as racy as before, however doing all the auto-accepts at once.

PR #6689, opened 2020-05-28
build: Clean up build.sh, add build.ps1

calmh:
### Purpose

Remove old cruft and add feature parity for Windows command line peeps.


timfallmk:
Some of the removed code was being [used by the homebrew formula](https://github.com/Homebrew/homebrew-core/pull/57963)

PR #6691, opened 2020-05-29
gui: Add controls for encryption

imsodin:
### Purpose

Do *not* look at the code please, it's neither functional nor pretty, just some html to get the "mockup screenshots" below. I just need some feedback on the UI choices :)

The idea is to control encryption from the share tabs in the folder and device edit modal. In the first two screenshot sharing and encryption is enabled in a single dropdown menu (pw still needs to be set). The second two screenshots are an option with just a checkbox for sharing and encryption is enabled if a password is set.

### sharing

Wide, normal folder:
image

Narrow, normal folder:
image

encrypted folder (i.e. can't encrypt devices):
image

---

### folder type

Can't be changed from/to encrypted after initial setup.

normal folder, can't change to encrypted:
image

encrypted folder, can't change type:
image


AudriusButkevicius:
Checkboxes are simpler.

calmh:
Yes, I agree. My preference is towards the bottom screenshot. The checkbox control sharing, and entering a password or not controls encryption.

I think "encryption disabled" is a confusing placeholder for the "not shared at all" case, I would think "not shared" or just the same (disabled) "enter password to encrypt" would be more clear.

I think it would be nice with an "eye" toggle on the right of the password field to make it a regular visible text box. Both for verification since there isn't a double entry thing and for the future when I've forgotten the password and want somewhere to see it and/or copy-paste from.

uok:
Thanks @imsodin for working on that!
I suggest a [switch](https://getbootstrap.com/docs/4.5/components/forms/#switches) (or something similar) to show/hide the password field. This way the user has a clear on/off interface. As placeholder I think "encryption password" is enough.

imsodin:
I updated/added screenshots in the top post. Hardly wins a design award but I hope ticks all the boxes that have been brought up so far. Please give another round of feedback on the overall setup/design, afterwards I'll do the same for the device modal and polish for actual code review.

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -189,13 +177,16 @@
...
                     <option value="receiveonly" translate>Receive Only</option>
+                    <option value="encrypted" ng-disabled="editingExisting" translate>Encrypted</option>
                   </select>
                   <p ng-if="currentFolder.type == 'sendonly'" translate class="help-block">Files are protected from changes made on other devices, but changes made on this device will be sent to the rest of the cluster.</p>
                   <p ng-if="currentFolder.type == 'receiveonly'" translate class="help-block">Files are synchronized from the cluster, but any changes made locally will not be sent to other devices.</p>
+                  <p ng-if="currentFolder.type == 'encrypted'" translate class="help-block">Receives only encrypted data, i.e. cannot read stored data. All remote devices need to be set up with the same password or be encrypted themselves to share this folder.</p>
+                  <p ng-if="editingExisting" translate class="help-block">Type cannot be changed from/to encrypted on an existing folder. You need to remove it, potentially clean up/decrypt data on disk, and re-add it.</p>
```
uok, line 189: Suggestion for easier reading/translation:
Folder type "encrypted" cannot be changed after saving. You need to remove the folder, optionally delete files on disk, and add it again.

uok:
Thanks! I suggest to hide the password field if the folder is not share with a device

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -189,13 +177,16 @@
...
                     <option value="sendreceive" translate>Send &amp; Receive</option>
                     <option value="sendonly" translate>Send Only</option>
                     <option value="receiveonly" translate>Receive Only</option>
+                    <option value="encrypted" ng-disabled="editingExisting" translate>Encrypted</option>
                   </select>
                   <p ng-if="currentFolder.type == 'sendonly'" translate class="help-block">Files are protected from changes made on other devices, but changes made on this device will be sent to the rest of the cluster.</p>
                   <p ng-if="currentFolder.type == 'receiveonly'" translate class="help-block">Files are synchronized from the cluster, but any changes made locally will not be sent to other devices.</p>
+                  <p ng-if="currentFolder.type == 'encrypted'" translate class="help-block">Receives only encrypted data, i.e. cannot read stored data. All remote devices need to be set up with the same password or be encrypted themselves to share this folder.</p>
```
uok, line 188: Suggestion for easier reading/translation:
Receives only encrypted data. All connected devices need to be set up with the same password.

imsodin:
The folder type is now called "Vault", because "encrypted" is simply overused and thus has a high potential for confusion (folder content can be plain text, but encrypted when sending to a untrusted device, and transport is anyway always encrypted, ...).

@uok I adjusted the strings, but slightly deviated from your proposals. Please have a look whether it makes sense to you like this.  
What is your motivation for hiding the password fields? I am not a fan of hiding elements. I find dis-/appearing elements distracting and we can't use the space for something else (I certainly don't want a mix of single and double column).

tomasz1986:
> The folder type is now called "Vault", because "encrypted" is simply overused and thus has a high potential for confusion (folder content can be plain text, but encrypted when sending to a untrusted device, and transport is anyway always encrypted, ...).

Just an opinion, but "vault" sounds very cryptic [sic], i.e. impossible to understand by normal computer users (without reading the documentation), especially when placed next to the very plain and simple "Send" and "Receive". "Encryption" and "encrypted", on the other hand, are widely used and understood, also outside of the computing world, if only because of the massive file encrypting ransomware outbreaks.

uok:
@imsodin 
strings
are more clear now - thanks!

hiding password fields
it prevents a simple UX error where user enters password but forgets to check the checkbox to share with device. Also the password field should not be an text indicator with "not shared" if a device is not shared - that is imho what the checkbox is for. 

vault
interesting suggestion, but I'm undecided. I agree with @tomasz1986 that it may be confusing for the average user. "Encrypted" is even understood by my parents :smile:

imsodin:
@tomasz1986 What you describe as the downside is actually what I considered the upside: Encrypted is used in so many contexts, it leaves the door wide open for anyone to misunderstand it. Lets say the user wants to create a folder to send data to a friend as a p2p backup scheme and wants to send data encrypted. Naturally the user uses the folder type "encrypted", because hey, that's what they want: Encrypt data - wrong. Or seeing three folder types and then another folder type called "encrypted", I can already see more support request along the lines of "Syncthing connections are not encrypted by default, that's terrible". Essentially so many things are encrypted, that it's no longer a unique enough term. Encrypted folder is not wrong, and I agree it's more meaningful, however I believe it's more likely to guide less knowledgeable people towards the right conclusion (with help texts and docs (latter admittedly being wishful thinking)) while people knowing what they are looking for might be thrown off for a second by vault, but I am quite sure will still recognise it as what they are looking for.  
Obviously there might by better terms than "vault", that's just the best @calmh and me agreed upon in a very short time :) 

>  it prevents a simple UX error where user enters password but forgets to check the checkbox to share with device. Also the password field should not be an text indicator with "not shared" if a device is not shared - that is imho what the checkbox is for.

One can't enter a pw on a unchecked folder - the password field is disabled then. The help text is meant to explain why it's disabled.

AudriusButkevicius:
I do agree, vault is a terrible idea, because first of all it's a software product.

tomasz1986:
I actually have an opposite feeling that some people will select "Vault" simply because it sounds cool, with no idea what the concept is about. Let us hope that I am wrong.

I am not saying that the name has to be "Encrypted" though, only that the term "Vault" is completely foreign to the majority of computer users. I personally believe that obscurity can easily become a double-edged sword here.

imsodin:
> I do agree, vault is a terrible idea, because first of all it's a software product.

As I said before, I am very much open to other names or arguments against the problems on "encrypted" I brought up. As to "first of all it's a software product": I think first of all it's an underground dungeon full of gold. Then again, there doesn't seem to be any "first of all" in general and not even for software: https://en.wikipedia.org/wiki/Vault Aka I really don't think various software using the word "vault" in their name or otherwise is a problem for us using it for an option. It might still be a terrible idea for another reason.

Maybe "locally-encrypted"? It's a mouthful, but with the addition of "locally" I hope to convey that it's not about connections, i.e not about transport encryption or outgoing data encryption (as opposed to the device encryption setting).

AudriusButkevicius:
I think encrypted is fine, add a banner message next to it explaining what it is and saying that the communication is encrypted anyway.

AudriusButkevicius:
Password protected?

imsodin:
I like that, as in protected from a local malicious actor. Are you fine with shortening that to just "protected"? I don't think it looses essential meaning and shorter is nicer.

AudriusButkevicius:
I think it's not obvious from the shorter name that there must be passwords involved somewhere.

uok:
compromise: "password protected" as label/folder type and "protected" for the internals

gui/default/assets/lang/lang-en.json
```diff
@@ -2,7 +2,7 @@
    "A device with that ID is already added.": "A device with that ID is already added.",
    "A negative number of days doesn't make sense.": "A negative number of days doesn't make sense.",
    "A new major version may not be compatible with previous versions.": "A new major version may not be compatible with previous versions.",
-   "A vault stores and syncs encrypted data. Folders on all connected devices need to be set up with the same password or be of type vault too.": "A vault stores and syncs encrypted data. Folders on all connected devices need to be set up with the same password or be of type vault too.",
+   "A receiveEncryptedOnly stores and syncs encrypted data. Folders on all connected devices need to be set up with the same password or be of type receiveEncryptedOnly too.": "A receiveEncryptedOnly stores and syncs encrypted data. Folders on all connected devices need to be set up with the same password or be of type receiveEncryptedOnly too.",
```
AudriusButkevicius, line 5: Seems like a bunch of bad fine replaces here.

imsodin:
The folder type is now "receive encrypted". Going back to using the generally loved "encrypted" combined with "receiving", hopefully making it clear that "encrypted" relates to received data. I hope that can get consensus :)

The code is now also in a functional (manually tested) and hopefully presentable state, i.e. ready for review - however only look at the diff in `./gui`: The PR is against the encryption branch, but contains the changes in `encryption-token` (as it requires the config changes from that).

gui/default/assets/lang/lang-en.json
```diff
@@ -134,6 +134,7 @@
    "Folder Label": "Folder Label",
    "Folder Path": "Folder Path",
    "Folder Type": "Folder Type",
+   "Folder type \"{%receiveEncrypted%}\" cannot be changed after adding the folder. You need to remove the folder, delete or decrypt the data on disk, and add the folder again.": "Folder type \"{{receiveEncrypted}}\" cannot be changed after adding the folder. You need to remove the folder, delete or decrypt the data on disk, and add the folder again.",
```
calmh, line 137: Could we not just say "Receive Encrypted" here without the magic token? It will be translated as part of the string as a whole?

gui/default/syncthing/core/editShareTemplate.html
```diff
@@ -0,0 +1,33 @@
...
+      <span ng-switch-when="device">{{folderLabel(id)}}</span>
+      <span ng-switch-when="folder">{{deviceName(findDevice(id))}}</span>
+    </span>
+  </label>
+</div>
+<div class="col-md-6">
+  <div class="input-group">
+    <span class="input-group-addon" ng-switch="!currentSharing.encPWs[id] || currentSharing.encPWs[id] === ''">
+      <span ng-switch-when='true' class="fas fa-fw fa-unlock" />
+      <span ng-switch-when='false' class="fas fa-fw fa-lock" />
```
calmh, line 14: Should we show a lock when the type is receive encrypted, even though there is no password entered and one cannot be entered?]

gui/default/assets/lang/lang-en.json
imsodin: I thought it was a nice touch to ensure consistent translations being used, given this is a "label/name", not just a normal language phrase. I.e. preventing the possibility that the string "Receive Encrypted" alone gets translated to something different than in the context of this sentence.

gui/default/syncthing/core/editShareTemplate.html
imsodin: Probably makes sense, consistent with a lock meaning "data exchanged with this device is encrypted".

gui/default/assets/lang/lang-en.json
calmh: Okay, let's leave it as-is.

AudriusButkevicius:
I'd hide the password box if the checkbox is unchecked.

AudriusButkevicius:
Also, I probably lost the plot here, but what does receive encrypted mean?
I was under the assumption that receive is receive, whether it's en/decrypted or not depends on whether the password is specified?
Why do we suddenly need this distinction?

calmh:
It eases consistency of configuration and passwords, and enforces lack of local changes.

imsodin:
Password fields are visible, but deactivated when not shared - with a help text saying it's not shared. Seems like people like hiding better? As stated before I personally like static, disabled elements better than elements that appear and disappear at will. However I don't care much either way and it's simple to change.

The password field is when you locally have the un-encrypted data, and encrypt that for the device you share the data with. The "receive encrypted" folder type is if you are on the other end, i.e. you do get encrypted data (in which case you obviously don't need to set any passwords).

AudriusButkevicius:
Sorry, I know the concepts of the application quite well, and I thought I followed the encryption plot here, but I am already sitting here like "what the fuck does this all mean".

Can you not derive the magical folder type internally if you need it?

I suspect this password per device will already confuse the fuck out of people. I understand it's more flexible, but it's more confusing and more error prone. I can see the influx of people with cries about mismatching passwords, data being unreadable, incorrect folder types being selected leading to fuck knows what.

I guess it would be easier if all folders were encrypted by default.

A simple checkbox "this is an encrypted folder" with an option to specify a en/de-cryption password would be much more simple, which when checked and with no password would show message saying this folder can only receive encrypted data bla bla.

You could put the value of that checkbox in CC message and blow up if stuff does not align.

AudriusButkevicius:
I guess there are 2 rants here, first is that we're going for "complicated by default" because of multiple passwords, second is that a folder types feel more heavy weight in their behaviour than a checkbox. 

First question that comes to my head is why isn't there a sendreceiveEncrypted sendonlyEnrcypted if receiveOnly needs a specialization?

imsodin:
> Can you not derive the magical folder type internally if you need it?

Sure, we can detect that we are encrypted by receiving the encryption token (or detecting encrypted indexes or whatever). What happens if the user creates a folder, with data in it, and then a device connects sending encrypted data? Hardly nuke all the local data, so warn and drop said connection. What do you write in the warning? You need to explain why it's a problem that you have data locally, and the remote wants to share encrypted data etc etc - I wouldn't know how to do that in warning, it's hard enough to do in docs (illustrations help though). It's much easier to tell the user: "Remote XY wants to send encrypted data, but your folder is not configured to receive encrypted data - go talk to XY or configure your folder right".

As the traditional folder types make no sense when you have encrypted data locally (you can't send anything, so you are essentially receive-only, but not quite, because you don't know anything about the real state of the folder) I considered it a good idea to extend the folder type. It might indeed be more natural to have a new setting, something like "locally-encrypted". I see how that makes just as much sense as a folder type. However I really do think it's necessary to make it an explicit characteristic of the folder in order to be able to give clear errors when misconfigurations happen. Which will happen - if you have a misconfiguration-proof solution, I am all ears. 

> Perhaps the per-device is password is a nice feature, but does it really need to be the default we shove down everyones throat?

You are charging an open door here - t's not like that wasn't considered: https://forum.syncthing.net/t/latest-encrypted-devices-proposal-what-how/14164/57  
Just that at the time my conclusion was the general sentiment is "someone will have a use-case for it, lets just support it". It would be entirely possible to show only a single PW in the UI and let any advanced user wanting separate PWs use edit the config directly (or have a switch in the UI to select single or multiple PWs).

> You could put the value of that checkbox in CC message and blow up if stuff does not align.

That's exactly what happens.

In the end it boils down to:
 - Present a single PW (or just have a single PW) or one per device  
 - How to present the fact that the local data is encrypted and any failure/misconfig states related to that to the user.

AudriusButkevicius:
I am not saying that we should not support it, but I don't think that many people will want this by default.

I think the average Joe who can't work out that folder IDs need to match on both sides, and fully rely on the "spice girls - wannabe" prompts will shit themselves when they see the "password bingo" we got going on for them.

But, now we have custom folder types adding a second final boss, to this "share a folder encrypted" game, which I feel effectively takes you to "I can't do anything until I've got my Phd in syncthing" territory. 

I guess the ways I see to simplify this is:
1. Have a checkbox for "content of the folder is encrypted with a password"
2. Have one password field. The password can only be left out empty if the folder is receive only. That password sets a per-folder password which is applied to every device (unless the device has it's own password in the config or something, which advanced people can craft themselves).
3. Maybe, but probably not, some sub-dialog if people want to go whole hog with per device passwords, but probably hand crafting XML is also good enough for those people.

Or if you guys feel this will be such a mainstream hit, then feel free to ignore me, and this looks good.

calmh:
So, yeah, we discussed this on the forum but we can discuss it again here. Though I'd prefer it happened on the forum. 🤷 

Some constraints we have on the implementation:

- Protocol and config wise there is a password per {folder, device} tuple. We can hide this but in the end the user needs to enter it somewhere. If it's just per folder then we either wouldn't have a field for the password in the device dialog (and setting up an encrypted folder becomes an inscrutable and annoying multi-dialog process) or we'd have the field and it *looks* per-device but isn't as it also changes the password for other devices... And also clobbers your suggested handcrafted XML if ever touched. Neither option seems great to me.

- On the side that stores the encrypted data we need some special support. There's a magic token to store, and there's special handling of scanning and such for the files on disk. Picking up non-encrypted data and sending it out is not acceptable. This is best represented by a folder type, internally. If we want to represent it in the UI as "no folder type, just a checkbox" (which disables the normal folder type selector) that's doable but seems odd to me. We can auto suggest the right folder type in the new folder sharing popup, but we can't usefully switch folder types once it's populated with any kind of data...

The suggestions above seem half baked to me since they result in new and other problems. If you can propose an actual workflow (as in "this is what the user will do when they set it up) that is better then I'm sure we're all ears. Currently it's "just" a matter of entering a password in the appropriate place in the device or folder editor, and setting up a folder on the other side to receive the data...

AudriusButkevicius:
I know I was part of the initial discussions, and everything seemed fine while I had the plot under control, but now having lost the plot, I stepped back, had a look at this again, hence why I brought this up. 

What I am proposing is:

1. Have a field in the folder dialog for the password for the folder.
2. Have a checkbox makes it clear that this folder will have it's stuff encrypted, that can drive the folder types internally, but perhaps it should be a flag on the folder in that case.
3. We don't need move/hide dialogs elements just present validation errors (empty password for send-receive is a nono, validation error on password)
4. Have a folder level password which is used as a fallback for every FolderDevice that does not have their own password.
5. Sharing a folder from the device dialog does what it did before, shares the folder with an empty password. The "fallback" mechanism saves the day and the device still ends up with the folder level password used. when the connection is set up.
6. For advanced people, they can go and setup a password for FolderDevice manually, there is nothing in the UI to control that, and nothing in the UI signifies that it's set.

imsodin:
In your (Audrius) scenario having both trusted (unencrypted) and untrusted devices for one folder would be an advanced usage, not configurable through the folder edit dialog, correct?  
I am thinking of a use case like several trusted devices locally (B-trusted), a central, hosted, untrusted device (A-trusted) and a/several trusted devices at a second location. I consider that a realistic, not very advanced use-case for encryption. Of course encrypting for trusted devices works, it's just useless overhead on both sides. However maybe that's an acceptable tradeoff between usability and performance.

If we don't want to make that performance tradeoff of encrypting for everyone, I see two steps for improvement based on both your input:

1. A single encryption control in the "Share With" tab:  
At the top of the share dialog there's a checkbox to make a folder locally encrypted as in the current "Receive Encrypted" folder type. This disables (removes) the PW fields in the dialog below. The folder type selector is disabled and shows "Receive encrypted" if the enc checkbox is ticked. Otherwise the usual three folder types are selectable.

2. Do not support different PWs for a single folder in the web UI. Builds on top of 1. and everything is still configurable in a single dialog and you can have both trusted/unencrypted and untrusted device on a single folder:

Folder:  
The checkbox at the top as in 1. now also has a pw field. If there's multiple different pws in config, the input is deactivated with a message saying why. Every device then gets two checkboxes: "Shared" and "Untrusted".

Device:  
No top control. Folders are separated in unencrypted and locally encrypted/"Receive Encrypted".
For unencrypted there are three controls per folder that's not "Receive Encrypted": The same "Shared" and "Untrusted" checkboxes plus a pw input. If the folder already has a PW, that's displayed and changing it is disabled. Thus you can add a new, untrusted device and immediately set the folder PW in a single dialog.  
For "Receive Encrypted", only the "Shared" checkbox is available.  
Alternatively no separating, and the "Receive Encrypted" devices just have deactivated/hidden "Shared" checkbox and pw input.

I am not entirely convinced by 2. I do not see the value of differing passwords per device, so hiding that is nice, but then again it has even more control elements than now (even though it's simple boolean elements versus string, which is clearly simpler). So yeah, I guess I need (and will get) some third opinions now :)

AudriusButkevicius:
So fuck it, I give up.

It's fine as it is.

imsodin:
What about the folder type discussion you brought up, which would be change 1. above. You kinda convinced me on that. Fuck that too? I mean I want this encryption stuff to be quite good, or at least ok, and your feedback is clearly helpful towards reaching that - and I don't quite know for sure where you stand from "I give up" ;)

AudriusButkevicius:
I guess I'm tired talking about this, hence effectively I don't care, hence I give up and I want to stop sinking time into this.

I still think we could get away with a single password. Those RPi zero folks can assume themselves to be advanced users and go into hand crafting xml to disable encryption for certain devices (whatever disabling means... magical sentinel value that means empty string?)

As for your proposal, it's not exactly clear to me what you are proposing, seems like a new control that doesn't do anything other than control some old control... That's probably even more confusing, as people will check that checkbox and then not know where to put the passwords going "but but but I want to send stuff encrypted, I selected encrypted and it says receive only".

calmh:
I think we just need to do *something*, see how it works. If the feedback is that it's too difficult then we can refactor. Too much theorizing at this point isn't super useful.

Issue #6674, opened 2020-05-23
Consider removing rolling hash support


calmh:
I realize I'm suggesting killing someones baby here. However, weak hashing has some non-zero costs: in calculating when hashing, in checking when syncing, and in code complexity. (Although the latter is not great, as it's proven stable enough to not require a lot of maintenance.)

The utility to compare to the cost is an apparent saving of 1% on data transfer, according to data.s.n:

<img width="966" alt="Screen Shot 2020-05-23 at 10 55 44" src="https://user-images.githubusercontent.com/125426/82726381-021aaa80-9ce4-11ea-8065-1b037a197146.png">

In my mind this effectively rounds to zero, so the question is if it's worth keeping around.

I'm aware that the utility is probably >> 1% to *some* users, possibly to the point of being mission-critical for an extremely small subset of users. The hashing cost is however carried by all.

(There are also many other features with < 1% usage where the same argument could be applied, probably. I'm focusing on this one just because it has a general runtime cost.)

AudriusButkevicius:
Did you evaluate how much of a cost is the non-zero cost? 

If its 1% cost for 1% gain, sounds pretty good to me.
Its also used for the request response validation, which is cheaper than sha256, so there is some utility there.

Also, we could put it back behind a flag, and let people switch it off if you are seeing cases where its detrimental.

calmh:
Good point.

greatroar:
I've been looking at the weakhash finder for some time to see how it could be improved, and I think I found two flaws in it. The algorithm is currently:

1. Finder scans through the file, recording matching offsets for a set of hash values. It always only records the first ten it finds.
2. Client code re-reads the matching blocks and verifies them using SHA-256.

The first flaw is the inefficiency from reading the file multiple times. The second, which I suspect may harm its effectiveness, is that it the finder is biased to the front of the file. If edits occur near the end, then the probability that the first ten matches are all false positives increases with the file size.

I believe both issues can be solved at once by redesigning the finder to be one-pass and stateless aside from the file offset. WDYT?

imsodin:
Single pass would certainly be nice and push the scales even a little more towards a better cost-benefit ratio :)

[...]

PR #6692, opened 2020-05-30
lib/model, lib/weakhash: Single-pass weakhash.Finder

greatroar:
Follow-up to #6674. This PR completely rewrites the weakhash.Finder, so it's a rather big patch. I do believe that the new code is more straightforward.

The Finder now does a single pass over the file, which is done prior to any alternative methods of finding blocks. Its new API is inspired by bufio.Scanner; it is entirely decoupled from SHA-256. There is no longer an arbitrary limit of ten hits per hash value.

The Finder still has to read back into its Reader to fetch the blocks, but it always reads the exact part of the file that it had previously read, so the block in question is almost certainly still in the filesystem cache. If the rollinghash library were updated or replaced so that the blocks can be copied out of its buffers, even this read could be eliminated.


lib/model/folder_sendrecv.go
```diff
@@ -1334,16 +1294,85 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
...
+	if len(byWeakhash) == 0 {
+		l.Debugf("not weak hashing %s. file did not contain any weak hashes", state.file.Name)
+		return state, nil
+	}
+
+	done := make([]bool, len(state.blocks))
+
+	for finder.Next() {
+		h, offset := finder.Match()
+		sha := sha256.Sum256(buf)
```
AudriusButkevicius, line 1333: Match() needs to return buf, because now buf is size of BlockSize(), yet last block can be smaller than that.

lib/model/folder_sendrecv.go
```diff
@@ -1334,16 +1294,85 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
...
+	// Remove the blocks we've done from blocks.
+	blocksleft := state.blocks[:0]
+	for _, indices := range byWeakhash {
+		for _, i := range indices {
+			if !done[i] {
+				blocksleft = append(blocksleft, state.blocks[i])
+			}
+		}
+	}
+	state.blocks = blocksleft
```
AudriusButkevicius, line 1371: I think this screws up the remaining block order?

I think you could just iterate over done?

lib/weakhash/weakhash.go
```diff
@@ -8,111 +8,125 @@ package weakhash
...
-	offsets := make(map[uint32][]int64)
-	for _, hashToFind := range hashesToFind {
-		offsets[hashToFind] = make([]int64, 0, maxWeakhashFinderHits)
+// NewFinder returns a Finder that initially has no hashes.
+// Call Add before Next to add hashes.
+//
+// The buffer buf will be filled for each match found. Its length is taken
+// to be the block size.
+func NewFinder(r io.ReaderAt, buf []byte) *Finder {
+	rr, ok := r.(io.Reader)
```
AudriusButkevicius, line 43: We should just change the interface to be what it needs to be

lib/weakhash/weakhash.go
```diff
@@ -8,111 +8,125 @@ package weakhash
...
+// When Next has returned true, the contents of the block found are in the
+// buffer handed to NewFinder.
+func (f *Finder) Next() bool {
+	if len(f.hashes) == 0 {
+		return false
 	}
 
-	for _, offset := range h.offsets[hash] {
-		_, err := h.reader.Seek(offset, io.SeekStart)
+	blocksize := int64(len(f.buf))
```
AudriusButkevicius, line 88: blockSize

lib/model/folder_sendrecv.go
AudriusButkevicius: We could return the size of buf, and then before hashing, make sure that the candidate blocks are of the expected size before wasting time doing sha256, but perhaps that's micro-optimization.

lib/weakhash/weakhash.go
```diff
@@ -8,111 +8,125 @@ package weakhash
...
+
+	if f.offset < blocksize {
+		// Initialize by reading the first blocksize bytes. ReadAt must
+		// return an error when it reads less than len(f.buf) bytes.
+		_, err := io.ReadFull(f.br, f.buf)
```
AudriusButkevicius, line 93: This will fail on last block presumably, that is not BlockSize sized. I guess it's ok.

lib/weakhash/weakhash.go
```diff
@@ -8,111 +8,125 @@ package weakhash
...
+		f.hf.Roll(bt)
+		f.offset++
+
+		h := f.hf.Sum32()
+		if _, ok := f.hashes[h]; ok {
+			// We have to read the block again here, because the rollinghash
+			// library does not provide access to its buffers. This is wasteful
+			// because the block is already in memory somewhere, but at least
+			// it will likely still be in the disk cache.
+			_, f.err = f.r.ReadAt(f.buf, f.offset-blocksize)
```
AudriusButkevicius, line 124: Feels like SeekReader would be a better option here? Atleast you'd avoid having to have two readers.

lib/weakhash/weakhash.go
```diff
@@ -8,111 +8,125 @@ package weakhash
...
-// the hashes provided, and returns a hash -> slice of offsets within reader
-// map, that produces the same weak hash.
-func Find(ctx context.Context, ir io.Reader, hashesToFind []uint32, size int) (map[uint32][]int64, error) {
-	if ir == nil || len(hashesToFind) == 0 {
-		return nil, nil
-	}
+	r    io.ReaderAt
+	sr   *io.SectionReader
+	br   *bufio.Reader
+	size int64 // File size.
```
AudriusButkevicius, line 29: never used

lib/weakhash/weakhash.go
```diff
@@ -8,111 +8,125 @@ package weakhash
...
 
-// Find finds all the blocks of the given size within io.Reader that matches
-// the hashes provided, and returns a hash -> slice of offsets within reader
-// map, that produces the same weak hash.
-func Find(ctx context.Context, ir io.Reader, hashesToFind []uint32, size int) (map[uint32][]int64, error) {
-	if ir == nil || len(hashesToFind) == 0 {
-		return nil, nil
-	}
+	r    io.ReaderAt
+	sr   *io.SectionReader
```
AudriusButkevicius, line 27: never used

AudriusButkevicius:
I thought that rolling hash library only keeps 64 or so bytes, which puzzled me as to how it works with that.

lib/weakhash/weakhash.go
greatroar: ReadAt is a single system call (pread) on Unix.

greatroar:
My experiments to determine the false positive probability of Adler-32 have so far yielded the following.

I chunk a file into 128kB-sized blocks and compute the Adler-32 hash of each block. I then search the file for these hashes and count the number of hits for each hash, skipping the blocks at offsets that are multiples of 128kB. I ran this on compressed files, which are very unlikely to contain actual dupicates

For files < 1GB, I can't find an instance where there are more than four hits for any hash value. However, for a 58GB (bzip2-compressed) file, I find that 446181 out of the 470559 hashes occur more than ten times. So somewhere in the multi-GB range, the current weakhash algorithm breaks down. It could, of course, be "fixed" by increasing maxWeakhashFinderHits, but unless I'm mistaken, the number of false positives increases quadratically with the file size...

lib/weakhash/weakhash.go
greatroar: Done.

lib/weakhash/weakhash.go
greatroar: Done.

lib/model/folder_sendrecv.go
```diff
@@ -1334,16 +1294,84 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
...
+	for finder.Next() {
+		h, offset := finder.Match()
+		sha := sha256.Sum256(buf)
+
+		for _, i := range byWeakhash[h] {
+			if !bytes.Equal(state.blocks[i].Hash, sha[:]) {
+				continue
+			}
+
+			_, err = f.limitedWriteAt(w, buf, offset)
```
AudriusButkevicius, line 1340: This should be state.blocks[i].Offset?

lib/model/folder_sendrecv.go
```diff
@@ -1334,16 +1294,84 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
...
+		select {
+		case <-f.ctx.Done():
+			return state, f.ctx.Err()
+		default:
+		}
+	}
+
+	// Remove the blocks we've done from blocks,
+	// preserving the order of the remaining ones.
+	blocksleft := state.blocks[:0]
```
AudriusButkevicius, line 1364: blocksLeft

lib/model/folder_sendrecv.go
greatroar: Good catch. I've added a check in sendReceiveFolder.copyByWeakHash so that the last block is never shown to the weakhash.Finder if it isn't of the right size.

lib/weakhash/weakhash.go
greatroar: This is only run on the first block, so I guess it always fails for files of < 128kB. I reckon that's fine.

lib/model/folder_sendrecv.go
greatroar: Yes, good point. Can the current weakhash test be extended in an easy way? I find the level of abstraction in that test quite daunting...

lib/weakhash/weakhash.go
```diff
@@ -8,111 +8,121 @@ package weakhash
...
-// return whether it wishes to continue iterating.
-func (h *Finder) Iterate(hash uint32, buf []byte, iterFunc func(int64) bool) (bool, error) {
-	if h == nil || hash == 0 || len(buf) != h.size {
-		return false, nil
+// Next returns true if it can find a match for any of f's hashes,
+// false if it encounters either an error or EOF.
+//
+// When Next has returned true, the contents of the block found are in the
+// buffer handed to NewFinder.
+func (f *Finder) Next() bool {
```
imsodin, line 79: I'd return an error only (with EOF meaning all good, but finished), or `(bool, error)`. Then you can drop the `Error` method and it's immediately evident on the loop whether an error occurred, without the api user needing to think of calling `Error` afterwards.

lib/model/folder_sendrecv.go
imsodin: Could you point out where/what exactly that check is, I fail to find it.

lib/model/folder_sendrecv.go
AudriusButkevicius: I guess you can write a new test that is less daunting that exercises this.

lib/model/folder_sendrecv.go
```diff
@@ -1334,16 +1294,93 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
...
+	defer file.Close()
+
+	buf := protocol.BufferPool.Get(state.file.BlockSize())
+	defer protocol.BufferPool.Put(buf)
+
+	finder := weakhash.NewFinder(file, buf)
+
+	byWeakhash := make(map[uint32][]int) // Block indices by weak hash.
+	for i, block := range state.blocks {
+		if int(block.Size) != len(buf) {
```
greatroar, line 1318: @imsodin Here's the check for the size of the last block.

lib/model/folder_sendrecv.go
```diff
@@ -1334,16 +1294,93 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
...
+			// Skip the last block if it isn't block.Size bytes.
```
imsodin, line 1319:
```suggestion
			// Skip the last block if it isn'tstate.file.BlockSize() bytes.
```

lib/model/folder_sendrecv_test.go
```diff
@@ -415,6 +416,87 @@ func TestWeakHash(t *testing.T) {
 	}
 }
 
+func TestWeakHashWrite(t *testing.T) {
```
AudriusButkevicius, line 419: Does this fail before the bug was fixed?

lib/model/folder_sendrecv_test.go
```diff
@@ -415,6 +416,87 @@ func TestWeakHash(t *testing.T) {
...
+		Size:   size,
+	}
+
+	// Change the order of blocks from the desired file.
+	var buf [protocol.MinBlockSize]byte
+	for i := 0; i < nblocks; i++ {
+		shifted := (i + 1) % nblocks
+		_, err := desired.ReadAt(buf[:], int64(shifted*protocol.MinBlockSize))
+		must(t, err)
+		_, err = source.WriteAt(buf[:], int64(i*protocol.MinBlockSize))
```
AudriusButkevicius, line 460: Out of curiosity it would be interesting to write the blocks not exactly at the block boundary, to verify that rolling actually finds it.

AudriusButkevicius: Ah, nvm, it already does that.

lib/model/folder_sendrecv.go
AudriusButkevicius: I guess this is a conscious decision, but it feels it wouldn't add much complexity to support non BlockSize sized blocks?

AudriusButkevicius:
What happens if I have a file full of zeroes (or ones or whatever), and I try to  copyByWeakHash that?

I suspect we'll end up doing len(fileSize) copies, copying the same block over and over and over, and keep rolling everything even when everything in []done is true.

I think there are a few problems, we might re-do blocks that are already done, and we never terminate early in case everything is done.

lib/model/folder_sendrecv_test.go
greatroar: Yes.

lib/weakhash/weakhash.go
greatroar: The API design is a conscious decision: it makes the loop check a single expression. Note that the error from finder.Err is special, in that it is logged and then ignored. Regarding EOF, that isn't an error, it's an expected condition; there are no partial results to handle as in io.Reader.Read. bufio.Scanner discards it for the same reason.

lib/model/folder_sendrecv.go
greatroar: Done, as you already noticed.

lib/model/folder_sendrecv.go
greatroar: That would require maintaining two Adler-32 rolling hashes (so two sliding windows). It can be done, but the gain is always at most one fewer block to pull.

greatroar:
I don't know what's up with the failing tests. It looks like other tests are timing out. Did I not clean up correctly?

AudriusButkevicius:
It's just timeouts, probably rerun would fix it.

imsodin:
If it's all jobs failing it's likely not just a flaky test. I think you don't call `copyDone()` on the state now? Tests might be looking for that.

greatroar:
I've added early termination, please check.

greatroar:
Some of the builds succeed, others fail in what seem to be unrelated tests. Where should I add a copyDone?

lib/model/folder_sendrecv.go
```diff
@@ -1336,23 +1336,35 @@ func (f *sendReceiveFolder) copyByWeakHash(state copyBlocksState, w io.WriterAt)
...
+			done[blockIdx] = true
+
+			// Remove this block from the candidates for h, and remove
+			// h from the Finder if no candidates for it remain.
+			if len(candidates) == 1 {
+				delete(byWeakhash, h)
+				finder.Remove(h)
+			} else {
+				candidates[i] = candidates[len(candidates)-1]
+				byWeakhash[h] = candidates[:len(candidates)-1]
```
AudriusButkevicius, line 1366: If we hit this condition multiple times you'll end up with the same block in multiple places.

```
candidates = [1, 2, 3, 4, 5]

// Hit on 1
candidates[0] = 5 // candidates is now  [5, 2, 3, 4, 5]
byWeakHash[h] = [5, 2, 3, 4]

// No hit on 2

// Hit on 3
candidates[2] = 5 // candidates is now  [5, 2, 5, 4, 5]
byWeakHash[h] = [5, 2, 5, 4] // you get saved by the fact len(candidates) is always len(originalCandidates)-1 at worst, hence you never lose 4.
```

greatroar: There's a break statement right after this, so we jump into the next iteration of the outer loop. That resets candidates for the next h found.

Maybe this can be made more obvious by doing

```
candidates := byWeakhash[h]
found := -1
for i, blockIdx := range candidates {
	if theSha256Matches {
		found = i
		break
	}
}

if found == -1 {
	continue
}

// do the actual write and remove candidate here

AudriusButkevicius: Ah yes, missed that, this now makes sense but brings me to the next question of why we don't want to try the same h for multiple candidates tho?

imsodin:
> Some of the builds succeed, others fail in what seem to be unrelated tests. Where should I add a copyDone?

It's all the jobs which run tests that fail.  
Probably now you'd insert it here (after `.copiedFrom...`: https://github.com/syncthing/syncthing/pull/6692/files#diff-cf783533b0c2c4fe7c3a325925917a56R1357 Previously the weakhash finder was handled in the `copierRoutine` block loop, where there's a `if found { copyDone }` at the end.

lib/model/folder_sendrecv.go
greatroar: I've set the PR back to draft status until I figure out how to do this right. I'm not even sure if it's correct anymore.

greatroar:
I submitted chmduquesne/rollinghash#10 in order to get rid of re-reading and to make SHA-256 verification inside the Finder easier.

PR #6693, opened 2020-05-31
lib/model, lib/versioner: Drive version cleanup from scanner (fixes #6313)

### LINKED ISSUE #6313: Improving disk access control in folders with file versioning
In home NAS servers it makes sense to spin down disks when they are not used to reduce power consumption. By enlarging the rescan interval in Syncthing to a reasonably large value, like for instance 24h, the disks can be spun down for a substantial portion of the day.

However, when staggered file versioning is used, disks spin up hourly to remove file versions in the `.stversions` folder that have expired.

It has been [suggested in the forum](https://forum.syncthing.net/t/synchthing-prevents-disk-spindown/14446/8) that using a single hook to update the staggered file versioning and rescan the folder provides control over disk access by adjusting a single parameter.

calmh:
This change:

- removes the service bits of versioners, instead having them rely on
  `Clean()` being called periodically by someone else,

- adds a `Clean()` call to the versioner after periodic scanning,

- moves the versioner reference from the folder implementation
  (`sendReceiveFolder`) to the `folder`, because `folder` is what is
  responsible for driving scans.

Hence versioner cleanup is run on (after) the regularly scheduled scan.
It doesn't run on the various forced or triggered scans.

(We should add a documentation note about this as there will be effects
to completely disabling scans, but that's not something we recommend
anyway.)


AudriusButkevicius:
I'd do this as a separate controllable loop timer in the folder loop, but I guess if this takes a while for some reason, it's not ideal that it blocks the folder.

You can also probably do folderRunnerTokens to folderRunnerToken.

Also whats the rationale for the change?

calmh:
The rationale is in the ticket that is being solved by this :) (To not perform folder disk I/O every hour when the user has already said they want scans only once a day/week/etc.) It's also actually less code and moving parts.

I don't think we need to drive and configure this separately from normal scanning; the amount of work we need to do here is generally going to be a lot less than for the full folder scan so piggybacking on that seems to make sense to me.

AudriusButkevicius:
Sure, I can just see cases where people complain that scans take a long time have it disabled, and now they'll complain that versions never get cleaned up, or they don't get cleaned up often enough.

imsodin:
In general I consider linking the two things (scanning folder and versioner cleanup) is a good thing. However I do share the concern about existing users being taken by surprise by this. I'd expect the use-case where you do use no/very long scan intervals is likely to be a case where versioning is used as well (always on device that doesn't change files themselves). Again I do think it's reasonable to change the interval to something suitable for cleanup on such device, I just don't expect users to realize they should change it from a changelog. Maybe that's worth adding a notification in the UI if we detect versioning and rescan intervals >24h/disabled.

calmh:
So how about this. I add a separate versioner cleaning interval which defaults to one hour (no change for existing users) and drive that from the folder loop. Users wanting something different can change the interval in either direction.

AudriusButkevicius:
Sounds good. Potentially new folder states? Or I guess this should be short enough not to need it.

lib/config/folderconfiguration.go
```diff
@@ -42,7 +42,8 @@ type FolderConfiguration struct {
 	AutoNormalize           bool                        `xml:"autoNormalize,attr" json:"autoNormalize" default:"true"`
 	MinDiskFree             Size                        `xml:"minDiskFree" json:"minDiskFree" default:"1%"`
 	Versioning              VersioningConfiguration     `xml:"versioning" json:"versioning"`
-	Copiers                 int                         `xml:"copiers" json:"copiers"` // This defines how many files are handled concurrently.
+	VersionCleanupIntervalS int                         `xml:"versionCleanupIntervalS" json:"versionCleanupIntervalS" default:"3600"` // Follows same min/max as scan interval
```
AudriusButkevicius, line 45: Should this live in the versioner configuration?

lib/model/folder.go
```diff
@@ -701,6 +722,24 @@ func (f *folder) scanTimerFired() {
...
+	}
+	defer f.ioLimiter.give(1)
+
+	f.setState(FolderCleaning)
+
+	if err := f.versioner.Clean(f.ctx); err != nil {
+		l.Infoln("Failed to clean versions in %s: %v", f.Description(), err)
+	}
+
+	f.versionCleanupTimer.Reset(f.versionCleanupInterval)
```
AudriusButkevicius, line 740: Does a change in this trigger a folder restart?

calmh: Looks like any folder config attribute without `restart:"false"` triggers a folder restart, and the versioning config is without attribute. So yeah, I expect any change to the versioning config should trigger a restart appropriately.

calmh:
So, in starting to document this I realised this is almost entirely useless because there is no way to set the new option in the GUI. Fixing that in a separate PR. Maybe we should have more of a docs-first approach to PRs in some cases.

PR #6696, opened 2020-06-01
lib/scanner: Reduce allocations in Blocks

greatroar:
Here's a small optimization for scanner.Blocks. It performs fewer allocations and gets a tiny, but consistent, throughput boost:

```
HashFile-8    88.3ms ± 1%    87.8ms ± 0%   -0.58%  (p=0.000 n=19+19)

HashFile-8    45.4kB ± 0%    45.3kB ± 0%   -0.21%  (p=0.000 n=20+18)

HashFile-8      28.0 ± 0%      24.0 ± 0%  -14.29%  (p=0.000 n=20+18)

HashFile-8   202MB/s ± 1%   203MB/s ± 0%   +0.58%  (p=0.000 n=19+19)
```

### DIFF
--- a/lib/scanner/blocks.go
+++ b/lib/scanner/blocks.go
@@ -12,6 +12,7 @@ import (
 	"hash"
 	"hash/adler32"
 	"io"
+	"sync"
 
 	"github.com/syncthing/syncthing/lib/protocol"
 	"github.com/syncthing/syncthing/lib/sha256"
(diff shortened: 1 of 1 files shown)


lib/scanner/blocks.go
```diff
@@ -105,9 +96,59 @@ func Blocks(ctx context.Context, r io.Reader, blocksize int, sizehint int64, cou
...
 
+type multiHash struct {
+	strong hash.Hash
+	weak   hash.Hash32
+}
+
+var multiHashPool sync.Pool
+
+func newMultiHash(useWeakHashes bool) *multiHash {
+	h, ok := multiHashPool.Get().(*multiHash)
```
AudriusButkevicius, line 111: Not sure what the purpose of this is, but we never put anything, so presumably it's always empty.
Also, we don't run many parallel Blocks() calls I think, the prevent read thrashing, so I wonder if there is any point in sync.Pool

calmh:
More generally this is something performing file I/O and hashing, I doubt the four allocations removed have any impact on the overall performance. (Cleanup can still be good if the end result is clearer or less code...)

lib/scanner/blocks.go
greatroar: The Put is at the end of Blocks, just above the return statement.

greatroar:
I can't claim this is a cleanup since the current code is pretty straightforward. Closing; the trouble of changing the code is not worth it.

PR #6699, opened 2020-06-02
lib/connections: Skip and warn on malformed URLs (fixes #6697)

### LINKED ISSUE #6697: Syncthing hangs with slightly malformed listen address
### Overview

With syncthing 1.5 a slightly malformed listen address (tcp4:/127.0.0.1:22000) causes the process to hang.  

Negative

Please be sure to include at least:

 - **Version:** syncthing 1.5.0 on Linux Mint 19.3 (based on Ubuntu 18.04)  with kernel 5.3.0-53
 - **Steps to reproduce:** Replace the listen address with tcp4:/127.0.0.1 and restart
 - **Errant behaviour:** Syncthing hangs.  Ctrl-C is detected but not actioned.  Kill -9 is required to terminate the process.
 - **Expected behaviour:** syncthing should detect malformed addresses and display an error message, especially addresses that are malformed in common ways

calmh:
### Purpose

Don't hang on malformed URLs.

### Testing

Manual.

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -601,14 +601,25 @@ func (s *service) CommitConfiguration(from, to config.Configuration) bool {
 			continue
 		}
 
-		if _, ok := s.listeners[addr]; ok {
-			seen[addr] = struct{}{}
-			continue
-		}
-
 		uri, err := url.Parse(addr)
 		if err != nil {
 			l.Infof("Parsing listener address %s: %v", addr, err)
+			continue
+		}
+
+		// Make sure we always have the canonical representation of the URL.
+		// This is for consistency as we use it as a map key, but also to
+		// avoid misunderstandings. We do not just use the canonicalized
+		// version, because an URL that looks very similar to a human might
+		// mean something entirely different to the computer (e.g.,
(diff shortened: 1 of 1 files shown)


lib/connections/service.go
```diff
@@ -601,14 +601,25 @@ func (s *service) CommitConfiguration(from, to config.Configuration) bool {
...
-		if err != nil {
-			l.Infof("Parsing listener address %s: %v", addr, err)
+		// Make sure we always have the canonical representation of the URL.
+		// This is for consistency as we use it as a map key, but also to
+		// avoid misunderstandings. We do not just use the canonicalized
+		// version, because an URL that looks very similar to a human might
+		// mean something entirely different to the computer (e.g.,
+		// tcp:/127.0.0.1:22000 in fact being equivalent to tcp://:22000).
+		if canonical := uri.String(); canonical != addr {
+			l.Warnf("Skipping malformed listener URL %q (not canonical)", addr)
```
imsodin, line 617: I'd print the canonical string to save the tiny brainwork of finding the mistake.

calmh: I considered doing that, but in the case from the original issue that would be `tcp:///127.0.0.1:22000` which would just cement the mistake

PR #6700, opened 2020-06-02
build: Add "changelog" command

calmh:
### Purpose

Adds a command to build.go that parses tags and tag messages to create a decent looking changelog. The purpose is to add something like `go run build.go --num-versions=10 changelog > CHANGES.txt` to the build/release process and thus automatically include a changelog in the release.

(Next sub-step is to also copy that file in to `gui/default/` and add a thing to show it in a notice on first start after upgrade.)

### DIFF
--- a/build.go
+++ b/build.go
@@ -49,6 +49,7 @@ var (
 	debugBinary   bool
 	coverage      bool
 	timeout       = "120s"
+	numVersions   = 5
 )
 
 type target struct {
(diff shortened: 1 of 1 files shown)


build.go
```diff
@@ -1240,3 +1256,62 @@ func protobufVersion() string {
...
+	return vers[:n], nil
+}
+
+func reverseStrings(ss []string) {
+	for i := 0; i < len(ss)/2; i++ {
+		ss[i], ss[len(ss)-1-i] = ss[len(ss)-1-i], ss[i]
+	}
+}
+
+func filterStrings(ss []string, op func(string) bool) []string {
```
imsodin, line 1284: op?

calmh: ..erator, or ..eration, as in the comparison operation...

PR #6703, opened 2020-06-02
lib/config: Add missing quic address in case of non-default port (fixes #6679)

### LINKED ISSUE #6679: quic listener is not added when default tcp port is blocked
If the default port is blocked (by another instance of syncthing), only the tcp and relay listeners are added in the config which is quite obvious from the code:
https://github.com/syncthing/syncthing/blob/baa38eea7a848ce693d8187e1a7a3ddb6a7fa55a/lib/config/config.go#L135-L138

I guess the issue is obvious enough to not need discussion in the forum.

artback:
This fix adds missing quic listener on same port as TCP incase of default port being blocked

fixes #6679

### DIFF
--- a/lib/config/config.go
+++ b/lib/config/config.go
@@ -133,7 +133,7 @@ func NewWithFreePorts(myID protocol.DeviceID) (Configuration, error) {
 		cfg.Options.RawListenAddresses = []string{"default"}
 	} else {
 		cfg.Options.RawListenAddresses = []string{
-			fmt.Sprintf("tcp://%s", net.JoinHostPort("0.0.0.0", strconv.Itoa(port))),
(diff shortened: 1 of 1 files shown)


lib/config/config.go
```diff
@@ -135,6 +135,7 @@ func NewWithFreePorts(myID protocol.DeviceID) (Configuration, error) {
 		cfg.Options.RawListenAddresses = []string{
 			fmt.Sprintf("tcp://%s", net.JoinHostPort("0.0.0.0", strconv.Itoa(port))),
```
AudriusButkevicius, line 136:
```suggestion
			util.Address("tcp", net.JoinHostPort("0.0.0.0", strconv.Itoa(port))),
```

AudriusButkevicius:
Thanks

PR #6709, opened 2020-06-03
lib/build: Add some env vars as synthetic build tags

calmh:
This adds some env vars to the long version string as if they were build
tags. The purpose is to better understand what code was running or not
in the version output, usage reporting and crash reports. In order to
prevent possible privacy issues the actual value of the variable is not
reported, just the fact that it was set to something non-empty.

Example:

	% ./bin/syncthing --version
	syncthing v1.6.1+47-g1eb104f3-buildtags "Fermium Flea" (go1.14.3 darwin-amd64) jb@kvin.kastelo.net 2020-06-03 07:25:46 UTC [stnoupgrade, use_badger]

### DIFF
--- a/lib/build/build.go
+++ b/lib/build/build.go
@@ -89,7 +89,7 @@ func LongVersionFor(program string) string {
 	date := Date.UTC().Format("2006-01-02 15:04:05 MST")
 	v := fmt.Sprintf(`%s %s "%s" (%s %s-%s) %s@%s %s`, program, Version, Codename, runtime.Version(), runtime.GOOS, runtime.GOARCH, User, Host, date)
 	for _, envVar := range envTags {
-		if v := os.Getenv(envVar); v != "" {
+		if os.Getenv(envVar) != "" {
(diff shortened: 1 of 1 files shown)


lib/build/build.go
```diff
@@ -79,6 +88,11 @@ func LongVersionFor(program string) string {
 	// This string and date format is essentially part of our external API. Never change it.
 	date := Date.UTC().Format("2006-01-02 15:04:05 MST")
 	v := fmt.Sprintf(`%s %s "%s" (%s %s-%s) %s@%s %s`, program, Version, Codename, runtime.Version(), runtime.GOOS, runtime.GOARCH, User, Host, date)
+	for _, envVar := range envTags {
+		if os.Getenv(envVar) != "" {
+			Tags = append(Tags, strings.ToLower(envVar))
```
AudriusButkevicius, line 93: Do we really want this as tags? Tags are usually good for filtering, this is probably just "extra" data?

PR #6710, opened 2020-06-03
cmd/stcrashreceiver: Pick up extra build tags and send to Sentry

calmh:
This extracts the extra tags from any `[foo]` stuff at the end of the
version and sends them to Sentry for indexing.

If I need to modify that regexp again I'll probably write a from scratch
tokenizer and parser for our version string instead...

### DIFF
--- a/cmd/stcrashreceiver/sentry.go
+++ b/cmd/stcrashreceiver/sentry.go
@@ -149,7 +149,7 @@ func parseReport(path string, report []byte) (*raven.Packet, error) {
 		pkt.Tags = append(pkt.Tags, raven.Tag{Key: "commit", Value: version.commit})
 	}
 	for _, tag := range version.extra {
-		pkt.Tags = append(pkt.Tags, raven.Tag{Key: "extra", Value: tag})
+		pkt.Tags = append(pkt.Tags, raven.Tag{Key: tag, Value: "1"})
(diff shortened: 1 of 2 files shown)


cmd/stcrashreceiver/sentry.go
```diff
@@ -148,22 +148,26 @@ func parseReport(path string, report []byte) (*raven.Packet, error) {
 	if version.commit != "" {
 		pkt.Tags = append(pkt.Tags, raven.Tag{Key: "commit", Value: version.commit})
 	}
+	for _, tag := range version.extra {
+		pkt.Tags = append(pkt.Tags, raven.Tag{Key: "extra", Value: tag})
```
AudriusButkevicius, line 152: I think it's a map serverside, so you want unique keys?

calmh: Are you sure? I haven't tried but I tried to google it and it seemed like multiple values should be supported. It's a bit of a pain in the ass otherwise, probably better then to send a single joined value

AudriusButkevicius: The idea is that it would should you the distribution between tags, so I am not sure what the distribution of "extra" means.

I think you can just do `badger=1`, `noupgrade=1`

PR #6713, opened 2020-06-04
gui: Ignore permissions isn't just about FAT

imsodin:
This will require a lot of re-translation, but I really think it's worth it - too many people interpreting the FAT as an exclusive reason to enable it.


gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -234,7 +234,7 @@
                 <div class="col-md-6 form-group">
                   <label translate>Permissions</label><br />
                   <input type="checkbox" ng-model="currentFolder.ignorePerms" /> <span translate>Ignore</span>
-                  <p translate class="help-block">File permission bits are ignored when looking for changes. Use on FAT file systems.</p>
+                  <p translate class="help-block">Access permissions are ignored when looking for changes. Useful on systems without (e.g. FAT) or incompatible permissions (e.g. synology).</p>
```
AudriusButkevicius, line 237: Access permissions is ambiguous of access to what, it's file permissions.
The "useful" makes it suggestive, where as "use" made it more instructional 

`systems with (ex. FAT)` looks like a cliff hanger.

Use on file systems with non-standard permissions (e.g FAT, synology)

calmh:
It's not just when looking for changes, it's also that we don't set them when syncing...

`Disables comparing and syncing file permissions. Useful on systems with nonexistent or custom permissions (e.g., FAT, exFAT, Synology, Android).`

gui/default/assets/lang/lang-en.json
```diff
@@ -151,6 +153,7 @@
    "Help": "Help",
    "Home page": "Home page",
    "However, your current settings indicate you might not want it enabled. We have disabled automatic crash reporting for you.": "However, your current settings indicate you might not want it enabled. We have disabled automatic crash reporting for you.",
+   "If you want to prevent other users on this computer to access Syncthing and through it your files, consider setting up authentication.": "If you want to prevent other users on this computer to access Syncthing and through it your files, consider setting up authentication.",
```
greatroar, line 156: "If you want to prevent other users on this computer from accessing Syncthing (and through it, your files)" ...

calmh: Despite how it looks, that phrase is not part of this PR. Though if it were to be changed, now would be a good time (before it gets too ingrained)

imsodin: https://github.com/syncthing/syncthing/pull/6716

PR #6443, opened 2020-03-22
all: Store pending devices and folders in database (fixes #7178)

acolomb:
### Purpose

As discussed in #5758 and mentioned [on the forum](https://forum.syncthing.net/t/manual-intervention-for-introducer-feature/12212/27?u=acolomb), storing the pending (offered from remote but not yet added locally) devices and folders in the XML configuration is not a nice and scalable design.  Instead, the information should live in the database, properly structured, and made available over dedicated API endpoints.

This is also ground work and practice to finally approach an acceptable implementation of the prototype in #5758, extending the concept to other devices we have heard about in `ClusterConfig` messages.  That is deliberately saved for another PR though.

### Testing

All code paths have undergone extensive manual testing within an existing setup of four instances (one compiled from this PR). Especially regarding the [clean-up of pending entries](https://github.com/syncthing/syncthing/pull/6443#issuecomment-631100468) upon config changes, both online through the `POST /rest/system/config` REST API and offline by modifying the XML while Syncthing is not running. Notifications on the GUI look as before, API endpoints have been verified with curl.

~~Unit tests would be rather complicated and were discussed as not strictly needed considering how low the importance of the handled information is.~~  Unit tests included, with a few basic test cases.

### Documentation

The envisioned API for this stuff is described in https://github.com/syncthing/docs/pull/498.

The commit messages contain much of the rationale for each change, so I'd be happy to keep them intact instead of squash-merging in a single commit.  If desired, I can rebase to clean out obvious fixup commits and end up with a nice patch series of self-contained changes.

### Breaking Changes For The User

Existing `<pendingDevice>` / `<pendingFolder>` entries in the XML config are not carried over to the database.  The corresponding notifications will disappear until the next connection attempt with the offering device.  It has been discussed that the low importance of this information does not warrant a separate logic for config --> database info migrations.  One possibility to minimize disruption would be to just keep the XML entries intact for some releases and hold back on commit 8ae24612397fea6182cfb3d6dcce592c78c6df5d until then.

PR #6714, opened 2020-06-04
lib/protocol: Avoid panic in DeviceIDFromBytes

acolomb:
### Purpose

Return an error from DeviceIDFromBytes() if the length does not match,
instead of panicking.

Change all callers to handle the potential error value.  If possible,
pass it up or handle the condition gracefully, and panic immediately
after the function call in case there is no obvious better reaction.

Based on a related discussion in #6443.

### Testing

Test suite passes, no functional changes expected.

### DIFF
--- a/cmd/strelaysrv/listener.go
+++ b/cmd/strelaysrv/listener.go
@@ -151,7 +151,11 @@ func protocolConnectionHandler(tcpConn net.Conn, config *tls.Config) {
 			case protocol.ConnectRequest:
 				requestedPeer, err := syncthingprotocol.DeviceIDFromBytes(msg.ID)
 				if err != nil {
-					panic(err)
+					if debug {
+						log.Println(id, "is looking for an invalid peer ID")
+					}
+					conn.Close()
+					continue
 				}
 				outboxesMut.RLock()
 				peerOutbox, ok := outboxes[requestedPeer]
--- a/cmd/strelaysrv/listener.go
+++ b/cmd/strelaysrv/listener.go
@@ -149,7 +149,10 @@ func protocolConnectionHandler(tcpConn net.Conn, config *tls.Config) {
 				protocol.WriteMessage(conn, protocol.ResponseSuccess)
 
 			case protocol.ConnectRequest:
-				requestedPeer := syncthingprotocol.DeviceIDFromBytes(msg.ID)
+				requestedPeer, err := syncthingprotocol.DeviceIDFromBytes(msg.ID)
+				if err != nil {
+					panic(err)
+				}
 				outboxesMut.RLock()
 				peerOutbox, ok := outboxes[requestedPeer]
 				outboxesMut.RUnlock()
--- a/cmd/strelaysrv/listener.go
+++ b/cmd/strelaysrv/listener.go
@@ -154,6 +154,7 @@ func protocolConnectionHandler(tcpConn net.Conn, config *tls.Config) {
 					if debug {
 						log.Println(id, "is looking for an invalid peer ID")
 					}
+					protocol.WriteMessage(conn, protocol.ResponseNotFound)
 					conn.Close()
 					continue
 				}
--- a/lib/db/meta.go
+++ b/lib/db/meta.go
@@ -52,7 +52,11 @@ func (m *metadataTracker) Unmarshal(bs []byte) error {
 
 	// Initialize the index map
 	for i, c := range m.counts.Counts {
-		m.indexes[metaKey{protocol.DeviceIDFromBytes(c.DeviceID), c.LocalFlags}] = i
+		dev, err := protocol.DeviceIDFromBytes(c.DeviceID)
+		if err != nil {
+			return err
+		}
+		m.indexes[metaKey{dev, c.LocalFlags}] = i
 	}
 	return nil
 }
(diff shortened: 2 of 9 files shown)


cmd/strelaysrv/listener.go
```diff
@@ -149,7 +149,10 @@ func protocolConnectionHandler(tcpConn net.Conn, config *tls.Config) {
 				protocol.WriteMessage(conn, protocol.ResponseSuccess)
 
 			case protocol.ConnectRequest:
-				requestedPeer := syncthingprotocol.DeviceIDFromBytes(msg.ID)
+				requestedPeer, err := syncthingprotocol.DeviceIDFromBytes(msg.ID)
+				if err != nil {
+					panic(err)
```
AudriusButkevicius, line 154: I am not sure we want to take the relay down because someone sent a malicious message.

acolomb: Neither am I. In this case it just mirrors the status quo, where `DeviceIDFromBytes()` itself would `panic()`.

What do you suggest we do instead?  I don't know the relay server code at all, so didn't see an obvious better alternative.

calmh: If possible, drop the connection and continue whatever the outer loop is / return from the handler (I didn't look closely).

AudriusButkevicius: I've noticed there are a few cases where we moved the panic out which we might have to reconsider.

acolomb: > If possible, drop the connection and continue whatever the outer loop is / return from the handler (I didn't look closely).

I'm not quite sure I understand the Go spec correctly.  When panicking inside a goroutine, will the whole program crash or just that specific routine?  If it is the latter, then the `panic()` call is appropriate, aborting the current request and continuing the connection accept loop.

calmh: No, the program crashes, unless specific circumstances (which don't apply here).

imsodin: Panic brings down everything, not just a goroutine, and shouldn't be used for control flow (like e.g. python exceptions). Closing the connection and aborting (return, break, ) on `err != nil` is the simple and correct way to go here.

acolomb: Something along these lines: 1fb2d81a318a840446c2ca3c0e754883e66b2b5d?

lib/relay/protocol/packets.go
```diff
@@ -56,7 +56,11 @@ type SessionInvitation struct {
 }
 
 func (i SessionInvitation) String() string {
-	return fmt.Sprintf("%s@%s", syncthingprotocol.DeviceIDFromBytes(i.From), i.AddressString())
+	device := "<invalid>"
+	if address, err := syncthingprotocol.DeviceIDFromBytes(i.From); err == nil {
+		device = address.String()
+	}
+	return fmt.Sprintf("%s@%s", device, i.AddressString())
```
acolomb, line 63: I'm not sure where this is even used.  Could the `<` and `>` characters cause trouble when used through the `GoStringer` interface?

lib/db/meta.go
```diff
@@ -392,7 +396,10 @@ func (m *countsMap) devices() []protocol.DeviceID {
 
 	for _, dev := range m.counts.Counts {
 		if dev.Sequence > 0 {
-			id := protocol.DeviceIDFromBytes(dev.DeviceID)
+			id, err := protocol.DeviceIDFromBytes(dev.DeviceID)
+			if err != nil {
+				panic(err)
```
acolomb, line 401: This "should never happen" I guess.  Unless there is database corruption or some piece of code put garbage in these DB entries.  IMO panicking is almost never the right thing to do, but I don't know this part of the code well enough to propose a better approach.

cmd/strelaysrv/listener.go
```diff
@@ -151,7 +151,11 @@ func protocolConnectionHandler(tcpConn net.Conn, config *tls.Config) {
 			case protocol.ConnectRequest:
 				requestedPeer, err := syncthingprotocol.DeviceIDFromBytes(msg.ID)
 				if err != nil {
-					panic(err)
+					if debug {
+						log.Println(id, "is looking for an invalid peer ID")
+					}
+					conn.Close()
```
AudriusButkevicius, line 157: add a `protocol.WriteMessage(conn, protocol.ResponseNotFound)` maybe, to be nicer to the caller.

acolomb: Done. 8b3270bd006c780354acddd65667edcac8c4589a

imsodin:
@AudriusButkevicius What did you mean by this?

> I've noticed there are a few cases where we moved the panic out which we might have to reconsider.

Otherwise looks good to merge to me.

acolomb:
Another try to clarify the cleanup semantics, this time as a flowchart (independent of maps / slices / whatever):

pending cleanup flowchart_2020-06-07

Note how the data needed for PendingDevices is an inverted (**unknown** vs. **known**) subset of the data collected for PendingFolders.  That's why I reuse the same DropListObserved.  The only other difference is **ignored** device IDs, which are always dropped of course.  My implementation adds them between the two cleanup loops.

So I'm full of hope that with better function / type names and comments the algorithm itself can stay.  As long as the basic design of collecting decisions first and then doing two passes for the different key types is acceptable.

PR #6715, opened 2020-06-05
lib/scanner: Save one stat call per file

imsodin:
Stating every file in the scanner seems entirely superfluous, as we are already given the file info when walking the filesystem.

### DIFF
--- a/lib/scanner/walk.go
+++ b/lib/scanner/walk.go
@@ -265,7 +265,7 @@ func (w *walker) walkAndHashFiles(ctx context.Context, toHashChan chan<- protoco
 
 		if ignoredParent == "" {
 			// parent isn't ignored, nothing special
-			return w.handleItem(ctx, path, toHashChan, finishedChan, skip)
+			return w.handleItem(ctx, path, info, toHashChan, finishedChan, skip)
 		}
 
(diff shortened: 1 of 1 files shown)


lib/scanner/walk.go
```diff
@@ -274,17 +274,20 @@ func (w *walker) walkAndHashFiles(ctx context.Context, toHashChan chan<- protoco
...
+			return w.handleItem(ctx, path, info, toHashChan, finishedChan, skip)
 		}
 
 		// The previously ignored parent directories of the current, not
 		// ignored path need to be handled as well.
-		if err = w.handleItem(ctx, ignoredParent, toHashChan, finishedChan, skip); err != nil {
-			return err
-		}
-		for _, name := range strings.Split(rel, string(fs.PathSeparator)) {
+		for _, name := range append([]string{""}, strings.Split(rel, string(fs.PathSeparator))...) {
```
calmh, line 282: explain the change?

imsodin: That's the removed `handleItem` outside of the loop. It was a duplicate call before, and now it would be even more duplication (due to stating), thus I opted for the admittedly slightly hacky option of adding an empty string to handle the base path too. Added a comment.

PR #6717, opened 2020-06-05
lib: Handle case-insensitive filesystems

imsodin:
https://github.com/syncthing/syncthing/wiki/Filesystem-Case-Sensitivity#detect-real-case-proposal


imsodin:
All the basic pieces for handling case-insensitivity are in place now. Disclaimer: I haven't done any testing, the `DirNames` approach to getting the right case might be super slow and there seems to be a problem on windows with `osutil.IsDeleted` (will investigate).

imsodin:
I ran the many files integration benchmark twice on master and this branch, and the results are all over the place. At least it's not obviously extremely much worse. However I'll run them later on an beefier, more idle machine than my laptop to hopefully get a bit more consistent results:

```
PR:

2020/06/15 13:06:09 Result: Wall time: 944.392072ms / MiB
2020/06/15 13:06:09 Result: 1.1e+03 KiB/s synced
2020/06/15 13:06:09 Receiver: Utime: 425.885258ms / MiB
2020/06/15 13:06:09 Receiver: Stime: 208.14287ms / MiB
2020/06/15 13:06:09 Receiver: MaxRSS: 98868 KiB
2020/06/15 13:06:09 Sender: Utime: 167.71058ms / MiB
2020/06/15 13:06:09 Sender: Stime: 44.580516ms / MiB
2020/06/15 13:06:09 Sender: MaxRSS: 101172 KiB

2020/06/15 14:02:21 Result: Wall time: 1.390849093s / MiB
2020/06/15 14:02:21 Result: 745 KiB/s synced
2020/06/15 14:02:21 Receiver: Utime: 716.717322ms / MiB
2020/06/15 14:02:21 Receiver: Stime: 391.476258ms / MiB
2020/06/15 14:02:21 Receiver: MaxRSS: 104944 KiB
2020/06/15 14:02:21 Sender: Utime: 222.357967ms / MiB
2020/06/15 14:02:21 Sender: Stime: 75.986225ms / MiB
2020/06/15 14:02:21 Sender: MaxRSS: 81760 KiB


master:

2020/06/15 13:20:16 Result: Wall time: 1.502113108s / MiB
2020/06/15 13:20:16 Result: 690 KiB/s synced
2020/06/15 13:20:16 Receiver: Utime: 478.989451ms / MiB
2020/06/15 13:20:16 Receiver: Stime: 242.398516ms / MiB
2020/06/15 13:20:16 Receiver: MaxRSS: 103960 KiB
2020/06/15 13:20:16 Sender: Utime: 184.897419ms / MiB
2020/06/15 13:20:16 Sender: Stime: 52.640032ms / MiB
2020/06/15 13:20:16 Sender: MaxRSS: 91844 KiB

2020/06/15 13:26:46 Result: Wall time: 1.375864329s / MiB
2020/06/15 13:26:46 Result: 754 KiB/s synced
2020/06/15 13:26:46 Receiver: Utime: 539.303741ms / MiB
2020/06/15 13:26:46 Receiver: Stime: 271.998967ms / MiB
2020/06/15 13:26:46 Receiver: MaxRSS: 97320 KiB
2020/06/15 13:26:46 Sender: Utime: 195.600322ms / MiB
2020/06/15 13:26:46 Sender: Stime: 56.701838ms / MiB
2020/06/15 13:26:46 Sender: MaxRSS: 96024 KiB
```

imsodin:
There's now a few basic unit tests and I tried syncing `test` and `Test` to windows and got:

grafik

Looks like it's working :)

imsodin:
I just had a somewhat lengthy debugging session after having issues with using this branch in production and it looks like there's a fundamental problem with the "lock-step-scanning":  
Walking in lockstep to detect deletions requires a common ordering for walking filesystem and database. However the filesystem is ordered lexically by path components while the db is ordered lexically by path. E.g. fs goes a -> a/a -> b-a while the db goes a -> b-a -> a/a.

Unless someone has a good idea how to make that a non-issue (I don't want to cache walk results to re-order a different way), I'll dig a deeper grave for the lockstepping idea this time and do a tree based case-insensitive deletion detection:  
Instead of getting the real path and stating every file in db, lazyly build a tree of fs path components while checking items in db. I.e. whenever a db item contains path components for which `DirNames` hasen't bee called yet, do that and extend the tree with the results.

AudriusButkevicius:
`path.replace("/", "\x00")` problem solved 🤣

calmh:
I tested this with some case-only renames on my Mac, and

1. it actually seems to work, in that the end result is correct and the devices consider themselves up to date, but
2. there are a lot of errors logged along the way.

When renaming one file from `cli_test.go` to `cli_test.GO` the other sides logs this:

```
[MRIW7] 14:34:25 INFO: Puller (folder default, item "cli_test.GO"): name to sync (cli_test.GO) conflicts with name on disk (cli_test.go)
```

I'm not sure why. Misunderstanding of the check somewhere in the rename shortcut? We miss the rename case somehow and apply the create before the delete?

When renaming a folder full of files from `foo` to `FOO` the other side logs this:

```
...
[I6KAH] 14:36:03 INFO: Puller (folder default, item "FOO/http_test.go"): checking temporary sync path: name to sync (FOO/.syncthing.http_test.go.tmp) conflicts with name on disk (foo/.syncthing.http_test.go.tmp)
[I6KAH] 14:36:03 INFO: Puller (folder default, item "FOO/symlink_test.go"): checking temporary sync path: name to sync (FOO/.syncthing.symlink_test.go.tmp) conflicts with name on disk (foo/.syncthing.symlink_test.go.tmp)
...
```

(and so on for each file)

Maybe because the files are handled before the directory? (The directory actually does change case on disk in the end.)

imsodin:
We currently first collect files to be synced and to be deleted dirs, and at the same time create dirs and symlinks, then sync files, then delete. We could collect case conflicts when creating dirs, and then search the list of directories and files to be deleted for a "case-insensitive match". That's probably doable, but I guess quite cumbersome. A lighter option would be to not log errors until the last pull iteration. That has the drawback that for very long sync operations, errors impeding sync aren't displayed live in the UI, but only after the entire thing is done (which especially for initial setup use-cases can easily be hours, even more). Maybe delaying is for the best anyway: That will also prevent quite a few transient errors like `no such file` (request triggers scan on remote) or `device disappeared` (conn drop while syncing).

calmh:
Heh, tab completion handled the branch after initial 🙄 :)

Yeah, I think holding these errors until the last iteration is good enough (don't we already do something similar for something else?).

imsodin:
Ok, did that then. Meaning we now emit all errors at once at the end of pulling. Which highlights an aspect unrelated to this PR: Do we really want to print a line for every single item that has an error when pulling/scanning? Anyway, it's another issue/PR.

In addition I added an `fs.CachedRealCaser`, which caches `DirNames` results for 10 seconds or until exceeding 1 million strings (both values derived from magic), after that completely resets. I wrote an integration benchmark which should show potential effects, however I am hitting flakyness (event not encountered) and haven't managed to nail that down, so no benchs yet.

calmh:
I had an API thought that I'm not sure if we've already considered and discarded. What if we added RealCase to the success path of Stat/Lstat, so that the Name() on the returned FileInfo is the corrected case variant? (Kinda like you'd have hoped the filesystem worked to begin with) And added caching for this specific part of the Lstat just like you've done, as part of the default filesystem implementation. The calling code would then need to jump fewer hoops, but maybe it has other effects

imsodin:
I thought about that too, but consequently we'd have to wrap every single method on `Filesystem` in case resolving, making sure to get everything right. That's at the least cumbersome, probably also not simple to get right. Also caching in the filesystem layer by default is quite likely to be problematic if not considered by the calling code, thus I think it's better if any caching is explicit.

AudriusButkevicius:
I think artificially making the filesystem case sensitive is probably fine?
Sure, every open and what not becomes case sensitive and incurs a RealCase or whatever call, but it's easier to reason about correctness of this.

calmh:
Oh I was absolutely not suggesting trying to enforce case sensitive semantics on top of a case insensitive filesystem. I was just thinking we could bake in the RealCase call into Stat and not have a separate method for it. Everything else would still be the same, just that instead of `if RealCase(someFile) != someFile` checks we'd have `if info.Name() != someFile` after a stat... No real wrapping required, just a question of tweaking the returned basicFileInfo or what it's called.

AudriusButkevicius:
Why would it not be easier at the fs level? Everything works like before just blows up if you try to open/create a file with a different case.

calmh:
Apart from having to wrap everything as mentioned, I suspect that we will run into weird things if we start lying. Consider things like

```
info, err := fs.Stat(somefile)
if fs.IsNotExist(err) {
  // great, it doesn't exist
  // we can fs.OpenFile(..., O_EXCL)
  // or, no need to archive the old file because there is none, we can just rename on top of it
}
```

which will break spectacularly when I ask for `foo`, get "does not exist" then try to do stuff but at the *actual* filesystem level because `Foo` exists.

This seems like a world of pain where we'll always have to keep in mind that what the fs operations tell us might not be the truth and nothing makes sense any more

imsodin:
If performance of case-resolving isn't an issue, baking it into `FileInfo.Name()` would indeed be nice and clean. If caching is needed, we'd have to add some boolean argument to the stat functions, to allow caching or not, which isn't particularly nice either. I really need to get some benchmarks going: Given a directory may have a huge amount of children and `RealCase` needs to loop all of them, I would like to have some kind of data to asses the impact.

calmh:
I think we could have the caching you already implemented. It would only affect the case-ness of the returned name after all, not the existence etc which would come from the regular stat results...

(BTW did I dream or misread or was there a Windows-specific RealCase before that is now gone? Assuming that that's fast out of the box and doesn't need caching it seems like it would be nice to have a Windows specialization even if the general method works there too. That doesn't have to be part of this PR though, it can be added as a later optimization.)

AudriusButkevicius:
Do we really think that the surface of handling IsNotExistsButExistsWithDifferentCase would be bigger than the existing peppering and pushing through caching things?

Nobody else uses our fs abstraction, and in my head there are a handful of places where this would need to be handled (move temp to place, versioner archive, rename dirs)

calmh:
Certainly I might be wrong. It happens. :)

I think *if* we were going to go that route we should

1) move the current realcase stuff into Stat(), with caching. For implementation cleanliness we might do this as a RealCaseFileSystem that wraps the BasicFileSystem and is the new thing we use by default

2) make all fs operations there do the realcase thing as part of the canonicalization we already do and then return a new fs.ErrCaseConflict when appropriate. Looking at our `if fs.IsNotExist(...)` cases it's probably essential that this new error is *not* a NotExist but something else which will pass through to our default / weird thing happening branches... Not sure if we'll need a specific `fs.IsCaseConflict(error) bool` for somewhere, maybe?

This might work. Whether it's cleaner or not remains to be seen I guess. Imsodin? What do you think? You've thought this through in more detail than we have...

imsodin:
> (BTW did I dream or misread or was there a Windows-specific RealCase before that is now gone? Assuming that that's fast out of the box and doesn't need caching it seems like it would be nice to have a Windows specialization even if the general method works there too. That doesn't have to be part of this PR though, it can be added as a later optimization.)

You didn't dream. I removed that because I thought of the current api as filesystem agnostic, thus only relying on `Filesystem.DirNames`. That implementation can easily be resurfaced (should be in an earlier commit in this PR even).

The main motivation for my current approach is to prevent "surprises". Caching names by default and only `FileInfo.Name` returning the correct case should work fine, but it's not obvious that it happens. I know, ideally all of us know, but someone "new" to Syncthing would hardly realize it. That's obviously a pretty theoretical consideration to apply (in practice, it is the three of us or otherwise we are around),  
Nevertheless from this standpoint I like wrapping the entire filesystem. Then the behaviour is consistent again, one doesn't need to know which parts may consider case and which don't.

It should be fairly easy to push case resolving and caching into basicFilesystem and expose it on fileinfo.name. I'll do that and then have a look at what it means to make the entire basicFilesystem case-sensitive with case-specific errors and whether the gotchas jump at me right away or not. I'll do that later today.

imsodin:
> It should be fairly easy to push case resolving and caching into basicFilesystem and expose it on fileinfo.name.

For once it's a "good" thing I use "should" so much: It isn't actually easy.  
First minor issue: `FileInfo.Name()` returns just the base of the path (i.e. output of `filepath.Base(path)`). We already get that wrong in two mock FSes we have. However that simple to fix: Always return the full path from `Name`, if a consumer wants just the base, let them use filepath.  
The major problem is that to do case resolving lazily on `FileInfo`, the `Name` method needs to get an error return. And that's ugly (huge rats tail of changes and feels weird that after a successful stat `.Name` can still fail). Doing it non-lazily is 1. not so efficient and 2. means caching needs more care: Tests fail left and right because tests do stat, create something, stat again. And in the second stat it uses the now incorrect `DirNames` results. I.e. cache needs to be dropped whenever something is created. And that last point I haven't done yet, as I need a break (and at this point I'll probably think more about making the FS case-sensitive before going ahead, as the initially consider "simple" change has already become quite complex).

AudriusButkevicius:
I'd start by by making the  filesystems case sensitive.

Namely, Stat/Open fails if the case given is wrong. It could return a special error that satisfies both `fs.IsNotExists(err)` and `fs.IsCaseConflict(err)`, we could have `err.(*CaseConflictErr).RealName()` to access the real path if needed.
 
I think I'd start without caching, make it expensive on each operation but correct.

Things like Open(w+)/Create/Rename/Mkdir would fail if there is a conflicting file, we'd then do the `is.IsCaseConflict()` check, get the real name, do whatever db switcheroo we need to fix it up.

Then we can do caching, but as the famous saying goes, cache invalidation is hard.
Thinking about this I don't think we can have a cache at all, because even if I cache that
`aaa/bbb/ccc/ddd` is actually `aAa/bBb/cCc/dDd`, or some the correct case for part of the path,
I don't control the user renaming `bBb` to `BbB`, so I'd have no way of knowing that I need to invalidate.

calmh:
I feel this discussion is going around in circles. We can absolutely not make a case mismatch mean `fs.IsNotExists(err) == true` because it's not the truth and will result in bad things. Caching will be needed because as part of scanning the files in `a/b/c/d` we will need to listdir and iterate `a`, `a/b`, `a/b/c` and `a/b/c/d`. That's `O(n_files * n_parents * n_other_children)` and won't fly when it just looks like a stat call.

I think the better option here is to make sure the current code is good enough and captures the mechanics we want with an acceptable level of caching. (It probably already is that.) We can then look at possible cleaner implementations. I do not want a theoretical purity argument here to block the solution to a significant, five year old pain point.

AudriusButkevicius:
In our context we care about the case, and it is the truth that the file under that case does not exist. Same way like it does not exist in a case sensitive filesystem. This is what we care about for detecting files disappearing and detecting they were renamed etc, etc.
The fact you can't create a file with a different case under the same path is a different class of problem.

Sure, we can get something that is "good enough" out for the sake of getting something out, but as you yourself say, we'll have to live with this potentially forever.

Regardless, I am not the one sinking time into this, so you may do was you wish, but I feel that we're bleeding too much knowledge and understanding of this into the application logic as it stands.

calmh:
I'm really suspicious against returning ErrNotExist when there is something in place that would be clobbered by a subsequent create or write. Perhaps things will be fine because those operations will then return ErrCaseConflict and we'll error out weirdly. We can try. For that to work we don't actually need the Name() magic that imsodin already found was trickier than expected...

@imsodin since you're the one running the show here, would you mind trying a variant where we pretend case sensitivity at the fs layer and return errors when trying to access an existing file with the wrong case? To just see if the behavior works and makes sense, we can indeed skip caching in the first iteration... (or use precisely the realcase and caching you already have...)

AudriusButkevicius:
Why would we need Name() magic?

calmh:
I was agreeing that we don't, if the whole thing moves into fs.

AudriusButkevicius:
Sorry, missed the not part.

Yes, those operations will return ErrCaseConflict, and that's exactly what we'll have to start handling stuff in the application logic.
Any destructive operations (open, rename, etc) will just fail if the case is wrong.

imsodin:
I haven't read your discussion in detail, but I have a prototype since this morning of more or less what you seem to have converged on. Due to some urgent family business couldn't sit down yet to get it into a pushable form. My last comment was just trying to detail why "move `CachedRealCaser` into stat" isn't as straight forward as I though. I am now working on a wrapper fs for basicfilesystem which returns `ErrCase` for all relevant methods if the name doesn't match what's on disk (I haven't thought about naming at all, just used the first thing coming to mind). I will hopefully have something presentable within the next hour :)

imsodin:
Still haven't managed to get it to work fully, but pushing a wip anyway in case you want to have a look at where I am heading with the basic filesystem wrapper that returns a case error for case conflicts. For now to me it looks like this (wrapper fs) is roughly the way forward or otherwise what I had previously (this PR at the time of 7a37fb6 ), i.e. the separate `CachedRealCaser` for case resolving. I gotta run now, intend to continue at earliest opportunity (I want to get this working :D ).

AudriusButkevicius:
Seems like some files are missing.

lib/fs/basicfs_case.go
```diff
@@ -0,0 +1,409 @@
...
+	basicFile
+	caseFS *caseBasicFilesystem
+}
+
+func (f caseBasicFile) Stat() (FileInfo, error) {
+	if err := f.caseFS.checkCase(f.name); err != nil {
+		return nil, err
+	}
+	return f.basicFile.Stat()
+}
```
AudriusButkevicius, line 271: Not used. I guess it's needed if the file case changes after opening, yet I wonder if stat on a normal filesystem fails if the file is renamed, or if it's done on the file descriptor.

AudriusButkevicius:
This looks a lot cleaner now, up to the point that it's so clean I am not sure if it works anymore.

imsodin:
Finally, finally, finally it's working. Not sure whether it still qualifies for "so clean it's doubtful that it works at all" ;)

The diff grew because of caching respectively having to invalidate cache on every file creation (moving stuff from osutil to fs) and before scans (done by dropping cache on the start of every walk), as we otherwise might not pick up changes on fast successive scans. And what cost me an hour or two was that we copy folder configs all the time, which also creates a filesystem copy: Thus caching can't happen by filesystem object, but needs to happen by filesystem root (that's what the `casers` map is for).

I still don't have benchmarks. As calmh wrote above somewhere, I don't think this flies without some sort of caching. And in this PR caching is definitely scarier as it applies to the entire FS, while for the earlier approach (passing case resolver object around, #6812) caching is "confined" to that object.

Stuff left to do: Scour code for hardcoded `BasicFilesystem` instances and add an UI element ("I promise my filesystem is case-sensitive") to switch to the non-case-resolved basic fs (I don't intend to actually expose it as a choice of filesystems, that would be mightily confusing).

lib/fs/basicfs.go
```diff
@@ -352,3 +353,11 @@ func (e *ErrWatchEventOutsideRoot) Error() string {
 func (f *BasicFilesystem) newErrWatchEventOutsideRoot(absPath string, roots []string) *ErrWatchEventOutsideRoot {
 	return &ErrWatchEventOutsideRoot{fmt.Sprintf("Watching for changes encountered an event outside of the filesystem root: f.root==%v, roots==%v, path==%v. This should never happen, please report this message to forum.syncthing.net.", f.root, roots, absPath)}
 }
+
+func (f *BasicFilesystem) WriteFile(name string, content []byte, perm FileMode) error {
+	name, err := f.rooted(name)
```
AudriusButkevicius, line 358: Hell no. This doesn't make sense to have on the filesystem interface.

imsodin: It's not on the `Filesystem` interface. I don't know why `fs.BasicFilesystem` is exported, but it's only used as `fs.Filesystem`. This is exposed as a utility function, i.e. `fs.WriteFile(fs.Filesystem, ...)`, which is a replacement for `ioutil.WriteFile`. It's in the filesystem package because of cache invalidation and as it seems like the logical place for it to go.

AudriusButkevicius: We use this in tests only no? If yes, it should stay in tests, and we should not expand public interface for tests.

lib/fs/basicfs_case_unix.go
```diff
@@ -0,0 +1,167 @@
...
+	caseCacheTimeout = 10 * time.Second
+	// When the number of names (all lengths of []string from DirNames)
+	// exceeds this, we drop the cache.
+	caseMaxCachedNames = 1 << 20
+)
+
+var casers map[string]*realCaser
+
+func init() {
+	casers = make(map[string]*realCaser)
```
AudriusButkevicius, line 29: Can just do a make on the same line as var

lib/fs/basicfs.go
imsodin: We do use ioutil.WriteFile in api, which isn't "filesystemified". It just felt like a natural extension that might be useful earlier or later anyway so why not have it. However I agree as it's not needed now, it's better to have just a testing util function instead.

AudriusButkevicius: WriteFile is not a native filesystem construct so it should not be part of the contract. You don't have os.WriteFile, you have ioutil.WriteFile, which is some helper package.

lib/fs/basicfs_case_unix.go
```diff
@@ -0,0 +1,163 @@
...
+		case <-r.caseTimerStop:
+			if !r.caseTimer.Stop() {
+				<-r.caseTimer.C
+			}
+			r.caseMut.Lock()
+			r.caseTimerStop = nil
+			r.caseMut.Unlock()
+		}
+	}()
+}
```
AudriusButkevicius, line 124: This seems like a lot of moving parts.

I'd have a single timer, which is reset back to 10 seconds every time realCase is called.

Why is this complexity needed?

lib/fs/basicfs_case.go
```diff
@@ -0,0 +1,263 @@
...
+package fs
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+)
+
+type ErrCase struct {
```
AudriusButkevicius, line 16: ErrCaseConflict

lib/fs/basicfs_case.go
```diff
@@ -0,0 +1,263 @@
...
+	given, real string
+}
+
+func (e *ErrCase) Error() string {
+	return fmt.Sprintf(`given name "%v" differs from name in filesystem "%v"`, e.given, e.real)
+}
+
+func IsErrCase(err error) bool {
```
AudriusButkevicius, line 24: IsErrCaseConflict

lib/fs/basicfs_case.go
```diff
@@ -0,0 +1,263 @@
...
+	e := &ErrCase{}
+	return errors.As(err, &e)
+}
+
+// caseBasicFilesystem is a BasicFilesystem with additional checks to make a
+// potentially case insensitive underlying FS behave like it's case-sensitive.
+type caseBasicFilesystem struct {
```
AudriusButkevicius, line 31: Nothing makes this bound to basicFilesystem (except for the windows specialization), so this should just be `caseSensitiveFilesystem` encapsulating a general underlying fs.Filesystem.


The newRealCaser is then passed underlying fs.Filesystem and can have a special case:
```
if underlying.Type() == Basic && runtime.GOOS == "windows" {
   return windowsBasedRealCaser
} else {
   return generalRealCaserThatWorksOnAnyFilesysmte
}
```

imsodin: And then you add a `makeCaseSensitive bool` argument to `NewFilesystem`?

lib/model/folder_recvonly_test.go
```diff
@@ -386,18 +386,3 @@ func setupROFolder(t *testing.T) (*model, *receiveOnlyFolder) {
 
 	return m, f
 }
-
-func writeFile(fs fs.Filesystem, filename string, data []byte, perm fs.FileMode) error {
```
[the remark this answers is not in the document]
imsodin: I didn't notice that I moved it, which is kind of the reason: When undoing moving this to `fs.WriteFile` I put it where I instinctively put model helper functions: testutils_test.go. Essentially: There's no pressing reason to move (now), but testutils is a more appropriate place

lib/fs/basicfs_case_unix.go
imsodin: The thing that requires some shenanigans is that I don't want to reset an already running timer to make sure the cache actually gets invalidated after timeout. Given "something" is needed, I used a channel which doubles as early cancellation on explicit case invalidation.

lib/fs/basicfs_case.go
AudriusButkevicius: So it depends on how we spin this.

Currently we default to wrapping everything, with the option to switch it off.
We could make the logic for wrapping happen in NewFilesystem based on some runtime.GOOS logic.
We could make it an option on the folder config I guess, which we then push down to NewFilesystem
Or we could have a magic type.

The thing I don't like about the magic type is that it's not an implementation of it's own, it's just a wrapper around an existing implementation. Now if we ever add let's say s3 as an implementation, you'll have to have caseS3 etc as different types that act just as wrappers and don't provide any functionality themselves.

We have walkFs and mtimeFs which perform a simillar function, but there are not types you can instantiated.

Sure, they are somewhat not optional, hence I am questioning whether this should be, and if yes, under what circumstances.

imsodin: I am now spinning it similar to mtimefs, i.e. casefs is now filesystem agnostic and you get it by calling `fs.NewCaseFilesystem(fs Filesystem)`. That happens in `FolderConfiguration.Filesystem()`, depending on `FolderConfiguration.CaseSensitiveFS` (which is false by default, meaning case resolving happens). I expect this to be a checkbox in the edit folder modal in the advanced tab.

lib/fs/debug_symlink_unix.go
```diff
@@ -6,16 +6,23 @@
...
 // symlinks when non-elevated.
-func DebugSymlinkForTestsOnly(oldname, newname string) error {
-	return os.Symlink(oldname, newname)
+func DebugSymlinkForTestsOnly(oldFs, newFs Filesystem, oldname, newname string) error {
+	if err := os.Symlink(filepath.Join(oldFs.URI(), oldname), filepath.Join(newFs.URI(), newname)); err != nil {
+		return err
+	}
+	if caseFs, ok := unwrapFilesystem(newFs).(*caseBasicFilesystem); ok {
+		caseFs.cleanCase()
+	}
```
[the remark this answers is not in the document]
imsodin: Yep. I mean the whole debugsymlinkforwhatever thing might not be needed anymore, as maybe go has some support for it by now (?), but it is in use for now and the missing cache invalidation did brake tests.

calmh:
This currently doesn't pick up case-only renames on my setup. I know imsodin is away so this is not a prod in that direction, more a memory note so that I can try to fix it next week or so. (Currently also in vacation mode.)

imsodin:
What exactly is the scenario? There's unit tests for a local rename being picked up by scanning and a remote rename being synced, which means if it's either of those the test isn't functional.

imsodin:
I have this running on my laptop and a windows testing machine and it works as I expect (renames on both sides just propagate, creating case conflicts on linux results in a sync error on windows). Mac is a different pair of shoes (case-insensitive and default case resolver, while linux is case-sensitive and windows uses a custom case-resolver), but basic functionality (case renames) should be covered by tests.

Also the integration benchmark I added now runs consistently. It still does not show a significant change. It's probably not a particularly good/relevant benchmark though. I'd expect most performance impact for systems  on "weird" or network filesystems.

imsodin:
I just noticed a bit of fan activity after resuming from suspend and it was due to the initial scan of syncthing, respectively the case-resolving:

[profile001.pdf](https://github.com/syncthing/syncthing/files/4977822/profile001.pdf)

Looks like the lowercasing respectively moving between strings and runes is quite costly.

calmh:
Retesting this it now actually works for me. I don't know if this is because the cache changes or if I did something wrong last time or something is flaky. I'll however run it some more with the assumption that it's all good and get into some actual review from my side :)

calmh:
I added a benchmark to do a scanner-like walk (first walk the filesystem, then stat each entry as if we already had them in the database);

```
jb@kvin:..ing/syncthing/lib/fs % go test -bench . -run none -benchtime 10s
goos: darwin
goarch: amd64
pkg: github.com/syncthing/syncthing/lib/fs
BenchmarkWalkCaseFakeFS10k/raw-8                     288          41836273 ns/op               602 B/entry               0.432 allocs/entry           2141 ns/entry     11771535 B/op     348491 allocs/op
BenchmarkWalkCaseFakeFS10k/case-8                    159          79047286 ns/op              1201 B/entry               0.781 allocs/entry           4045 ns/entry     23473706 B/op     602603 allocs/op
```

One "op" here is a complete double-walk over a 10k files fakefs. There's more than 10k things looked at because there's a directory structure based on the random file names, so somewhere between 10k and 30k actual filesystem entries in total. The "/entry" numbers are per such filesystem entry.

So this adds about 600 bytes and 4000 ns overhead to each entry inspected. The 4000 ns is about a factor two of the non-CaseFS benchmark, but we should keep in mind that this is doing zero actual I/O or hashing or database operations so we are measuring just FakeFS overhead vs FakeFS+CaseFS overhead.

I think we're good here, performance wise.

lib/fs/casefs.go
```diff
@@ -0,0 +1,450 @@
...
+		node.results[comp] = child
+		node.children[n] = child
+		node = child
+		out = filepath.Join(out, n)
+	}
+
+	return out, nil
+}
+
+func (r *defaultRealCaser) startCaseResetTimerLocked() {
```
calmh, line 394: This thing could be slightly simplified by not using a timer and a routine for the rest, instead recording the creation time of the cache. If it's too old or too large on entry into realCase we drop the cache. The downside would be that a cache is not automatically expired when we stop calling realCase -- it would stick around until next call. A workaround for that could be a permanent ticker routine that once every few seconds checks if the cache is expired and removes it. This is similar to what happens now except there would be a permanent routine and ticker instead of the selects and stuff.

I don't think the current implementation is necessarily bad, but it is a few things to take in.

imsodin: I avoided a persistent goroutine because a filesystem never gets closed, thus we'd leak a goroutine every time a filesystem is (re-)created.

calmh: Good point

PR #6719, opened 2020-06-07
lib/connections, lib/nat: Correctly dis-/enable nat (fixes #6552)

### LINKED ISSUE #6552: panic: Stop called more than once on ... created by nat.Service
If disable NAT, enable NAT, then disable NAT again Syncthing will crash. The reason is that we repeatedly add and remove the service from a supervisor, which isn't a supported pattern since a service shouldn't be restarted after it's been explicitly stopped. ~~We should create a new service instead.~~ Oh man that thing is wired deeply into the connection stuff. Probably we need to not stop the entire service but just tell it to lay low?

Sentry Issue: [SYNCTHING-MS](https://sentry.syncthing.net/syncthing/syncthing/issues/683/?referrer=github_integration)

//@calmh

imsodin:
I added exported methods to control the nat service activity. The alternative would have been to subscribe the nat service to config changes, but this option felt a bit more straight forward/has less moving part.

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -123,7 +123,7 @@ type service struct {
 	tlsDefaultCommonName string
 	limiter              *limiter
 	natService           *nat.Service
-	natServiceToken      *suture.ServiceToken
+	natServiceEnabled    bool
 	evLogger             events.Logger
 
 	listenersMut       sync.RWMutex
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
Sorry, but this just adds a variable, and a bunch of methods that are never used?

Also, it's not clear to me why the old method did not work.

imsodin:
> If disable NAT, enable NAT, then disable NAT again Syncthing will crash. The reason is that we repeatedly add and remove the service from a supervisor, which isn't a supported pattern since a service shouldn't be restarted after it's been explicitly stopped.

AudriusButkevicius:
Sorry, hard to follow to tickets via phone.

I think handing the responsibility of the config value to the nat service makes more sense, reduces extra variables

AudriusButkevicius:
Glancing at the code, I think moving handling of this into the service itself is definately cleaner.

imsodin:
Finefine, done - I guess I missed my chance of sneaking in an easy, non-optimal fix by messing up the first commit :P

AudriusButkevicius:
I don't see any new commits, so not sure what "done" means here?

lib/connections/service.go
```diff
@@ -652,16 +651,6 @@ func (s *service) CommitConfiguration(from, to config.Configuration) bool {
 	}
 	s.listenersMut.Unlock()
 
-	if to.Options.NATEnabled && s.natServiceToken == nil {
-		l.Debugln("Starting NAT service")
-		token := s.Add(s.natService)
```
AudriusButkevicius, line 657: This should still happen somewhere.

PR #6723, opened 2020-06-08
lib/db: Remove unused keyer methods

greatroar:
This removes two unused methods.

### DIFF
--- a/lib/db/keyer_test.go
+++ b/lib/db/keyer_test.go
@@ -58,21 +58,9 @@ func TestGlobalKey(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	fld2, ok := db.keyer.FolderFromGlobalVersionKey(key)
-	if !ok {
-		t.Error("should have been found")
-	}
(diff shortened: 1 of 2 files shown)


lib/db/keyer_test.go
```diff
@@ -46,36 +46,6 @@ func TestDeviceKey(t *testing.T) {
...
-	}
-
-	fld2, ok := db.keyer.FolderFromGlobalVersionKey(key)
-	if !ok {
-		t.Error("should have been found")
-	}
-	if !bytes.Equal(fld2, fld) {
-		t.Errorf("wrong folder %q != %q", fld2, fld)
-	}
-	name2 := db.keyer.NameFromGlobalVersionKey(key)
```
imsodin, line 68: I wouldn't remove the whole test - this function is still in use.

greatroar: Restored.

greatroar:
Failing test seems unrelated.

PR #6726, opened 2020-06-09
all: Make all error implementations pointer types

greatroar:
This matches the convention of the stdlib ([json.MarshalerError](https://godoc.org/encoding/json#MarshalerError), [net.AddrError](https://godoc.org/net#AddrError), [os.PathError](https://godoc.org/os#PathError)) and avoids ambiguity: when customErr{} and &customErr{} both implement error, client code needs to check for both. The current lib/db/backend.{IsClosed,IsNotFound} do exactly that.

Memory use should remain the same, since storing a non-pointer type in an interface value still copies the value to the heap:

```
// foo.go
package foo

type customErr struct{}

func (customErr) Error() string { return "" }

func Foo() error { return customErr{} }
```

```
$ go build -gcflags=-m foo.go
# command-line-arguments
./foo.go:6:6: can inline customErr.Error
./foo.go:8:6: can inline Foo
./foo.go:8:36: customErr literal escapes to heap
<autogenerated>:1: inlining call to customErr.Error
<autogenerated>:1: .this does not escape
<autogenerated>:1: leaking param: .this
```

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -522,7 +522,7 @@ type errNoUpgrade struct {
 	current, latest string
 }
 
-func (e errNoUpgrade) Error() string {
+func (e *errNoUpgrade) Error() string {
 	return fmt.Sprintf("no upgrade available (current %q >= latest %q).", e.current, e.latest)
 }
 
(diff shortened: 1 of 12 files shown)


greatroar:
I also just noticed that in lib/relay/client/methods.go, there is an "incorrect response code" error being constructed by fmt.Errorf, while the other errors with that message have a custom type. Is this an oversight?

https://github.com/syncthing/syncthing/blob/53af64a2a4fc939d5758c5a7c34248eed63cc6de/lib/relay/client/methods.go#L108

AudriusButkevicius:
I don't think anything acts on the type, so it probably doesn't matter much.

greatroar:
There's a check for it in TestRelay.

greatroar:
I caught one more error type, lib/discover.lookupError. I changed that from

```go
type lookupError struct {
	error
	cacheFor time.Duration
}
```

to

```go
type lookupError struct {
	msg      string
	cacheFor time.Duration
}
```

with a separate Error method. The thing here is that embedding the error type implicitly adds an Error method to both the pointer and the non-pointer types.

PR #6727, opened 2020-06-09
Implement encrypted devices (continued)

imsodin:
This PR is the continuation/equivalent of https://github.com/syncthing/syncthing/pull/6214 on a branch in the syncthing/syncthing repo. Everything on there still applies.

For a list of ongoing/outstanding tasks and more notes, please refer to this wiki page (and extend/update as appropriate): https://github.com/syncthing/syncthing/wiki/Untrusted-devices-(encryption)-status

Companion docs PR: https://github.com/syncthing/docs/pull/480


ya-isakov:
Am I right, that file names and so on are encrypted with AES, but file content is encrypted with XChaCha20-Poly1305? Documentation states that AES-GCM is used everywhere.
P.S. I'm completely agree with usage of ChaCha20, just trying to clarify a bit.

imsodin:
@ya-isakov 

> Am I right, that file names and so on are encrypted with AES, but file content is encrypted with XChaCha20-Poly1305? Documentation states that AES-GCM is used everywhere.

The documentation does reference chacha where it's used as far as I can see: https://deploy-preview-480--syncthing-docs.netlify.app/specs/untrusted.html

ya-isakov:
But in this [documentation](https://docs.syncthing.net/branch/untrusted/html/users/untrusted.html) it is said that
`Encryption is AES-256-GCM with a key derived from the password and folder ID using scrypt.`

imsodin:
Right, I was looking at the spec. The user docs indeed need updating in many places, I'll replace the gcm reference with something mentioning chacha and aes-siv. Feel free to comment (or contribute) directly on the docs PR if you notice anything else: https://github.com/syncthing/docs/pull/480

@AudriusButkevicius I was able to share files between an untrusted test instance with this PR, i.e. this is now ready for review if you'd like to dive in ;)

AudriusButkevicius:
Started going through this, but as expected this is large and will take a while. 

What state do we believe this is in, is there stuff left to do?

imsodin:
Of course.

From my side there's nothing left to do on this PR. As I wrote above, I need to update user docs, mostly to add set up instructions with screenshots (plus some info might be outdated). I intend to do that soon (as in by wednesday for sure).

lib/model/model.go
```diff
@@ -1564,12 +1764,15 @@ func (m *model) Request(deviceID protocol.DeviceID, folder, name string, size in
 	if err := readOffsetIntoBuf(folderFs, name, offset, res.data); fs.IsNotExist(err) {
 		l.Debugf("%v REQ(in) file doesn't exist: %s: %q / %q o=%d s=%d", m, deviceID, folder, name, offset, size)
 		return nil, protocol.ErrNoSuchFile
+	} else if err == io.EOF && len(hash) == 0 {
+		// Read beyond end of file when we can't verify the hash -- this is
+		// a padded read for an encrypted file. It's fine.
```
AudriusButkevicius, line 1769: How does this work? We lie about the block sizes in encrypted files, so clients ask to read more than what's available?

lib/model/model.go
```diff
@@ -1057,9 +1153,40 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
 		if !ok {
 			// Shouldn't happen because !cfg.Paused, but might happen
 			// if the folder is about to be unpaused, but not yet.
+			l.Debugln("ccH: no fset", folder.ID)
 			continue
 		}
 
+		ccDeviceRemote, hasDevice := ccDevicesRemote[folder.ID]
+		ccDeviceLocal, hasDeviceLocal := ccDevicesLocal[folder.ID]
```
AudriusButkevicius, line 1161: hasRemoteDevice, hasLocalDevice

lib/model/model.go
```diff
@@ -171,9 +174,16 @@ var (
...
+	errIgnoredFolderRemoved     = errors.New("folder no longer ignored")
+	errReplacingConnection      = errors.New("replacing connection")
+	errStopped                  = errors.New("Syncthing is being stopped")
+	errEncInvConfigLocal        = errors.New("can't encrypt data for a device when the folder type is receiveEncrypted")
+	errEncInvConfigRemote       = errors.New("remote has encrypted data and encrypts that data for us - this is impossible")
+	errEncNotEncryptedLocal     = errors.New("folder is announced as encrypted, but not configured thus")
+	errEncNotEncryptedRemote    = errors.New("folder is configured to be encrypted but not announced thus")
+	errEncNotEncryptedUntrusted = errors.New("device is untrusted, but configured to receive not encrypted data")
+	errEncPW                    = errors.New("different passwords used")
+	errEncReceivedToken         = errors.New("need to reinitialise after receiving encryption token")
```
AudriusButkevicius, line 186: wtf does this mean?

lib/model/model.go
```diff
@@ -1057,9 +1153,40 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
...
+		ccDeviceLocal, hasDeviceLocal := ccDevicesLocal[folder.ID]
+
+		if err := m.ccCheckEncryption(cfg, folderDevice, ccDeviceRemote, ccDeviceLocal, hasDevice, hasDeviceLocal, deviceCfg.Untrusted); err != nil {
+			sameError := false
+			if devs, ok := m.folderEncFailures[folder.ID]; ok {
+				sameError = devs[deviceID] == err
+			} else {
+				m.folderEncFailures[folder.ID] = make(map[protocol.DeviceID]error)
+			}
+			m.folderEncFailures[folder.ID][deviceID] = err
```
AudriusButkevicius, line 1170: IT almost feels like we're writing javascript now.
Maps of maps :(

I guess this can't be handled by the folder and the state kept on the folder, we have to keep piling it onto the model?

lib/model/model.go
```diff
@@ -1057,9 +1153,40 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
...
+			msg := fmt.Sprintf("Failure checking encryption consistency with device %v for folder %v: %v", deviceID, cfg.Description(), err)
```
AudriusButkevicius, line 1171: What is encryption consistency? and how do I restore it?

I think this talks about disagreement on encryption passwords not matching?

lib/model/model.go
```diff
@@ -1068,81 +1195,81 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
...
+
+		// This is the other side's description of themselves. We
+		// check to see that it matches the IndexID we have on file,
+		// otherwise we drop our old index data and expect to get a
+		// completely new set.
+
+		switch {
+		case !hasDevice:
+			l.Debugf("Device %v folder %s sent no info about themselves", deviceID, folder.Description())
+			fallthrough
```
AudriusButkevicius, line 1234: All this hasDevice hasDeviceLocal stuff seems just invalid state. I suspect we didn't allow it before, and now we're starting to handle this. Why?

lib/model/model.go
```diff
@@ -1151,46 +1278,94 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
...
-		if ok {
-			m.progressEmitter.temporaryIndexSubscribe(conn, tempIndexFolders)
-		}
+func (m *model) ccCheckEncryption(fcfg config.FolderConfiguration, folderDevice config.FolderDeviceConfiguration, ccDeviceRemote, ccDeviceLocal protocol.Device, hasDeviceRemote, hasDeviceLocal, deviceUntrusted bool) error {
+	hasTokenRemote := hasDeviceRemote && len(ccDeviceRemote.EncPwToken) > 0
+	hasTokenLocal := hasDeviceLocal && len(ccDeviceLocal.EncPwToken) > 0
+	isEncRemote := folderDevice.EncryptionPassword != ""
+	isEncLocal := fcfg.Type == config.FolderTypeReceiveEncrypted
+
+	if !isEncRemote && !isEncLocal && deviceUntrusted {
```
AudriusButkevicius, line 1290: This transience of untrustedness for encrypted folders is weird.
It seems if I have
A (plain) -> B (encrypted) -> C (encrypted), B still requires C to be marked as untrusted?

lib/model/model.go
```diff
@@ -1151,46 +1278,94 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
...
+		// Noone cares about encryption here
+		return nil
 	}
 
-	if changed {
-		if err := m.cfg.Save(); err != nil {
-			l.Warnln("Failed to save config", err)
+	if isEncRemote && isEncLocal {
+		// Should never happen, but config racyness and be safe.
+		return errEncInvConfigLocal
```
AudriusButkevicius, line 1301: I guess I am missing how this works, but I thought both sides have to agree on encryption?

lib/model/model.go
```diff
@@ -1151,46 +1278,94 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
...
+			if err := writeEncToken(ccToken, fcfg); err != nil {
+				return err
+			}
+			m.fmut.Lock()
+			m.folderEncPwTokens[fcfg.ID] = ccToken
+			m.fmut.Unlock()
+			// We can only announce ourselfs once we have the token,
+			// thus we need to resend CCs now that we have it.
+			m.closeConns(fcfg.DeviceIDs(), errEncReceivedToken)
+			return errEncReceivedToken
```
AudriusButkevicius, line 1363: Sorry, but I have no idea what this is. I can't seem to pin it to anything that was previously discussed.

lib/protocol/encryption.go
```diff
@@ -0,0 +1,521 @@
...
+	maxPathComponent      = 200              // characters
+	encryptedDirExtension = ".syncthing-enc" // for top level dirs
+	miscreantAlgo         = "AES-SIV"
+)
+
+// The encryptedModel sits between the encrypted device and the model. It
+// receives encrypted metadata and requests from the untrusted device, so it
+// must decrypt those and answer requests by encrypting the data.
+type encryptedModel struct {
+	Model
```
AudriusButkevicius, line 41: I wonder if this is not dangerous, and this should have a reference to a model but not embed it.

The moment we add a new method to the model, we'll automatically call through to the underlying model if we forget to implement it here, which probably not what we want?

lib/protocol/encryption.go
```diff
@@ -0,0 +1,521 @@
...
+	return rawResponse{enc}, nil
+}
+
+func (e encryptedModel) DownloadProgress(deviceID DeviceID, folder string, updates []FileDownloadProgressUpdate) error {
+	if _, ok := e.folderKeys[folder]; !ok {
+		return e.Model.DownloadProgress(deviceID, folder, updates)
+	}
+
+	// The updates contain nonsense names and sizes, so we ignore them.
+	return nil
```
AudriusButkevicius, line 117: I guess we can decrypt the names and adjust the sizes?

Or otherwise not send them at all?

lib/protocol/encryption.go
```diff
@@ -0,0 +1,521 @@
...
+	if _, ok := e.folderKeys[folder]; !ok {
+		e.Connection.DownloadProgress(ctx, folder, updates)
+	}
+
+	// No need to send these
+}
+
+func (e encryptedConnection) ClusterConfig(config ClusterConfig) {
+	// TODO: Filter/clean the outgoing ClusterConfig?
+	e.Connection.ClusterConfig(config)
```
AudriusButkevicius, line 191: Is the TODO still relevant?

lib/protocol/encryption.go
```diff
@@ -0,0 +1,521 @@
...
+
+func (e encryptedModel) ClusterConfig(deviceID DeviceID, config ClusterConfig) error {
+	// TODO: Filter/clean the incoming ClusterConfig?
+	return e.Model.ClusterConfig(deviceID, config)
+}
+
+// The encryptedConnection sits between the model and the encrypted device. It
+// encrypts outgoing metadata and decrypts incoming responses.
+type encryptedConnection struct {
+	Connection
```
AudriusButkevicius, line 128: Same comment as for model embedding.

lib/protocol/encryption.go
```diff
@@ -0,0 +1,521 @@
...
+		},
+	}
+
+	// Construct the fake block list. Each block will be blockOverhead bytes
+	// larger than the corresponding real one and have an encrypted hash.
+	// Very small blocks will be padded upwards to minPaddedSize.
+	//
+	// The encrypted hash becomes just a "token" for the data -- it doesn't
+	// help verifying it, but it lets the encrypted device do block level
+	// diffs and data reuse properly when it gets a new version of a file.
```
AudriusButkevicius, line 237: Not sure why but this triggers my security spider senses, because I know things like DARE, every time the file changes, everything in the files changes, so no reuse is possible.

I guess it's a compromise we're making, and that might be fine, but the nonce of the data changes, changing the encrypted payload, yet the fake hash stays the same, making me doubt the usefulness of the nonce, as effectively you know two differently looking encrypted payloads represent the same data, effectively providing an oracle.

proto/lib/config/folderconfiguration.proto
```diff
@@ -14,8 +14,9 @@ import "lib/fs/copyrangemethod.proto";
 import "ext.proto";
 
 message FolderDeviceConfiguration {
-    bytes device_id     = 1 [(ext.goname) = "DeviceID", (ext.xml) = "id,attr", (ext.json) = "deviceID", (ext.device_id) = true];
-    bytes introduced_by = 2 [(ext.xml) = "introducedBy,attr", (ext.device_id) = true];
+    bytes  device_id           = 1 [(ext.goname) = "DeviceID", (ext.xml) = "id,attr", (ext.json) = "deviceID", (ext.device_id) = true];
+    bytes  introduced_by       = 2 [(ext.xml) = "introducedBy,attr", (ext.device_id) = true];
+    string encryption_password = 3 ;
```
AudriusButkevicius, line 19: extra space before ;

proto/lib/protocol/bep.proto
```diff
@@ -74,6 +74,7 @@ message Device {
     bool            introducer                 = 7;
     uint64          index_id                   = 8 [(gogoproto.customname) = "IndexID", (gogoproto.customtype) = "IndexID", (gogoproto.nullable) = false];
     bool            skip_introduction_removals = 9;
+    bytes           enc_pw_token               = 10;
```
AudriusButkevicius, line 77: enc_pw_tkn or shorter ept for maximum obscurity please.

proto/lib/protocol/bep.proto
```diff
@@ -164,6 +166,7 @@ message Request {
     bytes  hash           = 6;
     bool   from_temporary = 7;
     uint32 weak_hash      = 8;
+    int32  block_no       = 9;
```
AudriusButkevicius, line 169: So this is for us to make it easier to find what's the real offset in the unencrypted file that we're after, but that's a deterministic function anyway, so I am not really use we need it?

It's going to be Offset/EncryptedBlockSize (where EncryptedBlockSize is deterministici as well)?

AudriusButkevicius (commented):
First round, I guess.

I don't get the whole random stuff written to a marker directory or the purpose that it serves.

gui/default/assets/lang/lang-en.json
```diff
@@ -60,6 +62,7 @@
    "Danger!": "Danger!",
    "Debugging Facilities": "Debugging Facilities",
    "Default Folder Path": "Default Folder Path",
+   "Delete Not Encrypted Items": "Delete Not Encrypted Items",
```
motiejus, line 65: Not Encrypted -> Plaintext?

Is "plaintext" user-friendly enough?

gui/default/assets/lang/lang-en.json
```diff
@@ -151,6 +155,8 @@
    "Introduced By": "Introduced By",
    "Introducer": "Introducer",
    "Inversion of the given condition (i.e. do not exclude)": "Inversion of the given condition (i.e. do not exclude)",
+   "Items have been added to this folder which are not encrypted.": "Items have been added to this folder which are not encrypted.",
+   "Items in Receive Encrypted": "Items in Receive Encrypted",
```
motiejus, line 159: Not sure what this means. Is 'Receive' a name of a place?

gui/default/assets/lang/lang-en.json
```diff
@@ -310,6 +321,7 @@
    "The following intervals are used: for the first hour a version is kept every 30 seconds, for the first day a version is kept every hour, for the first 30 days a version is kept every day, until the maximum age a version is kept every week.": "The following intervals are used: for the first hour a version is kept every 30 seconds, for the first day a version is kept every hour, for the first 30 days a version is kept every day, until the maximum age a version is kept every week.",
    "The following items could not be synchronized.": "The following items could not be synchronized.",
    "The following items were changed locally.": "The following items were changed locally.",
+   "The following unexpected items are present.": "The following unexpected items are present.",
```
motiejus, line 324: `Unexpected items found`?

gui/default/assets/lang/lang-en.json
```diff
@@ -374,11 +388,15 @@
    "You have no ignored folders.": "You have no ignored folders.",
    "You have unsaved changes. Do you really want to discard them?": "You have unsaved changes. Do you really want to discard them?",
    "You must keep at least one version.": "You must keep at least one version.",
+   "You should never add or change anything locally in a \"{%receiveEncrypted%}\" folder.": "You should never add or change anything locally in a \"{{receiveEncrypted}}\" folder.",
```
motiejus, line 391: `locally` is implied, thus delete?

gui/default/assets/lang/lang-en.json
```diff
@@ -374,11 +388,15 @@
...
    "days": "days",
    "directories": "directories",
+   "enter password to encrypt sent data": "enter password to encrypt sent data",
```
motiejus, line 394: it is impossible to encrypt "sent" data. Perhaps "data to be sent", or just "enter password for data encryption"?

lib/config/config.go
```diff
@@ -430,6 +430,22 @@ func (cfg *Configuration) DeviceMap() map[protocol.DeviceID]DeviceConfiguration
 	return m
 }
 
+// FolderPasswords returns the folder passwords set for this device, for
+// folders that have an encryption password set.
+func (cfg Configuration) FolderPasswords(device protocol.DeviceID) map[string]string {
+	res := make(map[string]string)
```
motiejus, line 436: pre-allocate map?

```
 	res := make(map[string]string, len(cfg.Folders))
```

lib/config/folderconfiguration.go
```diff
@@ -253,3 +263,7 @@ func (f *FolderConfiguration) CheckAvailableSpace(req int64) error {
 	}
 	return fmt.Errorf("insufficient space in %v %v", fs.Type(), fs.URI())
 }
+
+func (f *FolderConfiguration) HasPlaintextData() bool {
```
motiejus, line 267: call `IsPlainText`? Because the current signature may imply that the folder may be half-encrypted.

Also, sometimes `FolderTypeReceiveEncrypted`, sometimes `HasPlaintextData()` is used to detect the folder type. Perhaps we don't need this function at all?

lib/scanner/walk.go
```diff
@@ -124,6 +127,16 @@ func (w *walker) walk(ctx context.Context) chan ScanResult {
...
+	if w.DetectChangesOnly {
+		go func() {
+			for file := range toHashChan {
+				finishedChan <- ScanResult{File: file}
+			}
+			close(finishedChan)
+		}()
+		return finishedChan
+	}
+
```
motiejus, line 139: This feels like a hack: a config option which essentially changes what the function is doing (much less than without the config option). Perhaps `walk` could be refactored to a few functions, where the function name, not the config option, is setting what the caller is expecting.

lib/model/model.go
imsodin: `err` has the specifics. It can mean a few things, from not matching PWs, to remote sending us encrypted stuff while we think we should get plain things, or the other way around, or ...  
Maybe just: "Incorrect encryption configuration for folder %v with device %v"?

lib/model/model.go
```diff
@@ -171,9 +174,16 @@ var (
...
-	errStopped              = errors.New("Syncthing is being stopped")
+	errIgnoredFolderRemoved     = errors.New("folder no longer ignored")
+	errReplacingConnection      = errors.New("replacing connection")
+	errStopped                  = errors.New("Syncthing is being stopped")
+	errEncInvConfigLocal        = errors.New("can't encrypt data for a device when the folder type is receiveEncrypted")
+	errEncInvConfigRemote       = errors.New("remote has encrypted data and encrypts that data for us - this is impossible")
+	errEncNotEncryptedLocal     = errors.New("folder is announced as encrypted, but not configured thus")
+	errEncNotEncryptedRemote    = errors.New("folder is configured to be encrypted but not announced thus")
+	errEncNotEncryptedUntrusted = errors.New("device is untrusted, but configured to receive not encrypted data")
+	errEncPW                    = errors.New("different passwords used")
```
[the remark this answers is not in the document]
imsodin: All of this errors are specific to a folder and a remote device (in connection with the local one). That common context is given where all of these errors are logged.

lib/model/model.go
```diff
@@ -2204,6 +2426,12 @@ func (m *model) generateClusterConfig(device protocol.DeviceID) protocol.Cluster
 				Introducer:  deviceCfg.Introducer,
 			}
 
+			if deviceCfg.DeviceID == m.id && hasEncToken {
+				protocolDevice.EncPwToken = encToken
```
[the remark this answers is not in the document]
imsodin: It's a "thing" derived from the password, that allows everyone (whether they know the pw or not, i.e. trusted or not) to check that a folder is only every encrypted with one single pw and emit an error otherwise.

gui/default/assets/lang/lang-en.json
imsodin: "Receive Encrypted" is a folder type. Anyways, clearly a terrible label. It means "Not encrypted/locally changed items in a receive-only folder (where nothing should ever change locally)". So we might as well use the same label as for receive-only folders: "Locally Changed Items".

gui/default/assets/lang/lang-en.json
imsodin: I like being explicit, leaves less space for user misinterpretation. And "change location" is a ubiquitous source of misunderstandings (e.g. versioning only applying to remote changes). And this string is used in a modal, no space issues there.

motiejus (commented):
It would be great to see more tests. To name a few:
- tests with added encrypted folders -- happy case.
- some contents are modified. Tests capture that files are not decrypted and appropriate errors are logged.
- tests on how two peers -- one understanding this feature, the other not -- will behave. E.g. what happens if I share an encrypted folder with a peer that has an older syncthing? The behavior might be simply covered by a simple protocol test, but I haven't looked deeply. I tried two incompatible peers, they kept disconnecting from each other, and I didn't see a user-friendly indication on either side.

lib/protocol/encryption.go
imsodin: I don't follow: You are saying it's possible for an untrusted node to know that two blocks, differing in their encrypted form due to different nonces, have the same payload, correct? How would they know that?

lib/model/folder.go
```diff
@@ -500,16 +507,28 @@ func (f *folder) scanSubdirs(subDirs []string) error {
...
+			// This is a "virtual" parent directory of encrypted files.
+			// We don't track it, but check if anything still exists
+			// within and delete it otherwise.
+			if fi.IsDirectory() && protocol.IsEncryptedParent(fi.Name) {
+				if names, err := mtimefs.DirNames(fi.Name); err == nil && len(names) == 0 {
+					mtimefs.Remove(fi.Name)
+				}
+				changes--
+				return
+			}
```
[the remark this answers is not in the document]
imsodin: The purpose is to have files in many dirs, not just all of them in one dir (the root dir). That's achieved by splitting using the first few filename chars as a parent dir. These dirs are ignored, i.e. they don't get a file info on scanning and we just create them in the puller and remove in scanner if empty. Seems like a functional and simple hack to me.

lib/model/model.go
```diff
@@ -2761,3 +2989,38 @@ func sanitizePath(path string) string {
 
 	return strings.TrimSpace(b.String())
 }
+
+func encTokenPath(cfg config.FolderConfiguration) string {
+	return filepath.Join(cfg.MarkerName, "syncthing-enc_pw_token")
```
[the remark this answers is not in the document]
imsodin: No specific reason, might also live in db. Should it?

lib/model/folder_recvenc.go
```diff
@@ -0,0 +1,111 @@
...
+
+	f.setState(FolderScanning)
+	defer f.setState(FolderIdle)
+
+	batch := newFileInfoBatch(func(fs []protocol.FileInfo) error {
+		f.updateLocalsFromScanning(fs)
+		return nil
+	})
+
+	snap := f.fset.Snapshot()
```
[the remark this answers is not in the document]
calmh: `snap` is pretty well ingrained in our codebase.

lib/protocol/encryption.go
AudriusButkevicius: The encrypted hash is encrypted deterministically for reuse purposes as the comment says. So I assume the encrypted hash would give away that it's the same plaintext data (as otherwise how can you reuse anything?)

lib/model/folder.go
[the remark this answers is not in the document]
AudriusButkevicius: What happens to directories that we announce on the plaintext side?

lib/model/model.go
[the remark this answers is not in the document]
AudriusButkevicius: Why does it need to be stored anywhere anyway? 


As far as I understand it's just a thing that is used to verify the sides agree on a password/verify the password hasn't changed?

lib/model/model.go
AudriusButkevicius: I guess if it can mean many things we should report each fault separately to tell the user where the problem lies. Saying that there is some undefined problem which might be one from this long list is not super helpful.

lib/protocol/encryption.go
calmh: When you request a data block it gets encrypted using a random nonce. You can request a block as many times as you like, getting a new random nonce and hence a new "variant" of the encrypted block. You can of course assume that the plaintext is the same every time you request it, unless the file changed on disk. I don't think this is a problem, this is generally how nonces work? The fact that we have a token of sorts that also says the plaintext didn't change is kind of secondary.

lib/protocol/encryption.go
calmh: My thought here was that we might want to be more restrictive in a clusterconfig towards an untrusted device, perhaps not mentioning other devices at all like we do normally.

lib/protocol/encryption.go
calmh: This originates from when the other side folder was just a normal one getting odd data, at this point it's a separate folder type so we should indeed not even send them.

lib/model/folder.go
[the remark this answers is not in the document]
imsodin: They are treated the same way, i.e. also using first few chars of the encrypted name as parent dirs. The "final" component then is just a dir instead of a file (symlinks are represented as dirs).

lib/protocol/encryption.go
calmh: agreed, there is no need to embed here

cmd/stdecryptfolder/stdecryptfolder.go
```diff
@@ -0,0 +1,154 @@
...
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/binary"
+	"log"
+	"os"
+	"path/filepath"
+
+	"github.com/alecthomas/kingpin"
```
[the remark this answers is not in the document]
calmh: This one is the best though.

cmd/stdecryptfolder/stdecryptfolder.go
```diff
@@ -0,0 +1,154 @@
...
+	var sizeBs [4]byte
+	if _, err := fd.ReadAt(sizeBs[:], info.Size()-4); err != nil {
+		return protocol.FileInfo{}, err
+	}
+
+	size := int(binary.BigEndian.Uint32(sizeBs[:]))
+	fiBs := make([]byte, size)
+	if _, err := fd.ReadAt(fiBs, info.Size()-4-int64(size)); err != nil {
+		return protocol.FileInfo{}, err
+	}
```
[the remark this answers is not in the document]
calmh: Having it at the front means we need to read and understand that offset for every block read when serving requests. Having it at the end means understanding it is optional, which is good as it's only used for reconstructing data without the presence of index info / in a disaster recovery scenario.

lib/model/model.go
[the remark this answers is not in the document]
imsodin: Trusted nodes create it on demand from the passwords. However untrusted nodes can't do that and they need to provide it in cluster configs, thus they need to persist it. That's the entire point of the token, if e.g. a trusted device connects to an untrusted one, the untrusted device needs to be able to present the token it got before when another trusted device connects later.

lib/model/model.go
imsodin: We are not logging an unspecific error or giving a list, but e.g. `Failure checking encryption consistency with device *somedevice* for folder *somefolder*: different encryption passwords used`. I am not saying the error messages are ideal, but I don't get where "report each fault separately" or any "long list" come into this.

cmd/stdecryptfolder/stdecryptfolder.go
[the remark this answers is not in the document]
AudriusButkevicius: Recli is something else, it just generates urfave/cli handlers reflectively.

proto/lib/protocol/bep.proto
calmh: It isn't really because the block size isn't deterministic, it depends on the file size now and the file size in the past, as we may have hashed it once drifted away from the original size.

cmd/stdecryptfolder/stdecryptfolder.go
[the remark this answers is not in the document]
AudriusButkevicius: So from reading the code I understood it was at the front, but you are saying it's at the back? If its at the back, how do we know where it starts?

lib/protocol/encryption.go
imsodin: And the comment is not referring to block reuse between different files, only from the old files. Block hashes are encrypted with the file key, i.e. the filename is used as a nonce and thus the same block in different files have different encrypted hashes.

lib/model/folder_recvenc.go
```diff
@@ -0,0 +1,111 @@
...
+func newReceiveEncryptedFolder(model *model, fset *db.FileSet, ignores *ignore.Matcher, cfg config.FolderConfiguration, ver versioner.Versioner, fs fs.Filesystem, evLogger events.Logger, ioLimiter *byteSemaphore) service {
+	return &receiveEncryptedFolder{newSendReceiveFolder(model, fset, ignores, cfg, ver, fs, evLogger, ioLimiter).(*sendReceiveFolder)}
+}
+
+func (f *receiveEncryptedFolder) Revert() {
+	f.doInSync(func() error { f.revert(); return nil })
+}
+
+func (f *receiveEncryptedFolder) revert() {
+	l.Infof("Reverting unexpected items in folder %v (receive-encrypted)", f.Description)
```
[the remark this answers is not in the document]
imsodin: `%q` is almost never a good idea for user facing strings, if you want to use double quotes use backticks around the string (e.g. a label like "corporate/accounting" shouldn't be displayed as "corporate\/accounting").

cmd/stdecryptfolder/stdecryptfolder.go
```diff
@@ -0,0 +1,154 @@
...
+
+		fi, err := protocol.DecryptFileInfo(encFi, folderKey)
+		if err != nil {
+			log.Printf("Decrypting metadata for %s: %v", path, err)
+			return nil
+		}
+
+		fileKey := protocol.FileKey(fi.Name, folderKey)
+
+		log.Println("Decrypting", fi.Name, "...")
```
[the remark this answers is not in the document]
imsodin: `protocol.FileInfo.Name` already is a full path (well relative to the folder root).

lib/model/model.go
```diff
@@ -1305,8 +1480,14 @@ func (m *model) handleDeintroductions(introducerCfg config.DeviceConfiguration,
...
 // AutoAcceptFolders set to true.
-func (m *model) handleAutoAccepts(deviceCfg config.DeviceConfiguration, folder protocol.Folder) (config.FolderConfiguration, bool) {
+func (m *model) handleAutoAccepts(deviceID protocol.DeviceID, folder protocol.Folder, ccDevicesRemote, ccDevicesLocal map[string]protocol.Device) (config.FolderConfiguration, bool) {
+	ccDevice, hasDevice := ccDevicesRemote[folder.ID]
+	ccDeviceUs, hasDeviceLocal := ccDevicesLocal[folder.ID]
 	if cfg, ok := m.cfg.Folder(folder.ID); !ok {
+		if !hasDevice || !hasDeviceLocal {
+			l.Infof("Failed to auto-accept folder %s from %s due to missing information on encryption from remote device", folder.Description(), deviceID)
+			return config.FolderConfiguration{}, false
+		}
```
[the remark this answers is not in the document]
imsodin: We didn't error on the cc if device info was missing, I just kept that behaviour where possible. When auto-accepting though I want to know whether the remote encrypts the folder to be added to us or not. I could also just go ahead and add it as normal send-receive when that info is missing.

lib/model/model.go
```diff
@@ -1320,25 +1501,44 @@ func (m *model) handleAutoAccepts(deviceCfg config.DeviceConfiguration, folder p
...
 			fcfg := config.NewFolderConfiguration(m.id, folder.ID, folder.Label, fs.FilesystemTypeBasic, filepath.Join(defaultPath, path))
 			fcfg.Devices = append(fcfg.Devices, config.FolderDeviceConfiguration{
-				DeviceID: deviceCfg.DeviceID,
+				DeviceID: deviceID,
 			})
 
-			l.Infof("Auto-accepted %s folder %s at path %s", deviceCfg.DeviceID, folder.Description(), fcfg.Path)
+			if len(ccDevice.EncPwToken) > 0 || len(ccDeviceUs.EncPwToken) > 0 {
+				fcfg.Type = config.FolderTypeReceiveEncrypted
+			}
```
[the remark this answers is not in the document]
imsodin: The token is always on the device that is encrypted. I.e. when the remote generates its cluster config, if it is a receive-encrypted folder (i.e. remote is untrusted), the remote token is not empty, otherwise if it encrypts data for us, the token for our device is not empty. That's also why it's impossible for both the token for us and the remote to be non-empty: That would mean the remote itself has encrypted data/is untrusted, and it actively encrypts data sent to us - not possible.

lib/model/model.go
```diff
@@ -1320,25 +1501,44 @@ func (m *model) handleAutoAccepts(deviceCfg config.DeviceConfiguration, folder p
...
 				return config.FolderConfiguration{}, false
 			}
 		}
+		if !hasDevice || !hasDeviceLocal {
+			l.Infof("Failed to auto-accept folder %s from %s due to missing information on encryption from remote device", folder.Description(), deviceID)
+			return config.FolderConfiguration{}, false
+		}
+		if cfg.Type == config.FolderTypeReceiveEncrypted {
+			if len(ccDevice.EncPwToken) == 0 && len(ccDeviceUs.EncPwToken) == 0 {
+				l.Infof("Failed to auto-accept folder %s from %s as the remote wants to send us un-encrypted data, but we are encrypted", folder.Description(), deviceID)
```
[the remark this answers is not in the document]
imsodin: untrusted is per folder and device, not just device. You can share both encrypted and unencrypted folders with a single device. There is an additional untrusted setting on devices, which if activated ensures that no unencrypted folders are ever shared with that device.

lib/model/model.go
imsodin: The last block can have arbitrary size, but untrusted devices don't know that size, thus just request the standard block size.

lib/protocol/encryption.go
AudriusButkevicius: So I am not sure that's ok from some previous discussions on the subject on DARE/sio with @aead because:

1. If I am able to inject a file into the system somehow, by seeing that file propagate, I can verify if a file with similar content was there before because the deterministic hash will match. I can also verify that two files are the same even if they are at different locations, encrypted with different nonces etc.
2. Because I can ask for the same thing over and over again, and I'll get the same plaintext encrypted with the same key just a different nonce. I am not sure that's ok. Sure, 24 bytes is not easy to bruteforce but still. 

I'd be nice for @aead to provide an opinion.
I know when I discussed this in the past any prospect of block reuse looked unlikely.

proto/lib/protocol/bep.proto
AudriusButkevicius: But we have the size given to us, so surely we can work out what block it's referring to based on it's size?

AudriusButkevicius: I mean imagine a world where syncthing is content addressed, and all requests are served by hash. I think that would still work without the need of block no et al.

lib/model/model.go
imsodin: We did not abort before when those were missing. However it looks like we just didn't check and assumed it would be always there. So it might indeed be sensible to throw an error if any of this is missing, but that's a change from previous behaviour (could do a separate PR to do it there, the less happening in this monstrosity of a PR the better).

cmd/stdecryptfolder/stdecryptfolder.go
```diff
@@ -0,0 +1,154 @@
...
+		if err != nil {
+			log.Printf("Reading %s: %v", path, err)
+			return nil
+		}
+		defer inFd.Close()
+
+		verifyHashes := true
+		if len(fi.Blocks) != len(encFi.Blocks) {
+			log.Printf("Preparing %s: warning: block list mismatch (%d encrypted, %d plaintext), possible metadata out of sync?", outPath, len(encFi.Blocks), len(fi.Blocks))
+			verifyHashes = false
```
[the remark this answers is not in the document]
AudriusButkevicius: The encrypted and decrypted form of the file should have the same number of blocks, so it's either truncated in both or neither.

lib/model/model.go
AudriusButkevicius: Right, ignore me, the specific error is in the message.

lib/model/model.go
imsodin: B requires that either 1. like in the cc from C, C marks itself as encrypted, or 2., like in the cc from A, A is not marked as encrypted, but B is. Also relevant:
https://github.com/syncthing/syncthing/pull/6727#discussion_r484730030

lib/model/model.go
imsodin: Still bad naming: `isEncryptionRemote` means "locally unencrypted data will be encrypted for this remote device" while `isEncryptedLocal` means "the local data is already encrypted". The two are mutually exclusive, because you can't (well shouldn't) encrypt again locally encrypted data.

lib/model/model.go
imsodin: When a receive-encrypted folder is first created, it couldn't have received the password token yet. Only once it receives the first cluster config from a remote, it does receive the token. However it must send that token for a connection to be successfully established, as the remote needs to be able to check that it has the right password.  
Here we just received that first token, thus close the connection to resend the cluster config with that token included.

lib/protocol/encryption.go
imsodin: 1. I still don't get how you can verify anything after injecting a file: The hash won't match with other files that have the same blocks, because:

```
encryptedBlockHash = AES-SIV(blockHash, fileKey)

fileKey = HKDF(SHA256, folderKey+filename, salt: "syncthing", info: nil)
```

2. Related question: Is there any gain in using a nonce/non-deterministic encryption at all then? As you state, the untrusted device anyway has the information to match encrypted data of the same plaintext (by knowing the deterministic encrypted hash). That's also true for encrypted file infos (knowing the deterministic encrypted filename). Even if that were fine in terms of security of the encryption (I don't know about that), as far as I understand there's no point in using a nonce in this scenario. Any in-transit attack scenarios don't apply to us, as transit encryption is covered by another layer. So we might as well use AES-SIV everywhere, which is designed for this scenario (no nonce, deterministic encryption).

imsodin:
Can you (original reviewer/commenter) please mark review comments as resolved if you feel they are. The PR is already huge and it's hard to keep track of what still needs addressing/discussion and anything to make that easier would be appreciated. Thanks!

AudriusButkevicius:
Sorry, I got a busy week ahead of me, so not sure how quickly I'll turn around resolving them.

motiejus:
> Can you (original reviewer/commenter) please mark review comments as resolved if you feel they are. The PR is already huge and it's hard to keep track of what still needs addressing/discussion and anything to make that easier would be appreciated. Thanks!

I believe this applies to me too, but github doesn't allow me to mark my comments as such. I am happy to comment "LGTM" on things I believe are resolved.

imsodin:
That's weird. Anyway, to not generate more comments/notifications just to mark them as done, you could also put a reaction/emoji on those comments which are done :)

motiejus:
> That's weird.

I guess that's only to members/maintainers of syncthing, but I agree everyone who left a review should be able to mark their comments as done. Can this be configurable in project settings?

> Anyway, to not generate more comments/notifications just to mark them as done, you could also put a reaction/emoji on those comments which are done :)

I'll use :rocket: for "done".

lib/protocol/encryption.go
aead: Hi, I don't think I understand the full scope / entire threat model. However, when a stream is split into fragments / blocks and each encrypted block is self-contained then there is a fundamental problem for data integrity:

Assume a file consisting of: `B0 || B1 || B2`. The encrypted file be: `C0 || C1 || C2`
Now, the file gets updated to `B0 || B1' || B2` (`B1 != B1'`). This leads to an encrypted file `C0 || C1' || C2` (`C1 != C1'`)

Now, a receiver has no way to determine what's the latest version of the file -  `C0 || C1 || C2` or `C0 || C1' || C2`
Therefore a malicious node can replace `C1'` with `C1` at any point in time and can re-distribute old content. If a node updates the file whenever `encryptedBlockHash` changes it would accept "outdated" content from a malicious node.
 
Also, the malicious node learns which parts of the file were changed. Depending on the file type this may allow plaintext recovery / usage analysis.

calmh: > Now, a receiver has no way to determine what's the latest version of the file

Why do you think so? The file metadata is encrypted and signed and sent in the usual way. Certainly a malicious device can simply refuse to pass on updated metadata, but it can't create a new metadata entry that refers to old data.

> the malicious node learns which parts of the file were changed

This is indeed the intentional tradeoff. The alternative is that the entire file needs to be send on any change, and that behavior is far enough away from Syncthing that there's really no point any more, I think. At that point you can simply use an encryption program of your choice and SCP instead.

aead: > Why do you think so? The file metadata is encrypted and signed and sent in the usual way. Certainly a malicious device can simply refuse to pass on updated metadata, but it can't create a new metadata entry that refers to old data.

Not sure whether I'm missing something here but what prevents the following flow:
```
 - Node_s encrypts file: C0 || C1 || C2  Enc(B0) || Enc(B1) || Enc(B2) and distributes C0, C1 and C2
 - Node_x receives  C0, C1 and C2 from Node_s
 - Node_r receives C0, C1 and C2 from Node_x and reconstructs B0 || B1 || B2 = Dec(C0) || Dec(C1) || Dec(C2)
 - Node_s updates and encrypts: C1' = Enc(B1') (B1 != B1')  and distributes C1
 - Node_x receives C1'  from Node_s
 - Node_x receives C1' from Node_x and updates  B1 as B1' = Dec(C1')

 - Node_m sends  C1 to Node_x
 - Node_x receives C1 and has no way to determine whether it's a "new" block or an "old" one.
 - Node_r receives C1 and only rejects it if there is some "version downgrade protection" in place that rejects a block
   if its version (or other mono. increasing info) is older then the current info.
``` 
An adversary can replay old information to all peers that cannot apply the downgrade protection due to encryption.

aead: I misunderstood that the blocks are not distributed themselves on an update event. However, if I've understood the description correctly the same problem still exists. Whether it's the actual data or the metadata - at the point when there are multiple valid "versions" of a data structure (`B1'` being the successor of `B1` - or `F'` being the fileinfo successor of `F`) an attacker can replay the old version.
Whether an attacker replays an old ciphertext block or an old ciphertext metadata structure - which the receiver then uses to fetch the ciphertext block - is basically the same. If there is a downgrade-protection in place then a receiver will not accept any plaintext (block) with a lower version number, and therefore, reject the replayed data.

This boils down to: If a node receives multiple encrypted file infos (one for each file update) such that it fetches , decrypts **and** accepts:
 1. `... || B1`
 2. `... || B1'`
 3. `... || B1`
then an attacker can cause re-distribution of old content.

> The question right now is whether it is a problem, that an adversary can request the same blocks again and again, and we'll send him the same plaintext encrypted with chacha and a different nonce every time. And given the adversary knows that it's the same plaintext, is there any point/advantage in using chacha (with nonce, non-deterministic) instead of AES-SIV (deterministic), which we use in other places where the result must be deterministic for the protocol to work.

From a conceptual point of view there should be no difference. AES-GCM-SIV (as nonce-misuse-resistant AEAD) achieves the same security properties as ChaCha20-Poly1305 as long as a nonce is used and should reveal whether two ciphertexts correspond to the same plaintext if no nonce is used.
(AES-GCM-SIV is probabilistic if a nonce / random AAD is specified and deterministic otherwise - See e.g.: https://tools.ietf.org/html/rfc8452#section-1)

calmh: > an attacker can replay the old version

They can, but the version vector identifies it as an old version. It will not cause the recipient to download the old version, it will cause them to consider the sender out of date.

aead: > They can, but the version vector identifies it as an old version. It will not cause the recipient to download the old version, it will cause them to consider the sender out of date.

:+1: 

> Here I'm not sure. :) I started with XChaCha20-Poly1305 because it seems like a modern construction suited for the purpose. AES-SIV came into the picture later for the specific cases where we need the property that a given plaintext corresponds to a certain ciphertext. It's possible we could use the latter for the former, but I saw no reason to switch. Is there a reason to switch?

Well, based on the comments above an adversary can request a encrypted block at any point in time. So, you provide an Oracle to the attacker. Since you use `XChaCha20-Poly1305` that uses 192 bit nonces (128 bit for HChaCha20 KDF and 96 bit for ChaCha20Poly1305) it is practically infeasible for an attacker to request blocks until nonce reuse becomes "likely". The attack would be:
```
1. Request C0 = Enc(B0)
2. Wait for B0 to change: B0' = B0 
3. Request C0' = Enc(B0')
4. Do some basic analysis to find out whether the nonce was reused - e.g. xor C0 and C0'
5. If the nonce was reused => goto 3. 
```
An attacker has to request a block at least `2^64` times to have a non-negligible chance of nonce reuse.
***
However, as an alternative you use AES-GCM-SIV (with or without a nonce). The advantage of AES-GCM-SIV is that
even though the nonce is 96 bits long and an attacker probably can trigger nonce reuse AES-GCM-SIV would only reveal
whether the block is equal to another block in the past in such a case.
However, the general considerations w.r.t. `(X)ChaCha20-Poly1305 vs AES-GCM` still applies - e.g. AES usually requires special hardware instructions for a secure & fast implementation.

So, I don't see a strong reason to use AES-GCM-SIV over XChaCha20-Poly1305 here. However, it is important that the nonce of non-deterministic scheme that does not provide misuse-resistance has to be sufficiently large. For instance, I think XChaCha20-Poly1305 would be fine while ChaCha20-Poly1305 (with 96 bit nonces) would not...

gui/default/assets/lang/lang-en.json
```diff
@@ -281,6 +289,7 @@
    "Start Browser": "Start Browser",
    "Statistics": "Statistics",
    "Stopped": "Stopped",
+   "Stores and syncs only encrypted data. Folders on all connected devices need to be set up with the same password or be of type \"{%receiveEncrypted%}\" too.": "Stores and syncs only encrypted data. Folders on all connected devices need to be set up with the same password or be of type \"{{receiveEncrypted}}\" too.",
```
AudriusButkevicius, line 292: The `stores` part is a false statement. My local data is not encrypted.

I guess the right term is encrypts outgoing, decrypts incoming data

gui/default/assets/lang/lang-en.json
```diff
@@ -335,10 +345,13 @@
    "Unavailable": "Unavailable",
    "Unavailable/Disabled by administrator or maintainer": "Unavailable/Disabled by administrator or maintainer",
    "Undecided (will prompt)": "Undecided (will prompt)",
+   "Unencrypted Items": "Unencrypted Items",
```
AudriusButkevicius, line 348: Again, we don't know if they are encrypted or not, might just be the wrong password.

gui/default/assets/lang/lang-en.json
```diff
@@ -374,11 +387,15 @@
    "You have no ignored folders.": "You have no ignored folders.",
    "You have unsaved changes. Do you really want to discard them?": "You have unsaved changes. Do you really want to discard them?",
    "You must keep at least one version.": "You must keep at least one version.",
+   "You should never add or change anything locally in a \"{%receiveEncrypted%}\" folder.": "You should never add or change anything locally in a \"{{receiveEncrypted}}\" folder.",
    "days": "days",
    "directories": "directories",
+   "enter password to encrypt data to be sent": "enter password to encrypt data to be sent",
```
AudriusButkevicius, line 393: Strange that this starts with a lowercase.
Also, this password is used for decryption as well, so "encrypt data to be sent" is just part of the story?

"Enter encryption password" ?

gui/default/assets/lang/lang-en.json
```diff
@@ -374,11 +387,15 @@
...
    "files": "files",
    "full documentation": "full documentation",
    "items": "items",
+   "not shared": "not shared",
+   "received data is already encrypted": "received data is already encrypted",
```
AudriusButkevicius, line 398: again, lowercase for both of these?

gui/default/index.html
```diff
@@ -852,6 +864,7 @@ <h4 class="panel-title">
   <ng-include src="'syncthing/core/aboutModalView.html'"></ng-include>
   <ng-include src="'syncthing/core/discoveryFailuresModalView.html'"></ng-include>
   <ng-include src="'syncthing/folder/removeFolderDialogView.html'"></ng-include>
+  <ng-include src="'syncthing/folder/deleteEncFolderDialogView.html'"></ng-include>
```
AudriusButkevicius, line 867: Enc->Encrypted maybe?

gui/default/syncthing/core/editShareTemplate.html
```diff
@@ -0,0 +1,35 @@
...
+          <input class="form-control input-sm" type="{{plain ? 'text' : 'password'}}" ng-model="encpws[id]" ng-required="untrusted" placeholder="{{'enter password to encrypt data to be sent' | translate}}" />
+        </span>
+        <span ng-switch-default>
+          <input class="form-control input-sm" type="password" placeholder="{{'not shared' | translate}}" disabled />
+        </span>
+      </span>
+    </span>
+    <span ng-switch="selected[id] && folderType !== 'receiveencrypted'" class="input-group-addon">
+      <span ng-switch-when='true'>
+        <span class="button fas fa-fw fa-eye" ng-click="togglePW()" />
```
AudriusButkevicius, line 28: PW

gui/default/syncthing/core/editShareTemplate.html
```diff
@@ -0,0 +1,35 @@
+<div class="col-md-6 checkbox">
+  <label for="sharedwith-{{id}}">
+    <input id="sharedwith-{{id}}" ng-model="selected[id]" type="checkbox" />
+    <span tooltip data-original-title="{{id}}">{{label}}</span>
+  </label>
+</div>
+<div class="col-md-6">
+  <div class="input-group">
+    <span class="input-group-addon" ng-switch="folderType !== 'receiveencrypted' && !encpws[id]">
```
AudriusButkevicius, line 9: encpws

gui/default/syncthing/core/syncthingController.js
```diff
@@ -831,7 +841,7 @@ angular.module('syncthing.core')
             if (status === 'unknown') {
                 return 'info';
             }
-            if (status === 'stopped' || status === 'outofsync' || status === 'error' || status === 'faileditems') {
+            if (status === 'stopped' || status === 'outofsync' || status === 'error' || status === 'faileditems' || status === 'localunencrypted') {
```
AudriusButkevicius, line 844: Again, it's not really localunencrypted, it's more like unexpected, which could be localadditions?

If we treat this (localadditions) as success for receiveonly, why are we treating this as a error for receiveencrypted?

gui/default/syncthing/core/syncthingController.js
```diff
@@ -2529,4 +2586,24 @@ angular.module('syncthing.core')
...
                     address.indexOf('unixs://') == 0);
         }
+
+    })
+    .directive('shareTemplate', function () {
+        return {
+            templateUrl: 'syncthing/core/editShareTemplate.html',
+            scope: {
+                selected: '=',
+                encpws: '=',
```
AudriusButkevicius, line 2596: encpws

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1847,25 +1868,25 @@ angular.module('syncthing.core')
...
                 };
             });
-            for (var deviceID in folderCfg.selectedDevices) {
-                if (folderCfg.selectedDevices[deviceID] === true) {
+            for (var deviceID in $scope.currentSharing.selected) {
+                if ($scope.currentSharing.selected[deviceID] === true) {
                     newDevices.push({
-                        deviceID: deviceID
+                        deviceID: deviceID,
+                        encryptionPassword: $scope.currentSharing.encryptionPasswords[deviceID],
```
AudriusButkevicius, line 1884: is the trailing comma actually fine in js?

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -63,27 +63,27 @@
...
+          <div class="form-group">
+            <div class="form-horizontal" ng-if="currentSharing.shared.length">
+              <label translate for="folders">Shared Folders</label>
+              <p class="help-block">
+                <span translate>Select the folders to share with this device.</span>&emsp;
+                <small><a href="#" ng-click="selectAllSharedFolders(true)" translate>Select All</a>&emsp;
+                  <a href="#" ng-click="selectAllSharedFolders(false)" translate>Deselect All</a></small>
+              </p>
+              <div class="form-group" ng-repeat="folder in currentSharing.shared">
+                <share-template selected="currentSharing.selected[folder.id]" encpws="currentSharing.encryptionPasswords[folder.id]" id="{{folder.id}}" label="{{folder.label}}" folder-type="{{folder.type}}" untrusted="{{device.untrusted}}" />
```
AudriusButkevicius, line 75: encpws

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -63,27 +63,27 @@
...
+            </div>
+            <div class="form-horizontal" ng-if="currentSharing.unrelated.length">
+              <label translate>Unshared Folders</label>
+              <p class="help-block">
+                <span translate>Select additional folders to share with this device.</span>&emsp;
+                <small><a href="#" ng-click="selectAllUnrelatedFolders(true)" translate>Select All</a>&emsp;
+                  <a href="#" ng-click="selectAllUnrelatedFolders(false)" translate>Deselect All</a></small>
+              </p>
+              <div class="form-group" ng-repeat="folder in currentSharing.unrelated">
+                <share-template selected="currentSharing.selected[folder.id]" encpws="currentSharing.encryptionPasswords[folder.id]" id="{{folder.id}}" label="{{folder.label}}" folder-type="{{folder.type}}" untrusted="{{device.untrusted}}" />
```
AudriusButkevicius, line 86: encpws

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -46,24 +46,18 @@
...
-            <div class="row">
-              <div class="col-md-4" ng-repeat="device in currentFolder.sharedDevices">
-                <div class="checkbox">
-                  <label>
-                    <input type="checkbox" ng-model="currentFolder.selectedDevices[device.deviceID]" /> {{deviceName(device)}}
-                  </label>
-                </div>
-              </div>
+            <div class="form-group" ng-repeat="device in currentSharing.shared">
+              <share-template selected="currentSharing.selected" encpws="currentSharing.encryptionPasswords" id="{{device.deviceID}}" label="{{deviceName(device)}}" folder-type="{{currentFolder.type}}" untrusted="{{device.untrusted}}" />
```
AudriusButkevicius, line 57: encpws

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -73,14 +67,8 @@
...
-            <div class="row">
-              <div class="col-md-4" ng-repeat="device in currentFolder.unrelatedDevices">
-                <div class="checkbox">
-                  <label>
-                    <input type="checkbox" ng-model="currentFolder.selectedDevices[device.deviceID]" /> {{deviceName(device)}}
-                  </label>
-                </div>
-              </div>
+            <div class="form-group" ng-repeat="device in currentSharing.unrelated" ng-init="id = device.deviceID; folder = currentFolder">
+              <share-template selected="currentSharing.selected" encpws="currentSharing.encryptionPasswords" id="{{device.deviceID}}" label="{{deviceName(device)}}" folder-type="{{currentFolder.type}}" untrusted="{{device.untrusted}}" />
```
AudriusButkevicius, line 71: encpws

lib/model/folder_sendrecv.go
```diff
@@ -1548,6 +1587,34 @@ func (f *sendReceiveFolder) performFinish(file, curFile protocol.FileInfo, hasCu
...
+	size := file.ProtoSize()
+	bs := make([]byte, 4+size)
+	n, err := file.MarshalTo(bs)
+	if err != nil {
+		return protocol.FileInfo{}, err
+	}
+	binary.BigEndian.PutUint32(bs[n:], uint32(n))
+	bs = bs[:n+4]
+
+	fd, err := f.fs.OpenFile(tempName, fs.OptWriteOnly|fs.OptAppend, 0600)
```
AudriusButkevicius, line 1604: Should we defer closing the file in shared puller state or something, so we wouldn't need to reopen it?

lib/model/model.go
AudriusButkevicius: Why don't we just pad the last block?

lib/model/model.go
```diff
@@ -1317,25 +1498,44 @@ func (m *model) handleAutoAccepts(deviceCfg config.DeviceConfiguration, folder p
...
+			return config.FolderConfiguration{}, false
+		}
+		if cfg.Type == config.FolderTypeReceiveEncrypted {
+			if len(ccDeviceRemote.EncryptionPasswordToken) == 0 && len(ccDeviceLocal.EncryptionPasswordToken) == 0 {
+				l.Infof("Failed to auto-accept folder %s from %s as the remote wants to send us un-encrypted data, but we are encrypted", folder.Description(), deviceID)
+				return config.FolderConfiguration{}, false
+			}
+		} else {
+			if len(ccDeviceRemote.EncryptionPasswordToken) > 0 || len(ccDeviceLocal.EncryptionPasswordToken) > 0 {
+				l.Infof("Failed to auto-accept folder %s from %s as the remote wants to send us encrypted data, but we are not encrypted", folder.Description(), deviceID)
```
AudriusButkevicius, line 1531: What does "we are not encrypted" mean?

lib/model/model.go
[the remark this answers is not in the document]
AudriusButkevicius: Why can't I have:

`<untrusted> <-> <untrusted>` but both know/use the password, effectively adding an extra encryption layer on top of the transport?

In that case you'd have tokens from both sides, and that would be fine, as long as they match.

lib/model/model.go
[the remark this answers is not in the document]
AudriusButkevicius: I think this is fine.

lib/model/model.go
AudriusButkevicius: So I need to think more about this, but something smells to me here.

lib/model/model.go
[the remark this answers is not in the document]
AudriusButkevicius: I guess this sort of reduces the usefulness of per-device-per-folder passwords, as for two plaintext devices to sync to a third encrypted device, they would have to use the same password.

Sure for a 4th device they could use a different one, which I guess is why we have this, but I wonder if it's worth it.

lib/model/model.go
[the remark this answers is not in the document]
AudriusButkevicius: Token implies it's used somehow, you exchange a token for something else.
This is more of a fingerprint.

lib/model/model.go
```diff
@@ -1054,9 +1150,40 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
...
+		ccDeviceLocal, hasCCDeviceLocal := ccDevicesLocal[folder.ID]
+
+		if err := m.ccCheckEncryption(cfg, folderDevice, ccDeviceRemote, ccDeviceLocal, hasCCDeviceRemote, hasCCDeviceLocal, deviceCfg.Untrusted); err != nil {
+			sameError := false
+			if devs, ok := m.folderEncryptionFailures[folder.ID]; ok {
+				sameError = devs[deviceID] == err
+			} else {
+				m.folderEncryptionFailures[folder.ID] = make(map[protocol.DeviceID]error)
+			}
+			m.folderEncryptionFailures[folder.ID][deviceID] = err
```
AudriusButkevicius, line 1167: Can this live in the folder rather than the model somehow?

AudriusButkevicius (commented):
More naming moaning. The most important stuff. Resolved stuff that was not relevant, and things that are still relevant and became outdated and hidden, recommented on them. There are still a few old outdated comment threads that are still relevant and not resolved.

gui/default/syncthing/core/syncthingController.js
maxhq: Yes they are (ECMAScript 5). And they are useful too because they prevent errors when adding more entries.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

gui/default/assets/lang/lang-en.json
imsodin: This describes the receive-only folder type. It does receive encrypted data, not decrypt, and stores encrypted data.

gui/default/syncthing/core/syncthingController.js
imsodin: I didn't notice local additions was treated like a success - I don't see how that makes sense and if anything, that should be changed.

lib/model/model.go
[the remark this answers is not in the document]
imsodin: > I guess this sort of reduces the usefulness of per-device-per-folder passwords, as for two plaintext devices to sync to a third encrypted device, they would have to use the same password.

That's not just because of password tokens though, that's intrinsic in the encryption protocol: If two trusted devices did use different passwords for a single untrusted device, they wouldn't sync with each other at all: Due to different keys, they'd generate different encrypted filenames for the same file. The untrusted file would then have two files for every "unencrypted" file and when one trusted device receives such a duplicate file from the other trusted device through the untrusted one, it won't be able to decrypt it.

Whether the possibility to have separate passwords for the same folder for different devices is useful is another matter. I think that was discussed on the forum thread and finally decided to have it.

lib/model/model.go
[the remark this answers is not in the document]
imsodin: I assume you mean the situation as in the first graphic here: https://docs.syncthing.net/branch/untrusted/html/users/untrusted.html  
That is possible, however both won't send a token for themselves, only for the remote. Because they themselves do not have encrypted data, they do encrypt outgoing data. I.e. for this scenario in this condition `ccDevice` (now `ccDeviceRemote`) will be empty.

lib/scanner/walk.go
imsodin: I introduced `WalkWithoutHashing` in https://github.com/syncthing/syncthing/pull/6727/commits/faa39d179385df44f03e193cb6b90e764ae6d414  
There wasn't much refactoring necessary, as code for walking the filesystem and detecting changes was already well separated from hashing files. The previous solution with the config option needed a single condition with 7 lines of code in it to trigger the different behaviour (those 7 lines now live in `walkWithoutHashing`).

lib/model/model.go
imsodin: These are connection/cluster config level errors. And folders do not handle that let allow know anything about it. Changing that would be entirely out of scope of this PR.

lib/model/model.go
imsodin: The sent data is already padded, because `res.data` is created with the requested size. Thus reading from file tries to read beyond the end of the file.

calmh:
#7055

PR #6735, opened 2020-06-10
lib: Walk fs and db simultaneously on scan

imsodin:
### Purpose

This is a respin of https://github.com/syncthing/syncthing/pull/4584, to address the [scanner part of the latest case insensitivity proposal](https://github.com/syncthing/syncthing/wiki/Filesystem-Case-Sensitivity#scanner).  I had to adjust a bit on stuff that changed since the original PR (around receive only and vector nulling on unignore), but it's mostly the same otherwise.

PR is marked as WIP, as it should probably be part of one big change regarding case sensitivity and it needs another round of debugging apparently. 

### Testing

One new unit tests from the old PR (haven't looked at it at all), other than that nothing.

### DIFF
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -187,6 +187,13 @@ func (s *Snapshot) WithPrefixedHaveTruncated(device protocol.DeviceID, prefix st
 	}
 }
 
+func (s *Snapshot) WithPrefixedHave(device protocol.DeviceID, prefix string, fn Iterator) {
+	l.Debugf("%s WithPrefixedHave(%v)", s.folder, device)
+	if err := s.t.withHave([]byte(s.folder), device[:], []byte(osutil.NormalizedFilename(prefix)), false, nativeFileIterator(fn)); err != nil && !backend.IsClosed(err) {
+		panic(err)
+	}
+}
+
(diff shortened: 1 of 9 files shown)


imsodin:
Actually this old method of detecting deletions at the same time as changes has a conceptual flaw, that is by now highlighted by the rename detection: Now index updates for deletions can be sent before changes/creations. Thus experiment closed - below is some more detail and the new approach compatible with rename detection.

To solve this issue with the proposed single walk to detect both changes and deletions we'd have to delay (cache) all detected deletions. As there's no limitation on the possible amount of deletions, that would require a db backend in case cached deletions exceed a certain size. That's doable, but in my opinion questionable if the gain of a single pass outweighs the loss in terms of additional db interactions.  
Of course we could also forgo rename efficiency and flush deletions if they exceed a certain size, but that feels like a way bigger drawback than doing two passes over the filesystem.  
My current opinion is that this single pass walking is the wrong approach. Instead I'll change the current second pass to be driven by walking the filesystem as well, instead of walking the db and stating stuff. That's still more efficient than the current approach and handles case correctly without additional changes. Only for the rename detection to work with case-insensitivity we'll need the "super-cow-stat".

PR #6737, opened 2020-06-11
lib: Walk fs and db simultaneously on scan

imsodin:
In the end I still want to do https://github.com/syncthing/syncthing/pull/6735, just a bit differently (shouldn't reason about and close PRs late at night xD ):   

This still walks the filesystem and db at the same time, however still does two passes. In the first detecting changes to items existing in the filesystem, and in the second detecting deletions (and deleted files being ignored). The difference is that due to walking the fs instead of stating this does correctly detect case renames on case-insensitive systems. And it's probably a bit more efficient as well, as it's now a single listdir call per directory instead of stating every single file and instead of individual db lookups for every item in the fs, it walks over keys in db.

The rename detection in the folder still relies on stating to detect deletions, i.e. will not work in case of case conflicts on case-insensitive systems. However that's just an optimisation that fails and will be solved once the "super-cow-stat" is in place.

The first commit is https://github.com/syncthing/syncthing/pull/6736, the latter two are just historical, no need to look at them separately.

### DIFF
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -187,6 +187,13 @@ func (s *Snapshot) WithPrefixedHaveTruncated(device protocol.DeviceID, prefix st
 	}
 }
 
+func (s *Snapshot) WithPrefixedHave(device protocol.DeviceID, prefix string, fn Iterator) {
+	l.Debugf("%s WithPrefixedHave(%v)", s.folder, device)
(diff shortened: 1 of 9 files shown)


imsodin:
I integrated this into a single PR for the case-insensitivity: https://github.com/syncthing/syncthing/pull/6717

PR #6741, opened 2020-06-15
lib/model: Check dir before deletion when pulling

imsodin:
### Purpose

The main fix in this PR is that an unscanned directory was deleted when pulling a deletion for said path. That likely didn't cause much real-world problems, as it would later fail if there's any content in the directory. Nevertheless it's wrong.

Another discovered issue: `scanIfItemChanged` scheduled a scan for an empty string whenever there wasn't anything in the db yet, but something on disk. That's because if the the db fileinfo `cur` is empty, `cur.Name` is empty too naturally. Again, that "just" meant needless scanning, the outcome was still correct.

### Testing

One new unit test for deleting an unscanned directory (which also fails on the empty string to be scanned).

One existing unit test failed due to that change, thus needed a scan before the scenario to make sure the scenario in said test is represented in the db.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -75,8 +75,30 @@ var (
 	contextRemovingOldItem      = "removing item to be replaced"
 )
 
+type dbUpdateType int
+
+func (d dbUpdateType) String() string {
+	switch d {
+	case dbUpdateHandleDir:
+		return "dbUpdateHandleDir"
+	case dbUpdateDeleteDir:
+		return "dbUpdateDeleteDir"
+	case dbUpdateHandleFile:
+		return "dbUpdateHandleFile"
+	case dbUpdateDeleteFile:
+		return "dbUpdateDeleteFile"
+	case dbUpdateShortcutFile:
+		return "dbUpdateShourtcutFile"
+	case dbUpdateHandleSymlink:
+		return "dbUpdateHandleSymlink"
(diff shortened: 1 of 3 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -1904,10 +1927,10 @@ func (f *sendReceiveFolder) deleteDirOnDisk(dir string, snap *db.Snapshot, scanC
 // scanIfItemChanged schedules the given file for scanning and returns errModified
 // if it differs from the information in the database. Returns nil if the file has
 // not changed.
-func (f *sendReceiveFolder) scanIfItemChanged(stat fs.FileInfo, item protocol.FileInfo, hasItem bool, scanChan chan<- string) (err error) {
+func (f *sendReceiveFolder) scanIfItemChanged(name string, stat fs.FileInfo, item protocol.FileInfo, hasItem bool, scanChan chan<- string) (err error) {
 	defer func() {
 		if err == errModified {
-			scanChan <- item.Name
+			scanChan <- name
```
calmh, line 1933: could have used stat.Name() instead of a new parameter? no big deal

imsodin: Not quite, because `stat.Name() == filepath.Base(name)`. Probably we should be consistent with using `name` for the last component and `path` otherwise - but that would be a pretty big cleanup job, because we aren't consistent at all on that o.O

PR #6745, opened 2020-06-15
[enc] replace IsSHA256 with folder type check

imsodin:
PR for encryption branch

This is a pure refactor to disable block finding/comparing based on folder type, thus eliminating any need for `HashIsSHA256`. To prevent `copierRoutine` from becoming even more nested, I moved the weak hash initialization into `initWeakHashFinder`.


lib/model/folder_sendrecv.go
```diff
@@ -1273,9 +1244,9 @@ func (f *sendReceiveFolder) copierRoutine(in <-chan copyBlocksState, pullChan ch
 			buf = protocol.BufferPool.Upgrade(buf, int(block.Size))
 
 			var found bool
-			if block.HashIsSHA256() {
+			if f.Type != config.FolderTypeReceiveEncrypted {
```
calmh, line 1247: Maybe for clarity hide this behind a `func (t FolderType) HasPlaintextData() bool` or so?

imsodin: I opted to put it onto the folder config, not type. Usage in the model is a bit more succinct and in case something changes in how we represent "being locally encrypted" in the process that won't affect model code like this.

Build status can be ignored, probably need to merge main into encryption again - will do.

calmh:
doesn't merge with main any more, but that needs fixing on the encryption branch, so merging here

PR #6746, opened 2020-06-15
all: Add copy-on-write filesystem support (fixes #4271)

### LINKED ISSUE #4271: Make use of filesystems that can handle copies efficiently.

AudriusButkevicius:
I can't merge into my own branch and have a PR here, so opening the full thing.
Only the last commit is interesting here.


AudriusButkevicius:
So I can't open a PR to the PR because the branch lives in my repo. Effectively, only the last commit matters.

lib/versioner/simple.go
```diff
@@ -19,9 +19,10 @@ func init() {
 }
 
 type simple struct {
-	keep       int
-	folderFs   fs.Filesystem
-	versionsFs fs.Filesystem
+	keep            int
+	folderFs        fs.Filesystem
+	versionsFs      fs.Filesystem
+	copyRangeMethod fs.CopyRangeMethod
```
calmh, line 25: This seems like it could just obey the folder copyrangemethod?

lib/fs/basicfs_copy_range_method.go
```diff
@@ -54,3 +54,7 @@ func (o *CopyRangeMethod) UnmarshalText(bs []byte) error {
 	}
 	return nil
 }
+
+func (o *CopyRangeMethod) ParseDefault(str string) error {
```
calmh, line 58: Unused?

lib/versioner/simple.go
AudriusButkevicius: I did that at first, but it was ugly pushing it down singular values from the folder. The alternative could be to push down whole folder config into the versioner factory method?

lib/fs/basicfs_copy_range_method.go
AudriusButkevicius: It's used by SetDefaults

lib/versioner/simple.go
calmh: Yeah, I guess try that. It would be odd to enable fancy copyranges on a folder and not gain the advantage on versioning automatically.

lib/fs/basicfs_copy_range.go
```diff
@@ -0,0 +1,60 @@
...
+func registerCopyRangeImplementation(copyMethod CopyRangeMethod, impl copyRangeImplementation) {
+	mut.Lock()
+	defer mut.Unlock()
+
+	l.Debugln("Registering " + copyMethod.String() + " copyRange method")
+
+	copyRangeMethods[copyMethod] = impl
+}
+
+// CopyRange tries copy the datto use the most optimal way to copy data between two files.
```
calmh, line 32: The datto, eh?

lib/fs/basicfs_copy_range.go
```diff
@@ -0,0 +1,60 @@
...
+// Takes size bytes at offset srcOffset from the source file, and copies the data to destination file at offset
+// dstOffset. If required, adjusts the size of the destination file to fit that much data.
+//
+// On Linux/BSD it tries to use ioctl and copy_file_range system calls, which if the underlying filesystem supports it
+// tries referencing existing data in the source file, instead of making a copy and taking up additional space.
+//
+// If that is not possible, the data will be copied using an in-kernel copy (copy_file_range fallback, sendfile),
+// oppose to user space copy, if those system calls are available and supported for the source and target in question.
+//
+// CopyRange does it's best to have no effect on src and dst file offsets (copy operation should not affect it).
```
calmh, line 42: `its`

lib/fs/basicfs_copy_range.go
```diff
@@ -0,0 +1,60 @@
...
+func CopyRange(copyMethod CopyRangeMethod, src, dst File, srcOffset, dstOffset, size int64) error {
+	if impl, ok := copyRangeMethods[copyMethod]; !ok {
```
calmh, line 44: I'd invert and have the ENOTSUP outside the if but 🤷

lib/fs/basicfs_copy_range_copyfilerange.go
```diff
@@ -0,0 +1,42 @@
...
+		//
+		// Also, even if explicitly not stated, the same is true for dstOffset
+		n, err := unix.CopyFileRange(int(src.Fd()), &srcOffset, int(dst.Fd()), &dstOffset, int(size), 0)
+		if n == 0 && err == nil {
+			return io.ErrUnexpectedEOF
+		}
+		if err != nil && err != syscall.EAGAIN {
+			return err
+		}
+		size -= int64(n)
```
calmh, line 39: if we got syscall.EAGAIN above we will happily reduce size and continue?

lib/fs/basicfs_copy_range_ioctl.go
```diff
@@ -0,0 +1,72 @@
...
+	// https://www.man7.org/linux/man-pages/man2/ioctl_ficlonerange.2.html
+	// If src_length is zero, the ioctl reflinks to the end of the source file.
+	if srcOffset+size == fi.Size() {
+		size = 0
+	}
+
+	if srcOffset == 0 && dstOffset == 0 && size == 0 {
+		// Optimization for whole file copies.
+		_, _, e1 := syscall.Syscall(syscall.SYS_IOCTL, dst.Fd(), FICLONE, src.Fd())
+		if e1 != 0 {
```
calmh, line 55: The double-take on that one... 😮 maybe call it `errNo` and (return `error(errNo)`?) just to be super duper explicit about wtf is going on here?

lib/fs/basicfs_copy_range_ioctl.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	params := fileCloneRange{
+		srcFd:     int64(src.Fd()),
+		srcOffset: uint64(srcOffset),
+		srcLength: uint64(size),
+		dstOffset: uint64(dstOffset),
+	}
+	_, _, e1 := syscall.Syscall(syscall.SYS_IOCTL, dst.Fd(), FICLONERANGE, uintptr(unsafe.Pointer(&params)))
+	if e1 != 0 {
```
calmh, line 68: likewise

lib/fs/basicfs_copy_range_sendfile.go
```diff
@@ -0,0 +1,67 @@
...
+		// in_fd.
+		n, err := syscall.Sendfile(int(dst.Fd()), int(src.Fd()), &srcOffset, int(size))
+		if n == 0 && err == nil {
+			err = io.ErrUnexpectedEOF
+		}
+		if err != nil && err != syscall.EAGAIN {
+			return err
+		}
+
+		size -= int64(n)
```
calmh, line 62: EAGAIN logic oddness

lib/fs/basicfs_copy_range_copyfilerange.go
AudriusButkevicius: EAGAIN usually means interrupted, which could copy some data... We're deducting n, so we'll know how much, if anything, was copied.

calmh: Ah, cool.

AudriusButkevicius: Looking at:
https://go-review.googlesource.com/c/go/+/229101/7/src/internal/poll/copy_file_range_linux.go

It seems it does not need any special handling.
Also, it seems io.Copy now does this behind the scenes.

AudriusButkevicius: Apparently there is no such thing as non-blocking file io, so there is no way that EAGAIN could be returned.

calmh: Well that was the theory about Stat() as well and that doesn’t match up with reality so I think handling this is good; unless failure results in fallback to userspace copies.

AudriusButkevicius: So it's not clear what's the value for N when EAGAIN returns.  
The docs state if n is -1, check errno, implying that if err == EAGAIN, then n = -1, so I am going for the safe option and not handling it.

I guess I could do
```
if err != nil && err != EAGAIN {
}
if n > 0 {
  size -= n
}
```

AudriusButkevicius: Updated to deduct n only if it's positive.

lib/fs/basicfs_copy_range.go
```diff
@@ -0,0 +1,57 @@
...
+
+package fs
+
+import (
+	"syscall"
+
+	"github.com/syncthing/syncthing/lib/sync"
+)
+
+var (
```
imsodin, line 15: Everything in this file except for the last method in principle isn't basicfs specific, i.e. should live in filesystem.go or filesystem_copy_range.go. Same for the methods listing file and the standard implementation.

lib/fs/basicfs_copy_range_allwithfallback.go
```diff
@@ -0,0 +1,21 @@
...
+
+package fs
+
+func init() {
+	registerCopyRangeImplementation(CopyRangeMethodAllWithFallback, copyRangeAllWithFallback)
+}
+
+func copyRangeAllWithFallback(src, dst File, srcOffset, dstOffset, size int64) error {
+	var err error
+	for _, method := range []CopyRangeMethod{CopyRangeMethodIoctl, CopyRangeMethodCopyFileRange, CopyRangeMethodSendFile, CopyRangeMethodStandard} {
```
imsodin, line 15: Run everything in `copyRangeMethods` except for `CopyRangeMethodAllWithFallback`.

And isn't basicfs specific either.

AudriusButkevicius: I want this in a specific order tho.

AudriusButkevicius:
We gud?

calmh:
Write a docs page on roughly what the different configs do?

AudriusButkevicius:
Also, when we add this in the change log, we should mention that this is an option you need to enable.

Atera:
I volunteered to test in the other PR. Would love to give this a spin. If the documentation's not done yet, could you briefly state what I need to tweak to enable the changes? I add a 'copyRangeMethod' to the folder options and set it to 'copy_file_range'? I seem to see options standard, ioctl, copy_file_range, sendfile, and all, but I'm not sure I'm looking in the right place or what these behaviours are.

AudriusButkevicius:
If you start a master build you'll see new option in folder config after the first save, change that option to one of the mechanisms as appropriate.

calmh:
Any docs on this available? I would like to point to a page with some guidance so that we can get some testing early in the 1.8.0 RC cycle as I expect we might need iterations on this...

AudriusButkevicius:
Yes, it's in the folder section.

calmh:
Right... Mind if I copy/move it to a separate advanced section, to make it easier to link & refer to?

AudriusButkevicius:
Sure, whatevs.

mksully22:
This pull request (commit 4812fd3ec13b627354f416073fa1ffca26b1187a) fails compile on ppc64le (compiling on Alpine).
A git bisect was used to narrow down the commit that was responsible. The Error reported:
```
--- FAIL: TestCopyRange (2.73s)
    --- FAIL: TestCopyRange/tmp (2.73s)
        --- FAIL: TestCopyRange/tmp/ioctl (0.34s)
            --- FAIL: TestCopyRange/tmp/ioctl/append_to_end (0.02s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/append_to_end_offsets_at_start (0.03s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/overwrite_part_of_destination_region (0.02s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/overwrite_all_of_destination (0.03s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/overwrite_part_of_destination (0.02s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/destination_gets_expanded_as_it_is_being_written_to (0.02s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/unaligned_source_offset_unaligned_size (0.01s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/unaligned_source_offset_aligned_size (0.03s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/unaligned_destination_offset_unaligned_size (0.03s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/unaligned_destination_offset_aligned_size (0.03s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/aligned_offsets_unaligned_size (0.03s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/last_block (0.01s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/whole_file_copy_block_aligned (0.01s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
            --- FAIL: TestCopyRange/tmp/ioctl/whole_file_copy_not_block_aligned (0.01s)
                filesystem_copy_range_test.go:332: inappropriate ioctl for device
FAIL
FAIL    github.com/syncthing/syncthing/lib/fs   3.214s

```

AudriusButkevicius:
There was a follow up PR in 1.8.0 that probably fixes this.

mksully22:
Thank you Audrius. I just re-attempted to build the latest (main branch) with git log showing the latest commit being:
'''
commit fd8bea6179efe2931cb90398dda754edc5bdb622 (HEAD -> main, origin/main, origin/HEAD)
Author: Simon Frei <freisim93@gmail.com>
Date:   Mon Aug 10 18:48:37 2020 +0200

    lib/osutil: Preserve perms in AtomicWriter (fixes #tbd) (#6885)

'''

It also fails in the same way at this level. Is the PR you are thinking of uncommitted? If so can you give me a link to it? I'd be happy to give it a try.

AudriusButkevicius:
For people finding this ticket in the future.
This is now implemented as an optional configuration, but the default behaviour remains the same, copy data into application memory.

You can find the docs for the feature here: https://docs.syncthing.net/advanced/folder-copyrangemethod.html

Issue #6748, opened 2020-06-16
Folder defaults are not applied consistently on config migrations / new folders


calmh:
We set some defaults in the FolderConfiguration struct definition. These are effectively only used for the initial "Sync" folder, because we only call SetDefaults in NewConfig and not when deserializing. The defaults for new folders are set in syncthingController.js. Defaults for new config options are never applied to existing folders.

Suggested cleanup:

- Some hack to make sure we apply defaults on unmarshalling a config from XML/JSON.
- A new REST API call `.../folder-defaults` that returns the default values to use to the GUI. Currently this can be just the compiled in defaults, but in the future one could imagine this being the result of compiled in defaults + user changeable ones...

imsodin:
Relevant information on configuration, mostly but not only defaults: https://github.com/syncthing/syncthing/wiki/Configuration

PR #6750, opened 2020-06-16
lib/config: Apply defaults to folders/devices on config load (ref #6748)

calmh:
This makes it so that `default:"..."` struct tags on devices and folders actually take effect during JSON/XML loading, like it does for general and GUI options.

This means we can add a field to a device/folder and have it take its default value on config load without writing a config migration. (I suspect we've missed this in the past because the current
behavior is unintuitive...)

### DIFF
--- a/lib/config/config.go
+++ b/lib/config/config.go
@@ -12,7 +12,6 @@ import (
 	"encoding/xml"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net"
 	"net/url"
 	"os"
@@ -109,8 +108,8 @@ func New(myID protocol.DeviceID) Configuration {
 	util.SetDefaults(&cfg.Options)
 	util.SetDefaults(&cfg.GUI)
 
-	// Can't happen.
 	if err := cfg.prepare(myID); err != nil {
+		// Can't happen.
 		l.Warnln("bug: error in preparing new folder:", err)
 		panic("error in preparing new folder")
 	}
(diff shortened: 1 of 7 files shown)


imsodin:
Having the magic default application on folders and devices only feels like potential for future mistakes/confusion. It's clear why it's only necessary for those (slice vs non-slice), but maybe it's worth doing this for all the config elements with defaults instead of setting defaults in the `config.Read...` methods. And it may even be possible to create a utility function for it (untested, and anyway optional):

```golang
func (cfg *DeviceConfiguration) UnmarshalJSON(data []byte) error {
	return util.UnmarshalJSONWithDefaults(cfg, data)
}

func UnmarshalJSONWithDefaults(target interface{}, data []byte) error {
	SetDefaults(target)
	type noCustomUnmarshal reflect.TypeOf(target)
	ptr := (*noCustomUnmarshal)(target)
	return json.Unmarshal(data, ptr)
}
```

calmh:
Unsure about the possibility to generalize, but to be clear this is not something that now applies to folders and devices *only*. The situation currently is that defaults-on-deserialize applies to everything *but* devices and folders -- that is, the main config, GUI config and options config all get defaults as expected. Devices and folders were the exception because way back when it happened we didn't sort out the unmarshal magic I just did. This change makes the behavior the same for all the types.

imsodin:
I am aware of that, I am only talking about where the defaults are applied: For folders and devices it's now on unmarshalling, for everything else it's on a dedicated config reading function (`json.Unmarshal(&OptionsConfiguration{}, bs)` has no defaults). I don't think we ever do (un)marshal options and gui separately from the entire config, but if we added an api for doing just that in the future, we'd have to think about doing the same unmarshalling magic for those too, and that's where my proposition to apply defaults on unmarshal for everything comes in.

calmh:
Right, okay. Let me look into that.

calmh:
Not for merging, need to consider the details a bit

AudriusButkevicius:
Looks good, yet I am a bit puzzled how we manage to hide the marshalling methods via a type alias, presumably the underlying method still resolves to the type behind the alias?

calmh:
Declaring a new type drops the methods. But this solution is ugly. I was thinking today that I might as well fork encoding/json+encoding/xml to add support for a method `{Un}MarshalContext()` that would take a context and pass it to new `MarshalJSONContext()` etc methods, which would help with keeping track of things during unmarshal -- like unmarshal could then know what `myID` is and be able to do related cleanup directly. Similarly one could support `PreUnmarshal(ctx)` and `PostUnmarshal(ctx)` hooks for things like setting defaults prior to unmarshal and doing validation post unmarshal without having to do ugly conversions to temporary types and stuff.

st-review:
:robot: *beep boop*

PR #6751, opened 2020-06-16
lib/db: Improve error message on meta inconsistency

imsodin:
Not a functional change, just don't be actively confusing towards users:

> 2020-06-16 03:10:04 Stored folder metadata for "y9qce-uafde" is 442306h10m4.252789s old; recalculating
> 
> I don’t know if the ‘old’ time is a red herring, but I restarted this instance yesterday whilst looking at this issue; it certainly wasn’t 50 years ago…

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -811,7 +811,12 @@ func (db *Lowlevel) getMetaAndCheck(folder string) *metadataTracker {
 func (db *Lowlevel) loadMetadataTracker(folder string) *metadataTracker {
 	meta := newMetadataTracker()
 	if err := meta.fromDB(db, []byte(folder)); err != nil {
-		l.Infof("No stored folder metadata for %q; recalculating", folder)
+		if err == errMetaInconsistent {
+			l.Infof("Stored folder metadata for %q is inconsistent; recalculating", folder)
+		} else {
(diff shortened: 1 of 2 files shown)


calmh:
This doesn't actually change the log message in the description though. Maybe we should catch `meta.Created().IsZero()` specifically, and then also round the time to some reasonable unit before printing?

imsodin:
I considered `meta.Created().IsZero()` meaning data inconsistency an implementational detail of the metadatatracker, thus caught that condition when loading metadata from disk. Now we'll never reach the time comparison when the time is zero. 

I'll fix the unit.

calmh:
Ah, sorry, I see it now. I wonder how that would even happen, I'd think that if the blob exists it would have a timestamp, and if it doesn't then there would be an error from db.Get. Anyhow, sure.

imsodin:
Ah sorry I should have mentioned the cause, I thought that was "obvious" (I stared so much at that code lately either changing it or hoping to find what causes all the support requests with inconconsistencies that it is for me): When an update makes a counter go below 0, the counter is set to 0 instead and the created time is set to 0 to make it recalculate the next time it's loaded.

calmh:
Aight aight, all makes sense :)

PR #6752, opened 2020-06-16
lib/rand: Various fixes

greatroar:
Here's three patches to lib/rand and its use in Syncthing:

* The first removes a cast.
* The second adds one bit of randomness to TLS certificates' serial numbers. Since NewCertificate also generates a random key, this should not be crucial, but getting this extra bit doesn't hurt code complexity.
* The third clarifies the status of random functions wrt. cryptographic security. Strength of the RNG is too weak a property to rely on.


greatroar:
It's terminology combined with my misreading of math/rand source code. Intn is secure. I'll update the PR.

Issue #6758, opened 2020-06-18
"Nightly" release channel


calmh:
The release candidates aren't generally broken enough for my taste, so I like to run the latest dev build. Currently this is a little bit annoying as I need to manually build and copy out the binary to my boxes, so I only upgrade them every now and then. This seems like a waste when we have a perfectly fine build available on the build server. There are several ways I could scratch my itch:

- Hack a script to pull the latest build from the server and upgrade to it. Advantage: I can do this now. Disadvantages: It depends on how the system is set up, how Syncthing runs and where Syncthing is stored. I don't want to make a Powershell variant for Windows.

- Make the build server sign the builds and create a separate upgrade feed for it, say https://build.syncthing.net/nightly.json. The Syncthing-side config then consists of 1) enabling upgrades to pre-releases, 2) changing the upgrade server URL. Advantages: Easy to do, no Syncthing code changes. Disadvantages: It's not just a dropdown selection in the config?

- As above but include it in the main feed and add a new option in Syncthing so you can choose "Nightly" where you currently choose between stable and release candidates. Advantage: everyone can now easily access this thing. Disadvantage: maybe everyone shouldn't? Requires writing actual code.

calmh:
Imma start with option two as that is convenient for me and doesn't commit to anything onerous in the long term.

AudriusButkevicius:
I'd prefer 3 over 2, as it's easier for users

calmh:
I think we can move to 3 from 2 while the opposite is more breaking.

calmh:
https://docs.syncthing.net/users/releases.html#nightly-builds

Issue #473, opened 2014-08-05
Self built version wants to downgrade


Nutomic:
I built syncthing from master (after v0.9.0) and ran it, and the web GUI shows the "Upgrade to v0.9.0" notice.

This should probably be disabled for dirty versions, or check against the first part of the version string only.

PR #6760, opened 2020-06-18
build: Change version strings to not have plus in them (ref #6758)

calmh:
Currently a random dev version has a version string like this:

	v1.7.0-rc.1+22-g946170f3f

That is, the tag name followed by a plus sign and the git describe
metadata (number of commits and hash) plus -dirty or -branchname in some
cases.

We introduced the plus sign in #473, where a dev version would
previously be called v0.9.0-42-gwhatever which is considered older than
v0.9.0 and hence caused a downgrade. The problem with the plus is that
per semver everything after the plus is ignored as build metadata, which
means we won't upgrade from v1.7.0-rc.1 to v1.7.0-rc.1+22-g946170f3f.

With this change the we use dots instead, and a "0-dev" marker, making
dev versions greater than the tag they come from but lesser than any
future real version:

	v1.7.0-rc.1.0-dev.22.g946170f3f

Or, after a release version:

	v1.7.0.0-dev.22.g946170f3f

The leading zero is to maintain semver-ish compliance in this case, as
we can't add non-numeric things before the dash, and we can't add a dash
directly after the tag version as v1.7.0-anything is older than v1.7.0
while we want to be newer...

With this we can still take any given released version and either bump
the last number or tack on a ".1" to become newer than the dev versions
in between.

### DIFF
--- a/build.go
+++ b/build.go
@@ -34,7 +34,6 @@ import (
 )
 
 var (
-	versionRe     = regexp.MustCompile(`-[0-9]{1,3}-g[0-9a-f]{5,10}`)
 	goarch        string
 	goos          string
 	noupgrade     bool
@@ -852,15 +851,36 @@ func getReleaseVersion() (string, error) {
 }
 
 func getGitVersion() (string, error) {
-	v, err := runError("git", "describe", "--always", "--dirty")
+	// The current version as Git sees it
+	bs, err := runError("git", "describe", "--always", "--dirty")
 	if err != nil {
 		return "", err
 	}
-	v = versionRe.ReplaceAllFunc(v, func(s []byte) []byte {
-		s[0] = '+'
-		return s
-	})
-	return string(v), nil
+	vcur := string(bs)
+
+	// The closest current tag name
+	bs, err = runError("git", "describe", "--always", "--abbrev=0")
+	if err != nil {
+		return "", err
+	}
+	v0 := string(bs)
+
+	versionRe := regexp.MustCompile(`-([0-9]{1,3}-g[0-9a-f]{5,10})`)
+	if m := versionRe.FindStringSubmatch(vcur); len(m) > 0 {
+		suffix := strings.ReplaceAll(m[1], "-", ".")
+
+		if strings.Contains(v0, "-") {
+			// We're based of a tag with a prerelease string. We can just
+			// add our dev stuff directly.
+			return fmt.Sprintf("%s.dev.%s", v0, suffix), nil
+		}
+
+		// We're based on a release version. We need to bump the patch
+		// version and then add a -dev prerelease string.
+		next := nextPatchVersion(v0)
+		return fmt.Sprintf("%s-dev.%s", next, suffix), nil
+	}
+	return vcur, nil
 }
 
 func getVersion() string {
@@ -1361,4 +1381,12 @@ func trimTagMessage(msg, tag string) string {
 		msg = msg[:beginSig]
 	}
 	return strings.TrimSpace(msg)
+}
+
+func nextPatchVersion(ver string) string {
+	parts := strings.SplitN(ver, "-", 2)
+	digits := strings.Split(parts[0], ".")
+	n, _ := strconv.Atoi(digits[len(digits)-1])
+	digits[len(digits)-1] = strconv.Itoa(n + 1)
+	return strings.Join(digits, ".")
 }
--- a/lib/build/build_test.go
+++ b/lib/build/build_test.go
@@ -27,7 +27,6 @@ func TestAllowedVersions(t *testing.T) {
 		{"v0.13.0-some-weird-but-allowed-tag", true},
 		{"v0.13.0-allowed.to.do.this", true},
 		{"v0.13.0+not.allowed.to.do.this", false},
-		{"v1.6.1.1-dev.80.gf9ac3f503-dirty", true},
 	}
 
 	for i, c := range testcases {
--- a/lib/build/build_test.go
+++ b/lib/build/build_test.go
@@ -27,6 +27,7 @@ func TestAllowedVersions(t *testing.T) {
 		{"v0.13.0-some-weird-but-allowed-tag", true},
 		{"v0.13.0-allowed.to.do.this", true},
 		{"v0.13.0+not.allowed.to.do.this", false},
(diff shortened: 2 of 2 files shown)


calmh:
Possibly we could also use `1-dev` to always be conceptually a dev version of the next following point release, even though we generally don't release that point release... That is, one could imagine that after 1.7.0 comes 1.7.0.1 and 1.7.0.1-dev.42 is then something inbetween 1.7.0 and 1.7.0.1... 🤷

imsodin:
I wouldn't have thought about anything like that, but now that you brought it up it kinda makes sense - so yeah lets use `.1-dev.` :)

build.go
```diff
@@ -852,13 +851,16 @@ func getReleaseVersion() (string, error) {
 }
 
 func getGitVersion() (string, error) {
+	// The current version as Git sees it
 	v, err := runError("git", "describe", "--always", "--dirty")
 	if err != nil {
 		return "", err
 	}
+
+	versionRe := regexp.MustCompile(`-[0-9]{1,3}-g[0-9a-f]{5,10}`)
```
imsodin, line 860: Why does this var need to be local now?

AudriusButkevicius:
So I know + and - are used to denote commits ahead and commits behind, for ordering purposes. I think even semver agrees with@ this. The new scheme looks like it says its behind rc, when it's actually ahead. We're in control how we handle this, so why not stay with +?

build.go
calmh: It doesn't need to, but it was just used in one place and having it nearer where it's used seemed clearer.

calmh:
No, that's very much not at all what semver says or our implementation does :)

> Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers [...] Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence.

(https://semver.org/#spec-item-10)

AudriusButkevicius:
Feels weird that the number of separators we use is a moving target, hurts my brain to process that.

calmh:
I’m open for better patterns, but I think special casing so that -dev means newer-than while -anythingelse means older-than is asking for trouble. These things will/may leak into any number of other packaging systems which won’t have our special understanding of version strings. So I don’t mind something aesthetically better, whatever that may be, but I don’t think breaking the version parsing to accomplish it is the right way.

calmh:
If anything I guess the one special case we could do is to parse plus-strings as newer. Everyone else should ignore them and they will only affect our nightlies, which then can only be delivered by auto upgrade. So the damage is limited. (But then I can’t launch nightlies until after 1.8.0)

imsodin:
What about just immediately bumping the patch digit after release to avoid going to 4 digits, i.e. `v1.7.1-dev.22.g946170f3f` instead of `v1.7.0.1-dev.22.g946170f3f`. I mean it's not wrong, even if we then decide to actually go to v1.8.0 - it's after v1.7.0 and before either v1.7.1 or v1.8.0.

> These things will/may leak into any number of other packaging systems which won’t have our special understanding of version strings.

They are anyway getting messed up by not using semver (e.g. slash being used by debian to denote debian specific version counter for a single upstream version, which makes them a dev build with all the associated non-nicenesses). If we want to be nicer to packagers, we should not use versions to determine behaviour (or only as a fallback). I don't think putting some meaning into a plus will wreak havoc. Having to wait for 1.8.0 is annoying though...

And I know the 99 thingy like this: For dev versions after 1.7.0 use 1.7.0.99-dev (next 1.7.1) respectively 1.7.99-dev (1.8.0). I don't like that though, the proposed 1.7.0.1 achieves all goals and is less confusing in my opinion.

calmh:
Bumping the patch digit sounds fine when the preceding version is a real release. I'm not sure what to do about 1.7.0-rc.1+20 though.

We can't bump the patch to say 1.7.1-dev.20 because that's newer than a 1.7.0-rc.2 we might release later. If we bump the rc number and add -dev we'd end up at 1.7.0-rc.2-dev.20 which is also bad (it's newer than 1.7.0-rc.2). We could have a special case to not bump when there is a dash and instead just add dotted parts, getting 1.7.0-rc.1.dev.20, but that's not particularly consistent then.

calmh:
I can try adding support for comparing with plus and see how that plays out.

AudriusButkevicius:
I guess we could increment the patch version/rc version? Or foor rc's we just drop the tag?

1.7.0+some -> 1.7.1-dev.99
1.7.0-rc1+some > 1.7.1-dev99 (or 1.7.0-rc2-dev)

imsodin:
At the moment I'd tend to taking `+` into account for version comparing and for now go with variant 2 in https://github.com/syncthing/syncthing/issues/6758 (custom upgrade server). Makes most sense to me: a fourth version digit is not semver compliant either, and ugly, while the plus bit comes from git, i.e. is a de-facto standard, so not very likely that anyones going to be taken by surprise by it.

imsodin:
Do we really need interoperability between RCs and nightlies, as in is it a problem that `1.7.1-dev.20` is strictly newer than `1.7.0-rc.2`? The two are concurrent anyway - one lives on a master the other on a release branch. If your on an outdated nightly build, you likely won't have any joy regardless. And using nightlies does mean being prepared for some pain in my opinion.

calmh:
If we're to remain somewhat semver, I don't see another option than doing that and accepting that the version numbering is sometimes odd. Let's see how that looks.

AudriusButkevicius:
We could just have 4th digit be build number.

Then everything that rolls off of nightly just gets -dev tag and does not care if it's rc or non-rc.
RCs keep rc tags.

So you end up with
1.7.0.35-
1.7.0.29-rc2
1.2.0.27-dev
1.7.0.26-rc1

You can still know where 27-dev falls between 26-rc1 and 29-rc2

calmh:
I made a combo:

>With this change the we instead either just add a dev suffix (if we're
> already on a prerelease version) or we wind the patch version and add a
> dev suffix.
>
>      v1.7.0-rc.1+23-gef3441bd6 => v1.7.0-rc.1.dev.23.gef3441bd6
>      v1.6.1+80-gef3441bd6      => v1.6.2-dev.22.g946170f3f
>
> This should preserve the ordering and keep versions semver-ish.

AudriusButkevicius:
That looks fine

PR #6762, opened 2020-06-18
all: Don't send locally changed items in index and add UI controls

imsodin:
PR for encryption

### Purpose

Currently we just delete everything in a receive encrypted folder that shouldn't be there. That's a super-tough foot-gun: Pick the wrong type or path on folder creation and boom, data gone. Now it's more like a receive only folder: We only delete stuff behind the `*randomChars*.syncthing-enc/` toplevel directories. Anything else that's there gives an indication in the UI that there are files present which shouldn't be, with a clean button to delete all of these.
  
I piggy-backed on the existing `FlagLocalReceiveOnly` - I probably should add `FlagLocalNotEncrypted`, right? Requires quite a bit of "duplication" in the model/api/controller, so delaying until I get a review/go-ahead with the general behaviour.
 
### Testing

Manual only.

### Screenshots

image
image
image


lib/api/api.go
```diff
@@ -703,6 +704,12 @@ func (s *service) postDBRevert(w http.ResponseWriter, r *http.Request) {
 	go s.model.Revert(folder)
 }
 
+func (s *service) postDBCleanEnc(w http.ResponseWriter, r *http.Request) {
+	var qs = r.URL.Query()
+	var folder = qs.Get("folder")
+	go s.model.CleanEnc(folder)
```
AudriusButkevicius, line 710: Why not just revert?
I'd argue that revert/override are already the same thing, (effectively it's still a revert just in a different location), so adding a new endpoint seems like a nono, you can branch what revert does based on the folder type.

uok:
Helpful addition! 
Please keep the wording consistent. So either "Unencrypted" (my preference) for error folder status, info text, action button and modal title - or "Not Encrypted". Thanks!

lib/api/api.go
imsodin: There's already override and revert, doing similar but clearly different things. This is again something "similar", but again different from both. The consistent continuation is adding a new endpoint. So you want to deprecate the override endpoint and switch to revert only? I am fine with that. I'd like to wait for @calmh's ack or not before going ahead (just because that's going to be a bunch of boring busywork that I don't want to just revert again).

calmh: I think we can keep Override (make remotes like local) and Revert (make local like remote). This, then, is a variant of Revert.

imsodin:
Reverting now also happens through the revert endpoint and as such I think it's ok to use the same local flag on files as for changed files on receive-only folders.

gui/default/index.html
```diff
@@ -504,9 +505,12 @@ <h4 class="panel-title">
                 <button type="button" class="btn btn-sm btn-danger pull-left" ng-click="override(folder.id)" ng-if="folderStatus(folder) == 'outofsync' && folder.type == 'sendonly'">
                   <span class="fas fa-arrow-circle-up"></span>&nbsp;<span translate>Override Changes</span>
                 </button>
-                <button type="button" class="btn btn-sm btn-danger pull-left" ng-click="revert(folder.id)" ng-if="canRevert(folder.id)">
+                <button type="button" class="btn btn-sm btn-danger pull-left" ng-click="revert(folder.id)" ng-if="folder.type === 'receiveonly' && hasLocalChanged(folder)">
                   <span class="fa fa-arrow-circle-down"></span>&nbsp;<span translate>Revert Local Changes</span>
                 </button>
+                <button type="button" class="btn btn-sm btn-danger pull-left" ng-click="revertEncModal(folder.id)" ng-if="folder.type === 'receiveencrypted' && hasLocalChanged(folder)">
+                  <span class="fa fa-arrow-lock"></span>&nbsp;<span translate>Revert Not Encrypted Items</span>
```
AudriusButkevicius, line 512: This is not really a revert, this would be a delete presumably, which is destructive?

We have revert local changes, which undoes changes but I suspect it does not delete extra files does it?

gui/default/syncthing/transfer/localChangedFilesModalView.html
```diff
@@ -1,8 +1,12 @@
-<modal id="localChanged" status="info" icon="fas fa-exclamation-circle" heading="{{'Locally Changed Items' | translate}}" large="yes" closeable="yes">
-  <div class="modal-body">
-    <p translate>
+<modal id="localChanged" status="{{localChangedType === 'receiveencrypted' ? 'warning' : 'info'}}" icon="fas fa-exclamation-circle" heading="{{localChangedHeading(localChangedType)}}" large="yes" closeable="yes">
+  <div class="modal-body" ng-switch="localChangedType">
+    <p ng-switch-when="receiveonly" translate>
       The following items were changed locally.
     </p>
+    <p ng-switch-when="receiveencrypted">
+      <span translate>The following items are present and not encrypted.</span>
```
AudriusButkevicius, line 7: So we don't know if they are encrypted or not, but are unexpected as they were not received.

lib/model/folder.go
```diff
@@ -450,11 +450,15 @@ func (f *folder) scanSubdirs(subDirs []string) error {
 		}
 	case f.Type == config.FolderTypeReceiveEncrypted:
 		batchAppend = func(fi protocol.FileInfo, _ *db.Snapshot) {
-			// Delete changed items and set zero version vector
-			// such that we get a correct copy again.
-			if err := mtimefs.RemoveAll(fi.Name); err != nil {
-				l.Debugf(`%v Failed to remove changed item "%v": %v`, f.Description(), fi.Name, err)
+			if protocol.IsEncryptedPath(fi.Name) {
+				// Delete changed, encrypted items.
```
AudriusButkevicius, line 454: What happens for stuff that's not under the encrypted path?

imsodin: They get the local flag change. That's what the revert button in the ui is about. Everything that's behind our sentinel gets deleted automatically, everything outside gets the flag such that it shows in UI to be deleted by user action (and gets marked as invalid etc).

gui/default/index.html
imsodin: Yes, revert local changes does delete locally added files as well as reverting files to the global state. I'll rename it to delete - specific is always better.

gui/default/syncthing/transfer/localChangedFilesModalView.html
imsodin: Any ideas for a understandable/readable sentence that means something? As opposed to what's there now or something like "The following unexpected items are present in the folder."

AudriusButkevicius: `The following items are present in the folder but were not received from remote devices`

or what you had makes sense too.

lib/model/folder.go
AudriusButkevicius: Are we kosher with stuff being deleted automatically?
We never delete stuff automatically.

imsodin: I know we don't. I thought this might be the one exception: Having files with garbage names and a top-level directory named `a.syncthing-encrypted` should make it highly unlikely that someone puts files in there (as in willfully stupid, not just stupid). At the same time the more I think about it, the less benefit I see in "auto-repairing" stuff (what would we be repairing?). So yeah, maybe we should just mark everything as invalid instead.

imsodin:
Changed it to never delete anything automatically. In addition fixed a related issue @calmh pointed out (out-of-PR-band): We can't ever send locally changed/created file infos, as these aren't encrypted and thus produce an error on the remote. Therefore I am now filtering these out in the index sender.

PR #6763, opened 2020-06-18
build, lib/upgrade: Compare build metadata after plus in version (ref #6758)

calmh:
This is an alternative to #6760. It makes only a small change to how we represent versions (dot vs dash in one place), but instead diverges more from semver in assigning meaning to the build metadata field. Personally I think this is less desirable because regardless of what we do with the version string in the other variant it's still parseable with a semver mentality, while here we're totally out on our own.

The only reason for the existence of the plus-part is to be not taken into account. Any other meaning can anyways be expressed in the release or prerelease parts...

With this change we take into account the stuff after "+" in version
comparisons, when all else is equal. Having something after the plus is
newer than not having something. If both versions have something after
the plus they are compared in the same manner as prerelease parts (after
"-") are compared: in dot separated components, numerically if both are
numbers or alphabetically if both are non-numbers. Having differing
types in the same position in the version string is confusing, don't do
that.

Also changes the build metadata to be "+commits.hash" instead of
"+commits-hash" in order to get a dot separator so the number of commits
gets compared numerically.


calmh:
Actually no, this is bullshit and I refuse to offer it for consideration.

PR #6764, opened 2020-06-18
lib/fs: Add support for Windows duplicate extents

AudriusButkevicius:
Another one

### DIFF
--- a/lib/fs/basicfs_copy_range_duplicateextents.go
+++ b/lib/fs/basicfs_copy_range_duplicateextents.go
@@ -1,0 +1,128 @@
+// Copyright (C) 2020 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// +build windows
+
+package fs
+
+import (
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+)
+
+func init() {
+	registerCopyRangeImplementation(CopyRangeMethodDuplicateExtents, copyRangeImplementationForBasicFile(copyRangeDuplicateExtents))
+}
+
+// Inspired by https://github.com/git-lfs/git-lfs/blob/master/tools/util_windows.go
+
+var (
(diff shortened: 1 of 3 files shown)


calmh:
How do you test any of this?

AudriusButkevicius:
I didn't, ran locally, it just said it doesn't support it so 🤷‍♂️ 
Don't have Windows Server 2016 to try it, nor do I run a SMB/CIFS server.

This is pretty much a copy of git-lfs code.

We could add it to UR and see if some internet hero tires to use it.

calmh:
Mmm... I mean we have a fairly lax attitude to testing in the best of cases, but these are all just "might make every file synced just a bunch of zeroes lol who knows" which feels more irresponsible than usual...

AudriusButkevicius:
It's more likely it fails than writes zeros.

Again, it's an advanced option, with no trivial way to enable it, so only people who really want will go for it, and if it's a shotgun, it's a shotgun, it wouldn't be the first one, modTimeWindowS and disableFsync are already your options there.

We can put in the docs that it's not tested, and some lunatic with ReFS can try it out and tell us if it works.

calmh:
Yeah there are options that can have bad results when used wrong. But here we have stuff that we don't know if it works or not, and if it doesn't it might eat your data. I don't think we can have that around even as an advanced option in good conscience. But we had this discussion with the previous PR as well.

I think what we need is a doc article that says 1) how to enable these things 2) in big red warning note, that they are untested and that we want feedback, and then a forum post saying these are new and untested options and we would like people with the requisite software & spirit of adventure to test and report back. 

With a couple of successful test reports I think we can agree that it does what it says on the tin, barring odd corner cases, which is like anything else.

Without any successful testing we don't know if we're holding the syscall upside down and copying zeroes back to the original or whatever.

AudriusButkevicius:
So I am working on the docs as we speak, but I don't expect people to go and build their own version of the application to be able to tell us.

calmh:
That's fine, as long as it's not an undocumented tasty-looking trap ready to spring on the unwary.

AudriusButkevicius:
To find the possible values people will have to go to the docs, and it will be declared as untested there.

Also, how do you feel about sneaking this into UR without asking for permission?

calmh:
I think that's fine. UR gets enabled with feature reporting, I think it's reasonable expectation that when a new feature is introduced it is included in the report. This is not a change in what we report to the extent that it changes the consent consideration.

PR #6768, opened 2020-06-20
build: Disable Cgo on windows

imsodin:
Cgo is optionally required by badger only (not for long any more hopefully https://github.com/dgraph-io/badger/issues/1162). Lets just not use that and prevent any hassle around getting a C compiler set up.

### DIFF
--- a/build.go
+++ b/build.go
@@ -500,10 +500,13 @@ func build(target target, tags []string) {
 func setBuildEnvVars() {
 	os.Setenv("GOOS", goos)
 	os.Setenv("GOARCH", goarch)
-	if goos == "windows" {
-		os.Setenv("CGO_ENABLED", "0")
-	} else {
-		os.Setenv("CC", cc)
+	os.Setenv("CC", cc)
+	if os.Getenv("CGO_ENABLED") == "" {
+		switch goos {
+		case "darwin", "solaris":
+		default:
+			os.Setenv("CGO_ENABLED", "0")
+		}
 	}
 }
 
--- a/build.go
+++ b/build.go
@@ -451,9 +451,7 @@ func install(target target, tags []string) {
 	}
 	os.Setenv("GOBIN", filepath.Join(cwd, "bin"))
 
-	os.Setenv("GOOS", goos)
-	os.Setenv("GOARCH", goarch)
-	os.Setenv("CC", cc)
+	setBuildEnvVars()
 
 	// On Windows generate a special file which the Go compiler will
 	// automatically use when generating Windows binaries to set things like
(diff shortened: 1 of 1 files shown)


build.go
```diff
@@ -501,6 +497,16 @@ func build(target target, tags []string) {
...
 }
 
+func setBuildEnvVars() {
+	os.Setenv("GOOS", goos)
+	os.Setenv("GOARCH", goarch)
+	if goos == "windows" {
+		os.Setenv("CGO_ENABLED", "0")
+	} else {
+		os.Setenv("CC", cc)
+	}
```
AudriusButkevicius, line 507: I think we always should set cc as that is passed via the command line, I believe.

I think the only want CGO_ENABLED on android no and maybe OS X for watcher stuff?
Presumably the same issue would happen on Linux if I did not have CGO enabled?

imsodin: It gets a bit murky there:

```$ grep -r 'import "C"' vendor/
vendor/golang.org/x/sys/unix/zsyscall_aix_ppc.go:import "C"
vendor/golang.org/x/sys/unix/mkerrors.sh:       echo 'import "C"'
vendor/golang.org/x/sys/unix/zsyscall_aix_ppc64_gccgo.go:import "C"
vendor/github.com/syncthing/notify/watcher_fen_cgo.go:import "C"
vendor/github.com/syncthing/notify/watcher_fsevents_cgo.go:import "C"
vendor/github.com/DataDog/zstd/zstd.go:import "C"
vendor/github.com/DataDog/zstd/zstd_stream.go:import "C"
vendor/github.com/DataDog/zstd/errors.go:import "C"
vendor/github.com/shirou/gopsutil/disk/disk_darwin_cgo.go:import "C"
```

We do have ppc builds, they might need (or profit from) cgo.

Feels like hand-holding cgo is at least as bad as creating a badger fork removing the cgo bits - though as mentioned that might be wasted effort soon.

AudriusButkevicius: The CC argument still holds tho

AudriusButkevicius: To be honest, my gut says that debian et all might complain when we suddenly require a c compiler.

calmh: Only set CGO_ENABLED=0 if the builder did not already set CGO_ENABLED to anything, I think.

AudriusButkevicius: Sounds good.

imsodin: Did that and never set it on solaris and darwin, as it's needed for notify there. That means building for ppc/ppc64/aix maybe requires the caller of build.go to set `CGO_ENABLED`, though probably that's not needed.

calmh: We already set CGO_ENABLED=0 for everything (I think) on the build server, so for us this is a no-op. The solaris etc builds end up being without notifications enabled.

imsodin: I wasn't aware of that, I thought we the mac builder was in place for that. Actually now I am confused why our mac builds even pass: I don't see us setting a `kqueue` build tag anywhere and then notify should try to build fsevents, which requires cgo.

PR #6770, opened 2020-06-20
lib/ur: Normalise contract between syncthing and ursrv

AudriusButkevicius:
Also:
1. Fixes potential overflows in 32bit UR server when computing stats
2. Fixes the preview diffing UI not working (angular scoping is a mess)

In theory we could also remove the urVersion >= stuff, as the "clear" functionality now un-sets any fields that are from future version.


AudriusButkevicius:
This now also makes it mongo web scale so js kids would love it, by migrating all the reports to jsonb field in postgres.

This means adding new fields to the report now requires no changes to ursrv (it still requires to roll out UR with new "contract" but I'd like the rollout of that to be easier)

Tested by taking 1000 rows from prod, loading to my local postgres, running the migration, and running the aggregations.
Stuff still seems to work.

Will require storage expansion atleast temporarily, as it will duplicate the data during the migration.

AudriusButkevicius:
Obviously the motivation here is to making the task of tracking new stuff easier.
We add features, yet never do the leg work for usage tracking, and I want to put an end to that.

AudriusButkevicius:
Also, in theory I could do the migration fully in SQL, not pulling the data to the client, but I have no way to do "omitempty" in SQL.

lib/ur/contract/contract.go
```diff
@@ -0,0 +1,468 @@
...
+import (
+	"database/sql/driver"
+	"encoding/json"
+	"errors"
+	"reflect"
+	"sort"
+	"strconv"
+	"time"
+
+	"github.com/lib/pq"
```
calmh, line 18: This makes syncthing binary now depend on pq driver and the sql stuff. Would be nice to avoid...

AudriusButkevicius: Do you know how? I could put the parts that depends on sqp behind build flags I guess.

calmh: Or back into cmd/ursrv

imsodin: The sql/driver isn't necessary, you don't need to "name" the interfaces to implement the relevant methods, the returned types just need to implement those interfaces. And to get rid of pq we need to "reimplement" an int64 array, which isn't super nice but hardly problematic. I wasn't that sure as it might sound here if it really works, thus I testes it and it seems to do :)

```diff
--- a/lib/ur/contract/contract.go
+++ b/lib/ur/contract/contract.go
@@ -7,27 +7,24 @@
 package contract
 
 import (
-	"database/sql/driver"
 	"encoding/json"
 	"errors"
 	"reflect"
 	"sort"
 	"strconv"
 	"time"
-
-	"github.com/lib/pq"
 )
 
 type IntMap map[string]int
 
-func (p IntMap) Value() (driver.Value, error) {
+func (p IntMap) Value() ([]byte, error) {
 	return json.Marshal(p)
 }
 
 func (p *IntMap) Scan(src interface{}) error {
 	source, ok := src.([]byte)
 	if !ok {
-		return errors.New("Type assertion .([]byte) failed.")
+		return errors.New("type assertion .([]byte) failed.")
 	}
 
 	var i map[string]int
@@ -40,6 +37,28 @@ func (p *IntMap) Scan(src interface{}) error {
 	return nil
 }
 
+type Int64Array []int64
+
+func (p Int64Array) Value() ([]byte, error) {
+	return json.Marshal(p)
+}
+
+func (p *Int64Array) Scan(src interface{}) error {
+	source, ok := src.([]byte)
+	if !ok {
+		return errors.New("type assertion .([]byte) failed.")
+	}
+
+	var i []int64
+	err := json.Unmarshal(source, &i)
+	if err != nil {
+		return err
+	}
+
+	*p = i
+	return nil
+}
+
 type Report struct {
 	// Generated
 	Received time.Time `json:"-"` // Only from DB
@@ -109,8 +128,8 @@ type Report struct {
 	UpgradeAllowedAuto   bool `json:"upgradeAllowedAuto,omitempty" since:"2"`
 
 	// V2.5 fields (fields that were in v2 but never added to the database
-	UpgradeAllowedPre bool          `json:"upgradeAllowedPre,omitempty" since:"2"`
-	RescanIntvs       pq.Int64Array `json:"rescanIntvs,omitempty" since:"2"`
+	UpgradeAllowedPre bool       `json:"upgradeAllowedPre,omitempty" since:"2"`
+	RescanIntvs       Int64Array `json:"rescanIntvs,omitempty" since:"2"`
 
 	// v3 fields
 
@@ -132,18 +151,18 @@ type Report struct {
 	CustomStunServers          bool   `json:"customStunServers,omitempty" since:"3"`
 
 	FolderUsesV3 struct {
-		ScanProgressDisabled    int           `json:"scanProgressDisabled,omitempty" since:"3"`
-		ConflictsDisabled       int           `json:"conflictsDisabled,omitempty" since:"3"`
-		ConflictsUnlimited      int           `json:"conflictsUnlimited,omitempty" since:"3"`
-		ConflictsOther          int           `json:"conflictsOther,omitempty" since:"3"`
-		DisableSparseFiles      int           `json:"disableSparseFiles,omitempty" since:"3"`
-		DisableTempIndexes      int           `json:"disableTempIndexes,omitempty" since:"3"`
-		AlwaysWeakHash          int           `json:"alwaysWeakHash,omitempty" since:"3"`
-		CustomWeakHashThreshold int           `json:"customWeakHashThreshold,omitempty" since:"3"`
-		FsWatcherEnabled        int           `json:"fsWatcherEnabled,omitempty" since:"3"`
-		PullOrder               IntMap        `json:"pullOrder,omitempty" since:"3"`
-		FilesystemType          IntMap        `json:"filesystemType,omitempty" since:"3"`
-		FsWatcherDelays         pq.Int64Array `json:"fsWatcherDelays,omitempty" since:"3"`
+		ScanProgressDisabled    int        `json:"scanProgressDisabled,omitempty" since:"3"`
+		ConflictsDisabled       int        `json:"conflictsDisabled,omitempty" since:"3"`
+		ConflictsUnlimited      int        `json:"conflictsUnlimited,omitempty" since:"3"`
+		ConflictsOther          int        `json:"conflictsOther,omitempty" since:"3"`
+		DisableSparseFiles      int        `json:"disableSparseFiles,omitempty" since:"3"`
+		DisableTempIndexes      int        `json:"disableTempIndexes,omitempty" since:"3"`
+		AlwaysWeakHash          int        `json:"alwaysWeakHash,omitempty" since:"3"`
+		CustomWeakHashThreshold int        `json:"customWeakHashThreshold,omitempty" since:"3"`
+		FsWatcherEnabled        int        `json:"fsWatcherEnabled,omitempty" since:"3"`
+		PullOrder               IntMap     `json:"pullOrder,omitempty" since:"3"`
+		FilesystemType          IntMap     `json:"filesystemType,omitempty" since:"3"`
+		FsWatcherDelays         Int64Array `json:"fsWatcherDelays,omitempty" since:"3"`
 	} `json:"folderUsesV3,omitempty" since:"3"`
 
 	GUIStats struct {
@@ -193,8 +212,8 @@ func New() *Report {
 	r.FolderUsesV3.FilesystemType = make(IntMap)
 	r.GUIStats.Theme = make(IntMap)
 	r.TransportStats = make(IntMap)
-	r.RescanIntvs = make(pq.Int64Array, 0)
-	r.FolderUsesV3.FsWatcherDelays = make(pq.Int64Array, 0)
+	r.RescanIntvs = make(Int64Array, 0)
+	r.FolderUsesV3.FsWatcherDelays = make(Int64Array, 0)
 	return r
 }
 
@@ -409,7 +428,7 @@ func (r *Report) FieldNames() []string {
 	}
 }
 
-func (r Report) Value() (driver.Value, error) {
+func (r Report) Value() (string, error) {
 	// This needs to be string, yet we read back bytes..
 	bs, err := json.Marshal(r)
 	return string(bs), err
@@ -461,7 +480,7 @@ func clear(v interface{}, since int) error {
 	return nil
 }
 
-func SortPqInt64Array(slice pq.Int64Array) {
+func SortInt64Array(slice Int64Array) {
 	sort.Slice(slice, func(a, b int) bool {
 		return slice[a] < slice[b]
 	})
diff --git a/lib/ur/usage_report.go b/lib/ur/usage_report.go
index 9c7b87dd8..a4bcbad20 100644
--- a/lib/ur/usage_report.go
+++ b/lib/ur/usage_report.go
@@ -153,7 +153,7 @@ func (s *Service) reportData(ctx context.Context, urVersion int, preview bool) (
 			l.Warnf("Unhandled versioning type for usage reports: %s", cfg.Versioning.Type)
 		}
 	}
-	contract.SortPqInt64Array(report.RescanIntvs)
+	contract.SortInt64Array(report.RescanIntvs)
 
 	for _, cfg := range s.cfg.Devices() {
 		if cfg.Introducer {
@@ -256,7 +256,7 @@ func (s *Service) reportData(ctx context.Context, urVersion int, preview bool) (
 			report.FolderUsesV3.FilesystemType[cfg.FilesystemType.String()]++
 			report.FolderUsesV3.FsWatcherDelays = append(report.FolderUsesV3.FsWatcherDelays, int64(cfg.FSWatcherDelayS))
 		}
-		contract.SortPqInt64Array(report.FolderUsesV3.FsWatcherDelays)
+		contract.SortInt64Array(report.FolderUsesV3.FsWatcherDelays)
 
 		guiCfg := s.cfg.GUI()
 		// Anticipate multiple GUI configs in the future, hence store counts.
```

AudriusButkevicius: So pq.Int64Array is not json, it's has a special way it gets serialized on the wire for postgres to put it in it's native array.

Anyways, I already worked it out with some build tag magic.
Every binary now injects a build tag with it's name, so we can gate files based on that.

AudriusButkevicius: It's a bit grim that you can't in theory just do `go install ursrv` but I guess that ship has sailed many moons ago.

AudriusButkevicius: It seems we pull in sql driver anyway:
```
/cygdrive/c/Gohome/src/github.com/syncthing/syncthing (ur-2) $ go mod why database/sql/driver
# database/sql/driver
github.com/syncthing/syncthing/lib/connections
github.com/lucas-clemente/quic-go
github.com/lucas-clemente/quic-go/qlog
github.com/francoispqt/gojay
database/sql
database/sql/driver
```

We don't pull in postgres driver via anything else tho, so I am starting to question whether it's worth it.

calmh: `database/sql/driver` is a small package so I wouldn't worry about it. `pq.Int64Array` is just a slice of int64 so I'm not sure why we don't just have `[]int64` in `contract` and have cmd/ursrv do a `whatever := pq.Int64Array(the things from contract)` where necessary? Or, in a more philosophical way of looking at it, why would the contract between syncthing and ursrv need to concern itself with the database engine in use?

AudriusButkevicius: Actually, because we store it as json, we in theory don't need it to be pg.Int64Array, but we still need it to load the existing data from pg's int array column for the purpose of the migration.

So maybe do the hack now, once the migration is done, delete the code for that purpose.

Or have extra logic in the migration code, which I guess is less evil.

AudriusButkevicius: So I see only two ways forward for this:

1. Keep existing build tag magic which removes both sql and pg (the PR as is)
2. Move the use of pg package to ursrv explicitly, changing Report not to have pg.Int64Array, and having to read them out into separate variables during the migration. This removes the use of pg package, but keeps the use of sql package in the shared package.

The changes to build.go I made I feel are generally useful, so perhaps we'd want to keep that?


Which one should I go for?

calmh: Number two feels better to me. Cleanups in build.go are probably fine but I don't like the program-name-as-tag thing. It sounds neat on the face of it but enables misuse with packages having optional parts compiled in and out on a whim. I think it would be better to keep packages focused enough that there's no need to have half of a package disappear based on who the intended consumer is.

AudriusButkevicius:
Right, how about this?

calmh:
lgtm apart from not building

calmh:
and I didn't look at any of the ursrv-side or migration things because ew sql

AudriusButkevicius:
there is not that much sql going on, so I guess I'd like it to be fully reviewed.

Also, obviously before deploying this, we should make a copy of the db, and expand the existing storage.

AudriusButkevicius:
Just ran a local migration again, still seems to work

AudriusButkevicius:
I could do the migration this way:
image

But omit empty is going to be a PITA.

calmh:
I've increased the volume size, we can probably just stop postgresql and tar.gz the whole dir to get a reasonable backup.

```
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda        197G   49G  139G  26% /mnt/data
```

AudriusButkevicius:
Do you want me to do the migration?
What's the process of landing latest bnaries on the machines?

calmh:
So apparently we auto deploy new ursrv:s and the migration is now in progress by itself

AudriusButkevicius:
Lol ok, no backups then I guess :D

AudriusButkevicius:
We might want to stop it, take a backup of the db and restart it?

calmh:
Yolo. Got some really popular features here though;

image

AudriusButkevicius:
Interesting.

PR #6774, opened 2020-06-21
lib/api, lib/model: Fixes around event request tracking

imsodin:
Two small things:

1. When refactoring the api I clearly didn't grok the function of the "minute-variable" in the folder summary service. Luckily I only changed the name but not semantics, so nothing broke but the name was completely off.

2. We don't register all calls to the event endpoint, e.g. not those to the `event/disk` endpoint. I don't see why we'd do that, thus I moved the `OnEventSub` call to the generic `getEvents` function.

### DIFF
--- a/lib/api/api.go
+++ b/lib/api/api.go
@@ -1202,7 +1202,6 @@ func (s *service) postDBIgnores(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *service) getIndexEvents(w http.ResponseWriter, r *http.Request) {
-	s.fss.OnEventRequest()
 	mask := s.getEventMask(r.URL.Query().Get("events"))
 	sub := s.getEventSub(mask)
 	s.getEvents(w, r, sub)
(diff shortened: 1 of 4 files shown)


lib/api/api.go
```diff
@@ -1213,7 +1213,9 @@ func (s *service) getDiskEvents(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *service) getEvents(w http.ResponseWriter, r *http.Request, eventSub events.BufferedSubscription) {
-	s.fss.OnEventRequest()
+	if eventSub.Mask()&events.FolderSummary != 0 {
```
calmh, line 1216: or `events.FolderCompletion`

lib/model/folder_summary.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/syncthing/syncthing/lib/util"
 )
 
-const minSummaryInterval = time.Minute
+const maxDurSinceLastEventReq = time.Minute
```
mikeee, line 25: `maxDurationSinceLastEventReq` for the sake of verbosity seems reasonable perhaps?

lib/events/events.go
```diff
@@ -238,6 +238,7 @@ type Event struct {
 type Subscription interface {
 	C() <-chan Event
 	Poll(timeout time.Duration) (Event, error)
+	Mask() EventType
```
calmh, line 241: Maybe `Allows(EventType) bool` would be cleaner (just moving the `& ... != 0`)?

imsodin: `Allows` feels kind of weird in this context, and e.g. `SubscribedTo` would be even worse. The mask value is also used as a map key in the api, i.e. there could be a use-case for actually getting the value. I think I'll choose simple over clean here.

PR #6777, opened 2020-06-21
 lib/nat: Make service termination faster

AudriusButkevicius:
Noticed it can take a while

### DIFF
--- a/lib/pmp/pmp.go
+++ b/lib/pmp/pmp.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/jackpal/gateway"
 	"github.com/jackpal/go-nat-pmp"
+	"github.com/pkg/errors"
 
 	"github.com/syncthing/syncthing/lib/nat"
 	"github.com/syncthing/syncthing/lib/util"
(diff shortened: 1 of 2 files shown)


lib/pmp/pmp.go
```diff
@@ -44,10 +45,18 @@ func Discover(ctx context.Context, renewal, timeout time.Duration) []nat.Device
...
-	_, err = c.GetExternalAddress()
-	if err != nil && strings.Contains(err.Error(), "Timed out") {
-		l.Debugln("Timeout trying to get external address, assume no NAT-PMP available")
-		return nil
+	err = util.CallWithContext(ctx, func() error {
+		_, ierr := c.GetExternalAddress()
+		return ierr
+	})
+	if err != nil {
+		if errors.Cause(err) == context.Canceled {
```
calmh, line 53: Does util specifically use `github.com/pkg/errors` for wrapping? If so I guess this is fine (but should be flagged for future refactor), otherwise `errors.Is` seems more appropriate here now

imsodin: FYI: pkg/errors is compatible with go1.13's errors.Is/As methods.

greatroar: Actually, compatibility was [withdrawn](https://github.com/pkg/errors/pull/220) in pkg/errors 0.9.1.

```go
package main

import (
	"fmt"
	"github.com/pkg/errors"
)

func main() {
	err := fmt.Errorf("wrapped: %w", fmt.Errorf("inner"))
	fmt.Println(errors.Cause(err)) // prints wrapped: inner
}
```

imsodin: Ok, not so nice - thanks for the hint. The other way around should still work as far as I see (i.e. you can use `errors.Is` on an error wrapped with `errors.Wrap`). So as we are using `fmt.Errorf` in places, we definitely shouldn't use `errors.Cause` anymore.

AudriusButkevicius: It's a wider fix that needs to happen through the codebase, hence I am still merging this.

Also, I am not sure this actually affects context.Cancelled, as it's created with `errors.New` so does not have Is, so the only way this could go wrong if something wrapped the error which then implemented `Is` and explicitly returned false.

PR #6778, opened 2020-06-21
lib/model: More connection closing

AudriusButkevicius:
Not sure wtf happened here. Did this never work sensibly?

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -166,11 +166,8 @@ var (
 	errDeviceIgnored     = errors.New("device is ignored")
 	errDeviceRemoved     = errors.New("device has been removed")
 	ErrFolderPaused      = errors.New("folder is paused")
-	errFolderUnpaused    = errors.New("folder is unpaused")
 	errFolderNotRunning  = errors.New("folder is not running")
 	errFolderMissing     = errors.New("no such folder")
-	errFolderRemoved     = errors.New("folder removed")
-	errFolderShared      = errors.New("folder (un)shared")
 	errNetworkNotAllowed = errors.New("network not allowed")
 	errNoVersioner       = errors.New("folder has no versioner")
 	// errors about why a connection is closed
@@ -2429,7 +2426,6 @@ func (m *model) CommitConfiguration(from, to config.Configuration) bool {
 				l.Infoln("Adding folder", cfg.Description())
 				m.newFolder(cfg)
 			}
-			m.closeConns(cfg.DeviceIDs(), errFolderShared)
 		}
 	}
 
@@ -2438,7 +2434,6 @@ func (m *model) CommitConfiguration(from, to config.Configuration) bool {
 		if !ok {
 			// The folder was removed.
 			m.removeFolder(fromCfg)
-			m.closeConns(fromCfg.DeviceIDs(), errFolderRemoved)
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
The current user experience:

image

lib/model/model.go
```diff
@@ -2426,6 +2429,7 @@ func (m *model) CommitConfiguration(from, to config.Configuration) bool {
 				l.Infoln("Adding folder", cfg.Description())
 				m.newFolder(cfg)
 			}
+			m.closeConns(cfg.DeviceIDs(), errFolderShared)
```
calmh, line 2432: newFolder calls closeConns, and a new paused folder shouldn't require closing

lib/model/model.go
```diff
@@ -2434,6 +2438,7 @@ func (m *model) CommitConfiguration(from, to config.Configuration) bool {
 		if !ok {
 			// The folder was removed.
 			m.removeFolder(fromCfg)
+			m.closeConns(fromCfg.DeviceIDs(), errFolderRemoved)
```
calmh, line 2441: removeFolder calls stopFolder which calls closeConns

lib/model/model.go
```diff
@@ -2445,13 +2450,19 @@ func (m *model) CommitConfiguration(from, to config.Configuration) bool {
...
+				m.closeConns(diff, errFolderShared)
+			}
 		}
 
 		// Emit the folder pause/resume event
 		if fromCfg.Paused != toCfg.Paused {
 			eventType := events.FolderResumed
 			if toCfg.Paused {
 				eventType = events.FolderPaused
+				m.closeConns(toCfg.DeviceIDs(), ErrFolderPaused)
```
calmh, line 2463: The folder is restarted when paused/resumed, which has closeConns calls

lib/model/model.go
```diff
@@ -2445,13 +2450,19 @@ func (m *model) CommitConfiguration(from, to config.Configuration) bool {
 		// Check if anything differs that requires a restart.
 		if !reflect.DeepEqual(fromCfg.RequiresRestartOnly(), toCfg.RequiresRestartOnly()) {
 			m.restartFolder(fromCfg, toCfg)
+			if diff := difference(fromCfg.DeviceIDs(), toCfg.DeviceIDs()); len(diff) > 0 {
```
calmh, line 2453: restartFolder calls closeConns

lib/model/model.go
AudriusButkevicius: It would in theory. For consistency we should send cc with paused=true

AudriusButkevicius:
I'll follow this up with some tests.

AudriusButkevicius:
I've added some tests and a fix for the one test case that is broken.

lib/model/model.go
```diff
@@ -496,6 +493,8 @@ func (m *model) restartFolder(from, to config.FolderConfiguration) {
 	}
 
 	m.stopFolder(from, fmt.Errorf("%v folder %v", errMsg, to.Description()))
+	// Need to send CC change to both from and to devices.
+	m.stopFolder(to, fmt.Errorf("%v folder %v", errMsg, to.Description()))
```
imsodin, line 497:
```suggestion
	err := fmt.Errorf("%v folder %v", errMsg, to.Description()))
	m.stopFolder(from, err)
	// Need to send CC change to both from and to devices.
	m.closeConns(to.DeviceIDs, err)
```

AudriusButkevicius:
Build servers seem to be broken, can't resolve github.com

PR #6787, opened 2020-06-22
all: Add untrusted config for devices

imsodin:
PR for the encryption branch

Adds an `Untrusted` config option to devices. If set it ensures that no data is sent to that device without being encrypted.

image
image

The first commit is a convenience change: We already do keep folder configs as a map in $scope, now we do the same for devices.


gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -133,6 +133,13 @@
               </div>
             </div>
           </div>
+          <div class="row">
+            <div class="form-group col-md-6">
+              <input type="checkbox" id="untrusted" ng-model="currentDevice.untrusted" />
+              <label for="untrusted" translate>Untrusted</label>
+              <p translate class="help-block">All folders shared with this device must be protected by a password, such that all sent data is encrypted.</p>
```
AudriusButkevicius, line 140: `such that all sent data is unreadable without the given password`?

Encrypted might just mean TLS...

imsodin:
Is this good to merge into the main encryption branch?

lib/model/model.go
```diff
@@ -1233,12 +1234,16 @@ func (m *model) ccHandleFoldersLocked(folders []protocol.Folder, deviceCfg confi
...
 
 // requires fmut read-lock
-func (m *model) ccCheckEncryptionLocked(fcfg config.FolderConfiguration, folderDevice config.FolderDeviceConfiguration, ccDevice, ccDeviceUs protocol.Device, hasDevice, hasUs bool) error {
+func (m *model) ccCheckEncryptionLocked(fcfg config.FolderConfiguration, folderDevice config.FolderDeviceConfiguration, ccDevice, ccDeviceUs protocol.Device, hasDevice, hasUs, deviceUntrusted bool) error {
 	hasTokenDev := hasDevice && len(ccDevice.EncPwToken) > 0
 	hasTokenUs := hasUs && len(ccDeviceUs.EncPwToken) > 0
 	isEncDev := folderDevice.EncryptionPassword != ""
 	isEncUs := fcfg.Type == config.FolderTypeReceiveEncrypted
 
+	if !isEncDev && !isEncUs && deviceUntrusted {
```
AudriusButkevicius, line 1243: The variable names are horrible (and were horrible from the previous PRs), but shouldn't this be:

`cfg.Type  == config.FolderTypeReceiveEncrypted && untrusted && folderDevice.EncryptionPassword == ""`

which is

`!isEncDev && isEncUs && deviceUntrusted`

lib/config/deviceconfiguration.go
```diff
@@ -30,6 +30,7 @@ type DeviceConfiguration struct {
 	IgnoredFolders           []ObservedFolder     `xml:"ignoredFolder" json:"ignoredFolders"`
 	PendingFolders           []ObservedFolder     `xml:"pendingFolder" json:"pendingFolders"`
 	MaxRequestKiB            int                  `xml:"maxRequestKiB" json:"maxRequestKiB"`
+	Untrusted                bool                 `xml:"untrusted" json:"untrusted"`
```
AudriusButkevicius, line 33: Could add UR stats straight away :D

lib/model/model.go
imsodin: I don't think so: When we receive encrypted content (`cfg.Type == config.FolderTypeReceiveEncrypted`), we don't need to encrypt to an untrusted devices (well actually we can't, it doesn't make sense and it's an error condition further down). This says: If we don't have encrypted data locally, nor encrypt our plain data for the given device and that device is untrusted, that's bad.

Just because they were in a previous PR doesn't mean I can't change them anymore - shoot away with better ones.

AudriusButkevicius: But isn't in that case we're just proxying what we got?
Why do we need a password in that case?

imsodin: We don't need a pw then: When we are proxying encrypted stuff `cfg.Type == config.FolderTypeReceiveEncrypted` is true and thus this error doesn't get emitted.

AudriusButkevicius: Right, get it. 

I think we should fix the Enc Pw stuff to more readable names, and us and dev shoudl be local and remote.

lib/ur/contract/contract.go
```diff
@@ -237,6 +241,8 @@ func (r *Report) FieldPointers() []interface{} {
 		&r.FolderUsesV3.AlwaysWeakHash, &r.FolderUsesV3.CustomWeakHashThreshold,
 		&r.FolderUsesV3.FsWatcherEnabled,
 
+		&r.DeviceUsesV3.Untrusted,
```
AudriusButkevicius, line 244: This is not needed, but if you're adding it, you should also add the column name,

imsodin: Ah yeah right, lazily searched for the wrong term (FolderUsesV3).instead of trying to understand. Why not necessary though?

AudriusButkevicius: This was used for reading columns, and inserting into json tables. The migration is done, we don't have the data that was behind those columns.

imsodin: Removed it. Maybe that code should be pruned or moved to a file with "legacy" in its name just to make it clear?

AudriusButkevicius: Agree, we can delete the code given the migration is done, I guess unless someone builds from source, or does a major version jump on their own UR server.

PR #6790, opened 2020-06-23
lib/ur: Zero our the value before unmarshal

AudriusButkevicius:
Who knew?!


AudriusButkevicius:
I've also added a commit with a comment, as I had the "create new pr" window open, and github seems to have cached what the head of the branch looks like. Weirds.

PR #6799, opened 2020-06-25
lib/protocol: Include WeakHash in BlocksHash

imsodin:
The resolution of way too much debugging is just a single line...

All my pre-adler weakhash files were being rehashed over and over again, even after applying the other two PR fixes from today. That's because on rehash the weak hashes changed, but the sha256 hashes didn't. And as the `BlocksHash` doesn't consider weak hashes, block deduplication kicked in and prevented the new block list from being saved. Thus when the file was requested again, the block hash still didn't match and it got scanned again.

This will prevent the scan rename detecting by blockhash map to kick in for new files, as existing files have the sha256-only bockshash, while new files include the weak hash. It's not critical, but also somewhat ugly - a migration won't hurt. I'll consider that tomorrow :)

Testing: I finally got my new node to sync up :)

### DIFF
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -745,6 +745,11 @@ func (db *schemaUpdater) updateSchemaTo14(_ int) error {
 		}
 	}
 
+	dbi.Release()
+	if err != dbi.Error() {
+		return err
+	}
+
 	return t.Commit()
 }
 
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -27,9 +27,10 @@ import (
 //   8-9: v1.4.0
 //   10-11: v1.6.0
 //   12-13: v1.7.0
+//   14: v1.8.0
 const (
-	dbVersion             = 13
-	dbMinSyncthingVersion = "v1.7.0"
+	dbVersion             = 14
+	dbMinSyncthingVersion = "v1.8.0"
 )
 
 var errFolderMissing = errors.New("folder present in global list but missing in keyer index")
@@ -95,6 +96,7 @@ func (db *schemaUpdater) updateSchema() error {
 		{10, db.updateSchemaTo10},
 		{11, db.updateSchemaTo11},
 		{13, db.updateSchemaTo13},
+		{14, db.updateSchemaTo14},
 	}
 
 	for _, m := range migrations {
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
Sorry, are we fixing a bug here or what?
It's not obvious to me whats happening here.

imsodin:
I am fixing my production problems. And am tired, thus probably less clear than I though I was, sorry.

Problem is that if block weak hashes only change when scanning, that doesn't get recorded in the db. That's because of block list duplication, which is based on `BlocksHash`, which previously didn't change when just the weak hash changes.

This bug only affects people with files that were scanned pre 2017 (aka probably only me). And it will affect everyone if we ever choose to switch weak hashes in the future.

I'll look into a test too tomorrow. Night.

AudriusButkevicius:
If the hash changed, why would it not get recorded in the database?

I wonder if this is going to produce a different side effect, where the block lists will go missing because the hash is now computed differently than before?

calmh:
I think this should be fine, and doesn't require a migration. Existing files and block hashes remain, on change a new block list will get a new-format hash, and it would anyway have changed 🤷

imsodin:
I believe a migration is necessary: Without it if you add a new node with prepopulated data on disk, we will create a conflict copy for every single file (as we use `BlocksHash` as a shortcut to comparing individual block sha256 to check for equivalence).

lib/db/schemaupdater.go
```diff
@@ -683,6 +685,84 @@ func (db *schemaUpdater) updateSchemaTo13(prev int) error {
 	return t.Commit()
 }
 
+func (db *schemaUpdater) updateSchemaTo14(_ int) error {
```
calmh, line 688: This seems overly complicated, isn't it just rewriteFiles plus a GC pass?

imsodin: That doesn't cover `BlockListMap`. However I just realized there's an alternative option that doesn't require a migration and addresses the conflict problem. It is a bit less efficient but more robust, as it also prevents conflicts e.g. between a device on an RC (already done the migration) and one on stable:

```diff
--- a/lib/protocol/bep_extensions.go
+++ b/lib/protocol/bep_extensions.go
@@ -276,9 +276,11 @@ func PermsEqual(a, b uint32) bool {
 
 // BlocksEqual returns true when the two files have identical block lists.
 func (f FileInfo) BlocksEqual(other FileInfo) bool {
-       // If both sides have blocks hashes then we can just compare those.
-       if len(f.BlocksHash) > 0 && len(other.BlocksHash) > 0 {
-               return bytes.Equal(f.BlocksHash, other.BlocksHash)
+       // If both sides have blocks hashes and they match, we are good. If they
+       // don't match still check individual block hashes to catch differences
+       // in weak hashes only (e.g. after switching weak hash algo).
+       if len(f.BlocksHash) > 0 && len(other.BlocksHash) > 0 && bytes.Equal(f.BlocksHash, other.BlocksHash) {
+               return true
        }
 
        // Actually compare the block lists in full.
```

Any gotchas coming to mind going that route?

calmh: Seems fine to me

AudriusButkevicius:
Man, these sort of changes scare the shit out of me. It looks like a one line change, but it can fuck everything up royally. I can't say if it looks fine as it seems quite wide reaching and I have not though through the scenarios.

lib/protocol/protocol_test.go
```diff
@@ -887,7 +887,8 @@ func TestBlocksEqual(t *testing.T) {
 		{blocksOne, nil, blocksOne, nil, true},          // blocks compared
 		{blocksOne, nil, blocksTwo, nil, false},         // blocks compared
 		{blocksOne, hashOne, blocksTwo, hashOne, true},  // hashes equal, blocks not looked at
-		{blocksOne, hashOne, blocksOne, hashTwo, false}, // hashes different, blocks not looked at
+		{blocksOne, hashOne, blocksOne, hashTwo, true},  // hashes different, blocks ompared
+		{blocksOne, hashOne, blocksTwo, hashTwo, false}, // hashes different, blocks ompared
```
calmh, line 891:
```suggestion
		{blocksOne, hashOne, blocksOne, hashTwo, true},  // hashes different, blocks compared
		{blocksOne, hashOne, blocksTwo, hashTwo, false}, // hashes different, blocks compared
```

PR #6808, opened 2020-07-02
lib/api, lib/model: Improve folder completion API (fixes #6075)

### LINKED ISSUE #6075: Feature: API check sync status of folder[s]
Sorry, I can not find a reliable, supported way to do an api call that reports about the sync status of one or all folder(s). I would like to do something on a folder with a bash script, but only if that folder is synced.

I found a few hints in the forum:
https://docs.syncthing.net/rest/db-completion-get.html looks like a simple `completion < 100` will do it, however my local computer gives me this result for a folder:
```json
{
  "completion": 0,
  "globalBytes": 2027069255,
  "needBytes": 2027069255,
  "needDeletes": 0,
  "needItems": 251738
}
```
Value of `completion` is 0 - why that? So that easy one will not work. Unfortunately I can not find any hint in the docs that helps me.

Then I found https://docs.syncthing.net/rest/db-status-get.html - here I _guess_ that `needBytes` could be a good indicator if a directory is synced, but I am not sure - I would like to not guess.

I have to dig much too deep for such a fundamentally important info - the syncthing api should make it much easier to get the sync status.

So I would like to propose an API call `GET rest/db/sync` that will give simple binary yes/no results about the sync status of all folders if called without options or of one folder if called with an optional folder name.

Meanwhile it would be very lovely if anybody would like to tell me what is the most reliable way to get the current sync status of a folder, thanks!

calmh:
This makes the /rest/db/completion call more useful and user friendly.

Prior to this change the only supported use was with both "folder" and
"device" parameters, where the device is a remote device. This then
gives the completion status for specifically that folder on that device.

With this change we also allow the following:

- Omitting the "device" parameter, in which case it means the local
  device (and returns the local completion)

- Explicitly specifying the local device ID, which means the same thing.

- Omitting the "folder" parameter, in which case it means all folders.

- Omitting both, which logically then means the completion percentage
  for all folders as a whole on this device.

This should be closer to what people expect usage wise, while also now
giving a single place to check for a completion percentage for the local
device or any remote device as a whole.

### DIFF
--- a/lib/api/api.go
+++ b/lib/api/api.go
@@ -676,13 +676,41 @@ func (s *service) getDBCompletion(w http.ResponseWriter, r *http.Request) {
 	var folder = qs.Get("folder")
 	var deviceStr = qs.Get("device")
 
-	device, err := protocol.DeviceIDFromString(deviceStr)
-	if err != nil {
-		http.Error(w, err.Error(), 500)
-		return
+	// We will check completion status for either the local device, or a
+	// specific given device ID.
+
+	device := protocol.LocalDeviceID
+	if deviceStr != "" {
+		var err error
+		device, err = protocol.DeviceIDFromString(deviceStr)
+		if err != nil {
+			http.Error(w, err.Error(), 500)
+			return
+		}
 	}
 
-	sendJSON(w, s.model.Completion(device, folder).Map())
+	// The user specifically asked for our own device ID. Internally that is
+	// known as protocol.LocalDeviceID so translate.
+	if device == s.id {
+		device = protocol.LocalDeviceID
+	}
+
+	var comp model.FolderCompletion
+	if folder != "" {
+		// We want completion for a specific folder.
+		comp = s.model.Completion(device, folder)
+	} else {
+		// We want completion for all folders shared with the device, as an
+		// aggregate. Grab completion for all folders shared with the device
+		// and aggregate.
+		for _, fcfg := range s.cfg.FolderList() {
+			if device == protocol.LocalDeviceID || fcfg.SharedWith(device) {
+				comp.Add(s.model.Completion(device, fcfg.ID))
+			}
+		}
+	}
+
+	sendJSON(w, comp.Map())
 }
 
 func (s *service) getDBStatus(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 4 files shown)


AudriusButkevicius:
Also, we're using it ourselves in the UI?

calmh:
I don't see any killing sprees being possible here. UI does its thing, which is based on folders and events so I don't think it needs to take advantage of this.

AudriusButkevicius:
Why would we not want to use this in the ui to have consistency with consumers?

calmh:
I don't understand what you're asking. Of course we use the completion call in the UI, but there we (need to) ask for device-and-folder specific values, so we don't need these extensions or they would have happened years ago.

lib/api/api.go
```diff
@@ -676,13 +676,39 @@ func (s *service) getDBCompletion(w http.ResponseWriter, r *http.Request) {
...
+	if folder != "" {
+		// We want completion for a specific folder.
+		comp = s.model.Completion(device, folder)
+	} else {
+		// We want completion for all folders as an aggregate.
+		for _, fcfg := range s.cfg.FolderList() {
+			if device == protocol.LocalDeviceID || fcfg.SharedWith(device) {
+				comp.Add(s.model.Completion(device, fcfg.ID))
+			}
+		}
```
imsodin, line 708: I'd probably have done the aggregation to model.go, thus not exposing `model.FolderCompletion` and keeping api code "dumb".

lib/model/model.go
```diff
@@ -765,26 +831,10 @@ func (m *model) Completion(device protocol.DeviceID, folder string) FolderComple
 		need.Bytes = 0
 	}
 
-	needRatio := float64(need.Bytes) / float64(tot)
-	completionPct := 100 * (1 - needRatio)
+	comp := NewFolderCompletion(counts.Bytes, need.Bytes, counts.Files+counts.Directories+counts.Symlinks, need.Files+need.Directories+need.Symlinks, need.Deleted)
```
AudriusButkevicius, line 834: Almost feel like this should be NewFolderCompeltion(global, need) and the summing happens internally.

calmh: yes

PR #6809, opened 2020-07-02
lib/fs: Set O_NOATIME flag on Linux

greatroar:
With O_NOATIME, every read call no longer needs to update the access time on a file. This gives much less disk traffic, especially on an initial scan.

### DIFF
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -220,6 +220,7 @@ func (f *BasicFilesystem) Open(name string) (File, error) {
 	if err != nil {
 		return nil, err
 	}
+	setNoatime(fd)
 	return basicFile{fd, name}, err
 }
 
@@ -232,6 +233,7 @@ func (f *BasicFilesystem) OpenFile(name string, flags int, mode FileMode) (File,
 	if err != nil {
 		return nil, err
 	}
+	setNoatime(fd)
 	return basicFile{fd, name}, err
 }
 
--- a/lib/fs/noatime_linux.go
+++ b/lib/fs/noatime_linux.go
@@ -1,0 +1,29 @@
+// Copyright (C) 2020 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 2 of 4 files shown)


AudriusButkevicius:
My gut is saying that people who really care about this, should mount the filesystem with no atime.
Hiding that we've accessed files is not a great default I think.

imsodin:
I had the same thought, but I also think it's a sensible default at the very least for "home-users" (I set noatime by default in my fstab). I also can't think of a use-case for knowing that syncthing accessed a file in a "professional" setting, but nevertheless there might be. If the gains of setting noatime are significant (I guess you made some rough tests @greatroar ?), maybe it's sensible to have noatime by default and an env var to activate setting atime?

calmh:
I also set noatime and can’t imagine a use for it. My objection is more that this oddly inverts an established default and removes the choice to actually use atimes. (I assume someone does.)

As with other odd looking changes, I’d like to hear about prior art. Is it common for Linux applications to manually set noatime themselves? Who does this and what was their rationale?

greatroar:
[rsync](https://github.com/WayneD/rsync/blob/c528f8d5c8aa7b16b20cda72a9f4119699890c28/syscall.c#L206) does this, but it requires a flag. So does [bup](https://github.com/bup/bup/blob/2c1f4acab0355b28c4f8fb9f8d0329ec6b2c2afa/lib/bup/_helpers.c#L1742). restic used to do this unconditionally until the functionality get inadvertently removed (restic/restic#2816). Unison doesn't seem to do this, as far as I can tell. I don't know about Dropbox.

I thought I had some good performance figures, but it seems like I screwed up and I was measuring the MtimeFS performance :frowning_face:. I'll try again with a full initial scan.

calmh:
I'd be surprised if one (cached) write would have a significant impact when we're reading a full file. But if so, could we do it by adding O_NOATIME to the OpenFile call in the scanner, and have a folder option for it (default on)?

greatroar:
Closing as I don't have time to run full benchmarks for this. Remounting is less intrusive, and newer Linux distros mount with relatime anyway.

Issue #6807, opened 2020-07-02
"File Pull Order" should be disabled (greyed out) for "Send Only" folders in the GUI


tomasz1986:
This has just come to my mind when fiddling with Syncthing options, but should "File Pull Order" not be disabled (greyed out) in the Web GUI when the folder is "Send Only"?

Right now, it can be set regardless of the folder type, but changing it does nothing for "Send Only" folders. I think that it will be better to grey it out in this case, just to avoid any unnecessary confusion.

tomasz1986:
The issue has been marked fixed, but I have just noticed that the "File Pull Order" is still shown in the GUI for "Send Only" folders, after unfolding the folder's details. This only happens when it has been set to a non-default value before setting the folder type to "Send Only".

I think that it should be hidden here too.

![image](https://user-images.githubusercontent.com/5626656/89048861-5ea4e080-d38b-11ea-9d77-e7e9ec95662d.png)

PR #6811, opened 2020-07-05
gui: Hide file pull order when folder is send only

jccroft1:
Fixes issue #6807 

### Testing

As stated by the below article, update default gui folder. Then run syncthing and setup new folder or edit existing folder, go to the Advanced tab and chage "Folder Type" to "Send Only" and observe "File Pull Order" is hidden.  
https://docs.syncthing.net/dev/web.html# 

NB: May need to force refresh to ensure html page isn't cached. 

### Screenshots

image

### DIFF
--- a/gui/default/syncthing/folder/editFolderModalView.html
+++ b/gui/default/syncthing/folder/editFolderModalView.html
@@ -207,7 +207,7 @@
                   <p ng-if="currentFolder.type == 'sendonly'" translate class="help-block">Files are protected from changes made on other devices, but changes made on this device will be sent to the rest of the cluster.</p>
                   <p ng-if="currentFolder.type == 'receiveonly'" translate class="help-block">Files are synchronized from the cluster, but any changes made locally will not be sent to other devices.</p>
                 </div>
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
I wonder if we should hide the control all together. Making it read only gives a false impression that there is some fixed mode for that, where in reality that setting has no effect.

calmh:
I'm OK with either, but would have preferred disabling it and setting the value to `N/A` or such. I'm generally not super excited about dialog elements appearing and disappearing based on other choices. One day you go looking for this option and it's not there and you wonder if you had an aneurism and forgot where it was supposed to be. If it's instead there but greyed it's more obvious that something else is preventing it from being set.

jccroft1:
> I'm OK with either, but would have preferred disabling it and setting the value to `N/A` or such. I'm generally not super excited about dialog elements appearing and disappearing based on other choices. One day you go looking for this option and it's not there and you wonder if you had an aneurism and forgot where it was supposed to be. If it's instead there but greyed it's more obvious that something else is preventing it from being set.

Created new PR #6821

PR #6814, opened 2020-07-06
lib: Use encoding.TextUnmarshaler for utils.SetDefaults

imsodin:
When adding a new default value for `fs.FilesystemType` for #6717 that the `defaultParser` interface duplicates `encoding.TextUnmarshaler` that config types already need to implement for xml/json encoding.


imsodin:
Yeah no. Apparently json encoding considers an object implementing `TextUnmarshaller` as exclusively marshalling through text and thus fails if it gets asked to unmarshal a byte representation that is not text, but generated through the struct tags.

PR #6821, opened 2020-07-08
gui: Disable file pull order when for send-only (fixes #6807)

### LINKED ISSUE #6807: "File Pull Order" should be disabled (greyed out) for "Send Only" folders in the GUI

jccroft1:
Fixes issue #6807

### Testing

As stated by the below article, update default gui folder. Then run syncthing and setup new folder or edit existing folder, go to the Advanced tab and chage "Folder Type" to "Send Only" and observe "File Pull Order" is set to disabled.  
https://docs.syncthing.net/dev/web.html# 

NB: May need to force refresh to ensure html page isn't cached. 

### Screenshots

image

### DIFF
--- a/gui/default/syncthing/folder/editFolderModalView.html
+++ b/gui/default/syncthing/folder/editFolderModalView.html
@@ -209,13 +209,16 @@
                 </div>
                 <div class="col-md-6 form-group">
                   <label translate>File Pull Order</label>
-                  <select class="form-control" ng-model="currentFolder.order">
+                  <select class="form-control" ng-model="currentFolder.order" ng-if="currentFolder.type != 'sendonly'">
                     <option value="random" translate>Random</option>
                     <option value="alphabetic" translate>Alphabetic</option>
                     <option value="smallestFirst" translate>Smallest First</option>
                     <option value="largestFirst" translate>Largest First</option>
                     <option value="oldestFirst" translate>Oldest First</option>
                     <option value="newestFirst" translate>Newest First</option>
+                  </select>
+                  <select class="form-control" ng-if="currentFolder.type == 'sendonly'" disabled>
+                    <option value="disabled" translate>Disabled</option>
                   </select>
                 </div>
               </div>


gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -199,24 +199,27 @@
                 <div class="col-md-6 form-group">
                   <label translate>Folder Type</label>
                   &nbsp;<a href="https://docs.syncthing.net/users/foldertypes.html" target="_blank"><span class="fas fa-question-circle"></span>&nbsp;<span translate>Help</span></a>
-                  <select class="form-control" ng-model="currentFolder.type">
+                  <select class="form-control" ng-model="currentFolder.type" ng-change="typeChange()">
```
imsodin, line 202: You shouldn't need all this code to react to a change here, angularjs takes care of changing values, i.e. the `ng-show` condition below changes if the folder type changes.

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -199,24 +199,27 @@
...
                     <option value="receiveonly" translate>Receive Only</option>
                   </select>
                   <p ng-if="currentFolder.type == 'sendonly'" translate class="help-block">Files are protected from changes made on other devices, but changes made on this device will be sent to the rest of the cluster.</p>
                   <p ng-if="currentFolder.type == 'receiveonly'" translate class="help-block">Files are synchronized from the cluster, but any changes made locally will not be sent to other devices.</p>
                 </div>
-                <div class="col-md-6 form-group">
+                <div class="col-md-6 form-group" >
                   <label translate>File Pull Order</label>
-                  <select class="form-control" ng-model="currentFolder.order">
+                  <select class="form-control" ng-model="currentFolder.order" ng-show="currentFolder.type != 'sendonly'">
```
imsodin, line 212: I think we almost exclusively use ng-if which entirely removes the element from the DOM, while this just hides it. I don't know whether it matters much, but it seems cleaner to remove it entirely than just hide it. You could even put the ng-if on the `option`s, not needing a second select. And I'd also put an `ng-disabled="currentFolder.type == 'sendonly'" on it.

gui/default/syncthing/folder/editFolderModalView.html
jccroft1: During my testing I had no luck relying on the ng-show. It still creates a new select option even with the fixes below. Am I missing something? 

image

```html
<select class="form-control" ng-model="currentFolder.order" ng-disabled="currentFolder.type == 'sendonly'">
    <option value="random" ng-if="currentFolder.type != 'sendonly'" translate>Random</option>
    <option value="alphabetic" ng-if="currentFolder.type != 'sendonly'" translate>Alphabetic</option>
    <option value="smallestFirst" ng-if="currentFolder.type != 'sendonly'" translate>Smallest First</option>
    <option value="largestFirst" ng-if="currentFolder.type != 'sendonly'" translate>Largest First</option>
    <option value="oldestFirst" ng-if="currentFolder.type != 'sendonly'" translate>Oldest First</option>
    <option value="newestFirst" ng-if="currentFolder.type != 'sendonly'" translate>Newest First</option>
    <option value="disabled" ng-if="currentFolder.type == 'sendonly'" translate>Disabled</option>
</select>
```

AudriusButkevicius:
I think this is OTT in what it tries to do.

If the folder type is receive only, we should just render a custom select with a single "Disabled" entry, without any binding to the model, and leave the existing value set on the model as is.

gui/default/syncthing/folder/editFolderModalView.html
imsodin: Ok, so angularjs is doing angularjs (as in being annoying). Then just what Audrius proposed below: Still two selects with an `ng-if` each.

PR #6828, opened 2020-07-11
lib/scanner: Less strict validation (fixes #6827)

### LINKED ISSUE #6827: syncthing 1.6.1 -> 1.7.0 fails on "failed validating data"

1.7.0 on both machines. Ubuntu 20.04. ZFS filesystem.

Keeps looping on trying to transfer a lot of (seemingly) random files. I've tried resetting databases on both sides, and neither helped.

If I grep on one failing file I can find a lot of these errors:
```
[EX3KG] 2020/07/10 22:40:35.363987 model.go:1526: DEBUG: model@0xc000500c00 REQ(in) failed validating data (<nil>): DEVICE_ID: "py2zx-sydjy" / ".kube/bin/kubectl-1.13.0" o=19791872 s=131072
[EX3KG] 2020/07/10 22:40:35.364247 model.go:1599: DEBUG: model@0xc000500c00 recheckFile: DEVICE_ID: "py2zx-sydjy" / ".kube/bin/kubectl-1.13.0"
[EX3KG] 2020/07/10 22:40:35.364309 model.go:1526: DEBUG: model@0xc000500c00 REQ(in) failed validating data (<nil>): DEVICE_ID: "py2zx-sydjy" / ".kube/bin/kubectl-1.13.0" o=19267584 s=131072
[EX3KG] 2020/07/10 22:40:35.364384 model.go:1599: DEBUG: model@0xc000500c00 recheckFile: DEVICE_ID: "py2zx-sydjy" / ".kube/bin/kubectl-1.13.0"
```

Sometimes these also show up in the log:

```
[EX3KG] 2020/07/10 22:40:31.469657 model.go:1453: DEBUG: model@0xc000500c00 REQ(in): DEVICE_ID: "py2zx-sydjy" / ".kube/bin/kubectl-1.13.0" o=19922944 s=131072 t=false
[EX3KG] 2020/07/10 22:40:31.469722 model.go:1453: DEBUG: model@0xc000500c00 REQ(in): DEVICE_ID: "py2zx-sydjy" / ".kube/bin/kubectl-1.13.0" o=20316160 s=131072 t=false
[EX3KG] 2020/07/10 22:40:31.469801 model.go:1453: DEBUG: model@0xc000500c00 REQ(in): DEVICE_IDF: "py2zx-sydjy" / ".kube/bin/kubectl-1.13.0" o=19529728 s=131072 t=false
```

I have a full debug log that I can send to someone in private if needed, but I'm not sure how safe it is to link all the debug info on github?

calmh:
This fixes the change in #6674 where the weak hash became a deciding
factor. Now we again just use it to accept a block, but don't take a
negative as meaning the block is bad.

### DIFF
--- a/lib/scanner/blocks.go
+++ b/lib/scanner/blocks.go
@@ -109,11 +109,11 @@ func Blocks(ctx context.Context, r io.Reader, blocksize int, sizehint int64, cou
 }
 
 // Validate quickly validates buf against the 32-bit weakHash, if not zero,
-// else against the cryptohash hash, if len(hash)>0.
-// It is satisfied if neither hash is given.
+// else against the cryptohash hash, if len(hash)>0. It is satisfied if
+// either hash matches or neither hash is given.
 func Validate(buf, hash []byte, weakHash uint32) bool {
-	if weakHash != 0 {
-		return adler32.Checksum(buf) == weakHash
+	if weakHash != 0 && adler32.Checksum(buf) == weakHash {
+		return true
 	}
 
 	if len(hash) > 0 {


AudriusButkevicius:
Its a bit grim as we lose out of cpu savings of not having to do sha... perhaps it would be better to fix up the hashes and put the workarounds away? Or atleast trigger a rehash to fix these over time.

calmh:
So, we do rehash the file if the mismatch is on our side, but here we're comparing against the weak hash given to us by the requestor. If they for whatever reason have an old or incorrect weak hash that doesn't match the block we will drop the request and they won't know their metadata is bad. I think it's better to serve them the block since the data is actually correct. In practice we'll only take this branch on old weakhashes or actually bad blocks, both of which should be rare enough in practice that the performance difference doesn't matter.

PR #6834, opened 2020-07-14
gui: Add GUI control for versioning cleanup interval (ref #6693)

calmh:
### Purpose

Actually allow someone to set the option in question.

### Testing

Clickety click

### Screenshot

Screenshot 2020-07-14 at 11 16 57


AudriusButkevicius:
Should we just put them side by side?

calmh:
It's also shown for staggered versioning which doesn't have the same "cleanout after" so it's not trivial

PR #6835, opened 2020-07-14
Modified the Docker entrypoint so that the Syncthing Docker image can be started as a non-root user

### LINKED ISSUE #6832: Support running syncthing Docker image as non-root user

The generated Docker image doesn't allow running the the image as a non-root user. Running as a non-root user is useful as it reduces permission given to the container when eg running in Kubernetes. This will reduce the attack surface.

Currently I'm rewriting /bin/entrypoint.sh as a workaround for this issue, but it would be good if such a local workaround weren't needed.

What currently happens:
<pre>
$ docker run -p 8384:8384 -p 22000:22000 \
>   -v "$(pwd)/data":/var/syncthing \
>   --user 1000:1000 \
>   syncthing/syncthing:latest
su-exec: setgroups(1000): Operation not permitted
</pre>

HansK-p:
Modified the Docker entrypoint so that the Syncthing Docker image can be started as a non-root user

### Purpose

Allow running the Syncthing Docker image as a non-root user (fixes #6832)

### Testing

Before change:
<pre>
# This works
rm -rf data && mkdir data
docker run -p 8384:8384 -p 22000:22000 \
  -v "$(pwd)/data":/var/syncthing \
  syncthing/syncthing:latest

# This fails even when the user starting the Docker daemon has uid/gid 1000
rm -rf data && mkdir data
docker run -p 8384:8384 -p 22000:22000 \
  -v "$(pwd)/data":/var/syncthing \
  --user 1000:1000 \
  syncthing/syncthing:latest
</pre>

After change:
<pre>
# This still works
rm -rf data && mkdir data
docker run -p 8384:8384 -p 22000:22000 \
  -v "$(pwd)/data":/var/syncthing \
  syncthing:latest # Local syncthing Docker image

# This also works when the user starting the Docker daemon has uid/gid 1000
rm -rf data && mkdir data
docker run -p 8384:8384 -p 22000:22000 \
  -v "$(pwd)/data":/var/syncthing \
  --user 1000:1000 \
  syncthing:latest # Local syncthing Docker image
</pre>

### Documentation

Ask me if you would like me to add a line about running the Docker image as a non-root user.

## Authorship

based on the commit metadata.

### DIFF
--- a/script/docker-entrypoint.sh
+++ b/script/docker-entrypoint.sh
@@ -2,6 +2,10 @@
 
 set -eu
 
-chown "${PUID}:${PGID}" "${HOME}" \
-  && exec su-exec "${PUID}:${PGID}" \
-     env HOME="$HOME" "$@"
+if [ "$(id -u)" = '0' ]; then
+  chown "${PUID}:${PGID}" "${HOME}" \
+    && exec su-exec "${PUID}:${PGID}" \
+       env HOME="$HOME" "$@"
+else
+  exec "$@"
+fi


calmh:
Thanks!

HansK-p:
And thank you for your fast followup.

PR #6837, opened 2020-07-16
gui: sharpen device icons (fixes #5579)

### LINKED ISSUE #5579: Avoid anti-aliasing artifacts for svg device icons
In many icons lines between the squares are visible. These are anti-aliasing artifacts and can be avoided with:
```html
<svg shape-rendering="optimizeSpeed">
```
(https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering)

### Version Information

Syncthing Version: v0.14.50
Browser Version: Chrome

aelindeman:
### Purpose

This commit will add the rule [`shape-rendering: crispEdges`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering) to the `.identicon` class to remove blurry grid lines within device icons.

This fixes #5579.

### Testing

No tests are included but can be added if necessary.

#### Before

image

#### After

image

### Documentation

No documentation changes are included but can be added if necessary.

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -241,6 +241,7 @@ identicon {
 .identicon {
     width: 1em;
     height: 1em;
+    shape-rendering: crispEdges;
 }
 
 a.toggler {


aelindeman:
If a markup change is preferred over a CSS one, `shape-rendering` could instead be added as an attribute alongside where the class is added: https://github.com/syncthing/syncthing/blob/fdf5a5c0d78413ba489c05b3fd08f3ed72493a3d/gui/default/syncthing/core/identiconDirective.js#L31

PR #6838, opened 2020-07-17
lib/db: Rework flush hooks

AudriusButkevicius:
This moves the preFlush/commit hooks to be a property of the transaction.

I noticed that flushes might happen during Put's, which can happen when updating block map, or simply during t.updateGlobal while iterating over updateLocalFiles/updateRemoteFiles, which might flush out the transaction, but not run any hooks hence would not serialise the metadata.

In theory is should never happen as Checkpoint flushes over dbFlushBatchMin (64k), where as Put flushes over dbFlushBatchMax (1mb), so I think it's unlikely that in a tight loop where we call checkpoint after every file we go from less than dbFlushBatchMin in the previous iteration and suddenly to over dbFlushBatchMax in a single iteration, but regardless, this feels like a more robust way to do this.

I started looking at this effectively due to:
https://forum.syncthing.net/t/v1-6-1-local-and-global-state-swapped-between-two-nodes/15209/8

It seems that 2 devices are both missing the same index entries, making a small revelation that this index fudge problem is probably originating on the sending side and not the receiving side.

PS: Also, flush every 64k? Seems like tiny transaction size.

### DIFF
--- a/lib/db/backend/backend.go
+++ b/lib/db/backend/backend.go
@@ -15,6 +15,8 @@ import (
 	"github.com/syncthing/syncthing/lib/locations"
 )
 
+type CommitHook func(WriteTransaction) error
+
 // The Reader interface specifies the read-only operations available on the
 // main database and on read-only transactions (snapshots). Note that when
 // called directly on the database handle these operations may take implicit
@@ -61,7 +63,7 @@ type ReadTransaction interface {
 type WriteTransaction interface {
 	ReadTransaction
 	Writer
-	Checkpoint(...func() error) error
+	Checkpoint() error
 	Commit() error
 }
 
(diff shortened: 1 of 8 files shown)


calmh:
> Also, flush every 64k? Seems like tiny transaction size.

I think we changed this lately to reduce memory usage when doing large operations that don't have natural checkpoints, like deleting an entire folder or some of the migrations.

(tl;dr on the actual change in my current state, will get to it, but it sounds reasonable, though probably not the root cause for the issue in question?)

AudriusButkevicius:
Yeah, not the root cause, but my gut says that if you setup up a burn test, where one side continuously sends new index entries from A to B, with either A or B occasionally restarting, it should be not that hard to reproduce.

PR #6841, opened 2020-07-19
dev: add github action file #5436

### LINKED ISSUE #5436: MultiArch for Docker

qinghon:
### Purpose

Add support MultiArch for Docker image, use `Github Action` for build

fixes #5436 

### Testing

build as Github Action and run test as arm64/amd64(no armv7 device but should be run).

### Documentation

The repository administrator needs to add the username and password of hub.docker.com to secrets as DOCKER_USERNAME and DOCKER_PASSWORD.

## Authorship

based on the commit metadata.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,71 @@
+
+name: buildx
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  buildx:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Prepare
+        id: prepare
+        run: |
+          DOCKER_IMAGE=${GITHUB_REPOSITORY}
+          DOCKER_PLATFORMS=linux/amd64,linux/arm/v7,linux/arm64
+          VERSION=edge
+
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+            VERSION=${GITHUB_REF#refs/tags/}
+          fi
+
+          TAGS="--tag ${DOCKER_IMAGE}:${VERSION}"
+          if [[ $VERSION =~ ^v[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ ]]; then
+            TAGS="$TAGS --tag ${DOCKER_IMAGE}:latest"
+          fi
+
+          echo ::set-output name=docker_image::${DOCKER_IMAGE}
+          echo ::set-output name=version::${VERSION}
+          echo ::set-output name=buildx_args::--platform ${DOCKER_PLATFORMS} \
+            --build-arg VERSION=${VERSION} \
+            --build-arg BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ') \
+            --build-arg VCS_REF=${GITHUB_SHA::8} \
+            ${TAGS} --file ./Dockerfile .
+      -
+        name: Set up Docker Buildx
+        uses: crazy-max/ghaction-docker-buildx@v3
+      -
+        name: Docker Buildx (build)
+        run: |
+          echo '${{ steps.prepare.outputs.buildx_args }}'|awk '{print $4}'|awk -F= '{print $2}' |tee RELEASE 
+          [[ -z "$(cat RELEASE)" ]] && exit 1
+          docker buildx build --output "type=image,push=false" ${{ steps.prepare.outputs.buildx_args }}
+      -
+        name: Docker Login
+        if: success() && github.event_name != 'pull_request'
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,71 @@
+
+name: buildx
+
+on:
```
calmh, line 4: This seems like something that should run always to verify the build, and just not push if it’s not a tag?

.github/workflows/main.yml
```diff
@@ -0,0 +1,71 @@
...
+          TAGS="--tag ${DOCKER_IMAGE}:${VERSION}"
+          if [[ $VERSION =~ ^v[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ ]]; then
+            TAGS="$TAGS --tag ${DOCKER_IMAGE}:latest"
+          fi
+
+          echo ::set-output name=docker_image::${DOCKER_IMAGE}
+          echo ::set-output name=version::${VERSION}
+          echo ::set-output name=buildx_args::--platform ${DOCKER_PLATFORMS} \
+            --build-arg VERSION=${VERSION} \
+            --build-arg BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ') \
```
calmh, line 37: Are these (build_date and vcs_ref used for anything?

.github/workflows/main.yml
```diff
@@ -0,0 +1,71 @@
...
+            --build-arg VCS_REF=${GITHUB_SHA::8} \
+            ${TAGS} --file ./Dockerfile .
+      -
+        name: Set up Docker Buildx
+        uses: crazy-max/ghaction-docker-buildx@v3
+      -
+        name: Docker Buildx (build)
+        run: |
+          echo '${{ steps.prepare.outputs.buildx_args }}'|awk '{print $4}'|awk -F= '{print $2}' |tee RELEASE
```
calmh, line 46: What does this RELEASE stuff do?

.github/workflows/main.yml
qinghon: Sure! If want test ,may be fork and push a tag .
If want enable nighty version，first need to modify the Dockerfile and add ARG.

.github/workflows/main.yml
qinghon: no ,anything not use the args, but if you want [this](https://github.com/syncthing/syncthing/pull/6841#discussion_r456991005) ,need it

.github/workflows/main.yml
qinghon: Do [this](https://github.com/syncthing/syncthing/blob/55147f5901ca8be3d6c06762bbec2f4fd6e42b82/build.go#L854), content is tag name. eg: `v1.8.0-rc.2`
If not ,[getGitVersion()](https://github.com/syncthing/syncthing/blob/55147f5901ca8be3d6c06762bbec2f4fd6e42b82/build.go#L862) will be get commit sha for version, it's not allowed.
I think this is a bug(even if you can use the RELEASE file to skip), because it will cause developers to be unable to checkout the tag to compile the docker image.

.github/workflows/main.yml
calmh: I don't really want a nightly, so not pushing anything to docker hub, but I want the build steps in the repo to be exercised on commits to `main` and pull requests.

.github/workflows/main.yml
calmh: If we need a version name somewhere, the canonical would be to get it from `go run build.go version` so that it's the same as the binary inside the image.

.github/workflows/main.yml
qinghon: I think so too, but in fact images built using the main branch do not work properly.

I have tried the following, but this requires a allowed version or debug mode:
```yml
on:
  pull_request:
    branches: main
  push:
    branches: main
  push:
    tags:
      - v*
```

.github/workflows/main.yml
qinghon: In github action checkout command is:
```bash
git checkout --progress --force refs/tags/v1.8.0-rc.5
```
but `build.go` get a commit sha name(eg: 5c01c3d8), I had to use the RELEASE file can be work.

calmh:
Solved in a different manner with buildx on the build server. Thanks for showing the way!

PR #6845, opened 2020-07-22
lib/model, lib/scanner, lib/weakhash: Single-pass weakhash.Finder

greatroar:
Successor to #6692.

This version uses a new [rolling hash library](/greatroar/rolling) that includes the file scanning logic and allows access to the scanner's buffer, so the weakhash.Finder is completely gone. I've moved its test to lib/model.

The scanner uses the fast non-rolling Adler-32 in this new library, which gives a small throughput boost:

```
Validate-8    67.7ms ± 0%    65.8ms ± 0%  -2.81%  (p=0.000 n=9+10)
HashFile-8    87.7ms ± 0%    86.4ms ± 0%  -1.48%  (p=0.000 n=10+8)

Validate-8     0.00B          0.00B         ~     (all equal)
HashFile-8    45.4kB ± 0%    45.4kB ± 0%    ~     (p=0.904 n=10+10)

Validate-8      0.00           0.00         ~     (all equal)
HashFile-8      28.0 ± 0%      28.0 ± 0%    ~     (all equal)

HashFile-8   203MB/s ± 0%   206MB/s ± 0%  +1.51%  (p=0.000 n=10+8)
```

The weakhash benchmarks are gone. The new library has its own benchmarks, which show that it can scan almost 60MB/s on my machine (more on a newer one, I suppose).


imsodin:
I like the API and doing everything in a single pass. What makes me a bit concerned is that we move from an old, established one-person-show to another, new one-person-show. And move the scanning logic there. Which makes sense content-wise, but its still "our" code in that we need to debug it if there come up problems and that's not shared with someone else (unless you have more plans for it?). Ideally I'd have hoped @chmduquesne to react to your PR and get the changes in there. That ship has probably sailed though. One high-level question thus: Do you have any plans/expectations for that module outside of Syncthing and/or what's your motivation for creating a separate module?

chmduquesne:
I can have a look later if you want. One thing, though: if you had complaints about my lib, why not file a ticket instead of creating an entirely new lib? I would favor joining forces, it makes the project stronger... The more maintainers the merrier! Happy to discuss.

AudriusButkevicius:
Just an FYI, I am on holiday atm, but I plan to look at this in depth end of next week, given existing code I think was mostly my nonsense.

One comment, which perhaps other maintainers disagree with, is that we're adding quite a lot of dependencies on @greatroar repos in fairly sensitive areas, so I'm starting to be afraid we might blindly merge some dependabot PR bumping the dependency breaking the world for us, as the code is those projects is not reviewed by us

imsodin:
@chmduquesne As far as I understand the main requirement for the previous iteration of this change (without a new lib) was https://github.com/chmduquesne/rollinghash/pull/10

chmduquesne:
Wait, I had not noticed this PR! I am very sorry, I changed my email system recently and I have I missed a few notifications. I will review this ASAP.

chmduquesne:
It seems like the API is quite different from my library, so I am not sure if I can really support you.

In the PR that I received, it looks like you are trying to access the content of the rolling window. It is true that there is an internal copy of this window kept in the rollinghash. This copy is continuously updated with each call to Roll(), in a circular buffer. Its only purpose is to know the value of the byte exiting the window: the idea of rollinghash was to avoid maintaining this buffer externally, because I found it annoying having to write h.Roll(enter, exit) when at a constant memory cost I could just write h.Roll(enter).

If you absolutely insist, I can add a way to extract this window (and add the WriteTo method suggested in chmduquesne/rollinghash#10). However, I am not sure this is very useful in the current context, since I don't see WriteTo used anywhere.

chmduquesne:
By the way, I know that adler32 is popular because it is one of the first rolling hashes ever designed and it got traction thanks to rsync, but I would recommend switching to buzhash64 instead. It has less collisions for the same performance. If you are interested in knowing why, there is a little undocumented tool called roll (https://github.com/chmduquesne/rollinghash/blob/master/roll/main.go) which can be invoked with the option -stats. I wrote it to know how often I could expect a checksum beginning with a given number of 0s when hashing random data. If your checksum is well balanced, you should expect a checksum starting with n 0 should be expected every 2^n bytes. Buzshash64 works very well for that. In fact, every checksum in my package behaves well, *except* adler32. If you dig in the maths, it is obvious why and this behavior is expected, but then you should know that adler32 is really far from being the best rolling checksum.

calmh:
@greatroar echoing the others, and adding personally, it would bring greater peace of mind when adding a dependency on a repo of yours if you had a name, or an email address, or had _existed anywhere_ longer than a few months. I don't see any problems with your contributions, but I also see no track record on which to base an assumption that these packages will be maintained in any fashion in the future. We can discuss these things offline, this may not be the correct forum, but it's a shared concern that has been brought up in the thread.

greatroar:
I'll remove my lib and send patches to @chmduquesne instead.

PR #6847, opened 2020-07-22
lib/model: Check folder error before sync-waiting (fixes #6793)

### LINKED ISSUE #6793: Error status should take precedence over Local Additions
I have a Receive Only folder that is the mount path of an encfs volume. When I reboot my device, I don't automatically mount them so I expectedly get "folder path missing" error. However, when I have local changes, that's the status I see. I believe the failure status ("Stopped" if I remember correctly) should be displayed instead of "Local Additions".

Steps to reproduce:
- Create a Receive Only folder
- Make local changes and wait for "Local Additions" to be displayed in the web UI.
- Delete the folder completely
- Restart Syncthing
- Observe that the folder status still shows as "Local Additions" and the error "folder path missing" is displayed in the folder details.

Screen Shot 2020-06-24 at 1 26 02 PM

Syncthing v1.6.1
macOS 11.0 (Big Sur) Beta

imsodin:
There's two related elements here: 1. `folder.pull` should not `defer f.setState(FolderIdle)`. The actual underlying pull already takes care of folder state (either error or idle). 2. For scanning and (real) pulling we always check folder health/error before setting status. We don't do the same for sync-waiting. This means an existing error gets "half-cleared": State changes to sync-waiting, but error remains (as errors are only touched by `setError`). Thus when the actual puller calls `setError`, it sees the same error before and after and thus returns early without setting the folder state back to error. Regardless of this issue I believe it makes sense to check folder health before acquiring the iolimiter token to get the error notification to the user faster.

I don't entirely get the separation of `setState` and `setError` though. Given `setError` is called with `nil`, it seems like one shouldn't call `setState` when an error is set/"active", but that's not enforced. If that's not the case and `setState` is meant to override errors, it doesn't do that (no `s.err = nil`). To me it would feel simpler and more robust if `setState` did clear previous error states. Does anyone have any insights into why things work as they do at the moment?

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -334,15 +334,37 @@ func (f *folder) pull() (success bool) {
 		return true
 	}
 
+	// Abort early (before acquiring a token) if there's a folder error
+	err := f.getHealthErrorWithoutIgnores()
+	f.setError(err)
+	if err != nil {
+		l.Debugln("Skipping pull of", f.Description(), "due to folder error:", err)
+		return false
+	}
+
 	f.setState(FolderSyncWaiting)
-	defer f.setState(FolderIdle)
 
 	if err := f.ioLimiter.takeWithContext(f.ctx, 1); err != nil {
+		f.setError(err)
 		return true
 	}
 	defer f.ioLimiter.give(1)
 
 	startTime := time.Now()
+
+	// Check if the ignore patterns changed.
+	oldHash := f.ignores.Hash()
+	defer func() {
+		if f.ignores.Hash() != oldHash {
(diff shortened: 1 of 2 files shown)


calmh:
Probably just evolution. Maybe setState and setError should be combined. Either we are in a "doing something" or "waiting for something" state, when there implicitly can't be an error, or we are done with something and the outcome determines if we are now in an error state or not...

imsodin:
I agree it could and should be improved, however I'd anyway like to get this fix in the way things work currently going. So when it gets improved it start from a sounder base.

PR #6848, opened 2020-07-23
change default marker folder name

### LINKED ISSUE #6844: Rename .stfolder to something more obvious
I was recently cleaning up my file system and noticed a `.stfolder` in one of my directories. I erroneously assumed this was a folder created by the Spring Tools Suite IDE, so I removed the directory.  Syncthing eventually let me know that the folder in question was no longer in sync due to a missing file/folder (which seemed to be .stfolder).

I think it would be more useful and blatantly obvious if the folder in question was called `.syncthing` to .`syncthing-folder`. Then I would know exactly which application/process created the folder. I believe it's better to be explicit than implicit - abbreviations could belong to any number of tools.

saimonn:
### Purpose

Attempt to fix #6844 

### Testing

`./build.sh test` passes, but I'm not sure if it covers all the users cases (esp. with previously existing `.stfolder`)
I let untouched a few test data with `<markerName>.stfolder</markerName>` on purpose

### Documentation

### DIFF
--- a/cmd/stcompdirs/main.go
+++ b/cmd/stcompdirs/main.go
@@ -91,7 +91,7 @@ func startWalker(dir string, res chan<- fileInfo, abort <-chan struct{}) chan er
 		if rn == "." {
 			return nil
 		}
-		if rn == ".stversions" || rn == ".stfolder" {
+		if rn == ".stversions" || rn == ".syncthing" {
 			return filepath.SkipDir
 		}
 
--- a/lib/api/testdata/config/config.xml
+++ b/lib/api/testdata/config/config.xml
@@ -19,7 +19,7 @@
         <disableTempIndexes>false</disableTempIndexes>
         <paused>false</paused>
         <weakHashThresholdPct>25</weakHashThresholdPct>
-        <markerName>.stfolder</markerName>
+        <markerName>.syncthing</markerName>
         <useLargeBlocks>true</useLargeBlocks>
     </folder>
     <folder id="¯\_(ツ)_/¯ Räksmörgås 动作 Адрес" label="" path="s12-1/" type="sendreceive" rescanIntervalS="10" fsWatcherEnabled="false" fsWatcherDelayS="10" ignorePerms="false" autoNormalize="true">
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -460,7 +460,7 @@ func TestFolderCheckPath(t *testing.T) {
 	}
 	testFs := fs.NewFilesystem(fs.FilesystemTypeBasic, n)
 
-	err = os.MkdirAll(filepath.Join(n, "dir", ".stfolder"), os.FileMode(0777))
+	err = os.MkdirAll(filepath.Join(n, "dir", ".syncthing"), os.FileMode(0777))
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -26,7 +26,7 @@ var (
 	ErrMarkerMissing    = errors.New("folder marker missing (this indicates potential data loss, search docs/forum to get information about how to proceed)")
 )
 
-const DefaultMarkerName = ".stfolder"
+const DefaultMarkerName = ".syncthing"
 
 type FolderConfiguration struct {
 	ID                      string                      `xml:"id,attr" json:"id"`
--- a/lib/config/migrations.go
+++ b/lib/config/migrations.go
@@ -144,7 +144,7 @@ func migrateToConfigV23(cfg *Configuration) {
 		permBits = 0700
 	}
 
-	// Upgrade code remains hardcoded for .stfolder despite configurable
+	// Upgrade code remains hardcoded for .syncthing despite configurable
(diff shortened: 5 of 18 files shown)


uok:
I think `.syncthing` as @calmh suggested should be fine.
Does this also tackle existing folder markers? What if user has .stfolder in folder A and .syncthing in folder B?

saimonn:
I updated the commit to use `.syncthing`
@uok I guess, if folder A and B have a different marker folder name, at least one of them must be using `<markerName>` in its config, which overrides the default name.

imsodin:
There's no transition for existing folders as far as I see. I'd actually not do any transition, but instead keep accepting/looking for both `.stfolder` (e.g. as `deprecatedDefaultMarkerName`) and `.syncthing`. Maybe `.syncthing-marker` would make sense, to make the relation to the `markerName` setting more clear (and maybe as a "marker" less likely for people to just delete it).

calmh:
I didn't read the change, but I expect it to be small. The current default marker name `.stfolder` gets set into the folder config at creation time, so changing the default doesn't affect any existing folders. I would expect that we can essentially just change the constant in two places and leave it at that.

AudriusButkevicius:
I think the change is fine, but you need a migration, also some of the files you are modifying are auto-generated.

imsodin:
We need to get to a consensus here, right now there's three maintainers saying three at least somewhat different things xD

What does it need a migration for? As calmh wrote it should be ok to just adjust the default marker name, existing folders will stay with ".stfolder" on disk and in config (which seems like the right thing to happen). Then there's a few places where those filenames are hardcoded (e.g. in `fs.IsInternal`), where I still think we should keep recognizing both .stfolder and the new marker name.

calmh:
I think changing the default for the future is fine. Migrating existing folders is unnecessary and will cause trouble. We don't know the cases where .stfolder exists but is read only and not created by us, or exists but is explicitly ignored by some other sync / cleanup tool, etc.

AudriusButkevicius:
Ok then this change needs to handle accepting and ignoring both

man/syncthing-config.5
```diff
@@ -98,7 +98,7 @@ The following shows an example of the default configuration file (IDs will diffe
         <disableTempIndexes>false</disableTempIndexes>
         <paused>false</paused>
         <weakHashThresholdPct>25</weakHashThresholdPct>
-        <markerName>.stfolder</markerName>
+        <markerName>.syncthing</markerName>
```
imsodin, line 101: Man pages are auto-generated from docs, i.e. these man page changes should happen in https://github.com/syncthing/docs instead.

AudriusButkevicius:
So I think this still requires a migration?

imsodin:
Why? Existing folders will keep using `.stfolder` as a marker. As Jakob wrote that's good in cases hich I believe for any custom setups involving that dir (e.g. other tools ignoring .stfolder) and/or read-only `.stfolder`. All new folders will get `.syncthing`. In the one place where we use hardcoded values (`fs.IsInternal`) we look for both (that function should probably be in lib/scanner anyway, as that's the only place it's used and there we could use the config value).

AudriusButkevicius:
Sure, we should attempt to rename it, if that fails, we keep checking .stfolder, but I don't see why we should encourage to leave it a half-way house.

The file is internal to us, so any tools relying on it shouldn't and will break with new folders anyway, but this is the price of doing this.

imsodin:
I don't think there's any clear winner between the two options. I agree with the half-way sentiment, and I think the arguments for "phase it over slowly without migration" are also valid. And in addition I don't think it matters much and thus have not much interest in discussing this to depth trying to reach a consensus. I am fine with either way, if I have to choose I'll choose firstly whatever the one implementing this wants and secondly no migration.

AudriusButkevicius:
So my concern is with leaving it at `.stfolder` is that if we ever decide to move ignores from the folder root, then suddently we'd have to add ugly handling for that, so I am just thinking ahead and think that we should do a migration.

To be honest, for future-proofing sake, I'd probably error the folder if the migration failed, atleast that gets rid of having to handle edge conditions in the future.

imsodin:
The folder marker name is configurable (`MarkerName`). If we move stuff into the folder marker, will move it into that custom name regardless of whether we do a transition now for `.stfolder` or not.

AudriusButkevicius:
I think these are separate usecases. The usecase when you want a custom marker name is when the filesystem is read-only, or you genuinely can't afford to have additional structures/files in the directory.

In those cases you won't have ignore patterns anyways because you don't want that extra file for the same reasons you don't want the default marker.

I guess you could point it at a custom marker for the purpose of not having to have the default marker, but at the point you create ignores you'll have extra files anyway, so you might as well switch back to the default marker (if the ignores happen to be inside of the marker).

ProactiveServices:
I wouldn't mind it being migrated and breaking external things because software that's still 0.x is still in the "break stuff as it evolves" stage, IMO.

AudriusButkevicius:
We'll we're not 0.x anymore, but again, this is a file internal to us, I can't really see anyone being dependant on it's name with the exception of android, which is in our hands to fix as we break this.

st-review:
:robot: *beep boop*

PR #6849, opened 2020-07-23
lib: Wrap suture to add service errors/termination (ref #5907)

imsodin:
This is a proposal to extend/wrap suture service to return an error and thus allow a service to signal that it's irrecoverably broken and thus the whole app should stop. It's essential variant 2. in https://forum.syncthing.net/t/shutdown-on-error/14908

The basic idea is pretty simple: Our service have `Serve() error` and when adding such a service to our supervisor, it wraps that handling the error and exposing just `Serve()` to suture. Thus the changes are minimal to how things work now, while the diff will obviously still be humongous. That's why I just implemented the wrapping and lib/api, to show how `FatalErr` replaces the controller object. Next would be to replace all sutures with our wrapping and after that we can make lib/db return errors instead of panicking, and the model then turns that into a `FatalErr`. However as said that's going to be lots of legwork so I'd like to do the conceptual review now and if successful do the rest and get it merge superfast to keep merge conflicts at bay. So: What do you think of this?

### DIFF
--- a/lib/api/api.go
+++ b/lib/api/api.go
@@ -32,7 +32,6 @@ import (
 	"time"
 
 	metrics "github.com/rcrowley/go-metrics"
-	"github.com/thejerf/suture"
 	"github.com/vitrun/qart/qr"
 	"golang.org/x/crypto/bcrypt"
 
@@ -49,11 +48,11 @@ import (
 	"github.com/syncthing/syncthing/lib/model"
 	"github.com/syncthing/syncthing/lib/protocol"
 	"github.com/syncthing/syncthing/lib/rand"
+	"github.com/syncthing/syncthing/lib/suturewrap"
 	"github.com/syncthing/syncthing/lib/sync"
 	"github.com/syncthing/syncthing/lib/tlsutil"
 	"github.com/syncthing/syncthing/lib/upgrade"
 	"github.com/syncthing/syncthing/lib/ur"
-	"github.com/syncthing/syncthing/lib/util"
 )
 
 // matches a bcrypt hash and not too much else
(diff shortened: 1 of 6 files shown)


lib/api/api.go
```diff
@@ -49,11 +48,11 @@ import (
 	"github.com/syncthing/syncthing/lib/model"
 	"github.com/syncthing/syncthing/lib/protocol"
 	"github.com/syncthing/syncthing/lib/rand"
+	"github.com/syncthing/syncthing/lib/suturewrap"
```
AudriusButkevicius, line 51: lib/service would be bettter

lib/suturewrap/suturewrap.go
```diff
@@ -0,0 +1,240 @@
...
+	"fmt"
+	stdsync "sync"
+	"time"
+
+	"github.com/syncthing/syncthing/lib/sync"
+
+	"github.com/thejerf/suture"
+)
+
+type FatalErr struct {
```
AudriusButkevicius, line 21: All of the AsService cruft should be here too

imsodin:
I'd very much appreciate a comment on whether you think this goes into the right direction or not. I really want to do this and start implementing sane error handling at the db level. I could already start with the sane error handling, but that means panicking in the code calling the db in all places that doesn't yet have sane handling respectively where a db error is truly fatal, which is even uglier than the status quo.

AudriusButkevicius:
Obviously, ideally we'd want this to be upstream, but given prior conversations I think it's probably right for us to use our own wrapper that suits our purpose.

I think it's fine for us to extend the API given we have no better choice, so this goes in the right direction, yet I do have nitpicks on the implementation.

imsodin:
Actually let me just drop a note upstream, asking if the concept would be something they could imagine incorporating it.

PR #6857, opened 2020-07-27
Remove unneeded dependencies

emad-elsaid:
### Purpose

After executing `go mod tidy` I found that some dependencies were not needed.

### Testing

running `./build.sh` didn't raise any errors

## Authorship

based on the commit metadata.


calmh:
Hi! go.sum does accumulate some cruft during upgrades, and does get periodically tidied. Having it around for a while doesn't hurt so I don't think we need separate PRs to clean it up, better for it to happen "organically" during normal dependency changes. Thanks for the thought!

emad-elsaid:
Thanks for the fast response

Issue #6855, opened 2020-07-26
panic: device present in global list but missing as device/fileinfo entry when scanning folder


HackaN:
I should begin with saying that this might be a duplicate of #6457, but that report didn't educate me much.

I am running syncthing on Ubuntu 20.04 and parallel to a system update (before actually performing the update) I got an error from Syncthing GTK saying that the "daemon exited too fast" and wanted me to provide a path for daemon binary.

I ran syncthing from terminal and I get several panic messages. First I got a panic about Config file version (31) that was newer than the supported version (28). I followed an existing error report from the Synthing forum and switched out the Ubuntu repo version for the syncthing's own repo, uninstalled/installed and rebooted (just to be sure changes would take effect) and am now faced with a new set of errors:

[start] 18:02:42 INFO: syncthing v1.7.1 "Fermium Flea" (go1.14.4 linux-amd64) deb@build.syncthing.net 2020-07-11 18:17:41 UTC
[**CLMWS**] 18:02:45 INFO: Completed initial scan of receiveonly folder "hackan Pixel Kamera" (pixel_3xdn foton)
panic: device present in global list but missing as device/fileinfo entry
[**monitor**] 18:02:45 WARNING: Panic detected, writing to "/home/hackan/.config/syncthing/panic-20200726-180245.log"
[monitor] 18:02:45 WARNING: Please check for existing issues with similar panic message at https://github.com/syncthing/syncthing/issues/
[monitor] 18:02:45 WARNING: If no issue with similar panic message exists, please create a new issue with the panic log attached
[monitor] 18:02:45 INFO: Reporting crash found in panic-20200726-180245.log (report ID bfe3c0d1) ...
[**monitor**] 18:02:45 INFO: Syncthing exited: exit status 2
[monitor] 18:02:46 WARNING: 4 restarts in 16.671957674s; not retrying further

I have seen suggestions about removing the index-folder and sort of starting over from scratch, but I have a large library and want to try to figure this out before attempting that as a solution... if that even would be a solution.

Let me know what other information I can provide.

calmh:
The actual panic information is in the mentioned file. It's apparently been auto reported so in theory we're already aware, but to get an answer to what's going on please attach or paste at least the first 100 lines or so from that panic log here.

HackaN:
[](https://cloud.woelkli.com/s/TWcFiE7L3fZ7Nqz)
Password: @?hHo-=67|L

HackaN:
[...]

PR #6861, opened 2020-07-28
lib/db: Include blocks in db check (ref #6855)

imsodin:
This is not a fix for #6855 or anything else, but given missing blocks are a thing, I don't see why we should do all the expensive db checking but omit blocklists from it.


Catfriend1:
Hi @imsodin , 
can I run this straight off the PR's teamcity build to get more diagnose info what's wrong with my cluster? I expect the panic to happen again as soon as Syncthing v1.8.0-rc.3 finishes the rescan on the node where it crashed constantly before.

PR #6863, opened 2020-07-28
lib/db: Don't count invalid items to global state (ref #6850)

### LINKED ISSUE #6850: Global state incorrect when ignoring local change
Steps to repro:

1. Create receive-only folder
2. Add some file, which is then displayed as a local change.
3. Add an ignore pattern matching the added file.

Result: Local state zero, global state still 1.

Expected: Both 0.

imsodin:
### Purpose

https://github.com/syncthing/syncthing/issues/6850#issuecomment-665031731:  
> [...] why do we include receive-only changed items in the global count in the first place? That seems completely wrong to me: The global state is what everyone agrees upon. Receive-only changed items are the complete antithesis to that. If we didn't add it to the global state, receive-only items would lead to a higher local than global state, which seems right. It also removes the need for the workaround added in #6852. Thus I propose to not add invalid items to the global state. 

This behaviour also fixes #6850, and in addition to the previous fix (which is reverted here) covers removing items from the global state as well.

### DIFF
--- a/lib/db/meta.go
+++ b/lib/db/meta.go
@@ -173,22 +173,27 @@ func (m *metadataTracker) addFile(dev protocol.DeviceID, f protocol.FileIntf) {
 	m.mut.Lock()
 	defer m.mut.Unlock()
 
-	m.dirty = true
-
 	m.updateSeqLocked(dev, f)
 
-	if f.IsInvalid() && f.FileLocalFlags() == 0 {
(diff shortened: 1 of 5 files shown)


calmh:
There's been a lot of mental back and forth on having invalid stuff in the global or not. I can't help but feel we might be overlooking some corner case here. But I guess we will see what it is when it happens.

PR #6868, opened 2020-07-30
lib/connections, lib/tlsutil: Handle certName in Go 1.15 (fixes #6867)

### LINKED ISSUE #6867: Certificate error when built with Go 1.15
When built with Go 1.15, every connection attempt results in 

```
[SYNO4] 10:08:38 WARNING: Bad certificate from ... at 172.16.33.3:22001-172.16.33.8:59376/tcp-server/TLS1.3-TLS_AES_128_GCM_SHA256:
   x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0
```

We should generate certificates with a SAN instead/in addition to the common name. However it would be great if we can connect without forcing new device IDs on everyone...

Setting the env var makes things work. However we would ideally not do that as it affects all other TLS connections as well.

calmh:
Our authentication is based on device ID (certificate fingerprint) but
we also check the certificate name for ... historical extra security
reasons. (I don't think this adds anything but it is what it is.) Since
that check breaks in Go 1.15 this change does two things:

- Adds a manual check for the peer certificate CommonName, and if they
  are equal we are happy and don't call the more advanced
  VerifyHostname() function. This allows our old style certificates to
  still pass the check.

- Adds the cert name "syncthing" as a DNS SAN when generating the
  certificate. This is the correct way nowadays and makes VerifyHostname()
  happy in Go 1.15 as well, even without the above patch.


lib/api/api.go
```diff
@@ -1752,12 +1752,21 @@ func checkExpiry(cert tls.Certificate) error {
...
+	if len(leaf.DNSNames) > 1 {
+		// The certificate has more DNS SANs attributes than we ever add, so
+		// we leave it alone.
+		return nil
+	}
+	if len(leaf.DNSNames) == 1 && leaf.DNSNames[0] != leaf.Issuer.CommonName {
+		// The one SAN is different from the issuer, so it's not one of our
+		// newer self signed certificates.
+		return nil
+	}
```
AudriusButkevicius, line 1769: Seems weird that checkExpiry function doesn't check expiry in so many edge cases, and returns no error... Defo feels deceptive...

calmh: Well it's a very special-purpose function, only used to check expiry for the specific purpose of whether we should regenerate the certificate. Possibly it's badly named, but nil in all these cases just means "leave the certificate alone, it's either fine or not our business to recreate it".

PR #6869, opened 2020-07-30
build: Update packages, now supports freebsd-arm64 builds

calmh:


calmh:
... except with this gopsutil is now broken for freebsd-arm instead. Looking at the issues on that repo it seems like a never ending saga of changing types. 🙄

PR #6872, opened 2020-08-01
lib/db: Log context on panic

imsodin:
Right now the only thing we know on panicking is that it panics, not on what. The function is clear from the traces, but not the file it panics on. Maybe that would show a pattern. May become obsolete by the next release depending on how fast the suture error handling and the subsequent db error handling comes to pass (unless there's another reason for an RC bump, then I'd propose also adding this commit (which I don't hope happens)).

### DIFF
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -13,6 +13,8 @@
 package db
 
 import (
+	"fmt"
+
 	"github.com/syncthing/syncthing/lib/db/backend"
 	"github.com/syncthing/syncthing/lib/fs"
 	"github.com/syncthing/syncthing/lib/osutil"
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
So if we panic, we have the stacktrace. What's the value in the warn before the panic?

imsodin:
The function arguments (file, folder, device, ...).

AudriusButkevicius:
If we think that is important, why we are not including that in the panic message itself?

imsodin:
Because we don't want personal info in the panic messages, as they get reported (there was even quite some cleanup on that part a while ago). This is just for debugging/supporting when a user asks about the panic. It's also not like I know it's important, but there's some chance that the file involved is somehow "special" or changed in some weird way or whatever.

Issue #6875, opened 2020-08-03
Copy range method "all" prevents sync


calmh:
Tested setting the "all" method on a folder on Mac to see what happens. I'm not sure if any of our advanced methods are supported, but then I'd expect a fallback to standard? Actual results:

```
[MRIW7] 09:36:21 INFO: Puller (folder default, item "whatever"): dst write: read /Users/jb/dev/syncthing/syncthing/test/s2/whatever: file already closed
[MRIW7] 09:36:21 INFO: default: Failed to sync 1 items
[MRIW7] 09:36:21 INFO: Folder default isn't making sync progress - retrying in 1m0s.
[MRIW7] 09:37:22 INFO: default: Failed to sync 1 items
[MRIW7] 09:37:22 INFO: Folder default isn't making sync progress - retrying in 1m0s.
[MRIW7] 09:38:22 INFO: default: Failed to sync 1 items
[MRIW7] 09:38:22 INFO: Folder default isn't making sync progress - retrying in 2m0s.
...
```

calmh:
Also on FreeBSD backed by ZFS with "all"

```
[start] 09:48:23 INFO: syncthing v1.8.0 "Fermium Flea" (go1.14.6 freebsd-amd64) teamcity@build.syncthing.net 2020-07-28 10:09:00 UTC
...
[MRIW7] 09:52:58 INFO: Puller (folder default, item "big"): dst write: read /home/jb/syncthing/test/s2/big: file already closed
[MRIW7] 09:52:58 INFO: default: Failed to sync 1 items
[MRIW7] 09:52:58 INFO: Folder default isn't making sync progress - retrying in 1m0s.
```

calmh:
Also linux 4.15.0 on ext4

```
[start] 10:10:15 INFO: syncthing v1.8.0 "Fermium Flea" (go1.14.6 linux-amd64) teamcity@build.syncthing.net 2020-07-28 10:09:00 UTC
...
[MRIW7] 10:11:40 INFO: Puller (folder default, item "big"): dst write: read /home/jb/src/syncthing/test/s2/big: file already closed
```

PR #6877, opened 2020-08-03
lib/fs: Unwrap mtimeFile, get fd the "correct" way (ref #6875)

AudriusButkevicius:
So the crux of the problem is the fact they are wrapped in mtimefs.

I thought about doing an interface assertion instead, but some of the implementations call Stat() which again might return basicFileInfo or mtimeFileInfo, so I am not sure I can  do an assertion on a function with yet another generic interface return value?

This also addresses stuff discussed in golang/go#24331

Let's see if this passes the tests.


calmh:
I fixed the freebsd builder so it's happy now.

We could have an unexported `interface { unwrap() FileInfo }` that FileInfo implementations can implement if they wrap a lower level implementation. Then we'd loop on `unwrap()` as long as the return is still assertable to that interface and finally check if it's basic or not.

It's not just the mtime one, there's also the new windows junction stuff, but maybe that only wraps junctions specifically?

calmh:
Or possibly we implement the SyscallFd stuff

AudriusButkevicius:
image

AudriusButkevicius:
image

AudriusButkevicius:
image

PR #6878, opened 2020-08-05
Update notifications.html

aliatiia:
small typo


AudriusButkevicius:
I don't think this is a typo, through in this context means "via" effectively.

calmh:
Yeah. Potentially it would be clearer with more commas but given translation efforts it is what is I think

PR #6883, opened 2020-08-08
all: Remove need to restart syncthing

AudriusButkevicius:
This removes the need to restart syncthing in almost all cases.

The only two left are:
1. Crash reporting - happens in the monitor (and monitor only re-reads the option on restart of the child process)
2. Database tuning - everything depends on this, so we pretty much have to tear down the whole application anyway.

### DIFF
--- a/lib/config/optionsconfiguration.go
+++ b/lib/config/optionsconfiguration.go
@@ -40,7 +40,7 @@ type OptionsConfiguration struct {
 	URInitialDelayS         int      `xml:"urInitialDelayS" json:"urInitialDelayS" default:"1800"`
 	RestartOnWakeup         bool     `xml:"restartOnWakeup" json:"restartOnWakeup" default:"true" restart:"true"`
 	AutoUpgradeIntervalH    int      `xml:"autoUpgradeIntervalH" json:"autoUpgradeIntervalH" default:"12" restart:"true"` // 0 for off
-	UpgradeToPreReleases    bool     `xml:"upgradeToPreReleases" json:"upgradeToPreReleases" restart:"true"`              // when auto upgrades are enabled
+	UpgradeToPreReleases    bool     `xml:"upgradeToPreReleases" json:"upgradeToPreReleases"`              // when auto upgrades are enabled
 	KeepTemporariesH        int      `xml:"keepTemporariesH" json:"keepTemporariesH" default:"24"`                        // 0 for off
 	CacheIgnoredFiles       bool     `xml:"cacheIgnoredFiles" json:"cacheIgnoredFiles" default:"false"`
 	ProgressUpdateIntervalS int      `xml:"progressUpdateIntervalS" json:"progressUpdateIntervalS" default:"5"`
--- a/lib/config/optionsconfiguration.go
+++ b/lib/config/optionsconfiguration.go
@@ -200,3 +200,7 @@ func (opts OptionsConfiguration) MaxConcurrentIncomingRequestKiB() int {
 	// Roll with it.
 	return opts.RawMaxCIRequestKiB
 }
+
+func (opts OptionsConfiguration) AutoUpgradeEnabled() bool {
+	return opts.AutoUpgradeIntervalH > 0
+}
--- a/lib/config/optionsconfiguration.go
+++ b/lib/config/optionsconfiguration.go
@@ -41,7 +41,7 @@ type OptionsConfiguration struct {
 	RestartOnWakeup         bool     `xml:"restartOnWakeup" json:"restartOnWakeup" default:"true"`
 	AutoUpgradeIntervalH    int      `xml:"autoUpgradeIntervalH" json:"autoUpgradeIntervalH" default:"12"` // 0 for off
 	UpgradeToPreReleases    bool     `xml:"upgradeToPreReleases" json:"upgradeToPreReleases"`              // when auto upgrades are enabled
-	KeepTemporariesH        int      `xml:"keepTemporariesH" json:"keepTemporariesH" default:"24"`                        // 0 for off
+	KeepTemporariesH        int      `xml:"keepTemporariesH" json:"keepTemporariesH" default:"24"`         // 0 for off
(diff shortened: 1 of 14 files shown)


AudriusButkevicius:
Actually, crash reporting does not require restarts after all, as it's checked on every attempt to report them, and the attempts only happen on restarts.

lib/discover/global.go
```diff
@@ -448,3 +449,15 @@ func (c *contextClient) Post(ctx context.Context, url, ctype string, data io.Rea
 	req.Header.Set("Content-Type", ctype)
 	return c.Client.Do(req)
 }
+
+func globalDiscoveryIdentity(addr string) string {
+	return "global discovery server " + addr
+}
+
+func ipv4DiscoveryIdentity(port int) string {
+	return fmt.Sprintf("IPv4 local broadcast discovery on port %d", port)
```
imsodin, line 458: Felt a bit weird at first to have such a nice human readable string as an identifier, then again it's actually nice :)  
I'd shorten it to `ipv4Identity` and so on.

lib/discover/manager.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/syncthing/syncthing/lib/sync"
+	"github.com/syncthing/syncthing/lib/util"
+)
+
+// The Manager aggregates results from multiple Finders. Each Finder has
+// an associated cache time and negative cache time. The cache time sets how
+// long we cache and return successful lookup results, the negative cache
+// time sets how long we refrain from asking about the same device ID after
+// receiving a negative answer. The value of zero disables caching (positive
+// or negative).
```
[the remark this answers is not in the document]
AudriusButkevicius: To be honest, I think all of this is an explanation how it works.

lib/discover/global.go
AudriusButkevicius: So there are 2 identities now, one is printed in the logs on startup, which is verbose, another one which is shown in the UI for errors (which we don't want to be long).

The one in the UI just says "IPv4 local" which doesn't distinguish the port, so is of no use to me for this purpose.
I guess I could change that, but I can't think of a way of keeping it short, yet still human readable, hence why we have 2 now.

The global discovery server names in the UI is just `global@http:/blabla` I guess we could have `local-ipv4@:21027`, but these things are not very friendly for the logs.

AudriusButkevicius:
I think the connection.NewService suggestion makes things worse, and breaks the nice dependency injection we have going on.

I'd be unable to supply my own implementation of discover.Finder to it anymore. I have an alternative approach, which is a wrapper around addressLister, which initially has no implementation, and then I later set the implementation once connection service is created.
 
This keeps the component APIs clean, and has a bit on ugly code in the wiring.

lib/config/optionsconfiguration.go
```diff
@@ -200,7 +200,3 @@ func (opts OptionsConfiguration) MaxConcurrentIncomingRequestKiB() int {
 	// Roll with it.
 	return opts.RawMaxCIRequestKiB
 }
-
-func (opts OptionsConfiguration) ShouldAutoUpgrade() bool {
-	return opts.AutoUpgradeIntervalH > 0
-}
```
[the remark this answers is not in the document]
AudriusButkevicius: I gave it a different name

lib/discover/global.go
imsodin: I missed that it was also used for logging, thought it was just a map key. Anyway sounds all good - I agree exposing more info like ports is not that valuable and potentially confusing (at the very least more complex).

lib/discover/manager.go
```diff
@@ -0,0 +1,297 @@
...
+			gd, err := NewGlobal(srv, m.cert, m.addressLister, m.evLogger)
+			if err != nil {
+				l.Warnln("Global discovery:", err)
+				continue
+			}
+
+			// Each global discovery server gets its results cached for five
+			// minutes, and is not asked again for a minute when it's returned
+			// unsuccessfully.
+			m.addLocked(identity, gd, 5*time.Minute, time.Minute)
```
imsodin, line 268: Give them magical values names please.

imsodin:
I guess we are getting a bit off-track here, given it's about not having to restart. And likely the issues have existed before this PR, the s/CachingMux/Manager/ change just prompted me to look at it - so feel free to skip any of this.  
If this were about a nice refactor and we want to keep dependency injection, I'd separate the discovery announcement and querying to remove the co-dependency with the connection service. And on separation of concerns it's weird that local discovery does it's own caching, and for global discovery the manager handles caching. I'd probably not do any caching on the finders and let the manager handle all of it.

AudriusButkevicius:
So I am not going down that rabbit hole in this PR. My goal is to remove the need to restart with as minimal code changes as possible.

If we want to refactor discovery, split stuff, give magic values constant names, I am fine with that, but not in this PR.

lib/syncthing/syncthing.go
```diff
@@ -279,43 +274,20 @@ func (a *App) startup() error {
...
 
-	// Start connection management
+	// Start discovery and connection management
 
-	connectionsService := connections.NewService(a.cfg, a.myID, m, tlsCfg, cachedDiscovery, bepProtocolName, tlsDefaultCommonName, a.evLogger)
-	a.mainService.Add(connectionsService)
+	// Chicken and egg, discovery manager depends on connection service to tell it what addresses it's listening on
+	// Connection service depends on discovery manager to get addresses to connect to.
+	// Create a wrapper that is then wired after they are both setup.
+	addrLister := lateAddressLister{}
```
imsodin, line 282:
```suggestion
	addrLister := &lateAddressLister{}
```

Catfriend1:
Hi,

can changes to "urSeen" / "urAccepted" also be applied without restarting Syncting after this PR got merged? The Android app currently has a restart of the SyncthingNative in place there when the user answered the "privacy" dialog.

AudriusButkevicius:
I don't think those required restart ever.

PR #1234, opened 2015-01-10
Small improvements for footer navbar (ref #1154)

uok:
Small improvements for footer navbar (ref #1154)
- add icons
- link to Syncthing wiki
- show footer navbar on small devices (non-fixed)

Issue #6527, opened 2020-04-13
Clean up after days feature for Simple File Versioning


jolange:
Hi!

Thanks for this nice tool, I like it a lot :-)

I searched the forums and issue tracker, but could not find any discussion on this:
I would like to have a "clean" feature for *Simple File Versioning*. I think something similar to the feature for [Trash Can File Versioning](https://docs.syncthing.net/users/versioning.html#trash-can-file-versioning) would be useful:

> A configuration option is available to clean the trash can from files older than a specified number of days. If this is set to a positive number of days, files will be removed when they have been in the trash can that long. Setting this to zero prevents any files from being removed from the trash can automatically.

Is there any reason to not have this for *Simple File Versioning*? I.e., remove versioned files from `.stdversions` older than XY days?

If you think this would be a desired feature, I could try to implement this (although I'm not really familiar with go, so I might take some time).

AudriusButkevicius:
I think if you want it time based, use trash can if you want it count based use simple, then there is also staggered which is the mix of the two, hence I think you just need to pick a different versioner.

jolange:
Thanks for your answer!

I don't quite agree with the characterization of the *Trash Can* approach as time based per se. I think it is the same as *Simple File Versioning* with just one versioned copy. Only the optional cleaning makes it time based and I think this would also be reasonable for *Simple File Versioning*.

Maybe let me explain my use case a bit more: I'd like to have several versions of deleted/modified files for the case that there are several modifications (in a short time) that I'd like to undo. On the other hand, I'd like to have cleaning after some time, e.g., after a month, because I will probably no longer need the old files. The *Staggered File Versioning* approach potentially creates too many versions for me.

So, *Simple File Versioning* is just fine for me. But I'd like to save storage space and not backup things forever. For example, if I delete a file (intentionally, because I really don't need it any longer), the file(s) in `.stdversions` would never get deleted at the moment. Of course I can do some manual cleaning from time to time by finding files older than XY days. But since there already is this feature for the *Trash Can* approach, I feel it would be straight-forward to add this feature here.

AudriusButkevicius:
I don't think the request is crazy, but my gut says that if the user wants something else than what is there, they should use external versioner and devise their own strategy. 

We will never have all the options to cater for everyone so I am inclined to push back unless you plan to contribute the code for this.

In this specific case I think its easier to setup a cron on the side with `find foo/.stversions -type f -mtime +30 -delete` (fake example check docs).

[...]

PR #6884, opened 2020-08-08
Created cleanup functionality for syncthing

rjpruitt16:
### Purpose

Describe the purpose of this change. If there is an existing issue that is
resolved by this pull request, ensure that the commit subject is on the form
`Some short description (fixes #1234)` where 1234 is the issue number.
https://github.com/syncthing/syncthing/issues/6527
Adding the trash can feature of deleting files after certain amount days to Simple File folder type.

### Testing

Describe what testing has been done, and how the reviewer can test the change
if new tests are not included.

I copied the test for trash can feature into simple feature. I did not write any test for gui.

### Screenshots

If this is a GUI change, include screenshots of the change. If not, please
feel free to just delete this section.

a picture of inputs for simple file feature

### Documentation

https://docs.syncthing.net/users/versioning.html#simple-file-versioning

Under simple file, it should probably be updated to say there is a feature for deleting folders automatically.

## Authorship

based on the commit metadata.

### DIFF
--- a/gui/default/syncthing/folder/editFolderModalView.html
+++ b/gui/default/syncthing/folder/editFolderModalView.html
@@ -2,100 +2,62 @@
   <div class="modal-body">
     <form role="form" name="folderEditor">
       <ul class="nav nav-tabs" ng-init="loadFormIntoScope(folderEditor)">
-        <li class="active"><a data-toggle="tab" href="#folder-general"><span class="fas fa-cog"></span> <span
-              translate>General</span></a></li>
-        <li><a data-toggle="tab" href="#folder-sharing"><span class="fas fa-share-alt"></span> <span
-              translate>Sharing</span></a></li>
-        <li><a data-toggle="tab" href="#folder-versioning"><span class="fas fa-copy"></span> <span translate>File
-              Versioning</span></a></li>
-        <li><a data-toggle="tab" href="#folder-ignores"><span class="fas fa-filter"></span> <span translate>Ignore
-              Patterns</span></a></li>
-        <li><a data-toggle="tab" href="#folder-advanced"><span class="fas fa-cogs"></span> <span
-              translate>Advanced</span></a></li>
+        <li class="active"><a data-toggle="tab" href="#folder-general"><span class="fas fa-cog"></span> <span translate>General</span></a></li>
+        <li><a data-toggle="tab" href="#folder-sharing"><span class="fas fa-share-alt"></span> <span translate>Sharing</span></a></li>
+        <li><a data-toggle="tab" href="#folder-versioning"><span class="fas fa-copy"></span> <span translate>File Versioning</span></a></li>
+        <li><a data-toggle="tab" href="#folder-ignores"><span class="fas fa-filter"></span> <span translate>Ignore Patterns</span></a></li>
+        <li><a data-toggle="tab" href="#folder-advanced"><span class="fas fa-cogs"></span> <span translate>Advanced</span></a></li>
       </ul>
       <div class="tab-content">
 
         <div id="folder-general" class="tab-pane in active">
-          <div class="form-group"
-            ng-class="{'has-error': folderEditor.folderLabel.$invalid && folderEditor.folderLabel.$dirty}">
+          <div class="form-group" ng-class="{'has-error': folderEditor.folderLabel.$invalid && folderEditor.folderLabel.$dirty}">
             <label for="folderLabel"><span translate>Folder Label</span></label>
-            <input name="folderLabel" id="folderLabel" class="form-control" type="text" ng-model="currentFolder.label"
-              value="{{currentFolder.label}}" />
+            <input name="folderLabel" id="folderLabel" class="form-control" type="text" ng-model="currentFolder.label" value="{{currentFolder.label}}" />
             <p class="help-block">
-              <span translate ng-if="folderEditor.folderLabel.$valid || folderEditor.folderLabel.$pristine">Optional
-                descriptive label for the folder. Can be different on each device.</span>
+              <span translate ng-if="folderEditor.folderLabel.$valid || folderEditor.folderLabel.$pristine">Optional descriptive label for the folder. Can be different on each device.</span>
             </p>
           </div>
-          <div class="form-group"
-            ng-class="{'has-error': folderEditor.folderID.$invalid && folderEditor.folderID.$dirty}">
+          <div class="form-group" ng-class="{'has-error': folderEditor.folderID.$invalid && folderEditor.folderID.$dirty}">
             <label for="folderID"><span translate>Folder ID</span></label>
-            <input name="folderID"
-              ng-readonly="editingExisting || (!editingExisting && currentFolder.viewFlags.importFromOtherDevice)"
-              id="folderID" class="form-control" type="text" ng-model="currentFolder.id" required=""
-              aria-required="true" unique-folder value="{{currentFolder.id}}" />
+            <input name="folderID" ng-readonly="editingExisting || (!editingExisting && currentFolder.viewFlags.importFromOtherDevice)" id="folderID" class="form-control" type="text" ng-model="currentFolder.id" required="" aria-required="true" unique-folder value="{{currentFolder.id}}" />
             <p class="help-block">
-              <span translate ng-if="folderEditor.folderID.$valid || folderEditor.folderID.$pristine">Required
-                identifier for the folder. Must be the same on all cluster devices.</span>
+              <span translate ng-if="folderEditor.folderID.$valid || folderEditor.folderID.$pristine">Required identifier for the folder. Must be the same on all cluster devices.</span>
               <span translate ng-if="folderEditor.folderID.$error.uniqueFolder">The folder ID must be unique.</span>
-              <span translate ng-if="folderEditor.folderID.$error.required && folderEditor.folderID.$dirty">The folder
-                ID cannot be blank.</span>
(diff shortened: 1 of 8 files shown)


imsodin:
Thanks for the contribution!

You need to revert the style change on the JS file. And in the html please follow the existing whitespace conventions (space around operators).

I also seem to remember some general argument against adding this to the simple file versioner, though I can't remember (or come up with something myself).

> I copied the test for trash can feature into simple feature. I did not write any test for gui.

Did you have to alter it significantly? Otherwise it's probably better to make a test function taking the versioner type as an argument to dedup (maybe also for the cleanup implementation? I didn't check myself yet at all).

rjpruitt16:
Is there a style guide for html and js I can see. I want to make sure I understand. Also, Are you saying adda fle call versioner_test.go and maybe write a table test for simple and trashcan versioning for DRY principles. I did not have to alter the test at all. It was the was the perfect commit for a go noob ;)

calmh:
There is no style guide beyond the surrounding code itself, unfortunately. 

Yeah if the test is unchanged it could be refactored to a function that takes a versioner and then call that with the different versions using t.Run for example.

rjpruitt16:
lol i see why you said revert the changes to js. I use an auto formatter. Sorry.

lib/versioner/simple_test.go
```diff
@@ -7,13 +7,14 @@
...
 
 import (
-	"github.com/syncthing/syncthing/lib/config"
 	"io/ioutil"
 	"math"
 	"path/filepath"
 	"testing"
 	"time"
 
+	"github.com/syncthing/syncthing/lib/config"
```
AudriusButkevicius, line 16: Surprised why this is needed. Nothing new in this file?

lib/versioner/simple.go
```diff
@@ -75,6 +80,50 @@ func (v simple) Restore(filepath string, versionTime time.Time) error {
 	return restoreFile(v.copyRangeMethod, v.versionsFs, v.folderFs, filepath, versionTime, TagFilename)
 }
 
-func (v simple) Clean(_ context.Context) error {
+func (v simple) Clean(ctx context.Context) error {
```
AudriusButkevicius, line 83: Is there a lot of common code between this function and the one in trashcan versioner?

(Sorry, on the phone hard to check myself).

If yes, perhaps extracting common logic into a shared utility function would be beneficial.

lib/versioner/simple_test.go
rjpruitt16: It's not new. My formatter just moved it down with the rest of the "github.com" imports. I would move it back, but I think it makes since to group these in the same location (even though it is sorting them alphabetically).

lib/versioner/simple.go
rjpruitt16: No. I always thought golang is a language of copy and pasting, but I guess inheritance is not necessary to be DRY. I will change.

rjpruitt16: lol I mean yes you right

lib/versioner/util.go
```diff
@@ -293,3 +294,51 @@ func findAllVersions(fs fs.Filesystem, filePath string) []string {
 
 	return versions
 }
+
+func CleanByDay(cleanoutDays int, folderFs fs.Filesystem, versionsFs fs.Filesystem, ctx context.Context) error {
```
AudriusButkevicius, line 298: This can be package private by being lowercase.

AudriusButkevicius: Also, go always has context as the first argument. I think in this case I'd have `(ctx, folderFs, versionerFs, days)` which somewhat gives them hierarcy of importance.

AudriusButkevicius: Actually, folderFs is not even used.

rjpruitt16: I want to read code as well as you. Thank you for teaching me so much. I promise to do better next time.

lib/versioner/versioner_test.go
```diff
@@ -0,0 +1,103 @@
+// Copyright (C) 2014 The Syncthing Authors.
```
AudriusButkevicius, line 1: 2020

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -144,7 +144,7 @@
               <span translate ng-if="folderEditor.externalCommand.$error.required && folderEditor.externalCommand.$dirty">The path cannot be blank.</span>
             </p>
           </div>
-          <div class="form-group" ng-if="currentFolder.fileVersioningSelector == 'staggered' || currentFolder.fileVersioningSelector == 'trashcan'" ng-class="{'has-error': folderEditor.versioningCleanupIntervalS.$invalid && folderEditor.versioningCleanupIntervalS.$dirty}">
+          <div class="form-group" ng-if="currentFolder.fileVersioningSelector" ng-class="{'has-error': folderEditor.versioningCleanupIntervalS.$invalid && folderEditor.versioningCleanupIntervalS.$dirty}">
```
imsodin, line 147:
```suggestion
          <div class="form-group" ng-if="currentFolder.fileVersioningSelector != 'none'" ng-class="{'has-error': folderEditor.versioningCleanupIntervalS.$invalid && folderEditor.versioningCleanupIntervalS.$dirty}">
```

lib/versioner/versioner_test.go
```diff
@@ -0,0 +1,103 @@
...
+		Path:           "testdata",
+		Versioning: config.VersioningConfiguration{
+			Params: map[string]string{
+				"cleanoutDays": "7",
+			},
+		},
+	}
+	testCasesVersioner := []Versioner{
+		newSimple(simpleCfg),
+		newTrashcan(trashCfg).(*trashcan),
```
imsodin, line 44: Why do you need the type assertion?

lib/versioner/versioner_test.go
```diff
@@ -0,0 +1,103 @@
...
+		{"testdata/.stversions/keep2/file1", false},
+		{"testdata/.stversions/keep2/file2", true},
+		{"testdata/.stversions/keep3/keepsubdir/file1", false},
+		{"testdata/.stversions/remove/file1", true},
+		{"testdata/.stversions/remove/file2", true},
+		{"testdata/.stversions/remove/removesubdir/file1", true},
+	}
+
+	for index, versioner := range testCasesVersioner {
+		t.Run(fmt.Sprintf("versioner trashcan clean up %d in %d", index, len(testCasesVersioner)), func(t *testing.T) {
```
imsodin, line 64: I'd rather store the versioners in a map using names as key and use those for logging here - more informative than e.g. "0 in 2".

lib/versioner/versioner_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	trashCfg := config.FolderConfiguration{
+		FilesystemType: fs.FilesystemTypeBasic,
+		Path:           "testdata",
+		Versioning: config.VersioningConfiguration{
+			Params: map[string]string{
+				"cleanoutDays": "7",
+			},
+		},
+	}
```
imsodin, line 41: Either I'm blind or this is an exact copy of `simpleCfg` and thus there' shouldn't be any need to have two separate ones.

lib/versioner/versioner_test.go
rjpruitt16: I saw written like that in the test previously. I assume somebody had a good reason for it, but after a second look, I removed it.

lib/versioner/versioner_test.go
rjpruitt16: I cannot pass the filename to the logging in t.Run(string, func(){}) since the looping of the file occurs inside the body of the t.Run function. Can you point me to an example of logging in syncing?

imsodin: What I mean is changing the current slice to a map like this

```
	testCasesVersioner := map[string]Versioner{
		"simple": newSimple(simpleCfg),
		"trashcan"; newTrashcan(trashCfg),
```

and then

```
	for versionerType, versioner := range testCasesVersioner {
		t.Run(fmt.Sprintf("%v versioner clean up", versionerType, len(testCasesVersioner)), func(t *testing.T) {
```

rjpruitt16:
I am not sure if that test is my fault. I check the build logs and it failed on a go get. Is there a way to trigger the build coverage restart without having to add something.

AudriusButkevicius:
It's unlikely your fault. It's not a required check anyway.

Give a shout once this is good to review again.

rjpruitt16:
its ready to be look at again

AudriusButkevicius (approved):

rjpruitt16:
I apologize.

rjpruitt16:
not ready yet. Give me 10 minutes

rjpruitt16:
Alright. The html has been elucidated through a checkout of the upstream/main instead my origin/main. I am ready for it be seen.

rjpruitt16:
:)

AudriusButkevicius:
Still looks good to me, but I'll leave for others to do the approval.

imsodin (approved):

AudriusButkevicius:
Bah, didn't fix the merge commit name :(

rjpruitt16:
yeah I wanted to do a git rebase so it does not look so sloppy but whatever. That was painful embarrassing effort on my part, but on the bright side, I doubt my next pull request be as sloppy. Thank you for teaching me so much and having patience with me.  @imsodin @calmh  @AudriusButkevicius I hope to be a productive member of the syncthing maintainers for years to come.

AudriusButkevicius:
Thanks for your patience with the picky old timers and thanks for the contribution.

PR #6885, opened 2020-08-09
lib/osutil: Preserve perms in AtomicWriter (fixes #6886)

### LINKED ISSUE #6886: chmod permission after editing .stignore via WebUI not preserved
Ref.: https://forum.syncthing.net/t/preserve-chmod-permission-after-editing-stignore-via-webui/15461

Reproducer:

chmod 0755 /path/to/folder/.stignore
Edit ignore patterns for “folder” from Syncthing’s Web UI
Syncthing always resets permission to 0600

What I like to suggest:

Keep the file permission if the user changed it. e.g. 0755 in my case.

Thank you.

Affects: all Syncthing versions since years (0.14.x ... current)

> imsodin
That might even qualify as a bug: Our atomic writer simply doesn’t care about permissions of the final product. In any case please file an issue on github

imsodin:
Preserves permissions if a file is replaced/truncated. Reported in https://forum.syncthing.net/t/preserve-chmod-permission-after-editing-stignore-via-webui/15461, waiting for the issue (number) to fix the commit msg.

### DIFF
--- a/lib/osutil/atomic.go
+++ b/lib/osutil/atomic.go
@@ -93,7 +93,12 @@ func (w *AtomicWriter) Close() error {
 		return err
 	}
 
-	err := w.fs.Rename(w.next.Name(), w.path)
+	info, err := w.fs.Lstat(w.path)
+	if err != nil && !fs.IsNotExist(err) {
+		w.err = err
+		return err
+	}
+	err = w.fs.Rename(w.next.Name(), w.path)
 	if runtime.GOOS == "windows" && fs.IsPermission(err) {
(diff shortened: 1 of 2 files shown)


lib/osutil/atomic.go
```diff
@@ -105,6 +110,12 @@ func (w *AtomicWriter) Close() error {
 		w.err = err
 		return err
 	}
+	if info != nil {
```
calmh, line 113: Info might theoretically be non-nil yet garbage if the file didn't exist. (As in I think info is undefined if err != nil)

PR #6893, opened 2020-08-13
lib/dialer: Try dialing without reuse in parallel (fixes #6892)

### LINKED ISSUE #6892: Syncthing 1.8.0 trying to connect via wrong network route
Related forum thread with details: https://forum.syncthing.net/t/syncthing-trying-to-connect-via-wrong-network-route/15476

Affected: Syncthing 1.8.0

Summary:
Syncthing explicitly specifies the source-IP when connecting to a remote device which ignores the routes configured in the operating system and can lead to problems connecting. Syncthing v.1.7.1 works fine in the same configuration.

AudriusButkevicius:
Reusing the listen address influences the routing decision, which is not what we want in some cases.

### DIFF
--- a/lib/dialer/public.go
+++ b/lib/dialer/public.go
@@ -159,6 +159,7 @@ func dialTwicePreferFirst(ctx context.Context, first, second dialFunc, firstName
 		case <-firstDone:
 			if firstErr == nil {
 				// First succeeded, no point doing anything in second
+				secondErr = errors.New("didn't dial")
 				close(secondDone)
 				return
 			}
--- a/lib/dialer/public.go
+++ b/lib/dialer/public.go
@@ -83,37 +83,15 @@ func dialContextWithFallback(ctx context.Context, fallback proxy.ContextDialer, 
 		return dialerConn{conn, newDialerAddr(network, addr)}, nil
 	}
 
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-	var proxyConn, fallbackConn net.Conn
-	var proxyErr, fallbackErr error
-	proxyDone := make(chan struct{})
-	fallbackDone := make(chan struct{})
-	go func() {
-		proxyConn, proxyErr = dialer.DialContext(ctx, network, addr)
-		l.Debugf("Dialing proxy result %s %s: %v %v", network, addr, proxyConn, proxyErr)
-		if proxyErr == nil {
-			proxyConn = dialerConn{proxyConn, newDialerAddr(network, addr)}
+	proxyDialFudgeAddress := func(ctx context.Context, network, address string) (net.Conn, error) {
+		conn, err := dialer.DialContext(ctx, network, addr)
+		if err != nil {
+			return nil, err
 		}
-		close(proxyDone)
-	}()
-	go func() {
-		fallbackConn, fallbackErr = fallback.DialContext(ctx, network, addr)
-		l.Debugf("Dialing fallback result %s %s: %v %v", network, addr, fallbackConn, fallbackErr)
-		close(fallbackDone)
-	}()
-	<-proxyDone
-	if proxyErr == nil {
-		go func() {
-			<-fallbackDone
-			if fallbackErr == nil {
-				_ = fallbackConn.Close()
-			}
-		}()
-		return proxyConn, nil
+		return dialerConn{conn, newDialerAddr(network, addr)}, err
 	}
-	<-fallbackDone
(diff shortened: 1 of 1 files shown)


lib/dialer/public.go
```diff
@@ -125,19 +125,52 @@ func DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
...
+	var reuseConn, nonReuseConn net.Conn
+	var reuseErr, nonReuseErr error
+	reuseDone := make(chan struct{})
+	nonReuseDone := make(chan struct{})
+	go func() {
+		dialer := &net.Dialer{
+			Control:   ReusePortControl,
+			LocalAddr: laddr,
+		}
+		reuseConn, reuseErr = dialContextWithFallback(ctx, dialer, network, addr)
```
imsodin, line 157: Earlier you checked the proxy env var already, so this will be a direct dial so might as well do that directly.

calmh:
This is not super pretty imho as it triggers a weird warning (though INFO level) on every connection...

```
[MRIW7] 10:28:57 INFO: Listen (BEP/tcp): TLS handshake: EOF
[MRIW7] 10:28:57 INFO: Established secure connection to ...etc
```

AudriusButkevicius:
wut, it shouldn't

calmh:
I'm sure it does, but on the receiving side it's two connections, one of which gets dropped during the TLS handshake...

AudriusButkevicius:
Right, I guess we could do it sequentially, rather than in parallel, but we'd potentially double the timeout.

imsodin:
> Alternatively, we could not log failed TLS handshakes (as it might as well be a port scan etc), but it's sometimes useful to observe when firewall gets in the way.

Wouldn't the connection not even get through to the listener when a firewall gets in the way, i.e. it never comes to the tls handshake?

AudriusButkevicius:
I've seen cases where the tcp handshake passes, then the firewall notices it's TLS by checking the first few bytes, and blocks it, because potentially it can't eavesdrop as it's configured.

lib/dialer/public.go
```diff
@@ -125,19 +103,63 @@ func DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
...
+	var firstConn, secondConn net.Conn
+	var firstErr, secondErr error
+	firstDone := make(chan struct{})
+	secondDone := make(chan struct{})
+	go func() {
+		firstConn, firstErr = first(ctx, network, address)
+		l.Debugf("Dialing %s result %s %s: %v %v", firstName, network, address, firstConn, firstErr)
+		close(firstDone)
+	}()
+	go func() {
```
calmh, line 148: Could we add a small delay here, like 1/3 the timeout or such, in the hope that in the normal case we never even get to the point of doing the second dial?

AudriusButkevicius: It will be a bit ugly having to reach into the context to get the deadline, but sure.

calmh: Well I mean a hardcoded 500ms would go a long way as well I think

lib/dialer/public.go
```diff
@@ -125,19 +103,73 @@ func DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
...
+		time.Sleep(sleep)
+		secondConn, secondErr = second(ctx, network, address)
```
calmh, line 159: Are we relying on that `first` succeeding will cause return and context cancellation and `second` then aborting early here? Otherwise I'd expect the sleep to be a select on `ctx`, `firstDone`, `time.After` so that we don't enter `second` if `first` succeeds

AudriusButkevicius: firstDone can also be closed if it fails, so selecting on that doesn't really tell us if we should proceed or not.

Trying to come up with logic that covers this, but it's pretty hairy.

lib/dialer/public.go
```diff
@@ -125,19 +103,85 @@ func DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
...
+		firstConn, firstErr = first(ctx, network, address)
+		l.Debugf("Dialing %s result %s %s: %v %v", firstName, network, address, firstConn, firstErr)
+		close(firstDone)
+	}()
+	go func() {
+		select {
+		case <-firstDone:
+			if firstErr == nil {
+				// First succeeded, no point doing anything in second
+				close(secondDone)
```
calmh, line 162: Should we set `secondErr = errors.New("didn't dial")` or something just in case? Otherwise LGTM

PR #6894, opened 2020-08-14
all: Subscribe to config changes on Serve

imsodin:
This is a first step towards the new suture API discussed in https://github.com/thejerf/suture/issues/39:

Currently we subscribe to config updates with different methods. The problematic one is when we subscribe in the constructor and unsubscribe in `Stop`. That won't be possible anymore with the new suture api, and regardless is a bad idea (the object being constructed is no guarantee that `Serve` and `Stop` will ever be called). Thus I added a `WithConfigSubscription` option to our service wrappers, that subscribes to config, takes care of the initial config setup and unsubscribes when `Serve` returns. To facilitate the initial config setup I changed `Wrapper.Subscribe` to return a copy of the config that's active at the time of subscribing.

That produced an import cycle on events, config and util package, thus I removed usage of events from config and split services into a new serviceutil package (just service doesn't work as a name as it's way too general, conflicts e.g. with connections.service). That has the nice side-effect of removing lots of `events.NoopLogger` usages all over. 

For supervisors that need to subscribe to config updates, it's a bit more complicated, as `ServeBackground` needs to work, where we don't know when `Serve` returns. However overloading supervisor methods to do setup/cleanup was always a hack and is actually unnecessary: The more elegant option is to add a tiny service to the supervisor that takes care of that. I did that on model and connections.service.

The entire diff is terribly long due to search&replace, but the relevant bit's are not that much: `serviceutil.WithConfigSubscription` and connected code (the rest is copy&paste, even if github doesn't mark it thus unfortunately), wrapper.go, connections.go and model.go.

### DIFF
--- a/cmd/stcli/main.go
+++ b/cmd/stcli/main.go
@@ -21,7 +21,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/syncthing/syncthing/lib/build"
 	"github.com/syncthing/syncthing/lib/config"
-	"github.com/syncthing/syncthing/lib/events"
 	"github.com/syncthing/syncthing/lib/locations"
 	"github.com/syncthing/syncthing/lib/protocol"
 	"github.com/urfave/cli"
@@ -85,7 +84,7 @@ func main() {
 		myID := protocol.NewDeviceID(cert.Certificate[0])
 
 		// Load the config
-		cfg, err := config.Load(locations.Get(locations.ConfigFile), myID, events.NoopLogger)
+		cfg, err := config.Load(locations.Get(locations.ConfigFile), myID)
 		if err != nil {
 			log.Fatalln(errors.Wrap(err, "loading config"))
 		}
--- a/cmd/strelaysrv/main.go
+++ b/cmd/strelaysrv/main.go
@@ -20,7 +20,6 @@ import (
 	"time"
 
 	"github.com/syncthing/syncthing/lib/build"
-	"github.com/syncthing/syncthing/lib/events"
 	"github.com/syncthing/syncthing/lib/osutil"
 	"github.com/syncthing/syncthing/lib/relay/protocol"
 	"github.com/syncthing/syncthing/lib/tlsutil"
@@ -184,7 +183,7 @@ func main() {
 		log.Println("ID:", id)
 	}
 
-	wrapper := config.Wrap("config", config.New(id), events.NoopLogger)
+	wrapper := config.Wrap("config", config.New(id))
 	wrapper.SetOptions(config.OptionsConfiguration{
 		NATLeaseM:   natLease,
 		NATRenewalM: natRenewal,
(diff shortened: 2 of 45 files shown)


AudriusButkevicius:
So a small glance at this and the following came up:
1. Moving the event logging out of config looks like a mistake. It introduces a ton of code, and opens up a new class of "whoops we forgot to send the event" errors.
2. Perhaps the package rename should be done separately in a separate PR before hand, as it's a huge source of noise in the PR as a whole. This is literally a 2 second job with GoLand if you use it (if you don't you should, we have OSS licenses I think).
3. I don't like the initCfg. I guess the purpose of it is to be able to distinguish between "first setup" and subsequent changes?
I guess if this is the way forward to subscribing to config changes (by implementing a service effectively), I'd either change the signature of Verify/CommitConfiguration to have `from *config.Configuration`, passing in a nil value when it's wiring up, or add a OnInitialization(cfg) to config.Comitter.

Also I am still not sure I agree with the analysis here.
Yes, stop is going away, but the lifetime of each Serve(ctx) is still well defined. 
I can't see why services that care about config changes can't just do `Sub` + `defer Unsub` inside of serve?

imsodin:
Yeah I guess I am already mixing up too many things here, will look at splitting it up.

1. That's a consequence of the import cycle. However I do also like it: Config being a "low-level" package I think it makes sense that it doesn't have to deal with comments. The number of places where we give it a `events.NoopLogger` is a testament that it's not inherently linked.

2. `svc` is used too. `svcutil`?

3. There's various different ways of initial setup config: Different ways of calling `CommitConfiguration` and custom initial setup code. This is just a convenience option for the service wrapper that subscribes to config updates and does an arbitrary initial config setup, such that it doesn't have to be done in every `Serve`.

As to the analysis: You're right, and I am not concerned with those services. I am with those that subscribe in the constructor and unsubscribe in `Stop`. And when refactoring I thought I might as well add a convenience option to the service wrappers that does just that. However that "convenience" has apparently a huge rats tail. I'll split this up/reduce the change.

AudriusButkevicius:
1. The amount of places we need to pass a logger in real code should be like ... one, when bootstrapping the config. The amount of times we have to call events.Log() because we no longer pass in the logger into the config is more than one. What's even worse is that this opens doors for a new class of bugs.  If you really want to save that one argument when calling Wrap in tests, split it into Wrap and WrapWithEventLogger if you are that concerned, alternatively, event logger could be wrapped in config.Commiter, and on CommitConfiguration could send the event.

2. I don't see a svc library. If we have structs called svc, we can rename them to service to keep svc for suture stuff.

3. Ideologically, services should be stateless, hence setup and tear down should all happen on serve. The whole passing references to functions as arguments made me want to vomit before when we did `foo.service = util.AsService(foo.isThisABirdOrAPlaneOrWhat)`, but I did not see a better way. I don't want to keep doing that if we can avoid it.
I would rather have things satisfy a new OnInitialization interface the svc package, where the service wrapper checks if the service passed implements that, and if it does, boot straps it, but please, not functions as variables.

PR #6895, opened 2020-08-15
build: Bump goleveldb to latest commit

imsodin:
There's been a few commits on it lately, including one PR from @calmh preventing a nil deref.


calmh:
x/sys has interactions with gopsutil which makes this not build on freebsd. Upgrading gopsutil as well solves that, but at least previously it broke one of the other architectures, there's an issue on it there.

imsodin:
issue: https://github.com/shirou/gopsutil/issues/912

proposed fix: https://github.com/shirou/gopsutil/pull/920

imsodin:
Now that we've upgrade gopsutil independently of this, this is good to go.

PR #6900, opened 2020-08-18
all: Use protobuf to generate config structs (fixes #6734)

### LINKED ISSUE #6734: Consider moving configuration to a proto contract
I think it would not be too hard to move structs to be generated from proto.

gogoproto supports additional tags as well as json overrides:
 https://github.com/gogo/protobuf/blob/master/test/tags/tags.proto

which should be sufficient for our purposes to replicate what we have now.

This proto file could then be used by other languages to generate serialisation code/classes for the config, and proto as a format natively supports json serialization, atleast in every language I used it.

I guess we can't make other clients not generate deprecated fields (even tho there is `deprecated` extension for that), but we could still ignore them when deserializing json by setting the json tag to `-`

Reason I am proposing this is because in non-dynamic languages, the classes you use to deserialize the data needs to map exactly to what the config has, otherwise you'll lose fields, and then potentially unset them when posting the config back.

In dynamic languages, this is not or a problem as you can just have dic of dics in python with random types and nobody cares.

Obviously, people might not use the latest version of tne contract, which could still lead to the same issue, but I guess we could make it more obvious in the release notes that people need to update the contract, especially when the default value of the field is not an empty value (urUrl for example)

AudriusButkevicius:
Tell me how much you hate me.

This was a long and manual effort of translating from one to another....

### What does this give us

This PR at this point is completely noop and gives us nothing.

Even if someone generated java classes from the proto files and serialized and posted the config using JSON, it would not work,
as we don't use proto json serialization, but we use `encoding/json`, which disagrees on the the field naming.

We could add a new rest endpoint that uses proto serialization, but it's probably better to build out a new API all together.

This does however open a possibility to build a GRPC API on top of this.

### How do I know I haven't screwed something up?

I wrote a small script `proto/scripts/dump_tags.go` that walked the config struct and dumped it's tags into a CSV.
I then compared the CSV of this PR with the CSV of main branch and ended up with this diff:
image

### Follow-ups 

This is also an opportunity to clean up the names of the variables atleast in the proto contract, so no more ururl, crurl etc.
Open to further suggestions here.

I potentially intend to follow up this PR removing most of the `(ext.goname)` stuff, because `CRURL` is super cryptic, and I suspect having magical Go names might not be kosher with grpc-gateway going forward, as I think it expects standard proto compiler generated fields. This is yet to be seen.

I guess this means LDAP, TLS, ID will become Ldap, Tls, Id, but I think it's an ok price to pay in our own codebase.

Fixes #6734


calmh:
Okidokie then... 😲 First impressions before getting into the weeds...

- Is there a technical reason for having a bajillion tiny proto files, one for each enum etc., instead of a larger config.proto? I think the latter would give an easier overview?

- Tiny details, but with standard gogoproto I can say `int32 some_field [(gogoproto.casttype) = "int")]` and get a field that serializes to int32 on the wire but is declared as `int` in the Go struct, which is nicer to use and avoids some conversions in other places. Perhaps we can do that in some places here.

I'll look closer later :)

AudriusButkevicius:
I kept it a thing per file because every thing has an extension file where we have marshal string and what not. Every one of those things was in a separate file before.

We can do cast to int, I can add that, but my gut says ti try and use as less of the gogoproto magic as possible that alerts names types etc, as I suspect it will cause issues with grpc gateway.

calmh:
Fair enough on the files. I think we should start out with retaining the simplicity (elsewhere) of simple `int` types, and if it becomes a problem down the line we can remove the casttype and use the wire type everywhere.

lib/config/deviceconfiguration.pb.go
```diff
@@ -0,0 +1,212 @@
+// Code generated by protoc-gen-gogo. DO NOT EDIT.
+// source: lib/config/deviceconfiguration.proto
+
+package config
+
+import (
+	fmt "fmt"
+	proto "github.com/gogo/protobuf/proto"
+	github_com_syncthing_syncthing_lib_protocol "github.com/syncthing/syncthing/lib/protocol"
+	protocol "github.com/syncthing/syncthing/lib/protocol"
```
imsodin, line 10: Something is weird here

imsodin (commented):
Wow, this is something indeed ;)

I have only tried to follow what's going on in the plugin and extensions so far, not looked at naming/types.

> I potentially intend to follow up this PR removing most of the (ext.goname) stuff, because CRURL is super cryptic, and I suspect having magical Go names might not be kosher with grpc-gateway going forward, as I think it expects standard proto compiler generated fields. This is yet to be seen.

I agree on the naming, however while technically we are "allowed" to do this on a minor version bump, it would be a pretty harsh move given the config is the main (only) way of controlling Syncthing through the API. If we did this, we might as well do the encoding/json to "proto-json" switch too. And carrying around both old and new names at the same time sounds like a nightmare (e.g. custom decoding taking an `ext.legacyName` into account). One option is to make a (rough) plan on where we want to go with the api, and announce that plan along with a preview of the new config (with preview == proto-generated json).

lib/config/deviceconfiguration.pb.go
AudriusButkevicius: Auto-generated code is auto-generated.
One is added for custom type conversion, the other one is added as a dependency of the proto type.

proto/ext.proto
```diff
@@ -0,0 +1,24 @@
+syntax = "proto2";
```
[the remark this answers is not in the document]
AudriusButkevicius: Extensions need to be proto2 I believe.

AudriusButkevicius:
The naming changes I am talking about only affect struct member names in go code, this has no effect on externally (json/xml tags stay the same).

lib/config/deviceconfiguration.pb.go
imsodin: Right, that's the `ext.device_id` thing. What's the difference to just using `(gogoproto.customtype) = "DeviceID"` like in bep.proto?

AudriusButkevicius: It would have to be a fully qualified path, so
`(gogoproto.customtype = "github.com/syncthing/syncthing/lib/protocol.DeviceID"`

imsodin:
The generation doesn't work for me:

```
Could not make proto path relative: lib/config/*.proto: No such file or directory
proto/generate.go:20: running "protoc": exit status 1
```

Looks like the issue is globbing not being supported by go generate: https://github.com/golang/go/issues/31591

How did you do it? :)

AudriusButkevicius:
Works for me on windows?

AudriusButkevicius:
I guess generate.go could invoke itself, and walk the dirs?

imsodin:
That's weird. Anyway do we need to use go generate? Couldn't `build.go proto` directly invoke protoc (walking or globbing to get all the proto files).

AudriusButkevicius:
It's quite a lot of generating, so extracting that out into a separate script that deals with just proto stuff makes sense to me?

AudriusButkevicius:
So should I try to fix all the things that need fixing or is this a lost cause?

imsodin:
In my opinion this is definitely not a lost cause. I mean it isn't really a gain so far, because all of the other puzzle pieces are still missing, but it looks maintainable and the huge work has already sunk in. And if an api-v2 really comes of this in the end, all the better.

calmh:
I think this is fine. The usual `go run build.go proto` appears to work for me. It hopefully doesn't break things, and it enables more things in the future.

Aesthetically I think it would be more pleasing with a single proto file like we do for bep, which would also be a single place for documentation comments and could be used to generate the config article in the docs, etc, but that's something that can happen in the future if we get there as there's no technical impact.

AudriusButkevicius:
I still argue for split files, so that each extension file ends up getting a .pb.go file.

lib/model/model.go
```diff
@@ -1781,7 +1781,9 @@ func (m *model) OnHello(remoteID protocol.DeviceID, addr net.Addr, hello protoco
 func (m *model) GetHello(id protocol.DeviceID) protocol.HelloIntf {
 	name := ""
 	if _, ok := m.cfg.Device(id); ok {
-		name = m.cfg.MyName()
+		if myCfg, ok := m.cfg.Device(m.id); ok {
```
calmh, line 1784:
```suggestion
		// Set our name (from the config of our device ID) only if we already know about the other side device ID.
		if myCfg, ok := m.cfg.Device(m.id); ok {
```

calmh: (because I was staring at the construction and wtf:ing)

calmh (commented):
lgtm I guess

proto/ext/ext.pb.go
```diff
@@ -0,0 +1,131 @@
+// Code generated by protoc-gen-gogo. DO NOT EDIT.
```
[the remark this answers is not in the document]
AudriusButkevicius: We need this to build the plugin, but I guess it could be ignored and generated on demand.

proto/lib/config/optionsconfiguration.proto
```diff
@@ -0,0 +1,66 @@
...
+    int32           nat_traversal_timeout_s                  = 17 [(ext.goname) = "NATTimeoutS", (ext.xml) = "natTimeoutSeconds", (ext.json) = "natTimeoutSeconds", (ext.default) = "10"];
+    int32           usage_reporting_accepted                 = 18 [(ext.goname) = "URAccepted", (ext.xml) = "urAccepted", (ext.json) = "urAccepted"];
+    int32           usage_reporting_seen                     = 19 [(ext.goname) = "URSeen", (ext.xml) = "urSeen", (ext.json) = "urSeen"];
+    string          usage_reporting_unique_id                = 20 [(ext.goname) = "URUniqueID", (ext.xml) = "urUniqueID", (ext.json) = "urUniqueId"];
+    string          usage_reporting_url                      = 21 [(ext.goname) = "URURL", (ext.xml) = "urURL", (ext.json) = "urURL", (ext.default) = "https://data.syncthing.net/newdata"];
+    bool            usage_reporting_unique_post_insecurely   = 22 [(ext.goname) = "URPostInsecurely", (ext.xml) = "urPostInsecurely", (ext.json) = "urPostInsecurely", (ext.default) = "false"];
```
imsodin, line 31: What's the significance of "unique"?

proto/ext/ext.pb.go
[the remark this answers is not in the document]
calmh: No that's fine then

imsodin (commented):
Yup, now it works here too - thanks. Weirdly  the formatting is different after running proto for one file only here:

```diff
--- a/lib/config/ldapconfiguration.pb.go
+++ b/lib/config/ldapconfiguration.pb.go
@@ -59,7 +59,9 @@ func init() {
        proto.RegisterType((*LDAPConfiguration)(nil), "config.LDAPConfiguration")
 }
 
-func init() { proto.RegisterFile("lib/config/ldapconfiguration.proto", fileDescriptor_9681ad7e41c73956) }
+func init() {
+       proto.RegisterFile("lib/config/ldapconfiguration.proto", fileDescriptor_9681ad7e41c73956)
+}
 
 var fileDescriptor_9681ad7e41c73956 = []byte{
        // 498 bytes of a gzipped FileDescriptorProto
```

Except for one question below, lgtm

proto/lib/config/optionsconfiguration.proto
AudriusButkevicius: Looks like a copy paste typo

imsodin (approved):

calmh:
boom

proto/scripts/protoc_plugin.go
```diff
@@ -0,0 +1,291 @@
...
+						// addresses -> address
+						xmlValue = strings.TrimSuffix(xmlValue, "es")
+					} else {
+						// devices -> device
+						xmlValue = strings.TrimSuffix(xmlValue, "s")
+					}
+				}
+				if deprecated {
+					xmlValue += ",omitempty"
+				}
```
calmh, line 270: holy shit the magic ✨  I didn't notice this stuff on first pass

AudriusButkevicius: Magic is magical

lib/config/config.go
```diff
@@ -177,29 +182,13 @@ func ReadJSON(r io.Reader, myID protocol.DeviceID) (Configuration, error) {
...
-	Folders        []FolderConfiguration `xml:"folder" json:"folders"`
-	Devices        []DeviceConfiguration `xml:"device" json:"devices"`
-	GUI            GUIConfiguration      `xml:"gui" json:"gui"`
-	LDAP           LDAPConfiguration     `xml:"ldap" json:"ldap"`
-	Options        OptionsConfiguration  `xml:"options" json:"options"`
-	IgnoredDevices []ObservedDevice      `xml:"remoteIgnoredDevice" json:"remoteIgnoredDevices"`
-	PendingDevices []ObservedDevice      `xml:"pendingDevice" json:"pendingDevices"`
-	XMLName        xml.Name              `xml:"configuration" json:"-"`
-
-	MyID            protocol.DeviceID `xml:"-" json:"-"` // Provided by the instantiator.
```
acolomb, line 199: > Tell me how much you hate me.

I hate you very much ;-)

Where can I find this piece of information now?

AudriusButkevicius: Its still on the wrapper I think.

acolomb: Never mind, I found it in `model.id` and agree it does not need to be magically inserted into the configuration struct.

PR #6907, opened 2020-08-19
lib/config, lib/fs: Make junction behaviour configurable (ref #6606)

imsodin:
See https://forum.syncthing.net/t/option-to-follow-directory-junctions-symbolic-links/14750/126 and following.

### DIFF
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -9,6 +9,7 @@ package fs
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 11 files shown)


AudriusButkevicius:
I think we do a config migration of this, setting to true on all existing folders, and defaulting to false.

imsodin:
Passing fs options as url-like query parameters might not be the best idea: A filesystem path does not necessarily conform to what `url.Parse` expects. An alternative is adding the usual `...opts Option` thing to `NewFilesystem` and then have `WithJunctionsAsDirs() Option`, though I believe Audrius has a bit of a disliking to that pattern (?).

AudriusButkevicius:
I don't mind options, I mind functions thrown around like they are variables, yet I think query string stuff is the right way to go here.

imsodin:
Well these options are functions disguised as options ;)  

Anyway my problem is that almost anything is allowed in paths, i.e. you need to do parsing yourself (`url.Parse` fails). At which point I'd rather not do the detour of encoding options into a pseudo-uri and then decoding the same options from that pseudo-uri, but pass them directly.

AudriusButkevicius:
so my original idea was to turn C:\foo\bar into file://c/foo/bar, but it never happened. If it did happen, this would be much easier.

This also would have allowed getting rid fo the type, as the schema dictates the type.

imsodin:
That makes sense, but being an unrelated change and quite possible to turn up weird cornercases, I used options, It's still easy enough to switch to "all uri initialisation" from here.

AudriusButkevicius (dismissed):

lib/fs/basicfs.go
```diff
@@ -23,13 +23,20 @@ var (
 	ErrNotRelative     = errors.New("not a relative path")
 )
 
+func WithJunctionsAsDirs() Option {
+	return func(fs Filesystem) {
+		fs.(*BasicFilesystem).junctionsAsDirs = true
```
calmh, line 28: Panic if I ask for a fakefs with junctionsAsDirs?

AudriusButkevicius (approved):

calmh (approved):

PR #6911, opened 2020-08-19
lib/model: Don't mark globally deleted items as recv-only (fixes #6910)

### LINKED ISSUE #6910: Local Additions in unshared Receive Only folders
Now, if this is really intended, please close the issue, but I think that it is a little bit unexpected that unshared Receive Only folders can even have local additions.

Case 1:

1. Create a new Receive Only folder and keep it unshared.
2. Add some files/folders to it.
3. The folder is now marked Local Additions.

Case 2:

1. Create a new Send & Receive folder and keep it unshared.
2. Add some files/folders to it.
3. Change the folder type to Receive Only.
4. Delete the previously added files.
5. The folder is now marked Local Additions, but the revert button does nothing (as the folder is unshared, and thus cannot "undelete" the removed files).

imsodin:


imsodin:
In addition to the new check I also removed a recently introduced, but ineffective and redundant check: Ineffective because it changed the flag on the wrong variable (`file` instead of `nf`) and redundant because for a deleted, receive-only file that noone else has, the global item is a receive-only undeleted file, which is already covered by another check.

calmh:
some conflictery

PR #6912, opened 2020-08-19
lib/nat: Don't hang on draining timer chan (fixes #6908)

### LINKED ISSUE #6908: NAT service fails to terminate during Shutdown
Since version 1.8.0, when-ever a device is told to shutdown syncthing, the process hangs for a bit and the log file prints out:

INFO: connections.Service: Service nat.Service@0xc000184600 failed to terminate in a timely manner.

Based on the logfiles, this takes about 9 secs to timeout. Aside from this slowing down the shutdown process, I am concerned with that service even running since I have NAT transversal disabled on all my syncthing devices. Perhaps this is likely a leftover bug from the 1.7 branch where the process was running even if not configured to.

I have this issue confirmed in Win7,10 & Linux, all running 1.8.0.

imsodin:
The timer chan might already be drained at this point.

### DIFF
--- a/lib/nat/service.go
+++ b/lib/nat/service.go
@@ -85,7 +85,10 @@ func (s *Service) serve(ctx context.Context) {
 		case <-timer.C:
 		case <-s.processScheduled:
 			if !timer.Stop() {
-				<-timer.C
+				select {
+				case <-timer.C:
+				default:
+				}
 			}
 		case <-ctx.Done():
 			timer.Stop()


AudriusButkevicius:
I think I'd just drop it on the floor all together?

How does it get to this code-path if it's disabled?

imsodin:
That's just the main for-select block in serve, where it's waiting for a config change that activates it (or if actived, waits on the timer or something else to trigger a round of natting).

AudriusButkevicius:
Sure, but I don't understand how that prevents it from stopping when it's disabled, as we should never even hit that branch of the select.

imsodin:
When a new mapping is added (which happens regardless of nat enabled or not) processing is scheduled, timer stopped and then before actually processing, there's the check if nat is even enabled. The advantage of doing it in the loop is that there doesn't need to be any locks protecting the `enabled` bool.

AudriusButkevicius (approved):

PR #6913, opened 2020-08-19
lib/model: Do db heavy ops without holding fmut

imsodin:
This is still about those deadlock false positives (https://sentry.syncthing.net/syncthing/syncthing/issues/29/?query=is%3Aunresolved%20deadlock%20detected), but this time instead of butchering the deadlock detection (#6909) I propose to acquire locks inside the loop when needed, instead of for the entire duration. The `FileSet.Drop` operation in there needs to potentially update a whole lot of global lists, which can take a while (longer than the deadlock interval apparently)


calmh (approved):
I guess...

Issue #6864, opened 2020-07-29
Folder incorrectly marked “Up to Date” after switching from “Receive Only” to “Send & Receive”


tomasz1986:
Steps to reproduce:

1. Create a new folder on Device A, set it to "Send Only", and add some files to it.
2. Share the folder with Device B.
3. Accept the folder on Device B and set its type to "Receive Only".
4. Delete some files from the folder on Device B. The folder is now marked "Local Additions".
5. Change the folder type on Device B to "Send & Receive".
6. Both folders show "Up to Date", but the files remain deleted, which is reflected in the "Local State" on Device B.

Device A:

![](https://user-images.githubusercontent.com/5626656/88773678-657df880-d1bd-11ea-91de-ad66643b5bec.png)

Device B:

![](https://user-images.githubusercontent.com/5626656/88775903-416fe680-d1c0-11ea-8f33-ab9a15a50050.png)

PR #6919, opened 2020-08-20
lib/model: Prevent localflag flipflopping on del. item (ref #6865)

imsodin:
Unfortunately, this is RC material:  
The fix for #6864 in #6865 was way too general: E.g. a deleted file in a receive-only folder, that doesn't exist anywhere else, gets the receive-only flag removed (because there's no local change). Thus it's local flags are different from the folder, which triggered the previous condition to readd the receive-only flag. And then the check for the first condition removes the flag again. And so on and so forth. Now the check is specific to the problem in #6864: If a deleted file has the receive-only flag set, but the folder is not receive-only, the flag is removed.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -637,11 +637,11 @@ func (f *folder) scanSubdirs(subDirs []string) error {
 				l.Debugln("marking deleted item that doesn't exist anywhere as not receive-only", file)
 				batchAppend(file.ConvertDeletedToFileInfo(), snap)
 				changes++
-			case file.IsDeleted() && file.LocalFlags != f.localFlags:
-				// No need to bump the version for a file that was
-				// and is deleted and just the local flags changed.
-				file.LocalFlags = f.localFlags
-				l.Debugln("changing localflags on deleted item", file)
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
This is becoming so complicated that I feel we'll soon need to start modelling this bullshit in alloy (https://alloytools.org/) to formally prove ourselves that all of these PRs is not just whack-a-mole.

imsodin:
I get the sentiment and I am looking forward to an alloy replica of bullshit ;)  
It is improving though. It's not just added complexity, it solves stuff. There's even a few new tests every now and then proving that. And I do remember times where it was for example just generally accepted that the remote status is whacky when you don't have ignores, and downright useless when you have. And if the moon was in the first quarter while venus and mars could be seen Syncthing unleashed havoc onto your systems. Now it's about not having to press the ugly red button when a locally deleted file also gets deleted from all remotes and the like (yeah, yeah, I am aware there's also bigger fish still swimming happily in the lake waiting to be fried, still, they got rarer and aren't that huge anymore).

PR #6925, opened 2020-08-24
all: Add failure reporting (fixes #7028)

### LINKED ISSUE #7028: Detect and optionally report non-fatal failures
Currently in many places where we encounter a fatal error or a condition that should never occur, we panic. And there's a mechanism in place to catch and locally store those panics for debugging, possibly reporting them if enabled. However there's also instances where a condition should never occur, but it isn't fatal (because we can repair or ignore and reasonable assume the program will keep working). At the same time these conditions may be symptoms of an underlying problem that needs fixing, thus there should be some way for such issues to get to our attention.  
I propose to emit events in such conditions and add a (configurable) option to send reports about them.

imsodin:
https://forum.syncthing.net/t/non-panic-reports-to-sentry/15483

This adds a new `Failure` event type and a service that tracks those events and sends them to the crash receiver if crash reporting is enabled. That service aggregates identical events before sending. Unfortunately I couldn't find any way to send many events or an event counter to sentry. Thus I just send aggregated events as a single event (with an custom `count` field). I don't think we lose much information that way.

There's a single example usage for now (non-increasing index detected) - I'll look for more places where we do "that shouldn't happen ever, but" "...we can fix it" or "...lets ignore it" and add events there too once the general approach is "approved".


imsodin:
Just in case: This PR is WIP because I intend to add failure reporting in more places if the general mechanism is the way to go. As in this is not ready to merge, but ready to review.

lib/failhandler/failhandler.go
```diff
@@ -0,0 +1,165 @@
+// Copyright (C) 2014 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package failhandler
```
calmh, line 7: This seems like it could be in the usage reporting package rather than its own thing?

lib/failhandler/failhandler.go
```diff
@@ -0,0 +1,165 @@
...
+	return err
+}
+
+func (h *handler) VerifyConfiguration(_, _ config.Configuration) error {
+	return nil
+}
+
+func (h *handler) CommitConfiguration(from, to config.Configuration) bool {
+	if from.Options.CREnabled != to.Options.CREnabled || from.Options.CRURL != to.Options.CRURL {
+		h.optsChan <- to.Options
```
calmh, line 136: With timeout, in case the other end is sleeping due to repeated send failure? Or buffered chan and drop with warning or something?

calmh (commented):
Looks reasonable to me, but I wonder if it doesn't make more sense to tie this to usage reporting than crash reporting...

imsodin:
It's somewhere in between the two. As this together with service termination is meant to replace the anti-pattern of `if unexpected { panic("lets die") }` I considered it more like crash reporting. In that it doesn't send any characteristics about your instance/usage, but events about harmful incidents. But I am sure the verdict can easily go either way.

lib/failhandler/failhandler.go
imsodin: If we treat it like usage reporting, then sure, otherwise it feels weird having them in the same place.

lib/failhandler/failhandler.go
imsodin: I do the sending in a goroutine now to keep process events and configs while sending.

cmd/stcrashreceiver/main.go
```diff
@@ -0,0 +1,110 @@
...
+
+	log.SetOutput(os.Stdout)
+	if err := http.ListenAndServe(*listen, mux); err != nil {
+		log.Fatalln("HTTP serve:", err)
+	}
+}
+
+func handleFailureFn(dsn string) func(w http.ResponseWriter, req *http.Request) {
+	return func(w http.ResponseWriter, req *http.Request) {
+		bs, err := ioutil.ReadAll(req.Body)
```
AudriusButkevicius, line 58: Limit reading a megabyte or so?

lib/failhandler/failhandler.go
```diff
@@ -0,0 +1,173 @@
...
+	// When a specific failure first occurs, it is delayed by minDelay. If
+	// more of the same failures occurs those are further delayed and
+	// aggregated for maxDelay.
+	minDelay    = 10 * time.Second
+	maxDelay    = time.Minute
+	sendTimeout = time.Minute
+)
+
+type Report struct {
+	Descr   string
```
AudriusButkevicius, line 34: Full name please.

lib/failhandler/failhandler.go
```diff
@@ -0,0 +1,173 @@
...
+	reqCtx, reqCancel := context.WithTimeout(ctx, sendTimeout)
+	defer reqCancel()
+	req, err := http.NewRequestWithContext(reqCtx, http.MethodGet, url, &b)
+	if err != nil {
+		l.Infoln("Failed to send failure report:", err)
+		return
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := http.DefaultClient.Do(req)
```
AudriusButkevicius, line 166: Does this handle proxy stuff correctly? I'd check what discovery does.

lib/model/model.go
```diff
@@ -2011,6 +2011,7 @@ func (s *indexSender) sendIndexTo(ctx context.Context) error {
 					l.Warnln("Failed repairing sequence entries:", dbErr)
 					panic("Failed repairing sequence entries")
 				} else {
+					s.evLogger.Log(events.Failure, "detected and repaired non-increasing sequence")
```
AudriusButkevicius, line 2014: Still at a loss how knowing this is useful?

I guess to verify when the bug has been fixed?

imsodin: Yep. We have no clue how bad the problem is right now, because we have an ugly mitigation. Lately there is the "device missing but present in global" thing that with actual db error handling might be mitigatable too, but I don't want to just mitigate without any reporting. Without reporting if a fix comes in (as it hopefully lately did for this exact problem), we have no clue whether the fix was effective or not.

imsodin:
I switched to connect this to usage reporting, just to be on the save side: Given when we use this mechanism, we clearly aren't crashing (even if it replaces a previous `panic`) - so crash reporting is wrong. I added the events in two more places in model. I expect more usages once doing actual error handling on the database (which is happening :) ). I'd also like to use it in lib/db, but that diff is huge as it adds a new parameter to `NewLowlevel`, so I want to do it in a separate PR (https://github.com/imsodin/syncthing/compare/all/failureReporting...imsodin:all/failureReporting-db).

cmd/stcrashreceiver/stcrashreceiver.go
```diff
@@ -155,8 +124,12 @@ func (r *crashReceiver) servePut(reportID, fullPath string, w http.ResponseWrite
 
 		go func() {
 			// There's no need for the client to have to wait for this part.
-			if err := sendReport(r.dsn, reportID, bs, user); err != nil {
-				log.Println("Failed to send report:", err)
+			pkt, err := parseCrashReport(reportID, bs)
+			if err != nil {
+				log.Println("Failed to parse crash report:", err)
+			}
+			if err := sendReport(r.dsn, pkt, user); err != nil {
```
AudriusButkevicius, line 131: In case or error, this would still send the report

AudriusButkevicius (approved):
other than that one comment, looks good.

AudriusButkevicius (dismissed):

lib/ur/failurereporting.go
```diff
@@ -0,0 +1,181 @@
+// Copyright (C) 2014 The Syncthing Authors.
```
calmh, line 1:
```suggestion
// Copyright (C) 2020 The Syncthing Authors.
```

lib/ur/failurereporting.go
```diff
@@ -0,0 +1,181 @@
...
+			descr := e.Data.(string)
+			if stat, ok := h.buf[descr]; ok {
+				stat.last = e.Time
+				stat.count++
+			}
+			h.buf[descr] = &stat{
+				first: e.Time,
+				last:  e.Time,
+				count: 1,
+			}
```
calmh, line 105: This first changes the stat in `h.buf[descr]` then overwrites it, so count is always 1 etc?

lib/ur/failurereporting.go
```diff
@@ -0,0 +1,181 @@
...
+						Version:     build.LongVersion,
+					})
+					delete(h.buf, descr)
+				}
+			}
+			if len(reports) > 0 {
+				// Lets keep process events/configs while it might be timing out for a while
+				go func() {
+					sendFailureReports(ctx, reports, url)
+					timer.Reset(minDelay)
```
calmh, line 123: timer is not goroutine safe, should reset from select loop

calmh (approved):

calmh:
looks good, too bad it doesn't compile

AudriusButkevicius (approved):

calmh:
This needs an issue describing the use case / problem being solved, for release notes

PR #6926, opened 2020-08-24
lib/syncthing: Also cleanup on startup error

imsodin:
I mostly implemented sutures new wip api (https://github.com/thejerf/suture/issues/39) so far and as is to be expected, the diff humongous and I touching all this stuff also uncovered some problems/possible improvements. This is one where we currently do not cleanup when an error occurs on startup (cleanup as in shutting down services and closing db). Now we first start the supervisor and routine waiting for it to exit and do said cleanup, before doing the startup operations.


imsodin:
I added a new condition to the existing startup failure unit test to prove that it actually fixes something (yes, it does fail on main :) ).

calmh (approved):

PR #6931, opened 2020-08-26
gui: Fix add folder bug (fixes #6930)

### LINKED ISSUE #6930: Can not add new folders on master.
I am running master because my system is on go 1.15, I pulled master this afternoon and everything launches fine but I can not add new folders on my linux system. The save button is unavailable (mouse cursor slashed circle) even if I use the default autogenerated directory and do not supply a label.

Gentoo Linux
Go 1.15
syncthing v1.9.0-rc.4.dev.4.gb19b5c95
qutebrowser (qtwebengine 5.15.0) and Firefox 79.0-r2

rjpruitt16:
### Purpose
Fix the add folder bug by providing a default value greater than 0. Also, my previous commit had a bug of not showing the intervals at all. 
https://github.com/syncthing/syncthing/issues/6930#issuecomment-680204973.


calmh:
The default here ought to be 3600 seconds (one hour)

imsodin:
And given you have contributed a few times (thanks a lot!), and hopefully will keep doing so: Could you please prefix the commit msg/PR title with the affected package in the future (https://docs.syncthing.net/dev/contributing.html#commits) - thanks.

rjpruitt16:
I also would like rephrase it. It is not actually a default value. It just giving it a nonzero value when nothing has been selected. I believe the bug has to do with it causing an invalid error on null folders. Would you prefer that when someone pulls up the form that they actually see 3600 or is it just good enough that one can create new folders? @calmh

rjpruitt16:
I tried doing a get rebase and became confused and scared. I thought it would be like a merge squash, but it was not working how I thought it would be. Is it good enough that my last commit follow the style guide or is there a better way to squash the commits?

calmh:
The only thing that matters is the end result diff after however many commits and whatnot. In the end we squash merge it into a single commit.

calmh (approved):

PR #6935, opened 2020-08-27
cmd/strelaypoolsrv: Fix relay shuffling

greatroar:
### Purpose

In strelaypoolsrv's handleGetRequest, when `cap(permanentRelays) >= len(permanentRelays) + len(knownRelays)`,

	append(permanentRelays, knownRelays...)

returns a slice of the array underlying permanentRelays. The subsequent rand.Shuffle then mixes the permanent and known relays. Worse, multiple goroutines may be shuffling the same slice concurrently.

I believe that as long as strelaypoolsrv is running on amd64 hardware, the worst things that can happen are

* duplicate relays being reported to clients, and
* permanent relays being forgotten over time until a restart.

The knownRelays array is never modified, so that is safe.

### DIFF
--- a/cmd/strelaypoolsrv/main.go
+++ b/cmd/strelaypoolsrv/main.go
@@ -314,8 +314,11 @@ func handleRequest(w http.ResponseWriter, r *http.Request) {
 
 func handleGetRequest(rw http.ResponseWriter, r *http.Request) {
 	rw.Header().Set("Content-Type", "application/json; charset=utf-8")
+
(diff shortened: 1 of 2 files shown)


ReidFubush (dismissed):
Sorry for the confusion love your response thank you

calmh (approved):
Good catch. I suspect this has gone unnoticed for us because we don't run the "official" servers with any permanent relays since many years.

PR #6939, opened 2020-08-27
lib/db, lib/model: Avoid accounting mishap on folder restart (fixes #6938)

### LINKED ISSUE #6938: Modifying config during first scan corrupts folder state
I have seen this problem several times but have not been able to reproduce until now.

How to reproduce:

1. Add a new folder, preferably with a lot of files (so that it takes some time to scan it).
2. While the folder is being scanned for the first time, modify some of its configuration. I changed `Max Conflicts` from `10` to `0`.
3. The folder state is incorrect.

image

Below is the same folder, which was just left to scan, with no config changes in between. The state is correct here.

image

Rescanning does nothing. The only way to fix the problem seems to be wiping out the database.

calmh:
### Purpose

We created a new fileset before stopping the folder during restart. When
we create that fileset it loads the current metadata and sequence
numbers from the database. But the folder may have time to update those
before stopping, leaving the new fileset with bad data.

This would cause wrong accounting (forgotten files) and potentially
sequence reuse causing files not sent to other devices.

This change reuses the fileset on restart, skipping the issue entirely.
It also moves the creation of the fileset back under the lock so there
should be no chance of concurrency issues here.

### Testing

Restarting a folder during initial scan no longer causes badness for me.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -452,6 +452,7 @@ func (m *model) restartFolder(from, to config.FolderConfiguration, cacheIgnoredF
 		l.Warnf("bug: folder restart cannot change ID %q -> %q", from.ID, to.ID)
 		panic("bug: folder restart cannot change ID")
 	}
+	folder := to.ID
 
 	// This mutex protects the entirety of the restart operation, preventing
 	// there from being more than one folder restart operation in progress
@@ -459,7 +460,7 @@ func (m *model) restartFolder(from, to config.FolderConfiguration, cacheIgnoredF
 	// because those locks are released while we are waiting for the folder
 	// to shut down (and must be so because the folder might need them as
 	// part of its operations before shutting down).
-	restartMut := m.folderRestartMuts.Get(to.ID)
+	restartMut := m.folderRestartMuts.Get(folder)
 	restartMut.Lock()
 	defer restartMut.Unlock()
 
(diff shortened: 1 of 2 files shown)


lib/model/model.go
```diff
@@ -492,8 +485,17 @@ func (m *model) restartFolder(from, to config.FolderConfiguration, cacheIgnoredF
...
+	// Cache the (maybe) existing fset before it's removed by cleanupFolderLocked
+	fset := m.folderFiles[to.ID]
+
 	m.cleanupFolderLocked(from)
 	if !to.Paused {
+		if fset == nil {
+			// Create a new fset. Might take a while and we do it under
+			// locking, but it's unsafe to create fset:s concurrently so
+			// that's the price we pay.
+			fset = db.NewFileSet(to.ID, to.Filesystem(), m.db)
```
calmh, line 497: Actually we could maybe release `fmut` for the duration since we have the restart lock, and are guaranteed that only CommitConfiguration could run at any given time ... maybe

imsodin: Are we guaranteed that only a single CommitConfiguration can be running at a time? I don't think it is, and that's the only way how fset could ever be nil. If e.g. a folder config change races with the same folder being paused afterwards, and somehow the pausing gets ahead of the change, then we might try to restart a folder that's already stopped. (we should base the "from-state" on the internal state, not the `from` config given to CommitConfiguration to avoid such problems, but that's a different matter)

calmh: No, actually we're not guaranteed that since it's called from cfg.Replace() which is for example called from inside ClusterConfig(). Potentially we could expand the restartMut to cover all config changes for a folder (currently it doesn't) for the meantime I'm happier keeping this inside fmut

imsodin (approved):

AudriusButkevicius:
I suspect this is a big source of bugs, yet I am not sure if its the only source.

PR #6941, opened 2020-08-29
lib/model: Adjust fileinfo size when adding the trailer

imsodin:
I did some manual testing of #6727 and together with the fixes pushed there directly, this is the last change required to make it "work for me" (as in I can share files to an encrypted folder without observing immediate brokenness). However this fix is a bit ugly, thus as a separate PR for discussion:  
When pulling changes on an encrypted folder we add a trailer to the file, containing the encrypted file info. Thus the size on disk is bigger than the size recorded in the file info. That gets picked up by the scanner as an illegal local change. My proposed fix is to just increase the fileinfo size by the length of the trailer. The ugly side-effect of this is that the local state will be bigger (in bytes) than the global one. I think that's not a problem (you may even consider it a feature, as it's the actual size on disk), but maybe there's another solution. Or we could hide the global byte count, given it's not correct anyway on encrypted folders.


calmh (approved):
Sounds sane lets try it

PR #6947, opened 2020-09-01
all: Implement suture v4-api

imsodin:
This PR adapts to the new suture API (jcontext branch, https://github.com/thejerf/suture/issues/39). The diff is huge, but the actual changes arent:  
The first commit just adds a convenience function to create a default supervisor spec to make the next commits nicer (no behaviour change).
The second, main commit implements the new API with minimal changes. Meaning I stayed as close as possible to how things worked before (e.g. moving code in `Stop` to a `OnSupervisorStop` utility service that executes that code when the supervisor and thus the service is stopped).  
The third commit changes lib/syncthing and lib/api to make use of the new interface. Meaning instead of getting a controller from lib/syncthing, api now exits with the new "fatal-error" that takes down all services and bubbles the error to lib/syncthing.

This seems to work (runs and can shutdown syncthing from web UI), however suture v4 isn't released yet. So there's some time to hone this out.

One followup to this will be exposing errors on lib/db and let consumers react to those errors, i.e. terminate through the suture api instead of panic unless the db error can be handled. I started on it and that's going to be huge change (tons of functions in db and model need to gain error returns...), so I'll open a PR against this PR when ready.

### DIFF
--- a/lib/api/api.go
+++ b/lib/api/api.go
@@ -53,6 +53,7 @@ import (
 	"github.com/syncthing/syncthing/lib/tlsutil"
 	"github.com/syncthing/syncthing/lib/upgrade"
 	"github.com/syncthing/syncthing/lib/ur"
+	"github.com/syncthing/syncthing/lib/util"
 )
 
 // matches a bcrypt hash and not too much else
@@ -81,7 +82,6 @@ type service struct {
 	connectionsService   connections.Service
 	fss                  model.FolderSummaryService
 	urService            *ur.Service
-	contr                Controller
 	noUpgrade            bool
 	tlsDefaultCommonName string
 	configChanged        chan struct{} // signals intentional listener close due to config change
@@ -89,15 +89,10 @@ type service struct {
 	startedOnce          chan struct{} // the service has started successfully at least once
 	startupErr           error
 	listenerAddr         net.Addr
+	exitChan             chan *util.FatalErr
 
 	guiErrors logger.Recorder
 	systemLog logger.Recorder
-}
-
-type Controller interface {
-	ExitUpgrading()
-	Restart()
-	Shutdown()
 }
 
 type Service interface {
@@ -106,7 +101,7 @@ type Service interface {
 	WaitForStart() error
 }
 
-func New(id protocol.DeviceID, cfg config.Wrapper, assetDir, tlsDefaultCommonName string, m model.Model, defaultSub, diskSub events.BufferedSubscription, evLogger events.Logger, discoverer discover.Manager, connectionsService connections.Service, urService *ur.Service, fss model.FolderSummaryService, errors, systemLog logger.Recorder, contr Controller, noUpgrade bool) Service {
+func New(id protocol.DeviceID, cfg config.Wrapper, assetDir, tlsDefaultCommonName string, m model.Model, defaultSub, diskSub events.BufferedSubscription, evLogger events.Logger, discoverer discover.Manager, connectionsService connections.Service, urService *ur.Service, fss model.FolderSummaryService, errors, systemLog logger.Recorder, noUpgrade bool) Service {
 	return &service{
 		id:      id,
 		cfg:     cfg,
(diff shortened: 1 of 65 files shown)


imsodin:
Test failures are due to a deadlock/race when adding a service and stopping the supervisor: https://github.com/thejerf/suture/issues/41

imsodin:
With suture v4 released, this is ready now.

AudriusButkevicius:
I'm stuck with a small screen in the rain until next week sadly.

lib/api/api.go
```diff
@@ -874,7 +870,13 @@ func (s *service) getDebugFile(w http.ResponseWriter, r *http.Request) {
 
 func (s *service) postSystemRestart(w http.ResponseWriter, r *http.Request) {
 	s.flushResponse(`{"ok": "restarting"}`, w)
-	go s.contr.Restart()
+
+	go func() {
```
calmh, line 874: `go`ing these feels kind of dirty. How about making the channel one-buffered and doing these sends with a select, maybe returning an error if we take the default clause (`"shutdown in progress"`)?

lib/model/folder_summary.go
```diff
@@ -180,7 +178,7 @@ func (c *folderSummaryService) listenForUpdates(ctx context.Context) {
 		case ev := <-sub.C():
 			c.processUpdate(ev)
 		case <-ctx.Done():
-			return
+			return nil
```
calmh, line 181: Is there a difference between returning nil or ctx.Error() in all these cases? (ctx.Error() would be less surprising to me)

calmh (commented):
looks reasonable

lib/model/folder_summary.go
imsodin: My thinking was that being cancelled isn't actually a failure but a shutdown as requested, and `err != nil` usually represents a failure. However I agree the existence of `context.Cancelled` implementing `error` challenges that assertion. Our service types have an `Error()` method which remember the error the function returns. I went looking for where we use that, and the only place I found is connections. So if the user e.g. disables relays they will get no error or a context cancelled error in the GUI next to the relay address. For dialers we do filter context cancelled errors:

```golang
func (s *connectionStatusHandler) setConnectionStatus(address string, err error) {
	if errors.Cause(err) == context.Canceled {
		return
	}
```

I'd go with returning `ctx.Error()` and expect anyone running a service to know that `context.Cancelled` is special. Respectively in most cases it won't be relevant anyway, because we are shutting down anyway.

calmh: Right, I hade a vague feeling that it was possible to cancel a context with an arbitrary error, but it's not so the only possible error returns are context.Canceled and context.DeadlineExceeded. I think it would make sense to return the error regardless; in all cases I can imagine the context cancelling should mean we are shutting down so no action will be taken based on it... At least the `case err := <- ctx.Done: return err` is visually common and unconfusing

calmh (dismissed):
lgtm, does it work? :)

imsodin:
Nope, shutting down is broken, it exits with code 1 and thus restarts. I assume I need to filter `context.Canceled` in `syncthing.App` - tbc

imsodin:
What prevented shutdown was that `util.FatalError` is special, wasn't considered in `syncthing.App.run`. It is an error, because it has to be for the suture API, but it can also represent a graceful shutdown or reboot, which isn't a failure.

Unrelated changes: Fixed a copy-paste mistake in api.go and log the error that caused an unclean shutdown in cmd/syncthing/main.go.

lib/syncthing/syncthing.go
```diff
@@ -353,6 +350,19 @@ func (a *App) run(ctx context.Context) {
 	close(a.stopped)
 }
 
+func (a *App) handleMainServiceError(err error) {
+	if err == nil || errors.Is(err, context.Canceled) {
+		return
+	}
+	if fatalErr, ok := err.(*util.FatalErr); ok {
```
calmh, line 357: `errors.As`?

calmh (approved):

PR #6948, opened 2020-09-01
BUILDHOST name changed for Arch

svetlemodry:
Hi all!
To make the Arch Linux syncthing package reproducible, I've fixed the BUILDHOST variable in the build instructions. So I belive this code needs to be changed too, to reflect the change.

Some more details are in the bug report [here](https://bugs.archlinux.org/task/67279). The change was implemented in the last  [PKGBUILD](https://github.com/archlinux/svntogit-community/commit/705393bbda70210db9825da85c031435d1f2a5dc#diff-8d0411b338c83cd8cd8ad9d9db127101).
Cheers!
J

### DIFF
--- a/cmd/ursrv/main.go
+++ b/cmd/ursrv/main.go
@@ -54,7 +54,7 @@ var (
 		{regexp.MustCompile("jenkins@build.syncthing.net"), "GitHub"},
 		{regexp.MustCompile("snap@build.syncthing.net"), "Snapcraft"},
 		{regexp.MustCompile("android-.*vagrant@basebox-stretch64"), "F-Droid"},
-		{regexp.MustCompile("builduser@svetlemodry"), "Arch (3rd party)"},
+		{regexp.MustCompile("builduser@archlinux"), "Arch (3rd party)"},
 		{regexp.MustCompile("synology@kastelo.net"), "Synology (Kastelo)"},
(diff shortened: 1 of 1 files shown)


calmh:
Oh so *you're* svetlemodry, well met! :) Thanks for noticing and letting us know. The regexp should probably match both old and new though, as we'll be seeing the old value for years to come as well.

svetlemodry:
Yep, that's me .)
Not sure what should be the correct value there though to cover both, so please fix as you see fit.

cmd/ursrv/main.go
```diff
@@ -54,7 +54,7 @@ var (
 		{regexp.MustCompile("jenkins@build.syncthing.net"), "GitHub"},
 		{regexp.MustCompile("snap@build.syncthing.net"), "Snapcraft"},
 		{regexp.MustCompile("android-.*vagrant@basebox-stretch64"), "F-Droid"},
-		{regexp.MustCompile("builduser@svetlemodry"), "Arch (3rd party)"},
+		{regexp.MustCompile("builduser@archlinux"), "Arch (3rd party)"},
```
calmh, line 57:
```suggestion
		{regexp.MustCompile("builduser@(archlinux|svetlemodry)"), "Arch (3rd party)"},
```

calmh (approved):

PR #6950, opened 2020-09-02
lib/db: Add local need check&repair

imsodin:
I hate this as much as everyone else if not more, but it turns out it is needed: Wracking my brain on https://forum.syncthing.net/t/syncthing-stuck-syncing-because-of-device-that-doesnt-exist/15501 and the out-of-sync folders with nothing in the out-of-sync list, I realized we do check and repair all kinds of things but never our "cached" list of files we need locally. In this PR I add the check and run it in the usual place. I ran it on my devices and on both it did repair stuff. I don't know what the underlying issue is. It would be nice to have https://github.com/syncthing/syncthing/pull/6925 to report if there's anything to repair, to know if this is an ongoing problem or if I have just been cleaning up relics of the past on my devices (wishful thinking).

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -1071,10 +1071,14 @@ func (db *Lowlevel) checkLocalNeed(folder []byte) (int, error) {
 	defer dbi.Release()
 
 	var needName string
-	needDone := !dbi.Next()
-	if !needDone {
-		needName = string(t.keyer.NameFromGlobalVersionKey(dbi.Key()))
+	var needDone bool
+	next := func() {
+		needDone = !dbi.Next()
+		if !needDone {
+			needName = string(t.keyer.NameFromGlobalVersionKey(dbi.Key()))
+		}
 	}
+	next()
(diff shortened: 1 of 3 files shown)


calmh:
Sounds reasonable, fails tests

imsodin:
And it wasn't a test needing an update, the check/repair was broken - I was lucky that this test caught the mistake... I added a dedicated unit test.

lib/db/lowlevel.go
```diff
@@ -1035,6 +1049,81 @@ func (db *Lowlevel) repairSequenceGCLocked(folderStr string, meta *metadataTrack
...
+	for !needDone {
+		repaired++
+		if err := t.Delete(dbi.Key()); err != nil {
+			return 0, err
+		}
+		l.Debugln("check local need: removing", needName)
+		next()
+	}
+
+	if err = t.Commit(); err != nil {
```
calmh, line 1120: Needs explicit check of iterator error and closing it, currently panics with `panic: Unclosed iterator at time of Txn.Discard` with Badger

PR #6959, opened 2020-09-07
build, lib/db: Upgrade Badger

calmh:


calmh:
Right, I forgot they don't test 32 bit builds

PR #6960, opened 2020-09-07
lib/db: Don't hang on double iterator release with Badger

calmh:
Since iterators must be released before committing or discarding a
transaction we have the pattern of both deferring a release plus doing
it manually. But we can't release twice because we track this with a
WaitGroup that will panic when there are more Done()s than Add()s. This
just adds a boolean to let an iterator keep track.

(This is needed to let the conversion back from Badger to LevelDB work, which is a required step in retiring the Badger experiment...)

### DIFF
--- a/lib/db/backend/badger_backend.go
+++ b/lib/db/backend/badger_backend.go
@@ -334,6 +334,7 @@ type badgerIterator struct {
 	first     []byte
 	last      []byte
 	releaseFn func()
+	released  bool
 	didSeek   bool
 	err       error
 }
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
I think it's pretty cheap for us to keep the code, as it's fairly slow moving.
Perhaps badger will get their stuff together and come back to planet earth.

If anything, we should make a public shaming post and post it on hackernews of how badger lost the ball.

There is also bbolt (which coreos/etc maintain) which might be a good candidate.

AudriusButkevicius (approved):

PR #6962, opened 2020-09-07
lib/db, lib/protocol: Never need empty-version entries (fixes #6961)

### LINKED ISSUE #6961: Accounting issue with receive-only deleted files causing spurious 95% completion for remote devices
An existing in-sync file can be changed and then deleted on a receive-only device. This results in two updates, (1) a file change and (2) a file delete, each with the invalid version (empty) and the invalid bit set. If the proper version is then withdrawn from the cluster (because all devices that had them are removed from the config, for example) we will "upgrade" this locally-changed deleted file to a regular deleted file (no invalid bit) but retain the invalid version to ensure it will never override a real file version, should it reappear.

The result in the database is a global file entry with the empty version, and the delete-bit unset. The delete bit is unset because it changed without the version changing between updates 1 and 2. 

A separate but related factor is how we handle delete records for files we don't have: we simply ignore them. This means that most other devices will not announce anything at all for this zero-version deleted file.

When running a database repair or metadata recovery we will look at all global file entries and compute the need requirements for all devices. At this point it looks like there is a valid non-deleted file that is not present in the announcements from other devices, hence it gets added to their "need" counter.

At this point the need counter is wrong and the device looks like "95% in sync" from our perspective...

calmh:
### Purpose

Avoid havoc when discovering locally-deleted-upgraded files during repair / need calculation...

### Testing

Fixes a broken database I had laying around.

@imsodin mentioned wanting to write a real test for this...

### DIFF
--- a/lib/protocol/bep_extensions.go
+++ b/lib/protocol/bep_extensions.go
@@ -189,10 +189,6 @@ func WinsConflict(f, other FileIntf) bool {
 	return f.FileVersion().Compare(other.FileVersion()) == ConcurrentGreater
 }
 
-func (f FileInfo) IsEmpty() bool {
-	return f.Version.Counters == nil
-}
-
 func (f FileInfo) IsEquivalent(other FileInfo, modTimeWindow time.Duration) bool {
 	return f.isEquivalent(other, modTimeWindow, false, false, 0)
 }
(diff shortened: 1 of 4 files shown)


lib/protocol/bep_extensions.go
```diff
@@ -189,10 +189,6 @@ func WinsConflict(f, other FileIntf) bool {
 	return f.FileVersion().Compare(other.FileVersion()) == ConcurrentGreater
 }
 
-func (f FileInfo) IsEmpty() bool {
-	return f.Version.Counters == nil
-}
```
calmh, line 194: (Dead code)

calmh:
Test LGTM

imsodin (approved):

PR #6969, opened 2020-09-08
lib/fs: Use generic case resolver on windows (fixes #6968)

### LINKED ISSUE #6968: Very slow scans on Windows in 1.9.0
Apparently the case sensitivity changes for Windows result in very slow scans. The general method used on Unixes however works fine on Windows too and is much faster. We should use that on Windows as well.

https://syncthing.community/t/1-9-0-time-for-rescan-increases-dramatically-on-windows/15580

imsodin:
The case sensitivity handling introduced in #6717 is really slow on windows. Originally reported in https://forum.syncthing.net/t/1-9-0-time-for-rescan-increases-dramatically-on-windows/15580. On my system a scan took 1min45s before and 28s after this PR for a ~70k item folder.


lib/fs/casefs.go
```diff
@@ -75,12 +75,7 @@ func NewCaseFilesystem(fs Filesystem) Filesystem {
 	caseFs := &caseFilesystem{
 		Filesystem: fs,
 	}
-	switch k.fstype {
-	case FilesystemTypeBasic:
-		caseFs.realCaser = newBasicRealCaser(fs)
-	default:
-		caseFs.realCaser = newDefaultRealCaser(fs)
-	}
+	caseFs.realCaser = newDefaultRealCaser(fs)
```
calmh, line 78: move into struct initializer?

imsodin: I was thinking it wouldn't hurt to keep it as a separate interface (as it's already done), in case any future filesystem wants to implement their own thing (or for mocking in testing).

calmh: I just mean

```
caseFs := &caseFilesystem{
    Filesystem: fs,
    realCaser: newDefaultRealCaser(fs),
}
```

imsodin: ... got it now. Is there a nice English saying equivalent to the German "Tomaten auf den Augen haben" (probably not having tomatoes on your eyes) :D

calmh (approved):

lib/fs/casefs.go
calmh: Not that I know of :D There's "har du knäck i lurarna" for the hearing equivalent...

PR #6971, opened 2020-09-08
lib/db, lib/syncthing: Don't repair DB on upgrade, but on error (fixes #6917)

### LINKED ISSUE #6917: Reconsider db check on upgrade
It used to be quite quick and painless; seconds on my computer. Now up to about five minutes on my computer. Other setups are worse:

```
<sphalerite> Is there any way to make "Checking db due to upgrade" faster? It takes hours on my phone.
```

I think it's the block loading that now happens...

imsodin:
No longer repair the db unconditionally on upgrade (fixes #6917). Instead when an error due to an inconsistency occurs, mark the db as needing repair. Then on next startup, do the repair.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -82,6 +82,12 @@ func NewLowlevel(backend backend.Backend, opts ...Option) *Lowlevel {
 	}
 	db.keyer = newDefaultKeyer(db.folderIdx, db.deviceIdx)
 	db.Add(util.AsService(db.gcRunner, "db.Lowlevel/gcRunner"))
+	misc := NewMiscDataNamespace(db)
+	if ok, _, _ := misc.Bool("repairDB"); ok {
+		l.Infoln("Database was marked for repair - this may take a while")
+		db.CheckRepair()
+		misc.Delete("repairDB")
+	}
 	return db
 }
 
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -11,6 +11,8 @@ import (
 	"context"
 	"encoding/binary"
 	"io"
+	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/dchest/siphash"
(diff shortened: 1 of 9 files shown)


calmh:
I think this is reasonable, but it doesn't feel like something that needs to leak outside the db package. I suggest:

- setting the repair bit outside the database itself, because it may not be writable or the write+panic might make it not flush (maybe?). Something like touching `databaseDir+".needsrepair"`?

- having the lowlevel itself check for this on open and run the repair before returning the database

calmh:
Oh and perhaps we should add a repair cycle before running migrations, but that's halfway unrelated.

lib/db/lowlevel.go
```diff
@@ -82,6 +86,14 @@ func NewLowlevel(backend backend.Backend, opts ...Option) *Lowlevel {
 	}
 	db.keyer = newDefaultKeyer(db.folderIdx, db.deviceIdx)
 	db.Add(util.AsService(db.gcRunner, "db.Lowlevel/gcRunner"))
+	if loc := backend.Location(); loc != "" {
+		path := filepath.Join(loc, needsRepairFile)
```
calmh, line 90: We should probably not assume that `loc` is a directory that we can do stuff in. We should be able to tack on an extension to loc though

lib/db/lowlevel.go
```diff
@@ -82,6 +86,14 @@ func NewLowlevel(backend backend.Backend, opts ...Option) *Lowlevel {
...
+		if _, err := os.Lstat(path); err == nil {
+			l.Infoln("Database was marked for repair - this may take a while")
+			db.CheckRepair()
```
calmh, line 93: CheckRepair no longer needs to be exported I think

imsodin: A model test was using it, but luckily that was replaceable by `WithRecheckInterval(time.Millisecond)`.

calmh (approved):

PR #6974, opened 2020-09-09
lib/fs: More efficient casefs cache

calmh:
This changes the cache to cache less things, yet retain the required
efficiency for our walk usecase. This uses less memory.

Specifically, instead of keeping result and child caches for each path
level, only keep a single cached child. In practice our operations are
depth-first, or almost depth-first, and then we retain the same hit
ratio for a smaller cache size.

I improved the benchmark so that it counts the Lstat and DirNames
operations performed, and they do not change significantly. The amount
of allocated memory is reduced by 20% and the walk itself is actually
slightly faster.

This also removes the clear based on number of cached names (as that is
not a thing any more) and the timer based clear (which was unused). This
means we'll retain the last cache state forever until it's cleared by a
write operation, but we did that before too and that state is now a lot
smaller...

The overhead compared to not using a casefs, for our typical "double
walk" (walk the tree then stat everything again) is 2x the dirnames we
would otherwise call, and no overhead on the stats (unchanged from old
implementation)

```
name                         old time/op         new time/op         delta
WalkCaseFakeFS100k/rawfs-8           306ms ± 1%          305ms ± 2%     ~     (p=0.182 n=9+10)
WalkCaseFakeFS100k/casefs-8          579ms ± 5%          557ms ± 1%   -3.77%  (p=0.000 n=10+10)

name                         old B/entry         new B/entry         delta
WalkCaseFakeFS100k/rawfs-8             590 ± 0%            590 ± 0%     ~     (all equal)
WalkCaseFakeFS100k/casefs-8          1.09k ± 0%          0.87k ± 0%  -19.98%  (p=0.000 n=10+10)

name                         old DirNames/entry  new DirNames/entry  delta
WalkCaseFakeFS100k/rawfs-8            0.51 ± 0%           0.51 ± 0%     ~     (all equal)
WalkCaseFakeFS100k/casefs-8           1.02 ± 0%           1.02 ± 0%     ~     (all equal)

name                         old DirNames/op     new DirNames/op     delta
WalkCaseFakeFS100k/rawfs-8           51.2k ± 0%          51.2k ± 0%     ~     (all equal)
WalkCaseFakeFS100k/casefs-8           102k ± 0%           102k ± 0%     ~     (all equal)

name                         old Lstat/entry     new Lstat/entry     delta
WalkCaseFakeFS100k/rawfs-8            3.02 ± 0%           3.02 ± 0%     ~     (all equal)
WalkCaseFakeFS100k/casefs-8           3.02 ± 0%           3.02 ± 0%     ~     (all equal)

name                         old Lstat/op        new Lstat/op        delta
WalkCaseFakeFS100k/rawfs-8            302k ± 0%           302k ± 0%     ~     (all equal)
WalkCaseFakeFS100k/casefs-8           302k ± 0%           302k ± 0%     ~     (all equal)

name                         old allocs/entry    new allocs/entry    delta
WalkCaseFakeFS100k/rawfs-8            15.7 ± 0%           15.7 ± 0%     ~     (all equal)
WalkCaseFakeFS100k/casefs-8           27.5 ± 0%           26.1 ± 0%   -5.09%  (p=0.000 n=10+10)

name                         old ns/entry        new ns/entry        delta
WalkCaseFakeFS100k/rawfs-8           2.02k ± 1%          2.02k ± 2%     ~     (p=0.163 n=9+10)
WalkCaseFakeFS100k/casefs-8          3.83k ± 5%          3.68k ± 1%   -3.77%  (p=0.000 n=10+10)

name                         old alloc/op        new alloc/op        delta
WalkCaseFakeFS100k/rawfs-8          89.2MB ± 0%         89.2MB ± 0%     ~     (p=0.364 n=9+10)
WalkCaseFakeFS100k/casefs-8          164MB ± 0%          131MB ± 0%  -19.97%  (p=0.000 n=10+10)

name                         old allocs/op       new allocs/op       delta
WalkCaseFakeFS100k/rawfs-8           2.38M ± 0%          2.38M ± 0%     ~     (all equal)
WalkCaseFakeFS100k/casefs-8          4.16M ± 0%          3.95M ± 0%   -5.05%  (p=0.000 n=10+10)
```

### DIFF
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -16,12 +16,9 @@ import (
 	"time"
 )
 
-// Both values were chosen by magic.
 const (
+	// How long to consider cached dirnames valid
 	caseCacheTimeout = time.Second
-	// When the number of names (all lengths of []string from DirNames)
-	// exceeds this, we drop the cache.
-	caseMaxCachedNames = 1 << 20
 )
 
 type ErrCaseConflict struct {
@@ -47,10 +44,45 @@ type fskey struct {
 	uri    string
 }
 
-var (
-	caseFilesystems    = make(map[fskey]Filesystem)
-	caseFilesystemsMut sync.Mutex
-)
+// caseFilesystemRegistry caches caseFilesystems and runs a routine to drop
+// their cache every now and then.
+type caseFilesystemRegistry struct {
+	fss          map[fskey]*caseFilesystem
+	mut          sync.Mutex
+	startCleaner sync.Once
+}
+
+func (r *caseFilesystemRegistry) get(fs Filesystem) *caseFilesystem {
+	r.mut.Lock()
+	defer r.mut.Unlock()
+
+	k := fskey{fs.Type(), fs.URI()}
+	caseFs, ok := r.fss[k]
+	if !ok {
+		caseFs = &caseFilesystem{
+			Filesystem: fs,
+			realCaser:  newDefaultRealCaser(fs),
+		}
+		r.fss[k] = caseFs
+		r.startCleaner.Do(func() {
+			go r.cleaner()
+		})
+	}
+
+	return caseFs
+}
+
+func (r *caseFilesystemRegistry) cleaner() {
+	for range time.NewTicker(time.Minute).C {
+		r.mut.Lock()
+		for _, caseFs := range r.fss {
+			caseFs.dropCache()
+		}
+		r.mut.Unlock()
+	}
+}
+
+var globalCaseFilesystemRegistry = caseFilesystemRegistry{fss: make(map[fskey]*caseFilesystem)}
 
 // caseFilesystem is a BasicFilesystem with additional checks to make a
 // potentially case insensitive underlying FS behave like it's case-sensitive.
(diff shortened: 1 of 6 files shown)


lib/fs/casefs.go
```diff
@@ -332,85 +324,47 @@ func (r *defaultRealCaser) realCase(name string) (string, error) {
...
+
+			node.expires = time.Now().Add(caseCacheTimeout)
+			node.child = nil
 		}
-		child, ok := node.children[n]
-		if !ok {
-			child = &caseNode{name: n}
+
+		// If we don't already have a correct cached child, try to find it.
+		if node.child == nil || node.child.name != comp {
```
calmh, line 348: Note this also means we will not cache bad lookups (case differs from the one in the filesystem), but we don't need to optimize for those because they are a rare case and will result in an error higher up in the call chain.

imsodin (approved):

imsodin:
Should look at the failing check before approving: `unwrapFilesystem` is currently only used/defined behind `// +build !windows`

imsodin (approved):

PR #6976, opened 2020-09-09
lib/model: Handle symlink deletion on windows (fixes #6972)

### LINKED ISSUE #6972: Deleting a folder containing a symlink results in a permanently out of sync on windows
I have seen this with 1.8.0 and later.

My share has 2 members - one linux and one windows 10.

The same files (soft links) are reported under out of sync items and failed items on the windows side.

To reproduce:

1) share a folder between linux and windows.
2) on linux:
cd \<some directory in the share\>
mkdir x
ln -s ~/.bashrc x/y
\<wait a few seconds to sync\>
rm -rf x

I have not tried this between 2 linux instances.

I understand that soft links are supposed to be ignored - and for my application that works fine. Adding the files to the ignore list might work, but I would rather not have to keep track of all the names.

imsodin:
We never have symlinks on windows, so no need to try to delete them (and get confused by it being in db but not on disk).

Tested manually.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -357,7 +357,12 @@ func (f *sendReceiveFolder) processNeeded(snap *db.Snapshot, dbUpdateChan chan<-
 				// files to delete inside them before we get to that point.
 				dirDeletions = append(dirDeletions, file)
 			} else if file.IsSymlink() {
-				f.deleteFile(file, snap, dbUpdateChan, scanChan)
+				if runtime.GOOS == "windows" {
+					// We never had it, just pretend we deleted it
+					dbUpdateChan <- dbUpdateJob{file, dbUpdateDeleteFile}
+				} else {
+					f.deleteFile(file, snap, dbUpdateChan, scanChan)
+				}
 			} else {
 				df, ok := snap.Get(protocol.LocalDeviceID, file.Name)
 				// Local file can be already deleted, but with a lower version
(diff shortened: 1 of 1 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -357,7 +357,12 @@ func (f *sendReceiveFolder) processNeeded(snap *db.Snapshot, dbUpdateChan chan<-
 				// files to delete inside them before we get to that point.
 				dirDeletions = append(dirDeletions, file)
 			} else if file.IsSymlink() {
-				f.deleteFile(file, snap, dbUpdateChan, scanChan)
+				if runtime.GOOS == "windows" {
+					// We never had it, just pretend we deleted it
+					dbUpdateChan <- dbUpdateJob{file, dbUpdateDeleteFile}
```
calmh, line 362: This might not strictly be true, if it's a file that we synced once that then changed to be a symlink and then finally was deleted -- we would still have the file around? Or maybe that case would have been handled differently to begin with or picked up by the scanner after we ignored the symlink change...?

imsodin: Good point - I moved the fix into `deleteFile`.

Unrelated to the PR but a similar problem: We also don't fully handle valid file to symlink on windows: We do mark it as invalid and synced, but we do not check if there's an old file and remove that. Quite likely on next scan we then pick up the old file and give it a new version, overriding the symlink. Need to check that out.

lib/model/folder_sendrecv.go
```diff
@@ -1976,7 +1976,7 @@ func (f *sendReceiveFolder) checkToBeDeleted(file, cur protocol.FileInfo, hasCur
 		return err
 	}
 	if fs.IsNotExist(err) {
-		if hasCur && !cur.Deleted {
+		if hasCur && !cur.Deleted && !cur.IsUnsupported() {
```
imsodin, line 1979: I wasn't sure whether this should be the generic test for unsupported flag or if it should be `runtime.GOOS == windows && cur.IsSymlink()`. Currently we only use the unsupported flag for this use case, so it applies. And if we keep using that flag for stuff we do *not* ever write to disk (which makes sense when it is unsupported), the generic check is sensible.

calmh (approved):

PR #6977, opened 2020-09-09
lib/model: Consider existing file when handling symlink on windows

imsodin:
### Purpose

When a file is replaced with a symlink (on a non-windows device), we currently leave that file alone on windows and add the symlink with the unsupported flag to the db. On next scan it will see the old file on disk, bump the version and thus reintroduce the old file to the remote, overriding the new symlink. Now before marking the symlink as unsupported on windows, we do check what is on disk and remove it (or rescan if changed).

Came up here: https://github.com/syncthing/syncthing/pull/6976#discussion_r485501779

### Testing

New unit test.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -387,6 +387,9 @@ func (f *sendReceiveFolder) processNeeded(snap *db.Snapshot, dbUpdateChan chan<-
 			}
 
 		case runtime.GOOS == "windows" && file.IsSymlink():
+			if err := f.handleSymlinkCheckExisting(file, snap, scanChan); err != nil {
+				f.newPullError(file.Name, fmt.Errorf("handling unsupported symlink: %w", err))
+			}
 			file.SetUnsupported(f.shortID)
 			l.Debugln(f, "Invalidating symlink (unsupported)", file.Name)
 			dbUpdateChan <- dbUpdateJob{file, dbUpdateInvalidate}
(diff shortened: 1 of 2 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -387,6 +387,9 @@ func (f *sendReceiveFolder) processNeeded(snap *db.Snapshot, dbUpdateChan chan<-
 			}
 
 		case runtime.GOOS == "windows" && file.IsSymlink():
+			if err := f.handleSymlinkCheckExisting(file, snap, scanChan); err != nil {
+				f.newPullError(file.Name, fmt.Errorf("handling unsupported symlink: %w", err))
+			}
```
calmh, line 392: return or continue or something?

calmh: maybe also a `// No reason to retry`, `changed--` as in one place above?

calmh:
This is concerning:

```
16:36:36 INFO: Puller (folder "default" (default), item "link"): syncing: handling unsupported symlink: CreateFile \\?\C:\BuildAgent\temp\buildTmp\syncthing_testFolder-921527387\link: The system cannot find the file specified.
```

lib/model/folder_sendrecv.go
imsodin: Do we really know here that a retry won't help? In the case above it's an invalid file, that will never work. But here it might be e.g. a case-conflicting file that is about to be deleted in the pull later on.

imsodin:
Where's that from? On the failed windows CI job for 88f1d1e I find `  lib\model\folder_sendrecv_test.go:1245:1: expected declaration, found '<<'`, which looks like a merge conflict.

calmh:
CI link below for me,

Screen Shot 2020-09-10 at 10 09 07

imsodin:
It was indeed a `no such file`, I erroneously switched a `IsNotExist` condition...

calmh (approved):

PR #6979, opened 2020-09-10
lib/fs: Optimize UnicodeLowercase

greatroar:
Most notably, it now detects all-lowercase files and returns these as-is. The tests have been expanded with two cases and are now used as a benchmark (admittedly a rather trivial one).

Benchmark results on Linux/amd64 (note that WalkCaseFakeFS100k creates all-lowercase filenames):

    name                           old time/op         new time/op         delta
    WalkCaseFakeFS100k/rawfs-8             388ms ± 1%          312ms ± 3%   -19.64%  (p=0.000 n=10+8)
    WalkCaseFakeFS100k/casefs-8            718ms ± 1%          587ms ± 1%   -18.30%  (p=0.000 n=10+10)
    UnicodeLowercaseMaybeChange-8         5.45µs ± 5%         4.57µs ± 1%   -16.20%  (p=0.000 n=10+10)
    UnicodeLowercaseNoChange-8            5.33µs ± 4%         3.09µs ± 1%   -42.01%  (p=0.000 n=10+10)

    name                           old B/entry         new B/entry         delta
    WalkCaseFakeFS100k/rawfs-8               590 ± 0%            530 ± 0%   -10.17%  (p=0.000 n=10+10)
    WalkCaseFakeFS100k/casefs-8              869 ± 0%            760 ± 0%   -12.54%  (p=0.000 n=10+10)
    
    name                           old allocs/entry    new allocs/entry    delta
    WalkCaseFakeFS100k/rawfs-8              15.7 ± 0%           12.1 ± 0%   -22.93%  (p=0.000 n=10+10)
    WalkCaseFakeFS100k/casefs-8             26.1 ± 0%           20.1 ± 0%   -22.99%  (p=0.000 n=10+10)
    
    name                           old ns/entry        new ns/entry        delta
    WalkCaseFakeFS100k/rawfs-8             2.57k ± 1%          2.06k ± 3%   -19.65%  (p=0.000 n=10+8)
    WalkCaseFakeFS100k/casefs-8            4.75k ± 1%          3.86k ± 1%   -18.63%  (p=0.000 n=10+10)
    
    name                           old alloc/op        new alloc/op        delta
    WalkCaseFakeFS100k/rawfs-8            89.2MB ± 0%         80.2MB ± 0%   -10.08%  (p=0.000 n=9+10)
    WalkCaseFakeFS100k/casefs-8            131MB ± 0%          115MB ± 0%   -12.54%  (p=0.000 n=9+10)
    UnicodeLowercaseMaybeChange-8           320B ± 0%           192B ± 0%   -40.00%  (p=0.000 n=10+10)
    UnicodeLowercaseNoChange-8              320B ± 0%             0B       -100.00%  (p=0.000 n=10+10)
    
    name                           old allocs/op       new allocs/op       delta
    WalkCaseFakeFS100k/rawfs-8             2.38M ± 0%          1.82M ± 0%   -23.44%  (p=0.002 n=8+10)
    WalkCaseFakeFS100k/casefs-8            3.95M ± 0%          3.04M ± 0%   -23.10%  (p=0.000 n=10+9)
    UnicodeLowercaseMaybeChange-8           16.0 ± 0%           10.0 ± 0%   -37.50%  (p=0.000 n=10+10)
    UnicodeLowercaseNoChange-8              16.0 ± 0%            0.0       -100.00%  (p=0.000 n=10+10)

### DIFF
--- a/lib/fs/folding.go
+++ b/lib/fs/folding.go
@@ -34,6 +34,9 @@ func UnicodeLowercase(s string) string {
 // Byte index of the first rune r s.t. lower(upper(r)) != r.
 func firstCaseChange(s string) int {
 	for i, r := range s {
+		if r <= unicode.MaxASCII && (r < 'A' || r > 'Z') {
+			continue
+		}
 		if unicode.ToLower(unicode.ToUpper(r)) != r {
 			return i
 		}
--- a/lib/fs/folding.go
+++ b/lib/fs/folding.go
@@ -7,13 +7,36 @@
 package fs
 
 import (
+	"strings"
 	"unicode"
+	"unicode/utf8"
 )
 
 func UnicodeLowercase(s string) string {
-	rs := []rune(s)
-	for i, r := range rs {
-		rs[i] = unicode.ToLower(unicode.ToUpper(r))
+	i := firstCaseChange(s)
+	if i == -1 {
+		return s
 	}
-	return string(rs)
+
+	var rs strings.Builder
+	// WriteRune always reserves utf8.UTFMax bytes for non-ASCII runes,
+	// even if it doesn't need all that space. Overallocate now to prevent
+	// it from ever triggering a reallocation.
+	rs.Grow(utf8.UTFMax - 1 + len(s))
+	rs.WriteString(s[:i])
+
+	for _, r := range s[i:] {
+		rs.WriteRune(unicode.ToLower(unicode.ToUpper(r)))
+	}
+	return rs.String()
+}
+
+// Byte index of the first rune r s.t. lower(upper(r)) != r.
+func firstCaseChange(s string) int {
+	for i, r := range s {
+		if unicode.ToLower(unicode.ToUpper(r)) != r {
+			return i
+		}
+	}
+	return -1
 }
--- a/lib/fs/folding_test.go
+++ b/lib/fs/folding_test.go
@@ -6,45 +6,71 @@
 
 package fs
 
-import "testing"
+import (
+	"testing"
+)
+
+var caseCases = [][2]string{
+	{"", ""},
+	{"hej", "hej"},
+	{"HeJ!@#", "hej!@#"},
+	// Western Europe diacritical stuff is trivial.
+	{"ÜBERRÄKSMÖRGÅS", "überräksmörgås"},
+	// As are ligatures.
+	{"Æglefinus", "æglefinus"},
+	{"Ĳssel", "ĳssel"},
+	// Cyrillic seems regular as well.
+	{"Привет", "привет"},
+	// Greek has multiple lower case characters for things depending on
+	// context; we should always choose the same one.
+	{"Ὀδυσσεύς", "ὀδυσσεύσ"},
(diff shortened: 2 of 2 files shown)


greatroar:
I must confess that I don't know what's causing

    UnicodeLowercaseMaybeChange-8           320B ± 0%           384B ± 0%   +20.00%  (p=0.000 n=10+10)

Overallocating a few bytes (say, 4 or 10) can decrease the alloc/op to 208B, but I don't understand why. The encoded output is never longer than the encoded input in the benchmark.

greatroar:
Found it, see updated benchmark results and comment in the code.

calmh:
So this is neat, though I think the benchmarks aren't useful here. The improvement is equally large for the `raw` case that doesn't use `casefs`, since the underlying fakefs itself uses UnicodeLowercase a lot to implement its fake case insensitivity.

lib/fs/folding.go
```diff
@@ -7,13 +7,36 @@
...
+	for _, r := range s[i:] {
+		rs.WriteRune(unicode.ToLower(unicode.ToUpper(r)))
+	}
+	return rs.String()
+}
+
+// Byte index of the first rune r s.t. lower(upper(r)) != r.
+func firstCaseChange(s string) int {
+	for i, r := range s {
+		if unicode.ToLower(unicode.ToUpper(r)) != r {
```
calmh, line 37: Maybe shortcut this even further by checking if `r` is an ASCII lowercase letter before doing the whole unicode dance?

greatroar:
> The improvement is equally large for the raw case that doesn't use casefs, since the underlying fakefs itself uses UnicodeLowercase a lot to implement its fake case insensitivity.

I'm not sure I follow completely. Should I remove the rawfs benchmark results from the commit message? The casefs ones? Both?

lib/fs/folding.go
greatroar: Good idea:

```
name                           old time/op    new time/op    delta
UnicodeLowercaseMaybeChange-8    4.59µs ± 2%    4.57µs ± 1%    ~     (p=0.197 n=10+10)
UnicodeLowercaseNoChange-8       3.26µs ± 1%    3.09µs ± 1%  -5.27%  (p=0.000 n=9+10)
```

Commit coming up.

lib/fs/folding.go
```diff
@@ -7,13 +7,39 @@
...
+		return s
 	}
-	return string(rs)
+
+	var rs strings.Builder
+	// WriteRune always reserves utf8.UTFMax bytes for non-ASCII runes,
+	// even if it doesn't need all that space. Overallocate now to prevent
+	// it from ever triggering a reallocation.
+	rs.Grow(utf8.UTFMax - 1 + len(s))
+	rs.WriteString(s[:i])
```
calmh, line 26: I don't think this works in the general case. `firstCaseChange` returns the index of the first rune to have its case changed, while `s[:i]` is counted in bytes. This should break for strings with lowercase non-ASCII characters in the beginning, which may not be any of our current test cases

calmh: My bad, this is fine.

calmh (approved):

PR #7001, opened 2020-09-23
lib/api: Add /rest/config endpoint (fixes #6540)

### LINKED ISSUE #6540: Config PATCH/PUT/DELETE API
Currently we just have the one config API endpoint, `POST /rest/system/config` to replace the whole thing. This isn't optimally easy to use from API clients that just want to change an option or two as you need a while get/change/post and there is also an inherent race condition.

It would be nice to have an API that can twiddle parts of the config easier and safer. From the outside it could look something like this:

- `PUT /rest/system/config/devices/5Y62GZS-...` creates or replaces a single device in its entirety
- `DELETE /rest/system/config/folders/default` deletes the default folder
- `PATCH /rest/system/config/devices/5Y62GZS-...` overwrites the properties present in the given object only, which could be something like `{"name": "Unused for now", "paused": true}`.
- `PUT /rest/system/config/options/localAnnounceEnabled` overwrites the named field in the config, in this case the JSON in the body needs to a boolean like `false`.
- ... and so on

Implementation wise we can mostly just take the current config, JSON serialize it, traverse the resulting object according to the request path, and then remove, replace or unmarshal on top of the resulting leaf object, then commit. (Folders and devices handled specifically as they are arrays in the config object and not dicts.) This would happen while holding the API config lock, so it's atomic from an API perspective.

imsodin:
In this PR I add new rest endpoints at `/rest/config` to allow for partial config reading/changes. I mostly follow what was proposed in #6540, with the exception that I don't support setting individual options like e.g. `/rest/config/options/maxsendkbps`. To do that you do a `PATCH` request on `/rest/config/options` with just the options you want to modify in the body. Individual folders and devices can be modified (or deleted) using e.g. `/rest/config/folders/*folderid*`. It's manual, as in if we add new elements to `config.Configuration`, this needs updating (as in register new endpoint and add some boilerplate) - I am fine with taking care of that.

I added tests for all methods, but by no means exhaustively covering all endpoints.

This will require documentation updates, I'll do that after the first round of (high-level) review.

Fixes #6540

### DIFF
--- a/lib/api/api.go
+++ b/lib/api/api.go
@@ -31,6 +31,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/julienschmidt/httprouter"
 	metrics "github.com/rcrowley/go-metrics"
 	"github.com/thejerf/suture"
 	"github.com/vitrun/qart/qr"
@@ -240,57 +241,80 @@ func (s *service) serve(ctx context.Context) {
 	s.cfg.Subscribe(s)
 	defer s.cfg.Unsubscribe(s)
 
+	restMux := httprouter.New()
+
 	// The GET handlers
-	getRestMux := http.NewServeMux()
-	getRestMux.HandleFunc("/rest/db/completion", s.getDBCompletion)           // [device] [folder]
-	getRestMux.HandleFunc("/rest/db/file", s.getDBFile)                       // folder file
-	getRestMux.HandleFunc("/rest/db/ignores", s.getDBIgnores)                 // folder
-	getRestMux.HandleFunc("/rest/db/need", s.getDBNeed)                       // folder [perpage] [page]
-	getRestMux.HandleFunc("/rest/db/remoteneed", s.getDBRemoteNeed)           // device folder [perpage] [page]
-	getRestMux.HandleFunc("/rest/db/localchanged", s.getDBLocalChanged)       // folder
-	getRestMux.HandleFunc("/rest/db/status", s.getDBStatus)                   // folder
-	getRestMux.HandleFunc("/rest/db/browse", s.getDBBrowse)                   // folder [prefix] [dirsonly] [levels]
-	getRestMux.HandleFunc("/rest/folder/versions", s.getFolderVersions)       // folder
-	getRestMux.HandleFunc("/rest/folder/errors", s.getFolderErrors)           // folder
-	getRestMux.HandleFunc("/rest/folder/pullerrors", s.getFolderErrors)       // folder (deprecated)
-	getRestMux.HandleFunc("/rest/events", s.getIndexEvents)                   // [since] [limit] [timeout] [events]
-	getRestMux.HandleFunc("/rest/events/disk", s.getDiskEvents)               // [since] [limit] [timeout]
-	getRestMux.HandleFunc("/rest/stats/device", s.getDeviceStats)             // -
-	getRestMux.HandleFunc("/rest/stats/folder", s.getFolderStats)             // -
-	getRestMux.HandleFunc("/rest/svc/deviceid", s.getDeviceID)                // id
-	getRestMux.HandleFunc("/rest/svc/lang", s.getLang)                        // -
-	getRestMux.HandleFunc("/rest/svc/report", s.getReport)                    // -
-	getRestMux.HandleFunc("/rest/svc/random/string", s.getRandomString)       // [length]
-	getRestMux.HandleFunc("/rest/system/browse", s.getSystemBrowse)           // current
-	getRestMux.HandleFunc("/rest/system/connections", s.getSystemConnections) // -
-	getRestMux.HandleFunc("/rest/system/discovery", s.getSystemDiscovery)     // -
-	getRestMux.HandleFunc("/rest/system/error", s.getSystemError)             // -
-	getRestMux.HandleFunc("/rest/system/ping", s.restPing)                    // -
-	getRestMux.HandleFunc("/rest/system/status", s.getSystemStatus)           // -
-	getRestMux.HandleFunc("/rest/system/upgrade", s.getSystemUpgrade)         // -
-	getRestMux.HandleFunc("/rest/system/version", s.getSystemVersion)         // -
-	getRestMux.HandleFunc("/rest/system/debug", s.getSystemDebug)             // -
-	getRestMux.HandleFunc("/rest/system/log", s.getSystemLog)                 // [since]
-	getRestMux.HandleFunc("/rest/system/log.txt", s.getSystemLogTxt)          // [since]
+	restMux.HandlerFunc(http.MethodGet, "/rest/db/completion", s.getDBCompletion)           // [device] [folder]
+	restMux.HandlerFunc(http.MethodGet, "/rest/db/file", s.getDBFile)                       // folder file
+	restMux.HandlerFunc(http.MethodGet, "/rest/db/ignores", s.getDBIgnores)                 // folder
+	restMux.HandlerFunc(http.MethodGet, "/rest/db/need", s.getDBNeed)                       // folder [perpage] [page]
+	restMux.HandlerFunc(http.MethodGet, "/rest/db/remoteneed", s.getDBRemoteNeed)           // device folder [perpage] [page]
+	restMux.HandlerFunc(http.MethodGet, "/rest/db/localchanged", s.getDBLocalChanged)       // folder
+	restMux.HandlerFunc(http.MethodGet, "/rest/db/status", s.getDBStatus)                   // folder
+	restMux.HandlerFunc(http.MethodGet, "/rest/db/browse", s.getDBBrowse)                   // folder [prefix] [dirsonly] [levels]
+	restMux.HandlerFunc(http.MethodGet, "/rest/folder/versions", s.getFolderVersions)       // folder
+	restMux.HandlerFunc(http.MethodGet, "/rest/folder/errors", s.getFolderErrors)           // folder
+	restMux.HandlerFunc(http.MethodGet, "/rest/folder/pullerrors", s.getFolderErrors)       // folder (deprecated)
+	restMux.HandlerFunc(http.MethodGet, "/rest/events", s.getIndexEvents)                   // [since] [limit] [timeout] [events]
+	restMux.HandlerFunc(http.MethodGet, "/rest/events/disk", s.getDiskEvents)               // [since] [limit] [timeout]
+	restMux.HandlerFunc(http.MethodGet, "/rest/stats/device", s.getDeviceStats)             // -
+	restMux.HandlerFunc(http.MethodGet, "/rest/stats/folder", s.getFolderStats)             // -
+	restMux.HandlerFunc(http.MethodGet, "/rest/svc/deviceid", s.getDeviceID)                // id
+	restMux.HandlerFunc(http.MethodGet, "/rest/svc/lang", s.getLang)                        // -
+	restMux.HandlerFunc(http.MethodGet, "/rest/svc/report", s.getReport)                    // -
+	restMux.HandlerFunc(http.MethodGet, "/rest/svc/random/string", s.getRandomString)       // [length]
+	restMux.HandlerFunc(http.MethodGet, "/rest/system/browse", s.getSystemBrowse)           // current
+	restMux.HandlerFunc(http.MethodGet, "/rest/system/connections", s.getSystemConnections) // -
+	restMux.HandlerFunc(http.MethodGet, "/rest/system/discovery", s.getSystemDiscovery)     // -
+	restMux.HandlerFunc(http.MethodGet, "/rest/system/error", s.getSystemError)             // -
+	restMux.HandlerFunc(http.MethodGet, "/rest/system/ping", s.restPing)                    // -
+	restMux.HandlerFunc(http.MethodGet, "/rest/system/status", s.getSystemStatus)           // -
+	restMux.HandlerFunc(http.MethodGet, "/rest/system/upgrade", s.getSystemUpgrade)         // -
+	restMux.HandlerFunc(http.MethodGet, "/rest/system/version", s.getSystemVersion)         // -
+	restMux.HandlerFunc(http.MethodGet, "/rest/system/debug", s.getSystemDebug)             // -
+	restMux.HandlerFunc(http.MethodGet, "/rest/system/log", s.getSystemLog)                 // [since]
+	restMux.HandlerFunc(http.MethodGet, "/rest/system/log.txt", s.getSystemLogTxt)          // [since]
 
 	// The POST handlers
-	postRestMux := http.NewServeMux()
-	postRestMux.HandleFunc("/rest/db/prio", s.postDBPrio)                          // folder file [perpage] [page]
-	postRestMux.HandleFunc("/rest/db/ignores", s.postDBIgnores)                    // folder
-	postRestMux.HandleFunc("/rest/db/override", s.postDBOverride)                  // folder
-	postRestMux.HandleFunc("/rest/db/revert", s.postDBRevert)                      // folder
-	postRestMux.HandleFunc("/rest/db/scan", s.postDBScan)                          // folder [sub...] [delay]
-	postRestMux.HandleFunc("/rest/folder/versions", s.postFolderVersionsRestore)   // folder <body>
-	postRestMux.HandleFunc("/rest/system/error", s.postSystemError)                // <body>
-	postRestMux.HandleFunc("/rest/system/error/clear", s.postSystemErrorClear)     // -
-	postRestMux.HandleFunc("/rest/system/ping", s.restPing)                        // -
-	postRestMux.HandleFunc("/rest/system/reset", s.postSystemReset)                // [folder]
-	postRestMux.HandleFunc("/rest/system/restart", s.postSystemRestart)            // -
-	postRestMux.HandleFunc("/rest/system/shutdown", s.postSystemShutdown)          // -
-	postRestMux.HandleFunc("/rest/system/upgrade", s.postSystemUpgrade)            // -
-	postRestMux.HandleFunc("/rest/system/pause", s.makeDevicePauseHandler(true))   // [device]
-	postRestMux.HandleFunc("/rest/system/resume", s.makeDevicePauseHandler(false)) // [device]
-	postRestMux.HandleFunc("/rest/system/debug", s.postSystemDebug)                // [enable] [disable]
+	restMux.HandlerFunc(http.MethodPost, "/rest/db/prio", s.postDBPrio)                          // folder file [perpage] [page]
+	restMux.HandlerFunc(http.MethodPost, "/rest/db/ignores", s.postDBIgnores)                    // folder
+	restMux.HandlerFunc(http.MethodPost, "/rest/db/override", s.postDBOverride)                  // folder
+	restMux.HandlerFunc(http.MethodPost, "/rest/db/revert", s.postDBRevert)                      // folder
+	restMux.HandlerFunc(http.MethodPost, "/rest/db/scan", s.postDBScan)                          // folder [sub...] [delay]
+	restMux.HandlerFunc(http.MethodPost, "/rest/folder/versions", s.postFolderVersionsRestore)   // folder <body>
+	restMux.HandlerFunc(http.MethodPost, "/rest/system/error", s.postSystemError)                // <body>
+	restMux.HandlerFunc(http.MethodPost, "/rest/system/error/clear", s.postSystemErrorClear)     // -
+	restMux.HandlerFunc(http.MethodPost, "/rest/system/ping", s.restPing)                        // -
+	restMux.HandlerFunc(http.MethodPost, "/rest/system/reset", s.postSystemReset)                // [folder]
+	restMux.HandlerFunc(http.MethodPost, "/rest/system/restart", s.postSystemRestart)            // -
+	restMux.HandlerFunc(http.MethodPost, "/rest/system/shutdown", s.postSystemShutdown)          // -
+	restMux.HandlerFunc(http.MethodPost, "/rest/system/upgrade", s.postSystemUpgrade)            // -
+	restMux.HandlerFunc(http.MethodPost, "/rest/system/pause", s.makeDevicePauseHandler(true))   // [device]
+	restMux.HandlerFunc(http.MethodPost, "/rest/system/resume", s.makeDevicePauseHandler(false)) // [device]
+	restMux.HandlerFunc(http.MethodPost, "/rest/system/debug", s.postSystemDebug)                // [enable] [disable]
+
+	// Config endpoints
+
+	configBuilder := &configMuxBuilder{
+		Router: restMux,
+		id:     s.id,
+		cfg:    s.cfg,
+		mut:    sync.NewMutex(),
+	}
+
+	configBuilder.registerConfig("/rest/config/")
+	configBuilder.registerConfigInsync("/rest/config/insync")
+	configBuilder.registerFolders("/rest/config/folders")
+	configBuilder.registerDevices("/rest/config/devices")
+	configBuilder.registerFolder("/rest/config/folders/:id")
+	configBuilder.registerDevice("/rest/config/devices/:id")
+	configBuilder.registerOptions("/rest/config/options")
+	configBuilder.registerLDAP("/rest/config/ldap")
+	configBuilder.registerGUI("/rest/config/gui")
+
+	// Deprecated config endpoints
+	configBuilder.registerConfigDeprecated("/rest/system/config") // POST instead of PUT
+	configBuilder.registerConfigInsync("/rest/system/config/insync")
 
 	// Debug endpoints, not for general use
 	debugMux := http.NewServeMux()
@@ -299,15 +323,14 @@ func (s *service) serve(ctx context.Context) {
 	debugMux.HandleFunc("/rest/debug/cpuprof", s.getCPUProf) // duration
 	debugMux.HandleFunc("/rest/debug/heapprof", s.getHeapProf)
 	debugMux.HandleFunc("/rest/debug/support", s.getSupportBundle)
-	getRestMux.Handle("/rest/debug/", s.whenDebugging(debugMux))
+	restMux.Handler(http.MethodGet, "/rest/debug/", s.whenDebugging(debugMux))
 
-	// A handler that splits requests between the two above and disables
-	// caching
-	restMux := noCacheMiddleware(metricsMiddleware(s.addConfigHandler(getPostHandler(getRestMux, postRestMux))))
+	// A handler that disables caching
+	noCacheRestMux := noCacheMiddleware(metricsMiddleware(restMux))
 
 	// The main routing handler
 	mux := http.NewServeMux()
-	mux.Handle("/rest/", restMux)
+	mux.Handle("/rest/", noCacheRestMux)
 	mux.HandleFunc("/qr/", s.getQR)
 
 	// Serve compiled in assets unless an asset directory was set (for development)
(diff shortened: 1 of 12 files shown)


AudriusButkevicius:
I think this is beating a dead horse. We should move to grpc for that and use grpc-web to generate rest side of that.

calmh:
Would that help though? I mean, just sprinkling grpc over something won't make us magically handle partial update requests and stuff like that, right, it's really just a matter of the wire protocol which seems orthogonal to this issue?

AudriusButkevicius:
The right way to fix this to address the tramping issue is the way k8s does it.

Resources (aka config) has a generation number, as you post it back, you keep the generation number so the backend can verify if you'll tramp something or not, and refuse your post.

Going to absurd levels of endpoints just makes this worse as you'll have people calling endpoints many times to change many settings, causing continuous restarts and strain on everything that happens to use the config.

I guess we could have more granular per folder per device grpc endpoints, and thats what I started working on, but not to absurd level where every config option has its own endpoint.

AudriusButkevicius:
I understand that you want something now, but I also see this as a great opportunity to build out the grpc utopia I came up with as part of this work, so I'd rather the effort was spent there oppose to beating a dead horse.

AudriusButkevicius:
I guess tldr, my stance is: don't think we should have endpoints per setting, we can have endpoints per entity, but that should be in grpc as I think of that as the future, and we should address tramping by generation numbers, oppose to silly granular endpoints.

There is a thing called fieldmask in proto which we could use to support patch behaviour and still not have tramping.

imsodin:
In my view config tramping/racyness is unrelated to this. This is only about convenience: Instead of handling/serializing a huge structure you get to handle just the thing you need to. Doing lots of config changes in quick succession is just as problematic with a single endpoint as with many endpoints.

And I get the point about the grpc-utopia. And I don't mean to go against that with this PR and I really don't think it does. I did not go directly with grpc, because just familiarizing myself with it takes time, not even thinking about polishing all the nits and grits to the point that it's usable. Extending the existing api is little effort with in my opinion clearly better UX.

AudriusButkevicius:
So if its not the tramping we want to address, why do we need this?

Seems like a high price to pay for some convenience? So whats the usecase where this is a "fly or die" feature?

imsodin:
High price being the additional code?

The use-case is: There's a huge cluster with a few orchestrated central nodes, where changes to devices (new ones, folder sharing) happen frequently. Meaning the config of those central nodes is huge. Thus it's nice not to have to juggle/serialize/... a huge blob every time a change happens. 

Not sure about how "fly or die"-ish this really is, but it feels like a sensible thing to want. And I yet fail to see the huge drawbacks/costs associated with this change that would outweigh the benefits, even if you consider the benefits to be small.

AudriusButkevicius:
> Thus it's nice not to have to juggle/serialize/... a huge blob every time a change happens.

I mean it's equally as nice not to have additional code/complexity that just handles something that is just "nice" but not really necessary.

Especially when that code lands on a pile of mud, ... which we'd like to untangle... so adding any additional code there makes the untangling in the future harder, as I/we can't guarantee that what you are adding now will be available in the new thing. 

And then we'll have that one user who will go "but, but, but, my workflow, it's all ruined".

Hence my point, these enhancements sound great, but they should be start of the new thing, so we know it's not going away when we've hit the utopia, and that one user is happy that we're not on the starting path of deprecating something that has been implemented just a month ago.

imsodin:
> I mean it's equally as nice not to have additional code/complexity that just handles something that is just "nice" but not really necessary.

Sure, improvement vs maintainability is always a trade-off. However in my obviously biased view, this doesn't introduce much complexity: It takes existing functionality on the entire config (take a request body, unmarshaling it on top of a struct and then committing that struct to config) and applies that to parts of the config. If we want to change the way we process/change configs, you'll have to change it for the existing functionality already. And that's going to be the hard thing. Once that's done, I don't see how using the same mechanism on subsections of the config increases complexity.

> Especially when that code lands on a pile of mud, ... which we'd like to untangle... so adding any additional code there makes the untangling in the future harder, as I/we can't guarantee that what you are adding now will be available in the new thing.

Is it really a question whether any new api will have the same monolith config get/post endpoint as the only way to control all of Syncthing instead of more granular control? In my mind that was one major requirement for anything new.

Also I don't agree on the pile of mud: If anything is muddy, it's lib/config and how we do changes and put them into effect. I change nothing about how that works (the changes to `config.Wrapper` only makes what is already there more consistent). I don't see us untangling/changing the current rest api, only eventually retire it. With emphasis on *eventually*: I don't think it makes sense at this point to prevent any improvement in the current api, just because there's an api with a better technology starting to emerge.

AudriusButkevicius:
The mud in the api comes not from the fact it's monolith or not monolith, it comes from the fact that there is no way to generate an api client, for any language, you always have to roll one by hand.

The fact that the API is poorly structured/monolith is the least of my worries.

When I mean untangling, I mean shipping v2 api with equivalent functionality of v1, hence growing v1 api surface will have an impact on what v2 will have to look like, which is what I want to avoid. 

As I said I think this could be a great opportunity to start work on v2, because we're adding these endpoints because of pain with v1, and the remedy to the pain could be in v2.

I understand you not wanting to work on v2 as it's new waters, but this feels like cutting corners just to get something shipped, as at some point someone will have to come back to this and redo it in v2, spending the effort twice, rather than once. But sure, you do you, even if we disagree.

I guess I can put a skeleton down for v2 we can build on, if this is your biggest concern doing it in v2.

imsodin:
One thing upfront: You are totally right that this is about getting something shipped. The v2 api might remedy everything, but I do not consider it something that will happen soon (as in within this year), neither would I want it to. We should be able to have the time to think about it and make good decisions. It's definitely not that I don't want to work on v2 at all. I was under the impression that you were considering or already doing such work, thus indeed didn't plan on starting myself, but if you want to talk about coordinating such an effort - lets do that.

In my understanding of your comments there's two main points: This increases the scope of any v2 api and it's effort spent in vain that would be better invested in v2.

As to effort to reach feature parity in v2: I probably didn't fully grok what v2 entails. The way I understood it, it's a solution that allows to generate api clients in various languages plus a better wire format. And that it still requires code in Go in Syncthing to interface between the generated API and our internals (config, model, whatever). If that were true, I again maintain that it isn't inherently harder to create a sensible solution to modify the entire config object vs parts of it. If there's a solution for the entire thing, I am confident I can create something for the parts (and I am willing to do so).

Effort spent in vain is somewhat moot on my side, it has happened, and now it's about whether there's someone willing to spend the effort to review it. I also don't think this PR is that much effort.

AudriusButkevicius:
So I did put a few bricks down on v2, but again, the amount of work I put in opensource is a anti-correlated with the amount of interesting stuff I get to do at work, and how much of a craving I have for my middle aged person video gaming addiction.

I guess if there was urgency, I could step it up, but I am not sure there is.

Anyways, even if I am sad that we keep building an ever growing tower of dung, especially when the argument for making a bigger dung pile is that something is "nice" oppose to something is "necessary" or "impossible" to do in the current api, the work is already done, so I guess I just have to close my eyes, and allow the shit to keep flowing.

calmh:
I'm not at all as sold on grpc and stuff as you Audrius and all this talk of "mud" and "piling on dung" and "absurd amounts" and whatnot is really grating and frankly bullshit in my view. There's nothing wrong with a simple REST API to GET and POST JSON objects. Sure, you need to hand craft your client, but come on, it's not precisely rocket science and we and the world have been doing it for ages.

However, at some point when the config grows it's rather annoying to have to treat it in its entirety all the time. We have concessions to this in the pause/resume endpoints that were initially "get your whole config, set a boolean, post the entire thing back again". Being able to patch the config in parts makes perfect sense to me, and even if we were to allow endpoints to dig down into each individual little value that would be essentially one endpoint with data in the path, not an "absurd amount of endpoints".

I've yet to see anything to indicate that the grpc+rest-gateway utopia improves any of this in a way that matters so I'm not sure why it would be a blocker to wait for it?

I'm also not getting if you're saying the current API is somehow crap and should be frozen for further changes and then thrown out in its entirety. If you are, this is something I disagree significantly with.

AudriusButkevicius:
Yes, I am saying the current API is crap, it's crap on a few different planes:

1. You have to hand-roll clients. Sorry to drop your baby on the floor, but what's the point of an api if it's close to impossible to consume. You have to spend hours creating classes that map 1 to 1 to the go code, and then PRAY you haven't missed or misstyped something, pray that we don't add something new that is crucial to set, or otherwise we might suddenly cause syncthing to bulldoze your data, like it was a case in android app, where we didn't update the java classes, and upon saving the config via the app, unset some setting which did some form of boo-boo. If that is not a definition of shit API, I am not sure what is... 

2. The API is monolithic


I am not saying we should now freeze this API and never add anything to it again because of that.
We will add new features that will need new endpoints `POST /rest/folder/gosuperfast`, and that's fine, as long as there is no alternative way of doing that.

I am however **against** adding new "comfort features", because they are "nice to have".

This PR tries to address point two, which is a valid pain point, and yes, we should have more granular endpoints, but by adding this comfort feature we are making v1 api more bearable, which will deincentivize people to migrate to v2, where bearability was supposed to be a great selling point.

I would have preferred for this change to have been a start of v2, where the bearability/comfort features like that are tought through and available, making an incentive to start an early (even if partial) migration for consumers.

The GRPC stuff will look exactly like your rest stuff now, less monolithic, more granular, but with a sensible openapi spec that people can generate a client from, so the whole "post objects via curl is cool" argument is moot.

I guess we're not all waving the same flag here, so I guess I am surprised that my go to proto changes made it through, if we're still strong believers in the success via expansion of the existing api.

Anyways, this is going nowhere, the work is already done, democracy wins, I'll shut up.

calmh:
I think I'm genuinely not really understanding the problems and solutions the same way you do. This might be because I don't use grpc and friends enough. My understanding:

1. Handrolling the config classes on the client is bad. When we add an attribute it gets lost in the roundtrip. (I fully agree with you here.)
2. We now have the config description in protobuf. So you don't need to hand roll the classes any more, you can generate them from the protobuf schemas and use JSON serialization. Sweet, point 1 is essentially solved. You seem to be beating a dead horse here.
3. Not re-generating the client classes based on our protobuf schemas for every release means you risk losing new attributes in the roundtrip.
4. Using grpc or whatever changes nothing about point 3, because the proto-binary serialization is essentially the same as the JSON one in this respect.

Using grpc gives you a schema definition of the endpoints in addition to the data, but really, implementing the endpoint as a pair of GET/POST methods with a well defined schema seems like quite a small part of *using* any new endpoint we might introduce.

So I get you think the current API is shit, and I disagree and would rather not wake up every morning to hearing it. Even so, by all means suggest an actual improvement but saying it should *remain* shit to make your future changes more palatable is, again, bullshit in my view.

calmh:
One more point about API consumability and grpc etc. In my experience few API users are "serious" enough to warrant the schemas and generated classes etc. Off the top of my head I can think of three (android, trayzor, gtk). In the wild API consumers I see all the time are things like

- powershell script that polls the completion endpoint and alarms when it's < 100%
- bash + curl to grab /connections and count them, feeding into a time series system
- bash + curl to check for local modifications and post to the revert endpoint

I fully expect this sort of integration vastly outnumbers the "big" ones, and we'll never hear about most of them. 

Our API is totally well consumable for this purpose, except for things like config changes that become cumbersome because it's a complex structure to get, modify, post back *when not even using a real programming language*. Making it simpler to do those kinds of changes will make life easier for a bunch of garden variety mini integrations. Adding grpc schemas etc won't do anything positive at all for these integrators...

AudriusButkevicius:
> We now have the config description in protobuf. So you don't need to hand roll the classes any more, you can generate them from the protobuf schemas and use JSON serialization. Sweet, point 1 is essentially solved. You seem to be beating a dead horse here.

Except that it's not. We don't use proto to json serialization, we use go struct with it's magical tags to json serialization, so even if we have proto contracts, nobody can actually use them, not until we have a new API that actually uses them.

> Not re-generating the client classes based on our protobuf schemas for every release means you risk losing new attributes in the roundtrip.
> Using grpc or whatever changes nothing about point 3, because the proto-binary serialization is essentially the same as the JSON one in this respect.

That's not true, atleast in my head and my vision, as the idea is that we'd have patch support using proto's FieldMask.
The grpc web stuff actually generates the field mask for you, so if you post `{"name": "foo"}` for a device via rest, it's able to just merge in that once field on the existing device struct without tramping stuff.

Also, binary proto (and maybe even json, I have not checked) supports carrying/preserving unknown fields, so even if the client doesn't understand that field, it's not actually lost on a round trip.

> Using grpc gives you a schema definition of the endpoints in addition to the data, but really, implementing the endpoint as a pair of GET/POST methods with a well defined schema seems like quite a small part of using any new endpoint we might introduce.

Again, maybe I am spoiled by what I have at work, but I never ever had to do that. I never ever had to write any code by hand to access any API. The client code, including the code that knows what endpoints are available and what parameters they take and what not is always provided to me by the generated client. I guess having to craft the URL and know which struct to post is not rocket science, but there is definately a gap in discoverability, and it's pretty frequent that people forget that they need to pass a few query params to get things to work. Or simply can't find the method they are looking for. So sure, it's usable, but far from providing a great developer/user/consumer experience.

> One more point about API consumability and grpc etc. In my experience few API users are "serious" enough to warrant the schemas and generated classes etc. 

I think you are wrong here, because if that was the case, nobody would have bothered to spend their time on:

https://github.com/JodusNodus/node-syncthing
https://www.npmjs.com/package/syncthing-node
https://github.com/blakev/python-syncthing
https://github.com/retgoat/syncthing-ruby
https://github.com/tokkenno/syncthing-dotnet
https://github.com/cskwrd/FlyTops.Syncthing
https://github.com/jaycefayne/syncthing-rs
https://github.com/jetho/syncthing-hs
https://github.com/burnnat/syncthing.dart
https://share.zabbix.com/cat-app/zabbix-syncthing

I don't work in the syncthing consulting business, so I can't say what people actually use and what they don't but I suspect what you see is also not the full picture, because I assume these things exist because someone needed them, and if they exist, there are probably people using them, you just haven't had an opportunity to work with users of these.

Someone spent time doing the work because of the choices that we're made a while ago on how we expose an API.
I am saying the choice is bad, and these libraries should not need to exist, or not in their current form where people hand craft urls (which if I am reading you correctly is ok in your books, but not in mine I guess).

> I fully expect this sort of integration vastly outnumbers the "big" ones, and we'll never hear about most of them.

Sure, I agree, but that's not because people think bash + sed is the rock and roll, it's because we provide nothing better.
If there was a cli easily available, they would use that, but it's not easily available.

> Even so, by all means suggest an actual improvement but saying it should remain shit to make your future changes more palatable is, again, bullshit in my view.

Things should not stay shit, we should improve things, but we should make the improvements in the thing we believe will be the future, not in the thing we believe we will want to start deprecating.

But as I said before, I am not sure if we're rocking the same flag here, as I feel like I am the only one here who thinks that hand crafting urls and bash + sed is not rock and roll at all.

lib/api/confighandler.go
```diff
@@ -0,0 +1,348 @@
...
+	c.HandleFunc(configBase+"folders/", c.handleFolder)
+	c.HandleFunc(configBase+"devices", c.handleDevices)
+	c.HandleFunc(configBase+"devices/", c.handleDevice)
+	c.HandleFunc(configBase+"options", c.handleOptions)
+	c.HandleFunc(configBase+"ldap", c.handleLDAP)
+	c.HandleFunc(configBase+"gui", c.handleGUI)
+
+	// Legacy
+	c.HandleFunc("/rest/system/config", c.handleConfig)
+	c.HandleFunc("/rest/system/config/insync", c.handleConfigInsync)
```
AudriusButkevicius, line 52: I guess this needs to be duplicated here and in addConfigHandler?

AudriusButkevicius, line 52: I guess this needs to be here and in addConfigHandler?

lib/api/confighandler.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+func (c *configHandler) handleConfig(w http.ResponseWriter, r *http.Request) {
+	if !checkMethod(w, r) {
+		return
+	}
+	if r.Method == http.MethodGet {
+		sendJSON(w, c.cfg.RawCopy())
+		return
+	}
```
AudriusButkevicius, line 64: I wonder if there couldn't be some abstraction that makes this easier to read
```
type HttpHandler struct {
   Get http.Handler
   Post
   Patch
}
```

then you do 

something like
```
m.HandleFunc("devices", fromHandler(HttpHandler{
    Get: withExtractFolderId(func(foo bar) { logic })
   Post: withExtractFolderId(func(bar, nuff) { logic })
})

func withExtractFolderIf(delegate func(folder, w, r)) func(w, r) {
    return func(w, r) {
         fid := extractFolderId(r)
         delegate(fid, w, r)
    }
))
``` 
any function pointer that's not set returns method not allowed.

Now there is a bunch of `if method == get: sentJson(); return` 

Or I mean, there are frameworks for route and method handling, just use a framework instead of reinventing the wheel, of stripping paths to extract folder ids etc.

AudriusButkevicius (approved):
So I tried to review this, but got dizzy trying to maintain a mental stack of which branches of the handler which http method hits.

Anyways, we're not using this, so YOLO.

lib/api/confighandler.go
imsodin: I didn't fully get to something just like you propose, but I tried to go into that direction - is readability any better? And used a framework, which is need indeed, I usually am just vary to add a dependency for a tiny usage like this.

lib/api/confighandler.go
```diff
@@ -0,0 +1,375 @@
...
+func (c *configHandler) registerConfigInsync(path string) {
+	legacyPath := "/rest/system/config/insync"
+	handler := func(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
+		sendJSON(w, map[string]bool{"configInSync": !c.cfg.RequiresRestart()})
+	}
+	c.GET(path, handler)
+	c.GET(legacyPath, handler)
+}
+
+func (c *configHandler) registerFolders(path string) {
```
AudriusButkevicius, line 80: It would make more sense if the path constants were declared/constructed in these methods, the only reason they are done outside is because of the base path, which can either become a prop of the handler, or be passed in.

This is  to be able to understand whether `/folder/:id` or `/folders` is used for this method.

lib/api/confighandler.go
```diff
@@ -0,0 +1,375 @@
...
+		}
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		c.finish(w, waiter)
+	}
+	c.POST(path, handler)
+	c.PUT(path, handler)
+}
```
AudriusButkevicius, line 124: According for rest-fu, PUT/POST should be used for singular resources.

I guess following the rest logic for PUT and POST I would interpret this as:
PUT - Update the folder list to look like this (adding/removing/modifying folders)
POST - Create the folders in the array?


I think this is somewhat gray ground, so I am not sure if we want to support PUT/POST for collection endpoints all together, but PUT and POST doing the same is definately not cool.

imsodin: I used both because from what I read `PUT` is right for our use-cases (e.g. all the operations are idempotent), but we still did use `POST` on the previous config endpoint. However I'll just keep `POST` on the anyway legacy `/rest/system/config` and use `PUT` everywhere else exclusively.

AudriusButkevicius (dismissed):

imsodin:
Saw some debug leftovers to clean-up, sorry for the review dismiss.

And as we now have github.com/julienschmidt/httprouter, we might as well use that for the other endpoints as well. Then there's no more need for the `getPostHandler` and `addConfigHandler` detours anymore: https://github.com/imsodin/syncthing/commit/0a6fdb125bef9a96d9659dce6539c3bfa5e3754f  
What do you think?

AudriusButkevicius (dismissed):
Again, so I am not looking at this in great details as I think it's a lost cause to some extent, but I guess it looks fine on the surface.

lib/api/api.go
```diff
@@ -309,7 +303,7 @@ func (s *service) serve(ctx context.Context) {
 
 	// A handler that splits requests between the two above and disables
 	// caching
-	restMux := noCacheMiddleware(metricsMiddleware(getPostHandler(getRestMux, postRestMux)))
+	restMux := noCacheMiddleware(metricsMiddleware(s.addConfigHandler(getPostHandler(getRestMux, postRestMux))))
```
calmh, line 306: Hiding the addConfigHandler under the caching rubric is sneaky. I'm unsure it even needs to be a method instead of being a section under the allready long list of handlefuncs? (I think there's some value in the list of endpoints being in one place in a long list.)

lib/api/api.go
```diff
@@ -446,6 +440,16 @@ func (s *service) CommitConfiguration(from, to config.Configuration) bool {
 	return true
 }
 
+func (s *service) addConfigHandler(handler http.Handler) http.Handler {
+	mux := http.NewServeMux()
+	configHandler := newConfigHandler(s.id, s.cfg)
+	mux.Handle(configBase, configHandler)
+	mux.Handle("/rest/system/config", configHandler)        // Legacy
+	mux.Handle("/rest/system/config/insync", configHandler) // Legacy
```
calmh, line 448: "Legacy" seems a bit harsh since it's actively used by the GUI

lib/api/confighandler.go
```diff
@@ -0,0 +1,369 @@
+// Copyright (C) 2014 The Syncthing Authors.
```
calmh, line 1:
```suggestion
// Copyright (C) 2020 The Syncthing Authors.
```

lib/api/confighandler.go
```diff
@@ -0,0 +1,369 @@
...
+
+	"github.com/julienschmidt/httprouter"
+	"golang.org/x/crypto/bcrypt"
+
+	"github.com/syncthing/syncthing/lib/config"
+	"github.com/syncthing/syncthing/lib/protocol"
+	"github.com/syncthing/syncthing/lib/sync"
+)
+
+const configBase = "/rest/config/"
```
calmh, line 23: This feels like it should be a parameter to newConfigHandler, like it is to the other HandleFoo methods that register handlers for things?

calmh: Having the "legacy" paths hardcoded in there isn't super pretty either. How about exposing at least the legacy operations as handler funcs on the configHandler and registering them like normal in the api service?

calmh (commented):
Just first look, will need to take closer look at the actual put/post/patch/delete stuff since I'm currently on an awkward device, but I promise to not take too long to do so

calmh:
Did you consider a more general reflect- or JSON-based approach to the sub-operations? That is, instead of specifically handling folders and devices and stuff, instead trying to take the path from the request as an object path inside the config and applying the relevant DELETE/PATCH/PUT on that location with the object directly? It could be that you did and this would be horrible and cumbersome, or it could be the that it isn't and fulfils the requirement in a smaller way...

imsodin:
I considered a generic implementation with reflect, but considered the complexity (and the increased chance of buggyness coming with that) not worth the benefit. It couldn't be fully generic, there would have to be special casing for accessing items by a key/id (e.g. .../folder/id). And json and go names would need translation. Restricting rest paths to top-level config elements and folders/devices by id seems like a good compromise between granularity and implementation simplicity.

lib/api/api.go
imsodin: I changed the GUI to use the new endpoints and changed the labelling to "deprecated".

imsodin:
I am now using `httprouter.Router` for the existing rest endpoints as well and thus moved the registering such that all rest paths are in the "long list" (including the legacy/deprecated ones).

lib/api/confighandler.go
```diff
@@ -0,0 +1,336 @@
...
+
+func (c *configMuxBuilder) registerFolders(path string) {
+	c.GET(path, func(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
+		sendJSON(w, c.cfg.FolderList())
+	})
+	c.PUT(path, func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+		c.mut.Lock()
+		defer c.mut.Unlock()
+		var waiter config.Waiter
+		folders := make([]config.FolderConfiguration, 0)
```
calmh, line 62: I don't think you need to `make` it? `var folders []config.FolderConfiguration` should be sufficient to unmarshal into. (Applies below multiple times as well)

lib/api/confighandler.go
```diff
@@ -0,0 +1,336 @@
...
+		err := unmarshalTo(r.Body, &folders)
```
calmh, line 63: Error here should really result in `http.StatusBadRequest` rather than internal server error I think. (Applies to all unmarshals.)

lib/api/confighandler.go
calmh: I think this is old and resolved, but my understanding of things is that if `/config/devices/` is a collection, POST to `/config/devices/` could add a new device, while PUT to `/config/devices/$someDeviceID` would change or add the device with that specific ID. PUT to `/config/devices/` to replace the entire collection of devices seems kosher.

lib/api/confighandler.go
```diff
@@ -0,0 +1,336 @@
...
+		if !ok {
+			http.Error(w, "No folder with given ID", http.StatusNotFound)
+			return
+		}
+		sendJSON(w, folder)
+	})
+	c.PUT(path, func(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+		c.adjustFolder(w, r, config.FolderConfiguration{})
+	})
+
```
calmh, line 108: Spacing inconsistency OCD triggered. I think we should have a space between all the method registrations...

lib/api/confighandler.go
```diff
@@ -0,0 +1,336 @@
...
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	c.finish(w, waiter)
+}
+
+func (c *configMuxBuilder) adjustGUI(w http.ResponseWriter, r *http.Request, newGUI config.GUIConfiguration) {
+	c.mut.Lock()
+	defer c.mut.Unlock()
+	oldGUI := c.cfg.GUI()
```
calmh, line 276: I think the thing with newGUI/oldGUI here is confusing given that newGUI is in fact oldGUI and we're grabbing a *new* oldGUI just to have a cache of the password? Better to just grab the password before unmarshal and pass that explicitly to checkGUIPassword (who doesn't really need to full config structs)?

lib/api/confighandler.go
```diff
@@ -0,0 +1,336 @@
...
+
+type configMuxBuilder struct {
+	*httprouter.Router
+	id  protocol.DeviceID
+	cfg config.Wrapper
+	mut sync.Mutex
+}
+
+func (c *configMuxBuilder) registerConfig(path string) {
+	c.GET(path, func(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
```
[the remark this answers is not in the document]
imsodin: Unfortunately `params` in the context of httprouter aren't url parameters, but path components like id in `../devices/:id`.

I don't really like the all caps method names anyway, so I'll switch to `Handle...`

lib/api/confighandler.go
calmh: This tiny thing remains and I still think it matters. :) (Just in case you missed it since everything else looks fine)

imsodin: Then I don't get it. I now pass passwords directly to `checkGUIPasswords` and query the wrapper directly for password in config right now (active, old) and pass the unmarshalled password as the new password.

calmh: Oh I was for some reason looking at an old version, but I have a suggestion for the new version as well, hang on

lib/api/confighandler.go
```diff
@@ -0,0 +1,377 @@
...
+	err := unmarshalTo(r.Body, &gui)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	if gui.Password, err = checkGUIPassword(c.cfg.GUI().Password, gui.Password); err != nil {
+		l.Warnln("bcrypting password:", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
```
calmh, line 329:
```suggestion
	oldPassword := gui.Password        
	err := unmarshalTo(r.Body, &gui)
	if err != nil {
		http.Error(w, err.Error(), http.StatusBadRequest)
		return
	}
	if gui.Password, err = checkGUIPassword(oldPassword, gui.Password); err != nil {
		l.Warnln("bcrypting password:", err)
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}
```

calmh: Eh it doesn't do the diff nicely, but what I means was just to cache the password in oldPassword instead of calling on cfg.GUI() once more

calmh (commented):
LGTM I think?

calmh (approved):

PR #7006, opened 2020-09-26
Fix translation errors

shellus:
### Purpose

Fix translation error


AudriusButkevicius:
Translations are managed on transifex

PR #7009, opened 2020-09-26
all: Move remaining protos to use the vanity plugin

AudriusButkevicius:
As part of sketching out the new API, I'm trying to avoid having to declare messages twice, which means that I'd like to expose db.Counts/protocol.FileInfoTruncated etc, in which case it makes sense for them to move under the vanity plugin.

I also address some poor naming of types and variables.
The proto standard says that we should always use snake_case as well as all enum values should be prefixed with the enum name (because in C enums are globals, so two things delcaring UNKNOWN would clash).

Also, because the vanity plugin does int32 -> int, a few types changes, but the changes should be non-lossy, as we're only doing int32 -> int, but not int64


AudriusButkevicius:
So it's strange that I had to fix the fuzz tests but following what was already there it sort of made sense, hence not sure how it  was not broken before to be honest.

calmh:
So having the vanity stuff is nice, but

> I'd like to expose db.Counts/protocol.FileInfoTruncated

makes me tingle all over. These are internal implementation specifics, suddenly leaking them into an API feels quite wrong.

AudriusButkevicius:
Sure, fine, we'll have copies.

proto/scripts/protoc_plugin.go
```diff
@@ -34,11 +34,11 @@ func main() {
...
 	vanity.ForEachFile(files, vanity.TurnOffGoUnkeyedAll)
 	vanity.ForEachFile(files, vanity.TurnOffGoSizecacheAll)
-	vanity.ForEachFile(files, vanity.TurnOffGoEnumStringerAll)
+	vanity.ForEachFile(files, vanity.TurnOnMarshalerAll)
+	vanity.ForEachFile(files, vanity.TurnOnUnmarshalerAll)
 	vanity.ForEachEnumInFiles(files, HandleCustomEnumExtensions)
 	vanity.ForEachFile(files, SetPackagePrefix("github.com/syncthing/syncthing"))
-
-	vanity.ForEachMessageInFiles(files, HandleCustomExtensions)
+	vanity.ForEachFile(files, HandleFile)
```
imsodin, line 41: What's the intention of this change? On the surface it looks like it doesn't actually change anything, just calls `ForEachMessageInFiles` on single files instead of all of them at once. And it makes the diff of `HandleCustomExtensions` almost impossible to read due to the indentation change.

AudriusButkevicius: So we now pass the file ref to HandleCustomExtensions and in the handling of (device_id) =  true do not prepend the type with package name if the file we are processing is in the same package. This is to prevent lib/protocol from importing itself.

AudriusButkevicius: Also, if you prepend `?w=1` to the url (or set it via the settings cog) you can ignore whitespace changes which makes indent changes a breeze to look at.

imsodin: Wow, that looks much nicer now (and makes the changes very clear) - thanks for the tip :)

imsodin (dismissed):
I guess fingers crossed that nothing breaks.

imsodin (approved):

AudriusButkevicius:
Needs an approval due to merge conflicts.

calmh (approved):

PR #7011, opened 2020-09-28
lib/fs: Be more clear about invalid file names (ref #7010)

calmh:
Add specific errors for the failures, resulting in this rather than just
the generic "invalid filename":

```
[MRIW7] 08:50:50 INFO: Puller (folder default, item "NUL"): syncing: filename is invalid: name is reserved
[MRIW7] 08:50:50 INFO: Puller (folder default, item "fail."): syncing: filename is invalid: name ends with space or period
[MRIW7] 08:50:50 INFO: Puller (folder default, item "sup:yo"): syncing: filename is invalid: name contains reserved character
[MRIW7] 08:50:50 INFO: default: Failed to sync 3 items
```

### DIFF
--- a/lib/fs/util.go
+++ b/lib/fs/util.go
@@ -15,7 +15,11 @@ import (
 	"strings"
 )
 
-var errNoHome = errors.New("no home directory found - set $HOME (or the platform equivalent)")
+var (
+	errInvalidSpacePeriod  = errors.New("name ends with space or period")
+	errInvalidReservedName = errors.New("name is reserved")
+	errInvalidReservedChar = errors.New("name contains reserved character")
+)
 
 func ExpandTilde(path string) (string, error) {
 	if path == "~" {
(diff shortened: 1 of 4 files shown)


lib/fs/util.go
```diff
@@ -15,7 +15,11 @@ import (
 	"strings"
 )
 
-var errNoHome = errors.New("no home directory found - set $HOME (or the platform equivalent)")
+var (
+	errInvalidSpacePeriod  = errors.New("name ends with space or period")
+	errInvalidReservedName = errors.New("name is reserved")
+	errInvalidReservedChar = errors.New("name contains reserved character")
```
AudriusButkevicius, line 21: I guess the use of word reserved is a bit vague. I'd say name contains charaters not permitted on Windows, or file name is not permitted on windows or something.

calmh: I refactored the errors a bit to be more clear, un-exported the things as they're not used outside of lib/fs currently

calmh: Exact wording is a bit difficult because the "name" is actually a "path component" but then it gets very technical

AudriusButkevicius (approved):

PR #7013, opened 2020-09-28
lib/fs: Add COM0 and LPT0 to invalid Windows file names

tomasz1986:
COM0 and LPT0 are not listed in the official Microsoft's documentation
at https://docs.microsoft.com/windows/win32/fileio/naming-a-file, but
in reality are also invalid in Windows Explorer.


tomasz1986:
These two are actually a little different from the others, because CMD can operate on them (without relying on the `\\?\` syntax, but Explorer still treats them as "invalid", so I think that they should be added to the list.

calmh (approved):

PR #7014, opened 2020-09-29
lib/model: Auto-resolve identical case-conflicts (fixes #6989)

### LINKED ISSUE #6989: Out of Sync error after case rename while ignored
How to reproduce:

1. Sync an SR folder between Device A and B.
2. Add a `test` file to the folder and let it sync.
3. Ignore `*` on Device B.
4. While ignored, rename `test` on Device B to `TEST`.
5. Unignore `*` on Device B.

image
image

imsodin:
This does what was proposed in #6989: When at the end of syncing a file there's a case-conflict, and the conflicting files have the same content, just keep one of the two names instead of throwing an error. This only handles files and case conflicts in the last path component, as directories are a can of worms: We'd have to check the equality of an entire filesystem sub-tree.

Extended a unit test to cover both the scenario with differing and equal contents on case-conflict.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -1558,7 +1558,37 @@ func (f *sendReceiveFolder) performFinish(file, curFile protocol.FileInfo, hasCu
 			return err
 		}
 	} else if !fs.IsNotExist(err) {
-		return err
+		// If there's a case conflict, check if the conflicting files
+		// have the same content and "merge" (choose one, delete the other).
+		var caseErr *fs.ErrCaseConflict
+		if !errors.As(err, &caseErr) {
+			return err
+		}
+		// We can't deal with case conflict in parents (would have to check
+		// equality of the entire filesystem subtree).
+		if filepath.Dir(file.Name) != filepath.Dir(caseErr.Real) {
+			return err
+		}
+		// Check that the case conflict is up-to-date in db
+		stat, statErr := f.fs.Lstat(caseErr.Real)
+		if statErr != nil {
+			return statErr
+		}
+		other, ok := snap.Get(protocol.LocalDeviceID, caseErr.Real)
+		if scanErr := f.scanIfItemChanged(other.Name, stat, other, ok, scanChan); scanErr != nil {
+			return scanErr
+		}
+		if other.FileType() != file.FileType() || !other.BlocksEqual(file) {
+			return err
+		}
+		// Arbitrary but deterministic way of choosing one of the two names.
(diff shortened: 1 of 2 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -1558,7 +1558,37 @@ func (f *sendReceiveFolder) performFinish(file, curFile protocol.FileInfo, hasCu
...
+			return err
+		}
+		// Arbitrary but deterministic way of choosing one of the two names.
+		if other.Name > file.Name {
+			file.Name = other.Name
+		}
+		if err := f.inWritableDir(f.fs.Remove, other.Name); err != nil {
+			return err
+		}
+		scanChan <- other.Name // Scan the deletion
```
calmh, line 1591: I don't fully understand what happens here, can you explain? We decide to retain either the old or new name, potentially replacing the name of the thing we are syncing. We remove the file under old name, and queue a scan. We then continue syncing, potentially putting the file into the old name (if we chose that) and updating the database under that name.

It seems to me that this is essentially a no-op, in that we won't get a database update for the thing we were syncing (we changed the name) and we replaced the file that was there with an identical one under the same name, and we're back to square one?

imsodin:
The current implementation introduces new bugs (see below), thus closing for now. If there's a viable solution to that, I'll reopen.

Writing more tests and thinking on this, I believe that "merging" (choosing one name arbitrarily) identical, case-conflicting files is not safe. Consider e.g. the following scenario with two case-insensitive devices A and B:

1. Both devices have file "foo" in sync.
2. Device A renames "foo" to "Foo", and for reasons the rename shortcut doesn't take effect (e.g. deletion is sent later)..

Current result:  
When pulling on device B in the first iteration there's a case-conflict error, then the deletion of "foo" happens and on the second iteration everything gets back in sync with "Foo".

Result with "merging":  
When the creation of "Foo" comes in, device B notices the case conflicts, sees that the file is the same as "foo", merges the files into "foo", deleting "Foo". Later "foo" is also deleted.

This problem might be prevented by checking deletions before merging and bumping the version of the merged file. However in my opinion that's way too much complexity (and potential to introduce bugs) consider the gain.

lib/model/folder_sendrecv.go
imsodin: You are perfectly right, changing the name of the fileinfo does nothing respectively the wrong thing. I corrected it, but thinking more about it there's still unresolved problems that let me doubt whether this is a good idea at all, see https://github.com/syncthing/syncthing/pull/7014#issuecomment-701277474

PR #7018, opened 2020-09-30
lib/model, lib/protocol: Send ClusterConfig on config change (fixes #7020)

### LINKED ISSUE #7020: Do not disconnect when changing folder settings
Currently we disconnect all related devices when changing folder settings, which then take effect once connecting again.

imsodin:
### Purpose

Instead of dropping a connection when the config changes, a new cluster config is generated and sent.

On the receiving end this means we need to track index senders to reset them when a new cluster config comes in. There was a commit on the progress emitter that it breaks on multiple CCs. I couldn't figure out though what is supposed to break, the operation has the exact same effect if called on a newly established connection or a super old one.

On the sending side sending of cluster configs happens in the toplevel function now. Otherwise we might send multiple CCs to the same device, which is wasteful. We did wait on the connection being closed previously, which we obviously don't anymore. This was done to prevent incoming indexes to happen after folder objects are removed from the model - I think we previously panicked then. Now if that happens we just drop the connection with an error, which is fine.

### Testing

Adjusted a few tests either to check that a CC is sent instead of connection dropped or to drop a device instead of changing a folder to actually prompt the connection to be closed.  
Also did non-intensive testing on the test setup, i.e. pause and unpause folders verifying that the connection doesn't drop.

### Documentation

Needs a change to the BEP to remove the "must not" on sending multiple cluster configs.

### DIFF
--- a/lib/model/fakeconns_test.go
+++ b/lib/model/fakeconns_test.go
@@ -35,6 +35,7 @@ type fakeConnection struct {
 	indexFn                  func(context.Context, string, []protocol.FileInfo)
 	requestFn                func(ctx context.Context, folder, name string, offset int64, size int, hash []byte, fromTemporary bool) ([]byte, error)
 	closeFn                  func(error)
+	clusterConfigFn          func(protocol.ClusterConfig)
 	mut                      sync.Mutex
 }
 
@@ -91,7 +92,13 @@ func (f *fakeConnection) Request(ctx context.Context, folder, name string, offse
 	return f.fileData[name], nil
 }
 
-func (f *fakeConnection) ClusterConfig(protocol.ClusterConfig) {}
+func (f *fakeConnection) ClusterConfig(cc protocol.ClusterConfig) {
+	f.mut.Lock()
+	defer f.mut.Unlock()
+	if f.clusterConfigFn != nil {
+		f.clusterConfigFn(cc)
+	}
+}
 
 func (f *fakeConnection) Ping() bool {
 	f.mut.Lock()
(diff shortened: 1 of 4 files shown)


AudriusButkevicius (approved):
Looks good, yet I have fear of race conditions.

I guess it'd be nice to have some sort of stress test that causes a avalanche of these messages, and verify we're not leaking routines, overwriting variables out of order, and handle things like "need to send cluster config but the device just disconnected or is about to have it's connection closed" and so on.

calmh (approved):
I couldn't make it misbehave by frenetically pressing the pause/resume all button on one side, so that's cool. (One side did get stuck showing the other out of sync, but a GUI reload fixed it so I think it's a GUI or event issue...)

There's a tiny wrinkle in that setting a device to be introducer doesn't take immediate effect, because it doesn't make them send us a new CC and we don't drop the connection. Once they make any change though it does take effect on the other side immediately though which is also cool. (Or when the connection is re-established for whatever reason.)

calmh:
One thing though, this *really* needs a corresponding enhancement issue for the changelog before merging.

imsodin:
Created #7020 

> There's a tiny wrinkle in that setting a device to be introducer doesn't take immediate effect, because it doesn't make them send us a new CC and we don't drop the connection. Once they make any change though it does take effect on the other side immediately though which is also cool. (Or when the connection is re-established for whatever reason.)

That's an additional challenge: It's similar for removed ignored folders, where I left it at dropping the connection to get a new CC from the remote. We'd need a new message to request that the remote resends the CC for that. However that's tricky for backwards compat, as it won't drop the connection if the remote doesn't know that new message. The only backwards compatible solution I can think of would be sending the new CC request message, then wait for some time for the CC to come in, and if nothing comes, drop the connection - and that's ugly and slow.

calmh:
We could still drop the connection on changes that require us to retroactively reprocess CC messages due to local policy changes... Or, for that matter, cache it and reprocess it in those cases. But I think the introducer wrinkle is acceptable for now, perhaps just should be noted in an issue.

Issue #3930, opened 2017-01-24
Problems when folder and/or home disk are on extremely large filesystem?


josefsachsconning:
(when `Minimum Free Disk Space` in folder advanced settings is set to default value of 1)
```
2017-01-24 13:34:16: Stopping folder "notes" (uxhud-2egzw) - folder has insufficient free space
```

(when Actions > Advanced > Options > minHomeDiskFreePct is set to default value of 1)
```
2017-01-24 13:42:53: Stopping folder "notes" (uxhud-2egzw) - home disk has insufficient free space
```

Works okay as long as I change both values to 0.
Here is filesystem information:
```
$ df .
Filesystem                                              1K-blocks     Used        Available Use% Mounted on
fs-whatever.efs.us-east-1.amazonaws.com://foobar 9007199254740992 44992512 9007199209748480   1% /home/foobar
```

### Version Information

Syncthing Version: v0.14.20, Linux (64 bit)
OS Version: Ubuntu 14.04

calmh:
Large, and more than 99% full. Fractional values are also allowed, such as 0.1 if you want to go up to 99.9% full. I don't see an issue here except possibly with the default value?

josefsachsconning:
It looks almost empty to me.

calmh:
Yeah, my first comment was incorrect because I didn't read properly and I deleted it. Not quick enough I guess. :)

9223372036854775807 is the maximum int64 value (2^63). That filesystem is 9007199254740992*1024 = 9223372036854775808 bytes, or thereabout, with rounding. In other words it's possible it exceeds an int64.

josefsachsconning:
I would have guessed that it would use an unsigned int.

calmh:
It might, but we don't.

AudriusButkevicius:
I think you can just set minHomeDiskFreePct  to 0?

calmh:
Yes, even says so in the initial report. Lets not torment this guy any more by not reading his issue :)

AudriusButkevicius:
This is what happens when a report contains more than a few lines :(

Sorry.

josefsachsconning:
Not complaining, really appreciate the help, but let me know if you prefer less information in future reports (if any).

calmh:
So the statvfs API gives us size of blocks and number of blocks. It looks like this filesystem synthesizes it's size to be exactly 2^63. That exceeds int64 when we do the multiplication, so we get a negative total size and things fall apart from there.

[...]

PR #7019, opened 2020-10-01
lib: Use uint64 for disk stats (ref #3930)

imsodin:
We use https://github.com/shirou/gopsutil to get disk stats and they expose the values as uint64 (https://github.com/shirou/gopsutil/blob/master/disk/disk.go#L11), which we then convert to int64 for the filesystem interface. It's a minor change on our side and filesystems pretending to be just as large as an int64 are apparently a thing (see https://github.com/syncthing/syncthing/issues/3930#issuecomment-701983666 and issue linked just above that comment). For anything larger https://github.com/shirou/gopsutil would have to use floats or different units in their API (we are already using floats for any further calculations).

### DIFF
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -245,11 +245,8 @@ func (f *FolderConfiguration) CheckAvailableSpace(req uint64) error {
 	if err != nil {
 		return nil
 	}
-	usage.Free -= req
-	if usage.Free > 0 {
-		if err := CheckFreeSpace(f.MinDiskFree, usage); err == nil {
-			return nil
-		}
+	if !checkAvailableSpace(req, f.MinDiskFree, usage) {
+		return fmt.Errorf("insufficient space in %v %v", fs.Type(), fs.URI())
 	}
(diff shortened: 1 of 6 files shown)


AudriusButkevicius (approved):

lib/config/folderconfiguration.go
```diff
@@ -235,7 +235,7 @@ func (f *FolderConfiguration) SharedWith(device protocol.DeviceID) bool {
 	return false
 }
 
-func (f *FolderConfiguration) CheckAvailableSpace(req int64) error {
+func (f *FolderConfiguration) CheckAvailableSpace(req uint64) error {
```
calmh, line 238: This breaks the check on L248-249, which is why we avoid unsigned integers to begin with. Change might still be desired, but requires careful auditing everywhere the values are used in any kind of arithmetic

calmh (changes requested):

lib/config/folderconfiguration.go
imsodin: Thanks for catching. Corrected and checked that `fs.Usage` values aren't involved in any arithmetics anywhere else.

calmh (approved):

Issue #7026, opened 2020-10-06
panic: key not found when repairing DB


imsodin:
https://sentry.syncthing.net/share/issue/b29b701f48f64fcf8ce06ef60e3cc765/

```
github.com/syncthing/syncthing/lib/db/lowlevel.go in func1 at line 818
github.com/syncthing/syncthing/lib/db/lowlevel.go in getMetaAndCheck at line 833
github.com/syncthing/syncthing/lib/db/lowlevel.go in loadMetadataTracker at line 856
github.com/syncthing/syncthing/lib/db/set.go in NewFileSet at line 46
```

We can't know for sure where the error is occuring, as the panic happens in defer, but given it happens on 1.10.0 only it's most likely the code added in https://github.com/syncthing/syncthing/pull/6950. The only places I see where a `key not found` could occur are the `Delete` calls. However the deleted keys are keys we just got by iterating, so they should exist.

The github-sentry integration seems to be broken

calmh:
(GitHub has been sending me angry emails about the sentry integration using some deprecated API. We may want to upgrade it at some point.)

imsodin:
Me too. And the sentry integration itself is deprecated, in favor of global apps (but it looks like we are already using that?): https://github.com/getsentry/sentry/tree/master/src/sentry_plugins/github Can't look into it as I don't have sentry admin access.

AudriusButkevicius:
The integration is only used to associate commits with reported events, which was always a hit and miss (mostly a miss), so I am not sure if it matters?

imsodin:
This issue has now suddenly disappeared from sentry. The link above points to something completely different and a search turns up only old stuff: https://sentry.syncthing.net/syncthing/syncthing/?query=key+not+found No clue what's going on.

calmh:
The reports are there in the original, so it wasn't just you hallucinating anyway.

https://crash.syncthing.net/reports/b4ef2d010bc68a3f674ab0833fb1a5054271a1a7da7585bc46563d04e5731cfe
https://crash.syncthing.net/reports/b49f60572e8984e2f96416d39db8fc225b71233f297ef318a045386b2ff0dd37
https://crash.syncthing.net/reports/b4e539b29889cccb80e72e97b9b77375e6270bcd93ece84dae1bd55bc10d7c48
https://crash.syncthing.net/reports/b436bbd7cd150b323c1758473381a4311fabdbf505f87a2f30dbb8859634e2f4
[...]

PR #7027, opened 2020-10-06
lib/db: Improve error handling checking db (ref #7026)

imsodin:
Adds context to the errors in `getMetaAndCheck` to know where it originates from.

Also ignores not found errors when deleting in `checkLocalNeed`. I don't think we should ever get such errors there, but they are the only possible source I see for the panics in #7026, and it doesn't hurt.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -10,6 +10,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
+	"fmt"
 	"io"
 	"os"
 	"time"
@@ -822,6 +823,7 @@ func (db *Lowlevel) getMetaAndCheck(folder string) *metadataTracker {
 	var fixed int
 	fixed, err = db.checkLocalNeed([]byte(folder))
 	if err != nil {
+		err = fmt.Errorf("checking local need: %w", err)
 		return nil
 	}
 	if fixed != 0 {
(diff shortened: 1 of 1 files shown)


lib/db/lowlevel.go
```diff
@@ -822,6 +823,7 @@ func (db *Lowlevel) getMetaAndCheck(folder string) *metadataTracker {
 	var fixed int
 	fixed, err = db.checkLocalNeed([]byte(folder))
 	if err != nil {
+		err = fmt.Errorf("checking local need: %w", err)
```
calmh, line 826: This set-a-local-var-and-return-nil-oops-we're-not-returning-after-all is some serious mindbending bullshit, though not a fault of this PR. :) But perhaps we can fix it, adding an error return here and making a `mustGetMetaAndCheck` that panics on that error? (Until we stop panicking on errors...)

imsodin: I already have a follow-up to https://github.com/syncthing/syncthing/pull/6925 that starts work on making db error handling less terrifying, and I hope suture lands soon so those pieces of the puzzle can also move into place. This really just is a tiny measure to make it at least it bit more transparently terrifying.

calmh: Fine. Worse is better, or something.

calmh (approved):

PR #7029, opened 2020-10-08
cmd/stcrashreceiver, lib/db: Improve panic message handling

imsodin:
This PR does a few related things regarding panics:

 - We have plenty of IO errors from leveldb including the db path, which become panics. Yes they should be errors, yes the goal is still to do that. In the meantime lets remove the paths from the panic message. And anyway, even with errors those would be fatal so the path will have to be filtered eventually.

- Added the panic message to the fingerprint. That's absolutely necessary for https://github.com/syncthing/syncthing/pull/6925, otherwise all events from it will be merged together (as they don't have traces). I also did it for panics to prevent merging different errors with the same trace. There's some pattern matching to prevent stuff like indexes or db paths (old clients without the change above will keep sending them) prevent matching events together.

 - Do not use the fingerprint for pattern matching for db corruption or fatal db io errors. It doesn't matter where those occur. This should reduce the noise in the sentry issue list. This could be extended to other kinds of errors that may occur regardless of time/place. 

No need to tell me the regexyness is terrible (and possibly terrifying), I know it is. I think the benefits are worth the ugliness though, but that's obviously open to discussion.

### DIFF
--- a/cmd/stcrashreceiver/sentry.go
+++ b/cmd/stcrashreceiver/sentry.go
@@ -128,8 +128,37 @@ func parseCrashReport(path string, report []byte) (*raven.Packet, error) {
 		"url": reportServer + path,
 	}
 	pkt.Interfaces = []raven.Interface{&trace}
+	pkt.Fingerprint = crashReportFingerprint(pkt.Message)
 
 	return pkt, nil
+}
+
+var (
+	indexRe          = regexp.MustCompile(`\[[-:0-9]+\]`)
+	sizeRe           = regexp.MustCompile(`(length|capacity) [0-9]+`)
+	ldbPosRe         = regexp.MustCompile(`(\(pos=)([0-9]+)\)`)
+	ldbChecksumRe    = regexp.MustCompile(`(want=0x)([a-z0-9]+)( got=0x)([a-z0-9]+)`)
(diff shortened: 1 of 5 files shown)


AudriusButkevicius (dismissed):

cmd/stcrashreceiver/sentry.go
```diff
@@ -128,10 +128,39 @@ func parseCrashReport(path string, report []byte) (*raven.Packet, error) {
...
+	message = ldbChecksumRe.ReplaceAllString(message, "${1}X${3}X")
+	message = ldbInternalKeyRe.ReplaceAllString(message, "${1}x${2}x")
+	message = ldbPathRe.ReplaceAllString(message, "$1 x: ")
+	if message != orig {
+		return []string{message}
+	}
+
+	message = indexRe.ReplaceAllString(message, "[x]")
+	message = sizeRe.ReplaceAllString(message, "$1 x")
+	return []string{"{{ default }}", message}
```
calmh, line 161: I assume the `{{ default }}` thing here means the backtrace, and then the modified message also becomes part of the fingerprint?

imsodin: Yes. I added a explanation and link to the docs in a comment.

calmh (approved):

calmh:
(crash receiver needs manual deploy, I'll take care of it this evening and report back)

PR #7038, opened 2020-10-17
lib/model: Handle index sender lifetime (fixes #7034)

### LINKED ISSUE #7034: Folders getting stuck in "Syncing" after a20c6ca after doing changes when folders are paused
After a20c6ca868db44327575440bfa3501f96ba3afcd, there is no disconnect on config change anymore, but folders seem to get stuck in the "Syncing" state.

How to reproduce:

1. Pause local folders.
2. Do changes in local folders.
3. Unpause local folders.

image
image

Once stuck, they seem to stay like that basically forever. There is also nothing in the logs.

I have to pause and resume the affected devices (and hence force a connection reset) in order to make the folders actually sync the changes.

Tested with `main` (`v1.10.1-dev.23.gd828adb6`).

imsodin:
Manages life-time of index senders. Previously they were only tied to the connection staying alive and new cluster configs, in which case index senders are terminated. Now index senders are also un-/paused together with folders and stopped when removing a folder. When unpausing the newly generated fileset is passed, thus they stay up-to-date with changes in the newly created folder service - which fixes #7034.

### DIFF
--- a/lib/model/indexsender.go
+++ b/lib/model/indexsender.go
@@ -106,7 +106,6 @@ func (s *indexSender) Complete() bool { return true }
 func (s *indexSender) resume(fset *db.FileSet) {
 	select {
 	case <-s.connClosed:
-		return
 	case s.resumeChan <- fset:
 	}
 }
@@ -114,7 +113,6 @@ func (s *indexSender) resume(fset *db.FileSet) {
 func (s *indexSender) pause() {
 	select {
 	case <-s.connClosed:
-		return
 	case s.pauseChan <- struct{}{}:
 	}
 }
@@ -237,8 +235,11 @@ func newIndexSenderRegistry(conn protocol.Connection, closed chan struct{}, sup 
 // If an index sender is already running, it will be stopped first.
 func (r *indexSenderRegistry) add(folder config.FolderConfiguration, fset *db.FileSet, local, remote protocol.Device) {
 	r.mut.Lock()
-	defer r.mut.Unlock()
+	r.addLocked(folder, fset, remote, local)
+	r.mut.Unlock()
+}
 
+func (r *indexSenderRegistry) addLocked(folder config.FolderConfiguration, fset *db.FileSet, local, remote protocol.Device) {
 	if is, ok := r.indexSenders[folder.ID]; ok {
 		r.sup.RemoveAndWait(is.token, 0)
 		delete(r.indexSenders, folder.ID)
@@ -311,6 +312,8 @@ func (r *indexSenderRegistry) add(folder config.FolderConfiguration, fset *db.Fi
 		fset:         fset,
 		prevSequence: startSequence,
 		evLogger:     r.evLogger,
+		pauseChan:    make(chan struct{}),
+		resumeChan:   make(chan *db.FileSet),
 	}
 	is.Service = util.AsService(is.serve, is.String())
 	is.token = r.sup.Add(is)
@@ -363,7 +366,7 @@ func (r *indexSenderRegistry) resume(folder config.FolderConfiguration, fset *db
 	defer r.mut.Unlock()
 
 	if info, ok := r.startInfos[folder.ID]; ok {
-		r.add(folder, fset, info.local, info.remote)
+		r.addLocked(folder, fset, info.local, info.remote)
 		delete(r.startInfos, folder.ID)
 		return
 	}
--- a/lib/model/indexsender.go
+++ b/lib/model/indexsender.go
@@ -347,7 +347,8 @@ func (r *indexSenderRegistry) remove(folder string) {
 	delete(r.startInfos, folder)
 }
 
-// remove stops a running index sender or removes one pending to be started.
+// removeAllExcept stops all running index senders and removes those pending to be started,
+// except mentioned ones.
 // It is a noop if the folder isn't known.
 func (r *indexSenderRegistry) removeAllExcept(except map[string]struct{}) {
 	r.mut.Lock()
--- a/lib/model/indexsender.go
+++ b/lib/model/indexsender.go
@@ -385,12 +385,15 @@ func (r *indexSenderRegistry) resume(folder config.FolderConfiguration, fset *db
 	r.mut.Lock()
 	defer r.mut.Unlock()
 
+	is, isOk := r.indexSenders[folder.ID]
 	if info, ok := r.startInfos[folder.ID]; ok {
+		if isOk {
+			r.sup.RemoveAndWait(is.token, 0)
+			delete(r.indexSenders, folder.ID)
+		}
 		r.addLocked(folder, fset, info.local, info.remote)
 		delete(r.startInfos, folder.ID)
-		return
-	}
-	if is, ok := r.indexSenders[folder.ID]; ok {
+	} else if isOk {
 		is.resume(fset)
 	}
 }
--- a/lib/model/indexsender.go
+++ b/lib/model/indexsender.go
@@ -347,6 +347,25 @@ func (r *indexSenderRegistry) remove(folder string) {
 	delete(r.startInfos, folder)
 }
 
+// remove stops a running index sender or removes one pending to be started.
+// It is a noop if the folder isn't known.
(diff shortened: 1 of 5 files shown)


lib/model/model.go
```diff
@@ -2572,6 +2636,22 @@ func (m *model) checkFolderRunningLocked(folder string) error {
 	return errFolderNotRunning
 }
 
+// lockBoth locks both fmut and pmut, thus ensuring that always happens in the
+// same order to prevent deadlocks. It doesn't matter in which order the locks
+// are released.
+func (m *model) lockBoth(fmutRead, pmutRead bool) {
```
calmh, line 2642: Not a fan. The multiple-boolean-parameters thing makes it hard to see at the call site what it means (no clue unless reading the signature) and I'm not sure the order thing has been a source of issues for a long time... If you think it really is then I would prefer `lockFandRLockP()` or whatever as more explicit. (Or consolidate both to a single lock for less trickiness to begin with.)

lib/model/model.go
```diff
@@ -1133,14 +1168,17 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
 			fset:         fs,
 			prevSequence: startSequence,
 			evLogger:     m.evLogger,
+			pauseChan:    make(chan struct{}),
```
calmh, line 1171: I think we've reached the point where `func newIndexSender(...) *indexSender` should definitely exist

lib/model/model.go
```diff
@@ -1908,12 +1951,18 @@ func (s *indexSender) serve(ctx context.Context) {
 				return
 			case <-evChan:
 			case <-ticker.C:
+			case <-s.pauseChan:
+				paused = true
```
calmh, line 1955: `s.fset = nil` ?

calmh (commented):
I'm slightly confused, or this is not the whole picture. Index senders should be stopped and started when *we* pause and unpause folders, and when the *other side* does it. I can see index senders being created in cluster config parsing only, but not for folders that are paused on either side. Remote unpausing might create an index sender when we parse the new cluster config, but if we were the one being paused I don't see it being created.

I think this needs a table with four states horizontally ("running on both sides at initial connection time", "paused on our side", "paused on their side", "paused on both sides") and then "live-paused by us", "live-paused by them", "live-resumed by us", "live-resumed by them" vertically, and then making sure the right thing happens in the right place in all cases :)

imsodin:
Right, the tricky part is that we need to receive a cluster config while our folder is not paused to setup an index sender initially, as we need to have the fileset. Which means setting up/starting index sender states needs to be approached from that angle (connection/cluster config), not folder runners/config changes. Stopping them based on a folder being paused/removed still makes sense.

calmh:
I'm thinking the map of maps of index senders might want to be a type of its own (`indexSenderRegistry`?) with appropriate start/pause/resume/stop methods to simplify. Then we do start/stop in cluster configs, possibly starting some senders in paused state directly, and pause/resume based on our own config changes...

lib/model/model.go
```diff
@@ -1050,8 +1036,9 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
 		}
 	}
 
-	var paused []string
-	indexSenders := make(map[string]*indexSender, len(cm.Folders))
+	resendClusterConfig := false
```
calmh, line 1043: This is nasty. I don't think it's reasonable to have *our* internal state when we resume a folder depend on the counterpart noticing the change in our pause status and repeating an identical cluster config. I mean, can you even document that requirement in the BEP spec with a straight face? 😁

imsodin:
I now added such a index sender registry that handles all index senders for one device, and on which one can add and remove folders (depending on if the folder is shared with us and running on the remote) and pause and resume folders (for locally paused folders to preserve sequence information). I also added a unit test checking if indexes are sent (or not) after changing config and pausing/resuming folders in cluster configs.

lib/model/indexsender.go
```diff
@@ -0,0 +1,399 @@
...
+		evLogger:     r.evLogger,
+		pauseChan:    make(chan struct{}),
+		resumeChan:   make(chan *db.FileSet),
+	}
+	is.Service = util.AsService(is.serve, is.String())
+	is.token = r.sup.Add(is)
+	r.indexSenders[folder.ID] = is
+}
+
+// addPaused stores the given info to start an index sender ones resume is called
```
calmh, line 323:
```suggestion
// addPaused stores the given info to start an index sender once resume is called
```

lib/model/indexsender.go
```diff
@@ -0,0 +1,399 @@
...
+}
+
+// resume unpauses an already running index sender or starts it, if it was added
+// while paused.
+// It is a noop if the folder isn't known.
+func (r *indexSenderRegistry) resume(folder config.FolderConfiguration, fset *db.FileSet) {
+	r.mut.Lock()
+	defer r.mut.Unlock()
+
+	if info, ok := r.startInfos[folder.ID]; ok {
```
calmh, line 387: For safety's sake and in keeping in line with the other things, should be a safe no-op or a restart with new fset if it's already running?

lib/model/indexsender.go
```diff
@@ -0,0 +1,399 @@
...
+	defer r.mut.Unlock()
+
+	if is, ok := r.indexSenders[folder]; ok {
+		r.sup.RemoveAndWait(is.token, 0)
+		delete(r.indexSenders, folder)
+	}
+	delete(r.startInfos, folder)
+}
+
+// remove stops a running index sender or removes one pending to be started.
```
calmh, line 350:
```suggestion
// removeAllExcept stops all running index senders and removes those pending to be started,
// except mentioned ones.
```

lib/model/model.go
```diff
@@ -1034,13 +1059,41 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
 			l.Infof("Unexpected folder %s sent from device %q; ensure that the folder exists and that this device is selected under \"Share With\" in the folder configuration.", folder.Description(), deviceID)
 			continue
 		}
+
+		var foundRemote, foundLocal bool
+		var remoteDevice, localDevice protocol.Device
```
calmh, line 1064:
```suggestion
		var remoteDeviceInfo, localDeviceInfo protocol.Device
```

This type (procotol.Device) isn't really awesomely named, confusion with a device ID is easy. :) Mitigating somewhat by variable naming?

lib/model/indexsender.go
```diff
@@ -0,0 +1,400 @@
...
+		r.sup.RemoveAndWait(is.token, 0)
+		delete(r.indexSenders, folder)
+	}
+	delete(r.startInfos, folder)
+}
+
+// removeAllExcept stops all running index senders and removes those pending to be started,
+// except mentioned ones.
+// It is a noop if the folder isn't known.
+func (r *indexSenderRegistry) removeAllExcept(except map[string]struct{}) {
```
calmh, line 353: Minor style thing, I'm not super happy about passing around complex types like this when a `except []string` would be clearer, but I realize that in this case we already have the map and save some work. I'm half minded to add a `stringSet` type for these things (we have on in authors.go) so that it's easy to take a slice and do `stringSetFromStrings(theSlice)` and work on that, and similarly pass on `theStringSet.strings()` when a slice is wanted. Maybe another day, no need to change anything here, I'm just thinking out loud.

lib/model/indexsender.go
imsodin: Right, that's better/safer. Though I am quite certain that it's not possible for a startInfo to be present and the folder being running at the same time as  startInfo is removed when starting, but again, no point relying on that.

lib/model/indexsender.go
imsodin: Honestly here the "list" (as in the general, not type-related term) is purpose built for this function, thus I create it as such a map. I could have also created a slice and passed that. I guess it's a python habit of avoiding loops at all cost :)

calmh (approved):

PR #7040, opened 2020-10-17
lib/db: Undo adding user info to panic msgs (ref #7029)

imsodin:
Removes context including user info like device IDs from panic message.

Was introduced in #7029, i.e. is RC material.

### DIFF
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -531,5 +531,6 @@ func fatalError(err error, opStr string, db *Lowlevel) {
 			}
 		}
 	}
-	warnAndPanic(fmt.Errorf("%v: %w:", opStr, err))
+	l.Warnf("Fatal error: %v: %v", opStr, err)
+	obfuscateAndPanic(err)
 }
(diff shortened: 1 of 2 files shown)


calmh (commented):
I must be blind, this change looks like a no-op to me, what's the active ingredient?

lib/db/set.go
```diff
@@ -531,5 +531,6 @@ func fatalError(err error, opStr string, db *Lowlevel) {
 			}
 		}
 	}
-	warnAndPanic(fmt.Errorf("%v: %w:", opStr, err))
+	l.Warnf("Fatal error: %v: %v", opStr, err)
```
calmh, line 534: Oh, it's in the `opStr` here is it? If so, all good.

calmh (approved):

lib/db/set.go
imsodin: Exactly, it contains info on method and the problematic arguments, and baking it into the error here made it turn up in the panic message.

PR #7044, opened 2020-10-19
lib/db: Remove need for the right dev removing globals (fixes #7036)

### LINKED ISSUE #7036: Shutting Syncthing down while pushing files to multiple devices corrupts the database?
Related to https://forum.syncthing.net/t/folders-up-to-date-but-stuck-in-syncing-with-empty-out-of-sync-items/15667.

How to reproduce:

1. Create and share an empty folder between 3 devices - A, B, and C. 
2. Add a few large files to the folder on A.
3. While the files are still being pushed from A to B and C, shut Syncthing down on A.
5. Restart Syncthing on A.
6. Wait for the previously interrupted synchronisation to finish.
7. C is stuck trying to sync the files to B, but they exist there already.

image
image
image
image

imsodin:
Huge thanks to @tomasz1986 for not letting loose on this issue until discovering a reproducer so... lacking a better word lets say intricate, and also providing all the logs and screenshots I could wish for. Really, this fix is your achievement!

Not much to say about the fix: Stupid little slip. The code in question was even test covered before, but apparently not checked.

There's one question: There's a good chance this affects a few users. Recalculating metadata when releasing the fix seems in order - how should we trigger that? A db transition would work, but is a bit ingenious as there isn't any actual transition. We could also add an "upgrade hook" to syncthing.go that can do arbitrary stuff when updating to specific version, and export the db repair function again to call it there. Better ideas?

### DIFF
--- a/lib/db/transactions.go
+++ b/lib/db/transactions.go
@@ -859,7 +859,7 @@ func (t readWriteTransaction) removeFromGlobal(gk, keyBuf, folder, device, file 
 			continue
 		}
 		if fv, have := fl.Get(dev[:]); Need(removedFV, have, fv.Version) {
-			meta.removeNeeded(deviceID, f)
+			meta.removeNeeded(dev, f)
 		}
 	}
 
--- a/lib/model/model_test.go
+++ b/lib/model/model_test.go
@@ -4142,6 +4142,46 @@ func TestCompletionEmptyGlobal(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


calmh (approved):

calmh:
The only issue I have with using a migration is that it then prevents downgrade. We could have a separate counter for non-schema-affecting migrations, we could artificially lower STRECHECKDBEVERY to something lower on upgrade, we could ... something

PR #7045, opened 2020-10-20
lib/connections: Fix LAN addresses begin advertised even when disabled (fixes #7035)

### LINKED ISSUE #7035: Devices Announcing LAN addresses even with AnnounceLANAddresses set to off/false
Ever since the 1.10 update all of my devices are announcing their LAN addresses, even after I switch the config to turn that feature off. Despite this, devices are remembering or still discovering the LAN addresses of other devices.

I was expecting that turning off the LAN announcing would return the devices to pre-1.10 operations, but that is not the case. I do not want devices announcing their LAN addresses to either the Global Discovery servers or to each other if they are on completely separate networks.

AudriusButkevicius:
A clear oversight. Verified manually by checking the logs.

### DIFF
--- a/lib/connections/quic_listen.go
+++ b/lib/connections/quic_listen.go
@@ -160,7 +160,7 @@ func (t *quicListener) URI() *url.URL {
 }
 
 func (t *quicListener) WANAddresses() []*url.URL {
-	uris := t.LANAddresses()
+	uris := []*url.URL{t.uri}
 	t.mut.Lock()
 	if t.address != nil {
 		uris = append(uris, t.address)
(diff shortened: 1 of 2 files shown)


calmh (commented):
So the configured URI is always considered both a LAN address and a WAN address and announced regardless? It seems to me we should maybe look at the address and classify it as either LAN or WAN depending on if it's unspecified, RF1918 private, or other?

calmh:
Or I guess we always announced the statically configured URI and should continue to do so?

AudriusButkevicius:
I think the idea behind the two functions is different. It's more like GlobalAddresses LocalAddresses, as that is effectively what it's used for.

calmh:
Potentially, but if the listen address is `tcp://192.168.0.2:22000` then it's clearly not a WAN address imho. Yet, we've apparently always announced the listen address verbatim regardless, so we can continue to do so now as well.

calmh (approved):

PR #7047, opened 2020-10-21
lib/db: Add mechanism to repair db without schema update (ref #7044)

imsodin:
This adds a similar db transition mechanism to the one already existing for schema changes. Schema transitions have in the past been "abused" for operations that did not actually change the db. This puts unnecessary limitations on downgrading. The added mechanism is meant to be used in these cases.

I also added the first such repair transition for #7044, which fixes a bug that could lead to corrupt metadata, and thus requires a metadata recalculation.

### DIFF
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -87,6 +87,10 @@ func (db *schemaUpdater) updateSchema() error {
 	if err != nil {
 		return err
 	}
+	// Cover versions before adding `dbMigrationVersion` (== 0) and possible future weirdness.
+	if prevMigration < prevVersion {
+		prevMigration = prevVersion
+	}
 
 	if prevVersion == dbVersion && prevMigration >= dbMigrationVersion {
 		return nil
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -28,9 +28,13 @@ import (
 //   10-11: v1.6.0
 //   12-13: v1.7.0
 //   14: v1.9.0
+//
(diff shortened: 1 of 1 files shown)


calmh:
I had a different but similar idea for this mechanism, which may better or worse... We could have each migration return the minimum schema version required to understand the new format, and we'd store that counter separately for the purpose of running migrations. Call the existing one `schemaVersion` and the new one `migratedSchemaVersion` for example, then the new migration in this case would return `schemaVersion=14` while being a migration to version 15. We'd store `{schemaVersion=14, migratedSchemaVersion=15}` in the database, old things could still open the database and we'd know that we'd applied migrations up to 15.

imsodin:
That's better indeed. It's both a smaller change and it's easier to understand in which order migrations are applied. The function naming isn't ideal: As migrations might now be schema upgrades or not, they aren't consistent anymore. Which isn't a problem in itself, but I could rename all of them to `migrationX` for consistency, but less meaning in names, if that's preferable.

lib/db/schemaupdater.go
```diff
@@ -83,64 +83,41 @@ func (db *schemaUpdater) updateSchema() error {
 		return err
 	}
 
-	prevRepairVersion, _, err := miscDB.Int64("dbRepairVersion")
+	prevMigration, _, err := miscDB.Int64("dbMigrationVersion")
 	if err != nil {
 		return err
 	}
```
calmh, line 90: `if prevMigration < prevVersion { prevMigration = prevVersion }` or something somewhere to cover legacy + impossible oddities?

calmh (approved):

PR #7049, opened 2020-10-21
gui: Refactor to make encryption diff smaller

imsodin:
A large part of the gui diff in https://github.com/syncthing/syncthing/pull/6727 are refactors to make my live easier or prevent duplication. To implement the feature flag and keep the diff minimal, I factored out all those changes that do not change behaviour and that I consider useful even if encryption wouldn't be there:

 - Devices are now tracked as a map internally.

 - Use `$scope.currentSharing` to track sharing changes instead of having "virtual attributes" on the device-/folderconfigs.

Testing: Un-/Sharing folders/devices still works.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -204,10 +204,10 @@
               </div>
               <div class="panel-body">
                 <p>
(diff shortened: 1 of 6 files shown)


calmh (approved):
javascript lol

Issue #62, opened 2014-02-09
Non-source / read only / untrusted / backup devices


calmh:
I.e. ones that we never accept updates from, but who can otherwise participate in the exchange as usual. Needs #63 to work securely.

dzarda:
I have found myself in the need of this feature. After reading through this, I got the idea of _setting the scan rate to 0._ This seems to get rid of the Hard-Drive-not-sleeping issue. Syncthing seems to accept this quite well - it disables HDD checking, CPU sits at 0%, so does IO activity.

Note: This is a per-directory approach.

Is this a documented feature? Or did I just create a feature out of thin air? :)

Nutomic:
@dzarda That's definitely working as intended (we use it to disable the scanner on Android where inotify is implemented). I'll update the wiki on this.

But I think it doesn't solve this issue, as it doesn't protect against malicious devices.

generalmanager:
There are many different use cases in which this could be useful. I'll describe a setup which includes the more common ones and their difficulties.
Please skip to the description at the end if you are familiar with the problems of the currently implemented "master" devices.

There are five devices:

1. desktop PC
2. laptop
3. smartphone
4. server
5. a friend's PC

There are four folders:

**1. A global share folder:**

Device | Folder rights
------------ | -------------
PC | RW
laptop | RW
smartphone | RW
server | RO
friend's PC | RW

This is currently not possible, because if all devices exept the server were "master" nodes, they would ignore each others updates. If no device is set to master mode, the RO device would have RW access tough.

**2. A folder containing the pictures from the phone:**

Device | Folder rights
------------ | -------------
PC | RO
laptop | RO
smartphone | RW
server | RO
friend's PC | --

This is possible with the current "master folders", but does not automatically overwrite changes on RO devices, which is very annoying in many cases.

**3. A folder to exchange files between you and a friend:**

Device | Folder rights
------------ | -------------
PC | RW
laptop | RW
smartphone | RW
server | RO
friend's PC | RW

This is not currently possible, same reasons as in **1.**

**4. A backup folder which you entrust your friends with, but you don't want them to make local changes, because it would destroy your backup:**

Device | Folder rights
------------ | -------------
PC | RW
laptop | RW
smartphone | --
server | RO
friend's PC | RO

[...]

PR #7055, opened 2020-10-23
all: Add untrusted folders behind feature flag (ref #62)

imsodin:
This is a proposal for adding untrusted folders (encryption) support behind a feature flag, as discussed in https://forum.syncthing.net/t/encryption-merging-scary-changes-feature-flags/15703

The PR branch is an extension of https://github.com/syncthing/syncthing/pull/6727, with the feature flag added and the stdecryptfolder tool removed. That still had outstanding PR reviews and isn't necessary to have in a first iteration for testing this. Other than that I believe I addressed all the PR reviews from there and this should be ready to merge. Of course except for the feature flag code, which no one has reviewed yet.

The feature flag determines which web UI is served: The "traditional" one or the new one with controls for untrusted folders. A missing feature flag does not prevent "manual" config changes that introduce untrusted folders. My reasoning being is that it not being available by default in UI is enough of a barrier to prevent wide adoption. I can add a mechanism in the config layer to prevent untrusted folders without the flag if deemed necessary.

The feature flag itself is just a string, such that it can be reused if we ever want to use a feature flag again, instead of creating ad deprecating a bool every time.

The statics server is adjusted, such that if the feature flag is set, it looks for files in a "untrusted" subdir and serves those instead of the "normal" files if present. Those files are adjusted full copies of the originals, i.e. if other UI changes happen, those need to be kept in sync by hand. That's ugly, but I consider it manageable (thus volunteering xD ).


calmh (dismissed):
I fixed some things. Merge?

calmh:
@imsodin please sanity check my latest merge commit

imsodin:
The noscript change isn't in the untrusted index.html - taking care of that.

imsodin:
Now the "UI copies" should match again.

> I fixed some things. Merge?

Thanks, that plus fixes look good, Yes :)

calmh (approved):

calmh:
TestIssue5063 is flaky and times out sometimes, reproducible with -count 100 for example, I didn't investigate yet if it's due to this change or just in general

imsodin:
Seems related, tests time out for me on this branch but not on main. Plus I just realized the recent index sender changes duplicate some encryption code, need to sort that out (as in a merge-leftover). Looking into it.

imsodin:
This PR made it worse, but the problem exists in v1.11.1 too: I messed up locking in the recent index changes. I'll file a separate PR on main and then take care of this PR.

calmh (approved):

PR #7056, opened 2020-10-24
gui: Fix undefined variables fallout from #7049.

acolomb:
Some accesses to the new $scope.devices object (instead of previously
array) used wrong / inexistent names for the index variable.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1105,7 +1105,7 @@ angular.module('syncthing.core')
         };
 
         $scope.setDevicePause = function (device, pause) {
-            $scope.devices[id].paused = pause;
+            $scope.devices[device].paused = pause;
             $scope.config.devices = $scope.deviceList();
             $scope.saveConfig();
         };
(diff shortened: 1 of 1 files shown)


gui/default/syncthing/core/syncthingController.js
```diff
@@ -1466,6 +1466,7 @@ angular.module('syncthing.core')
                 return;
             }
 
+	    var id = $scope.currentDevice.deviceID
             delete $scope.devices[id];
```
imsodin, line 1470: Missing indentation. I'd directly put the expression as the key, without using the intermediate variable.

imsodin (commented):
Thanks a lot for catching this!

gui/default/syncthing/core/syncthingController.js
acolomb: Whitespace issue fixed, that was my editor inserting a tab for 8 spaces...

The id variable is used again twice inside the loop below, so I thought caching it locally helps readability.

imsodin (approved):

PR #7061, opened 2020-10-27
gui: Readd check if device exists (ref #7059)

imsodin:
This reverts PR #7059 / commit c7d40ccbaee867b579ccb291b258fd28299b3464.

I was speeding when merging that: There is no looping, it's just a safety check that the device in question actually exists.


calmh (approved):

acolomb:
@imsodin Sorry, that was a really dumb oversight of mine. Must have been confused with some other branch that I didn't notice it was actually an `if` there on main.

Issue #7064, opened 2020-10-28
fatal error: runtime.newosproc


tomasz1986:
The panic happened without any user intervention. Syncthing had just been doing its job, and I only found the log file later.

The log comes from Syncthing v1.10.0 (x86-32) running in Windows 10 Enterprise 2016 LTSB x86.

```
Panic at 2020-10-28T21:25:20+09:00
fatal error: runtime.newosproc

runtime stack:
runtime.throw(0x120fa12, 0x11)
	runtime/panic.go:1116 +0x64
runtime.newosproc(0x17ecafc0)
	runtime/os_windows.go:794 +0x146
runtime.newm1(0x17ecafc0)
	runtime/proc.go:1829 +0xbd
runtime.newm(0x13e5b7c, 0x12c2c000, 0x4a7, 0x0)
	runtime/proc.go:1808 +0x81
runtime.startm(0x12c2c000, 0x1)
	runtime/proc.go:1965 +0xac
runtime.handoffp(0x12c2c000)
	runtime/proc.go:2004 +0x28c
runtime.retake(0x68d8ec28, 0x4587, 0x0)
	runtime/proc.go:4805 +0x171
runtime.sysmon()
	runtime/proc.go:4713 +0x275
runtime.mstart1()
	runtime/proc.go:1172 +0x9b
runtime.mstart()
	runtime/proc.go:1137 +0x51
```

[panic-20201028-212520.log](https://github.com/syncthing/syncthing/files/5452199/panic-20201028-212520.log)

AudriusButkevicius:
This is the OS refusing to start a thread when go asked for it.

None of the code is ours and if anything would be a go runtime bug, but more likely issues with the os running out of handles or something.

tomasz1986:
It seems that the reason may be https://devblogs.microsoft.com/oldnewthing/?p=34773.

>Often I see people asking why they can’t create more than around 2000 threads in a process. The reason is not that there is any particular limit inherent in Windows. Rather, the programmer failed to take into account the amount of address space each thread uses.
>
>A thread consists of some memory in kernel mode (kernel stacks and object management), some memory in user mode (the thread environment block, thread-local storage, that sort of thing), plus its stack. (Or stacks if you’re on an Itanium system.)
>
>Usually, the limiting factor is the stack size.
>
>This program will typically print a value around 2000 for the number of threads.
>
>Why does it give up at around 2000?
>
[...]

PR #7069, opened 2020-10-30
lib/config: Sanity checks on MaxConcurrentWrites (ref #7064)

imsodin:
This changes the meaning of `MaxConcurrentWrites <= 0` to mean default and additional puts an upper limit on the config value too. I guess the latter is going to be controversial, as it's clearly user error. My reasoning is that I'd rather check for that than have to explain the error to a user. And the limit of 64 is entirely randomly chosen, I do not know anything about concurrent IO, so if there is a technical reason to change that or drop it alltogether, please let me know.

### DIFF
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -129,6 +129,7 @@ func TestDeviceConfig(t *testing.T) {
 				WeakHashThresholdPct: 25,
 				MarkerName:           DefaultMarkerName,
 				JunctionsAsDirs:      true,
+				MaxConcurrentWrites:  maxConcurrentWritesDefault,
 			},
 		}
 
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -26,7 +26,11 @@ var (
 	ErrMarkerMissing    = errors.New("folder marker missing (this indicates potential data loss, search docs/forum to get information about how to proceed)")
 )
 
-const DefaultMarkerName = ".stfolder"
+const (
+	DefaultMarkerName          = ".stfolder"
+	maxConcurrentWritesDefault = 2
+	maxConcurrentWritesLimit   = 64
+)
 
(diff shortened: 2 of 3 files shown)


calmh (commented):
I think the `2` needs to be named somewhere, and we should skip the Infof. Also we have in some cases done this type of clamping in a method on the options type, I wonder if it might not be cleaner to have all such logic centralized there...

imsodin:
It's a mixture at present: Sometimes we replace 0 with the default value in config and othertimes we leave 0 and use a default value outside of config. The former has the advantage that everything is in one place and the stored config value is the same as the effective one, while the latter keeps the stored value the same as what the user specified, making it obvious if it's the default value and preventing possible confusion when the stored value is different from the specified one. I think the former is better and thus moved this change to the config.

tomasz1986:
The upper limit sounds kind of low though, especially considering superfast storage like NVME RAID or similar. I am no storage expert by any means, but even my very slow SD card has been able to withstand unlimited `MaxConcurrentWrites` so far with only 1 crash (https://github.com/syncthing/syncthing/issues/7064).

tomasz1986:
I am just saying that, as Simon also mentioned in the first post, it would be good to have some kind of a technical reason of limiting the setting at `64` and not `32` or `128`, for example.

imsodin:
That's not what I meant: I invited anyone that has concrete arguments/requirements for a different upper limit to state them. I did not mean to open a discussion on it. This is about preventing the worst in case of misconfiguration (or messing up default values). If anyone does think we should debate parallel io, then please don't do that on this PR.

Good to merge?

calmh (approved):

calmh:
I really have no opinion on what amount of concurrent I/O people would or could run with, and see no reason to limit it for that reason. But in practice the amount of concurrent I/O will also be the lower bound on the number of running threads and those are neither free nor unlimited, so an upper limit makes sense. I doubt that a lack of more concurrent I/O will ever be the performance bottleneck.

PR #7017, opened 2020-09-30
gui: add advance config port mapping to gui

rjpruitt16:
### Purpose

https://github.com/syncthing/syncthing/issues/4824

Gives the user the opportunity to display a link to web gui remote machines. The user can choose the port and set if the link will display.

### Testing

I just check to see if the link was shown depending on the setting.

### Screenshots

Screen Shot 2020-09-29 at 11 13 53 PM
Screen Shot 2020-09-29 at 11 34 37 PM

### Documentation

https://github.com/syncthing/docs/pull/573

## Authorship

Your name and email will be added automatically to the AUTHORS file
based on the commit metadata.

PR #7079, opened 2020-11-01
lib/api: Allow OPTIONS method in CORS preflight request handling.

acolomb:
### Purpose

Extracted from #7017, I think this change is of general utility.  It reports the HTTP OPTIONS method as part of the allowed methods for the API.  That is normally used as part of the CORS preflight check when doing a GET or POST request.  When just trying to check API availability, a deliberate OPTIONS request is enough, but currently it fails because of the returned `Access-Control-Allow-Methods` header.

I don't see any security risks here and was hoping to get this small change into a release before #7017 is finalized, so that the probing done there actually works with as many instances as possible.

### Testing

Verified that the header is set: `curl -X OPTIONS http://localhost:8384/ -v`.  Tested with Firefox (as part of #7017) that it no longer fails with a [CORS error](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMethodNotFound?utm_source=devtools&utm_medium=firefox-cors-errors&utm_campaign=default) when doing an OPTIONS request from within the JS code.

### DIFF
--- a/lib/api/api.go
+++ b/lib/api/api.go
@@ -504,7 +504,7 @@ func corsMiddleware(next http.Handler, allowFrameLoading bool) http.Handler {
 			// Add a generous access-control-allow-origin header for CORS requests
 			w.Header().Add("Access-Control-Allow-Origin", "*")
 			// Only GET/POST Methods are supported
-			w.Header().Set("Access-Control-Allow-Methods", "GET, POST")
+			w.Header().Set("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
 			// Only these headers can be set
 			w.Header().Set("Access-Control-Allow-Headers", "Content-Type, X-API-Key")
 			// The request is meant to be cached 10 minutes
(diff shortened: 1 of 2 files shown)


lib/api/api.go
```diff
@@ -504,7 +504,7 @@ func corsMiddleware(next http.Handler, allowFrameLoading bool) http.Handler {
 			// Add a generous access-control-allow-origin header for CORS requests
 			w.Header().Add("Access-Control-Allow-Origin", "*")
 			// Only GET/POST Methods are supported
-			w.Header().Set("Access-Control-Allow-Methods", "GET, POST")
+			w.Header().Set("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
```
AudriusButkevicius, line 507: Comment above needs to be updated.

acolomb: Right, sorry.

AudriusButkevicius (approved):

PR #7081, opened 2020-11-01
lib/api: Fix /rest/config path and add methods to cors (ref #7001)

imsodin:
Just noticed through #7079 that there's a method whitelist and thus added the new methods used in the `/rest/config` endpoints. And this also fixes an issue with `/rest/config` the only endpoint added in #7001 that I didn't add a test for. Weirdly enough adding the test wouldn't have helped: It doesn't fail (neither does the UI, which uses it too), while it does fail using curl. Anyway, works with curl too now.

### DIFF
--- a/lib/api/api.go
+++ b/lib/api/api.go
@@ -302,7 +302,7 @@ func (s *service) serve(ctx context.Context) {
 		mut:    sync.NewMutex(),
 	}
 
-	configBuilder.registerConfig("/rest/config/")
+	configBuilder.registerConfig("/rest/config")
 	configBuilder.registerConfigInsync("/rest/config/insync")
 	configBuilder.registerFolders("/rest/config/folders")
 	configBuilder.registerDevices("/rest/config/devices")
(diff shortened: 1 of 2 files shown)


acolomb:
That whitelist applies only to cross-origin requests AFAIK. Regular GUI usage is not affected by that, since the script making the requests comes from the same origin as the request target URL. As far as I understand, the browser takes care of disallowing requests from other script origins. So curl not doing it would be expected, since the "origin" is the user's command line.

Does that make sense? I'm no expert on CORS, and don't quite see what security is gained by the headers Syncthing sends...

calmh (approved):

calmh:
Afaik this does not improve security, it on the contrary allows requests from non-GUI origins that would otherwise be blocked by the browser. For example, some Javascript running somewhere using the API key to do things.

acolomb:
> Afaik this does not improve security, it on the contrary allows requests from non-GUI origins that would otherwise be blocked by the browser. For example, some Javascript running somewhere using the API key to do things.

I meant I don't see what security gain the CORS check brings for Syncthing. As it effectively poses no restrictions, even more so with this change. Having the API open for all script origins is okay, I just wasn't sure if this function in Syncthing was actually *meant* to increase security somehow.

PR #7091, opened 2020-11-03
lib/ur: Fix panics in failure-reporting (fixes #7090)

### LINKED ISSUE #7090: panics in failure reporting
https://sentry.syncthing.net/syncthing/syncthing/issues/4783

```
panic: interface conversion: interface {} is nil, not string

github.com/syncthing/syncthing/lib/ur/failurereporting.go in serve at line 97

    			}
    			url = opts.CRURL + "/failure"
    		case e := <-h.evChan:
    			descr := e.Data.(string)
    			if stat, ok := h.buf[descr]; ok {
    				stat.last = e.Time
    				stat.count++

github.com/syncthing/syncthing/lib/util/utils.go in Serve at line 300
github.com/thejerf/suture@v3.0.2+incompatible/supervisor.go in func1 at line 600
```

The channel is closed when unsubscribing, thus we receive from it. The channel was supposed to be set to nil on unsubscribe, but that doesn't happen.

imsodin:
The main fix is in setting the channel variable to nil when unsubscribing (line 97). In addition I removed that var from the struct, because it's only used in `serve` and for good measure added a check that the channel wasn't closed when receiving from it. And finally added a context escape hatch to the goroutine doing the initial config setup.

EDIT: Actually if putting in extra safeties, lets do it right and also check the type conversion itself. And report if either of those two checks fail, because they shouldn't ever fail so I want to know if they do.

### DIFF
--- a/lib/ur/failurereporting.go
+++ b/lib/ur/failurereporting.go
@@ -102,7 +102,7 @@ outer:
 			url = opts.CRURL + "/failure"
 		case e, ok := <-evChan:
 			if !ok {
-				// Just to be save - shouldn't ever happen, as
+				// Just to be safe - shouldn't ever happen, as
 				// evChan is set to nil when unsubscribing.
 				h.addReport(evChanClosed, time.Now())
 				evChan = nil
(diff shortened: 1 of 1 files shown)


lib/ur/failurereporting.go
```diff
@@ -86,25 +92,29 @@ outer:
...
-				stat.count++
-			} else {
-				h.buf[descr] = &failureStat{
-					first: e.Time,
-					last:  e.Time,
-					count: 1,
-				}
+		case e, ok := <-evChan:
+			if !ok {
+				// Just to be save - shouldn't ever happen, as
```
calmh, line 105:
```suggestion
				// Just to be safe - shouldn't ever happen, as
```

calmh (approved):

PR #7095, opened 2020-11-06
lib/api, lib/db: Add file debug endpoint

imsodin:
The `/rest/db/file` endpoint already has a good amount of helpful info when debugging weird out-of-sync issues, but it doesn't give the full picture. I added a `/rest/debug/file` endpoint which in addition to the info already contained in `/rest/db/file` also prints the global versions. Thus one can see at a glance all involved devices and at which versions they are (plus if ignored and/or deleted). I named all the involved method with `Debug` and the api data is a string, as it is an internal implementation detail and should not be used for anything but debugging.


calmh:
Is this better than adding more info to the existing file endpoint?

imsodin:
I didn't want to do that because of the global version list being an implementation detail. While unlikely, I don't want anyone to start depending on that info. On a `/rest/debug` endpoint it seems fine to e.g. change the format at any time, while on `/lib/db/file` not so much.

calmh (approved):

PR #7102, opened 2020-11-09
gui: Add warning when JavaScript is disabled in Web browser (fixes #7099)

### LINKED ISSUE #7099: Add a html message to GUI when javascript is disabled
If the browser loading the syncthing web GUI has something like noscript blocking javascript, the user is presented with a blank page. It would be helpful if a fallback html message stated that javascript is required for the page to load correctly.

tomasz1986:
When using a Web browser with JavaScript either disabled or unavailable,
show a warning to let the user know that the Web GUI requires JS in
order to operate.

To achieve this, add a `div` that wraps both the navbar and the main
content, and then move the CSS class `ng-cloak` from the `html` element to
that `div`. This way, only the JavaScript-dependent part is hidden when
JS is unavailable, and not the whole website, as it is the case right
now. Then, add a `<noscript>` element right at the start of the `body`
element, so that the warning is also shown right away in text-based Web
browsers. The `noscript` element includes a stripped down version of the
navbar showing only the Syncthing logo, and then a container with the
warning itself. Lastly, leave the footer untouched and always visible,
because it does not rely on JavaScript at all.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -53,7 +53,7 @@
             </div>
             <div class="panel-body">
               <p>
-              <span translate>The Syncthing admin interface requires JavaScript to function correctly. Please enable JavaScript in your Web browser and try again.</span>
+              The Syncthing admin interface requires JavaScript. Please enable JavaScript in your Web browser and try again.
               </p>
             </div>
           </div>
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -7,7 +7,7 @@
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
 -->
-<html lang="en" ng-app="syncthing" ng-controller="SyncthingController" class="ng-cloak">
+<html lang="en" ng-app="syncthing" ng-controller="SyncthingController">
 <head>
   <meta charset="utf-8"/>
   <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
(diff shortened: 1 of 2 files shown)


tomasz1986:
This is how the warning looks like:

image

I know that I am duplicating the Syncthing logo code, but the alternative way is to either to add the `ng-cloak` class to all the other navbar elements (which is complex and also requires all future additions to remember about it) and the main div container, or add a div wrapper with `ng-cloak` over both the navbar and the main container in order to hide the "normal" content when JS is disabled, which looks kind of ugly and also increases the code indentation even more.

In the end, I think that adding the `ng-cloak` class to the existing navbar and div container elements, and then using a separate `noscript` element to display the warning is simpler and more maintainable. Please correct me if I am mistaken here.

gui/default/index.html
```diff
@@ -29,786 +29,821 @@
...
+              <h3 class="panel-title">
+                <div class="panel-icon">
+                  <span class="fas fa-exclamation-circle"></span>
+                </div>
+                <span translate>Warning!</span>
+              </h3>
+            </div>
+            <div class="panel-body">
+              <p>
+              <span translate>The Syncthing admin interface requires JavaScript to function correctly. Please enable JavaScript in your Web browser and try again.</span>
```
calmh, line 56:
```suggestion
              <span>The Syncthing admin interface requires JavaScript. Please enable JavaScript in your Web browser and try again.</span>
```

calmh: (no translation is going to happen here because no javascript, and slightly shorter.)

calmh (approved):

gui/default/index.html
```diff
@@ -29,786 +29,821 @@
...
+              <h3 class="panel-title">
+                <div class="panel-icon">
+                  <span class="fas fa-exclamation-circle"></span>
+                </div>
+                <span translate>Warning!</span>
+              </h3>
+            </div>
+            <div class="panel-body">
+              <p>
+              The Syncthing admin interface requires JavaScript. Please enable JavaScript in your Web browser and try again.
```
calmh, line 56:
```suggestion
              The Syncthing admin interface requires JavaScript. Please enable JavaScript in your web browser and try again.
```

calmh (approved):

gui/default/index.html
```diff
@@ -29,786 +29,821 @@
...
+    <div class="container content">
+      <div class="row">
+        <div class="col-md-12">
+          <div class="panel panel-danger">
+            <div class="panel-heading">
+              <h3 class="panel-title">
+                <div class="panel-icon">
+                  <span class="fas fa-exclamation-circle"></span>
+                </div>
+                <span translate>Warning!</span>
```
tomasz1986, line 51: Already merged, but I guess that there should be no `translate` here either, right?

calmh: Indeed; fixed.

PR #7107, opened 2020-11-09
lib/model: Fix locking when resending cluster-configs

imsodin:
The current code violates the "fmut-pmut-locking-order". I can't find any deadlock panics pointing at this nor is there a big increase in deadlocks panics on sentry - maybe I got lucky and this somehow doesn't have consequences.


imsodin:
What made this low (no?) consequence, is that we only acquire a read-lock on fmut here. The only place where we acquire both a full fmut lock and a pmut lock is in `removeFolder`. And that's called from `CommitConfiguration` as well, thus is unlikely to happen at the same time.

calmh (approved):

PR #6068, opened 2019-10-08
Update lucas-clemente/quic-go sum

popey:
I was building an unrelated PR of syncthing, and the sum for lucas-clemente/quic-go needed updating. So doing this separately, another PR to follow.

### Purpose

Fixes the build, which fails when the checksum for this component is wrong, which it is.

```
get "gopkg.in/ldap.v2": found meta tag get.metaImport{Prefix:"gopkg.in/ldap.v2", VCS:"git", RepoRoot:"https://gopkg.in/ldap.v2"} at https://gopkg.in/ldap.v2?go-get=1                                        
go: extracting gopkg.in/ldap.v2 v2.5.1                                                                                                                                                                       
go: downloading gopkg.in/asn1-ber.v1 v1.0.0-20181015200546-f715ec2f112d                                                                                                                                      
Fetching https://gopkg.in/asn1-ber.v1?go-get=1                                                                                                                                                               
verifying github.com/lucas-clemente/quic-go@v0.12.0: checksum mismatch                                                                                                                                       
        downloaded: h1:dYHUyB50gEQlK3KqytmNySzuyzAcaQ3iuI2ZReAfVrE=                                                                                                                                          
        go.sum:     h1:TRbvZ6F++sofeGbh+Z2IIyIOhl8KyGnYuA06g2yrHdI=                                                                                                                                          
exit status 1                                                                                                                                                                                                
exit status 1 

```

### Testing

I did a test build in a clean lxc container running Ubuntu 16.04.
`go run build.go `
Without this patch, the build fails.

Issue #7033, opened 2020-10-13
Default Folder uses different config than user-created folders


tomasz1986:
How to reproduce:

1. Create a new folder in a fresh installation of Syncthing.
2. Compare the configuration of the newly created folder, and the Default Folder.

Default Folder
```
<folder id="default" label="Default Folder" path="" type="sendreceive" rescanIntervalS="3600" fsWatcherEnabled="true" fsWatcherDelayS="10" ignorePerms="false" autoNormalize="true">
		<filesystemType>basic</filesystemType>
		<device id="" introducedBy=""></device>
		<minDiskFree unit="">0</minDiskFree>
		<versioning>
				<cleanupIntervalS>0</cleanupIntervalS>
		</versioning>
		<copiers>0</copiers>
		<pullerMaxPendingKiB>0</pullerMaxPendingKiB>
		<hashers>0</hashers>
		<order>random</order>
		<ignoreDelete>false</ignoreDelete>
		<scanProgressIntervalS>0</scanProgressIntervalS>
		<pullerPauseS>0</pullerPauseS>
		<maxConflicts>-1</maxConflicts>
		<disableSparseFiles>false</disableSparseFiles>
		<disableTempIndexes>false</disableTempIndexes>
		<paused>false</paused>
		<weakHashThresholdPct>25</weakHashThresholdPct>
		<markerName>.stfolder</markerName>
		<copyOwnershipFromParent>false</copyOwnershipFromParent>
		<modTimeWindowS>0</modTimeWindowS>
		<maxConcurrentWrites>2</maxConcurrentWrites>
		<disableFsync>false</disableFsync>
		<blockPullOrder>standard</blockPullOrder>
		<copyRangeMethod>standard</copyRangeMethod>
		<caseSensitiveFS>false</caseSensitiveFS>
		<junctionsAsDirs>false</junctionsAsDirs>
</folder>
```

User-created folder
```
<folder id="test" label="Default Folder" path="" type="sendreceive" rescanIntervalS="3600" fsWatcherEnabled="true" fsWatcherDelayS="10" ignorePerms="false" autoNormalize="true">
		<filesystemType>basic</filesystemType>
		<device id="" introducedBy=""></device>
		<minDiskFree unit="%">1</minDiskFree>
		<versioning>
				<cleanupIntervalS>0</cleanupIntervalS>
		</versioning>
		<copiers>0</copiers>
		<pullerMaxPendingKiB>0</pullerMaxPendingKiB>
[...]

PR #7131, opened 2020-11-19
all: Add configurable defaults (fixes #4224, fixes #6086)

### LINKED ISSUE #4224: Editable default values for folders, devices

### LINKED ISSUE #6086: Default Folder uses different config than user-created folders
While setting up a new folders I noticed that there are different defaults for `maxConflicts`

https://github.com/syncthing/syncthing/blob/999d4a0e2397c44863c04f87596a484ab4f41c39/gui/default/syncthing/core/syncthingController.js#L69

https://github.com/syncthing/syncthing/blob/f6f696c6c51d0a5d27e548b814f6efb6e5af5e8d/lib/config/folderconfiguration.go#L53

https://github.com/syncthing/syncthing/blob/6b74cdc6138a0e07f1954eb03ad479438082a8b1/man/syncthing-rest-api.7#L130

Expected value
-1

imsodin:
The PR is still WIP because of the to-dos below, but it is usable as it is right now.

The branch lives on top of https://github.com/syncthing/syncthing/pull/7127 (first three commits at the time of writing) - I would suggest looking at that first, however if you'd rather have one huge diff, it's available here :) 

The defaults in proto are represented as normal `FolderConfiguration`/`DeviceConfiguration` messages. I added a new `ext.nodefault` proto extension/tag to mark fields which are not configurable (like the id). Those are used to ensure that they are at their zero value when committing a config change. Using the same type allows using it directly as a template for new folders/devices and means there's no duplication.

The defaults can be obtained and set through the rest api on `rest/config/defaults/folder` resp. `.../device` and that endpoint is used instead of the hardcoded values in the GUI (fixes #6068).

### Changed defaults

If the defaults in the GUI did not match those in the config, I chose the GUI ones as those are the ones that were really in effect:

Folder:
 - max conflicts: -1 -> 10
 - min disk free: nothing/unlimited -> 1%

### To-do

 - [x] GUI to modify defaults. Plan is to reuse existing folder and device modals. (Required to really fix #4224).

 - [x] Port gui to untrusted.

optional:

  - [ ] Somehow handle versioning defaults. Currently you can set default versioning for new folders including parameters, but not default parameters for the different versioning types. Not sure whether that's really necessary and the GUI versioning code is a bit messy/verbose, so haven't looked into that yet.

### Testing

I extended the test checking defaults and did some manual fiddling around with config and UI, which looked good (but was likely far from comprehensive).


tomasz1986:
Does this also fix https://github.com/syncthing/syncthing/issues/6748 and https://github.com/syncthing/syncthing/issues/7033?

> Folder:
> 
> * max conflicts: -1 -> 10
> * min disk free: nothing/unlimited -> 1%

What about `maxConcurrentWrites` (supposed to be `2`, but the GUI sets it to `0`)?

imsodin:
That's not directly affected by this PR: The GUI didn't set it at all, like many other fields, which results in zero values. Handling that for `maxConcurrentWrites` has already been fixed in https://github.com/syncthing/syncthing/pull/7069

imsodin:
I adjusted the folder/device edit modals and accompanying JS to handle editing defaults too. Basically hid items for which defaults don't make sense (ids, remove buttons, ...).

They are accessible through the settings modal: Clicking on a button opens another modal to edit the settings, and dismissing that modal drops you back to the settings modal.

image

This is ready for review. Only thing left is mirroring the changes in the "untrusted GUI code".

tomasz1986:
> They are accessible through the settings modal: Clicking on a button opens another modal to edit the settings, and dismissing that modal drops you back to the settings modal.

I have been trying to test, but nothing happens when trying to click any of those buttons. I have compiled from https://github.com/imsodin/syncthing/tree/all/defaultConfig and also checked a few different browsers (Chromium, IE, Palemoon), but there is no difference. No reaction at all, and no JS errors either.

imsodin:
Works fine for me on firefox and chromium (linux) - no idea what the problem could be. There's no new source file I have forgotten to check in, nor do I have a "gui" dir in my config (you probably don't either, as you're seeing the new buttons). To debug you could inspect the buttons using developer tools, they should have attributes like `ng-click=editFolderDefaults()`. If that's there, find that function (or it's absence) in `syncthingController.js` and add a breakpoint to see what happens (or doesn't).

The gui changes are now also applied to the untrusted files, i.e. this is fully ready now. Except of course we need to find out why it doesn't work for @tomasz1986  :grimacing:

tomasz1986:
I have recompiled with the newest commits, and it is working now! I am not sure what the problem was about.

However, I have tried to fiddle around with the actual settings and something seems to be wrong…

1. I set the folder defaults as follows.

image

image

2. This is what I get when trying to create a new folder.

image

image

imsodin:
Thanks a lot @tomasz1986 !

I removed `Options.DefaultFolderPath`, that is now `Defaults.Folder.Path`. In the usage reporting I kept the variable, though it's a bit of an outlier now. Respectively we might add a new metric that checks for `reflect.DeepEqual(Defaults.Folder, 

> Also, just a side question, but I can see the current device's ID included under the `defaults` element in `config.xml`. Is there any reason or requirement for it to be there?

The local directory always is part of every folder, so to me it makes sense that it's also part of the default folder. All the gui and go code would add the local device anyway immediately if it was missing (i.e. there would have to be special code just to keep it out of the "default folder").

tomasz1986:
I am not sure whether this is a complete version, but I have tested the current state anyway.

Right now, the File Versioning and Advanced settings seem to be applied fine. The folder ID is again placed in the proper field. The Folder Label is still empty though.

There is also this strange glitch.

1. Set the default folder path to `.\abc`.
2. Start adding new folder, but cancel before saving.
3. The default folder path has now changed to `.\abc\abc`

imsodin:
> I am not sure whether this is a complete version, but I have tested the current state anyway.

Except for the glitches you find - it is :)

> There is also this strange glitch.

I cannot reproduce. Steps I took:

1. Open folder defaults, set default folder path to `\.abc`, save.
2. Click add folder. The displayed path is `\.abc/*folder-id*`.
3. Click close.

Result: The path in both the add folder dialog and the default folder config dialog is still `\.abc/*folder-id*` resp. `\.abc`

Probably unrelated: `.\abc` is strange to me. Is `\` a path separator here? And I do think relative paths work, as in "something happens" (and assume they are relative to the dir from which syncthing was started?), but it's not "on purpose", as in the desired behaviour is not defined and thus might change -> I would not recommend using it.

tomasz1986:
It must be `.\abc` (or possibly `./abc`, but not `\.abc` :grin:), and yes, the `\` is just a separator here, relative to the path, where Syncthing has been started.

Regardless of the relative paths, I think that I have found the culprit though.

1. Set the default folder label to `abc`
2. Set the default folder path to `.`

image

3. Try creating a new folder, but do not save it.
4. The default folder path has changed to `.\abc`.

image

The label is basically added to the path. Is this how it should behave? If yes, then why do we need both the label and the path?

I have also found another problem. Nothing is displayed under Sharing, even though I have multiple devices connected.

image

tomasz1986:
I have tested the commit in its current state. The folder path and folder sharing issues are gone :clap:.

The only remaining problem that I can see at the moment is that default folder label does not seem to work at all. When trying to add a folder, the label is just empty, regardless of the default one.

tomasz1986:
I have tested the newest compilation, and everything seems to work as intended 👍.

I have just two questions/comments.

1. If you compare the folder defaults screen and the add folder screen, then there is one obvious difference. That is, there is no "Ignore patterns" tab in the former. I am not asking you to do it, but would there be any technical reason not to include the default ignore patterns in `config.xml` too?

2. #6068 seems be a closed PR related to a bug in QUIC. Is it really related/fixed by this commit?

imsodin:
1. That mostly just feels out of scope for this PR: It takes existing config elements to store defaults. Ignores are not part of config. That could be made part of config, or as a separate file. Then there's the question whether we want defaults as in a template, or additionally "common" ignore patterns, as in patterns that act as if they were included. Basically a soon as you introduce ignores, it gets complicated/bikesheddy. Lets get this done and then we'll see what comes after ;)
2. That's a typo indeed, should have been #6086. And #6748 is at least related too.

tomasz1986:
1. Yeah, I was just thinking about what would be nice to have, regardless of the technical complexity. Especially since if you do integrate ignores like that, then then the question comes up, why do we need `.stignore` at all? Anyhow, as someone with tons of folders, I am definitely looking forward to having this PR implemented, no matter if without ignores.

2. Please add #7033 to the mix too 🙂. It is already partially fixed (regarding `maxConcurrentWrites`), and this commit fixes the rest.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1856,37 +1935,46 @@ angular.module('syncthing.core')
...
-            $scope.ignores.disabled = false;
-            $scope.editFolderModal();
         };
 
+        function addFolderInit(folderID) {
+            $scope.editingExisting = false;
+            $scope.editingDefaults = false;
+            return $http.get(urlbase + '/config/defaults/folder').then(function(p) {
+                $scope.currentFolder = p.data;
+                for (var k in $scope.versioningDefaults) {
```
AudriusButkevicius, line 1963: Why are these not stored in the folder config?

imsodin: Two parts: One was a bug in the current PR, the other is a conscious omission:  
Bug: The default versioning type including params is stored in `config.defaults.folder.versioning`, however wasn't applied when adding a new folder. That's now fixed, along with moving the gui-only parameters into `currentFolder._guiVersioning`.  
Omission: Versioning parameters are currently "free-form" (`map[string]string`), which prevents using `ext.default` in proto. The consistent thing to do would be creating proto types for all of these and adding a `Config.Defaults.VersioningParams` with a new map-like struct (type -> params). That just feels like a lot of work for little gain (defaults for something that's off by default).

tomasz1986:
Do you think that it would be safe to use this commit as it is in production right now?

I have been doing a lot of folder reorganisation and such recently, and their configuration is a little of a struggle. Do you expect any major changes in the future commits that could break the current state (e.g. changes to the `config.xml` structure, etc.)?

calmh (approved):
Awesome

acolomb:
Was trying to check this out for local inspection.  @imsodin I can't find this branch in your fork anymore?

calmh:
The set of configurable options in the default settings dialog is somewhat limited; should these new pseudo-device/folders perhaps show up in the advanced config editor as well?

PR #7132, opened 2020-11-19
lib/model: Index racyness between conns and folders (ref #7100)

imsodin:
More fixes regarding races between folders being started and cluster configs being processed based on the logs provided in https://forum.syncthing.net/t/device-stays-out-of-sync-until-restart/15891

So far I have looked at incoming cluster-configs only: #7100. That fix works, but it's not an unclean fix: Index senders are started without index info based, as the folder/device is new. That's correct for the starting sequence, but it's still bad because we don't store the index id. Meaning on a successive connect, we will re-exchange all index data. That is fixed by always storing the index data, even if we don't share the folder so far.

There's also an issue with sending cluster configs: When a folder is not paused, but hasn't been started yet, we send an info that says that the folder is running, but there's no index id/sequence. The remote then drops all info on us, while we happily start sending from a non-zero starting sequence once the folder is started. Thus the remote will be missing index info from us. For as long as we don't reset deltas. This problem likely existed for a long time, not just since sending multiple CCs like the problem above that are related to receiving CCs.  
The fix for this is to pretend the folder is paused if it isn't running, even if it isn't paused in config. Then when it is actually started we'll resend the cluster-config including the proper index info.

### Testing

"Alex from the forum" (aka the bug-reproducing log-delivering hero) will see if this prevents them from making remote devices out-of-sync.

### DIFF
--- a/lib/model/indexsender.go
+++ b/lib/model/indexsender.go
@@ -242,12 +242,6 @@ func (r *indexSenderRegistry) add(folder config.FolderConfiguration, fset *db.Fi
 	r.mut.Unlock()
 }
 
-func (r *indexSenderRegistry) addNew(folder config.FolderConfiguration, fset *db.FileSet) {
-	r.mut.Lock()
-	r.startLocked(folder.ID, fset, 0)
-	r.mut.Unlock()
-}
-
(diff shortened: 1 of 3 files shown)


imsodin:
Closing in favour of https://github.com/syncthing/syncthing/pull/7133 and https://github.com/syncthing/syncthing/pull/7134: Same production code changes, but the two independent issues are separated. There's also two separate unit tests for the two issues now.

PR #7135, opened 2020-11-20
lib/db: Reset index-id when dropping device

imsodin:
When we drop all index information on a device, we should also make sure their index id is zero for consistency. This ensures a subsequent full index exchange (which we need, as we just dropped all info) even if something goes wrong (like incoming index updates after dropping: #7134).

### DIFF
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -74,6 +74,13 @@ func (s *FileSet) Drop(device protocol.DeviceID) {
 		// announced from the remote is newer than our current sequence
 		// number.
 		s.meta.resetAll(device)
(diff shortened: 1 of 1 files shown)


calmh (approved):
I suspect this is an unnecessary no-op as `meta.resetAll` should remove the sequence, and announcing index-id + sequence=0 is functionally equivalent to not announcing an index id...

imsodin:
Yes, when everything is working correctly it is a noop. However if we do receive index updates after dropping (which shouldn't happen but can happen), the sequence will then become non-zero -> no longer equivalent.

PR #7136, opened 2020-11-20
gui: Remove probing for remote GUI address (ref #7017)

calmh:
This removes the probing for the remote side, instead making it so we
simply construct the address based on the currently connected address,
if any, and let the user sort out whether it works or not.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -806,8 +806,8 @@
                         <th><span class="fas fa-fw fa-desktop"></span>&nbsp;<span translate>Remote GUI</span></th>
                         <td class="text-right" ng-attr-title="Port {{deviceCfg.remoteGUIPort}}">
                           <!-- Apply RFC6874 encoding for IPv6 link-local zone identifier -->
-                          <a ng-if="idToRemoteGUI[deviceCfg.deviceID]" href="{{idToRemoteGUI[deviceCfg.deviceID].replace('%', '%25')}}">{{idToRemoteGUI[deviceCfg.deviceID]}}</a>
-                          <span ng-if="!idToRemoteGUI[deviceCfg.deviceID]">Unreachable</span>
+                          <a ng-if="hasDeviceGUIAddr(deviceCfg)" href="{{deviceGUIAddr(deviceCfg).replace('%', '%25')}}">{{deviceGUIAddr(deviceCfg)}}</a>
+                          <span ng-if="!hasDeviceGUIAddr(deviceCfg)">Unreachable</span>
                         </td>
                       </tr>
                     </tbody>
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -806,8 +806,8 @@
                         <th><span class="fas fa-fw fa-desktop"></span>&nbsp;<span translate>Remote GUI</span></th>
                         <td class="text-right" ng-attr-title="Port {{deviceCfg.remoteGUIPort}}">
                           <!-- Apply RFC6874 encoding for IPv6 link-local zone identifier -->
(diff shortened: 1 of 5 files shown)


imsodin (dismissed):

gui/default/index.html
```diff
@@ -806,8 +806,8 @@ <h4 class="panel-title">
                         <th><span class="fas fa-fw fa-desktop"></span>&nbsp;<span translate>Remote GUI</span></th>
                         <td class="text-right" ng-attr-title="Port {{deviceCfg.remoteGUIPort}}">
                           <!-- Apply RFC6874 encoding for IPv6 link-local zone identifier -->
-                          <a ng-if="idToRemoteGUI[deviceCfg.deviceID]" href="{{idToRemoteGUI[deviceCfg.deviceID].replace('%', '%25')}}">{{idToRemoteGUI[deviceCfg.deviceID]}}</a>
-                          <span ng-if="!idToRemoteGUI[deviceCfg.deviceID]">Unreachable</span>
+                          <a ng-if="hasDeviceGUIAddr(deviceCfg)" href="{{deviceGUIAddr(deviceCfg).replace('%', '%25')}}">{{deviceGUIAddr(deviceCfg)}}</a>
+                          <span ng-if="!hasDeviceGUIAddr(deviceCfg)">Unreachable</span>
```
acolomb, line 810: If we don't do any probing, I don't think that "Unreachable" is a sensible information.  If you mean to indicate there is no suitable connection to construct an address from, maybe we should change the wording. Something like "No connection"?

And I thought the discussion turned out in favour of the "remote GUI" naming, also for variables / functions?

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1140,6 +1065,28 @@ angular.module('syncthing.core')
             return '?';
         };
 
+        $scope.hasDeviceGUIAddr = function (deviceCfg) {
```
acolomb, line 1068: `hasRemoteGUIAddress`?

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1140,6 +1065,28 @@ angular.module('syncthing.core')
...
+            if (!deviceCfg.remoteGUIPort)
+                return false;
+            var conn = $scope.connections[deviceCfg.deviceID];
+            return conn && conn.connected && conn.address && conn.type.indexOf('Relay') == -1;
+        };
+
+        $scope.deviceGUIAddr = function (deviceCfg) {
```
acolomb, line 1075: `remoteGUIAddress`?

acolomb (changes requested):
Looks good overall, just some wording suggestions.

gui/default/index.html
calmh: Perhaps "Unknown"?

acolomb: Actually we do know.  Either there is no connection, or it's known not to be related to the host itself, but a relay server. So "No connection" is still my favorite, for the second case "No direct connection" would be more accurate.

calmh: Well, there is a remote device and it has a GUI, we just don't know how to connect to it. We also use it for other things like unknown folder status so it's already there and translated. "No [direct] connection" would be more like the underlying reason we don't know the GUI address.

acolomb: But the heading "Remote GUI" is not about an address.  The fact is that we do not know how to (hopefully) reach it, which is somewhere between "Unreachable" and "Unknown".  Maybe get more explicit and say "Address unknown"?

imsodin (approved):

acolomb:
My last review comment was not addressed yet.  Please reconsider the wording because an "Unknown" remote GUI is not a sensible information.

calmh:
Well, I at least disagree, I think it's fine and quite unambiguous.

PR #7145, opened 2020-11-24
Ignore gui

### LINKED ISSUE #4729: Overview of all included and/or excluded files

christianprescott:
### Purpose

This change adds a browser for the global state of a folder and reflects which files and directories are ignored. I aim to close #4729, not to build next gen ignores.

It does:
present a browser of files and directories
show each file's status as unaffected, ignored, or included (by a !negated pattern)
match files to simple patterns (anchored at folder root, identifying a specific file)
use client-side parsing and matching for patterns
make use of /rest/db/browse in its current form
have some tests
It will soon:
edit ignore patterns with a checkbox or toggle button
detect advanced patterns, and decline to show the browser view when present
be better integrated with the ignores UI - I don't intend to keep it in its own tab
It does not:
handle "advanced" patterns - wildcards, case insensitivity, etc
prepare for larger scale - large number of files or ignore patterns, pagination
use fancytree for the browser view

### Testing

This change includes Javascript tests! I am happy to split their addition out into a separate issue. I am not familiar with the project's build practices and would need some direction to get them running in CI.

Tests should help build confidence in a complex client side pattern-matching change. See the 'matchingPatterns' block of browseController.test.js for the examples demonstrating which patterns match which files. The controller is tested thoroughly but template is not - I think that's acceptable, it's relatively thin.

To test this change manually, open the folder edit modal and go to the "browse" tab.

### Documentation

I have a change prepared to add a Testing section to syncthing/docs dev/web.rst. Just a couple sentences and a code block describing how to run the karma tests.

It doesn't look like there's a home for any documentation on using the browser. Maybe the Ignoring Files page?


AudriusButkevicius:
So I haven't looked at the code, but I always felt that the only sensible way to do this would be a tree view, as otherwise there will be tons of clicking on the breadcrumb line just to navigate

christianprescott:
This table+breadcrumbs UI just developed organically as I hacked at this. Tree requires a more complex structure of files, or loading them all at once, and seemed limiting when this might require some unusual button behavior. At least those were the assumptions I made without having tried to use fancytree :) I could be convinced it's the superior method!

christianprescott:
Peek 2020-11-30 14-03

Here's the current state of the change, viewing and modifying simple ignore patterns. I've updated my github notification settings so will be responding more promptly.

AudriusButkevicius:
@calmh @imsodin wdyt?

calmh:
Fancy. Looks reasonable to me. I'd probably put the patterns above the "tree" (or whatever we call it) widget, as the pattern text box is fixed size and the tree widget can have unbounded height? (Didn't look at any code either.)

imsodin:
You a wizard? :)

The current breadcrumbs approach seems quite usable to me. I'd probably prefer a tree too - less up-and-down mouse movement (even though I wouldn't use the mouse :P ). And a very cursory glance at fancytree suggests it does what we need: tri-state selection and lazy-loading. However it might have other drawbacks and increased complexity. Basically with my current, limited knowledge of this I consider both approaches viable.

Some points I noted on a very cursory glance at the code:

 - I didn't see any ordering logic. E.g. if I first include "/foo/bar" and then ignore `/foo`, it looks like `/foo` will be inserted first. Which makes the latter `!/foo/bar` ineffective.

 - Due to problems like the above, I believe it's really hard to let the same patterns be used for manual patterns and to be controlled through this new GUI. There have been proposals of "comment fences", to separate the automagically managed patterns from manual once. Maybe that's helpful/needed here - I'll need to have a close look at the logic (and maybe add some nasty test-cases - great to see tests :) ) to get a clearer picture for myself.

christianprescott:
Not a wizard just an unemployed web dev

OK I'll continue with the breadcrumbs UI for now, but will experiment with fancytree and see if the same features are compatible. To @imsodin 's points:

1. Yes, it is naive in this respect can create make some non-obvious changes. It should be possible to insert them with respect to specificity - I'll add a test :)
2. I plan to add a switch between basic (browser) and advanced (textarea) modes as discussed in #5132, showing only one at a time and disabling access to the basic view entirely when advanced patterns are present. I think that addresses your concern with manual patterns? It hides the text box for scroll concerns, ensures we only modify pattern set when they're simple enough to understand (no need for comment fence), and avoids problems with rendering them in the browse.

gui/default/syncthing/folder/browseController.js
```diff
@@ -0,0 +1,56 @@
...
+        // Reference to browse data for the current folder
+        self.browse = undefined;
+        self.fileMatches = undefined;
+
+        $scope.$watch(function() {
+            return self.folder.id;
+        }, function (newId) {
+            if (newId) {
+                self.browse = Browse.forFolder(newId);
+                self.fileMatches = FileMatches.forFolder(newId);
```
christianprescott, line 26: This `forFolder` pattern made sense back when I thought there would be a need to cache data for multiple folders at once but there is not. It has become repetitive and verbose, clumsily used in different ways for each service. I will remove this so the service holds just one dataset at a time. If there's ever a need to cache or separate the data it can be added in a more object oriented fashion.

package.json
```diff
@@ -0,0 +1,8 @@
+{
```
christianprescott, line 1: javascript lol. It's conventional to commit this and the accompanying 3K lines of json lockfile, but it feels out of place in a golang project, where the npm dependencies are used only for the test environment and not in build output. Not sure if there's a better way to structure these test dependencies.

christianprescott:
Peek 2020-12-02 20-28

Here is what that advanced mode toggle looks like.

I will try out fancytree later this week. Otherwise this is feature complete as much as I planned it and my remaining todo items are small - css, error messaging, translations.

I do have some general questions:
1. What's up with untrusted/? Lots of similar files. Do I need to make changes there?
2. Will I need to take any action to get strings into the catalog/transifex?
3. Just noticed the `pathSeparator` value in the API. Are there problems on some platforms with my using '/' for paths?

tomasz1986:
I have not tested this yet, but how do you deal with a lot of folders/files in a folder? Are all of them displayed at once?

What happens if you have e.g. 10,000 files in a folder? Is the GUI able to withstand displaying all that at once? What is the RAM overhead?

christianprescott:
As it stands, I don't deal with them. The rest/db/browse endpoint does not
offer pagination, so a folder with 10,000 files will render a table with
10,000 rows in the order returned by the API. For a large folder that might
be CPU intensive for the browser! I don't think it is unreasonably memory
hungry though.

christianprescott:
Peek 2020-12-06 18-30

Here is the change using fancytree. 
	+ Maintains tree structure without having to keep a matching data structure ourselves. yay!
	+ Should have better performance out of the box when making changes to ignores
	- Model is not bound to angular/view state, so it's tricky to integrate with other parts of the app. Might be prone to bugs but that concern may be unfounded. (for example, I'm having difficulty resetting the tree when changing folder edit modals)
	- Less unit testable - but practically only the `toggle` and rendering functions are uncovered
So this seems just fine if you'd prefer the tree UI. I still need to fix the bug mentioned above.

AudriusButkevicius:
Looks good. I know there were some issues with resetting the trees state that I had elsewhere, I think in version restore UI, so perhaps there is something there to be inspired by.

Also, I think fancy tree was just a suggestion of something we used before that seem to have worked, but I think you have more expertise in the UI area than all of the maintainers put together, so perhaps there are better, angular specific components that we could use.

I guess we use an ancient version of angular so it might be hard to find something that works, but I know for sure that angular material has something like this that would work.

One thing that looks strange in the example is the fact that the checkboxes are left aligned, rather than aligned with the item, but I guess that's how fancy tree renders it?

I guess if you feel it's polished enough, let us know, and we'll start digging through the code.

christianprescott:
I am certainly satisfied enough with fancytree to favor it here over adding a different tree package. This is my first encounter with angular so I'll leave version upgrade or component library for another day :)

This is ready for review.

gui/default/syncthing/folder/ignoresService.js
```diff
@@ -0,0 +1,207 @@
...
+        function parsePattern(line) {
+            var stripResult = stripPrefix(line.trim());
+            var prefixes = stripResult[0];
+            var hasPrefix = prefixes['(?i)'] || prefixes['(?d)'];
+            var path = toPath(stripResult[1]);
+            var matchFunc = !hasPrefix && matcher(path);
+
+            return {
+                text: line,
+                isSimple: !!matchFunc,
```
AudriusButkevicius, line 125: Again, unfamiliar with js, is this a not not, or not null?

gui/default/syncthing/folder/ignoresService.js
```diff
@@ -0,0 +1,207 @@
...
+            self.data.text = 'Loading...';
+            self.data.error = null;
+            self.data.disabled = true;
+            return getIgnores(folderId).then(function (response) {
+                self.data.text = response.map(function(r) { return r.text; }).join('\n');
+                self.data.patterns = response;
+                self.data.disabled = false;
+                return self.data;
+            }).catch(function (err) {
+                self.data.text = '';
```
AudriusButkevicius, line 51: Doesn't set data.error?

gui/default/syncthing/folder/ignoreTreeService.js
```diff
@@ -0,0 +1,100 @@
...
+            };
+        };
+
+        function toggle(node) {
+            var absPath = '/' + node.data.file.path;
+            if (node.data.match) {
+                var match = node.data.match;
+                if (absPath === match.path) {
+                    // match is exact match to this file, remove match from patterns
+                    Ignores.removePattern(match.text);
```
AudriusButkevicius, line 72: Shouldn't this check if the pattern is negated?
Sure, it might be an exact match, but it might be negated, at which point removing it would do the wrong thing?

Also, I guess "toggle" doesn't actually say whether it was enabled or disabled, so we're just assuming that we got things right from the get go?

gui/default/syncthing/folder/ignoreTreeService.js
```diff
@@ -0,0 +1,100 @@
...
+
+            return {
+                // Fancytree keys
+                title: file.name,
+                selected: (!match || match.isNegated),
+                key: file.path,
+                lazy: !file.isFile,
+                folder: !file.isFile,
+                // Data keys
+                file: file,
```
AudriusButkevicius, line 61: I think file should be path here? As we're calling buildNode for directories too?
But then you have path.path, so perhaps this should be entry or something like that.

gui/default/syncthing/folder/ignoreTreeService.js
```diff
@@ -0,0 +1,100 @@
...
+            if (node.data.match) {
+                var match = node.data.match;
+                if (absPath === match.path) {
+                    // match is exact match to this file, remove match from patterns
+                    Ignores.removePattern(match.text);
+                } else {
+                    // match is parent directory of file
+                    // If the parent pattern is negated, add pattern ignoring this file
+                    var prefix = match.isNegated ? '' : '!';
+                    Ignores.addPattern(prefix + absPath);
```
AudriusButkevicius, line 77: Same question, how do we know we're not unignoring here?

AudriusButkevicius (commented):
I've raised some questions, which I guess mostly arise from the complexity this tries to tackle, but in general looks good.
The code is neat and clean.

I guess once we've settled on the comments I might want to take this for a spin to see how this works before we merge this.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1822,24 +1808,27 @@ angular.module('syncthing.core')
...
-                    $scope.ignores.disabled = false;
-                })
-                .error(function (err) {
-                    $scope.ignores.text = $translate.instant("Failed to load ignore patterns.");
-                    $scope.emitHTTPError(err);
-                });
+            $scope.ignores = Ignores.data;
+            $scope.currentFolder.ignoreIsEditingAdvanced = true;
+            Ignores.refresh($scope.currentFolder.id).then((response) => {
+                $scope.currentFolder.ignoreIsBasic = response.patterns.every(function (p) { return p.isSimple; });
```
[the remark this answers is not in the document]
christianprescott: MDN reports it's widely available, going back as far as IE9 circa 2011. Is that acceptable? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

gui/default/syncthing/folder/ignoresService.js
```diff
@@ -0,0 +1,207 @@
...
+                matchFunc: matchFunc,
+            };
+        };
+
+        // Adapted from lib/ignore/ignore.go@parseLine
+        function stripPrefix(line) {
+            var seenPrefix = { '!': false, '(?i)': false, '(?d)': false };
+
+            var found;
+            while (found !== null) {
```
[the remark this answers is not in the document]
christianprescott: Agreed, this is clunky. I will revise it.

gui/default/syncthing/core/syncthingController.js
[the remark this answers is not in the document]
tomasz1986: I do not think that anyone tests Syncthing in anything older than IE11… and I am probably the only one testing in that one 😉.

gui/default/syncthing/folder/ignoresService.js
christianprescott: This is a not not. Coercing the absence/presence of matchFunc into a boolean.

gui/default/syncthing/folder/ignoresService.js
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        // Return a function that can be applied to match the pattern
+        function matcher(line) {
+            if (line.length === 0) return neverMatch;
+            if (line.indexOf('//') === 0) return neverMatch; // comment
+            if (line.indexOf('/') !== 0) return null; // not a root line
+            if (line.length > 1 && line.charAt(line.length - 1) === '/') return null; // trailing slash
+
+            line = line.replaceAll(/\\[\*\?\[\]\{\}]/g, '') // remove properly escaped characters for this evaluation
```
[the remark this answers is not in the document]
christianprescott: This function is inspecting the pattern to determine if we can match it at all. For this change, there is no attempt to match glob patterns, only patterns with plain file path and anchored at folder root with a leading slash. So here we're only doing the matching to check if the pattern is more complex than that, and if it is we return no matcher function (null) and indicate this is an "advanced" pattern we won't attempt to render in the UI or modify the pattern.

I'll improve the comment on this function. The "identifies simple patterns" block in ignoresService.test.js should help understand what patterns this service does and does not understand/render.

gui/default/syncthing/folder/ignoreTreeService.js
AudriusButkevicius: Yes, the issue is about the name of the variable.

gui/default/syncthing/core/syncthingController.js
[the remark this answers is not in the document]
AudriusButkevicius: Fine.

gui/default/syncthing/folder/browseService.js
```diff
@@ -0,0 +1,71 @@
...
+
+            var items = [];
+            for (var name in data) {
+                var isFile = Array.isArray(data[name]);
+                var item = {
+                    name: name,
+                    path: pathPrefix.concat([name]).join('/'),
+                    isFile: isFile
+                };
+                items.push(item);
```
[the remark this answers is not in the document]
AudriusButkevicius: I can try and propose a PR myself for the changes that I think should happen, I guess whether changes to use it would need to be made in this PR or a follow up PR would depend on the merge order, if that's ok with you?

gui/default/syncthing/folder/browseService.js
```diff
@@ -0,0 +1,71 @@
...
+
+        function browseList(data, prefix) {
+            var pathPrefix = []
+            if (prefix.length > 0) {
+                pathPrefix.push(prefix);
+            }
+
+            var items = [];
+            for (var name in data) {
+                var isFile = Array.isArray(data[name]);
```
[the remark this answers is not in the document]
AudriusButkevicius: Jesus, I put that endpoint in years ago, I wish I could travel in time and slap myself for such poor design.

gui/default/syncthing/folder/ignoresService.js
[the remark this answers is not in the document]
AudriusButkevicius: I think the explanation you gave in your reply to my comment could be copied above that code line, as it made it much more clear.

gui/default/syncthing/folder/ignoreTreeService.js
AudriusButkevicius: I guess I need to try this out to understand how it works and how it interacts with existing patterns, but this raises a few questions/points.

1. It seems that in the case above, we would add a redundant pattern, is that the case, or the example is just to explain handling of pre-existing text patterns placed there by the user?
2. We should try to "collapse" patterns as much as we can, as it's N files in a folder matched against M patterns.
3. What happens when the existing text based patterns created by a user are a mess? I guess the tree ui does not try to simplify it?

gui/default/syncthing/folder/browseService.js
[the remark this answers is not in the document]
christianprescott: Yes that's ok with me

gui/default/syncthing/folder/ignoreTreeService.js
christianprescott: Got it - there are a number of places file is referred to this way, I'll go with entry instead

gui/default/syncthing/folder/ignoreTreeService.js
christianprescott: 1. It is to explain handling of pre-existing text pattern, but it's _possible_ you could end up with redundant patterns from the UI, with the right sequence of changes.
2. Collapsing would resolve that issue. I'll investigate that. The bones are there to determine when one pattern's path is the child of another so I hope it is not too complicated.
3. The UI will make no changes to the patterns until a file is toggled in the UI. Even for messy patterns, it should be representing their effects accurately though!

christianprescott:
I have not yet addressed the "collapse redundant patterns" discussion. It should be doable and with an acceptable performance cost - because even for large pattern sets the one time collapsing modification to patterns should pay off with fewer pattern comparisons when updating rendered file matches in the UI.

I hesitate because the collapse could be a destructive action. Imagine you have ignored dozens of specific directories inside a parent directory. Ignoring the parent directory would collapse the patterns matching the children and they'd be lost if you toggled the parent back to included. There is the option to Close the modal and cancel the edit, if you noticed the unintended effects and wanted to restore the child patterns. I suppose it's more clear than the repeated checkbox clicks described in the review comment.

christianprescott:
Peek 2021-01-17 17-47

Here are some rules being "collapsed" when a parent one is modified. Note the child patterns are destroyed. When you're ready to try it out give some attention to whether toggling a file or directory has the effect you expect, because now IgnoresService makes some extra modifications to the patterns for this feature.

All comments have been addressed, ready for another round of review.

AudriusButkevicius:
Sorry, I've been really bad at keeping up with github stuff lately. I'll try to get to this soon.

gui/default/syncthing/folder/ignoresService.js
```diff
@@ -0,0 +1,257 @@
...
+            return line;
+        }
+
+        // Inspect the pattern to determine if it is simple enough to match,
+        // then return a function that can be applied to match a path.
+        function chooseMatcher(line) {
+            if (line.length === 0) return neverMatch;
+            if (line.indexOf('//') === 0) return neverMatch; // comment
+            if (line.indexOf('/') !== 0) return null; // not a root line
+            if (line.length > 1 && line.charAt(line.length - 1) === '/') return null; // trailing slash
```
AudriusButkevicius, line 203: There is also `#include` directive that doesn't seem to be handled here.

AudriusButkevicius: Also, we have crazy stuff like:
```
foo/*/bar                   # any directory at the same level
```
or even
```
pics/**/Thumbs.db           # any directory at any level
```
Which probably needs to abort if it sees it?

gui/default/syncthing/folder/ignoresService.js
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        /*
+         * pattern matching functions
+         */
+
+        // A prefix match is anchored at folder root and is an exact match to
+        // one file or directory
+        function prefixMatch(filePath) {
+            var patternPath = this.path
```
AudriusButkevicius, line 241: This is magical, I assume this is some prototype voodoo that adds the method on the object?

AudriusButkevicius (commented):
So I filed my PR #7306 which makes the API slightly more sane given we're starting to use it.

I somewhat skimmed the code and it looks fine to me on the surface. One part could be a separate PR, but I am not too fussed.

There is obviously a lot of potential for edge cases but we won't know till we battle test it I guess.

@calmh @imsodin what's your view on how to get this merged?

I guess this is a new extra feature with limited damage surface, as even if it doesn't work, users can go to the old way, so I think it's fairly safe to merge it?

gui/default/index.html
```diff
@@ -385,22 +385,22 @@ <h4 class="panel-title">
                   <table class="table table-condensed table-striped table-auto">
                     <tbody>
                       <tr ng-show="folder.label != undefined && folder.label.length > 0">
-                        <th><span class="fas fa-fw fa-info-circle"></span>&nbsp;<span translate>Folder ID</span></th>
-                        <td class="text-right no-overflow-ellipse">{{folder.id}}</td>
+                        <th class="col-fixed"><span class="fas fa-fw fa-info-circle"></span>&nbsp;<span translate>Folder ID</span></th>
+                        <td class="col-flex text-right no-overflow-ellipse">{{folder.id}}</td>
```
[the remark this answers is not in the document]
christianprescott: It's related in that we want to use the same table cell styles here and in the fancytree table, but don't want the style tied to the cell type nor the order of the cells. So the styles in overrides.css are refactored to use an explicit class name "col-fixed" instead of the element type "th".

It could be a separate PR, but it would be a strange one because the refactor has no effect on its own. I will investigate a different approach to this CSS, with the previous behavior as default and the option to specify a class and override it.

gui/default/syncthing/folder/ignoresService.js
christianprescott: Yes, `this` is magical :) Technically prototype isn't involved here, simply calling a function on an object sets `this` to the object it's called on. It can be bad form to use `this` when the context is ambiguous, but in this case the matchFunc assignment and call are done in a way that is [clearly defined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this#as_an_object_method) so I'm okay with it.
(An alternative approach might be to use `prefixMatch.call(pattern, filePath)` to set `this` to `pattern` explicitly, but that has the same effect as `pattern.prefixMatch(filePath)` while being harder to read)

gui/default/syncthing/folder/ignoresService.js
christianprescott: The two wildcard examples are handled on lines 205-211. After removing special characters that have been escaped `\*` we check for special characters including `*`. If any are present, returning null indicates this is an advanced pattern and we don't know how to match it to files.

There is an important distinction between the neverMatch function and null (no function): the former is for a pattern our UI can safely work around as we add and remove patterns, the latter is for a pattern too complex to render or modify with the UI. Presence of advanced patterns is used in syncthingController.js to determine whether to disable the UI entirely, because adding patterns on the wrong side of a glob could have unintended effects.

This is an ugly function. For now, I'm not sure of a simpler way to inspect the pattern than a series of `if`s for each check. I hope the comments and tests (ignoreService.test.js:264 in this case) can help.

christianprescott (commented):
I think I can remove the CSS noise from the diff, and I'll take a look at #include and the API PR.

gui/default/syncthing/folder/ignoresService.js
christianprescott: Regarding #include, I think we have to disable the browse UI when it is present.

I was curious whether the API offers enough information to render it in some way. The rest/db/ignores endpoint reflects the patterns from the file in the `expanded` field, but `ignore` field contains only the text representation "#include patterns.txt". Perhaps if there was a way to link the #include line to the patterns inside, we could show the effects and... disable modifications to paths affected by the #included file? Insert new patterns before the #include command? Not sure. This is speculation about something to support in the future - for now an "#include" command will be parsed as an advanced pattern and have the same effects as a glob pattern.

calmh:
I haven't followed the technicals or done any review (and don't necessarily intend to) but I agree that as a GUI-only change with a fallback (there will still be a text format editor?) it's limited risk. Merge when you guys are happy. If this now depends on #7306 or should take advantage of #7306, I guess merge that first?

imsodin (commented):
I do like this code - and I haven't even looked at the tests yet :)

A though about future extension - not necessary to address in this PR, but to keep in mind to prevent the current implementation going against that (it doesn't as far as I see):  
I expect having some sentinel comment(s) to mark the patterns managed by the tree to become necessary soon. One basic use-case that I expect people using the tree to expect, is ignoring file extensions. That's not currently possible, as those aren't simple patterns right now.

gui/default/syncthing/folder/ignoresService.js
```diff
@@ -0,0 +1,257 @@
...
+        'use strict';
+
+        var self = this;
+
+        // public definitions
+
+        self.data = {
+            // Text representation of ignore patterns. Updated when patterns
+            // are added or removed, but modifying `text`` does not update
+            // `patterns`.
```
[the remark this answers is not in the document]
christianprescott: No. This comment is kind of misleading, I will update to: "modifying text does not update patterns until parseText is called"

After editing text in advanced mode and calling parseText, *all* lines are parsed into pattern objects. No data is lost, and when making changes the tree only modifies the patterns simple enough to understand.

gui/default/syncthing/folder/ignoresService.js
```diff
@@ -0,0 +1,257 @@
...
+            // Find the last pattern more specific than newPattern. Because
+            // the first matching pattern is applied to a file, we need to
+            // insert newPattern in a position that doesn't override existing
+            // patterns.
+            var afterIndex = findLastIndex(self.data.patterns, function(pattern) {
+                return patternMoreSpecificThan(pattern, newPattern);
+            });
+            self.data.patterns.splice(afterIndex + 1, 0, newPattern);
+
+            // Remove any more specific patterns so the new pattern has the intended effect.
```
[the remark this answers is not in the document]
christianprescott: No. The way it works currently, making a change to a parent directory overwrites all patterns that apply to child directories. We can discuss a way to not overwrite the more specific patterns - there are some weird UI interactions to consider.

gui/default/syncthing/folder/ignoresService.js
```diff
@@ -0,0 +1,257 @@
...
+        // The changes made to the path make it more specific than the original
+        // pattern. It is useful for prefix matcher. Should a glob matcher be
+        // added this path should not be used to match because it does not
+        // represent the complete pattern. The most common effect of the changes
+        // is matching patterns like `*`, `**`, or `/*` without a special case
+        function toPath(line) {
+            // Add a leading slash when the pattern begins with wildcards
+            // because a pattern beginning with * or ** will match entries at
+            // the root (as well as those in child directories)
+            line = line.replace(/^\*+/, '/$&'); // `$&` inserts the matched substring
```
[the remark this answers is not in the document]
christianprescott: If there are still `*`s after trimming trailing ones on :193, then yes it will be ignored in chooseMatcher. Practically speaking, :189 is necessary to turn pattern `*` into `/*` so it is interpreted as the path `/` and matches all files.

gui/default/syncthing/folder/ignoresService.js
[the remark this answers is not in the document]
imsodin: Ah right. I missed the `parseIgnores` on losing focus in the html.

gui/default/syncthing/folder/ignoresService.js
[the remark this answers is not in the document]
imsodin: You are one step ahead - I wasn't even thinking about that.  
Overwriting child input on toggling a parent seems fine. I was worrying that the (internal) behaviour (recursing into ignored directories) breaks the tree - however as you sort with the least specific last, that's not an issue at all. I just needed a moment to get there :)

christianprescott:
Updated for new db/browse response structure.

Re:future Yes comment fence could be used one day. The complexity of handling patterns like globs or extensions comes from the UI making unintended changes, having widespread in the syncthing folder beyond the specific file the user is toggling. I can imagine one day adding a dialog for these patterns to confirm changes, explaining the impact. Or option to ignore other files with the same name or extension as the current one. Just speculating.

AudriusButkevicius:
@imsodin should we just merge this?

imsodin (commented):
Took it for a spin and it's really nice :)

The blue blackground of the tree was a bit distracting to me. I'd keep it white like everything else.

> It doesn't look like there's a home for any documentation on using the browser. Maybe the Ignoring Files page?

There's something in the introduction, but it's quite outdated I think (haven't looked at it in a long time). I think it fits best into ignoring files indeed.

> @imsodin should we just merge this?

Generally yes.  
We should talk about the build process for tech-ui and tests here, as both require npm. Having tests without running seems wasteful. We could back it out of the PR until we decided how to handle it, and add it later.

tomasz1986:
I have just done some testing with the ignore GUI. Here are some of my observations.

1. Windows path backslashes `\` seem to be completely unsupported. Trying to use them in paths automatically disables the `Basic` button.

   image

   Just a side note, but when the button becomes disabled, it would be nice to have some kind of an indicator as to why it is unavailable. Otherwise the users will likely be confused.

2. The whole "Edit Folder" GUI becomes extremely laggy with a large number of files/folders. I have tried to use it with 10,000 files, and it is barely usable. I am testing on a decent hardware in Chromium, so the situation will likely be much worse on slower devices. CPU spikes to 100% and RAM usage goes up significantly upon trying to view the ignore patterns.

3. Numbers are sorted as follows.

   image

   It would be nice to have them sorted by their numerical value instead, but I am not sure what exactly is responsible for the sorting behaviour.

4. There are some misalignment issues in the tree,

   image

   but the same is true when browsing Versioning, so I am not sure whether anything can be done about this here.

schuellerf:
Awesome feature - I found this code when I was motivated to write something similar myself.
@christianprescott any plans for the two features:
 * Show the "Basic" view in "readonly mode" when patterns are too complex to _edit_ in "basic mode"?
   (I think it should be pretty easy to just _show_ which files are included but avoid changing the check marks)
 * Switch between local and global state view?
   as this is important for me to see which local files are excluded

Jip-Hop:
Awesome development! I've been wishing for this functionality for years 🙂 As a stress test I synced a folder containing the Firefox source code. I expanded folders until I had more than 60.000 items in the tree. No problems (de)selecting items.

However I encountered one bug. When expanding a directory in the tree, it may use the wrong folder id when calling the rest api. You need two folders to test this. In the first folder browse the tree by expanding directories. Works fine. Then try the same on the second folder. It will fail to browse more than one level. You can see in the network requests panel that it uses the wrong folder id (id of the first folder even though we're inside the second folder now). In the screenshot you can see the current folder id at the top, and the wrong folder id used in the api call.

Screenshot 2021-04-13 at 19 19 37

Besides this I'd recommend a few usability improvements.

At the moment it looks like a whole folder is included because it's checked and marked blue, even though some of its sub folders are excluded. I think it's useful to distinguish 3 states:
1. Excluded, including all children (checkbox unchecked)
2. Partly excluded, but some of its children are checked (checkbox unchecked or partially checked, maybe different background color?)
3. Fully included, including all children (checkbox checked)

The tree is under the "Ignore Patterns" tab, but what we're selecting is being included (not ignored). This is a bit contradictory in my mind. I think a heading/label above the tree could help explain, for example: "Select items to sync".

Currently this ui is not presented when accepting a shared folder. So I'd have to manually set `.stignore` rules to `**` when accepting the folder and then go into folder settings to select folders I want to sync in the tree view. Not a big issue as I can work around it.

I'd be super happy if the current functionality would be merged. 🥳

christianprescott:
Thanks @tomasz1986 
1. Thank you for including the complete error message :) looks like I let some ES6 slip in those two places. Fixed. If there are further problems with this I'll look into adding some linting rules.
2. Changed to use the path separator from the /rest/system/status endpoint. I don't have a Windows machine on hand to test, but I would expect this change to work with Windows separators, and also across OS's? e.g. Loading the UI from a Windows server to a Linux browser.
3. **Performance issues are still outstanding**. Let's see what I can do about paginating client side. I suspect lots of time is spent by fancytree processing and rendering rows, not necessarily by fetching files. I will investigate ways to have fancytree do its loading in smaller batches using the existing syncthing API.
4. They're presented in the same value returned by the API. I would prefer not to make that change client side, especially if pagination becomes involved
5. Fixed here and in Versions tree. Vertically centered the icons.

@schuellerf 
1. No plans. In my eyes, until all types of patterns are supported by the UI the presence of any advanced patterns in a "readonly" mode makes it misleading because it would present an inaccurate/incomplete picture of ignored files. Besides, I think a more worthwhile way to support complex patterns would be to include them in the API, with the actual golang matching engine telling the client which pattern matches a file.
2. API support for this does not exist yet. /rest/system/browse almost does - but only presents directories. It's an interesting distinction though because when a file is ignored on all devices globally it disappears! And can't be included again until the ignoring pattern is removed. So this would be a good enhancement.

@Jip-Hop 
Good catch with the folder change - fixed.
Some descriptive text has been added: "Selected items will be synced, other items ignored" but we're still under the Ignore Patterns tab for now.
Using the current db/browse endpoint we can't display folder contents until the folder is present on our own syncthing device, so this isn't possible yet. However one enhancement I saw mentioned elsewhere was a toggle for top-level "ignore root" which would be pretty simple.

Jip-Hop:
Thanks! I like the descriptive text.

Tested again, this time on Windows. I confirm the folder switching is now working properly 🙂 

However (un)checking doesn't change the checkbox any more and duplicate ignore rules are added. See screenshots for results after clicking a checkbox multiple times.

Capture
Capture2

Would indeed be nice to show a top-level "ignore root" button (similar to manually typing the `**` rule). Currently `Basic` mode under `Ignore Patterns` is disabled in the `Add Folder` modal. I also suggest some explanation that after clicking `Save` it's possible to select which files/folders to sync form the Folder -> Edit -> Ignore Patterns -> Basic menu.

Have you considered multi-stage checkboxes, to not show 'fully checked' if at lease one child is excluded?

tomasz1986:
> Thanks @tomasz1986
> 
> 1. Thank you for including the complete error message :) looks like I let some ES6 slip in those two places. Fixed. If there are further problems with this I'll look into adding some linting rules.
> 2. Changed to use the path separator from the /rest/system/status endpoint. I don't have a Windows machine on hand to test, but I would expect this change to work with Windows separators, and also across OS's? e.g. Loading the UI from a Windows server to a Linux browser.
> 3. **Performance issues are still outstanding**. Let's see what I can do about paginating client side. I suspect lots of time is spent by fancytree processing and rendering rows, not necessarily by fetching files. I will investigate ways to have fancytree do its loading in smaller batches using the existing syncthing API.
> 4. They're presented in the same value returned by the API. I would prefer not to make that change client side, especially if pagination becomes involved
> 5. Fixed here and in Versions tree. Vertically centered the icons.

I have done some quick testing.

1. The GUI loads in IE 11, and it is possible to edit advanced ignore patterns, but basic ones do not load. There are still some JS errors.

   ```
   TypeError: Object doesn't support property or method 'find'
      at Anonymous function (http://localhost:8391/syncthing/core/syncthingController.js:1965:17)
      at processQueue (http://localhost:8391/vendor/angular/angular.js:13337:11)
      at Anonymous function (http://localhost:8391/vendor/angular/angular.js:13353:27)
      at Scope.prototype.$eval (http://localhost:8391/vendor/angular/angular.js:14589:9)
      at Scope.prototype.$digest (http://localhost:8391/vendor/angular/angular.js:14405:15)
      at Scope.prototype.$apply (http://localhost:8391/vendor/angular/angular.js:14694:13)
      at Anonymous function (http://localhost:8391/vendor/angular/angular.js:14984:11)
      at completeOutstandingRequest (http://localhost:8391/vendor/angular/angular.js:4959:7)
      at Anonymous function (http://localhost:8391/vendor/angular/angular.js:5347:7)
   ````

2. The `\` separators seem to work now, but there are issues with duplicate patterns and non-working checkboxes mentioned above by @Jip-Hop, so it is kind of difficult to test right now :wink:.

4. The same problems affect the Versions modal (see https://github.com/syncthing/syncthing/pull/7395 and https://github.com/syncthing/syncthing/issues/7393). I have tried some basic HTML and CSS tweaks, but there has been no difference so far. It is just the whole loading and processing that is super heavy, both on CPU and RAM. I would say that either pagination or some kind of lazy load will be required to make it usable with larger numbers of files. If you manage to achieve this here, then I'm quite sure that the same could be applied to the Versions modal to fix those issues there too 🙂.

   Also, I have added some tweaks to contain the box and make it resizable in https://github.com/syncthing/syncthing/pull/7395, which probably could be adapted to both the Versions and the Ignore Patterns tree at the same time.

5. I have done the same with

   ```
   span.fancytree-expander,
   span.fancytree-icon {
       vertical-align: middle !important;
   }
   ```
   in https://github.com/syncthing/syncthing/pull/7395. Not saying which code is better, but if my commit is applied first, then you probably will not need to have any icon aligning tweaks here (and vice versa).

st-review:
:robot: *beep boop*

acolomb:
I think this would be a really useful addition.  @christianprescott are you still interested in finishing the implementation?

acolomb:
Another ping @christianprescott, since you seem to have been active on GitHub again lately.  This being one of the most requested features in Syncthing, would you be willing to start working on it again?  I see you switched jobs around the time this PR went quiet.  Maybe now is a better time?

We also have some related progress toward a selective sync GUI using ignore patterns, see #9619 and its immediate user: https://forum.syncthing.net/t/beta-test-my-new-ios-app-for-syncthing/22457/1

PR #7149, opened 2020-11-24
build: Ignore noscript in translation-script

imsodin:
There were some strings without the translation attribute/directive and the script was complaining about the new noscript stuff (which can't be translated).

### DIFF
--- a/script/translate.go
+++ b/script/translate.go
@@ -35,7 +35,7 @@ func generalNode(n *html.Node, filename string) {
 	if n.Type == html.ElementNode {
 		if n.Data == "translate" { // for <translate>Text</translate>
 			translate = true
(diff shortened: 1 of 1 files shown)


calmh:
This adds translations for strings I removed in #7136 so I'd rather merge that first...

script/untrusted-gui.patch
```diff
@@ -0,0 +1,275 @@
+diff --git a/gui/default/index.html b/gui/default/untrusted/index.html
```
calmh, line 1: Unintentional?

imsodin: definitely! wtf?

calmh (approved):

PR #7150, opened 2020-11-24
all: Handle errors opening db/creating file-set (ref #5907)

imsodin:
This is a first step towards graceful error handling in the db layer. It introduces error return values to `NewLowlevel` and `NewFileSet`. That requires a mechanism in the model to handle fatal errors when starting folders, which is just a one-buffered channel for fatal errors that the model service receives from to initiate shutdown.

This means that errors which were previously panics and thus reported now result in a graceful shutdown, i.e. no panic reporting. Thus I introduced failure reporting in the db. The principle is that every package is responsible for reporting it's own failures, even if that failure results in a returned error. The reasoning is, that only at the error source you can know if the error needs reporting and whether it's actually safe to report (e.g. doesn't contain paths or ids).

Previously such fatal failures would hardly ever get reported, as the reporting service gets torn down due to the fatal failure before it sends them. Now unreported failures are written to disk when the service stops and these failures are then reported on next startup.


calmh:
zomg

ProactiveServices:
> zomg

@imsodin is a beast and nothing can stop him.

AudriusButkevicius:
> The reasoning is, that only at the error source you can know if the error needs reporting and whether it's actually safe to report (e.g. doesn't contain paths or ids).

Why not check if the error satisfies some reportable interface or something?

Having to push down the whole reporting mechanism stack down so deep feels icky.

imsodin:
> Why not check if the error satisfies some reportable interface or something?
> 
> Having to push down the whole reporting mechanism stack down so deep feels icky.

There are conditions in the db package where I'd like to generate a report, but that are handled internally. Like metadata counters becoming negative or empty globals. That's stuff that should never happen, but also isn't bad enough to justify making it fatal (or just recoverable).

lib/model/model.go
```diff
@@ -252,24 +254,41 @@ func NewModel(cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersio
...
+	close(m.started)
+
+	var err error
+	select {
+	case <-ctx.Done():
+	case err = <-m.fatalChan:
+	}
+
+	m.cfg.Unsubscribe(m)
+	m.closeAllConnectionsAndWait()
```
calmh, line 268: Defer these and return err/nil directly from the select, which becomes last in the method? I think this would be clearer than having err just be implicitly nil because it's unset, and keeps the unsubscribe with the subscribe.

lib/model/model.go
```diff
@@ -252,24 +254,41 @@ func NewModel(cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersio
...
 	clusterConfigDevices := make(deviceIDSet, len(m.cfg.Devices()))
 	for _, folderCfg := range m.cfg.Folders() {
 		if folderCfg.Paused {
 			folderCfg.CreateRoot()
 			continue
 		}
-		m.newFolder(folderCfg, cacheIgnoredFiles)
+		err := m.newFolder(folderCfg, cacheIgnoredFiles)
+		if err != nil {
+			m.fatal(err)
```
calmh, line 283: This will deadlock, because fatalChan is unbuffered and we only start reading from it after initFolders returns? Perhaps have initFolders return err instead and let the caller handle it?

lib/model/model.go
```diff
@@ -505,7 +533,11 @@ func (m *model) restartFolder(from, to config.FolderConfiguration, cacheIgnoredF
 			// Create a new fset. Might take a while and we do it under
 			// locking, but it's unsafe to create fset:s concurrently so
 			// that's the price we pay.
-			fset = db.NewFileSet(folder, to.Filesystem(), m.db)
+			var err error
+			fset, err = db.NewFileSet(folder, to.Filesystem(), m.db)
+			if err != nil {
+				return fmt.Errorf("restarting folder: %w", err)
```
calmh, line 539:
```suggestion
				return fmt.Errorf("restarting %s: %w", to.Decription(), err)
```

It may not be consistent, but I think some / many / all of these errors end up in m.fatal() and don't explicitly get the affected folder name added as context before printing / crashing / etc... So maybe we should always add the description to the error, worst case we get double the folder name in the final output but 🤷

lib/ur/failurereporting.go
```diff
@@ -65,37 +67,20 @@ type failureStat struct {
...
-		case <-ctx.Done():
-		}
-	}()
 	h.cfg.Subscribe(h)
 	defer h.cfg.Unsubscribe(h)
 
-	var url string
+	url, sub, evChan := h.applyOpts(h.cfg.Options(), nil)
+
+	h.handleOldFailureReports(ctx, url, sub != nil)
```
calmh, line 75: This is really a PR of its own hidden in here

calmh (commented):
Good stuff

lib/ur/failurereporting.go
imsodin: https://github.com/syncthing/syncthing/pull/7164

lib/model/model.go
```diff
@@ -253,24 +255,44 @@ func NewModel(cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersio
...
 func (m *model) serve(ctx context.Context) error {
-	// Add and start folders
+	defer m.closeAllConnectionsAndWait()
+
+	if err := m.initFolders(); err != nil {
+		close(m.started)
+		return fatalErr(err)
+	}
+
+	m.cfg.Subscribe(m)
```
calmh, line 265: We could miss a config change after initFolders but before we subscribe, so perhaps subscribe before initFolders?

lib/model/model.go
```diff
@@ -253,24 +255,44 @@ func NewModel(cfg config.Wrapper, id protocol.DeviceID, clientName, clientVersio
...
+	defer m.cfg.Unsubscribe(m)
+	close(m.started)
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case err := <-m.fatalChan:
+		return err
```
calmh, line 273: Seeing as we're never going to return a non-fatal error we could do the wrapping just the once in the return here and have the channel just take normal errors from anywhere

lib/model/model.go
```diff
@@ -2891,3 +2935,14 @@ func writeEncryptionToken(token []byte, cfg config.FolderConfiguration) error {
 		Token:    token,
 	})
 }
+
+func fatalErr(err error) error {
```
calmh, line 2939: This look like a `util.NewFatal(error, status) error` kind of thing

calmh: (I also hate the util package and think we should split it into svcutil and probably configutil or something... for another day.)

lib/model/model.go
imsodin: Probably better that way round, yes. However really there should be a hook or something on `Subscribe` to initialize the subscriber with the config that's in place while subscribing. Otherwise you can still have the race that a config change comes in while we are still initialising folders (which isn't even so unlikely, given creating the filesets and recalculating metadata can take a while).

calmh (commented):
LGTM

calmh (approved):

PR #7151, opened 2020-11-24
lib/model: Simplify access to Folder and Device configuration.

acolomb:
Use the config wrapper's accessor methods with a device / folder argument instead of looking up in a temporary map of the same data.

### Purpose

Slight performance optimization by not re-creating the maps repeatedly.

### Testing

None so far, as it seems an obvious change.


calmh (approved):

acolomb:
Thanks.

PR #7155, opened 2020-11-25
lib/model: Fix enc file size when pulling (fixes #7152)

### LINKED ISSUE #7152: Restart of Syncthing with receive-encrypted folder creates sync-conflicts, adds as unexpected items
1. Share a folder with encryption from A to B
2. On B, in the accept dialog, set the folder to receive-encrypted and disable watching, scan only once a day
3. Let files sync (this required a restart, because the other side has the "doesn't start index sender" bug.)
4. Observe everything is beautiful.
5. Restart Syncthing on B.

Screenshot 2020-11-24 at 23 21 45

Screenshot 2020-11-24 at 23 23 57

It seems every restart adds about one new folder content as sync conflicts and then unexpected items.

(In this, A is v1.12.0-rc.3.dev.22.g5d2c83a7 and B is v1.12.0-rc.3.dev.27.ge6595c1a)

imsodin:
The problem in #7152 is the trailer we add to the file containing the encrypted file info and the connected file size change: We adjust that when actually pulling the file. On index retransfer, we get a new global file, as the version always changes when encrypting a file info. That triggers a pull, but as the contents didn't change, that can be shortcutted. In the shortcut we do not adjust the filesize. So on next scan, the scanner notices the changed file size and thus the file gets marked as locally changed.

Now the file size adjustment happens in the db updater routine of the puller, not in the shared puller state.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -1695,6 +1695,7 @@ func (f *sendReceiveFolder) dbUpdaterRoutine(dbUpdateChan <-chan dbUpdateJob) {
 		return nil
 	}
 
+	recvEnc := f.Type == config.FolderTypeReceiveEncrypted
 loop:
 	for {
 		select {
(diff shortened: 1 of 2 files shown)


calmh:
Excellent detective work

calmh (approved):

PR #7156, opened 2020-11-25
lib/db: Do not reset index-id when dropping device (ref #7135)

imsodin:
This reverts commit 641b7aee384f9490e747abd1c61a4b987018de4a respectively PR #7135.

I created a unit test to make sure that behaviour doesn't erroneously get changed again in the future, and added a comment there why this behaviour is correct/needed:

```
// Dropping the index ID on Drop is bad, because Drop gets called when receiving
// an Index (as opposed to an IndexUpdate), and we don't want to loose the index
// ID when that happens.
```

### DIFF
--- a/lib/db/set.go
+++ b/lib/db/set.go
@@ -74,13 +74,6 @@ func (s *FileSet) Drop(device protocol.DeviceID) {
 		// announced from the remote is newer than our current sequence
 		// number.
 		s.meta.resetAll(device)
-		// Also reset the index ID, as we do want a full index retransfer
-		// if we ever reconnect, regardless of if the index ID changed.
-		if err := s.db.setIndexID(device[:], []byte(s.folder), 0); backend.IsClosed(err) {
-			return
-		} else if err != nil {
-			fatalError(err, opStr, s.db)
-		}
 	}
 
 	t, err := s.db.newReadWriteTransaction()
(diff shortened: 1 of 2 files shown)


calmh (approved):

AudriusButkevicius:
Late to the party and the database model is not one of my strong feats, but I have questions about this.

Seems like we should not call drop in Index (not IndexUpdate) to address this?

As far as I understand Index is just used as a first message for no good reason nowadays I guess, as it's same as IndexUpdate.

The problem we have is that we need to distinguish from "this is a brand new indeX" vs "this is the first message for the continuation of the index you've already seen"?

The reason I feel this is strange is because I feel that this introduces a inconsistencies between what we actually have and what we say we have, if I understand it correctly.

imsodin:
> The problem we have is that we need to distinguish from "this is a brand new indeX" vs "this is the first message for the continuation of the index you've already seen"?

We do that already. If the remote announces an id we know and we don't need to start sending from scratch, we start directly with an IndexUpdate. Dropping all information we already have when getting an Index is correct - we don't want to keep stale index entries if the remote decided to send us everything again.

AudriusButkevicius:
So why is dropping the id a problem, given we're going to re-set it anyway immediately afterwards?
Or are you saying the ordering is somehow screwed up where we set the id, drop the old content if any, accept new content?

imsodin:
Yes, ordering is... well not screwed, just the other way around: We receive and set the index id on ClusterConfig, and ClusterConfig is the first message always, i.e. before Index. So we mustn't reset the index-ID on Index.

PR #7160, opened 2020-11-26
gui: Fix invalid HTML syntax.

acolomb:
While starting work on some GUI code, I noticed some invalid HTML constructs.  Using Emacs' `nxml-mode`, I found some pointers to more syntax errors, across all `.html` files under `gui/default/`.  Unfortunately this can not easily be automated because some of the AngularJS constructs and `translate` attributes are not strictly XML-conformant, but correctly follow the respective package's syntax.

Some of these are arguably cosmetic, as long as no parser trips over them.  Like for the comparison operators leading to stray `<` characters, I propose a rather simple solution which should still provide the highest readability.

TODO
----
* [ ] The last commit is still in FIXME stage, as I don't know how to best solve the issue of forbidden div-elements within headings.
* [ ] This needs to be mirrored to the `untrusted` GUI when finished.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -474,7 +474,7 @@
                               <span class="fas fa-eye-slash"></span>&nbsp;<span translate>Failed to setup, retrying</span>
                             </span>
                           </div>
-                          <div ng-if="0 >= folder.rescanIntervalS">
+                          <div ng-if="folder.rescanIntervalS <= 0">
                             <span ng-if="!folder.fsWatcherEnabled" tooltip data-original-title="{{'Disabled periodic scanning and disabled watching for changes' | translate}}">
                               <span class="far fa-clock"></span>&nbsp;<span translate>Disabled</span>&ensp;
                               <span class="fas fa-eye-slash"></span>&nbsp;<span translate>Disabled</span>
@@ -517,7 +517,7 @@
                       <tr ng-if="folderStats[folder.id].lastScan">
                         <th><span class="far fa-fw fa-clock"></span>&nbsp;<span translate>Last Scan</span></th>
                         <td translate ng-if="folderStats[folder.id].lastScanDays >= 365" class="text-right">Never</td>
-                        <td ng-if="365 > folderStats[folder.id].lastScanDays" class="text-right">
+                        <td ng-if="folderStats[folder.id].lastScanDays < 365" class="text-right">
                           <span>{{folderStats[folder.id].lastScan | date:'yyyy-MM-dd HH:mm:ss'}}</span>
                         </td>
                       </tr>
@@ -551,7 +551,7 @@
                     <button type="button" class="btn btn-default btn-sm" ng-click="restoreVersions.show(folder.id)" ng-if="folder.versioning.type">
                       <span class="fas fa-undo"></span>&nbsp;<span translate>Versions</span>
                     </button>
-                    <button type="button" class="btn btn-sm btn-default" ng-click="rescanFolder(folder.id)" ng-disabled="0 > ['idle', 'stopped', 'unshared', 'outofsync', 'faileditems', 'localadditions'].indexOf(folderStatus(folder))">
+                    <button type="button" class="btn btn-sm btn-default" ng-click="rescanFolder(folder.id)" ng-disabled="['idle', 'stopped', 'unshared', 'outofsync', 'faileditems', 'localadditions'].indexOf(folderStatus(folder)) < 0">
                       <span class="fas fa-refresh"></span>&nbsp;<span translate>Rescan</span>
                     </button>
                     <button type="button" class="btn btn-sm btn-default" ng-click="editFolder(folder)">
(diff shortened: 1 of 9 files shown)


tomasz1986:
Angular is invalid HTML per se, but yeah, there is plenty of invalid code in the GUI. Just check each HTML file in the W3C Validator (https://validator.w3.org).

However, there is no need to change `<br>` to `<br />`>. The former is totally valid, unless you are doing XHTML.

acolomb:
Staying close to XHTML has some value at least when using the a generic XML checker tool like I did. Especially since the usual HTML5 tools will report even more false positives as you confirmed.

tomasz1986:
> Staying close to XHTML has some value at least when using a generic XML checker tool like I did. Especially since the usual HTML5 tools will report even more false positives as you confirmed.

I would like to disagree with this point. HTML5 has diverged too much to be able to use any standard XML validator on it. Unless the code has been XHTML from the very beginning, I personally do not see any benefits. The code, as it is right now, is completely invalid as far as XHTML goes anyway. Even with XHTML itself, I would be cautious with applying XML tools to it (see the `<hr`> comment). And I am saying this as a fan of both XML and XHTML.

The problems reported by the HTML5 tools are not false positives though. The code does in fact have a lot of errors. The problem is that I do not think that it is possible to really do anything meaningful with them as long as Angular is used to generate the content.

acolomb:
You're right, the `<hr>` thing was a late night mistake... I see that now.

Regarding XHTML style, I would prefer consistency and the `<hr />` style would actually have helped me to see the intention while navigating the code.

But I separated the commits for different "error types", so no problem to revert individually.

gui/default/index.html
```diff
@@ -474,7 +474,7 @@ <h4 class="panel-title">
                               <span class="fas fa-eye-slash"></span>&nbsp;<span translate>Failed to setup, retrying</span>
                             </span>
                           </div>
-                          <div ng-if="folder.rescanIntervalS <= 0">
+                          <div ng-if="0 >= folder.rescanIntervalS">
```
wweich, line 477: I don't think the syncthing project prefers yoda conditions

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -22,7 +22,7 @@
                 </ul>
               </p>
               <p class="help-block">
-                <span translate ng-if="deviceEditor.deviceID.$valid || deviceEditor.deviceID.$pristine">The device ID to enter here can be found in the "Actions > Show ID" dialog on the other device. Spaces and dashes are optional (ignored).</span>
+                <span translate ng-if="deviceEditor.deviceID.$valid || deviceEditor.deviceID.$pristine">The device ID to enter here can be found in the "Actions &gt; Show ID" dialog on the other device. Spaces and dashes are optional (ignored).</span>
```
wweich, line 25: this will break the current translations as the original string changes

calmh: We could possibly handle this in the tooling, translating back and forth as needed... (so that strings towards transifex get html entities resolved and removed, and re-added again when importing the translated strings.)

gui/default/index.html
calmh: I fucking hate yoda conditions. I also think it's probably a dead horse, because as long as we use angular with these types of expressions all over the place it's unlikely all future contributors will understand to not use less-than. Worst case they will see some examples like this and think we want yoda everywhere... Iff we wanted to fix this (which I'm not sure it's worth) I guess one way would be to move the expressions into the controller instead...

acolomb: Okay, let's revert this part then.  The other approach would be to wrap it like `!(x >= x)`, but that's even uglier.

gui/default/syncthing/device/editDeviceModalView.html
calmh: The translation script is an ugly mess, but I think this should pretty much cover it:

```diff
--- a/script/translate.go
+++ b/script/translate.go
@@ -94,7 +94,7 @@ func inTranslate(n *html.Node, filename string) {
 }

 func translation(v string) {
-       v = strings.TrimSpace(v)
+       v = strings.TrimSpace(html.UnescapeString(v))
        if _, ok := trans[v]; !ok {
                av := strings.Replace(v, "{%", "{{", -1)
                av = strings.Replace(av, "%}", "}}", -1)
```

any HTML entities in the original source get parsed on the way to the translation JSON. I don't think we need to do anything in particular in the other direction (at least we haven't so far).

acolomb: Should I apply that change in this PR then?

calmh: Yeah, try that, and run `go run build.go translate` on the changes afterwards (ideally it should not change the lang-en.json file).

calmh: Actually, thinking about it one more time, it could be that the angular translate plugin fails to pick it up and we need to leave the html entities in the "key" string. So I guess see if translation still works, and if it doesn't it will need some adjustment.

acolomb: After `go run build.go translate`, my local copy of `lang-en.json` is empty.  With or without that change.

acolomb: I don't know why the `build.go` script fails, apparently the redirect of stdout does not work.  I always get an empty `lang-en-new.json` file.  But running `translate.go` does output correctly to stdout.

No change in content for `lang-en.json` with that patch applied.

st-review:
:robot: *beep boop*

PR #7161, opened 2020-11-27
gui: Restore Select / Deselect All buttons in device sharing tab.

acolomb:
### Purpose

Refactoring in #7049 broke the "Select All" and "Deselect All" buttons under the Sharing tab for editing devices.  It was supposed to work just like for the folder editing modal, but the distinction between already shared and not shared is not made in the edit device modal yet.

Fixing that properly (by splitting just like it is done in the edit folder modal) will come in a separate PR, but I wanted to get this one out as a quick regression fix.

### Testing

Both buttons confirmed working when editing an existing device.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1438,6 +1438,18 @@ angular.module('syncthing.core')
             $('#editDevice').modal();
         };
 
+        $scope.selectAllFolders = function () {
+            angular.forEach($scope.folders, function (_, id) {
+                $scope.currentSharing.selected[id] = true;
+            });
+        };
+
+        $scope.deSelectAllFolders = function () {
+            angular.forEach($scope.folders, function (_, id) {
+                $scope.currentSharing.selected[id] = false;
+            });
+        };
+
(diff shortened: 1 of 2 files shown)


acolomb:
Note that this cannot be applied to the `untrusted` GUI variant as-is.  The splitting into two categories seems to have happened there already.  I have a PR almost ready to properly fix this in the default GUI, so that one should rather be replicated onto `untrusted`.

calmh (approved):

acolomb:
This should probably go to the release branch as well, as it is fixes a regression in v1.12.

acolomb:
@calmh Would you consider giving me write access to the syncthing main repo?  That way I could add labels, milestones, link issues etc.  I promise not to bypass the review process ;-)

Should have sent this PR against the `release` branch in the first place, sorry.

calmh:
I've sent an invitation. However, please only ever file PRs against `main`, and ideally leave the milestones as is as I set and verify them as part of the release process. :)

acolomb:
Thanks @calmh.  So you will merge the commit over to release?

calmh:
Yep

PR #7162, opened 2020-11-27
gui: Split folders into two categories on the sharing tab for devices.

acolomb:
This is the proper fix for the problem mentioned in #7161.  While the former is meant as a hot-fix, this one actually changes the GUI layout and functionality.

When editing folders, the checkboxes have been split into two categories for some time now.  This one mirrors the same to adding / editing a device.  Also properly fixes copy-paste errors introduced with #7049.

This PR does not include replicating the changes to the `untrusted` GUI, as it is nontrivial in this case.

### DIFF
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -66,7 +66,7 @@
           <div class="row">
             <div class="col-md-12">
               <div class="form-group" ng-if="currentSharing.shared.length">
-                <label translate for="folders">Currently Shared Folders for Device</label>
+                <label translate for="folders">Shared Folders</label>
                 <p class="help-block">
                   <span translate>Deselect folders to stop sharing with this device.</span>&emsp;
                   <small><a href="#" ng-click="selectAllSharedFolders(true)" translate>Select All</a>&emsp;
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -68,7 +68,7 @@
               <div class="form-group" ng-if="currentSharing.shared.length">
                 <label translate for="folders">Currently Shared Folders for Device</label>
                 <p class="help-block">
-		  <span translate>Deselect folders to stop sharing with this device.</span>&emsp;
+                  <span translate>Deselect folders to stop sharing with this device.</span>&emsp;
                   <small><a href="#" ng-click="selectAllSharedFolders(true)" translate>Select All</a>&emsp;
                     <a href="#" ng-click="selectAllSharedFolders(false)" translate>Deselect All</a></small>
                 </p>
@@ -92,9 +92,9 @@
                   <small><a href="#" ng-click="selectAllUnrelatedFolders(true)" translate>Select All</a>&emsp;
                     <a href="#" ng-click="selectAllUnrelatedFolders(false)" translate>Deselect All</a></small>
                 </p>
-		<p class="help-block" ng-if="folderList().length == 0">
-		  <span translate>There are no folders to share with this device.</span>
-		</p>
+                <p class="help-block" ng-if="folderList().length == 0">
+                  <span translate>There are no folders to share with this device.</span>
+                </p>
                 <div class="row">
                   <div class="col-md-4" ng-repeat="folder in currentSharing.unrelated">
                     <div class="checkbox">
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -65,15 +65,38 @@
           </div>
           <div class="row">
             <div class="col-md-12">
-              <div class="form-group">
-                <label translate for="folders">Share Folders With Device</label>
+              <div class="form-group" ng-if="currentSharing.shared.length">
+                <label translate for="folders">Currently Shared Folders for Device</label>
                 <p class="help-block">
-                  <span translate>Select the folders to share with this device.</span>&emsp;
-                  <small><a href="#" ng-click="selectAllFolders()" translate>Select All</a>&emsp;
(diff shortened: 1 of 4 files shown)


calmh:
Sounds good, needs minor conflict resolving now. I don't think we need to fine tune this now for the untrusted case, it's good enough to fix it in the old GUI.

imsodin:
Either my short-term memory is failing me bad, or that's already how it looks like in the untrusted branch. There I deduplicated this part of the folder and device edit modals in `editShareTemplate.html` respectively the `share-template` directive.

acolomb:
> Either my short-term memory is failing me bad, or that's already how it looks like in the untrusted branch. There I deduplicated this part of the folder and device edit modals in `editShareTemplate.html` respectively the `share-template` directive.

Yes, sort of.  That's what I noticed after trying to merge my changes into untrusted.  Best way forward I think is to merge this, then adjust the untrusted copy to have the smallest needed diff.

imsodin:
Ah I thought this was about the separation in already shared and unrelated devices, but it's about the (de-)select all buttons. Which indeed don't work for device in untrusted. Thanks for catching all of this.

acolomb:
It is indeed about the separation.  You merged a (nonfunctional) version of the JS part, but didn't adjust the HTML (outside untrusted) accordingly.  Which was fixed by #7161 as an immediate hot-fix.

imsodin:
Again, thanks for that. I am talking about the untrusted bit now:

> This PR does not include replicating the changes to the untrusted GUI, as it is nontrivial in this case.

The problem is the same copy-paste error you just fixed in the "classic" JS. All that's needed is:

```diff
--- a/gui/default/untrusted/syncthing/core/syncthingController.js
+++ b/gui/default/untrusted/syncthing/core/syncthingController.js
@@ -1457,16 +1457,16 @@ angular.module('syncthing.core')
         };
 
         $scope.selectAllSharedFolders = function (state) {
-            var devices = $scope.currentSharing.shared;
-            for (var i = 0; i < devices.length; i++) {
-                $scope.currentSharing.selected[devices[i].deviceID] = !!state;
+            var folders = $scope.currentSharing.shared;
+            for (var i = 0; i < folders.length; i++) {
+                $scope.currentSharing.selected[folders[i].id] = !!state;
             }
         };
 
         $scope.selectAllUnrelatedFolders = function (state) {
-            var devices = $scope.currentSharing.unrelated;
-            for (var i = 0; i < devices.length; i++) {
-                $scope.currentSharing.selected[devices[i].deviceID] = !!state;
+            var folders = $scope.currentSharing.unrelated;
+            for (var i = 0; i < folders.length; i++) {
+                $scope.currentSharing.selected[folders[i].id] = !!state;
             }
         };
```

acolomb:
Right, but on the HTML side you added something related to encrypted devices, which I didn't examine in detail yet.  I'll  I dive into it and update this PR to cover the untrusted GUI as well.

acolomb:
@imsodin merging the HTML changes is complex, but making progress.  I see you added a new template for that.  Big question: How can I test the untrusted GUI at all?

imsodin:
I still don't get why you believe you need to merge anything. Just the diff above makes it fully functional for me.

Anyways go to actions > advanced > options > featureflags and enter untrusted. Then reload the gui and you have the untrusted gui.

acolomb:
> I still don't get why you believe you need to merge anything. Just the diff above makes it fully functional for me.

Not fully, as another change was needed in `$scope.addDevice()` to make it work.  Besides that, I tried to port any **changes** done on the main GUI to the untrusted variant.  I thought the goal was to keep the diff between those two only encryption-related.

> Anyways go to actions > advanced > options > featureflags and enter untrusted. Then reload the gui and you have the untrusted gui.

Thanks, that works.

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -65,15 +65,38 @@
           </div>
           <div class="row">
             <div class="col-md-12">
-              <div class="form-group">
-                <label translate for="folders">Share Folders With Device</label>
+              <div class="form-group" ng-if="currentSharing.shared.length">
+                <label translate for="folders">Currently Shared Folders for Device</label>
```
imsodin, line 69: I'd prefer

```suggestion
                <label translate for="folders">Shared Folders</label>
```

as in the untrusted. There's no need to mention the context of a device, when in the device edit modal.

gui/default/untrusted/syncthing/core/syncthingController.js
```diff
@@ -1433,40 +1433,36 @@ angular.module('syncthing.core')
...
-                } else {
-                    $scope.currentSharing.unrelated.push($scope.folders[folderID]);
-                }
-                $scope.currentSharing.selected[folderID] = found;
-            }
-
+            });
+            $scope.currentSharing.unrelated = $scope.folderList().filter(function (n) {
+                return !$scope.currentSharing.selected[n.id];
+            });
```
imsodin, line 1450: As far as I understand this does not change behaviour. What's the reasoning for this change?

gui/default/syncthing/device/editDeviceModalView.html
acolomb: Right, that's probably too much.  I was trying to align it with the edit folder modal, which has a very similar heading.  Will change.

gui/default/untrusted/syncthing/core/syncthingController.js
acolomb: As I said, just trying to keep the diff between default and untrusted as small as possible, without having two different implementations of the same algorithm.  I picked the one from default because it was easier to grasp for me at least.

acolomb:
Is that a random test failure?  I can't make out a reason from the build log.

imsodin (approved):
Looks good.

acolomb:
Anything else for this to be merged?

PR #7164, opened 2020-11-27
lib/ur: Send unreported failures on shutdown

imsodin:
If a fatal error should be reported, chances are good we are getting shut down before they can be reported. To mitigate that unsent reports are written to disk when the service stops and these failures are then reported on next startup.


calmh:
Is there reason to believe we have time to write them to disk? (And if there is, why don't we have time to send them upstream instead?)

imsodin:
Suture gives a service 10s to terminate, which is hopefully enough to write to disk. However shutdown should be as fast as possible, and given the pretty long timeouts we have on all these network sending tasks, I classified them as too slow to happen on shutdown. In normal circumstances that's probably not the case, and we could try to send for e.g. 100ms and only write to disk if that fails.

lib/ur/failurereporting.go
```diff
@@ -160,6 +167,56 @@ func (h *failureHandler) addReport(descr string, evTime time.Time) {
...
+		return
+	}
+	reports, err := readFailures(path)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			l.Infoln("Failed to read previous failures:", err)
+		}
+		return
+	}
+	if err := sendFailureReports(ctx, reports, url); err != nil {
```
calmh, line 185: delete the file after it's sent, or it will get sent repeatedly?

calmh:
I'm kinda sorta feeling this is too much machinery for quite little gain, but i'm not going to actively fight it.

AudriusButkevicius:
Has the failure stuff proven to be useful thus far?

imsodin:
It just feels wrong to not report most fatal errors by design.

However I agree storing stuff on disk is too much machinery for the gain. To still have some chance of getting reports sent on shutdown, it now tries to send reports on shutdown, with a timeout of half the service timeout (which is now accessible as `svcutil.ServiceTimeout`).

And for reasons I acted on `util` containing too much different stuff and move bits relevant to services to svcutil. While a trivial change, it blows up the diff drowning the relevant bits. Tell me if you think it would be better in a separate PR.

calmh (dismissed):

calmh:
Is fine, requires reviving... :/

calmh (dismissed):

calmh (approved):

PR #7167, opened 2020-11-27
lib/model: Unflake TestRequestReceiveEncryptedLocalNoSend

imsodin:
That new test is super flaky: https://build.syncthing.net/project.html?projectId=Syncthing&testNameId=-4005993063680092438&tab=testDetails  
Couldn't repro locally, but adding the callback to the connection after passing the connection to the model seems like a very probably culprit.


calmh (approved):
yolo

Issue #7146, opened 2020-11-24
QUIC writes directly to stdout


imsodin:
Running on main I get lines like this in the Syncthing log:

```
2020/11/24 10:52:01 Failed to sufficiently increase receive buffer size. Was: 208 kiB, wanted: 2048 kiB, got: 416 kiB.
```

Grepping my `$GOPATH/pkg/mod` puts the blame on quic. I haven't looked into it at any further at all, thus keeping `needs-trage` for now. Likely introduced in https://github.com/syncthing/syncthing/pull/7139, thus `skip-changelog`.

imsodin:
https://github.com/lucas-clemente/quic-go/issues/2929

RubenKelevra:
Well, this doesn't really fix the underlying issue. Better to include the real fix:

https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size

imsodin:
I don't know what "this" you are referring to, but this issue is about QUIC logging to the standard logger, which ends up directly in stdout instead of in our logging system.

The particular line mentioned as an example is about udp buffer size indeed. The new version of QUIC even prints a link to the wiki page you posted in the error message. However that's not something Syncthing can control, as adjusting it requires root privileges.

bt90:
> However that's not something Syncthing can control, as adjusting it requires root privileges.

Syncthing *could* package a sysctl config file to raise the buffer size. But that would only make sense if we get hit by this.

bt90:
@imsodin https://github.com/bt90/syncthing/tree/sysctl_udp

PR #7168, opened 2020-11-28
cmd/syncthing: Discard default logger output (ref #7146)

imsodin:
The logs on my nightly devices get filled with this quic stuff that they write directly to the default logger of the `log` package (#7146). I hope they will reconsider that practice, but until then lets discard it to not drown relevant logging. Alternatively we could redirect it to a new debug logging facility (e.g. `logger` or `external`) but that seems pointless as noone will ever remember it exists or notice when debugging, given how generic it would be.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -47,6 +47,13 @@ import (
 	"github.com/pkg/errors"
 )
 
+func init() {
+	// Temporary (hopefully) workaround due to QUIC writing unconditionally
+	// to the default logger.
+	// https://github.com/syncthing/syncthing/issues/7146
+	log.SetOutput(ioutil.Discard)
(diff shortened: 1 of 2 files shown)


imsodin:
Actually we can't do this in the logger package, as some packages in `cmd/...` use the `log` package and most likely import `logger`, thus their logging would be silenced too -> moved `init()` to `cmd/syncthing`.

calmh:
I'm sort of thinking that maybe the cure is worse than the decease here. Some spurious logging from quic is annoying, but will hopefully get fixed? With this in place we'll never know about similar things in the future, or panics in the http server...

imsodin:
That just happened (partially): https://github.com/syncthing/syncthing/pull/7180

PR #7177, opened 2020-12-02
lib/connections: Refactor connection loop

calmh:
This breaks out some methods from the connection loop to make it simpler to manage and understand.

Some slight simplifications to remove the `seen` variable (we can filter `nextDial` based on whether times are in the future or not, so we don't need to track `seen`) and adding a minimum loop interval (5s) in case some dialer goes haywire and requests a 0s redial interval or such.

Otherwise no behavioral changes.


lib/connections/service.go
```diff
@@ -336,176 +338,200 @@ func (s *service) handle(ctx context.Context) error {
...
-			if connected && ct.Priority() == bestDialerPrio {
-				// Things are already as good as they can get.
-				continue
-			}
+		// The actual sleep time will have been the time until the next dial
+		// scheduled in nextDial, clamped by stdConnectionLoopSleep as we
+		// don't want to sleep too long (config changes might happen).
+		if sleep < minConnectionLoopSleep {
+			sleep = minConnectionLoopSleep
+		}
```
AudriusButkevicius, line 383: This talks about too long but sets the sleep to a bigger value?

lib/connections/service.go
```diff
@@ -336,176 +338,200 @@ func (s *service) handle(ctx context.Context) error {
...
-				uri, err := url.Parse(addr)
-				if err != nil {
-					s.setConnectionStatus(addr, err)
-					l.Infof("Parsing dialer address %s: %v", addr, err)
-					continue
-				}
+		// Don't connect to devices that are already connected using the
+		// best possible dialer priority...
+		connection, connected := s.model.Connection(deviceCfg.DeviceID)
+		if connected && connection.Priority() == bestDialerPrio {
```
AudriusButkevicius, line 425: This turned from >= to ==

AudriusButkevicius (approved):
5 beers in, this looks good.

AudriusButkevicius (dismissed):

lib/connections/service.go
```diff
@@ -336,176 +338,203 @@ func (s *service) handle(ctx context.Context) error {
...
+		s.dialDevices(ctx, now, cfg, bestDialerPrio, nextDial, isInitialRampup)
 
-			deviceID := deviceCfg.DeviceID
-			if deviceID == s.myID {
-				continue
-			}
+		// The sleep time is until the next dial scheduled in nextDial,
+		// clamped by stdConnectionLoopSleep as we don't want to sleep too
+		// long (config changes might happen).
+		sleep := filterAndFindSleepDuration(nextDial, now)
```
imsodin, line 373: Maybe combine with block below to not do this when `!isInitialRampup`

lib/connections/service.go
```diff
@@ -336,176 +338,203 @@ func (s *service) handle(ctx context.Context) error {
...
 
-			if deviceCfg.Paused {
-				continue
-			}
+		// .. unless we are un the initial rampup time, in which case we
```
imsodin, line 375:
```suggestion
		// .. unless we are in the initial rampup time, in which case we
```

lib/connections/service.go
```diff
@@ -336,176 +338,203 @@ func (s *service) handle(ctx context.Context) error {
...
+		// override the sleep interval for the first few passes through the
+		// loop.
+		if isInitialRampup {
+			sleep = initialRampup
+			initialRampup *= 2
+		}
 
-			ct, connected := s.model.Connection(deviceID)
+		// ... while making sure not to loop to quickly either.
```
imsodin, line 383:
```suggestion
		// ... while making sure not to loop too quickly either.
```

lib/connections/service.go
```diff
@@ -336,176 +338,203 @@ func (s *service) handle(ctx context.Context) error {
...
+	for _, df := range dialers {
+		if df.Valid(cfg) != nil {
+			continue
+		}
+		if prio := df.Priority(); first || prio < bestDialerPrio {
+			first = false
+			bestDialerPrio = prio
+		}
+	}
+	return bestDialerPrio
```
imsodin, line 410: This now returns zero for no dialer vs highest possible int32 before. It won't change any behaviour though, as the only condition using `bestDialerPrio` is a `==`, which will never match either 0 or highest int32. Still no dialer resulting in highest prio is a bit surprising.

lib/connections/service.go
```diff
@@ -336,176 +338,203 @@ func (s *service) handle(ctx context.Context) error {
...
-					l.Debugf("Not dialing %s via %v as sleep is %v, next dial is at %s and current time is %s", deviceID, addr, sleep, nextDialAt, now)
-					continue
-				}
-				// If we fail at any step before actually getting the dialer
-				// retry in a minute
-				nextDial[nextDialKey] = now.Add(time.Minute)
+		// Don't connect to devices that are already connected using the
+		// best possible dialer priority...
+		connection, connected := s.model.Connection(deviceCfg.DeviceID)
+		if connected && connection.Priority() == bestDialerPrio {
```
imsodin, line 428: Unrelated to this PR: This should be `>=`. If we e.g. listen on tcp but disabled dialing on tcp, we don't need to dial lower-prio methods when a tcp connection is already established.

Just saw that later on in the dialer loop we do use the `>=` condition between dialer prio and current connection prio.

lib/connections/service.go
```diff
@@ -336,176 +338,203 @@ func (s *service) handle(ctx context.Context) error {
...
-				if len(deviceCfg.AllowedNetworks) > 0 {
-					if !IsAllowedNetwork(uri.Host, deviceCfg.AllowedNetworks) {
-						s.setConnectionStatus(addr, errors.New("network disallowed"))
-						l.Debugln("Network for", uri, "is disallowed")
-						continue
-					}
-				}
+func (s *service) resolveDialTargets(ctx context.Context, now time.Time, cfg config.Configuration, deviceCfg config.DeviceConfiguration, nextDial map[string]time.Time, initial bool) []dialTarget {
+	deviceID := deviceCfg.DeviceID
+	connection, connected := s.model.Connection(deviceID)
```
imsodin, line 441: Conn is already obtained just before calling this function and could thus be passed in.

calmh:
PTAL

imsodin (approved):

PR #7185, opened 2020-12-07
gui: Fix blank device name under recent changes.

elesiuta:
### Purpose

Fixes the device showing up as blank under recent changes if the friendly name is known. This also appears to have been mentioned in the [forum](https://forum.syncthing.net/t/device-name-is-blank-in-recent-changes/15970).

If the friendly name is unknown, eg if the remote device is removed after making a change, the short ID is still shown properly both before and after this change.


calmh (approved):

calmh:
Thanks!

PR #7186, opened 2020-12-08
lib/connections: Handle QUIC not being available

calmh:
This does two things:

- Exclude QUIC from go1.16 builds, automatically, for now, since it
  doesn't work and just panics.

- Provide some fake listeners and dialers when QUIC is disabled.

These fake listeners and dialers indicate that they are disabled and
unsupported, which silences "Dialing $address: unknown address scheme:
quic" type of stuff which is not super helpful to the user.

### DIFF
--- a/lib/connections/quic_dial.go
+++ b/lib/connections/quic_dial.go
@@ -4,7 +4,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
-// +build go1.12,!noquic,!go1.16
+// +build go1.14,!noquic,!go1.16
 
 package connections
 
--- a/lib/connections/quic_dial.go
+++ b/lib/connections/quic_dial.go
@@ -4,7 +4,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
-// +build go1.12,!noquic
+// +build go1.12,!noquic,!go1.16
 
 package connections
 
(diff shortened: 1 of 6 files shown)


lib/connections/quic_dial.go
```diff
@@ -4,7 +4,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
-// +build go1.12,!noquic
+// +build go1.12,!noquic,!go1.16
```
imsodin, line 7: While at it: Quic/qtls apparently removed support for go1.13 a while ago, so the actual lower bound is now go1.14:

https://github.com/marten-seemann/qtls/commit/6befe96d4cf783737540378ecef36e171cdca49c

AudriusButkevicius:
I'd move the unsupported things into it's own generic file, and then just leave the inits in the quic specific file, in case we want to reuse them.

calmh:
Indeed, and we already had a mostly suitable type no less. Did some minor cleanup and refactoring on the error comparisons.

AudriusButkevicius (approved):

lib/connections/quic_unsupported.go
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2020 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// +build noquic go1.16
```
imsodin, line 7:
```suggestion
// +build noquic !go1.14 go1.16
```

imsodin (approved):

PR #7188, opened 2020-12-10
lib: Apply config changes sequentially (ref #5298)

imsodin:
This is a proposal to make config changes in the current system less prone to races. It is thus related to #5298, however clearly does not address all the issues brought up there nor is it intending to.

It was prompted by the review comment about the order of initialising folders in the model and subscribing to config changes here: https://github.com/syncthing/syncthing/pull/7150#discussion_r537010460  
That's inherently racy as it is, regardless of the order in which order it's done.

The proposed change here is that changes to the config only go ahead once all calls to `.CommitConfiguration` due to the previous change returned. If multiple config changes are outstanding, they will applied in the same order as the config change method (e.g. `SetFolder`) was called.  
In addition with `.Subscribe` returning the currently active config, this allows a subscriber to subscribe to config changes, use the returned config for initialisation, and then start processing config changes. Thus starting with subscribing, they get all config changes sequentially in the correct order. Which I have done for `model`.

The obvious potential for problems here is locking. Previously anyone changing the config was only blocked until all previous config changes were dispatched asynchronously. Now blocking happens until all previous config changes took effect. If a subscriber is changing the config while holding a lock, that the subscriber also acquires on `CommitConfiguration`, that can deadlock. However doing that is also problematic today (can cause concurrent `CommitConfiguration` calls, which is also very bad) and we are aware of it, as e.g. evident in this comment:

```golang
	// Needs to happen outside of the fmut, as can cause CommitConfiguration
```

I added an explicit clause to the comment on the `Committer` interface, which obviously doesn't guarantee anything, but might increase awareness.


lib/config/wrapper.go
```diff
@@ -183,11 +196,29 @@ func (w *wrapper) RawCopy() Configuration {
...
+		w.currentCounter = 0
+	}
+	myCounter := w.replaceCounter
+	w.replaceCounter++
+
+	for myCounter > w.currentCounter {
+		w.replaceCond.Wait()
+	}
+
+	w.replaceCond.L.Unlock()
```
calmh, line 219: This whole thing is non-obvious in what it accomplishes and could use a couple of lines of comments about what & why ("ensures config changes are applied in order because ..."). Also I'd use int64 and not require any logic about it wrapper (borderline non-required even in the int32 case). Oh, and assuming I understood the sequencing purpose correctly, the `initChange` name is kinda nondescript. `awaitQueuedOrder` or something?

lib/config/wrapper.go
```diff
@@ -203,9 +234,17 @@ func (w *wrapper) replaceLocked(to Configuration) (Waiter, error) {
 		}
 	}
 
-	w.cfg = to
+	waiter := w.notifyListeners(from.Copy(), to.Copy())
+	go func() {
```
calmh, line 238: If we return due to an error in any of the conditions above we'll deadlock forever because we increased replaceCounter but currentCounter can never keep up?

(I wonder if this counter-and-cond thing is the best way to represent a config  change queue... I'm thinking a service loop reading from a `chan func(*Config) *Config` that is processed serially, and the config change methods just push their operation to that channel? Make the channel 1000-buffered or something, and if sending to the channel fails it means the config change operation returned an error...)

calmh (commented):
Hmm

imsodin:
> Hmm

Actually I wanted to talk about that (the concept) mainly. I just started to explain what I had in mind without code, and thought that wouldn't be comprehensible. Thus I created an explanation plus code to compare if the explanation isn't fully comprehensible.  
I guess the review (thanks) means that at the least you didn't immediately think of a catch that invalidates the concept :)

calmh:
I think the concept is sound. In general I think 1) starting a subscription, 2) getting the config, 3) continuing with life, should be a working sequence of operations for a subscriber even without this, as long as you're guaranteed to get notifications of all config changes that happen after the subscribe call returns. However, sequencing all config changes so that they are actually applied in the order they were asked for still seems worthwhile.

imsodin:
I refactored (rewrote) as suggested with a buffered chan. There's still a bit of machinery as I didn't want to just spawn a goroutine on `Wrap` that's never terminated (neither do I want config to be a service). The machinery seemed sound and understandable while writing, I hope it stands the test of review (and/or a nights sleep and looking at it again) :)

AudriusButkevicius (commented):
So I understand what this tries to do, but I doubt it squashes a lot of bugs.

I think our bigger problem is things like
```
cfg = wcfg.Copy()
cfg.Fiddle = true
wcfg.Save(...)
```
which would not be solved by this, because someone else might have done SetFolders between Copy() and Save() which would now be lost.

The same problem applies at a smaller level too, whereby
```
devs := wcfg.Devices()
devs.forEach { it.Fiddle = true }
wcfg.SetDevices(devs)
```

Suddenly loses any devices added between Devices() and SetDevices().

The few ways I can think of solving this is turning config modification around.
All the types you read now like Devices() etc are immutable (probably return interfaces),
and the only way to mutate stuff is to do something like:
```
wcfg.Modify(func (cfg config.Configuration) {
    cfg.Fiddle = true
})
```
or 
```
leImmutableDevice.Modify(func (mutable)  {
   mutable.Fiddle = true
})
```
which is obviously serialized and potentially blocking, and probably PITA to use.

The other option would be remove ability to modify collections in bulk.

So no SetDevices, just AddDevice/RemoveDevice etc, then each struct gains a version, and on each read/modify/write we check if the version being written for this device matches the version currently in the config to catch outdated data writes, and prevent those.
Sprinkle retry logging everywhere.

Or idk, switch to some sort of CRDT library, plus counters?

imsodin:
> So I understand what this tries to do, but I doubt it squashes a lot of bugs.

I agree. What you explained is a much bigger source of bugs. However solving it is also hard, while this is simple, and can happen now. I do believe the issue solved here might realistically cause bugs at startup, as it's quite realistic that a user changes stuff in the UI while folders are still rebuilding metadata. That immeadiately creates a race between handling folders in `serve` and `CommitConfiguration`.

> Suddenly loses any devices added between Devices() and SetDevices().

Not that it changes much, but `SetDevices` is more like `AddOrReplaceDevices`: It doesn't remove devices not in the given device-list.

calmh:
I see the thing with avoiding a serve() method, but it's a new design that we don't have anywhere else that's less easy to see is obviously correct. I think having and actual serve loop and combining that with Audrius' API suggestion would be better. That is, something like

```
type Wrapper struct {
  queue chan func(*Configuration)
  ...
} // there can be only one Wrapper instance

func (w* Wrapper) Serve() {
  for fn := range funcs { // actually a select probably
    fn(&w.cfg) // mutates
    // notify subscribers
    // save
  }
}
```

We can then be sure that things are serialized because there is only the one loop, and it can do things like set a timer to do a save one second after a config change, so multiple changes get aggregated, etc...

We could still have methods like RemoveDevice and SetDevice. If SetDevice takes a device config it's not 100% safe against multiple mutators touching the same device at the same time, but really that's a design error regardless. If we really care then the Set... functions could of course take mutator functions like `func(*DeviceConfiguration)` and such to do the actual tweaking of only specific attributes...

imsodin:
Now `config.wrapper` is a service and modification happens through

```
type ModifyFunction func(*Configuration) bool

	Modify(ModifyFunction) (Waiter, error)
```

The end state is quite palatable, the diff... less so. Though a lot of it is boilerplate in tests.

acolomb:
Sorry if this is not directly related, but I read about possibly aggregating config changes before calling `CommitConfiguration()` above.  I found [one place](https://github.com/imsodin/syncthing/blob/d3a403343dc6d9f2960454d801383bef806c73d0/lib/model/model.go#L2713) where that might more easily trigger an existing bug: 

```go
    		// Ignored folder was removed, reconnect to retrigger the prompt.
		if !toCfg.Paused && len(fromCfg.IgnoredFolders) > len(toCfg.IgnoredFolders) {
			closeDevices = append(closeDevices, deviceID)
		}
```

If a folder is unignored and another one ignored in the same modification cycle, this check will not notice and neglect to replace the connection.  I don't see an easy way except a deep comparison to catch this.

imsodin:
The new config wrapper service saves the config every 5 seconds if there are any changes now. Thus I removed all the `.Save` calls that did not act on the error except for logging. Explicitly calling `.Save` is still possible and used in the rest api, where an error is returned if saving fails.

The PR is ready from my side.

lib/config/wrapper.go
```diff
@@ -179,11 +206,77 @@ func (w *wrapper) RawCopy() Configuration {
...
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+
+		var waiter Waiter = noopWaiter{}
+		var err error
+
+		w.mut.Lock()
+		to := w.cfg.Copy()
+		if e.modifyFunc(&to) {
```
calmh, line 250: As an alternative to returning a boolean we could serialize the config before and after the function call and compare to see if they changed something relevant or not. Then perhaps the modifier could/should return an error, for cases where it starts modifying the config but then realizes something went wrong and needs to bail out... but if that is not a thing that can happen currently we could leave the modifier without return value?

lib/config/wrapper.go
```diff
@@ -179,11 +206,77 @@ func (w *wrapper) RawCopy() Configuration {
...
+		select {
+		case <-done:
+			if !saveTimerRunning {
+				saveTimer.Reset(minSaveInterval)
+				saveTimerRunning = true
+			}
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
```
calmh, line 273: Seems to me we could start the save timer directly when the modify returns and save some small complexity?

lib/syncthing/syncthing.go
```diff
@@ -277,24 +281,26 @@ func (a *App) startup() error {
 	a.mainService.Add(discoveryManager)
 	a.mainService.Add(connectionsService)
 
-	// Candidate builds always run with usage reporting.
+	a.cfg.Modify(func(cfg *config.Configuration) bool {
```
calmh, line 284: Are we sure mainService is running at this point? If not, this will hang, right?

calmh (commented):
I like this

lib/config/wrapper.go
imsodin: It seems a bit ugly to me to serialize or reflect to see if the config changed, when the caller knows whether they changed the config or not. Not that it matters performance wise. Well that's my reasoning, I am not attached to either solution though :1234:

lib/config/wrapper.go
imsodin: Waiting for `done` happens to serialize modifications, i.e. only process a change once the previous one is fully in effect. Using that to restart the timer only afterward is just a side-effect. However regardless of complexity, starting it immediatly on the change feels like the better call: I don't see any reason to e.g. wait for a slow `CommitConfiguration` before saving.

lib/syncthing/syncthing.go
imsodin: Yes and yes. Starting the api in this same function would already hang: It's waiting for the server to come up.

lib/config/wrapper.go
calmh: I was thinking it a bit ugly that I had to say whether I modified the config or not, when the config is there so the caller can plainly see if I did or not. :) It also wasn't 100% certain what the bool return meant at first glance.

Since it's just for saving we could even just remember the last serialized version and when going to save check if it's the same and then skip it.

But if there are cases where we start changing the config and then return false because we change our mind then this is legitimate. (I didn't check if we ever do that, some of the auto accept and introducer stuff is convoluted.)

(I'm also not heavily invested in this, it's fine as is, really.)

imsodin: There's a bit of ugliness everywhere :)

> Since it's just for saving we could even just remember the last serialized version and when going to save check if it's the same and then skip it.

It's also to skip the subscriber verify and commit stages if unchanged.

> But if there are cases where we start changing the config and then return false because we change our mind then this is legitimate. (I didn't check if we ever do that, some of the auto accept and introducer stuff is convoluted.)

> It also wasn't 100% certain what the bool return meant at first glance.

That tips it in favor of not having a bool to me: Less potential for confusion.

lib/config/commit_test.go
```diff
@@ -43,9 +43,8 @@ func (validationError) String() string {
 
 func replace(t testing.TB, w Wrapper, to Configuration) {
 	t.Helper()
-	waiter, err := w.Modify(func(cfg *Configuration) bool {
+	waiter, err := w.Modify(func(cfg *Configuration) {
 		cfg = &to
```
calmh, line 47: This can't possibly work?

lib/config/commit_test.go
```diff
@@ -93,9 +92,8 @@ func TestReplaceCommit(t *testing.T) {
 
 	w.Subscribe(validationError{})
 
-	_, err := w.Modify(func(cfg *Configuration) bool {
+	_, err := w.Modify(func(cfg *Configuration) {
 		cfg = &Configuration{Version: 3}
```
calmh, line 96: Likewise

lib/model/testutils_test.go
```diff
@@ -339,9 +337,8 @@ func newFileSet(t testing.TB, folder string, fs fs.Filesystem, ldb *db.Lowlevel)
 
 func replace(t testing.TB, w config.Wrapper, to config.Configuration) {
 	t.Helper()
-	waiter, err := w.Modify(func(cfg *config.Configuration) bool {
+	waiter, err := w.Modify(func(cfg *config.Configuration) {
 		*cfg = to
```
calmh, line 341: 👍

calmh (commented):
LGTM apart from that which is impossible

imsodin:
The explanation for the impossibility: The entire test file with the "defect" is one large no-op. The only test within is skipped. I was unable to make the test fail (after fixing), so I'll unskip it to see what CI has to say.

calmh:
Aah it's `TestReplaceCommit`! I remember that one, I think I was the one finally adding the skip. 😬

imsodin:
And, do you want to give it a new chance? Or is it too 😬

calmh:
Who knows, this change might even fix it?! Let's see.

calmh (approved):

cmd/syncthing/main.go
```diff
@@ -618,20 +618,20 @@ func syncthingMain(runtimeOptions RuntimeOptions) {
 	// environment variable is set.
 
 	if build.IsCandidate && !upgrade.DisabledByCompilation && !runtimeOptions.NoUpgrade {
-		l.Infoln("Automatic upgrade is always enabled for candidate releases.")
```
calmh, line 621: This deadlocks because we're not running the service yet...

PR #7202, opened 2020-12-16
lib: Consistently set suture logging

imsodin:
I got some log lines like this in my testing logs:

```
2020/12/16 11:57:51 broadcastBeacon: Failed service 'Service@0xc000046500 created by broadcastBeacon@0xc000272460/reader' (1.000000 failures of 2.000000), restarting: true, error: listen udp4 :21027: bind: address already in use
```

The supervisor for local discovery didn't set the suture event (log) hook, thus it got logged to stdout via `log`.

To prevent that oversight I added an `eventHook` argument to our `util.Spec` convenience function.

### DIFF
--- a/lib/api/api_test.go
+++ b/lib/api/api_test.go
@@ -119,7 +119,7 @@ func TestStopAfterBrokenConfig(t *testing.T) {
 	defer os.Remove(token)
 	srv.started = make(chan string)
 
-	sup := suture.New("test", util.Spec(func(e suture.Event) { l.Debugln(e) }))
+	sup := suture.New("test", util.SpecWithDebugLogger(l))
 	sup.Add(srv)
 	ctx, cancel := context.WithCancel(context.Background())
 	sup.ServeBackground(ctx)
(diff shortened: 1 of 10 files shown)


acolomb (approved):
I've seen such messages as well.  With these changes, they have indeed vanished.  Nice approach :-)

lib/api/api_test.go
```diff
@@ -119,7 +119,7 @@ func TestStopAfterBrokenConfig(t *testing.T) {
 	defer os.Remove(token)
 	srv.started = make(chan string)
 
-	sup := suture.New("test", util.Spec())
+	sup := suture.New("test", util.Spec(func(e suture.Event) { l.Debugln(e) }))
```
AudriusButkevicius, line 122: I think this would be less ugly if we had:
```
util.SpecWithDebugLogger(l)
util.SpecWithInfoLogger(l)
```
or
```
util.SpecWithLogger(l, log.INFO_LEVEL)
```

AudriusButkevicius (approved):

PR #7204, opened 2020-12-17
lib/config: Remove deprecated pending entries from config (ref #6443)

imsodin:
Removes the pending entries in config, which live in db since #6443.  
We discussed that there's no need to transfer existing pending entries in config over to db as that's low-value info and will get reintroduced on next connect (and would also be hard to do due to package dependency issues), but they still should be erased.

### DIFF
--- a/lib/config/config.go
+++ b/lib/config/config.go
@@ -30,7 +30,7 @@ import (
 
 const (
 	OldestHandledVersion = 10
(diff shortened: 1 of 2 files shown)


acolomb (approved):
Looks good. Did some testing with my dummy nodes and it worked just as expected.

Issue #7179, opened 2020-12-05
Pending devices and folders should have an optional expiry date


acolomb:
Currently pending device or folder notifications stay active basically forever.  For folders no longer offered, there is #5187 to stop showing them.  In addition, there should be an option to set a maximum age for such notifications.  It would be an opt-in setting of the general configuration section, with zero as default value meaning "never expire" (similar to the watch interval setting).  Other values could be given in hours or days to enable the feature.  If the observed timestamp is older than the configured age, the notification just disappears.

Implementation wise, this could live in the backend, or just a check within the GUI - to be discussed.  The latter would need a separate endpoint to "dismiss" pending notifications instead of ignoring or accepting them (which is a useful thing in itself).

AudriusButkevicius:
I agree for stuff that is no longer offered, we should not show it, but for everything else, the user should action these things, so I disagree.

You haven't actually explained why would anyone want this.

acolomb:
> You haven't actually explained why would anyone want this.

Let me give an example: User A has several devices, and opens the GUI very seldomly on some of them.  User B wants to share something to A and gets told the wrong device ID W by accident.  They notice the mistake and B removes the wrong (not yet accepted) device W and adds the correct ID.  Now the notification "B's device wants to connect" stays forever on the wrong device W.  When User A looks at the GUI later, the notification shows a long outdated intention.

Arguably not many users will opt-in to the feature, but those who like cleanliness might welcome the timed automatic cleanup to more easily see what's still relevant.  I don't think it needs a very complex implementation, so I think it's still in proportion to the (small) benefit.

AudriusButkevicius:
Both sides know there was some screw up, so there are no surprises there that need machine intervention.

I'd rather not have the code for this, as this only affects 0.01% of the user base, happens 0.01% of the time, yet the code will be there forever

acolomb:
Part of my argument was that it's not a single click if you're not already signed into the GUI of some device.  I think many people almost never look at the GUI, be it on an unattended NAS for example or one of my colleagues who don't really get this whole Syncthing at all but are just happy the folders are always in sync.

But I see your point, let's just hear some other opinions if this would be desired.  I don't have any strong ambitions here, just that my planned solution to #5187 would already get us very close to allow this.  So more of a "what else could we do with this code" suggestion :-)

AudriusButkevicius:
If you are not signed in, and not in the UI, whether the notification is there or not does not matter to you.

[...]

PR #7205, opened 2020-12-17
lib/model: Forget pending folders no longer announced in ClusterConfig (fixes #5187)

### LINKED ISSUE #5187: Remove pending folders if remote device no longer announces them
This is somewhat niche and not critical, but could be considered a minor bug - all in all I labeled it as enhancement. A folder that was announced by a remote, then got removed on the remote, keeps being displayed as a new folder. It would be nice if pending folders in the config would be checked when receiving the cluster config and no longer pending devices were removed. This issue is relevant to .51 RC, so if fixed soon it could be included, but I don't consider it a necessity.

acolomb:
### Purpose

When receiving a `ClusterConfig` message, walk through possible pending folder DB entries for that device and discard any that were not just touched.  If there were any entries that now expired, generate a new API event `FolderOfferCancelled` and let the GUI refresh the `/rest/cluster/pending/*` endpoints.

### Testing

Manually tested by adding and afterwards removing a pending folder from another device. The notification disappeared almost instantly in reaction to the ClusterConfig exchange.

### Documentation

New event type is added to the API documentation, see syncthing/docs#592.


acolomb:
Note that the implementation uses the already recorded timestamp to avoid building up new complexities.  In addition, that makes it easy to define a "grace period" if ever needed, or something along the lines of #7179.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -272,6 +272,11 @@ angular.module('syncthing.core')
             $scope.pendingFolders[arg.data.folder] = pendingFolder;
         });
 
+        $scope.$on(Events.FOLDER_OFFER_CANCELLED, function (event, arg) {
```
AudriusButkevicius, line 275: I don't think we should have a cancel event.

Effectively what we care about, is when the folder list is updated.
So far we used to rely on folder_rejected event/config_saved, but perhaps it now should be pending_folders_changed, or something.
If you really care what was added what was removed the "data" part of the event could contain that.

lib/model/model.go
```diff
@@ -1338,6 +1339,14 @@ func (m *model) ccHandleFolders(folders []protocol.Folder, deviceCfg config.Devi
 	}
 
 	indexSenders.removeAllExcept(seenFolders)
+	if expired, err := m.db.ExpirePendingFolders(deviceID, handleTime); err != nil {
```
AudriusButkevicius, line 1342: I don't like this mechanism, as it's not obvious from this what gets added or removed.
I think you should explicitly, check what's in the CC, check what's in the database, get a list of what needs to be removed, remove it, potentially debug log it.

gui/default/syncthing/core/syncthingController.js
acolomb: So far there was only `FolderRejected` which contained all details so that the GUI updated its list in-place without querying the endpoint (formerly config, now cluster) again. I didn't want to make a breaking event API change, and reusing the event with that name would be confusing anyway.

However, no-longer-pending folders should happen very rarely and in my opinion querying the endpoint again and replacing the whole list is justified (as is done on config changes).

So we can just rename the event to `PendingFoldersChanged` or even more generic `ClusterPendingChanged`. But I don't think removing `FolderRejected` should be touched or replaced. Not sure if you were suggesting that.

lib/model/model.go
acolomb: > I don't like this mechanism, as it's not obvious from this what gets added or removed.

Strange, I like it very much for its efficiency and simplicity.  A one-line comment and maybe renaming the function to `ExpirePendingFoldersBeforeTime` should be enough to make it clearer?

> I think you should explicitly, check what's in the CC, check what's in the database, get a list of what needs to be removed, remove it, potentially debug log it.

That's surely more code and more runtime overhead, but I don't see what we actually gain from juggling with lists here?  IMO the current solution is dead simple and gets the job done, no need for anything more complex.

> Now it's some sort of magical "now" value that goes in, and magic happens underneath, with no clarify what effect that has.

Nothing magical: Note when the CC is received, update all contained folders' timestamps, delete pending folders which were not just touched.  Will add a comment to make that more obvious.

gui/default/syncthing/core/syncthingController.js
acolomb: We can use the `ClusterPendingChanged` event also after removing stuff in `cleanPending()`.  It's currently handled through the `ConfigSaved` event, but that coupling leads to unnecessary API queries when nothing actually changed.

AudriusButkevicius: Why do we have to wedge in the term cluster here?

There are some pending folders and pending devices, how does adding cluster suddenly give this more meaning?

Syncthing is already heavy on new terminology, so I think we should strive to minimise new magical words that we expose to the users/external developers, as they cause confusion.

lib/model/model.go
```diff
@@ -1338,6 +1339,15 @@ func (m *model) ccHandleFolders(folders []protocol.Folder, deviceCfg config.Devi
...
 
 	indexSenders.removeAllExcept(seenFolders)
+	// All current pending folders were touched above, so discard any with older timestamps
+	if expired, err := m.db.RemovePendingFoldersBeforeTime(deviceID, handleTime); err != nil {
+		l.Infof("Could not clean up pending folder entries: %v", err)
+	} else if expired > 0 {
+		m.evLogger.Log(events.ClusterPendingChanged, map[string]interface{}{
+			"device": deviceID.String(),
+			"count":  expired,
+		})
```
AudriusButkevicius, line 1349: I think it would be more useful if it contained what was added/removed etc, so that receivers of the event would not even need to go to the other endpoints to reconcile the current state.

lib/db/observed.go
```diff
@@ -111,6 +111,42 @@ func (db *Lowlevel) RemovePendingFolder(id string) {
...
+			continue
+		} else if err = of.Unmarshal(bs); err != nil {
+			// Keep invalid entries
+			continue
+		} else if !of.Time.Before(oldest) {
+			// Keep entries younger or equal to the given timestamp
+			continue
+		}
+		if err := db.Delete(iter.Key()); err != nil {
+			l.Infof("Failed to remove pending folder entry: %v", err)
```
AudriusButkevicius, line 140: Should be warning?

lib/db/observed.go
```diff
@@ -111,6 +111,42 @@ func (db *Lowlevel) RemovePendingFolder(id string) {
...
+	var count uint
+	for iter.Next() {
+		var bs []byte
+		var of ObservedFolder
+		if bs, err = db.Get(iter.Key()); err != nil {
+			// Keep inaccessible entries
+			continue
+		} else if err = of.Unmarshal(bs); err != nil {
+			// Keep invalid entries
+			continue
```
AudriusButkevicius, line 134: Why keep all of these dodgy things, and skim over them in silence?

lib/model/model.go
```diff
@@ -1338,6 +1339,15 @@ func (m *model) ccHandleFolders(folders []protocol.Folder, deviceCfg config.Devi
 	}
 
 	indexSenders.removeAllExcept(seenFolders)
+	// All current pending folders were touched above, so discard any with older timestamps
+	if expired, err := m.db.RemovePendingFoldersBeforeTime(deviceID, handleTime); err != nil {
+		l.Infof("Could not clean up pending folder entries: %v", err)
+	} else if expired > 0 {
+		m.evLogger.Log(events.ClusterPendingChanged, map[string]interface{}{
```
AudriusButkevicius, line 1346: I think this should also be emitted in places where we currently emit folder_rejected or device_rejected, to effectively have one event that represents all pending stuff changes.

gui/default/syncthing/core/syncthingController.js
acolomb: The term ClusterPending refers to the API endpoints `/rest/cluster/pending/*`.  Its intended meaning is "refreshing info from those endpoints should be done now, as it has changed in an unspecified way".

Cluster is used in the docs IIRC and to me is related to the structure which devices and shared folders are connected to each other.

lib/model/model.go
acolomb: See above, I think trying to wrap everything into the event is duplicated effort. The event should just be regarded as a trigger to refresh the info. Handling the changes individually is also prone to get the local (GUI) cache out of sync with the endpoint data. 

Having a full-fledged FolderRejected event in addition is due to legacy and keeping compatibility. It's good to include the additional details there so it's usable immediately, not strictly requiring a refresh.

lib/db/observed.go
acolomb: Right, thanks.

gui/default/syncthing/core/syncthingController.js
imsodin: I agree in principle. However only "pending" feels very obscure too - what is "pending"? And I can't easily come up with a convincing alternative for the meaning of "linked devices and folders" or "introducing/altering device/folder pairings". Maybe "pairing"? Maybe that's more of a common/comprehensible word than "cluster" and could work as `PendingPairingChanged` and `/rest/pairing/pending/folders`. I am not terribly convinced though.

lib/db/observed.go
acolomb: Might just as well delete them, just like `PendingFolders()` does. I did not plan on that since it's already done there and I find it confusing if the functions removes anything other than it was requested to do.

So should I remove them here?

gui/default/syncthing/core/syncthingController.js
acolomb: I need to get my "candidate folders / devices" branch cleaned up and ready for PR. Because that's always in the back of my head when searching for the right terminology and it will also live under `/rest/cluster/candidate/*`. The endpoint name btw. is now already in the docs, no?

@imsodin are you unhappy with the endpoint naming or just the event name? It's meant to spell out explicitly that exactly this endpoint (subtree) needs refreshing.

lib/model/model.go
acolomb: I am actually happy with additions / updates being handled without an extra API query and removals (or anything else) having a separate refresh trigger event. For consistency we should send the latter as well for additions. But I couldn't see a nice way for the consumer to avoid a needless refresh after already handling a FolderRejected event right before.

So shall we:

1. Send both events for a while, with one marked deprecated and having the same contents.

2. Switch the GUI to the new event immediately.

3. Handle contained detailed changes in-place, and fall back to an API refresh if they are not provided.

4. Remove the old event completely at some point.

The desired granularity is not obvious to me. We could for example unify FolderRejected and DeviceRejected, or keep them separate in the new structure?

gui/default/syncthing/core/syncthingController.js
imsodin: The candidate thing just plays into the same: It could also be `/rest/pairing/candidate/...`, or just `/rest/candidate/...`. All my the already mentioned points apply to that as well. And I am still not super happy with either "cluster" or "pairing". Then again, maybe we are bikeshedding here and should just accept "cluster" as a fact. I don't care much either way, I just tried to express some ideas to hopefully advance the discussion towards a solution.

They (endpoints and events) definitely need to be consistent. However the new `/rest/cluster/...` stuff hasn't been released, so now is the time to change it if we want to.

acolomb: I slightly prefer cluster over pairing, but could live with either. Since cluster is in fact mentioned several times in the documentation, I thought it was common terminology on the developer front.

AudriusButkevicius: The folder and device is pending sharing?

I don't like candidate or pairing as it's yet a new word.

If we converge on something else, we should definitely update end rest urls.

lib/db/observed.go
AudriusButkevicius: I guess warn and remove? Definately not sit in silence and do nothing.

lib/model/model.go
AudriusButkevicius: I think we keep old events as they were, and new event should contain everything the consumer needs (in order to avoid hitting the endpoint), meaning list of additions removals etc.

I think the epiphany is that someday the ui will be completely event sourced, and the endpoints will only be called once on page load to bootstrap.

gui/default/syncthing/core/syncthingController.js
acolomb: Candidate stuff is (will be) in fact new functionality, so a new term seems justified. It's closely related to pending though, both derived from the ClusterConfig message contents. Guess that's where I got the "cluster" from...

I agree this is turning into bikeshedding though, so will try to focus on agreed changes until someone comes up with a better term that could lead to consensus.

lib/model/model.go
acolomb: Regarding the GUI code, I read that deleting properties from a JS object is bad for browser performance. So on that end, should it still do a filter-and-replace dance or is it okay to simply `delete(pendingFolders[folderID], deviceID)`?

lib/model/model.go
acolomb: Thanks for your comments above @AudriusButkevicius, but what's your take on granularity? Coalesce addition and removal into one "changed" event I guess? But keep device and folder changes in separate events?

AudriusButkevicius: I think it's fine to have it in a single event, as both deletions amd additions happen at the same time, upon connecting, so it's likely we'll have both additions and removals populated.

acolomb: Currently we send the `...Rejected` events for each thing individually.  You're suggesting to collect that and send one event per ClusterConfig?

acolomb:
I incorporated the changes for dropping invalid entries and drafted the new event types with all needed details included.

The GUI is not adapted yet, and the code in model obviously still too verbose. Just wanted to ask for feedback on the event contents, as the amount of code looks a little scary to me now compared to the original suggestion.

lib/model/model.go
AudriusButkevicius: Idk, I don't think we microoptimise our js code.

lib/model/model.go
AudriusButkevicius: I suggest to leave the old event as it was, and just send the new one in the new format. You can even send multiple new ones, but they should allow the state to be event sourced.

acolomb: Hope the current draft meets that expectation.  Still needs some cleanup, but it works in principle, using the two new events (`Pending...Changed`).  Old ones (`...Rejected`) are in fact untouched, just no longer used in our GUI.

acolomb:
This should be ready for another review.  Docs PR https://github.com/syncthing/docs/pull/592 updated as well to show the new event structure.

lib/model/model.go
acolomb: Addressed in latest commits.

lib/db/observed.go
```diff
@@ -111,6 +111,41 @@ func (db *Lowlevel) RemovePendingFolder(id string) {
...
+				of.Label, device.Short(), of.Time)
+		} else {
+			// Keep entries younger or equal to the given timestamp
+			continue
+		}
+		if err := db.Delete(iter.Key()); err != nil {
+			l.Warnf("Failed to remove pending folder entry: %v", err)
+		}
+		folderID := string(db.keyer.FolderFromPendingFolderKey(iter.Key()))
+		res = append(res, folderID)
```
AudriusButkevicius, line 144: I think we only want to append in the "expired" case?
Or do we really want to report folders that we can't read/unmarshal/etc as removed?
I guess in theory they are removed, but if Get(iter.Key()) failed, not sure what we are removing?

lib/model/model.go
```diff
@@ -1298,6 +1300,12 @@ func (m *model) ccHandleFolders(folders []protocol.Folder, deviceCfg config.Devi
 				l.Warnf("Failed to persist pending folder entry to database: %v", err)
 			}
 			indexSenders.addPending(cfg, ccDeviceInfos[folder.ID])
+			updatedPending = append(updatedPending, map[string]string{
+				"folder":      folder.ID,
+				"folderLabel": folder.Label,
```
AudriusButkevicius, line 1305: I think I'd have either `id, label` or `folderId, folderLabel`,

lib/model/model.go
```diff
@@ -2755,27 +2790,41 @@ func (m *model) cleanPending(existingDevices map[protocol.DeviceID]config.Device
 			// at all (but might become pending again).
 			l.Debugf("Discarding pending removed folder %v from all devices", folderID)
 			m.db.RemovePendingFolder(folderID)
+			removedPendingFolders = append(removedPendingFolders, map[string]string{
+				"folder": folderID,
```
AudriusButkevicius, line 2794: again, id or folderId.

lib/model/model.go
```diff
@@ -2755,27 +2790,41 @@ func (m *model) cleanPending(existingDevices map[protocol.DeviceID]config.Device
...
-					m.db.RemovePendingFolderForDevice(folderID, deviceID)
+					goto removeFolderForDevice
 				}
 			}
+			continue
+		removeFolderForDevice:
+			m.db.RemovePendingFolderForDevice(folderID, deviceID)
+			removedPendingFolders = append(removedPendingFolders, map[string]string{
+				"folder": folderID,
+				"device": deviceID.String(),
```
AudriusButkevicius, line 2817: folderId deviceId

lib/model/model.go
```diff
@@ -2755,27 +2790,41 @@ func (m *model) cleanPending(existingDevices map[protocol.DeviceID]config.Device
...
+			})
 		}
 	}
+	if len(removedPendingFolders) > 0 {
+		m.evLogger.Log(events.PendingFoldersChanged, map[string]interface{}{
+			"removed": removedPendingFolders,
```
AudriusButkevicius, line 2823: Would we not want removedPendingFolders to be an array of objects that have a folderId, oppose to an array of strings, just in case in the future we add some extra information about the removal (such as when it was first added etc).

lib/model/model.go
```diff
@@ -2784,14 +2833,21 @@ func (m *model) cleanPending(existingDevices map[protocol.DeviceID]config.Device
...
+			goto removeDevice
 		}
+		continue
+	removeDevice:
+		m.db.RemovePendingDevice(deviceID)
+		removedPendingDevices = append(removedPendingDevices, deviceID.String())
+	}
+	if len(removedPendingDevices) > 0 {
+		m.evLogger.Log(events.PendingDevicesChanged, map[string]interface{}{
+			"removed": removedPendingDevices,
```
AudriusButkevicius, line 2849: Ditto about objects

AudriusButkevicius:
In principle looks good, small comments.

lib/model/model.go
acolomb: `id` might be confused with the event's sequence ID.  It's otherwise only used in `DeviceConnected` and `DeviceDisconnected` so far.

`folderID` is used once in `LocalChangeDetected`.

Most other event types agree on `device` and `folder` fields for the IDs. Should we try to stay consistent to that, or set a new standard starting here?

lib/model/model.go
acolomb: Hm? In this case, it *is* an array of objects, because in most cases folder *and* device ID must be specified.

lib/model/model.go
acolomb: Okay I'll make it consistent with the removed folders case.

lib/db/observed.go
acolomb: >  Or do we really want to report folders that we can't read/unmarshal/etc as removed?

Hard question.  It should never make a difference because every time the current list is requested through `PendingFolders()`, that will already weed out invalid entries.  OTOH, that happens without `cleanPending()` ever noticing so will also *not* be reported in the event.

For the sake of lower code complexity I'd prefer the current approach.  If it ever reports something as removed which the GUI didn't have in its cached list, that will fail gracefully.  It's also our only chance to report this over the event API and certainly doesn't hurt for keeping the consumer's cache in sync.

imsodin: At the very least it's wrong the report something as removed when `Delete` failed. I'd also tend to only report truly expired entries. And doesn't feel any more complex to move the two lines above into the second if block above.

lib/model/model.go
imsodin: `folder` in JS sometimes means ID, but there's also places where it's the entire config. Using `folderId` removes that ambiguity and that outweighs consistency with other events to me (respectively we might want to add a `folderId` field to those as well - not this PR though).

acolomb: Okay, will change. For consistency, it should be `folderID` and `deviceID` though, with all-caps ID?

lib/db/observed.go
acolomb: You're right about failed deletions.  That problem isn't solved by moving the `append()` call before the actual `Delete()` though.  I conditionalized it now based on the folderID variable, which is not populated for invalid entries.

acolomb:
Should we try to get this into v1.13.0?  It would bring a user-visible benefit of the (hopefully invisible) change #6443.

Please indicate if there are any further change requests beyond the addressed review comments.

imsodin (commented):
Looks good to merge from my side, but I want @AudriusButkevicius to have another look in view of the previews review comments.

And I just need to say it again, even though I can live with the current state:  
I consider the control flow as easier to follow with small functions as opposed to the `goto`-`continue`-constructs in this PR.

acolomb:
> I consider the control flow as easier to follow with small functions as opposed to the `goto`-`continue`-constructs in this PR.

Thanks for your review.  I see your point about avoiding `goto`, and I won't argue in its favor, but wanted to explain my reasoning:

* With a separate function, you'd still need the `continue` after each invocation so it's more duplication.
* Considering model.go has over 3000 lines, adding another function wouldn't help get an overview.
* Local state would need to be passed to the function, making for a longer read overall.

acolomb:
@AudriusButkevicius, just pinging in case you overlooked the review request.  Could you please have another look to wrap this up?

Note that the `/rest/cluster/...` API endpoints are now in the v1.13.0-rc.1, so the naming discussion lost some degrees of freedom :-/

AudriusButkevicius:
Sorry, I have ~50 unread github notifications, so going through them slowly.

AudriusButkevicius (approved):
LGTM

acolomb:
Thanks.  Will do a follow-up shortly to mirror the GUI changes to untrusted.

PR #7206, opened 2020-12-17
gui: Apply changes to untrusted (ref #6443)

imsodin:
Applied the GUI changes from #6443 to the untrusted files.


acolomb (approved):
Thanks for merging over the changes, I hadn't thought of that anymore.  Looks good to me, now it's a rather small diff for the encryption stuff.

PR #7209, opened 2020-12-18
lib/fs: Correct wrapping order for meaningful log-caller

imsodin:
With `fs` debug logging we print information about what called the fs method. However for case- and mtime-fileysstems, this information is entirely useless, as it just points at the overriding method in those wrappers (e.g. chmod just says it was called by chmod). To prevent that I added a utility function that ensures proper wrapping order:

```golang
// wrapFilesystem should always be used when wrapping a Filesystem.
// It ensures proper wrapping order, which right now means:
// `logFilesystem` needs to be the outermost wrapper for caller lookup.
func wrapFilesystem(fs Filesystem, wrapFn func(Filesystem) Filesystem) Filesystem {
```

### DIFF
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -98,7 +98,9 @@ type caseFilesystem struct {
 // case-sensitive one. However it will add some overhead and thus shouldn't be
 // used if the filesystem is known to already behave case-sensitively.
 func NewCaseFilesystem(fs Filesystem) Filesystem {
-	return globalCaseFilesystemRegistry.get(fs)
+	return wrapFilesystem(fs, func(ifs Filesystem) Filesystem {
+		return globalCaseFilesystemRegistry.get(ifs)
+	})
 }
 
 func (f *caseFilesystem) Chmod(name string, mode FileMode) error {
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -52,7 +52,7 @@ type caseFilesystemRegistry struct {
 	startCleaner sync.Once
 }
 
-func (r *caseFilesystemRegistry) get(fs Filesystem) *caseFilesystem {
+func (r *caseFilesystemRegistry) get(fs Filesystem) Filesystem {
 	r.mut.Lock()
 	defer r.mut.Unlock()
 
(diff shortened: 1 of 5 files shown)


lib/fs/casefs.go
```diff
@@ -98,7 +98,9 @@ type caseFilesystem struct {
 // case-sensitive one. However it will add some overhead and thus shouldn't be
 // used if the filesystem is known to already behave case-sensitively.
 func NewCaseFilesystem(fs Filesystem) Filesystem {
-	return globalCaseFilesystemRegistry.get(fs)
+	return wrapFilesystem(fs, func(ifs Filesystem) Filesystem {
+		return globalCaseFilesystemRegistry.get(ifs)
+	})
```
calmh, line 103:
```suggestion
	return wrapFilesystem(fs, globalCaseFilesystemRegistry.get)
```

if we change the return type of `get` to Filesystem

lib/fs/filesystem.go
```diff
@@ -260,6 +260,21 @@ func Canonicalize(file string) (string, error) {
 	return file, nil
 }
 
+// wrapFilesystem should always be used when wrapping a Filesystem.
+// It ensures proper wrapping order, which right now means:
+// `logFilesystem` needs to be the outermost wrapper for caller lookup.
```
calmh, line 265: We have two different wrapping orders of logfs and walkfs, depending on whether we want to debug above or below walkfs... Probably that does not matter, and maybe we should remove the `walkfs` trace option as this is potentially confusing

calmh (approved):
LGTM unless you wanna tweak

lib/fs/filesystem.go
imsodin: Having both `fs` and `walkfs` has merits: The former will just print one debug line that it will start walking now, the latter will also print every filesystem operation that occurs while walking - I think I remember I needed that huge walking verbosity once. And the unwrapping added in this PR is correct for both cases: It's appropriate for the filesystem calls originating in the walking code to give the fs walking code as the caller.

lib/ignore/ignore_test.go
```diff
@@ -1192,3 +1192,31 @@ func TestEmptyPatterns(t *testing.T) {
 		}
 	}
 }
+
+func TestUnignoredParent(t *testing.T) {
```
calmh, line 1196: Random test case is random?

calmh (approved):

PR #7212, opened 2020-12-18
lib: Close underlying conn in protocol (fixes #7165)

### LINKED ISSUE #7165: Connections aren't actually closed when closing a protocol connection
Came up starting in https://github.com/syncthing/syncthing/pull/7141#issuecomment-733247434, short summary: A protocol connection doesn't control the underlying tls connection, thus it cannot close it. All the methods that supposedly close it, just tell the model that it's closed and potentielly send a Closed BEP msg, but don't actually close the connection.

The type hierarchy looks like this:  
model has: `connections.Service` implemented by `completeConn{internalConn, protocol.Connection}` and `internalConn` are wrappers of the concrete underlying tls connection type (tcp, quic, ...).  
The protocol connections gets a reader and a writer based on `internalConn`. An option would be to instead pass it a `io.ReadWriteCloser`, based on the same.

imsodin:
### Purpose

From https://github.com/syncthing/syncthing/issues/7165:
> Came up starting in #7141 (comment), short summary: A protocol connection doesn't control the underlying tls connection, thus it cannot close it. All the methods that supposedly close it, just tell the model that it's closed and potentielly send a Closed BEP msg, but don't actually close the connection.

What I did is to move the entire connection interface to lib/protocol, and pass the necessary pieces (closer and connection info) to from lib/connections to lib/protocol. The main advantage is that the protocol can now close the underlying connection on any error. Additional it's also a bit cleaner, removing some type aggregation and an interface in lib/connections.

### Testing

No new tests. The changes are just adjustments to the new API.  
My testing is nothing more than: Connections still do drop and reconnect on pause/unpause.

### DIFF
--- a/lib/testutils/testutils.go
+++ b/lib/testutils/testutils.go
@@ -8,6 +8,7 @@ package testutils
 
 import (
 	"errors"
+	"net"
 	"sync"
 )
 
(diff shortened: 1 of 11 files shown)


lib/testutils/testutils.go
```diff
@@ -52,3 +53,49 @@ func (rw *NoopRW) Read(p []byte) (n int, err error) {
 func (rw *NoopRW) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
+
+type NoopCloser struct{}
```
calmh, line 57: `ioutil.NopCloser(nil)` but we can also have our own, skips the nil reader baggage 🤷

calmh (approved):

PR #7216, opened 2020-12-20
gui: Allow device removal even if possibly reintroduced (fixes #5426)

### LINKED ISSUE #5426: Old device removed from introducer isn't removable from other devices
As explained in [this forum post](https://forum.syncthing.net/t/unable-to-remove-device-that-has-been-removed-from-introducer/12638)

OLD_DEVICE has been introduced to REMOTE_DEVICE in the past by INTRODUCER_DEVICE
OLD_DEVICE has since been removed from INTRODUCER_DEVICE
REMOTE_DEVICE can't remove OLD_DEVICE because "this device will be reintroduced by INTRODUCER_DEVICE"

Syncthing Version: v1.0.0 (previously same bug on 0.14.54)
OS Version: Ubuntu 16.04
Browser Version: Firefox latest / Safari latest

capture d ecran 2019-01-03 a 17 38 21

capture d ecran 2019-01-03 a 17 40 40

hiqua:
### Purpose

Fixes #5426 according to https://github.com/syncthing/syncthing/issues/5426#issuecomment-748635865

Not approved yet, but since the PR is small enough, I went ahead in case you think it makes sense.

### Testing

Not tested yet.

### DIFF
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -37,10 +37,6 @@
             <p translate ng-if="currentDevice.deviceID == myID" class="help-block">Shown instead of Device ID in the cluster status. Will be advertised to other devices as an optional default name.</p>
             <p translate ng-if="currentDevice.deviceID != myID" class="help-block">Shown instead of Device ID in the cluster status. Will be updated to the name the device advertises if left empty.</p>
           </div>
-          <div class="form-group" ng-if="willBeReintroducedBy">
-            <label translate for="removal">Device Removal</label>
-              <p class="text-danger">When you remove this device, it might be automatically reintroduced by {{ willBeReintroducedBy }}.</p>
-          </div>
         </div>
         <div id="device-sharing" class="tab-pane">
           <div class="row">
@@ -114,10 +110,6 @@
               </div>
             </div>
           </div>
-          <div class="form-group" ng-if="willBeReintroducedBy">
-            <label translate for="removal">Device Removal</label>
-              <p class="text-danger">When you remove this device, it might be automatically reintroduced by {{ willBeReintroducedBy }}.</p>
-          </div>
         </div>
         <div id="device-advanced" class="tab-pane">
           <div class="row form-group">
@@ -164,13 +156,13 @@
               </div>
             </div>
           </div>
-          <div class="form-group" ng-if="willBeReintroducedBy">
-            <label translate for="removal">Device Removal</label>
-              <p class="text-danger">When you remove this device, it might be automatically reintroduced by {{ willBeReintroducedBy }}.</p>
-          </div>
         </div>
       </div>
     </form>
+    <div class="form-group" ng-if="willBeReintroducedBy">
+      <label translate for="removal">Device Removal</label>
+      <p class="text-danger">When you remove this device, it might be automatically reintroduced by {{ willBeReintroducedBy }}.</p>
+    </div>
   </div>
   <div class="modal-footer">
     <button type="button" class="btn btn-primary btn-sm" ng-click="saveDevice()" ng-disabled="deviceEditor.$invalid">
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -175,10 +175,7 @@
       <span class="fas fa-times"></span>&nbsp;<span translate>Close</span>
     </button>
     <div ng-if="editingExisting" class="pull-left">
-      <button type="button" class="btn btn-warning btn-sm" ng-if="willBeReintroducedBy" tooltip data-original-title="{{ willBeReintroducedBy }} {{'might reintroduce this device' | translate}}.">
-        <span class="fas fa-minus-circle"></span>&nbsp;<span translate>Remove</span>
-      </button>
-      <button type="button" class="btn btn-warning btn-sm" data-toggle="modal" data-target="#remove-device-confirmation" ng-if="!willBeReintroducedBy">
+      <button type="button" class="btn btn-warning btn-sm" data-toggle="modal" data-target="#remove-device-confirmation">
         <span class="fas fa-minus-circle"></span>&nbsp;<span translate>Remove</span>
       </button>
     </div>
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -161,7 +161,7 @@
     </form>
     <div class="form-group" ng-if="willBeReintroducedBy">
       <label translate for="removal">Device Removal</label>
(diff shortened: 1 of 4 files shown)


imsodin:
I tend to agree that given there is not unlikely path to end up in this state (device not removable in GUI, but it could be removed), it's better not to disable the remove button. However I think then the message about the device that might reintroduce it should be a directly displayed as a warning at the bottom of the modal body, not just a tooltip. Or alternatively make it a confirmation dialog on clicking remove, but that seems like too much work for the gain.

hiqua:
@imsodin  Is this what you had in mind?

2020-12-21-111051_891x265_scrot

imsodin:
Yep.  
Just an idea on the wording - that's really meant as an idea only and you are very welcome to not like it and stick with yours ;) :  
"When you remove this device, it might be automatically reintroduced by $dev".

hiqua:
What's the difference between `default/untrusted/...` and `default/...`? Do I need to do the changes in both?

imsodin:
1. If you don't want to know, you don't need to read 2. - I can take care of it once this is good to merge (or after merge).
2. It's code for the untrusted feature preview. It's basically a duplicate of the files directly in `default/...` with some additional changes. See https://forum.syncthing.net/t/early-testing-for-untrusted-encrypted-devices/15873

hiqua:
Ah I didn't make the connection with the upcoming untrusted device feature, ok then I'll try to apply the changes there as well.

imsodin (commented):
Could you move the removal block outside of the form/tabs div, such that it doesn't need to be repeated every time.

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -164,13 +156,13 @@
...
-          <div class="form-group" ng-if="willBeReintroducedBy">
-            <label translate for="removal">Device Removal</label>
-              <p class="text-danger">When you remove this device, it might be automatically reintroduced by {{ willBeReintroducedBy }}.</p>
-          </div>
         </div>
       </div>
     </form>
+    <div class="form-group" ng-if="willBeReintroducedBy">
+      <label translate for="removal">Device Removal</label>
+      <p class="text-danger">When you remove this device, it might be automatically reintroduced by {{ willBeReintroducedBy }}.</p>
```
imsodin, line 164:
```suggestion
      <p class="text-danger" translate translate-value-introducer="{{ willBeReintroducedBy }}">When you remove this device, it might be automatically reintroduced by {%introducer%}.</p>
```

hiqua:
Let me know if there's anything else I should do.

gui/default/untrusted/syncthing/device/editDeviceModalView.html
```diff
@@ -146,6 +146,10 @@
         </div>
       </div>
     </form>
+    <div class="form-group" ng-if="willBeReintroducedBy">
+      <label translate for="removal">Device Removal</label>
+      <p class="text-danger">When you remove this device, it might be automatically reintroduced by {{ willBeReintroducedBy }}.</p>
```
calmh, line 151: Missing translate handling here

gui/default/untrusted/syncthing/device/editDeviceModalView.html
```diff
@@ -158,7 +162,7 @@
       <span class="fas fa-times"></span>&nbsp;<span translate>Close</span>
     </button>
     <div ng-if="editingExisting" class="pull-left">
-      <button type="button" class="btn btn-warning btn-sm disabled" ng-if="willBeReintroducedBy" tooltip data-original-title="This device will be reintroduced by {{ willBeReintroducedBy }}">
+      <button type="button" class="btn btn-warning btn-sm" ng-if="willBeReintroducedBy" tooltip data-original-title="{{ willBeReintroducedBy }} might reintroduce this device.">
```
calmh, line 165: ... as well as in all the tooltips, though I see it also wasn't translated previously. Perhaps we never translate in tooltips because it doesn't work? Or we use the ` | translate` angular thing?

gui/default/untrusted/syncthing/device/editDeviceModalView.html
calmh: Not entirely sure this merits `class="text-danger"` on the entire text. We don't have danger-colored texts in other places, do we? All in all the removal is harmless so I think the danger is minimal.

gui/default/syncthing/device/editDeviceModalView.html
hiqua: The interpolation on introducer does not seem to be working, are we missing something?

gui/default/untrusted/syncthing/device/editDeviceModalView.html
hiqua: @imsodin handled the translation, I just forgot to sync the untrusted files.

gui/default/untrusted/syncthing/device/editDeviceModalView.html
hiqua: See previous comment. But there's still a problem with the interpolation.

imsodin:
I completely lost track of this PR, sorry about the wait.

The interpolation issue is likely because of the missing translation string. You can run `go run build.go translate` locally which will update the translation strings.

I just tried to reproduce the issues and see how it looks, and realized we already have a confirmation dialog for removing devices. In that case I would not add a notice to the edit device dialog (there's already a lot of content there), but add a notice to the confirmation dialog (removeDeviceDialogView.html).

Thanks in advance for your patience and generally your contribution!

hiqua:
@imsodin ok I'll look into that, thanks a lot for the feedback. I actually wanted to move forward later today so the timing is perfect. In any case I should be able to update this in a few days at the latest.

hiqua:
I'll update the untrusted part once we have the final version.

gui/default/syncthing/device/editDeviceModalView.html
```diff
@@ -159,6 +159,10 @@
         </div>
       </div>
     </form>
+    <div class="form-group" ng-if="willBeReintroducedBy">
+      <label translate for="removal">Device Removal</label>
+      <p class="help-block" translate translate-value-introducer="{{ willBeReintroducedBy }}">When you remove this device, it might be automatically reintroduced by {%introducer%}.</p>
+    </div>
```
imsodin, line 165: I didn't properly explain what I meant in my last comment: I intended to remove/replace this part with the notice in the remove device dialog, i.e. to remove this div block.

imsodin:
lgtm, apart from translations (suggest you undo current changes and run go run build.go translate again) and untrusted.

gui/default/syncthing/device/editDeviceModalView.html
hiqua: Sorry it was clear, I just went too fast and should have double-checked.

imsodin (approved):

PR #7223, opened 2020-12-21
lib/config, lib/connections: Add optional connection limits (fixes #7176)

### LINKED ISSUE #7176: Active connections min/max setting
This is a draft for a feature to handle the following two use cases.

1. I have a lot of low powered devices and want them to communicate in (something that feels like) a full mesh, but without the overhead of each device keeping a hundreds or thousands of connections running.

2. I have a lot of clients (>1000) connecting to a small number of servers (<10), for load balancing purposes. Clients should automatically balance connections over the servers in question so they can handle one going down, without ending up with too many clients connecting to any given server, potentially overloading it.

Currently this is difficult because while it's easy to configure full mesh it's difficult for Syncthing to scale to it. The client-server example requires some fancy configuration management to direct clients to specific servers, and possibly reconfiguring them once a server goes down.

Now lets assume we had two configuration options available:

- Minimum active connections (minC)
- Maximum active connections (maxC)

Where minC <= maxC and both default to unbounded / infinitely large. These control the connection service in terms of establishing outgoing and incoming connections. Specifically,

- The connection loop runs as usual to establish outgoing connections while the number of active connections is less than minC. Once minC has been reached we no longer attempt to establish further outgoing connections.
- Incoming connections are accepted as normal as long as we have fewer than maxC connections. Once maxC is reached we stop accepting further incoming connections. (Similar to a paused device.)

For some fairness to work out we also make one adjustment to the connection loop:

- Every time we run the connection loop, the order of devices we connect to is shuffled.

Our two use cases above then become quite simple.

For case one we configure all our devices for full mesh and set minC=10, maxC=20. Each device will end up connected to between ten and twenty other devices. (It's possible that this might result in partitioned sets of devices, but it doesn't seem especially likely. Some calculations could probably be done on how to set minC/maxC for a given cluster size to achieve a certain probability of it being fully connected...)

For case two with for example 1000 clients and 5 servers we want about 200 connections per server. We also want to handle one server going down, leaving us with 250 connections per remaining server. We can configure the servers to connect to all clients, and clients to connect to all servers. Clients get maxC=1 -- keep precisely one active connection. Servers get minC=200, maxC=250. Hence the servers will attempt to "grab" 200 clients each, clients will attempt to establish one connection to any server. Given randomization it's likely that the distribution ends up fairly even, but it's possible some servers get up to 250 clients and some other servers fewer than 200. If a server goes down there is "space" for the clients to migrate to another server.

calmh:
This adds two new configuration options:

    // The number of connections at which we stop trying to connect to more
    // devices, zero meaning no limit. Does not affect incoming connections.
    ConnectionLimitEnough int

    // The maximum number of connections which we will allow in total, zero
    // meaning no limit. Affects incoming connections and prevents
    // attempting outgoing connections.
    ConnectionLimitMax int

These can be used to limit the number of concurrent connections in various ways.

I'm aware this change is not everyone's cup of tea, however the damage in the code is quite limited if the option isn't used.


lib/connections/service.go
```diff
@@ -439,7 +460,24 @@ func (s *service) dialDevices(ctx context.Context, now time.Time, cfg config.Con
...
+	rand.Shuffle(queue)
+
+	// Cut the queue down to the limit, if any.
+	if connectionLimit > 0 && len(queue) > allowAdditional {
+		l.Debugf("Dialing %d devices out of %d because we've reached the connection limit (%d)", allowAdditional, len(queue), connectionLimit)
+		queue = queue[:allowAdditional]
+	}
+
+	for _, entry := range queue {
+		if conn, ok := s.dialParallel(ctx, entry.id, entry.targets); ok {
```
imsodin, line 480: Dials may fail, thus instead of limiting the number ahead of dialing I'd count dialed devices, and stop once the limit is reached.

imsodin (commented):
I like how it is quite simple. My only worry is that it will require more complexity in the future to accommodate users. I assume this is meant very much as an advanced feature that is given as is and the user needs to provide the thoughts and research to make it work for their specific use-case. However I doubt that will be honored by users, and if put in place without thinking, I can see complications. Like sharing one folder with only one device, all other folders with a lot of other devices. Then the user might expect Syncthing to be smart enough to always connect to that one device even with limits.

For mobile devices or any other devices that are only intermittently connected I'd expect quite a lot of benefit (less index transfer/db writes) from caching a `ConnectionLimitEnough`-long list of devices that were most recently connected, to dial those first. Could be RAM only for disconnects, or even persisted in db to carry over the benefit Syncthing being stopped.

I also started looking into graph theory a bit to get an idea about the probability of a fully-connected mesh with a given number of nodes and edges per node. Didn't yet get anywhere conclusive - will write if I do, and maybe someone knowledgeable can chime in before that :)

proto/lib/config/optionsconfiguration.proto
```diff
@@ -68,3 +69,14 @@ message OptionsConfiguration {
     double          min_home_disk_free_pct = 9005 [deprecated = true];
     int32           max_concurrent_scans   = 9006 [deprecated = true];
 }
+
+message ConnectionLimits {
```
calmh, line 73: This seems minor enough that it should live in the same file as OptionsConfiguration -- but it could be broken out into a separate file if someone insisted.

calmh:
Moving this into a separate message means it's no longer available in the advanced config dialog. Tongue-in-cheek, that solves the issue of someone breaking their setup inadvertently using this... However I don't really think we need to make it more difficult than it already is to find advanced options. I'll consider trying to extend the config editor to support this.

lib/connections/service.go
```diff
@@ -440,8 +466,49 @@ func (s *service) dialDevices(ctx context.Context, now time.Time, cfg config.Con
...
+			break
+		}
+		idx++
+	}
+	if idx < len(queue)-1 {
+		rand.Shuffle(queue[idx:])
+	}
+
+	// Perform dials according to the queue, stopping when we've reached the
+	// allowed additional number of connections (if limited).
```
calmh, line 503: There's a bug here in that we might establish a connection here only to have it dropped in a few milliseconds due to the other side having paused us or reached a limit.

We can't easily predict the future here so one idea that strikes me is that we should remember a connection being dropped in the handshake and penalise that device in future connection attempts (sorting it at the back of the queue).

Ideally we should send `Close` messages with a cause when devices are paused and such, so we can know this on the receiving side and reduce connection attempts towards that device in general...

imsodin: Sending a close message definitely is a good idea.

The general problem of counting soon to fail connections could also be solved by sending a result channel together with the connection on `s.conns`, on which the handle-loop sends a bool whether the connection has been accepted/successful or not.

calmh: I was thinking of adding a statistic for connection life time, tracked the same way we track "last seen", updated in Closed(). When sorting devices for connection purposes we can then take into account the last connection life time, or average of the last 5 connection life times, or something like that. If it's less than a second we can conclude the device has issues talking to us and try them less often.

I don't think the channel thing will work with the current setup, as on the dialing side the connection looks accepted and  goes all the way through model.AddConnection() and such. Then it gets torn down some short interval later, probably while we're preparing or sending cluster configs.

AudriusButkevicius:
Why wouldn't we keep this in the advanced config?

Seems like a foot gun for 99.99% of the users, and you are putting it next to the umbrellas.

calmh:
I think that ties into imsodins work on defaults and the resulting thoughts on actual structure and sanity in the config values, grouping things that are conceptually multiple knobs on the same aspect of something. I don't have strong opinions on it either way, but it seems reasonable to group things somewhat, like we already to with gui params and ldap settings.

AudriusButkevicius:
I guess it's a half-way house. My personal opinion is that we should either refactor the config as a while, or keep piling on how it is, because atleast the way it is now, our existing stuff already works with the "piling" aspect.

imsodin:
Having a separate message indeed originated from thinking about defaults and the need for multiple values for "one semantic config element" (e.g. because of values where the default is non-zero, but zero is a valid value (disabled), which is unfortunate as the zero value might mean either undefined (i.e. use default) or disabled). And then I'd consider it cleaner to have a combined object with multiple values instead of multiple objects with a common name only varying in a suffix (as in `somethingValue`, somethingEnabled`, `somethingModifier`, ...). However I guess it's true that there's not much if any gain as long as it's a "half-way house" and it is not consistent, and thus not having a separate message wins due to less code (for now).

calmh:
OK, separate values again, and now with better connection sorting.

lib/connections/structs.go
```diff
@@ -193,9 +194,11 @@ type genericListener interface {
 type Model interface {
 	protocol.Model
 	AddConnection(conn protocol.Connection, hello protocol.Hello)
+	NumConnections() int
 	Connection(remoteID protocol.DeviceID) (protocol.Connection, bool)
 	OnHello(protocol.DeviceID, net.Addr, protocol.Hello) error
 	GetHello(protocol.DeviceID) protocol.HelloIntf
+	DeviceStatistics() (map[string]stats.DeviceStatistics, error)
```
imsodin, line 201: Now that it's added to the public interface, maybe it's the time to change the key from `string` to `protocol.DeviceID`?

lib/model/model.go
```diff
@@ -2049,10 +2057,14 @@ func (m *model) OnHello(remoteID protocol.DeviceID, addr net.Addr, hello protoco
...
-			return errNetworkNotAllowed
-		}
+	if len(cfg.AllowedNetworks) > 0 && !connections.IsAllowedNetwork(addr.String(), cfg.AllowedNetworks) {
+		// The connection is not from an allowed network.
+		return errNetworkNotAllowed
+	}
+
+	if max := m.cfg.Options().ConnectionLimitMax; max > 0 && m.NumConnections() >= max {
+		// We're not allowed to accept any more connections.
+		return errConnLimitReached
```
imsodin, line 2067: I'd add debug logs in both cases explaining why we drop the connection.

imsodin (dismissed):
I think this is a good and useful change in itself, I am just still worried about the incorrect use/support aspect:  
We already do have options that we do not want users to use, or that are useful if used correctly, but what's correct isn't obvious and incorrect use is bad. It's likely this will be another one of those. I'd love if we had some `DoNotTouchOptions` section in the config that we document as "provided as is, no guarantees if it breaks your setup and data". Maybe a docs page that lists all of those options or gives a blanket statement like "If you change any options not availabe through web UI folder/device editor dialogs, you are on your own" would do tool.

Two minor suggestions for the code below.

lib/model/model.go
calmh: I did that initially, but we already always Info() it:

https://github.com/syncthing/syncthing/blob/e6f0ed65be02d1ed278989f5d9fca948880e4ebd/lib/connections/service.go#L276-L280

calmh:
I hear you (both/all) about the possibility of making a not-properly-connected cluster using this. However I'm not convinced it's any worse than all the other ways I can misconfigure my setup so that it doesn't work, and there's literally a myriad of those. There is a large red do-not-touch-unless-you-know-why banner at the top of the advanced config dialog too 🤷

calmh:
Docs say to not use the option ^ :)

imsodin (approved):

PR #7242, opened 2020-12-30
lib/connections, lib/model: Track last connection duration (ref #7223)

calmh:
This adds a statistic to track the last connection duration per device. It isn't used for much in this PR, but it's available for #7223 to use in deciding how to order device connection attempts (deprioritizing devices that just dropped our connection the last time).

Also some minor refactoring of the device stat thing to make it less string based.


lib/stats/device.go
```diff
@@ -10,26 +10,34 @@ import (
...
+
+const (
+	lastSeenKey     = "lastSeen"
+	connDurationKey = "lastConnDuration"
 )
 
 type DeviceStatistics struct {
-	LastSeen time.Time `json:"lastSeen"`
+	LastSeen                time.Time `json:"lastSeen"`
+	LastConnectionDurationS float64   `json:"lastConnectionDurationS"`
```
calmh, line 24: A `time.Duration` marshals as an integer nanosecond count which isn't super intuitive on the receiving API side, so going with seconds here instead.

AudriusButkevicius: Why not handle it as time.Duration properly?
Feels confusing seeing .Seconds() and .Nanoseconds() in the same thing.

AudriusButkevicius (commented):
So this tracks the last duration, not duration per address which I question how useful it is.

lib/stats/device.go
calmh: We do need subsecond precision, because I want to tell when a connection is closed immediately due to other side having paused us, for example. That's zero seconds, which would otherwise be impossible to differentiate from the "no value" zero. (We could use `-1*time.Second` or some sentinel for that but that's also ugly.)

I made it seconds here instead of time.Duration because time.Duration marshals to for example 169461293376 while I'd much rather see 169.461293376 in the API output...

calmh:
I don't think this needs to be per address. At least I see no reason to care about what happened to some specific address long ago, I just want to know if it seems possible to keep a connection established to another device at all or if it keeps getting closed within less than a second every time it's established.

imsodin:
Per address could help with connections that get established, but then soon after terminated (by some evil AV/firewall). At least I got the impression such scenarios exist from some forum threads, maybe it's not actually a thing in reality.

calmh:
Do we need the complexity to cover for that case though, necessarily? I mean, I can add stats per address, I don't have any immediate plans for using it but we can have it. It'll require garbage collecting as well because addresses can be more or less random per connection attempt and will accumulate.

imsodin (approved):
Actually the info about per address connection time is already there for with this PR, through `protocol.ConnectionInfo.EstablishedAt`. If we wanted to do something about specific addresses failing, we could.

PR #7272, opened 2021-01-13
gui: Fix log-tailing behavior when scrolled to top (fixes #7267)

### LINKED ISSUE #7267: GUI log tailing is broken if closed when scrolled up
The log window pauses the live display ("tailing") of the log when scrolled up. If the log window is closed in this state and then reopened, it's empty and says "scroll down to continue", but there is no log data so no scrolling can happen.

Screen Shot 2021-01-12 at 08 38 19

Choongkyu:
### Purpose

Fixes reported issue #7267

### Testing

1) Open logging modal
2) Scroll to top to stop log-tailing
3) Close modal
4) Reopen modal and see log-tailing resumes and displays log instead of blank textarea as shown in issue repoort

### DIFF
--- a/gui/default/syncthing/core/logViewerModalView.html
+++ b/gui/default/syncthing/core/logViewerModalView.html
@@ -29,7 +29,7 @@
     </div>
   </div>
   <div class="modal-footer">
-    <button type="button" class="btn btn-default btn-sm" ng-click="logging.close()">
+    <button type="button" class="btn btn-default btn-sm" data-dismiss="modal">
       <span class="fas fa-times"></span>&nbsp;<span translate>Close</span>
     </button>
   </div>
--- a/gui/default/syncthing/core/logViewerModalView.html
+++ b/gui/default/syncthing/core/logViewerModalView.html
@@ -29,7 +29,7 @@
     </div>
   </div>
   <div class="modal-footer">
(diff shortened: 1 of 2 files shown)


wweich (changes requested):
Instead of creating a new function in `logging`, you should just put `$scope.logging.paused = false;` inside the hidden event handler. Then there is no need to change the markup.

calmh:
Or set paused to false at the top of show()?

lib/fs/casefs.go
```diff
@@ -84,11 +84,23 @@ func (r *caseFilesystemRegistry) get(fs Filesystem) Filesystem {
...
+		toProcess := make([]*caseFilesystem, 0, len(r.fss))
 		for _, caseFs := range r.fss {
-			caseFs.dropCache()
+			toProcess = append(toProcess, caseFs)
 		}
 		r.mut.RUnlock()
+
+		for _, caseFs := range toProcess {
+			caseFs.dropCache()
+		}
```
imsodin, line 103: Unrelated stuff sneaked in.

imsodin (changes requested):

lib/fs/casefs.go
Choongkyu: thanks for catching that! fixed.

calmh (approved):

imsodin (approved):

Issue #7270, opened 2021-01-12
Data requests block API requests that also touch the filesystem


calmh:
All folder I/O goes through the casefs for a given folder, where things get serialized due to the locking on the realCaser cache. Some API requests do folder I/O, such as /rest/db/status which tries to load the ignore patterns.

(This specific case is easy to solve, because it doesn't need to load ignore patterns, it can be satisfied with whatever is already loaded instead. I'll file a PR to improve that shortly.)

However the slightly larger issue is that the realCase can be quite slow (it does potentially lots of listdir calls) and having everything pass through a mutex that waits for that is not great. It means for example that CheckPath() that we call all over the place gets blocked by requests. Possibly we could make the cache better by using two phase locking (first a read lock which is fast and concurrent if the cache matches, then a write lock only if we have a cache miss).

For that to work we'd need to slightly re-engineer it as the cache today covers the whole path down to the leaf, and only one such path at a time (though with all parents' info). This means that if we look up `/foo/bar/baz/quux` and then afterwards `/foo/bar/baz/asdf` this is a cache miss. In practice we should probably cache the initial .../quux call and when we get the next call we follow the cache all the way down to the leaf and there just use the cached directory info in the parent instead of updating the leaf... This might make more sense looking at the code. It's not entirely trivial.

PR #7276, opened 2021-01-13
lib/fs: Double lock and don't store leafs in realCaser (ref #7270)

imsodin:
Two-phase locking alone is pretty straight forward to add, but isn't helpful on its own:

Currently we have the directory contents as an array for every path component of the cached entry. So if we look up `/foo/bar/baz/quux` and then afterwards `/foo/bar/baz/asdf` we don't need to make an fs call, but we do need to loop through the cached childs of `/foo/bar/baz/` and then we update the cache to have `asdf` as leaf instead of `quux`.

In this PR the result of the last component are not stored, i.e. `asdf` and `quux` in the example above. That would remove the need for modification/write locks when stating lots of files in the same directory. However if another call on an unrelated path comes in between, that will change the cache and thus require acquiring write-locks. Where this does help is for fs operations in the folder root, e.g. stating `.stfolder`, which will now no longer cause any modification/lock and also will not drop the cache anymore. In the wast majority of cases where the operations are on unrelated paths (e.g. requests), we'll still modify/lock/drop cache. To fix that, some tree-ish option would help, where cache can be kept in more than a single branch. Problem there is that the size of that cache can grow fast.

### DIFF
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -370,44 +370,72 @@ func (r *defaultRealCaser) realCase(name string) (string, error) {
 		return out, nil
 	}
 
-	r.mut.Lock()
-	defer r.mut.Unlock()
+	r.mut.RLock()
+	defer r.mut.RUnlock()
 
 	node := r.root
-	for _, comp := range strings.Split(name, string(PathSeparator)) {
-		if node.dirNames == nil || node.expires.Before(time.Now()) {
-			// Haven't called DirNames yet, or the node has expired
+	pathComponents := strings.Split(name, string(PathSeparator))
+	lastIndex := len(pathComponents) - 1
 
-			var err error
-			node.dirNames, err = r.fs.DirNames(out)
-			if err != nil {
-				return "", err
-			}
-
-			node.dirNamesLower = make([]string, len(node.dirNames))
-			for i, n := range node.dirNames {
-				node.dirNamesLower[i] = UnicodeLowercase(n)
-			}
-
-			node.expires = time.Now().Add(caseCacheTimeout)
-			node.child = nil
+	var err error
+	for _, comp := range pathComponents[:lastIndex] {
+		node, out, err = r.getChildNodeRLocked(node, out, comp, false)
+		if err != nil {
+			return "", err
 		}
+	}
 
-		// If we don't already have a correct cached child, try to find it.
-		if node.child == nil || node.child.name != comp {
-			// Actually loop dirNames to search for a match.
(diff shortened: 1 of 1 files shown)


lib/fs/casefs.go
```diff
@@ -370,44 +370,72 @@ func (r *defaultRealCaser) realCase(name string) (string, error) {
...
+func (r *defaultRealCaser) getChildNodeRLocked(node *caseNode, path, childName string, leaf bool) (*caseNode, string, error) {
+	var err error
+
+	if node.dirNames == nil || node.expires.Before(time.Now()) {
+		// Haven't called DirNames yet, or the node has expired
+
+		r.mut.RUnlock()
+		r.mut.Lock()
+
+		node.dirNames, err = r.fs.DirNames(path)
```
calmh, line 405: We could break out the dirnames+lowercasing to a function and call that before acquiring the write lock, somewhat reducing the time we're holding the write lock.

lib/fs/casefs.go
```diff
@@ -370,44 +370,72 @@ func (r *defaultRealCaser) realCase(name string) (string, error) {
...
+	if node.child == nil || node.child.name != childName {
+		// Actually loop dirNames to search for a match.
+		childName, err = findCaseInsensitiveMatch(childName, node.dirNames, node.dirNamesLower)
+		if err != nil {
+			return nil, "", err
+		}
+		if !leaf {
+			r.mut.RUnlock()
+			r.mut.Lock()
+			node.child = &caseNode{name: childName}
```
calmh, line 432: Similarly we could call that function to fill the dirnames here as well (outside the lock), thus avoiding having to grab the write lock twice for every new node?

calmh (commented):
Looks simple enough, though I'm generally not super excited about calling a function under a lock that then releases that lock and re-acquires it. It's easy to look at the call site and think "this function is called under the lock, there will be no interleaving of other lockers" and be completely fooled because there may be all kinds of releases and locks by other routines while inside...

I had a redesign of this started this afternoon with a sync.Map instead, keeping a node per path key. That does have the downside that it can grow a lot though. Maybe an actual LRU/LFU cache of some kind would be better here...

imsodin:
I'd go with a combo of your ideas (map and lru&lfu) and use https://godoc.org/github.com/hashicorp/golang-lru#TwoQueueCache. Looks like a widely used and actively maintained project, so we don't need to think about size limiting ourselves. I doubt it will have a huge impact performance wise, some overhead vs saving a few calls by dropping unused entries instead of everything (if we did our own size limiting).

PR #7281, opened 2021-01-14
lib/model: Pull when folder leaves error state (fixes #7280)

### LINKED ISSUE #7280: Stopped folder not synchronizing after a rescan
Seen on Syncthing v1.12.1, Windows 10 64 bit.

When a folder is not available when starting Syncthing (e.g. on removable storage), it will be in the "Stopped" state with an appropriate error message "folder path missing".  If the physical path becomes available again, a (manual or timed) rescan will usually suffice to get it back into a working state.

However, when a remote device made changes to the folder while it was stopped, the rescan will lead to an "Out of Sync" state and the changes are not pulled from a remote device.  They are correctly listed as out of sync items.  Further rescans do not help, only pausing and unpausing the folder makes it work again.

I expect it to become "Up to Date" again after triggering a rescan.  It used to be like that, so looks like a regression in the recently reworked remote index handling.

### Steps to reproduce

1. Stop Synchthing on the local device, which mutually shares a folder with a remote device.
2. Move the shared folder path somewhere else so ST won't find it anymoe.
3. Create a new file in the folder on the remote side.
4. Start local ST again and find the folder "Stopped" and a warning notification.
5. Wait for the remote device to be connected again.
6. Move the shared folder back into its original location.
7. Press the "Rescan" button on the local folder. (Optionally dismiss the notification first.)
8. Find the folder in "Out of Sync" state, with the new file listed among unsynchronized items.
9. To fix, pause and then unpause the folder.  Watch it syncing properly and pulling the new file.

imsodin:
Schedule a pull when a folder error is cleared. Pulls are only done on demand (incoming indexes or scanned changes), which means a pull may not happen for a long time if the folder was out-of-sync while being in error state.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -981,6 +981,7 @@ func (f *folder) setError(err error) {
 		}
 	} else {
 		l.Infoln("Cleared error on folder", f.Description())
+		f.SchedulePull()
 	}
 
 	if f.FSWatcherEnabled {


acolomb:
Sorry I have limited time for testing right now, but the fix looks reasonable so I expect it to be a no-brainer.

acolomb (approved):
Tested it on a demo setup, comparing v1.12.1 verbatim and with this patch applied.  Works as expected.

acolomb:
Tested on main branch as well.

acolomb:
I assume this is a simple enough change to be picked for v1.13.0? @calmh

calmh:
1.13.0 is tagged and out in RC, so, no.

acolomb:
It's not a regression between v1.12.0, but has already existed for longer.  I only found the time to investigate and report it today, but it was definitely not this way half a year ago.  Might have regressed somewhere around v1.10 or v1.11 maybe.

calmh:
Yep. Then it's not material for a new 1.13 RC.

acolomb:
Alright. Had just hoped to enjoy the fix in about two weeks instead of six on my production machines :-)

PR #7285, opened 2021-01-16
cmd/syncthing: Provide early startup for config service (ref #7188)

calmh:
Don't hang on startup.

(We should push a new nightly after merging...)

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -45,6 +45,7 @@ import (
 	"github.com/syncthing/syncthing/lib/upgrade"
 
 	"github.com/pkg/errors"
+	"github.com/thejerf/suture/v4"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
Shouldn't the supervisor live on the app, in the ambition that we could tear down nicely?

I guess the event logger lived outside already, hence why that is not needed?

This does change what people wrapping this in C would need to do tho?

imsodin (approved):
Thanks for the catch!

It makes sense that the lifetime of objects passed to the app/`syncthing.New` is controlled outside of it. After all you could provide your own implementation of both interfaces. We have convenience methods in `lib/syncthing/util.go`, so having something there to handle the lifetime of the "default wrapper and event-logger" including starting would make sense. Nothing for this hotfix/PR though.

> This does change what people wrapping this in C would need to do tho?

They now need to use the same method they already use to run `events.Logger` on `config.Wrapper` as well.

imsodin:
Nightly is deployed and an announcement post about the manual intervention required made: https://forum.syncthing.net/t/nightly-build-requires-manual-intervention-2021-01-16/16195 (not sure how many nightly users we really have - would be interesting for usage reporting).

PR #7286, opened 2021-01-16
lib/fs: Cache all real-case results (fixes #7270)

### LINKED ISSUE #7270: Data requests block API requests that also touch the filesystem

imsodin:
As discussed in https://github.com/syncthing/syncthing/pull/7276 I implemented storing the case-resolver results in an lru&lfu cache (github.com/hashicorp/golang-lru).

A solution without golang-lru, i.e. doing locking and counting ourselves, is not much more complex. That solution is in the intermediary commit: aee7ea1717ad3e93c4e81414783ed0ec2febb741 / diff: https://github.com/syncthing/syncthing/pull/7286/files/aa1ffcaebfa12b4b6b5e6df53a3a7f7621674845

I extended the existing case-fs benchmark (on second walk we stat every path component, not just full path) and I added a new benchmark that does an unrelated lstat every 1000. item in the regular walk. To simulate stuff like api/incoming requests. To make benchmarking more relevant, I added read-locking without changing caching in a first commit, and compare against that.

The results are pretty clear for number of `DirNames` calls: Small, but significant reduction with random other calls, and no difference with or without golang-lru (and no changes for "just" walking):

lock vs with:

```
name                                          old DirNames/op     new DirNames/op     delta
WalkCaseFakeFS100k/casefs-otherOpEvery1000-8           104k ± 0%           102k ± 0%   -1.32%  (p=0.000 n=10+18)
```

lock vs without:

```
name                                          old DirNames/op     new DirNames/op     delta
WalkCaseFakeFS100k/casefs-otherOpEvery1000-8           104k ± 0%           102k ± 0%   -1.32%  (p=0.000 n=10+18)
```

With a cache is a bit slower:

without vs with

```
name                                          old ns/entry        new ns/entry        delta
WalkCaseFakeFS100k/casefs-8                           5.36k ±15%          6.82k ± 5%  +27.19%  (p=0.000 n=19+10)
WalkCaseFakeFS100k/casefs-otherOpEvery1000-8          5.38k ± 9%          6.92k ± 7%  +28.72%  (p=0.000 n=18+9)
```

but that's on a fake filesystem. On a real one I'd expect actual filesystem calls to dominate.

In the end it's not a performance-question, but about slightly less code (and control) with a new dependency versus locking and counting ourselves.

There's also the cache size constant, which is basically arbitrary. If it goes against your arbitrarity sense or you do have an actual argument for some number, please say so.


lib/fs/casefs.go
```diff
@@ -370,77 +377,87 @@ func (r *defaultRealCaser) realCase(name string) (string, error) {
...
+			num := len(dirNames)
+			node = &caseNode{
+				children:    make(map[string]struct{}, num),
+				lowerToReal: make(map[string]string, num),
+			}
+			lastLower := ""
+			for _, n := range dirNames {
+				node.children[n] = struct{}{}
+				lower := UnicodeLowercase(n)
+				if lower != lastLower {
```
calmh, line 409: something is weird here, lastLower is never set and I'm not sure what it's supposed to do.

lib/fs/casefs.go
```diff
@@ -370,77 +377,87 @@ func (r *defaultRealCaser) realCase(name string) (string, error) {
...
-	node := r.root
+	var node *caseNode
 	for _, comp := range strings.Split(name, string(PathSeparator)) {
-		if node.dirNames == nil || node.expires.Before(time.Now()) {
+		// Check cache
+		v, ok := r.cache.Get(out)
+		if ok {
+			node = v.(*caseNode)
+			if node.expires.Before(time.Now()) {
+				r.purgeRecursive(out, node.children)
```
calmh, line 387: Technically we only need to purge recursively if the casing actually changed at this level, otherwise we can let child cache entries live on their own... Not sure if that makes anything easier or harder.

calmh: We can also ignore the recursive purge entirely... Children entries could then outlive their parent, but only by at most one cache time, so an entry that lives for one second might refer to parent information up to two seconds old... I think that would be fine?

calmh (commented):
Looks reasonable. One thing with the lack of locking, we might get ten incoming block requests for the same file which is not cached, and it's quite likely those will proceed concurrently. They will all see a cache miss and do the work with listdir etc., then setting the cache item ten times, for each level.

One possible way to avoid this that might or might not be worth the machinery could be to have a wrapper around the cache. When getting it would create a new caseNode instance and set it, if one were not present in the cache. This caseNode would have an update rwlock that is taken when the caseNode need to be filled with dir info. Looking at a returned cache entry would then use double locking, first r-lock to read and then change to a w-lock if it needs to be filled. Filling it would then happen only once with the others instead waiting on that to complete... 🤷

lib/fs/casefs.go
calmh: Thinking further, this might not even be a problem since the cache is keyed on "out" (real case)? The children would become unreachable since we wouldn't find them by traversal any more, and get expelled from the cache at some point.

lib/fs/casefs.go
imsodin: I forgot to assign inside the if here. It's there because in a case-sensitive fs there can be multiple files with the same lowercase, and previously we assigned the first (plus it's a few less assignements). It doesn't matter in practice because a direct lookup will always work in a case-sensitive fs.

imsodin:
> Looks reasonable. One thing with the lack of locking, we might get ten incoming block requests for the same file which is not cached, and it's quite likely those will proceed concurrently. They will all see a cache miss and do the work with listdir etc., then setting the cache item ten times, for each level.

A problem is, we'd then be modifying a pointer without a lock, which is a race with those other routines coming afterwards. To fix that we'd have to move calling `DirNames` inside that lock.

And if we are adding any more machinery, especially our own locking, I'd prefer the variant without the external cache: Not much more complexity (if at all) and it's in our own control.

imsodin:
Now there's a wrapper around the cache that provides a `getExpireAdd` method which checks the cache, and adds a new entry if missing or expired atomically. That doesn't fully address the problem of multiple routines querying the same path concurrently, and thus potentially causing multiple calls to `DirNames` yet: For that some race-free option to check if `DirNames` still needs to be called is required. I opted to go with a buffered channel for that: The first to receive from it calls `DirNames` and then closes the channel. All routines coming later will block until the chan is closed and do nothing on receiving nothing. The same could be achieved by double locking on every node, but that's extremely slow (like an order of magnitude slower).

lib/syncthing/syncthing.go
```diff
@@ -121,6 +121,10 @@ func (a *App) Start() error {
 }
 
 func (a *App) startup() error {
+	if cfgService, ok := a.cfg.(suture.Service); ok {
```
calmh, line 124: ?

calmh (commented):
LGTM otherwise

calmh (approved):

PR #7303, opened 2021-01-25
lib/svcutil: Simplify doneService

greatroar:
With this patch, OnSupervisorDone no longer allocates.


calmh:
Wether it allocates or not really isn't relevant, but it's a minus-diff, so 🤷 👍

calmh (approved):

PR #7304, opened 2021-01-25
lib/stun: Close connection from defer, not goroutine

greatroar:
The goroutine waited for the context to be done, but that is exactly Serve's exit condition. The only behavioral change, in the absence of panics, should be that Serve now waits for the connection to be closed before it returns.

### DIFF
--- a/lib/stun/stun.go
+++ b/lib/stun/stun.go
@@ -109,9 +109,8 @@ func (s *Service) Serve(ctx context.Context) error {
 	defer func() {
 		s.setNATType(NATUnknown)
 		s.setExternalAddress(nil, "")
+		_ = s.stunConn.Close()
 	}()
-
-	util.OnDone(ctx, func() { _ = s.stunConn.Close() })
 
 	timer := time.NewTimer(time.Millisecond)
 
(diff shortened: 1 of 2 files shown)


imsodin:
I didn't check this yet: As far as I remember there are operations that do not use contexts and thus might block a long time. Closing `stunConn` when the context is cancelled unblocks those and thus speeds up shutdown. That might however no longer be the case nowadays, but needs to be checked before merging.

greatroar:
Oh, I hadn't realized that. You mean other code detects that it's using a closed connection?

calmh:
it unblocks pending Read()s and similar, yes.

imsodin:
What I was thinking of was stuff like this:

https://github.com/syncthing/syncthing/blob/70427b6844b46c6c41cf11b150e98c570c962d0b/lib/stun/stun.go#L183

https://github.com/syncthing/syncthing/blob/70427b6844b46c6c41cf11b150e98c570c962d0b/lib/stun/stun.go#L275

The first is already wrapped in `util.CallWithContext`, the second isn't. So I'd say keep your changes in this PR and additionally wrap the second in `util.CallWithContext` as well.

PR #7306, opened 2021-01-26
lib/model: Make /browse endpoint return sane objects

AudriusButkevicius:
This compliments #7145 of not having a crazy API.

We ourselves don't use this endpoint, nor could I find any use of it in:
1. synctrayzor
2. gtk
3. syncthingtray
4. syncthing-tray

So I guess the only people we could break is someone somehow using this for their pet peeves?

I guess a release note is enough to cover this.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,7 @@ require (
 	github.com/shirou/gopsutil/v3 v3.20.11
 	github.com/syncthing/notify v0.0.0-20201210100135-17de26665ddc
 	github.com/syndtr/goleveldb v1.0.1-0.20200815071216-d9e9293bd0f7
+	github.com/thejerf/suture v4.0.0+incompatible
 	github.com/thejerf/suture/v4 v4.0.0
 	github.com/urfave/cli v1.22.4
 	github.com/vitrun/qart v0.0.0-20160531060029-bf64b92db6b0
--- a/go.mod
+++ b/go.mod
@@ -37,7 +37,6 @@ require (
 	github.com/shirou/gopsutil/v3 v3.20.11
 	github.com/syncthing/notify v0.0.0-20201210100135-17de26665ddc
 	github.com/syndtr/goleveldb v1.0.1-0.20200815071216-d9e9293bd0f7
-	github.com/thejerf/suture v4.0.0+incompatible
 	github.com/thejerf/suture/v4 v4.0.0
 	github.com/urfave/cli v1.22.4
 	github.com/vitrun/qart v0.0.0-20160531060029-bf64b92db6b0
--- a/lib/api/api.go
+++ b/lib/api/api.go
@@ -246,7 +246,7 @@ func (s *service) Serve(ctx context.Context) error {
 	restMux.HandlerFunc(http.MethodGet, "/rest/db/remoteneed", s.getDBRemoteNeed)             // device folder [perpage] [page]
 	restMux.HandlerFunc(http.MethodGet, "/rest/db/localchanged", s.getDBLocalChanged)         // folder
 	restMux.HandlerFunc(http.MethodGet, "/rest/db/status", s.getDBStatus)                     // folder
-	restMux.HandlerFunc(http.MethodGet, "/rest/db/browse", s.getDBBrowse)                     // folder [prefix] [dirsOnly] [levels]
+	restMux.HandlerFunc(http.MethodGet, "/rest/db/browse", s.getDBBrowse)                     // folder [prefix] [dirsonly] [levels]
 	restMux.HandlerFunc(http.MethodGet, "/rest/folder/versions", s.getFolderVersions)         // folder
 	restMux.HandlerFunc(http.MethodGet, "/rest/folder/errors", s.getFolderErrors)             // folder
 	restMux.HandlerFunc(http.MethodGet, "/rest/folder/pullerrors", s.getFolderErrors)         // folder (deprecated)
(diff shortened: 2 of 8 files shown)


lib/model/model.go
```diff
@@ -2510,15 +2510,35 @@ func (m *model) Revert(folder string) {
 	runner.Revert()
 }
 
-func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly bool) map[string]interface{} {
+type TreeEntry struct {
+	Name     string                `json:"name"`
+	ModTime  time.Time             `json:"modTime"`
+	Size     int64                 `json:"size"`
+	Type     protocol.FileInfoType `json:"type"`
+	Children []*TreeEntry          `json:"children,omitempty"`
```
imsodin, line 2518: Why list and not a map?

lib/protocol/bep_extensions.go
```diff
@@ -394,3 +395,20 @@ func VectorHash(v Vector) []byte {
...
+	if err := json.Unmarshal(data, &s); err != nil {
+		return err
+	}
+	n, ok := FileInfoType_value[s]
+	if !ok {
+		return errors.New("invalid value: " + s)
+	}
+	*x = FileInfoType(n)
+	return nil
+}
```
imsodin, line 414: This adds some more potential to break something, e.g. a wrapper that shows out-of-sync files with a file type indication. Not very likely either though.

And this looks so generic I'd expect some `asstring` proto option - turns out doesn't exist because jsonpb does it by default.

imsodin (commented):
just nits

lib/model/model.go
AudriusButkevicius: Because maps suck for REST APIs.

Hypothetically, if you tried to consume this from let's say C, you'd get extra pain because of maps.

lib/api/api.go
```diff
@@ -710,14 +710,14 @@ func (s *service) getDBBrowse(w http.ResponseWriter, r *http.Request) {
 	qs := r.URL.Query()
 	folder := qs.Get("folder")
 	prefix := qs.Get("prefix")
-	dirsonly := qs.Get("dirsonly") != ""
+	dirsOnly := qs.Get("dirsOnly") != ""
```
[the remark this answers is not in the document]
AudriusButkevicius: Fair.

lib/model/model.go
```diff
@@ -2510,15 +2510,35 @@ func (m *model) Revert(folder string) {
...
 	m.fmut.RLock()
 	files, ok := m.folderFiles[folder]
 	m.fmut.RUnlock()
 	if !ok {
 		return nil
 	}
 
-	output := make(map[string]interface{})
+	root := &TreeEntry{
+		Name:     "ROOT",
```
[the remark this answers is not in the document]
AudriusButkevicius: No reason, I can remove it.

I had some debug code that printed out the parent e.g, `looking for "foo" in parent "ROOT"`  which made it more readable.

lib/model/model.go
```diff
@@ -2538,42 +2558,39 @@ func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsonly
...
-				directory, ok := last[path]
-				if !ok {
-					newdir := make(map[string]interface{})
-					last[path] = newdir
-					last = newdir
-				} else {
-					last = directory.(map[string]interface{})
+				child := findByName(parent.Children, path)
+				if child == nil {
+					panic(fmt.Errorf("could not find child '%s' for path '%s' in parent '%s'", path, f.Name, parent.Name))
```
imsodin, line 2573: If this really should be fatal, please use `model.fatal`. However I am not sure this should be fatal: In theory a remote might send us index updates in an ugly order, where a child comes in a batch before the parent. We could just skip it.

AudriusButkevicius: So this is used in the http handler, so it would just return a 500 etc. But you're right I guess (value, err) is a better approach here.

go.mod
```diff
@@ -37,6 +37,7 @@ require (
 	github.com/shirou/gopsutil/v3 v3.20.11
 	github.com/syncthing/notify v0.0.0-20201210100135-17de26665ddc
 	github.com/syndtr/goleveldb v1.0.1-0.20200815071216-d9e9293bd0f7
+	github.com/thejerf/suture v4.0.0+incompatible
```
imsodin, line 40: That shouldn't be here. Also in go.sum.

test/http_test.go
```diff
@@ -263,6 +263,6 @@ func BenchmarkAPI_db_status(b *testing.B) {
 	benchmarkURL(b, "/rest/db/status?folder=default")
 }
 
-func BenchmarkAPI_db_browse_dirsonly(b *testing.B) {
-	benchmarkURL(b, "/rest/db/browse?folder=default&dirsonly=true")
+func BenchmarkAPI_db_browse_dirsOnly(b *testing.B) {
+	benchmarkURL(b, "/rest/db/browse?folder=default&dirsOnly=true")
```
imsodin, line 267: `&dirsOnly` needs to be `&dirsonly` again as well.

imsodin (commented):
Small oversights to be fixed. And one entirely optional suggestion. Otherwise lgtm.

go.mod
AudriusButkevicius: My ide keeps adding it, not sure why it can't deal with the suture upgrade. Doesn't like the /v4 import. I think it should be suture, and go.mod just makes sure that resolves to v4m

imsodin: Since modules the import path must include the major version suffix (e.g. it's possible to use both v3 and v4 of the same library in a module). Maybe a leftover uncommitted file that still imports "github.com/thejerf/suture"?

lib/model/model.go
```diff
@@ -2510,15 +2510,34 @@ func (m *model) Revert(folder string) {
...
+	return nil
+}
+
+func (m *model) GlobalDirectoryTree(folder, prefix string, levels int, dirsOnly bool) ([]*TreeEntry, error) {
 	m.fmut.RLock()
 	files, ok := m.folderFiles[folder]
 	m.fmut.RUnlock()
 	if !ok {
-		return nil
+		return nil, errFolderMissing
```
imsodin, line 2535: Entirely optional suggestion:  
That results in a 500 on the api, while a 404 would be more appropriate (we do that on other endpoints if we don't find a folder/path). We could export `errFolderMissing` and check the error in the api to decide on 404 or 500.

imsodin (approved):

PR #7308, opened 2021-01-27
lib/stun: Inline util.OnDone, comment on its purpose

greatroar:
This is an alternative to #7304: it inlines util.OnDone at its only call site and replaces its comment by one that explains why the separate goroutine is necessary.

I didn't add the second CallWithContext as I didn't find the resulting code much clearer than this version. In fact, I'm wondering if the CallWithContext around Discover cannot be removed instead?

Closes #7304.

### DIFF
--- a/lib/stun/stun.go
+++ b/lib/stun/stun.go
@@ -111,7 +111,12 @@ func (s *Service) Serve(ctx context.Context) error {
 		s.setExternalAddress(nil, "")
 	}()
 
-	util.OnDone(ctx, func() { _ = s.stunConn.Close() })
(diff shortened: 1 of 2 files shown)


calmh (approved):

calmh:
This seems like a change that costs more in coding and review time than it's worth, and now you have a merge conflict to resolve... But have at it. :)

greatroar:
Merge conflict resolved. Feel free to close both PRs if you don't think either is worth the effort. Pushing context support to the go-stun lib would be the only truly clean solution, I guess...

calmh (approved):

PR #7310, opened 2021-01-28
lib/scanner: Allocate structure for final partial block

greatroar:
scanner.Blocks promises, in a comment, to allocate space "once and for all" if it knows the file size. It only did so, however, when the block size divided the file size. The pre-allocation now works as intended for files of all sizes.

Benchmark results on Linux/amd64, using updated benchmark for old and new:

```
HashFile-8    88.6ms ± 1%    88.3ms ± 1%   -0.33%  (p=0.046 n=19+19)

HashFile-8   201MB/s ± 1%   202MB/s ± 1%   +0.33%  (p=0.044 n=19+19)

HashFile-8    59.4kB ± 0%    46.1kB ± 0%  -22.47%  (p=0.000 n=14+20)

HashFile-8      29.0 ± 0%      27.0 ± 0%   -6.90%  (p=0.000 n=20+20)
```

### DIFF
--- a/lib/scanner/blocks.go
+++ b/lib/scanner/blocks.go
@@ -30,7 +30,7 @@ func Blocks(ctx context.Context, r io.Reader, blocksize int, sizehint int64, cou
 	}
 
 	hf := sha256.New()
-	hashLength := hf.Size()
+	const hashLength = sha256.Size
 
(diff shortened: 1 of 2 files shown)


greatroar:
The test failure on Windows seems to be unrelated.

calmh (approved):

PR #7315, opened 2021-01-31
lib/model: Optimize rename detection in scanner

imsodin:
A user reported very slow scans and looking at a crazy volume of FS debug logs revealed that the rename detection is inefficient if a lot of data is newly scanned that contains a lot of duplicates. Not a very common case, but still something that's simple to improve on by caching not only files that were already detected as deleted/renamed, but also files that were already checked with `IsDeleted` during rename detection.

I also added a check for the filename of the file for which we want to find duplicates and "delete" the cache map afterwards, as it isn't used any more but the function might live on for quite some time.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -549,7 +549,7 @@ func (f *folder) scanSubdirs(subDirs []string) error {
 	}
 
 	f.clearScanErrors(subDirs)
-	alreadyUsed := make(map[string]struct{})
+	alreadyUsedOrExisting := make(map[string]struct{})
 	for res := range fchan {
 		if res.Err != nil {
 			f.newScanError(res.Path, res.Err)
@@ -571,7 +571,7 @@ func (f *folder) scanSubdirs(subDirs []string) error {
 		switch f.Type {
 		case config.FolderTypeReceiveOnly, config.FolderTypeReceiveEncrypted:
 		default:
-			if nf, ok := f.findRename(snap, res.File, alreadyUsed); ok {
+			if nf, ok := f.findRename(snap, res.File, alreadyUsedOrExisting); ok {
 				batchAppend(nf, snap)
 				changes++
 			}
@@ -581,6 +581,10 @@ func (f *folder) scanSubdirs(subDirs []string) error {
 	if err := batch.flush(); err != nil {
 		return err
 	}
+
+	// Might have grown large, isn't used anymore and this function may keep
+	// running for some time.
+	alreadyUsedOrExisting = nil
 
 	if len(subDirs) == 0 {
 		// If we have no specific subdirectories to traverse, set it to one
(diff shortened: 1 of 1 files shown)


calmh (approved):

lib/model/folder.go
```diff
@@ -582,6 +582,10 @@ func (f *folder) scanSubdirs(subDirs []string) error {
 		return err
 	}
 
+	// Might have grown large, isn't used anymore and this function may keep
+	// running for some time.
+	alreadyUsedOrExisting = nil
```
greatroar, line 587: I know I'm late to the party, but this doesn't actually do anything, as you can verify using

```
go build -gcflags=-S |& sha256sum
```

with this line active and commented out.

calmh: I was actually hoping that the computer would already notice that it's not used and consider it eligible for garbage collection, so hopefully that's actually the case. 👍

greatroar: That's not what happens, though. The compiler detects the variable no longer being useful, which allows it to elide the nil assignment, but the function keeps the reference to the map. If you look at the generated assembly (for amd64), you'll find the map construction as

```
CALL    runtime.makemap_small(SB)
MOVQ    (SP), AX
MOVQ    AX, "".alreadyUsedOrExisting+1352(SP)
```

The stack address 1352(SP) is referenced only once more, when alreadyUsedOrExisting is passed to f.findRename. It's never zeroed out, so the map will survive GC until the function returns.

calmh: I wonder if that's actually true though. The existence of things like runtime.KeepAlive to tweak the liveness detection makes it seem to me like the object should be eligible for collection as soon as it's not live any more. If it is the case that it's still live it seems like an issue that should be reported somewhere as the liveness detection then only gets in the way of getting garbage collected.

calmh: At least a manual `runtime.GC()` call will collect things that are not in used after that point in the function, I assume a normal automatic GC will as well.

greatroar: runtime.KeepAlive solves an unrelated problem, where object liveness must be artificially extended because of a finalizer or system call.

And how would manual runtime.GC know that it can collect this map, if the stack is keeping it alive?

imsodin: Not to take away from this interesting question, however from a practical standpoint I was anyway thinking that `folder.scanSubdirs` is a way to long function: I'll split it into logical pieces (detecting new/changed items, detecting deleted/ignored items - maybe more) and thus create a nice scope for this variable.

calmh: It's not unrelated, runtime.KeepAlive is needed because the compiler does liveness analysis. The stack doesn't keep it alive as you assume, it can be garbage collected while still in scope if it's not used later in the scope. You can see this happening by setting a finalizer, but the finalizer doesn't *cause* it to happen.

greatroar: The compiler can never do an *exact* liveness analysis, so the program may have to keep some objects alive longer than necessary. The compiler also assumes functions are short, so it doesn't inject code to delete references mid-way through. I'm not sure if you're confusing stack and scope or you're assuming I do, but the stack is the primary mechanism that keeps objects alive. That's where the GC starts its scans.

KeepAlive solves the problem that the runtime *may* collect an object too early (when it is still "referenced" by a non-pointer), not that it *will* collect objects as soon as the last reference can be statically determined to be gone.

PR #7316, opened 2021-01-31
lib/model: Correct pull progress for small files (fixes #7263)

### LINKED ISSUE #7263: GUI displays Syncing (100%, 0 B) when syncing byte-sized files
1. Create a folder on Device A with byte-sized tiny files. In my test, I used 10,000 of < 10-byte files.
3. While syncing, Device B displays `Syncing (100%, 0 B)`, even though the files are not `0 B`, and also not at `100%`.

image

Also, once the synchronisation has finished, the GUI on Device A is stuck in the following state, even though the actual folders are in sync. The glitch/error does go away after pausing and unpausing the folder on both devices.

image

Not sure how useful they are, but I also have logs from both devices with `STTRACE=db,model` enabled.

[DeviceA-syncthing.log](https://github.com/syncthing/syncthing/files/5793496/DeviceA-syncthing.log)
[DeviceA-syncthing.0.log](https://github.com/syncthing/syncthing/files/5793497/DeviceA-syncthing.0.log)
[DeviceA-syncthing.1.log](https://github.com/syncthing/syncthing/files/5793498/DeviceA-syncthing.1.log)

[DeviceB-syncthing.log](https://github.com/syncthing/syncthing/files/5793499/DeviceB-syncthing.log)
[DeviceB-syncthing.0.log](https://github.com/syncthing/syncthing/files/5793500/DeviceB-syncthing.0.log)
[DeviceB-syncthing.1.log](https://github.com/syncthing/syncthing/files/5793502/DeviceB-syncthing.1.log)
[DeviceB-syncthing.2.log](https://github.com/syncthing/syncthing/files/5793503/DeviceB-syncthing.2.log)

Tested with Syncthing v1.12.1.

imsodin:
Now we return a progress of `blockSize / 2` for files with a single block. That means that the reported progress is potentially larger than the total amount of bytes to sync in total. That leads to the 100% progress reported from the start reported in #7263.

Because I felt like it I adjusted the bytes estimate to be based on the probability that the small block is part of the given number of blocks, based on the total number of blocks in the file. That returns the actual file size for single block files, which fixes the above problem (which could obviously have been achieved in a much simpler way). For larger files the estimate will be a bit "better", probably entirely unnoticeably better.

### DIFF
--- a/lib/model/sharedpullerstate.go
+++ b/lib/model/sharedpullerstate.go
@@ -376,6 +376,7 @@ func (s *sharedPullerState) Progress() *pullerProgress {
 	defer s.mut.RUnlock()
 	total := s.reused + s.copyTotal + s.pullTotal
 	done := total - s.copyNeeded - s.pullNeeded
+	file := len(s.file.Blocks)
 	return &pullerProgress{
 		Total:               total,
 		Reused:              s.reused,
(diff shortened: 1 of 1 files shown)


calmh (approved):
Neat

PR #7319, opened 2021-02-01
lib/protocol: Hide repeated data blocks in a given file

calmh:
Adds the block offset as additional data into the AES-SIV encryption of the block hash, thus making it not possible to correlate blocks at different offsets within a file. (These are never decrypted so there is no "other side" to this change.)

### DIFF
--- a/lib/protocol/encryption.go
+++ b/lib/protocol/encryption.go
@@ -11,7 +11,9 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"encoding/base32"
+	"encoding/binary"
 	"errors"
+	"fmt"
 	"io"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


imsodin:
> (These are never decrypted so there is no "other side" to this change.)

Why actually? We could send and then decrypt the hash in requests from untrusted to trusted, thus being able to check the data. Maybe that doesn't happen to save a bit of overhead?

calmh:
In theory we probably could, yes. Maybe we already do send it and just lack the verification, I'm not sure.

calmh:
I think we do, let me add that.

calmh:
There, now we actually decrypt the hash as well.

imsodin (commented):
In the `copierRoutine` in the puller we already skip searching by weak hash for receive-encrypted folders. ~~Now we can also skip searching by hash.~~ We can't - we still want to reuse unchanged blocks.

imsodin (approved):

PR #7325, opened 2021-02-03
gui: do not touch .stignore when not needed (fixes #7284)

### LINKED ISSUE #7284: Do not touch .stignore on saving folder config when ignore patterns have not changed
Right now, the modtime of `.stignore` gets changed every time we click on the "Save" button in the "Edit Folder" window. Even if there are no changes to the actual patterns, the modtime still gets updated. Is there any specific reason for this behaviour?

If not, then I would argue that when the patterns have not changed, the modtime should remain unchanged too.

Tested with Syncthing v1.12.1.

hiqua:
### Purpose

Do not touch .stignore when not needed (fixes #7284)

### Testing

* saved without changes (mtime unchanged)
* saved with real changes (mtime changed)
* saved with whitespace changes (mtime unchanged)

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -2019,7 +2019,11 @@ angular.module('syncthing.core')
             $scope.folders[folderCfg.id] = folderCfg;
             $scope.config.folders = folderList($scope.folders);
 
-            if (ignoresLoaded && $scope.editingExisting && ignores !== folderCfg.ignores) {
+            function arrayEquals(a, b) {
+              return a.length === b.length && a.every(function(v, i) { return v === b[i] });
+            }
+
+            if (ignoresLoaded && $scope.editingExisting && !arrayEquals(ignores, folderCfg.ignores)) {
                 saveIgnores(ignores);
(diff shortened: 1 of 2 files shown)


calmh (commented):
Sounds good. We don't yet have any arrow-syntax functions in here, which may or may not be unnecessarily conservative, but I'd suggest old style function syntax for the time being.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -2019,7 +2019,11 @@ angular.module('syncthing.core')
             $scope.folders[folderCfg.id] = folderCfg;
             $scope.config.folders = folderList($scope.folders);
 
-            if (ignoresLoaded && $scope.editingExisting && ignores !== folderCfg.ignores) {
+            function arrayEquals(a, b) {
+              return a.length === b.length && a.every((v, i) => v === b[i]);
```
calmh, line 2023: dat arrow though

hiqua: took that one in the knee

calmh (approved):

PR #7326, opened 2021-02-03
lib/ignore: Store cache timestamps as Unix ns counts

greatroar:
This halves the size of cache entries on amd64 (32 bytes -> 16 bytes). A quick simulation suggest the total memory footprint should get about a third smaller: I put strconv.Itoa(j) for j = 0 through 10000 in a cache and saw the allocation rate go down from 1.91 to 1.31 MiB.

### DIFF
--- a/lib/ignore/cache.go
+++ b/lib/ignore/cache.go
@@ -21,7 +21,7 @@ type cache struct {
 
 type cacheEntry struct {
 	result Result
-	access time.Time
+	access int64 // Unix nanosecond count. Sufficient until the year 2262.
 }
 
 func newCache(patterns []Pattern) *cache {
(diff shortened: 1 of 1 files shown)


calmh:
Literally noone uses ignore caching; the better cleanup might be to remove it entirely.

lib/ignore/cache.go
```diff
@@ -21,7 +21,7 @@ type cache struct {
 
 type cacheEntry struct {
 	result Result
-	access time.Time
+	access int64 // Unix nanosecond count. Sufficient until the year 2262.
```
calmh, line 24: changing the order of these should also reduce the size of the struct a fair bit I think

greatroar: It goes in a map, where it's rounded up to the alignment size, so it would only matter on 32-bit platforms.

calmh: Are you sure? I thought map values were stored packed.

greatroar:
I'll see if I can excise the cache entirely.

lib/ignore/cache.go
greatroar: I don't have a reference to the runtime code handy, but some quick benchmarking suggests that a map with values of type `struct { x uint64; y byte }` is as big as one of `struct { x, y uint64 }`.

AudriusButkevicius (approved):
Lgtm

PR #7327, opened 2021-02-03
all: Automatic/disabled folder-config when receive-encrypted

imsodin:
> Another wish: it would be great, when a peer that is intended to be receiveEncrypted receives a membership notification, that he was warned to switch to Advanced tab to set himself “Receive Encrypted” before “Save” as we won’t give him the password.

https://forum.syncthing.net/t/early-testing-for-untrusted-encrypted-devices/15873/17

When rejecting a folder/adding a pending folder, record if the remote wants to send us encrypted data or not. Then act on that in the GUI by automatically setting the folder type to receive-encrypted.

It is possible to share a folder between two devices where both have the PW and send data encrypted between them. In this case the folder type would be wrong. However I don't consider that problematic: 1. While possible that setup is pointless. 2. The user will need to enter the password, and thus notice the incorrect folder type as the PW entry is not possible on recv-enc.

It also disables ignores, versioning and ignore permissions in the folder edit modal. And ensures that `IgnorePerms == true` in Go and JS.


calmh:
>  go: github.com/syncthing/syncthing/lib/config: package github.com/golang/protobuf/ptypes/timestamp imported from implicitly required module; to add missing requirements, run:
> 00:29:19 go get github.com/golang/protobuf/ptypes/timestamp@v1.4.3

I think a go mod tidy should fix this, the requirements have changed in Go 1.16 so that it never modifies go.mod implicitly during build.

gui/default/untrusted/syncthing/core/syncthingController.js
```diff
@@ -1911,36 +1912,48 @@ angular.module('syncthing.core')
             }
         };
 
+        function addFolderInit(recvEnc) {
+            $scope.editingExisting = false;
+            $scope.currentFolder = angular.copy($scope.folderDefaults);
+            if (recvEnc) {
+                $scope.currentFolder.type = "receiveencrypted";
+                $scope.currentFolder.ignorePerms = true;
```
acolomb, line 1920: Isn't this something the backend should enforce if permission synchronization is not useful for encrypted folders?  Also applies to versioning settings.

I think such things should not be the GUI's responsibility.

gui/default/untrusted/syncthing/core/syncthingController.js
```diff
@@ -1911,36 +1912,48 @@ angular.module('syncthing.core')
...
-                $scope.ignores.error = null;
-                $scope.ignores.disabled = false;
                 $scope.editFolderModal();
             });
         };
 
         $scope.addFolderAndShare = function (folder, folderLabel, device) {
-            $scope.editingExisting = false;
-            $scope.currentFolder = angular.copy($scope.folderDefaults);
+            var pendingFolder = $scope.pendingFolders[folder];
```
acolomb, line 1939: If the addFolderAndShare function needs access to the pendingFolder's details, we could pass the corresponding `offeredBy[k]` object instead of the folderLabel parameter.

gui/default/untrusted/syncthing/folder/editFolderModalView.html
```diff
@@ -16,6 +16,7 @@
             <input name="folderLabel" id="folderLabel" class="form-control" type="text" ng-model="currentFolder.label" value="{{currentFolder.label}}" />
             <p class="help-block">
               <span translate ng-if="folderEditor.folderLabel.$valid || folderEditor.folderLabel.$pristine">Optional descriptive label for the folder. Can be different on each device.</span>
+              {{currentFolder.type}} {{currentFolder.type === 'receiveencrypted'}}
```
acolomb, line 19: What does this do?  Debugging leftover maybe?

lib/config/folderconfiguration.go
```diff
@@ -226,6 +226,12 @@ func (f *FolderConfiguration) prepare(myID protocol.DeviceID, existingDevices ma
 	} else if f.MaxConcurrentWrites > maxConcurrentWritesLimit {
 		f.MaxConcurrentWrites = maxConcurrentWritesLimit
 	}
+
+	if f.Type == FolderTypeReceiveEncrypted {
+		f.IgnorePerms = true
+		f.Versioning.Type = ""
+		f.Versioning.Params = nil
+	}
```
acolomb, line 234: I see, it is in fact enforced by the backend.  Maybe we should have an API endpoint to receive the default folder config with "encrypted" as a parameter?

acolomb (commented):
Just looked at the code, no actual testing.  See suggestions below.

acolomb:
During `go run build.go proto`, the following dependency is added to `go.mod`:

	github.com/golang/protobuf v1.4.3

Should this be included in the PR?  It's one of three references to protobuf in that file, with different versions.

lib/config/folderconfiguration.go
imsodin: Sounds reasonable,. Any kind of defaults aren't available though as of yet: https://github.com/syncthing/syncthing/pull/7131

gui/default/untrusted/syncthing/core/syncthingController.js
imsodin: It is enforced in the backend. The thing is this is a new folder config created in the GUI - it doesn't come from the backend. And the settings must already be correct when presented to the user. See also my answer to your other comment: https://github.com/syncthing/syncthing/pull/7327#discussion_r570083050

lib/config/folderconfiguration.go
acolomb: So this can go into a follow-up PR once #7131 is done.  Should we open an issue as reminder?

acolomb: In addition, does the "Watch for changes" configuration make sense for receive-encrypted folders?

acolomb (approved):
Looks good, although I don't have any encrypted folders set up to actually see it in action.

lib/config/folderconfiguration.go
calmh: In my setup I would disable watching and set the rescan interval to 86400. Maybe it makes sense to offer that here as well?

imsodin: I'll not disable it entire, but use those as defaults.

lib/config/folderconfiguration.go
```diff
@@ -226,6 +226,12 @@ func (f *FolderConfiguration) prepare(myID protocol.DeviceID, existingDevices ma
 	} else if f.MaxConcurrentWrites > maxConcurrentWritesLimit {
 		f.MaxConcurrentWrites = maxConcurrentWritesLimit
 	}
+
+	if f.Type == FolderTypeReceiveEncrypted {
+		f.IgnorePerms = true
+		f.Versioning.Type = ""
+		f.Versioning.Params = nil
```
imsodin, line 233: Versioning I am not as certain any more. It should just work in principle when it comes to versioning items in the code. Recovering items will be somewhat painful to users: All the UI recovering is likely useless, and on disk won't be trivial either. At the same time it's possible (once the de-cryption tooling is there, and that will happen). And it's quite likely, that encrypted nodes will be used as central hubs, which makes them suitable for versioning (all changes there will be remote changes).

What are your thoughts on disabling versioning or not for receive-encrypted folders?

lib/config/folderconfiguration.go
imsodin: That means the UI also needs to know the "default default" (without user configuration), depending on some folder parameters. Now there's "default defaults" depending on recv-enc, but already before there was e.g. the rescan interval "default default" depending on watching being enabled. That's doable but I am at a loss regarding naming right now: "DefaultFolder` or `/rest/config/defaults/folder` is the name for user-configured defaults. What would be a good name to get a "default default"?    
`/rest/config/defaults/folder/default` feels disingenuous and obtuse. `/rest/config/defaults/unconfigured/folder` isn't quite nice either. Ideas?

calmh: Holy shit. Just hardcode something? :D

lib/config/folderconfiguration.go
calmh: I think we should default to it off, but not lock it down any further at this point.

calmh:
Right. I'm not totally in the loop on the observed stuff nor the GUI really at the moment, fire at will, I bless this PR from a distance and in principle 🙏

imsodin:
> Also, cookie for managing to flake out TestConfigChanges on Linux as well!

Ha! And an attempt at reproducing is brewing in some branch.

@acolomb Would you mind looking at the (additional) pending changes I made? I tested them "in practice", so it's not entirely broken - still more eyes is good.

lib/config/folderconfiguration.go
acolomb: @imsodin That wasn't easy to follow, but I think I see what you're trying to achieve.  My thought was simpler:  When we have (configurable or even constant) defaults available from an API endpoint, that request gets an optional parameter for the folder type.  If given, the returned defaults are adjusted with anything that would be overruled by the backend anyway because of the folder type.  So if e.g. `ignorePerms` is a must (or strongly advised) for encrypted folders, then the GUI does not need to handle it specially, but just preselect the choice as given in the API response.  Of course the user could still change it (unless we really hide the option for encrypted folders) but the backend would revert to the required choice anyway.

So no "default defaults" for the GUI to compare against, but only the user's configured defaults, with some precautions applied based on the given folder type hint.  Does that make sense?

lib/config/folderconfiguration.go
acolomb: Sounds reasonable, but I can't oversee the consequences because of limited experience with how versioning really works.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -271,19 +272,20 @@ angular.module('syncthing.core')
 
             if (arg.data.added) {
                 arg.data.added.forEach(function (rejected) {
-                    var offeringDevice = {
+                    var observedFolder = {
```
acolomb, line 275: I can see why you might want to rename this, matching the DB's `ObservedFolder` type.  But for consistency in the GUI code, I would keep the `offeringDevice` naming, because the rejection is always specific to the device, and `observedFolder` next to `pendingFolder` is more confusing IMO.

Just a suggestion though, I'm okay with the rename if it's clearer to you this way.

lib/config/folderconfiguration.go
imsodin: That's endless: You could also have a parameter for fs watcher enabled or not, as that also influences the scan interval. And there might be more. Unless there's an elegant solution to cover it all, I'd rather keep it how it is now (some values in the UI). And even if there's an elegant solution: Unless it's super simple/small - not in this PR.

acolomb (approved):
> @acolomb Would you mind looking at the (additional) pending changes I made? I tested them "in practice", so it's not entirely broken - still more eyes is good.

Except for the IMO unnecessary variable rename, looks like it won't break anything.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1811,12 +1813,18 @@ angular.module('syncthing.core')
             $scope.currentFolder.path = pathJoin($scope.config.defaults.folder.path, newvalue);
         });
 
-        $scope.fsWatcherToggled = function () {
+        $scope.setFSWatcherIntervalDefault = function () {
+            var defaultRescanIntervals = [60, 3600];
+            if (defaultRescanIntervals .indexOf($scope.currentFolder.rescanIntervalS) >= 0) {
```
acolomb, line 1818: Extra space.

Didn't you mean to return when the value is NOT in the array?

gui/default/syncthing/core/syncthingController.js
imsodin: You're right about consistency, and it's unrelated to this PR - I'll revert it.

The name was very confusing to me though: The object contains no device info, but e.g. `label`, which doesn't label a device, but a folder. To me  `observedFolder` is better, or maybe `deviceFolder` would also make sense.

lib/config/folderconfiguration.go
acolomb: > Unless it's super simple/small - not in this PR.

Definitely agree.

gui/default/syncthing/core/syncthingController.js
acolomb: I totally understand the confusion.  `offerDetails` maybe?  That would make line 288 readable again.

gui/default/untrusted/syncthing/core/syncthingController.js
```diff
@@ -1828,12 +1830,31 @@ angular.module('syncthing.core')
             $scope.currentFolder.path = pathJoin($scope.config.defaults.folder.path, newvalue);
         });
 
-        $scope.fsWatcherToggled = function () {
+        $scope.setFSWatcherIntervalSDefault = function () {
+            var defaultRescanIntervals = [60, 3600, 3600*24];
+            if (defaultRescanIntervals .indexOf($scope.currentFolder.rescanIntervalS) === -1) {
```
acolomb, line 1835: Extra space here as well.

gui/default/untrusted/syncthing/core/syncthingController.js
```diff
@@ -1828,12 +1830,31 @@ angular.module('syncthing.core')
             $scope.currentFolder.path = pathJoin($scope.config.defaults.folder.path, newvalue);
         });
 
-        $scope.fsWatcherToggled = function () {
+        $scope.setFSWatcherIntervalSDefault = function () {
```
acolomb, line 1833: The function name differs between untrusted and default.  The "S" needs to go here.

gui/default/untrusted/syncthing/core/syncthingController.js
```diff
@@ -1993,13 +2015,20 @@ angular.module('syncthing.core')
             });
         };
 
-        $scope.addFolderAndShare = function (folderID, folderLabel, device) {
+        $scope.addFolderAndShare = function (folderID, pendingFolder, device) {
```
acolomb, line 2018: I suggest porting the argument change to the default GUI as well, so the diff is again only encryption related.

acolomb (changes requested):
Some more diff between default and untrusted, with one actual bug.

acolomb (approved):

PR #7328, opened 2021-02-04
all: Remove ignore cache

greatroar:
This removes ignore result caching, as suggested by @calmh in #7326.

The config option is still recognized, but ignored. Manpage syncthing-config.5 needs an update.


AudriusButkevicius:
There is no reason to keep the config variable if it does nothing.

imsodin:
Do we really want to remove it? 2.2% of the reporting users (~1.6k) use it. Probably a lot of them do so randomly, but there might also be "that special setup" that does so conciously with tons of ignore patterns and will see a significant performance impact.

If we do remove it, this needs a an issue to be closed and included in the release notes.

AudriusButkevicius:
Idk, it doesn't cost is to have it, as in, the code does not sprawl into different areas and cause maintenance burden, so not exactly sure of the rationale behind @calmh's suggestion.

calmh:
I don't think I had a rationale beyond "do we really need to tweak features almost nobody uses", and if we do, then perhaps evaluating the need for it first. I haven't seen a profile in a long time that shows ignore matching to be a bottle neck.

greatroar:
I looked at the benchmark in the tests only so far, and that suggests a factor 10 improvement from caching, from 960ns down to 90ns on my machine. 960ns is still terribly fast.

However, when I extend the benchmark to match 20000 filenames against 10000 distinct patterns in each iteration, I get the following timings:

```
name           time/op
Match-8         11.0s ± 0%
MatchCached-8  2.60ms ± 0%
```

... that's a factor 4e6 improvement. So I guess the question is, do people write/generate such big ignore files? The experience in restic is that [some people need this](restic/restic#2694), but Syncthing is not restic.

imsodin:
I wouldn't rule out that there is such a use case, and I agree with Audrius statement that there's no real need to get rid of it. I'd keep it.

greatroar:
Looks like merging the patterns into one with the glob library's {foo,bar} syntax doesn't alleviate this much.

AudriusButkevicius:
Ok, so I guess we agree that we can keep this as a trade off @calmh ?

In which case the original pr halving the memory seems to make sense?

calmh:
Sure

PR #7330, opened 2021-02-04
cmd/syncthing: Refactor command line parsing

calmh:
This change does the following:

- Changes to use the kong command line parser instead of the flag
  package from the standard library. Kong is awesome; I'm deliberately
  using only a fraction of the features here in order to make the change
  smaller to begin with. Feel free to read up on
  https://github.com/alecthomas/kong and then come back here, I'll wait.

- Makes the current mode of operation a subcommand of syncthing,
  `syncthing serve`. The purpose of this is to enable other syncthing
  subcommands in the near future.

- Attempts to clean up and combine environment variables with command
  line flags to a greater extent. This means creating a number of new
  command line flags that were previously environment variables only,
  and passing them explicitly to where they're needed instead of by
  environment magic.  There's probably more that could be done here, but
  I didn't want to go overboard.

This is a backwards compatible change, with a single exception. That is,
any valid command line from before this change should be a valid command
line now as well, and the same environment variables still do the same
things.

The exception is `syncthing --help` which, being a top level help
command, now lists the available subcommands instead of the previous
output. The new way to get the regular help output is `syncthing serve
--help`.

Kong, like most other modern command line parsers, doesn't understand
single-dashed long options (instead it has short options, which we
currently don't have any). To facilitate the transition we handle the
case where no subcommand is given and translate invalid -options to new
style --options. This mean `syncthing -home foo` is accepted and
equivalent to `syncthing serve --home foo`. Note that you *can't* say
`syncthing serve -home foo` -- once you've opted in to saying an
explicit subcommand it's expected that you follow the new syntax and use
proper long options.

For reference, this is the new help output:

```
% syncthing --help
Usage: syncthing <command>

Flags:
  -h, --help    Show context-sensitive help.

Commands:
  serve
    Run Syncthing

Run "syncthing <command> --help" for more information on a command.
```

and then:

```
% syncthing serve --help
Usage: syncthing serve

Run Syncthing

Flags:
  -h, --help                                      Show context-sensitive help.

      --allow-newer-config                        Allow loading newer than current config version
      --audit                                     Write events to audit file
      --auditfile=PATH                            Specify audit file (use "-" for stdout, "--" for stderr)
      --browser-only                              Open GUI in browser
      --conf=PATH                                 Set configuration directory (config and keys)
      --data=PATH                                 Set data directory (database and logs)
      --device-id                                 Show the device ID
      --generate=PATH                             Generate key and config in specified dir, then exit
      --gui-address=URL                           Override GUI address (e.g. "http://192.0.2.42:8443")
      --gui-apikey=API-KEY                        Override GUI API key
      --hide-console                              Hide console window (Windows only)
      --home=PATH                                 Set configuration and data directory
      --log-file=PATH                             Log file name (see below)
      --log-flags=BITS                            Select information in log line prefix (see below)
      --log-max-old-files=N                       Number of old files to keep (zero to keep only current)
      --log-max-size=BYTES                        Maximum size of any file (zero to disable log rotation)
      --no-browser                                Do not start browser
      --no-restart                                Do not restart Syncthing when exiting due to API/GUI command, upgrade, or crash ($STNORESTART)
      --no-default-folder                         Don't create the "default" folder on first startup ($STNODEFAULTFOLDER)
      --no-upgrade                                Disable automatic upgrades ($STNOUPGRADE)
      --paths                                     Show configuration paths
      --paused                                    Start with all devices and folders paused
      --unpaused                                  Start with all devices and folders unpaused
      --upgrade                                   Perform upgrade
      --upgrade-check                             Check for available upgrade
      --upgrade-to=URL                            Force upgrade directly from specified URL
      --verbose                                   Print verbose log output
      --version                                   Show version
      --debug-block-profiler                      Write block profiles to block-$pid-$timestamp.pprof every 20 seconds ($STBLOCKPROFILE)
      --debug-cpu-profile                         Write a CPU profile to cpu-$pid.pprof on exit ($CPUPROFILE)
      --debug-db-indirect-gc-interval=DURATION    Database indirection GC interval ($STGCINDIRECTEVERY)
      --debug-db-recheck-interval=DURATION        Database metadata recalculation interval ($STRECHECKDBEVERY)
      --debug-deadlock-timeout=SECONDS            Used for debugging internal deadlocks ($STDEADLOCKTIMEOUT)
      --debug-gui-assets-dir=PATH                 Directory to load GUI assets from ($STGUIASSETS)
      --debug-heap-profile                        Write heap profiles to heap-$pid-$timestamp.pprof each time heap usage increases ($STHEAPPROFILE)
      --debug-network-profiler-url=URL            Network profiler listen URL ($STPROFILER)
      --debug-perf-stats                          Write running performance statistics to perf-$pid.csv (Unix only) ($STPERFSTATS)
      --reset-database                            Reset the database, forcing a full rescan and resync
      --reset-deltas                              Reset delta index IDs, forcing a full index exchange

The -logflags value is a sum of the following:

(same logflags output)

Development Settings
--------------------

The following environment variables modify Syncthing's behavior in ways that
are mostly useful for developers. Use with care. See also the --debug-* options
above.

 STTRACE           A comma separated string of facilities to trace. The valid
                   facility strings are listed below.

 STDEADLOCKTIMEOUT Used for debugging internal deadlocks; sets debug
                   sensitivity. Use only under direction of a developer.

 STLOCKTHRESHOLD   Used for debugging internal deadlocks; sets debug
                   sensitivity.  Use only under direction of a developer.

 STHASHING         Select the SHA256 hashing package to use. Possible values
                   are "standard" for the Go standard library implementation,
                   "minio" for the github.com/minio/sha256-simd implementation,
                   and blank (the default) for auto detection.

 GOMAXPROCS        Set the maximum number of CPU cores to use. Defaults to all
                   available CPU cores.

 GOGC              Percentage of heap growth at which to trigger GC. Default is
                   100. Lower numbers keep peak memory usage down, at the price
                   of CPU usage (i.e. performance).

Debugging Facilities
--------------------

(same debugging facilities)
```

I removed the environment variables that are now seen as command line
flags from the "Development Settings" section.


cmd/syncthing/main.go
```diff
@@ -143,218 +111,211 @@ Debugging Facilities
...
+
+	// Debug options below
+	DebugBlockProfiler        bool          `env:"STBLOCKPROFILE" help:"Write block profiles to block-$pid-$timestamp.pprof every 20 seconds"`
+	DebugCPUProfile           bool          `help:"Write a CPU profile to cpu-$pid.pprof on exit" env:"CPUPROFILE"`
+	DebugDBIndirectGCInterval time.Duration `env:"STGCINDIRECTEVERY" help:"Database indirection GC interval"`
+	DebugDBRecheckInterval    time.Duration `env:"STRECHECKDBEVERY" help:"Database metadata recalculation interval"`
+	DebugDeadlockTimeout      int           `placeholder:"SECONDS" env:"STDEADLOCKTIMEOUT" help:"Used for debugging internal deadlocks"`
+	DebugGUIAssetsDir         string        `placeholder:"PATH" help:"Directory to load GUI assets from" env:"STGUIASSETS"`
+	DebugHeapProfile          bool          `env:"STHEAPPROFILE" help:"Write heap profiles to heap-$pid-$timestamp.pprof each time heap usage increases"`
+	DebugNetworkProfilerURL   string        `placeholder:"URL" env:"STPROFILER" help:"Network profiler listen URL"`
```
imsodin, line 174: I'd rename the various profilers to `DebugProfile...`, to group them under their common denominator.

imsodin (commented):
Looks good.

>     * There's probably more that could be done here, but
>       I didn't want to go overboard.

What was the measure for what is "overboard" or not?  
E.g. the `STTRACE` one is something that is used quite often, I'd say at least as often as the profiling options, so I'd expect to see it as a `--debug-...` too.

calmh:
STTRACE specifically is tricky because it often debugs things that happen during init, like in the mutex package where it really needs to be set before we start creating mutexes... I think refactoring that so that each package has a `SetTracing(...)` that are called from `main` would be large and probably complicated, so the poster boy for "going overboard" for this PR. :)

calmh:
```
      --debug-db-indirect-gc-interval=DURATION    Database indirection GC interval ($STGCINDIRECTEVERY)
      --debug-db-recheck-interval=DURATION        Database metadata recalculation interval ($STRECHECKDBEVERY)
      --debug-deadlock-timeout=SECONDS            Used for debugging internal deadlocks ($STDEADLOCKTIMEOUT)
      --debug-gui-assets-dir=PATH                 Directory to load GUI assets from ($STGUIASSETS)
      --debug-perf-stats                          Write running performance statistics to perf-$pid.csv (Unix only) ($STPERFSTATS)
      --debug-profile-block                       Write block profiles to block-$pid-$timestamp.pprof every 20 seconds ($STBLOCKPROFILE)
      --debug-profile-cpu                         Write a CPU profile to cpu-$pid.pprof on exit ($CPUPROFILE)
      --debug-profile-heap                        Write heap profiles to heap-$pid-$timestamp.pprof each time heap usage increases ($STHEAPPROFILE)
      --debug-profiler-listen=ADDR                Network profiler listen address ($STPROFILER)
      --reset-database                            Reset the database, forcing a full rescan and resync
      --reset-deltas                              Reset delta index IDs, forcing a full index exchange
```

imsodin (approved):

AudriusButkevicius (approved):

PR #7331, opened 2021-02-04
all: Failure reporting fixes

imsodin:
Failure reports should be arriving on sentry, but didn't: Forgot the crash reporting url has a path already, `newcrash`.

Also empty failure reports are sent on shutdown. And `GET` used for sending failure reports - which works as we don't care, but is still weird. And finally: There's lots of "failure report is not a string" failure reports, because errors from the db have not been stringified.

### DIFF
--- a/lib/ur/failurereporting.go
+++ b/lib/ur/failurereporting.go
@@ -194,7 +194,7 @@ func sendFailureReports(ctx context.Context, reports []FailureReport, url string
 
 	reqCtx, reqCancel := context.WithTimeout(ctx, sendTimeout)
 	defer reqCancel()
-	req, err := http.NewRequestWithContext(reqCtx, http.MethodPut, url, &b)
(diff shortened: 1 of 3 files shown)


lib/ur/failurereporting.go
```diff
@@ -192,7 +194,7 @@ func sendFailureReports(ctx context.Context, reports []FailureReport, url string
 
 	reqCtx, reqCancel := context.WithTimeout(ctx, sendTimeout)
 	defer reqCancel()
-	req, err := http.NewRequestWithContext(reqCtx, http.MethodGet, url, &b)
+	req, err := http.NewRequestWithContext(reqCtx, http.MethodPut, url, &b)
```
calmh, line 197: Technically I think this ought to be a POST ("send data/command to this resource") rather than PUT ("replace the content of this resource")?

calmh (approved):

calmh (approved):

PR #7332, opened 2021-02-05
cmd/syncthing: Add decrypt subcommand

calmh:
This adds the `syncthing decrypt` subcommand that is used to
(offline-)decrypt or just verify the contents of an encrypted folder.

```
% syncthing decrypt --help
Usage: syncthing decrypt <path>

Decrypt or verify an encrypted folder

Arguments:
  <path>    Path to encrypted folder

Flags:
  -h, --help                Show context-sensitive help.

      --to=PATH             Destination directory, when decrypting
      --verify              Don't decrypt, just verify that the files are valid
      --password=STRING     Folder password for decryption / verification ($FOLDER_PASSWORD)
      --folder-id=STRING    Folder ID of the encrypted folder, if it cannot be determined automatically
      --continue            Continue processing next file in case of error, instead of aborting
      --verbose             Show verbose progress information
      --token-path=PATH     Path to the token file within the folder (used to determine folder ID)
```

**N.B.** this is based on the kong branch, so needs merging after that, and also illustrates how nice it is to add subcommands in this manner.

### DIFF
--- a/cmd/syncthing/decrypt/decrypt.go
+++ b/cmd/syncthing/decrypt/decrypt.go
@@ -1,0 +1,245 @@
+// Copyright (C) 2021 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Package decrypt implements the `syncthing decrypt` subcommand.
+package decrypt
+
+import (
+	"encoding/binary"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"path/filepath"
+
+	"github.com/syncthing/syncthing/lib/fs"
+	"github.com/syncthing/syncthing/lib/protocol"
+	"github.com/syncthing/syncthing/lib/scanner"
+)
+
+type CLI struct {
+	Path      string `arg:"" required:"1" help:"Path to encrypted folder"`
+	To        string `xor:"mode" placeholder:"PATH" help:"Destination directory, when decrypting"`
+	Verify    bool   `xor:"mode" help:"Don't decrypt, just verify that the files are valid"`
+	Password  string `help:"Folder password for decryption / verification" env:"FOLDER_PASSWORD"`
+	FolderID  string `help:"Folder ID of the encrypted folder, if it cannot be determined automatically"`
+	Continue  bool   `help:"Continue processing next file in case of error, instead of aborting"`
+	Verbose   bool   `help:"Show verbose progress information"`
+	TokenPath string `placeholder:"PATH" help:"Path to the token file within the folder (used to determine folder ID)"`
+
+	folderKey *[32]byte
+}
+
+type storedEncryptionToken struct {
+	FolderID string
+	Token    []byte
+}
+
+func (c *CLI) Run() error {
+	log.SetFlags(0)
+
+	if c.To == "" && !c.Verify {
+		return fmt.Errorf("must set --to or --verify")
+	}
+
+	if c.TokenPath == "" {
+		// This is a bit long to show as default in --help
+		c.TokenPath = ".stfolder/syncthing-encryption_password_token"
+	}
+
+	if c.FolderID == "" {
+		// We should try to figure out the folder ID
+		folderID, err := c.getFolderID()
+		if err != nil {
+			log.Println("No --folder-id given and couldn't read folder token")
+			return fmt.Errorf("getting folder ID: %w", err)
+		}
+		c.FolderID = folderID
+	}
+
+	if c.Verbose {
+		log.Println("Folder ID:", c.FolderID)
+	}
+
+	c.folderKey = protocol.KeyFromPassword(c.FolderID, c.Password)
+
+	return c.walk()
+}
+
+// walk finds and processes every file in the encrypted folder
+func (c *CLI) walk() error {
+	srcFs := fs.NewFilesystem(fs.FilesystemTypeBasic, c.Path)
+	var dstFs fs.Filesystem
+	if c.To != "" {
+		dstFs = fs.NewFilesystem(fs.FilesystemTypeBasic, c.To)
+	}
+
+	return srcFs.Walk("/", func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if !info.IsRegular() {
+			return nil
+		}
+		if fs.IsInternal(path) {
+			return nil
+		}
+
+		return c.withContinue(path, c.process(srcFs, dstFs, path))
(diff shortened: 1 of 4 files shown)


calmh:
So for the record I ran this in verify mode on my two existing encrypted folders, and one passes without comment (phew), the other threw a few errors:

```
jb@lun0:~ $ sudo ./syncthing decrypt --verify --password=youwish /data/syncthing/kastelo-untrusted --continue
Warning: Projects - Documents/.849C9593-D756-4E56-8D6E-42412F2A707B: encrypted block 0 (1064 bytes): chacha20poly1305: message authentication failed
Warning: Marketing - Documents/some-word-doc.docx: plaintext block 0 (95263 bytes) failed validation after decryption
Warning: Holding - Documents/.849C9593-D756-4E56-8D6E-42412F2A707B: encrypted block 0 (1064 bytes): chacha20poly1305: message authentication failed
Warning: Design - Documents/.849C9593-D756-4E56-8D6E-42412F2A707B: encrypted block 0 (1064 bytes): chacha20poly1305: message authentication failed
Warning: Seychelles - Documents/.849C9593-D756-4E56-8D6E-42412F2A707B: encrypted block 0 (1064 bytes): chacha20poly1305: message authentication failed
Warning: Admin HR - Documents/.849C9593-D756-4E56-8D6E-42412F2A707B: encrypted block 0 (1064 bytes): chacha20poly1305: message authentication failed
Warning: Marketing - Documents/.849C9593-D756-4E56-8D6E-42412F2A707B: encrypted block 0 (1064 bytes): chacha20poly1305: message authentication failed
Warning: Marketing - Documents/some-other-word-doc.docx: plaintext block 1 (6986 bytes) failed validation after decryption
Warning: Fastigheter - Documents/.849C9593-D756-4E56-8D6E-42412F2A707B: encrypted block 0 (1064 bytes): chacha20poly1305: message authentication failed
```

So two word documents in one folder where the data decrypted correctly but didn't match the (plaintext) block hashes. This is not a great sign, these are not documents that would change very often, and in one case it was only one block out of two while the word format is a zip file, so it should change entirely or not at all... I suspect either hardware or other error has hit those files at some point; refreshing them from the source caused the error to go away.

The other guid-named files are files that for some reason are stuck with "generic error" on the download side, I didn't check why yet, but it's a problem higher in the chain than encryption.

So at least the check seems to check out.

acolomb:
If the command `decrypt --verify` does not really decrypt to a persistent target, I think `--verify-only` would be a better option name.

cmd/syncthing/decrypt/decrypt.go
```diff
@@ -0,0 +1,261 @@
...
+	if _, err := io.ReadFull(fd, trailer); err != nil {
+		return nil, err
+	}
+
+	var plainFi protocol.FileInfo
+	if err := plainFi.Unmarshal(trailer); err != nil {
+		return nil, err
+	}
+
+	return &plainFi, nil
```
imsodin, line 260: s/plain/enc/

cmd/syncthing/decrypt/decrypt.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+
+		plainFd, err = dstFs.Create(plainFi.Name)
+		if err != nil {
+			return fmt.Errorf("%s: %w", plainFi.Name, err)
+		}
+	}
+
+	if err := c.decryptFile(encFi, &plainFi, encFd, plainFd); err != nil {
+		return fmt.Errorf("%s: %w", plainFi.Name, err)
```
imsodin, line 166: `plainFd` doesn't get closed here.

cmd/syncthing/decrypt/decrypt.go
```diff
@@ -0,0 +1,261 @@
...
+			return fmt.Errorf("encrypted block %d (%d bytes): %w", i, encBlock.Size, err)
+		}
+
+		// Decrypt it
+		dec, err := protocol.DecryptBytes(buf, fileKey)
+		if err != nil {
+			return fmt.Errorf("encrypted block %d (%d bytes): %w", i, encBlock.Size, err)
+		}
+
+		// Verify the block against the expected plaintext
```
imsodin, line 200:
```suggestion
		// Verify the block size against the expected plaintext
```

cmd/syncthing/decrypt/decrypt.go
```diff
@@ -0,0 +1,261 @@
...
+		if !scanner.Validate(dec, plainBlock.Hash, 0) {
+			// The block decrypted correctly but fails the hash check. This
+			// is odd and unexpected, but it it's still a valid block from
+			// the source. The file might have changed while we pulled it?
+			err := fmt.Errorf("plaintext block %d (%d bytes) failed validation after decryption", i, plainBlock.Size)
+			if c.Continue {
+				log.Printf("Warning: %s: %v", plainFi.Name, err)
+			} else {
+				return err
+			}
```
imsodin, line 219: Maybe a relict from before the `withContinue` wrapper existed? Or I don't see the relevant difference of this error to the other encryption related errors in this function.

imsodin (commented):
There's a whole

cmd/syncthing/decrypt/decrypt.go
calmh: This is intentional, but not well explained. For file-fatal errors we return and let `withContinue` handle it, but this is a "soft" error where the file might not be entirely toast just because a blocks fails the hash check. It decrypted correctly, meaning it was decrypted and authenticated with the correct key, so it was certainly created by someone with the right key and *should* be mostly valid.

I still consider read, decryption, and write errors file-fatal though. Maybe we could have a `--continue-harder` for those cases, when we really want to get a file with just some blocks in it.

Also I see I don't delete failed files, which should maybe happen.

calmh:
I spent several tens of seconds thinking of a succinct and precise way to express this, but I'm not sure it's successful. Open for better suggestions. Needs to convey that 1. no plaintext ever hits storage and 2. files are still decrypted and compared to their expected plaintext hashes after decryption...

`--verify-only         Don't write decrypted files to disk (but verify plaintext hashes)`

acolomb:
> `--verify-only         Don't write decrypted files to disk (but verify plaintext hashes)`

That's clear and concise to me.

uok:
maybe switch texts - what it does do first and what it doesn't in bracket?

cmd/syncthing/decrypt/decrypt.go
imsodin: Agree, I'd delete failed items. Thus one has only valid files in the target directory. One can still investigate failed files with the warnings afterwards if desired. Though looking at the sample run output you posted, there's no reference to the encrypted filename. That should probably be there. Not only for investigation, but also so one can remove those files to re-sync them with the correct contents if necessary.

I don't see the value of `--continue-harder`

imsodin (approved):

PR #7336, opened 2021-02-05
lib/db: Fix and improve removing entries from global (ref #6501)

imsodin:
Due to a support request about a panic in drop I was looking at the db code once again. And did find *something*, though I don't see how it would cause a panic. It still is incorrect and I have a test to prove that :)

In the code updating the needs, the check only looks at whether a device needed the file with the old global (using `removedFV` which is confusing, but at this point equal to `globalFV` - that's not the problem). However the device might still need the file now, compared to the new global.

### DIFF
--- a/lib/db/db_test.go
+++ b/lib/db/db_test.go
@@ -951,6 +951,49 @@ func TestDuplicateNeedCount(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


calmh:
Confusing

calmh (approved):

PR #7342, opened 2021-02-08
build: Fix strings in versioninfo for Windows (fixes #7340)

### LINKED ISSUE #7340: Change name from "Open Source Continuous File Synchronization" to "Syncthing" in Windows Firewall
Right now, Syncthing is added to Windows Firewall as

```
Name: Open Source Continuous File Synchronization
Description: Open Source Continuous File Synchronization
```

I am not sure where exactly this is configured, but it would be useful to have the name recorded as "Syncthing", so that it is easy to find it in the firewall settings.

image

**Edit**: The problem can also be seen when adding the program to Windows Firewall.

image

The name is "Open Source Continuous File Synchronization", and the publisher is "Unknown". I can see that these strings seem to be added through `build.go`, so I will try to experiment and check whether this can be fixed easily.

tomasz1986:
Add missing required strings and modify the description, so that it says
"Syncthing - Open Source Continuous File Synchronization", as apparently
the Windows Firewall uses this string both for the name and description.

Ref: https://docs.microsoft.com/windows/win32/menurc/versioninfo-resource


tomasz1986:
Before:

image

After:

image

imsodin:
If possible I'd rather not set the company field. I saw in the linked docs that the company is "required", however we didn't put it before and it worked. Is there any real downside to not setting it?

tomasz1986:
Yeah, it is responsible for the "publisher" field, which otherwise is shown as "Unknown" (see the screenshot above).

imsodin (approved):
Right, definitely better than "Unknown" - missed that. And given it matches what we have in the copyright headers "The Syncthing Authors" seems appropriate.

tomasz1986:
Yes, I did actually consult the copyright notes when adding this information 🙂.

PR #7351, opened 2021-02-10
build: Add build process for newgui

calmh:
This adds the opt-in flag `--with-new-gui` (or `$BUILD_NEW_GUI`) for building and including the "new GUI" which is now, inventively, in the `newgui` directory. It reuses our lazy asset rebuild stuff to only trigger the `npm` steps if they are actually required.

**Note!** This again puts the GUI in the place of a *theme*, meaning we shouldn't do this in production until there is a way to switch back to a standard theme.


AudriusButkevicius:
Can we pick a better theme name than newgui?

build.go
```diff
@@ -764,11 +766,47 @@ func rebuildAssets() {
 }
 
 func lazyRebuildAssets() {
-	if shouldRebuildAssets("lib/api/auto/gui.files.go", "gui") || shouldRebuildAssets("cmd/strelaypoolsrv/auto/gui.files.go", "cmd/strelaypoolsrv/gui") {
+	shouldRebuild := shouldRebuildAssets("lib/api/auto/gui.files.go", "gui") ||
+		shouldRebuildAssets("cmd/strelaypoolsrv/auto/gui.files.go", "cmd/strelaypoolsrv/gui")
+
+	if withNewGUI {
+		shouldRebuild = buildNewGUI() || shouldRebuild
+	}
```
imsodin, line 774: Could insert a notice here like `Not building experimental *insertFancyBetterName*; use -with-new-gui to build it`.

calmh:
I knew the naming would be a thing, I agree `newgui` is crap :) When deciding on a name we should also change the name in the actual npm project as that affects the generated folder name. I guess `next-gen-gui` is slightly nicer?

AudriusButkevicius:
Sounds good.

imsodin:
Yes - boring, technocratic names :D  
I was thinking `gui-next`, but I like the little note of "gravity" with generation in `next-gen-gui`. Really, I am fine with anything.

calmh:
Commit grew since it's now a rename of the gui directory as well, plus a `s/tech-ui/next-gen-gui/` in package.json and angular.json to make the build reasonable. The README is wrong and out of date but I don't have the energy to fix it right now.

calmh:
PTAL? we might merge this to silence the gui build failure on everything else, as I adjusted the build for this PR...

AudriusButkevicius (approved):

calmh:
Merge manually because I accidentally the entire branch and then fumbled it. 🤷

PR #7357, opened 2021-02-11
lib/versioner: Improve error messages (fixes #7354)

### LINKED ISSUE #7354: Error message when external versioning fails is ambiguous
If external versioning fails to move a file, we get the following error.

The problem is that the error message, i.e. `exit status 255` means nothing to the user. It would be better to have something specifically saying that the file could not be versioned or such.

Tested with Syncthing v1.13.1.

imsodin:
Attaches stderr (shortened if long) to the error when external versioner fails and adds a `*type* versioner: *operation*:` prefix to all errors returned for context (e.g. `simple versioner: get versions: permission denied`). 

All the change outside of `lib/versioner` is just noop boilerplate to use `errors.Is` in one test instead of fragile string testing.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -88,7 +88,7 @@ type Model interface {
 	SetIgnores(folder string, content []string) error
 
 	GetFolderVersions(folder string) (map[string][]versioner.FileVersion, error)
-	RestoreFolderVersions(folder string, versions map[string]time.Time) (map[string]string, error)
+	RestoreFolderVersions(folder string, versions map[string]time.Time) (map[string]error, error)
 
 	DBSnapshot(folder string) (*db.Snapshot, error)
 	NeedFolderFiles(folder string, page, perpage int) ([]db.FileInfoTruncated, []db.FileInfoTruncated, []db.FileInfoTruncated, error)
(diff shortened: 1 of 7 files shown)


calmh:
Keeping in mind that even if this is probably fine, the gui by default cuts at the last colon and only shows that part...

calmh (dismissed):

tomasz1986:
I have compiled and tested the PR, and this is how the GUI looks right now.

image

The relevant log entry:

```
[7ZL7H] DEBUG: sendreceive/su9ww-gwwzx@0xc000028380 Deleting file  %&()
[7ZL7H] DEBUG: sendreceive/su9ww-gwwzx@0xc000028380 new error for  %&(): delete file: external versioner: archive: exit status 255
[7ZL7H] DEBUG: sendreceive/su9ww-gwwzx@0xc000028380 changed 1 on try 3
[7ZL7H] INFO: Puller (folder "test1" (su9ww-gwwzx), item " %&()"): syncing: delete file: external versioner: archive: exit status 255
```

lib/model/model.go
```diff
@@ -2637,11 +2637,11 @@ func (m *model) RestoreFolderVersions(folder string, versions map[string]time.Ti
 		return nil, errNoVersioner
 	}
 
-	restoreErrors := make(map[string]string)
+	restoreErrors := make(map[string]error)
 
 	for file, version := range versions {
 		if err := ver.Restore(file, version); err != nil {
-			restoreErrors[file] = err.Error()
+			restoreErrors[file] = err
```
AudriusButkevicius, line 2644: What happens when you pass errors to json serialiser?
I assume this now might become an object and break the UIs expectations?

AudriusButkevicius (dismissed):

AudriusButkevicius (approved):

PR #7358, opened 2021-02-12
gui: Add Docs help links to each setting in Advanced Configuration

tomasz1986:
Add a link next to each setting's label to its explanation in the Docs.
This way, it is easy to quickly check what the setting is about without
going to the Docs site separately and searching for it manually.

### DIFF
--- a/gui/default/syncthing/settings/advancedSettingsModalView.html
+++ b/gui/default/syncthing/settings/advancedSettingsModalView.html
@@ -16,7 +16,7 @@
           <div class="panel-body">
             <form class="form-horizontal" role="form">
               <div ng-repeat="(key, value) in advancedConfig.gui" ng-init="type = inputTypeFor(key, value)" ng-if="type != 'skip'" class="form-group">
-                <label for="guiInput{{$index}}" class="col-sm-4 control-label">{{key | uncamel}}&nbsp;<a href="https://docs.syncthing.net/users/config.html#gui-{{key | lowercase}}" target="_blank"><span class="fas fa-question-circle"></span></a></label>
+                <label for="guiInput{{$index}}" class="col-sm-4 control-label">{{key | uncamel}}&nbsp;<a href="{{docsURL('users/config.html#config-option-gui.')}}{{key | lowercase}}" target="_blank"><span class="fas fa-question-circle"></span></a></label>
                 <div class="col-sm-8">
                   <input ng-if="inputTypeFor(key, value) == 'list'" id="guiInput{{$index}}" class="form-control" type="text" ng-model="advancedConfig.gui[key]" ng-list />
                   <input ng-if="inputTypeFor(key, value) != 'list'" id="guiInput{{$index}}" class="form-control" type="{{inputTypeFor(key, value)}}" ng-model="advancedConfig.gui[key]" />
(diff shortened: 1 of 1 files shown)


tomasz1986:
This is how it looks like.



WIP, as this will only work if 

1. https://github.com/syncthing/docs/pull/607 is implemented.

2. all settings are actually documented. Right now, a few are still missing from the Docs.

imsodin:
This is likely to generate links pointing at nothing. In my opinion a link to https://docs.syncthing.net/users/config.html at the top would have the same gain with less fragility/busy-work (adding the config titles).

tomasz1986:
> This is likely to generate links pointing at nothing.

Hmm, I would not really say so. There are some missing entries, but the vast majority of the options are actually present in the Docs.

> In my opinion a link to [docs.syncthing.net/users/config.html](https://docs.syncthing.net/users/config.html) at the top would have the same gain with less fragility/busy-work (adding the config titles).

Accessing the Docs is not a big deal. The problem is that searching through that page is a major pain. I am just speaking from my own experience, but especially on touch screen devices, looking for individual entries is a nightmare.

imsodin:
Right, and it's only a missing anchor - leaves the user still on the right page.

(We should just move the docs repo into the main one and add a lint step to ensure all options are present. Or even better autogenerate from a proto comment or something. Anyway, I am digressing :) ).

st-review:
:robot: *beep boop*

acolomb:
I would like to see this happen as a very useful addition.  To avoid the burden of having to maintain sync between the anchor links into the docs repo, how about just using the Sphinx-generated search facility, e.g.:

https://docs.syncthing.net/search.html?q=CaseSensitiveFS

That at least gives clickable results and is not even limited to the `*.rst` files in `config/` and `advanced/`.  Maybe there is also a way to make Sphinx highlight a certain search term within a specific target page through a URL parameter?

tomasz1986:
I personally prefer direct links, as the search may return too many results for some of the options, with the actual configuration page listed somewhere down below the less relevant ones. The links here use the same IDs as are used in `config.xml`, so as long as the Docs follow the convention (which they should, for the sake of consistency), I don't think there would be any real burden when it comes to maintenance. It was mentioned before, but even if an ID link becomes broken for some reason, it will still redirect to the configuration page, which I would say is acceptable.

What I would like to do in addition to this one though would to be have some highlighting in CSS for `:target` in the Docs, so that the linked option with its description gets automatically highlighted in yellow or something, for better visibility.

acolomb:
Reopening as this will be very useful once https://github.com/syncthing/docs/pull/710 is merged.

acolomb:
Please migrate to the versioned links once https://github.com/syncthing/syncthing/pull/8108 is merged.

tomasz1986:
> Please migrate to the versioned links once #8108 is merged.

I've migrated the links 🙂.

acolomb:
We removed the .html extension on all links, please do the same here.

tomasz1986:
> We removed the .html extension on all links, please do the same here.

Done.

acolomb:
Some don't work:

https://docs.syncthing.net/users/config#config-option-device.addresses
https://docs.syncthing.net/users/config#config-option-device.allowednetworks
https://docs.syncthing.net/users/config#config-option-device.deviceid
https://docs.syncthing.net/users/config#config-option-device.ignoredfolders
https://docs.syncthing.net/users/config#config-option-folder.devices
https://docs.syncthing.net/users/config#config-option-gui.usetls
https://docs.syncthing.net/users/config#config-option-options.alwayslocalnets
https://docs.syncthing.net/users/config#config-option-options.crurl
https://docs.syncthing.net/users/config#config-option-options.featureflags
https://docs.syncthing.net/users/config#config-option-options.globalannounceservers
https://docs.syncthing.net/users/config#config-option-options.listenaddresses
https://docs.syncthing.net/users/config#config-option-options.stunservers
https://docs.syncthing.net/users/config#config-option-options.unackednotificationids

The links are generated based on the JSON names used in the GUI.  The documentation is written for the XML representation.  Sometimes they use different name tags, or the plural / singular form differs (e.g. multiple `<address>` elements vs. one `addresses: [ ... ]` array).

acolomb:
I think we can handle that gracefully by defining aliases on the individual `option` directives in the docs.  That's a manual process, but only few options to touch overall.

acolomb:
It was actually easy to solve:  https://github.com/syncthing/docs/pull/724

calmh (approved):
Seems fine to me. Leaving unmerged in case I missed something that should be tweaked what with the ongoing docs changes, otherwise merge at will.

acolomb:
Yeah it doesn't really depend on the docs changes (https://github.com/syncthing/docs/pull/724).  Until that one is merged, some generated links here will just point at the config page in general.  No big deal.

PR #7360, opened 2021-02-12
gui: Disable Cleanup Interval with External File Versioning as unsupported

tomasz1986:
As for right now, the External File Versioning does not support Cleanup
Interval. Therefore, the option should no be available at all when using
it.

Ref: https://forum.syncthing.net/t/16346


tomasz1986:
I am doing this on the assumption that Cleanup Interval does not work with External Versioning, judging by the forum thread and my own code checking. If this is incorrect, please let me know.

Before:



After:

AudriusButkevicius (approved):

PR #7362, opened 2021-02-13
gui: Fix setting external versioning command (fixes #7361)

### LINKED ISSUE #7361: External versioning command does not save from the Web GUI
1. Try to set File Versioning to external with any command, e.g.

   

2. Click "Save".
3. Re-open the "Edit Folder" window and go to "File Versioning" again.
4. The command is blank. It is missing from `config.xml` also.

   

I though that it could be my change (https://github.com/syncthing/syncthing/commit/5e9c7bc022040555d56860148352e5862737889f), but no, even after reverting the commit, the problem is still here. Basically, v1.13.1 works, but `main` (`80fc238be`) does not.

What is interesting is that if we add the command through `config.xml` directly, then it does stick (until we try to save any changes to the folder, that is, as then it gets cleared again).

imsodin:
Regression introduced in https://github.com/syncthing/syncthing/pull/7131

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1916,7 +1916,7 @@ angular.module('syncthing.core')
                 break;
             case "external":
                 $scope.currentFolder._guiVersioning.selector = "external";
-                $scope.currentFolder.externalCommand = currentVersioning.params.command;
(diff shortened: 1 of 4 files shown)


tomasz1986:
There is a typo in the title: "verisoning" -> "versioning".

calmh (approved):

PR #7364, opened 2021-02-14
cmd/syncthing: Add cli as a subcommand (fixes #6566, fixes #4719)

### LINKED ISSUE #4719: Package syncthing-cli as Debian package, if/when it's deemed functional
Is it possible to get the CLI in the apt/debian package? 

I would like to script against syncthing for remote use cases.

### LINKED ISSUE #6566: Ship stcli in main package
I can't see why we don't package it.

Alternatively, we could merge it into the main binary, potentially something like
```
syncthing cli <args to cli>
```

I can do the legwork if we deem this appropriate.

AudriusButkevicius:


calmh:
Neat! Is it possible to make `syncthing cli --help` equivalent with `syncthing cli help`? As it is it's not obvious how to get to the "real" help otherwise. And that help refers to just `syncthing [options] command <args>` or something, omitting the "cli" subcommand that's also necessary.

AudriusButkevicius:
done

imsodin:
Do we really want the `cli` intermediate command"? Calling `syncthing config ...` directly seems simpler and cleaner to me.

AudriusButkevicius:
That is harder to do

imsodin:
Right, there's e.g. the common flags. Maybe it could be called `api` or `rest` instead to convey their common denominator?

AudriusButkevicius:
Sure, I am happy to call it api, rest doesn't mean much to people outside of development.

AudriusButkevicius:
The common flag stuff is moot, or atleast kong docs suck at explaining on how to have global flags.
Also, things like decrypt don't care about ui address, so it would make no sense to have them global.

imsodin:
Well, it's anyway bikeshedding I started here, not worth spending time on - sorry about that :)  
I am fine with either `cli` or `api`.

go.mod
```diff
@@ -40,6 +40,7 @@ require (
 	github.com/shirou/gopsutil/v3 v3.20.11
 	github.com/syncthing/notify v0.0.0-20201210100135-17de26665ddc
 	github.com/syndtr/goleveldb v1.0.1-0.20200815071216-d9e9293bd0f7
+	github.com/thejerf/suture v4.0.0+incompatible
 	github.com/thejerf/suture/v4 v4.0.0
```
calmh, line 44: We should probably have just one suture dependency?

calmh (commented):
LGTM. I think `cli` is fine. There are aspects about the actual CLI I want to improve afterwards, but that's for afterwards.

go.mod
AudriusButkevicius: idk why but my IDE and go command line just go mad about this stuff. Lemme back out all the go.mod changes.

calmh (approved):

AudriusButkevicius:
Re changes you want. What changes are you after?

calmh:
Primarily I think the common methods to look up file metadata, folder status, and I'd like to remove the "reset" operation which is kind of easy and prominently accessible yet very destructive...

PR #7365, opened 2021-02-15
gui: Hide non-functional Versions button when using External Versioning

tomasz1986:
The button does nothing when the External Versioning is being used, so
it should not be displayed at all to avoid confusing the users.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -548,7 +548,7 @@
                     <button ng-if="folder.paused" type="button" class="btn btn-sm btn-default" ng-click="setFolderPause(folder.id, false)">
                       <span class="fas fa-play"></span>&nbsp;<span translate>Resume</span>
                     </button>
-                    <button type="button" class="btn btn-default btn-sm" ng-click="restoreVersions.show(folder.id)" ng-if="folder.versioning.type">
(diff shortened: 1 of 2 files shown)


tomasz1986:
Before:



After:

gui/default/index.html
```diff
@@ -548,7 +548,7 @@ <h4 class="panel-title">
                     <button ng-if="folder.paused" type="button" class="btn btn-sm btn-default" ng-click="setFolderPause(folder.id, false)">
                       <span class="fas fa-play"></span>&nbsp;<span translate>Resume</span>
                     </button>
-                    <button type="button" class="btn btn-default btn-sm" ng-click="restoreVersions.show(folder.id)" ng-if="folder.versioning.type">
+                    <button type="button" class="btn btn-default btn-sm" ng-click="restoreVersions.show(folder.id)" ng-if="folder.versioning.type != 'external'">
```
AudriusButkevicius, line 551: Doesn't this make the button show up when versioning is disabled all together?

tomasz1986: Of course it does… :flushed:

Should be fixed now.

AudriusButkevicius (approved):

Issue #7359, opened 2021-02-12
Pause All button should not pause the own device


tomasz1986:
1. Use the "Pause All" button to pause all devices.
2. All devices in `config.xml`, including the own one, which normally has no option to be paused, have been changed to `<paused>true</paused>`.
3. Resume devices one by one by unpausing each of them separately.
4. The own device stays `<paused>true</paused>`.
5. The only way to get it back to `<paused>false</paused>` is to use the "Resume All" button (or modify the config file).

I would say that the "Pause All" button should not affect the own device at all, simply because as far as I understand, it does not even have the capability to be paused.

This issue does not seem to affect Syncthing itself, as the synchronisation still works, but it does affect 3rd party wrappers like Syncthing Tray (see https://forum.syncthing.net/t/yet-another-syncthing-tray/8502/102).

Martchus:
> I would say that the "Pause All" button should not affect the own device at all, simply because as far as I understand, it does not even have the capability to be paused.

That's how I'd understand it as well. It would be great if a Syncthing developer could confirm our understanding. If it is correct I suppose Syncthing Tray (like any other GUI wrapper) should ignore the own device being configured as paused.

Martchus:
I've also noticed that one can pause (and resume) the own device via the REST-API. That also seems unintended and should likely be prevented.

AudriusButkevicius:
Pause on self doesn't do anything.

Off the top of my head, the only reason we have "self" as a separate device is for ourselves is for things like name, so there isn't really a good reason to even have it nowaadays I think.

Martchus:
I've changed Syncthing Tray to ignore the paused flag on the own device: https://github.com/Martchus/syncthingtray/commit/851edeaafcf12d719047be8eb609c2ded9405753

---

You have been changing the ticket title and I think that's good. Of course that's a breaking API change. It means the own device will not show up in the list of devices anymore in Syncthing Tray (and any other GUI wrappers which simply display the list of devices as-is). I personally find it acceptable because the own device provided very little information and at least in Syncthing Tray the own device ID is visible via an extra button anyways. At the time when writing the code for Syncthing Tray I've already found it weird to have "extra logic" within the GUI just to sort out the own device (to give it a special status) and of course the same counts for the additional code I've just been adding to ignore the "paused" flag.

tomasz1986:
I would like to add one more observation. It seems that the "Resume All" is supposed to be displayed in place of "Pause All". However, if we do as follows:

[...]

PR #7366, opened 2021-02-15
gui: Ignore own device when pausing/resuming all devices (ref #7359)

tomasz1986:
Do not change the state of the current device when using the "Pause All"
and "Resume All" buttons. Also, ignore the state of the current device
when checking whether all devices are paused, so that the two buttons
appear and disappear properly, depending only the state of remote
devices.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1732,21 +1732,21 @@ angular.module('syncthing.core')
         };
 
         $scope.setAllDevicesPause = function (pause) {
-            for (var id in $scope.otherDevices()) {
-                $scope.otherDevices()[id].paused = pause;
-            };
+            $scope.otherDevices().forEach(function (device) {
+                device.paused = pause;
+            });
             $scope.config.devices = deviceList($scope.otherDevices());
             $scope.saveConfig();
         }
 
         $scope.isAtleastOneDevicePausedStateSetTo = function (pause) {
-            for (var id in $scope.otherDevices()) {
-                if ($scope.otherDevices()[id].paused == pause) {
-                    return true;
+            var paused = false;
+            $scope.otherDevices().forEach(function (device) {
+                if (device.paused == pause) {
+                    paused = true;
                 }
-            }
-
-            return false
+            });
+            return paused;
         }
 
         $scope.errorList = function () {
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1732,16 +1732,16 @@ angular.module('syncthing.core')
         };
 
         $scope.setAllDevicesPause = function (pause) {
-            for (var id in $scope.devices) {
-                $scope.devices[id].paused = pause;
+            for (var id in $scope.otherDevices()) {
+                $scope.otherDevices()[id].paused = pause;
             };
-            $scope.config.devices = deviceList($scope.devices);
+            $scope.config.devices = deviceList($scope.otherDevices());
             $scope.saveConfig();
         }
 
         $scope.isAtleastOneDevicePausedStateSetTo = function (pause) {
-            for (var id in $scope.devices) {
-                if ($scope.devices[id].paused == pause) {
+            for (var id in $scope.otherDevices()) {
+                if ($scope.otherDevices()[id].paused == pause) {
                     return true;
                 }
             }
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1735,7 +1735,7 @@ angular.module('syncthing.core')
             $scope.otherDevices().forEach(function (device) {
                 device.paused = pause;
             });
-            $scope.config.devices = deviceList($scope.otherDevices());
+            $scope.config.devices = deviceList($scope.devices);
             $scope.saveConfig();
         }
 
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -1733,7 +1733,7 @@ angular.module('syncthing.core')
 
         $scope.setAllDevicesPause = function (pause) {
             $scope.otherDevices().forEach(function (device) {
-                device.paused = pause;
+                $scope.devices[device.deviceID].paused = pause;
             });
             $scope.config.devices = deviceList($scope.devices);
             $scope.saveConfig();
@@ -1742,7 +1742,7 @@ angular.module('syncthing.core')
         $scope.isAtleastOneDevicePausedStateSetTo = function (pause) {
             var paused = false;
             $scope.otherDevices().forEach(function (device) {
-                if (device.paused == pause) {
+                if ($scope.devices[device.deviceID].paused == pause) {
                     paused = true;
                 }
             });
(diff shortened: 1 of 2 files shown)


AudriusButkevicius (approved):

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1733,15 +1733,17 @@ angular.module('syncthing.core')
 
         $scope.setAllDevicesPause = function (pause) {
             for (var id in $scope.devices) {
-                $scope.devices[id].paused = pause;
+                if ($scope.devices[id].deviceID !== $scope.myID) {
+                  $scope.devices[id].paused = pause;
+                }
```
calmh, line 1738: Since this was required in multiple places and we're liable to forget it in the future, maybe wrap the condition into an iterator like `$scope.forAllRemoteDevices(fn)` or so?

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1733,15 +1733,17 @@ angular.module('syncthing.core')
 
         $scope.setAllDevicesPause = function (pause) {
             for (var id in $scope.devices) {
-                $scope.devices[id].paused = pause;
+                if ($scope.devices[id].deviceID !== $scope.myID) {
```
imsodin, line 1736: That could just be
```suggestion
                if (id !== $scope.myID) {
```

gui/default/syncthing/core/syncthingController.js
imsodin: Seems sensible - what's your "status" on this, @tomasz1986?

tomasz1986: I have tried a different approach on this. What do you think about the current iteration?

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1732,16 +1732,16 @@ angular.module('syncthing.core')
         };
 
         $scope.setAllDevicesPause = function (pause) {
-            for (var id in $scope.devices) {
-                $scope.devices[id].paused = pause;
+            for (var id in $scope.otherDevices()) {
+                $scope.otherDevices()[id].paused = pause;
```
imsodin, line 1736:
```suggestion
                $scope.devices[id].paused = pause;
```

Same below

imsodin (changes requested):

gui/default/syncthing/core/syncthingController.js
tomasz1986: I have tried changing the code like that, but I am getting tons of these JS errors.

```
TypeError: Cannot read property 'paused' of undefined
    at Scope.$scope.isAtleastOneDevicePausedStateSetTo (syncthingController.js:1744)
    at Object.$parseFunctionCall [as get] (angular.js:12493)
    at Scope.$digest (angular.js:14423)
    at angular.js:14984
    at angular.js:5347
```

The modified code:

``` javascript
        $scope.setAllDevicesPause = function (pause) {
            for (var id in $scope.otherDevices()) {
                $scope.devices[id].paused = pause;
            };
            $scope.config.devices = deviceList($scope.otherDevices());
            $scope.saveConfig();
        }

        $scope.isAtleastOneDevicePausedStateSetTo = function (pause) {
            for (var id in $scope.otherDevices()) {
                if ($scope.devices[id].paused == pause) {
                    return true;
                }
            }

            return false
        }
```

imsodin: Ah, `otherDevices()` returns a list, i.e. the iterator variable is not a device-id, but a simple index. You can use `otherDevices.foreach(function (device) {...});` to iterate then.

tomasz1986: I have modified the code and it does work, but please check it and let me know what you think about it now.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1732,21 +1732,21 @@ angular.module('syncthing.core')
...
 
         $scope.setAllDevicesPause = function (pause) {
-            for (var id in $scope.devices) {
-                $scope.devices[id].paused = pause;
-            };
-            $scope.config.devices = deviceList($scope.devices);
+            $scope.otherDevices().forEach(function (device) {
+                device.paused = pause;
+            });
+            $scope.config.devices = deviceList($scope.otherDevices());
```
imsodin, line 1738: In the `foreach` you are modifying in place. That might work, i.e. modifies the corresponding object in `$scope.devices` directly. I'd still do that explicitly, i.e. modify `$scope.devices[device.DeviceID] = paused`. Then in the end you still need assign all devices to the config not just the "others".

tomasz1986: I have fixed the all devices assigning part, but if I try to do the explicit modification, e.g.

```
$scope.devices[device.DeviceID].paused = pause;
```

or

```
$scope.devices[device].paused = pause;
```

then I get

```
Cannot set property 'paused' of undefined
```

errors.

imsodin: Because I messed up the camelCase that is used in JSland: `$scope.devices[device.deviceID].paused = pause;`. Generally the console that shows you the error also lets you interact with the objects, i.e. you can see what `device` is and what its members are named.

tomasz1986: And I actually searched for a similar syntax in the rest of the code, and found it, but failed to notice the case difference too :hushed:. I just use the DevTools in Chromium, but my JS console skills are extremely basic.

I think that I have fixed everything now.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1732,21 +1732,21 @@ angular.module('syncthing.core')
...
+            var paused = false;
+            $scope.otherDevices().forEach(function (device) {
+                if ($scope.devices[device.deviceID].paused == pause) {
+                    paused = true;
                 }
-            }
-
-            return false
+            });
+            return paused;
```
imsodin, line 1749: I'd just `continue` if `== $scope.myID` for less iteration.

tomasz1986: To be fair, I'm not really sure where I should add the check. Do you mean that it should be added to the current `forEach` method, or rather that the current one should be replaced with a different code?

imsodin: Replace the foreach, basically going back to the previous code, with added handling `myID`. Something like the following (beware of mistakes - I wrote it in the github text editor).

```js
            for (var id in $scope.devices) {
                if ($scope.devices[id].paused === pause && id !== $scope.myID) {
                    return true;
                }
            }
            return false
```

It was a early return in a single loop before. Now it's one full loop in `otherDevices`, then another full loop over it's results, even if `paused` is already true.

st-review:
:robot: *beep boop*

acolomb:
@tomasz1986 are you still going to finish this?

tomasz1986:
> @tomasz1986 are you still going to finish this?

I don't think so. My JS skills are too weak to do this cleanly, and the whole "self" device is to be removed anyway (see https://github.com/syncthing/syncthing/issues/7359), so I'm not sure if it's worth it to spend a lot of time on this.

If you want to do the fix instead, please feel free to go ahead 😉.

PR #7367, opened 2021-02-15
gui: Hide the Rescan All button when no folders exist

tomasz1986:
If there are no folders present, show only the "Add Folder" button, and
hide the "Rescan All" button. Only show the latter when at least one
folder exists.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -570,7 +570,7 @@
             <button type="button" class="btn btn-sm btn-default" ng-click="setAllFoldersPause(false)" ng-if="isAtleastOneFolderPausedStateSetTo(true)">
               <span class="fas fa-play"></span>&nbsp;<span translate>Resume All</span>
             </button>
-            <button type="button" class="btn btn-sm btn-default" ng-click="rescanAllFolders()" ng-disabled="!isAtleastOneFolderPausedStateSetTo(false)">
+            <button type="button" class="btn btn-sm btn-default" ng-click="rescanAllFolders()" ng-if="isAtleastOneFolderPausedStateSetTo(true) || isAtleastOneFolderPausedStateSetTo(false)" ng-disabled="!isAtleastOneFolderPausedStateSetTo(false)">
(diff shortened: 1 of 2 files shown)


tomasz1986:
Before:



After:

calmh (approved):

gui/default/index.html
```diff
@@ -570,7 +570,7 @@ <h4 class="panel-title">
             <button type="button" class="btn btn-sm btn-default" ng-click="setAllFoldersPause(false)" ng-if="isAtleastOneFolderPausedStateSetTo(true)">
               <span class="fas fa-play"></span>&nbsp;<span translate>Resume All</span>
             </button>
-            <button type="button" class="btn btn-sm btn-default" ng-click="rescanAllFolders()" ng-disabled="!isAtleastOneFolderPausedStateSetTo(false)">
+            <button type="button" class="btn btn-sm btn-default" ng-click="rescanAllFolders()" ng-if="isAtleastOneFolderPausedStateSetTo(true) || isAtleastOneFolderPausedStateSetTo(false)" ng-disabled="!isAtleastOneFolderPausedStateSetTo(false)">
```
calmh, line 573: It seems like there ought to exist an easier criteria here, like the set of folders actually being non-empty, instead of iterating twice over the folders that do exist...

tomasz1986: Yeah, I know that this is hackish, but I could not come up with anything better using the current code. My JS skills are kind of limited, so I may have very likely missed something obvious though.

calmh: We have an `isEmptyObject` function so I think something like `ng-if="!isEmptyObject(folders)"` should do it

tomasz1986: I could not figure out the `isEmptyObject` method, but I have analysed the rest of the code, and in the end I have taken a different approach with `.length > 0` (which is already used in many other places in `index.html`.). I hope that this is fine now.

calmh (approved):

PR #7372, opened 2021-02-17
lib/connections: Allow QUIC with Go 1.16

calmh:
This appears to work for me locally, even though there isn't a newer released version of quic-go yet, and the master branch of quic-go has references to a qtls package for 1.16 that is not pulled in here...


calmh:
(Builders are 1.16 now, I had to change `GO386=387` to `GO386=softfloat` and exclude the new `ios` and `openbsd/mips` ports that didn't build.)

AudriusButkevicius (approved):

Issue #7166, opened 2020-11-27
Remote device address extremely long when connection failed


tomasz1986:
This is not exactly a bug per se, but is it really necessary to list all these addresses under a remote device when the connection has failed?

![image](https://user-images.githubusercontent.com/5626656/100464901-c25cd180-3111-11eb-8d20-1cad24071ea5.png)

I mean, it is nice to have all the details listed, but are these `quic://` and `relay://` address with any significance to the user?

calmh:
How useful it is to show discovered addresses is probably arguable, but if we assume that it's useful I don't see an argument that some are more useful than others... in this I think there being lots of addresses is useful information by itself. So yeah, I think the current state is adequate and desirable.

tomasz1986:
Personally, the reason why this a little annoying is that the information which I am usually looking for when the device is disconnected is `Last seen` (and possibly `Introducer`). Because of the superlong address, I am always forced to scroll down to get to those fields.

What is your opinion about moving the `Last seen` field above the `Address`?

imsodin:
Or maybe fold the list if it exceeds some limit, and show a "show X more" link at the bottom of it to expand the remaining addresses.

PR #7373, opened 2021-02-17
gui: Show "Last seen" at the top when device is disconnected (ref #7166)

tomasz1986:
Move the "Last seen" field to the very top in the device information.
This way, if a device has disconnected unexpectly, we can quickly check
the time when it was last available. Right now, due to the very long
address field, it is usually necessary to scroll down in order to view
the "Last seen" field.


tomasz1986:
Before:

image

After:

image

calmh (approved):

PR #7375, opened 2021-02-18
lib: Use counterfeiter to mock interfaces in tests

imsodin:
I was looking for test tooling to time tests (found [gotestsum](https://github.com/gotestyourself/gotestsum)) and in the process stumbled over [counterfeiter](https://github.com/maxbrunsfeld/counterfeiter): It generates functional test classes for any given interface. As I was annoyed by having to change the various `lib/api/mocked_...` types to make tests pass in the past, I very much appreciate the idea of having those generated automatically. There's more boilerplate to get rid of in the fake connection in model testing. And the mocked classes provide lots of functionality (controlling what functions do, providing info on which functions were called including arguments, ...) which we could make use of (I guess a lot of the request_test.go stuff could be simplified.

The mocks are created with go:generate, i.e. there's now more assets to generate. I refactored the lazy asset building to work on a list of assets, and rebuild those which are necessary. An asset specifies a list of source and destination paths, which are all walked to get the newest time (and missing root means rebuild too).

There's one ugly hack: The mocked classes always include a reference to the interface. That can lead to cyclic imports. And it seems unlikely that's going to change given the upstream history: https://github.com/maxbrunsfeld/counterfeiter/issues/121. Thus I added a generation script to remove this reference from the generated code for one protocol interface that we do use in protocol tests.


calmh:
This seems like a fair amount of complexity for something that isn't super hard to do manually, but allright. However I think we should treat it the same as protobuf stuff -- generate explicitly and check in, so that `go test` more or less works out of the box for most packages.

imsodin:
I see that. It's definitely not hard, just a bit annoying. And I liked this "mock-generator", so wanted to make use of it. Having done the change myself I am not a good judge for the trade-off between complexity/tediousness of the generation vs convenience/functionality of the mocks, as I am set to underestimate the former. I don't want to push this through, if you feel it's too much baggage or something.

calmh:
No I kinda like adding the mock behavior as required in the test rather than having it somewhere else in the mock, so this is fine to me. Just need to look at it again.

script/prune_mocks.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		err = exec.Command("goimports", "-w", path).Run()
+		if err != nil {
+			log.Fatal(err)
+		}
+		return nil
+	})
+}
+
+func pruneInterfaceCheck(path string, size int64) error {
```
calmh, line 44: This is kind of meta, but this seems like a fragile function for several reasons. One obvious alternative would be to pass it through a filter that removes lines starting with `var _ =`, though of course you gain some extra juggling of a temp file and whatnot.

meta/copyright_test.go
```diff
@@ -66,6 +66,9 @@ func checkCopyright(path string, info os.FileInfo, err error) error {
 		if copyrightRe.MatchString(scanner.Text()) {
 			return nil
 		}
+		if scanner.Text() == "// Code generated by counterfeiter. DO NOT EDIT." {
```
calmh, line 69: More general: https://github.com/golang/go/issues/13560#issuecomment-288457920

calmh (commented):
I'm cool with the meat of the change. In fact, it's better since relevant functionality is moved closer to the tests that requires it. 👍

script/prune_mocks.go
```diff
@@ -42,26 +45,37 @@ func main() {
...
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(strings.TrimSpace(line), "var _ ") {
+			continue
 		}
-		if i := bytes.LastIndex(buf, searched); i != -1 {
-			pos += int64(i)
-			break
+		if _, err := tmp.WriteString(line + "\n"); err != nil {
+			fmt.Println(1)
```
calmh, line 67:
```suggestion
```

script/prune_mocks.go
```diff
@@ -42,26 +45,37 @@ func main() {
...
+			os.Remove(tmp.Name())
+			return err
 		}
-		pos -= chunk
 	}
-	return fd.Truncate(pos)
+
+	if err := fd.Close(); err != nil {
+		fmt.Println(2)
```
calmh, line 74:
```suggestion
```

calmh (approved):

PR #7381, opened 2021-02-18
lib/fs: Consider options in case-fs caching (fixes #7371)

### LINKED ISSUE #7371: junctionsAsDirs requires Syncthing restart to trigger
1. Add a new folder to Syncthing under Windows.
2. Enable `junctionsAsDirs` for the newly added folder.
3. Create a new junction in the folder using `mklink /j`.
4. Rescan the folder.
5. The junction has not been detected.

It seems that enabling `junctionsAsDirs` requires Syncthing to restart in order to actually trigger. After the restart, new junctions continue to be properly detected, so the issue seems to be limited only to enabling the option for the very first time.

Is this expected? If yes, then I will add this information to the Docs. If not, then I think that this should classify as a bug.

Tested with `main` (`v1.14.0-rc.1.dev.133.gfb078068`).

imsodin:
This adds an url query like `id` internal member to `fs.Options`. The case-fs then creates a query-like options-string and uses that for caching in addition to the path.

### DIFF
--- a/lib/fs/basicfs.go
+++ b/lib/fs/basicfs.go
@@ -27,12 +27,15 @@ var (
 )
 
 func WithJunctionsAsDirs() Option {
-	return func(fs Filesystem) {
-		if basic, ok := fs.(*BasicFilesystem); !ok {
-			l.Warnln("WithJunctionsAsDirs must only be used with FilesystemTypeBasic")
-		} else {
-			basic.junctionsAsDirs = true
-		}
+	return Option{
+		apply: func(fs Filesystem) {
+			if basic, ok := fs.(*BasicFilesystem); !ok {
+				l.Warnln("WithJunctionsAsDirs must only be used with FilesystemTypeBasic")
+			} else {
+				basic.junctionsAsDirs = true
+			}
+		},
+		id: "junctionsAsDirs=true",
 	}
 }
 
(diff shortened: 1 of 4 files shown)


lib/fs/basicfs.go
```diff
@@ -27,12 +27,15 @@ var (
...
-		}
+	return Option{
+		apply: func(fs Filesystem) {
+			if basic, ok := fs.(*BasicFilesystem); !ok {
+				l.Warnln("WithJunctionsAsDirs must only be used with FilesystemTypeBasic")
+			} else {
+				basic.junctionsAsDirs = true
+			}
+		},
+		id: "junctionsAsDirs=true",
```
AudriusButkevicius, line 38: You don't really need the =true part, the presence of the value on the url is already "true'ish"

AudriusButkevicius (approved):

calmh:
I was thinking a file system would represent its options as parameters in the returned URI(), in which case the original cache implementation would have worked. Perhaps that means we must be able to parse options from the uri as well?

imsodin:
> Perhaps that means we must be able to parse options from the uri as well?

I didn't want to get into that: On the one hand it's an intrusive change (config change) and it is problematic in fringe cases, as the "query-control-characters! `?` and `&` are valid in paths already. I also don't like the concept in general really: It means we need to take care to use `URI` where options are necessary, but not where it reaches users. There we need to somehow transform between user-friendly options and the uri style queries. That's why I made the option ids part internal to the `fs` package.

calmh (approved):
OK

PR #7386, opened 2021-02-20
gui: Handle info labels that are longer than available space (fixes #944)

### LINKED ISSUE #944: Handle info labels that are longer than available space
@calmh: Hello, thanks for your work with syncthing, and sorry for my bad english!

I use a Samsung Galaxy S4 active android 4.2.2 with syncthing 0.5.7.

The handy has two repositories s4-phone and s4-card)
If i start the gui you can see the display with some overwritten line.

2014_11_8_0_48_43_265_pm

May be a shorter description or a line break could solve the problem.
(For me it is not so important, but i want to show it to you)

Thanks for your work

tomasz1986:
Apply to table headers the same code as already used for table data.
This way, the headers will be either pushed to the next line, or cut
with an ellipsis if the single word is too long.

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -120,7 +120,8 @@ table.table-dynamic {
     word-break: break-all;
 }
 
-table.table-condensed td {
+table.table-condensed td,
+table.table-condensed th {
     overflow: hidden;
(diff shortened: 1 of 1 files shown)


tomasz1986:
Before:



After:

calmh (approved):

PR #7388, opened 2021-02-20
gui: Remove download animation from Out of Sync Items modal (fixes #3322)

### LINKED ISSUE #3322: GUI "Out of sync" modal uses massive amounts of CPU in browser
I don't know what's going on here. Is it the animated progress bars? Polling and parsing updates? When showing this:

screen shot 2016-06-17 at 08 53 05

That browser tab uses almost a full CPU core (constantly):

screen shot 2016-06-17 at 08 53 18

After closing the modal, usage is reasonable:

screen shot 2016-06-17 at 08 53 35

screen shot 2016-06-17 at 08 53 37

v0.13.7, Safari/Mac latest

tomasz1986:
Remove the animation due to its excessive CPU usage, especially when a
large number of files is being downloaded and listed at the same time.
Also, remove the stripes, as they serve no purpose in the now-static
progress bar.

### DIFF
--- a/gui/default/syncthing/transfer/neededFilesModalView.html
+++ b/gui/default/syncthing/transfer/neededFilesModalView.html
@@ -7,7 +7,7 @@
       <div class="progress-bar" style="width: 20%"><span translate class="show">Copied from original</span></div>
       <div class="progress-bar progress-bar-info" style="width: 20%"><span translate class="show">Copied from elsewhere</span></div>
       <div class="progress-bar progress-bar-warning" style="width: 20%"><span translate class="show">Downloaded</span></div>
-      <div class="progress-bar progress-bar-danger progress-bar-striped active" style="width: 20%"><span translate class="show">Downloading</span></div>
+      <div class="progress-bar progress-bar-danger" style="width: 20%"><span translate class="show">Downloading</span></div>
     </div>
 
     <hr />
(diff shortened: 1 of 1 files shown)


tomasz1986:
No screenshots, as I would have to record a video/gif showing the animations, but this basically makes the download progress bar solid red.

AudriusButkevicius:
I disagree with the change in general, animating a progress bar is cheap, and only a problem for people on prehistoric hardware.

imsodin:
I tend to agree with removing: Firstly because I dislike 99.9% of all web-animation (it's usually somewhere on a spectrum between pointless to detrimental/distracting), and secondly because I apparently run pre-historic hardware: I recently added a styles to disable a tiny animation on a site I have open continuously because it triggered the fan (I recentlyish upgraded to a thinkpad t480, which is super-flashy new for me xD ).

tomasz1986:
I personally disable animations and transitions with custom CSS on all websites (because many of them make me physically dizzy), but regardless of any individual reasons, this particular modal uses a lot of CPU (comparatively) even without the animation, e.g. due to the constantly changing file list. Because of this, I would say that that the more we can reduce the CPU usage here, the better.

Although, to tell the truth, I will admit that I cannot really see a noticeable difference on Ryzen G4350, which I use for these tests.

calmh (approved):

calmh:
Not taking the trouble to try to reproduce, the change is trivial and the functionality equivalent with or without animation so 🤷

PR #7389, opened 2021-02-20
lib/fs: Expose fs option on interface (fixes #7385, ref #7381)

### LINKED ISSUE #7385: infinite filesystem recursion due to missing file id
I'm trying to use syncthing with a folder exposed by cryptomator. 

The problem is that I'm getting for all folders / files "infinite filesystem recursion" as the virtual filesystem of cryptomator seems to not generate a unique file id for each file.

As example:
X:>fsutil file queryfileid keyfiles
File ID is 0x00000000000000000000000000000000

X:>fsutil file queryfileid pictures
File ID is 0x00000000000000000000000000000000

As far as I have seen syncthing relies on the unique file id for internal purposes. So far so good. However, I've created an issue on the cryptomator side (see https://github.com/cryptomator/cryptomator/issues/1540), but as I understood having a unique file id (provided by the file systems) is nothing you can rely on, according to the cryptomator developers. 

I'm not sure if this is true or not, but it seems to be true, as e.g. almost all tools can work with a cryptomator drive without issues - but syncthing. 

As syncthing is a greate / cool tool, I'd like to bring this to your attention.

Syncthing in use: V1.13.1, Windows 10 20H2.

imsodin:
Fixes #7385 by only checking for infinite recursion if the junctions-as-dirs option is enabled.

Passing around options like I did in #7381 is tedious/repetitive. I thus changed the way options work from passing function `With...` style arguments to const/bitmask options. This allows knowing what options are active on any given filesystem, thus we don't need to pass options to the case- and walk-fs (and the case-fs can directly use option value for the key instead of constructing some arbitrary string).

One choice I am unsure about: I kept the variable length `opts ...FilesystemOptions` argument, even though multiple options can be passed by bitwise-or-ing options. I left that such that all `NewFilesystem` calls without options don't need to be changed to have an additional `, 0` (or something like `0, fs.OptionNone`) argument.

### DIFF
--- a/lib/fs/filesystem.go
+++ b/lib/fs/filesystem.go
@@ -179,6 +179,12 @@ var IsPermission = os.IsPermission
 // IsPathSeparator is the equivalent of os.IsPathSeparator
 var IsPathSeparator = os.IsPathSeparator
 
+// Option modifies a filesystem at creation. An option might be specific
+// to a filesystem-type.
+//
+// ID is used to identify options with the same effect, i.e. must be different
+// for options with different effects. Meaning if an option has parameters, a
+// representation of those must be part of the returned string.
 type Option interface {
 	ID() string
 	apply(Filesystem)
--- a/lib/fs/filesystem.go
+++ b/lib/fs/filesystem.go
@@ -47,7 +47,7 @@ type Filesystem interface {
 	Usage(name string) (Usage, error)
 	Type() FilesystemType
 	URI() string
-	Options() FilesystemOptions
+	Options() []Option
 	SameFile(fi1, fi2 FileInfo) bool
 }
 
@@ -179,16 +179,12 @@ var IsPermission = os.IsPermission
 // IsPathSeparator is the equivalent of os.IsPathSeparator
 var IsPathSeparator = os.IsPathSeparator
 
-type FilesystemOptions uint
+type Option interface {
+	ID() string
+	apply(Filesystem)
+}
 
-const (
-	OptionJunctionsAsDirs FilesystemOptions = 1 << iota
-)
-
-// NewFilesystem creates a filesystem of given fsType with its root at uri.
-// Options can be passed through multiple optional options. It is also possible
-// to pass a combination of FilesystemOptions in a single argument.
-func NewFilesystem(fsType FilesystemType, uri string, opts ...FilesystemOptions) Filesystem {
+func NewFilesystem(fsType FilesystemType, uri string, opts ...Option) Filesystem {
 	var fs Filesystem
 	switch fsType {
 	case FilesystemTypeBasic:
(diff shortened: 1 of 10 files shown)


calmh:
With this you've suddenly limited options to being booleans, and having the same meaning between different fs implementations. I'm not a fan. Consider the fakefs for example that take several numerical options. Casefs also could/should take numerical options to control its caching parameters. 

I can see it being useful to ask a filesysten about its options though, and it seems like we'd need something like a string representation there, probably looking an awful lot like a query string.

imsodin:
I incorporated the latest feedback: Options are now distinct types implementing an interface, and the options used when creating a filesystem can be obtained with the new filesystem method `Options() []Option`:

```golang
type Option interface {
	ID() string
	apply(Filesystem)
}
```

lib/fs/filesystem.go
```diff
@@ -178,9 +179,9 @@ var IsPermission = os.IsPermission
 // IsPathSeparator is the equivalent of os.IsPathSeparator
 var IsPathSeparator = os.IsPathSeparator
 
-type Option struct {
-	apply func(Filesystem)
-	id    string
+type Option interface {
+	ID() string
```
calmh, line 183:
```suggestion
	String() string
```

Might as well be a stringer? Feels more like a string representation of an option, especially if there's a value, than a unique ID of some sort.

calmh (commented):
Sounds good

lib/fs/filesystem.go
imsodin: It has to be IDish, as in every option with the same `ID` should have the same effect. That's required for the case-fs caching. I imagined that if an option has a parameter, it needs to be a query-like string (`name=param` if only a single parameter is allowed, or `name?param1=value1&param2=value2`). Implementing `Stringer` is anyways always a good thing, will do, I'd still like to keep `ID` to signify in the method name that it is more than just an "informal" string representation. I added a comment explaining the purpose of `ID` to the `Option` interface.  
Does that make sense? Or maybe be more specific and have both a `Name() string` and `Params() uri.Values` method on `Options` - feels a bit over-engineered though.

calmh: Right, that's where I was thinking that intuitively to me for some hypothetical optional like `lstatCacheTime=30s` the "ID" would be `statCacheTime` and the string representation would be `lstatCacheTime=30s` and it's the latter we would use when caching because two options with the same ID but with different values are different. But we don't have that today and the distinction doesn't exist yet so it's all hypothetical and based on feeling. :)

imsodin: `String` it is - and if it ever gets non-hypothetical, we can reopen the discussion (or not :) ).

calmh (approved):

PR #7390, opened 2021-02-20
gui: Fix versioning date filter popup position (fixes #4755)

### LINKED ISSUE #4755: Date selector dialog on restore option partially hidden
**- What happened**

Clicked on **Versions** > **Filter** by date and tried to set a time range.  This is what I saw.

screenshot from 2018-02-13 10-59-32

The top of the dialog box is obscured by the address bar. This also happens in chromium.

**- What operating system, operating system version and version of Syncthing you are running**

Syncthing v0.14.45-rc.1, FreeBSD (64 bit) 
Firefox 58.0.2 (64-bit)
Chromium Version 64.0.3282.140 (Official Build) Built on Ubuntu , running on Ubuntu 17.10 (64-bit)

### LINKED ISSUE #5408: Selection of time in versions GUI not possible without editing the string inside the textfield
Steps to reproduce:
- Create a folder with versioning 
- Click on the folder and click the versions button
- click on the date textfield
- select Custom Range
- click on one of the drop down selectors for time
- The list of numbers appears and disappears a very short time after it --> no time for choose an other time

I'm using the latest Firefox and ST v0.14.54 on Linux

tomasz1986:
Change the position and the parent element of the calendar popup, so
that it spawns inside the Restore Versions modal. This way, even on
small phone screens it is possible to scroll down to view the whole
calendar.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -2378,12 +2378,13 @@ angular.module('syncthing.core')
                             });
 
                             $("#restoreVersionDateRange").daterangepicker({
+                                parentEl: '#restoreVersions',
                                 timePicker: true,
                                 timePicker24Hour: true,
                                 timePickerSeconds: true,
                                 autoUpdateInput: true,
                                 opens: "left",
-                                drops: "up",
+                                drops: "down",
                                 startDate: minDate,
                                 endDate: maxDate,
                                 minDate: minDate,
--- a/gui/default/untrusted/syncthing/core/syncthingController.js
+++ b/gui/default/untrusted/syncthing/core/syncthingController.js
@@ -2418,12 +2418,13 @@ angular.module('syncthing.core')
                             });
 
                             $("#restoreVersionDateRange").daterangepicker({
+                                parentEl: '#restoreVersions',
                                 timePicker: true,
                                 timePicker24Hour: true,
                                 timePickerSeconds: true,
                                 autoUpdateInput: true,
                                 opens: "left",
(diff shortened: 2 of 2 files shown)


tomasz1986:
Before:

See the linked issue.

After:

AudriusButkevicius:
I don't think this actually fixes it?
If the list is large, it will just be hidden at the bottom instead of the top.

tomasz1986:
Yeah, I am sorry for the noise. This is going to require more time to be fixed properly. I will write back once the PR is ready.

tomasz1986:
I have fixed it. It took me some time to figure this out, but the fix itself is super simple.

This is how it looks on small screens.

https://user-images.githubusercontent.com/5626656/108623558-f3377300-7482-11eb-8917-96c8b4e1332f.mp4

AudriusButkevicius:
Does it still work if you keep the direction as up?
Reason I ask is because most of the time the list is not going to be empty, so the thing opening downwards will always require to scroll.

tomasz1986:
Unfortunately, not without additional tweaks. I will try to have another look at it maybe tomorrow.

tomasz1986:
I have looked for an obvious way to make it work, but it seems that some JS/CSS hackery would be required to achieve this.

In the case #7395 is accepted, would it be acceptable to have the calendar pop up at the bottom? If the file list is contained like that, then the calendar should always be visible on larger screens. It will require scrolling on mobile, but I would say that it is still way better than the current state, i.e. being cut off and unusable.

st-review:
:robot: *beep boop*

PR #7395, opened 2021-02-22
gui: Tweak the Restore Versions modal for better usability

tomasz1986:
Add various tweaks to the Restore Versions modal in order to make it
more usable in general, and also to prevent the content from breaking
when used with long filenames and on small resolutions.

The specific changes are as follows.

1. Contain the file list in a resizable box, so that only 5.5 items are
   visible initially, while the rest is available through scrolling.
   This prevents the whole body from being pushed down by the list, and
   also keeps the filter and confirmation buttons visible on the screen.
2. Forcefully break long filenames with no spaces, so that they always
   fit on the screen.
3. Change the position of the restore buttons' menu popup to the left,
   so that it does not go out of the screen on small resolutions.
4. Widen the filter input fields, so that there is more space for the
   text, and especially the data range fits without being cut off.

### Testing

Tested in Chromium, Firefox, and IE11*.

\* The browser does not support `resize: both`, so the container is not user-resizable, but it is still usable nevertheless.

### Screenshots

Before:

After:

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -304,6 +304,73 @@ ul.three-columns li, ul.two-columns li {
     z-index: 980;
 }
 
+/**
+ * Fancytree tweaks
+ */
+
+table.fancytree-ext-table {
+  width: 100%
+}
+
+span.fancytree-expander,
+span.fancytree-icon {
+    vertical-align: middle;
+}
+
+/* Break long names not to push the restore buttons off the screen. */
+span.fancytree-title {
+    word-break: break-all;
+}
+
+/**
+ * Restore Versions modal tweaks
+ */
+
+/* Contain the file list in a fix-sized box with a scrollbar. By
+   default, display 5.5 items at once (depends on the browser, font, and
+   DPI). The container can be resized by the user at will. */
+#restoreTree-container {
+    height: auto;
+    overflow-y: auto;
+    resize: both;
+    /* Size of a single item, 39px + 12px (.form-control padding). */
+    min-height: 51px;
+    /* Size of 5.5 items, (39px * 5.5) + 6px (.form-control padding). */
+    max-height: 220.5px;
+}
+
+/* Disable the max-height limit set above when resized manually. */
+#restoreTree-container[style*="height"] {
+    max-height: none;
+}
+
+/* The outline is ugly and serves no purpose here. */
+#restoreTree {
+    outline: none;
+}
+
+/* Hide the empty table head, so that it takes no vertical space. */
+#restoreTree > thead {
+    display: none;
+}
+
+/* Reduce the space between the restore button and menu on mobile. */
+#restoreTree .dropdown-toggle {
+    margin-bottom: 0;
+}
+
+/* Move the button to the left, so that it is visible on mobile. */
+#restoreTree .dropdown-menu {
+    left: auto;
+    right: 0;
+}
+
+/* Widen the filter inputs to fit more text, esp. the date range. */
+#restoreVersions .form-group,
+#restoreVersions .form-control {
+    width: 100%;
+}
+
 /** Footer nav on small devices **/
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
Why not min-wheigh max-height?

On my 4k vertical monitor this will look like a joke.

tomasz1986:
> Why not min-wheigh max-height?

How would you do this?

I have tried playing around with different approaches, but the only way to make it work like that was to limit the height of the whole modal, and this is not good, because it breaks when the resolution is too low to fit the whole thing vertically.

> On my 4k vertical monitor this will look like a joke.

I took the inspiration from the Ignore Patterns. The container is roughly the same size (or actually a little bit larger). As with the Ignore Patterns, it is also resizable by dragging from the bottom-right corner.

gui/default/assets/css/overrides.css
```diff
@@ -304,6 +304,73 @@ ul.three-columns li, ul.two-columns li {
...
+ * Restore Versions modal tweaks
+ */
+
+/* Contain the file list in a fix-sized box with a scrollbar. By
+   default, display 5.5 items at once (depends on the browser, font, and
+   DPI). The container can be resized by the user at will. */
+#restoreTree-container {
+    height: auto;
+    overflow-y: auto;
+    resize: both;
```
imsodin, line 335: What's the use-case to resize horizontally? Looks weird when you expand the area beyond its parent horizontally in the video.

tomasz1986: I actually simply followed the same code as the Ignore Patterns use, but this does help on small screens when dealing with super-long filenames. Otherwise, just 1 file can take the whole screen.

imsodin:
Are the problems about the modal size already addressed, or if not do they need to be addressed in this PR? Just asking because I believe the overall UX improvement here is quite good.

tomasz1986:
> Are the problems about the modal size already addressed, or if not do they need to be addressed in this PR? Just asking because I believe the overall UX improvement here is quite good.

They are not addressed, unfortunately. This is how it looks on a vertical 1920x1080 screen.



Not great, although the file list is resizable, so this is not a tragedy. Still, I will perhaps give this another try this week and see if I can add some tweaks without messing around with the HTML too much.

st-review:
:robot: *beep boop*

acolomb:
@tomasz1986 do you still intend to finish this work?

tomasz1986:
> @tomasz1986 do you still intend to finish this work?

This one is tricky. I managed to make it work for different screen media queries before, but it was done only using hardcoded pixels in CSS. It did work, but would require too much maintenance, so I decided not to go that way.

I'm not sure, but I think that the modal should simply resize itself following the vertical screen size, but a) leave the same amount of spacing on both top and bottom, and b) only resize when there is actual content inside it that needs to be scrolled. Otherwise, there is no need for the resize.

If you have an idea how to do this cleanly without hardcoding the pixels, I would be very helpful for any tips 🙂.

tomasz1986:
I have fixed all the issues. The list now fits the whole screen (when needed). I took a bruteforce yet simple approach of calculating the height of all other elements and then reducing the file list height by it. Three different values are applied following the bootstrap media query breakpoints.

image
image
image

Would it be possible to re-open the PR?

acolomb:
Just pushing new commits will reopen it I believe.

tomasz1986:
> Just pushing new commits will reopen it I believe.

I'd pushed the commits already before writing the post, but nothing has changed here. It seems the PR needs to be reopened first for any updates to show up.

acolomb:
I can't reopen either. 

Try to check out the PR head from GitHub and then rebase your new commits onto that. Once reopened, you can still rebase and force push if needed.

imsodin:
Just open a new PR from the same branch. Github says the branch has been force pushed or recreated, and won't reopen because of that. Not sure why that is, but that's not really important when a simple solution is to just re-create the PR.

PR #7407, opened 2021-02-25
all: Fix versioning path handling

imsodin:
The code in GUI uses a wrong key for the versioning path: It should be `fsPath`, but used is `versionsPath`. I changed the fs related options (path and type) to be members of `VersioningConfiguration` directly instead of the "free-form" string map entries. That doesn't prevent any such mixups in JS land, but at least makes it unambiguous in Go (and anything that will be based on the protos :) ).


calmh (approved):

imsodin:
While the bug is old, exposing the versioning path in the folder edit dialog is new to v1.14.0 - that's why it was detected now. Thus I'd release a v1.14.0-rc.3 with this PR plus #7384 (it's a small change and omitting would cause conflicts). Agreed? @AudriusButkevicius @calmh

calmh:
Sure

PR #7410, opened 2021-02-26
gui: Allow to resize command in External Versioning

tomasz1986:
Use textarea instead of input in the command field for the External File
Versioning. This way, it is possible to resize it at will, so that we
can input and view even a very long command in full, without it being
cut off and forcing us to scroll horizontally.

### Testing

Tested in IE 11, Chromium 88, and Firefox ESR 78.7.1.

### Screenshots

Before:

After:

### DIFF
--- a/gui/default/syncthing/folder/editFolderModalView.html
+++ b/gui/default/syncthing/folder/editFolderModalView.html
@@ -140,7 +140,7 @@
           <div class="form-group" ng-if="currentFolder._guiVersioning.selector=='external'" ng-class="{'has-error': folderEditor.externalCommand.$invalid && folderEditor.externalCommand.$dirty}">
             <p translate class="help-block">An external command handles the versioning. It has to remove the file from the shared folder. If the path to the application contains spaces, it should be quoted.</p>
             <label translate for="externalCommand">Command</label>
-            <input name="externalCommand" id="externalCommand" class="form-control" type="text" ng-model="currentFolder._guiVersioning.externalCommand" required="" aria-required="true" />
+            <textarea name="externalCommand" id="externalCommand" class="form-control" rows="1" ng-model="currentFolder._guiVersioning.externalCommand" required="" aria-required="true" />
             <p class="help-block">
               <span translate ng-if="folderEditor.externalCommand.$valid || folderEditor.externalCommand.$pristine">See external versioning help for supported templated command line parameters.</span>
               <span translate ng-if="folderEditor.externalCommand.$error.required && folderEditor.externalCommand.$dirty">The path cannot be blank.</span>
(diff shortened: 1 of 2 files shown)


gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -140,7 +140,7 @@
           <div class="form-group" ng-if="currentFolder._guiVersioning.selector=='external'" ng-class="{'has-error': folderEditor.externalCommand.$invalid && folderEditor.externalCommand.$dirty}">
             <p translate class="help-block">An external command handles the versioning. It has to remove the file from the shared folder. If the path to the application contains spaces, it should be quoted.</p>
             <label translate for="externalCommand">Command</label>
-            <input name="externalCommand" id="externalCommand" class="form-control" type="text" ng-model="currentFolder._guiVersioning.externalCommand" required="" aria-required="true" />
+            <textarea name="externalCommand" id="externalCommand" class="form-control" rows="3" ng-model="currentFolder._guiVersioning.externalCommand" required="" aria-required="true" />
```
imsodin, line 143: Why 3 rows by default?  
Personally I think the most sensible use of this field is to execute a script in a file, not write an entire script. Thus I'd keep it single-lined by default.

tomasz1986: 1 row is really not enough on small screens, even if we just call a file or such.

(Chromium emulated Samsung Galaxy Fold)

image

What I would actually like to do is to simply have the area automatically resize itself following the content. This is possible with some basic vanilla JS, e.g. https://stackoverflow.com/a/24676492. However, I am not familiar enough with the Syncthing's JS structure. If I wanted to add this kind of code, which file exactly should it go to?

imsodin: Not sure this is worth the effort: I doubt anyone writes long commands on a small phone screen. If you do want to try it out, basically all our JS goes into syncthingController.js. Which is obviously not a nice solution, but it is how it is. If it's short enough you might be able to inline it with angularjs interpolation (use `ngChange` or put it inside `{...}`. Or you could take the high road and create some "auto-resizing textarea" directive.

tomasz1986: I will give it a shot. Not necessarily for small screens, but for myself really, as as you can see, I like to put whole scripts there, and right now I have to copy and paste them from a text editor. The goal is basically to have only 1 line displayed by default, and then just autoresize once the content exceeds it (plus having an option to resize manually).

tomasz1986: I have taken a look at a JS solution, but with my limited skills, this will require some time, so what I am going to do is to work on it as a separate commit later.

As for this one though, I would say that it is ready as it is, i.e. just a simple change to make the input field manually resizable.

imsodin (approved):

PR #7412, opened 2021-02-26
Treat *.localhost as valid localhost addresses

wbolster:
This loosens the ‘is this localhost?’ check to include *.localhost host
names.

This allows for clearer (hence better) names to be used in browsers,
e.g. when accessing a remote syncthing instance ‘foo’ using a ssh port
forward, one can use foo.localhost to remind oneself which one is which.
💡 Without these changes, Syncthing shows a ‘Host check error’ when
pointing a browser at http://foo.localhost/, and with these changes, the
interface loads as usual.

The .localhost top level domain is a reserved top-level domain (RFC 2606):

> The ".localhost" TLD has traditionally been statically defined in
> host DNS implementations as having an A record pointing to the
> loop back IP address and is reserved for such use.  Any other use
> would conflict with widely deployed code which assumes this use.
> – https://tools.ietf.org/html/rfc2606

As Wikipedia puts it:

> This allows the use of these names for either documentation purposes
or in local testing scenarios. – https://en.wikipedia.org/wiki/.localhost

On Linux systems, systemd-resolved resolves *.localhost, on purpose:
https://www.freedesktop.org/software/systemd/man/systemd-resolved.service.html

See also #4815, #4816.

### DIFF
--- a/lib/api/api.go
+++ b/lib/api/api.go
@@ -1772,8 +1772,13 @@ func addressIsLocalhost(addr string) bool {
 		// There was no port, so we assume the address was just a hostname
 		host = addr
 	}
-	switch strings.ToLower(host) {
-	case "localhost", "localhost.":
+	host = strings.ToLower(host)
+	switch {
+	case host == "localhost":
+		return true
+	case host == "localhost.":
+		return true
+	case strings.HasSuffix(host, ".localhost"):
 		return true
 	default:
 		ip := net.ParseIP(host)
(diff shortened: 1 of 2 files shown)


wbolster:
fwiw, this should help with the issue described by @bigtonylewis in #4815, which was sort-of addressed in #4816 by @calmh even though [‘this won't solve [...] problem’](https://github.com/syncthing/syncthing/issues/4815#issuecomment-373281380) 😄

calmh (approved):

calmh:
Thanks!

wbolster:
yw. i now realize my change doesn't handle "*localhost." with a trailing dot, while the code explicitly handles 'localhost.". is this worth a follow-up pull request, or would no one care?

calmh:
I doubt it's commonly occurring.

PR #7417, opened 2021-03-01
Package sysctl configuration to raise UDP buffer size on Linux

bt90:
### Purpose

This PR provides a sysctl configuration to increase the maximum receive socket buffer on Linux as suggest by https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size

The configuration gets placed in `/usr/lib/sysctl.d` according to [sysctl.d(5)](https://manpages.ubuntu.com/manpages/focal/en/man5/sysctl.d.5.html). The prefix `30-` causes the file to be applied early as the processing is done in lexicographic order. This should be low enough that configuration files of other packages can still overwrite this value.

I picked the `30-` prefix so that the file can be also used for manual installs in `/etc/sysctl.d` according to the older naming schema in Ubuntu Bionic:

>  Other packages install their files in the 30-*.conf range, to override system defaults.

I also modified the `post-upgrade` script to call `deb-systemd-invoke restart procps.service`

See also #7146

### DIFF
--- a/etc/linux-sysctl/README.md
+++ b/etc/linux-sysctl/README.md
@@ -1,0 +1,21 @@
+sysctl configuration to raise UDP buffer size
+===================
+Installation
+-----------
+**Please note:** When you installed syncthing using the official deb package, you can skip the copying.
+
+Copy the file `30-syncthing.conf` to `/etc/sysctl.d/` (root permissions required).
+
+In a terminal run
+```
+sudo sysctl -q --system
+```
+to apply the sysctl changes.
+
+
+Verification
+----------
+You can verify that the new limit is active using
+```
+sysctl net.core.rmem_max
+```
--- a/script/post-upgrade
+++ b/script/post-upgrade
@@ -1,3 +1,4 @@
 #!/bin/sh
 
+deb-systemd-invoke restart procps.service
 pkill -HUP -x syncthing || true
--- a/script/post-upgrade
+++ b/script/post-upgrade
@@ -1,4 +1,8 @@
 #!/bin/sh
 
-deb-systemd-invoke restart procps.service
+if command -v deb-systemd-invoke > /dev/null; then
+	deb-systemd-invoke restart procps.service
+else
+	sysctl -q --system
+fi
 pkill -HUP -x syncthing || true
--- a/build.go
+++ b/build.go
@@ -115,6 +115,7 @@ var targets = map[string]target{
 			{src: "etc/linux-systemd/system/syncthing@.service", dst: "deb/lib/systemd/system/syncthing@.service", perm: 0644},
 			{src: "etc/linux-systemd/system/syncthing-resume.service", dst: "deb/lib/systemd/system/syncthing-resume.service", perm: 0644},
 			{src: "etc/linux-systemd/user/syncthing.service", dst: "deb/usr/lib/systemd/user/syncthing.service", perm: 0644},
+			{src: "etc/linux-sysctl/30-syncthing.conf", dst: "deb/usr/lib/sysctl.d/30-syncthing.conf", perm: 0644},
 			{src: "etc/firewall-ufw/syncthing", dst: "deb/etc/ufw/applications.d/syncthing", perm: 0644},
 			{src: "etc/linux-desktop/syncthing-start.desktop", dst: "deb/usr/share/applications/syncthing-start.desktop", perm: 0644},
 			{src: "etc/linux-desktop/syncthing-ui.desktop", dst: "deb/usr/share/applications/syncthing-ui.desktop", perm: 0644},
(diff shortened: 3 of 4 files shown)


etc/linux-sysctl/README.md
```diff
@@ -0,0 +1,21 @@
...
+===================
+Installation
+-----------
+**Please note:** When you installed syncthing using the official deb package, you can skip the copying.
+
+Copy the file `30-syncthing.conf` to `/etc/sysctl.d/` (root permissions required).
+
+In a terminal run
+```
+sudo systemctl restart procps
```
calmh, line 11: Do we need to be this systemd specific here, instead of for example `sysctl -q --system` to load the things directly?

calmh (commented):
Is the default on linux low for a reason, such that we for example run the risk of bricking someones RPi due to this change?

etc/linux-sysctl/README.md
bt90: Works both. I can change it.

bt90:
> Is the default on linux low for a reason, such that we for example run the risk of bricking someones RPi due to this change?

I couldn't really find any reliable source which would explain the reasoning behind the default size but ` net.core.rmem_max` controls the *maximum* size of the receiver buffer. If the application doesn't request a higher buffer size then `net.core.rmem_default` applies which is 208 kiB.

Tuned seems to do something similar for Oracle RDBMS: https://github.com/redhat-performance/tuned/blob/master/profiles/oracle/tuned.conf

bt90:
Relevant performance test: https://github.com/lucas-clemente/quic-go/issues/2586#issuecomment-639247615

script/post-upgrade
```diff
@@ -1,3 +1,4 @@
 #!/bin/sh
 
+deb-systemd-invoke restart procps.service
```
imsodin, line 3: Could also use `sysctl` directly.

imsodin (commented):
> Edit: quic-go seems to request only 2048 kiB judging from the log provided here: #7146 (comment) so we could lower it a bit.

I wouldn't base anything on that, it's a very specific and uncontrolled sample. Unless there's reasons to believe increasing the limit is problematic, I'd just go with what quic suggest.

script/post-upgrade
otbutz: Just following best practices. `/etc/sysctl.d/README.sysctl`:

> After making any changes, please run "service procps reload" (or, from a Debian package maintainer script "deb-systemd-invoke restart procps.service").

But i can check if `deb-systemd-invoke` is available and fallback to `sysctl`

imsodin: Sounds good. I was thinking about people e.g. on devuan or some other systemd-less system. Not sure whether they have a drop-in for `deb-systemd-invoke` to do the non-systemd equivalent (I myself am blissfully using systemd without really understanding all the fuss - probably got lucky in having been exposed to linux too late to form strong opinions there :) ).

bt90: Done. But it seems that even the systemd hating zealots from Devuan provide deb-systemd-invoke 😛

bt90: The check follows Debians best packing practices guide: https://www.debian.org/doc/manuals/developers-reference/best-pkging-practices.html#best-practices-for-maintainer-scripts

bt90:
Installed the deb generated by the buildserver on my Ubuntu 20.04 server and the sysctl setting is getting properly applied:

```
systemd[1]: Starting Apply Kernel Variables...
```

```
$ sysctl net.core.rmem_max
net.core.rmem_max = 2097152
```

But i still get this log output:
> 2021/03/02 20:30:43 connection doesn't allow setting of receive buffer size. See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.

Anybody willing to give it a spin?

imsodin:
> Anybody willing to give it a spin?

Same here: I applied the change manually and restarted Syncthing and got the same message again.

The problem is that we wrap the udp conn for stun, and that wrapped object fails here:  
https://github.com/lucas-clemente/quic-go/blob/dd9f8e4a2b0f231f88ed7ed7a199ccd6a34e01af/packet_handler_map.go#L59  
That can be fixed by `lib/stun` working with an extended interface based on `net.PacketConn` that also has `SetReadBuffer(bytes int) error`.

bt90:
Maybe also relevant: https://github.com/lucas-clemente/quic-go/issues/2607

Would probably be a good idea to pass a connection which might later profit from future performance optimisations.

imsodin:
> Would probably be a good idea to pass a connection which might later profit from future performance optimisations.

That's basically what I proposed by extending the connection interface with the relevant method(s). Do you want to look into doing that?

bt90:
Sorry, i have zero experience with go 😅

imsodin:
Together with https://github.com/syncthing/syncthing/pull/7432 I don't get the error anymore.

calmh (approved):

bt90:
@calmh is this changelog worthy?

calmh:
It would have been if it had closed an issue... (those are what goes into the changelog)

PR #7424, opened 2021-03-03
lib: Return error from db.FileSet.Snapshot (fixes #7419, ref #5907)

### LINKED ISSUE #7419: panic "Snapshot(): database is closed" on start after wiping database
I had this panic today when starting Syncthing after wiping the database. By this I mean manually deleting `index-v0.14.0.db` after completely shutting Syncthing down, and then trying to start it again. The program still probably restarted itself after the panic and continued to work, and I only noticed the log many hours later.

```
[7LDIB] 2021/03/02 15:37:18 WARNING: Fatal error: ccmsc-zsmeh Snapshot(): database is closed
Panic at 2021-03-02T15:37:19+09:00
panic: database is closed

goroutine 150 [running]:
github.com/syncthing/syncthing/lib/db.fatalError(0x1b4ca40, 0x20f50f4, 0x160e1280, 0x16, 0x128c01e0)
	github.com/syncthing/syncthing/lib/db/set.go:508 +0x14b
github.com/syncthing/syncthing/lib/db.(*FileSet).Snapshot(0x14fa8720, 0x187459b0)
	github.com/syncthing/syncthing/lib/db/set.go:159 +0x220
github.com/syncthing/syncthing/lib/model.(*folder).pull(0x129ad340, 0x12c9bd00)
	github.com/syncthing/syncthing/lib/model/folder.go:333 +0x10d
github.com/syncthing/syncthing/lib/model.(*folder).Serve(0x129ad340, 0x1b56460, 0x150e26a0, 0x0, 0x0)
	github.com/syncthing/syncthing/lib/model/folder.go:173 +0x9a3
github.com/thejerf/suture/v4.(*Supervisor).runService.func2(0x128f7bc0, 0x6, 0x332ab610, 0x129ad340, 0x1b56460, 0x150e26a0, 0x150dca30, 0x150ea400)
	github.com/thejerf/suture/v4@v4.0.0/supervisor.go:546 +0x5f
created by github.com/thejerf/suture/v4.(*Supervisor).runService
	github.com/thejerf/suture/v4@v4.0.0/supervisor.go:534 +0x12a
```

[panic-20210302-153719.log](https://github.com/syncthing/syncthing/files/6069916/panic-20210302-153719.log)

Syncthing v1.13.x (self-compiled with a few custom commits, see https://github.com/tomasz1986/syncthing/commits/tomasz86-v1.13.0)

imsodin:
### Purpose

Another step towards better database error handling (#5907), fixing #7419: Now `db.FileSet.Snapshot` returns an error instead of panicking.  
Most of the diff is boilerplate of adding error returns all over the place. One larger change regards folder error handling: Previously we did not bubble errors consistently (as in already existing errors unrelated to this PR). Now we do need errors in `Serve` to shut the service down in case of fatal errors. I thus opted to bubble "folder-level" errors (as opposed to item level errors, e.g. errors scanning/pulling an item) up to the main loop in the service. That means the folder state is set with any error that reaches that point and isn't fatal. 

### Testing

Existing unit tests pass and the UI still comes up - haven't done any manual testing beyond that.

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -178,7 +178,7 @@ func (f *folder) Serve(ctx context.Context) error {
 		case <-f.pullFailTimer.C:
 			var success bool
 			success, err = f.pull()
-			if err == nil && success && f.pullPause < 60*f.pullBasePause() {
+			if (err != nil || !success) && f.pullPause < 60*f.pullBasePause() {
 				// Back off from retrying to pull
 				f.pullPause *= 2
 			}
@@ -396,7 +396,7 @@ func (f *folder) pull() (success bool, err error) {
 
 	success, err = f.puller.pull()
 
-	if success && err != nil {
+	if success && err == nil {
 		return true, nil
 	}
 
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -27,6 +27,7 @@ import (
 	"github.com/syncthing/syncthing/lib/protocol"
 	"github.com/syncthing/syncthing/lib/scanner"
 	"github.com/syncthing/syncthing/lib/stats"
+	"github.com/syncthing/syncthing/lib/svcutil"
 	"github.com/syncthing/syncthing/lib/sync"
 	"github.com/syncthing/syncthing/lib/util"
 	"github.com/syncthing/syncthing/lib/versioner"
@@ -87,7 +88,7 @@ type syncRequest struct {
 }
 
 type puller interface {
-	pull() bool // true when successful and should not be retried
+	pull() (bool, error) // true when successful and should not be retried
 }
 
 func newFolder(model *model, fset *db.FileSet, ignores *ignore.Matcher, cfg config.FolderConfiguration, evLogger events.Logger, ioLimiter *byteSemaphore, ver versioner.Versioner) folder {
@@ -164,16 +165,20 @@ func (f *folder) Serve(ctx context.Context) error {
 	initialCompleted := f.initialScanFinished
 
 	for {
+		var err error
+
 		select {
 		case <-f.ctx.Done():
 			close(f.done)
 			return nil
 
 		case <-f.pullScheduled:
-			f.pull()
+			_, err = f.pull()
 
 		case <-f.pullFailTimer.C:
-			if !f.pull() && f.pullPause < 60*f.pullBasePause() {
+			var success bool
+			success, err = f.pull()
+			if err == nil && success && f.pullPause < 60*f.pullBasePause() {
 				// Back off from retrying to pull
 				f.pullPause *= 2
 			}
@@ -181,18 +186,19 @@ func (f *folder) Serve(ctx context.Context) error {
 		case <-initialCompleted:
 			// Initial scan has completed, we should do a pull
 			initialCompleted = nil // never hit this case again
-			f.pull()
+			_, err = f.pull()
 
 		case <-f.forcedRescanRequested:
-			f.handleForcedRescans()
+			err = f.handleForcedRescans()
 
 		case <-f.scanTimer.C:
 			l.Debugln(f, "Scanning due to timer")
-			f.scanTimerFired()
+			err = f.scanTimerFired()
 
 		case req := <-f.doInSyncChan:
 			l.Debugln(f, "Running something due to request")
-			req.err <- req.fn()
+			err = req.fn()
+			req.err <- err
 
 		case next := <-f.scanDelay:
 			l.Debugln(f, "Delaying scan")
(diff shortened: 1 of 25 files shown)


lib/model/folder.go
```diff
@@ -164,52 +165,64 @@ func (f *folder) Serve(ctx context.Context) error {
...
 
 		case <-f.pullScheduled:
-			f.pull()
+			_, err = f.pull()
 
 		case <-f.pullFailTimer.C:
-			if !f.pull() && f.pullPause < 60*f.pullBasePause() {
+			var success bool
+			success, err = f.pull()
+			if err == nil && success && f.pullPause < 60*f.pullBasePause() {
```
calmh, line 181: This inverts the condition on `success` -- we shouldn't back off when pull was successfull

lib/model/folder_sendrecv.go
```diff
@@ -582,7 +589,7 @@ func (f *sendReceiveFolder) handleDir(file protocol.FileInfo, snap *db.Snapshot,
 	// that don't result in a conflict.
 	case err == nil && !info.IsDir():
 		// Check that it is what we have in the database.
-		curFile, hasCurFile := f.model.CurrentFolderFile(f.folderID, file.Name)
```
calmh, line 585: This (passing around snap) seems unrelated. An oversight from long ago?

lib/model/model.go
```diff
@@ -891,21 +891,29 @@ func (m *model) Completion(device protocol.DeviceID, folder string) FolderComple
...
-	return comp
+	return comp, nil
 }
 
-func (m *model) folderCompletion(device protocol.DeviceID, folder string) FolderCompletion {
+func (m *model) folderCompletion(device protocol.DeviceID, folder string) (FolderCompletion, error) {
 	m.fmut.RLock()
-	rf, ok := m.folderFiles[folder]
+	err := m.checkFolderRunningLocked(folder)
+	rf := m.folderFiles[folder]
```
calmh, line 907: This looks like it could still be something that doesn't exist (rf becomes nil)?

lib/model/model.go
```diff
@@ -2675,17 +2706,25 @@ func (m *model) Availability(folder string, file protocol.FileInfo, block protoc
...
+	snap, err := fs.Snapshot()
+	if err != nil {
+		return nil, err
+	}
 	defer snap.Release()
+
+	return m.availabilityWithInfo(cfg, snap, file, block), nil
+}
+
+func (m *model) availabilityWithInfo(cfg config.FolderConfiguration, snap *db.Snapshot, file protocol.FileInfo, block protocol.BlockInfo) []Availability {
```
calmh, line 2721: `...WithSnapshot` seems more specific (or maybe even *In*Snapshot), or is there some other info we're looking for here?

lib/model/folder.go
```diff
@@ -374,23 +389,23 @@ func (f *folder) pull() (success bool) {
...
+		return false, err
 	}
 
-	success = f.puller.pull()
+	success, err = f.puller.pull()
 
-	if success {
-		return true
+	if success && err != nil {
+		return true, nil
```
greatroar, line 400: Shouldn't this be `success && err == nil`?

lib/model/folder.go
```diff
@@ -341,25 +357,24 @@ func (f *folder) pull() (success bool) {
...
 
 	// Send only folder doesn't do any io, it only checks for out-of-sync
 	// items that differ in metadata and updates those.
 	if f.Type != config.FolderTypeSendOnly {
 		f.setState(FolderSyncWaiting)
 
 		if err := f.ioLimiter.takeWithContext(f.ctx, 1); err != nil {
-			f.setError(err)
-			return true
+			return true, err
```
greatroar, line 377: This is the only instance of true + non-nil error being returned. Isn't it easier to drop the bool return value and check for timeout/context canceled in the retrying logic?

lib/model/folder_sendrecv.go
imsodin: That's intentional `model.CurrentFolderFile` now returns an error, which I don't want to have to handle here, because the folder might be missing. Plus it needlessly creates a new snapshot (i.e. yes, the change would also make sense without this PR)..

lib/model/model.go
imsodin: When `checkFolderRunning` returns `nil`, it exists for sure. We do the same in other places.

lib/model/model.go
imsodin: The unspecific "info" qualifier refers to already having the folder config and snapshot instead of needing to get them from the model, and `WithConfigAndSnapshot` is a mouthful.

lib/model/folder.go
imsodin: Also in the end: A pull can result in no success (files couldn't be synced) without an error.

lib/model/model.go
calmh: Well... I'd still go with something like `snapshotAvailability` and skip mentioning the config, at least it says more that it acts on a point in time than just "with info" which is sort of saying nothing. :)

lib/model/folder.go
greatroar: Oh right, in line 226. Too bad.

calmh (approved):

PR #7435, opened 2021-03-05
lib/api: Sanitize names used in certificates (fixes #7434)

### LINKED ISSUE #7434: Non-DNS-like computer name fails creation of HTTPS certificate
```
[4HCF3] 00:45:04 WARNING: Failed starting API: create cert: x509: "Vicenç-PC" cannot be encoded as an IA5String
```

No API means startup failure, so Syncthing shuts down.

calmh:


lib/api/api.go
```diff
@@ -153,6 +156,10 @@ func (s *service) getListener(guiCfg config.GUIConfiguration) (net.Listener, err
 		if err != nil {
 			name = s.tlsDefaultCommonName
 		}
+		name, err = sanitizedHostname(name)
```
otbutz, line 159: Wouldn't punnycode encoding be easier?

calmh: Maybe, maybe not? We'd still need a sanitization step beforehand as punycode/IDN deals with well formed (but international) domain names, and there's nothing to say that our input is well formed, and the result would look more like crap in case someone actually looked at it. 🤷

imsodin (approved):

lib/api/api.go
```diff
@@ -1860,3 +1867,32 @@ func errorString(err error) *string {
 	}
 	return nil
 }
+
+// sanitizedHostname returns the given name in a suitable form for use as
+// the common name in a certificate, or an error.
+func sanitizedHostname(name string) (string, error) {
```
greatroar, line 1873: This version is doing some unnecessary work and the error return from transform.String isn't handled. The following is shorter and passes all your tests:

```go
// sanitizedHostname returns the given name in a suitable form for use as the
// common name in a certificate, or the empty string if no such form can be
// constructed.
func sanitizedHostname(name string) string {
	t := transform.Chain(
		// Split runes with diacritics into base character and mark.
		norm.NFD,
		// Leave only [A-Za-z0-9-.].
		runes.Remove(runes.Predicate(func(r rune) bool {
			return r > unicode.MaxASCII ||
				!unicode.IsLetter(r) && !unicode.IsNumber(r) &&
					r != '.' && r != '-'
		})),
		runes.Map(unicode.ToLower))

	name, _, err := transform.String(t, name)
	if err != nil {
		return ""
	}

	// Name should not start or end with a dash or dot.
	return strings.Trim(name, "-.")
}
```

I've been so bold as to replace the error return by just returning the empty string, since the actual error is ignored by the caller.

calmh: That's better, I'll steal that version. I'll keep the error return though, even if it's technically unnecessary - it signals to any caller that the returned string may not be valid in all cases, which you might not otherwise know by just looking at the signature.

imsodin (dismissed):
There's a conflict - otherwise lgtm.

PR #7441, opened 2021-03-08
gui: Allow to translate theme names

tomasz1986:
Allow to translate theme names, so that they match the GUI language
rather than being always displayed in English, as it is the case right
now.

### DIFF
--- a/gui/default/syncthing/settings/settingsModalView.html
+++ b/gui/default/syncthing/settings/settingsModalView.html
@@ -164,7 +164,7 @@
               <div class="form-group">
                 <label translate>GUI Theme</label>
                 <select class="form-control" ng-model="tmpGUI.theme" ng-if="themes.length > 1">
-                  <option ng-repeat="theme in themes.sort()" value="{{ theme }}">
+                  <option translate ng-repeat="theme in themes.sort()" value="{{ theme }}">
                     {{ themeName(theme) }}
                   </option>
                 </select>


tomasz1986:
This works if I add theme names to a language json file manually, but I'm really not sure how the interaction with Transifex operates. Is this enough to add a new string to the translation, or does it require a different approach?

calmh:
We have tooling to add things automatically to lang-en.json, but things can also be added there manually and will be allowed to stay there. That file is read by transifex automatically.

I'm not sure this is entirely great as-is though, as you'll get rather context-free strings to translate. And perhaps we shouldn't translate the autofixed "theme name" but rather a key like `theme-key-default` to something human readable like `The default theme`, I don't know...

tomasz1986:
> I'm not sure this is entirely great as-is though, as you'll get rather context-free strings to translate.

That is true, although there are several strings like that already, e.g. "Off", etc. If there is a better way to do this, I'm all for it.

Although, to tell the truth, even if there is a context, it is still often necessary to search for the particular string in the HTML. For instance, I have been going through the Polish translation recently, and in many cases it is impossible to accurately translate things like adjectives, etc. without the related nouns (due to differences in gender, plurality, etc.). Same goes for other English words that don't have direct counterparts, and thus need to be translated differently depending on the usage.

st-review:
:robot: *beep boop*

tomasz1986:
I still think it would be nice to translate the theme names, even if only these short strings. This is because at the moment there are quite a few languages listed on Transifex that are marked as translated in 100% already. However, even in those languages, the only part of the software remaining untranslated are these theme names (apart from debugging options, error logs, and similar).

In contrary to some of the other strings, themes are something that any user is likely going to encounter. Therefore, leaving these untranslated makes the software feel less "complete". I'd say this is especially true when using non-Latin languages, because then these theme names really do stick out.

acolomb:
Note that your approach here causes a translation string `{{ themeName(theme) }}` to be added.  That's not desired nor does it work as intended.

I proposed a different approach based on the `theme-name-...` proposal above in #8006.

PR #7447, opened 2021-03-09
cmd/syncthing: Correct name of HiddenConsole flag (fixes #7446)

### LINKED ISSUE #7446: The no-console flag seems to have stopped working on Windows

It does not.

### Required information

**Syncthing** v1.15.0-rc.1. (I am unsure how I have ended up on a release candidate.)

**OS**: Windows 10

**Browser**: Firefox 86.0

What happens:

```
> C:\Portable_apps\syncthing\syncthing.exe -no-console -no-browser
> syncthing.exe: error: unknown flag --no-console
```

[Edited.]

imsodin:
This fixes #7446 by adding the `name:"no-console"` tag. I also moved it into a separate file with build tags, such that it only shows up on windows as before, because why not.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -139,6 +139,7 @@ var entrypoint struct {
 
 // serveOptions are the options for the `syncthing serve` command.
 type serveOptions struct {
+	buildServeOptions
 	AllowNewerConfig bool   `help:"Allow loading newer than current config version"`
 	Audit            bool   `help:"Write events to audit file"`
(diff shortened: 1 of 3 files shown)


tomasz1986:
Just a side note, but right now Syncthing refuses to start outright if it detects an invalid flag. Would it cause any harm to simply ignore such and continue executing anyway? This would at least allow to keep running the software in such situations without any manual intervention.

calmh (approved):
I was too lazy to do this apparently

PR #7453, opened 2021-03-11
lib/api: Fix body of renamed config/restart-required endpoint (ref #7402)

imsodin:
Just renaming the endpoint (#7402) isn't enough: The body also needs to change, both in key (`configInSync`) and reversing the value.

### DIFF
--- a/lib/api/confighandler.go
+++ b/lib/api/confighandler.go
@@ -51,6 +51,12 @@ func (c *configMuxBuilder) registerConfigInsync(path string) {
 	})
 }
 
+func (c *configMuxBuilder) registerConfigRequiresRestart(path string) {
+	c.HandlerFunc(http.MethodGet, path, func(w http.ResponseWriter, _ *http.Request) {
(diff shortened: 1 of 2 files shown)


lib/api/confighandler.go
```diff
@@ -51,6 +51,12 @@ func (c *configMuxBuilder) registerConfigInsync(path string) {
 	})
 }
 
+func (c *configMuxBuilder) registerConfigRequiresRestart(path string) {
+	c.HandlerFunc(http.MethodGet, path, func(w http.ResponseWriter, _ *http.Request) {
+		sendJSON(w, map[string]bool{"config-requires-restart": c.cfg.RequiresRestart()})
```
calmh, line 56:
```suggestion
		sendJSON(w, map[string]bool{"requiresRestart": c.cfg.RequiresRestart()})
```

I believe we use camelcase (and can increase brevity)?

calmh (approved):

PR #7463, opened 2021-03-12
lib/fs: Check both old and new path when renaming (fixes #7426)

### LINKED ISSUE #7426: Rename with different case triggers delete on android
Today I noticed a file unexpectedly disappearing from a shared directory on my mac. This seems to happen when I rename a file using different casing. Steps to reproduce the issue:

- in a shared folder, using my terminal: `touch FOO.txt`
- `mv FOO.txt foo.txt`
- foo.txt is gone from my mac.

Here's the relevant activity log:

```
Samsung Galaxy S10e Bas | deleted | file | Bas | foo.txt | 2021-03-03 18:57:18
MacBook Bas | deleted | file | Bas | FOO.txt | 2021-03-03 18:57:02
MacBook Bas | modified | file | Bas | foo.txt | 2021-03-03 18:57:02
MacBook Bas | modified | file | Bas | FOO.txt | 2021-03-03 18:56:42
```

It is as if my phone sees the rename as a delete of the original file, then echoes that delete back to its peers.

I'm using Syncthing 1.6.1-1 on mac OS Catalina (10.15.6) and Syncthing v1.13.0 on my Samsung Galaxy S10e.

imsodin:
I can't think of any reason why we shouldn't test both the source and target path for case-conflicts, so I assume that was an oversight. It's not problematic in our usage, because in the puller we always check both source and target files against db (i.e. stat them) before renaming, but it still seems like the right/save thing to do.

### DIFF
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -226,6 +226,9 @@ func (f *caseFilesystem) Rename(oldpath, newpath string) error {
 	if err := f.checkCase(oldpath); err != nil {
 		return err
 	}
+	if err := f.checkCase(newpath); err != nil {
+		return err
+	}
 	if err := f.Filesystem.Rename(oldpath, newpath); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


calmh:
This would prevent a case-only rename from happening at all? (except on case sensitive filesystems)

imsodin:
Good point, tests agree :)  
Then lets cover that scenario, whie still checking both.

calmh (approved):

PR #7468, opened 2021-03-13
gui: Use grey background for disabled options in form-control

tomasz1986:
Disabled options are currently barely distinguishable from enabled
ones. This changes their background to grey, following the Bootstrap
defaults already used for disabled \<select\>.

### Screenshots

Before:

After:

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -437,3 +437,9 @@ ul.three-columns li, ul.two-columns li {
 .form-horizontal {
     margin-bottom: 10px;
 }
+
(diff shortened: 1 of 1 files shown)


imsodin:
Why does bootstrap's disabled css not apply?

tomasz1986:
It is applied to whole disabled `<select>` (e.g. when we disable File Pull Order for Send Only folders), but isn't applied to individual options that are disabled. I think that having the same grey background for the options is better. Otherwise, browser default styles are used, which only make the font colour a little lighter (which really looks the same on a typical TN screen). It makes the user wonder why such an option cannot be clicked.

imsodin (approved):

Issue #7466, opened 2021-03-13
Receive-encrypted fails after a while: file with empty block list


Z3r0byte:
### Does your log mention database corruption?
No

### Include required information
I have a setup of two machines and one untrusted server all running 1.14.0. I have enabled the untrusted feature flag on all machines.  
When syncing between the two trusted machines, everything goes fine and both machines indicate they're up to date. But, when syncing with the untrusted server using the receive-encrypted type, all connections start failing. I am syncing two folders and for both folders these kind of error messages start appearing on the untrusted server:
`[J47TR] 13:37:50 INFO: Connection to BWYS6XV-XXXXXXXXXXXXXXXXXXXXXXX-RP6EJAB at 141.xxx.xxx.xxx:22000-xxx.xxx.xxx.xxx:5068/tcp-client/TLS1.3-TLS_CHACHA20_POLY1305_SHA256 closed: protocol error: index update: "9.syncthing-enc/45/F31BH51JPLO2N7T7MJ03AQ56Q0IIM1CTNE9A38NRNAOIU19NB5LGKF093B6M6D4U6SP5LKTN5NFA9GHN1PK4SK8LFD0K1R9R9S0CQKQQ86T1D7P1HG415P20JD299GNPTJPD54QSP0DKMF9T9D4SRO116INLTOC7DMIKUN1F6AED4D2P27NG": file with empty block list`

I have tried pausing one folder to see if it was an error specific to one folder, but neither are able to fully sync after a couple of hundred MB's. I also tried changing the pull order of the folders, but to no avail. Removing and re adding the folders also did not work. Since the filename is encrypted, I also don't know what folder exactly it is getting stuck on. It does keep giving the same error with the same file name.

If there is any additional information needed, let me know and I'll try my best to give it to you.

imsodin:
I think I know what the issue is: Do you have ignore patterns?  
I believe we drop the invalid flag from file infos sent, which results in a valid looking file without blocks. That then causes this error. Will repro and fix.

And general info on testing the untrusted feature:  
You should run release candidates, i.e. v1.15.0-rc.3 at the time of writing. As this issue shows, it is still in testing and there will be fixes for it, that you'll want to get promptly.

Z3r0byte:
Thanks for your quick response. I indeed do have ignore patterns. I also thought of this and removed those patterns, but the issue persisted (but maybe I didn't give syncthing enough time to rescan). I will switch to rc builds and keep testing from there.

PR #7470, opened 2021-03-13
lib/protocol: Improve messages when an error occurs receiving (ref #7466)

imsodin:
The error messages have no info on which folders are affected, see e.g. #7466 

>  [J47TR] 13:37:50 INFO: Connection to BWYS6XV-XXXXXXXXXXXXXXXXXXXXXXX-RP6EJAB at 141.xxx.xxx.xxx:22000-xxx.xxx.xxx.xxx:5068/tcp-client/TLS1.3-TLS_CHACHA20_POLY1305_SHA256 closed: protocol error: index update: "9.syncthing-enc/45/F31BH51JPLO2N7T7MJ03AQ56Q0IIM1CTNE9A38NRNAOIU19NB5LGKF093B6M6D4U6SP5LKTN5NFA9GHN1PK4SK8LFD0K1R9R9S0CQKQQ86T1D7P1HG415P20JD299GNPTJPD54QSP0DKMF9T9D4SRO116INLTOC7DMIKUN1F6AED4D2P27NG": file with empty block list

And they aren't consistent in whether they include "receiving" and "message". I always added the former (the same method names exist for sending) and removed the latter.

### DIFF
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -1058,40 +1058,29 @@ func (c *rawConnection) lz4Decompress(src []byte) ([]byte, error) {
 	return decoded, nil
 }
 
-type ProtocolError commonError
+type unwrapError interface {
+	error
+	Unwrap() error
+}
+
+type ProtocolError struct {
+	unwrapError
+}
 
 func newProtocolError(err error, msgContext string) *ProtocolError {
-	return (*ProtocolError)(newCommonError(err, msgContext))
-}
-
-func (e *ProtocolError) Error() string {
-	return fmt.Sprintf("protocol error on %v: %v", e.msgContext, e.err)
-}
-
-type HandleError commonError
-
-func newHandleError(err error, msgContext string) *HandleError {
-	return (*HandleError)(newCommonError(err, msgContext))
-}
-
-func (e *HandleError) Error() string {
-	return fmt.Sprintf("handling %v: %v", e.msgContext, e.err)
-}
-
-type commonError struct {
-	err        error
-	msgContext string
-}
-
-func newCommonError(err error, msgContext string) *commonError {
-	return &commonError{
-		err:        err,
-		msgContext: msgContext,
+	return &ProtocolError{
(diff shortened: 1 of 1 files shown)


lib/protocol/protocol.go
```diff
@@ -450,59 +450,59 @@ func (c *rawConnection) dispatcherLoop() (err error) {
...
-				return fmt.Errorf("protocol error: index message in state %d", state)
+				return indexError(fmt.Errorf("invalid state %d", state), "index", msg.Folder)
 			}
 			if err := checkIndexConsistency(msg.Files); err != nil {
-				return errors.Wrap(err, "protocol error: index")
+				return indexError(err, "index", msg.Folder)
 			}
 			if err := c.handleIndex(*msg); err != nil {
-				return fmt.Errorf("receiving index: %w", err)
+				return indexError(err, "index", msg.Folder)
```
calmh, line 459: The error returns from `handle...` should not be wrapped as `protocol error: ...` I think

calmh (commented):
I think there's a case to be made here for a `type ProtocolError struct { op, message string }` that formats itself consistently and applies to where we got something we shouldn't and a separate `type HandlerError { op string, inner error }` that wraps an error from the handler (which I think is a better name than `receiver` really, given there's so much other "receiving" going here that it's confusing otherwise)...

lib/protocol/protocol.go
```diff
@@ -1078,3 +1057,51 @@ func (c *rawConnection) lz4Decompress(src []byte) ([]byte, error) {
...
+
+type ProtocolError struct {
+	unwrapError
+}
+
+func newProtocolError(err error, msgContext string) *ProtocolError {
+	return &ProtocolError{
+		unwrapError: fmt.Errorf("protocol error on %v: %w", msgContext, err).(unwrapError),
+	}
+}
```
calmh, line 1074:
```suggestion
func protocolError(err error, msgContext string) error {
	return fmt.Errorf("protocol error on %v: %w", msgContext, err)
}
```

I mean, at this point we don't use the types for formatting or recognition, so we could skip having them :)

calmh (approved):

PR #7472, opened 2021-03-13
lib/model: Don't use ignore patterns for recv-enc folders (fixes #7469)

### LINKED ISSUE #7469: .stignore should be ignored in Receive Encrypted folders
Right now, if we create a new Receive Encrypted folder and try to edit it later, the Ignore Patterns tab is disabled.

However, if we do instead

1. Click "Add Folder".
2. Go to "Ignore Patterns" and set them to `*`.
3. Go to "Advanced" and set the "Folder Type" to "Receive Encrypted".
4. Click "Save".

then we end up with

which can only be fixed by deleting `.stignore` manually.

I'm not sure what the best way to fix this is, but I would personally propose to have Syncthing completely ignore the `.stignore` file when the folder type is set to "Receive Encrypted". This would likely require changes in the actual Go code though, and not just the GUI.

imsodin:


imsodin:
I'd like to get another RC released, and include a fix for #7459. Could someone please review this PR (it's small ;) ).

lib/model/model.go
```diff
@@ -2105,6 +2107,9 @@ func (m *model) SetIgnores(folder string, content []string) error {
 	if !ok {
 		return fmt.Errorf("folder %s does not exist", cfg.Description())
 	}
+	if cfg.Type == config.FolderTypeReceiveEncrypted {
+		return nil
+	}
```
AudriusButkevicius, line 2113: This seems lossy, as you can still create the file manually, so don't see much point in doing this.

AudriusButkevicius (approved):

PR #7473, opened 2021-03-13
gui: Improve Receive Encrypted folder type requirements text

tomasz1986:
The current text gives an impression that we are currently using a
Receive Encrypted folder, even if we are not. Thus, make the current
text displayed only when the folder is in fact Receive Encrypted, and
add a new string to be displayed when using different folder types.


tomasz1986:
I have also changed `===` to `==` in the other line to make it consistent with the rest of the code.

However, I'm still not sure about how the translation works. Right now, `"{%receiveEncrypted%}"` doesn't get changed in the newly added string, while it does in the existing ones. Will this fix itself or do I need to add something more to the code?

calmh (approved):

PR #7476, opened 2021-03-14
lib/db, lib/model: Allow needing invalid files (fixes #7474)

### LINKED ISSUE #7474: Device tries to push ignored files to a remote device in a 3-device setup
I'm sorry for the title, but I'm not sure how to describe this. For reference, see https://forum.syncthing.net/t/folder-stuck-in-sync-and-non-matching-local-and-global-states/16431.

1. Add a new folder with some files on Device A.
3. When adding the folder on Device B, ignore everything with `*`.
4. Share the folder from Device B to Device C.
5. Device B is now stuck trying to push the ignored files to Device C.

Device A:

Device B:

Device C:

imsodin:
### Purpose

Scenario (https://github.com/syncthing/syncthing/issues/7474#issuecomment-798881296):

> The reason is that we have an "efficiency" measure: We do not pull items that are not present at all locally and globally invalid or deleted. B sees valid items from A, thus their global is valid. C only sees invalid items from B, thus their global is invalid and they don't pull. B doesn't know about C's global, thus shows C as out of sync. The fix here is to always pull invalid files ("pull" as in create a local invalid entry too).

I adjusted `Need` to consider invalid files needed, and added a shortcut to the puller to send invalid files directly to the db. If at some point we do create a local file, that "trumps" the invalid file as usual (bumps the version of the invalid file, uninvalidates it, ...).

I also added a db migration to trigger a repair, otherwise the changes don't take effect.

### Testing

Unit test checking that we do pull and send back an invalid file info.  
Needed to add some machinery to the fake connection to correctly send invalid files.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -846,6 +846,10 @@ func (db *Lowlevel) getMetaAndCheck(folder string) (*metadataTracker, error) {
 	db.gcMut.RLock()
 	defer db.gcMut.RUnlock()
 
+	return db.getMetaAndCheckGCLocked(folder)
+}
+
+func (db *Lowlevel) getMetaAndCheckGCLocked(folder string) (*metadataTracker, error) {
 	fixed, err := db.checkLocalNeed([]byte(folder))
 	if err != nil {
 		return nil, fmt.Errorf("checking local need: %w", err)
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -21,7 +21,7 @@ import (
 // do not put restrictions on downgrades (e.g. for repairs after a bugfix).
 const (
 	dbVersion             = 14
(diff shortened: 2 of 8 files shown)


calmh (approved):

tomasz1986:
> I also added a db migration to trigger a repair, otherwise the changes don't take effect.

I have tested the commit, but I did have to reset the database in order for the errors to go away. Was is supposed to fix itself after replacing the binary?

imsodin:
Yes, should have fixed itself. On which device did you test it? You would have to do it on the device shown as out of sync, i.e. device C in your scenario in #7474.

tomasz1986:
I had replaced the binary on all three. However, I have also now tried to reproduce the behaviour again with a completely new folder, and it did fix itself, so I would call this solved 🙂.

PR #7484, opened 2021-03-16
cmd/syncthing/cli: Remove db reset and adjust others

imsodin:
Split off from: https://github.com/syncthing/syncthing/pull/7454

> I removed the db reset operation: I don't think that should be so easily available. I also added a warning to the override operation.

And use the new/non-legacy endpoint to check if a restart is required due to config changes.

### DIFF
--- a/cmd/syncthing/cli/operations.go
+++ b/cmd/syncthing/cli/operations.go
@@ -28,18 +28,13 @@ var operationCommand = cli.Command{
 			Action: expects(0, emptyPost("system/shutdown")),
 		},
 		{
-			Name:   "reset",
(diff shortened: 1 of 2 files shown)


tomasz1986:
The Docs have a few references to this one, so they will need an update if the commit gets merged.

imsodin:
Can you point to an example? I am not aware of any docs with `syncthing cli` references.

tomasz1986:
Never mind :flushed:. I got confused and thought that this was about the standard binary and the regular `-reset-database` flag. I'm sorry for the noise.

AudriusButkevicius (approved):

PR #7503, opened 2021-03-18
cmd/syncthing: Add more commands to cli

imsodin:
This adds the most used debugging endpoint and commands to pause/resume folders/devices. My general aim is to eventually provide the most used operations as dedicated commands, while "details" can be handled through `cli config ...`. So maybe add commands for sharing.

One thing I am not sure about: I went with two separate commands to pause/resume. An alternative would be a single `toggle-paused` command or a single `paused` command with a third boolean argument. I don't like the former very much, as the user needs to know the current state to achieve the right outcome. I am undecided between the latter and what is proposed in the PR (two commands).


acolomb:
> I am undecided between the latter and what is proposed in the PR (two commands).

I'd prefer two commands, as proposed.

cmd/syncthing/cli/show.go
```diff
@@ -40,5 +45,43 @@ var showCommand = cli.Command{
...
+			Usage:     "Show whether the folder/device is paused",
+			ArgsUsage: folderArgsUsage,
+			Action:    expects(2, paused),
+		},
 	},
 }
+
+func paused(c *cli.Context) error {
+	switch c.Args()[0] {
+	case "devices", "folders":
```
acolomb, line 59: Singular again.

cmd/syncthing/cli/show.go
```diff
@@ -40,5 +45,43 @@ var showCommand = cli.Command{
...
+	response, err := client.Get(strings.Join(append([]string{"config"}, c.Args()...), "/"))
+	if err != nil {
+		return err
+	}
+	bytes, err := responseToBArray(response)
+	if err != nil {
+		return err
+	}
+	var paused bool
+	if c.Args()[0] == "device" {
```
acolomb, line 73: This check doesn't work, it always treats the structure as a folder.  I wonder why `Unmarshal()` works, though?

acolomb (dismissed):
Did some poking on the API using this subcommand.  Some things not working as expected.

cmd/syncthing/cli/debug.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/urfave/cli"
+)
+
+var debugCommand = cli.Command{
+	Name:     "debug",
+	HideHelp: true,
+	Usage:    "Debug command group",
+	Subcommands: []cli.Command{
+		{
+			Name:   "enable",
```
AudriusButkevicius, line 21: Aren't these driver by the config, so you could do `config gui debug set true`?

cmd/syncthing/cli/operations.go
```diff
@@ -38,6 +39,18 @@ var operationCommand = cli.Command{
 			ArgsUsage: "[folder id]",
 			Action:    expects(1, foldersOverride),
 		},
+		{
+			Name:      "pause",
```
AudriusButkevicius, line 43: Ditto, driven by the config I think?

imsodin:
My intention was to have "shortcuts" for common tasks. But as Audrius exemplified above, the config commands are already quite short, and nicer anyway (e.g. you can get the folder ids, you don't need to know them). I'll remove the redundant bits and instead add some more debug commands, to make that command group less lonely.

imsodin:
I removed all the commands that are redundant with the `config` subcommand and instead added profiling and what was previously `cmd/stindex` to the `debug` command group.

cmd/syncthing/cli/debug.go
```diff
@@ -0,0 +1,51 @@
...
+			Usage:     "Save a profile to help figuring out what Syncthing does.",
+			ArgsUsage: "cpu | heap",
+			Action:    expects(1, profile()),
+		},
+	},
+}
+
+func debugFile() cli.ActionFunc {
+	return func(c *cli.Context) error {
+		return dumpOutput(fmt.Sprintf("debug/file?folder=%v&file=%v", c.Args()[0], c.Args()[1]))(c)
```
AudriusButkevicius, line 38: Could/should do path normalization.

cmd/syncthing/cli/index.go
```diff
@@ -0,0 +1,54 @@
...
+			Usage:  "Print the entire db",
+			Action: expects(0, dump),
+		},
+		{
+			Name:   "dumpsize",
+			Usage:  "Print the db size of different categories of information",
+			Action: expects(0, dumpsize),
+		},
+		{
+			Name:   "idxck",
```
AudriusButkevicius, line 31: Just "check"?

cmd/syncthing/cli/main.go
```diff
@@ -150,20 +150,16 @@ func Run() error {
...
 			operationCommand,
 			errorsCommand,
+			debugCommand,
 		},
 	}}
 
-	tty := isatty.IsTerminal(os.Stdin.Fd()) || isatty.IsCygwinTerminal(os.Stdin.Fd())
-	if !tty {
+	// tty := isatty.IsTerminal(os.Stdin.Fd()) || isatty.IsCygwinTerminal(os.Stdin.Fd())
+	if false {
```
AudriusButkevicius, line 162: ???

imsodin: Lol, my git foo: When running in delve/debugger, this terminal business fails. I was too lazy to look into what the purpose of it all is and ended up committing the workaround.

cmd/syncthing/cli/index.go
```diff
@@ -8,47 +8,31 @@ package cli
...
-	"github.com/syncthing/syncthing/lib/locations"
 )
 
 var indexCommand = cli.Command{
 	Name:  "index",
 	Usage: "Show information about the index (database)",
 	Subcommands: []cli.Command{
 		{
-			Name:   "dump",
+			Name:   "indexDump",
```
AudriusButkevicius, line 18: this is already a index subcommand. why not just dump?
```syncthing cli index indexDump` sounds silly,.

File: cmd/syncthing/cli/index.go [hunk 9]
```diff
@@ -8,47 +8,31 @@ package cli
...
 			Usage:  "Print the entire db",
-			Action: expects(0, dump),
+			Action: expects(0, indexDump),
 		},
 		{
-			Name:   "dumpsize",
+			Name:   "indexDumpsize",
```
AudriusButkevicius, line 23: What do we do with multiWord parameters elsewhere? are they camelCased? kebab-cased etc?
Just want to make sure we stay consistent.

cmd/syncthing/cli/utils.go
```diff
@@ -127,3 +130,20 @@ func prettyPrintResponse(c *cli.Context, response *http.Response) error {
...
+	for i := range bs {
+		if bs[i] == 0 {
+			return string(bs[:i])
+		}
+	}
+	return string(bs)
+}
+
+func normalizePath(path string) string {
+	return filepath.ToSlash(filepath.Clean(path))
```
AudriusButkevicius, line 148: YAGNI, but wonder if we should handle people giving us the full absolute path etc, and trimming the folder path prefix.

imsodin: I am not convinced we need to that. And if we do, then I'd argue that should happen on the api side of things.

AudriusButkevicius (approved):

imsodin:
I didn't notice this introduces an issue deepsource considers "security" relevant (and thus persists an ugly red cross to our commit checks): https://deepsource.io/gh/syncthing/syncthing/issue/GO-S0904/occurrences

Their description:

> Detects possible usage of Logger for user defined data that could break log pattern.
> 
> Detects usage of methods from the packages "log" and "go.uber.org/zap" where the arguments could be tainted and untrusted.
> 
> Audit the arguments of the logger for user defined input(s) to prevent attacks.
> Examples
> Bad practice
> 
> ```
> // input() returns user defined data
> unsafe := input()
> 
> // If unsafe contains characters like '\r', '\n', etc. then it might break the log pattern
> // Manual auditing is required
> log.Println(unsafe)
> ```

We do print "unsafe" data all the time (folder labels, device names, file names, ...) and I never considered that problematic. And still don't.

Apparently this can only be ignored in their web interface, not the toml in the repo: https://deepsource.io/blog/releases-issue-actions/ (at least I didn't find anything on it). And I don't have the rights to do so. If you agree this is not an actual issue, please ignore it on there or give me permission to do so myself :)

AudriusButkevicius:
I mean, for a cli thats a bit BS anyway, as if you cam run cli you can do whatever on the machine amyway.

PR #7505, opened 2021-03-19
all: Encrypted fileinfo trailer needs to be in wire format

imsodin:
This issue came up in https://forum.syncthing.net/t/how-to-restore-decrypt-an-encrypted-untrusted-folder/16520/11

If the untrusted devices is running windows, the `decrypt` tool errors on decrypting metadata. Because we wrote that metadata in native format, while the protocol code expects wire format (slashes).

### DIFF
--- a/lib/model/sharedpullerstate.go
+++ b/lib/model/sharedpullerstate.go
@@ -352,8 +352,8 @@ func (s *sharedPullerState) finalizeEncrypted() error {
 	wireFile := s.file
 	wireFile.Name = osutil.NormalizedFilename(wireFile.Name)
 
-	bs := make([]byte, encryptionTrailerSize(s.file))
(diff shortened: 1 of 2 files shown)


lib/model/sharedpullerstate.go
```diff
@@ -346,6 +347,11 @@ func (s *sharedPullerState) finalClose() (bool, error) {
 // folder from encrypted data we can extract this FileInfo from the end of
 // the file and regain the original metadata.
 func (s *sharedPullerState) finalizeEncrypted() error {
+	// Here the file is in native format, while encryption happens in
+	// wire format (always slashes).
+	wireFile := s.file
```
calmh, line 352: creates `wireFile` but doesn't actually write it to the trailer

calmh (approved):

Issue #7198, opened 2020-12-15
Removing a folder and re-adding it again (untrusted) makes it not sync until restart


calmh:
Most likely because the other side has cached the fact that we already had index information and won't send it again, even though we've dropped our index data.

imsodin:
I am unable to repro. What I did:

1. Have a folder with a few files and dirs in sync between devices A and B - connection active.
2. Remove the folder on A.
3. Readd the folder on A with same ID but pointing at a new, empty path.

Result: Files get synced.

What did I do differently?

calmh:
So, in my case this was an untrusted folder, being removed and then rm-rf:ed on the untrusted side because it had become full of crap, then re-added. I don't think that should be matter? Will retest.

BeayemX:
Maybe the problem is related to #7076 ?

calmh:
(This happened again yesterday, again with an untrusted folder, so it's real and still there. Still requires investigation.)

PR #7510, opened 2021-03-21
lib: Handle adding enc folders on an existing conn (fixes #7509)

### LINKED ISSUE #7509: Sharing the first folder to an untrusted device on an established connection sends plain-text
Was discovered in https://forum.syncthing.net/t/problem-adding-second-trusted-device-connected-to-untrusted-device/16538

> 
> Prerequisites
> 
>     Devices A+B: Discovery and nat traversal deactivated, Relaying on (Windows x32)
>     Device U: Discovery, nat traversal and relaying deactivated (Ubuntu 20.04 64bit)
>     All Devices on v1.15.0-rc.5, Windows machines using SyncTrayzor
>     Devices A and B are connected to U. Status: Connected (unused)
> 
> Steps to reproduce
> 
>     Add new Folder to Device A and share it with encryption password
>     Receive Folder on U and add it.
> 
> => U will now show “Up to date” but will not receive any data. A will show “Syncing” for the device and “up to date” for the folder while not sending any data.
> 
>     Pause the new Folder on A and resume it
> 
> => A and U will now sync but U will now have the unencrypted data at the chosen path.

A connection handles multiple folders, every single can be encrypted or not. Which folders are or not encrypted is set when the connection is created. At the time the code was written that was ok, as we dropped a connections when any folder changed. Now we don't anymore, thus adding an encrypted folder doesn't make that connection encrypted until the connection is dropped (restart, device pause, ...).

Should this go into the changelog? It's definitely a "note-worthy" bug, but the untrusted feature is still in testing/preview. Not sure.

The point about having to pause and resume the folder to make anything happen is unrelated (did repro but don't yet know why it happens).

imsodin:
Fixes #7509 by adding a method to the connection to set new folder passwords. This means that any connection may become (partly) encrypted. Thus there's no longer two separate factory methods, but every connection is a `encryptedConnection`.

The passwords on the connection are (re-)set when sending cluster-configs, meaning when a new connection is added, when a new folder is added or when a folder changes.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1535,7 +1535,7 @@ func (m *model) sendClusterConfig(ids []protocol.DeviceID) {
 	// Generating cluster-configs acquires fmut -> must happen outside of pmut.
 	for _, conn := range ccConns {
 		cm, passwords := m.generateClusterConfig(conn.ID())
-		conn.ResetFolderPasswords(passwords)
+		conn.SetFolderPasswords(passwords)
 		go conn.ClusterConfig(cm)
 	}
 }
@@ -2247,7 +2247,7 @@ func (m *model) AddConnection(conn protocol.Connection, hello protocol.Hello) {
 
 	// Acquires fmut, so has to be done outside of pmut.
 	cm, passwords := m.generateClusterConfig(deviceID)
-	conn.ResetFolderPasswords(passwords)
+	conn.SetFolderPasswords(passwords)
 	conn.ClusterConfig(cm)
 
 	if (device.Name == "" || m.cfg.Options().OverwriteRemoteDevNames) && hello.DeviceName != "" {
@@ -2407,8 +2407,8 @@ func (m *model) numHashers(folder string) int {
 	return 1
 }
 
-// generateClusterConfig returns a ClusterConfigMessage that is correct for
-// the given peer device
+// generateClusterConfig returns a ClusterConfigMessage that is correct and the
+// set of folder passwords for the given peer device
 func (m *model) generateClusterConfig(device protocol.DeviceID) (protocol.ClusterConfig, map[string]string) {
 	var message protocol.ClusterConfig
 
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -293,7 +293,7 @@ func (m *model) initFolders(cfg config.Configuration) error {
 	ignoredDevices := observedDeviceSet(m.cfg.IgnoredDevices())
 	m.cleanPending(cfg.DeviceMap(), cfg.FolderMap(), ignoredDevices, nil)
 
-	m.resendClusterConfig(clusterConfigDevices.AsSlice())
+	m.sendClusterConfig(clusterConfigDevices.AsSlice())
 	return nil
 }
 
@@ -1510,7 +1510,7 @@ func (m *model) ccCheckEncryption(fcfg config.FolderConfiguration, folderDevice 
 			m.fmut.Unlock()
 			// We can only announce ourselfs once we have the token,
 			// thus we need to resend CCs now that we have it.
-			m.resendClusterConfig(fcfg.DeviceIDs())
+			m.sendClusterConfig(fcfg.DeviceIDs())
 			return nil
 		}
 	}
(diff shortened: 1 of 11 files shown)


lib/model/model.go
```diff
@@ -2409,13 +2409,15 @@ func (m *model) numHashers(folder string) int {
 
 // generateClusterConfig returns a ClusterConfigMessage that is correct for
 // the given peer device
```
calmh, line 2411:
```suggestion
// generateClusterConfig returns a ClusterConfigMessage and the set of folder passwords for
// the given peer device
```

lib/protocol/encryption.go
```diff
@@ -135,42 +136,46 @@ func (e encryptedModel) ClusterConfig(deviceID DeviceID, config ClusterConfig) e
...
 }
 
 // The encryptedConnection sits between the model and the encrypted device. It
 // encrypts outgoing metadata and decrypts incoming responses.
 type encryptedConnection struct {
 	ConnectionInfo
-	conn       Connection
-	folderKeys map[string]*[keySize]byte // folder ID -> key
+	conn *rawConnection
+	em   *encryptedModel
```
calmh, line 148: I think this should be a pointer to the folderKeyRegistry directly instead of reaching into the encryptedModel

lib/protocol/encryption.go
```diff
@@ -590,3 +595,27 @@ func isEncryptedParentFromComponents(pathComponents []string) bool {
...
+	keys map[string]*[keySize]byte // folder ID -> key
+	mut  sync.RWMutex
+}
+
+func newFolderKeyRegistry() *folderKeyRegistry {
+	return &folderKeyRegistry{
+		keys: make(map[string]*[keySize]byte),
+	}
+}
+
```
calmh, line 609: Don't really need a constructor for this one, nil map is fine until reset() happens

lib/protocol/encryption.go
```diff
@@ -590,3 +595,27 @@ func isEncryptedParentFromComponents(pathComponents []string) bool {
...
+func (r *folderKeyRegistry) get(folder string) (*[keySize]byte, bool) {
+	r.mut.RLock()
+	key, ok := r.keys[folder]
+	r.mut.RUnlock()
+	return key, ok
+}
+
+func (r *folderKeyRegistry) reset(passwords map[string]string) {
```
calmh, line 617: `setPasswords`? `reset` sounds like remove all passwords to me

lib/protocol/protocol.go
```diff
@@ -140,6 +140,7 @@ type RequestResponse interface {
 
 type Connection interface {
 	Start()
+	ResetFolderPasswords(passwords map[string]string)
```
calmh, line 143: Likewise, maybe `SetFolderPasswords`?

calmh (commented):
Also fixes #7198 I think

lib/protocol/encryption.go
imsodin: I was just wondering whether one might misunderstand the end-result: It is exactly `passwords` but it could also be the previous state with the elements of `passwords` applied on top of it (i.e. missing passwords remaining). `reset` feels a bit more clear in that respect than just `set`. Maybe `replace` instead?

imsodin:
#7198 sounds like the issue I am seeing when reproducing the issue here (having to pause/resume the folder before indexes get exchanged). And I am still seeing it with the PR applied.

lib/protocol/encryption.go
calmh: I think `set` implies replace, but I'm fine with `replace` as well.

lib/protocol/encryption.go
imsodin: I don't really want a nil-check in `folderKeyRegistry.get` though.

calmh: Why would you need one?

imsodin: I'd expect `key, ok := r.keys[folder]` to panic. And without checking I am not 100% that `model.AddConnection` is guaranteed to initialize the passwords before any message comes in from a remote. Expressing that thought makes it clear that I should keep initializing passwords when creating the connection. Likely it's all irrelevant practically, as the early messages aren't encrypted, but still.

calmh: It doesn't panic though, a nil map is equivalent to a non-nil but empty map when reading.

imsodin: Cool, wasn't aware of that. I didn't really internalize that aspect of go (treating nil as a valid state), I treat nil as "do not touch".
Any ways, it does now get initialized at connection construction time with a valid map of passwords.

calmh (approved):

PR #7517, opened 2021-03-22
cmd/strelaysrv: Increase default strelaysrv network buffer size (fixes #7514)

### LINKED ISSUE #7514: Increase default strelaysrv network buffer size

no

The default network buffer size is not a multiple of 64K per default, which is the maximum TCP packet size.
This causes unneccesary fragmentation, slowing down large file uploads when a relay is involved.
Changing this value to 64K (65536) significantly improves upload speed on fast connections.

- strelaysrv v0.14.49, linux amd64

AudriusButkevicius:
Increases throughput at the cost of more memory per connection

### DIFF
--- a/cmd/strelaysrv/main.go
+++ b/cmd/strelaysrv/main.go
@@ -233,6 +233,7 @@ func main() {
 	uri, err := url.Parse(fmt.Sprintf("relay://%s/?id=%s&pingInterval=%s&networkTimeout=%s&sessionLimitBps=%d&globalLimitBps=%d&statusAddr=%s&providedBy=%s", mapping.Address(), id, pingInterval, networkTimeout, sessionLimitBps, globalLimitBps, statusAddr, providedBy))
 	if err != nil {
 		log.Fatalln("Failed to construct URI", err)
+		return
 	}
 
 	log.Println("URI:", uri.String())
--- a/cmd/strelaysrv/main.go
+++ b/cmd/strelaysrv/main.go
@@ -99,7 +99,7 @@ func main() {
 	flag.IntVar(&natRenewal, "nat-renewal", 30, "NAT renewal frequency in minutes")
 	flag.IntVar(&natTimeout, "nat-timeout", 10, "NAT discovery timeout in seconds")
 	flag.BoolVar(&pprofEnabled, "pprof", false, "Enable the built in profiling on the status server")
-	flag.IntVar(&networkBufferSize, "network-buffer", 2048, "Network buffer size (two of these per proxied connection)")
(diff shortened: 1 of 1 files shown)


calmh:
Sorry, looking more closely at this I don't think this is appropriate. It's already a flag so the user can optimize as they like, lots of relays have a hundred sessions which by my mental count would cost 13+ gigs of RAM. So no, I think the current default is fine despite the better performance with a bigger buffer.

I guess an alternative would be some sort of smarter shared buffer handling that only used buffer space when required, but that's not what we have.

AudriusButkevicius:
This is only needed for active sessions, maximum we have at the moment is 1115, which is
(1115 * 65335 * 2) / 1<<20 = ~138mb

AudriusButkevicius:
I guess even setting it to 32k or 16k would be beneficial over the current 2k.

calmh:
I'm a dumbass and off by a factor 1024. This is fine.

cmd/strelaysrv/main.go
```diff
@@ -233,6 +233,7 @@ func main() {
 	uri, err := url.Parse(fmt.Sprintf("relay://%s/?id=%s&pingInterval=%s&networkTimeout=%s&sessionLimitBps=%d&globalLimitBps=%d&statusAddr=%s&providedBy=%s", mapping.Address(), id, pingInterval, networkTimeout, sessionLimitBps, globalLimitBps, statusAddr, providedBy))
 	if err != nil {
 		log.Fatalln("Failed to construct URI", err)
+		return
```
calmh, line 236: Not necessary but 🤷

calmh (approved):

cmd/strelaysrv/main.go
AudriusButkevicius: Your fancy lint started crying.

calmh: Yeah what it means is they the defer doesn’t run due to Fatalln, but they doesn’t matter much

AudriusButkevicius: 🤷‍♂️ 

```
In these cases, in case of failure, defer never gets called.

Example:

defer os.Remove(filename)
if bad {
    log.Fatalf("something bad happened")
}
should be written as:

defer os.Remove(filename)
if bad {
    log.Printf("something bad happened")
    return
}
```

calmh: Println + return is a normal return so the defer runs. Fatalln is an immediate exit so no defer gets a chance to run.  But the defer is a context cancel so we have no reason to care.

PR #7522, opened 2021-03-25
gui: Add confirmation dialog for revert/override (fixes #7520)

### LINKED ISSUE #7520: Ask for confirmation before reverting/overriding a folder
I lost work recently clicking the 'Revert local changes?' button on a receive only folder, expecting a dialogue box would list the affected files and seek my confirmation before reverting or deleting.  To my surprise, the file I'd mistakenly saved in the wrong folder was instantly deleted.  Everone's expectations are different of course but the potential for accidental data loss here is huge.

No versioning strategy was employed on this particular folder but even if it had been, I don't believe it would have helped since versioning strategies only apply to changes coming in from other devices.

imsodin:
Generalises the confirmation dialog for deleting items added to receive-encrypted folders to also handle reverting/overriding receive/send-only folders (fixes #7520).


gui/default/syncthing/folder/revertOverrideView.html
```diff
@@ -0,0 +1,33 @@
...
+      </p>
+      <p translate>
+        Are you sure you want to permanently delete all these files?
+      </p>
+    </div>
+    <div ng-switch-default>
+      <p>
+        <span translate>Warning</span>:</br>
+        <span ng-switch="revertOverrideParams.type">
+          <span ng-switch-when="override" translate>The folder content on other devices will be overwritten to become identical with this device. E.g. files not present here will be deleted on other devices.</span>
```
tomasz1986, line 16:
```suggestion
          <span ng-switch-when="override" translate>The folder content on other devices will be overwritten to become identical with this device, e.g. files not present here will be deleted on other devices.</span>
```

The "E.g." at the beginning of the second sentences looks kind of weird. How about using a comma, or if the sentence becomes too long, then just starting the second sentence with writing "for example" or "for instance" in full?

gui/default/syncthing/folder/revertOverrideView.html
```diff
@@ -0,0 +1,33 @@
...
+          <span ng-switch-default translate>The folder content on this device will be overwritten to become identical with all other devices. E.g. files newly added here will be deleted.</span>
```
tomasz1986, line 17:
```suggestion
          <span ng-switch-default translate>The folder content on this device will be overwritten to become identical with other devices. E.g. files newly added here will be deleted.</span>
```

There is no "all" in the preceding override dialogue, so how about dropping it from here too? Or is perhaps it necessary to underline the point? I think that the sentence sounds fine without the word.

gui/default/syncthing/folder/revertOverrideView.html
calmh: I think the second sentence is fine without the e.g. as well, just saying that files here/there will be deleted.

calmh:
build.go translate

acolomb (approved):
I really like the unifying design. One minor nit-pick / question, ignore at will.

acolomb (approved):

PR #7525, opened 2021-03-26
lib/db: Handle indirection error repairing sequences (fixes #7026)

### LINKED ISSUE #7026: panic: key not found when repairing DB

imsodin:
#7026 is about unspecific `key not found` errors on sentry. The only one that still gets events is https://sentry.syncthing.net/share/issue/270bbc681e3943f7a121268553e6f2e5/, and that's what I am addressing here:

`panic: repairing sequences: filling Blocks: key not found`

When repairing sequences all local files are deserialized, and apparently sometimes the blocks are indirected (or look like it), but theirs no entry for the hash. Instead of failing db repair and thus making the db entirely broken, the defective file info is now removed from the db - to be rescanned later.

Some related changes: I moved checking/repairing globals out of `recalcMeta` into `getMetaAndCheck`. Having that happen as a side effect of recalculating metadata is plain confusing (and I probably wrote that code :) ).  
And the db migration 15 (which uses `recalcMeta`), is redundant with 16 (the meta just gets recalculated twice) - thus removed.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -441,13 +441,14 @@ func (db *Lowlevel) dropDeviceFolder(device, folder []byte, meta *metadataTracke
 	return t.Commit()
 }
 
-func (db *Lowlevel) checkGlobals(folder []byte) (int, error) {
+func (db *Lowlevel) checkGlobals(folderStr string) (int, error) {
 	t, err := db.newReadWriteTransaction()
 	if err != nil {
 		return 0, err
 	}
 	defer t.close()
 
+	folder := []byte(folderStr)
 	key, err := db.keyer.GenerateGlobalVersionKey(nil, folder, nil)
 	if err != nil {
 		return 0, err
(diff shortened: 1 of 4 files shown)


lib/db/lowlevel.go
```diff
@@ -1022,6 +1032,25 @@ func (db *Lowlevel) repairSequenceGCLocked(folderStr string, meta *metadataTrack
...
+				if err != nil {
+					return 0, err
+				}
+				if err := t.Delete(sk); err != nil {
+					return 0, err
+				}
+				if err := t.Delete(it.Key()); err != nil {
+					return 0, err
+				}
+			}
```
calmh, line 1053: it also needs to be removed from globals?

calmh (approved):

PR #7530, opened 2021-03-28
lib, gui: Default ignores for new folders (fixes #7428)

### LINKED ISSUE #7428: add Ignore Patterns to Folder Defaults
It would be nice to input a default set of Ignore Patterns for newly added folders that you would otherwise manually apply to every share you add. E.g.:

__Settings__ :arrow_right: __Default Configuration__ :arrow_right: __Edit Folder Defaults__ :arrow_right: __Ignore Patterns__

For example, I always add this by default. Sometimes I forget, causing certain peers to exceed their free disk space.

```
(?d).Trash-*
(?d).sync.ffs_db
(?d)tmp
node_modules
bower_components
lost+found
*.bak
*.delme
```

This template would then be copied to the added folder ignore patterns where it can be edited. A new installation of Syncthing could add some sane defaults for the user, or just leave it empty.

imsodin:
Adds default ignores. Those are applied when a folder is auto-accepted. When adding a folder in the web UI, it checks if there is a pre-existing `.stignore` file at the new folder path (through a new `/rest/system/readignores` endpoint), and if not uses the defaults.

The defaults are stored in the config folder in a file named `default-stignore`. The can be modified through the `/rest/db/defaultignores` endpoint. `/rest/db` hardly is the right place for this, but `/rest/db/ignores` already exists. And it's not obvious to me where to put it: Putting it into an own group feels like overkill (`/rest/ignores/folder/folder-id` and `/rest/ignores/default`) and moving to `/rest/system` isn't much better than `/rest/db`. Suggestions for a better endpoint and filename are welcome :) 

Testing: I can change defaults and they appear when adding a new folder.


tomasz1986:
I have just done some very quick testing.

1. When editing the default ignore patterns, the path displayed at the bottom is wrong.
   
   image

2. The GUI shows this error if the default ignore file doesn't exist.

   image

3. The `default-stignore` files is created hidden on Windows. I don't think it should be?

4. After deleting the `default-stignore` file manually, the GUI still shows the previously added patterns. Trying to clear/modify them results in an error.
   
   image


Also, not really an issue, but how about naming the file something like `.stignore-default` instead of `default-stignore`? Just to make it closer to `.stignore`.

calmh:
Why not in fact store the defaults in the db (or config)? I get that there's a symmetry with the regular ignore patterns being files on disk, but I think we may want to start moving away from that.

imsodin:
> Why not in fact store the defaults in the db (or config)? I get that there's a symmetry with the regular ignore patterns being files on disk, but I think we may want to start moving away from that.

I personally enjoy being able to edit ignore patterns in a plain file. Outside of dev the one thing I touch the most about my Syncthing clients is their ignore pattern files :)  
What's the benefit of/intention for moving it to config/db?

> Also, not really an issue, but how about naming the file something like .stignore-default instead of default-stignore? Just to make it closer to .stignore.

I agree `stignore-default` makes more sense. The leading dot is to hide a file, and I don't want to hide it in the syncthing config dir.

calmh:
I can see the draw of editing files on disk, but that's an argument for a .foldercfg.yaml in every folder. :) As is there are advantages to ignore patterns in files (#include, mainly) and there's history to deal with, but I don't see a reason for keeping *default* ignore patterns physically different from default everything-else...

imsodin:
Ok, "default everything-else" then means config, probably `defaults` -> `ignores`. Which then is a potentially very long text field.

acolomb:
> As is there are advantages to ignore patterns in files (#include, mainly) and there's history to deal with, but I don't see a reason for keeping _default_ ignore patterns physically different from default everything-else...

I agree that it doesn't need to be a separate file, but should just go to the `config.xml` file.  This way everything default-config-related is to be found in the same place.  Please *do not* store it in the database.  I think having the assurance that deleting the database folder works as a kind of "reset" for transient data, but never touches explicit user-chosen options is comforting and useful to handle support.

If it should in fact end up as a separate config file, I'd prefer `stignore-default`, no dot, definitely not hidden on any OS.

tomasz1986:
> As is there are advantages to ignore patterns in files (#include, mainly) and there's history to deal with, but I don't see a reason for keeping _default_ ignore patterns physically different from default everything-else...

The default ignore patterns also can have the `#include` in them, can't them? :wink:

However, if the default patterns are stored in the config file, is there any reason not to store each folder's patterns like that too? Of course, this would increase the config file size quite significantly.

Another problem with using the `config.xml` though is that some characters (e.g. `&`) will have to be escaped.

calmh:
Storing folder ignore patterns in the config is a rabbit hole that we should not get down into in this issue. The issues of migrations and include patterns and whatnot are non-trivial and I'd be happy to hash out a solution to that -- on the forum.

*Default* ignore pattens are however just a string. You can type "#include" into it, but we're not going to parse the default ignore patterns (until they're written to an .stignore file somewhere and become regular ignore patterns) so it doesn't matter.

tomasz1986:
Using two different formats for the default and the actual patterns will make it impossible to do a simple copy & paste on a file level though. One will have to either use the GUI/API or reformat the existing patterns manually in order to paste them into the `config.xml` (and the other way round).

This is not the case with the rest of the default folder settings, where one can simply copy and replace the relevant parts inside the `config.xml` file itself. If the per-folder ignore patterns are to stay in a file, I would personally prefer to have the default ones in a file also (as the commit currently does), just for the sake of consistency and the easy manipulation mentioned above…

imsodin:
I moved the default ignores into config.xml. Besides the various good arguments made already for either option (plain file or config.xml based storage), the config option also has the benefit of a simpler implementation: No additional responsibilities for the model (a separate `ignore.Matcher` instance plus file on disk) and straight-forward integration into the api (at `config/defaults/ignores`).  
I personally somewhat lament losing the simplicity of modifying a plain text file for ignores here, so I added `syncthing cli operations default-ignores` which accepts a path to a file containing ignore patterns, which are then applied as default ignores to the config :)

cmd/syncthing/cli/client.go
```diff
@@ -54,7 +56,7 @@ func (c *APIClient) Endpoint() string {
 	return url
 }
 
-func (c *APIClient) Do(req *http.Request) (*http.Response, error) {
+func (c *APIClient) Request(req *http.Request) (*http.Response, error) {
```
calmh, line 59: `Do` matches https://golang.org/pkg/net/http/#Client.Do so this rename is slightly confusing to me

tomasz1986:
This is kind of a corner case, but the default ignore patterns are not used in the following situation.

1. Set ignore patterns to `123`.
1. Create a folder `test` with ignore patterns set to `abc`.
2. Start creating yet another folder, and first set its path to the same as `test`.
3. The ignore patterns for the new folder have been set to `abc`, which is expected.
4. Without saving the changes, now change the folder path to a completely new one.
5. The ignore patterns are still `abc`, and not the default `123`.

imsodin:
Ready for (re-)review.

tomasz1986:
This is happening with an existing folder, but I have compiled the binary, started it, and immediately saw that even though the folder has some ignore patterns set,

image

the Ignore Patterns tab displays nothing.

image

The patterns are displayed if I use the current `main` though.

imsodin:
Thanks, typo spotted and fixed.

lib/config/config.go
```diff
@@ -603,3 +603,20 @@ func ensureZeroForNodefault(empty interface{}, target interface{}) {
...
 }
+
+// IgnoresFromFile return Ignores with Lines initiated with the lines of the file at give path.
+func IgnoresFromFile(path string) (Ignores, error) {
+	content, err := ioutil.ReadFile(path)
+	if err != nil {
+		return Ignores{}, err
+	}
+	return Ignores{
+		Lines: strings.Split(string(content), "\n"),
```
calmh, line 614: We probably need to handle Windows line endings here, possibly by just strings.TrimSpace()ing every line

lib/api/api.go
```diff
@@ -1280,27 +1282,43 @@ func (s *service) getDBIgnores(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
 
-	err = s.model.SetIgnores(qs.Get("folder"), data["ignore"])
-	if err != nil {
+	s.getDBIgnores(w, r)
+}
+
+func (s *service) readIgnores(w http.ResponseWriter, r *http.Request) {
+	ignores, err := config.IgnoresFromFile(r.URL.Query().Get("path"))
```
calmh, line 1301: This is a generic file reader (readIgnores("/etc/passwd")), not sure if that matters or not as it's anyway a thing Syncthing can read and hence sync if asked to, but we should consider.

calmh (approved):

lib/config/config.go
imsodin: Yep. Used `bufio.Scanner` instead to let it handle such issues.

lib/api/api.go
imsodin: Should we rename the method to something like "readfile" to make it explicit? I don't think there's a sensible way around having such an endpoint and still sanely handle ignores on new folders (i.e. not clobber any existing .stignore files).

calmh: We could maybe have a `clobber: false` attribute on the set-ignores endpoint, which only sets the ignores if the file is not already present... And then use that if the user creates a folder with the default unmodified ignores...

imsodin: If we want to get rid of this, I'd propose delaying setting ignore patterns to after the folder is created. That already happens behind the scenes by first adding the folder paused, then applying ignores, then unpausing - I'd just reflect that in the UI too. So when adding a folder instead of an input field, there would just a be a short explanation and a checkbox. If ticked, one can edit the ignores after clicking save (or add or whatever the confirmation button is).

calmh: That checkbox would leave the folder paused, you mean? And at that point we go through the normal read-ignores and/or use the default? The user then unpauses the folder when done. I guess that would work.

Having an attribute to set-ignores to say "only write if there aren't already any ignores" would also work though I think.

imsodin: I meant that, but automatically: User clicks save, all input elements get frozen and the ignores checkbox switched to an ignores input populated through the usual "give me this folders ignore patterns"  endpoint. Then they click save again and the folder gets unpaused. Lets do that actually. 

And that kind of flag seems like a mixed bag UX wise: I tell syncthing to set some ignores, and then it doesn't actually do that to not clobber existing ones.

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -173,8 +173,7 @@
             <dd><span translate>Comment, when used at the start of a line</span></dd>
           </dl>
           <hr />
```
tomasz1986, line 175:
```suggestion
          <hr ng-if="!editingDefaults" />
```

The `hr` is not needed when the next line is hidden, so I would hide it too. Otherwise there is a superfluous horizontal line there.

imsodin:
Here it is. When adding a folder now you get just a checkbox in the ignores tab:

Ignore tab of add folder modal with checkbox to add ignores later

Then after clicking save it switches to the following:

Ignore tab of add folder modal after saving, where you actually enter patterns

Other tabs are disabled in this state. Behind the scenes the folder is already added, but paused. Once you click save again the ignores are applied and then the folder is resumed.

The code is as ugly as expected when touching significant amounts of it. Lots of the diff is replacing `$scope.editingExisting` and `$scope.editingDefaults` (and adding a new folder is implicitly represented by both being false) with `$scope.currentFolder._editing` which is a string that's either `add`, `add-ignores` (the new state introduced in this PR), `existing` or `defaults`.

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -2,44 +2,44 @@
...
+        <li ng-class="{'disabled': currentFolder._editing == 'add-ignores'}"><a data-toggle="tab" href="{{currentFolder._editing == 'add-ignores' ? '' : '#folder-sharing'}}"><span class="fas fa-share-alt"></span> <span translate>Sharing</span></a></li>
+        <li ng-class="{'disabled': currentFolder._editing == 'add-ignores'}"><a data-toggle="tab" href="{{currentFolder._editing == 'add-ignores' ? '' : '#folder-versioning'}}"><span class="fas fa-copy"></span> <span translate>File Versioning</span></a></li>
+        <li ng-class="{'disabled': currentFolder._recvEnc}"><a data-toggle="tab" href="{{currentFolder._recvEnc ? '' : '#folder-ignores'}}"><span class="fas fa-filter"></span> <span translate>Ignore Patterns</span></a></li>
+        <li ng-class="{'disabled': currentFolder._editing == 'add-ignores'}"><a data-toggle="tab" href="{{currentFolder._editing == 'add-ignores' ? '' : '#folder-advanced'}}"><span class="fas fa-cogs"></span> <span translate>Advanced</span></a></li>
```
imsodin, line 9: This is an abomination. Bootstrap depends on `form > ul > li` so it isn't straight forward replaceable with larger, more verbose structures (e.g. a diff with an `ng-if`). Probably could be made a bit less abominable using angularjs directives.

tomasz1986:
It seems impossible to create a new folder with the checkbox unticked in the newest iteration 😉. It gets stuck with an "Unknown" state, and then disappears after refreshing the GUI.

imsodin:
Forgot this was still open - resolved conflicts. Would be good to merge this before new conflicts come up :)

tomasz1986:
Something is broken in the "Edit Device Defaults".

image

The other tabs in the "Edit Folder" modal are "disabled", but you can still click them (or tab with the keyboard) and change focus this way. A cosmetic issue, but it is what it is.

image

Just a comment, as I guess there is no other way to implement this without the checkbox and such, but the overall experience does feel a little bit clunky…

tomasz1986:
Will the Android app cooperate with this one? I mean the native interface, not the Web GUI. I can test, maybe tomorrow, if needed 😉.

tomasz1986:
Yeah, just curious what would happen if you first set default ignores in the Web GUI, and then try to create folder using the native one.

st-review:
:robot: *beep boop*

imsodin:
This is still ready for review and merge - I just forgot about it as well in the last two months. Maybe lets try to get it merged in <1a :D

tomasz1986:
I'm giving this another try at the moment.

1. The folder doesn't seem paused when the new modal asking to add ignore patterns pops up. What I mean is that the folder is already shared and can be accepted on the remote device while the user is still inputting the ignore patterns on the first device.
2. There are JS errors when opening and saving a remote device modal.
	```
	TypeError: Cannot read properties of undefined (reading '_editing')
			at Scope.$scope.editDeviceModalIcon (syncthingController.js:1606)
			at $parseFunctionCall (angular.js:12493)
			at Object.expressionInputWatch (angular.js:12902)
			at Scope.$digest (angular.js:14423)
			at angular.js:14984
			at angular.js:5347
	```
3. Default ignores fail to apply when using "auto accept".
image
	Note that the auto-accepted folder in my testing has the same name as an existing folder, so Syncthing automatically uses the ID as its foldername instead.

imsodin:
I blame other changes that happened in the meantime (which means in reality I likely messed up, but it might also be due to other changes) :D 

Thanks for testing @tomasz1986 , I'll look into it.

tomasz1986:
No problem 🙂.

I'd love to have this implemented to be able to use at work computers. There are too many devices and folders to always remember to add ignore patterns manually there.

I'd also like to add one more thing regarding folder pausing. Even if the folder is set to start "paused" by default, with this PR it seems to always start unpaused.

imsodin:
All these issues were fixed, and I didn't encounter any new ones playing around (though I again didn't play around very thoroughly).

@tomasz1986 Does it behave fine for you?

tomasz1986:
I have found two issues with the newest iteration.

1. Creating a folder without any ignore patterns results in
image

2. If you tick "add ignore patterns" and click "save" first, but then click "close" in the new modal instead of "save", the folder is kept paused instead of being automatically unpaused.

imsodin:
Thanks again, fixed those too.

tomasz1986:
I have tested the newest version. I think there is a problem with the "remove on cancel" approach.

1. The folder is already being shared when the "add ignore patterns" modal is on the screen. As things are right now, the folder is automatically removed after clicking "cancel". However, the remote device can already accept it in between, which will likely create confusion.

2. This is related to the above, but when using "auto accept", the folder will almost surely be created on the remote device while still editing ignore patterns on the first device. If we click "cancel" and the folder is automatically removed, we will be left with a "ghost" folder on the remote device.

How about simply leaving the folder as it is after clicking "cancel", just without any ignore patterns?

imsodin:
Yes, I see. I incorrectly thought the paused folder wouldn't be sent in cluster-configs - but of course it is.

The semantics of "close" in this context aren't straight forward though: In any other context, it means leave as is. However "as is" isn't well defined on a new folder. If there was already a `.stignore`, I definitely wouldn't expect "close" to replace them with no patterns. And even if there aren't, and the displayed patterns are the default ones, it seems more consistent to me to set the default ignores on "close" - after all the user actively clicked "add patterns" before. Does this make any sense?

tomasz1986:
Yeah, of course I should've been more concrete in my wording 😜.

We've got three cases, right?

1. Pre-existing `.stignore` - "close" results in keeping it as is.
2. No `.stignore` but default patterns set - "close" results in filling `.stignore` with them.
3. No `.stignore` and no default patterns - "close" results in no `.stignore` created.

imsodin:
Ok, remove on cancel is gone and patterns are now saved when the modal is close/dismissed too.

tomasz1986:
I've gone through another round of testing.

I think there is still a case, where the behaviour may be unclear or unexpected. Let's say the user first ticks "add ignore patterns" and gets moved to the new modal. There, the default or pre-existing patterns are already present. The user removes or modifies them, but after doing so, they choose to "close" the modal rather than "save".

The current behaviour seems to be that the user-typed patterns are saved regardless and the default or pre-existing ones are wiped out no matter what. On the other hand, I would assume the user will likely be expecting that their typed patterns would replace the default or pre-existing only on "save", and not on "close" (or dismiss). The same applies to no pre-existing patterns, actually. The user may just type some gibberish for testing, but then decide to "close" or dismiss the modal. They will likely not expect such "test" patterns to be saved.

What do you think?

PS There's also a corner case, where the user closes or refreshes the GUI before clicking either of the buttons. There, no patterns are saved at all, but I think this can be called expected, right?

tomasz1986:
I'm not sure if this is the final version or not, but I've tested the newest iteration, and I can't find anything unexpected anymore 🙂.

Is there anything else that would prevent this PR from being merged?

imsodin:
For the past few iterations you have been the judge that it wasn't final. So if you are happy now, it's a review away from being merged :)

acolomb:
Had a rough look over the changes. But it's a bit complicated because of the new `_editing` attribute replacing the previous booleans. I need to take a closer look when back at the laptop.

One thing for the config proto, I'd name the message IgnorePatterns instead of Ignores.

imsodin:
> Had a rough look over the changes. But it's a bit complicated because of the new `_editing` attribute replacing the previous booleans. I need to take a closer look when back at the laptop.

Yeah, I didn't want to add yet another mutually exclusive status boolean, thus switched to something "enum-like". I added `editingExisting()` and `editingDefaults()` convenience functions to make the diff a bit less terrible.

> One thing for the config proto, I'd name the message IgnorePatterns instead of Ignores.

That's a bit of a pickle. Our naming there isn't great, but I think consisntent so far: The strings as present in the .stignore (now also config) are called `lines`, the string representation when those were parsed are named `patterns`. And when we are talking about the entire thing, it's "ignores". It would probably be nicer to name the entire thing `ignore-patterns` and then e.g. `input` and `parsed`? Anyway that would have to be changed in a lot of places, so I am hesitant to do anything about that in this PR.

gui/default/syncthing/folder/editFolderModalView.html
```diff
@@ -2,44 +2,44 @@
...
             <p class="help-block">
               <span translate ng-if="folderEditor.folderLabel.$valid || folderEditor.folderLabel.$pristine">Optional descriptive label for the folder. Can be different on each device.</span>
             </p>
           </div>
-          <div ng-if="!editingDefaults" class="form-group" ng-class="{'has-error': folderEditor.folderID.$invalid && folderEditor.folderID.$dirty}">
+          <div ng-if="!editingDefaults()" class="form-group" ng-class="{'has-error': folderEditor.folderID.$invalid && folderEditor.folderID.$dirty}">
             <label for="folderID"><span translate>Folder ID</span></label>
-            <input name="folderID" ng-readonly="editingExisting || (!editingExisting && currentFolder.viewFlags.importFromOtherDevice)" id="folderID" class="form-control" type="text" ng-model="currentFolder.id" required="" aria-required="true" unique-folder value="{{currentFolder.id}}" />
+            <input name="folderID" ng-readonly="['existing', 'add'].indexOf(currentFolder.Editing) != -1" id="folderID" class="form-control" type="text" ng-model="currentFolder.id" required="" aria-required="true" unique-folder value="{{currentFolder.id}}" />
```
acolomb, line 23: Use your new `has()` function.  And it should be `_editing`, no?

lib/fs/util.go
```diff
@@ -228,3 +229,22 @@ func isVolumeNameOnly(parts []string) bool {
 	isUNCVolumeName := len(parts) == 4 && strings.HasSuffix(parts[3], ":")
 	return isNormalVolumeName || isUNCVolumeName
 }
+
+// ReadLines is a utility to read a file at path in filesystem and returns
+// returns its contents as a string-slice of lines.
```
acolomb, line 234: Typo, double "returns"

lib/fs/util.go
```diff
@@ -228,3 +229,22 @@ func isVolumeNameOnly(parts []string) bool {
...
+	fd, err := filesystem.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer fd.Close()
+	scanner := bufio.NewScanner(fd)
+	var lines []string
+	for scanner.Scan() {
+		lines = append(lines, scanner.Text())
+	}
```
acolomb, line 245: Should we somehow limit the number of lines?  It's kind of a DoS attack vector to put a huge file as .stignore and then trying to add the folder will possibly jam Syncthing be eating all RAM?

acolomb (changes requested):
Found some errors in the new implementation.

Docs for the new API endpoint and the config entry still need to be written.

lib/fs/util.go
imsodin: There's no remote action involved, just the local user running a cli command. And what's the difference of opening the file in an editor or here, memory wise?

gui/default/syncthing/core/syncthingController.js
```diff
@@ -1578,14 +1578,16 @@ angular.module('syncthing.core')
...
+            switch ($scope.currentDevice._editing) {
+            case "existing":
                 title += $translate.instant("Edit Device");
-            } else {
+                break;
+            case "add":
                 title += $translate.instant("Add Device");
+                break;
+            case "defaults":
+                return $translate.instant("Edit Device Defaults");
```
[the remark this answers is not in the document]
imsodin: I don't get what this changes? `title` is empty at this point.

lib/fs/util.go
acolomb: Guess the only difference is that the file is is not being consciously opened, but a side effect of adding a Syncthing folder. Of course it's unlikely that someone else has put a huge file with that name somewhere for others to trip over.

gui/default/syncthing/core/syncthingController.js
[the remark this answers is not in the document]
acolomb: All other cases do not assume that the title is initially empty. Only this case would break if someone were to add something to the initial definition.

gui/default/syncthing/core/syncthingController.js
```diff
@@ -2033,39 +2055,61 @@ angular.module('syncthing.core')
...
-            $http.get(urlbase + '/config/defaults/folder')
-                 .success(function (data) {
-                     $scope.currentFolder = data;
-                     $scope.editingExisting = false;
-                     $scope.editingDefaults = true;
-                     editFolder();
-                 })
-                 .error($scope.emitHTTPError);
+            $q.all([
+                $http.get(urlbase + '/config/defaults/folder').then(function (response) {
```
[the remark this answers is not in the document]
imsodin: The `.success` is a deprecated angularjs thing, `.then` is the generic promises api. Thus I use `.then` whenever I anyway have to change something.

imsodin:
Thanks for the review @acolomb - I applied or responded to all the comments.

cmd/syncthing/cli/operations.go
```diff
@@ -38,6 +41,12 @@ var operationCommand = cli.Command{
 			ArgsUsage: "[folder id]",
 			Action:    expects(1, foldersOverride),
 		},
+		{
+			Name:      "default-ignores",
+			Usage:     "Set the default ignores (config) from a file",
+			ArgsUsage: "path",
```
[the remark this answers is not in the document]
acolomb: @imsodin missed this?

gui/default/syncthing/core/syncthingController.js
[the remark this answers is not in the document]
acolomb: And the suggested change was also to include a comment in the line returning.  It's easy to miss that it won't execute the rest of the function.

lib/fs/util.go
imsodin: This is only used when loading ignores from a file on disk to config.xml with `syncthing cli`. I also removed this function and put the code directly into `.../cli/operations.go`, I think I originally created this util function in earlier versions of this PR when had been used in multiple places.

cmd/syncthing/cli/operations.go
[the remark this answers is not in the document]
imsodin: Yes I did miss it. I am not a fan of all-caps and tend to avoid it wherever possible. Just had a look at `man man` and they say italic, respectively they actually use underscore. Other man-pages enclose it in `<>`. Anyway, I am consistent with one of the inconsistent options in this PR, so an issue for a different PR?

gui/default/syncthing/core/syncthingController.js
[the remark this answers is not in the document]
imsodin: I moved this condition to the beginning of the function as before the PR - seems the least missable.

lib/fs/util.go
acolomb: Ah sorry my confusion.

cmd/syncthing/cli/operations.go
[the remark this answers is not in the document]
acolomb: Alright.

acolomb (approved):

cmd/syncthing/cli/operations.go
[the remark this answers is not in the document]
acolomb: I just checked again and in `cmd/syncthing/main.go` we use all-caps for arguments consistently.  The documentation has angle brackets throughout.  It's mainly the cli component where it's inconsistent.

acolomb:
Nice to see this merged. Don't forget to update the docs though :wink:

imsodin:
I have it on my todo - if anyone feels like documenting it themselves, I'd also be happy :)
https://github.com/syncthing/docs/issues/712

PR #7535, opened 2021-03-30
lib/db: Fix comparison of pending folder timestamps (fixes #7532)

### LINKED ISSUE #7532: Removing pending folders is flaky
I didn't get a folder invitation that I should have, and adding a bit of additional debugging info I get the following:

```
Mar 30 13:19:48 homeserver syncthing[9815]: [BPDFD] INFO: Adding/updating pending folder f4vae-zg2uv (enc) from device BXAEQNX, last seen 2021-03-30 13:19:48 +0200 CEST
Mar 30 13:19:48 homeserver syncthing[9815]: [BPDFD] INFO: Unexpected folder "enc" (f4vae-zg2uv) sent from device "BXAEQNX-MC4GCEG-YQZGMDJ-4ME4SDW-TGHNYWO-UD2NWHL-XTJA3C4-NMC65QL"; ensure ...
Mar 30 13:19:48 homeserver syncthing[9815]: [BPDFD] INFO: Removing stale pending folder f4vae-zg2uv (enc) from device BXAEQNX, last seen 2021-03-30 11:19:48 +0000 UTC, oldest 2021-03-30 13:19:49 +020
```

`time.Now` called in the db after being called in model can apparently return an earlier time, or rounding to a second is buggy/does something unexpected. In any case calling `time.Now` for every entry is pointless: We do know which folders were updated, so one could pass a list/map of those and drop all other entries. Or just drop all entries before processing the cluster config. Or pass the timestamp from model to `RemovePendingFoldersBeforeTime`.

https://github.com/syncthing/syncthing/pull/7205
@acolomb

acolomb:
Regression introduced in #7459, this one instance of Round() was overlooked and not changed to Truncate().

### DIFF
--- a/lib/db/observed.go
+++ b/lib/db/observed.go
@@ -124,7 +124,7 @@ func (db *Lowlevel) RemovePendingFoldersBeforeTime(device protocol.DeviceID, old
 		return nil, err
 	}
 	defer iter.Release()
-	oldest = oldest.Round(time.Second)
+	oldest = oldest.Truncate(time.Second)
 	var res []string
 	for iter.Next() {
 		var of ObservedFolder


acolomb:
This was a regression introduced after v1.15.0-rc.3.  Should probably go to another rc?

PR #7536, opened 2021-03-30
docker: Raise UDP buffer size in example

bt90:
### Purpose

Raises the UDP buffer size limit similar to #7417 for docker deployments if the host network isn't used.

### Testing

I did not test it but it is explicitly stated in the docker documentation:

https://docs.docker.com/engine/reference/commandline/run/#configure-namespaced-kernel-parameters-sysctls-at-runtime

### DIFF
--- a/README-Docker.md
+++ b/README-Docker.md
@@ -13,7 +13,8 @@ altered with the ``PUID`` and ``PGID`` environment variables.
 
 ```
 $ docker pull syncthing/syncthing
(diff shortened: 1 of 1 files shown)


imsodin:
Shouldn't we just do that in the dockerfile, i.e. when creating the image, instead of telling the user to?

imsodin (approved):
Thanks.

calmh (approved):

PR #7537, opened 2021-03-30
lib/db, lib/model: Refactor removing expired pending folders

imsodin:
Remove expired pending folders without relying on timing. Also changed the update function signature to take a `ObservedFolder` instead of `string, string`, such that one can't switch label and id by mistake.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1306,13 +1306,17 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
 }
 
 func (m *model) ccHandleFolders(folders []protocol.Folder, deviceCfg config.DeviceConfiguration, ccDeviceInfos map[string]*indexSenderStartInfo, indexSenders *indexSenderRegistry) ([]string, map[string]struct{}, error) {
-	handleTime := time.Now()
 	var folderDevice config.FolderDeviceConfiguration
 	tempIndexFolders := make([]string, 0, len(folders))
 	paused := make(map[string]struct{}, len(folders))
 	seenFolders := make(map[string]struct{}, len(folders))
 	updatedPending := make([]updatedPendingFolder, 0, len(folders))
 	deviceID := deviceCfg.DeviceID
+	expiredPending, err := m.db.PendingFoldersForDevice(deviceID)
+	if err != nil {
+		l.Infof("Could not get pending folders for cleanup: %v", err)
+	}
+	of := db.ObservedFolder{Time: time.Now().Truncate(time.Second)}
 	for _, folder := range folders {
 		seenFolders[folder.ID] = struct{}{}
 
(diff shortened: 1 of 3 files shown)


lib/model/model.go
```diff
@@ -1412,18 +1418,16 @@ func (m *model) ccHandleFolders(folders []protocol.Folder, deviceCfg config.Devi
...
-	if err != nil {
-		l.Infof("Could not clean up pending folder entries: %v", err)
+	for folder := range expiredPending {
+		m.db.RemovePendingFolderForDevice(folder, deviceID)
 	}
 	if len(updatedPending) > 0 || len(expiredPending) > 0 {
-		expiredPendingList := make([]map[string]string, len(expiredPending))
-		for i, folderID := range expiredPending {
-			expiredPendingList[i] = map[string]string{
+		expiredPendingList := make([]map[string]string, 0, len(expiredPending))
```
acolomb, line 1425: Out of curiosity, what's the benefit of setting the length to zero, then `append()`ing?  Just readability?

acolomb (commented):
Code looks fine, but haven't tested it yet.  Just some questions below.

lib/model/model.go
AudriusButkevicius: `make([]Foo, 10)` allocates a slice with pre-allocated 10 items, so appending would result in the 11th item (first 10 having zero'ed values)

`0, len(foo)` allocates 0 sized slice with len(foo) capacity, so we are sure append will not need to re-allocate and copy the slice as we're appending (because it runs out of capacity)

acolomb: I understand that difference, but the previous approach was `L := make(..., 10)` and then *update* items `L[i]`, skipping `append()` completely.  Same overall effect, but any other disadvantages?

AudriusButkevicius: I don't think so, I prefer the latter as you don't need a variable for the index.

acolomb:
I don't see a need for this change, as the current mechanism works fine after the regression was fixed in #7535.  The idea here was to put a marker on each pending entry as we update them, then iterate through all of them and remove those without the marker.  Using the timestamp for the "marker" seems natural to me.  That scheme is simple and efficient.  The problem in #7532 was caused by a partial change from `.Round()` to `.Truncate()`, but at least those calls were not scattered across different files, deliberately, to avoid relying on different packages' rounding preference.

The proposal here duplicates the existing DB entries into a list in memory, to discard from that while updating and removing all remaining entries.  I don't really see a big benefit, as a more efficient solution is already in place.  The signature change is alright, and your implementation looks clean.  So I won't stand in the way, but just think this is unnecessary.

calmh:
The change removes code and simplifies, which is good. Assuming it does what it's supposed to, it's good with me.

calmh (approved):

PR #7539, opened 2021-03-30
gui: Modal dialog for listeners and discovery status.

### LINKED ISSUE #7743: Show discovery and listener status when not failed
In the GUI, the listener and discovery status is displayed as a count of active mechanisms. In case of a failure, details can be seen in a modal on click. The modal should be available always and show non-error status details as well.

acolomb:
### Purpose

Discovery failure info has been moved to a separate modal view in #4048.  The same treatment is in order for the listener failures. Additionally, it would be nice to know what these entries mean even when there are no failures. As discussed [on the forum](https://forum.syncthing.net/t/first-contact-what-are-listener/16526/14).

Replace the discoveryFailuresModalView with a unified connectivityStatusModalView, which is also suitable for displaying
listener failures and status for diagnosis.  Include some short descriptions about what each category means.

Summarize the running listeners in addition to the failed ones in a new array $scope.listenersRunning.

Fixes #7743.

**TODO:**
* [x] Include additional information about running (not failed) discovery mechanisms in the `/rest/system/status` API response.  Currently the discovery status dialog is empty without failures, because the info is just not there.
* [x] Possibly change the label for "Listeners" and "Discovery" as discussed in the forum thread.
* [x] Use different icons for the two dialog variants.

### Testing

* Provoked a discovery failure by enabling local discovery for two instances on the same machine --> shows error.
* Provoked a listener failure by manually setting a port below 1024 for the QUIC listener --> shows error.
* Verified that no errors are displayed when all is good.

### Screenshots

Screenshot_2021-03-30 devtest Syncthing_discoveryFailures
Screenshot_2021-03-30 devtest Syncthing_listenerFailures

### Documentation

Minor GUI usability change, no changed or removed functionality.
- [x] Slight change of API response for `GET /rest/system/status` to include `null` discovery "errors".

**~~If the labels end up being revised, screenshots and / or docs need to be revisited.~~**


acolomb:
> * Include additional information about running (not failed) discovery mechanisms in the `/rest/system/status` API response.  Currently the discovery status dialog is empty without failures, because the info is just not there.

Should I try to extend that API response within this PR?

imsodin:
I like the dialog and the info presented.

> Should I try to extend that API response within this PR?

Yes. both a new member with a list of discovery methods or including discovery methods with nil errors in `discoveryErrors` would make sense to me.

I'll look at the wording etc. once the naming discussion on the forum is done.

acolomb:
I noticed the same problem as mentioned in [an older PR](https://github.com/syncthing/syncthing/pull/7160#discussion_r532097273): When running `go run build.go translate`, the `lang-en.json` file always ends up empty.  What am I missing?

acolomb:
> Yes. both a new member with a list of discovery methods or including discovery methods with nil errors in `discoveryErrors` would make sense to me.

I went for the latter, which causes the smallest API change.  It is subtle though and prone to silent breakage, so maybe a formal API revision (renaming the field) would be better?

The `DiscoveryStatusEntry` stuff just mirrors what is there for listeners.  If that approach looks good, we might just remove the previous `ChildErrors()` method right away?

I still can't figure out what [goes wrong](https://github.com/syncthing/syncthing/pull/7539#issuecomment-819066132) with `go run build.go translate`, hence can't really test whether translation of the new dialog works :shrug:

gui/default/syncthing/core/connectivityStatusModalView.html
```diff
@@ -0,0 +1,54 @@
+<modal id="connectivity-status" status="{{connectivityStatusParams.status}}" icon="fas fa-fw {{connectivityStatusParams.type == 'listeners' ? 'fa-sitemap' : 'fa-map-signs'}}" heading="{{connectivityStatusParams.heading}}" large="no" closeable="yes">
+  <div class="modal-body" ng-switch="connectivityStatusParams.type">
+    <div ng-switch-when="listeners">
+      <div ng-if="listenersRunning.length > 0">
+	<p translate>
+	  The following network ports and protocols are available for other devices to connect to this one:
```
imsodin, line 6: Maybe use "listening" here on purpose to make it clear what listening means in this context for anyone confused about the term. Something like "Syncthing is listening on the following addresses for connection attempts from other devices:".

The "ports and protocols" formulation seems a bit out of place/too detailed. Below it's "ports or protocols", which isn't better. Both are just part of an address, and address is a less technical term that I'd consider more likely for any user to grasp. Basically I'd prefer using "address".

gui/default/syncthing/core/connectivityStatusModalView.html
```diff
@@ -0,0 +1,54 @@
...
+	  <li ng-repeat="listener in listenersRunning">{{listener}}</li>
+	</ul>
+      </div>
+      <div ng-if="listenersFailed.length > 0">
+	<p translate>
+	  Some network ports or protocols could not be enabled to accept connections:
+	</p>
+	<ul>
+	  <li ng-repeat="listener in listenersFailed">{{listener}}</li>
+	</ul>
```
imsodin, line 18: Indentation.

imsodin: Same below, and in one place it's 4 instead of 2 spaces. Easier to just let some formatter do its magic instead of commenting on each of them :)

gui/default/syncthing/core/connectivityStatusModalView.html
```diff
@@ -0,0 +1,54 @@
+<modal id="connectivity-status" status="{{connectivityStatusParams.status}}" icon="fas fa-fw {{connectivityStatusParams.type == 'listeners' ? 'fa-sitemap' : 'fa-map-signs'}}" heading="{{connectivityStatusParams.heading}}" large="no" closeable="yes">
+  <div class="modal-body" ng-switch="connectivityStatusParams.type">
+    <div ng-switch-when="listeners">
+      <div ng-if="listenersRunning.length > 0">
```
imsodin, line 4: Maybe display some text for `== 0`, saying that no listeners are active and connections can only be established by this device dialing others (or maybe leave out the second bit, might not be helpful/understandable).  
Similar thing for discovery.

lib/discover/manager.go
```diff
@@ -34,8 +34,11 @@ import (
 type Manager interface {
 	FinderService
 	ChildErrors() map[string]error
+	DiscoveryStatus() map[string]DiscoveryStatusEntry
```
imsodin, line 37: As `ChildErrors` returns all children/finders/discoverers, also if there's no error, I'd just remove it and only have a single method that returns status.

Naming wise we usually leave out `Discovery` here, as it's redundant with the package name - same as it isn't `DiscoveryManager`. Just `Status` doesn't convey that it's multiple ones. As we do already use the term `Finder` in exported parts of the package: maybe `FinderStatus`.

lib/discover/manager.go
```diff
@@ -34,8 +34,11 @@ import (
...
 }
 
+type DiscoveryStatusEntry *string
```
imsodin, line 40: What's the purpose of this instead of an error?

I mean I know about the json serialization shenanigans on the api side, but that doesn't (shouldn't) impact the choices here.

gui/default/index.html
```diff
@@ -641,28 +641,22 @@ <h4 class="panel-title">
                       </td>
                     </tr>
                     <tr>
-                      <th><span class="fas fa-fw fa-sitemap"></span>&nbsp;<span translate>Listeners</span></th>
+                      <th><span class="fas fa-fw fa-sitemap"></span>&nbsp;<span translate>Active Listening Addresses</span></th>
```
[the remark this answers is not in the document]
acolomb: I just went with the suggestion from @calmh here.  Having two (some / total) counts justifies to label what the difference in numbers signifies IMO.

gui/default/syncthing/core/connectivityStatusModalView.html
acolomb: Sorry, that's just my emacs HTML mode defaulting to mixed tabs and spaces.  Need to get that under control soon, as I often forget to `untabify`.

gui/default/syncthing/core/connectivityStatusModalView.html
acolomb: Good idea. Added for both listeners and discovery.

imsodin: Shouldn't the condition also include failed ones, i.e. running + failed == 0?

lib/discover/manager.go
imsodin: There's already a `errorString` convenience function to convert `error` to something json-friendly. I don't think it's sensible to avoid errors in go apis just to avoid that. And for future extensibility (e.g. adding fields), shouldn't it be `type DiscoveryStatusEntry struct{ err error }` to end up in a json object?

lib/api/api.go
```diff
@@ -1017,16 +1017,9 @@ func (s *service) getSystemStatus(w http.ResponseWriter, r *http.Request) {
...
-			discoMethods++
-			if err != nil {
-				discoErrors[disco] = err.Error()
-			}
-		}
-		res["discoveryMethods"] = discoMethods
-		res["discoveryErrors"] = discoErrors
+		discoStatus := s.discoverer.FinderStatus()
+		res["discoveryMethods"] = len(discoStatus) // DEPRECATED: Redundant, only for backwards compatibility, should be removed.
+		res["discoveryErrors"] = discoStatus
```
imsodin, line 1022: I'd add a "discoveryStatus" key and mark "discoveryErrors" as deprecated too, as the errors bit is misleading (I assumed it was only discovery methods with errors at first).

gui/default/index.html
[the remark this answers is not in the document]
imsodin: That doesn't change much that I am not a fan of making it longer like this. I am also not dead set on not having it. Just dreading e.g. the desig bug report about ultra-long labels not wrapping nicely (probably from a German for "Aktivierte zuhörende Addressen") :) 

As to having two counts and active: I could understand "active" in this context to mean active / available, while it is working / active. In my opinion it's the job of the modal to make clear what those two numbers mean.

gui/default/syncthing/core/connectivityStatusModalView.html
acolomb: That would show it only when no methods are *configured* at all.  I think the text also applies when none are running because all configured ones have failed.

gui/default/index.html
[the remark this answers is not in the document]
tomasz1986: > Just dreading e.g. the desig bug report about ultra-long labels not wrapping nicely (probably from a German for "Aktivierte zuhörende Addressen") :)

Long labels should wrap fine in the current state, but they don't look very nice on small screens. I would agree that 1-2 words maximum would indeed be more preferable.

lib/discover/manager.go
acolomb: There is even an `errorStringMap()` which seems to do just the right thing.  Will look into it and then probably revert everything back to just using `ChildErrors()` again.

I didn't switch to a struct right away to avoid a breaking change in the JSON API.  I get the feeling you're not reading my commit messages in full, which usually explain such things, e.g. https://github.com/syncthing/syncthing/pull/7539/commits/c793e08b1229ebf604ea8d78fea7440abab9aedf and https://github.com/syncthing/syncthing/pull/7539/commits/177058353f55d5e311265085810b0d881389b2e0 in this case.  Just a hint so there might be less questions on the next review :-)

lib/api/api.go
acolomb: This comes down to breaking the API or not.  I chose the path of least breakage when you said either would be okay.  Just trying not to increase the required effort for wrapper developers needlessly.  If you're more comfortable with different fields, I'll be happy to clean them up.

That would be only `discoveryStatus` then, replacing both keys.  In that case use child objects vor values as well?

gui/default/index.html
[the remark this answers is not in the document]
acolomb: I understood "active" to mean working, not configured.  Dunno about the translations, in the settings dialog it doesn't seem overly long.  Anyway, I was hoping for an opinion from @calmh, because he suggested these labels.

lib/api/api.go
imsodin: I meant adding a duplicate entry that's named `discoveryStatus` and has the same content as `discoveryErrors`, keeping `discoveryErrors` to not break api (but add a deprecation comment to be addressed whenever a v2 happens).

gui/default/index.html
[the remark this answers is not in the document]
imsodin: I like the modal. I don't like the name change and am not interested in discussing it. If the name change is reverted, I am happy to do a final round of review and merge this. You can still bring up the name change in a separately.

lib/api/api.go
acolomb: > It's not important though, it may also stay as is and one has to realize that it also contains empty errors for discovery methods that run successfully.

Then I'd say let's leave this in "backward-compatible with slight potential hiccup" state and clean up / rename whenever the deprecation notice on `discoveryMethods` will get addressed.

gui/default/index.html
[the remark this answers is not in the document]
acolomb: Sorry, I was under the impression that this was the consensus from the forum discussion.  You either didn's speak up loud enough there, or have a minority opinion ;-)

acolomb:
Added https://github.com/syncthing/docs/pull/647 regarding the API change.

lib/api/api.go
imsodin: Actually I didn't realize the api filtered out nil errors before. In that case, what's the downside of keeping the current `discoveryErrors` for backwards compat and add `discoveryStatus` with all methods? What we do now is not backwards compatible, and might trip api consumers that will now display a discovery as failed, even though it's running.  
Sorry for the confusion.

acolomb: > In that case, what's the downside of keeping the current `discoveryErrors` for backwards compat and add `discoveryStatus` with all methods?

Unnecessary duplication.

So should the new `discoveryStatus` be an dict of objects then, like `connectionServiceStatus`?  That would be easier to extend later on just in case.  Or just a plain string -> error mapping as we have now?

acolomb: Done.

acolomb:
I think this is worthy to be mentioned in the next change log.  Should I open a corresponding issue?

calmh:
Yes!

acolomb:
All reviews addressed AFAICT.  Anything else before this can be merged?

imsodin (approved):

PR #7552, opened 2021-04-06
build: Use nfpm for packaging (fixes #7548)

### LINKED ISSUE #7548: Upgrading syncthing-relaysrv to 1.15.0 fails in post-installation
* Versions: Upgrading the to the latest release (from candidate to stable) on ARM
* What happened: The dist-upgrade fails with an error in a post-installation script
* What was expected: your run-of-the-mill upgrade

Output:

```
Setting up syncthing-relaysrv (1.15.0) ...
/var/lib/dpkg/info/syncthing-relaysrv.postinst: 11: /var/lib/dpkg/info/syncthing-relaysrv.postinst: Syntax error: "else" unexpected
dpkg: error processing package syncthing-relaysrv (--configure):
 installed syncthing-relaysrv package post-installation script subprocess returned error exit status 2
Errors were encountered while processing:
 syncthing-relaysrv
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

imsodin:
Switches from fpm to nfpm to create debian packages.  
Written in go, no docker image or ruby required. Seems well maintained and if a version is faulty, it's easy to "control" through modules.

nfpm doesn't have the functionality to handle systemd units. I thus created a unified post-inst script using templating for all our packages. It restarts running systemd services if any, otherwise it sends `HUP`.

I changed the build tag for `build.go` to `build` such that `go mod tidy` doesn't remove the modules needed there (some new go1.16 thing).

Testing: Installed the syncthing package, did restart and still works. And diffing the new extracted .deb against the old only shows expected differences.

This also means we can retire the syncthing/builder docker image, once TC is adjusted to not use that.


script/deb-post-inst.template
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+command -v deb-systemd-invoke > /dev/null
+has_systemd=$?
+
+if [ $has_systemd -ne 0 ]; then
+	deb-systemd-invoke --daemon-reload
```
calmh, line 7: This is called if we *don't* have systemd?

calmh (commented):
Looks nice, possibly apart from the fact  that build.go now depends on a lot of stuff which will get downloaded when it's run, potentially causing all kinds of political ruckus, even when we're not building deb:s. Maybe this should go into a separate build-deb.go or something...

script/deb-post-inst.template
```diff
@@ -0,0 +1,14 @@
...
+fi
+
+if [ $has_systemd -ne 0 ] && [ -n "$(systemctl list-units --plain --no-legend {{.Service}})" ]; then
```
bt90, line 10:
```suggestion
if [ $has_systemd -eq 0 ] && deb-systemd-helper was-enabled {{.Service}} ; then
```

or something like that.

imsodin: `{{Service}}` can containg globbing, e.g. `syncthing@*.service`. Thus `is-enabled` doesn't work. Of course I totally agree that it is ugly, so I am very open to better ideas, just couldn't come up with them myself.

imsodin:
> Looks nice, possibly apart from the fact that build.go now depends on a lot of stuff which will get downloaded when it's run, potentially causing all kinds of political ruckus, even when we're not building deb:s. Maybe this should go into a separate build-deb.go or something...

go1.16 got more annoying (also known as strict) with regards to go.mod: It won't update it anymore on `go run`. Thus if you keep ignoring `build.go` and thus not adding the deps to go.mod, it will fail with something like:

```
build.go:36:2: no required module provides package github.com/goreleaser/nfpm/v2; to add it:
        go get github.com/goreleaser/nfpm/v2
build.go:37:2: no required module provides package github.com/goreleaser/nfpm/v2/deb; to add it:
        go get github.com/goreleaser/nfpm/v2/deb
build.go:38:2: no required module provides package github.com/goreleaser/nfpm/v2/files; to add it:
        go get github.com/goreleaser/nfpm/v2/files
```

I don't think there's any way around this as long as the source stays in the same module. We could create a file like `tools.go` with a `// +build: tools` tag to make it clearly visible (apparently that's a convention), where the imports that are just for building are listed (might also include some proto stuff, definitely that windows executable stuff, ...).

imsodin:
Looks like this is for go1.17. Would have to revert to nfpm v1 for go1.15 compat, which is quite different (well, semver), e.g. no file permissions, and likely unmaintained.

I'll file a new PR with just the post-inst script changes, replacing the systemd features and thus als ofixing #7552.

PR #7560, opened 2021-04-09
lib/api: Log the remote address on login attempts.

acolomb:
### Purpose

This enables usage of the audit log to e.g. automatically block remote addresses from connecting after repeated login failures, as requested [on the forum](https://forum.syncthing.net/t/can-the-syncthing-web-gui-report-authentication-errors/16642/11).

### Testing

Manual verification of the audit log contents when the API authorization is enabled.

### Documentation

**TODO**: No docs PR prepared yet, waiting on feedback whether this is the right approach or if there are any other security implications.

### DIFF
--- a/lib/api/api_auth.go
+++ b/lib/api/api_auth.go
@@ -29,10 +29,11 @@ var (
 	sessionsMut = sync.NewMutex()
 )
 
-func emitLoginAttempt(success bool, username string, evLogger events.Logger) {
+func emitLoginAttempt(success bool, username, address string, evLogger events.Logger) {
 	evLogger.Log(events.LoginAttempt, map[string]interface{}{
 		"success":  success,
 		"username": username,
+		"remoteAddress": address,
 	})
 }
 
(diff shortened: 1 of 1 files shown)


calmh:
Somewhat of a rabbit hole, I think if we do this we at minimum needs to handle x-forwarded-for headers, but perhaps then also not trust those headers if not sent by a proxy (127.0.0.1 or configured?)

AudriusButkevicius:
This is in response to a specific forum thread, yet reading that thread I felt that the right approach there was to put a reverse proxy, and use standard nginx/apache scrapers that feed into fail2ban etc.

calmh:
So this is one of those cases then where I think an issue describing the use case to be solved would be a good thing.

AudriusButkevicius:
Someone wants to write a fail2ban scraper that iptables blocks addresses that try to brute force basic auth.

For that purpose, the trust of X-Forwarded-For also has to be behind a config flag, as the attacker could send X-Forwarded-For to confuse syncthing, while it's actually not behind a reverse proxy.

acolomb:
I don't have an application for this myself, but I think the information could still be useful.  If the remote address is that of a reverse proxy, that's already valuable, as we know it's not someone else.  Deciding to block that whole proxy instead of the address for which the request was forwarded is up to whoever wants to write such rules.  Keep in mind Syncthing does not block anything by itself, but just gives a pointer to the immediate connecting host.

AudriusButkevicius (approved):
I guess this doesn't do any harm.

PR #7566, opened 2021-04-12
Making modals a bit more consistent

pvinis:
### Purpose

I tried to work on https://github.com/syncthing/syncthing/issues/6408 but I couldn't make it work. angular and jquery and all that is weird.. we should move the gui to react :D. 

Anyway, I tried to fix that issue but I couldn't, but on the way there, I made the 4 modals more consistent, so the code is the same for all four. It might get out of sync again if people change only one, so some funcs might have been better, but I didn't want to overcomplicate before I could actually fix the issue.

I added a `defaultPerPage` that the first time you open the modal is activated, and if you change it to say `25`, it will stay `25` on that modal.

### Testing

I tested it locally.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -11,6 +11,7 @@ angular.module('syncthing.core')
         var navigatingAway = false;
         var online = false;
         var restarting = false;
+		var defaultPerPage = 10;
 
         function initController() {
             LocaleService.autoConfigLocale();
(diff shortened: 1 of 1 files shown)


imsodin (commented):
Just in case I missed something: How does this make the modals more consistent? I only see that it persists the amount of items per page like you described.

pvinis:
It kinda makes thigs the same for all modals. Some were resetting things, some not. Some were assigning the return value of refresh, some not. Now they are all doing the same thing.

I feel like this would be better as some function that does a thing and then gets the modal and vars as args and does the same thing to all, but I'm not super familiar with jquery and backbone etc. I kinda wish this gui was in react 😅.

imsodin:
Just in case that wasn't clear: I am not opposed to this change, just wanted a clarification (which I got in the last comment). If the indentation issues (and now conflicts) are addressed this is lgtm.

st-review:
:robot: *beep boop*

PR #7567, opened 2021-04-12
all: Remove untrusted feature flag (fixes #109)

### LINKED ISSUE #109: Support for file encryption (e.g. non-trusted servers)

imsodin:
No longer hide the web UI controls for the new untrusted/encrypted
device feature. Testing hasn't been very widespread, but there has been
some and quite a few bugs have been caught and fixed. I believe its time
to not hide it anymore, and cautiously recommend usage. E.g. mention
that the feature hasn't been widely used yet and anyone using it is an
early adopter, but drop the bit about not using it with production data.
We can maybe stress the need for backups in general and especially
using this.

To remove the feature flag I added a new type of config migration like on dbs:
Changes something, buti not the schema.


calmh:
Cool, but I don't think we should do the migration schema version thing. It's a whole new dimension (and new field in the config, so this by itself is an API change anyway...) for not that much real use. Not many will have set the feature flag, and for those that have we can simply let if life on, it will be ignored and does no harm.

calmh (approved):

PR #7575, opened 2021-04-13
api: Log warnings on API authorization failures.

acolomb:
This is intended as an enhancement for #7560, to make failed login attempts more easily visible even when the audit log is disabled.  Inspired by [this forum discussion](https://forum.syncthing.net/t/can-the-syncthing-web-gui-report-authentication-errors/16642).  The log level is kept to Info to avoid message panels in an already connected GUI.

### DIFF
--- a/lib/api/api_auth.go
+++ b/lib/api/api_auth.go
@@ -57,8 +57,7 @@ func basicAuthAndSessionMiddleware(cookieName string, guiCfg config.GUIConfigura
 
 		l.Debugln("Sessionless HTTP request with authentication; this is expensive.")
 
-		error := func(reason string) {
-			l.Infoln("API authorization failed: ", reason)
+		error := func() {
 			time.Sleep(time.Duration(rand.Intn(100)+100) * time.Millisecond)
 			w.Header().Set("WWW-Authenticate", "Basic realm=\"Authorization Required\"")
 			http.Error(w, "Not Authorized", http.StatusUnauthorized)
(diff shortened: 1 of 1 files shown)


calmh (commented):
This'll emit a flurry of warnings when a browser visits, as the browser won't send credentials until it gets the unauthorized status back. Really only the case at the bottom (where we call `emitLoginAttempt(false, ...)`) is an auth failure. Perhaps the logging should happen in that function instead.

acolomb:
The only occasion I could produce a "flood" of about 8 failure messages, was when restarting Syncthing with the browser page still open, but after the password was changed.  Otherwise it's always just one entry saying "wrong authorization type", then the browser stopped connecting to ask for the credentials.  What I observed was at least not excessive logging, although the no authorization case could arguably be disregarded.  I think the other messages might help diagnosing problems.

I don't have a strong opinion, so if you think these are too much, I will reduce to only the "wrong user / password" case.

imsodin:
> Really only the case at the bottom (where we call `emitLoginAttempt(false, ...)`) is an auth failure.

I concur.

AudriusButkevicius:
Likewise.

lib/api/api_auth.go
```diff
@@ -97,6 +97,7 @@ func basicAuthAndSessionMiddleware(cookieName string, guiCfg config.GUIConfigura
 
 		if !authOk {
 			emitLoginAttempt(false, username, r.RemoteAddr, evLogger)
+			l.Infoln("Wrong credentials supplied during API authorization")
```
imsodin, line 100: You could also include the remote address, then someone wanting to scrape logs can do it directly without the need for `--audit`?

calmh: I'd also still do this in `emitLoginAttempt`. Reduces ambiguity, more future proof.

imsodin (approved):

PR #7591, opened 2021-04-22
cmd/syncthing: Don't fail early on api setup error (fixes 7558)

imsodin:
Instead of immediately exiting when encountering an error setting up the API client, the error is stored and returned if the api client is used. That ensures that the error does not come up if no operation involving the API is happening (like just querying `--help`).

To do so I added an interface for the api client and an `errorAPIClient` type that returns the error from setup when on all of it's methods.

### DIFF
--- a/cmd/syncthing/cli/config.go
+++ b/cmd/syncthing/cli/config.go
@@ -46,13 +46,24 @@ func getConfigCommand(f *apiClientFactory) (cli.Command, error) {
 		HideHelp:    true,
 		Usage:       "Configuration modification command group",
 		Subcommands: commands,
+		Before:      h.configBefore,
 		After:       h.configAfter,
 	}, nil
 }
 
+func (h *configHandler) configBefore(c *cli.Context) error {
+	for _, arg := range c.Args() {
+		if arg == "--help" || arg == "-h" {
+			return nil
+		}
+	}
+	return h.err
+}
+
 func (h *configHandler) configAfter(c *cli.Context) error {
 	if h.err != nil {
-		return h.err
+		// Error was already returned in configBefore
+		return nil
 	}
 	if reflect.DeepEqual(h.cfg, h.original) {
 		return nil
--- a/cmd/syncthing/cli/config.go
+++ b/cmd/syncthing/cli/config.go
@@ -1,0 +1,76 @@
+// Copyright (C) 2021 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package cli
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+
+	"github.com/AudriusButkevicius/recli"
+	"github.com/pkg/errors"
+	"github.com/syncthing/syncthing/lib/config"
+	"github.com/urfave/cli"
+)
+
+type configHandler struct {
+	original, cfg config.Configuration
+	client        APIClient
+	err           error
+}
+
+func getConfigCommand(f *apiClientFactory) (cli.Command, error) {
+	h := new(configHandler)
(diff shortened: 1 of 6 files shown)


AudriusButkevicius:
This is actually better than anything else I came up with.

I wonder if we could go a step further and generate a dummy config if looking up a proper one fails, to still wire the subcommands, so things like `cli config devices --help` still works etc, just returns some single garbage folder and single garbage device which works for help purposes, and then when it gets to handling the result of the command, just aborts because it knows it used a placeholder config.

As a side venture, you could also turn getClient into getClientFactory which the commands themselves call (which then constructs the client). I guess this doesn't help in any way as we retrieve the config for the purpose of generating commands of devices available to start with, so this will be called regardless even in case of --help.

Also, if we fail to call the API to get the config, we could try loading it from disk, so the CLI would still work, but instead of fiddling the api, it would fiddle the file on disk.

imsodin:
Nice ideas. I did all of that except for managing the config on disk when the api is down. Definitely doable (and useful), but I don't want to do it in this PR (just because I don't want to do it right now :) ).

It still shows the error when passing `--help`. I tried to check for the help flag in `afterConfig`, but `cli.Context.IsSet("help")` always returns false and I didn't manage to get that information in any way.

cmd/syncthing/cli/utils.go
```diff
@@ -51,7 +51,7 @@ func indexDumpOutput(url string) cli.ActionFunc {
 
 func saveToFile(url string) cli.ActionFunc {
 	return func(c *cli.Context) error {
-		client := c.App.Metadata["client"].(*APIClient)
+		client := c.App.Metadata["client"].(APIClient)
```
AudriusButkevicius, line 54: I think this is now never set, hence should fail?

AudriusButkevicius:
I think help might be a global flag, so you might need to check `GlobalIsSet`.

Also found the following in the code:
```
func checkCommandHelp(c *Context, name string) bool {
	if c.Bool("h") || c.Bool("help") {
		_ = ShowCommandHelp(c, name)
		return true
	}

	return false
}

func checkSubcommandHelp(c *Context) bool {
	if c.Bool("h") || c.Bool("help") {
		_ = ShowSubcommandHelp(c)
		return true
	}

	return false
}
```

imsodin:
I tried all of that (also with `cli.HelpFlag.Name()`) - I don't have any interest in debugging why it doesn't work. I had a quick look at the code and it's convoluted - printing the error together with the help doesn't seem that bad to me.

cmd/syncthing/cli/config.go
```diff
@@ -0,0 +1,76 @@
...
+	}, nil
+}
+
+func (h *configHandler) configAfter(c *cli.Context) error {
+	if h.err != nil {
+		return h.err
+	}
+	if reflect.DeepEqual(h.cfg, h.original) {
+		return nil
+	}
```
AudriusButkevicius, line 59: If you moved this above the h.err check, --help should not modify the config, in which case we'd end up not returning the error.

AudriusButkevicius: You need a util.FillNil on the default config it seems, but other than that, seems to work for me.

You can check for "--help" in ctx.Args() apparently, but I feel the DeepEquals early termination is probably less hacky.

imsodin: That's not ok because it means we don't return the error on an actual (read) operation, but just the default value and success.

AudriusButkevicius: In that case you don't even want to run the command at all.

How about adding a Before handler, if it doesn't have `--help` or `-h` in args, and error is set, just abort the command all together.

And then remove the error check in the After command.

AudriusButkevicius (approved):

cmd/syncthing/cli/config.go
imsodin: Just came to the same conclusion after posting my last comment. However I kept the error check, because `After` is called regardless of whether `Before` returned a non-nil error or not.

Issue #7379, opened 2021-02-18
"Warning, this path is a subdirectory of an existing folder" does not appear when editing only folder label


tomasz1986:
1. Share a new folder between devices A and B.
2. On Device B, try to add the folder under the same path as one of the existing folders.

3. Set the path by editing only the folder label. No warning appears.

   ![1](https://user-images.githubusercontent.com/5626656/108396906-bc7f1400-725a-11eb-947d-4b2107d25de8.png)

4. Set the path by editing the folder path directly. The warning shows up.

   ![2](https://user-images.githubusercontent.com/5626656/108396967-cc96f380-725a-11eb-90e2-e4cee5e259d3.png)

**Edit**: The same is likely true for all the other similar warnings.

PR #7595, opened 2021-04-23
gui: Handle empty path in validation (ref #7379)

imsodin:
The validation now can also kick in when the path is empty, which produced a JS error.


imsodin:
This needs to go into the final v1.16.0 RC tomorrow, as it's an issue that didn't previously exist (introduced in #7379).

calmh (approved):

calmh:
Noted on the RC

PR #7596, opened 2021-04-23
lib/model: Handle invalid needed items on send-only (ref #7476)

imsodin:
I noticed a weird failure event on sentry: `got deleted file that doesn't exist locally as needed when pulling on send-only`
https://sentry.syncthing.net/syncthing/syncthing/issues/10537/?environment=Stable&referrer=alert_email
Its informational only - the condition is not acted upon (the item is just ignored). It will lead to incorrect remote out-of-sync status for the affected device on other devices.

Most likely the cause is #7476, where now invalid files are considered needed. In that PR I only handled that case in send-receive folders, and forgot about the send-only folder having its own `pull` method.

I also took the opportunity to use `fileInfoBatch` instead of the "legacy manual way" (no behaviour change).

### DIFF
--- a/lib/model/folder_sendonly.go
+++ b/lib/model/folder_sendonly.go
@@ -37,8 +37,10 @@ func (f *sendOnlyFolder) PullErrors() []FileError {
 
 // pull checks need for files that only differ by metadata (no changes on disk)
 func (f *sendOnlyFolder) pull() (bool, error) {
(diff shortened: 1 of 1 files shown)


imsodin:
I'd like for this to go into the final v1.16.0 RC, as it relates to remote out-of-sync status (aka fixes a cause) and that's basically the main (and almost only) kind of issue we are debugging at the moment in support topics.

calmh (approved):

PR #7597, opened 2021-04-24
Avoid Semicolon Autoinsertion

0verk1ll:
### Purpose

This addresses recommendations from LGTM about Semicolon insertion (https://lgtm.com/rules/4860080/). This change makes the JavaScript more consistent and predictable.

### Testing

I have not performed any testing. I do not believe the function of the program will be changed by this pull request. Automated testing should be sufficient for this PR.

### Documentation

This will not change the way the code runs for the user.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -42,7 +42,7 @@ angular.module('syncthing.core')
         $scope.pendingFolders = {};
         $scope.progress = {};
         $scope.version = {};
-        $scope.needed = {}
+        $scope.needed = {};
         $scope.neededFolder = '';
         $scope.failed = {};
         $scope.localChanged = {};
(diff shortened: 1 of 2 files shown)


gui/default/syncthing/core/syncthingController.js
```diff
@@ -1769,7 +1769,7 @@ angular.module('syncthing.core')
             }
 
             return false
```
greatroar, line 1771: Shouldn't there be a semicolon here too?

0verk1ll: Yes, I added the semicolon.
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/return, "The `return` statement is affected by automatic semicolon insertion (ASI)."

calmh (approved):

PR #7598, opened 2021-04-25
all: Deprecate TLS 1.2 on sync connections (fixes #7594)

### LINKED ISSUE #7594: Consider removing support for TLS <1.3 on sync connections
Historically we've been fairly aggressive about adopting and then mandating secure crypto on the connection. TLS 1.2 is still alive and well, but we have a lot of cipher suite magic to ensure we get a secure connection. We could remove all that, simplifying our TLS config, and say that sync connections (*not* GUI/API connections) are now TLS 1.3 only.

The oldest Syncthing release that understands TLS 1.3 is 1.3.0 from October 1, 2019. By current data.s.n measurements we have ~94.6% of the population on 1.3.0 or newer. I would expect that those who do run old versions probably do that for a reason and probably do it somewhat consistently as well (that is, old talking to old rather than old talking to newest). Connections to older devices will be break with `Listen (BEP/tcp): TLS handshake: tls: client offered only unsupported versions: [303]`. ("303" is TLS 1.2, the other side will say the same but about "304" (TLS 1.3) instead.)

I'm not aware of any non-Syncthing BEP implementations we would need to be compatible with.

Nonetheless it would be the first intentional compatibility break for a long time, and in theory would require a Syncthing 2.0 release. (I *don't* think we should do a 2.0, because that would imply incompatibility with 1.3.0+ as well.) We could also retain the TLS 1.2 uglyness hidden behind a feature flag or environment var as a bailout for a while.

ref #7593

calmh:
This makes us use TLS 1.3+ on sync connections by default. A new option
`insecureAllowOldTLSVersions` exists to allow communication with TLS
1.2-only clients (roughly Syncthing 1.2.2 and older). Even with that
option set you get a slightly simplified setup, with the cipher suite
order fixed instead of auto detected.


AudriusButkevicius:
Is this backwards compatible with old clients? I assume they support 1.3, hence would try to negotiate that regardless?

calmh:
I discussed that in the corresponding ticket, but in short 1.3.0+ are compatible, 1.2.2 and lower fails to connect. (Plus/minus a little for other packagers as this depends on having build with Go 1.12, not our code.)

AudriusButkevicius (approved):

PR #7599, opened 2021-04-25
build: List go:generate tools in tools.go

greatroar:
This reuses [conventional go build machinery](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) to simplify build.go, as suggested by @imsodin at #7552. `go generate` just works, without having genxdr or counterfeiter installed. `go mod tidy` no longer removes counterfeiter.

There's more work to be done on this front, in particular wrt. protobufs, but I figured I'd send a PR to test the waters and see if you want to move in this direction at all.

### DIFF
--- a/build.go
+++ b/build.go
@@ -202,18 +202,6 @@ var targets = map[string]target{
 	},
 }
 
-// These are repos we need to clone to run "go generate"
-
-type dependencyRepo struct {
-	path   string
-	repo   string
-	commit string
-}
-
-var dependencyRepos = []dependencyRepo{
-	{path: "xdr", repo: "https://github.com/calmh/xdr.git", commit: "08e072f9cb16"},
-}
-
 func initTargets() {
 	all := targets["all"]
 	pkgs, _ := filepath.Glob("cmd/*")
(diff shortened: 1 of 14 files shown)


calmh:
Sounds good 👍

AudriusButkevicius (approved):
lgtm

imsodin (changes requested):
Why is the `github.com/gogo/protobuf` repo still needed - it's also part of the new tools.go and thus should be provided like that, shouldn't it?

greatroar:
Protobufs takes more work due to the need to import a protobuf file from the gogo/protobuf repo. I'll leave the PR at this for now.

greatroar:
Oh and the protoc compiler (which has to still be installed separately) needs to invoke the protoc-gen-gogo command, so that has to be compiled and in its PATH.

imsodin (approved):
Thanks for the explanation, lgtm too.

greatroar:
Rebased on main. By the way, you may want to re-run `build.go proto` and check in the result. It changes two imports on main.

greatroar:
Rebased again.

imsodin (approved):

PR #7604, opened 2021-04-26
Trigger connection loop on config device addition (fixes #7600)

### LINKED ISSUE #7600: Fast connect to new devices following config update
We are using syncthing as part of a larger system, as an end-to-end solution for file synchronization between two [or more] endpoints. It is working well for us, but we have encountered a problem with the time it takes syncthing to register that a new device was added and connect to it.

In our system, syncthing is running in the background, and at a random time we may change the devices to which we sync the folder[s]. We do this using the REST API - `/rest/config/devices`. Following that we listen on events that inform us that the connection has succeeded and initial content sync completed.

The problem is that after the config update it can take up to a minute (`stdConnectionLoopSleep`) for syncthing to actually reach out to the remote device, and in an interactive context that is a **long** time. This affects all users of the REST API - I think reducing the time it takes to implement config updates would be useful to everyone 😄 

From reading around the code, it seems a relatively simple solution of adding a "deviceConfigUpdated" channel to the select in the loop in `lib/connections/service.go:connect`, and sending a signal from the REST api handlers once the config is updated, should solve. I'd be happy to work on this 🙂

gsaraf:
### Purpose
In order to fix issue #7600, we use the existing config commit listener in the connections service to check for new devices. If any exist, we signal on a new channel for the connection loop to stop sleeping.

There is technically a small edge case between the channel check and the config copy where if the config is updated during that time we would end up with two dialDevices immediately following each other. This is extremely low probability, so not addressed currently.

### Testing
1) Run syncthing in the background
2) Use api to add new device (POST to `/rest/config/devices`)
3) See from GUI that device is immediately connected to.

I'd be happy to make any changes needed. Thanks!

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -129,17 +129,18 @@ type service struct {
 	*suture.Supervisor
 	connectionStatusHandler
 
-	cfg                  config.Wrapper
-	myID                 protocol.DeviceID
-	model                Model
-	tlsCfg               *tls.Config
-	discoverer           discover.Finder
-	conns                chan internalConn
-	bepProtocolName      string
-	tlsDefaultCommonName string
-	limiter              *limiter
-	natService           *nat.Service
-	evLogger             events.Logger
+	cfg                    config.Wrapper
+	deviceAddressesChanged chan struct{}
+	myID                   protocol.DeviceID
+	model                  Model
+	tlsCfg                 *tls.Config
+	discoverer             discover.Finder
+	conns                  chan internalConn
+	bepProtocolName        string
+	tlsDefaultCommonName   string
+	limiter                *limiter
+	natService             *nat.Service
+	evLogger               events.Logger
 
 	listenersMut       sync.RWMutex
 	listeners          map[string]genericListener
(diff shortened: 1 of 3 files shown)


lib/connections/service.go
```diff
@@ -733,6 +739,21 @@ func (s *service) CommitConfiguration(from, to config.Configuration) bool {
...
 
+func (s *service) checkAndSignalConnectLoopOnUpdatedDevices(from, to config.Configuration) {
+	oldDevices := make(map[protocol.DeviceID]config.DeviceConfiguration, len(to.Devices))
+	for _, dev := range from.Devices {
+		oldDevices[dev.DeviceID] = dev
+	}
+
+	for _, dev := range to.Devices {
+		oldDev, ok := oldDevices[dev.DeviceID]
+		if !ok || !reflect.DeepEqual(oldDev.Addresses, dev.Addresses) {
```
AudriusButkevicius, line 750: Using reflect for comparing string slices is grim. I'd expect something to exist in the standard library.

imsodin: I don't think that exists and we already did implement this super elaborate function in `api_test.go` as `equalStrings`. Could move that to the `lib/utils` package for everyone to use.

gsaraf: Awesome, I'll move it to lib/utils.

lib/connections/service.go
```diff
@@ -130,6 +130,7 @@ type service struct {
 	connectionStatusHandler
 
 	cfg                  config.Wrapper
+	configDevicesUpdated chan bool
```
AudriusButkevicius, line 133: Feel free to ignore this, but it could be a zero sized `struct{}`.

AudriusButkevicius (dismissed):
lgtm

lib/connections/service.go
imsodin: Equally ignorable (cause bike-shedding): `deviceAddressesChanged` would be more accurate in my opinion.

AudriusButkevicius (dismissed):

lib/connections/service.go
```diff
@@ -129,17 +129,18 @@ type service struct {
...
-	tlsCfg               *tls.Config
-	discoverer           discover.Finder
-	conns                chan internalConn
-	bepProtocolName      string
-	tlsDefaultCommonName string
-	limiter              *limiter
-	natService           *nat.Service
-	evLogger             events.Logger
+	cfg                    config.Wrapper
+	deviceAddressesChanged chan struct{}
```
calmh, line 133: The order of the early variables in the struct matches those of passed in parameters, I'd put this further down (maybe after the `evLogger`) as it's an internally constructed thing.

lib/connections/service.go
```diff
@@ -733,6 +738,25 @@ func (s *service) CommitConfiguration(from, to config.Configuration) bool {
 	return true
 }
 
+func (s *service) checkAndSignalConnectLoopOnUpdatedDevices(from, to config.Configuration) {
+	oldDevices := make(map[protocol.DeviceID]config.DeviceConfiguration, len(to.Devices))
```
calmh, line 742:
```suggestion
	oldDevices := make(map[protocol.DeviceID]config.DeviceConfiguration, len(from.Devices))
```

calmh (approved):

PR #7609, opened 2021-04-29
lib: Revert needing invalid files (fixes #7608, ref #7476)

### LINKED ISSUE #7474: Device tries to push ignored files to a remote device in a 3-device setup

### LINKED ISSUE #7608: Files ignored on one remote do not get synced
Came up in https://forum.syncthing.net/t/syncthing-lost-one-file-out-of-a-dozen-on-one-node/16721.

Reproducer:

Initial State: Devices A, B share a folder. A has file F, B ignores file F.

1. Connect additional device C to B, wait for index exchange and sync/pull to finish.
2. Connect C to A.

Expected: File F gets synced to C.

Result: File F does not get synced to C.

This is a regression introduced in https://github.com/syncthing/syncthing/pull/7476, which in turn fixes https://github.com/syncthing/syncthing/issues/7474.

imsodin:
Unfortunately the solution for #7474 implemented in #7476 created a new, more severe problem: Files that should be synced, aren't: #7608. I couldn't come up with a way to fix #7474 without introducing this regression that doesn't add a lot of complexity. Thus I reverted that change and thus re-introduce #7474. As that is only a "presentation issue" that's clearly the more desirable issue to have.

Most of the diff here is due to the migration needed to resolve issues for users already affected by #7608. As I needed a file info batch in this migration, I moved the `fileInfoBatch` from model to db.

### Testing

I introduced a new unit test reproducing #7608 and removed the test for #7474.

### DIFF
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -20,7 +20,7 @@ import (
 // do not put restrictions on downgrades (e.g. for repairs after a bugfix).
 const (
 	dbVersion             = 14
-	dbMigrationVersion    = 16
+	dbMigrationVersion    = 17
 	dbMinSyncthingVersion = "v1.9.0"
 )
 
@@ -101,6 +101,7 @@ func (db *schemaUpdater) updateSchema() error {
 		{13, 13, "v1.7.0", db.updateSchemaTo13},
 		{14, 14, "v1.9.0", db.updateSchemaTo14},
 		{14, 16, "v1.9.0", db.checkRepairMigration},
+		{14, 17, "v1.9.0", db.migration17},
 	}
 
 	for _, m := range migrations {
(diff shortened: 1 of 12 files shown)


tomasz1986:
That is unfortunate…

Is https://github.com/syncthing/syncthing/issues/7474 going to be reopened then? Maybe something or someone will come up later and make it possible to fix it. I can live with the cosmetic issue, but at the same time it makes the affected devices appear permanently out of sync, which is less than ideal…

calmh (approved):

PR #7624, opened 2021-05-01
lib/fs, lib/api, lib/model: Expose mtime remappings as part of /db/file

AudriusButkevicius:
Also fixes cli interpreting status codes incorrectly,
It assumes that 403 (Forbidden (Debugging disabled)) as a invalid api key.

### DIFF
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -659,7 +659,6 @@ func (fs *fakeFs) SameFile(fi1, fi2 FileInfo) bool {
 	return ok && fi1.ModTime().Equal(fi2.ModTime()) && fi1.Mode() == fi2.Mode() && fi1.IsDir() == fi2.IsDir() && fi1.IsRegular() == fi2.IsRegular() && fi1.IsSymlink() == fi2.IsSymlink() && fi1.Owner() == fi2.Owner() && fi1.Group() == fi2.Group()
 }
 
-
 func (fs *fakeFs) underlying() (Filesystem, bool) {
 	return nil, false
 }
--- a/lib/fs/fakefs.go
+++ b/lib/fs/fakefs.go
@@ -27,7 +27,7 @@ import (
 // see readShortAt()
 const randomBlockShift = 14 // 128k
 
-// fakefs is a fake filesystem for testing and benchmarking. It has the
+// fakeFs is a fake filesystem for testing and benchmarking. It has the
 // following properties:
 //
 // - File metadata is kept in RAM. Specifically, we remember which files and
@@ -37,7 +37,7 @@ const randomBlockShift = 14 // 128k
 // - File contents are generated pseudorandomly with just the file name as
 //   seed. Writes are discarded, other than having the effect of increasing
 //   the file size. If you only write data that you've read from a file with
-//   the same name on a different fakefs, you'll never know the difference...
+//   the same name on a different fakeFs, you'll never know the difference...
 //
 // - We totally ignore permissions - pretend you are root.
 //
@@ -51,9 +51,9 @@ const randomBlockShift = 14 // 128k
 //     insens=b   "true" makes filesystem case-insensitive Windows- or OSX-style (default false)
 //     latency=d  to set the amount of time each "disk" operation takes, where d is time.ParseDuration format
 //
-// - Two fakefs:s pointing at the same root path see the same files.
+// - Two fakeFs:s pointing at the same root path see the same files.
 //
-type fakefs struct {
+type fakeFs struct {
 	counters    fakefsCounters
 	uri         string
 	mut         sync.Mutex
@@ -82,10 +82,10 @@ type fakefsCounters struct {
 
 var (
 	fakefsMut sync.Mutex
-	fakefsFs  = make(map[string]*fakefs)
+	fakefsFs  = make(map[string]*fakeFs)
 )
 
-func newFakeFilesystem(rootURI string, _ ...Option) *fakefs {
+func newFakeFilesystem(rootURI string, _ ...Option) *fakeFs {
 	fakefsMut.Lock()
 	defer fakefsMut.Unlock()
 
@@ -102,7 +102,7 @@ func newFakeFilesystem(rootURI string, _ ...Option) *fakefs {
 		return fs
 	}
 
-	fs := &fakefs{
+	fs := &fakeFs{
 		uri: "fake://" + rootURI,
 		root: &fakeEntry{
 			name:      "/",
@@ -183,7 +183,7 @@ type fakeEntry struct {
 	content   []byte
 }
 
-func (fs *fakefs) entryForName(name string) *fakeEntry {
+func (fs *fakeFs) entryForName(name string) *fakeEntry {
 	// bug: lookup doesn't work through symlinks.
 	if fs.insens {
 		name = UnicodeLowercase(name)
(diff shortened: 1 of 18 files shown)


lib/fs/mtimefs.go
```diff
@@ -211,42 +215,44 @@ func (f mtimeFile) Stat() (FileInfo, error) {
...
 	return f.File
 }
 
-// The dbMtime is our database representation
-
-type dbMtime struct {
-	real    time.Time
-	virtual time.Time
+// The DBMtime is our database representation
+type DBMtime struct {
```
calmh, line 238: If we expose this we should give it a less crappy name, maybe something like `MtimeMapping` or so?

calmh (commented):
Neat

lib/fs/filesystem.go
```diff
@@ -16,6 +16,18 @@ import (
...
+const (
+	FilesystemVariantBasic FilesystemVariant = iota
+	FilesystemVariantMtime
+	FilesystemVariantCase
+	FilesystemVariantError
+	FilesystemVariantWalk
+	FilesystemVariantFake
+	FilesystemVariantLog
+)
+
```
imsodin, line 30: Currently this has overlap with `FilesystemType`, as in basic and fake exist there too. To make it clearly distinct from `FilesystemType`, this could be `FilesystemWrapperType` and basic and fake become just `FilesystemWrapperTypeNoWrapper` (or anything similar, naming being hard as always). Respectively basic and fake don't need a wrapper-type at all, as they wrap nothing, and just return false on their `underlying` and `variant` methods.

lib/fs/fakefs.go
```diff
@@ -51,9 +51,9 @@ const randomBlockShift = 14 // 128k
 //     insens=b   "true" makes filesystem case-insensitive Windows- or OSX-style (default false)
 //     latency=d  to set the amount of time each "disk" operation takes, where d is time.ParseDuration format
 //
-// - Two fakefs:s pointing at the same root path see the same files.
+// - Two fakeFs:s pointing at the same root path see the same files.
 //
-type fakefs struct {
+type fakeFs struct {
```
imsodin, line 56: Usually we fully capitalize abbrevs, i.e. `fakeFS`, like in `MtimeFS`.

imsodin (commented):
Exposing the wrapping business feels bad. That's a somewhat ugly implementation detail. What about having a function `fs.MtimeMapping(file string) (fs.MtimeMapping, error)` that returns `fs.ErrNotMtimeFS` if called on not an mtime fs?

AudriusButkevicius:
```
fs.MtimeMapping(file string) (fs.MtimeMapping, error)
```

I agree the unwrapping stuff is not great, but I think that is even more ugly, because suddenly every filesystem exposes some sort of method which is just a specialisation to one of them?

Unless you are suggesting:
```
func MtimeMapping(fs.Filesystem, file string) (fs.MtimeMapping, error)
```

Which I guess keeps the cruft internal and is not horrible.

I think the methods for unwrapping, and indication of the type makes things cleaner in general, even if we decide to not expose them, because it allows to pick how much unwrapping you want to do.

Ideally, I'd pass a type to the unwrap method, but because type references don't exist in go, I've opted for enums.

lib/fs/filesystem.go
AudriusButkevicius: Agree

imsodin:
> Unless you are suggesting:
> 
> ```
> func MtimeMapping(fs.Filesystem, file string) (fs.MtimeMapping, error)
> ```

That was what I meant :)

Alternatively maybe even a bit cleaner: Have an interface

```golang
type MtimeMapper interface {
	MtimeMapping(file string) (MtimeMapping, error)
}
```

and then use it like

```golang
	if mapper, ok := mtimeFs.(fs.MtimeMapper); !ok {
		return fs.MtimeMapping{}, errors.New("unwrapping failed")
	} else {
		return mapper.MtimeMapping(file)
	}
```

lib/fs/filesystem.go
```diff
@@ -273,9 +288,9 @@ func Canonicalize(file string) (string, error) {
 // It ensures proper wrapping order, which right now means:
 // `logFilesystem` needs to be the outermost wrapper for caller lookup.
 func wrapFilesystem(fs Filesystem, wrapFn func(Filesystem) Filesystem) Filesystem {
-	logFs, ok := fs.(*logFilesystem)
+	logFs, ok := UnwrapFilesystem(fs, FilesystemWrapperTypeLog)
```
imsodin, line 291: I'd keep the direct type assertion: This functions is about the order of wrapping, while `UnwrapFilesystem` doesn't care about the order and just gets the wrapper from any "depth". And as the comment above explains, the logging wrapper needs to be the outermost at all times.

AudriusButkevicius:
That would not work, as mtimeFs might be wrapped in logFs/walkFs etc, which embeds Filesystem, so I don't think it would ever satisfy this interface.

lib/fs/folding_test.go
```diff
@@ -25,7 +25,7 @@ var caseCases = [][2]string{
 	// context; we should always choose the same one.
 	{"Ὀδυσσεύς", "ὀδυσσεύσ"},
 	{"ὈΔΥΣΣΕΎΣ", "ὀδυσσεύσ"},
-	// German ß doesn't really have an upper case variant, and we
+	// German ß doesn't really have an upper case wrapperType, and we
```
imsodin, line 28: Global search and replace artifact.

lib/fs/folding_test.go
```diff
@@ -38,7 +38,7 @@ var caseCases = [][2]string{
 	{"العَرَبِيَّة", "العَرَبِيَّة"},
 	// Neither does Hebrew.
 	{"עברית", "עברית"},
-	// Nor Chinese, in any variant.
+	// Nor Chinese, in any wrapperType.
```
imsodin, line 41: Also search and replace gone rogue.

lib/fs/mtimefs.go
```diff
@@ -211,42 +215,44 @@ func (f mtimeFile) Stat() (FileInfo, error) {
...
 			FileInfo: info,
-			mtime:    virtual,
+			mtime:    mtimeMapping.Virtual,
 		}
 	}
 
 	return info, nil
 }
 
+// Used by copyRange to unwrap the real file to access the
```
imsodin, line 232: Unfinished comment.

lib/fs/mtimefs.go
```diff
@@ -211,42 +215,44 @@ func (f mtimeFile) Stat() (FileInfo, error) {
...
 func (f mtimeFile) unwrap() File {
 	return f.File
 }
 
-// The dbMtime is our database representation
-
-type dbMtime struct {
-	real    time.Time
-	virtual time.Time
+// The MtimeMapping is our database representation
```
imsodin, line 237: Also quite abruptly terminated comment.

imsodin:
> That would not work, as mtimeFs might be wrapped in logFs/walkFs etc, which embeds Filesystem, so I don't think it would ever satisfy this interface.

Yes it does, as long as it's embedded (and not a member and the type implements `Filesystem` independently, which wouldn't really be a wrapper anymore): The type assertion checks all available methods, not just what's available on the interface.

imsodin (approved):

calmh (approved):

Issue #7636, opened 2021-05-06
Investigate QUIC performance


bt90:
I've been tempted to test QUIC performance after a few reports of slow connections on the forum.

Test setup:
* target: small server running Ubuntu 20.04
  * Intel(R) Celeron(R) G3930 @ 2.90GHz, 8GB RAM
  * connected via LAN to my router
  * Syncthing DB on a SSD
* source: desktop running Windows 10
  * Intel(R) Core(R) i7-6700K@ 4.00GHz, 32GB RAM
  * connected via AC Wifi which caps at around ~600MBit/s according to iperf
  * no thirdparty firewall or AV solution
* sample data: a few big video files

Results:
* TCP: ~560MBit/s which is roughly the cap of my Wifi connection
* QUIC: ~**8.5MBit/s**

Observations:
* low CPU usage
* no packet receive or receive buffer errors according to `netstat -naus`

Packet lengths:
![grafik](https://user-images.githubusercontent.com/1926175/117353678-b50df500-aeb0-11eb-99fc-3ca90a4cc68b.png)

bt90:
Flow graph:

Server: 192.168.178.2
Desktop: 192.68.178.26

![grafik](https://user-images.githubusercontent.com/1926175/117354602-cdcada80-aeb1-11eb-8d12-b4a813691d1b.png)

AudriusButkevicius:
Are you planning to look into it, tweak it/test it?

We use it out of the box and expect it to work, if it doesn't work as expected, I don't think it's worth having an issue here, as we are not going to fix it, and  it should probably be brought up with the quic people as I don't think we have any expertise here.

If you were to do that, I guess it would be best if you wrote a synthetic benchmark using their tools, just to role out syncthing.
If it does seem that it's something that syncthing does, then we could look into it.

bt90:
> Are you planning to look into it, tweak it/test it?

Sadly not a go developer but i'm happy to test.

> We use it out of the box and expect it to work, if it doesn't work as expected, I don't think it's worth having an issue here, as we are not going to fix it, and it should probably be brought up with the quic people as I don't think we have any expertise here.

quic-go is used by caddy and IPFS so I doubt it's supposed to be *that* slow. Maybe @marten-seemann can help here.

AudriusButkevicius:
You could try compiling syncthing with the following diff applied and see if it makes it any better:
```diff
diff --git a/lib/connections/quic_listen.go b/lib/connections/quic_listen.go
[...]

PR #7648, opened 2021-05-09
lib/connections: Add connection benchmarks, and allow binding to port zero addresses

AudriusButkevicius:
This still has the stun glue, so quic will have sad speeds.

### DIFF
--- a/lib/connections/quic_listen.go
+++ b/lib/connections/quic_listen.go
@@ -13,7 +13,6 @@ import (
 	"crypto/tls"
 	"net"
 	"net/url"
-	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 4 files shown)


AudriusButkevicius:
This now also adds support for the OOB stuff that quic starts using in later versions.

Also, picks up a new version of pfilter where I've fixed the buffer corruption bug that was causing abysmal rates.

Fixes #7636

calmh:
This "Do A, B, C" PR format where C seems like an important bugfix that's not to me obvious where it hides in the diff is not great. Also the main thing seems to be adding the methods talked about and reverted in the previous commits but that's not mentioned. Split it up?

AudriusButkevicius:
Split it into two, this one adds benchmarks (and support for `:0` binds)
The other one updates pfilter and adds oob support.

calmh (approved):

lib/connections/quic_listen.go
```diff
@@ -99,19 +98,30 @@ func (t *quicListener) serve(ctx context.Context) error {
 	go svc.Serve(ctx)
 
 	registry.Register(t.uri.Scheme, wrapped)
-	defer registry.Unregister(t.uri.Scheme, wrapped)
```
imsodin, line 102: Unregister doesn't happen anymore in case of an error in `quic.Listen`, as the defer is moved below that.j

PR #7653, opened 2021-05-09
gui: Move the QR code button next to device ID in editDevice modal.

acolomb:
### Purpose

Trying to avoid one source of confusion when using QR codes to mutually add devices, where the local ID / QR code is needed but the user ended up with the remote device's ([forum link](https://forum.syncthing.net/t/great-technology-bad-user-interface/16781/4?u=acolomb)).

The "Show QR" button shows a different representation of the device ID input field contents, so move it right next to that one.  There are now two possible incarnations because in the `editingExisting` case, the read-only device ID is not actually an HTML input.  Treat both cases as Bootstrap input groups with attached button.

Note this removes access to the "Show QR" function when any other tab than "General" is active.  That is intentional to better guide the user about which device ID will be shown (not the local one).

### Testing

Tested in Firefox: Button is shown and works both when adding and editing a device.  Does not show up when editing device default config.  Resizing the window results in sane wrapping of long device IDs, as seen in the second screenshot.

### Screenshots

Screenshot_2021-05-09 devtest Syncthing_addDevice2
Screenshot_2021-05-09 devtest Syncthing_editDevice2

### Documentation

The button is not mentioned in the docs, but can be seen in [some screenshots](https://docs.syncthing.net/intro/getting-started.html).  Is there a preferred way / setup to update those?

### DIFF
--- a/gui/default/assets/css/overrides.css
+++ b/gui/default/assets/css/overrides.css
@@ -443,3 +443,9 @@ ul.three-columns li, ul.two-columns li {
     background-color: #eeeeee;
     opacity: 1;
 }
+
+/* Make a "well" look more like a readonly text input when grouped with a button */
+.input-group .well-sm {
+  padding-top: 6px;
+  padding-bottom: 6px;
+}
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -11,7 +11,14 @@
           <div ng-if="!editingDefaults" class="form-group" ng-class="{'has-error': deviceEditor.deviceID.$invalid && deviceEditor.deviceID.$dirty}" ng-init="loadFormIntoScope(deviceEditor)">
             <label translate for="deviceID">Device ID</label>
             <div ng-if="!editingExisting">
-              <input name="deviceID" id="deviceID" class="form-control text-monospace" type="text" ng-model="currentDevice.deviceID" required="" valid-deviceid list="discovery-list" aria-required="true" />
+              <div class="input-group">
+                <input name="deviceID" id="deviceID" class="form-control text-monospace" type="text" ng-model="currentDevice.deviceID" required="" valid-deviceid list="discovery-list" aria-required="true" />
+                <div class="input-group-btn">
+                  <button type="button" class="btn btn-default btn-sm" data-toggle="modal" data-target="#idqr" ng-disabled="!deviceEditor.deviceID.$valid">
+                    <span class="fas fa-qrcode"></span>&nbsp;<span translate>Show QR</span>
+                  </button>
+                </div>
+              </div>
               <datalist id="discovery-list">
                 <option ng-repeat="id in discovery" value="{{id}}" />
               </datalist>
@@ -29,7 +36,14 @@
                 <span translate ng-if="deviceEditor.deviceID.$error.unique && deviceEditor.deviceID.$dirty">A device with that ID is already added.</span>
               </p>
             </div>
-            <div ng-if="editingExisting" class="well well-sm text-monospace" select-on-click>{{currentDevice.deviceID}}</div>
+            <div ng-if="editingExisting" class="input-group">
+              <div class="well well-sm text-monospace form-control" style="height: auto;" select-on-click>{{currentDevice.deviceID}}</div>
+              <div class="input-group-btn">
+                <button type="button" class="btn btn-default btn-sm" data-toggle="modal" data-target="#idqr">
+                  <span class="fas fa-qrcode"></span>&nbsp;<span translate>Show QR</span>
+                </button>
+              </div>
+            </div>
           </div>
           <div class="form-group">
             <label translate for="name">Device Name</label>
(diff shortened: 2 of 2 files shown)


tomasz1986:
Not exactly related to the change, but I wanted to test the commit and compiled Syncthing from your branch, but then I got an error when trying to run the compiled binary:

```
Invalid version string "c3fef4a7";
        does not match regexp ^v\d+\.\d+\.\d+(-[a-z0-9]+)*(\.\d+)*(\+\d+-g[0-9a-f]+)?(-[^\s]+)?$
```

I wonder what this is about…

imsodin (dismissed):
I like the new location.

Minor (optional) suggestion: I'd probably merge the `editingExisting` cases and just disable the input field if `editingExisting == true` to avoid some duplication.

tomasz1986:
To be frank, this doesn't look the best in terms of graphic design… The functionality is definitely there though. Would it be possible to increase the button's height a little bit to match the Device ID field (when not wrapped, that is).

imsodin:
This is the only place this well thing is used except for the QR modal, there's no related custom css and according to bootstrap it's just a bit of styling ("inset effect") - I'd just remove it and make this a plain input-group.

acolomb:
> This is the only place this well thing is used except for the QR modal, there's no related custom css and according to bootstrap it's just a bit of styling ("inset effect") - I'd just remove it and make this a plain input-group.

I can get the readonly `<input>` element to work in principle.  But the `select-on-click` directive needs to be changed (or an alternative added) for that.  Text wrapping definitely doesn't work this way, which was neat with the "well" solution.  So switching to an `<input>` in the QR modal as well would lose some functionality, but results in only one `select-on-click` implementation.

So what's important here?  `select-on-click` behavior?  Text wrapping?  Cutting duplication between well and input?  Seems we can't have all at once without seriously complicating the code :-(

imsodin:
Totally lost track of this - annoying bootstrap issues of little consequence don't get much attention (bad excuse, I know). Anyway:

The wrapping should still work if you keep the `height: auto` on the text input, right? The button probably won't scale too, and likely it isn't as easy as putting some more `height: auto`s on that and the parent, but at least it looks fine for large displays and ok-ish for small ones.

I don't understand how removing the well would be problematic with regard to `select-on-click`?

acolomb:
> The wrapping should still work if you keep the `height: auto` on the text input, right?

No, `<input>` is always a single-line entry.  See https://stackoverflow.com/a/5286671/9294540 for example.  Multi-line is only possible with `<textarea>`, which opens up another can of worms because we don't want multi-line text to be *entered*.

> I don't understand how removing the well would be problematic with regard to `select-on-click`?

The `select-on-click` behavior needs a different implementation when used on an `<input>` tag.  If we change that, we need to use `<input>` inside the QR code modal also, where it's definitely always read-only.  That results in not wrapping the textual ID on narrow screens anymore.  Or we keep the "well" there, but remove its select-on-click behavior.  Or we add another attribute `select-on-click-input` to have both methods available.

I was going with the least invasive change so far, now not sure which parts are important to you considering it will add much code to have them all.

imsodin:
Thanks for the info. That (no wrapping on input) is a very weird limitation, but hey it's web-dev - what do I expect :)  
I am confused now because it is already an input field for `!editingExisting`. And I don't understand what the `select-on-click` directive even does at all. Seems entirely superfluous, both the value in the qr-modal and the input field is populated with the usual angularjs mechanisms.  
Given the understanding I have now, I'd remove `select-on-click` and make everything an input, living with not displaying the full ID on narrow screens. However I am likely not understanding something here, so staying as is is fine for me too. If using well, the height discrepancy between the button and well is the padding, so a `padding: 6px 12px` (as in form-control) would fix that.

acolomb:
> I am confused now because it is already an input field for `!editingExisting`. And I don't understand what the `select-on-click` directive even does at all. Seems entirely superfluous, both the value in the qr-modal and the input field is populated with the usual angularjs mechanisms.

The `<input>` is editable and therefore used in appropriate contexts.  For read-only contexts, the assumption seems to be that copying the device ID is the most commonly needed operation, therefore `select-on-click` works like in your browser address bar: No need to drag the mouse, but the whole text content is already selected for copying when you click anywhere in the field.  That's a useful UX detail IMHO.

> Given the understanding I have now, I'd remove `select-on-click` and make everything an input, living with not displaying the full ID on narrow screens. However I am likely not understanding something here, so staying as is is fine for me too.

Wrapping on small screens is also a useful behavior.  Such as when you want to tell or verify the device ID in a phone call, not having to scroll around.  So I tend to keep both behaviors, as the PR now does.

> If using well, the height discrepancy between the button and well is the padding, so a `padding: 6px 12px` (as in form-control) would fix that.

I don't quite see where you mean this needs to be added.  Will experiment a little if it can be matched.

acolomb:
Yea, web development is not fun.  Changing the padding on the `<button>` would not help, as it still doesn't grow vertically to match the well height.  Even with `box-sizing: border-box; height: 100%; min-height: 100%;` it will not grow however I tried.  Disabled `padding`, `line-height` as well.  Absolute pixel height is applied correctly, but no way to fill the parent container height (which is perfectly fine at 44 px and grows with the well).  Dunno, it [should work](https://stackoverflow.com/questions/485827/css-100-height-with-padding-margin).  Tested with Firefox only so far.

I propose to accept this slight misalignment and continue with life.  Bootstrap at its best :shrug:

imsodin:
Thanks for the explanations, agreed on that. Regarding my css tinkering:

> I don't quite see where you mean this needs to be added. Will experiment a little if it can be matched.

I tried it directly in the browser dev tools, but this bit in a stylesheet should do the what I mean:

```css
.input-group .well {
  padding-top: 6px;
  padding-bottom: 6px;
}
```

With that they at least align when there's no wrapping.

acolomb:
Applied your style fix in a slight variation, affecting only the `.well-sm` class.

The last commit fixes an annoyance with that (not well-formed) HTML file that always confused my editor.  It does not affect generated translation strings.

imsodin (approved):

PR #7654, opened 2021-05-10
lib/connections: Update pfilter to pick up bugfix/oob stuff, support OOB connections (fixes #7636)

### LINKED ISSUE #7636: Investigate QUIC performance

AudriusButkevicius:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/syncthing/syncthing
 
 require (
-	github.com/AudriusButkevicius/pfilter v0.0.0-20210218141631-7468b85d810a
+	github.com/AudriusButkevicius/pfilter v0.0.0-20210509224425-1301e3144acb
 	github.com/AudriusButkevicius/recli v0.0.5
 	github.com/alecthomas/kong v0.2.16
 	github.com/bkaradzic/go-lz4 v0.0.0-20160924222819-7224d8d8f27e
@@ -28,7 +28,7 @@ require (
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
 	github.com/klauspost/cpuid/v2 v2.0.6 // indirect
 	github.com/lib/pq v1.10.1
-	github.com/lucas-clemente/quic-go v0.19.3
+	github.com/lucas-clemente/quic-go v0.20.1
 	github.com/maruel/panicparse v1.6.1
 	github.com/mattn/go-isatty v0.0.12
 	github.com/maxbrunsfeld/counterfeiter/v6 v6.3.0
@@ -57,3 +57,5 @@ require (
 )
 
 go 1.14
+
+replace github.com/lucas-clemente/quic-go v0.20.1 => github.com/lucas-clemente/quic-go v0.19.3
--- a/lib/connections/quic_listen.go
+++ b/lib/connections/quic_listen.go
@@ -131,7 +131,7 @@ func (t *quicListener) serve(ctx context.Context) error {
 	defer l.Infof("QUIC listener (%v) shutting down", udpConn.LocalAddr())
 
 	t.mut.Lock()
-	t.laddr = packetConn.LocalAddr()
+	t.laddr = udpConn.LocalAddr()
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -57,3 +57,5 @@ require (
 )
 
 go 1.14
+
+replace github.com/lucas-clemente/quic-go v0.20.1 => github.com/lucas-clemente/quic-go v0.19.3
```
calmh, line 61: wat

AudriusButkevicius: pfilter uses quic in tests/benchmarks. The version I use is newer than the one we use here, so pulling in pfilter updates quic-go. I'm forcing it back to the original version we had.

lib/connections/quic_listen.go
```diff
@@ -82,18 +82,24 @@ func (t *quicListener) OnExternalAddressChanged(address *stun.Host, via string)
...
 
-	svc, conn := stun.New(t.cfg, t, packetConn)
+	udpConn, err := net.ListenUDP(network, udpAddr)
+	if err != nil {
+		l.Infoln("Listen (BEP/quic):", err)
+		return err
+	}
+	defer func() { _ = udpConn.Close() }()
+
+	svc, conn := stun.New(t.cfg, t, udpConn)
```
calmh, line 98: I'm really confused about how this thing works. We use pfilter to separate out stun packages from other data. `stun.New` returns the "other data" connection which we call `conn` here and then never use again, except for the deferred close. Instead the `stunQUICWrapper` wraps the underlying UDP conn directly. This should result in two readers of that underlying conn: STUN via pfilter, and QUIC itself. It seems random which packets would end up where...

go.mod
calmh: Does pfilter require the new features in QUIC .20? It seems like it should depend on the older version otherwise, for specifically this reason...

AudriusButkevicius: It depends on it purely for the purpose of tests, but go.mod does not differentiate between test and non-test dependencies.

lib/connections/quic_listen.go
AudriusButkevicius: stunQUICWrapper wraps both, udpConn and conn, though there is a problem here that I've noticed here.
We should be using ReadMsgUDP/WriteMsgUDP of the net.PacketConn returned by the stun.New, but we instead circumvent that.

go.mod
calmh: Well this "we depend on quic-go 0.20.1" and later "haha just kidding when I said 0.20.1 I mean 0.19.3" is really weird... For better or worse Go does the MVS thing and so ideally dependencies that don't care about the specifics should depend on some lower version and let the main module push it upwards, instead of the other way around...

lib/connections/quic_listen.go
AudriusButkevicius: I think I fixed it?!

AudriusButkevicius: I'll step through stuff in the debugger later to day to verify.

AudriusButkevicius:
Verified in the debugger, after updating to a newer version of quic, it does the right thing with the oob stuff.

This PR now contains #7657, but this is hopefully to avoid merge conflicts once that one lands.

PR #7659, opened 2021-05-10
cmd/uraggregate: Handle malformed data, dont abort on error (fixes #7639)

### LINKED ISSUE #7639: Some usage reporting stats missing since June 2020
All performance info (data transferred, memory used, ...) and users joined/left are missing on data.syncthing.net. That coincides with https://github.com/syncthing/syncthing/pull/6770. I only did a quick check in the JS console that it isn't a graph creation issue, .e. that the data is actually missing in the data obtained from `ursrv`. I haven't checked whether the data gets lost while aggregating incoming reports or when querying the db from the UI.

AudriusButkevicius:


calmh (approved):

calmh:
I tweaked the "publish ursrv" job to also handle uraggregate and made the corresponding changes on the server so this gets live.

PR #7660, opened 2021-05-11
gui: Make device name clickable in navbar to show the ID.

acolomb:
### Purpose

Add a QR code icon next to the device name and make all that a link to show the own device's ID with QR code.  A tooltip indicates what the link does.  Inspired by [this forum thread](https://forum.syncthing.net/t/great-technology-bad-user-interface/16781/6).

This duplicates the Actions > Show ID entry but easier to discover.  That button is still kept to satisfy existing user habits.

### Screenshots

image

### Documentation

Currently no need to adjust docs.  Only if the "Show ID" menu entry should be removed as well, the [Getting Started guide](https://docs.syncthing.net/intro/getting-started.html) would need an update.

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -74,11 +74,11 @@
           <img class="logo hidden visible-xs" src="assets/img/favicon-default.png" height="32" alt=""/>
         </span>
         <p class="navbar-text hidden-xs" ng-class="{'hidden-sm':upgradeInfo && upgradeInfo.newer}">
-	  <a href="" data-toggle="modal" data-target="#idqr" ng-click="currentDevice=thisDevice()" tooltip data-original-title="{{'Show ID' | translate}}" data-placement="bottom">
-	    {{thisDeviceName()}}
-	    &nbsp;<span class="fas fa-fw fa-qrcode"></span>
-	  </a>
-	</p>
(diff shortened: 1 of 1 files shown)


imsodin:
In my opinion the link css in the title looks a bit ugly. However without it discoverability of the QR is likely small. A possible alternative would be to add an additional line to the local and remote device tables with the device ID (possibly short ID, with long on hover and long + QR on click).

acolomb:
Well the screenshot shows the "hover" state, including tooltip.  Regularly the device name looks just as always, not underlined.

imsodin:
It's a normally styled `<a>` element, e.g. looking like this in the bright theme:

image

acolomb:
Ah sorry I only looked at the dark theme, which I use to distinguish my production instance from the test build... Thanks for the `.navbar-link` hint, it looks better now with the default theme.

Regarding the links in each device section, I like that idea.  For remote devices it would be an alternative or maybe addition to #7653, so I'd leave this PR as is.  Or do you guys think we will get too many ways to discover the QR code then?

imsodin:
> Regarding the links in each device section, I like that idea. For remote devices it would be an alternative or maybe addition to #7653, so I'd leave this PR as is. Or do you guys think we will get too many ways to discover the QR code then?

I personally would prefer a row for device ID/QR on local/remote devices over this PR. So if you are interested in making that change, I would welcome merging that instead, dropping the link in the title.

acolomb:
Replaced by #7728, which is hopefully just as easy to discover.

PR #7667, opened 2021-05-11
cmd/uraggregate: Cap memory size at a petabyte (fixes #7661)

### LINKED ISSUE #7661: uraggregate: Wonky charts
Seems we should add some filtering, as someone is doing some manual reporting with clearly the wrong units.

AudriusButkevicius:
After this is merged and rolled out we should do:

`DELETE FROM performance WHERE day > '2020-08-11'` and let it re-aggregate;

Fixes #7661

### DIFF
--- a/cmd/uraggregate/main.go
+++ b/cmd/uraggregate/main.go
@@ -287,6 +287,8 @@ func aggregatePerformance(db *sql.DB, since time.Time) (int64, error) {
 			DATE_TRUNC('day', Received) > $1
 			AND DATE_TRUNC('day', Received) < DATE_TRUNC('day', NOW())
 			AND Report->>'version' like 'v_.%'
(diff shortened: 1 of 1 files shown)


calmh (approved):

AudriusButkevicius:
I assume this is not deployed?

Should we setup CI for these things to roll off of master or something?

imsodin:
It was deployed, but looks like the sql snippet wasn't run - done now and data looks sensible again. Thanks for the quick fix.

Personally I'd rather just ssh+build+copy+kill once or twice a year instead of setting up CI/CD.

AudriusButkevicius:
I also noticed we use a ton of `date_trunc('day', bah)` which seems to dodge using an index, so all of these queries are just repeated sequence scans across the whole table.

PR #7671, opened 2021-05-13
lib/model: Don't use empty folder cfg for index sender (fixes #7649)

### LINKED ISSUE #7649: Incorrect local and global states after ignoring and unignoring files
I'm sorry for the so-so title, but this is a little bit complicated. I have experienced the issue in my real network, and now I have managed to reproduce it.

1. Create a new Send Only folder on Device 1 with a few files in it, and share it with Device 2.
2. Accept the folder on Device 2 as Send & Receive, but set the ignore patterns to `*`.

   image
   image

5. Add a new file to the folder on Device 2, and unignore it with `!`.

   image

6. Change the folder type on Device 1 to Send & Receive.

   image

7. Remove the `!` exclusion from the ignore patterns on Device 2, so that everything is ignored again with `*`.

   image

   As you can see in the screenshot, the global state on Device 2 has been broken.

8. Clear the ignore patterns on Device 2, so that there are no ignore patterns on both devices now.
9. The final state looks like this, i.e both folders seem to think that they are up-to-date, but the are not.

   image
   image

**Edit**: Resetting deltas on Device 2 seems to get the folder states back to normal.

imsodin:
~~Adds more debugging to analyse https://github.com/syncthing/syncthing/issues/7649~~

Still contains the debugging, but mainly now a fix for #7649.

The diff has too little content to see what the problem is, so here is a larger section:

```diff
 		cfg, ok := m.cfg.Folder(folder.ID)
 		if ok {
 			folderDevice, ok = cfg.Device(deviceID)
 		}
 		if !ok {
 			indexSenders.remove(folder.ID)
 			if deviceCfg.IgnoredFolder(folder.ID) {
 				l.Infof("Ignoring folder %s from device %s since we are configured to", folder.Description(), deviceID)
 				continue
 			}
 			delete(expiredPending, folder.ID)
 			of.Label = folder.Label
 			of.ReceiveEncrypted = len(ccDeviceInfos[folder.ID].local.EncryptionPasswordToken) > 0
 			if err := m.db.AddOrUpdatePendingFolder(folder.ID, of, deviceID); err != nil {
 				l.Warnf("Failed to persist pending folder entry to database: %v", err)
 			}
-			indexSenders.addPending(cfg, ccDeviceInfos[folder.ID])
+			indexSenders.addPending(folder.ID, ccDeviceInfos[folder.ID])
```

The `cfg` at this point is just an empty/default object. Thus in `addPending` when using `cfg.ID`, that's an empty string.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -587,6 +587,18 @@ func (db *Lowlevel) dropFolderIndexIDs(folder []byte) error {
 	return t.Commit()
 }
 
+func (db *Lowlevel) dropIndexIDs() error {
+	t, err := db.newReadWriteTransaction()
+	if err != nil {
+		return err
+	}
+	defer t.close()
+	if err := t.deleteKeyPrefix([]byte{KeyTypeIndexID}); err != nil {
+		return err
+	}
+	return t.Commit()
+}
+
 func (db *Lowlevel) dropMtimes(folder []byte) error {
 	key, err := db.keyer.GenerateMtimesKey(nil, folder)
 	if err != nil {
--- a/lib/db/schemaupdater.go
+++ b/lib/db/schemaupdater.go
@@ -20,7 +20,7 @@ import (
 // do not put restrictions on downgrades (e.g. for repairs after a bugfix).
 const (
 	dbVersion             = 14
-	dbMigrationVersion    = 17
+	dbMigrationVersion    = 18
 	dbMinSyncthingVersion = "v1.9.0"
 )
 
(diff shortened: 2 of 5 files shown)


calmh (dismissed):
Nice

imsodin:
Actually there's enough support queries that might be caused by this bug to warrant a one-time delta index drop with this bugfix -> added a migration for that.

AudriusButkevicius (approved):
Finally, perhaps we can now die in peace.

PR #7672, opened 2021-05-13
lib/connections: Reduce default quic redial interval (fixes #7471)

### LINKED ISSUE #7471: Improve UDP hole punching
As discussed in https://forum.syncthing.net/t/a-few-questions-about-quic-hole-punching-and-stun we should turn up dialing frequency for QUIC endpoints to ensure that stateful firewalls keep our punched holes open.

I suggest that we aim for maxium delay between outgoing UDP packets of about 20s. This should be low enough for most firewalls to keep the port constantly open. e.g Linux netfilter has a default timeout of 30s. But i fear that consumer routers might even have lower limits so 20s should be a good compromise.

AudriusButkevicius:


calmh (approved):
(I recall there are some limits applied by the outer connection loop as well, didn't check if those interfere with the intention here.)

AudriusButkevicius:
So I can see that it defaults to 1min sleep if something fails, but I could not find anything that would clamp the sleep time to be less than something.

AudriusButkevicius:
Ah, there is a sub 5s clamp.

Regardless,, this turns into 20s into 32 seconds because we capture now at the start of dialing, so probably needs more thought

AudriusButkevicius:
Actually, it's just the first loop, so this is fine.

bt90:
@AudriusButkevicius https://www.bountysource.com/issues/98271887-improve-udp-hole-punching 😉

Issue #5385, opened 2018-12-17
Deleted files show up as failed scans


calmh:
Since we added reporting of scan failures to the GUI, I get things like this:

<img width="1222" alt="screenshot 2018-12-17 at 13 42 02" src="https://user-images.githubusercontent.com/125426/50087832-a0f2cb80-0201-11e9-8e74-51cfba414889.png">

This popped up when I deleted the test file and it got reported by fs notifications. "no such file" is expected under the circumstances, so I think we're being too eager to report an error somewhere in the stack.

Also, the scan failure doesn't go away until you do a manual rescan *and reload the GUI*.

PR #7675, opened 2021-05-14
lib/scanner: Do not drop all not-exist-errors and debug logging

imsodin:
Another debugging PR for https://forum.syncthing.net/t/german-umlaute-like-aou-seems-to-make-trouble/16804/12


imsodin:
This fixes an actual usability bug: We drop all `fs.IsNotExist` errors. We do that because we might be asked to scan something that doesn't exist, which isn't actually an error (https://github.com/syncthing/syncthing/issues/5385). However having the check there at the end means it also drops legitimate errors, like in this case normalization errors. Thus I moved the `fs.IsNotExist` check to the error check in `walkAndHashFiles`.

calmh (approved):

PR #7676, opened 2021-05-14
Add static favicon

### LINKED ISSUE #7638: favicon not working Firefox & derivative browsers
Firefox, Pale Moon, and similar browsers do not display the initial favicon (or at all) since the favicon link is dynamically created by evaluating javascript expressions. 

The bug is due to some FireFox forks not re-loading the scripted favicon link after the page first loading. Chrome and forks do, but this is non-standard behavior. 

See https://forum.syncthing.net/t/fixing-syncthing-favicons-in-firefox/15726 for description of cause and work around. 

I suggest that the initial favicon link not result in a 404, as noted by calmh at https://forum.syncthing.net/t/fixing-syncthing-favicons-in-firefox/15726/2. 

If it is desirable to have the favicon change based on status, the favicon link can be updated as at present. However, changing the favicon after page load does not appear to be standard behavior, so many browsers may not display the change.

bt90:
### Purpose

This PR adds a static favicon with the path `/favicon.ico` and exempts it from the authentication logic. This allows Bitwarden and probably other tools to pick up the favicon even if the web UI requires authentication.

~~TODO: modify index.html to first load the static favicon in order to solve #7638~~

### Testing

Successfully loaded the favicon without logging in.


bt90:
Had a quick look at the HTML/JS code and decided not to touch it.

AudriusButkevicius:
I feel that this could be made prettier. The fact that some list of files that we permit lives in the basic auth middleware sounds sad.

Perhaps something like:
```
authMiddleware(handler)
noAuthMiddleware(handler, authMiddleware, []string{"/noauth1", "/noauth2"})
```

bt90:
>  some list of files

Do we have any other files which should bypass authentication?

> I feel that this could be made prettier. 

While i agree that a separate method would be prettier, that's about the limit of what I dare contribute without go experience 😅

bt90:
> That's a pretty huge favicon. Any reason not to use the <5KiB one in cmd/ursrv/static/assets/img/favicon.png?

Just reusing our current ico:

https://github.com/syncthing/syncthing/blob/main/assets/logo.ico

The ico format also bundles multiple resolutions.

greatroar:
Not a big deal (it compresses to 50KiB), it's just that all the other favicons are way smaller.

Regarding the Go code, the favicon skips the auth but does still send the device ID in the X-Syncthing-Id header.

bt90:
> Regarding the Go code, the favicon skips the auth but does still send the device ID in the X-Syncthing-Id header.

Good catch. Probably better to do it like @AudriusButkevicius suggested and handle this in a separate function.

greatroar:
Inserting this between the IsAuthEnabled and UseTLS checks seems to work:

```go
faviconMux := http.NewServeMux()
faviconMux.HandleFunc("/favicon.ico", func(w http.ResponseWriter, r *http.Request) {
	io.WriteString(w, "insert icon here")
})
faviconMux.Handle("/", handler)
handler = faviconMux
```

Maybe the statics handler can be put there?

st-review:
:robot: *beep boop*

PR #7678, opened 2021-05-15
lib/fs: Handle normalization on macOS (fixes #7677)

### LINKED ISSUE #7677: UTF-8 normalization doesn't work on macOS
When Syncthing sees a filename which is well formed UTF-8 but unexpected normalization form for the system it corrects the normalization form on disk. This currently doesn't happen on macOS, instead the file remains unscanned and "invisible" to Syncthing.

calmh:
This one was somewhat tricky; macOS does normalization-insensitive lookups so Lstat will find a file regardless of the normalization on disk vs in lookup, but out caseFS is more sensitive.

Initially I thought we should simply normalize in caseFS on macOS, but that means we return
normalization issues as case conflicts. That then means we should allow renames over case conflicts but that's risky as macOS can also have case sensitive filesystems.

Instead I changed checkCase to handle "not found" from the case insensitive lookup in the same manner as "not found" from Lstat. This works and I don't see any obvious negative side effects...

### DIFF
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -361,7 +361,6 @@ func (f *caseFilesystem) checkCase(name string) error {
 		}
 		return err
 	}
-
 	return f.checkCaseExisting(name)
 }
 
@@ -412,8 +411,7 @@ func (r *defaultRealCaser) realCase(name string) (string, error) {
 		return realName, nil
 	}
 
-	comps := PathComponents(name)
-	for _, comp := range comps {
+	for _, comp := range PathComponents(name) {
 		node := r.cache.getExpireAdd(realName)
 
 		node.once.Do(func() {
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -368,11 +368,6 @@ func (f *caseFilesystem) checkCase(name string) error {
 // stating the file.
 func (f *caseFilesystem) checkCaseExisting(name string) error {
 	realName, err := f.realCase(name)
-	if IsNotExist(err) {
-		// It did exist just before -> cache is outdated, try again
-		f.dropCache()
-		realName, err = f.realCase(name)
-	}
 	if err != nil {
 		return err
 	}
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	lru "github.com/hashicorp/golang-lru"
+	"golang.org/x/text/unicode/norm"
 )
 
 const (
@@ -376,7 +377,10 @@ func (f *caseFilesystem) checkCaseExisting(name string) error {
 	if err != nil {
 		return err
 	}
-	if realName != name {
+	// We normalize the normalization (hah!) of the strings before
+	// comparing, as we don't want to treat a normalization difference as a
+	// case conflict.
+	if norm.NFC.String(realName) != norm.NFC.String(name) {
 		return &ErrCaseConflict{name, realName}
 	}
 	return nil
@@ -409,7 +413,7 @@ func (r *defaultRealCaser) realCase(name string) (string, error) {
 	}
 
 	comps := PathComponents(name)
-	for i, comp := range comps {
+	for _, comp := range comps {
 		node := r.cache.getExpireAdd(realName)
 
 		node.once.Do(func() {
@@ -426,7 +430,7 @@ func (r *defaultRealCaser) realCase(name string) (string, error) {
 			lastLower := ""
 			for _, n := range dirNames {
 				node.children[n] = struct{}{}
-				lower := UnicodeLowercase(n)
+				lower := UnicodeLowercaseNormalized(n)
 				if lower != lastLower {
 					node.lowerToReal[lower] = n
 					lastLower = n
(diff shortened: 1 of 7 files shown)


AudriusButkevicius (approved):

imsodin (changes requested):
I am a bit concerned about the catch-all workaround. It is correct for the purposes of this PR, however it's not guaranteed that such an error only occurs under exactly these circumstances. Returning this error here (status quo) seems like a good property in general. The fix for #7677 would then be to check the error from `checkCase(newpath)` for `IsNotExist` in `caseFilesystem.Rename` and continue if true - like we already do for detecting a case-conflict for case-only renames.

imsodin:
Or maybe just drop `ErrNotExist` unconditionally like we already do in the `LStat` just above. If nothing exists with the given name, that's not a case-conflict, so `err == nil` seems right.

calmh:
I think the prior assumption is that isnotexist can't happen from the case resolving code if the Lstat succeeded. macOS breaks that assumption in this case, I'm not sure if there are other possibilities (cache issues?). I can think more deeply about it tomorrow.

calmh:
So...

`(*caseFilesystem) checkCase()` doesn't have a comment specifying exactly what it does, but the code implies it should return an error only for case-related problems; hence nil if the thing we're looking at doesn't exist according to `Lstat()`. Then `(*caseFilesystem) checkCaseExisting()` does the same thing on the assumption that the thing we're looking at *does* in fact exist according to Lstat, calling `(*defaultRealCaser) realCase()` to figure out the canonical casing of the thing. It even checks specifically for IsNotExist and tries again, on the assumption that our caches were outdated...

Given all this, `realCase()` returning `ErrNotExist` for a thing that `Lstat()` can see is really surprising. Essentially one of them must be wrong. Frankly I think it's `Lstat()` that's wrong, and in fact our `(*caseFilesystem) Lstat()` instead returns `ErrNotExist` here because of all the above... But that's not what we use internally in `caseFileSystem` of course.

So yeah, I think maybe the proper thing is for `(*caseFilesystem) checkCaseExisting()` to return nil on not found, as even though the situation is strange it is at least not a *case* error as far as checkCaseExisting can determine? (Without guarding with `runtime.GOOS`)

The only other FS I know of that enforces normalization is ZFS with `utf8only=on` and `normalization=<something>` which should in theory behave exactly like macOS on this point -- it stores whatever is given as the name on disk, but lookups and comparisons are normalized.

calmh:
Of course, doing that then means we do *not* return ErrNotExist when we can't find things, so the "caches were invalid, let's try again" case in checkCase() doesn't happen and instead we return the unmodified name as given by the user as assumed correct...

imsodin:
I am clearly worrying about super-nice/unlikely cases, just tell me if I should stop:  
This might now mask a case conflict, if a parent directory has such a normalization problem. And the removed condition where we dropped the cache on `IsNotExist` is still a possible scenario (nothing to do with normalization, just an item got deleted since cache creation).

Initially I thought it would make sense to handle utf-8 normalization problems like case-problems, as both are a case of "filesystem treats two distinct strings as the same path". However that's not a problem for unicode normalization, because Syncthing also does consider the two normalizations as the same and already has measures in place to detect/prevent collisions.

What do you think about checking normalisations:

```diff
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	lru "github.com/hashicorp/golang-lru"
+	"golang.org/x/text/unicode/norm"
 )
 
 const (
@@ -440,7 +441,12 @@ func (r *defaultRealCaser) realCase(name string) (string, error) {
 			lower := UnicodeLowercase(comp)
 			comp, ok = node.lowerToReal[lower]
 			if !ok {
-				comp, ok = node.lowerToReal[NativeFilename(lower)]
+				for _, n := range unicodeNorms {
+					comp, ok = node.lowerToReal[n.String(lower)]
+					if ok {
+						break
+					}
+				}
 				if !ok {
 					return "", ErrNotExist
 				}
@@ -453,6 +459,13 @@ func (r *defaultRealCaser) realCase(name string) (string, error) {
 	return realName, nil
 }
 
+var unicodeNorms []norm.Form = []norm.Form{
+	norm.NFC,
+	norm.NFD,
+}
+
 func (r *defaultRealCaser) dropCache() {
 	r.cache.Purge()
 }
```

calmh:
I tried something similar initially (normalizing the keys of the `lowerToReal` map), which caused other issues as it resulted in the normalization difference being returned as a case conflict... Though perhaps if we did that and checked if the difference is only in normalization before producing a case conflict. Let me try.

calmh:
Though this begs the question of how we handle actual normalization conflicts on filesystems where that can happen...

lib/fs/casefs.go
```diff
@@ -375,7 +376,10 @@ func (f *caseFilesystem) checkCaseExisting(name string) error {
 	if err != nil {
 		return err
 	}
-	if realName != name {
+	// We normalize the normalization (hah!) of the strings before
```
imsodin, line 379: Obligatory xkcd: https://xkcd.com/927/  
(And new standard: standard == normalisation (== normalization)).

imsodin (approved):
> Though this begs the question of how we handle actual normalization conflicts on filesystems where that can happen... That _should_ be considered a case conflict, though I'm not sure how we'd tell the difference, probably it requires `SameFile` and stuff... Or maybe that is already correctly handled by the normalization code and the scanner.

Don't you already know about normalization only (i.e. not case) conflicts with the norm check after `checkCase`. I.e. when the names are equal, all good, otherwise when normalized equal we have a normalization conflict and otherwise there's a case conflict. However I think it should already work fine as is (ignoring normalization errors), as we normalize/native-ize everything in scanner/puller.

calmh:
I don't know any more, luckily we have users who will undoubtedly find the remaining corner cases.

PR #7679, opened 2021-05-15
cmd/uraggregate: Optimise queries

AudriusButkevicius:
So this actually uses the index.

Seems postgres can't work out that it can use an index scan if something is done with the indexed field.
Even if I disable seq scans on the optimiser, it still picks a seq scan.

Beware, I think this is equivalent, but off by one errors are super hard, hence give it a proper review.

### DIFF
--- a/cmd/uraggregate/main.go
+++ b/cmd/uraggregate/main.go
@@ -47,7 +47,7 @@ func main() {
 func runAggregation(db *sql.DB) {
 	since := maxIndexedDay(db, "VersionSummary")
 	log.Println("Aggregating VersionSummary data since", since)
-	rows, err := aggregateVersionSummary(db, since.Add(24 * time.Hour))
+	rows, err := aggregateVersionSummary(db, since.Add(24*time.Hour))
 	if err != nil {
 		log.Println("aggregate:", err)
 	}
@@ -62,7 +62,7 @@ func runAggregation(db *sql.DB) {
 
 	since = maxIndexedDay(db, "Performance")
 	log.Println("Aggregating Performance data since", since)
-	rows, err = aggregatePerformance(db, since.Add(24 * time.Hour))
+	rows, err = aggregatePerformance(db, since.Add(24*time.Hour))
 	if err != nil {
 		log.Println("aggregate:", err)
 	}
@@ -70,7 +70,7 @@ func runAggregation(db *sql.DB) {
 
 	since = maxIndexedDay(db, "BlockStats")
 	log.Println("Aggregating BlockStats data since", since)
-	rows, err = aggregateBlockStats(db, since.Add(24 * time.Hour))
+	rows, err = aggregateBlockStats(db, since.Add(24*time.Hour))
 	if err != nil {
 		log.Println("aggregate:", err)
 	}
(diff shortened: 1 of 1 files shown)


cmd/uraggregate/main.go
```diff
@@ -163,7 +163,7 @@ func setupDB(db *sql.DB) error {
 
 func maxIndexedDay(db *sql.DB, table string) time.Time {
 	var t time.Time
-	row := db.QueryRow("SELECT MAX(Day) FROM " + table)
+	row := db.QueryRow("SELECT MAX(DATE_TRUNC('day', Day)) FROM " + table)
```
AudriusButkevicius, line 166: This change is not actually needed, but just playing it safe.

AudriusButkevicius:
We could also try upgrading to latest and greatest postgres and see if it helps.

cmd/uraggregate/main.go
```diff
@@ -62,15 +62,15 @@ func runAggregation(db *sql.DB) {
...
+	rows, err = aggregatePerformance(db, since.Add(24*time.Hour))
 	if err != nil {
 		log.Println("aggregate:", err)
 	}
 	log.Println("Inserted", rows, "rows")
 
 	since = maxIndexedDay(db, "BlockStats")
 	log.Println("Aggregating BlockStats data since", since)
-	rows, err = aggregateBlockStats(db, since)
+	rows, err = aggregateBlockStats(db, since.Add(24*time.Hour))
```
calmh, line 73: I think the since here should be now().add(24 hours).truncate(24 hours) for the > comparisons in the sql to be correct

calmh: Maybe that’s equivalent to what’s already done since “since” should be a midnight timestamp…

cmd/uraggregate/main.go
```diff
@@ -315,8 +315,8 @@ func aggregateBlockStats(db *sql.DB, since time.Time) (int64, error) {
 		SUM((Report->'blockStats'->>'copyElsewhere')::numeric) AS CopyElsewhere
 		FROM ReportsJson
 		WHERE
-			DATE_TRUNC('day', Received) > $1
-			AND DATE_TRUNC('day', Received) < DATE_TRUNC('day', NOW())
+			Received > $1
+			AND Received < DATE_TRUNC('day', NOW())
```
calmh, line 319: These all seem reasonable

cmd/uraggregate/main.go
AudriusButkevicius: That's the idea, that is why I've added the truncate on getting the since.

calmh (approved):

AudriusButkevicius:
Whats the appetite for trying yo upgrade the postgres? 

I've got most of whats needed ready.

imsodin:
I wouldn't mind having a backup and wanted to propose making an off-site backup (data isn't critical, still nice to have) but didn't quite grasp how much it is. Maybe we should add a retention limit for individual reports and just keep the aggregates the time before?

AudriusButkevicius:
As it stands it's around 150GB.
I don't see the harm in keeping the history forever. I think you could have some nice ad-hoc insights at some point.

PR #7686, opened 2021-05-17
as for international practice, china replace with china mainland

sandacn:
### Purpose

Due to the particularity of China, the general international practice is to use `China Mainland`.
You can refer to Apple's official website: https://www.apple.com/choose-country-region/

### Testing

done

### Screenshots

WX20210517-104929

## Authorship

sandacn
sandacn#gmail.com


calmh:
This file is auto generated based on what Transifex calls things.

PR #7689, opened 2021-05-17
lib/model: Ensure indexes are only received after checking IDs (ref #7649)

imsodin:
### Purpose

This change would have prevented the db/index corruption even if the bug that caused #7649 is present. It ensures that indexes are only ever received after checking IDs, regardless of the order methods are called in model. Getting the order right in the model is hard, as it involves multiple, intertwined inputs (connections being established/closed, local config changes, remote config changes, ...).

The main change here is to move receiving indexes/writing them to the db into what is currently the `indexSender` (renamed to `indexHandler`). Checking indexIDs happens when starting/creating an index handler, thus it's now impossible to receive indexes before checking IDs.

Git doesn't pick up the the rename as there's too many changed lines, however the diff is actually quite sane with one function moved verbatim. Thus I recommend looking at the diff without the first commits:  
https://github.com/syncthing/syncthing/pull/7689/files/adfb666f486503e6f63803de3a2831adaa4fa9a9..0e7a416c77bbd80556b1646be6073e7f6c9c7152
 
### Testing

Due to the stricter conditions on receiving indexes a lot of tests now require a (fake) connection to work. Somehow that resulted in more usage of `fakeFS` and triggered a crash there, thus the lone change to lib/fs.


lib/model/model.go
```diff
@@ -2253,17 +2215,31 @@ func (m *model) AddConnection(conn protocol.Connection, hello protocol.Hello) {
...
+		if fcfg.SharedWith(deviceID) {
+			folderStates[id] = &indexHandlerFolderState{
+				cfg:    fcfg,
+				fset:   m.folderFiles[id],
+				runner: m.folderRunners[id],
+			}
+		}
+	}
+	m.fmut.RUnlock()
+	m.indexHandlers[deviceID] = newIndexHandlerRegistry(conn, m.deviceDownloads[deviceID], closed, folderStates, m.Supervisor, m.evLogger)
```
calmh, line 2242: I'm not super excited about exposing the `indexHandlerFolderState` as a thing outside the `indexHandler`, it feels very much like an internal type. The other usages are easy as it's just used to pass essentially two parameters that could be broken out, but this map thing is something else. Perhaps we could solve this by creating the index handler registry and then in the loop calling something like `registry.RegisterFolderState(cfg, fset, runner)`

lib/model/model.go
```diff
@@ -1214,9 +1192,9 @@ func (m *model) ClusterConfig(deviceID protocol.DeviceID, cm protocol.ClusterCon
 
 	// Assemble the device information from the connected device about
 	// themselves and us for all folders.
-	ccDeviceInfos := make(map[string]*indexSenderStartInfo, len(cm.Folders))
+	ccDeviceInfos := make(map[string]*indexHandlerStartInfo, len(cm.Folders))
```
calmh, line 1195: Similarly here I wonder if we could avoid building and carrying around a bunch of `indexHandlerStartInfo`s and instead just tell the registry about them at the point were we would now build the startinfo?

calmh (commented):
This seems reasonable, though I can't see I understand all the intricacies. One reason for that may be that it's still very intertwined with the model, with both the old startinfo type and the new folderstate type floating around in various places in the model. I think if we do this breaking-out thing it would be much nicer if it were possible to encapsulate that state more in the indexHandler{Registry} so that we register stuff by calling appropriate methods on those instead of essentially building their state from the outside...

lib/model/model.go
imsodin: That type is badly named, I renamed it to `clusterConfigDeviceInfo`: It isn't exclusively used for index handling, it also contains the encryption info, is used for auto-accepting folders, ...  
We can't push the info immediately to the index-handler and delegate hashing out what to do with it directly, as it would e.g. create a brief race condition between sending indexes and checking encryption later in `ClusterConfig` (which might need to prevent the connection). Generally the index-handler would have to check for the same conditions that are already checked in `ccHandleFolders`, so I would expect more duplication than simplification there.  
Or maybe I am just too deep into all of this and need to get some distance to grasp the bigger picture.

imsodin:
The review comments should be addressed now (and model - index-sender a bit less entangled now).

And in the meantime an issue came up that "upgraded" this PR from nice-to-have/hardening, to fixing a known (even though niche) issue: https://forum.syncthing.net/t/issues-when-ommiting-password-when-syncing-to-untrusted-device/16835

calmh (approved):

PR #7690, opened 2021-05-17
lib/api, lib/config: Apply defaults before deserializing json

imsodin:
This applies defaults before unmarshaling for `PUT`/`POST` on `/rest/config/...`. The intention being, that if a caller omits a field, that is set to a proper default, not just the type null/default value. This can e.g. typically happen if a new config field is added, that the caller doesn't know about. For modifying existing configs, one could just use `PATCH` to achieve this. However not for e.g. adding a new folder.

Caveat: This only applies to the api/JSON: Config on disk/xml works as before, i.e. no defaults applied there. However this is not problematic, as we only use that internally and we always "know" about all fields.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -2117,7 +2117,7 @@ angular.module('syncthing.core')
                 folderCfg.versioning.params.command = '' + folderCfg._guiVersioning.externalCommand;
                 break;
             default:
-                delete folderCfg.versioning;
+                folderCfg.versioning.type = '';
             }
             delete folderCfg._guiVersioning;
 
--- a/lib/api/confighandler.go
+++ b/lib/api/confighandler.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/syncthing/syncthing/lib/config"
 	"github.com/syncthing/syncthing/lib/protocol"
+	"github.com/syncthing/syncthing/lib/util"
 )
 
 type configMuxBuilder struct {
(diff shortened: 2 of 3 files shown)


calmh:
Neat, but also scary. I'm sure we have cases where the absence of something is meant to disable it -- versioning being one such thing I think. Does it work to disable versioning from the GUI if the default says something else?

imsodin:
You are right of course. I only had a new idea concerning the deserialisation and decided to ignore that it doesn't apply to XML, however I missed that crucial bit: There's lots of config values where the zero value/absence does not mean "undefined" but "disabled", and there's no way to distinguish. I guess we'll have to live with that until we do a major (breaking) config change.

For the use-case of adding a new folder with proper defaults, without having to know about all the config fields, there could be e.g. a `POST` endpoint at `/rest/defaults/folder?new` that requires a body containing at least a folder id and path, possible more fields, to create a new folder based on defaults.

calmh:
Well, I mean, for most things we can distinguish. We can send `false`, `""`, `[]`, `0` and whatnot and probably do so. Versioning also has a `{type: "none"}` representation, it's just that we don't use it? I Don't think this is conceptually impossible/dead in the water.

calmh:
XML is worse because it's difficult to override a non-empty slice at all but JSON can do that.

calmh:
The rest is up to us to define, if we say that absent keys means to use the default then that's the thing. There's no inherent reason that absent must mean disabled/zero/etc. 

Though, possibly, we should consider wether this should only apply to the new config API. On the one hand that defeats the purpose of auto-enrolling old wrappers in new features, but on the other hand it avoids breaking people who depend on the current behavior.

imsodin:
My reasoning there was consistency between the serialisation formats we use, and in XML it's a pain if possible to distinguish missing from default, and in protobuf there is no such thing as missing afaik. However that's no good reason to not do this (apply defaults), because it concerns the API, i.e. JSON. And if we ever do use protobuf for the API we should just do a v2 and also change the config to handle "enabledness" explicitly.

Type none is only a gui thing, in go land `Type != ""` is used to check if versioning is enabled. The fix is actually simple: Just don't delete the entire versioning config in the gui, just set the type to the empty string. Hardly very nice, but consistent with what we do now.

> Though, possibly, we should consider wether this should only apply to the new config API. On the one hand that defeats the purpose of auto-enrolling old wrappers in new features, but on the other hand it avoids breaking people who depend on the current behavior.

This almost doesn't change the old config API: When pasting a full config, there's no configured defaults.The only change is that now basic (?) defaults (`(ext.default)` in proto) is also applied to folders and devices (that was already applied to everything else).

gui/default/syncthing/core/syncthingController.js
```diff
@@ -2117,7 +2117,7 @@ angular.module('syncthing.core')
                 folderCfg.versioning.params.command = '' + folderCfg._guiVersioning.externalCommand;
                 break;
             default:
-                delete folderCfg.versioning;
+                folderCfg.versioning.type = '';
```
calmh, line 2120: This means we retain various options like staggered times and whatnot but with empty time? Or is that cleaned up in some other place? (I guess `folderCfg.versioning = {type: ''} // nothing` would resolve this otherwise.)

calmh (approved):

calmh (approved):

calmh:
This definitely requires a change note

PR #7692, opened 2021-05-17
Fix UnicodeLowercaseNormalized on lowercase NFD

greatroar:
#7678 changed UnicodeLowercase to do normalization, but missed the first return statement.


calmh (approved):
Ah, thanks

PR #7694, opened 2021-05-17
cmd/syncthing/cli: Update recli, fix stdin handling (fixes #7685, fixes #7673)

### LINKED ISSUE #7673: bug: cli subcommand is stuck on non-interactive shell
Syncthing version: v1.16.1

I've been trying to use `syncthing cli` over SSH to configure my hosts (ansible-like provisioning software, but it's details). 
E.g. `ssh hostname 'syncthing cli config options local-ann-port set 1234'`.

**Expected**: it applies the configuration 

**Result**: it gets stuck with no output

**Investigation**: 

- sshing onto the host and running command in terminal seemed to work perfectly. I tried `strace` and it seemed that when running over ssh, the command is indefinitely waiting. After trying with `ssh -t` instead (which worked), I figured it had to do with TTY, searched in the code and it seems it's this bit to blame

https://github.com/syncthing/syncthing/blob/ef4b8a2cf86d067c59a31ab3de20ffc0dca6e877/cmd/syncthing/cli/main.go#L103-L106

I figured I could try providing stdin output instead and indeed: `echo 'config options local-ann-port set 1234' | ssh hostname 'syncthing cli'` -- that worked.

**Resolution**

- might be nice to write `waiting for commands input` to stderr or something like that before trying to read so at least it's clear to user what's happening
- might be nice to add something about it to `syncthing cli --help'? Since it's kind of unexpected, especially considering that it all works as expected when you test it in the terminal.
- not sure what was the intent of this stdin reading functionality (didn't find docs), so hard to tell what's the right way to actually solve it..
  I don't know Go apis, but wonder whether it could check that there is no stdin and skip the reading part? Seems wrong that it gets stuck on it

Happy to help with the former two if it's welcome, not sure about the latter since I don't know Go well. Thanks!

### LINKED ISSUE #7685: CLI: strconv.ParseInt error when adding new device via CLI
I tried adding a new device using the CLI and found that the command resulted in a strconv.ParseInt error. Odd, because the internal format is a DeviceID which I think is an array of bytes. Quoting the device-id in the cli results in the same error.

command: `syncthing cli config devices add --device-id=XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX --name="MyName"`

The error I get is: `Incorrect Usage: invalid value "XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX" for flag -device-id: strconv.ParseInt: parsing "XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX": invalid syntax` and the device is not added to Syncthing. 

Note: I've masked the device-id above.

### initial troubleshooting

Assuming I'm not making some obvious mistake in running the command: I suspect that [recli](https://github.com/AudriusButkevicius/recli) is mistakenly trying to convert the input from a string to an Int. But I haven't had time to further dive into it.

### Required information

* Version: 1.16.1
* installed via apt on Debian 4.19.152-1 (device is a intel NUC)

AudriusButkevicius:
Makes stdin handling less magical, and picks up an update to recli which fixes handling TextUnmarshal interface on structs when field is a non-pointer.

### DIFF
--- a/cmd/syncthing/cli/main.go
+++ b/cmd/syncthing/cli/main.go
@@ -8,13 +8,13 @@ package cli
 
 import (
 	"bufio"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"strings"
 
 	"github.com/alecthomas/kong"
 	"github.com/flynn-archive/go-shlex"
-	"github.com/mattn/go-isatty"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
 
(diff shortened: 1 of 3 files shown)


cmd/syncthing/cli/main.go
```diff
@@ -97,38 +97,41 @@ func Run() error {
...
+					args := make([]string, len(os.Args)-1)
+					copy(args, os.Args)
+
+					fmt.Println("Reading commands from stdin...", args)
+					scanner := bufio.NewScanner(os.Stdin)
+					for scanner.Scan() {
+						input, err := shlex.Split(scanner.Text())
+						if err != nil {
+							return errors.Wrap(err, "parsing input")
+						}
```
lodenrogue, line 119: This continue is kind of awkward. Made me pause for a second to figure out what the code was doing. Here's a suggestion:

Instead of the continue try this:

```go
if len(input) != 0 {
    err = app.Run(append(args, input...))
    if err != nil {
        return err
    }
}
```

lodenrogue (changes requested):

cmd/syncthing/cli/main.go
imsodin: The continue is consistent with the style in use in Syncthing's code base: Keep things flat and return/continue/break early. Maybe it's a Go thing, maybe just "us" - in some other code-bases this seems to be frowned upon (only return/... at the end). I like it this way.

imsodin (approved):

PR #7712, opened 2021-05-26
all: Allow dismissing pending devices / folders without ignoring (fixes #7700)

### LINKED ISSUE #7700: Add rest endpoint(s) to dismiss pending folder/device invitations
See https://github.com/syncthing/syncthing/issues/7179#issuecomment-843853381 and following. Before moving pending folders/devices to the db, they could be dismissed by modifying the config. Now there's endpoints to get these (GET `/rest/cluster/pending/...`) but no way to dismiss. E.g. handling `DELETE` on the same endpoints (either with some payload or ID in the url to only dismiss specific entries) would probably do. However it's done, dismissing those should be possible through the api.

acolomb:
### Purpose

Adds a "Not Now" button to the pending device and pending folder panel in the GUI.  Tooltips on this one and the "Ignore" buttons explain what the difference is.  This uses a new API in the backend, with `DELETE` methods on the existing endpoints.  Fixes #7700.

As a collateral, existing error handling for removing pending stuff is slightly changed.  We no longer include stuff in the `Pending*Changed` events where the removal actually failed.

Split up into several commits for easier review.

### Testing

Manual testing in the GUI for both a pending folder and a pending device.  Removing a pending folder for all announcing devices is only possible through the API right now and has been tested via `curl`.

### Screenshots

Screenshot_2021-05-26 devtest Syncthing_pendingNotNow

### Documentation

See https://github.com/syncthing/docs/pull/645 for the API description.


tomasz1986:
One problem with important tooltips on clickable buttons is that they will be inaccessible on touch screen devices, as clicking the button will also trigger its action. Although, personally, I'm not sure if the tooltips are actually needed though? The buttons are kind of self-explanatory…

Also, I haven't tested how this looks in practice, but wouldn't the button fit better on the right, so that all three are kept together in one place?

acolomb:
I wanted to separate the two "final" actions from the "just get out of my way" visually. It's done that way in some dialogs IIRC. But I'm no expert on ui design guidelines, do you know any best practice rules for this?

The tooltips are not strictly needed, I agree. But they don't hurt either for devices which can display on hover?

tomasz1986:
> I wanted to separate the two "final" actions from the "just get out of my way" visually. It's done that way in some dialogs IIRC. But I'm no expert on ui design guidelines, do you know any best practice rules for this?

I'm no expert either. I only know that unrelated stuff shouldn't be put together like that, but here all of them are kind of from the same family. I think that my discomfort with the buttons being spread out like that is the fact that these pop-ups can get very wide on larger screens, making the distance between the buttons even wider. This is not the case on pop-ups inside modals, e.g. the Edit Folder one, etc. This or that, it probably won't matter much anyway :wink:.

> The tooltips are not strictly needed, I agree. But they don't hurt either for devices which can display on hover?

Personally, I'm not a big fan of them. I have tonnes of folders, and when configuring a new device or such, I have to click on those buttons a lot. Having a tooltip pop up each time the mouse flies over the button will probably get annoying very quickly.

AudriusButkevicius (approved):
Lgtm in principle.

AudriusButkevicius:
Although I agree, the options of the possible actions (buttons), should be grouped.

lib/db/observed.go
```diff
@@ -26,11 +26,13 @@ func (db *Lowlevel) AddOrUpdatePendingDevice(device protocol.DeviceID, name, add
...
 }
 
-func (db *Lowlevel) RemovePendingDevice(device protocol.DeviceID) {
+func (db *Lowlevel) RemovePendingDevice(device protocol.DeviceID) error {
 	key := db.keyer.GeneratePendingDeviceKey(nil, device[:])
 	if err := db.Delete(key); err != nil {
 		l.Warnf("Failed to remove pending device entry: %v", err)
+		return err
 	}
+	return nil
```
imsodin, line 35: This now both handles (warning log) and returns the error. Ideally both happens at the same place, e.g. no warning on the api (the error is passed on to the caller) and warning wherever it's called in the model.

lib/db/observed.go
```diff
@@ -79,32 +81,35 @@ func (db *Lowlevel) AddOrUpdatePendingFolder(id string, of ObservedFolder, devic
...
 	key, err := db.keyer.GeneratePendingFolderKey(nil, device[:], []byte(id))
 	if err != nil {
-		return
+		return err
 	}
 	if err := db.Delete(key); err != nil {
 		l.Warnf("Failed to remove pending folder entry: %v", err)
+		return err
 	}
+	return nil
```
imsodin, line 93: Same as above regarding handling/passing on errors.

lib/db/observed.go
```diff
@@ -79,32 +81,35 @@ func (db *Lowlevel) AddOrUpdatePendingFolder(id string, of ObservedFolder, devic
...
 	}
 	defer iter.Release()
 	for iter.Next() {
 		if id != string(db.keyer.FolderFromPendingFolderKey(iter.Key())) {
 			continue
 		}
-		if err := db.Delete(iter.Key()); err != nil {
+		if err = db.Delete(iter.Key()); err != nil {
 			l.Warnf("Failed to remove pending folder entry: %v", err)
 		}
```
imsodin, line 110: And once more.

gui/default/index.html
```diff
@@ -208,11 +208,14 @@ <h3 class="panel-title">
                 </p>
               </div>
               <div class="panel-footer clearfix">
+                <button type="button" class="btn btn-sm btn-default pull-left" ng-click="dismissPendingDevice(deviceID)" tooltip data-original-title="{{'Do not add it to the ignore list, so this notification may recurr.' | translate}}">
+                  <span class="far fa-clock"></span>&nbsp;<span translate>Not Now</span>
```
imsodin, line 212: I tend to prefer "Dismiss" over "Not Now" - to me the latter implicates it will come up again (e.g. just hidden until refresh/some time passed). While in reality it might come up again, but it's not guaranteed.

imsodin (commented):
I too would prefer all the control elements grouped.

gui/default/index.html
uok: or "Later"

acolomb:
> I think that my discomfort with the buttons being spread out like that is the fact that these pop-ups can get very wide on larger screens, making the distance between the buttons even wider.

IIUC the width is limited to 1170 px for the panels on large screens (>= 1200 px).  That's what the screenshot shows.  But there seems to be a consensus to keep them on one side, so I will change to that.

gui/default/index.html
acolomb: Changed to "Dismiss".

lib/db/observed.go
acolomb: I think it logically belongs to `db` logging, as it concerns an error from the database layer.  The `model` log facility is used for so many things already.  Or do you think filtering is not a concern because warnings always get through?

acolomb: By the way, model passes on the error to api, but that really seems like the wrong place to me.  I just don't feel comfortable duplicating these log calls on every call site when they could easily live within the called function.

lib/db/observed.go
imsodin: Right, I think a single warning is enough then.

lib/db/observed.go
imsodin: This is clearly a trade-off between having the logging in just one place vs having a single place that's responsible for the error.  
Now there's places where there is a `err != nil` check, and the response is `continue`. That looks like the error is just ignored. While in fact a warning is printed to the user. I wouldn't check the inner function for that, as it returned the error to me.

And the api needn't print a warning, same principle there: The error is passed on in the http response. It's now the callers responsibility do do something with it.

lib/db/observed.go
acolomb: As in "Failed to remove <count> pending folder entries: <last error>"?

lib/db/observed.go
acolomb: 1. I am going for DRY principle here.  2. If we ever get any of these warnings in a log, I think it's best pointing at the `db` package, what matters to me is what the user sees and that is rightfully described as "db: Failed...".

You seem much more worried about the developer's experience when reading the code.  Is that bikeshed really worth duplicating the same code line 7 times?  If it's unclear why the returned error is not logged from `model`, I can add comments explaining that.  Summary: The DB has *failed* to `RemovePendingBlah`, but the model has *successfully* done `cleanPending` or `ccHandleFolders`, as in there's nothing else we can do if deleting fails on a low level.  It might succeed next time or the DB might blow up to pieces soon, but the `model` does not exhibit any problem.  Hence the only thing that *can* be done (warn so the user notices the SSD is falling out) should happen close to where the error happened (where DB backend meets our application).  There is no gain in moving the log message.

So, not checking a returned error is harmful, I agree.  But deliberately ignoring it *for the model's goal* only needs a code comment at max IMO.

imsodin: 1. Do you think you can convince someone by just mentioning one of the most basic principles of our trade?  
Maybe my perception is completely off, but all I perceive from you is push-back, without even considering and addressing any points that don't align with your view. Probably a limitation of the online/written only medium we have to work with here. Anyway, I'll adjust my attitude to suit: Less focus on trying to establish common ground, more me voicing what I want - it's not something I like, but seems necessary.  
2. Warnings don't point anywhere based on where they are "placed", unless you enable additional log flags that include filename+line. If you want context on an error, use `fmt.Errorf("context: %w", err)`.

Yes, I value consistency. And here that means: Either handle the error in the db package (warn and continue), not returning an error at all, or don't handle the error, returning the error to the caller and not warning inside the db. I prefer the latter. And I don't care if it means 7 additional lines of code. Btw if you place comments there, you also end up with +7 lines of code.

acolomb: > 1. Do you think you can convince someone by just mentioning one of the most basic principles of our trade?

I just don't understand how this "most basic principle" seems to be the least important to you in this specific case?  I asked about what gain this would bring and sorry, I fail to understand how duplicating the same logging call would be *more* consistent.  If the function ever gets reused, I would find it more consistent if its failures are logged independent of where it is called from.

Your principle "handle internally **OR** report to caller" sounds fine in general.  But consequently, `model` should not log it either, because the error is passed back to API, then to the requesting HTTP client.  So logging in either package is **not** "handling" the error, but just a helpful additional hint to diagnose related problems.  And given how unlikely the error is, that hint should never be missing whether any future call site remembers to copy the error message along with the function call, or not.  The return value is only added just in case the caller does care whether it succeeded.

> Maybe my perception is completely off, but all I perceive from you is push-back, without even considering and addressing any points that don't align with your view.

Maybe we're equally stubborn.  My perception is that I have previously accepted your and others' feedback for several changes where I didn't agree.  And in some cases (like here) I just try to defend my reasoning because it's not some random oversight, but a well thought-through decision from my side.

> Probably a limitation of the online/written only medium we have to work with here. Anyway, I'll adjust my attitude to suit: Less focus on trying to establish common ground, more me voicing what I want - it's not something I like, but seems necessary.

It's definitely sometimes harder to discuss in writing, and I had wished for clear statements what you wanted on some occasions.  On others, I feel like my arguments were being overread or misread.  So let's just both focus on good written communication.

Here we genuinely disagree on which arguments are more important in the trade-off.  Since @AudriusButkevicius had already approved this PR with the current logging behavior, I was hoping for a third opinion whether there is a pressing need to change it.  If my opinion ends up representing the minority, that's okay, I will adapt.  But so far only one of three people see a problem with this code, so pardon my hesitation on a request where I personally see more trouble than benefit.

Could we postpone any further discussion to a follow-up PR, so that the uncontroversial feature part here reaches a release?  That's what I meant with "bikeshed", this PR is much more than adding an error return value.

> 2. Warnings don't point anywhere based on where they are "placed", unless you enable additional log flags that include filename+line. If you want context on an error, use `fmt.Errorf("context: %w", err)`.

Right, sorry I was confused about the default log format.  Having the package name in log lines for filtering would be useful, do you agree?

> Btw if you place comments there, you also end up with +7 lines of code.

Latest commit f2dc2e99c0699119da1c66875f6825b37e4963e4 adds comments only where the error is ignored on the model side, the other places look pretty standard and self-explanatory.

lib/db/observed.go
acolomb: Here I need a clarification what you mean.  Printing one message per failed operation is helpful I think.  But unsure if you meant the opposite, logging only one overall error message even if only one of several iterations failed.

lib/db/observed.go
imsodin: > Your principle "handle internally **OR** report to caller" sounds fine in general. But consequently, `model` should not log it either, because the error is passed back to API, then to the requesting HTTP client.

That's exactly what I proposed:

> This now both handles (warning log) and returns the error. Ideally both happens at the same place, e.g. no warning on the api (the error is passed on to the caller) and warning wherever it's called in the model.

> And the api needn't print a warning, same principle there: The error is passed on in the http response. It's now the callers responsibility do do something with it.

Anyway, just going into two more points:

> Maybe we're equally stubborn. 

I definitely don't disagree on that :)  
And I am truly impressed (and amazed) at your patience/stamina.

> Could we postpone any further discussion to a follow-up PR, so that the uncontroversial feature part here reaches a release? That's what I meant with "bikeshed", this PR is much more than adding an error return value.

That's true - this is definitely a net positive. The thing with postponing things, especially contentious points, is that they usually never get addressed then. So better to do it right where the come up. In this case I realized on re-reading that my issues with the error handling are more fundamental and were present already before this PR -> lets merge and I'll have a look myself.

imsodin (approved):

lib/db/observed.go
acolomb: > That's exactly what I proposed:
> 
> > This now both handles (warning log) and returns the error. Ideally both happens at the same place, e.g. **no warning** on the api (the error is passed on to the caller) **and warning** wherever it's called **in the model**.

Then I misunderstood because that sentence might be missing a "no" after the final "and" (emphasis added).

Thanks for merging.

PR #7716, opened 2021-05-28
build, lib/model: Add flag to run tests without -short and fix failure

imsodin:
This adds a `-long` flag to `build.go` to run all the slow test, i.e. not set `-short`.

To my surprise running with `-long` did result in a test failure, luckily just a cosmetic one: When I improved the encryption error messages, I swapped to error messages.  
We could run the coverage tests with `-long`, then it won't require much more CI time, but we at least run those tests on on platform (linux, the others being a lot slower). Long tests took 2m7s on my machine.

### DIFF
--- a/build.go
+++ b/build.go
@@ -52,7 +52,7 @@ var (
 	coverage       bool
 	long           bool
 	timeout        = "120s"
-	longTimeout    = "240s"
+	longTimeout    = "600s"
 	numVersions    = 5
 	withNextGenGUI = os.Getenv("BUILD_NEXT_GEN_GUI") != ""
 )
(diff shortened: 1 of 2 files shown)


build.go
```diff
@@ -374,7 +377,12 @@ func test(tags []string, pkgs ...string) {
 	lazyRebuildAssets()
 
 	tags = append(tags, "purego")
-	args := []string{"test", "-short", "-timeout", timeout, "-tags", strings.Join(tags, " ")}
+	args := []string{"test", "-tags", strings.Join(tags, " ")}
+	if !long {
+		timeout = longTimeout
```
calmh, line 382: Logic error, we're in the -short path here

build.go
```diff
@@ -50,7 +50,9 @@ var (
 	benchRun       string
 	debugBinary    bool
 	coverage       bool
+	long           bool
 	timeout        = "120s"
+	longTimeout    = "240s"
```
calmh, line 55: I'd make this even longer

calmh (changes requested):

calmh (approved):

PR #7720, opened 2021-05-29
lib/db: Clean up Timer and wait for logging before return in GC

greatroar:
<del>The GC's logging goroutine races with its parent on the variables (discarded|matched)(Blocks|Versions). This new version only accesses the variables in one goroutine, so it's safe. It also cleans up its Timer.</del>

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -693,20 +693,19 @@ func (db *Lowlevel) gcIndirect(ctx context.Context) (err error) {
 
 	var discardedBlocks, matchedBlocks, discardedVersions, matchedVersions int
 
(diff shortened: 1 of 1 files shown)


calmh:
I don't actually think it's a race though? It accesses the variables only after internalCtx is cancelled, and if the parent context weren't cancelled. The internalCtx is cancelled by a defer on the gc runner.

greatroar:
Oh, you're right. Channel communication is a kind of barrier. I thought it only synchronized access to values sent over the channel. I'll leave the PR open with a new commit message, but feel free to close it.

imsodin (approved):

PR #7726, opened 2021-05-31
gui, lib: Fix tracking deleted locally-changed on encrypted (fixes #7715)

### LINKED ISSUE #7708: irreversible Local Additions after re-adding the same folder as Receive Only
This is a long-standing problem that I have experienced multiple times, but only now I have managed to reproduce it.

1. On Device 1, create a new folder called `test1`.
2. In the folder `test1`, create a folder called `1`.
3. In the folder `1`, create a file called `1`.
3. On Device 2, create a new folder called `test2`.
4. In the folder `test2`, create a folder called `2`.
5. In the folder `2`, create a file called `2`.

The first steps are done on the file system level, not in Syncthing.

5. On Device 1, add the `test1` folder to Syncthing, and share it with Device 2.
6. On Device 2, accept the folder as SR, pointing it at the `test2` folder.
7. The contents of the two folders have been merged and synced.

   image
   image

8. On Device 2, remove the folder from Syncthing.
9. Then, re-add it again, but this time as RO.
10. The folder is now marked as "Local Additions".

    image
    image
    image

11. Reverting local changes doesn't clear all the errors either.

    image

    ```
    INFO: Revert: directory has been deleted on a remote device but is not empty; the contents are probably ignored on that remote device, but not locally
    ```

Logs with `db` and `model`:

[syncthing1.log](https://github.com/syncthing/syncthing/files/6527626/syncthing1.log)
[syncthing2.log](https://github.com/syncthing/syncthing/files/6527627/syncthing2.log)

### LINKED ISSUE #7715: Deleted encrypted files don't show up as locally changed in web UI
If you delete an encrypted file in a receive-only folder, i.e. do that manually in the filesystem, that won't register in the UI. It will look like everything is ok.

Reason is that we have an ugly hack to satisfy two requirements:

 - Locally changed items from a receive-encrypted folder must never be sent to remotes.
 - Locally added items that are deleted again aren't a problem, thus shouldn't show as local changes in the web UI.

The ugly hack is to ignore all items that have the locally-changed flag set and are deleted. That's fine for actually spurious files, that were removed, but obviously not for files that should be there.

PR incoming.

imsodin:
This PR removes the ugly hack explained in #7715:

> The ugly hack is to ignore all items that have the locally-changed flag set and are deleted. That's fine for actually spurious files, that were removed, but obviously not for files that should be there.

Instead it adds the non-ugly, but somewhat scary ability to remove file infos from the database. We already do that, but only in bulk when a device changes index-id/is removed. It isn't rocket science, it's just the half of `updateLocalFiles` that removes the old file. The scary bit is just that the code it touches is already scary.  
An alternative to this approach would be adding a new flag (e.g. `FlagLocalReceiveEncryptedDeleted`) and mark locally-changed, deleted items in receive-encrypted folders with that flag. This sidesteps the database changes, but instead of getting rid of useless file-infos, it keeps them with a new flag. That's why I'd prefer the db variant.

To facilitate removing as well as updating files while scanning, I added a `scanBatch` type with `Update` and `Remove` methods, that replaces the `batchAppendFunction` that was passed around as a function argument before.  

### Testing

Added one new test that ensures that spurious files added to either receive-only or receive-encrypted folders, that are subsequently deleted, don't show up as locally-changed. And extended an existing test to check the same thing.


calmh:
Can you explain in human words what the mechanism here is? I don't fully follow.

imsodin:
I can try:

Two requirements with whatever the solution is (https://github.com/syncthing/syncthing/issues/7715#issue-905695427):

>   -  Locally changed items from a receive-encrypted folder must never be sent to remotes.
>   -  Locally added items that are deleted again aren't a problem, thus shouldn't show as local changes in the web UI.

**Status quo:**

- For receive-encrypted folders we filter out items with `FlagLocalReceiveOnly` when sending indexes. We keep that flag even if a file is deleted and only ever existed locally. For receive-only folders we remove the flag, because a deleted file that doesn't exist anywhere is not a concern, neither for sent indexes nor should it show up in the web UI as a locally changed item.
- The web UI and `Model.LocalChangedFolderFiles` filter out deleted items with `FlagLocalReceiveOnly` for receive-encrypted folders.

Problem: Valid files that exist globally, but the user mistakenly deleted locally, aren't shown as missing in the UI and thus can't be recreated (by reverting).

**Status PR:**

Instead of removing the flag for deleted, locally changed files that only ever existed locally, like we do now for receive-only folders, we delete the item from the database.

 - The item isn't in the database, thus isn't going to be sent as index.
 - Same here: No need to ignore/filter anything towards the web UI, the item simply is gone.

**How is it done:**

Add a new method `(s *FileSet) RemoveLocalItems(items []string)`.  
On the db side, I re-use the code from `updateLocalFiles` that deals with removing the existing item from the db.  
On the model/folder/scanner side, I extend the batch that holds scanned/changed items to be updated, to also hold items to be removed from the db (`scanBatch`). Same mechanism: Once the limit is reached (1000 items), `RemoveLocalFiles` is called. What previously was `scanSubdirsBatchAppendFunc`, is now `(b *scanBatch) Update`.

calmh:
I need to think about this a little bit more, but I don't think we can just remove things from the database. Those entries will have taken up a sequence number, removing them and doing a meta recalculation would rewind the sequence number which is not great.

imsodin:
That's a good point: Recalculating metadata and not resetting delta indexes seems like a dangerous thing to do in general. Even today where there *should* be no way the sequence number decreases when recalculating metadata, I think it would be better to ensure it doesn't (or reset delta indexes).

st-review:
:robot: *beep boop*

imsodin:
Hmm, no. This is actually still a relevant bug-fix. Just unfortunately touching the db, i.e. an annoying one. @calmh If you'd prefer the alternative with the new flag (see below) instead of removing file-infos from db, I could create a PR for that (I'd still prefer the current approach, but very much understand the hesitance to touch/change this core/database part):

> An alternative to this approach would be adding a new flag (e.g. FlagLocalReceiveEncryptedDeleted) and mark locally-changed, deleted items in receive-encrypted folders with that flag. This sidesteps the database changes, but instead of getting rid of useless file-infos, it keeps them with a new flag. That's why I'd prefer the db variant.

calmh:
I don't like either option as it's all intrusive, complicated and annoying. :) But thinking about it again I think it's fine to remove things from the database that should be unseen. At some point I want to revive my PR to remove old deleted items as well, in order to prevent infinite index growth, and that's similar.

The sequence number concern is there, but if it happens (which would only be if the removed item is the latest one) it should trigger the "out of sync with realize" path in delta indexes and force a full retransmission, fixing the issue.

calmh:
So with that out of the way I think this is fine; is it good for review (still?) or needs something to revive?

imsodin:
The "sequence index going backwards on recalculation"-issue is even explicitly handled now: https://github.com/syncthing/syncthing/pull/7770

It was ready before, had a quick look over it now given it's been a long while, looks good and passes tests still after merging - ready for review.  
Plus I very much agree with the following xD
> I don't like either option as it's all intrusive, complicated and annoying. :)

calmh (approved):

Issue #934, opened 2014-11-03
Harmonize units used for transfer rates (limit vs current etc)


brunorochan:
![sync1](https://cloud.githubusercontent.com/assets/9532129/4883876/9b2439e8-6366-11e4-86fe-86cfd74d67ba.png)
![sync2](https://cloud.githubusercontent.com/assets/9532129/4883878/a865c3ce-6366-11e4-920c-493cbf966c68.png)

As you guys can see, upload and download limits are set, but Syncthing always ignore them. Please, help.

Issue #900, opened 2014-10-23
Units of data rates and file sizes


uok:
Display data rates in KB/s, MB/s, ... (not kbps, mbps) and files sizes in KB, MB, GB, ... (not MiB, GiB) since these units are more common and intuitive.

label: [enhancement]

PR #7727, opened 2021-05-31
gui: Default to bps for download/upload rates

### LINKED ISSUE #7725: Display down- and upload rates with bps per default
Download and upload rates describe network bandwidth usage and should therefore use the appropriate unit by default. The current default of B/s is confusing for new users as they will compare the values with speedtest results or ratings of their network gear which all use bps.

Case in point: https://forum.syncthing.net/t/please-help-how-can-i-get-speed-higher-than-2-mb-s/16859/11

bt90:
Fixes #7725 by displaying download- and upload rates with `bps` per default.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -49,7 +49,7 @@ angular.module('syncthing.core')
         $scope.scanProgress = {};
         $scope.themes = [];
         $scope.globalChangeEvents = {};
-        $scope.metricRates = false;
+        $scope.metricRates = true;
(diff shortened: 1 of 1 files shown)


acolomb (approved):

imsodin:
https://github.com/syncthing/syncthing/issues/7725#issuecomment-852983373 by @calmh:

> No. We used to do precisely this. It caused no end of complaints and confusion. #900 #934 and I'm sure all kinds of threads on the forum from way back.

I am pretty sure in the end either option will find users that are confused by them.

PR #7728, opened 2021-05-31
gui: Show ID under each device section (local / remote), clickable for QR code

acolomb:
### Purpose

Show the abbreviated device ID as clickable link, leading to the respective `idqrModalView` showing.  Supersedes / replaces #7660.

### Screenshots

Screenshot_2021-05-31 devtest Syncthing_IDQR lines

### Documentation

Not strictly required except for outdated screenshots.  All previous ways to access the modal still remain.


tomasz1986:
How about adding a small help tooltip like the one used for discovery failures?

image

The text could say "Click to see full identification string" or something in this vein.

acolomb:
> How about adding a small help tooltip like the one used for discovery failures?

I thought you were not a fan of tooltips ;-)  Will add one.

gui/default/index.html
```diff
@@ -682,6 +682,14 @@ <h4 class="panel-title">
                       <th><span class="far fa-fw fa-clock"></span>&nbsp;<span translate>Uptime</span></th>
                       <td class="text-right">{{system.uptime | duration:"m"}}</td>
                     </tr>
+                    <tr>
+                      <th><span class="fas fa-fw fa-qrcode"></span>&nbsp;<span translate>Identification</span></th>
```
imsodin, line 686: Let's keep calling it "Device ID" to keep consistency with e.g. the device edit dialog.

gui/default/index.html
```diff
@@ -682,6 +682,14 @@ <h4 class="panel-title">
...
+                      <td class="text-right">
+                        <span tooltip data-original-title="{{'Click to see full identification string and QR code.' | translate}}">
+                          <a href="" style="color:inherit" ng-click="showDeviceIdentification(thisDevice())">{{deviceShortID(deviceCfg.deviceID)}}</a>
```
imsodin, line 689: We use link styling for clickable items in other rows, so I'd do the same here. Currently I don't see any hint that there's an action when clicking the short device ID.

imsodin (commented):
Nice!

gui/default/index.html
tomasz1986: Isn't "Device ID" a little superfluous (being located under my device and other devices? "Identification" in the long form is actually used right there in the linked pop-up. Just "ID" is also an option though.

tomasz1986 (commented):
> > How about adding a small help tooltip like the one used for discovery failures?
> 
> I thought you were not a fan of tooltips ;-) Will add one.

Not a fan in general, but there isn't really any place for additional explanation here, and the adjacent links already utilise this style of tooltips, so for the sake of consistency, I would say that it's better to have it than not 😉.

gui/default/index.html
imsodin: Yeah I am mostly worried about translations. For English the link between "ID" and "Identification" is indeed obvious, but I wouldn't be surprised if "Device ID" and "Identification" get translated to terms that aren't obviously matched. Thus I'd prefer a bit of redundancy to ensure consistency.

tomasz1986: This will vary from language to language, but I can confirm that, e.g. in Korean, "ID" is translated as "ID" (i.e. is left untranslated), but "identification" is translated to a Korean word. There is no direct connection between the two, other than the placement in the Web GUI.

gui/default/index.html
acolomb: Done.

gui/default/index.html
acolomb: My thinking was also that "Device" is just repeating the context, and "ID" is just an abbreviation of "Identifier" or "Identification".  Seeing that the modal that opens on click already had "Device Identification" as its title, I adapted to that *also for consistency*.

Changed to "Device ID".

imsodin: I agree with that reasoning (I wasn't aware "Device Identification" is already in use/translated). Feel free to revert that change, I'll re-approve. Or merge as is.

imsodin (dismissed):

gui/default/index.html
acolomb: Reverted, thanks.

imsodin (approved):

PR #7739, opened 2021-06-02
lib/model: Fix passwords on receive-enc needing token (ref #7518)

imsodin:
I was testing some encryption related UI changes (not this PR) when suddenly all hell broke loose: The trusted side treated received encrypted file indexes like normal files. Reason is that the change in #7198 is just plain broken, luckily it only matters if you have encrypted folders both ways (folder1: A untrusted <-> B, folder2: A <-> B untrusted). That is uncommon enough that no-one seems to have run into it yet - yeii for not cleaning up old test folders on test setups.

The completely broken part is returning early from `generateClusterConfig` with `passwords == nil`, and then ignoring that error in `AddConnection`, thus going ahead and calling `Connection.SetPasswords` with a nil map.

I basically reverted #7518 and to keep #7198 fixed, I added a new check to `CommitConfiguration`: If a receive-encrypted folder changes, and it doesn't have the token yet, don't just resend a cluster config, but drop the connection (to get the token).

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -192,12 +192,9 @@ var (
 	errEncryptionPassword                 = errors.New("different encryption passwords used")
 	errEncryptionTokenRead                = errors.New("failed to read encryption token")
 	errEncryptionTokenWrite               = errors.New("failed to write encryption token")
-	errEncryptionNeedToken                = errors.New("require password token for receive-encrypted token")
 	errMissingRemoteInClusterConfig       = errors.New("remote device missing in cluster config")
 	errMissingLocalInClusterConfig        = errors.New("local device missing in cluster config")
 	errConnLimitReached                   = errors.New("connection limit reached")
-	// messages for failure reports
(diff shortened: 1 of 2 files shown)


lib/model/model.go
```diff
@@ -2939,11 +2935,13 @@ func (m *model) CommitConfiguration(from, to config.Configuration) bool {
 
 	m.pmut.RLock()
 	for _, id := range closeDevices {
+		delete(clusterConfigDevices, id)
 		if conn, ok := m.conn[id]; ok {
 			go conn.Close(errDevicePaused)
```
calmh, line 2940: We might be sending multiple closes here for the same connection; I'm going to assume that's OK.

calmh (approved):

lib/model/model.go
imsodin: Yes, that's a noop then (there's a `sync.Once` in `protocol.Connection`).

PR #7750, opened 2021-06-05
lib/db: Use a more concurrent GC (fixes #7722)

### LINKED ISSUE #7722: Database GC can block for a long time
https://forum.syncthing.net/t/syncthing-stuck-gui-seemingly-responsive-but-nothing-is-happening/16828

Syncthing can become completely unresponsive when GC happens. Ideally that doesn't take long (e.g. ~10s for a largish setup with DB on SSD for me), but if DB is on slow storage it takes a long time. Since https://github.com/syncthing/syncthing/pull/7707 there's now at least an indication in the log that GC is happening and might render Syncthing unresponsive until it's finished.

Options to improve the situation:
 - Don't do db compaction. Might be a good idea anyway, considering the comment on the method at goleveldb that pretty much advises against using it.
 - Do most of the GC without holding the global mutex, and only acquire that for the final part when actually deleting the entries.

calmh:
This changes the GC mechanism so that the first pass (which reads all
FileInfos to populate bloom filters with block & version hashes) can
happen concurrently with normal database operations.

The big gcMut still exists, and we grab it temporarily to block all
other modifications while we set up the bloom filters. We then release
the lock and let other things happen, with those other things also
updating the bloom filters as required. Once the first phase is done we
again grab the gcMut, knowing that we are the sole modifier of the
database, and do the cleanup.

I also removed the final compaction step.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -73,7 +73,6 @@ type Lowlevel struct {
 
 	blockFilter   *bloomFilter
 	versionFilter *bloomFilter
-	filtersMut    sync.Mutex
 }
 
 func NewLowlevel(backend backend.Backend, evLogger events.Logger, opts ...Option) (*Lowlevel, error) {
@@ -89,7 +88,6 @@ func NewLowlevel(backend backend.Backend, evLogger events.Logger, opts ...Option
 		recheckInterval:    recheckDefaultInterval,
 		oneFileSetCreated:  make(chan struct{}),
 		evLogger:           evLogger,
-		filtersMut:         sync.NewMutex(),
 	}
 	for _, opt := range opts {
 		opt(db)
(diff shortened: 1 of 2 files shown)


lib/db/lowlevel.go
```diff
@@ -826,16 +832,31 @@ func (db *Lowlevel) gcIndirect(ctx context.Context) (err error) {
...
+	return nil
+}
+
+func (db *Lowlevel) recordIndirectionHashesForFile(f *protocol.FileInfo) {
+	db.recordIndirectionHashes(IndirectionHashesOnly{BlocksHash: f.BlocksHash, VersionHash: f.VersionHash})
+}
 
-	return db.Compact()
+func (db *Lowlevel) recordIndirectionHashes(hs IndirectionHashesOnly) {
+	db.filtersMut.Lock()
```
imsodin, line 845: As the lock only protects concurrent access to the blockfilters, but not the blockfilter variables (that's `gcMut`), we could first do the nil checks and abort early, thus preventing pointless locking.

tomasz1986:
The commit removes the INFO logging added by @imsodin some time ago. Is it because GC will be so quick (or done concurrently in the background) that the logging is not necessary anymore? I'm asking, because the reason for adding the logging seemed to be to have some kind of an indicator that GC was ongoing, as otherwise the GUI just seemed stuck with no reason.

calmh:
It can still take a while, just the critical section is shorter; I reinstated and moved the logging.

lib/db/lowlevel.go
greatroar: Blobloom has a SyncFilter type that uses atomic operations for Add and Has. It can be used to omit this lock entirely.

calmh: @imsodin Right!

@greatroar Thanks, I wasn't aware. In this case though we really only need the locking on the Add() operation.

calmh: Looking at it I see it's probably more efficient to use a SyncFilter anyway. Doing that instead. 👍

imsodin (approved):

PR #7753, opened 2021-06-07
lib/connections: Trigger dialer when connection gets closed

imsodin:
When a connection is closed (config change, pausing/resuming, error, ...) it may take a while for a connection to be re-established (up to a minute by default). When just running unsupervised that's fine, but when setting something up in the UI or debugging it's annoying. Now the connection service tracks the lifetime of established connections and triggers dialling when it's closed.

That requires removing all keys in `nextDialAt` for a device. To simplify that, I refactored it to have a (typed) double map instead of a composed `*deviceID*/*addr*` string key.

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -142,10 +142,13 @@ type service struct {
 	natService           *nat.Service
 	evLogger             events.Logger
 
-	deviceAddressesChanged chan struct{}
-	listenersMut           sync.RWMutex
-	listeners              map[string]genericListener
-	listenerTokens         map[string]suture.ServiceToken
+	dialNow           chan struct{}
+	dialNowDevices    map[protocol.DeviceID]struct{}
+	dialNowDevicesMut sync.Mutex
+
+	listenersMut   sync.RWMutex
+	listeners      map[string]genericListener
+	listenerTokens map[string]suture.ServiceToken
 }
 
 func NewService(cfg config.Wrapper, myID protocol.DeviceID, mdl Model, tlsCfg *tls.Config, discoverer discover.Finder, bepProtocolName string, tlsDefaultCommonName string, evLogger events.Logger) Service {
@@ -166,10 +169,13 @@ func NewService(cfg config.Wrapper, myID protocol.DeviceID, mdl Model, tlsCfg *t
 		natService:           nat.NewService(myID, cfg),
 		evLogger:             evLogger,
 
-		deviceAddressesChanged: make(chan struct{}, 1),
-		listenersMut:           sync.NewRWMutex(),
-		listeners:              make(map[string]genericListener),
-		listenerTokens:         make(map[string]suture.ServiceToken),
+		dialNowDevicesMut: sync.NewMutex(),
+		dialNow:           make(chan struct{}, 1),
+		dialNowDevices:    make(map[protocol.DeviceID]struct{}),
+
+		listenersMut:   sync.NewRWMutex(),
+		listeners:      make(map[string]genericListener),
+		listenerTokens: make(map[string]suture.ServiceToken),
 	}
 	cfg.Subscribe(service)
 
@@ -324,6 +330,13 @@ func (s *service) handle(ctx context.Context) error {
 		rd, wr := s.limiter.getLimiters(remoteID, c, isLAN)
 
 		protoConn := protocol.NewConnection(remoteID, rd, wr, c, s.model, c, deviceCfg.Compression, s.cfg.FolderPasswords(remoteID))
+		go func() {
+			<-protoConn.Closed()
+			s.dialNowDevicesMut.Lock()
+			s.dialNowDevices[remoteID] = struct{}{}
+			s.scheduleDialNow()
+			s.dialNowDevicesMut.Unlock()
+		}()
 
 		l.Infof("Established secure connection to %s at %s", remoteID, c)
 
@@ -334,7 +347,7 @@ func (s *service) handle(ctx context.Context) error {
 
 func (s *service) connect(ctx context.Context) error {
 	// Map of when to earliest dial each given device + address again
-	nextDialAt := make(map[string]time.Time)
+	nextDialAt := make(nextDialRegistry)
 
 	// Used as delay for the first few connection attempts (adjusted up to
 	// minConnectionLoopSleep), increased exponentially until it reaches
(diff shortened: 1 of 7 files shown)


calmh:
This might have been easier than the whole clusterconfig-on-existing-connection-let's-refactor-all-the-index-senders thing ;]

lib/connections/service.go
```diff
@@ -324,6 +330,13 @@ func (s *service) handle(ctx context.Context) error {
 		rd, wr := s.limiter.getLimiters(remoteID, c, isLAN)
 
 		protoConn := protocol.NewConnection(remoteID, rd, wr, c, s.model, c, deviceCfg.Compression, s.cfg.FolderPasswords(remoteID))
+		go func() {
+			<-protoConn.Closed()
+			s.dialNowDevicesMut.Lock()
+			s.dialNowDevices[remoteID] = struct{}{}
+			s.scheduleDialNow()
+			s.dialNowDevicesMut.Unlock()
+		}()
```
calmh, line 339: Could we get stuck in a fast infinite connection loop here if the remote side is dropping us due to paused or so?

calmh (commented):
Wouldn't it be sufficient to just schedule an immediate dial loop on config changes?

lib/connections/service.go
imsodin: Yep, added a one minute minimum delay on closed connections. Pausing is done at hello stage, i.e. before this stage, but there still might be problems with cluster-config or a plainly misbehaving remote.

lib/model/fakeconns_test.go
```diff
@@ -28,13 +28,16 @@ func newFakeConnection(id protocol.DeviceID, model Model) *fakeConnection {
 		id:         id,
 		model:      model,
 		closed:     make(chan struct{}),
+		closeOnce:  sync.Once{},
```
calmh, line 31: no need to be so explicit

```suggestion
```

lib/connections/service.go
```diff
@@ -324,6 +331,16 @@ func (s *service) handle(ctx context.Context) error {
 		rd, wr := s.limiter.getLimiters(remoteID, c, isLAN)
 
 		protoConn := protocol.NewConnection(remoteID, rd, wr, c, s.model, c, deviceCfg.Compression, s.cfg.FolderPasswords(remoteID))
+		go func() {
+			// Don't immedately redial if the connection is dropped quickly.
+			delay := time.NewTimer(closedConnectionRedialMinimumAge)
+			<-protoConn.Closed()
+			<-delay.C
```
calmh, line 338: Could check the connection's `EstablishedAt()` and just skip the rest here if it's too young. Once we're up to waiting a minute and such we can anyway let the regular redial intervals handle it I think.

lib/connections/service.go
```diff
@@ -1050,3 +1075,32 @@ func (s *service) validateIdentity(c internalConn, expectedID protocol.DeviceID)
 
 	return nil
 }
+
+type nextDialRegistry map[protocol.DeviceID]map[string]time.Time
+
+func (r nextDialRegistry) get(device protocol.DeviceID, addr string) (time.Time, bool) {
+	if _, ok := r[device]; !ok {
+		return time.Time{}, false
+	}
```
calmh, line 1084: not required

lib/connections/service.go
```diff
@@ -1050,3 +1075,32 @@ func (s *service) validateIdentity(c internalConn, expectedID protocol.DeviceID)
...
+	if _, ok := r[device][addr]; !ok {
+		return time.Time{}, false
+	}
+	return r[device][addr], true
```
calmh, line 1088: Also not really required as this could be shortened to

```
t := r[device][addr]
return t, !t.IsZero()
```

or just letting the caller handle the fact that the time might be zero -- which might happen by default this this effectively asks for a dial in the distant past?

lib/connections/service.go
```diff
@@ -380,7 +397,14 @@ func (s *service) connect(ctx context.Context) error {
 		l.Debugln("Next connection loop in", sleep)
 
 		select {
-		case <-s.deviceAddressesChanged:
+		case <-s.dialNow:
+			// Remove affected devices from nextDialAt to dial immediately,
+			// regardless of when we last dialed it.
+			s.dialNowDevicesMut.Lock()
+			for device := range s.dialNowDevices {
+				delete(nextDialAt, device)
```
calmh, line 405: This should be a method on the registry now

lib/connections/service.go
imsodin: I replaced it alltogether: The benefit was, that despite the delay, for any connection closed after the initial delay, a reconnect would happen instantaneously instead of after up to the regular redial interval (1 minute). However that indeed still delays legitimate connection drops short after establishing a connection (e.g. due to cluster-config stuff). I thus added a cooldown mechanism to the redial loop instead.

calmh:
Instead of making the dialcooldownentry a separate entity, could we teach the next dial-registry to keep some history and moderate dials to something like not more than three times in the last two minutes or something? We could then ask it for a dial "now" and either get that or get a time that slightly in the future depending on how much dialing we've done recently? This whole thing is starting to feel a bit over engineered, but that way it would be limited to one place where we have logic for next dial per address

imsodin:
Definitely nicer with everything in one place and abstracted away from the main loop - thanks for the hint. It's all in the registry now. And I get the over-engineered bit, I just really want to have the immediate dial functionality, and it feels adequately prudent to guard against constantly redialling.

lib/connections/service.go
```diff
@@ -379,12 +392,23 @@ func (s *service) connect(ctx context.Context) error {
...
+		case <-s.dialNow:
+			// Remove affected devices from nextDialAt to dial immediately,
+			// regardless of when we last dialed it (there's cool down in the
+			// registry for too many repeat dials).
+			s.dialNowDevicesMut.Lock()
+			for device := range s.dialNowDevices {
+				nextDialAt.redialDevice(device, now)
+			}
+			s.dialNowDevices = make(map[protocol.DeviceID]struct{})
+			s.dialNowDevicesMut.Unlock()
```
calmh, line 406: cancel the timeout timer here for cleanliness?

lib/connections/service.go
```diff
@@ -1050,3 +1057,89 @@ func (s *service) validateIdentity(c internalConn, expectedID protocol.DeviceID)
...
+		}
+		if len(dev.nextDial) == 0 {
+			delete(r, id)
+		}
+		if dev.attempts > 0 {
+			interval := dialCoolDownInterval
+			if dev.attempts >= dialCoolDownMaxAttemps {
+				interval = dialCoolDownDelay
+			}
+			if now.Before(dev.coolDownIntervalStart.Add(interval)) {
```
calmh, line 1139: Maybe this should be an After, maybe I'm not following the logic -- either way it could be cleared by a couple of tests on the cooldown logic :)

calmh (approved):

PR #7769, opened 2021-06-14
gui: Fix typo in new tooltip.

acolomb:
Fixes a typo introduced in #7712.  That happened after the last release, but has been included in the `-rc.1` version and propagated to transifex.

### DIFF
--- a/gui/default/assets/lang/lang-en.json
+++ b/gui/default/assets/lang/lang-en.json
@@ -95,7 +95,7 @@
    "Discovery": "Discovery",
    "Discovery Failures": "Discovery Failures",
    "Dismiss": "Dismiss",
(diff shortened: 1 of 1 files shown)


acolomb:
Would it be okay to push such a small fix directly to main, skipping a PR?

AudriusButkevicius (approved):

AudriusButkevicius:
I don't think anyone should be modifying main branch.
I think the only person that does that is @calmh during releases or in case of emergency.

acolomb:
Alright, thanks for clarifying.

tomasz1986:
Transifex has now two strings listed at the moment.

image

Is this something that is going to get fixed by itself?

acolomb:
Sorry I think git has not picked up the modification in index.html, only in the language file. My bad for committing an incomplete fix.

PR #7770, opened 2021-06-16
lib/db: Check if sequences change when repairing metadata

imsodin:
Topic came up in https://github.com/syncthing/syncthing/pull/7726#issuecomment-855205470:

calmh:
> [...] Those entries will have taken up a sequence number, removing them and doing a meta recalculation would rewind the sequence number which is not great.

imsodin:
> That's a good point: Recalculating metadata and not resetting delta indexes seems like a dangerous thing to do in general. Even today where there should be no way the sequence number decreases when recalculating metadata, I think it would be better to ensure it doesn't (or reset delta indexes).

I added a step at the end of the db check/repair that compares meta-sequences for all devices between before and after repair. If any of them are changed, the respective index id is reset.

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -954,7 +954,7 @@ func (db *Lowlevel) getMetaAndCheckGCLocked(folder string) (*metadataTracker, er
 		}
 	}
 
-	if err := db.checkSequencesUnchanged([]byte(folder), oldMeta, meta); err != nil {
+	if err := db.checkSequencesUnchanged(folder, oldMeta, meta); err != nil {
 		return nil, fmt.Errorf("checking for changed sequences: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


lib/db/lowlevel.go
```diff
@@ -1281,6 +1286,57 @@ func (db *Lowlevel) checkLocalNeed(folder []byte) (int, error) {
...
+			return err
+		}
+		return t.Delete(key)
+	}
+
+	if oldMeta.Sequence(protocol.LocalDeviceID) != meta.Sequence(protocol.LocalDeviceID) {
+		if err := deleteIndexID(protocol.LocalDeviceID); err != nil {
+			return err
+		}
+		l.Infoln("Local sequence changed while repairing - dropping delta indexes")
```
calmh, line 1311: Might want to mention the folder in these messages as I guess it might happen for several folders if it happens at all

calmh (commented):
I think it's overkill to drop index IDs for lower than expected remote IDs, it just means we didn't write the metadata record in time and will get duplicates of those updates (which should be fine...)?  Anyway it might be nice with failure counters here as well so we can know what happens (and in which direction the problem was (lower/higher), maybe).

calmh (dismissed):

imsodin:
> I think it's overkill to drop index IDs for lower than expected remote IDs, it just means we didn't write the metadata record in time and will get duplicates of those updates (which should be fine...)? 

Should indeed be fine. I think it *should* not happen that we do write file infos to db, but not metadata, as we write both at the same time. However who knows what the actual db layer does.

calmh (approved):

PR #7771, opened 2021-06-16
gui: Fix race in online event callback (fixes #7733)

### LINKED ISSUE #7733: TypeError: Cannot read property ‘urAccepted’ of undefined
```
logbar.js:11 TypeError: Cannot read property 'urAccepted' of undefined
    at syncthingController.js:135
    at angular.js:9452
    at processQueue (angular.js:13337)
    at angular.js:13353
    at Scope.$eval (angular.js:14589)
    at Scope.$digest (angular.js:14405)
    at angular.js:14984
    at angular.js:5347
console.<computed> @ logbar.js:11
```

Ref: https://forum.syncthing.net/t/typeerror-cannot-read-property-uraccepted-of-undefined/16844

imsodin:
The various refresh functions that aren't debounced now return a promise, and the online event handler uses those to prevent racing (and thus potentially accessing undefined stuff) with the refresh functions it started.

I found a similar issue in `saveConfig`, where the callback is called before `refreshConfig` is actually done.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -109,46 +109,49 @@ angular.module('syncthing.core')
 
             console.log('UIOnline');
 
-            refreshSystem();
-            refreshDiscoveryCache();
-            refreshConfig();
-            refreshCluster();
-            refreshConnectionStats();
             refreshDeviceStats();
             refreshFolderStats();
             refreshGlobalChanges();
             refreshThemes();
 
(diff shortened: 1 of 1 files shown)


calmh (commented):
There's some diff highlighting that looks like indentation may be mixed tabs and spaces, but otherwise this all seems fine to me.

calmh (approved):

imsodin:
That's just github highlighting what the changes are between lines - it's 4 spaces, no tabs :)

PR #7772, opened 2021-06-16
gui, lib: Handle pw adding remote encrypted folder (fixes #7705)

### LINKED ISSUE #7705: Accepting a receive encrypted folder share should prompt the user to input a password
1. On Device 1, create a new folder with some files and share it with Device 2.
2. On Device 2, accept the new folder as SR, and also share it with password with Device 3.
3. On Device 3, accept the new folder as RE, and also share it with Device 1.
4. On Device 1, the Share Folder pop-up shows.
   image
5. However, clicking the "Share" button results in errors on Device 1 and Device 3.
   image
   image

For better user experience, I think that it would be nice to have Syncthing prompt the user on Device 1 to input the password, especially since s/he may not even know that the folder is set to RE on the remote device. I would also argue that the error message is way too technical for an average user to understand what is even going on :wink:.

imsodin:
Goal: When adding/sharing a folder in the UI, and the remote is untrusted (has the folder as receive-encrypted), the user is prompted to enter the encryption password (unless the local folder is also receive-encrypted).

To do that the information whether the remote has encrypted data is persisted in the pending folder entries in db. The folder modal checks for those in addition to the untrusted config setting to decide if the user must enter the password. And sharing a folder also checks the pending entry to open the folder edit modal in the sharing tab in case the remote is encrypted (instead of silently adding it "normally", resulting in a connection error).


gui/default/syncthing/core/syncthingController.js
```diff
@@ -2114,11 +2122,19 @@ angular.module('syncthing.core')
...
+                $scope.folders[folder].devices.push({
+                    deviceID: device
+                });
+                $scope.config.folders = folderList($scope.folders);
+                $scope.saveConfig();
+            } else {
+                // Open edit folder dialog to enter encryption password
+                $scope.editFolderExisting(folderCfg);
+                $scope.currentSharing.selected[device] = true;
+                $('.nav-tabs a[href="#folder-sharing"]').tab('show');
```
calmh, line 2136: This is not a beautiful coupling of presentation and logic...

calmh (approved):

gui/default/syncthing/core/syncthingController.js
imsodin: Made the tab to show on open a function parameter, such that this can be done in `editFolderModal`. Not sure if this act of beautification (if it is) isn't entirely lost in the overall ... non-beauty of `syncthingController.js` :)

calmh: Yeah I don't think this makes it much better or worse really :)

calmh (approved):

PR #7775, opened 2021-06-17
Added firewalld service files

lclutz:
### Purpose

Provide firewalld service files for users using Syncthing on systems with the firewalld service running.
These files are the firewalld equivalent of the UFW application presets already provided in this repo.

### Testing

N/A

### Screenshots

N/A

### Documentation

N/A

### DIFF
--- a/etc/firewall-firewalld/syncthing.xml
+++ b/etc/firewall-firewalld/syncthing.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Syncthing</short>
+  <description>Syncthing file synchronization</description>
+  <port protocol="udp" port="22000"/>
+  <port protocol="udp" port="21027"/>
+</service>
+
--- a/etc/firewall-firewalld/README.md
+++ b/etc/firewall-firewalld/README.md
@@ -1,0 +1,25 @@
+Firewalld services
+==================
+Installation
+------------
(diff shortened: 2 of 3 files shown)


etc/firewall-firewalld/syncthing.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Syncthing</short>
+  <description>Syncthing file synchronization</description>
+  <port protocol="udp" port="22000"/>
```
ProactiveServices, line 5: You also need to allow tcp/22000 and locally, 21027/udp (if there is such a thing, in firewalld) as per the [firewall docs](https://docs.syncthing.net/users/firewall.html#firewall-setup).

ProactiveServices:
I'm unfamiliar with firewalld but the current Syncthing [docs](https://docs.syncthing.net/users/firewall.html#firewalld) say: "If you are using Firewalld it has included support for syncthing (since version 0.5.0, January 2018), and you can enable it with..."

lclutz:
> I'm unfamiliar with firewalld but the current Syncthing [docs](https://docs.syncthing.net/users/firewall.html#firewalld) say: "If you are using Firewalld it has included support for syncthing (since version 0.5.0, January 2018), and you can enable it with..."

You're right! I didn't even notice that firewalld already ships with presets for syncthing. I've always configured it myself. Sorry, my mistke. I hope I didn't take up too much of your time. This PR is obsolete I'll close it.

PR #7776, opened 2021-06-17
build: Add codecov upload script to repo

imsodin:
Coverage CI jobs are failing since a while (curl missing - huh?). They still show up as successful, because the failure is within a bash command substitution, and those errors aren't passed on. I think the `bash <(curl something)` is ugly anyway (and has been shown as unsafe recently), so I'd prefer having that script in the repo and manually upgrading it (basically whenever uploading breaks, hopefully almost never).

### DIFF
--- a/script/codecov-upload.sh
+++ b/script/codecov-upload.sh
@@ -1,0 +1,1883 @@
+#!/usr/bin/env bash
+
+# Apache License Version 2.0, January 2004
+# https://github.com/codecov/codecov-bash/blob/master/LICENSE
+
+set -e +o pipefail
+
+VERSION="1.0.3"
+
+codecov_flags=( )
+url="https://codecov.io"
+env="$CODECOV_ENV"
+service=""
+token=""
+search_in=""
+# shellcheck disable=SC2153
+flags="$CODECOV_FLAGS"
+exit_with=0
+curlargs=""
+curlawsargs=""
+dump="0"
+clean="0"
+curl_s="-s"
+name="$CODECOV_NAME"
+include_cov=""
+exclude_cov=""
+ddp="$HOME/Library/Developer/Xcode/DerivedData"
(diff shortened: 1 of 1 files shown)


script/codecov-upload.sh
```diff
@@ -0,0 +1,1883 @@
...
+      if (length($3) > 0) {
+         vn=""; if (indent > 0) {vn=(vn)(vname[0])("_")}
+         printf("%s%s%s=\"%s\"\n", "'"$prefix"'",vn, $2, $3);
+      }
+   }'
+}
+
+if [ $# != 0 ];
+then
+  while getopts "a:A:b:B:cC:dD:e:f:F:g:G:hi:J:k:Kn:p:P:Q:q:r:R:s:S:t:T:u:U:vx:X:Zz:N:-" o
```
calmh, line 263: This could use some comments

script/codecov-upload.sh
```diff
@@ -0,0 +1,1883 @@
...
+else
+  urlencoded_slug=$(urlencode "$slug_o")
+fi
+
+query="branch=$branch\
+       &commit=$commit\
+       &build=$([ "$build_o" = "" ] && echo "$build" || echo "$build_o")\
+       &build_url=$build_url\
+       &name=$(urlencode "$name")\
+       &tag=$([ "$tag_o" = "" ] && echo "$tag" || echo "$tag_o")\
```
calmh, line 1116: Lacks urlencoding

calmh (changes requested):
A bit complex, lacks some tests

calmh (approved):
Just kidding.

imsodin:
I was wondering why you are reviewing so much on holidays, now I am suspecting you might be are experiencing bit of Audrius' holiday boredom :D

PR #7780, opened 2021-06-21
gui: Make listener/discovery modal more discoverable

imsodin:
The only way to discover that there is a modal showing info on the listeners/discovery is to hover them. That's unnecessarily confusing, see e.g. https://forum.syncthing.net/t/android-cant-find-truenas-and-vice-versa/16977/10 . Now they look like normal links (same as e.g. the device id)


AudriusButkevicius (approved):

acolomb:
Note that this change only applies when there is a partial failure.  All methods running (x/x) still displays in green, all failed (0/x) in red.  I'm fine with the change though, just a hint.

PR #7783, opened 2021-06-22
lib/connections: Dial devices in parallel

imsodin:
Improves time to connection/sync in the following scenario:  
Multiple devices are added at the same time through API. One of them happens to be not dialable (firewall), the others are. This devices also happens to be first. Now dialing will take 2x10s (tcp, quic) to time out on the first device until it tries dialing the second one (and immediately succeeds). This leads to an unnecessarily long time to connection (and thus start of sync/transfer).

The main change is straight forward: Dial all devices concurrently. There are some counting shenanigans due to the amount of connections limit. And I added context checks when sending the connection and in the loop.


lib/connections/service.go
```diff
@@ -490,15 +490,43 @@ func (s *service) dialDevices(ctx context.Context, now time.Time, cfg config.Con
...
+				done = numConns >= allowAdditional
+			}
+			dialCond.L.Unlock()
+			if done {
 				break
 			}
 		}
+		numDials++
+		dialWG.Add(1)
+		go func(entry dialQueueEntry) {
```
calmh, line 516: I think we should have a semaphore here with a reasonable limit (32?) so that we don't cause issues or explode our FD usage when dialling thousands of devices

calmh (commented):
Seems fine in principle

lib/connections/service.go
imsodin: Sounds reasonable. Given the FDs apply per dial, not device, I moved the semaphore/limitation there. And to prevent the case that one device has so many addresses to dial that it "clogs up" the semaphore, there's a two step limitation: 8 per device, 64 globally. I reused `model.byteSemaphore` for that which is now `util.Semaphore`, and introduced `util.MultiSemaphore` which combines multiple semaphores the same way `model.newLimitedRequestResponse` already did. That blew up the diff, but really it's just renaming.

lib/connections/service.go
```diff
@@ -490,14 +492,40 @@ func (s *service) dialDevices(ctx context.Context, now time.Time, cfg config.Con
...
+			defer dialWG.Done()
+			conn, ok := s.dialParallel(dialCtx, entry.id, entry.targets, dialSemaphore)
+			if !ok {
+				return
+			}
+			numConnsMut.Lock()
 			if allowAdditional > 0 && numConns >= allowAdditional {
-				break
+				dialCancel()
```
calmh, line 518: this could happen after numConns++ below, as we can determine already then that no further dialing will be allowed to succeed

lib/connections/service.go
```diff
@@ -995,6 +1025,7 @@ func (s *service) dialParallel(ctx context.Context, deviceID protocol.DeviceID,
 					res <- conn
 				}
 				wg.Done()
+				sema.Give(1)
```
calmh, line 1028: Better defer these, having them hanging way down here gives me the heebie-jeebies

calmh: Also, wait, we sema.Take outside the loop but sema.Give inside?

lib/util/semaphore.go
```diff
@@ -0,0 +1,148 @@
+// Copyright (C) 2018 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package util
```
calmh, line 7: (We should split `package util` into `structutil`, `netutil`, `stringsutil`, `syncutil` to regain some semblance of order instead of random-bag-of-crap... but not in this PR...)

lib/connections/service.go
imsodin: We need to check before sending as well, otherwise there's a race between the context and "sending too many connections". I now wrapped sending in a condition and added a condition below numConns++

lib/util/semaphore.go
imsodin: We already have a few of these. Actually, a semaphore might also belong to `lib/sync`? Given I am moving it around in this PR, we might as well move it to an appropriate location now.

calmh (approved):

Issue #7764, opened 2021-06-13
panic: deadlock detected at fmut


tomasz1986:
I have found these panic logs on one of my devices. It seems that Syncthing was able to recover and continued to work on its own, so until now I wasn't even aware that they were occurring.

```
panic: deadlock detected at fmut

goroutine 33 [running]:
github.com/syncthing/syncthing/lib/model.(*deadlockDetector).Watch.func1(0xc00026c890, 0x2926444f128, 0xc00002a330, 0x2009211, 0x4)
	github.com/syncthing/syncthing/lib/model/util.go:61 +0x3a5
created by github.com/syncthing/syncthing/lib/model.(*deadlockDetector).Watch
	github.com/syncthing/syncthing/lib/model/util.go:37 +0xbe

goroutine 1 [chan receive, 95 minutes]:
github.com/syncthing/syncthing/lib/syncthing.(*App).Wait(...)
	github.com/syncthing/syncthing/lib/syncthing/syncthing.go:371
main.syncthingMain(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	github.com/syncthing/syncthing/cmd/syncthing/main.go:721 +0xa2f
main.serveOptions.Run(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	github.com/syncthing/syncthing/cmd/syncthing/main.go:407 +0x3f8
reflect.Value.call(0x1ffd120, 0x28d9600, 0x293, 0x2008e39, 0x4, 0xc000151c38, 0x0, 0x0, 0x146ce34, 0x1ffd120, ...)
	reflect/value.go:476 +0x907
reflect.Value.Call(0x1ffd120, 0x28d9600, 0x293, 0xc000151c38, 0x0, 0x0, 0x1f75320, 0xc000270b90, 0x186)
	reflect/value.go:337 +0xc5
github.com/alecthomas/kong.callMethod(0x200886f, 0x3, 0x1ffd120, 0x28d9600, 0x199, 0x1ffd120, 0x28d9600, 0x293, 0xc00025fe00, 0xc0000b0700, ...)
	github.com/alecthomas/kong@v0.2.16/callbacks.go:71 +0x4ba
github.com/alecthomas/kong.(*Context).RunNode(0xc0000b0700, 0xc00023e1c0, 0x0, 0x0, 0x0, 0x0, 0x0)
[...]

PR #7785, opened 2021-06-23
all: Send deadlocks as failures, crash only as a last resort

imsodin:
https://github.com/syncthing/syncthing/issues/7764

> I used to regularly check f/pmut "deadlocks" but all my samples look like false positives. Actually I think we should make our reaction less painful: What do you think about having the same kind of timeout as now, but it only results in a failure event (which I'd extend to be able to transport goroutine traces), and only after a much longer time do we actually panic. I'd still keep the panic for cases like e.g. introducing a real deadlock in a new RC release, which would otherwise require a manual upgrade (or at least restart) to resolve. My hope is to reduce the amount of false positive lock-panics and thus make it less likely for goleveldb to fail and even if the db is fine, it's also nicer on devices already being slow, if we don't interrupt their slow progress every now and then.

Now after the timeout, we send a failure report. Then we wait for 10 times the timeout, after which we send another failure report, try to shut down and only if that doesn't work (after all, it might indeed be a deadlock), panic.

To make those failure reports actually helpful, I added a goroutine field to failure reports. Those are saved on the crash-receiver server similarly to how crash reports are saved now. I also added a generic string map for extraneous info that might be added to a failure report (used for timeout length in deadlock detection).


calmh:
I'm not convinced we need something this elaborate... Do we get a lot of deadlock failure reports? I can agree on removing the crash given most of the reports are false alarms, instead just keeping failure report plus a warning in the log and leaving it at that... If we did that we could enable it in release builds as well.

imsodin:
There are not many deadlock reports any more, 0-10 per day in the last 14 days.

My motivation for extending the failure reports with goroutines is two-fold:  
I wanted goroutine traces for unclean shutdowns (services or database doesn't terminate within timeout) - there may be more instances. And now deadlock reports without goroutine traces would be pretty much useless. I did find a lot of locking improvements through deadlock reports in the past. Granted, there haven't been any useful reports in quite a while.

And independent of that I'd like to retain the crashing in candidate releases, in case we introduce a new deadlock: Without crashing the user needs to manually intervene to get unstuck, instead of Syncthing upgrading itself after the crash.

So in my opinion there's two options here:

1. The elaborate one with goroutines, i.e. PR status quo.
2. Keep crashing like we do now, just increase the timeout (because I really don't think we should crash after only 20min).

Opinions? Am I missing a "middle-ground" step here?

lib/model/util.go
```diff
@@ -13,57 +13,95 @@ import (
...
-						msg += "\n===" + otherName + "===\n" + otherHolder.Holders()
-					}
+			select {
+			case <-warn.C:
+				failure := ur.FailureDataWithGoroutines(fmt.Sprintf("potential deadlock detected at %s (short timeout)", name))
+				failure.Extra["timeout"] = d.warnTimeout.String()
+				d.evLogger.Log(events.Failure, failure)
+			case <-done:
+				warn.Stop()
+				fatal.Stop()
```
calmh, line 66: Maybe make the inner loop a method so we can defer these on creation?

lib/model/util.go
```diff
@@ -13,57 +13,95 @@ import (
...
+				failure := ur.FailureDataWithGoroutines(err.Error())
+				failure.Extra["timeout"] = d.fatalTimeout.String()
+				var others string
+				for otherName, otherHolder := range d.otherHolders() {
+					others += "===" + otherName + "===\n" + otherHolder + "\n"
 				}
-				panic(msg)
+				others = others[:len(others)-1]
+				failure.Extra["other-holders"] = others
+				d.otherHolders()
```
calmh, line 81: This is called again for some side effect...?

lib/model/util.go
```diff
@@ -13,57 +13,95 @@ import (
...
+			case <-fatal.C:
+				err := fmt.Errorf("potential deadlock detected at %s (long timeout)", name)
+				failure := ur.FailureDataWithGoroutines(err.Error())
+				failure.Extra["timeout"] = d.fatalTimeout.String()
+				var others string
+				for otherName, otherHolder := range d.otherHolders() {
+					others += "===" + otherName + "===\n" + otherHolder + "\n"
 				}
-				panic(msg)
+				others = others[:len(others)-1]
```
calmh, line 79: What are we cutting off the end here?

lib/model/util.go
```diff
@@ -13,57 +13,95 @@ import (
...
+				time.Sleep(time.Minute)
+				panic(fmt.Sprintf("%v:\n%v", err, others))
+			case <-done:
+				fatal.Stop()
 			}
 		}
 	}()
 }
 
+func (d *deadlockDetector) otherHolders() map[string]string {
```
calmh, line 95: Looks like this might as well return a single formatted string if that's how we always treat the resulting map

calmh (commented):
I buy the argument for the elaborateness

lib/model/util.go
imsodin: An empty line.

calmh (approved):

PR #7790, opened 2021-06-25
lib/protocol: Write uncompressible messages uncompressed

greatroar:
When a message becomes bigger after compression, the uncompressed version can be sent instead to not waste bandwidth and receiver CPU time. I've been keeping track of compressed message sizes and I've found it extremely common that metadata messages become larger from compression; it's only the longer ones that actually benefit (so maybe the threshold should be increased too).

I've merged the various send*Message methods so that there's less copying of compressed messages and less reallocation when compression fails. The cost of that is having to serialize the header of the uncompressed version unconditionally.

### DIFF
--- a/lib/protocol/protocol.go
+++ b/lib/protocol/protocol.go
@@ -523,7 +523,7 @@ func (c *rawConnection) readMessageAfterHeader(hdr Header, fourByteBuf []byte) (
 		// Nothing
 
 	case MessageCompressionLZ4:
-		decomp, err := c.lz4Decompress(buf)
+		decomp, err := lz4Decompress(buf)
 		BufferPool.Put(buf)
 		if err != nil {
 			return nil, errors.Wrap(err, "decompressing message")
@@ -754,7 +754,7 @@ func (c *rawConnection) writeCompressedMessage(msg message) error {
 		return errors.Wrap(err, "marshalling message")
 	}
 
-	compressed, err := c.lz4Compress(buf)
+	compressed, err := lz4Compress(buf)
 	if err != nil {
 		BufferPool.Put(buf)
 		return errors.Wrap(err, "compressing message")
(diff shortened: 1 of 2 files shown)


lib/protocol/protocol.go
```diff
@@ -823,26 +753,63 @@ func (c *rawConnection) writeUncompressedMessage(msg message) error {
...
+
+		// Header length
+		binary.BigEndian.PutUint16(buf, uint16(hdrSize))
+		// Header
+		if _, err := hdr.MarshalTo(buf[2:]); err != nil {
+			return errors.Wrap(err, "marshalling header")
+		}
+		// Message length
+		binary.BigEndian.PutUint32(buf[2+chdrSize:], uint32(compressedSize))
+	}
```
imsodin, line 803: I think it would be simpler/cleaner if this code would stay in a separate function, e.g. `compressedMessage(msg message, uncompressedSize int) ([]byte, bool, error)` that returns false if compressing increases the size. Obviously it's entirely possible I am missing something that makes this annoying/ugly.

lib/protocol/protocol.go
```diff
@@ -823,26 +753,63 @@ func (c *rawConnection) writeUncompressedMessage(msg message) error {
...
-		BufferPool.Put(buf)
 		return errors.Wrap(err, "marshalling header")
 	}
 	// Message length
 	binary.BigEndian.PutUint32(buf[2+hdrSize:], uint32(size))
 	// Message
 	if _, err := msg.MarshalTo(buf[2+hdrSize+4:]); err != nil {
-		BufferPool.Put(buf)
 		return errors.Wrap(err, "marshalling message")
 	}
```
imsodin, line 769: This part is has no effect/is unneccessary if the message is sent compressed.

greatroar: It does (except for header marshaling), because this determines what needs to be compressed. The original code also marshaled twice in the compressed case.

lib/protocol/protocol.go
greatroar: That means messages may have to be marshalled three times: once to know what needs to be compressed, once compressed, then again to send the uncompressed message.

I can try to separate the compression stuff into a separate method, though.

greatroar: Please see the updated PR.

imsodin (approved):

PR #7791, opened 2021-06-25
lib/model: Compare all items with global on scan (fixes #7740)

### LINKED ISSUE #7740: Incorrect local state when using negated patterns inside ignored parent folder on both sides
I have finally managed to reproduce https://forum.syncthing.net/t/folder-stuck-in-sync-and-non-matching-local-and-global-states/16431.

1. On Device 1, create a folder called `test1`.
2. In the folder, create a folder `test`, and then inside it another folder `test`.
3. The folder structure now looks like `test1\test\test`.
4. Add the `test1` folder to Syncthing with the following ignore patterns.
   ```
   !/*/
   *
   ```
   As expected, both the subfolder and its parent folder have been indexed.
   
   image

5. Share the folder with Device 2, and accept it there with the same ignore patterns as above.
6. The folder state on Device 2 is correct, but the local state on Device 1 has changed. The parent folder has been ignored, despite its child being unignored.

   image
   image

Logs with `model` and `db`:

[syncthing1.log](https://github.com/syncthing/syncthing/files/6589230/syncthing1.log)
[syncthing2.log](https://github.com/syncthing/syncthing/files/6589231/syncthing2.log)

One thing to note is that the folder states are correct when not using the same ignore patterns on the other side. It seems that it is only when the same negate/ignore patterns are applied to both sides that the issue occurs.

imsodin:
First two commits are the same as on https://github.com/syncthing/syncthing/pull/7726: Replaces the `batchAppendFunc` being passed around as a function argument with a dedicated `scanBatch` type that has an `append` method.

The main change fixing #7740 is, that we now compare every new, scanned file with the current global file. If they are equivalent, the global file info is used, thus saving the detour over syncing to resolve the conflict (that isn't actually a conflict). This also benefits initial scan scenarios, where both sides have the data: If files are scanned after the index is received from the remote, they will immediately be in sync (instead of looking out of sync until all of the scan is done and a sync happens to resolve the conflicts).

### DIFF
--- a/lib/model/folder.go
+++ b/lib/model/folder.go
@@ -541,46 +541,33 @@ func (f *folder) newScanBatch() *scanBatch {
 // Append adds the fileinfo to the batch for updating, and does a few checks.
 // It returns false if the checks result in the file not going to be updated or removed.
 func (b *scanBatch) Append(fi protocol.FileInfo, snap *db.Snapshot) bool {
-	// Resolve items which are identical with the global state.
-	switch b.f.Type {
-	case config.FolderTypeReceiveOnly:
-		switch gf, ok := snap.GetGlobal(fi.Name); {
-		case !ok:
-		case gf.IsEquivalentOptional(fi, b.f.modTimeWindow, false, false, protocol.FlagLocalReceiveOnly):
-			// What we have locally is equivalent to the global file.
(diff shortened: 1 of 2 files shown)


calmh (approved):
Cool

PR #7794, opened 2021-06-26
lib/fs: Set expiry after DirNames in case-fs

imsodin:
Slightly refactors the case-filesystem with the main goal of setting the expiry after calling `DirNames`. As that can take a long time for large directories and slow systems (see https://github.com/syncthing/syncthing-android/issues/1630). As the expiry needs to be set while holding the cache lock, I moved the entire node creation including calling `DirNames` into `newCaseNode` - thus the `sync.Once` isn't needed any more.

### DIFF
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -386,7 +386,6 @@ func (f *caseFilesystem) checkCaseExisting(name string) error {
 }
 
 type defaultRealCaser struct {
-	fs    Filesystem
 	cache caseCache
 }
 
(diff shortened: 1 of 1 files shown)


Catfriend1:
Hi,

is this ready to be cherry picked and tested on Android?

imsodin:
Not sure, I don't know what the criterion for that might be. The state is like any PR: I think it's a good, and working change (that's why I PRed), it passes tests (see CI) and it hasn't been reviewed.

auanasgheps:
Would this improve the situation on Android 11?

calmh (approved):

PR #7798, opened 2021-06-28
lib/connections: Fully dial resumed devices

imsodin:
This improves/extends the mechanism to dial changed devices introduced in https://github.com/syncthing/syncthing/pull/7604: So far it covers newly added devices and addresses. It does not cover resumed devices, as they are still subject redial delays (e.g. when quickly pausing/unpausing a device to trigger a reconnect). Now new or previously paused are added to `dialNowDevices`, thus not subject to any delays.

### DIFF
--- a/lib/connections/service.go
+++ b/lib/connections/service.go
@@ -786,13 +786,24 @@ func (s *service) CommitConfiguration(from, to config.Configuration) bool {
 
 func (s *service) checkAndSignalConnectLoopOnUpdatedDevices(from, to config.Configuration) {
 	oldDevices := from.DeviceMap()
+	dial := false
+	s.dialNowDevicesMut.Lock()
 	for _, dev := range to.Devices {
+		if dev.Paused {
+			continue
+		}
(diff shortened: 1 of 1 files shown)


lib/connections/service.go
```diff
@@ -786,13 +786,24 @@ func (s *service) CommitConfiguration(from, to config.Configuration) bool {
...
+		}
 		oldDev, ok := oldDevices[dev.DeviceID]
-		if !ok || !util.EqualStrings(oldDev.Addresses, dev.Addresses) {
-			s.scheduleDialNow()
-			break
+		if !ok || oldDev.Paused || !util.EqualStrings(oldDev.Addresses, dev.Addresses) {
+			if !ok || oldDev.Paused {
+				s.dialNowDevices[dev.DeviceID] = struct{}{}
+			}
+			dial = true
```
calmh, line 800: This nesting is ugly and looks like a mistake... Maybe flatten it to two separate branches, a second one just for the equalstrings?

calmh (approved):

PR #7810, opened 2021-07-05
lib/connections: Resolve IPv6 for quic6:// peers

### LINKED ISSUE #7809: Syncthing attemps IPv4 addresses for quic6:// peers

Environment: syncthing 1.17.0 installed from [Arch Linux official package](https://archlinux.org/packages/community/x86_64/syncthing/)

Any steps to reproduce the problem:

* Create a DNS server that gives both an IPv4 address (e.g., `192.168.xxx.yyy`) and an IPv6 address (e.g., `fd0a:b:c:d::xxx:yyy`) for the internal domain name `server.internal`.
* Use the domain name as the peer address (e.g., `quic6://server.internal:22000`) in Syncthing

What happened:

Syncthing attempts IPv4-compatible addresses (e.g., `[::ffff:192.168.xxx.yyy]:22000`) for the peer as per `strace`.

```
sendto(13, <some long data>, 1252, 0, {sa_family=AF_INET6, sin6_port=htons(22000), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "::ffff:192.168.xxx.yyy", &sin6_addr), sin6_scope_id=0}, 28
```

What you expected to happen instead:

Syncthing connect to the normal IPv6 address (e.g., `[fd0a:b:c:d::xxx:yyy]:22000`)

Possible fix:

I noticed that in `quic_dial.go`, the network "udp" is used for resolving addresses [1]. On the other hand, in `quic_listen.go`, the proper network is used based on the URI scheme [2]. If I patch syncthing use the same approach in `quic_dial.go`, everything works fine.

```Diff
diff --git a/lib/connections/quic_dial.go b/lib/connections/quic_dial.go
index 83c6246f5..a952d1ecd 100644
--- a/lib/connections/quic_dial.go
+++ b/lib/connections/quic_dial.go
@@ -13,6 +13,7 @@ import (
 	"crypto/tls"
 	"net"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/lucas-clemente/quic-go"
@@ -45,7 +46,9 @@ type quicDialer struct {
 func (d *quicDialer) Dial(ctx context.Context, _ protocol.DeviceID, uri *url.URL) (internalConn, error) {
 	uri = fixupPort(uri, config.DefaultQUICPort)
 
-	addr, err := net.ResolveUDPAddr("udp", uri.Host)
+	network := strings.ReplaceAll(uri.Scheme, "quic", "udp")
+
+	addr, err := net.ResolveUDPAddr(network, uri.Host)
 	if err != nil {
 		return internalConn{}, err
 	}
```

[1] https://github.com/syncthing/syncthing/blob/1921533c4cae1873d3cd63a28a590075ac7a48fc/lib/connections/quic_dial.go#L48
[2] https://github.com/syncthing/syncthing/blob/1921533c4cae1873d3cd63a28a590075ac7a48fc/lib/connections/quic_listen.go#L84-L86

yan12125:
### Purpose

Before this patch, IPv4-compatible addresses (`::ffff:aaa.bbb.ccc.ddd`)
may be used if a `quic6://some.domain:port` is specified and both IPv4 and
IPv6 addresses exist for that domain name.

Fixes #7809

### Testing

Follow "steps to reproduce the problem" as described in #7809, and check if I got connections with normal IPv6 addresses.

## Authorship

based on the commit metadata.

### DIFF
--- a/lib/connections/quic_dial.go
+++ b/lib/connections/quic_dial.go
@@ -13,6 +13,7 @@ import (
 	"crypto/tls"
 	"net"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/lucas-clemente/quic-go"
@@ -45,7 +46,9 @@ type quicDialer struct {
 func (d *quicDialer) Dial(ctx context.Context, _ protocol.DeviceID, uri *url.URL) (internalConn, error) {
 	uri = fixupPort(uri, config.DefaultQUICPort)
 
-	addr, err := net.ResolveUDPAddr("udp", uri.Host)
+	network := strings.ReplaceAll(uri.Scheme, "quic", "udp")
+
+	addr, err := net.ResolveUDPAddr(network, uri.Host)
 	if err != nil {
 		return internalConn{}, err
 	}


acolomb (commented):
I don't know how the Go library handles it in the rest of the code, but at least here we should treat the search string case-insensitive. URI parsers must accept any case variation on the scheme part. Or is it already normalized in the `uri` data structure?

yan12125:
I didn't find documentation about it. From Go source codes, the scheme is indeed normalized to lower cases: https://github.com/golang/go/blob/go1.16.5/src/net/url/url.go#L531

AudriusButkevicius (approved):
lgtm

acolomb (approved):

yan12125:
Thanks for reviewing!

PR #7819, opened 2021-07-13
lib/relay/client: Stricter typing and remove unused code

greatroar:
In particular, this makes it impossible to accidentally stack dynamicClients.

commonClient.invitations was protected by a mutex, but never set after construction. I moved that mutex; I also changed it to a vanilla stdlib mutex because its use is so trivial, but I can of course change that back if you want.

### DIFF
--- a/lib/relay/client/client.go
+++ b/lib/relay/client/client.go
@@ -11,58 +11,43 @@ import (
 
 	"github.com/syncthing/syncthing/lib/relay/protocol"
 	"github.com/syncthing/syncthing/lib/svcutil"
-	"github.com/syncthing/syncthing/lib/sync"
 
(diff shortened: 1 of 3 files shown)


imsodin (approved):
Looks good generally. I don't understand how removing the factory prevents stacking dynamic clients respectively I don't see how that could have happened before already.

greatroar:
There's no code that creates nested dynamicClients, but it was structurally possible because dynamicClient.client was a RelayClient (interface type). Nested dynamicClients are now prohibited statically.

imsodin:
Ah ok right, that's not related to removing the "explicit" factory, and moving the pattern into `NewClient` - that change is probably just about simplicity/"taste", possibly bit less code then.

PR #7829, opened 2021-07-19
lib/fs: Improve case conflict error message (fixes #7827)

### LINKED ISSUE #7827: The error message "given name ... differs from filesystem name ..." does not help users to resolve the issue
I am running Syncthing on Android (1.18.0) and on Arch Linux (1.17.1) to sync my smartphone's storage to my notebook. On my smartphone, I get a lot of messages like

```
2021-07-18 10:27:29 Puller (folder "Speicher" (t66yq-r9o4p), item "Music/Einzel/The King's Singers/This Is The King's Singers"): syncing: checking file to be replaced: given name "Music/Einzel/The King's Singers/This Is The King's Singers" differs from name in filesystem "Music/Einzel/The King's Singers/This Is the King's Singers"
```

I don't know what to do with this information. What is the "given name"? How can a file have a different name than its file system name? Can this error message be reformulated to something that tells me as a user how to resolve this?

tomasz1986:
Use non-technical language in the message to make clear what exactly
the error is talking about. Also, mention a possible solution on how to
fix the problem.

### DIFF
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -29,7 +29,7 @@ type ErrCaseConflict struct {
 }
 
 func (e *ErrCaseConflict) Error() string {
-	return fmt.Sprintf(`remote "%v" uses different upper or lowercase than local "%v"; change the casing on either side to fix`, e.Given, e.Real)
+	return fmt.Sprintf(`remote "%v" uses different upper or lowercase characters than local "%v"; change the casing on either side to fix`, e.Given, e.Real)
 }
 
 func IsErrCaseConflict(err error) bool {
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -29,7 +29,7 @@ type ErrCaseConflict struct {
 }
 
 func (e *ErrCaseConflict) Error() string {
-	return fmt.Sprintf(`remote "%v" is spelled in a different case than local "%v"; use identical spelling on both sides to fix`, e.Given, e.Real)
+	return fmt.Sprintf(`remote "%v" uses different upper or lowercase than local "%v"; change the casing on either side to fix`, e.Given, e.Real)
 }
 
 func IsErrCaseConflict(err error) bool {
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -29,7 +29,7 @@ type ErrCaseConflict struct {
 }
 
 func (e *ErrCaseConflict) Error() string {
-	return fmt.Sprintf(`given name "%v" differs from name in filesystem "%v"`, e.Given, e.Real)
+	return fmt.Sprintf(`remote "%v" is spelled in a different case than local "%v"; use identical spelling on both sides to fix`, e.Given, e.Real)
 }
 
 func IsErrCaseConflict(err error) bool {
--- a/lib/fs/casefs.go
+++ b/lib/fs/casefs.go
@@ -29,7 +29,7 @@ type ErrCaseConflict struct {
 }
 
 func (e *ErrCaseConflict) Error() string {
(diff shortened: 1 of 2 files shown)


rasa:
@tomasz1986 Can you correct the [test](https://github.com/syncthing/syncthing/blob/df482763001435aa5050d983e29f2ac2efceb2aa/lib/model/folder_sendrecv_test.go#L1207) too, so the build passes?

lib/fs/casefs.go
```diff
@@ -29,7 +29,7 @@ type ErrCaseConflict struct {
 }
 
 func (e *ErrCaseConflict) Error() string {
-	return fmt.Sprintf(`given name "%v" differs from name in filesystem "%v"`, e.Given, e.Real)
+	return fmt.Sprintf(`remote "%v" is spelled in a different case than local "%v"; use identical spelling on both sides to fix`, e.Given, e.Real)
```
calmh, line 32:
```suggestion
	return fmt.Sprintf(`remote "%v" uses a different upper-/lowercase variant than local "%v"; use identical casing on both sides to fix`, e.Given, e.Real)
```

(This might be my english-as-secondary-language personality, but "spelling" doesn't sound like what I'd consider a case difference... Someone with a primary language that uses case as a significant language feature might have another opinion...)

tomasz1986: Well, I took the text from https://github.com/syncthing/syncthing/issues/7827#issuecomment-882494860. The idea seems to be to use layman's terms instead of more technical language. "Spelling" does seem to be used in the context of case, if you look it up in the Internet (e.g. https://www.ibm.com/support/pages/node/4931943, "The HATS Drop-down widget was matching multiple values in the list if the values were the same but were spelled using a different case."), so I'm not sure about this one. I'm not a native speaker either though.

rasa: As a native speaker, I think
`remote "%v" uses a different upper-/lowercase variant than local "%v"; use identical casing on both sides to fix`
is fine, but perhaps we could be a little more explicit:
`remote "%v" uses a different upper-/lowercase variant than local "%v"; change the casing on either side to fix`
?

tomasz1986: I have reworded and shortened the message slightly. What do you think about it now?

rasa: `uses different upper or lowercase` sounds mildly incomplete, like the word `characters` should be added. 
Without wanting to be pedantic, how about:
`remote "%v" differs only by case from local "%v"; change the casing of either file name to fix`
?

tomasz1986: I have added "characters" 😉. I actually like it a lot, as it gives more context to the user, especially a non-English speaking one to know what to look for. I have left the rest unchanged. I don't want to add "file" in there, as the problem may be about a folder, not a file.

lib/fs/casefs.go
```diff
@@ -29,7 +29,7 @@ type ErrCaseConflict struct {
 }
 
 func (e *ErrCaseConflict) Error() string {
-	return fmt.Sprintf(`given name "%v" differs from name in filesystem "%v"`, e.Given, e.Real)
+	return fmt.Sprintf(`remote "%v" uses different upper or lowercase characters than local "%v"; change the casing on either side to fix`, e.Given, e.Real)
```
imsodin, line 32: While pedantic, it currently doesn't say that you have to change the casing such that the two are equal. Should be obvious, but if it can be helped, nothing should be considered obvious :) 

```suggestion
	return fmt.Sprintf(`remote "%v" uses different upper or lowercase characters than local "%v"; change the casing on either side to match the other`, e.Given, e.Real)
```

tomasz1986: Fixed 👍.

imsodin (approved):

PR #7840, opened 2021-07-22
cmd/stcrashreceiver: Sanitize failure report fingerprints

imsodin:
Sentry (and my mailbox) is getting flooded with database corruption events because they can since a while occur as failures instead of panics, and the fingerprint isn't sanitized on failures. Now it is.


calmh (approved):

calmh:
(needs manual deploy)

imsodin:
done

PR #7843, opened 2021-07-22
lib/model: Don't try to delete deleted item on revert

imsodin:
Reported here by @tomasz1986. We try to delete a file that doesn't exist, which unsurprisingly fails.


calmh (approved):

tomasz1986:
> Reported here by @tomasz1986.

Just for the sake of completion (because I think that a link was supposed to be included there?), the issue was reported on the forum in https://forum.syncthing.net/t/irreversible-local-additions-the-system-cannot-find-the-path-specified/17120.

PR #7847, opened 2021-07-23
gui: Improve revert/override/delete warning modals

tomasz1986:
1. Change each modal title text to match the action that is being
   executed (i.e. "Revert" to "Revert Local Additions", "Override" to
   "Override Changes", "Delete" to "Delete Unexpected Items").
2. Change the icons to match the icons used by each action (i.e. arrow-
   circle-down for Revert, arrow-circle-up for Override). Replace the
   broken lock icon for Delete with minus-circle.
3. Rearrange the order in the modal HTML code to simplify it a little.
4. Use red text for warnings, following the warning present in the
   Advanced Configuration.

### Screenshots

Before (Revert Local Changes):

image

After (Revert Local Changes):

image

Before (Override Changes):

image

After (Override Changes):

image

Before (Delete Unexpected Items):

image

After (Delete Unexpected Items):

image


tomasz1986:
On a side note, if this PR is accepted, then the "Revert" and "Override" strings added to translate in https://github.com/syncthing/syncthing/commit/7ec76095e6abd6c84b60c8cfe1f814188f1f9777 will probably be no longer necessary.

imsodin (commented):
Personally I don't like the red text, I find it less readable. And I'd say the red title and button should get the danger message across.

And how did the reordering make things simpler? The `default` case being first is unexpected to me.

AudriusButkevicius:
Agree on the red text being ugly.

tomasz1986:
> Personally I don't like the red text, I find it less readable. And I'd say the red title and button should get the danger message across.

> Agree on the red text being ugly.

Well, I only followed the example of Advanced Configuration.

image

Is this one less readable and/or ugly too? I can change the text to black, although I personally find the red one better (when viewed in the actual browser, as screenshots may not do it enough justice).

> And how did the reordering make things simpler? The `default` case being first is unexpected to me.

If I remember correctly, I wanted to reuse the "Warning!" message on the top for all the dialogues. Also, I personally found the `default` case not being the first unexpected. Why should the default not be the first case though?

calmh:
One thing I would like in order to make these clearer, to the extent that's possible, is to make the buttons represent the taken action -- that is, `Revert` and `Cancel` instead of `Yes` and `No` for example.

As for the red text, yeah, no, we don't do that. We do it on the advanced config dialog specifically because it's not really a warning dialog by itself and to call out that part of it as a warning.

tomasz1986:
@imsodin 
> And how did the reordering make things simpler? The `default` case being first is unexpected to me.

I have read up the Angular documentation on the `ng-switch`, and of course you were right about the order. I have fixed the code to follow the standard order, and also tried to clearly split the three cases - `deleteEnc`, `override`, and `default` (revert).

@calmh 
> One thing I would like in order to make these clearer, to the extent that's possible, is to make the buttons represent the taken action -- that is, `Revert Changes` and `Cancel` instead of `Yes` and `No` for example.

I have changed the final questions to be very specific on what the consequences are. The button text has also been changed. I have tried not to touch the already used strings, so that the existing translations don't need to be redone.

> As for the red text, yeah, no, we don't do that.

The red text is gone now 😉.

---

What do you think of these?

image
image
image

calmh:
Seems fine to me

imsodin (approved):

PR #7848, opened 2021-07-23
gui: Move dismiss button after add/ignore buttons

tomasz1986:
Currently, the dismiss button is displayed as the first of the three
buttons. However, the most common action that the user wants to do when
sharing a new folder is to add it on a different device. Because of
this, the add button should be displayed first as the most prominent of
the three. The ignore button is the opposite of the add button, and also
results in a parmenent action, hence it makes sense to lump the two
together. Thus, the dismiss button should be moved to the last place as
an alternative to the two main actions, when the user is yet unsure what
they want to do with the notification.

### Screenshots

Before:

image

After:

image

### DIFF
--- a/gui/default/index.html
+++ b/gui/default/index.html
@@ -209,14 +209,14 @@
               </div>
               <div class="panel-footer clearfix">
                 <div class="pull-right">
-                  <button type="button" class="btn btn-sm btn-default" ng-click="dismissPendingDevice(deviceID)" tooltip data-original-title="{{'Do not add it to the ignore list, so this notification may recur.' | translate}}">
-                    <span class="far fa-clock"></span>&nbsp;<span translate>Dismiss</span>
-                  </button>
                   <button type="button" class="btn btn-sm btn-success" ng-click="addDevice(deviceID, pendingDevice.name)">
                     <span class="fas fa-plus"></span>&nbsp;<span translate>Add Device</span>
                   </button>
                   <button type="button" class="btn btn-sm btn-danger" ng-click="ignoreDevice(deviceID, pendingDevice)" tooltip data-original-title="{{'Permanently add it to the ignore list, suppressing further notifications.' | translate}}">
                     <span class="fas fa-times"></span>&nbsp;<span translate>Ignore</span>
+                  </button>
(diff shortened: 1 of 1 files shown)


acolomb:
I originally placed the Dismiss button on the far left on its own, but that idea was not received well at the time. I still think it would be a better solution than just shuffling them around here.

tomasz1986:
That is an alternative, yes 🙂. I personally do like them lumped together as related, and also to avoid unnecessary mouse movements, since these pop-ups get very wide on larger screens.

Also, there can be different opinions on the order of the buttons. I have managed to find the following guidelines by Apple.

> **Position buttons as expected.** Any buttons in the bottom right of a dialog should dismiss the dialog. An action button, which initiates the dialog’s primary action, should be farthest to the right. A Cancel button should be to the immediate left of the action button. If a third dismissal button exists, it should be to the left of the Cancel button. A Help button should be farthest to the left. A button that affects the contents of the dialog itself, such as a button for expanding the dialog to reveal additional details, should be to the right of the Help button, or farthest to the left if no Help button is present.

source: https://developer.apple.com/design/human-interface-guidelines/macos/windows-and-views/dialogs

They propose the order of 'help-dismiss-cancel-action', however here in Syncthing the standard button order is 'action-cancel', which is the opposite of the Apple's one, so even if we were to follow such guidelines, I think that it makes sense to place 'dismiss' to the right of 'cancel' (in reverse to the Apple's order).

Regardless of anything else though, I have been adding a lot of folders here and there recently, and in my personal opinion, having the 'dismiss' button right in the first place just doesn't feel right, at least in my eyes 😉, hence the PR.

AudriusButkevicius (approved):
Personally, I prefer them clumped together (less mouse movement) and I prefer Add being the first button, but it's just a personal preference. I'm leaving it to others to shoot this down.

bugith:
BTW, about this nice new feature ... please make the button to clear the notification as intended ;-)

calmh:
LGTM. If any button should be moved out to the far corner that'd usually be the "danger" one. (Which is probably "ignore" in this specific case?)

PR #2478, opened 2015-11-16
Warn the user if they're running with an insecure looking setup (fixes #2139)

calmh:
This shows an angry warning in the UI when the GUI is configured to listen on something other than localhost and there is no auth configured. It can be disabled by setting the hidden `insecureAdminAccess` setting to true, at which point we instead emit a warning at startup (which also pops up in the GUI, but can be dismissed).

screen shot 2015-11-16 at 21 41 09

PR #7852, opened 2021-07-27
lib/syncthing: Remove the warning when insecureAdminAccess is enabled

ncfavier:
### Purpose

My use case is having Syncthing listen on a private WireGuard address. It's annoying to have this warning in my face even after explicitly enabling the setting that starts with "insecure".

See https://github.com/syncthing/syncthing/pull/2478#discussion_r45117737 for prior discussion.

### Testing

Trivial change, no testing done.

### Documentation

https://github.com/syncthing/docs/pull/656

### DIFF
--- a/lib/syncthing/syncthing.go
+++ b/lib/syncthing/syncthing.go
@@ -414,10 +414,6 @@ func (a *App) setupGUI(m model.Model, defaultSub, diskSub events.BufferedSubscri
 		return nil
 	}
 
-	if guiCfg.InsecureAdminAccess {
-		l.Warnln("Insecure admin access is enabled.")
-	}
-
 	summaryService := model.NewFolderSummaryService(a.cfg, m, a.myID, a.evLogger)
 	a.mainService.Add(summaryService)
 
--- a/man/syncthing-config.5
+++ b/man/syncthing-config.5
@@ -750,7 +750,7 @@ If set, this is the API key that enables usage of the REST interface.
 .TP
 .B insecureAdminAccess
 If true, this allows access to the web GUI from outside (i.e. not localhost)
-without authorization. A warning will displayed about this setting on startup.
+without authorization.
 .TP
 .B theme
 The name of the theme to use.


AudriusButkevicius:
I don't see how this is addressing the concerns for why this was added in the first place.

 It addresses your use case, which is great, but potentially leaves many users with insecure setup and no warning, so I don't think this should be merged.

ncfavier:
Remember that there are two levels to this: when `insecureAdminAccess` is disabled (and the admin access is insecure), users get the big red "Danger!" box. Then, if they decide to enable `insecureAdminAccess`, they get a second, dismissable warning instead.

This PR removes the second warning, not the first.

calmh:
This boils down to a matter of opinion; my opinion is that the warning fulfils a useful purpose and should stay.

ncfavier:
Then, as for all warnings, there should at least be an option to disable it, say `reallyInsecureAdminAccess`. But that sounds pretty insecure, so we'd better warn the user about it. But then we need another option to disable that warning, let's name it `yesIAmSureIWantToEnableInsecureAdminAccess`. [...]

Jokes aside, what "useful purpose" do you have in mind?

tomasz1986:
> Jokes aside, what "useful purpose" do you have in mind?

This is the same case as with the warning about running as root, and the one about synchronising Synching's own config files. I personally do agree that these warnings get annoying very quickly if you are forced to see them on every start of the application. I doubt that you will succeed in persuading the developers to add additional options in order to silence them though :wink:.

If you don't want to bother with modifying the code and compiling a custom build, then I would suggest to just use a custom theme with CSS that will hide them in the Web GUI.

ncfavier:
> This is the same case as with the warning about running as root, and the one about synchronising Synching's own config files.

The difference is that those warnings don't occur after the user has enabled an advanced configuration option basically saying "I know what I'm doing", which is the case here.

AudriusButkevicius:
I think having a special env var to suppress these might be an option?

It doesn't bloat the config and is still usable by computer savvy people.

PR #7854, opened 2021-07-28
Fix config deadlock when deleting a paused folder

yyogo:
### Purpose

This PR fixes a bug which causes the config service to deadlock when trying to delete a folder which has been paused for over a minute. The API call block and future attempts to modify the config will block without ever succeeding. This was caused by two bugs  in the `indexHandler` routine:

* it was blocking while waiting for the resume signal, so it could not handle the context being cancelled
* the cond was not signaled when resuming, so even resuming before deletion would not fix the issue

This PR fixes these bugs by waiting for the resume signal asynchronously, while also checking if the context is cancelled.

### Testing

**Follow these steps to reproduce the bug or verify the fix:**

1. Have a connected device with a shared folder `x`
2. Pause the folder:
   `curl -d '{"paused":true}' -X PATCH -H "X-API-Key: apikey" 'http://localhost:8384/rest/config/folders/x'`
3. Wait 60 seconds
4. (optional) resume the folder:
    `curl -d '{"paused":false}' -X PATCH -H "X-API-Key: apikey" 'http://localhost:8384/rest/config/folders/x'`
5. Remove  the folder using:
    `curl -X DELETE -H "X-API-Key: apikey" 'http://localhost:8384/rest/config/folders/x'`

**Before fix is applied:** The API call blocked perpetually. Future attempts to modify the config also block.
**After fix is applied:** The API call returns successfully and the folder is deleted. Future attempts to modify the config return succesfully

### DIFF
--- a/lib/model/indexhandler.go
+++ b/lib/model/indexhandler.go
@@ -111,20 +111,42 @@ func newIndexHandler(conn protocol.Connection, downloads *deviceDownloadState, f
 	}
 }
 
-func (s *indexHandler) Serve(ctx context.Context) (err error) {
-	l.Debugf("Starting index handler for %s to %s at %s (slv=%d)", s.folder, s.conn.ID(), s.conn, s.prevSequence)
-	defer func() {
-		err = svcutil.NoRestartErr(err)
-		l.Debugf("Exiting index handler for %s to %s at %s: %v", s.folder, s.conn.ID(), s.conn, err)
-	}()
-
-	// We need to send one index, regardless of whether there is something to send or not
+// waitForFileset waits for the handler to resume and fetches the current fileset.
+func (s *indexHandler) waitForFileset(ctx context.Context) *db.FileSet {
 	s.cond.L.Lock()
-	for s.paused {
+	defer s.cond.L.Unlock()
+
+	for s.paused && ctx.Err() == nil {
 		s.cond.Wait()
 	}
 	fset := s.fset
-	s.cond.L.Unlock()
+
+	return fset
+}
+
+func (s *indexHandler) Serve(ctx context.Context) (err error) {
+	l.Debugf("Starting index handler for %s to %s at %s (slv=%d)", s.folder, s.conn.ID(), s.conn, s.prevSequence)
+	stop := make(chan struct{})
+
+	defer func() {
+		err = svcutil.NoRestartErr(err)
+		l.Debugf("Exiting index handler for %s to %s at %s: %v", s.folder, s.conn.ID(), s.conn, err)
+		close(stop)
+	}()
+
+	// Broadcast the pause cond when the context quits
+	go func() {
+		select {
+		case <-ctx.Done():
+			s.cond.L.Lock()
+			s.cond.Broadcast()
+			s.cond.L.Unlock()
+		case <-stop:
+		}
+	}()
+
+	// We need to send one index, regardless of whether there is something to send or not
+	fset := s.waitForFileset(ctx)
 	err = s.sendIndexTo(ctx, fset)
 
 	// Subscribe to LocalIndexUpdated (we have new information to send) and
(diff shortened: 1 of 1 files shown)


lib/model/indexhandler.go
```diff
@@ -111,6 +111,42 @@ func newIndexHandler(conn protocol.Connection, downloads *deviceDownloadState, f
 	}
 }
 
+// waitCondWithContext waits for cond to signal, or for the context to finish.
+// If the context is cancelled, the cond will broadcast and an error will be returned.
+// cond must be locked when calling this function.
+func waitCondWithContext(cond *sync.Cond, ctx context.Context) error {
```
AudriusButkevicius, line 117: Probably should be CondLocked rather than Cond to follow other naming patterns.

lib/model/indexhandler.go
```diff
@@ -111,6 +111,42 @@ func newIndexHandler(conn protocol.Connection, downloads *deviceDownloadState, f
 	}
 }
 
+// waitCondLockedWithCtx waits for cond to signal, or for the context to finish.
+// If the context is cancelled, the cond will broadcast and an error will be returned.
+// cond must be locked when calling this function.
+func waitCondLockedWithCtx(cond *sync.Cond, ctx context.Context) error {
+	c := make(chan struct{}, 1)
```
imsodin, line 118: Whats the 1-buffering for?

imsodin (commented):
Good catch, thanks!

Suggestion:  
Instead of creating a goroutine and channel on every `Wait`, you could have a single goroutine in `Serve` that broadcasts on `ctx.Done()` and check for `ctx.Done()` inside the for-loop after waiting.

lib/model/indexhandler.go
yyogo: > Good catch, thanks!
> 
> Suggestion:
> Instead of creating a goroutine and channel on every `Wait`, you could have a single goroutine in `Serve` that broadcasts on `ctx.Done()` and check for `ctx.Done()` inside the for-loop after waiting.

Ah, good idea. will do!

> Whats the 1-buffering for?

Not really needed, you're right.

yyogo:
> Good catch, thanks!
> 
> Suggestion:
> Instead of creating a goroutine and channel on every `Wait`, you could have a single goroutine in `Serve` that broadcasts on `ctx.Done()` and check for `ctx.Done()` inside the for-loop after waiting.


Ah, good idea. will do!

lib/model/indexhandler.go
```diff
@@ -111,20 +111,42 @@ func newIndexHandler(conn protocol.Connection, downloads *deviceDownloadState, f
...
+func (s *indexHandler) waitForFileset(ctx context.Context) *db.FileSet {
+	s.cond.L.Lock()
+	defer s.cond.L.Unlock()
+
+	for s.paused && ctx.Err() == nil {
+		s.cond.Wait()
+	}
+	fset := s.fset
+
+	return fset
```
imsodin, line 124: This returns a `nil` fileset if the context is cancelled, without the caller knowing about it.

lib/model/indexhandler.go
```diff
@@ -111,20 +111,42 @@ func newIndexHandler(conn protocol.Connection, downloads *deviceDownloadState, f
...
+		close(stop)
+	}()
+
+	// Broadcast the pause cond when the context quits
+	go func() {
+		select {
+		case <-ctx.Done():
+			s.cond.L.Lock()
+			s.cond.Broadcast()
+			s.cond.L.Unlock()
```
imsodin, line 143: You don't need to lock to broadcast.

imsodin (changes requested):

imsodin (approved):

imsodin:
@calmh @AudriusButkevicius Should I add that to the current RC? It's late (release is on next Tuesday), however it's also a serious issue (with a simple fix).

calmh:
I'm fine with that

AudriusButkevicius:
Sounds good.

PR #7864, opened 2021-08-02
lib/config: Allow sub-second watcher delay (fixes #7859)

### LINKED ISSUE #7859: Allow sub-second watcher delay (use case: remote development)
I'm doing "remote development", where I edit sources locally, and they're compiled/executed on a remote, faster machine.

I'm currently using lsyncd. Syncthing can _almost_ cover this use case; unfortunately, it introduces unnecessary latency due to `fsWatcherDelayS`, which can't be set lower than one second. Depending on circumstances, this reduces or even cancels out the advantage of compiling/executing on the faster machine.

Ideally, it would be possible to specify a fractional `fsWatcherDelayS`, or an integer `fsWatcherDelayMs`.

tomasz1986:
Allow the watcher delay to take fractional values, effectively allowing
for much shorter delays. The minimum value is limited at 0.01, which
effectively translates to 10ms. This is required in order to guarantee
that there is still enough time to aggregate multiple single change
events.


imsodin:
No delay at all isn't ok. notify sometimes creates multiple events for a single change, at least those should be aggregated.

gsaraf:
@imsodin What is the minimum delay that would be fine in your opinion? Would 100ms be ok?

imsodin:
I just don't want == 0, as that means multiple triggered scans by design if there are duplicate events. Even back to back events in our tests are reported in <1ms, so a 10ms lower limit should be very save and that's already below human perception.

bard:
I opened the original #7859 and was going to add a comment, but @imsodin said it already. I originally suggested that `fsWatcherDelayS` could be changed to either fractional seconds or milliseconds so as to be able to set it very low, but leaving it as seconds and making it so a value of "0" actually translates to something like "10ms" is absolutely fine for the case I explained, requires (as far as I can see) minimal source changes, and no config migration.

tomasz1986:
I've given this a shot. `fsWatcherDelayS` stays as it was and keeps using seconds, but it gets converted to milliseconds later in the code. When set to `0`, a hard floor of 10ms is used instead.

lib/config/folderconfiguration.go
```diff
@@ -188,7 +188,7 @@ func (f *FolderConfiguration) prepare(myID protocol.DeviceID, existingDevices ma
 		f.RescanIntervalS = 0
 	}
 
-	if f.FSWatcherDelayS <= 0 {
+	if f.FSWatcherDelayS < 0 {
```
imsodin, line 191: Lets not change the semantics that `0` means default.

lib/watchaggregator/aggregator.go
```diff
@@ -438,8 +438,16 @@ func (a *aggregator) CommitConfiguration(from, to config.Configuration) bool {
 }
 
 func (a *aggregator) updateConfig(folderCfg config.FolderConfiguration) {
-	a.notifyDelay = time.Duration(folderCfg.FSWatcherDelayS) * time.Second
-	a.notifyTimeout = notifyTimeout(folderCfg.FSWatcherDelayS)
+	// Convert FSWatcherDelayS to milliseconds, and use a minimum delay
+	// of 10ms when set to zero to still have time to aggregate multiple
+	// single change events.
+	if folderCfg.FSWatcherDelayS == 0 {
```
imsodin, line 444:
```suggestion
	if folderCfg.FSWatcherDelayS < 0.01 {
```

imsodin (commented):
Without a rename `FSWatcherDelayS` needs to remain seconds. The option was to change it to float in the proto so it stays compatible with current values and one can specify fractions of a second.

bard:
Keeping in mind once again that I have limited experience with Go and even less with this codebase, so apologies if I'm not making any sense: I thought implementing that strategy could be limited to something like this:

```diff
  func notifyTimeout(eventDelayS int) time.Duration {
          shortDelayS := 10
          shortDelayMultiplicator := 6
          longDelayS := 60
          longDelayTimeout := time.Duration(1) * time.Minute
          if eventDelayS < shortDelayS {
                  return time.Duration(eventDelayS*shortDelayMultiplicator) * time.Second
          }
          if eventDelayS < longDelayS {
                  return longDelayTimeout
          }
-         return time.Duration(eventDelayS) * time.Second
+         if (eventDelayS === 0) {
+                 // Clamp at 10ms minimum
+                 return time.Duration(10) * time.Millisecond
+         } else {
+                 return time.Duration(eventDelayS) * time.Second
+         }
}
```

So basically just special-casing `eventDelayS == 0`.

lib/watchaggregator/aggregator.go
```diff
@@ -458,16 +466,16 @@ func updateInProgressSet(event events.Event, inProgress map[string]struct{}) {
 // air, they were just considered as a sensible compromise between fast updates and
 // saving resources. For short delays the timeout is 6 times the delay, capped at 1
 // minute. For delays longer than 1 minute, the delay and timeout are equal.
-func notifyTimeout(eventDelayS int) time.Duration {
-	shortDelayS := 10
+func notifyTimeout(eventDelayMs int) time.Duration {
```
calmh, line 469: This thing should take a `time.Duration` to begin with to reduce the unit conversion gymnastics. Then we get a common `a.notifyTimeout = notifyTimeout(a.notifyDelay)` after the if-clauses above.

tomasz1986:
I've modified the code to follow the suggestion by @bard, because it seems much simpler, while doing the same job.

@imsodin 
> Lets not change the semantics that `0` means default.

Is it really bad to just allow `0`? With the old code, trying to use `0` simply resets the value to the default `10`. It's not like you can have `0` actually set there in `config.xml`. With this commit, `0` would simply mean `0` (or 10ms in reality), and negative values would still reset the value to `10`. What do you think?

What I basically mean that `0` in this case is not the same as `0` with e.g. `maxFolderConcurrency`, where the actual value is dynamic or something like that.

imsodin:
> Is it really bad to just allow 0?

Maybe. Zero values and defaults are a minefield (if you are interested, some thoughts on that and other things are here: https://github.com/syncthing/syncthing/wiki/Configuration). Maybe this is fine, maybe it isn't - it doesn't matter because the actually simple and least surprising option is keeping the current behaviour, just allowing fractions of a second as a value. Simple in terms of just needing some special casing somewhere in the code is *not* a good criterium.

bard:
> Simple in terms of just needing some special casing somewhere in the code is _not_ a good criterium.

From my understanding, the value in question is meant to capture user intent, not a software contract.

Theoretically, if the value were part of a contract between different modules of syncthing, or between syncthing and other programs, I'd agree with you — if one module/program promises 500 X's and delivers 510 X's, it makes the job of the module/program on the other side more complicated. ("Theoretically" because, when dealing with time, the case isn't that clear-cut — distributed systems have little choice but to accept and plan for variability, or be fragile.)

When it comes to user intent, I think it's important to stay aware of what the number (or the string, or any vehicle of meaning) is supposed to encode. If I tap my phone's screen, and my fingertip ends up touching just one pixel below a hyperlink, where nothing clickable is present, is the browser's code doing me a favor and "special-casing" by activating the hyperlink, or is it fulfilling its contract with me, which consists in interpreting my intent correctly?

Ideally, user configuration values take intent into account as well. If I say, "sync at 10 seconds", I'm not really saying "sync at 10000ms" — 9500ms or 10500ms or anything in between captures my intent just as correctly. If I want to say "sync immediately", to say "sync at 0 seconds" captures my intent fine, too.

All IMHO, of course.

(Perhaps the crux here is just that a user value shouldn't wait too long and until it's too deep into the stack before intent is extracted from it?)

imsodin:
I incorrectly assumed the `0` value is documented to be default - it isn't. Nevertheless it is the case for some other config options and it has been the case for this one previously - I want to keep that behaviour. I don't want e.g. some script that used 0 to break. I don't understand why we are discussing this - there is a simple, backwards compatible solution, and I haven't heard any arguments against it yet.

bard:
@imsodin if you mean allowing fractions of a second as a value, it was also my first suggestion in [the original issue](https://github.com/syncthing/syncthing/issues/7859#issue-957243402), and [the argument against it](https://github.com/syncthing/syncthing/issues/7859#issuecomment-890390080) was the cost of migrating configuration.

imsodin:
https://github.com/syncthing/syncthing/issues/7859#issuecomment-890390682

> It might be mostly free to change the type to float and keep the seconds unit.

tomasz1986:
> It might be mostly free to change the type to float and keep the seconds unit.

I have tried to convert the value from int to float, and it does seem to work, but don't really have confidence regarding my modifications. I had to edit quite some code related to usage reporting, including migration, and I don't know whether it works or not.

Also, if I set the type to `float` in `folderconfiguration.proto`, it automatically uses `float32`, but there is no `sortFloats32` that could be used instead of `sortInts` in https://github.com/syncthing/syncthing/pull/7864/files#diff-dbcda7c99edad4fd4234bbc68e60511d0751bc10ea9a0e96618cfbca6eb73f95. I found https://gist.github.com/harihar/83670077ec532858ba1c390a65d3d91f and applied the solution from the stackoverflow answer linked there, but I'm not sure if this is the right way. On a side note, trying to use `float64` in `folderconfiguration.proto` results in complaints about it being not defined.

When it comes to generating proto itself, I used the precompiled binary mentioned in https://forum.syncthing.net/t/exec-protoc-executable-file-not-found-in-path/17307, so I also hope that the generated code is fine.

bard:
> [#7859 (comment)](https://github.com/syncthing/syncthing/issues/7859#issuecomment-890390682)
> 
> > It might be mostly free to change the type to float and keep the seconds unit.

Ah, fair enough, I didn't realize that addressed the objection, I thought the migration would be necessary regardless.

calmh:
`float64` is spelled `double` in protobuf, though I don't think the precision is essential. For usage reporting I think the simplest (though not 100% correct) solution is to simply round the value to an integer for now -- possibly rounding up, so that 0.1s gets reported as 1s (which is notably different from the default and from zero).

imsodin:
Looks like this needs a `go mod tidy` with go1.17 to build due to the new handling of indirect dependencies there.

tomasz1986:
> `float64` is spelled `double` in protobuf, though I don't think the precision is essential. For usage reporting I think the simplest (though not 100% correct) solution is to simply round the value to an integer for now -- possibly rounding up, so that 0.1s gets reported as 1s (which is notably different from the default and from zero).

Just to confirm, I should revert the usage reporting changes, keeping the value as `int`, and only do a conversion from `float32` to `int` with rounding up beforehand. Is this correct?

calmh:
That sounds reasonable to me, yes.

st-review:
:robot: *beep boop*

calmh (approved):

PR #7866, opened 2021-08-03
lib/connections: Always run a simple connection test

calmh:
### Purpose

Make sure connection establishment does not panic under normal circumstances.


AudriusButkevicius:
I assume this is a reaction to the arch thing? Or is there something more sinister?

imsodin (approved):
It's passing now :)  
I removed a few debug statements, the part about not supporting reuse port and one goroutine - @calmh can you check please that those changes make sense to you.

calmh:
LGTM!

PR #7869, opened 2021-08-03
lib/model: Don't consider hashes pulling on recv-enc

imsodin:
There is an issue reported about slowness/huge memory usage while syncing to an encrypted folder here: https://forum.syncthing.net/t/untrusted-device-never-finishes-syncing-very-high-memory-usage/17187  
The memory dump points at `HashFile` inside `handleFile`. We shouldn't ever hash files in a receive-encrypted folder, as hashes there aren't really hashes but just encrypted strings basically.

In this PR I factor out the parts that deal with hashes from `handleFile` and skip those if the folder is receive-encrypted. 

I have still no clue how this could cause such exorbitant memory usage.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -1075,57 +1075,27 @@ func (f *sendReceiveFolder) renameFile(cur, source, target protocol.FileInfo, sn
 func (f *sendReceiveFolder) handleFile(file protocol.FileInfo, snap *db.Snapshot, copyChan chan<- copyBlocksState) {
 	curFile, hasCurFile := snap.Get(protocol.LocalDeviceID, file.Name)
 
-	have, _ := blockDiff(curFile.Blocks, file.Blocks)
-
 	tempName := fs.TempName(file.Name)
 
 	populateOffsets(file.Blocks)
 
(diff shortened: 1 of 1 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -1075,57 +1075,27 @@ func (f *sendReceiveFolder) renameFile(cur, source, target protocol.FileInfo, sn
...
-	tempBlocks, err := scanner.HashFile(f.ctx, f.mtimefs, tempName, file.BlockSize(), nil, false)
-	if err != nil {
-		var caseErr *fs.ErrCaseConflict
-		if errors.As(err, &caseErr) {
-			if rerr := f.mtimefs.Rename(caseErr.Real, tempName); rerr == nil {
-				tempBlocks, err = scanner.HashFile(f.ctx, f.mtimefs, tempName, file.BlockSize(), nil, false)
-			}
-		}
+	if f.Type != config.FolderTypeReceiveEncrypted {
+		haveBlocks, _ := blockDiff(curFile.Blocks, file.Blocks)
```
calmh, line 1087: The blockDiff is fine though, for encrypted files.

calmh (approved):
Seems reasonable

PR #7873, opened 2021-08-05
lib/db/backend: Prevent panic in errors.As

imsodin:
https://sentry.syncthing.net/organizations/syncthing/issues/13738
https://sentry.syncthing.net/organizations/syncthing/issues/13739

### DIFF
--- a/lib/db/backend/backend.go
+++ b/lib/db/backend/backend.go
@@ -142,12 +142,12 @@ type errNotFound struct{}
 func (*errNotFound) Error() string { return "key not found" }
 
 func IsClosed(err error) bool {
-	var e *errClosed
+	e := &errClosed{}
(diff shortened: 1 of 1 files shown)


AudriusButkevicius (approved):

calmh:
There's some bad analysis here, this code is not the problem.

https://play.golang.org/p/GxBNFCZMoIx

imsodin:
I indeed jumped to conclusions here (based on other places in the code assigning a value to the underlying variable). The line number in `errors.As` also clearly points to something else: https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/errors/wrap.go;l=91  
And that looks like it should never be nil.

calmh:
🤷

PR #7895, opened 2021-08-19
build: Update x/sys to fix new issue in go 1.17

bn4t:
Updating x/sys so that this project supports go 1.17.

More context: https://github.com/Homebrew/homebrew-core/pull/83413


imsodin (approved):
Syncthing builds and runs on macos with go1.17, so that's not a problem for the linked PR (I commented there on what I suspect is: quic). However they also link to https://github.com/golang/go/issues/45702 which is a serious issue - thanks for that :)

PR #7898, opened 2021-08-19
lib/connections: Remove future go build constraints on quic

imsodin:
I checked and the quic folks already supported go1.17 since beta, so I'd say it's fairly save to stop guarding against future go releases (and forgetting to adjust it on every go release - I would have forgotten if not for the PR from the brew people). While at it I also bumped the lower constraint to go1.15 which is the oldest they support on the quic version we have (not that it matters much as we don't support go1.14 anyway).

### DIFF
--- a/lib/connections/quic_dial.go
+++ b/lib/connections/quic_dial.go
@@ -4,8 +4,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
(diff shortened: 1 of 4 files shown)


calmh (approved):

calmh:
Fully agree

AudriusButkevicius:
The build tags are confusing. Feels like its 1.15 only, but I assume its 1.15 and up?

calmh:
Yeah

PR #7564, opened 2021-04-11
build: Do not use --deb-systemd with fpm (fixes #7548)

imsodin:
Reuses the bits from https://github.com/syncthing/syncthing/pull/7552 that aren't related to nfpm (which requires go1.16), to fix #7548. Fixing as in not using `--deb-systemd` that was broken upstream, and doesn't get any attention, but restarting services in a script of our own. This is also a behaviour change, as `--deb-systemd` enables the service on install. I personally like that change: I dislike programs that enable their daemons automatically (thought that seems to be standard practice).

No testing yet: I was too lazy to install ruby or build/download the docker - will test after CI built something.

PR #7899, opened 2021-08-22
build: Readd systemd files to deb releases (ref #7564)

imsodin:
Systemd service files for stdiscosrv are missing: https://forum.syncthing.net/t/newer-builds-for-stdiscosrv/17247/3. Apparently the dysfunctional option removed in #7564 also took care of placing the service files, not only starting them.

I also moved all our units files into `/usr/lib/...` instead of `/lib`. We already do that for user system services files and I see no reason to distribute it in various different ones. Doesn't change anything for users, systemd looks in both places.


AudriusButkevicius (dismissed):

GermanCoding:
I fear that the move from /lib to /usr/lib this might break existing enabled services.

When enabling a service, systemd creates a symlink, e.g

`/etc/systemd/system/multi-user.target.wants/syncthing@syncthing.service → /lib/systemd/system/syncthing@.service`

This symlink is specifically pointed to /lib/... so I'm not sure what happens when the file moves from /lib/ to /usr/lib? Is the link updated automatically?

I believe this may not be the case, because just recently the stdiscosrv upgrade from apt.s.n to Debian official did something very similar and in that case, my symlinks broke - this caused quite a bit of havoc (service no longers starting, no longer able to view service status etc) until I disabled and then re-enabled said service (which first removed then re-created the symlink in the new location).

I haven't tested this change specifically, so I don't know for sure what's going to happen.

AudriusButkevicius (approved):

PR #7922, opened 2021-09-01
lib/model: Add missing error assignment + Remove unused argument

greatroar:
In (*model.sendReceiveFolder).handleSymlink, the error return from handleSymlinkCheckExisting wasn't being assigned, so the following check always got nil. Also removed an unused argument on a (*model.sendReceiveFolder).shortcutFile.

### DIFF
--- a/lib/model/folder_sendrecv.go
+++ b/lib/model/folder_sendrecv.go
@@ -387,7 +387,7 @@ func (f *sendReceiveFolder) processNeeded(snap *db.Snapshot, dbUpdateChan chan<-
 				// We are supposed to copy the entire file, and then fetch nothing. We
 				// are only updating metadata, so we don't actually *need* to make the
 				// copy.
-				f.shortcutFile(file, curFile, dbUpdateChan)
+				f.shortcutFile(file, dbUpdateChan)
 			} else {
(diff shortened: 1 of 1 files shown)


lib/model/folder_sendrecv.go
```diff
@@ -743,7 +743,7 @@ func (f *sendReceiveFolder) handleSymlink(file protocol.FileInfo, snap *db.Snaps
 		return
 	}
 
-	if f.handleSymlinkCheckExisting(file, snap, scanChan); err != nil {
+	if err = f.handleSymlinkCheckExisting(file, snap, scanChan); err != nil {
```
calmh, line 746:
```suggestion
	if err := f.handleSymlinkCheckExisting(file, snap, scanChan); err != nil {
```

is the more common spelling, you variant is equivalent of course but makes me starting looking for possible side effects of assigning to err in the outer scope

greatroar: I was just doing what the comment told me :)

calmh: Oh so there was a side effect, intended. I'm not sure that solution is beautiful exactly, but it's not your fault at least. :)

PR #7926, opened 2021-09-03
all: Unused args, retvals, assignments

greatroar:
Also two typos. These were all found by golangci-lint or DeepSource.

### DIFF
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -516,7 +516,11 @@ func checkUpgrade() (upgrade.Release, error) {
 }
 
 func upgradeViaRest() error {
-	cfg, _ := loadOrDefaultConfig(protocol.EmptyDeviceID, events.NoopLogger)
+	cfg, err := loadOrDefaultConfig(protocol.EmptyDeviceID, events.NoopLogger)
+	if err != nil {
+		return err
+	}
+
 	u, err := url.Parse(cfg.GUI().URL())
 	if err != nil {
 		return err
--- a/cmd/syncthing/main.go
+++ b/cmd/syncthing/main.go
@@ -412,7 +412,7 @@ func (options serveOptions) Run() error {
 }
 
 func openGUI(myID protocol.DeviceID) error {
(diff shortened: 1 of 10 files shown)


lib/db/transactions.go
```diff
@@ -714,7 +713,7 @@ func (t readWriteTransaction) updateGlobal(gk, keyBuf, folder, device []byte, fi
 			// The inserted file is the global file
 			global = file
 		} else {
-			keyBuf, global, err = t.getGlobalFromFileVersion(keyBuf, folder, name, true, globalFV)
+			keyBuf, global, _ = t.getGlobalFromFileVersion(keyBuf, folder, name, true, globalFV)
```
calmh, line 716: I suspect that we should in fact have some sort of error handling here, so the bug is larger than just unused retval

imsodin: Generally I want to have a good look at all this db changes, maybe there's some insights (bugs) to be discovered.

imsodin: On a second, slower look it all looks entirely benign. Just as noted here where a `if err != nil { return ..., err}` has been forgotten previously.

cmd/syncthing/main.go
```diff
@@ -516,7 +516,7 @@ func checkUpgrade() (upgrade.Release, error) {
 }
 
 func upgradeViaRest() error {
-	cfg, _ := loadOrDefaultConfig(protocol.EmptyDeviceID, events.NoopLogger, true)
+	cfg, _ := loadOrDefaultConfig(protocol.EmptyDeviceID, events.NoopLogger)
```
AudriusButkevicius, line 519: probably should have an error check?

lib/db/transactions.go
greatroar: Inserted an error check.

cmd/syncthing/main.go
greatroar: Done.

imsodin (approved):

PR #7931, opened 2021-09-07
lib/model, lib/protocol: Don't generate encrypted filenames that are reserved on Windows (fixes #7808)

### LINKED ISSUE #7808: Bug: Receive encrypted on Windows can have illegal file names.
Running Syncthing v1.17.0 on Windows and receiving encrypted data sometimes results in filenames that are not allowed on the platform:

image

This causes some data to be unsyncable until either the file is renamed (good luck finding out what file) or the encryption key is rotated (causing a full resync) and hoping you won;t hit the bug again.

calmh:
### Purpose

Avoid creating reserved filenames by trivial escaping.

### Testing

I don't know? Hence this is draft. How do we test this?

### Badness

This handles the practicalities, but not really all the metadata. If we already have announced a file under a given bad name, the next update will use a better name, but the old metadata will be around forever, causing a conflict if a new trusted device is set up and receives all metadata from the untrusted device. There will be two entries representing the same trusted-side file.

### DIFF
--- a/lib/model/model.go
+++ b/lib/model/model.go
@@ -1879,7 +1879,22 @@ func (m *model) Request(deviceID protocol.DeviceID, folder, name string, blockNo
 		// file has finished downloading.
 	}
 
-	if info, err := folderFs.Lstat(name); err != nil || !info.IsRegular() {
+	info, err := folderFs.Lstat(name)
+	if fs.IsNotExist(err) && folderCfg.Type == config.FolderTypeReceiveEncrypted {
+		// This might be a request for an encrypted file name with a Windows
+		// reserved path component that we've escaped on the other side, but
+		// the file we have on disk predates that escaping.
+		if unesc := protocol.UnescapeWindowsReserved(name); unesc != name {
+			if unescInfo, unescEerr := folderFs.Lstat(unesc); unescEerr == nil {
+				// The file exists under the old filename encoding. Rename
+				// it and carry on from there.
+				_ = folderFs.Rename(unesc, name)
+				info = unescInfo
+				err = nil
+			}
+		}
+	}
+	if err != nil || !info.IsRegular() {
 		// Reject reads for anything that doesn't exist or is something
 		// other than a regular file.
 		l.Debugf("%v REQ(in) failed stating file (%v): %s: %q / %q o=%d s=%d", m, err, deviceID, folder, name, offset, size)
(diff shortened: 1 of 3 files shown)


rasa:
@calmh Since users will never be opening encrypted files with Windows Explorer, there is no need to rename the files, as Windows can save them as is. We just have code that rejects them to allow users to easily open them. So this issue could be solved by simply removing the check the reserved name check for encrypted folders. So something like (untested):
https://github.com/syncthing/syncthing/blob/main/lib/fs/basicfs.go#L45 :
```
type BasicFilesystem struct {
	root            string
	junctionsAsDirs bool
	allowReservedFilenames bool
}

func (o *OptionAllowReservedFilenames) apply(fs Filesystem) {
	if basic, ok := fs.(*BasicFilesystem); !ok {
		l.Warnln("AllowReservedFilenames must only be used with FilesystemTypeBasic")
	} else {
		basic.allowReservedFilenames = true
	}
}

func (o *OptionAllowReservedFilenames) String() string {
	return "allowReservedFilenames"
}
```
https://github.com/syncthing/syncthing/blob/main/lib/config/folderconfiguration.go#L45 :
```
func (f FolderConfiguration) Filesystem() fs.Filesystem {
	// This is intentionally not a pointer method, because things like
	// cfg.Folders["default"].Filesystem() should be valid.
	var opts []fs.Option
	if f.FilesystemType == fs.FilesystemTypeBasic && f. {
		opts = append(opts, new(fs.OptionJunctionsAsDirs))
	}
	if f.FilesystemType == fs.FilesystemTypeBasic && f.Type == FolderTypeReceiveEncrypted {
		opts = append(opts, new(fs.OptionAllowReservedFilenames))
	}
```
and then move the reserved name check to inside the BasicFilesystem. Something along the lines of:
https://github.com/syncthing/syncthing/blob/019e0ce875d06efe70613e14575fb99352ca2e2f/lib/fs/basicfs.go#L396-L433

st-review:
:robot: *beep boop*

tomasz1986:
Is there any obstacle not to merge this PR?

The original bug can be very annoying, and there is no workaround, so it would be nice to have the fix implemented.

> Since users will never be opening encrypted files with Windows Explorer, there is no need to rename the files, as Windows can save them as is. We just have code that rejects them to avoid causing GUI users grief. So this issue could be solved by simply removing the reserved name check for encrypted folders.

The user will still be unable to delete the whole RE folder using Windows Explorer, so in my opinion, the reserved name check is needed here too.

imsodin:
I should finally look at @rasa pull request again. Even if we wouldn't want to use that mechanism in general, I think it's the right location to solve this problem: I don't particularly like windows weirdness and details being addressed in the protocol layer. That means all devices get to deal with it. Escaping it in the filesystem layer means only affected devices get to deal with it. And if something breaks or changes are needed later, only those devices are affected.

PR #7933, opened 2021-09-08
chore: Quiet numerous DeepSource reported issues

rasa:
### Purpose

Quiet DeepSource issues reported in
https://deepsource.io/gh/syncthing/syncthing/run/1f029f7e-c9e4-40a1-9b59-d1a8a4d39f1f/go/

### Testing

All tests pass.

### Comments

I chose to quiet things like
```
func (fs *errorFilesystem) Chmod(name string, mode FileMode) error { return fs.err }
```
by prefixing it with:
```
// skipcq: RVV-B0012 : parameter 'name' seems to be unused, consider removing or renaming it as _
```
as I didn't want to rewrite someone else's code as:
```
func (fs *errorFilesystem) Chmod(_ string, _ FileMode) error { return fs.err }
```
without direction.

### DIFF
--- a/lib/fs/basicfs_test.go
+++ b/lib/fs/basicfs_test.go
@@ -33,8 +33,7 @@ func TestChmodFile(t *testing.T) {
 	path := filepath.Join(dir, "file")
 	defer os.RemoveAll(dir)
 
-	// skipcq: GSC-G302 : Expect file permissions to be 0600 or less
-	defer os.Chmod(path, 0666)
+	defer os.Chmod(path, 0666) #skipcq
 
 	fd, err := os.Create(path)
 	if err != nil {
@@ -42,8 +41,7 @@ func TestChmodFile(t *testing.T) {
 	}
 	fd.Close()
 
-	// skipcq: GSC-G302 : Expect file permissions to be 0600 or less
-	if err := os.Chmod(path, 0666); err != nil {
+	if err := os.Chmod(path, 0666); err != nil { #skipcq
 		t.Error(err)
 	}
 
@@ -229,8 +227,7 @@ func TestDirNames(t *testing.T) {
 	sort.Strings(testCases)
 
 	for _, sub := range testCases {
-		// skipcq: GSC-G301 : Expect directory permissions to be 0750 or less
-		if err := os.Mkdir(filepath.Join(dir, sub), 0777); err != nil {
+		if err := os.Mkdir(filepath.Join(dir, sub), 0750); err != nil {
 			t.Error(err)
 		}
 	}
--- a/lib/fs/basicfs_test.go
+++ b/lib/fs/basicfs_test.go
@@ -33,6 +33,7 @@ func TestChmodFile(t *testing.T) {
 	path := filepath.Join(dir, "file")
 	defer os.RemoveAll(dir)
 
+	// skipcq: GSC-G302 : Expect file permissions to be 0600 or less
 	defer os.Chmod(path, 0666)
 
 	fd, err := os.Create(path)
@@ -41,6 +42,7 @@ func TestChmodFile(t *testing.T) {
 	}
 	fd.Close()
 
+	// skipcq: GSC-G302 : Expect file permissions to be 0600 or less
 	if err := os.Chmod(path, 0666); err != nil {
 		t.Error(err)
 	}
@@ -74,6 +76,7 @@ func TestChownFile(t *testing.T) {
 	path := filepath.Join(dir, "file")
 	defer os.RemoveAll(dir)
 
+	// skipcq: GSC-G302 : Expect file permissions to be 0600 or less
 	defer os.Chmod(path, 0666)
 
 	fd, err := os.Create(path)
@@ -226,6 +229,7 @@ func TestDirNames(t *testing.T) {
 	sort.Strings(testCases)
 
 	for _, sub := range testCases {
+		// skipcq: GSC-G301 : Expect directory permissions to be 0750 or less
 		if err := os.Mkdir(filepath.Join(dir, sub), 0777); err != nil {
 			t.Error(err)
 		}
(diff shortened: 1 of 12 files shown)


calmh (commented):
Whole I appreciate the effort you've put in to write this, I don't think we should add these magic comments. In my opinion they're ugly, and static checkers come and go. We've used several previously that are no longer a thing, and probably deep source will wander the same path. At that point I don't want hundreds of cryptic incantations to it "littered" all over the code.

I also think it's fine to let it complain about things that are just it's opinion, or not applicable in the condition it complains about it. The intention is for a human to consider it and draw a conclusion one way or the other.

rasa:
Understood. I will remove the comments from the PR, which will require code changes to quiet some of the issues.

While I appreciate your opinion, please consider:
1. the comments indicate a human has reviewed the item and determined it's not an issue.
1. having DeepSource saying most PRs are broken ( ❌ ) with sometimes hundreds of reported issues will discourage us from bothering to review its reports, as 99% of the time it's just noise, that we'll assume someone else has already reviewed.
1. wading thru pages of the same errors over and over again is not conducive to actually finding issues. Most users will simply not bother.
1. the comments can be stripped out with a single: `sed -i -Ee '|^\s*//\s*skipcq:|d; s|//\s*skipcq:.*||'`.
1. the comments can be added at the end of the line instead.

I couldn't find anything that allows us to disable reporting specific issues in the .deepsource.toml file. Most other linters have this feature, so maybe this functionality exists, but isn't well documented.

AudriusButkevicius:
Skimmed some of the comments, we should just fix the code where it makes sense, ie, not to shadow built ins, rename things to _ etc, oppose to adding comments to silence it.

rasa:
> Skimmed some of the comments, we should just fix the code where it makes sense, ie, not to shadow built ins, rename things to _ etc, oppose to adding comments to silence it.
> 
> And ignore the ones that make no sense.

Agreed, thanks. 

If I find a way to globally (or per file) silence specific issues that we may never want to see (say file permissions) I will propose it.

AudriusButkevicius:
To be honest, a good chunk of them are also noise, that we should just ignore and not have comments for.

rasa:
Ok, I removed all the full line comments, and fixed most of the issues in the code. I did add `//skipcq` in a few test files, *at the end of the line*. I also added a few in non-test files, to quiet "weak rand" and "func name differs only in case" issues.

If you want me to remove these comments too, lemme know if everywhere, or just in non-test files. Thanks.

lib/fs/basicfs_test.go
```diff
@@ -33,15 +33,15 @@ func TestChmodFile(t *testing.T) {
 	path := filepath.Join(dir, "file")
 	defer os.RemoveAll(dir)
 
-	defer os.Chmod(path, 0666)
+	defer os.Chmod(path, 0666) //skipcq
```
calmh, line 36: Rather than adding these I think we should just tell deepsource to ignore lax permissions in test files

lib/fs/basicfs_watch_test.go
```diff
@@ -126,17 +126,17 @@ func TestWatchRename(t *testing.T) {
 	name := "rename"
 
 	old := createTestFile(name, "oldfile")
-	new := "newfile"
+	new_ := "newfile"
```
calmh, line 129: Underscores are ugly; if we don't want to use "new" then something like "newName", "newFile" or whatever, please.

calmh (commented):
Replacing parameter names with underscores makes the code worse, in my opinion, except in a couple of specific exceptions.

For stub methods it's obvious that the parameters are unused, because it's a method stub without a body. Having the parameter names makes the code more natural and readable.

For methods implementing an interface the signature is decided by the interface and it *might* be useful to underscore a parameter to clearly indicate it's unused. Like, a ServeHTTP method that doesn't touch the request might have a `ServeHTTP(_ *http.Request, ...` because it's super clear what the parameter is and it might be useful info that we never touch it.

For Chtimes in fakefs for example I *don't* like the change because "_ time.Time, mtime time.Time` just tells me there's some other time passed that we apparently don't care about, but leaves it a mystery what that time is and why we ignore it. 

Guiding principle: we're writing code for humans to read -- Deepsource doesn't care, and it's wrong.

lib/fs/basicfs_test.go
AudriusButkevicius: Or move the permissions value to a variable

rasa: > For stub methods it's obvious that the parameters are unused, because it's a method stub without a body. Having the parameter names makes the code more natural and readable.

Agreed, but is `//skipcq` acceptable to quiet these? In test files only, or all files?

> For Chtimes in fakefs for example I _don't_ like the change because "_ time.Time, mtime time.Time` just tells me there's some other time passed that we apparently don't care about, but leaves it a mystery what that time is and why we ignore it.

Understood. Restore `atime` but add `//skipcq` to quiet?

> Rather than adding these I think we should just tell deepsource to ignore lax permissions in test files

Sorry, but how? DeepSource's docs show quieting specific lines of code, or ignoring entire files. I don't see a way to ignore specific errors for entire files. I've asked DeepSource if this is possible on their Slack #ideas channel.

> Or move the permissions value to a variable.

Good idea. We'll see if that works.

Note that our [contributing page](https://docs.syncthing.net/dev/contributing.html#tests) may need to be tweaked a bit, as it currently says `That the output from go lint and go vet is clean.` Based on your feedback, I can submit a PR to freshen this statement to clarify what is being asked of our developers.

lib/fs/basicfs_watch_test.go
rasa: Agreed, thx.

rasa:
Closing due to age. Please reopen if needed.

PR #7936, opened 2021-09-09
lib/model: Don't reset db while folder is running (fixes #7935)

### LINKED ISSUE #7935: Panic after POST /rest/system/reset
Logs:
```
[REDAC] 20:49:17 INFO: Cleaning data for folder "fored-acted"
[REDAC] 20:49:18 WARNING: Fatal error: fored-acted GetGlobal(ndock): device present in global list but missing as device/fileinfo entry
panic: device present in global list but missing as device/fileinfo entry
```

What I did:
```sh
HOST="redacted" CONFDIR=~/.config/syncthing APIKEY="$(~/.local/bin/xq -r '.configuration.gui.apikey' "$CONFDIR/config.xml")" curl --retry 10 --retry-delay 30 -s -H "Content-Type: application/json" -H "X-API-Key: $APIKEY" -X POST "https://$HOST/rest/system/reset?folder=fored-acted"
Bad Gateway # (it's behind a reverse proxy)
^C
```

 - Return a positive response like [202](https://http.cat/202): [Accepted](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202) before panicking.

imsodin:
Resetting the all db info on a folder while the same folder is running (and accessing its db state) is clearly not safe.


lib/model/model.go
```diff
@@ -2766,7 +2766,37 @@ func (m *model) BringToFront(folder, file string) {
...
+		return
+	}
+
+	// Stop the folder holding restart mutex to prevent any other concurrent
+	// changes from interfering.
+	restartMut := m.folderRestartMuts.Get(folder)
+	restartMut.Lock()
+	defer restartMut.Unlock()
+
+	m.RemoveAndWait(token, 0)
```
calmh, line 2787: This seems in principle unsafe, I see no reason that the folder isn't already restarted or reconfigured or removed after we release fmut and before we reach this point. Do we need to allow resetting a running folder at all? It seems to me like we could do all this under fmut, and return an error in case the folder is running...

imsodin: Right, the restart mutes protects from pausing, but one could still entirely remove/readd the folder at the same time. Giving the responsibility to pause the folder to the caller is a good idea :)

AudriusButkevicius (approved):

calmh (approved):

calmh:
Someone should try to figure out why depsource thinks you introduced 49 new antipatterns and 40 "bug risks" with this commit, at some point... It looks rather innocuous to me.

PR #7939, opened 2021-09-10
lib/config: Decouple VerifyConfiguration from Committer

greatroar:
... and remove 8/10 implementations, which were no-ops. This saves code and time copying configurations.

### DIFF
--- a/lib/config/wrapper.go
+++ b/lib/config/wrapper.go
@@ -31,14 +31,14 @@ const (
 
 var errTooManyModifications = errors.New("too many concurrent config modifications")
 
-// The Committer interface is implemented by objects that need to know about
-// or have a say in configuration changes.
+// The Committer and Verifier interfaces are implemented by objects
+// that need to know about or have a say in configuration changes.
 //
 // When the configuration is about to be changed, VerifyConfiguration() is
-// called for each subscribing object, with the old and new configuration. A
-// nil error is returned if the new configuration is acceptable (i.e. does not
-// contain any errors that would prevent it from being a valid config).
-// Otherwise an error describing the problem is returned.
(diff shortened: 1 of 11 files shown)


lib/config/wrapper.go
```diff
@@ -55,11 +55,16 @@ var errTooManyModifications = errors.New("too many concurrent config modificatio
...
-	VerifyConfiguration(from, to Configuration) error
 	CommitConfiguration(from, to Configuration) (handled bool)
 	String() string
 }
 
+// A Verifier can determine if a new configuration is acceptable.
+// See the description for Committer, above.
+type Verifier interface {
+	VerifyConfiguration(from, to Configuration) error
+}
```
AudriusButkevicius, line 66: This is probably missing String() which would be used printing the debug message.

Also, the fact that we do this magically concerns me, because its enough to change the argument type and things will stop working, so for this that do implement it, we should have a compile time assertion that they satisfy the interface.

AudriusButkevicius: Or alternative, provide a NoopVerifier that all thing embed (which I guess just brings us to existing behaviour)

greatroar: Type-asserting to Verifier doesn't remove the String method. Debugln passes its argument to Sprintln, which type-asserts to fmt.Stringer.

I put in the assertions, though.

calmh: I don't think the concern is about the String() method being removed or inaccessible, but that the code that runs the Verifier expects it to always be there so it could / should be part of the interface. However it's just used for debugging, so 🤷

calmh (approved):

PR #7943, opened 2021-09-12
lib/config: Check that marker is inside folder (fixes #7942)

### LINKED ISSUE #7942: Improve error message about folder marker outside of folder
A folder marker point at the folder root or even outside of it (e.g. `../foo/bar`) is accepted. This is unsafe, as it breaks the check against fully disappeared folder contents. *If* that check should be possible to be disabled, then that should be an explicit option (I don't think it should be).

imsodin:


lib/config/folderconfiguration.go
```diff
@@ -206,6 +207,15 @@ func (f *FolderConfiguration) prepare(myID protocol.DeviceID, existingDevices ma
 	if f.MarkerName == "" {
 		f.MarkerName = DefaultMarkerName
 	}
+	if f.MarkerName == "." {
+		l.Warnf(`Setting the folder marker name to "." is not allowed, reverting to default "%v"`, DefaultMarkerName)
+		f.MarkerName = DefaultMarkerName
+	}
+	markerPath := filepath.Join(f.Path, f.MarkerName)
+	if s := strings.TrimPrefix(markerPath, f.Path); s == markerPath || s == "" {
```
AudriusButkevicius, line 215: What about something like `foo/..` or `./.` etc?
Don't you need a filepath.Clean or something?

imsodin: `filepath.Join` already cleanes the result. Added those as testcases.

AudriusButkevicius (dismissed):

calmh (commented):
Are we certain we want to forbid dot as marker? It's worked for quite a while, and mostly does what it says on the tin as long as the folder itself is not a mount point. 

With this change folders already using dot as market will stop.

lib/config/folderconfiguration.go
```diff
@@ -206,6 +207,15 @@ func (f *FolderConfiguration) prepare(myID protocol.DeviceID, existingDevices ma
...
+		l.Warnf(`The folder marker name "%v" is not located inside the folder path, reverting to default "%v"`, f.MarkerName, DefaultMarkerName)
```
greatroar, line 216: %q may catch some configuration errors that "%v" doesn't.

tomasz1986:
> With this change folders already using dot as marker will stop...

Yeah, I don't personally use it, but the `.` marker seems to be the only way to sync read-only folders with dynamic content (i.e. there's no static file that could be used as one). Or is there perhaps another way to do this?

lib/config/folderconfiguration.go
imsodin: I don't understand. `%q` escapes various characters, I just want to put quotes around the string that may contain spaces.

tomasz1986:
For some reason, the error message is displayed twice.

image

imsodin:
Only a single folder? You could enable config, maybe model logging to see if e.g. config gets applied twice - however this issue really isn't worth that kind of attention.

I now added context to the logs about which folder it concerns, and tuned it down: It's no longer a warning but only info. The warning otherwise pops up whenever any config changes happen, which is plain annoying. The info log will likely not get noticed by users, but at least it's there if anyone posts logs when debugging something.

tomasz1986:
> Only a single folder? You could enable config, maybe model logging to see if e.g. config gets applied twice - however this issue really isn't worth that kind of attention.

Yeah, it's just a test instance that starts completely empty.

Here you go: [syncthing.log](https://github.com/syncthing/syncthing/files/7159505/syncthing.log)

imsodin:
The issue is, that `Config.prepare` is called twice: Once when deserializing json in the api and once on the config wrapper on modification. That inherently leads to double-logging, which should be addressed before adding any "nice-to-have" info logging like in this PR. Issue for that: https://github.com/syncthing/syncthing/issues/7957

lib/config/folderconfiguration.go
calmh: I think greatroars point is that with %q it'll be obvious if the string contains for example a unicode nonbreaking space instead of a normal space, etc, which in some theoretical situation could be confusing if there was a file on disk with a normal space and we're debugging why Syncthing doesn't see it...

PR #7945, opened 2021-09-13
gui: Sort already shared devices in edit folder modal (fixes #7940).

### LINKED ISSUE #7940: "Currently Shared With Devices" list in the folder Sharing tab should be sorted alphabetically
The list of devices under "Currently Shared With Devices" appears not to be sorted alphabetically. I'm not sure what the order is actually defined by right now. Is it by device ID?

image

On the other hand, the "Unshared Devices" list right below the current one is sorted alphabetically, so I believe that there is no reason not to sort this one alphabetically too. Not a huge deal, but it can become bothersome quickly if you have lots of devices.

acolomb:
### Purpose

The list of unshared devices is built from deviceList(), which sorts
the result.  But the shared devices are listed in undefined order from
the unsorted currentFolder.devices array.  Apply the same sorting
after building the array used in the dialog.

### Testing

Manual GUI verification.

### DIFF
--- a/gui/default/syncthing/core/syncthingController.js
+++ b/gui/default/syncthing/core/syncthingController.js
@@ -799,6 +799,7 @@ angular.module('syncthing.core')
                 }
                 $scope.currentSharing.selected[n.deviceID] = true;
             });
+            $scope.currentSharing.shared.sort(deviceCompare);
             $scope.currentSharing.unrelated = $scope.deviceList().filter(function (n) {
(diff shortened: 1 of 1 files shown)


calmh (approved):

StoyanDimitrov:
Don't know if desired or not, but it seems like the sorting is done case in insensitive maner:

acolomb:
Yes, actually case sensitive. All capital letters first, then all lowercase.  It's the same sorting as in other places.

Issue #7445, opened 2021-03-08
request: add windows/arm64 build for Windows on ARM


rubin55:
### Include required information
Really like its sister issue #7128, add windows/arm64 support and provide a download for Syncthing on Windows on ARM.

Go is very close to official build support and there's an official build that's also being used for Wireguard, see: 

https://github.com/golang/go/issues/36439
https://github.com/thongtech/go-windows-arm

rubin55:
Just a note for when you guys get to this. You will run into two issues:

1. go-ole: https://github.com/go-ole/go-ole/issues/202
2. goversioninfo: https://github.com/golang/go/issues/38020

I was able to fix 1. by simply creating the variant and variant-date files for arm with the correct build-directive. I was not able to fix the error I got with goversion info (`unknown ARM relocation type 7`). It should be something obvious though. The golang issue does not provide a fix (it's about someone building on Linux on ARM, where goversioninfo seems to be invoked unintendedly).

rasa:
https://github.com/go-ole/go-ole/pull/223 has been merged. Once https://github.com/shirou/gopsutil/pull/1132 is merged, we should be able to target windows/arm and windows/arm64.

calmh:
Builds with the OLE update:

https://build.syncthing.net/buildConfiguration/Syncthing_BuildWindows/110898?buildTab=artifacts&guest=1

Didn't look at what breakage the gopsutil stuff is.

PR #7947, opened 2021-09-15
Add windows/arm{,64} support

rasa:
### Purpose

See #7445 .

https://github.com/go-ole/go-ole/pull/223 has been merged. 
Once https://github.com/shirou/gopsutil/pull/1132 is merged, we should be able to target windows/arm and windows/arm64.

### DIFF
--- a/build.go
+++ b/build.go
@@ -733,7 +733,13 @@ func shouldBuildSyso(dir string) (string, error) {
 
 	sysoPath := filepath.Join(dir, "cmd", "syncthing", "resource.syso")
 
-	if _, err := runError("goversioninfo", "-o", sysoPath); err != nil {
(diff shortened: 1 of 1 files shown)


rasa:
Interesting that https://data.syncthing.net/ shows one Windows user already patched go-ole and gopsutil locally, and is running on arm.

rubin55:
> 
> 
> Interesting that https://data.syncthing.net/ shows one Windows user already patched go-ole and gopsutil locally, and is running on arm.

:-) That would be me! I've been running Syncthing on Windows on ARM (ARM 32-bit for about 6ish months now, see #7445).

AudriusButkevicius (approved):

PR #7962, opened 2021-09-24
lib/db: Replace SipHash with hash/maphash

greatroar:
SipHash was introduced in #6826 because maphash wasn't available in Go 1.13. Syncthing now requires 1.14, so we can drop one dependency. In Go 1.17, maphash is also significantly faster than SipHash (see /restic/restic#3519: restic also hashes SHA-256s).

The go.mod file has some unrelated changes from go mod tidy. Maybe there should be a CI test for that?

### DIFF
--- a/lib/db/lowlevel.go
+++ b/lib/db/lowlevel.go
@@ -12,18 +12,16 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"io"
+	"hash/maphash"
 	"os"
 	"regexp"
 	"time"
 
-	"github.com/dchest/siphash"
 	"github.com/greatroar/blobloom"
 	"github.com/syncthing/syncthing/lib/db/backend"
 	"github.com/syncthing/syncthing/lib/events"
 	"github.com/syncthing/syncthing/lib/fs"
 	"github.com/syncthing/syncthing/lib/protocol"
-	"github.com/syncthing/syncthing/lib/rand"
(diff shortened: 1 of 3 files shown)


lib/db/lowlevel.go
```diff
@@ -865,38 +863,36 @@ func (db *Lowlevel) recordIndirectionHashes(hs IndirectionHashesOnly) {
...
-// runs and colliding keys are not kept indefinitely.
+// The randomization in maphash should ensure that we get different collisions
+// across runs, so colliding keys are not kept indefinitely.
 func (b *bloomFilter) hash(id []byte) uint64 {
 	if len(id) != sha256.Size {
 		panic("bug: bloomFilter.hash passed something not a SHA256 hash")
 	}
-	return siphash.Hash(b.k0, b.k1, id)
+	var h maphash.Hash
+	h.SetSeed(b.seed)
```
calmh, line 893: Intuitively it seems like it would make sense to create the `maphash.Hash` once and store it in the `bloomFilter`, and just `Reset()` it before each use? We can even use a zero `maphash.Hash` as it'll seed itself on first use.

greatroar: That's not thread-safe.

calmh: No, that's fine, it's obvious from the syncfilter and whatnot if one actually reads the code.

calmh (approved):

PR #7964, opened 2021-09-25
lib/rand: Optimizations

greatroar:
rand.secureSource.Uint64 no longer allocates. rand.String uses a strings.Builder. Benchmark results on linux/amd64:

```
name            old time/op    new time/op    delta
SecureSource-8    69.1ns ± 3%    51.7ns ± 3%   -25.21%  (p=0.000 n=20+10)
String-8          2.66µs ± 2%    1.95µs ± 1%   -26.61%  (p=0.000 n=10+10)

name            old alloc/op   new alloc/op   delta
SecureSource-8     8.00B ± 0%     0.00B       -100.00%  (p=0.000 n=20+10)
String-8            288B ± 0%       32B ± 0%   -88.89%  (p=0.000 n=10+10)

name            old allocs/op  new allocs/op  delta
SecureSource-8      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=20+10)
String-8            33.0 ± 0%       1.0 ± 0%   -96.97%  (p=0.000 n=10+10)
```

### DIFF
--- a/lib/rand/securesource.go
+++ b/lib/rand/securesource.go
@@ -21,6 +21,7 @@ import (
 type secureSource struct {
 	rd  io.Reader
 	mut sync.Mutex
+	buf [8]byte
 }
 
 func newSecureSource() *secureSource {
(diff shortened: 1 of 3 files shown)


lib/rand/securesource.go
```diff
@@ -47,18 +46,18 @@ func (s *secureSource) Read(p []byte) (int, error) {
...
-
 	// Read eight bytes of entropy from the buffered, secure random number
 	// generator. The buffered reader isn't concurrency safe, so we lock
 	// around that.
 	s.mut.Lock()
-	_, err := io.ReadFull(s.rd, buf[:])
-	s.mut.Unlock()
+	defer s.mut.Unlock()
+
+	buf, err := s.rd.Peek(8) // Peek fills the buffer if needed.
```
calmh, line 55: While this is clever, could we not accomplish approximately the same thing by moving the `[8]byte` into the `secureSource` and the not have to depend on the reader being precisely a bufio.Reader and precisely how Peek/Discard word?

greatroar: Yes, at a minor cost in throughput:

```
name            old time/op    new time/op    delta
SecureSource-8    47.8ns ± 1%    51.7ns ± 3%  +8.21%  (p=0.000 n=18+10)
String-8          1.83µs ± 1%    1.95µs ± 1%  +6.62%  (p=0.000 n=10+10)
```

It's your call. To be sure, those comments are there to make some promises in the Peek/Discard descriptions explicit, they're not implementation details.

greatroar: Never mind. I just realized that Peek only ever does one Read, so this panics on intermittent failures. Pushed a version with a buffer in the secureSource.

calmh (approved):

Issue #7963, opened 2021-09-25
Improve filesystem detection


bt90:
https://forum.syncthing.net/t/syncthing-keeps-attempting-to-resync-files-that-have-not-changed/17359/10

The current implementation uses gopsutils `disk.Usage()` which tries to detect the filesystem using statfs on Unix systems. This has some quirks as noted by @tomasz1986 

Could we iterate over the mountpoints instead in order to get a better detection?

https://pkg.go.dev/github.com/shirou/gopsutil/disk#Partitions

calmh:
Could we not remove it entirely and put the responsibility of detecting android quirks on the android wrapper?

bt90:
I'm not even sure if this is a pure Andoid problem. Looks like statfs generally has some rough edges.

calmh:
Probably; but we only do the autodetection on android.

imsodin:
I agree: If there's something specifically useful/necessary for android, lets do that there. And if it's about generally improving fs detection, that should be a request in gopsutil - I don't think we want to start re-implementing that in syncthing.

PR #7966, opened 2021-09-25
lib/config: Always set mtime window to 2 on Android unless ext detected (ref #7963)

tomasz1986:
The current detection is flawed, because it looks for a few specific
file systems like "msdos" or "fat" to set the mtime window, while in
reality Android seems to report names like "fuseblk", which can stand
for fat, ext4, or even f2fs.

At the moment, we set the mtime window only for a few known names used
for the fat filesystem. With this change, we take a safer approach of
always setting the time window unless we explicitly detect file systems
like ext2/ext3/ex4, which are known not to experience issues with moving
timestamps on Android.

### DIFF
--- a/lib/config/folderconfiguration.go
+++ b/lib/config/folderconfiguration.go
@@ -62,7 +62,7 @@ func (f FolderConfiguration) ModTimeWindow() time.Duration {
 		if usage, err := disk.Usage(f.Filesystem().URI()); err != nil {
 			dur = 2 * time.Second
 			l.Debugf(`Detecting FS at "%v" on android: Setting mtime window to 2s: err == "%v"`, f.Path, err)
-		} else if strings.Contains(strings.ToLower(usage.Fstype), "ext2") || strings.Contains(strings.ToLower(usage.Fstype), "ext3") || strings.Contains(strings.ToLower(usage.Fstype), "ext4") {
+		} else if strings.HasPrefix(strings.ToLower(usage.Fstype), "ext2") || strings.HasPrefix(strings.ToLower(usage.Fstype), "ext3") || strings.HasPrefix(strings.ToLower(usage.Fstype), "ext4") {
 			l.Debugf(`Detecting FS at %v on android: Leaving mtime window at 0: usage.Fstype == "%v"`, f.Path, usage.Fstype)
 		} else {
 			dur = 2 * time.Second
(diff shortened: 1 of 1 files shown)


bt90:
What about f2fs?

tomasz1986:
> What about f2fs?

I don't think it's detected at all. My systems report either nothing, or `fuseblk` on the `/sdcard` partition. It will have to suffer the 2s mtime window, unfortunately. It likely does already now though (when detected as `""`).

lib/config/folderconfiguration.go
```diff
@@ -62,11 +62,11 @@ func (f FolderConfiguration) ModTimeWindow() time.Duration {
 		if usage, err := disk.Usage(f.Filesystem().URI()); err != nil {
 			dur = 2 * time.Second
 			l.Debugf(`Detecting FS at "%v" on android: Setting mtime window to 2s: err == "%v"`, f.Path, err)
-		} else if usage.Fstype == "" || strings.Contains(strings.ToLower(usage.Fstype), "fat") || strings.Contains(strings.ToLower(usage.Fstype), "msdos") {
+		} else if strings.Contains(strings.ToLower(usage.Fstype), "ext2") || strings.Contains(strings.ToLower(usage.Fstype), "ext3") || strings.Contains(strings.ToLower(usage.Fstype), "ext4") {
```
rasa, line 65: How about flipping it around to:
`if strings.Contains("ext2ext3ext4", strings.ToLower(usage.Fstype)) {`
?

bt90:
```
if strings.HasPrefix(strings.ToLower(usage.Fstype), "ext") {
```

tomasz1986: > How about flipping it around to:
> `if strings.Contains("ext2ext3ext4", strings.ToLower(usage.Fstype)) {`
> ?

This won't work, will it? What we're actually looking for is likely `ext2/ext3` and/or possibly `ext2/ext3/ext4`.

> `if strings.HasPrefix(strings.ToLower(usage.Fstype), "ext") {`

This probably would be fine, but I personally think it's still safer to look for the values specifically. Who knows what Android will report there in the future 😕. Another possibility could be go regex and use something like `^ext[2-4]`, but I'm not sure if there's any real benefit over the current method.

calmh (approved):
I would have gone with just looking for the `ext` prefix but I don't really care strongly. 👍

Issue #7965, opened 2021-09-25
Trash Can File Versioning always throws errors on restore (even when succeeded)


tomasz1986:
1. Add a new folder and share it between Device 1 and 2.
2. Add a new file to the folder on Device 1 and let it sync.
3. Enable Trash Can File Versioning on Device 1.
4. On Device 2, delete the file, and let it sync.
5. On Device 1, try to restore the file using the GUI.
   ![image](https://user-images.githubusercontent.com/5626656/134776671-be4ffd72-6b34-4328-9b2b-0786e5533bf3.png)
6. Despite the error, the file has actually been restored correctly.

AudriusButkevicius:
Can you hover on the error and see what the rest of the message says?

tomasz1986:
```
trashcan versioner: restore: rename \\?\R:\syncthing-test\folders\test1\.stversions\~syncthing~1.txt.tmp \\?\R:\syncthing-test\folders\test1\.stversions\1.txt: The system cannot find the file specified.
```

tomasz1986:
The issue seems to actually be very old.

I've done more testing, and it is present as soon as at least Syncthing v1.2.0. In the earlier versions, the GUI doesn't actually list any files in the Restore Versions modal, so I cannot test there.

tykisson:
New install of v1.18.4, Linux (64-bit ARM)
I can confirm that I receive the same error but then the file did appear on Device 1/2.

Going back to the versions list the file is no longer listed in version. Deleting the file from Device 1 does sync over to Device 2 and still does not appear on the versions list. The test file now appears to be lost.

tomasz1986:
I think this issue may be a duplicate of https://github.com/syncthing/syncthing/issues/6253. I'm not 100% sure though, because the error messages seem to differ, yet the behaviour itself seems identical, i.e. error on the screen, but working in reality.

Are the two issues about the exact same thing, or just similar?

AudriusButkevicius:
Seems they are about the same thing. I'll close the other one given this has more context.

tomasz1986:
I think the problem here is related to the logic behind

https://github.com/syncthing/syncthing/blob/30fa462e33f025b46e32d18d4a31b640ed23fd71/lib/versioner/trashcan.go#L66-L91

Basically, the error only shows up if the original file no longer exists. In other words, it occurs only when trying to restore a deleted file. On the other hand, if we try to restore a modification of a file that still exists in its original location, then there are no errors.

[...]

Issue #7393, opened 2021-02-21
Restore Versions modal extremely laggy and RAM-hungry with many items


tomasz1986:
1. Add 5,000 files to `.stversions`.
2. Open the Restore Versions modal.
3. It is game over. Loading the list takes a long time in itself, but the problem is that even after it has been loaded, the GUI is stuck. It is impossible to refresh or do anything else really. The CPU usage is high, and RAM can go up to as much as 1 GB or more. Also, scrolling the file list in the modal is super laggy too.

I have been trying to tweak the code, and have found out that removing

https://github.com/syncthing/syncthing/blob/1ed0116147b1c285df58fcec1930c6cf69917ab5/gui/default/syncthing/core/syncthingController.js#L2336-L2339

greatly helps with the lag. The RAM usage is still very high, but scrolling the list is much smoother, and the GUI does not get stuck anymore. Why exactly does Angular need to be forced to redraw here?

AudriusButkevicius:
Because angular has no idea that the tree structure underneath has changed, and the tree itself is not an angular component to notify about the change.

tomasz1986:
How exactly does it impact the Restore Version modal? I have tried to use it without that code, and restoring files seems to work the same.

AudriusButkevicius:
I don't remember exactly, but I suspect that without this, either the html does not match the data, or changes to the UI are not reflected in the model.

tomasz1986:
I've been playing with the Restore Versions modal a lot in the recent days, and what actually slows down everything is the insertion of version picker related HTML.

https://github.com/syncthing/syncthing/blob/198028d627952366d535024cf6fc4a4de07f5db7/gui/default/syncthing/core/syncthingController.js#L2426-L2430

If we either delete the template or simply zero the HTML files, the rendering becomes very fast. For instance, without the HTML insertion, it takes only ~3 seconds to render 10000 files (in Chromium). With the HTML insertion, the browser simply freezes and has to be killed to become usable again.

No solution yet, just sharing observations.

ryannathans:
I have ~35000 files in .stversions (staggered backups on ~5000 files) and it's **impossible** to restore files using the web UI (I have a 5950X CPU, plenty of RAM and it's still not helping)

All restores have to be done manually by SSHing into the syncthing instance and copying the file back to the working directory

PR #7972, opened 2021-09-26
gui: Tweak the Restore Versions modal for better usability

tomasz1986:
Add various tweaks to the Restore Versions modal in order to make it
more usable in general. In the process, upgrade both fancytree and
daterangepicker to their newest versions.

The important changes in details are as follows.

1. Wrap the table in a container, which is user-resizable, by default
   occupying as much space as needed within the screen. This prevents
   the generation of a giant list pushing everything down the screen,
   and also keeps the filter, date picker, and buttons visible.
2. Apply various usability tweaks, including support for long filenames,
   ensuring proper functionality on all resolutions, widening the input
   fields to fit more text, etc.
3. Upgrade the theme to match the rest of the Web GUI, including using
   the same set of icons. Ensure that everything works correctly in all
   supported themes.
4. Upgrade fancytree from 2.28.1 to 2.38.0, and daterangepicker from
   3.0.0 to 3.1. In the process, remove all pre-defined fancytree themes
   and create our own in css/assets instead. This is because it takes
   much less code to write a theme from scratch for our use case than to
   override the complex pre-defined one.

## Screenshots Before & After

<details>
<summary>Light Theme (desktop)</summary>

image
image
</details>

<details>
<summary>Light Theme (tablet)</summary>

</details>

<details>
<summary>Light Theme (phone)</summary>

</details>

<details>
<summary>Black Theme (desktop)</summary>

image
image
</details>

<details>
<summary>Black Theme (tablet)</summary>

</details>

<details>
<summary>Black Theme (phone)</summary>

</details>

<details>
<summary>Dark Theme (desktop)</summary>

image
image
</details>

<details>
<summary>Dark Theme (tablet)</summary>

</details>

<details>
<summary>Dark Theme (phone)</summary>

</details>


tomasz1986:
I still want to add some minor fixes, but I'd like to comment on a few things first.

1. The PR is large due to upgrading both fancytree and daterangepicker. I thought that it would be a good idea to upgrade them, because our versions were a few years old. These two components are only used in the Restore Versions modal anyway, so doing so won't affect anything else. There are no actual changes when it comes to the latter apart from a version bump, but in the case of fancytree, I have completely replaced the theme with a custom one that I believe fits Syncthing's UI better.
2. The calendar is still broken (see https://github.com/syncthing/syncthing/pull/7390) and also doesn't support the black and dark themes. I've tried to fix it without digging deep down the code, but couldn't really do much, so for now I'm planning to deal with it later in a separate PR.
3. There are still major performance issues with large numbers of files (see https://github.com/syncthing/syncthing/issues/7393), but this also will have to be addressed separately. This is fortunately limited mainly to cases, where we have a few thousands of files in single folder.

tomasz1986:
I'd say the commit is ready for review.

I've added a few more fixes, and also updated the screenshots to show the final version. I've simplified the container, removed some unneeded CSS, which resulted in getting rid of the inner border, and also gaining more horizontal space (which is critical here).

Also, as commented by @imsodin at https://github.com/syncthing/syncthing/pull/7395#discussion_r583587628, I've limited the container to be user-resizable only vertically. The reason is that if we allow horizontal resizing, then the user can drag the box as far as they can, and once that happens, it overflows the screen, and there is no way to go back, because the Web GUI disallows horizontal scrolling. Thus, I think it's just safer to keep the container in the modal. On a side note, the same probably should be applied to the Ignore Patterns textarea.

Just in case this PR is accepted, I'd like to ask, if possible, not to squash the two commits, and rather merge them separately. The reason is that there is a bug in the current version of fancytree that prevents folder child counters from being shown in IE11. I've reported the bug to the developer, but right now there's no fix, so I've applied a quick workaround myself. I think it would be better to keep this specific commit separate, so that if fancytree does get updated later, it will be easy to track the fix down, revert it, and then perform a clean upgrade.

st-review:
:robot: *beep boop*

calmh:
I'll take a look in the new year!

calmh (approved):
This seems fine to me and a significant improvement! Could I ask you to do a quick search for "restore" "GUI" type issues (they are legion) and see which of them are legitimately closed by this change?

tomasz1986:
Thank you for the positive feedback 🙂.

I've gone through all open issues and tried to gather everything related to Versioning:

- [Configurable restore path](https://github.com/syncthing/syncthing/issues/4670)
- [Date selector dialog on restore versions partially hidden](https://github.com/syncthing/syncthing/issues/4755)
- [Selection of time in versions GUI not possible without editing the string inside the textfield](https://github.com/syncthing/syncthing/issues/5408)
- [Restoring versions reports an error but actually works](https://github.com/syncthing/syncthing/issues/6253)
- ["Restore Versions" GUI breaks when path exists as both directory and file](https://github.com/syncthing/syncthing/issues/7068)
- [Restore Versions modal extremely laggy and RAM-hungry with many items](https://github.com/syncthing/syncthing/issues/7393)
- [Trash Can File Versioning always throws errors on restore (even when succeeded)](https://github.com/syncthing/syncthing/issues/7965)
- [Restore Versions file filter should have an option to disable case sensitivity](https://github.com/syncthing/syncthing/issues/7973)

I don't think this PR fixes any of them, unfortunately :slightly_frowning_face:. The issues are mostly about functionality, while the PR is mainly a visual overhaul. One component that isn't addressed here, but should be later, is the calendar, as it lacks support for themes and has issues on small screens, but I'm planning to work on it separately later.

All in all, I think this PR is ready as it is now, with no "ref" or "fixes" next to it.

calmh:
Then so be it! I thought it might have resolved some of the gui related ones but could pinpoint it. :) thanks for looking!

PR #7976, opened 2021-09-27
gui: Allow text wrapping in remove folder/device modals

### LINKED ISSUE #7975: Copy: Warning message inaccurate: Web interface: "Remove folder"
### Setup

 - syncthing v1.12.1-ds1 "Fermium Flea" (go1.15.4 linux-amd64) debian@debian 2021-02-04 16:26:39 UTC
 - Firefox 92.0

### Steps

- Web interface
- Folders
- Add folder
- Folder label "st-test-create-and-remove-dot-stfolder"
- Save
- `$ ls -a st-test-create-and-remove-dot-stfolder`
- `.stfolder`
- Web interface
- Click [st-test-create-and-remove-dot-stfolder]
- [Edit]
- [Remove]
- [Yes]

### Actual result

> Are you sure you want to remove folder st-test-create-and-remove-dot-stfolder?
> 
> No files will be deleted as a result of this operation.

### Expected result

> Are you sure you want to stop syncing st-test-create-and-remove-dot-stfolder with this device?
> 
> The folder itself will still exist

Or similar

It removes `.stfolder`, so saying that no files will be deleted seems at best inaccurate. =)

(This kind of copy: https://en.m.wikipedia.org/wiki/Copy_(written))

### Code

Was added into [default/syncthing/folder/removeFolderDialogView.html](https://github.com/syncthing/syncthing/blob/main/gui/default/syncthing/folder/removeFolderDialogView.html) by https://github.com/syncthing/syncthing/commit/205426a9c6bb3b60a06f4aece95fd0562cd9910e#diff-a484dbffd054b295d5730b6d812ae1094e71d3a2db114c064b47669f699c7e90R7

tomasz1986:
Remove the embedded CSS that prevents the lines from wrapping, resulting
in device and folder names being hidden on small screens.

### Screenshots

The old code prevents wrapping, which cuts the lines on mobile. I'd say this is very bad, because the folder path and/or device names end up being hidden.

### DIFF
--- a/gui/default/syncthing/device/removeDeviceDialogView.html
+++ b/gui/default/syncthing/device/removeDeviceDialogView.html
@@ -1,9 +1,11 @@
 <modal id="remove-device-confirmation" status="warning" icon="fas fa-question-circle" heading="{{'Remove Device' | translate}}" large="no" closeable="yes">
   <div class="modal-body">
-    <p ng-model="currentDevice.name" style="overflow: hidden; text-overflow: ellipsis; white-space: nowrap;">
-      <span translate translate-value-name="{{currentDevice.name}}">Are you sure you want to remove device {%name%}?</span>
+    <p ng-model="currentDevice.name" translate translate-value-name="{{currentDevice.name}}">
+      Are you sure you want to remove device {%name%}?
     </p>
-    <p ng-if="willBeReintroducedBy" translate translate-value-reintroducer="{{willBeReintroducedBy}}">{%reintroducer%} might reintroduce this device.</p>
+    <p ng-if="willBeReintroducedBy" translate translate-value-reintroducer="{{willBeReintroducedBy}}">
+      {%reintroducer%} might reintroduce this device.
+    </p>
   </div>
   <div class="modal-footer">
     <button type="button" class="btn btn-warning pull-left btn-sm" data-dismiss="modal" ng-click="deleteDevice()">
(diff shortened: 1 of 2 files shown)


gui/default/syncthing/device/removeDeviceDialogView.html
```diff
@@ -1,9 +1,11 @@
 <modal id="remove-device-confirmation" status="warning" icon="fas fa-question-circle" heading="{{'Remove Device' | translate}}" large="no" closeable="yes">
   <div class="modal-body">
-    <p ng-model="currentDevice.name" style="overflow: hidden; text-overflow: ellipsis; white-space: nowrap;">
-      <span translate translate-value-name="{{currentDevice.name}}">Are you sure you want to remove device {%name%}?</span>
+    <p ng-model="currentDevice.name" translate translate-value-name="{{currentDevice.name}}">
+      Are you sure you want to remove device {%name%} from Syncthing?
+    </p>
+    <p ng-if="willBeReintroducedBy" translate translate-value-reintroducer="{{willBeReintroducedBy}}">
+      {%reintroducer%} might reintroduce this device.
```
Jonta, line 7: I don't know what %reintroducer% is
- A device
- An account
- Either of those two
- Other

Will this be clear enough for the user? =)

calmh:
The fixes to wrapping are obviously good; the wording changes I'm not so sure about. I agree the new wording is a bit better, but I'm not sure it's better *enough* to warrant breaking translations for all languages.

tomasz1986:
> The fixes to wrapping are obviously good; the wording changes I'm not so sure about. I agree the new wording is a bit better, but I'm not sure it's better _enough_ to warrant breaking translations for all languages.

Yeah, but what should we do then? Should we close the issue and the PR as not really needed? I can move the wrapping fixes to a separate commit if that's the case.

calmh:
Yes, I think so, or just refine this to be about the layout changes. The existing "no files" ambiguity is acceptable to me.

tomasz1986:
Fixed the PR to cover only the wrapping fixes 👍 .

calmh (approved):

PR #8006, opened 2021-10-17
gui: Translate theme names in settings

acolomb:
### Purpose

Supersedes #7441 with a different approach.  Add each subdirectory of the guiDir as a translation candidate string.
The key is prefixed with "theme-name-" and the default English translation corresponds to the directory name turned to title case.  Disable the automatic name mangling in the GUI JS code in favor of just looking up the translation. 

### Testing

The second commit shows the effect of running `go run build.go translate` and manually adds a German translation for testing.  **It should be reverted before merging!**  Just let the automatic translation scripts handle it instead.

Verified to show translated names in German language setting and no visible change in the English setting.  Other, yet untranslated languages show the capitalized theme directory name.

### To-Do

* [ ] Check what happens for other languages in Transifex where there is no translation yet.  Will it copy the key or just apply the English translation as a fallback?
* [x] This will not work correctly for themes supplied in the `STGUIASSETS` directory.  Maybe a check could be added if the translated string is still prefixed with `theme-name-` and fall back to the old capitalization code in JS?

### DIFF
--- a/script/translate.go
+++ b/script/translate.go
@@ -147,7 +147,7 @@ func collectThemes(basePath string) {
 		if f.IsDir() {
 			key := "theme-name-" + f.Name()
 			if _, ok := trans[key]; !ok {
- 				name := strings.Title(f.Name())
+				name := strings.Title(f.Name())
 				trans[key] = name
 			}
 		}


calmh:
(I did see the note about reverting the manual translation changes, but chose not to worry about it as it'll get overwritten in the end.)

PR #8022, opened 2021-10-29
gui: Display identicons for discovered device IDs.

acolomb:
### Purpose

In the Add Device modal dialog, there is a list of up to five locally discovered device IDs.  Choosing from them is a great time-saver, but identifying which is the wanted one requires comparing (partial) device ID strings.  Especially when integrating a new device into an existing cluster, one can more easily remember the identicon shown on one of the existing cluster devices.  Make it easy to recognize that device in the discovered ID list by replacing the bullet points with the corresponding identicon.

### Testing

Open the Add Remote Device modal and admire the beautifully listed identicons.  Requires not yet added remote devices on the local network.

### Screenshots

Screenshot 2021-10-29 at 16-28-17 devtest Syncthing_add identicons

### DIFF
--- a/gui/default/syncthing/device/editDeviceModalView.html
+++ b/gui/default/syncthing/device/editDeviceModalView.html
@@ -25,7 +25,11 @@
               <p class="help-block" ng-if="discovery && discovery.length !== 0">
                 <span translate>You can also select one of these nearby devices:</span>
                 <ul>
-                  <li ng-repeat="id in discovery"><a href="#" ng-click="currentDevice.deviceID = id">{{id}}</a></li>
+                  <li ng-repeat="id in discovery" style="list-style-type: none;">
+                    <a href="#" ng-click="currentDevice.deviceID = id">
+                      <identicon data-value="id"></identicon>&nbsp;&nbsp;{{id}}
+                    </a>
+                  </li>
                 </ul>
               </p>
               <p class="help-block">


calmh (approved):
Neat

tomasz1986:
> one can more easily remember the identicon shown on one of the existing cluster devices.

Just a comment, and I don't know if it's only me, but to tell the truth, I can never remember the identicons 😅. I think I've learnt to simply ignore them when looking at my device list.

acolomb:
> Just a comment, and I don't know if it's only me, but to tell the truth, I can never remember the identicons.

Me neither, in the long term.  But just today I was looking at one device's list of remote devices and tried to add some of those to a new device, whose screen was 1 m away.  Sitting in front of the new device, I could glance at the other's screen and see the identicons.  That's when I wished I wouldn't have to jump back and forth to expand each remote device section to see its ID and match it against the discovered list.  With this change in place, one can stay put at the new device's screen :slightly_smiling_face:

Another use case is telling somebody else in the same room / on the phone: "Choose the one whose icon looks like a rocket." instead of "Click on the string starting with 9KJDF9-KJDF..."

tomasz1986:
I agree in the first point completely, but when it comes to

> Another use case is telling somebody else in the same room / on the phone: "Choose the one whose icon looks like a rocket." instead of "Click on the string starting with 9KJDF9-KJDF..."

I usually work with either computer-illiterate people or seniors, and frankly speaking, I have strong doubts anyone will be able to see a rocket (or any other distinguishable shape, as a matter of fact) in those icons 🤣.

calmh:
Well `GGOZ` above looks a little like a rocket. Or maybe a penis. `4RIUC` is a little dog. `BT452` is a space invader.

PR #8023, opened 2021-10-29
lib/model: Optimize jobQueue performance and memory use

greatroar:
By truncating time.Time to an int64 nanosecond count, we lose the ability to precisely order timestamps before 1678 or after 2262, but we gain (linux/amd64, Go 1.17.1):

```
name                      old time/op    new time/op    delta
JobQueuePushPopDone10k-8    2.85ms ± 5%    2.29ms ± 2%  -19.80%  (p=0.000 n=20+18)
JobQueueBump-8              34.0µs ± 1%    29.8µs ± 1%  -12.35%  (p=0.000 n=19+19)

name                      old alloc/op   new alloc/op   delta
JobQueuePushPopDone10k-8    2.56MB ± 0%    1.76MB ± 0%  -31.31%  (p=0.000 n=18+13)

name                      old allocs/op  new allocs/op  delta
JobQueuePushPopDone10k-8      23.0 ± 0%      23.0 ± 0%     ~     (all equal)
```

Results for BenchmarkJobQueueBump are with the fixed version, which no longer depends on b.N for the amount of work performed. rand.Rand.Intn is cheap at ~10ns per iteration.

The choice in Go 1.16, the current minimum required version, is between seconds and nanoseconds. When Go 1.17 becomes the minimum, the time.Time.UnixMicro or UnixMilli methods can be used instead to strike a different balance between timescale and precision.

### DIFF
--- a/lib/model/queue.go
+++ b/lib/model/queue.go
@@ -23,7 +23,7 @@ type jobQueue struct {
 type jobQueueEntry struct {
 	name     string
 	size     int64
-	modified time.Time
+	modified int64
 }
 
 func newJobQueue() *jobQueue {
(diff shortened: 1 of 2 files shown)


calmh:
Yeah deepsource can go pound sand in this case.

calmh (approved):

PR #8025, opened 2021-10-29
lib/syncthing: Clean up / refactor LoadOrGenerateCertificate() utility function.

acolomb:
Just some cleanup around the `LoadOrGenerateCertificate()` function.  It should adhere to the given arguments instead of delegating to the `locations` package.  This is technically a no-op change, because the only place calling this function passes exactly the same values.

I've chosen to keep the arguments although in principle the `locations` package already provides the desired paths.  Many other functions here take a path as argument and mix it with some `locations.Get()` values.  If you think the latter should be used consistently, I will switch over to that method as far as possible.

Also a no-op is to factor out the second half of that function and reuse it in `cmd/syncthing`.  That previously had its own independent (but matching) values for certificate name and lifetime, which is prone to get out of sync.

### DIFF
--- a/lib/syncthing/utils.go
+++ b/lib/syncthing/utils.go
@@ -27,14 +27,18 @@ import (
 func LoadOrGenerateCertificate(certFile, keyFile string) (tls.Certificate, error) {
 	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
 	if err != nil {
-		l.Infof("Generating ECDSA key and certificate for %s...", tlsDefaultCommonName)
-		return tlsutil.NewCertificate(
-			certFile, keyFile,
-			tlsDefaultCommonName,
-			deviceCertLifetimeDays,
-		)
+		return GenerateCertificate(certFile, keyFile)
 	}
 	return cert, nil
+}
+
+func GenerateCertificate(certFile, keyFile string) (tls.Certificate, error) {
+	l.Infof("Generating ECDSA key and certificate for %s...", tlsDefaultCommonName)
+	return tlsutil.NewCertificate(
+		certFile, keyFile,
+		tlsDefaultCommonName,
+		deviceCertLifetimeDays,
+	)
 }
 
 func DefaultConfig(path string, myID protocol.DeviceID, evLogger events.Logger, noDefaultFolder bool) (config.Wrapper, error) {
--- a/lib/syncthing/utils.go
+++ b/lib/syncthing/utils.go
@@ -34,11 +34,7 @@ func LoadOrGenerateCertificate(certFile, keyFile string) (tls.Certificate, error
 
 func GenerateCertificate(certFile, keyFile string) (tls.Certificate, error) {
 	l.Infof("Generating ECDSA key and certificate for %s...", tlsDefaultCommonName)
-	return tlsutil.NewCertificate(
-		certFile, keyFile,
-		tlsDefaultCommonName,
-		deviceCertLifetimeDays,
-	)
+	return tlsutil.NewCertificate(certFile, keyFile, tlsDefaultCommonName, deviceCertLifetimeDays)
 }
 
 func DefaultConfig(path string, myID protocol.DeviceID, evLogger events.Logger, noDefaultFolder bool) (config.Wrapper, error) {
--- a/lib/syncthing/utils.go
+++ b/lib/syncthing/utils.go
@@ -25,15 +25,11 @@ import (
 )
 
 func LoadOrGenerateCertificate(certFile, keyFile string) (tls.Certificate, error) {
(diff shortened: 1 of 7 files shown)


calmh (commented):
Good fix on the arguments, that's a bug. However, this seems like it could live in `tlsutil` as it's just a thin wrapper around another `tls` / `tlsutil` call?

AudriusButkevicius:
Also, while we are at it, could we separate the generation from persisting?

There was a number of times when I needed certs in tests, and the only way to get them, is to persist them on disk. There is no reason why this can't generate certs in memory, and then if needed, pass them to a different function to persist to disk.

acolomb:
> Good fix on the arguments, that's a bug. However, this seems like it could live in `tlsutil` as it's just a thin wrapper around another `tls` / `tlsutil` call?

I wouldn't necessarily move it, so that the `tlsDefaultCommonName` and `deviceCertLifetimeDays` constants can stay unexported members of the `lib/syncthing` package.  The same tlsutil function is used for the API cert, but with different lifetime and commonName settings.  So it makes sense to keep the wrappers in the corresponding package.

> Also, while we are at it, could we separate the generation from persisting?

Will look into it.  That makes the wrapper functions just a tiny bit less thin, then I'd keep them.

acolomb:
It's not quite easy to persist the key and cert to PEM files once they have been converted to a `tls.Certificate` structure AFAICS.  And passing around other data structures like `pem.Block` would make the utility functions less useful because they don't actually create a TLS certificate.  So I opted to just make the saving part optional by passing empty strings for `keyFile` and / or `certFile` and thus keep the `NewCertificate()` function API compatible.

If you like this approach, I can continue and remove the use of temporary files in tests where not needed.

lib/tlsutil/tlsutil.go
```diff
@@ -117,42 +117,52 @@ func NewCertificate(certFile, keyFile, commonName string, lifetimeDays int) (tls
...
+		if err = pem.Encode(keyOut, keyBlock); err != nil {
+			return tls.Certificate{}, errors.Wrap(err, "save key")
+		}
+		if err = keyOut.Close(); err != nil {
+			return tls.Certificate{}, errors.Wrap(err, "save key")
+		}
 	}
 
-	return tls.LoadX509KeyPair(certFile, keyFile)
+	return tls.X509KeyPair(pem.EncodeToMemory(certBlock), pem.EncodeToMemory(keyBlock))
```
AudriusButkevicius, line 165: I dislike the pass empty string to not save. Why can't it just be two methods? I guess you can't go from tls.Cert back to pem blocks?

acolomb: Yes, at least it wasn't obvious to me how to do that. Something happens with certificate chains and parsing and you cannot recover the original form. If you find a way, I'm all ears.

imsodin: It's about the call signature, where both empty strings and two methods isn't particularly nice, but I too would prefer the latter. Something like e.g.

```
func NewCertificate(certFile, keyFile string, commonName string, lifetimeDays int) (tls.Certificate, error)

func NewCertificateMemoryOnly(commonName string, lifetimeDays int) (tls.Certificate, error)
```

Or alternatively export `tlsutil.generateCertificate`, and have

```
func NewCertificate(certBlock, keyBlock *pem.Block) (tls.Certificate, error)

func SaveCertificate(certBlock, keyBlock *pem.Block) error
```
or possibly encapsulate the two `*pem.Block` variables in a type in tlsutil to not even expose that "implementation detail".

acolomb: I find it strange if a caller needs to call `GenerateCertificate()` and then `NewCertificate()` to actually use it.  Especially when the `tlsutil` package is obviously meant to make the process easier.

So I went with a separate function `NewCertificateInMemory()`. OK?

lib/syncthing/utils.go
```diff
@@ -25,22 +25,22 @@ import (
...
 	return cert, nil
 }
 
+func GenerateCertificate(certFile, keyFile string) (tls.Certificate, error) {
+	l.Infof("Generating ECDSA key and certificate for %s...", tlsDefaultCommonName)
+	return tlsutil.NewCertificate(
+		certFile, keyFile,
+		tlsDefaultCommonName,
+		deviceCertLifetimeDays,
+	)
```
imsodin, line 41: Please use the opportunity to remove the line-breaks - almost the entire code-base doesn't break up long lines, relying on editors for (soft) line-breaking.

acolomb: Done.

imsodin (approved):

acolomb:
Thanks @imsodin.

PR #8027, opened 2021-10-31
lib/api, lib/config: Remove config.ReadJSON (fixes #7957)

### LINKED ISSUE #7957: Consistently check config only once
Currently the config is checked through `.prepare` multiple times: Both when calling `ReadJSON` and when modifying the config in the wrapper. I propose to remove the checking on deserialization/reading and doing it only inside the wrapper (when wrapping or modifying/replacing a config). Needs some carefulness to ensure a "raw" config object loaded from disk isn't used anywhere before wrapping it. That usecase bould be handled by exposing `.prepare` as e.g. `CheckConfig(cfg *Configuration)`.

imsodin:
When posting a config through the api, `.prepare` is called twice, both in `ReadJSON` and when modifying the posted config in `config.Wrapper`. And `ReadJSON` does some default handling that `ReadXML` doesn't, and which duplicates logic in `lib/api/confighandler.go`. It's also used only there (and in tests) and all the other config objects don't have such a method -> removed it.

### DIFF
--- a/lib/api/confighandler.go
+++ b/lib/api/confighandler.go
@@ -289,7 +289,7 @@ func (c *configMuxBuilder) adjustConfig(w http.ResponseWriter, r *http.Request) 
 	var to config.Configuration
 	util.SetDefaults(&to)
 	if err := json.Unmarshal(bs, &to); err != nil {
-		http.Error(w, err.Error(), http.StatusBadRequest)
+		http.Error(w, fmt.Sprintf("unmarshalling config: %s", err), http.StatusBadRequest)
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


lib/api/confighandler.go
```diff
@@ -280,13 +280,41 @@ func (c *configMuxBuilder) registerGUI(path string) {
 }
 
 func (c *configMuxBuilder) adjustConfig(w http.ResponseWriter, r *http.Request) {
-	to, err := config.ReadJSON(r.Body, c.id)
+	bs, err := ioutil.ReadAll(r.Body)
```
greatroar, line 283:
```suggestion
	bs, err := io.ReadAll(r.Body)
```

See #7971.

calmh:
The Windows posting failure seems consistent, not sure why...

st-review:
:robot: beep boop

To permanently exempt this PR from bot nagging, add the `slow-pr` label.

Issue #8021, opened 2021-10-26
Provide a way to preset GUI credentials with password hashing


acolomb:
Syncthing's default configuration does not have GUI authentication set up.  As noted [on the forum](https://forum.syncthing.net/t/the-danger-zone-free-files-compliments-of-this-ignoramus/17378/7) and in https://github.com/SynoCommunity/spksrc/issues/4875, it would be advisable to set a user / password combination before first starting Syncthing.  Especially on headless systems (like NAS) which need to have a liberal GUI listen address and therefore just installing and starting Syncthing would open a big security hole without authentication.

Presetting the credentials is possible in principle through a barebones `config.xml`, but in practice turning a password into a bcrypt hash is not always easy.  Neither should any external tooling depend on Syncthing's internal config XML structure.

The suggestion is to add a new `syncthing cli` command, or a separate binary altogether, which accepts a user name and password from the command line or standard input.  It then patches that into a specified `config.xml` file, re-using the internal configuration handling code.  Ideally it should also indicate through its exit code whether the config file is valid and normalize it when writing.

The existing method through the REST API cannot be used for this, as it needs to happen before the background service is started.  At least with the SPK packages for a Synology NAS, possibly also in other deployment scenarios.

AudriusButkevicius:
I'd be wary of passing passwords on the commandline, given things like bash history etc. Reading it from stdin would be better.

acolomb:
> I'd be wary of passing passwords on the commandline, given things like bash history etc. Reading it from stdin would be better.

That is a first-class consideration I have already implemented in my prototype.  Will make a PR when it's cleaned up.

acolomb:
I'm a little bit stuck in my implementation when trying to actually modify the config.  If there is no file yet, `syncthing.DefaultConfig()` gives us a `config.Wrapper`.  I would do the same when it does exist by calling `config.Load()`.  But that wrapper doesn't really work for modifications unless I start its `Serve()` method with a new context etc.  That seems overly complicated all in all, just to open an XML file, modify two tags, and write it back.  Note this subcommand is just a simple, single-threaded script context where there are no services running and thus no Committers to take care of.

Do you think using the `ReadXML()` / `WriteXML()` functions directly for the generate subcommand would suffice?  Or is there a simpler way to actually make use of the `config.Wrapper`?

calmh:
Getting a wrapper, stating it's serve method and doing modification seem the cleanest solution to me.

acolomb:
[...]

PR #8028, opened 2021-10-31
lib/config: Move the bcrypt password hashing to GUIConfiguration

acolomb:
### Purpose

This is part of the solution for #8021, but makes sense as an isolated change.

What hash is used to store the password should ideally be an implementation detail, so that every user of the GUIConfiguration object automatically agrees on how to handle it.  That is currently distribututed over the confighandler.go and api_auth.go files, plus tests.

Add the `SetHasedPassword()` / `CompareHashedPassword()` API to keep the hashing method encapsulated.  Add a separate test for it and adjust other users and tests.  Remove all deprecated imports of the bcrypt package.

### Testing

`TestGUIPasswordHash()` added for the new config API.  Adjusted `lib/api/api_auth_test.go` to the different `authBasic()` function signature.  All tests still passing.

### DIFF
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -746,6 +746,9 @@ func TestGUIPasswordHash(t *testing.T) {
 	if err := c.HashAndSetPassword(testPass); err != nil {
 		t.Fatal(err)
 	}
+	if c.Password == testPass {
+		t.Error("Password hashing resulted in plaintext")
+	}
 
 	if err := c.CompareHashedPassword(testPass); err != nil {
 		t.Errorf("No match on same password: %v", err)
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -743,8 +743,7 @@ func TestGUIPasswordHash(t *testing.T) {
 	var c GUIConfiguration
 
 	testPass := "pass"
-	_, err := c.SetHashedPassword(testPass)
-	if err != nil {
+	if err := c.SetHashedPassword(testPass); err != nil {
 		t.Fatal(err)
 	}
 
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -739,6 +739,25 @@ func TestGUIConfigURL(t *testing.T) {
 	}
 }
 
+func TestGUIPasswordHash(t *testing.T) {
+	var c GUIConfiguration
+
+	testPass := "pass"
+	_, err := c.SetHashedPassword(testPass)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := c.CompareHashedPassword(testPass); err != nil {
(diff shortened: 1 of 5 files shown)


lib/config/guiconfiguration.go
```diff
@@ -113,6 +115,23 @@ func (c GUIConfiguration) URL() string {
 	return u.String()
 }
 
+// SetHashedPassword hashes the given plaintext password and stores the new hash.  A raw
+// copy of the updated hash bytes is returned as well.
```
calmh, line 119: I think it could return just the error?

calmh (commented):
Lgtm otherwise

lib/config/guiconfiguration.go
```diff
@@ -113,6 +115,22 @@ func (c GUIConfiguration) URL() string {
 	return u.String()
 }
 
+// SetHashedPassword hashes the given plaintext password and stores the new hash.
+func (c *GUIConfiguration) SetHashedPassword(password string) error {
```
AudriusButkevicius, line 119: SetHashed implies to accept an already hashed password.

HashAndSetPassword would be better?

lib/config/config_test.go
```diff
@@ -739,6 +739,24 @@ func TestGUIConfigURL(t *testing.T) {
...
+	var c GUIConfiguration
+
+	testPass := "pass"
+	if err := c.SetHashedPassword(testPass); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := c.CompareHashedPassword(testPass); err != nil {
+		t.Errorf("No match on same password: %v", err)
+	}
```
AudriusButkevicius, line 752: Should probably test c.Password != testPass

acolomb: Sorry, I'm confused what you mean.  Compare the hash and the plaintext?

AudriusButkevicius: If you made HashAndSetPassword and CompareHashedPassword completely noop (ie set the password as plaintext), the test would still pass.

acolomb: Okay understood.

lib/config/guiconfiguration.go
```diff
@@ -113,6 +115,22 @@ func (c GUIConfiguration) URL() string {
 	return u.String()
 }
 
+// SetHashedPassword hashes the given plaintext password and stores the new hash.
+func (c *GUIConfiguration) HashAndSetPassword(password string) error {
+	hash, err := bcrypt.GenerateFromPassword([]byte(password), 0)
+	if err == nil {
+		c.Password = string(hash)
+	}
+	return err
```
imsodin, line 124:
```suggestion
	if err != nil {
		return err
	}
	c.Password = string(hash)
	return nil
```

imsodin (approved):

PR #8037, opened 2021-11-03
Normalize CLI options to always use two dashes.

acolomb:
Consistently use double dashes for option flags, as the CLI help output does as well.

Fix typos:
 * `-conf` :arrow_right: `--config`
 * `-data-dir` :arrow_right:  `--data`
 * `-verify.` :arrow_right: `--verify-only`

Applies also to tests running the syncthing binary for consistency.  Other binaries (`stevents`, `stsigtool`, `build.go`, `stdisco`, `strelaysrv`) are not affected as they output single-dash options in their own CLI help.

### Testing

No functional changes, only affects help strings and uses options already parsed correctly in test, just now with two dashes.

### Documentation

Related: https://github.com/syncthing/docs/pull/692

### DIFF
--- a/cmd/syncthing/cli/main.go
+++ b/cmd/syncthing/cli/main.go
@@ -75,7 +75,7 @@ func Run() error {
 			Usage: "Set configuration and data directory",
 		},
 		cli.StringFlag{
-			Name:  "conf",
+			Name:  "config",
 			Value: "PATH",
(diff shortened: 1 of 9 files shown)


imsodin (approved):

acolomb:
Sorry I just found some more inconsistencies to squeeze in here.  Thanks for the review anyway.

calmh (approved):

PR #8039, opened 2021-11-06
lib/api: http.Request.BasicAuth instead of custom code

greatroar:
This is just cleanup, except that the stdlib method also recognizes case variants of the string "Basic " for closer conformance to RFC 7235.


calmh (approved):

calmh:
Newfangled convenience functions :)

PR #8042, opened 2021-11-07
lib/nat: Fix race condition in Mapping

greatroar:
The locking protocol in nat.Mapping was racy:

* Mapping.addressMap RLock'd, but then returned a shared map, so the lock didn't do anything.
* Operations inside Service.{verifyExistingMappings,acquireNewMappings} would lock the map for every update, but that means callers to Mapping.ExternalAddresses can be looping over the map while the Service methods are concurrently modifying it.

The Service methods now lock the map once and release the lock only when done.

Also, subscribers no longer get the added and removed addresses, because none of them were using the information. This was changed for a previous attempt to retain the fine-grained locking and not reverted because it simplifies the code.

### DIFF
--- a/lib/nat/service.go
+++ b/lib/nat/service.go
@@ -125,11 +125,15 @@ func (s *Service) process(ctx context.Context) (int, time.Duration) {
 
 	s.mut.RLock()
 	for _, mapping := range s.mappings {
-		if mapping.expires.Before(time.Now()) {
+		mapping.mut.RLock()
+		expires := mapping.expires
+		mapping.mut.RUnlock()
+
+		if expires.Before(time.Now()) {
 			toRenew = append(toRenew, mapping)
 		} else {
 			toUpdate = append(toUpdate, mapping)
-			mappingRenewIn := time.Until(mapping.expires)
+			mappingRenewIn := time.Until(expires)
 			if mappingRenewIn < renewIn {
 				renewIn = mappingRenewIn
 			}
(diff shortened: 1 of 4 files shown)


lib/nat/service.go
```diff
@@ -206,41 +206,39 @@ func (s *Service) RemoveMapping(mapping *Mapping) {
...
+		mapping.mut.Lock()
+		defer mapping.mut.Unlock()
 
-	newAdded, newRemoved = s.acquireNewMappings(ctx, mapping, nats)
-	added = append(added, newAdded...)
-	removed = append(removed, newRemoved...)
+		change = s.verifyExistingMappings(ctx, mapping, nats, renew)
+		add := s.acquireNewMappings(ctx, mapping, nats)
+		change = change || add
+	}()
```
imsodin, line 220: The intention here is to be able to `defer` the lock - or is there more to this function? Otherwise that seems a bit unnecessary to mean, compared to lock-unlock unclosing just 2-3 "plan" lines (no control flow shenanigans).

lib/nat/structs.go
```diff
@@ -26,55 +26,43 @@ type Mapping struct {
 	mut          sync.RWMutex
 }
 
+// Must be called with m.mut locked.
 func (m *Mapping) setAddress(id string, address Address) {
```
imsodin, line 30: The convention in Syncthing's code base is to suffix functions that must be called under a lock with `Locked`, makes it very obvious and hard to forget if calling from some place. Same applies to the verify and acquire functions above.

imsodin (commented):
There's also `Mapping.expires` which is accessed and modified without acquiring a lock in `Service` - at first glance that looks wrong. Might still be safe, but then it shouldn't be grouped with the `mut` in the declaration - did you check that?

lib/nat/structs.go
greatroar: Done, comments removed.

lib/nat/service.go
greatroar: Yes, I wanted to defer the unlock so it happens even when something panics (can happen for out-of-memory, in theory). Reverted to a simple lock-change-unlock.

greatroar:
I addressed your comments and fixed all accesses to expires. It can be written in Mapping.RemoveMapping, which is called from tcpListener.serve.

imsodin (approved):

lib/nat/service.go
```diff
@@ -206,41 +210,36 @@ func (s *Service) RemoveMapping(mapping *Mapping) {
...
+
+	if change || add {
+		mapping.notify()
 	}
 }
 
-func (s *Service) verifyExistingMappings(ctx context.Context, mapping *Mapping, nats map[string]Device, renew bool) ([]Address, []Address) {
-	var added, removed []Address
-
+func (s *Service) verifyExistingMappings(ctx context.Context, mapping *Mapping, nats map[string]Device, renew bool) (change bool) {
```
AudriusButkevicius, line 228: Shouldn't these also be "locked"?

greatroar: Done. I did s/Mappings/Locked/, because in the terminology of this package, each handles a single Mapping.

calmh (approved):

PR #8043, opened 2021-11-08
cmd/syncthing: Implement generate as a subcommand with optional API credential setting (fixes #8021)

### LINKED ISSUE #8021: Provide a way to preset GUI credentials with password hashing

acolomb:
### Purpose

The `--generate` option for the `syncthing serve` subcommand allows creating an initial configuration and keypair given a home / config directory.  This functionality is mirrored to a new subcommand `syncthing serve` with the following extensions:

1. Do not require a directory to be specified, but use the default location.
2. Accept the standard `--home` or `--config` options to override the default location, as used in the `serve` subcommand.
3. Accept `--gui-user=...` argument to preconfigure the generated config for GUI authentication.  This overwrites the setting should the `config.xml` file already exist.
4. Accept `--gui-password=...` argument to preconfigure the generated config for GUI authentication.  This overwrites the setting should the `config.xml` file already exist.  The password can be read from the standard input stream by specifying `-` as value.  Hashing the password is automatically taken care of.

Besides tidying up the command line structure, the main use case for this is setting up GUI credentials from an installation wizard on headless systems (like NAS), which otherwise would start out with the GUI wide open.

### Testing

Manual testing only: Generated new cert and config in empty home dir.  Adjusted GUI user and password.  Password read from pipe, standard input (terminated via EOF or line break), fails when empty password is piped in.  Existing config stays untouched if neither `--gui-...` option is given.  Giving the existing password hash as password again skips the modification.  Old `--generate` option still works unmodified.  Verified correct usage / help messages for both `syncthing serve` and `syncthing generate`.

### Screenshots

```
$ ./bin/syncthing generate --help
Usage: syncthing generate

Generate key and config, then exit

Flags:
  -h, --help                   Show context-sensitive help.

      --config=PATH            Set configuration directory (config and keys)
      --home=PATH              Set configuration and data directory
      --no-default-folder      Don't create the "default" folder on first startup ($STNODEFAULTFOLDER)
      --gui-user=STRING        Override GUI authentication user name
      --gui-password=STRING    Override GUI authentication password (use - to read from standard input)
```

### Documentation

See https://github.com/syncthing/docs/pull/694.

### DIFF
--- a/cmd/syncthing/generate/generate.go
+++ b/cmd/syncthing/generate/generate.go
@@ -110,38 +110,35 @@ func Generate(confDir, guiUser, guiPassword string, noDefaultFolder bool) error 
 	go cfg.Serve(ctx)
 	defer cancel()
 
-	guiCfg := cfg.GUI()
+	var updateErr error
 	waiter, err := cfg.Modify(func(cfg *config.Configuration) {
-		if changed := updateGUIAuthentication(&guiCfg, guiUser, guiPassword); changed {
-			cfg.GUI = guiCfg
-		}
+		updateErr = updateGUIAuthentication(&cfg.GUI, guiUser, guiPassword)
 	})
 	if err != nil {
 		return fmt.Errorf("modify config: %w", err)
 	}
 
 	waiter.Wait()
+	if updateErr != nil {
+		return updateErr
+	}
 	if err := cfg.Save(); err != nil {
 		return fmt.Errorf("save config: %w", err)
 	}
 	return nil
 }
 
-func updateGUIAuthentication(guiCfg *config.GUIConfiguration, guiUser, guiPassword string) bool {
-	changed := false
+func updateGUIAuthentication(guiCfg *config.GUIConfiguration, guiUser, guiPassword string) error {
 	if guiUser != "" && guiCfg.User != guiUser {
 		guiCfg.User = guiUser
 		log.Println("Updated GUI authentication user name:", guiUser)
-		changed = true
 	}
 
-	if guiPassword != "" && guiCfg.Password != guiPassword {
+	if guiPassword != "" {
 		if err := guiCfg.HashAndSetPassword(guiPassword); err != nil {
-			log.Fatal("Failed to set GUI authentication password.")
-		} else {
-			log.Println("Updated GUI authentication password.")
-			changed = true
+			return fmt.Errorf("Failed to set GUI authentication password: %w", err)
 		}
+		log.Println("Updated GUI authentication password.")
 	}
-	return changed
+	return nil
 }
--- a/cmd/syncthing/generate/generate.go
+++ b/cmd/syncthing/generate/generate.go
@@ -1,0 +1,62 @@
+// Copyright (C) 2021 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Package generate implements the `syncthing generate` subcommand.
+package generate
+
+import (
+	"crypto/tls"
+	"log"
+	"os"
+
+	"github.com/pkg/errors"
+
+	"github.com/syncthing/syncthing/lib/events"
+	"github.com/syncthing/syncthing/lib/fs"
+	"github.com/syncthing/syncthing/lib/locations"
+	"github.com/syncthing/syncthing/lib/protocol"
+	"github.com/syncthing/syncthing/lib/syncthing"
+)
+
+func Generate(confDir string, noDefaultFolder bool) error {
+	dir, err := fs.ExpandTilde(confDir)
+	if err != nil {
+		return err
+	}
+
+	if err := syncthing.EnsureDir(dir, 0700); err != nil {
+		return err
+	}
+	locations.SetBaseDir(locations.ConfigBaseDir, dir)
+
+	var myID protocol.DeviceID
+	certFile, keyFile := locations.Get(locations.CertFile), locations.Get(locations.KeyFile)
+	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
+	if err == nil {
+		log.Println("WARNING: Key exists; will not overwrite.")
+	} else {
+		cert, err = syncthing.GenerateCertificate(certFile, keyFile)
+		if err != nil {
+			return errors.Wrap(err, "create certificate")
+		}
+	}
+	myID = protocol.NewDeviceID(cert.Certificate[0])
+	log.Println("Device ID:", myID)
+
+	cfgFile := locations.Get(locations.ConfigFile)
+	if _, err := os.Stat(cfgFile); err == nil {
+		log.Println("WARNING: Config exists; will not overwrite.")
+		return nil
+	}
+	cfg, err := syncthing.DefaultConfig(cfgFile, myID, events.NoopLogger, noDefaultFolder)
+	if err != nil {
+		return err
+	}
+	if err := cfg.Save(); err != nil {
+		return errors.Wrap(err, "save config")
+	}
+	return nil
+}
--- a/cmd/syncthing/generate/generate.go
+++ b/cmd/syncthing/generate/generate.go
@@ -15,8 +15,6 @@ import (
 	"log"
 	"os"
 
-	"github.com/pkg/errors"
-
 	"github.com/syncthing/syncthing/cmd/syncthing/cmdutil"
 	"github.com/syncthing/syncthing/lib/config"
 	"github.com/syncthing/syncthing/lib/events"
@@ -61,7 +59,7 @@ func (c *CLI) Run() error {
 	}
 
 	if err := Generate(c.ConfDir, c.GUIUser, c.GUIPassword, c.NoDefaultFolder); err != nil {
-		return errors.Wrap(err, "Failed to generate config and keys")
+		return fmt.Errorf("Failed to generate config and keys: %w", err)
 	}
 	return nil
 }
@@ -85,7 +83,7 @@ func Generate(confDir, guiUser, guiPassword string, noDefaultFolder bool) error 
 	} else {
 		cert, err = syncthing.GenerateCertificate(certFile, keyFile)
 		if err != nil {
-			return errors.Wrap(err, "create certificate")
+			return fmt.Errorf("create certificate: %w", err)
 		}
 	}
 	myID = protocol.NewDeviceID(cert.Certificate[0])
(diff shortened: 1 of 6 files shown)


cmd/syncthing/generate/generate.go
```diff
@@ -0,0 +1,149 @@
...
+	defer cancel()
+
+	guiCfg := cfg.GUI()
+	waiter, err := cfg.Modify(func(cfg *config.Configuration) {
+		if changed := overrideGUIAuthentication(&guiCfg, guiUser, guiPassword); changed {
+			cfg.GUI = guiCfg
+		}
+	})
+	if err != nil {
+		return errors.Wrap(err, "modify config")
```
calmh, line 122:
```suggestion
		return fmt.Errorf("modify config: %w", err)
```

spelling of this changed in recent Go, I realize this is just copied/moved from the old code but might as well update in the process. (github.com/pkg/errors should not be used in new code)

(Same in all the other `errors.Wrap` calls, I just picked one at random.)

cmd/syncthing/generate/generate.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	go cfg.Serve(ctx)
+	defer cancel()
+
+	guiCfg := cfg.GUI()
+	waiter, err := cfg.Modify(func(cfg *config.Configuration) {
+		if changed := overrideGUIAuthentication(&guiCfg, guiUser, guiPassword); changed {
+			cfg.GUI = guiCfg
```
calmh, line 118: This assignment seems a no-op since overrideGUIAuthentication took a pointer a changed the object itself

cmd/syncthing/generate/generate.go
```diff
@@ -0,0 +1,149 @@
...
+		return errors.Wrap(err, "save config")
+	}
+	return nil
+}
+
+func overrideGUIAuthentication(guiCfg *config.GUIConfiguration, guiUser, guiPassword string) bool {
+	changed := false
+	if guiUser != "" && guiCfg.User != guiUser {
+		guiCfg.User = guiUser
+		log.Println("Overriding GUI authentication user name:", guiUser)
```
calmh, line 136: Minor nit, but "Overriding" feels to me like something is being temporarily overridden, not permanently changed. If we need to say anything at all in the success case I'd suggest something like "updated" or "changed".

calmh (commented):
Looks good

cmd/syncthing/generate/generate.go
acolomb: Hm. I believe the Wrapper's `GUI()` method returns a copy though, not the original `GUIConfiguration` object?  That's why we take its address here.  The same pattern can be found in `lib/api/confighandler.go` for example.

acolomb:
Regarding documentation, I could pack that into https://github.com/syncthing/docs/pull/693, or wait for that to be merged and make a separate PR.

cmd/syncthing/generate/generate.go
imsodin: I believe there's just no need for the `guiCfg` object assigned outside of `Modify`. You could just pass `&cfg.GUI` into `updateGUIAuthentication`.

cmd/syncthing/generate/generate.go
```diff
@@ -0,0 +1,147 @@
...
+
+func updateGUIAuthentication(guiCfg *config.GUIConfiguration, guiUser, guiPassword string) bool {
+	changed := false
+	if guiUser != "" && guiCfg.User != guiUser {
+		guiCfg.User = guiUser
+		log.Println("Updated GUI authentication user name:", guiUser)
+		changed = true
+	}
+
+	if guiPassword != "" && guiCfg.Password != guiPassword {
```
imsodin, line 138: This is comparing a hash against a plain-text password, isn't it?  
Probably the field should be named `Password` but `PasswordHash`.

lib/syncthing/utils.go
```diff
@@ -24,6 +24,29 @@ import (
...
+	fs := fs.NewFilesystem(fs.FilesystemTypeBasic, dir)
+	err := fs.MkdirAll(".", mode)
+	if err != nil {
+		return err
+	}
+
+	if fi, err := fs.Stat("."); err == nil {
+		// Apprently the stat may fail even though the mkdirall passed. If it
+		// does, we'll just assume things are in order and let other things
+		// fail (like loading or creating the config...).
```
imsodin, line 37: I know this isn't yours, but this sounds mysterious, and not in good way.

cmd/syncthing/generate/generate.go
acolomb: You're right on both points.  I personally think that renaming is not worth the effort.

The comparison is a special case, because I wanted to skip writing if the password didn't actually change.  But the hash changes randomly (salted), so this is the only case we can actually detect.  If you think it's not useful, we can drop it though.

cmd/syncthing/generate/generate.go
```diff
@@ -0,0 +1,147 @@
...
+		if err := guiCfg.HashAndSetPassword(guiPassword); err != nil {
+			log.Fatal("Failed to set GUI authentication password.")
```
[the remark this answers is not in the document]
acolomb: No, the `ModifyFunction` type in `lib/config` has no return value.  I'm not sure it's worth the effort to pass an `*Error` pointer to it somehow?  The boolean here is just used to optimize whether the `GUIConfiguration` replacement can be skipped.

lib/syncthing/utils.go
acolomb: Yeah, I just moved the code, didn't really understand this comment either.  Could we postpone examination and fixing to another PR?  I at least wouldn't know what circumstances to test and under what conditions such behavior may have been observed.  Might be worth keeping that commit separate so tracking that it was only moved is more obvious?

cmd/syncthing/generate/generate.go
acolomb: In this special case, nothing else is running that might modify the config in parallel.  But in the general case, the `GUI()` method makes a copy while holding the lock, to make sure it's an atomic snapshot.  Should we simplify here based on the single-thread assumption?

cmd/syncthing/generate/generate.go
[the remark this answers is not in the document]
imsodin: That's just a closure passed in. Getting the error out of that is as simple as declaring the error outside of the function and then assigning inside it.

cmd/syncthing/generate/generate.go
imsodin: What would you detect? The hash being equal to the plain-text password isn't something that should ever happen, and if it does then because someone manually added it to the config, in which case it should also be replaced.

cmd/syncthing/generate/generate.go
imsodin: The `cfg *config.Configuration` you get as an argument is already a copy - that would be really bad if that was a pointer to the actual config object in use.

cmd/syncthing/generate/generate.go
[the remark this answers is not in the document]
acolomb: Hm.  Maybe I just don't get the timing of it.  Should I check for that passed-out error after the `Wait()` call then?

cmd/syncthing/generate/generate.go
acolomb: I see your concern and that makes sense to avoid.  My thought was about the opposite case, supplying an already hashed password.

imsodin: Ah right, now I get it - if for some arcane reason the user gets the hash from config and passes it to this function, it should just leave the password alone. Yeah I guess why not.

cmd/syncthing/generate/generate.go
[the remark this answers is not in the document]
acolomb: Committed a simplified closure, please check if that's more to your taste.

cmd/syncthing/generate/generate.go
acolomb: > and if it does then because someone manually added it to the config, in which case it should also be replaced.

That case is really not nice though.  I'm undecided whether leaving a (wrongfully) unhashed password in the config when setting it again is wiser than trying to cater for this very hypothetical corner case...

acolomb:
Anything else I can do to get this across the finish line?  Currently waiting for some comment replies from @imsodin.

imsodin (approved):

cmd/syncthing/generate/generate.go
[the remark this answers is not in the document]
imsodin: It doesn't matter where you do it - `Wait` returns immediately if there's no change. I would have probably checked it right after checking `err`.

cmd/syncthing/generate/generate.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/syncthing/syncthing/lib/locations"
+	"github.com/syncthing/syncthing/lib/osutil"
+	"github.com/syncthing/syncthing/lib/protocol"
+	"github.com/syncthing/syncthing/lib/syncthing"
+)
+
+type CLI struct {
+	cmdutil.CommonOptions
+	GUIUser     string `placeholder:"STRING" help:"Specify new GUI authentication user name"`
+	GUIPassword string `placeholder:"STRING" help:"Specify new GUI authentication password (use - to read from standard input)"`
```
imsodin, line 31: Just noticed while reviewing the docs PR: Maybe we shouldn't allow to specify a pw as an argument at all, only stdin or enter it interactively.

cmd/syncthing/generate/generate.go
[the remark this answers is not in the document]
acolomb: I'm sure `Wait()` will return immediately, but `Modify()` may do so as well, before the `Serve()` goroutine has finished?  That's why I thought checking after `Wait()` is the only way to make sure the async function has completed running.

cmd/syncthing/generate/generate.go
acolomb: I think that would be surprising to most people.  It's easy to work around using `echo foo | ...`, but that has the same security implications (though if someone can read your shell history, they can just as well access your files directly, and I think non-interactive shells don't even save history).  So in an automated use-case, why let the user jump through that extra hoop?

Another UX point: When being asked for a password on the console, I'd expect it to not get echoed while typing.  This feature here is meant mainly for piping (`ssh-askpass | syncthing ...`).  So avoiding the complexities of terminal settings is the goal here.

acolomb:
Oh I just noticed merging this when the RC is already out maybe wasn't the best idea.  Sorry.

PR #8052, opened 2021-11-22
lib/config, lib/ignore: Write Windows line endings (fixes #7115)

### LINKED ISSUE #7115: Use CRLF instead of LF for new lines in config.xml and .stignore on Windows
Right now, Syncthing seems to always use LF for line endings in `config.xml` and `.stignore`. This is a problem in Windows when trying to edit those using the built-in Notepad, which until very recently had no support for other line endings than CRLF. Support for them has only been added in the very recent versions of Windows 10.

The problem is that when opening those files in Notepad under older (<= 1809?) versions of Windows 10 or any other older version of Windows for that matter (including also the newest Windows Server 2019), the content of those files is displayed in a long single line. This makes it basically impossible to review and edit.

I would like to suggest that these files should automatically revert to CRLF line endings when Syncthing is being run under Windows.

Ref: https://devblogs.microsoft.com/commandline/extended-eol-in-notepad

calmh:
Write config.xml and .stignore with appropriate line endings.

### DIFF
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -595,6 +595,36 @@ func TestNewSaveLoad(t *testing.T) {
 	}
 }
 
+func TestWindowsLineEndings(t *testing.T) {
+	if runtime.GOOS != "windows" {
+		t.Skip("Windows specific")
+	}
+
+	path := "testdata/temp.xml"
+	os.Remove(path)
+	defer os.Remove(path)
+
+	intCfg := New(device1)
+	cfg := wrap(path, intCfg, device1)
+	defer cfg.stop()
+
+	err := cfg.Save()
+	if err != nil {
+		t.Error(err)
+	}
+
+	bs, err := os.ReadFile(path)
+	if err != nil {
+		t.Error(err)
+	}
+
+	unixLineEndings := bytes.Count(bs, []byte("\n"))
+	windowsLineEndings := bytes.Count(bs, []byte("\r\n"))
+	if unixLineEndings == 0 || windowsLineEndings != unixLineEndings {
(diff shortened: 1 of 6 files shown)


lib/ignore/ignore_test.go
```diff
@@ -1193,3 +1193,36 @@ func TestEmptyPatterns(t *testing.T) {
...
+	}
+
+	lines := "foo\nbar\nbaz\n"
+
+	ffs := fs.NewFilesystem(fs.FilesystemTypeBasic, ".")
+	m := New(ffs)
+	if err := m.Parse(strings.NewReader(lines), ".stignore"); err != nil {
+		t.Fatal(err)
+	}
+	if err := WriteIgnores(ffs, ".stignore", m.Lines()); err != nil {
```
imsodin, line 1209: As far as I see this creates an `.stignore` in lib/ignores and doesn't remove it. Generally I'd also prefer `ioutil.TempDir`.

lib/config/config_test.go
```diff
@@ -595,6 +595,36 @@ func TestNewSaveLoad(t *testing.T) {
...
 }
 
+func TestWindowsLineEndings(t *testing.T) {
+	if runtime.GOOS != "windows" {
+		t.Skip("Windows specific")
+	}
+
+	path := "testdata/temp.xml"
+	os.Remove(path)
+	defer os.Remove(path)
```
imsodin, line 605: I'd prefer `ioutil.TempDir`.

imsodin (commented):
Looks good, one comment about a possibly not deleted file in a unit test.

imsodin (approved):

PR #8053, opened 2021-11-22
meta: Add test for deprecated package ioutil

calmh:
Forbid usage of deprecated package `ioutil`.

### DIFF
--- a/meta/forbidden_words_test.go
+++ b/meta/forbidden_words_test.go
@@ -1,0 +1,52 @@
+// Copyright (C) 2021 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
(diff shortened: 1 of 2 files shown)


meta/forbidden_words_test.go
```diff
@@ -0,0 +1,52 @@
+// Copyright (C) 2021 The Syncthing Authors.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Checks for authors that are not mentioned in AUTHORS
```
imsodin, line 7: not here :)

imsodin (approved):

meta/forbidden_words_test.go
calmh: And not even where I copied it from...

imsodin (approved):

